PR #2577, opened 2026-07-20
remove paragraph marker tooltip from power

Sebastian-ubs:
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2577)

### DIFF
--- a/extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx
+++ b/extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx
@@ -1804,28 +1804,34 @@ globalThis.webViewComponent = function PlatformScriptureEditor({
       );
     }
 
+    const editorTree = (
+      <EditorKeyboardShortcuts editorRef={editorRef}>
+        <Editorial
+          ref={editorRef}
+          scrRef={scrRef}
+          onScrRefChange={setScrRefNoScroll}
+          options={options}
+          logger={logger}
+          onUsjChange={isReadOnly ? undefined : handleEditorialUsjChange}
+          onSelectionChange={handleSelectionChange}
+          onStateChange={(state) => {
+            setCanUndo(state.canUndo);
+            setCanRedo(state.canRedo);
+            setBlockMarker(state.blockMarker);
+            setContextMarker(state.contextMarker);
+          }}
+        />
+      </EditorKeyboardShortcuts>
+    );
+
     return (
       <>
         {workaround}
-        <ParagraphMarkerTooltipOverlay>
-          <EditorKeyboardShortcuts editorRef={editorRef}>
-            <Editorial
-              ref={editorRef}
-              scrRef={scrRef}
-              onScrRefChange={setScrRefNoScroll}
-              options={options}
-              logger={logger}
-              onUsjChange={isReadOnly ? undefined : handleEditorialUsjChange}
-              onSelectionChange={handleSelectionChange}
-              onStateChange={(state) => {
-                setCanUndo(state.canUndo);
-                setCanRedo(state.canRedo);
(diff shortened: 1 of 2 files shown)


tjcouch-sil (approved):
Thanks!

extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx
```diff
@@ -1804,28 +1804,34 @@ globalThis.webViewComponent = function PlatformScriptureEditor({
...
-              onStateChange={(state) => {
-                setCanUndo(state.canUndo);
-                setCanRedo(state.canRedo);
-                setBlockMarker(state.blockMarker);
-                setContextMarker(state.contextMarker);
-              }}
-            />
-          </EditorKeyboardShortcuts>
-        </ParagraphMarkerTooltipOverlay>
+        {isPowerMode ? (
```
lyonsil, line 1830: **Editor can remount and blank out because the overlay wrapper is gated on `isPowerMode`**

Gating `ParagraphMarkerTooltipOverlay` on `isPowerMode` puts a *different React element type* at this fragment slot depending on a live setting value. When `isPowerMode` transitions while the editor is past its loading guards, React unmounts/remounts the whole `EditorKeyboardShortcuts > Editorial` subtree — losing undo/redo history, selection, cursor, and scroll. Worse, the remounted `Editorial` stays **blank** until the next navigation: it takes no `usj` prop and is only seeded via `editorRef.setUsj()` from `useEditorPdpSync`, whose effect doesn't re-run when `usjFromPdp` is unchanged (a mode flip doesn't change it). This is exactly the remount the reverse portal at `:1973` exists to prevent.

Reachability I checked:

- **Runtime mode switch** is mostly moot — flipping `platform.interfaceMode` reloads the whole dock layout and recreates the editor web view (`simpleLayout`'s hardcoded id `3cf575f0` isn't in the saved/`testLayout` power layout), so `isPowerMode` never flips in a surviving instance.
- **Mount-time race** is the live path: `useSetting('platform.interfaceMode', 'simple')` returns the default `'simple'` first, then updates to `'power'`. A power-mode editor whose `usjFromPdp` resolves before the setting does mounts in the simple (overlay) branch, seeds content, then remounts blank when the setting flips. Narrow (the setting is warm-cached, USJ is a cold fetch) but real — and power mode is still WIP, so it would ship latent.

Suggested fix — keep the overlay always mounted and gate the tooltip via a prop, so the editor's element type/position never changes with the mode:

```tsx
return (
  <>
    {workaround}
    <ParagraphMarkerTooltipOverlay enabled={!isPowerMode}>{editorTree}</ParagraphMarkerTooltipOverlay>
  </>
);
```

(the overlay keeps its same root `<div className="tw:relative">` in both modes and just skips the tooltip + listeners when `enabled` is false).

This one change also clears the **4 other issues** the automated review flagged: the "type into a blanked editor overwrites the chapter" data-integrity path, the policy-leaked-to-call-site altitude nit, the inline nested-ternary style inconsistency, and the now-stale `_editor-overrides.scss` wrapper-chain comment (the overlay is present in both modes again).

lyonsil (commented):
> @lyonsil reviewed 1 file and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2577)* status: all files reviewed, 1 unresolved discussion (waiting on Sebastian-ubs).

lyonsil (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2577#-:-P10VgFEEE_GddZI645b:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on Sebastian-ubs).

PR #2579, opened 2026-07-21
Speed up startup by rendering project pickers from metadata

lyonsil:
# Code Review Summary

**Branch**: worktree-startup-perf-investigation

**Base**: origin/main

**Date**: 2026-07-20

**Files changed**: 52

## Overview

This branch improves Platform.Bible startup time. The author first instrumented the startup path
(a gated `PT_STARTUP_MARKS` flag, per-process `markStartup` primitives, and a `startup-waterfall`
CLI parser) and found the dominant cost was opening every Paratext project to read its `Settings.xml`
just to enumerate a handful of display fields for the project pickers. The fix carries those display
fields (`name`, `fullName`, `language`, `languageTag`, `isEditable`, `isPublished`) on
`ProjectMetadata` and populates them cheaply during project enumeration via
`ScrTextExtensions.GetRawParatextSetting` — reading the raw settings parameter dictionary directly
instead of invoking the real per-project getters, whose side effects (LDML parse for `languageTag`,
settings persistence, expensive object construction) were the actual expense. The renderer project
picker and the Home/New-Tab web views then render from a single filtered `getMetadataForAllProjects`
fetch instead of an N×5 per-project `pdp.getSetting` fan-out. A new `platform.onDidChangeProjects`
network event keeps those surfaces live when the project set changes. Measured result: ~5-6 seconds
saved on the author's dev machine.

The changes are additive at the public-API surface, keep the TS↔C# contract in sync, and are
documented to preserve the previously-rendered output. The substitution's correctness rests on a set
of deliberate, documented divergences from the real getters (see Suggested Review Focus).

## API Changes

- **`@papi/core`** (via `shared/models/project-metadata.model`): added six optional fields to
  `ProjectMetadataWithoutFactoryInfo` (re-exported as `ProjectMetadata` /
  `ProjectMetadataWithoutFactoryInfo`): `name?`, `fullName?`, `language?`, `languageTag?`,
  `isEditable?`, `isPublished?` — all `string`/`boolean`.
- **`papi-shared-types` (`NetworkEvents`)**: added event `'platform.onDidChangeProjects': undefined`
  (no payload).
- **`shared/models/project-lookup.service-model`**: added export
  `normalizeProjectId(projectId: string): string`.
- **`node/utils/util`**: added export `isStartupMarksEnvVariableSet: () => boolean`.
- **`shared/data/platform.data`**: added constants `STARTUP_MARKS_QUERY_PARAMETER`,
  `STARTUP_MARK_PREFIX`, `STARTUP_MARK_PROCESS_START`, `STARTUP_MARK_MAIN_PROCESS_TAG`.
- **`shared/global-this.model`**: added global `var startupMarks: boolean`.
- No changes to `lib/platform-bible-react/`, `lib/platform-bible-utils/`, or any extension
  `types/*.d.ts`. All changes are additive (new optional fields, new exports, a new no-payload
  event); `git diff` confirms zero removed or modified export lines.

None.

- [x] **`use-local-projects.hook.ts` had no test** despite adding genuinely new behavior (the
  `platform.onDidChangeProjects`-driven refetch, the promise-current/`isMounted` guards, and the
  catch path that must clear the loading flag on a failed fetch). _(fixed during review: stood up a
  vitest harness for `platform-get-resources` — mirroring `platform-scripture-editor` — and added
  `use-local-projects.hook.test.ts`; 14 tests covering initial-loading, success mapping,
  rejected-fetch-still-clears-loading + warn, `onDidChangeProjects` refetch, `enabled:false` skip,
  `refetchTriggers` re-run, unmount/stale-promise guard, exact filter args, and the setting-error
  fallback.)_
- [x] **`ScrTextExtensions.cs` doc comments referenced `ParatextProjectDataProvider.cs` by absolute
  line range** (`~1608-1612`, `~1686-1707`, `~1591-1594`, `~1603-1606`), which rot as that ~2600-line
  file changes. _(fixed during review: removed all four line-number ranges, keeping the durable
  method/getter references — `ParatextProjectDataProvider.GetProjectSetting`, the
  `platform.isPublished`/`languageTag`/`isEditable` getters; verified none remain in `c-sharp/`.)_

_Author response: both were requested fixes and were implemented and verified during the review._

### Minor — Consider

- [ ] ~~`normalizeProjectId` is exported from `project-lookup.service-model` but not re-exported
  through any `@papi/` module, so out-of-tree extensions can't reach it.~~ _(Author: overkill to
  re-export for something this simple; it's a core-internal helper for the picker hook.)_
- [ ] ~~Two id-normalization strategies coexist in `project-lookup.service-model` —
  `normalizeProjectId` (`toUpperCase`) vs. `areProjectIdsEqual` (`localeCompare` accent-insensitive);
  latent inconsistency if the id format ever broadens beyond hex GUIDs.~~ _(Author: the uppercase
  normalization is deliberately the map-key / case-insensitive form for cases where `localeCompare`
  doesn't fit; consistent id handling was otherwise error-prone. No bug today with hex-GUID ids.)_
- [x] **`ThreadingUtils.ObserveTaskLoggingErrorsToStderr` had no direct test.** _(fixed during review:
  added `c-sharp-tests/ThreadingUtilsTests.cs` — 3 tests: single fault logs description+exception to
  stderr; `WhenAll` multi-fault logs every fault (pins the `AggregateException` behavior); success
  logs nothing. Uses a signaling capture writer to await the fire-and-forget continuation
  deterministically.)_
- [x] **`metadataToLocalProjectInfo` (exported pure mapper) was untested.** _(fixed during review:
  covered by 5 cases in the new `use-local-projects.hook.test.ts` — straight-through mapping, id
  fallback for name/fullName, fullName→name fallback, `isPublished`/`language` defaults, explicit-value
  preservation.)_
- [x] **The `excludePdpFactoryIds` setting-resolution block (`useSetting` + `useMemo` +
  `isPlatformError` guard + identical warning) was duplicated verbatim in `home.web-view.tsx` and
  `new-tab.web-view.tsx`.** _(fixed during review: folded the resolution into `useLocalProjects`
  (option 1, the author's choice) — the hook now reads the shared
  `platformGetResources.excludePdpFactoryIdsInHome` setting internally with a stable module-level
  default; both views dropped the block, and New Tab shed its now-unused `useSetting`/`isPlatformError`/
  `logger` imports (grep-verified). The hook test gained a `useSetting` mock and a PlatformError-fallback
  case.)_
- [ ] ~~`metadataToProjectItem(m)` in `use-project-picker-data.hook.ts` uses a single-letter param `m`,
  inconsistent with its sibling `metadataToLocalProjectInfo(data)`.~~ _(Author: `data`/`m` for a single
  datum is conventional; low visibility, no one will notice.)_
- [x] **Backward-facing comments narrating the migration** ("the old (pre-Lazy) code…", "the old picker
  behavior…", "matches the prior effect's early return", "matching the pre-metadata behavior…") at
  several sites, per the project's forward-facing-comment convention. _(fixed during review: swept all
  new/changed comments across the diff and rewrote them forward-facing (or removed where nothing useful
  remained), preserving load-bearing rationale like the `language`-not-`languageTag` guard reason.
  7 files touched, comment-only, build + typecheck verified.)_
- [ ] ~~`LocalParatextProjects` now implements `IDisposable` (owns a `FileSystemWatcher` + debounce
  `Timer`) but the production owner in `Program.cs` never disposes it.~~ _(Author: `IDisposable` was
  added to silence an analyzer warning; disposal isn't needed for the process-lifetime singleton.
  Note: the `IDisposable` addition also introduced a CS0108 regression that WAS fixed — see below.)_

_Author response: findings 1, 2, 6, and 8 dismissed with the reasons above; 3, 4, 5, and 7 were
fixed during review._

### Additional fixes surfaced during review (not in the original findings)

- [x] **CS0108 regression + latent test resource leak.** Adding `IDisposable` to `LocalParatextProjects`
  made the pre-existing test helper `TestLocalParatextProjectsInTempDir.Dispose()` hide the base
  `Dispose()` (CS0108) and never call it, leaking the base's `FileSystemWatcher`/`Timer` in tests.
  _(fixed during review: made `LocalParatextProjects.Dispose()` `virtual`; the helper now `override`s it
  and calls `base.Dispose()` first, dropping the redundant `IDisposable` re-declaration. `dotnet build`
  now emits zero CS0108.)_
- [x] **`platform-get-resources/tsconfig.lint.json` missing `exclude` override.** The new test file fell
  outside the extension's ESLint type-aware project (the base tsconfig excludes `*.test.ts`, and
  `tsconfig.lint.json` overrode `include` but not `exclude`). _(fixed during review: added
  `"exclude": ["node_modules", "**/temp-build"]`, matching sibling extensions `platform-scripture-editor`
  and `legacy-comment-manager`; typecheck unaffected.)_

#### Shared Regions Modified

None. No `#region shared with <url>` markers overlap any changed lines.

#### Extension Config Changes

None triggering propagation. The only `extensions/` source changes are in
`platform-get-resources/src/` (plus that extension's `package.json` test-script and `tsconfig.json`/
`tsconfig.lint.json` test-file wiring). No `webpack.config.ts` or `.eslintrc*` changes, so no propagation
to `paranext-extension-template` / `paranext-multi-extension-template` is owed.

## Positive Observations

- `lib/papi-dts/papi.d.ts` correctly regenerated (not hand-edited); every added declaration and its
  TSDoc match the source change.
- The new optional `ProjectMetadata` fields carry precise absence semantics — notably `isEditable`
  ("absent is NOT `false`; test `isEditable !== false`") — and the picker consumers honor it.
- Cross-language contract kept in sync and verified: C# `ProjectMetadata` gained the six matching fields
  with `[JsonIgnore(WhenWritingNull)]` + CamelCase serialization, and the C# event constant
  `PROJECTS_CHANGED_EVENT_TYPE` equals the TS `'platform.onDidChangeProjects'`.
- The `resolveLanguage` guard change (`language.startsWith('%')` → `!language`) is correct against the new
  data source (C# returns raw-or-null, never the `%…%` placeholder the old per-PDP path returned), and
  keeps the language column blank for language-less projects rather than falsely showing "English".
- Strong C# test coverage accompanies the new logic (`LocalParatextProjectsWatcherTests`,
  `ScrTextExtensionsTests`, `ParatextProjectDataProviderFactoryBaseTests`, `StartupTimingTests`).
- New TS utilities are tested (`startup-timing.util`, `startup-waterfall.util`); the renderer picker hook
  test suite was expanded from 7 to 14 tests.
- `tsconfig.erb.json` closes a real type-check gap for the `startup-waterfall` CLI/parser, wired into
  `npm run typecheck`.
- Good DRY consolidation: `ProjectSettingsNames.TryParseParatextBoolean`,
  `ScrTextExtensions.GetRawParatextSetting`, the shared `useLocalProjects` hook, and `normalizeProjectId`.
- The `ParatextProjectDataProviderFactoryBase` rewrite from double-checked-locking to a per-project
  `Lazy<>` in a `ConcurrentDictionary` is cleaner and better-scoped, with fault-eviction so failed
  lookups self-heal; it incidentally fixed a latent off-by-one (`A-Y` → `A-Z`) and moved to thread-safe
  `Random.Shared`.
- Both `useLocalProjects` and the renderer hook clear the loading flag even on a rejected fetch (no more
  infinite spinner).
- No hardcoded user-visible strings introduced.

## Interview Notes

- **Stated purpose**: reduce startup time. The author instrumented the startup path, identified that
  opening every project to read `Settings.xml` (to enumerate a few display fields) was the biggest time
  block, and used `getMetadataForAllProjects` as the enumeration jump-off point — substituting a cheap
  raw-parameter-dictionary read for those fields.
- **Design understanding**: the author explained the mechanism clearly and without deferring to AI.
  They specifically called out `languageTag` as the deceptive-complexity case because reading it the
  naive way can trigger LDML file parsing/analysis (and settings persistence), which
  `ScrTextExtensions.GetLanguageTag` deliberately avoids. Measured ~5-6s saved on their dev machine.
- **Author-flagged caveat (for the reviewer)**: there may be edge cases where language IDs are reported
  differently when a project's `Settings.xml` or LDML is incorrect or inconsistent. This aligns with the
  deliberate, documented divergences in `GetLanguageTag` (e.g. a corrupt/unloadable LDML does not trigger
  `CreateLayoutEngine`'s coerce-to-English path during enumeration; a null/empty resolved id is coerced to
  `en` without persisting) and in `GetIsEditable` (a malformed `Editable` value reports `false` rather than
  throwing). Worth scrutinizing that these divergences are all acceptable.
- **Decisions**: dismissed the re-export (finding 1), the dual id-normalization note (finding 2), the
  single-letter param (finding 6), and the non-disposal of the singleton (finding 8); requested tests for
  the hook and `ObserveTaskLoggingErrorsToStderr`, the comment sweep, and (for finding 5) folding the
  setting resolution into the hook.
- No unresolved Critical/Important findings; the author demonstrated understanding of every area
  discussed.

## In-Review Quality Check

Changes were made during the interview, so the full quality suite was re-run afterward:

- **typecheck** — PASS (all workspaces).
- **lint** — PASS after one real fix. Repo-wide `npm run lint` hits a known worktree-nesting ESLint
  artifact (`@typescript-eslint` resolves in both the worktree and the parent `/home/lyonsm/paranext-core`
  checkout because the worktree `.eslintrc.js` has no `root: true`) — an environment artifact, not a code
  issue. Linting the changed files directly surfaced one genuine violation (the new test file fell outside
  the extension's lint tsconfig project); fixed via the `tsconfig.lint.json` `exclude` override above.
- **prettier** — PASS (changed files already formatted).
- **npm test** — PASS for all in-review files (`use-project-picker-data` 14, `startup-timing.util` 6,
  `startup-waterfall.util` 9, `use-local-projects.hook` 14). 3 unrelated `lib/platform-bible-react`
  storybook chromium tests timed out (environmental; no lib sources changed).
- **csharpier** — PASS (288 files).
- **dotnet build** — PASS: 0 errors, zero CS0108, only the 5 pre-existing warnings.
- **dotnet test** — 1528 passed / 6 skipped / 2 failed. Both failures are
  `LocalParatextProjectsWatcherTests` FileSystemWatcher tests failing with
  `IOException: inotify instance limit (1024) reached` — a host resource exhaustion (another process on
  the machine held ~1002/1024 inotify instances), not a code defect. No non-inotify C# failures.

## Suggested Review Focus

- [ ] **Parity of the raw-read substitution vs. the real PDP getters.** `ScrTextExtensions.GetRawParatextSetting`
  / `GetLanguageTag` / `GetIsEditable` reproduce the getters' values without their side effects, but with
  documented deliberate divergences. Confirm each divergence is acceptable: corrupt/unloadable LDML not
  coercing `languageTag` to "en" during enumeration; null/empty language id coerced to "en" without
  persisting; malformed `Editable` reporting `false` instead of throwing. **The author explicitly flagged
  possible language-ID edge cases when `Settings.xml`/LDML are inconsistent — cover this directly.**
- [ ] **The live `platform.onDidChangeProjects` refresh path** — new no-payload event emitted from
  `LocalParatextProjects`; drives refetch in both web views (via `useLocalProjects`) and the renderer
  picker. Verify it fires on add/remove/rename (S/R, DBL download) and doesn't storm.
- [ ] **`useProjectPickerData` refresh-generation split** — cheap web-view-driven vs. expensive
  metadata-driven generations, the shared-promise cache keyed by generation, and the miss-then-single-refresh
  guard (`missRefreshedProjectIdRef`) against an infinite refresh loop.
- [ ] **Setting-name mismatch** — `platformGetResources.excludePdpFactoryIdsInHome` is now shared by both
  Home and New Tab (the New Tab surface reuses the Home-named setting). Functionally fine; the name reads
  slightly oddly. Renaming a registered setting key was intentionally left out of scope.
- [ ] **Pre-existing backward-facing comments (optional cleanup, out of this PR's scope)** — the comment
  sweep intentionally left comments that already existed at the base commit: `ProjectSummary.cs`,
  `ManageBooks/ProjectFilterServiceTests.cs`, `Projects/ParatextProjectDataProviderCommentTests.cs`.

AI-assisted — [session](https://claude.ai/code/session_01NB1H9evDHbvNZvu44g9TMr)

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2579)


src/shared/models/project-metadata.model.ts
```diff
@@ -16,6 +16,35 @@ export type ProjectMetadataWithoutFactoryInfo = {
    * project.
    */
   projectInterfaces: ProjectInterfaces[];
+  /**
+   * Short display name of the project. Absent when the producing Project Data Provider Factory does
+   * not supply it; consumers should fall back to {@link id}.
+   */
+  name?: string;
```
tjcouch-sil, line 25: What do you think of throwing `@experimental` on all these properties? Seems like big and somewhat fundamental changes, so it feels nice to ease them in rather than commit right now.

README.md
```diff
@@ -272,6 +272,35 @@ Platform.Bible core extensions are found in the `extensions` folder. Please foll
...
+npm run startup-waterfall                              # reads the default main.log for your platform
+npm run startup-waterfall -- --log=/path/to/main.log   # or point at a specific capture
+npm run startup-waterfall -- --help                    # usage
+```
+
+Notes:
+
+- **Use a packaged build for numbers you intend to compare.** In dev, the .NET provider runs under `dotnet watch`, which inflates the C# portion of the timeline and is not representative.
+- `main.log` accumulates across launches; the tool shows only the latest run and warns when it drops older ones. For a clean capture, delete (or copy aside) the log between runs.
+- For the design rationale and recorded baseline/after measurements, see the startup-performance specs under `docs/superpowers/specs/` (e.g. `2026-07-10-startup-baseline.md`, `2026-07-11-startup-phase3-after.md`).
```
tjcouch-sil: This seems out of date.

tsconfig.erb.json
```diff
@@ -0,0 +1,22 @@
+{
+  // Type-check gate for the startup-waterfall parser + CLI under .erb/scripts, which the root
+  // tsconfig's `include` ("src", "assets") leaves out. Without this, a type error in the parser or
+  // CLI would keep CI green and surface only at manual runtime. Scoped to these files (not all of
+  // .erb/scripts) because the older scripts there carry pre-existing type issues from their own
+  // heavy deps and are out of scope for this gate. The colocated *.util.test.ts is not listed for
```
tjcouch-sil, line 6: Might be worth considering changing the name to something like `tsconfig.waterfall.json` since this intentionally does not check everything in `.erb`. But maybe the idea is "we want it to at some point", so maybe you want to keep the name.

tjcouch-sil (changes requested):
Thanks! Lots of good work here. I'd like to see `@experimental` on the API. Let me know your thoughts.

README.md
lyonsil: Agreed and fixed. I previously swept code comments for backward-facing references explaining why we did something vs. forward-facing comments that can assist maintainers. I missed the README in that sweep.

tsconfig.erb.json
lyonsil: I intentionally left this as `erb` with the idea that this paves the way for more scripts in the same directory to be covered. I am a bit afraid that if I rename to `tsconfig.waterfall.json` then the next time we add a script, AI will create `tsconfig.other-script-name.json` and start a proliferation of `tsconfig` files.

src/shared/models/project-metadata.model.ts
lyonsil: Good idea - done. I put this on some other things added but missed these. I also missed another event that is now also marked as `experimental`. I think it should all be covered now.

lyonsil (commented):
Thanks for the feedback! I made some updates based on your comments. Let me know if you want to discuss anything else or have other suggestions.

> @lyonsil made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2579#-:-Oy9_lri56XPWRMGZtta:bnh4jzl)* status: 0 of 73 files reviewed, 3 unresolved discussions (waiting on tjcouch-sil).

tjcouch-sil (approved):
Thanks for the good updates! I dunno if Tom is planning to review.

lyonsil:
I chatted with Tom. He has a pref-related PRD coming up and will adjust things as needed/helpful in his work.

PR #2576, opened 2026-07-18
PT-4175: First-run state machine and modal shell

katherinejensen00:
## Summary

PT-4175 delivers the **foundation** for the simple-mode first-run setup wizard: a persisted `platform.firstRunComplete` flag and a startup state machine that gates the app behind a **non-dismissable modal** until onboarding completes. It ships the shell, the gate, and four **placeholder** steps that the sibling tickets (PT-4176 Language / PT-4177 Identify / PT-4178 Sync consent / PT-4179 Sync progress) fill in by swapping entries in the shell's `stepComponents` map. Additive and backward-compatible; the four review passes found **no Critical or Important issues**.

## What changed & where to look

Authored source below (`lib/papi-dts/papi.d.ts` and `lib/platform-bible-react/dist/**` in the diff are **regenerated build output**, not hand-edited).

**State machine — start here**
- `src/renderer/services/first-run.reducer.ts` — pure `decideFirstRun(firstRunComplete, wizardActive, registrationValidity)` → `showApp` / `completeThenShowApp` (existing-user migration) / `waitForRegistration` / `startWizard(step)`. The whole routing decision in one function; its test mirrors every row.
- `src/renderer/services/first-run-store.ts` — renderer store that feeds the reducer and exposes a subscribable status. Owns: idempotent startup resolve (StrictMode-safe), migration backfill, resume-to-sync after the registration relaunch, cache-non-clobber on read failure, retry, and the power-mode fast-path.
- `src/renderer/services/resolve-registration-validity.ts` — bounded (15s) tri-state check of `paratextRegistration.doesUserHaveValidRegistration`.

**UI**
- `src/renderer/components/first-run/first-run-overlay.component.tsx` — the full-viewport non-dismissable gate (loading / error+Retry / wizard) at `Z_INDEX_FIRST_RUN`.
- `.../first-run-shell.component.tsx` — wizard chrome (title, "Step X of Y", Back/Skip/Next footer, busy + inline error) and the injectable `stepComponents` map siblings use to swap in real steps.
- `.../steps/*.placeholder.component.tsx` — the four temporary placeholders; `first-run-step-props.model.ts` — the step contract.

**Wiring & settings**
- `src/renderer/app.component.tsx` — mounts the overlay and kicks off resolution on startup.
- `src/main/startup-tasks.ts` — gates simple-mode auto-sync on `firstRunComplete` (a fresh user never syncs before consenting).
[...]

PR #2580, opened 2026-07-21
PT-4176: First-run UI language picker

katherinejensen00:
image

## Summary

Adds the **first screen of the first-run setup wizard: a UI-language picker.** It shows a searchable list of the interface languages that actually have setup-dialog translations, each displayed by its **autonym (its own name, in its own script)** — e.g. `English`, `Español`, `中文（简体）`. Picking one sets `platform.interfaceLanguage` and the **wizard re-renders live in that language** (no restart).

"Which languages qualify" is **auto-detected**: a language appears once it has ≥90% of the setup-dialog (`%firstRun_*%`) keys translated — no hand-maintained list. Today that's English + the locales translated in this PR; Khmer is deliberately deferred until translated.

> **Stacked PR:** based on `worktree-pt-4175` (#2576), so this diff is **PT-4176 only**. It will retarget to `main` automatically when #2576 merges.

### Follow-up added to this branch: default to the OS language on a fresh first run

On a **fresh first run**, if the user's OS/browser locale has enough setup-dialog localization to **qualify for the picker** (the same ≥90% gate), the wizard now **opens in that language** instead of English — so the picker's preselection already matches the user's system. Region-only locales resolve to the right script via `Intl.Locale(...).maximize()`: `zh-CN`→`zh-hans`, `zh-TW`→`zh-hant`, `fr-CA`→`fr`. Non-qualifying locales stay English.

It's **once-only and never overrides an explicit choice** (tied to the `wizardActive` false→true transition), **best-effort** (any failure logs a warning and falls back to English without blocking the wizard), and adds **no new PAPI surface** — it reuses `getSetupDialogLanguages()` and a pure renderer matcher. Because the setting is seeded *before* the wizard mounts, the wizard's strings resolve straight to the OS language, so the live-render bridge above never has to fire a change for the first paint. (Commits: `pickBestSetupLanguage` matcher, then the first-run-store seed.)

## What changed — where to look

| Area | Key file(s) | What it does |
| --- | --- | --- |
| **Detection (backend)** | `extension-host/services/setup-dialog-languages.util.ts` (+ test) | Pure function: given the English baseline + each locale's keys, returns the languages at ≥90% coverage (English always included). Easy to read in isolation. |
| ↳ exposed via | `extension-host/services/localization.service-host.ts`, `shared/services/localization.service-model.ts` | New `getSetupDialogLanguages()` / `SetupDialogLanguages` PAPI data type, mirroring the existing `AvailableInterfaceLanguages`. |
| **Live re-render** | `extension-host/services/localization.service-host.ts` (`subscribeToInterfaceLanguageChanges`) | When `platform.interfaceLanguage` changes, re-emits the localized-string data types so all mounted UI re-renders in the new language. **App-wide** (also affects settings/popover), guarded against no-op emits, subscribed after engine registration. |
| **Autonym data** | `extension-host/data/language-details.data.ts` (+ test) | Extracted the hardcoded autonym map out of the host; added in-script autonyms for `zh-hans`/`zh-hant`. The test fails the build if a shipped locale lacks a real (non-code) autonym. |
| **Picker component** | `lib/platform-bible-react/.../interface-language-picker/interface-language-picker.component.tsx` (+ test/stories/data) | Reusable searchable list (shadcn `Command`). Searches autonym + other names; shows each autonym with `dir="auto"`; scales to 100+ (see stories). |
| **Wizard step** | `renderer/components/first-run/steps/language.component.tsx` (+ test) | Wires the picker to the setting; always keeps the current selection visible (with its real autonym); immediate-apply; gates Next while loading. Swapped into the shell (`first-run-shell.component.tsx`), replacing the placeholder. |
| **OS-language default** (follow-up) | `renderer/services/pick-best-setup-language.ts` (+ test), `renderer/services/first-run-store.ts` (+ test) | Pure `Intl`-based matcher (maximize + progressive base-fallback, POSIX-locale normalization, casing-preserving) + a once-only, best-effort seed in the first-run store that defaults `platform.interfaceLanguage` to a qualifying OS locale before the wizard paints. |
| **Localization** | `assets/localization/{en,es,fr,zh-hans,zh-hant}.json`, `metadata.json` | Picker strings + setup-dialog translations (see caveat below). Placeholder key **deprecated, not deleted**. |
| **Generated — skip** | `lib/papi-dts/papi.d.ts`, `lib/platform-bible-react/dist/**` | Build outputs (`build:types` / `build:basic`), committed intentionally. No need to review line-by-line. |

**Suggested review order:** the detection util → its host wiring → the picker component → the wizard step. That follows the data from "which languages qualify" to "what the user sees."

## Reviewer notes

- **Translations need native review — now more load-bearing.** `es/fr/zh-hans/zh-hant` setup-dialog strings are machine-assisted drafts; please have the localization team confirm (especially `zh-*`). Khmer is intentionally omitted (won't meet the threshold yet). With the OS-language default (below), a matching user is now **auto-shown** the translated wizard rather than opting in via the picker — a rough auto-selected translation is a worse first impression than English, so this review gates the default's value.
- **One decision to sanity-check:** the live-render bridge re-renders the *whole app* on a language change (all mounted localized UI, not just the wizard) — intended, and guarded against redundant emits.
- **OS-language default — accepted risks (best-effort, first-run only).** The seed runs once and is best-effort: if the qualifying-language lookup is slow/incomplete at that instant or fails, the profile stays English with only a logged warning and no retry (once-only, so it's not reattempted on later launches). This degrades safely to English. It also does not remove the first-render English flash inherent to async-resolved localized strings — it only removes the extra English→bridge→OS hop. Both are documented deliberate trade-offs.
- **Not runtime-verified:** the first-run wizard is gated by machine registration, so the flow couldn't be smoke-tested on a registered dev machine. Behavior is covered by tests instead: the detection util (threshold/boundary/exclusion), the live-bridge (seed/no-op/change), and the picker + step (autonym rendering, selection, search, error fallbacks).
- **Tests:** every new non-trivial file has a colocated test; all green (renderer + picker suites, typecheck, lint).

<details>
<summary><b>Full automated review summary</b> (findings, interview notes, follow-ups) — for the review meeting</summary>

# Code Review Summary

**Branch**: pt-4176-language-picker

**Base**: origin/main (review scope narrowed to the PT-4176 work only: `11ffb70e9d4..HEAD`; the PT-4175 first-run-shell commits also on this branch were excluded by author direction)

**Date**: 2026-07-21

**Review model**: Claude Opus 4.8 (1M context)

**Files changed**: 31 (27 source + 4 generated `platform-bible-react` dist bundles; `lib/papi-dts/papi.d.ts` is also generated)

## Overview

PT-4176 adds the first-run wizard's **language-picker step**: a searchable, scrollable list of UI languages shown by their **autonym in their own script**, offering only the languages that have setup-dialog (`%firstRun_*%`) localizations (auto-detected at ≥90% key coverage; English always qualifies). Selecting a language sets `platform.interfaceLanguage` and the wizard **re-renders live** in the chosen language.

The work spans: a reusable `InterfaceLanguagePicker` component (shadcn `Command`) in `platform-bible-react`; a backend detection method (`getSetupDialogLanguages`) exposed as a new `SetupDialogLanguages` PAPI data type, backed by a pure, unit-tested util; an app-wide "live re-render" bridge in the localization host (re-emits localized strings when `platform.interfaceLanguage` changes); an extracted, guard-tested `languageDetails` autonym map (with `zh-hans`/`zh-hant` in-script autonyms added); and the wizard step wiring, replacing the placeholder. Setup-dialog keys were translated into `es/fr/zh-hans/zh-hant` (machine-assisted drafts pending native review); Khmer deferred.

This branch was already through two prior review rounds (specialized reviewers + an 8-angle `/code-review`) before this `/review-paratext` pass; findings below are what this pass surfaced.

## API Changes

- `lib/platform-bible-react` (`src/index.ts`): new exports from `interface-language-picker.component` — `InterfaceLanguagePicker` (default→named), `INTERFACE_LANGUAGE_PICKER_STRING_KEYS`, `type InterfaceLanguagePickerLocalizedStrings`, `type InterfaceLanguagePickerProps`.
- `shared/services/localization.service-model` (surfaced in generated `papi.d.ts`): `LocalizationDataDataTypes` gains `SetupDialogLanguages: DataProviderDataType<undefined, Record<string, LanguageInfo>, never>`; `ILocalizationService` gains `getSetupDialogLanguages(): Promise<Record<string, LanguageInfo>>` and an always-throws `setSetupDialogLanguages()`.
- No changes to `platform-bible-utils` public surface; no `@papi/*` signature changes; no extension type-declaration changes. All changes are additive — no deleted exports, no signature narrowing, no default-behavior changes.

None.

- [x] **Autonym text rendered without per-item bidi direction** (`interface-language-picker.component.tsx`) — a mixed-script list (English, العربية, `中文（简体）`) inheriting one container direction can misrender RTL or parenthesized autonyms. _(fixed during review: wrapped the autonym in `<span dir="auto">` so each lays out per its own script.)_

[Author response: agreed; fixed in-review.]

### Minor — Consider

- [x] **`languages` prop doc / tag contract** — kept the JSDoc as "keyed by **BCP-47 tag**" (the analysis suggested softening to "locale tag"; the author confirmed BCP-47 is the proper identifier and the display name/autonym is a separate concern, so the contract should not be softened). _(resolved by discussion; no code change.)_
- [x] **`package-lock.json` `name` flipped to the worktree name** (install-in-worktree artifact) _(fixed during review: reverted so the lockfile keeps `paranext-core`.)_
- [x] **Stories used an inline structural type** instead of the exported `LanguageInfo` _(fixed during review: import `LanguageInfo`.)_
- [x] **Sample `languages` fixtures duplicated** across the picker story and test _(fixed during review: extracted `SAMPLE_LANGUAGES` into the colocated `interface-language-picker.data.ts`, shared by the test and the FewLanguages story.)_
- [x] **`AvailableInterfaceLanguages` default reused the setup-named constant** _(fixed during review: renamed `DEFAULT_SETUP_LANGUAGES` → `ENGLISH_FALLBACK_LANGUAGES` and used it for both data-type defaults.)_
- [ ] ~~Instruction copy says "the app" rather than a product name~~ _(author: acceptable as neutral, pre-registration first-run copy; leaving.)_
- [ ] ~~`getSetupDialogLanguages` return depends on `LanguageInfo` imported from `platform-bible-react` (not `@papi/core`)~~ _(matches the existing `getAvailableInterfaceLanguages`/`AvailableInterfaceLanguages` precedent; non-issue.)_

[Author response: fixed the five actionable minors during review; dismissed the two noted above with rationale.]

#### Shared Regions Modified

None. No changed source file contains a `#region shared with` marker.

#### Extension Config Changes

None. No files under `extensions/` were changed; the backend logic lives in `src/extension-host/` (core).

## Positive Observations

- Clean, additive API that mirrors the existing `AvailableInterfaceLanguages` data type precisely (same selector/data/error shape, paired throwing setter) — no novel pattern.
- Strong separation of concerns: pure, unit-tested `computeSetupDialogLanguages` util + thin host adapter; `languageDetails` extracted to its own `data.ts` with a build-failing guard test that requires every shipped locale to have a real (non-code) autonym.
- Behavior-focused tests (autonym rendering, selection marking, `onChange` tag, search by each name source, `otherNames` matched-but-never-rendered, hide-search-when-≤1, threshold boundary + exclusion, live-bridge seed/no-op/change) rather than implementation-mirroring; jsdom stubs carry justified lint-suppression comments.
- Accessibility handled: localized `aria-label` on the search input; selection via `aria-current` + visually-hidden localized "Selected" label + `data-checked` (reusing the vendored check, no double icon); the `aria-current`-vs-cmdk-`aria-selected` distinction documented at the call site.
- Correct immediate-apply model (self-reversing selection, live re-render, failed write logged not swallowed); content-first Next-gating during load.
- Localization correct: all user-visible strings localized across all six locale files; the retired placeholder key deprecated in `metadata.json` (not deleted) per the immutability rule.
- Microcopy follows the guides: sentence case, proper `…` ellipsis, second-person "your".
- Thorough Storybook coverage: FewLanguages, OneLanguage, ManyLanguages (120 + scroll), SelectedMidList, RtlAndLongAutonyms.

## Interview Notes

- **Purpose** (author): the first-run UI language picker (PT-4176) — searchable, in-script autonyms, filtered to setup-dialog-localized languages, immediate-apply with live re-render.
- **Scope decision** (author): review PT-4176 only (`11ffb70..HEAD`); PT-4175's first-run shell is separate.
- **Key design decisions** the author directed across the work: percentage (≥90%) threshold keyed off the `%firstRun_*%` namespace; searchable shadcn `Command` (not MUI, not a plain list) to scale to 100+; in-script autonym display; backend detection; app-wide live-render bridge; ship setup-key translations now (km deferred).
- **BCP-47 tag contract** (author): BCP-47 is the proper language identifier and the picker's display name (autonym) is a separate concern; the JSDoc keeps "BCP-47 tag" (not softened). On the raw-vs-canonical casing (`zh-hans` vs `zh-Hans`): the author decided this is a **non-issue and needs no change** — for the shipped locales the only difference is casing, it's the same language/text, and BCP-47 is case-insensitive. No follow-up.
- No areas were deferred to AI without understanding; the author drove the design decisions throughout.

## In-Review Quality Check

Changes were made during the review (the Important bidi fix + five minor fixes). After them:

- **Tests**: renderer `language.component.test.tsx` 8/8 pass; `platform-bible-react` picker suite 11/11 pass.
- **Lint**: 0 errors on changed files; one prettier comment-wrap warning in `interface-language-picker.data.ts` auto-fixed.
- **Typecheck**: `tsc -p ./tsconfig.json` clean, 0 errors.
- **Build**: `platform-bible-react` dist regenerated (`build:basic`) so the renderer consumes the updated component; `papi.d.ts` was resynced earlier in the branch.

No unfixable failures.

## Suggested Review Focus

- [ ] **Live-render bridge is app-wide by design.** Any `platform.interfaceLanguage` change now re-emits `LocalizedString`/`LocalizedStrings` for every subscriber (settings, popover, all web views), guarded against no-op/seed emits and subscribed only after engine registration. Confirm this app-wide behavior is desired.
- [ ] **Translations pending native review.** `es/fr/zh-hans/zh-hant` setup-dialog strings are machine-assisted drafts; Khmer is deferred (won't meet the threshold until translated).
- [ ] **Generated artifacts committed intentionally.** `lib/platform-bible-react/dist/**` and `lib/papi-dts/papi.d.ts` are build outputs committed in-branch; re-running `build:basic` / `build:types` should produce no diff.
- [ ] **Runtime not observed.** The first-run wizard is gated by machine registration, so an automated e2e / manual runtime smoke of the wizard flow was not feasible on a registered machine; the cross-process behavior is covered by the bridge/detection/integration tests instead.

</details>

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2580)


Sebastian-ubs:
Hi Katherine, how does this language picker ui relate to
- Language selection in the user profile
- Language selection in the settings ui
- Language filtering in the Simple download resources ui
- Language filtering in the Power get resources ui
We should try to use similar patterns (therefore likely a similar component) for most of them (likely filtering be another than selecting app locale)

katherinejensen00:
@Sebastian-ubs I made a separate Jira work item to reuse it in Settings, https://paratextstudio.atlassian.net/browse/PT-4234. This PR applies specifically to the setup wizard. I will add user profile and resource language filtering to that existing work item.

rolfheij-sil:
🤖 Claude:

### Code review

Found 2 issues:

1. New public API is missing the `@experimental` TSDoc tag (team convention for all new public-API surface): the `SetupDialogLanguages` data type and `getSetupDialogLanguages`/`setSetupDialogLanguages` in the localization service model (all reaching the generated `papi.d.ts`), and the four new `InterfaceLanguagePicker` exports from platform-bible-react.

https://github.com/paranext/paranext-core/blob/bd7b2bf6e9938022f2f22cf86ef86619f18411f6/src/shared/services/localization.service-model.ts#L36-L38

https://github.com/paranext/paranext-core/blob/bd7b2bf6e9938022f2f22cf86ef86619f18411f6/lib/platform-bible-react/src/index.ts#L171-L176

2. `package-lock.json`'s top-level `name` is still `"pt-4176-language-picker"`. The PR description's review summary says this was reverted to `paranext-core`, but the revert isn't in the current diff; left as is, the mismatched name will churn in future lockfile updates.

https://github.com/paranext/paranext-core/blob/bd7b2bf6e9938022f2f22cf86ef86619f18411f6/package-lock.json#L1-L3

Lower-confidence item worth a look:

- A failed `setInterfaceLanguage` write in the language step is only logged (`logger.warn`) with no user-facing feedback. The UX guideline (`applying-changes.mdx`, Feedback table) says failures should always surface an error toast or inline message, and the accompanying test pins the log-only behavior in place.

https://github.com/paranext/paranext-core/blob/bd7b2bf6e9938022f2f22cf86ef86619f18411f6/src/renderer/components/first-run/steps/language.component.tsx#L66-L75

Minor notes that did not make the confidence cut:

- The testing helper `implementLocalizationDataProviderEngine` calls `subscribeToInterfaceLanguageChanges()` without any `registerEngine`, contradicting that method's own "Call this AFTER the engine is registered" doc (harmless in tests only because the mock bypasses registration; the production path in `initialize()` gets the order right).
- Two new PT-4176 references were added in source content (`language-details.data.ts` comment, `metadata.json` deprecation message) while this same PR deliberately strips ticket refs from other comments; existing deprecation messages on main do not cite tickets.

jolierabideau (changes requested):
Thanks Katherine!

Informational only (related to Sebastian's comment above I think) but Claude pointed out - 

**Selected-language indicator vs. Component Choices guideline.** Selection is shown only by an end-side check mark (inherited from the vendored cmdk `CommandItem`) plus a visually-hidden "Selected" label. The `Guidelines/Component Choices` page calls for a start-side indicator and a selected-row background (`tw:bg-muted/50`) for single-select controls. The guideline itself marks the start-side point as "up for UX discussion," so this is informational — worth confirming with UX whether the picker should match the convention or is exempt as a distinct list component.

> @jolierabideau partially reviewed 64 files and all commit messages, and made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2580#-:-OyZ6D5Y8jm9rzf1qeFW:be0euss)* status: all files reviewed, 3 unresolved discussions (waiting on katherinejensen00).

___
*[`.eslintrc.js` line 2 at r2](https://reviewable.io/reviews/paranext/paranext-core/2580#-OyZ5MxHCOT_3GmaK3tF:-OyZ5MxHCOT_3GmaK3tG:bewhwll) ([raw file](https://github.com/paranext/paranext-core/blob/bd7b2bf6e9938022f2f22cf86ef86619f18411f6/.eslintrc.js#L2)):*
> ```JavaScript
> module.exports = {
>   root: true,
> ```

If I understand correctly, this is a repo wide change to not have eslint search in parent dirs. Is this intentional for this PR?
___
*[`lib/platform-bible-react/src/components/advanced/interface-language-picker/interface-language-picker.component.tsx` line 119 at r2](https://reviewable.io/reviews/paranext/paranext-core/2580#-OyZ5lCm6DdotygOCqMo:-OyZ5lCm6DdotygOCqMp:b-wrrcw) ([raw file](https://github.com/paranext/paranext-core/blob/bd7b2bf6e9938022f2f22cf86ef86619f18411f6/lib/platform-bible-react/src/components/advanced/interface-language-picker/interface-language-picker.component.tsx#L119)):*
> ```XML
>               key={tag}
>               value={tag}
>               keywords={keywords}
> ```

NIT From Claude

**`interface-language-picker.component.tsx:119` — `keywords={keywords}` on `CommandItem` looks inert.** The `Command` is configured with `shouldFilter={false}`, and the real matching is done by the `visibleEntries` memo (lines ~77–81). Since cmdk's `keywords` only feeds its internal (disabled) filter, this prop appears to do nothing. Worth dropping it or adding a one-line note so a future reader doesn't assume cmdk is doing the matching. (Flagging for your confirmation rather than asserting — depends on cmdk internals.)
___
*[`src/renderer/components/first-run/steps/language.component.tsx` line 36 at r2](https://reviewable.io/reviews/paranext/paranext-core/2580#-OyZ6kJj5ytJUTEBIrrX:-OyZ6kJj5ytJUTEBIrrY:b-umtli1) ([raw file](https://github.com/paranext/paranext-core/blob/bd7b2bf6e9938022f2f22cf86ef86619f18411f6/src/renderer/components/first-run/steps/language.component.tsx#L36)):*
> ```XML
>   const primaryLanguage = safeInterfaceLanguage[0] ?? 'en';
> 
>   const [setupLanguagesPossiblyError, , isLoading] = useData(
> ```

NIT From Claude 

**Repeated `isPlatformError(x) ? FALLBACK : x` unwrap** for both `useData` results in `language.component.tsx:36–50`. A tiny local helper would DRY it, but the inline form is arguably clearer — judgment call, not a defect.

katherinejensen00:
Fixed in 2f613c8bd86. Details:

**1. `@experimental` tags (both locations)** — Added `/** @experimental */` to:
- `SetupDialogLanguages` data type, `getSetupDialogLanguages`, and `setSetupDialogLanguages` in `localization.service-model.ts` (API specific to the first-run wizard; matches the pattern guide's "APIs designed for one particular UI or use case" criterion)
- `InterfaceLanguagePicker`, `InterfaceLanguagePickerProps`, `InterfaceLanguagePickerLocalizedStrings`, and `INTERFACE_LANGUAGE_PICKER_STRING_KEYS` in the component source
- Regenerated `papi.d.ts` (build:types) and `platform-bible-react/dist` (build:basic) to surface the tags in the generated output

**2. `package-lock.json` name** — Reverted to `paranext-core`. (The PR review summary said this was fixed during review, but the commit didn't actually land — it was only run locally without staging. Fixed now.)

**3. Error toast for failed `setInterfaceLanguage` (lower-confidence item)** — The applying-changes guideline is clear ("Failure: always"), so implemented it:
- Added `%firstRun_language_setFailed%` to en/es/fr/zh-hans/zh-hant locale files (machine-assisted drafts, consistent with the other setup-dialog strings already pending native review)
- The catch block now calls `toast.error(strings['%firstRun_language_setFailed%'])` in addition to `logger.warn`
- Test updated to assert both the warn log and the toast

**Minor notes (no code change):**
- The `implementLocalizationDataProviderEngine` test-helper order: acknowledged. The production path in `initialize()` gates correctly; the helper bypasses registration intentionally (it's test infrastructure). No change needed.
- PT-4176 ticket references in `language-details.data.ts` comment and `metadata.json` deprecation message: the deprecation message is directly analogous to the `%commentEditor_cancelButton_tooltip%` entry already in `metadata.json` (which also cites its replacement key without a ticket). The `language-details.data.ts` comment was a deliberate call to explain why the guard test exists. Leaving both as-is.

jolierabideau (commented):
> @jolierabideau resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2580)* status: 40 of 67 files reviewed, 2 unresolved discussions (waiting on katherinejensen00).

katherinejensen00 (commented):
Thank you for your help and the insightful feedback, Jolie!

> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2580#-:-Oy_9OPB2ZFgYCU56pRn:bd85qt1)* status: 40 of 67 files reviewed, 2 unresolved discussions (waiting on jolierabideau).

___
*[`.eslintrc.js` line 2 at r2](https://reviewable.io/reviews/paranext/paranext-core/2580#-OyZ5MxHCOT_3GmaK3tF:-Oy_9GmS-_O6yfYMlSMK:bovs13g) ([raw file](https://github.com/paranext/paranext-core/blob/bd7b2bf6e9938022f2f22cf86ef86619f18411f6/.eslintrc.js#L2)):*

If I understand correctly, this is a repo wide change to not have eslint search in parent dirs. Is this intentional for this PR?
</blockquote></details>

Yes, intentional — it was added to fix a duplicate-plugin ESLint error that showed up when running in the nested git worktree setup this branch uses. Without `root: true`, ESLint walks up into the parent directory and loads the config a second time, causing a plugin-registered-twice error. It's also just correct for any root config in a monorepo; there's nothing meaningful above it to find. No change to behavior on a normal checkout.
___
*[`lib/platform-bible-react/src/components/advanced/interface-language-picker/interface-language-picker.component.tsx` line 119 at r2](https://reviewable.io/reviews/paranext/paranext-core/2580#-OyZ5lCm6DdotygOCqMo:-Oy_9NJ4Dp60rjawv2l4:bt2vr1b) ([raw file](https://github.com/paranext/paranext-core/blob/bd7b2bf6e9938022f2f22cf86ef86619f18411f6/lib/platform-bible-react/src/components/advanced/interface-language-picker/interface-language-picker.component.tsx#L119)):*

NIT From Claude

**`interface-language-picker.component.tsx:119` — `keywords={keywords}` on `CommandItem` looks inert.** The `Command` is configured with `shouldFilter={false}`, and the real matching is done by the `visibleEntries` memo (lines ~77–81). Since cmdk's `keywords` only feeds its internal (disabled) filter, this prop appears to do nothing. Worth dropping it or adding a one-line note so a future reader doesn't assume cmdk is doing the matching. (Flagging for your confirmation rather than asserting — depends on cmdk internals.)
</blockquote></details>

Fixed.

jolierabideau (approved):
> @jolierabideau partially reviewed 27 files and all commit messages, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2580)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

jolierabideau (approved):
> @jolierabideau reviewed 6 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2580)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2589, opened 2026-07-23
PT-4178: Sync consent step for first-run wizard

katherinejensen00:
image

https://github.com/user-attachments/assets/dffa1168-40b5-49e6-935c-76136d1e26a7

## Summary

- Adds `platform.firstRunSyncSkipped: boolean` hidden setting (default `false`)
- Renderer (`first-run-store.ts`): on \"Skip for now\", writes `firstRunComplete` first, then `firstRunSyncSkipped` — ordering ensures a crash between the two writes leaves sync enabled rather than permanently suppressed
- Main process (`startup-tasks.ts`): gates simple-mode startup sync behind both `firstRunComplete` and `firstRunSyncSkipped`; fail-open on unreadable setting (proceeding with sync is consent-safe when the user may never have skipped)
- Replaces the `sync-consent.placeholder.component.tsx` placeholder with a real `SyncConsentStep` component (static heading + body; shell owns footer buttons)
- Adds new localization key `%firstRun_button_skipForNow%` = \"Skip for now\" for the sync consent skip button; `%firstRun_button_skip%` restored to \"Skip setup\" (existing value — key immutability rule)

## Architecture notes

**Dual-signal design:** `platform.firstRunSyncSkipped` is the durable gate (main process reads it at startup). A renderer-only localStorage hint (`platform-bible.firstRunSyncSkipped`) is also written as a lightweight cache for any future in-wizard logic — it is explicitly NOT the gating signal and not read by startup-tasks.

**Fail-open/fail-closed asymmetry is intentional:** An unreadable `firstRunComplete` defaults to skipping sync (consent-safe — better to delay sync than expose an un-consented user). An unreadable `firstRunSyncSkipped` defaults to syncing (consent-safe — better to sync than permanently suppress a user who never chose to skip).

**`platform.firstRunSyncSkipped` is permanent** (hidden, never reset by core). A user who skips and later wants automatic startup sync must edit the settings file manually or clear their profile — there is no UI reset path. This is a product decision captured in the JSDoc.

**Hidden case for cross-view sync:** not applicable — `SyncConsentStep` renders static body content only; no geometry-driven sync.

## Skip path: empty-state behavior

_(Requires manual smoke test — deferred to human reviewer. See task-5-brief.md for the test script.)_

After \"Skip for now\" the wizard completes and the simple-mode layout opens with no synced projects. Startup auto-sync is suppressed on all subsequent launches via `platform.firstRunSyncSkipped=true`. Manual Send/Receive is unaffected.

## Test plan

- [x] `npm test` — all tests pass (pre-existing `pt9-css-converter` prettier timeout is unrelated to these changes)
- [x] `npm run lint` — 0 errors
- [x] `npm run typecheck` — 0 errors
- [x] `notification-display.test.tsx` flaky post-teardown error fixed (pre-existing; Sonner auto-dismiss timer drained before jsdom teardown with `vi.useFakeTimers({ shouldAdvanceTime: true })`)
- [ ] Manual smoke test: start app in simple mode as fresh user, walk wizard to sync-consent step, verify heading/body, click \"Skip for now\", confirm app opens without crash, relaunch and confirm startup sync did not fire (check logs for `platform.firstRunSyncSkipped is true`)

## Key findings fixed from \`/code-review max\` (48 agents, 3.7M tokens)

- **Crash-window ordering** (CONFIRMED): `firstRunSyncSkipped` was written before `firstRunComplete` — a crash between the two would leave the skip gate permanently armed from the aborted session. Fixed: `markFirstRunComplete()` now runs first.
- **Missing setting declaration tests** (CONFIRMED): Added `platform.firstRunSyncSkipped` test block to `core-settings-info.data.test.ts` (mirrors `firstRunComplete` pattern: hidden, default false, booleanValidator)
- **Stale mock key** (CONFIRMED): `first-run-overlay.component.test.tsx` still referenced `%firstRun_step_syncConsent_placeholder%` after the key was removed. Fixed.
- **Incomplete throw-test assertions** (CONFIRMED): The \"still completes when syncSkipped write throws\" test didn't assert the write was attempted or that localStorage was written. Fixed.
- **Missing Next-on-syncConsent test** (CONFIRMED): No test verified that clicking Next navigates without calling `completeFirstRun`. Fixed.
- **Unnecessary test scaffolding** (CONFIRMED): Three `completeFirstRun` tests called `resolveFirstRunState()` unnecessarily. Simplified.
- **Wrong mock tuple shape** (CONFIRMED): `useLocalizedStrings` mock returned `[strings]` instead of `[strings, false]`. Fixed.

## Key findings fixed from \`/review-paratext\` author interview

- **Localization key mutation** (CONFIRMED by 3 agents): `%firstRun_button_skip%` value was changed in-place from \"Skip setup\" to \"Skip for now\", violating the key immutability rule. Fixed: restored original value, added new `%firstRun_button_skipForNow%` key.
- **`_props` unused-vars lint error**: `SyncConsentStep` used `_props` prefix which isn't configured as an ESLint ignore pattern. Fixed with empty destructuring `({}: FirstRunStepProps)`.
- **`require-disable-comment` lint error**: Justification on `eslint-disable-next-line` was inline rather than on a separate line above. Fixed.
- **Write-order test missing**: No test covered the crash-safe `firstRunComplete`-before-`firstRunSyncSkipped` invariant. Added explicit ordering test to `first-run-store.test.ts`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2589)


rolfheij-sil:
🤖 Claude:

### Code review

Found 3 issues:

1. The `platform.firstRunSyncSkipped` write has no self-heal: a transient settings-write failure is only logged, and nothing ever re-reads or re-persists the value, so the user's "Don't sync yet" choice is silently lost and auto-sync fires on every later launch. The sibling `firstRunComplete` write got exactly this treatment after review on #2576 (self-heal in `resolveInternal`, same file), and the localStorage cache this PR writes for the skip flag (`SYNC_SKIPPED_KEY`) is never read anywhere — the self-heal ingredient exists but is unused.

https://github.com/paranext/paranext-core/blob/b64a6a3a029e3c784070eefe7d61c373da2d9f80/src/renderer/services/first-run-store.ts#L230-L239

2. The new public settings key `platform.firstRunSyncSkipped` is missing the `@experimental` TSDoc tag (team convention for all new public-API items reaching `papi.d.ts`). The sibling `platform.firstRunComplete` from #2576 lacks it too, so worth fixing both while here.

https://github.com/paranext/paranext-core/blob/b64a6a3a029e3c784070eefe7d61c373da2d9f80/src/declarations/papi-shared-types.ts#L282-L289

3. `%firstRun_step_syncConsent_placeholder%` (already on main) is deleted outright from `en.json` with no deprecation/fallback entry in `metadata.json`. The Localization Guide ("Existing Strings Are Immutable") requires keeping the old key and adding a metadata mapping; #2580 handles the retired `%firstRun_step_language_placeholder%` that way.

https://github.com/paranext/paranext-core/blob/b64a6a3a029e3c784070eefe7d61c373da2d9f80/assets/localization/en.json#L41-L44

- Leftovers from the in-flight relabel to "Don't sync yet": `%firstRun_button_skipForNow%` in `en.json` (line 33) is defined but never referenced, and five comments still describe the abandoned "Skip for now" label — `startup-tasks.ts` line 178, `papi-shared-types.ts` line 283 (public API TSDoc, mirrored in `papi.d.ts`), `first-run-store.ts` line 217, and `startup-tasks.test.ts` line 122.
- The permanent skip has neither friction before (plain ghost button, no confirmation or consequence preview) nor recovery after (no UI path back; only a hand-edit of the settings file) — `applying-changes.mdx` asks for one of the two for hard-to-reverse changes. If this was an explicit product decision, a pointer to that decision would help future readers.
- The new `afterEach` in `notification-display.test.tsx` (lines 102-108) drains Sonner timers outside `act()`, deterministically emitting ~36 "not wrapped in act(...)" warnings (tests still pass). Branch `pt-4179-storybook-renderer-story-stubs` contains an act()-safe fix for the same timer leak (`act(() => toast.dismiss())`), so two competing fixes are currently in flight.
- The shell/step contract docs are now stale: `first-run-step-props.model.ts` still says "The shell owns the footer buttons... a step renders only its body", and `FirstRunShell`'s docstring still says it owns "the footer (Back / Skip / Next)" — both predate `stepManagesOwnFooter` and the Skip removal introduced here.

Minor note that did not make the confidence cut: `SyncConsentStep.handleSync` re-implements the shell's `runAction` busy/error shape almost line-for-line; nothing shared exists to reuse today, but worth unifying if a shared wrapper gets extracted.

tombogle:
@katherinejensen00 If possible add a video.

jolierabideau (changes requested):
> @jolierabideau reviewed 21 files and all commit messages, and made 9 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2589)* status: all files reviewed, 9 unresolved discussions (waiting on katherinejensen00).

___
*[`.storybook/main.ts` line 169 at r1](https://reviewable.io/reviews/paranext/paranext-core/2589#-OyZD6hP63o1o8ifU-Sd:-OyZD6hP63o1o8ifU-Se:b-bq0vrp) ([raw file](https://github.com/paranext/paranext-core/blob/b64a6a3a029e3c784070eefe7d61c373da2d9f80/.storybook/main.ts#L169)):*
> ```XML
>         // keys from the bundled en.json (same source as the @papi stubs above). Exact-match ($)
>         // key so submodule imports are unaffected.
>         '@renderer/hooks/papi-hooks$': join(__dirname, 'papi-stubs/renderer-hooks.ts'),
> ```

NIT From Claude

**Redundant Storybook resolver config.** [.storybook/main.ts:162-183](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/.storybook/main.ts#L162-L183) Both a webpack `resolve.alias` exact-match key and a `NormalModuleReplacementPlugin` point at the same `@renderer/hooks/papi-hooks` stub. If the plugin is what actually works (because `TsconfigPathsPlugin` resolves `@renderer/*` first), the alias is effectively dead config. Confirm which is authoritative and drop the redundant one, or add a one-line note that both are genuinely required and must be kept in lockstep.
___
*[`.storybook/papi-stubs/renderer-hooks.ts` at r1](https://reviewable.io/reviews/paranext/paranext-core/2589#-OyZD1DG40vIhL1jNiBe:-OyZD1DG40vIhL1jNiBf:b-idhkgc) ([raw file](https://github.com/paranext/paranext-core/blob/b64a6a3a029e3c784070eefe7d61c373da2d9f80/.storybook/papi-stubs/renderer-hooks.ts)):*
NIT From Claude

**Two Storybook stub surfaces overlap.** [.storybook/papi-stubs/renderer-hooks.ts](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/.storybook/papi-stubs/renderer-hooks.ts) Near-identical inert implementations of `useLocalizedStrings`/`useSetting`/`useData`/`useProjectData`/`useProjectSetting`/`useProjectDataProvider` already exist in `.storybook/papi-stubs/frontend-react.ts`. Since `@renderer/hooks/papi-hooks` re-exports the PAPI frontend-react hooks, the new stub could re-export the shared inert impls rather than re-declaring them.
___
*[`assets/localization/en.json` line 32 at r1](https://reviewable.io/reviews/paranext/paranext-core/2589#-OyZCcCWCG1wzhNKzDiO:-OyZCcCWCG1wzhNKzDiP:b-e1effs) ([raw file](https://github.com/paranext/paranext-core/blob/b64a6a3a029e3c784070eefe7d61c373da2d9f80/assets/localization/en.json#L32)):*
> ```JSON
>   "%firstRun_button_retry%": "Retry",
>   "%firstRun_button_dontSyncYet%": "Don't sync yet",
>   "%firstRun_button_skip%": "Skip setup",
> ```

NIT From Claude 

**`%firstRun_button_skip%` ("Skip setup") is now a fully orphaned key.** [en.json:32](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/assets/localization/en.json#L32) This branch removed the shell's generic Skip button and its `KEYS` reference, so this key has zero references in `src`/`lib`/tests at HEAD (verified: it _was_ referenced at the merge base). This is a **different** orphan than the `skipForNow` one already noted. Same disposition question as the placeholder key the earlier review flagged — remove it or handle it per the immutability/deprecation rule in `metadata.json`.
___
*[`assets/localization/en.json` line 42 at r1](https://reviewable.io/reviews/paranext/paranext-core/2589#-OyZDPZ43AvM2QbtMk_1:-OyZDPZ43AvM2QbtMk_2:b-q5m26) ([raw file](https://github.com/paranext/paranext-core/blob/b64a6a3a029e3c784070eefe7d61c373da2d9f80/assets/localization/en.json#L42)):*
> ```JSON
>   "%firstRun_step_identify_placeholder%": "Identify: name + registration code (coming soon)",
>   "%firstRun_step_language_placeholder%": "Language picker (coming soon)",
>   "%firstRun_step_syncConsent_body%": "Your projects are stored on a shared server. Syncing brings your work up to date and shares it with your team.",
> ```

NIT From Claude

**Copy vocabulary inconsistency.** [en.json:42-43](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/assets/localization/en.json#L42-L43) Heading is "Sync your **data**" while the body says "Your **projects**… your **work**." Prefer one noun for the same thing within a step ("projects" is more concrete).
___
*[`src/renderer/components/first-run/first-run-shell.component.tsx` line 95 at r1](https://reviewable.io/reviews/paranext/paranext-core/2589#-OyZClNTEw52_KHOrLr3:-OyZClNTEw52_KHOrLr4:bns0gja) ([raw file](https://github.com/paranext/paranext-core/blob/b64a6a3a029e3c784070eefe7d61c373da2d9f80/src/renderer/components/first-run/first-run-shell.component.tsx#L95)):*
> ```XML
>   );
> 
>   const onSkip = useMemo(
> ```

NIT From Claude

**Generic shell hardcodes `step === 'syncConsent'` in two places.** [first-run-shell.component.tsx:95-105](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/first-run/first-run-shell.component.tsx#L95-L105) Both `onSkip` and `stepManagesOwnFooter` name-check the specific step, coupling the step-agnostic shell to one step and splitting footer ownership (placeholder steps use the shell footer; `syncConsent` renders its own). Consider driving footer ownership from a step-declared capability (e.g. a `managesOwnFooter` flag on the step contract) so future real steps don't each require editing the shell. (This is the code counterpart to the stale contract-docs item in the earlier review.)
___
*[`src/renderer/components/first-run/wizard-step-form.component.tsx` at r1](https://reviewable.io/reviews/paranext/paranext-core/2589#-OyZCWLS5aEBidDN7yfV:-OyZCWLS5aEBidDN7yfW:bgd7il2) ([raw file](https://github.com/paranext/paranext-core/blob/b64a6a3a029e3c784070eefe7d61c373da2d9f80/src/renderer/components/first-run/wizard-step-form.component.tsx)):*
Do we need a Storybook story for this?
___
*[`src/renderer/components/first-run/steps/sync-consent-step.component.tsx` line 23 at r1](https://reviewable.io/reviews/paranext/paranext-core/2589#-OyZDK7T0OTinVyO-VYk:-OyZDK7T0OTinVyO-VYl:bw1fi0e) ([raw file](https://github.com/paranext/paranext-core/blob/b64a6a3a029e3c784070eefe7d61c373da2d9f80/src/renderer/components/first-run/steps/sync-consent-step.component.tsx#L23)):*
> ```XML
>  * Sync consent wizard step. Presents "Sync" and "Don't sync yet" options.
>  *
>  * "Sync" fires `paratextBibleSendReceive.syncProjects` (all projects — all-or-nothing per PT-4261)
> ```

NIT From Claude

**TSDoc describes an interstitial the wiring can't show.** [sync-consent-step.component.tsx:23-24](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/first-run/steps/sync-consent-step.component.tsx#L23-L24) The doc says "The sync progress interstitial (PT-4179) shows while sync runs," but `handleSync` `await`s `onSync()` to completion _before_ calling `onNext()` to advance to the `syncProgress` step — so the progress step is only reached after sync has already finished; the busy state is the local button `Spinner`. Reconcile the comment (or the wiring) when PT-4179 lands.
___
*[`src/renderer/components/first-run/steps/sync-consent-step.component.tsx` line 71 at r1](https://reviewable.io/reviews/paranext/paranext-core/2589#-OyZDX53Bk_driDn6i4L:-OyZDX53Bk_driDn6i4M:blxoqne) ([raw file](https://github.com/paranext/paranext-core/blob/b64a6a3a029e3c784070eefe7d61c373da2d9f80/src/renderer/components/first-run/steps/sync-consent-step.component.tsx#L71)):*
> ```XML
>       secondaryButton={
>         !isSyncing && onSkip ? (
>           <Button variant="ghost" onClick={onSkip}>
> ```

NIT From Claude

**Decline action styled as lowest-emphasis `ghost`.** [sync-consent-step.component.tsx:71](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/first-run/steps/sync-consent-step.component.tsx#L71) For a genuine consent decision where declining is a legitimate (and permanent) choice, a bounds-less `ghost` button lowers its discoverability relative to primary "Sync." Consider `outline` (as the Back button already uses) so the two consent options read as more balanced.
___
*[`src/renderer/components/first-run/steps/sync-consent-step.component.tsx` line 72 at r1](https://reviewable.io/reviews/paranext/paranext-core/2589#-OyZCO7A9Dih7_UFVv-Y:-OyZCO7A9Dih7_UFVv-Z:b-56rey0) ([raw file](https://github.com/paranext/paranext-core/blob/b64a6a3a029e3c784070eefe7d61c373da2d9f80/src/renderer/components/first-run/steps/sync-consent-step.component.tsx#L72)):*
> ```XML
>         !isSyncing && onSkip ? (
>           <Button variant="ghost" onClick={onSkip}>
>             {strings['%firstRun_button_dontSyncYet%']}
> ```

From Claude 

**"Don't sync yet" label implies the choice is temporary, but it's permanent.** [sync-consent-step.component.tsx:72](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/first-run/steps/sync-consent-step.component.tsx#L72) · [en.json:31](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/assets/localization/en.json#L31) `onSkip` persists `platform.firstRunSyncSkipped = true`, which suppresses startup auto-sync **from then on** — it's never re-offered (per the PR's own "permanent, no UI reset path" note). The word "yet" tells the user "not now, I'll be asked again later," which contradicts the actual behavior. This is separate from the earlier review's friction/recovery point — that was about _mechanisms_; this is the label itself misrepresenting the outcome. Either reword to reflect permanence, or add body copy clarifying startup sync stays off until manually re-enabled.

tombogle (changes requested):
> @tombogle reviewed all commit messages and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2589)* status: all files reviewed, 10 unresolved discussions (waiting on jolierabideau and katherinejensen00).

___
*[`src/renderer/components/first-run/steps/sync-consent-step.component.tsx` line 72 at r1](https://reviewable.io/reviews/paranext/paranext-core/2589#-OyZCO7A9Dih7_UFVv-Y:-OyZGOVr2XucIYYtBfDO:b-d7lleb) ([raw file](https://github.com/paranext/paranext-core/blob/b64a6a3a029e3c784070eefe7d61c373da2d9f80/src/renderer/components/first-run/steps/sync-consent-step.component.tsx#L72)):*

From Claude 

**"Don't sync yet" label implies the choice is temporary, but it's permanent.** [sync-consent-step.component.tsx:72](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/first-run/steps/sync-consent-step.component.tsx#L72) · [en.json:31](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/assets/localization/en.json#L31) `onSkip` persists `platform.firstRunSyncSkipped = true`, which suppresses startup auto-sync **from then on** — it's never re-offered (per the PR's own "permanent, no UI reset path" note). The word "yet" tells the user "not now, I'll be asked again later," which contradicts the actual behavior. This is separate from the earlier review's friction/recovery point — that was about _mechanisms_; this is the label itself misrepresenting the outcome. Either reword to reflect permanence, or add body copy clarifying startup sync stays off until manually re-enabled.
</blockquote></details>

I was starting to type this up myself, but I see Claude already found it. Upgrading to blocker.
___
*[`assets/localization/en.json` line 31 at r1](https://reviewable.io/reviews/paranext/paranext-core/2589#-OyZGbCK336ADjclSTOu:-OyZFoyj3IQJQfYCZb-_:b-w92m0o) ([raw file](https://github.com/paranext/paranext-core/blob/b64a6a3a029e3c784070eefe7d61c373da2d9f80/assets/localization/en.json#L31)):*
> ```JSON
>   "%firstRun_button_next%": "Next",
>   "%firstRun_button_retry%": "Retry",
>   "%firstRun_button_dontSyncYet%": "Don't sync yet",
> ```

No Spanish equivalent. I think this comes after the UI locale choice, right? We should supply a Spanish string. (But first need to address the issue of what the UI should actually say in English.)

katherinejensen00:
Addressed in commit 60f9f700168. Here's what was fixed and one item I pushed back on:

**High-confidence findings — fixed:**

1. **Self-heal for `platform.firstRunSyncSkipped`** — Added the same cache-driven re-persist pattern that `firstRunComplete` uses. In `resolveInternal()`, when `firstRunComplete` is true and the localStorage cache says the user skipped sync but the setting still reads `false`, the write is re-attempted. Added 3 covering tests (`first-run-store.test.ts`).

2. **`%firstRun_step_syncConsent_placeholder%` deletion** — Restored the key to `en.json` with its original value and added a `deprecationInfo` entry to `metadata.json` per the key-immutability rule.

3. **`@experimental` on settings** — Pushed back. After checking the codebase, `@experimental` is applied only to *command handlers* in `papi-shared-types.ts`, never to settings. `platform.interfaceMode` (explicitly noted as experimental) doesn't use the tag either. Applying it here would be the first instance for a setting — inconsistent precedent. Not implemented.

**Lower-confidence findings — fixed:**

- **Unused `%firstRun_button_skipForNow%`** — Removed from `en.json`. The key was never on `main`, so the immutability rule doesn't apply.
- **Stale "Skip for now" comments** — Updated to "Don't sync yet" in `startup-tasks.ts`, `startup-tasks.test.ts`, `papi-shared-types.ts`, `first-run-store.ts`, and `papi.d.ts` (regenerated).
- **Stale shell/step contract docs** — Updated `first-run-shell.component.tsx` docstring (removed "Skip" from footer list, noted `stepManagesOwnFooter`) and `first-run-step-props.model.ts` docstring (clarified that steps with custom footers receive navigation callbacks directly).

**Not addressed:**

- **`afterEach` in `notification-display.test.tsx`** — Pre-existing issue; the PR description documents the chosen approach. Branch `pt-4179` has a competing fix in flight — leaving this to avoid a conflict.
- **Shell/step `SyncConsentStep.handleSync` vs shell `runAction` duplication** — Reviewer noted this as a minor observation below the confidence cut. No shared abstraction exists today; deferring until something to generalize against materializes.

katherinejensen00:
Addressed in commit cd82e94032a. Rebased on main in the same commit.

**Tom's comments — both fixed:**

1. **\"Don't sync yet\" label implies temporary** (blocker) — Renamed `%firstRun_button_dontSyncYet%` → `%firstRun_button_skipSync%` with label **\"Skip automatic sync\"**. \"Yet\" implied the choice was temporary and the user would be re-offered; \"Skip automatic sync\" accurately describes that startup auto-sync is suppressed permanently on all subsequent launches. All comment references (TSDoc, inline comments, test descriptions, stories) updated to match.

2. **No Spanish equivalent** — Added Spanish translations for the three strings introduced in this PR that appear on the sync-consent step: heading (`%firstRun_step_syncConsent_heading%`), body (`%firstRun_step_syncConsent_body%`), and the new skip button (`%firstRun_button_skipSync%`). Since this step follows the locale-selection step, all three need to render in the user's chosen language.

tombogle:
*[`assets/localization/en.json` line 42 at r1](https://reviewable.io/reviews/paranext/paranext-core/2589#-OyZDPZ43AvM2QbtMk_1:-OyZp8HJ5TvJabF8F43x:be0zhcx) ([raw file](https://github.com/paranext/paranext-core/blob/b64a6a3a029e3c784070eefe7d61c373da2d9f80/assets/localization/en.json#L42)):*

NIT From Claude

**Copy vocabulary inconsistency.** [en.json:42-43](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/assets/localization/en.json#L42-L43) Heading is "Sync your **data**" while the body says "Your **projects**… your **work**." Prefer one noun for the same thing within a step ("projects" is more concrete).
</blockquote></details>

Go with "projects". I think the existing wording muddies the water about the two-way nature of the sync. How about:

-   Your projects are stored on a shared server. Syncing updates your copy and shares your changes with your team.
-   Your team's projects are stored on a shared server. Syncing updates your copy and shares your changes with your team.
-   Projects are stored on a shared server. Syncing updates your copy and shares your changes with your team.

At this point, do we know for sure that the user a) has any projects, b) that they are on the server, c) is part of a team?

tombogle:
*[`assets/localization/en.json` line 42 at r1](https://reviewable.io/reviews/paranext/paranext-core/2589#-OyZDPZ43AvM2QbtMk_1:-OyZrVI98BC_fuhtXo9Y:b-k3rqm7) ([raw file](https://github.com/paranext/paranext-core/blob/b64a6a3a029e3c784070eefe7d61c373da2d9f80/assets/localization/en.json#L42)):*

Go with "projects". I think the existing wording muddies the water about the two-way nature of the sync. How about:

-   Your projects are stored on a shared server. Syncing updates your copy and shares your changes with your team.
-   Your team's projects are stored on a shared server. Syncing updates your copy and shares your changes with your team.
-   Projects are stored on a shared server. Syncing updates your copy and shares your changes with your team.

At this point, do we know for sure that the user a) has any projects, b) that they are on the server, c) is part of a team?
</blockquote></details>

If not, we could consider an option like:
-   **Syncing updates your local copy and uploads your changes to the shared server.** (Doesn't assume projects or a team.)
-   **If you're working on shared projects, syncing updates your local copy and uploads your changes.** (Explicitly conditional.)
-   **When working on shared projects, syncing updates your local copy and shares your changes with others.** (Avoids assuming the user has a team.)
-   **Syncing keeps your local copy up to date with the shared server and uploads your changes.** (Server-focused, no team assumption.)

tombogle:
*[`assets/localization/es.json` line 148 at r3](https://reviewable.io/reviews/paranext/paranext-core/2589#-OyZmE9a5x1ejWpGne-n:-OyZmE9a5x1ejWpGne-o:b-rwrrq4) ([raw file](https://github.com/paranext/paranext-core/blob/cd82e94032af7fa92145b9c57256c91b1d8f9f2d/assets/localization/es.json#L148)):*
> ```JSON
>   "%extension_failed_to_start%": "No se pudo iniciar la extensión: {extensionName}",
>   "%firstRun_button_skipSync%": "Omitir sincronización automática",
>   "%firstRun_step_syncConsent_body%": "Tus proyectos están almacenados en un servidor compartido. Sincronizar mantiene tu trabajo actualizado y lo comparte con tu equipo.",
> ```

This mistake alerted me to my failure to push a branch with improved instructions for Claude regarding Spanish strings. That [PR](https://github.com/paranext/paranext-core/pull/2616) is up now. There are a few issues with this, but no point in trying to localize until we are sure we have the final form in English. Ping me once that is settled, and I can help ensure we have a usable Spanish version.

merchako:
> Go with "projects"

"Sync your projects" is good as long as that's actually what's being synced. "Data" or "work" can be usefully vague if there's a bunch of non-project data that also needs to be synced.

katherinejensen00:
Addressed in commit 00bf473572e.

**Body copy** (commit 00bf473572e):
- Heading: "Sync your data" → **"Sync your projects"** (vocabulary consistency — Tom's "go with projects")
- Body: → **"When working on shared projects, syncing updates your local copy and shares your changes with others."** (Tom's own suggestion — conditional phrasing, no team assumption)

**Spanish**: Removed all three `firstRun` Spanish keys. English body copy is now settled but Spanish will wait for Tom's review as offered. Spanish style guidance from PR #2616 is noted for when we do add it.

tombogle (changes requested):
> @tombogle reviewed 3 files and all commit messages, and made 7 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2589)* status: 8 of 25 files reviewed, 16 unresolved discussions (waiting on jolierabideau and katherinejensen00).

___
*[`assets/localization/en.json` line 32 at r1](https://reviewable.io/reviews/paranext/paranext-core/2589#-OyZCcCWCG1wzhNKzDiO:-Oy_1zhIB317zVQXdv4d:b3r7p8f) ([raw file](https://github.com/paranext/paranext-core/blob/b64a6a3a029e3c784070eefe7d61c373da2d9f80/assets/localization/en.json#L32)):*

NIT From Claude 

**`%firstRun_button_skip%` ("Skip setup") is now a fully orphaned key.** [en.json:32](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/assets/localization/en.json#L32) This branch removed the shell's generic Skip button and its `KEYS` reference, so this key has zero references in `src`/`lib`/tests at HEAD (verified: it _was_ referenced at the merge base). This is a **different** orphan than the `skipForNow` one already noted. Same disposition question as the placeholder key the earlier review flagged — remove it or handle it per the immutability/deprecation rule in `metadata.json`.
</blockquote></details>

I'm not 100% sure I understand the deprecation policy, but it seems that any localization key added within a sprint should be able to be removed/renamed within that same sprint, especially when it is clearly intended for a very narrow purpose. There is essentially no chance that we'd be breaking an extension.
___
*[`assets/localization/en.json` line 42 at r1](https://reviewable.io/reviews/paranext/paranext-core/2589#-OyZDPZ43AvM2QbtMk_1:-Oy_4BWh9vZ-xawj6rfW:ba9ki5u) ([raw file](https://github.com/paranext/paranext-core/blob/b64a6a3a029e3c784070eefe7d61c373da2d9f80/assets/localization/en.json#L42)):*

If not, we could consider an option like:
-   **Syncing updates your local copy and uploads your changes to the shared server.** (Doesn't assume projects or a team.)
-   **If you're working on shared projects, syncing updates your local copy and uploads your changes.** (Explicitly conditional.)
-   **When working on shared projects, syncing updates your local copy and shares your changes with others.** (Avoids assuming the user has a team.)
-   **Syncing keeps your local copy up to date with the shared server and uploads your changes.** (Server-focused, no team assumption.)
</blockquote></details>

English looks good. Here are your Spanish strings:
Al trabajar en proyectos compartidos, la sincronización actualiza la copia local y comparte sus cambios con otras personas.
___
*[`-- commits` line 132 at r1](https://reviewable.io/reviews/paranext/paranext-core/2589#-OyZlBJa1xX02FjFQyyY:-OyZlBJa1xX02FjFQyyZ:b-4yxsbi):*
When the dust settles and you prepare to squash, ensure that the commit comments are consistent with the final text to avoid confusion. (Could document the options that were tried and discarded if you deem it worthwhile.
___
*[`assets/localization/en.json` line 33 at r4](https://reviewable.io/reviews/paranext/paranext-core/2589#-Oy_6ftRDXAWstuhB9NQ:-Oy_6ftRDXAWstuhB9NR:bl132zw) ([raw file](https://github.com/paranext/paranext-core/blob/00bf473572e76b0c59122b47f3ebd72b1dc02fe8/assets/localization/en.json#L33)):*
> ```JSON
>   "%firstRun_button_skip%": "Skip setup",
>   "%firstRun_button_skipSync%": "Skip automatic sync",
>   "%firstRun_button_sync%": "Sync",
> ```

"%firstRun\_button\_skipSync%": "Omitir la sincronización automática",
"%firstRun\_button\_sync%": "Sincronizar"
___
*[`assets/localization/en.json` line 43 at r4](https://reviewable.io/reviews/paranext/paranext-core/2589#-Oy_4StICTJvuwRRA5BV:-Oy_4StICTJvuwRRA5BW:bpzv6up) ([raw file](https://github.com/paranext/paranext-core/blob/00bf473572e76b0c59122b47f3ebd72b1dc02fe8/assets/localization/en.json#L43)):*
> ```JSON
>   "%firstRun_step_syncConsent_body%": "When working on shared projects, syncing updates your local copy and shares your changes with others.",
>   "%firstRun_step_syncConsent_heading%": "Sync your projects",
> ```

Sincronizar proyectos
___
*[`assets/localization/en.json` line 390 at r4](https://reviewable.io/reviews/paranext/paranext-core/2589#-Oy_51L_-ZJHnBImEG-o:-Oy_51L_-ZJHnBImEG-p:bi9yam9) ([raw file](https://github.com/paranext/paranext-core/blob/00bf473572e76b0c59122b47f3ebd72b1dc02fe8/assets/localization/en.json#L390)):*
> ```JSON
>   "%settings_platform_comments_enabled_label%": "Comments enabled (experimental)",
>   "%settings_platform_firstRunComplete_label%": "Setup complete",
>   "%settings_platform_firstRunSyncSkipped_label%": "Sync skipped on first run",
> ```

Sincronización omitida durante la configuración inicial
___
*[`assets/localization/metadata.json` line 31 at r4](https://reviewable.io/reviews/paranext/paranext-core/2589#-Oy_5BsQ8eualQIC0Tmr:-Oy_5BsQ8eualQIC0Tms:b-uth01o) ([raw file](https://github.com/paranext/paranext-core/blob/00bf473572e76b0c59122b47f3ebd72b1dc02fe8/assets/localization/metadata.json#L31)):*
> ```JSON
>       "message": "Replaced by %firstRun_step_syncConsent_heading% and %firstRun_step_syncConsent_body% when the real sync-consent step was implemented in PT-4178."
>     }
>   },
> ```

This was added in this sprint and explicitly marked as a placeholder. I doubt we actually need deprecation info, but whatever.

tombogle (commented):
> @tombogle reviewed 3 files and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2589)* status: 11 of 25 files reviewed, 17 unresolved discussions (waiting on jolierabideau and katherinejensen00).

___
*[`src/renderer/services/first-run-store.ts` line 168 at r4](https://reviewable.io/reviews/paranext/paranext-core/2589#-Oy_93bl-KV1BPp1GpRN:-Oy_93bl-KV1BPp1GpRO:bnjajbz) ([raw file](https://github.com/paranext/paranext-core/blob/00bf473572e76b0c59122b47f3ebd72b1dc02fe8/src/renderer/services/first-run-store.ts#L168)):*
> ```XML
>           }
>         } catch (e) {
>           logger.warn(`Failed to re-persist platform.firstRunSyncSkipped: ${getErrorMessage(e)}`);
> ```

Not sure I totally grasp this, but is it really trying to "re-persist" (previously succeeded at persisting) or did it just "re-fail" (previously failed to persist and no it just failed again)?

katherinejensen00:
Addressed in commit edfb79f472b.

Right — "re-persist" was ambiguous. In this path the original write (in `completeFirstRun`) failed, so the self-heal is a retry of a failed write, not a second write of an already-persisted value. Renamed to `Self-heal write of platform.firstRunSyncSkipped failed` to make that explicit.

tombogle:
*[`lib/papi-dts/papi.d.ts` line 4599 at r6](https://reviewable.io/reviews/paranext/paranext-core/2589#-Oy_PMqD-EzVIiECOdyG:-Oy_PMqD-EzVIiECOdyH:b8dguhi) ([raw file](https://github.com/paranext/paranext-core/blob/e36608a8e0b2c792d59f466dfcb4687e78f4903c/lib/papi-dts/papi.d.ts#L4599)):*
> ```XML
>      * sync; manual Send/Receive is unaffected. Never reset by core.
>      */
>     'platform.firstRunSyncSkipped': boolean;
> ```

If it's not too late to rename this, I think the name is a bit confusing. Purports to be related only to the first-run, but if I understand correctly it really means that the user chose *during the first run* to skip automatic S/R during startup (but not during shutdown?).

katherinejensen00:
Done in commit 42b0f86f0a5: renamed to `platform.suppressStartupSync`. Agreed — the old name made it sound like a one-time first-run event rather than a permanent opt-out of automatic startup sync.

katherinejensen00 (commented):
Thank you Tom and Jolie for all of your help!!

> @katherinejensen00 made 19 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2589#-:-Oy_e6tlBQweQ0ZiUYZO:bpana2z)* status: 6 of 25 files reviewed, 18 unresolved discussions (waiting on jolierabideau and tombogle).

___
*[`-- commits` line 132 at r1](https://reviewable.io/reviews/paranext/paranext-core/2589#-OyZlBJa1xX02FjFQyyY:-Oy_c8YL4uI4eH3PU5QU:bbxjpln):*

When the dust settles and you prepare to squash, ensure that the commit comments are consistent with the final text to avoid confusion. (Could document the options that were tried and discarded if you deem it worthwhile.
</blockquote></details>

Good point. Will make sure the final is less confusing.
___
*[`.storybook/main.ts` line 169 at r1](https://reviewable.io/reviews/paranext/paranext-core/2589#-OyZD6hP63o1o8ifU-Sd:-Oy_cJHC93d7k-0-eKbv:bgzmklr) ([raw file](https://github.com/paranext/paranext-core/blob/b64a6a3a029e3c784070eefe7d61c373da2d9f80/.storybook/main.ts#L169)):*

NIT From Claude

**Redundant Storybook resolver config.** [.storybook/main.ts:162-183](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/.storybook/main.ts#L162-L183) Both a webpack `resolve.alias` exact-match key and a `NormalModuleReplacementPlugin` point at the same `@renderer/hooks/papi-hooks` stub. If the plugin is what actually works (because `TsconfigPathsPlugin` resolves `@renderer/*` first), the alias is effectively dead config. Confirm which is authoritative and drop the redundant one, or add a one-line note that both are genuinely required and must be kept in lockstep.
</blockquote></details>

You're right — the `resolve.alias` at line 169 is dead. As the comment block just below it explains, `TsconfigPathsPlugin` takes precedence for `@renderer/*` paths, so only the `NormalModuleReplacementPlugin` actually fires. The two entries also point to _different_ files (`renderer-hooks.ts` vs the real `renderer-papi-hooks.ts`). I'll remove the dead alias entry — and `renderer-hooks.ts`, which is only referenced by it.
___
*[`.storybook/papi-stubs/renderer-hooks.ts` at r1](https://reviewable.io/reviews/paranext/paranext-core/2589#-OyZD1DG40vIhL1jNiBe:-Oy_cNqo4InZaP_tsd5r:b1aps2d) ([raw file](https://github.com/paranext/paranext-core/blob/b64a6a3a029e3c784070eefe7d61c373da2d9f80/.storybook/papi-stubs/renderer-hooks.ts)):*

NIT From Claude

**Two Storybook stub surfaces overlap.** [.storybook/papi-stubs/renderer-hooks.ts](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/.storybook/papi-stubs/renderer-hooks.ts) Near-identical inert implementations of `useLocalizedStrings`/`useSetting`/`useData`/`useProjectData`/`useProjectSetting`/`useProjectDataProvider` already exist in `.storybook/papi-stubs/frontend-react.ts`. Since `@renderer/hooks/papi-hooks` re-exports the PAPI frontend-react hooks, the new stub could re-export the shared inert impls rather than re-declaring them.
</blockquote></details>

`renderer-hooks.ts` is the dead stub pointed to by the dead `resolve.alias` above — the active stub is `renderer-papi-hooks.ts`, which already takes exactly the approach you describe: it re-exports every real hook unchanged and overrides only `useLocalizedStrings`. Removing the alias removes `renderer-hooks.ts` with it, so no duplication remains.
___
*[`assets/localization/en.json` line 31 at r1](https://reviewable.io/reviews/paranext/paranext-core/2589#-OyZGbCK336ADjclSTOu:-Oy_cUfT-doAq-HHq0xU:b-gy12dp) ([raw file](https://github.com/paranext/paranext-core/blob/b64a6a3a029e3c784070eefe7d61c373da2d9f80/assets/localization/en.json#L31)):*

No Spanish equivalent. I think this comes after the UI locale choice, right? We should supply a Spanish string. (But first need to address the issue of what the UI should actually say in English.)
</blockquote></details>

English is now settled. Thanks for providing the Spanish strings in your later review — I'll add them to `es.json`.
___
*[`assets/localization/en.json` line 32 at r1](https://reviewable.io/reviews/paranext/paranext-core/2589#-OyZCcCWCG1wzhNKzDiO:-Oy_cR7_76Wl2qTjcypQ:b-9crrjq) ([raw file](https://github.com/paranext/paranext-core/blob/b64a6a3a029e3c784070eefe7d61c373da2d9f80/assets/localization/en.json#L32)):*

I'm not 100% sure I understand the deprecation policy, but it seems that any localization key added within a sprint should be able to be removed/renamed within that same sprint, especially when it is clearly intended for a very narrow purpose. There is essentially no chance that we'd be breaking an extension.
</blockquote></details>

Good catch, Jolie! and since this key was already on `main` before this PR (the old generic Skip button), the sprint-window exception doesn't apply. I'll add a `deprecationInfo` entry to `metadata.json`, the same way I handled `%firstRun_step_syncConsent_placeholder%`.

Tom, The sprint-window exception applies to keys that haven't shipped to `main` yet. `%firstRun_button_skip%` was already on `main` before this PR (it was the old generic "Skip setup" button), so I'll add a `deprecationInfo` entry to `metadata.json` to stay consistent with the policy.
___
*[`assets/localization/en.json` line 42 at r1](https://reviewable.io/reviews/paranext/paranext-core/2589#-OyZDPZ43AvM2QbtMk_1:-Oy_cr7BCeNMxdF4Ufag:b-saven2) ([raw file](https://github.com/paranext/paranext-core/blob/b64a6a3a029e3c784070eefe7d61c373da2d9f80/assets/localization/en.json#L42)):*

English looks good. Here are your Spanish strings:
Al trabajar en proyectos compartidos, la sincronización actualiza la copia local y comparte sus cambios con otras personas.
</blockquote></details>

Fixed. heading is now "Sync your projects" and body reads "When working on shared projects, syncing updates your local copy and shares your changes with others."

Thanks, Tom, for the Spanish localizations!
___
*[`assets/localization/en.json` line 33 at r4](https://reviewable.io/reviews/paranext/paranext-core/2589#-Oy_6ftRDXAWstuhB9NQ:-Oy_cmr23W_eImNqVQGW:b-9n6v5j) ([raw file](https://github.com/paranext/paranext-core/blob/00bf473572e76b0c59122b47f3ebd72b1dc02fe8/assets/localization/en.json#L33)):*

"%firstRun\_button\_skipSync%": "Omitir la sincronización automática",
"%firstRun\_button\_sync%": "Sincronizar"
</blockquote></details>

Thanks for all of these! Adding them to `es.json`. One note on line 390: the setting was renamed from `firstRunSyncSkipped` to `suppressStartupSync` in 42b0f86f0a5, so the corresponding label key is now `%settings_platform_suppressStartupSync_label%` ("Suppress automatic startup sync"). Does "Sincronización omitida durante la configuración inicial" still fit the new label, or would you like to revisit the wording given the rename?
___
*[`assets/localization/en.json` line 43 at r4](https://reviewable.io/reviews/paranext/paranext-core/2589#-Oy_4StICTJvuwRRA5BV:-Oy_cw-A7r-yJl1-3gl1:b-cn55et) ([raw file](https://github.com/paranext/paranext-core/blob/00bf473572e76b0c59122b47f3ebd72b1dc02fe8/assets/localization/en.json#L43)):*

Sincronizar proyectos
</blockquote></details>

Thanks! Added.
___
*[`assets/localization/en.json` line 390 at r4](https://reviewable.io/reviews/paranext/paranext-core/2589#-Oy_51L_-ZJHnBImEG-o:-Oy_cwnx7fqh3xkKJVkZ:b-cn55et) ([raw file](https://github.com/paranext/paranext-core/blob/00bf473572e76b0c59122b47f3ebd72b1dc02fe8/assets/localization/en.json#L390)):*

Sincronización omitida durante la configuración inicial
</blockquote></details>

Thanks! Added.
___
*[`assets/localization/metadata.json` line 31 at r4](https://reviewable.io/reviews/paranext/paranext-core/2589#-Oy_5BsQ8eualQIC0Tmr:-Oy_d-ub2V2nLNuWvl2u:b-ud264t) ([raw file](https://github.com/paranext/paranext-core/blob/00bf473572e76b0c59122b47f3ebd72b1dc02fe8/assets/localization/metadata.json#L31)):*

This was added in this sprint and explicitly marked as a placeholder. I doubt we actually need deprecation info, but whatever.
</blockquote></details>

Agreed it feels like overkill for a placeholder — but since it shipped to `main` before this PR, I'm keeping the entry for policy consistency. 
___
*[`lib/papi-dts/papi.d.ts` line 4599 at r6](https://reviewable.io/reviews/paranext/paranext-core/2589#-Oy_PMqD-EzVIiECOdyG:-Oy_dE375HTjKU9J_PlR:b-i10qlo) ([raw file](https://github.com/paranext/paranext-core/blob/e36608a8e0b2c792d59f466dfcb4687e78f4903c/lib/papi-dts/papi.d.ts#L4599)):*

If it's not too late to rename this, I think the name is a bit confusing. Purports to be related only to the first-run, but if I understand correctly it really means that the user chose *during the first run* to skip automatic S/R during startup (but not during shutdown?).
</blockquote></details>

You're right — "re-persist" was ambiguous. In this path the original write (in `completeFirstRun`) failed, so this is a retry of that failed write, not a second write of a value that already landed. Updated so the log now reads "Self-heal write of `platform.suppressStartupSync` failed" to make that explicit.
___
*[`src/renderer/components/first-run/first-run-shell.component.tsx` line 95 at r1](https://reviewable.io/reviews/paranext/paranext-core/2589#-OyZClNTEw52_KHOrLr3:-Oy_dNkZ0UOuY1HgJXwH:bis0flq) ([raw file](https://github.com/paranext/paranext-core/blob/b64a6a3a029e3c784070eefe7d61c373da2d9f80/src/renderer/components/first-run/first-run-shell.component.tsx#L95)):*

NIT From Claude

**Generic shell hardcodes `step === 'syncConsent'` in two places.** [first-run-shell.component.tsx:95-105](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/first-run/first-run-shell.component.tsx#L95-L105) Both `onSkip` and `stepManagesOwnFooter` name-check the specific step, coupling the step-agnostic shell to one step and splitting footer ownership (placeholder steps use the shell footer; `syncConsent` renders its own). Consider driving footer ownership from a step-declared capability (e.g. a `managesOwnFooter` flag on the step contract) so future real steps don't each require editing the shell. (This is the code counterpart to the stale contract-docs item in the earlier review.)
</blockquote></details>

Solid idea and preps us for future add-ons. Fixed.
___
*[`src/renderer/components/first-run/wizard-step-form.component.tsx` at r1](https://reviewable.io/reviews/paranext/paranext-core/2589#-OyZCWLS5aEBidDN7yfV:-Oy_deR-77v-oIa8IMo1:bqxada9) ([raw file](https://github.com/paranext/paranext-core/blob/b64a6a3a029e3c784070eefe7d61c373da2d9f80/src/renderer/components/first-run/wizard-step-form.component.tsx)):*

Do we need a Storybook story for this?
</blockquote></details>

`WizardStepForm` is a thin presentational layout — heading, body slot, error, and a button row. It's exercised through `SyncConsentStep`'s stories (Default, Syncing, WithBack), which together show all the button slots and the error path. I can add a standalone story if you'd prefer explicit coverage in isolation, but I didn't think it added much beyond what's already visible there.
___
*[`src/renderer/components/first-run/steps/sync-consent-step.component.tsx` line 23 at r1](https://reviewable.io/reviews/paranext/paranext-core/2589#-OyZDK7T0OTinVyO-VYk:-Oy_dhLpD69x0jI90ROK:b264sjk) ([raw file](https://github.com/paranext/paranext-core/blob/b64a6a3a029e3c784070eefe7d61c373da2d9f80/src/renderer/components/first-run/steps/sync-consent-step.component.tsx#L23)):*

NIT From Claude

**TSDoc describes an interstitial the wiring can't show.** [sync-consent-step.component.tsx:23-24](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/first-run/steps/sync-consent-step.component.tsx#L23-L24) The doc says "The sync progress interstitial (PT-4179) shows while sync runs," but `handleSync` `await`s `onSync()` to completion _before_ calling `onNext()` to advance to the `syncProgress` step — so the progress step is only reached after sync has already finished; the busy state is the local button `Spinner`. Reconcile the comment (or the wiring) when PT-4179 lands.
</blockquote></details>

Correct — that line was aspirational, not descriptive. I'll update the TSDoc to reflect current behavior: sync runs to completion with the local button spinner, then `onNext()` advances past the consent step. PT-4179 will reconcile the wiring when it lands.
___
*[`src/renderer/components/first-run/steps/sync-consent-step.component.tsx` line 71 at r1](https://reviewable.io/reviews/paranext/paranext-core/2589#-OyZDX53Bk_driDn6i4L:-Oy_dkbt4zHUwEQ8xd8B:b-6pzmhj) ([raw file](https://github.com/paranext/paranext-core/blob/b64a6a3a029e3c784070eefe7d61c373da2d9f80/src/renderer/components/first-run/steps/sync-consent-step.component.tsx#L71)):*

NIT From Claude

**Decline action styled as lowest-emphasis `ghost`.** [sync-consent-step.component.tsx:71](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/first-run/steps/sync-consent-step.component.tsx#L71) For a genuine consent decision where declining is a legitimate (and permanent) choice, a bounds-less `ghost` button lowers its discoverability relative to primary "Sync." Consider `outline` (as the Back button already uses) so the two consent options read as more balanced.
</blockquote></details>

Good point — `outline` makes the two choices feel like peers, which fits better for a permanent decision. I'll change "Skip automatic sync" to `variant="outline"`.
___
*[`src/renderer/components/first-run/steps/sync-consent-step.component.tsx` line 72 at r1](https://reviewable.io/reviews/paranext/paranext-core/2589#-OyZCO7A9Dih7_UFVv-Y:-Oy_dpAfBPtQw6vn-2Rb:bhm4h0n) ([raw file](https://github.com/paranext/paranext-core/blob/b64a6a3a029e3c784070eefe7d61c373da2d9f80/src/renderer/components/first-run/steps/sync-consent-step.component.tsx#L72)):*

I was starting to type this up myself, but I see Claude already found it. Upgrading to blocker.
</blockquote></details>

Fixed. Renamed to "Skip automatic sync", which accurately describes that startup auto-sync is suppressed permanently on all subsequent launches. Thanks Tom and Jolie!
___
*[`src/renderer/services/first-run-store.ts` line 168 at r4](https://reviewable.io/reviews/paranext/paranext-core/2589#-Oy_93bl-KV1BPp1GpRN:-Oy_dv6W9ZrfVWIXFmmh:b-i10qlo) ([raw file](https://github.com/paranext/paranext-core/blob/00bf473572e76b0c59122b47f3ebd72b1dc02fe8/src/renderer/services/first-run-store.ts#L168)):*

Not sure I totally grasp this, but is it really trying to "re-persist" (previously succeeded at persisting) or did it just "re-fail" (previously failed to persist and no it just failed again)?
</blockquote></details>

You're right — "re-persist" was ambiguous. In this path the original write (in `completeFirstRun`) failed, so this is a retry of that failed write, not a second write of a value that already landed. Updated so the log now reads "Self-heal write of `platform.suppressStartupSync` failed" to make that explicit.
___
*[`assets/localization/es.json` line 148 at r3](https://reviewable.io/reviews/paranext/paranext-core/2589#-OyZmE9a5x1ejWpGne-n:-Oy_e2F1DDYpCRE2eoVx:b269j2l) ([raw file](https://github.com/paranext/paranext-core/blob/cd82e94032af7fa92145b9c57256c91b1d8f9f2d/assets/localization/es.json#L148)):*

This mistake alerted me to my failure to push a branch with improved instructions for Claude regarding Spanish strings. I'm working on getting that PR up now. There are a few issues with this, but no point in trying to localize until we are sure we have the final form in English. Ping me once that is settled, and I can help ensure we have a usable Spanish version.
</blockquote></details>

Sounds good. Thanks for all your help with the localizations, Tom!

jolierabideau (approved):
> @jolierabideau reviewed 23 files and all commit messages, and resolved 7 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2589)* status: all files reviewed, 11 unresolved discussions (waiting on tombogle).

tombogle (changes requested):
> @tombogle reviewed 14 files and all commit messages, made 6 comments, and resolved 9 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2589)* status: 16 of 25 files reviewed, 7 unresolved discussions (waiting on jolierabideau and katherinejensen00).

___
*[`lib/papi-dts/papi.d.ts` line 4599 at r6](https://reviewable.io/reviews/paranext/paranext-core/2589#-Oy_PMqD-EzVIiECOdyG:-Oyb1GETAtpfLdB5TtMS:bfpgk7n) ([raw file](https://github.com/paranext/paranext-core/blob/e36608a8e0b2c792d59f466dfcb4687e78f4903c/lib/papi-dts/papi.d.ts#L4599)):*

You're right — "re-persist" was ambiguous. In this path the original write (in `completeFirstRun`) failed, so this is a retry of that failed write, not a second write of a value that already landed. Updated so the log now reads "Self-heal write of `platform.suppressStartupSync` failed" to make that explicit.
</blockquote></details>

Looking at this with fresh eyes and in light of the newly worded settings label, could we turn this into a positive setting instead? *If* this is ever going to be shown so that the user has a chance to change his mind (which seems pretty likely), it will be much easier for them to interpret if expressed positively: `platform.enableStartupSync` or `platform.syncOnStartup`. 
___
*[`src/renderer/components/first-run/first-run-shell.component.tsx` line 16 at r9](https://reviewable.io/reviews/paranext/paranext-core/2589#-Oyb3pgHAg561D7wk541:-Oyb3pgHAg561D7wk542:b-g94d0x) ([raw file](https://github.com/paranext/paranext-core/blob/fb9cc85cb8250ac728c962affbecfbb42dda973c/src/renderer/components/first-run/first-run-shell.component.tsx#L16)):*
> ```XML
> export const STEP_ORDER: FirstRunStep[] = ['language', 'identify', 'syncConsent', 'syncProgress'];
> 
> /** Default step bodies. Sibling tickets replace individual entries with their real step. */
> ```

I think this comment is now outdated
___
*[`src/renderer/components/first-run/first-run-shell.component.tsx` line 101 at r9](https://reviewable.io/reviews/paranext/paranext-core/2589#-Oyb51vY29Pq9BMdgTrS:-Oyb51vY29Pq9BMdgTrT:b-r6rug7) ([raw file](https://github.com/paranext/paranext-core/blob/fb9cc85cb8250ac728c962affbecfbb42dda973c/src/renderer/components/first-run/first-run-shell.component.tsx#L101)):*
> ```XML
>     () =>
>       StepComponent.managesSkip
>         ? () => runAction(() => completeFirstRun({ syncSkipped: true }))
> ```

We're abstracting the question of "which step is this", but then we're still subtly hardcoding the answer in `syncSkipped`. If this were some other step, it would need to be `someOtherStepSkipped`. Maybe should be `skippedStep` of type `integer | undefined`. As it is, it *seems* to be generic but really isn't. (I'm not even sure that in all conceivable startup wizards, skipping a step should automatically end the wizard, but that's an additional abstraction we don't currently need.)
___
*[`src/renderer/components/first-run/first-run-step-props.model.ts` line 22 at r9](https://reviewable.io/reviews/paranext/paranext-core/2589#-Oyb7gcG7oWmTUZPHZJK:-Oyb7gcG7oWmTUZPHZJL:b-iixmf7) ([raw file](https://github.com/paranext/paranext-core/blob/fb9cc85cb8250ac728c962affbecfbb42dda973c/src/renderer/components/first-run/first-run-step-props.model.ts#L22)):*
> ```XML
>   /**
>    * Skip the rest of setup and finish (persists `platform.suppressStartupSync = true`). Provided
>    * only to steps that declare {@link FirstRunStepComponent.managesSkip}. The callback may be async;
> ```

This comment also reveals the generic vs. specific confusion. If multiple steps were to allow skipping, we couldn't say here specifically which setting would be persisted.
___
*[`assets/localization/en.json` line 396 at r9](https://reviewable.io/reviews/paranext/paranext-core/2589#-Oyb0gp69Sbfn3Zl2_hQ:-Oyb0gp69Sbfn3Zl2_hR:by2k20z) ([raw file](https://github.com/paranext/paranext-core/blob/fb9cc85cb8250ac728c962affbecfbb42dda973c/assets/localization/en.json#L396)):*
> ```JSON
>   "%settings_platform_comments_enabled_label%": "Comments enabled (experimental)",
>   "%settings_platform_firstRunComplete_label%": "Setup complete",
>   "%settings_platform_suppressStartupSync_label%": "Suppress automatic startup sync",
> ```

This is an improvement, but if we could turn the setting positive, we could avoid the semi-jargony "Suppress" and make this easier to understand (and change later if actually displayed as a mutable setting): Perform automatic sync on startup
___
*[`assets/localization/es.json` line 520 at r9](https://reviewable.io/reviews/paranext/paranext-core/2589#-Oyb24Mk8ZryPCNAuRjz:-Oyb24Mk8ZryPCNAuRk-:b-epojn3) ([raw file](https://github.com/paranext/paranext-core/blob/fb9cc85cb8250ac728c962affbecfbb42dda973c/assets/localization/es.json#L520)):*
> ```JSON
>   "%settings_platform_requestTimeout_description%": "Número de segundos que se permite para que las solicitudes internas se completen. Las solicitudes que tarden más de este tiempo serán abortadas. Establezca en 0 para nunca abortar solicitudes.",
>   "%settings_platform_requestTimeout_label%": "Tiempo de espera de la solicitud",
>   "%settings_platform_suppressStartupSync_label%": "Sincronización omitida durante la configuración inicial",
> ```

Given the new English (which is an improvement): Suprimir la sincronización automática al iniciar
(but see my comment about making it positive instead:
- Realizar sincronización automática al iniciar
- OR
- Sincronizar automáticamente al iniciar)

katherinejensen00 (commented):
Thanks for helping to make the localizations and the code better, Tom!

> @katherinejensen00 made 7 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2589#-:-OydKMR02IgcpYHf-TUa:b-pjpunw)* status: 5 of 25 files reviewed, 7 unresolved discussions (waiting on jolierabideau and tombogle).

___
*[`assets/localization/en.json` line 396 at r9](https://reviewable.io/reviews/paranext/paranext-core/2589#-Oyb0gp69Sbfn3Zl2_hQ:-OydJrPC6HafrG4U2363:bnhdz4y) ([raw file](https://github.com/paranext/paranext-core/blob/fb9cc85cb8250ac728c962affbecfbb42dda973c/assets/localization/en.json#L396)):*

This is an improvement, but if we could turn the setting positive, we could avoid the semi-jargony "Suppress" and make this easier to understand (and change later if actually displayed as a mutable setting): Perform automatic sync on startup
</blockquote></details>

Changed to `"Perform automatic sync on startup"`. Key renamed from `%settings_platform_suppressStartupSync_label%` to `%settings_platform_syncOnStartup_label%` to match the setting rename.
___
*[`assets/localization/es.json` line 520 at r9](https://reviewable.io/reviews/paranext/paranext-core/2589#-Oyb24Mk8ZryPCNAuRjz:-OydJmH1Ae4zWyHCSA8D:b-epo627) ([raw file](https://github.com/paranext/paranext-core/blob/fb9cc85cb8250ac728c962affbecfbb42dda973c/assets/localization/es.json#L520)):*

Given the new English (which is an improvement): Suprimir la sincronización automática al iniciar
(but see my comment about making it positive instead:
- Realizar sincronización automática al iniciar
- OR
- Sincronizar automáticamente al iniciar)
</blockquote></details>

Updated to `"Sincronizar automáticamente al iniciar"`. Thanks for the Spanish help!
___
*[`lib/papi-dts/papi.d.ts` line 4599 at r6](https://reviewable.io/reviews/paranext/paranext-core/2589#-Oy_PMqD-EzVIiECOdyG:-OydKJ5D2nYbnRDpJYN0:b-6tl212) ([raw file](https://github.com/paranext/paranext-core/blob/e36608a8e0b2c792d59f466dfcb4687e78f4903c/lib/papi-dts/papi.d.ts#L4599)):*

Looking at this with fresh eyes and in light of the newly worded settings label, could we turn this into a positive setting instead? *If* this is ever going to be shown so that the user has a chance to change his mind (which seems pretty likely), it will be much easier for them to interpret if expressed positively: `platform.enableStartupSync` or `platform.syncOnStartup`.
</blockquote></details>

Fixed. Renamed to `platform.syncOnStartup` with default `true` (sync enabled). Flipped the conditional in `startup-tasks.ts` to `=== false`, flipped the default in `core-settings-info.data.ts`, and regenerated `papi.d.ts`. First-run store now sets `platform.syncOnStartup` to `false` (rather than `suppressStartupSync` to `true`) when the user skips.
___
*[`src/renderer/components/first-run/first-run-shell.component.tsx` line 16 at r9](https://reviewable.io/reviews/paranext/paranext-core/2589#-Oyb3pgHAg561D7wk541:-OydK80b1wTrBnbP6-n0:b-886gsx) ([raw file](https://github.com/paranext/paranext-core/blob/fb9cc85cb8250ac728c962affbecfbb42dda973c/src/renderer/components/first-run/first-run-shell.component.tsx#L16)):*

I think this comment is now outdated
</blockquote></details>

Fixed
___
*[`src/renderer/components/first-run/first-run-shell.component.tsx` line 101 at r9](https://reviewable.io/reviews/paranext/paranext-core/2589#-Oyb51vY29Pq9BMdgTrS:-OydK9z81qk8C662xerv:b-2ohnml) ([raw file](https://github.com/paranext/paranext-core/blob/fb9cc85cb8250ac728c962affbecfbb42dda973c/src/renderer/components/first-run/first-run-shell.component.tsx#L101)):*

We're abstracting the question of "which step is this", but then we're still subtly hardcoding the answer in `syncSkipped`. If this were some other step, it would need to be `someOtherStepSkipped`. Maybe should be `skippedStep` of type `integer | undefined`. As it is, it *seems* to be generic but really isn't. (I'm not even sure that in all conceivable startup wizards, skipping a step should automatically end the wizard, but that's an additional abstraction we don't currently need.)
</blockquote></details>

Good point. Changed to `completeFirstRun({ skippedStep: step })` where `step` is the shell's current step state — no hardcoded string. The type now uses the existing `FirstRunStep` union.
___
*[`src/renderer/components/first-run/first-run-step-props.model.ts` line 22 at r9](https://reviewable.io/reviews/paranext/paranext-core/2589#-Oyb7gcG7oWmTUZPHZJK:-OydJxIVBgrJNQPcQSbT:box9g75) ([raw file](https://github.com/paranext/paranext-core/blob/fb9cc85cb8250ac728c962affbecfbb42dda973c/src/renderer/components/first-run/first-run-step-props.model.ts#L22)):*

This comment also reveals the generic vs. specific confusion. If multiple steps were to allow skipping, we couldn't say here specifically which setting would be persisted.
</blockquote></details>

The TSDoc now no longer references `suppressStartupSync` — the model describes the generic step-props contract and doesn't mention the setting at all.

tombogle (commented):
> @tombogle reviewed 22 files, made 2 comments, and resolved 6 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2589)* status: all files reviewed, 3 unresolved discussions (waiting on katherinejensen00).

___
*[`src/renderer/components/first-run/first-run-shell.component.tsx` line 154 at r10](https://reviewable.io/reviews/paranext/paranext-core/2589#-OydP6hM8Ol-ZViMOQJh:-OydP6hM8Ol-ZViMOQJi:b-sde306) ([raw file](https://github.com/paranext/paranext-core/blob/3a962f2992c0742ab042dbf09ac2ccebe52774c7/src/renderer/components/first-run/first-run-shell.component.tsx#L154)):*
> ```XML
>           // reason, the shell will need to accept a skip-label callback from that step.
>           <Button variant="ghost" onClick={onSkip} disabled={isBusy}>
>             {strings['%firstRun_button_skipSync%']}
> ```

Not ideal, but okay
___
*[`src/renderer/components/first-run/wizard-step-form.component.tsx` line 7 at r10](https://reviewable.io/reviews/paranext/paranext-core/2589#-OydQkyq-GvAiV0lOle-:-OydQkyq-GvAiV0lOle0:brm6rml) ([raw file](https://github.com/paranext/paranext-core/blob/3a962f2992c0742ab042dbf09ac2ccebe52774c7/src/renderer/components/first-run/wizard-step-form.component.tsx#L7)):*
> ```XML
>   children: ReactNode;
>   error?: string;
>   /** Back button node, rendered on the left. Omit when no back navigation is available. */
> ```

Ideally use "leading" and "trialing". But this is just a comment, so not a huge deal.

tombogle (approved):
> @tombogle reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2589)* status: all files reviewed, 3 unresolved discussions (waiting on katherinejensen00).

PR #11, opened 2023-01-31
Add debugging to GHA workflows

irahopkinson:
- [manually trigger a workflow](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow). This option will show at step 5.
- in the GH Actions UI Console use the SSH command printed every 5 seconds to connect to the build agent.
- only the actor can SSH. Must have public key registered on GH.

PR #2590, opened 2026-07-23
PT-4020 improve tooltip styling for "Delete Verse Number" two-step confirmation

tombogle:
# Code Review Summary

**Branch**: pt-4020-improve-tooltip-styling

**Base**: origin/main

**Date**: 2026-07-27

**Files changed**: 34

## Overview

This branch started as a fix for the styling on the two-step verse-number-deletion confirmation
tooltip (PT-4020), then grew in scope as the underlying "press key again to confirm" mechanism
proved non-trivial enough to be worth generalizing. The extension's `verse-delete-tooltip`
directory was renamed to `two-step-delete-tooltip`, and its rendering was extracted into a new,
reusable `DestructiveKeyConfirmation` component in `platform-bible-react` — both to make rapid
UX-driven style iteration easier and to make the pattern available to other extensions. A new
`keyboard-util.ts` helper (`getLocalizeKeyForPhysicalKey`) was added to `platform-bible-utils` so
physical key names can be localized generically rather than duplicated per caller. Supporting this
also required customizing the shadcn `tooltip.tsx` (a new `arrowClassName` prop and reworked
arrow positioning/clip-path math) so tooltips could carry a destructive-themed border, per PT-4236.
Work happened across two sessions on this branch: this paranext-core session, and a separate
session working directly in the `platform-bible-react` (PBR) library folder.

## API Changes

- `lib/platform-bible-react`: Added new export `DestructiveKeyConfirmation` (default + named) and
  its `DestructiveKeyConfirmationProps` type, from `components/basics/destructive-key-confirmation.component.tsx`.
- `lib/platform-bible-react`: `TooltipContent` (in `shadcn-ui/tooltip.tsx`) gained a new optional
  prop `arrowClassName?: string` (additive/non-breaking); the arrow's translate/clip-path CSS was
  also reworked (visual-only, not a signature change — it only has a visible effect when a caller
  uses `arrowClassName` to give the arrow a border, as `DestructiveKeyConfirmation` does).
- `lib/platform-bible-utils`: Added new export `getLocalizeKeyForPhysicalKey(key: NameablePhysicalKey): LocalizeKey`
  and new type export `NameablePhysicalKey` (`'Backspace' | 'Delete'`), from `keyboard-util.ts`.
- `lib/platform-bible-react/dist` and `lib/platform-bible-utils/dist`: rebuilt and committed
  consistently with the above `src` changes.
- No changes to `papi.d.ts`, `@papi/*` modules, or any extension `src/types/*.d.ts` files.

- [ ] ~~`extensions/src/platform-scripture-editor/contributions/localizedStrings.json`: the rename
  of `verseDelete_deleteSelection`/`verseDelete_removeVerseMarker` to `twoStepDelete_*` deleted the
  old, previously-existing keys outright instead of keeping them alongside the new ones (or adding
  a `metadata.fallbackKey` mapping), which on its face conflicts with the project's "existing
  localization keys are immutable" rule.~~ _(author's explanation: these keys were added earlier
  this week and have never shipped in a released version — the "never delete published keys" rule
  protects already-released strings, which doesn't apply here. The old keys were specific enough to
  this exact context that reuse elsewhere is highly unlikely, and the team has allowed this kind of
  rename for not-yet-released keys before.)_

Author confirmed this was intentional and gave a clear rationale; no code change made.

- [x] **`lib/platform-bible-utils/src/keyboard-util.ts`'s `getLocalizeKeyForPhysicalKey` always
  returned the localized *word* for a key ("Backspace"/"Delete") with no macOS-specific symbol
  (`⌫`/`⌦`) branching**, contrary to the Keyboard shortcuts guideline's "Preferred representations
  by OS" table. _(fixed during review: added `getConfirmingKeyDisplayLabel(intent, localizedWord, isMac)`
  to `two-step-delete-tooltip.utils.ts`, wired to `isMacOs()` from `platform-bible-react` in the
  overlay component, with new unit tests in `two-step-delete-tooltip.utils.test.ts` written
  test-first.)_
- [x] **`destructive-key-confirmation.component.tsx` forced `tw:max-w-none tw:whitespace-nowrap`**,
  which could clip or overflow the hint on a narrow webview with longer-locale text (e.g. Spanish),
  instead of allowing responsive wrapping. _(fixed during review: removed both overrides so the
  tooltip falls back to `TooltipContent`'s default `max-w-xs` + normal wrapping — single line when
  it fits comfortably, wraps otherwise, per the author's stated intent.)_
- [x] **`two-step-delete-tooltip-overlay.component.tsx` reimplemented scroll-container discovery
  inline** (walking `parentElement` ancestors checking `overflowY`) instead of reusing the existing
  `findScrollContainer(fromElement, { requireOverflow })` helper in `editor-dom.util.ts`, which the
  sibling `ParagraphMarkerTooltipOverlay` (that this component's own doc comment says it "mirrors")
  already uses for the identical purpose. _(fixed during review: switched to
  `findScrollContainer(positionAnchor, { requireOverflow: false }) ?? positionAnchor`.)_

Author agreed with all three and requested fixes; all three were implemented, tested, and verified
via typecheck/lint/test during the review.

### Minor — Consider

- [ ] ~~`destructive-key-confirmation.stories.tsx`'s `SelectionDeletion` story text ("Press {key}
  again to delete **this** selection") drifts slightly from the shipped copy ("Press {key} again to
  delete selection").~~ _(author's explanation: the component is designed to be reused with
  different text; keeping the story and shipped text identical is a nice-to-have for quick
  illustrative screenshots, not a requirement. There's a genuine UX tension between brevity and
  clarity here — dropping articles/pronouns can be harder for non-native English speakers — and the
  exact production wording may still change after further UX review.)_
- [ ] ~~`assets/localization/es.json`: the Spanish translation for the renamed
  `removeVerseMarker` key changed wording ("quitar el marcador de versículo" → "borrar este número
  de versículo") alongside the key rename.~~ _(author's explanation: the Storybook version shows a
  literal `\v` marker in the text, so that translation was chosen to match it; the real app doesn't
  show the raw marker, and the Spanish copy may change further after UX testing/review.)_

Author reviewed both and declined to change either, with reasoned context recorded above.

#### Shared Regions Modified

None.

#### Extension Config Changes

None.

## Positive Observations

- The new `DestructiveKeyConfirmation`/`getLocalizeKeyForPhysicalKey` public API additions are
  clean, additive, non-breaking, with thorough TSDoc and correctly exported types.
- `dist/` artifacts for both `platform-bible-react` and `platform-bible-utils` were rebuilt and
  committed in lockstep with the `src/` changes.
- The shadcn `tooltip.tsx` customization is fully covered by detailed `// CUSTOM` comments
  explaining what/why, including an honest "KNOWN LIMITATION" callout for the unresolved
  left/right-side arrow clip-path math.
- Good test coverage for the new pure-function utilities (`keyboard-util.test.ts`,
  `two-step-delete-tooltip.utils.test.ts`), including edge cases (missing/unrecognized attributes,
  scroll-offset math).
- The `verse-delete-tooltip` → `two-step-delete-tooltip` rename left no stray references
  (`VerseDeleteTooltipOverlay`, old directory name) anywhere in the extension.
- New/renamed exports were inserted in correct alphabetical order in
  `lib/platform-bible-react/src/index.ts` and `lib/platform-bible-utils/src/index.ts`.
- `destructive-key-confirmation.stories.tsx` ships a thorough story set (Default,
  SelectionDeletion, NoArrow, Closed) including a locale-simulation control and measured (not
  hand-placed) anchor positioning mirroring the real component's technique.
- `DestructiveKeyConfirmation` correctly follows the Tooltips guideline (uses `Kbd`, not
  parentheses) and the Applying Changes guideline — the two-step "press again to confirm" pattern
  is a well-chosen, lightweight friction mechanism for an irreversible destructive action.
- The destructive-tinted styling (`text-destructive`/`border-destructive`/`bg-destructive/10`) is
  applied with a clear semantic justification rather than arbitrary custom styling.

## Interview Notes

**Stated purpose**: from the user's perspective, the essence of this change is a "simple"
styling/presentation fix to the two-step verse-number-deletion confirmation tooltip (PT-4020,
PT-4236). Internally, though, the scope grew: the tooltip's rendering was extracted into a
reusable `DestructiveKeyConfirmation` component in `platform-bible-react`, both as a pragmatic move
(faster style iteration, more accessible to UX) and a forward-looking one (making the "press again
to confirm" pattern available to other extensions, since the anchor/arm/detect mechanics proved
non-trivial). The arm/anchor/detect mechanisms themselves are essentially unchanged from before —
just refactored to live behind the new component's boundary. Supporting a bordered tooltip variant
required customizing the shadcn `tooltip.tsx` (new `arrowClassName` prop, reworked arrow
positioning). Part of the utility logic moved into `platform-bible-utils` specifically to make key
naming both localizable and generically reusable outside this one extension.

**Findings discussion**:
- The Critical localization-key-deletion finding was intentional and well-justified: the renamed
  keys were added earlier this week and never shipped in a release, so the "never delete published
  keys" rule doesn't apply; precedent exists for this kind of pre-release rename.
- All three Important findings (macOS key symbols, tooltip wrap behavior, scroll-container
  duplication) were acknowledged and fixed live during the review.
- Both Minor findings were dismissed with clear, specific reasoning (see Findings section above)
  rather than left unaddressed by default.

**Design understanding**: asked given the branch's size (34 files). The author gave a clear,
confident walkthrough covering why the component was extracted, why the shadcn tooltip needed
customization, and why utility code moved to `platform-bible-utils` — no gaps or deferrals to AI
were observed; the author demonstrated full ownership of the design.

**Unresolved items**: none.

**Wrap-up**: no additional context or concerns raised.

## In-Review Quality Check

All three Important-finding fixes were verified after implementation:

- **Library rebuild**: `lib/platform-bible-react/src/` changed (the tooltip `max-w`/`whitespace`
  fix), so `npm run build:platform-bible-react` was run; confirmed `dist/index.js/cjs/d.ts`,
  `experimental.js/cjs`, and the content-hashed `resizable-*` chunk files actually changed.
- **`npm run typecheck`**: passed cleanly across all workspaces after the rebuild.
- **`npm run lint`**: one auto-fixable issue surfaced (a `prettier/prettier` warning on a comment
  starting with the intentionally-lowercase "macOS's"); `eslint --fix` would have "fixed" it by
  incorrectly capitalizing to "MacOS's" (a `prettier-plugin-jsdoc` sentence-capitalization
  artifact), so the comment was reworded instead to avoid the false-positive trigger while keeping
  correct branding. Lint is clean after the reword. Two pre-existing, unrelated `no-console`
  warnings in `use-effective-resource-reference-list.hook.ts` were left untouched.
- **Format** (changed files only): no source files needed reformatting.
- **`npm test`**: full suite passed with 0 failures (321 test files); specifically confirmed
  `keyboard-util.test.ts`, `two-step-delete-tooltip.utils.test.ts`, and
  `destructive-key-confirmation.stories.tsx` all pass.

No unfixable failures.

## Suggested Review Focus

- [ ] Confirm the reviewer agrees that the deleted (not-yet-released) localization keys are fine to
  drop without a fallback mapping, given they were added this week and never shipped.
- [ ] Spot-check the new macOS key-symbol behavior (`⌫`/`⌦` via `getConfirmingKeyDisplayLabel`) on
  an actual Mac, since it couldn't be visually verified in this session.
- [ ] Sanity-check the tooltip wrap behavior change (removed `max-w-none`/`whitespace-nowrap`) with
  a long Spanish string on a narrow webview to confirm it wraps cleanly rather than looking broken
  mid-`Kbd`.
- [ ] Note that both dismissed Minor findings (Storybook story text, Spanish wording) may still be
  revisited pending further UX review — flagged here for awareness, not action.

---

This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2590)


tombogle:
image

tombogle:
image

tombogle:
@merchako I'm working to see if I can get the "arrow" to appear so that the tooltip points to the verse number in question. (It's mostly working now.) Assuming that is possible, and maybe even if not, I'm thinking maybe the message should change to say "...again to remove _this_ verse...". Since they can't actually see the *marker* and since this is a true *deletion*, not merely removing/detaching, I'd lean toward this wording:
`Press <blah> again to delete this verse number.`

tombogle:
Not yet pushed, but here's how it looks with the arrow in storybook:
image
And in the app:
image

merchako:
1. “This”  isn’t needed.
2.  Seeing the outline of diamond that’s used for the arrow is a problem.

merchako:
Hey @tombogle — re: pointing the tooltip nub at a *meaningful* part of the selection (not the whole bounding rectangle). This is a **geometry** concern sitting next to your **styling** PR, so it's fine as a follow-up rather than something #2590 has to carry. Sharing a diagnosis + a drop-in approach so you (and your agents) don't have to re-derive it.

## Why the nub can't aim today

`sync()` anchors the tooltip to `.verse-selected` via `computeAnchorRect` → `getBoundingClientRect()` — one axis-aligned box:

```ts
// verse-delete-tooltip.utils.ts — computeAnchorRect()
const markerRect = marker.getBoundingClientRect();   // ONE box; no per-line info
return { top: …, left: …, width: markerRect.width, height: markerRect.height };
```

For a multi-verse range there can be several `.verse-selected` spans and `querySelector` takes the first; either way a single bounding box can't express where a **ragged multi-line selection starts** (mid-first-line) or **ends** (mid-last-line). That's the editor dev's "I just get a rectangle" problem exactly.

## The good news: the geometry is already reachable in core

I verified against the editor's **merged `main`** (`libs/shared-react/src/plugins/usj/StructureKeyboardPlugin.tsx`):

- It publishes the armed signal your component reads — `root.setAttribute("data-verse-delete-intent"/"-kind", …)` and the `verse-selected` class (lines ~250, ~49–50).
- For the **selection** kind it **arms without touching the selection** — the branch only calls `setArmedState({ kind: "selection", … })` + `preventDefault()`; there's **no** `$setSelection`/`removeText` (those live in the *verse/paragraph* branch). Comment in source: *"Range/text selection that includes a verse marker: arm without altering the selection."*

So while armed, the **live DOM Selection is intact**, in the **same document** your overlay already queries (the editor is a Lexical `contenteditable`, not a nested iframe — that's also why `positionAnchor.querySelector('.verse-selected')` works today). That gives you real per-line geometry with **no editor change**:

```ts
const range = positionAnchor.ownerDocument.getSelection()?.getRangeAt(0);
range.getClientRects();   // one rect PER LINE — rects[0] = start line, rects.at(-1) = end line
```

Those are precisely the "top-left of start / bottom-right of end" corners we want to aim at.

> Sanity check if your yalc-linked editor differs from `main`: select across a verse, arm it, and in the webview console run
> `const s=getSelection(); console.log(s.rangeCount, s.isCollapsed, s.getRangeAt(0)?.getClientRects().length)`.
> Expect `rangeCount ≥ 1`, `isCollapsed false`, `length ≥ 1`. (If it ever comes back collapsed, the endpoints are still available editor-side as `anchor`/`focus` `PointSnapshot`s in `ArmedDelete` — ping me.)

## The fix

Keep the marker anchor for `kind === 'verse'` (it's a `NodeSelection` with no text range — aiming at the chip is right). For `kind === 'selection'`, build a zero-width caret at the reading-end and let your existing invisible `TooltipTrigger` do the rest.

New util (mirrors `computeAnchorRect`; add a case to `verse-delete-tooltip.utils.test.ts`):

```ts
/**
 * Zero-width caret AnchorRect at the selection's reading-END (preferEnd) or START, in
 * positionAnchor content coords. Reads the LIVE DOM selection (intact while armed). Returns
 * undefined when there's no usable range — including a display:none / inactive tab, where
 * getClientRects() yields zeros (see .claude/rules/cross-view-sync-hidden-views.md).
 */
export function computeSelectionCaretRect(
  positionAnchor: HTMLElement,
  preferEnd: boolean,
): AnchorRect | undefined {
  const selection = positionAnchor.ownerDocument.getSelection();
  if (!selection || selection.rangeCount === 0 || selection.isCollapsed) return undefined;

  const rects = Array.from(selection.getRangeAt(0).getClientRects()).filter((r) => r.height > 0);
  if (rects.length === 0) return undefined; // hidden tab → zero rects; bail

  const line = preferEnd ? rects[rects.length - 1] : rects[0];
  const isRtl = getComputedStyle(positionAnchor).direction === 'rtl';
  const caretX = preferEnd === !isRtl ? line.right : line.left; // reading-end: right in LTR, left in RTL

  const anchor = positionAnchor.getBoundingClientRect();
  return { top: line.top - anchor.top, left: caretX - anchor.left, width: 0, height: line.height };
}
```

In `sync()`:

```ts
const hint = readArmedHint(root);
if (!hint) { setArmed(undefined); return; }

if (hint.kind === 'selection') {
  const rect = computeSelectionCaretRect(positionAnchor, /* preferEnd */ true);
  if (!rect) { setArmed(undefined); return; }
  setArmed({ ...hint, ...rect });
  return;
}

const marker = positionAnchor.querySelector<HTMLElement>('.verse-selected');
if (!marker) { setArmed(undefined); return; }
setArmed({ ...hint, ...computeAnchorRect(marker, positionAnchor) });
```

**Arrow note:** your unpushed change enables the nub; today the committed code is `align="start"` / `showArrow={false}`. With a zero-width caret trigger, `align="center"` puts the arrow on the caret x, and keeping `side="bottom"` makes it point at the **end**; Radix collision-flips it above when there's no room — which satisfies "prefer the end." (Optional full rule — point at *start* when it flips above — read the content's resolved `data-side` and pass `preferEnd = side === 'bottom'`. Not required.)

**Watch-outs:** keep the kind split (verse = NodeSelection, no range); the `rects.length === 0` guard doubles as the hidden-view guard; Radix tooltips don't steal focus, so opening won't clear the selection; coords are viewport-space like `markerRect` today, so the existing "subtract positionAnchor rect, no scrollTop" reasoning carries over.

---

<details><summary>📋 Prompt for an AI coding agent (copy-paste)</summary>

```
Implement selection-aware nub placement for the verse-delete tooltip in paranext-core.

FILES (paranext-core):
- extensions/src/platform-scripture-editor/src/verse-delete-tooltip/verse-delete-tooltip.utils.ts
- .../verse-delete-tooltip.utils.test.ts
- .../verse-delete-tooltip-overlay.component.tsx

BACKGROUND (verified against editor main): while a "selection" kind two-step delete is armed, the
editor arms WITHOUT altering the selection, so the live DOM Selection is intact in the same document
the overlay renders in. The overlay currently anchors to `.verse-selected` via computeAnchorRect ->
getBoundingClientRect(), a single box that loses where a ragged multi-line selection starts/ends.

OPTIONAL SANITY CHECK before coding (if the linked editor build might differ from main): arm a
selection and evaluate in the webview `getSelection().rangeCount`, `.isCollapsed`,
`.getRangeAt(0).getClientRects().length`; expect >=1, false, >=1. If collapsed, STOP and report.

TASK:
1. Add `computeSelectionCaretRect(positionAnchor, preferEnd)` to the utils file:
   - `const selection = positionAnchor.ownerDocument.getSelection()`
   - return undefined if no selection || rangeCount === 0 || isCollapsed
   - `const rects = Array.from(range.getClientRects()).filter(r => r.height > 0)`; return undefined
     if empty (this also covers display:none hidden tabs)
   - `const line = preferEnd ? rects.at(-1) : rects[0]`
   - `const isRtl = getComputedStyle(positionAnchor).direction === 'rtl'`
   - `const caretX = preferEnd === !isRtl ? line.right : line.left`
   - return a zero-width AnchorRect in positionAnchor content coords (subtract
     positionAnchor.getBoundingClientRect()), height = line.height
2. In the overlay's sync(), branch on hint.kind: for 'selection' use
   computeSelectionCaretRect(positionAnchor, true) (bail to undefined if it returns undefined); for
   everything else keep the existing `.verse-selected` + computeAnchorRect path unchanged.
3. Keep side="bottom". If the arrow is enabled, use align="center" so it sits on the caret x.
4. Add a unit test mirroring the existing computeAnchorRect tests: mock a selection with multiple
   getClientRects; assert the rect sits at the last line's trailing edge (LTR) and the first line's
   edge when preferEnd=false; assert RTL flips the edge; assert empty rects -> undefined.
5. Do NOT touch the editor repo. Do NOT change the 'verse' kind's anchoring. Run
   `npm run typecheck && npm run lint` and the utils test file before finishing.

CONSTRAINTS: match surrounding code style and TSDoc density; no new deps; the rects.length===0 guard
is the documented hidden-view handling (.claude/rules/cross-view-sync-hidden-views.md).
```

</details>

tombogle:
> pointing the tooltip nub at a _meaningful_ part of the selection (not the whole bounding rectangle)

Noted as a new nice-to-have subtask in Jira.

tombogle:
Storybook Delete *this* selection - top/center
image

tombogle:
Storybook Spanish Backspace selection - bottom/start
image

tombogle:
https://github.com/user-attachments/assets/9f1a135a-c8c4-4e70-b672-362626edefaf

katherinejensen00 (commented):
## 🤖 Claude AI Review — PR #2590

Multi-perspective review covering code quality, accessibility, architecture, clarity, scope, and tests. Four warnings posted as inline comments; eleven suggestions listed here.

| Severity | Count |
|----------|-------|
| ⚠️ Warning (inline) | 4 |
| 💡 Suggestion (below) | 11 |
| 🚫 Blocking | 0 |

---

### 💡 Suggestions

**1. `lib/platform-bible-react/src/stories/basics/destructive-key-confirmation.stories.tsx:88` — `useLayoutEffect` missing `anchorPreview` dep**
The dep array is `[selectionStart, selectionEnd]` only. Switching the Storybook control from `'marker'` to `'selection'` without changing an offset never triggers re-measurement — `measuredRect` stays `undefined`, `isAnchorReady` stays `false`, and the tooltip stays permanently closed. Fix: add `anchorPreview` to the dep array.

**2. `lib/platform-bible-react/src/components/basics/destructive-key-confirmation.component.tsx:29` — `side` prop TSDoc missing KNOWN LIMITATION**
The TSDoc says only "Tooltip placement side. Defaults to 'bottom'." — no mention that `'left'`/`'right'` produce a broken/borderless arrow. Since this is a public `platform-bible-react` export, a one-sentence caveat in the TSDoc would prevent silent style degradation for future consumers who don't read implementation comments.

**3. `lib/platform-bible-utils/src/keyboard-util.ts:39` — localization asset coupling + single-word constraint undocumented**
Adding a member to `NameablePhysicalKey` silently produces a `%physicalKey_<newkey>%` key with no translation until every locale file is manually updated. The TSDoc invites growth ("Add to this union as UI needs") but doesn't note that only single-word key names work: a compound key like `'ArrowLeft'` would produce `%physicalKey_arrowleft%` (mashed, unmatched). A test asserting every union member has a `%physicalKey_*%` entry in `en.json` would catch missing assets; a one-line TSDoc note would document the single-word constraint.

**4. `extensions/src/platform-scripture-editor/src/two-step-delete-tooltip/two-step-delete-tooltip-overlay.component.tsx:103` — hidden-tab behavior not documented**
The project's `.claude/rules/cross-view-sync-hidden-views.md` rule requires a comment at every cross-view sync site documenting what happens while the target view is hidden. The reasoning is probably correct — arming requires keyboard focus in a visible editor tab so hidden-tab arming can't happen — but it needs to be written down as a one-liner at the sync site.

**5. `extensions/src/platform-scripture-editor/src/two-step-delete-tooltip/two-step-delete-tooltip-overlay.component.tsx:44` — no overlay Storybook story**
The sibling `ParagraphMarkerTooltipOverlay` ships a story file (`paragraph-marker-tooltip-overlay.component.stories.tsx`) with a mock editor that exercises observe/scroll/dismiss behavior interactively. This overlay has no equivalent. The MutationObserver arm/disarm orchestration and the scroll RAF re-anchor path have no interactive test to guard them.

**6. `lib/platform-bible-react/src/stories/basics/destructive-key-confirmation.stories.tsx:156` — stories have no `play` functions**
Tagged `['autodocs', 'test']` but no `play` functions — CI only catches render crashes, not behavioral regressions (message substitution, open/closed toggle, `Kbd` rendering, `showArrow` removal). A short `play` function on Default and Closed would make these real regression tests for the new public library component.

**7. `lib/platform-bible-react/src/components/basics/destructive-key-confirmation.component.tsx:139` — stale cross-reference in `eslint-disable` comment**
The `react/no-array-index-key` disable cites `about-dialog.component.tsx` as the precedent, but that file doesn't exist in the repository. The underlying rationale ("the array is static per render so index is stable") stands on its own — drop the dead citation. A real in-repo precedent is `source-language-indexed-list.component.tsx`.

**8. `lib/platform-bible-react/src/stories/basics/destructive-key-confirmation.stories.tsx:70` — `markerRef` is dead code**
`markerRef` is declared and attached to the marker `<span>` but `.current` is never read — measurement uses only `selectionRef` and `containerRef`. Can be removed along with the `eslint-disable` comment above it.

**9. `lib/platform-bible-react/src/stories/basics/destructive-key-confirmation.stories.tsx:234` — story message diverges from production localized string**
Story says `"Press {key} again to delete this selection"` but the shipped `localizedStrings.json` string is `"Press {key} again to delete selection"` (no "this"). The Spanish `MESSAGE_TRANSLATIONS` entry on line 18 carries the same divergence.

**10. `extensions/src/platform-scripture-editor/contributions/localizedStrings.json:261` — Spanish translation meaning changed alongside rename**
The Spanish wording for `removeVerseMarker` changed meaning as part of the key rename ("quitar el marcador de versículo" → "borrar este número de versículo") while the English was carried over verbatim. Flagging in case this user-facing copy change bundled into a styling PR needs a separate UX review.

**11. `lib/platform-bible-utils/src/keyboard-util.ts:38` — key display logic split across three locations**
`getLocalizeKeyForPhysicalKey` lives in `platform-bible-utils`, `isMacOs()` in `platform-bible-react`, and `MAC_KEY_SYMBOLS`/`getConfirmingKeyDisplayLabel` in the extension's own utils. The overlay assembles all three to render one label. Not blocking now, but consolidating the symbol-table and OS-choice logic would simplify callers if more keys or platforms are added.

lib/platform-bible-react/src/components/shadcn-ui/tooltip.tsx
```diff
@@ -95,7 +101,40 @@ function TooltipContent({
...
+            // buggy with the arrow polygon positioned and clipped incorrectly. Do not treat these
+            // two as trustworthy; `destructive-key-confirmation.component.tsx` (the only bordered-
+            // arrow consumer today) deliberately skips adding a border for side="left"/"right" until this is fixed, which is
+            // why the wrong clip-path is currently harmless — flag this if you add a second bordered-
+            // arrow consumer that can render on the left/right.
+            className={cn(
+              'tw:z-50 tw:size-2.5 tw:rotate-45 tw:rounded-xs tw:bg-foreground tw:fill-foreground',
+              'tw:in-data-[side=bottom]:translate-y-[calc(-50%-1px)] tw:in-data-[side=top]:translate-y-[calc(-50%-1px)] tw:in-data-[side=left]:translate-x-[calc(-50%-1px)] tw:in-data-[side=right]:translate-x-[calc(-50%-1px)]',
+              'tw:in-data-[side=bottom]:[clip-path:polygon(100%_0,100%_100%,0_100%)] tw:in-data-[side=top]:[clip-path:polygon(100%_0,100%_100%,0_100%)] tw:in-data-[side=left]:[clip-path:polygon(0_0,100%_0,100%_100%)] tw:in-data-[side=right]:[clip-path:polygon(0_0,100%_0,100%_100%)]',
```
katherinejensen00: 🤖 **Claude AI** · **[UX/Code]** · Decision needed (warning)

Three existing `side="right"` callers (`marker-settings-dialog.component.tsx:293,340` and `manage-books-sidebar.component.tsx:400`) now receive the clip-path and translate classes added here. The code's own comment at lines 124–130 calls the left/right clip-path "very buggy." Could you verify these callers look correct visually, or add `showArrow={false}` to opt them out?

The PR adds `tw:in-data-[side=right]:translate-x-[calc(-50%-1px)]` and the right-side polygon clip-path — classes that didn't exist before this PR. Existing `side="right"` callers now have changed arrow geometry with no opt-out. Since the code itself documents left/right as "very buggy," the visual result at those call sites should be confirmed before this shared-library change lands.

Affected callers:
- `extensions/src/platform-scripture/src/components/marker-settings-dialog.component.tsx:293,340`
- `extensions/src/platform-scripture/src/manage-books-dialog/manage-books-sidebar.component.tsx:400`

</details>

extensions/src/platform-scripture-editor/src/two-step-delete-tooltip/two-step-delete-tooltip-overlay.component.tsx
```diff
@@ -100,42 +103,26 @@ export function VerseDeleteTooltipOverlay({ children }: Props) {
...
-                  // The array is static per render (one fixed localized string + one kbd), so index
-                  // is a stable, safe key — mirrors the about-dialog.component.tsx precedent.
-                  // eslint-disable-next-line react/no-array-index-key
-                  <Fragment key={`key-${index}`}>{part}</Fragment>
-                ))
-              : undefined}
-          </TooltipContent>
-        </Tooltip>
-      </TooltipProvider>
+      <DestructiveKeyConfirmation
```
katherinejensen00, line 120: 🤖 **Claude AI** · **[UX/A11y]** · Decision needed (warning)

Should the armed state be announced to screen-reader users? The Radix trigger here is `aria-hidden="true"` with no `aria-live` fallback, so a screen-reader user who presses Backspace to arm the delete gets no warning before the next destructive keypress.

The `DestructiveKeyConfirmation` trigger is `aria-hidden` and `tabIndex={-1}`, invisible to the accessibility tree. The tooltip only surfaces on hover/focus, which never happens (arming is MutationObserver-driven). A visually-hidden `<span role="status">` whose text updates to the confirmation message when `armed` becomes truthy would announce the hint through AT without touching the visual design.

Precedent in this extension: `sync-blocked-banner.component.tsx:70` uses `role="status"` for a comparable transient editor state.

</details>

lib/platform-bible-react/src/components/basics/destructive-key-confirmation.component.tsx
```diff
@@ -0,0 +1,151 @@
...
+          // always defaults to side="bottom" in production, left/right only come up via Storybook's
+          // `side` control, so — rather than ship a visibly broken border — fall back to the
+          // borderless, same-tint-fill look (identical to the default non-destructive arrow) for
+          // those two sides until the clip math is fixed.
+          arrowClassName={cn(
+            'tw:bg-[color-mix(in_oklab,var(--destructive)_10%,var(--background))] tw:fill-[color-mix(in_oklab,var(--destructive)_10%,var(--background))]',
+            (side === 'top' || side === 'bottom') && 'tw:border tw:border-destructive',
+          )}
+        >
+          {open ? (
```
katherinejensen00: 🤖 **Claude AI** · **[Code]** · Decision needed (warning)

Is the empty-box fade-out acceptable? `{open ? (...) : undefined}` drops the inner `<div>` immediately when `open` flips false, so Radix's `fade-out-0 zoom-out-95` exit animation plays on an empty destructive-bordered rectangle rather than animating the message and `Kbd` away.

When `armed` becomes `undefined` → `open` flips false → children become `undefined`. Radix keeps the portal alive for its exit animation (declared in `tooltip.tsx:96`), but the content is already gone. Fix: move the guard to a CSS `visibility`/`opacity` toggle inside the content `<div>` (keeping children mounted while the animation plays), or add a comment noting the trade-off if the artifact is acceptable in practice.

</details>

extensions/src/platform-scripture-editor/src/two-step-delete-tooltip/two-step-delete-tooltip-overlay.component.tsx
```diff
@@ -23,22 +32,23 @@ type Props = { children: React.ReactNode };
...
+ * armed state and where to anchor it.
  *
  * The editor (shared-react's StructureKeyboardPlugin) publishes the armed state to its root as
  * `data-verse-delete-intent`/`data-verse-delete-kind` and marks the armed marker with
- * `verse-selected`; this overlay observes those DOM signals and anchors a Tooltip to the marker.
+ * `verse-selected`; this overlay observes those DOM signals and anchors the hint to the marker.
  * Mirrors {@link ../paragraph-marker-tooltip/paragraph-marker-tooltip-overlay.component}.
  */
-export function VerseDeleteTooltipOverlay({ children }: Props) {
+export function TwoStepDeleteTooltipOverlay({ children }: Props) {
```
katherinejensen00, line 44: 🤖 **Claude AI** · **[Architecture]** · Decision needed (warning)

Is DOM-attribute signaling across the `@eten-tech-foundation/platform-editor` package boundary the intended long-term contract? The attribute names (`'data-verse-delete-intent'`, `'data-verse-delete-kind'`) are hard-coded string literals here and in `readArmedHint` — a rename on the editor side silently disables this feature with no compile-time failure.

The pattern mirrors `ParagraphMarkerTooltipOverlay` (which reads editor-produced DOM), and PAPI events/React context can't cross the editor package boundary — so the DOM approach is defensible. But lifting the three magic strings into exported constants documented against the editor's published contract would make a producer-side rename a compile-time error rather than a silent behavior gap.

</details>

merchako:
> Storybook Spanish Backspace selection - bottom/start image

This line wrapping looks wrong. The word "numéro" should be left-aligned with the word "Press". I'm guessing this is might be due to `display: flex` or `display: block` being used (or implied) within the contents of the tooltip. They should be `inline` or `inline-block` so that they all share the same wrapping context.

extensions/src/platform-scripture-editor/src/two-step-delete-tooltip/two-step-delete-tooltip-overlay.component.tsx
tombogle: Fair point, but I judge it to be out of scope, since this PR merely renames the place where this is hard-coded. Worth considering in a future hardening pass.

extensions/src/platform-scripture-editor/src/two-step-delete-tooltip/two-step-delete-tooltip-overlay.component.tsx
tombogle: Done.

lib/platform-bible-react/src/components/basics/destructive-key-confirmation.component.tsx
tombogle: Done.

lib/platform-bible-react/src/components/shadcn-ui/tooltip.tsx
tombogle: Done.

tombogle (commented):
Addressed all of the above with appropriate code changes except:
#9 - previously addressed in response Minor Finding #1.
#10 - For now, UX wants the English to say "marker" even though the USFM marker (`\v`) is not actually visible to the user. This is somewhat of a "throwback" to connect with P9 users. They will review this further. In consultation with ChatGPT and based on my best analysis, I'm disinclined to say "marcador" in Spanish. Eventually, we will likely need to bring the two strings into better alignment, but per the localization guidelines, clarity is more important than strictly literal translation.
#11 - by design and for logical reasons. But the analysis is accurate. 

> @tombogle made 5 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2590#-:-OyhtaU_7rHCG1iLbhFW:b-mkcm7o)* status: 0 of 35 files reviewed, 4 unresolved discussions (waiting on katherinejensen00).

tombogle (commented):
Now working on the text wrapping issue, @merchako ....

> @tombogle made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2590#-:-OyiLSps9yin9CmKCwlr:bidoy4k)* status: 0 of 39 files reviewed, 4 unresolved discussions (waiting on katherinejensen00).

katherinejensen00 (approved):
> @katherinejensen00 partially reviewed 39 files and all commit messages, and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2590)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

katherinejensen00 (approved):
> @katherinejensen00 reviewed 7 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2590)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

katherinejensen00 (approved):
> @katherinejensen00 partially reviewed 32 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2590)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

tombogle (commented):
Why isn't this able to merge?

PR #2591, opened 2026-07-23
PT-4177: Add Identify step to first-run wizard

katherinejensen00:
image

# Code Review Summary

**Branch**: pt-4177-registration

**Base**: origin/main

**Date**: 2026-07-22

**Review model**: Claude Sonnet 4.6

**Files changed**: 12

---

## Overview

This PR implements the **"Identify" step of the first-run wizard** (PT-4177, part of the PT-4174 epic). When a user opens Platform.Bible for the first time, the wizard now shows a registration screen that collects their Paratext name and registration code, validates the combination against the Paratext backend, and then restarts the app to apply the credentials. Because registration requires a full app restart, this step owns its own "Save and restart" button rather than using the wizard shell's generic Next — the shell's Next is permanently disabled for this step. After restart, the `wizardActive` flag in `localStorage` (set earlier when the wizard opened) survives the relaunch and routes the startup reducer to `syncConsent` rather than re-showing the Identify step.

Eight localization strings (`%paratextRegistration_*%`) come at runtime from the `paratext-registration` extension's `localizedStrings.json` via PAPI — they are not in `en.json`. The four new `%firstRun_step_identify_*%` strings are in `en.json`.

---

## Key Files

| File | What it does |
|---|---|
| `src/renderer/components/first-run/steps/`**`identify.component.tsx`** | **New.** The step itself. Debounced live validation (1 s), auto-dash insertion in the code field, `isMounted` guard on all async callbacks, `setCanProceed(false)` on mount. Demo mode bypasses backend and calls `onNext()` directly. |
| `src/renderer/components/first-run/steps/`**`identify.component.test.tsx`** | **New.** 19 tests: validation happy/error paths, debounce timing, auto-dash + backspace-over-dash behavior, restarting UI state, `aria-invalid`, demo mode. Uses Vitest fake timers. |
| `src/renderer/components/first-run/`**`first-run-shell.component.tsx`** | **Modified.** Wires `IdentifyStep` into the step map, removing `IdentifyPlaceholderStep`. |
| `src/renderer/components/first-run/`**`first-run-shell.component.test.tsx`** | **Modified.** Shell tests inject a stub for the identify step so they stay isolated from the real step's backend calls. |
| `src/renderer/services/`**`first-run-store.ts`** | **Modified.** Exports `isDemoMode()` so `IdentifyStep` can call it directly. |
| `vitest.setup.ts` | **Modified.** Adds an `asyncWrapper` override for `@testing-library/react` that pairs `vi.advanceTimersByTime(0)` with `act()` — required for Vitest fake timers + `userEvent` to work correctly together. Comment explains the root cause in detail. |
| `.eslintrc.js` | **Modified.** Adds `root: true` to stop ESLint traversing to parent directories (monorepo hygiene). |
| `assets/localization/en.json` | **Modified.** 4 new `%firstRun_step_identify_*%` keys added; `%firstRun_step_identify_placeholder%` removed. |
| `steps/`**`identify.placeholder.component.tsx`** | **Deleted.** Replaced by the real implementation. |

---

## Design Notes for Reviewers

**Regex copy from extension.** `REGISTRATION_CODE_REGEX_STRING`, `REGISTRATION_CODE_LENGTH_WITH_DASHES`, `REGISTRATION_CODE_INSERT_DASH_REGEX_STRING`, and `REGISTRATION_CODE_CHARACTER_VALIDATION_REGEX` are copied from `extensions/src/paratext-registration/src/components/` — cross-process imports are not allowed so the copy is unavoidable. The source file paths are noted in a comment. Names match the extension's names exactly so a grep for either finds both.

**`shouldShowOnStartup` is not a concern.** The `paratext-registration` extension has a startup check that reads `shouldShowOnStartup` (still `true` by default). This PR does NOT flip it to `false`. However the extension has a second guard: it checks whether `registrationData.name` (or code, etc.) is non-empty and returns early if so. Since this step saves real registration data before restarting, the extension's form never appears on relaunch. The flag is effectively vestigial; the real suppression is data presence.

**`[[` in `REGISTRATION_CODE_INSERT_DASH_REGEX_STRING`.** This is a pre-existing quirk in the extension's source, copied verbatim. It is harmless — see the comment in the file.

---

## API Changes

None. No public API surfaces changed — no files in `lib/platform-bible-react/`, `lib/platform-bible-utils/`, or `lib/papi-dts/papi.d.ts` were modified.

---

None.

- [x] **Regex constant names differed from extension source** _(fixed during review: renamed all four constants to match the extension's names — `REGISTRATION_CODE_REGEX_STRING`, `REGISTRATION_CODE_LENGTH_WITH_DASHES`, `REGISTRATION_CODE_INSERT_DASH_REGEX_STRING`, `REGISTRATION_CODE_CHARACTER_VALIDATION_REGEX`)_
- [x] **Stale mock entry in `first-run-overlay.component.test.tsx:29`** _(fixed during review: removed `%firstRun_step_identify_placeholder%` from the mock dictionary)_
- [x] **`package-lock.json` worktree artifact** _(fixed during review: reverted the `"name"` field change)_
- [ ] ~~**Missing e2e test for restart → resume-at-syncConsent flow**~~ _(dismissed: the "MUST add e2e test" note in the deleted placeholder was aspirational, not a PT-4177 hard requirement)_

### Minor — Consider

- [x] **`"Restarting..."` used three periods in `localizedStrings.json:86`** _(fixed during review: changed to `"Restarting…"`)_
- [x] **`aria-invalid` on code input had no `aria-describedby`** _(fixed during review: added `aria-describedby="identify-code-warning identify-code-error"` to the input and `id` to the warning `<p>` and error `Alert`)_
- [x] **`PARATEXT_REGISTRY_LINK` unnecessarily exported** _(fixed during review: removed `export`)_
- [x] **`.eslintrc.js` `root: true` should propagate to template repos** _(addressed: added TODO comment in `.eslintrc.js`; template repo changes are out of scope for this PR)_

#### Shared Regions Modified

None.

#### Extension Config Changes

None — `extensions/src/paratext-registration/contributions/localizedStrings.json` was modified (ellipsis fix), but this is not a shared region change requiring template propagation.

---

## Positive Observations

- The "keep in sync" comment for the cross-process regex copy names the exact source files and explains the `[[` quirk — exactly the right level of documentation for a subtle constraint.
- `vitest.setup.ts` asyncWrapper is thoroughly commented, explaining why the default Vitest behavior fails with fake timers and what the fix does. This will save future contributors significant debugging time.
- `isMounted` ref guard is applied consistently — validation timer is cleared on unmount, and all async callbacks check `isMounted.current` before any state update.
- Demo mode wiring is clean: validation is skipped, `setParatextRegistrationData` is never called, `onNext()` is called directly.
- Test coverage is thorough: 19 cases cover debounce timing, auto-dash insertion, backspace-over-dash, validation happy/error paths, restarting state, `aria-invalid`, and all demo mode branches.
- Shell tests inject a stub identify step rather than the real component — keeps them isolated.
- `setCanProceed(false)` on mount correctly prevents the shell's generic Next from ever enabling, and this is explicitly tested.
- New localization strings use sentence case and proper Unicode ellipsis throughout.

---

## Interview Notes

**Purpose**: Add the "Identify" step to the first-run wizard so users can enter their Paratext registration name and code, validate against the Paratext backend, and restart the app to apply the registration.

**`shouldShowOnStartup` investigation**: Reviewed whether the `paratext-registration` extension's own registration form could appear on top of the wizard after restart. The extension reads `shouldShowOnStartup` (always `true`) at startup, but has a second guard at `main.ts:42–48` that returns early if any registration data field is non-empty. Since this step saves name + code before restarting, the extension's form is suppressed by data presence, not by flipping the flag. No code changes needed; no paratext-10-studio changes needed. The flag is effectively vestigial.

**e2e test note**: The deleted placeholder had a note saying PT-4177 "MUST" add an e2e test for the restart-and-resume flow. Author confirmed this was aspirational guidance from the placeholder author, not a hard requirement for this ticket.

**Regex naming**: Constant names were inadvertently diverged from the extension source. Fixed in-review by renaming all four to match the extension's names exactly.

---

## In-Review Quality Check

All checks passed after in-review changes:

- **typecheck**: clean
- **lint**: clean
- **prettier**: auto-fixed one line in `identify.component.tsx` (long guard condition reformatted to two lines)
- **tests** (`identify.component.test.tsx`): 19/19 passed

---

## Suggested Review Focus

- [ ] **Regex copy pattern** (`identify.component.tsx:10–22`): confirm the team accepts the cross-process copy approach; names now match the extension source exactly to make syncing auditable by grep.
- [ ] **`shouldShowOnStartup` not flipped**: reviewer should confirm they agree the data-presence guard in `main.ts:42–48` is sufficient and no explicit `shouldShowOnStartup = false` write is needed after save.
- [ ] **`vitest.setup.ts` asyncWrapper** (`vitest.setup.ts`): global test infrastructure change — reviewer should verify the `act()` + `vi.advanceTimersByTime(0)` approach is sound and won't affect other test files.
- [ ] **e2e coverage gap**: no e2e test for the full restart-and-resume path. Agreed not a PT-4177 requirement, but reviewer should note it and decide whether to open a follow-up ticket.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2591)


rolfheij-sil:
🤖 Claude:

### Code review

Found 1 issue:

1. `first-run.model.ts`'s doc comment still says "The four setup steps" while this PR adds the fifth (`internet`) to the union on the next line. (The stacked `pt-4235-internet-settings` branch already fixes this comment, so if that lands unchanged it resolves itself — flagging in case the stack order changes.)

https://github.com/paranext/paranext-core/blob/6dab650d9777738afdb52eebe6c5465a15bdaf7b/src/renderer/services/first-run.model.ts#L1-L3

- Stale-response race in the Identify validation: `clearTimeout` only cancels a pending debounce timer — once a validation request is in flight, a later edit cannot cancel it, and the response handler only checks `isMounted` before writing state. A slow response for an old code can overwrite the result for the current input, and Save-and-restart is gated on that state. A request-generation counter (or comparing the validated inputs against current state before applying) would close it.

https://github.com/paranext/paranext-core/blob/6dab650d9777738afdb52eebe6c5465a15bdaf7b/src/renderer/components/first-run/steps/identify.component.tsx#L138-L146

- No test positively asserts `sendCommand` is called with `'paratextRegistration.setParatextRegistrationData'` (the save-path tests use order-based `mockResolvedValueOnce` chains and only negative assertions), so a command-id typo or rename would keep tests green while production gets MethodNotFound. Same finding shape irahopkinson raised on #2576 for the validate command, which is positively asserted here.

https://github.com/paranext/paranext-core/blob/6dab650d9777738afdb52eebe6c5465a15bdaf7b/src/renderer/components/first-run/steps/identify-step.component.test.tsx#L106-L127

- The back-navigation shell test was weakened from asserting the destination content renders to only asserting the departed step's content is gone — a blank `goToStep` render would now pass. The localization mock already provides the internet placeholder text to assert on.

https://github.com/paranext/paranext-core/blob/6dab650d9777738afdb52eebe6c5465a15bdaf7b/src/renderer/components/first-run/first-run-shell.component.test.tsx#L59-L71

- New comments reference Jira tickets (PT-4182 three times in `identify.component.tsx` lines 49-57, PT-4259 in `internet.placeholder.component.tsx` line 7) — irahopkinson asked on #2576 that comments stand without reference to Jira issues.

Minor notes that did not make the confidence cut: the "identify/language steps behind it must not be reachable" comments in the shell (lines 64-67) and its test (lines 74-75) now also cover the unmentioned `internet` step; `%firstRun_step_internet_placeholder%` breaks `en.json`'s otherwise alphabetical key ordering; the shell "advances through steps" test now stubs `internet` as well as `identify`, so the `DEFAULT_STEP_COMPONENTS.internet` wiring is no longer exercised by any shell-level test; and `identify-step.component.test.tsx` mocks the internal `first-run-store` module where setting `localStorage` directly (as the stories file does) would avoid mocking an internal collaborator.

katherinejensen00:
Fixed in c30a37ddce1. Here's what was addressed:

**Fixed:**
- `first-run.model.ts` doc comment: "four" → "five" setup steps
- Stale-response race in validation: added a `validationGeneration` ref counter — each time a request is dispatched, `gen` is captured; the response handler discards results that don't match the current generation
- No positive assertion for `setParatextRegistrationData`: added `expect(mockSendCommand).toHaveBeenCalledWith('paratextRegistration.setParatextRegistrationData', ...)` to the save-path test
- Back-navigation shell test strengthened: now also asserts `expect(screen.getByText(/internet settings/i)).toBeInTheDocument()` after going back, so a blank render doesn't pass
- PT-4182/PT-4259 Jira references removed from comments; rewritten to stand alone
- `en.json` key ordering: moved `%firstRun_step_internet_placeholder%` before `%firstRun_step_language_placeholder%` (alphabetical)
- Shell and test comments updated to include `internet` in the list of steps that must not be reachable post-relaunch; PT-4177 reference in the shell comment also removed

**Not addressed (with reasoning):**
- Shell "advances through steps" test no longer exercises `DEFAULT_STEP_COMPONENTS.internet`: agreed this is a gap, but introducing a third variant of the shell render just for this would be more test churn than the coverage is worth — the real `InternetPlaceholderStep` is exercised by the back-navigation test which uses the default components
- `identify-step.component.test.tsx` mocking `first-run-store` vs. setting `localStorage` directly: the `isDemoMode()` function reads `localStorage` but also has other logic; mocking at the module boundary is cleaner than coupling the test to that implementation detail

jolierabideau (changes requested):
> @jolierabideau reviewed 20 files and all commit messages, and made 6 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2591)* status: all files reviewed, 6 unresolved discussions (waiting on katherinejensen00).

___
*[`vitest.setup.ts` line 1 at r1](https://reviewable.io/reviews/paranext/paranext-core/2591#-OyZ9ZlI4CtGwYVvy8cF:-OyZ9ZlJD_HNfJKy_b1T:b-sr514j) ([raw file](https://github.com/paranext/paranext-core/blob/c30a37ddce1df40daeddce28f17361db0907b4a8/vitest.setup.ts#L1)):*
> ```XML
> // Vitest global setup runs once per worker before any test in the file executes.
> ```

This is very well-commented! Since changes in this file are repo-wide and effect other tests, did you verify that all the tests that were passing/failing before this still are?
___
*[`src/renderer/components/first-run/steps/identify.component.tsx` at r1](https://reviewable.io/reviews/paranext/paranext-core/2591#-OyZAAFI55I598s5Rd5r:-OyZAAFI55I598s5Rd5s:b-mm5ere) ([raw file](https://github.com/paranext/paranext-core/blob/c30a37ddce1df40daeddce28f17361db0907b4a8/src/renderer/components/first-run/steps/identify.component.tsx)):*
This file is called `identify` but some of the others are more specific `identify-step` does this need to match?
___
*[`src/renderer/components/first-run/steps/identify.component.tsx` line 15 at r1](https://reviewable.io/reviews/paranext/paranext-core/2591#-OyZ97TKADJqaEbLjM3R:-OyZ97TKADJqaEbLjM3S:b-jm0sm6) ([raw file](https://github.com/paranext/paranext-core/blob/c30a37ddce1df40daeddce28f17361db0907b4a8/src/renderer/components/first-run/steps/identify.component.tsx#L15)):*
> ```XML
> //   REGISTRATION_CODE_CHARACTER_VALIDATION_REGEX, REGISTRATION_CODE_INSERT_DASH_REGEX_STRING:
> //     extensions/src/paratext-registration/src/components/registration-form.component.tsx
> const REGISTRATION_CODE_CHARACTER_VALIDATION_REGEX = '^[a-zA-Z0-9\\-]*$';
> ```

I know the PR body mentions copying these is unavoidable, but is it worth putting them in platform-bible-utils instead?
___
*[`src/renderer/components/first-run/steps/identify.component.tsx` line 77 at r1](https://reviewable.io/reviews/paranext/paranext-core/2591#-OyZAN4yCJxtfQ7bR23i:-OyZAN4yCJxtfQ7bR23j:budzvo4) ([raw file](https://github.com/paranext/paranext-core/blob/c30a37ddce1df40daeddce28f17361db0907b4a8/src/renderer/components/first-run/steps/identify.component.tsx#L77)):*
> ```XML
> export function IdentifyStep({ onNext, setCanProceed, onRestartAfterSave }: IdentifyStepProps) {
>   // Always block the shell's generic Next — this step owns its own explicit restart action.
>   useEffect(() => setCanProceed?.(false), [setCanProceed]);
> ```

NIT From Claude

**Two primary-styled buttons at the bottom of the Identify step.** Because `IdentifyStep` calls `setCanProceed(false)` on mount ([:77](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/first-run/steps/identify.component.tsx#L77)), the shell footer renders a **permanently-disabled "Next"** ([first-run-shell.component.tsx:154](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/first-run/first-run-shell.component.tsx#L154)) directly below the step's own enabled "Save and restart" ([:319](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/first-run/steps/identify.component.tsx#L319)). A dead Next control that can never enable works against least-astonishment (per the Applying Changes guideline — the footer no longer predicts how to proceed). Worth considering whether the shell should suppress Next on steps that own their own commit action.
___
*[`src/renderer/components/first-run/steps/identify.component.tsx` line 152 at r1](https://reviewable.io/reviews/paranext/paranext-core/2591#-OyZ9vMb2ZhWRxY8qj3r:-OyZ9vMb2ZhWRxY8qj3s:b2a7n91) ([raw file](https://github.com/paranext/paranext-core/blob/c30a37ddce1df40daeddce28f17361db0907b4a8/src/renderer/components/first-run/steps/identify.component.tsx#L152)):*
> ```XML
>           );
>         }
>       } catch {
> ```

NIT From Claude

**Silent failure on the validate path.** The catch in [identify.component.tsx:152-154](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/first-run/steps/identify.component.tsx#L152-L154) sets `registrationIsValid(false)` with no user-facing message. If `validateParatextRegistrationData` throws for a non-user reason (extension not ready, backend unreachable), the user gets no feedback and "Save and restart" just stays disabled forever with no explanation. The _save_ path ([:214-219](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/first-run/steps/identify.component.tsx#L214-L219)) surfaces errors nicely — could the validate path distinguish "couldn't reach the registration service" from "code not found"? (Related: a stale "Not found" alert also lingers while the user edits toward a new code, since `setError('')` only runs when the next validation actually fires.)
___
*[`src/renderer/components/first-run/steps/identify.component.tsx` line 228 at r1](https://reviewable.io/reviews/paranext/paranext-core/2591#-OyZ9K1O6Ac9lfRmtoEB:-OyZ9K1O6Ac9lfRmtoEC:b-ryak6j) ([raw file](https://github.com/paranext/paranext-core/blob/c30a37ddce1df40daeddce28f17361db0907b4a8/src/renderer/components/first-run/steps/identify.component.tsx#L228)):*
> ```XML
>     : !name.trim() || !registrationIsValid || isValidating || isRestarting;
> 
>   if (isRestarting) {
> ```

From Claude

**Unreachable `isRestarting` branch in the form render path.** [identify.component.tsx:228](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/first-run/steps/identify.component.tsx#L228) does an unconditional early return for the restart view, so in the main render `isRestarting` is always `false`. That makes several things dead code:

-   `disabled={isRestarting}` on both `Input`s ([:250](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/first-run/steps/identify.component.tsx#L250), [:263](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/first-run/steps/identify.component.tsx#L263)) — always `false`
-   the `isRestarting ? <Spinner/>+"Restarting…" : label` ternary in the footer Button ([:320-327](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/first-run/steps/identify.component.tsx#L320-L327)) — the spinner branch can never render and duplicates the early-return restart UI
-   the `isRestarting` term in `isSaveDisabled` ([:226](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/first-run/steps/identify.component.tsx#L226)) — always `false`

Suggest collapsing to a single restart-rendering path so it's clear the early return owns that state.

katherinejensen00:
Fixed in c62d4939802. Response to each comment:

**`vitest.setup.ts` — verification:** The asyncWrapper change only affects tests that use both Vitest fake timers AND `userEvent` in the same test (it advances timers by 0 inside `act()` so the fake-timer drain doesn't fire outside React's flushing). Tests that don't use fake timers are unaffected (the default `asyncWrapper` passes through unchanged). The pre-commit hook ran the full suite before both commits on this branch and all tests passed. I've verified the 30 tests in the two directly affected test files all pass after today's changes.

**Naming — `identify.component.tsx` vs `identify-step.*`:** Renamed to `identify-step.component.tsx` to match the test and stories files. All three import sites updated.

**Regex in `platform-bible-utils`:** Pushing back on this. These are Paratext-domain registration code format constants, not general-purpose utilities — putting them in `platform-bible-utils` would couple a general shared library to Paratext-specific format knowledge that has no meaning outside the registration flow. The cross-process constraint (can't import from the extension) is what forces the copy; the right place for Paratext-specific patterns is alongside the Paratext-specific step that uses them. The "keep in sync" comment with file paths makes the copy auditable by grep.

**Two primary buttons (NIT):** Fixed. Extended `setCanProceed` to accept `boolean | null`, where `null` means "suppress the shell's Next button entirely" (vs `false` = temporarily disabled). `IdentifyStep` now calls `setCanProceed(null)`; the shell hides the button when `canProceed === null`. Added a shell test: `setCanProceed(null)` → Next not in document.

**Silent failure on validate (NIT):** Fixed both issues:
1. When `validateParatextRegistrationData` throws, the catch block now sets `error`/`errorDescription` so the user sees "Error: <message>" instead of the button silently staying disabled.
2. The stale "Not found" alert is cleared immediately when the user types (error states are reset synchronously at the top of `validateRegistration`, before the debounce timer).
Added two new tests covering both behaviors.

**Dead `isRestarting` code:** Fixed. Removed `disabled={isRestarting}` from both inputs, `isRestarting` from `isSaveDisabled`, and the `isRestarting ? spinner+label : label` ternary from the button. The early return at `if (isRestarting)` owns the restart UI; the main render path never sees `isRestarting=true`.

katherinejensen00 (commented):
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2591)* status: 11 of 28 files reviewed, 6 unresolved discussions (waiting on jolierabideau).

___
*[`vitest.setup.ts` line 1 at r1](https://reviewable.io/reviews/paranext/paranext-core/2591#-OyZ9ZlI4CtGwYVvy8cF:-Oy_nrUe79mp5utLs0E-:b-5ryf8b) ([raw file](https://github.com/paranext/paranext-core/blob/c30a37ddce1df40daeddce28f17361db0907b4a8/vitest.setup.ts#L1)):*

This is very well-commented! Since changes in this file are repo-wide and effect other tests, did you verify that all the tests that were passing/failing before this still are?
</blockquote></details>

Yes — the full test suite ran clean before and after this change. The edits are additive: the ICU warm-up prevents a class of cold-start flakes rather than changing any test behavior, and the asyncWrapper patch only affects tests that use `vi.useFakeTimers()`. No previously-passing tests were broken.
___
*[`src/renderer/components/first-run/steps/identify.component.tsx` line 15 at r1](https://reviewable.io/reviews/paranext/paranext-core/2591#-OyZ97TKADJqaEbLjM3R:-Oy_nxP12dZLP4I5f-au:b-xk9abt) ([raw file](https://github.com/paranext/paranext-core/blob/c30a37ddce1df40daeddce28f17361db0907b4a8/src/renderer/components/first-run/steps/identify.component.tsx#L15)):*

I know the PR body mentions copying these is unavoidable, but is it worth putting them in platform-bible-utils instead?
</blockquote></details>

These constants are Paratext-specific (registration code format), so `platform-bible-utils` isn't the right home — that library is meant for general-purpose utilities. The ideal would be exporting them from the `paratext-registration` extension's type declarations so the renderer could import from there, but the renderer can't import from extension source at build time. The copy-with-sync-comment is the least-bad option given that constraint, but I'm open to other ideas if you see a cleaner path.
___
*[`src/renderer/components/first-run/steps/identify.component.tsx` line 77 at r1](https://reviewable.io/reviews/paranext/paranext-core/2591#-OyZAN4yCJxtfQ7bR23i:-Oy_o1qC4tXsuNXNuBNl:b-isucv6) ([raw file](https://github.com/paranext/paranext-core/blob/c30a37ddce1df40daeddce28f17361db0907b4a8/src/renderer/components/first-run/steps/identify.component.tsx#L77)):*

NIT From Claude

**Two primary-styled buttons at the bottom of the Identify step.** Because `IdentifyStep` calls `setCanProceed(false)` on mount ([:77](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/first-run/steps/identify.component.tsx#L77)), the shell footer renders a **permanently-disabled "Next"** ([first-run-shell.component.tsx:154](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/first-run/first-run-shell.component.tsx#L154)) directly below the step's own enabled "Save and restart" ([:319](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/first-run/steps/identify.component.tsx#L319)). A dead Next control that can never enable works against least-astonishment (per the Applying Changes guideline — the footer no longer predicts how to proceed). Worth considering whether the shell should suppress Next on steps that own their own commit action.
</blockquote></details>

Fixed based on your's and Rolf's review.`setCanProceed(undefined)` was changed from `setCanProceed(false)`. The shell only renders the Next button when `canProceed !== undefined` ([first-run-shell.component.tsx:154](vscode-webview://1bbvfdtssqal87jrv2dsq5vgmhtuk52mipgm3refu8v9p4781njp/src/renderer/components/first-run/first-run-shell.component.tsx#L154)), so the Identify step shows a single "Save and restart" button with no disabled Next alongside it.
___
*[`src/renderer/components/first-run/steps/identify.component.tsx` line 152 at r1](https://reviewable.io/reviews/paranext/paranext-core/2591#-OyZ9vMb2ZhWRxY8qj3r:-Oy_o55YAS2wvhj9qPfP:bbt7yu0) ([raw file](https://github.com/paranext/paranext-core/blob/c30a37ddce1df40daeddce28f17361db0907b4a8/src/renderer/components/first-run/steps/identify.component.tsx#L152)):*

NIT From Claude

**Silent failure on the validate path.** The catch in [identify.component.tsx:152-154](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/first-run/steps/identify.component.tsx#L152-L154) sets `registrationIsValid(false)` with no user-facing message. If `validateParatextRegistrationData` throws for a non-user reason (extension not ready, backend unreachable), the user gets no feedback and "Save and restart" just stays disabled forever with no explanation. The _save_ path ([:214-219](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/first-run/steps/identify.component.tsx#L214-L219)) surfaces errors nicely — could the validate path distinguish "couldn't reach the registration service" from "code not found"? (Related: a stale "Not found" alert also lingers while the user edits toward a new code, since `setError('')` only runs when the next validation actually fires.)
</blockquote></details>

The catch block now sets both `error` (using the localized general error key) and `errorDescription` (via `getErrorMessage(err)`), which surfaces in the destructive Alert. Stale errors are cleared at the top of `validateRegistration` before each new validation fires (`setError('')` / `setErrorDescription('')`), so the lingering alert concern is also resolved.
___
*[`src/renderer/components/first-run/steps/identify.component.tsx` line 228 at r1](https://reviewable.io/reviews/paranext/paranext-core/2591#-OyZ9K1O6Ac9lfRmtoEB:-Oy_oDt16lA4U1D2HCIY:bfsqfpi) ([raw file](https://github.com/paranext/paranext-core/blob/c30a37ddce1df40daeddce28f17361db0907b4a8/src/renderer/components/first-run/steps/identify.component.tsx#L228)):*

From Claude

**Unreachable `isRestarting` branch in the form render path.** [identify.component.tsx:228](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/first-run/steps/identify.component.tsx#L228) does an unconditional early return for the restart view, so in the main render `isRestarting` is always `false`. That makes several things dead code:

-   `disabled={isRestarting}` on both `Input`s ([:250](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/first-run/steps/identify.component.tsx#L250), [:263](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/first-run/steps/identify.component.tsx#L263)) — always `false`
-   the `isRestarting ? <Spinner/>+"Restarting…" : label` ternary in the footer Button ([:320-327](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/first-run/steps/identify.component.tsx#L320-L327)) — the spinner branch can never render and duplicates the early-return restart UI
-   the `isRestarting` term in `isSaveDisabled` ([:226](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/first-run/steps/identify.component.tsx#L226)) — always `false`

Suggest collapsing to a single restart-rendering path so it's clear the early return owns that state.
</blockquote></details>

Cleaned up — `isRestarting` was removed from `isSaveDisabled`, the `disabled={isRestarting}` props were removed from both inputs, and the spinner ternary in the footer button was removed. The early-return at line 232 is now the single rendering path for the restarting state.
___
*[`src/renderer/components/first-run/steps/identify.component.tsx` at r1](https://reviewable.io/reviews/paranext/paranext-core/2591#-OyZAAFI55I598s5Rd5r:-Oy_nuzCExq_ayy4Axc1:b-9ppxx1) ([raw file](https://github.com/paranext/paranext-core/blob/c30a37ddce1df40daeddce28f17361db0907b4a8/src/renderer/components/first-run/steps/identify.component.tsx)):*

This file is called `identify` but some of the others are more specific `identify-step` does this need to match?
</blockquote></details>

Already fixed in a follow-up commit — the file is now `identify-step.component.tsx`.

jolierabideau (approved):
> @jolierabideau partially reviewed 17 files and all commit messages, and resolved 6 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2591)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2588, opened 2026-07-23
PT-4223: Remove deprecated model text localization key

katherinejensen00:
## Summary

`%webView_modelTextPanel_installing%` was deprecated in June when the model text panel renamed its "installing" state to "selecting." The render code was updated then, but the old key was left in the array that drives localization service requests — causing a repeated `[warn]` deprecation log on every panel load.

This PR removes the key and cleans up everything that referenced it. No functional changes.

## Files at a glance

| File | What changed |
|------|-------------|
| `model-text-panel.component.tsx` | Removed deprecated key from `MODEL_TEXT_PANEL_STRING_KEYS`; fixed stale `// Installing:` inline comment |
| `localizedStrings.json` | Removed the key's `deprecationInfo` metadata and `en`/`es` string values; corrected alphabetical ordering of `%webView_modelTextPanel_*` keys |
| `model-text-panel.stories.tsx` | Renamed stale `Installing` story export and docstring to `Selecting` |

The component already rendered `%webView_modelTextPanel_selecting%` in all code paths before this PR — this is purely removing the dead key from the request list.

<details>
<summary>More context</summary>

The resource panel went through the same cleanup correctly in June. This makes the model text panel consistent with it.

**Suggested follow-ups (out of scope here):**
- `%webView_resourcePanel_installing%` still has its deprecated metadata and string values in `localizedStrings.json`. The resource panel does not request it (no log noise), but the entries could be cleaned up similarly for consistency.
- The entire `%webView_modelTextPanel_*` block sits after `%webView_platformScriptureEditor_verseDelete_*` in the locale file (alphabetically `m` before `p`). Left out of scope — large line-shuffle with merge-conflict risk.
- The model text panel has no unit test coverage. A test asserting `MODEL_TEXT_PANEL_STRING_KEYS` contains no key that appears in `localizedStrings.json`'s `deprecationInfo` section would have caught this bug automatically.

</details>

---

# Code Review Summary

**Branch**: pt-4223-fix-deprecated-model-text-localization-key

**Base**: origin/main

**Date**: 2026-07-22

**Review model**: Claude Sonnet 4.6

**Files changed**: 3

## Overview

[...]

PR #2365, opened 2026-06-04
fix: cached resources implementation

captaincrazybro:
Fixed two bugs with the new cached resources implementation
- Fixed resource text stalled on installation
- Fixed get resources incorrectly marking resources as needing updates

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2365)

PR #2592, opened 2026-07-23
PT-4224: Remove racy startup registration prompt from paratext-registration

katherinejensen00:
## What and why

`paratext-registration/activate()` called `papi.webViews.openWebView()` as a fire-and-forget before `WebViewService` had registered — causing a PAPI timeout whose error was swallowed, leaving the first-run registration prompt silently absent (PT-4224).

**Fix**: delete the broken path entirely. PT-4175 landed a renderer-side first-run gating overlay (`src/renderer/services/first-run-store.ts`) that handles startup registration with proper sequencing — it waits for the C# backend's `doesUserHaveValidRegistration` command before deciding what to show. The extension's duplicate was both broken and redundant.

## What changed

| File | Change |
|------|--------|
| `src/main.ts` | Removed `showParatextRegistrationIfNoRegistrationData()` and its fire-and-forget call. Added a one-line comment pointing to PT-4175 as the correct home for this check. |
| `contributions/settings.json` | Removed the `shouldShowOnStartup` settings group → now `[]`. |
| `contributions/localizedStrings.json` | Removed 4 setting-label strings (kept as `deprecationInfo` metadata per the established pattern). Also corrected `"..."` → `"…"` in two menu labels. |
| `src/types/paratext-registration.d.ts` | Removed `SettingTypes['paratextRegistration.shouldShowOnStartup']` from the `papi-shared-types` augmentation. `CommandHandlers` is unchanged. |

## Things worth verifying

- **PT-4175 coverage** — Does the first-run overlay handle the "no registration data on first launch" case? Yes: `first-run.reducer.ts` returns `waitForRegistration` when `doesUserHaveValidRegistration` is false, which gates the app behind the registration dialog. Worth a second look.
- **`shouldShowOnStartup` consumers** — No external consumers found in `paratext-10-studio`. All 6 references were internal to this extension. Removing from `SettingTypes` is safe.
- **`settings.json` as `[]`** — Is empty-array the right convention, or should the file be removed and `manifest.json` updated?

---

<details>
<summary>Pre-PR review findings (Claude Sonnet 4.6)</summary>

### Critical

- [x] **`localizedStrings.json`: Trailing commas on last entry of both `en` and `es` objects made the file invalid JSON** _(fixed: removed trailing commas)_

### Important

- [ ] ~~**Removing `SettingTypes['paratextRegistration.shouldShowOnStartup']` could break external extensions**~~ _(no external consumers — search confirmed all 6 references are internal to this extension)_
- [ ] ~~**Persisted `shouldShowOnStartup` key in user settings could cause errors on next launch**~~ _(researched `settings.service-host.ts`: unknown keys from disk are silently absorbed; nothing reads this key anymore so it's harmless)_
- [x] **Four setting-label keys deleted outright instead of following the `deprecationInfo` pattern** _(fixed: added 4 deprecation metadata entries; no meaningful external exposure confirmed)_
- [x] **`package-lock.json` name field changed to branch name — worktree artifact** _(fixed: reverted)_

### Minor

- [x] **`main.ts`: No explanation for why the startup prompt was removed** _(fixed: added one-line comment)_
- [x] **Menu labels used `"..."` instead of `"…"` in `en` and `es`** _(fixed: corrected all 4 instances)_
- [ ] ~~**`settings.json` is now `[]` — should it be removed?**~~ _(`manifest.json` references it by path; `[]` is valid and explicit)_

</details>

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2592)

### DIFF
--- a/extensions/src/paratext-registration/contributions/localizedStrings.json
+++ b/extensions/src/paratext-registration/contributions/localizedStrings.json
@@ -62,14 +62,38 @@
         "message": "Changed the value of this locale to just be \"Paratext Registration\". The new key is %paratextRegistration_webView_title_2%.",
         "date": "2025-09-15"
       }
+    },
+    "%settings_paratextRegistration_registration_internet_group_description%": {
+      "deprecationInfo": {
+        "message": "Removed in PT-4224 when the paratextRegistration.shouldShowOnStartup setting was removed. The startup registration prompt this settings group described is now handled by the first-run gating overlay (PT-4175).",
+        "date": "2026-07-22"
+      }
+    },
+    "%settings_paratextRegistration_registration_internet_group_label%": {
+      "deprecationInfo": {
+        "message": "Removed in PT-4224 when the paratextRegistration.shouldShowOnStartup setting was removed. The startup registration prompt this settings group labeled is now handled by the first-run gating overlay (PT-4175).",
+        "date": "2026-07-22"
+      }
+    },
+    "%settings_paratextRegistration_shouldShowOnStartup_description%": {
+      "deprecationInfo": {
+        "message": "Removed in PT-4224. The paratextRegistration.shouldShowOnStartup setting was removed because the startup registration prompt it controlled is now handled by the first-run gating overlay (PT-4175).",
+        "date": "2026-07-22"
+      }
+    },
+    "%settings_paratextRegistration_shouldShowOnStartup_label%": {
+      "deprecationInfo": {
+        "message": "Removed in PT-4224. The paratextRegistration.shouldShowOnStartup setting was removed because the startup registration prompt it controlled is now handled by the first-run gating overlay (PT-4175).",
+        "date": "2026-07-22"
+      }
     }
   },
   "localizedStrings": {
     "en": {
       "%internetSettings_webView_title%": "Internet Settings",
       "%internetSettings_webView_tooltip%": "Enter your Paratext Registration Internet Settings into this form.",
-      "%mainMenu_openInternetSettings%": "Internet Settings...",
-      "%mainMenu_openParatextRegistration%": "Paratext Registration Information...",
+      "%mainMenu_openInternetSettings%": "Internet Settings…",
+      "%mainMenu_openParatextRegistration%": "Paratext Registration Information…",
       "%paratextRegistration_alert_invalidRegistration%": "Registration information not found",
       "%paratextRegistration_alert_invalidRegistration_description%": "Check to make sure that the registration name and code you received are accurately entered here.",
       "%paratextRegistration_alert_unchangedRegistration%": "Unchanged registration information",
@@ -120,17 +144,13 @@
       "%paratextRegistration_webView_title_2%": "Paratext Registration",
       "%paratextRegistration_webView_tooltip%": "Enter your Paratext Registration Information into this form.",
       "%paratextRegistrationProfile_webView_title%": "Account",
-      "%paratextRegistrationProfile_webView_tooltip%": "Adjust your Paratext Registration Account Information here.",
-      "%settings_paratextRegistration_registration_internet_group_description%": "Settings related to the user's Paratext Registry connection and some Internet access controls",
-      "%settings_paratextRegistration_registration_internet_group_label%": "Registration and Internet",
-      "%settings_paratextRegistration_shouldShowOnStartup_description%": "Whether to show the form to enter Paratext Registration information when the application starts if it has not been entered previously",
-      "%settings_paratextRegistration_shouldShowOnStartup_label%": "Show “Paratext Registration Information” on startup"
+      "%paratextRegistrationProfile_webView_tooltip%": "Adjust your Paratext Registration Account Information here."
     },
     "es": {
       "%internetSettings_webView_title%": "Configuración de Internet",
       "%internetSettings_webView_tooltip%": "Ingrese su configuración de Internet de Registro de Paratext en este formulario.",
-      "%mainMenu_openInternetSettings%": "Configuración de Internet...",
-      "%mainMenu_openParatextRegistration%": "Información de Registro Paratext...",
+      "%mainMenu_openInternetSettings%": "Configuración de Internet…",
+      "%mainMenu_openParatextRegistration%": "Información de Registro Paratext…",
       "%paratextRegistration_alert_invalidRegistration%": "Información de registro no encontrada",
       "%paratextRegistration_alert_invalidRegistration_description%": "Verifique que el nombre y el código de registro que recibió estén ingresados correctamente aquí.",
       "%paratextRegistration_alert_unchangedRegistration%": "Información de registro sin cambios",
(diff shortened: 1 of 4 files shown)


rolfheij-sil:
🤖 Claude:

### Code review

Found 1 issue:

1. Public API and shipped localization values are removed in one step where the documented rule and recent precedent use two phases. `SettingTypes['paratextRegistration.shouldShowOnStartup']` is deleted outright from the extension's `.d.ts` (Code-Style-Guide: "Deprecate rather than remove when possible... retain the old export marked deprecated rather than deleting it" — the same construct was handled that way for `platform.commentsEnabled`), and the four setting-label strings had their en/es values deleted in the same commit that added their `deprecationInfo` (the established pattern keeps values alongside `deprecationInfo` and removes them in a later cleanup — exactly how `%webView_modelTextPanel_installing%` was retired in #2365 → #2588 this same week). If the one-step break is judged necessary here (the whole feature is gone and no external consumers were found), the guide asks for it to be announced in #platform-changes.

https://github.com/paranext/paranext-core/blob/ddaa027a56e62b5a5ec6435fadd1fafb60c142f0/extensions/src/paratext-registration/contributions/localizedStrings.json#L64-L90

- The `"..."` → `"…"` edits to `%mainMenu_openInternetSettings%` and `%mainMenu_openParatextRegistration%` mutate existing key values in place (en and es), which the Localization Guide's "Existing Strings Are Immutable (CRITICAL)" rule prohibits — and the change runs backwards relative to the Code-Style-Guide, which prescribes literal `"..."` on menu items that open dialogs; every other `%mainMenu_*%` value in the repo uses ASCII dots, so this also leaves `platform-get-resources`' menu item inconsistent.
- New Jira ticket references in long-lived content: the replacement comment in `main.ts` cites PT-4175, and all four new `deprecationInfo` messages cite PT-4224/PT-4175, while the existing deprecation messages in the repo cite none — the same "comments should stand without reference to Jira issues" feedback irahopkinson gave on #2576. https://github.com/paranext/paranext-core/blob/ddaa027a56e62b5a5ec6435fadd1fafb60c142f0/extensions/src/paratext-registration/src/main.ts#L27-L29

Minor notes that did not make the cut: the new comment's "Startup registration is handled by the renderer-side first-run overlay" slightly overstates today's state — the overlay's Identify step is still a placeholder until #2591 merges, so until then a first-run user is gated but cannot register through it; and the PR body's verification note says the reducer "returns waitForRegistration when doesUserHaveValidRegistration is false", but 'invalid' actually routes to startWizard (waitForRegistration is the 'unknown' path) — the simple-mode outcome is still covered, just via a different mechanism than described. Also checked and ruled out: power-mode users were not regressed by this removal (the removed prompt was opt-in-off by default and broken by the race; power mode is deliberately never gated by first-run, and the menu-driven registration form remains).

jolierabideau (approved):
> @jolierabideau reviewed 4 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2592)* status: all files reviewed, 2 unresolved discussions (waiting on katherinejensen00).

___
*[`extensions/src/paratext-registration/contributions/localizedStrings.json` at r1](https://reviewable.io/reviews/paranext/paranext-core/2592#-OyZaenV5Lv5plXHGXQo:-OyZaenV5Lv5plXHGXQp:b8rb0ka) ([raw file](https://github.com/paranext/paranext-core/blob/cc1ed90f0f1933c3e65959a1452ec3d5a0625ed9/extensions/src/paratext-registration/contributions/localizedStrings.json)):*
NIT From Claude

**`localizedStrings.json` — inconsistent/verbose phrasing across the four new `deprecationInfo` messages (Minor).** Two entries use _"Removed in PT-4224 when the paratextRegistration.shouldShowOnStartup setting was removed. The startup registration prompt this settings group {described/labeled} is now handled by…"_ and the other two use _"Removed in PT-4224. The … setting was removed because…"_. Four sibling entries describing the same removal read inconsistently, and all four are noticeably longer than the existing one-sentence deprecation messages in this file (e.g. _"The registration in Paratext 10 Studio is no longer shared with Paratext 9"_). Normalize to one concise phrasing (and, per the bot's point, drop the Jira refs).
___
*[`extensions/src/paratext-registration/src/main.ts` line 28 at r1](https://reviewable.io/reviews/paranext/paranext-core/2592#-OyZaZ4h4vYzwssckLTn:-OyZaZ4h4vYzwssckLTo:b4s81yj) ([raw file](https://github.com/paranext/paranext-core/blob/cc1ed90f0f1933c3e65959a1452ec3d5a0625ed9/extensions/src/paratext-registration/src/main.ts#L28)):*
> ```XML
> }
> 
> // Startup registration is handled by the renderer-side first-run overlay — opening WebViews here races WebViewService.
> ```

NIT From Claude

**comment placement & length (Minor).** The new `// Startup registration is handled by the renderer-side first-run overlay…` comment sits directly above `export async function activate(...)`, so it reads like a doc comment for `activate`, but it's really explaining the _absence_ of the fire-and-forget call that used to live _inside_ `activate` (right after the `context.registrations.add(...)` block). Consider moving it to that old call site inside the function body where the rationale is anchored to the thing it's about. It's also a single 131-char line vs. the ~100-char norm in the rest of the file — worth wrapping.

katherinejensen00 (commented):
Thanks for the review, Jolie!

> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2592#-:-Oy_N72u78ASqkr5YmNn:bql5ayq)* status: 2 of 4 files reviewed, 2 unresolved discussions (waiting on jolierabideau).

___
*[`extensions/src/paratext-registration/contributions/localizedStrings.json` at r1](https://reviewable.io/reviews/paranext/paranext-core/2592#-OyZaenV5Lv5plXHGXQo:-Oy_N4oEERVnqBnvzI3q:bt2vr1b) ([raw file](https://github.com/paranext/paranext-core/blob/cc1ed90f0f1933c3e65959a1452ec3d5a0625ed9/extensions/src/paratext-registration/contributions/localizedStrings.json)):*

NIT From Claude

**`localizedStrings.json` — inconsistent/verbose phrasing across the four new `deprecationInfo` messages (Minor).** Two entries use _"Removed in PT-4224 when the paratextRegistration.shouldShowOnStartup setting was removed. The startup registration prompt this settings group {described/labeled} is now handled by…"_ and the other two use _"Removed in PT-4224. The … setting was removed because…"_. Four sibling entries describing the same removal read inconsistently, and all four are noticeably longer than the existing one-sentence deprecation messages in this file (e.g. _"The registration in Paratext 10 Studio is no longer shared with Paratext 9"_). Normalize to one concise phrasing (and, per the bot's point, drop the Jira refs).
</blockquote></details>

Fixed.
___
*[`extensions/src/paratext-registration/src/main.ts` line 28 at r1](https://reviewable.io/reviews/paranext/paranext-core/2592#-OyZaZ4h4vYzwssckLTn:-Oy_N60-5tt2vrrgsA-o:bt2vr1b) ([raw file](https://github.com/paranext/paranext-core/blob/cc1ed90f0f1933c3e65959a1452ec3d5a0625ed9/extensions/src/paratext-registration/src/main.ts#L28)):*

NIT From Claude

**comment placement & length (Minor).** The new `// Startup registration is handled by the renderer-side first-run overlay…` comment sits directly above `export async function activate(...)`, so it reads like a doc comment for `activate`, but it's really explaining the _absence_ of the fire-and-forget call that used to live _inside_ `activate` (right after the `context.registrations.add(...)` block). Consider moving it to that old call site inside the function body where the rationale is anchored to the thing it's about. It's also a single 131-char line vs. the ~100-char norm in the rest of the file — worth wrapping.
</blockquote></details>

Fixed.

PR #1841, opened 2025-09-10
PT-3417: Created FootnoteItem/list components and corresponding stories

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/1841)

PR #2600, opened 2026-07-23
Suppress jsdom CSS-parse dumps that hang Linux CI unit tests

lyonsil:
## Summary

The `npm unit tests` step intermittently **hangs for hours on the Linux CI runner** (it does not just run slow), while Windows and macOS finish in ~3-4 min. Separately, every run's log is bloated with enormous single-line CSS strings. This PR shows both symptoms are the **same root cause** and fixes them at the source.

### Why review this

Not part of the current epic - this is CI infrastructure. It is worth reviewer time now because the hang is actively blocking PRs: a stalled Linux job holds the check "in progress" and, with no `timeout-minutes` set, can burn up to the default 360-minute (6 h) job budget before GitHub kills it.

## Diagnosis

- `platform-bible-react` component tests render components (e.g. the experimental scripture/footnote editor) that inject a `<style>` containing the full **Tailwind v4.2.2** stylesheet.
- **jsdom 26's** CSS engine (`rrweb-cssom`) cannot parse Tailwind v4's modern syntax (`@layer`, `@property`, `@supports` with `oklch()`/`color-mix()`). On failure jsdom raises a `jsdomError` whose `detail` is the **entire ~192 KB stylesheet**, and its `VirtualConsole` forwards that to `console.error`.
- That happens **~50x per run ≈ 9.6 MB, ~89% of the log** - identical on all three OSes.
- On **Linux only**, the volume of these giant writes intermittently **deadlocks the vitest worker→main→runner output pipeline**. Evidence: two independent stuck runs both froze at the byte-identical point - mid-write of the **32nd** dump - inside the first (core) suite; Windows/macOS drained the same output fine. The stall was invisible in "completed run" samples because a stalled run never completes.

## Changes

- `vitest.setup.ts` (shared by the root **core** suite and the **platform-bible-react** suite): wrap `console.error` to drop jsdom's `"Could not parse CSS stylesheet"` errors at the source - in the worker, before vitest formats/serializes them. jsdom cannot parse this CSS regardless and no test relies on its computed styles. All other `console.error` output passes through untouched.

Suppressing at the source removes both the log spam **and** the giant writes that caused the Linux deadlock. (An `onConsoleLog` reporter filter would hide the log text but still let the 192 KB cross the worker→main boundary, so it would not reliably fix the hang.)

## Testing

- [x] Core suite locally: **37 CSS dumps (~megabytes) → 0**, log for one representative test file **396 KB → 3.2 KB**, all **1151 tests still pass**.
- [x] `eslint`, `prettier --check`, and a strict `tsc` on the changed logic all clean.
- [ ] **CI is the real test** for the Linux hang (not locally reproducible - it is CI-pipeline-specific). This PR exists to confirm the Linux `npm unit tests` step stays green and fast.

## AI Involvement

Diagnosis and this one-file change were produced with Claude Code (Opus 4.8) under my guidance: it correlated the stuck in-progress runs, traced the jsdom/Tailwind parse failure, wrote the console filter, and verified the before/after locally. I reviewed the root-cause reasoning and the diff.

## Risk Level

Low - test-infra only, no production code. Worst case is that a genuine future CSS-parse error is not surfaced in test logs; that is acceptable since jsdom cannot parse this CSS and no assertion depends on it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2600)

### DIFF
--- a/vitest.setup.ts
+++ b/vitest.setup.ts
@@ -93,3 +93,45 @@ configure({
     }
   },
 });
+
+// ---------------------------------------------------------------------------
+// Suppress jsdom's "Could not parse CSS stylesheet" errors.
+//
+// jsdom's CSS engine (rrweb-cssom) cannot parse Tailwind v4's modern syntax
+// (@layer, @property, @supports with oklch()/color-mix(), ...). Whenever a
+// component under test injects a <style> containing the Tailwind stylesheet
+// (e.g. the experimental scripture/footnote editor), jsdom fails to parse it
+// and reports a jsdomError whose `detail` property is the ENTIRE ~192 KB
+// stylesheet. jsdom's VirtualConsole forwards that to console.error, so every
+// such render dumps ~192 KB to the test output - roughly 50x per full run
+// (~9.6 MB, ~89% of the CI log).
+//
+// Beyond the log bloat, the sheer volume of these giant writes intermittently
+// deadlocked the vitest worker->main->runner output pipeline on the Linux CI
+// runner, hanging the "npm unit tests" step for hours (it froze mid-write of
+// the ~32nd dump every time). Windows and macOS drained the same output fine.
+//
+// jsdom cannot parse this CSS regardless, and no test relies on its computed
+// styles, so we drop these specific errors at the source - in the worker,
+// before vitest formats and serializes them - which removes both the log spam
+// and the writes that caused the hang. All other console.error output passes
+// through untouched. jsdom's VirtualConsole does a live property lookup on the
+// console object at emit time, so reassigning console.error here is honored.
+const CSS_PARSE_ERROR_MESSAGE = 'Could not parse CSS stylesheet';
+function isJsdomCssParseError(arg: unknown): boolean {
+  if (arg instanceof Error) {
+    if (arg.message.includes(CSS_PARSE_ERROR_MESSAGE)) return true;
+    return 'type' in arg && arg.type === 'css parsing';
+  }
+  return typeof arg === 'string' && arg.includes(CSS_PARSE_ERROR_MESSAGE);
+}
+// The no-console rule guards against stray logging in production code. This is a test-only
+// setup file that must wrap the global console.error to drop jsdom's CSS-parse noise (see the
+// block comment above), so the rule does not apply to the two references below.
+/* eslint-disable no-console */
+const originalConsoleError = console.error.bind(console);
+console.error = (...args: unknown[]) => {
+  if (args.some(isJsdomCssParseError)) return;
+  originalConsoleError(...args);
+};
+/* eslint-enable no-console */


tjcouch-sil (commented):
Nice! I have a few concerns, though: do we know when or why this started happening? Is this an appropriate course of action given the reason this started happening? Why is this only occurring on linux? I don't know of anything recent that would have obviously started this happening. Did it start months ago when we did the shadcn and Tailwind upgrade?

lyonsil:
Good questions - dug into the history, and your instinct is right.

**When/why it started:** It started with the Tailwind 4 + shadcn upgrade (`74a7dd5`, ~May 12), not anything jsdom-side. jsdom has been on 26 since Feb 2025, and the tests that inject the stylesheet (footnote list/editor, #1841/#1904) rendered fine under Tailwind v3 for months. What broke parsing is Tailwind v4's modern CSS - `@property`, `@layer`, `oklch()`, `color-mix()` in `theme.css`/`utilities.css` - which jsdom's CSS engine (`rrweb-cssom`) can't parse. On each parse failure jsdom hands the *entire* ~192 KB sheet to `console.error`. So: no new test, no new jsdom - the v4 syntax meeting the pre-existing jsdom is the trigger.

**Why only Linux:** The parse failure and the 192 KB dumps happen identically on all three OSes - that's the log bloat you see everywhere. Only the *hang* is Linux-specific, and it's not about the CSS: it's the vitest worker->main output pipe. The volume of giant writes intermittently hits pipe/IPC backpressure that deadlocks on Linux but drains fine on the Windows/macOS runners. I don't have an instrumented proof of the exact kernel mechanism, but the evidence is strong - two independent stuck runs both froze at the byte-identical point (mid-write of the ~32nd dump).

**Is suppression the right course:** I think so, because the underlying parse failure is an upstream jsdom/`rrweb-cssom` limitation we can't fix from here, and the alternatives are worse - we're not giving up Tailwind v4 syntax, and stripping the stylesheet out of component tests is brittle. No test asserts on computed styles from that sheet, so nothing of value is lost. The filter is narrow (the "Could not parse CSS stylesheet" message / `css parsing` error type only), and doing it in the worker before serialization is deliberately the right *layer*: it kills both the log bloat and the writes that cause the hang. An `onConsoleLog` reporter filter would hide the text but still let the 192 KB cross the worker boundary, so it wouldn't reliably fix the hang. When a future jsdom can parse v4 CSS, we just drop the filter.

And even setting the Linux hang aside: these dumps are ~89% of the test log (~9.6 MB/run). That noise makes investigating a *real* test failure genuinely hard - the actual error is buried in megabytes of duplicated stylesheet. Cutting it is worth doing on that basis alone, independent of whether it's the direct cause of the hang.

tjcouch-sil (approved):
Gotcha. Thanks for explaining!

PR #2602, opened 2026-07-23
Hide Home icon & menu items in Simple mode, plus toolbar styling

captaincrazybro:
# Code Review Summary

**Branch**: hide-home-icon-and-menu-items-simple-mode

**Base**: origin/main

**Date**: 2026-07-23

**Files changed**: 39

## Overview

This branch implements ticket T-C ("hide the Home icon button and trim the scripture-editor's
top menu in Simple mode") plus five additional Simple-mode-only toolbar styling enhancements
requested by the author after initial planning: a taller toolbar (`tw:h-14`), a ghost variant +
trigger chevron + fit-content width on the top `BookChapterControl`, and a rounded pill container
+ divider around the back/forward reference-history buttons. Power mode is required to render
pixel-identical to before in every case.

Mechanically, this is two independent changes: (1) an `isHiddenInSimple` flag on the shared
`MenuItemBase` type, read by `menu-data.service-host.ts` via a live `platform.interfaceMode`
subscription to filter menu items in both `getMainMenu()` and `getWebViewMenu()`; and (2) a set of
`isPowerMode`-gated props/classNames applied directly at each toolbar control's point of use, with
two shared components (`BookChapterControl`, `NavigationHistoryButtons`) gaining small opt-in
props (defaulting to off) so only this toolbar's usage is affected.

During the review interview, one additional real bug was found and fixed: the workspace-updating
overlay's top inset was never updated for the new taller Simple-mode toolbar (a gap the original
plan had explicitly flagged as deliberately deferred), and a test-coverage gap in the menu-item
filtering (only `topMenu` was exercised, not `contextMenu`) was closed. Two minor style nits were
also fixed, and one design-guide deviation (the BCV chevron's icon choice) was consciously kept
as-is per the author's explicit call to prioritize visual parity with the project picker's own
chevron over the app's single-select combobox convention.

Screenshot:
Screenshot from 2026-07-23 18-40-43

## API Changes

- `lib/platform-bible-react` (`BookChapterControlProps`): Added optional `showTriggerChevron?: boolean` prop (default `false`) — renders an `IconSelector` chevron at the end of the trigger when `true`.
- `lib/platform-bible-react` (`NavigationHistoryButtonsProps`): Added optional `groupClassName?: string` prop (default `undefined`, applied to the outer `ButtonGroup`) and optional `showDivider?: boolean` prop (default `false`, renders a `ButtonGroupSeparator` between the back/forward buttons).
- `lib/platform-bible-utils` (`MenuItemBase`): Added optional `isHiddenInSimple?: boolean` field, plus a matching `isHiddenInSimple` property in `menuDocumentSchema`'s menu-item JSON schema.
- No changes to `papi.d.ts`, `@papi/*` modules, or any extension `.d.ts` type-declaration files.

None.

- [x] **`src/renderer/components/overlays/overlay-workspace-updating.component.tsx` hardcoded `top: 48`, unaffected by the new taller Simple-mode toolbar** _(fixed during review: `WorkspaceUpdatingOverlayPresentational` now takes a required `isPowerMode` prop; `WorkspaceUpdatingOverlay` computes it via `useIsPowerMode()`. `top` is now `48` in Power mode / `56` in Simple mode. Updated the Storybook story (added `isPowerMode` to `Default`, added a new `SimpleMode` story) and added two new tests pinning both values.)_
- [x] **`menu-data.service-host.test.tsx`'s Simple-mode filtering tests only exercised `topMenu`, not `contextMenu`** _(fixed during review: extended `MOCK_MENU_DATA_WITH_HIDDEN_ITEM`'s `contextMenu.items` with a hidden entry and added two parallel tests — excludes in simple mode, includes in power mode.)_
- [ ] ~~`BookChapterControl`'s new chevron uses tabler's `IconSelector`, but `component-choices.mdx` prescribes lucide's `ChevronDown` for a single-select combobox that closes on selection (confirmed `BookChapterControl` does), and every sibling `advanced/` component uses `lucide-react` for this affordance — `IconSelector` also introduces a second icon library into this file~~ _(author's explanation: wants the BCV chevron to visually match the project picker's own dropdown chevron exactly, as an intentional design choice — not an oversight. If the project picker's icon changes in the future, this chevron should be updated to match it too.)_
- [ ] ~~`book-chapter-control.types.ts`'s JSDoc for `showTriggerChevron` still says it renders "a `ChevronsUpDown` indicator," but the implementation renders `IconSelector`~~ _(author's explanation: same as above — the icon-choice tradeoff was consciously kept, so the doc mismatch is an accepted side effect of that decision rather than a bug to fix right now.)_
- [x] **Chevron used `tw:ml-2` (physical margin) instead of the logical `tw:ms-2`, inconsistent with this same component's own RTL-awareness (`Direction`/`readDirection` usage elsewhere in the file) and `direction.stories.tsx`'s explicit "Bad example" callout** _(fixed during review: changed to `tw:ms-2`.)_

Author defended the icon-choice and stale-doc findings together as one intentional tradeoff (visual parity with the project picker over the single-select combobox guideline), and asked only for the RTL margin issue and the two coverage/behavior gaps to be fixed. All fixes verified with passing tests and clean lint.

### Minor — Consider

- [x] **Redundant `!isPowerMode &&` ternary in `platform-bible-toolbar.tsx`'s `SelectTrigger` className — the whole `<Select>` block is already gated by an outer `{!isPowerMode && (...)}`, so the condition was always true at that point** _(fixed during review: simplified to an unconditional className string. Confirmed via `git blame` that the outer gate pre-dates this branch, but the redundant inner ternary was introduced by this branch's own commit.)_
- [x] **`tw:bg-muted/60` is a one-off opacity shade on the reference-history pill container; every other `bg-muted/*` usage in the codebase is `/50`, `/80`, or `/30`, and `theming.mdx` calls out `/50` as the shade to reuse** _(fixed during review: changed to `tw:bg-muted/50` in both `reference-history-buttons.component.tsx` and the `groupClassName` JSDoc example in `navigation-history-buttons.component.tsx`.)_

Author asked for both to be fixed after confirming (via `git blame`) that the SelectTrigger redundancy was introduced by this branch rather than pre-existing.

#### Shared Regions Modified

None — no `#region shared with <url>` markers found in any changed file.

#### Extension Config Changes

- [n/a] `extensions/src/platform-scripture-editor/contributions/menus.json` — extension-specific content change (added `isHiddenInSimple: true` to one menu item), not a config/build file; no propagation needed.

## Positive Observations

- All new optional props (`showTriggerChevron`, `groupClassName`, `showDivider`, `isHiddenInSimple`) default to values that preserve prior behavior — genuinely non-breaking, each backed by an explicit "default unchanged" test.
- `isHiddenInSimple`'s type and JSON schema were updated together and kept in sync; `menu-document-combiner.test.ts` confirms the flag validates and survives contribution combination.
- `filterItemsForSimpleMode` correctly runs after `defaultWebViewTopMenu`/`defaultWebViewContextMenu` are merged into each web view's menu, so default-menu items are covered by the Simple-mode filter too, not just per-webview-contributed items.
- `getDockLayoutOuterInset`'s new 56px Simple-mode top inset is cross-tested against the toolbar's new `tw:h-14`, with both the Power-mode (48px) and Simple-mode (56px) cases pinned by tests, directly verifying the "Power mode pixel-identical" requirement — and the code comments in both files cross-reference each other so the coupling is discoverable (this is exactly the coupling that was missed for the workspace-updating overlay, caught and fixed during this review).
- `platform-bible-toolbar.test.tsx` explicitly asserts Power-mode values (no `ghost`, no chevron, `tw:w-96`) alongside the new Simple-mode values, not just the new path.
- Comprehensive test coverage: every production file changed in this diff has a corresponding test-file change in the same diff.
- No hardcoded user-facing strings were introduced; the only new UI addition is an icon, not text.
- The `menu-data.service-host.ts` live-mode-change subscription (`#subscribeToInterfaceMode`) is well-isolated, documented, and guards races via an `isDisposed` check, explicitly mirroring a documented pattern from `web-view.service-host.ts`.
- The pill divider correctly uses `ButtonGroup`/`ButtonGroupSeparator` rather than manual spacing, matching `design-principles.mdx`'s guidance.
- The Home button keeps its required tooltip even though it's now gated behind `isPowerMode`.

## Interview Notes

Author's stated purpose matched the PURPOSE given at the start of this review (T-C plus five
styling enhancements, Power mode pixel-identical throughout).

Key decisions and reasoning surfaced during the interview:

- Confirmed the workspace-updating-overlay inset gap was the same "known, deliberately
  out-of-scope" issue called out in the original implementation plan, and chose to fix it now
  rather than leave it deferred further.
- Requested the additional `contextMenu` filtering test after confirming the gap was real (the
  existing tests only ever exercised an empty `contextMenu.items` fixture).
- On the chevron icon-choice tension: the UX analysis pass recommended lucide's `ChevronDown` per
  `component-choices.mdx` (single-select comboboxes that close on selection should use a
  down-only chevron, not the up-down "selector" glyph), but the author explicitly chose to keep
  tabler's `IconSelector` instead, prioritizing exact visual parity with the project picker's own
  dropdown chevron over the app's documented single-select convention. The author was explicit
  that this is an intentional, acknowledged tradeoff — not an oversight — and that the chevron
  should be revisited if the project picker's own icon ever changes. The stale JSDoc naming
  (`ChevronsUpDown` instead of `IconSelector`) was left as-is for the same reason.
- Verified via `git blame` (at the author's request) that the redundant `SelectTrigger` ternary
  was newly introduced by this branch before agreeing to fix it — the author wanted to distinguish
  pre-existing code from new redundancy before deciding.

Design understanding (Step 3.3): asked to walk through the most complex or non-obvious part of the
changes, the author summarized the branch as "removes the Home button for 10 Simple, adds some
enhancements to the styling in the top tab bar, and then limits the items in the project menu in
10 Simple" — an accurate purpose-level summary. This didn't walk through the underlying mechanism
(the `isHiddenInSimple` live-settings-subscription plumbing, or the toolbar-height/dock-layout-inset
coupling) in detail, but the author demonstrated working knowledge of that mechanism-level detail
through their specific, technically-grounded decisions earlier in the interview (correctly
recalling the overlay gap from the original plan, reasoning precisely about the chevron tradeoff,
and using `git blame` to verify provenance before agreeing to a fix). No areas of the changes were
deferred to AI or left unexplained — the author was engaged throughout and, notably, explicitly
overrode one of the automated analysis findings with clear, stated reasoning rather than accepting
it by default.

## In-Review Quality Check

All Step 4.1 checks passed after the in-review fixes:

- **Library rebuild**: `npm run build:platform-bible-react` was run (both `book-chapter-control.component.tsx` and `navigation-history-buttons.component.tsx` changed); produced new `dist/` diffs — including `dist/experimental.{cjs,js}`, which an earlier manual rebuild during the interview had missed — all re-staged.
- **Typecheck**: skipped per this project's established convention (lint + tests are sufficient; typecheck is too slow to use as a completion gate).
- **Lint**: `npm run lint` passed with 0 errors repo-wide (worked around this worktree's nested-directory ESLint config-resolution quirk by temporarily setting `root: true`, reverted afterward — confirmed no diff on `.eslintrc.js`). Two pre-existing `no-console` warnings remain in an unrelated file not touched by this review.
- **Format**: Prettier run against only the files changed in this review reported all in-scope files unchanged.
- **Tests**: `npm test` passed, exit code 0, every workspace green — no assertion failures.
- **Final state**: all fixes staged (`git add`), nothing committed yet — confirmed via `git status --short`.

## Suggested Review Focus

- [ ] The BCV chevron's icon choice (`IconSelector` vs. the design guide's `ChevronDown` for single-select comboboxes) — the author explicitly overrode the UX-guide recommendation to match the project picker visually. Worth a quick gut-check on whether that's the right long-term call, since the JSDoc will remain slightly stale (`ChevronsUpDown` named but `IconSelector` used) as an accepted tradeoff.
- [ ] The `isHiddenInSimple` menu-filtering mechanism (flag + live `platform.interfaceMode` subscription in `menu-data.service-host.ts`) — a new pattern for this codebase; worth confirming the live-update-on-mode-change behavior once in person.
- [ ] The now-mode-aware workspace-updating overlay inset — a bug fixed during this review (previously flagged as a known, deferred gap); worth a quick visual check of the "Updating project view" spinner in Simple mode during a project switch.

AI-assisted — [session 1](<session URL>)

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2602)


lib/platform-bible-utils/src/extension-contributions/menus.model.ts
```diff
@@ -56,6 +56,12 @@ export type MenuItemBase = OrderedItem & {
   tooltip?: LocalizeKey;
   /** Additional information provided by developers to help people who perform localization */
   localizeNotes: string;
+  /**
+   * Set to `true` to hide this menu item while the app is in Simple interface mode
+   * (`platform.interfaceMode === 'simple'`). Omit (or set to `false`) for items that should show in
+   * both modes — most items need no value here at all.
+   */
+  isHiddenInSimple?: boolean;
```
tjcouch-sil: Instead of this narrow property that works for you but not for power or for other modes in the future, what do you say to reworking this property to be something like `hiddenInterfaceModes: InterfaceMode[]` (not sure I love the name) and making an `InterfaceMode: 'simple' | 'power'` type which is the type union of interface modes (then you can replace here too https://github.com/paranext/paranext-core/blob/main/src/declarations/papi-shared-types.ts#L275)?

tjcouch-sil (commented):
Nice! I didn't do a whole review or anything; just saw the title and was interested in how you did the hiding in simple. Just left a suggestion for you :) thanks!

katherinejensen00 (changes requested):
## Review — 2 items to address before merge

Nice, tightly-scoped PR, and the write-up already caught most of what I'd flag. Two latent gaps remain in how the Simple-mode `isHiddenInSimple` guarantee is enforced for the **main menu**. Neither is triggered by the single (web-view) flag this PR ships, but both should be closed or explicitly documented before this becomes the mechanism other contributors build on.

### 🚫 1. macOS native menu bar bypasses the Simple-mode filter
**`src/extension-host/services/menu-data.service-host.ts:85` (`getUnlocalizedMainMenu`)**

`filterItemsForSimpleMode` is applied to `getMainMenu`/`getWebViewMenu` but not `getUnlocalizedMainMenu`/`subscribeUnlocalizedMainMenu`, which is what builds the macOS native application menu (`subscribeCurrentMacosMenubar`, registered unconditionally on darwin). The moment any **main-menu** item sets `isHiddenInSimple: true`, it will still appear in the macOS app menu in Simple mode while being hidden everywhere else. Since the tests assert main-menu filtering works, this is an inconsistency. Please either apply the filter in the unlocalized getter too, or add a comment documenting that the macOS menu is intentionally out of scope.

### 🚫 2. Main-menu filtering doesn't live-update on a mode switch
**`src/renderer/components/platform-bible-toolbar.tsx:214` (main-menu `usePromise`)**

`#subscribeToInterfaceMode` fires `notifyUpdate('*')` so menus stay live on a mode change, but the toolbar loads the main menu via `usePromise` keyed on `[updateMenuData]`, which doesn't subscribe to the data provider — so the update never reaches it (it re-pulls only when the menu is reopened). The web-view path uses `useData` and does update live. The "stays live if the user switches modes" comment therefore over-promises for the main menu. Please subscribe the main menu to menu-data updates, or adjust the comment/behavior expectation.

Both are latent today (only a web-view item uses the flag, and that path is correct), so converting these to tracked follow-ups instead of blocking is reasonable if you'd prefer.

katherinejensen00 (commented):
## Optional / non-blocking suggestions

None of these block merge — polish and forward-looking notes.

**3. [Minor] Stale TSDoc icon name** — `lib/platform-bible-react/src/components/advanced/book-chapter-control/book-chapter-control.types.ts:107`. The `showTriggerChevron` doc says it renders a `ChevronsUpDown` indicator, but it renders tabler's `IconSelector`. Keeping `IconSelector` for parity with the toolbar's project-picker `<Select>` is a fine call — just reword the doc so it doesn't name a component that isn't used.

**4. [Minor] 48/56 toolbar-height coupling duplicated 3×** — `src/renderer/components/docking/platform-dock-layout-positioning.util.ts:96` (also the toolbar `h-12`/`h-14` class and the workspace-updating overlay). Coupling is enforced only by comments today; a shared `TOOLBAR_HEIGHT` constant pair that both the px insets and the class map derive from would make drift impossible.

**5. [Minor] Per-call allocation** — `src/extension-host/services/menu-data.service-host.ts:70`. `getMainMenu`/`getWebViewMenu` rebuild a filtered object graph on every call, even in Power mode where nothing changed. The filtered result only changes on a mode flip (already tracked), so it could be computed once in the subscription callback. (Also drops the previous referential stability of the returned menu — no consumer relies on it today.)

**6. [Minor] Untested error branch** — `src/extension-host/services/menu-data.service-host.ts:165`. The `isPlatformError(newMode)` warn-and-return path in the subscription callback has no test coverage (the live-update test casts the callback to a non-error signature).

**7. [Nit] Spurious initial notify** — `src/extension-host/services/menu-data.service-host.ts:154`. The initial `notifyUpdate('*')` fires even in Power mode (no state change); the change-callback correctly guards on an actual change — the seed could too.

**8. [Nit] Fragile test flush** — `src/extension-host/services/menu-data.service-host.test.tsx:293`. `await Promise.resolve()` ×2 across six tests couples to the exact microtask count of the fire-and-forget subscription; a `flushMicrotasks()`/readiness helper would be sturdier against an added `await`.

**9. [Nit] Implicit variant default** — `src/renderer/components/platform-bible-toolbar.tsx:466`. `triggerVariant={isPowerMode ? undefined : 'ghost'}` relies on `BookChapterControl`'s default being `'outline'`; naming both branches (`isPowerMode ? 'outline' : 'ghost'`) is clearer and won't silently change if the default moves.

**10. [Nit] Story misalignment** — `src/renderer/components/overlays/overlay-workspace-updating.stories.tsx`. The new `SimpleMode` story's simulated toolbar strip is hardcoded `tw:h-12` (48px), but the overlay insets 56px in Simple mode, so they don't line up in that story (Storybook-only).

lib/platform-bible-utils/src/extension-contributions/menus.model.ts
captaincrazybro: Done.

captaincrazybro (commented):
> @captaincrazybro made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2602)* status: 0 of 42 files reviewed, 1 unresolved discussion (waiting on tjcouch-sil).

katherinejensen00 (commented):
Two nits on the latest round (interface-mode generalization + live-update fix). Both non-blocking — the blockers and the doc fix all look correctly resolved, and the `useData` subscription is a clean improvement over the old poll.

src/extension-host/services/menu-data.service-host.ts
```diff
@@ -96,6 +141,49 @@ class MenuDataDataProviderEngine
...
+   * async): failures are logged, not thrown, and `isDisposed` guards against setting state or
+   * leaking a subscription if this engine is disposed before the async work resolves — mirrors the
+   * `subscribeToInterfaceMode` pattern in `web-view.service-host.ts`'s `registerDockLayout`.
+   */
+  #subscribeToInterfaceMode(): void {
+    const subscribe = async () => {
+      try {
+        const initialMode = await settingsService.get('platform.interfaceMode');
+        if (this.isDisposed) return;
+        this.currentMode = initialMode;
```
katherinejensen00, line 157: Nit: the initial read assigns the raw result directly (`this.currentMode = initialMode`), but the subscription callback just below guards `isPlatformError(newMode)` and bails. Since `settingsService.get('platform.interfaceMode')` can resolve to a `PlatformError`, `currentMode` — typed `InterfaceMode` — can transiently hold a `PlatformError` here. It degrades safely to 'show everything' (same as the old `=== 'simple'` behavior), so there's no behavior change, but mirroring the callback's guard (or keeping the `'power'` default on error) would keep the field honest to its declared type.

lib/platform-bible-utils/src/extension-contributions/menus.model.ts
```diff
@@ -16,6 +16,9 @@ export type OrderedItem = {
   order: number;
 };
 
+/** An interface mode a menu item can be hidden in. */
+export type InterfaceMode = 'simple' | 'power';
```
katherinejensen00, line 27: Nit: this `InterfaceMode` duplicates the canonical `SettingTypes['platform.interfaceMode']` (which `src/renderer/hooks/use-interface-mode.hook.ts` derives from rather than re-declaring). `platform-bible-utils` can't import app-level `SettingTypes` due to dependency layering, so a literal copy is probably unavoidable — but the two unions can silently drift if a third interface mode is ever added. A short `// keep in sync with platform.interfaceMode` comment would mitigate.

katherinejensen00 (approved):
Approving — both blockers from the earlier review are resolved:

- **macOS native menu bar** now applies the interface-mode filter via `getUnlocalizedMainMenu` (with a test pinning macOS-menubar consistency).
- **Main-menu live-update** now uses a `useData(...).MainMenu` subscription instead of a one-shot fetch, so interface-mode and localization changes reach it without reopening the menu — and it retires the old `%`-label `setTimeout` poll.

Nice bonus that the `isHiddenInSimple` → `hiddenInterfaceModes` generalization landed here too, and the stale `ChevronsUpDown` doc is fixed. The two remaining items are non-blocking nits (initial-read `isPlatformError` guard; `InterfaceMode` duplicating the canonical setting type) — fine to address or leave as you see fit.

src/extension-host/services/menu-data.service-host.ts
captaincrazybro: Skip, not needed. Settings service doesn't return `PlatformError`

captaincrazybro (commented):
> @captaincrazybro made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2602)* status: 0 of 42 files reviewed, 3 unresolved discussions (waiting on katherinejensen00 and tjcouch-sil).

captaincrazybro (commented):
> @captaincrazybro resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2602)* status: 0 of 42 files reviewed, 2 unresolved discussions (waiting on katherinejensen00 and tjcouch-sil).

katherinejensen00 (approved):
Thanks for making those changes, Levi, and for improving the Simple layout!

captaincrazybro (commented):
> @captaincrazybro resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2602)* status: 0 of 42 files reviewed, 1 unresolved discussion (waiting on tjcouch-sil).

captaincrazybro (commented):
Looks good

lib/platform-bible-utils/src/extension-contributions/menus.model.ts
```diff
@@ -16,6 +16,16 @@ export type OrderedItem = {
...
 
+/**
+ * An interface mode a menu item can be hidden in.
+ *
+ * Keep in sync with `SettingTypes['platform.interfaceMode']` in
+ * `src/declarations/papi-shared-types.ts` — this package can't import that app-level type
+ * (dependency layering runs the other way), so this is an independently declared, structurally
+ * identical union rather than a shared one.
+ */
+export type InterfaceMode = 'simple' | 'power';
```
tjcouch-sil, line 27: Instead of having to say "keep this in sync with the setting", could you instead use this type for the setting too? Then I think they wouldn't get out-of-sync, right?

Sorry if I missed something! I'm on mobile because my power is out.

tjcouch-sil (commented):
Nice! Just one last little thought. It's hard to review on the mobile interface, so I may have missed something. Anyway, feel free to merge whenever; I don't want to block this. Note that I didn't review anything but the menu definition change.

PR #2603, opened 2026-07-23
PT-4208: Add icon-only Simple-mode tabs with live-measured collapse

captaincrazybro:
# Code Review Summary

**Branch**: pt-4208-implement-simple-tab-icons

**Base**: origin/main

**Date**: 2026-07-23

**Files changed**: 59

## Overview

This branch implements icon-only Simple-mode tabs: Column 3 (Resources & Tools) tabs get per-source icons that adapt to theme and selection state, and the column collapses each tab to icon-only when it no longer has room for icon+text — measured live against the tab bar's real content width rather than a fixed pixel threshold, so it stays correct across locales, tab counts, and label lengths. The branch also includes related Simple-mode dock layout/styling work (divider rendering, tab padding, `TabToolbar` overflow handling) and a shared `useTabIconSelection` hook extracted into `platform-bible-react` for reuse across three extension web views.

During review, four findings were fixed in place: icon-only tabs now get a distinguishing `aria-label` instead of a generic "Tab" label for every collapsed tab; a redundant tooltip that mirrored an already-visible title was suppressed; a dead pass-through wrapper was simplified; and a missing test for the `MutationObserver` recompute path was added. Three duplication findings (a theme-subscription hook, an `isPowerMode` hook, and a visibility-detection reimplementation) were investigated and initially dismissed — the first two because extracting them into `platform-bible-react` would violate the library's deliberate, already-documented "no `@papi/frontend` dependency" boundary, and the third because closer inspection showed the two implementations are not actually equivalent (different state contracts, different failure modes on transient zero-width panels).

Two accidentally-committed docs belonging to the separate, already-merged PT-3958 branch were removed after the initial review. In a follow-up pass, the author revisited the visibility-detection finding, and after further reconsideration of the failure-mode analysis, it was implemented after all (see Post-Review Follow-Up Fixes below). The author also brought four additional, independently-diagnosed bugs to this same follow-up pass — each verified against the actual code (and, where relevant, live in a running instance) before being fixed or, in one case, correctly identified as not currently reachable.

## API Changes

- `lib/platform-bible-react`: Added new exports `useTabIconSelection()`, `pickTabIconUrl()`, and `TabIconUrls` type (from new file `src/hooks/use-tab-icon-selection.hook.ts`, re-exported via `src/index.ts`).
- `lib/platform-bible-react`: Modified `TabToolbar` (`src/components/advanced/tab-toolbar/tab-toolbar.component.tsx`) — internal Tailwind class changes only (`tw:border`→`tw:border-b`, `tw:flex-wrap`→`tw:flex-nowrap`, dropped `tw:h-full` on the three area wrappers); no prop/signature change.
- `lib/platform-bible-react/dist/*`: Rebuilt output matching the source changes above, committed as its own dedicated commit.
- `lib/platform-bible-utils`: No changes.
- `papi.d.ts` / `@papi/*` modules: No changes. `isClosable` and `iconUrl` (both newly used by several extensions' web-view providers in this branch) were already present on `WebViewDefinition` before this branch.
- Extensions' type declaration files (`extensions/src/*/src/types/*.d.ts`): No changes.

None.

- [x] **`useTabIconSelection` (`lib/platform-bible-react/src/hooks/use-tab-icon-selection.hook.ts`) reimplemented tab-selection detection via a 500ms poll of `frameElement.offsetParent`, instead of reusing the already-exported `useViewVisibility` hook**, which detects the same rc-dock `display:none` condition event-driven via `IntersectionObserver`. _(Initially dismissed during the interview — `useViewVisibility` returns a strict boolean with no "unknown" state, while `pickTabIconUrl` needs a genuine three-state result for a real fallback icon variant, and `IntersectionObserver`'s default root is the iframe's own viewport, which seemed like it could report "not intersecting" for the genuinely-active tab during a column resize. On revisiting in a follow-up pass, both concerns didn't hold up: `useViewVisibility`'s synchronous initial value means the "unknown" state was already just a bootstrap artifact of the old technique, not something worth preserving, and `IntersectionObserver` only reports non-intersecting at genuinely zero width, which `panelLock.minWidth` prevents in practice. Fixed during follow-up: `useTabIconSelection` now calls `useViewVisibility()` directly; `pickTabIconUrl` keeps its existing signature for standalone testability. 6 tests rewritten to exercise the new `IntersectionObserver`-based path, revert-tested.)_
- [ ] ~~The same ~15-line "subscribe to `papi.themes.subscribeCurrentTheme`, guard the disposed-unmount race, unsubscribe on cleanup" block is duplicated verbatim across `resource-text-panel.web-view.tsx`, `scripture-text-grid.web-view.tsx`, and `comment-list.web-view.tsx`.~~ _(dismissed — this is a documented, deliberate decision recorded in the design spec (`docs/superpowers/specs/2026-07-22-implement-simple-tab-icons-design.md`): `platform-bible-react` has zero dependency on `@papi/frontend` today, and adding one to share this subscription would be "a real architectural change, not something to introduce silently." Author chose to preserve that boundary.)_
- [ ] ~~The `isPowerMode` detection block (`useSetting('platform.interfaceMode', ...)` → error-check → `useMemo`) is now duplicated a 4th time across extension web views, with zero extraction into a shared hook.~~ _(dismissed — same architectural reason as above: `useSetting` is imported from `@papi/frontend/react`, so a shared hook would hit the identical "no PAPI dependency" boundary. Author chose to preserve the boundary here too.)_
- [x] **Icon-only tabs lost their distinguishing `aria-label`** — all four Column 3 tabs announced as the same generic "Tab" string to a screen reader once collapsed, since the only visible differentiator (the title text) is exactly what got hidden. _(fixed during review: `aria-label` now falls back to the resolved title when `isIconOnly` is true, keeping the generic label otherwise; 2 regression tests added and revert-tested.)_

Author defended two of the three duplication findings as a deliberate, already-documented architectural boundary (both remain dismissed); the visibility-detection finding was fixed in a follow-up pass after reconsideration; the aria-label finding was fixed in-review.

### Minor — Consider

- [ ] ~~The 500ms polling in `useTabIconSelection` runs for the lifetime of every web view using the hook, even while backgrounded.~~ _(dismissed — superseded by the Important-finding discussion above; the polling approach may be the correct tradeoff for this specific "which tab is selected" question.)_
- [x] **Dead pass-through wrapper** in `platform-dock-layout.component.tsx:136`: `loadTab={(savedTabInfo) => loadTab(savedTabInfo)}` called `loadTab` with exactly the argument it already accepts (leftover from a prior commit that removed a third argument but didn't restore the plain form). _(fixed during review: simplified to `loadTab={loadTab}`.)_
- [x] **Missing test coverage for the `MutationObserver` recompute path** in `platform-tab-title.component.tsx` — only the `ResizeObserver` path was exercised by tests. _(fixed during review: added a test that mutates a sibling `.platform-tab-title-measure` clone's DOM content directly and asserts the icon-only recompute fires without any panel resize; revert-tested.)_
- [x] **Tooltip mirrored the tab title unconditionally**, including at full width where the title is already visible on-screen, going against the "don't repeat the visible label" tooltip guideline. _(fixed during review: `PlatformTabTitle` now suppresses the tooltip when it would only repeat the already-visible title and the tab isn't icon-only; still shows when icon-only, or when the tooltip conveys something beyond the title. 3 tests added and revert-tested.)_
- [ ] ~~`TabToolbar`'s `flex-wrap` → `flex-nowrap` fix (a real, well-verified overlap-bug fix) trades garbled overlap for silent clipping with no overflow affordance for toolbar buttons that don't fit.~~ _(dismissed — author confirmed this was already verified: no Power-mode `TabToolbar` consumer relies on wrapping to stay reachable at the 300px minimum width.)_
- [ ] ~~The icon-only threshold is computed via `ResizeObserver`/`MutationObserver` rather than the CSS `@container` queries the Responsiveness guideline recommends.~~ _(acknowledged, documented only — no code change needed. Already extensively justified in-code: `container-type` size containment broke rc-dock's content-driven ancestor sizing, confirmed via CDP investigation recorded in the component's comments.)_
- [ ] ~~No Storybook story or documentation page demonstrates the new icon-only/responsive Column 3 behavior, and the existing Simple-mode prototype story is now stale relative to production.~~ _(dismissed — author said no need.)_

Three of seven minor findings were fixed in-review (dead wrapper, missing MutationObserver test, redundant tooltip); the rest were dismissed with author confirmation or already-documented rationale.

#### Shared Regions Modified

None. No `#region shared with <url>` markers were found overlapping changed lines in the reviewed files.

#### Extension Config Changes

- [n/a] No changes to `extensions/` config/build files (`webpack.config.ts`, `tsconfig.json`, `package.json`, `.eslintrc*`) in this diff — all extension changes are source/asset files.

## Post-Review Follow-Up Fixes

Four additional bugs, independently diagnosed by the author, were verified and fixed in a follow-up pass after the initial review (each verified against the actual code — and, where relevant, live in a running instance of this branch — before any fix was made):

- [x] **`useTabIconSelection` reimplemented visibility polling instead of reusing `useViewVisibility`** — see the reopened Important finding above.
- [x] **Text Collection's `isClosable` was hardcoded `false`, unlike its sibling Column 3 providers (`isClosable: interfaceMode === 'power'`)** — `getTabGroup()` routes `isClosable === false` tabs to `TAB_GROUP_RESOURCES`, but `getGroups(isPowerMode=true)` never registers that group, so the tab pointed at an unconfigured rc-dock group in Power mode. Fixed in `extensions/src/platform-scripture-editor/src/main.ts` to match the sibling providers; verified against the already-correctly-parameterized `getTabGroup`/`getGroups` test suite plus the full `platform-scripture-editor` suite (542 tests).
- [x] **Icon effects in `resource-text-panel.web-view.tsx` and `comment-list.web-view.tsx` skipped `updateWebViewDefinition` entirely on a Power-mode early return**, rather than explicitly clearing `iconUrl` — since the merge only touches keys present in the update object, skipping the call left a previously-set Simple-mode icon in place after a runtime Simple→Power transition instead of reverting to text-only. Fixed in both files to call `updateWebViewDefinition({ iconUrl: undefined })` before returning. `tooltip` did not need the same fix in `resource-text-panel.web-view.tsx` — a separate effect there already unconditionally re-evaluates it on every `isPowerMode` change.
- [x] **`platform-tab-title.component.tsx`'s icon-only effect never reset `isIconOnly` to `false` on its Power-mode early return** — since `isPowerMode` is a live subscription (not a one-time read), the effect does re-run on a runtime mode switch, but left the stale Simple-mode value in place. Both modes render the tab title through the same code path that applies the icon-only class/aria-label/tooltip-suppression logic, so a tab collapsed to icon-only in Simple mode stayed stuck that way after switching to Power mode — most concretely reviving the "tooltip mirrors an already-visible title" bug fixed earlier in this same review. Fixed by calling `setIsIconOnly(false)` before the early return; regression test added and revert-tested.
- [ ] ~~`parseFloat(getComputedStyle(navList).columnGap)` could return `NaN` if `column-gap` computed as the string `"normal"` (no explicit gap rule), silently zeroing the icon-only threshold's gap term via the `|| 0` fallback.~~ _(investigated and not fixed — the premise doesn't hold for the current code: `dock-layout-wrapper.component.scss` already sets `gap: vars.$space--xs` unconditionally on `.dock-nav-list` via the `dock-style-platform-bible` class every panel receives (confirmed live via CDP: Column 3's `.dock-nav-list` computes `columnGap: "4px"`, not `"normal"`). Author agreed to skip the defensive hardening given the corrected premise.)_

## Positive Observations

- Exceptionally thorough code comments throughout the SCSS/TSX changes (`platform-tab-title.component.tsx`, `dock-layout-wrapper.simple-mode.scss`, `_simple-mode.scss`) — each documents not just what a rule does but which alternatives were tried, why they failed (with CDP-verified specifics), and what invariant must hold for the current approach to keep working.
- Clean, deliberate extraction of `useTabIconSelection`/`pickTabIconUrl`/`TabIconUrls` into a shared `platform-bible-react` hook, replacing a near-identical duplicate that previously lived only in `platform-scripture-editor`, now reused identically across three call sites (`scripture-text-grid.web-view.tsx`, `resource-text-panel.web-view.tsx`, `comment-list.web-view.tsx`) — old file and its test fully removed, no dangling references.
- `papi.d.ts` was correctly left untouched — the feature reuses pre-existing `isClosable`/`iconUrl` fields rather than growing the PAPI surface unnecessarily.
- The `platform-bible-react` `dist/` rebuild was committed as its own dedicated commit, matching the source diff exactly, rather than hand-edited.
- `TabToolbar`'s CSS-only change is backed by two new regression tests that directly assert against the specific overlap bug being fixed.
- `platform-dock-layout-positioning.util.ts`'s `getGroups`/`getTabGroup`/`getDockLayoutOuterInset` refactor is a clean, well-tested, pure-function style, with both callers correctly updated and solid test coverage for power/simple-mode branches and edge cases.
- `extensions/src/platform-scripture-editor/src/model-text-panel.component.tsx`'s new truncated header reuses the existing `useTruncationTooltip` hook and wires `onFocus`/`onBlur` alongside pointer handlers, so keyboard users can reach the tooltip too.
- New divider/selection styling in `dock-layout-wrapper.simple-mode.scss` consistently uses existing theme tokens (`var(--border)`, `var(--tab-front-blur)`) rather than hardcoded colors.
- The `pointer-events: none` fix on `.platform-tab-tooltip` was found via a genuine CDP-driven repro (tooltip content stealing a pointerdown meant for the tab underneath), not guessed.
- No new hardcoded UI strings — every new title/tooltip usage reuses pre-existing, unchanged localization keys.
- Exceptionally thorough test coverage across nearly every new/modified logic path, including several explicit regression tests documenting real bugs found via CDP investigation (e.g., the "measures live rather than caching a past reading" test).
- No shadcn/ui files touched, no cross-process boundary violations (no extension code imports `@renderer/*`), no hand-edited `papi.d.ts`.

## Interview Notes

**Stated purpose**: Icon-only Simple-mode tabs — Column 3's tabs collapse to icon-only when the column narrows, measured live rather than via a fixed threshold, plus related Simple-mode dock layout/styling work.

**Key interview outcomes**:

- Three Important findings (visibility-detection duplication, theme-subscription duplication, isPowerMode duplication) were raised, challenged by the author, and on independent re-verification initially found to be either factually incorrect (the visibility-detection case) or already-considered, deliberate architectural decisions documented in the design spec (the two PAPI-dependency-boundary cases). The two PAPI-dependency-boundary findings remain dismissed; the visibility-detection finding was revisited by the author in a follow-up pass, re-examined, and implemented (see Post-Review Follow-Up Fixes) once the earlier failure-mode concerns didn't hold up under closer scrutiny.
- One Important finding (missing per-tab `aria-label` on icon-only tabs) and three Minor findings (dead wrapper, missing `MutationObserver` test, redundant tooltip) were fixed in-review, each with new tests added and revert-tested per TDD discipline.
- Two Minor findings were dismissed based on the author's prior verification (TabToolbar overflow affordance) or explicit preference (no Storybook update needed); one was acknowledged as an already-documented, deliberate guideline deviation requiring no further action.

**Design understanding (Step 3.3) — flag for reviewer**: When asked to walk through the most complex/non-obvious part of the change, the author described a three-tier model: icon+text when there's room, icon-only when there isn't, and — if there still isn't enough room — the tabs falling back to "a dropdown menu" reusing "the RCDock overflow system." An exhaustive code search of the actually-installed rc-dock v3.3.2 (every file in its `lib/` directory) found no such overflow/more-tabs mechanism anywhere — the only dropdown usage in the library is the unrelated tab right-click context menu, and the tab bar's enclosing `.dock-nav-wrap` is set to `overflow: hidden` (silent clipping, not a dropdown). Live verification of this specific behavior was attempted three times during the session (including deliberately setting Column 3's `panelLock.minWidth` to 100px to force the scenario) but was disrupted each time by app-instance mix-ups across worktrees (a different worktree's Platform.Bible instance kept auto-restarting and being mistaken for this branch's instance) and one script/connection failure. The author ultimately confirmed the dropdown behavior is present after a verified-correct instance of this branch was running with the 100px override; this reviewer's own attempted independent verification did not complete (the CDP connection dropped as the instance restarted) before the author's confirmation. **This finding rests on the author's direct observation, not an independently reproduced verification by this review** — worth a quick, direct check in the review meeting, especially since a genuine "overflow dropdown" would need to originate from somewhere other than rc-dock's own tab-bar code (rc-dock v3.3.2 has none), so it's worth confirming exactly what mechanism is producing it.
- With today's fixed 4 tabs and Column 3's normal 300px `panelLock.minWidth`, 4 icon-only tabs only need ~204px, so this fallback tier can't currently be reached in practice regardless of which mechanism is actually responsible for it.

## In-Review Quality Check

- **Typecheck**: passed cleanly, no errors.
- **Lint**: repo-wide `npm run lint` could not run to completion due to a pre-existing, structural issue unrelated to this review (ESLint's legacy config cascade finds two distinct `@typescript-eslint` plugin installations because this worktree is nested inside the main repo's directory tree). The three in-review files were verified lint-clean by invoking ESLint directly against them (`--no-eslintrc --config .eslintrc.js --resolve-plugins-relative-to .`) — zero errors or warnings.
- **Format**: all three in-review files were already correctly formatted; no changes needed.
- **Tests**: full suite passed (1146 tests) except one pre-existing, unrelated failure in `simple-layout.data.test.ts` caused by a stray uncommitted `panelLock.minWidth: 100` edit in `simple-layout.data.ts` left over from live-verification testing during the interview (see Design Understanding note above) — this was not part of the in-review file set, was reverted after being found, and is not reflected in the final diff. All 21 tests in the updated `platform-tab-title.component.test.tsx` passed, including the new aria-label, tooltip, and MutationObserver tests.
- **Auto-fix applied**: a duplicated 3-line comment block above the `aria-label` JSX attribute in `platform-tab-title.component.tsx` (introduced during in-review editing) was found and removed by the quality-check pass; re-verified clean afterward.

## Suggested Review Focus

- [ ] **Confirm the tab-overflow fallback mechanism directly** — the author's mental model includes a "dropdown menu" tier reusing rc-dock's overflow system, but a full source search of the installed rc-dock version found no such mechanism, and this reviewer's own independent verification attempt didn't complete. Worth walking through together, since whatever is producing this behavior (if it truly exists) isn't coming from rc-dock's tab-bar code as currently understood.
- [ ] Spot-check the two remaining dismissed "duplication" findings (theme subscription, isPowerMode) against the reasoning recorded above — both were dismissed on the `platform-bible-react`/`@papi/frontend` architectural boundary, not accepted at face value, but a second pair of eyes is worthwhile given it now affects 4 independent call sites across 2 extensions.
- [ ] Confirm the `aria-label`, tooltip-suppression, and dead-wrapper fixes made during this review match the intended behavior when exercised live (not just via unit tests).
- [ ] **Confirm the four post-review follow-up fixes match intended behavior live**, especially the `isClosable`/rc-dock-group fix (Text Collection in Power mode) and the icon-clearing/icon-only-reset fixes (toggle Simple↔Power mode at runtime with a Column 3 tab already collapsed to icon-only, and confirm the icon and tooltip both revert correctly).

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2603)


src/renderer/components/docking/platform-tab-title.component.tsx
```diff
@@ -280,9 +294,127 @@ export function PlatformTabTitle({
...
+  //    settles to "however big my current content is." Confirmed via CDP: once collapsed to
+  //    icon-only (or once comfortably fitting), `.dock-nav-wrap`'s clientWidth got stuck reporting
+  //    its own small content size and never grew even when the column was widened dramatically
+  //    (tested up to a 3000px window) — useless for detecting "is there now enough room to
+  //    re-expand." `.dock-panel` doesn't have this problem: it's the actual resizable column,
+  //    confirmed (both here and by the earlier hardcoded-threshold version) to track the true
+  //    available width correctly in both directions.
+  const [isIconOnly, setIsIconOnly] = useState(false);
+  useEffect(() => {
+    if (isPowerMode) return undefined;
```
katherinejensen00: **[Finding 1 / High]** This early return correctly skips observer setup in Power mode, but it never calls `setIsIconOnly(false)` before returning. Both modes render `titleWithTooltip` through the same code path that applies the icon-only class, so if `isIconOnly` was `true` in Simple mode it remains permanently `true` after a Simple→Power runtime switch (`useIsPowerMode()` is a live subscription — a mode change re-runs this effect with the stale state in place).

**Fix:** call `setIsIconOnly(false)` before `return undefined`, or add a leading `useEffect([isPowerMode])` that resets it whenever Power mode becomes active.

extensions/src/platform-scripture-editor/src/scripture-text-grid.web-view.tsx
```diff
@@ -315,34 +319,10 @@ globalThis.webViewComponent = function ScriptureTextGridWebView({
...
-    // rc-dock fires no event we can hook from inside the iframe on tab switches, so poll cheaply.
-    const id = window.setInterval(update, 500);
-    return () => window.clearInterval(id);
-  }, []);
-
+  const gridIconUrl = useTabIconSelection(isDarkTheme, TAB_ICON_URLS);
   useEffect(() => {
-    updateWebViewDefinition({ iconUrl: pickTabIconUrl(isDarkTheme, isTabSelected, TAB_ICON_URLS) });
-  }, [isDarkTheme, isTabSelected, updateWebViewDefinition]);
+    updateWebViewDefinition({ iconUrl: gridIconUrl });
```
katherinejensen00, line 324: **[Finding 2 / High]** `updateWebViewDefinition` is called here without an `if (isPowerMode) return` guard. `useTabIconSelection` has no mode awareness and resolves an icon URL in both modes, so this effect stamps an `iconUrl` onto the tab definition in Power mode too.

Both sibling web views (`resource-text-panel`, `comment-list`) guard with `if (isPowerMode) return` before this call.

**Fix:** add `if (isPowerMode) return;` as the first statement of this effect, matching the sibling pattern.

extensions/src/platform-scripture-editor/src/resource-text-panel.web-view.tsx
```diff
@@ -160,6 +164,49 @@ globalThis.webViewComponent = function ResourceTextPanel({
...
+    };
+  }, []);
+
+  const tabIconUrls =
+    resourceType === 'ScriptureResource' ? BIBLE_TEXTS_ICON_URLS : COMMENTARIES_ICON_URLS;
+  const tabIconUrl = useTabIconSelection(isDarkTheme, tabIconUrls);
+  useEffect(() => {
+    // Power mode: no tab icon, exactly as today — do not call updateWebViewDefinition at all so an
+    // already-set iconUrl (there shouldn't be one) is never touched.
+    if (isPowerMode) return;
```
katherinejensen00: **[Finding 3 / High]** This early return is correct for skipping Power-mode icon updates, but when `isPowerMode` transitions `false → true` at runtime the effect re-runs and exits here without clearing the `iconUrl` (and `tooltip`) that a previous Simple-mode run already set.

`updateWebViewDefinition({ iconUrl: undefined })` *would* clear it — the merge function checks `key in updateInfo`, so a present-but-`undefined` value writes through.

**Fix:** call `updateWebViewDefinition({ iconUrl: undefined, tooltip: undefined })` before this `return` to reset the Simple-mode-only fields on a mode switch.

src/renderer/components/docking/platform-tab-title.component.tsx
```diff
@@ -280,9 +294,127 @@ export function PlatformTabTitle({
...
+      const fullContentWidth = Array.from(measureClones).reduce(
+        (total, clone) => total + clone.getBoundingClientRect().width,
+        0,
+      );
+      // The gap the tab bar's own flex container puts between sibling tabs isn't part of any one
+      // tab's own measured width, so it isn't captured by summing the clones above — read it live
+      // from `.dock-nav-list`'s own CSS instead of hardcoding it, so it can't drift out of sync
+      // with that stylesheet.
+      const navList = panel.querySelector('.dock-nav-list');
+      const interTabGapPx = navList ? parseFloat(getComputedStyle(navList).columnGap) || 0 : 0;
```
katherinejensen00, line 398: **[Finding 4 / High]** `parseFloat('normal')` returns `NaN` — `'normal'` is what Chrome returns for `column-gap` when no explicit gap rule is applied to `.dock-nav-list`, and neither stylesheet sets one. `NaN || 0` silently zeroes `interTabGapPx`, so `interTabGapsTotal` is also 0. The icon-only threshold (`fullContentWidth + interTabGapsTotal + TAB_BAR_CHROME_OVERHEAD_PX`) is then understated, potentially preventing tabs from ever collapsing to icon-only even when the column is genuinely too narrow.

**Fix:** `const gap = parseFloat(getComputedStyle(navList).columnGap); const safeGap = Number.isFinite(gap) ? gap : 0;`

src/renderer/components/docking/dock-layout-wrapper.component.tsx
```diff
@@ -21,17 +26,19 @@ export const DockLayoutWrapper = forwardRef(function DockLayoutWrapper(
   { loadTab, saveTab, onLayoutChange, defaultLayout, style }: DockLayoutWrapperProps,
   ref: ForwardedRef<DockLayout> | undefined,
 ) {
+  const isPowerMode = useIsPowerMode();
   return (
     <DockLayout
       ref={ref}
-      groups={GROUPS}
+      groups={getGroups(isPowerMode)}
```
katherinejensen00, line 33: **[Finding 5 / Medium]** `getGroups` returns a new object on every call, including a freshly created `createElement(ChevronsUpDown)` for `moreIcon`. `DockLayout` extends `React.PureComponent` and uses shallow prop comparison, so a new reference always fails the equality check — forcing a `fixLayoutData` O(n layout) traversal on every parent re-render even when `isPowerMode` has not changed.

**Fix:** `groups={useMemo(() => getGroups(isPowerMode), [isPowerMode])}`

extensions/src/platform-scripture-editor/src/main.ts
```diff
@@ -949,12 +960,20 @@ const scriptureTextGridWebViewProvider: IWebViewProvider = {
...
+    const titleLocalizedStrings = await papi.localization.getLocalizedStrings({
+      localizeKeys: [SCRIPTURE_TEXT_GRID_TITLE_KEY],
+    });
     return {
       ...savedWebView,
-      // Icon-only tab: no visible text label, just the "Text Collection" tooltip (the web view keeps
-      // this in sync). The initial empty title avoids flashing a label before the web view runs.
-      title: '',
-      tooltip: '%webView_scriptureTextGrid_title_multiple%',
+      title: SCRIPTURE_TEXT_GRID_TITLE_KEY,
```
katherinejensen00, line 1011: **[Finding 6 / Medium]** The other providers (`ScriptureEditorWebViewFactory`, `createResourceTextPanelProvider`) gate `isClosable` on `interfaceMode === 'power'`. This provider sets `isClosable: false` unconditionally (line 980, just below), so `getTabGroup()` routes the Text Collection tab to `TAB_GROUP_RESOURCES` even in Power mode. But `getGroups(isPowerMode=true)` only registers `TAB_GROUP` — `TAB_GROUP_RESOURCES` is absent — leaving this tab with no registered group config in Power mode.

**Fix:** change `isClosable: false` to `isClosable: interfaceMode === 'power'`, matching the pattern of the other providers.

src/renderer/components/docking/platform-dock-layout-storage.util.ts
```diff
@@ -784,7 +784,7 @@ export function updateWebViewDefinition(
 
   const updatedTabData = createRCDockTabFromTabInfo(
```
katherinejensen00, line 785: **[Finding 7 / Medium]** When `updateWebViewDefinition` is called and `isClosable` has changed (e.g. after a mode switch), `createRCDockTabFromTabInfo` assigns the tab a new `group` string. However, rc-dock's `updateTab` re-inserts the tab into the *same panel* — the panel keeps its original `group`. The tab's `group` field and its containing panel's `group` then disagree. rc-dock's drag-drop validation reads the tab's `group` to decide which panels accept it, so a mismatch here can allow a tab to be dragged into panels it should be locked out of.

src/renderer/components/docking/platform-dock-layout-positioning.util.ts
```diff
@@ -35,20 +39,110 @@ const MAX_FLOAT_HEIGHT_FRACTION = 0.85;
...
+}
+
+/**
+ * WebViewTypes that make up Simple mode's fixed 3-column layout, mapped to the rc-dock group each
+ * is confined to while pinned there. Kept in sync with the webViewTypes hardcoded in
+ * `simple-layout.data.ts` (Columns 1/2) — Column 3 also includes `scriptureTextGrid` (Text
+ * Collection), which isn't part of that static default layout but joins Column 3 at runtime once
+ * enabled.
+ */
+const FIXED_LAYOUT_WEBVIEW_GROUPS: Record<string, string> = {
```
katherinejensen00, line 100: **[Finding 8 / Medium]** Five of the six entries here are bare string literals copied from extension code (`'legacyCommentManager.commentListPanel'`, `'platformScriptureEditor.bibleTexts'`, etc.). Only `SCRIPTURE_EDITOR_WEBVIEW_TYPE` is protected by a shared exported constant. If any extension renames its `webViewType`, this map goes stale silently at runtime — the renamed tab gets `TAB_GROUP` (draggable, no column lock) in Simple mode with no compile-time error.

**Fix:** export typed constants from each extension and import them here, or add drift-guard tests asserting these strings match the extensions' own type declarations.

src/renderer/components/docking/simple-layout.data.ts
```diff
@@ -14,6 +15,8 @@ export const simpleLayout: LayoutBase = {
         size: 1,
         children: [
           {
+            group: HEADLESS_GROUP,
+            panelLock: { minWidth: 300 },
```
katherinejensen00, line 19: **[Finding 9 / Medium]** `panelLock` is a `PlatformDockLayout`-level extension field, not part of rc-dock's own `PanelData` schema. `loadLayout` in `platform-dock-layout-storage.util.ts` calls rc-dock's `loadLayout` verbatim without re-merging these static defaults. If rc-dock's serializer omits `panelLock` during `saveLayout()` (because it is outside rc-dock's own schema), the column minimum-width constraint silently disappears after the first restart.

Worth a quick check: call `dockLayout.saveLayout()` in the console and confirm `panelLock` is present in the output. If absent, re-apply static defaults in the load path.

src/renderer/components/docking/dock-layout-wrapper.component.tsx
```diff
@@ -21,17 +26,19 @@ export const DockLayoutWrapper = forwardRef(function DockLayoutWrapper(
...
       // DockLayout requires LayoutData, but it needs to be LayoutBase in case we use loadTab
       /* eslint-disable no-type-assertion/no-type-assertion */
       defaultLayout={
         (defaultLayout as LayoutData) || { dockbox: { mode: 'horizontal', children: [] } }
       }
       style={style}
-      dropMode="edge"
+      // Simple mode: omit dropMode so tabs can't be dropped onto panel edges to split panels.
+      dropMode={isPowerMode ? 'edge' : undefined}
```
katherinejensen00, line 41: **[Finding 10 / Medium]** The comment says omitting `dropMode` prevents panel-edge drops in Simple mode, which is load-bearing for column isolation. This is an empirical observation — if rc-dock's internal default happens to be `'edge'`, omitting the prop would not help.

If this is the primary mechanism preventing panel splitting in Simple mode, it warrants either an integration test that confirms the behavior or a comment citing the specific rc-dock source being relied upon.

extensions/src/legacy-comment-manager/src/comment-list.web-view.tsx
```diff
@@ -11,16 +11,24 @@ import {
   Sonner,
   sonner,
   usePromise,
+  useTabIconSelection,
```
katherinejensen00, line 14: **[Finding 11 / Medium]** This file imports both `useTabIconSelection` (500ms polling of `frameElement.offsetParent`) and `useViewVisibility` (IntersectionObserver-based) — two overlapping mechanisms for detecting tab visibility.

The `useTabIconSelection` JSDoc says polling is the only option, but `useViewVisibility` shows IntersectionObserver works from inside rc-dock iframes. The signals differ subtly (`offsetParent === null` under `display:none` vs. zero intersection area) but should be equivalent for tab-selected detection. With 4 Simple-mode web views using the polling hook, there are 4 always-on 500ms timers running even on hidden tabs (rc-dock keeps inactive panes mounted).

Worth verifying equivalence and potentially consolidating to one observation mechanism.

src/renderer/components/docking/platform-tab-title.component.tsx
```diff
@@ -280,9 +294,127 @@ export function PlatformTabTitle({
...
+      const interTabGapsTotal = Math.max(0, measureClones.length - 1) * interTabGapPx;
+      const neededWidth = fullContentWidth + interTabGapsTotal + TAB_BAR_CHROME_OVERHEAD_PX;
+      setIsIconOnly(neededWidth > panel.getBoundingClientRect().width);
+    };
+    const resizeObserver = new ResizeObserver(recomputeIsIconOnly);
```
katherinejensen00, line 403: **[Finding 13 / Low]** Each of the (up to 4) Column 3 tabs installs its own `ResizeObserver` + `MutationObserver` on the shared `.dock-panel` ancestor. Every DOM mutation fires all N observers independently, each calling `querySelectorAll`, `getBoundingClientRect`, and `getComputedStyle` — O(N²) DOM reads per event burst.

Currently bounded by the fixed 4-tab max in Column 3, so low priority. A low-friction improvement: debounce `recomputeIsIconOnly` with `requestAnimationFrame` so simultaneous firings collapse to a single DOM read.

src/renderer/components/docking/platform-dock-layout-positioning.util.ts
katherinejensen00, line 100: **[Finding 14 / Low]** Adding a new fixed-column webview to `simple-layout.data.ts` without a corresponding entry here produces no error — the tab silently gets `TAB_GROUP` (draggable, no column lock) in Simple mode. The comment notes the two must be kept in sync manually, but there is no enforcement.

A lightweight test that exports the registered webViewType keys from this map and asserts they match the `webViewType` values in `simple-layout.data.ts` would catch any addition that misses the map.

extensions/src/platform-scripture-editor/src/main.ts
katherinejensen00, line 1011: **[Finding 18 / Scope]** The PT-4208 spec states Text Collection "already relies on being unconditionally icon-only" and that "stays that way — not new behavior introduced by this spec." This line restores a real localized title, so the tab now participates in the icon+title→icon-only collapse rather than being permanently icon-only.

This is arguably better UX consistency. If it was intentionally approved during implementation, noting the spec deviation in the PR description would help reviewers confirm it is expected scope rather than accidental.

lib/platform-bible-react/src/components/advanced/tab-toolbar/tab-toolbar.component.tsx
```diff
@@ -78,16 +78,16 @@ export function TabToolbar({
         />
       )}
       {startAreaChildren && (
-        <div className="tw:flex tw:h-full tw:shrink tw:grow-[10] tw:flex-row tw:flex-wrap tw:items-start tw:gap-x-1 tw:gap-y-2 tw:overflow-clip">
+        <div className="tw:flex tw:shrink tw:grow-[10] tw:flex-row tw:flex-nowrap tw:items-start tw:gap-x-1 tw:gap-y-2 tw:overflow-clip">
```
katherinejensen00, line 81: **[Finding 19 / Scope]** The PT-3958 spec for `TabToolbar` authorized removing `tw:h-full` from the wrapper divs. This change also adds `tw:flex-nowrap` (prevents toolbar items from wrapping) and changes `tw:border` → `tw:border-b` on the outer container (line 70).

Since `TabToolbar` is shared, these additional styling changes affect Power-mode consumers (Checks panel, Enhanced Resources) beyond the Simple-mode scope. If they were intentionally reviewed, noting that they go beyond the PT-3958 spec would help future maintainers understand why the spec and the code differ.

docs/superpowers/plans/2026-07-20-simple-layout-styling-adjustments.md
```diff
@@ -0,0 +1,1049 @@
+# 10Simple Layout Styling Adjustments Implementation Plan
```
katherinejensen00: **[Finding 20 / Scope]** `docs/superpowers/` is gitignored (`.gitignore` line 63), signaling that these are ephemeral planning artifacts not intended for the shared repo history. These files appear to have been force-committed.

If these documents are genuinely useful reference material, consider moving them to a non-gitignored location (e.g., `.context/`) or linking them in the PR description. Otherwise, `git rm --cached docs/superpowers/` before merge would honor the gitignore intent.

katherinejensen00:
**[Finding 12 / Low — file not in diff, so posting as general comment]**

**File:** `src/stories/platform/ten-layout-shared.tsx` — `SIMPLE_GROUPS` constant

`SIMPLE_GROUPS` registers `TAB_GROUP` and `HEADLESS_GROUP` but omits `TAB_GROUP_RESOURCES`. Production `getGroups(false)` registers all three, so the Storybook simple-mode story does not replicate Column 3's group isolation — resource tabs can be dragged across columns in the story, meaning visual regressions in Simple-mode column isolation won't be caught there.

Simplest fix: replace the hand-maintained `SIMPLE_GROUPS`/`POWER_GROUPS` constants with direct calls to `getGroups(isPowerMode)` so the story stays automatically in sync as the group config evolves.

katherinejensen00 (commented):
Last blocker — once this is fixed I can approve.

extensions/src/platform-scripture-editor/src/scripture-text-grid.web-view.tsx
katherinejensen00, line 324: **[Finding 2 / High — last blocker]** The same fix applied to `resource-text-panel` and `comment-list` in the latest batch was not applied here. In Power mode, `useTabIconSelection` now uses `useViewVisibility` (always a concrete `boolean`), so it will return `true` when the Text Collection tab is active and stamp an icon onto the Power-mode tab definition.

```typescript
useEffect(() => {
  if (isPowerMode) {
    updateWebViewDefinition({ iconUrl: undefined });
    return;
  }
  updateWebViewDefinition({ iconUrl: gridIconUrl });
}, [isPowerMode, gridIconUrl, updateWebViewDefinition]);
```

(`isPowerMode` will need to be destructured from the webview props, same as in the sibling components.)

extensions/src/legacy-comment-manager/src/comment-list.web-view.tsx
captaincrazybro: Done.

extensions/src/platform-scripture-editor/src/main.ts
captaincrazybro: I did end up needing to change this a bit...

captaincrazybro: Done.

extensions/src/platform-scripture-editor/src/resource-text-panel.web-view.tsx
captaincrazybro: Done.

extensions/src/platform-scripture-editor/src/scripture-text-grid.web-view.tsx
captaincrazybro: Done.

lib/platform-bible-react/src/components/advanced/tab-toolbar/tab-toolbar.component.tsx
captaincrazybro: It's fine...

src/renderer/components/docking/dock-layout-wrapper.component.tsx
captaincrazybro: Done.

src/renderer/components/docking/dock-layout-wrapper.component.tsx
captaincrazybro: Edge is distinct from the `default` option

src/renderer/components/docking/platform-dock-layout-positioning.util.ts
captaincrazybro: It is VERY unlikely we'll rename the web view types

captaincrazybro: Skip. not necessary

src/renderer/components/docking/platform-dock-layout-storage.util.ts
captaincrazybro: So on switch of interface mode it causes a full layout rebuild, so this concern doesn't matter...

src/renderer/components/docking/platform-tab-title.component.tsx
captaincrazybro: Done.

src/renderer/components/docking/platform-tab-title.component.tsx
captaincrazybro: Not needed...

src/renderer/components/docking/platform-tab-title.component.tsx
captaincrazybro: skip

src/renderer/components/docking/simple-layout.data.ts
captaincrazybro: Already checked...

docs/superpowers/plans/2026-07-20-simple-layout-styling-adjustments.md
captaincrazybro: Done.

captaincrazybro (commented):
> @captaincrazybro made 16 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2603)* status: 0 of 61 files reviewed, 17 unresolved discussions (waiting on katherinejensen00).

extensions/src/platform-scripture-editor/src/scripture-text-grid.web-view.tsx
captaincrazybro: The text collection tab icon should be showed in both simple and power, same with the comments tab.

katherinejensen00 (approved):
Nice work on this, Levi! The live tab bar measurement system is particularly well-engineered. Two optional follow-ups: (1) parseFloat NaN in recomputeIsIconOnly - use Number.isFinite; (2) getGroups without useMemo in dock-layout-wrapper. Neither blocks merge. Approving!

katherinejensen00 (approved):
> @katherinejensen00 reviewed 61 files and all commit messages, and resolved 17 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2603)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

katherinejensen00 (approved):

irahopkinson (approved):
:lgtm: approving to overcome code-owner limitation

> @irahopkinson reviewed 61 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2603#-:-OyaUjpvBO9YBLi7cxja:bm33rhb)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

katherinejensen00 (approved):
> @katherinejensen00 reviewed 8 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2603)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2610, opened 2026-07-25
PT-4179: Add sync progress wizard step to first-run flow

katherinejensen00:
image

image

Adds the fourth and final first-run wizard step: a sync progress screen that shows a spinner/progress bar while syncing and a "Sync complete" screen when done. Finish is disabled until the full sync cycle completes.

## Review-driven deltas (added 2026-07-26)

Two changes from the 2026-07-25 UX review:

**1. Interstitial presentation (PT-4257)**

`syncProgress` is now excluded from the stepper count and suppressed from Back navigation:
- Numbered steps show "Step N of 3" (was "of 4")
- Step indicator is absent on syncProgress
- Back button is absent on syncProgress even when reached via navigation from syncConsent
- Two new exported constants: `INTERSTITIAL_STEPS` (Set) and `NUMBERED_STEPS` (filtered array), both exported for downstream use (e.g. PT-4257 stepper)

**2. Per-project name rows**

Rows populate below the global bar as each project's sync begins and persist into the completion state as a summary:
- Each unique `progressText` from a determinate `onSyncProgress` event creates a row (spinner while syncing, ✓ when the next project arrives)
- Indeterminate events (null/undefined `progressValue`) are ignored — their `progressText` is a localized status message, not a project name
- All rows flip to done when `isSyncing: false`
- `role="status"` narrowed to wrap only the heading/body in completion state (not the row list) — screen readers don't re-read all project names on the heading transition
- Two new Storybook stories: `MultiProject` (GreekNT done, TPTS syncing) and `AllDoneWithProjects` (completion + both rows done)

Note: rows are name-only (no model text, language, or resource counts) — those fields are not available from live S/R events. A richer per-project event would be needed for that detail (deferred).

## Key files

| File | What it does |
|---|---|
| `steps/sync-progress.component.tsx` | New step — subscribes to two S/R network events, manages sync state, renders progress UI |
| `first-run-shell.component.tsx` | `goToStep` now always sets `canProceed=false` (previously hardcoded per step); adds spinner on disabled Finish for the last step; `INTERSTITIAL_STEPS`/`NUMBERED_STEPS` for interstitial treatment |
| `.storybook/papi-stubs/` + `main.ts` | Two new stub files wired via `NormalModuleReplacementPlugin` so startup-wizard stories don't crash on a 10 s WebSocket timeout (no PAPI backend in Storybook) |
| `steps/sync-progress.component.stories.tsx` | 5 stories (Default/MidSync/AllDone/MultiProject/AllDoneWithProjects) using injectable event props + local `PlatformEventEmitter` pairs — no PAPI needed |
| `steps/sync-progress.component.test.tsx` | 18 tests: state transitions, stale-event guard, belt-and-suspenders, recovery timeout, ARIA, listener cleanup, per-project row accumulation |
| `assets/localization/en.json` | New sync-progress localization keys; completion body string tightened (removed hardcoded button-label sentence) |

## Three non-obvious decisions worth verifying

1. **`hasSyncStartedRef` is a ref, not state.** `onSyncStateChanged` is fire-and-forget — a trailing `isSyncing: false` can arrive before or after mount. Without the guard, a lone `false` would immediately enable Finish (sync appeared to be "already done"). Using a ref also keeps the handler identity stable so `useEvent` never re-subscribes on each render.

2. **30-second recovery timeout.** If sync completes *before* this step mounts, no events arrive at all. The timeout enables Finish after 30 s with no events as a fallback. It's a no-op if any event arrived (same ref guards it). Is 30 s the right value? Worth discussing.

3. **Injectable event props (`onSyncStateChangedEvent`, `onSyncProgressEvent`).** Production callers omit them; Storybook and tests inject local emitters. This is why the stories and tests can drive the component to any state without a live backend.

## Please verify

- [ ] **PT-4178 (sync consent)**: does the consent step call `sendReceiveProjects` (which fires `onSyncStateChanged`)? This step depends on that — it won't work with `syncProjects`.
- [ ] **30 s timeout** — weigh in if you have a sense of real-world sync startup latency.
- [ ] **Spinner on disabled Finish** — the shell now shows a spinner for *any* last step while `!canProceed`. Fine for sync (always async), but may surprise a future last step that gates on user action.
- [ ] **Row granularity** — rows accumulate from `progressText` transitions in determinate events. If the S/R extension fires one message per file (not per project), rows may be noisy; if one message for the whole batch, only one row appears. The `MultiProject` Storybook story demonstrates the intended multi-row behavior.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2610)


rolfheij-sil:
🤖 Claude:

### Code review

No issues met the high-confidence bar, but several verified lower-confidence concerns are worth a look:

1. The new e2e suite invalidates its own precondition when run: `preConfigureSettings({ 'platform.firstRunComplete': false })` runs once in `beforeAll`, but the isolated fixture launches a fresh Electron per test against the same shared `dev-appdata/data/settings.json`. The "skip setup" test completes the first run and writes `firstRunComplete: true` back to that file, so the following "Sync progress" test's `waitForWizard()` times out deterministically. This is latent today only because CI runs `test:e2e:smoke`, not `tests/isolated/`. `comment.fixture.ts`'s worker-scoped pattern (restore only after the app closes) guards against exactly this hazard.

https://github.com/paranext/paranext-core/blob/716be0505436eec1c2aaa521c1674cb3af1cb71b/e2e-tests/tests/isolated/first-run/first-run.spec.ts#L21-L31

2. Cross-PR contract mismatch with #2589: the comment here says the sync-consent step calls `paratextBibleSendReceive.sendReceiveProjects` before advancing, but #2589's real sync-consent step calls `syncProjects` and only advances after the sync has completed. If that holds post-integration, the events this step is built around may never arrive while it is mounted, leaving the 30-second recovery timeout as the only path that enables Finish. Worth reconciling the two PRs' handoff (and the command name in the comment) explicitly.

https://github.com/paranext/paranext-core/blob/716be0505436eec1c2aaa521c1674cb3af1cb71b/src/renderer/components/first-run/steps/sync-progress.component.tsx#L81-L85

3. When sync completes, Finish flips from disabled to enabled but focus is never moved to it. On an interstitial with no other controls (Back is suppressed) and an automatic transition, keyboard and screen-reader users are not told the only actionable control just became available. Same shape as the overlay retry-button ref-focus fix from the #2576 review.

https://github.com/paranext/paranext-core/blob/716be0505436eec1c2aaa521c1674cb3af1cb71b/src/renderer/components/first-run/steps/sync-progress.component.tsx#L110-L115

4. The shell comment says interstitials are covered by "the step's own heading and role=status", but in the syncing state this step renders a plain heading with no live region — `role="status"` only wraps the completion branch — so entering the step announces nothing to screen readers.

https://github.com/paranext/paranext-core/blob/716be0505436eec1c2aaa521c1674cb3af1cb71b/src/renderer/components/first-run/first-run-shell.component.tsx#L125-L128

5. Row accumulation de-duplicates only against the immediately previous project name (`lastProjectNameRef`), so a non-consecutively recurring `progressText` (retry or interleaved progress) appends a second row with the same `key={row.name}` — duplicate React keys and unreliable reconciliation. The existing test covers only consecutive duplicates.

https://github.com/paranext/paranext-core/blob/716be0505436eec1c2aaa521c1674cb3af1cb71b/src/renderer/components/first-run/steps/sync-progress.component.tsx#L152-L164

6. `first-run-step-props.model.ts`'s module doc still describes the old default-enabled model ("if it needs to gate progress, calls `setCanProceed(false)`") while this PR flips the shell to reset `canProceed` to false on every navigation. The property-level doc was updated but the top-of-file doc now contradicts it — a sibling-step author reading only the top doc would ship a step whose Next never enables.

https://github.com/paranext/paranext-core/blob/716be0505436eec1c2aaa521c1674cb3af1cb71b/src/renderer/components/first-run/first-run-step-props.model.ts#L1-L6

Minor notes that did not make the cut: the "Sibling tickets (PT-4176/77/78/79)" note in the same file is now partially stale (this PR is PT-4179); and the identify-placeholder follow-up note was trimmed from "PT-4177 MUST add an end-to-end test... (no re-show/loop)" to vaguer wording pointing at a unit test as the model — the load-bearing part of that note was the E2E requirement and the relaunch-loop failure mode.

jolierabideau (changes requested):
> @jolierabideau reviewed 27 files and all commit messages, and made 6 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2610)* status: all files reviewed, 6 unresolved discussions (waiting on katherinejensen00).

___
*[`assets/localization/en.json` line 41 at r1](https://reviewable.io/reviews/paranext/paranext-core/2610#-OyZGZ3dBXP0EwV9ogOu:-OyZGZ3dBXP0EwV9ogOv:b-e9oapz) ([raw file](https://github.com/paranext/paranext-core/blob/716be0505436eec1c2aaa521c1674cb3af1cb71b/assets/localization/en.json#L41)):*
> ```JSON
>   "%firstRun_step_syncProgress_body%": "Setting up your projects. This may take a few minutes.",
>   "%firstRun_step_syncProgress_complete_body%": "Your projects are ready.",
> ```

From Claude

**Microcopy vocabulary is inconsistent within the step.** In-progress heading is _"Syncing your **data**"_ while the body says _"Setting up your **projects**"_ and completion says _"Your **projects** are ready."_ ([en.json](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/assets/localization/en.json)). The Vocabulary guideline establishes "project" as the term for a user's work; "your data" reads as a different subject. Suggest aligning the heading, e.g. _"Syncing your projects."
___
*[`lib/platform-bible-react/src/stories/guidelines/applying-changes.mdx` line 44 at r1](https://reviewable.io/reviews/paranext/paranext-core/2610#-OyZH4GVE-ietyv-EN-m:-OyZH4GVE-ietyv-EN-n:b-blb5yg) ([raw file](https://github.com/paranext/paranext-core/blob/716be0505436eec1c2aaa521c1674cb3af1cb71b/lib/platform-bible-react/src/stories/guidelines/applying-changes.mdx#L44)):*
> ```MDX
> ### Primary models
> 
> | Model                           | What it means                                                                                                                                                                         | Use it when                                                                                                                | Example                                                      |
> ```

I saw these on one of the other PRs too, are these format changes in the 6 mdx files deliberate or Claude? 
___
*[`src/renderer/components/first-run/steps/sync-progress.component.tsx` line 58 at r1](https://reviewable.io/reviews/paranext/paranext-core/2610#-OyZGS2T-RTAybnKppog:-OyZGS2T-RTAybnKppoh:b782t3r) ([raw file](https://github.com/paranext/paranext-core/blob/716be0505436eec1c2aaa521c1674cb3af1cb71b/src/renderer/components/first-run/steps/sync-progress.component.tsx#L58)):*
> ```XML
>           ) : (
>             // Decorative: row text provides the accessible label.
>             <span aria-hidden="true">✓</span>
> ```

Should this be a Lucide react icon?
___
*[`src/renderer/components/first-run/steps/sync-progress.component.tsx` line 119 at r1](https://reviewable.io/reviews/paranext/paranext-core/2610#-OyZGKYf1evPCuMq5hC8:-OyZGKYf1evPCuMq5hC9:b-13unht) ([raw file](https://github.com/paranext/paranext-core/blob/716be0505436eec1c2aaa521c1674cb3af1cb71b/src/renderer/components/first-run/steps/sync-progress.component.tsx#L119)):*
> ```XML
>   // step mounted and we missed the completion event. Enable Finish so the user is never stuck.
>   useEffect(() => {
>     const id = setTimeout(() => {
> ```

From Claude

**30s recovery timeout can enable Finish prematurely on a slow-starting sync.** [sync-progress.component.tsx:118-123](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/first-run/steps/sync-progress.component.tsx#L118-L123) arms the timeout at mount and only guards on "no event yet" (`!hasSyncStartedRef.current`). But the body copy itself says _"This may take a few minutes"_ — if the **first** S/R event legitimately takes longer than 30s to arrive (connection/network latency before the first `onSyncProgress`/`onSyncStateChanged`), the timeout fires and enables Finish while sync hasn't actually run. This is a _different_ mechanism from the bot's #2 (which is about events never arriving at all because consent may call `syncProjects`). Consider (a) resetting the timeout on each received event so it only covers the genuinely-quiet case, or (b) a longer window. Corollary edge: once _any_ event is seen, there's no secondary fallback — a dropped terminal `isSyncing:false` after start leaves Finish disabled indefinitely. Both are documented tradeoffs, so this is a "worth a second look" not a blocker.
___
*[`src/renderer/components/first-run/steps/sync-progress.component.tsx` line 164 at r1](https://reviewable.io/reviews/paranext/paranext-core/2610#-OyZGuuk3uBhJXW8ioE4:-OyZGuuk3uBhJXW8ioE5:bn4k827) ([raw file](https://github.com/paranext/paranext-core/blob/716be0505436eec1c2aaa521c1674cb3af1cb71b/src/renderer/components/first-run/steps/sync-progress.component.tsx#L164)):*
> ```XML
>         lastProjectNameRef.current = text;
>         setRows((prev) => [
>           ...prev.map((r) => (r.status === 'syncing' ? { ...r, status: 'done' as const } : r)),
> ```

NIT From Claude

**Inconsistent loop-variable naming in the same file.** [sync-progress.component.tsx:145](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/first-run/steps/sync-progress.component.tsx#L145) and [:164](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/first-run/steps/sync-progress.component.tsx#L164) use single-letter `r` for a `ProjectRow`, while `ProjectRowList` at [:51](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/first-run/steps/sync-progress.component.tsx#L51) uses the clearer `row` for the same type. Rename `r` → `row`.
___
*[`src/stories/keyboard-shortcuts.mdx` line 15 at r1](https://reviewable.io/reviews/paranext/paranext-core/2610#-OyZHMoT3hWXBmCm4-xY:-OyZHMoT3hWXBmCm4-xZ:b-u3gkyo) ([raw file](https://github.com/paranext/paranext-core/blob/716be0505436eec1c2aaa521c1674cb3af1cb71b/src/stories/keyboard-shortcuts.mdx#L15)):*
> ```MDX
> each is handled in code.
> 
> For how to *display* a shortcut in the UI (per-OS formatting conventions), see the
> ```

Is this a deliberate change?

katherinejensen00:
Addressed in f40318a. Details per item:

**1 (e2e isolation)** — Fixed: `beforeAll`/`afterAll` → `beforeEach`/`afterEach`. The "skip setup" test calls `completeFirstRun()` which writes `firstRunComplete: true` to the shared settings file; subsequent tests in the same suite need a fresh reset before each Electron launch.

**2 (cross-PR handoff)** — Updated the JSDoc on `SyncProgressStep` to document both scenarios: if sync-consent fires `sendReceiveProjects` (non-blocking) before advancing, events arrive here; if it calls `syncProjects` and waits for completion before advancing, no events arrive and the 30 s timeout is the only path to Finish. Added a note that PT-4178 must reconcile the exact handoff. I can't change the logic without knowing #2589's actual behavior — leaving it for the integration pass.

**3 (focus on Finish)** — Added `finishButtonRef` and a `useEffect` in the shell that watches `[canProceed, isLastStep]` and calls `finishButtonRef.current?.focus()` on the `false → true` transition on the last step. Same shape as the #2576 overlay retry-button fix.

**4 (missing live region in syncing state)** — Wrapped the syncing-state heading and body in `<div role="status">`, mirroring the completion branch. Progress indicator stays outside (changes too frequently to announce). Added a test verifying the syncing state has `role="status"` with the heading text.

**5 (row dedup)** — Removed `lastProjectNameRef`; the guard moved inside `setRows`'s functional update as `if (prev.some(r => r.name === text)) return prev`. This handles both consecutive and non-consecutive recurrences. Added a test for the non-consecutive case (GreekNT → TPTS → GreekNT should stay 2 rows, not 3).

**6 (module doc)** — Updated the top-level doc in `first-run-step-props.model.ts` to describe the default-disabled model: navigation always resets to disabled; steps that want Next/Finish enabled on mount call `setCanProceed(true)` in their own mount effect.

katherinejensen00 (commented):
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2610)* status: 19 of 28 files reviewed, 6 unresolved discussions (waiting on jolierabideau).

___
*[`assets/localization/en.json` line 41 at r1](https://reviewable.io/reviews/paranext/paranext-core/2610#-OyZGZ3dBXP0EwV9ogOu:-Oy_qGNG-TON_pckBD6b:b-c65phn) ([raw file](https://github.com/paranext/paranext-core/blob/716be0505436eec1c2aaa521c1674cb3af1cb71b/assets/localization/en.json#L41)):*

From Claude

**Microcopy vocabulary is inconsistent within the step.** In-progress heading is _"Syncing your **data**"_ while the body says _"Setting up your **projects**"_ and completion says _"Your **projects** are ready."_ ([en.json](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/assets/localization/en.json)). The Vocabulary guideline establishes "project" as the term for a user's work; "your data" reads as a different subject. Suggest aligning the heading, e.g. _"Syncing your projects."
</blockquote></details>

Good catch — "projects" is the right term throughout. Updated `%firstRun_step_syncProgress_heading%` to "Syncing your projects."
___
*[`lib/platform-bible-react/src/stories/guidelines/applying-changes.mdx` line 44 at r1](https://reviewable.io/reviews/paranext/paranext-core/2610#-OyZH4GVE-ietyv-EN-m:-Oy_qPGrA6j4jgMWDdKz:b-grwfno) ([raw file](https://github.com/paranext/paranext-core/blob/716be0505436eec1c2aaa521c1674cb3af1cb71b/lib/platform-bible-react/src/stories/guidelines/applying-changes.mdx#L44)):*

I saw these on one of the other PRs too, are these format changes in the 6 mdx files deliberate or Claude?
</blockquote></details>

Claude says "These are Prettier normalization from the `npm run format` pass — it converts `_italic_` to `*italic*` and adjusts Markdown table column widths to align cells. No content changes. I'll add a note to the commit if that helps reviewers skip them in future diffs."
___
*[`src/renderer/components/first-run/steps/sync-progress.component.tsx` line 58 at r1](https://reviewable.io/reviews/paranext/paranext-core/2610#-OyZGS2T-RTAybnKppog:-Oy_qVZc22MC8cnUCoSU:bsv6pqp) ([raw file](https://github.com/paranext/paranext-core/blob/716be0505436eec1c2aaa521c1674cb3af1cb71b/src/renderer/components/first-run/steps/sync-progress.component.tsx#L58)):*

Should this be a Lucide react icon?
</blockquote></details>

Good point! Yes, it should be. Fixed.
___
*[`src/renderer/components/first-run/steps/sync-progress.component.tsx` line 119 at r1](https://reviewable.io/reviews/paranext/paranext-core/2610#-OyZGKYf1evPCuMq5hC8:-Oy_q_ak2IliwSJNGy10:b-gf446p) ([raw file](https://github.com/paranext/paranext-core/blob/716be0505436eec1c2aaa521c1674cb3af1cb71b/src/renderer/components/first-run/steps/sync-progress.component.tsx#L119)):*

From Claude

**30s recovery timeout can enable Finish prematurely on a slow-starting sync.** [sync-progress.component.tsx:118-123](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/first-run/steps/sync-progress.component.tsx#L118-L123) arms the timeout at mount and only guards on "no event yet" (`!hasSyncStartedRef.current`). But the body copy itself says _"This may take a few minutes"_ — if the **first** S/R event legitimately takes longer than 30s to arrive (connection/network latency before the first `onSyncProgress`/`onSyncStateChanged`), the timeout fires and enables Finish while sync hasn't actually run. This is a _different_ mechanism from the bot's #2 (which is about events never arriving at all because consent may call `syncProjects`). Consider (a) resetting the timeout on each received event so it only covers the genuinely-quiet case, or (b) a longer window. Corollary edge: once _any_ event is seen, there's no secondary fallback — a dropped terminal `isSyncing:false` after start leaves Finish disabled indefinitely. Both are documented tradeoffs, so this is a "worth a second look" not a blocker.
</blockquote></details>

Both tradeoffs are real. I've documented them in the JSDoc at the top of `SyncProgressStep` (handoff uncertainty + timeout-as-recovery). I'm inclined to leave the 30s as-is for now: the exact sync handoff is still unresolved pending PT-4178, and resetting the timeout on each event (option a) would add meaningful complexity against an uncertain spec. If PT-4178's integration pass shows a slow-start pattern in practice, we can extend the window or adopt option (a) there. Agreed that the secondary fallback gap (dropped terminal `isSyncing: false`) is a known hole — also noted in the JSDoc.
___
*[`src/renderer/components/first-run/steps/sync-progress.component.tsx` line 164 at r1](https://reviewable.io/reviews/paranext/paranext-core/2610#-OyZGuuk3uBhJXW8ioE4:-Oy_qgJpECZfCVA8YNlv:bt2vr1b) ([raw file](https://github.com/paranext/paranext-core/blob/716be0505436eec1c2aaa521c1674cb3af1cb71b/src/renderer/components/first-run/steps/sync-progress.component.tsx#L164)):*

NIT From Claude

**Inconsistent loop-variable naming in the same file.** [sync-progress.component.tsx:145](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/first-run/steps/sync-progress.component.tsx#L145) and [:164](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/first-run/steps/sync-progress.component.tsx#L164) use single-letter `r` for a `ProjectRow`, while `ProjectRowList` at [:51](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/first-run/steps/sync-progress.component.tsx#L51) uses the clearer `row` for the same type. Rename `r` → `row`.
</blockquote></details>

Fixed.
___
*[`src/stories/keyboard-shortcuts.mdx` line 15 at r1](https://reviewable.io/reviews/paranext/paranext-core/2610#-OyZHMoT3hWXBmCm4-xY:-Oy_qhMzDVcHlp5kPbK9:b-udjm1) ([raw file](https://github.com/paranext/paranext-core/blob/716be0505436eec1c2aaa521c1674cb3af1cb71b/src/stories/keyboard-shortcuts.mdx#L15)):*

Is this a deliberate change?
</blockquote></details>

Similar to the other MDX files — Prettier normalized `_display_` to `*display*` during the format pass. No content change.

jolierabideau (approved):
> @jolierabideau reviewed 10 files and all commit messages, and resolved 6 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2610)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2613, opened 2026-07-27
Don't let a startup settings-fetch timeout crash the whole data provider

imnasnainaec:
## Issue

A real dev-mode log showed `ParanextDataProvider` (the .NET backend) crashing shortly after startup:

```
System.TimeoutException: Task "SettingService.GetSetting for platform.paratextDataLastRegistryDataCachedTimes" did not complete quickly enough.
   at Paranext.DataProvider.ThreadingUtils.GetTaskResult[T](...)
   at Paranext.DataProvider.Services.SettingsService.GetSetting[T](...)
   at Paranext.DataProvider.Services.PersistedParatextDataSettings..ctor(PapiClient papiClient)
   at Paranext.DataProvider.Program.Main()
```

`PersistedParatextDataSettings.LastRegistryDataCachedTimes` and `PersistedPtxUtilsSettings.MementoData` each call `SettingsService.GetSetting` directly in a property initializer, with no try/catch. A timeout there (this call races the rest of Electron/webpack dev startup, using the hardcoded `ThreadingUtils.DefaultTimeout` rather than any configured timeout) propagates straight out of `Program.Main()` uncaught, killing the whole process. `dotnet watch` only restarts on a file change, not on an unexpected exit, so once this happens the backend is dead for the rest of the session.

## Solution

- Wrap the `GetSetting` calls in both classes' property initializers in the same catch-log-and-fall-back-to-empty pattern `SettingsService.Initialize` already uses around its own `GetSetting` call, via a new shared `SettingsService.TryGetSetting` helper (also used by both classes, replacing near-duplicate private helpers). `TryGetSetting` treats a legitimate null result as success — only an exception counts as failure.
- Guard `SafeSave()` against overwriting real persisted data with the empty fallback: while the load is unconfirmed, `SafeSave` re-checks reachability (not a one-time flag, so a transient startup-race timeout doesn't permanently block saving for the rest of the session) and merges the confirming read's value in as the base — in-memory changes win on key conflicts — instead of discarding it.
- Note: since the fallback starts empty, this merge can resurrect a persisted entry the app never loaded into memory (so a deliberate "drop this entry" can't be distinguished from "never saw it") — an accepted tradeoff, since the alternatives are crashing or clobbering everything.
- Regression tests for both classes and the shared helper, using the existing `DummyPapiClient`/`DummySettingsService` test harness (which needed a small fix itself: its "set" handler returned a raw `bool` instead of the `JsonElement` wire response `SetSetting` expects — never previously exercised by any test).

---

Drafted with Claude Sonnet 5.

---

Devin review: https://app.devin.ai/review/paranext/paranext-core/pull/2613

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2613)

### DIFF
--- a/c-sharp/Services/PersistedParatextDataSettings.cs
+++ b/c-sharp/Services/PersistedParatextDataSettings.cs
@@ -7,10 +7,8 @@ internal class PersistedParatextDataSettings : IParatextDataSettings
 {
     private readonly PapiClient _papiClient;
 
-    // True until the settings service is confirmed reachable, so SafeSave doesn't overwrite the
-    // previously-persisted value with this instance's fallback-to-empty default. Re-checked (not
-    // permanent) so a transient startup-race timeout doesn't block saving for the rest of the
-    // session once the service comes up.
+    // SetSetting is a blind overwrite, so SafeSave must not save the fallback-to-empty default over
+    // real data. Re-checked, not latched: one startup timeout shouldn't kill saving the session.
     private bool _loadUnconfirmed;
 
     public PersistedParatextDataSettings(PapiClient papiClient)
@@ -40,7 +38,7 @@ internal class PersistedParatextDataSettings : IParatextDataSettings
                 return;
             _loadUnconfirmed = false;
 
-            // The confirming read may have found entries from before this session that our
+            // The confirming read may have found entries from before this session that the
             // fallback-to-empty default never saw. Keep them, but let in-memory changes win.
             if (persistedValue != null)
                 foreach (var (key, value) in persistedValue)
--- a/c-sharp/Services/PersistedParatextDataSettings.cs
+++ b/c-sharp/Services/PersistedParatextDataSettings.cs
@@ -3,37 +3,34 @@ using PtxUtils;
 
 namespace Paranext.DataProvider.Services;
 
-internal class PersistedParatextDataSettings(PapiClient papiClient) : IParatextDataSettings
+internal class PersistedParatextDataSettings : IParatextDataSettings
 {
-    public SerializableStringDictionary LastRegistryDataCachedTimes { get; set; } =
-        GetLastRegistryDataCachedTimes(papiClient);
+    private readonly PapiClient _papiClient;
 
-    // A timeout here must not crash the whole data provider at startup, so this mirrors the
-    // try/catch SettingsService.Initialize already uses around its own GetSetting call.
-    private static SerializableStringDictionary GetLastRegistryDataCachedTimes(
-        PapiClient papiClient
-    )
+    // Set when the initial load fails (e.g. times out), so SafeSave doesn't overwrite the
+    // previously-persisted value with this instance's fallback-to-empty default.
+    private readonly bool _loadFailed;
+
+    public PersistedParatextDataSettings(PapiClient papiClient)
     {
-        try
-        {
-            return SettingsService.GetSetting<SerializableStringDictionary>(
-                    papiClient,
-                    Settings.PARATEXT_DATA_LAST_REGISTRY_DATA_CACHED_TIMES
-                ) ?? [];
-        }
-        catch (Exception ex)
-        {
-            Console.WriteLine(
-                $"Error getting {Settings.PARATEXT_DATA_LAST_REGISTRY_DATA_CACHED_TIMES}: {ex}"
-            );
-            return [];
-        }
+        _papiClient = papiClient;
+        _loadFailed = !SettingsService.TryGetSetting(
+            papiClient,
+            Settings.PARATEXT_DATA_LAST_REGISTRY_DATA_CACHED_TIMES,
+            out SerializableStringDictionary? value
+        );
+        LastRegistryDataCachedTimes = value ?? [];
     }
+
+    public SerializableStringDictionary LastRegistryDataCachedTimes { get; set; }
 
     public void SafeSave()
     {
+        if (_loadFailed)
+            return;
+
         SettingsService.SetSetting(
-            papiClient,
+            _papiClient,
             Settings.PARATEXT_DATA_LAST_REGISTRY_DATA_CACHED_TIMES,
             LastRegistryDataCachedTimes
         );
--- a/c-sharp/Services/PersistedParatextDataSettings.cs
+++ b/c-sharp/Services/PersistedParatextDataSettings.cs
@@ -6,10 +6,29 @@ namespace Paranext.DataProvider.Services;
 internal class PersistedParatextDataSettings(PapiClient papiClient) : IParatextDataSettings
 {
     public SerializableStringDictionary LastRegistryDataCachedTimes { get; set; } =
-        SettingsService.GetSetting<SerializableStringDictionary>(
-            papiClient,
-            Settings.PARATEXT_DATA_LAST_REGISTRY_DATA_CACHED_TIMES
-        ) ?? [];
+        GetLastRegistryDataCachedTimes(papiClient);
+
+    // A timeout here must not crash the whole data provider at startup, so this mirrors the
+    // try/catch SettingsService.Initialize already uses around its own GetSetting call.
+    private static SerializableStringDictionary GetLastRegistryDataCachedTimes(
+        PapiClient papiClient
+    )
+    {
+        try
+        {
+            return SettingsService.GetSetting<SerializableStringDictionary>(
+                    papiClient,
+                    Settings.PARATEXT_DATA_LAST_REGISTRY_DATA_CACHED_TIMES
+                ) ?? [];
+        }
+        catch (Exception ex)
+        {
+            Console.WriteLine(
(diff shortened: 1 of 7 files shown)


c-sharp/Services/PersistedParatextDataSettings.cs
```diff
@@ -6,10 +6,29 @@ namespace Paranext.DataProvider.Services;
...
+                    papiClient,
+                    Settings.PARATEXT_DATA_LAST_REGISTRY_DATA_CACHED_TIMES
+                ) ?? [];
+        }
+        catch (Exception ex)
+        {
+            Console.WriteLine(
+                $"Error getting {Settings.PARATEXT_DATA_LAST_REGISTRY_DATA_CACHED_TIMES}: {ex}"
+            );
+            return [];
```
imnasnainaec: 🤖 **Claude AI** · **[Contracts]** **[Architecture]** · Decision needed (warning)

On a transient settings-fetch timeout, should `SafeSave()` be suppressed (e.g. via a "loaded successfully" flag) so it can't overwrite the previously-persisted `LastRegistryDataCachedTimes` with the empty fallback — or is a redundant registry re-cache an acceptable trade-off for this key?

The try/catch now falls back to `[]` instead of crashing (good), but `SafeSave()` (lines 33-40) later persists the in-memory `LastRegistryDataCachedTimes` back via `SettingsService.SetSetting`. So a save that occurs after a failed load overwrites the stored registry-cache times with the empty fallback — a silent load-failure→clobber path that did not exist before (the timeout previously crashed startup). There is no flag distinguishing "genuinely empty" from "load failed".

For *this* key the cost is benign (a redundant registry re-cache). Flagging it mainly because the sibling `MementoData` shares the exact pattern with higher stakes, and because the `SettingsService.Initialize` precedent this mirrors does **not** have the problem — its fallback value is never persisted.

`SafeSave` is invoked by the external ParatextData framework via `IParatextDataSettings`, so this code does not control its timing relative to the failed load.

_Independently reported by the Contracts and Architecture perspectives._
</details>

c-sharp/Services/PersistedPtxUtilsSettings.cs
```diff
@@ -4,11 +4,25 @@ namespace Paranext.DataProvider.Services;
...
+        {
+            return SettingsService.GetSetting<SerializableStringDictionary>(
+                    papiClient,
+                    Settings.PTX_UTILS_MEMENTO_DATA
+                ) ?? [];
+        }
+        catch (Exception ex)
+        {
+            Console.WriteLine($"Error getting {Settings.PTX_UTILS_MEMENTO_DATA}: {ex}");
+            return [];
```
imnasnainaec: 🤖 **Claude AI** · **[Contracts]** **[Architecture]** · Decision needed (warning)

Same fallback→`SafeSave` interaction as `PersistedParatextDataSettings`, but higher-stakes here: is resetting the user's stored mementos on a transient startup timeout acceptable, or should `SafeSave()` suppress the write when the initial load failed?

`MementoData` holds PtxUtils mementos (e.g. internet settings, form state) rather than regenerable cache timestamps. After a transient startup timeout the in-memory `MementoData` falls back to `[]`, and the framework-invoked `SafeSave()` (lines 39-42) will then persist `[]` over the user's previously-stored mementos, resetting them. `SafeSave` has no way to tell "the setting really was empty" from "the read timed out".

The empty value itself is contract-legal (the pre-existing `?? []` null path already produced it) — the concern is the *save* behavior after a load failure. Keep the resolution consistent with whatever is decided for `PersistedParatextDataSettings`.

`SafeSave` is invoked by the external PtxUtils framework via `IPtxUtilsSettings`, not by this codebase, so its timing relative to the failed load is not controlled here.

_Independently reported by the Contracts and Architecture perspectives._
</details>

c-sharp/Services/PersistedParatextDataSettings.cs
```diff
@@ -6,10 +6,29 @@ namespace Paranext.DataProvider.Services;
 internal class PersistedParatextDataSettings(PapiClient papiClient) : IParatextDataSettings
 {
     public SerializableStringDictionary LastRegistryDataCachedTimes { get; set; } =
-        SettingsService.GetSetting<SerializableStringDictionary>(
-            papiClient,
-            Settings.PARATEXT_DATA_LAST_REGISTRY_DATA_CACHED_TIMES
-        ) ?? [];
+        GetLastRegistryDataCachedTimes(papiClient);
+
+    // A timeout here must not crash the whole data provider at startup, so this mirrors the
```
imnasnainaec: 🤖 **Claude AI** · **[Clarity]** · Decision needed (suggestion)

Is the near-identical try/catch helper + duplicated comment across the two settings files acceptable, or should it be factored into a shared `SettingsService` helper?

This private static "get setting, fall back to empty on exception" helper — including the word-for-word identical two-line WHY comment — is duplicated in `PersistedPtxUtilsSettings.GetMementoData`. Only the setting key and property name differ. A shared helper (e.g. `SettingsService.GetSettingOrEmpty<T>(papiClient, key)` that logs and returns `[]`) would keep the rationale in one place and keep both settings classes in sync if the fallback behavior ever changes.
</details>

imnasnainaec (commented):
## 🤖 [Claude AI Review] — PR #2613

Multi-perspective automated review. Panel: security, scope, contracts, tests, architecture, clarity (`ux` auto-dropped — no TS/React changes). **No blocking findings** — posted as a comment.

| Security     | 0        | 0       | 0          |
| Scope        | 0        | 0       | 0          |
| Contracts    | 0        | 2       | 0          |
| Tests        | 0        | 0       | 3          |
| Architecture | 0        | 2       | 0          |
| Clarity      | 0        | 0       | 6          |
| **Total**    | **0**    | **4**   | **9**      |

The 4 warnings are Contracts + Architecture independently converging on the **same** issue in each source file → **2 unique warning findings**, both inline below. Security and Scope returned clean.

### Confirmed clean

- **Security** — imports standard (none blocked); settings read via PAPI; no secrets (only setting-key names + log text as literals); no injection. Exception text is logged but setting *values* are not, so no payload leakage. Fallback-to-empty is a safe default.
- **Scope** — both source diffs add only the try/catch wrapper; `SafeSave`/`UpgradeNeeded`/`EnableFormSnapping` untouched. The happy-path `Constructor_SettingPresent_ReturnsStoredValue` tests legitimately guard that wrapping didn't break the normal read path. No scope creep, no new dependencies.

### Informational (no inline comment — no action likely needed)

🤖 **Claude AI** · **[Tests]** **[Clarity]** (suggestion)

Both test files use `[TestCase()]` on parameterless tests; the codebase standard is `[Test]` (1384 uses across 103 files; only 3 files use `[TestCase()]`). Non-idiomatic, no behavioral impact.

🤖 **Claude AI** · **[Tests]** (suggestion)

The fail-path test (`PersistedParatextDataSettingsTests.cs:34`, and its sibling) was **verified falsifiable** — it genuinely exercises the catch path today: `DummySettingsService` throws on a missing key → faulted task → `ThreadingUtils.GetTaskResult` rethrows → the source `catch` runs. So it is a valid regression guard. Mild robustness gap: the assertions (`DoesNotThrow` + `Is.Empty`) can't distinguish "catch swallowed a throw" from "`GetSetting` returned null and `?? []` coalesced". If the harness ever returned null instead of throwing, the test would keep passing even with the `catch` removed. Optional hardening: make the fake throw a distinctive exception and assert it was reached.

🤖 **Claude AI** · **[Clarity]** (suggestion)

The `settings` local is declared `null`, assigned inside the `Assert.DoesNotThrow` lambda, then dereferenced with `settings!`. Constructing directly (`var settings = new ...(_client)`) and asserting on the result — a throw already fails the test — removes the `null`/`!` dance. Applies to both test files, line 34.

🤖 **Claude AI** · **[Clarity]** (suggestion)

The `_client`/`_settingsService` fields plus SetUp/TearDown are byte-for-byte identical across both new test files; a shared base fixture would remove the boilerplate. Acceptable to leave for test independence.

---

<sub>Generated by `/review-pr`. Inline comments below lead with the decision needed. Run `/triage-feedback 2613` to consolidate reviewer feedback.</sub>

imnasnainaec (commented):
> @imnasnainaec reviewed all commit messages and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2613)* status: 0 of 7 files reviewed, all discussions resolved.

imnasnainaec:
Devin chat take-away to consider:

> The PR trades a crash for graceful degradation without true retry. For the registry-cache key that's clearly fine. For mementos it's the higher-stakes case — no re-read, and (at head) the confirming read is discarded rather than merged, so persisted-but-unloaded mementos can be dropped on the next save.
>
> If genuine recovery were wanted, the fix is small and already half-present: use the value the SafeSave reachability read already fetches (merge it in, or reload into memory) instead of discarding it.

PR #2617, opened 2026-07-27
PT-4219: e2e smoke test — first-run wizard overlay renders on fresh profile

katherinejensen00:
Adds an isolated e2e smoke test that verifies the first-run wizard overlay renders and gates app access for a fresh Simple-mode user (PT-4175).

**One file changed:** `e2e-tests/tests/isolated/first-run-wizard.spec.ts` (new, ~80 lines). No production code changes.

---

## What the test does

Launches Electron with a fresh ephemeral user-data dir (no localStorage). Waits for the app to be ready, then asserts a `role="dialog"` is visible. The test accepts any of three overlay states — loading spinner, wizard, or registration-error — because all three confirm the gate is rendering and blocking app access.

## Key design decisions worth a second look

**Why `isolated.fixture` instead of `app.fixture`**
`app.fixture` pre-seeds `platform.firstRunComplete: true` to skip the wizard for every other test. This test deliberately needs the wizard to show, so it uses `isolated.fixture` (one Electron per test, fresh temp user-data dir).

**Why `preConfigureSettings` with explicit `firstRunComplete: false`**
The settings seed writes to the shared `dev-appdata/data/settings.json` file *before* Electron launches (Playwright runs `beforeEach` before test-scoped fixture setup). `firstRunComplete: false` is explicit so a developer who has already run through the wizard on their machine doesn't silently suppress the gate. `interfaceMode: 'simple'` prevents the power-mode bypass path in `resolveInternal`.

**Why `beforeEach`/`afterEach` and not `beforeAll`/`afterAll`**
The `electronApp` fixture is test-scoped (one Electron per `test()` call). `beforeEach`/`afterEach` match that scope: each test gets a clean write and a guaranteed restore. A throw in `beforeEach` propagates as a visible test failure rather than silently leaving dev-appdata dirty.

**Known limitation**
On a developer machine with a running Paratext installation, `resolveInternal` may complete quickly (registration succeeds), call `completeThenShowApp`, and dismiss the overlay before the assertion. This does not affect CI, where there is no live Paratext install.

## Test plan

- [ ] `npm run test:e2e:isolated -- --grep "first-run"` passes on a fresh checkout
- [ ] `npm run test:e2e:isolated -- --grep "first-run"` passes on a machine that has previously completed the wizard

AI-assisted — session URL not available for this session

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2617)

### DIFF
--- a/e2e-tests/tests/isolated/first-run-wizard.spec.ts
+++ b/e2e-tests/tests/isolated/first-run-wizard.spec.ts
@@ -10,8 +10,10 @@
  *
  * `app.fixture` pre-seeds `platform.firstRunComplete: true` to bypass the wizard for smoke tests.
  * This test deliberately needs the wizard to appear, so it uses `isolated.fixture` which performs
- * no such pre-seeding, and seeds only `platform.interfaceMode: 'simple'` to keep the app in the
- * mode where first-run gating applies (power mode bypasses the gate).
+ * no such pre-seeding, and seeds `platform.interfaceMode: 'simple'` (so the async resolveInternal
+ * path doesn't bypass the gate for power-mode users) plus `platform.firstRunComplete: false` (so a
+ * developer who has completed the wizard locally — writing firstRunComplete: true to dev-appdata —
+ * does not silently suppress the gate in this test).
  *
  * ## Why the dialog is always visible for a fresh profile
  *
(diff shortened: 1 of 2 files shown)


tombogle (approved):
> @tombogle reviewed 2 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2617)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2618, opened 2026-07-28
PT-4260: Unify wizard form component — WizardStepForm, IdentifyStep, SyncConsentStep

katherinejensen00:
## Summary

- Extracts a shared `WizardStepForm` presentational layout (heading + children + error + Back/secondary/primary button row) used by all wizard steps
- Replaces `identify.placeholder` and `sync-consent.placeholder` with full working implementations: `IdentifyStep` (name + code input, PAPI validation, save-and-restart) and `SyncConsentStep` (Sync / Skip automatic sync)
- Wires both into `FirstRunShell`'s `DEFAULT_STEP_COMPONENTS`; adds 8 new `en.json` localization keys

## Key files

| File | What it does |
|---|---|
| [`wizard-step-form.component.tsx`](src/renderer/components/first-run/wizard-step-form.component.tsx) | New shared layout: heading, children, optional error, Back left / secondary+primary right |
| [`identify-step.component.tsx`](src/renderer/components/first-run/steps/identify-step.component.tsx) | Paratext registration name + code, debounced PAPI validation, save-and-restart |
| [`sync-consent-step.component.tsx`](src/renderer/components/first-run/steps/sync-consent-step.component.tsx) | Sync / Skip automatic sync; fires `paratextBibleSendReceive.syncProjects` |
| [`first-run-shell.component.tsx`](src/renderer/components/first-run/first-run-shell.component.tsx) | `DEFAULT_STEP_COMPONENTS` now contains real steps instead of placeholders |
| [`assets/localization/en.json`](assets/localization/en.json) | 8 new keys (`%firstRun_step_identify_*`, `%firstRun_step_syncConsent_*`, `%firstRun_button_sync/skipSync%`) |

## Design decisions worth knowing

**Shell owns the footer, steps own their content.** `WizardStepForm` renders Back/secondary/primary inside the step body. The shell _also_ renders its own footer (Back/Skip/Next). For `SyncConsentStep`, this means both footers are visible simultaneously — acknowledged in a `// Known limitation` comment, tracked by PT-4257 (`managesOwnFooter`). `setCanProceed(false)` on mount disables the shell Next so users can't bypass the step's own actions.

**`IdentifyStep` calls `platform.restart` directly.** After saving registration data the app must relaunch so the C# layer picks up the new credentials. An injectable `onRestartAfterSave` prop enables test isolation without calling the real command.

**Registration code constants are copied from the `paratext-registration` extension** (with a `// keep in sync` comment and source paths cited). Exporting them from the extension for cross-package import would require a public API surface change — left for a follow-up.

## Tests

- 20 unit tests for `IdentifyStep` (validation, demo mode, save/restart, error paths)
- 10 unit tests for `SyncConsentStep` (sync success/failure, skip, back, busy state)
- 9 unit tests for `WizardStepForm` (layout, button slots, error slot)
- Shell navigation tests refactored to use `DUMMY_STEPS` (decoupled from real step implementations)

## Not in scope

`SyncProgressStep` (PT-4258) — the `sync-progress.placeholder` is intentionally kept.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2618)


src/renderer/components/first-run/wizard-step-form.component.tsx
```diff
@@ -27,9 +35,16 @@ export function WizardStepForm({
   return (
     <div className="tw:flex tw:flex-col tw:gap-3">
       <h2 className="tw:text-base tw:font-semibold">{heading}</h2>
-      <div className="tw:text-sm tw:text-muted-foreground">{children}</div>
-      {error && <p className="tw:text-sm tw:text-destructive">{error}</p>}
+      <div>{children}</div>
```
irahopkinson, line 44: **[1 — highest priority] Removing the wrapper styling silently restyles the already-shipped Sync consent screen.**

```diff
- <div className="tw:text-sm tw:text-muted-foreground">{children}</div>
+ <div>{children}</div>
```

The new TSDoc says "each step owns its own text colors" — but the component's *other* consumer was never updated to own them. `sync-consent-step.component.tsx:75` passes a bare `<p>{strings['%firstRun_step_syncConsent_body%']}</p>` with no classes, and is unchanged in this PR. Its body copy now renders at default browser size and default foreground colour instead of `text-sm` / `text-muted-foreground`.

That's an unrequested visual regression on a screen this PR doesn't otherwise touch — in a PR whose purpose is visual consistency. Both review axes flagged this independently.

**Fix**: either restore the wrapper classes and let `IdentifyStep` override locally, or add `tw:text-sm tw:text-muted-foreground` to `SyncConsentStep`'s `<p>` in the same change.

src/renderer/components/first-run/wizard-step-form.component.tsx
```diff
@@ -27,9 +35,16 @@ export function WizardStepForm({
...
+      {/* aria-live="assertive" so screen readers announce errors immediately — errors appear after an
+          async operation (e.g. sync failure) while focus stays on the primary button. */}
+      {error && (
+        <p className="tw:text-sm tw:text-destructive" aria-live="assertive" role="alert">
+          {error}
+        </p>
+      )}
       <div className="tw:flex tw:justify-between tw:gap-2">
+        {/* Empty div when backButton is omitted — intentional justify-between spacer to keep secondary+primary right-aligned. */}
```
irahopkinson: **[2] This Back-left button row is dead in production — the shipped wizard still shows two right-aligned rows.**

The component implements PT-4257 correctly (`tw:justify-between`, back slot on the left). But `git grep 'backButton\|secondaryButton' -- src/` returns hits only in this file, its `.stories.tsx` and its `.test.tsx` — **no step passes either prop**.

The shell renders its own row instead (`first-run-shell.component.tsx:173`):

```tsx
<div className="tw:flex tw:justify-end tw:gap-2">   // Back, Skip and Next all right-aligned
```

So on Identify the user sees "Save and restart" right-aligned in the body, and "Back" right-aligned on a **separate row below it** — the opposite of PT-4257's *"Back on the left; Skip/Next together on the right."*

**Fix**: either wire `onBack`/`onSkip` into the slots and drop the shell's duplicate row, or delete the two slots until a consumer needs them (they currently exist only to satisfy the stories and tests).

src/renderer/components/first-run/wizard-step-form.component.tsx
```diff
@@ -27,9 +35,16 @@ export function WizardStepForm({
   return (
     <div className="tw:flex tw:flex-col tw:gap-3">
       <h2 className="tw:text-base tw:font-semibold">{heading}</h2>
-      <div className="tw:text-sm tw:text-muted-foreground">{children}</div>
-      {error && <p className="tw:text-sm tw:text-destructive">{error}</p>}
+      <div>{children}</div>
+      {/* aria-live="assertive" so screen readers announce errors immediately — errors appear after an
+          async operation (e.g. sync failure) while focus stays on the primary button. */}
+      {error && (
+        <p className="tw:text-sm tw:text-destructive" aria-live="assertive" role="alert">
```
irahopkinson: **[minor] This error paragraph is now byte-for-byte identical to the shell's** (`first-run-shell.component.tsx:168`).

In a PR about unifying wizard presentation, the two error renderers diverged and were then hand-synchronised rather than extracted. Worth a tiny `WizardError`, or have the shell reuse this slot, so the next a11y tweak lands once.

Also: `role="alert"` already implies `aria-live="assertive"`, so the pair is redundant in both copies.

src/renderer/components/first-run/first-run-step-props.model.ts
```diff
@@ -17,8 +17,10 @@ export interface FirstRunStepProps {
   /** Return to the previous step. Absent on the first step (Language). */
   onBack?: () => void;
   /**
-   * Skip the rest of setup and finish. Present only when `setCanSkip(true)` has been called by the
-   * current step.
+   * Skip the rest of setup and finish, marking sync as skipped. The shell always passes a defined
+   * function; the type is optional to support standalone usage (Storybook, unit tests) where the
+   * shell is absent. `setCanSkip` controls whether the shell's own footer Skip button is shown;
+   * steps that manage their own footer (e.g. SyncConsentStep) call `onSkip` directly instead.
```
irahopkinson: **[3] This TSDoc states two things that are false.**

1. *"The shell always passes a defined function"* — `first-run-shell.component.tsx:118-121` is `useMemo(() => (canSkip ? () => … : undefined), …)`, so `onSkip` is `undefined` until a step calls `setCanSkip(true)`.
2. *"steps that manage their own footer (e.g. SyncConsentStep) call `onSkip` directly instead"* — `git grep onSkip -- src/` shows `SyncConsentStep` never references it. It calls `setCanSkip?.(true)` (`sync-consent-step.component.tsx:42`) and relies on the shell's footer button.

This doc was added by `3d2940756ab` for a SyncConsentStep design that `b2d0286a5ca` then reverted — the doc survived the revert. The documented contract now inverts the real one, so a step author following it would call `onSkip()` without `setCanSkip(true)` and hit an undefined call.

Also, *"marking sync as skipped"* over-specifies: the shell calls `completeFirstRun({ skippedStep: step })` for whatever step is current.

src/renderer/components/first-run/steps/identify-step.component.tsx
```diff
@@ -5,6 +5,7 @@ import { Alert, AlertDescription, AlertTitle, Button, Input, Spinner } from 'pla
 import { getErrorMessage, LocalizeKey } from 'platform-bible-utils';
 import { AlertCircle, CircleCheck } from 'lucide-react';
 import { ChangeEvent, useEffect, useRef, useState } from 'react';
+import { WizardStepForm } from '../wizard-step-form.component';
```
irahopkinson, line 8: **[4] `useEffect` here can paint an *enabled* Next button for one frame.**

`identify-step.component.tsx:76` is `useEffect(() => setCanProceed?.(undefined), [setCanProceed])`, whereas `SyncConsentStep` uses `useLayoutEffect` for the identical job, with the comment *"Hide the shell's generic Next button before the first paint so it never flashes visible."*

The shell's `canProceed` initialises to `useState(true)` (`first-run-shell.component.tsx:64`) and renders Next whenever `canProceed !== undefined`, with `disabled={!canProceed}` (`:186-187`). `goToStep` isn't called for the entry step, so when the wizard *enters* at `identify` (the resume path) the shell paints an **enabled** Next for a frame — clicking it would skip registration.

Two sibling steps solving the identical problem two different ways is exactly what PT-4260 exists to remove. **Fix**: switch to `useLayoutEffect` (import change is on this line).

src/renderer/components/first-run/steps/identify-step.component.tsx
```diff
@@ -219,9 +225,13 @@ export function IdentifyStep({ onNext, setCanProceed, onRestartAfterSave }: Iden
       // The process terminates here; setIsRestarting(true) above keeps the button in "Restarting…"
       // until the app exits.
       await (onRestartAfterSave ?? (() => commandService.sendCommand('platform.restart')))();
+      // In production, platform.restart relaunches the process and this line never runs.
```
irahopkinson: **[5] This comment is factually wrong, and the line below it is reachable in production.**

`platform.restart`'s handler (`src/main/main.ts:982-996`) is:

```ts
async () => {
  if (!willRestart) { willRestart = true; app.relaunch({ … }); }
  app.quit();
}
```

It resolves as soon as `app.quit()` is *invoked* — and its own comment references *"the few seconds it takes `app.quit` to run."* So the JSON-RPC reply can reach the renderer before the window is destroyed, flipping the "Restarting…" overlay back to a live form with "Save and restart" re-enabled during the quit window. Re-clicking is harmless thanks to the `willRestart` guard, but the flicker is a regression on the pre-PR behaviour, and the comment will mislead the next maintainer.

**Fix**: gate the reset on the injected path so production is untouched — `if (onRestartAfterSave && isMounted.current) setIsRestarting(false);` — and correct the comment.

---

Separately, and just so it isn't lost: PT-4177's UX review (2026-07-25) says *"Prefer handling the restart via the **single-restart-at-the-end** approach (PT-4182) rather than a per-step restart."* The default path here still relaunches from step 3. That's fine if PT-4182 tracks it (it does) — the `onRestartAfterSave` seam makes batching straightforward later.

src/renderer/components/first-run/steps/identify-step.component.tsx
```diff
@@ -245,11 +255,14 @@ export function IdentifyStep({ onNext, setCanProceed, onRestartAfterSave }: Iden
   }
 
   return (
-    <div className="tw:flex tw:flex-col tw:gap-4">
-      <h2 className="tw:m-0 tw:text-sm tw:font-normal">
-        {strings['%firstRun_step_identify_heading%']}
-      </h2>
-
+    <WizardStepForm
```
irahopkinson, line 274: **[6] `IdentifyStep` bypasses `WizardStepForm`'s `error` slot entirely.**

Only `heading` and `primaryButton` are passed. Errors render as two `<Alert variant="destructive">` blocks inside `children` (`:321-335`), while `sync-consent-step.component.tsx:67` passes `error={error}` and gets a plain `<p className="tw:text-sm tw:text-destructive" role="alert">`.

So the two screens the ticket exists to unify present validation failures in visibly different ways, and only the sync one gets the live-region announcement. PT-4260's DoD asks for *"consistent styling/**validation**/button layout."*

**Fix**: pick one presentation and push it into `WizardStepForm`. The `Alert` is the richer option — it already has title + description and `role="alert"` from `shadcn-ui/alert.tsx:44`.

src/renderer/components/first-run/steps/identify-step.component.tsx
```diff
@@ -136,6 +142,8 @@ export function IdentifyStep({ onNext, setCanProceed, onRestartAfterSave }: Iden
         return;
       }
       setIsValidating(true);
+      // Read from ref so we always get the latest strings even if PAPI delivered them mid-wait.
+      const s = stringsRef.current;
```
irahopkinson: **[nit]** `const s` — Code-Style-Guide asks for clear variable names and no unestablished abbreviations. `latestStrings` reads better next to the `s['%paratextRegistration_alert_invalidRegistration%']` use sites just below.

irahopkinson (commented):
## Review — PT-4260 (two-axis: standards conformance + spec fidelity)

### ⚠️ Rebase first

**This branch is based on tombogle's #2590 (`pt-4020-improve-tooltip-styling`), not on `main`.** Rebasing onto `main` is worth doing before anything else here:

- "Files changed" currently shows ~800 lines of someone else's work, making this PR look ~3× its real size.
- The copy carried here is **stale** — it predates #2590's screen-reader accessibility fix (`87f956bb16e`), its `Kbd` line-wrapping fix (`9616e67e795`), and a 101-line test file. If #2618 merged first, it would land the older version and regress #2590.
- Debris leaked into your *own* commits during rebases (items 9 and 10 below). Separately, commit `2cf495eb978` is labelled "test(pt9-css-converter): increase prettier-conformant test timeout to 30 s" but touches only `first-run-shell.component.tsx` and adds four merge-conflict-marker lines. HEAD is clean and the squash-merge hides it, so nothing reaches `main` — but commit-by-commit review and `git bisect` are broken until the rebase.

**Everything below was reviewed against `6a56a574015`** (the parent of your first commit), so it covers your work only — none of the findings are about #2590's code.

### Prioritised

| # | Issue | Where | Axis |
|---|---|---|---|
| 1 | `WizardStepForm` children wrapper de-styled — silently regresses the **untouched** Sync consent screen | `wizard-step-form.component.tsx:38` | both |
| 2 | PT-4257's Back-left row is dead in production; the shell still renders two right-aligned rows | `wizard-step-form.component.tsx:47` | spec |
| 3 | `onSkip` TSDoc states two things that are false | `first-run-step-props.model.ts:23` | both |
| 4 | Entering at `identify` paints an **enabled** Next for one frame (`useEffect` vs `useLayoutEffect`) | `identify-step.component.tsx:8` | spec |
| 5 | `setIsRestarting(false)` — the comment justifying it is factually wrong | `identify-step.component.tsx:228` | spec |
| 6 | `IdentifyStep` bypasses the `error` slot, so the two "unified" screens show errors differently | `identify-step.component.tsx:258` | both |
| 7 | 6 of 9 new `WizardStepForm` tests are non-falsifiable; the layout contract is untested | `wizard-step-form.component.test.tsx` | both |
| 8 | `es.json` skipped by the locale backfill (the guide mandates en **and** es) | `assets/localization/fr.json` | standards |
| 9 | Rebase debris: unrelated stories file added by `b2d0286a5ca` | `two-step-delete-tooltip-overlay.component.stories.tsx` | scope |
| 10 | Rebase debris: manual `.mdx` emphasis flip (Prettier won't re-normalise — `.prettierignore:72`) | `src/stories/keyboard-shortcuts.mdx:15` | scope |

Items 1 and 6 were found independently by both review axes, which makes them the highest-confidence findings. Remaining lower-severity notes (naming, `strings` vs `stringsRef` consistency, test-timer hygiene, `DUMMY_STEPS` coverage) are inline.

### Correction to an earlier read

I initially thought PT-4260's DoD clause *"one shared form component used by **both** registration and sync screens"* was unmet. That was wrong — `sync-consent-step.component.tsx:6` does import `WizardStepForm` (it landed via PT-4178). The DoD is satisfied at HEAD; this PR contributes the registration half.

### Verified clean

- DoD "Storybook story" — 5 stories covering every slot including `AllSlots`.
- Copied registration constants (`identify-step.component.tsx:16-23`) match the `paratext-registration` extension verbatim, source paths cited, `'[['` quirk documented. No drift.
- Stale-response handling in `validateRegistration` — generation counter claimed before every guard, `isMounted` checked before each state write, timer cleared on unmount and at the top of each call. Correct.
- The `stringsRef` staleness guard correctly replaces captured-closure reads in the debounce callback.
- Save-error clearing on re-typing closes the gap where a stale alert could coexist with a new validation error.
- Every `%firstRun_*%` key referenced by production renderer code exists in `en.json`; no raw key leaks to the UI.
- `dist/` at HEAD is consistent with `src/` at HEAD for both libraries.

**Not run**: the test suite (review worktree has no `node_modules`), so the two test-timing findings are read-only analyses, not demonstrated failures.

---
AI-assisted review (Claude Opus 5). Localization key deprecation/immutability process was treated as out of scope for the in-flight `%firstRun_*%` placeholder keys.

src/renderer/components/first-run/steps/identify-step.component.tsx
```diff
@@ -219,9 +225,13 @@ export function IdentifyStep({ onNext, setCanProceed, onRestartAfterSave }: Iden
...
       // until the app exits.
       await (onRestartAfterSave ?? (() => commandService.sendCommand('platform.restart')))();
+      // In production, platform.restart relaunches the process and this line never runs.
+      // If the restart resolves without relaunching (e.g. a test stub), reset the spinner so the
+      // UI doesn't get stuck.
+      if (isMounted.current) setIsRestarting(false);
     } catch (err) {
       if (!isMounted.current) return;
-      setSaveError(strings['%general_error_title%']);
+      setSaveError(stringsRef.current['%general_error_title%']);
```
irahopkinson: **[minor] Mixed `strings` / `stringsRef.current` access makes it unclear which is authoritative.**

The ref is genuinely needed only inside the `setTimeout` callback in `validateRegistration`, whose closure can outlive the render. `saveAndRestart` is recreated every render, so its `strings` closure is already current — reading the ref here just makes the reader wonder why.

Suggest using `strings` in `saveAndRestart` and scoping the ref (and its justification comment) to the debounce path.

src/renderer/components/first-run/wizard-step-form.component.test.tsx
```diff
@@ -0,0 +1,102 @@
...
+    render(
+      <WizardStepForm
+        heading="h"
+        primaryButton={<button type="button">OK</button>}
+        error="Something failed"
+      >
+        body
+      </WizardStepForm>,
+    );
+    expect(screen.getByText('Something failed')).toBeInTheDocument();
```
irahopkinson: **[7] Six of the nine tests here can't fail, and the one contract worth pinning isn't tested.**

The three "when omitted" tests query a literal that the test never supplies:

```ts
it('does not render error text when error prop is omitted', …
  expect(screen.queryByText('Something failed')).not.toBeInTheDocument());
```

`'Something failed'` is never passed in that test, so it passes no matter what the component does — same shape for the `/back/i` and `/skip/i` pairs. Testing-Guide's Test Quality Checklist asks for *"Falsifiable — test fails when the implementation is broken."* The other three assert that a `ReactNode` passed as a prop appears in the output, i.e. React's own behaviour ("What NOT to Test → Framework behavior").

Meanwhile the thing this component exists to guarantee — PT-4257's Back-left / secondary+primary-right **ordering**, and the error's `role="alert"` — has no assertion. The button row could be reordered without a failing test.

**Fix**: replace the vacuous negatives with `queryByRole('alert')` and a DOM-order assertion over the button row.

src/renderer/components/first-run/steps/identify-step.component.test.tsx
```diff
@@ -41,10 +41,17 @@ const mockIsDemoMode = vi.mocked(firstRunStore.isDemoMode);
 
 const VALID_CODE = 'ABCDEF-ABCDEF-ABCDEF-ABCDEF-ABCDEF';
 
+// userEvent instance created once at module scope (before any fake-timer installation).
+// advanceTimers references vi.advanceTimersByTimeAsync by function reference — vitest replaces
+// setTimeout/clearTimeout with fakes in beforeEach and we advance manually in each test.
+const user = userEvent.setup({ advanceTimers: vi.advanceTimersByTimeAsync });
```
irahopkinson: **[minor] Module-scope `userEvent` shared across all 20 tests, plus two competing timer-advance mechanisms.**

Creating `user` before `useFakeTimers` is the documented workaround, so the ordering is right — but one instance is now shared across every test in the file and its pointer/keyboard state isn't reset by RTL cleanup. And `advanceTimers: vi.advanceTimersByTimeAsync` here plus `shouldAdvanceTime: true` below both drive the clock; the comments acknowledge both without saying why both are needed.

Suggest a per-test `userEvent.setup()` created after the fake-timer install, and picking one advance mechanism. I couldn't run the suite in the review worktree, so this is hygiene rather than a demonstrated flake.

src/renderer/components/first-run/steps/identify-step.component.test.tsx
```diff
@@ -41,10 +41,17 @@ const mockIsDemoMode = vi.mocked(firstRunStore.isDemoMode);
...
+
 beforeEach(() => {
-  vi.useFakeTimers();
   mockSendCommand.mockReset();
   mockIsDemoMode.mockReturnValue(false);
+  // shouldAdvanceTime: true lets userEvent's internal scheduler timers tick in real time
+  // while still allowing manual timer control for debounce assertions.
+  vi.useFakeTimers({ shouldAdvanceTime: true });
```
irahopkinson: **[minor] `shouldAdvanceTime: true` reintroduces a wall-clock dependency into the debounce assertions.**

It advances the fake clock in step with real time. The debounce tests then assert the error is *absent* before calling `vi.advanceTimersByTime(INVALID_CODE_DISPLAY_DEBOUNCE_MS + 1)` — so if typing takes more than `INVALID_CODE_DISPLAY_DEBOUNCE_MS` (1000 ms) of wall-clock on a loaded CI runner, the debounce fires on its own and the pre-advance assertion fails.

Given `32b7a3a31a5` was specifically about fixing timing flakiness, worth confirming this doesn't trade one flake for another.

src/renderer/components/first-run/steps/identify-step.component.test.tsx
```diff
@@ -138,7 +144,7 @@ describe('IdentifyStep', () => {
       .mockResolvedValueOnce(true) // validateParatextRegistrationData
       .mockResolvedValueOnce(undefined); // setParatextRegistrationData
     const onRestartAfterSave = vi.fn().mockReturnValue(new Promise<never>(() => {}));
-    const user = userEvent.setup({ advanceTimers: vi.advanceTimersByTime });
+
```
irahopkinson, line 200: **[minor] The one behaviour change in `saveAndRestart` has no test.**

Both restart tests stub with `new Promise(() => {})`, which never settles — so `if (isMounted.current) setIsRestarting(false);` (component `:231`) never executes in any test. That line is the sole behavioural change to this function, and it's what the prop's new TSDoc documents (*"If it resolves without actually relaunching the app, the component resets the spinner overlay"*).

A test that resolves `onRestartAfterSave` and asserts the overlay clears would pin it.

src/renderer/components/first-run/first-run-shell.component.test.tsx
```diff
@@ -106,11 +89,25 @@ vi.mock('platform-bible-react', () => {
...
+// navigation tests at a stub step that never re-enables Next.
+function makeDummyStep(label: string): ComponentType<FirstRunStepProps> {
+  function DummyStep({ setCanProceed }: FirstRunStepProps) {
+    useEffect(() => setCanProceed?.(true), [setCanProceed]);
+    return <p>{label}</p>;
+  }
+  return DummyStep;
 }
 
+const DUMMY_STEPS: Record<FirstRunStep, ComponentType<FirstRunStepProps>> = {
```
irahopkinson, line 171: **[minor] The `DUMMY_STEPS` refactor improves isolation but drops the integration this PR actually changes.**

Nothing now verifies that `DEFAULT_STEP_COMPONENTS` maps `identify` → the real `IdentifyStep` with a working `setCanProceed(undefined)` handshake — which is precisely what changed here.

The file also now mixes conventions: six tests still use `DEFAULT_STEP_COMPONENTS` (`:186, :198, :240, :252, :292`), and the `command.service` / `network.service` mocks are retained for paths that no longer need them.

Worth keeping one test on the real components for the entry-step handshake.

assets/localization/fr.json
```diff
@@ -140,6 +143,16 @@
   "%firstRun_language_title%": "Choisissez votre langue",
   "%firstRun_loading%": "Démarrage de la configuration…",
   "%firstRun_loading_detail%": "Vérification de vos informations d'enregistrement. Cela peut prendre un moment.",
+  "%firstRun_step_identify_heading%": "Saisissez vos informations d'enregistrement",
```
irahopkinson, line 146: **[8] `es.json` was skipped by this backfill.**

These `%firstRun_step_identify_*%` keys were added to `fr`, `zh-hans` and `zh-hant` — but not to `es`. Localization-Guide → Language Requirements: *"**Always provide both `en` AND `es` translations.** Both are required for the build to pass."*

`es.json` is currently missing 10 `%firstRun_*%` keys, including all four `%firstRun_step_identify_*%` that the rewritten `IdentifyStep` renders, so Spanish users fall back to English for the whole Identify step.

The gap pre-dates this PR, so it isn't introduced here — but this commit's stated purpose is closing translation gaps, and it closes them for the three optional languages while skipping the mandated one. (I couldn't find a build check that actually enforces the guide's "required to pass", so that wording may be aspirational; the documented rule stands either way.)

extensions/src/platform-scripture-editor/src/two-step-delete-tooltip/two-step-delete-tooltip-overlay.component.stories.tsx
```diff
@@ -0,0 +1,149 @@
+import type React from 'react';
```
irahopkinson: **[9] Rebase debris — this file belongs to #2590, not PT-4260.**

`git show b2d0286a5ca --stat` lists **only** this file, despite a 12-line commit message about restoring `SyncConsentStep`. It should drop out when the branch is rebased onto `main`.

src/stories/keyboard-shortcuts.mdx
```diff
@@ -12,7 +12,7 @@ defined by bundled extensions, and the shortcuts defined inside the shared `plat
 component library. It is the one source of truth for what shortcuts exist, what they do, and where
 each is handled in code.
 
-For how to _display_ a shortcut in the UI (per-OS formatting conventions), see the
+For how to *display* a shortcut in the UI (per-OS formatting conventions), see the
```
irahopkinson: **[10] Rebase debris — unrelated emphasis flip** (`_display_` → `*display*`).

Worth noting this one is a *manual* edit rather than tooling: `.prettierignore:72` excludes `*.mdx`, so nothing will re-normalise it. Drop the hunk during the rebase.

assets/localization/fr.json
katherinejensen00: Fixed — added all four `%firstRun_step_identify_*%` keys to `es.json`.

extensions/src/platform-scripture-editor/src/two-step-delete-tooltip/two-step-delete-tooltip-overlay.component.stories.tsx
katherinejensen00: Fixed — removed the file. It slipped in during the rebase onto #2590; it has no place in this PR.

src/renderer/components/first-run/first-run-shell.component.test.tsx
katherinejensen00: Fixed — added one integration test that renders `FirstRunShell` with `entryStep="identify"` and `DEFAULT_STEP_COMPONENTS`, then asserts Next is absent. This covers the `useLayoutEffect`/`setCanProceed` handshake with the real component.

src/renderer/components/first-run/wizard-step-form.component.tsx
katherinejensen00: Fixed — added `tw:text-sm tw:text-muted-foreground` directly to `SyncConsentStep`'s `<p>` instead of restoring the wrapper. This way IdentifyStep's form field labels keep the default foreground without needing an override, and each step owns its own text styling as the TSDoc says.

src/renderer/components/first-run/wizard-step-form.component.tsx
katherinejensen00: Fixed — the error slot now uses `<Alert variant="destructive">` with `AlertTitle` and an optional `AlertDescription` prop, which also resolves item 6. The shell's top-level error is a different concern (it shows a wizard-level load failure), so they're no longer equivalent.

src/renderer/components/first-run/wizard-step-form.component.tsx
katherinejensen00: Fixed — removed `backButton` and `secondaryButton` entirely. Neither has a caller in the current codebase, so the cleaner approach is to drop the slots rather than leave them dormant. The shell keeps its own footer row for Back/Next/Skip; PT-4257 can wire the step-owned layout when a step actually needs it.

src/renderer/components/first-run/wizard-step-form.component.test.tsx
katherinejensen00: Fixed — the catch block now uses `strings` directly. The `stringsRef` pattern is scoped to the `validateRegistration` `setTimeout` callback where it's genuinely needed (the timeout closure can outlive the render by up to 1 s); `saveAndRestart` is recreated on every render so its closure is always current at call time.

src/renderer/components/first-run/steps/identify-step.component.tsx
katherinejensen00: Fixed — reverted to the simpler, accurate version. `SyncConsentStep` uses `setCanSkip(true)` and never touches `onSkip` directly; the shell is not guaranteed to pass a defined function.

src/renderer/components/first-run/steps/identify-step.component.tsx
katherinejensen00: Fixed — renamed to `latestStrings`.

src/renderer/components/first-run/steps/identify-step.component.tsx
katherinejensen00: Fixed — gated the reset on `if (onRestartAfterSave && isMounted.current)` so production behavior (spinner stays until the process terminates) is unchanged, while test stubs that resolve without restarting get the UI reset. Updated the comment to reflect the actual production timeline you described.

src/renderer/components/first-run/steps/identify-step.component.tsx
katherinejensen00: Fixed — the catch block now uses `strings` directly. The `stringsRef` pattern is scoped to the `validateRegistration` `setTimeout` callback where it's genuinely needed (the timeout closure can outlive the render by up to 1 s); `saveAndRestart` is recreated on every render so its closure is always current at call time.

src/renderer/components/first-run/steps/identify-step.component.tsx
katherinejensen00: Fixed — both error states (validation and save failure) are now merged into `activeError`/`activeErrorDescription` and passed through WizardStepForm's `error` and `errorDescription` props. The two are mutually exclusive (any keystroke clears both via `validateRegistration`), so the merge is safe. WizardStepForm now renders the error as `<Alert variant="destructive">` with title + optional description.

src/renderer/components/first-run/steps/identify-step.component.test.tsx
katherinejensen00: Fixed — moved to a `setupUser()` factory called inside each test body after `vi.useFakeTimers()` runs, so every test gets a fresh instance with the correct fake-timer context. Removed `shouldAdvanceTime: true`; `advanceTimers: vi.advanceTimersByTimeAsync` on the per-test instance is sufficient.

src/renderer/components/first-run/steps/identify-step.component.test.tsx
katherinejensen00: Fixed — removed `shouldAdvanceTime: true`. The per-test `setupUser()` with `advanceTimers: vi.advanceTimersByTimeAsync` handles userEvent's scheduling without a wall-clock dependency.

src/renderer/components/first-run/steps/identify-step.component.test.tsx
katherinejensen00: Fixed — added a test that provides a resolving `onRestartAfterSave`, waits for the "Save and restart" button to reappear after clicking, and asserts the "Restarting…" overlay is gone.

src/renderer/components/first-run/first-run-step-props.model.ts
katherinejensen00: Fixed — reverted to the simpler, accurate version. `SyncConsentStep` uses `setCanSkip(true)` and never touches `onSkip` directly; the shell is not guaranteed to pass a defined function.

src/stories/keyboard-shortcuts.mdx
katherinejensen00: Fixed — reverted to `_display_`. Good catch that `.prettierignore` excludes `.mdx` so it wouldn't self-correct.

katherinejensen00 (commented):
> @katherinejensen00 made 17 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2618)* status: 0 of 50 files reviewed, 17 unresolved discussions (waiting on irahopkinson).

assets/localization/fr.json
irahopkinson: Partly — thanks, the four `%firstRun_step_identify_*%` keys are in and `es.json` went from 22 to 26 `%firstRun_*%` keys.

It's still 5 behind `fr` / `zh-hans` / `zh-hant` (31 each): the four `%firstRun_step_syncProgress_*%` keys and `%firstRun_button_continueWithoutSetup%`. Same backfill gap, and `es` is the one language the Localization Guide mandates alongside `en`.

Happy for that to be a follow-up rather than a blocker on this PR — your call. Flagging so it doesn't quietly disappear.

extensions/src/platform-scripture-editor/src/two-step-delete-tooltip/two-step-delete-tooltip-overlay.component.stories.tsx
irahopkinson: Reopening this one — the ground shifted underneath us, through no fault of the fix.

When you replied, deleting the file was correct. But **#2590 merged to `main` on 30 Jul** (`fddbf8342b7`), so `two-step-delete-tooltip-overlay.component.stories.tsx` is now legitimately on `main` — and this branch still carries the deletion.

- `git diff origin/main...HEAD` shows `161 -` for it (pure deletion, not a rename).
- The file is absent at HEAD, while its siblings `two-step-delete-tooltip-overlay.component.tsx`, `two-step-delete-tooltip.utils.ts` and `two-step-delete-tooltip.utils.test.ts` are all present.
- It's the only remaining file in this diff outside `src/renderer/` and `assets/localization/`.

Merging as-is would remove Tom's Storybook stories from `main`. Restoring it should be a one-liner:

```bash
git checkout origin/main -- extensions/src/platform-scripture-editor/src/two-step-delete-tooltip/two-step-delete-tooltip-overlay.component.stories.tsx
```

After that the diff should be purely `src/renderer/` + `assets/localization/`, which is what PT-4260 ought to touch.

src/renderer/components/first-run/first-run-shell.component.test.tsx
irahopkinson: Verified at `:547` — `render(<FirstRunShell entryStep="identify" stepComponents={DEFAULT_STEP_COMPONENTS} />)` asserting Next is absent.

That covers the real `useLayoutEffect` / `setCanProceed(undefined)` handshake against the actual `IdentifyStep`, which is the integration the `DUMMY_STEPS` refactor had removed. Exactly the right shape.

src/renderer/components/first-run/wizard-step-form.component.tsx
irahopkinson: Verified at `sync-consent-step.component.tsx:83` — `<p className="tw:text-sm tw:text-muted-foreground">`.

Agreed this is the better of the two options I suggested: `IdentifyStep`'s field labels keep the default foreground without needing an override, and the TSDoc's "each step owns its own text colors" is now actually true rather than aspirational.

src/renderer/components/first-run/wizard-step-form.component.tsx
irahopkinson: Verified — the step-level error is now `<Alert variant="destructive">` with `AlertTitle` and an optional `AlertDescription`, while the shell's `<p>` stays a wizard-level load failure.

Agreed they're no longer the same concern, so there's nothing left to extract. Dropping this.

src/renderer/components/first-run/wizard-step-form.component.tsx
irahopkinson: Verified — and the code is better than the reply describes, so worth correcting for anyone reading this thread later.

`secondaryButton` is gone, but `backButton` was **kept** (`wizard-step-form.component.tsx:16-21`) and is now genuinely wired: `IdentifyStep:86` calls `setManagesOwnFooter?.(true)` and `:278` passes `backButton`. That hooks into the `managesOwnFooter` seam PT-4257 landed on `main` (#2619), so the shell no longer stacks a second row and Back really is flush-left.

That resolves the finding properly rather than by deletion — which is the outcome I was hoping for. Nothing to change.

src/renderer/components/first-run/wizard-step-form.component.test.tsx
irahopkinson: Same Reviewable mix-up as the `useLayoutEffect` thread — the reply text here is the `stringsRef` answer.

The fix itself is in, and goes past what I asked for: the vacuous `queryByText('Something failed')` negatives are replaced with `getByRole('alert')` / `queryByRole('alert')` (`:50-71`), and there's a new `renders heading before children before error before button in DOM order` test at `:115`. That last one pins the layout contract, which was the actual gap — the button row can no longer be reordered without a failing test.

Marking satisfied on the code.

src/renderer/components/first-run/steps/identify-step.component.tsx
irahopkinson: Heads-up: the reply text on this thread is the `onSkip` TSDoc answer — it looks like Reviewable landed it on the wrong discussion.

The actual fix did ship, though: `identify-step.component.tsx:84` is now `useLayoutEffect`, and the same effect also calls `setManagesOwnFooter?.(true)`. Entering the wizard at `identify` can no longer paint an enabled Next for a frame.

Marking satisfied on the code.

src/renderer/components/first-run/steps/identify-step.component.tsx
irahopkinson: Verified — `latestStrings` at `:157`.

src/renderer/components/first-run/steps/identify-step.component.tsx
irahopkinson: Verified at `:242` — `if (onRestartAfterSave && isMounted.current) setIsRestarting(false);`.

Production path is untouched, test stubs still get their reset, and the comment now describes the real `app.quit()` timeline instead of claiming the line is unreachable.

src/renderer/components/first-run/steps/identify-step.component.tsx
irahopkinson: Verified — `saveAndRestart` reads `strings` directly now, and the ref is confined to the `validateRegistration` timeout closure, which is exactly the one place it's load-bearing. The reasoning in your reply matches what the code does.

src/renderer/components/first-run/steps/identify-step.component.tsx
irahopkinson: Verified. I checked the mutual-exclusivity claim specifically, because the merge depends on it being true — and it holds: `validateRegistration` clears `error`, `errorDescription`, `saveError` **and** `saveErrorDescription` on every call (`:137-142`), so `activeError = error || saveError` can't mask a live second error. `activeErrorDescription = error ? errorDescription : saveErrorDescription` pairs correctly with that precedence.

Both screens now route through the same `Alert`, which was the point of the finding.

src/renderer/components/first-run/steps/identify-step.component.test.tsx
irahopkinson: Verified — per-test `setupUser()` called inside each test body after `vi.useFakeTimers()`, so there's no shared instance carrying pointer/keyboard state across the file.

src/renderer/components/first-run/steps/identify-step.component.test.tsx
irahopkinson: Verified — `shouldAdvanceTime` is gone, so the debounce assertions no longer race the wall clock on a loaded runner. One advance mechanism, per-test instance. Good.

src/renderer/components/first-run/steps/identify-step.component.test.tsx
irahopkinson: Verified — `clears the spinner overlay when onRestartAfterSave resolves` at `:222`. That's the test that actually executes the `if (onRestartAfterSave && isMounted.current)` line, which previously never ran in any test because both restart stubs used a never-settling promise.

src/renderer/components/first-run/first-run-step-props.model.ts
irahopkinson: Verified — both inaccurate claims are gone, and the `setCanSkip` / `onSkip` split is now described accurately.

src/stories/keyboard-shortcuts.mdx
irahopkinson: Verified — back to `_display_`.

irahopkinson (changes requested):
## Round 2 — re-reviewed against the rebased branch

Thanks, this is a thorough set of fixes. **Rebase confirmed done**: base is now `57bc81792a3`, i.e. current `main` including the merged #2619 (PT-4257) and #2590 (PT-4020). The diff is down to 14 files, all `src/renderer/` + `assets/localization/` — with one exception, below.

**15 of 17 verified fixed**, 1 partial, 1 that has turned into a regression through no fault of the fix.

### 🔴 One blocker — the stories-file deletion

Deleting `two-step-delete-tooltip-overlay.component.stories.tsx` was the right call when you replied. But #2590 merged to `main` on 30 Jul (`fddbf8342b7`), so that file is now legitimately on `main` and this branch still carries the deletion (`161 -`, pure delete, file absent at HEAD while its three siblings are present). Merging as-is would remove Tom's Storybook stories. Details and the one-line restore are in that discussion.

### ⚠️ One partial

`es.json` got the four `%firstRun_step_identify_*%` keys (22 → 26) but is still 5 behind `fr`/`zh` (31): the four `%firstRun_step_syncProgress_*%` plus `%firstRun_button_continueWithoutSetup%`. Fine as a follow-up — flagged so it isn't lost.

### Worth calling out

- **Item 2 landed better than its reply describes.** `secondaryButton` went, but `backButton` was *kept* and wired to PT-4257's `managesOwnFooter` seam (`IdentifyStep:86` + `:278`). So Back is genuinely flush-left now rather than the finding being closed by deletion — the outcome I was hoping for.
- **Two replies landed on the wrong threads** (Reviewable ordering): the `useLayoutEffect` thread got the `onSkip` answer, and the `WizardStepForm` test thread got the `stringsRef` answer. Both underlying fixes shipped correctly, so no action — noted in each thread so the history reads straight.
- **Item 7 went past the ask** — as well as replacing the vacuous negatives with `getByRole('alert')`, there's a new DOM-order test pinning the Back-left / primary-right layout contract.
- **I re-verified the load-bearing claim** on the `activeError` merge: `validateRegistration` clears all four error states on every call (`:137-142`), so the merge genuinely cannot mask a live second error.

### Leftovers, non-blocking

Two of Tom's commits (`caf8ba40536`, `c3101a1c594`) are still in the branch history, but contribute nothing to the diff now that #2590 is on `main` — harmless under squash-merge.

CI: CodeQL and both Analyze jobs pass; the three platform builds were still pending when I looked.

Once the stories file is restored I'm happy for this to go in.

---
AI-assisted review (Claude Opus 5).

> @irahopkinson+AI made 18 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2618#-:-OypTTvfESMt1zFSoEJj:bw5mrla)* status: 0 of 50 files reviewed, 17 unresolved discussions (waiting on irahopkinson).

irahopkinson (commented):
One thing to note, depending on how the squash and merge is done, Tom might get all the credit since he has the first commit. You can ensure you get the credit if you squash the branch yourself and make sure you are the author.

> @irahopkinson partially reviewed 50 files and all commit messages, made 1 comment, and resolved 15 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2618#-:-OypUtci9S69zt3ayVRO:b-715arz)* status: all files reviewed, 2 unresolved discussions (waiting on katherinejensen00).

extensions/src/platform-scripture-editor/src/two-step-delete-tooltip/two-step-delete-tooltip-overlay.component.stories.tsx
katherinejensen00: Restored in 2b3a03c8051 via `git checkout origin/main -- extensions/src/platform-scripture-editor/src/two-step-delete-tooltip/two-step-delete-tooltip-overlay.component.stories.tsx` — exactly the one-liner you suggested. You're right that #2590 (fddbf8342b7) landing on main flipped the correct action from delete to restore. The file is back at HEAD alongside its three siblings, and `git diff origin/main...HEAD` is now purely `src/renderer/` + `assets/localization/`.

assets/localization/fr.json
katherinejensen00: Done, not deferred — added the remaining 5 keys to `es.json` in 2b3a03c8051: the four `%firstRun_step_syncProgress_*%` keys and `%firstRun_button_continueWithoutSetup%`. `es` now matches `fr`/`zh-hans`/`zh-hant` at 31 `%firstRun_*%` keys. Since the Localization Guide mandates `es` alongside `en`, closing the gap here rather than punting it to a follow-up.

katherinejensen00 (commented):
I am fine with Tom getting the credit, but that is a good note. Also I restored the storybook file. Thanks for catching how it almost got messed up in the rebase!

> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2618#-:-OypiNSs6iIcAC7g6I0_:be0k7im)* status: 41 of 50 files reviewed, 1 unresolved discussion (waiting on irahopkinson and irahopkinson+AI).

assets/localization/fr.json
irahopkinson: Verified — `es.json` is at 31 `%firstRun_*%` keys, matching `fr` / `zh-hans` / `zh-hant` exactly. Appreciate you closing it here rather than taking the follow-up I offered.

extensions/src/platform-scripture-editor/src/two-step-delete-tooltip/two-step-delete-tooltip-overlay.component.stories.tsx
irahopkinson: Verified in `2b3a03c8051` — `git diff origin/main...HEAD` no longer lists the file at all, so it's byte-identical to `main`. Tom's stories are safe.

Unblocking. Thanks for turning it round quickly.

irahopkinson (commented):
## Round 3 — all clear ✅

Both open items verified fixed in `2b3a03c8051`:

- **Stories file restored** — no longer appears in `git diff origin/main...HEAD` at all, i.e. byte-identical to `main`.
- **`es.json` complete** — 31 `%firstRun_*%` keys, matching `fr` / `zh-hans` / `zh-hant`.

### Two things you did beyond the ask

- **Dropped the `dist/` debris entirely.** There's now zero `dist/` churn. Your reasoning checks out — the branch has no `lib/` source changes, so the full file list vs `main` is 13 files, all under `src/renderer/` or `assets/localization/`. Resetting to `main` was right, and it makes the PR reviewable at a glance.
- **The conditional `aria-describedby`.** I checked this one rather than taking it on trust, since dangling ARIA references are easy to get subtly wrong: the guard (`showInvalidCode ? 'identify-code-warning' : undefined`) matches the render condition exactly, so the attribute can no longer point at a non-rendered element, and the backend-error case is covered by the Alert's `role="alert"`. Correct.

### Status

CI is fully green now — all three platform builds plus both Analyze jobs and CodeQL (they were pending at my last pass).

All 17 discussions resolved from my side. Nothing outstanding — happy for this to go in. Nice work on the turnaround.

On the squash credit: understood, and no action needed — noting only that Tom's two commits are still first in the branch history, so that outcome holds unless you squash it yourself.

---
AI-assisted review (Claude Opus 5).

> @irahopkinson+AI made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2618#-:-OypvUASDCDx1YneERgG:b-scg82i)* status: 41 of 50 files reviewed, 1 unresolved discussion (waiting on irahopkinson).

irahopkinson (approved):
:lgtm:

> @irahopkinson reviewed 9 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2618#-:-OypvB7s2IWWakWmQvzv:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

irahopkinson (approved):
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2618)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2619, opened 2026-07-28
PT-4257: first-run wizard stepper and shell navigation

katherinejensen00:
# Code Review Summary

**Branch**: pt-3257-stepper

**Base**: origin/main

**Date**: 2026-07-28

**Review model**: Claude Sonnet 4.6

**Files changed**: 23

## Overview

This PR adds the `WizardStepper` presentational component to `platform-bible-react` and wires it into `FirstRunShell` — the modal chrome for Platform.Bible's first-run setup wizard (PT-4174 epic). The shell manages navigation across four numbered steps (`language` → `internet` → `identify` → `syncConsent`) plus a post-consent interstitial (`syncProgress`). A new `NumberedStep` type (vs the wider `FirstRunStep`) enforces at compile time that `syncProgress` cannot be added to `STEP_ORDER`. The shell handles locale-aware numeral formatting, accessibility (sr-only `aria-live` for step announcements), and a two-layer `isBusy` guard against double-invocation of `completeFirstRun`.

The `syncConsent` step uses a distinct "Finish" label (restored during review — the original removal was an unintended rebase artifact). All step placeholder copy was cleaned of an embedded Jira ticket reference, and step placeholder localization keys were added to all four non-English locale files.

## API Changes

- `lib/platform-bible-react`: Added `WizardStepper` component export
- `lib/platform-bible-react`: Added `WizardStepperProps` type export

Both are non-breaking additions with no changes to existing exports.

None.

- [x] **`%firstRun_step_internet_placeholder%` missing from non-English locale files** _(fixed during review: added all five `firstRun_step_*_placeholder` keys to fr, es, zh-hans, zh-hant — the other step placeholder keys were also absent from non-English files, so all five were added together)_

- [x] **`%firstRun_button_finish%` removed — "Next" used on `syncConsent` instead** _(fixed during review: key restored to all five locale files with original translations; shell now shows "Finish" on syncConsent and "Next" on all other steps)_

- [x] **`"Internet settings (PT-4259)"` — Jira ticket reference in user-visible copy** _(fixed during review: changed to `"Internet settings (coming soon)"` in all five locale files and both test mocks)_

[Author response: Finding 1 — oversight; author chose to add all step placeholder keys to non-English files rather than just the internet key. Finding 2 — unintentional removal, described as "a bad rebase decision." Finding 3 — author confirmed Jira ticket numbers should never appear in user-facing text.]

### Minor — Consider

- [x] **`WizardStepper` export out of alphabetical order in `lib/platform-bible-react/src/index.ts`** _(fixed during review: moved after `TextFieldProps`)_

- [x] **No test for `InternetPlaceholderStep` in `steps.placeholder.test.tsx`** _(fixed during review: added test case matching the existing `SyncConsentPlaceholderStep` pattern)_

- [x] **Empty `<div>` placeholder on left side of footer when `onBack` is undefined** _(fixed during review: removed `justify-between` + empty wrapper div; right group now uses `ms-auto` to push to the right edge, so layout is clean whether or not Back is present)_

- [ ] **Error display on `syncProgress` has no visible recovery UI** — when `completeFirstRun()` throws on the `syncProgress` step, the error paragraph renders below the step component but the footer buttons are hidden. The error is visible but there is no obvious retry path in the current placeholder. Deferred — the real `SyncProgressStep` (PT-4179) will own its error handling UI.

[Author response: Asked for fixes on minor 1–3. Minor 4 deferred for UX/PT-4179 to address since the real SyncProgressStep will have its own error recovery design.]

#### Shared Regions Modified

None.

#### Extension Config Changes

None.

## Positive Observations

- `WizardStepper` marks itself `aria-hidden="true"` and delegates screen-reader step announcements to an always-mounted `aria-live="polite"` paragraph in the shell — the live region stays mounted even on `syncProgress` (with empty content) so screen readers don't lose it during navigation.
- Button variant hierarchy is correct: Back (`outline`), Skip (`ghost`), Next/Finish (`default`) — matches the design guide's primary-accept/secondary/discard hierarchy.
- Button order is correct: Back (discard) on the left, Next (confirm) on the right.
- "Skip setup" is an outcome-named label rather than a bare "Skip."
- Sentence case used throughout all button labels and headings.
- `WizardStepper` circles use semantic color pairing: active uses `tw:bg-primary tw:text-primary-foreground`, past uses `tw:bg-muted tw:text-muted-foreground`.
- `STEP_ORDER` type tightened from `FirstRunStep[]` to `NumberedStep[]` — TypeScript now prevents `syncProgress` from being accidentally added to the ordered step list.
- `DEFAULT_LOCALE` extracted to module scope to avoid a new array allocation on every render (since `useSetting` compares by reference).
- `goToStep` resets per-step chrome (`error`, `isBusy`, `canProceed`) atomically with the step navigation, preventing stale state from bleeding across steps.
- `WizardStepper` has thorough unit tests: 6 cases covering rendering, active/past state, over/underflow clamping, locale handling, and empty-locale fallback.
- `FirstRunShell` tests substantially expanded: cover `syncProgress` interstitial path, reentrancy guard, sr-only step indicator, `setCanProceed(false)` mount-effect interaction, and entry-floor back navigation.
- Dist declaration file regenerated and in sync with source (`WizardStepper` and `WizardStepperProps` present in `index.d.ts`).
- `WizardStepperProps` uses only primitive types — straightforward for extension developers consuming the component.

## Interview Notes

**Purpose**: PT-4257 (child of PT-4174 first-run wizard epic) — add the wizard chrome (`FirstRunShell`: title, step indicator, footer navigation) and the `WizardStepper` visual circle component. Five placeholder step bodies are included; sibling tickets replace individual entries with real steps.

**Key design decisions the author explained:**

- *`syncProgress` is not in `STEP_ORDER`*: It's a post-consent interstitial, not a user-navigable step. The `NumberedStep | syncProgress` type split enforces this at compile time and prevents the back-navigation floor from being set to syncProgress.
- *`%firstRun_button_finish%` removed*: Unintentional rebase artifact — restored during review.
- *Step placeholder strings in non-English locales*: Oversight — author chose to add all five step placeholder keys to all four non-English files rather than just the new internet key.
- *Jira ticket in placeholder string*: Acknowledged as wrong; all instances cleaned to "(coming soon)".

**Minor 4 disposition**: Error display on `syncProgress` deferred — author stated the real `SyncProgressStep` (PT-4179) will own error handling; designing recovery UI around a placeholder is premature.

## In-Review Quality Check

Changes were made during the interview. Quality checks run after all fixes:

- **Library rebuild** (`npm run build:platform-bible-react`): Succeeded. TSDoc added to `WizardStepperProps` interface (required by TypeDoc); dist files regenerated and staged.
- **Typecheck**: Passed. One pre-existing error in `src/main/services/app.service-host.ts` (`Cannot find module buildInfo.json`) — present on `main`, not caused by this branch.
- **Lint**: Auto-fixed one prettier formatting issue in `first-run-shell.component.tsx` (ternary line break). Pre-existing `import/no-unresolved` for `buildInfo.json` in `app.service-host.ts` — not caused by this branch.
- **Format**: Clean after lint auto-fix.
- **Tests**: Three test cases in `first-run-shell.component.test.tsx` were clicking `/next/i` to navigate `syncConsent → syncProgress`, but the button is now labeled "Finish" — fixed to click `/finish/i`. All 14 shell tests pass. Pre-existing intermittent timeouts in `network-object-status.service.test.ts` and `network.service.test.ts` are unrelated to this branch.

## Suggested Review Focus

- [ ] **Finish label on syncConsent**: Is "Finish" the right final label, or should it be something more specific (e.g., "Start sync")? The `%firstRun_button_finish%` key was originally introduced with PT-4175 and removed accidentally — confirm the label is appropriate before merge.
- [ ] **`syncProgress` error recovery (minor 4)**: The current placeholder shows the error paragraph but no retry UI. Verify that PT-4179's `SyncProgressStep` will handle this, or decide whether a shell-level fallback is needed.
- [ ] **Step placeholder keys in non-English locales**: Confirm the decision to add all five step placeholder keys to fr/es/zh-hans/zh-hant is intentional (these are all English-language placeholder strings — they will display in English regardless of the user's language setting).

---

## Summary

Implements the wizard modal stepper (PT-4257, child of PT-4174). Adds the visual step-circle indicator component and the shell that owns all wizard chrome — step navigation, Back/Skip/Next footer, sr-only screen-reader announcements, and busy/error state for the async `completeFirstRun` call.

**Key files at a glance:**

| File | What it is |
|------|-----------|
| `lib/platform-bible-react/src/components/basics/wizard-stepper.component.tsx` | New presentational component: locale-aware numeral circles, `aria-hidden` (visual only) |
| `src/renderer/components/first-run/first-run-shell.component.tsx` | Wizard chrome: title, step indicator, WizardStepper, footer (Back/Skip/Next), navigation state machine |
| `src/renderer/services/first-run.model.ts` | Adds `NumberedStep` type split from `FirstRunStep` |
| `src/renderer/components/first-run/steps/internet.placeholder.component.tsx` | PT-4259 slot |

The four existing placeholder steps (`language`→real in PT-4176, `internet`/`identify`/`syncConsent`/`syncProgress` still placeholder) slot in via `stepComponents` so each ticket can land independently.

## Design notes

<details>
<summary>Things worth a second look</summary>

**`syncProgress` is a numbered-step exception.** It's a post-consent interstitial, so it's excluded from the numbered-step count. `NUMBERED_STEPS: NumberedStep[]` is the typed source of truth (adding an interstitial to it is a compile error), and `STEP_ORDER: FirstRunStep[]` is derived as `[...NUMBERED_STEPS, 'syncProgress']` — so the two can't drift, and an interstitial is any step at/beyond `NUMBERED_STEPS.length`. The shell hides its footer on the interstitial.

**Screen-reader accessibility is split across two elements.** `WizardStepper` is `aria-hidden` (visual circles only). A separate `aria-live="polite"` sr-only `<p>` always stays mounted — if it unmounted on `syncProgress`, screen readers would lose the live region on the next navigation. The paragraph content goes empty (`''`) rather than unmounting.

**`locale || 'en'` uses `||` not `??`.** An empty-string locale crashes `Intl.NumberFormat('')` in V8. The comment at the call site explains this.

**`isBusy` guard is ref-backed.** The `<Button disabled>` on the footer handles the shell's button, but step components call `onNext`/`onSkip` directly (e.g. the real syncProgress step calls `onNext` when sync completes). React batches state, so two calls in one tick would both read a stale `isBusy` from the render closure — the guard lives in `runAction` and is backed by `isBusyRef` (updated synchronously), so it covers both the `onNext` and `onSkip` async paths.

**`entryStep` sets the Back floor.** The startup reducer resumes post-relaunch users at `syncConsent`. `entryIndex` prevents backing into already-completed steps — important because the real `identify` step calls `platform.restart`, so backing into it would re-trigger the relaunch loop.

</details>

## Test plan

- [ ] Storybook: `First run/FirstRunShell` — all 5 entry-step stories (`Language`, `Internet`, `Identify`, `SyncConsent`, `SyncProgress`), check centering in `CenteredInModal`
- [ ] Storybook: `Basics/WizardStepper` — verify circles, active step highlight, locale variants
- [ ] Unit tests: 14 shell tests + 7 WizardStepper tests + 7 overlay tests all pass (`npm test`)

## Review-fix update

Addressed the Reviewable feedback (see per-comment replies). Notable points for reviewers:

- **`WizardStepper` API simplified**: dropped the overlapping `formatLabel` prop (locale-only now), removed the dead `locale = 'en'` default param, switched the upcoming-circle outline to `border-input`, added a `data-state` hook, and moved the component into a `basics/wizard-stepper/` folder.
- **Step order typing**: `NUMBERED_STEPS: NumberedStep[]` is now the typed source and `STEP_ORDER` is derived from it (`INTERSTITIAL_STEPS` removed). (The earlier claim that `STEP_ORDER` itself was `NumberedStep[]` was inaccurate — corrected above.)
- **Double action row (Reviewable comment on the footer)**: the shell now supports `setManagesOwnFooter` so a step that owns its footer suppresses the shell footer (no stacked rows). This is the shell half of the fix that #2618 flagged as "tracked by PT-4257". The step-side adoption for `identify`/`sync-consent` lives on **#2618 (pt-4260)** and the two must land together — this PR does not edit those sibling-owned steps.
- Removed 20 dead English placeholder localization keys from the non-English locale files; reverted an unrelated `.prettierignore` change (to be handled in its own PR); added tests for `usePrimaryInterfaceLanguage`, the `managesOwnFooter` path, and the `onSkip` re-entrancy guard.

AI-assisted

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2619)


jolierabideau (approved):
> @jolierabideau reviewed 22 files and all commit messages, and made 16 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2619)* status: all files reviewed, 16 unresolved discussions (waiting on katherinejensen00).

___
*[`.prettierignore` line 82 at r1](https://reviewable.io/reviews/paranext/paranext-core/2619#-OyiaCeaEMw4Z6jCkxh3:-OyiaCeaEMw4Z6jCkxh4:b-pww1ft) ([raw file](https://github.com/paranext/paranext-core/blob/9165e15a734edd9f7acc6e1655959f288d4a29fa/.prettierignore#L82)):*
> ```ignore
> CLAUDE.md
> 
> # AI session artifacts (gitignored; exempt from format/lint)
> ```

Claude finding

The `.prettierignore` commit doesn't do what it says**
`.prettierignore:81-85`, commit 9165e15 ("local format:check fails when session files exist on disk")

`package.json` runs Prettier with an explicit `--ignore-path .prettierignorerun` in `format:check`
(line 67), `format:core` (line 68), and every `lint-staged` entry (lines 118-123). An explicit
`--ignore-path` replaces Prettier's default list, so the three new entries are never read by
`npm run format:check` — the reported problem is still there. Confirmed:

    npx prettier --check --ignore-path .prettierignorerun docs/superpowers .review .superpowers
    → Code style issues found in 56 files.

`.claude` / `.context` / `CLAUDE.md` are in *both* ignore files; these three need the same treatment.
Two smaller problems with the same hunk:

- The comment "(gitignored; exempt from format/lint)" is wrong twice. `git ls-files docs/superpowers`
  returns 5 tracked, committed markdown files — `.gitignore` has no effect on tracked paths, so this
  silently exempts real repo docs going forward. And `.prettierignore` has no bearing on ESLint.
- It's unrelated to PT-4257. Given it needs a correctness fix anyway, I'd pull it into its own small
  PR so the fix gets reviewed on its merits.
___
*[`assets/localization/es.json` line 169 at r1](https://reviewable.io/reviews/paranext/paranext-core/2619#-Oyi_h2I2gXJ5UP-rXRs:-Oyi_h2I2gXJ5UP-rXRt:b-3kaqe5) ([raw file](https://github.com/paranext/paranext-core/blob/9165e15a734edd9f7acc6e1655959f288d4a29fa/assets/localization/es.json#L169)):*
> ```JSON
>   "%firstRun_step_language_placeholder%": "Language picker (coming soon)",
>   "%firstRun_step_syncProgress_placeholder%": "Sync progress (coming soon)",
> ```

Claude finding

The 20 added localization entries are dead, English, and partly orphaned**
`assets/localization/{es,fr,zh-hans,zh-hant}.json`

- No production code reads any `%firstRun_step_*_placeholder%` key. All five step bodies are real
  implementations now (`steps/language.component.tsx`, `internet-settings-step.component.tsx`,
  `identify-step.component.tsx`, `sync-consent-step.component.tsx`, `sync-progress.component.tsx`).
  The only repo-wide references are two test mocks (`first-run-overlay.component.test.tsx:27`,
  `setup-dialog-languages.util.test.ts:38`) and the *exclusion* regex at
  `src/extension-host/services/setup-dialog-languages.util.ts:14`.
- `%firstRun_step_{identify,internet,syncProgress}_placeholder%` don't exist in `en.json` at all.
  Localization-Guide "Blocking Issues" #2 names exactly this: keys in one language but not the other.
- `%firstRun_step_language_placeholder%` is already recorded as deprecated in
  `assets/localization/metadata.json:38` — "The Language step now has a real implementation
  (PT-4176). This placeholder string is no longer used." This PR re-adds it to four locale files.
- The values are verbatim English inside translated files. Per the guide's fallback chain
  (`BACKUP_LANGUAGE = 'en'` is always appended), that's redundant — and harmful, because an English
  value is indistinguishable from a real translation in a translator export, and "(coming soon)" is
  dev scaffolding shipping as non-English user microcopy.

No functional regression (the util deliberately excludes these from the threshold baseline), so this
is dead data rather than a bug — but the fix is to delete all 20 lines, not to add English to
`en.json`. Looks like residue the "real steps" rebases left behind.
___
*[`lib/platform-bible-react/src/components/basics/wizard-stepper.component.tsx` line 18 at r1](https://reviewable.io/reviews/paranext/paranext-core/2619#-OyicSUi2AoQKQBQ3N-X:-OyicSUj8YeVV7iDz__A:b4zibnh) ([raw file](https://github.com/paranext/paranext-core/blob/9165e15a734edd9f7acc6e1655959f288d4a29fa/lib/platform-bible-react/src/components/basics/wizard-stepper.component.tsx#L18)):*
> ```XML
>   /**
>    * Optional numeral formatter for step-circle labels. When supplied, the component skips its
>    * internal {@link NumberFormat} allocation and calls this function instead — useful when the
> ```

NIT `wizard-stepper.component.tsx:18` — `{@link NumberFormat}` is copied verbatim into
  `dist/index.d.ts:2528`, where `NumberFormat` isn't imported or declared. Dangling link in the
  generated TypeDoc; use plain `` `NumberFormat` ``.
___
*[`lib/platform-bible-react/src/components/basics/wizard-stepper.component.tsx` line 21 at r1](https://reviewable.io/reviews/paranext/paranext-core/2619#-OyiaR4d6u73cYpNu4tU:-OyiaR4d6u73cYpNu4tV:b-vpr3oa) ([raw file](https://github.com/paranext/paranext-core/blob/9165e15a734edd9f7acc6e1655959f288d4a29fa/lib/platform-bible-react/src/components/basics/wizard-stepper.component.tsx#L21)):*
> ```XML
>    * caller already holds a formatter for the same locale. Takes precedence over `locale`.
>    */
>   formatLabel?: (n: number) => string;
> ```

Claude finding

`WizardStepper` ships two overlapping ways to do one job**
`lib/platform-bible-react/src/components/basics/wizard-stepper.component.tsx:11-21`

`locale` and `formatLabel` cover the same concern, with `formatLabel` silently winning. The stated
justification (skip the internal `NumberFormat` allocation) doesn't hold — the component already
memoizes the formatter by locale at line 38, so it's one allocation per locale change either way.
This is a published `platform-bible-react` export, so removing a prop later is a breaking change, and
right now the only consumer passes `formatLabel`, leaving the `locale` path exercised only by tests
and stories. I'd drop `formatLabel`, pass `locale={locale}` from the shell, and delete the shell's
`fmt`/`formatStep`/`formattedStepCount` plumbing along with it. Note also that `locale?: string` is a
new prop convention for this library — no other component under `src/components/` has one (they take
`localizedStrings` and leave locale resolution to the caller), so worth a nod from the library owners.



_Code quote:_
```XML
  /**
   * BCP 47 locale tag for numeral formatting in the circle labels. E.g. `'ar'` → ١٢٣٤. Defaults to
   * `'en'`. Ignored when {@link WizardStepperProps.formatLabel} is provided.
   */
  locale?: string;
  /**
   * Optional numeral formatter for step-circle labels. When supplied, the component skips its
   * internal {@link NumberFormat} allocation and calls this function instead — useful when the
   * caller already holds a formatter for the same locale. Takes precedence over `locale`.
   */
  formatLabel?: (n: number) => string;
```
___
*[`lib/platform-bible-react/src/components/basics/wizard-stepper.component.tsx` line 33 at r1](https://reviewable.io/reviews/paranext/paranext-core/2619#-OyibqD93Kg3Wlzot0Ey:-OyibqD93Kg3Wlzot0Ez:b2gsrmy) ([raw file](https://github.com/paranext/paranext-core/blob/9165e15a734edd9f7acc6e1655959f288d4a29fa/lib/platform-bible-react/src/components/basics/wizard-stepper.component.tsx#L33)):*
> ```XML
>   currentStep,
>   totalSteps,
>   locale = 'en',
> ```

NIT `wizard-stepper.component.tsx:33,36` — the `locale = 'en'` default param is dead; `locale || 'en'`
  on the next line already handles `undefined` *and* `''`. Keep the `||` (it's the one the test covers).
___
*[`lib/platform-bible-react/src/components/basics/wizard-stepper.component.tsx` line 56 at r1](https://reviewable.io/reviews/paranext/paranext-core/2619#-Oyj0OWeEGJvq4ZLUItN:-Oyj0OWeEGJvq4ZLUItO:b3p0xd9) ([raw file](https://github.com/paranext/paranext-core/blob/9165e15a734edd9f7acc6e1655959f288d4a29fa/lib/platform-bible-react/src/components/basics/wizard-stepper.component.tsx#L56)):*
> ```XML
>             {stepNum > 1 && <div className="tw:h-px tw:flex-1 tw:bg-border" />}
>             <div
>               className={cn(
> ```

NIT `wizard-stepper.component.tsx:56-63` — completed and not-yet-reached circles differ only by fill
  (`bg-muted` vs `border border-muted-foreground`), both with `text-muted-foreground`, so a filled muted
  circle reads more "disabled" than "done", and progress is conveyed by color alone. A check mark for
  completed steps would fix both. Separately, `border-muted-foreground` uses a *foreground* token as a
  border color — shadcn uses `border-border`/`border-input`.
___
*[`lib/platform-bible-react/src/components/basics/wizard-stepper.component.stories.tsx` at r1](https://reviewable.io/reviews/paranext/paranext-core/2619#-Oyid55bB5K_z-W3-f1d:-Oyid55bB5K_z-W3-f1e:b-qj7y4p) ([raw file](https://github.com/paranext/paranext-core/blob/9165e15a734edd9f7acc6e1655959f288d4a29fa/lib/platform-bible-react/src/components/basics/wizard-stepper.component.stories.tsx)):*
NIT  `wizard-stepper.component.stories.tsx` is the only `*.component.stories.tsx` under
  `lib/platform-bible-react/src/components/` — the other 19 (including both neighbours in `basics/`)
  are `*.stories.tsx`. Also, per `.claude/rules/architecture/react-patterns.md` ("folder when 2+ of:
  test, story, types, utils") this should live in `basics/wizard-stepper/`.
___
*[`lib/platform-bible-react/src/components/basics/wizard-stepper.component.test.tsx` at r1](https://reviewable.io/reviews/paranext/paranext-core/2619#-OyieGfT6fEIn3GSr4ZU:-OyieGfT6fEIn3GSr4ZV:b-he89sr) ([raw file](https://github.com/paranext/paranext-core/blob/9165e15a734edd9f7acc6e1655959f288d4a29fa/lib/platform-bible-react/src/components/basics/wizard-stepper.component.test.tsx)):*
`wizard-stepper.component.test.tsx` — six of seven tests assert on Tailwind classes
  (`.tw\:rounded-full`, `tw:bg-primary`, `tw:bg-muted`), so a theme/class refactor breaks them with no
  behavior change. A `data-state="active" | "complete" | "upcoming"` attribute on each circle gives the
  tests a stable hook and is useful to consumers too.
___
*[`src/renderer/components/first-run/first-run-shell.component.tsx` line 116 at r1](https://reviewable.io/reviews/paranext/paranext-core/2619#-Oyiax86CKjYtggVsgYs:-Oyiax86CKjYtggVsgYt:bkpb1e9) ([raw file](https://github.com/paranext/paranext-core/blob/9165e15a734edd9f7acc6e1655959f288d4a29fa/src/renderer/components/first-run/first-run-shell.component.tsx#L116)):*
> ```XML
>   }, []);
> 
>   const onNext = useCallback(() => {
> ```

Claude finding

The `isBusy` re-entrancy guard isn't actually re-entrant-safe**
`first-run-shell.component.tsx:116-124`

`if (isBusy) return` reads `isBusy` out of the render closure, so two synchronous `onNext` calls in the
same tick both see `false` — which is precisely the case the comment describes ("step calls onNext
programmatically while busy"). The test passes because `await userEvent.click` flushes a render between
the two clicks. If the guard is meant to be real, back it with a ref (`isBusyRef.current`); if the
footer's `disabled` prop is considered sufficient, say so in the comment and simplify.
___
*[`src/renderer/components/first-run/first-run-shell.component.tsx` line 162 at r1](https://reviewable.io/reviews/paranext/paranext-core/2619#-OyikJnB46ongr6JXFvZ:-OyikJnB46ongr6JXFv_:b-8ahhtg) ([raw file](https://github.com/paranext/paranext-core/blob/9165e15a734edd9f7acc6e1655959f288d4a29fa/src/renderer/components/first-run/first-run-shell.component.tsx#L162)):*
> ```XML
> 
>   return (
>     <div className="tw:flex tw:min-h-screen tw:items-center tw:justify-center tw:py-8">
> ```

NIT `first-run-shell.component.tsx:162` — `tw:min-h-screen` is viewport-relative, but the only production parent is already `tw:h-screen tw:overflow-auto`. `tw:min-h-full` behaves identically there and doesn't misbehave if the shell is ever embedded in something smaller.
___
*[`src/renderer/components/first-run/first-run-shell.component.tsx` line 170 at r1](https://reviewable.io/reviews/paranext/paranext-core/2619#-Oyj0oWx7QJrkceJaRlf:-Oyj0oWx7QJrkceJaRlg:bkqt9t0) ([raw file](https://github.com/paranext/paranext-core/blob/9165e15a734edd9f7acc6e1655959f288d4a29fa/src/renderer/components/first-run/first-run-shell.component.tsx#L170)):*
> ```XML
>           {/* Keep aria-live paragraph always mounted so screen readers don't lose the live region
>               when navigating to/from syncProgress. Content is empty ('') on syncProgress. */}
>           <p className="tw:sr-only" aria-live="polite">
> ```

NIT `first-run-shell.component.tsx:170-172` — the visible "Step X of N" text is now `sr-only`, leaving the `aria-hidden` circles as the only visible progress cue. Reads intentional, but it's a user-visible reduction; worth an explicit UX sign-off.
___
*[`src/renderer/components/first-run/first-run-shell.component.tsx` line 192 at r1](https://reviewable.io/reviews/paranext/paranext-core/2619#-OyiapVU0TWisnHpMhjX:-OyiapVU0TWisnHpMhjY:b-d92b5w) ([raw file](https://github.com/paranext/paranext-core/blob/9165e15a734edd9f7acc6e1655959f288d4a29fa/src/renderer/components/first-run/first-run-shell.component.tsx#L192)):*
> ```XML
>         {error && <p className="tw:text-sm tw:text-destructive">{error}</p>}
> 
>         <div className="tw:flex tw:items-center tw:justify-between">
> ```

Claude finding

Two stacked action rows on the steps that own their primary button**
`first-run-shell.component.tsx:192-218` vs `wizard-step-form.component.tsx:32-38`

`WizardStepForm` (PT-4260) already defines exactly this row — `flex justify-between`, `<div>{backButton}</div>`,
then a right-hand group. The shell's new footer reimplements it, and the steps that call
`setCanProceed(undefined)` render their own row on top of it:

- **Sync consent** — `Sync` sits right-aligned in the body row, then the shell's row below holds `Back`
  at the far left and ghost `Skip automatic sync` at the right. The bottom-right anchor, where users
  look for the primary action, is occupied by the skip.
- **Identify** — the shell's row is `Back` alone with an empty right half.

Worth noting the split predates this PR (the old footer was `justify-end`), but moving `Back` to the
far left makes it plainly visible, and this PR owns the chrome. Either render Back/Skip through
`WizardStepForm`, or have steps hand their primary button up to the shell.
___
*[`src/renderer/components/first-run/first-run-shell.component.tsx` line 209 at r1](https://reviewable.io/reviews/paranext/paranext-core/2619#-Oyj-rdt-MX6F35BEAJm:-Oyj-rdt-MX6F35BEAJn:b-5gqxzg) ([raw file](https://github.com/paranext/paranext-core/blob/9165e15a734edd9f7acc6e1655959f288d4a29fa/src/renderer/components/first-run/first-run-shell.component.tsx#L209)):*
> ```XML
>             )}
>             {canProceed !== undefined && (
>               <Button ref={finishButtonRef} onClick={onNext} disabled={!canProceed || isBusy}>
> ```

NIT 
`first-run-shell.component.tsx:209-214` — a `Spinner` renders inside the *disabled* Finish button
  whenever `isLastStep && !canProceed`, i.e. when the user has triggered nothing and can't press it. The step body already provides inline status (`sync-progress.component.tsx` `role="progressbar"` +`role="status"`). Consider gating the spinner on `isBusy` only.
___
*[`src/renderer/components/first-run/first-run-shell.component.test.tsx` line 363 at r1](https://reviewable.io/reviews/paranext/paranext-core/2619#-Oyj0zNA0GXrxesDVpjw:-Oyj0zNA0GXrxesDVpjx:b3v6kb6) ([raw file](https://github.com/paranext/paranext-core/blob/9165e15a734edd9f7acc6e1655959f288d4a29fa/src/renderer/components/first-run/first-run-shell.component.test.tsx#L363)):*
> ```XML
>   });
> 
>   it('goes back from internet to language (entry-floor: Back is gone on the first step)', async () => {
> ```

NIT`first-run-shell.component.test.tsx:363` — `setCanProceed: setProc` swaps one opaque abbreviation for another (`Proc` = process/procedure/proceed?). `setCanProceedProp` if the alias is only there to avoid  shadowing.
___
*[`src/renderer/hooks/use-primary-interface-language.hook.ts` at r1](https://reviewable.io/reviews/paranext/paranext-core/2619#-Oyib7fvDntenj_u0UJR:-Oyib7fvDntenj_u0UJS:b9z61tk) ([raw file](https://github.com/paranext/paranext-core/blob/9165e15a734edd9f7acc6e1655959f288d4a29fa/src/renderer/hooks/use-primary-interface-language.hook.ts)):*
Claude finding

`usePrimaryInterfaceLanguage` has no test**
`src/renderer/hooks/use-primary-interface-language.hook.ts`

Both branches the JSDoc calls out deliberately — the `isPlatformError` → `'en'` path and the
`|| 'en'` empty-string path guarding `Intl.NumberFormat('')`'s `RangeError` — are unexercised;
`first-run-shell.component.test.tsx` mocks `useSetting` to `[['en'], vi.fn()]`, bypassing the hook
entirely. Sibling hooks in that directory have tests (`use-is-project-auto-sync-blocked.hook.test.ts`,
`use-project-picker-data.hook.test.ts`, `use-window-controls-overlay.hook.test.ts`). Three cases —
error, empty, normal — would lock in the documented intent.
___
*[`src/renderer/services/first-run.model.ts` line 5 at r1](https://reviewable.io/reviews/paranext/paranext-core/2619#-OyiacPO42oSYz2JRkNf:-OyiacPO42oSYz2JRkNg:b-7bawd3) ([raw file](https://github.com/paranext/paranext-core/blob/9165e15a734edd9f7acc6e1655959f288d4a29fa/src/renderer/services/first-run.model.ts#L5)):*
> ```XML
> 
> /** All wizard steps. 'syncProgress' is a post-consent interstitial — not in STEP_ORDER. */
> export type FirstRunStep = NumberedStep | 'syncProgress';
> ```

Claude finding

`NumberedStep` buys no safety, and its JSDoc states the invariant backwards**
`src/renderer/services/first-run.model.ts:2-5`, `first-run-shell.component.tsx:21-33`

The doc comment says `'syncProgress'` is "not in STEP_ORDER" — but `STEP_ORDER` (shell line 21) *does*
include it; what excludes it is the runtime `NUMBERED_STEPS` filter at line 33. Someone trusting that
comment will mis-reason about navigation. Meanwhile `NumberedStep` is exported but has zero consumers
repo-wide (its only use is composing `FirstRunStep` two lines below), and `NUMBERED_STEPS` still infers
as `FirstRunStep[]`, so the two sources of truth for "which steps are numbered" can drift with no
compile error. Either make the type load-bearing —

    export const NUMBERED_STEPS: NumberedStep[] = ['language', 'internetSettings', 'identify', 'syncConsent'];
    export const STEP_ORDER: FirstRunStep[] = [...NUMBERED_STEPS, 'syncProgress'];

which also lets `INTERSTITIAL_STEPS` go away — or drop the export and fix the comment. (Heads up: the
PR description claims `STEP_ORDER` was tightened to `NumberedStep[]`; it wasn't.)



_Code quote:_
```XML
export type NumberedStep = 'language' | 'internetSettings' | 'identify' | 'syncConsent';

/** All wizard steps. 'syncProgress' is a post-consent interstitial — not in STEP_ORDER. */
export type FirstRunStep = NumberedStep | 'syncProgress';
```

katherinejensen00 (commented):
Thank you, Jolie, for all your help with this review! Great findings

> @katherinejensen00 made 17 comments and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2619#-:-Oyo0tzJ1FZ7C_T17R_n:bvhaner)* status: 4 of 25 files reviewed, 15 unresolved discussions (waiting on jolierabideau).

___
*[`src/renderer/components/first-run/first-run-shell.component.tsx` line 116 at r1](https://reviewable.io/reviews/paranext/paranext-core/2619#-Oyiax86CKjYtggVsgYs:-Oyo0Wwh6msOizFo_enw:bpigty0) ([raw file](https://github.com/paranext/paranext-core/blob/9165e15a734edd9f7acc6e1655959f288d4a29fa/src/renderer/components/first-run/first-run-shell.component.tsx#L116)):*

Claude finding

The `isBusy` re-entrancy guard isn't actually re-entrant-safe**
`first-run-shell.component.tsx:116-124`

`if (isBusy) return` reads `isBusy` out of the render closure, so two synchronous `onNext` calls in the
same tick both see `false` — which is precisely the case the comment describes ("step calls onNext
programmatically while busy"). The test passes because `await userEvent.click` flushes a render between
the two clicks. If the guard is meant to be real, back it with a ref (`isBusyRef.current`); if the
footer's `disabled` prop is considered sufficient, say so in the comment and simplify.
</blockquote></details>

Fixed with an `isBusyRef`. I put the check inside `runAction` so it also covers `onSkip` (which had the same gap) — added tests for both the `onNext` and `onSkip` same-tick cases.
___
*[`src/renderer/components/first-run/first-run-shell.component.tsx` line 162 at r1](https://reviewable.io/reviews/paranext/paranext-core/2619#-OyikJnB46ongr6JXFvZ:-Oyo0_379xCzKxk2nOnN:b-o9qugn) ([raw file](https://github.com/paranext/paranext-core/blob/9165e15a734edd9f7acc6e1655959f288d4a29fa/src/renderer/components/first-run/first-run-shell.component.tsx#L162)):*

NIT `first-run-shell.component.tsx:162` — `tw:min-h-screen` is viewport-relative, but the only production parent is already `tw:h-screen tw:overflow-auto`. `tw:min-h-full` behaves identically there and doesn't misbehave if the shell is ever embedded in something smaller.
</blockquote></details>

Changed to `min-h-full`.
___
*[`src/renderer/components/first-run/first-run-shell.component.tsx` line 170 at r1](https://reviewable.io/reviews/paranext/paranext-core/2619#-Oyj0oWx7QJrkceJaRlf:-Oyo0b4W6bUCStbzL8Ke:b-7kyq23) ([raw file](https://github.com/paranext/paranext-core/blob/9165e15a734edd9f7acc6e1655959f288d4a29fa/src/renderer/components/first-run/first-run-shell.component.tsx#L170)):*

NIT `first-run-shell.component.tsx:170-172` — the visible "Step X of N" text is now `sr-only`, leaving the `aria-hidden` circles as the only visible progress cue. Reads intentional, but it's a user-visible reduction; worth an explicit UX sign-off.
</blockquote></details>

Restored the visible "Step X of N" count — it now doubles as the `aria-live` announcement (kept always-mounted so the live region survives navigation to/from the interstitial). Easier to pull back to sr-only later than to add, so erring toward showing it.
___
*[`src/renderer/components/first-run/first-run-shell.component.tsx` line 192 at r1](https://reviewable.io/reviews/paranext/paranext-core/2619#-OyiapVU0TWisnHpMhjX:-Oyo0dtNAk6zhfrVM3YH:b-9ydv4n) ([raw file](https://github.com/paranext/paranext-core/blob/9165e15a734edd9f7acc6e1655959f288d4a29fa/src/renderer/components/first-run/first-run-shell.component.tsx#L192)):*

Claude finding

Two stacked action rows on the steps that own their primary button**
`first-run-shell.component.tsx:192-218` vs `wizard-step-form.component.tsx:32-38`

`WizardStepForm` (PT-4260) already defines exactly this row — `flex justify-between`, `<div>{backButton}</div>`,
then a right-hand group. The shell's new footer reimplements it, and the steps that call
`setCanProceed(undefined)` render their own row on top of it:

- **Sync consent** — `Sync` sits right-aligned in the body row, then the shell's row below holds `Back`
  at the far left and ghost `Skip automatic sync` at the right. The bottom-right anchor, where users
  look for the primary action, is occupied by the skip.
- **Identify** — the shell's row is `Back` alone with an empty right half.

Worth noting the split predates this PR (the old footer was `justify-end`), but moving `Back` to the
far left makes it plainly visible, and this PR owns the chrome. Either render Back/Skip through
`WizardStepForm`, or have steps hand their primary button up to the shell.
</blockquote></details>

This is the split #2618 flagged as "tracked by PT-4257." The shell side is here: a step that owns its footer now calls `setManagesOwnFooter(true)` and the shell renders no footer of its own (Back/Skip are still handed down for the step to place in its own `WizardStepForm` row). The step-side adoption for `identify`/`sync-consent` is on #2618 and the two land together — I didn't edit those sibling-owned steps from this PR.
___
*[`src/renderer/components/first-run/first-run-shell.component.tsx` line 209 at r1](https://reviewable.io/reviews/paranext/paranext-core/2619#-Oyj-rdt-MX6F35BEAJm:-Oyo0h-j0844OzmleMCq:b-el4koy) ([raw file](https://github.com/paranext/paranext-core/blob/9165e15a734edd9f7acc6e1655959f288d4a29fa/src/renderer/components/first-run/first-run-shell.component.tsx#L209)):*

NIT 
`first-run-shell.component.tsx:209-214` — a `Spinner` renders inside the *disabled* Finish button
  whenever `isLastStep && !canProceed`, i.e. when the user has triggered nothing and can't press it. The step body already provides inline status (`sync-progress.component.tsx` `role="progressbar"` +`role="status"`). Consider gating the spinner on `isBusy` only.
</blockquote></details>

Done — the spinner is gated on `isBusy` only now; the last step surfaces its own progress (e.g. SyncProgressStep's progressbar).
___
*[`src/renderer/components/first-run/first-run-shell.component.test.tsx` line 363 at r1](https://reviewable.io/reviews/paranext/paranext-core/2619#-Oyj0zNA0GXrxesDVpjw:-Oyo0kvD8qB52j78biS0:b-vm8ivd) ([raw file](https://github.com/paranext/paranext-core/blob/9165e15a734edd9f7acc6e1655959f288d4a29fa/src/renderer/components/first-run/first-run-shell.component.test.tsx#L363)):*

NIT`first-run-shell.component.test.tsx:363` — `setCanProceed: setProc` swaps one opaque abbreviation for another (`Proc` = process/procedure/proceed?). `setCanProceedProp` if the alias is only there to avoid  shadowing.
</blockquote></details>

Renamed to `setCanProceedProp`.
___
*[`src/renderer/hooks/use-primary-interface-language.hook.ts` at r1](https://reviewable.io/reviews/paranext/paranext-core/2619#-Oyib7fvDntenj_u0UJR:-Oyo0oogCEiZPszNcSPg:b-bapxk4) ([raw file](https://github.com/paranext/paranext-core/blob/9165e15a734edd9f7acc6e1655959f288d4a29fa/src/renderer/hooks/use-primary-interface-language.hook.ts)):*

Claude finding

`usePrimaryInterfaceLanguage` has no test**
`src/renderer/hooks/use-primary-interface-language.hook.ts`

Both branches the JSDoc calls out deliberately — the `isPlatformError` → `'en'` path and the
`|| 'en'` empty-string path guarding `Intl.NumberFormat('')`'s `RangeError` — are unexercised;
`first-run-shell.component.test.tsx` mocks `useSetting` to `[['en'], vi.fn()]`, bypassing the hook
entirely. Sibling hooks in that directory have tests (`use-is-project-auto-sync-blocked.hook.test.ts`,
`use-project-picker-data.hook.test.ts`, `use-window-controls-overlay.hook.test.ts`). Three cases —
error, empty, normal — would lock in the documented intent.
</blockquote></details>

Added `use-primary-interface-language.hook.test.ts` covering exactly those three (PlatformError → `'en'`, empty-string → `'en'`, normal → first tag).
___
*[`src/renderer/services/first-run.model.ts` line 5 at r1](https://reviewable.io/reviews/paranext/paranext-core/2619#-OyiacPO42oSYz2JRkNf:-Oyo0sJH1hrKH2hl9mE6:b-jfqwoc) ([raw file](https://github.com/paranext/paranext-core/blob/9165e15a734edd9f7acc6e1655959f288d4a29fa/src/renderer/services/first-run.model.ts#L5)):*

Claude finding

`NumberedStep` buys no safety, and its JSDoc states the invariant backwards**
`src/renderer/services/first-run.model.ts:2-5`, `first-run-shell.component.tsx:21-33`

The doc comment says `'syncProgress'` is "not in STEP_ORDER" — but `STEP_ORDER` (shell line 21) *does*
include it; what excludes it is the runtime `NUMBERED_STEPS` filter at line 33. Someone trusting that
comment will mis-reason about navigation. Meanwhile `NumberedStep` is exported but has zero consumers
repo-wide (its only use is composing `FirstRunStep` two lines below), and `NUMBERED_STEPS` still infers
as `FirstRunStep[]`, so the two sources of truth for "which steps are numbered" can drift with no
compile error. Either make the type load-bearing —

    export const NUMBERED_STEPS: NumberedStep[] = ['language', 'internetSettings', 'identify', 'syncConsent'];
    export const STEP_ORDER: FirstRunStep[] = [...NUMBERED_STEPS, 'syncProgress'];

which also lets `INTERSTITIAL_STEPS` go away — or drop the export and fix the comment. (Heads up: the
PR description claims `STEP_ORDER` was tightened to `NumberedStep[]`; it wasn't.)
</blockquote></details>

Fixed both: `NUMBERED_STEPS` is now typed `NumberedStep[]` and `STEP_ORDER` is derived from it (`[...NUMBERED_STEPS, 'syncProgress']`), so the two can't drift and `INTERSTITIAL_STEPS` is gone (`isInterstitial` is now `index >= NUMBERED_STEPS.length`). Corrected the backwards JSDoc, and fixed the PR-description claim.
___
*[`.prettierignore` line 82 at r1](https://reviewable.io/reviews/paranext/paranext-core/2619#-OyiaCeaEMw4Z6jCkxh3:-Oyo-h2u2M7TD3n0GH-L:b-p1w88d) ([raw file](https://github.com/paranext/paranext-core/blob/9165e15a734edd9f7acc6e1655959f288d4a29fa/.prettierignore#L82)):*

Claude finding

The `.prettierignore` commit doesn't do what it says**
`.prettierignore:81-85`, commit 9165e15 ("local format:check fails when session files exist on disk")

`package.json` runs Prettier with an explicit `--ignore-path .prettierignorerun` in `format:check`
(line 67), `format:core` (line 68), and every `lint-staged` entry (lines 118-123). An explicit
`--ignore-path` replaces Prettier's default list, so the three new entries are never read by
`npm run format:check` — the reported problem is still there. Confirmed:

    npx prettier --check --ignore-path .prettierignorerun docs/superpowers .review .superpowers
    → Code style issues found in 56 files.

`.claude` / `.context` / `CLAUDE.md` are in *both* ignore files; these three need the same treatment.
Two smaller problems with the same hunk:

- The comment "(gitignored; exempt from format/lint)" is wrong twice. `git ls-files docs/superpowers`
  returns 5 tracked, committed markdown files — `.gitignore` has no effect on tracked paths, so this
  silently exempts real repo docs going forward. And `.prettierignore` has no bearing on ESLint.
- It's unrelated to PT-4257. Given it needs a correctness fix anyway, I'd pull it into its own small
  PR so the fix gets reviewed on its merits.
</blockquote></details>

Confirmed — `format:check`/`lint-staged` all pass `--ignore-path .prettierignorerun`, so the `.prettierignore` entries were dead. I've reverted the change out of this PR entirely
___
*[`assets/localization/es.json` line 169 at r1](https://reviewable.io/reviews/paranext/paranext-core/2619#-Oyi_h2I2gXJ5UP-rXRs:-Oyo-ub3EAvomPmLqWU7:b-v123pj) ([raw file](https://github.com/paranext/paranext-core/blob/9165e15a734edd9f7acc6e1655959f288d4a29fa/assets/localization/es.json#L169)):*

Claude finding

The 20 added localization entries are dead, English, and partly orphaned**
`assets/localization/{es,fr,zh-hans,zh-hant}.json`

- No production code reads any `%firstRun_step_*_placeholder%` key. All five step bodies are real
  implementations now (`steps/language.component.tsx`, `internet-settings-step.component.tsx`,
  `identify-step.component.tsx`, `sync-consent-step.component.tsx`, `sync-progress.component.tsx`).
  The only repo-wide references are two test mocks (`first-run-overlay.component.test.tsx:27`,
  `setup-dialog-languages.util.test.ts:38`) and the *exclusion* regex at
  `src/extension-host/services/setup-dialog-languages.util.ts:14`.
- `%firstRun_step_{identify,internet,syncProgress}_placeholder%` don't exist in `en.json` at all.
  Localization-Guide "Blocking Issues" #2 names exactly this: keys in one language but not the other.
- `%firstRun_step_language_placeholder%` is already recorded as deprecated in
  `assets/localization/metadata.json:38` — "The Language step now has a real implementation
  (PT-4176). This placeholder string is no longer used." This PR re-adds it to four locale files.
- The values are verbatim English inside translated files. Per the guide's fallback chain
  (`BACKUP_LANGUAGE = 'en'` is always appended), that's redundant — and harmful, because an English
  value is indistinguishable from a real translation in a translator export, and "(coming soon)" is
  dev scaffolding shipping as non-English user microcopy.

No functional regression (the util deliberately excludes these from the threshold baseline), so this
is dead data rather than a bug — but the fix is to delete all 20 lines, not to add English to
`en.json`. Looks like residue the "real steps" rebases left behind.
</blockquote></details>

Removed all 20 lines from fr/es/zh-hans/zh-hant. Confirmed nothing reads them at runtime (the util excludes `%firstRun_step_*_placeholder%` from the baseline), so no functional change.
___
*[`lib/platform-bible-react/src/components/basics/wizard-stepper.component.tsx` line 18 at r1](https://reviewable.io/reviews/paranext/paranext-core/2619#-OyicSUi2AoQKQBQ3N-X:-Oyo025h6PwvWeMrPwg8:b-qpfqea) ([raw file](https://github.com/paranext/paranext-core/blob/9165e15a734edd9f7acc6e1655959f288d4a29fa/lib/platform-bible-react/src/components/basics/wizard-stepper.component.tsx#L18)):*

NIT `wizard-stepper.component.tsx:18` — `{@link NumberFormat}` is copied verbatim into
  `dist/index.d.ts:2528`, where `NumberFormat` isn't imported or declared. Dangling link in the
  generated TypeDoc; use plain `` `NumberFormat` ``.
</blockquote></details>

Resolved — that JSDoc block went away with the `formatLabel` removal (comment 4), so there's no dangling link left.
___
*[`lib/platform-bible-react/src/components/basics/wizard-stepper.component.tsx` line 21 at r1](https://reviewable.io/reviews/paranext/paranext-core/2619#-OyiaR4d6u73cYpNu4tU:-Oyo-zGF72yRCI4yg-a4:beyr0mm) ([raw file](https://github.com/paranext/paranext-core/blob/9165e15a734edd9f7acc6e1655959f288d4a29fa/lib/platform-bible-react/src/components/basics/wizard-stepper.component.tsx#L21)):*

Claude finding

`WizardStepper` ships two overlapping ways to do one job**
`lib/platform-bible-react/src/components/basics/wizard-stepper.component.tsx:11-21`

`locale` and `formatLabel` cover the same concern, with `formatLabel` silently winning. The stated
justification (skip the internal `NumberFormat` allocation) doesn't hold — the component already
memoizes the formatter by locale at line 38, so it's one allocation per locale change either way.
This is a published `platform-bible-react` export, so removing a prop later is a breaking change, and
right now the only consumer passes `formatLabel`, leaving the `locale` path exercised only by tests
and stories. I'd drop `formatLabel`, pass `locale={locale}` from the shell, and delete the shell's
`fmt`/`formatStep`/`formattedStepCount` plumbing along with it. Note also that `locale?: string` is a
new prop convention for this library — no other component under `src/components/` has one (they take
`localizedStrings` and leave locale resolution to the caller), so worth a nod from the library owners.
</blockquote></details>

Agreed — dropped `formatLabel`; the component is `locale`\-only now, and the shell passes `locale={locale}` and no longer carries the `formatStep` plumbing. On the `locale?: string` convention: it's the minimal surface for internal numeral formatting here, but happy to switch to a `localizedStrings`\-style approach if the library owners prefer — flagging for their call.
___
*[`lib/platform-bible-react/src/components/basics/wizard-stepper.component.tsx` line 33 at r1](https://reviewable.io/reviews/paranext/paranext-core/2619#-OyibqD93Kg3Wlzot0Ey:-Oyo06QK5LZysCZS-W31:b-ks9ej1) ([raw file](https://github.com/paranext/paranext-core/blob/9165e15a734edd9f7acc6e1655959f288d4a29fa/lib/platform-bible-react/src/components/basics/wizard-stepper.component.tsx#L33)):*

NIT `wizard-stepper.component.tsx:33,36` — the `locale = 'en'` default param is dead; `locale || 'en'`
  on the next line already handles `undefined` *and* `''`. Keep the `||` (it's the one the test covers).
</blockquote></details>

Removed the default param; `locale || 'en'` is the single fallback (covers `undefined` and `''`).
___
*[`lib/platform-bible-react/src/components/basics/wizard-stepper.component.tsx` line 56 at r1](https://reviewable.io/reviews/paranext/paranext-core/2619#-Oyj0OWeEGJvq4ZLUItN:-Oyo0Ak49lZvBWS3HB5X:b-83vrdr) ([raw file](https://github.com/paranext/paranext-core/blob/9165e15a734edd9f7acc6e1655959f288d4a29fa/lib/platform-bible-react/src/components/basics/wizard-stepper.component.tsx#L56)):*

NIT `wizard-stepper.component.tsx:56-63` — completed and not-yet-reached circles differ only by fill
  (`bg-muted` vs `border border-muted-foreground`), both with `text-muted-foreground`, so a filled muted
  circle reads more "disabled" than "done", and progress is conveyed by color alone. A check mark for
  completed steps would fix both. Separately, `border-muted-foreground` uses a *foreground* token as a
  border color — shadcn uses `border-border`/`border-input`.
</blockquote></details>

Fixed the border token → `border-input`. Completed steps now show a check mark (via `data-state="complete"`) instead of the numeral, so "done" reads by shape, not color alone. The stepper stays `aria-hidden`; progress is announced through the shell's sr-only live region.
___
*[`lib/platform-bible-react/src/components/basics/wizard-stepper.component.test.tsx` at r1](https://reviewable.io/reviews/paranext/paranext-core/2619#-OyieGfT6fEIn3GSr4ZU:-Oyo0FYO3TtEe3zUyQXD:b-9isqco) ([raw file](https://github.com/paranext/paranext-core/blob/9165e15a734edd9f7acc6e1655959f288d4a29fa/lib/platform-bible-react/src/components/basics/wizard-stepper.component.test.tsx)):*

`wizard-stepper.component.test.tsx` — six of seven tests assert on Tailwind classes
  (`.tw\:rounded-full`, `tw:bg-primary`, `tw:bg-muted`), so a theme/class refactor breaks them with no
  behavior change. A `data-state="active" | "complete" | "upcoming"` attribute on each circle gives the
  tests a stable hook and is useful to consumers too.
</blockquote></details>

Added `data-state="active|complete|upcoming"` to each circle and retargeted the tests to it, so a theme/class refactor no longer breaks them.
___
*[`lib/platform-bible-react/src/components/basics/wizard-stepper.component.stories.tsx` at r1](https://reviewable.io/reviews/paranext/paranext-core/2619#-Oyid55bB5K_z-W3-f1d:-Oyo0LkPExXHGg2RaB0-:b-9isqco) ([raw file](https://github.com/paranext/paranext-core/blob/9165e15a734edd9f7acc6e1655959f288d4a29fa/lib/platform-bible-react/src/components/basics/wizard-stepper.component.stories.tsx)):*

NIT  `wizard-stepper.component.stories.tsx` is the only `*.component.stories.tsx` under
  `lib/platform-bible-react/src/components/` — the other 19 (including both neighbours in `basics/`)
  are `*.stories.tsx`. Also, per `.claude/rules/architecture/react-patterns.md` ("folder when 2+ of:
  test, story, types, utils") this should live in `basics/wizard-stepper/`.
</blockquote></details>

Added `data-state="active|complete|upcoming"` to each circle and retargeted the tests to it, so a theme/class refactor no longer breaks them.

jolierabideau (approved):
> @jolierabideau partially reviewed 21 files and all commit messages, and resolved 15 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2619)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2438, opened 2026-06-18
feat: add the /investigate-prd PRD-investigation capability

rolfheij-sil:
## Summary

Adds the **`/investigate-prd`** capability: a command that reads a Shape-Up PRD, fans out a read-only investigation (PT9 source archaeology for ported features, a PT10 reuse sweep always), answers or routes the open questions by audience, and writes a short brief whose centerpiece is a **proposed work-item breakdown mapped to the PRD's non-negotiables (NN-n) and nice-to-haves (NTH-n)**. Once the product owner and implementation owner agree on the brief, the companion **`/prd-to-jira`** command creates the epic + work items in Jira, behind a mandatory dry-run approval. A third sibling, **`/refine-prd`**, coaches Epic Leads through refining a PRD draft in product language before engineering ever picks it up.

Originally built in ai-prompts (`general/`, PR #329, closed in favor of this PR); moved here so it lives directly in paranext-core. Based on `main` — the embedded profile it builds on (#2436) is already merged.

*(Body updated 2026-08-05 to describe the PR as merged (`bd0fd30e9c4`) — two review rounds changed it substantially; see "How review changed it" below.)*

## What merged

- `.claude/commands/investigate-prd.md` — the thin orchestrating command: interpret → soundness gate (interactive, with a headless carve-out that carries open questions into the brief as unconfirmed assumptions) → investigate → completeness check → answer/route questions by audience → brief (seven sections) → checkpoint → optional Jira handoff. Repos are resolved from a portable `{ROOT}` convention (parent of the checkout, `PT_REPOS_ROOT` override) with an existence-based preflight that warns instead of silently degrading, and a standards-freshness probe.
- `.claude/commands/prd-to-jira.md` — creates the epic (parent issue carrying the non-negotiable ↔ ticket mapping table) plus one work item per proposed item, mirroring the shape of epic PT-4024. Gated on explicit approval of the full drafts; never emits time estimates; drafts offline when Jira is unreachable and gates on section titles, not numbers.
- `.claude/commands/refine-prd.md` — the PRD-refinement coach for Epic Leads: soundness findings, per-requirement testability, PT9-reality checks against the bundled reference, what PT10 already ships, appetite sanity — product language only.
[...]

PR #2626, opened 2026-08-03
Correct open* command declarations: parameter is the triggering web view id, not a project id

rolfheij-sil:
## Summary

The `platform-scripture.d.ts` declarations for the four inventory commands (`openCharactersInventory`, `openRepeatedWordsInventory`, `openMarkersInventory`, `openPunctuationInventory`) and `openChecksSidePanel` documented a `projectId` parameter. Every implementation actually takes the **triggering web view's id** and resolves the project internally via `papi.webViews.getOpenWebViewDefinition` — an integrator following the declared contract would pass a project id and get a silent no-op (the handler probes it as a web view id, finds nothing, and resolves `undefined`).

This renames the type-literal parameters to match the implementations (`webViewId` / `editorWebViewId`) and adds TSDoc stating the resolution semantics for each. `openManageBooks` (genuinely dual-purpose) and `openFind` were already documented correctly and are untouched.

Doc-only change: parameter names in TypeScript type literals carry no type-level meaning, so no caller can break; no runtime change; no new API items (so no new `@experimental` tags per convention).

## Origin

Identified in the PR #2438 review as the likely source of a wrong signature in that PR's authoring rule (since fixed there): https://github.com/paranext/paranext-core/pull/2438#discussion_r3693310158

This PR intentionally trails #2438 (which corrected the rule text) but merges independently of it.

https://claude.ai/code/session_01Qu9i2XxnVHwSHrc4W5vTic

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2626)

## Fix round (2026-08-04)

All five review notes addressed (`ce02eaeefab` + `7060e6fdcad`):

1-3. The five declaration blocks now carry full `@param`/`@returns` TSDoc per the Code-Style-Guide API-surface requirements; the quoted-string `{@link}` references are inlined; the `openChecksSidePanel` block states its failure mode like its siblings.
4. `openChecksSidePanel`'s runtime metadata param renamed to `editorWebViewId` — following the implementation and the `.d.ts` rather than the siblings, since the handler specifically needs an *editor* web view (scroll-group resolution). Verified the metadata feeds OpenRPC introspection only; invocation is positional.
5. `openPunctuationInventory` now registers the same `method` metadata block as its three siblings.

Bonus: the five `open*` param summaries no longer describe the argument as tied to a project — same ambiguity, same fix. (The equivalent summaries on `openMarkersChecklist` and `openFind` were left alone as out of this PR's five-command scope.)

Typecheck verified: identical error sets before/after the commits (280 pre-existing environment/branch errors, zero new).

**Addendum (2026-08-04, `959d00f9e89`):** `openFind` carried the same metadata param-name mismatch as item 4 (implementation `editorWebViewId`, metadata `webViewId`) — swept in to complete the set. Only `openMarkersChecklist`'s old-style summary now remains untouched (no name mismatch there).

### DIFF
--- a/extensions/src/platform-scripture/src/main.ts
+++ b/extensions/src/platform-scripture/src/main.ts
@@ -593,8 +593,8 @@ export async function activate(context: ExecutionActivationContext) {
         ],
         result: {
           name: 'return value',
-          summary: 'The ID of the opened markers checklist web view',
-          schema: { type: 'string' },
+          summary: 'The ID of the opened markers checklist web view, or undefined if not opened',
+          schema: { type: ['string', 'null'] },
         },
       },
     },
@@ -642,8 +642,8 @@ export async function activate(context: ExecutionActivationContext) {
         ],
         result: {
           name: 'return value',
-          summary: 'The id of the opened Manage Books web view',
-          schema: { type: 'string' },
+          summary: 'The ID of the opened Manage Books web view, or undefined if not opened',
+          schema: { type: ['string', 'null'] },
         },
       },
     },
--- a/extensions/src/platform-scripture/src/main.ts
+++ b/extensions/src/platform-scripture/src/main.ts
@@ -187,6 +187,11 @@ async function openMarkersChecklist(webViewId: string | undefined): Promise<stri
     projectId = webViewDefinition?.projectId;
   }
 
+  if (!projectId) {
+    logger.debug('No project!');
+    return undefined;
+  }
+
   const options: ChecklistWebViewOptions = { projectId };
   return papi.webViews.openWebView(
     markersChecklistWebViewType,
--- a/extensions/src/platform-scripture/src/main.ts
+++ b/extensions/src/platform-scripture/src/main.ts
@@ -659,9 +659,10 @@ export async function activate(context: ExecutionActivationContext) {
       summary: 'Open the find UI',
       params: [
         {
-          name: 'webViewId',
+          name: 'editorWebViewId',
           required: false,
-          summary: 'The ID of the web view tied to the project that we are searching in',
+          summary:
+            'The ID of the triggering editor web view; the project to search in is resolved from it',
           schema: { type: 'string' },
         },
       ],
--- a/extensions/src/platform-scripture/src/main.ts
+++ b/extensions/src/platform-scripture/src/main.ts
@@ -417,8 +417,8 @@ export async function activate(context: ExecutionActivationContext) {
         ],
         result: {
           name: 'return value',
-          summary: 'The ID of the opened characters inventory web view',
-          schema: { type: 'string' },
+          summary: 'The ID of the opened characters inventory web view, or undefined if not opened',
+          schema: { type: ['string', 'null'] },
         },
       },
     },
@@ -452,8 +452,9 @@ export async function activate(context: ExecutionActivationContext) {
         ],
         result: {
           name: 'return value',
-          summary: 'The ID of the opened repeated words inventory web view',
-          schema: { type: 'string' },
+          summary:
+            'The ID of the opened repeated words inventory web view, or undefined if not opened',
+          schema: { type: ['string', 'null'] },
         },
       },
     },
@@ -488,8 +489,8 @@ export async function activate(context: ExecutionActivationContext) {
         ],
         result: {
           name: 'return value',
-          summary: 'The ID of the new open markers inventory web view',
-          schema: { type: 'string' },
+          summary: 'The ID of the opened markers inventory web view, or undefined if not opened',
+          schema: { type: ['string', 'null'] },
         },
       },
     },
@@ -535,8 +536,9 @@ export async function activate(context: ExecutionActivationContext) {
         ],
         result: {
           name: 'return value',
-          summary: 'The ID of the opened punctuation inventory web view',
-          schema: { type: 'string' },
+          summary:
+            'The ID of the opened punctuation inventory web view, or undefined if not opened',
+          schema: { type: ['string', 'null'] },
         },
       },
     },
@@ -562,8 +564,8 @@ export async function activate(context: ExecutionActivationContext) {
         ],
         result: {
           name: 'return value',
-          summary: 'The ID of the new checks side panel web view',
-          schema: { type: 'string' },
+          summary: 'The ID of the opened checks side panel web view, or undefined if not opened',
+          schema: { type: ['string', 'null'] },
         },
       },
     },
@@ -631,7 +633,7 @@ export async function activate(context: ExecutionActivationContext) {
         summary: 'Open the unified Manage Books dialog (FN-008)',
         params: [
           {
-            name: 'projectIdOrWebViewId',
+            name: 'webViewIdOrProjectId',
             required: false,
             summary:
               'Either the active editor web view id (resolves its project) or a literal project id; omit to open with the project picker visible.',
@@ -668,8 +670,8 @@ export async function activate(context: ExecutionActivationContext) {
       ],
       result: {
         name: 'return value',
-        summary: 'The ID of the new find web view',
-        schema: { type: 'string' },
+        summary: 'The ID of the find web view, or undefined if not opened',
+        schema: { type: ['string', 'null'] },
       },
     },
   });
--- a/extensions/src/platform-scripture/src/main.ts
+++ b/extensions/src/platform-scripture/src/main.ts
@@ -410,7 +410,8 @@ export async function activate(context: ExecutionActivationContext) {
           {
             name: 'webViewId',
             required: false,
-            summary: 'The ID of the web view tied to the project that the inventory is for',
+            summary:
+              'The ID of the triggering web view; the project the inventory is for is resolved from it',
             schema: { type: 'string' },
           },
         ],
@@ -444,7 +445,8 @@ export async function activate(context: ExecutionActivationContext) {
           {
             name: 'webViewId',
             required: false,
-            summary: 'The ID of the web view tied to the project that the inventory is for',
+            summary:
+              'The ID of the triggering web view; the project the inventory is for is resolved from it',
             schema: { type: 'string' },
           },
         ],
@@ -479,7 +481,8 @@ export async function activate(context: ExecutionActivationContext) {
           {
             name: 'webViewId',
             required: false,
-            summary: 'The ID of the web view tied to the project that the inventory is for',
+            summary:
+              'The ID of the triggering web view; the project the inventory is for is resolved from it',
             schema: { type: 'string' },
           },
         ],
@@ -518,6 +521,25 @@ export async function activate(context: ExecutionActivationContext) {
   const openPunctuationInventoryPromise = papi.commands.registerCommand(
     'platformScripture.openPunctuationInventory',
     openPlatformPunctuationInventory,
+    {
+      method: {
+        summary: 'Open the punctuation inventory',
+        params: [
+          {
+            name: 'webViewId',
+            required: false,
+            summary:
+              'The ID of the triggering web view; the project the inventory is for is resolved from it',
+            schema: { type: 'string' },
+          },
+        ],
+        result: {
+          name: 'return value',
+          summary: 'The ID of the opened punctuation inventory web view',
+          schema: { type: 'string' },
+        },
+      },
+    },
   );
   const punctuationInventoryWebViewProviderPromise = papi.webViewProviders.registerWebViewProvider(
     punctuationInventoryWebViewType,
@@ -531,9 +553,10 @@ export async function activate(context: ExecutionActivationContext) {
         summary: 'Open the checks side panel',
         params: [
           {
-            name: 'webViewId',
+            name: 'editorWebViewId',
             required: false,
-            summary: 'The ID of the web view tied to the project that the checks are for',
+            summary:
+              'The ID of the editor web view the checks side panel is opened for; the project and scroll group are resolved from it',
             schema: { type: 'string' },
           },
         ],
--- a/extensions/src/platform-scripture/src/main.ts
+++ b/extensions/src/platform-scripture/src/main.ts
@@ -221,11 +221,12 @@ async function openMarkersChecklistSettings(): Promise<void> {
 }
 
 /**
- * FN-008 (2026-05-01): Open the unified Manage Books dialog as a tab web view. The optional
- * argument is either an editor's `webViewId` (from a scripture-editor menu) or a literal project id
- * — we probe with `papi.webViews.getOpenWebViewDefinition` and fall back to treating the value as a
- * project id when the probe returns `undefined`. When the caller provides no id (e.g. main-menu
- * invocation) the dialog opens with the project picker visible.
+ * FN-008 (2026-05-01): Open the unified Manage Books dialog as a centered floating window. The
+ * optional argument is either an editor's `webViewId` (from a scripture-editor menu) or a literal
+ * project id — we probe with `papi.webViews.getOpenWebViewDefinition`, and if it doesn't resolve to
+ * a web view with a project (it throws, finds nothing, or finds a web view with no project), we
+ * fall back to treating the value as a literal project id. When the caller provides no id (e.g.
+ * main-menu invocation) the dialog opens with the project picker visible.
  */
 async function openManageBooks(
   webViewIdOrProjectId: string | undefined,
--- a/extensions/src/platform-scripture/src/types/platform-scripture.d.ts
+++ b/extensions/src/platform-scripture/src/types/platform-scripture.d.ts
@@ -2391,24 +2391,52 @@ declare module 'papi-shared-types' {
 
     'platformScripture.invalidateCheckResults': (details: CheckResultsInvalidated) => Promise<void>;
 
+    /**
+     * Open the characters inventory web view. The single optional argument is the **triggering web
+     * view's id** (e.g. the editor tab the command was invoked from) — not a project id. The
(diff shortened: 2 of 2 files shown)


extensions/src/platform-scripture/src/types/platform-scripture.d.ts
```diff
@@ -2391,24 +2391,52 @@ declare module 'papi-shared-types' {
 
     'platformScripture.invalidateCheckResults': (details: CheckResultsInvalidated) => Promise<void>;
 
+    /**
+     * Open the characters inventory web view. The single optional argument is the **triggering web
```
lyonsil: **1. Missing `@param` / `@returns` on all five new blocks** (low)

`papi-shared-types` augmentations in extension `.d.ts` files are explicitly listed as API surface in [`Code-Style-Guide.md#api-surface-tsdoc-requirements`](.context/standards/Code-Style-Guide.md), which requires a `@param` per parameter and a `@returns` covering sentinels and edge cases.

Concrete effect: an integrator reading IntelliSense on these five commands still gets no statement of what the resolved `string` actually is (the id of the newly opened web view). On a PR whose sole purpose is making these declarations self-documenting, this is the convention it was meant to satisfy — worth finishing:

```ts
/**
 * Open the characters inventory web view. …
 *
 * @param webViewId Id of the triggering web view (e.g. the editor tab the command was invoked
 *   from) — not a project id. The project is resolved from it internally.
 * @returns Id of the newly opened characters inventory web view, or `undefined` if no web view id
 *   was provided or the web view has no project.
 */
```

extensions/src/platform-scripture/src/types/platform-scripture.d.ts
```diff
@@ -2391,24 +2391,52 @@ declare module 'papi-shared-types' {
...
+     * without opening anything.
+     */
     'platformScripture.openCharactersInventory': (
-      projectId?: string | undefined,
+      webViewId?: string | undefined,
     ) => Promise<string | undefined>;
 
+    /**
+     * Open the repeated-words inventory web view. Argument semantics are identical to
+     * {@link 'platformScripture.openCharactersInventory'}: the triggering web view's id, from which
```
lyonsil: **2. `{@link 'quoted string'}` is not a resolvable declaration reference** (low) — same pattern at lines 2416 and 2425.

TSDoc declaration references don't accept a quoted string literal, so this renders as inert text rather than a navigable link. It's the only such usage in the repo — every other `{@link}` in this file targets a bare symbol (e.g. `{@link ICheckRunner}`).

Combined with the style guide's "assume readers cannot look up anything outside the comment itself", the three delegating comments leave an IntelliSense reader with a dead cross-reference *and* no argument description in its place. Inlining the one sentence of semantics would be both valid and self-contained.

extensions/src/platform-scripture/src/types/platform-scripture.d.ts
```diff
@@ -2391,24 +2391,52 @@ declare module 'papi-shared-types' {
...
+     * the project is resolved internally.
+     */
     'platformScripture.openPunctuationInventory': (
-      projectId?: string | undefined,
+      webViewId?: string | undefined,
     ) => Promise<string | undefined>;
 
+    /**
+     * Open the checks side panel next to a scripture editor. The single optional argument is the
```
lyonsil: **3. This is the only one of the five blocks that omits the failure mode** (low)

`extensions/src/platform-scripture/src/main.ts:163-165`: when the id is absent or the resolved web view has no project, the handler logs and returns `undefined` without opening a panel. The other four blocks state this explicitly; a caller reading only this one would expect a panel id unconditionally.

extensions/src/platform-scripture/src/types/platform-scripture.d.ts
```diff
@@ -2391,24 +2391,52 @@ declare module 'papi-shared-types' {
...
+     * **triggering editor's web view id** — not a project id. The handler resolves the editor's
+     * project and scroll group internally via `papi.webViews.getOpenWebViewDefinition` and places
+     * the panel relative to that editor tab.
+     */
     'platformScripture.openChecksSidePanel': (
-      projectId?: string | undefined,
+      editorWebViewId?: string | undefined,
```
lyonsil, line 2457: **4. Name alignment is only half done for `openChecksSidePanel`** (low) — the actual defect is at `extensions/src/platform-scripture/src/main.ts:534`, which is outside this diff, so anchoring here.

The `.d.ts` now says `editorWebViewId`, but the runtime `registerCommand` metadata still declares `params[0].name: 'webViewId'`. That metadata — not the `.d.ts` — is what a network/JSON-RPC integrator actually introspects, so the surface this PR set out to correct stays inconsistent there.

(The four inventory commands don't have this problem; their metadata already says `webViewId` — `main.ts:411`, `445`, `480`.)

extensions/src/platform-scripture/src/types/platform-scripture.d.ts
```diff
@@ -2391,24 +2391,52 @@ declare module 'papi-shared-types' {
...
     ) => Promise<string | undefined>;
 
+    /**
+     * Open the punctuation inventory web view. Argument semantics are identical to
+     * {@link 'platformScripture.openCharactersInventory'}: the triggering web view's id, from which
+     * the project is resolved internally.
+     */
     'platformScripture.openPunctuationInventory': (
-      projectId?: string | undefined,
+      webViewId?: string | undefined,
```
lyonsil, line 2443: **5. `openPunctuationInventory` is registered with no `method` metadata at all** (low, pre-existing) — at `extensions/src/platform-scripture/src/main.ts:518-521`, outside this diff, so anchoring here.

Its three siblings each register with a full `method` block (`main.ts:406`, `440`, `475`); punctuation is registered with no options argument at all. An integrator enumerating commands over the network gets no summary, no param, and no result description for it — the same "declared contract doesn't match reality" theme this PR addresses, and a one-block fix if you want to finish the set.

lyonsil (commented):
Doc-only review of the five `open*` declarations.

**The central claim checks out.** All five handlers take a web view id and resolve the project via `papi.webViews.getOpenWebViewDefinition`, which returns `undefined` for an unknown id (`src/renderer/services/web-view.service-host.ts:1214`) rather than throwing — so the documented silent no-op is accurate. The only invocation sites are the scripture-editor tab menus, which pass `webViewId`; the runtime `registerCommand` metadata already used `webViewId`, so the `.d.ts` really was the outlier. No generated artifact mirrors these declarations, so nothing is left stale.

No correctness bugs. Five low-severity notes inline, all about the documentation surface this PR set out to fix. Two of them (items 4 and 5) are defects in `main.ts`, which isn't part of this diff — they're anchored on the nearest related `.d.ts` line with the real location called out.

extensions/src/platform-scripture/src/types/platform-scripture.d.ts
rolfheij-sil: 🤖 Claude: Fixed in `ce02eaeefab` — all five blocks now carry `@param`/`@returns` per the Code-Style-Guide requirement, with your template's semantics: web-view-id-not-project-id, internal resolution via `papi.webViews.getOpenWebViewDefinition`, and the `undefined` sentinel with the nothing-opens note.

extensions/src/platform-scripture/src/types/platform-scripture.d.ts
rolfheij-sil: 🤖 Claude: Fixed in `ce02eaeefab` — the three quoted-string `{@link}`s are inlined as one-sentence semantics; no quoted-string `{@link}` remains in the file.

extensions/src/platform-scripture/src/types/platform-scripture.d.ts
rolfheij-sil: 🤖 Claude: Fixed in `ce02eaeefab` — the `openChecksSidePanel` block now states the return-`undefined` (nothing opens) failure mode like its four siblings.

extensions/src/platform-scripture/src/types/platform-scripture.d.ts
rolfheij-sil: 🤖 Claude: Fixed in `7060e6fdcad` — metadata renamed to `editorWebViewId`, following the implementation parameter and the panel's editor-specific scroll-group requirement; its summary corrected in the same pass. Verified before renaming: the `method` block feeds OpenRPC introspection only and command invocation is positional, so the rename is purely schema-descriptive. While verifying, we found `openFind` carried the same pre-existing mismatch (implementation param `editorWebViewId`, metadata `webViewId`) — swept in via `959d00f9e89` to complete the set.

extensions/src/platform-scripture/src/types/platform-scripture.d.ts
rolfheij-sil: 🤖 Claude: Fixed in `7060e6fdcad` — `openPunctuationInventory` now registers the full sibling-matching `method` block. While in the file, the inventory commands' param summaries were also de-ambiguated ("the ID of the triggering web view; the project … is resolved from it") to close the same project-id confusion at the metadata level.

extensions/src/platform-scripture/src/types/platform-scripture.d.ts
```diff
@@ -2391,24 +2391,70 @@ declare module 'papi-shared-types' {
...
+     *   tab.
+     * @returns Id of the newly opened checks side panel web view, or `undefined` if no editor web
+     *   view id was provided or the web view has no project (nothing is opened in that case).
+     */
     'platformScripture.openChecksSidePanel': (
-      projectId?: string | undefined,
+      editorWebViewId?: string | undefined,
     ) => Promise<string | undefined>;
 
     /**
```
lyonsil, line 2460: **A. `openFind`'s `.d.ts` still claims a behavior the handler doesn't have** (low)

Line 2463, just below: "Pass `undefined` to open without an editor context." It doesn't open without one — `main.ts:290-293`:

```ts
if (!projectId) {
  logger.debug('No project!');
  return undefined;
}
```

Pre-existing, and fairly out of scope until `959d00f9e89` swept `openFind`'s metadata in to complete the set. With that commit in, this is the one remaining false statement about the very contract this PR corrects, sitting two lines from the block it corrected — same one-sentence fix as the other five.

extensions/src/platform-scripture/src/main.ts
```diff
@@ -518,6 +521,25 @@ export async function activate(context: ExecutionActivationContext) {
...
+            name: 'webViewId',
+            required: false,
+            summary:
+              'The ID of the triggering web view; the project the inventory is for is resolved from it',
+            schema: { type: 'string' },
+          },
+        ],
+        result: {
+          name: 'return value',
+          summary: 'The ID of the opened punctuation inventory web view',
```
lyonsil: **B. The `method` result blocks still promise a plain string** (low)

All six blocks this PR touches declare `summary: 'The ID of the … web view'` with `schema: { type: 'string' }`, but every handler can resolve `undefined` (`openInventory` at `main.ts:135-137`, `openChecksSidePanel` at `163-166`, `openFind` at `290-293`).

That's the same declared-contract-vs-reality gap this PR exists to close, on the surface #4 established a network integrator actually introspects: #3's fix now states the failure mode in the TS surface, while the RPC surface still says "you get an id."

There's an in-repo precedent for the accurate form — `extensions/src/platform-enhanced-resources/src/main.ts:116-120`, same open-a-web-view-from-an-editor shape:

```ts
result: {
  name: 'return value',
  summary: 'Web view id for the new Enhanced Resource web view, or undefined if not created',
  schema: { type: ['string', 'null'] },
},
```

Worth doing at minimum on this newly authored block, and ideally on the five siblings, since they're all already in the diff.

extensions/src/platform-scripture/src/main.ts
```diff
@@ -636,9 +659,10 @@ export async function activate(context: ExecutionActivationContext) {
       summary: 'Open the find UI',
       params: [
         {
-          name: 'webViewId',
+          name: 'editorWebViewId',
```
lyonsil, line 670: **C. Nit: `openManageBooks`'s metadata param name is reversed relative to its implementation** (low) — actual site is `main.ts:634`, outside this diff, so anchoring here.

Metadata says `projectIdOrWebViewId`; the implementation and the `.d.ts` both say `webViewIdOrProjectId`. Same class as #4 — word order only, and harmless since dispatch is positional. Flagging solely because `959d00f9e89` went after the whole set; skip it if you'd rather stop here.

lyonsil (commented):
Fix round reviewed — all five items are properly addressed, and the two claims worth checking both hold up.

- #4's "metadata feeds introspection only; invocation is positional" is correct: both dispatchers do `method(...params)` (`src/main/services/rpc-server.ts:193`, `src/client/services/rpc-client.ts:173`) over `RequestParams = Array<any>` (`src/shared/data/rpc.model.ts:49`). No by-name binding, no param-name-keyed logic, no OpenRPC snapshot tests — the renames are safe. `editorWebViewId` also matches existing precedent for this exact shape (`platform-enhanced-resources/src/main.ts:109`).
- The new `@param`/`@returns` text matches the handlers (`main.ts:124-145`, `147-180`), the quoted `{@link}`s are gone repo-wide, and the file is Prettier/`prettier-plugin-jsdoc` clean.

Three low-severity follow-ups below, none blocking.

rolfheij-sil:
🤖 Claude: All three follow-ups fixed, and the round's independent self-review then
extended the set: all eight open* result blocks are now `['string','null']`-honest (the
two initially excluded can also resolve `undefined` via `openWebView` itself), all eight
declarations carry full self-contained `@param`/`@returns`, and two more instances of the
"claims a behavior the handler doesn't have" class fell out — the `.d.ts` and an
implementation comment both said openManageBooks opens as a tab web view; it deliberately
opens as a centered 1100×720 float (main.ts:253). One deliberate behavior change rode
along, decided with Rolf: openMarkersChecklist gains the missing-project guard every
sibling already had — archaeology showed its projectless launch was a rejected
anti-pattern, not a feature (details in the result-schema thread). Five commits, each
gated on tsc/prettier/eslint/hooks. Per-thread details inline.

extensions/src/platform-scripture/src/types/platform-scripture.d.ts
rolfheij-sil: 🤖 Claude: Fixed in 6793226a0ea — the sentence is gone and the block is now full
`@param`/`@returns`, stating the real failure mode (no editor web view id, or no project
resolvable from it → resolves `undefined`, nothing opens), verified against
main.ts:290-293. The self-review round then tightened the same block (b62ad53e8e7): it now
documents the reuse behavior — an existing find web view is reused, reloaded only if the
resolved project differs (main.ts:313-315), and brought to front — so callers know the
returned id may belong to a pre-existing web view; and the `@param` no longer frames
resolution as menu-only, since there's a programmatic caller too (the Ctrl+F handler at
platform-scripture-editor.web-view.tsx:1170).

extensions/src/platform-scripture/src/main.ts
rolfheij-sil: 🤖 Claude: Done for the six in 6793226a0ea in the precedent form — and then extended to
all eight in b62ad53e8e7: an independent review of the fix round disproved my exclusion of
openManageBooks/openMarkersChecklist. Both end in a bare `return
papi.webViews.openWebView(...)`, and openWebView itself is declared `Promise<WebViewId |
undefined>` ("or undefined if the provider did not create a WebView",
web-view.service-model.ts:60-64), so all eight can resolve `undefined`. Wire check while
in there: `undefined` really does become `null` on the wire (serialization.ts:23-31;
rpc.model.ts:174-179 forces `result = null` for JSON-RPC compliance), so
`['string','null']` is the true wire contract; nothing validates result schemas at
runtime, consistent with #4 last round. Their .d.ts blocks also got the full
`@param`/`@returns` treatment (6636cf06841) — and one real find surfaced there:
openMarkersChecklist was the only open* handler with no missing-project guard, opening a
projectless checklist. We dug into whether that was designed: PT9's ChecklistsTool can't
launch projectless (non-null `ScrText` invariant; its project combo only *retargets* an
already-open tool), and the porting decision docs show the projectless-picker launch was
an explicitly rejected fallback from an early mainMenu wiring mistake — 0e90c701d72 fixed
the menu half but the handler guard was never added. So f0ff5286cc9 completes that fix
with the sibling-style guard (no id, or unresolvable → logs, resolves `undefined`,
nothing opens) — the one deliberate runtime change in this PR. All four markers-checklist
e2e specs invoke via the editor's own menu (always a project-bearing id), so none can
regress.

extensions/src/platform-scripture/src/main.ts
rolfheij-sil: 🤖 Claude: Fixed in 6793226a0ea — metadata now reads `webViewIdOrProjectId`, matching the
.d.ts and implementation. While completing that block's TSDoc (6636cf06841) one more
instance of item A's class fell out: the .d.ts said openManageBooks "opens the dialog as a
tab web view" — it deliberately opens as a centered floating window (main.ts:253) — fixed
there plus the last remaining comment instance (aa4d3d212e5).

lyonsil (approved):
:lgtm: - Please make sure when you squash/merge you rewrite the final commit title and details to cover the expanded scope compared to the original commit.

> @lyonsil reviewed 2 files and all commit messages, made 1 comment, and resolved 8 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2626#-:-OzMXQ8k432P0HXyaCUn:b-hn9m0z)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2627, opened 2026-08-03
Add /verify-standards mechanical doc-claim checker

rolfheij-sil:
## Motivation

A staleness audit of #2438's content (153 files, ~5,700 externally-verifiable claims) found three distinct rot classes:

1. **Status claims** ("PR #x in flight") — 100% of tracked instances had already expired
2. **Line/identifier cites** — drift with every refactor (42% of the PT9 inventory's line-cites already point into changed files)
3. **Never-true claims** — fabricated tooling/keys/thresholds that no maintenance cadence can catch later

This command is the enforcement engine for all three: a read-only mechanical sweep verifying cited repo paths, backticked code identifiers, `npm run` scripts, `%localization%` keys, in-flight status phrases, and playwright/dotnet test-config names against `origin/main` — with exemptions for frozen-record files and explicitly-aspirational claims.

**Cadence:** run before each epic's investigation phase (stale standards produce confidently wrong briefs), and on any PR adding standards content — the generation-time gate that keeps never-true claims from landing.

Test-run output re-finds the audit's confirmed fabrications (Stryker config, `test:mutation` scripts, phantom `%general_save%` keys, renamed `s_paratextProjectInterfaces`) with low noise.

Stacked on #2438 (`ai/investigate-prd-command`); rebases onto main after it merges.

https://claude.ai/code/session_01Qu9i2XxnVHwSHrc4W5vTic

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2627)

## Fix round (2026-08-04)

All 11 review items addressed in `45217b6d6d0` + `106d290bee5`:

- **1 (gate reads origin/main)** — fixed: doc text now reads from the working tree (`git ls-files` + file reads), existence checks stay on `origin/main`; Step 1 prose and the completion checklist now state this split, plus a "run from the branch you're gating" note. Verified: the sweep now scans this branch's own files (75 vs main's 74).
- **2 (silent failure)** — fixed: checked `sh_ck` for reference lookups, `FATAL` + exit 1 on unresolvable `origin/main` or empty scope (reproduced in test), and Step 2 instructs the runner to stop and report on `FATAL` rather than triaging.
- **3 (dead fence guard)** — fixed with the narrower policy, now documented in Exemptions: localization-key and path checks are suppressed inside fences; identifier, npm, dotnet, and playwright checks stay live (fenced commands are recipes readers execute, per Step 3's own instruction).
- **4 (hash sort)** — fixed: the probe extracts dates and takes the **oldest** stamp (weakest-link risk), not the newest.
- **5 (empty grep undefined)** — fixed: scope widened to all stamped surfaces + `CLAUDE.md`, and an empty result is a spelled-out "treat as unverified" warning.
- **6 (aspirational too broad)** — fixed: trailing `<!-- aspirational: … -->` marker only; the "(or prose saying so)" escape removed from the contract.
- **7 (unanchored example filter)** — fixed: exact-set match (`Old/Other/Some/My/Example/Variant/Root`) + `fullmatch` on `Foo|Bar|Baz` stems.
- **8 (in-flight prose FP)** — fixed: a status finding now requires a PR reference on the line; the `CLAUDE.md` write-gate false positive is gone.
- **9 (npm inventory)** — fixed: `extensions/src/*/package.json` included.
- **10 (frozen regex)** — fixed **with one deviation from the suggested patch**: the alternation also anchors the repo's actual canonical fence form (`> **Frozen record** — …`, the form `docs-durability.md` prescribes), which the suggested regex missed; skips now print the matched marker text.
- **11 (double-read + string sort)** — fixed: PR refs are collected from the in-hand text in the main loop and int-sorted; no per-match re-reads.

Measured on identical main-scope input: 150 → 91 candidates (fence/prose false-positive classes eliminated); wall-clock ~unchanged (identifier greps dominate — the eliminated per-ref re-reads were not the bottleneck).

**Addendum (2026-08-04, `f72b24307e5`):** the fetch line's exit status (the second half of review item 2) is now surfaced — a failed fetch over an existing stale `origin/main` prints a dated staleness WARN carried into the report and checklist.

## Squash message

Curated per `/squash-message` (review item 7/7) — paste over GitHub's pre-filled text when squash-merging. The retracted runtime claim is excluded; the fix-round "150 → 91 candidates" measurement is also left out as journey-not-merged-state (say the word to restore it as a bullet).

```text
Add /verify-standards, a read-only mechanical sweep that gates the
working tree's standards corpus — cited repo paths, backticked code
identifiers, npm scripts, dotnet test-category names, playwright
project names, localization keys, and undated in-flight status
claims — against origin/main. Existence checks (does this
path/identifier/config name actually exist) run against origin/main;
the doc text being checked is read from the working tree, so a PR's
own edits get gated, not just main. Cadence: before each epic's
investigation phase, and as the generation-time gate on PRs that add
standards content.

/investigate-prd gains a companion standards-freshness preflight
probe: it surfaces the oldest verification stamp across all stamped
surfaces (not the newest, and not a raw commit-hash sort) and
separately names any surface — .claude/rules, .claude/agents,
.claude/commands, CLAUDE.md — that has never been stamped at all,
so one fresh-looking date can't hide an unchecked surface. It
suggests running /verify-standards once the oldest stamp is older
than ~30 days.

A few invariants maintainers should know:
- An unresolvable reference or an empty verification scope FATALs
  the run and stops rather than reporting a false-clean pass; a
  disabled or empty check (no dotnet categories found, no playwright
  projects found) WARNs and announces itself instead of silently
  turning off, and the runner protocol carries WARNs as coverage
  gaps rather than clean passes. A failed fetch against a stale
  origin/main also WARNs, dated to the ref's last-fetched time.
- A fence guard suppresses localization-key and path checks inside
  fenced code blocks (those are recipes, not prose claims) while
  the npm-script, dotnet-category, playwright-project, and
  status checks stay live there; the identifier check fires on
  backticked tokens wherever they appear, including inside fences.
- Frozen-record files and explicitly-aspirational claims (marked
  with a trailing HTML comment) are exempted; an in-flight status
  claim is only flagged when its line lacks a PR reference.

Running the sweep against the standards corpus itself caught live
rot in Testing-Guide.md, fixed here:
- a playwright invocation citing `--project=development`, which has
  never existed in e2e-tests/playwright.config.ts — corrected to
  `isolated`, the standalone per-feature project the example
  actually describes.
- an invented Smoke/Critical/Full/Integration C# test-category
  scheme with time budgets nothing ever measured — replaced with the
  11 categories that actually exist in c-sharp-tests, pointing at
  .claude/skills/test-runner/categories.md instead of duplicating
  it, with corrected `dotnet test` / `npm run test:core` invocations
  and an invented feature.smoke.test.ts file-naming convention
  (consumed by nothing on main) removed.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01Qu9i2XxnVHwSHrc4W5vTic
```

### DIFF
--- a/.claude/commands/investigate-prd.md
+++ b/.claude/commands/investigate-prd.md
@@ -79,17 +79,27 @@ unreachable, **stop and warn the user before investigating**:
 
 **Preflight — standards freshness.** The stamped docs this investigation leans on carry
 `Verified against paranext-core origin/main` stamps. Find the **oldest** stamp date (the weakest
-link is what governs risk):
+link is what governs risk) **and** the surfaces that carry no stamp at all:
 
 ```bash
+# Oldest stamp date across the surfaces that have stamps
 grep -rh "Verified against paranext-core" .context/standards/ .claude/skills/ .claude/rules/ .claude/agents/ .claude/commands/ CLAUDE.md 2>/dev/null | grep -oE '20[0-9]{2}-[0-9]{2}-[0-9]{2}' | sort | head -1
+# Surfaces with no stamp at all - they contribute no date, so the line above cannot reveal them.
+# The date anchor is deliberate: it ignores prose that merely quotes the stamp phrase (this file does).
+for d in .context/standards .claude/skills .claude/rules .claude/agents .claude/commands CLAUDE.md; do
+  grep -rqE "Verified against paranext-core.*20[0-9]{2}-[0-9]{2}-[0-9]{2}" "$d" 2>/dev/null || echo "UNSTAMPED: $d"
+done
 ```
 
-If it is older than ~30 days, tell the user in one line — "standards last verified {date};
-consider running `/verify-standards` first" — and proceed. If the grep returns **nothing**, that
-is a warning, not a pass: say "no verification stamps found; treat the standards as unverified
-and consider running `/verify-standards` first" — and proceed. Don't run the verification
-yourself; it's a separate, deliberate pass ([`/verify-standards`](verify-standards.md)).
+Report **both** results in the same one-line message — an unstamped surface has never been
+verified, and averaging it away behind a fresh date is exactly the false confidence this probe
+exists to prevent: "standards last verified {date}; {list of UNSTAMPED surfaces} carry no stamps"
+(drop the second clause when nothing is unstamped). Add "consider running `/verify-standards`
+first" when the date is older than ~30 days **or** any surface is unstamped, then proceed. If the
+date grep returns **nothing**, that is a warning, not a pass: say "no verification stamps found;
+treat the standards as unverified and consider running `/verify-standards` first" — and proceed.
+Don't run the verification yourself; it's a separate, deliberate pass
+([`/verify-standards`](verify-standards.md)).
 
 An unreachable **PT10** repo degrades `pt10-reuse-scout` (affects every PRD — it always runs);
 an unreachable **`{ROOT}/Paratext`** degrades `pt9-archaeologist` (only matters if the PRD ports
--- a/.claude/commands/investigate-prd.md
+++ b/.claude/commands/investigate-prd.md
@@ -77,6 +77,13 @@ unreachable, **stop and warn the user before investigating**:
 > coverage shrinks — I'll fall back to the bundled inventory and flag the gaps. Continue
 > anyway, or fix and rerun?
 
+**Preflight — standards freshness.** The standards and skills this investigation leans on carry
+`Verified against paranext-core origin/main` stamps. Check the newest one
+(`grep -rh "Verified against paranext-core" .context/standards/ .claude/skills/ | sort | tail -1`).
+If it is older than ~30 days, tell the user in one line — "standards last verified {date};
+consider running `/verify-standards` first" — and proceed. Don't run the verification yourself;
+it's a separate, deliberate pass ([`/verify-standards`](verify-standards.md)).
+
 An unreachable **PT10** repo degrades `pt10-reuse-scout` (affects every PRD — it always runs);
 an unreachable **`{ROOT}/Paratext`** degrades `pt9-archaeologist` (only matters if the PRD ports
 PT9 features). Wait for the user. If they choose to proceed, continue and let the agents degrade
--- a/.claude/commands/investigate-prd.md
+++ b/.claude/commands/investigate-prd.md
@@ -77,12 +77,19 @@ unreachable, **stop and warn the user before investigating**:
 > coverage shrinks — I'll fall back to the bundled inventory and flag the gaps. Continue
 > anyway, or fix and rerun?
 
-**Preflight — standards freshness.** The standards and skills this investigation leans on carry
-`Verified against paranext-core origin/main` stamps. Check the newest one
-(`grep -rh "Verified against paranext-core" .context/standards/ .claude/skills/ | sort | tail -1`).
+**Preflight — standards freshness.** The stamped docs this investigation leans on carry
+`Verified against paranext-core origin/main` stamps. Find the **oldest** stamp date (the weakest
+link is what governs risk):
+
+```bash
+grep -rh "Verified against paranext-core" .context/standards/ .claude/skills/ .claude/rules/ .claude/agents/ .claude/commands/ CLAUDE.md 2>/dev/null | grep -oE '20[0-9]{2}-[0-9]{2}-[0-9]{2}' | sort | head -1
+```
+
 If it is older than ~30 days, tell the user in one line — "standards last verified {date};
-consider running `/verify-standards` first" — and proceed. Don't run the verification yourself;
-it's a separate, deliberate pass ([`/verify-standards`](verify-standards.md)).
+consider running `/verify-standards` first" — and proceed. If the grep returns **nothing**, that
+is a warning, not a pass: say "no verification stamps found; treat the standards as unverified
+and consider running `/verify-standards` first" — and proceed. Don't run the verification
+yourself; it's a separate, deliberate pass ([`/verify-standards`](verify-standards.md)).
 
 An unreachable **PT10** repo degrades `pt10-reuse-scout` (affects every PRD — it always runs);
 an unreachable **`{ROOT}/Paratext`** degrades `pt9-archaeologist` (only matters if the PRD ports
--- a/.claude/commands/verify-standards.md
+++ b/.claude/commands/verify-standards.md
@@ -26,10 +26,17 @@ frozen records or pinned-snapshot corpora with their own provenance rules.
 ## Exemptions
 
 - **Frozen records**: any in-scope file whose first 30 lines contain `Status: frozen record` or a
-  provenance fence (a blockquote naming itself a historical/point-in-time record) is skipped
-  entirely; the report lists it as skipped.
-- **Aspirational claims**: a line ending in `<!-- aspirational: not yet in repo -->` (or prose
-  saying so) is exempt from identifier/path resolution — it documents intent, not current code.
+  provenance fence — a blockquote starting `**Frozen record**` or naming itself a
+  historical/point-in-time record — is skipped entirely; the report lists each skip with the
+  marker text it matched, so an accidental freeze is visible.
+- **Aspirational claims**: a line **ending in** `<!-- aspirational: not yet in repo -->` is exempt
+  from that line's checks — it documents intent, not current code. The marker is the only
+  recognized form; prose saying "aspirational" does not exempt anything (a mechanical gate can't
+  honor prose).
+- **Fenced code blocks**: localization-key and path checks are suppressed inside ``` fences
+  (example keys and illustrative paths are legitimate there), but identifier, npm-script, dotnet
+  category, and playwright project checks stay live — a fenced command or code sample presents
+  itself as real and readers execute it, so a nonexistent name in one is a finding, not noise.
 
 ## Step 1: Refresh reference state
 
@@ -38,13 +45,17 @@ git fetch origin main --quiet
 git -C "$(git rev-parse --show-toplevel)/../Paratext" rev-parse --short HEAD 2>/dev/null || echo "PT9 checkout absent — PT9-path checks will be skipped"
 ```
 
-All existence checks below run against `origin/main` (not the working tree), so results are
-meaningful on any branch.
+The sweep reads the **doc text from the working tree** (so a PR branch's added or edited content
+is what gets gated — run it from the branch you're gating) while all **existence checks** (paths,
+scripts, identifiers, keys) run against `origin/main`, so "does this exist" answers don't depend
+on your branch state.
 
 ## Step 2: Run the mechanical sweep
 
 Run this from the repo root. It is self-contained and read-only; it prints one `findings:` line
-per candidate issue, grouped by file at the end.
+per candidate issue, grouped by file at the end. **If the sweep prints `FATAL`, stop and report
+the failure to the caller — do not triage, do not tick checklist boxes; a failed run verifies
+nothing.**
 
 ```bash
 python3 - <<'EOF'
@@ -55,8 +66,13 @@ PT9 = os.path.normpath(os.path.join(REPO, "..", "Paratext"))
 SCOPE = [".context/standards", ".claude/rules", ".claude/skills", ".claude/agents", ".claude/commands", "CLAUDE.md"]
 
 def sh(args): return subprocess.run(args, capture_output=True, text=True).stdout
+def sh_ck(args):
+    r = subprocess.run(args, capture_output=True, text=True)
+    if r.returncode != 0: raise SystemExit(f"FATAL: {' '.join(args)} failed: {r.stderr.strip()[:200]}")
+    return r.stdout
 
-main_files = set(sh(["git", "-C", REPO, "ls-tree", "-r", "--name-only", "origin/main"]).splitlines())
+main_files = set(sh_ck(["git", "-C", REPO, "ls-tree", "-r", "--name-only", "origin/main"]).splitlines())
+if not main_files: raise SystemExit("FATAL: could not read origin/main - run `git fetch origin main` (check the remote name) and rerun")
 suffix = collections.defaultdict(list)
 for p in main_files: suffix[p.rsplit("/", 1)[-1]].append(p)
 pt9_files = set()
@@ -66,7 +82,7 @@ pt9_suffix = collections.defaultdict(list)
 for p in pt9_files: pt9_suffix[p.rsplit("/", 1)[-1]].append(p)
 
 scripts = set()
-for pkg in ["package.json", "extensions/package.json"] + [p for p in main_files if p.startswith("lib/") and p.endswith("/package.json") and p.count("/") == 2]:
+for pkg in ["package.json", "extensions/package.json"] + [p for p in main_files if p.startswith("lib/") and p.endswith("/package.json") and p.count("/") == 2] + [p for p in main_files if p.startswith("extensions/src/") and p.endswith("/package.json") and p.count("/") == 3]:
     if pkg in main_files:
         try: scripts |= set(json.loads(sh(["git", "-C", REPO, "show", f"origin/main:{pkg}"])).get("scripts", {}).keys())
         except Exception: pass
@@ -88,22 +104,31 @@ RE_LOC = re.compile(r'(%[A-Za-z0-9_.]+%)')
 RE_INFLIGHT = re.compile(r'not yet merged|in[- ]flight|unmerged|awaiting merge|pending merge', re.I)
 RE_PRREF = re.compile(r'(?:\bPR\s*#|pull/|[A-Za-z][A-Za-z0-9-]*#)(\d{2,5})\b')
 RE_DATE = re.compile(r'\b20\d\d-\d\d(?:-\d\d)?\b')
-RE_ASPIRATIONAL = re.compile(r'aspirational', re.I)
-RE_FROZEN = re.compile(r'Status:\s*frozen record|historical record|point-in-time (?:record|plan)', re.I)
+RE_ASPIRATIONAL = re.compile(r'<!--\s*aspirational:[^>]*-->\s*$')
+RE_FROZEN = re.compile(r'^\s*(?:>.*)?Status:\s*frozen record|^\s*>\s*\**(?:Frozen record|.*historical record|.*point-in-time (?:record|plan))', re.I | re.M)
 SKIP_TOKENS = {"TODO", "README", "LICENSE", "ROOT", "ARGUMENTS", "SLUG", "HEAD", "PATH", "URL", "PT9_MAP", "PT9_REFERENCES", "PRD_PATH", "ASPECTS", "PT9_CLAIMS", "DEPTH", "PT_REPOS_ROOT"}
 
-def is_frozen(text): return bool(RE_FROZEN.search("\n".join(text.splitlines()[:30])))
+def frozen_marker(text):
+    m = RE_FROZEN.search("\n".join(text.splitlines()[:30]))
+    return m.group(0).strip()[:60] if m else None
 
 findings, skipped = collections.defaultdict(list), []
-targets = [f for f in main_files if any(f == s or f.startswith(s + "/") for s in SCOPE) and f.endswith(".md")]
+wt_files = sh_ck(["git", "-C", REPO, "ls-files"]).splitlines()
+targets = [f for f in wt_files if any(f == s or f.startswith(s + "/") for s in SCOPE) and f.endswith(".md")]
+if not targets: raise SystemExit(f"FATAL: no in-scope .md files found under {SCOPE}")
+pr_refs = set()
 for rel in sorted(targets):
-    text = sh(["git", "-C", REPO, "show", f"origin/main:{rel}"])
-    if is_frozen(text): skipped.append(rel); continue
+    with open(os.path.join(REPO, rel), encoding="utf-8") as fh: text = fh.read()
+    fm = frozen_marker(text)
+    if fm: skipped.append(f'{rel} - matched "{fm}"'); continue
+    pr_refs |= set(int(n) for n in RE_PRREF.findall(text))
     in_fence = False
     for i, ln in enumerate(text.splitlines(), 1):
-        if ln.strip().startswith("```"): in_fence = not in_fence
+        if ln.strip().startswith("```"):
+            in_fence = not in_fence
+            continue
         if RE_ASPIRATIONAL.search(ln): continue
-        for m in RE_PATHTOK.finditer(ln):
+        for m in RE_PATHTOK.finditer(ln) if not in_fence else []:
             p = m.group(1).lstrip("./")
             if "*" in p or "{" in p or p.startswith("http"): continue
             if "path/to" in p or "hello-world" in p or p.startswith("foo/"): continue
@@ -116,16 +141,17 @@ for rel in sorted(targets):
         for m in RE_NPM.finditer(ln):
             if m.group(1) not in scripts:
                 findings[rel].append(f":{i} [npm] `npm run {m.group(1)}` not in any package.json")
-        for m in RE_LOC.finditer(ln):
+        for m in RE_LOC.finditer(ln) if not in_fence else []:
             if m.group(1) not in loc_keys:
                 findings[rel].append(f":{i} [l10n] key `{m.group(1)}` not in en.json or contributions")
-        if RE_INFLIGHT.search(ln) and not RE_DATE.search(ln):
+        if RE_INFLIGHT.search(ln) and RE_PRREF.search(ln) and not RE_DATE.search(ln):
             findings[rel].append(f":{i} [status] undated in-flight claim: {ln.strip()[:90]}")
         for m in RE_TICK.finditer(ln):
             tok = m.group(1).strip()
             if tok in SKIP_TOKENS or tok.islower() or len(tok) < 4: continue
             if tok.startswith("$") or re.fullmatch(r'[A-Z0-9_]+', tok): continue  # shell/template placeholders
-            if re.match(r'(Foo|Bar|Baz|Old|Other|Some|My|Example|Variant|Root)[A-Z]?', tok.rsplit(".", 1)[-1].rstrip("()")) and tok.rsplit(".", 1)[-1].rstrip("()") not in ("Root",): continue  # example-name convention
+            bare_tok = tok.rsplit(".", 1)[-1].rstrip("()")
+            if bare_tok in {"Old", "Other", "Some", "My", "Example", "Variant", "Root"} or re.fullmatch(r'(Foo|Bar|Baz)[A-Za-z]*', bare_tok): continue  # example-name convention
             if not RE_IDENT.match(tok): continue
             bare = tok.rstrip("()").rsplit(".", 1)[-1]
             if len(bare) < 4 or bare.islower(): continue
@@ -140,8 +166,7 @@ for rel in sorted(targets):
                 findings[rel].append(f":{i} [playwright] project `{m}` not in playwright-cdp.config.ts")
 
 print(f"scanned {len(targets)} files; skipped {len(skipped)} frozen records: {skipped}")
-prs = sorted({m for rel in targets for m in RE_PRREF.findall(sh(['git','-C',REPO,'show',f'origin/main:{rel}'])) if not is_frozen(sh(['git','-C',REPO,'show',f'origin/main:{rel}']))})
-print(f"PR refs to spot-check with gh (state vs claim): {prs}")
+print(f"PR refs to spot-check with gh (state vs claim): {sorted(pr_refs)}")
 for rel in sorted(findings):
     print(f"\n{rel}")
     for f in findings[rel]: print(f"  {f}")
@@ -175,7 +200,8 @@ The sweep favors precision but still produces noise. For each candidate:
 
 ## Completion checklist
 
-- [ ] `git fetch origin main` ran; checks were against `origin/main`, not the working tree
+- [ ] `git fetch origin main` ran; doc text was read from the working tree, existence checks ran
+      against `origin/main`
 - [ ] Sweep script ran to completion; frozen-record skips listed
 - [ ] Every PR reference spot-checked with `gh`
 - [ ] Findings triaged into misleads-agent / wrong-detail / cosmetic with evidence
--- a/.claude/commands/verify-standards.md
+++ b/.claude/commands/verify-standards.md
@@ -34,9 +34,13 @@ frozen records or pinned-snapshot corpora with their own provenance rules.
   recognized form; prose saying "aspirational" does not exempt anything (a mechanical gate can't
   honor prose).
 - **Fenced code blocks**: localization-key and path checks are suppressed inside ``` fences
-  (example keys and illustrative paths are legitimate there), but identifier, npm-script, dotnet
-  category, and playwright project checks stay live — a fenced command or code sample presents
-  itself as real and readers execute it, so a nonexistent name in one is a finding, not noise.
+  (example keys and illustrative paths are legitimate there). The **command** checks — npm script,
+  dotnet test category, and playwright project — match raw line text, so they stay live inside
+  fences: a fenced command presents itself as runnable and readers execute it, so a nonexistent
+  script, category, or project in one is a finding, not noise. The **identifier** check works
+  differently: it only fires on backticked tokens, and code inside a fence is not backticked, so
+  identifiers are checked wherever they are backticked — in practice, prose. An identifier that
+  appears only inside a fenced code sample is not checked at all.
 
 ## Step 1: Refresh reference state
 
@@ -96,9 +100,13 @@ for lf in [p for p in main_files if p == "assets/localization/en.json" or p.ends
     try: loc_keys |= set(re.findall(r'"(%[^"%]+%)"', sh(["git", "-C", REPO, "show", f"origin/main:{lf}"])))
     except Exception: pass
 
-pw_cfg = sh(["git", "-C", REPO, "show", "origin/main:e2e-tests/playwright-cdp.config.ts"])
+pw_cfg = sh_ck(["git", "-C", REPO, "show", "origin/main:e2e-tests/playwright.config.ts"])
 pw_projects = set(re.findall(r"name:\s*['\"]([^'\"]+)['\"]", pw_cfg))
 cats = set(re.findall(r'\[Category\("([^"]+)"\)\]', sh(["git", "-C", REPO, "grep", "-h", r"\[Category(", "origin/main", "--", "c-sharp-tests"])))
+# Both inventories gate their own check (`if cats and ...` / `if pw_projects and ...`), so an empty
+# one would silently disable it. Say so instead.
+for label, inv in (("dotnet test categories (c-sharp-tests)", cats), ("playwright projects (playwright.config.ts)", pw_projects)):
+    if not inv: print(f"WARN: {label} inventory is empty - that check is disabled this run")
 
 RE_PATHTOK = re.compile(r'(?<![\w/.])((?:[A-Za-z0-9_@-]+/)+[A-Za-z0-9_@.-]+\.[A-Za-z]{1,5})(?![\w/])')
 RE_TICK = re.compile(r'`([^`\n]{2,80})`')
@@ -122,7 +130,10 @@ targets = [f for f in wt_files if any(f == s or f.startswith(s + "/") for s in S
 if not targets: raise SystemExit(f"FATAL: no in-scope .md files found under {SCOPE}")
 pr_refs = set()
 for rel in sorted(targets):
-    with open(os.path.join(REPO, rel), encoding="utf-8") as fh: text = fh.read()
+    try:
+        with open(os.path.join(REPO, rel), encoding="utf-8") as fh: text = fh.read()
+    except FileNotFoundError:
+        raise SystemExit(f"FATAL: {rel} is tracked but missing from the working tree - commit, restore, or stash before running the gate")
     fm = frozen_marker(text)
     if fm: skipped.append(f'{rel} - matched "{fm}"'); continue
     pr_refs |= set(int(n) for n in RE_PRREF.findall(text))
@@ -165,9 +176,11 @@ for rel in sorted(targets):
         for cat in re.findall(r'--filter[= ]"?Category=([A-Za-z0-9]+)', ln):
             if cats and cat not in cats:
                 findings[rel].append(f":{i} [dotnet] test category `{cat}` not found in c-sharp-tests")
-        for m in re.findall(r'--project[= ]([A-Za-z0-9-]+)', ln):
-            if pw_projects and m not in pw_projects:
-                findings[rel].append(f":{i} [playwright] project `{m}` not in playwright-cdp.config.ts")
+        if "playwright test" in ln:  # `--project` is also tsc's and dotnet's flag - don't match those
+            for m in re.findall(r'--project[= ]([A-Za-z0-9-]+)', ln):
+                if pw_projects and m not in pw_projects:
+                    findings[rel].append(f":{i} [playwright] project `{m}` not in playwright.config.ts")
+    if in_fence: print(f"WARN: {rel} ends inside an unclosed fence - path/l10n checks were suppressed from the last ``` onward")
 
 print(f"scanned {len(targets)} files; skipped {len(skipped)} frozen records: {skipped}")
 print(f"PR refs to spot-check with gh (state vs claim): {sorted(pr_refs)}")
--- a/.claude/commands/verify-standards.md
+++ b/.claude/commands/verify-standards.md
@@ -1,0 +1,182 @@
+---
+description: Mechanically verify the repo's Claude-facing docs (.context/standards, .claude rules/skills/agents/commands, CLAUDE.md) against the live code — cited paths, identifiers, npm scripts, localization keys, status claims. Read-only; prints a findings report. Run before each epic's investigation phase and on any PR that adds standards content.
+---
+
+# Verify Standards
+
+Mechanically check every externally-verifiable claim in the repo's Claude-facing documentation
+against the live repo, and report what no longer (or never did) hold. **This command is read-only
+by design**: it never edits files, never posts to GitHub or Jira, and its output is a console
+report for the caller to act on.
+
+Why it exists: a staleness audit of this content found three rot classes — status claims that
+expire in days ("PR #x in flight"), line/identifier cites that drift with refactors, and claims
+that were never true at all. The first two need a re-verification cadence; the third needs a
+**generation-time gate**: new standards claims must pass this command, or carry an explicit
+aspirational marker (see "Exemptions").
+
+## Scope
+
+Checked: `.context/standards/**`, `.claude/rules/**`, `.claude/skills/**`, `.claude/agents/**`,
+`.claude/commands/**`, and the root `CLAUDE.md`.
+
+Not checked: `.context/research/**`, `.context/designs/**`, `.context/plans/**` — these are
+frozen records or pinned-snapshot corpora with their own provenance rules.
+
+## Exemptions
+
+- **Frozen records**: any in-scope file whose first 30 lines contain `Status: frozen record` or a
+  provenance fence (a blockquote naming itself a historical/point-in-time record) is skipped
+  entirely; the report lists it as skipped.
+- **Aspirational claims**: a line ending in `<!-- aspirational: not yet in repo -->` (or prose
+  saying so) is exempt from identifier/path resolution — it documents intent, not current code.
+
+## Step 1: Refresh reference state
+
+```bash
+git fetch origin main --quiet
+git -C "$(git rev-parse --show-toplevel)/../Paratext" rev-parse --short HEAD 2>/dev/null || echo "PT9 checkout absent — PT9-path checks will be skipped"
+```
+
+All existence checks below run against `origin/main` (not the working tree), so results are
+meaningful on any branch.
+
+## Step 2: Run the mechanical sweep
+
+Run this from the repo root. It is self-contained and read-only; it prints one `findings:` line
+per candidate issue, grouped by file at the end.
+
+```bash
+python3 - <<'EOF'
+import json, os, re, subprocess, collections
+
+REPO = subprocess.run(["git", "rev-parse", "--show-toplevel"], capture_output=True, text=True).stdout.strip()
+PT9 = os.path.normpath(os.path.join(REPO, "..", "Paratext"))
+SCOPE = [".context/standards", ".claude/rules", ".claude/skills", ".claude/agents", ".claude/commands", "CLAUDE.md"]
+
+def sh(args): return subprocess.run(args, capture_output=True, text=True).stdout
+
+main_files = set(sh(["git", "-C", REPO, "ls-tree", "-r", "--name-only", "origin/main"]).splitlines())
+suffix = collections.defaultdict(list)
+for p in main_files: suffix[p.rsplit("/", 1)[-1]].append(p)
+pt9_files = set()
+if os.path.isdir(PT9):
+    pt9_files = set(sh(["git", "-C", PT9, "ls-files"]).splitlines())
+pt9_suffix = collections.defaultdict(list)
+for p in pt9_files: pt9_suffix[p.rsplit("/", 1)[-1]].append(p)
+
+scripts = set()
+for pkg in ["package.json", "extensions/package.json"] + [p for p in main_files if p.startswith("lib/") and p.endswith("/package.json") and p.count("/") == 2]:
+    if pkg in main_files:
+        try: scripts |= set(json.loads(sh(["git", "-C", REPO, "show", f"origin/main:{pkg}"])).get("scripts", {}).keys())
+        except Exception: pass
+
+loc_keys = set()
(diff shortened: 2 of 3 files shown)


lyonsil (commented):
Review of the two changed files at `4374ed9`. I extracted the embedded Python sweep and **ran it against this repo** (45s, 60 files scanned, 114 candidate findings) rather than only reading the diff, so most of what follows is empirical.

11 inline comments below, numbered `N/11` and ordered by severity in that numbering.

**Items 1-3 decide whether the command does its stated job:**

1. The sweep reads doc text from `origin/main`, so a PR's added standards content is never scanned — including this PR's own 15 branch-only files.
2. An unresolvable `origin/main` reports `0 candidate findings`, exit 0 — a failed run is indistinguishable from a clean one.
3. The fenced-code guard (`in_fence`) is dead code, which accounts for most of the 114 candidates.

**Items 4-11** are correctable in place: hash-vs-date sorting and an undefined empty case in the `/investigate-prd` preflight, and six regex/inventory issues in the sweep where the implementation is looser or narrower than the documented contract.

Nothing here argues against the approach — the sweep found real rot in `CLAUDE.md` and the standards on its first run, which is the point.

.claude/commands/verify-standards.md
```diff
@@ -0,0 +1,182 @@
...
+RE_PRREF = re.compile(r'(?:\bPR\s*#|pull/|[A-Za-z][A-Za-z0-9-]*#)(\d{2,5})\b')
+RE_DATE = re.compile(r'\b20\d\d-\d\d(?:-\d\d)?\b')
+RE_ASPIRATIONAL = re.compile(r'aspirational', re.I)
+RE_FROZEN = re.compile(r'Status:\s*frozen record|historical record|point-in-time (?:record|plan)', re.I)
+SKIP_TOKENS = {"TODO", "README", "LICENSE", "ROOT", "ARGUMENTS", "SLUG", "HEAD", "PATH", "URL", "PT9_MAP", "PT9_REFERENCES", "PRD_PATH", "ASPECTS", "PT9_CLAIMS", "DEPTH", "PT_REPOS_ROOT"}
+
+def is_frozen(text): return bool(RE_FROZEN.search("\n".join(text.splitlines()[:30])))
+
+findings, skipped = collections.defaultdict(list), []
+targets = [f for f in main_files if any(f == s or f.startswith(s + "/") for s in SCOPE) and f.endswith(".md")]
```
lyonsil: **1/11 — blocking: the sweep can never see the content it is meant to gate.**

Both the file list (this line) and the file text (line 100) come from `origin/main`. Step 1's rationale at lines 41-42 is sound for the *right-hand* side of every check (does this path/script/identifier exist?), but it got applied to the *left-hand* side too (which prose do we read?).

That breaks the headline purpose at lines 172-174: "Run it on any PR that adds or edits standards/rules/skills content: new claims must come out clean." A PR's added content is by definition not on `origin/main`, so it is never read. Measured on this PR: the head branch has 75 in-scope `.md` files, `origin/main` has 60 — the 15 branch-only files are skipped, including `verify-standards.md` itself. The gate reports clean, and checklist line 178 ("checks were against `origin/main`, not the working tree") certifies the very thing that broke it.

Split the two sides:

```python
targets = [f for f in sh(["git", "-C", REPO, "ls-files"]).splitlines() if <same scope filter>]
...
with open(os.path.join(REPO, rel), encoding="utf-8") as fh: text = fh.read()
```

Leave `main_files`, `suffix`, `scripts`, `loc_keys`, `cats`, `pw_projects` and the `git grep origin/main` identifier lookups exactly as they are, then reword checklist line 178 to what is actually true: doc text from the working tree, existence checks against `origin/main`.

.claude/commands/verify-standards.md
```diff
@@ -0,0 +1,182 @@
...
+python3 - <<'EOF'
+import json, os, re, subprocess, collections
+
+REPO = subprocess.run(["git", "rev-parse", "--show-toplevel"], capture_output=True, text=True).stdout.strip()
+PT9 = os.path.normpath(os.path.join(REPO, "..", "Paratext"))
+SCOPE = [".context/standards", ".claude/rules", ".claude/skills", ".claude/agents", ".claude/commands", "CLAUDE.md"]
+
+def sh(args): return subprocess.run(args, capture_output=True, text=True).stdout
+
+main_files = set(sh(["git", "-C", REPO, "ls-tree", "-r", "--name-only", "origin/main"]).splitlines())
```
lyonsil: **2/11 — blocking: a failed reference lookup produces a clean bill of health.**

`sh()` (line 57) returns stdout only — no `check=True`, no stderr, no returncode — and the `git fetch origin main --quiet` on line 37 is a bare bash line whose exit status nobody reads. Every downstream set derives from `main_files`, so one unresolvable ref collapses the sweep into silence rather than an error: `targets` empty, loop body never runs, and it prints

```
scanned 0 files; skipped 0 frozen records: []
PR refs to spot-check with gh (state vs claim): []

0 candidate findings
```

exit 0. I reproduced this by pointing the extracted script at a nonexistent ref. Realistic triggers: a fork whose upstream remote isn't named `origin`, a shallow clone, CI with no network, a sandbox that blocks the fetch. It's the worst failure mode available to a gate — indistinguishable from success, and the operator then ticks all five checklist boxes on a run that verified nothing.

```python
if not main_files:
    raise SystemExit("FATAL: could not read origin/main - run `git fetch origin main` (check the remote name) and rerun")
if not targets:
    raise SystemExit(f"FATAL: no in-scope .md files found under {SCOPE}")
```

Also worth a checked `sh` variant (`subprocess.run(..., check=True)`) for calls where empty output is meaningless, keeping the tolerant one only where empty is a legitimate answer (`git grep` misses, absent PT9 checkout).

.claude/commands/verify-standards.md
```diff
@@ -0,0 +1,182 @@
...
+SKIP_TOKENS = {"TODO", "README", "LICENSE", "ROOT", "ARGUMENTS", "SLUG", "HEAD", "PATH", "URL", "PT9_MAP", "PT9_REFERENCES", "PRD_PATH", "ASPECTS", "PT9_CLAIMS", "DEPTH", "PT_REPOS_ROOT"}
+
+def is_frozen(text): return bool(RE_FROZEN.search("\n".join(text.splitlines()[:30])))
+
+findings, skipped = collections.defaultdict(list), []
+targets = [f for f in main_files if any(f == s or f.startswith(s + "/") for s in SCOPE) and f.endswith(".md")]
+for rel in sorted(targets):
+    text = sh(["git", "-C", REPO, "show", f"origin/main:{rel}"])
+    if is_frozen(text): skipped.append(rel); continue
+    in_fence = False
```
lyonsil, line 146: **3/11 — blocking: the fenced-code guard is dead code.**

`in_fence` is initialized here and toggled on line 104, and never read — two references in the whole script. So every fenced example is linted as prose, which is the dominant noise source in the real 60-file run (114 candidates): `%foo%`, `%key%`, `%s%`, `%n%`, `%setting_label%`, `%oldKey_thatNeedsReplacement%`, `%APPDATA%` reported as missing l10n keys, plus `[path]` findings for `bible/main.log` and `XDG_CONFIG_HOME/Electron/logs/main.log`. All inside fences. That undercuts the low-noise claim and Step 3's premise that the output "favors precision."

Secondary: the toggle fires *before* the line is processed, so once wired up the opening ``` line would be classed as inside the fence — correct by accident today, but you want an explicit `continue`.

The bigger question is whether to restore it at all. The `[ident]` check is arguably *most* valuable inside fenced examples, and Step 3 (lines 158-159) explicitly says "if an example presents itself as real code, a nonexistent identifier is a finding, not noise" — blanket fence-skipping contradicts that instruction. Narrower fix: suppress only the checks that are meaningless in code, keep `[ident]` live.

```python
if ln.strip().startswith("```"):
    in_fence = not in_fence
    continue
...
if not in_fence:
    for m in RE_LOC.finditer(ln): ...
    for m in RE_PATHTOK.finditer(ln): ...
```

Either way, decide it deliberately and document it — the Exemptions section doesn't mention fences at all.

.claude/commands/investigate-prd.md
```diff
@@ -77,6 +77,13 @@ unreachable, **stop and warn the user before investigating**:
 > coverage shrinks — I'll fall back to the bundled inventory and flag the gaps. Continue
 > anyway, or fix and rerun?
 
+**Preflight — standards freshness.** The standards and skills this investigation leans on carry
+`Verified against paranext-core origin/main` stamps. Check the newest one
+(`grep -rh "Verified against paranext-core" .context/standards/ .claude/skills/ | sort | tail -1`).
```
lyonsil: **4/11 — the freshness check sorts by commit hash, not date.**

The pipeline assumes the date is the sort key. The actual stamp format is:

```
> Verified against paranext-core origin/main `998ca09a087` — 2026-08-03.
```

After the constant prefix the next varying field is an 11-char hex hash, so `sort` orders by hash and the date is never consulted. Both directions fail: docs re-verified today at `0ab…` alongside stale ones at `998…` returns the stale line (spurious warning); the mirror case — old work at `fff…`, fresh at `0ab…` — reports the docs fresh when they aren't, which is the direction that hurts, since the point is to stop `/investigate-prd` building a brief on rotten standards.

```bash
grep -rh "Verified against paranext-core" .context/standards/ .claude/skills/ \
  | grep -oE '20[0-9]{2}-[0-9]{2}-[0-9]{2}' | sort | tail -1
```

Also reconsider `tail -1`: "newest stamp" answers a question nobody has. Risk is governed by the weakest link, so `head -1` (oldest stamp) is the number that should drive the 30-day warning.

.claude/commands/investigate-prd.md
```diff
@@ -77,6 +77,13 @@ unreachable, **stop and warn the user before investigating**:
...
+If it is older than ~30 days, tell the user in one line — "standards last verified {date};
```
lyonsil: **5/11 — no defined behavior when the grep returns nothing.**

The instruction branches only on "If it is older than ~30 days," so empty output leaves the agent to proceed silently, exactly as if the standards had just been verified. Not hypothetical: `origin/main` today contains the string `Verified against paranext-core` **zero** times, so any branch cut from main runs this preflight, gets nothing, and continues without a word. Same after any doc reorg that moves files out of the two grepped directories.

Spell the case out as a warning, not a pass — e.g. "no verification stamps found; treat the standards as unverified and consider `/verify-standards` first."

Related scope gap on the line above: the grep covers `.context/standards/` and `.claude/skills/`, while `/verify-standards` scopes `.claude/rules`, `.claude/agents`, `.claude/commands` and `CLAUDE.md` as well. Those four can rot indefinitely without ever moving the date this preflight reports.

.claude/commands/verify-standards.md
```diff
@@ -0,0 +1,182 @@
...
+
+RE_PATHTOK = re.compile(r'(?<![\w/.])((?:[A-Za-z0-9_@-]+/)+[A-Za-z0-9_@.-]+\.[A-Za-z]{1,5})(?![\w/])')
+RE_TICK = re.compile(r'`([^`\n]{2,80})`')
+RE_IDENT = re.compile(r'^[A-Za-z_$][A-Za-z0-9_$]*(?:\(\))?$|^[A-Za-z_$][A-Za-z0-9_$]*(?:\.[A-Za-z_$][A-Za-z0-9_$]*)+$')
+RE_NPM = re.compile(r'\bnpm run ([A-Za-z0-9:_-]+)')
+RE_LOC = re.compile(r'(%[A-Za-z0-9_.]+%)')
+RE_INFLIGHT = re.compile(r'not yet merged|in[- ]flight|unmerged|awaiting merge|pending merge', re.I)
+RE_PRREF = re.compile(r'(?:\bPR\s*#|pull/|[A-Za-z][A-Za-z0-9-]*#)(\d{2,5})\b')
+RE_DATE = re.compile(r'\b20\d\d-\d\d(?:-\d\d)?\b')
+RE_ASPIRATIONAL = re.compile(r'aspirational', re.I)
```
lyonsil: **6/11 — the aspirational exemption is far broader than documented.**

A bare case-insensitive `aspirational` anywhere on a line, and the `continue` at line 105 sits at the top of the per-line body — so it disables *all* of that line's checks: path, npm, l10n, status, identifier. The documented contract (lines 31-32) is a trailing `<!-- aspirational: not yet in repo -->` marker.

So a line reading "This is not aspirational — `FooProvider` ships today, see `src/.../foo.provider.ts`" is fully exempt. Self-defeating too: prose *describing* the exemption disables checking on its own line.

```python
RE_ASPIRATIONAL = re.compile(r'<!--\s*aspirational:[^>]*-->\s*$')
```

The "(or prose saying so)" escape hatch in lines 31-32 should go at the same time — a mechanical gate can't honor it without exactly this failure.

.claude/commands/verify-standards.md
```diff
@@ -0,0 +1,182 @@
...
+        for m in RE_LOC.finditer(ln):
+            if m.group(1) not in loc_keys:
+                findings[rel].append(f":{i} [l10n] key `{m.group(1)}` not in en.json or contributions")
+        if RE_INFLIGHT.search(ln) and not RE_DATE.search(ln):
+            findings[rel].append(f":{i} [status] undated in-flight claim: {ln.strip()[:90]}")
+        for m in RE_TICK.finditer(ln):
+            tok = m.group(1).strip()
+            if tok in SKIP_TOKENS or tok.islower() or len(tok) < 4: continue
+            if tok.startswith("$") or re.fullmatch(r'[A-Z0-9_]+', tok): continue  # shell/template placeholders
+            if re.match(r'(Foo|Bar|Baz|Old|Other|Some|My|Example|Variant|Root)[A-Z]?', tok.rsplit(".", 1)[-1].rstrip("()")) and tok.rsplit(".", 1)[-1].rstrip("()") not in ("Root",): continue  # example-name convention
```
lyonsil: **7/11 — the example-name filter is an unanchored prefix match, and it drops real identifiers.**

`re.match` anchors at the start only, `[A-Z]?` is optional, and there's no `\b` or `$` — so anything merely *beginning* with one of those strings is exempted: `OldTestamentBooks`, `SomeoneService`, `MyProjectDataProvider`, `Variants`, `ExampleDataProvider`, `Others`, `RootLayout` (the `("Root",)` guard only rescues the exact token `Root`).

Severity comes from lines 172-174, which name never-true claims as "the one rot class no maintenance cadence can catch later." A false negative here is therefore permanent — a fabricated `OldProjectSettings` cite passes clean on every future run. That's strictly worse than a false positive, which Step 3 triage catches in seconds.

```python
EXAMPLE_EXACT = {"Old", "Other", "Some", "My", "Example", "Variant", "Root"}
bare_tok = tok.rsplit(".", 1)[-1].rstrip("()")
if bare_tok in EXAMPLE_EXACT or re.fullmatch(r'(Foo|Bar|Baz)[A-Za-z]*', bare_tok): continue
```

.claude/commands/verify-standards.md
```diff
@@ -0,0 +1,182 @@
...
+pw_cfg = sh(["git", "-C", REPO, "show", "origin/main:e2e-tests/playwright-cdp.config.ts"])
+pw_projects = set(re.findall(r"name:\s*['\"]([^'\"]+)['\"]", pw_cfg))
+cats = set(re.findall(r'\[Category\("([^"]+)"\)\]', sh(["git", "-C", REPO, "grep", "-h", r"\[Category(", "origin/main", "--", "c-sharp-tests"])))
+
+RE_PATHTOK = re.compile(r'(?<![\w/.])((?:[A-Za-z0-9_@-]+/)+[A-Za-z0-9_@.-]+\.[A-Za-z]{1,5})(?![\w/])')
+RE_TICK = re.compile(r'`([^`\n]{2,80})`')
+RE_IDENT = re.compile(r'^[A-Za-z_$][A-Za-z0-9_$]*(?:\(\))?$|^[A-Za-z_$][A-Za-z0-9_$]*(?:\.[A-Za-z_$][A-Za-z0-9_$]*)+$')
+RE_NPM = re.compile(r'\bnpm run ([A-Za-z0-9:_-]+)')
+RE_LOC = re.compile(r'(%[A-Za-z0-9_.]+%)')
+RE_INFLIGHT = re.compile(r'not yet merged|in[- ]flight|unmerged|awaiting merge|pending merge', re.I)
```
lyonsil, line 122: **8/11 — `RE_INFLIGHT` fires on unrelated prose, permanently.**

`in[- ]flight` matches the ordinary English phrase with no requirement that the line also reference a PR. The real run produced:

```
CLAUDE.md:177 [status] undated in-flight claim: in-flight write count, and an arm generation…
```

That's the Send/Receive write-gate description — a concurrency term, not a status claim — and it will be re-emitted on every run until someone rewords `CLAUDE.md` to appease the linter. Recurring known-noise findings are how a report stops being read.

`RE_PRREF` on the next line already recognizes the thing that makes a status claim a status claim:

```python
if RE_INFLIGHT.search(ln) and RE_PRREF.search(ln) and not RE_DATE.search(ln):
```

.claude/commands/verify-standards.md
```diff
@@ -0,0 +1,182 @@
...
+suffix = collections.defaultdict(list)
+for p in main_files: suffix[p.rsplit("/", 1)[-1]].append(p)
+pt9_files = set()
+if os.path.isdir(PT9):
+    pt9_files = set(sh(["git", "-C", PT9, "ls-files"]).splitlines())
+pt9_suffix = collections.defaultdict(list)
+for p in pt9_files: pt9_suffix[p.rsplit("/", 1)[-1]].append(p)
+
+scripts = set()
+for pkg in ["package.json", "extensions/package.json"] + [p for p in main_files if p.startswith("lib/") and p.endswith("/package.json") and p.count("/") == 2]:
```
lyonsil: **9/11 — the npm inventory omits extension packages, but the finding text claims otherwise.**

Root, `extensions/`, and `lib/*/` only — the per-extension manifests under `extensions/src/*/package.json` are never read. The finding emitted at line 118 nevertheless asserts `not in any package.json`, which is a false statement inside a tool whose whole job is catching false statements in docs.

Confirmed against `origin/main`: `start:production` is defined *only* in `extensions/src/*/package.json`, so any doc citing `npm run start:production` gets reported as broken.

```python
+ [p for p in main_files if p.startswith("extensions/src/") and p.endswith("/package.json") and p.count("/") == 3]
```

Extending beats reweording the message — the docs do cite extension scripts.

.claude/commands/verify-standards.md
```diff
@@ -0,0 +1,182 @@
...
+RE_PATHTOK = re.compile(r'(?<![\w/.])((?:[A-Za-z0-9_@-]+/)+[A-Za-z0-9_@.-]+\.[A-Za-z]{1,5})(?![\w/])')
+RE_TICK = re.compile(r'`([^`\n]{2,80})`')
+RE_IDENT = re.compile(r'^[A-Za-z_$][A-Za-z0-9_$]*(?:\(\))?$|^[A-Za-z_$][A-Za-z0-9_$]*(?:\.[A-Za-z_$][A-Za-z0-9_$]*)+$')
+RE_NPM = re.compile(r'\bnpm run ([A-Za-z0-9:_-]+)')
+RE_LOC = re.compile(r'(%[A-Za-z0-9_.]+%)')
+RE_INFLIGHT = re.compile(r'not yet merged|in[- ]flight|unmerged|awaiting merge|pending merge', re.I)
+RE_PRREF = re.compile(r'(?:\bPR\s*#|pull/|[A-Za-z][A-Za-z0-9-]*#)(\d{2,5})\b')
+RE_DATE = re.compile(r'\b20\d\d-\d\d(?:-\d\d)?\b')
+RE_ASPIRATIONAL = re.compile(r'aspirational', re.I)
+RE_FROZEN = re.compile(r'Status:\s*frozen record|historical record|point-in-time (?:record|plan)', re.I)
```
lyonsil: **10/11 — `RE_FROZEN` can silently disable a whole file on an incidental phrase.**

The documented contract (lines 28-30) is `Status: frozen record` **or a provenance blockquote**. The regex implements the first literally but the second as bare unanchored phrases anywhere in the first 30 lines. Any file whose intro happens to say "commit history is a historical record" — plausible in `Git-Guide.md` — has every check skipped.

The only trace is membership in the `skipped` list, which line 142 prints as a raw Python list with no reason attached, so a reader can't distinguish a deliberate freeze from an accidental one.

```python
RE_FROZEN = re.compile(r'^\s*(?:>.*)?Status:\s*frozen record|^\s*>.*(?:historical record|point-in-time (?:record|plan))', re.I | re.M)
```

And print `skipped` as `rel — matched "<marker text>"` lines rather than a bare list.

.claude/commands/verify-standards.md
```diff
@@ -0,0 +1,182 @@
...
+                findings[rel].append(f":{i} [ident] `{tok}` grep-resolves nowhere")
+        for cat in re.findall(r'--filter[= ]"?Category=([A-Za-z0-9]+)', ln):
+            if cats and cat not in cats:
+                findings[rel].append(f":{i} [dotnet] test category `{cat}` not found in c-sharp-tests")
+        for m in re.findall(r'--project[= ]([A-Za-z0-9-]+)', ln):
+            if pw_projects and m not in pw_projects:
+                findings[rel].append(f":{i} [playwright] project `{m}` not in playwright-cdp.config.ts")
+
+print(f"scanned {len(targets)} files; skipped {len(skipped)} frozen records: {skipped}")
+prs = sorted({m for rel in targets for m in RE_PRREF.findall(sh(['git','-C',REPO,'show',f'origin/main:{rel}'])) if not is_frozen(sh(['git','-C',REPO,'show',f'origin/main:{rel}']))})
```
lyonsil: **11/11 — PR-ref collection re-reads every file twice, and sorts numbers as strings.**

`git show` runs once in the generator and again in the `if not is_frozen(sh(...))` guard — and the guard is evaluated *per matched ref*, so a file with five PR references is fetched and re-scanned five extra times. This sits in the slowest phase of a run that already takes ~45s, and the text is already in hand inside the loop above.

Separately, `sorted()` over strings printed `['210', '2219', '2590']` in the real run. It reads plausibly only because those three happen to sort that way; add a two-digit ref and `'99'` lands after `'2590'`.

```python
prs = set()
...  # inside the per-file loop, after the is_frozen check
prs |= set(RE_PRREF.findall(text))
...
print(f"PR refs to spot-check with gh (state vs claim): {sorted(prs, key=int)}")
```

.claude/commands/verify-standards.md
rolfheij-sil: 🤖 Claude: Fixed in `45217b6d6d0` — doc text now comes from the working tree (`git ls-files` + `open()`); every existence check stays on `origin/main` as you specified. The checklist line is reworded to that true split, and Step 1 gained "run from the branch you're gating." Extraction test confirms: the sweep now scans the branch's own added files.

.claude/commands/verify-standards.md
rolfheij-sil: 🤖 Claude: Fixed in `45217b6d6d0` — checked `sh` variant plus both FATALs as suggested, and Step 2 now instructs the runner to stop and report on FATAL rather than triage. Verified: a bad ref exits 1 with the message. The other half of your finding — the fetch line's unread exit status — is covered in `f72b24307e5`: a failed fetch over an existing stale ref now prints a dated staleness WARN that the checklist carries into the report.

.claude/commands/verify-standards.md
rolfheij-sil: 🤖 Claude: Fixed in `45217b6d6d0` with your narrower option, deliberately extended one notch: `l10n`/`path` checks are suppressed inside fences, but `ident`, `npm`, `dotnet`, and `playwright` checks stay live there — fenced commands are recipes readers execute, so a fabricated script name in a fence is a finding, per Step 3's own instruction. The policy is documented in Exemptions. Measured effect: the fence and undated-status noise classes are gone (150 → 91 candidates on identical input: −53 fenced l10n/path, −9 undated-status, +3 from the tightened filters).

.claude/commands/investigate-prd.md
rolfheij-sil: 🤖 Claude: Fixed in `106d290bee5` — the date-extraction pipeline as suggested, and `head -1`: the oldest stamp drives the warning, per your weakest-link argument.

.claude/commands/investigate-prd.md
rolfheij-sil: 🤖 Claude: Fixed in `106d290bee5` — explicit no-stamps branch ("no verification stamps found; treat the standards as unverified…") and the grep widened to all six stamped-or-stampable surfaces. One observation back: `.claude/rules/` deliberately received no stamps in the #2438 fix round yet carries code claims — flagging it as a candidate for stamping in the next `/verify-standards` pass rather than this PR.

.claude/commands/verify-standards.md
rolfheij-sil: 🤖 Claude: Fixed in `45217b6d6d0` — trailing HTML-comment marker required; "(or prose saying so)" removed from the contract. Checked: of main's four prose "aspirational" uses, three carry nothing checkable; one gains a single triage-discardable candidate (`Stryker/Stryker.NET` parses as a path on the very line saying Stryker isn't set up).

.claude/commands/verify-standards.md
rolfheij-sil: 🤖 Claude: Fixed in `45217b6d6d0` — exact-set plus `fullmatch` on the Foo/Bar/Baz stems as suggested. Measured cost: at most three new triage-discardable candidates on today's main, against permanently silent false negatives — your trade.

.claude/commands/verify-standards.md
rolfheij-sil: 🤖 Claude: Fixed in `45217b6d6d0` — `RE_INFLIGHT` now requires a PR reference on the line; the CLAUDE.md write-gate false positive is gone from the run.

.claude/commands/verify-standards.md
rolfheij-sil: 🤖 Claude: Fixed in `45217b6d6d0` — extension package.jsons added with your glob (verified `count("/") == 3` matches the layout); the false-positive class you identified can no longer occur (no in-scope doc cites `start:production` today, so it was latent rather than live).

.claude/commands/verify-standards.md
rolfheij-sil: 🤖 Claude: Fixed in `45217b6d6d0` with one deviation: your anchored regex plus `> **Frozen record**` in the blockquote alternation — that's the canonical fence `docs-durability.md` prescribes and the #2438 fix round actually wrote, which the suggested pattern would have missed. Skips now print `rel — matched "<marker>"` as you asked.

.claude/commands/verify-standards.md
rolfheij-sil: 🤖 Claude: Fixed in `45217b6d6d0` — PR refs are collected from the in-hand worktree text during the main loop (folded with item 1's change, no re-reads) and int-sorted. Wall-clock is ~unchanged in a controlled A/B — the identifier greps dominate, so the ~84 eliminated `git show` calls were noise; the commit message's "runtime halved" line predates that measurement and overstates it.

.claude/commands/verify-standards.md
```diff
@@ -0,0 +1,212 @@
...
+    if pkg in main_files:
+        try: scripts |= set(json.loads(sh(["git", "-C", REPO, "show", f"origin/main:{pkg}"])).get("scripts", {}).keys())
+        except Exception: pass
+
+loc_keys = set()
+for lf in [p for p in main_files if p == "assets/localization/en.json" or p.endswith("/contributions/localizedStrings.json")]:
+    try: loc_keys |= set(re.findall(r'"(%[^"%]+%)"', sh(["git", "-C", REPO, "show", f"origin/main:{lf}"])))
+    except Exception: pass
+
+pw_cfg = sh(["git", "-C", REPO, "show", "origin/main:e2e-tests/playwright-cdp.config.ts"])
```
lyonsil: **1/7 - blocking-ish: the playwright check is a permanent silent no-op, and it is hiding a live finding.**

`e2e-tests/playwright-cdp.config.ts` has no `projects:` array at all - it is a single-config file (`testDir`, `testIgnore`, `use`, ...). The real projects are in `e2e-tests/playwright.config.ts`:

```
name: 'smoke'  |  name: 'isolated'  |  name: 'enhanced-resources'
```

So `pw_projects` is the empty set on every run, and the `if pw_projects and ...` guard on line 169 collapses the check into silence - no finding, no warning, no trace. The finding text on line 170 names the wrong file too. This is item 2's failure mode scoped to one check: a lookup that resolves to nothing is indistinguishable from a clean pass. Both the front-matter and the PR body list playwright config names among what this command verifies, so the command currently makes a claim about itself that isn't true.

It is not academic. `.context/standards/Testing-Guide.md:1334` cites:

```
npx playwright test e2e-tests/tests/{feature}/ --config=e2e-tests/playwright.config.ts --project=development --reporter=list
```

There is no `development` project. A reader who runs that command gets nothing - the exact rot class this tool exists to catch, sitting in the corpus, invisible because the check that would have caught it is inert.

One caution on the fix: repointing the path alone is not enough, because the regex is not scoped to playwright invocations. On today's main it also captures `extensions` from `npx tsc --noEmit --project extensions/src/<ext>/tsconfig.json` and `c-sharp-tests` from `dotnet watch test --project c-sharp-tests/` - four new false positives the moment `pw_projects` becomes non-empty. Gate the match on the line actually being a playwright invocation, e.g.:

```python
pw_cfg = sh_ck(["git", "-C", REPO, "show", "origin/main:e2e-tests/playwright.config.ts"])
...
if "playwright test" in ln:
    for m in re.findall(r'--project[= ]([A-Za-z0-9-]+)', ln):
        if pw_projects and m not in pw_projects:
            findings[rel].append(f":{i} [playwright] project `{m}` not in playwright.config.ts")
```

.claude/commands/verify-standards.md
```diff
@@ -0,0 +1,212 @@
...
+pw_projects = set(re.findall(r"name:\s*['\"]([^'\"]+)['\"]", pw_cfg))
+cats = set(re.findall(r'\[Category\("([^"]+)"\)\]', sh(["git", "-C", REPO, "grep", "-h", r"\[Category(", "origin/main", "--", "c-sharp-tests"])))
```
lyonsil, line 111: **2/7 - the same silent-disable shape survives for `cats`.**

`if cats and cat not in cats` on line 166 has the identical property as the playwright guard: if the `git grep` for `[Category(` ever returns nothing - directory renamed, attribute style changed, grep failing for any reason - the dotnet check turns itself off and the report looks the same as a clean one.

It works today (I get 11 categories: Acceptance, Contract, Critical, DiskVerification, EdgeCase, ErrorPath, GoldenMaster, Infrastructure, Integration, Invariant, Regression), so this is latent rather than live. But item 2 was about a class, not an instance, and the class is still here in two places. The cheap close is to make an empty inventory *announce itself* rather than disappear:

```python
for label, inv in (("dotnet categories", cats), ("playwright projects", pw_projects)):
    if not inv: print(f"WARN: {label} inventory is empty - that check is disabled this run")
```

Two lines, and the report can never again imply coverage it didn't have.

.claude/commands/verify-standards.md
```diff
@@ -0,0 +1,212 @@
...
+    m = RE_FROZEN.search("\n".join(text.splitlines()[:30]))
+    return m.group(0).strip()[:60] if m else None
+
+findings, skipped = collections.defaultdict(list), []
+wt_files = sh_ck(["git", "-C", REPO, "ls-files"]).splitlines()
+targets = [f for f in wt_files if any(f == s or f.startswith(s + "/") for s in SCOPE) and f.endswith(".md")]
+if not targets: raise SystemExit(f"FATAL: no in-scope .md files found under {SCOPE}")
+pr_refs = set()
+for rel in sorted(targets):
+    with open(os.path.join(REPO, rel), encoding="utf-8") as fh: text = fh.read()
```
lyonsil: **3/7 - a tracked-but-missing file raises a traceback, not a `FATAL`.**

`git ls-files` lists tracked paths, which is not the same set as paths present on disk - an unstaged deletion, an interrupted rebase, or a sparse checkout all leave entries here with no file behind them. I reproduced it by moving `Git-Guide.md` aside:

```
Traceback (most recent call last):
  File "sweep.py", line 60, in <module>
    with open(os.path.join(REPO, rel), encoding="utf-8") as fh: text = fh.read()
FileNotFoundError: [Errno 2] No such file or directory: '.../.context/standards/Git-Guide.md'
```

The exit code is fine (1), so this is not the dangerous silent variety. But Step 2 tells the runner "if the sweep prints `FATAL`, stop and report" - a bare traceback isn't a `FATAL`, so the instruction doesn't cover the case, and the operator is left improvising. Fold it into the convention you already established:

```python
try:
    with open(os.path.join(REPO, rel), encoding="utf-8") as fh: text = fh.read()
except FileNotFoundError:
    raise SystemExit(f"FATAL: {rel} is tracked but missing from the working tree - commit, restore, or stash before running the gate")
```

.claude/commands/verify-standards.md
```diff
@@ -0,0 +1,212 @@
...
+    fm = frozen_marker(text)
+    if fm: skipped.append(f'{rel} - matched "{fm}"'); continue
+    pr_refs |= set(int(n) for n in RE_PRREF.findall(text))
+    in_fence = False
+    for i, ln in enumerate(text.splitlines(), 1):
+        if ln.strip().startswith("```"):
```
lyonsil, line 148: **4/7 - `in_fence` is load-bearing now, so a stuck fence silently disables two checks for the rest of a file.**

Not a defect today - I checked all 75 in-scope files and every one has an even ``` count with no four-backtick fences, so nothing is currently mis-classified. Flagging it because the risk profile changed with this fix: while the guard was dead the toggle couldn't hurt anything, and now an odd count or a ```` ```` wrapper (which `startswith("```")` also matches) latches `in_fence` true and suppresses every subsequent path and l10n check in that file - with no output distinguishing it from a genuinely clean file.

Given this command's own standard for silent no-ops, a one-line assertion after the per-file loop would keep it honest:

```python
if in_fence: print(f"WARN: {rel} ends inside an unclosed fence - path/l10n checks were suppressed from the last ``` onward")
```

Worth doing while the corpus is clean, since that's when it's free.

.claude/commands/verify-standards.md
```diff
@@ -0,0 +1,212 @@
...
+  provenance fence — a blockquote starting `**Frozen record**` or naming itself a
+  historical/point-in-time record — is skipped entirely; the report lists each skip with the
+  marker text it matched, so an accidental freeze is visible.
+- **Aspirational claims**: a line **ending in** `<!-- aspirational: not yet in repo -->` is exempt
+  from that line's checks — it documents intent, not current code. The marker is the only
+  recognized form; prose saying "aspirational" does not exempt anything (a mechanical gate can't
+  honor prose).
+- **Fenced code blocks**: localization-key and path checks are suppressed inside ``` fences
+  (example keys and illustrative paths are legitimate there), but identifier, npm-script, dotnet
+  category, and playwright project checks stay live — a fenced command or code sample presents
```
lyonsil: **5/7 - nit: the Exemptions rationale oversells what stays live inside fences.**

The text says identifier checks stay live in fences because "a fenced command or code sample presents itself as real". That holds for the command checks (npm, dotnet, playwright), which match raw line text. It does not hold for identifiers: `RE_TICK` requires backticks, and fenced code is not backticked, so identifiers in a fenced code *sample* are not checked in fences or in prose.

My probe file confirms the split - inside a fenced ```ts block:

```
const x = new TotallyFabricatedProviderName();   -> no finding
`YetAnotherFabricatedIdentifier`                 -> :20 [ident] grep-resolves nowhere
```

The behavior is defensible (checking every token in a code block would be a noise firehose); it's the stated reason that's off, and only for the "code sample" half. Since this PR's whole currency is claims that survive checking, suggest narrowing the sentence to what's actually delivered - fenced *commands* are checked; identifiers are checked wherever they're backticked, which in practice means prose.

.claude/commands/investigate-prd.md
```diff
@@ -0,0 +1,314 @@
...
+> checkouts live somewhere else, set `PT_REPOS_ROOT` to that directory. Without them my
+> coverage shrinks — I'll fall back to the bundled inventory and flag the gaps. Continue
+> anyway, or fix and rerun?
+
+**Preflight — standards freshness.** The stamped docs this investigation leans on carry
+`Verified against paranext-core origin/main` stamps. Find the **oldest** stamp date (the weakest
+link is what governs risk):
+
+```bash
+grep -rh "Verified against paranext-core" .context/standards/ .claude/skills/ .claude/rules/ .claude/agents/ .claude/commands/ CLAUDE.md 2>/dev/null | grep -oE '20[0-9]{2}-[0-9]{2}-[0-9]{2}' | sort | head -1
```
lyonsil, line 86: **6/7 - the widened grep can't see the surfaces that have no stamps at all.**

Verified working: on this branch the probe returns `2026-08-03`, and it is genuinely the oldest (all 17 stamps carry that date). The `head -1` change and the no-stamps branch both do what they say.

The residual is that widening the grep only helps where stamps exist. Today's distribution across the six surfaces you now grep:

```
10  .context/standards/       7  .claude/skills/       1  .claude/commands/
 0  .claude/rules/            0  .claude/agents/       0  CLAUDE.md
```

So `.claude/rules/` and `.claude/agents/` contribute nothing, and the probe reports a confident, fresh `2026-08-03` while two of the six surfaces have never been verified at all - `.claude/rules/` in particular carries code claims, as you noted. Your plan to stamp them in a later pass is the right call and I'm not asking for it here. But until then the probe is silently blind to a third of its own scope, which is the same shape as the findings above.

Cheap interim close - report the gaps rather than averaging them away:

```bash
for d in .context/standards .claude/skills .claude/rules .claude/agents .claude/commands CLAUDE.md; do
  grep -rq "Verified against paranext-core" "$d" 2>/dev/null || echo "UNSTAMPED: $d"
done
```

Then the one-line message to the user can read "standards last verified {date}; .claude/rules and .claude/agents carry no stamps", which is the true statement.

lyonsil (commented):
## Fix round verified — all 11 items land

I re-ran the sweep from the head of this branch (extracted the heredoc, ran it against a clean worktree) and unit-tested each changed regex. **All 11 items are genuinely fixed**, including the deviation on item 10: I tested my suggested `RE_FROZEN` against the repo's canonical `> **Frozen record** - ...` fence and it does *not* match, while yours does. Your version is the correct one - thanks for catching that rather than applying the patch as given.

Spot-checks worth recording:

- **1** - 75 files scanned from the branch (vs main's 60), and `verify-standards.md` now gates itself and comes out clean.
- **2** - bad ref reproduces `FATAL: ... Not a valid object name`, exit 1. The addendum's fetch WARN prints as designed.
- **3** - probe file confirms the policy: fenced path/`%key%` suppressed, prose path/`%key%` flagged, fenced `npm run` and `Category=` still flagged.
- **7** - all seven of the false negatives I listed (`OldTestamentBooks`, `SomeoneService`, `MyProjectDataProvider`, `Variants`, `ExampleDataProvider`, `Others`, `RootLayout`) are now caught; `Old`/`Root`/`FooProvider` still exempt.
- **9** - every extension manifest on main sits at `count("/") == 3`; none missed.
- **11** - output is `[210, 2219, 2590, 2626]`, int-sorted.

Seven follow-ups below. Only the first is worth blocking on; the rest are small.

---

**7/7 - the commit message for `45217b6d6d0` still claims a speedup you have since retracted.**

Its body ends `Measured: 150 -> 91 candidates on identical input, runtime roughly halved.` Your reply on item 11 retracts exactly that: *"Wall-clock is ~unchanged in a controlled A/B ... the commit message's 'runtime halved' line predates that measurement and overstates it."*

We squash-merge, and GitHub's default squash body concatenates the commit messages - so as things stand, a known-false measured claim lands in `main`'s history, inside the PR whose stated purpose is stopping never-true claims from landing. Please curate the squash message (`/squash-message`) so the retracted half doesn't survive the merge. The `150 -> 91` half is fine; it's the runtime clause that needs to go.

rolfheij-sil:
🤖 Claude: All seven addressed — six on the branch, the squash message curated per 7/7
(added to the PR body). Two structural notes: the branch is rebased onto main (#2438's
squash-merge had retargeted the base and bloated this diff to 100+ duplicate files; it's
back to the real ones — apologies for any comments GitHub now marks outdated), and the
repaired playwright check plus triage surfaced real rot beyond your example: Testing-Guide's
whole C# test-category scheme (Smoke/Full speed tiers with time budgets) arrived with the
generic profile embed (865a2a19446) and never described this repo — replaced with the 11
categories that exist, each verified, and two broken commands in the same block fixed.
Final sweep on the branch: 109 candidates, zero WARNs. Two independent review-verify
rounds ran over these fixes; the second caught my own item-5 rewrite overcorrecting —
details in that thread.

.claude/commands/verify-standards.md
rolfheij-sil: 🤖 Claude: Fixed in a0c7c12947c per your snippet, one deviation: the config read uses
`sh_ck`, so if e2e-tests/playwright.config.ts ever disappears from origin/main the gate
FATALs loudly rather than becoming another silent empty inventory — loud-beats-silent
seemed right here; say the word if you'd prefer a WARN. Gating verified per-line against
the corpus: your predicted false positives were exactly four (your two plus
test-runner/SKILL.md:279,282 and Testing-Guide.md:1439's `dotnet watch test --project`),
all suppressed, with Testing-Guide.md:1334 kept as the single real finding. That line is
fixed in 7ddbe94e1ce: `--project=isolated`, path moved inside its testDir — `development`
never existed (`git log -S` is empty), `smoke` has no per-feature subdirectories so a
`{feature}` filter under it matches nothing, and `isolated` is the project both the config
comment and tests/isolated/README.md designate for feature e2e tests. The "(CI)" label on
that example also went — CI actually runs `npm run test:e2e:smoke`.

.claude/commands/verify-standards.md
rolfheij-sil: 🤖 Claude: Added in a0c7c12947c, with labels naming the sources ("dotnet test categories
(c-sharp-tests)", "playwright projects (playwright.config.ts)"). Mutation-tested: a bogus
attribute name + the CDP config repoint → both WARNs print where previously the two real
findings just vanished. I left `scripts`/`loc_keys` out of the loop deliberately — they
have no `if inv and` guard, so an empty inventory there fails loud (mass findings), not
silent. Follow-through in 57394586ccb: Step 2 and the completion checklist now carry a
WARN protocol (a run with warnings is not a clean pass) — otherwise the silent no-op
would have reappeared one layer up, in the runner's own reporting.

.claude/commands/verify-standards.md
rolfheij-sil: 🤖 Claude: Applied in a0c7c12947c and tested both directions with a tracked file moved
aside (.claude/agents/feature-mapper.md): pre-fix, bare FileNotFoundError traceback with
exit 1 and no FATAL token; post-fix, the FATAL line with commit/restore/stash guidance,
exit 1, and no traceback (SystemExit is handled before traceback printing).

.claude/commands/verify-standards.md
rolfheij-sil: 🤖 Claude: Added in a0c7c12947c and probed with both latch shapes — a stray trailing
fence and a four-backtick wrapper (odd toggle count): the WARN names the file in both
cases, and the unmodified 75-file corpus prints zero WARNs, so no false alarms. An
independent re-review additionally fixture-tested the leak-across-files scenario (a file
ending inside an open fence, followed by a frozen-record file that `continue`s before the
reset, followed by a normal file): the per-file reset holds — no leak. The WARN is also
now part of the runner protocol (57394586ccb), same as item 2.

.claude/commands/verify-standards.md
rolfheij-sil: 🤖 Claude: You were right — and my first rewrite (a0c7c12947c) overcorrected into the
opposite falsehood: it claimed an identifier appearing only inside a fenced code sample
is never checked, but the identifier check has no fence guard, and backticks nest inside
fences — six of the current 109 findings are exactly that (e.g.
.claude/agents/review-analyzer.md:232-233, a fenced report template). An independent
review of the fix round caught it. Final text (57394586ccb) states precisely what fires
where: path/l10n suppressed inside fences; the command checks and the undated-status
check live on raw line text everywhere; the identifier check backtick-gated — live in
prose and on backticks nested inside fences (templates, code comments), never seeing
unbackticked fenced code.

.claude/commands/investigate-prd.md
rolfheij-sil: 🤖 Claude: Applied in 4d72c3e1d95, one deviation from the snippet: the unstamped-surface
grep is anchored on the full dated stamp form rather than the plain phrase — because the
plain phrase self-matches investigate-prd.md's own prose describing the stamp. (That's
also what the "1 × .claude/commands" in your distribution was: the probe file quoting
itself, not a stamp.) Date-anchored, the honest count is four unstamped surfaces:
.claude/rules, .claude/agents, .claude/commands, and CLAUDE.md. The probe prints the
oldest stamp (2026-08-03) plus those UNSTAMPED lines, the user-facing message template
requires both facts, and the /verify-standards nudge fires when the date is stale or
anything is unstamped. Verified the anchor can't self-match (the committed regex literal
contains no digits after "20"), and `.claude/rules`/`.claude/agents` stamping stays a
next-pass item as agreed.

rolfheij-sil:
🤖 Claude: Agreed — curated squash message added to the PR body (Squash message section),
drafted per /squash-message from the final branch state. The runtime clause is gone
entirely; I also left out the accurate "150→91 candidates" half, since it's a fix-round
measurement rather than merged-state behavior — say the word if you'd like that half back
as a bullet. Related branch hygiene in the same round: rebased onto main (the #2438
squash-merge had retargeted this PR's base, ballooning the diff to 100+ duplicate files),
so the commit you cited now lives at a different SHA; the curated message is what should
land on main regardless.

.claude/commands/verify-standards.md
lyonsil: FATAL seems appropriate

lyonsil:
Verified the fix round at `57394586ccb` by extracting the sweep and mutation-testing each of the six items rather than reading the diff. All six land: the playwright line-gate suppresses exactly the four false positives you name and still fires on a mutated project name (`:1330 [playwright] project 'development' not in playwright.config.ts`); both empty-inventory WARNs print with their source labels; the missing-file case gives `FATAL` + exit 1 with no traceback; the stuck-fence WARN names the file and does not leak into the next one; and classifying all 109 findings by fence position gives 6 in-fence, all `ident`, 0 in-fence path/l10n - exactly what the rewritten Exemptions text now says. The `Testing-Guide.md` category rewrite checks out independently too (the 11 names match `git grep '\[Category('` on main exactly; no root `.sln`, so `dotnet test c-sharp-tests/` is the right form; `test:core = vitest`, so `-- --run -t` forwards where the old `npm test -- --testNamePattern` reached only the workspace suites).

Two corrections to the **squash message** - same class as item 7/7, text bound for `main`'s history.

**1. "dated to the ref's last-fetched time" is not what the WARN prints.**

`git log -1 --format=%cs origin/main` prints the tip **commit's** committer date, not when the ref was fetched. They coincide in my clone today (tip `2026-08-05`; last fetch `2026-08-05 10:48`), but they diverge whenever `main` has not moved since your fetch - the WARN then reports a date older than the ref's actual freshness, which is the wrong direction for a staleness warning.

```diff
   gaps rather than clean passes. A failed fetch against a stale
-  origin/main also WARNs, dated to the ref's last-fetched time.
+  origin/main also WARNs, dated to origin/main's tip commit.
```

Step 1's own WARN string carries the same looseness ("existence checks will run against the last-fetched `origin/main` from {date}") and deserves the same reword. If last-fetched really is the number you want to report, `git reflog show --date=short origin/main | head -1` has it - with the `%cs` form as the fallback for clones whose remote-tracking reflog is empty.

**2. `Critical` and `Integration` are not invented categories.**

Both exist on main - 16 and 38 uses respectively. What was invented is the speed-tier framing, the time budgets, and the `Smoke`/`Full` names. As written, a reader can come away believing `Critical` and `Integration` were fabrications too.

```diff
-- an invented Smoke/Critical/Full/Integration C# test-category
-  scheme with time budgets nothing ever measured — replaced with the
-  11 categories that actually exist in c-sharp-tests, pointing at
+- an invented speed-tier scheme — time budgets nothing ever
+  measured, over two category names that never existed (Smoke,
+  Full) and two that do (Critical, Integration) — replaced with
+  the 11 categories that actually exist in c-sharp-tests, pointing at
```

**FYI, one footnote on your item-1 reply.** `name: 'development'` did exist in `e2e-tests/playwright.config.ts`, on Martijn's original E2E-infra branch (`667407e25df`, alongside a `production` project). That branch is not an ancestor of `main` - it landed squashed as `d14dedefbf5` with `smoke` as the only project, which is why a history search scoped to `main` comes back empty. So "never existed" is true of `main`'s history, which is the history that matters here; it just is not true of the full ref graph. No change needed - flagging it only so the claim does not surprise anyone who searches `--all` later.

lyonsil (approved):
:lgtm: - Thanks for putting this together. It's a good idea. Note the proposed squash message has a couple of small things to adjust from Claude's review. The actual changes look good as-is.

> @lyonsil reviewed 120 files and all commit messages, made 1 comment, and resolved 17 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2627#-:-OzMc-Lg7vNlqWsZRbfq:b-l88ap1)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

Issue #518, opened 2023-10-03
Design a settings dialog


mikelothers:
Create mockups for settings in Platform.Bible

PR #2629, opened 2026-08-03
PT-4021: Sync \li/\lim gutter indent CSS into _usj-nodes.scss + coverage test

katherinejensen00:
image

## Summary

`extensions/src/platform-scripture-editor/src/_usj-nodes.scss` is a hand-maintained copy of scripture-editors' `packages/platform/src/usj-nodes.css`. It had drifted and was missing the `--para-indent` / `--verse-text-start` entries for the entire `\li`/`\lim` family, so those hanging-indent markers rendered with the gutter marker overlapping the paragraph text in Platform.Bible.

This PR:
1. **Adds the missing `\li`/`\lim` gutter entries**, mirroring scripture-editors commit `846ad99` (PT-4021).
2. **Adds a coverage test** so the copy can't silently drift again — the upstream file already has an equivalent test, but the downstream copy that actually ships in Platform.Bible had none, which is exactly how it drifted.

## Reviewer orientation (two commits)

- **`_usj-nodes.scss`** — the CSS: `\li`/`\li1` → `10vw`, `\li2`/`\lim`/`\lim1` → `15vw`, `\li3`/`\lim2` → `20vw`, `\li4`/`\lim3` → `25vw`, `\lim4` → `30vw` (new block), and all `\li*`/`\lim*` → `--verse-text-start: -7.5vw`. Values come from the USFM LeftMargin spec (inches × 20), matching the upstream fix. The entries were slotted into the existing `.psc-gutter-markers.text-spacing` value-groups (this copy has a small pre-existing divergence — a `[dir='ltr']` variant on the `pi2`/`pi3` selectors — so it isn't a byte-copy of upstream).
- **`usj-nodes-scss-coverage.test.ts`** — the guard: parses the SCSS and asserts every USFM indented/hanging marker has the corresponding custom-property **setter** (it strips comments, ignores `var()` reads, and fails loudly if it parses nothing). Runs under the extensions vitest config.

## Relationship to the editor-side fix

The Lexical cursor fix and the editor-side CSS for PT-4021 live in **scripture-editors** (separate PR). This is the **paranext-core consuming side**: the `platform-scripture-editor` extension uses its own copy of the stylesheet, so the fix has to be synced here too. Longer term, generating or otherwise auto-syncing this file from the editor package would remove the manual-copy drift risk entirely — out of scope here.

## Testing

```
npx vitest run --config extensions/vitest.config.ts usj-nodes-scss-coverage
```
→ both assertions pass against the updated stylesheet.

Linked to https://github.com/eten-tech-foundation/scripture-editors/pull/518

AI-assisted (Claude Code).

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2629)

### DIFF
--- a/extensions/src/platform-scripture-editor/src/_usj-nodes.scss
+++ b/extensions/src/platform-scripture-editor/src/_usj-nodes.scss
@@ -2569,14 +2569,16 @@ span.read img {
 }
 
 /* Indent compensation — mirrors the margin-left values above.
-   Only active when .text-spacing is present (same condition as the source rules). */
+   Only active when .text-spacing is present (same condition as the source rules).
+   pi2 and pi3 use the same primary-indent value in both LTR and RTL (10vw and 15vw
+   respectively), so no direction restriction is needed here. */
 .psc-gutter-markers.text-spacing .usfm_io,
 .psc-gutter-markers.text-spacing .usfm_io1,
 .psc-gutter-markers.text-spacing .usfm_ili,
 .psc-gutter-markers.text-spacing .usfm_ili1,
 .psc-gutter-markers.text-spacing .usfm_li,
 .psc-gutter-markers.text-spacing .usfm_li1,
-.psc-gutter-markers.text-spacing[dir='ltr'] .usfm_pi2 {
+.psc-gutter-markers.text-spacing .usfm_pi2 {
   --para-indent: 10vw;
 }
 .psc-gutter-markers.text-spacing .usfm_q,
@@ -2589,7 +2591,7 @@ span.read img {
 .psc-gutter-markers.text-spacing .usfm_li2,
 .psc-gutter-markers.text-spacing .usfm_lim,
 .psc-gutter-markers.text-spacing .usfm_lim1,
-.psc-gutter-markers.text-spacing[dir='ltr'] .usfm_pi3 {
+.psc-gutter-markers.text-spacing .usfm_pi3 {
   --para-indent: 15vw;
 }
 .psc-gutter-markers.text-spacing .usfm_qm,
--- a/extensions/src/platform-scripture-editor/src/_usj-nodes.scss
+++ b/extensions/src/platform-scripture-editor/src/_usj-nodes.scss
@@ -1,4 +1,7 @@
-/* Copied from https://github.com/eten-tech-foundation/scripture-editors/blob/ba0e846b3f11bea5720c4aa1a3486b69feb6b7e0/packages/platform/src/usj-nodes.css */
+/* Based on https://github.com/eten-tech-foundation/scripture-editors/packages/platform/src/usj-nodes.css
+   (originally synced from commit ba0e846b3f1). Local additions: li/lim --para-indent and
+   --verse-text-start entries in .psc-gutter-markers.text-spacing, mirroring scripture-editors
+   commit 846ad99 (PT-4021 \li/\lim gutter overlap fix). */
 
 /* stylelint-disable */
 
@@ -2567,6 +2570,8 @@ span.read img {
 .psc-gutter-markers.text-spacing .usfm_io1,
 .psc-gutter-markers.text-spacing .usfm_ili,
 .psc-gutter-markers.text-spacing .usfm_ili1,
+.psc-gutter-markers.text-spacing .usfm_li,
+.psc-gutter-markers.text-spacing .usfm_li1,
 .psc-gutter-markers.text-spacing[dir='ltr'] .usfm_pi2 {
   --para-indent: 10vw;
 }
@@ -2577,16 +2582,26 @@ span.read img {
 .psc-gutter-markers.text-spacing .usfm_q4,
 .psc-gutter-markers.text-spacing .usfm_io2,
 .psc-gutter-markers.text-spacing .usfm_ili2,
+.psc-gutter-markers.text-spacing .usfm_li2,
+.psc-gutter-markers.text-spacing .usfm_lim,
+.psc-gutter-markers.text-spacing .usfm_lim1,
 .psc-gutter-markers.text-spacing[dir='ltr'] .usfm_pi3 {
   --para-indent: 15vw;
 }
 .psc-gutter-markers.text-spacing .usfm_qm,
 .psc-gutter-markers.text-spacing .usfm_qm1,
-.psc-gutter-markers.text-spacing .usfm_io3 {
+.psc-gutter-markers.text-spacing .usfm_io3,
+.psc-gutter-markers.text-spacing .usfm_li3,
+.psc-gutter-markers.text-spacing .usfm_lim2 {
   --para-indent: 20vw;
 }
-.psc-gutter-markers.text-spacing .usfm_io4 {
+.psc-gutter-markers.text-spacing .usfm_io4,
+.psc-gutter-markers.text-spacing .usfm_li4,
+.psc-gutter-markers.text-spacing .usfm_lim3 {
   --para-indent: 25vw;
+}
+.psc-gutter-markers.text-spacing .usfm_lim4 {
+  --para-indent: 30vw;
 }
 .psc-gutter-markers.text-spacing .usfm_ipi,
 .psc-gutter-markers.text-spacing .usfm_imi,
@@ -2629,7 +2644,17 @@ span.read img {
 }
 .psc-gutter-markers.text-spacing .usfm_ili,
 .psc-gutter-markers.text-spacing .usfm_ili1,
-.psc-gutter-markers.text-spacing .usfm_ili2 {
+.psc-gutter-markers.text-spacing .usfm_ili2,
+.psc-gutter-markers.text-spacing .usfm_li,
+.psc-gutter-markers.text-spacing .usfm_li1,
+.psc-gutter-markers.text-spacing .usfm_li2,
+.psc-gutter-markers.text-spacing .usfm_li3,
+.psc-gutter-markers.text-spacing .usfm_li4,
+.psc-gutter-markers.text-spacing .usfm_lim,
+.psc-gutter-markers.text-spacing .usfm_lim1,
+.psc-gutter-markers.text-spacing .usfm_lim2,
+.psc-gutter-markers.text-spacing .usfm_lim3,
+.psc-gutter-markers.text-spacing .usfm_lim4 {
   --verse-text-start: -7.5vw;
 }
 .psc-gutter-markers.text-spacing .usfm_iq,
--- a/extensions/src/platform-scripture-editor/src/_usj-nodes.scss
+++ b/extensions/src/platform-scripture-editor/src/_usj-nodes.scss
@@ -1,7 +1,11 @@
-/* Based on https://github.com/eten-tech-foundation/scripture-editors/packages/platform/src/usj-nodes.css
-   (originally synced from commit ba0e846b3f1). Local additions: li/lim --para-indent and
-   --verse-text-start entries in .psc-gutter-markers.text-spacing, mirroring scripture-editors
-   commit 846ad99 (PT-4021 \li/\lim gutter overlap fix). */
+/* Based on scripture-editors' packages/platform/src/usj-nodes.css:
+   https://github.com/eten-tech-foundation/scripture-editors/blob/main/packages/platform/src/usj-nodes.css
+   (originally synced from commit ba0e846b3f1:
+   https://github.com/eten-tech-foundation/scripture-editors/commit/ba0e846b3f1).
+   Local additions: li/lim --para-indent and --verse-text-start entries in
+   .psc-gutter-markers.text-spacing, mirroring scripture-editors commit 846ad99
+   (https://github.com/eten-tech-foundation/scripture-editors/commit/846ad99,
+   \li/\lim gutter overlap fix). */
 
 /* stylelint-disable */
 
--- a/extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts
+++ b/extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts
@@ -1,0 +1,139 @@
+// @vitest-environment node
+import { readFileSync } from 'node:fs';
+import { dirname, resolve } from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { describe, expect, it } from 'vitest';
+
+// `_usj-nodes.scss` is a hand-maintained copy of scripture-editors'
+// packages/platform/src/usj-nodes.css. That upstream file has its own coverage test
+// (usj-nodes.test.ts); this is the equivalent guard for the downstream copy that actually
+// ships in Platform.Bible, so the two can't silently drift (which reintroduced the \li/\lim
+// gutter overlap once already). Keep the marker lists below in sync with the upstream test.
+
+const dir = dirname(fileURLToPath(import.meta.url));
+// Strip CSS comments so their text can't be mistaken for selectors or declarations.
+const css = readFileSync(resolve(dir, '_usj-nodes.scss'), 'utf-8').replace(/\/\*[\s\S]*?\*\//g, '');
+
+/**
+ * Collects all `.usfm_<marker>` names that appear in a `.psc-gutter-markers` rule block that SETS
+ * `property` (a `${property}:` declaration, not a `var(--property)` read).
+ */
+function getGutterMarkers(property: string): Set<string> {
+  const markers = new Set<string>();
+  const ruleBlock = /([^{}]+)\{([^}]+)\}/g;
+  let match;
+  // eslint-disable-next-line no-cond-assign
+  while ((match = ruleBlock.exec(css)) !== null) {
+    const [, selectors, declarations] = match;
+    if (!selectors.includes('psc-gutter-markers') || !declarations.includes(`${property}:`))
+      continue;
+    const markerName = /\.usfm_([a-z0-9]+)/g;
+    let m;
+    // eslint-disable-next-line no-cond-assign
+    while ((m = markerName.exec(selectors)) !== null) markers.add(m[1]);
+  }
+  return markers;
+}
+
+// USFM standard LeftMargin values -> vw (formula: inches x 20).
+// Source: https://github.com/ubsicap/usfm/blob/master/sty/usfm.sty
+// Every marker here must have a --para-indent entry in .psc-gutter-markers.text-spacing.
+const PARA_INDENT_MARKERS = new Set([
+  // 0.25" -> 5vw
+  'ipi',
+  'imi',
+  'pmo',
+  'pm',
+  'pmc',
+  'pmr',
+  'pi',
+  'pi1',
+  'mi',
+  // 0.5" -> 10vw
+  'io',
+  'io1',
+  'ili',
+  'ili1',
+  'li',
+  'li1',
+  'pi2',
+  // 0.75" -> 15vw
+  'q',
+  'q1',
+  'q2',
+  'q3',
+  'q4',
+  'io2',
+  'ili2',
+  'li2',
+  'lim',
+  'lim1',
+  'pi3',
+  // 1.0" -> 20vw
+  'qm',
+  'qm1',
+  'io3',
+  'li3',
+  'lim2',
+  // 1.25" -> 25vw
+  'io4',
+  'li4',
+  'lim3',
+  // 1.5" -> 30vw
+  'lim4',
+]);
+
+// Markers with a negative FirstLineIndent (hanging indent) -> --verse-text-start.
+// Source: https://github.com/ubsicap/usfm/blob/master/sty/usfm.sty
+const VERSE_TEXT_START_MARKERS = new Set([
+  'q',
+  'q1',
+  'q2',
+  'q3',
+  'q4',
+  'qm',
+  'qm1',
+  'qm2',
+  'qm3',
+  'iq',
+  'iq1',
+  'iq2',
+  'iq3',
+  'ili',
+  'ili1',
+  'ili2',
+  'li',
+  'li1',
+  'li2',
+  'li3',
+  'li4',
+  'lim',
+  'lim1',
+  'lim2',
+  'lim3',
+  'lim4',
+]);
+
+describe('_usj-nodes.scss .psc-gutter-markers.text-spacing coverage', () => {
+  it('every USFM indented marker has a --para-indent entry', () => {
+    const covered = getGutterMarkers('--para-indent');
+    // Guard against a silently-broken parser (no matches -> every marker would look "missing").
+    expect(covered.size, 'no --para-indent gutter markers parsed').toBeGreaterThan(0);
+    const missing = [...PARA_INDENT_MARKERS].filter((m) => !covered.has(m));
+    expect(
+      missing,
+      `Add --para-indent entries to .psc-gutter-markers.text-spacing for: ${missing.join(', ')}`,
+    ).toEqual([]);
+  });
+
+  it('every hanging-indent marker has a --verse-text-start entry', () => {
+    const covered = getGutterMarkers('--verse-text-start');
+    // Guard against a silently-broken parser (no matches -> every marker would look "missing").
+    expect(covered.size, 'no --verse-text-start gutter markers parsed').toBeGreaterThan(0);
+    const missing = [...VERSE_TEXT_START_MARKERS].filter((m) => !covered.has(m));
+    expect(
+      missing,
+      `Add --verse-text-start entries to .psc-gutter-markers.text-spacing for: ${missing.join(', ')}`,
+    ).toEqual([]);
+  });
+});
--- a/extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts
+++ b/extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts
@@ -5,14 +5,17 @@ import { fileURLToPath } from 'node:url';
 import { describe, expect, it } from 'vitest';
 
 // `_usj-nodes.scss` is a hand-maintained copy of scripture-editors'
-// packages/platform/src/usj-nodes.css. That upstream file has its own coverage test
-// (usj-nodes.test.ts); this is the equivalent guard for the downstream copy that actually
-// ships in Platform.Bible, so the two can't silently drift (which reintroduced the \li/\lim
-// gutter overlap once already). Keep the marker lists below in sync with the upstream test.
+// packages/platform/src/usj-nodes.css. The downstream copy had no guard against drifting from
+// upstream, which is how the \li/\lim gutter entries went missing and the overlap regressed. This
+// test is that guard: keep the marker lists below in sync with upstream usj-nodes.css.
 
 const dir = dirname(fileURLToPath(import.meta.url));
-// Strip CSS comments so their text can't be mistaken for selectors or declarations.
-const css = readFileSync(resolve(dir, '_usj-nodes.scss'), 'utf-8').replace(/\/\*[\s\S]*?\*\//g, '');
+// Strip comments so their text can't be mistaken for selectors or declarations. Block comments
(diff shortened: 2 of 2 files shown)


imnasnainaec (commented):
> @imnasnainaec made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2629)* status: 0 of 2 files reviewed, all discussions resolved.

___
*[`extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts` line 21 at r1](https://reviewable.io/reviews/paranext/paranext-core/2629#-OzDC3PyDfM3iaxyeW3l:-OzDC3PyDfM3iaxyeW3m:b-ytwt1s) ([raw file](https://github.com/paranext/paranext-core/blob/8f881c6fcce19a62098195ee62d6f7adab46577d/extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts#L21)):*
Devin observations:

> if gutter rules are ever nested (e.g. inside a media query or SCSS nesting) they would silently be dropped, and only the covered.size > 0 guard would remain as protection
> 
> coverage test only checks presence, not values
> 
> a marker moved outside the `.text-spacing` group would go unnoticed
___
*[`extensions/src/platform-scripture-editor/src/_usj-nodes.scss` line 1 at r1](https://reviewable.io/reviews/paranext/paranext-core/2629#-OzDBNIC95VBzZSsVxoZ:-OzDBNIC95VBzZSsVxo_:b-ibzu51) ([raw file](https://github.com/paranext/paranext-core/blob/8f881c6fcce19a62098195ee62d6f7adab46577d/extensions/src/platform-scripture-editor/src/_usj-nodes.scss#L1)):*
> ```SCSS
> /* Based on https://github.com/eten-tech-foundation/scripture-editors/packages/platform/src/usj-nodes.css
> ```

While moving away from permalink to a general repo path seems intentional, this url is 404.
___
*[`extensions/src/platform-scripture-editor/src/_usj-nodes.scss` line 4 at r1](https://reviewable.io/reviews/paranext/paranext-core/2629#-OzDBQWoCxWMqKzSVj2v:-OzDBQWoCxWMqKzSVj2w:b-iaohgv) ([raw file](https://github.com/paranext/paranext-core/blob/8f881c6fcce19a62098195ee62d6f7adab46577d/extensions/src/platform-scripture-editor/src/_usj-nodes.scss#L4)):*
> ```SCSS
>    (originally synced from commit ba0e846b3f1). Local additions: li/lim --para-indent and
>    --verse-text-start entries in .psc-gutter-markers.text-spacing, mirroring scripture-editors
>    commit 846ad99 (PT-4021 \li/\lim gutter overlap fix). */
> ```

It might be nice to have a url to accompany this commit hash.

katherinejensen00 (commented):
Thanks for the review!

> @katherinejensen00 made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2629#-:-OzIKnfH2bj355eyV2T8:b-9qdoz3)* status: 0 of 2 files reviewed, all discussions resolved.

___
*[`extensions/src/platform-scripture-editor/src/_usj-nodes.scss` line 1 at r1](https://reviewable.io/reviews/paranext/paranext-core/2629#-OzDBNIC95VBzZSsVxoZ:-OzIKkAS0FHK4Id6Nqac:b8ymmxe) ([raw file](https://github.com/paranext/paranext-core/blob/8f881c6fcce19a62098195ee62d6f7adab46577d/extensions/src/platform-scripture-editor/src/_usj-nodes.scss#L1)):*
<details><summary><i>Previously, imnasnainaec (D. Ror.) wrote…</i></summary><blockquote>

While moving away from permalink to a general repo path seems intentional, this url is 404.
</blockquote></details>

Good catch — fixed to a working general path: `https://github.com/eten-tech-foundation/scripture-editors/blob/main/packages/platform/src/usj-nodes.css`
___
*[`extensions/src/platform-scripture-editor/src/_usj-nodes.scss` line 4 at r1](https://reviewable.io/reviews/paranext/paranext-core/2629#-OzDBQWoCxWMqKzSVj2v:-OzIKmwoAh3_xjXtnpcX:b-ik4uwf) ([raw file](https://github.com/paranext/paranext-core/blob/8f881c6fcce19a62098195ee62d6f7adab46577d/extensions/src/platform-scripture-editor/src/_usj-nodes.scss#L4)):*
<details><summary><i>Previously, imnasnainaec (D. Ror.) wrote…</i></summary><blockquote>

It might be nice to have a url to accompany this commit hash.
</blockquote></details>

Added — both commit hashes now link to their commits (`ba0e846b3f1` and `846ad99`).
___
*[`extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts` line 21 at r1](https://reviewable.io/reviews/paranext/paranext-core/2629#-OzDC3PyDfM3iaxyeW3l:-OzIKgzREsPrO8CTJrlt:b-d16o3n) ([raw file](https://github.com/paranext/paranext-core/blob/8f881c6fcce19a62098195ee62d6f7adab46577d/extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts#L21)):*
<details><summary><i>Previously, imnasnainaec (D. Ror.) wrote…</i></summary><blockquote>

Devin observations:

> if gutter rules are ever nested (e.g. inside a media query or SCSS nesting) they would silently be dropped, and only the covered.size > 0 guard would remain as protection
> 
> coverage test only checks presence, not values
> 
> a marker moved outside the `.text-spacing` group would go unnoticed
</blockquote></details>

Thanks — all three are addressed in the rewritten test:

-   **Nesting:** added `nestingProblems()`, which walks brace depth and fails loudly if any `--para-indent`/`--verse-text-start` setter sits deeper than one block (i.e. wrapped in an `@media`/`@keyframes`/SCSS nest). Verified it fires end-to-end by nesting a rule.
-   **Values:** the test now asserts the exact `vw` value for each marker, not just presence.
-   **`.text-spacing` scope:** extraction now requires _both_ `.psc-gutter-markers` and `.text-spacing` on the selector, so a marker moved out of that group reads as uncovered and fails.

imnasnainaec (commented):
> @imnasnainaec reviewed 1 file and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2629)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on katherinejensen00).

___
*[`extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts` line 15 at r2](https://reviewable.io/reviews/paranext/paranext-core/2629#-OzIUAWq5hRHCGKOf-SP:-OzIUAWq5hRHCGKOf-SQ:b-bcwn4z) ([raw file](https://github.com/paranext/paranext-core/blob/b48912ef45d550c1cb40a7b4112efc73f56b3ab2/extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts#L15)):*
:pick: Devin nitpick:

> The parser strips only `/* ... */` comments. In a .scss file, a commented-out declaration such as `// --para-indent: 10vw;` inside a `.psc-gutter-markers.text-spacing` block would still be seen as a live setter by `getGutterMarkerValues`, so the guard could pass on a rule that no longer ships. Low probability today (the file is a byte-ish copy of a plain CSS file with no `//` comments), but worth a one-line addition if the file is ever hand-edited in SCSS style.
___
*[`extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts` line 103 at r2](https://reviewable.io/reviews/paranext/paranext-core/2629#-OzIUYiv715O5b3gFhQf:-OzIUYiv715O5b3gFhQg:bce5ous) ([raw file](https://github.com/paranext/paranext-core/blob/b48912ef45d550c1cb40a7b4112efc73f56b3ab2/extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts#L103)):*
:pick: Slight scope-creep by Devin:

> The expected `--para-indent` map covers `qm`/`qm1` but not `qm2`/`qm3`, which do get `--verse-text-start` overrides and have their own margins in the stylesheet. If those markers are genuinely missing a `--para-indent` setter, the same gutter-overlap symptom this PR fixes for `\li`/`\lim` would still exist for `\qm2`/`\qm3`, and the new guard won't catch it because they aren't listed. Worth confirming against the upstream test's marker list.

katherinejensen00 (commented):
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2629)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on imnasnainaec).

___
*[`extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts` line 15 at r2](https://reviewable.io/reviews/paranext/paranext-core/2629#-OzIUAWq5hRHCGKOf-SP:-OzJ-E1x9znRg2x3BEYc:b-nomi3j) ([raw file](https://github.com/paranext/paranext-core/blob/b48912ef45d550c1cb40a7b4112efc73f56b3ab2/extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts#L15)):*
<details><summary><i>Previously, imnasnainaec (D. Ror.) wrote…</i></summary><blockquote>

:pick: Devin nitpick:

> The parser strips only `/* ... */` comments. In a .scss file, a commented-out declaration such as `// --para-indent: 10vw;` inside a `.psc-gutter-markers.text-spacing` block would still be seen as a live setter by `getGutterMarkerValues`, so the guard could pass on a rule that no longer ships. Low probability today (the file is a byte-ish copy of a plain CSS file with no `//` comments), but worth a one-line addition if the file is ever hand-edited in SCSS style.
</blockquote></details>

Good call — the parser now strips `//` line comments too (after the block-comment pass, so the header URLs are already gone). Verified a commented-out `// --para-indent: …;` inside a gutter block is no longer read as a live setter.
___
*[`extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts` line 103 at r2](https://reviewable.io/reviews/paranext/paranext-core/2629#-OzIUYiv715O5b3gFhQf:-OzJ-8MlA2vj9QwEThAg:b-kdqpi9) ([raw file](https://github.com/paranext/paranext-core/blob/b48912ef45d550c1cb40a7b4112efc73f56b3ab2/extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts#L103)):*
<details><summary><i>Previously, imnasnainaec (D. Ror.) wrote…</i></summary><blockquote>

:pick: Slight scope-creep by Devin:

> The expected `--para-indent` map covers `qm`/`qm1` but not `qm2`/`qm3`, which do get `--verse-text-start` overrides and have their own margins in the stylesheet. If those markers are genuinely missing a `--para-indent` setter, the same gutter-overlap symptom this PR fixes for `\li`/`\lim` would still exist for `\qm2`/`\qm3`, and the new guard won't catch it because they aren't listed. Worth confirming against the upstream test's marker list.
</blockquote></details>

Confirmed it's a real gap: on the companion PR #518 branch (the source of the `\li`/`\lim` fix), `\qm2`/`\qm3` get only `--verse-text-start` and no gutter `--para-indent`, even though they share the same `margin-left: 20vw` as `\qm`/`\qm1` — which _do_ get `--para-indent: 20vw`. So their gutter marker overlaps the text, same symptom as `\li`/`\lim`. It's pre-existing upstream rather than something this PR introduces, and adding the entries here would diverge this copy from its source, so I've kept it out of this sync and filed **PT-4313** to fix it upstream and sync down. (The audit for that ticket also turned up the same gap for `\iq*`, `\ph*`, `\tr*`, and a few others.)

PT-4313: [https://paratextstudio.atlassian.net/browse/PT-4313](https://paratextstudio.atlassian.net/browse/PT-4313)

imnasnainaec (commented):
> @imnasnainaec reviewed all commit messages, made 3 comments, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2629)* status: 1 of 2 files reviewed, 3 unresolved discussions (waiting on katherinejensen00).

___
*[`extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts` line 18 at r3](https://reviewable.io/reviews/paranext/paranext-core/2629#-OzMB-rC5M-ypSYRmMmj:-OzMB-rC5M-ypSYRmMmk:bdkzh60) ([raw file](https://github.com/paranext/paranext-core/blob/1d52488b3b92979f96f63d6a3919a92e417a6326/extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts#L18)):*
Devin concern:
> extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts:16
**Line-comment stripping would corrupt the stylesheet parse if a protocol-relative or full URL ever appears in a declaration**
The `//` strip runs over the whole file after block comments are removed. Today every `//` sits inside the header block comment, so this is safe, but a future `url(https://…)` or `url(//…)` in a declaration would have the rest of that line (including its `;` and possibly `}`) deleted, silently shifting brace depth and block boundaries and making the coverage checks unreliable. A cheap hardening is to only strip `//` at a line start / after whitespace and not inside `url(...)`.
___
*[`extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts` line 61 at r3](https://reviewable.io/reviews/paranext/paranext-core/2629#-OzMBBaa4c_fp2cFke5c:-OzMBBaa4c_fp2cFke5d:b-x74snw) ([raw file](https://github.com/paranext/paranext-core/blob/1d52488b3b92979f96f63d6a3919a92e417a6326/extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts#L61)):*
Devin nitpick:
> extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts:R61-78
> 
> **Coverage map is last-wins and ignores extra/unexpected markers**
> 
> `getGutterMarkerValues` overwrites earlier values with later ones (`values.set(marker, value)`), which matches CSS cascade only when specificity is equal — true today for all these rules, but a higher-specificity earlier rule would make the test report the wrong "actual". The test also only checks expected markers; a marker given an indent that the USFM spec doesn't call for (or a stale entry left behind after upstream removes it) would not be flagged. Adding a reverse check (actual keys ⊆ expected keys) would make the drift guard symmetric.
___
*[`extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts` line 103 at r3](https://reviewable.io/reviews/paranext/paranext-core/2629#-OzMBdWmEkRXXj5NtBgS:-OzMBdWmEkRXXj5NtBgT:bfw4jg7) ([raw file](https://github.com/paranext/paranext-core/blob/1d52488b3b92979f96f63d6a3919a92e417a6326/extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts#L103)):*
Devin RTL concern:
> extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts:R103-110
> 
> **Test asserts pi2/pi3 indents that only exist for LTR, so the RTL gap stays invisible**
> 
> `EXPECTED_PARA_INDENT` lists `pi2` (10vw) and `pi3` (15vw), but the stylesheet only defines them under `.psc-gutter-markers.text-spacing[dir='ltr']` (`extensions/src/platform-scripture-editor/src/_usj-nodes.scss:2579`, `:2592`). The coverage filter only requires the selector text to contain `psc-gutter-markers` and `text-spacing`, so the `[dir='ltr']` qualifier is ignored and these entries pass even though RTL documents get no indent compensation for `\pi2`/`\pi3` (their base margins are set for both directions at `:1400`-ish style rules). The guard therefore reads stronger than it actually is for direction-qualified rules; worth either narrowing the filter or dropping the `[dir='ltr']` divergence so LTR/RTL behave alike.

katherinejensen00 (commented):
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2629)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on imnasnainaec).

___
*[`extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts` line 18 at r3](https://reviewable.io/reviews/paranext/paranext-core/2629#-OzMB-rC5M-ypSYRmMmj:-OzMsNn3--NZFUbgD5sl:b5o1uz) ([raw file](https://github.com/paranext/paranext-core/blob/1d52488b3b92979f96f63d6a3919a92e417a6326/extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts#L18)):*
<details><summary><i>Previously, imnasnainaec (D. Ror.) wrote…</i></summary><blockquote>

Devin concern:
> extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts:16
**Line-comment stripping would corrupt the stylesheet parse if a protocol-relative or full URL ever appears in a declaration**
The `//` strip runs over the whole file after block comments are removed. Today every `//` sits inside the header block comment, so this is safe, but a future `url(https://…)` or `url(//…)` in a declaration would have the rest of that line (including its `;` and possibly `}`) deleted, silently shifting brace depth and block boundaries and making the coverage checks unreliable. A cheap hardening is to only strip `//` at a line start / after whitespace and not inside `url(...)`.
</blockquote></details>

The strip ran over the whole file, so a future `url(https://…)` or `url(//…)` in a declaration would have deleted the rest of that line and shifted brace depth. Changed the regex to only strip `//` at line-start or after whitespace (`/(^|\s)\/\/.*$/gm` → `'$1'`), so `//` inside a value is left intact.
___
*[`extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts` line 61 at r3](https://reviewable.io/reviews/paranext/paranext-core/2629#-OzMBBaa4c_fp2cFke5c:-OzMsRHE-_-kZFfGlOMC:b-abxae) ([raw file](https://github.com/paranext/paranext-core/blob/1d52488b3b92979f96f63d6a3919a92e417a6326/extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts#L61)):*
<details><summary><i>Previously, imnasnainaec (D. Ror.) wrote…</i></summary><blockquote>

Devin nitpick:
> extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts:R61-78
> 
> **Coverage map is last-wins and ignores extra/unexpected markers**
> 
> `getGutterMarkerValues` overwrites earlier values with later ones (`values.set(marker, value)`), which matches CSS cascade only when specificity is equal — true today for all these rules, but a higher-specificity earlier rule would make the test report the wrong "actual". The test also only checks expected markers; a marker given an indent that the USFM spec doesn't call for (or a stale entry left behind after upstream removes it) would not be flagged. Adding a reverse check (actual keys ⊆ expected keys) would make the drift guard symmetric.
</blockquote></details>

Added `unexpectedMarkers()` (`actual ⊆ expected`) and wired it into both assertions, making the drift guard symmetric: an indent the USFM spec doesn't call for, or a stale entry after upstream removes a marker, now fails the test.
___
*[`extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts` line 103 at r3](https://reviewable.io/reviews/paranext/paranext-core/2629#-OzMBdWmEkRXXj5NtBgS:-OzMtLDyCDQ9LtvDRXJ3:b-q2s0ty) ([raw file](https://github.com/paranext/paranext-core/blob/1d52488b3b92979f96f63d6a3919a92e417a6326/extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts#L103)):*
<details><summary><i>Previously, imnasnainaec (D. Ror.) wrote…</i></summary><blockquote>

Devin RTL concern:
> extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts:R103-110
> 
> **Test asserts pi2/pi3 indents that only exist for LTR, so the RTL gap stays invisible**
> 
> `EXPECTED_PARA_INDENT` lists `pi2` (10vw) and `pi3` (15vw), but the stylesheet only defines them under `.psc-gutter-markers.text-spacing[dir='ltr']` (`extensions/src/platform-scripture-editor/src/_usj-nodes.scss:2579`, `:2592`). The coverage filter only requires the selector text to contain `psc-gutter-markers` and `text-spacing`, so the `[dir='ltr']` qualifier is ignored and these entries pass even though RTL documents get no indent compensation for `\pi2`/`\pi3` (their base margins are set for both directions at `:1400`-ish style rules). The guard therefore reads stronger than it actually is for direction-qualified rules; worth either narrowing the filter or dropping the `[dir='ltr']` divergence so LTR/RTL behave alike.
</blockquote></details>

Correct, and this exposed a real rendering bug rather than just a weak assertion. Checked upstream `usj-nodes.css`: it applies pi2/pi3 gutter `--para-indent` in **both** directions and explicitly notes they're direction-independent (same 10vw/15vw primary indent for LTR and RTL). The local `.psc-gutter-markers.text-spacing[dir='ltr']` qualifier on pi2/pi3 was drift from upstream, so RTL documents got **no** gutter indent compensation for `\pi2`/`\pi3` — exactly the kind of regression this coverage test exists to catch, and the filter's substring match was hiding it.

Fixed by dropping the `[dir='ltr']` divergence so LTR/RTL behave alike (matching upstream). Note this is the base-margin rules' _gutter compensation_ only — the base `[dir='ltr']`/`[dir='rtl']` margin split for pi2/pi3 is upstream-correct and left untouched.

To keep the filter from masking this again, added `directionQualifiedGutterRules()`, which fails if any gutter `--para-indent`/`--verse-text-start` rule carries a `[dir=…]` qualifier. Verified it fails when the qualifier is re-introduced, then restored the fix.

imnasnainaec (commented):
> @imnasnainaec reviewed 1 file and all commit messages, made 1 comment, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2629)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on katherinejensen00).

___
*[`extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts` line 147 at r4](https://reviewable.io/reviews/paranext/paranext-core/2629#-OzN4Iid3j4c6DuCe9Vk:-OzN4Iid3j4c6DuCe9Vl:bna8akk) ([raw file](https://github.com/paranext/paranext-core/blob/e3973d0707a035ba5b5eb75a745d73c44fb19fbf/extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts#L147)):*
> ```XML
>   '30vw': ['lim4'], // 1.5"
> });
> ```

Possibly out-of-scope Devin concern:

extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts:R139-171

**Marker labels still overlap text for several other indented paragraph styles, and the new guard does not catch it**

The list of paragraph styles the new check requires an indent entry for (`EXPECTED_PARA_INDENT` at `extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts:139-146`) omits many styles that are indented in the stylesheet, so the check passes while those styles keep the exact overlap problem this change set out to fix.
Impact: In the editor's margin view, the small style labels for introduction poetry, deeper-level poetry, hanging paragraphs and table rows still sit on top of the paragraph text.

<details>
<summary>Which markers are indented in the stylesheet but have no gutter compensation</summary>

Comparing the non-gutter `.text-spacing[dir='ltr'] .usfm_X { margin-left: … }` rules with the `.psc-gutter-markers.text-spacing` `--para-indent` groups, these markers have a non-zero left margin but no `--para-indent` setter: `ipq`, `imq`, `ipr` (5vw), `iq`, `iq1`, `iq2`, `iq3`, `qm2`, `qm3` (20vw), `qd` (5vw), `ph`, `ph1` (10vw), `ph2` (15vw), `ph3` (20vw), `p2` (2.5vw), `tr`, `tr1` (10vw), `tr2` (15vw), `psi` (5vw). Similarly `ph`, `ph1`, `ph2`, `ph3`, `tr`, `tr1`, `tr2` have `text-indent: -5vw` but no `--verse-text-start`.

Because `unexpectedMarkers()`/`valueMismatches()` (`extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts:93-114`) only compare against the hand-written expected maps, a marker that is absent from BOTH the stylesheet and the expected map is silently uncovered — precisely the drift mode the test claims to prevent.

</details>

katherinejensen00 (commented):
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2629)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on imnasnainaec).

___
*[`extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts` line 147 at r4](https://reviewable.io/reviews/paranext/paranext-core/2629#-OzN4Iid3j4c6DuCe9Vk:-OzNjiMt741GKlLeC0oO:bdrdwh7) ([raw file](https://github.com/paranext/paranext-core/blob/e3973d0707a035ba5b5eb75a745d73c44fb19fbf/extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts#L147)):*
<details><summary><i>Previously, imnasnainaec (D. Ror.) wrote…</i></summary><blockquote>

Possibly out-of-scope Devin concern:

extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts:R139-171

**Marker labels still overlap text for several other indented paragraph styles, and the new guard does not catch it**

The list of paragraph styles the new check requires an indent entry for (`EXPECTED_PARA_INDENT` at `extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts:139-146`) omits many styles that are indented in the stylesheet, so the check passes while those styles keep the exact overlap problem this change set out to fix.
Impact: In the editor's margin view, the small style labels for introduction poetry, deeper-level poetry, hanging paragraphs and table rows still sit on top of the paragraph text.

<details>
<summary>Which markers are indented in the stylesheet but have no gutter compensation</summary>

Comparing the non-gutter `.text-spacing[dir='ltr'] .usfm_X { margin-left: … }` rules with the `.psc-gutter-markers.text-spacing` `--para-indent` groups, these markers have a non-zero left margin but no `--para-indent` setter: `ipq`, `imq`, `ipr` (5vw), `iq`, `iq1`, `iq2`, `iq3`, `qm2`, `qm3` (20vw), `qd` (5vw), `ph`, `ph1` (10vw), `ph2` (15vw), `ph3` (20vw), `p2` (2.5vw), `tr`, `tr1` (10vw), `tr2` (15vw), `psi` (5vw). Similarly `ph`, `ph1`, `ph2`, `ph3`, `tr`, `tr1`, `tr2` have `text-indent: -5vw` but no `--verse-text-start`.

Because `unexpectedMarkers()`/`valueMismatches()` (`extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts:93-114`) only compare against the hand-written expected maps, a marker that is absent from BOTH the stylesheet and the expected map is silently uncovered — precisely the drift mode the test claims to prevent.

</details>
</blockquote></details>

Real observation, and I verified the list against the stylesheet — those markers do have a base margin with no gutter `--para-indent`, so their labels overlap. But this is the same upstream-completeness gap I confirmed for `\qm2`/`\qm3` earlier in this review, now tracked in **PT-4313** — whose audit already extends to exactly these (`\iq*`, `\ph*`, `\tr*`, and the rest).

The distinction that keeps it out of this PR: it isn't _drift_. This copy faithfully matches upstream `usj-nodes.css`, which itself lacks that gutter compensation. This test's job is to catch the copy diverging from upstream (the regression that lost `\li`/`\lim`), not to assert full USFM indent completeness. Adding the missing entries here would diverge the copy from its source, and adding the symmetric "every base-margin marker must have gutter compensation" check you're describing would go red immediately against upstream's own gaps — so both belong in PT-4313, applied upstream and synced down, not layered on here. I'll add that symmetric base-vs-gutter check as part of PT-4313 once upstream is complete.

PT-4313: [https://paratextstudio.atlassian.net/browse/PT-4313](https://paratextstudio.atlassian.net/browse/PT-4313)

imnasnainaec (approved):
> @imnasnainaec reviewed 1 file and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2629)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2631, opened 2026-08-03
PT-4302: Fix false 'no internet' error on first-run registration probe

katherinejensen00:
# Code Review Summary

**Branch**: PT-4302

**Base**: origin/main

**Date**: 2026-08-03

**Files changed**: 17

## Overview

PT-4302 fixes a first-run onboarding bug: on upgrade, the gate probed the C# registration
provider once, and a transient `'unknown'` (provider still starting during a busy upgrade) was
routed straight to an error screen that falsely blamed the network. The branch (1) retries a
transient `'unknown'` before giving up (default 3 attempts, 2s backoff, 15s per-attempt timeout)
and rewords the microcopy to stop blaming the connection; (2) hardens the retry/logging and adds a
test that guards the reworded error-body string against a silent blank-screen regression; and (3)
adds a loading watchdog that reveals a "continue without finishing setup" escape hatch after ~15s
so a genuinely-stuck startup is never a dead end, backed by a store-level generation guard so a
late-settling probe can neither clobber the user's status nor leave persisted wizard state (the
guard suppresses a superseded run's durable writes, not just its in-memory status update).

The branch is six focused commits (retry-probe fix → review-feedback hardening → loading watchdog
→ `retryUntil` extraction + escape-hatch reword → persisted-write gating → routing
`requestWithRetry` through the shared `retryUntil`) rebased onto the latest `origin/main`.

## API Changes

No public (extension-facing) API changed. `lib/platform-bible-utils` was not modified.

Internal-only changes:
- New internal shared util `src/shared/utils/retry.util.ts` exporting `retryUntil(attempt, isDone, { maxAttempts?, delayMs? })`. Not part of any `@papi/`-exported surface.
- `resolveRegistrationValidity()` (renderer-internal) signature widened with two optional, behavior-preserving params (`maxAttempts`, `retryDelayMs`).
- `requestWithRetry()` (`src/shared/data/rpc.model.ts`, main-process RPC dispatch) now routes through `retryUntil` instead of a hand-rolled loop; behavior preserved, the unused `wait` import dropped, and it gained its first unit tests.
- New localization keys `%firstRun_button_continueWithoutFinishingSetup%` and `%firstRun_loading_slow%`; `%firstRun_error_body%` and `%firstRun_button_continueWithoutSetup%` deprecated in `metadata.json` (replaced by `%firstRun_error_body_2%` and `%firstRun_button_continueWithoutFinishingSetup%`).

None.

None.

### Minor — Consider

- [x] **`first-run-store.ts` — `resolutionGeneration` guard was status-only.** The guard bumped by `continueWithoutRegistration` originally suppressed only the in-memory `setStatus`, not the durable writes on the `startWizard` path (`WIZARD_ACTIVE_KEY`, the interface-language seed). _(fixed during review: first added a comment on `completeFirstRun` explaining why it needn't supersede; then strengthened `resolveInternal` so the generation gate also suppresses the durable persisted writes of a superseded run — so a user who bails from the watchdog can't be resumed mid-wizard on the next launch. Covered by a new store test.)_
- [x] **Microcopy phrasing mismatch (button vs. body).** The button read "Continue without setup" while the body strings say "continue without finishing setup". _(fixed during review: unified on "finishing setup" — added `%firstRun_button_continueWithoutFinishingSetup%` in all 5 locales, deprecated the old key in `metadata.json`, and repointed both usages (error screen + watchdog). The old button string was already on `origin/main` from PT-4175, so a new-key-and-deprecate was required rather than an in-place edit.)_
- [ ] ~~Deprecated `%firstRun_error_body%` left in the locale files.~~ _(Dismissed: the key was introduced by PT-4175 and is already on `origin/main`, so keeping it plus a `metadata.json` deprecation entry is the required immutable-strings convention — not dead-string cruft. The compliance and style passes independently confirmed this is the correct pattern.)_

[Author response: Author directed both fixes. For the button, author explicitly chose the convention-safe new-key-and-deprecate path after confirming the old key predates this PR. Author independently identified and fixed the durable-write gap in the generation guard, and requested the `requestWithRetry` de-duplication described below.]

#### Shared Regions Modified

#### Extension Config Changes

None. No files under `extensions/` and no build/CI/dependency files changed.

## Positive Observations

- Localization key-change convention followed exactly: reworded strings get a new key with the old one retained and deprecated in `metadata.json` (dated, with rationale) — not edited in place. The new-key suffix matches established repo precedent.
- All new user-visible text is localized across all five locale files (`en`, `es`, `fr`, `zh-hans`, `zh-hant`); no hardcoded strings introduced. Sentence-case, second-person voice, true U+2026 ellipsis and U+2014 em dash all conform to the UX guidelines.
- Strong, behavior-focused test coverage: retry-then-succeed, exhaustion, definitive-answer short-circuit, non-positive `maxAttempts` clamp, default wiring, and a fake-timer test proving the backoff is awaited and skipped after the last attempt; store tests proving `continueWithoutRegistration` supersedes a slow in-flight resolution AND blocks its durable writes; a component fake-timer test for the ~15s escape reveal; and an assertion that the reworded error body actually renders (guards against a blank error screen from a key typo).
- The `resolutionGeneration` guard is a correct fix for the late-settling-probe race and gates both the in-memory status and the durable persisted writes (`WIZARD_ACTIVE_KEY`, the interface-language seed), so a superseded run can't resume the wizard on the next launch — both paths tested.
- `retryUntil` extraction reuses the existing `wait` utility, is well-documented (names its call sites and the deliberately-excluded deadline/abort case), and now backs two real callers (`resolveRegistrationValidity` and `requestWithRetry`) — the latter previously untested and now covered by five new cases.
- The loading-watchdog escape hatch is deliberately placed outside the `role="status"` live region so it is not re-announced; the error state moves focus to Retry on mount — good accessibility detail.
- `SlowLoading` Storybook story (`slowRevealMs: 0`) makes the otherwise-15s-delayed state reviewable without waiting.

## Interview Notes

- **Stated purpose** (confirmed by author): the three-part fix described in the Overview.
- **Findings resolution**: two minor findings fixed in-review, one dismissed with a correct convention-based rationale. The author demonstrated clear understanding throughout — driving the scoping decision on whether deprecated keys could be removed based on whether they originated in this PR (they did not), and independently identifying that the generation guard needed to cover durable writes, not just in-memory status.
- **In-review refactor (`retryUntil`)**: extracted at the author's request to de-duplicate a retry pattern the earlier deep review found in three places. Home decision: placed in `src/shared/utils/retry.util.ts` rather than `platform-bible-utils`, because (a) `platform-bible-utils` is a published, extension-facing library and this is an internal helper, and (b) that library cannot currently be built in this worktree (pre-existing missing `@types/jest`/`@types/node`). `resolveRegistrationValidity` migrated; behavior preserved (all resolver tests green).
- **Second caller added** (author-requested): `requestWithRetry` (`src/shared/data/rpc.model.ts`, main-process RPC dispatch — called by `rpc-server.ts` and `rpc-websocket-listener.ts`) was migrated onto `retryUntil` as part of this PR. Because that function was previously untested, the migration adds five unit tests (success / non-MethodNotFound short-circuit, retry-then-succeed, give-up after `MAX_REQUEST_ATTEMPTS` returning the response, per-attempt + give-up logging) — a net coverage gain on hot-path code — and removes a dead, unreachable `throw`. This gives `retryUntil` two real callers.
- **Still deliberately excluded**: `requestSessionSyncWithBootRetry` (`startup-tasks.ts`) — a fundamentally different retry shape (monotonic `performance.now()` deadline, abort/freshness gates, variable backoff) that would be distorted by forcing it into the fixed-attempts helper.
- No areas of author non-understanding or AI-deferral were observed.

## In-Review Quality Check

Changes were made during review (two finding fixes, the `retryUntil` extraction + `requestWithRetry`
migration, and the persisted-write guard strengthening). Results:
- **Typecheck** (`tsc --noEmit`, repo-wide): clean for all changed files. Two pre-existing errors remain in `src/main` (a missing generated `release/app/buildInfo.json` and an unrelated `installExtensions` overload) — neither touched by this branch.
- **Lint** (`eslint`): clean on all changed files (0 errors).
- **Format** (`prettier` / lint-staged on commit): all changed files formatted.
- **Tests**: 79 affected tests pass — `retry.util.test.ts` (7), `resolve-registration-validity.test.ts` (12), `first-run-store.test.ts` (40), `first-run-overlay.component.test.tsx` (8), `rpc.model.test.ts` (12, including 5 new `requestWithRetry` cases).
- **Environment caveat**: `npm run build:platform-bible-utils` fails in this worktree on missing `@types/jest`/`@types/node` (pre-existing, unrelated to this branch). Not needed for this branch because `retryUntil` lives in `src/shared`, not the library. A full repo-wide `npm run lint` / `npm test` was not run exhaustively; verification was per changed file.

## Suggested Review Focus

- [ ] **Retry budget / UX**: for a genuinely-down provider the gate can spin ~32–49s before resolving; the ~15s loading watchdog mitigates this by offering an escape, but confirm the overall budget is acceptable.
- [ ] **`retryUntil` home**: `src/shared/utils/retry.util.ts` vs `platform-bible-utils`. Confirm the internal-helper placement is preferred over growing the published extension API.
- [ ] **RPC hot-path change**: `requestWithRetry` (`rpc.model.ts`) now routes through `retryUntil`. It is on the main-process dispatch path (`rpc-server.ts`, `rpc-websocket-listener.ts`); the change is behavior-preserving and now unit-tested (it was previously untested), but worth a focused look since it touches core RPC.
- [ ] **`resolutionGeneration` guard**: the store-level supersede mechanism now gates both status and durable writes — verify the invariant (and the `completeFirstRun` exemption) reads correctly.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2631)


jolierabideau (changes requested):
Thanks Katherine! Great job on the localization work and test cases! Left some comments from Claude. Only one is blocking - `retryUntil` seems like something that should live in PBU but maybe there was a problem with that?

> @jolierabideau reviewed 18 files and all commit messages, and made 17 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2631#-:-OzCWqNfD8WjW7cNzsUC:b7u5vde)* status: all files reviewed, 16 unresolved discussions (waiting on katherinejensen00).

___
*[`assets/localization/en.json` line 38 at r1](https://reviewable.io/reviews/paranext/paranext-core/2631#-OzCWPCt7GYIY1zbszi4:-OzCWPCt7GYIY1zbszi5:bvftv31) ([raw file](https://github.com/paranext/paranext-core/blob/bfbeb7f204175057874af4ff4c975974ba1484df/assets/localization/en.json#L38)):*
> ```JSON
>   "%firstRun_description%": "Complete the initial setup to start using {%product_name%}.",
>   "%firstRun_error_body%": "{%product_name%} couldn't reach the registration service. Check your connection and try again, or continue without finishing setup for now.",
>   "%firstRun_error_body_2%": "{%product_name%} couldn't verify your registration yet. It may still be starting up — retry in a moment, or continue without finishing setup for now.",
> ```

Take it or leave it: `_2` matches what we've done before (`%userProfile_networkSettings_2%`), so you're on solid ground. But these keys stick around forever, and something like `%firstRun_error_body_providerStartingUp%` would tell a future reader _why_ it replaced the original. `_2` tells them nothing.
___
*[`assets/localization/zh-hans.json` line 126 at r1](https://reviewable.io/reviews/paranext/paranext-core/2631#-OzCWVfkB7CDbWVu1_Jh:-OzCWVfkB7CDbWVu1_Ji:b9vc6hh) ([raw file](https://github.com/paranext/paranext-core/blob/bfbeb7f204175057874af4ff4c975974ba1484df/assets/localization/zh-hans.json#L126)):*
> ```JSON
>   "%Book.LAO%": "老底嘉书",
>   "%firstRun_button_back%": "返回",
>   "%firstRun_button_continueWithoutFinishingSetup%": "不完成设置继续",
> ```

In both Chinese locales the new value is character-for-character identical to the deprecated key's ("不完成设置继续" / "不完成設定繼續"). That's probably correct — the Chinese already said "without finishing setup" — but a translator opening this file now sees two identical entries and may well "fix" one of them. A `notes` entry in `metadata.json` would head that off.

While you're in the locale files: the new Spanish and French body copy is noticeably longer than the English, so it'd be worth an in-app look to make sure it doesn't overflow that centered `max-w-md` block.
___
*[`src/renderer/components/first-run/first-run-overlay.component.tsx` line 52 at r1](https://reviewable.io/reviews/paranext/paranext-core/2631#-OzCVGA2DsZdP1kVidPs:-OzCVGA2DsZdP1kVidPt:bewwf98) ([raw file](https://github.com/paranext/paranext-core/blob/bfbeb7f204175057874af4ff4c975974ba1484df/src/renderer/components/first-run/first-run-overlay.component.tsx#L52)):*
> ```XML
>  * resolve, short enough that a genuinely-stuck startup isn't a dead end.
>  */
> const REGISTRATION_SLOW_REVEAL_MS = 15_000;
> ```

This 15s is the same 15s as `REGISTRATION_RESOLVE_TIMEOUT_MS`, and the doc comment right here says so ("~15s ≈ one full probe attempt"). Since that constant is already exported and lives in the same process, deriving from it would keep the two in sync — otherwise the next person to tune the probe timeout quietly breaks the relationship you just documented.
___
*[`src/renderer/components/first-run/first-run-overlay.component.tsx` line 75 at r1](https://reviewable.io/reviews/paranext/paranext-core/2631#-OzCWb2LBuUOx61uNfde:-OzCWb2LBuUOx61uNfdf:bth6pl6) ([raw file](https://github.com/paranext/paranext-core/blob/bfbeb7f204175057874af4ff4c975974ba1484df/src/renderer/components/first-run/first-run-overlay.component.tsx#L75)):*
> ```XML
>   // against a stale state.
>   const [loadingIsSlow, setLoadingIsSlow] = useState(false);
>   useEffect(() => {
> ```

The reset and `clearTimeout` paths in this effect aren't covered. The comment says the watchdog re-arms on each loading entry, which is the right behavior — a test that advances past the threshold, flips to `error`, then back to `loading` would prove it isn't stuck latched on.
___
*[`src/renderer/components/first-run/first-run-overlay.component.tsx` line 91 at r1](https://reviewable.io/reviews/paranext/paranext-core/2631#-OzCV5kc3z6pfkf8rXGX:-OzCV5kc3z6pfkf8rXGY:b8120rb) ([raw file](https://github.com/paranext/paranext-core/blob/bfbeb7f204175057874af4ff4c975974ba1484df/src/renderer/components/first-run/first-run-overlay.component.tsx#L91)):*
> ```XML
>   // eslint-disable-next-line no-null/no-null
>   const retryButtonRef = useRef<HTMLButtonElement>(null);
>   useEffect(() => {
> ```

Heads up — this existing focus effect now has a new way to bite. Probing can run for ~49s total (three 15s attempts plus two 2s backoffs), so the escape hatch is sitting on screen and tab-reachable for a good 20–35s before the flip to `error`. If someone had deliberately tabbed over to "Continue without finishing setup", that flip snatches their focus away to Retry. Skipping the forced focus when something in the gate is already focused would avoid it.
___
*[`src/renderer/components/first-run/first-run-overlay.component.tsx` line 116 at r1](https://reviewable.io/reviews/paranext/paranext-core/2631#-OzCUVYTBpRTrTNm8aii:-OzCUVYTBpRTrTNm8aij:b-lnnxdi) ([raw file](https://github.com/paranext/paranext-core/blob/bfbeb7f204175057874af4ff4c975974ba1484df/src/renderer/components/first-run/first-run-overlay.component.tsx#L116)):*
> ```XML
> 
>         {status.kind === 'loading' && (
>           <div className="tw:flex tw:h-full tw:flex-col tw:items-center tw:justify-center tw:gap-4">
> ```

This container has no horizontal padding, while the error branch uses `tw:p-8` (and `FULL_SCREEN_CONTENT` zeroes padding out). The new "taking longer" sentence is a lot longer than the old loading copy — longer still in Spanish and French — so in a narrow window it'll press right up against the edges. `tw:max-w-md` only helps on wide screens. Mind adding matching padding here?

One more thing about this container: since it's `justify-center`, the spinner and both lines of text jump upward the moment the escape block appears. It's a sudden shift on a screen the user has been staring at, and it moves things out from under their cursor. Reserving the space ahead of time (or anchoring the spinner group) would make it feel much calmer.
___
*[`src/renderer/components/first-run/first-run-overlay.component.tsx` line 127 at r1](https://reviewable.io/reviews/paranext/paranext-core/2631#-OzCSaweA2hQnXqo3-jk:-OzCSaweA2hQnXqo3-jl:bwn55pa) ([raw file](https://github.com/paranext/paranext-core/blob/bfbeb7f204175057874af4ff4c975974ba1484df/src/renderer/components/first-run/first-run-overlay.component.tsx#L127)):*
> ```XML
>                 the role="status" live region so it isn't re-announced. continueWithoutRegistration
>                 supersedes the in-flight resolution so its late result can't reopen the gate. */}
>             {loadingIsSlow && (
> ```

Screen reader users won't know this appeared. I see the comment about keeping it outside `role="status"` so it doesn't get re-announced, and that instinct makes sense — but the result is that it's never announced at all. Nothing in the live region changes, and focus doesn't move, so after 15s a new button quietly shows up and the only way to find it is to guess and start tabbing.

Good news: you can have both. A polite live region announces _additions_ by default, so if you wrap just this revealed block in its own `aria-live="polite"`, it gets announced once and the spinner copy above never re-reads. The error branch already does the careful thing by moving focus to Retry — I'd love the watchdog to get the same treatment. Right now the escape hatch only really exists for sighted users.
___
*[`src/renderer/components/first-run/first-run-overlay.component.tsx` line 127 at r1](https://reviewable.io/reviews/paranext/paranext-core/2631#-OzCVPmg7PV9if3XanpR:-OzCVPmg7PV9if3XanpS:bqaen2i) ([raw file](https://github.com/paranext/paranext-core/blob/bfbeb7f204175057874af4ff4c975974ba1484df/src/renderer/components/first-run/first-run-overlay.component.tsx#L127)):*
> ```XML
>                 the role="status" live region so it isn't re-announced. continueWithoutRegistration
>                 supersedes the in-flight resolution so its late result can't reopen the gate. */}
>             {loadingIsSlow && (
> ```

Small one: this button is now an exact copy of the one in the error branch — same variant, same handler, same label key. Pulling it into a little local component would keep them from drifting, and would give the accessibility fix above a single place to live.
___
*[`src/renderer/components/first-run/first-run-overlay.component.tsx` line 132 at r1](https://reviewable.io/reviews/paranext/paranext-core/2631#-OzCUQ304DFZJambYBc7:-OzCUQ31Ctty-7l_B5vo:b-6nxp8m) ([raw file](https://github.com/paranext/paranext-core/blob/bfbeb7f204175057874af4ff4c975974ba1484df/src/renderer/components/first-run/first-run-overlay.component.tsx#L132)):*
> ```XML
>                   {strings['%firstRun_loading_slow%']}
>                 </p>
>                 <Button variant="ghost" onClick={() => continueWithoutRegistration()}>
> ```

`ghost` is exactly right on the error screen, where it sits next to a filled Retry and is genuinely the lesser option. Here it's the only thing on the screen, and ghost is our quietest variant — barely visible edges. Someone who's been watching a spinner for 15 seconds should be able to _see_ the exit. `secondary` or `outline` would serve them better.
___
*[`src/renderer/components/first-run/first-run-overlay.component.test.tsx` line 214 at r1](https://reviewable.io/reviews/paranext/paranext-core/2631#-OzCUJS558R4iJLVfp_Q:-OzCUJS6Dxwxrot5uBiw:b-88cmjc) ([raw file](https://github.com/paranext/paranext-core/blob/bfbeb7f204175057874af4ff4c975974ba1484df/src/renderer/components/first-run/first-run-overlay.component.test.tsx#L214)):*
> ```XML
>   });
> 
>   it('reveals a continue-without-setup escape after loading stays slow (PT-4302)', () => {
> ```

This test checks that the button shows up, but never clicks it — and `continueWithoutRegistration` doesn't appear anywhere else in this file either. So neither escape hatch (this one or the error screen's) has a test proving the button actually calls the store. Your store-side supersession tests are great, but nothing ties the button to them, which means a missing or wrong `onClick` would sail through green.

One line should do it, matching the retry test right above: `expect(store.continueWithoutRegistration).toHaveBeenCalledOnce()`.
___
*[`src/renderer/services/first-run-store.ts` line 149 at r1](https://reviewable.io/reviews/paranext/paranext-core/2631#-OzCW4MEBrmItuDBnQq0:-OzCW4MEBrmItuDBnQq1:b-tm2sfw) ([raw file](https://github.com/paranext/paranext-core/blob/bfbeb7f204175057874af4ff4c975974ba1484df/src/renderer/services/first-run-store.ts#L149)):*
> ```XML
> 
> async function resolveInternal(generation: number): Promise<void> {
>   // True once a user action superseded this run mid-flight (e.g. "continue without setup" from the
> ```

Tiny accuracy thing on this comment: it says the gate keeps a superseded run from persisting state "including the interface-language seed", but `seedInterfaceLanguageFromOsLocale()` runs at line 268 and the `isSuperseded()` check comes after it at 271 — so a bail landing inside that function's own awaits can't stop the seed write. The window is very narrow (we're already past the probe by then) and the seed is harmless, so I'd just soften the wording rather than move the guard. The claim is a bit stronger than what the code delivers.
___
*[`src/renderer/services/resolve-registration-validity.ts` line 60 at r1](https://reviewable.io/reviews/paranext/paranext-core/2631#-OzCVh6b0gnagFwssKi5:-OzCVh6b0gnagFwssKi6:b-ifoz4p) ([raw file](https://github.com/paranext/paranext-core/blob/bfbeb7f204175057874af4ff4c975974ba1484df/src/renderer/services/resolve-registration-validity.ts#L60)):*
> ```XML
>  * so `'unknown'` is never a connectivity problem. On a busy upgrade/cold start the provider can
>  * take a while to come up, so we retry rather than immediately stranding a registered, connected
>  * user on the error screen (PT-4302). A definitive `'valid'`/`'invalid'` returns right away.
> ```

Code reference in comment
___
*[`src/renderer/services/resolve-registration-validity.ts` line 67 at r1](https://reviewable.io/reviews/paranext/paranext-core/2631#-OzCVe_R2OHmGbB1Ft21:-OzCVe_R2OHmGbB1Ft22:bx0w4ac) ([raw file](https://github.com/paranext/paranext-core/blob/bfbeb7f204175057874af4ff4c975974ba1484df/src/renderer/services/resolve-registration-validity.ts#L67)):*
> ```XML
>  * throughout.
>  */
> export async function resolveRegistrationValidity(
> ```

Three optional positional params is getting awkward — to override the third you have to restate the first two, and the tests already show the pain (`resolveRegistrationValidity(1000, undefined, 0)`). Nothing in production passes any arguments, so switching to an options object costs nothing right now, and it'd match the shape you used for `retryUntil({ maxAttempts, delayMs })` and `completeFirstRun(options?)` in this same PR.
___
*[`src/renderer/services/resolve-registration-validity.ts` line 81 at r1](https://reviewable.io/reviews/paranext/paranext-core/2631#-OzCVlr53nXEDsn1ca7t:-OzCVlr53nXEDsn1ca7u:b-ug1h7d) ([raw file](https://github.com/paranext/paranext-core/blob/bfbeb7f204175057874af4ff4c975974ba1484df/src/renderer/services/resolve-registration-validity.ts#L81)):*
> ```XML
>   if (validity === 'unknown') {
>     logger.warn(
>       `Could not resolve registration validity after ${Math.max(1, maxAttempts)} attempt(s); the provider is likely still starting up.`,
> ```

This `Math.max(1, maxAttempts)` repeats the clamp `retryUntil` already does internally — the comment just above even mentions it. Two copies of one rule tend to drift, and when they do, this log message starts reporting a number that isn't true. Clamping once into a local and using it in both places would settle it.
___
*[`src/shared/data/rpc.model.ts` line 219 at r1](https://reviewable.io/reviews/paranext/paranext-core/2631#-OzCW-5t15kjA5v12amO:-OzCW-5t15kjA5v12amP:bakh0ul) ([raw file](https://github.com/paranext/paranext-core/blob/bfbeb7f204175057874af4ff4c975974ba1484df/src/shared/data/rpc.model.ts#L219)):*
> ```XML
>     async (attemptNumber) => {
>       const response = await requestCallback();
>       const isMissingHandler =
> ```

The MethodNotFound check now exists twice, once here and once inverted in the `isDone` predicate below. Hand-inverted twins are the kind of thing that drifts under a later edit. One shared predicate used as `isMissingHandler(response)` and `(r) => !isMissingHandler(r)` would make them impossible to disagree.
___
*[`src/shared/utils/retry.util.ts` line 1 at r1](https://reviewable.io/reviews/paranext/paranext-core/2631#-OzCWFjN103yjVzXt_gF:-OzCWFjN103yjVzXt_gG:b-zfg898) ([raw file](https://github.com/paranext/paranext-core/blob/bfbeb7f204175057874af4ff4c975974ba1484df/src/shared/utils/retry.util.ts#L1)):*
> ```XML
> import { wait } from 'platform-bible-utils';
> ```

Not an objection, just want this to be a real decision: `retryUntil` is the same species as `wait` and `waitForDuration`, which live in `platform-bible-utils`. Keeping it in `src/shared/utils` also means it now surfaces as a new `declare module 'shared/utils/retry.util'` in `papi.d.ts`. Both homes are defensible — but if part of the reason was that `platform-bible-utils` wouldn't build locally, that's a tooling problem we should fix rather than let it pick the architecture.

katherinejensen00 (commented):
Thank you for the helpful code review comments, Jolie! 

The blocking comment is resolved — `retryUntil` now lives in `platform-bible-utils` and the rest are fixed too.

> @katherinejensen00 made 17 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2631#-:-OzDEbpJ-bQnEQd5Kfhu:bmqp4sm)* status: 3 of 33 files reviewed, 16 unresolved discussions (waiting on jolierabideau).

___
*[`assets/localization/en.json` line 38 at r1](https://reviewable.io/reviews/paranext/paranext-core/2631#-OzCWPCt7GYIY1zbszi4:-OzDDNKB8o9HQBFN1mGp:b351g7r) ([raw file](https://github.com/paranext/paranext-core/blob/bfbeb7f204175057874af4ff4c975974ba1484df/assets/localization/en.json#L38)):*

Take it or leave it: `_2` matches what we've done before (`%userProfile_networkSettings_2%`), so you're on solid ground. But these keys stick around forever, and something like `%firstRun_error_body_providerStartingUp%` would tell a future reader _why_ it replaced the original. `_2` tells them nothing.
</blockquote></details>

Fixed— renamed to `%firstRun_error_body_providerStartingUp%` across all five locales and updated the deprecation pointer in `metadata.json`. Since `_2` was introduced in this same (unmerged) PR and never shipped or went out to translators, an in-place rename is safe here — the immutability rule protects keys already on `main`. Agreed the descriptive name reads better; thanks.
___
*[`assets/localization/zh-hans.json` line 126 at r1](https://reviewable.io/reviews/paranext/paranext-core/2631#-OzCWVfkB7CDbWVu1_Jh:-OzDDTp_4f1liLsmMYBF:bcnsffk) ([raw file](https://github.com/paranext/paranext-core/blob/bfbeb7f204175057874af4ff4c975974ba1484df/assets/localization/zh-hans.json#L126)):*

In both Chinese locales the new value is character-for-character identical to the deprecated key's ("不完成设置继续" / "不完成設定繼續"). That's probably correct — the Chinese already said "without finishing setup" — but a translator opening this file now sees two identical entries and may well "fix" one of them. A `notes` entry in `metadata.json` would head that off.

While you're in the locale files: the new Spanish and French body copy is noticeably longer than the English, so it'd be worth an in-app look to make sure it doesn't overflow that centered `max-w-md` block.
</blockquote></details>

Added a `notes` entry on `%firstRun_button_continueWithoutFinishingSetup%` in `metadata.json` spelling out that the zh-Hans/zh-Hant values intentionally match the deprecated key ("不完成设置继续" / "不完成設定繼續"), so it reads as expected rather than a copy error. Good call. On the es/fr length: I've matched the loading container's padding to the error screen (`p-8`) and anchored the escape block so it can't crowd the edges, but I'll still do the in-app pass you suggested before merge to confirm the longer copy doesn't overflow.
___
*[`src/renderer/components/first-run/first-run-overlay.component.tsx` line 52 at r1](https://reviewable.io/reviews/paranext/paranext-core/2631#-OzCVGA2DsZdP1kVidPs:-OzDDXQzAe30issPFBB8:b-smgaak) ([raw file](https://github.com/paranext/paranext-core/blob/bfbeb7f204175057874af4ff4c975974ba1484df/src/renderer/components/first-run/first-run-overlay.component.tsx#L52)):*

This 15s is the same 15s as `REGISTRATION_RESOLVE_TIMEOUT_MS`, and the doc comment right here says so ("~15s ≈ one full probe attempt"). Since that constant is already exported and lives in the same process, deriving from it would keep the two in sync — otherwise the next person to tune the probe timeout quietly breaks the relationship you just documented.
</blockquote></details>

Fixed— `REGISTRATION_SLOW_REVEAL_MS` now equals the imported `REGISTRATION_RESOLVE_TIMEOUT_MS` and the comment notes the derivation, so retuning the probe timeout keeps the watchdog in step automatically.
___
*[`src/renderer/components/first-run/first-run-overlay.component.tsx` line 75 at r1](https://reviewable.io/reviews/paranext/paranext-core/2631#-OzCWb2LBuUOx61uNfde:-OzDDc33DQMtCHESDytk:b-4wgls9) ([raw file](https://github.com/paranext/paranext-core/blob/bfbeb7f204175057874af4ff4c975974ba1484df/src/renderer/components/first-run/first-run-overlay.component.tsx#L75)):*

The reset and `clearTimeout` paths in this effect aren't covered. The comment says the watchdog re-arms on each loading entry, which is the right behavior — a test that advances past the threshold, flips to `error`, then back to `loading` would prove it isn't stuck latched on.
</blockquote></details>

Added exactly that test — "re-arms the slow-loading watchdog after loading → error → loading": it advances past the threshold, flips to `error`, returns to `loading`, asserts the hatch is gone (proving the reset fired), then advances again to prove a fresh timer re-armed.
___
*[`src/renderer/components/first-run/first-run-overlay.component.tsx` line 91 at r1](https://reviewable.io/reviews/paranext/paranext-core/2631#-OzCV5kc3z6pfkf8rXGX:-OzDDkToEJh2UrrJoCKZ:bv710gu) ([raw file](https://github.com/paranext/paranext-core/blob/bfbeb7f204175057874af4ff4c975974ba1484df/src/renderer/components/first-run/first-run-overlay.component.tsx#L91)):*

Heads up — this existing focus effect now has a new way to bite. Probing can run for ~49s total (three 15s attempts plus two 2s backoffs), so the escape hatch is sitting on screen and tab-reachable for a good 20–35s before the flip to `error`. If someone had deliberately tabbed over to "Continue without finishing setup", that flip snatches their focus away to Retry. Skipping the forced focus when something in the gate is already focused would avoid it.
</blockquote></details>

Fixed — the error-mount focus effect now skips moving focus to Retry when a control in the gate already has focus (focus is trapped in the modal, so any focused button is one of ours). It only pulls focus to Retry when it would otherwise land on the non-interactive container.
___
*[`src/renderer/components/first-run/first-run-overlay.component.tsx` line 116 at r1](https://reviewable.io/reviews/paranext/paranext-core/2631#-OzCUVYTBpRTrTNm8aii:-OzDDqRGC3JYXdwGm0jw:bw0bnbd) ([raw file](https://github.com/paranext/paranext-core/blob/bfbeb7f204175057874af4ff4c975974ba1484df/src/renderer/components/first-run/first-run-overlay.component.tsx#L116)):*

This container has no horizontal padding, while the error branch uses `tw:p-8` (and `FULL_SCREEN_CONTENT` zeroes padding out). The new "taking longer" sentence is a lot longer than the old loading copy — longer still in Spanish and French — so in a narrow window it'll press right up against the edges. `tw:max-w-md` only helps on wide screens. Mind adding matching padding here?

One more thing about this container: since it's `justify-center`, the spinner and both lines of text jump upward the moment the escape block appears. It's a sudden shift on a screen the user has been staring at, and it moves things out from under their cursor. Reserving the space ahead of time (or anchoring the spinner group) would make it feel much calmer.
</blockquote></details>

Both addressed — added `p-8` to match the error screen, and anchored the escape hatch to the bottom (out of the centered flow) so the spinner group no longer shifts when it appears. I'll confirm the calmer feel in-app alongside the es/fr length check.
___
*[`src/renderer/components/first-run/first-run-overlay.component.tsx` line 127 at r1](https://reviewable.io/reviews/paranext/paranext-core/2631#-OzCSaweA2hQnXqo3-jk:-OzDDwaW8yHfC3O_PJ1B:bvi9m7) ([raw file](https://github.com/paranext/paranext-core/blob/bfbeb7f204175057874af4ff4c975974ba1484df/src/renderer/components/first-run/first-run-overlay.component.tsx#L127)):*

Screen reader users won't know this appeared. I see the comment about keeping it outside `role="status"` so it doesn't get re-announced, and that instinct makes sense — but the result is that it's never announced at all. Nothing in the live region changes, and focus doesn't move, so after 15s a new button quietly shows up and the only way to find it is to guess and start tabbing.

Good news: you can have both. A polite live region announces _additions_ by default, so if you wrap just this revealed block in its own `aria-live="polite"`, it gets announced once and the spinner copy above never re-reads. The error branch already does the careful thing by moving focus to Retry — I'd love the watchdog to get the same treatment. Right now the escape hatch only really exists for sighted users.
</blockquote></details>

Fixed both ways — the revealed block now has its own `aria-live="polite"` region (separate from the spinner's `role="status"`, so the spinner copy isn't re-read), and focus moves to the escape button when it reveals, mirroring the error screen's Retry treatment. The new store-call test also clicks it, so the wiring is covered.
___
*[`src/renderer/components/first-run/first-run-overlay.component.tsx` line 127 at r1](https://reviewable.io/reviews/paranext/paranext-core/2631#-OzCVPmg7PV9if3XanpR:-OzDE0aR0ZpDEzZeqi87:b-22dj1p) ([raw file](https://github.com/paranext/paranext-core/blob/bfbeb7f204175057874af4ff4c975974ba1484df/src/renderer/components/first-run/first-run-overlay.component.tsx#L127)):*

Small one: this button is now an exact copy of the one in the error branch — same variant, same handler, same label key. Pulling it into a little local component would keep them from drifting, and would give the accessibility fix above a single place to live.
</blockquote></details>

Extracted `ContinueWithoutSetupButton` (a small `forwardRef` local component) used by both branches, so the label and `continueWithoutRegistration` handler have one home. It takes a `variant` prop so the two contexts can still differ per your next comment.
___
*[`src/renderer/components/first-run/first-run-overlay.component.tsx` line 132 at r1](https://reviewable.io/reviews/paranext/paranext-core/2631#-OzCUQ304DFZJambYBc7:-OzDE6OS2u7IduMUMm2Q:b-bv5f99) ([raw file](https://github.com/paranext/paranext-core/blob/bfbeb7f204175057874af4ff4c975974ba1484df/src/renderer/components/first-run/first-run-overlay.component.tsx#L132)):*

`ghost` is exactly right on the error screen, where it sits next to a filled Retry and is genuinely the lesser option. Here it's the only thing on the screen, and ghost is our quietest variant — barely visible edges. Someone who's been watching a spinner for 15 seconds should be able to _see_ the exit. `secondary` or `outline` would serve them better.
</blockquote></details>

Fixed — the loading-watchdog button is now `variant="outline"`; the error-screen one stays `ghost` since it sits beside the filled Retry as the lesser option.
___
*[`src/renderer/components/first-run/first-run-overlay.component.test.tsx` line 214 at r1](https://reviewable.io/reviews/paranext/paranext-core/2631#-OzCUJS558R4iJLVfp_Q:-OzDEAyI5DpRb1agJPt3:brrtay0) ([raw file](https://github.com/paranext/paranext-core/blob/bfbeb7f204175057874af4ff4c975974ba1484df/src/renderer/components/first-run/first-run-overlay.component.test.tsx#L214)):*

This test checks that the button shows up, but never clicks it — and `continueWithoutRegistration` doesn't appear anywhere else in this file either. So neither escape hatch (this one or the error screen's) has a test proving the button actually calls the store. Your store-side supersession tests are great, but nothing ties the button to them, which means a missing or wrong `onClick` would sail through green.

One line should do it, matching the retry test right above: `expect(store.continueWithoutRegistration).toHaveBeenCalledOnce()`.
</blockquote></details>

Added the click-and-assert to the watchdog test, and the same assertion to the error-screen test, so both escape hatches are now proven to reach the store (I added `continueWithoutRegistration` to the store mock for this). Good catch — a dead `onClick` would have sailed through before.
___
*[`src/renderer/services/first-run-store.ts` line 149 at r1](https://reviewable.io/reviews/paranext/paranext-core/2631#-OzCW4MEBrmItuDBnQq0:-OzDEEbkDlHm-fDOY1j5:b-49rf0y) ([raw file](https://github.com/paranext/paranext-core/blob/bfbeb7f204175057874af4ff4c975974ba1484df/src/renderer/services/first-run-store.ts#L149)):*

Tiny accuracy thing on this comment: it says the gate keeps a superseded run from persisting state "including the interface-language seed", but `seedInterfaceLanguageFromOsLocale()` runs at line 268 and the `isSuperseded()` check comes after it at 271 — so a bail landing inside that function's own awaits can't stop the seed write. The window is very narrow (we're already past the probe by then) and the seed is harmless, so I'd just soften the wording rather than move the guard. The claim is a bit stronger than what the code delivers.
</blockquote></details>

Softened as suggested — the comment now says the guard covers the wizard-resume writes (`WIZARD_ACTIVE_KEY`, `firstRunComplete`) and notes the interface-language seed is only guarded at its boundaries: a bail landing inside the seed's own awaits can still write it, which is harmless (OS-matched UI language, never resumes the wizard). Left the guard where it is per your note.
___
*[`src/renderer/services/resolve-registration-validity.ts` line 60 at r1](https://reviewable.io/reviews/paranext/paranext-core/2631#-OzCVh6b0gnagFwssKi5:-OzDEIMg3Drz2ddh54h1:b-fara6a) ([raw file](https://github.com/paranext/paranext-core/blob/bfbeb7f204175057874af4ff4c975974ba1484df/src/renderer/services/resolve-registration-validity.ts#L60)):*

Code reference in comment
</blockquote></details>

No action needed — this looks like an automated annotation with no specific request; the comment's `resolveRegistrationValidity` reference is intentional. Let me know if you meant something more here.
___
*[`src/renderer/services/resolve-registration-validity.ts` line 67 at r1](https://reviewable.io/reviews/paranext/paranext-core/2631#-OzCVe_R2OHmGbB1Ft21:-OzDELzl6ThFmDL6wiuO:bjiwsy5) ([raw file](https://github.com/paranext/paranext-core/blob/bfbeb7f204175057874af4ff4c975974ba1484df/src/renderer/services/resolve-registration-validity.ts#L67)):*

Three optional positional params is getting awkward — to override the third you have to restate the first two, and the tests already show the pain (`resolveRegistrationValidity(1000, undefined, 0)`). Nothing in production passes any arguments, so switching to an options object costs nothing right now, and it'd match the shape you used for `retryUntil({ maxAttempts, delayMs })` and `completeFirstRun(options?)` in this same PR.
</blockquote></details>

Fixed — `resolveRegistrationValidity` now takes a single options object (`{ timeoutMs?, maxAttempts?, retryDelayMs? }`) with per-field defaults, matching the `retryUntil`/`completeFirstRun` shape. Updated all the test call sites (including the awkward `(1000, undefined, 0)` one).
___
*[`src/renderer/services/resolve-registration-validity.ts` line 81 at r1](https://reviewable.io/reviews/paranext/paranext-core/2631#-OzCVlr53nXEDsn1ca7t:-OzDEQG88450cCQNIXDe:bluzowb) ([raw file](https://github.com/paranext/paranext-core/blob/bfbeb7f204175057874af4ff4c975974ba1484df/src/renderer/services/resolve-registration-validity.ts#L81)):*

This `Math.max(1, maxAttempts)` repeats the clamp `retryUntil` already does internally — the comment just above even mentions it. Two copies of one rule tend to drift, and when they do, this log message starts reporting a number that isn't true. Clamping once into a local and using it in both places would settle it.
</blockquote></details>

Fixed — clamp once into `effectiveMaxAttempts` and pass that to both `retryUntil` and the log message, so the reported count can never disagree with the actual budget.
___
*[`src/shared/data/rpc.model.ts` line 219 at r1](https://reviewable.io/reviews/paranext/paranext-core/2631#-OzCW-5t15kjA5v12amO:-OzDEUGJ1y6dJ0B45DNa:b-x96o05) ([raw file](https://github.com/paranext/paranext-core/blob/bfbeb7f204175057874af4ff4c975974ba1484df/src/shared/data/rpc.model.ts#L219)):*

The MethodNotFound check now exists twice, once here and once inverted in the `isDone` predicate below. Hand-inverted twins are the kind of thing that drifts under a later edit. One shared predicate used as `isMissingHandler(response)` and `(r) => !isMissingHandler(r)` would make them impossible to disagree.
</blockquote></details>

Fixed — extracted a single `isMissingHandler(response)` predicate, used as `isMissingHandler(response)` on the attempt side and `(r) => !isMissingHandler(r)` in the stop condition, so the two can't drift.
___
*[`src/shared/utils/retry.util.ts` line 1 at r1](https://reviewable.io/reviews/paranext/paranext-core/2631#-OzCWFjN103yjVzXt_gF:-OzDEXq135xqgMhN2kc1:b-gtsrh4) ([raw file](https://github.com/paranext/paranext-core/blob/bfbeb7f204175057874af4ff4c975974ba1484df/src/shared/utils/retry.util.ts#L1)):*

Not an objection, just want this to be a real decision: `retryUntil` is the same species as `wait` and `waitForDuration`, which live in `platform-bible-utils`. Keeping it in `src/shared/utils` also means it now surfaces as a new `declare module 'shared/utils/retry.util'` in `papi.d.ts`. Both homes are defensible — but if part of the reason was that `platform-bible-utils` wouldn't build locally, that's a tooling problem we should fix rather than let it pick the architecture.
</blockquote></details>

Moved it to `platform-bible-utils` (in `util.ts`, right next to `wait`/`waitForDuration`), with its tests, and updated both callers. You were right that the local-build worry shouldn't drive the architecture — PBU builds and its full test suite passes cleanly here, so that concern doesn't hold. As a bonus, `papi.d.ts` regenerated to _remove_ the old `declare module 'shared/utils/retry.util'` surface with nothing added, since PBU functions aren't individually re-declared there.

jolierabideau (approved):
> @jolierabideau partially reviewed 31 files and all commit messages, and resolved 16 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2631)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2634, opened 2026-08-04
PT-4300: Fix first-run Internet Settings startup race

katherinejensen00:
## Summary

On a cold first launch, the wizard's **Internet Settings** step could reach the .NET data provider before it had registered its command handlers — so `getParatextDataInternetSettings` failed instantly with JSON-RPC **-32601** ("method not found") and dropped the user straight into an error screen. Navigating slowly avoided it, which is the tell for a startup race.

The step now **re-issues the command while the error is still "method not found," up to a ~30 s wall-clock budget**, instead of giving up on the first miss. While waiting it shows a delayed reassurance message; only once the budget is spent does it surface a **friendly, localized** error + Retry (the raw RPC error is logged, not shown to the user).

Two small refactors fell out of the fix: a **`useDelayedFlag`** hook and a shared **`StepLoading`** component.

## Where to review (roughly by importance)

| File | What it does |
| --- | --- |
| **`steps/internet-settings-step.component.tsx`** | **The fix.** `load()` is now a wall-clock retry loop — re-issue on method-not-found (matched via the shared `getJsonRpcRequestErrorMessagePrefix` producer, *not* a hardcoded `-32601`) until the budget, then a friendly error + Retry. Guarded against stale/overlapping loads. |
| **`…internet-settings-step.component.test.tsx`** | Race coverage: retry across more than one provider window, budget expiry, the delayed message, stale-load supersession, and the non-retryable-error path. |
| **`hooks/use-delayed-flag.hook.ts`** (+ test) | New hook — returns `true` once a condition has held for N ms; replaces the component's hand-rolled `setTimeout`/`clearTimeout`. |
| **`first-run/step-loading.component.tsx`** (+ story) | New shared spinner + message panel; de-dupes an identical inline block and adds an `aria-live` region. |
| **`steps/identify-step.component.tsx`** | Adopts `StepLoading` for its restart panel (gains the `aria-live` it was missing). |
| **`assets/localization/*.json`** | Two new keys across all 5 locales. |

## Key decisions

- **Why retry at all** — the command layer already retries a missing handler ~10 s, but registration can outlast a single window on a slow cold start, so we re-issue to span several. (The sibling `resolve-registration-validity.ts` handles the same race for a call it *can* abandon; this step needs the data, so it re-issues rather than degrading.)
- **Budget ≈ 30 s** is an educated guess (the sibling uses 15 s for its abandonable call) — flagged below for confirmation.
- **Friendly error copy** replaces the raw `-32601` in the UI; the raw message is `logger.warn`'d for debugging.

## Testing

119 tests pass (first-run dir + the new hook); typecheck clean; changed files lint clean.

## For the reviewer's attention (follow-ups, not blockers)

- **~30 s budget** — confirm against a real cold-start registration time; the constant is trivial to tune.
- **es/fr/zh-hans/zh-hant translations** of the two new strings are best-effort and want a native/translator pass.

<details>
<summary>📋 Full automated review summary (findings, positives, interview notes)</summary>

# Code Review Summary

**Branch**: pt-4300-internet-bug · **Base**: origin/main · **Date**: 2026-08-03 · **Review model**: Claude Opus 4.8 · **Files changed**: 13

## Findings

**Critical:** None. **Important:** None.

**Minor:**
- [x] **[UX] "data services" was technical/unfamiliar user-facing vocabulary** _(fixed during review: reworded to "Getting things ready…" and "We couldn't get things ready. Please try again in a moment." across all 5 locales, with test + story updates)_
- [x] **[Style] connecting microcopy hardcoded as a story literal could drift from `en.json`** _(fixed during review: updated the `StepLoading` story sample and the internet-settings story docstring)_
- [ ] ~~**[Localization] es/fr/zh-hans/zh-hant translations should be confirmed by a translator**~~ _(author: leave for now — best-effort machine translations; native pass to follow)_
- [ ] ~~**[API] method-not-found classified by message-substring rather than a machine-readable code**~~ _(author: leave for now — intended anti-drift pattern using the shared `getJsonRpcRequestErrorMessagePrefix` producer; revisit only if a richer error marker becomes available)_

## Positive Observations

- Method-not-found match derived from the shared `getJsonRpcRequestErrorMessagePrefix` producer (anti-drift); test fixtures build the same prefix from the same producer, so tests can't silently drift from production.
- Strong, behavior-focused tests: 5 falsifiable `useDelayedFlag` cases, plus a startup-race block covering multi-window retry, budget expiry, the delayed message, stale-load supersession, and the non-retryable path.
- `useDelayedFlag` removes hand-rolled timer bookkeeping (React-managed cleanup); `StepLoading` genuinely de-dupes an identical inline block **and** adds an `aria-live` region the inline version lacked.
- Load/save failures surface an inline destructive Alert (never silent); raw RPC error is logged, not shown. Localization present in all 5 locales; key naming follows the `%firstRun_step_<step>_<purpose>%` convention. Conventional file placement; no dead imports; stale-write guards (`loadGeneration` + `isMounted`).

## Interview Notes

- Purpose confirmed: fix the first-run Internet Settings startup race, plus the `useDelayedFlag` / `StepLoading` refactors.
- No Critical/Important findings; two Minor items fixed in-review, two dismissed with rationale.
- Author-flagged follow-ups: the ~30 s startup budget is an educated guess (vs the sibling's 15 s) pending a real cold-start measurement; the machine translations need a native/translator pass.

## In-Review Quality Check

- Typecheck passed (only 2 known pre-existing, unrelated errors). Tests: 119 pass (9 files). The 8 changed TS/TSX files lint clean (2 prettier warnings auto-fixed).
- Environment caveat: `npm run lint` can't complete end-to-end here (its `build:types` step fails on missing `@types/jest`/`@types/node` — an incomplete install, not a code defect). That step also spuriously regenerated `lib/papi-dts/papi.d.ts`, which was reverted; it is not part of this change. CI, with a full install, will run lint cleanly.

## Suggested Review Focus

- [ ] Startup budget value (~30 s vs the sibling's 15 s) — confirm against a real cold-start measurement.
- [ ] New/updated es/fr/zh-hans/zh-hant translations — native/translator pass.
- [ ] Reworded copy ("Getting things ready…" / "We couldn't get things ready…") — confirm it reads well.

</details>

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2634)


tombogle (changes requested):
> @tombogle reviewed 5 files and all commit messages, and made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2634)* status: 5 of 14 files reviewed, 3 unresolved discussions (waiting on katherinejensen00).

___
*[`assets/localization/en.json` line 51 at r1](https://reviewable.io/reviews/paranext/paranext-core/2634#-OzHRhv05LKj2Hi_rR4P:-OzHRhv05LKj2Hi_rR4Q:b-xnctti) ([raw file](https://github.com/paranext/paranext-core/blob/93d29bbcd14bf2d002f0119654bb706e29f26e5a/assets/localization/en.json#L51)):*
> ```JSON
>   "%firstRun_step_identify_validatingCode%": "Checking your registration…",
>   "%firstRun_step_internetSettings_connecting%": "Getting things ready…",
>   "%firstRun_step_internetSettings_loadError%": "We couldn't get things ready. Please try again in a moment.",
> ```

The user just waited 30 whopping seconds. Will they actually believe that waiting another "moment" is the key to success? I know, I'd be thinking, "I'm toast!"
Can this be skipped? If so, what are the implications of skipping? If they try again, are they possibly in for another 30s, or will it just be a quick check and fail? In practice, are you seeing situations where we might actually have to wait ~30 seconds, or are we just being generous?
The vagueness of "Getting things ready" (or just "Please wait") might be fine for an initial wait-state message, but if we are still stuck 30 seconds later, I think the user is going to need some kind of reassurance. Is there anything we can tell them about what is going wrong that will set their mind at ease (or tip them off to a problem they might be able to fix?) about the delay without swamping them with meaningless technical details? For instance, if this is happening because they have other processes that are using most of the available memory or doing horrendous disk-churning activities. do they need to know that so they can consider killing something that is killing their system?
___
*[`assets/localization/es.json` line 172 at r1](https://reviewable.io/reviews/paranext/paranext-core/2634#-OzHPh3Z-wv1BirTeL75:-OzHPh3_0x3UeN9SVrUm:b-uk3n3c) ([raw file](https://github.com/paranext/paranext-core/blob/93d29bbcd14bf2d002f0119654bb706e29f26e5a/assets/localization/es.json#L172)):*
> ```JSON
>   "%firstRun_step_identify_validatingCode%": "Verificando su registro…",
>   "%firstRun_step_internetSettings_connecting%": "Preparando todo…",
>   "%firstRun_step_internetSettings_loadError%": "No pudimos preparar todo. Vuelva a intentarlo en un momento.",
> ```

"Preparing everything" sounds at once, awkward, vague, and cosmically over-promising.
I would either say
Preparando…
OR
Espere un momento… (Wait a moment…)
The former is close to the French. If you're willing to depart further from the English, though, "Espere un momento…" is really what you're asking the user to do while the handlers become available.
___
*[`src/renderer/components/first-run/step-loading.component.tsx` line 23 at r1](https://reviewable.io/reviews/paranext/paranext-core/2634#-OzHX0XE7CcKpuEK1Awk:-OzHX0XE7CcKpuEK1Awl:bshp9j4) ([raw file](https://github.com/paranext/paranext-core/blob/93d29bbcd14bf2d002f0119654bb706e29f26e5a/src/renderer/components/first-run/step-loading.component.tsx#L23)):*
> ```XML
>           The top margin is applied only when populated so the spinner-only state stays tight. */}
>       <p
>         className={`tw:text-sm tw:text-muted-foreground${message ? 'tw:mt-4' : ''}`}
> ```

Missing a space. Should be:
```
className={`tw:text-sm tw:text-muted-foreground${message ? ' tw:mt-4' : ''}`}
```
OR
```
className={cn(
  'tw:text-sm',
  'tw:text-muted-foreground',
  message && 'tw:mt-4',
)}
```

katherinejensen00 (commented):
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2634)* status: 5 of 14 files reviewed, 3 unresolved discussions (waiting on tombogle).

___
*[`assets/localization/en.json` line 51 at r1](https://reviewable.io/reviews/paranext/paranext-core/2634#-OzHRhv05LKj2Hi_rR4P:-OzIQILA72FaNRjXUPdf:b9rwzfr) ([raw file](https://github.com/paranext/paranext-core/blob/93d29bbcd14bf2d002f0119654bb706e29f26e5a/assets/localization/en.json#L51)):*

The user just waited 30 whopping seconds. Will they actually believe that waiting another "moment" is the key to success? I know, I'd be thinking, "I'm toast!"
Can this be skipped? If so, what are the implications of skipping? If they try again, are they possibly in for another 30s, or will it just be a quick check and fail? In practice, are you seeing situations where we might actually have to wait ~30 seconds, or are we just being generous?
The vagueness of "Getting things ready" (or just "Please wait") might be fine for an initial wait-state message, but if we are still stuck 30 seconds later, I think the user is going to need some kind of reassurance. Is there anything we can tell them about what is going wrong that will set their mind at ease (or tip them off to a problem they might be able to fix?) about the delay without swamping them with meaningless technical details? For instance, if this is happening because they have other processes that are using most of the available memory or doing horrendous disk-churning activities. do they need to know that so they can consider killing something that is killing their system?
</blockquote></details>

Good push on the copy — I've reworded both strings across all 5 locales to be more honest and reassuring:

-   Waiting: _"Setting things up for the first time. This can take a moment…"_ — names _why_ it's slow (first-run setup) rather than a vague "please wait".
-   Error: _"Setup is taking longer than expected. Please try again — it usually works right away."_ — the "right away" is deliberate, because of your retry question below.

On your specific questions:

-   **Is Retry another 30s?** Almost always no. Retry re-runs the same load, and by the time the error shows, the provider's handlers have essentially always finished registering — so the very first attempt on Retry resolves quickly. Only a still-unregistered provider would consume the budget again, which is why the new copy promises "right away."
-   **Is 30s realistic?** It's generous on purpose, not a measured typical wait — the common cold start clears in a few seconds; the budget just has to outlast a slow registration window. I've left it flagged as tunable pending a real cold-start measurement.
-   **Can it be skipped?** Not today — on failure the step offers Retry and Next stays disabled. Adding a skip means finishing first-run with internet settings unconfigured (defaults applied), which is a product decision I'd rather not make unilaterally in this bug-fix PR. Happy to file a follow-up if you think it's worth it.
-   **Surfacing memory/disk as the cause?** I looked at this and decided against it: we can't actually attribute the delay to memory or disk from here (the fetch only knows "the handler isn't registered yet"), so any such message would be a guess that could send the user chasing the wrong problem. I kept the raw RPC error in the log for debugging instead.
___
*[`assets/localization/es.json` line 172 at r1](https://reviewable.io/reviews/paranext/paranext-core/2634#-OzHPh3Z-wv1BirTeL75:-OzIQPf6CHYACzwY5OrG:b-2dm5q2) ([raw file](https://github.com/paranext/paranext-core/blob/93d29bbcd14bf2d002f0119654bb706e29f26e5a/assets/localization/es.json#L172)):*

"Preparing everything" sounds at once, awkward, vague, and cosmically over-promising.
I would either say
Preparando…
OR
Espere un momento… (Wait a moment…)
The former is close to the French. If you're willing to depart further from the English, though, "Espere un momento…" is really what you're asking the user to do while the handlers become available.
</blockquote></details>

Ha ha ha. Agreed!

Agreed — "todo" was doing too much. Since the English now leads with the first-time-setup framing, I matched that and dropped "todo":

-   Waiting: _"Preparando la configuración inicial. Esto puede tardar un momento…"_
-   Error: _"La preparación está tardando más de lo esperado. Vuelva a intentarlo; suele funcionar de inmediato."_

This keeps the honest "first-time setup can take a moment" tone without the over-promise. Still best-effort machine translation pending a native/translator pass, so please flag anything that reads awkwardly.
___
*[`src/renderer/components/first-run/step-loading.component.tsx` line 23 at r1](https://reviewable.io/reviews/paranext/paranext-core/2634#-OzHX0XE7CcKpuEK1Awk:-OzIQSWG-HL90tCaiiyZ:bb3t09u) ([raw file](https://github.com/paranext/paranext-core/blob/93d29bbcd14bf2d002f0119654bb706e29f26e5a/src/renderer/components/first-run/step-loading.component.tsx#L23)):*

Missing a space. Should be:
```
className={`tw:text-sm tw:text-muted-foreground${message ? ' tw:mt-4' : ''}`}
```
OR
```
className={cn(
  'tw:text-sm',
  'tw:text-muted-foreground',
  message && 'tw:mt-4',
)}
```
</blockquote></details>

Fixed — went with your `cn(...)` option, because the plain-space version doesn't survive our tooling: `prettier-plugin-tailwindcss` normalizes class strings and strips a hand-inserted separating space on format, so `' tw:mt-4'` reformats back to `'tw:mt-4'` and re-glues the classes. `cn` passes each class as a separate argument, so the plugin can't collapse them:

```
className={cn('tw:text-sm tw:text-muted-foreground', message && 'tw:mt-4')}
```

(I also added `cn` to the three `platform-bible-react` test mocks that render this component.)

tombogle (changes requested):
> @tombogle reviewed 15 files and all commit messages, made 8 comments, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2634)* status: all files reviewed, 8 unresolved discussions (waiting on katherinejensen00).

___
*[`assets/localization/en.json` line 50 at r2](https://reviewable.io/reviews/paranext/paranext-core/2634#-OzJfqhO4bGQwubKq0jW:-OzJfqhO4bGQwubKq0jX:bxpaf61) ([raw file](https://github.com/paranext/paranext-core/blob/69e706c58d9737744635219a2058048bde5aa7d8/assets/localization/en.json#L50)):*
> ```JSON
>   "%firstRun_step_identify_registryLink%": "Visit Paratext Registry",
>   "%firstRun_step_identify_validatingCode%": "Checking your registration…",
>   "%firstRun_step_internetSettings_connecting%": "Setting things up for the first time. This can take a moment…",
> ```

Consider: `Setting things up for the first time. This can take a few moments…` (I'm not sure how long a moment is, but it might be less than 30 seconds.)
___
*[`assets/localization/es.json` line 171 at r2](https://reviewable.io/reviews/paranext/paranext-core/2634#-OzJYqVcCiRBZPKFLvFw:-OzJYqVcCiRBZPKFLvFx:byox4z7) ([raw file](https://github.com/paranext/paranext-core/blob/69e706c58d9737744635219a2058048bde5aa7d8/assets/localization/es.json#L171)):*
> ```JSON
>   "%firstRun_step_identify_registryLink%": "Visite el Registro de Paratext",
>   "%firstRun_step_identify_validatingCode%": "Verificando su registro…",
>   "%firstRun_step_internetSettings_connecting%": "Preparando la configuración inicial. Esto puede tardar un momento…",
> ```

Configurando por primera vez. Esto puede tardar unos momentos…
___
*[`assets/localization/es.json` line 172 at r2](https://reviewable.io/reviews/paranext/paranext-core/2634#-OzJevgwAlwPzVYjy5-A:-OzJevgwAlwPzVYjy5-B:b-3h5q85) ([raw file](https://github.com/paranext/paranext-core/blob/69e706c58d9737744635219a2058048bde5aa7d8/assets/localization/es.json#L172)):*
> ```JSON
>   "%firstRun_step_identify_validatingCode%": "Verificando su registro…",
>   "%firstRun_step_internetSettings_connecting%": "Preparando la configuración inicial. Esto puede tardar un momento…",
>   "%firstRun_step_internetSettings_loadError%": "La preparación está tardando más de lo esperado. Vuelva a intentarlo; suele funcionar de inmediato.",
> ```

La configuración inicial está tardando más de lo esperado. Vuelva a intentarlo; normalmente se resuelve enseguida.
___
*[`src/renderer/components/first-run/step-loading.component.tsx` line 21 at r2](https://reviewable.io/reviews/paranext/paranext-core/2634#-OzJfIAw5Gi_CFHCqY6f:-OzJfIAw5Gi_CFHCqY6g:b-r5ubt7) ([raw file](https://github.com/paranext/paranext-core/blob/69e706c58d9737744635219a2058048bde5aa7d8/src/renderer/components/first-run/step-loading.component.tsx#L21)):*
> ```XML
>       {/* The live region is kept always mounted (empty until there's a message) so it exists BEFORE
>           its text appears — otherwise screen readers don't announce a message that pops in later.
>           The top margin is applied only when populated (via cn, not string concatenation — the
> ```

NIT: I'd probably remove the excessively detailed AI comment about why you used `cn`. We use `cn` a lot, and nobody is going to think it's weird that you did.
___
*[`src/renderer/components/first-run/steps/internet-settings-step.component.tsx` line 24 at r2](https://reviewable.io/reviews/paranext/paranext-core/2634#-OzJhIRZ1n7NMl19TvtV:-OzJhIRZ1n7NMl19TvtW:boa46mt) ([raw file](https://github.com/paranext/paranext-core/blob/69e706c58d9737744635219a2058048bde5aa7d8/src/renderer/components/first-run/steps/internet-settings-step.component.tsx#L24)):*
> ```XML
> 
> // `internetSettings_*` keys come from the paratext-registration extension, `firstRun_*` from core
> // (assets/localization). Both merge in the combiner, so the extension keys need no en.json entry.
> ```

I think *maybe* I understand what this is trying to say, but it's a little vague and confusing. Those localization keys *are* indeed in a file called `en.json`.
___
*[`src/renderer/components/first-run/steps/internet-settings-step.component.tsx` line 41 at r2](https://reviewable.io/reviews/paranext/paranext-core/2634#-OzJiIUf5xRkb6Ts7cK3:-OzJiIUf5xRkb6Ts7cK4:b-zfya1e) ([raw file](https://github.com/paranext/paranext-core/blob/69e706c58d9737744635219a2058048bde5aa7d8/src/renderer/components/first-run/steps/internet-settings-step.component.tsx#L41)):*
> ```XML
> );
> // Each call is bounded by the ~30 s client request timeout; method-not-found returns in ~10 s
> // (command-layer retry), so several attempts fit, while a full client timeout consumes the budget in
> ```

If the "attempts" here are intended to mean the *outer* retry attempts, then "so up to three outer retries fit within the budget" would be more accurate. In reality, it will likely only be 2 because of overhead. If we're really talking about the total attempts (inner and outer) then "several" feels pretty accurate, but maybe the wording should be tightened to make it clear what attempts we mean.
___
*[`src/renderer/hooks/use-delayed-flag.hook.ts` line 14 at r2](https://reviewable.io/reviews/paranext/paranext-core/2634#-OzJjrxkBS3ZveUcFpNX:-OzJjrxl03byJS-FKckJ:bhv0mvt) ([raw file](https://github.com/paranext/paranext-core/blob/69e706c58d9737744635219a2058048bde5aa7d8/src/renderer/hooks/use-delayed-flag.hook.ts#L14)):*
> ```XML
>  * @param delayMs How long `active` must stay `true` before this returns `true`
>  */
> export function useDelayedFlag(active: boolean, delayMs: number): boolean {
> ```

if `delayMs` is a UX-tuned number that is likely to be a constant in the common case, maybe we should make it optional and supply a default rather than making every caller reason as to what the number should eb or look at other callers to find the UX-approved delay.
___
*[`src/renderer/hooks/use-delayed-flag.hook.test.ts` line 15 at r2](https://reviewable.io/reviews/paranext/paranext-core/2634#-OzJkMGs3RHExVGsKsmV:-OzJkMGs3RHExVGsKsmW:bi29a0h) ([raw file](https://github.com/paranext/paranext-core/blob/69e706c58d9737744635219a2058048bde5aa7d8/src/renderer/hooks/use-delayed-flag.hook.test.ts#L15)):*
> ```XML
> 
>   it('stays false until active has been true for the full delay', () => {
>     const { result } = renderHook(() => useDelayedFlag(true, 2000));
> ```

All of these tests pass 2000. The implementation could ignore the parameter and hard-code 2000 and the tests would still pass.

katherinejensen00 (commented):
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2634)* status: all files reviewed, 8 unresolved discussions (waiting on tombogle).

___
*[`assets/localization/en.json` line 50 at r2](https://reviewable.io/reviews/paranext/paranext-core/2634#-OzJfqhO4bGQwubKq0jW:-OzMrVwI4veXm6IrR2YR:bi5ptki) ([raw file](https://github.com/paranext/paranext-core/blob/69e706c58d9737744635219a2058048bde5aa7d8/assets/localization/en.json#L50)):*

Consider: `Setting things up for the first time. This can take a few moments…` (I'm not sure how long a moment is, but it might be less than 30 seconds.)
</blockquote></details>

Fixed— changed "a moment" to "a few moments" in en.json (and kept fr/zh-hans/zh-hant consistent with the plural). "Moments" sets a more honest expectation for a wait that can run longer.
___
*[`assets/localization/es.json` line 171 at r2](https://reviewable.io/reviews/paranext/paranext-core/2634#-OzJYqVcCiRBZPKFLvFw:-OzMrZQl1WfmFuhRC9YH:b-ojnymt) ([raw file](https://github.com/paranext/paranext-core/blob/69e706c58d9737744635219a2058048bde5aa7d8/assets/localization/es.json#L171)):*

Configurando por primera vez. Esto puede tardar unos momentos…
</blockquote></details>

Applied your wording verbatim.
___
*[`assets/localization/es.json` line 172 at r2](https://reviewable.io/reviews/paranext/paranext-core/2634#-OzJevgwAlwPzVYjy5-A:-OzMr_jM8SGfg3ycvb5v:b-ojnymt) ([raw file](https://github.com/paranext/paranext-core/blob/69e706c58d9737744635219a2058048bde5aa7d8/assets/localization/es.json#L172)):*

La configuración inicial está tardando más de lo esperado. Vuelva a intentarlo; normalmente se resuelve enseguida.
</blockquote></details>

Applied your wording verbatim.
___
*[`src/renderer/components/first-run/step-loading.component.tsx` line 21 at r2](https://reviewable.io/reviews/paranext/paranext-core/2634#-OzJfIAw5Gi_CFHCqY6f:-OzMrc2I30e0yoJHycyc:bajb8dy) ([raw file](https://github.com/paranext/paranext-core/blob/69e706c58d9737744635219a2058048bde5aa7d8/src/renderer/components/first-run/step-loading.component.tsx#L21)):*

NIT: I'd probably remove the excessively detailed AI comment about why you used `cn`. We use `cn` a lot, and nobody is going to think it's weird that you did.
</blockquote></details>

Fair — trimmed it. The comment now just explains the always-mounted live region (the non-obvious a11y bit) and drops the `cn` justification.
___
*[`src/renderer/components/first-run/steps/internet-settings-step.component.tsx` line 24 at r2](https://reviewable.io/reviews/paranext/paranext-core/2634#-OzJhIRZ1n7NMl19TvtV:-OzMreFvEEefinLIxph_:b-p5qw4y) ([raw file](https://github.com/paranext/paranext-core/blob/69e706c58d9737744635219a2058048bde5aa7d8/src/renderer/components/first-run/steps/internet-settings-step.component.tsx#L24)):*

I think *maybe* I understand what this is trying to say, but it's a little vague and confusing. Those localization keys *are* indeed in a file called `en.json`.
</blockquote></details>

You're right, that was muddled — the point is narrowly about `%internetSettings_button_retry%`, which the paratext-registration extension localizes itself, so this core step doesn't need to add it to core's `assets/localization`. Reworded to say exactly that.
___
*[`src/renderer/components/first-run/steps/internet-settings-step.component.tsx` line 41 at r2](https://reviewable.io/reviews/paranext/paranext-core/2634#-OzJiIUf5xRkb6Ts7cK3:-OzMrg8y9VMg5YQg6cw5:b-du3er2) ([raw file](https://github.com/paranext/paranext-core/blob/69e706c58d9737744635219a2058048bde5aa7d8/src/renderer/components/first-run/steps/internet-settings-step.component.tsx#L41)):*

If the "attempts" here are intended to mean the *outer* retry attempts, then "so up to three outer retries fit within the budget" would be more accurate. In reality, it will likely only be 2 because of overhead. If we're really talking about the total attempts (inner and outer) then "several" feels pretty accurate, but maybe the wording should be tightened to make it clear what attempts we mean.
</blockquote></details>

Meant the outer re-issues. Tightened the comment to "roughly two or three outer re-issues fit within this budget," and noted that a call hitting the full ~30 s client timeout consumes the budget in a single attempt — so it's clear which attempts we're counting.
___
*[`src/renderer/hooks/use-delayed-flag.hook.ts` line 14 at r2](https://reviewable.io/reviews/paranext/paranext-core/2634#-OzJjrxkBS3ZveUcFpNX:-OzMriGsBbSAXaTI5gXN:bec8hk) ([raw file](https://github.com/paranext/paranext-core/blob/69e706c58d9737744635219a2058048bde5aa7d8/src/renderer/hooks/use-delayed-flag.hook.ts#L14)):*

if `delayMs` is a UX-tuned number that is likely to be a constant in the common case, maybe we should make it optional and supply a default rather than making every caller reason as to what the number should eb or look at other callers to find the UX-approved delay.
</blockquote></details>

Fixed— `delayMs` is now optional and defaults to an exported `DEFAULT_DELAY_MS` (2000) documented on the hook. The internet-settings caller just calls `useDelayedFlag(!settings && !loadFailed)` now, and its local `CONNECTING_MESSAGE_DELAY_MS` constant is gone.
___
*[`src/renderer/hooks/use-delayed-flag.hook.test.ts` line 15 at r2](https://reviewable.io/reviews/paranext/paranext-core/2634#-OzJkMGs3RHExVGsKsmV:-OzMrm_K5EEoTPHzDaBe:b-i4zinl) ([raw file](https://github.com/paranext/paranext-core/blob/69e706c58d9737744635219a2058048bde5aa7d8/src/renderer/hooks/use-delayed-flag.hook.test.ts#L15)):*

All of these tests pass 2000. The implementation could ignore the parameter and hard-code 2000 and the tests would still pass.
</blockquote></details>

Good catch — the tests couldn't have caught a hard-coded delay. Reworked them to use distinct values (500 and 5000) at their exact boundaries, so ignoring `delayMs` now fails, plus a separate test that omitting the arg falls back to `DEFAULT_DELAY_MS`.

tombogle (approved):
> @tombogle reviewed 14 files and all commit messages, and resolved 8 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2634)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2635, opened 2026-08-04
`isCharacterMarker` helper and character-marker item generator

jolierabideau:
# Code Review Summary

**Branch**: `is-character-marker-helper-and-item-generator`

**Base**: `origin/main` (merge-base `4a51de3`)

**Date**: 2026-08-03

**Files changed**: 15 (5 of them tracked `lib/platform-bible-utils/dist/` build output)

## Overview

This branch implements `PT-XXX-C` from the character-level-markers implementation plan: the
`isCharacterMarker` helper in `platform-bible-utils` and the `generateCharacterMarkerMenuListItems`
item generator for the character-marker menu, plus the menu's remove-row label in `en` and `es`. Both
functions are pure and DOM-free, and neither is mounted — the control that consumes them is
`PT-XXX-D`. The one user-visible surface (the remove row) cannot render today, because it is emitted
only when `EditorRef` supplies a `removeCharacterMarker` operation and no such operation exists yet.

The plan's §1.6 "Simple mode only / inert in Power" requirement is the dominant constraint on this
ticket, and it is satisfied in a way a reviewer can verify from the diff alone: `isBlockMarker` is
byte-for-byte unmodified, `generateInlineMarkerMenuListItems` is untouched (the `\` keydown menu is
Power's entire character-marker story), no existing call site was retargeted through the new helper or
generator, and the only wiring into the running app is the localized-key spread in the editor web view
— which §1.6 names as the single deliberate, accepted Power-mode side effect. Net effect in Power mode:
zero new entry points, zero changed entry points.

Analysis found **no critical issues**. Four important findings were raised and all four were resolved
during the review — three by code changes, one by a deliberate scope decision. Of 15 minor findings,
7 were fixed and 8 recorded.

## API Changes

- `lib/platform-bible-utils`: Added `isCharacterMarker(marker: string): boolean` export
  (`src/markers/usfm-markers.ts:3082`, re-exported from `src/index.ts:55`, declared in
  `dist/index.d.ts`). Export name set went 134 → 135.
- `lib/platform-bible-utils`: `MarkerCategoryType as CategoryType` changed from a **type-only** export
  to a **value** export (`src/index.ts:281`), so consumers can compare `marker.category` at runtime.
  The public name is unchanged, so this is additive — no consumer breaks. (Changed during this review.)
- `lib/platform-bible-react`: No API change. One Storybook story added (`CharacterMarkerRemoveRow`).
- `@papi/` modules (`lib/papi-dts/papi.d.ts`): No change, and none owed — `papi.d.ts` holds no
  reference to `usfmMarkers`, `isBlockMarker`, or `MarkerType`, only named type imports.
- Extensions' declared type surfaces (`extensions/src/*/src/types/*.d.ts`): No change.
- Extension-internal module surface (not part of the declared type surface): new module
  `extensions/src/platform-scripture-editor/src/character-marker-menu.utils.ts` exports
  `generateCharacterMarkerMenuListItems()` and `CHARACTER_MARKER_MENU_STRING_KEYS`.
  `platform-scripture-editor.utils.ts` gains only a file-header comment.

None.

- [x] **The extension-host / React module boundary was documented only in a commit message.** The
  generator needs a `lucide-react` icon; `main.ts` imports `platform-scripture-editor.utils.ts`, so a
  runtime UI value there loads in the extension host, whose module shim rejects any `require` other
  than `papi` — the whole extension fails to activate and no editor opens. Verified that nothing
  enforces it: the extension's `no-restricted-imports` blocks only `shared/*`, `renderer/*`,
  `extension-host/*`, `node/*`, `client/*`, `main/*`; CI's blocking smoke suite documents that opening
  an editor is out of its scope (`e2e-tests/tests/smoke/ui-interaction.spec.ts:151-154`).
  _(fixed during review: added a forward-facing file-header comment to `character-marker-menu.utils.ts`
  stating the constraint and why the module exists, plus a counterpart warning in
  `platform-scripture-editor.utils.ts` — the file where the violation would actually be written —
  noting that type-only imports are erased and fine while value imports are fatal. Author declined the
  ESLint rule and smoke assertion for this ticket.)_
- [ ] ~~**Third near-verbatim copy of the parent-scoped `usfmMarkers.children` → `MarkerMenuItem[]`
  block**, alongside `generateInlineMarkerMenuListItems`
  (`platform-scripture-editor.utils.ts:420-452`) and the footnote editor's differently-scoped copy
  (`footnote-editor.utils.ts:22-43`). All three share the two early returns, the
  `Object.entries(children).forEach` accumulation, the identical title-localization fallback, and the
  identical `(a.marker ?? a.title).localeCompare(...)` sort. The varying part is injectable (a marker
  predicate plus an action factory).~~ _(Author elected to record it as follow-up rather than extract
  now. Note that a full three-way collapse is not available to this plan regardless: §1.6 and PRD
  no-go #1 both place the footnote editor's copy off-limits because it renders in Power mode, so only
  the two `platform-scripture-editor` copies are in reach. No `TODO` was added in code because the
  plan's ticket IDs are still placeholders and an ID-less `TODO` gives a future reader nothing to act
  on.)_
- [x] **The en/es parity guard did not scale past its one hardcoded key.**
  `localized-strings.test.ts` re-declared the key as a literal instead of importing the
  `CHARACTER_MARKER_MENU_STRING_KEYS` tuple it already exports, so a second key added to the tuple
  would have shipped with no `es` value and a green test — and missing `es` translations are a blocking
  issue per the Localization guide. _(fixed during review: the guard is now driven by
  `CHARACTER_MARKER_MENU_STRING_KEYS.forEach`, matching the `SYNC_BLOCKED_KEYS.forEach` shape at
  `legacy-comment-manager/src/localized-strings.test.ts:72-80`. Also added the Spanish sentence-case
  assertion the same precedent carries. 4 assertions per key, passing.)_
- [ ] **Picking the already-applied marker is silently inert** — the action closes the menu and does
  nothing, the row looks enabled, and nothing indicates which marker is currently applied
  (`character-marker-menu.utils.ts:98-114`). Against the design guide that is two misses:
  `Guidelines/Applying Changes` requires feedback whenever an effect is not visible, and
  `Guidelines/Component Choices` expects a selection surface to show the current selection.
  **Deliberately deferred to `PT-XXX-D`, after investigating the cheap fix and finding it unavailable:**
  neither existing `MarkerMenu` flag works. `isDisallowed` does not merely disable — it *hides* the row
  while the search box is empty whenever any allowed item exists
  (`marker-menu.component.tsx:148-155`), so the applied marker would vanish from the list entirely.
  `isDeprecated` stays visible and disables correctly, but its trailing label is
  `%markerMenu_deprecated_label%`, so the row would be labeled "Deprecated" — a user-visible falsehood
  in both languages. Rendering it correctly therefore requires an **additive `MarkerMenuItem` field**,
  which is precisely the `MarkerMenu` change §1.6 assigns to `PT-XXX-D`, and a boolean added now would
  likely be superseded by that ticket's partial/all/none tri-state rather than extended. The inertness
  is documented in a code comment and pinned by a test.

### Minor — Consider

- [x] **`Object.entries(markerDetails.children).forEach(([, markers]) => …)`** discarded the key with a
  destructuring hole (copied from the analogue). _(fixed during review: now `Object.values(...)`.)_
- [x] **`characterMarkerActions` was named for actions** but one of its three members
  (`currentCharacterMarker`) is state, not an action — the TSDoc conceded it. _(fixed during review:
  renamed to `characterMarkerOptions`. Deliberately **not** the analyzer's suggested
  `characterMarkerContext`, because "context" collides with `contextMarker`, which appears throughout
  this domain and means something different.)_
- [x] **`makeMockEditorRef` was copied verbatim** from `platform-scripture-editor.utils.test.ts:22-32`
  including a `formatPara` spy no test in this file uses, with a doc comment saying "the generators"
  (plural) though the file tests one. _(fixed during review: spy dropped, comment singularized.)_
- [x] **Missing the Spanish sentence-case assertion** the `legacy-comment-manager` precedent carries
  (`localized-strings.test.ts:61-65`). _(fixed during review: added, driven off the same key tuple.)_
- [x] **`MarkerCategoryType` was exported type-only** (as `CategoryType`) while the very next line made
  `MarkerType` a *value* export with a comment explaining why. `isCharacterMarker` is the repo's first
  runtime comparison on `Marker.category`, and its documented contract is stated in terms of the
  `DivisionMarks` category, so a consumer could type `usfmMarkers[m].category` but not compare it
  against `CategoryType.DivisionMarks` at runtime. _(fixed during review: moved to the value export
  beside `MarkerType` under its existing public name, so no consumer breaks. Verified no in-repo
  consumer of `CategoryType` existed. `dist/` rebuilt — `index.js`/`index.cjs` gained the runtime
  export; `index.d.ts` needed no change because its dts bundler already emitted the enum declaration
  and re-exported it regardless of type-only status.)_
- [x] **No Storybook deliverable demonstrated the new remove row**, so its label/icon pairing could not
  be seen anywhere — the app cannot render it and no story showed it, while `Guidelines/UI Review`
  makes a story the deliverable for UI work. _(fixed during review: added a
  `CharacterMarkerRemoveRow` story to `marker-menu.stories.tsx` showing the remove row leading a
  sorted marker list, with a docs note that the icon must be explicit because `MarkerMenu` falls back
  to a `Ban` glyph. The story's `play` asserts the remove row renders first and is enabled, and it
  **is** exercised by `vitest`: `.storybook/preview.ts` sets project-level `tags: ['autodocs', 'test']`,
  Storybook merges tags additively (removal requires `'!test'`), and `vitest.config.ts` runs
  `storybookTest` with `tags: { include: ['test'] }` — so this `meta`'s `tags: ['autodocs']` adds to
  the project tags rather than replacing them. Confirmed by running
  `npx vitest run --project=storybook src/stories/advanced/marker-menu.stories.tsx` → 3 tests passed,
  covering both this `play` and `DisallowedMarker`'s.)_
- [x] **The design doc §2/§3 and the plan's `PT-XXX-C` section still named the old location** (append
  to `platform-scripture-editor.utils.ts`), stale after the module move — `PT-XXX-D`'s implementer
  would have looked in the wrong file. _(fixed during review: both updated to name
  `character-marker-menu.utils.ts` / `.test.ts`, and both now state the extension-host import
  constraint as the reason rather than leaving it as an unexplained deviation. Files remain untracked
  per the author's standing preference.)_
- [ ] ~~**`isCharacterMarker` went straight to the stable surface** rather than `src/experimental.ts`.~~
  _(Defensible and left as-is: it sits beside its stable analogue `isBlockMarker` and the design
  commits to keeping the predicate data-driven. Flagged so the choice is conscious rather than
  defaulted — if OQ-10 later narrows the predicate instead of filtering at the menu layer, that is a
  breaking behavior change on a stable export carrying a deprecation obligation.)_
- [ ] ~~**Placement: a `character-marker-menu/` folder** would match the extension's per-feature
  colocation convention (`paragraph-marker-tooltip/`, `two-step-delete-tooltip/`) and avoid a rename
  when `PT-XXX-D` adds the consuming component.~~ _(Flat `src/` placement is also well-precedented
  (`scripture-text-grid-*.utils.ts`), so this is a preference rather than a violation. Worth deciding
  at `PT-XXX-D`, when the component actually arrives.)_
- [ ] ~~**`LocalizedStringsFile` / `readLocalizedStrings` duplicated** from
  `legacy-comment-manager/src/localized-strings.test.ts:16-28`.~~ _(Following the precedent is the
  right call — the Code-Style-Guide names `extensions/src/common` as the home for shared extension code
  and it does not exist. Recorded so a **third** copy triggers extracting it.)_
- [ ] ~~**The structure-protection assertion (`items.every((item) => !item.isDisallowed)`) can only
  pass at this layer**, since the generator takes no `isStructureProtected` and has no path that could
  set the field.~~ _(Deliberate, and the exemption is structural rather than asserted-by-flag. Useful
  as a pin against a future addition, but it does not establish the behavioral claim; that belongs at
  the integration tier §4.3 already schedules for `PT-XXX-D`. Recorded so the acceptance-criterion
  checkbox is not read as stronger coverage than it is — the sibling test in `usfm-markers.test.ts`
  makes the equivalent caveat explicitly in a comment.)_
- [ ] ~~**The `...CHARACTER_MARKER_MENU_STRING_KEYS` spread in the web view is the one uncovered new
  line.**~~ _(No test harness exists for that 2100-line file; building one for a single spread is not
  proportionate. Informational.)_
- [ ] **Commit `65f05890b21` says "Add isCharacterMarker to PBU"** — an undefined project-coined
  initialism; the Code-Style-Guide lists `USFM`/`RTL`/`PAPI` as the established set and asks that
  others be spelled out. Later commits correctly write `platform-bible-utils`. **Worth fixing in the
  squash message, since that is what lands on `main`.**
- [ ] **The remove row's noun may not match its neighbours.** The label says "character marker" while
  the marker descriptions rendered in the *same* menu say "character **style**"
  (`%markerMenu_marker_bd_description%` = "A character style, use bold text"), and Paratext 9 calls
  these character styles. Countervailing: the plan's own Terminology table says to prefer "character
  marker" over "character style", and the extension already uses "markers" in user-facing text. **Folded
  into the OQ-11 ack going to UX** rather than decided here — the string is unrenderable today, and
  changing it later requires a new key plus a `metadata.fallbackKey`.
- [ ] ~~**The remove row is an action label while sibling rows are glossary-style descriptions**, so
  reading top-to-bottom the first row is an action and the rest are dictionary entries.~~
  _(Pre-existing — `generateInlineMarkerMenuListItems` uses the same descriptions — and matching it is
  the right call for consistency. Flagged only so UX sees the mismatch the new top row makes visible.)_

#### Shared Regions Modified

None. Every changed non-`dist` file was scanned for `#region shared with` markers; none contain any.

#### Extension Config Changes

- [n/a] `extensions/src/platform-scripture-editor/contributions/localizedStrings.json` —
  extension-specific string content, not shared build/config infrastructure. No propagation needed.

No `package.json`, `tsconfig*.json`, `webpack.config.*`, `.eslintrc*`, `.erb/configs/`, or
`.github/workflows/` files changed anywhere in the diff, so there are no dependency, script, or CI
changes to propagate.

## Positive Observations

- **The §1.6 inertness claim is verifiable from the diff alone**, which is exactly what the acceptance
  criteria asked for: `isBlockMarker` untouched (the new function appended strictly below it),
  `generateInlineMarkerMenuListItems` unchanged, no call site retargeted, and the single web-view edit
  adding only localize keys behind an explanatory comment.
- **Every factual claim in the new TSDoc holds against the real data**, verified by executing the built
  package: 58 `MarkerType.Character` markers; exactly four in `DivisionMarks` (`ca`, `v`, `va`, `vp`)
  and no others; 54 accepted, every one with `hasEndMarker: true`; zero markers satisfy both
  `isCharacterMarker` and `isBlockMarker`; every accepted marker has a `description`; `f`, `fe`, `x`
  are all `MarkerType.Note`.
- **The four-marker correction is implemented as a category test, not a name list**, so `va`/`vp`/`ca`
  are correctly rejected — the trap being that `hasEndMarker` excludes exactly one of the four and
  silently keeps three.
- **`dist/` is a genuine rebuild, not hand-edited**: one export name added across the whole surface,
  chunk hashes unchanged, and the remaining `index.js` churn is minifier identifier renaming with no
  semantic drift.
- **Tests run against the real `usfmMarkers`** rather than a mock, per the plan's acceptance criterion,
  and the data-invariant tests state what they do *not* establish — unusually careful test commentary.
- **The generator's test suite pins the non-obvious contracts**, not the easy ones: the
  merge-independently case (marker applied, no change callback → still adds), the deliberately-inert
  already-applied branch, the childless-parent early return beating the remove row, and the raw
  `contextMarker` of `'p'` guard.
- **Localization is clean**: a new key rather than an edited meaning, correct alphabetical position in
  both `en` and `es`, correct `%webView_{feature}_{elementDescription}%` naming, sentence case, no
  ellipsis (right — the action is immediate), and "Quitar" consistent with the neighbouring
  `%webView_scriptureTextGrid_viewOptions_removeFromList%`, keeping "remove" distinct from "delete".
- **`RemoveFormatting` is a defensible, deliberate icon choice** — `Guidelines/Design Principles`
  reserves `Trash2` for discarding content, and this action removes only the marker while leaving its
  text, so `Trash2` would over-signal loss.
- **All comments on the branch are forward-facing**, per the repo's comment rule — constraints,
  invariants, non-obvious rationale, and "this pins the current answer so narrowing it later fails
  here and gets re-decided." No PR/ticket references for work done in this PR, no review-round notes,
  no stage tags, no dated development notes. Audited explicitly at the author's request.
- **The module boundary fix is a real architectural boundary, not a workaround** — a module imported by
  both the extension host and a web view structurally cannot hold runtime UI values.

## Interview Notes

**Stated purpose**: resolve `PT-XXX-C` — the `isCharacterMarker` helper and the character-marker item
generator.

**The author's own framing of the hard part.** Asked which part a reviewer should actually think about,
the author separated the two non-obvious decisions by kind, and this distinction is the most useful
thing in this write-up:

- **The module split is a diagnosis** — "unpleasant to find, but once you see the runtime error the
  conclusion is forced." The author traced the four-step chain and identified precisely what makes it
  non-obvious: the first three steps are invisible in the source. An icon is a runtime value TypeScript
  cannot elide (which is why the file looked fine before — its other UI-adjacent imports were
  type-only); extension builds externalize dependencies so the bundle emits a bare `require('react')`;
  the host shim rejects it; the entire extension throws on activation. Web views escape because
  `main.ts` pulls them in as `?inline` strings, forming a separate graph where React is legal. The
  author's stated concern is not the fix but its **invisibility**: nothing enforces it — not the build,
  not lint, not the type system — and a header comment in two files is the whole safety net.
- **The category exclusion is a bet on data shape**, and the author explicitly nominated it as the
  reviewable claim. The setup: the obvious `type === MarkerType.Character` is wrong because `v`, `va`,
  `vp`, `ca` are typed `Character` despite being structure markers. The genuinely non-obvious part is
  why `category` is the *only* discriminator available — **the intuitive fallback `hasEndMarker` fails
  in a way that looks like it works.** `v` has `hasEndMarker: false` so it gets excluded, and a
  reasonable person would stop there believing the problem solved; but `va`, `vp`, and `ca` all have
  `hasEndMarker: true`, identical to every genuine character marker. So `hasEndMarker` catches exactly
  one of four and silently keeps three.

**The trade the author put in front of the reviewer**, in their own terms — the two options fail in
opposite directions:

- **A name list fails closed.** Someone adds another numbering marker → it leaks into the character
  menu as a selectable style. Visible, wrong, reported.
- **The category rule fails open.** Someone files a genuine body-text character marker under
  `DivisionMarks` → it silently disappears from the menu. No error, nothing to notice.

The author verified the invariant over all 186 markers and was candid that on today's data the category
rule is *exactly equivalent* to the `v`/`va`/`vp`/`ca` name list — "It buys nothing today. Its entire
value is in future behavior, and its entire risk is in future data." Their position: `DivisionMarks` is
semantically about document structure, so a styling marker landing there would already be a
data-modeling error — the right bet, but a bet.

**The author found a real gap in their own tests, unprompted**, and it was the sharpest observation of
the interview: the two invariant tests guard only the direction that fails loudly. `only accepts
markers that have an end marker` checks *accepted* markers, so a genuine character marker filed under
`DivisionMarks` is rejected, never enters either filter, both tests still pass, and the marker quietly
vanishes. **The one direction the design's fragility actually lies in was the untested one.** Fixed
during the review: added `excludes exactly the chapter and verse numbering markers by category`,
asserting `DivisionMarks ∩ Character === ['ca','v','va','vp']`. Falsified before keeping it — an
injected styling marker under `DivisionMarks` produced
`expected [ 'ca', 'v', 'va', 'vp', 'zz' ] to deeply equal [ 'ca', 'v', 'va', 'vp' ]`.

**Author-directed process constraints**, both applied:

- All comments added or changed on this branch must be **forward-facing** (the repo's comment rule).
  The author authorized the boundary comments only after confirming they pass the strip-the-PR-context
  test, and then asked for a full audit of every comment on the branch. Audit result: clean, nothing
  cut.
- Plan and PRD docs stay **untracked**; staging was done by explicit path throughout, never `git add -A`.

**No areas of deferral or uncertainty.** The author did not defer any part of these changes to an AI
tool, and explained every flagged decision without prompting. On the one finding they declined to fix
(the inert applied-marker row), they made an explicit scope judgment — keep the additive `MarkerMenu`
change in `PT-XXX-D` where §1.6 assigns it — rather than expressing uncertainty. The author also
correctly challenged one of my findings as imprecise: I framed the "character marker vs character
style" wording issue as a label naming a concept, and they pushed back asking where the text actually
renders. They were right — it is an action label on a row that cannot render today, and the prototype
uses neither noun (its equivalent row is `(none)`, and "Character Style" is only a button tooltip), so
the finding is a sub-question of the already-open OQ-11 rather than a standalone issue.

## In-Review Quality Check

All checks clean after the in-review changes. **3157 tests passing, 0 failures.**

| Check | Result |
| :---- | :---- |
| `npm run typecheck` | Clean, all workspaces |
| `npm run lint` (repo-wide) | **0 errors.** 4 warnings, all pre-existing in files not on this branch's diff (`model-text-panel.component.tsx`, `use-effective-resource-reference-list.hook.ts`) |
| Prettier (changed files only) | Reformatted 1 file — a doc-comment rewrap in `marker-menu.stories.tsx` |
| Tests — root | 41 files / 593 passed |
| Tests — `extensions` | 77 files / 1148 passed |
| Tests — `platform-bible-utils` | 28 files / 427 passed, 1 skipped |
| Tests — `platform-bible-react` | 127 files / 989 passed |
| `npm run build:platform-bible-utils` | Ran after the `src/index.ts` export change; `dist/index.js` and `index.cjs` picked up the `CategoryType` runtime export |
| `dotnet test c-sharp-tests/` | Skipped — no C# changed on this branch |

No unfixable failures. One coverage caveat, stated rather than hidden: the web-view
`CHARACTER_MARKER_MENU_STRING_KEYS` spread remains the one uncovered new line.

_(Correction: an earlier revision of this write-up also listed the new story's `play` function as
uncovered, on the assumption that this `meta`'s `tags: ['autodocs']` dropped the `'test'` tag. That
was wrong — Storybook merges tags additively across project/meta/story, and `.storybook/preview.ts`
already sets `tags: ['autodocs', 'test']` project-wide, so both this story's `play` and
`DisallowedMarker`'s do run under `vitest`. Verified:
`npx vitest run --project=storybook src/stories/advanced/marker-menu.stories.tsx` → 3 tests passed.)_

## Suggested Review Focus

Prioritized agenda for the author-reviewer meeting:

- [ ] **The `DivisionMarks` category bet — the author's own nominated review item, and the right place
  to start.** Is `DivisionMarks` a closed structural category that will never gain a styling marker?
  The rule is exactly equivalent to a `v`/`va`/`vp`/`ca` name list on today's data, so this is purely a
  choice about which failure direction you prefer: a name list fails closed and visibly, the category
  rule fails open and silently. The new invariant test converts the silent direction into a build
  failure — confirm the reviewer agrees that is sufficient mitigation for taking the open-failure side.
- [ ] **The unenforced extension-host module boundary.** Two header comments are the entire safety net
  against a change that causes total extension activation failure at runtime, with no build, lint, or
  type-level signal. The author declined the ESLint `no-restricted-imports` override and the smoke-test
  assertion for this ticket — worth a second opinion on whether a comment is enough for a failure mode
  this severe, especially as `PT-XXX-D` through `PT-XXX-G` will each add UI code to this extension.
- [ ] **The inert already-applied marker row, and what `PT-XXX-D` owes it.** Deferred deliberately, but
  the reviewer should confirm the shape: neither existing `MarkerMenu` flag can express "applied"
  (`isDisallowed` hides the row, `isDeprecated` mislabels it), so `PT-XXX-D` must add the field — and
  its partial/all/none tri-state should be designed to cover this case rather than sitting beside it.
- [ ] **OQ-11 / §11 → A8, now with a third sub-question for UX.** Does the remove row read *(none)*
  (the prototype's wording) or "Remove character marker"? Is `RemoveFormatting` the right icon? And —
  added by this review — should the row's noun be "character marker" or "character **style**", given
  the description rows directly beneath it say "A character style, use bold text"? All three are
  acks/cosmetic and none blocks merge, but all three become expensive after the strings ship.
- [ ] **Whether `isCharacterMarker` belongs on the stable surface** rather than `src/experimental.ts`,
  given that OQ-10 (hiding non-styling character markers such as `qs`, `qac`, `xt`, `rq`) is still
  open. If that ever lands as a narrowing of the predicate rather than a menu-layer filter, it is a
  breaking change on a stable export.
- [ ] **Follow-up, agreed during review, not done here: the three-way duplication** of the
  parent-scoped children → `MarkerMenuItem[]` block. Only the two `platform-scripture-editor` copies
  are in reach — §1.6 and PRD no-go #1 both put the footnote editor's copy off-limits. Decide whether
  to extract before `PT-XXX-D` adds a fourth consumer. Needs a real ticket ID.
- [ ] **Squash-message fix:** commit `65f05890b21` uses the undefined initialism "PBU".

AI-assisted — reviewed with `/review-paratext` (Claude Opus 5)

---

This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2635)


katherinejensen00 (changes requested):
Thank you for adding better character level support, Jolie! Also, Claude was impressed by your testing. It said, "the code is sound and unusually well-tested". Great work!

> @katherinejensen00 reviewed 15 files and all commit messages, and made 9 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2635#-:-OzCX_FBC36w9NIICR49:b-oecuyp)* status: all files reviewed, 8 unresolved discussions (waiting on jolierabideau).

___
*[`extensions/src/platform-scripture-editor/src/character-marker-menu.utils.ts` line 102 at r1](https://reviewable.io/reviews/paranext/paranext-core/2635#-OzCWZFk3CRNAoMYLvO0:-OzCWZFk3CRNAoMYLvO1:b-cnc9va) ([raw file](https://github.com/paranext/paranext-core/blob/1633e06bc67c7b6951db220995180f58ebc42e24/extensions/src/platform-scripture-editor/src/character-marker-menu.utils.ts#L102)):*
> ```XML
>       : undefined;
> 
>   const markerMenuItems: MarkerMenuItem[] = [];
> ```

**\[Low\] Simplify the item build and the sort.**

The `const markerMenuItems = []` + `Object.values(...).forEach(push(...))` is a `flatMap`: `Object.values(markerDetails.children).flatMap(markers => markers.filter(isCharacterMarker).map((marker): MarkerMenuItem => …))`. And at line 130 the comparator's `?? a.title` is dead — every generated item has a `marker` and the remove row is prepended after sorting — so `a.marker.localeCompare(b.marker)` is both simpler and truer to the invariant.
___
*[`extensions/src/platform-scripture-editor/src/character-marker-menu.utils.ts` line 123 at r1](https://reviewable.io/reviews/paranext/paranext-core/2635#-OzCVRmZC4AT4TX78zkI:-OzCVRmZC4AT4TX78zkJ:b6w1yg0) ([raw file](https://github.com/paranext/paranext-core/blob/1633e06bc67c7b6951db220995180f58ebc42e24/extensions/src/platform-scripture-editor/src/character-marker-menu.utils.ts#L123)):*
> ```XML
>               if (currentCharacterMarker && changeCharacterMarker)
>                 changeCharacterMarker(currentCharacterMarker, marker);
>               else editorRef.current?.insertMarker(marker);
> ```

**\[Medium\] Please verify that "add instead of change" can't nest character markers.**

When a marker is applied but `changeCharacterMarker` isn't wired, picking a different marker falls through to `insertMarker`, inserting a second character marker over an existing span. It's unreachable today (nothing passes `currentCharacterMarker`), and I couldn't inspect `@eten-tech-foundation/platform-editor` to see whether `insertMarker` nests, replaces, or rejects here. Worth confirming against the editor before PT-XXX-D passes a real `currentCharacterMarker`, since nested/overlapping character markers are invalid in some USFM consumers.
___
*[`extensions/src/platform-scripture-editor/src/localized-strings.test.ts` line 15 at r1](https://reviewable.io/reviews/paranext/paranext-core/2635#-OzCWmro8JW9NR9zc9_H:-OzCWmro8JW9NR9zc9_I:b8gm7kg) ([raw file](https://github.com/paranext/paranext-core/blob/1633e06bc67c7b6951db220995180f58ebc42e24/extensions/src/platform-scripture-editor/src/localized-strings.test.ts#L15)):*
> ```XML
>   // JSON.parse returns `any`; asserting the known shape of localized strings contribution files
>   // eslint-disable-next-line no-type-assertion/no-type-assertion
>   const stringsFile = JSON.parse(readFileSync(stringsFilePath, 'utf-8')) as LocalizedStringsFile;
> ```

**\[Low\] You can drop this `eslint-disable` entirely.**

`JSON.parse` returns `any`, which assigns to a typed target without a cast, so `const stringsFile: LocalizedStringsFile = JSON.parse(readFileSync(...), 'utf-8')` satisfies the type with no `as` and no suppression — which is what the eslint-disable-discipline rule prefers.
___
*[`extensions/src/platform-scripture-editor/src/localized-strings.test.ts` line 43 at r1](https://reviewable.io/reviews/paranext/paranext-core/2635#-OzCVg5l4uD4hrrwI8p0:-OzCVg5l4uD4hrrwI8p1:b3cxvab) ([raw file](https://github.com/paranext/paranext-core/blob/1633e06bc67c7b6951db220995180f58ebc42e24/extensions/src/platform-scripture-editor/src/localized-strings.test.ts#L43)):*
> ```XML
>       const es = localizedStrings.es[key];
>       expect(es.charAt(0)).toMatch(/[A-ZÁÉÍÓÚÜÑ]/);
>       expect(es.slice(1)).toBe(es.slice(1).toLowerCase());
> ```

**\[Low\] This sentence-case check will reject correct future translations and can throw instead of failing cleanly.** 

`es.slice(1) === es.slice(1).toLowerCase()` forbids _any_ non-initial capital, so a future key whose Spanish legitimately contains a proper noun or an acronym (e.g. `USFM`) fails a correct translation. Related: `charAt(0)` won't match a Spanish opening `¿`/`¡`; a missing `es` value makes `es.charAt(0)` throw a `TypeError` rather than report the missing string; and it uses `toLowerCase` not `toLocaleLowerCase`. It's copied from `legacy-comment-manager`, so this is really a shared-pattern gap — worth loosening (or extracting) before more keys ride on it.
___
*[`extensions/src/platform-scripture-editor/src/character-marker-menu.utils.ts` line 137 at r1](https://reviewable.io/reviews/paranext/paranext-core/2635#-OzCVH9r0CbOWU-R_Z7i:-OzCVH9r0CbOWU-R_Z7j:bzbgmsr) ([raw file](https://github.com/paranext/paranext-core/blob/1633e06bc67c7b6951db220995180f58ebc42e24/extensions/src/platform-scripture-editor/src/character-marker-menu.utils.ts#L137)):*
> ```XML
>   // menu's search box is empty: the row has no marker code to match on, so a search can only match
>   // its title, and `MarkerMenu` renders title matches after every code match.
>   if (!currentCharacterMarker || !removeCharacterMarker) return sortedMarkerMenuItems;
> ```

**\[Medium\] The menu can return a "remove only" list, which contradicts the documented `@returns`.

The `@returns` says the result is `[]` when the parent "contributes no character markers." But for the 14 parents whose children are all non-character (`mt`, `qs`, `cp`, `h`, `h1`–`h3`, `ide`, `b`, `mte1`, `ms3`, `mt1`–`mt3`), the childless early-return is bypassed, the filter yields `[]`, and if a marker is applied with `removeCharacterMarker` supplied you get `[removeRow]` — a menu that offers only "Remove" and nothing to add. It's latent (needs the remove op), but before PT-XXX-D either return `[]` when there are no character children, or update the doc to describe the remove-only case.
___
*[`extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts` line 34 at r1](https://reviewable.io/reviews/paranext/paranext-core/2635#-OzCWQ1SAL7M57KKQ3Um:-OzCWQ1SAL7M57KKQ3Un:bb0g34x) ([raw file](https://github.com/paranext/paranext-core/blob/1633e06bc67c7b6951db220995180f58ebc42e24/extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts#L34)):*
> ```XML
> import { ScriptureRange } from 'platform-scripture-editor';
> import type { SharedProjectsInfo } from 'platform-scripture';
> import { MutableRefObject } from 'react';
> ```

**\[Low\] One-line hardening for the import boundary: make these `import type`.** 

`MutableRefObject`, `EditorRef`, and `MarkerMenuItem` are bare value imports used only in type positions — safe today only because the compiler elides them. Switching to `import type { … }` makes the erasure structural: a future value-use becomes a compile error _at the import line_ instead of a silent runtime `require('react')`. Given this is the file whose header warns about exactly this, it's a fitting place to practice it.
___
*[`extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts` line 1 at r1](https://reviewable.io/reviews/paranext/paranext-core/2635#-OzCV-oG57q3UvWgSzxz:-OzCV-oH-eUGKQ3orYv2:b-2o4iaw) ([raw file](https://github.com/paranext/paranext-core/blob/1633e06bc67c7b6951db220995180f58ebc42e24/extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts#L1)):*
> ```XML
> /*
> ```

**\[Medium\] The extension-host import boundary is documented but not enforced — and it's a hard-fail.**

The header is accurate: a _value_ import of a UI module into this host-reachable file crashes activation. I confirmed the exact mechanism — `lucide-react`/`platform-bible-react` aren't webpack-externalized, so they're bundled, but they call `forwardRef`/`createContext` at module-eval, which transitively emits `require('react')`; `react` **is** externalized and **isn't** on the host shim's allowlist (`extension.service.ts:1328-1354`), so it throws "Requiring other than papi is not allowed" and no editor opens. The whole safety net is prose, and PT-XXX-D..G will keep adding UI to this extension. Consider a `no-restricted-imports` rule (or an import-graph vitest test over `main.ts`'s transitive deps) so this fails in CI instead of at runtime. The cheapest first step is #6 below.
___
*[`lib/platform-bible-react/src/stories/advanced/marker-menu.stories.tsx` line 12 at r1](https://reviewable.io/reviews/paranext/paranext-core/2635#-OzCW0ZWEiLhDHm1qoKo:-OzCW0ZWEiLhDHm1qoKp:bbohcbf) ([raw file](https://github.com/paranext/paranext-core/blob/1633e06bc67c7b6951db220995180f58ebc42e24/lib/platform-bible-react/src/stories/advanced/marker-menu.stories.tsx#L12)):*
> ```XML
>   title: 'Advanced/MarkerMenu',
>   component: MarkerMenu,
>   tags: ['autodocs'],
> ```

**\[Low\] The PR description's coverage caveat is inverted — the story's `play` _does_ run.** 

The write-up says the `play` "is not exercised by vitest" because the meta lacks `'test'`. But `.storybook/preview.ts` sets a global `tags: ['autodocs','test']`, Storybook merges tags additively (you remove one with `'!test'`), and `vitest.config.ts` includes `'test'` — so this story's `play` **and** the `DisallowedMarker` precedent's both execute. Good news for coverage; just worth correcting the note so the assumption doesn't propagate.

jolierabideau (commented):
Thank you!

> @jolierabideau made 2 comments and resolved 7 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2635#-:-OzD1FAt5xGy4cjsRkBf:b-9755fn)* status: all files reviewed, 1 unresolved discussion (waiting on katherinejensen00).

___
*[`extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts` line 1 at r1](https://reviewable.io/reviews/paranext/paranext-core/2635#-OzCV-oG57q3UvWgSzxz:-OzD17mLAYVvQpEywBSD:b-896fix) ([raw file](https://github.com/paranext/paranext-core/blob/1633e06bc67c7b6951db220995180f58ebc42e24/extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts#L1)):*

**\[Medium\] The extension-host import boundary is documented but not enforced — and it's a hard-fail.**

The header is accurate: a _value_ import of a UI module into this host-reachable file crashes activation. I confirmed the exact mechanism — `lucide-react`/`platform-bible-react` aren't webpack-externalized, so they're bundled, but they call `forwardRef`/`createContext` at module-eval, which transitively emits `require('react')`; `react` **is** externalized and **isn't** on the host shim's allowlist (`extension.service.ts:1328-1354`), so it throws "Requiring other than papi is not allowed" and no editor opens. The whole safety net is prose, and PT-XXX-D..G will keep adding UI to this extension. Consider a `no-restricted-imports` rule (or an import-graph vitest test over `main.ts`'s transitive deps) so this fails in CI instead of at runtime. The cheapest first step is #6 below.
</blockquote></details>

Done.

katherinejensen00 (approved):
> @katherinejensen00 reviewed 5 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2635)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2425, opened 2026-06-17
Improve performance when switching to Simple

Sebastian-ubs:
# Code Review Summary

**Branch**: improve_simple-power_switching

**Base**: origin/main

**Date**: 2026-08-17

**Review model**: Claude Sonnet 5

**Files changed**: 21

## Overview

This branch speeds up the Power → Simple interface-mode switch by baking the target project directly into a cloned `simpleLayout` (via `buildSimpleLayoutForProject`) instead of restoring an empty layout and paying a placeholder-webview mount + reload cycle. A renderer-side cache (`last-opened-project-cache.ts`) lets the common case resolve the target project synchronously; a cold-start path falls back to an async recents-provider lookup, trying each recent project in turn instead of giving up after the first.

The branch has gone through five review rounds. The first four — Sebastian/tjcouch's file-count trim, `/review-paratext`'s Critical-bug fix, Matt Lyons' 48-finding deep review plus six follow-up asks, and a 15-finding adversarial re-review (N1–N15) from rolfheij-sil — are already reflected in earlier commits. **This update is a catch-up pass, not a new round**: the PR body had gone stale relative to the branch. Since the last summary was written (2026-08-13), five more commits landed, fixing five of that round's own open Minor findings (N7, N11, N12, N13, N14) and verifying four more as moot (#36, #19, #17, #21) — none of which had been reflected back into the PR body's checkboxes. A sixth commit fixed a genuine regression (found live, in a follow-up review) that was never added to the findings list at all. This round re-confirms all of that against the actual commit history, runs a fresh four-pass analysis on the current diff, and fixes the one new Important finding it surfaced.

## API Changes

- `extensions/src/platform-scripture-editor/src/types/platform-scripture-editor.d.ts`: `platformScriptureEditor.finalizeProjectSwitch: (projectId: string) => Promise<void>` — a new, purely additive extension command, now marked `@experimental` (fixed this round via N7, see below). No existing exports modified or removed.
- No changes to `lib/platform-bible-react/`, `lib/platform-bible-utils/`, or `papi.d.ts` / any other `extensions/src/*/src/types/*.d.ts`.

### Critical — Must address before merge

None.

### Important — Should address before merge

[...]

PR #2636, opened 2026-08-04
Scope Simple-mode startup sync to the active project

tombogle:
## Summary

- Simple-mode startup sync currently calls `syncProjects` with `undefined` `projectIds`, which the C# side treats as "sync every shared project the account has ever touched anywhere on this machine" - not scoped to this app's own project set. On a machine that also has other Paratext 9 projects checked out, those get swept into the same startup sync, adding real time to the sequential per-project modification-status check the C# side runs (confirmed via a captured startup waterfall: this sync was ~60% of a ~55s cold start).
- Startup now resolves `platformScripture.recentlyOpenedProjects` (the same signal the Simple-mode default-active-project picker already uses) and scopes the sync to that list when it's non-empty (steady-state: the user already has an active project). Falls back to today's broad "sync everything" behavior when the list is empty (zero-state: no project opened in Simple mode yet), preserving the "discover a project already synced via Paratext 9" behavior for a brand-new user.
- `getMostRecentProjectId()` is refactored to share a new `recently-opened-project.util.ts` helper instead of duplicating the data-provider read, so this resolution and the Power->Simple layout fast-path's resolution can't drift apart.

This targets `improve_simple-power_switching` (#2425) rather than `main` since it builds on that branch's Simple-mode work and should be reviewed on its own merits, but will most likely land after #2425 merges.

Companion change: a small patch to `paratext-10-studio`'s private fork threads this same `projectIds` scoping into the C# `GetSharedProjects()`/`SyncProjectsCore` call (using an existing but previously-unused `SRProjectSelectionOptions.ForcedSelectedProjectIds` mechanism) - without it, the C# side still enumerates and status-checks every account-wide project regardless of what TS passes. Verified manually end-to-end in a paratext-10-studio release build: startup no longer status-checks unrelated projects, and is meaningfully faster.

## Test plan

- [x] `npm run typecheck:core` clean
- [x] Targeted `eslint` clean on all changed files
- [x] Full `npm test`: 39 files / 572 tests pass (both before and after rebasing onto the latest `improve_simple-power_switching`)
- [x] New tests: `recently-opened-project.util.test.ts` (3 cases), 2 new cases in `startup-tasks.test.ts` (steady-state scoped call, zero-state fallback preserved)
- [x] Manual end-to-end verification against a real `paratext-10-studio` release build (see companion patch PR)

AI-assisted — built with Claude Code.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2636)

### DIFF
--- a/src/main/startup-tasks.test.ts
+++ b/src/main/startup-tasks.test.ts
@@ -10,6 +10,7 @@ import { settingsService } from '@shared/services/settings.service';
 import * as commandService from '@shared/services/command.service';
 import * as networkService from '@shared/services/network.service';
 import { logger } from '@shared/services/logger.service';
+import { getRecentlyOpenedProjectIds } from '@shared/utils/recently-opened-project.util';
 import { performStartupTasks, STARTUP_SYNC_RETRY_BUDGET_MS } from './startup-tasks';
 
 vi.mock('@shared/services/settings.service', () => ({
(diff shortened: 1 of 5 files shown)


tombogle:
Closing in favor of a redesign. The scoping approach here (guess via `recentlyOpenedProjects` and forward up to 5 ids to `syncProjects`) has a real gap: it's an unconditional guess, not a targeted fix, and can still sync several projects when only one is actually needed. The better fix is to trigger a precisely-scoped sync at the exact point the active project is actually decided - in `openDefaultActiveProjectIfApplicable` (the Default Active Project Picker), mirroring the pattern `syncOnProjectSwitch` already uses for interactive project switches - rather than guessing earlier from the main process before that decision has been made.

That mechanism doesn't depend on `improve_simple-power_switching` (confirmed: `syncOnProjectSwitch` and the picker code are already on `main`), so the follow-up will target `main` directly instead of this branch.

The companion patch PR (paratext-10-studio#171, `GetSharedProjects(restrictToProjectIds)`) stays valid as-is - it's general-purpose scoping infrastructure that the new approach will also rely on.

PR #2638, opened 2026-08-04
feat: persist window layouts, bounds and displays (PT-4285)

rolfheij-sil:
> **Consolidated 2026-08-07:** this PR's commits now ride #2639 (see its body for the combined record). Closed unmerged; branch retained so per-step diffs stay linkable.

---

Implements PT-4285: every window's dock layout, bounds, and display persist across restarts, using the storage design decided on the ticket — **one structure holding all window layouts in order** (`window-layouts.json` in `userData`), owned and written only by the main process. Position in the list is the identity; window ids stay runtime-only.

### How it works

- At startup, main reads the structure and creates one window per entry in order, handing each window its entry. In **Simple mode only the main window is restored** (Simple is single-window per the epic's mode-switching spec); unrestored entries are preserved in the structure untouched, so a Simple session can never destroy a Power user's secondary windows.
- Renderers load their layout from their window's entry and **push** reconciled layout changes to main over internal (non-public) request handlers; main aggregates and debounces disk writes (temp-file + rename). At quit, each window's close handler flushes the full structure before shutdown tasks run; a **deliberate** window close instead rewrites the structure without that window, so it does not come back.
- Bounds are captured per window keeper-style (normal-state only, debounced) and restored through a visibility check against the currently connected displays — a window whose monitor went away comes back somewhere visible. **`electron-window-state` is removed**; the old `window-state.json` is still read once so an upgrade keeps the user's window placement.
- Legacy rule: with no structure file, exactly one window is restored from the pre-multi-window `localStorage` layout. This replaces the per-window prefixed-key persistence this branch previously carried (which cloned window 1's layout into every new window via its legacy-migration fallback) — per the ticket's "Note for PR #2621", the keyed model is now fully superseded.
- A window created mid-session starts **empty** (decided on the ticket); a deliberately-empty window open at quit is restored (empty). Phantom protection: layouts are reconciled on push (duplicate/orphaned tabs, empty panels stripped), and at load an entry whose tabs were *all* phantoms is dropped — it cannot resurrect a window — while an entry saved tab-less legitimately restores one.

### Verification

- **52 new unit tests** (persistence service 19, renderer host 12, bounds util 7, layout reconciliation 14), all behaviour-level; full suite green; typecheck green; `npm run build:types` produces an **empty `papi.d.ts` diff** — nothing new reaches the public API.
[...]

PR #2637, opened 2026-08-04
test: add multi-window lifecycle e2e tests (PT-4287)

rolfheij-sil:
> **Consolidated 2026-08-07:** this PR's commits now ride #2639 (see its body for the combined record). Closed unmerged; branch retained so per-step diffs stay linkable.

---

Adds behaviour-level e2e coverage for the multi-window work delivered in #2621, as the groundwork slice of PT-4287. Three tests in `e2e-tests/tests/isolated/multi-window/`, each launching its own Electron instance:

1. **Second-window lifecycle** — a second window starts clean (window-scoped web view ids, no duplicate registrations at warn/error severity), generic window-service calls follow focus to window 2 and back, and closing the secondary window neither quits the app nor runs shutdown tasks.
2. **Hosting takeover** — closing the window hosting the app-global theme and scroll-group services hands hosting to the surviving window, proven by live PAPI reads plus a scroll-group write round-trip against the survivor (not just cached reads).
3. **Quit with two windows** — a real `app.quit()` runs the shutdown tasks exactly once (not per-window, not zero) and the process exits cleanly with code 0.

All assertions are outside-observable (PAPI over WebSocket, page content, user-facing log lines), deliberately: the plumbing underneath is expected to be refactored per the PT-4275 architecture discussion, and these tests are the safety net for that refactor. No `src/` or fixture changes — only the new spec and a README index line.

### Falsifiability

Every test keys on the power-mode shutdown session-sync log marker (pre-configured `platform.interfaceMode: 'power'`), which fires exactly once per shutdown-task run — making "shutdown did not run" (test 1) and "ran exactly once" (test 3) real assertions rather than vacuous ones. Regressing id scoping, focus routing, either takeover subscription, the unreachable-object sweep, the write path, or the single-shutdown latch each fails a specific documented assertion.

### Notes for reviewers

- Test 3 watches the OS process `exit` event instead of Playwright's `ElectronApplication` `close` event: a graceful dev-mode quit leaves the .NET watcher child holding inherited stdio pipes, which that event waits on indefinitely. The leftover process group is reaped after the exit assertion.
- Headless compositors (observed on WSLg/Weston) honor a window's initial activation but ignore programmatic re-activation, so the focus helper escalates after 10 s to delivering the `focus` notification at the Electron boundary — everything the app does with the event still runs for real.
- Verified with two consecutive full green runs (3/3, no retries; ~55 s per run).

### Scope vs the ticket

[...]

PR #2621, opened 2026-07-28
PT-4277: Per-window PAPI service scoping and focus routing

rolfheij-sil:
# Multi-window support: per-window PAPI service scoping and focus routing

Platform.Bible assumes a single window in about a dozen load-bearing places — menus, notifications, focus tracking, keyboard shortcuts, app shutdown, and the layout system each exist exactly once. Today a second renderer cannot even start: four of the renderer's startup calls claim globally-unique PAPI names, and `network-object.service.ts` throws on duplicate registration.

This PR makes those single-window assumptions per-window, so a second window starts cleanly and behaves correctly. It revives the hackathon spike from #2051, which had drifted 317 commits behind main and whose test plan was never run. The first commit is that original port unchanged; the second squashes the follow-up work already on that PR; everything after is the revival — rebasing onto current main, then making multi-window actually work.

## What this changes

**Renderer-hosted services register under window-scoped names.** The web view, window, dialog, and notification services and the renderer-hosted commands register as `WebViewService-1`, `platform.openSettings-1`, and so on. The main process republishes each generic name as a proxy that routes to the focused window, so existing callers and `papi.d.ts` keep working unchanged.

**App-global services keep exactly one host.** Scroll groups and the theme are genuinely app-wide, so one renderer hosts each and the others attach — and a surviving window takes the name over when the hosting window closes.

**Network events survive multiple sources.** The scroll-group and web-view events moved to the platform's existing `MultiSourceNetworkEvents` mechanism. As single-source emitters, only the first window to start could emit, so every later window's emits were dropped.

**Web view ids are scoped per window.** Ids baked into the shared layout constants meant two windows materialised web views with identical ids, colliding on `webViewMessage:{id}`.

Running it with two and three windows — which the original spike never did — is what found the two defects no amount of reading would have: a multi-window quit skipped the shutdown Send/Receive entirely, and seven event emitters plus the per-web-view message handlers collided on the second window.

## Tickets

| Ticket | Relationship |
| --- | --- |
| [PT-4277](https://paratextstudio.atlassian.net/browse/PT-4277) | **The delivery ticket.** This PR satisfies its Definition of Done except the two gaps below |
| [PT-4286](https://paratextstudio.atlassian.net/browse/PT-4286) | The multi-window quit / shutdown Send/Receive data-loss bug is routed here by PT-4277, and is fixed in this PR |
| [PT-4270](https://paratextstudio.atlassian.net/browse/PT-4270) | Cross-window ordering of navigation commands — deferred, not started |
[...]

PR #2639, opened 2026-08-04
Multi-window e2e fixtures, layout persistence, per-window UI tests

rolfheij-sil:
Three multi-window slices that were separate PRs, now one review unit: behaviour-level e2e coverage for the multi-window work delivered in #2621 (**PT-4287**), per-window dock-layout/bounds/display persistence across restarts (**PT-4285**), and a regression spec pinning the per-window UI behaviours the delivered architecture already resolves structurally (**PT-4280**). Together: one new persistence service in main plus its renderer host, `electron-window-state` removed, 52 new unit tests, and three e2e specs (6 tests) covering second-window lifecycle, hosting takeover, quit, layout persistence across three sequential launches, and per-window overlays/dialogs/notifications/navigation/scroll groups.

Consolidated from #2637/#2638 on 2026-08-07 to shrink the review stack; per-part incremental diffs remain viewable:

- Part 1 (PT-4287, formerly #2637) — https://github.com/paranext/paranext-core/compare/pt-1891-multi-window...pt-4287-multi-window-e2e-fixtures
- Part 2 (PT-4285, formerly #2638) — https://github.com/paranext/paranext-core/compare/pt-4287-multi-window-e2e-fixtures...pt-4285-window-layout-persistence
- Part 3 (PT-4280, this branch) — https://github.com/paranext/paranext-core/compare/pt-4285-window-layout-persistence...pt-4280-per-window-ui-regression

**Stacked on #2621** (`pt-1891-multi-window`). Merge order: **#2621 → #2639 → #2649 → #2651**. Do not merge before #2621; rebases onto main as the stack merges.

---

## Part 1 — e2e lifecycle fixtures (PT-4287, formerly #2637)

Adds behaviour-level e2e coverage for the multi-window work delivered in #2621, as the groundwork slice of PT-4287. Three tests in `e2e-tests/tests/isolated/multi-window/`, each launching its own Electron instance:

1. **Second-window lifecycle** — a second window starts clean (window-scoped web view ids, no duplicate registrations at warn/error severity), generic window-service calls follow focus to window 2 and back, and closing the secondary window neither quits the app nor runs shutdown tasks.
2. **Hosting takeover** — closing the window hosting the app-global theme and scroll-group services hands hosting to the surviving window, proven by live PAPI reads plus a scroll-group write round-trip against the survivor (not just cached reads).
3. **Quit with two windows** — a real `app.quit()` runs the shutdown tasks exactly once (not per-window, not zero) and the process exits cleanly with code 0.

All assertions are outside-observable (PAPI over WebSocket, page content, user-facing log lines), deliberately: the plumbing underneath is expected to be refactored per the PT-4275 architecture discussion, and these tests are the safety net for that refactor. This part touched no `src/` or fixture code — only the new spec and a README index line.

### Falsifiability

Every test keys on the power-mode shutdown session-sync log marker (pre-configured `platform.interfaceMode: 'power'`), which fires exactly once per shutdown-task run — making "shutdown did not run" (test 1) and "ran exactly once" (test 3) real assertions rather than vacuous ones. Regressing id scoping, focus routing, either takeover subscription, the unreachable-object sweep, the write path, or the single-shutdown latch each fails a specific documented assertion.

### Notes for reviewers

- Test 3 watches the OS process `exit` event instead of Playwright's `ElectronApplication` `close` event: a graceful dev-mode quit leaves the .NET watcher child holding inherited stdio pipes, which that event waits on indefinitely. The leftover process group is reaped after the exit assertion.
- Headless compositors (observed on WSLg/Weston) honor a window's initial activation but ignore programmatic re-activation, so the focus helper escalates after 10 s to delivering the `focus` notification at the Electron boundary — everything the app does with the event still runs for real.
- Verified with two consecutive full green runs (3/3, no retries; ~55 s per run).

### Scope vs the ticket

PT-4287's full Definition of Done covers epic features that don't exist yet (create-via-menu, dock back, persistence, mixed-DPI matrix; depends on PT-4281/PT-4285). This part is the harness-groundwork slice covering what #2621 actually delivers; the rest fills in as those tickets land.

### Windows note (2026-08-04)

During native Windows verification of the stack, two of this spec's tests (`multi-window.spec.ts` — the takeover and quit tests' `createSecondWindow` call) timed out waiting for the second window on their **first attempt** in all three runs, and passed **every retry** (retries run in a fresh Playwright worker, so it correlates with worker reuse). Linux runs are unaffected. Not yet investigated; recorded in the epic's deferred-followups notes.

---

## Part 2 — window layout persistence (PT-4285, formerly #2638)

Implements PT-4285: every window's dock layout, bounds, and display persist across restarts, using the storage design decided on the ticket — **one structure holding all window layouts in order** (`window-layouts.json` in `userData`), owned and written only by the main process. Position in the list is the identity; window ids stay runtime-only.

### How it works

- At startup, main reads the structure and creates one window per entry in order, handing each window its entry. In **Simple mode only the main window is restored** (Simple is single-window per the epic's mode-switching spec); unrestored entries are preserved in the structure untouched, so a Simple session can never destroy a Power user's secondary windows.
- Renderers load their layout from their window's entry and **push** reconciled layout changes to main over internal (non-public) request handlers; main aggregates and debounces disk writes (temp-file + rename). At quit, each window's close handler flushes the full structure before shutdown tasks run; a **deliberate** window close instead rewrites the structure without that window, so it does not come back.
- Bounds are captured per window keeper-style (normal-state only, debounced) and restored through a visibility check against the currently connected displays — a window whose monitor went away comes back somewhere visible. **`electron-window-state` is removed**; the old `window-state.json` is still read once so an upgrade keeps the user's window placement.
- Legacy rule: with no structure file, exactly one window is restored from the pre-multi-window `localStorage` layout. This replaces the per-window prefixed-key persistence this branch previously carried (which cloned window 1's layout into every new window via its legacy-migration fallback) — per the ticket's "Note for PR #2621", the keyed model is now fully superseded.
- A window created mid-session starts **empty** (decided on the ticket); a deliberately-empty window open at quit is restored (empty). Phantom protection: layouts are reconciled on push (duplicate/orphaned tabs, empty panels stripped), and at load an entry whose tabs were *all* phantoms is dropped — it cannot resurrect a window — while an entry saved tab-less legitimately restores one.

### Verification

- **52 new unit tests** (persistence service 19, renderer host 12, bounds util 7, layout reconciliation 14), all behaviour-level; full suite green; typecheck green; `npm run build:types` produces an **empty `papi.d.ts` diff** — nothing new reaches the public API.
- **Live e2e**: new `window-layout-persistence.spec.ts` runs three sequential launches into one preserved profile — quit with two windows → both restored (main with its layout, secondary empty) at saved sizes with the persisted file holding exact pre-quit placements → deliberately close the secondary → relaunch restores exactly one. The existing multi-window spec was adapted to lock the empty-start behaviour explicitly. Both specs green in two consecutive full runs each, no retries.
- Also fixed en route: the e2e global setup only built the dev main-process bundle when missing, so e2e runs could silently test a **stale main process** against a fresh renderer; it now rebuilds on staleness (own commit).

### Notes for reviewers

- Consciously started despite the ticket's declared PT-4278/PT-4281 dependencies (rationale on the ticket): PT-4278 only moves *where* the orchestration lives later; PT-4281 is when secondary windows gain content to round-trip. Until then secondary windows persist existence/bounds/display.
- Not covered live, with reasons in the spec header: legacy single-window upgrade (unit-covered), monitor-gone re-placement (pure function, unit-covered), window *position* restore (WSLg assigns positions itself; position is asserted exactly on the save side, size on the restore side per the app's own on/off-display rule), tab-bearing secondary windows (needs PT-4281).
- The `activate` (macOS) path re-runs the full restore; not verifiable on this machine.

### Windows-native verification (2026-08-04)

The branch was additionally verified on native Windows (fresh clone in a Windows-side folder, win32 Node, real windowing): full unit suite and typecheck green, and the persistence e2e passed twice — with **exact position restore** (settled placement == requested; the persisted file held both windows' exact pre-quit bounds, position included) and both windows exercising the **on-display exact-saved-size restore branch** that WSLg cannot reach. The upgrade test honored the legacy keeper-file size. Two Windows-only defects surfaced and are fixed here: POSIX separators hardcoded in a unit-test expectation, and e2e teardown crashing on a disposed `ElectronApplication` after a graceful quit. Run at 100% DPI only; >100% scaling remains for the epic's manual mixed-DPI matrix (PT-4287).

---

## Part 3 — per-window UI regression tests (PT-4280, this branch)

Delivers PT-4280 — which a premise audit found had shrunk to a testing task: the ticket's four claimed per-document-singleton defects were authored against pre-multi-window code, and the delivered architecture (every window is a full renderer with its own document, React root, OverlayHost, Toaster, and per-window module state) resolves all of them structurally. This part pins those behaviours so they can't regress, and settles the audit's one uncertain edge with a live probe.

### What's pinned (one spec, one Electron launch, two windows)

1. An overlay raised from window 2's web view (via the iframe's inherited papi) renders in window 2's document only.
2. A modal dialog via the generic `dialog:showDialog` lands in the focused window; Escape resolves the pending request.
3. A notification from work in window 2 appears in window 2 with window 1 minimized — the ticket's explicit DoD case.
4. The navigation-target fallback never reaches across windows: window 2's BCV control stays disabled while window 1 has an editor; each fresh web view attaches only in the window that was focused when it was opened.
5. Scroll groups: two windows on different groups each follow their own group's reference while sharing the app-global group set (window 1 on A, window 2 on C; navigation in each group moves only its subscribers, across windows).

**Probe result (documented, no product change needed):** a notification sent while *no* window has focus routes to the last-focused window even when minimized, renders there, and is visible on restore — matching single-window semantics.

### Notes for reviewers

- The spec doubles as adjudication of the premise audit: any wrong "already correct" verdict fails its assertion. All six scenarios passed in two consecutive runs plus a full multi-window-folder run (6/6, no retries).
- No production code changes in this part; no `papi.d.ts` impact. Two shared helpers moved from `multi-window.spec.ts` into `multi-window.util.ts`.
- WSLg cannot reliably report a programmatic minimize back through `isMinimized()`, so scenario 3's minimize is requested-and-logged while the product claim under test (focus-based routing) is asserted deterministically.

---

## Deep review (2026-08-05)

**What ran:** a 25-agent adversarially-verified `/code-review` at high effort, plus four `/review-paratext` analyzers (api, style, compliance, ux) — over the full #2637→#2638→#2639 stack range, i.e. exactly the range this consolidated PR now covers.

**Analyzer verdicts:** zero critical/important issues across all four analyzers; minors only, all addressed or consciously acknowledged (see below).

### Findings in Part 1 (PT-4287)

- None anchored to Part 1's own diff — all 10 workflow findings landed in code introduced later in the stack (Parts 2 and 3).
- Part 1's specs later gained shared helpers from Part 3 (the quit epilogue and editor-open helper).
- Its known Windows-only first-attempt flake note (documented in the isolated e2e README) stands and was not touched by this review.

### Findings in Part 2 (PT-4285)

- CONFIRMED data-loss: a debounced write scheduled after the quit flush survived window removal and could rewrite `window-layouts.json` with closed windows trimmed.
- CONFIRMED data-loss: any failed `windowLayout:get` fell back to the shared legacy `localStorage` layout, cloning it into secondary/mid-session windows.
- CONFIRMED durability regression: the written structure was snapshotted at enqueue time, dropping a layout pushed while the quit flush waited behind an in-flight write.
- PLAUSIBLE race: rapid macOS dock re-activations could run `restoreWindows` concurrently and duplicate the restored window set.
- PLAUSIBLE stale read: `loadWindowLayouts` read the file without draining the in-flight write chain.
- CONFIRMED upgrade gap: layouts saved under the base branch's per-window prefixed `localStorage` keys were ignored.
- CONFIRMED e2e-infra defect: the stale-bundle check in `global-setup.ts` missed bundle inputs (`lib/platform-bible-utils/src`, `.erb/configs`).
- PLAUSIBLE e2e-infra defect: the teardown catch in `helpers.ts` treated any `process()` throw as "already exited", skipping the kill escalation.
- Cleanup: two copy-pasted saved-layout tab walkers consolidated into one.
- Minor (analyzer): pushed layouts were stored verbatim, relying on renderer-side reconciliation — now reconciled server-side too.
- Minor (analyzer): defensive branches (`assignEntryToWindow` guards; save-request invalid args) were untested — four new tests pin them.
- Minor (analyzer): duplicated 4-line bounds-capture cancel blocks — consolidated into one closure.

**Fixing commit:** 840c086e560021b8ba371e79666b5b8c4fd9621c — "fix: close quit-time persistence races found in deep review" (PT-4285).

### Findings in Part 3 (PT-4280)

- Cleanup: `per-window-ui.spec.ts` forked the fixture helper's scripture-editor open-retry loop (and had dropped its initial-layout race guard).
- Minor (analyzer): the graceful-quit epilogue was repeated six times across specs, with two hand-rolled first-window waits.

**Fixing commit:** 9eb079b367f4abb29d207aaca5c01958d6c3327b — "test: dedupe quit epilogues and share the editor-open helper" (PT-4280).

### Shared verification and interview

**Verification:** 14 new unit tests (6 written red-first), full suite + typecheck + lint green, 6/6 multi-window e2e tests green in one run with zero retries, and an empty `papi.d.ts` diff.

**Author interview:** Interview self-answered from session context by the AI assistant under the author's instruction; no questions required escalation. Deliberate decisions on record: empty-start windows, Simple-mode single-window restore, log-marker shutdown assertions, storage design per the ticket.

---

## Round-T2 fixes (2026-08-10)

Three approved changes from the latest review round:

- **macOS quit re-issue** — every window's `close` handler calls `preventDefault()` to run its shutdown work, which cancels an in-flight Electron quit; on darwin nothing re-issued it, so Cmd+Q destroyed every window and left the process resident forever (`will-quit` never ran, and the still-set quit latch made dock reactivation refuse too). `window-all-closed` now re-issues the quit when one was requested (`isAppQuitRequested()`), keeping the macOS keep-resident convention only when no quit is in flight. The re-issued quit walks the normal `before-quit` → `will-quit` teardown, exactly the path the unconditional `app.quit()` already takes on Linux/Windows. Darwin verification is manual (Cmd+Q); the suite's Linux quit path is unchanged by construction.
- **Quit-timeout diagnostics** — the multi-window e2e's 120 s quit timeout used to reject with a bare message while the suites' captured app output — the one artifact that can name the statement shutdown is stuck on — was discarded. The timeout error now carries the last 60 lines of the capture, and the budget docblock states explicitly that "a healthy quit lands well under a minute" holds only in this suite's no-S/R configuration (with an S/R extension registered, the sync's own bound is 600 s, past the whole budget).
- **Simple-mode menu gating** — the "Open new window" main-menu entry is hidden in Simple mode via `hiddenInterfaceModes: ["simple"]` (menu-only by design; the `platform.createWindow` command itself is not gated). Pinned red-first by two new tests asserting against the real shipped menu document: the item absent from the main menu in `simple` mode, present in `power` mode.

---

AI-assisted — [session](https://claude.ai/code/session_01Lf3XmoA9StqLNDiMBqZVTy)

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2639)


e2e-tests/tests/isolated/multi-window/per-window-ui.spec.ts
```diff
@@ -0,0 +1,459 @@
...
+    // the empty-start behaviour itself is locked by multi-window.spec.ts.
+    await expectWindowDockEmpty(page2);
+    await expect(page2.locator(BCV_TRIGGER).first()).toBeDisabled();
+    logStep(`window ${window2Id} up; both BCV controls disabled with no navigable tabs anywhere`);
+
+    // ── Navigation target follows only the OWN window's tabs ───────────────────────────────────
+    // Open an editor in window 1 (the generic web-view command routes to the focused window).
+    await waitForAtLeastOneProjectMetadata(WEBSOCKET_PORT, 60_000);
+    await focusWindowAndWaitForRouting(electronApp, window1Id);
+    const editorId1 = await openScriptureEditorForProject(mainPage, SAMPLE_WEB_PROJECT_ID);
```
rolfheij-sil, line 289: 🤖 Claude: Cleanup: the spec forked the fixture helper's scripture-editor open-retry loop (and had dropped its initial-layout race guard). Fixed in 9eb079b367f4abb29d207aaca5c01958d6c3327b: the fixture helper (`openScriptureEditorForProject`) now takes a target page and returns the opened web-view id; the spec uses it directly. The loadLayout-race guard is restored for window 1, with a documented `skipInitialLayoutGuard` opt-out for window 2 (an empty secondary window that never gets an initial iframe for the guard to wait on).

e2e-tests/tests/isolated/multi-window/multi-window.util.ts
```diff
@@ -351,4 +429,28 @@ export async function quitAppAndWaitForExit(
...
 
+/**
+ * The graceful-quit epilogue shared by the multi-window suites: trigger a real quit and wait for
+ * the OS process to exit (see {@link quitAppAndWaitForExit}), assert the exit was clean (code 0, no
+ * signal), then sweep everything the given capture recorded for {@link FAULT_MARKERS} and for
+ * warn/error-severity duplicate registrations ({@link DUPLICATE_REGISTRATION_PATTERN}).
+ *
+ * @param label Prefix for the step-log line recording the exit (e.g. `'phase 1'`).
+ */
+export async function quitAndExpectCleanExit(
```
rolfheij-sil, line 451: 🤖 Claude: Minor (analyzer): the graceful-quit epilogue (trigger quit, wait for exit, assert clean exit, sweep fault markers) was repeated six times across specs; two specs also hand-rolled their own first-window waits. Fixed in 9eb079b367f4abb29d207aaca5c01958d6c3327b: a shared `quitAndExpectCleanExit()` helper is used everywhere the epilogue was repeated, and `waitForAppPages` is used for the first-window waits.

rolfheij-sil (commented):
🤖 Claude: Deep review results for this PR (part of the #2637→#2638→#2639 stack). A 25-agent adversarially-verified code-review workflow plus four /review-paratext analyzers (api/style/compliance/ux) ran over the full stack range. The analyzers found zero critical/important issues (minors only). The workflow found 2 findings anchored to this PR's diff, both cleanups, already fixed in commit 9eb079b367f4abb29d207aaca5c01958d6c3327b ("test: dedupe quit epilogues and share the editor-open helper"). See the 2 inline comments below for detail. Verification: full unit suite + typecheck + lint green, all 6 multi-window e2e tests green in one run with zero retries, and an empty `papi.d.ts` diff.

timothy-mccormack:
### Code review

Found 1 issue (minor, docs-only):

1. The `WindowLayoutGetResponse` JSDoc says `empty` covers "one whose saved layout is unusable", but no code path produces that. A phantom-only entry is dropped from `entries` in `loadWindowLayouts` before any window is created, so `windowLayout:get` is never called for it; and the main entry is always kept with an emptied `dockbox`, so it comes back as `kind: 'entry'`. The condition that actually yields `empty` for a restored window is an entry with no `layout` key at all (a bounds-only secondary), which is a different thing from "unusable". Both behaviors are pinned by existing tests (`drops a phantom-only entry so it cannot resurrect a window`, `an entry with no saved layout restores an empty window, not the legacy layout`) — it's the wording that's off, not the code.

https://github.com/paranext/paranext-core/blob/7c504f2678458a982d842b72aacb096e22998743/src/shared/data/window-layout-persistence.model.ts#L61-L64

Everything else checked out. Reviewed for CLAUDE.md compliance (e2e test placement, `no-null`, `no-type-assertion` disables all carry reasons, no hardcoded secrets), obvious bugs, git history against the `electron-window-state` removal, prior review comments on #2621/#2637/#2638, and code-comment contracts. The quit/close flush ordering, write serialization, and `fileSlots`/`trackedWindows` bookkeeping all hold up.

rolfheij-sil:
🤖 Claude: Verified your finding limb by limb against `handleGetLayoutRequest` — it holds, and the
wording is now fixed in `26992d794d8`.

Confirming each step:

- A phantom-only **secondary** entry is dropped by the filter in `loadWindowLayouts`, so no window
  is created and `windowLayout:get` is never called for it. Correct.
- The main entry is exempt from that filter (`index === mainIndex` short-circuits), and
  `reconcileSavedLayout` always returns an object — an emptied `dockbox` is still truthy — so
  `tracked.layout` is set and the answer is `kind: 'entry'`. Correct.
- What actually yields `empty` for a *restored* window is an entry whose `layout` key is absent —
  a bounds-only secondary — which, as you say, is a different thing from "unusable". Correct.

One nuance on "no code path produces that": `parseEntry` reads the layout through `asRecord`, which
rejects any non-object value, not just an absent key. So a structure holding `"layout": null` (or a
string, or a number) on a secondary entry does land on `empty` — an unusable saved layout producing
`empty` is reachable, just not by the route the old wording implied. Marginal, and it doesn't change
your conclusion: the sentence described a behaviour that isn't what the code decides on.

While tracing it we found two more problems in the same block, so the correction goes further than
the `empty` bullet:

- The `legacy` bullet was also wrong, and more consequentially. It read "no layout has ever been
  captured for this window's entry", which describes every layout-less entry — but `usesLegacyLayout`
  is set only for the single window of a legacy startup and for the **main** entry. A layout-less
  *secondary* deliberately gets `empty`, because falling back there would clone the pre-multi-window
  layout into it. The doc as written described the bug the code exists to avoid.
- The two defensive `empty` returns (a request with no window id, and one for an untracked window)
  were undocumented.

The rewritten block now spells out all three kinds, both `legacy` cases and why a secondary is
excluded, both defensive returns, and the phantom-only rule including the main-entry exception you
identified.

Separately: since this was the last open item, we ran an independent max-depth review over the
branch before finalising. It found no critical or important issues — it corroborated the quit/close
flush ordering, write serialisation, and `fileSlots`/`trackedWindows` bookkeeping you checked — and
surfaced three low-to-medium hardening items, each fixed here rather than deferred:

- `loadLayout()`'s no-argument path became async in this PR (it now awaits `windowLayout:get`, up to
  ~4 s with retries) and ends by replacing the whole dock, with nothing stopping a second load from
  starting meanwhile. An interface-mode switch during that window could finish first and then be
  overwritten by the earlier load's answer. Now generation-guarded: a superseded load drops its
  result instead of applying it (`4dd7841441a`).
- `reconcileNode` dropped tabs without touching the panel's `activeId`, so a reconciled layout could
  be persisted holding an active-tab reference to a tab no longer in it. It is now re-pointed at the
  first surviving tab, matching the convention `window-scoped-web-view-ids.util.ts` already uses
  (`a0934bc9321`).
- The write chain had no `.catch`: `buildStructure` runs inside the chained `.then`, outside
  `writeStructureToDisk`'s guard, so a throw there would have left `writeChain` rejected for the rest
  of the session and surfaced as a rejection from `writeNow` — including the quit-time flush. The
  chain now catches and resolves, and the quit path treats a persistence failure as non-fatal so it
  still reaches the session sync (`a63fd185b20`).

Each is small; red-first tests cover the first two, and consolidating persistence onto a single
error boundary made the chain's swallow tests actually exercise it.

One e2e-infra fix rides along in the same range (`87a80fdbc40`): the stale-bundle check added
earlier in this PR rebuilt via `prestart`, which emits `preload.bundle.dev.js` — a file nothing
reads. Unpackaged, the app loads `.erb/dll/preload.js`, which only the separate preload config
produces, so a run could use a preload bundle left over from an earlier checkout. The check now
covers both bundles.

Could you submit the formal GitHub approval when you're satisfied with the correction?

src/main/services/window-layout-persistence.service.ts
```diff
@@ -0,0 +1,506 @@
...
+ */
+const LEGACY_WINDOW_STATE_FILE_NAME = 'window-state.json';
+
+/** How long to coalesce updates before writing the structure to disk */
+const WRITE_DEBOUNCE_MS = 500;
+
+/**
+ * What startup (or macOS re-activation) should create:
+ *
+ * - `restore`: one window per entry — or only the `mainEntryIndex` entry's window in simple interface
```
rolfheij-sil, line 52: 🤖 Claude: Anchoring the "what should happen on a Power → Simple switch with more than one window open" question here, because this docblock is the layout data model any transition rule has to work with.

What this PR ships: ONE layout structure, not one per mode. In Simple mode startup restores only the `mainEntryIndex` entry's window and leaves every other entry in the file untouched (:11, :52-53, and the plan built at :253). Main is not mode-blind about it either — it reads `platform.interfaceMode` before restoring (`src/main/main.ts:1332-1340`) and again in the startup and shutdown tasks (`src/main/shutdown-tasks.ts:94-104,166-175`).

Two gaps are real, and they are narrower than "nothing enforces single-window Simple":

1. Nothing subscribes to `platform.interfaceMode`, so a LIVE switch to Simple leaves the windows that are already open exactly where they are.
2. "Open new window" is contributed un-gated (`src/extension-host/data/menu.data.json:43-49`, contributed on #2621), so a user in Simple mode can open a second window today. That makes "Simple is single-window" a restore-time property right now, not an invariant — worth knowing before any review finding is closed on the strength of it.

Giving each mode its own layout data — separate data per mode including which windows are open, and on a transition keeping the window the user clicked in, closing the rest, opening the new mode's layout — would REPLACE the model this docblock describes rather than extend it. That is a decision for PT-4285's owner and UX more than for us, and it is why we have not started it.

The cheapest piece is separable: gating "Open new window" while in Simple mode removes the way a user reaches a second window, without touching the data model, and could go first on its own.

Questions:
(a) How do you want to proceed on the per-mode layout data?
(b) Are you happy for the "Open new window" gating to go first on its own, while (a) finds its owner?

rolfheij-sil:
🤖 Claude: For the reviewer — open questions on this PR

**1. Simple-mode transitions and the layout data model.** See the inline thread at `src/main/services/window-layout-persistence.service.ts:52`. This PR ships one layout structure for both modes (Simple restores only the main entry, the other entries preserved untouched); giving each mode its own structure would replace that model rather than extend it, so it needs the PT-4285 owner and UX rather than just us. The separable half — gating "Open new window" while in Simple mode — removes the way a user reaches a second window and could go first on its own. Questions are in the thread: how to proceed, and whether the gating change can go first.

**2. Where the re-triage that follows the transition rule lives.** Thirteen findings from the round-4 review on #2621 — the ones grouped under its proposed fixes B and D (https://github.com/paranext/paranext-core/pull/2621#pullrequestreview-4877614064) — are parked until the Simple-mode transition rule is decided. They are tracked on a running small-items list kept with this epic's working notes rather than in Jira, because they have no scheduled home yet: a ticket gets created once the work is scheduled, not before. They must not be re-triaged before the transition rule lands, because several of them are only "not a bug" once Simple genuinely is single-window, which today it is not (see the thread above). No question here — flagging the ordering so it is visible.

**Current state notes on this PR** (facts, for reading the diff)

- Startup reads `platform.interfaceMode` before deciding what to restore (`src/main/main.ts:1332-1340`); in Simple mode only the `mainEntryIndex` entry's window is created and every other entry stays in the file (`src/main/services/window-layout-persistence.service.ts:11,52-53,253`).
- The shutdown/startup tasks read the same setting (`src/main/shutdown-tasks.ts:94-104,166-175`), so mode-dependent behaviour in main is not limited to restore.
- A secondary entry with no layout stays empty rather than cloning the legacy layout (`window-layout-persistence.service.ts:69-75,295`), so a restore cannot mint a duplicate of a tab another window still holds.
- "Open new window" (`src/extension-host/data/menu.data.json:43-49`) is contributed on #2621 and is not gated by interface mode.

**Per-part diffs:** the three compare links at the top of the PR body split this branch into its parts if you would rather walk them separately.

src/main/services/window-layout-persistence.service.ts
timothy-mccormack: Picking this up informally — not a listed reviewer, so treat (a) as deferring to the people you named.

**(b) Yes, and it's cheaper than "a separable piece of work" — the gating mechanism already exists and is already live.**

- `filterItemsForInterfaceMode()` (`src/extension-host/services/menu-data.service-host.ts:35-39`) drops any item whose `hiddenInterfaceModes` includes the current mode.
- The menu data service subscribes to `platform.interfaceMode` and re-notifies on change (`:146-182`).
- Covered by tests, including `menu data updates live when platform.interfaceMode changes after subscribe resolves` (`menu-data.service-host.test.tsx:405`).
- **Nothing in `menu.data.json` uses it yet** — zero occurrences in the file.

So gating "Open new window" looks like one field on the existing entry:

```json
{
  "label": "%mainMenu_openNewWindow%",
  "localizeNotes": "Application main menu > Project > Open new window",
  "group": "platform.projectMisc",
  "order": 1,
  "command": "platform.createWindow",
  "hiddenInterfaceModes": ["simple"]
}
```

Because the subscription is already live, the item would disappear the moment a user switches to Simple — so this closes gap 2 as a live property, not a restart-time one. The macOS menubar path is covered too: `getUnlocalizedMainMenu` applies the same filter as `getMainMenu` (`:92`, with a comment at `:91` saying a hidden item would otherwise still appear there), and `getWebViewMenu` filters both its `topMenu` and `contextMenu` (`:112`, `:118`).

**One premise to correct, because it changes the cost estimate:** the comment says "Nothing subscribes to `platform.interfaceMode`, so a LIVE switch to Simple leaves the windows that are already open exactly where they are." The conclusion about windows is right, but something does subscribe — the menu data service, live. The narrower accurate claim is that **nothing in main subscribes for window lifecycle**, which is the gap that matters for (a).

**(a)** Nothing to add beyond agreeing it is a data-model decision rather than a review comment. One thing worth flagging for whoever owns it: gating "Open new window" makes "Simple is single-window" true for *new* sessions, but a user who was in Power with three windows and switches to Simple still has three windows until restart. So (b) narrows the gap without closing it, and the transition rule still has to say what happens to already-open windows.

On the parked round-4 findings — no objection to the ordering, and the point that several are only "not a bug" once Simple genuinely is single-window is the right reason to hold them. Noting that (b) landing on its own does **not** make them re-triageable, for the reason in the paragraph above.

timothy-mccormack:
Verified `26992d794d8` against the code rather than the description. It holds, and it goes further than what I flagged.

**Your `asRecord` nuance is right, and my "no code path produces that" was too absolute.** `asRecord` returns `undefined` for any non-object value, not just an absent key (`window-layout-persistence.service.ts:104-110`), so `parseEntry` maps `"layout": null` — or a string, or a number — to `layout: undefined` (`:146-158`). On a secondary that leaves `usesLegacyLayout` false (`:296`), and `handleGetLayoutRequest` falls through to `{ kind: 'empty' }` (`:466-468`). An unusable stored layout does reach `empty`; it just isn't the phantom-only route the old wording implied. The rewritten bullet names exactly that case ("its stored value is not object-shaped"), so the doc and the code now agree.

**The `legacy` correction is the more valuable half, and I missed it.** `usesLegacyLayout` is set in only two places — the legacy-startup single window (`:310`) and `slot.entry.layout === undefined && entryIndex === mainSlotIndex` (`:296`) — so the old "no layout has ever been captured for this window's entry" did describe the clone-the-legacy-layout-into-every-secondary bug the code exists to prevent. That was a worse defect than the `empty` bullet I raised.

I also checked the four commits that rode along, since they postdate my review:

- `a63fd185b20` — correct, and the placement argument is the substance of it: `buildStructure` ran inside the chained `.then` but outside `writeStructureToDisk`'s `try`, so only a boundary at `enqueueWrite` covers both. A throw there would have left `writeChain` permanently rejected and surfaced as a rejection out of the quit-time flush.
- `4dd7841441a` — the generation guard is checked after the mode read *and* threaded into `getPersistedLayout` as `isSuperseded`, which is the part that matters: the retry path is the multi-second window, so a guard that only bracketed the mode read would have left the real race open.
- `a0934bc9321` — re-pointing `activeId` at the first surviving tab rather than deleting it matches what `window-scoped-web-view-ids.util.ts` already does, and applying it on every node instead of only panels is right given the input is arbitrary JSON off disk.
- `87a80fdbc40` — building the bundle the unpackaged app actually loads (`.erb/dll/preload.js`) rather than the one `prestart` emitted.

One open item, and it is the only thing I would want closed before a formal approval rather than after: **this branch has no CI signal at all.** `gh pr checks 2639` reports no checks on `pt-4280-per-window-ui-regression`, at this head (`a63fd185b20`) and at every head before it. Every green result on this PR — unit suite, typecheck, lint, the 6 multi-window e2e tests, the red-first tests on the two fixes above — is a locally-run claim in the PR description and commit messages. I have read the diffs and they support the claims, but reading a diff is not running a suite, and a +4.9k-line change to window lifecycle and persistence is not where I would want that gap. If checks are expected to run on this branch and are not, that looks worth chasing before merge; if they only run against `main` because of the stacked base, saying so explicitly would close it.

Nothing else outstanding from my side on the docblock.

timothy-mccormack:
### macOS e2e run — unmeasured platform, reporting what it does

Not a defect claim. You've confirmed the 6/6 green run wasn't on macOS, so this is the first measurement of that platform rather than evidence anything here regressed. Posting it because PT-4287 owns cross-platform verification and this is the data it doesn't have yet.

**Environment:** macOS 13.7.8 (Apple M1 Max), dotnet 8.0.421, clean checkout of `a63fd185b201722ab32d3f976fb6bc429413f143` (current head), no local modifications.

**What passes on macOS:**

| Check | Result |
|---|---|
| `npm run test:core -- --run` | 139 files, **1722/1722 pass** |
| `npm run typecheck` | exit 0 |
| `npm run lint` | exit 0 (5 warnings, all in `platform-scripture-editor`, untouched by this PR) |

**What doesn't:** `npm run test:e2e:isolated multi-window`

```
Running 6 tests using 1 worker
✓ second window starts empty, focus routing follows the focused window…   (36.5s)
✓ closing the window hosting the app-global services hands hosting over    (12.1s)
✘ quitting with two windows open runs the shutdown tasks exactly once…     (2.2m)
✓ overlays, dialogs, notifications, and navigation targets stay…           (36.6s)
✘ windows, layouts, and bounds survive relaunch; a deliberately closed…    (2.4m)
✘ a pre-multi-window profile upgrades to one window with its legacy…       (2.2m)

3 failed, 3 passed (15.1m)   exit 1
```

All six failures — three tests, each on first attempt and on retry #1 — are the same error:

```
Error: Electron process did not exit within 120 s of app.quit()
  at isolated/multi-window/multi-window.util.ts:415
```

The trace shows the app reaching `[multi-window +7.9s] both windows up; quitting`, then producing **no further output at all**. Still `exitCode=null` when the fixture force-killed it two minutes later.

**Ruled out:** stray processes before or after (none), missing/broken .NET (dotnet 8.0.421 present, `c-sharp/bin` built), flake (identical on every retry), bad checkout (unit/typecheck/lint green on the same tree). Electron launches, two windows open, PAPI round-trips and the UI works — the three passing tests demonstrate that.

**The part that makes this ambiguous, and why I'm not calling it a regression:** the three failing tests are exactly the three that call `quitAndExpectCleanExit`; the three that pass never quit gracefully, because the fixture SIGKILLs the process group as its normal teardown (visible in their logs too). And `quitAppAndWaitForExit` is **new in this PR** — no pre-existing e2e test asserts that the app exits cleanly on any platform. So there is no macOS baseline to compare against, and the hang may well predate this branch entirely. What's demonstrable is narrower: this PR is the first thing to assert graceful quit, and on macOS that assertion fails.

Worth noting `quitAppAndWaitForExit`'s own docblock already anticipates platform-specific quit trouble — the `.NET watcher child holding the inherited pipe write-ends` note, raised there for Linux. Whatever holds the process open here may be related.

Happy to instrument the shutdown path and pin down what's actually blocking exit if that's useful — it needs a real debugging pass rather than a quick look, so say if you'd like it.

timothy-mccormack (approved):

src/main/services/window-layout-persistence.service.ts
rolfheij-sil: 🤖 Claude: You're right, and the claim as I posted it was wrong — correcting it here plainly.

I wrote that nothing subscribes to `platform.interfaceMode`. In fact there are eight live
subscribers — the menu data service you named, plus two renderer core services (the dock layout
reload and the window service's navigation target), a renderer hook, and four extension consumers.
Your correction was right in substance and understated the extent.

The narrower claim you proposed is the accurate one, verified: nothing in `src/main` subscribes to
`platform.interfaceMode`. All four main-process touches are one-shot `settingsService.get` calls
at startup/shutdown occasions; the only settings subscription in `src/main` at all is
`platform.zoomFactor` (main.ts:1272). So the conclusion the original claim was supporting — that
the main process doesn't react to mode changes for window lifecycle — survives, but only in that
narrower form, and I should have written it that way the first time.

rolfheij-sil: 🤖 Claude: Landed as you suggested (4ebac21b0e7): one `"hiddenInterfaceModes": ["simple"]` field on
the "Open new window" entry, with a red-first test asserting the item is absent from the main menu
in Simple and present in Power. Two notes for the record: the mechanism was already in production
use (the scripture editor's menus.json uses the same field), so this rides shipped precedent; and
this hides the UI route only — `platform.createWindow` stays registered and callable over PAPI,
so "Simple cannot create a second window" is deliberately NOT an invariant this buys. Per your
own point, the parked single-window findings stay parked on the mode-transition rule, not on this.

rolfheij-sil:
🤖 Claude: You found a real shipping bug, and it's now fixed on this PR — details below, plus one
ask that replaces the instrumentation pass you kindly offered.

Everything structural in your report checked out: the :415 assertion, exactly 3 of 6 tests going
through quitAndExpectCleanExit, the suite being genuinely new with no macOS baseline, and the
process really still being alive when exitCode=null was read.

What we traced from source: the window close handler calls event.preventDefault() unconditionally
(it has to, to run shutdown work before the window goes away), and on Electron that cancels an
in-flight quit. The handler then destroys the window, window-all-closed fires — and that handler
re-issues app.quit() only off macOS. So on macOS, ⌘Q tears down every window, nothing re-issues
the quit, will-quit never runs, and the app stays resident forever with zero windows. The dock
icon can't reopen a window in that state either. A second ⌘Q exits, which is probably why this
was never noticed.

Three corrections to details of your report, none of which change your conclusion:

1. "May well predate this branch" is too cautious — it definitely does. The structure is
   byte-identical at the merge base, on main, and at this head, and `git log -S` bottoms out at
   the original electron-react-boilerplate import. This is a repo-age defect your suite detected.
2. The "no further output at all" line in your evidence can't support an inference about where the
   app stopped: Playwright never forwards the Electron child's stdout into the report, and our
   fixture's own capture was read only after a successful exit — on the failure path it was
   collected and then thrown away. That's a fixture bug, and it's also fixed in this round: the
   120 s timeout error now carries the tail of the captured app output (9159bc16b5d), so a
   failing run names the stuck statement instead of going silent.
3. The .NET-watcher/pipe lead is a dead end by design: those pipe write-ends govern the `close`
   event, and the fixture deliberately watches `exit` — exitCode=null actually rules that
   mechanism out rather than implicating it.

The fix (461030400a0): window-all-closed re-issues the quit when one is in flight —
`if (process.platform !== 'darwin' || isAppQuitRequested()) app.quit()`. The latch is set in
before-quit, so the macOS red-button convention (close all windows, app stays resident) is
untouched; only a real requested quit is re-issued.

The ask, instead of the shutdown-instrumentation pass you offered: one link in this chain —
Electron treating a cancelled window close as a cancelled quit — we reasoned from source but could
not measure without a Mac. Could you pull the branch and ⌘Q the dev app? Before the fix it should
need a second ⌘Q (or leave the app resident); with the fix it should exit on the first press. If
you have the time, re-running the three failing isolated tests would confirm it end to end. That's
a much smaller use of your time than what you offered, and it settles the one thing we couldn't.

rolfheij-sil:
🤖 Claude: Closing the CI question you raised — your second explanation was the right one, and
it's now moot in the good direction.

The mechanism: both workflows with a `pull_request` trigger filter on the BASE branch
(`branches: ['main', ...]`), and this PR's base was the multi-window foundation branch from the
day it opened until that PR merged — so no event in this branch's life ever matched the filter.
Not actor-approval gating, not paths-ignore; we checked the alternatives.

Since the base auto-retargeted to main, this branch now gets real runs — you can see its
first-ever Test/CodeQL runs on the PR. Honest caveat: the current runs are red, and it's not this
branch — every OS job dies in "Install packages" on an upstream breakage (the scripture-editors
`platform-yalc` branch was force-rebuilt and its dependency change breaks the yalc install path
with "TypeError: expand is not a function"). main is red with the identical failure; the editor
team is being pinged. When their fix lands, a re-run should go green with no change on our side.

Also worth saying: main has required status checks (CodeQL + the three OS builds), so this PR
physically can't merge until those run green — the gap you flagged closes itself as a matter of
branch policy, not memory. And thank you for approving ahead of this being closed; noted and
appreciated.

timothy-mccormack:
### macOS ⌘Q verification — the fix holds, and the pre-fix behaviour is confirmed as described

Ran the A/B you asked for. Both halves behave exactly as your source-level reasoning predicted, including the two secondary symptoms.

**Environment:** macOS 13.7.8 (Apple M1 Max), dev mode (`npm start`), single window, clean tree, no local modifications.

#### Before the fix — `window-all-closed` without `isAppQuitRequested()`

One ⌘Q press:

- All windows disappeared from screen — visually indistinguishable from a quit.
- The main process was **still alive**:

```
$ pgrep -fl "paranext-core/node_modules/electron/dist/Electron.app"
14278 .../Electron --require .../electronmon/src/hook.js .
14327 .../Electron Helper -r ts-node/register/transpile-only .../extension-host.ts
14328 .../Electron Helper (GPU) --type=gpu-process
14329 .../Electron Helper --type=utility --utility-sub-type=network.mojom.NetworkService
```

- Clicking the Dock icon opened **no window** — the stuck quit latch making `activate` refuse, as you described.
- A **second** ⌘Q exited cleanly (`pgrep` then returned nothing).

Worth noting 14327: the extension host was still resident too, so this is not merely an idle main process — `will-quit`'s graceful teardown demonstrably never ran.

One accuracy note on which build this was. The pre-fix session under test was built from `a63fd185b20` (the pre-rebase head I had checked out) rather than `6427eac86c8`. `src/main/main.ts` is byte-identical between those two commits, so the `window-all-closed` code exercised is exactly the pre-fix version — but I'd rather state the commit I actually ran than the one I intended to.

#### With the fix — `4ebac21b0e7`

Fresh `npm start` (so `prestart` rebuilt the main bundle), single ⌘Q press:

- App terminated completely on the **first** press.
- No Activity Monitor entry.
- No running-indicator dot in the Dock.
- `pgrep -fl "paranext-core/node_modules/electron/dist/Electron.app"` → no output.

#### Isolated e2e suite — now green on macOS

Re-ran `npm run test:e2e:isolated multi-window` on `4ebac21b0e7`, same machine as the 10 Aug report:

```
Running 6 tests using 1 worker
  ✓ 1 multi-window.spec.ts:303 › second window starts empty, focus routing follows…      (32.6s)
  ✓ 2 multi-window.spec.ts:420 › closing the window hosting the app-global services…     (13.8s)
  ✓ 3 multi-window.spec.ts:527 › quitting with two windows open runs the shutdown…       (15.1s)
  ✓ 4 per-window-ui.spec.ts:260 › overlays, dialogs, notifications, and navigation…      (39.0s)
  ✓ 5 window-layout-persistence.spec.ts:375 › windows, layouts, and bounds survive…      (1.2m)
  ✓ 6 window-layout-persistence.spec.ts:587 › a pre-multi-window profile upgrades…       (50.8s)

  6 passed (3.8m)
```

Against 10 Aug's `3 failed, 3 passed (15.1m)` — and the runtime collapse from 15.1m to 3.8m is itself the signature, since the old figure was mostly three 120 s quit timeouts plus their retries.

The three that previously failed are exactly the three that call `quitAndExpectCleanExit`, and each now reports a real graceful exit rather than a force-kill:

```
[multi-window +12.4s] quit: exited with code 0 signal undefined
[teardown] Closing Electron app... pid=29549 exitCode=0 signalCode=null
```

```
[window-layout-persistence +26.6s] phase 1: exited with code 0 signal undefined
[window-layout-persistence +48.2s] phase 2: exited with code 0 signal undefined
[window-layout-persistence +72.9s] phase 3: exited with code 0 signal undefined
```

```
[window-layout-upgrade +18.2s] launch A: exited with code 0 signal undefined
[window-layout-upgrade +50.8s] launch B: exited with code 0 signal undefined
```

Every one of those eight graceful quits ends in `exitCode=0` with no SIGKILL escalation — where before, the fixture waited the full 120 s and force-killed with `exitCode=null`. The three tests that never quit gracefully (1, 2, 4) still take the SIGKILL teardown path, unchanged.

That closes the end-to-end half: the 120 s `app.quit()` timeout at `multi-window.util.ts:415` was the darwin quit defect, not a fixture or environment problem.

Incidental: the run also exercised your new staleness check — it reported *"Development preload bundle is older than main-process sources. Rebuilding…"* and rebuilt before launching, which is the `87a80fdbc40` / `b1c961522c5` path doing its job.

#### Verdict

Confirmed on real hardware: Electron does treat a `preventDefault()`-ed window close as a cancelled quit, and `461030400a0` re-issuing the quit from `window-all-closed` fixes it. All three pre-fix symptoms — resident process, refusing Dock icon, second-⌘Q escape — reproduced before the change and are gone after it.

Both halves of the ask are covered: the manual ⌘Q A/B settles the Electron-level mechanism, and the isolated suite going 6/6 confirms it end to end on the platform that was previously unmeasured.

#### Scope

The manual A/B was dev mode with a single window — the bug reproduces with one window by construction, since `window-all-closed` fires on the last window regardless of count. The two-window path is covered by the e2e suite rather than by hand. Run at default display scaling; the mixed-DPI matrix remains PT-4287's.

Issue #171, opened 2023-05-08
Create/update dev docs that explain the project folder structure


lyonsil:
This is probably done most naturally as markdown files inside the project itself, but it doesn't have to be.

PR #2640, opened 2026-08-05
Scope Simple-mode sync to the project the picker actually opens

tombogle:
# Code Review Summary

**Branch**: scope-sync-to-picker-decision

**Base**: origin/main

**Date**: 2026-08-28

**Files changed**: 6

## Overview

This branch scopes the automatic Send/Receive sync that runs when Simple mode starts up so it targets exactly the project that is actually becoming active, instead of syncing every project the user's account has access to. Previously, `src/main/startup-tasks.ts` unconditionally called `syncProjects(undefined)` on every Simple-mode cold start, which (via the companion, already-merged paratext-10-studio PR #171) triggered a server-side "modification status" check across the user's entire account. Now, `startup-tasks.ts` only fires that broad bootstrap sync in the true zero-state case (no project ever opened in Simple mode). Everywhere else, the Default Active Project Picker (`platform-scripture-editor.utils.ts`) fires a precisely-scoped `syncOnProjectSwitch(papi, projectId, undefined)` at the exact moment it resolves and opens a project — whether from the recents list (`tryOpenFromRecentlyOpened`) or the Send/Receive fallback candidate (`openDefaultActiveProjectIfApplicable`). A new shared helper, `getRecentlyOpenedProjectIds()`, gives `startup-tasks.ts` a process-agnostic way to check whether any project has ever been opened, reusing the same underlying data provider the picker already reads.

This PR's scoping only has full effect together with the private paratext-10-studio patch (PR #171, merged) that narrows the C# `GetSharedProjects`/`ResolveSharedProjects` implementation to actually honor a restricted project-id set — without that companion change, each individual sync call would still be expensive even when scoped to one project.

## API Changes

None. No files under `lib/platform-bible-react/`, `lib/platform-bible-utils/`, `papi.d.ts`, or any extension `src/types/*.d.ts` are touched by this diff. All six changed files are internal implementation; `getRecentlyOpenedProjectIds` is not re-exported from any `@papi/` module or public barrel, and the data provider it consumes (`platformScripture.recentlyOpenedProjects`) is pre-existing and unmodified.

None.

- [ ] ~~`src/shared/utils/recently-opened-project.util.ts:16-27` / `src/main/startup-tasks.ts:201-205`: The new zero-state gate has no bounded retry for the `platformScripture.recentlyOpenedProjects` data provider becoming available, unlike this same file's own `STARTUP_SYNC_RETRY_BUDGET_MS` mechanism (120s), which exists specifically because live E2E testing (2026-07-16) observed extension-host activation sometimes exceed the shared `networkService.request` retry ceiling (~9-10s, `MAX_REQUEST_ATTEMPTS` × `REQUEST_ATTEMPT_WAIT_TIME_MS`). Verified: `dataProviderService.get` → `networkObjectService.get` → `getRemoteNetworkObjectFunctions` → `networkService.request`, the exact same short-ceiling path. On a slow cold boot where the extension host isn't ready within ~9s, `getRecentlyOpenedProjectIds()` returns `[]` even for a returning user with real history, and the zero-state branch fires the old broad `syncProjects(undefined)`.~~ _(Author's explanation: confirmed as a real race, but not a regression — the fallback is exactly the pre-existing behavior, so the worst case is "PR's optimization doesn't apply this boot," never "broken." Deliberately not extending the `STARTUP_SYNC_RETRY_BUDGET_MS` pattern to this new gate in this PR, to keep the change minimal and reviewable; flagged below as a suggested fast-follow rather than a blocker.)_
- [ ] ~~`src/main/startup-tasks.ts:201-205`: Architectural consequence of the same gate — a returning user with recent-project history no longer has a session-long fallback broad sync if the Default Active Project Picker never manages to open any of them (e.g. every locally-known shared project is still `'new'`/`'unregistered'`, or Send/Receive never activates that session). Previously the unconditional startup sync could resolve exactly this kind of stuck state.~~ _(Author's explanation: known and intentional — this is the same edge case already tracked separately as PT-4307 ("picker finds nothing to load"), deliberately deferred out of this PR's scope during earlier design discussion.)_
- [ ] ~~`src/shared/utils/recently-opened-project.util.ts:16-27`: Duplicates the fetch idiom (`dataProviderService.get('platformScripture.recentlyOpenedProjects')` → `getRecentProjects(undefined)` → array-check → catch-and-log) already present in `getMostRecentUsableProjectId` at `src/renderer/services/web-view.service-shard.ts:1764-1771` (added in already-merged PR #2425).~~ _(Author's explanation: acknowledged minor duplication. Consolidating would mean modifying already-merged PR #2425 code that's outside this PR's stated scope (picker sync scoping); deferred as optional cleanup rather than done as a drive-by refactor here.)_

[Author response: all three Important findings were reviewed and explained by the author (with independent verification of the technical claims in finding 1) rather than fixed in this pass — see explanations above and Suggested Review Focus below.]

### Minor — Consider

None.

#### Shared Regions Modified

None.

#### Extension Config Changes

None.

## Positive Observations

- The gating logic in `startup-tasks.ts` is well-commented and correctly ordered relative to the existing mode/first-run/consent checks.
- `syncOnProjectSwitch` is reused as-is (not modified) at both new call sites in `platform-scripture-editor.utils.ts`, correctly gated behind a successful `openScriptureEditor`/recent-project-open before firing, matching the existing `main.ts` project-switch call site's fire-and-forget, error-isolated contract.
- Test coverage is thorough and directly verifies the new behavior: `startup-tasks.test.ts` covers both the zero-state (fires) and steady-state (skips) branches; `recently-opened-project.util.test.ts` covers the happy path, provider-unavailable, and provider-throws branches; `platform-scripture-editor.utils.test.ts` verifies the scoped `syncProjects` call fires with exactly the opened project ID from both the recents-path and S/R-fallback-path, including an updated call-count assertion with a clear explanatory comment.
- `src/shared/utils/recently-opened-project.util.ts` is named and organized consistently with sibling files in the same directory, and correctly uses `@shared/services/data-provider.service` (the established cross-process pattern) rather than crossing a process boundary improperly.
- No new UI-facing strings, no new components, and no build/config files were touched, so localization, Storybook, and build-config compliance are not applicable.
- No backward-facing dev-history comments (in-PR ticket refs, dated notes) were introduced.
- Prettier and ESLint both pass clean on all six changed/added files.

## Interview Notes

**Stated purpose**: Scope the automatic Simple-mode startup/project-switch Send/Receive sync to the specific project actually being opened, instead of running an account-wide sync — cutting Simple-mode cold-start time.

**Key design decisions**:
- The sync trigger point moves from a single unconditional call in `startup-tasks.ts` to the Default Active Project Picker itself, since the picker is the only place that actually knows, at the moment it opens a project, which single project is becoming active. `startup-tasks.ts` is narrowed to a pure zero-state bootstrap.
- This is deliberately distinct from `finalizeProjectSwitch` (added by already-merged PR #2425), which only fires from a live Power→Simple mode switch (`handleSwitchToSimpleMode` in `web-view.service-shard.ts`) during an already-running session — it never runs during cold boot, so it does not overlap with or obviate the two new picker-triggered `syncOnProjectSwitch` calls added here.
- This branch was rebased (force-pushed) onto latest `main` after PR #2425 merged; the rebase was a clean zero-conflict rebase with no code changes needed, confirming the two pieces of work are complementary rather than overlapping.
- The companion C#-side scoping fix (paratext-10-studio PR #171, merged) is required for this PR's project-id narrowing to actually reduce sync cost — that patch narrows `GetSharedProjects`/`ResolveSharedProjects` in `ParatextProjectSendReceiveService.cs` to honor a restricted project-id set before the expensive per-project modification-status check runs, not just filter results afterward.

**Findings resolution**: all three Important findings were explained rather than fixed in this pass; see the Findings section above. None were flagged as unresolved — the author (with independent code verification for finding 1) provided a specific technical rationale for each.

**Manual verification** (from earlier testing this session, prior to this specific rebase): `PT_STARTUP_MARKS=true` + `npm run startup-waterfall` on a packaged release build, cross-referenced against direct PAPI RPC calls and toast inspection, showed Simple-mode startup time improve from an initial ~55s baseline to under 29s, and confirmed unrelated Paratext-9-only projects no longer appear in the startup toast stream.

## Suggested Review Focus

- [ ] Finding 1 (slow-boot fallback race in `getRecentlyOpenedProjectIds`): agree the fallback-to-broad-sync behavior is acceptable to ship as-is (never worse than pre-PR), or require a bounded-retry fix (mirroring `STARTUP_SYNC_RETRY_BUDGET_MS`) before merge?
- [ ] Finding 2 (lost safety-net sync when the picker can't resolve any recent project): confirm PT-4307 fully covers this scope, or whether a distinct ticket is warranted.
- [ ] Finding 3 (duplicate recents-fetch idiom vs. `getMostRecentUsableProjectId`): optional follow-up consolidation, not blocking.
- [ ] Confirm the two-repo dependency (this PR + paratext-10-studio PR #171) is acceptable — this PR alone changes *which* project(s) get synced, but the per-sync cost reduction depends on the separate, already-merged private patch.

---

## Summary

Supersedes #2636 with a better-founded design (see discussion there for why).

- The Default Active Project Picker (`openDefaultActiveProjectIfApplicable` /
  `tryOpenFromRecentlyOpened`) now fires the existing `syncOnProjectSwitch` helper for exactly the
  project it resolves and opens — whether that's the top `recentlyOpenedProjects` entry or the
  Send/Receive-fallback candidate. This reuses the same fire-and-forget sync path already used on
  interactive project switches, so it inherits its existing behavior (and, combined with
  [paratext-10-studio#171](https://github.com/paranext/paratext-10-studio/pull/171)'s C#-side
  `GetSharedProjects` scoping, its speed) for free.
- `src/main/startup-tasks.ts`'s Simple-mode branch is narrowed to only run the broad,
  account-wide bootstrap sync in the genuine zero-state case (no project ever recorded in
  `platformScripture.recentlyOpenedProjects`). Whenever a recent project is already recorded, the
  picker owns firing a correctly-scoped sync once it knows which project it's actually opening —
  no more guessing at up to 5 candidate ids up front.
- Added `src/shared/utils/recently-opened-project.util.ts`, a small process-agnostic helper
  (`dataProviderService.get(...)` works the same from main or renderer) so `startup-tasks.ts`
  doesn't duplicate the recents lookup the picker already does.

## Why this design over #2636

The previous approach had `startup-tasks.ts` pass its own guess at up to 5 recent project ids into
`syncProjects`. That still risked syncing several projects instead of exactly the one about to be
opened, and had no reliable signal in the main process for which project is "the" active one. This
version instead defers to the one place that *does* know: the picker, at the exact moment it
resolves and opens a project.

## Test plan

- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm test` (core suite)
- [x] `cd extensions && npx vitest run src/platform-scripture-editor/src/platform-scripture-editor.utils.test.ts`
- [x] Manual: verified in a Paratext 10 Studio release build that startup no longer syncs
      unrelated Paratext-9-only projects and Simple-mode project switching still syncs correctly.

Follow-on (deliberately out of scope here, tracked separately): if every recorded recent project
turns out missing/inaccessible on disk, the picker currently has nothing to open and no automatic
recovery sync fires. Tracked as a lower-priority follow-up.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2640)

### DIFF
--- a/extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.test.ts
+++ b/extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.test.ts
@@ -1026,6 +1026,41 @@ describe('openDefaultActiveProjectIfApplicable', () => {
     expect(mockWarn).not.toHaveBeenCalled();
   });
 
+  it('fires a scoped sync for the S/R-fallback project it opens', async () => {
+    // Same rationale as the recents-path case: the picker knows exactly which project it just
+    // opened, so it (not some upstream guesser) is the right place to trigger a scoped sync.
+    const { papi, mockGetSetting, mockGetAllOpenWebViewDefinitions, mockSendCommand } =
+      createPickerMocks();
+    mockGetSetting.mockResolvedValue('simple');
+    mockGetAllOpenWebViewDefinitions.mockResolvedValue(
+      asWebViews([{ webViewType: SCRIPTURE_EDITOR_WEBVIEW_TYPE, projectId: undefined }]),
+    );
+    mockSendCommand.mockImplementation(async (commandName: string) => {
+      if (commandName === 'paratextBibleSendReceive.getSharedProjects') {
+        return {
+          proj1: {
+            id: 'proj1',
+            name: 'P1',
+            fullName: 'Project 1',
+            language: 'en',
+            editedStatus: 'edited',
+            lastSendReceiveDate: '2025-06-01T00:00:00Z',
+          },
+        };
+      }
+      if (commandName === 'platformScriptureEditor.openScriptureEditor') return 'opened-webview-id';
+      if (commandName === 'paratextBibleSendReceive.syncProjects') return undefined;
+      return undefined;
+    });
+
+    const outcome = await openDefaultActiveProjectIfApplicable(papi);
+
+    expect(outcome).toBe('filled');
+    expect(mockSendCommand).toHaveBeenCalledWith('paratextBibleSendReceive.syncProjects', [
+      'proj1',
+    ]);
+  });
+
   it('picks the editable project even when an Observer-only project has a newer lastSendReceiveDate', async () => {
     const {
       papi,
@@ -1717,6 +1752,7 @@ describe('openDefaultActiveProjectIfApplicable', () => {
     setRecentProjects(['proj-recent']);
     mockSendCommand.mockImplementation(async (commandName: string) => {
       if (commandName === 'platformScriptureEditor.openScriptureEditor') return undefined;
+      if (commandName === 'paratextBibleSendReceive.syncProjects') return undefined;
       throw new Error(`Unexpected command in test: ${commandName}`);
     });
 
@@ -1729,6 +1765,36 @@ describe('openDefaultActiveProjectIfApplicable', () => {
     );
     expect(mockSendCommand).not.toHaveBeenCalledWith('paratextBibleSendReceive.getSharedProjects');
     expect(mockRecordProjectOpened).toHaveBeenCalledWith('proj-recent');
+  });
+
+  it('fires a scoped sync for the project opened from recents', async () => {
+    // The picker knows, with certainty, exactly which project just became Simple mode's active
+    // project — so it should sync exactly that one, the same way syncOnProjectSwitch does for an
+    // interactive project switch, rather than a caller upstream guessing from a candidate list.
+    const {
+      papi,
+      mockGetSetting,
+      mockGetAllOpenWebViewDefinitions,
+      mockSendCommand,
+      setRecentProjects,
+    } = createPickerMocks();
+    mockGetSetting.mockResolvedValue('simple');
+    mockGetAllOpenWebViewDefinitions.mockResolvedValue(
+      asWebViews([{ webViewType: SCRIPTURE_EDITOR_WEBVIEW_TYPE, projectId: undefined }]),
+    );
+    setRecentProjects(['proj-recent']);
+    mockSendCommand.mockImplementation(async (commandName: string) => {
+      if (commandName === 'platformScriptureEditor.openScriptureEditor') return undefined;
+      if (commandName === 'paratextBibleSendReceive.syncProjects') return undefined;
+      throw new Error(`Unexpected command in test: ${commandName}`);
+    });
+
+    const outcome = await openDefaultActiveProjectIfApplicable(papi);
+
+    expect(outcome).toBe('filled');
+    expect(mockSendCommand).toHaveBeenCalledWith('paratextBibleSendReceive.syncProjects', [
+      'proj-recent',
+    ]);
   });
 
   it('tries each recent project in order and opens the first one that succeeds', async () => {
@@ -1750,13 +1816,16 @@ describe('openDefaultActiveProjectIfApplicable', () => {
         if (projectId === 'proj-gone') throw new Error('Project not found');
         return undefined;
       }
+      if (commandName === 'paratextBibleSendReceive.syncProjects') return undefined;
       throw new Error(`Unexpected command in test: ${commandName}`);
     });
 
     const outcome = await openDefaultActiveProjectIfApplicable(papi);
 
     expect(outcome).toBe('filled');
-    expect(mockSendCommand).toHaveBeenCalledTimes(2);
+    // Two openScriptureEditor attempts (proj-gone fails, proj-alive succeeds) plus one scoped sync
+    // for the project that actually opened.
+    expect(mockSendCommand).toHaveBeenCalledTimes(3);
     expect(mockSendCommand).toHaveBeenCalledWith(
       'platformScriptureEditor.openScriptureEditor',
       'proj-gone',
@@ -1862,6 +1931,7 @@ describe('openDefaultActiveProjectIfApplicable', () => {
     setRecentProjects(['proj-recent']);
     mockSendCommand.mockImplementation(async (commandName: string) => {
       if (commandName === 'platformScriptureEditor.openScriptureEditor') return undefined;
+      if (commandName === 'paratextBibleSendReceive.syncProjects') return undefined;
       throw new Error(`Unexpected command in test: ${commandName}`);
     });
     mockRecordProjectOpened.mockRejectedValue(new Error('Storage write failed'));
--- a/extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.test.ts
+++ b/extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.test.ts
@@ -1014,6 +1014,41 @@ describe('openDefaultActiveProjectIfApplicable', () => {
     expect(mockWarn).not.toHaveBeenCalled();
   });
 
+  it('fires a scoped sync for the S/R-fallback project it opens', async () => {
+    // Same rationale as the recents-path case: the picker knows exactly which project it just
+    // opened, so it (not some upstream guesser) is the right place to trigger a scoped sync.
+    const { papi, mockGetSetting, mockGetAllOpenWebViewDefinitions, mockSendCommand } =
+      createPickerMocks();
+    mockGetSetting.mockResolvedValue('simple');
+    mockGetAllOpenWebViewDefinitions.mockResolvedValue(
+      asWebViews([{ webViewType: SCRIPTURE_EDITOR_WEBVIEW_TYPE, projectId: undefined }]),
+    );
+    mockSendCommand.mockImplementation(async (commandName: string) => {
+      if (commandName === 'paratextBibleSendReceive.getSharedProjects') {
+        return {
+          proj1: {
+            id: 'proj1',
+            name: 'P1',
+            fullName: 'Project 1',
+            language: 'en',
+            editedStatus: 'edited',
+            lastSendReceiveDate: '2025-06-01T00:00:00Z',
+          },
+        };
+      }
+      if (commandName === 'platformScriptureEditor.openScriptureEditor') return 'opened-webview-id';
+      if (commandName === 'paratextBibleSendReceive.syncProjects') return undefined;
+      return undefined;
+    });
+
+    const outcome = await openDefaultActiveProjectIfApplicable(papi);
+
+    expect(outcome).toBe('filled');
+    expect(mockSendCommand).toHaveBeenCalledWith('paratextBibleSendReceive.syncProjects', [
+      'proj1',
+    ]);
+  });
+
   it('picks the editable project even when an Observer-only project has a newer lastSendReceiveDate', async () => {
     const {
       papi,
@@ -1705,6 +1740,7 @@ describe('openDefaultActiveProjectIfApplicable', () => {
     setRecentProjects(['proj-recent']);
     mockSendCommand.mockImplementation(async (commandName: string) => {
       if (commandName === 'platformScriptureEditor.openScriptureEditor') return undefined;
+      if (commandName === 'paratextBibleSendReceive.syncProjects') return undefined;
       throw new Error(`Unexpected command in test: ${commandName}`);
     });
 
@@ -1717,6 +1753,36 @@ describe('openDefaultActiveProjectIfApplicable', () => {
     );
     expect(mockSendCommand).not.toHaveBeenCalledWith('paratextBibleSendReceive.getSharedProjects');
     expect(mockRecordProjectOpened).toHaveBeenCalledWith('proj-recent');
+  });
+
+  it('fires a scoped sync for the project opened from recents', async () => {
+    // The picker knows, with certainty, exactly which project just became Simple mode's active
+    // project — so it should sync exactly that one, the same way syncOnProjectSwitch does for an
+    // interactive project switch, rather than a caller upstream guessing from a candidate list.
+    const {
+      papi,
+      mockGetSetting,
+      mockGetAllOpenWebViewDefinitions,
+      mockSendCommand,
+      setRecentProjects,
+    } = createPickerMocks();
+    mockGetSetting.mockResolvedValue('simple');
+    mockGetAllOpenWebViewDefinitions.mockResolvedValue(
+      asWebViews([{ webViewType: SCRIPTURE_EDITOR_WEBVIEW_TYPE, projectId: undefined }]),
+    );
+    setRecentProjects(['proj-recent']);
+    mockSendCommand.mockImplementation(async (commandName: string) => {
+      if (commandName === 'platformScriptureEditor.openScriptureEditor') return undefined;
+      if (commandName === 'paratextBibleSendReceive.syncProjects') return undefined;
+      throw new Error(`Unexpected command in test: ${commandName}`);
+    });
+
+    const outcome = await openDefaultActiveProjectIfApplicable(papi);
+
+    expect(outcome).toBe('filled');
+    expect(mockSendCommand).toHaveBeenCalledWith('paratextBibleSendReceive.syncProjects', [
+      'proj-recent',
+    ]);
   });
 
   it('tries each recent project in order and opens the first one that succeeds', async () => {
@@ -1738,13 +1804,16 @@ describe('openDefaultActiveProjectIfApplicable', () => {
         if (projectId === 'proj-gone') throw new Error('Project not found');
         return undefined;
       }
+      if (commandName === 'paratextBibleSendReceive.syncProjects') return undefined;
       throw new Error(`Unexpected command in test: ${commandName}`);
     });
 
     const outcome = await openDefaultActiveProjectIfApplicable(papi);
 
     expect(outcome).toBe('filled');
-    expect(mockSendCommand).toHaveBeenCalledTimes(2);
+    // Two openScriptureEditor attempts (proj-gone fails, proj-alive succeeds) plus one scoped sync
+    // for the project that actually opened.
+    expect(mockSendCommand).toHaveBeenCalledTimes(3);
     expect(mockSendCommand).toHaveBeenCalledWith(
       'platformScriptureEditor.openScriptureEditor',
       'proj-gone',
@@ -1850,6 +1919,7 @@ describe('openDefaultActiveProjectIfApplicable', () => {
     setRecentProjects(['proj-recent']);
     mockSendCommand.mockImplementation(async (commandName: string) => {
       if (commandName === 'platformScriptureEditor.openScriptureEditor') return undefined;
+      if (commandName === 'paratextBibleSendReceive.syncProjects') return undefined;
       throw new Error(`Unexpected command in test: ${commandName}`);
     });
     mockRecordProjectOpened.mockRejectedValue(new Error('Storage write failed'));
--- a/extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts
+++ b/extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts
@@ -677,6 +677,9 @@ async function tryOpenFromRecentlyOpened(
               `Default active project picker: failed to record recently-opened project ${projectId}: ${getErrorMessage(e)}`,
             );
           }
+          // Fire-and-forget, mirroring the project-switch call site in main.ts. No outgoing
+          // project — nothing was previously active in Simple mode this session.
+          syncOnProjectSwitch(papi, projectId, undefined);
           return 'filled' as const;
         } catch (e) {
           papi.logger.debug(
@@ -834,6 +837,9 @@ export async function openDefaultActiveProjectIfApplicable(
         `Default active project picker: failed to record recently-opened project ${top.id}: ${getErrorMessage(e)}`,
       );
     }
+    // Fire-and-forget, mirroring the project-switch call site in main.ts. No outgoing project —
+    // nothing was previously active in Simple mode this session.
+    syncOnProjectSwitch(papi, top.id, undefined);
   }
 
   return hasFailed ? 'failed' : 'filled';
--- a/extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts
+++ b/extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts
@@ -716,6 +716,9 @@ async function tryOpenFromRecentlyOpened(
               `Default active project picker: failed to record recently-opened project ${projectId}: ${getErrorMessage(e)}`,
             );
           }
+          // Fire-and-forget, mirroring the project-switch call site in main.ts. No outgoing
+          // project — nothing was previously active in Simple mode this session.
+          syncOnProjectSwitch(papi, projectId, undefined);
           return 'filled' as const;
         } catch (e) {
           papi.logger.debug(
@@ -875,6 +878,9 @@ export async function openDefaultActiveProjectIfApplicable(
         `Default active project picker: failed to record recently-opened project ${top.id}: ${getErrorMessage(e)}`,
       );
     }
+    // Fire-and-forget, mirroring the project-switch call site in main.ts. No outgoing project —
+    // nothing was previously active in Simple mode this session.
+    syncOnProjectSwitch(papi, top.id, undefined);
   }
 
   return hasFailed ? 'failed' : 'filled';
--- a/src/main/startup-tasks.test.ts
+++ b/src/main/startup-tasks.test.ts
@@ -10,6 +10,7 @@ import { settingsService } from '@shared/services/settings.service';
 import * as commandService from '@shared/services/command.service';
 import * as networkService from '@shared/services/network.service';
 import { logger } from '@shared/services/logger.service';
+import { getRecentlyOpenedProjectIds } from '@shared/utils/recently-opened-project.util';
 import { performStartupTasks, STARTUP_SYNC_RETRY_BUDGET_MS } from './startup-tasks';
 
 vi.mock('@shared/services/settings.service', () => ({
@@ -28,11 +29,16 @@ vi.mock('@shared/services/logger.service', () => ({
   logger: { debug: vi.fn(), info: vi.fn(), warn: vi.fn(), error: vi.fn() },
 }));
 
+vi.mock('@shared/utils/recently-opened-project.util', () => ({
+  getRecentlyOpenedProjectIds: vi.fn(),
+}));
+
 const mockSettingsGet = vi.mocked(settingsService.get);
 const mockSendCommand = vi.mocked(commandService.sendCommand);
 const mockRequestNoRetry = vi.mocked(networkService.requestNoRetry);
 const mockLoggerDebug = vi.mocked(logger.debug);
 const mockLoggerWarn = vi.mocked(logger.warn);
+const mockGetRecentlyOpenedProjectIds = vi.mocked(getRecentlyOpenedProjectIds);
 
 /**
  * Builds a rejection shaped like what `networkService`'s request plumbing actually throws for a
@@ -67,6 +73,10 @@ beforeEach(() => {
   vi.clearAllMocks();
   mockSendCommand.mockResolvedValue(undefined);
   mockRequestNoRetry.mockResolvedValue(undefined);
+  // Zero-state default (no recently-opened project) so every existing test's "fires the broad
+  // bootstrap sync" expectation stays valid unless a test explicitly overrides this to exercise
+  // the steady-state (picker handles it) path.
+  mockGetRecentlyOpenedProjectIds.mockResolvedValue([]);
 });
 
 describe('performStartupTasks', () => {
@@ -80,13 +90,22 @@ describe('performStartupTasks', () => {
     expect(mockSendCommand).not.toHaveBeenCalled();
   });
 
-  it('fires syncProjects with no project IDs when interface mode is simple', async () => {
+  it('fires the broad bootstrap sync when no project has been opened in simple mode yet (zero-state)', async () => {
     stubSettings({ mode: 'simple', firstRunComplete: true });
+    mockGetRecentlyOpenedProjectIds.mockResolvedValue([]);
     await performStartupTasks();
     expect(mockSendCommand).toHaveBeenCalledWith(
       'paratextBibleSendReceive.syncProjects',
       undefined,
     );
+    expect(mockRequestNoRetry).not.toHaveBeenCalled();
+  });
+
+  it('does NOT fire a startup sync when a recent project exists (the picker syncs it once it opens)', async () => {
+    stubSettings({ mode: 'simple', firstRunComplete: true });
+    mockGetRecentlyOpenedProjectIds.mockResolvedValue(['project-a']);
+    await performStartupTasks();
+    expect(mockSendCommand).not.toHaveBeenCalled();
     expect(mockRequestNoRetry).not.toHaveBeenCalled();
   });
 
--- a/src/main/startup-tasks.ts
+++ b/src/main/startup-tasks.ts
@@ -16,6 +16,7 @@ import {
 import { JSONRPCErrorCode } from 'json-rpc-2.0';
 import type { SettingTypes } from 'papi-shared-types';
 import { getErrorMessage, wait } from 'platform-bible-utils';
+import { getRecentlyOpenedProjectIds } from '@shared/utils/recently-opened-project.util';
 
 /**
  * How long (ms) to keep retrying `runScheduledSessionSync` while it's still unregistered before
@@ -106,10 +107,12 @@ type StartupSyncTriggerOutcome = ScheduledSessionSyncResult | 'skipped-stale' | 
  * Runs initialization tasks (currently: triggering an initial project sync) shortly after the app
  * finishes starting up.
  *
- * In Simple mode: requests a sync of all locally-known shared projects so the user sees the latest
- * content as soon as they open the app. All errors are swallowed — the S/R extension may not be
- * installed (e.g. Platform.Bible), the command may not yet be registered, or the sync may fail.
- * Startup must never be blocked or visibly affected by this.
+ * In Simple mode: only fires a bootstrap sync of every locally-known shared project when no project
+ * has been opened in Simple mode yet (zero-state) — otherwise this is a no-op, because the Default
+ * Active Project Picker syncs exactly the project it resolves and opens once it does. All errors
+ * are swallowed — the S/R extension may not be installed (e.g. Platform.Bible), the command may not
+ * yet be registered, or the sync may fail. Startup must never be blocked or visibly affected by
+ * this.
  *
  * In Power mode: requests a sync of just the projects scheduled "On startup/shutdown" via the S/R
  * extension's `runScheduledSessionSync` command. Same error-swallowing contract as Simple mode — if
@@ -191,12 +194,21 @@ async function performStartupTasksInternal(signals?: StartupTasksSignals): Promi
     return;
   }
 
-  // Simple mode: sync all locally-known shared projects (no project IDs = "sync all" per the
-  // C# `String[]? projectIds` contract). The C# S/R command registers asynchronously during
-  // startup; `sendCommand` will wait (with retry on missing handler) until it's available or
-  // times out. `undefined` as the single arg serializes as `null` in the JSON-RPC params array
-  // — matching the "sync all" sentinel on the C# side.
-  logger.debug('Startup sync starting');
+  // Simple mode: only bootstrap-sync here when no project has been opened yet (zero-state) — the
+  // Default Active Project Picker syncs exactly the project it resolves and opens once it does
+  // (see syncOnProjectSwitch calls in platform-scripture-editor.utils.ts), so guessing a scope
+  // here would either duplicate that sync or sync projects that aren't actually becoming active.
+  const recentProjectIds = await getRecentlyOpenedProjectIds();
+  if (recentProjectIds.length > 0) {
+    logger.debug('Startup sync skipped: a recent project exists; the picker will sync it');
+    return;
+  }
+
+  // Zero-state: no project IDs = "sync all" per the C# `String[]? projectIds` contract, so a
+  // brand-new user can discover a project they already have via Paratext 9. The C# S/R command
+  // registers asynchronously during startup; `sendCommand` will wait (with retry on missing
+  // handler) until it's available or times out.
+  logger.debug('Startup sync starting (zero-state bootstrap)');
   try {
     await commandService.sendCommand('paratextBibleSendReceive.syncProjects', undefined);
     logger.debug('Startup sync complete');
--- a/src/shared/utils/recently-opened-project.util.test.ts
+++ b/src/shared/utils/recently-opened-project.util.test.ts
@@ -1,0 +1,52 @@
+import { describe, expect, it, vi } from 'vitest';
+import { dataProviderService } from '@shared/services/data-provider.service';
+import { logger } from '@shared/services/logger.service';
+import { getRecentlyOpenedProjectIds } from '@shared/utils/recently-opened-project.util';
+
+vi.mock('@shared/services/data-provider.service', () => ({
+  dataProviderService: { get: vi.fn() },
+}));
+
+vi.mock('@shared/services/logger.service', () => ({
+  logger: { debug: vi.fn(), info: vi.fn(), warn: vi.fn(), error: vi.fn() },
+}));
+
+const mockGet = vi.mocked(dataProviderService.get);
+const mockLoggerWarn = vi.mocked(logger.warn);
+
+describe('getRecentlyOpenedProjectIds', () => {
+  it('returns the recents list from the data provider, most-recent first', async () => {
+    // The mock returns a minimal stand-in for the full data provider — only the method this
+    // helper calls needs to be present.
+    // eslint-disable-next-line no-type-assertion/no-type-assertion
+    mockGet.mockResolvedValue({
+      getRecentProjects: vi.fn().mockResolvedValue(['project-a', 'project-b']),
+    } as never);
+
+    const result = await getRecentlyOpenedProjectIds();
+
+    expect(result).toEqual(['project-a', 'project-b']);
+  });
+
+  it('returns an empty array when the data provider is unavailable', async () => {
+    mockGet.mockResolvedValue(undefined);
+
+    const result = await getRecentlyOpenedProjectIds();
+
+    expect(result).toEqual([]);
+  });
+
+  it('returns an empty array and logs a warning when the data provider read throws', async () => {
+    // A read failure must be distinguishable in logs from the genuine zero-state (no project ever
+    // opened) — both currently resolve to [], but callers treat that as "no active project yet",
+    // and a silently-swallowed failure would be indistinguishable from a real zero-state.
+    mockGet.mockRejectedValue(new Error('network object not registered'));
+
+    const result = await getRecentlyOpenedProjectIds();
+
+    expect(result).toEqual([]);
+    expect(mockLoggerWarn).toHaveBeenCalledWith(
+      expect.stringContaining('network object not registered'),
+    );
+  });
+});
--- a/src/shared/utils/recently-opened-project.util.ts
+++ b/src/shared/utils/recently-opened-project.util.ts
@@ -1,0 +1,28 @@
+import { dataProviderService } from '@shared/services/data-provider.service';
+import { logger } from '@shared/services/logger.service';
+import { getErrorMessage } from 'platform-bible-utils';
+
+/**
+ * Reads the ordered list of project ids the user has most recently opened as their main/active
+ * project in Simple mode (most-recent first; see `platformScripture.recentlyOpenedProjects`).
+ * Process-agnostic — `dataProviderService` resolves the same underlying data provider whether
+ * called from the renderer or the main process.
+ *
+ * @returns The recents list, or an empty array if the data provider is unavailable or the read
+ *   fails for any reason (e.g. not yet registered during cold boot). A read failure is logged
+ *   (unlike an unavailable provider, which is a routine early-startup state) so it stays
+ *   distinguishable from a genuine zero-state in logs.
+ */
+export async function getRecentlyOpenedProjectIds(): Promise<string[]> {
+  try {
+    const recentsProvider = await dataProviderService.get(
+      'platformScripture.recentlyOpenedProjects',
+    );
+    if (!recentsProvider) return [];
+    const recents = await recentsProvider.getRecentProjects(undefined);
+    return Array.isArray(recents) ? recents : [];
+  } catch (e) {
+    logger.warn(`Could not read recentlyOpenedProjects; treating as empty: ${getErrorMessage(e)}`);
+    return [];
+  }
+}


lyonsil (commented):
## Review: scoping Simple-mode startup sync

36 inline items below, grouped roughly as: the two new `syncOnProjectSwitch` calls and what they
duplicate (1-10), picker-internal issues the change surfaces (11-18), the new main-process gate
(19-24), the new shared util (25-29), and test falsifiability (30-36).

Method note, since it changed two of my conclusions: I verified the runtime behavior of concurrent
`syncProjects` against the PT10 Studio overlay, not just paranext-core. `RunWithSyncNotification`
rejects a concurrent sync with `FAILED_PRECONDITION` rather than queuing or deduplicating it. So the
duplicate is **not** a doubled sync or a doubled editing block - it is a call that is almost always
refused and logged as a warning. I also verified the two non-falsifiable tests by mutation rather
than by reading them.

Three further items (a pre-existing recents-cache bug this change increases exposure to, a TS/C#
contract divergence, and the missing ADR entry) plus my suggested changes are in a separate top-level
comment, since they are not anchorable to lines in this diff.

Nothing here is a blocker I would hold the PR on unilaterally - items 1-2 and 21 are the ones I
think are worth resolving before merge.

_(AI-assisted, with my guidance - applies to this review and all inline comments in it.)_

extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts
```diff
@@ -677,6 +677,9 @@ async function tryOpenFromRecentlyOpened(
               `Default active project picker: failed to record recently-opened project ${projectId}: ${getErrorMessage(e)}`,
             );
           }
+          // Fire-and-forget, mirroring the project-switch call site in main.ts. No outgoing
+          // project — nothing was previously active in Simple mode this session.
+          syncOnProjectSwitch(papi, projectId, undefined);
```
lyonsil, line 682: **1. Duplicate sync (recents path).**

This call is redundant. The picker only runs when an empty Scripture Editor exists, so
`allScriptureEditors.length > 0`, so `resolveOpenEditorDispatch` returns `replace-tab` (`:558`), so
`needsOverlay` is true (`main.ts:307`), so `main.ts:366` **already** fired
`syncOnProjectSwitch(papi, projectId, undefined)` inside the `openScriptureEditor` command awaited at
`:669`. The arguments match exactly: the tab being replaced is the empty editor, so
`outgoing?.projectId` is `undefined` at `main.ts:360-364`.

It is not a doubled sync, though. The PT10 Studio implementation of `syncProjects` is exclusive by
rejection - `RunWithSyncNotification` does `if (!_sendReceiveSemaphore.Wait(0))` and throws
`FAILED_PRECONDITION` ("A Send/Receive is already in progress"). `main.ts:366` fires first and still
holds the semaphore while `openOrUpdateRelatedPanels` / `applyForProject` / `openWebView` run, so
this call almost always loses the race and throws.

Net effect: a near-always-rejected call that logs a false warning (see item 17) on essentially every
Simple-mode startup, while the PR's headline mechanism is delivered entirely by pre-existing code.

extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts
```diff
@@ -834,6 +837,9 @@ export async function openDefaultActiveProjectIfApplicable(
         `Default active project picker: failed to record recently-opened project ${top.id}: ${getErrorMessage(e)}`,
       );
     }
+    // Fire-and-forget, mirroring the project-switch call site in main.ts. No outgoing project —
+    // nothing was previously active in Simple mode this session.
+    syncOnProjectSwitch(papi, top.id, undefined);
```
lyonsil, line 842: **2. Duplicate sync (S/R-fallback path).**

Same as item 1, via the `openScriptureEditor` awaited at `:818`. Same semaphore-rejection behavior,
same false warning.

extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts
```diff
@@ -677,6 +677,9 @@ async function tryOpenFromRecentlyOpened(
               `Default active project picker: failed to record recently-opened project ${projectId}: ${getErrorMessage(e)}`,
```
lyonsil, line 677: _Anchored here because GitHub only accepts inline comments on lines inside a diff hunk; the code this refers to is at **`:669`**._

**3. A resolved-but-`undefined` open is treated as success.**

`platformScriptureEditor.openScriptureEditor` is declared `Promise<string | undefined>`, where
`undefined` means no WebView was opened (`open()` does return the id). The `try` here only catches
rejection, so a resolved `undefined` falls straight through to `recordProjectOpened`, the new sync,
and `return 'filled'`.

Consequences: a bad id gets pinned at the head of recents (which `startup-tasks.ts:201` now trusts on
the next boot - see item 21), a sync fires with no editor on screen, and the picker reports success
for a still-empty editor.

extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts
```diff
@@ -834,6 +837,9 @@ export async function openDefaultActiveProjectIfApplicable(
         `Default active project picker: failed to record recently-opened project ${top.id}: ${getErrorMessage(e)}`,
```
lyonsil, line 837: _Anchored here because GitHub only accepts inline comments on lines inside a diff hunk; the code this refers to is at **`:818`**._

**4. Same unchecked return on the S/R-fallback path.**

`openedScriptureEditor = true` at `:819` on any non-throwing resolve, including a resolved
`undefined`. Same consequences as item 3.

extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts
lyonsil, line 682: **5. The dispatch-kind guard was dropped.**

`main.ts:307` deliberately syncs only when `interfaceMode === 'simple' && dispatch.kind ===
'replace-tab'`, and `main.ts:294-296` documents why `focus-existing` must not sync ("No project swap
occurs, so neither the transition overlay nor a sync is needed"). This call is unconditional.

At startup the stray sync is masked by the semaphore rejection, but on the mid-session re-fill paths
this driver documents at `:856-860` there is no competing sync - so it actually runs, raising an S/R
and the editing block for a project the user never switched to.

extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts
lyonsil, line 842: **6. Same missing dispatch-kind guard on the S/R-fallback path.** See item 5.

extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts
```diff
@@ -677,6 +677,9 @@ async function tryOpenFromRecentlyOpened(
               `Default active project picker: failed to record recently-opened project ${projectId}: ${getErrorMessage(e)}`,
             );
           }
+          // Fire-and-forget, mirroring the project-switch call site in main.ts. No outgoing
```
lyonsil, line 680: **7. `outgoingProjectId` is hardcoded `undefined` on a premise the code contradicts.**

The comment asserts "nothing was previously active in Simple mode this session," but
`startDefaultProjectPicker`'s own TSDoc at `:856-860` documents mid-session re-fills: the macOS
soft-close-then-reopen path re-creates the Scripture Editor without a `projectId` and re-triggers the
picker via `onDidUpdateWebView`.

On those paths a project *was* active, potentially with unpushed edits, and skipping the outgoing
`sendReceiveProjects` flush leaves them unsent.

extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts
```diff
@@ -834,6 +837,9 @@ export async function openDefaultActiveProjectIfApplicable(
         `Default active project picker: failed to record recently-opened project ${top.id}: ${getErrorMessage(e)}`,
       );
     }
+    // Fire-and-forget, mirroring the project-switch call site in main.ts. No outgoing project —
```
lyonsil, line 840: **8. Same false premise in the S/R-fallback copy of the comment.** See item 7.

extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts
lyonsil, line 682: **9. No freshness bound on the picker's sync.**

The picker is a long-lived subscription (`onDidOpenWebView` / `onDidUpdateWebView` /
`onSyncStateChanged`, `:854-863`), not a startup-only routine. The macOS re-fill path documented at
`:856-860` can fire this deep sync hours into a session with no apparent cause - and mid-session
there is no competing sync to reject it, so it runs.

This is exactly what `STARTUP_SYNC_FRESHNESS_WINDOW_MS` (`startup-tasks.ts:38-53`) exists to prevent
on the core side, and this path has no equivalent.

extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts
lyonsil, line 842: **10. Same missing freshness bound on the S/R-fallback path.** See item 9.

extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts
lyonsil, line 677: _Anchored here because GitHub only accepts inline comments on lines inside a diff hunk; the code this refers to is at **`:655`**._

**11. Main and the picker read the same provider at different times and can disagree.**

`startup-tasks.ts:201` reads `platformScripture.recentlyOpenedProjects` through
`networkService.request`'s retry policy (10 x 1000 ms) and may see `['proj-A']`, so it skips the
broad sync. This read runs inside `platformScriptureEditor.activate()` with no such retry, can get
`undefined`, and falls through to the S/R fallback - which opens and syncs `proj-B` (the
most-recently-S/R'd project).

Result: `proj-A` gets no sync at all, and no broad sync covers it either.

lyonsil, line 677: _Anchored here because GitHub only accepts inline comments on lines inside a diff hunk; the code this refers to is at **`:658`**._

**12. Two divergent readers of one provider.**

This site uses `?? []` and logs at `debug` (`:694-696`). The new
`recently-opened-project.util.ts:23,25` uses `Array.isArray(recents) ? recents : []` and logs at
`warn`. Neither uses the repo's canonical `isPlatformError` guard, which
`use-project-picker-data.hook.ts:185` applies to this same provider.

These two readers now independently decide whether a startup sync fires, so disagreement between
them is precisely the failure in items 11 and 21.

extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts
lyonsil, line 837: _Anchored here because GitHub only accepts inline comments on lines inside a diff hunk; the code this refers to is at **`:809`**._

**13. The Observer-only fallback makes the sync a silent no-op.**

`const topCandidate = editable[0] ?? observerOnly[0]` can select an Observer-only project, which the
backend then skips: `ParatextProjectSendReceiveService.cs:137-141` says "Non-editable and unknown IDs
are skipped."

Worth surfacing because the TS declaration core's callers actually read
(`src/@types/paratext-bible-send-receive/index.d.ts:372-375`) omits the editable filter entirely,
saying only "Unknown project IDs are skipped" - so nothing in TypeScript tells you this call does
nothing. See the top-level comment for the contract fix.

extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts
```diff
@@ -677,6 +677,9 @@ async function tryOpenFromRecentlyOpened(
               `Default active project picker: failed to record recently-opened project ${projectId}: ${getErrorMessage(e)}`,
             );
           }
+          // Fire-and-forget, mirroring the project-switch call site in main.ts. No outgoing
+          // project — nothing was previously active in Simple mode this session.
+          syncOnProjectSwitch(papi, projectId, undefined);
           return 'filled' as const;
```
lyonsil, line 683: **14. The record-then-sync block is copy-pasted verbatim.**

`:673-683` and `:829-843` are byte-identical, comment included.
`openDefaultActiveProjectIfApplicable` already funnels both paths to a single
`return hasFailed ? 'failed' : 'filled'` at `:845` - the one choke point where "filled implies
synced" could be enforced once instead of twice.

extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts
lyonsil, line 837: _Anchored here because GitHub only accepts inline comments on lines inside a diff hunk; the code this refers to is at **`:814-815`**._

**15. `hasFailed` and `openedScriptureEditor` are exact complements.**

One bit of state tracked in two mutable booleans that can never disagree: `hasFailed` is set only in
the `catch` that `openedScriptureEditor = true` is the `try`-path counterpart of. One of them can go.

extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts
```diff
@@ -834,6 +837,9 @@ export async function openDefaultActiveProjectIfApplicable(
         `Default active project picker: failed to record recently-opened project ${top.id}: ${getErrorMessage(e)}`,
       );
     }
+    // Fire-and-forget, mirroring the project-switch call site in main.ts. No outgoing project —
+    // nothing was previously active in Simple mode this session.
+    syncOnProjectSwitch(papi, top.id, undefined);
   }
 
   return hasFailed ? 'failed' : 'filled';
```
lyonsil, line 845: _Anchored here because GitHub only accepts inline comments on lines inside a diff hunk; the code this refers to is at **`:970-971`**._

**16. `syncOnProjectSwitch`'s TSDoc is now false.**

It states it is "Only called in simple mode after a replace-tab project switch." The two callers this
PR adds are startup/re-fill auto-opens, not project switches. If the calls stay, the doc needs to
describe both callers and the guarantees each does (and does not) provide.

lyonsil, line 845: _Anchored here because GitHub only accepts inline comments on lines inside a diff hunk; the code this refers to is at **`:987`**._

**17. The warn text misattributes startup-sync failures.**

`Project-switch sync: incoming sync for ${incomingProjectId} failed: ...` is emitted regardless of
caller. With the calls this PR adds, this is the line that will log
"A Send/Receive is already in progress" on essentially every Simple-mode startup (item 1), labelled
as a project switch that never happened. That is an actively misleading triage signal.

lyonsil, line 845: _Anchored here because GitHub only accepts inline comments on lines inside a diff hunk; the code this refers to is at **`:861-863`**._

**18. The documented retry trigger does not work as described.**

This TSDoc describes `paratextBibleSendReceive.onSyncStateChanged` as handling "a sync finishing"
generally. But `shared-layout-receiver.model.ts:171-172` and `sync-state.ts:33-36` state it fires only
for **manual** Send/Receives - the C#-served `syncProjects` this file calls emits nothing.

So the "re-check after the first sync settles" recovery for `'no-candidate'` does not fire on its own
sync. This matters because item 19's remaining permanent hole depends on that recovery existing.

src/main/startup-tasks.ts
```diff
@@ -191,12 +194,21 @@ async function performStartupTasksInternal(signals?: StartupTasksSignals): Promi
...
-  // startup; `sendCommand` will wait (with retry on missing handler) until it's available or
-  // times out. `undefined` as the single arg serializes as `null` in the JSON-RPC params array
-  // — matching the "sync all" sentinel on the C# side.
-  logger.debug('Startup sync starting');
+  // Simple mode: only bootstrap-sync here when no project has been opened yet (zero-state) — the
+  // Default Active Project Picker syncs exactly the project it resolves and opens once it does
+  // (see syncOnProjectSwitch calls in platform-scripture-editor.utils.ts), so guessing a scope
+  // here would either duplicate that sync or sync projects that aren't actually becoming active.
+  const recentProjectIds = await getRecentlyOpenedProjectIds();
+  if (recentProjectIds.length > 0) {
```
lyonsil, line 202: **19. The skip is unconditional but the compensating sync is conditional.**

This returns for every Simple-mode session with a non-empty recents list, on the premise that "the
picker will sync it." But the picker only syncs on its `'filled'` paths -
`'wrong-mode'`, `'no-empty'`, `'no-send-receive'`, `'no-candidate'` and `'failed'` (five of the six
`DefaultProjectPickerOutcome` values) fire no sync at all.

Most self-heal via the unfiltered `onDidOpenWebView` trigger. The genuinely permanent hole is "every
recent project is gone from disk **and** every shared project is `new`/`unregistered`" ->
`'no-candidate'`, with no data change to retry against, and (per item 18) no working sync-completion
retry either.

lyonsil, line 202: **20. Every project except the picker's one choice loses its startup sync.**

Pre-PR, the no-ID `syncProjects` covered all locally-present shared projects. Post-PR a second
project - a back-translation, or a second translation the user is not opening today - syncs neither
at startup (skipped here) nor at shutdown (`performSimpleModeShutdownSync` covers only the open
writable editor), so it can go many sessions without a sync.

If that is the intended trade, it is worth stating in the TSDoc at `:110-115`, which currently reads
as though nothing is lost.

src/main/startup-tasks.ts
```diff
@@ -191,12 +194,21 @@ async function performStartupTasksInternal(signals?: StartupTasksSignals): Promi
...
-  // C# `String[]? projectIds` contract). The C# S/R command registers asynchronously during
-  // startup; `sendCommand` will wait (with retry on missing handler) until it's available or
-  // times out. `undefined` as the single arg serializes as `null` in the JSON-RPC params array
-  // — matching the "sync all" sentinel on the C# side.
-  logger.debug('Startup sync starting');
+  // Simple mode: only bootstrap-sync here when no project has been opened yet (zero-state) — the
+  // Default Active Project Picker syncs exactly the project it resolves and opens once it does
+  // (see syncOnProjectSwitch calls in platform-scripture-editor.utils.ts), so guessing a scope
+  // here would either duplicate that sync or sync projects that aren't actually becoming active.
+  const recentProjectIds = await getRecentlyOpenedProjectIds();
```
lyonsil, line 201: **21. This gate probably never engages at cold boot, and fails toward the behavior the PR removes.**

`getRecentlyOpenedProjectIds()` -> `dataProviderService.get` -> `networkObjectService.get` ->
`networkService.request`, which retries `MAX_REQUEST_ATTEMPTS` (10) x `REQUEST_ATTEMPT_WAIT_TIME_MS`
(1000 ms) = a ~9-10 s ceiling (`rpc.model.ts:24,31`). Meanwhile `performStartupTasks` is fired at
`main.ts:254`, immediately after `extensionHostService.start()` merely **forks** the host
(`main.ts:234`), and the provider is registered at line 705 of `platform-scripture`'s 757-line
`activate()`.

If the provider is not up in time this reads `[]`, which is indistinguishable from a genuine
zero-state, and falls through to the broad sync-all at `:213` - while the picker also fires its own
scoped sync. So on slow machines the PR's benefit is not realized and startup sync is merely delayed
by up to ~10 s, with an extra sync collision.

**Caveat on my evidence:** the live-E2E datum in this file's own doc at `:23-27` (2026-07-16,
activation exceeding that exact ceiling) is about the **send-receive** extension, not
`platform-scripture`. It is suggestive of the same cold-boot race, not proof for this provider. Worth
measuring with `PT_STARTUP_MARKS` before settling on a gate policy.

lyonsil, line 201: **22. A multi-second network wait was added to a path that honors neither `abortSignal` nor the freshness window.**

The Power branch checks `signals?.abortSignal?.aborted` and `STARTUP_SYNC_FRESHNESS_WINDOW_MS`
precisely so a late trigger does not fire into a quitting app or raise the editing block on an editor
the user has already started typing in. The Simple branch checks neither, and this line roughly
doubles its latency.

Quit inside that window and this still issues sync-everything against a connection
`networkService.shutdown()` is tearing down - the exact case `abortSignal` was added for.

src/main/startup-tasks.ts
```diff
@@ -191,12 +194,21 @@ async function performStartupTasksInternal(signals?: StartupTasksSignals): Promi
     return;
   }
 
-  // Simple mode: sync all locally-known shared projects (no project IDs = "sync all" per the
-  // C# `String[]? projectIds` contract). The C# S/R command registers asynchronously during
-  // startup; `sendCommand` will wait (with retry on missing handler) until it's available or
-  // times out. `undefined` as the single arg serializes as `null` in the JSON-RPC params array
-  // — matching the "sync all" sentinel on the C# side.
-  logger.debug('Startup sync starting');
+  // Simple mode: only bootstrap-sync here when no project has been opened yet (zero-state) — the
```
lyonsil, line 197: _Anchored here because GitHub only accepts inline comments on lines inside a diff hunk; the code this refers to is at **`:186`**._

**23. Simple-mode startup sync now runs almost entirely outside the consent gate.**

`platform.syncOnStartup` has exactly one consumer repo-wide - `:186`, just above. There are zero hits
under `extensions/src/` or in either extension repo; same for `platform.firstRunComplete`.

The picker's sync calls were already ungated (via `main.ts:366`), but by returning at `:202` whenever
recents is non-empty, this PR makes the **gated** path unreachable for every returning Simple-mode
user. The consent gate now governs only the zero-state bootstrap, and a user who chose "Skip
automatic sync" still gets a sync from the picker.

src/main/startup-tasks.ts
```diff
@@ -191,12 +194,21 @@ async function performStartupTasksInternal(signals?: StartupTasksSignals): Promi
...
+  // Default Active Project Picker syncs exactly the project it resolves and opens once it does
+  // (see syncOnProjectSwitch calls in platform-scripture-editor.utils.ts), so guessing a scope
+  // here would either duplicate that sync or sync projects that aren't actually becoming active.
+  const recentProjectIds = await getRecentlyOpenedProjectIds();
+  if (recentProjectIds.length > 0) {
+    logger.debug('Startup sync skipped: a recent project exists; the picker will sync it');
+    return;
+  }
+
+  // Zero-state: no project IDs = "sync all" per the C# `String[]? projectIds` contract, so a
```
lyonsil, line 207: **24. The zero-state comment is misleading, and the rewrite dropped the one fact worth keeping.**

- "so a brand-new user can discover a project" names a user who cannot reach this line: a brand-new
  user has `firstRunComplete === false` and returns at `:178`. Their broad sync comes from
  `sync-consent-step.component.tsx:21`, which fires the identical call.
- "discover" is the wrong verb - the C# contract says the no-ID form syncs only projects "already
  present locally (i.e., not new)".
- The old comment carried a genuinely non-obvious fact that this rewrite dropped: `undefined`
  serializes as `null` in the JSON-RPC params array, which is what hits the C# sync-all sentinel.
  Worth restoring.

src/shared/utils/recently-opened-project.util.ts
```diff
@@ -0,0 +1,28 @@
...
+ *   fails for any reason (e.g. not yet registered during cold boot). A read failure is logged
+ *   (unlike an unavailable provider, which is a routine early-startup state) so it stays
+ *   distinguishable from a genuine zero-state in logs.
+ */
+export async function getRecentlyOpenedProjectIds(): Promise<string[]> {
+  try {
+    const recentsProvider = await dataProviderService.get(
+      'platformScripture.recentlyOpenedProjects',
+    );
+    if (!recentsProvider) return [];
```
lyonsil, line 21: **25. The dominant failure mode is the one that logs nothing.**

`if (!recentsProvider) return [];` is the branch a cold-boot / not-yet-registered provider actually
takes, and it is silent. That directly contradicts this function's own TSDoc at `:11-13`, which
claims a failure "is logged (unlike an unavailable provider, which is a routine early-startup state)
so it stays distinguishable from a genuine zero-state in logs."

As written, the case most likely to produce a wrong `[]` is precisely the case you cannot see in the
logs.

src/shared/utils/recently-opened-project.util.ts
```diff
@@ -0,0 +1,28 @@
...
+    const recents = await recentsProvider.getRecentProjects(undefined);
+    return Array.isArray(recents) ? recents : [];
+  } catch (e) {
+    logger.warn(`Could not read recentlyOpenedProjects; treating as empty: ${getErrorMessage(e)}`);
```
lyonsil, line 25: **26. Fail-open here inverts the discipline the caller's own file documents.**

Three distinct states - provider absent, read threw, genuine zero-state - all collapse to `[]`, and
`startup-tasks.ts:202` escalates `[]` to the single most destructive available action (sync every
locally-known shared project).

`startup-tasks.ts:139-143` explicitly refuses this exact reasoning for an unreadable
`interfaceMode`: "When we can't tell the mode, the safe default is to skip the automatic startup sync
this session and warn." Same uncertainty, same blast radius, opposite default.

src/shared/utils/recently-opened-project.util.ts
```diff
@@ -0,0 +1,28 @@
...
+ * called from the renderer or the main process.
+ *
+ * @returns The recents list, or an empty array if the data provider is unavailable or the read
+ *   fails for any reason (e.g. not yet registered during cold boot). A read failure is logged
+ *   (unlike an unavailable provider, which is a routine early-startup state) so it stays
+ *   distinguishable from a genuine zero-state in logs.
+ */
+export async function getRecentlyOpenedProjectIds(): Promise<string[]> {
+  try {
+    const recentsProvider = await dataProviderService.get(
```
lyonsil, line 18: **27. `dataProviderService.get` leaves permanent residue in main for a one-shot boolean.**

`get` registers a revocable proxy in `networkObjectRegistrations` and wires a
`<provider>:onDidUpdate` network-event subscription, neither of which is ever disposed here. The main
process will therefore receive and discard every recents update for the lifetime of the process,
just to answer "is this list empty?" once at boot.

`dataProviderService.hasKnown` answers the same question with neither the proxy nor the
subscription.

src/shared/utils/recently-opened-project.util.ts
```diff
@@ -0,0 +1,28 @@
...
+      'platformScripture.recentlyOpenedProjects',
```
lyonsil, line 19: **28. The provider-name literal is now hardcoded in a fifth place.**

Existing copies: `platform-bible-toolbar.tsx:180`, `use-project-picker-data.hook.ts:181`,
`platform-scripture-editor.utils.ts:655` and `:832`. The owning extension keeps its constant private
(`recently-opened-projects.service.ts:20` -
`RECENTLY_OPENED_PROJECTS_DATA_PROVIDER_NAME`, not exported).

The repo's precedent for mirroring an extension identifier into core is
`SCRIPTURE_EDITOR_WEBVIEW_TYPE`, which is documented as the single core-side copy. Exporting the
constant and importing it would be better than adding a fifth string.

src/shared/utils/recently-opened-project.util.ts
```diff
@@ -0,0 +1,28 @@
+import { dataProviderService } from '@shared/services/data-provider.service';
```
lyonsil, line 1: **29. This file is in the wrong bucket.**

`.claude/rules/architecture/shared-patterns.md` reserves `src/shared/` for "Code for ALL processes."
The only non-test importer is `src/main/startup-tasks.ts:19`, and the two pre-existing renderer
readers of this same provider (`platform-bible-toolbar.tsx:180`,
`use-project-picker-data.hook.ts:181`) were not migrated to it.

`src/node/` or `src/main/` matches the actual sharing. If the intent is that the renderer readers
migrate here later, that is worth saying in the TSDoc - otherwise this reads as `shared/` by default
rather than by design.

extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.test.ts
```diff
@@ -1014,6 +1014,41 @@ describe('openDefaultActiveProjectIfApplicable', () => {
...
+      }
+      if (commandName === 'platformScriptureEditor.openScriptureEditor') return 'opened-webview-id';
+      if (commandName === 'paratextBibleSendReceive.syncProjects') return undefined;
+      return undefined;
+    });
+
+    const outcome = await openDefaultActiveProjectIfApplicable(papi);
+
+    expect(outcome).toBe('filled');
+    expect(mockSendCommand).toHaveBeenCalledWith('paratextBibleSendReceive.syncProjects', [
```
lyonsil, line 1047: **30. Mutation-proven: this assertion cannot catch a duplicate sync.**

`toHaveBeenCalledWith` matches any call in the mock's history, so it passes whether the sync fires
once or five times. I verified it: duplicating the `syncOnProjectSwitch` call at
`platform-scripture-editor.utils.ts:842` leaves **97/97 tests green**.

Given that the central risk in this change is "the picker syncs a project that was already synced
upstream," a `toHaveBeenCalledTimes` assertion here would be doing the actual work.

extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.test.ts
```diff
@@ -1719,6 +1755,36 @@ describe('openDefaultActiveProjectIfApplicable', () => {
...
+    mockSendCommand.mockImplementation(async (commandName: string) => {
+      if (commandName === 'platformScriptureEditor.openScriptureEditor') return undefined;
+      if (commandName === 'paratextBibleSendReceive.syncProjects') return undefined;
+      throw new Error(`Unexpected command in test: ${commandName}`);
+    });
+
+    const outcome = await openDefaultActiveProjectIfApplicable(papi);
+
+    expect(outcome).toBe('filled');
+    expect(mockSendCommand).toHaveBeenCalledWith('paratextBibleSendReceive.syncProjects', [
```
lyonsil, line 1783: **31. Same non-falsifiable assertion on the recents-path test.** See item 30.

extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.test.ts
```diff
@@ -1738,13 +1804,16 @@ describe('openDefaultActiveProjectIfApplicable', () => {
...
       throw new Error(`Unexpected command in test: ${commandName}`);
     });
 
     const outcome = await openDefaultActiveProjectIfApplicable(papi);
 
     expect(outcome).toBe('filled');
-    expect(mockSendCommand).toHaveBeenCalledTimes(2);
+    // Two openScriptureEditor attempts (proj-gone fails, proj-alive succeeds) plus one scoped sync
+    // for the project that actually opened.
+    expect(mockSendCommand).toHaveBeenCalledTimes(3);
```
lyonsil, line 1816: **32. This pre-existing count assertion is the only duplicate-detecting test in the file.**

Duplicating the sync at *both* call sites produces exactly one failure across the whole suite, and it
is this line - the count the PR itself had to bump from 2 to 3 - not either of the two new tests.

That is a useful signal about where the real coverage is: the tests added for this behavior have none
of it.

src/shared/utils/recently-opened-project.util.test.ts
```diff
@@ -0,0 +1,52 @@
...
+      getRecentProjects: vi.fn().mockResolvedValue(['project-a', 'project-b']),
+    } as never);
+
+    const result = await getRecentlyOpenedProjectIds();
+
+    expect(result).toEqual(['project-a', 'project-b']);
+  });
+
+  it('returns an empty array when the data provider is unavailable', async () => {
+    mockGet.mockResolvedValue(undefined);
```
lyonsil, line 32: **33. Mutation-proven non-falsifiable.**

I deleted `if (!recentsProvider) return [];` from
`recently-opened-project.util.ts:21` and this file still passes **3/3**. The resulting TypeError is
swallowed by the outer `catch`, which returns `[]` anyway - so the test cannot tell the guard from its
absence.

`expect(mockLoggerWarn).not.toHaveBeenCalled()` would fix it, and would simultaneously pin the
logging contract that item 25 says is currently violated.

src/shared/utils/recently-opened-project.util.test.ts
```diff
@@ -0,0 +1,52 @@
...
+}));
+
+vi.mock('@shared/services/logger.service', () => ({
+  logger: { debug: vi.fn(), info: vi.fn(), warn: vi.fn(), error: vi.fn() },
+}));
+
+const mockGet = vi.mocked(dataProviderService.get);
+const mockLoggerWarn = vi.mocked(logger.warn);
+
+describe('getRecentlyOpenedProjectIds', () => {
```
lyonsil, line 17: **34. No `beforeEach` / `vi.clearAllMocks()`.**

`vitest.config.ts` sets neither `clearMocks` nor `mockReset`, so mock state accumulates across cases
in this file. Verified empirically: `mockGet` reaches 3 recorded calls by the last test.

Harmless today, but it blocks the exact assertion that would fix item 33 - any
`not.toHaveBeenCalled()` check would be reading state left over from an earlier case.

src/main/startup-tasks.test.ts
```diff
@@ -90,6 +101,14 @@ describe('performStartupTasks', () => {
     expect(mockRequestNoRetry).not.toHaveBeenCalled();
   });
 
+  it('does NOT fire a startup sync when a recent project exists (the picker syncs it once it opens)', async () => {
```
lyonsil, line 104: **35. This test passes for the wrong reason.**

It asserts that no sync fired, but never asserts *why*. Make `getRecentlyOpenedProjectIds` reject
instead of resolving `['project-a']` and it still passes - the rejection is caught by
`performStartupTasks`'s outer try/catch and no command is sent either way.

Every sibling skip test in this file pins the reason string (e.g. `:146-148`). Asserting
`mockLoggerDebug` was called with "Startup sync skipped: a recent project exists" would make this one
falsifiable, and would distinguish "gate engaged" from "everything blew up quietly" - which is
exactly the failure mode item 21 describes.

src/main/startup-tasks.test.ts
```diff
@@ -28,11 +29,16 @@ vi.mock('@shared/services/logger.service', () => ({
   logger: { debug: vi.fn(), info: vi.fn(), warn: vi.fn(), error: vi.fn() },
 }));
 
+vi.mock('@shared/utils/recently-opened-project.util', () => ({
```
lyonsil, line 32: **36. Mocks an internal collaborator rather than the boundary.**

`.claude/rules/testing/tdd-discipline.md`: "Mock only at external boundaries, not internal
collaborators." `getRecentlyOpenedProjectIds` is a same-repo helper introduced by this PR; the actual
boundary is `dataProviderService`. As written, renaming the provider or changing the getter's arity
leaves every test in this file green.

This is also the fifth module-level `vi.mock` in the file, past Testing-Guide.md's "more than 3 mocks
- reconsider the design" threshold.

lyonsil:
## Three items that can't be anchored inline, plus suggested changes

Companion to the [inline review](https://github.com/paranext/paranext-core/pull/2640#pullrequestreview-4866440584) (items 1-36). These three live outside this PR's diff.

### 37. `extensions/src/platform-scripture/src/recently-opened-projects.service.ts:93` - the cache this PR now reads earliest can be poisoned, and the repair overwrites real data

`loadCache()` opens with `if (this.cache) return this.cache;`. `[]` is truthy, so an empty result is memoized with no invalidation path, and `readRawRecentlyOpenedProjects` swallows every storage failure into `'[]'`.

This is **pre-existing, not introduced here** - the picker was already the first caller. What this PR changes is exposure: `startup-tasks.ts:201` makes the main process the earliest caller, at the point in boot where the extension host's storage layer is least warm.

The consequence is worse than a stale read. `recordProjectOpened` (`:80-90`) reads the poisoned `[]`, computes `[projectId, ...[]]`, and writes it - **overwriting the user's real on-disk recents list**. One racy read at boot can silently truncate recents to a single entry.

Worth its own issue regardless of what happens to this PR. Two candidate fixes: sentinel the cache on `undefined` rather than truthiness, and don't let a read failure populate the cache at all.

### 38. `src/@types/paratext-bible-send-receive/index.d.ts:372-375` - the TS contract omits a filter the C# applies

`ParatextProjectSendReceiveService.cs:137-141`: *"filters for editable projects and S/Rs them ... **Non-editable** and unknown IDs are skipped."*

`index.d.ts:372-375` - the declaration core's callers actually read - says only *"Unknown project IDs are skipped."*

So a caller passing an Observer-only project id gets a silent no-op with nothing in TypeScript to warn them. That is exactly the path inline item 13 describes (`platform-scripture-editor.utils.ts:809` can select an Observer-only project as `topCandidate`). Bringing the `.d.ts` wording in line with the C# is the cheap half of that fix.

### 39. `.context/standards/Architecture-Decisions.md` - no entry for this decision

Repo `CLAUDE.md`: *"When any code work surfaces a significant architecture decision - choosing among viable approaches, introducing a new pattern or top-level structure ... or deciding where a feature lives - record it."*

This PR moves ownership of Simple-mode sync scope from core to an extension, and introduces a new core-reads-extension-data-provider coupling. That is squarely in scope. The log still ends at ADR-0006.

---

## Suggested changes

**1. Delete both new `syncOnProjectSwitch` calls (`:682`, `:842`).** Highest value, lowest risk - `openScriptureEditor` already fires the correctly-scoped sync, with the dispatch-kind guard the picker lacks. Collapses inline items 1, 2, 5, 6, 7, 8, 9, 10, 14, 16, 17.

Pair it with two small fixes so the paths where `main.ts` deliberately suppresses the sync are covered on purpose rather than by accident:

- Guard `main.ts:283`'s `platform.interfaceMode` read with `isPlatformError` before passing it to `resolveOpenEditorDispatch`. Today a failed read silently drops `needsOverlay`, and the sync with it.
- Check `openScriptureEditor`'s return value at `:669` and `:818` (items 3, 4). A resolved `undefined` means no editor opened - don't record recents, don't sync, don't report `'filled'`. This is also what makes it safe to stop syncing on the `focus-existing` path.

**2. Hold the main-side gate (`startup-tasks.ts:201-205`) until it's measured.** Capture how long `platform-scripture.activate()` actually takes to reach its provider registration (`main.ts:705` of 757) at cold boot on a slow target, with `PT_STARTUP_MARKS`. That number decides whether this gate is an optimization or a coin flip, and it can't be read off the code.

One thing I'd steer away from: `paratextBibleSendReceive.syncOpenProjects` looks like the natural primitive here, but it isn't usable at this call site. `performStartupTasks` is fired at `main.ts:254`, right after the extension-host fork and **before `createWindow`** - there are no open web views yet, so it would resolve to `syncProjects([])`, an explicit no-op. Using it would require inventing a layout-settled signal core doesn't currently have.

**3. If the gate stays, three changes make it honest.** Return tri-state (`string[] | 'unknown'`) instead of collapsing three distinct states onto `[]`; use `dataProviderService.hasKnown` so main stops holding a proxy and a live subscription to answer a boolean (item 27); and log the `!recentsProvider` branch (item 25).

Then decide explicitly what `'unknown'` maps to, and write it down. Note that mapping it to *skip* may make the Simple-mode startup sync effectively dead code on slow machines - which may well be fine (the zero-state bootstrap is also fired by `sync-consent-step.component.tsx:21`, and `firstRunComplete` already gates this path), but it should be a deliberate deletion rather than a side effect.

**4. Make the four non-falsifiable tests falsifiable** (items 30-36): a `toHaveBeenCalledTimes` assertion on each new sync test; `expect(mockLoggerWarn).not.toHaveBeenCalled()` plus a `beforeEach(vi.clearAllMocks)` in the util test; a reason-string assertion on the new gate test; and mock `dataProviderService` rather than `getRecentlyOpenedProjectIds`.

**5. Relocate `recently-opened-project.util.ts`** to `src/node/` or `src/main/`, and export `RECENTLY_OPENED_PROJECTS_DATA_PROVIDER_NAME` from the owning extension so the fifth hardcoded copy of that string becomes an import (items 28, 29).

---

Method note on two conclusions I revised while checking my own work: I verified the runtime behavior of concurrent `syncProjects` against the PT10 Studio overlay rather than paranext-core alone. `RunWithSyncNotification` rejects a concurrent sync with `FAILED_PRECONDITION` instead of queuing or deduplicating it - so the duplicate is a call that gets refused and logged, **not** a doubled sync and **not** a doubled editing block. I'd previously described it as both. I also verified the non-falsifiable tests by mutating the source and re-running, not by reading them.

tombogle:
Closing this rather than continuing to iterate on it — writing up why for the record, and because it settles a lot of the review above.

## tl;dr

Matt's review turned out to be substantively right, and following it all the way through changed what this PR should even be. Rather than fix in place, closing now and re-opening a much smaller PR once #2700 lands.

## What we found

The PR's headline mechanism — two new `syncOnProjectSwitch` calls in the picker (`platform-scripture-editor.utils.ts:721,883`) — is redundant with pre-existing code (Matt's items 1/2). Tracing it directly: the picker only calls `openScriptureEditor` when it already found an empty editor slot, which means `resolveOpenEditorDispatch` always resolves to `replace-tab`, which means `needsOverlay` is `true`, which means `main.ts`'s existing `open()` already fires `syncOnProjectSwitch(papi, projectId, undefined)` — the identical call, before the picker's own call even runs. Items 5–10, 14, 16, 17 follow from that same redundancy.

More importantly: PT-4483's actual DoD item 1 ("fires exactly one scoped sync, not account-wide") is *already* satisfied without any of this PR's new picker-side code — it's satisfied by that same pre-existing `main.ts` call, combined with the already-merged C# scoping fix in [paratext-10-studio#171](https://github.com/paranext/paratext-10-studio/pull/171), which is what actually made the account-wide `GetSharedProjects`/`ResolveSharedProjects` enumeration honor a scoped id (previously it silently ran account-wide regardless of what was requested — that's the real ~55s→29s startup fix). This PR was solving a problem the C# patch had already solved from the other end.

Separately: item 24 turned out to be based on stale docs, not stale code. `syncProjects`'s `.d.ts` TSDoc doesn't mention the `isFirstTimeSync` special case in the actual P10 Studio override (`SyncProjectsCore`) — when every locally-known shared project is still `new`, the no-ID call *does* sync everything, including the new ones. The current zero-state comment in `startup-tasks.ts` is correct; the contract docs (both the C# stub's XML comment and the TS `.d.ts`) are what's out of date. Worth fixing regardless of what happens to the rest of this PR.

And item 23 (consent bypass) is real but pre-existing — not introduced by this PR — and #2700 (PT-4369) is already fixing it at the actual source (`syncOnProjectSwitch` itself, plus a shared `first-run-consent.util.ts` gate), which is a better fix location than anything scoped to this PR.

## Why close instead of fix in place

The only genuinely new, still-needed piece of this PR is the `startup-tasks.ts` zero-state narrowing (skip the broad bootstrap sync once a project has ever been recorded) plus the `recently-opened-project.util.ts` helper it needs — both real, but both need real fixes (silent-failure/fail-open behavior, a leaked data-provider subscription, a 5th hardcoded provider-name literal, wrong directory per `src/shared/` vs `src/node/` conventions, and a couple of non-falsifiable tests).

#2700 ("PT-4369: Gate automatic sync on first-run consent"), open now and reviewed, directly modifies both files this trimmed scope would touch — `syncOnProjectSwitch` in `platform-scripture-editor.utils.ts` and the gating structure in `startup-tasks.ts` (via a new shared `first-run-consent.util.ts`). Landing changes here first would hand #2700 an avoidable conflict in exactly its two focus files.

## Next steps

- PT-4483 stays open; will re-scope once #2700 merges and reopen against the new baseline. Likely a small diff: the zero-state gate + the util's bug fixes + the C#/TS doc corrections for `syncProjects`.
- Matt's review remains the design reference for that follow-up PR — items 25–29 (the util), 33/34 (its tests), and the `isFirstTimeSync` doc gap are still live action items; most of the rest (1–2, 5–10, 14, 16–18, 30–32, 36) are resolved by this closure, since the code they're about won't exist in the new PR.

Thanks for the extremely thorough review — it's a big part of why we caught this before it shipped.

PR #2644, opened 2026-08-05
PT-4305: Re-show setup wizard when registration becomes invalid

katherinejensen00:
# Code Review Summary

**Branch**: pt-4305-registration-recheck-on-startup

**Base**: origin/main

**Date**: 2026-08-05

**Files changed**: 15 (14 in scope; `package-lock.json` is out of scope — its only change is the top-level `"name"` field, a benign npm-in-worktree artifact)

## Overview

Re-shows the first-run wizard's registration step at startup for an already-onboarded (Simple-mode)
user whose Paratext registration has become invalid (expired, revoked, or the registration file
removed/corrupted). A completed user still reaches the app immediately (`setStatus({ kind: 'app' })`),
and a **non-blocking, fired-not-awaited** background re-check (`startBackgroundRegistrationRecheck`
in `first-run-store.ts`) then resolves registration validity via the existing
`resolveRegistrationValidity()` helper. Only a **definitive `'invalid'`** re-raises the wizard at the
`identify` step; a down/slow backend resolves to `'unknown'` and is ignored, so an outage never
re-onboards an established user. The re-check runs at most once per startup and swallows its own
errors, so it can never block or crash startup.

Users can opt out via a new core boolean setting `platform.showRegistrationReminderOnStartup`
(default `true`), toggled either from the Settings UI or from a "Don't show this on startup again"
checkbox in the re-register view. The re-register view also adds a "Continue without registration"
escape hatch and an explanatory notice, and reuses existing `platform-bible-react` components. The
pure `decideFirstRun` reducer and the fresh-user onboarding flow are unchanged.

This branch was also hardened by a prior multi-agent `/code-review` pass before this review (see
Interview Notes).

## API Changes

- `@papi/core` (`SettingTypes` in `papi-shared-types.ts`, regenerated into `papi.d.ts`): **added**
  `platform.showRegistrationReminderOnStartup: boolean` — a new user setting (default `true`) with a
  `booleanValidator` entry, a `platformSettings` contribution, and matching `_label`/`_description`
  localization keys. This is the only public API-surface change. Purely additive — no exports
  removed, no signatures narrowed. (The new optional `allowContinueWithoutRegistration` prop lives on
  internal `src/renderer/` first-run components, not public API surface.)

None.

- [x] **Settings label used first person ("Remind me to re-register at startup")**, contrary to the Vocabulary guideline (use second person, not "my"/"me") and inconsistent with sibling settings (e.g. "Perform automatic sync on startup"). _(fixed during review: changed `%settings_platform_showRegistrationReminderOnStartup_label%` to "Show re-registration reminder at startup")_

[Author response: Agreed it was a genuine improvement; fixed in-review.]

### Minor — Consider

- [x] **`consumeJustRegisteredFlag()` ran after the `reminderSuppressed` early-return** in `startBackgroundRegistrationRecheck`, so a suppressed launch left `JUST_REGISTERED_KEY` stale — a latent edge case where a much-later re-enable could swallow one legitimate wizard raise. _(fixed during review: moved the consume above the suppression check so the flag is consumed exactly once per startup; added a test `consumes the just-registered flag even when the reminder is suppressed`)_
- [x] **Re-register "no longer valid" notice rendered as bold body text, not an `Alert`** — a genuine problem state deserving appropriate visual weight and screen-reader announcement (the file already uses `Alert` for the valid-registration confirmation). _(fixed during review: notice now renders via `Alert`/`AlertTitle` with an `AlertCircle` icon)_
- [x] **Suppression checkbox could diverge from the persisted setting on a failed write** — the optimistic local `suppressReminder` state stayed set even if `settingsService.set` threw (caught + logged). _(fixed during review: `onToggleSuppressReminder` now reverts the local state on write failure; added a test `reverts the suppression checkbox when the settings write fails`)_
- [ ] ~~`platform.showRegistrationReminderOnStartup` is user-visible (not `isHidden`), unlike its adjacent siblings `platform.firstRunComplete` / `platform.syncOnStartup`~~ _(Intended: a deliberate design decision so a user who suppressed the reminder has a Settings-UI path to re-enable it. Unlike the machine-managed sibling settings, this is a genuine user preference.)_
- [ ] ~~The registration-code `placeholder="XXXXXX-XXXXXX-XXXXXX-XXXXXX-XXXXXX"` is a hardcoded, non-localized user-visible string~~ _(Pre-existing and unchanged by this PR; it is a Latin-letter format mask, not translatable prose — out of scope for this change.)_

[Author response: The two open-then-dismissed minors are intentional/pre-existing as noted; the other three minors were fixed in-review.]

#### Shared Regions Modified

None. (No `#region shared with` markers appear in any changed file.)

#### Extension Config Changes

None. (No files under `extensions/` were changed.)

## Positive Observations

- Reuses the existing `resolveRegistrationValidity()` tri-state helper rather than re-implementing registration probing — no duplication of validity logic.
- `consumeJustRegisteredFlag()` cleanly DRYs the previously-inlined read-then-clear pair, shared between the mutually-exclusive fresh-user and completed-user startup paths (no double-consume).
- Correctness matches intent: non-blocking (`startBackgroundRegistrationRecheck` fired without `await`), once-per-startup module guard (reset in `resetFirstRunStore`), `'unknown'` treated as "stay in app" (outage-safe), explicit-`false` suppression with a show-by-default fallback, no stubs/TODOs/dead code.
- Correct apply-model separation: the immediate-apply "Don't show again" checkbox is isolated from the explicit-apply "Save and restart" form by a divider (with a code comment citing the rationale); the restart-triggering registration change correctly uses explicit apply.
- `backButton` written as an `if/else` with a comment explaining it avoids `no-nested-ternary`, rather than suppressing the lint rule.
- New setting wired end-to-end and consistently (type + validator + contribution + label/description), following the sibling-setting pattern; `papi.d.ts` regenerated (not hand-edited); no dangling `%...%` references.
- Thorough test coverage: the re-check across all branches (invalid→wizard, valid/unknown→stay, suppression, fail-open on read error, undefined-default, the one-launch `justRegistered` guard including next-launch re-raise, power-mode skip, error swallowing); re-register UI (escape hatch, checkbox true/false toggling, notice gating, write-failure revert); and the new setting's default/visibility/validator.
- Storybook coverage: a new `ReRegisterMode` story (with a `play` assertion) covers the new component variant.
- Button microcopy names outcomes ("Save and restart", "Continue without registration"), sentence case throughout, escape hatch appropriately de-emphasized (`ghost` variant), proper `…` ellipsis characters.
- No breaking changes; the `cross-view-sync-hidden-views` rule does not apply (data-driven state machine, no geometry/scroll side effects).

## Interview Notes

- **Stated purpose**: Re-show the setup wizard's registration step at startup for an already-onboarded (Simple-mode) user whose Paratext registration has become invalid, without regressing startup latency or re-onboarding an established user on a transient backend hiccup, and with a way to suppress the reminder that is respected (PT-4305). The design was pre-approved in the ticket.
- **Deliberate decisions the author made** (beyond the original design):
  - The `justRegistered` guard in the completed-user re-check is an intentional addition: `IdentifyStep` sets `JUST_REGISTERED_KEY` on every save, but the completed-user branch returns before the fresh-user code that consumes it, so without the guard a transient `'invalid'` immediately after a re-register would re-nag the user. This review further hardened it (consume once per startup regardless of suppression).
  - The new setting is intentionally **visible** (not `isHidden`) so a user who suppressed the reminder can re-enable it from Settings.
  - The gate intentionally appears over active work after the (up to ~15s) validity probe; this is mitigated by the "Continue without registration" escape hatch and the suppression checkbox.
- **Prior hardening (before this review pass)**: a multi-agent `/code-review` had already un-exported the internal re-check, DRY'd `consumeJustRegisteredFlag`, removed a double-negative, added the `ReRegisterMode` story, added the invalid-registration notice + checkbox separation, and substantially hardened the test suite (including fixing a mock-isolation issue).
- **In-review changes**: all four findings the author chose to act on were fixed during this review (see Findings). The author confirmed they wanted every genuine improvement applied.
- **Wrap-up**: Author had nothing additional to flag.
- **Understanding**: The author demonstrated clear understanding of the design and its tradeoffs throughout; no areas were deferred to AI or left unexplained.

## In-Review Quality Check

Four fixes + two verifying tests were applied during the review, then the quality gate was run:

- **typecheck**: clean — only the pre-existing, unrelated `src/main/services/app.service-host.ts` → `buildInfo.json` (TS2307) error remains (a missing build artifact in the worktree, present before this branch).
- **lint (ESLint)**: clean.
- **format (Prettier, changed files)**: clean — "All matched files use Prettier code style!"
- **full test suite**: 1411/1418 passed. The 7 failures are in **unrelated** files (`network.service.test.ts`, `dialog.service-host.test.ts`, `scroll-group.service-host.test.ts`, and two others) and are **pre-existing timing flakes** — all sampled failing files pass in isolation (35/35). They are not caused by this branch. This branch's own tests (first-run-store, identify-step, first-run-shell, core-settings-info) all pass.

## Suggested Review Focus

- [ ] Confirm the settings-UI exposure of `platform.showRegistrationReminderOnStartup` (visible, not hidden) matches product intent — it diverges deliberately from the hidden sibling first-run settings.
- [ ] Sanity-check the startup-latency claim: the background re-check is fired-not-awaited after `showApp`, so completed users are never blocked; confirm this holds in the packaged app.
- [ ] Review the UX of the gate re-appearing over active work after the validity probe (up to ~15s), and whether the escape hatch + suppression checkbox are sufficient mitigation.
- [ ] Note the 7 pre-existing flaky test failures in unrelated services are an existing repo condition (they pass in isolation), not introduced here — worth a separate ticket if not already tracked.

AI-assisted — Claude Code (Opus 4.8, VS Code extension; no session URL available)

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2644)


captaincrazybro (approved):
> @captaincrazybro reviewed 14 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2644)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2645, opened 2026-08-05
Add chapter-number button for blank chapters in Scripture editor

captaincrazybro:
# Code Review Summary

**Branch**: t-a-chapter-number-reinstatement

**Base**: origin/main

**Date**: 2026-08-05

**Files changed**: 5 (plus one pre-existing, unrelated uncommitted file — see Interview Notes)

## Overview

Adds an "Add chapter number" button and empty-state view to the Scripture editor for genuinely blank chapters (no `\c` marker, no verses) in Simple mode. When the current chapter is blank, the editor canvas is replaced by a message ("This chapter is empty.") and a button that, on click, builds and inserts a `\c` + `\v 1..N` scaffold — matching the chapter's real versification — via the editor's `applyUpdate` API. Power mode is unaffected (unchanged current behavior). This branch also includes a fix wave addressing findings from an earlier whole-branch review: a re-entrancy guard against double-clicking the button (which previously could write the scaffold to USFM twice), dedicated tooltip copy for the disabled state, tighter scoping of a refocus effect, and gating the button on whether real versification data exists for the current chapter.

A companion fix in the sibling `eten-tech-foundation/scripture-editors` repo (not part of this diff) corrects a related `\c` marker-menu bug and a real undo-history regression; that work is tracked and reviewed separately since it's a different git repository.

Screenshot of populate chapter button:
image

## API Changes

None. All changed files are extension-internal (`extensions/src/platform-scripture-editor/`); nothing in `lib/platform-bible-react/`, `lib/platform-bible-utils/`, `@papi/*`, or this extension's own `.d.ts` type-declaration module changed. The new exported helpers (`isChapterBlank`, `buildChapterScaffoldOps`, `canAddChapterNumber`, `resolveAddChapterNumberClick`) and `EmptyChapterView` are internal implementation details, not part of any public API surface.

None.

- [x] **Button label "Add Chapter Number" is Title Case, violating the sentence-case guideline** — inconsistent with every other action label in the same localization file ("Insert comment", "Lock structure", etc.). _(fixed during review: changed to "Add chapter number" in `contributions/localizedStrings.json`)_
- [x] **`handleAddChapterNumber` had no defense-in-depth check against `isStructureProtected`** — relied solely on the button's `disabled` prop, unlike the paragraph and inline marker menus in the same file, which both re-check protection state inside their action callback and call `notifyStructureProtected()` as a documented second layer. _(fixed during review: added the same check to `handleAddChapterNumber`)_
- [ ] **Refocus effect doesn't account for the dock tab being hidden/inactive** — if a user clicks "Add chapter number" and switches to a different tab before the ~300ms insert round trip completes, `editorRef.current?.focus()` fires on an invisible pane. Per `.claude/rules/cross-view-sync-hidden-views.md`, this is a silent no-op in practice (Electron/browsers don't focus elements in a `display:none` tree), so the likely user-visible effect is just "cursor isn't positioned as expected if you come back to this tab later" — not a crash or data issue. The established pattern elsewhere in this codebase (`useViewVisibility` from `platform-bible-react`) isn't used here. Left open for the reviewer to decide whether this is worth hardening now or as a follow-up.
- [ ] **Missing Storybook story and component test for `EmptyChapterView`** — the sibling `structure-protection-button.component.tsx` (a very similar disabled-button-with-tooltip pattern) ships both. This is the same "no test for Task 2/3" tradeoff that was explicitly decided during this feature's planning (see Interview Notes) — left open here since the reviewer may weigh it differently than the plan did.
- [ ] **Duplicated ~12-line accessibility wrapper (disabled-button-with-tooltip pattern)** — `EmptyChapterView` is now the second copy of this exact boilerplate in this extension (the first is the Paragraph Selection button in `platform-scripture-editor.web-view.tsx`). No shared helper exists for it in either the extension or `platform-bible-react`. Left open — worth a small extraction, but not urgent.
- [ ] **No responsive wrap/stack fallback for the message+button row** — the Responsiveness guideline requires no horizontal scrollbar down to a 300px-wide WebView; the current row (`tw:flex tw:items-center tw:gap-2`, no wrap) could overflow at that width. Left open.

_Author response: the two fixed findings were clear-cut style/pattern-consistency issues, fixed directly. The four remaining open findings are all real but lower-urgency (a narrow edge-case UX gap, two "nice to have" test/reuse investments, and a responsive-design polish item) — recorded for the human reviewer to triage rather than auto-fixed, given the scope this branch has already grown to across its implementation and two prior fix-review cycles._

### Minor — Consider

- [x] **The uncommitted "KNOWN CAVEAT" comment about undo reliability ships the feature anyway** — this is a known, already-tracked risk (see Interview Notes), not new information, but flagged again here since two independent analysis passes both surfaced it.

_Author response: on further reflection, this isn't actually a meaningful risk for this specific action. Populating an already-blank chapter with the `\c`/`\v` scaffold has no real content to lose if undo doesn't work — there's no legitimate scenario where a user needs to *undo* this rather than simply deleting the inserted chapter/verse markers again, which is no worse a recovery path than undo would have been. Not pursuing a tracked ticket for it; the code comment at `handleAddChapterNumber` documents the mechanism (and its two root causes, one fixed upstream, one open) for whoever eventually bumps the vendored package version. See Interview Notes for the full investigation history._

#### Shared Regions Modified

None.

#### Extension Config Changes

None.

## Positive Observations

- The new pure helper functions (`isChapterBlank`, `buildChapterScaffoldOps`, `canAddChapterNumber`, `resolveAddChapterNumberClick`) are well-isolated, thoroughly unit-tested with real edge-case coverage (e.g. a chapter with a `\c` marker but no verses is correctly *not* treated as blank, to avoid a duplicate `\c` insert), and documented with TSDoc explaining *why*, not just *what*.
- The re-entrancy guard (`isInsertInFlightRef`) and the scoped refocus guard (`didInsertScaffoldRef`) are both correct and clearly commented — genuinely tricky async UI state handled with care, distinguishing "this click's insert landed" from any other cause of the same state transition (navigation, remote update, Send/Receive sync).
- `getEndVerse` returning `0` (chapter 0, or no versification entry) is treated as a real, reachable case throughout — both the gating logic and its tests reflect this, not just a defensive afterthought.
- Consistent `!isPowerMode` guarding keeps this Simple-mode-only feature from touching Power mode anywhere.
- `EmptyChapterView` follows this extension's established presentational-component conventions exactly (props-only, frozen localization-key array + type, `localize()` fallback) — matches `structure-protection-button.component.tsx` and `share-layout-button.component.tsx`.
- All new user-visible strings are correctly routed through localization with no hardcoded text.
- The *other* hidden-view concern (`Editorial` staying `tw:hidden` while the chapter is blank) is explicitly called out and dismissed with a comment citing the exact project rule file — exactly the kind of decidable-from-the-diff documentation that rule asks for.
- Button is hidden entirely (not shown-disabled-with-no-explanation) when versification data isn't loaded or the project is read-only.
- No ellipsis on the button label — correct, since the click completes the action immediately with no further choice required.

## Interview Notes

This review was run autonomously (the author/session owner stepped away to manually test the running app while this analysis ran), so the "interview" was self-conducted using the full context already established earlier in this session: a written design spec, an implementation plan executed task-by-task with per-task review, a discovered-and-fixed undo-history bug in a sibling repo, a full whole-branch review, and a complete fix-review cycle addressing that review's findings.

- **Stated purpose**: matches the PR description above.
- **The uncommitted `dock-layout-wrapper.component.tsx` change is NOT part of this branch's work.** It predates this session's changes (confirmed: it was already modified, uncommitted, before any of this branch's 4 commits were made) and its purpose is unknown to this review — it appears to memoize `getGroups(isPowerMode)` in the dock layout, which is plausible as an unrelated in-progress fix, but has not been explained or attributed. **This should not be swept into this PR's commit** — flagged explicitly rather than silently touched, per this session's standing instruction to leave it alone.
- **The undo-reliability caveat is a known, already-investigated gap, not new information for this review.** During implementation, manual end-to-end testing found that Undo does not yet reliably restore the inserted scaffold. Two rounds of empirical investigation (in the sibling `scripture-editors` repo) found and fixed one real cause (a `viewOptions` memoization bug) and identified — but did not chase down — a second, different cause, most likely `paranext-core`'s own `use-editor-pdp-sync.hook.ts` calling `EditorRef.setUsj()` on the PDP-echo success path, which also wipes Lexical's undo history. On a later review pass, the person directing this session concluded this gap doesn't warrant a tracked ticket after all: populating an already-blank chapter has no real content at stake, so there's no legitimate case for needing to undo it specifically — the recovery path (delete the inserted markers) is trivially available regardless of whether undo works. The code comment documenting the mechanism stays for whoever bumps the vendored package version later, but this is not treated as an open risk requiring follow-up tracking.
- **The vendored-package version gap is real and separately tracked.** `extensions/src/platform-scripture-editor/package.json` still pins `~0.8.14`; the sibling repo's fixes (the `\c` marker fix and the `viewOptions` fix) are committed there but not yet published to a real npm version, so neither fix ships in a normal `npm install` outside this development machine's manually-patched `.yalc` copy. This is intentionally deferred — a real version bump/publish is a separate, later action owned by whoever manages that repo's releases, not something this branch or this review should force.
- **No areas of the implementation were deferred to an AI without understanding** — every design decision in this feature (the empty-state UI placement, the `applyUpdate`-over-`setUsj` choice to avoid wiping annotations, the blank-detection rule requiring neither a chapter node nor verses, the Simple-mode-only scoping) traces to an explicit decision made and reasoned through during this session's planning and review cycles, documented in this session's own implementation ledger.

## In-Review Quality Check

Two findings were fixed during this review (sentence-case button label; structure-protection defense-in-depth check). Quality checks on the amended files:

- **Lint**: one Prettier formatting warning on the dependency-array reformatting in `platform-scripture-editor.web-view.tsx`; auto-fixed with `prettier --write`. Clean after.
- **Format**: clean after the fix above.
- **Test**: `platform-scripture-editor` workspace — 581/581 passing (39 files), no regressions from either fix.
- **Typecheck**: not run separately (per this session's established practice of treating lint + build/test as sufficient verification for changes of this scope).

Both fixes were committed as `6c2baa8e7ce` ("fix: sentence-case button label and structure-protection defense-in-depth"), on top of the branch's existing 4 commits. Not pushed — this branch is being kept as-is per an explicit decision, pending further manual testing.

## Suggested Review Focus

- [ ] **The unrelated, uncommitted `dock-layout-wrapper.component.tsx` change** — confirm its origin and intent before this PR is finalized; it should likely be split out into its own commit/PR with its own explanation, or explicitly excluded from this one.
- [x] **The undo-reliability gap** — not treated as blocking: populating an already-blank chapter has nothing to lose if undo doesn't work, so there's no real scenario where a user needs to undo this specific action rather than just deleting the markers again. The code comment at `handleAddChapterNumber` documents the mechanism for context; no tracked ticket is planned.
- [ ] **The hidden-dock-tab refocus edge case** — decide whether to harden with `useViewVisibility` now or accept the documented no-op behavior.
- [ ] **Test/story coverage for `EmptyChapterView`** — this extension's own convention (see `structure-protection-button.component.tsx`) suggests both; this branch's planning explicitly chose to skip them for this component, but the reviewer may want to weigh in on whether that tradeoff still holds now that the component has grown a defense-in-depth structure-protection path.

---

**Note:** this branch depends on a companion fix in the sibling `eten-tech-foundation/scripture-editors` repo (see that repo's PR) for the \`c\` marker-menu fix. That repo's PR also carries a real undo-history fix, not yet published to a real npm version — `extensions/src/platform-scripture-editor/package.json` still pins `~0.8.14`. Undo for the newly-inserted chapter/verse scaffold is not fully reliable until that dependency is bumped, but this is not treated as a blocker: populating an already-blank chapter has no real content to lose, so there is no scenario where undoing this specific action matters — the recovery path (delete the inserted markers) is available either way.

**Rebase note (2026-08-07):** rebased onto current `main`. Upstream PR #2648 (already merged) independently built the same disabled-button-plus-tooltip abstraction this branch introduced (`DisabledActionTooltip`) — as an extension-local `DisabledTooltipWrapper`, consumed by the new character-marker control, with its own docstring flagging the paragraph switcher (one of `DisabledActionTooltip`'s three call sites) as its intended next adopter. Rather than shipping both, consolidated: moved `DisabledTooltipWrapper` into `platform-bible-react` (it now has cross-extension consumers), refactored `DisabledActionTooltip` to compose it internally, and updated the character-marker control's import — deleting the extension-local duplicate.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2645)


katherinejensen00 (changes requested):
Thanks for your good work on this, Levi! Just a few things to consider. 

> @katherinejensen00 reviewed 5 files and all commit messages, and made 8 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2645#-:-OzMi4Oy4hYFAxDXNCxh:b-92nptq)* status: all files reviewed, 7 unresolved discussions (waiting on captaincrazybro).

___
*[`extensions/src/platform-scripture-editor/src/empty-chapter-view.component.tsx` line 63 at r1](https://reviewable.io/reviews/paranext/paranext-core/2645#-OzMi0RuAl6K9H2e6K-5:-OzMi0RuAl6K9H2e6K-6:b8x49jh) ([raw file](https://github.com/paranext/paranext-core/blob/6c2baa8e7ce3e18447003c6ef2c71a9a80bc703e/extensions/src/platform-scripture-editor/src/empty-chapter-view.component.tsx#L63)):*
> ```XML
>         <span>{localize(localizedStrings, MESSAGE_KEY)}</span>
>         {showButton && (
>           <TooltipProvider>
> ```

This disabled-button-inside-a-focusable-wrapper + tooltip pattern (with its `jsx-a11y/no-noninteractive-tabindex` suppression) is the same idiom as the Paragraph-Selection button in the web view and two spots in `find.component.tsx` — four copies now, each carrying its own lint disable. Worth pulling into a small shared `platform-bible-react` component so an a11y fix lands once. The pattern itself is correct here (a disabled button can't host its own tooltip, so the focusable wrapper is the right call) — it's just duplicated.
___
*[`extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts` line 1079 at r1](https://reviewable.io/reviews/paranext/paranext-core/2645#-OzMhpda2ZFotA_PU5cS:-OzMhpda2ZFotA_PU5cT:b-e9i2tg) ([raw file](https://github.com/paranext/paranext-core/blob/6c2baa8e7ce3e18447003c6ef2c71a9a80bc703e/extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts#L1079)):*
> ```XML
>  * @returns The Delta operations, in insertion order.
>  */
> export function buildChapterScaffoldOps(chapterNum: number, lastVerseNum: number): DeltaOp[] {
> ```

Two things on the generated scaffold:

1. No enclosing paragraph. The ops are `\c` + bare `\v \v …` with no `para` op, but real chapter USJ nests verses inside a `<para style="p">`. `DeltaOp.insert` is loosely typed (`Record<string, unknown>`), so nothing flags this at compile time. Could you check what actually gets saved? If `applyUpdate` doesn't auto-wrap, the USFM comes out as `\c 1 \v 1 \v 2 …` with no `\p`, which Paratext and downstream tools won't expect. The unit test asserts only the op-array shape, not that the editor accepts it or that the saved USFM is well-formed.

2. Versification fidelity. The linear `1..lastVerse` loop assumes contiguous verses; for versifications with excluded or segmented verses it'll emit verses that shouldn't exist. The C# side already scaffolds chapters from versification (`ScriptureTemplateService.GetCVs`, which walks `NextVerse`) and already has an "effectively empty chapter" check (`UsfmChapterScaffolding.HasContentBeyondScaffolding`). A lower-fidelity reimplementation in the renderer means the two can drift. Re-homing is bigger than this PR, but worth a conscious call — going through a PDP/command peer of the create-book path would also make the insert idempotent (no re-entrancy guard needed) and keep undo in one place.
___
*[`extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts` line 1096 at r1](https://reviewable.io/reviews/paranext/paranext-core/2645#-OzMhuVWEhf9Bf0B4_gz:-OzMhuVWEhf9Bf0B4_h-:bmeg041) ([raw file](https://github.com/paranext/paranext-core/blob/6c2baa8e7ce3e18447003c6ef2c71a9a80bc703e/extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts#L1096)):*
> ```XML
>  * @returns `true` if the button should be enabled, `false` otherwise.
>  */
> export function canAddChapterNumber(lastVerse: number): boolean {
> ```

Minor: `canAddChapterNumber` is just `lastVerse > 0`, and `resolveAddChapterNumberClick` is a three-outcome union that its single caller immediately re-expands into three branches. If the isolated unit tests are the reason they're extracted, no problem keeping them; otherwise they read as a bit more indirection than one call site needs.
___
*[`extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx` line 1301 at r1](https://reviewable.io/reviews/paranext/paranext-core/2645#-OzMgcTm86c1reWvu8qN:-OzMgcTm86c1reWvu8qO:b-zef45q) ([raw file](https://github.com/paranext/paranext-core/blob/6c2baa8e7ce3e18447003c6ef2c71a9a80bc703e/extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx#L1301)):*
> ```XML
>   }, [saveUsjToPdpRaw]);
> 
>   const isBlankChapter = useMemo(() => isChapterBlank(usjFromPdp ?? defaultUsj), [usjFromPdp]);
> ```

`useData` doesn't reset its value to the default when the selector changes — it keeps the previous chapter's USJ until the new subscription resolves, and the spinner guard further down only catches the first-load case (`usjFromPdp === defaultUsj` by reference), not navigation. So during a chapter change `isBlankChapter` is computed against stale content: navigating from a blank chapter into a populated one briefly shows "This chapter is empty." + the button over the loading editor. Because the button is live in that window, a click can drive `applyUpdate` into a chapter that's resolving to real content — a path to a duplicate `\c`. Can we gate the empty view on a fresher signal (e.g. also require the subscription to be settled for the current `scrRef`)?

Nit: `usjFromPdp ?? defaultUsj` — `usjFromPdp` is typed non-optional and the hook already defaults to `defaultUsj`, so it's never `undefined` here; the `?? defaultUsj` is dead.
___
*[`extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx` line 1303 at r1](https://reviewable.io/reviews/paranext/paranext-core/2645#-OzMhE9l3ULW-Os_Ql6V:-OzMhE9l3ULW-Os_Ql6W:bpa82f3) ([raw file](https://github.com/paranext/paranext-core/blob/6c2baa8e7ce3e18447003c6ef2c71a9a80bc703e/extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx#L1303)):*
> ```XML
>   const isBlankChapter = useMemo(() => isChapterBlank(usjFromPdp ?? defaultUsj), [usjFromPdp]);
> 
>   const lastVerse = useMemo(
> ```

Cross-book timing bug: `lastVersesInCurrentBook` comes from `usePromise`, which preserves the previous value while refetching, but `currentBookNum` (which `getEndVerse` guards on) updates synchronously with `scrRef.book`. So right after navigating to a new book, `getEndVerse(newBook, ch)` passes its book-match guard while the array still holds the *old* book's counts — returning the previous book's verse count. If someone scaffolds a blank chapter in that window, `buildChapterScaffoldOps` emits the wrong number of `\v` markers. The `lastVersesInCurrentBook !== undefined` check in `showButton` doesn't catch it, since the preserved value is defined — just stale.
___
*[`extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx` line 1343 at r1](https://reviewable.io/reviews/paranext/paranext-core/2645#-OzMhJxi8NBCEXko8rUM:-OzMhJxi8NBCEXko8rUN:bagjxqh) ([raw file](https://github.com/paranext/paranext-core/blob/6c2baa8e7ce3e18447003c6ef2c71a9a80bc703e/extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx#L1343)):*
> ```XML
>     isInsertInFlightRef.current = true;
>     didInsertScaffoldRef.current = true;
>     // KNOWN CAVEAT: undo reliability for this insert depends on unreleased fixes in the vendored
> ```

Shipping with undo known-unreliable is a real gap for a data-mutating action — the apply-model guideline wants a recovery path for changes that aren't self-reversing. It traces back to routing the insert through the editor's edit-sync + undo history rather than a data-layer write (see my note on `buildChapterScaffoldOps`). Could we get a tracked follow-up referenced for the dependency bump, so the fix isn't carried only in a code comment?
___
*[`extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx` line 1358 at r1](https://reviewable.io/reviews/paranext/paranext-core/2645#-OzMg08sAyAw3J7saHXa:-OzMg08sAyAw3J7saHXb:blvtv7a) ([raw file](https://github.com/paranext/paranext-core/blob/6c2baa8e7ce3e18447003c6ef2c71a9a80bc703e/extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx#L1358)):*
> ```XML
>   // Simple-mode only, so the button can't render in Power mode today — the guard is currently
>   // redundant but documents that invariant and protects against a future refactor.
>   useEffect(() => {
> ```

These two guard refs (`isInsertInFlightRef`, `didInsertScaffoldRef`) are component-wide but only reset when `isBlankChapter` flips to `false`, so they leak across chapters:

- **Stuck button (blank→blank):** click Add on a blank chapter, then navigate to another blank chapter within the insert round trip. `isBlankChapter` stays `true`, this effect early-returns, and the ref never resets — so on the new chapter every click resolves to `already-in-flight` and the button silently does nothing until a non-blank chapter is visited. A freshly created book of empty chapters is exactly this scenario.
- **Stuck button (rejected/reverted save):** if the write is rejected (sync write-gate revert, permissions), the chapter stays blank, the guard never resets, and the button is dead for that chapter until the tab is reopened. There's no timeout/failure reset.
- **Focus steal:** navigate to a *populated* chapter mid-flight and this effect fires `editorRef.current?.focus()` on that chapter, pulling focus off wherever the user put it during navigation.

All three come from the refs not being scoped to `scrRef`. Could we reset both in a `useEffect` keyed on `[scrRef.book, scrRef.chapterNum]` (and on save failure)? That resolves all three together. A test that clicks, navigates blank→blank, and asserts the button still works would lock it in.

Two smaller things while here: the refs are always set and reset together, so a single ref could do both jobs; and this effect re-triggers `focus()` but not `scrollToVerse`, so a blank chapter opened at verse > 1 lands scrolled to the top after un-hiding (the scroll effects don't re-fire because `scrRef` is unchanged).

extensions/src/platform-scripture-editor/src/empty-chapter-view.component.tsx
```diff
@@ -0,0 +1,69 @@
...
+  isStructureProtected,
+  showButton,
+  onAddChapterNumber,
+}: EmptyChapterViewProps) {
+  return (
+    <div className="tw:flex tw:items-start tw:justify-start tw:h-full tw:px-4 tw:pt-4">
+      <div className="tw:flex tw:items-center tw:gap-2">
+        <span>{localize(localizedStrings, MESSAGE_KEY)}</span>
+        {showButton && (
+          <DisabledActionTooltip
```
captaincrazybro, line 57: **Fixed.** Extracted this disabled-button/wrapper/tooltip idiom (and its two other copies — the Paragraph Selection button in `platform-scripture-editor.web-view.tsx`, and the ReplaceAll/Replace buttons in `find.component.tsx`) into a shared `DisabledActionTooltip` component in `platform-bible-react`. All three sites now delegate to it, so the a11y wrapper logic and its lint suppression live in one place.

extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx
```diff
@@ -1296,6 +1307,91 @@ globalThis.webViewComponent = function PlatformScriptureEditor({
...
   }, [saveUsjToPdpRaw]);
 
+  // `useProjectData`'s underlying `useData` hook doesn't reset its value back to the default when
+  // the selector (here, `scrRef`) changes — it keeps the previous chapter's USJ until the new
+  // subscription's first update lands. Gating on `isUsjFromPdpLoading` (which DOES flip `true` as
+  // soon as the chapter changes, ahead of the round trip) prevents `isBlankChapter` from being
+  // computed against stale content: without this, navigating from a blank chapter into a populated
+  // one could briefly show the empty-chapter view (and its live, clickable button) over the
+  // still-loading real content.
+  const isBlankChapter = useMemo(
```
captaincrazybro, line 1307: **Fixed.** `isBlankChapter` is now also gated on the `ChapterUSJ` subscription's `isLoading` flag (the third element of the `useProjectData(...).ChapterUSJ(...)` tuple, previously dropped). While a chapter navigation's subscription hasn't yet delivered fresh data, `isBlankChapter` is forced `false`, so the empty-chapter view (and its live button) can no longer render against the previous chapter's stale USJ.

extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx
```diff
@@ -442,7 +445,15 @@ globalThis.webViewComponent = function PlatformScriptureEditor({
...
   }, [versificationPdp, currentBookNum]);
-  const [lastVersesInCurrentBook] = usePromise(fetchLastVersesInCurrentBook, undefined);
+  // `preserveValue: false` clears the value the instant `currentBookNum` changes (rather than
+  // waiting for the new book's fetch to resolve). Without this, `usePromise`'s default of
+  // preserving the previous value would let `getEndVerse` briefly return the OLD book's verse
+  // counts for the NEW book — `currentBookNum` updates synchronously with `scrRef.book`, but this
+  // array would otherwise lag behind until the refetch lands, and a click during that window would
+  // scaffold the wrong number of `\v` markers.
+  const [lastVersesInCurrentBook] = usePromise(fetchLastVersesInCurrentBook, undefined, {
+    preserveValue: false,
```
captaincrazybro, line 445: **Fixed.** Passed `{ preserveValue: false }` to `usePromise` here, so `lastVersesInCurrentBook` clears the instant `currentBookNum` changes instead of serving the previous book's verse-count array until the new book's fetch resolves. Confirmed via `usePromise`'s source that `preserveValue` defaults to `true` (keeps the stale value across a promise-factory-identity change) — that default was exactly the gap you flagged, since `getEndVerse`'s book-number guard can't catch it: by the time of that check, the book number already matches the new book, but the array behind it didn't yet.

extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx
```diff
@@ -1296,6 +1307,91 @@ globalThis.webViewComponent = function PlatformScriptureEditor({
...
+  );
+
+  // Tracks a scaffold insert this component triggered that hasn't yet been observed to land
+  // (`isBlankChapter` flipping to `false`) or been superseded by navigating away. Serves two roles
+  // that are always set and reset together: (1) re-entrancy guard for the whole applyUpdate -> save
+  // -> PDP-echo round trip (~300ms measured) — while `true`, a second click is a no-op instead of
+  // inserting the scaffold twice; (2) "the chapter just stopped being blank because of THIS insert"
+  // signal for the refocus effect below, distinguishing that from any other cause of the same
+  // transition (ordinary navigation, a remote/collaborative update, a Send/Receive sync landing).
+  const pendingScaffoldInsertRef = useRef(false);
```
captaincrazybro, line 1324: **Fixed.** Merged `isInsertInFlightRef`/`didInsertScaffoldRef` into this single `pendingScaffoldInsertRef` (they were always set/reset together, per your smaller suggestion) and added a `useEffect` right below it, keyed on `[scrRef.book, scrRef.chapterNum]`, that resets the ref on every chapter-identity change. This closes all three scenarios you traced: the button no longer gets stuck on a second blank chapter (the ref no longer leaks across navigation), it no longer stays stuck after a rejected/reverted save once the user navigates away and back, and the refocus effect can no longer steal focus on an unrelated chapter — the reset effect's chapter-change trigger always lands in an earlier commit than the refocus effect's own `isBlankChapter`-driven trigger, since the latter depends on the (slower) loading-flag flip rather than the (immediate) `scrRef` change. One gap not closed: a rejected save while the user stays on the *same* chapter with no navigation at all still has no timeout/failure-driven reset — that would need an explicit save-failure signal plumbed into this component, which doesn't exist yet on this path. Left as a known follow-up rather than blocking this fix.

extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts
```diff
@@ -1052,3 +1053,88 @@ export async function openOrUpdateRelatedPanels(
...
+/**
+ * Builds the Delta operations that insert a blank `\c` + `\v 1..N` scaffold — one chapter marker
+ * followed by one verse marker per verse, each with empty text content. Intended for
+ * `EditorRef.applyUpdate` when reinstating a chapter number in an effectively-blank chapter.
+ *
+ * @param chapterNum The chapter number to insert (matching the chapter shown in the BCV control).
+ * @param lastVerseNum The last verse number for this chapter, per the project's versification.
+ * @returns The Delta operations, in insertion order.
+ */
+export function buildChapterScaffoldOps(chapterNum: number, lastVerseNum: number): DeltaOp[] {
```
captaincrazybro, line 1108: Skipped. Traced the paragraph-wrapping half through `$applyUpdate`/`$insertNodeAtCharacterOffset` in scripture-editors: inserting chapter+verse embeds with no intervening `para` op does get auto-wrapped in a single shared `ImpliedParaNode` (verified the index bookkeeping collapses correctly across the whole verse run, not one ImpliedParaNode per verse), and `ImpliedParaNode.getMarker()` reports `"p"` — the same default a real paragraph would use — so this is likely fine, not the bare `\c 1 \v 1 \v 2` you were concerned about. Couldn't fully close the loop to the actual serialized USFM without a live app instance, so this isn't proven, just de-risked by source tracing. The versification-fidelity half (linear `1..lastVerseNum`, ignoring excluded/segmented verses) is a real, verified gap — re-homing through a PDP/command peer of the C# scaffolding (as you suggested) is bigger than this PR and left as a known limitation for now.

extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts
```diff
@@ -1052,3 +1053,88 @@ export async function openOrUpdateRelatedPanels(
...
+/**
+ * Whether the "Add Chapter Number" button should be enabled for a chapter with the given last verse
+ * number (as returned by `getEndVerse`). `getEndVerse` returns `0` for chapter 0 or any chapter
+ * with no versification entry — a reachable case (e.g. front-matter references use `chapterNum:
+ * 0`), not just a defensive one — and the button must not render as clickable then.
+ *
+ * @param lastVerse The chapter's last verse number, from `getEndVerse(book, chapterNum)`.
+ * @returns `true` if the button should be enabled, `false` otherwise.
+ */
+export function canAddChapterNumber(lastVerse: number): boolean {
```
captaincrazybro, line 1125: Skipped. Kept as-is — the isolated-unit-test rationale you named is exactly why these were extracted (both are pure and directly tested), so this is a style preference rather than a defect. Happy to inline them if you'd still prefer that after seeing the tests.

extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx
```diff
@@ -1296,6 +1307,91 @@ globalThis.webViewComponent = function PlatformScriptureEditor({
...
+      // `showButton` below already gates on `canAddChapterNumber(lastVerse)`, so this should be
+      // unreachable in normal use. Warn instead of silently returning so a future gap in that gate
+      // is diagnosable rather than presenting as "the button did nothing."
+      logger.warn(
+        `handleAddChapterNumber: no versification entry for ${scrRef.book} chapter ${scrRef.chapterNum}; ignoring click`,
+      );
+      return;
+    }
+    pendingScaffoldInsertRef.current = true;
+    // KNOWN CAVEAT: undo reliability for this insert depends on unreleased fixes in the vendored
```
captaincrazybro, line 1360: Skipped (for now). This is a fair process point rather than a code defect — the gap is real and already documented here and in both PR descriptions, but you're right that it should have an actual tracked ticket rather than living only in this comment. That's a call for the team/ticket-owner to make, not something I can create from this session. Leaving it open until a ticket reference can be added here.

katherinejensen00 (commented):
Thanks for the thorough follow-ups, Levi — I re-checked each against the code and the three behavioral fixes (the `DisabledActionTooltip` dedup, `preserveValue: false`, and the `isUsjFromPdpLoading` gate) all hold up. No concerns there.

That leaves a few open items. I'm fine merging with any of these deferred if you'd rather track them separately — just want the decision to be explicit rather than implicit:

1.  **Undo unreliability** — you agreed this warrants a real ticket rather than living in a code comment. Could you drop a ticket # into the comment before merge, or confirm you'd like to merge without one?
2.  **Scaffold correctness** — the para-wrapping is de-risked by source tracing but not verified against saved USFM, and versification fidelity (excluded/segmented verses) is a known gap. Comfortable shipping as a documented limitation, as long as we're consciously choosing that.
3.  **Guard-ref residual** — the rejected-save-on-same-chapter-with-no-navigation case still has no failure reset. Fine as a follow-up; just flagging it's still live.

Two small ones that I think slipped through — quick to clear if you agree:

4.  `usjFromPdp ?? defaultUsj` — `usjFromPdp` is non-optional here, so the `?? defaultUsj` is dead.
5.  The refocus effect re-fires `focus()` but not `scrollToVerse`, so a blank chapter opened at verse > 1 lands scrolled to the top.

Your call on all of these — happy to approve once you've either addressed them or confirmed which you're intentionally deferring.

> @katherinejensen00 reviewed 14 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2645#-:-OzNtTiF7W9VTvW0qQEV:b-fd4owq)* status: all files reviewed, 14 unresolved discussions (waiting on captaincrazybro).

katherinejensen00 (commented):
> @katherinejensen00 resolved 7 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2645)* status: all files reviewed, 7 unresolved discussions (waiting on captaincrazybro).

extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx
```diff
@@ -1296,6 +1307,97 @@ globalThis.webViewComponent = function PlatformScriptureEditor({
...
 
+  // `useProjectData`'s underlying `useData` hook doesn't reset its value back to the default when
+  // the selector (here, `scrRef`) changes — it keeps the previous chapter's USJ until the new
+  // subscription's first update lands. Gating on `isUsjFromPdpLoading` (which DOES flip `true` as
+  // soon as the chapter changes, ahead of the round trip) prevents `isBlankChapter` from being
+  // computed against stale content: without this, navigating from a blank chapter into a populated
+  // one could briefly show the empty-chapter view (and its live, clickable button) over the
+  // still-loading real content.
+  const isBlankChapter = useMemo(
+    () => !isUsjFromPdpLoading && isChapterBlank(usjFromPdp ?? defaultUsj),
```
captaincrazybro, line 1308: Re-verified this one rather than taking either of our previous passes at face value: reproduced the exact `tsc` error by temporarily removing `?? defaultUsj` again (`Argument of type 'Usj | undefined' is not assignable to parameter of type 'Usj'`), then checked every other use of `usjFromPdp` in this file — lines 1485, 1632, 1838, 1926, and 2148 all already guard it as possibly falsy (`if (usjFromPdp && ...)`, `if (!usjFromPdp || ...)`, etc). So this isn't dead per the type checker or per how the rest of the file already treats this value — kept as-is. Happy to dig further if you have a specific reason to believe the type is wrong, but wanted to show the check rather than just re-assert the same conclusion.

extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx
```diff
@@ -1296,6 +1307,97 @@ globalThis.webViewComponent = function PlatformScriptureEditor({
...
+  // also flip `isBlankChapter` to `false`) do not steal focus or scroll. Also gated on
+  // `!isPowerMode`: the empty-chapter-view feature (and this button) is Simple-mode only, so the
+  // button can't render in Power mode today — the guard is currently redundant but documents that
+  // invariant and protects against a future refactor.
+  useEffect(() => {
+    if (isBlankChapter) return;
+    // The round trip the in-flight guard above was waiting for has completed, regardless of what
+    // caused this transition.
+    if (!isPowerMode && pendingScaffoldInsertRef.current) {
+      scrollToVerse(scrRef);
```
captaincrazybro, line 1385: **Fixed.** Added a `scrollToVerse(scrRef)` call right here, alongside `.focus()`. Confirmed the mechanism you described: the 'scroll the selected verse' effect further down is keyed on `scrRef`, which is unchanged across this insert (same chapter, no navigation), so it never re-fires on its own once `Editorial` un-hides. Called synchronously (no setTimeout, unlike that other effect) since `applyUpdate` updates the Lexical editor state directly and should have already rendered the new verse markers into the DOM by the time the PDP round trip completes and `isBlankChapter` flips — the other effect's delay is about waiting for Lexical's initial load, a different concern than this one.

captaincrazybro:
Responding to the three non-code items from the last review pass:

1. **Undo unreliability** — filing a real ticket now rather than merging with just the code comment. Will follow up here with the ticket number once it's created.
2. **Scaffold correctness** — confirmed: shipping the linear verse-loop (and the de-risked-but-not-fully-proven paragraph wrapping) as a documented known limitation, consciously rather than by default.
3. **Guard-ref residual** (rejected-save-on-same-chapter-with-no-navigation) — agreed, no action needed now; noted as still-live.

Also pushed a fix for the two remaining inline items (re-verified the `?? defaultUsj` claim against `tsc` and the file's own existing patterns — still not dead — and added the missing `scrollToVerse` call) — replies posted on those threads.

captaincrazybro:
Update on point 1 (undo unreliability) — walking this back. On reflection, the scope here is narrow enough that I don't think a tracked ticket is necessary: this only affects undoing the "Add Chapter Number" scaffold insert itself (an already-blank chapter gaining a \`\c\` + \`\v 1..N\` skeleton with no real text yet), not general editing in a populated chapter. If the insert doesn't undo cleanly, the practical recovery is deleting the chapter number and verse markers again — mildly annoying, not data-lossy. The code comment at the insert site already documents the caveat and its two root causes (one fixed in scripture-editors #526, one open in \`use-editor-pdp-sync.hook.ts\`) for whoever picks up the editor version bump later. Comfortable merging without a separate ticket for this.

katherinejensen00 (commented):
Sounds good. Thanks for looking into those. This PR is done!

> @katherinejensen00 reviewed 1 file and all commit messages, made 1 comment, and resolved 5 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2645#-:-OzOLiIL5S0yrqwvlEmc:b5ewz7t)* status: all files reviewed, 4 unresolved discussions (waiting on captaincrazybro).

katherinejensen00 (approved):
> @katherinejensen00 resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2645)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

katherinejensen00 (approved):
> @katherinejensen00 partially reviewed 15 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2645)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #17, opened 2023-02-01
Added proper icon files

tjcouch-sil:
- Renamed pt-react.png to icon.png so electron-builder will hopefully pick it up for all platforms
- Added other icon files in a new assets-excluded folder whose contents are not included in the electron-builder package (so we can put whatever we want in there)

Resolves #6

PR #2648, opened 2026-08-06
Add placement-agnostic character-marker control (Simple mode only)

jolierabideau:
# Code Review Summary

**Branch**: character-marker-control

**Base**: origin/main

**Date**: 2026-08-07

**Files changed**: 26

## Overview

This branch implements Character-marker control (placement-agnostic).
It adds `CharacterMarkerControl` (a `Popover` + `MarkerMenu` trigger showing the character marker at
the caret), `CharacterMarkerToolbar` (the slot-based container that carries the Simple-mode gate),
`useCharacterMarkerState` (composes editor state into the control's props), `useIsPowerMode`, and
`computeCharacterMarkerCoverage` (the single coverage analysis feeding the `(mixed)` label, the
per-row selection state, and eventually `PT-XXX-B4`'s gap list). In `platform-bible-react`,
`MarkerMenuItem` gains two optional fields — `selectionState` and `isDisabled` — plus a start-side
selection indicator.

The two constraints the plan cares most about hold, and are asserted rather than assumed. The
`MarkerMenuItem` additions are purely additive and inert when absent, which matters because
`MarkerMenu` also serves the `\` keydown menu and the footnote editor, both of which render in Power
mode. And `getUsj()` — which serializes the whole chapter — is reachable only from `onOpen`, pinned
by tests rather than by convention. Placement-agnosticism is confirmable from the diff alone: no
`TabToolbar` import, no geometry calls, no positioning CSS.

**Nothing here is mounted in production yet.** The web view imports only the string-keys constant;
mounting is `PT-XXX-D1`/`D2`. So this branch is test- and Storybook-verifiable, not manually
demo-able. That matches the plan, but it means the reviewer should not expect a running demo.

## API Changes

- **`lib/platform-bible-react` (`MarkerMenuItem`)**: added optional `selectionState?: 'all' |
  'partial' | 'none'`. Display-only; affects neither visibility nor selectability. The indicator and
  `aria-checked` render only when the field is supplied.
- **`lib/platform-bible-react` (`MarkerMenuItem`)**: added optional `isDisabled?: boolean`, folded
  into the existing `disabled={isDisallowed || isDeprecated || isDisabled}`. Renders no trailing
  label (unlike the other two, it describes the consumer's inability to act, not the marker).
- **`lib/platform-bible-react` (`MARKER_MENU_STRING_KEYS`)**: added
  `%markerMenu_searchPlaceholder_character%`, joining `_insert` and `_paragraph`.
- **`lib/platform-bible-utils`**: no changes.
- **`lib/papi-dts/papi.d.ts`**: no changes — no `@papi/` surface touched.
- **Extension type declaration files**: no changes.
- **Extension-internal exports** (not a public API surface): `CharacterMarkerControl`,
  `CharacterMarkerToolbar`, their prop types, `CHARACTER_MARKER_CONTROL_STRING_KEYS`,
  `DisabledTooltipWrapper`, `computeCharacterMarkerCoverage` + its types, `useCharacterMarkerState`,
  `useIsPowerMode`. `correctEditorUsjVersion` moved from module-private in the web view to exported
  from `platform-scripture-editor.utils.ts`.

None.

- [x] **A partially-covered selection made every menu row silently inert** — `resolveCurrentMarker`
      treats a marker covering only part of the selection as "applied", so with
      `changeCharacterMarker` not yet wired every row's action became a no-op with no feedback.
      _(fixed during review: kept the inert semantics, which are correct and plan-blessed —
      `insertMarker` nests rather than replaces, verified upstream — and instead marked the inert
      rows `isDisabled` so a click is refused visibly instead of swallowed. Four tests, including
      "disables nothing when no marker is applied, so adding still works".)_
- [x] **The menu never closed on selection, so focus never returned to the editor** — `MarkerMenu`
      wires `onSelect` straight to `item.action`, and the host's `closeMarkersMenu` was `() => {}`.
      _(fixed during review: the host wraps each item's action to also close; `handleOpenChange` now
      ignores no-op transitions so the editor refocus doesn't reach Radix as a second close and
      double-fire `onClose`. Three tests, including keyboard activation via ArrowDown+Enter.)_
- [x] **The visible trigger label could not truncate** — a bare text child of a `tw:shrink-0
      tw:whitespace-nowrap` Button overflows instead of clipping, and a placement wrapper could not
      fix it from `className`. _(fixed during review: label wrapped in a `tw:truncate` span, Button
      given `tw:min-w-0 tw:shrink`, tooltip gains truncation as a third trigger via
      `useTruncationTooltip`. Tooltip deliberately left uncontrolled — see Interview Notes.)_
- [x] **Mixed single-select / multi-select signals** — `ChevronDown` + one trigger value says
      single-select; checkbox-like squares plus a menu that stays open say multi-select. _(fixed
      during review: closing on selection restored the single-select behavior, and UX decided
      2026-08-06 to show a check for both `'all'` and `'partial'` and nothing for `'none'`.
      `aria-checked="mixed"` is retained so screen-reader users keep the distinction the visual now
      drops.)_
- [x] **Two names for the same thing in one control** — the trigger said "Character marker" while
      the search placeholder said "character style". _(fixed during review: settled on the PRD's
      term per the plan's Terminology table — "Search character markers" / "Buscar marcadores de
      carácter" — and dropped the trailing full stop.)_
- [x] **User-visible separators hardcoded instead of translated** — `" - "` in the label and `": "`
      in the aria-label were baked into code, against the Localization Guide and against this
      extension's own `formatEditorTitle` precedent. _(fixed during review: added
      `%..._label_format%` and `%..._ariaLabel_format%` in `en` + `es`, interpolated with
      `formatReplacementString`. The existing parity test asserts es ≠ en, which cannot hold for
      format strings, so a separate guard asserts both locales carry every placeholder.)_
- [ ] ~~`useIsPowerMode` duplicates the `platform.interfaceMode` read already in
      `use-structure-protection-state.hook.ts:68-71`, leaving two independent readers in one
      extension that can diverge on error handling.~~ _(Author: deliberate — consolidating it will be
      a follow-up. Not done here to avoid touching Power-mode-adjacent code on this branch.)_
- [ ] ~~The Simple-mode gate lives only in `CharacterMarkerToolbar`, yet `CharacterMarkerControl` is
      also exported, so a wrapper mounting it directly would silently lose the "not rendered at all
      in Power mode" guarantee.~~ _(Author: remove the export if unused, otherwise leave the code
      alone. Verified the export is load-bearing — the test file mounts it bare in five tests, the
      story file uses it, and `useCharacterMarkerState` imports `CharacterMarkerControlProps` to
      derive its return type — so the code is unchanged. Partly mitigated by the folder split, which
      moved the gate into a file named for it.)_

### Minor — Consider

- [x] Two `dist/experimental.js`/`.cjs` files carried 18 lines of unrelated build churn with zero
      feature content. _(fixed during review: reverted to `main`; verified 0 feature-related lines.)_
- [x] `CharacterMarkerSelectionState` was exported with no consumers while the literal union was
      written inline in 5+ places. _(fixed during review: the alias is now used at every site, plus a
      private `Exclude<…, 'none'>` variant for `markerStates`, which genuinely is only
      `'all' | 'partial'`.)_
- [x] `useIsPowerMode`'s TSDoc cited "the editor web view's own memo", which does not exist. _(fixed
      during review: now points at `useStructureProtectionState`, the real comparable read.)_
- [x] `tw:rounded-[3px]` arbitrary value instead of the scale's `tw:rounded-sm`. _(fixed during
      review: the element carrying it was removed by the UX indicator change.)_
- [x] Flash-of-wrong-mode — defaulting to `'simple'` while the setting resolved meant Power would
      paint the toolbar's children on first frame and then unmount them. _(fixed during review:
      `useSetting` exposes a 4th `isLoading` element, so `useIsPowerMode` now returns
      `boolean | undefined` and the toolbar gates on `isPowerMode !== false`. Tested both
      directions.)_
- [x] Two exported components in one file, with the mode gate buried in a file named "control".
      _(fixed during review: `character-marker-control/` folder created with `git mv`;
      `CharacterMarkerToolbar` split into `character-marker-toolbar.component.tsx` with its own
      tests.)_
- [x] The disabled-tooltip wrapper was copy-pasted from the paragraph switcher. _(fixed during
      review: extracted as `DisabledTooltipWrapper` with its own tests, used by the new control
      only — see Interview Notes for why the paragraph switcher was left alone.)_
- [x] The search-placeholder key sat in the extension while its two siblings live in the lib.
      _(fixed during review: moved to `%markerMenu_searchPlaceholder_character%` in
      `platform-bible-react` and added to `MARKER_MENU_STRING_KEYS`; the web view already preloads
      that list, so no new registration was needed.)_
- [x] `MarkerMenu` had no `.stories.tsx` in the shared library despite gaining a new affordance.
      _(fixed during review: added, covering `'all'`/`'partial'`/`'none'`, mixed, the no-field
      legacy case, `isDisabled`, the deprecated/disallowed states, and a narrow-width long-title
      story.)_
- [x] Row titles did not truncate in the now-200px popover. _(fixed during review: title and
      subtitle truncate, container given `tw:min-w-0 tw:flex-1`, native `title` for hover reveal,
      matching the `scope-selector` / `project-selector` precedent.)_
- [x] The `computeCoverage` doc comment overstated the `hasUncovered` invariant for a collapsed
      caret on a non-text node. _(fixed during review: rewritten to name all three sources of the
      "no information" result.)_
- [x] A caret inside a note reported `hasUncovered: true`, conflating "excluded from coverage" with
      "unmarked text is selected". _(fixed during review: now returns the "no information" result so
      the hook falls back to `contextMarker`, with a test verified falsifiable against the old
      behavior.)_
- [ ] **Test `STRINGS` in `character-marker-control.component.test.tsx` still hardcodes copies of the
      `en` values**, so a future copy revision leaves the tests passing against stale text. The two
      other test files examined (`use-character-marker-state.hook.test.ts`,
      `character-marker-menu.utils.test.ts`) were found NOT to have this problem — their English
      strings are deliberate input fixtures whose whole point is asserting the lookup resolved to
      them, and two tests specifically assert the fallback to the raw key. Only the control test is
      genuinely affected; left open rather than fixed, since the story file's
      `getLocalizedStrings(...)` approach would couple the assertions to the contribution file.
- [ ] ~~The new leading selection indicator diverges from `CommandItem`'s trailing `data-checked`
      check, used by `interface-language-picker`.~~ _(Not a defect: `Guidelines/Component Choices`
      explicitly says "start side checkbox (not on the end side like the default New York style)",
      so this code follows the guideline and the language picker is the outlier. No doubling occurs —
      `data-checked` is never set, asserted by test.)_
- [ ] ~~The sync-blocked tooltip reuses a string saying "Send/Receive" where the product name is now
      "Sync".~~ _(Author: pre-existing, not introduced by this branch — do not touch.)_
- [ ] ~~The whole vertical slice has no production consumer yet.~~ _(Matches the plan — mounting is
      `PT-XXX-D1`/`D2`. Not dead code, but see the Overview note: nothing here is runtime-verifiable
      on this branch.)_

#### Shared Regions Modified

None. All four analysis agents scanned the changed files for `#region shared with` markers; none
present.

#### Extension Config Changes

`.github/workflows/` files were modified.

## Positive Observations

- The `MarkerMenuItem` inertness contract is **asserted, not assumed**: tests verify that an item
  without `selectionState` renders no affordance and sets no `aria-checked`, that `data-checked` is
  never set (so `CommandItem`'s trailing check stays hidden and cannot double up), and that an item
  without `isDisabled` stays clickable. This is exactly what the plan demands for the two consumers
  that render in Power mode.
- The `getUsj()` cost constraint is **structural rather than conventional** — `getUsj` is only
  reachable from `onOpen` — and is pinned by two tests asserting it is never called on selection
  change and exactly once per menu open.
- Placement-agnosticism is confirmable from the diff alone, as the plan asked: no `TabToolbar`
  import, no `getBoundingClientRect`, no positioning CSS. `menuAlign` / `menuDirection` /
  `isLabelHidden` push those decisions to the caller.
- `menuDirection` and the conditional `dir` spread address a real RTL trap (Radix portals to
  `document.body`, which inherits no direction in this app) rather than assuming `align="end"`
  mirrors for free — and the TSDoc explains the mechanism precisely.
- The disabled-state a11y handling is genuinely thought through: because a disabled button is not
  focusable and cannot host a tooltip, the wrapper becomes focusable and named *only* while disabled.
  The tooltip text is `aria-hidden` to avoid double-announcing a value that is already the accessible
  name.
- Localization is complete and parity-tested in both `en` and `es`, with a guard that also asserts
  the Spanish value differs from the English one — catching untranslated copy-paste.
- Reuse is real, not nominal: `MarkerMenu`, the paragraph-switcher popover pattern, the
  structure-protection-button key-list/`localize` convention, and `useTruncationTooltip` are all
  picked up rather than reinvented. `correctEditorUsjVersion` was moved rather than duplicated, with
  a type-only import preserving the extension host's no-runtime-import boundary.
- Zero edits under `shadcn-ui/`. The 220px list cap is a descendant override on the popover
  (`tw:**:data-[slot=command-list]:max-h-[220px]`) precisely so the shared component stays untouched,
  with the reasoning recorded at the call site.
- Non-obvious decisions are documented where they are made, not only in the PR: why `isMixed` is
  deliberately not memoized, why `dir` is spread conditionally, why the tooltip text is `aria-hidden`.
- The keyboard-shortcuts-catalog rule correctly does not apply — no keyboard handler was added
  (verified by grep, the only "keydown" is in a prose comment).

## Interview Notes

**Stated purpose**: satisfy `PT-XXX-D` from the character-level markers implementation plan.

**Key decisions the author made, and their reasoning:**

- **`useStructureProtectionState` consolidation deferred.** Deliberate — the author does not want to
  touch Power-mode-adjacent code on this branch, and will consolidate the duplicated
  `platform.interfaceMode` read in a follow-up.
- **`CharacterMarkerControl` stays exported.** The author's instruction was "remove the export if we
  don't need it, otherwise leave the code alone." Verification showed three real consumers, so the
  code is unchanged.
- **Selection indicator vocabulary.** The author relayed a UX-team decision: show a check on every
  marker that is on the selection — including a partial one — and nothing for unselected. The
  Component Choices guideline marks the exact glyph as "up for UX discussion", so the team's call
  governs. The tri-state survives in `aria-checked` for assistive tech.
- **"Follow code patterns"** was the author's guidance on two open questions (the truncation tooltip,
  and the file-layout convention), rather than a specific preference — resolved by finding the
  in-repo precedents (`useTruncationTooltip` in `model-text-panel` / `project-selector`; the
  Colocation Principle plus the extension's own `paragraph-marker-tooltip/` folder).
- **`Send/Receive` vs `Sync` naming left alone** because the string is pre-existing and not
  introduced by this branch.

**Decisions made during review that the reviewer should scrutinize:**

- **The inert-row semantics were kept, not changed.** The `api` agent proposed narrowing
  `resolveCurrentMarker` to fully-covered-only. That was rejected during review for two reasons
  worth re-checking: it would delete the remove row in exactly the U3 case the plan's §3 note was
  written to fix, and falling through to `insertMarker` produces nested markup rather than the
  replacement the user meant. The fix applied instead makes the inertness *visible*. **If the
  reviewer disagrees with that reading, this is the decision to re-open.**
- **The truncation tooltip is uncontrolled**, diverging from `project-selector`, which controls
  `open`. Rationale (recorded in a code comment): project-selector had to, because cmdk's
  `CommandItem` defeats Radix's pointer/focus auto-detection; this trigger is an ordinary button, and
  controlling `open` would cost the focus-driven tooltip the disabled wrapper depends on. The
  accepted trade-off is that keyboard focus alone does not reveal a truncated label — acceptable
  because the full text is already the accessible name.
- **`DisabledTooltipWrapper` is used by the new control only.** Extracting it and adopting it in the
  paragraph switcher too would have edited code that renders in Power mode, contradicting §1.6. The
  duplication is therefore knowingly left in the paragraph switcher, for a follow-up.

**Unresolved items**: none. Every Critical/Important finding was either fixed or explained.

**Areas the author did not understand**: none of a substantive nature. The author asked what action
three minor findings called for (the note-caret coverage signal, the leading-vs-trailing check
convention, and the no-production-consumer observation) rather than deferring on how their own code
works; each was explained and then either fixed or dismissed on the merits.

## In-Review Quality Check

Run repo-wide after all in-review changes:

- **`npm run lint`** — exit 0. Zero errors. Four warnings remain, all pre-existing and in files this
  branch does not touch (`model-text-panel.component.tsx` duplicate imports ×2,
  `use-effective-resource-reference-list.hook.ts` `no-console` ×2).
- **Prettier** (changed files only) — clean, no reformatting needed.
- **`npm test`** — all suites pass. `platform-scripture-editor`: 48 files / 673 tests.
  `platform-bible-react`: 1023 tests. Root: 1418 tests. No failures anywhere.
- **`npm run typecheck`** — **2 errors, pre-existing and unrelated to this branch.**
  `platform-scripture-editor.web-view.tsx` lines 12 and 14 import `PARAGRAPH_STRUCTURE_VIEW_MODE` and
  `StructureProtectionMode` from `@eten-tech-foundation/platform-editor`, which the locally
  yalc-linked version does not export. Verified byte-identical on `origin/main` and untouched by this
  branch — a stale local link, not a regression. **Worth confirming green in CI.**

## Suggested Review Focus

- [ ] **The inert-row decision.** On a selection a character marker partially covers, every other
      marker row is now *disabled* rather than inserting. Is "visible refusal" the right interim
      behavior until `PT-XXX-B2` (replace) and `PT-XXX-B4` (extend) land, or should adding a
      different marker over a partially-marked selection be allowed to nest? This is the one place
      the review overrode an analyzer's recommendation.
- [ ] **`aria-checked="mixed"` with a plain check glyph.** UX chose a binary visual; the accessible
      state stays tri-state. Confirm that divergence is what UX intended, since a screen-reader user
      and a sighted user now receive different granularity.
- [ ] **The Power-mode guarantee end-to-end.** `useIsPowerMode` now returns `boolean | undefined` and
      the toolbar renders nothing until the setting resolves. Worth confirming this is the right
      shape rather than, say, a `{ isPowerMode, isLoading }` tuple — and that no future caller reads
      `undefined` as falsy.
- [ ] **The two duplications knowingly left in place**: the `platform.interfaceMode` read in
      `useStructureProtectionState`, and the disabled-tooltip wrapper in the paragraph switcher. Both
      are deliberate deferrals to protect Power-mode code on this branch. Agree on who files the
      follow-ups.
- [ ] **Nothing is mounted yet.** No manual verification is possible on this branch — review against
      the tests and the Storybook stories, and treat runtime behavior as unverified until
      `PT-XXX-D1`/`D2`.
- [ ] **The pre-existing typecheck failure.** Confirm CI is green; if it is not, the yalc-linked
      editor package needs refreshing before merge.

---

This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2648)


katherinejensen00 (changes requested):
Thanks for your great work on this, Jolie!

Here is Claude's summary:
This is unusually strong work — the coverage logic is sound (I traced the collapsed-caret and nested-marker paths and they're correct), the shared-`MarkerMenu` change is genuinely additive and inert for its three Power-mode consumers, and the `import type` move creates no runtime leak into the extension host (the boundary test passes). The findings below are almost all edge-cases and polish, not mainline defects.

**Bottom line:** No blockers. The cluster worth acting on before merge is **#1–#3 + #5** (the reversed/marker-object coverage edges and the missing degrade test) and **#8** (the localization parity gap) — small, concrete, and each closes a "ships green but wrong" path. **#9–#13** are worth a look but several are AT/UX judgment calls best confirmed with the relevant owner rather than changed blind. Everything else is optional polish. And to be candid about my own review: my first-pass "hyphen" and "use `LanguageStrings`" notes were wrong (the antagonistic pass caught both), and several items others might raise — `undefined` return, dist churn, identical `_format` strings — are verified non-issues I'd push back on if a reviewer flags them.

For easy reference, here is the whole list in one spot:
**1\. Coverage silently treats a backwards selection as a caret.** When a selection inside one text node has its offsets reversed (`start.offset > end.offset`), `slice(start, end)` returns `''`, so `totalLength` is 0 and the collapsed-caret branch reports the enclosing markers as `'all'` instead of measuring the selected characters. If the editor can ever hand you anchor-after-focus offsets, coverage will be wrong for right-to-left drags. Worth confirming whether `platform-editor` normalizes selection direction; if it does, this is moot and a one-line comment saying so would close it. → [character-marker-coverage.utils.ts:140](vscode-webview://1bbvfdtssqal87jrv2dsq5vgmhtuk52mipgm3refu8v9p4781njp/extensions/src/platform-scripture-editor/src/character-marker-coverage.utils.ts#L140) · **Low–Moderate**

**2\. The reversed/unreachable-range degrade trims with the wrong node's offset.** The comment says an unreachable end "degrades to the start node alone," but the single-segment branch then does `slice(startOffset, endOffset)` using the _end_ node's offset against the _start_ node's text — an arbitrary prefix, not the whole start node. → [character-marker-coverage.utils.ts:144](vscode-webview://1bbvfdtssqal87jrv2dsq5vgmhtuk52mipgm3refu8v9p4781njp/extensions/src/platform-scripture-editor/src/character-marker-coverage.utils.ts#L144) · **Low**

**3\. A selection ending on a marker object (not its text) undercounts that marker.** If `end.jsonPath` resolves to a character `MarkerObject` rather than its text child, the forward walk stops on the object before descending, so its text never enters `segments` and its coverage is undercounted. → [character-marker-coverage.utils.ts:125](vscode-webview://1bbvfdtssqal87jrv2dsq5vgmhtuk52mipgm3refu8v9p4781njp/extensions/src/platform-scripture-editor/src/character-marker-coverage.utils.ts#L125) · **Low**

**4\. Coverage never re-syncs while the menu is open.** Coverage is computed once in `onOpen`. If the selection changes while the menu stays open (e.g. an external scroll-group update re-renders the view), the trigger label and per-row check states keep showing the open-time selection until close. If that freeze is intentional, a one-line note at the `setCoverage` site and a small test would lock it in; if not, it's a stale-state bug. → [use-character-marker-state.hook.ts:169](vscode-webview://1bbvfdtssqal87jrv2dsq5vgmhtuk52mipgm3refu8v9p4781njp/extensions/src/platform-scripture-editor/src/use-character-marker-state.hook.ts#L169) · **Low**

**5\. The reversed-range branch has no test — which is what lets #1 and #2 hide.** Every coverage test uses a forward range, so `!reachedEnd` and the single-segment offset trim are never exercised. A test with `end` before `start` in document order would pin the intended degrade and catch #1/#2. → [character-marker-coverage.utils.test.ts](vscode-webview://1bbvfdtssqal87jrv2dsq5vgmhtuk52mipgm3refu8v9p4781njp/extensions/src/platform-scripture-editor/src/character-marker-coverage.utils.test.ts) · **Moderate**

**6\. No multi-segment test actually trims the last segment.** Every multi-segment case sets `end.offset` to the last segment's full length, so `slice(0, endOffset)` is always a no-op — the trailing-trim branch runs but never discriminates. One case ending mid-word in a trailing marked run would make it meaningful. → [character-marker-coverage.utils.test.ts](vscode-webview://1bbvfdtssqal87jrv2dsq5vgmhtuk52mipgm3refu8v9p4781njp/extensions/src/platform-scripture-editor/src/character-marker-coverage.utils.test.ts) · **Low–Moderate**

**7\. The "caret in unmarked text" path is untested.** Only caret-in-marker and caret-in-note are covered; the plain-text `hasUncovered: true` branch has no test. (The logic is correct — I traced it — this is just a guard against future regressions.) → [character-marker-coverage.utils.test.ts](vscode-webview://1bbvfdtssqal87jrv2dsq5vgmhtuk52mipgm3refu8v9p4781njp/extensions/src/platform-scripture-editor/src/character-marker-coverage.utils.test.ts) · **Low**

**8\. Two new localized keys aren't parity-checked, so an untranslated one could ship.** The control-keys block in the parity test checks a hardcoded 4-of-8 subset rather than the exported `CHARACTER_MARKER_CONTROL_STRING_KEYS`, so `%…_syncEditBlocked_banner%` and `%markerMenu_searchPlaceholder_character%` are never checked for en/es parity — the precise gap this test exists to close. Driving the loop off the exported constant fixes it and stays correct as keys are added. → [localized-strings.test.ts](vscode-webview://1bbvfdtssqal87jrv2dsq5vgmhtuk52mipgm3refu8v9p4781njp/extensions/src/platform-scripture-editor/src/localized-strings.test.ts) · **Low–Moderate**

**9\. The enabled trigger can advertise a tooltip that isn't mounted.** `useTruncationTooltip` is built for a _controlled_ `<Tooltip open>`, but here the Tooltip is uncontrolled and only `<TooltipContent>` is conditionally rendered. Radix still opens on hover/focus of an enabled, non-truncated trigger and sets `aria-describedby` pointing at content that isn't in the DOM — a dangling reference a screen reader will try to resolve. Either control `open` with the hook's value, or assert `aria-describedby` is absent in that state. → [character-marker-control.component.tsx:327](vscode-webview://1bbvfdtssqal87jrv2dsq5vgmhtuk52mipgm3refu8v9p4781njp/extensions/src/platform-scripture-editor/src/character-marker-control/character-marker-control.component.tsx#L327) · **Low–Moderate**

**10\. The disabled-state tooltip trigger uses a role that may not announce its label.** While disabled, the wrapper becomes a focusable `role="group"` named by `aria-label`. `group` is a non-interactive container role, and several screen readers don't announce a container's own label on focus — so the "why is this disabled?" text (the whole point of the wrapper) may be silent. `role="button"` + `aria-disabled="true"` stays focusable and is reliably announced. Please verify on a real screen reader before changing, since this is AT-dependent. → [disabled-tooltip-wrapper.component.tsx:50](vscode-webview://1bbvfdtssqal87jrv2dsq5vgmhtuk52mipgm3refu8v9p4781njp/extensions/src/platform-scripture-editor/src/disabled-tooltip-wrapper.component.tsx#L50) · **Low–Moderate**

**11\. `aria-checked` coexists with cmdk's own `aria-selected` on every row.** cmdk's `CommandItem` is `role="option"` and already sets `aria-selected` for its keyboard cursor; adding `aria-checked` puts two selection semantics on the same row, and `none → aria-checked="false"` labels ~40 rows "not checked." It's a deliberate, tested choice — I'd just confirm with whoever owns AT/UX that a screen-reader user hearing both is the intended readout, and consider omitting `aria-checked` on `none` rows. → [marker-menu.component.tsx:158](vscode-webview://1bbvfdtssqal87jrv2dsq5vgmhtuk52mipgm3refu8v9p4781njp/lib/platform-bible-react/src/components/advanced/marker-menu.component.tsx#L158) · **Low**

**12\. The RTL-safe default is inverted for the future placement wrapper.** `menuDirection` defaults to `undefined`, which resolves to physical `ltr` (the `localStorage` key it reads has no writer in the web view). So an RTL project whose wrapper passes `menuAlign="end"` but forgets `menuDirection` gets a menu that collides off the trailing edge — the dangerous case is the forgot-a-prop default. It's well documented; you might just make the wrapper pass `menuDirection` unconditionally, or default it from the resolved project direction. → [character-marker-control.component.tsx:116](vscode-webview://1bbvfdtssqal87jrv2dsq5vgmhtuk52mipgm3refu8v9p4781njp/extensions/src/platform-scripture-editor/src/character-marker-control/character-marker-control.component.tsx#L116) · **Low**

**13\. The menu list is re-filtered and re-sorted on every open/close.** `markerMenuItems` depends on `currentMarker`/`coverage`, which change each open/close, so the `filter + localeCompare` over the block's children re-runs even though the list itself only depends on `blockMarker`/strings/callbacks. Splitting list-generation (memoized on those) from the per-row `isDisabled`/`selectionState` stamping avoids the repeated sort. → [use-character-marker-state.hook.ts:207](vscode-webview://1bbvfdtssqal87jrv2dsq5vgmhtuk52mipgm3refu8v9p4781njp/extensions/src/platform-scripture-editor/src/use-character-marker-state.hook.ts#L207) · **Low–Moderate**

**14\. Two spots hand-roll `cn()`.** `['tw:…', className].filter(Boolean).join(' ')` appears in both the control and the toolbar; `cn` is already exported from `platform-bible-react` (and imported by this component's own sibling files). Besides being the house helper, `cn`'s tailwind-merge de-dupes conflicting utilities — the exact class-precedence issue the comment at line 267 is worrying about. → [character-marker-control.component.tsx:268](vscode-webview://1bbvfdtssqal87jrv2dsq5vgmhtuk52mipgm3refu8v9p4781njp/extensions/src/platform-scripture-editor/src/character-marker-control/character-marker-control.component.tsx#L268) & [character-marker-toolbar.component.tsx:35](vscode-webview://1bbvfdtssqal87jrv2dsq5vgmhtuk52mipgm3refu8v9p4781njp/extensions/src/platform-scripture-editor/src/character-marker-control/character-marker-toolbar.component.tsx#L35) · **Low**

**15\. `coveringMarkers` memo is redundant.** It re-computes `Object.keys(coverage.markerStates)` that `resolveCurrentMarker` already derives, and only feeds the intentionally-non-memoized `isMixed` block — a plain `const` does the same for less. → [use-character-marker-state.hook.ts:177](vscode-webview://1bbvfdtssqal87jrv2dsq5vgmhtuk52mipgm3refu8v9p4781njp/extensions/src/platform-scripture-editor/src/use-character-marker-state.hook.ts#L177) · **Nit**

**16\. The "no-information" test is duplicated.** `computeCoverage` re-derives by hand what the coverage module already encodes as `EMPTY_COVERAGE`; an exported `isEmptyCoverage` keeps that definition in one place. → [use-character-marker-state.hook.ts:66](vscode-webview://1bbvfdtssqal87jrv2dsq5vgmhtuk52mipgm3refu8v9p4781njp/extensions/src/platform-scripture-editor/src/use-character-marker-state.hook.ts#L66) · **Nit**

**17\. A few comments point at placeholder tickets.** `PT-XXX-B2/B4` and `PT-XXX-D1/D2` were never filled in, so they reference nothing. Keep the behavioral rationale, drop or substitute the tokens. → [use-character-marker-state.hook.ts:128](vscode-webview://1bbvfdtssqal87jrv2dsq5vgmhtuk52mipgm3refu8v9p4781njp/extensions/src/platform-scripture-editor/src/use-character-marker-state.hook.ts#L128) & [platform-scripture-editor.web-view.tsx:165](vscode-webview://1bbvfdtssqal87jrv2dsq5vgmhtuk52mipgm3refu8v9p4781njp/extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx#L165) · **Low**

**18\. A few comments are so long the key line gets lost.** The 17-line tooltip block (188-204) bundles four ideas mid-JSX; the `aria-hidden` block (329-335) repeats a principle stated in two other places; and the "verified by grep" instruction (137-141) now matches its own prose. The content is good — it's a distribution problem. Splitting/de-duping these three keeps the load-bearing sentences skimmable. → [character-marker-control.component.tsx:188](vscode-webview://1bbvfdtssqal87jrv2dsq5vgmhtuk52mipgm3refu8v9p4781njp/extensions/src/platform-scripture-editor/src/character-marker-control/character-marker-control.component.tsx#L188) (and 329, 137) · **Nit**

**19\. Two deliberate duplications need real owners, not prose follow-ups.** The second `platform.interfaceMode` reader (`useIsPowerMode` vs `useStructureProtectionState`) and the disabled-tooltip pattern still inline in the paragraph switcher are both fine to defer — but "follow-up" in a comment tends to evaporate; a tracked ticket for each would make the deferral real. (Note: the two interface-mode readers have intentionally different contracts, so there's no divergence bug today.) → [use-is-power-mode.hook.ts:25](vscode-webview://1bbvfdtssqal87jrv2dsq5vgmhtuk52mipgm3refu8v9p4781njp/extensions/src/platform-scripture-editor/src/use-is-power-mode.hook.ts#L25) · **Low**

**20\. The gate lives one layer out from the exported control.** `CharacterMarkerControl` is exported (rightly — it has three consumers), but the Simple-mode gate is in `CharacterMarkerToolbar`, so a future wrapper that mounts the control directly loses the "never in Power" guarantee. A short note on the export naming the toolbar as its required host, or a test asserting no bare mount, would protect it. → [character-marker-control.component.tsx:143](vscode-webview://1bbvfdtssqal87jrv2dsq5vgmhtuk52mipgm3refu8v9p4781njp/extensions/src/platform-scripture-editor/src/character-marker-control/character-marker-control.component.tsx#L143) · **Low**

> @katherinejensen00 reviewed 26 files and all commit messages, and made 21 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2648#-:-OzS5SCr3kSoHR8jiffq:b-a6rx5p)* status: all files reviewed, 20 unresolved discussions (waiting on jolierabideau).

___
*[`lib/platform-bible-react/src/components/advanced/marker-menu.component.tsx` line 158 at r1](https://reviewable.io/reviews/paranext/paranext-core/2648#-OzSFWb51OpzH_Djp2pM:-OzSFWb51OpzH_Djp2pN:b-a0zx75) ([raw file](https://github.com/paranext/paranext-core/blob/b95824d68fd676ff199c5fe9f26e8ae31876c0c0/lib/platform-bible-react/src/components/advanced/marker-menu.component.tsx#L158)):*
> ```XML
>       // Never pair this with `data-checked`: CommandItem renders its own trailing check for that,
>       // which would double the checkmark.
>       aria-checked={
> ```

**11\. `aria-checked` coexists with cmdk's own `aria-selected` on every row.** cmdk's `CommandItem` is `role="option"` and already sets `aria-selected` for its keyboard cursor; adding `aria-checked` puts two selection semantics on the same row, and `none → aria-checked="false"` labels ~40 rows "not checked." It's a deliberate, tested choice — I'd just confirm with whoever owns AT/UX that a screen-reader user hearing both is the intended readout, and consider omitting `aria-checked` on `none` rows.
___
*[`extensions/src/platform-scripture-editor/src/character-marker-coverage.utils.ts` line 125 at r1](https://reviewable.io/reviews/paranext/paranext-core/2648#-OzS6Yth65XTE-U8A-R7:-OzS6Yth65XTE-U8A-R8:bezo7q2) ([raw file](https://github.com/paranext/paranext-core/blob/b95824d68fd676ff199c5fe9f26e8ae31876c0c0/extensions/src/platform-scripture-editor/src/character-marker-coverage.utils.ts#L125)):*
> ```XML
>   usjRW.findNextMatchingNode(startLocation, ({ node, documentLocation }) => {
>     if (isString(node)) segments.push({ jsonPath: documentLocation.jsonPath, text: node });
>     if (documentLocation.jsonPath === endJsonPath) {
> ```

**3\. A selection ending on a marker object (not its text) undercounts that marker.** If `end.jsonPath` resolves to a character `MarkerObject` rather than its text child, the forward walk stops on the object before descending, so its text never enters `segments` and its coverage is undercounted.
___
*[`extensions/src/platform-scripture-editor/src/character-marker-coverage.utils.ts` line 144 at r1](https://reviewable.io/reviews/paranext/paranext-core/2648#-OzS6SxL3bK5uw8_IIOZ:-OzS6SxL3bK5uw8_IIO_:b-6r0xyj) ([raw file](https://github.com/paranext/paranext-core/blob/b95824d68fd676ff199c5fe9f26e8ae31876c0c0/extensions/src/platform-scripture-editor/src/character-marker-coverage.utils.ts#L144)):*
> ```XML
>       segments[0].text = segments[0].text.slice(startOffset);
>       const last = segments[segments.length - 1];
>       last.text = last.text.slice(0, endOffset);
> ```

**2\. The reversed/unreachable-range degrade trims with the wrong node's offset.** The comment says an unreachable end "degrades to the start node alone," but the single-segment branch then does `slice(startOffset, endOffset)` using the _end_ node's offset against the _start_ node's text — an arbitrary prefix, not the whole start node.
___
*[`extensions/src/platform-scripture-editor/src/character-marker-coverage.utils.ts` line 140 at r1](https://reviewable.io/reviews/paranext/paranext-core/2648#-OzS6KkX0Eg-puc3CWDS:-OzS6KkX0Eg-puc3CWDT:betaa1p) ([raw file](https://github.com/paranext/paranext-core/blob/b95824d68fd676ff199c5fe9f26e8ae31876c0c0/extensions/src/platform-scripture-editor/src/character-marker-coverage.utils.ts#L140)):*
> ```XML
>     const endOffset = selection.end?.offset;
>     if (segments.length === 1) {
>       segments[0].text = segments[0].text.slice(startOffset, endOffset);
> ```

**1\. Coverage silently treats a backwards selection as a caret.** When a selection inside one text node has its offsets reversed (`start.offset > end.offset`), `slice(start, end)` returns `''`, so `totalLength` is 0 and the collapsed-caret branch reports the enclosing markers as `'all'` instead of measuring the selected characters. If the editor can ever hand you anchor-after-focus offsets, coverage will be wrong for right-to-left drags. Worth confirming whether `platform-editor` normalizes selection direction; if it does, this is moot and a one-line comment saying so would close it. → **Low–Moderate**
___
*[`extensions/src/platform-scripture-editor/src/character-marker-coverage.utils.test.ts` at r1](https://reviewable.io/reviews/paranext/paranext-core/2648#-OzS755L4UJpkjtNVUXx:-OzS755L4UJpkjtNVUXy:bbmit5a) ([raw file](https://github.com/paranext/paranext-core/blob/b95824d68fd676ff199c5fe9f26e8ae31876c0c0/extensions/src/platform-scripture-editor/src/character-marker-coverage.utils.test.ts)):*
**7\. The "caret in unmarked text" path is untested.** Only caret-in-marker and caret-in-note are covered; the plain-text `hasUncovered: true` branch has no test. (The logic is correct — I traced it — this is just a guard against future regressions.)
___
*[`extensions/src/platform-scripture-editor/src/character-marker-coverage.utils.test.ts` at r1](https://reviewable.io/reviews/paranext/paranext-core/2648#-OzS6x3Y7mM-uYUmwb20:-OzS6x3Y7mM-uYUmwb21:bxf77wl) ([raw file](https://github.com/paranext/paranext-core/blob/b95824d68fd676ff199c5fe9f26e8ae31876c0c0/extensions/src/platform-scripture-editor/src/character-marker-coverage.utils.test.ts)):*
**5\. The reversed-range branch has no test — which is what lets #1 and #2 hide.** Every coverage test uses a forward range, so `!reachedEnd` and the single-segment offset trim are never exercised. A test with `end` before `start` in document order would pin the intended degrade and catch #1/#2.
___
*[`extensions/src/platform-scripture-editor/src/disabled-tooltip-wrapper.component.tsx` line 50 at r1](https://reviewable.io/reviews/paranext/paranext-core/2648#-OzSFKON4xnH5wPHUYhc:-OzSFKON4xnH5wPHUYhd:b-dx88th) ([raw file](https://github.com/paranext/paranext-core/blob/b95824d68fd676ff199c5fe9f26e8ae31876c0c0/extensions/src/platform-scripture-editor/src/disabled-tooltip-wrapper.component.tsx#L50)):*
> ```XML
>       // eslint-disable-next-line react/jsx-props-no-spreading
>       {...props}
>       role={isDisabled ? 'group' : undefined}
> ```

**10\. The disabled-state tooltip trigger uses a role that may not announce its label.** While disabled, the wrapper becomes a focusable `role="group"` named by `aria-label`. `group` is a non-interactive container role, and several screen readers don't announce a container's own label on focus — so the "why is this disabled?" text (the whole point of the wrapper) may be silent. `role="button"` + `aria-disabled="true"` stays focusable and is reliably announced. Please verify on a real screen reader before changing, since this is AT-dependent.
___
*[`extensions/src/platform-scripture-editor/src/character-marker-coverage.utils.test.ts` at r1](https://reviewable.io/reviews/paranext/paranext-core/2648#-OzS71NJ6G9tVOLc4QoC:-OzS71NJ6G9tVOLc4QoD:b-h40fy9) ([raw file](https://github.com/paranext/paranext-core/blob/b95824d68fd676ff199c5fe9f26e8ae31876c0c0/extensions/src/platform-scripture-editor/src/character-marker-coverage.utils.test.ts)):*
**6\. No multi-segment test actually trims the last segment.** Every multi-segment case sets `end.offset` to the last segment's full length, so `slice(0, endOffset)` is always a no-op — the trailing-trim branch runs but never discriminates. One case ending mid-word in a trailing marked run would make it meaningful.
___
*[`extensions/src/platform-scripture-editor/src/localized-strings.test.ts` at r1](https://reviewable.io/reviews/paranext/paranext-core/2648#-OzS79AxD8adiLTQq2gT:-OzS79AxD8adiLTQq2gU:b-6y4uj0) ([raw file](https://github.com/paranext/paranext-core/blob/b95824d68fd676ff199c5fe9f26e8ae31876c0c0/extensions/src/platform-scripture-editor/src/localized-strings.test.ts)):*
**8\. Two new localized keys aren't parity-checked, so an untranslated one could ship.** The control-keys block in the parity test checks a hardcoded 4-of-8 subset rather than the exported `CHARACTER_MARKER_CONTROL_STRING_KEYS`, so `%…_syncEditBlocked_banner%` and `%markerMenu_searchPlaceholder_character%` are never checked for en/es parity — the precise gap this test exists to close. Driving the loop off the exported constant fixes it and stays correct as keys are added.
___
*[`extensions/src/platform-scripture-editor/src/use-character-marker-state.hook.ts` line 207 at r1](https://reviewable.io/reviews/paranext/paranext-core/2648#-OzSFn5w0aSbKsQe6z2r:-OzSFn5w0aSbKsQe6z2s:b3odqkk) ([raw file](https://github.com/paranext/paranext-core/blob/b95824d68fd676ff199c5fe9f26e8ae31876c0c0/extensions/src/platform-scripture-editor/src/use-character-marker-state.hook.ts#L207)):*
> ```XML
>   }
> 
>   const markerMenuItems = useMemo(
> ```

**13\. The menu list is re-filtered and re-sorted on every open/close.** `markerMenuItems` depends on `currentMarker`/`coverage`, which change each open/close, so the `filter + localeCompare` over the block's children re-runs even though the list itself only depends on `blockMarker`/strings/callbacks. Splitting list-generation (memoized on those) from the per-row `isDisabled`/`selectionState` stamping avoids the repeated sort.
___
*[`extensions/src/platform-scripture-editor/src/use-character-marker-state.hook.ts` line 177 at r1](https://reviewable.io/reviews/paranext/paranext-core/2648#-OzSGM_WEjwKUp7g7BNJ:-OzSGM_WEjwKUp7g7BNK:b-uzgwon) ([raw file](https://github.com/paranext/paranext-core/blob/b95824d68fd676ff199c5fe9f26e8ae31876c0c0/extensions/src/platform-scripture-editor/src/use-character-marker-state.hook.ts#L177)):*
> ```XML
>   }, [editorRef]);
> 
>   const coveringMarkers = useMemo(() => Object.keys(coverage?.markerStates ?? {}), [coverage]);
> ```

**15\. `coveringMarkers` memo is redundant.** It re-computes `Object.keys(coverage.markerStates)` that `resolveCurrentMarker` already derives, and only feeds the intentionally-non-memoized `isMixed` block — a plain `const` does the same for less.
___
*[`extensions/src/platform-scripture-editor/src/use-character-marker-state.hook.ts` line 66 at r1](https://reviewable.io/reviews/paranext/paranext-core/2648#-OzSGTRvEG1XBDHBK2dx:-OzSGTRvEG1XBDHBK2dy:b55bdrn) ([raw file](https://github.com/paranext/paranext-core/blob/b95824d68fd676ff199c5fe9f26e8ae31876c0c0/extensions/src/platform-scripture-editor/src/use-character-marker-state.hook.ts#L66)):*
> ```XML
>   if (!selection || !editorUsj) return undefined;
> 
>   const computed = computeCharacterMarkerCoverage(correctEditorUsjVersion(editorUsj), selection);
> ```

**16\. The "no-information" test is duplicated.** `computeCoverage` re-derives by hand what the coverage module already encodes as `EMPTY_COVERAGE`; an exported `isEmptyCoverage` keeps that definition in one place.
___
*[`extensions/src/platform-scripture-editor/src/character-marker-control/character-marker-control.component.tsx` line 143 at r1](https://reviewable.io/reviews/paranext/paranext-core/2648#-OzSGzcS7rDA7dGBe8rF:-OzSGzcS7rDA7dGBe8rG:b-l0i06n) ([raw file](https://github.com/paranext/paranext-core/blob/b95824d68fd676ff199c5fe9f26e8ae31876c0c0/extensions/src/platform-scripture-editor/src/character-marker-control/character-marker-control.component.tsx#L143)):*
> ```XML
>  * default-enabled path the other tests here already exercise.
>  */
> export function CharacterMarkerControl({
> ```

**20\. The gate lives one layer out from the exported control.** `CharacterMarkerControl` is exported (rightly — it has three consumers), but the Simple-mode gate is in `CharacterMarkerToolbar`, so a future wrapper that mounts the control directly loses the "never in Power" guarantee. A short note on the export naming the toolbar as its required host, or a test asserting no bare mount, would protect it.
___
*[`extensions/src/platform-scripture-editor/src/use-character-marker-state.hook.ts` line 128 at r1](https://reviewable.io/reviews/paranext/paranext-core/2648#-OzSGZdkAksa269q5c-6:-OzSGZdkAksa269q5c-7:b-vtf72) ([raw file](https://github.com/paranext/paranext-core/blob/b95824d68fd676ff199c5fe9f26e8ae31876c0c0/extensions/src/platform-scripture-editor/src/use-character-marker-state.hook.ts#L128)):*
> ```XML
>     // `insertMarker` NESTS rather than replaces: picking the covering marker again has no defined
>     // meaning yet (toggle off? extend? — `PT-XXX-B4`), and picking a different one needs a replace
>     // operation that does not exist until `PT-XXX-B2`. Both are correct as behavior and wrong as
> ```

**17\. A few comments point at placeholder tickets.** `PT-XXX-B2/B4` and `PT-XXX-D1/D2` were never filled in, so they reference nothing. Keep the behavioral rationale, drop or substitute the tokens.
___
*[`extensions/src/platform-scripture-editor/src/use-character-marker-state.hook.ts` line 169 at r1](https://reviewable.io/reviews/paranext/paranext-core/2648#-OzS6ncy5YToQQ6__t2o:-OzS6ncy5YToQQ6__t2p:by4dv2d) ([raw file](https://github.com/paranext/paranext-core/blob/b95824d68fd676ff199c5fe9f26e8ae31876c0c0/extensions/src/platform-scripture-editor/src/use-character-marker-state.hook.ts#L169)):*
> ```XML
> 
>   const onOpen = useCallback(() => {
>     setCoverage(computeCoverage(editorRef, getSelection));
> ```

**4\. Coverage never re-syncs while the menu is open.** Coverage is computed once in `onOpen`. If the selection changes while the menu stays open (e.g. an external scroll-group update re-renders the view), the trigger label and per-row check states keep showing the open-time selection until close. If that freeze is intentional, a one-line note at the `setCoverage` site and a small test would lock it in; if not, it's a stale-state bug.
___
*[`extensions/src/platform-scripture-editor/src/character-marker-control/character-marker-control.component.tsx` line 116 at r1](https://reviewable.io/reviews/paranext/paranext-core/2648#-OzSFevP9kUoVF_vFs5S:-OzSFevP9kUoVF_vFs5T:byt8s8b) ([raw file](https://github.com/paranext/paranext-core/blob/b95824d68fd676ff199c5fe9f26e8ae31876c0c0/extensions/src/platform-scripture-editor/src/character-marker-control/character-marker-control.component.tsx#L116)):*
> ```XML
>    * behavior every pre-existing consumer already had.
>    */
>   menuDirection?: 'ltr' | 'rtl';
> ```

**12\. The RTL-safe default is inverted for the future placement wrapper.** `menuDirection` defaults to `undefined`, which resolves to physical `ltr` (the `localStorage` key it reads has no writer in the web view). So an RTL project whose wrapper passes `menuAlign="end"` but forgets `menuDirection` gets a menu that collides off the trailing edge — the dangerous case is the forgot-a-prop default. It's well documented; you might just make the wrapper pass `menuDirection` unconditionally, or default it from the resolved project direction.
___
*[`extensions/src/platform-scripture-editor/src/character-marker-control/character-marker-control.component.tsx` line 329 at r1](https://reviewable.io/reviews/paranext/paranext-core/2648#-OzSGjXo1ve_IX-uUtGx:-OzSGjXo1ve_IX-uUtGy:btsx0ds) ([raw file](https://github.com/paranext/paranext-core/blob/b95824d68fd676ff199c5fe9f26e8ae31876c0c0/extensions/src/platform-scripture-editor/src/character-marker-control/character-marker-control.component.tsx#L329)):*
> ```XML
>         {isTooltipShown && (
>           <TooltipContent>
>             {/* `aria-hidden`: the tooltip is the VISUAL channel only. Its text is already the
> ```

**18\. A few comments are so long the key line gets lost.** The 17-line tooltip block (188-204) bundles four ideas mid-JSX; the `aria-hidden` block (329-335) repeats a principle stated in two other places; and the "verified by grep" instruction (137-141) now matches its own prose. The content is good — it's a distribution problem. Splitting/de-duping these three keeps the load-bearing sentences skimmable.
___
*[`extensions/src/platform-scripture-editor/src/character-marker-control/character-marker-control.component.tsx` line 327 at r1](https://reviewable.io/reviews/paranext/paranext-core/2648#-OzSFB6J7K-EPidHU1BW:-OzSFB6J7K-EPidHU1BX:b-qyqxxe) ([raw file](https://github.com/paranext/paranext-core/blob/b95824d68fd676ff199c5fe9f26e8ae31876c0c0/extensions/src/platform-scripture-editor/src/character-marker-control/character-marker-control.component.tsx#L327)):*
> ```XML
>           </DisabledTooltipWrapper>
>         </TooltipTrigger>
>         {isTooltipShown && (
> ```

**9\. The enabled trigger can advertise a tooltip that isn't mounted.** `useTruncationTooltip` is built for a _controlled_ `<Tooltip open>`, but here the Tooltip is uncontrolled and only `<TooltipContent>` is conditionally rendered. Radix still opens on hover/focus of an enabled, non-truncated trigger and sets `aria-describedby` pointing at content that isn't in the DOM — a dangling reference a screen reader will try to resolve. Either control `open` with the hook's value, or assert `aria-describedby` is absent in that state.
___
*[`extensions/src/platform-scripture-editor/src/use-is-power-mode.hook.ts` line 25 at r1](https://reviewable.io/reviews/paranext/paranext-core/2648#-OzSGrlg8s3BVTiOQi58:-OzSGrlg8s3BVTiOQi59:b9c7jp6) ([raw file](https://github.com/paranext/paranext-core/blob/b95824d68fd676ff199c5fe9f26e8ae31876c0c0/extensions/src/platform-scripture-editor/src/use-is-power-mode.hook.ts#L25)):*
> ```XML
>  *   the setting resolves.
>  */
> export function useIsPowerMode(): boolean | undefined {
> ```

**19\. Two deliberate duplications need real owners, not prose follow-ups.** The second `platform.interfaceMode` reader (`useIsPowerMode` vs `useStructureProtectionState`) and the disabled-tooltip pattern still inline in the paragraph switcher are both fine to defer — but "follow-up" in a comment tends to evaporate; a tracked ticket for each would make the deferral real. (Note: the two interface-mode readers have intentionally different contracts, so there's no divergence bug today.)
___
*[`extensions/src/platform-scripture-editor/src/character-marker-control/character-marker-control.component.tsx` line 268 at r1](https://reviewable.io/reviews/paranext/paranext-core/2648#-OzSGBETBezq264l6SwT:-OzSGBEU7-0j6bYsMY45:b9swown) ([raw file](https://github.com/paranext/paranext-core/blob/b95824d68fd676ff199c5fe9f26e8ae31876c0c0/extensions/src/platform-scripture-editor/src/character-marker-control/character-marker-control.component.tsx#L268)):*
> ```XML
>                   // label below can actually truncate; a shrink-0 button just overflows instead.
>                   // The consumer's class comes last so it still wins on any conflict.
>                   className={['tw:min-w-0 tw:shrink', className].filter(Boolean).join(' ')}
> ```

**14\. Two spots hand-roll `cn()`.** `['tw:…', className].filter(Boolean).join(' ')` appears in both the control and the toolbar; `cn` is already exported from `platform-bible-react` (and imported by this component's own sibling files). Besides being the house helper, `cn`'s tailwind-merge de-dupes conflicting utilities — the exact class-precedence issue the comment at line 267 is worrying about.

jolierabideau (commented):
> @jolierabideau made 7 comments and resolved 14 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2648)* status: 16 of 27 files reviewed, 6 unresolved discussions (waiting on katherinejensen00).

___
*[`extensions/src/platform-scripture-editor/src/character-marker-coverage.utils.ts` line 140 at r1](https://reviewable.io/reviews/paranext/paranext-core/2648#-OzS6KkX0Eg-puc3CWDS:-OzSntR13rMSygkDLUeM:b-qiuu7u) ([raw file](https://github.com/paranext/paranext-core/blob/883391e25e4926d30e344f665ba4cd7ff2c8711d/extensions/src/platform-scripture-editor/src/character-marker-coverage.utils.ts#L140)):*

**1\. Coverage silently treats a backwards selection as a caret.** When a selection inside one text node has its offsets reversed (`start.offset > end.offset`), `slice(start, end)` returns `''`, so `totalLength` is 0 and the collapsed-caret branch reports the enclosing markers as `'all'` instead of measuring the selected characters. If the editor can ever hand you anchor-after-focus offsets, coverage will be wrong for right-to-left drags. Worth confirming whether `platform-editor` normalizes selection direction; if it does, this is moot and a one-line comment saying so would close it. → **Low–Moderate**
</blockquote></details>

`platform-editor` normalizes direction before this code sees it. `$getUsjSelectionFromEditor` takes `start` from focus and `end` from anchor when `selection.isBackward()`, so a right-to-left drag still arrives start-before-end. Taking your escape hatch — added the one-line note so the next reader doesn't re-derive it.
___
*[`extensions/src/platform-scripture-editor/src/character-marker-coverage.utils.ts` line 144 at r1](https://reviewable.io/reviews/paranext/paranext-core/2648#-OzS6SxL3bK5uw8_IIOZ:-OzSnyYQD6kXjqLTmjoy:bbnc6fh) ([raw file](https://github.com/paranext/paranext-core/blob/883391e25e4926d30e344f665ba4cd7ff2c8711d/extensions/src/platform-scripture-editor/src/character-marker-coverage.utils.ts#L144)):*

**2\. The reversed/unreachable-range degrade trims with the wrong node's offset.** The comment says an unreachable end "degrades to the start node alone," but the single-segment branch then does `slice(startOffset, endOffset)` using the _end_ node's offset against the _start_ node's text — an arbitrary prefix, not the whole start node.
</blockquote></details>

You're right the code contradicted its own comment. I couldn't write a test for it, and I'd rather say so than ship a fake one: on the `!reachedEnd` path `segments` is always exactly one node, and for a single node the old and new trims produce identical output in every case — if the old trim yields empty text, `totalLength` is 0 and the caret branch returns the same states the normal path would. Verified for all three ancestor cases. Code-vs-comment defect, not a behavior defect.
___
*[`extensions/src/platform-scripture-editor/src/disabled-tooltip-wrapper.component.tsx` line 50 at r1](https://reviewable.io/reviews/paranext/paranext-core/2648#-OzSFKON4xnH5wPHUYhc:-OzSp6O7EWsDmuc_qixZ:b4kh347) ([raw file](https://github.com/paranext/paranext-core/blob/883391e25e4926d30e344f665ba4cd7ff2c8711d/extensions/src/platform-scripture-editor/src/disabled-tooltip-wrapper.component.tsx#L50)):*

**10\. The disabled-state tooltip trigger uses a role that may not announce its label.** While disabled, the wrapper becomes a focusable `role="group"` named by `aria-label`. `group` is a non-interactive container role, and several screen readers don't announce a container's own label on focus — so the "why is this disabled?" text (the whole point of the wrapper) may be silent. `role="button"` + `aria-disabled="true"` stays focusable and is reliably announced. Please verify on a real screen reader before changing, since this is AT-dependent.
</blockquote></details>

Deferring precisely because you flagged it as AT-dependent. No real screen reader here, and swapping to `role="button"` + `aria-disabled="true"` blind trades a maybe-problem for a definite change. Happy to make it the moment we have a reading from NVDA/JAWS/VoiceOver.
___
*[`extensions/src/platform-scripture-editor/src/localized-strings.test.ts` at r1](https://reviewable.io/reviews/paranext/paranext-core/2648#-OzS79AxD8adiLTQq2gT:-OzSoXucAmjgGdQEhukc:bup9kzc) ([raw file](https://github.com/paranext/paranext-core/blob/883391e25e4926d30e344f665ba4cd7ff2c8711d/extensions/src/platform-scripture-editor/src/localized-strings.test.ts)):*

**8\. Two new localized keys aren't parity-checked, so an untranslated one could ship.** The control-keys block in the parity test checks a hardcoded 4-of-8 subset rather than the exported `CHARACTER_MARKER_CONTROL_STRING_KEYS`, so `%…_syncEditBlocked_banner%` and `%markerMenu_searchPlaceholder_character%` are never checked for en/es parity — the precise gap this test exists to close. Driving the loop off the exported constant fixes it and stays correct as keys are added.
</blockquote></details>

**FIXED** at `:65-82` — with one correction Correction: `%…_syncEditBlocked_banner%` _is_ defined in this extension's `contributions/localizedStrings.json` for both locales — genuinely checkable, genuinely unchecked. That was the real gap. `%markerMenu_searchPlaceholder_character%` is not (it ships in `lib/platform-bible-react/src/localizedStrings.json`), and the two `_format` keys are already covered by the block at `:99-116` that deliberately drops the differs-from-English assertion — so a straight swap would have introduced three false failures. The loop now drives off the exported constant minus a documented three-key exclusion list. One consequence: importing the constant pulled React into this node-environment test and broke collection, so the keys moved to `character-marker-control/character-marker-control.const.ts:45`; the component re-exports, so no consumer changed.
___
*[`extensions/src/platform-scripture-editor/src/use-character-marker-state.hook.ts` line 169 at r1](https://reviewable.io/reviews/paranext/paranext-core/2648#-OzS6ncy5YToQQ6__t2o:-OzSo9Oi93nQFeGbT7KP:b-b479c) ([raw file](https://github.com/paranext/paranext-core/blob/883391e25e4926d30e344f665ba4cd7ff2c8711d/extensions/src/platform-scripture-editor/src/use-character-marker-state.hook.ts#L169)):*

**4\. Coverage never re-syncs while the menu is open.** Coverage is computed once in `onOpen`. If the selection changes while the menu stays open (e.g. an external scroll-group update re-renders the view), the trigger label and per-row check states keep showing the open-time selection until close. If that freeze is intentional, a one-line note at the `setCoverage` site and a small test would lock it in; if not, it's a stale-state bug.
</blockquote></details>

Intentional, documented
___
*[`extensions/src/platform-scripture-editor/src/use-is-power-mode.hook.ts` line 25 at r1](https://reviewable.io/reviews/paranext/paranext-core/2648#-OzSGrlg8s3BVTiOQi58:-OzSpkpYAqWt68ZjcOpf:b-9s1f0z) ([raw file](https://github.com/paranext/paranext-core/blob/883391e25e4926d30e344f665ba4cd7ff2c8711d/extensions/src/platform-scripture-editor/src/use-is-power-mode.hook.ts#L25)):*

**19\. Two deliberate duplications need real owners, not prose follow-ups.** The second `platform.interfaceMode` reader (`useIsPowerMode` vs `useStructureProtectionState`) and the disabled-tooltip pattern still inline in the paragraph switcher are both fine to defer — but "follow-up" in a comment tends to evaporate; a tracked ticket for each would make the deferral real. (Note: the two interface-mode readers have intentionally different contracts, so there's no divergence bug today.)
</blockquote></details>

**DEFERRED** — blocked on ticket numbers Agreed "follow-up" evaporates. But `TODO(PT-XXXX)` would reintroduce exactly the dangling placeholder #17 asked me to remove. Deferred until two tickets exist: (1) consolidating the `platform.interfaceMode` readers, (2) adopting `DisabledTooltipWrapper` in the paragraph switcher.
___
*[`lib/platform-bible-react/src/components/advanced/marker-menu.component.tsx` line 158 at r1](https://reviewable.io/reviews/paranext/paranext-core/2648#-OzSFWb51OpzH_Djp2pM:-OzSpDne4yA3O0_byvYC:b-me3fm0) ([raw file](https://github.com/paranext/paranext-core/blob/883391e25e4926d30e344f665ba4cd7ff2c8711d/lib/platform-bible-react/src/components/advanced/marker-menu.component.tsx#L158)):*

**11\. `aria-checked` coexists with cmdk's own `aria-selected` on every row.** cmdk's `CommandItem` is `role="option"` and already sets `aria-selected` for its keyboard cursor; adding `aria-checked` puts two selection semantics on the same row, and `none → aria-checked="false"` labels ~40 rows "not checked." It's a deliberate, tested choice — I'd just confirm with whoever owns AT/UX that a screen-reader user hearing both is the intended readout, and consider omitting `aria-checked` on `none` rows.
</blockquote></details>

unchanged Your suggestion taken literally: needs the AT/UX owner to confirm the intended readout, including whether `none` rows should omit `aria-checked` rather than announce ~40 rows as "not checked."

katherinejensen00 (approved):
> @katherinejensen00 reviewed 11 files and all commit messages, and resolved 6 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2648)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2649, opened 2026-08-06
Dispose-on-disconnect and typed service-shard discovery (PT-4275)

rolfheij-sil:
Steps 1 and 2 of the multi-window architecture plan (**PT-4275**), now one review unit. Step 1 replaces the reachability probe with authoritative death announcements — dispose-on-disconnect — collapsing "host window died" → "objects declared gone" from 11.1 s to 1–2 ms. Step 2 then stops the main process from finding a window's service implementation by rebuilding the name that window registered under, and finds it by what the registration says it is: typed `objectType` discovery plus a `windowId` attribute, indexed once, with the `*.service-router.ts` / `*.service-shard.ts` vocabulary and its ADRs written down. Between them they delete the probe, the per-close sweeps, `EVENT_NAME_ON_DID_CLOSE_WINDOW`, and the `${name}-${windowId}` string as an unchecked cross-process contract.

Consolidated from #2647 on 2026-08-07 to shrink the review stack; per-part incremental diffs remain viewable:

- Part 1 (step 1, formerly #2647) — https://github.com/paranext/paranext-core/compare/pt-4280-per-window-ui-regression...pt-4275-dispose-on-disconnect
- Part 2 (step 2, this branch) — https://github.com/paranext/paranext-core/compare/pt-4275-dispose-on-disconnect...pt-4275-router-shard-discovery

**Stacked on #2639** (`pt-4280-per-window-ui-regression`). Merge order: **#2621 → #2639 → #2649 → #2651**. Draft until the stack below merges.

---

## Part 1 — dispose-on-disconnect (step 1, formerly #2647)

Step 1 of the multi-window architecture plan (PT-4275): replace the reachability probe with authoritative death announcements — dispose-on-disconnect.

**Design** (per the lead dev's plan, followed exactly with one documented deviation): `rpc-server`'s `onWebSocketClose` already deletes the departing client's methods; it now hands that list to a local event (parsing nothing), exposed through the `createRpcHandler` seam and re-exported from `network.service`, and main's `network-object.service` filters it into the existing `object:onDidDisposeNetworkObject` events that every process already handles. The announcement is emitted after the teardown it describes, so it structurally cannot race it — the race the old close-announcement design carried (and we proved live) is gone by construction.

**Deleted:** the reachability probe (`forgetUnreachableRemoteObjects` and its public surface), the three per-close sweeps, the bounded re-sweep mitigation, and `EVENT_NAME_ON_DID_CLOSE_WINDOW` end-to-end. The app-global takeovers (theme, scroll group) now ride dispose events alone.

**Measured** (live, two windows, closing the hosting window):

- host death → objects declared gone: **11.1 s → 1–2 ms**
- host death → both singletons answering again: **~9.2 s → ~1.0 s** (residual is the driver's poll cadence; re-registration itself is 50–86 ms)
- window reload: survivor uninterrupted; reloaded window recovers

**Also:** the network-object hooks (`useDataProvider` et al.) now recover after a provider disposal instead of latching `undefined` forever; `usePromise` surfaces factory rejections; `PlatformEventEmitter.emitIsolated` (`@experimental`) isolates subscriber throws at every hop of the new chain; a single-source event name is test-proven re-registrable after socket cleanup (a question the plan left open).

Adversarially reviewed (2 independent passes) with a consolidated 14-item fix round; all findings addressed or explicitly documented. Full battery: unit suites green across workspaces, multi-window e2e 6/6 including quit, typecheck/lint clean, `papi.d.ts` regenerated (2 items removed, additions all `@experimental` and platform-internal-documented).

---

## Part 2 — typed service-shard discovery and router/shard naming (step 2, this branch)

Step 2 of the multi-window architecture plan: the main process stops finding a window's service implementation by rebuilding the name that window registered under, and finds it by what the registration says it is.

**Typed `objectType` discovery replaces name parsing.** Every window-scoped service registers with an object type of its own (`'webViewServiceShard'`, `'notificationServiceShard'`, `'windowServiceShard'`) and a `windowId` attribute — `src/shared/models/service-shard.model.ts`. `createServiceShardIndex` (`src/main/services/service-shard-index.ts`) subscribes once to the network object create/dispose announcements, filters on that type, and keeps a `windowId → shard` map. Window readiness listens to the same index instead of parsing a window id back out of a network object id. The scoped registration names are unchanged; only discovery moved.

That removes the `${name}-${windowId}` string as an unchecked cross-process contract from the router side, removes the `as` assertions that defeated the typed provider-name system, and stops a router paying the network service's registration retry to learn that a window it asked about was never there.

**Router / shard vocabulary, per the architecture plan.** Main-process files that claim a generic name and forward became `*.service-router.ts`; renderer files that implement one window's copy became `*.service-shard.ts`, and their start functions were renamed to match. The pattern, its rules, and its exceptions are written down in `.context/standards/Architecture.md` § "Service router and service shard", with `adr-shard-discovery-by-type` in `.context/standards/Architecture-Decisions.md`.

**Behavior is preserved.** The window service router's relay is byte-identical after normalizing the rename. No test was dropped; the shard-index fixture announces shards under a name deliberately unrelated to the service name, so a router that rebuilt a scoped name would now fail its tests.

### Factory deferral

The plan's §5.3 asks for one shared router factory. It was built, measured, and then deferred — the plan's author explicitly invited disagreement on this step, so here is the measurement rather than an assertion.

| | lines |
|---|---|
| factory + its test | **+271** |
| duplication it removed | ~25 |
| forwarding methods it generated across the whole codebase | **1** (notification's `send`) |

The WebView router passed an empty forwarded list and used the factory purely as a registration wrapper; the window router cannot use it at all (it is a data provider); the command router registers request handlers rather than a network object.

The type argument did not hold up either. A plain `const router: ServiceType = { ... }` annotation already made TypeScript name any member a router failed to publish, for free. Assembling the object at runtime from two halves destroys that — hence `as unknown as T` — and the conditional-type coverage guard put in its place was measurably **weaker** than the annotation it replaced: `{ alpha: undefined, beta: undefined }` compiled clean, producing a router with zero callable members that registers nothing and fails at runtime with "no handler". The factory also needed three eslint suppressions; the annotation needs none.

So both routers are plain typed literals plus `networkObjectService.set`, and the one genuine deduplication stays as `createTargetShardResolver` (`src/main/services/target-shard-resolver.util.ts`). Typed discovery — the index, the model, the object-type and attribute registrations, the readiness wiring — is untouched; that is what this step was for.

Worth building when there are three or more genuine forwarding routers to design it against, which is what moving the renderer-hosted commands into per-service routers is expected to produce. Designing it against n=1 produced the shape above.

### Please look at: fan-outs now refuse to answer for a window they could not ask

The fan-outs ask only windows that have registered their services, but they used to answer as though the windows they skipped were not there — a window with a dozen editors open came back byte-identical to a window that does not exist. At quit that meant the shutdown sync selected nothing, found no projects, and logged a debug-only `skipped`: the quietest line in the log on the run most likely to have dropped a user's unsynced work.

A tracked-but-unready window is now reported as a window that could not be asked. `getAllOpenWebViewDefinitions` refuses to answer rather than pass off an incomplete list, the shutdown sync reports `partial`, and an operation naming a web view no longer falls back to the focused window when the owner was the window nothing asked.

**The cost, observed live:** during startup there is always a window that has not registered yet, so a cold start now produces a `warn` in main and, because `platform-scripture-editor`'s default active project picker calls `getAllOpenWebViewDefinitions` at activation, a `warn` from the extension host whose own comment reserves that path for "PAPI plumbing bugs". Nothing breaks — the picker is idempotent and re-runs on the next web view event, and the app comes up correctly — but it is per-startup noise on a healthy run, which is the opposite of what making the log trustworthy was for. Flagging it explicitly: the honest-reporting change is right for the shutdown path, and whether the startup window deserves the same volume is worth a second opinion.

### Verification

- `npm run typecheck` clean; `npm test` green across all workspaces; `npm run format:check` clean; eslint clean on all 79 changed lintable files (root + per-extension).
- `npm run build:types` regenerates `lib/papi-dts/papi.d.ts` with an empty diff. This part's own `papi.d.ts` diff against Part 1 is prose/TSDoc only — six comment hunks, no signature change, no new exported symbol.
- e2e `--project=isolated e2e-tests/tests/isolated/multi-window`: 6 passed.
- Live smoke over PAPI: app starts, `platform.createWindow` opens a second window, both windows register their window-scoped services and their WebView service shard, and the router's fan-out answers with definitions from both (6 + 6) while each shard also answers for itself.

## Deferred, with owners

- **PT-4281 — the blur-aware focus guard (`isApplicationFocused()`).** The raise gate on this PR
  reads `getFocusedWindowId() !== undefined`. That flag latches on the first focus and nothing
  clears it, because no blur handler exists, so it establishes only that some window has been
  focused at least once — it cannot tell whether this app holds focus at the moment of a raise. The
  docs and the test name on this PR now say exactly that, rather than claiming a guarantee the code
  does not provide.

  The real guard is **already implemented in PT-4281 (#2670)**, which stacks directly on this work:
  a `blur` handler in `main.ts` feeds `isApplicationFocused()`, both raise paths consult it, and
  `web-view.service-router.move-raise-guard.test.ts` exercises it in both states. Once that merges,
  the stronger claim becomes true and the wording here can be tightened to match.

- **PT-4281 — the deep-link raise.** Not extended here — **delivered on PT-4281 (#2670)**, per the
  ruling that it be wired through `focusWindow` in one review with the guard. `handleUri` now
  raises through the service rather than calling `BrowserWindow` methods directly, which is what
  earns it the flash-on-denial behaviour and makes a raise a no-op rather than a throw on a window
  that has closed.

  It is deliberately **not** gated on `isApplicationFocused()` the way the two in-app raise paths
  are. That gate exists to stop the app pulling itself in front of whatever the user is working in;
  a deep link runs precisely when the app does not own the foreground, so the gate would suppress
  every raise the path exists to perform. The user asked for that window by following the link.

---

AI-assisted — [session](https://claude.ai/code/session_01Lf3XmoA9StqLNDiMBqZVTy)

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2649)


src/main/services/web-view.service-router.ts
```diff
@@ -0,0 +1,433 @@
...
+// Router methods that route to the focused window's WebView service shard
+
+async function openWebView(
+  webViewType: WebViewType,
+  layout?: Layout,
+  options?: OpenWebViewOptions,
+): Promise<WebViewId | undefined> {
+  // If an existingId is provided, search all windows for the webview's owner
+  if (options?.existingId) {
+    const owner = await findOwner(options.existingId, 'openWebView');
```
rolfheij-sil: 🤖 Claude: One thing this line does not cover, and it decides whether "one instance for the whole app" is what the code actually does: the `existingId: '?'` wildcard never reaches a merged list.

Traced end to end at this commit:

1. A truthy `existingId` sends the literal `'?'` into `findOwner` (this line).
2. `findOwner` asks each ready window `getOpenWebViewDefinition('?')` (:130).
3. The shard resolves by exact id — `getWebViewTabInfoById` in `platform-dock-layout-storage.util.ts`, no wildcard handling — so every window answers "no".
4. `findOwner` returns undefined and the call falls through to the focused window (:238).
5. The wildcard is honoured only there, inside that one shard (`src/renderer/services/web-view.service-shard.ts:2153-2160`, via `findFirstWebViewDefinitionByType`).

So every view that uses the wildcard rather than a concrete id can exist once per window today: Manage Books (`extensions/src/platform-scripture/src/main.ts:250,265,306` — its own comment says "one dialog at a time"), the singleton Column-3 comment panel (`legacy-comment-manager/src/main.ts:160,196`), Home and New Tab (`platform-get-resources/src/main.ts:237,254`), the registration views (`paratext-registration/src/main.ts:16,24`), and the Simple-mode layout receiver (`platform-scripture-editor/src/shared-layout-receiver.model.ts:50`). The call sites that read `getAllOpenWebViewDefinitions()` are unaffected — those do consult the merged list.

Fix shape we would take, routing-only and with no new surface: special-case `'?'` in `openWebView` — read `getAllOpenWebViewDefinitionsWithReachability()` (already in this file, :282), take the first definition of that `webViewType`, then `findOwner` its concrete id and route there. No new shard method, no new `papi.d.ts` symbol, no new experimental-marker decision, and it composes with the layout-target routing at :231-235 and the cross-window raise at :201-212.

Deliberately not built yet, because it is your call. It is one commit either way.

Questions:
(a) Is app-wide resolution of the `'?'` wildcard the semantics you want here?
(b) If yes, does it land on this PR or on #2651? We would rather put it on #2651: changing this file here restacks #2651's 56 commits, and #2651 edits this same file.

rolfheij-sil:
🤖 Claude: For the reviewer — open questions on this PR (plus the stack-wide ones, gathered here since this is the first architecture PR in the walk)

**1. `existingId: '?'` resolves inside one window.** See the inline thread at `src/main/services/web-view.service-router.ts:223`. The wildcard never reaches a merged list: `findOwner` asks each window for the literal id `'?'`, every window answers no, and the call falls through to the focused window where the shard honours the wildcard locally. Manage Books, the singleton comment panel, Home, New Tab, the registration views and the Simple-mode layout receiver can therefore exist once per window today. The fix is routing-only with no new surface. Questions are in the thread: is app-wide resolution the intended semantics, and does the fix land here or on #2651?

**2. Committing the design document for the window-scoped reachability read** (the shape we worked up for the round-4 review's proposed fix D on #2621, https://github.com/paranext/paranext-core/pull/2621#pullrequestreview-4877614064). That read is not being added, so the document that designed it is the record of a road not taken; we are keeping it as a dated document in the PRD folder outside the repo rather than committing it to `.context/designs/`, on the reasoning that `.context/` is for decisions that constrain the code that exists (and our standing convention keeps feature design docs in the PRD folder). Nothing is lost: both of its durable outputs are already in the repo — the predicate fix in this PR (a window that could not be asked is now reported as such rather than as one that answered nothing, `web-view.service-router.ts:99-150`) and the re-classification of the mechanical `getAllOpenWebViewDefinitions()` call sites as correct under app-wide uniqueness. Question: if you would rather have it in-repo anyway, say so and it goes to `.context/designs/` with a header recording the declination.

**3. Scoped network-object ids — settled, no question here.** The `${name}-${windowId}` ids stay as they are rather than becoming opaque, and making them opaque — the follow-up we had flagged as possibly needed — is confirmed unnecessary. No work planned.

**Current state notes on this PR** (facts, for reading the diff)

- `openWebView` routes by the layout's `targetTabId` (`panel` and `replace-tab`) through the same `findOwner` the `existingId` path uses — after `existingId`, before focus (`web-view.service-router.ts:216-239`).
- A layout target no window can be asked about does not fail the open: it warns and falls through to the focused window (`:170-186`). The `existingId` search still throws in that situation, because guessing wrong there mints a second copy of a view meant to be unique, while guessing wrong on placement costs placement only.
- A panel whose target tab is nowhere is now placed as a layout with no target would be, rather than throwing (`src/renderer/components/docking/platform-dock-layout-storage.util.ts:1010-1030`); `replace-tab` still refuses, since replacing a tab that is not there means nothing.
- A cross-window open raises the owning `BrowserWindow` (`web-view.service-router.ts:201-212`) — only when something actually opened, only when the open crossed windows, and only while this app already holds focus, so it can never take focus from another application. The renderer-initiated `shouldBringToFront` path is deliberately not wired, documented at the call site.
- `set()` refuses a network object id that is another registered id followed by a dot, in both directions, scoped to locally hosted registrations (`src/shared/services/network-object.service.ts:226,689-692`).
- The `render-process-gone` handler's comment names the signal a crashed window actually recovers through, `onDidRegisterWindowServiceShard` (`src/main/main.ts:677-680`).
- The disconnect-announcement ordering test you asked for is landed in `a43c30f39a7`, hardened in `ac494303467` (`src/main/services/__tests__/rpc-disconnect-announcement-ordering.test.ts`): it pins that no macrotask or I/O turn separates a client socket's close from the announcement's `ws.send` — the corrected invariant, since the path does take one microtask hop through an already-resolved `initialize()`.
- Experimental markers: six shards carry them (web-view, window, notification via the shared docs const, dialog, Usersnap, BookChapterControl), as do the three routers added on #2651 (dialog, Usersnap, BookChapterControl) and the two main-process hosts moved there (theme, scroll group) — those five live at #2651's head, not this PR's. Dialog is object-level on the shard and per-handler on the router. The eight pre-existing published command names keep their published flags, pinned by tests.

**Per-part diffs:** the two compare links at the top of the PR body split this branch into step 1 and step 2.

rolfheij-sil:
🤖 Claude: One environment note before you walk the router: the cross-window raise (`openWebViewInOwningWindow` → `focusWindow`) is implemented, fires on the right conditions, and is **verified working on native Windows** (live repro today: opening a view owned by a background window brought that window forward). The same repro on **WSLg does nothing visible** — traced end to end: the gates pass and `window.focus()` is called cleanly, but WSLg's compositor refuses client-initiated activation (it keeps `_NET_ACTIVE_WINDOW` permanently null and no focus event ever returns). So if you test on a WSLg dev setup and the window doesn't raise, it's the platform, not the code. We're considering `flashFrame(true)` alongside the raise as graceful degradation for denied activations — opinion welcome.

tjcouch-sil:
Answers to your three questions, one ruling reversal, and two confirmations.

**1. `existingId: '?'` — app-wide, and land the fix on this PR.** Multiple windows don't change
the fact that there is one app: `'?'` searches for one web view in that one app, so per-window
resolution is a bug under the same app-wide-uniqueness ruling as everything else. Implementation
is your choice — fan out asking each shard, or answer it from whatever the router can already see —
as long as the result is app-wide. It belongs on this PR with the other routing fixes.

**2. The FIX-D design doc — agreed, keep it out of the repo.** It's the design for a declined API,
not overarching principles; a road-not-taken doc in the repo just goes stale. Your convention
argument (`.context/` constrains code that exists; feature designs live in the PRD folder) is
right, and both durable outputs already landed. My earlier "commit it" predates the read being
declined — withdrawn.

**3. Scoped ids — confirmed settled.** Non-opaque, experimental-marked, no work planned.

**Ruling reversal: `TabLayout.parentTabGroupId` must route.** I read the carve-out comment and I'm
overriding it. "The only call site today is in-window by construction" doesn't hold across the
extension ecosystem — extension developers can pass `parentTabGroupId` right now, and a router
that honors every layout field except one is exactly the bug that surfaces randomly a year from
now and confuses everyone. The API needs to work. The fix is the dock-containment shard query
("does this window's dock contain tab/tab-group X") — on the shard's wire surface, off the public
interface, covered by the object-level experimental marker like the rest. Building it also closes
a second gap in what shipped: `findLayoutTargetOwner` resolves via the web-view owner lookup, so a
`targetTabId` naming a non-web-view tab currently falls through to the focused window too. One
query covers both.

**Confirmed: the cross-window raise fires on found-existing, which is the case I care about.**
"Open Comments List" from window 2 when the list lives in window 1 must shift OS focus to
window 1 even though nothing new opened — traced it, the shard returns the existing id after
raising the tab, so the router's raise condition fires. Keep that property.

**flashFrame: yes, ship it as the degradation for denied activations** — the auto-cancel on focus
makes it near-invisible on successful raises in theory. Please verify by hand on native Windows
that the blink isn't visually annoying before it merges.

---

src/main/services/web-view.service-router.ts
rolfheij-sil: 🤖 Claude: Fixed on this PR, taking your option (a) — fan out and ask each shard — because the router has nothing to answer from: `service-shard-index.ts` deliberately keeps only network object ids and no definition cache, so option (b) costs *two* fan-outs (a merged read to find a definition of the right type, then `findOwner` to learn whose window it is) and the merged read throws on partial coverage. (a) needed no new shard method: `findOwner` now takes an `OwnerMatcher` — `{ kind: 'id' }` or `{ kind: 'type' }` — and `openWebView` passes the type matcher when `existingId === '?'`. Both kinds share the reachability accounting, the logging and the `WebViewOwner` shape.

Correcting what we told you: `'?'` resolving per-window was only half of it. `findOwner` threw "some windows were unreachable" whenever any window could not be asked, and `openWebView`'s `existingId` branch never caught it — so while a second window was still starting, every `'?'` open in the app failed outright rather than falling through. Generalising the matcher fixed the resolution; the throw needed its own call, and it is now per-caller: a call that would create still refuses to guess (opening blind is what mints the second copy of a view meant to be unique), while `createNewIfNotFound: false` gets `undefined` back — which is true, and costs a probe nothing. `getOpenWebViewDefinition` now short-circuits on a found owner before the unreachable check too, matching the other two paths.

Two traps building it surfaced, both fixed here:

- **Window creation order.** `getReadyWindowIds()` is creation order, so taking the first match would have preferred an old background window over the one the user is in. `findOwner` now waits for every window to answer, then prefers the routing-target window and falls back to lowest window id — deterministic whichever shard replies first.
- **The raise ignored `bringToFront`.** `openWebViewInOwningWindow` raised whenever the shard returned an id, and `legacy-comment-manager` runs `{ existingId: '?', createNewIfNotFound: false, bringToFront: false }` as a passive probe. The moment `'?'` started routing app-wide, that probe would have begun stealing OS focus to another window. The raise is now gated on `bringToFront !== false`, and `OpenWebViewOptions.bringToFront`'s TSDoc says so.

rolfheij-sil:
🤖 Claude: All four off-thread items are done — question 1 is answered in its own thread on `web-view.service-router.ts`. One of them needs a decision from you: the reasoning behind `flashFrame` does not hold, details at the bottom.

**Question 2 — the FIX-D design doc.** Agreed, it stays out of the repo, and there is no work. The earlier round-trip came from our sequencing, not from you.

**Question 3 — scoped ids.** Confirmed settled. No work planned.

**Ruling reversal — `parentTabGroupId` must route.** You were right, and our carve-out's premise was false rather than merely thin. The "+" click is not in-window by construction: it sends a command to the extension host, which calls `papi.webViews.openWebView`, which re-enters the router, which resolves the target window at that *later* instant — and on a miss, `addTabToDock` placed the tab without the group it was given and logged nothing at all.

Implemented as you specified. `WebViewServiceShard` (new `src/shared/models/web-view.service-shard.model.ts`) extends the public `WebViewServiceType` with `dockContainsTab`, so the query is on the shard's wire surface, off the public interface, and covered by the shard registration's object-level `x-experimental`. `papi.d.ts` gains exactly one member — `PapiDockLayout.containsTab`, tagged `@experimental`. `findLayoutTargetOwner` now asks dock containment instead of web-view ownership, which closes your second gap in the same query: a `targetTabId` naming a settings, about or error tab used to resolve as unowned and fall through (and for `replace-tab`, fail the open outright in whichever window it landed in). It is latent, not live — no in-repo caller passes a non-web-view `targetTabId` today — but it is one query either way. `addTabToDock` now warns when it falls through.

One fact came out of building it that changes what the routing can promise: **rc-dock mints tab-group ids per renderer process** — a module-level counter in `Algorithm.js` that every window starts at zero and increments independently — so several windows routinely hold the same tab-group id. It is not app-unique. Routing by it can therefore only ever *prefer* the right window, never identify it, which is why the routing-target preference in `findLayoutTargetOwner` is load-bearing here rather than a tie-break. That is now written down in `getLayoutTargetTabId`'s rationale.

**Confirmation — the cross-window raise on found-existing.** Property preserved, and now genuinely pinned. The existing test could not have caught a regression: its fake shard's `openWebView` always resolved the placeholder `'opened'` instead of the existing id, so a well-meaning "raise only when something new opened" gate would have left every test green. The fake now returns the found id and the test asserts on it. The raise also picked up a `bringToFront` gate (see question 1's thread) so a passive probe cannot steal focus; nothing changes for a user-initiated open.

**Confirmation — `flashFrame`.** Shipped, but the reasoning you approved it on does not hold, and you should have that before it merges:

- Electron 39.8.8 documents no auto-cancel on focus. `flashFrame(flag)` is only "starts or stops flashing the window"; whether the taskbar button settles on activation is undocumented platform behavior, not a guarantee.
- The raise is already gated on `getFocusedWindowId() !== undefined` — this app already owns the foreground when it runs — which is exactly the Win32 condition under which activation is *permitted*. So the flash fires overwhelmingly on **successful** raises, not denied ones. The premise is closer to inverted than to merely unverified.
- There is no denial signal to gate on: `BrowserWindow.focus()` returns void and Electron emits no event for a refusal. The only alternative is a poll-with-timeout heuristic that races the user clicking elsewhere. So it is unconditional or nothing.

It is in unconditionally, in `window-state.service.ts:focusWindow`, with that reasoning in the comment, and I will verify it by hand on native Windows before it merges as you asked — with the caveat that the successful-raise case is the one I can reliably reproduce.

**One ask.** The path where denial genuinely is likely — the deep-link raise in `main.ts`, which fires when the app may not own the foreground and says so in its own comment — calls `focus()` on a raw `BrowserWindow` and never goes through `focusWindow`, so it gets none of this. Do you want it extended there? We left it alone rather than deciding unilaterally, since it is outside this PR's scope.

rolfheij-sil:
🤖 Claude: Correcting myself on `flashFrame` before this merges — one of the three grounds I gave you was wrong, and it was the load-bearing one.

I told you the raise is "already gated on `getFocusedWindowId() !== undefined` — this app already owns the foreground when it runs", and concluded the flash fires overwhelmingly on successful raises, so your premise was "closer to inverted than to merely unverified."

That gate does not mean what I said it means. `setFocusedWindowId` has exactly one caller — a `focus` handler on each window — and `focusedWindowId` is cleared only when that specific window is removed. **There is no `blur` handler anywhere in `src/main`.** So the value is set the first time any window is focused and stays set for the rest of the session, whether or not Platform.Bible owns the OS foreground. The gate says "a window has been focused at some point", not "we own the foreground now".

So denial is genuinely reachable, and your reasoning stands better than my refutation did. The concrete path: an extension re-opens a web view by id while the user is typing in another application — `openWebView` resolves the owning window, and we run `restore()` + `focus()` + `flashFrame(true)` against a foreground we do not hold. Windows refuses the activation, and the flash is the only thing the user gets. That is exactly the case you described.

That same finding is also a defect in its own right: the guard is commented as what stops us stealing focus from another application, and it does not do that. We can and do yank the app in front of whatever the user is working in. I am treating that as a separate fix.

**The hand check, and what it actually showed.** Rolf ran it on native Windows. The raised window came to the front **and** its taskbar button flashed about five times before stopping on its own. I had predicted the flash would be invisible on a successful raise; it is not, and the code explains why: `focus()` runs on the line *before* `flashFrame(true)`, and `flashFrame(false)` is never called anywhere in the repo. Activation happens first, so there is no later activation event to auto-cancel the flash and nothing ever stops it explicitly — the taskbar just flashes out its count.

**Which suggests a better fix than either of us proposed:** call `flashFrame(true)` *before* `focus()`. Then a successful activation cancels the flash itself, and only a denied one keeps flashing. Silent on the common path, informative on exactly the path you added it for. I would like to make that change unless you object — it is a one-line reorder and it makes the degradation do what you wanted rather than what we shipped.

The two other grounds I gave you are unaffected and still hold: Electron documents no auto-cancel behaviour we can rely on, and there is still no denial signal to gate on — `BrowserWindow.focus()` returns void and no event follows a refusal. So it remains unconditional-or-nothing, and I now think unconditional is right.

Sorry for the noise on this one. The original analysis was confident and wrong in a way a five-minute read of the focus plumbing would have caught, and it is better you have it before the merge than after.

rolfheij-sil:
🤖 Claude: A second correction on `flashFrame`, and this one retires the rest of my argument. I gave you three grounds. Two were wrong.

**Ground two** I already retracted above: the raise is not gated on our owning the foreground, because `setFocusedWindowId` has one caller and nothing clears it on blur.

**Ground three was also wrong.** I told you "there is no denial signal to gate on — `BrowserWindow.focus()` returns void and Electron emits no event for a refusal." There is one. `isFocused()`, read immediately after `focus()`, distinguishes the two.

Only ground one survives: Electron documents no auto-cancel we can rely on. That one is now proven rather than argued — see variant B below.

**What the hand check actually showed.** Rolf ran three variants on native Windows:

| Variant | Code | Result on a successful raise |
|---|---|---|
| A — what we shipped | `focus()` then `flashFrame(true)` | window came forward **and flashed ~5×** |
| B | `flashFrame(true)` then `focus()` | **still ~5 flashes** — Windows does not cancel a flash on activation |
| C | `flashFrame(true)`, `focus()`, then `flashFrame(false)` if `isFocused()` | **no flash at all** |

A disproves my prediction that the flash would be invisible on a successful raise — it is not, and `flashFrame(false)` is never called anywhere in the repo, so nothing stops it. B disproves the idea that ordering alone could fix it. C works.

**So the shipped behaviour now does what you asked for.** `focusWindow` flashes, focuses, and cancels the flash when `isFocused()` reports the activation landed. Silent on the common path; flashing only where Windows genuinely refused and the flash is the whole signal the user gets. The failure mode is benign: if `isFocused()` ever reads false on a success — another platform, a later Electron — we flash, which is merely the previous behaviour.

**A related defect worth knowing about, with an owner.** The guard I cited is commented as what stops us stealing focus from another application, and it cannot do that: it is true from the first time any window is focused until that window closes. So an extension re-opening a web view by id while the user is typing elsewhere could pull the app in front of them. That is not fixed here — PT-4281 adds a blur-aware `isApplicationFocused()` to the same module, which makes the guard real. The two compose: its guard decides whether to raise, this decides what the raise does afterwards.

**One question I did not decide for you.** `PapiNetworkEventEmitter.emitIsolated` ran `assertNotDisposed()` and the network send outside the isolation, so a throwing send aborted the fan-out before any local subscriber was told — the failure mode `emitIsolated` exists to prevent. The send is now isolated and its failure logged. But the method's TSDoc already promises that a failure to reach the network is reported where the network callback was supplied, so a `networkSubscriber` that throws synchronously has broken its own contract. Should that surface to the caller — perhaps through `handleSubscriberError` with a sentinel index — or is a `logger.error` enough, given the emitter does not know its own event type and so logs less specifically than the network callback would? Your call; I did not want to invent a channel.

Twice wrong on the same feature is not a good record, and both times the correction came from running something rather than reasoning about it. The parts of your original instinct I argued against were the parts that held.

tjcouch-sil:
Verified this round end to end: every ruling landed correctly, and several landed better than
asked — the generalized `OwnerMatcher` with deterministic multi-claimant resolution, the
`bringToFront` gate, the falsifiability fix on the raise test, and the three-variant `flashFrame`
study. That empirical work is exactly how I want disagreements like that settled; variant C is
approved as shipped, Windows-only empirics acknowledged.

Two asks and two decisions:

**1. The focus-steal claims are still shipped as truth, and the real fix has no durable home.**
Your own correction established that `getFocusedWindowId() !== undefined` cannot prevent stealing
focus from another application — no blur handler exists, the value latches on first focus. But
three places still assert that it does: `openWebViewInOwningWindow`'s TSDoc ("never take it from
another application"), the **public** `bringToFront` TSDoc that lands in `papi.d.ts` ("never
stealing focus from outside the app"), and the test named `'does not take focus from another
application'`. Please make all three honest now — the gate prevents raising before any window has
ever been focused, nothing more — and record the deferral durably: the blur-aware
`isApplicationFocused()` plan currently lives only in a PR comment; it needs to be in PT-4281's
Jira description and this PR's body. A false claim on the extension-facing API surface is the one
kind of doc bug we can't leave to memory.

**2. `getOpenWebViewDefinitionsForWindow` contradicts its own `@throws` contract for a
once-ready window.** The doc says it throws when a window "was serving requests but could not be
asked," but the guard is `isWindowReady`, and a crashed/abandoned window has been marked
not-ready — so the shard-missing path returns `[]` instead of throwing. Consumer impact: in
`shutdown-tasks.ts`, closing a window whose renderer died makes the per-window closing sync
silently select nothing, with no warning — while the quit path reports the same situation
correctly via `abandonedWindowIds`. Either align the code with the contract (report the
ever-ready-but-unaskable case) or fix the contract and make the closing-sync reader warn; silent
empty on a data-adjacent path is the one option that isn't fine.

**Decision — `emitIsolated`:** `logger.error` is enough; keep the behavior you shipped and
pinned. A network callback that throws synchronously has broken its own reporting contract, and
inventing a sentinel-index channel to re-report it would grow the emitter's API to serve a
caller bug. If a real consumer someday needs the signal, design it then.

**Decision — the deep-link raise:** don't extend it in this PR. Wire it through `focusWindow` as
part of PT-4281, alongside the `isApplicationFocused()` guard — the deep link is precisely the
may-not-own-the-foreground path, so it gets flash-on-denial for free the moment the real guard
exists, and the two changes belong in one review. Please note it in PT-4281's description with
the rest.

rolfheij-sil:
🤖 Claude: All four items are landed, one of them differently than you asked — and one of your premises has changed since you wrote.

**1. The focus-steal claims are now honest.** All three said the gate prevents stealing focus from another application; none of them could. Fixed at source in `web-view.model.ts` and regenerated into `papi.d.ts` (the public one you singled out), in the raise doc on `openWebViewInOwningWindow`, and in the test, now named `'does not raise before any window has been focused'` — which is what it actually pins. The docs now describe what the gate does establish: no raise fires before some window has been focused, which is what keeps one from happening during startup. Nothing claims more than that.

**2. `isApplicationFocused()` is not a plan in a PR comment — it is built, on #2670.** This is the premise that moved. PT-4281 ships a `blur` handler in `main.ts` feeding `isApplicationFocused()`, the guard on both raise paths in the router, and `web-view.service-router.move-raise-guard.test.ts` exercising it true and false. I verified that against #2670's tree and confirmed it with the session that owns it. So the durable home you asked for is a merged PR rather than a Jira line. #2649's body records the split explicitly: this PR's gate is weak, the real guard lands next, and the wording here tightens when it does.

**3. The deep-link raise — delivered on PT-4281 (#2670).**

> _Edited after posting._ This item originally said the work was outstanding. That was true when I wrote it and stopped being true shortly afterwards, in `2cd6460ac24`. Correcting the comment rather than leaving you to act on a stale one. (The version before that said it was already done, which was my error — I had matched `focusWindow(` inside `raiseMoveTarget`, the move-target raise, and never checked which binding it resolved to. `handleUri` did at that point declare a local `const focusWindow = …` that shadowed the imported service.)

`handleUri` now raises through the `focusWindow` service instead of calling `BrowserWindow` methods directly. That is what earns it the flash-on-denial behaviour, and it makes a raise a no-op rather than a throw on a window that has already closed — so a raise can never drop the URI before it is dispatched.

It is deliberately **not** gated on `isApplicationFocused()` the way the two in-app raise paths are. That gate exists to stop the app pulling itself in front of whatever the user is working in; a deep link runs precisely when the app does not own the foreground, so the gate would suppress every raise the path exists to perform. The user asked for that window by following the link.

**4. `getOpenWebViewDefinitionsForWindow` now matches its contract.** Took the align-the-code option. The guard was `isWindowReady`; it is now `wasWindowEverReady`, so a window that was serving and has since died reports that it could not be asked, while a never-ready window still answers none. The consumer needed no change at all — `shutdown-tasks.ts` already had the loud warning for this case; it was simply unreachable for a dead renderer, which is your "silent empty on a data-adjacent path" arriving as a no-op. The regression test is mutation-verified: reverting the guard turns exactly that one test red and nothing else.

**5. `emitIsolated` — kept as shipped**, per your decision. No sentinel-index channel.

**Two things to know before you re-read.** This branch was rebased onto main (18 commits) after it went CONFLICTING; CI is green on all three platforms. And main merged its own ADR-0013 through ADR-0018 while the stack was open, so ours renumbered: **0013→0019** here, **0014/0015/0016→0020/0021/0022** on #2651, and **0017→0023** on #2670. Your "amend ADR-0016" therefore means **ADR-0022** — see #2651 for that one.

tjcouch-sil:
One item before merge: **the PR body cites pre-renumber ADR numbers** ("ADR-0011" in the Part-2
prose — the stack's entries are now ADR-0026 and up after main claimed the intervening numbers).
This matters precisely because we squash-merge: the PR body becomes the permanent commit message
on `main`, so merging as-is bakes citations that point at unrelated entries into history. Two-minute
body edit; the repo files themselves are already consistent.

rolfheij-sil:
🤖 Claude: Fixed, and the stack moved under it twice while I was doing so — the numbers in the body are not the ones you quoted any more.

**What you flagged.** The Part-2 prose cited ADR-0011 for this branch's entry. You were right that it matters at squash-merge, and it was worse than a dangling reference: main had claimed 0011 in the meantime, so the citation resolved to a real but unrelated decision (the character-marker entry). A reader following it would have landed somewhere plausible and wrong.

**What it is now.** This branch's entry is **ADR-0029** — "Service routers discover shards by network-object type". The body cites that, and I verified it against the live PR rather than against my own edit.

**Why the number is not the one you would have predicted.** Between your review and this reply, `main` gained nine commits and claimed 0026, 0027 and 0028 for its own decisions. That made #2649 `CONFLICTING` outright — not a stale-citation problem but an unmergeable one — so the branch is rebased onto current `main` (`1cf0623e4fe`) and the entry moved 0026 → 0029. All 69 commits replayed; one was dropped deliberately, an intermediate "renumber to 0019" commit that became empty once I renumbered straight to the final value. I confirmed it was the only one by diffing sorted commit subjects before and after.

**Verified after the rebase:** `tsc --noEmit` clean; core suite 161 files / 2040 tests; CI green on Windows, macOS and Ubuntu, and I checked the run's `headSha` matches the branch tip so it is not a stale green.

**The part worth your attention is the pattern, not this instance.** The ADR numbers on this stack have now gone stale three times, and the PR bodies twice in a single day — the second time within an hour of my fixing them. The number a branch writes is not the number it merges with, and no tool catches the drift: a stale citation compiles, lints and reads fine. We are proposing a small standards change so this stops recurring — treating the ADR number as a property of the merge rather than of the branch, with a re-check immediately before merge instead of at write time. It will come to you separately as its own change, off `main`, rather than riding this PR.

One related gap that fell out of the same work and is worth knowing independently: **a `papi.d.ts` diff cannot see a wire-only surface.** A network-object registration missing `'x-experimental': true` regenerates a byte-identical `papi.d.ts`, so the check we all actually run would never catch it. Nothing is missing on this PR — I checked all six shard registrations, and the one that looked unmarked carries its marker in its documentation object — but the blind spot is real and is part of the same proposed change.

rolfheij-sil (commented):
> @rolfheij-sil resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2649)* status: 0 of 153 files reviewed, all discussions resolved.

tjcouch-sil (approved):

PR #2655, opened 2026-08-07
PT-4189: Footnote pane editing components + PT9 line fidelity

lyonsil:
## Summary

Package 1 of [PT-4189](https://paratextstudio.atlassian.net/browse/PT-4189) (WI-4 of
[PT-4186](https://paratextstudio.atlassian.net/browse/PT-4186), "Donna edits text using Standard
view"): the `platform-bible-react` components that let a footnote be edited **in place in the
footnotes pane**, PT9-style, instead of in a popover. Library-only and additive — no web-view
wiring yet. Package 2 (`platform-scripture-editor`) is gated on WI-6 merging `standard-view`
into `main`.

### Why review this

Part of the current epic (PT-4186, Standard-view editing). This is the library half; the
web-view half cannot start until WI-6 lands, so reviewing this now unblocks that sequencing.

## Changes

**`FootnoteEditor` — inline mode** (`inline?: boolean`)

- Fluid width (no popover width-lock), no Save button; edits apply live to the parent editor via
  debounced `replaceEmbedUpdate`, with a flush on unmount.
- New `initialCaretPosition` prop (`'end'` or a mapped offset).
- Popover-mode behavior is pinned unchanged by existing tests.

**`FootnoteList` — row-swap seam**

- `editingFootnoteIndex` + `renderEditingFootnote` swap one row for a consumer-supplied editor.
  The list never imports an editor; composition happens in the consumer.
- `onFootnoteEditRequested` carries a caret position mapped from the click point
  (caret-where-you-clicked, per PT9).
- Editing-row highlight; roving focus skips the editing row; arrow keys are no longer hijacked
  from the embedded editor.

**PT9 line-item fidelity** (last commit, `62c089ab72c`)

Audited each rendered row against PT9's notes-pane renderer (`StandardNotes.xslt`,
`Standard.xslt`, `Base.xslt`) and closed the gaps:

- Closing markers on runs the source closed; `\+` prefix on nested character markers; the `\cat`
  category run; PT9's `notetext` bidi wrapper; the empty-note placeholder; no-break marker
  separators.
- A leading `\fr`/`\xo` target reference now renders **inline** at the head of the note text
  rather than in its own aligned column, so a row reads as one continuous USFM run as PT9's pane
  does. `FootnoteList` drops from three grid columns to two. **This is a deliberate visual
  change** from the earlier commits on this branch.
- Fixed a caret-mapping bug: the row counted USFM marker text toward the caret offset while the
  editor excluded it (Lexical renders markers as `contenteditable=false` decorators), so any
  offset captured with markers shown landed early in the editor by the length of the preceding
  marker text.
- Keys are now the sibling index. No content-derived key can be unique — a note may hold sibling
  runs with the same marker *and* the same text, which was emitting duplicate-key warnings.

**Storybook**

- `Demo/Scripture Editor/Footnotes Pane` composes the editor + pane with production props, and
  defaults to the PT9 Standard-view configuration this feature targets (markers visible, callers
  verbatim), with a control to switch to the formatted-view configuration.
- Supersedes draft PR #2153, whose workaround-based design story this replaces with real props.
  That PR should be closed with a pointer here.

## Known gaps (tracked, TODO-tagged in code)

- [PT-4322](https://paratextstudio.atlassian.net/browse/PT-4322) — marker attributes
  (`\xt |link-href=...`) are still dropped. The correct fix reuses
  `UsjReaderWriter.closingMarkerToUsfm` rather than hand-rolling the attribute rules; the ticket
  explains why. It also covers a related caret drift for notes with bare text children.

## AI Involvement

AI-assisted, with my guidance, throughout. Claude did the PT9 source archaeology (reading the
notes-pane XSLT and comparing it line-by-line against the rendered rows), wrote the components
and their tests test-first, and drafted PT-4322. I directed the scope, made the design calls
(inlining the reference, always showing the category, what to defer), and reviewed the output.
Commits carry `Co-authored-by` trailers. Session URLs are not available for these sessions, so
the per-session links our convention asks for are omitted.

## Testing

- [x] `platform-bible-react` suite green (136 files / 1054 tests)
- [x] `platform-enhanced-resources` green (193 tests) — it also consumes `FootnoteList`, so this
      is the non-breaking check for the other consumer
- [x] Repo-wide `npm run typecheck` and `npm run lint` clean (0 errors)
- [ ] **Manual verification of the pane's appearance not yet done.** The fidelity work is pinned
      by unit tests and Storybook, but nobody has looked at it in the running app — worth a pass
      given the layout change.
- Not run: `dotnet test` (no C# changes).

Note on CI: repo-wide `npm test` intermittently fails a handful of browser-based Storybook
stories with ~15-30s timeouts under parallel contention — a different set each run, none
footnote-related, all passing in isolation. Please don't read those as regressions from this PR.

## Risk Level

**Medium** — additive for existing consumers, but `FootnoteItem`/`FootnoteList` rendering changed
shape (three columns to two, plus extra marker text), so anything depending on the old row layout
will look different. The one behavioral change to a shipped path is the caret-offset origin, which
was measurably wrong before.

[PT-4189]: https://paratextstudio.atlassian.net/browse/PT-4189?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PT-4186]: https://paratextstudio.atlassian.net/browse/PT-4186?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PT-4322]: https://paratextstudio.atlassian.net/browse/PT-4322?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2655)


tjcouch-sil:
Superseded by #2805, which carries every commit of this branch (merged with `main` at 8352e84c047) plus the `platform-scripture-editor` wiring, so the whole PT-4189 feature can be reviewed in one PR. Closing this one to avoid two review passes.

PR #2656, opened 2026-08-08
PT-4201: E2E clipboard USFM round-trip spec

tjcouch-sil:
## Hand-QA follow-ups (2026-09-16/17)

The editor fixes for the first hand-QA pass landed in [scripture-editors #13](https://github.com/paranext/scripture-editors/pull/13) (`ddea1549`+`40d06486` selection snap at decorator glyphs, `a1dede31` html carrier = USFM bytes + Paratext 9 `usfm:`-comment decode + protected-mode paste rules, dist `e91c930c`). This PR's spec gained four steps that drive those behaviours through the real browser and OS clipboard — the parts no jsdom test can see:

- a real mouse drag that **starts on a figure's `\fig ` glyph** copies the whole figure, beginning at the glyph's first byte (Chromium clamps a drag begun inside a `contenteditable="false"` island to that island, so the figure is the most such a drag can select — documented in the step);
- a drag that **ends inside the `|src="…"` attribute run** copies the whole run and nothing past it;
- copying a footnote puts the note's USFM bytes in **`text/html` as well as `text/plain`**, with no `data-caller` residue (what Paratext 9's html-first paste reads);
- pasting a **Paratext 9 clipboard** (`text/plain` = the caller glyph, `text/html` = P9's `exclude` caller span with the escaped `<!--usfm:…-->` comment) materializes a real note and never inserts the glyph as text.

`test.setTimeout(480_000)` on top of `test.slow()`: eight steps' paste/poll allowances no longer fit the 360 s budget after a ~85 s cold launch.

Run against scripture-editors `e91c930c` linked into this checkout: **1 passed (2.4 min), first attempt** — all eight steps, including the four original ones.

Merge order is unchanged: #13 must land (and the staged editor revision follow) before this spec can pass on `main`.

# Code Review Summary

**Branch**: `standard-view-pt-4201` (both repos)

**Base**: `main` — paranext-core `origin/main`, scripture-editors `eten-tech-foundation/main`

**Date**: 2026-09-15

**Files changed**: 1 (paranext-core, committed) + 40 (scripture-editors)

**PRs**: [paranext-core #2656](https://github.com/paranext/paranext-core/pull/2656) · [scripture-editors #532](https://github.com/eten-tech-foundation/scripture-editors/pull/532)

## Overview

PT-4201 makes Standard-view copy/paste round-trip USFM faithfully. The editor half (scripture-editors #532) makes `text/plain` byte-faithful USFM, stops an empty copy from overwriting the clipboard with `@lexical/clipboard`'s `#` placeholder, keeps opaque constructs (figure/sidebar/periph/ref/optbreak) intact across the internal clipboard flavor, replaces a blanket NBSP→`~` mapping with a positional rule, strips pasted `\c`/`\id` before they poison the save loop, and routes attribute-run pastes as plain text. The paranext-core half (#2656) adds one Playwright spec driving that round trip through the real OS clipboard.

This was an automated review: five `review-analyzer` passes (api/style/compliance across both repos), each finding verified against the code before being acted on. Every finding below is fixed, refuted, dismissed with a reason, or explicitly left open. Two of the analyzers' Important findings turned out to be genuine defects the branch's own 4,200-test suite could not see, and both now have pins that go red when the fix is reverted.

## API Changes

**paranext-core**: None. The only committed file is a new e2e spec.

**scripture-editors** — the published surface is `packages/platform` only; `libs/shared`, `libs/shared-react`, `libs/test-data` are `"private": true`.

- `packages/platform/src/index.ts` and `packages/platform/etc/platform-editor.api.md`: **no change**.
- `EditorRef.copy()` / `EditorRef.cut()`: signatures unchanged, **behavior changed** — with nothing selected they now leave the clipboard untouched instead of writing `#`.
- Private workspace libs: `shared-react` gained `copySelection`, `cutSelection`, `registerEmptyCopyGuard`, `$hasCopyableSelection`; `shared` exported `tableCellMarkerWithSpan`.
- Package-internal: `normalizePastedNbsp`, `stripPastedChapterAndBookId`, `$selectionToUsfmText` added; `PastePayload` narrowed to `{text, isInternal}`; `MarkerEditContext` gained `pasteRebuildArmed` / `pastePendedKeys` / `structureProtectionMode`; `isParaKindMarker` / `isCharKindMarker` moved to a new leaf module `markerKind.utils.ts`.

None.

**scripture-editors**

- [x] **`stripPastedChapterAndBookId` destroyed pasted user content past the chapter number.** `/\\(?:c|id)(?![\w-])[^\n\\]*/g` ran to the next `\` or newline, so pasting `x \c 5 y` stripped to `x ` — the `y` silently dropped. The tokenizer takes only the next whitespace-delimited word as the chapter number (`getNextWord`, `usfmFragmentToUsj.ts:372-376`), so everything after it is ordinary content. _(fixed during review: split into `CHAPTER_TOKEN` (marker + separator + one word, stopping at a backslash) and `BOOK_ID_TOKEN` (unchanged — an `\id` line's whole remainder genuinely is payload); updated the pin in `markerPasteFidelity.test.tsx` that had encoded the loss, and added direct pins. Reverting the regex turns the new pin red with `expected 'x ' to be 'x  y'`.)_
- [x] **`Editor.tsx`'s `structureProtectionMode` pass-through had no test — deleting the line left all 4,234 tests green.** Both `Editor` and `MarkerEditPlugin` default the prop to `"off"`, so a dropped pass-through silently starves `StructureKeyboardPlugin`'s sanitizer on a protected document's paste. _(fixed during review: threaded the option through `mountStandardViewEditor` and added a real-`Editor` paste pin. Verified falsifiable — deleting `Editor.tsx:1160` turns it red while the plugin-level pin stays green. Note the option lives on `EditorOptions`, not as a top-level `<Editor>` prop.)_
- [x] **The empty-copy rule was re-implemented inline instead of reusing the predicate this branch created.** `$handleCopyForStandardView` spelled `(!selection || selection.isCollapsed())` — provably the exact negation of shared-react's `$hasCopyableSelection()` — leaving two layers whose agreement is load-bearing resting on two independently-spelled predicates in two packages. _(fixed during review: the platform layer now calls `!$hasCopyableSelection()`, which also gives that export its first consumer outside its own module.)_
- [x] **Four new marker regexes bypassed `markerName.pattern.ts` ("the single authority"), and two of the spellings already disagreed.** `AFTER_MARKER_NBSP`/`BEFORE_MARKER_NBSP` used `[a-z0-9-]` + `i`; `LEADING_MARKER_LITERAL` and the `\c`/`\id` token used `[\w-]`. `\w` includes `_`, so `\my_marker` was a marker to one and not the other — its separator NBSP survived into content as a literal `~` exactly where the tokenizer expects a space. _(fixed during review: all four derive from `ENGINE_MARKER_NAME_BYTES`. New pin goes red on revert with `expected '\my_marker~light' to be '\my_marker light'`.)_
- [x] **Spec-document breadcrumbs and provenance narration in comments** — `Deferred item 2`, `S2's own documented case`, `measured`, `confirmed by reading LexicalClipboard.dev.js`, `Regression classes closed by`, `inherited rather than introduced`, `rather than papered over`. The repo's own `CLAUDE.md:275` forbids spec section numbers in code. _(fixed during review: rules kept, provenance and IDs stripped, across `whitespaceDisplay.plugin.utils.ts`, `UnknownNode.ts`, `attributeContextPasteFidelity.test.tsx`.)_
- [x] **`$`-sigil on two pure `string → string` functions.** `CLAUDE.md:269` requires every exported `$`-prefixed function to document its calling context; neither had one, because neither has a context. _(fixed during review: renamed to `normalizePastedNbsp` / `stripPastedChapterAndBookId` across sources, tests, and the semantics doc.)_

**paranext-core**

- [ ] **The spec cannot pass against the editor `main` builds, so merge order matters.** `$stripPastedChapterAndBookId` and `$selectionToUsfmText` do not exist on `origin/platform-yalc`. CI is unaffected (`test.yml` runs only `test:e2e:smoke`), but after this merges to `main`, `npm run test:e2e:isolated scripture-editor` fails for every developer until #532 is on `platform-yalc`. **Land #532 and move `platform-yalc` first.**
- [x] **Hard-coded `Control+` chords made the spec unrunnable on macOS.** `Control+C/V/Z` and `Control+Home/End` are OS editing chords, not app accelerators — the only other bare `Control+` uses in the suite (`verse-navigation-shortcuts.spec.ts`) genuinely are accelerators. _(fixed during review: per-platform key constants following `navigation-history.spec.ts:43-45`.)_
- [x] **The copy step could pass on clipboard contents it did not write.** Nothing cleared the OS clipboard before `Control+C`, and a retry attempt that clears the poll then fails the one-shot `\p ` check leaves real USFM there for the next attempt. _(fixed during review: a sentinel is written immediately before the copy.)_
- [x] **The undo step was three negative assertions with no positive control.** An empty or mid-replacement editor satisfies all of them, and the PDP's debounced save echoes by replacing content wholesale. _(fixed during review: added `toContainText('Amittai')` as the positive control, plus the missing closer-count assertion.)_
- [x] **No negative control (the Revert Test).** _(addressed during review by static analysis rather than a rebuild: `$selectionToUsfmText` and `$stripPastedChapterAndBookId` are absent from `origin/platform-yalc` while `$handleCopyForStandardView` is present, and the pre-fix paste gate returns `false` for any payload without an NBSP. So **step 4 (`\c` strip) is the true fix-pin; steps 1–3 are characterization pins** that likely pass pre-fix too. Recorded here so nobody mistakes the spec for a four-way regression net.)_

### Minor — Consider

**paranext-core**

- [x] `.first().toBeAttached()` could not catch glyph doubling on paste _(fixed: `toHaveCount(1)` on both opener and closer — Jonah has zero `\nd` going in, so the exact count is free)_
- [x] The undo step checked only the opening span, so a half-reverted pair would pass _(fixed: both halves asserted back to 0)_
- [x] The OS clipboard leak was undocumented _(fixed: new bullet in `e2e-tests/CLAUDE.md`'s "State that leaks between runs" — it is machine-wide and **not** contained by `isolatedProjectRoot` or the temp `userDataDir`, unlike everything else the spec touches)_
- [x] `tests/isolated/README.md` documented every subset except `scripture-editor/` _(fixed; pre-existing, one line)_
- [x] Backward-facing comment: `"Yahweh" — the word used in earlier drafts of this spec` _(fixed: keeps the durable fact that scripture vocabulary cannot serve as a sentinel, drops the drafting history)_
- [ ] **The `Control+Z` is not synchronized with the PDP's ~700ms debounced save/echo.** Mitigated by the positive control added above, but the ordering hazard is real and unresolved. See "In-Review Quality Check" for what two live runs showed.
- [ ] ~~The `press('Home')` between paste and undo should be dropped~~ _(dismissed: settles are history-merged via `$settleWithoutOwnUndoEntry`, which the analyzer itself verified, so it cannot split the undo step; the comment already marks it belt-and-braces.)_
- [ ] ~~`dev-packages.json` redirects every developer install and CI leg to a feature branch~~ _(refuted: working-tree-only — `git diff <MERGE_BASE> HEAD -- dev-packages.json` is empty, nothing staged. Both core analyzers looked at this and the second reached the same conclusion. Deliberately excluded from the review commit.)_

**scripture-editors**

- [x] `$hasCopyableSelection` was an export with no consumer outside its own module _(resolved by the reuse fix above)_
- [x] `isParaKindMarker` / `isCharKindMarker` were byte-identical but for one `MarkerType` comparison, in a module that exists precisely so the two tiers cannot disagree _(fixed: one private `isKindMarker` body, two thin wrappers; the both-kinds-true unknown fallback is preserved and now documented)_
- [x] `$selectionToUsfmText` mirrors `RangeSelection.getTextContent()` without naming the version it was mirrored from _(fixed: pinned to lexical 0.43.0 with an instruction to re-check on upgrade)_
- [x] **The clipboard semantics doc lived in a gitignored directory.** `.gitignore:67` ignores `docs/superpowers`; the file was tracked only by a force-add, so future edits to the durable contract would never appear in `git status`. _(fixed: `git mv` to `docs/clipboard-semantics.md`, alongside `docs/standard-view-invariants.md`, matching the repo's flat topic-named docs; the one code reference now carries the real path.)_
- [x] Deferred item 6 pointed at `docs/superpowers/specs/2026-07-07-standard-view-followups.md`, which does not exist _(fixed: replaced with a plain "tracked separately")_
- [x] The test-mapping table still described the E2E spec as "written, not yet run in this task" _(fixed: it runs and passes)_
- [ ] ~~`UnknownNode.ts`'s new block is 88% comment~~ _(dismissed: the block encodes a genuinely subtle invariant — why child-count is the test, and why excluding a wrapper makes `$appendNodesToJSON` hoist its children into a convincing lie. This repo is deliberately comment-heavy and the reasoning is not inferable from the code.)_
- [ ] `$noteCallerText`'s `caller || GENERATOR_NOTE_CALLER` fallback is untested — no fixture has a note without an explicit caller
- [ ] `$isNoteInternalDisplaySeparator` and `$collapsedNoteCategoryBytes` run the same three-step probe, differing only in the last check's sign — deferred as a refactor beyond this PR's scope
- [ ] `$handlePasteForStandardView` has four positional params, three optional only so tests can call it with one — an options object would read better; deferred
- [ ] `whitespaceDisplay.plugin.utils.ts` is now 787 lines with 10 of 11 exports being clipboard concerns — splitting out `standardViewClipboard.utils.ts` would match the module boundary to the concern; deferred (large diff churn on a branch already under review)
- [ ] **`EditorRef.copy()`/`cut()` behavior change deserves a release note** — a host that calls `copy()` at a collapsed caret used to get a clipboard write (`#`) and now gets nothing. Strictly better, but there is no compile-time signal for a consumer that built a workaround.

#### Shared Regions Modified

None. No changed file in either repo contains a `#region shared with` marker.

#### Extension Config Changes

- [n/a] paranext-core: no files under `extensions/` changed.
- [n/a] scripture-editors: no `extensions/` tree, and no build/tooling config changed.

## Positive Observations

- **Load-bearing claims in the doc comments hold up against the installed Lexical source.** An analyzer independently re-derived them: `$appendNodesToHTML` returns early on a null `element` before consuming the `excludeFromCopy` value; `$appendNodesToJSON` passes `'html'` and hoists excluded nodes' children; `copyToClipboard(editor, null)` really does append `createTextNode('#')` while `$copyToClipboardEvent` bails before `preventDefault()` on a collapsed selection. That is the exact `#`-leak mechanism the guard closes.
- **Falsifiability across the new tests is unusually good.** Copy-side pins assert exact full USFM strings the old blanket `getTextContent().replaceAll(NBSP, " ")` cannot produce, and three places carry explicit negative controls that isolate the fix as the only variable (`tier2Rebuild.utils.test.tsx:405`, `markerPasteFidelity.test.tsx:521`, `attributeContextPasteFidelity.test.tsx:251`).
- **The 2SA fixtures are exact regeneration output.** Re-running `GENERATE_TEST_DATA=1` plus prettier reproduced the committed files byte-identically, and `2sa.lexical.hidden.ts` correctly shows no change.
- Putting the empty-copy guard on `COPY_COMMAND`/`CUT_COMMAND` at `COMMAND_PRIORITY_LOW` is the right seam — one registration covers the shortcuts, the context menu and `EditorRef`, and it verifiably sits above rich-text's `EDITOR`-priority fallback and below Standard view's own `HIGH` claim.
- `markerKind.utils.ts` is a genuinely good extraction: it removes the Tier 1 ↔ Tier 2 import cycle and documents why a narrower `type === MarkerType.Paragraph` check would have been wrong.
- The e2e spec's step 4 is exemplary — a mid-line `${token} \c 99 ` payload matching the deliberately-unanchored strip regex, with three independent probes, one of them (the toolbar reference) read entirely outside the editor DOM.
- Every locator in the spec resolves against real editor code, and `nd` is a genuinely clean sentinel (the bundled Jonah SFM contains zero `\nd`), so the attach and zero-count assertions are non-vacuous.
- No `eslint-disable`, `@ts-expect-error`, type assertions, `console.*`, `.only`, or untracked TODOs added anywhere in either diff.

## Interview Notes

Run in automated mode at the author's request — no interview was conducted. The author asked for each finding to be considered on its merits, verified, and then fixed if small-to-medium, low-risk and worthwhile; otherwise refuted or dismissed with a reason. Judgment calls made autonomously:

- **Fixed** anything that was a verified correctness defect, a stated-convention violation, or a missing/weak test — including two defects the existing 4,200-test suite could not see (the `\c` over-strip and the untested `Editor` pass-through), both now pinned with tests confirmed to go red on revert.
- **Deferred** pure refactors with large diff churn and no correctness content (the module split, the options-object signature, the note-separator merge) as out of scope for a branch already in review.
- **Did not re-home PR #532.** Active scripture-editors development moved to `paranext/scripture-editors` under PT-4500 — that repo carries the `pt-4313`, `pt-4370` and `pt-4615` branches and commits `packages/*/dist` — while #532 still targets `eten-tech-foundation`. Moving a PR across repos is the author's call and the transfer guide is theirs to exercise.
- **Did not widen shared e2e timeouts** in response to the flake found below; the gate is shared infrastructure used by every isolated spec.

## In-Review Quality Check

**scripture-editors** — all gates green after every change:

- `nx run-many -t test` (`--skip-nx-cache --pool=forks --maxWorkers=4`): **shared 697 passed · shared-react 1674 passed / 1 skipped · platform-editor 1864 passed / 4 skipped**. Baseline before this review was 697 / 1674+1 / 1857+4, so the review added **7 tests** and broke none.
- `nx run-many -t typecheck lint`: clean — **0 errors**. The 28 lint warnings are pre-existing `no-console` / `no-loop-func` in files this branch does not touch (cross-referenced against the changed-file list).
- `nx format:check`: clean.
- Each new pin was **verified falsifiable** by reverting its fix and watching it go red with the expected message.

**paranext-core**:

- `tsc -p e2e-tests/tsconfig.json`, `eslint`, `prettier --check`: all clean.
- **The e2e spec was run twice against a live app.** Run 1: **passed, 4.0m**. Run 2: **failed then passed on retry — Playwright reports it flaky.**
- **The flake is not in the clipboard logic.** It failed at `clipboard-usfm-round-trip.spec.ts:77` (`makeSampleProjectEditable()`), in the *shared* backend-readiness gate at `e2e-tests/fixtures/helpers.ts:942`: `PAPI method "object:platform.Paratext-pdpf.getProjectDataProviderId" not listed in rpc.discover within 120000ms`. That is the C# data provider failing to register its PDP factory in time on a machine under heavy load, before any clipboard step runs. It affects every isolated spec that calls that helper, and is not caused by this branch.
- Worth noting for the budget: `test.slow()` gives 360s, and a passing run took 240s while the setup gate alone can consume 120s. The margin is thinner than the previously-recorded 1.7m suggests.

## Suggested Review Focus

- [ ] **Merge order is a hard constraint**: scripture-editors #532 must land *and* reach `platform-yalc` before paranext-core #2656, or `npm run test:e2e:isolated scripture-editor` goes red on `main` for every developer. CI will not catch this (only `test:e2e:smoke` runs there).
- [x] **Which repo should #532 land in?** PT-4500 moved active development to `paranext/scripture-editors`; this PR still targets `eten-tech-foundation`. Transferring it also means rebuilding the committed `packages/*/dist` that the paranext repo carries and `eten-tech-foundation` does not.
- [ ] **The narrowed `\c` strip is a deliberate behavior change.** `x \c 5 y` now keeps the `y`. Confirm that keeping trailing content is what you want in every paste shape, not just the tested ones.
- [ ] **The e2e flake.** It is a pre-existing shared-fixture readiness gate, but this spec's 360s budget leaves less headroom than the others. Decide whether to raise the gate, or accept the retry.
- [ ] **`EditorRef.copy()`/`cut()`** is a host-visible behavior change with no compile-time signal — does it need a release note?
- [ ] Remaining documented deferrals in `docs/clipboard-semantics.md` (items 10–14: the stranded-decorator class across figure/table/sidebar/periph/ref, table/milestone corpus losses, the dropped html-carrier NBSP rule) are unchanged by this review and still open.

---

## Summary

The isolated-Electron e2e spec for PT-4201 (copy/paste USFM fidelity), exercising the **real OS clipboard** via the Electron clipboard bridge: copy-out asserting valid USFM bytes, an external `\nd` paste settling into real marker spans, one-step undo, and the pasted-`\c` strip guarantee (chapter count stays 1).

**This spec runs and passes** (1.7m), against the fixed editor from the companion PR [scripture-editors #532](https://github.com/eten-tech-foundation/scripture-editors/pull/532). It was originally committed unrun.

> **Merge order.** The `\c`-strip step asserts behavior that exists only in #532. Merge that first (and let `platform-yalc` pick it up) — otherwise this spec fails for anyone who runs it from a clean checkout. The working-tree `dev-packages.json` revision flip stays uncommitted on purpose.

### Why review this

Part of the current epic (Donna edits text using Standard view, PT-4186): this is the end-to-end gate on the clipboard fidelity work.

## Changes

- New `e2e-tests/tests/isolated/scripture-editor/clipboard-usfm-round-trip.spec.ts` (one `test()`, `test.step` sub-scenarios, follows the `type-through-save-echo` idiom).
- Selector fix: the strip-guarantee probe targeted `p.chapter-marker`, a class no editor code emits; `ChapterNode` renders `<p class="chapter usfm_c" data-marker="c">`. The assertion was silently meaningless until the spec actually ran.

## Review follow-ups (this round)

- **The `\c`-strip step had no positive signal that the paste landed.** All three of its assertions were already true of the pre-paste document — the chapter count was pinned at 1 two lines above, `not.toContainText('\c 99')` holds for untouched Jonah, and the toolbar already read "Jonah 1" — so Playwright's first immediate poll satisfied every one of them and the step went green whether or not the paste was ever processed. A regression in the engine's strip would not have been caught. The payload now leads with a token the step waits for first, the way the `\nd` step above already did. It also puts the `\c` mid-line rather than on a line of its own, which is the shape a start-of-line-anchored strip would miss.
- Dropped the stale "this spec has not been run" header and corrected a comment that named a chapter class the code does not use (it reproduced exactly the wrong class the previous commit had removed from the locator).

## AI Involvement

Spec generated and iterated by Claude Code (Opus 5) under subagent-driven development with adversarial review. Review rounds caught vacuous assertions against native WEB text and two wrong DOM probes.

## Testing

- [x] Spec passes: `npm run test:e2e:isolated tests/isolated/scripture-editor/clipboard-usfm-round-trip.spec.ts` (1.7m), against a freshly built + yalc-linked `platform-editor` and a rebuilt renderer DLL
- [x] `tsc --noEmit -p e2e-tests/tsconfig.json` clean, eslint + prettier clean
- [ ] Two **pre-existing** specs in this subset fail on this machine (`formatted-default-simple-mode`, `paragraph-style-trigger-column-floor`) — both time out waiting for `button[aria-label="book-chapter-trigger"]` to become clickable. This branch touches no product code, and the clipboard and power-mode specs click that same button fine; not proven unrelated on a clean checkout, so flagging rather than claiming clean.

## Code-review round (2026-09-17)

A `/code-review` pass over this PR and its scripture-editors companion (#13). Two fixes here; the substantive ones are editor-side.

- **The attribute-run step measured its drag boxes before a `click()` that scrolls.** `openerBox` and `attributeBox` were read, then `figureTailGlyph.click()` ran to collapse the previous step's selection. Playwright's `click()` performs actionability checks including scroll-into-view, so any scroll there invalidated both boxes and the drag aimed at whatever text had moved into those coordinates — surfacing as a misleading 15s `toContain` timeout rather than a wrong-selection failure. The click now happens before the measurements.
- **`test.slow()` above `test.setTimeout(480_000)` was dead.** `setTimeout` replaces the budget outright rather than scaling it (`TimeoutManager.slow()` multiplies at call time; the later `setTimeout` overwrites), so the 3× marking did nothing while the comment read as though both applied. Removed, with the reasoning folded into one comment.

`tsc --noEmit -p e2e-tests/tsconfig.json`, eslint and prettier clean.

**No spec changes were needed for the editor-side behaviour change** in #13 (paste no longer deletes the host paragraph's marker): this spec's Paratext 9 step asserts the note token appears, and its copy steps assert clipboard bytes — neither depends on the paragraph count around the insertion point. The merge-order constraint below is unchanged.

## Review round (lyonsil, 2026-09-18) — `bea33ed9ce1`

- The undo step checks its positive control before the negatives, so a momentarily empty editor cannot satisfy them all (#1).
- The attribute-run drag releases on one of the run's own client rects, not the bounding-box centre that falls between the lines of a wrapped run (#2).
- `requireElementBox` rejects an off-screen target, so a mis-aimed drag fails as that rather than as a 15s clipboard timeout (#7).
- The `\fig ` opener and the caption suffix are value comparisons that show the bytes on failure (#9, #10).
- The raw `text/plain` read says why it is not NBSP-normalized (#11).
- ADR: the Paratext 9 `NormalizeTokenUsfm` citation states its condition instead of "every", and the `\b` observation names the rule that closed it (#4, #5).

The spec passed against an editor build containing scripture-editors #13's review fixes (on the automatic retry, after the first attempt hit the known cold-start `useWebViewState` crash at its first paste step); `tsc --noEmit -p e2e-tests/tsconfig.json`, eslint and prettier clean.

## Re-run against scripture-editors #13's second review round (2026-09-22)

That round changed two things this spec exercises — `StructureKeyboardPlugin`'s CUT refusal moved to `COMMAND_PRIORITY_CRITICAL`, and `EditorRef.formatPara`/`insertNote` now commit discretely — so the spec was re-run against a build carrying it (`9dd93da5`). Green in 3.3 minutes on the first attempt, no retry needed, every step including the reworked drag and undo steps. The spec itself is unchanged.

## Risk Level

Low — test-only addition; no production code.

AI-assisted — Claude Code session.

https://claude.ai/code/session_01QCjocx9uZcPqB2esPh9r6f

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2656)

### DIFF
--- a/e2e-tests/tests/isolated/scripture-editor/clipboard-usfm-round-trip.spec.ts
+++ b/e2e-tests/tests/isolated/scripture-editor/clipboard-usfm-round-trip.spec.ts
@@ -44,6 +44,13 @@ test.use({
   electronLaunchOptions: { isolatedProjectRoot: true, envOverrides: { DEV_NOISY: 'false' } },
 });
 
+// A fabricated all-caps token that cannot occur in the bundled WEB text, so `toContainText` /
+// `not.toContainText` are unambiguous (same rationale as type-through-save-echo.spec.ts's
+// TYPETHROUGHALPHA/BETA). "Yahweh" — the word used in earlier drafts of this spec — actually
+// occurs natively throughout Jonah 1 (vv. 1, 3, 4, 9, 10, 14 x3, 16 x2, 17 in the bundled WEB SFM),
+// which would make the paste-landed assertion pass vacuously and the undo assertion unsatisfiable.
+const PASTE_TOKEN = 'CLIPROUNDTRIPALPHA';
+
 test.describe('scripture editor clipboard USFM round trip', () => {
   test('Standard view copy/paste round-trips byte-faithful USFM through the real OS clipboard', async ({
     mainPage,
@@ -93,15 +100,20 @@ test.describe('scripture editor clipboard USFM round trip', () => {
     });
 
     await test.step('pasting external USFM-looking text tokenizes into editor marker spans', async () => {
-      await electronApp.evaluate(({ clipboard }) => clipboard.writeText('\\nd Yahweh\\nd* '));
+      await electronApp.evaluate(
+        ({ clipboard }, token) => clipboard.writeText(`\\nd ${token}\\nd* `),
+        PASTE_TOKEN,
+      );
       await editorInput.click();
       await editorInput.press('Control+End');
       await editorInput.press('Control+V');
-      await expect(editorInput).toContainText('Yahweh', { timeout: 20_000 });
+      await expect(editorInput).toContainText(PASTE_TOKEN, { timeout: 20_000 });
 
-      // Depart the just-pasted text to trigger the pend->settle rebuild (same idiom as
-      // attribute-display-settle.spec.ts): moving the caret off it is what turns raw marker syntax
-      // into a rendered MarkerNode span rather than leaving it as literal pending text.
+      // A fully-terminated paste (`\nd …\nd* `, both opener and closer present) rebuilds
+      // synchronously at paste time — unlike a typed span mid-edit, there is no pending tier it
+      // needs to depart to settle (contrast attribute-display-settle.spec.ts, where departure is
+      // required because typing stops mid-span). This Home press is belt-and-braces only, not a
+      // required step, in case caret position affects rendering some other way.
       await editorInput.press('Home');
 
       await expect(editorFrame.locator('span.opening[data-marker="nd"]').first()).toBeAttached({
@@ -114,26 +126,31 @@ test.describe('scripture editor clipboard USFM round trip', () => {
 
     await test.step('undoing the paste is a single step', async () => {
       await editorInput.press('Control+Z');
-      await expect(editorInput).not.toContainText('Yahweh', { timeout: 20_000 });
+      await expect(editorInput).not.toContainText(PASTE_TOKEN, { timeout: 20_000 });
       await expect(editorFrame.locator('span.opening[data-marker="nd"]')).toHaveCount(0, {
         timeout: 20_000,
       });
     });
 
     await test.step('pasting a `\\c` chapter marker cannot corrupt the open chapter', async () => {
+      // Chapter markers render as a block-level `<p class="chapter-marker usfm_c"
+      // data-marker="c">` (ChapterNode.createDOM), not an inline `span.opening` — the loaded
+      // chapter has exactly one going in, and a corrupted paste could either add a second one or
+      // replace/remove the existing one, so pin the baseline before pasting.
+      const chapterMarker = editorFrame.locator('p.chapter-marker[data-marker="c"]');
+      await expect(chapterMarker).toHaveCount(1);
+
       await electronApp.evaluate(({ clipboard }) => clipboard.writeText('\\c 99 '));
       await editorInput.click();
       await editorInput.press('Control+End');
       await editorInput.press('Control+V');
 
-      // Give the paste a moment to settle, then assert the engine's strip guarantee three
-      // independent ways: no `\c` marker span was created, no literal `\c` bytes landed as text
-      // either, and the toolbar's active reference — read entirely outside the editor DOM — is
-      // still Jonah 1. Any of the three would fail if a stray "\c 99" were (mis)tokenized as a
-      // real chapter marker.
-      await expect(editorFrame.locator('span.opening[data-marker="c"]')).toHaveCount(0, {
-        timeout: 20_000,
-      });
+      // Assert the engine's strip guarantee three independent ways: the chapter-marker block
+      // count is unchanged (no second one added, the existing one not corrupted away), no literal
+      // `\c 99` bytes landed as text either, and the toolbar's active reference — read entirely
+      // outside the editor DOM — is still Jonah 1. Any of the three would fail if a stray
+      // "\c 99" were (mis)tokenized as a real chapter marker.
+      await expect(chapterMarker).toHaveCount(1, { timeout: 20_000 });
       await expect(editorInput).not.toContainText('\\c 99');
       await expect(bcvTrigger).toContainText('Jonah 1', { timeout: 10_000 });
     });
--- a/e2e-tests/tests/isolated/scripture-editor/clipboard-usfm-round-trip.spec.ts
+++ b/e2e-tests/tests/isolated/scripture-editor/clipboard-usfm-round-trip.spec.ts
@@ -1,0 +1,141 @@
+/**
+ * E2E for the real OS-clipboard round trip of Standard-view USFM: copying editor content puts
+ * byte-faithful USFM (including `\v` and `\p` marker bytes) on the operating-system clipboard;
+ * pasting external USFM-looking text back in tokenizes it into the editor's marker spans; the paste
+ * and the rebuild it triggers are a single undo step; and pasting a `\c` chapter marker cannot
+ * corrupt the open chapter, because the engine strips `\c`/`\id` bytes on external paste.
+ *
+ * THIS SPEC IS WRITTEN BUT HAS NOT BEEN RUN. The Electron app instance and the ports it needs
+ * (1212, 8876, 9223) were held by a separate session, and the editor build this spec exercises had
+ * not yet been linked into the running app, at the time this file was written. It is verified by
+ * `tsc --noEmit` and `eslint` only — no attempt was made to launch the app or execute the test.
+ * Tracked as PT-4201; run with `npm run test:e2e:isolated
+ * scripture-editor/clipboard-usfm-round-trip` once the app/ports are free and the editor build is
+ * current, then iterate on selectors/timing only — a behavior failure at this layer means the
+ * underlying engine work regressed, not this spec.
+ *
+ * The "external application" side of the round trip is Electron's main-process `clipboard` module,
+ * reached via `electronApp.evaluate(({ clipboard }) => ...)`. That reads and writes the real OS
+ * clipboard from OUTSIDE the renderer entirely — bypassing `navigator.clipboard` and its renderer
+ * permission surface — which is exactly the external round trip this spec needs to exercise.
+ *
+ * ONE test() per spec file on purpose: the isolated fixture is test-scoped, and a SECOND Electron
+ * instance launched against the shared webpack renderer dev server has a documented failure mode
+ * where new dock tabs never render (see isolated.fixture.ts). Sub-scenarios are test.step()s
+ * sharing the one instance, in sequence, so each step's assertions run against the state the prior
+ * step left behind (the undo step, in particular, depends on the paste step having landed).
+ *
+ * Runs against an isolated project root, so the only project is the bundled sample WEB (installed
+ * by the C# backend into the empty root): `npm run test:e2e:isolated scripture-editor`.
+ */
+import { test, expect } from '../../../fixtures/isolated.fixture';
+import {
+  makeSampleProjectEditable,
+  navigateToolbarBcv,
+  openEditableScriptureEditorForProject,
+  SAMPLE_WEB_PROJECT_ID,
+  waitForHomeTab,
+} from '../../../fixtures/scripture-editor-helpers';
+
+// Power mode so the editor opens in Standard view (marker-editable) — the surface this feature
+// targets. DEV_NOISY=false keeps the normal Home layout (see standard-default-power-mode.spec.ts).
+test.use({
+  interfaceMode: 'power',
+  electronLaunchOptions: { isolatedProjectRoot: true, envOverrides: { DEV_NOISY: 'false' } },
+});
+
+test.describe('scripture editor clipboard USFM round trip', () => {
+  test('Standard view copy/paste round-trips byte-faithful USFM through the real OS clipboard', async ({
+    mainPage,
+    electronApp,
+  }) => {
+    // Heavy isolated test (own Electron instance, backend-readiness gates, several clipboard round
+    // trips each awaiting a settle/save). 3x "slow" budget — see standard-default-power-mode.spec.ts.
+    test.slow();
+
+    await waitForHomeTab(mainPage);
+    await makeSampleProjectEditable();
+    const editorId = await openEditableScriptureEditorForProject(mainPage, SAMPLE_WEB_PROJECT_ID);
+    const editorFrame = mainPage.frameLocator(`iframe[data-web-view-id="${editorId}"]`);
+    await editorFrame.locator('.editor-container').waitFor({ timeout: 60_000 });
+
+    // Jonah 1 (same book used by type-through-save-echo.spec.ts) — a small single chapter that
+    // opens with a `\p` paragraph marker immediately followed by `\v 1`.
+    await navigateToolbarBcv(mainPage, 'Jonah 1:1');
+    const editorInput = editorFrame.locator('.editor-input.marker-editable');
+    await expect(editorInput).toBeAttached({ timeout: 60_000 });
+    await expect(editorInput).toContainText('Amittai', { timeout: 60_000 });
+
+    // The active reference, read from the main toolbar — used below as an independent signal (not
+    // derived from the editor DOM at all) that a paste never changed the open chapter.
+    const bcvTrigger = mainPage.locator('button[aria-label="book-chapter-trigger"]').first();
+
+    await test.step('copying a selection puts byte-faithful USFM on the OS clipboard', async () => {
+      await editorInput.click();
+      await editorInput.press('Control+Home');
+      await editorInput.press('Shift+Control+End');
+      await editorInput.press('Control+C');
+
+      // The renderer's copy handler may forward to the OS clipboard asynchronously, so poll rather
+      // than reading once — a single premature read could observe a stale/empty clipboard.
+      let clipboardText = '';
+      await expect
+        .poll(
+          async () => {
+            clipboardText = await electronApp.evaluate(({ clipboard }) => clipboard.readText());
+            return clipboardText;
+          },
+          { timeout: 15_000 },
+        )
+        .toMatch(/\\v 1 /);
+      // Paragraph-marker bytes survived to the OS clipboard too, not just the verse marker.
+      expect(clipboardText).toMatch(/\\p /);
+    });
+
+    await test.step('pasting external USFM-looking text tokenizes into editor marker spans', async () => {
+      await electronApp.evaluate(({ clipboard }) => clipboard.writeText('\\nd Yahweh\\nd* '));
+      await editorInput.click();
+      await editorInput.press('Control+End');
+      await editorInput.press('Control+V');
+      await expect(editorInput).toContainText('Yahweh', { timeout: 20_000 });
+
+      // Depart the just-pasted text to trigger the pend->settle rebuild (same idiom as
+      // attribute-display-settle.spec.ts): moving the caret off it is what turns raw marker syntax
+      // into a rendered MarkerNode span rather than leaving it as literal pending text.
+      await editorInput.press('Home');
+
+      await expect(editorFrame.locator('span.opening[data-marker="nd"]').first()).toBeAttached({
+        timeout: 20_000,
+      });
+      await expect(editorFrame.locator('span.closing[data-marker="nd"]').first()).toBeAttached({
+        timeout: 20_000,
+      });
+    });
+
+    await test.step('undoing the paste is a single step', async () => {
+      await editorInput.press('Control+Z');
+      await expect(editorInput).not.toContainText('Yahweh', { timeout: 20_000 });
+      await expect(editorFrame.locator('span.opening[data-marker="nd"]')).toHaveCount(0, {
+        timeout: 20_000,
+      });
+    });
+
+    await test.step('pasting a `\\c` chapter marker cannot corrupt the open chapter', async () => {
+      await electronApp.evaluate(({ clipboard }) => clipboard.writeText('\\c 99 '));
+      await editorInput.click();
+      await editorInput.press('Control+End');
+      await editorInput.press('Control+V');
+
+      // Give the paste a moment to settle, then assert the engine's strip guarantee three
+      // independent ways: no `\c` marker span was created, no literal `\c` bytes landed as text
+      // either, and the toolbar's active reference — read entirely outside the editor DOM — is
+      // still Jonah 1. Any of the three would fail if a stray "\c 99" were (mis)tokenized as a
+      // real chapter marker.
+      await expect(editorFrame.locator('span.opening[data-marker="c"]')).toHaveCount(0, {
+        timeout: 20_000,
+      });
+      await expect(editorInput).not.toContainText('\\c 99');
+      await expect(bcvTrigger).toContainText('Jonah 1', { timeout: 10_000 });
+    });
+  });
+});


tjcouch-sil:
# Code Review Summary

**Branch**: `standard-view-pt-4201` (both repos)

**Base**: `main` — paranext-core `origin/main`, scripture-editors `eten-tech-foundation/main`

**Date**: 2026-09-15

**Files changed**: 1 (paranext-core, committed) + 40 (scripture-editors)

**PRs**: [paranext-core #2656](https://github.com/paranext/paranext-core/pull/2656) · [scripture-editors #532](https://github.com/eten-tech-foundation/scripture-editors/pull/532)

## Overview

PT-4201 makes Standard-view copy/paste round-trip USFM faithfully. The editor half (scripture-editors #532) makes `text/plain` byte-faithful USFM, stops an empty copy from overwriting the clipboard with `@lexical/clipboard`'s `#` placeholder, keeps opaque constructs (figure/sidebar/periph/ref/optbreak) intact across the internal clipboard flavor, replaces a blanket NBSP→`~` mapping with a positional rule, strips pasted `\c`/`\id` before they poison the save loop, and routes attribute-run pastes as plain text. The paranext-core half (#2656) adds one Playwright spec driving that round trip through the real OS clipboard.

This was an automated review: five `review-analyzer` passes (api/style/compliance across both repos), each finding verified against the code before being acted on. Every finding below is fixed, refuted, dismissed with a reason, or explicitly left open. Two of the analyzers' Important findings turned out to be genuine defects the branch's own 4,200-test suite could not see, and both now have pins that go red when the fix is reverted.

## API Changes

**paranext-core**: None. The only committed file is a new e2e spec.

**scripture-editors** — the published surface is `packages/platform` only; `libs/shared`, `libs/shared-react`, `libs/test-data` are `"private": true`.

- `packages/platform/src/index.ts` and `packages/platform/etc/platform-editor.api.md`: **no change**.
- `EditorRef.copy()` / `EditorRef.cut()`: signatures unchanged, **behavior changed** — with nothing selected they now leave the clipboard untouched instead of writing `#`.
- Private workspace libs: `shared-react` gained `copySelection`, `cutSelection`, `registerEmptyCopyGuard`, `$hasCopyableSelection`; `shared` exported `tableCellMarkerWithSpan`.
- Package-internal: `normalizePastedNbsp`, `stripPastedChapterAndBookId`, `$selectionToUsfmText` added; `PastePayload` narrowed to `{text, isInternal}`; `MarkerEditContext` gained `pasteRebuildArmed` / `pastePendedKeys` / `structureProtectionMode`; `isParaKindMarker` / `isCharKindMarker` moved to a new leaf module `markerKind.utils.ts`.

## Findings

None.

**scripture-editors**

- [x] **`stripPastedChapterAndBookId` destroyed pasted user content past the chapter number.** `/\\(?:c|id)(?![\w-])[^\n\\]*/g` ran to the next `\` or newline, so pasting `x \c 5 y` stripped to `x ` — the `y` silently dropped. The tokenizer takes only the next whitespace-delimited word as the chapter number (`getNextWord`, `usfmFragmentToUsj.ts:372-376`), so everything after it is ordinary content. _(fixed during review: split into `CHAPTER_TOKEN` (marker + separator + one word, stopping at a backslash) and `BOOK_ID_TOKEN` (unchanged — an `\id` line's whole remainder genuinely is payload); updated the pin in `markerPasteFidelity.test.tsx` that had encoded the loss, and added direct pins. Reverting the regex turns the new pin red with `expected 'x ' to be 'x  y'`.)_
- [x] **`Editor.tsx`'s `structureProtectionMode` pass-through had no test — deleting the line left all 4,234 tests green.** Both `Editor` and `MarkerEditPlugin` default the prop to `"off"`, so a dropped pass-through silently starves `StructureKeyboardPlugin`'s sanitizer on a protected document's paste. _(fixed during review: threaded the option through `mountStandardViewEditor` and added a real-`Editor` paste pin. Verified falsifiable — deleting `Editor.tsx:1160` turns it red while the plugin-level pin stays green. Note the option lives on `EditorOptions`, not as a top-level `<Editor>` prop.)_
- [x] **The empty-copy rule was re-implemented inline instead of reusing the predicate this branch created.** `$handleCopyForStandardView` spelled `(!selection || selection.isCollapsed())` — provably the exact negation of shared-react's `$hasCopyableSelection()` — leaving two layers whose agreement is load-bearing resting on two independently-spelled predicates in two packages. _(fixed during review: the platform layer now calls `!$hasCopyableSelection()`, which also gives that export its first consumer outside its own module.)_
- [x] **Four new marker regexes bypassed `markerName.pattern.ts` ("the single authority"), and two of the spellings already disagreed.** `AFTER_MARKER_NBSP`/`BEFORE_MARKER_NBSP` used `[a-z0-9-]` + `i`; `LEADING_MARKER_LITERAL` and the `\c`/`\id` token used `[\w-]`. `\w` includes `_`, so `\my_marker` was a marker to one and not the other — its separator NBSP survived into content as a literal `~` exactly where the tokenizer expects a space. _(fixed during review: all four derive from `ENGINE_MARKER_NAME_BYTES`. New pin goes red on revert with `expected '\my_marker~light' to be '\my_marker light'`.)_
- [x] **Spec-document breadcrumbs and provenance narration in comments** — `Deferred item 2`, `S2's own documented case`, `measured`, `confirmed by reading LexicalClipboard.dev.js`, `Regression classes closed by`, `inherited rather than introduced`, `rather than papered over`. The repo's own `CLAUDE.md:275` forbids spec section numbers in code. _(fixed during review: rules kept, provenance and IDs stripped, across `whitespaceDisplay.plugin.utils.ts`, `UnknownNode.ts`, `attributeContextPasteFidelity.test.tsx`.)_
- [x] **`$`-sigil on two pure `string → string` functions.** `CLAUDE.md:269` requires every exported `$`-prefixed function to document its calling context; neither had one, because neither has a context. _(fixed during review: renamed to `normalizePastedNbsp` / `stripPastedChapterAndBookId` across sources, tests, and the semantics doc.)_

**paranext-core**

- [ ] **The spec cannot pass against the editor `main` builds, so merge order matters.** `$stripPastedChapterAndBookId` and `$selectionToUsfmText` do not exist on `origin/platform-yalc`. CI is unaffected (`test.yml` runs only `test:e2e:smoke`), but after this merges to `main`, `npm run test:e2e:isolated scripture-editor` fails for every developer until #532 is on `platform-yalc`. **Land #532 and move `platform-yalc` first.**
- [x] **Hard-coded `Control+` chords made the spec unrunnable on macOS.** `Control+C/V/Z` and `Control+Home/End` are OS editing chords, not app accelerators — the only other bare `Control+` uses in the suite (`verse-navigation-shortcuts.spec.ts`) genuinely are accelerators. _(fixed during review: per-platform key constants following `navigation-history.spec.ts:43-45`.)_
- [x] **The copy step could pass on clipboard contents it did not write.** Nothing cleared the OS clipboard before `Control+C`, and a retry attempt that clears the poll then fails the one-shot `\p ` check leaves real USFM there for the next attempt. _(fixed during review: a sentinel is written immediately before the copy.)_
- [x] **The undo step was three negative assertions with no positive control.** An empty or mid-replacement editor satisfies all of them, and the PDP's debounced save echoes by replacing content wholesale. _(fixed during review: added `toContainText('Amittai')` as the positive control, plus the missing closer-count assertion.)_
- [x] **No negative control (the Revert Test).** _(addressed during review by static analysis rather than a rebuild: `$selectionToUsfmText` and `$stripPastedChapterAndBookId` are absent from `origin/platform-yalc` while `$handleCopyForStandardView` is present, and the pre-fix paste gate returns `false` for any payload without an NBSP. So **step 4 (`\c` strip) is the true fix-pin; steps 1–3 are characterization pins** that likely pass pre-fix too. Recorded here so nobody mistakes the spec for a four-way regression net.)_

### Minor — Consider

**paranext-core**

- [x] `.first().toBeAttached()` could not catch glyph doubling on paste _(fixed: `toHaveCount(1)` on both opener and closer — Jonah has zero `\nd` going in, so the exact count is free)_
- [x] The undo step checked only the opening span, so a half-reverted pair would pass _(fixed: both halves asserted back to 0)_
- [x] The OS clipboard leak was undocumented _(fixed: new bullet in `e2e-tests/CLAUDE.md`'s "State that leaks between runs" — it is machine-wide and **not** contained by `isolatedProjectRoot` or the temp `userDataDir`, unlike everything else the spec touches)_
- [x] `tests/isolated/README.md` documented every subset except `scripture-editor/` _(fixed; pre-existing, one line)_
- [x] Backward-facing comment: `"Yahweh" — the word used in earlier drafts of this spec` _(fixed: keeps the durable fact that scripture vocabulary cannot serve as a sentinel, drops the drafting history)_
- [ ] **The `Control+Z` is not synchronized with the PDP's ~700ms debounced save/echo.** Mitigated by the positive control added above, but the ordering hazard is real and unresolved. See "In-Review Quality Check" for what two live runs showed.
- [ ] ~~The `press('Home')` between paste and undo should be dropped~~ _(dismissed: settles are history-merged via `$settleWithoutOwnUndoEntry`, which the analyzer itself verified, so it cannot split the undo step; the comment already marks it belt-and-braces.)_
- [ ] ~~`dev-packages.json` redirects every developer install and CI leg to a feature branch~~ _(refuted: working-tree-only — `git diff <MERGE_BASE> HEAD -- dev-packages.json` is empty, nothing staged. Both core analyzers looked at this and the second reached the same conclusion. Deliberately excluded from the review commit.)_

**scripture-editors**

- [x] `$hasCopyableSelection` was an export with no consumer outside its own module _(resolved by the reuse fix above)_
- [x] `isParaKindMarker` / `isCharKindMarker` were byte-identical but for one `MarkerType` comparison, in a module that exists precisely so the two tiers cannot disagree _(fixed: one private `isKindMarker` body, two thin wrappers; the both-kinds-true unknown fallback is preserved and now documented)_
- [x] `$selectionToUsfmText` mirrors `RangeSelection.getTextContent()` without naming the version it was mirrored from _(fixed: pinned to lexical 0.43.0 with an instruction to re-check on upgrade)_
- [x] **The clipboard semantics doc lived in a gitignored directory.** `.gitignore:67` ignores `docs/superpowers`; the file was tracked only by a force-add, so future edits to the durable contract would never appear in `git status`. _(fixed: `git mv` to `docs/clipboard-semantics.md`, alongside `docs/standard-view-invariants.md`, matching the repo's flat topic-named docs; the one code reference now carries the real path.)_
- [x] Deferred item 6 pointed at `docs/superpowers/specs/2026-07-07-standard-view-followups.md`, which does not exist _(fixed: replaced with a plain "tracked separately")_
- [x] The test-mapping table still described the E2E spec as "written, not yet run in this task" _(fixed: it runs and passes)_
- [ ] ~~`UnknownNode.ts`'s new block is 88% comment~~ _(dismissed: the block encodes a genuinely subtle invariant — why child-count is the test, and why excluding a wrapper makes `$appendNodesToJSON` hoist its children into a convincing lie. This repo is deliberately comment-heavy and the reasoning is not inferable from the code.)_
- [ ] `$noteCallerText`'s `caller || GENERATOR_NOTE_CALLER` fallback is untested — no fixture has a note without an explicit caller
- [ ] `$isNoteInternalDisplaySeparator` and `$collapsedNoteCategoryBytes` run the same three-step probe, differing only in the last check's sign — deferred as a refactor beyond this PR's scope
- [ ] `$handlePasteForStandardView` has four positional params, three optional only so tests can call it with one — an options object would read better; deferred
- [ ] `whitespaceDisplay.plugin.utils.ts` is now 787 lines with 10 of 11 exports being clipboard concerns — splitting out `standardViewClipboard.utils.ts` would match the module boundary to the concern; deferred (large diff churn on a branch already under review)
- [ ] **`EditorRef.copy()`/`cut()` behavior change deserves a release note** — a host that calls `copy()` at a collapsed caret used to get a clipboard write (`#`) and now gets nothing. Strictly better, but there is no compile-time signal for a consumer that built a workaround.

### Template Propagation

#### Shared Regions Modified

None. No changed file in either repo contains a `#region shared with` marker.

#### Extension Config Changes

- [n/a] paranext-core: no files under `extensions/` changed.
- [n/a] scripture-editors: no `extensions/` tree, and no build/tooling config changed.

## Positive Observations

- **Load-bearing claims in the doc comments hold up against the installed Lexical source.** An analyzer independently re-derived them: `$appendNodesToHTML` returns early on a null `element` before consuming the `excludeFromCopy` value; `$appendNodesToJSON` passes `'html'` and hoists excluded nodes' children; `copyToClipboard(editor, null)` really does append `createTextNode('#')` while `$copyToClipboardEvent` bails before `preventDefault()` on a collapsed selection. That is the exact `#`-leak mechanism the guard closes.
- **Falsifiability across the new tests is unusually good.** Copy-side pins assert exact full USFM strings the old blanket `getTextContent().replaceAll(NBSP, " ")` cannot produce, and three places carry explicit negative controls that isolate the fix as the only variable (`tier2Rebuild.utils.test.tsx:405`, `markerPasteFidelity.test.tsx:521`, `attributeContextPasteFidelity.test.tsx:251`).
- **The 2SA fixtures are exact regeneration output.** Re-running `GENERATE_TEST_DATA=1` plus prettier reproduced the committed files byte-identically, and `2sa.lexical.hidden.ts` correctly shows no change.
- Putting the empty-copy guard on `COPY_COMMAND`/`CUT_COMMAND` at `COMMAND_PRIORITY_LOW` is the right seam — one registration covers the shortcuts, the context menu and `EditorRef`, and it verifiably sits above rich-text's `EDITOR`-priority fallback and below Standard view's own `HIGH` claim.
- `markerKind.utils.ts` is a genuinely good extraction: it removes the Tier 1 ↔ Tier 2 import cycle and documents why a narrower `type === MarkerType.Paragraph` check would have been wrong.
- The e2e spec's step 4 is exemplary — a mid-line `${token} \c 99 ` payload matching the deliberately-unanchored strip regex, with three independent probes, one of them (the toolbar reference) read entirely outside the editor DOM.
- Every locator in the spec resolves against real editor code, and `nd` is a genuinely clean sentinel (the bundled Jonah SFM contains zero `\nd`), so the attach and zero-count assertions are non-vacuous.
- No `eslint-disable`, `@ts-expect-error`, type assertions, `console.*`, `.only`, or untracked TODOs added anywhere in either diff.

## Interview Notes

Run in automated mode at the author's request — no interview was conducted. The author asked for each finding to be considered on its merits, verified, and then fixed if small-to-medium, low-risk and worthwhile; otherwise refuted or dismissed with a reason. Judgment calls made autonomously:

- **Fixed** anything that was a verified correctness defect, a stated-convention violation, or a missing/weak test — including two defects the existing 4,200-test suite could not see (the `\c` over-strip and the untested `Editor` pass-through), both now pinned with tests confirmed to go red on revert.
- **Deferred** pure refactors with large diff churn and no correctness content (the module split, the options-object signature, the note-separator merge) as out of scope for a branch already in review.
- **Did not re-home PR #532.** Active scripture-editors development moved to `paranext/scripture-editors` under PT-4500 — that repo carries the `pt-4313`, `pt-4370` and `pt-4615` branches and commits `packages/*/dist` — while #532 still targets `eten-tech-foundation`. Moving a PR across repos is the author's call and the transfer guide is theirs to exercise.
- **Did not widen shared e2e timeouts** in response to the flake found below; the gate is shared infrastructure used by every isolated spec.

## In-Review Quality Check

**scripture-editors** — all gates green after every change:

- `nx run-many -t test` (`--skip-nx-cache --pool=forks --maxWorkers=4`): **shared 697 passed · shared-react 1674 passed / 1 skipped · platform-editor 1864 passed / 4 skipped**. Baseline before this review was 697 / 1674+1 / 1857+4, so the review added **7 tests** and broke none.
- `nx run-many -t typecheck lint`: clean — **0 errors**. The 28 lint warnings are pre-existing `no-console` / `no-loop-func` in files this branch does not touch (cross-referenced against the changed-file list).
- `nx format:check`: clean.
- Each new pin was **verified falsifiable** by reverting its fix and watching it go red with the expected message.

**paranext-core**:

- `tsc -p e2e-tests/tsconfig.json`, `eslint`, `prettier --check`: all clean.
- **The e2e spec was run twice against a live app.** Run 1: **passed, 4.0m**. Run 2: **failed then passed on retry — Playwright reports it flaky.**
- **The flake is not in the clipboard logic.** It failed at `clipboard-usfm-round-trip.spec.ts:77` (`makeSampleProjectEditable()`), in the *shared* backend-readiness gate at `e2e-tests/fixtures/helpers.ts:942`: `PAPI method "object:platform.Paratext-pdpf.getProjectDataProviderId" not listed in rpc.discover within 120000ms`. That is the C# data provider failing to register its PDP factory in time on a machine under heavy load, before any clipboard step runs. It affects every isolated spec that calls that helper, and is not caused by this branch.
- Worth noting for the budget: `test.slow()` gives 360s, and a passing run took 240s while the setup gate alone can consume 120s. The margin is thinner than the previously-recorded 1.7m suggests.

## Suggested Review Focus

- [ ] **Merge order is a hard constraint**: scripture-editors #532 must land *and* reach `platform-yalc` before paranext-core #2656, or `npm run test:e2e:isolated scripture-editor` goes red on `main` for every developer. CI will not catch this (only `test:e2e:smoke` runs there).
- [ ] **Which repo should #532 land in?** PT-4500 moved active development to `paranext/scripture-editors`; this PR still targets `eten-tech-foundation`. Transferring it also means rebuilding the committed `packages/*/dist` that the paranext repo carries and `eten-tech-foundation` does not.
- [ ] **The narrowed `\c` strip is a deliberate behavior change.** `x \c 5 y` now keeps the `y`. Confirm that keeping trailing content is what you want in every paste shape, not just the tested ones.
- [ ] **The e2e flake.** It is a pre-existing shared-fixture readiness gate, but this spec's 360s budget leaves less headroom than the others. Decide whether to raise the gate, or accept the retry.
- [ ] **`EditorRef.copy()`/`cut()`** is a host-visible behavior change with no compile-time signal — does it need a release note?
- [ ] Remaining documented deferrals in `docs/clipboard-semantics.md` (items 10–14: the stranded-decorator class across figure/table/sidebar/periph/ref, table/milestone corpus losses, the dropped html-carrier NBSP rule) are unchanged by this review and still open.

tjcouch-sil:
**Rebased onto `origin/main`** (was 16 commits behind) — clean, 5 commits replayed, all three changed files intact. Gates green: `tsc -p e2e-tests/tsconfig.json`, `eslint`, `prettier --check`.

**The editor half moved repositories.** scripture-editors#532 is closed; the work now lives at **https://github.com/paranext/scripture-editors/pull/13**, under `paranext`, where Paratext 10 develops `platform-editor` — matching this repo's own `dev-packages.json`, whose `cloneUrl` the PT-4500 flip already pointed there.

**Merge order still applies, unchanged:** paranext/scripture-editors#13 must land *and* reach `platform-yalc` before this PR. The spec asserts behavior (`$stripPastedChapterAndBookId`, `$selectionToUsfmText`) that does not exist on `platform-yalc` today, so merging this first turns `npm run test:e2e:isolated scripture-editor` red on `main` for every developer. CI here will not catch that — `test.yml` runs only `test:e2e:smoke`.

.context/standards/Architecture-Decisions.md
```diff
@@ -3172,6 +3172,46 @@ step, no automation. Just a record.
...
+  host's own content re-tagged. It was added to un-skip a copy→paste round-trip test whose target is
+  an empty `\p` scaffold, and the "stray empty paragraph" it removed was that scaffold; it was never
+  recorded or reviewed as a product decision. Keeping it out of TYPED input then required a paste
+  provenance apparatus (a per-commit armed flag plus a set of pended node keys, to carry "this
+  rebuild came from a paste" across a deferred settle), which in turn let the read-only settle and
+  the mutating rebuild disagree about a deferred paste's shape.
+- **Decision:** A paste inserts what was pasted and nothing else. It never deletes a marker the user
+  did not select, and it never invents one they did not paste. Two paragraph markers in a row is what
+  the bytes say, and the line splits — leaving an empty predecessor — exactly as typing the same
+  bytes does. This matches Paratext 9, whose `NormalizeTokenUsfm` (`ParatextData/UsfmToken.cs`) emits
```
lyonsil, line 3193: **#4** - low · checked and confirmed

The decisions log states that Paratext 9 emits a line break before every Paragraph token, which its source does not do unconditionally.

**What happens:** `.context/standards/Architecture-Decisions.md:3193-3194` cites `NormalizeTokenUsfm` (`ParatextData/UsfmToken.cs`) as emitting "a line break before every Paragraph token", and uses that as the Paratext 9 parity warrant for the decision. In the method the `\r\n` is appended only when `dest.Length > 0` and `tokensHaveWhitespace` is false (`UsfmToken.cs:838-844`), so the first token in a document gets none and the whitespace-preserving mode never inserts one.

**Why it matters:** the omitted condition cannot reach the case the sentence is warranting - two paragraph markers in a row always have content before the second one and always take the default call path (`UsfmToken.cs:826`), while the whitespace-preserving mode is reached only from `ConvertToParatext7Format` (`UsfmToken.cs:659`), a USFM 3.0-to-2.0 conversion unrelated to this decision. So this is a precision gap in a durable record rather than a wrong warrant - but `.claude/rules/docs-durability.md` asks that a behaviour stated in a living doc be checked against the live source.

**Fix:** in `.context/standards/Architecture-Decisions.md`, replace the sentence "This matches Paratext 9, whose `NormalizeTokenUsfm` (`ParatextData/UsfmToken.cs`) emits a line break before every Paragraph token." with: "This matches Paratext 9, whose `NormalizeTokenUsfm` (`ParatextData/UsfmToken.cs`) writes a line break before a Paragraph token whenever the output already has content and the tokens are not already carrying their own whitespace (its legacy USFM-2.0-conversion mode skips this) - exactly the two-markers-in-a-row case here." The existing citation already gives file plus symbol with no line number, so it needs no other change.

_Other findings in this file: #5_

.context/standards/Architecture-Decisions.md
```diff
@@ -3172,6 +3172,46 @@ step, no automation. Just a record.
...
+  paragraph), P9 keeps the empty paragraph anyway, and it would have preserved the whole provenance
+  apparatus for a case nobody reaches. Keep the dedup and fix the settle divergence instead —
+  rejected: it fixes the symmetry of a rule that should not exist.
+- **Consequences:** pasting a whole-paragraph copy at a paragraph's content start now leaves an empty
+  paragraph ahead of the pasted one, which is a visible change and the P9-parity shape. Round-trip
+  tests that seeded an empty host and expected the paste to swallow it now assert the host and strip
+  it. `isPasteRebuild`, `Tier2Context.pasteRebuildArmed` and `pastePendedKeys` are gone, so the
+  read-only settle (`$settledUsj`) and the mutating rebuild call `$buildParaFragment` with identical
+  arguments and can no longer disagree — Invariant IV holds by construction rather than by
+  bookkeeping. Removing the dedup also surfaced a defect it had masked: every `\b` in a pasted
```
lyonsil, line 3209: **#5** - low · checked and confirmed

The decisions log reports a defect around `\b` without saying, where it reports it, whether that defect is fixed.

**What happens:** `.context/standards/Architecture-Decisions.md:3209-3211` records that every `\b` in a pasted document gained an empty `\p` in front of it - 16 in the 2sa fixture - as a bare observation. The entry's Status is "Accepted" (:3178) and nothing beside the observation states its disposition.

**Why it matters:** the Decision bullet three sentences earlier (:3195-3197) already carries the fix - the injection skip covers a paragraph whose own text already opens with a paragraph-marker literal, and `\b`'s text is nothing but that literal - so a reader taking the bullets in order is not left without an answer. A reader skimming only Consequences is, and this log elsewhere pairs "surfaced" with "and fixed" (:149).

**Fix:** in the Consequences bullet, tie the case to the rule already stated in the Decision bullet instead of leaving it as a bare observation - end "...never resolved the injected prefix away" with "- closed by the same injection-skip rule above (`$suppliesOwnParaMarker` in `markerEditDeletion.utils.ts`): `\b`'s own text is nothing but the marker literal, so it qualifies via that rule's end-of-run case." Per `.claude/rules/docs-durability.md`, date the status claim and describe the carrying work accurately - scripture-editors #13 is OPEN as of 2026-09-18, so do not write that it is merged.

_Other findings in this file: #4_

e2e-tests/tests/isolated/scripture-editor/clipboard-usfm-round-trip.spec.ts
```diff
@@ -0,0 +1,551 @@
...
+}
+
+/**
+ * The on-screen box of `locator`, failing with `description` instead of handing back a null box.
+ *
+ * `boundingBox()` reports MAIN-FRAME coordinates even for an element inside an iframe — the very
+ * space `mainPage.mouse` works in — so a drag between two elements of the editor's web view needs
+ * no iframe-offset arithmetic of its own.
+ */
+async function requireElementBox(locator: Locator, description: string): Promise<ElementBox> {
```
lyonsil, line 136: **#7** - low · checked and confirmed

A drag can be aimed at coordinates outside the window, and the failure then reports only a clipboard timeout.

**What happens:** `requireElementBox` (`clipboard-usfm-round-trip.spec.ts:136-140`) rejects only a `null` box. `boundingBox()` returns real main-frame coordinates for a rendered element even when it is scrolled out of view, so `figure.scrollIntoViewIfNeeded()` at :365 centring the FIGURE does not guarantee the `\nd*` tail glyph measured at :367-370 is on screen. `mainPage.mouse.move` at :375-381 then dispatches at an off-screen point, selects nothing, and the copy leaves the sentinel in place until the poll at :387-395 times out at 15s.

**Why it matters:** neither of the two call sites triggers this today - `scrollIntoViewIfNeeded` uses CDP's centring variant, and both the tail glyph and the attribute run sit immediately next to the figure in document flow - so this is a latent gap in the helper's diagnostics rather than a live failure. When it does fire, the run costs 15s and reports "expected to contain ... received CLIPROUNDTRIPSENTINELNOTCOPIED", which names the clipboard rather than the cause.

**Fix:** make `requireElementBox` reject a box that falls outside the window - compare against `locator.page().evaluate(() => ({ width: innerWidth, height: innerHeight }))`, which needs no `mainPage` parameter because `locator.page()` resolves to the same page - and throw naming `description`, so an off-screen drag target fails as an off-screen drag target instead of as a clipboard timeout.

_How this was checked: `Locator.boundingBox()` (backed by CDP `DOM.getBoxModel` in `node_modules/playwright-core/lib/server/chromium/crPage.js:865-878`) only returns `null` when the box model itself is unavailable - for example when the node is not rendered at all - not because the element is scrolled out of the visible viewport; a rendered element that is merely off-screen still yields real coordinates. `requireElementBox` (`clipboard-usfm-round-trip.spec.ts:136-140`) checks only for that `null` case, so a drag endpoint computed from a box outside the window would pass through unflagged. `figure.scrollIntoViewIfNeeded()` at :365 and :409 uses CDP's centring variant, which in practice keeps the immediately-adjacent tail glyph (:367-370) and attribute run (:421-423) on screen because both sit right next to the figure in document flow - so this is a latent gap in the helper's diagnostics rather than a live failure at either call site today. If it did trigger, the fallout is as described: a silent off-screen drag, no selection made, and a 15s poll timeout reporting the clipboard sentinel instead of the real cause._

_Other findings in this file: #1, #2, #9, #10, #11_

e2e-tests/tests/isolated/scripture-editor/clipboard-usfm-round-trip.spec.ts
```diff
@@ -0,0 +1,551 @@
...
+        timeout: 20_000,
+      });
+      await expect(editorFrame.locator('span.closing[data-marker="nd"]')).toHaveCount(1, {
+        timeout: 20_000,
+      });
+    });
+
+    await test.step('undoing the paste is a single step', async () => {
+      await editorInput.press(UNDO_KEY);
+      await expect(editorInput).not.toContainText(PASTE_TOKEN, { timeout: 20_000 });
```
lyonsil, line 269: **#1** - medium · checked and confirmed

The undo step reports success even when the undo never happened, if the editor is momentarily empty when the step runs.

**What happens:** `UNDO_KEY` is pressed at `clipboard-usfm-round-trip.spec.ts:268`, then three negative assertions run at :269, :271 and :274. `expect(locator).not.toContainText(...)` returns as soon as a single poll satisfies it, so a transient empty or mid-replacement editor satisfies all three at once. The positive control meant to rule that out - `toContainText('Amittai')` - runs afterwards at :282, by which time the content is back, so it passes too and the step goes green having measured nothing.

**Why it matters:** this step is the only guard that a paste and the rebuild it triggers collapse into one undo. `type-through-save-echo.spec.ts:86-92` and `formatted-default-simple-mode.spec.ts:46-48` both put the positive control before the negatives for exactly this hazard; the comment at :277-281 names the hazard but places the fix where it cannot close it.

**Fix:** move the `toContainText('Amittai', { timeout: 20_000 })` assertion from :282 to above :269, so the step polls past any empty window before the negative assertions run.

_How this was checked: Playwright's `expect(locator).not.toContainText(...)` takes an immediate one-shot read and returns success the instant the condition holds - it does not require the state to be stable (`node_modules/playwright-core/lib/server/frames.js:1151-1179`: `_expectInternal(..., true)` fires before the retry loop, and `retryWithProgressAndTimeouts` at :1169 returns as soon as a single poll satisfies the condition). At `clipboard-usfm-round-trip.spec.ts:268-276` the three negative assertions run immediately after the undo keystroke, and the comment at :277-281 correctly identifies that a transient empty or mid-replacement editor satisfies all three vacuously - but the positive control that would rule that out sits at :282, after the hazard it is meant to guard against, so it proves nothing about the state the negatives actually measured. `type-through-save-echo.spec.ts:86-92` establishes the same ordering convention explicitly, and `formatted-default-simple-mode.spec.ts:46-48` repeats it._

_Other findings in this file: #2, #7, #9, #10, #11_

lyonsil (commented):
A review of the new Playwright clipboard round-trip spec and the decisions-log entry added alongside it. 8 findings: 2 medium, 6 low.

Each comment carries the severity the reviewer settled on, plus a status - `checked and confirmed` means the finding was verified against the code on this branch. Every finding ends with a `Fix:` naming the file and the symbol the change should land on, so each one can be acted on without reading the others.

e2e-tests/tests/isolated/scripture-editor/clipboard-usfm-round-trip.spec.ts
```diff
@@ -0,0 +1,551 @@
...
+            return clipboardText;
+          },
+          { timeout: 15_000 },
+        )
+        .toContain(FIGURE_USFM);
+      // The press landed a few pixels INTO the opener glyph, so the copied bytes can only begin with
+      // the glyph's first byte if the engine grew the selection out over the whole read-only glyph
+      // rather than starting mid-glyph — the poll above proves the run is complete, this proves
+      // where it starts.
+      expect(clipboardText.startsWith('\\fig ')).toBe(true);
```
lyonsil, line 400: **#9** - low · checked and confirmed

When the figure-opener copy starts at the wrong byte, the failure message says only "expected true, received false".

**What happens:** `expect(clipboardText.startsWith('\\fig ')).toBe(true)` at `clipboard-usfm-round-trip.spec.ts:400` collapses the clipboard contents to a boolean before the matcher sees them, so nothing about the actual bytes reaches the report. The sibling step deliberately avoids this at :456 (`clipboardText.slice(-FIGURE_ATTRIBUTE_BYTES.length)` compared with `toBe`) and says so in its comment at :452-455.

**Why it matters:** this is the assertion that pins WHERE the copied run starts - the step's whole point - so when it fails it is the one place a reader needs the bytes.

**Fix:** change :400 to `expect(clipboardText.slice(0, '\\fig '.length)).toBe('\\fig ')`, deriving the slice length from the literal rather than hardcoding `5`, matching the length-derived slicing already used at :456.

_How this was checked: `expect(clipboardText.startsWith('\\fig ')).toBe(true)` at `clipboard-usfm-round-trip.spec.ts:400` does collapse the copied bytes into a boolean before the matcher runs, so a failure here only ever reports "expected true, received false" with no indication of what the clipboard actually held. The sibling assertion at :456 (`clipboardText.slice(-FIGURE_ATTRIBUTE_BYTES.length)).toBe(FIGURE_ATTRIBUTE_BYTES)`) avoids exactly this by comparing the string itself, and its own comment at :452-455 states that is the point. The whole file is new in this change, so :400 is itself newly authored rather than pre-existing code the change merely touches._

_Other findings in this file: #1, #2, #7, #10, #11_

e2e-tests/tests/isolated/scripture-editor/clipboard-usfm-round-trip.spec.ts
```diff
@@ -0,0 +1,551 @@
...
+      // Derived from the opener's box because the caption is a bare text node with no box of its
+      // own; deriving it that way also survives the figure's bytes wrapping onto a second line,
+      // where a point interpolated toward the attribute run would land on the wrong line entirely.
+      // A keyboard selection cannot stand in for this: the mouse path through a read-only glyph is
+      // what the behaviour is about.
+      await dragSelect(
+        mainPage,
+        openerBox.x + openerBox.width + 3,
+        openerBox.y + openerBox.height / 2,
+        attributeBox.x + attributeBox.width / 2,
```
lyonsil, line 437: **#2** - medium · checked and confirmed

The step that measures a selection ending inside a figure's attribute run can aim its drag release outside that run entirely.

**What happens:** :421-424 reads `figureAttributeRun.boundingBox()` and :437-438 releases the drag at that box's centre. `boundingBox()` resolves through CDP's `DOM.getBoxModel`, whose border quad is the UNION of an inline element's per-line client rects - so when `|src="avnt016.jpg" size="span"` (30 bytes, :73) wraps onto two lines, the union's vertical centre lands on the seam between them, on neither line's glyphs.

**Why it matters:** :428-431 already derives the START point from the opener's box specifically to survive that wrap; the end point has no equivalent protection. `attribute-display-settle.spec.ts:143-146` picks single-visual-line verses at this same 1280x800 window to dodge soft-wrap ambiguity, so the team already treats wrapping as live at this size. Off target, the step fails at the 15s poll (:443-451) reporting only that the clipboard still holds the sentinel.

**Fix:** derive the release point from ONE client rect of the run, then add the editor iframe's own `boundingBox()` offset before handing it to `mainPage.mouse`. `getClientRects()` read through `Locator.evaluate()` returns IFRAME-relative coordinates and does not do the translation `boundingBox()` does (:131-134), so dropping that offset mistargets a different way. Read the rects with `figureAttributeRun.evaluate((el) => Array.from(el.getClientRects()).map((r) => ({ x: r.x, y: r.y, width: r.width, height: r.height })))`, take the last, and release at `iframeBox.x + rect.x + rect.width / 2`, `iframeBox.y + rect.y + rect.height / 2`.

_How this was checked: `Locator.boundingBox()` calls into `DOM.getBoxModel` (`node_modules/playwright-core/lib/server/chromium/crPage.js:865-878`), whose `border` quad matches `getBoundingClientRect()` semantics - the union of an inline element's per-line client rects, not the first one. A headless-Chromium probe against a wrapped inline span nested in an iframe confirmed this: `getClientRects()` returned two 22px line rects (y:70-92 and y:93-115, iframe-relative) while `boundingBox()` returned a single 45px box whose vertical centre sits exactly on the seam between them, landing on neither line's glyphs. `clipboard-usfm-round-trip.spec.ts:421-424` reads that union box and :437-438 aims the release at its centre, so a two-line wrap of the 30-byte attribute run at :73 reproduces the mistarget. `attribute-display-settle.spec.ts:143-146` in the same suite deliberately picks short single-visual-line verses at the same 1280x800 window to dodge soft-wrap ambiguity, and this spec's own comment at :424-427 defends the start point against exactly this failure mode while leaving the end point undefended._

_Other findings in this file: #1, #7, #9, #10, #11_

e2e-tests/tests/isolated/scripture-editor/clipboard-usfm-round-trip.spec.ts
```diff
@@ -0,0 +1,551 @@
...
+      // the bytes actually were when it does not.
+      expect(clipboardText.slice(-FIGURE_ATTRIBUTE_BYTES.length)).toBe(FIGURE_ATTRIBUTE_BYTES);
+      // Nor did it over-expand past the run to the `\fig*` closer, which sits immediately after it.
+      expect(clipboardText).not.toContain('\\fig*');
+      // The caption side is a real, non-empty tail of the caption: the attribute run is pinned to the
+      // end above, so these are exactly the bytes between where the drag started and where the run
+      // begins, and `toContain` then holds only for a genuine substring of the caption.
+      const copiedCaption = clipboardText.slice(0, -FIGURE_ATTRIBUTE_BYTES.length);
+      expect(copiedCaption).not.toBe('');
+      expect(FIGURE_CAPTION_TOKEN).toContain(copiedCaption);
```
lyonsil, line 464: **#10** - low · checked and confirmed

The caption check in the attribute-run step passes on a single copied character.

**What happens:** :462 slices everything before the attribute bytes into `copiedCaption`, :463 requires it non-empty, and :464 asserts `expect(FIGURE_CAPTION_TOKEN).toContain(copiedCaption)` - satisfied by ANY substring of `CLIPROUNDTRIPFIG`, including one that could never arise from this drag.

**Why it matters:** low on its own, since the byte-exact tail comparison at :456 still pins the load-bearing "whole attribute run" behaviour. But the comment at :459-461 presents :464 as showing these are "exactly the bytes between where the drag started and where the run begins", and a substring check does not show that.

**Fix:** replace the membership check at :464 with a suffix check plus a minimum length - `expect(FIGURE_CAPTION_TOKEN.endsWith(copiedCaption)).toBe(true);` and `expect(copiedCaption.length).toBeGreaterThan(1);`. A suffix is the correct relation here: the attribute run is pinned intact at its end (:456) and the drag's mouse-down point can only land at or after the caption's first character, so any imprecision truncates the caption from the front, never the back.

_How this was checked: The comment at `clipboard-usfm-round-trip.spec.ts:459-461` reads "The caption side is a real, non-empty tail of the caption… these are exactly the bytes between where the drag started and where the run begins" - a claim about tail structure, backed by an assertion at :464 (`expect(FIGURE_CAPTION_TOKEN).toContain(copiedCaption)`) that only checks substring membership. Geometrically the claim about the value is right: since the attribute run is pinned intact at its end (:456) and the caption is ordinary editable text at its start, any imprecision in the drag's start x-coordinate can only truncate `copiedCaption` from the front, so on every legitimately-passing run it is in fact a genuine suffix of `FIGURE_CAPTION_TOKEN` - but `toContain` does not verify that; it would equally accept a non-suffix substring such as "OUND" that could never legitimately arise from this drag. Confirmed by direct simulation over click offsets 0 through 16 into "CLIPROUNDTRIPFIG": every resulting value satisfies `endsWith`, while a synthetic non-suffix passes `toContain` and fails `endsWith`._

_Other findings in this file: #1, #2, #7, #9, #11_

e2e-tests/tests/isolated/scripture-editor/clipboard-usfm-round-trip.spec.ts
```diff
@@ -0,0 +1,551 @@
...
+          { timeout: 15_000 },
+        )
+        .toContain(NOTE_USFM);
+      // The html is the USFM bytes, not an export of the editor's own DOM. Every DOM export of a note
+      // caller carries `data-caller` (ImmutableNoteCallerNode.exportDOM), so its absence is what
+      // separates the two — and it is the one attribute that cannot survive a bytes-only flavour.
+      expect(clipboardHtml).not.toContain('data-caller');
+      // Both flavours say the same thing, so a consumer's flavour preference cannot change the
+      // document it receives.
+      const clipboardText = await electronApp.evaluate(({ clipboard }) => clipboard.readText());
```
lyonsil, line 517: **#11** - low · checked and confirmed

The two flavour checks in the footnote step normalize whitespace differently, and nothing in the spec says why.

**What happens:** the html poll at `clipboard-usfm-round-trip.spec.ts:498-510` compares against `NOTE_USFM` only after normalizing `&nbsp;` and `\u00a0` to spaces (:506), with a comment explaining that a clipboard intermediary can re-serialize them. The plain-flavour check that closes the step reads `clipboard.readText()` raw at :517 and compares the same byte string at :518 with no normalization and no comment.

**Why it matters:** the asymmetry is correct and deliberate - the editor's copy handler builds `text/plain` with an unconditional NBSP-to-space replace, so that flavour cannot carry one, while `text/html` deliberately keeps NBSP at fragment edges and in runs of two or more for rich-text fidelity. Nothing at :517 records that, so the next reader sees two comparisons of the same bytes done two ways and cannot tell whether one is an oversight.

**Fix:** leave :517-518 unnormalized and add a comment above :517 saying why: `text/plain` is built via an unconditional NBSP-to-space replace in the editor's copy handler (`$getStandardViewClipboardData`, `whitespaceDisplay.plugin.utils.ts`), so it is guaranteed never to carry one and is deliberately compared raw - unlike `text/html`, which keeps NBSP at fragment edges and in runs of two or more and needs the :506 normalization for that reason. Do not unify the two checks: normalizing the text flavour would hide a real regression if that replace ever stopped running.

_Other findings in this file: #1, #2, #7, #9, #10_

e2e-tests/tests/isolated/scripture-editor/clipboard-usfm-round-trip.spec.ts
lyonsil: Refinement on the `Fix:` above - the suffix check reads better as a value comparison, so a failure shows the bytes instead of `expected true, received false`:

```ts
expect(copiedCaption).toBe(FIGURE_CAPTION_TOKEN.slice(-copiedCaption.length));
expect(copiedCaption.length).toBeGreaterThan(1);
```

The first line is exactly equivalent to `endsWith`: `"OUND"` gets expected `"PFIG"`, `"XCLIPROUNDTRIPFIG"` gets expected `"CLIPROUNDTRIPFIG"`, and an empty string gets the whole token back, because `slice(-0)` is `slice(0)`. It also matches the shape already used at :456 (`actual.slice(-CONST.length)).toBe(CONST)`), sliced from the other end.

The second line is doing this finding's real work rather than decorating it: `"G"` passes the suffix check on its own (expected `"G"`, received `"G"`), and a one-character copy is exactly the case in question.

If the drag reliably copies the whole caption, the stronger assertion is simply `expect(copiedCaption).toBe(FIGURE_CAPTION_TOKEN)` - one line, no floor needed. Whether the 3px offset at :435 lands in the first half of the caption's opening glyph (caret snaps before it, whole caption selected) or the second half (snaps after, leaving `"LIPROUNDTRIPFIG"`) is font-metric dependent and not settled here. Logging `copiedCaption` on a single run settles it; if it is the full token, prefer exact equality.

e2e-tests/tests/isolated/scripture-editor/clipboard-usfm-round-trip.spec.ts
tjcouch-sil: Fixed in bea33ed9ce1 — the `Amittai` positive control now runs right after the undo key, before the three negatives, with the comment moved to explain why it has to come first.

e2e-tests/tests/isolated/scripture-editor/clipboard-usfm-round-trip.spec.ts
tjcouch-sil: Fixed in bea33ed9ce1 — the release point is now the centre of the run's LAST client rect (read through `Locator.evaluate`), offset by the editor iframe's own box (`editorFrame.owner()`); it throws if the run has no rects. The start point is unchanged.

.context/standards/Architecture-Decisions.md
tjcouch-sil: Fixed in bea33ed9ce1 with your wording. The same over-claim was also in a scripture-editors test comment and in `docs/clipboard-semantics.md`; both are corrected in scripture-editors #13 (b17f235c).

.context/standards/Architecture-Decisions.md
tjcouch-sil: Fixed in bea33ed9ce1 — the Consequences bullet now ties the `\b` case to `$suppliesOwnParaMarker`'s end-of-run case, with no merge claim.

e2e-tests/tests/isolated/scripture-editor/clipboard-usfm-round-trip.spec.ts
tjcouch-sil: Fixed in bea33ed9ce1 — `requireElementBox` now rejects a box whose centre falls outside the window (read via `locator.page()`), naming the target; a box only partly clipped at the edge, with its centre on screen, still passes.

e2e-tests/tests/isolated/scripture-editor/clipboard-usfm-round-trip.spec.ts
tjcouch-sil: Fixed in bea33ed9ce1 — `expect(clipboardText.slice(0, '\\fig '.length)).toBe('\\fig ')`.

e2e-tests/tests/isolated/scripture-editor/clipboard-usfm-round-trip.spec.ts
tjcouch-sil: Fixed in bea33ed9ce1 with your refinement: `expect(copiedCaption).toBe(FIGURE_CAPTION_TOKEN.slice(-copiedCaption.length))` plus the length floor. I kept the floor rather than exact equality, since whether the 3px offset lands before or after the caption's first glyph is font-dependent, as you noted.

e2e-tests/tests/isolated/scripture-editor/clipboard-usfm-round-trip.spec.ts
tjcouch-sil: Fixed in bea33ed9ce1 — comment above the raw `text/plain` read: that flavour is built with an unconditional NBSP-to-space inversion (`$selectionToUsfmText`), so it is compared raw on purpose, unlike the html poll.

tjcouch-sil:
Thanks — all eight findings are fixed in bea33ed9ce1 (replies on each thread). The spec passed against an editor build containing scripture-editors #13's review fixes; the first attempt hit the known cold-start `useWebViewState` crash at its first paste step and the automatic retry went green through every step, including the reworked drag and undo steps.

_(AI-assisted)_

lyonsil (approved):
:lgtm:

> @lyonsil reviewed 4 files and all commit messages, made 1 comment, and resolved 8 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2656#-:-P2608LL82615ManFoz-:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2660, opened 2026-08-10
Mount the character-marker control in the Simple editor's gutter

jolierabideau:
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2660)

---

# Code Review Summary

**Branch**: mount-character-control-in-gutter

**Base**: origin/main

**Date**: 2026-08-10

**Files changed**: 18 (against `origin/main` after merging it in; the review began against the stale merge base `a6ddb13`, where 37 files differed)

## Overview

Implements `PT-XXX-D2` — "Variant B, Side" placement from the character-level markers plan: the
character-marker control is mounted in a reserved gutter on the inline-end edge of the Simple-mode
editor's text column, tracking the line the caret is on. The gutter space is *reserved* rather than
floated over, via a Simple-scoped `padding-inline-end` in `_simple-mode.scss`, so "the bar never
overlaps project text" is a property of the construction rather than arithmetic that happens to hold.
Only the vertical axis is computed in JS; horizontal placement is `inset-inline-end: 0`, which mirrors
for RTL and is immune to paragraph indentation. Serves the PRD's non-negotiable N1 (add a character
marker) and H1.

The review's headline finding was that the branch was cut before PR #2648 landed a **later revision
of the same work** on `main`, as a directory rather than a flat module — which git would have merged
into two coexisting `CharacterMarkerControl` implementations with no conflict marker. That merge has
been done and the D2 work re-targeted onto `main`'s `character-marker-control/`. Three analyzer
findings were resolved by the merge itself, since `main`'s revision already fixed them.

## API Changes

- `lib/platform-bible-react` — **no net change against `origin/main`.** The branch's earlier
  `MarkerMenuItem.selectionState` variant was superseded by `main`'s, which is a superset
  (`selectionState`, plus `isDisabled`, `%markerMenu_searchPlaceholder_character%` in
  `MARKER_MENU_STRING_KEYS`, a two-glyph indicator per the 2026-08-06 UX decision, and
  `min-w-0`/`truncate` responsiveness fixes). `main`'s version was taken wholesale.
- `lib/papi-dts/papi.d.ts` — unchanged. No `@papi/` module changed. No `papi-shared-types` augmentation.
- No localized-string keys added or removed against `origin/main`.
- Extension-internal only (module-level exports, not the extension's declared public API):
  - `character-marker-bar/` (new): `CharacterMarkerBar`, `CharacterMarkerBarOverlay`, `computeBarTop`,
    `resolveActiveLineRect`, `EDITOR_PARA_SELECTOR`, `ActiveLine`, `useEditorSelectionVersion`
  - `editor-dom.util.ts`: `clampTopToVisibleArea`, `measureBaselineOffset`, `BASELINE_PROBE_ATTRIBUTE`
  - `CharacterMarkerBarOverlayProps.bar` changed from required to **optional** (see Critical findings)

- [x] **Branch built on a stale merge base; `origin/main` (#2648) already contained a superseding
  revision of nearly every file touched.** `main` ships `character-marker-control/` as a directory
  with `menuDirection` **required**; the branch had a flat module with it optional. Git saw no shared
  history for those paths, so a merge would have kept both implementations *without a conflict
  marker*, and landing the branch's `marker-menu.component.tsx` would have removed two shipped public
  API members. _(fixed during review: merged `origin/main`; took its version of the control, coverage
  utils, state hook, power-mode hook, `MarkerMenu`, localized strings and `dist/`; deleted the
  branch's flat control files; re-targeted `character-marker-bar/` onto the directory and made
  `textDirection` required to satisfy the now-required `menuDirection`.)_

[Author response: Confirmed the branch predated #2648 and asked for the merge to be done first, with
a landable D2 as the outcome. Before resolving, every conflicting pair was diffed to check what
taking `main`'s side would lose — the verdict was that `main` is a strict superset on the shared
layer (no exported symbol, type member, behavior, test assertion or documented constraint lost), and
it additionally fixes two coverage bugs the branch had: the end-path-names-a-`MarkerObject`
undercount, and a caret inside a note reporting uncovered text rather than "no information".]

- [x] **The interface-mode ternary remounted the entire editor.** The web view chose between
  `ParagraphMarkerTooltipOverlay` and `CharacterMarkerBarOverlay` at the same tree position; React
  reconciles by element type, so any mode flip unmounted and remounted `editorTree`, losing Lexical's
  undo history and the scroll position. This was not only a settings-toggle case: `useSetting`
  returns its `'simple'` default before the stored value arrives, so a Power-mode user's editor
  remounted on **every open**. _(fixed during review: `CharacterMarkerBarOverlay.bar` is now optional
  and the same overlay is mounted in both modes, so only the slot content changes. Four tests added,
  including one asserting the editor's DOM node **identity** survives the slot filling in and
  emptying — a remount produces a new node even though the markup is identical.)_
- [x] **`_simple-mode.scss`'s two gates vs. the bar's one gate.** The reservation matched
  `.editor-container-simple .usfm.psc-gutter-markers` while the bar rendered on `!isPowerMode` alone;
  the comment documented this as a risk deferred to whoever makes Simple view options selectable.
  _(fixed during review — but NOT the way the analyzer proposed; see the author response below.)_
- [ ] ~~**`useIsPowerMode` duplicates the inline `isPowerMode` memo still in the web view.**~~
  _(`main` has the same duplication; the hook's only consumer there is `CharacterMarkerToolbar`.
  Adopting the hook in the web view means handling its `boolean | undefined` third state at ~6 other
  call sites — BCV controls, view options, the `editor-container-simple` class — which changes
  Power-mode chrome behavior on load and is beyond what D2 needs. Deferred deliberately; see
  Follow-ups.)_
- [ ] ~~**`localized-strings.test.ts` hand-copied `CHARACTER_MARKER_CONTROL_STRING_KEYS`**, so a new
  key silently escaped the en/es parity guard.~~ _(Resolved by the merge: `main` extracts the key list
  into a DOM-free `character-marker-control.const.ts` that both the component and the node-environment
  test import — exactly the fix the analyzer proposed.)_
- [ ] **Icon-only trigger's tooltip carries no indication of what the control is.** In the shipped
  gutter placement the trigger is icon-only, so the tooltip is the sighted user's only readout — but
  `tooltipText = isDisabled ? disabledTooltip : label`, so it reads bare `(none)`, `(mixed)` or
  `bd - Bold`. The `Tooltips` guideline requires a tooltip on an icon button to "add meaning an icon
  alone cannot convey". The accessible name already composes the full `Character marker: bd - Bold`
  via `ARIA_LABEL_FORMAT_KEY`; using that same string for the visual tooltip when `isLabelHidden`
  would fix it. **Left open deliberately: this is `main`'s control, shared with the toolbar placement,
  not this branch's code.**
- [x] **No story for `CharacterMarkerBar`** — the wired composition (state hook + toolbar + control +
  RTL/menu-align decisions), and the one variant a reviewer would most want to see rendered. Both its
  siblings had stories. _(fixed during review: added `character-marker-bar.component.stories.tsx` at
  the real 64px gutter width, with no-marker, applied, mixed, sync-blocked and RTL cases.)_

[Author response: Author approved fixing the remount, the gate mismatch and the missing story, and
chose to flag the tooltip for the reviewer rather than modify `main`'s control mid-branch.

**On the gate mismatch — the analyzer's proposed fix was wrong and was not applied.** It suggested
broadening the CSS to `.editor-container-simple .usfm`. Investigation showed `_simple-mode.scss` is
compiled into every web view in this extension, and both `resource-text-panel.web-view.tsx:615` and
`model-text-panel.component.tsx:458` apply `editor-container-simple` **unconditionally** to an editor
that sets no `hasGutterParaMarkers` — so the broader selector would have put 64px of dead padding on
the Bible Texts / Commentaries panel and the model text panel. It would also have dropped the RTL rule
from specificity (0,4,0) to (0,3,0), tying `_usj-nodes.scss`'s mirror rule and losing on source order,
silently killing the RTL reservation. The scope-guard test would not have caught either. The second
gate is **load-bearing, not incidental**. Fixed from the other side instead: the bar is now gated on
`!isPowerMode && viewOptions.hasGutterParaMarkers` — the same two conditions the CSS selects on — and
the stylesheet comment was corrected to say so.]

### Minor — Consider

- [x] **Changelog-style comments** narrating superseded values (`not the gap-2/8px this comment
  previously assumed`, `The 5em (80px) this replaces…`, `the earlier 2.5em estimate undercounted`),
  repeated in `simple-mode.scss.test.ts`. _(fixed during review: part of a branch-wide forward-facing
  comment audit — see Interview Notes.)_
- [x] **`clampTopToVisibleArea` had no direct unit test** — only indirect coverage through
  `computePosition` and `computeBarTop`. _(fixed during review: five value tests, pinning the clamp
  ORDER and the 1px anchor height, which neither caller's suite pins on its own since both assert
  through their own additional offsets. One test uses an asymmetric anchor/scroll-container fixture so
  passing the wrong element — the documented failure mode — is caught.)_
- [x] **Internal spec-artifact references in test names and comments** (`U3's worked example`,
  `the spec's §4.1`) — meaningless to a reader who never saw the plan. _(fixed during review; these
  had been removed in the comment audit and came back when the merge took `main`'s files.)_
- [ ] ~~Search-placeholder key was extension-local rather than a `%markerMenu_searchPlaceholder_*%`
  sibling~~ _(resolved by the merge — now `%markerMenu_searchPlaceholder_character%` in
  `platform-bible-react`.)_
- [ ] ~~`label` concatenated with a hardcoded ASCII hyphen rather than a localized format string~~
  _(resolved by the merge — `main` added `LABEL_FORMAT_KEY` and `ARIA_LABEL_FORMAT_KEY`.)_
- [ ] ~~Key naming inconsistency (`characterMarkerMenu_` vs `characterMarkerControl_`)~~ _(resolved
  by the merge along with the key itself.)_
- [ ] ~~Redundant Power-mode gate inside `CharacterMarkerToolbar`~~ _(not redundant on `main`: the
  gate is `isPowerMode !== false`, which also withholds render while the setting is unresolved,
  preventing a frame of Simple-only UI in a Power session.)_
- [ ] ~~Large unrelated churn in committed `dist/` build output~~ _(resolved: `dist/` was taken from
  `main` verbatim, and a final commit dropped stray `dist/experimental.*` churn so the build output
  matches its source — no `lib` source differs from `main`.)_
- [ ] **`CharacterMarkerControlLocalizedStrings` is `{[key: string]: string | undefined}`**, which
  `LanguageStrings` (keyed by `LocalizeKey`) is not assignable to — so the bar performs a no-op
  `Object.fromEntries` clone plus a `useMemo` to keep it off the caret-move render path. Typing the
  prop as `LanguageStrings` would delete both. **`main`'s file** — deferred, and now documented as a
  deliberate deferral at the workaround site.
- [ ] **No named `MarkerSelectionState` export from `platform-bible-react`** — the
  `'all' | 'partial' | 'none'` union is re-declared in four places. **`main`'s file** — deferred.
- [ ] **File organization**: `character-marker-coverage.utils.ts` and `use-character-marker-state.hook.ts`
  sit at `src/` root while the control and the bar are in feature directories. Left alone since `main`
  just established that layout.
- [ ] **`(mixed)` / `(none)` are lowercase parentheticals** — a PT9 idiom that reads oddly standing
  alone in a tooltip and deviates from the sentence-case capitalization guideline. Suggested
  `None`/`Mixed` or `No marker`/`Multiple markers`. **UX call.**
- [ ] **64px is ~21% of the text column at the 300px minimum tab width** from the responsiveness
  guideline. The reservation is unconditional in Simple mode (deliberately, to avoid reflow). **Worth
  UX confirming the narrow case.**
- [ ] **Popover uses stock Radix dismissal only** — `dismissal-patterns.mdx` asks each new surface to
  add the cross-iframe focus supplement, since pointer-based outside detection fails silently when the
  user clicks into a sibling WebView. Verified no surface in the repo implements it yet, so this is a
  pre-existing repo-wide gap rather than a regression; noted only because this is a new surface.
- [ ] **The plan and PRD this work is measured against are untracked**, so a reviewer cannot see N1/H1.
  Quote them in the PR description.

[Author response: Author asked for the two cheap, branch-owned items (direct `clampTopToVisibleArea`
coverage and the spec-artifact de-references) to be fixed, and for the rest to be recorded for
follow-up. Items in `main`'s files were deliberately not widened into this branch.]

#### Shared Regions Modified

None. No `#region shared with` markers appear in any changed file.

#### Extension Config Changes

None. No `package.json`, `tsconfig.json`, `webpack.config.ts` or `.eslintrc*` under `extensions/` was
modified — all extension edits are under `platform-scripture-editor/src/`.

## Positive Observations

- **The hidden-view rule is genuinely handled, not merely dismissed.** `CharacterMarkerBarOverlay`
  uses `useViewVisibility`, defers layout-dependent work behind `isVisibleRef`, collapses repeats into
  a single `pendingCatchUpRef`, and consumes it in a `useLayoutEffect` before paint. `recompute` keeps
  a deliberately stable identity so the catch-up is the *only* repair path and therefore observable —
  and three tests cover it. This is the rule's own reference shape.
- `measureBaselineOffset` returning `undefined` rather than `0` under `display: none` is exactly the
  right distinction, and the reason is documented at the function.
- `clampTopToVisibleArea` is a real de-duplication: the vertical-clamp math was lifted out of
  `paragraph-marker-tooltip.utils.ts` and shared rather than copy-pasted, with `computePosition`'s
  behavior preserved verbatim.
- **The reservation and the bar's width read one custom property** (`--psc-character-marker-bar-width`,
  declared once on `.editor-container-simple`), with no fallback literal — so they cannot drift.
- `simple-mode.scss.test.ts` scans the stylesheet source to pin both the Simple-scope prefix on every
  selector and the absolute-unit reservation — a class of regression no runtime test could catch.
- The popover's list-height cap is applied as a descendant override on the consumer's own
  `PopoverContent` via the stable `data-[slot=command-list]` attribute, with a comment explaining that
  `MarkerMenu` must stay untouched because Power-mode consumers share it — the right way to avoid a
  shadcn customization. No `shadcn-ui/` file was touched.
- RTL is designed for rather than retrofitted: `insetInlineEnd`, an explicit `menuDirection` threaded
  to the portaled popover, and a specificity-aware RTL padding rule whose (0,3,0)→(0,4,0) reasoning
  checks out against the source.
- Disabled state greys out rather than unmounting as the caret crosses a chapter number, and the
  explanatory tooltip stays reachable for keyboard and screen-reader users — a case usually dropped.
- The sync-blocked tooltip reuses the shipped `%…syncEditBlocked_banner%` string rather than
  introducing a second phrasing.
- No new keyboard handlers were introduced (the selection hook listens to `selectionchange`), so the
  keyboard-shortcuts catalog correctly needs no update.

## Interview Notes

**Stated purpose**: implement issue D2 from the character-level markers implementation plan — mount
the character-marker control in the editor gutter — serving the PRD "Simple — Saroj easily works with
character level markers".

**Author-directed work during the review**, beyond responding to findings:

1. **Reverted ADR-0007** from `.context/standards/Architecture-Decisions.md`. The author's reasoning:
   the gutter placement may be replaced by an overlay modal, so the decision is not settled enough to
   belong in an append-only decision log. The now-dangling `see ADR-0007` pointer in
   `_simple-mode.scss` was removed with it; the arithmetic rationale it introduced stands alone.
2. **Commissioned a branch-wide forward-facing comment audit** against an explicit rule (strip the PR
   context — does the comment still help?). 20 comments were cut or rewritten across four parallel
   passes. Categories found: ticket/spec/story IDs for work done in this PR (`PT-XXX-D2`, `U1:`,
   `U3's worked example`, `the spec's §4.1`); superseded-value narration; pre-PR consumer behavior
   ("renders exactly as it did"); comments arguing with a reviewer ("this assertion is load-bearing,
   not incidental"); and prototype references. **Two were also factually stale**, not just
   backward-facing: a web-view comment said "one extra key" for a six-key list, and an overlay test
   called its own new tests "pre-existing". Forward-facing notes were deliberately kept — the
   `_simple-mode.scss` gate risk, the RTL specificity math, the hidden-view catch-up notes the repo
   rule requires at each sync site, the `menuDirection`/Radix-portal rationale, and the test
   falsifiability notes.

- **Merge before fixing.** The author chose to merge `origin/main` first rather than fix findings
  against the stale base, since six of the fixes touched files `main` had already replaced. Correct
  call — fixing the flat control file would have fixed a file that does not survive the merge.
- **Landable D2, despite the possible pivot.** The author explicitly asked for a landable D2 rather
  than parking the branch as a spike, while noting they hope to use an overlay modal instead and will
  "come back to it" if not.
- **Scope discipline on `main`'s files.** The author consistently chose to flag findings in `main`'s
  code (tooltip composition, `LocalizedStrings` typing, `MarkerSelectionState` export) rather than
  widen this branch into them.

**Two analyzer findings were wrong, and are worth noting because they were caught by verification
rather than by review judgment:**

- The proposed `_simple-mode.scss` broadening would have caused a visible regression in two other
  web views (detailed above). Applied as suggested, it would have shipped.
- A claimed flash of the bar in Power mode during setting resolution does not occur —
  `CharacterMarkerToolbar` already gates on `isPowerMode !== false`. No change was made.

**Author understanding**: the author directed the technical decisions throughout — choosing the merge
ordering, the resolution strategy, which findings to fix versus flag, and rejecting scope creep into
`main`'s files. No areas were deferred to AI or left unexplained. The author declined to raise
additional context at wrap-up.

## In-Review Quality Check

Run after each of the four review commits, not only at the end:

- `npm run typecheck` — **clean.** One error surfaced immediately after the merge (the overlay story's
  harness missing the now-required `menuDirection`) and was fixed.
- `npm run lint` — **clean, 0 errors.** 4 warnings remain, all pre-existing and untouched by this
  branch (`import/no-duplicates` ×2 in `resource-reference.utils.ts`, `no-console` ×2).
- `npx prettier --write` on changed files only — applied; no repo-wide format run, so no unrelated
  files were reformatted.
- **Tests** — 741 `platform-scripture-editor` tests and 1023 `platform-bible-react` tests passing.
  Five bar tests failed immediately after the merge (the localized `ariaLabel`/`label` format patterns
  and the relocated search-placeholder key) and were fixed.
- **Falsifiability check** — the new "reads no geometry when there is nothing to position" test was
  verified to fail when the early return it covers is removed, rather than assumed to be meaningful.

## Suggested Review Focus

- [ ] **The merge resolution is the largest single thing to check.** `origin/main` (#2648) shipped a
  later revision of the same work and its version was taken wholesale on the shared layer. The
  per-file salvage analysis concluded nothing was lost, but that conclusion is worth a second pair of
  eyes on `character-marker-coverage.utils.ts` and `use-character-marker-state.hook.ts` in particular,
  which differed by 59 and 68 lines.
- [ ] **`_simple-mode.scss`'s double gate.** The analyzer proposed broadening it and that would have
  regressed two other web views. Confirm the corrected reasoning holds, and that gating the bar on
  `viewOptions.hasGutterParaMarkers` is the right way to keep the CSS and React conditions in step.
- [ ] **The overlay's optional `bar` slot.** Confirm the remount fix is right in principle — that
  keeping one element type mounted in both modes is preferable to the previous absent-from-the-tree
  gating, given the reserved space is governed by CSS independently.
- [ ] **Icon-only tooltip content** (open finding). Should the visual tooltip carry
  `Character marker: bd - Bold` rather than bare `bd - Bold`? The fix lands in `main`'s control.
- [ ] **UX, not code**: the `(mixed)`/`(none)` lowercase parentheticals, and whether 64px (~21% of the
  text column at the 300px minimum tab width) is acceptable in narrow tabs.
- [ ] **Direction of travel.** The author may replace this gutter placement with an overlay modal. If
  that is likely, decide whether to land D2 as-is or hold it — it affects how closely the gutter
  geometry is worth scrutinizing now.
- [ ] **Follow-ups deliberately deferred**, all in `main`'s files: adopt `useIsPowerMode` in the web
  view and delete the duplicated inline memo (needs the `boolean | undefined` third state handled at
  ~6 call sites); type `CharacterMarkerControlLocalizedStrings` as `LanguageStrings` to delete a
  no-op clone and its memo; export a named `MarkerSelectionState` from `platform-bible-react`.

AI-assisted — Claude Opus 5


katherinejensen00 (changes requested):
Thanks for your hard work on this, Jolie!

```markdown
Claude summary:

**Please look at before merge:**
1. The empty-chapter comment credits `useViewVisibility` for the `tw:hidden` case, but that hook watches `document.body`/the whole tab, not the hidden subtree — so a Simple-mode editor mounting straight onto a blank chapter positions against zero-geometry, latches `hasPositionedRef`, and the ResizeObserver catch-up then short-circuits at `if (hasPositionedRef.current) return`, leaving the bar at `top: 0` until the first click. Details inline.
2. Nothing re-runs `recompute` when the `bar` slot fills on a live Power→Simple switch; it currently self-heals only because the mode switch incidentally reflows the editor and trips the ResizeObserver. A one-line effect keyed on `bar !== undefined` would make it deterministic. Details inline.

**Smaller items (inline):** a loading-window mismatch between the two `isPowerMode` derivations, the multi-line-selection anchoring being undocumented, the new search-placeholder string dropping the trailing period its siblings carry (en + es), a few per-frame layout-read micro-inefficiencies, the `localizedStrings` clone-to-launder-a-type, and some comment de-duplication (with a caveat: keep the per-site coupling comments, since a repo rule wants them checkable from the diff).

One housekeeping check: my local `tsc` flags `hasGutterParaMarkers does not exist on ViewOptions`, but that's a stale local `node_modules` artifact (the field is established and documented) — can you just confirm CI typecheck is green?

Nice work overall.
```

> @katherinejensen00 reviewed 21 files and all commit messages, and made 22 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2660#-:-Ozh71vi-WImLdS09JNP:b-wno26s)* status: all files reviewed, 21 unresolved discussions (waiting on jolierabideau).

___
*[`assets/localization/en.json` line 314 at r1](https://reviewable.io/reviews/paranext/paranext-core/2660#-Ozh9sEB4lztYAnFU-ho:-Ozh9sEB4lztYAnFU-hp:b-48lnab) ([raw file](https://github.com/paranext/paranext-core/blob/c26cce158caa03d8a5a9d384ee758dd92450278f/assets/localization/en.json#L314)):*
> ```JSON
>   "%markerMenu_noResults%": "No results found.",
>   "%markerMenu_searchPlaceholder%": "Type a style or search.",
>   "%markerMenu_searchPlaceholder_character%": "Search character markers",
> ```

NIT The three sibling placeholders that share this same `MarkerMenu` search field all end with a period — `"Type a style or search."`, `"Search to insert a style."`, `"Search to change paragraph style."` — and the English phrasing follows a `"Search to <verb> a <thing>."` pattern. This one is `"Search character markers"` (no period, different pattern), and the es.json entry (`"Buscar marcadores de carácter"`, line 430) drops the period too. Worth matching the sibling style in both languages for consistency when switching menus.
___
*[`extensions/src/platform-scripture-editor/src/_simple-mode.scss` line 53 at r1](https://reviewable.io/reviews/paranext/paranext-core/2660#-OzhBgWBE0t0lFxZQ5lQ:-OzhBgWBE0t0lFxZQ5lR:b-koc5hr) ([raw file](https://github.com/paranext/paranext-core/blob/c26cce158caa03d8a5a9d384ee758dd92450278f/extensions/src/platform-scripture-editor/src/_simple-mode.scss#L53)):*
> ```SCSS
>   padding-block: 0 !important;
> }
> ```

NIT These comments are genuinely good, but the reserve-vs-float story, the 64px chrome arithmetic, the px-vs-em argument, and the webfont caveat are each retold across `_simple-mode.scss`, the web-view JSX, `character-marker-bar.component.tsx`, and the overlay — ~30–40 lines of restatement. Suggest making this SCSS block the canonical home for the *arithmetic* and pointing the others at it.

One caveat so this doesn't over-correct: keep the *per-site coupling* comments (the gating-parity note at the web-view `bar` gate, especially). Those are what make the CSS↔JS invariant checkable from a diff that touches only that file, which `.claude/rules/cross-view-sync-hidden-views.md` asks for. Trim the repeated arithmetic/webfont tangents, not the coupling invariants.
___
*[`extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx` line 409 at r1](https://reviewable.io/reviews/paranext/paranext-core/2660#-Ozh7dWV6uso4VORyTZW:-Ozh7dWV6uso4VORyTZX:bgvhwvk) ([raw file](https://github.com/paranext/paranext-core/blob/c26cce158caa03d8a5a9d384ee758dd92450278f/extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx#L409)):*
> ```XML
>   const [interfaceModePossiblyError] = useSetting('platform.interfaceMode', 'simple');
> 
>   const isPowerMode = useMemo(() => {
> ```

NIT This local `isPowerMode` has no `isLoading` guard, so during the `platform.interfaceMode` load window it reads the `'simple'` default → `false` and passes a real `<CharacterMarkerBar>` at the gate below (line 2002). Inside, `CharacterMarkerToolbar` uses the three-state `useIsPowerMode()`, which returns `undefined` while loading and renders nothing. Net: an empty, positioned 64px bar container paints for the loading frames before the toolbar fills it (or before it unmounts if the resolved mode is Power).

It self-heals, and it's consistent with the web-view's existing load-time Simple-layout flash — but it partly defeats the flash-prevention the three-state hook exists for. Gating the `bar` with the same `useIsPowerMode()` three-state value would close it.
___
*[`extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx` line 1975 at r1](https://reviewable.io/reviews/paranext/paranext-core/2660#-Ozh7JSX3QLhNnYBOpVV:-Ozh7JSX3QLhNnYBOpVW:b-hx2uwn) ([raw file](https://github.com/paranext/paranext-core/blob/c26cce158caa03d8a5a9d384ee758dd92450278f/extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx#L1975)):*
> ```XML
>           />
>         )}
>         {/* The empty-chapter view HIDES this subtree rather than replacing it, for the same reason
> ```

This comment says the overlay's `useViewVisibility` deferral "already covers being under `display:none`" — but `useViewVisibility` observes `document.body`/the iframe viewport (whole-tab visibility), not this inner `tw:hidden` subtree. So while the chapter is blank but the tab is active, `isViewVisible` is `true` and the overlay runs its *visible* recompute path against zero rects, anchors to the first paragraph's zero rect, and sets `hasPositionedRef = true`.

When the chapter fills in and this subtree un-hides, the ResizeObserver's `recompute` finds no caret selection and hits `if (hasPositionedRef.current) return` (character-marker-bar-overlay.component.tsx:179) *before* the first-paragraph re-anchor — so the bar stays at `top: 0` until the first click. This is the geometry-under-`display:none` case `.claude/rules/cross-view-sync-hidden-views.md` warns about.

Two things: (1) fix the comment to name the mechanism that actually recovers (the ResizeObserver, not the visibility deferral), and (2) consider resetting `hasPositionedRef` when the subtree un-hides so the first-paragraph anchor can run without a caret.
___
*[`extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx` line 2002 at r1](https://reviewable.io/reviews/paranext/paranext-core/2660#-OzhB3YY4EyvvqimGVby:-OzhB3YY4EyvvqimGVbz:b-5q3ymr) ([raw file](https://github.com/paranext/paranext-core/blob/c26cce158caa03d8a5a9d384ee758dd92450278f/extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx#L2002)):*
> ```XML
>           <CharacterMarkerBarOverlay
>             bar={
>               isPowerMode || !viewOptions.hasGutterParaMarkers ? undefined : (
> ```

NIT The bar and its reserved gutter space are kept in lockstep across three sites (this gate, the `.editor-container-simple .usfm.psc-gutter-markers` selector in `_simple-mode.scss`, and the toolbar's own gate) by comments alone. The defense-in-depth is good, but per the repo's "documented intent doesn't survive a refactor; a test does," a small test asserting the bar renders under exactly the `!isPowerMode && hasGutterParaMarkers` the SCSS matches would protect the highest-coupling point in the PR.
___
*[`extensions/src/platform-scripture-editor/src/character-marker-bar/character-marker-bar-overlay.component.tsx` line 27 at r1](https://reviewable.io/reviews/paranext/paranext-core/2660#-OzhAfxi0IiIr2znYVG7:-OzhAfxi0IiIr2znYVG8:b442jkm) ([raw file](https://github.com/paranext/paranext-core/blob/c26cce158caa03d8a5a9d384ee758dd92450278f/extensions/src/platform-scripture-editor/src/character-marker-bar/character-marker-bar-overlay.component.tsx#L27)):*
> ```XML
>  */
> function measureTriggerIconCenter(barContainer: HTMLElement): number | undefined {
>   const icon = barContainer.querySelector('svg');
> ```

NIT Not a live bug — today the trigger's first SVG is the `Type` icon, and both icons share the same vertical center in the `items-center` flex row, so which SVG is returned can't change the result. But the overlay's props doc says the shell "knows nothing about character markers … could anchor anything," and this hard-codes an `svg` alignment probe. A future slot occupant leading with a different element would silently misalign. A `data-` attribute or a ref the slot declares would keep the abstraction honest.
___
*[`extensions/src/platform-scripture-editor/src/character-marker-bar/character-marker-bar-overlay.component.tsx` line 103 at r1](https://reviewable.io/reviews/paranext/paranext-core/2660#-OzhC5VK1sE5zgVXDLRO:-OzhC5VK1sE5zgVXDLRP:b-n0jq6d) ([raw file](https://github.com/paranext/paranext-core/blob/c26cce158caa03d8a5a9d384ee758dd92450278f/extensions/src/platform-scripture-editor/src/character-marker-bar/character-marker-bar-overlay.component.tsx#L103)):*
> ```XML
>   // subset would silently reuse an offset measured under different metrics.
>   //
>   // This key does NOT catch a webfont finishing an asynchronous load: `getComputedStyle().fontFamily`
> ```

NIT The "computed-style can't see a webfont face swap" explanation appears here, at the `loadingdone` effect, and in the recompute body. Keep the full version only at the `loadingdone` effect (its natural home) and reduce this one to a one-line pointer.
___
*[`extensions/src/platform-scripture-editor/src/character-marker-bar/character-marker-bar-overlay.component.tsx` line 164 at r1](https://reviewable.io/reviews/paranext/paranext-core/2660#-OzhBZesCtoPoem4OwpG:-OzhBZesCtoPoem4OwpH:bxvvjxd) ([raw file](https://github.com/paranext/paranext-core/blob/c26cce158caa03d8a5a9d384ee758dd92450278f/extensions/src/platform-scripture-editor/src/character-marker-bar/character-marker-bar-overlay.component.tsx#L164)):*
> ```XML
> 
>     const editorRoot = positionAnchor.querySelector<HTMLElement>(EDITOR_ROOT_SELECTOR);
>     if (!editorRoot) return;
> ```

NIT Unlike the hidden-view case, if `bar` is present but `.editor-input` isn't in the DOM yet, `recompute` returns with no self-correction — it relies on an incidental later event. Mitigated at the real mount site (the bar only renders once USJ has loaded), so this is a latent contract gap rather than a live bug; a comment noting the mount-site guarantee would make the assumption explicit.
___
*[`extensions/src/platform-scripture-editor/src/character-marker-bar/character-marker-bar-overlay.component.tsx` line 204 at r1](https://reviewable.io/reviews/paranext/paranext-core/2660#-OzhA74l7aY7thumUi_p:-OzhA74l7aY7thumUi_q:bh7jbm7) ([raw file](https://github.com/paranext/paranext-core/blob/c26cce158caa03d8a5a9d384ee758dd92450278f/extensions/src/platform-scripture-editor/src/character-marker-bar/character-marker-bar-overlay.component.tsx#L204)):*
> ```XML
>     // default treats it as 0, i.e. the bar stays top-aligned, and the next recompute tries again.
>     const barContainer = barContainerRef.current;
>     const iconCenter = barContainer ? measureTriggerIconCenter(barContainer) : undefined;
> ```

NIT `iconCenter` only feeds the `if (metricsChanged)` block (line 222), but `measureTriggerIconCenter` (2 `getBoundingClientRect` reads) runs on every recompute frame, even when the font-metrics cache is warm and the result is discarded. Moving the icon measurement inside the `metricsChanged` guard drops steady-state reads from ~6 to ~4 per scroll frame.
___
*[`extensions/src/platform-scripture-editor/src/character-marker-bar/character-marker-bar-overlay.component.tsx` line 226 at r1](https://reviewable.io/reviews/paranext/paranext-core/2660#-OzhAC168Cj_kSG-NeQE:-OzhAC168Cj_kSG-NeQF:bhqmlhq) ([raw file](https://github.com/paranext/paranext-core/blob/c26cce158caa03d8a5a9d384ee758dd92450278f/extensions/src/platform-scripture-editor/src/character-marker-bar/character-marker-bar-overlay.component.tsx#L226)):*
> ```XML
>           // line-box strut matches the real paragraph's — no filler text needed, the strut alone is
>           // what gets measured.
>           measuringElement.style.fontFamily = fontFamily;
> ```

NIT These six style writes are immediately followed by `measureBaselineOffset`'s `getBoundingClientRect`, forcing a synchronous layout flush. Harmless once, but it re-triggers whenever the caret crosses a metrics boundary (e.g. body text ↔ an `\mt1` heading), so holding an arrow key across that boundary thrashes layout each frame. Minor — noting in case it's cheap to batch.
___
*[`extensions/src/platform-scripture-editor/src/character-marker-bar/character-marker-bar-overlay.component.tsx` line 256 at r1](https://reviewable.io/reviews/paranext/paranext-core/2660#-Ozh7U9bCc99Z6z1oODd:-Ozh7U9bCc99Z6z1oODe:b-9sweg5) ([raw file](https://github.com/paranext/paranext-core/blob/c26cce158caa03d8a5a9d384ee758dd92450278f/extensions/src/platform-scripture-editor/src/character-marker-bar/character-marker-bar-overlay.component.tsx#L256)):*
> ```XML
>   const scheduleRecompute = useCallback(() => {
>     cancelAnimationFrame(rafIdRef.current);
>     rafIdRef.current = requestAnimationFrame(recompute);
> ```

NIT When the overlay first mounts in Power mode, `bar` is `undefined`, so `recompute` early-returns at line 145 and never sets `top`. On a live Power→Simple switch the `bar` slot fills in, but no effect re-runs `recompute`: this mount effect is keyed on `[recompute, scheduleRecompute]` (both stable), and the selection/visibility/fonts effects don't fire on a slot change. The absolutely-positioned bar container also doesn't change `positionAnchor`'s size, so the ResizeObserver isn't guaranteed to fire either.

In practice it self-heals because the switch adds `editor-container-simple` to the root, reflows `.usfm`, and usually trips the ResizeObserver — but that's incidental. A small effect keyed on `bar !== undefined` that calls `scheduleRecompute()` would make the first paint deterministic. (`useIsPowerMode`'s own doc says the live toggle is supported, so this path is real.)
___
*[`extensions/src/platform-scripture-editor/src/character-marker-bar/character-marker-bar-overlay.component.tsx` line 267 at r1](https://reviewable.io/reviews/paranext/paranext-core/2660#-OzhAnpx-naBr5gIVKVB:-OzhAnpx-naBr5gIVKVC:bzb8tg1) ([raw file](https://github.com/paranext/paranext-core/blob/c26cce158caa03d8a5a9d384ee758dd92450278f/extensions/src/platform-scripture-editor/src/character-marker-bar/character-marker-bar-overlay.component.tsx#L267)):*
> ```XML
>     // scrollHeight and clientHeight are both 0 — so "actually overflowing right now" would be the
>     // wrong criterion. getComputedStyle still works under display:none, so discovery stays valid.
>     scrollContainerRef.current =
> ```

NIT The math is nicely shared (`clampTopToVisibleArea`, `findScrollContainer`, `measureBaselineOffset`), but ~30 lines of scroll-container discovery + listener + ResizeObserver + rAF-cleanup wiring are copy-pasted from `ParagraphMarkerTooltipOverlay`. Defensible at two consumers — but the toolbar comment anticipates footnote/comment controls, and a third one would be the point to extract a `useGutterAnchor` hook so a lifecycle fix doesn't have to be mirrored by hand.
___
*[`extensions/src/platform-scripture-editor/src/character-marker-bar/character-marker-bar-overlay.component.tsx` line 267 at r1](https://reviewable.io/reviews/paranext/paranext-core/2660#-OzhBTyx33Bf3XC7VSTU:-OzhBTyx33Bf3XC7VSTV:b6fibsu) ([raw file](https://github.com/paranext/paranext-core/blob/c26cce158caa03d8a5a9d384ee758dd92450278f/extensions/src/platform-scripture-editor/src/character-marker-bar/character-marker-bar-overlay.component.tsx#L267)):*
> ```XML
>     // scrollHeight and clientHeight are both 0 — so "actually overflowing right now" would be the
>     // wrong criterion. getComputedStyle still works under display:none, so discovery stays valid.
>     scrollContainerRef.current =
> ```

NIT (Same line as #13.) The scroll container is resolved once on mount and never re-resolved when the editor subtree is re-parented by the footnotes reverse-portal. Safe today only because the footnotes `ResizablePanel` uses `overflow: hidden`, so discovery resolves to the same outer container either way — but it's an assumption worth a note, since it's inherited from the sibling overlays rather than re-derived.
___
*[`extensions/src/platform-scripture-editor/src/character-marker-bar/character-marker-bar-overlay.component.tsx` line 280 at r1](https://reviewable.io/reviews/paranext/paranext-core/2660#-OzhA0QIEaIXlwHUP-tt:-OzhA0QIEaIXlwHUP-tu:b-mip865) ([raw file](https://github.com/paranext/paranext-core/blob/c26cce158caa03d8a5a9d384ee758dd92450278f/extensions/src/platform-scripture-editor/src/character-marker-bar/character-marker-bar-overlay.component.tsx#L280)):*
> ```XML
>     // tooltip needs no equivalent because it is transient. Without this, dragging the dock
>     // splitter leaves the bar at a stale top.
>     const resizeObserver = new ResizeObserver(() => recompute());
> ```

NIT Scroll and selection route through `scheduleRecompute` (rAF-coalesced), but the ResizeObserver calls `recompute` directly. A dock-splitter drag then runs the full ~6-layout-read recompute per resize tick with no per-frame cap, and calling `setTop` synchronously inside the observer callback risks the benign-but-noisy "ResizeObserver loop" console error (which can surface in E2E). Routing this through `scheduleRecompute` fixes both.
___
*[`extensions/src/platform-scripture-editor/src/character-marker-bar/character-marker-bar-overlay.component.tsx` line 308 at r1](https://reviewable.io/reviews/paranext/paranext-core/2660#-OzhAISw8NDbS6enmh5i:-OzhAISw8NDbS6enmh5j:bxu1rki) ([raw file](https://github.com/paranext/paranext-core/blob/c26cce158caa03d8a5a9d384ee758dd92450278f/extensions/src/platform-scripture-editor/src/character-marker-bar/character-marker-bar-overlay.component.tsx#L308)):*
> ```XML
> 
>     const handleFontsLoaded = () => {
>       fontMetricsRef.current = undefined;
> ```

NIT `loadingdone` fires for every completed font batch in the document, including fonts loaded by unrelated panels. Each one clears this editor's baseline cache and forces a full re-measure on the next recompute, which defeats the "measure once" intent in a busy multi-panel layout. Correct, just broader than needed — no action required unless you want to narrow it.
___
*[`extensions/src/platform-scripture-editor/src/character-marker-bar/character-marker-bar-overlay.component.tsx` line 373 at r1](https://reviewable.io/reviews/paranext/paranext-core/2660#-OzhBv8uCOTd7-PoyBc8:-OzhBv8uCOTd7-PoyBc9:bedu6ek) ([raw file](https://github.com/paranext/paranext-core/blob/c26cce158caa03d8a5a9d384ee758dd92450278f/extensions/src/platform-scripture-editor/src/character-marker-bar/character-marker-bar-overlay.component.tsx#L373)):*
> ```XML
>             ref={barContainerRef}
>             data-testid="character-marker-bar-container"
>             className="tw:absolute tw:pointer-events-none"
> ```

NIT This can collapse to one line without losing the point, e.g. `// '0px' string, not 0: a bare numeric 0 renders unitless.` The react-dom internals reference is more than the reader needs here.
___
*[`extensions/src/platform-scripture-editor/src/character-marker-bar/character-marker-bar-overlay.component.tsx` line 391 at r1](https://reviewable.io/reviews/paranext/paranext-core/2660#-OzhBOzW-iX7Iik54exM:-OzhBOzW-iX7Iik54exN:b-dy981k) ([raw file](https://github.com/paranext/paranext-core/blob/c26cce158caa03d8a5a9d384ee758dd92450278f/extensions/src/platform-scripture-editor/src/character-marker-bar/character-marker-bar-overlay.component.tsx#L391)):*
> ```XML
>             style={{
>               top,
>               insetInlineEnd: '0px',
> ```

NIT `insetInlineEnd: 0` resolves against `positionAnchor` (several wrappers up), while the reserved gutter is `padding-inline-end` on `.usfm`. Their edges coincide today only because every wrapper in between is full-width with no inline padding and `.editor-container` sets `max-width: none`. A future "reading width" / max-width on the column would push the bar out of the reserved gutter. A one-line guard comment here (or an assertion) would flag the dependency.
___
*[`extensions/src/platform-scripture-editor/src/character-marker-bar/character-marker-bar-overlay.component.test.tsx` line 67 at r1](https://reviewable.io/reviews/paranext/paranext-core/2660#-OzhAOZhE2NZSI44Gw1a:-OzhAOZhE2NZSI44Gw1b:boi18cv) ([raw file](https://github.com/paranext/paranext-core/blob/c26cce158caa03d8a5a9d384ee758dd92450278f/extensions/src/platform-scripture-editor/src/character-marker-bar/character-marker-bar-overlay.component.test.tsx#L67)):*
> ```XML
>   // vi.spyOn rather than direct assignment so afterEach's vi.restoreAllMocks() actually restores
>   // jsdom's own implementation — a bare assignment would leak the stub for the rest of the file.
>   vi.spyOn(Element.prototype, 'getBoundingClientRect').mockImplementation(function stub(
> ```

NIT This subscription (for the re-render side effect) plus the overlay's own `useEditorSelectionVersion()` means two `document` `selectionchange` listeners and two renders per caret move. They're legitimately different consumers, and the cost is small — but since the overlay already re-renders the slot on its own version bump, this one is arguably redundant. Fine to leave; worth consolidating to a shared parent subscription if a third consumer shows up.
___
*[`extensions/src/platform-scripture-editor/src/character-marker-bar/character-marker-bar.component.tsx` line 80 at r1](https://reviewable.io/reviews/paranext/paranext-core/2660#-OzhAZFMAqIJn8DjFs79:-OzhAZFMAqIJn8DjFs7A:bjg0223) ([raw file](https://github.com/paranext/paranext-core/blob/c26cce158caa03d8a5a9d384ee758dd92450278f/extensions/src/platform-scripture-editor/src/character-marker-bar/character-marker-bar.component.tsx#L80)):*
> ```XML
>   // the same object has to be handed over re-keyed. Memoized so it is not rebuilt on every caret
>   // move (`useEditorSelectionVersion` re-renders this component constantly).
>   const controlLocalizedStrings = useMemo(
> ```

NIT This clone exists only because the control declares `[key: string]` while `LanguageStrings` is `[k: LocalizeKey]` (both value types resolve to `string`) — as your own TSDoc notes, typing `CharacterMarkerControlLocalizedStrings` as `LanguageStrings` "would delete both the clone and the memo." Deferring it because the control is shared with the toolbar placement is reasonable; could you file a follow-up so the workaround doesn't get copy-pasted by the next placement?
___
*[`extensions/src/platform-scripture-editor/src/character-marker-bar/character-marker-bar.utils.ts` line 7 at r1](https://reviewable.io/reviews/paranext/paranext-core/2660#-OzhAtQTCLrSfwxb3KT4:-OzhAtQTCLrSfwxb3KT5:bepl6b4) ([raw file](https://github.com/paranext/paranext-core/blob/c26cce158caa03d8a5a9d384ee758dd92450278f/extensions/src/platform-scripture-editor/src/character-marker-bar/character-marker-bar.utils.ts#L7)):*
> ```XML
>  * `ParagraphMarkerTooltipOverlay` uses, so both overlays agree on what a paragraph is.
>  */
> export const EDITOR_PARA_SELECTOR = '.para[class*="usfm_"]';
> ```

NIT The doc comment says this "matches the selector `ParagraphMarkerTooltipOverlay` uses, so both overlays agree" — but that overlay still spells `.para[class*="usfm_"]` as a raw literal in three places. A single exported constant imported by both would enforce the agreement the comment currently only asserts.
___
*[`extensions/src/platform-scripture-editor/src/character-marker-bar/character-marker-bar.utils.ts` line 56 at r1](https://reviewable.io/reviews/paranext/paranext-core/2660#-Ozh9j-i6A_A3QT2wyWB:-Ozh9j-j1s8pJkOISVCJ:b-wwkyyp) ([raw file](https://github.com/paranext/paranext-core/blob/c26cce158caa03d8a5a9d384ee758dd92450278f/extensions/src/platform-scripture-editor/src/character-marker-bar/character-marker-bar.utils.ts#L56)):*
> ```XML
>   if (!para) return undefined;
> 
>   const caretRect = selection.getRangeAt(0).getBoundingClientRect();
> ```

NIT This uses `anchorNode` (line 48) plus the union rect of range 0. For a collapsed caret that's exact, but for a multi-line drag-selection the union rect's `top` is the *first* selected line, so the bar sits at the top of the selection while the user extends toward the far (focus) end — which contradicts the "line the caret is on" contract in the TSDoc (it only discusses the collapsed and degenerate cases). Either anchor to `focusNode`/the focus end, or document that a non-collapsed selection anchors to the selection top by design.

jolierabideau:
Worked through all 21. Both of the "look at before merge" items were real and are fixed with tests; the rest are below, including four I'm pushing back on.

## Fixed

**1. Empty-chapter comment + the `hasPositionedRef` latch** (`web-view.tsx:1975`) — correct on both counts, and the latch is reachable. `isChapterBlank` only means "no chapter or verse node", so a prose-only chapter still renders `.para` elements for the first-paragraph anchor to latch onto.

Fixed more generally than resetting the flag on un-hide: `recompute` now bails when the editor root reports a 0×0 rect, so *any* `display: none` ancestor is handled, not just this one call site. Nothing needs to schedule the repair — `display: none` collapses the observed anchor to a 0×0 box, so regaining layout is itself a resize and the ResizeObserver re-runs it. The comment now names the ResizeObserver as the recovery mechanism instead of crediting `useViewVisibility`, and explains that the hook reports whole-tab visibility.

New test: `does not latch a position measured while an ancestor gives the editor no layout`. Verified falsifiable — it fails when the guard is removed.

**2. No recompute when the `bar` slot fills** (`overlay:256`) — confirmed exactly as described; the mount effect's deps are both identity-stable and nothing else observes the slot. Added the one-line effect keyed on `bar !== undefined` (presence, not identity, so caret moves don't retrigger it). New test `positions the bar as soon as the slot fills, without waiting for an incidental resize`, also verified falsifiable.

**3. Search placeholder** (`en.json:314`) — right, and it's this branch's string rather than one inherited from `main`, so it was ours to fix. Now `"Search to change character style."` / `"Buscar para cambiar el estilo de carácter."`, matching the sibling pattern and punctuation in both languages.

**4. Gate-parity test** (`web-view:2002`) — agreed the highest-coupling point deserved more than comments. Added a third case to `simple-mode.scss.test.ts` asserting both halves: the compound `.editor-container-simple .usfm.psc-gutter-markers` selector on the CSS side and the `isPowerMode || !viewOptions.hasGutterParaMarkers` gate on the React side. Deliberately a source scan in the same spirit as the existing scope guard — the mount site is too deep in the web view to render in isolation, and a refactor that restates the gate differently *should* fail here rather than pass quietly.

**5. `measureTriggerIconCenter` on every frame** (`overlay:204`) — right, both reads were discarded with a warm cache. Moved inside the `metricsChanged` guard. Resolving the icon before the probe still holds, so a bar with no icon never touches the measuring element.

**6. ResizeObserver bypassing `scheduleRecompute`** (`overlay:280`) — fixed; it now routes through the rAF coalescer like scroll and selection, which also keeps `setTop` out of the observer callback. This made the existing resize test need a frame flush, which is a fair signal that the behavior actually changed.

**7. `EDITOR_PARA_SELECTOR`** (`bar.utils:7`) — you're right that the doc asserted an agreement nothing enforced. Moved the constant to `editor-dom.util.ts` (already the shared home both overlays import geometry helpers from) and replaced all three raw literals in `ParagraphMarkerTooltipOverlay`.

**8. Comment fixes** — `react-dom` internals reference collapsed to one line (`overlay:373`); the `insetInlineEnd` dependency on every intermediate wrapper being full-width now has its guard comment (`overlay:391`); the mount-site guarantee behind the un-self-correcting `.editor-input` return is stated (`overlay:164`); and the once-only scroll-container resolution now names the footnotes re-parent and why `overflow: hidden` on the `ResizablePanel` is what makes it safe (`overlay:267`).

**9. Multi-line selection** (`bar.utils:56`) — took the "document it" branch. The TSDoc now says a non-collapsed selection anchors to the selection top by design, and why: the bar reports coverage over the *whole* selection, so pinning it to the stable end keeps it from chasing the pointer while the thing it describes is still being chosen. Happy to switch to `focusNode` if you'd rather it track the moving end.

## Pushing back

**`isPowerMode` loading window** (`web-view:409`) — I don't think anything paints here. The container that mounts during the load window is `pointer-events-none` and empty: `CharacterMarkerToolbar` renders `null` while `useIsPowerMode()` is `undefined`, so there's no border, background or icon, just a zero-content absolutely-positioned box. The gutter *reservation* is driven by `editor-container-simple`, which is the pre-existing Simple-layout flash rather than anything this bar adds. Gating on the three-state value would also mean handling `boolean | undefined` at the ~6 other places the local memo feeds, which changes Power-mode chrome behavior on load — more than this buys. Say the word if you've actually seen it flash and I'll re-open it.

**Comment de-duplication** (`_simple-mode.scss:53`, `overlay:103`) — I follow the argument and I take the caveat about keeping the per-site coupling comments. I'd rather not do the consolidation sweep in this PR: the arithmetic comments sit in three files that a "point at the canonical home" edit would touch without changing behavior, and the gutter placement may be replaced with an overlay modal, in which case most of this text is deleted rather than relocated. If the placement sticks, I'll do the sweep as its own change where it's reviewable on its own terms.

**Batching the six style writes** (`overlay:226`) — real, but it's one forced flush only when the caret crosses a metrics boundary, and it's already behind the `metricsChanged` guard, which the icon-measurement fix above just made stricter. Batching would mean a second off-DOM element or a cached cssText, which costs more clarity than the flush costs frames.

**`loadingdone` breadth** (`overlay:308`) and **double `selectionchange`** (`test:67`) — agreed on both, and agreed with your own read that neither needs action now. Narrowing `loadingdone` means diffing the loaded face set against the editor's stack, which is more machinery than a cache invalidation deserves.

## Deferred, with reasons

**`useGutterAnchor` extraction** (`overlay:267`) — completely agree this is the right move at three consumers, and that the toolbar comment anticipating footnote/comment controls makes that likely. Holding at two, per the second-consumer rule.

**`svg` alignment probe** (`overlay:27`) — fair that it undercuts the "could anchor anything" claim in the props doc. Since it can't misbehave with today's only slot occupant, I'd rather introduce the `data-` attribute contract when there's a second occupant to design it against, so the attribute name isn't a guess.

**`localizedStrings` clone-to-launder-a-type** (`bar.component:80`) — will file the follow-up. The deferral is already documented at the workaround site, and the fix (typing `CharacterMarkerControlLocalizedStrings` as `LanguageStrings`) lands in `main`'s control, which the toolbar placement shares.

## The typecheck question

Not a stale `node_modules` on your end — CI genuinely cannot answer it right now, and neither can this PR. All three Build jobs are failing at the **Install packages** step, in `volta run pnpm nx devpub platform-editor` (the yalc dev-package link), which is well before anything on this branch is compiled. `main` fails identically at the same step on its latest commit, so it's pre-existing infrastructure breakage rather than anything here.

Locally, on this branch: `npm run typecheck` is clean across every workspace, `npm run lint` is 0 errors (4 warnings, all pre-existing and untouched — 2 `import/no-duplicates` in `resource-reference.utils.ts`, 2 `no-console`), prettier clean on changed files, and the extension's full suite is 753 passing.

jolierabideau:
Correction to my comment above: on item 1 (the empty-chapter comment at `web-view.tsx:1975`) I said the comment had been updated to name the ResizeObserver instead of crediting `useViewVisibility`. It had not — `a6b949d70a3` fixed the *behavior* but never touched `platform-scripture-editor.web-view.tsx`, so the sentence you flagged as wrong was still sitting there. Fixed in `b092e8c7238`; it now says the deferral does **not** cover this case and names the zero-geometry guard and the ResizeObserver as what actually recovers.

Everything else in that comment holds as written.

katherinejensen00 (approved):
:lgtm:

> @katherinejensen00 reviewed 9 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2660#-:-OziDXrj3f5Fv534b8pU:bnfp4nl)* status: all files reviewed, 21 unresolved discussions (waiting on jolierabideau).

PR #2662, opened 2026-08-10
PT-4312: Allow selecting a non-production registry server

katherinejensen00:
https://github.com/user-attachments/assets/2a4b578a-74ca-4528-a3f7-1711472d56e1

# Code Review Summary

**Branch**: PT-4312-multiple-registry-servers

**Base**: origin/main

**Date**: 2026-08-07

**Files changed**: ~24 (≈15 source/test/story + 9 generated `dist/` bundles; `package-lock.json` change reverted)

## Overview

Lets developers and testers point Platform.Bible at a non-production Paratext registry server, and makes the "Visit Paratext Registry" link open the site that matches the selected server.

- **PT-4312** — the server toggle (already in the collapsed **"Developer only"** section) now offers **Production / Development / Test**. Production stays the default; **QA is intentionally not shown** (it remains a valid persisted value that displays as Production).
- **PT-3620** — both "Visit Paratext Registry" links (the first-run step and the registration dialog) now resolve their URL from the selected server via one backend command, instead of being hardcoded to production.

**One source of truth:** a single C# command (`getParatextRegistryUrl`) maps the selected `ServerType` → registry site URL; both front ends call it and fall back to the production URL if it's unavailable, so the link is never blank.

## Where to look (reviewer's map)

Start with the first two rows — everything else follows from them. The `dist/` bundles are **generated build output; don't hand-read them** (they're committed because extensions consume the built package).

| Area | File | What it does |
| --- | --- | --- |
| **Backend (the core logic)** | `c-sharp/Users/ParatextRegistrationService.cs` | New `GetRegistryUrl(ServerType)` switch (dev/test/qa/prod) + the `getParatextRegistryUrl` PAPI command that reads the selected server. |
| **Backend tests** | `c-sharp-tests/Users/ParatextRegistrationServiceTests.cs` | Pins all four `ServerType` → URL mappings. |
| **Command contract** | `extensions/src/paratext-registration/src/types/paratext-registration.d.ts` | Declares the new `() => Promise<string>` command. |
| **Server toggle (shared UI)** | `lib/platform-bible-react/.../developer-section/developer-section.component.tsx` | Adds the **Test** option; QA stays hidden; keeps the "hidden value → Production" escape hatch. |
| **Link surface 1 (first run)** | `src/renderer/components/first-run/steps/identify-step.component.tsx` | Fetches the server-aware URL via `usePromise`; production fallback; skips in demo mode. |
| **Link surface 2 (registration dialog)** | `extensions/src/paratext-registration/src/paratext-registration.web-view.tsx` | Same fetch pattern via `usePromise`. |
| **Security hardening** | `lib/platform-bible-react/.../markdown-renderer.component.tsx` | Adds `rel="noopener noreferrer"` to `target="_blank"` links (the dialog link flows through here). |
| **Constant rename** | `extensions/src/paratext-registration/src/utils.ts` | `PARATEXT_REGISTRY_LINK` → `PRODUCTION_REGISTRY_URL` (the fallback). |
| Tests / stories / e2e | `*.test.tsx`, `developer-section.stories.tsx` (×2), `identify-step.component.stories.tsx` (new), `e2e-tests/.../internet-settings.spec.ts` | Coverage for the toggle, the link fetch, and the Test option. The new IdentifyStep stories include one that visually documents the registry link following the selected server (Test). |

## API Changes

- **`paratext-registration` extension** (`.../types/paratext-registration.d.ts`): Added `CommandHandlers['paratextRegistration.getParatextRegistryUrl']: () => Promise<string>` to the `papi-shared-types` augmentation. New, non-breaking; returns a primitive (no new types).
- **`lib/platform-bible-react` — `DeveloperSection` (`@experimental`)**: No signature change. Internal `ServerType` alias/JSDoc updated to surface `Test`; `DEVELOPER_SECTION_STRING_KEYS` gained `'%paratextRegistration_label_serverType_option_Test%'`. The exported `selectedServer: ServerType` / `onServerChange: (server: ServerType) => void` shapes are unchanged (the `ServerType` union already contained `'Test'`).
- **`lib/platform-bible-react` — `MarkdownRenderer`**: No signature change. Rendered `<a>` now emits `rel="noopener noreferrer"` when `anchorTarget === '_blank'` (behavior only).
- **`lib/platform-bible-utils`**: None.
- **`lib/papi-dts/papi.d.ts`**: None (extension command handlers live in the extension's own `.d.ts`, not the core bundle).
- **(Non-API, noted)** `extensions/.../utils.ts`: exported const `PARATEXT_REGISTRY_LINK` renamed to `PRODUCTION_REGISTRY_URL` (internal utils module; all in-repo references updated).

None.

- [x] **Stale second Storybook file** — `lib/platform-bible-react/src/stories/advanced/developer-section.stories.tsx` (separate from the extension's copy) omitted the Test mock-string key (so the new toggle rendered unlabeled) and had comments claiming Test collapses to Production. _(fixed during review: added a `TestActive` story, added the `Test` key to its `localizedStrings` mock, corrected the QA-collapse and disabled-toggle comments.)_
- [x] **`package-lock.json` root-name flip** — the only diff was `"name": "paranext-core"` → the worktree branch name, an accidental `npm install`-in-worktree artifact. _(fixed during review: reverted; not committed.)_

[Author response: Author approved fixing both — pure cleanups, no design tradeoff — and asked to also fix anything else that's a genuine improvement.]

### Minor — Consider

- [ ] ~~The registration web view's `fetchRegistryUrl` has no test~~ _(the repo does not unit-test web views — its sibling `internet-settings.web-view` is also untested; the logic is trivial and mirrors the fully-tested identify-step copy. Covered by manual verification.)_
- [ ] ~~The production URL literal is duplicated in three places (extension `utils.ts`, renderer step, C# `_` arm)~~ _(unavoidable across two languages/packages; the C# `GetRegistryUrl` is canonical and the two TS copies only shadow it for the offline/error fallback.)_
- [ ] ~~`fetchRegistryUrl`'s body is near-identical across the two front ends~~ _(they live in different packages; both already reuse the shared `usePromise` hook — a further shared abstraction isn't warranted for two consumers.)_
- [ ] ~~The "Visit Paratext Registry" label doesn't name the selected environment~~ _(changing it needs new localized strings and a UX decision — out of scope for this feature; flagged for the reviewer.)_
- [ ] ~~The new command's TSDoc could state the URL is environment-dependent~~ _(the `@returns` already says "for the server currently selected," which conveys this.)_
- [ ] ~~The `utils.ts` const rename is technically breaking~~ _(internal extension util, not public API; no out-of-repo importer.)_

[Author response: Author reviewed all six and did not request changes; each is informational or out of scope.]

#### Shared Regions Modified

None. (No `#region shared with` markers overlap the changed lines.)

#### Extension Config Changes

None. Only extension source/type-declaration/test/story files changed — no `webpack.config.ts`, `tsconfig.json`, `package.json`, or `.eslintrc*` under `extensions/`, so no propagation to the extension templates is needed.

## Positive Observations

- **One source of truth**: the server→URL mapping lives only in C# (`GetRegistryUrl`), exposed via one command; both front ends call it — no duplicated URL map in TS.
- **Thorough tests**: C# pins all four `ServerType` mappings; the DeveloperSection tests cover Test-active, click-Test, and the subtle "clicking active Test must NOT re-route to Production (unlike hidden QA)" edge case; the identify-step test was refactored to route command mocks by name (so the mount-time URL fetch can't consume a queued validation mock).
- **Localization**: uses the pre-existing `Test` key (en + es); no hardcoded user-visible strings introduced.
- **Security hardening**: `rel="noopener noreferrer"` added for `_blank` anchors, with an explanatory comment.
- **Clean reuse**: adopts the shared `usePromise` hook with stable module-scope callbacks; the `PARATEXT_REGISTRY_LINK` → `PRODUCTION_REGISTRY_URL` rename is clearer and the old symbol is gone everywhere.
- **Stable public API**: `ServerType` already included `'Test'`, so the exported prop shapes didn't change; the new command conforms to the existing sibling command convention (parameterless, resolves context from ParatextData internally).
- **Storybook coverage** (author-added): the new `IdentifyStep` stories cover the empty/valid/invalid/restart states and — most usefully for this feature — a story that demonstrates the "Visit Paratext Registry" link resolving to the selected server's site, documenting the PT-3620 behavior visually.

## Interview Notes

- **Purpose** (author's words): "Allow selecting a non-production Paratext registry server (Production/Development/Test; QA hidden, Production default), and make the 'Visit Paratext Registry' link open the registry site matching the selected server. (PT-4312 + PT-3620.)"
- The author directed the key design decisions throughout: QA stays hidden but its link still resolves to the QA site (logical consistency); reuse the existing `usePromise` hook rather than a bespoke effect; keep server selection subtle inside the "Developer only" disclosure; a single backend command as the source of truth.
- The author approved the two in-review fixes and asked to address any further genuine improvements; the six minor findings were reviewed and judged informational/out-of-scope.
- No unresolved items; the author demonstrated clear understanding of the changes (design walkthrough skipped as unnecessary). Nothing was deferred to AI.

## In-Review Quality Check

Changes made during review:
1. Fixed the stale lib `developer-section.stories.tsx` (added `TestActive`, added the Test mock key, corrected comments).
2. Reverted the `package-lock.json` name change.
3. In the author-added `identify-step.component.stories.tsx`, moved each `beforeEach` from under `parameters` to the story level. Storybook reads `beforeEach` as a top-level annotation; nested under `parameters` it is inert, so the `sendCommand`/demo-mode mocks never installed and the play assertions (e.g. the Test-registry-link story) would not hold. This matches the sibling `internet-settings-step.stories.tsx` pattern.

- **Format (prettier)**: clean — both story files already/again formatted.
- **Lint (eslint, lib workspace)**: clean on the lib story (0 errors). The renderer story's eslint run repeatedly timed out in this environment (~8 min); the change is purely structural (unwrapping `parameters`) and prettier-clean, so CI lint is the backstop.
- **Typecheck**: type-safe by construction — `'Test'` is a valid `ServerType`; `beforeEach` is a valid Story annotation (the sibling uses it). Full `typecheck` shows only the pre-existing, environmental `release/app/buildInfo.json` error (present on CI), unrelated to these changes.
- **Tests**: the DeveloperSection component test (12/12) is unaffected by Storybook-only changes.
- Storybook files are not bundled into `dist`, so no library rebuild was needed.

## Suggested Review Focus

- [ ] **Registry URLs** — confirm `registry-dev` / `registry-test` / `registry-qa.paratext.org` are the correct hosts (dev confirmed by PT-3620; test/qa follow the `registry-dev` naming convention and are verified manually, not in CI).
- [ ] **QA behavior (deliberate)** — a persisted QA value shows as Production in the toggle but its link resolves to the QA site. Confirm that's the intended trade-off.
- [ ] **Registration-dialog link** — this second link surface is covered by manual verification, not a unit test (per the repo's no-web-view-test convention). Worth a manual click-through.
- [ ] **`dist/` diff** — 9 generated bundles are included (the repo's convention is to commit the built `platform-bible-react`); they're not meant to be hand-read.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2662)


tombogle:
I notice in this video that on Step 2, the heading shows the Title as "Set up Platform.Bible" with the subtitle "Step 2 of 4", but then with no further explanation, the first radio button just says "Unrestricted". Under that is explanatory text that makes it reasonably clear *what* is not restricted, but the user kind of has to read from the bottom up to make sense of it. This is really a @paranext/paranext-ux question, but it seems to me that something is needed higher up to set the stage for the decision to be made.

tombogle:
FYI: regarding the package-lock problem, #2688 should prevent this kind of churn.

tombogle (changes requested):
> @tombogle partially reviewed 31 files and all commit messages, and made 7 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2662)* status: all files reviewed, 7 unresolved discussions (waiting on katherinejensen00).

___
*[`lib/platform-bible-react/src/components/advanced/developer-section/developer-section.component.tsx` line 47 at r1](https://reviewable.io/reviews/paranext/paranext-core/2662#-P-GfOPG9AF-DO7R1-bg:-P-GfOPG9AF-DO7R1-bh:b-cf75kh) ([raw file](https://github.com/paranext/paranext-core/blob/0cf5f2b99352dacb6027298329f76fd511e60ff8/lib/platform-bible-react/src/components/advanced/developer-section/developer-section.component.tsx#L47)):*
> ```XML
> }: DeveloperSectionProps) {
>   const [isExpanded, setIsExpanded] = useState(false);
>   // QA (and any unknown value) is not surfaced in this UI; it displays as Production.
> ```

I see comments and PR notes throughout explaining that QA resolves to production in the UI. But it actually implies a different URL. What is this all about? Is it explained somewhere? How/where/why would the setting ever be QA?
___
*[`c-sharp-tests/Users/ParatextRegistrationServiceTests.cs` line 7 at r1](https://reviewable.io/reviews/paranext/paranext-core/2662#-P-G_tzME8g5DTYEwTtZ:-P-G_tzME8g5DTYEwTt_:b-tjmt5k) ([raw file](https://github.com/paranext/paranext-core/blob/0cf5f2b99352dacb6027298329f76fd511e60ff8/c-sharp-tests/Users/ParatextRegistrationServiceTests.cs#L7)):*
> ```Smalltalk
> 
> [TestFixture]
> internal class ParatextRegistrationServiceTests
> ```

Can't help but wonder if this is taking TDD to a ridiculous extreme. If this rather trivial function deserves a test at all, this would be more compact and descriptive:



_Code snippet:_
```Smalltalk
[TestCase(ServerType.Development, ExpectedResult = "https://registry-dev.paratext.org/")]
[TestCase(ServerType.Test, ExpectedResult = "https://registry-test.paratext.org/")]
[TestCase(ServerType.QualityAssurance, ExpectedResult = "https://registry-qa.paratext.org/")]
[TestCase(ServerType.Production, ExpectedResult = "https://registry.paratext.org/")]
public string GetRegistryUrl_ReturnsExpectedUrl(ServerType server) =>
    ParatextRegistrationService.GetRegistryUrl(server);
```
___
*[`extensions/src/paratext-registration/src/types/paratext-registration.d.ts` line 28 at r1](https://reviewable.io/reviews/paranext/paranext-core/2662#-P-GZHKK6rzVmehA2sW_:-P-GZHKK6rzVmehA2sWa:b-d69m3m) ([raw file](https://github.com/paranext/paranext-core/blob/0cf5f2b99352dacb6027298329f76fd511e60ff8/extensions/src/paratext-registration/src/types/paratext-registration.d.ts#L28)):*
> ```XML
> 
>   /**
>    * Set of servers to which to connect in `ParatextData.dll`
> ```

Though not changed in this PR, I wonder if it might be helpful to update this comment to read:
The Paratext Registry environment that corresponds to a particular server connection in `ParatextData.dll`.
Sadly, as named "ServerType" is woefully vague, but rather than paying the high cost of renaming it, I think we get some benefit from trying to clearly explain its actual meaning.
___
*[`extensions/src/paratext-registration/src/paratext-registration.web-view.tsx` line 19 at r1](https://reviewable.io/reviews/paranext/paranext-core/2662#-P-GccHfDpNUnJOKocQh:-P-GccHfDpNUnJOKocQi:b1ok4ef) ([raw file](https://github.com/paranext/paranext-core/blob/0cf5f2b99352dacb6027298329f76fd511e60ff8/extensions/src/paratext-registration/src/paratext-registration.web-view.tsx#L19)):*
> ```XML
> /**
>  * Fetches the registry site URL for the selected server, falling back to production so the link is
>  * never blank or broken. Module-scope so it is a stable `usePromise` callback.
> ```

This wording is technically **too strong**. The fallback ensures the link is never _blank_ when this function succeeds or throws, but it doesn't guarantee that the resulting link isn't broken. If we someday retire one or more of those URLs, or the URL is unreachable (e.g., because access is blocked in the user's host country), the user will experience it as broken even though it resolved to a non-blank URL.
I don't know that we can or would necessarily want to do something about that, but at least the comment should reflect reality.
___
*[`extensions/src/paratext-registration/src/types/paratext-registration.d.ts` line 159 at r1](https://reviewable.io/reviews/paranext/paranext-core/2662#-P-GePx66knem4dJWj2G:-P-GePx66knem4dJWj2H:b-gwbup4) ([raw file](https://github.com/paranext/paranext-core/blob/0cf5f2b99352dacb6027298329f76fd511e60ff8/extensions/src/paratext-registration/src/types/paratext-registration.d.ts#L159)):*
> ```XML
>     ) => Promise<void>;
>     /**
>      * Gets the Paratext Registry website URL for the currently-selected registry server.
> ```

NIT: Maybe "...for the currently-selected registry environment." would be better. The user selects the environment in which they wish to operate. That *implies* a particular registry server, but in the future it could imply other things (e.g., soon, which analytics project token to use).
___
*[`src/renderer/components/first-run/steps/identify-step.component.tsx` line 48 at r1](https://reviewable.io/reviews/paranext/paranext-core/2662#-P-GgkGPEPHEz_a2XKpH:-P-GgkGPEPHEz_a2XKpI:b-ar56u9) ([raw file](https://github.com/paranext/paranext-core/blob/0cf5f2b99352dacb6027298329f76fd511e60ff8/src/renderer/components/first-run/steps/identify-step.component.tsx#L48)):*
> ```XML
> /**
>  * Fetches the registry site URL for the selected server, falling back to production so the link is
>  * never blank or broken (and skips the lookup entirely in demo mode). Module-scope so it is a
> ```

Ditto re: "broken"
___
*[`src/renderer/components/first-run/steps/identify-step.component.test.tsx` line 357 at r1](https://reviewable.io/reviews/paranext/paranext-core/2662#-P-Gi-AQ5t8oqhGg6EPh:-P-Gi-AQ5t8oqhGg6EPi:b-3xjup6) ([raw file](https://github.com/paranext/paranext-core/blob/0cf5f2b99352dacb6027298329f76fd511e60ff8/src/renderer/components/first-run/steps/identify-step.component.test.tsx#L357)):*
> ```XML
>       expect(mockSendCommand).toHaveBeenCalledWith('paratextRegistration.getParatextRegistryUrl'),
>     );
>     // Never blank/broken: the failed lookup leaves the link on the production fallback.
> ```

ditto: "broken"

katherinejensen00 (commented):
Agreed — having to read bottom-up to understand the choice isn't right. That step came in with PT-4235 and isn't touched here, and new framing copy needs localized strings plus a UX call, so I'd rather not decide it inside a registry-server PR. I'll raise it with @paranext/paranext-ux separately. I created a thread in Discord to ask UX, https://discord.com/channels/892072317436448768/1451250952236568677/1539466229679525919

Good to know — fixing it at the source beats catching it in review each time. The lock change here was reverted.

> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2662#-:-P-Mt83K6h0qGd2KMrUB:b-n0jbmy)* status: 19 of 31 files reviewed, 7 unresolved discussions (waiting on tombogle).

___
*[`extensions/src/paratext-registration/src/paratext-registration.web-view.tsx` line 19 at r1](https://reviewable.io/reviews/paranext/paranext-core/2662#-P-GccHfDpNUnJOKocQh:-P-MskiJEwxuA6-fxsF4:b-m7vrnh) ([raw file](https://github.com/paranext/paranext-core/blob/0cf5f2b99352dacb6027298329f76fd511e60ff8/extensions/src/paratext-registration/src/paratext-registration.web-view.tsx#L19)):*

This wording is technically **too strong**. The fallback ensures the link is never _blank_ when this function succeeds or throws, but it doesn't guarantee that the resulting link isn't broken. If we someday retire one or more of those URLs, or the URL is unreachable (e.g., because access is blocked in the user's host country), the user will experience it as broken even though it resolved to a non-blank URL.
I don't know that we can or would necessarily want to do something about that, but at least the comment should reflect reality.
</blockquote></details>

Fair — and pointed, given the URLs I'd hardcoded were themselves unreachable. Reworded to claim only what it delivers:

Fetches the registry site URL for the selected server environment, falling back to production so the link always has a target rather than going blank. That is all the fallback guarantees — whether the target is reachable is a separate matter (a retired host, or access blocked from the user's network or country, still reads as a broken link).
___
*[`extensions/src/paratext-registration/src/types/paratext-registration.d.ts` line 28 at r1](https://reviewable.io/reviews/paranext/paranext-core/2662#-P-GZHKK6rzVmehA2sW_:-P-MsbjS-2aY7r9pevoG:b-ce03f3) ([raw file](https://github.com/paranext/paranext-core/blob/0cf5f2b99352dacb6027298329f76fd511e60ff8/extensions/src/paratext-registration/src/types/paratext-registration.d.ts#L28)):*

Though not changed in this PR, I wonder if it might be helpful to update this comment to read:
The Paratext Registry environment that corresponds to a particular server connection in `ParatextData.dll`.
Sadly, as named "ServerType" is woefully vague, but rather than paying the high cost of renaming it, I think we get some benefit from trying to clearly explain its actual meaning.
</blockquote></details>

Agreed on both counts. I went a bit broader than your wording because it turns out ServerType isn't registry-specific — probing ParatextData, one value swings Send/Receive, Registry, Paratext Live, and Digital Bible Library together:

The server environment ParatextData.dll operates in. Selecting one picks the whole matching set of back ends — Send/Receive archive, Paratext Registry, Paratext Live, and Digital Bible Library — not just a single server. ParatextData.dll owns those mappings, and several environments deliberately share a back end (for example, Development, Test, and QualityAssurance all use the same registry).

I called out the shared-back-end part explicitly since that's the genuinely surprising bit.
___
*[`extensions/src/paratext-registration/src/types/paratext-registration.d.ts` line 159 at r1](https://reviewable.io/reviews/paranext/paranext-core/2662#-P-GePx66knem4dJWj2G:-P-MstRY2GW_lgNIilyg:b-a42glr) ([raw file](https://github.com/paranext/paranext-core/blob/0cf5f2b99352dacb6027298329f76fd511e60ff8/extensions/src/paratext-registration/src/types/paratext-registration.d.ts#L159)):*

NIT: Maybe "...for the currently-selected registry environment." would be better. The user selects the environment in which they wish to operate. That *implies* a particular registry server, but in the future it could imply other things (e.g., soon, which analytics project token to use).
</blockquote></details>

Taken — and your reasoning is already true today, not just in future: one ServerType selects Send/Receive, Registry, Paratext Live, and DBL together. So I used "server environment" rather than "registry environment", since the environment is the broader thing and the registry is one consequence of it. Both the summary and the @returns now say it.
___
*[`lib/platform-bible-react/src/components/advanced/developer-section/developer-section.component.tsx` line 47 at r1](https://reviewable.io/reviews/paranext/paranext-core/2662#-P-GfOPG9AF-DO7R1-bg:-P-Ms6_a0zH4go6ae3FU:b-gxf0h3) ([raw file](https://github.com/paranext/paranext-core/blob/0cf5f2b99352dacb6027298329f76fd511e60ff8/lib/platform-bible-react/src/components/advanced/developer-section/developer-section.component.tsx#L47)):*

I see comments and PR notes throughout explaining that QA resolves to production in the UI. But it actually implies a different URL. What is this all about? Is it explained somewhere? How/where/why would the setting ever be QA?
</blockquote></details>

Pushing on this turned up a bug, so I'm glad you asked.

How it could ever be QA: nothing in Platform.Bible's UI writes that value — this toggle only offers Production/Development/Test. It can only arrive from ParatextData's persisted InternetSettings.xml, either through the internet-settings API (which accepts the full ServerType enum) or from a settings file that already had it. The "display it as Production" handling predates this PR — it came in with PT-4235; this PR only added Test.

The different-URL part was a real defect. I checked ParatextData 9.5.0.22 directly: InternetAccess.RegistryServer returns https://registry.paratext.org for Production and https://registry-dev.paratext.org for Development, Test and QualityAssurance. Only two registry hosts exist. The registry-test.paratext.org and registry-qa.paratext.org I'd hardcoded don't resolve at all, so picking Test gave a dead link to a server the app never talks to — the opposite of what PT-3620 wanted.

Fixed by deleting the switch and returning InternetAccess.RegistryServer, so the link always matches the registry the rest of the app uses. I've also documented at the toggle where a QA value can come from.
___
*[`src/renderer/components/first-run/steps/identify-step.component.tsx` line 48 at r1](https://reviewable.io/reviews/paranext/paranext-core/2662#-P-GgkGPEPHEz_a2XKpH:-P-MswXO-WqvlForN8mp:b-meozo9) ([raw file](https://github.com/paranext/paranext-core/blob/0cf5f2b99352dacb6027298329f76fd511e60ff8/src/renderer/components/first-run/steps/identify-step.component.tsx#L48)):*

Ditto re: "broken"
</blockquote></details>

Same rewording here, plus the demo-mode note it already carried.
___
*[`src/renderer/components/first-run/steps/identify-step.component.test.tsx` line 357 at r1](https://reviewable.io/reviews/paranext/paranext-core/2662#-P-Gi-AQ5t8oqhGg6EPh:-P-Mszl451iK1YmDn1yA:b7ldm7f) ([raw file](https://github.com/paranext/paranext-core/blob/0cf5f2b99352dacb6027298329f76fd511e60ff8/src/renderer/components/first-run/steps/identify-step.component.test.tsx#L357)):*

ditto: "broken"
</blockquote></details>

Reworded. While in there I found the test itself couldn't fail: it asserted the href equals the production URL, which is also usePromise's initial value, so it passed whether or not the catch ever ran. The usePromise stand-in was also dropping undefined results, unlike the real hook, which hid the regression. Fixed both — verified by deleting the catch's fallback and confirming the test now fails with href: null.
___
*[`c-sharp-tests/Users/ParatextRegistrationServiceTests.cs` line 7 at r1](https://reviewable.io/reviews/paranext/paranext-core/2662#-P-G_tzME8g5DTYEwTtZ:-P-MsNg962x82_sefDQl:b-i49skf) ([raw file](https://github.com/paranext/paranext-core/blob/0cf5f2b99352dacb6027298329f76fd511e60ff8/c-sharp-tests/Users/ParatextRegistrationServiceTests.cs#L7)):*

Can't help but wonder if this is taking TDD to a ridiculous extreme. If this rather trivial function deserves a test at all, this would be more compact and descriptive:
</blockquote></details>

My bad. You were right, Tom, that it didn't deserve a test — and worse, it was pinning URLs that turn out to be wrong (see the QA thread). 

Rather than convert it to TestCase, the whole thing is gone: GetParatextRegistryUrl now returns InternetAccess.RegistryServer, so ParatextData owns the mapping and there's none of our logic left to test. ParatextRegistrationServiceTests.cs is deleted.

tombogle:
*[`extensions/src/paratext-registration/src/types/paratext-registration.d.ts` line 28 at r1](https://reviewable.io/reviews/paranext/paranext-core/2662#-P-GZHKK6rzVmehA2sW_:-P-PAuz-1lL8fJcFUW_4:b-z23j46) ([raw file](https://github.com/paranext/paranext-core/blob/0cf5f2b99352dacb6027298329f76fd511e60ff8/extensions/src/paratext-registration/src/types/paratext-registration.d.ts#L28)):*

Agreed on both counts. I went a bit broader than your wording because it turns out ServerType isn't registry-specific — probing ParatextData, one value swings Send/Receive, Registry, Paratext Live, and Digital Bible Library together:

The server environment ParatextData.dll operates in. Selecting one picks the whole matching set of back ends — Send/Receive archive, Paratext Registry, Paratext Live, and Digital Bible Library — not just a single server. ParatextData.dll owns those mappings, and several environments deliberately share a back end (for example, Development, Test, and QualityAssurance all use the same registry).

I called out the shared-back-end part explicitly since that's the genuinely surprising bit.
</blockquote></details>

Good sleuthing!

tombogle (commented):
> @tombogle reviewed 12 files and all commit messages, made 3 comments, and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2662)* status: all files reviewed, 5 unresolved discussions (waiting on katherinejensen00).

___
*[`lib/platform-bible-react/src/components/advanced/developer-section/developer-section.component.tsx` line 47 at r1](https://reviewable.io/reviews/paranext/paranext-core/2662#-P-GfOPG9AF-DO7R1-bg:-P-PBtNcCoVC46sQJZmE:b-ybd0ml) ([raw file](https://github.com/paranext/paranext-core/blob/0cf5f2b99352dacb6027298329f76fd511e60ff8/lib/platform-bible-react/src/components/advanced/developer-section/developer-section.component.tsx#L47)):*

Pushing on this turned up a bug, so I'm glad you asked.

How it could ever be QA: nothing in Platform.Bible's UI writes that value — this toggle only offers Production/Development/Test. It can only arrive from ParatextData's persisted InternetSettings.xml, either through the internet-settings API (which accepts the full ServerType enum) or from a settings file that already had it. The "display it as Production" handling predates this PR — it came in with PT-4235; this PR only added Test.

The different-URL part was a real defect. I checked ParatextData 9.5.0.22 directly: InternetAccess.RegistryServer returns https://registry.paratext.org for Production and https://registry-dev.paratext.org for Development, Test and QualityAssurance. Only two registry hosts exist. The registry-test.paratext.org and registry-qa.paratext.org I'd hardcoded don't resolve at all, so picking Test gave a dead link to a server the app never talks to — the opposite of what PT-3620 wanted.

Fixed by deleting the switch and returning InternetAccess.RegistryServer, so the link always matches the registry the rest of the app uses. I've also documented at the toggle where a QA value can come from.
</blockquote></details>

You say, "InternetAccess.RegistryServer returns [https://registry.paratext.org](https://registry.paratext.org) for Production and [https://registry-dev.paratext.org](https://registry-dev.paratext.org) for Development, Test and QualityAssurance." but the code still returns "Production" if selectedServer is QA. I'm not marking this as blocking, because I'm guessing you thought about this and it is the right thing to do (and was pre-existing). But worth a quick double-check (and maybe a clarifying explanation as to why QA (or other unknown values) map to "Production" since falling back to non-production feels like a safer choice for a scenario where and unknown/unexpected option appears.) Since this is really a rabbit trail for this PR, you could also just add a REVIEW comment or create a Jira issue to have this investigated further.
___
*[`c-sharp/Users/ParatextRegistrationService.cs` line 237 at r2](https://reviewable.io/reviews/paranext/paranext-core/2662#-P-PEht30i7e2Thncxuv:-P-PEht490SB_w7_Z-DX:b-xv2xjs) ([raw file](https://github.com/paranext/paranext-core/blob/447eae3ad48a79770213e0412cb10b4c57643aea/c-sharp/Users/ParatextRegistrationService.cs#L237)):*
> ```Smalltalk
>     /// <para>
>     /// ParatextData owns this mapping, so the link always points at the registry the rest of the
>     /// app is actually talking to. Today it resolves Production to
> ```

This is okay, but "Today" never goes out of date. You'd have to use Blame to see what "Today" referred to. You could put in an actual date, just omit this part of the comment, or change it to be an "example" URL with no stated guarantee of correspondence to reality: `(e.g., "https://registry.paratext.org" or"https://registry-dev.paratext.org)`
___
*[`src/renderer/components/first-run/steps/identify-step.component.stories.tsx` line 9 at r2](https://reviewable.io/reviews/paranext/paranext-core/2662#-P-PDKrK93m3tK-NPR2Q:-P-PDKrK93m3tK-NPR2R:b-ll9e1a) ([raw file](https://github.com/paranext/paranext-core/blob/447eae3ad48a79770213e0412cb10b4c57643aea/src/renderer/components/first-run/steps/identify-step.component.stories.tsx#L9)):*
> ```XML
> const DEMO_MODE_KEY = 'platform-bible.firstRunDemoMode';
> // The registry site a non-production environment resolves to. ParatextData maps Development, Test,
> // and QualityAssurance all to this one host; only Production differs. Used here to show the "Visit
> ```

See earlier comment. This is potentially confusing if ParatextData maps QA to registry-dev.paratext.org, but *we* reinterpret QA as Production.

tombogle (approved):
> @tombogle resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2662)* status: all files reviewed, 3 unresolved discussions (waiting on katherinejensen00).

katherinejensen00 (commented):
> @katherinejensen00 resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2662)* status: 10 of 42 files reviewed, all discussions resolved (waiting on tombogle).

PR #2586, opened 2026-07-22
docs: forward-facing over backward-facing comment rule + review sweep

jolierabideau:
This branch started as a narrow rule banning Jira IDs in code comments. That fixed a *symptom* of
a broader problem: AI-authored diffs systematically over-produce **backward-facing** comments —
notes that only record how the code reached its current state during a PR and mean nothing to a
future reader. Jira IDs are the most common case, but PR numbers, review-finding IDs, stage/epic
tags, and dated dev notes rot the same way. The old framing was also both too narrow (missed those)
and too broad (would wrongly flag legitimately *forward-facing* `PT-XXXX` references — an open
follow-up, or a non-inlinable deep link).

So the rule is now built on the real principle: **prefer forward-facing comments over
backward-facing ones**. A comment earns its place only if it helps someone who reads the code later
and never saw this PR.

## Changes

- **Renamed** `.claude/rules/code-quality/no-jira-in-comments.md` →
  `forward-facing-comments.md` and rewrote it around the strip-the-PR-context test, with explicit
  cut/keep lists and the forward-facing carve-outs (`TODO(PT-XXXX)`, non-inlinable deep links,
  constraints, rationale). The closing guidance is stated as an explicit author action, with the
  PR-level aggregation scoped to the reviewing case.
- **`pr-reviewer-clarity.md`** (`/review-pr`): added a "Backward-Facing Comments" check
  (severity `warning`), and reconciled the pre-existing "all findings include file paths and line
  numbers" checklist item with the file-level (no-`line`) exception for pervasive findings. The
  carve-out is anchored to the rule's "survives the merge" test rather than an exhaustive list of
  permitted forms, so a bare forward-facing rationale isn't flagged. The adjacent
  "Leftover TODOs/FIXMEs" check is scoped to *untracked* TODOs so the two checks can't return
  opposite verdicts on the same `TODO(PT-XXXX)`.
- **`review-analyzer.md`** (`/review-paratext`, Priority 6): added a matching backward-facing bullet
  carrying the same carve-outs and the same "survives the merge" framing, so legitimate references
  aren't flagged.
- **`.context/standards/Code-Style-Guide.md`**: PR #2621 (merged 2026-08-10, after this branch
  opened) added an independently-worded statement of the same rule under
  `## TypeScript-specific Guidelines`. Rather than ship the rule stated in full in two places, that
  bullet is now a one-line pointer to the rule, moved up into the language-agnostic principles
[...]

PR #2509, opened 2026-07-06
feat(scripture-text-grid): A4 Add ResourceCell with offline cell

jolierabideau:
# Code Review Summary

**Branch**: pt-4052-scripture-text-grid-verse-cell

**Base**: origin/main

**Date**: 2026-07-09

**Review model**: Claude Fable 5

**Files changed**: 27

## Overview

PR 1 of PT-4052 (A4 — ScriptureTextGrid renderer) for the PT-4047 epic "Multi-Resource View in Resource Viewer": adds the `ResourceCell` cell renderer for the Scripture Text Grid in `platform-scripture-editor` — a presentational/connected split (`ResourceCellView` / `ResourceCell`) rendering the focused chapter (or a sliced single verse in verse mode) via `Editorial`, with offline cell states (downloading/failed), per-cell text direction from the resource's own `platform.textDirection` setting, and verse-slicing utilities handling combined verses (PT-3495) and empty/verse-0 states (PT-3133). Because A3 (data layer) has not landed, the branch also builds what the grid needs from that layer: the `scripture-text-grid-contents` selector/mutation helpers, the C# per-user shown-by-default overlay with first-open init, auto-promote of admin-flagged resources into the download path, and a server-side admin gate on the two project-scope settings (`dataVersion` bumped 1.0.0 → 1.1.0 with `ExtraData` forward-compat).

Note for reviewers: `docs/superpowers/prd/PT-4052-A4-chapter-first-description.md` is a **non-adopted draft** proposing a chapter-first pivot that defers the verse view. Per the author, the verse view is **not** deferred — it is in scope (matching the branch name and the parent issue's "Default view mode: Verse" decision). Don't treat that draft as the source of truth.

## API Changes

All public-API changes are in `extensions/src/platform-scripture/src/types/platform-scripture.d.ts` (module `platform-scripture`). No changes to `lib/platform-bible-react/`, `lib/platform-bible-utils/`, or `@papi/` modules in `papi.d.ts`.

- `platform-scripture`: Added exported type `ShownByDefaultOverlay` (`{ [resourceId: string]: boolean }`)
- `platform-scripture`: Added `ShownByDefaultOverlay` data type to `UserTextConnectionSettingsProjectInterfaceDataTypes`
[...]

PR #2663, opened 2026-08-10
PT-4061: show verse 1 at a verse-0 reference in the Text Collection

katherinejensen00:
Resolves [PT-3133](https://paratextstudio.atlassian.net/browse/PT-3133). Jira: [PT-4061](https://paratextstudio.atlassian.net/browse/PT-4061).

AI-assisted — session link to be added.

## What and why

A verse-0 reference means "everything preceding verse 1" — intros, titles, Psalm `\d` superscriptions — which a one-verse-tall Text Collection cell cannot render usefully. Paratext 9 shows verse 1 there, and PT-3133's written acceptance adopts that.

A4 (#2509) had shipped a "No text for this verse" empty state instead. That was Todd's suggestion on PT-3133, which Ian had called reasonable but explicitly "a different use case" from this one. Nobody flagged the divergence, so PT-3133 sat looking fixed while contradicting its own acceptance. **Ian re-confirmed PT9 parity on 2026-08-05** before this reversal was written.

Fall-forward is **display-only**: the scroll group stays at verse 0, so the Scripture Editor is not pulled off the intro the user clicked into to get here.

## Please scrutinize this part

**The editor write-back guard** (`resource-cell.component.tsx`). It is the subtlest thing in the diff, and the reasoning around it was wrong twice before landing here — see round 3.

Where it ended up: the guard is **defense-in-depth, not a fix for a live report**. Against the editor this repo actually builds (`dev-packages/scripture-editors` `packages/platform` 0.8.15, yalc-linked by `postinstall` → `link-dev-packages`), `$resolvePosition` refuses to describe a position in a document with no `BookNode` and no `ChapterNode` — upstream invariant I5 — and `sliceUsjToVerse` drops both. So the plugin is silent in verse mode and the `viewMode === 'verse'` branch is unreachable today.

It is kept because it costs nothing and is the right shape if slices ever become addressable, and because the mismatch is genuinely ours to own: we hand the editor verse 1's text while telling it verse 0. What it is **not** is evidence that a write-back currently occurs.

## Changes

- **`resolveDisplayVerseNum`** — verse 0 → verse 1, every other verse unchanged. `sliceUsjToVerse` stays deliberately mechanical: the rule is about the *reference*, not the USJ, and the accessible-name site needs it without having any USJ.
- **Write-back guard** — described above.
- **Grid accessible name** resolves the same way, so the announcement names the verse actually displayed. It names the row, not its contents (a resource lacking verse 1 shows the empty state under the same label).
- **Chapter surfaces stay exempt by design** — chapter mode, the chapter-context split, the single-resource grid path, and the Resource Viewer all show verse-0 front matter directly. A comment at the Resource Viewer's chapter fetch records why it must **not** consume the slice; wiring it up would blank the intro and was the original ticket's (incorrect) instruction.
- **ADR-0019** records the decision, the rejected alternatives, and the condition to revisit.

## Three accepted trade-offs

All are inherent to the decision rather than defects, and are documented in ADR-0019 rather than fixed:

1. Verse 0 and verse 1 now render identically with the same accessible name, so stepping across that boundary produces no perceptible change. The alternative is a row of ghost text at every chapter boundary.
2. Verse-0 content is not shown in verse view. It stays one click away via the chapter-context split, which renders the chapter unsliced — that escape hatch is what makes the trade acceptable.

3. **Verse 0 becomes unobservable inside the grid, including to assistive tech.** At MAT 5:0 the row label announces "MAT 5:1" and the cells show verse 1, so a screen-reader user gets no in-grid signal that the shared reference is 5:0. Naming the row after a verse the cells demonstrably do not show would be worse. The boundary cues live outside the grid: the BCV control still displays `5:0`, and its previous-verse button is **disabled** there (an announced state, not an inert control — pinned by `'Is disabled when at verse 0'` in `book-chapter-control.navigation.test.ts`).

## Testing

705 unit tests pass. Typecheck and lint clean on all changed files.

New coverage: `resolveDisplayVerseNum` (0→1, identity otherwise); verse 0 displays verse 1; front matter (intro + superscription + heading) never leaks into the cell; empty state when verse 1 is absent too, without scanning forward to verse 2; combined opener emits its marker once; the editor receives the unresolved reference; the write-back is swallowed at verse 0 but still reported for normal verses; chapter mode keeps front matter. The `sliceUsjToVerse(usj, 0) → empty` contract is pinned, since `resolveDisplayVerseNum` exists only because the slicer stays mechanical.

**Verified in the running app (2026-08-11).** Full manual pass, 10/10. Most importantly, clicking inside a cell at Luke 1:0 leaves the reference at 1:0, and the Luke 5:0 chapter-jump case holds — those two are the defect review caught, which no unit test could reach. Also confirmed: Psalm 3:0 shows only verse 1 with no superscription or heading leaking in, missing verses still show the ghost text, and the chapter surfaces (chapter view, single-resource path, Resource Viewer) are unchanged.

## Review round 2 (2026-08-13)

Five inline comments from @imnasnainaec, all addressed. Four are documentation-only; one is a one-line comment correction in `platform-bible-react`.

- **Content-aware fall-forward** is now in ADR-0019's rejected alternatives as **(d)**, with the reasoning: it would need the chapter USJ that the accessible-name site does not have, and per-resource content-awareness would make the same reference fall forward in one cell but not its neighbor.
- **Why the write-back guard is not upstream.** ADR-0019 now states this was considered and rejected on the merits, not deferred. `ScriptureReferencePlugin` is bidirectional (it also moves the caret to `scrRef`), and these cells depend on its reporting for click-to-sync at every normal verse — so an `isReadonly` gate would break read-only navigation *and* our own click-to-sync. The mismatch is host-made, so the host owns it. Verified against `@eten-tech-foundation/platform-editor` ~0.8.14.
- **The a11y cost is now stated** as accepted rather than implied — see trade-off 3 above, recorded in both ADR-0019 and a comment at the accessible-name site.
- **The `bounds` dead end is filed:** [PT-4379](https://paratextstudio.atlassian.net/browse/PT-4379), linked from ADR-0019. It is not a one-line fix — `ScriptureBounds` is built async over PAPI, and `BookChapterControl` lives in `platform-bible-react` with no PAPI access, so it must be threaded as a prop through every consumer of the BCV control.
- **"Second consumer" → "second consumer _of this rule_"**, with the reason: whole-chapter surfaces handle verse 0 under a deliberately different rule, so promoting on a surface count would turn an intentional difference into apparent drift.

**One correction to my own earlier text:** ADR-0019 previously called the previous-verse button "inert" at verse 0 and said fall-forward removed its "last remaining cue." Both were wrong — `isNoOpNavigation` disables the button there, and a test pins it. The ADR now says so and notes the correction. Separately, the comment at `book-chapter-control.navigation.ts` attributed this gap to PT-4143, which is actually about global keyboard shortcuts overriding context-sensitive ones; that citation now points at PT-4379.

[PT-3133]: https://paratextstudio.atlassian.net/browse/PT-3133?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PT-4061]: https://paratextstudio.atlassian.net/browse/PT-4061?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Review round 3 (2026-08-16)

Five inline comments from @imnasnainaec on the editor mechanism this PR cited, plus one ⛏️. **He was right and I was wrong**, for a reason worth recording.

**I had the wrong editor.** My local `node_modules` held a stale npm 0.8.14 tarball, so I traced `$findAndSetChapterAndVerse` and its `parseInt(chapterNode?.getNumber() ?? "1", 10)` chapter-1 fallback and wrote them up as the live mechanism — in the ADR, the guard comment, the test comments, and this description. That plugin is not what runs. `postinstall` → `link-dev-packages` builds `dev-packages/scripture-editors` `packages/platform` at **0.8.15** and yalc-links it over `node_modules`; the 0.8.14 in `package-lock.json` is a placeholder that link replaces. CI does the same thing on every run.

I spent a while arguing the opposite from npm metadata (0.8.15 isn't published, `latest` is 0.8.14, and `main` imports four APIs 0.8.14 doesn't export). All true, and all beside the point — the repo consumes the editor from source, so the registry never enters into it. The four "missing" APIs were just my stale link.

**So, corrected throughout:**

- The guard is **defense-in-depth**, not a fix for a live report — `$resolvePosition` won't describe an unaddressable document (I5), the slice is exactly that, and the `viewMode === 'verse'` branch is unreachable today. **Luke 5:0 never becomes Luke 1:1.** Wording follows Ian's suggested replacement closely.
- Bidirectionality now cites `$moveCaretToVerseStart` and `$resolvePosition` → `onSelectionSettled` → `report`.
- **"relies on at every normal verse" was wrong**, now scoped to this grid's chapter mode and the Resource Viewer — both feed a whole chapter, so the document is addressable.
- The two write-back tests no longer claim to describe plugin behavior. `Editorial` is mocked; they pin *our* side of the channel, with constructed payloads, and say so.
- The version-comparison passage built on the same error is gone.

**Verified against the right source this time:** `dev-packages/.../ScriptureReferencePlugin.tsx` differs from upstream `main` only by a caret-placement tweak in `$moveCaretToVerseStart` — exactly as Ian said.

**Also in this round:** rebased onto latest `main`; dropped a diff-justifying parenthetical from `book-chapter-control.navigation.ts`; and added `'does not swallow the write-back in chapter mode at verse 0'`, which pins the one part of the guard that is *not* hypothetical — chapter mode is addressable, so the plugin really does report there, and every prior test held `viewMode` at `'verse'`. Verified falsifiable by reverting the check.

**Not taken:** the ⛏️ on `platform-scripture-editor.web-view.tsx`'s stale `versificationStr` comment. Confirmed correct (`positionToScrRef` carries the host's `versificationStr`), but it is outside this diff and Ian noted nothing needs changing. Happy to take it here if preferred.

**Correction to my own earlier text:** the round-2 testing note claimed a manual pass confirmed "normal-verse click-to-sync still works (so the guard is not too broad)." Wrong for verse mode — nothing reports there at all. What the manual pass actually covered is chapter mode, and the not-too-broad property is now pinned by the test above instead.

## Review round 4 (2026-08-19)

Two ⛏️ nitpicks from @imnasnainaec, both taken in one comment-only commit. No behavior change; 24 tests in the touched test file pass and Prettier is clean on both files.

- **ADR paragraph reflowed.** The `isReadonly` sub-bullet had a half-length line ending mid-sentence — leftover from an edit that cut the sentence's opening and never rewrapped the tail. Wrapped to the file's width end to end.
- **`lastFedUsjText`'s comments folded into its doc block** — and the doc line was indeed stale: it still promised `''` when nothing was fed, which stopped being true when the throw went in last round. The explanation of the throw is now the doc block's second paragraph, and the summary line no longer contradicts the code.

## Rebase (2026-08-22)

Rebased onto `main` three times over the course of this round; the two ⛏️ nitpicks from round 4 are in, plus one follow-up and one piece of bookkeeping.

- **`lastFedUsjText` no longer ends `?? ''`.** The throw already covers the never-fed case, so the fallback could only mask `JSON.stringify(undefined)` — which nothing feeds. Returning `undefined` there makes `.toContain`/`.not.toContain` fail loudly instead of passing vacuously, which is what the doc block promises.
- **The ADR is now ADR-0019, not ADR-0013.** `main` landed ADR-0013 through ADR-0018 while this branch was in review (`InstalledExtensions.packaged`, the analytics layer, per-web-view Ctrl+F, shadcn `Empty`, launch parameters, and the withdrawn launch token). Renumbered to the next free slot; the four citations in the grid's source comments and every reference in this description move with it. The ADR text itself is unchanged.
- **Three files this branch touches were also touched by PT-4341 and PT-4111** (`platform-scripture-editor.web-view.tsx`, `resource-text-panel.web-view.tsx`, `scripture-text-grid.component.tsx`). All auto-merged; verified the `versificationStr` comment, the Resource Viewer chapter-fetch comment, and the a11y comment plus `resolveDisplayVerseNum` call all survived intact.

**Verified on the final base:** `npm run typecheck` clean repo-wide; `npm run lint` 0 errors (4 pre-existing warnings, none in this diff); 159 tests in `scripture-text-grid/` and 21 in `book-chapter-control.navigation.test.ts` pass. Diff vs `main` is unchanged in shape — 11 files, 518 insertions.

**Comment sweep for #2586's rule.** `.claude/rules/code-quality/forward-facing-comments.md` landed after this branch was written, so every comment this branch adds or changes got the strip-the-PR-context test. Seven sites were narrating development history; a keyword grep over the added comment lines caught four that a read-through missed.

Cut: the "that stale rationale was corrected here" clause and the dated "Verified 2026-08-16" note; how a stale 0.8.14 tarball made an earlier version of that comment wrong; the `PT-3133`/`PT-4061` citations in `resource-text-panel.web-view.tsx` (tickets this PR closes — the constraint they carried now reads "single-verse surfaces resolve verse 0 to verse 1; whole-chapter surfaces like this one must not (see ADR-0019)"); "which is how a vacuous assertion slipped in here once"; "Verse 0 **no longer** reaches here"; "which two tests enter but none **previously** left"; and "more often **than before**".

Kept, because they point forward: the ADR-0019 pointers, the open PT-4379 follow-up, the PT-3495 regression an assertion guards, and the version constraint explaining why the `versificationStr` fallback exists at all. Test titles are untouched — the rule covers comments, and `PT-3133` in a `describe` reads as requirement traceability rather than development history. Say the word if you'd rather those go too.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2663)

[PT-4379]: https://paratextstudio.atlassian.net/browse/PT-4379?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.context/standards/Architecture-Decisions.md
+++ b/.context/standards/Architecture-Decisions.md
@@ -760,3 +760,66 @@ step, no automation. Just a record.
   should not be assumed to remount anything.
 - **Source:** PT-4111 `/review-paratext` code review. Withdrawn after PR #2691 review traced
   `srcNonce` to its use site.
+
+## ADR-0013: Verse 0 resolves to verse 1 on single-verse display surfaces (display-only)
+
+- **Date:** 2026-08-05
+- **Status:** Accepted
+- **Context:** A verse-0 reference means "everything preceding verse 1" — book/chapter intros,
+  titles, outlines, Psalm `\d` superscriptions. It is reachable three ways: placing the cursor in
+  pre-verse-1 editor content (`usj-reader-writer.ts` reports `verseNum: 0`); the toolbar
+  previous-verse button, which calls `getPreviousVerseRef` **without** `bounds`
+  (`book-chapter-control.navigation.ts`), so its no-versification branch floors at verse 0 in any
+  chapter, not just chapter 1; and typing a `C:0` reference. A one-verse-tall surface has no useful
+  way to render that content, so the Text Collection showed an empty cell at Luke 1:0 —
+  [PT-3133](https://paratextstudio.atlassian.net/browse/PT-3133). PT9's
+  Text Collection shows verse 1 there instead. A4/PT-4052 (PR #2509) shipped a "No text for this
+  verse" ghost-text state, contradicting PT-3133's written acceptance (display verse 1); the
+  divergence went unflagged until PT-4061.
+- **Decision:** Single-verse display surfaces resolve a verse-0 reference to verse 1 via
+  `resolveDisplayVerseNum` (`extensions/src/platform-scripture-editor/src/scripture-text-grid/verse-display.utils.ts`),
+  confirmed by Ian Hewerdine 2026-08-05. Three constraints make this safe:
+  - **Display-only, and it needs an explicit guard.** The resolved verse is never written back to
+    the scroll group — writing it back would yank the Scripture Editor off the intro the user came
+    from. This does not happen for free: `Editorial`'s `ScriptureReferencePlugin` mounts even when
+    `isReadonly` is set (gated only on `scrRef && onScrRefChange`) and reports any selection whose
+    verse differs from `scrRef`. Fall-forward makes that mismatch permanent, so `ResourceCell`
+    swallows the echo. Without the guard one click writes verse 1 — and, because the slice drops
+    chapter chrome, chapter 1, turning Luke 5:0 into Luke 1:1.
+  - **Chapter surfaces are exempt.** Anything rendering a whole chapter shows verse-0 front matter
+    directly: the Text Collection's chapter mode, its chapter-context split, its single-resource
+    path (`ScriptureTextGrid` renders one shown resource as `viewMode="chapter"`, so verse-0
+    behavior differs between one and several resources), and the Resource Viewer
+    (`resource-text-panel.web-view.tsx`). Correct behavior, signed off by Ian on PT-3133.
+  - **Genuinely-missing verses keep the ghost text.** Fall-forward applies only to verse 0. A verse
+    absent from a resource (an NT-only resource at an OT reference, an untranslated verse) still
+    renders "No text for this verse".
+  Accessible names resolve the same way, so the announcement names the verse the cells display. It
+  names the row, not its contents: a resource lacking verse 1 shows the empty state under the same
+  label, and a combined opener renders a "1-3" verse number under a label saying "1".
+- **Alternatives:** (a) *Keep the ghost-text empty state* — rejected: it blanks every cell at once
+  exactly when the user crosses a chapter or book boundary, so the tool reads as broken at the
+  moment it should be most useful; it also contradicts PT-3133's written acceptance. (b) *Normalize
+  the reference — forbid verse 0 in the Text Collection, or write verse 1 back to the scroll group*
+  (Todd Hoatson's suggestion on PT-3133) — rejected: the scroll group is shared, so it would move
+  every other view off the intro. (c) *Put the rule inside `sliceUsjToVerse`* — rejected: the
+  decision is about the **reference**, not the USJ, and the accessible-name site needs it without
+  having any USJ; burying a copy there would centralize nothing while making
+  `sliceUsjToVerse(usj, 0)` silently return verse 1.
+- **Consequences:** verse-0 content is not shown in verse view — deliberately. It stays one click
+  away via the chapter-context split, which renders the chapter unsliced; that escape hatch is what
+  makes the trade acceptable, so **revisit if the chapter-context split is ever removed or made
+  non-obvious**. Verse 0 and verse 1 now render identically and carry the same accessible name, so
+  stepping backward across that boundary produces no visible or announced change — accepted, since
+  the alternative is a row of ghost text at every chapter boundary. Note this is not one silent step
+  but a **silent dead end**: with no `bounds`, `getPreviousVerseRef` returns the same
+  `{chapterNum, verseNum: 0}` ref every time, so once the toolbar's previous-verse button lands on
+  verse 0 in a chapter > 1 it is inert, and fall-forward removes its last remaining cue. The pin
+  predates this decision; fixing it means passing `bounds` at that call site
+  (`book-chapter-control.navigation.ts`), which is out of scope here. Any future single-verse surface
+  must call `resolveDisplayVerseNum` — `sliceUsjToVerse` is deliberately mechanical and slices a raw
+  verse 0 to nothing (pinned by a test). Note the helper is currently private to
+  `platform-scripture-editor`; a surface in another extension cannot import it, so **promote it to
+  `lib/platform-bible-utils/src/scripture/` (which already owns `getPreviousVerseRef`, the producer
+  of verse-0 refs) at the second consumer** rather than re-typing the rule.
+- **Source:** PT-4061 (B3), which resolves PT-3133; Ian Hewerdine confirmed parity 2026-08-05.
--- a/.context/standards/Architecture-Decisions.md
+++ b/.context/standards/Architecture-Decisions.md
@@ -782,10 +782,27 @@ step, no automation. Just a record.
   - **Display-only, and it needs an explicit guard.** The resolved verse is never written back to
     the scroll group — writing it back would yank the Scripture Editor off the intro the user came
     from. This does not happen for free: `Editorial`'s `ScriptureReferencePlugin` mounts even when
-    `isReadonly` is set (gated only on `scrRef && onScrRefChange`) and reports any selection whose
-    verse differs from `scrRef`. Fall-forward makes that mismatch permanent, so `ResourceCell`
-    swallows the echo. Without the guard one click writes verse 1 — and, because the slice drops
-    chapter chrome, chapter 1, turning Luke 5:0 into Luke 1:1.
+    `isReadonly` is set (`Editor.tsx` gates it on `scrRef && onScrRefChange` only) and reports any
+    selection whose verse differs from `scrRef`. Fall-forward makes that mismatch permanent, so
+    `ResourceCell` swallows the echo. Without the guard one click writes verse 1 — and, because the
+    slice drops chapter chrome, `$findAndSetChapterAndVerse`'s
+    `parseInt(chapterNode?.getNumber() ?? '1', 10)` yields chapter 1, turning Luke 5:0 into
+    Luke 1:1.
+
+    **The guard belongs in the consumer, not upstream in the plugin** (raised in review of #2663;
+    verified against `@eten-tech-foundation/platform-editor` ~0.8.14). Gating the plugin on
+    `isReadonly` was considered and is rejected on the merits, not merely deferred: the plugin is
+    **bidirectional** — `$moveCursorToVerseStart` applies `scrRef` to the caret, and
+    `$findAndSetChapterAndVerse` reports the caret back — and read-only surfaces need both halves.
+    Not mounting it when read-only would break navigation-to-verse in every read-only editor (the
+    Resource Viewer, this grid's chapter mode) and would break read-only click-to-sync, which
+    **these very cells rely on at every normal verse**. `isReadonly` is therefore the wrong
+    predicate: a read-only editor reporting its caret position is correct behavior, not the bug.
+    The bug is narrower and entirely host-made — *we* hand the editor verse 1's USJ while telling it
+    verse 0, so the only component that knows the reference is a deliberate lie is the one that told
+    it. Whoever creates the mismatch owns swallowing it. The generalizable fix, if a third surface
+    ever wants one, is not an `isReadonly` gate but for the promoted helper to carry the guard with
+    it (see Consequences) — the rule and its guard are one unit.
   - **Chapter surfaces are exempt.** Anything rendering a whole chapter shows verse-0 front matter
     directly: the Text Collection's chapter mode, its chapter-context split, its single-resource
     path (`ScriptureTextGrid` renders one shown resource as `viewMode="chapter"`, so verse-0
@@ -805,21 +822,57 @@ step, no automation. Just a record.
   every other view off the intro. (c) *Put the rule inside `sliceUsjToVerse`* — rejected: the
   decision is about the **reference**, not the USJ, and the accessible-name site needs it without
   having any USJ; burying a copy there would centralize nothing while making
-  `sliceUsjToVerse(usj, 0)` silently return verse 1.
+  `sliceUsjToVerse(usj, 0)` silently return verse 1. (d) *Fall forward only when verse 0 has no
+  content of its own* — a content-aware rule (raised in review of #2663). The distinction it draws is
+  real in the source: a `\d` superscription is one short line that a one-verse-tall cell renders
+  fine, whereas book/chapter intros and outlines are what genuinely cannot be rendered there. Under
+  it, PT9 parity would hold where it actually matters (the intro case), `<` from Psalms verse 1 would
+  produce a visible change instead of a silent one, and no real Scripture would be hidden in verse
+  view. Rejected on cost and consistency: the helper would need the chapter USJ, and the
+  accessible-name site in `ScriptureTextGrid` has none — it names the row before any resource has
+  fetched anything — which is exactly why the rule is reference-only. Making it content-aware would
+  either force a USJ through the naming path or split the rule in two (a USJ-aware version for cells,
+  a reference-only version for labels), and per-resource content-awareness would make the *same*
+  reference fall forward in one cell and not in its neighbor, so a row of cells would disagree about
+  which verse they are showing. Worth revisiting if the superscription case draws real complaints,
+  since it is the one place this decision hides genuine content.
 - **Consequences:** verse-0 content is not shown in verse view — deliberately. It stays one click
   away via the chapter-context split, which renders the chapter unsliced; that escape hatch is what
   makes the trade acceptable, so **revisit if the chapter-context split is ever removed or made
   non-obvious**. Verse 0 and verse 1 now render identically and carry the same accessible name, so
-  stepping backward across that boundary produces no visible or announced change — accepted, since
-  the alternative is a row of ghost text at every chapter boundary. Note this is not one silent step
-  but a **silent dead end**: with no `bounds`, `getPreviousVerseRef` returns the same
-  `{chapterNum, verseNum: 0}` ref every time, so once the toolbar's previous-verse button lands on
-  verse 0 in a chapter > 1 it is inert, and fall-forward removes its last remaining cue. The pin
-  predates this decision; fixing it means passing `bounds` at that call site
-  (`book-chapter-control.navigation.ts`), which is out of scope here. Any future single-verse surface
-  must call `resolveDisplayVerseNum` — `sliceUsjToVerse` is deliberately mechanical and slices a raw
-  verse 0 to nothing (pinned by a test). Note the helper is currently private to
-  `platform-scripture-editor`; a surface in another extension cannot import it, so **promote it to
+  stepping backward across that boundary produces no visible or announced change **within the grid**
+  — accepted, since the alternative is a row of ghost text at every chapter boundary.
+
+  **Verse 0 becomes unobservable inside the Text Collection, including to assistive tech** (raised
+  in review of #2663). At a verse-0 reference the row label announces "MAT 5:1", the cells render
+  verse 1, and nothing in the grid states that the shared reference is actually 5:0 — so a
+  screen-reader user has no in-grid signal that they are at the boundary. This is a genuine accepted
+  cost of naming the row after what it displays, not an oversight: the alternative is a label that
+  announces a verse the cells demonstrably do not show, which is worse. Two things keep it from
+  being a true silent dead end, and both live outside the grid: the BCV control still displays the
+  real reference (`5:0`), and its **previous-verse button is disabled** at verse 0, which is an
+  announced state change rather than an inert control — `isNoOpNavigation` in
+  `book-chapter-control.navigation.ts` disables a step that would return the same ref, pinned by the
+  `'Is disabled when at verse 0'` test in `book-chapter-control.navigation.test.ts`. (An earlier
+  draft of this ADR called the button "inert" and claimed fall-forward removed its "last remaining
+  cue"; that was wrong on both counts and is corrected here.) What the button cannot do is roll
+  backward to the previous chapter's last verse: with no `bounds`, `getPreviousVerseRef` floors at
+  `{chapterNum, verseNum: 0}` in every chapter rather than only chapter 1, so the reference dead-ends
+  there. That pin predates this decision and is now tracked as
+  [PT-4379](https://paratextstudio.atlassian.net/browse/PT-4379) — out of scope here because the fix
+  is not passing an argument but threading an async, PAPI-built `ScriptureBounds` through a
+  `platform-bible-react` component into every consumer of the BCV control.
+
+  Any future single-verse surface must call `resolveDisplayVerseNum` — `sliceUsjToVerse` is
+  deliberately mechanical and slices a raw verse 0 to nothing (pinned by a test) — and must carry the
+  write-back guard with it. The helper is currently private to `platform-scripture-editor`, so a
+  surface in another extension cannot import it; **promote it to
   `lib/platform-bible-utils/src/scripture/` (which already owns `getPreviousVerseRef`, the producer
-  of verse-0 refs) at the second consumer** rather than re-typing the rule.
+  of verse-0 refs) at the second consumer _of this rule_** rather than re-typing it. The qualifier is
+  load-bearing: "second consumer" counted as surfaces would be wrong, because other surfaces already
+  handle verse-0 references under deliberately different rules — review of #2663 cites the
+  interlinearizer extension, which treats verse 0 as tokenizable content to be kept rather than
+  routed around, because it renders a whole chapter. A whole-chapter surface should look like the
+  chapter-surface exemption above, not like this rule. Promoting on a surface count would turn a
+  deliberate single-verse-vs-whole-chapter difference into apparent drift from a shared util.
 - **Source:** PT-4061 (B3), which resolves PT-3133; Ian Hewerdine confirmed parity 2026-08-05.
--- a/.context/standards/Architecture-Decisions.md
+++ b/.context/standards/Architecture-Decisions.md
@@ -782,21 +782,36 @@ step, no automation. Just a record.
   - **Display-only, and it needs an explicit guard.** The resolved verse is never written back to
     the scroll group — writing it back would yank the Scripture Editor off the intro the user came
     from. This does not happen for free: `Editorial`'s `ScriptureReferencePlugin` mounts even when
-    `isReadonly` is set (`Editor.tsx` gates it on `scrRef && onScrRefChange` only) and reports any
-    selection whose verse differs from `scrRef`. Fall-forward makes that mismatch permanent, so
-    `ResourceCell` swallows the echo. Without the guard one click writes verse 1 — and, because the
-    slice drops chapter chrome, `$findAndSetChapterAndVerse`'s
-    `parseInt(chapterNode?.getNumber() ?? '1', 10)` yields chapter 1, turning Luke 5:0 into
-    Luke 1:1.
+    `isReadonly` is set (`Editor.tsx` gates it on `scrRef && onScrRefChange` only) and
+    `$findAndSetChapterAndVerse` reports any selection whose **chapter or verse** disagrees with
+    `scrRef` — a chapter mismatch when the document has a chapter node, or a `scrRef` verse outside
+    the selected verse marker's range. Fall-forward makes that disagreement permanent
+    (`isVerseInRange(0, '1')` is false on every click), so `ResourceCell` swallows the echo. Without
+    the guard one click writes verse 1 — and, because the slice drops chapter chrome,
+    `parseInt(chapterNode?.getNumber() ?? '1', 10)` yields chapter 1, turning Luke 5:0 into Luke 1:1.
 
-    **The guard belongs in the consumer, not upstream in the plugin** (raised in review of #2663;
-    verified against `@eten-tech-foundation/platform-editor` ~0.8.14). Gating the plugin on
+    **Version-pinned:** traced against `@eten-tech-foundation/platform-editor` **0.8.14**, which is
+    what `package-lock.json` resolves and npm's current `latest` (as of 2026-08-16; 0.8.15 is not
+    published). This matters because upstream `main` in `eten-tech-foundation/scripture-editors` has
+    since rewritten the plugin around a `$resolvePosition` / `onSelectionSettled` / `report` machine
+    whose `$resolvePosition` returns `undefined` for a document with no `BookNode` and no
+    `ChapterNode` (pinned upstream as invariant I5). A verse-mode slice is exactly that document, so
+    **on the release that lands, the plugin will report nothing here and this guard becomes
+    defense-in-depth** rather than a live fix. Neither the guard nor this ADR changes at that point;
+    re-verify at the next version bump. (Raised in review of #2663 against an unreleased build.)
+
+    **The guard belongs in the consumer, not upstream in the plugin.** Gating the plugin on
     `isReadonly` was considered and is rejected on the merits, not merely deferred: the plugin is
     **bidirectional** — `$moveCursorToVerseStart` applies `scrRef` to the caret, and
-    `$findAndSetChapterAndVerse` reports the caret back — and read-only surfaces need both halves.
-    Not mounting it when read-only would break navigation-to-verse in every read-only editor (the
-    Resource Viewer, this grid's chapter mode) and would break read-only click-to-sync, which
-    **these very cells rely on at every normal verse**. `isReadonly` is therefore the wrong
+    `$findAndSetChapterAndVerse` reports the caret back (`$moveCaretToVerseStart` and
+    `$resolvePosition` after the upstream rewrite) — and read-only surfaces need both halves. Not
+    mounting it when read-only would break navigation-to-verse in every read-only editor and would
+    break read-only click-to-sync, which **this grid's chapter mode and the Resource Viewer rely on**.
+    (Scoped deliberately: click-to-sync is *not* load-bearing in verse mode. There the slice holds
+    exactly the verse `scrRef` names, so a click resolves to that same verse, the plugin finds no
+    disagreement, and nothing is reported — a verse cell has nowhere to sync TO. That predates this
+    decision: `sliceUsjToVerse` has dropped chapter chrome since #2509. It is not a lost capability
+    and is deliberately not filed.) `isReadonly` is therefore the wrong
     predicate: a read-only editor reporting its caret position is correct behavior, not the bug.
     The bug is narrower and entirely host-made — *we* hand the editor verse 1's USJ while telling it
     verse 0, so the only component that knows the reference is a deliberate lie is the one that told
@@ -870,9 +885,14 @@ step, no automation. Just a record.
   `lib/platform-bible-utils/src/scripture/` (which already owns `getPreviousVerseRef`, the producer
   of verse-0 refs) at the second consumer _of this rule_** rather than re-typing it. The qualifier is
   load-bearing: "second consumer" counted as surfaces would be wrong, because other surfaces already
-  handle verse-0 references under deliberately different rules — review of #2663 cites the
-  interlinearizer extension, which treats verse 0 as tokenizable content to be kept rather than
-  routed around, because it renders a whole chapter. A whole-chapter surface should look like the
-  chapter-surface exemption above, not like this rule. Promoting on a surface count would turn a
-  deliberate single-verse-vs-whole-chapter difference into apparent drift from a shared util.
+  handle verse-0 references under deliberately different rules. The worked example is the
+  interlinearizer extension (`sillsdev/interlinearizer-extension`, `InterlinearizerLoader.tsx`, the
+  `activeScrRef` memo): it renders a whole chapter and treats verse 0 as tokenizable content, so it
+  **keeps** verse 0 whenever a segment covers it — its USJ extractor emits a synthetic verse-0 scope
+  with SID `"<book> <chapter>:0"` for pre-verse-1 content — and falls back to verse 1 only when the
+  loaded book has no verse-0 segment for that chapter. That is content-awareness this rule
+  deliberately rejects (alternative (d) above), affordable there precisely because the whole book is
+  already parsed into segments before a reference is resolved. A whole-chapter surface should look
+  like the chapter-surface exemption above, not like this rule. Promoting on a surface count would
+  turn a deliberate single-verse-vs-whole-chapter difference into apparent drift from a shared util.
 - **Source:** PT-4061 (B3), which resolves PT-3133; Ian Hewerdine confirmed parity 2026-08-05.
--- a/.context/standards/Architecture-Decisions.md
+++ b/.context/standards/Architecture-Decisions.md
@@ -790,15 +790,23 @@ step, no automation. Just a record.
     the guard one click writes verse 1 — and, because the slice drops chapter chrome,
     `parseInt(chapterNode?.getNumber() ?? '1', 10)` yields chapter 1, turning Luke 5:0 into Luke 1:1.
 
-    **Version-pinned:** traced against `@eten-tech-foundation/platform-editor` **0.8.14**, which is
-    what `package-lock.json` resolves and npm's current `latest` (as of 2026-08-16; 0.8.15 is not
-    published). This matters because upstream `main` in `eten-tech-foundation/scripture-editors` has
-    since rewritten the plugin around a `$resolvePosition` / `onSelectionSettled` / `report` machine
-    whose `$resolvePosition` returns `undefined` for a document with no `BookNode` and no
-    `ChapterNode` (pinned upstream as invariant I5). A verse-mode slice is exactly that document, so
-    **on the release that lands, the plugin will report nothing here and this guard becomes
-    defense-in-depth** rather than a live fix. Neither the guard nor this ADR changes at that point;
-    re-verify at the next version bump. (Raised in review of #2663 against an unreleased build.)
+    **Which editor build this describes (as of 2026-08-16).** The trace above is against
+    `@eten-tech-foundation/platform-editor` **0.8.14** — what `package-lock.json` resolves, what npm
+    serves as `latest`, and what is installed locally. But 0.8.14 was published 2026-04-14 and this
+    repo's `main` already imports editor APIs it does not export (`PARAGRAPH_STRUCTURE_VIEW_MODE`,
+    `StructureProtectionMode`, `ViewOptions.hasGutterParaMarkers`, `EditorRef.removeCharacterMarker`),
+    so **the effective target is a newer build than the lockfile pins** and the lockfile should not be
+    read as authoritative here.
+
+    In that newer build the plugin is rewritten around a `$resolvePosition` / `onSelectionSettled` /
+    `report` machine whose `$resolvePosition` returns `undefined` for a document with no `BookNode`
+    and no `ChapterNode` (upstream invariant I5). A verse-mode slice is exactly that document, so
+    there the plugin reports **nothing** here. Both states are recorded deliberately, because the
+    guard is right under either: a live fix against 0.8.14's `$findAndSetChapterAndVerse`, and
+    defense-in-depth against the rewrite — the thing that keeps the write-back closed if a future
+    editor makes chrome-free slices addressable again. Re-verify at the next version bump; do not
+    delete the guard on the strength of the rewrite alone. (Both mechanisms surfaced in review of
+    #2663.)
 
     **The guard belongs in the consumer, not upstream in the plugin.** Gating the plugin on
     `isReadonly` was considered and is rejected on the merits, not merely deferred: the plugin is
--- a/.context/standards/Architecture-Decisions.md
+++ b/.context/standards/Architecture-Decisions.md
@@ -761,7 +761,7 @@ step, no automation. Just a record.
 - **Source:** PT-4111 `/review-paratext` code review. Withdrawn after PR #2691 review traced
   `srcNonce` to its use site.
 
-## ADR-0013: Verse 0 resolves to verse 1 on single-verse display surfaces (display-only)
+## ADR-0019: Verse 0 resolves to verse 1 on single-verse display surfaces (display-only)
 
 - **Date:** 2026-08-05
 - **Status:** Accepted
--- a/.context/standards/Architecture-Decisions.md
+++ b/.context/standards/Architecture-Decisions.md
@@ -813,13 +813,13 @@ step, no automation. Just a record.
     where the slice is unaddressable and no click reports anything. That predates this decision —
     `sliceUsjToVerse` has dropped chapter chrome since #2509 — and is not a lost capability, since a
     verse cell holds only the verse you are already on and so has nowhere to sync TO. Deliberately
-    not filed.) `isReadonly` is therefore the wrong
-    predicate: a read-only editor reporting its caret position is correct behavior, not the bug.
-    The bug is narrower and entirely host-made — *we* hand the editor verse 1's USJ while telling it
-    verse 0, so the only component that knows the reference is a deliberate lie is the one that told
-    it. Whoever creates the mismatch owns swallowing it. The generalizable fix, if a third surface
-    ever wants one, is not an `isReadonly` gate but for the promoted helper to carry the guard with
-    it (see Consequences) — the rule and its guard are one unit.
+    not filed.) `isReadonly` is therefore the wrong predicate: a read-only editor reporting its
+    caret position is correct behavior, not the bug. The bug is narrower and entirely host-made —
+    *we* hand the editor verse 1's USJ while telling it verse 0, so the only component that knows
+    the reference is a deliberate lie is the one that told it. Whoever creates the mismatch owns
+    swallowing it. The generalizable fix, if a third surface ever wants one, is not an `isReadonly`
+    gate but for the promoted helper to carry the guard with it (see Consequences) — the rule and
+    its guard are one unit.
   - **Chapter surfaces are exempt.** Anything rendering a whole chapter shows verse-0 front matter
     directly: the Text Collection's chapter mode, its chapter-context split, its single-resource
     path (`ScriptureTextGrid` renders one shown resource as `viewMode="chapter"`, so verse-0
--- a/.context/standards/Architecture-Decisions.md
+++ b/.context/standards/Architecture-Decisions.md
@@ -787,15 +787,20 @@ step, no automation. Just a record.
 
     **The guard is defense-in-depth, not a fix for a live report.** `$resolvePosition` refuses to
     describe a position in a document with no `BookNode` and no `ChapterNode` (upstream invariant I5),
-    and `sliceUsjToVerse` drops both — so today the plugin is silent in verse mode, and the
+    and `sliceUsjToVerse` drops both — so the plugin is silent in verse mode, and the
     `viewMode === 'verse'` branch of `handleScrRefChange` is unreachable. It is kept because it costs
     nothing and is the right shape if a future editor makes slices addressable; do not read it as
-    evidence that a write-back currently occurs. Verified against the editor this repo actually
-    builds: `dev-packages/scripture-editors` `packages/platform` at **0.8.15**, which `postinstall` →
-    `link-dev-packages` builds and yalc-links over `node_modules`. (The npm-published 0.8.14 that
-    `package-lock.json` names is a placeholder the link replaces; an earlier draft of this ADR
-    described 0.8.14's `$findAndSetChapterAndVerse` and its chapter-1 fallback as the live mechanism,
-    which was wrong — that plugin is not what runs here. Corrected in review of #2663.)
+    evidence that a write-back currently occurs.
+
+    Verified 2026-08-16 against `@eten-tech-foundation/platform-editor` **0.8.15**, in both places it
+    can be read: the published npm package, and `dev-packages/scripture-editors` `packages/platform`,
+    which `postinstall` → `link-dev-packages` builds and yalc-links over `node_modules`. They agree
+    on this mechanism (the vendored copy trails published 0.8.15 by one caret-placement line in
+    `$moveCaretToVerseStart`). **Verify against the linked build, not `package-lock.json`** — the lock
+    still named 0.8.14 when this was written, and reading that stale tarball is exactly how an earlier
+    draft of this ADR came to describe `$findAndSetChapterAndVerse` and its chapter-1 fallback as the
+    live mechanism. That was wrong; that plugin does not exist in 0.8.15. Corrected in review of
+    #2663.
 
     **The guard belongs in the consumer, not upstream in the plugin.** Gating the plugin on
     `isReadonly` was considered and is rejected on the merits, not merely deferred: the plugin is
--- a/.context/standards/Architecture-Decisions.md
+++ b/.context/standards/Architecture-Decisions.md
@@ -779,47 +779,36 @@ step, no automation. Just a record.
 - **Decision:** Single-verse display surfaces resolve a verse-0 reference to verse 1 via
   `resolveDisplayVerseNum` (`extensions/src/platform-scripture-editor/src/scripture-text-grid/verse-display.utils.ts`),
   confirmed by Ian Hewerdine 2026-08-05. Three constraints make this safe:
-  - **Display-only, and it needs an explicit guard.** The resolved verse is never written back to
+  - **Display-only, and it carries an explicit guard.** The resolved verse is never written back to
     the scroll group — writing it back would yank the Scripture Editor off the intro the user came
-    from. This does not happen for free: `Editorial`'s `ScriptureReferencePlugin` mounts even when
-    `isReadonly` is set (`Editor.tsx` gates it on `scrRef && onScrRefChange` only) and
-    `$findAndSetChapterAndVerse` reports any selection whose **chapter or verse** disagrees with
-    `scrRef` — a chapter mismatch when the document has a chapter node, or a `scrRef` verse outside
-    the selected verse marker's range. Fall-forward makes that disagreement permanent
-    (`isVerseInRange(0, '1')` is false on every click), so `ResourceCell` swallows the echo. Without
-    the guard one click writes verse 1 — and, because the slice drops chapter chrome,
-    `parseInt(chapterNode?.getNumber() ?? '1', 10)` yields chapter 1, turning Luke 5:0 into Luke 1:1.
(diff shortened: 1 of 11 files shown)


Sebastian-ubs:
> Paratext 9 shows verse 1 there

Confirmed.

> Fall-forward is display-only: the scroll group stays at verse 0, so the Scripture Editor is not pulled off the intro the user clicked into to get here.

Yes, that is the expected behavior

imnasnainaec (commented):
Five inline comments. None blocks a merge on its own. The two closest to blocking are: the request for an issue link on the `bounds` dead end; and the question of whether the write-back guard belongs in `ScriptureReferencePlugin` rather than in each consumer that will need it.

Where these came from: they surfaced while comparing this PR's verse-0 model against the interlinearizer extension's, which treats verse 0 as tokenizable content on a whole-chapter surface. That comparison is the origin of comments 1 and 5 — the two codebases have made deliberately different choices about verse 0, and the ADR is the right place for that to be on the record.

Disclosure: the inline comments were drafted by Claude and partially reviewed by me.

extensions/src/platform-scripture-editor/src/scripture-text-grid/verse-display.utils.ts
```diff
@@ -1,5 +1,20 @@
...
+ * 1, everything else shows itself.
+ *
+ * Verse 0 is everything preceding verse 1 (intros, titles, Psalm superscriptions), which a
+ * one-verse-tall cell cannot render usefully — so Paratext 9 shows verse 1 there and PT-3133 adopts
+ * that. Display-only: callers must not write the resolved verse back to the scroll group.
+ *
+ * Chapter surfaces must NOT call this — they show verse-0 front matter directly. Full rationale and
+ * rejected alternatives: ADR-0013.
+ */
+export function resolveDisplayVerseNum(verseNum: number): number {
```
imnasnainaec, line 21: The alternatives in ADR-0013 are all-or-nothing: keep the ghost text, normalize the reference, or move the rule into the slicer. There's a fourth the ADR doesn't consider — fall forward only when verse 0 has no content of its own.

That distinction is real in the source: a `\d` superscription is one short line that fits a one-verse-tall cell fine, while book/chapter intros and outlines are what genuinely can't be rendered here. Under a content-aware rule, PT9 parity holds where it actually matters (the intro case), `<` from verse 1 in Psalms produces a visible change instead of a silent one, and we stop hiding real scripture in verse view.

The cost is not small and may well sink it: the helper would need the USJ, and the accessible-name site has none — which is precisely why the ADR keeps the rule reference-only. So I'm not asking for the change. I am asking for it in the rejected-alternatives list with that reasoning, because it's the first thing a reader will think of and the ADR currently reads as though it wasn't considered.

extensions/src/platform-scripture-editor/src/scripture-text-grid/resource-cell.component.tsx
```diff
@@ -155,13 +156,33 @@ export function ResourceCell({
...
+
+  // Editorial's reference plugin is active even when read-only, and it reports any selection whose
+  // verse doesn't match `scrRef`. Under fall-forward that mismatch is permanent — we hand it verse
+  // 1's text while telling it verse 0 — so a single click would push verse 1 into the shared scroll
+  // group, dragging the Scripture Editor off the intro the user came from. Worse, the slice carries
+  // no chapter marker, so the plugin defaults to chapter 1 and Luke 5:0 would report Luke 1:1.
+  // Swallow that echo: fall-forward is display-only. Non-fallen-forward cells report normally.
+  const handleScrRefChange = useCallback(
+    (nextScrRef: SerializedVerseRef) => {
+      if (viewMode === 'verse' && isFallenForward) return;
```
imnasnainaec, line 190: This guard is correct and the defect it prevents is real. My question is whether it's in the right place.

The root cause is that `ScriptureReferencePlugin` mounts on a read-only editor and reports selections — `Editor.tsx` gates it on `scrRef && onScrRefChange` only. Fixing it here means every future single-verse surface has to independently rediscover this and write the same guard, and the ADR already anticipates a second consumer.

Was gating the plugin on `isReadonly` considered and rejected, or just not reached? If it's out of scope for this PR that's fine, but it should be said out loud somewhere, because right now the ADR presents a consumer-side guard as the design rather than as a workaround.

extensions/src/platform-scripture-editor/src/scripture-text-grid/scripture-text-grid.component.tsx
```diff
@@ -326,8 +327,10 @@ export function ScriptureTextGrid({
   // control (ArrowUp/ArrowDown) — the grip stops click/key propagation so reordering never triggers
   // the listitem's chapter-context activation. The Scripture reference is constant across rows in a
   // render, so format it once for the accessible name (`"{name}, {ref}"`, or just the name when no
-  // template is provided).
-  const reference = formatScrRef(scrRef);
+  // template is provided). Resolved so the announcement names the verse the cells actually display:
+  // at a verse-0 reference they show verse 1. This names the row, not its contents — a resource that
+  // lacks verse 1 shows the empty state under the same label.
+  const reference = formatScrRef({ ...scrRef, verseNum: resolveDisplayVerseNum(scrRef.verseNum) });
```
imnasnainaec, line 342: Agreed the label should match what the cells render — announcing "5:0" over verse-1 text would be worse. But note what this leaves a screen-reader user with at verse 0: the label says "MAT 5:1", the cells show verse 1, and nothing anywhere indicates the shared reference is actually 5:0.

Combined with the previous-verse button being inert at verse 0, `<` becomes both silently non-functional and silently mislabeled — the user gets no signal at all that they've hit the boundary.

The ADR frames this as "names the row, not its contents," which covers the resource-lacking-verse-1 case but not this one. Worth a sentence acknowledging that verse 0 is now unobservable to assistive tech, so that it reads as an accepted cost rather than an oversight.

.context/standards/Architecture-Decisions.md
```diff
@@ -415,3 +415,65 @@ step, no automation. Just a record.
...
+  away via the chapter-context split, which renders the chapter unsliced; that escape hatch is what
+  makes the trade acceptable, so **revisit if the chapter-context split is ever removed or made
+  non-obvious**. Verse 0 and verse 1 now render identically and carry the same accessible name, so
+  stepping backward across that boundary produces no visible or announced change — accepted, since
+  the alternative is a row of ghost text at every chapter boundary. Note this is not one silent step
+  but a **silent dead end**: with no `bounds`, `getPreviousVerseRef` returns the same
+  `{chapterNum, verseNum: 0}` ref every time, so once the toolbar's previous-verse button lands on
+  verse 0 in a chapter > 1 it is inert, and fall-forward removes its last remaining cue. The pin
+  predates this decision; fixing it means passing `bounds` at that call site
+  (`book-chapter-control.navigation.ts`), which is out of scope here. Any future single-verse surface
```
imnasnainaec: This paragraph identifies a real upstream bug — `book-chapter-control.navigation.ts` calls `getPreviousVerseRef` without `bounds`, so the no-versification branch floors at `{chapterNum, verseNum: 0}` in every chapter and returns the same ref forever — and then correctly scopes it out.

The problem is that this PR makes that bug *harder to find*: the ADR says so itself, "fall-forward removes its last remaining cue."

Out of scope is fine. Unfiled is not, when the same change removes the symptom that would have led someone to it. Can we get an issue and link it here?

.context/standards/Architecture-Decisions.md
```diff
@@ -415,3 +415,65 @@ step, no automation. Just a record.
...
+  must call `resolveDisplayVerseNum` — `sliceUsjToVerse` is deliberately mechanical and slices a raw
+  verse 0 to nothing (pinned by a test). Note the helper is currently private to
+  `platform-scripture-editor`; a surface in another extension cannot import it, so **promote it to
+  `lib/platform-bible-utils/src/scripture/` (which already owns `getPreviousVerseRef`, the producer
+  of verse-0 refs) at the second consumer** rather than re-typing the rule.
```
imnasnainaec: Worth tightening to "the second consumer of this rule."

There's already another surface handling verse-0 references — the interlinearizer extension — and it deliberately uses a different rule: it keeps verse 0 when the chapter has real verse-0 content and falls forward only when it doesn't, because verse 0 is tokenizable content there rather than something to route around.

As written, someone counting surfaces rather than rules will promote this into `platform-bible-utils` and the divergence will start looking like drift from a shared util instead of a deliberate difference between a single-verse and a whole-chapter surface.

.context/standards/Architecture-Decisions.md
katherinejensen00: Filed as **PT-4379**, linked from the ADR. Agreed on the principle — scoping out is fine, leaving it unfiled isn't.

Two findings while writing it up. First, the premise we were both working from was partly mine and partly wrong: the ADR's "removes its last remaining cue" line was inaccurate, since the button is disabled at verse 0 with a test pinning it. The dead end is real — you can't reach the previous chapter's last verse — but it isn't cue-less, and the ADR now says that correctly.

Second, the comment at that call site claimed this was already tracked in PT-4143. It isn't — PT-4143 is about global keyboard shortcuts clobbering context-sensitive ones. So it read as filed while being unfiled, which is worse than being obviously unfiled. That citation now points at PT-4379.

On why it stays out of scope: it isn't passing an argument. `ScriptureBounds` is built async over PAPI by `getScriptureBounds`, and `BookChapterControl` lives in `platform-bible-react` with no PAPI access — so it has to become a prop threaded through every consumer of the BCV control, none of which fetch versification today. That's in the ticket.

.context/standards/Architecture-Decisions.md
katherinejensen00: Good catch — the failure mode you describe (a deliberate difference reading as drift from a shared util) is exactly what would have happened. Tightened to "the second consumer **of this rule**," with a note that the qualifier is load-bearing and why: whole-chapter surfaces belong with the chapter-surface exemption above, not with this rule, so promoting on a surface count would be the wrong trigger.

One transparency note: I couldn't locate the interlinearizer in any of the repos I have checked out locally, so I've attributed the example to your review rather than citing it to a file — I didn't want to assert implementation details I hadn't read. If you can point me at the repo I'll add a concrete citation, which would make the contrast much more useful to the next reader than my paraphrase.

extensions/src/platform-scripture-editor/src/scripture-text-grid/resource-cell.component.tsx
katherinejensen00: Fair challenge — it was reached, but only in my head, which is the same as not reached. Now written out in ADR-0013, and I went back to the editor source to make sure the answer was right rather than convenient.

It's rejected on the merits. The plugin is bidirectional: `$moveCursorToVerseStart` applies `scrRef` to the caret and `$findAndSetChapterAndVerse` reports it back, and read-only surfaces need both halves. An `isReadonly` gate would break navigation-to-verse in every read-only editor (Resource Viewer, this grid's chapter mode) **and** break read-only click-to-sync — which these very cells depend on at every normal verse. So `isReadonly` is the wrong predicate; a read-only editor reporting its caret is correct behavior, not the bug.

The bug is narrower and entirely ours: we hand the editor verse 1's USJ while telling it verse 0. The only component that knows the reference is a deliberate lie is the one that told it. On your real concern — rediscovery — I've made the guard travel with the rule: the promotion note now says the helper and its guard move together as one unit, so consumer three inherits both instead of rederiving the guard. Verified against `platform-editor` ~0.8.14.

extensions/src/platform-scripture-editor/src/scripture-text-grid/scripture-text-grid.component.tsx
katherinejensen00: Agreed and recorded — in ADR-0013, at the naming site, and as trade-off 3 in the PR description, stated as an accepted cost rather than left implied.

One correction while checking this, and it's against my own earlier text rather than yours: the previous-verse button isn't inert at verse 0 — it's **disabled**. `isNoOpNavigation` disables any step that would return the same ref, and there's a test pinning it (`'Is disabled when at verse 0'`). So the user does get a boundary signal, and a screen reader announces the disabled state; the BCV control also still displays the real `5:0`. The ADR previously claimed fall-forward removed the button's "last remaining cue" — that was wrong, and it now says so explicitly.

That narrows it accurately: the loss is confined to _inside the grid_, where nothing reveals the real reference. The cues that survive live outside it. What the button genuinely cannot do is roll back to the previous chapter — that's the real gap, now PT-4379.

extensions/src/platform-scripture-editor/src/scripture-text-grid/verse-display.utils.ts
katherinejensen00: You're right that it reads as unconsidered, and it's the strongest of the alternatives. Added as **(d)** with your framing intact — including that it would hold parity where it actually matters and stop hiding real Scripture.

I recorded two costs beyond the one you named. Splitting the rule is the obvious one: the naming site would need either a USJ threaded into it or a second reference-only variant. The one that decided it for me is that content-awareness is inherently _per-resource_ — the same reference would fall forward in one cell and not in its neighbor, so a row of cells would disagree about which verse they're showing, under a single shared label. That seemed worse than the trade we're making.

I also noted it's worth revisiting if the superscription case draws real complaints, since that's the one place this decision hides genuine content.

katherinejensen00 (commented):
Thank you for the review and the helpful suggestions!

> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2663#-:-OzxM0Gp6jJb_iyaxLrx:bgzd6ac)* status: 0 of 10 files reviewed, 5 unresolved discussions (waiting on imnasnainaec).

.context/standards/Architecture-Decisions.md
imnasnainaec: https://github.com/sillsdev/interlinearizer-extension

imnasnainaec (commented):
> @imnasnainaec reviewed 3 files and all commit messages, made 2 comments, and resolved 5 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2663)* status: 3 of 10 files reviewed, 1 unresolved discussion (waiting on katherinejensen00).

___
*[`lib/platform-bible-react/src/components/advanced/book-chapter-control/book-chapter-control.navigation.ts` line 42 at r2](https://reviewable.io/reviews/paranext/paranext-core/2663#-P--JaaV3Rz89TCnGryx:-P--JaaV3Rz89TCnGryy:b-cfr6dm) ([raw file](https://github.com/paranext/paranext-core/blob/075c81cc82c932bfd75cc0066755fdf2f0d1f0b3/lib/platform-bible-react/src/components/advanced/book-chapter-control/book-chapter-control.navigation.ts#L42)):*
> ```XML
>   // is no versification source here, so verse navigation keeps its pre-versification
>   // floor-at-0 / unbounded-increment behavior within a book. Making the buttons versification-aware
>   // to match the keyboard commands is tracked in PT-4379. (This previously cited PT-4143, which is
> ```

:pick: This added parenthetical is justifying the diff, but doesn't really belong in the final product.

.context/standards/Architecture-Decisions.md
```diff
@@ -415,3 +415,119 @@ step, no automation. Just a record.
...
+- **Decision:** Single-verse display surfaces resolve a verse-0 reference to verse 1 via
+  `resolveDisplayVerseNum` (`extensions/src/platform-scripture-editor/src/scripture-text-grid/verse-display.utils.ts`),
+  confirmed by Ian Hewerdine 2026-08-05. Three constraints make this safe:
+  - **Display-only, and it needs an explicit guard.** The resolved verse is never written back to
+    the scroll group — writing it back would yank the Scripture Editor off the intro the user came
+    from. This does not happen for free: `Editorial`'s `ScriptureReferencePlugin` mounts even when
+    `isReadonly` is set (`Editor.tsx` gates it on `scrRef && onScrRefChange` only) and reports any
+    selection whose verse differs from `scrRef`. Fall-forward makes that mismatch permanent, so
+    `ResourceCell` swallows the echo. Without the guard one click writes verse 1 — and, because the
+    slice drops chapter chrome, `$findAndSetChapterAndVerse`'s
```
imnasnainaec: **The cited mechanism is from a different editor package**

`$findAndSetChapterAndVerse` and the `parseInt(chapterNode?.getNumber() ?? '1', 10)` chapter-1 fallback do not exist in `@eten-tech-foundation/platform-editor`. They live in `packages/scribe/src/editor/plugins/ScriptureReferencePlugin.tsx` in `eten-tech-foundation/scripture-editors` — the Scribe editor. This package publishes from `packages/platform`, whose plugin was rewritten around a `$resolvePosition` / `onSelectionSettled` / `report` machine. (A code search for `findAndSetChapterAndVerse` across that repo returns exactly one hit, in scribe.)

The rewritten plugin does not fall back to chapter 1 for a chrome-free slice — it declines to report at all:

```ts
const bookNode = $getFirstBookNode();
const chapterNode = $findThisChapter(startNode);
if (!chapterNode && !bookNode) return undefined;   // unaddressable
const chapterNum = chapterNode ? parseInt(chapterNode.getNumber() ?? "1", 10) : 1;
```

`onSelectionSettled` then returns at `if (!position) return;`. Since `sliceUsjToVerse` drops both `book` and `chapter` nodes and `usj-editor.adaptor.ts` synthesizes neither, a verse-mode slice hits that first line every time. Upstream pins it as invariant I5 with a named test: "emits nothing for a selection in a document with no book and no chapter (unaddressable)" (`packages/platform/src/editor/ScriptureReferencePlugin.test.tsx`). The `?? "1"` default only applies when a chapter node exists but has no number.

So Luke 5:0 does not become Luke 1:1 — nothing is reported. Suggested replacement for the last sentence:

```
`ResourceCell` swallows the echo. The guard is defense-in-depth rather than a fix for a
live report: `$resolvePosition` refuses to describe a position in a document with no
BookNode and no ChapterNode (invariant I5), and the slice has neither, so today the
plugin is silent in verse mode. The guard is what keeps that true if a future
platform-editor makes slices addressable.
```

.context/standards/Architecture-Decisions.md
```diff
@@ -415,3 +415,119 @@ step, no automation. Just a record.
...
+    `parseInt(chapterNode?.getNumber() ?? '1', 10)` yields chapter 1, turning Luke 5:0 into
+    Luke 1:1.
+
+    **The guard belongs in the consumer, not upstream in the plugin** (raised in review of #2663;
+    verified against `@eten-tech-foundation/platform-editor` ~0.8.14). Gating the plugin on
+    `isReadonly` was considered and is rejected on the merits, not merely deferred: the plugin is
+    **bidirectional** — `$moveCursorToVerseStart` applies `scrRef` to the caret, and
```
imnasnainaec: :pick: **Stale symbol names in the bidirectionality argument**

The two halves are now `$moveCaretToVerseStart` (applies `scrRef` to the caret) and `$resolvePosition` → `onSelectionSettled` → `report` (reports the caret back). `$moveCursorToVerseStart` and `$findAndSetChapterAndVerse` are the Scribe plugin's names.

The conclusion holds — the plugin is genuinely bidirectional, and gating it on `isReadonly` would break navigation-to-verse in the Resource Viewer and in this grid's chapter mode. Only the citations need updating.

.context/standards/Architecture-Decisions.md
```diff
@@ -415,3 +415,119 @@ step, no automation. Just a record.
...
+    `$findAndSetChapterAndVerse` reports the caret back — and read-only surfaces need both halves.
+    Not mounting it when read-only would break navigation-to-verse in every read-only editor (the
+    Resource Viewer, this grid's chapter mode) and would break read-only click-to-sync, which
+    **these very cells rely on at every normal verse**. `isReadonly` is therefore the wrong
```
imnasnainaec: **"these very cells rely on at every normal verse" is true only in chapter mode**

In chapter mode the cell feeds the full chapter USJ, which carries book and chapter nodes, so the plugin reports normally and click-to-sync works. In verse mode the slice is unaddressable (the I5 guard in `$resolvePosition`), so no click in a verse cell reports anything — at verse 0 or at any normal verse.

This matters because the claim is load-bearing for rejecting the upstream `isReadonly` gate. The rejection still stands on the Resource Viewer and chapter mode alone; the sentence just needs scoping to those, e.g. "which this grid's chapter mode and the Resource Viewer rely on".

Separately: verse-mode click-to-sync not working is pre-existing — `sliceUsjToVerse` has dropped chapter chrome since #2509 — so it isn't a regression here. Probably worth its own ticket if the product expects a click in a verse cell to move the scroll group.

extensions/src/platform-scripture-editor/src/scripture-text-grid/resource-cell.component.tsx
```diff
@@ -155,13 +156,40 @@ export function ResourceCell({
...
-  // omits it — the chapter is identical across verses, but the slice is not).
+  // Only the USJ fed to the editor is resolved — `scrRef` passes through untouched. Keying the memo
+  // on the resolved verse (not scrRef.verseNum) also keeps 1:0 -> 1:1 from re-feeding identical
+  // content.
+  const displayVerseNum = resolveDisplayVerseNum(scrRef.verseNum);
+  const isFallenForward = displayVerseNum !== scrRef.verseNum;
+
+  // Editorial's reference plugin is active even when read-only, and it reports any selection whose
+  // verse doesn't match `scrRef`. Under fall-forward that mismatch is permanent — we hand it verse
+  // 1's text while telling it verse 0 — so a single click would push verse 1 into the shared scroll
```
imnasnainaec: **Same three corrections needed in this comment block**

Against `@eten-tech-foundation/platform-editor` 0.8.15 (installed; `~0.8.14` in `package.json`):

- "reports any selection whose verse doesn't match `scrRef`" — `positionMatchesScrRef` compares book, chapter, **and** verse. Any of the three differing triggers a report.
- "the slice carries no chapter marker, so the plugin defaults to chapter 1 and Luke 5:0 would report Luke 1:1" — no chapter **and** no book means unaddressable, so `$resolvePosition` returns `undefined` and nothing is reported. Chapter 1 is never emitted.
- "these very cells depend on its reporting for click-to-sync at every normal verse" — true in chapter mode, not in verse mode.

The practical consequence is that the `viewMode === 'verse'` branch below is unreachable with the shipped plugin. I'd keep the guard — it's free, and it's the right shape if a future version makes slices addressable — but the comment should say that's what it is, rather than describing a write-back that cannot currently occur. Otherwise the next person to touch this reasons from a mechanism that isn't there.

extensions/src/platform-scripture-editor/src/scripture-text-grid/resource-cell.component.test.tsx
```diff
@@ -215,6 +281,106 @@ describe('ResourceCell viewMode', () => {
...
+
+  it('hands the editor the unresolved reference — only the displayed text falls forward', async () => {
+    renderResourceCell({ viewMode: 'verse', scrRef: verse0, chapterUsj: twoVerseChapterUsj });
+    await waitFor(() => expect(setUsjSpy).toHaveBeenCalled());
+    expect(capturedEditorScrRef).toHaveBeenLastCalledWith(expect.objectContaining({ verseNum: 0 }));
+  });
+
+  // Editorial's reference plugin runs even when read-only and reports any selection whose verse
+  // doesn't match scrRef. Under fall-forward that mismatch is permanent, so without a guard a click
+  // writes verse 1 — and, because the slice drops chapter chrome, chapter 1 — into the scroll group.
```
imnasnainaec: **Both tests in this pair assert against a fabricated plugin payload**

`Editorial` is mocked wholesale, so `getEditorScrRefChange()?.({ book: 'GEN', chapterNum: 1, verseNum: 1 })` is whatever we choose to send. The comment below it states it is "What the real plugin reports for a click on verse-1 text in a chapter-chrome-free slice" — the real plugin reports nothing there (`$resolvePosition` bails on a document with no BookNode and no ChapterNode; pinned upstream as invariant I5). The same applies to "still reports editor selection changes when the verse did not fall forward": in verse mode that report never arrives either.

The tests are still worth keeping — they pin `ResourceCell`'s side of the contract, which is what protects us if the plugin's addressability changes. Suggest dropping the "what the real plugin reports" framing and the "because the slice drops chapter chrome, chapter 1" clause here, and saying instead that these exercise the host-side write-back channel directly. As written, a reader takes them as evidence about plugin behavior that they don't provide.

imnasnainaec (commented):
Reviewed the verse-0 write-back guard against the editor package this repo actually ships. The guard's behavior is fine; the reasoning recorded around it cites a plugin we don't use.

Short version: `$findAndSetChapterAndVerse` and its `parseInt(chapterNode?.getNumber() ?? '1', 10)` chapter-1 fallback belong to `packages/scribe` in `eten-tech-foundation/scripture-editors`. `@eten-tech-foundation/platform-editor` publishes from `packages/platform`, whose plugin was rewritten around a `$resolvePosition` / `onSelectionSettled` / `report` machine. For a document with no BookNode and no ChapterNode — exactly what `sliceUsjToVerse` produces — `$resolvePosition` returns `undefined` and nothing is reported at all. Chapter 1 is never emitted, and the `viewMode === 'verse'` branch of `handleScrRefChange` is unreachable today.

I'd keep the guard: it costs nothing and it's the right shape if slices ever become addressable. But the ADR, the code comment, and the two new tests should stop describing a write-back that cannot currently occur, since the next person to touch this will reason from a mechanism that isn't there.

Checked against platform-editor 0.8.15 as installed, diffed against `packages/platform/src/editor/ScriptureReferencePlugin.tsx` on `main` (only delta is a caret-placement tweak in `$moveCaretToVerseStart`), plus the upstream regression test `emits nothing for a selection in a document with no book and no chapter (unaddressable)`.

One more that sits outside this diff, so it can't be inlined:

> :pick: `extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx:1230` — the comment "ScriptureReferencePlugin returns {book, chapterNum, verseNum} without versificationStr" is no longer accurate. `positionToScrRef` in 0.8.15 rides the host's `versificationStr` along on every position report, pinned upstream by "carries the host's versificationStr on position reports". The `??` fallback is harmless either way, so nothing to change — flagging so the stale rationale doesn't get cited as a constraint later.

---

Authored by Claude Opus 5 (1M context); partially checked by me.

.context/standards/Architecture-Decisions.md
katherinejensen00: Thank you — citation's in. InterlinearizerLoader.tsx, the activeScrRef memo: keeps verse 0 whenever a segment covers it (the extractor emits a synthetic verse-0 scope with SID "<book> <chapter>:0" for pre-verse-1 content), and falls back to verse 1 only when the chapter has no verse-0 segment.
▎
▎ Much better than my paraphrase, because it's precisely the content-aware rule alternative (d) rejects — affordable there because the whole book is segmented before a reference is resolved, which the naming site here can't do.

.context/standards/Architecture-Decisions.md
katherinejensen00: I dug into this and I think we're both partly right — with your half being the more consequential one.
▎
▎ $findAndSetChapterAndVerse and the ?? "1" fallback do exist in platform-editor, in 0.8.14: node_modules/@eten-tech-foundation/platform-editor/src/editor/ScriptureReferencePlugin.tsx. That's what package-lock.json resolves and what npm serves as latest, which is why I cited it.
▎
▎ But that doesn't settle it in my favour, because the lockfile isn't authoritative here. 0.8.14 was published 2026-04-14, and main already imports four editor APIs it doesn't export — PARAGRAPH_STRUCTURE_VIEW_MODE, StructureProtectionMode, ViewOptions.hasGutterParaMarkers, EditorRef.removeCharacterMarker. All four fail tsc on a clean origin/main checkout. So the build we actually run is newer than the pin, and the plugin you read is very likely the live one.
▎
▎ Rather than pick, the ADR now records both, with the guard justified under either: a live fix against 0.8.14's path, defense-in-depth against the rewrite. I took your suggested wording nearly verbatim for the second half, plus a note not to delete the guard on the strength of the rewrite alone.
▎
▎ Separately worth a conversation: main currently doesn't typecheck against its own lockfile. Happy to open that as its own thread.

.context/standards/Architecture-Decisions.md
katherinejensen00: Both name sets are now in, labelled by build — $moveCursorToVerseStart / $findAndSetChapterAndVerse for 0.8.14, $moveCaretToVerseStart / $resolvePosition for the rewrite. See my reply on the mechanism thread for why I kept both rather than swapping. Agreed the conclusion is unaffected either way.

.context/standards/Architecture-Decisions.md
katherinejensen00: You're right, and it's scoped now to this grid's chapter mode and the Resource Viewer.
▎
▎ Worth noting it holds under 0.8.14 too, by a different route: in verse mode the slice contains exactly the verse scrRef names, so a click resolves to that same verse, isVerseInCurrentRange is true, and nothing is reported. Same observable behavior either way, so the scoping is right regardless of which build ships.
▎
▎ On the ticket — I've left it unfiled, and the ADR now says why rather than staying silent. In verse mode the cell only ever holds the verse you're already on, so a click has nowhere to sync to; nothing is lost, which is what makes it unlike PT-4379 where the button genuinely can't reach the previous chapter. If you read it as a real gap I'll file it, just say the word.

extensions/src/platform-scripture-editor/src/scripture-text-grid/resource-cell.component.tsx
katherinejensen00: All three are in the comment block now.
▎
▎ Two land as you wrote them: chapter-or-verse rather than verse alone, and click-to-sync scoped to chapter mode.
▎
▎ The third I've handled differently. The viewMode === 'verse' branch is reachable against 0.8.14 — I traced it: no chapter node → selectedChapterNum 1, verse marker "1" → isVerseInRange(0, '1') false → reports {LUK, 1, 1}. Since the build we run is probably newer than the pin, the comment now documents both paths and says the guard is right either way. I think that gets at your actual concern: the next person shouldn't reason from a mechanism that isn't there — in either direction.

extensions/src/platform-scripture-editor/src/scripture-text-grid/resource-cell.component.test.tsx
katherinejensen00: Agreed and done. The "what the real plugin reports" framing and the chapter-1 clause are gone; the comment now says these exercise the host-side write-back channel and pin ResourceCell's side of the contract, with the payload described as modeled on 0.8.14 rather than offered as evidence about the plugin.
▎
▎ Your point also turned up something: every test around that guard held viewMode at 'verse', so dropping the viewMode === 'verse' check would have broken chapter-mode click-to-sync at every verse 0 with nothing failing. Added 'does not swallow the write-back in chapter mode at verse 0', and confirmed it fails when the check is reverted.

katherinejensen00 (commented):
You're right, and the reason I pushed back is worth writing down so nobody repeats it.
▎
▎ I had the wrong editor. My local node_modules held a stale npm 0.8.14 tarball, so I traced $findAndSetChapterAndVerse and its chapter-1 fallback and wrote them up as the live mechanism — ADR, code comment, test comments, PR description. I then argued the point from npm metadata: 0.8.15 isn't published, latest is 0.8.14, and main imports four APIs 0.8.14 doesn't export. All true, all irrelevant — this repo builds the editor from dev-packages/scripture-editors and yalc-links 0.8.15 over node_modules in postinstall. Your CI logs it on every run. The four "missing" APIs were just my stale link. "The editor package this repo actually ships" was exactly the right framing and I read it as npm when you meant the build.
▎
▎ All five corrections are in, following your suggested wording closely:
▎
▎ - The guard is defense-in-depth, not a fix for a live report — $resolvePosition won't describe a document with no BookNode and no ChapterNode (I5), the slice is exactly that, and the viewMode === 'verse' branch is unreachable today. Luke 5:0 never becomes Luke 1:1. It stays for the reason you gave: free, and the right shape if slices become addressable.
▎ - Bidirectionality now cites $moveCaretToVerseStart and $resolvePosition → onSelectionSettled → report.
▎ - "relies on at every normal verse" scoped to chapter mode and the Resource Viewer, with why — they feed a whole chapter, so the document is addressable.
▎ - Both write-back tests now say they pin our side of the channel with constructed payloads, not plugin behavior.
▎ - The version-comparison passage is deleted rather than reworded.
▎
▎ I re-verified against dev-packages/.../ScriptureReferencePlugin.tsx — only delta from upstream main is the caret tweak in $moveCaretToVerseStart, as you said.
▎
▎ On the ⛏️: confirmed, positionToScrRef carries the host's versificationStr, so the ?? is a no-op and the stated rationale is stale. Left alone since it's outside this diff and you flagged it as informational — say the word and I'll fix the comment here rather than leave a wrong one sitting.
▎
▎ One thing your review changed beyond what you asked: it prompted me to check the guard's other half, and every existing test held viewMode at 'verse'. Dropping the viewMode === 'verse' check would have broken chapter-mode sync at every verse 0 with nothing failing — which matters more now that chapter mode is the only place the plugin actually reports. Added 'does not swallow the write-back in chapter mode at verse 0' and confirmed it fails when the check is reverted.

> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2663#-:-P-CNDt86roS-bG2PHEk:b-pxs75t)* status: 3 of 10 files reviewed, 6 unresolved discussions (waiting on imnasnainaec).

___
*[`lib/platform-bible-react/src/components/advanced/book-chapter-control/book-chapter-control.navigation.ts` line 42 at r2](https://reviewable.io/reviews/paranext/paranext-core/2663#-P--JaaV3Rz89TCnGryx:-P-BJTB15cFu8aBJbV2L:b276j94) ([raw file](https://github.com/paranext/paranext-core/blob/075c81cc82c932bfd75cc0066755fdf2f0d1f0b3/lib/platform-bible-react/src/components/advanced/book-chapter-control/book-chapter-control.navigation.ts#L42)):*
<details><summary><i>Previously, imnasnainaec (D. Ror.) wrote…</i></summary><blockquote>

:pick: This added parenthetical is justifying the diff, but doesn't really belong in the final product.
</blockquote></details>

Agreed — removed. It was there to explain the diff to you, and that job ends when the PR merges; git blame carries "it used to say PT-4143" for anyone who needs it. The comment now just cites PT-4379.
▎
▎ One deliberate exception, flagging it so it doesn't look like I missed the same pattern: the equivalent parenthetical in ADR-0013 (the earlier draft calling the previous-verse button "inert") stays. That file is an append-only decision log where the convention is to mark superseded reasoning rather than delete it — a reader who saw the earlier claim needs to know it was withdrawn, since the ADR is the durable record and has no blame trail worth reading. Happy to drop it too if you'd rather keep that rule strictly to code.

katherinejensen00 (commented):
Retracting the version argument I made across these threads — @imnasnainaec was right and I was wrong, for a reason worth naming.
▎
▎ I had the wrong editor. My local node_modules held a stale npm 0.8.14 tarball, so I traced $findAndSetChapterAndVerse and its chapter-1 fallback and defended them as the live mechanism. They aren't: postinstall → link-dev-packages builds dev-packages/scripture-editors at 0.8.15 and yalc-links it over node_modules — CI logs it every run. "The editor package this repo actually ships" was the right framing; I read it as the registry when you meant the build.
▎
▎ Now checkable by anyone, since 0.8.15 published yesterday: npm pack @eten-tech-foundation/platform-editor@0.8.15 → src/editor/ScriptureReferencePlugin.tsx has $resolvePosition, onSelectionSettled, the I5 guard at if (!chapterNode && !bookNode) return undefined, and no $findAndSetChapterAndVerse.
▎
▎ So, taken as written rather than half-accepted: the guard is defense-in-depth, the plugin is silent in verse mode, the viewMode === 'verse' branch is unreachable, and Luke 5:0 never becomes Luke 1:1. ADR, guard comment, test comments and PR description all say that now. Sorry for the detour.
▎
▎ Also added while checking: 'does not swallow the write-back in chapter mode at verse 0' — every prior test held viewMode at 'verse', so dropping that check would have broken chapter-mode sync silently. That matters more now that chapter mode is the only place the plugin reports.

> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2663#-:-P-FTyVC1xPI71xhcBcg:b-k3g85g)* status: 3 of 10 files reviewed, 6 unresolved discussions (waiting on imnasnainaec).

extensions/src/platform-scripture-editor/src/scripture-text-grid/resource-cell.component.test.tsx
```diff
@@ -215,6 +281,133 @@ describe('ResourceCell viewMode', () => {
...
+    expect(lastFedUsjText()).toContain('verse one');
+    expect(lastFedUsjText()).not.toContain('verse two'); // still a single-verse slice
+    expect(screen.queryByText(/no text for this verse/i)).not.toBeInTheDocument();
+  });
+
+  it('falls back to the empty state when the chapter has no verse 1 either', async () => {
+    renderResourceCell({ viewMode: 'verse', scrRef: verse0, chapterUsj: noVerseOneChapterUsj });
+    expect(await screen.findByText(/no text for this verse/i)).toBeInTheDocument();
+    // Must not scan forward to the next available verse — verse 2 is not what 1:0 means.
+    expect(lastFedUsjText()).not.toContain('verse two');
```
imnasnainaec: **This assertion can't fail**

`lastFedUsjText()` returns `''` whenever `setUsj` was never called — `setUsjSpy.mock.lastCall ?? []` leaves `fedUsj` undefined, and `JSON.stringify(undefined) ?? ''` is `''`. And by the time this line runs, the preceding `findByText` has already established that the empty state rendered, which means the effect early-returned and `setUsj` was never called. So `''` is the only value this line can ever see, whatever the slicer does.

`expect(setUsjSpy).not.toHaveBeenCalled()` would actually pin "does not scan forward to the next available verse" — and it's what the verse-99 test above already uses.

extensions/src/platform-scripture-editor/src/scripture-text-grid/resource-cell.component.test.tsx
```diff
@@ -215,6 +285,135 @@ describe('ResourceCell viewMode', () => {
...
+
+  it('verse mode at verse 0 displays verse 1, not the empty state', async () => {
+    renderResourceCell({ viewMode: 'verse', scrRef: verse0, chapterUsj: twoVerseChapterUsj });
+    await waitFor(() => expect(setUsjSpy).toHaveBeenCalled());
+    expect(lastFedUsjText()).toContain('verse one');
+    expect(lastFedUsjText()).not.toContain('verse two'); // still a single-verse slice
+    expect(screen.queryByText(/no text for this verse/i)).not.toBeInTheDocument();
+  });
+
+  it('falls back to the empty state when the chapter has no verse 1 either', async () => {
```
imnasnainaec, line 304: :pick: **Nothing covers leaving the verse-empty state**

`ResourceCellView` swaps the editor out entirely for the ghost-text label when `isVerseEmpty` (`resource-cell-view.component.tsx:211-218`), so `editorRef.current` is null while it shows. Two tests enter that state; none leaves it.

Worth adding one that navigates from a verse the resource lacks to one it has and asserts `setUsj` receives the new slice. That pins the refeed after `Editorial` remounts, which lands only because React assigns refs before effects run — a real ordering dependency that nothing currently guards.

Fall-forward makes the empty↔ready transition reachable at chapter boundaries more often, so this path gets exercised more than it used to.

extensions/src/platform-scripture-editor/src/scripture-text-grid/verse-display.utils.ts
imnasnainaec, line 21: :pick: **Unstated precondition on the verse number**

This special-cases exactly `0`, so a negative or fractional verse passes through untouched and then slices to nothing — surfacing as the ghost-text empty state rather than as anything diagnosable.

Not reachable today: `SerializedVerseRef.verseNum` comes from `@sillsdev/scripture`, and previous-verse navigation floors at 0. But ADR-0013 earmarks this helper for promotion to `lib/platform-bible-utils/src/scripture/`, where callers outside this extension lose that guarantee.

A TSDoc line stating the input is a non-negative integer would be enough, matching how `parseVerseRange` just below already documents its non-numeric behavior. A runtime guard would be premature.

imnasnainaec (commented):
Two nitpicks, nothing blocking. Both are about durability rather than current behavior: a test-coverage gap around leaving the verse-empty state, and an unstated precondition on a helper that's earmarked for promotion into `platform-bible-utils`.

The corrected guard rationale reads accurately against the linked 0.8.15 build now.

---

Authored by Claude Opus 5 (1M context), inspired by Devin; reviewed and pared down by me.

imnasnainaec (commented):
> @imnasnainaec reviewed 4 files and all commit messages, and resolved 7 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2663)* status: 7 of 10 files reviewed, 2 unresolved discussions (waiting on katherinejensen00).

extensions/src/platform-scripture-editor/src/scripture-text-grid/resource-cell.component.test.tsx
katherinejensen00: Confirmed on every step, and fixed — thank you. Swapped to `expect(setUsjSpy).not.toHaveBeenCalled()`.

I mutation-checked it rather than trusting the reasoning: patching `sliceUsjToVerse` to fall through to the next present verse when the requested one is empty now fails this test. The old assertion passed under that same mutant, so it was pinning nothing at all — exactly as you said.

I also grepped the other five `.not.toContain(...)` sites rather than eyeballing them. Each one follows a positive `toContain`, which proves the feed happened, so they're falsifiable. Yours was the only vacuous one.

One thing beyond what you asked, so flagging it for pushback: I made `lastFedUsjText()` throw instead of returning `''` when `setUsj` was never called. The silent `''` is what made the bad assertion look reasonable, and nothing else in the file wants that fallback — every caller asserts against a feed that definitely happened. Happy to revert to just the assertion swap if you'd rather keep the helper dumb.

extensions/src/platform-scripture-editor/src/scripture-text-grid/resource-cell.component.test.tsx
katherinejensen00: Confirmed the premise and added it — 'refeeds the slice when navigating out of the empty state into a verse that exists', navigating 1:99 → 1:2 and asserting the new slice arrives.
▎
▎ Mutation-checked rather than assumed: removing verseSlice from the effect's dependency array fails this test and nothing else in the file, so the transition really was unguarded.
▎
▎ Worth recording one dead end, since it shaped the test. My first mutant was if (!editorRef.current) return; in place of the optional chaining — which passes, because it's semantically identical to what's already there. The ref-ordering dependency you name isn't separately falsifiable from inside the component; what is pinnable is the observable consequence, that the refeed happens at all on that transition. That's what the test asserts, with your ordering rationale written above it as the reason it works.

extensions/src/platform-scripture-editor/src/scripture-text-grid/verse-display.utils.ts
katherinejensen00: Added, taking the runtime-guard call with it — @param verseNum now states the non-negative-integer expectation, what actually happens if it's violated (falls through, slices to nothing, surfaces as the empty state rather than an error), and that the guarantee is caller-side today but lost on promotion.
▎
▎ Agreed on no runtime guard: the helper is a two-line pure function, and a throw would be the first thing a promoted version had to reconsider anyway. The promotion note in ADR-0013 is where that decision belongs if it's ever revisited.

katherinejensen00 (commented):
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2663)* status: 7 of 11 files reviewed, 2 unresolved discussions (waiting on imnasnainaec and PARAM).

imnasnainaec (commented):
2 minor nitpicks.

> @imnasnainaec reviewed 3 files and all commit messages, made 3 comments, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2663#-:-P-Q80W12EYWmNzMhE7Y:bqk2sev)* status: 9 of 11 files reviewed, all discussions resolved.

___
*[`.context/standards/Architecture-Decisions.md` line 472 at r5](https://reviewable.io/reviews/paranext/paranext-core/2663#-P-Q7NnsBc2hcz-Hp9Rc:-P-Q7NnsBc2hcz-Hp9Rd:b-cvagxf) ([raw file](https://github.com/paranext/paranext-core/blob/1d6c8aafd84cd4b32e74c80da87f129d62c7b49b/.context/standards/Architecture-Decisions.md#L472)):*
> ```Markdown
>     verse cell holds only the verse you are already on and so has nowhere to sync TO. Deliberately
>     not filed.) `isReadonly` is therefore the wrong
>     predicate: a read-only editor reporting its caret position is correct behavior, not the bug.
> ```

:pick: This sub-bullet paragraph needs to be reflowed. It has a half-length line ending in a mid-sentence line-break.
___
*[`extensions/src/platform-scripture-editor/src/scripture-text-grid/resource-cell.component.test.tsx` line 163 at r5](https://reviewable.io/reviews/paranext/paranext-core/2663#-P-Q6Bxe5vw19IxOB7u4:-P-Q6Bxe5vw19IxOB7u5:bmk1pnp) ([raw file](https://github.com/paranext/paranext-core/blob/1d6c8aafd84cd4b32e74c80da87f129d62c7b49b/extensions/src/platform-scripture-editor/src/scripture-text-grid/resource-cell.component.test.tsx#L163)):*
> ```XML
>   // Throws rather than returning '' when nothing was fed: an empty string silently satisfies every
>   // `.not.toContain(...)`, which is how a vacuous assertion slipped in here once. If the editor was
>   // never fed, assert `expect(setUsjSpy).not.toHaveBeenCalled()` instead of inspecting content.
> ```

:pick: The comments within this function should probably be merged into the (now-stale) doc line preceding the function.

.context/standards/Architecture-Decisions.md
```diff
@@ -415,3 +415,141 @@ step, no automation. Just a record.
...
+    **bidirectional** — `$moveCaretToVerseStart` applies `scrRef` to the caret, and `$resolvePosition`
+    → `onSelectionSettled` → `report` reports the caret back — and read-only surfaces need both
+    halves. Not mounting it when read-only would break navigation-to-verse in every read-only editor
+    and would break read-only click-to-sync, which **this grid's chapter mode and the Resource Viewer
+    rely on** — those feed a whole chapter, so the document carries the book and chapter nodes
+    `$resolvePosition` needs. (Scoped deliberately: click-to-sync is *not* load-bearing in verse mode,
+    where the slice is unaddressable and no click reports anything. That predates this decision —
+    `sliceUsjToVerse` has dropped chapter chrome since #2509 — and is not a lost capability, since a
+    verse cell holds only the verse you are already on and so has nowhere to sync TO. Deliberately
+    not filed.) `isReadonly` is therefore the wrong predicate: a read-only editor reporting its
```
katherinejensen00, line 816: Done. The short line was left over from an earlier edit that cut the sentence's opening and never reflowed the tail. The paragraph is now wrapped to the file's width end to end (`b80c9e5`).

extensions/src/platform-scripture-editor/src/scripture-text-grid/resource-cell.component.test.tsx
```diff
@@ -93,11 +117,58 @@ const twoVerseChapterUsj = usxStringToUsj(`<?xml version="1.0" encoding="utf-8"?
...
   mockUseProjectData.mockReturnValue({ ChapterUSJ: () => [value, vi.fn(), isLoading] });
 }
 
+/**
+ * Serialized USJ most recently handed to the editor.
+ *
+ * Throws rather than returning '' when nothing was fed: an empty string silently satisfies every
+ * `.not.toContain(...)`, which is how a vacuous assertion slipped in here once. If the editor was
+ * never fed, assert `expect(setUsjSpy).not.toHaveBeenCalled()` instead of inspecting content.
+ */
```
katherinejensen00, line 165: Done, and thank you for catching that the doc line was stale — it still promised `''` when nothing was fed, which stopped being true when the throw went in last round. The in-body comments are now the doc block's second paragraph, and the summary line no longer contradicts the code (`b80c9e5`).

katherinejensen00 (commented):
Both nitpicks taken in b80c9e5 — comment-only, no behavior change. 24 tests in the touched file still pass; Prettier clean on both files.

imnasnainaec (approved):
> @imnasnainaec reviewed 2 files and all commit messages, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2663)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

katherinejensen00:
@imnasnainaec — you signed this off before three pushes landed, so here is exactly what changed since. All comment- and bookkeeping-only; no behavior change, and the shape of the diff is identical (11 files, 518 insertions).

1. **`lastFedUsjText` no longer ends `?? ''`** (`2a083f1`). The throw covers the never-fed case, so the fallback could only mask `JSON.stringify(undefined)`, which nothing feeds. Returning `undefined` makes `.toContain`/`.not.toContain` fail loudly instead of passing vacuously — which is what the doc block you asked me to fix already promises.
2. **The ADR is now ADR-0019** (`04c47ff`). `main` landed ADR-0013 through ADR-0018 while this was in review, so the number this branch claimed was taken twice over. The ADR text is unchanged; the four citations in the grid's source comments and every reference in the PR description moved with it.
3. **Comment sweep for #2586's forward-facing rule** (`0804d1f`), which merged after this branch was written. Seven sites cut, including two you reviewed in their current form — the "that stale rationale was corrected here" clause and "which is how a vacuous assertion slipped in here once". Details in the PR description; the ADR-0019 pointers, the open PT-4379 follow-up and the PT-3495 regression note all stay, since those point forward. Test titles left alone.

Also rebased three times (PT-4341, PT-4111 and friends). Three files this branch touches were touched by those PRs; all auto-merged, and I verified the `versificationStr` comment, the Resource Viewer chapter-fetch comment, and the a11y comment plus `resolveDisplayVerseNum` call each survived intact.

Verified on the final base: `npm run typecheck` clean repo-wide, `npm run lint` 0 errors (4 pre-existing warnings, none in this diff), 159 tests in `scripture-text-grid/` and 21 in `book-chapter-control.navigation.test.ts` passing.

One environment note that may save you an hour, since it is machine-wide rather than specific to this branch: `link-dev-packages` exits 0 but yalc-publishes `platform-editor` and `scripture-utilities` **src-only** unless something has already built their `dist` — `devpub` is just `yalc push`, and nothing in that chain runs a build. With src-only copies, `moduleResolution: "Node"` cannot resolve the packages at all, so typecheck reports a dozen unresolved-module errors, ESLint reports `import/no-unresolved`, and Vitest resolves the package's `src/index.ts` and dies on the editor repo's internal `"shared"` alias. Fix is `pnpm nx extract-api` for both packages, then `devpub` + link. Worth knowing that `yalc push` propagates to **every** linked worktree at once — mine printed five, so one bad link run breaks all of them.

imnasnainaec (approved):
> @imnasnainaec reviewed 7 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2663)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2665, opened 2026-08-10
Wire character-marker removal into the menu and bar

jolierabideau:
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2665)

---

# Code Review Summary

**Branch**: `remove-character-marker`

**Base**: `origin/mount-character-control-in-gutter` (PR #2660)

**Date**: 2026-08-10

**Files changed**: 10 (the E1 work; see *Reading this diff* below)

## Overview

Wires character-marker **removal** into the character-marker menu and bar — issue E1 of the
character-level-markers plan, under the PRD *"Saroj easily works with character-level markers"*.
It adds `useRemoveCharacterMarker` (the removal action, wrapped in a version-history snapshot and a
Send/Receive guard), drives the menu's remove rows and per-row selection states from the selection's
per-marker coverage, and forwards that coverage into the menu generator so a row's action and its
displayed state are decided in one place.

The review began against the wrong base and that dominated the first pass. The branch had forked
from an older state of PR #2660 and then merged `main`, so it never picked up #2660's review fixes —
leaving a duplicate, pre-squash copy of the character-marker control at
`src/character-marker-control.component.tsx` that the bar was actually mounting. Merging
`origin/mount-character-control-in-gutter` resolved that and four findings with it. Everything below
is against the corrected base.

## Reading this diff

GitHub shows **97 files / +11,420 / −830** because #2660 has not merged `main` yet while this branch
has. All of PT-4277's multi-window work (#2621) therefore appears in the comparison. **The E1 work is
10 files**, all under `extensions/src/platform-scripture-editor/` plus `Architecture-Decisions.md`.
The noise disappears once #2660 lands on `main`.

## API Changes

None public. One extension-internal signature widened:

- `useCharacterMarkerState` / `generateCharacterMarkerMenuListItems`: `removeCharacterMarker`
  widened from `(marker: string) => void` to `(marker?: string) => void`. Calling with no argument
  removes every character marker the selection covers. A widening, so non-breaking.
- Localize key renamed before ever shipping: `%…_characterMarkerMenu_removeAllMarkers%` →
  `%…_characterMarkerMenu_removeMarkers%`.

- [x] **Duplicate, stale copy of the character-marker control at `src/character-marker-control.component.tsx`, which `CharacterMarkerBar` was mounting** — cost the menu-closes-on-select fix, the localized `label_format`/`ariaLabel_format` strings, the required `menuDirection`, `useTruncationTooltip`, `DisabledTooltipWrapper`, and the double-`onClose` guard; also orphaned `character-marker-control/character-marker-toolbar.component.tsx`. Typecheck and tests both passed, because the two prop shapes are structurally compatible and both test files pass side by side. _(fixed during review: merged `origin/mount-character-control-in-gutter`; git resolved the duplicate to a deletion and the bar now imports from `character-marker-control/`)_
- [x] **Menu search field would have rendered a raw localize key** — the stale control read `%webView_platformScriptureEditor_characterMarkerMenu_searchPlaceholder%`, which the branch re-added to `localizedStrings.json`, while the web view preloads `%markerMenu_searchPlaceholder_character%` from `character-marker-control.const.ts`. The stale key was never resolved at runtime. _(fixed during review: resolved by the same merge; the orphaned key is gone from `localizedStrings.json`)_

Author response: both were artifacts of the branch having forked from an older state of #2660. The
author identified the wrong-base problem themselves — that insight is what turned the review around.

- [x] **The catch-all remove row overstated what it does.** It read "Remove all character markers" but issues one argument-less call, and the editor's `$getCharNodeToRemove` matches only the *innermost* `CharNode` when the marker is `undefined` — so a nested stack takes one pass per layer. _(fixed during review: renamed the key to `…_removeMarkers%` and reworded to "Remove character markers" / "Quitar marcadores de carácter"; logic unchanged, so undo stays one step per activation. Recorded as ADR-0011.)_
- [ ] ~~Silent partial refusal on the catch-all row: `$splitCharNodeAroundTargets` returns `undefined` for a run whose innermost marker is nested and only partly covered, and that run is skipped with no notification.~~ _(Author agreed to leave it. Detecting it needs an outcome signal from the editor — a return value from `removeCharacterMarker`, or discrete-update-plus-re-derive — which is an editor-package change and outside the PRD's 2-week appetite. A before/after `getUsj()` comparison cannot substitute: it reads the cached `editedUsjRef` that the non-discrete update does not refresh, so it would report "unchanged" for successful removals too. `isMarkerRowInert` already disables `'partial'` rows, so the reachable case is narrow. Recorded as ADR-0011's revisit condition.)_

### Minor — Consider

- [x] **`selectionState` on the remove rows emitted a false `aria-checked`.** `MarkerMenu` maps the prop onto `aria-checked` (`'all'`→true, `'partial'`→'mixed', `'none'`→false) plus a checkbox affordance, and its contract is "how much of the selection *this marker* covers" — a question a markerless action row does not pose. On a fully-marked selection the honest coverage reading is `'none'`, which announces as "not checked" beside an action certain to remove markers. _(fixed during review: dropped from **both** remove rows — they occupy the same either/or slot, so fixing only the new one would make that slot render inconsistently. Touches #2660's row, which is still an open PR, so the correction lands with the feature.)_
- [x] **The `markerMenuItems` memo comment mis-attributed its cost** — it billed the coverage-forwarding refactor for a re-run that commit did not introduce, inviting a future reader to reverse a good refactor to recover a saving that was never spent there. _(fixed during review: reworded to explain why coverage is an **input** to the rows rather than a decoration — it gates whether the remove row exists at all, and a `.map()` over finished items cannot rewrite an action closure — and to record memo-splitting as the escape hatch if the sort ever measures)_
- [x] **`notify()`'s own failure path was uncovered** — every other branch of the hook had a test. _(fixed during review: added a test asserting both notifying sites still resolve when `papi.notifications.send` rejects. Verified falsifiable — removing `notify`'s `try/catch` fails that test and only that test.)_
- [x] **The one-layer-removal decision lived only in a code comment and a commit message** — and this feature had no ADR at all, the branch-local one having been dropped upstream. _(fixed during review: recorded as ADR-0011 with rejected alternatives and a revisit condition)_
- [ ] ~~`isMarkerRowInert` mirrors the row action's branch conditions condition-for-condition — two places to keep in sync.~~ _(Left as-is. All branches traced and they currently agree; the two sit adjacent and the mirroring is documented as deliberate, and no cheap structural fix beats that.)_

Author response: the author asked for all four fixable items to be applied, and accepted the
reasoning for leaving the two dismissed ones.

#### Shared Regions Modified

None.

#### Extension Config Changes

None — no config or build files were touched in `extensions/`.

## Positive Observations

- **The removal action follows the established destructive-edit precedent exactly** — version-history
  snapshot before the edit, best-effort, with the same `ERROR_UNIMPLEMENTED` handling as the
  insert-footnote and insert-cross-reference paths.
- **The sync-blocked guard is belt-and-braces with a test**, covering the race where a Send/Receive
  starts while the menu is already open.
- **A real ordering assertion**, not just "both were called": the snapshot test uses
  `invocationCallOrder` to prove the snapshot precedes the removal.
- **The comment explaining why a before/after `getUsj()` comparison cannot detect a refusal** is the
  kind of negative result that saves the next person a day.
- **Removing every covered marker is one argument-less call, not a loop**, with the undo-atomicity
  reasoning written down at the site.
- **The localized-strings parity test is driven off the exported key list**, so a key added to the
  menu is covered without anyone remembering to update the test.
- **Coverage handles nesting properly** — `getAncestorCharacterMarkers` walks the full ancestor
  chain, which is what makes per-marker rows an exact path for a nested marker.

## Interview Notes

**Stated purpose**: wire removal into the character marker — issue E1 of
`docs/plans/2026-07-27-character-level-markers-implementation-plan.md`.

**The author's most valuable contribution was diagnosing the review's own premise.** When presented
with two Critical findings, they identified that the branch should be based on
`mount-character-control-in-gutter` — which was correct, and collapsed four findings at once. That
is a stronger result than answering the findings would have been.

**Design understanding — demonstrated in depth.** Asked why the coverage-forwarding refactor was
worth its cost, the author read the refactor commit and every prior version of the hook, then gave a
three-part answer: coverage decides whether the remove row *exists* (`resolveCurrentMarker` returns
`undefined` for two-or-more covering markers, so the mixed case had no `currentMarker` to gate on and
a post-hoc stamp cannot add a row); coverage decides what the *action* does, and a `.map()` can
overwrite fields but cannot rewrite a closure, so "keep the split and teach the stamper" was never
actually available without re-implementing the nesting reasoning in a second file; and the old
stamper was already re-deriving the generator's emission rules from outside, which was the smell that
the boundary sat in the wrong place. All three verified correct.

**One factual disagreement, resolved by scope.** The author argued the memo comment billed the
refactor for a cost the design already owed, citing identical dep arrays across `c5d809aa06a`. That
is correct *for that commit* — its parent `319fc75ce0b` had already collapsed the two memos. But at
the review base the generator memo is keyed **without** `coverage` (deliberately, per its own
comment), with only a cheap `.map()` keyed on it — so the sort genuinely did not re-run on open/close
until a sibling E1 commit changed that. The comment was accurate for the branch as a whole; the
attribution was the misleading part, and it has been reworded to describe the design instead.

**No areas were deferred to AI or left unexplained.** The author's corrections were specific,
sourced to commits, and independently verifiable.

## In-Review Quality Check

Run after every change, all synchronous:

- `npm run typecheck` — **clean** (exit 0)
- `npm run lint` — **clean**, 0 errors (4 warnings, all pre-existing)
- Extension test suite — **776 passed / 54 files**
- Falsifiability check on the one new test — **verified**: removing `notify`'s `try/catch` fails that
  test and only that test

No unfixable failures.

## Reviewer-introduced regression, caught and fixed

Flagging explicitly because no automated check would have caught it. The merge resolved
`.context/standards/Architecture-Decisions.md` by taking the gutter branch wholesale — but that
branch predates PT-4277 (#2621), which added **ADR-0007 through ADR-0010** on `main`. The resolution
silently dropped all four. Restored from `origin/main`, verified correct: the gutter branch's copy is
byte-identical to its base, and main's is that base plus a pure 113-line addition. It surfaced only
because the next free ADR number was being looked up.

## Suggested Review Focus

- [ ] **The `selectionState` removal touches PR #2660's remove row**, not just the new one. Confirm
      that is acceptable while #2660 is still open, or ask for the single-row half to be backed out.
- [ ] **One-layer removal (ADR-0011)** — is "Remove character markers" the right label, and is
      one-activation-per-nesting-layer acceptable for the PRD's user? The alternative (a loop) costs
      one undo entry per layer.
- [ ] **The silent partial refusal**, dismissed above with reasoning. Worth a second opinion on
      whether the narrow reachable case justifies leaving it unreported.
- [ ] **Merge order** — this PR targets `mount-character-control-in-gutter` (#2660). It should land
      after #2660, and the 97-file comparison shrinks to 10 once #2660 has `main` merged in.
- [ ] **`isMarkerRowInert`'s mirroring** of the action's branches — a second reader on whether the
      adjacency-plus-comment approach is enough to keep them from drifting.


katherinejensen00 (changes requested):
Ordered by severity. Each lists the changed file/line to attach a comment.

**1\. Snapshot is committed before the removal even on the no-op paths** — _Low_ [use-remove-character-marker.hook.ts:97](vscode-webview://1bbvfdtssqal87jrv2dsq5vgmhtuk52mipgm3refu8v9p4781njp/extensions/src/platform-scripture-editor/src/character-marker-bar/use-remove-character-marker.hook.ts#L97) `commitSnapshot()` runs unconditionally before `editorRef.current?.removeCharacterMarker()`. If the editor ref is null (`?.` no-ops) or the editor silently declines a partial-nested removal (your own documented case), a "Before removing character marker" version-history point is written for an edit that never happened, with no notification. This mirrors the insert-footnote precedent, and the reachable menu paths are narrow (the partial-decline row is disabled by `isMarkerRowInert`), so it's genuinely low — but insert can't no-op, whereas removal can. Whether empty snapshots actually accumulate depends on whether `commitChanges` no-ops on an unchanged tree; worth a one-line acknowledgment at the site, or gating the snapshot on a non-null ref.

**2\. The pre-destructive-edit snapshot logic now exists in three places** — _Low (reuse)_ [use-remove-character-marker.hook.ts:49](vscode-webview://1bbvfdtssqal87jrv2dsq5vgmhtuk52mipgm3refu8v9p4781njp/extensions/src/platform-scripture-editor/src/character-marker-bar/use-remove-character-marker.hook.ts#L49) `commitSnapshot` duplicates the exact `commitChanges` + `true` + `ERROR_UNIMPLEMENTED`\-sentinel + best-effort-warn block already inlined twice in the web view (insert-footnote, insert-cross-reference). A change to that contract must now be made in three spots. A single host-safe helper would collapse all three — the ideal fix, though it means touching the web view's existing sites.

**3\. The sync-blocked notification is a second copy of `notifySyncEditBlocked`** — _Low (reuse)_ [use-remove-character-marker.hook.ts:34](vscode-webview://1bbvfdtssqal87jrv2dsq5vgmhtuk52mipgm3refu8v9p4781njp/extensions/src/platform-scripture-editor/src/character-marker-bar/use-remove-character-marker.hook.ts#L34) `notify()` re-emits the same `%…error_syncEditBlocked%` key at `severity: 'warning'` — exactly what the web view's `notifySyncEditBlocked` exists to centralize ("so the severity/message cannot drift across the several call sites"). It's a cross-module `useCallback`, so literal reuse isn't trivial, but the single-source invariant is now defeated for this key.

**4\. The "mixed coverage" predicate is duplicated verbatim across two files** — _Low (reuse)_ [use-character-marker-state.hook.ts:185](vscode-webview://1bbvfdtssqal87jrv2dsq5vgmhtuk52mipgm3refu8v9p4781njp/extensions/src/platform-scripture-editor/src/use-character-marker-state.hook.ts#L185) `coveringMarkers.length > 1 || (coveringMarkers.length > 0 && coverage.hasUncovered)` appears identically here (as `isMixed`) and in `character-marker-menu.utils.ts` (as `isMixedCoverage`). If "mixed" ever changes, the trigger's `(mixed)` label and the menu's remove-all row could disagree. `character-marker-coverage.utils.ts` already exports `isEmptyCoverage` "so that test lives in one place" — an `isMixedCoverage` helper beside it fits the established pattern.

**5\. `%…error_syncEditBlocked%` is preloaded twice** — _Nit_ [use-remove-character-marker.hook.ts:19](vscode-webview://1bbvfdtssqal87jrv2dsq5vgmhtuk52mipgm3refu8v9p4781njp/extensions/src/platform-scripture-editor/src/character-marker-bar/use-remove-character-marker.hook.ts#L19) (and the literal at [platform-scripture-editor.web-view.tsx:190](vscode-webview://1bbvfdtssqal87jrv2dsq5vgmhtuk52mipgm3refu8v9p4781njp/extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx#L190)) `REMOVE_CHARACTER_MARKER_STRING_KEYS` includes `SYNC_BLOCKED_KEY`, which `EDITOR_LOCALIZED_STRINGS` also lists literally. Harmless at runtime, but a dedup "cleanup" of either could look safe and silently break the other. (This does confirm the sync-blocked string is loaded into the bar — the wiring works; it's just doubly-listed.)

**6\. `selectionState` is computed three times, one form divergent** — _Nit_ [character-marker-menu.utils.ts:201](vscode-webview://1bbvfdtssqal87jrv2dsq5vgmhtuk52mipgm3refu8v9p4781njp/extensions/src/platform-scripture-editor/src/character-marker-menu.utils.ts#L201) Display (L192) and `isMarkerRowInert` (L194) use `coverage.markerStates[marker] ?? 'none'`; the `action()` at L201 omits `?? 'none'`. Equivalent today (both `undefined` and `'none'` hit the same branch), but the `isMarkerRowInert` doc says the paths "mirror condition-for-condition" — hoist one normalized `const` so a future `'none'`\-specific branch can't diverge silently.

**7\. Double-negative branch guard** — _Nit_ [character-marker-menu.utils.ts:211](vscode-webview://1bbvfdtssqal87jrv2dsq5vgmhtuk52mipgm3refu8v9p4781njp/extensions/src/platform-scripture-editor/src/character-marker-menu.utils.ts#L211) `else if (selectionState !== 'partial' && selectionState !== 'all')` is the branch that must _never_ nest a duplicate marker; a positive allowlist (`=== 'none' || === undefined`) states intent directly and excludes any future 4th enum value by default rather than admitting it into the insert path.

**8\. One-layer-removal rationale duplicated in 4 places** — _Nit (docs)_ [character-marker-menu.utils.ts:36](vscode-webview://1bbvfdtssqal87jrv2dsq5vgmhtuk52mipgm3refu8v9p4781njp/extensions/src/platform-scripture-editor/src/character-marker-menu.utils.ts#L36) The "one call, not a loop / named marker matches at any depth / undefined = innermost" explanation is restated in full in ADR-0011, this key's JSDoc, the `removeRow` inline block (L252), and the state-hook memo. Keep ADR-0011 authoritative; shorten the three code copies to a pointer. (Comments here are excellent but over-volumed overall — the `aria-checked` and `getUsj()`\-staleness notes earn their length; these repeats don't.)

**9\. Mid-phrase comment wrap** — _Nit_ [character-marker-menu.utils.ts:207](vscode-webview://1bbvfdtssqal87jrv2dsq5vgmhtuk52mipgm3refu8v9p4781njp/extensions/src/platform-scripture-editor/src/character-marker-menu.utils.ts#L207) `…would nest / instead of / extending.` reads awkwardly across the break — reflow.

**10\. `localized-strings.test.ts` comment now describes the wrong shipping source** — _Low (doc accuracy)_ [localized-strings.test.ts:71](vscode-webview://1bbvfdtssqal87jrv2dsq5vgmhtuk52mipgm3refu8v9p4781njp/extensions/src/platform-scripture-editor/src/localized-strings.test.ts#L71) This is genuine branch work (no missing base commit here), so it _won't_ self-resolve. An E1 commit reworded the exclusion comment to say `%markerMenu_searchPlaceholder_character%` "Ships in `platform-bible-react`'s own localized strings." It actually ships from core `assets/localization/*.json`; PBR's copy is a Storybook-only pseudo-localized mock. This reverts the clarification from `ebcd2a2c210` and reinstates the exact trap that fix removed. Restore the accurate wording ("do not treat a definition in PBR's `localizedStrings.json` as evidence the string ships").

**11\. Bar component test mocks away the very wiring it appears to verify** — _Low (test quality)_ [character-marker-bar.component.test.tsx:41](vscode-webview://1bbvfdtssqal87jrv2dsq5vgmhtuk52mipgm3refu8v9p4781njp/extensions/src/platform-scripture-editor/src/character-marker-bar/character-marker-bar.component.test.tsx#L41) `vi.mock('./use-remove-character-marker.hook', …)` replaces the hook with a bare spy, so nothing asserts the bar actually passes `projectId` into it — drop the `projectId={projectId}` prop and every test still passes. Also, the module-level `removeCharacterMarker` spy is **not** cleared in `afterEach` ([:97](vscode-webview://1bbvfdtssqal87jrv2dsq5vgmhtuk52mipgm3refu8v9p4781njp/extensions/src/platform-scripture-editor/src/character-marker-bar/character-marker-bar.component.test.tsx#L97)) while sibling state is; a future `toHaveBeenCalledTimes`/`not.toHaveBeenCalled` assertion will see leaked calls. Add `removeCharacterMarker.mockClear()`.

> @katherinejensen00 reviewed 97 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2665#-:-OziEidz1F6cvwEoyLzs:by5214m)* status: all files reviewed, 1 unresolved discussion (waiting on jolierabideau).

___
*[`extensions/src/platform-scripture-editor/src/character-marker-bar/use-remove-character-marker.hook.ts` line 97 at r1](https://reviewable.io/reviews/paranext/paranext-core/2665#-OziF8LH7dXn1cN9UJBP:-OziF8LH7dXn1cN9UJBQ:b-8p5gvf) ([raw file](https://github.com/paranext/paranext-core/blob/fa121980e87928ac4f9ac9e5147df2f140f28db7/extensions/src/platform-scripture-editor/src/character-marker-bar/use-remove-character-marker.hook.ts#L97)):*
> ```XML
>       }
> 
>       await commitSnapshot(projectId, localizedStrings[COMMIT_MESSAGE_KEY] ?? COMMIT_MESSAGE_KEY);
> ```

`commitSnapshot()` runs unconditionally before `editorRef.current?.removeCharacterMarker()`. If the editor ref is null (`?.` no-ops) or the editor silently declines a partial-nested removal (your own documented case), a "Before removing character marker" version-history point is written for an edit that never happened, with no notification. This mirrors the insert-footnote precedent, and the reachable menu paths are narrow (the partial-decline row is disabled by `isMarkerRowInert`), so it's genuinely low — but insert can't no-op, whereas removal can. Whether empty snapshots actually accumulate depends on whether `commitChanges` no-ops on an unchanged tree; worth a one-line acknowledgment at the site, or gating the snapshot on a non-null ref.

jolierabideau:
Worked through all 11 items — pushed in 4eaa9738ecf. Nine are fixed, one needs no action, and one turned out to be a bigger problem than described. Verification: `npm run typecheck` clean, `npm run lint` zero errors (the 4 remaining warnings are pre-existing, in `model-text-panel.component.tsx` and `use-effective-resource-reference-list.hook.ts`), full test suite green.

Two of these are worth reading before the rest.

### #6 was real for a different reason than stated — and it was already broken, not just fragile

Hoisting the normalized `selectionState` immediately produced a TypeScript error on the `=== 'none'` test. `markerStates` is a `Record<string, CharacterMarkerCoverageState>`, so TS types an arbitrary-key lookup as **always present** — which means `?? 'none'` was dead code *at the type level in all three places*, and TS narrowed `selectionState` to `'all' | 'partial' | undefined`.

So the divergence you spotted wasn't only a future hazard. The type system could not see the `'none'` case at any of the three sites, which is precisely why the omission at L201 was invisible. At runtime the fallback does fire — absence from the map is how coverage encodes "covers none of the selection". The lookup is now typed explicitly `| undefined` so the type matches runtime, with a comment marking that annotation as load-bearing rather than decorative.

### #10 needs no action — it self-resolved after all

You were right that r1 (`fa121980e87`) reverted `ebcd2a2c210`'s clarification; I confirmed that revert. But the prediction that it "won't self-resolve" didn't hold: `22f28862846` (the #2660 squash-merge) restored the accurate wording, and HEAD already reads "Ships in the core `assets/localization/*.json` … do not treat a definition there as evidence the string ships." Nothing to change.

---

### #1 — snapshot on the no-op paths → **fixed** (gated, not just annotated)

Took the stronger option. The removal now resolves `editorRef.current` **before** the snapshot and returns early if it's null, so the reachable no-op no longer writes a "Before removing character marker" restore point for an edit that can't happen. Your framing was the deciding factor: insert can't no-op, removal can.

The partial-nested decline is still undefended, and now says so explicitly at the site — including that a snapshot *is* still taken for it. Defending it needs the outcome signal ADR-0011 already records as absent. New test `takes no snapshot when the editor is not mounted`; I mutation-checked it (reverted the gate, watched it fail) so it's genuinely falsifiable.

### #2 + #3 + #5 — one shared module → **fixed at the root**

Went with the full extraction you called ideal, including the web view's pre-existing sites. New `editor-side-effects.utils.ts`:

- `commitVersionHistorySnapshot(projectId, message, editDescription)` — replaces **all three** copies. The `ERROR_UNIMPLEMENTED` contract now exists once.
- `notifySyncEditBlocked(localizedStrings)` — takes the strings, not a resolved message, so `SYNC_EDIT_BLOCKED_KEY` is spelled **exactly once in the codebase**. A caller can't reach the notice while spelling the key itself.

That also fixes **#5** structurally rather than by tidying: the web view lists the key via the const, and the removal hook deliberately no longer re-lists it (with a comment saying why). The "dedup either list and silently break the other" trap you described is now unreachable, not just currently-absent.

The web view's `notifySyncEditBlocked` survives as a thin `useCallback` that only binds `localizedStrings` for dependency-list stability. Recorded as **ADR-0012**, including why a hook was the wrong shape here (the insert paths call these from a message handler, not at render).

### #4 — mixed-coverage predicate → **fixed**

`isMixedCoverage` now sits beside `isEmptyCoverage` in `character-marker-coverage.utils.ts`, exactly the pattern you pointed at, consumed by both the trigger label and the menu's remove-row decision. The "deliberately NOT read off `currentCharacterMarker`" rationale moved to the helper so it's stated once.

### #7 — double-negative guard → **fixed**

Now a positive allowlist (`=== 'none' || === undefined`), with the two excluded states and their reasons named beneath it. As you noted, a future fourth enum value is now excluded by default instead of admitted into the insert path.

### #8 — duplicated rationale → **fixed**, with one correction

Trimmed to ADR-0011 pointers. One note: the state-hook memo comment you cited isn't a copy of this rationale — it explains *why the memo is keyed on coverage*, a different subject, so I left it. The actual fourth copy was the `removeCharacterMarker` option JSDoc at `character-marker-menu.utils.ts:144`. Same fix, different target. Point taken on comment volume overall.

### #9 — comment wrap → **fixed**

Reflowed. Also reflowed the `getUsj()`-staleness comment, since editing around it left a mid-phrase break.

### #11 — test mocks away its own wiring → **fixed, both halves**

You were right that dropping `projectId={projectId}` left every test passing — I verified by doing exactly that. The mock now records its options via `vi.fn<...>()`, and a new test asserts the bar forwards `projectId` into the hook; mutation-checked the same way as #1. `removeCharacterMarker.mockClear()` added, plus the new spy, in `afterEach`.

jolierabideau (commented):
> @jolierabideau made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2665)* status: 4 of 99 files reviewed, 1 unresolved discussion (waiting on katherinejensen00).

___
*[`extensions/src/platform-scripture-editor/src/character-marker-bar/use-remove-character-marker.hook.ts` line 97 at r1](https://reviewable.io/reviews/paranext/paranext-core/2665#-OziF8LH7dXn1cN9UJBP:-OzkzMeiE_JbQuzbk_E2:b-t1xij3) ([raw file](https://github.com/paranext/paranext-core/blob/fa121980e87928ac4f9ac9e5147df2f140f28db7/extensions/src/platform-scripture-editor/src/character-marker-bar/use-remove-character-marker.hook.ts#L97)):*

`commitSnapshot()` runs unconditionally before `editorRef.current?.removeCharacterMarker()`. If the editor ref is null (`?.` no-ops) or the editor silently declines a partial-nested removal (your own documented case), a "Before removing character marker" version-history point is written for an edit that never happened, with no notification. This mirrors the insert-footnote precedent, and the reachable menu paths are narrow (the partial-decline row is disabled by `isMarkerRowInert`), so it's genuinely low — but insert can't no-op, whereas removal can. Whether empty snapshots actually accumulate depends on whether `commitChanges` no-ops on an unchanged tree; worth a one-line acknowledgment at the site, or gating the snapshot on a non-null ref.
</blockquote></details>

Took the stronger option. The removal now resolves `editorRef.current` **before** the snapshot and returns early if it's null, so the reachable no-op no longer writes a "Before removing character marker" restore point for an edit that can't happen. Your framing was the deciding factor: insert can't no-op, removal can.

The partial-nested decline is still undefended, and now says so explicitly at the site — including that a snapshot _is_ still taken for it. Defending it needs the outcome signal ADR-0011 already records as absent. New test `takes no snapshot when the editor is not mounted`; I mutation-checked it (reverted the gate, watched it fail) so it's genuinely falsifiable.

jolierabideau:
Following up on the open question in item 1 — "whether empty snapshots actually accumulate depends on whether `commitChanges` no-ops on an unchanged tree." I'd fixed the item without answering that, so: checked, and it resolves to the bad branch.

`commitChanges` **does** have the no-op behavior, but it's opt-out:

```
'paratextBibleSendReceive.commitChanges': (projectId, comment, forceCommit?) => Promise<boolean>
// "Unless `forceCommit` is `true`, will only commit if there are changes/revisions detected."
```

Every pre-edit snapshot in the app passes `forceCommit: true` — insert-footnote, insert-cross-reference, character-marker removal, and Find's pre-replace snapshot. It has to, since the point of a pre-edit snapshot is to capture state *before* the edit lands, when by definition nothing has changed yet. (Find's two *post*-edit commits pass `false`, which is the consistent reading of the same flag.)

So an unguarded no-op leaves a real empty restore point, not one the backend quietly drops. That's what makes the ref gate load-bearing rather than tidy, and it means the remaining partial-decline path definitely writes an empty snapshot rather than possibly writing one.

One thing that bounds it: plain Platform.Bible's C# handler is an unimplemented stub that throws `PlatformUnimplementedException`, so core hits the `ERROR_UNIMPLEMENTED` path and writes nothing at all. This only surfaces in Paratext 10 Studio, where the patch supplies the real implementation.

No behavior change from this — the gate and the documented partial-decline case were already the right fix, and detecting the decline still needs an editor outcome signal that doesn't exist. But the mechanism is now recorded at the site instead of asserted, in 2111274e6dd, so the next reader doesn't have to re-derive it or assume the backend dedupes.

katherinejensen00 (approved):
Thanks for all your work to make the code review changes! Approved

> @katherinejensen00 reviewed 95 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2665#-:-Ozl9cr-5NlzsEhojUFT:bor9pge)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2667, opened 2026-08-11
Keep the Button press nudge from clobbering caller translates

imnasnainaec:
## Summary

`Button`'s pressed style nudges the button down 1px with `tw:active:not-aria-[haspopup]:translate-y-px`. Every Tailwind v4 translate utility writes the same `--tw-translate-y` custom property and then sets the `translate` shorthand from it, so the nudge *replaces* a caller's translate rather than adding to it — and `:active` outranks the caller's unprefixed utility. A `Button` centered with `tw:top-1/2 tw:-translate-y-1/2` — a common way to place an icon button inside an input — therefore jumps down half its own height for as long as it is held.

Compiled both utilities with this repo's Tailwind to confirm the collision:

```css
.tw\:-translate-y-1\/2 {
  --tw-translate-y: calc(calc(1 / 2 * 100%) * -1);
  translate: var(--tw-translate-x) var(--tw-translate-y);
}
.tw\:active\:not-aria-\[haspopup\]\:translate-y-px {
  &:active { &:not(*[aria-haspopup]) {
    --tw-translate-y: 1px;
    translate: var(--tw-translate-x) var(--tw-translate-y);
  } }
}
```

Moving the nudge to the `transform` property fixes it for every caller at once: per css-transforms-2 the browser applies `translate`, `rotate` and `scale` first and `transform` after, so the two compose instead of one overwriting the other. Callers get centering-plus-press with no workaround and no comment explaining one. `tw:transition-all` is already in the same base string, so the press animation is unchanged.

### Why review this

Not part of the current epic. Found while reviewing sillsdev/interlinearizer-extension#214, where two icon buttons jumped mid-click and were worked around downstream — the cause is here, and every consumer that centers a `Button` with a translate will hit it independently. `lib/platform-bible-react/src/stories/shadcn-ui/tooltip.stories.tsx:989` is a live instance in our own Storybook, so we currently demonstrate the broken idiom to anyone copying the pattern.

## Changes

- `button.tsx` — `tw:active:not-aria-[haspopup]:translate-y-px` → `tw:active:not-aria-[haspopup]:transform-[translateY(1px)]`, with a `// CUSTOM:` annotation that names what still collides (see the tradeoff note below). The original line arrived unannotated with the preset apply in a61ca3b913c, so without the annotation the next `shadcn apply` would silently revert this.
- `button.tsx` — annotated the pre-existing `--radius-md` → `--tw-radius-md` renames in the `xs`/`sm`/`icon-xs`/`icon-sm` variants, which came in unannotated with 55256831f97. `src/index.css` imports the theme with `prefix(tw)`, which emits `@theme` variables prefixed, so the boilerplate's unprefixed name resolves to nothing and those sizes lose their rounding entirely. Same reasoning as above: unannotated, so a preset apply reverts them silently.
- `button.test.tsx` (new) — two tests, one per silently-revertible customization: the nudge stays on `transform` and is not an `active:`-modified translate utility, and the clamped sizes keep `var(--tw-radius-md)`. Both fail when the corresponding customization is reverted (checked).
- `button.stories.tsx` — new `PressNudgeWithCallerTranslate` story: an icon button centered with `tw:-translate-y-1/2` inside the package's own `Input`, the exact idiom the old nudge broke.

No call sites needed changing: callers that center with translate utilities are fixed by this, and the one `-translate-y-1/2` clear-button in `platform-scripture`'s find component is a raw `<button>`, not this one.

## Testing

- [x] `npm run typecheck` in `lib/platform-bible-react` — passes
- [x] `prettier --check` on the changed files — passes (`shadcn-ui/` is ESLint-ignored)
- [x] Recompiled the prefixed class with the repo's Tailwind and confirmed it emits `transform: translateY(1px)` and no longer touches `--tw-translate-y`
- [x] `vitest --project=unit button.test.tsx` — 2 passed; re-run with both customizations reverted to confirm both fail
- [x] `vitest --project=storybook button.stories.tsx` — 17 passed, including the new story
- [ ] Visual check still wanted: open `Shadcn/Button` → `PressNudgeWithCallerTranslate`, hold the clear button, confirm it stays vertically centered and still nudges 1px

## Notes for the reviewer

- **The tradeoff this makes.** Compiling against the repo's Tailwind: `translate-*` writes `translate`, `scale-*` writes `scale`, and plain `rotate-*` writes `rotate` — all separate properties, so all three now compose with the nudge. What writes `transform` itself, and so still collides with it, is `skew-*`, `rotate-x/y/z-*`, the `transform`/`transform-gpu`/`transform-[…]` utilities, and inline `style={{ transform }}` (which now wins, since inline styles beat classes). No `Button` in this repo carries any of those. That is the right way round — the common idiom composes and the rare one loses — and the `// CUSTOM:` block records it. One second-order effect: because the individual properties are applied first, the 1px is measured in the button's own space, so a caller `rotate-*` or `scale-*` would rotate or scale the nudge.
- Not rebuilding `dist/` here; happy to add a rebuild if the PT team wants this out to consumers with this PR rather than the next build.
- interlinearizer's `ArcOverlay` works around the old behavior downstream and should move back to translate utilities once this lands.

## Why the tests are unit tests and not a play function

The pressed geometry is the thing worth testing, and a browser play function looks like where it could run for real — the `storybook` vitest project runs stories in Chromium and this file is already in it. It cannot work: Storybook's play-context `userEvent` dispatches synthetic events, and `:active` is a browser input state that untrusted events do not set. Measured it — a `[MouseLeft>]` press in that project leaves `button.matches(':active') === false` and the position delta at exactly `0`, with or without the fix, so the assertion would be non-falsifiable in the direction that matters. (`@storybook/addon-vitest` imports only `server` from `@vitest/browser/context`, not its CDP-backed `userEvent`.) Real geometry coverage would need a browser-mode test outside the storybook project, which is a new vitest project and out of scope.

What is left to assert deterministically is the class itself, and jsdom does that faster and without Playwright. So the split is: unit tests pin the classes, story is the human check for the geometry.

## Risk Level

Low — one utility class, no API change. Worst case is the 1px press nudge not rendering for some caller, which is cosmetic and visible in Storybook.

## AI Involvement

AI-assisted. Claude diagnosed the custom-property collision, verified it by compiling both utilities against the repo's own Tailwind, wrote the class change and the `// CUSTOM:` annotations, and added the tests and story. Opened as a draft pending my own review of the diagnosis and the visual check above.

Devin review: https://app.devin.ai/review/paranext/paranext-core/pull/2667

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2667)

### DIFF
--- a/lib/platform-bible-react/src/components/shadcn-ui/button.test.tsx
+++ b/lib/platform-bible-react/src/components/shadcn-ui/button.test.tsx
@@ -5,32 +5,27 @@ import '@testing-library/jest-dom';
 import { Button } from '@/components/shadcn-ui/button';
 
 describe('Button', () => {
-  // Callers position icon buttons with translate utilities — `tw:-translate-y-1/2` to center one
-  // inside an input's end padding, for example. Every Tailwind translate utility writes the same
-  // `--tw-translate-y` custom property, so a pressed-state translate replaces the caller's
-  // centering rather than adding to it, and `:active` outranks the caller's unprefixed utility:
-  // the button drops half its own height for as long as it is held. Keeping the nudge on
-  // `transform` lets the two compose, since the browser applies `translate` before `transform`.
-  //
-  // jsdom cannot enter `:active` and computes no Tailwind CSS, so these pin the class rather than
-  // the rendered geometry. That is enough to fail if a future `shadcn apply` reverts the utility,
-  // which is the regression worth catching. `PressNudgeWithCallerTranslate` in `button.stories.tsx`
-  // is the manual check for the geometry itself.
+  // jsdom computes no Tailwind CSS and cannot enter `:active`, so these pin the class strings a
+  // `shadcn apply` would revert. `PressNudgeWithCallerTranslate` in `button.stories.tsx` is the
+  // manual check for the geometry they stand in for.
+
   it('puts the pressed-state nudge on transform, not on a translate utility', () => {
     render(<Button>Press me</Button>);
 
     const { className } = screen.getByRole('button');
 
-    expect(className).toContain('tw:active:not-aria-[haspopup]:transform-[translateY(1px)]');
+    expect(className).toMatch(/active:\S*transform-\[translateY/);
+    // The trailing hyphen is load-bearing: it matches the reverted `translate-y-px` and not
+    // `translateY(1px)`.
     expect(className).not.toMatch(/active:\S*translate-/);
   });
 
-  it('keeps a caller translate alongside the pressed-state nudge', () => {
-    render(<Button className="tw:-translate-y-1/2">Press me</Button>);
+  it('keeps the prefixed radius variable on the sizes that clamp their radius', () => {
+    render(<Button size="xs">Press me</Button>);
 
     const { className } = screen.getByRole('button');
 
-    expect(className).toContain('tw:-translate-y-1/2');
-    expect(className).toContain('tw:active:not-aria-[haspopup]:transform-[translateY(1px)]');
+    expect(className).toContain('var(--tw-radius-md)');
+    expect(className).not.toMatch(/var\(--radius-md\)/);
   });
 });
--- a/lib/platform-bible-react/src/components/shadcn-ui/button.test.tsx
+++ b/lib/platform-bible-react/src/components/shadcn-ui/button.test.tsx
@@ -1,0 +1,36 @@
+// @vitest-environment jsdom
+import { describe, it, expect } from 'vitest';
+import { render, screen } from '@testing-library/react';
+import '@testing-library/jest-dom';
+import { Button } from '@/components/shadcn-ui/button';
+
+describe('Button', () => {
+  // Callers position icon buttons with translate utilities — `tw:-translate-y-1/2` to center one
+  // inside an input's end padding, for example. Every Tailwind translate utility writes the same
+  // `--tw-translate-y` custom property, so a pressed-state translate replaces the caller's
+  // centering rather than adding to it, and `:active` outranks the caller's unprefixed utility:
+  // the button drops half its own height for as long as it is held. Keeping the nudge on
+  // `transform` lets the two compose, since the browser applies `translate` before `transform`.
+  //
+  // jsdom cannot enter `:active` and computes no Tailwind CSS, so these pin the class rather than
+  // the rendered geometry. That is enough to fail if a future `shadcn apply` reverts the utility,
+  // which is the regression worth catching. `PressNudgeWithCallerTranslate` in `button.stories.tsx`
+  // is the manual check for the geometry itself.
+  it('puts the pressed-state nudge on transform, not on a translate utility', () => {
+    render(<Button>Press me</Button>);
+
+    const { className } = screen.getByRole('button');
+
+    expect(className).toContain('tw:active:not-aria-[haspopup]:transform-[translateY(1px)]');
+    expect(className).not.toMatch(/active:\S*translate-/);
+  });
+
+  it('keeps a caller translate alongside the pressed-state nudge', () => {
+    render(<Button className="tw:-translate-y-1/2">Press me</Button>);
+
+    const { className } = screen.getByRole('button');
+
+    expect(className).toContain('tw:-translate-y-1/2');
+    expect(className).toContain('tw:active:not-aria-[haspopup]:transform-[translateY(1px)]');
+  });
+});
--- a/lib/platform-bible-react/src/components/shadcn-ui/button.tsx
+++ b/lib/platform-bible-react/src/components/shadcn-ui/button.tsx
@@ -37,6 +37,11 @@ const buttonVariants = cva(
       size: {
         default:
           'tw:h-8 tw:gap-1.5 tw:px-2.5 tw:has-data-[icon=inline-end]:pe-2 tw:has-data-[icon=inline-start]:ps-2',
+        // CUSTOM: Renamed 'var(--radius-md)' to 'var(--tw-radius-md)' in the rounding utilities for
+        // the 'xs', 'sm', 'icon-xs', and 'icon-sm' sizes. 'src/index.css' imports Tailwind with
+        // 'prefix(tw)', which emits every '@theme' variable under a 'tw' prefix, so the
+        // boilerplate's unprefixed name resolves to nothing. That invalidates the whole 'min()' and
+        // leaves those four sizes with square corners.
         xs: 'tw:h-6 tw:gap-1 tw:rounded-[min(var(--tw-radius-md),10px)] tw:px-2 tw:text-xs tw:in-data-[slot=button-group]:rounded-lg tw:has-data-[icon=inline-end]:pe-1.5 tw:has-data-[icon=inline-start]:ps-1.5 tw:[&_svg:not([class*=size-])]:size-3',
         sm: 'tw:h-7 tw:gap-1 tw:rounded-[min(var(--tw-radius-md),12px)] tw:px-2.5 tw:text-[0.8rem] tw:in-data-[slot=button-group]:rounded-lg tw:has-data-[icon=inline-end]:pe-1.5 tw:has-data-[icon=inline-start]:ps-1.5 tw:[&_svg:not([class*=size-])]:size-3.5',
         lg: 'tw:h-9 tw:gap-1.5 tw:px-2.5 tw:has-data-[icon=inline-end]:pe-2 tw:has-data-[icon=inline-start]:ps-2',
--- a/lib/platform-bible-react/src/components/shadcn-ui/button.tsx
+++ b/lib/platform-bible-react/src/components/shadcn-ui/button.tsx
@@ -13,12 +13,12 @@ import { cn } from '@/utils/shadcn-ui/utils';
 const buttonVariants = cva(
   // CUSTOM: Added 'pr-twp' at the front of the base class string to apply Platform.Bible's
   // Tailwind CSS scope isolation. All Button instances inherit this via buttonVariants.
-  // CUSTOM: Changed the pressed-state nudge from 'tw:active:not-aria-[haspopup]:translate-y-px' to
-  // 'tw:active:not-aria-[haspopup]:transform-[translateY(1px)]' so it composes with caller
-  // positioning. Every Tailwind translate utility writes the same '--tw-translate-y' custom
-  // property, so the nudge replaced rather than added to a caller's translate: a Button centered
-  // with 'tw:-translate-y-1/2' jumped down half its own height while held. 'transform' is a separate
-  // property that the browser applies after 'translate', so the two now stack.
+  // CUSTOM: Moved the pressed-state nudge off 'tw:active:not-aria-[haspopup]:translate-y-px' onto
+  // 'transform'. Every Tailwind translate utility writes the same '--tw-translate-y', so the nudge
+  // replaced a caller's centering translate rather than adding to it. The browser applies
+  // 'translate', 'rotate' and 'scale' before 'transform', so those compose (a caller 'rotate-*' or
+  // 'scale-*' does rotate or scale the 1px); caller 'skew-*', 'rotate-x/y/z-*', 'transform*'
+  // utilities and inline style transforms still override it.
   'pr-twp tw:group/button tw:inline-flex tw:shrink-0 tw:items-center tw:justify-center tw:rounded-lg tw:border tw:border-transparent tw:bg-clip-padding tw:text-sm tw:font-medium tw:whitespace-nowrap tw:transition-all tw:outline-none tw:select-none tw:focus-visible:border-ring tw:focus-visible:ring-3 tw:focus-visible:ring-ring/50 tw:active:not-aria-[haspopup]:transform-[translateY(1px)] tw:disabled:pointer-events-none tw:disabled:opacity-50 tw:aria-invalid:border-destructive tw:aria-invalid:ring-3 tw:aria-invalid:ring-destructive/20 tw:dark:aria-invalid:border-destructive/50 tw:dark:aria-invalid:ring-destructive/40 tw:[&_svg]:pointer-events-none tw:[&_svg]:shrink-0 tw:[&_svg:not([class*=size-])]:size-4',
   {
     variants: {
@@ -38,10 +38,10 @@ const buttonVariants = cva(
         default:
           'tw:h-8 tw:gap-1.5 tw:px-2.5 tw:has-data-[icon=inline-end]:pe-2 tw:has-data-[icon=inline-start]:ps-2',
         // CUSTOM: Renamed 'var(--radius-md)' to 'var(--tw-radius-md)' in the rounding utilities for
-        // the 'xs', 'sm', 'icon-xs', and 'icon-sm' sizes. 'src/index.css' imports Tailwind with
-        // 'prefix(tw)', which emits every '@theme' variable under a 'tw' prefix, so the
-        // boilerplate's unprefixed name resolves to nothing. That invalidates the whole 'min()' and
-        // leaves those four sizes with square corners.
+        // the sizes that clamp their radius. 'src/index.css' imports Tailwind with 'prefix(tw)',
+        // which emits every '@theme' variable under a 'tw' prefix, so the boilerplate's unprefixed
+        // name resolves to nothing. That invalidates the whole 'min()' and leaves those sizes with
+        // square corners.
         xs: 'tw:h-6 tw:gap-1 tw:rounded-[min(var(--tw-radius-md),10px)] tw:px-2 tw:text-xs tw:in-data-[slot=button-group]:rounded-lg tw:has-data-[icon=inline-end]:pe-1.5 tw:has-data-[icon=inline-start]:ps-1.5 tw:[&_svg:not([class*=size-])]:size-3',
         sm: 'tw:h-7 tw:gap-1 tw:rounded-[min(var(--tw-radius-md),12px)] tw:px-2.5 tw:text-[0.8rem] tw:in-data-[slot=button-group]:rounded-lg tw:has-data-[icon=inline-end]:pe-1.5 tw:has-data-[icon=inline-start]:ps-1.5 tw:[&_svg:not([class*=size-])]:size-3.5',
         lg: 'tw:h-9 tw:gap-1.5 tw:px-2.5 tw:has-data-[icon=inline-end]:pe-2 tw:has-data-[icon=inline-start]:ps-2',
--- a/lib/platform-bible-react/src/components/shadcn-ui/button.tsx
+++ b/lib/platform-bible-react/src/components/shadcn-ui/button.tsx
@@ -13,7 +13,13 @@ import { cn } from '@/utils/shadcn-ui/utils';
 const buttonVariants = cva(
   // CUSTOM: Added 'pr-twp' at the front of the base class string to apply Platform.Bible's
   // Tailwind CSS scope isolation. All Button instances inherit this via buttonVariants.
-  'pr-twp tw:group/button tw:inline-flex tw:shrink-0 tw:items-center tw:justify-center tw:rounded-lg tw:border tw:border-transparent tw:bg-clip-padding tw:text-sm tw:font-medium tw:whitespace-nowrap tw:transition-all tw:outline-none tw:select-none tw:focus-visible:border-ring tw:focus-visible:ring-3 tw:focus-visible:ring-ring/50 tw:active:not-aria-[haspopup]:translate-y-px tw:disabled:pointer-events-none tw:disabled:opacity-50 tw:aria-invalid:border-destructive tw:aria-invalid:ring-3 tw:aria-invalid:ring-destructive/20 tw:dark:aria-invalid:border-destructive/50 tw:dark:aria-invalid:ring-destructive/40 tw:[&_svg]:pointer-events-none tw:[&_svg]:shrink-0 tw:[&_svg:not([class*=size-])]:size-4',
+  // CUSTOM: Changed the pressed-state nudge from 'tw:active:not-aria-[haspopup]:translate-y-px' to
+  // 'tw:active:not-aria-[haspopup]:transform-[translateY(1px)]' so it composes with caller
+  // positioning. Every Tailwind translate utility writes the same '--tw-translate-y' custom
+  // property, so the nudge replaced rather than added to a caller's translate: a Button centered
+  // with 'tw:-translate-y-1/2' jumped down half its own height while held. 'transform' is a separate
+  // property that the browser applies after 'translate', so the two now stack.
+  'pr-twp tw:group/button tw:inline-flex tw:shrink-0 tw:items-center tw:justify-center tw:rounded-lg tw:border tw:border-transparent tw:bg-clip-padding tw:text-sm tw:font-medium tw:whitespace-nowrap tw:transition-all tw:outline-none tw:select-none tw:focus-visible:border-ring tw:focus-visible:ring-3 tw:focus-visible:ring-ring/50 tw:active:not-aria-[haspopup]:transform-[translateY(1px)] tw:disabled:pointer-events-none tw:disabled:opacity-50 tw:aria-invalid:border-destructive tw:aria-invalid:ring-3 tw:aria-invalid:ring-destructive/20 tw:dark:aria-invalid:border-destructive/50 tw:dark:aria-invalid:ring-destructive/40 tw:[&_svg]:pointer-events-none tw:[&_svg]:shrink-0 tw:[&_svg:not([class*=size-])]:size-4',
   {
     variants: {
       variant: {
--- a/lib/platform-bible-react/src/stories/shadcn-ui/button.stories.tsx
+++ b/lib/platform-bible-react/src/stories/shadcn-ui/button.stories.tsx
@@ -1,5 +1,6 @@
 import type { Meta, StoryObj } from '@storybook/react-vite';
 import { expect, fn } from 'storybook/test';
+import { X } from 'lucide-react';
 import { Button } from '@/components/shadcn-ui/button';
 
 const meta: Meta<typeof Button> = {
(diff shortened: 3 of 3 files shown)


lib/platform-bible-react/src/components/shadcn-ui/button.tsx
```diff
@@ -13,7 +13,13 @@ import { cn } from '@/utils/shadcn-ui/utils';
...
+  // 'tw:active:not-aria-[haspopup]:transform-[translateY(1px)]' so it composes with caller
+  // positioning. Every Tailwind translate utility writes the same '--tw-translate-y' custom
+  // property, so the nudge replaced rather than added to a caller's translate: a Button centered
+  // with 'tw:-translate-y-1/2' jumped down half its own height while held. 'transform' is a separate
+  // property that the browser applies after 'translate', so the two now stack.
+  'pr-twp tw:group/button tw:inline-flex tw:shrink-0 tw:items-center tw:justify-center tw:rounded-lg tw:border tw:border-transparent tw:bg-clip-padding tw:text-sm tw:font-medium tw:whitespace-nowrap tw:transition-all tw:outline-none tw:select-none tw:focus-visible:border-ring tw:focus-visible:ring-3 tw:focus-visible:ring-ring/50 tw:active:not-aria-[haspopup]:transform-[translateY(1px)] tw:disabled:pointer-events-none tw:disabled:opacity-50 tw:aria-invalid:border-destructive tw:aria-invalid:ring-3 tw:aria-invalid:ring-destructive/20 tw:dark:aria-invalid:border-destructive/50 tw:dark:aria-invalid:ring-destructive/40 tw:[&_svg]:pointer-events-none tw:[&_svg]:shrink-0 tw:[&_svg:not([class*=size-])]:size-4',
```
imnasnainaec, line 22: 🤖 **Claude AI** · **[Scope]** · Decision needed (warning)

Should `lib/platform-bible-react/dist/` be rebuilt and committed in this PR? Without it neither the new class nor its compiled CSS rule reaches the renderer, the in-repo extensions, or the external consumer (sillsdev/interlinearizer-extension#214) that reported the bug.

The PR's primary goal does not reach any consumer without a `dist/` rebuild, and the stated justification for skipping it ("the committed build is already stale for this file") does not hold.

`dist/` is deliberately committed — `lib/platform-bible-react/.gitignore:6` reads `# dist # We want these commited so developers don't need to build the package.` The renderer app plus every extension resolve the library through `file:` deps into `package.json`'s `exports` → `dist/index.js`.

The committed `dist/index.js` still ships both the old class on the button **and** the old compiled CSS rule:

```css
.tw\:active\:not-aria-\[haspopup\]\:translate-y-px:active:not([aria-haspopup]){--tw-translate-y:1px;translate:var(--tw-translate-x) var(--tw-translate-y)}
```

Because PBR's Tailwind scans only PBR's own source and inlines the compiled stylesheet into `dist/index.js`, the new `transform-[translateY(1px)]` utility has no CSS rule anywhere until `dist` is rebuilt — so consumers keep the exact bug this PR fixes.

Verified directly: `grep -c 'translate-y-px' dist/index.js` → 1; `grep -c 'translateY(1px)' dist/index.js` → 0.

On the staleness claim: the last PBR source commit before this branch (824065f9c1a) touched only `*.stories.tsx`, which are not part of the lib build, and the last commit that did change bundled source (ca07ccadc6f) shipped a rebuilt `dist/` in the same commit. **`dist` was in sync at the PR base; this PR is what makes it stale.**

Notably, no in-repo product code centers a `Button` with a translate utility — the only `Button` + `translate-*` hits are `button.stories.tsx:314` (this PR's new story), `tooltip.stories.tsx:992` (a story), and `button.test.tsx:29`. `search-bar.component.tsx:80-86` centers its clear Button with `tw:inset-y-0 tw:my-auto` instead. So the entire practical benefit of this fix accrues to out-of-repo consumers, who can only ever consume the published `dist/` — they have no path to the source. That makes the rebuild the difference between this PR shipping a fix and shipping a no-op.

**Evidence:** `.context/standards/Code-Style-Guide.md`, "Rebuild `platform-bible-react` After Editing Its Sources": "Extensions (and the renderer) import `platform-bible-react` as a package — type resolution goes through the committed build artifact ... The rebuilt `dist/` changes are normally part of the same PR that changes the library source."

**Cross-ref:** `package.json:143` and `extensions/package.json:84` wire consumers to `file:` → `dist/`

</details>

lib/platform-bible-react/src/stories/shadcn-ui/button.stories.tsx
```diff
@@ -298,3 +299,32 @@ export const ComplexInteraction: Story = {
     });
   },
 };
+
+export const PressNudgeWithCallerTranslate: Story = {
+  render: () => (
+    <div className="tw:relative tw:w-64">
+      <input
```
imnasnainaec: 🤖 **Claude AI** · **[UX]** · Decision needed (warning)

Should the story render `<Input className="tw:w-full tw:ps-2 tw:pe-8" …>` from `@/components/shadcn-ui/input` (as the analogous `SearchBarClearButton` story already does), or is there a reason to keep the raw element here?

The new story hand-rolls a raw `<input>` with copied Tailwind classes instead of using the package's own `Input` component. The copy has already drifted from the real thing:

- `tw:border-border` where `Input` uses `tw:border-input`
- `tw:bg-background` where `Input` uses `tw:bg-transparent tw:dark:bg-input/30`
- omits `Input`'s `tw:focus-visible:border-ring tw:focus-visible:ring-3 tw:focus-visible:ring-ring/50` entirely

So in dark mode the story's field renders on the page surface rather than the input surface, and focusing it shows the browser default outline rather than the design-system ring. For a story whose entire job is to show a Button correctly positioned *inside an input*, the surrounding chrome should be the actual input, and it will then track future `Input` height/padding changes instead of silently going stale.

`Input` deliberately drops `tw:w-full` so callers control width, so the story only needs to add `tw:w-full` plus its padding overrides.

**Cross-ref:** `lib/platform-bible-react/src/stories/shadcn-ui/tooltip.stories.tsx:982` — the existing `SearchBarClearButton` story composes this exact pattern using the real `Input` component

</details>

lib/platform-bible-react/src/components/shadcn-ui/button.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+    expect(className).toContain('tw:active:not-aria-[haspopup]:transform-[translateY(1px)]');
+    expect(className).not.toMatch(/active:\S*translate-/);
+  });
+
+  it('keeps a caller translate alongside the pressed-state nudge', () => {
+    render(<Button className="tw:-translate-y-1/2">Press me</Button>);
+
+    const { className } = screen.getByRole('button');
+
+    expect(className).toContain('tw:-translate-y-1/2');
```
imnasnainaec: 🤖 **Claude AI** · **[Tests]** · Decision needed (warning)

Should test 2 be deleted as dead weight (test 1 already catches the revert), kept purely as executable documentation of the caller-`className` scenario, or promoted to the existing `storybook` browser project where the composed geometry can actually be measured?

The second test adds no regression coverage over the first. Ran the Revert Test the Testing-Guide requires — swapped `transform-[translateY(1px)]` back to `tw:active:not-aria-[haspopup]:translate-y-px` in `button.tsx` and re-ran the file. Test 2 did fail, but **only on line 34**, which is a verbatim duplicate of line 24 in test 1. **Line 33 passed against the buggy implementation.**

The reverted `className` vitest printed was:

```
... tw:active:not-aria-[haspopup]:translate-y-px ... tw:-translate-y-1/2
```

Both classes survive `cn` because tailwind-merge only collapses classes with identical modifier sets, and `active:not-aria-[haspopup]:` differs from no modifier. So line 33 asserts a tailwind-merge property that held before the bug, during the bug, and after the fix — it is "Framework behavior / testing someone else's code" per the guide's *What NOT to Test* table, not Button behavior.

The composition claim the test's name makes ("keeps a caller translate alongside the pressed-state nudge") is genuinely untestable in jsdom, which computes no Tailwind CSS and cannot enter `:active`.

Verified empirically, not inferred. `button.tsx` was restored to the PR head afterward; the working tree is clean.

**Evidence:** `.context/standards/Testing-Guide.md` — "What NOT to Test" (Framework behavior: "Testing someone else's code"); "Test Quality Checklist" (Falsifiable — test fails when the implementation is broken)

**Cross-ref:** `lib/platform-bible-react/src/utils/shadcn-ui/utils.ts:8` — `extendTailwindMerge({ prefix: 'tw' })`; the merge behavior line 33 exercises lives in tailwind-merge, and `utils.test.ts` already covers `cn` directly

</details>

lib/platform-bible-react/src/components/shadcn-ui/button.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+    expect(className).toContain('tw:active:not-aria-[haspopup]:transform-[translateY(1px)]');
+  });
+});
```
imnasnainaec, line 31: 🤖 **Claude AI** · **[Tests]** · Decision needed (suggestion)

Add a `--tw-radius-md` pin to this file now, or accept that only one of the two customizations documented in this diff has revert protection?

Asymmetric regression coverage for the two silently-revertible customizations this PR touches. The PR's justification for a class-assertion unit test is "it will fail if a future `shadcn apply` reverts the utility" — and in the same diff it adds a CUSTOM comment (`button.tsx:40-44`) documenting the `var(--radius-md)` → `var(--tw-radius-md)` rename, which is revertible by exactly the same mechanism and whose failure mode is arguably more visible (square corners on the `xs`/`sm`/`icon-xs`/`icon-sm` sizes) yet entirely silent. Nothing pins it.

The same jsdom test already in this file extends to it in three lines: render `<Button size="xs" />`, then assert the `className` contains `var(--tw-radius-md)` and does not match `/var\(--radius-md\)/`. That assertion has the identical falsifiability profile to the nudge assertions — no new infrastructure, no animation timing.

The same `var(--tw-radius-md)` pattern also appears in `select.tsx`, `toggle.tsx`, and `toggle-group.tsx`, so the exposure is wider than Button — but those files are outside this PR and a Button-only pin is a reasonable first step.

**Evidence:** `.context/standards/Testing-Guide.md` — "When to Add Tests": "you need tests that would actually catch regressions"

**Cross-ref:** `lib/platform-bible-react/src/components/shadcn-ui/button.tsx:40` — CUSTOM comment added by this PR describing the rename, with no corresponding test

</details>

lib/platform-bible-react/src/components/shadcn-ui/button.tsx
```diff
@@ -31,6 +37,11 @@ const buttonVariants = cva(
       size: {
         default:
           'tw:h-8 tw:gap-1.5 tw:px-2.5 tw:has-data-[icon=inline-end]:pe-2 tw:has-data-[icon=inline-start]:ps-2',
+        // CUSTOM: Renamed 'var(--radius-md)' to 'var(--tw-radius-md)' in the rounding utilities for
```
imnasnainaec, line 40: 🤖 **Claude AI** · **[Clarity]** **[Scope]** · Decision needed (suggestion)

Should `icon-xs`/`icon-sm` get their own short `// CUSTOM:` note so the annotation sits immediately before each changed entry, or is the single block above `xs` acceptable given the four entries share one cause?

This annotation sits immediately above `xs` but documents four sizes, two of which are not adjacent to it: `lg` (line 47) and `icon` (line 48) sit between `sm` and `icon-xs`. Someone editing `icon-sm` on lines 51-52, or running a `shadcn apply` diff scoped to that entry, sees `--tw-radius-md` with no nearby explanation and no signal that reverting it is a regression — which is the exact failure mode this PR is trying to prevent.

The prose also enumerates the member list ("the `xs`, `sm`, `icon-xs`, and `icon-sm` sizes" … "those four sizes"), which goes stale silently if a size is added or removed; "the sizes that clamp their radius" is stable phrasing.

Grouping the four together in one block is genuinely nicer to read than four copies; the tradeoff is discoverability at the two distant edit sites.

**On whether this belongs in the PR at all** — it does. It is comment-only (zero runtime effect; the renamed variables landed earlier in 55256831f97 and are unchanged here), it lives in the same shadcn file the PR already has open, and it remediates a violation of a repo rule that admits no exceptions. Splitting it out would cost more review cycles than it saves. The claim itself checks out: `src/index.css` imports Tailwind with `prefix(tw)`, and `dist/index.js` confirms the emitted variable is `--tw-radius-md`. The one cost is squash-message clarity — the merged commit will describe two unrelated things.

**Evidence:** `.claude/rules/code-quality/shadcn-discipline.md`: annotation "placed immediately before the changed code"; comment-rules: never restate in prose a count or member list that code declares

**Cross-ref:** `lib/platform-bible-react/src/components/shadcn-ui/context-menu.tsx:76,149,154` — the same z-index fix is annotated separately at each of its three non-adjacent sites rather than once

</details>

imnasnainaec (commented):
## 🤖 [Claude AI Review] — PR #2667

Reviewer panel: `security`, `scope`, `ux`, `tests`, `clarity` (default panel, auto-narrowed — no C# files, so `contracts` and `architecture` were dropped).

The diagnosis in the PR body is correct and I re-verified the mechanism independently: `translate-*` compiles to the `translate` property and `transform-[…]` to `transform`, the browser applies `translate` before `transform`, and the two now compose. The common case (Button with no caller translate) is visually identical — same 1px offset, and both properties establish a stacking context and a containing block, so nothing changes there. `tw:transition-all` covers both properties, so the press animation is unchanged.

Two findings were verified by running things rather than reading them, and both hold.

|-------------|----------|---------|------------|
| Security    | 0        | 0       | 0          |
| Scope       | 0        | 1       | 3          |
| UX          | 0        | 1       | 3          |
| Tests       | 0        | 1       | 3          |
| Clarity     | 0        | 0       | 5          |
| **Total distinct** | **0** | **3** | **8**      |

Five findings needing a decision are posted inline. The rest are below.

### Headline: the fix does not reach any consumer without a `dist/` rebuild

This is the inline comment on `button.tsx:22`, repeated here because it decides whether the PR ships a fix or a no-op. `dist/` is deliberately committed (`.gitignore:6`), and `dist/index.js` still contains `translate-y-px` (1 hit) and zero hits for `translateY(1px)`. The PR body's justification for skipping the rebuild — "the committed build is already stale for this file" — does not hold: the last pre-branch commit touching PBR source (824065f9c1a) changed only `*.stories.tsx`, which are not part of the lib build, and the last commit that did change bundled source (ca07ccadc6f) shipped a rebuilt `dist/` in the same commit. `dist` was in sync at the PR base. Since no in-repo product code centers a Button with a translate utility, the entire practical benefit accrues to out-of-repo consumers, who can only consume `dist/`.

---

### Informational findings

🤖 **Claude AI** · **[Scope]** **[UX]** **[Clarity]** (suggestion) — `button.tsx:16-21`

Three overlapping observations on the new `// CUSTOM:` block, pulling in opposite directions — worth deciding as one.

*It under-claims one thing:* the comment explains the collision that was removed but not the one that was introduced, and the new one is subtler. In Tailwind v4, `rotate-*` and `skew-*` write the `transform` property, not a dedicated one — the compiled CSS confirms it (`transform:var(--tw-rotate-x,) … var(--tw-skew-y,)`) — whereas `translate-*` compiles to `translate:` and `scale-*` to `scale:`. So the fix trades a clobber of caller `translate-*` for a clobber of caller `rotate-*`/`skew-*`/`transform-gpu`/arbitrary `transform-[…]`. Second, because CSS applies `translate`/`rotate`/`scale` before `transform`, the nudge is now expressed in the element's *local* space: a Button with `tw:rotate-90` would nudge sideways instead of down, one with `tw:scale-50` would nudge 0.5px. Neither was true of the old nudge.

No caller in the repo hits any of this today — the only `tw:transform` uses are bare, on non-Button icons (`search-bar.component.tsx:61`, `direction.stories.tsx:102,112`), and no Button carries `scale-*`/`rotate-*` — so the trade is a clear net win. But the constraint is now invisible at the call site, and the `// CUSTOM:` block is the durable record a future `shadcn apply` reviewer will consult.

*It over-explains another:* at 6 lines this is the longest `// CUSTOM:` block in the whole `shadcn-ui/` directory (every other one is 2-4 lines, with a single 5-line outlier in `alert.tsx`), and it says the same thing twice — the concrete repro ("jumped down half its own height while held") and the closing mechanism sentence ("`transform` is a separate property that the browser applies after `translate`") are two statements of one fact. "so the two now stack" also reads as diff narration rather than a description of the code as it stands.

Suggested resolution: cut the doubled explanation, spend the reclaimed line naming what composes (`translate`, `scale`) and what does not (`rotate`, `skew`, `transform-*`, inline `style` transform). Net length roughly unchanged, strictly more information.

*Cross-ref:* `context-menu.tsx:76` — comparable non-obvious CSS fix annotated in 2 lines including the ticket number (PT-3877)

---

🤖 **Claude AI** · **[Scope]** **[Tests]** (suggestion) — `button.stories.tsx:303`

The success criterion — "pressed Button stays vertically centered while held" — has no automated verification, and the PR's framing that the geometry "could not run for real" is stronger than the constraint actually is.

The repo already runs stories against real Chromium: `vitest.config.ts:38-60` defines a `storybook` project with `browser.provider: 'playwright'`, this file's meta carries `tags: ['autodocs', 'test']`, and eight `play`-driven stories in this very file already run there — two of which already assert Button classes. So the browser lane exists and this component is already in it.

The cited Testing-Guide caution ("Avoid Animation-Sensitive Assertions in Story Interaction Tests") is accurate for the naive approach — `toHaveStyle` mid-transition would flake — but it is a caution about *how* to assert, not proof that the geometry cannot be asserted. A held press is reachable via `userEvent.pointer({ keys: '[MouseLeft>]', target: button })`, and the transition race has a deterministic seam rather than a sleep: `await Promise.all(button.getAnimations().map((a) => a.finished))` before reading `getBoundingClientRect()`. The signal is also enormous rather than marginal — with the bug the `icon-xs` button drops ~13px, with the fix ~1px — so the assertion would not be tuning-sensitive.

Not arguing this must land in this PR; the class-string tests do catch the regression the PR is most worried about, and the story's docs string honestly states it is the manual check. Raising it so the gap is a deliberate choice rather than an assumed impossibility.

---

🤖 **Claude AI** · **[UX]** (suggestion) — `button.stories.tsx:313`

The story renders an enabled button whose accessible name promises an action ("Clear search") but which has no `onClick`, next to a `readOnly` input that gives no visual signal that typing is refused. Keyboard and screen-reader users on the autodocs page can Tab to the button, activate it, and get no feedback — a broken affordance in a page that documents the component. `fn` is already imported at line 2, so `onClick={fn()}` costs nothing; alternatively drive `defaultValue` from state and have the button clear it, which also makes the demo match the real `SearchBar` behavior. Keep the input editable either way — `:active` on the button is unaffected by input interactivity, so nothing about the pressed-state demo is lost.

The other display-only stories in this file (`VariantsDemo`, `SizesDemo`) read as static swatches; this one reads as a working search field, which sets the expectation that it works.

---

🤖 **Claude AI** · **[UX]** **[Clarity]** (suggestion) — `button.stories.tsx:325-326`

The story description is a single ~500-character, five-sentence paragraph where every other `docs.description.story` in this file is one short sentence (lines 53, 74, 104). It mixes three audiences: the reader instruction ("Hold it down: it should nudge 1px and stay vertically centered"), the CSS rationale, and the test-strategy note. The one line a docs reader actually needs is buried in the middle. Storybook renders these as markdown, so a lead sentence plus a short bulleted "why" would read far better.

Two specifics: sentence three is the third copy of the `--tw-translate-y` explanation in this PR, and sentence five names `button.test.tsx`, which in turn names this story at `button.test.tsx:17` — a circular cross-file pair that goes stale silently if either file is renamed. The pointer from the test to the story is the more useful direction of the pair.

Separately, the claim that `tw:-translate-y-1/2` is "the standard way to center an icon button inside an input" overstates repo convention: the package's own `SearchBar` centers its clear button with `tw:inset-y-0 tw:my-auto` (`search-bar.component.tsx:84`), not a translate. Softening to "one common way" avoids steering extension authors — or a future `SearchBar` refactor — toward the pattern that triggered this bug.

---

🤖 **Claude AI** · **[Tests]** **[Clarity]** (suggestion) — `button.test.tsx:24-25`

Of the two assertions in test 1, line 25 is the durable one and line 24 is the brittle one. Line 25 encodes the actual rule ("the nudge must not be an `active:`-modified translate utility"); line 24 pins one exact spelling of the fix. A behavior-preserving rewrite — `transform-[translate(0,1px)]`, an added `!`, a different modifier order — breaks line 24 while the pressed geometry is unchanged. Line 25 already catches the stated `shadcn apply` revert on its own (confirmed by the revert run); line 24's only independent detection value is "nudge deleted entirely", which a positive regex (`expect(className).toMatch(/active:\S*transform-\[translateY/)`) would catch just as well while surviving cosmetic rewrites.

The regex on line 25 is correct in both directions — verified, not eyeballed. Against the reverted class `tw:active:not-aria-[haspopup]:translate-y-px`: `active:` matches, `\S*` spans `not-aria-[haspopup]:`, then `translate-` matches. Against the current class it does not match, because `transform-[translateY(1px)]` contains `translateY(` and never the literal `translate-`. `\S*` cannot cross whitespace, so it is confined to one class token. No false positive today.

But it passes only because of that one-character distinction — the trailing hyphen in `translate-` not matching the capital Y in `translateY(` — which is easy to misread as a bug, and which the next editor could silently break by "simplifying" the pattern to `/active:\S*translate/`. A short note ("the trailing hyphen is load-bearing") would make it legible. The 12-line block above the test explains the CSS mechanism but says nothing about how this assertion is constructed.

---

🤖 **Claude AI** · **[Clarity]** (suggestion) — `button.test.tsx:8-13`

The first paragraph of the test's comment block restates the mechanism already explained in `button.tsx:16-21` — shared `--tw-translate-y`, the caller's centering being replaced, `transform` composing. That is the second of three copies of the same explanation in this PR (the third being the story description). The second paragraph (lines 15-18) is the part that earns its place: the jsdom limitation and why the assertions pin a class rather than geometry is reasoning the test code cannot show. Dropping the first paragraph leaves the test comment saying only what is true of the test.

It would also bring the block in line with the sibling test file, where each `it` carries its own 3-4 line note rather than one block covering both. As written the block sits above the first `it` but describes both, so it reads as a comment on the first test only.

*Cross-ref:* `context-menu.test.tsx:49,69` — per-`it` comments, each scoped to the case it precedes

---

<sub>Generated by `/review-pr`. Security review found nothing. AI review is advisory — human approval still required.</sub>

lib/platform-bible-react/src/components/shadcn-ui/button.tsx
imnasnainaec: Leaving them together for read/edit simplicity.

lib/platform-bible-react/src/components/shadcn-ui/button.tsx
imnasnainaec: `../dist/` is already stale, so let's not rebuild it unless somebody on the PT team request it.

imnasnainaec (commented):
> @imnasnainaec reviewed all commit messages and resolved 5 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2667)* status: 0 of 3 files reviewed, all discussions resolved.

katherinejensen00 (changes requested):
Thanks for fixing this button bug, Danny! Just one small thing to look at and then you are all set. Thanks for doing so many rounds of review already

> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2667#-:-Ozr0KG67IQno3ZuZk62:b-z26dq5)* status: all files reviewed, 1 unresolved discussion (waiting on imnasnainaec).

___
*[`lib/platform-bible-react/src/stories/shadcn-ui/button.stories.tsx` line 310 at r2](https://reviewable.io/reviews/paranext/paranext-core/2667#-Ozr-szJ-MEJ-pA6cWXX:-Ozr-szJ-MEJ-pA6cWXY:blnj1ep) ([raw file](https://github.com/paranext/paranext-core/blob/cd1bba253f859e1cd671723e875b14fd0d9830b9/lib/platform-bible-react/src/stories/shadcn-ui/button.stories.tsx#L310)):*
> ```XML
>       <Button
>         aria-label="Clear search"
>         className="tw:absolute tw:inset-e-1 tw:top-1/2 tw:-translate-y-1/2"
> ```

**`tw:inset-e-1` in the new story is not a real Tailwind utility — the clear button won't be positioned at the input's end.** · **Low** 

The whole `shadcn-ui/` directory positions inset-inline-end with `tw:end-*` (e.g. `tw:absolute tw:top-2 tw:end-2`); there are **zero** `inset-e-*` usages in the repo, and Tailwind v4 has no `inset-e-*` (it's `end-*`). So `tw:inset-e-1` generates no CSS: with only `tw:absolute tw:top-1/2 tw:-translate-y-1/2` taking effect, the X renders at the container's start edge, not inside the input's end. The vertical-centering-under-press behavior the story exists to demonstrate still shows, so it's cosmetic — but this story is explicitly the "correct idiom to copy," and the PR's own rationale is that we shouldn't "demonstrate the broken idiom to anyone copying the pattern." Change to `tw:end-1`.

imnasnainaec (commented):
> @imnasnainaec reviewed all commit messages and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2667)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on katherinejensen00).

___
*[`lib/platform-bible-react/src/stories/shadcn-ui/button.stories.tsx` line 310 at r2](https://reviewable.io/reviews/paranext/paranext-core/2667#-Ozr-szJ-MEJ-pA6cWXX:-OzrepXK5dd0iTUkgfNo:b-fp369d) ([raw file](https://github.com/paranext/paranext-core/blob/cd1bba253f859e1cd671723e875b14fd0d9830b9/lib/platform-bible-react/src/stories/shadcn-ui/button.stories.tsx#L310)):*

**`tw:inset-e-1` in the new story is not a real Tailwind utility — the clear button won't be positioned at the input's end.** · **Low** 

The whole `shadcn-ui/` directory positions inset-inline-end with `tw:end-*` (e.g. `tw:absolute tw:top-2 tw:end-2`); there are **zero** `inset-e-*` usages in the repo, and Tailwind v4 has no `inset-e-*` (it's `end-*`). So `tw:inset-e-1` generates no CSS: with only `tw:absolute tw:top-1/2 tw:-translate-y-1/2` taking effect, the X renders at the container's start edge, not inside the input's end. The vertical-centering-under-press behavior the story exists to demonstrate still shows, so it's cosmetic — but this story is explicitly the "correct idiom to copy," and the PR's own rationale is that we shouldn't "demonstrate the broken idiom to anyone copying the pattern." Change to `tw:end-1`.
</blockquote></details>

Changed to `tw:end-1` in a5a7fc500a3 — agreed the story should show the idiom the components actually use.

One clarification: `tw:inset-e-1` is a real utility. Compiled against our pinned Tailwind 4.2.2, both emit identical CSS:

```css
.tw\:end-1     { inset-inline-end: calc(var(--tw-spacing) * 1); }
.tw\:inset-e-1 { inset-inline-end: calc(var(--tw-spacing) * 1); }
```

So the clear button was positioned correctly before this change. It's also already in use — `tooltip.stories.tsx:981` (`inset-s-3`) and `:992` (`inset-e-0`).

Worth a decision: Tailwind's language server treats `inset-e-*` as canonical and now warns on `tw:end-1` — that hint is where the `inset-e-1` came from in the first place. Nothing in CI or the AGENT files enforces either direction. Settle on `end-*` as house style, or follow the tooling?

_(drafted by Claude, modified by me)_

katherinejensen00 (approved):
> @katherinejensen00 reviewed 1 file and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2667)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

Sebastian-ubs:
> Headline: the fix does not reach any consumer without a `dist/` rebuild

Right. Without building `platform-bible-react` this will not reach any consumer

Issue #1, opened 2023-01-23
Scaffold Core - Combine Electron, React, Edge, C#


tjcouch-sil:
As a developer, I want to develop on a base of code projects including the relevant technologies for the software so I can develop the software.

- Combine ERB and Edge
- Add C# solution that integrates well with the Node stuff

tjcouch-sil:
https://github.com/paranext/paranext-core/tree/scaffold-core

tjcouch-sil:
It seems very challenging to get electron-edge-js to work.

We had to set some csproj settings to match the [quick start project](https://github.com/agracio/electron-edge-js-quick-start) so it would include many C# dlls and the Edge dll provided in node_modules.

We discovered that electron-react-boilerplate has particular challenges with native module dependencies as detailed here:
https://electron-react-boilerplate.js.org/docs/adding-dependencies/#module-structure
https://www.electronjs.org/docs/latest/tutorial/using-native-node-modules
https://webpack.js.org/configuration/externals/
So we had to add electron-edge-js as a dependency in release/app/package.json instead. Then its postinstall runs @electron/rebuild to rebuild the native dependency for the current version of Electron.

@electron/rebuild was saying it couldn't get the abi for Electron 22.0.0. We had to uninstall and reinstall @electron/rebuild to rebuild electron-edge-js as a native dependency. If this does not work in the future, try adding a new node-abi as a dependency override:
https://github.com/electron/rebuild/issues/1031
https://stackoverflow.com/questions/15806152/how-do-i-override-nested-npm-dependency-versions

Then node-gyp was saying it couldn't find Python. So I installed 3.6.x as detailed in the [Building on Windows](https://github.com/agracio/edge-js#building-on-windows) section of edge-js readme. `choco install python` (may need to add `--version=3.6.8` if latest doesn't work)

Then it said it couldn't find a Visual Studio installation to use, so I used the Visual Studio Installer to install "Desktop development with C++".

irahopkinson:
You probably already know this and this may not be related to the issue above but [electron-edge-js](https://github.com/agracio/electron-edge-js) has to have a matching electron and nodejs version. Thats why the `main` branch has electron v20 and the `.nvmrc` file specifies node v16.15. If you update the electron version you should also update the nodejs version (and the `.nvmrc` file) you are using to the one specified by [electron-edge-js](https://github.com/agracio/electron-edge-js).

tjcouch-sil:
[...]

PR #2669, opened 2026-08-12
Add --wrap to run-e2e-wsl.sh and correct the e2e run recipes

rolfheij-sil:
## Summary

`e2e-tests/run-e2e-wsl.sh` exists to keep local e2e runs off the Windows desktop (WSLg renders
every Electron window the suite launches straight onto it), but it hardcoded the smoke suite — so
any other invocation, like the isolated multi-window suite, had to call `npx playwright` directly
and take over the desktop for the whole run.

The script now runs any command in the virtual display when you ask for it with `--wrap`. The
no-command form keeps its original behaviour: the smoke suite, with arguments forwarded.

### Why review this

Not part of the current epic — local-tooling quality of life, explicitly requested by Rolf
(2026-08-12) after desktop-rendered e2e runs kept interrupting him. Worth reviewer time now because
it changes a script other developers may share, and because the review turned up a set of
documented e2e recipes that do not work.

## Changes

**`run-e2e-wsl.sh`**

- `--wrap <command>` selects passthrough; anything else goes to the smoke run. The first version
  inferred this by asking whether the first argument was a command on `PATH`, which answers the
  wrong question: `run-e2e-wsl.sh find` ran GNU `find` from the repo root and exited 0 having run
  no tests, and a first word that was *not* on `PATH` (a typo, `pnpm`, `playwright` without `npx`)
  fell through to the smoke suite with its words injected as path filters. Both were silent. An
  explicit opt-in also matches `run-isolated.mjs`, which dispatches on a closed vocabulary and
  fails loudly.
- The banner now prints after the mode is decided, names the resolved command, and goes to stderr,
  so a wrapped run's stdout stays consumable.
- `exec xvfb-run`, so Ctrl-C reaches it rather than orphaning it behind this shell; and the
  duplicated `xvfb-run` literal is gone.
- Usage comment corrected: it named an npm script that does not exist, omitted `--config`, omitted
  the extra `--` that npm requires before Playwright flags, advertised `--debug` (unusable on an
  invisible display), and claimed results identical to WSLg (a bare Xvfb has no window manager).

**Docs and config** — correcting e2e run recipes that do not work, found while establishing which
suites the wrap actually helps:

- `npm run test:e2e:isolated all` and `npm run test:e2e:all` cannot pass, and are no longer
  presented as if they can. `tests/isolated/title-bar/` and `tests/isolated/navigation-history/`
  use `fixtures/cdp.fixture.ts`, which attaches to a running app, while the `isolated` project's
  `globalSetup` aborts when port 8876 is bound — the port `refresh.sh` waits on. No app and the
  attach fails; an app and global setup aborts before any test runs. `playwright-cdp.config.ts`
  cannot run them either: it `testIgnore`s `**/isolated/**`.
- The `enhanced-resources` instructions had the same problem and now route through
  `playwright-cdp.config.ts`, which has no `globalSetup`.
- `find-replace.spec.ts` documented two recipes that do not work (the bare form runs nothing; the
  other exits 1 with `Unknown subset`).
- Each isolated subset now records how it gets its Electron — per test, per worker, or by attaching
  — instead of the blanket "each test launches its own instance", which was false for four of them.
- The usage banner gained the path form, the only route to the three spec files directly under
  `tests/isolated/` that belong to no subset.
- A comment on the CI Linux e2e step explains why its Xvfb geometry differs, and why adding
  `--server-args` there would not enable a `cdp.fixture` job. CI's runtime arguments are unchanged.

## Testing

- Every misfire from the review re-run against the fixed script: `find`, `true`, `--`, `-p` and an
  unknown first word all route to the smoke suite, where a non-matching filter fails loudly.
- `--wrap` runs the given command inside the display with stdout clean; `--wrap` with no command
  exits 2.
- The default no-argument form is unchanged.
- Corrected recipes checked by collecting their tests rather than by inspection.

## Follow-up, not in this PR

`title-bar/` and `navigation-history/` remain unrunnable. Fixing that means moving them out of the
`isolated` project or giving `globalSetup` a narrow opt-out — a structural decision, so this PR
documents the behaviour rather than changing it.

## Risk Level

Low — local developer tooling and documentation. CI's behaviour is unchanged; the script's default
path behaves as before.

AI-assisted — [session 1](https://claude.ai/code/session_01Lf3XmoA9StqLNDiMBqZVTy), [session 2](https://claude.ai/code/session_01QyoVc1Cd9L2RcuDGE3T3FT)

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2669)

### DIFF
--- a/.claude/skills/test-runner/SKILL.md
+++ b/.claude/skills/test-runner/SKILL.md
@@ -143,7 +143,7 @@ Prerequisite: App running with `--remote-debugging-port=9223` (the `app-runner` 
 ```bash
 npm stop  # Port 8876 must be free
 npm run test:e2e:smoke                 # the CI smoke project
-npm run test:e2e:isolated [subset]     # per-test isolated Electron suite
+npm run test:e2e:isolated [subset]     # isolated Electron suite (per test or per worker)
 # raw invocation, if needed: npx playwright test --config=e2e-tests/playwright.config.ts --project=smoke
 ```
 
--- a/.claude/skills/test-runner/SKILL.md
+++ b/.claude/skills/test-runner/SKILL.md
@@ -20,7 +20,7 @@ Run and analyze tests for Platform.Bible (paranext-core) with structured output.
 | C# by category | `dotnet test c-sharp-tests/ --filter "Category=Contract"` |
 | Watch mode | `npm run test:core -- path/to/dir` (vitest watches by default) |
 | E2E (CDP, running app) | `npx playwright test --config=e2e-tests/playwright-cdp.config.ts` |
-| E2E (standalone) | `npm run test:e2e:smoke` (CI project) or `npm run test:e2e:isolated [subset]` |
+| E2E (standalone) | `npm run test:e2e:smoke` (CI project) or `npm run test:e2e:isolated <subset>` |
 
 ## TypeScript Tests (Vitest)
 
@@ -143,7 +143,11 @@ Prerequisite: App running with `--remote-debugging-port=9223` (the `app-runner` 
 ```bash
 npm stop  # Port 8876 must be free
 npm run test:e2e:smoke                 # the CI smoke project
-npm run test:e2e:isolated [subset]     # isolated Electron suite (per test or per worker)
+npm run test:e2e:isolated <subset>     # isolated Electron suite (per test or per worker). The
+                                       # bare form and `all` do not run: the bare form lists the
+                                       # subsets and exits 1, and `all` includes title-bar/ and
+                                       # navigation-history/, which need an app this project's
+                                       # global setup refuses to start alongside
 # raw invocation, if needed: npx playwright test --config=e2e-tests/playwright.config.ts --project=smoke
 ```
 
--- a/.context/standards/Testing-Guide.md
+++ b/.context/standards/Testing-Guide.md
@@ -1329,7 +1329,9 @@ npx playwright show-report e2e-tests/playwright-report
 # `npm run test:e2e:smoke`), `enhanced-resources` → tests/enhanced-resources/**.
 npx playwright test e2e-tests/tests/isolated/{feature}/ --config=e2e-tests/playwright.config.ts --project=isolated --reporter=list
 
-# WSL2: wrap any of the above so Electron windows do not take over the Windows desktop.
+# WSL2: wrap a standalone-mode run so its Electron windows do not take over the Windows desktop.
+# Only works when the app is launched inside the wrap — CDP mode attaches to an app started by
+# ./.erb/scripts/refresh.sh, which on Linux already runs it under its own Xvfb.
 # A bare Xvfb has no window manager, so compositor-dependent suites can behave differently.
 e2e-tests/run-e2e-wsl.sh --wrap npx playwright test e2e-tests/tests/isolated/{feature}/ --config=e2e-tests/playwright.config.ts --project=isolated
 ```
--- a/.context/standards/Testing-Guide.md
+++ b/.context/standards/Testing-Guide.md
@@ -1328,6 +1328,10 @@ npx playwright show-report e2e-tests/playwright-report
 # tests/isolated/** (most feature tests), `smoke` → tests/smoke/** (what CI runs, via
 # `npm run test:e2e:smoke`), `enhanced-resources` → tests/enhanced-resources/**.
 npx playwright test e2e-tests/tests/isolated/{feature}/ --config=e2e-tests/playwright.config.ts --project=isolated --reporter=list
+
+# WSL2: wrap any of the above so Electron windows do not take over the Windows desktop.
+# A bare Xvfb has no window manager, so compositor-dependent suites can behave differently.
+e2e-tests/run-e2e-wsl.sh --wrap npx playwright test e2e-tests/tests/isolated/{feature}/ --config=e2e-tests/playwright.config.ts --project=isolated
 ```
 
 ### Failure Analysis
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -165,11 +165,12 @@ jobs:
       - name: npm unit tests
         run: npm test
 
-      # No --server-args, so this runs at Xvfb's 1280x1024x8 default, unlike
-      # e2e-tests/run-e2e-wsl.sh which pins 1920x1080x24. That is fine for the smoke suite
-      # (it does not use cdp.fixture), but fixtures/cdp.fixture.ts hard-fails screenshots below
-      # Full HD — add --server-args="-screen 0 1920x1080x24" here before running any
-      # cdp.fixture-based project in CI.
+      # No --server-args, so this runs at Xvfb's 1280x1024x8 default. Fine here: the smoke
+      # suite launches its own Electron and does not use cdp.fixture. A cdp.fixture-based job
+      # cannot be added by putting --server-args on this step — that fixture attaches over port
+      # 9223 to an app started elsewhere, so this xvfb-run would never govern that app's window.
+      # Such a job has to start the app and Playwright under one Xvfb at 1920x1080x24, which is
+      # what fixtures/cdp.fixture.ts requires for screenshots.
       - name: E2E tests (Linux)
         if: ${{ matrix.os == env.OS_LINUX }}
         run: xvfb-run --auto-servernum npm run test:e2e:smoke
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -165,6 +165,11 @@ jobs:
       - name: npm unit tests
         run: npm test
 
+      # No --server-args, so this runs at Xvfb's 1280x1024x8 default, unlike
+      # e2e-tests/run-e2e-wsl.sh which pins 1920x1080x24. That is fine for the smoke suite
+      # (it does not use cdp.fixture), but fixtures/cdp.fixture.ts hard-fails screenshots below
+      # Full HD — add --server-args="-screen 0 1920x1080x24" here before running any
+      # cdp.fixture-based project in CI.
       - name: E2E tests (Linux)
         if: ${{ matrix.os == env.OS_LINUX }}
         run: xvfb-run --auto-servernum npm run test:e2e:smoke
--- a/README.md
+++ b/README.md
@@ -463,7 +463,10 @@ All `test:e2e:*` scripts are there for running variations of the playwright end-
   the `title-bar` and `navigation-history` subsets attach to a running app over CDP, which the
   project's own global setup refuses to start alongside
 - `test:e2e-cdp` runs tests that require the application UI to be open already
-- `test:e2e:all` runs all configured tests
+- `test:e2e:all` runs every project in `e2e-tests/playwright.config.ts`. Like
+  `test:e2e:isolated all` it cannot currently pass: `enhanced-resources`, `title-bar` and
+  `navigation-history` all attach to a running app, which that config's global setup refuses to
+  start alongside
 
 ## Storybook
 
--- a/README.md
+++ b/README.md
@@ -456,12 +456,12 @@ All `test:e2e:*` scripts are there for running variations of the playwright end-
 - `test:e2e:smoke-wsl` runs the same smoke tests using a "hidden UI" on Linux instead of a visible
   UI. Its script also takes `--wrap <command>` to run any other e2e command the same way, e.g.
   `e2e-tests/run-e2e-wsl.sh --wrap npm run test:e2e:isolated multi-window`
-- `test:e2e:isolated` runs tests that mostly require a separate application instance for each
-  test. These tests are organized into feature subsets under `e2e-tests/tests/isolated/`; run
-  `npm run test:e2e:isolated all` for every test, `npm run test:e2e:isolated <subset>` for one
-  feature, or pass no arguments to list the available subsets (no tests are run in that case).
-  `all` needs the app already running via `./.erb/scripts/refresh.sh`, because the `title-bar`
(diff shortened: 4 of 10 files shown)


katherinejensen00 (changes requested):
Read through this one carefully and verified the load-bearing claims by running them on real Linux (WSL2 bash) rather than by reading.

What's right: the motivation is real, and the no-arg default path is genuinely byte-equivalent — command -v "" fails, so test:e2e:smoke-wsl is unaffected. "$@" is quoted correctly throughout, and the file kept LF endings and mode 755.

The core concern is the dispatch predicate on line 35. command -v "$1" asks "is this token known to the shell?", not "did the user mean a command?" — and it misfires silently in both directions.

Correctness — the items with a verified trigger and a small fix:

1. L35 — find produces a false green. e2e-tests/tests/isolated/find/ is a real test subset, so find is plausible muscle memory from npm run test:e2e:isolated find. command -v find succeeds → passthrough runs GNU find with no args from the repo root: ~1.2M lines on stdout, exit 0, zero tests run, banner buried. The old script forwarded it to Playwright and failed loudly. This is a regression toward exactly the false green run-isolated.mjs:40-41 was written to prevent. I grepped every directory name under e2e-tests/tests/ against coreutils/findutils//bin//usr/bin/builtins — find is the only collision, so that list is complete. But the general shape has nastier members: ./run-e2e-wsl.sh true → exit 0, zero tests, zero output.
2. L35 — missing -- terminator on command -v. Verified: --, -v, -V, -p, -pv, -vv all exit 0. Against xvfb-run's real getopt string (+ae:f:hn:lp:s:w:): ./run-e2e-wsl.sh -- --grep smoke starts Xvfb then exits 127; -p is eaten as --xauth-protocol. One-line fix, verified end to end: command -v -- "$1" >/dev/null 2>&1. (The >/dev/null 2>&1 half matters too — &> is a bashism that under dash parses as background-job + redirect, making the condition unconditionally true.) Note this does not fix #1 — that needs an explicit opt-in, not a PATH probe.
3. L39 — the inverse misclassification fails silently. A first word not on PATH (typo, pnpm, playwright without npx) falls through to smoke with its words injected as path filters. ./run-e2e-wsl.sh playwright test --project=isolated runs smoke and isolated (--project is repeatable), and bare test matches every path under tests/. &> /dev/null discards the only diagnostic.
4. L36 — passthrough inserts no --, so npm swallows flags. Verified: npm run <script> --grep foo delivers ARGV ["foo"]. The form documented on L12 walks straight into this; L39 still inserts -- automatically, so nothing signals that it's now mandatory.
5. L12 — npm run test:e2e does not exist. Real names: test:e2e-cdp, test:e2e:all, test:e2e:isolated, test:e2e:smoke, test:e2e:smoke-wsl. Copy-paste → Xvfb starts, then Missing script. The nearest substitute isn't safe either: test:e2e:all includes enhanced-resources, which needs Marble resources.
6. L11 — npx playwright test ... documented without --config, and L30 already cd'd to the repo root where no root config exists. Testing-Guide.md:1316: "IMPORTANT: Always use the --config flag when running Playwrigwhere else in the repo spells it out.
7. L32 — the banner fires before the dispatch and is identical in both modes, which is what makes all of the above silent by construction. It should also go to stderr — passthrough is the first mode whose stdout is consumable, and --reporter=json > results.json currently writes the banner before the {.     L36/L38 — no trap and no exec. Non-interactive bash doesn't for Ctrl-C orphans xvfb-run → Xvfb → npm → Playwright → Electron plus the detached dev server (global-setup.ts:172-193). Each cancelled run leaks an Xvfb holding a /tmp/.X<N>-lock and keeps port 8876 held, so the next run dies with "Port 8876 is already in use" — and the stale pid file gets handed to process.kill(-pid, 'SIGTERM').
                                                                                                                                                      ems 2, 7 and 8 collapse into one rewrite:

command -v -- "${1-}" >/dev/null 2>&1 || set -- npm run test:e2e:smoke -- "$@"
echo "Running with virtual display (Xvfb): $*" >&2
exec xvfb-run --auto-servernum --server-args="-screen 0 1920x1080x24" "$@"

That also removes the duplicated xvfb-run literal and makes [ $# -gt 0 ] unnecessary (command -v "" already exits 1).

Also noted, lower priority:                                                                                                                           
- --project=smoke guarantee dropped in passthrough while the entry point is still named test:e2e:smoke-wsl and README.md:446 still says "the same smoke tests".
- L10's --debug can't work in an invisible display — it expands to PWDEBUG=1 --timeout=0 --headed, the Inspector attaches to Electron, and there's no globalTimeout: unbounded hang, nothing clickable.
- "identical results" (L4-5) is unsubstantiated and contradicted in-repo — fixtures/isolated.fixture.ts:60-71 force-resizes because "On headless Linux (xvfb) or WSL2 the default window can be very small", and the multi-window suite you validated against is written around WSLg compositor behavior (window-layout-persistence.spec.ts:55,139, multi-window.util.ts:190). A bare Xvfb normally has no window manager. 6/6 passing is good evidence but isn't equivalence — could the claim be softened?
- L30's cd precedes the dispatch, so cwd decides the branch for a relative path, and relative paths inside a wrapped command resolve against the repo root.
  Undocumented.
- "Wrap ANY e2e command" over-promises for CDP — playwright-cdp.config.ts:37 has no globalSetup and expects a running app, so the wrapper hands it an empty second display. Self-nesting isn't guarded either.
- Env-prefixed commands fail the gate — the repo's own idiom (ref=... npm start).
- The xvfb-run literal now lives in four places with a fifth divergent variant in CI (test.yml:170 has no --server-args → 1280x1024x8 vs 1920x1080x24 here). That geometry is load-bearing: fixtures/cdp.fixture.ts:64-65 hard-fails below 1920x1080.
- L17-27 hard-exits before argument parsing, blocking display-fre) and hosts with a usable DISPLAY. .erb/scripts/refresh.sh:19-26degrades gracefully instead.
- No -e/--error-file — xvfb-run defaults it to /dev/null, so on the exact WSLg setup L4-5 targets, Xvfb failures surface only as "Xvfb failed to start".
- Concurrent wrapped runs destroy each other — private displays, ver.pid and SingletonLock are global, and teardown's npm run stopkills matching processes machine-wide.
- Altitude: global-setup.ts is the single funnel every project passes through, and fixtures/helpers.ts:163 builds the Electron child env from process.env — setting DISPLAY there (or an E2E_VIRTUAL_DISPLAY=1 switch) woulor every suite and CI for free. As written, every other entry point still steals the desktop.
- Convention: this is the inverse of its own sibling — run-isolated.mjs:55-67 dispatches on a closed vocabulary and fails loudly.
- Docs: nothing was updated. README.md:446, the WSL2 setup sectiotill presents X-Server windows as expected), e2e-tests/CLAUDE.md'srun table, playwright.config.ts:29-39, tests/isolated/README.md:18 and Testing-Guide.md:1314+ are all unaware of the new mode, and no *-wsl npm script was added.

> @katherinejensen00 reviewed 1 file and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2669#-:-P-ouRtr4bCXWuSDkX9B:blprx70)* status: all files reviewed, 1 unresolved discussion (waiting on rolfheij-sil).

___
*[`e2e-tests/run-e2e-wsl.sh` line 35 at r1](https://reviewable.io/reviews/paranext/paranext-core/2669#-P-oo0VZ8bM28BhAUfOG:-P-oo0VZ8bM28BhAUfOH:b-kyac9s) ([raw file](https://github.com/paranext/paranext-core/blob/cf189374ede24dd92b419374258c7c5ba08f3dc2/e2e-tests/run-e2e-wsl.sh#L35)):*
> ```Shell
> # A first argument that is a command name means "wrap this command"; anything else keeps the
> # original behavior of forwarding the arguments to the smoke run
> if [ $# -gt 0 ] && command -v "$1" &> /dev/null; then
> ```

1. L35 — find produces a false green. e2e-tests/tests/isolated/find/ is a real subset, so find is plausible muscle memory. command -v find succeeds → passthrough runs GNU find with no args from the repo root: 1,243,192 lines on stdout, rc=0, banner buried, zero tests run. The old script passed find to Playwright, which fails loudly on a non-matching filter. A determinist

katherinejensen00 (approved):
> @katherinejensen00 reviewed 10 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2669)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

rolfheij-sil:
🤖 Claude: Thanks — this was a genuinely useful review, and the central point stands: the dispatch predicate was the wrong question to ask. Everything is addressed. Branch rebased onto main first (it was 23 behind), then eight commits of fixes.

**One deliberate deviation from your suggested rewrite.** You proposed keeping `command -v` with a `--` terminator. I went with an explicit `--wrap` opt-in instead, because your own note on item 1 says the PATH probe needs replacing rather than repairing — `--wrap` *is* that explicit opt-in, and it turns out to be the shorter diff as well as the more complete fix. It closes items 1 and 3, which the terminator alone leaves open, and it matches `run-isolated.mjs`, which dispatches on a closed vocabulary and fails loudly rather than guessing. Everything else in your rewrite (stderr banner, `exec`, dropping the duplicated literal, dropping `[ $# -gt 0 ]`) is in.

### Correctness items

1. **`find` false green** — fixed. Confirmed against main, where `tests/isolated/find/` exists; on the pre-rebase branch it did not, which is why the trigger only appears after rebasing. Reproduced it: exit 0, ~196k lines, zero tests. Now routes to the smoke run and fails loudly (`No tests found`, rc=1), the pre-feature behaviour. Your completeness check held — `find` is the only directory name under `e2e-tests/tests/` that collides with a PATH command.
2. **Missing `--` terminator** — moot, there is no longer a `command -v` on user input. Verified your list first: `--`, `-v`, `-V`, `-p`, `-pv`, `-vv` all passed the old gate. One correction: the `dash` parenthetical does not apply here — the script is `#!/bin/bash`, so `&>` was not a bug, only less portable. The `--` half was real.
3. **Inverse misclassification** — fixed by the same change; there is no silent fallthrough left to misfire.
4. **npm swallowing flags** — documented. Confirmed by probe: `npm run <script> --grep foo` delivers `["foo"]`, and `--project=isolated` delivers `[]`. The default (non-`--wrap`) form still inserts `--` for you; wrapped npm scripts need their own, and the usage comment says so.
5. **`npm run test:e2e` does not exist** — fixed, real script names throughout.
6. **`npx playwright test` without `--config`** — fixed; every documented form carries it.
7. **Banner before dispatch, on stdout** — fixed. It now prints after the mode is decided, names the resolved command, and goes to stderr so a wrapped run's stdout stays consumable.
8. **No `trap`, no `exec`** — `exec` is in, and the duplicated `xvfb-run` literal is gone.

### Lower-priority

**Taken:** `--debug` removed from the usage block; the "identical results" claim softened to name the WSLg/Xvfb window-manager difference; the CDP over-promise corrected; the `cd`-to-repo-root behaviour documented; env-prefixed commands now work (`--wrap env FOO=bar …`).

**Declined, with reasons:**

- **Altitude (`DISPLAY` in `global-setup.ts`)** — this is the one I most wanted to take, but nothing in `e2e-tests/` spawns an X server. Setting `DISPLAY` without one points at nothing, so the route means owning an Xvfb lifecycle inside the test harness — more machinery than the wrapper, not less.
- **`-e/--error-file`** — implemented, then backed out. It floods every *successful* run with ~20 lines of benign Xvfb/xkbcomp chatter, and `exec` forecloses post-processing it. There is a comment telling you to add it when debugging an Xvfb startup failure.
- **CI geometry** — I did not change the running args. CI's smoke suite does not use `cdp.fixture`, so 1280x1024x8 is harmless today, and raising a *blocking* job to 1920x1080x24 with no demonstrated need risks disturbing responsive-layout behaviour. There is now a comment at the step explaining the divergence — and, more usefully, that adding `--server-args` there would *not* enable a `cdp.fixture` job, because that fixture attaches to an app started elsewhere which this `xvfb-run` never governs.
- **Concurrent-run isolation, self-nesting guard, graceful degradation** — pre-existing harness properties rather than anything this change introduces; and degrading to a visible display would defeat the script's purpose.

### Docs

Updated `README.md`, `e2e-tests/CLAUDE.md`, the isolated README and the Testing Guide. No `*-wsl` npm script added — `--wrap` generalises it, which is the point of the change; a per-suite script for each would be the boilerplate it replaces.

### One thing worth your attention, found while fixing the docs

Correcting the "wrap any e2e command" wording meant working out which suites actually launch Electron. That surfaced a pre-existing problem unrelated to this PR: **`tests/isolated/title-bar/` and `tests/isolated/navigation-history/` cannot be run at all.** They use `fixtures/cdp.fixture.ts`, which attaches to a running app — but the `isolated` project's `globalSetup` aborts when port 8876 is bound (`global-setup.ts:121`), which is precisely the port `refresh.sh` waits on. And `playwright-cdp.config.ts`, which has no `globalSetup`, `testIgnore`s `**/isolated/**`. No app → `connectOverCDP` fails; app → global setup aborts before any test runs.

So `npm run test:e2e:isolated all` and `npm run test:e2e:all` cannot pass, and the documented `enhanced-resources` recipe was broken the same way (it works via the CDP config, which is what the docs now say). This PR documents the situation rather than papering over it; the structural fix — moving those two subsets, or giving `globalSetup` a narrow opt-out — is a separate decision.

### Note on the review text

A few passages arrived with characters dropped — item 7 runs into an unnumbered eighth item after a long run of spaces, and there are gaps in the Testing-Guide quote, the `refresh.sh` bullet and the Altitude bullet. I reconstructed the intent and believe I got all of it, but if something was lost entirely I would rather hear it than guess.

PR #2651, opened 2026-08-07
Host the app-global services and renderer commands in main (PT-4275)

rolfheij-sil:
Steps 3, 4 and 5 of the multi-window plan (**PT-4275**, §6 and §7), now one review unit. Everything app-global that a renderer used to own moves into the main process: the **scroll group service** (step 3), the **theme service** (step 4), and every **command and dialog request a window used to host** (step 5). Each renderer's copy of the two services becomes what the Service/Service Host pattern already calls it — a local cache of the host's state — and the per-window commands become router registrations in main that call a method on the window's service shard.

The end state is the one **Architecture.md §2** asked for: platform code in the renderer registers **no command names and no fixed request names at all**; the only names it still registers are derived from a per-instance web view id, which no two windows can hold. Getting there deletes `adr-singleton-services-elect-host-window`'s election, the takeover paths, the re-arm in every consumer, the cross-window mirrors, `command.service-router.ts`, both renderer-hosted name registries, both startup coverage checks, and ~1,760 lines of election/takeover machinery. Recorded in `adr-scroll-group-hosted-in-main`, `adr-theme-hosted-in-main` and `adr-renderer-registers-no-names`, with amendments to `adr-generic-name-routing-proxies` and `adr-shard-discovery-by-type` and two Patterns.md entries. Also riding here: a latent core **webpack `resolve.modules` bug** this work is the first to hit (see Part 1), which is extractable to its own PR against `main` if you would rather land it separately.

Consolidated from #2650 on 2026-08-07 to shrink the review stack; per-step incremental diffs remain viewable:

- Part 1 (step 3, scroll group, formerly #2650) — https://github.com/paranext/paranext-core/compare/80251974f28...f470e860732
- Part 2 (step 4, theme) — https://github.com/paranext/paranext-core/compare/f470e860732...d96a04a3729
- Part 3 (step 5, commands and dialogs) — https://github.com/paranext/paranext-core/compare/d96a04a3729...a5a47aae889
- Part 4 (2026-08-10 review round: TJ's two routing rulings, plus the small fixes his review and ours turned up) — https://github.com/paranext/paranext-core/compare/a5a47aae889...5b3233121be

**Stacked on #2649** (`pt-4275-router-shard-discovery`). Merge order: **#2621 → #2639 → #2649 → #2651**. Draft until the stack below merges.

---

## Part 1 — host the scroll group service in main (step 3)

[...]

PR #2670, opened 2026-08-13
PT-4281: 'window' layout, moving web views, cross-window reuse

rolfheij-sil:
# Summary

Adds the `'window'` layout disposition and everything needed to live with it: web views can
open into their own OS window, move between windows carrying their state, and find their
existing instances across every window instead of duplicating. Emptied windows close
themselves — except the last one, which docks Home, because windows are equal siblings and
closing the last one would exit the app. The saved window arrangement now restores in full,
however many windows were open at exit.

Serves PT-4281 (epic PT-4275); the primitives PT-4282's menus and PT-4287's automation
build on.

### Why review this

Part of the current multi-window epic. Stacked on #2651 (merge order #2649 → #2651 → this).

## What this delivers

- **`{ type: 'window' }` layout** (experimental): opens a web view in a window created for
  it, orchestrated by main (pending-content window → routed open → un-mark). In Simple mode
  — single-window by design — it degrades to a tab. A window whose routed content never
  arrives is closed again, never leaked; a provider that declines the open closes the window
  and resolves `undefined` (the API's established decline signal) rather than rejecting. If
  a reuse search cannot conclude whether an instance already exists, the open degrades to a
  tab rather than minting a duplicate into a brand-new OS window.

- **`targetWindowId` on `OpenWebViewOptions`** (experimental): open into a named window.
  Runtime-only handle; unknown ids error rather than silently falling back; contradictory
  combinations (`'window'` layout, `replace-tab`) fail fast; and a window that is already
  closing is refused as a target, so content cannot be routed into a window that is about to
  disappear.

- **Move primitive** (experimental): `platform.moveWebViewToNewWindow(webViewId)` and
  `platform.moveWebViewToWindow(webViewId, targetWindowId)`, plus a "Move tab to new window"
  tab context-menu item (Power mode, web view tabs). A move closes the view in its source
  window and reopens it from the captured definition, carrying its `useWebViewState` state
  across (state lives in per-window storage, so the move transfers it explicitly).

  **The id returned by the move is the authoritative post-move id** — for a view restored
  from a saved layout it can differ from the one passed in, so callers must use the returned
  value rather than assuming the original still resolves. Consumers see a close event then an
  open event; a held web view controller must be re-acquired after a move.

  On failure there is a recovery ladder: source window (unless it is already closing) →
  focused window → reject, with the captured definition logged. Close-source-first is
  load-bearing: one-instance views could never open in the target while the source instance
  lives. A move that fails from the tab context menu now tells the user so, instead of
  appearing to do nothing. A move whose adopt call times out but whose content actually
  landed is reconciled with the target and completes, rather than reopening the view a
  second time.

- **Window arrangement restores in full.** Quitting with several windows open brings all of
  them back — position, size and tabs — instead of only some. Whether a closed window's slot
  survives is now an explicit decision at the call site: a window the user closes during a
  session gives up its slot, while windows closed because the application itself is going
  down keep theirs, which is what makes a multi-window quit restore correctly. Persisted file
  format is unchanged.

- **Emptied windows** ask main whether to close or dock Home (`emptied-by-removal` /
  `born-empty`): any window emptying auto-closes unless it is the last real window, which
  docks Home instead — fixing the empty-new-window bug where Home opened as an undockable
  float. Floating/maximized tabs count as live content (a dock emptied behind a float docks
  Home locally, never closes the window), and a window that is showing a fallback layout
  docks Home locally rather than becoming a dead blank frame. Closing, pending-content, and
  abandoned windows (a renderer that crash-looped past its reload budget) are all excluded
  from the "last window" arithmetic, so overlapping closes, failed moves, and dead renderers
  can never exit the app or let a window that cannot be worked in stand in as the last one.
  These decisions are serialized and re-check the window immediately before closing it, so
  two windows emptying at once cannot take the app to zero. That count is one exported,
  directly-tested rule — `main.ts` composes none of it.

- **Cross-window reuse**: `existingId` reuse finds a view in any window and raises it (the
  raise honors `bringToFront: false` for passive probes, and consults real application focus
  rather than a last-focused id that is always set — so it does not steal focus when the app
  is in the background); `existingProjectId` (experimental) scopes an `existingId: '?'`
  search to a project — the filter narrows candidates before the routing-target preference,
  so a multi-window match still lands on the expected window. The app-wide `'?'` search
  itself and its per-caller reachability policy land in #2649; this PR carries the
  project-scoped extension and the consumers. The legacy comment manager's per-project memory
  map — whose stale eviction on cross-window moves was the defect — is deleted; the dock
  layouts are the reuse authority.

- **Raising a window is one path now.** The raise gate is the blur-aware `isApplicationFocused()`,
  replacing a last-focused id that latched the first time any window was focused and was never
  cleared — so it established only that some window had been focused once, never that this app
  held the foreground. That guard is the deferral #2649's body records against this ticket. The
  deep-link handler now raises through the same `focusWindow` service rather than raw
  `BrowserWindow` calls; it had declared a local binding of that very name which shadowed the
  import, so it was the one raise path getting neither the tracked-list lookup nor the
  flash-on-denial. It is deliberately **not** placed behind `isApplicationFocused()`: a deep link
  arrives while the app is in the background by definition, so that gate would suppress every raise
  the path exists to perform. Denial-and-flash is the intended outcome there — the lead dev's
  ruling on #2649 was to wire it through `focusWindow` here, "so it gets flash-on-denial for free
  the moment the real guard exists."

- **Hardening** along the way: replace-tab fails atomically before provider side effects
  exist when its target is in another window; a stale initial-layout answer no longer wipes
  content adopted mid-load; a failed reload no longer tears down the live view it left
  docked; an unreadable interface mode fails a move loudly instead of reporting a silent
  no-op; the open-project-tabs seed retries with warnings instead of silently swallowing a
  startup race.

## Design decisions worth reviewer attention

- Provider-declined `'window'` opens resolve `undefined` after closing the created window —
  a deliberate refinement over the spec's original "reject" wording, matching the API's
  existing decline signal.
- Unknown `webViewId` on a move throws; same-window moves and Simple-mode
  move-to-new-window are no-ops returning the id (bogus ids still error first).
- What a single-window move-out leaves behind depends on where the view went. Moving it to a
  NEW window leaves TWO windows — the moved view, plus Home docked in the window it left —
  because a window still waiting for its content is never counted by the last-window decision,
  so the emptying source sees itself as the last one standing. Moving it into a window that is
  already open and holding content closes the source instead, since that destination does
  count. Both are covered end to end. The same exclusion is what stops any ordering of a failed
  move from leaving zero windows (Rolf-ruled).
- Window layout tracking keeps a single list of persisted slots, each optionally bound to a
  live window, rather than reconciling separate live/persisted collections by runtime window
  id. Electron does not reuse `BrowserWindow.id` within a process, so slot-to-window binding
  is stable for the life of the process.
- Known and accepted for now: switching from Power to Simple mode with several windows open
  does not close the extra windows — Simple mode's single-window expectation is enforced at
  startup only. The intended behavior is captured in the epic's design notes and depends on
  the in-flight mode-transition work (#2425).
- A user-clicked-X window close is never announced to that window's renderer, so a move
  landing inside its close-task window is narrowed rather than impossible.

## Verification

- Unit tests throughout, written red-first; the race-sensitive pieces (closing-set algebra,
  recovery ladder, close-before-adopt ordering, filter idempotence) carry dedicated
  standalone suites. Where a test pins a rule that matters, it was mutation-verified: the
  rule was deliberately broken and the test confirmed red.
- The multi-window isolated e2e suite runs green headless on the final tree, re-run after
  every fix wave and every rebase rather than citing an earlier green.
- Live smoke runs with screenshot evidence, including exact-webViewId proof that one
  comment-list instance is raised rather than duplicated across windows, and is found in its
  new window after a live move. Highlights below.
- Full battery green: typecheck, lint, format, root and workspace tests, per-extension tsc.

### Evidence highlights

<details>
<summary><b>Move a web view to a new window</b> (command path): before — after (source) — after (new window)</summary>

</details>

<details>
<summary><b>Tab context menu</b>: "Move tab to new window" on a web view tab, and the result window</summary>

</details>

<details>
<summary><b>Single-window move-out</b>: ends with two windows — the moved view plus Home in the old window</summary>

</details>

<details>
<summary><b>Cross-window reuse</b>: a reuse open from window 1 raises the existing instance in window 2 instead of duplicating</summary>

</details>

## Merge

Squash-merge. Stacked on #2651 — do not merge before it; rebases as the stack merges.

AI-assisted — [session 1](https://claude.ai/code/session_01Lf3XmoA9StqLNDiMBqZVTy)

---

## Review and verification

This branch went through a four-dimension review — API/correctness, style/architecture,
coverage/compliance, and UX — followed by per-commit review of every fix the round produced.

**Findings were verified before being acted on, not taken as reported.** Test-quality claims were
settled by mutation: break the rule the test names, and see whether it goes red. Correctness claims
were traced to the decisive line or refuted there. That mattered — several confident findings were
wrong on their deciding claim and were dropped, and one was attributed to this branch when the same
behaviour predated it.

### What the review changed

**Defects fixed in the platform:**

- An emptied window decided its own close from a window count read before its content re-check.
  A window closed by the user during that re-check could take the whole application down with it —
  the outcome the last-window rule exists to prevent.
- An open did not wait for an in-flight layout load, so a web view opened during an interface-mode
  switch was thrown away by the load, silently: the close events are diffed against a reading taken
  before it arrived, so nothing reported it and its state leaked.
- The same silent wipe reached settings tabs and non-modal dialogs, which are not web views and so
  appear in no reporting at all. Both now wait, at their routed entry points.
- A move whose adopt expired after the renderer had already docked the tab left the same web view
  live in two windows, sharing one controller — so closing either copy disposed the survivor's.
- A move to a new window spent a full cold renderer start with the web view open in no window, and
  failed if that start ran long. The window is now readied before the source tab is captured, so a
  slow start costs a delay rather than the tab.
- A failed move told the user the same thing whether nothing had changed, the tab had gone to a
  window they did not pick, the tab was gone entirely, or the outcome could not be determined.
  Those are four different messages now.
- Waiting at the entry point was not enough on its own. A web view's route to the dock runs a
  provider in the extension host on the way — arbitrary extension code of no bounded duration — so a
  layout load starting while a provider thought landed the finished web view in a dock the load then
  replaced, silently. The wait and the refusal now also run where content actually enters the dock,
  with nothing left in between for the answer to go stale in.
- A reload whose web view left the dock while it waited no longer puts it back, behind a controller
  the load's close event had already disposed. A refusal that lands after the provider ran now
  unwinds what the provider registered, rather than leaving a controller and persisted state behind
  in a process that outlives the window.

**Rules that were true in the code but not pinned by anything** — window-bounds fallbacks (losing
one discards the placement a user last left a window in), the authoritative id a move returns, the
release of a departed window's mark, and several guards added during this round itself. Each is now
covered by a test that fails when the rule is broken.

**Moving a web view between windows is covered end to end**, by both routes a caller has — the tab
context menu and the two commands — against real Electron windows. It had none before, and the
ordering it exercises is the one least served by tests with stand-ins.

**Documentation corrected where it contradicted the code**, including two guideline pages that
still described the platform as single-window, and a passage that told extension authors an
`existingId: '?'` open fails when it actually goes ahead.

### Deliberately not done here

- **A window created while the user is in another application still takes the foreground.**
  Withholding activation from a window nobody asked for is **PT-4465**, deliberately deferred
  rather than attempted again here: focus state cannot distinguish "the user is elsewhere" from
  "the user closed every window and is clicking the dock icon", which is the path that creates a
  window through `app.on('activate')`. The caller knows and must declare it, which is what PT-4465
  builds. Recorded as `adr-window-activation-is-declared-not-inferred` so the inference is not
  reached for again before that ticket lands.

- **No keyboard route to "Move tab to new window".** The context-menu trigger is a non-focusable
  descendant of the tab, so Menu/Shift+F10 never reaches it — a limitation the pre-existing "Float
  tab" item shares. The fix is a real entry point rather than a trigger tweak, which is
  **PT-4282**'s subject.
- **Power → Simple with several windows open** does not close the extra windows; the single-window
  expectation is enforced at startup only. Depends on the in-flight mode-transition work (#2425).
- A macOS-only case where closing the last window and opening a new one duplicates the window
  entry is recorded on **PT-4286**, whose subject is app lifecycle. **Pre-existing — the window
  tracking on this branch neither introduced nor fixed it**, and it needs a Mac to reproduce.

### Worth a reviewer's attention

- **This PR adds two decision entries.** `adr-window-activation-is-declared-not-inferred`
  records why whether a new window activates is declared by its caller and never inferred from
  focus state, scoped to activating a new window since focus state remains the right input for
  raising an existing one. `adr-main-orchestrates-real-windows` records why multi-window uses real
  `BrowserWindow`s orchestrated by main rather than rc-dock's `windowbox`, with the spike evidence
  behind it. The log identifies entries by a slug chosen at write time rather than by a number, so
  this one no longer moves when a sibling branch lands first. It held three different numbers
  before that convention arrived and merged under none of them, which is why it carries no
  `**Formerly:**` line.
- **`Architecture.md`'s service-router row is amended in this PR** (its own commit, so it can be
  taken or dropped on its own). The row said a router "holds no logic", which has not described the
  shape since routers existed — choosing which window answers is a decision only main can make. The
  replacement says what a router does own, and keeps the constraint that subsystem-sized policy
  belongs in its own main-process module beside the router. **By that standard this router still
  owes that move**, which is a deliberate acknowledgement rather than an oversight.
- **A visible consequence of the move reorder**: the new window now appears before the source tab
  closes, so for the length of the capture the user sees an empty window beside their tab.
- **A failed move's rejection carries a machine-readable marker** naming where the web view ended
  up. It rides in the error message because a rejection crossing processes arrives as a code and a
  message and nothing else. It is documented on both move commands, and callers must strip it
  before showing the message to a user.
- A modal dialog is settled when its window unloads. Its promise lives only in the overlay that
  shows it, it never joins the map this window's unload rejects from, and the request timeout is
  lifted for showing a dialog — so a modal on screen when its window is destroyed used to leave its
  requestor waiting for the rest of the session, and that requestor is usually in a process that
  outlives the window. It is rejected rather than resolved as cancelled, because nobody cancelled
  anything.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2670)


tjcouch-sil:
Reviewed the full PR segment plus how it sits under the stack's rulings. Verdict first: this is
high-quality, conformant work — the three primitives live entirely inside the router/shard
architecture, no renderer registers a global name, nothing persists a `BrowserWindow.id`, the
`isApplicationFocused()` guard is real and tested both ways, and the move machinery's race
handling (in-flight registry, timed-out-adopt probing, recovery ladder that always rejects with a
machine-readable marker) is unusually rigorous. Also noted with approval: PT-4281 flagged that
adopting rc-dock's `windowbox` would relax the deliberate `window.open` hardening, and this PR
sidestepped that entirely by creating real windows through main. No hardening layer was touched.

Rulings and asks, numbered:

**1. The router-row rewording is accepted; the extraction it prescribes becomes a ticket.**
"Owns only what no single window can" is the honest statement of what a router is — the move
subsystem's state (the in-flight registry, pending-content lifecycle, abandoned-window close,
recovery ladder) is genuinely cross-window and belongs in main. But the router file should read as
routing again: lift the move subsystem into its own main-process module beside the router, as the
PR body itself concedes it owes. Follow-up ticket, not this PR — tracked and done, like the rest.

**2. `captureAndCloseWebView` and `adoptWebView` must validate their arguments.** Both are
wire-reachable from any process, and the bar is the one `setDetachedScrRef`'s own doc states: a
registered network object method checks rather than trusts. Today the capture id is unchecked and
`adoptWebView` writes the caller-supplied `state` blob into window storage before anything
validates the definition — and a pre-provider failure leaves that blob behind, so an arbitrary
caller can overwrite any web view id's stored state with unvalidated data even when the adopt then
fails. String-check the capture id; shape-check the adopt bundle (id, webViewType, state an
object); and do not seed state until validation passes.

**3. `captureAndCloseWebView` must wait out an in-flight layout load.** Every dock-entering path
waits (`waitForLayoutLoadToSettle` in openWebView, adoptWebView, addTab, the dialog shard) — the
dock-leaving half of the move does not. A capture racing a mode-switch load can remove the tab and
then have the finishing load resurrect it from the entry it was loaded with, while the adopt opens
it in the target — two windows, one id, the exact state the timed-out-adopt machinery exists to
prevent. Narrow window, one-line fix matching its siblings, plus a test.

**4. The capture's "live-state merge" contradicts the state wiring this branch ships — justify it
with the real path, or delete it.** I traced the wiring: `useWebViewState` writes go through
`setWebViewStateSync` → `updateWebViewDefinitionSync({ state })` — the dock definition is the live
authority — and the wrapper writes through to the state service, whose own docs say it is a
startup/persistence helper. Reads come from the definition too (`getFullWebViewStateSync`). So the
capture comment ("the live state lives in this window's storage, not on the dock's definition")
asserts the reverse of the shipped data flow, and the operation it justifies is a REPLACE
(`captured.state = liveState`), not a merge — meaning if any path ever advances definition state
without the write-through, the capture overwrites fresher state with staler on exactly the
operation built to preserve state. If there is a real path where storage is fresher than the
definition, name it in the comment and pin it with a test; otherwise drop the replace and correct
the comment. A wrong data-flow comment on a cross-window state operation is how the next person
builds on the wrong model.

**5. The stale doc block on your own headline guard.** `openWebViewInOwningWindow`'s TSDoc still
describes the removed latch ("nothing clears it… the blur-aware check that would make this safe is
TODO(PT-4281)") while the code beneath it calls the `isApplicationFocused()` this PR shipped.
Update the doc to describe the guard that exists.

**6. Window creation should respect the same focus principle as raises.** The guard stops raises
of existing windows while the user is in another application, but a `{type:'window'}` open or
`moveWebViewToNewWindow` from an extension while the app is backgrounded still creates a window
that takes OS focus — and unlike the deep link, nobody clicked anything. Backgrounded creations
should show without activating, plus the flash. Fine as a small follow-up in the same guard
family; say which way you take it.

**7. Lengthen the abandonment/content-arrival timeouts.** The 2-second bound on "did content
arrive since your report?" before closing a window is tuned for fast machines. Make it at least
5 seconds, and review the adopt probe's cadence (3×2 s) against the same principle: a messed-up
window sticking around a few more seconds is cheap; something the user tried to open failing to
open because their computer is a bit slower than we anticipated is expensive.

**8. Refuse emptiness reports for untracked windows.** Any process can report any window id; a
fabricated id that reaches the close decision poisons the handler's closing set, and when Electron
later mints a real window with that id, its first emptiness report is answered "closing" (no close
ever comes) and the renderer latches, refusing content forever. Cheap fix: validate the reported
id against the tracked window list before deciding anything.

**9. The reuse-race error is misleading.** On an `existingId` reuse hit with a `{type:'window'}`
layout, the original layout is passed to the owning window's shard; if the view leaves that window
between the owner search and the shard call, the shard proceeds to create and `addTabToDock`
throws "received a 'window' layout; these are handled by the main process" — a routing-contract
alarm for what is actually a lost race. Make that path fail as what it is.

**10. Accepted, for the record:** the emptiness handler's close-on-unverifiable-re-check bias
(with the longer bound from item 7 — the layout slot survives a wrong close, a zombie window has
no recovery); `targetWindowId`'s arrival — this is the windowId option my earlier ruling said
"follows when the open-in-a-chosen-window UX arrives," and this PR is that UX, so it lands as
planned; and PT-4351's sequencing after this PR, confirmed against the actual conflict shape.

**11. Design direction, recorded here and to be recorded in the PRD: platform-assigned window ids
replace Electron's on the PAPI.** Every window-id surface today exports Electron's `BrowserWindow.id`
semantics, and every mention needs a caveat ("runtime-only handle, reused across sessions, never
persist") — that is another system's dirty laundry on our API. Main already mints and threads a
`windowId` everywhere the platform speaks (URL param, shard attributes, routing, commands); mint a
main-owned counter instead, keep one internal map from platform id to `BrowserWindow`, and persist
the counter so ids are never reused across launches — which deletes every "never persist" caveat
and leaves room for a logical window to survive a recreated `BrowserWindow`. The swap must REPLACE
the Electron ids on the PAPI, not join them — one id namespace, main translating internally. Do it
while the whole surface is still `@experimental` and has no external consumers: its own small PR
right after this one, or in this PR if you size it as small as the existing plumbing suggests. A
window-enumeration API waits until a consumer exists, and if durable per-window identity is ever
wanted beyond this, that is the moment to design it.

tjcouch-sil:
While working through review feedback on the standard-view branch we hit a gap this PR seems best placed to settle: there's no supported way for a renderer or web view to learn **its own window id**.

Concrete case: the experimental app-window input event (and overlay dismissal keyed off it) carries no window identity, so with multiple windows an Escape/pointer-down in one window can dismiss or resolve an overlay belonging to another. To filter window-scoped events correctly, code running inside a web view (or the renderer) needs something like `papi.windows.getCurrentWindowId()` / a window id exposed on `@papi/frontend` — the per-window service scoping you're building here looks like it has the pieces.

Does PT-4281 (or a planned follow-up in PT-4275) introduce that primitive? If not, could it? We'd use it immediately to close the multi-window overlay cross-talk gap recorded in the `adr-app-window-input-kind-only` entry of Architecture-Decisions.md.

rolfheij-sil:
🤖 Claude: Two factual corrections before you re-review — both about claims in the review rather than about the asks, and the first corrects something in this PR's favour that shouldn't be.

**1. "Nothing persists a `BrowserWindow.id`" is true of this diff but not of the branch.** Two places persist it today:

- `localStorage` keys are prefixed with it — `` const prefixedKey = `${getWindowIdOrThrow()}_${key}` `` (`src/renderer/services/local-storage.service.ts:10`)
- web view ids carry a `-w<windowId>` suffix that is written into saved layouts — `` const scopedId = `${stripWindowScopeFromWebViewId(tab.id)}-w${globalThis.windowId}` `` (`src/renderer/components/docking/window-scoped-web-view-ids.util.ts:57`)

The storage service's own comment already acknowledges the instability: *"Electron's `BrowserWindow.id` is not guaranteed to be stable across restarts, so the legacy key must remain as a fallback (durable per-window identity is PT-4285's scope)."*

This **strengthens** your platform-minted-id direction rather than weakening it — persisted-and-reusable ids are precisely the hazard your caveat warns about — which is why you should have it before you re-review rather than after.

Worth knowing alongside it: **PT-4285 carries a recorded decision (2026-07-28) that window ids stay runtime-only** — "the position in the persisted list *is* the identity". That predates your ruling and you can't have seen it. The conflict may well be narrower than it looks, since never *reusing* a number is not the same as *identifying* a window across restarts, but it is a decision on an in-progress ticket rather than an open question, so it wants reconciling rather than implementing around.

**2. One ask shipped without a test, deliberately.** The capture now waits out an in-flight layout load, and that ordering is **not** unit-tested: driving a load in flight needs a harness the shard does not expose today, and a test that could not fail would be worse than none. The commit message says the same. Flagging it here so the fix-round report to follow is not read as implying coverage it does not have.

A fuller fix-round report follows once the remaining items are pushed.

rolfheij-sil:
🤖 Claude: On a web view learning its own window id — it is cheaper than it looks, it is now ticketed, and there is one thing we need from you before the shape can be settled.

**Tracked as PT-4464**, "Mint platform-assigned window ids in main and replace Electron's across the window-id surface", under PT-4275. Your item 11 and this ask are the same surface, so they land together rather than separately — details below.

**It is cheaper than the obvious route.** A web view's `about:srcdoc` iframe is a separate realm, so `globalThis.windowId` inside it is `undefined`, and the instinct is to add it to the `imports` allowlist that `openOrReloadWebView` injects. That turns out to be unnecessary. The same injected string does `window.papi = window.parent.papi`, which binds the **parent realm's object itself, not a copy** — so any papi function that closes over `globalThis.windowId` already resolves to the hosting window when called from inside a web view. That is how `papi.window` works from a web view today. An accessor hung off the frontend papi therefore inherits correct window scoping for free, with no change to the allowlist and no change to the iframe plumbing — which matters, because that allowlist is immediately followed by `delete window.parent; delete window.top; delete window.frameElement;` and is not somewhere to add hand-maintained entries lightly.

**Why it is folded into the id change rather than shipped first.** If platform-assigned ids are going to replace Electron's on the PAPI, then `windowId` changes meaning in the same breath. Exposing it to web views first would publish `BrowserWindow.id` semantics to a new class of consumer and then take them away. Same surface, same commit.

**What we need from you: we cannot find the consumer.** Searching `src/`, `lib/`, `extensions/src/`, `docs/`, `.context/` and `e2e-tests/` for `app-window`, `appWindow`, `window input`, `onDidReceiveInput` and `Input` in `papi.d.ts` returns nothing — there is no app-window input event on this branch or on `main`. So we think it lives on the standard-view branch, which we cannot read.

Could you point us at it, or paste the event's shape? Specifically: **what identity does it carry today, and who subscribes?** The accessor's form — string vs number, on `papi.window` or a new `papi.windows` — depends on how it gets consumed, and PT-4464 has to make that call.

**A note on the ADR you cited, because the numbers have moved under all of us.** You referenced ADR-0029 for the overlay cross-talk gap. When you wrote that, ADR-0029 was this stack's "Renderer platform code registers no command or request names" — but `main` claimed 0026–0028 while the stack was open, so after last night's rebase that entry is now **ADR-0032**, and 0029 is #2649's "Service routers discover shards by network-object type". Neither describes app-window input announcements, so we do not think the pointer was to one of ours in the first place — but worth flagging, since it is the same drift you caught in our PR bodies.

**One thing worth saying plainly so "let us settle the shape first" is not read as foot-dragging on a live bug: there is no cross-window leak in shipped code.** Overlay dismissal today is entirely renderer-realm-local — `registerAutoDismissListeners` in `overlay.service-host.ts` uses this window's own listeners and this window's own focus subject, and never crosses the network. The hazard you describe appears only once an input signal is broadcast over PAPI, which is presumably what your branch does. The urgency is there, not on `main`.

**The nearest identity-free path that does exist here, in case it is relevant:** the global shortcut handler in `createWindow` (`src/main/main.ts`) dispatches commands with no window identity at all, and carries two `TODO(PT-4143)` comments saying exactly that — "this preventDefault is global and focus-blind — it fires for any focus context, including inside WebView iframes". If your event is meant to replace or complement that path, PT-4143 may be where it belongs.

rolfheij-sil:
🤖 Claude: Fix round for items 1-10, on `2cab0ddf6b5`. Item 11 is untouched — it now has a ticket, see below.

## Fixed

**2 — argument validation on both wire-reachable halves.** The capture checks its id. The adopt shape-checks the bundle **before the state seed**, which is the part that mattered: the seed persists immediately, so validating after it would still have let an arbitrary caller park a blob under any web view id. Two further gaps surfaced while covering it — the check accepted arrays (`typeof []` is `'object'`; it now uses the same `isSerializable` the state setter uses), and a **well-formed** bundle whose provider then threw *or declined* still left the seed behind. Both paths now undo it. The decline path is worth naming: it returns `undefined` rather than throwing, so handling only the throw would have left half the hole open.

**3 — the capture waits out an in-flight layout load**, placed ahead of reading the dock rather than after it, since the wait is exactly the stretch in which a re-register detaches an already-read dock reference.

**4 — the comment is corrected and the replace removed.** You were right that it asserted the reverse of the wiring. On the hazard itself: `updateWebViewDefinitionSync` mirrors every state write into storage, and the only code bypassing that wrapper writes `scrollGroupScrRef` rather than `state` — so on every reachable path the replace substituted an equal value. **Not a live data-loss bug.** The fix is the comment plus deleting a redundant operation, not a behaviour change, and I would rather say so than claim a save that did not happen.

**5 — the stale doc block** on `openWebViewInOwningWindow` now describes the blur-aware guard that exists.

**7 — both bounds widened** (content re-check 2s → 5s; adopt probe 3×2s → 4×3s), with the reasoning recorded on the constants: the two ways of being wrong are not symmetrical. Waiting too long costs a pause; giving up too early closes a window whose content was arriving, or drops a move into recovery while the target is still finishing.

**8 — emptiness reports naming an untracked window are refused before any decision**, and the dependency is required rather than optional so no future composer can drop the check silently. One correction to my own reasoning in that comment: Electron ids ascend and are **never reused** within a process, so the id to fabricate is a *future* window's — report 7 while only 1-3 exist — and the latch hazard you described is reachable that way.

**9 — the reuse race** now reports a lost race instead of the dock's routing-contract alarm.

**Wire-contract TSDoc** on both move primitives updated for the new failure modes — the interface doc is what a caller in another process reads, and it described neither the checks nor the wait.

## Not fixed, deliberately

**6 — reverted from this PR; now PT-4465.** Two attempts both failed by reading focus state. The second broke the macOS dock-click path: after the last window closes nothing holds focus there either, so a window the user had *explicitly asked for* would have appeared unfocused and flashing. **Focus state cannot answer the question the guard asks** — whether a person in this app asked for this window. The caller can, and every relevant call site knows. PT-4465 takes it with that shape rather than a third heuristic here.

**`targetWindowId` is not validated against the tracked window list** in the router, unlike the two shard methods above. Noted rather than done: it is the router's validation, and it is the same surface as the id rework, so it is recorded on **PT-4464** to move with that.

**1 — the move-subsystem extraction is PT-4463.** Pure refactor, after this merges.

## Verification

typecheck, lint, `format:check` and the full unit suite green; `papi.d.ts` regenerates with no diff; the multi-window isolated e2e **9/9**; CI green on ubuntu, macOS and Windows for this SHA. The ADR renumbered again as the stack rebased — it is **ADR-0033** as of this commit.

A correction to a claim in the review itself is in my earlier comment: "nothing persists a `BrowserWindow.id`" holds for this diff but not for the branch.

tjcouch-sil:
Answers to everything outstanding, including the event shape you were blocked on.

**The app-window input event, since you can't read the branch.** It's
`platform.onDidAppWindowInput` — a network event **main** emits from per-window
`before-input-event` (Escape keydown) and `before-mouse-event` (mousedown) hooks in `main.ts`,
with payload `{ kind: 'mouseDown' | 'escape' }` and deliberately nothing else — the model's
security note restricts it to the two overlay-dismissal gestures with no key identity or
coordinates, so hooks that see every frame's input (including other extensions' web views) can't
become a keylogger. Subscribers today: the renderer's `overlay.service-host.ts` and the palette
component in `platform-bible-utils`. Your no-leak-in-shipped-code observation is correct — the
hazard begins when this event broadcasts, which is my branch's doing. The multi-window fix: add
`windowId` to the payload (that's routing identity, not input detail — amend the do-not-broaden
security note deliberately when you touch it), and give renderer/web-view code an accessor to
filter by.

**The accessor lives on `papi.window`. Do not create `papi.windows`.** The singular service's
semantics are already "the window relevant to you," which is exactly what the overlay filter
needs; a plural namespace earns existence only if real multi-window APIs (enumeration,
cross-window queries) ever arrive, and we've deferred those until a consumer exists.

**PT-4464's design, settled: platform window ids are GUID strings.** Not Electron's numbers, and
not a persisted counter. Main mints a GUID per window; the window-state module owns the only
GUID→`BrowserWindow` map, and everything else — including main's other services — speaks the
GUID. What this buys, concretely: the id becomes a *string* where Electron's is a number, so the
compiler flags every call site the migration misses and nobody can ever confuse the two
namespaces again; global uniqueness needs no persisted state; nothing about a GUID implies
ordering (window order lives in the window list, where it belongs); and ids stop being guessable,
which closes in depth the fabricate-a-future-window-id class you fixed in the emptiness handler.
Dev discoverability is already covered — the id rides the window's URL, visible at the top of
DevTools. The event's `windowId` and the accessor's return type are both `string` accordingly.

**Durability: I recommend durable** — persist the GUID in the layout slot, minting one on first
load for existing layouts, and amend PT-4285's "position is the identity" decision in the same
change (position remains the *ordering*). That also dissolves the PT-4285 conflict you flagged.
If you'd rather ship session-scoped GUIDs first, that's acceptable at your discretion **provided
the upgrade path stays clean and backwards-compatible** — which the GUID design gives you by
construction: slots gain an optional GUID field later, absent means mint fresh. Your
`targetWindowId`-validation note on PT-4464 is subsumed by the same change (the type flips to the
platform id, validated against the tracked map).

**PT-4465 — endorsed as ticketed.** The macOS dock-click evidence settles it: after the last
window closes, nothing holds focus, so no focus-state heuristic can distinguish "a person asked
for this window" from "an extension did this in the background." The caller knows; the caller
declares. Reverting the attempts rather than shipping a third heuristic was right.

**The Node port of the Testing-Guide gate — keep it; don't revert.** Your three reasons hold
(only Python invocation in the repo; warn-and-continue means no gate anywhere, since CI never
runs it; no clean cross-platform interpreter story), and the `process.exit`-discards-piped-stdout
find — diagnostics vanishing on exactly the path they exist for — justified the rewrite on its
own.

**ADR numbering: PR #2720 already solved it — stand down the merge-time-number proposal.** Main
now identifies decisions by stable kebab-case slugs chosen at write time (with "Formerly:
ADR-NNNN" lines), so nothing is reserved, renumbered, or collidable. Your stack's numbered
entries just need slugifying on the next rebase — mechanical. (The `adr-app-window-input-kind-only`
reference in my earlier comment was one of these slugs, not a number.) The **wire-surface blind
spot you found survives on its own and is worth building**: a missing wire-only `x-experimental`
regenerates a byte-identical `papi.d.ts`, so no diff check can see it. Shape it as a **snapshot
of the full wire surface** (the OpenRPC document) that PRs diff the way they diff `papi.d.ts` —
not as a marker-policing rule, since plenty of wire methods legitimately live off `papi.d.ts`
without being experimental.

**One nit from the fix round:** the capture-race fix's deliberately-missing test is recorded only
in the commit message. Put the one-liner in the code too — "untestable until the shard exposes a
load-in-flight harness; a test is owed then" — so whoever builds that harness knows.

Your two corrections of my review's claims are both accepted — the persisted `BrowserWindow.id`
usages you surfaced argue *for* PT-4464, and flagging the untested wait before the fix-round
report rather than after is how I want gaps handled. Everything else in the fix round verified
clean against the code on my side.

rolfheij-sil:
🤖 Claude: Thank you — that unblocks the two things I could not resolve from this side, and the
GUID ruling settles a tension I had flagged rather than papering over it.

**The nit is fixed.** The "no test yet" note now sits in the code beside the wait it describes,
in `captureAndCloseWebView`, rather than only in the commit message:

```
// Not unit-tested: driving a load in flight needs a load-in-flight harness the shard does not
// expose to tests. A test is owed once that harness exists.
```

**ADR slugifying is already done.** The stack was rebased onto the post-#2720 main this morning,
and all three branches now carry zero numbered headings — this PR's entry is
`adr-main-orchestrates-real-windows`, with no `Formerly:` line because it never merged under a
number. So there is no mechanical pass left owing on the next rebase.

**On GUID window ids — agreed, including durable.** Two things from this PR's side worth having
on the record. First, this PR's window-id surface is numeric (`targetWindowId` on the move
commands and on `OpenWebViewOptions`), and every item of it is `@experimental`; PT-4464 flips the
type rather than adding a second namespace, which is exactly what those markers exist to permit.
We would rather not churn this PR to pre-empt it — the compiler-flags-every-call-site property
you describe is worth more applied once, in the PR that does the migration, than spread across
two. Second, durable GUIDs dissolve the PT-4285 conflict in the direction I had argued for, and
they remove the counter's failure modes rather than hardening them: a persisted counter has to
answer what happens when its storage is missing, corrupt, or unwritable, and every answer to that
is worse than not needing one.

**Four digit-shaped assumptions the GUID change will have to land on.** PT-4464's own ticket
flagged these as constraints "each of which fails silently otherwise", so they need a home in the
new plan rather than a rediscovery. You have already answered the ordering one — window order
lives in the window list — which leaves three:

- **Owner tie-breaks.** `web-view.service-router.ts:263` and `:379` pick a winner with
  `matches.sort((a, b) => a.windowId - b.windowId)[0]`. On strings that subtraction is `NaN`, and
  a `NaN` comparator leaves the array in its **original** order — so the winner becomes whichever
  window's answer arrived first, which is the non-determinism the sort was added to remove. It
  fails silently and looks like it still works. The fix is not a string comparison but ordering by
  the window list's creation order, which is where you say window order belongs. Both sites are
  pre-existing, not this PR's.
- **Window-scope suffix.** `window-scoped-web-view-ids.util.ts:19` scopes a web view's id per
  window with `/-w\d+$/`, and `stripWindowScopeFromWebViewId` is what makes a moved view's id
  authoritative. A GUID suffix contains hyphens and letters, so the pattern stops matching and the
  strip silently no-ops — `/-w\d+$/.test('myview-we1b0-aaa')` is `false`. This PR adds a
  deliberate copy of that regex to its e2e helper, kept in sync by a comment, so the two move
  together. Pre-existing apart from the mirror.
- **Shard registration.** A window id reaches `getServiceShardAttributes` and
  `getServiceShardWindowId` in `service-shard.model.ts`, both of which are numeric by type today —
  the latter is `typeof windowId === 'number' ? windowId : undefined`, so a GUID would resolve to
  `undefined` and route nowhere rather than fail loudly. On PT-4464's own branch the string
  overload has already been removed, so this is two type flips rather than a parser change.

Not on the list, deliberately: `PREFIXED_DOCK_LAYOUT_KEY_PATTERN` in the shard has the same
`(\d+)` shape but is a **legacy reader** — it only ever matches keys written by the numeric-era
builds, which stay digits forever, and nothing writes that shape any more. It should stay
digit-only; widening it to accept GUIDs is a way to break the legacy layout fallback rather than a
migration step.

None of this argues against GUIDs — the id-shape reasons you give are stronger than the
convenience they cost. These are just the places where the migration is a rewrite rather than a
re-type, and the first two fail quietly rather than loudly, which is why they are worth naming
before the plan is written.

**The accessor on `papi.window`, singular — understood**, and the reasoning generalises: the
plural namespace earns existence when enumeration or cross-window queries have a consumer, not
before. It lands with PT-4464 so its return type is `string` from the first commit rather than
flipping under an extension that already adopted it.

**On `platform.onDidAppWindowInput` gaining `windowId`:** agreed that routing identity is a
different kind of thing from input detail, and amending the do-not-broaden note deliberately —
rather than quietly widening the payload — is the right way to do it. One ordering note: the
event's `windowId` and the accessor should ship as GUID strings together, so nothing consumes a
numeric window id from a shipped event and has to be migrated afterwards.

**The wire-surface snapshot:** agreed it stands on its own, and your framing is better than the
rule I proposed — a snapshot of the OpenRPC document diffed the way `papi.d.ts` is diffed catches
the blind spot without asserting that every wire method ought to be experimental. It needs a
ticket rather than riding this stack; flagging it here so it is not lost.

**PT-4465 and the Testing-Guide port:** noted, nothing further from me on either.

rolfheij-sil:
🤖 Claude: Following up on the "not unit-tested" comment you asked for on the capture-order fix — a review pass challenged it, and the challenge was right. The comment is wrong and I would like to replace it with the test.

What it claims is that driving a load in flight needs a harness the shard does not expose to tests. The real situation is narrower: `loadLayout` only registers a load as in-flight when the dock already held web views when that load began (`webViewsBeforeLoad.length > 0`), and every dock stand-in in `web-view.service-shard.test.ts` hard-codes `getAllWebViewDefinitions: () => []`. So no test in that file had ever produced a tracked load-in-flight, and `waitForLayoutLoadToSettle` had never been exercised. The rest of the harness was already there — `holdGetLayout()` parks the saved-layout request, and the interface-mode switch re-enters `loadLayout` — which the "a layout change while the switch to power is still loading pushes nothing" test already uses.

With a dock stand-in that reports its contents honestly, the test is straightforward: park a second load, call `captureAndCloseWebView` during it, land the load with a layout that drops that web view, and assert the capture resolves `undefined` and never calls `removeTabFromDock`. It passes, and it fails with `expected { id: 'target-web-view', … } to be undefined` when the `await waitForLayoutLoadToSettle()` is removed — so it pins the ordering rather than just exercising the path. The other 52 tests in the file are untouched by that mutation.

It lands with the rebase push, with the comment removed.

tjcouch-sil:
Reviewed the delta since the last fix round. Tip is clean — the capture-race test is real and
mutation-verified, all round-5 fixes still hold after both rebases, and the new mid-move
visibility work is sound. Five asks and two closures.

**1. Record the focus-guard principle durably — it was re-derived from scratch a third time this
round.** Two commits (`de053f9967e`, `8fbcd6c3ce9`) landed a pure focus-state heuristic gating
window creation — the same shape the earlier item-6 attempts took, and it fails the identical
recorded case: after the last window closes on macOS, nothing holds focus, so the dock-click
restore the user explicitly asked for would have appeared inactive and flashing. `ab67e6e256b`
caught it and removed it before tip, with the commit message restating PT-4465's principle
correctly. Good outcome, but the only place that principle exists is a commit message squash-merge
will collapse — `git grep PT-4465` across the whole branch is empty, the PR body doesn't mention
it, the ADR is silent. Please add an ADR entry (deferred capability + why focus state can never
answer this question) and a PR-body callout naming PT-4465, so the fourth attempt — by someone
with no memory of the first three — doesn't happen.

**2. File the wire-surface-snapshot ticket.** Your comment said it "needs a ticket rather than
riding this stack; flagging it here so it is not lost" — I checked Jira and there is no matching
issue. It's currently lost exactly the way you were guarding against.

**3. `adoptWebView` can delete a currently-docked view's live state on a hostile call.** The
seed-undo on provider decline/throw (`web-view.service-shard.ts` ~3669–3686) doesn't check
whether the target id already names a view this window holds docked. A shape-valid wire call
adopting that id overwrites its live state, then deletes it outright if the provider then
declines or throws — the exact loss `openOrReloadWebView`'s still-docked branch is built to avoid.
Unreachable through a legitimate move (capture always removes the source tab first), but reachable
under the same "reachable from any process, checks rather than trusts" threat model these
hardening commits were written for. Refuse the adopt (or at least skip the delete) when the id is
already docked here.

**4. The mid-move fold-in should dedupe against both id spellings.** `capturedDefinition.id` is
the only key `getAllOpenWebViewDefinitionsWithReachability` dedupes on, but the move record
already tracks two (`webViewIds`), and a landed-late adopt can answer with a renamed authoritative
id. During the late-adopt probe window (up to ~12s at the widened bounds) a whole-app read could
count the moving view twice. Cheap fix: dedupe against `move.webViewIds`.

**5. The wiring-latch `AsyncVariable` starts its 30s timer at router module import, not at first
wait.** A boot slower than that with no waiter yet rejects unobserved (unhandled-rejection noise
in main), and a caller arriving after that point gets an instant rejection even if wiring lands
moments later. Low severity — not boot-fatal — but worth a start-on-first-wait latch or a
pre-attached rejection handler so the timer means what callers assume it means.

**Closure — log-level asymmetry.** The owner search still logs mid-move at `warn` while the new
fold-in was deliberately downgraded to `debug` for the identical expected-condition reason
(`40109f9d029`). Pre-existing line, but your own rationale argues for `debug` there too — align
it while you're in the area.

**Closure — `Layout`/`FloatLayout` re-exported from `@papi/core` without `@experimental`.**
Accepted as correct: both are pre-existing stable types newly reachable by name, and the
experimental `'window'` arm is tagged where it's actually declared. No change needed.

rolfheij-sil:
🤖 Claude: All five asks are done, and both closures are settled. Two of them came out slightly
different from what you proposed, both times because the code said something the ask did not —
details below rather than buried in commits.

**1 — the focus-guard principle is now recorded.** `a06e47de863` adds
`adr-window-activation-is-declared-not-inferred` to `Architecture-Decisions.md`, and the PR body
now names PT-4465. The entry states the constraint as current fact (the question is whether a
person asked for this window, which only the caller knows), names both failure modes that make
focus state unable to answer it, and points at PT-4465 for the design. It is explicit that none of
the mechanism is wired yet — `createWindow` takes no intent flag today — so a reader looking for it
is sent to the ticket rather than left hunting the code for something that is not there.

One thing I scoped deliberately, because an unscoped version would have been actively harmful:
the entry is about **activating a new window**, not about raising an existing one. Focus state is
the right input for a raise and is used that way today — `isApplicationFocused()` gates the
cross-window open raise and the move raise, and `handleUri` is just as deliberately not gated on
it, for the reason its own comment gives. An entry that condemned focus state generally would read
as an argument against those guards, and whoever tidied that inconsistency would remove correct
code. The entry says so explicitly.

**2 — the wire-surface snapshot exists as a PR, not a ticket.** It is
[#2734](https://github.com/paranext/paranext-core/pull/2734), open since the 28th, and its body
cites your framing of it. So it was not lost, though your check was reasonable given it never
reached Jira — it went straight to a branch instead. Nothing further owed here unless you would
rather it be tracked in Jira as well.

**3 — the adopt is refused outright rather than only skipping the delete.** `910078226a5`. Taking
the weaker half of your suggestion would not have been enough: the seed runs *before* the provider,
so a bundle naming a docked view overwrites that view's live state whether or not the delete
later fires. The check therefore sits before the seed, after the waits — the dock can change while
`waitForLayoutLoadToSettle` parks, so it has to speak for the moment the seed happens rather than
the moment the adopt arrived, which is why the close guard beside it is doubled for the same reason.

It throws rather than answering `undefined`, because `undefined` already means "the provider
declined" and the recovery ladder absorbs that as an ordinary no. This is never ordinary. It also
means that if the ladder ever re-tries a window where a timed-out adopt actually landed, the adopt
is refused instead of putting one web view id live in two windows.

I did not add the second guard on the delete. To justify it I would need the id to become docked
*during* the provider call, and after this refusal an adopt is never a reload of an already-docked
view — which is the case `openOrReloadWebView`'s still-docked branch describes. I could not
construct a test that fails without it, so I left it out rather than add code no test can hold
honest.

**4 — the fold-in dedupes on both ids, and the fix is narrower than it looks.** `d0a8de93faf`. It
now matches every id the move tracks, which fixes the case where the window holding the view
answers with the spelling the move started from — a recovery back into the source window, once that
window has re-scoped it.

Worth being straight about what it does not fix: a *target* window that scopes the adopted view to
a suffix of its own answers under a third spelling that no move record holds, and that read would
still double-count. Matching stripped forms on both sides would catch it — and would also merge two
genuinely separate views that share an unscoped id, which Simple mode produces by loading one layout
into every window. Since the fold-in exists to stop a view going missing, I took the match that
cannot make the read wrong in the other direction. Say the word if you would rather have the broad
one.

**5 — the wiring bound now starts when a call waits.** `bbffac13360`. You were right about the
cause: `AsyncVariable` schedules its timeout in its constructor, so building the latch at module
scope started the clock at import. The latch is now built by the first call that actually waits, and
cleared again when a wait fails, so each waiting call gets the bound measured from its own arrival —
which also means nothing exists to reject when wiring simply beats every caller, so the unobserved
rejection cannot happen rather than being handled.

One residual: two callers waiting at once share a latch, so the second is bounded from the first's
arrival. Giving each its own means a per-call timeout race rather than a shared latch, which seemed
more machinery than a not-boot-fatal bound earns.

**Closure — log level aligned.** `d0a8de93faf`. The owner search's mid-move miss is `debug` now,
for the reason you gave; the refusal it reports is unchanged, since the level was only ever about
how loudly to say it.

**Closure — `Layout`/`FloatLayout`.** Agreed and nothing changed.

Every behaviour change above is covered by a test that fails without it, and each was checked by
mutation rather than by having been written: the adopt test fails if the guard moves below the
seed, and the latch test fails specifically if the clear-on-failure is dropped. Full battery green
at the tip — typecheck, lint, format, and the main and renderer service suites (1371 tests);
`papi.d.ts` regenerates unchanged, as expected for a round that touches no public surface.

tjcouch-sil:
Verified all five against the code at tip. Four are right, two of them better than what I asked
for. One argument does not hold, and it is the one you used to skip work — details below.

**Accepted, with credit where it's due:**

- **The ADR scoping is a genuine improvement on my ask.** I asked for an entry recording that focus
  state can't answer whether a person asked for a window; you scoped it to *activating a new*
  window and said outright that focus state remains the right input for a raise, naming the two
  guards and `handleUri`'s deliberate exception. You're right that an unscoped entry would have
  read as an argument against correct code, and the next person to tidy that "inconsistency" would
  have deleted the guards. Entry, failure modes, PT-4465 citations, and the "none of it is wired
  yet" statement all check out.
- **Refusing the adopt outright rather than only skipping the delete was the right read.** The
  seed runs before the provider, so my weaker suggestion would have left the overwrite intact and
  fixed only the second half. Throwing rather than returning `undefined` is also correct — the
  recovery ladder treats `undefined` as an ordinary provider decline.
- **The wiring latch** is structurally sound now: one construction site, one caller, awaited in the
  same microtask, so the unobserved rejection can't exist rather than being handled. The shared-latch
  residual is benign; leave it.
- **The fold-in, and a correction to me:** your narrow choice was right and my implied alternative
  was wrong. Deduping by stripped form even *among tracked ids* would make a moved `home-w2` match
  an unrelated `home-w4` and silently drop a genuinely-missing view — wrong in the same direction,
  just as a false negative. Both halves of your residual analysis verified: a target window really
  can mint a third spelling after re-scoping, and the Simple-mode collision is real (module-constant
  layouts hand every window the same baked ids).
- **#2734** — my "it's lost" was wrong; it went to a branch instead of Jira, which is fine. It's
  also better than the ask: demonstrating the blind spot with a deleted marker producing a
  byte-identical `papi.d.ts` beats asserting it, and covering the C# side wasn't requested.

**One ask back — the delete-path guard, which I'd like fixed here.** Your reason for leaving it out
was that after the new refusal, an adopt can never be a reload of an already-docked view. That's
true at the check and false at the delete: both deletes run after an unbounded `await` on
extension-host provider code, and nothing re-reads the dock in between. The layer directly beneath
already refuses that same delete under that same condition — `openOrReloadWebView`'s catch
preserves state when the id is still docked, on the grounds that deleting it "would gut a view the
user still sees" — and `adoptWebView`'s catch then deletes unconditionally, undoing it.

Two constructions where the id becomes docked *during* the provider call:

1. An adopt times out client-side but keeps running in the target window, parked in the provider.
   The probe gives up, and the recovery ladder's focused-window rung readopts into that same
   window. Adopt #2 passes the new guard, seeds, and calls the provider. Adopt #1 returns and docks
   the view live. Adopt #2 declines — and its cleanup deletes the live view's state.
2. A `platform.interfaceMode` change fires a layout load in every window, re-docking the persisted
   layout while an adopt is parked in its provider. Same ending.

Same two-line shape as the layer below: re-read `getDockLayout()` before each delete and skip when
the id is docked. On the "no test could hold it honest" point — the test is constructible (park the
provider, dock the id from the stand-in, release, assert no delete), but it needs
`shardOverDockLayout`'s `getWebViewDefinition` fixed first: it currently answers for *any* id, so
the new guard's test pins the guard's position but not that it checks the right id. Worth fixing
regardless.

**One to ticket, not to fix here.** The fold-in's `definitionIds` bookkeeping has the same
Simple-mode collision on the other side of the predicate: two concurrent moves whose captured ids
strip to the same value (`home-w2`, `home-w3` → both `home`) make the first fold-in suppress the
second, dropping a genuinely-missing view from that read. It's pre-existing — your commit didn't
introduce it — and it needs two simultaneous same-named Simple-mode moves, so a follow-up ticket
alongside the residual you already documented is the right home for it.

**Housekeeping:** the branch is CONFLICTING against main, and it's only the ADR log — `main`
appended `adr-node-dom-globals-polyfill` where this branch appended its two entries. Keep all
three, any order; nothing else conflicts.

tjcouch-sil (approved):
What's left seems so small that I'm going to go ahead and approve and let you take care of the last little bit on your own. Merge when finished unless you think I might disagree about something. Thanks!

rolfheij-sil:
🤖 Claude: Both asks are done, and your correction on the housekeeping note was the one thing that needed correcting in the other direction — the conflict was wider than the ADR log.

**The delete-path guard — you were right on all three points, including the one I had argued against.**

I re-read it before changing anything. The refusal at the top of `adoptWebView` speaks for the moment the adopt arrived; both cleanup deletes run after an unbounded `await` on provider code, and nothing re-read the dock in between. And the asymmetry you named is exactly as you describe: `openOrReloadWebView`'s catch already preserves state when the id is still docked, and `adoptWebView`'s catch then deleted unconditionally, undoing it one layer up.

Both deletes now go through a helper that re-reads `getDockLayout()` and skips when the id is present. One judgement call inside it worth flagging: a dock read that itself throws is treated as *still holding* the view, so the delete is skipped and a warning is logged. An orphaned state entry is recoverable and invisible to the user; a gutted live view is neither.

**On the test — you were right that the stand-in had to be fixed first, and it was worse than "answers for any id".** The move suite's dock stand-in was `getWebViewDefinition: () => webViewDefinition`, taking no parameter at all. It now answers only for the id its definition carries, and takes a setter so a test can dock a view while an adopt is parked in its provider.

Fixing it turned one existing test red, which was the right outcome: `capture closes the tab it captured` asked `captureAndCloseWebView` for the *unscoped* id and relied on the stand-in answering anyway. Its comment said it was pinning that "the close goes through the definition's own id rather than the id the caller asked with" — but capture looks the raw id up with no scoping, so in a real dock the definition's id always equals the id asked for, and that distinction cannot arise. It now asks with the id the dock holds, which still pins that capture closes the tab it captured.

**One thing I want to be precise about:** the new test *is* the reproduction of your two constructions, not a substitute for one. It parks the provider, docks the id from the stand-in while it is parked, releases, and asserts no delete. Construction 1 (a concurrent adopt lands) and construction 2 (a layout load re-docks) are the same state transition from different causes, so one test covers both mechanisms. I did not run either live — the measurement that would settle them empirically is this test, which is why it exists.

Mutation-checked both ways: forcing the guard never to skip fails the new test; forcing it always to skip fails the two existing tests pinning the delete on the decline path and the throw path — so it cannot pass by deleting nothing.

**The fold-in `definitionIds` collision:** verified and agreed. In Simple mode the layouts are module constants, so every window is handed the same baked ids; both moves' tracked ids then contain the same stripped form, and the first fold-in's bookkeeping makes the second look already-present. A ticket is drafted and waiting on Rolf; I will add its key here once it exists rather than name one that does not.

**One correction back, on the housekeeping note.** The branch conflicted with `main` in **four** files, not only the decision log: `.context/standards/Architecture-Decisions.md`, `src/main/main.ts`, `src/main/startup-tasks.ts` and `src/shared/data/rpc.model.test.ts`. Two measurements agreed — a test merge and a real rebase, which stopped nine times across six commits.

It matters because "nothing else conflicts" invites resolving by taking one side. The first `main.ts` hunk is an import block where both sides added a symbol, so the correct resolution is the union — taking either side alone drops an import and breaks the build. And `startup-tasks.ts` was a genuine edit-vs-edit: `main` moved the method-not-found matcher into `rpc.model.ts` while this branch moved the timed-out matcher there, each keeping the other locally. `rpc.model.ts` merged cleanly and now exports both, so the resolution was to import both and delete both local copies; `isMethodNotFoundError` is gone, superseded by `main`'s `isJsonRpcMethodNotFoundError`, which is the same matcher plus an optional `requestType` argument.

The decision-log entry is now identified by slug — `adr-main-orchestrates-real-windows` — which this branch had already done in a later commit. Two earlier renumber commits became empty and dropped out.

**Also worth knowing, since it was invisible to every check we run.** The rebase left `lib/platform-bible-react/dist/` internally inconsistent: only the entry bundles conflicted, so they were resolved from this branch while the content-hashed chunks beside them came from the new base, leaving `index.js` importing a chunk that does not exist in the tree. Typecheck, lint and the whole unit suite passed over it. Rebuilding the package from the merged source fixed it, in its own commit.

rolfheij-sil:
🤖 Claude: Fixed in `638ce7dc70c` on #2758.

The fold-in's cross-record bookkeeping is gone, and the guard that remains now matches the name the move was asked for rather than the stripped one it captured. Deleting that guard reddens exactly the double-count test; reverting it to the stripped match reddens exactly the two cases it exists for. So a first fold-in can no longer make the next read as already-present, and a window still reporting an adopted view under the stripped name can no longer hide an in-flight one.

Reviewing your finding turned up a second, older way to lose a view, which we fixed here rather than describe: a window holding both an adopted, unscoped `home` and its own `home-w2` — the state any window is in after receiving a move — dropped the in-flight view on a single drag, because the capture strips to `home` and the window still reported the other view under that name. That guard was byte-identical on `main`, so it predated this work; it is gone now, and pinned by a test that fails against the old predicate.

One residual, deliberate. While a move record is still open and the target already reports the view — a late-landing adopt, about nine seconds of probing — the view is reported twice. That reaches the shutdown read, where `getWritableEditorProjectIds` collapses it through a `Set`, and also `getAllOpenWebViewDefinitions` on the PAPI surface, which returns the list as-is. One thing to be clear about for anyone consuming that read: an extension must not deduplicate it by id. Ids are not unique across windows — every window's default Home tab reduces to `home`, and this branch's own test pins two live views both answering to `id: 'home'`, told apart only by the project each is showing — so deduping by id would drop a real view. Type-plus-project is not a sound key either: two windows showing the same project in the same web view type are legitimately identical under it. Until a view has an identity that survives a move there is nothing safe to collapse on, so the guidance is to tolerate a repeat. We still took a visible duplicate over a silent drop, because a drop is invisible to the caller and a duplicate is not.

The root of all of it is that a web view has no name that survives a move: the capture strips the source window's scope, the adopt reopens the view under that stripped spelling, and the next layout load re-scopes it — so one view wears three names across a move while every window's default Home tab shares one. The fix is a per-view identity minted at the adopt and carried on the definition, which is a change to the adopt path rather than to this read. It is written up in our small-items ledger as item A16, with the three failed in-read attempts and the diff of the closest one, so a fourth attempt starts from what is already known.

PT-4351 and PT-4352 are unscheduled for this epic.

PR #2672, opened 2026-08-13
fix(project-selector): row tooltip and filter dropdown render behind popover

Sebastian-ubs:
## Summary

Inside `ProjectSelector`, the row tooltip and the filter dropdown were portaled out of the selector's popover (both are Radix portals to `document.body`), so they stack against it purely by z-index. Two mismatched constants meant they lost that fight:

- Shadcn `PopoverContent` defaults to `Z_INDEX_ABOVE_DOCK` (600).
- `ProjectSelector` pinned both the row `TooltipContent` and the `FilterMenu`'s `DropdownMenuContent` to `Z_INDEX_OVERLAY` (400).

Result: hovering a row or opening the filter menu produced overlays that appeared **behind** the popover.

Fix: introduce a local `Z_INDEX_ABOVE_SELECTOR_POPOVER = Z_INDEX_ABOVE_DOCK + 50` and use it for both overlays. Mirrors the `+ 50` pattern already used in [src/renderer/components/projects/project-picker.component.tsx](https://github.com/paranext/paranext-core/blob/main/src/renderer/components/projects/project-picker.component.tsx) (`Z_INDEX_MODAL + 50`) and stays below `Z_INDEX_FIRST_RUN` (700).

Considered but not done: hoisting the constant into [lib/platform-bible-react/src/components/z-index.ts](https://github.com/paranext/paranext-core/blob/main/lib/platform-bible-react/src/components/z-index.ts). Only two callsites in one file needed it, so I kept it colocated with a comment explaining the layering. Happy to promote it if a reviewer prefers.

## Test plan

- [ ] `npm --workspace lib/platform-bible-react run storybook` → `Advanced/Project Selector`. Open any story that has multiple rows, hover a row: tooltip appears above the popover. Open the filter icon: dropdown appears above the popover.
- [ ] Repeat in `MultiProject` for the "Show selected only" toggle inside the filter dropdown.
- [ ] Confirm no regression in the manage-books dialog consumer (`extensions/src/platform-scripture/src/manage-books.web-view.tsx`) — the popover still sits above the dialog, and its nested overlays now sit above it too.
- [ ] Typecheck (`npx tsc --noEmit -p lib/platform-bible-react/tsconfig.json`) and lint (`npm --workspace lib/platform-bible-react run lint`) — both clean.

AI-assisted.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2672)

Screenshot 2026-08-13 153718
Screenshot 2026-08-13 153711

PR #2673, opened 2026-08-13
feat(project-selector): grouping options (open tabs, last used, language, versification, type)

Sebastian-ubs:
# Code Review Summary

**Branch**: pr-2673-handover (tracks `sebastian/feat/project-selector-more-grouping-options`, PR #2673)

**Base**: origin/main (merge-base `aec038b3f63`)

**Date**: 2026-09-14

**Files changed**: 54 (35 source + 19 generated `dist/`)

## Overview

Reworks `ProjectSelector`'s grouping support from a fixed set of component-owned props into a
consumer-driven descriptor API: a consumer passes `availableGroupings` — an array of
`ProjectSelectorGrouping` objects, each saying how to bucket a row, what to call the bucket, how to
order buckets, and what to do with rows it cannot classify — and the row data travels in an untyped
`customData` bag packed by `makeProjectSelectorCustomData` in `platform-bible-utils`. The component
never interprets `customData`; only the descriptor supplied by the same consumer reads it. Every
in-repo picker (checklist, checks side panel, Find, manage-books, settings sidebar) is then wired to
the groupings its own data can actually populate, via a per-surface `*_GROUPING_IDS` allow-list that
a new coverage test mechanically enforces. Roughly 16 public props and 6 localized-string fields are
removed, and two localization keys renamed, with no deprecation cycle — sanctioned by
`experimental.ts`'s own stability disclaimer.

**This is a handover branch.** It carries 36 commits authored by someone else; the reviewing
developer picked it up rather than writing it. The interview was therefore framed around what has
been verified versus inherited, not around design intent. All four analysis passes rated the branch
**no critical issues**, and all four independently praised the same thing: the grouping-coverage
test is a genuinely good answer to the type hole that an untyped `customData` bag opens, and it is
unusually honest in its own comments about what it does not cover.

Twenty findings were addressed with code changes during this review; one Important finding was
verified as a false positive and dismissed; one is deliberately left open as a follow-up.

## API Changes

`lib/platform-bible-react` (`platform-bible-react/experimental` — `ProjectSelector`):

- **Added exports**: `type ProjectSelectorGrouping`, `type ProjectSelectorLocalizedStringKey`,
  `type ProjectSelectorStringLookup`, `type BuiltInGroupingStrings`, `type SelectionGroupingStrings`,
  `PROJECT_SELECTOR_STRING_KEYS`, `buildProjectSelectorLocalizedStrings`,
  `buildBuiltInGroupingStrings`, `buildSelectionGroupingStrings`, `makeBuiltInGroupings`,
  `makeSelectionGrouping`, `defaultGroupings`, and (added during this review) `NO_GROUPING`.
- **Added props** on `ProjectSelectorProps`: `availableGroupings?: readonly ProjectSelectorGrouping[]`,
  `defaultGrouping?: string | 'none'`.
- **Added** `ProjectSelectorProject.customData?: Readonly<Record<string, unknown>>`.
- **Removed props** (BREAKING): `showSelectedOnly`, `onChangeShowSelectedOnly`,
  `filterShowSelectedOnly`, `selectAll`, `groupByVersification`, `priorityVersificationId`,
  `versificationUnknownSectionHeading`, `defaultGroupByOpenTabs`, `filterGroupByOpenTabs`,
  `groupByOpenTabs`, `onChangeGroupByOpenTabs`, `filterSectionLabel`, `hideFilterMenu`,
  `buttonPlaceholder`, `commandEmptyMessage`, `ariaLabel`, `hideTriggerChevron`, `alignDropDown`,
  `popoverContentClassName`, `popoverContentStyle`. (`buttonClassName` was restored.)
- **Modified** `ProjectSelectorLocalizedStrings`: removed `filterAriaLabel`, `filterSectionLabel`,
  `filterGroupByOpenTabs`, `filterShowSelectedOnly`, `versificationUnknownSectionHeading`,
  `selectAll`; added `ariaLabel`, `buttonPlaceholder`, `commandEmptyMessage`, `groupByAriaLabel`,
  `groupByNone`, `autoOpenTabsGroupingLabel`, `autoSelectionGroupingLabel`,
  `autoSelectionSelectedSectionHeading`, `autoSelectionUnselectedSectionHeading`.
- `SettingsSidebar`: no prop change (internal rewiring only).

`lib/platform-bible-utils`:

- **Added exports**: `PROJECT_SELECTOR_CUSTOM_DATA_KEYS`, `makeProjectSelectorCustomData()`,
  `recencyMapFromOrderedIds()`, `type ProjectSelectorCustomDataShape`.

**Localization (BREAKING)**: `%projectSelector_filterAriaLabel%` → `%projectSelector_groupByAriaLabel%`,
`%projectSelector_filterGroupNone%` → `%projectSelector_groupByNone%`. Nine new
`%manageBooks_projectType_*%` keys were added during this review (en only).

`lib/papi-dts/papi.d.ts`: no changes required — neither library is declared there.

None.

- [x] **The checks side panel fetched localized strings and then ignored them.**
  `checks-side-panel.component.tsx` added `...PROJECT_SELECTOR_STRING_KEYS` to its key array with a
  comment asserting the caller "merges its own placeholder/ariaLabel on top", but never called
  `buildProjectSelectorLocalizedStrings` — so ~9 picker-chrome strings fell through to hardcoded
  English for a Spanish user, in a panel that had already fetched the Spanish values. Partly
  pre-existing (merge-base passed only the same three overrides) but widened here, because the new
  grouping menu adds new user-visible chrome. _(fixed during review: spread
  `buildProjectSelectorLocalizedStrings(localizedStrings)` before the three panel-specific
  overrides, mirroring `find.component.tsx` exactly; the line-47 comment is now true)_
- [x] **Deprecated localization keys were still being fetched at runtime.** Eight
  `%manageBooks_projectSelector_*%` keys and three Find keys remained in their fetch arrays with no
  remaining readers (verified per key by grep across `extensions/src`, `lib/*/src`, `src`), while
  `localization.service-host.ts:330-334` logs a `logger.warn` per deprecated key — ~8-11 warnings
  per manage-books/Find mount for strings never rendered. _(fixed during review: ten dead keys
  removed from the two arrays; the two versification keys, which are genuinely still read, kept)_
- [x] **Four sibling keys were retired a different way in the same file.**
  `%manageBooks_projectSelector_filterAriaLabel%`, `_filterGroupByOpenTabs%`, `_filterSectionLabel%`
  and `_filterShowSelectedOnly%` were deleted outright with no `metadata.deprecationInfo`, while
  eleven siblings in the same change got a full deprecation entry with `fallbackKey`. _(fixed during
  review: `deprecationInfo` added for all four, three with a `fallbackKey` to the superseding
  platform key, and `filterShowSelectedOnly` with an explanation instead since it has no successor)_
- [x] **The manage-books `type` grouping shipped raw PT9 enum values as user-visible section
  headings.** `typeName: p.projectType` put developer tokens (`Standard`, `BackTranslation`,
  `StudyBibleAdditions`, …) straight into section headings — untranslated, PascalCase, outside the
  localization system, and in violation of the sentence-case rule. `'type'` is genuinely offered
  (`MANAGE_BOOKS_PROJECT_SELECTOR_GROUPING_IDS:342`), so this was live in the Copy "From" picker and
  the sidebar. _(fixed during review: nine `%manageBooks_projectType_*%` keys added (en only), a
  `PROJECT_TYPE_LOCALIZE_KEYS` map and a memo resolve them consumer-side — mirroring the
  versification grouping that already did this correctly — and `toManageBooksSelectorRows` takes an
  optional `projectTypeNames` map, keeping the raw value as a fallback so an unmapped PT9 type still
  gets a heading)_
- [x] **The default grouping was resolved once at mount, contradicting its own documented contract.**
  `availableGroupings` is a `useMemo` and recomputes, but `activeGrouping` was a `useState`
  initializer. A consumer that omits `availableGroupings` and loads `openTabs` asynchronously
  resolved to `'none'` and never re-resolved — the `openTabs` radio appears in the menu while the
  list stays flat. Latent rather than live (no in-repo picker hits it), but it replaced an explicit
  `defaultGroupByOpenTabs` prop with mount-timing-dependent behavior, and the most likely consumer
  to hit it is out-of-repo. _(fixed during review: `pickedGrouping` state tracks whether the user has
  chosen; until then a `resolvedGrouping` memo keeps following its inputs. Two regression tests
  added. **Honesty note:** only the first discriminates — mounting with empty `openTabs`, then
  supplying them, then asserting the section heading appears, fails against the unfixed component
  (verified by reverting the fix: it times out at 5s). The second, which asserts a user's explicit
  choice survives a later prop change, passes both before and after; it guards against
  over-correcting into re-resolving on every change, and is not evidence the fix works.)_
- [x] **The `customData` key constants only bound the writer half.**
  `PROJECT_SELECTOR_CUSTOM_DATA_KEYS` documents "reference these constants … so a rename here
  surfaces at every callsite", and `makeProjectSelectorCustomData` honoured it — but the reader side
  (`project-selector.groupings.ts`) hardcoded `customData?.lastUsedAt`, `.language`, `.type`,
  `.typeName`, so a rename would have broken the groupings silently. _(fixed during review: the
  built-in groupings now read through the constants)_
- [x] **The component hand-built two grouping descriptors instead of using the shared factories.**
  _(fixed during review, with the finding's premise corrected: the analysis claimed this would
  "collapse four props off the public surface", but the component receives the **resolved**
  `ProjectSelectorLocalizedStrings` bag, not the raw `%key%` lookup the string builders need — so the
  four `auto*` fields are the only channel for localizing the auto-derived path and deleting them
  would break it. The narrower real fix was applied: a new `makeOpenTabsGrouping` factory, used by
  both `makeBuiltInGroupings` and the component's auto-derive path, plus `makeSelectionGrouping` at
  the call site, so each descriptor has one definition)_
- [x] **245 lines of new public grouping API had no test file.** No `project-selector.groupings.test.ts`
  existed; `rows.test.ts` only exercised locally-built descriptor literals. The consequential gap was
  ~20 key-string→field mappings in the three `build*Strings` functions, where a mistyped key silently
  degrades to English with nothing failing. _(fixed during review: new `project-selector.groupings.test.ts`, 18 tests,
  zero lint suppressions. The highest-value assertions feed the builders a lookup whose value for
  every key IS that key, so each produced field names the key it was built from and a mistyped key
  fails loudly instead of degrading to English. Also pins the `type` heading rule (first non-empty
  `typeName` wins, else the raw key), `lastUsed`'s `compareSections` ordering, and the
  presence-flag-not-sort-key semantics. The two `!` non-null assertions this initially needed were
  removed by restructuring into throwing helpers rather than disabling the lint rule)_
- [x] **The recency wiring was duplicated verbatim across four web views**, comments included,
  including the load-bearing "normalize BOTH sides" gotcha that four copies each had to get right.
  _(fixed during review: extracted `useProjectRecencyMap(logLabel)` into
  `extensions/src/platform-scripture/src/hooks/`, alongside `use-open-project-tabs.ts`; the only
  per-site difference was the log prefix, which is parameterized, so no site's behavior changed)_
- [x] **The "Select all" deprecation message records a reason that is factually wrong.** Surfaced by
  the author's question during the interview, not by the analysis passes. Two different
  justifications were shipped: the localization manifest said *"it was redundant with the Selection
  grouping"*, while the component comment said *"selecting every project mounts a data subscription
  per project … a performance hazard"*. The first is false — `partitionBySelection` only filters and
  sorts rows into two display buckets; it selects nothing, so it cannot substitute for a bulk-select
  action. The manifest message is what a downstream extension author sees when the deprecation
  warning fires, so it was the only explanation most readers would ever find. _(fixed during review:
  the manifest message now states the real, performance-based reason)_
- [ ] ~~`manage-books-sidebar` lost `buttonClassName` (`tw:h-8`/`tw:font-normal`) with no
  replacement, leaving it inconsistent with `settings-sidebar`.~~ _(Verified false positive. The
  removed string was `tw:h-8 tw:font-normal` + `tw:w-full` + narrow `tw:px-0.5 tw:text-xs`, and the
  component's base className at `project-selector.component.tsx:1005` now carries
  `tw:flex tw:h-8 tw:w-full tw:min-w-0 tw:shrink! … tw:font-normal` with
  `isTriggerNarrow && 'tw:px-0.5 tw:text-xs'`. Complete migration, not a drop. Recorded so it is not
  re-raised.)_
- [ ] **"Clear all" is a bulk destructive action with no recovery path.** It wipes the whole
  multi-selection with no confirmation and no undo, and this PR removed "Select all", which was the
  only affordance that could restore a large selection. `.claude/rules/ux/apply-model.md` asks for a
  recovery path on anything "destructive, costly, hard to reverse, bulk" — Clear all is all four.
  **Deliberately left open as a follow-up at the author's direction.** Note that the performance
  argument for removing Select all does *not* argue against this: an undo that re-selects the N pairs
  that were subscribed a moment earlier adds no new subscription load. Suggested shape: the button
  flips to "Undo (N)" until the popover closes.

[Author response: The author directed fixes for every Important finding except the Clear all undo,
which they asked to be recorded as a potential follow-up. They also challenged the framing of the
Select all removal, which is what surfaced the incorrect deprecation message above — a finding none
of the four analysis passes caught.]

### Minor — Consider

- [x] **The group-by trigger lost its active-state affordance** (`aria-pressed` + highlight) relative
  to merge-base, so neither sighted nor screen-reader users could tell from the button whether a
  grouping was active. _(fixed during review: `aria-pressed` restored with the merge-base styling;
  `GroupByMenu` already received `activeGrouping`, so nothing new was threaded through)_
- [x] **The group-by icon button used a native `title` tooltip**, 400 lines after the same file
  explicitly rejects `title` for the main trigger as "inconsistent with the app's shadcn tooltip
  styling". _(fixed during review: converted to the shadcn `Tooltip`. No stacking change was needed
  or made — `TooltipContent` sets the tooltip tier itself, and the documented
  `Z_INDEX_ABOVE_POPOVER` override on `DropdownMenuContent` is untouched)_
- [x] **The Open button set `title` identical to its visible label** — the exact "redundant tooltip"
  case in `Guidelines/Tooltips`. _(fixed during review: `title` removed; `aria-label` retained)_
- [x] **`'Selected'`/`'Unselected'` English defaults were hardcoded in three places.** _(fixed during
  review: two exported constants in `project-selector.rows.ts`, the lowest layer, consumed by all
  three; no import cycle)_
- [x] **`NO_GROUPING` was unexported and `'none'` undocumented as reserved**, so consumers had to
  hardcode the literal and a consumer grouping with `id: 'none'` would silently collide with the None
  radio. _(fixed during review: exported, re-exported from `experimental.ts`, and `'none'` added to
  the reserved-ids TSDoc alongside `'openTabs'` and `'selection'`)_
- [x] **`recencyMapFromOrderedIds` last-wins on duplicate ids**, assigning the *lowest* score to the
  most recent entry; duplicates are reachable because callers pass `.map(normalizeProjectId)`.
  _(fixed during review: first occurrence now wins, TSDoc states the rule, and a new test covers it —
  the test fails against the old behavior)_
- [x] **Backward-facing review-finding IDs in comments** — `I1`, `I2`, `I12` across 12 sites in both
  `lib/` and `extensions/`, plus two `describe()` titles. These are this PR's own development
  artifacts, not tickets. _(fixed during review: all 12 rewritten to state the durable fact; the
  forward-facing `normalizeProjectId` pointers were kept and only the bare IDs dropped. A broad
  `\bI[0-9]+\b` sweep confirms none remain)_
- [x] **Comments claimed recency "orders" the `lastUsed` grouping**, when `lastUsedAt` is a pure
  presence flag whose magnitude is never compared. _(fixed during review: corrected at every site,
  matching the wording sites that already had it right. A deterministic sweep found four further
  hits that were deliberately left — they describe the recents *service's* list order, which genuinely
  is ordered. The presence-flag semantics is now pinned by a test)_
- [x] **The coverage test's "there is no list to forget to update" claim overstated the mechanism** —
  the checklist appends `makeSelectionGrouping` outside the filtered list, and the Create "Based on"
  picker passes a bespoke `versification` grouping, so neither reaches the suite. _(fixed during
  review: the NOT-COVERED block now names both and the claim is scoped to what it actually
  guarantees. No assertion was weakened; still 9/9)_
- [x] **No ADR was recorded for the headline decision**, which `CLAUDE.md` requires for exactly this
  class of choice. _(fixed during review: `adr-project-selector-consumer-driven-groupings` added at
  its byte-order slug position — verified, and the whole file re-verified as byte-ordered — covering
  the descriptor API, the untyped-`customData` tradeoff, the rejected alternatives, and the fact that
  the compensating coverage test is extension-scoped and cannot see the same defect introduced inside
  the library)_
- [ ] **`partitionByGrouping` silently drops rows** when `getGroupKey` returns `undefined`/`''` and
  the grouping sets no `unknownSectionHeading` — including from search results. Documented, and no
  in-repo grouping hits it, but a consumer-defined one makes projects vanish. Left open deliberately:
  changing the default is a semantic change to a documented API and is a reviewer decision, not a
  cleanup.
- [ ] **`defaultGroupings` (camelCase) reads oddly** beside `PROJECT_SELECTOR_STRING_KEYS` and
  collides conceptually with the new `defaultGrouping` **prop**. Left open: renaming a published
  export is itself a breaking change and belongs in the same decision as the rest of the API surface.
- [ ] **`recencyMapFromOrderedIds` is exported unprefixed** into a ~200-export flat namespace despite
  existing solely to feed `ProjectSelectorProject.customData`. Left open for the same reason.
- [ ] **`%projectSelector_groupByAriaLabel%` and `%projectSelector_groupSectionLabel%` are
  byte-identical** in both locales ("Group by" / "Agrupar por"), doubling translator work for no user
  benefit. Left open: merging them is a localization-API decision.
- [ ] **Selecting the first project inserts the Clear-all row**, shifting every row down under the
  pointer mid-multi-select; and with the Selection grouping active, clicking a row re-sorts the list
  under the pointer. Neither is a regression, but the second is now the default multi-select
  experience rather than an opt-in. Left open as design questions.
- [ ] **`makeSelectionGrouping` has no Storybook story** despite being newly public; it appears only
  via the auto-derive path in `MultiProject`. Its behavior is now covered by unit tests, so this is
  documentation rather than coverage.
- [ ] **`experimental.ts` re-exports the grouping symbols through the component's pass-through block**
  rather than importing from `project-selector.groupings` directly. Cosmetic.

[Author response: The author asked for all relevant minor findings to be fixed. Ten were; the seven
left open are those where the "fix" is a breaking rename, a semantic change to a documented API, or a
design decision that belongs to the reviewer rather than to a review pass.]

#### Shared Regions Modified

None. All four analysis passes swept the changed files for `#region shared with` markers; no hits.

#### Extension Config Changes

None. No `package.json`, `tsconfig*.json`, `webpack.config.*`, or `.eslintrc*` under `extensions/`
was changed, and no root build/CI config changed.

## Positive Observations

- `project-selector-grouping-coverage.test.ts` is the standout. Because `customData` is untyped,
  nothing in the type system connects "this picker offers Language" to "this picker packs a
  language". Driving each surface's own exported `*_GROUPING_IDS` constant through its own row
  builder turns a dead menu item into a test failure rather than a UI-only bug. It is also unusually
  honest about its own limits — it names what it does not cover, explains why it reproduces
  `partitionByGrouping`'s bucketing rather than widening `experimental.ts` for one test, and names
  the two ways its local partitioner diverges plus why neither can change a verdict.
- The `*_GROUPING_IDS` allow-lists each carry a specific, data-grounded justification for every
  built-in they **omit** (Find: "the recents list is capped at 5"; manage-books: "one
  project-settings round-trip per row"), rather than silently offering everything.
- Localization migration was complete and mechanically verifiable: every `%projectSelector_*%`
  literal read anywhere in the component or groupings module is declared in
  `PROJECT_SELECTOR_STRING_KEYS` and present in **both** `en.json` and `es.json` with real Spanish —
  zero read-but-undeclared, zero declared-but-unread, zero missing translations.
- `%markersChecklist_toolbar_comparativeTexts%` → `%..._comparativeProjects%` is handled exactly
  right for a *meaning* change: a new key, the old one retained with `deprecationInfo` +
  `fallbackKey`, and the type union updated.
- Z-index handling is correct against `.claude/rules/ux/z-index-tiers.md`: zero new overrides in
  `src/`, and the one surviving `Z_INDEX_ABOVE_POPOVER` override carries exactly the call-site
  justification rule 2 asks for, naming the surfaces it was verified against.
- `hideTriggerChevron`/`buttonClassName` were migrated properly rather than dropped — the narrow-rail
  styling now lives in the component behind a measured `NARROW_TRIGGER_THRESHOLD_PX`, with the
  `ResizeObserver`-absent case degrading to a one-shot measurement instead of throwing, and a comment
  explaining why it measures border-box rather than `contentRect`.
- `availableGroupings` omitted vs `[]` vs length-1 vs length-≥2 is a four-way contract, and all four
  branches are spelled out in the prop's TSDoc with the concrete consumer that motivates each.
- Five new Storybook stories cover the grouping surface specifically, so the descriptor API's
  contract is demonstrable in the browser.

## Interview Notes

**This is a handover branch.** `pr-2673-handover` tracks a fork remote and is exactly in sync with
PR #2673's head; the 36 commits were authored by someone else. The reviewing developer did not write
this code and did not walk through its design — which is the expected and correct situation for a
handover, but the reviewer should know it rather than read the interview as a design defence. No
design-understanding question was put to the author for that reason.

The author's participation was decisive and directive: they approved fixes for every Important
finding except one, and asked for all relevant minor findings to be fixed.

**The single most valuable moment was the author's own question.** Asked whether to implement a
Clear-all undo, they instead asked *"Why was Select All removed?"* Chasing that down revealed that
two contradictory justifications had shipped — a correct performance argument in a code comment, and
a false "redundant with the Selection grouping" claim in the localization deprecation manifest, which
is the one a downstream extension author would actually see. **None of the four analysis passes
caught this.** It was found because the author declined to accept a framing and asked for the
underlying reason.

The author also chose to leave the Clear-all undo as a follow-up rather than take it in this PR —
a scoping call, not a gap in understanding.

**Corrections made to the analysis during review** (recorded so the reviewer does not re-raise them):

- The UX pass claimed the PR description overstated a z-index fix. It did not: #2672 merged
  2026-09-11, *before* this branch's merge-base, which is exactly what the description says.
- The style pass claimed `manage-books-sidebar` lost trigger styling with no replacement. It did not;
  the component internalizes every one of those classes.
- The style pass claimed removing four `auto*` string fields would simplify the public surface. It
  would have broken localization of the auto-derived grouping path, because the component receives the
  resolved strings bag rather than the raw key lookup.

**Not verified by anyone, and worth a reviewer's attention**: the manual Storybook sweep. The
automated story tests pass, but the grouping menus have not been stepped through by hand in a
browser, so a purely visual regression that no test covers could still be present. This was already
flagged in the PR description and remains true.

## In-Review Quality Check

All commands below were re-run by the reviewer directly, not taken from subagent reports.

| Check | Result |
|---|---|
| `npm run typecheck:workspaces` | **exit 0** |
| `npm run typecheck:core` | **exit 0** |
| `npm run lint --workspace extensions` | **exit 0** |
| `npm run lint --workspace lib/platform-bible-react` | **exit 0** |
| `npm run lint --workspace lib/platform-bible-utils` | **exit 0** (1 pre-existing warning, 0 errors) |
| `npm --workspace platform-scripture run test` | **563 passed / 30 files** |
| `npm --workspace lib/platform-bible-utils run test` | **623 passed, 1 skipped / 33 files** |
| `npm --workspace lib/platform-bible-react run test` | **1509 passed / 163 files, npm exit 0** |

Both libraries were rebuilt and `dist/` staged, since extensions compile against the committed build
rather than source; the new `NO_GROUPING` export was confirmed present in `dist/experimental.js`
rather than assumed.

**Two issues resolved during the quality pass, both worth knowing about:**

1. **A Storybook dependency-cache failure that was not a stale cache.** Three story files failed at
   import with `does not provide an export named 'PROJECT_SELECTOR_CUSTOM_DATA_KEYS'`. The cache
   regenerated with an identical hash and the identical error, and both `dist/index.js` and
   `dist/index.cjs` were confirmed to export the symbol — so the diagnosis had to be corrected. The
   real cause was a second cache at `lib/platform-bible-react/node_modules/.cache/storybook` (not the
   repo-root one). After clearing it: 163/163 files, 1509 tests, exit 0. Worth knowing because any
   contributor adding a new cross-package export will hit exactly this and it presents as a source
   error.
2. **`typecheck:erb` / `lint:scripts` fail on `spdx-license-list`, and the PR description's account of
   it needs sharpening.** The description tells reviewers the root `npm run lint` is an `&&` chain
   that fails at its first link and therefore silently skips the workspace lints — that part is
   correct and worth keeping. But the underlying `TS2307` is not a pre-existing *code* problem:
   `spdx-license-list` **is** declared in `package.json` (6.12.0) and is simply **not installed** in
   this checkout, and the branch touches neither `.erb/` nor `package.json`. So CI with a clean
   install would very likely be green here, and the failure should not be carried into the PR as a
   known defect.

## Suggested Review Focus

- [ ] **The breaking change, and who absorbs it.** ~16 props and 6 string fields removed with no
      deprecation cycle. `experimental.ts` sanctions this in its own contract and nothing in-repo
      still passes the removed props, but **Paratext 10 Studio is out-of-repo and was not checked**.
      Confirm whether it consumes `ProjectSelector`, and if so who lands the migration and when.
- [ ] **The central design tradeoff: untyped `customData` + a convention, instead of types.** This is
      the decision the whole PR rests on, and it is now recorded in
      `adr-project-selector-consumer-driven-groupings`. The known residual risk is stated there and
      is worth discussing explicitly: the grouping-coverage test lives in `platform-scripture`, so a
      data-reading grouping added *inside* `platform-bible-react` — e.g. to the settings sidebar,
      whose public `ProjectInfo` prop carries only an id and a name — would be dead on arrival and
      invisible to the test. The durable fix is a check inside `ProjectSelector` itself. Is that
      worth doing now, or is the allow-list convention sufficient?
- [ ] **"Clear all" with no undo** — the one Important finding deliberately left open. Decide whether
      the "Undo (N)" affordance lands in this PR or as follow-up work.
- [ ] **The manual Storybook sweep**, which nobody has done. The grouping menus have not been
      exercised by hand in a browser.
- [ ] **The seven open minor findings**, each left open because the fix is a breaking rename, a
      semantic change to a documented API, or a design call — specifically `partitionByGrouping`
      silently dropping unclassifiable rows, and the two duplicate-value localization keys.
- [ ] **Handover context.** The reviewing developer did not author these 36 commits. Areas worth
      walking through together rather than assuming: the row/partition model in
      `project-selector.rows.ts`, and the four-way `availableGroupings` contract (omitted vs `[]` vs
      one vs many), since those are what a future consumer will get wrong first.

---

Reworks `ProjectSelector`'s grouping support into a consumer-driven API and wires every in-repo picker to the groupings its own data can actually populate.

This description replaces the original August one, which described the pre-rework design (typed `type`/`lastUsedAt` fields, a "Show selected only" checkbox). Neither of those is what lands here. (The `defaultGrouping` prop that description also mentioned *does* land — `project-selector.component.tsx:321`, resolved at `:638-644`.)

## The grouping API

Groupings are no longer an enum the component owns. A consumer passes `availableGroupings` — an array of `ProjectSelectorGrouping` **descriptor objects**, each of which says how to bucket a row, what to call the bucket, how to order buckets, and what (if anything) to do with rows it can't classify.

Row data travels in an untyped `customData` bag on each `ProjectSelectorProject`, packed by `makeProjectSelectorCustomData` in `platform-bible-utils`. The component never interprets `customData`; only the descriptor that the same consumer supplied reads it. That is what lets a surface offer "Language" without the component needing a language field, and what lets a surface omit a grouping it has no data for.

`makeBuiltInGroupings` / `buildBuiltInGroupingStrings` build the five stock descriptors (`openTabs`, `lastUsed`, `language`, `type`, plus `selection` via `makeSelectionGrouping`) from the shared `%projectSelector_*%` localization block, so consumers wire them in one line and then filter to the ids they can populate.

The string builders take a loose `%key%`-indexed record, so a consumer hands `useLocalizedStrings`' return value straight through with no narrowing or type assertion.

## BREAKING: public props removed

Relative to `main`, these `ProjectSelectorProps` members are gone:

| Removed | Superseded by |
| --- | --- |
| `showSelectedOnly`, `onChangeShowSelectedOnly`, `filterShowSelectedOnly` | the built-in `selection` grouping, which buckets rows into Selected / Unselected |
| `selectAll` | the `selection` grouping (see the note below) |
| `groupByVersification`, `priorityVersificationId`, `versificationUnknownSectionHeading` | a consumer-supplied descriptor |
| `defaultGroupByOpenTabs`, `filterGroupByOpenTabs`, `groupByOpenTabs`, `onChangeGroupByOpenTabs` | the `openTabs` built-in descriptor |
| `filterSectionLabel` | `%projectSelector_groupSectionLabel%` |
| `hideFilterMenu` | omit `availableGroupings` (or pass `[]`) |
| `buttonPlaceholder`, `commandEmptyMessage`, `ariaLabel` | the same-named fields inside the `localizedStrings` bag — move `ariaLabel="…"` to `localizedStrings={{ ariaLabel: "…" }}` |
| `hideTriggerChevron`, `alignDropDown`, `popoverContentClassName` | — (presentation is now the component's) |
| `popoverContentStyle` | — (presentation is now the component's) |

`popoverContentStyle` did have an in-repo caller: the settings sidebar passed `popoverContentStyle={{ zIndex: Z_INDEX_OVERLAY }}` (`settings-sidebar.component.tsx:177`) to pin the popover to z-400. #2672 already dropped that override on `main`, and removing the prop makes the behavior unconditional: the popover now takes the `PopoverContent` default of `Z_INDEX_ABOVE_DOCK` (600) instead of 400, so it stacks above the dock rather than behind it. That is a fix, not a regression.

**"Select all" was removed for a second, independent reason**: selecting every project mounts a data subscription per project, which is a real performance hazard on a large installation. It is not coming back as a bulk toggle.

`buttonClassName` was **restored** — the rework dropped it, which silently regressed the settings sidebar's button layout.

## BREAKING: localization keys renamed

- `%projectSelector_filterAriaLabel%` → `%projectSelector_groupByAriaLabel%`
- `%projectSelector_filterGroupNone%` → `%projectSelector_groupByNone%`

Four dead `%manageBooks_projectSelector_filter*%` keys were removed.

**User-visible string change (markers checklist).** The comparative-texts picker's trigger label and aria-label now read **"Select comparative projects"** instead of "Select comparative texts", because the picker selects projects. Per the immutability rule, the existing `%markersChecklist_toolbar_comparativeTexts%` key keeps its original English value and gains a `fallbackKey` stub; the new value lives on a new key, `%markersChecklist_toolbar_comparativeProjects%`, which is what the web view now reads. Existing translations of the old key continue to resolve for anyone still referencing it.

## Per-surface grouping matrix

Each picker offers only the groupings its own row builder can populate. Exclusions are deliberate, not oversights:

| Surface | Offers | Why not the rest |
| --- | --- | --- |
| Find (default) | `openTabs`, `language` | No project-type source on the wire. `lastUsed` excluded because Find lists only **open** projects against a 5-entry recents cap, so the grouping degenerates to a single bucket. |
| Find (Simple interface) | `language` | Same exclusions, plus `openTabs`: this branch passes an empty `openTabs` list to suppress scroll-group badges, so the open-tabs section has no eligible row and the grouping collapses to one undifferentiated bucket. |
| Checks side panel | `openTabs`, `lastUsed`, `language` | No project-type source. |
| Manage-books sidebar | `openTabs`, `lastUsed`, `type` | No language on the wire, and the per-project fan-out that would fetch it was deliberately removed for performance. |
| Checklist (both pickers) | `openTabs`, `lastUsed`, `language` | No project-type source. The comparative-texts picker additionally offers `selection`. |
| Manage-books Copy "From" | `openTabs`, `type` | Recency would cost a re-fetch on every recents change while the dialog is open. |
| Settings sidebar | none | `platform-bible-react` is PAPI-free by design, so the sidebar has no source for any grouping dimension. |

## Fixes

**Markers checklist panel crash.** `useData('platformScripture.recentlyOpenedProjects').RecentProjects` yields `string[] | PlatformError`. The unguarded value reached a helper that calls `.forEach`, throwing during render and blanking the panel. Now guarded with `isPlatformError`; the grouping degrades to "no recency" rather than crashing.

**Project-id casing.** The recents service stores project ids verbatim, while canonical ids are uppercase and the open-tabs hook lowercases them. A raw lookup therefore matched nothing and silently routed every project into "Other". Both sides now normalize before comparing.

**Comparative-texts grouping.** Open-tabs grouping is restored on the comparative-texts picker rather than disabled. `project-multi` keys rows by `(projectId, scrollGroupId)`, but comparative-texts storage is per-project with no scroll group, so a stored ref matched neither row of a project open in two scroll groups: both rendered unselected, and clicking one added a duplicate instead of toggling. Fixed by collapsing to one row per project (lowest scroll group) and de-duplicating on the way back.

**Trade-off, called out for review:** a project open in both scroll group A and scroll group B now shows only "(A)" in this picker. The trigger under-reports where the project is open. Recorded as `adr-per-project-selection-collapses-scroll-groups` in `.context/standards/Architecture-Decisions.md`.

## New guard

A coverage test fails the build when a picker offers a grouping its own row builder cannot populate — the failure mode that produced the all-"Other" and single-bucket menus above. Each picker's offered grouping ids are an **exported constant** that both the component and the test read, so the advertised set and the checked set cannot drift.

## Deferred

- **A general project-type data source**, so Find and the checks side panel could offer Type. Both are excluded today purely for lack of data.
- **Splitting `%manageBooks_copy_sourcePlaceholder%`**, so the Copy "From" picker's aria-label can differ from its placeholder text.
- **A residual structural risk.** The settings sidebar is safe only because it omits `availableGroupings` entirely. The risk is narrow and specific: a grouping backed by real `openTabs` is always fine, but a **data-reading** grouping (`language`, `type`, `lastUsed`) added there would be dead on arrival, because `platform-bible-react` is PAPI-free and the sidebar's public `ProjectInfo` prop carries only an id and a name — there is nothing in the rows for those groupings to read. That defect would live *inside the shared library*, where this extension-scoped coverage test cannot see it. The durable fix is a check inside `ProjectSelector` itself.

## Verification

- `npm run typecheck` — `typecheck:core`, `typecheck:e2e`, `typecheck:workspaces` all clean. `typecheck:erb` fails on one pre-existing `TS2307` (cannot find module `spdx-license-list`), unrelated to this branch. The same missing module also trips the separate ESLint rule `import/no-unresolved`, which fails `lint:scripts` — two different tools reporting one missing dependency, not one error counted twice.
- `npm --workspace platform-scripture run test -- --run src/` — 560 passed / 30 files.
- `npm --workspace lib/platform-bible-react run test -- --run src/` — 1391 passed / 158 files.
- `npm --workspace lib/platform-bible-utils run test -- --run src/` — 622 passed, 1 skipped / 33 files.
- `npm run lint` on `extensions`, `lib/platform-bible-react`, `lib/platform-bible-utils` — zero errors, run **per workspace**.
- **Operational note for reviewers:** the root `npm run lint` is an `&&` chain, and its first link fails on the `spdx-license-list` `import/no-unresolved` error above. Everything after that link never runs, so the root command reports a failure that looks pre-existing while silently skipping the workspace lints — it hid 9 real lint errors during this work. Run each workspace's lint individually (`npm run lint --workspace extensions`, `… --workspace lib/platform-bible-react`, `… --workspace lib/platform-bible-utils`) rather than trusting the root command.
- **C# is untouched by this work**, so `dotnet test c-sharp-tests/` was not run.
- **Not verified: the manual Storybook sweep.** The automated story tests pass, but the grouping menus were not stepped through by hand in a browser, so a purely visual regression that no unit test covers could still be present. Worth a few minutes from whoever reviews the UI.

`lib/platform-bible-react/dist` and `lib/platform-bible-utils/dist` are regenerated in this branch; extensions compile against the committed `dist`, so the rename above only takes effect with it.

AI-assisted.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2673)


jolierabideau (commented):
## Review summary

Thanks for this — the API direction is right, and the deprecation hygiene is good. Replacing a mutually-exclusive boolean pair with a single `defaultGrouping` value is the change this component wanted, `defaultGroupByOpenTabs` keeps a truthful `@deprecated` tag with a `{@link}` to its replacement and exact `false` → flat-list semantics, and the three new `ProjectSelectorProject` fields are optional and propagated uniformly across all four `computeRows` sites. The rationale comments (why the `useState` initializer is deliberately mount-only, why the fallback is *not* `availableGroupings[0]`, why the radio items omit `onSelect={preventDefault}` while the checkbox below keeps it) are the genuinely forward-facing kind a maintainer would otherwise have to re-derive. The committed `dist/` output is consistent and complete, and `index.d.ts` is correctly unchanged since `ProjectSelector` ships only from the `experimental` entry point.

Two themes drive most of what follows: the removal of `groupByVersification` dropped a capability the new API can't express, and the feature ships enabled everywhere while no in-repo code produces the data two of the five groupings need.

---

### Critical

**`groupByVersification`'s locked-grouping behavior cannot be reproduced with the new API** — `project-selector.component.tsx:709-719`

The old prop did two things: force versification grouping *and* hide the filter menu entirely (`{!props.groupByVersification && <FilterMenu …>}`), so the user could not leave that mode. `defaultGrouping="versification"` only sets the *initial* value — the user can switch away, at which point `priorityVersificationId` stops applying.

The natural way to lock it, `availableGroupings={[]} defaultGrouping="versification"`, silently resolves to `'none'`, because the initializer gates on `availableGroupings.includes(props.defaultGrouping)` and then falls through to `availableGroupings.includes('openTabs') ? 'openTabs' : 'none'`. And `availableGroupings={['versification']}` still renders a `None` radio the user can pick. So there is no prop combination that expresses "grouped by versification, no menu".

Two non-breaking fixes, either or both:

1. Retain `groupByVersification?: boolean` as `@deprecated`, mapping internally to a locked versification grouping — symmetrical with the treatment `defaultGroupByOpenTabs` got, whose asymmetry is otherwise unexplained.
2. Honor `defaultGrouping` even when absent from `availableGroupings` — treat `availableGroupings` as *menu content* rather than a validity whitelist, so `availableGroupings={[]}` + `defaultGrouping="versification"` gives forced grouping with no menu.

Worth deciding first whether the "Based on" picker actually needs the lock. If being able to switch away is an intentional improvement, that's a fine answer — it just needs saying explicitly, and then this becomes a documentation item.

---

### Important

**Two of the five groupings have no data producer anywhere in the repo, yet are enabled in every picker.** A repo-wide grep (excluding `dist/`) finds zero producers of `lastUsedAt` and no relevant producers of `type`/`typeName` (the only `typeName` hits are `platform-scripture-editor` annotation styles). So in every in-repo picker — checks side panel, settings sidebar, manage-books sidebar, Copy "From", Create "Based on" — choosing **Type** files every row under one `"Unknown type"` heading and **Last used** under one `"Other"` heading. This is the same failure `partitionAndSort` already guards against for open-tabs grouping, with the comment "A lone 'Other projects' heading without a partner section reads as a bug". Fix by passing explicit `availableGroupings` per call site, and/or applying that same guard in the three new partition functions — when the only section produced is the unknown/other bucket, emit a `flat` section instead.

**The migrated caller lost its restriction and gained a menu it never had** — `manage-books-dialog.component.tsx:2489-2510`, `:2202-2214`. The filter menu used to be suppressed for the versification picker; the new gate is `availableGroupings.length > 0 || mode === 'project-multi'`, and neither call site passes `availableGroupings`. `allOtherProjectsAsPS` (line 743) supplies only `id`/`shortName`/`fullName`/`versificationId`/`versificationName`; `copyFromProjectsAsPS` (line 775) only `id`/`shortName`/`fullName`. Suggest `availableGroupings={['versification']}` on Create "Based on" and `availableGroupings={['openTabs']}` on Copy "From" — matching the prop's own documented purpose ("Pass a subset to hide the ones your data doesn't support").

**Nine new user-facing strings are unlocalized in the only consumer that localizes the selector** — `manage-books.web-view.tsx:435-476` hand-builds `ProjectSelectorLocalizedStrings` key-by-key and wasn't extended for `filterGroupNone`, `filterGroupByLanguage`, `filterGroupByLastUsed`, `filterGroupByVersification`, `filterGroupByType`, `languageUnknownSectionHeading`, `typeUnknownSectionHeading`, `lastUsedRecentSectionHeading`, `lastUsedOtherSectionHeading`. `contributions/localizedStrings.json:219-232` has no matching `%manageBooks_projectSelector_*%` keys either, so all nine fall back to hardcoded English in every locale. They also need adding to the key list in `manage-books-dialog.types.ts` (~line 240).

**The two changed default strings never reach users** — `contributions/localizedStrings.json:222,225` still hold `"By open tabs"` and `"Group"`, and the extension's values win over the library defaults. So in the shipping manage-books dialog the menu reads "Group / None / By open tabs / Language / Last used / …" — one item phrased as a sentence fragment among nouns, under a heading that no longer matches the radio semantics. The PR's claimed polish ("Group by", dropped `"By "` prefix) is invisible in the product. This is a rewording of the same concept, so updating the existing values is the right move (plus the fallbacks at `manage-books.web-view.tsx:446,448-451`).

**`ProjectSelectorGroupingOption` isn't exported from the package entry point** — `experimental.ts:15-20` re-exports only `ProjectSelectorProject`, `ProjectSelectorOpenTab`, `ProjectSelectorProjectPair`, `ProjectSelectorLocalizedStrings`, and `dist/experimental.d.ts:212` declares the type with no `export`. Extension developers can pass an inline literal to `availableGroupings` but can't name the type to declare a variable, prop, or helper — which is exactly what the new story does, and it only works because it imports the internal `@/components/...` path extensions can't use.

**No test changes**, against ~189 new lines in `project-selector.rows.ts` and a new 4-branch resolution chain. `project-selector.rows.test.ts` (38 `it`s) and `project-selector.component.test.tsx` (11 `it`s) both already exist and were untouched. Uncovered: `partitionByLanguage` bucketing/ordering/elision; `partitionByType` including the "adopt the first non-empty `typeName` observed" branch at `rows.ts:583` (the only non-obvious line in the file); `partitionByLastUsed` ordering and the `typeof … === 'number'` split that deliberately keeps `lastUsedAt: 0` in the first bucket; the whole `defaultGrouping` → `includes` → `defaultGroupByOpenTabs === false` → `'openTabs'`/`'none'` chain; `availableGroupings={[]}`; and `priorityVersificationId` being ignored under other groupings. At minimum please pin the resolution chain — that's exactly where the Critical finding lives. (For what it's worth, `partitionByVersification` was already untested and no `groupByVersification` tests existed before, so nothing was deleted or left dangling — the gap is purely additive.)

**`partitionByLanguage` and `partitionByType` are near-verbatim copies of `partitionByVersification`** — `project-selector.rows.ts:466-670`. All three bucket into a `Map` by a key, derive a label, sort buckets with `compareRows`, sort sections by `label.localeCompare(…, { sensitivity: 'base' })` (three identical copies of that line, at 501/546/591), then append a trailing unknown bucket. The only differences are the key/label accessors, the `kind` string, and versification's priority pin. One generic `partitionByKey(rows, { kind, keyOf, labelOf, unknownLabel, priorityKey? })` would collapse ~120 lines to ~40 and make a sixth grouping a one-liner. Note the copy has already drifted: versification uses `if (!existing.label && …)` (unreachable, since `label = name ?? id` is never empty) while type uses `if (existing.label === key && row.typeName)`, which makes a bucket's label depend on row order when siblings disagree.

**The grouping option set isn't exhaustiveness-checked; a sixth option needs edits in nine places** — `project-selector.component.tsx:589-608,791-808` and `rows.ts:392`. Both switches end in a `default` that swallows a new member with no compile error: `groupingLabel` falls through to `return option`, silently rendering the raw camelCase key as a UI label, and the `sections` switch merges `case 'none': default:` so a new grouping silently renders flat. `isGroupingChoice` is a hand-written six-way string chain independent of the union, so it needs editing too. The nine sites: the union, `DEFAULT_GROUPINGS`, `isGroupingChoice`, `groupingLabel`, `ProjectSelectorLocalizedStrings`, `DEFAULT_STRINGS`, the `sections` switch, `RowSection['kind']`, `sectionHeading`. Consider deriving the union from `DEFAULT_GROUPINGS` (`as const` + `typeof DEFAULT_GROUPINGS[number]`), deriving the guard from that array, and replacing the `default` arms with an exhaustive-`never` assertion so the compiler enumerates the rest. Cheapest to do now, while there are only five.

**`partitionByLanguage` keys off the display string, departing from the pattern its two siblings use** — `project-selector.rows.ts:591-625`. `partitionByVersification` keys on `versificationId` and labels with `versificationName`; `partitionByType` keys on `type` and labels with `typeName`. `ProjectRow` already carries `languageCode` (`rows.ts:132`) and it's unused here, though it *is* used for display, search, and tooltips elsewhere in the component. Consequence: two rows both tagged `languageCode: 'en'` split into separate sections if one says `"English"` and another `"english"`/`"Eng"`, and section labels are whatever unlocalized string the caller passed. Key on `languageCode` with `language` as the label (falling back to the code) — or add a comment saying why display-name keying is deliberate.

**The `"Recently used"` heading misrepresents its contents** — `project-selector.rows.ts:672-694`. Every row carrying any `lastUsedAt` goes in it; there's no time window, so a project last opened three years ago sits under "Recently used" (the story fixture files a 20-day-old item there). As written the option is a *sort*, not a grouping — its only grouping effect is segregating rows with missing data. Either real recency buckets (Today / This week / This month / Older, which is the shape users expect from "Last used") or a heading that states the actual predicate. `"Other"` is also vague for "never used". This one's a product-copy decision more than a code nit.

**Ad-hoc z-index arithmetic bypasses the curated scale** — `project-selector.component.tsx:85-87`. `Z_INDEX_ABOVE_SELECTOR_POPOVER = Z_INDEX_ABOVE_DOCK + 50` lands an unnamed layer (650) between `Z_INDEX_ABOVE_DOCK` (600) and `Z_INDEX_FIRST_RUN` (700), where `components/z-index.ts` — an explicit single-source-of-truth ladder, mirrored by `src/renderer/styles/_vars.scss` — documents nothing. The underlying fix is legitimate: `popover.tsx` pins `PopoverContent` to 600 via a `// CUSTOM:` block, so the old `Z_INDEX_OVERLAY` (400) genuinely rendered the nested dropdown and row tooltip behind the popover. But the documented approach for a nested overlay is the other direction — `dismissal-patterns.mdx` says to portal nested popovers into the surface's own content node with `PopoverPortalContainerProvider` "rather than letting them default to `document.body`", which fixes stacking *and* keeps dismissal containment correct. If the constant stays, please name it in `z-index.ts` with the usual doc comment and import it.

**The four copies of the project→row field-mirroring block grew again** — `project-selector.rows.ts:267-370`. This diff adds the same three lines (`type`, `typeName`, `lastUsedAt`) at four separate `computeRows` sites, on top of the `versificationId`/`versificationName` pair already duplicated four times. The next optional field needs four more edits, and the failure mode if one is missed is silent data loss — a grouping that appears to do nothing. A `mirrorProjectFields(project)` helper to spread would close it.

**The free-form-`type` decision isn't recorded in the architecture-decision log.** The ~30 lines of TSDoc on `ProjectSelectorProject.type` (and the PR body) argue for deliberately not constraining `type` to a union, weighing PT9 `ProjectType` and DBL `ResourceType` against duplication and drift. That's the "choosing among viable approaches" case `CLAUDE.md` → "Recording Architecture Decisions" asks to be captured in `.context/standards/Architecture-Decisions.md`, in byte-order slug position. Moving the rationale there also resolves the secondary concern that a public API doc is carrying design-doc prose whose concrete value lists will drift from their sources of truth — the contract (free-form key, grouped by exact equality, caller owns the taxonomy) is what the TSDoc needs.

---

### Minor

- `component.tsx:646-660` — the `DropdownMenuRadioGroup` has no accessible name. Radix renders a bare `role="group"` and the adjacent `DropdownMenuLabel` ("Group by") isn't programmatically associated, so a screen reader announces "Open tabs, radio button, 1 of 6" with no indication of what's being chosen. The sibling `BookGridGroupByToggle` sets `aria-label={labels.groupBy}` on its `ToggleGroup` — worth mirroring.
- `component.tsx` — nothing announces the regroup. No `aria-live` region, and since the menu now closes on selection Radix returns focus to the filter icon button rather than the `CommandInput`, so a keyboard/SR user lands on a button, hears nothing about the restructured list, and must Shift-Tab back before arrow keys navigate it.
- `component.tsx:568-577` — `isGroupingChoice` re-lists the union by hand, so adding a member yields a guard that rejects the new value with no compile error.
- `component.tsx:265-272` vs `709-719` — the JSDoc blocks disagree on precedence. `defaultGroupByOpenTabs` says "Superseded silently if both are set", but the initializer only returns early when `defaultGrouping` is *valid*; if `defaultGrouping` names an unavailable option, control falls through and the deprecated prop does take effect. Fix the wording, or short-circuit on `props.defaultGrouping !== undefined`.
- `component.tsx:705-719` — stale state if `availableGroupings` shrinks after mount: `activeGrouping` is mount-initialized only (deliberately, per the comment), so a caller narrowing the array leaves the component grouping by an option no longer in the menu, with the radio group showing nothing selected. Either reconcile to `'none'`, or note that callers must keep the array stable — the existing comment covers the `defaultGrouping` case but not this one.
- `component.tsx:574,262` — `GroupingChoice` is the de facto public prop type but is unexported, and `defaultGrouping` re-spells the union inline rather than using the alias defined ~300 lines later in the same file. `book-grid.component.tsx:62` folds `'none'` into one exported union (`BookGridGroupBy = 'canon' | 'status' | 'none'`) — same shape, worth matching.
- `component.tsx:97-107,160-166` — key naming: five keys follow `filterGroupBy<Option>` but the None item is `filterGroupNone`; `filterGroupByNone` would make the key set mechanically derivable from the option union. Also `lastUsedOtherSectionHeading: 'Other'` sits beside `otherProjectsSectionHeading: 'Your projects & resources'` — two unrelated "other" buckets whose names don't distinguish them.
- `rows.ts:451-494` — `type`/`typeName` breaks the `<thing>Id`/`<thing>Name` pairing that `versificationId`/`versificationName` establishes two fields above; `typeId`/`typeName` would keep the convention and read less ambiguously than a bare `type` on a data object.
- `component.tsx:793,807` — `partitionAndSort(filteredRows, true)` / `(…, false)` reads as a mystery flag now that the caller dispatches on a named grouping. The two branches are effectively two functions; splitting them (`partitionByOpenTabs` / `flatSection`) would make the switch arms self-describing.
- `rows.ts:386-392` — `RowSection`'s TSDoc still says `'flat' means no section header (grouping toggle off)`; there's no longer a toggle. Separately, `platform-bible-utils` already exports `groupBy(items, keySelector)` (`util.ts:146`) and a `Collator` wrapper (`intl/intl-collator.ts`) while the new partition functions hand-roll the bucket loop and use raw `localeCompare` — staying consistent with the surrounding file is defensible, but if the generic partitioner above happens, building it on `groupBy` is the reuse-conformant version.
- Section headings go into `CommandGroup heading`, whose shadcn styling (`command.tsx:185`) has no truncation, while rows in the same list get the truncation-tooltip treatment via `useTruncationTooltip`. In the 320px popover a long localized language/type name wraps to multiple lines with no tooltip; `responsiveness.mdx` asks for a tooltip wherever text truncates and warns menu-like entries should truncate rather than wrap.
- The filter trigger's accessible name is still `filterAriaLabel` ("Filter") although the menu is now predominantly a "Group by" radio group with one filter toggle.
- `stories.tsx` — the new story descriptions use pre-PR label text (`'By open tabs'`, `'By last used'` in `RestrictedGroupingOptions`) and call the trigger "the funnel icon" when it renders lucide `Filter`.
- `stories.tsx:393-570` — no story for `defaultGrouping="versification"` + `priorityVersificationId`, and no fixture in the file carries `versificationId`/`versificationName` at all. That's the grouping with the only non-alphabetical section order (the pinned bucket) and the one the extension actually uses, so it's the most valuable to document visually. Pre-existing gap, but this PR is well-placed to close it.

---

### Also good

- Only one in-repo caller used `groupByVersification` and it was migrated in the same commit; a repo-wide grep (excluding `dist/`) returns zero hits, and no existing test asserted the old checkbox behavior.
- `isGroupingChoice` narrows the raw `string` from `onValueChange` instead of asserting, respecting the repo's `no-type-assertion` rule.
- `partitionByLastUsed` uses `typeof row.lastUsedAt === 'number'` rather than a truthiness check, correctly keeping a `0` timestamp in the first bucket.
- Empty-bucket elision is consistent across all three new partition functions.
- Closing the menu on grouping selection while keeping it open for "Show selected only" mirrors `component-choices.mdx` (single-select closes, multi-select doesn't).
- Every new string is sentence case per `capitalization.mdx`, and none carries an ellipsis — correct per `ellipses.mdx`, since selecting a grouping acts immediately.
- Reusing `DropdownMenuRadioGroup`/`DropdownMenuRadioItem` inherits existing RTL awareness (`readDirection()`, logical `ps-`/`pe-`/`end-` utilities), so no new directionality debt.
- Nothing under `components/shadcn-ui/` was touched, so the `// CUSTOM:` rule doesn't apply — the z-index fix was correctly confined to the consuming component.
- The three new stories follow the file's conventions, include a fixture deliberately missing `type`/`lastUsedAt` to exercise the unknown buckets, and pin `NOW` to a fixed epoch for determinism.
- No build-config or dependency movement, and CI is green on all three platforms.

---

Happy to talk through any of these, especially the Critical one and whether `DEFAULT_GROUPINGS` should really be all five given the data situation.

*AI-assisted review (Claude Opus 5); findings verified against the diff and `main` by hand.*

Sebastian-ubs:
I am currently reworking this

Sebastian-ubs:
@jolierabideau can you take it from here? Currently it looks like this - (I intentionally removed "Select all" by the way)
grafik

extensions/src/platform-scripture/src/manage-books-dialog/manage-books-dialog.component.tsx
```diff
@@ -2260,19 +2307,18 @@ export function ManageBooksDialog({
...
-                          )}
-                          localizedStrings={projectSelectorLocalizedStrings}
-                          // Mirror the prior <SelectTrigger> "primary fill while empty" affordance —
-                          // the picker reads as a call-to-action until a source project is set.
-                          buttonClassName={cn(
-                            'tw:h-8 tw:w-full',
-                            !copySourceId &&
-                              'tw:border-primary tw:bg-primary tw:text-primary-foreground tw:hover:bg-primary/90',
-                          )}
+                          localizedStrings={{
```
Sebastian-ubs, line 2349: Not sure if we should really pull these into localized strings @jolierabideau , please check

extensions/src/platform-scripture/src/checklist.web-view.tsx
```diff
@@ -655,20 +699,85 @@ global.webViewComponent = function ChecklistWebView({
...
+          // eslint-disable-next-line no-type-assertion/no-type-assertion
+          projectSelectorResolvedStrings as ProjectSelectorResolvedStrings,
+        ),
+      ),
+    ],
+    [projectSelectorResolvedStrings],
+  );
+
+  // Selection grouping is auto-added by ProjectSelector for `project-multi` mode. We pass
+  // `openTabs={[]}` here on purpose: comparative-texts selection is stored per-project (no
```
Sebastian-ubs: Not sure this is right. I guess we should still be able to use open tabs as a grouping and refrain from duplicate selection

lib/platform-bible-react/src/components/advanced/project-selector/project-selector.groupings.ts
```diff
@@ -0,0 +1,122 @@
...
+  languageLabel?: string;
+  typeLabel?: string;
+  lastUsedRecentSectionHeading?: string;
+  lastUsedOtherSectionHeading?: string;
+  languageUnknownSectionHeading?: string;
+  typeUnknownSectionHeading?: string;
+};
+
+// The two bucket keys the built-in `lastUsed` grouping partitions into.
+const RECENT = 'recent';
```
Sebastian-ubs, line 87: Do those need localisation?

lib/platform-bible-react/src/components/advanced/project-selector/project-selector.component.tsx
```diff
@@ -77,57 +85,96 @@ export type {
...
+  ariaLabel?: string;
+  /** Trigger fallback text when nothing is selected. */
+  buttonPlaceholder?: string;
+  /** "No results" message inside the popover when the search has no matches. */
+  commandEmptyMessage?: string;
+  /** Placeholder for the popover's search input. */
   searchPlaceholder?: string;
-  /** Accessible label for the filter menu icon button. Defaults to `"Filter"`. */
+  /** Accessible label + `title` for the filter menu icon button. */
   filterAriaLabel?: string;
```
Sebastian-ubs: Probably this should be renamed to groupBy, maybe even using a different icon

lib/platform-bible-react/src/components/advanced/project-selector/project-selector.component.tsx
```diff
@@ -77,57 +85,96 @@ export type {
...
   groupSectionLabel?: string;
-  /** Filter menu: section heading for the filter toggles. Defaults to `"Filter"`. */
-  filterSectionLabel?: string;
-  /** Filter menu: "By open tabs" item under the Group section. Defaults to `"By open tabs"`. */
-  filterGroupByOpenTabs?: string;
-  /** Filter menu: multi-only item under the Filter section. Defaults to `"Show selected only"`. */
-  filterShowSelectedOnly?: string;
-  /** Section heading for the Open tabs section. Defaults to `"Opened project & resource tabs"`. */
+  /** Filter menu: "None" grouping radio item — the "no grouping" option. */
+  filterGroupNone?: string;
```
Sebastian-ubs: Probably remove the word "filter"

lib/platform-bible-react/src/components/advanced/project-selector/project-selector.component.tsx
```diff
@@ -77,57 +85,96 @@ export type {
...
+   * `"A"`).
    */
   boundButClosedTooltip?: string;
-  /** Label of the "Open" button shown on bound-but-closed rows. Defaults to `"Open"`. */
+  /** Label of the "Open" button shown on bound-but-closed rows. */
   openButtonLabel?: string;
-  /** Multi-select: "Select all" button. Defaults to `"Select all"`. */
-  selectAll?: string;
-  /** Multi-select: "Clear all" button. Defaults to `"Clear all"`. */
+  /** Multi-select: "Clear all" button (shown only when at least one pair is selected). */
```
Sebastian-ubs, line 163: Test this: the ui should not look blank nor jump

lib/platform-bible-react/src/components/advanced/project-selector/project-selector.component.tsx
```diff
@@ -137,6 +184,115 @@ function resolveStrings(
...
+  '%projectSelector_filterAriaLabel%',
+  '%projectSelector_groupSectionLabel%',
+  '%projectSelector_filterGroupNone%',
+  '%projectSelector_openTabsSectionHeading%',
+  '%projectSelector_otherProjectsSectionHeading%',
+  '%projectSelector_boundButClosedTooltip%',
+  '%projectSelector_openButtonLabel%',
+  '%projectSelector_clearAll%',
+  '%projectSelector_grouping_openTabs_label%',
+  '%projectSelector_grouping_lastUsed_label%',
```
Sebastian-ubs: If we pass those default grouping ones in here - how customizable is it really - and should we not use the same customizing mechanism instead?

lib/platform-bible-react/src/components/advanced/project-selector/project-selector.component.tsx
```diff
@@ -137,6 +184,115 @@ function resolveStrings(
...
+    clearAll: strings['%projectSelector_clearAll%'],
+  };
+}
+
+/**
+ * Convert the raw `%projectSelector_*%` resolved strings into the labels + section-heading strings
+ * that {@link makeBuiltInGroupings} accepts. Pair with {@link buildProjectSelectorLocalizedStrings}
+ * to wire the whole picker from a single {@link PROJECT_SELECTOR_STRING_KEYS} call.
+ */
+export function buildBuiltInGroupingStrings(
```
Sebastian-ubs: I would expect this to live not on here, but in selector-groupings.ts

lib/platform-bible-react/src/components/advanced/project-selector/project-selector.component.tsx
```diff
@@ -912,18 +1077,23 @@ export function ProjectSelector(props: ProjectSelectorProps) {
     <PopoverTrigger asChild>{triggerButton}</PopoverTrigger>
   );
 
+  // The filter menu only exists to let the user switch between groupings. With 0 or 1 groupings
+  // there is nothing to switch between (single-grouping lock), so drop the funnel button entirely.
+  const showFilterMenu = availableGroupings.length > 1;
```
Sebastian-ubs: Rename to grouping / groupby menu

lib/platform-bible-react/src/components/advanced/project-selector/project-selector.component.tsx
```diff
@@ -936,35 +1106,33 @@ export function ProjectSelector(props: ProjectSelectorProps) {
...
                 />
               )}
             </div>
-            {props.mode === 'project-multi' && (
-              <div className="tw:flex tw:justify-between tw:border-b tw:py-2 tw:pe-4 tw:ps-2">
-                <Button variant="ghost" size="sm" onClick={handleSelectAll}>
-                  {`${strings.selectAll} (${allPairs.length.toString()})`}
-                </Button>
+            {props.mode === 'project-multi' && props.selection.pairs.length > 0 && (
+              // Right-aligned "Clear all" only. Select all was removed — it is redundant with the
```
Sebastian-ubs: It's not redundant, it is risky for performance to select all projects

extensions/src/platform-scripture/src/checklist.web-view.tsx
jolierabideau: Agreed, and this is fixed — open tabs is back as a grouping on the comparative-texts picker rather than disabled.

The underlying bug was the duplicate selection you're pointing at: `project-multi` keys rows by `(projectId, scrollGroupId)`, but comparative-texts storage is per-project with no scroll group, so a stored ref matched neither row of a project open in two scroll groups — both rendered unselected, and clicking one added a duplicate instead of toggling.

Fix is to collapse to one row per project (lowest scroll group) and de-duplicate on the way back: `comparativeOpenTabByProject` / `collapsedComparativeOpenTabs` in `extensions/src/platform-scripture/src/checklist.web-view.tsx:750-762`. `openTabs` is now in the offered set (`CHECKLIST_PROJECT_SELECTOR_GROUPING_IDS`, same file `:166`).

Trade-off worth knowing: a project open in both scroll group A and B now shows only "(A)" in this picker, so the trigger under-reports where it's open. Recorded as `adr-per-project-selection-collapses-scroll-groups` in `.context/standards/Architecture-Decisions.md`.

lib/platform-bible-react/src/components/advanced/project-selector/project-selector.groupings.ts
jolierabideau: Yes, and they are now. `BuiltInGroupingStrings` is the *input* to the built-ins — every field is filled from the central `%projectSelector_grouping_*%` localization block via `buildBuiltInGroupingStrings` (`project-selector.groupings.ts:198`, keys listed at `:13`). The English literals in the file are only last-resort fallbacks for an unlocalized mount (e.g. a bare Storybook render), not production copy.

lib/platform-bible-react/src/components/advanced/project-selector/project-selector.component.tsx
jolierabideau: Done, both parts. The strings are `groupByAriaLabel` / `groupByNone` (`project-selector.component.tsx:129,133`, defaults `:172,174`, keys `%projectSelector_groupByAriaLabel%` / `%projectSelector_groupByNone%` at `:206,208`), and the icon changed from the funnel to lucide's `Group` (`:18`).

Pinned by a test: `group-by menu naming` in `project-selector.component.test.tsx:365` asserts the menu trigger is named "Group by" and that no button matching `/filter/i` exists.

lib/platform-bible-react/src/components/advanced/project-selector/project-selector.component.tsx
jolierabideau: Done — "filter" is gone from the whole surface: props, localization keys, component names and comments. `filterAriaLabel` → `groupByAriaLabel`, `filterGroupNone` → `groupByNone`, `filterSectionLabel` removed, `FilterMenu` → `GroupByMenu`. The `not.toBeInTheDocument()` assertion on `/filter/i` in `project-selector.component.test.tsx:365` keeps it from creeping back.

lib/platform-bible-react/src/components/advanced/project-selector/project-selector.component.tsx
jolierabideau: Tested, both halves.

**Not blank** — `first paint` in `project-selector.component.test.tsx:426` renders the selector with *no* `localizedStrings` at all and asserts the trigger has a non-empty accessible name and non-empty text. That's backed by the deliberate fallbacks at `project-selector.component.tsx:163-166`: `ariaLabel` and `buttonPlaceholder` have English last-resort defaults precisely so the trigger is never empty before localized strings resolve.

**No jump** — `project-multi trigger shape at 1 vs 2+ selected` (`:388`) asserts the count renders as the same Badge element at one selected as at two, which is the shape that would otherwise change between renders.

lib/platform-bible-react/src/components/advanced/project-selector/project-selector.component.tsx
jolierabideau: Fair point, and the rework does exactly that — the built-ins now go through the *same* mechanism as a custom grouping, with no special path in the component.

`makeBuiltInGroupings(strings)` (`project-selector.groupings.ts:103`) returns plain `ProjectSelectorGrouping` descriptors, structurally identical to anything a consumer writes by hand, and they arrive through the ordinary `availableGroupings` prop. The component doesn't know which are "built-in". A consumer composes them freely — e.g. the comparative-texts picker does `[...makeBuiltInGroupings(...), makeSelectionGrouping(...)]`, and each surface filters the list down to the ids its own row builder can populate.

What stays central is only the *vocabulary*: the `%projectSelector_grouping_*%` keys, so every picker in the app says "Language" the same way. A consumer that wants different wording passes its own `label` on the descriptor.

lib/platform-bible-react/src/components/advanced/project-selector/project-selector.component.tsx
jolierabideau: Agreed — moved. `PROJECT_SELECTOR_STRING_KEYS`, `ProjectSelectorLocalizedStringKey`, `readProjectSelectorString`, `buildBuiltInGroupingStrings` and `buildSelectionGroupingStrings` all live in `project-selector.groupings.ts` now (`:13`, `:38`, `:56`, `:198`, `:231`), re-exported from the component for the public entry point.

`buildProjectSelectorLocalizedStrings` stayed in `project-selector.component.tsx:197` on purpose — it builds the component's own `ProjectSelectorLocalizedStrings` bag, not grouping data, so it belongs next to the type it populates.

lib/platform-bible-react/src/components/advanced/project-selector/project-selector.component.tsx
jolierabideau: Renamed — `FilterMenu` → `GroupByMenu` (`project-selector.component.tsx:585`, props type `:578`) and `showFilterMenu` → `showGroupByMenu` (`:1038`).

lib/platform-bible-react/src/components/advanced/project-selector/project-selector.component.tsx
jolierabideau: You're right, and the comment was wrong — reworded to say what actually motivated the removal. `project-selector.component.tsx:1075-1077` now reads:

> Right-aligned "Clear all" only. There is deliberately no "Select all": selecting every project mounts a data subscription per project, so on a large installation it is a performance hazard rather than a convenience.

Called out in the PR description too, so it doesn't come back as a bulk toggle.

extensions/src/platform-scripture/src/manage-books-dialog/manage-books-dialog.component.tsx
jolierabideau: Checked. They aren't being *pulled into* localized strings — they're still resolved from the consumer's own key, `t('%manageBooks_copy_sourcePlaceholder%', 'Select project')`. What changed is only the delivery shape: the standalone `ariaLabel` / `buttonPlaceholder` props were removed from `ProjectSelectorProps`, so per-picker copy now merges on top of the shared bag:

```tsx
localizedStrings={{
  ...projectSelectorLocalizedStrings,
  ariaLabel: t('%manageBooks_copy_sourcePlaceholder%', 'Select project'),
  buttonPlaceholder: t('%manageBooks_copy_sourcePlaceholder%', 'Select project'),
}}
```

That keeps one channel for every string the selector renders instead of two, and keeps picker-role copy on the consumer's key rather than in the shared `%projectSelector_*%` block.

The one genuine wart is that a single key feeds both the aria-label and the visible placeholder. Splitting `%manageBooks_copy_sourcePlaceholder%` into two keys is listed as deferred in the PR description — happy to do it here if you'd rather not defer it. Leaving this thread open for your call.

katherinejensen00 (approved):
Thanks for all your hard work on this! Claude was impressed and only found low level warnings and suggestions, nothing worth holding the PR up for. Approved 🎉

> @katherinejensen00 partially reviewed 60 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2673#-:-P1_bftR31x7XPBoGDNU:b-m0nvt3)* status: all files reviewed, 8 unresolved discussions (waiting on Sebastian-ubs).

katherinejensen00 (approved):
> @katherinejensen00 resolved 8 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2673)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2674, opened 2026-08-13
fix(settings): show project full name in settings sidebar ProjectSelector

Sebastian-ubs:
## Summary

The settings sidebar's `<ProjectSelector>` was rendering each project as a single-line row (`shortName` only). Every other `<ProjectSelector>` consumer in the app (manage-books, checks side panel) shows a two-line row: short name over a muted full name. This PR aligns the settings sidebar with that pattern.

### Root cause

`SettingsSidebar` accepts a public `ProjectInfo` shape with a single `projectName` string. Its internal adapter mirrored that string into both `shortName` and `fullName` on the `ProjectSelectorProject`, which trips [ProjectSelector's `fullName === shortName` de-dup rule](https://github.com/paranext/paranext-core/blob/main/lib/platform-bible-react/src/components/advanced/project-selector/project-selector.component.tsx) (present to avoid rendering the same string twice) — so the second line was correctly suppressed. That behaviour is right for callers that genuinely only have one name; the settings sidebar isn't one of them.

### Fix, at the source

- **Extend `SettingsSidebar.ProjectInfo`** with an optional `projectFullName`. The adapter now uses `projectFullName ?? projectName` for `fullName`. When callers upgrade to supply it, rows render two lines. When they don't, the de-dup rule still collapses to a single line — no visual regression.
- **[settings-tab.component.tsx](src/renderer/components/settings-tabs/settings-tab.component.tsx)** — the only in-repo consumer of `SettingsSidebar` — replaces its single-setting `getProjectName` helper with `getProjectNames`, which fetches `platform.name` **and** `platform.fullName` in parallel from the `platform.base` PDP. This mirrors the pattern in [`extensions/src/platform-scripture/src/checks-side-panel.web-view.tsx:43-48`](https://github.com/paranext/paranext-core/blob/main/extensions/src/platform-scripture/src/checks-side-panel.web-view.tsx#L43-L48). Empty `platform.fullName` is collapsed to `undefined` so projects without one still render single-line.

## Test plan

- [ ] Open the settings tab in-app with a project that has distinct `platform.name` and `platform.fullName` (e.g. any manage-books-visible standard project). Confirm the sidebar's project picker rows now show short name over full name.
- [ ] Open the same tab with a legacy project that has no `platform.fullName` set (or an empty one). Confirm the row still renders as a single line (short name only).
- [ ] `npm run typecheck:core` — clean.
- [ ] `npm --workspace lib/platform-bible-react run lint` — clean.
- [ ] `npm --workspace lib/platform-bible-react run test -- --run src/stories/advanced/settings/settings-sidebar.stories.tsx` — 6 story smoke-tests pass.

## Scope

This PR only touches the settings-tab / SettingsSidebar path. Consumers of `SettingsSidebar` outside this repo (if any) don't need to change — `projectFullName` is optional and the fallback preserves prior behaviour.

AI-assisted.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2674)

before:
image

after:
image


jolierabideau (changes requested):
## Summary review

Thanks for this — it's a well-scoped fix and the pattern instincts are right where it counts. The trigger change reuses the existing canonical `triggerLabelFormat="shortNameAndFullName"` prop rather than hand-rolling a `"{short} - {full}"` label, the `ProjectInfo` extension is purely additive so downstream consumers don't have to move, and the single-line fallback leans on `ProjectSelector`'s existing `fullName === shortName` de-dup instead of reimplementing it. `getProjectNames` even matches the name and parallel-fetch shape of the three existing precedents, and the committed `dist/` is properly regenerated. Good instincts throughout.

I did run into two things I think are worth resolving before this goes in, plus some smaller pattern notes. Happy to be wrong on any of these — and glad to pair on the first one if it's easier.

### 1. The empty-string guard may not catch the case it's aimed at

I think this one's worth a close look:

```ts
const projectFullName =
  typeof projectFullNameRaw === 'string' && projectFullNameRaw.length > 0
    ? projectFullNameRaw : undefined;
```

Unless I'm misreading the flow, an *unset* `platform.fullName` doesn't come back as `''`. `ParatextProjectDataProvider.GetProjectSetting` falls through (`c-sharp/Projects/ParatextProjectDataProvider.cs:1783-1786`, *"Setting not found, so get the default value"*) to `ProjectSettingsService.GetDefault`, and the contributed default is the localize key `'%project_full_name_missing%'` (`src/extension-host/data/core-project-settings-info.data.ts:16-19`), which resolves to **`*Name Missing*`** (`assets/localization/en.json:368`; `*Falta el Nombre*` in es).

That's a non-empty string, so it would survive the guard and the sidebar would render `ESV - *Name Missing*` in the trigger, with a muted `*Name Missing*` second row line. It looks like only a project with a literally-empty `FullName` in `Settings.xml` takes the `''` path.

I suspect what happened is that two similar-looking paths got conflated — the empty-string shape is definitely real, it just comes from the C# `ProjectSummary.fullName` **wire field** that manage-books consumes (`manage-books.web-view.tsx:111`, *"Empty when unset"*) rather than from `pdp.getSetting`. Easy one to hit.

One option: compare against `papi.projectSettings.getDefault('platform.fullName')` and collapse a match to `undefined` — or simply drop the value when it equals `projectName`, since the selector de-dups that anyway. Whichever you prefer.

For what it's worth, this looks latent in `checks-side-panel.web-view.tsx` and `find.web-view.tsx` too, which pass `platform.fullName` straight through — so it's not something this PR introduced. This just happens to be the first place it lands in a trigger label where it becomes visible. Might be worth a follow-up issue for those two rather than widening this PR.

### 2. A single failing project could empty the whole sidebar list

`getProjectNames` adds a second `getSetting` inside `Promise.all`, and the caller's `Promise.all` over every project id doesn't have a per-project guard either. If any one project rejects, the outer promise rejects wholesale, `usePromise` catches it (`use-promise.hook.ts:75-85`) and leaves the default `[]` — so the picker would come up empty with nothing surfaced to the user.

This is a slightly new exposure: before this change only `platform.name` was fetched, which every PDP special-cases, whereas `platform.fullName` is handled by fewer engines and resolves through `getDefault`.

The nice thing is the precedent you cited in the PR body already solves this — `checks-side-panel.web-view.tsx:54` uses `pdp.getSetting('platform.language').catch(() => undefined)` with the rationale *"degrades to an unknown language bucket rather than failing the whole lookup."* You matched that helper's name and its parallel fetch; adding its `.catch(() => undefined)` (and possibly `Promise.allSettled` on the outer loop) would complete the pattern.

### Smaller pattern notes

- **`getProjectNames` return shape** — the three precedents (`checks-side-panel.web-view.tsx:46`, `find.web-view.tsx:132`, `checklist.web-view.tsx:656`) all return `{ shortName, fullName }`; this one returns `{ projectName, projectFullName }`. Aligning would make the four copies read as visibly the same helper. To be clear, writing a new local copy is absolutely the right call — the extension copies can't cross the build boundary into `src/renderer/`, and `checks-side-panel.web-view.tsx:42-44` documents that deliberately. Just the naming.
- **`ProjectInfo` vs `ProjectSelectorProject`** — with `projectFullName` added, `ProjectInfo` is now a field-for-field rename of the library's own `ProjectSelectorProject` (`projectId`/`projectName`/`projectFullName` vs `id`/`shortName`/`fullName`), with a `useMemo` whose only job is that rename. Every other consumer in the repo speaks `shortName`/`fullName` directly. Totally fine to keep if backward compatibility is the reason — in which case that'd be a useful thing to say in the comment on the `useMemo`, since it's what justifies the adapter's existence.
- **Comment placement** — the comment there explains the row layout, but the behavioral change in that hunk is the new `triggerLabelFormat` line, which is uncommented. Also, of the two selectors named, only manage-books uses `shortNameAndFullName`; the checks side panel uses the default `shortName` trigger and gets its two-line rows from the default row renderer.
- Tiny one: "behaviour" → "behavior" to match the rest of the repo.

### The new stories

I may be missing the intent here, so correct me if so — but `ShortNameTriggerLabel` and `ShortNameTriggerLabelNoScrollGroups` both exercise `triggerLabelFormat="shortName"`, which is the pre-existing default and isn't what this PR changes. The `shortNameAndFullName` path this PR adopts already has the `WideTriggerLabel` story plus three assertions in `project-selector.component.test.tsx`. The second story also overlaps `SimpleFlatList` fairly closely, differing mainly by `openTabs={[]}`.

Meanwhile the path this PR does introduce isn't covered anywhere: `settings-sidebar.stories.tsx` gave all five fixtures a distinct `projectFullName`, so nothing exercises the absent-`projectFullName` single-line fallback the new TSDoc describes. Leaving one fixture without it would pin that compatibility claim right where the component lives, which might be more valuable than the two new stories.

If you do keep them, they're built inline in both `render` bodies — the file's convention is module-scope fixtures (`sampleProjects`, `sampleOpenTabs`).

### Tests

`settings-sidebar.component.tsx` doesn't have a test file yet, and `getProjectNames` isn't covered — `settings-tab.component.test.tsx` mocks `getMetadataForAllProjects` to `async () => []` so that path isn't reached, though the `projectDataProviders.get` scaffolding is already in place. `project-selector.component.test.tsx` is a good working recipe for a small render-and-assert (trigger reads `"ESV - English Standard Version"`; falls back to one line without `projectFullName`). A three-case test around the normalization would also give issue #1 somewhere to live.

Small thing: the rename left a stale `getProjectName` reference in that test file's mock comment.

### Accessibility (worth considering)

The trigger passes `ariaLabel={projectsSidebarGroupLabel}` ("Project settings") through to the `role="combobox"` button. Since `aria-label` overrides text content in the accessible-name computation, a screen reader announces just *"Project settings, combobox"* — neither name comes through. It's pre-existing wiring and arguably out of scope, but this PR is what makes that label carry information worth announcing, so it might be a nice moment to fix it. Dropping `ariaLabel` (the `SidebarGroupLabel` already supplies group context) or switching to `aria-labelledby` would do it.

### Truncation (just worth eyeballing)

The sidebar is `16rem`, and after the icon, gaps, padding and chevron the label has roughly 150–170px to work with, so `"NIV - New International Version"` clips to around `"NIV - New Internat…"`. Not a guideline violation — `ProjectSelector` wraps the trigger in a tooltip with the full string, and short-name-first means the recognizable part survives at any width, which is the right ordering. Since the popover rows already show the full name on their own line as soon as you open it, it might be worth a quick look in-app to decide whether the second field earns its place in the trigger.

### Housekeeping

The branch is ~95 commits behind `main`, and `main` has since removed the `popoverContentStyle={{ zIndex: Z_INDEX_OVERLAY }}` override from `settings-sidebar.component.tsx` (per the z-index tier rule about consumers not overriding an overlay's z-index). That's the same hunk this PR touches, so there'll likely be a small conflict on rebase — keep `main`'s removal and your `triggerLabelFormat` line.

Also noticed the test plan checkboxes are still unchecked; the two manual in-app steps would be a good sanity check on #1 once it's sorted.

Thanks again — the core approach here is solid, and most of this is polish around the edges.

Sebastian-ubs:
closing as superseded by #2822

PR #2675, opened 2026-08-13
fix(project-selector): move row tooltip to right side; simplify content

Sebastian-ubs:
## Summary

Two changes to the row tooltip inside `<ProjectSelector>`'s popover:

### 1. Anchor moves from top to right

`side=\"top\"` frequently overlapped the row directly above (or the search bar for the topmost row). `side=\"right\"` puts the tooltip clear of the popover's content column and reads as a canonical row-details affordance. Alignment shifts from `text-center` to `text-start` to match side-anchored tooltip convention.

### 2. Content is restructured into three simple rows

Old content was a bold full name, then a couple of case-specific conditional lines. New content is a fixed three-row layout (with an optional fourth row for a disabled reason):

- **Full name** — bold, first row.
- **Language** — with `(languageCode)` muted in parentheses when both are present. Handles partial data (`language`-only or `languageCode`-only rows still render sensibly with no stray parenthesis).
- **Scroll group** — one line that covers every case the row can be in:
  - bound-but-closed rows: caller's `boundButClosedTooltip` template with the letter substituted, rendered italic.
  - open-tab rows (`project-multi` / `projectScrollGroup` modes with a `scrollGroupId`): `{scrRefLabel} ({letter})` when the caller supplied a scripture-ref label, else just the letter.
  - `project`-mode rows: comma-joined letters of all open scroll groups (`A, B`), mirroring the chip strip on the row.
- **Disabled reason** — retained as a muted-italic fourth row when present. Surfaces information not otherwise visible on the row.

## Test plan

- [ ] Storybook → **Advanced/Project Selector**: hover rows in every story (`SingleProject`, `MultiProject`, `ScrollGroupBinding`, `PerRowDisabled`, `NoProjects`). Tooltip anchors right of the row, doesn't overlap the row above.
- [ ] `SingleProject`: hover a row whose language and languageCode are both set → tooltip shows `English (en-US)`. Hover a row with a scroll group open → tooltip shows the letter list. Truncate the container so the row text clips — tooltip still opens on hover (unchanged truncation-open behaviour).
- [ ] `ScrollGroupBinding`: close a tab to force a bound-but-closed row → its tooltip's scroll-group line renders italic with the caller's `boundButClosedTooltip` string.
- [ ] `PerRowDisabled`: hover a disabled row → the muted italic `disabledReason` still surfaces on its own line under the scroll-group line.
- [ ] `npm --workspace lib/platform-bible-react run test -- --run src/components/advanced/project-selector/` — 53 tests pass.
- [ ] `npm --workspace lib/platform-bible-react run lint` — clean.

## Relation to other in-flight PRs

Touches the same file as [#2672](https://github.com/paranext/paranext-core/pull/2672) (z-index fix) and [#2673](https://github.com/paranext/paranext-core/pull/2673) (grouping options). Based on `main`, so this PR is independent and can merge in any order — expect a small mechanical rebase against whichever of those merges last.

AI-assisted.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2675)

before
image

after

Screenshot 2026-08-13 193528
Screenshot 2026-08-13 193517
Screenshot 2026-08-13 193511

### DIFF
--- a/lib/platform-bible-react/src/components/advanced/project-selector/project-selector.component.tsx
+++ b/lib/platform-bible-react/src/components/advanced/project-selector/project-selector.component.tsx
@@ -429,38 +429,63 @@ function ProjectRowView({ row, mode, strings, onClick, onOpen, selectedRowRef }:
   const letter =
     row.scrollGroupId !== undefined ? scrollGroupLetterFromMap(row.scrollGroupId) : undefined;
 
-  const tooltipBoundBut =
-    row.isBoundButClosed && letter
-      ? strings.boundButClosedTooltip.replace('{group}', letter)
-      : undefined;
+  // Single scroll-group tooltip line covering every case the row can be in. Bound-but-closed
+  // rows use the caller's `boundButClosedTooltip` template (with the letter substituted in) and
+  // render italic. Open-tab rows show `{scrRefLabel} ({letter})` when a ref label is available,
+  // otherwise just the letter. `project`-mode rows aggregate multiple open scroll groups into a
+  // comma-joined list to mirror the chip strip on the right side of the row.
+  let scrollGroupLine: string | undefined;
+  let scrollGroupLineIsItalic = false;
+  if (row.isBoundButClosed && letter) {
+    scrollGroupLine = strings.boundButClosedTooltip.replace('{group}', letter);
+    scrollGroupLineIsItalic = true;
+  } else if (letter) {
+    scrollGroupLine = row.scrollGroupScrRefLabel
+      ? `${row.scrollGroupScrRefLabel} (${letter})`
+      : letter;
+  } else if (row.openGroups.length > 0) {
+    scrollGroupLine = row.openGroups.map(scrollGroupLetterFromMap).join(', ');
+  }
+
+  // Compose the language line as a single string so it renders as one row (with the code, when
+  // present, muted in parentheses). Handles both partial-data cases (only language, only code)
+  // without emitting a stray parenthesis.
+  let languageLine: ReactNode;
+  if (row.language && row.languageCode) {
+    languageLine = (
+      <>
+        {row.language}
+        <span className="tw:text-muted-foreground"> ({row.languageCode})</span>
+      </>
+    );
+  } else if (row.language) {
+    languageLine = row.language;
+  } else if (row.languageCode) {
+    languageLine = row.languageCode;
+  }
 
   return (
     <Tooltip open={isHovered} delayDuration={400}>
       <TooltipTrigger asChild>{rowNode}</TooltipTrigger>
       <TooltipContent
-        side="top"
-        align="center"
+        // side="right" — hovered rows sit inside the selector's popover; a top-side tooltip
+        // frequently overlapped the row above (or the search bar for the topmost row). Right-side
+        // places the tooltip clear of the popover's content column and reads as a canonical
+        // "row-details" affordance.
+        side="right"
+        align="start"
         sideOffset={8}
         collisionPadding={16}
-        className="tw:max-w-xs tw:text-center"
+        className="tw:max-w-xs tw:text-start"
         style={{ zIndex: Z_INDEX_OVERLAY }}
       >
         <div className="tw:font-semibold">{row.fullName}</div>
-        {tooltipHasLanguage && (
-          <div className="tw:text-sm">
-            {row.language}
-            {row.languageCode && (
-              <span className="tw:text-muted-foreground"> ({row.languageCode})</span>
-            )}
+        {languageLine && <div className="tw:text-sm">{languageLine}</div>}
+        {scrollGroupLine && (
+          <div className={cn('tw:text-sm', scrollGroupLineIsItalic && 'tw:italic')}>
+            {scrollGroupLine}
           </div>
         )}
-        {!row.isBoundButClosed && row.scrollGroupScrRefLabel && letter && (
-          <div className="tw:text-sm">
-            {row.scrollGroupScrRefLabel}
-            <span className="tw:text-muted-foreground"> ({letter})</span>
-          </div>
-        )}
-        {tooltipBoundBut && <div className="tw:text-sm tw:italic">{tooltipBoundBut}</div>}
         {row.isDisabled && row.disabledReason && (
           <div className="tw:text-sm tw:italic tw:text-muted-foreground">{row.disabledReason}</div>
         )}
--- a/lib/platform-bible-react/src/components/advanced/project-selector/project-selector.component.tsx
+++ b/lib/platform-bible-react/src/components/advanced/project-selector/project-selector.component.tsx
@@ -476,7 +476,18 @@ function ProjectRowView({ row, mode, strings, onClick, onOpen, selectedRowRef }:
         align="start"
         sideOffset={8}
         collisionPadding={16}
-        className="tw:max-w-xs tw:text-start"
+        // `flex-col items-start` overrides shadcn TooltipContent's base `inline-flex items-center
+        // gap-1.5` so the info lines stack vertically instead of laying out as columns. `gap-0.5`
+        // gives them modest breathing room. See `shadcn-ui/tooltip.tsx` for the base classes.
+        className="tw:max-w-xs tw:flex-col tw:items-start tw:gap-0.5 tw:text-start"
+        // `showArrow={false}` — shadcn's TooltipContent arrow renders as an unclipped rotated
+        // square (a diamond) on `side="left"`/`"right"`. The tooltip file explicitly documents
+        // this: top/bottom get a clip-path that reduces the diamond to a triangle, but left/right
+        // deliberately keep the plain diamond ("no consumer today needs a bordered left/right
+        // arrow"). That renders as a small rectangle poking out of the tooltip — not the pointer
+        // shape we want. Suppressing the arrow entirely is the pragmatic fix; the tooltip's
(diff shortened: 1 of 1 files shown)


jolierabideau:
Reviewed with a focus on adherence to our established code patterns. Nice, tightly scoped change — the four-branch consolidation genuinely simplifies the JSX, and the partial-data language handling is a real fix. A few things to address, all in `lib/platform-bible-react/src/components/advanced/project-selector/project-selector.component.tsx`.

## Important

**1. `side="right"` is hardcoded and not RTL-aware** — line 475

Radix's `side` is a *physical* side, not a logical one. In an RTL UI the popover's row content mirrors to the left, so a right-anchored tooltip lands back on top of the rows — reintroducing exactly the overlap this PR is fixing.

`readDirection()` from `@/utils/dir-helper.util` is the settled convention for this in this library: `scroll-group-selector`, `book-chapter-control`, `navigation-history-buttons`, `scripture-results-viewer`, `search-bar`, `tabs-vertical`, and even `shadcn-ui/tabs.tsx` and `shadcn-ui/slider.tsx` all use it. This file doesn't yet, but its siblings in `advanced/` do.

```tsx
const direction: Direction = readDirection();
// ...
side={direction === 'rtl' ? 'left' : 'right'}
```

Worth noting the `showArrow={false}` rationale already covers both `left` and `right`, so nothing else needs to change.

**2. New user-visible strings are composed in code, bypassing this file's own localization convention** — lines 444, 447

The component defines `ProjectSelectorLocalizedStrings` + `DEFAULT_STRINGS` and routes every other user-facing string through it, with caller overrides. The two new compositions don't:

```tsx
? `${row.scrollGroupScrRefLabel} (${letter})`   // parenthesized-suffix format baked in
scrollGroupLine = row.openGroups.map(scrollGroupLetterFromMap).join(', ');   // list separator baked in
```

Neither the `X (Y)` format nor `", "` as a list separator is universal across locales. These want keys alongside `boundButClosedTooltip` — e.g. `scrollGroupWithRefTooltip: '{ref} ({group})'` and a `scrollGroupListSeparator`.

**3. No test coverage for the new branching**

The change adds a four-branch scroll-group selector (lines 437-448) and a three-branch language composer (lines 452-465). `project-selector.component.test.tsx` has 11 behavior tests and asserts nothing about tooltip content today.

The partial-data cases the PR description calls out (`language`-only, `languageCode`-only rendering "without a stray parenthesis") are precisely the kind of thing that should be pinned by a test — right now nothing stops a future refactor from regressing them. The existing RTL harness in that file supports it directly.

## Minor

**4.** `strings.boundButClosedTooltip.replace('{group}', letter)` (line 440) — our convention is `formatReplacementString` from `platform-bible-utils` (see Localization-Guide). Pre-existing, but this PR rewrites the line, so it's a cheap moment to fix.

**5. Styling asymmetry between the two new lines.** `languageLine` is a `ReactNode` and keeps its muted `(en-US)` parenthetical; `scrollGroupLine` is a plain `string`, so `(A)` loses the `tw:text-muted-foreground` treatment it effectively had before. The differing types are the root cause — making both `ReactNode` would restore the parity.

**6.** `hasExtraTooltipContent` (line 319) wasn't extended for the new `openGroups` branch. I think that's *correct* — the chip strip is already visible on the row, so it isn't information the tooltip is uniquely surfacing — but it's undocumented, and it reads like an oversight. One line of comment would stop someone "fixing" it later.

**7. Comment accuracy** (line ~479). `tw:flex-col` does not override `inline-flex` — they're separate tailwind-merge groups, so display stays `inline-flex` and only the direction changes. The rendering is right; the explanation isn't. (`items-start` and `gap-0.5` do override as described.)

**8.** The seven-line `showArrow={false}` comment (lines 483-489) restates rationale already documented in `shadcn-ui/tooltip.tsx` lines 122-128. Duplicated rationale drifts when one copy changes — a one-line pointer to the source would do.

**9.** `tw:max-w-xs` (line 482) is redundant — `TooltipContent`'s base class list already sets it.

**10.** `let scrollGroupLine` plus a parallel `scrollGroupLineIsItalic` flag (lines 437-438): italic is derivable from `row.isBoundButClosed`, or a small helper returning `{ text, isItalic }` would remove the mutable pair.

## Done well

- `showArrow={false}` uses the existing `CUSTOM` prop rather than editing the shadcn file — correctly respects the shadcn edit-discipline rule. The stated arrow behaviour also checks out against `tooltip.tsx`.
- `text-start` / `align="start"` / `items-start` are logical properties rather than `text-left` — the right instinct, which is what makes the hardcoded `side="right"` stand out.
- The class overrides do resolve correctly through the prefix-aware `cn`, so the shadcn base-class override works as intended.
- Comments explain *why* rather than *what*, and all five stories named in the test plan exist.

---

*Note: I reviewed the diff and surrounding code but did not run lint or the test suite locally.*

*AI-assisted review.*

jolierabideau (changes requested):
Reviewed with a focus on adherence to our established code patterns. Nice, tightly scoped change — the four-branch consolidation genuinely simplifies the JSX, and the partial-data language handling is a real fix. A few things to address, all in `lib/platform-bible-react/src/components/advanced/project-selector/project-selector.component.tsx`.

## Important

**1. `side="right"` is hardcoded and not RTL-aware** — line 475

Radix's `side` is a *physical* side, not a logical one. In an RTL UI the popover's row content mirrors to the left, so a right-anchored tooltip lands back on top of the rows — reintroducing exactly the overlap this PR is fixing.

`readDirection()` from `@/utils/dir-helper.util` is the settled convention for this in this library: `scroll-group-selector`, `book-chapter-control`, `navigation-history-buttons`, `scripture-results-viewer`, `search-bar`, `tabs-vertical`, and even `shadcn-ui/tabs.tsx` and `shadcn-ui/slider.tsx` all use it. This file doesn't yet, but its siblings in `advanced/` do.

```tsx
const direction: Direction = readDirection();
// ...
side={direction === 'rtl' ? 'left' : 'right'}
```

Worth noting the `showArrow={false}` rationale already covers both `left` and `right`, so nothing else needs to change.

**2. New user-visible strings are composed in code, bypassing this file's own localization convention** — lines 444, 447

The component defines `ProjectSelectorLocalizedStrings` + `DEFAULT_STRINGS` and routes every other user-facing string through it, with caller overrides. The two new compositions don't:

```tsx
? `${row.scrollGroupScrRefLabel} (${letter})`   // parenthesized-suffix format baked in
scrollGroupLine = row.openGroups.map(scrollGroupLetterFromMap).join(', ');   // list separator baked in
```

Neither the `X (Y)` format nor `", "` as a list separator is universal across locales. These want keys alongside `boundButClosedTooltip` — e.g. `scrollGroupWithRefTooltip: '{ref} ({group})'` and a `scrollGroupListSeparator`.

**3. No test coverage for the new branching**

The change adds a four-branch scroll-group selector (lines 437-448) and a three-branch language composer (lines 452-465). `project-selector.component.test.tsx` has 11 behavior tests and asserts nothing about tooltip content today.

The partial-data cases the PR description calls out (`language`-only, `languageCode`-only rendering "without a stray parenthesis") are precisely the kind of thing that should be pinned by a test — right now nothing stops a future refactor from regressing them. The existing RTL harness in that file supports it directly.

## Minor

**4.** `strings.boundButClosedTooltip.replace('{group}', letter)` (line 440) — our convention is `formatReplacementString` from `platform-bible-utils` (see Localization-Guide). Pre-existing, but this PR rewrites the line, so it's a cheap moment to fix.

**5. Styling asymmetry between the two new lines.** `languageLine` is a `ReactNode` and keeps its muted `(en-US)` parenthetical; `scrollGroupLine` is a plain `string`, so `(A)` loses the `tw:text-muted-foreground` treatment it effectively had before. The differing types are the root cause — making both `ReactNode` would restore the parity.

**6.** `hasExtraTooltipContent` (line 319) wasn't extended for the new `openGroups` branch. I think that's *correct* — the chip strip is already visible on the row, so it isn't information the tooltip is uniquely surfacing — but it's undocumented, and it reads like an oversight. One line of comment would stop someone "fixing" it later.

**7. Comment accuracy** (line ~479). `tw:flex-col` does not override `inline-flex` — they're separate tailwind-merge groups, so display stays `inline-flex` and only the direction changes. The rendering is right; the explanation isn't. (`items-start` and `gap-0.5` do override as described.)

**8.** The seven-line `showArrow={false}` comment (lines 483-489) restates rationale already documented in `shadcn-ui/tooltip.tsx` lines 122-128. Duplicated rationale drifts when one copy changes — a one-line pointer to the source would do.

**9.** `tw:max-w-xs` (line 482) is redundant — `TooltipContent`'s base class list already sets it.

**10.** `let scrollGroupLine` plus a parallel `scrollGroupLineIsItalic` flag (lines 437-438): italic is derivable from `row.isBoundButClosed`, or a small helper returning `{ text, isItalic }` would remove the mutable pair.

## Done well

- `showArrow={false}` uses the existing `CUSTOM` prop rather than editing the shadcn file — correctly respects the shadcn edit-discipline rule. The stated arrow behaviour also checks out against `tooltip.tsx`.
- `text-start` / `align="start"` / `items-start` are logical properties rather than `text-left` — the right instinct, which is what makes the hardcoded `side="right"` stand out.
- The class overrides do resolve correctly through the prefix-aware `cn`, so the shadcn base-class override works as intended.
- Comments explain *why* rather than *what*, and all five stories named in the test plan exist.

---

*Note: I reviewed the diff and surrounding code but did not run lint or the test suite locally.*

*AI-assisted review.*

PR #2154, opened 2026-03-27
Pt 3897 replace preview

katherinejensen00:
https://paratextstudio.atlassian.net/browse/PT-3897
image

- Create a replace preview view options picker as per this V0: https://v0.app/chat/preview-options-picker-s08sC9IIAfJ
- Fix history (I broke it when I implemented auto search)
- Fix some styling that Sebastian pointed out
- Improve keyboard navigation so the editor navigates to the reference without losing the cursor to the editor

Demo Video (had to zip it because it was just a little bit over the size limit for files):
[replace-preview-options-picker.zip](https://github.com/user-attachments/files/26355442/replace-preview-options-picker.zip)

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2154)

---

<a href="https://app.devin.ai/review/paranext/paranext-core/pull/2154" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open with Devin
  </picture>
</a>

---

# Code Review Summary

**Branch**: pt-3897-replace-preview

**Base**: origin/main

**Date**: 2026-04-06

**Review model**: Claude Sonnet 4.6

**Files changed**: 38

## Overview

This PR delivers a replace-preview options picker for the Find/Replace UI, allowing users to customize how replacement previews are displayed (layout, highlight shape, highlight color, monospace font, and whitespace markers). It also bundles a significant set of bug fixes and improvements discovered during PT-3895 work: fixing a blank-screen crash on kebab menu click, fixing the replace button layout in search results, reducing editor focus stealing on arrow key navigation, overhauling the search history system (it was entirely broken by the PT-3895 autosearch feature), adding the ability to select text in the editor and open Find with Ctrl+F to search for it, adding a UX-requested whitespace reference story documenting PT9 vs PT10 differences, and fixing incorrect fonts and heading styles in Find/Replace dropdowns and popovers. New and improved Storybook stories are included throughout.

The replace preview picker has passed UX review and is intended for near-term user testing. It is expected to be simplified or removed once winning options are identified.

## API Changes

- `lib/platform-bible-react` (`RecentSearchesProps<T>`): Added two optional props — `open?: boolean` and `onOpenChange?: (open: boolean) => void` — enabling controlled mode for the popover. Fully backward-compatible.
[...]

PR #2677, opened 2026-08-14
PT-4341: Open Find from any scripture tab type

katherinejensen00:
https://github.com/user-attachments/assets/8aa7de79-fcc3-4b94-b484-4d1f3f651308

# Code Review Summary

**Branch**: pt-4341-find-from-any-tab

**Base**: origin/main

**Date**: 2026-08-13

**Files changed**: 74

## Overview

This branch delivers the Find/Replace feature for Scripture projects and, on top of it, **PT-4341 — "open Find from any scripture tab type."** Find can now be invoked (Ctrl+F) not just from the Scripture editor but from the read-only reference panels (model text, Bible text, commentaries), and it searches the **resource actually displayed** in that panel — not the panel's own container project. In Simple mode a Find tab is seeded into the Column-3 layout and brought to the front on invoke. The change extends the `platformScripture.openFind` command with an explicit source-project parameter, gates the editor-controller id so a non-editor panel never triggers a ~20s controller hang, and adds a shared `useOpenFindShortcut` hook plus pure, unit-tested resolution helpers.

The four analysis passes (api/correctness, style/patterns, compliance/coverage, UX design guide) found **no critical or important correctness issues**. One important compliance gap (a missing Storybook story) and several minor items were surfaced; the actionable ones were fixed during the review (below).

## API Changes

- `lib/platform-bible-utils`: added `DEBOUNCE_CANCELED_ERROR_MESSAGE` const and `DebouncedFunction<TFunc>` type; `debounce()` return type changed from a plain callable to `DebouncedFunction<TFunc>` (callable signature preserved; adds a `cancel()` method) — additive / backward-compatible.
- `lib/platform-bible-react`: `Button` gained a `'subtle'` variant (widens `ButtonProps['variant']`); `RecentSearchesProps.buttonVariant` retyped to `ButtonProps['variant']` and gained optional `open?` / `onOpenChange?` props; `ResultsCard` internal-only behavior change (no prop/signature change).
- `extensions/src/platform-scripture` (`platform-scripture.d.ts`): added `FindHistoryDataTypes` and `IFindHistoryDataProvider` types and the `'platformScripture.findHistory'` project-data-provider interface; `platformScripture.openFind` gained two trailing optional params `selectedText?` and `sourceProjectId?` (backward-compatible); updated return-value JSDoc.
- `FindWebViewOptions` (exported interface, not a `@papi/` surface): added optional `initialSearchText?`.
- `lib/papi-dts/papi.d.ts`: unchanged (no `@papi/` module surface changed).

None.

- [x] **[Compliance] `ReplacePreviewOptions` had no Storybook story** _(fixed during review: added `replace-preview-options.stories.tsx` with `Default` + `InlineGreyBlueMonospace` stories; localized-string mapping replicates `find.component.tsx`, no hardcoded strings)_

[Author response: agreed and asked to add the story now; done in-review.]

### Minor — Consider

- [x] **[Style] Find-highlight amber `rgb(255 213 0 / 40%)` was a hardcoded literal duplicated across two rules in `_editor-overrides.scss`** _(fixed during review: extracted to a single `$find-result-highlight` SCSS variable)_
- [x] **[Style] Five bare `.catch(() => {})` swallowed errors silently in `find.web-view.tsx`** (history persistence, annotation cleanup, result-preview select/annotate) _(fixed during review: added `logger.warn(\`…${getErrorMessage(e)}\`)` to each)_
- [x] **[Compliance/UX] Dead localization keys `%webView_find_searchPlaceholderShort%` and `%webView_find_clearSearch%`** — present in both locales but unreferenced in source _(fixed during review: removed from the en and es blocks; JSON re-validated)_
- [ ] ~~**[UX] The replace-preview-options trigger label "View" is ambiguous** (verb vs. noun); a clearer label like "Display" was suggested~~ _(left as-is for the UX team to decide the copy)_
- [ ] ~~**[UX] Color-pair labels use title case ("Red / Cyan", "Red / Green", "Grey / Blue")** where the sentence-case rule would give "Red / cyan"~~ _(intentional symmetric styling for color pairs)_
- [ ] ~~**[Compliance] `electronmon` devDependency bumped `^2.0.3` → `^2.0.4`**~~ _(incidental to the feature, but retained — a helpful dev-start improvement)_
- [ ] ~~**[Compliance] The e2e suite doesn't exercise the PT-4341 panel Ctrl+F triggers**~~ _(deferred; a full e2e needs installed DBL-resource fixtures, and the trigger logic is covered by unit + hook tests)_
- [ ] ~~**[API] `openFind` now takes three positional optionals** (`editorWebViewId?`, `selectedText?`, `sourceProjectId?`), which is easy to mis-order at future call sites~~ _(conforms to the established sibling `open*` command convention; each param is documented in the `.d.ts` TSDoc)_

[Author response: fixed M3/M4/M8 in-review; dismissed the rest with the reasons above.]

### Template Propagation

None. No `#region shared with <url>` markers overlap changed lines, and no extension config/build files changed (the only build-config change is the root `package.json` `electronmon` bump).

## Positive Observations

- The `openFind` resolution logic is extracted into a pure, separately unit-tested `resolveFindInvocation`, with a well-documented reason for **not** forwarding `editorWebViewId` to non-editor panels (avoids a ~20s `useWebViewController` hang), and the mirrored `SCRIPTURE_EDITOR_WEBVIEW_TYPE` literal is backed by a pinning test.
- New pure logic is uniformly unit-tested: `resolveFindInvocation`, `getOpenFindTriggerArgs`, `useOpenFindShortcut` (jsdom keydown/no-op/cleanup), `FindHistoryDataProviderEngine` (invalid JSON, cap, dedup, per-project keys), and `debounce.cancel`.
- The keyboard-shortcuts catalog (`scripture-find`) was correctly updated per the catalog rule — new context/comment and a `locations` entry pointing at `use-open-find-shortcut.hook.ts`.
- `FindHistoryDataProviderEngine` follows the standard `DataProviderEngine` pattern with storage injected for testability without an `ExecutionToken`, reusing prior storage keys for data continuity.
- Localization is complete and parallel across en/es; `…` (proper ellipsis) and sentence-case (`"Replace all"`) refinements were applied consistently, editing existing keys rather than minting new ones.
- The `subtle` Button variant carries a proper `// CUSTOM:` annotation and a matching Storybook story; `RecentSearchesProps.buttonVariant` retyped to `ButtonProps['variant']` keeps it in sync automatically.
- Accessibility: the preview-options control has both a visible label and `aria-label`; grouped radio/checkbox controls use semantic `<fieldset>`/`<legend>`; the popover width (288px) stays within the 300px narrow-panel target.
- The Simple-mode Column-3 Find tab seeding is paired with updated `simple-layout.data.test.ts` assertions (tab count and webViewType count), and the literal `webViewType` is documented as needing to match `findWebViewType`.
- The `debounce()` change is cleanly backward-compatible (callable preserved; `cancel()` + `DEBOUNCE_CANCELED_ERROR_MESSAGE` additive, with a clear cancellation-rejection contract).

## Interview Notes

- **Purpose** (author-confirmed): PT-4341 — enable opening Find from any scripture tab type (model-text, Bible-text, commentary panels), not just the scripture editor; resolve the correct displayed-resource as the search source; and dock/seed Find as a Column-3 tab in Simple mode. Built on the broader Find/Replace feature work already on this branch.
- The author directed this work closely, including commissioning adversarial reviews and a `/code-review max` pass (which caught and fixed a real Important defect earlier on the branch — a stale `editorWebViewId` resurrected on a panel-triggered Find reload). On the design walkthrough the author answered **"all solid"** and had **nothing further to flag** — no areas of uncertainty and no deferral to AI; understanding is demonstrated.
- In-review changes made: added the `ReplacePreviewOptions` Storybook story; extracted the find-highlight color to a SCSS variable; added `logger.warn` to five silent `.catch` handlers; removed two dead localization keys.

## In-Review Quality Check

Changes were made during the interview, so quality checks were run on the four affected files:

- **Prettier**: one auto-fix (a long `.catch(...)` line wrapped in `find.web-view.tsx`); the other three files were already clean.
- **Typecheck**: zero new errors in `platform-scripture` and `platform-scripture-editor`.
- **Stylelint**: clean on `_editor-overrides.scss` (the new `$find-result-highlight` variable raised no issues).
- **ESLint**: clean on `find.web-view.tsx` and the new story.
- **Tests**: 79 find-related tests pass (`open-find.utils`, `find-trigger`, `find-history.data-provider`, `find.utils`, `use-open-find-shortcut`, `model-text-panel.component`).

No unfixable failures. (Pre-existing, unrelated failures in `eslint-plugin-paranext` rule tests and `lib/platform-bible-react` Storybook/chromium stories are not from this branch and were left untouched.)

## Suggested Review Focus

Prioritized areas for the author–reviewer meeting. The author reported no uncertainties ("all solid"); these are the subtlest pieces worth a second set of eyes:

- [ ] **`editorWebViewId` type-gate + Find-provider presence-check** — Find forwards an editor-controller id only for real editor tabs, and a panel-triggered reload *clears* a stale editor id (the fix that avoids a wrong-editor coupling / ~20s `useWebViewController` hang). Confirm the reload-vs-restore distinction holds (restore omits the `editorWebViewId` key, so the saved value is preserved).
- [ ] **Displayed-resource-as-source** — the panels pass the *selected resource's* project id (`modelResourceProjectId` / `resourceProjectId`), never the panel's own container project. Confirm this is right for every resource type (DBL Bible text, DBL commentary, project reference).
- [ ] **Seeded Column-3 Find tab + `existingId:'?'` bring-to-front** — relies on Simple mode being stateless (re-read from `simpleLayout`, never persisted) so the seeded tab is focused in place rather than spawning a second Find panel; worth a cold-start manual smoke.
- [ ] **Related follow-up filed**: PT-4372 (make `getWebViewController` fast-reject a known-absent/wrong-type controller instead of hanging ~20s) — the platform-level fix behind the Find-specific `editorWebViewId` gate.
- [ ] Scope note: this branch is the whole Find/Replace feature, not only PT-4341; the broader Find UI (replace-preview options, recent searches, results card) is in-scope for review too.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2677)


jolierabideau (changes requested):
Thanks Katherine! I think we also need to add the Find menu items to the model texts, bible texts, and commentaries web views. Unless that is specifically left for a follow up PR. Also Claude pointed out that there is a 5th scripture tab type - Scripture Text Grid, just want to document as an open question if we should add this to that one as well. I would assume maybe? But can be done in a follow up.

> @jolierabideau reviewed 24 files and all commit messages, and made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2677#-:-P-FavtIDVjfHk9_eWgG:b-ehg4tp)* status: all files reviewed, 2 unresolved discussions (waiting on katherinejensen00).

___
*[`extensions/src/platform-scripture-editor/src/use-open-find-shortcut.hook.ts` line 24 at r1](https://reviewable.io/reviews/paranext/paranext-core/2677#-P-FbGowAmzMzdOzUhZj:-P-FbGowAmzMzdOzUhZk:b-xpwckw) ([raw file](https://github.com/paranext/paranext-core/blob/299a66dc149a4da09af875805a03aa69601b8f9b/extensions/src/platform-scripture-editor/src/use-open-find-shortcut.hook.ts#L24)):*
> ```XML
>       if (!event.ctrlKey || event.key.toLowerCase() !== 'f') return;
>       // Intercept Ctrl+F like the editor's handler; a panel with no resource resolved then no-ops below.
>       event.preventDefault();
> ```

When no resource is resolved yet, this swallows Ctrl+F and does nothing — no panel, no message, not even a log line. PT‑4336's NN‑1D is precisely "always see feedback… there is no situation where a search term is entered but the result list stays or becomes empty" and the demo bullet is "Find never goes silent." A silent keystroke in a Bible-text tab that is still loading its resource is the same trust failure in a different place. Suggest either opening Find with the panel's own project as a fallback, or at minimum `logger.debug` + letting the event through, so the behavior is diagnosable.
___
*[`src/renderer/components/docking/simple-layout.data.ts` line 97 at r1](https://reviewable.io/reviews/paranext/paranext-core/2677#-P-FbL17Dr20sJIERHn2:-P-FbL17Dr20sJIERHn3:b5e181x) ([raw file](https://github.com/paranext/paranext-core/blob/299a66dc149a4da09af875805a03aa69601b8f9b/src/renderer/components/docking/simple-layout.data.ts#L97)):*
> ```XML
>                 },
>               },
>               {
> ```

Is this PR branched off of Matt's? If so, ignore this comment as it probably should go on his PR. If not, we may need to see what his solution is before this is merged.

The seeded tab carries `state: {}` and no `projectId`, and `FindWebViewProvider.getWebView` resolves `projectId = getWebViewOptions.projectId || savedWebView.projectId || undefined` — so on a cold Simple-mode start the Find tab exists with **no project**. AC 4 says invoking Find brings the tab to front "with the resolved source pre-selected", which is fine, but what does the tab render _before_ the first invoke? Worth confirming it shows the placeholder/empty state rather than an error or a spinner — again NN‑1D territory. Also: this adds a 4th always-mounted web view iframe to every Simple-mode startup; was startup timing checked (`PT_STARTUP_MARKS`)?

jolierabideau (changes requested):
> @jolierabideau made 11 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2677)* status: all files reviewed, 12 unresolved discussions (waiting on katherinejensen00).

___
*[`src/renderer/components/docking/simple-layout.data.ts` line 97 at r1](https://reviewable.io/reviews/paranext/paranext-core/2677#-P-FbL17Dr20sJIERHn2:-P-FcXQd3qMrV8yMo7Vw:b-5mjjyk) ([raw file](https://github.com/paranext/paranext-core/blob/299a66dc149a4da09af875805a03aa69601b8f9b/src/renderer/components/docking/simple-layout.data.ts#L97)):*

Is this PR branched off of Matt's? If so, ignore this comment as it probably should go on his PR. If not, we may need to see what his solution is before this is merged.

The seeded tab carries `state: {}` and no `projectId`, and `FindWebViewProvider.getWebView` resolves `projectId = getWebViewOptions.projectId || savedWebView.projectId || undefined` — so on a cold Simple-mode start the Find tab exists with **no project**. AC 4 says invoking Find brings the tab to front "with the resolved source pre-selected", which is fine, but what does the tab render _before_ the first invoke? Worth confirming it shows the placeholder/empty state rather than an error or a spinner — again NN‑1D territory. Also: this adds a 4th always-mounted web view iframe to every Simple-mode startup; was startup timing checked (`PT_STARTUP_MARKS`)?
</blockquote></details>

Claude also pointed out 

The file two entries above already imports `SCRIPTURE_EDITOR_WEBVIEW_TYPE` from `@shared/models/web-view.model` — and [web-view.model.ts:310-316](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/shared/models/web-view.model.ts#L310-L316) exists for exactly this purpose, with the same "core code cannot import extension source, so the value is mirrored here as the single core-side copy" rationale. The established pattern is a named, documented export in that model file, not an inline literal in layout data. Please add `FIND_WEBVIEW_TYPE` there and reference it here, so the next core-side consumer doesn't mint a 2nd copy.
___
*[`extensions/src/platform-scripture/contributions/localizedStrings.json` at r1](https://reviewable.io/reviews/paranext/paranext-core/2677#-P-Fer8c4-f8EMK3Sw2i:-P-Fer8c4-f8EMK3Sw2j:b-a9w6va) ([raw file](https://github.com/paranext/paranext-core/blob/299a66dc149a4da09af875805a03aa69601b8f9b/extensions/src/platform-scripture/contributions/localizedStrings.json)):*
Verified: `%webView_find_searchPlaceholderShort%` is genuinely unreferenced, so the removal is safe. (Note the PR summary in the description claims `%webView_find_clearSearch%` was removed too — it wasn't, and shouldn't be; it's used at `find.component.tsx:567,578`. Just a stale description.)

One thing I'd ask the author directly: the PR description is a `/review-paratext` self-review whose "Files changed: 74" and API-changes list describe the whole Find/Replace branch, but `main` now has that work (#2154) and this PR is a 24-file delta. The description should be regenerated so the reviewer isn't reading findings about code that isn't in the diff.
___
*[`extensions/src/platform-scripture/src/main.ts` line 305 at r1](https://reviewable.io/reviews/paranext/paranext-core/2677#-P-Fe6tEAt7-YZEo-_Im:-P-Fe6tEAt7-YZEo-_In:b-hctfr5) ([raw file](https://github.com/paranext/paranext-core/blob/299a66dc149a4da09af875805a03aa69601b8f9b/extensions/src/platform-scripture/src/main.ts#L305)):*
> ```XML
>   if (!projectId) {
>     logger.debug('No project!');
>     return undefined;
> ```

**`extensions/src/platform-scripture/src/main.ts:305-311` (`editorScrollGroupId` in `options`) together with `find.web-view-provider.ts:47` (`scrollGroupScrRef: getWebViewOptions.editorScrollGroupId`)**

The provider assigns the incoming scroll group **unconditionally** — there's no `?? savedWebView.scrollGroupScrRef`. Resource panels deliberately have no scroll group ([platform-scripture-editor/src/main.ts:1082-1084](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/extensions/src/platform-scripture-editor/src/main.ts#L1082-L1084): _"Intentionally does not force scrollGroupScrRef in simple mode… not scroll-synced with the scripture editor"_), and your own test fixture models `panelDef` with `scrollGroupScrRef: undefined`. So: open Find from the editor (joins scroll group 0) → Ctrl+F in a Bible-text tab → `shouldReloadExistingFind` returns true → reload with `editorScrollGroupId: undefined` → **the already-open Find panel is dropped out of its scroll group**. Was that verified? If it's intended, the provider should say so; if not, either preserve the saved group or pass the panel's group through.
___
*[`extensions/src/platform-scripture/src/find/find-web-view-state.utils.ts` line 29 at r1](https://reviewable.io/reviews/paranext/paranext-core/2677#-P-FeG4k7llAwNZNyRxI:-P-FeG4k7llAwNZNyRxJ:b-kmaj0l) ([raw file](https://github.com/paranext/paranext-core/blob/299a66dc149a4da09af875805a03aa69601b8f9b/extensions/src/platform-scripture/src/find/find-web-view-state.utils.ts#L29)):*
> ```XML
>         ? getWebViewOptions.editorWebViewId
>         : savedWebView.state?.editorWebViewId,
>     ...(getWebViewOptions.initialSearchText
> ```

The clear-vs-preserve distinction rides on _key presence_ across a process boundary. I traced it and it does survive: `rpc.model.ts` uses `serialize`/`deserialize` from `platform-bible-utils`, which map `undefined → null → undefined` and keep the key — so the behavior is correct today. But it's load-bearing on a non-obvious property of one serializer, it isn't mentioned in the doc comment, and no test covers the round trip. Suggest either a comment naming that dependency plus a serialize/deserialize round-trip test, or an explicit signal (e.g. `clearEditorWebViewId?: boolean`) that can't be lost in transit.
___
*[`extensions/src/platform-scripture/src/find/open-find.utils.ts` line 7 at r1](https://reviewable.io/reviews/paranext/paranext-core/2677#-P-FcfHY42Zpi_ewllF2:-P-FcfHY42Zpi_ewllF3:b98yth9) ([raw file](https://github.com/paranext/paranext-core/blob/299a66dc149a4da09af875805a03aa69601b8f9b/extensions/src/platform-scripture/src/find/open-find.utils.ts#L7)):*
> ```XML
>  * not import the platform-scripture-editor constant) — it must equal
>  * `SCRIPTURE_EDITOR_WEBVIEW_TYPE` in
>  * `extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts`.
> ```

Introducing the named const is an improvement, but `platform-scripture` already has five bare `'platformScriptureEditor.react'` literals — including [find.web-view.tsx:260](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/extensions/src/platform-scripture/src/find.web-view.tsx#L260), which is the exact `useWebViewController(...)` call this new gate exists to protect, plus `checks-side-panel.web-view.tsx:56,139`, `checklist.web-view.tsx:631`, and `manage-books.web-view.tsx:61`. If the const is worth a pinning test, it's worth being the one copy in the extension — at least convert `find.web-view.tsx:260` in this PR.
___
*[`extensions/src/platform-scripture/src/find/open-find.utils.ts` line 70 at r1](https://reviewable.io/reviews/paranext/paranext-core/2677#-P-FeKXj5qi0-xVoIIfp:-P-FeKXj5qi0-xVoIIfq:b-5xycla) ([raw file](https://github.com/paranext/paranext-core/blob/299a66dc149a4da09af875805a03aa69601b8f9b/extensions/src/platform-scripture/src/find/open-find.utils.ts#L70)):*
> ```XML
>  */
> export function shouldReloadExistingFind(
>   existingFind: SavedWebViewDefinition | undefined,
> ```

Any Ctrl+F with a live text selection now forces a full reload of an already-open Find panel. That's how the pre-fill reaches it (pre-existing), but from a resource panel the common case _is_ "text selected" — so panel Ctrl+F will remount Find and discard in-progress results/scope each time. Given NN‑1D ("no situation where a search term is entered but the result list… becomes empty"), worth confirming the reload doesn't flash an empty result list.
___
*[`extensions/src/platform-scripture-editor/src/find-trigger.util.ts` line 42 at r1](https://reviewable.io/reviews/paranext/paranext-core/2677#-P-FctVUBddJsZfhQJHe:-P-FctVUBddJsZfhQJHf:bsjpd79) ([raw file](https://github.com/paranext/paranext-core/blob/299a66dc149a4da09af875805a03aa69601b8f9b/extensions/src/platform-scripture-editor/src/find-trigger.util.ts#L42)):*
> ```XML
>  * (Find then opens without pre-filling).
>  */
> export function getEditorOpenFindArgs(
> ```

This wrapper exists to normalize `selectedText ?? ''` and returns an object whose fields are immediately destructured back into positional args at the call site — the editor call got _less_ readable (`findArgs.webViewId` where `webViewId` was in scope), and it carries two tests that assert `?? ''`. Either inline it, or better: have the editor path call `getOpenFindTriggerArgs` too (passing the editor's own project id as the source) so both triggers share one code path. Related: `getOpenFindTriggerArgs`'s `webViewId: string | undefined` branch is unreachable from both call sites — `id` is always supplied by `WebViewProps`.
___
*[`extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx` line 1158 at r1](https://reviewable.io/reviews/paranext/paranext-core/2677#-P-Fclv2-Q_lEqPWz4gN:-P-Fclv2-Q_lEqPWz4gO:bm7iyyc) ([raw file](https://github.com/paranext/paranext-core/blob/299a66dc149a4da09af875805a03aa69601b8f9b/extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx#L1158)):*
> ```XML
>       }
> 
>       // Find dialog trigger listener
> ```

Two divergences at the site the PR touched: (a) the editor keeps its own inline Ctrl+F branch instead of using the new `useOpenFindShortcut` hook, so there are now two Ctrl+F→openFind implementations to keep in sync; (b) the editor's `papi.commands.sendCommand(...)` is still a floating promise with no `.catch`, while the new hook logs a `logger.warn` on failure. Same call, same failure mode, different handling — pick one.
___
*[`extensions/src/platform-scripture-editor/src/resource-text-panel.web-view.tsx` line 333 at r1](https://reviewable.io/reviews/paranext/paranext-core/2677#-P-Fcyuz0EVyUxiBE7NZ:-P-Fcyuz0EVyUxiBE7N_:b-oij9wb) ([raw file](https://github.com/paranext/paranext-core/blob/299a66dc149a4da09af875805a03aa69601b8f9b/extensions/src/platform-scripture-editor/src/resource-text-panel.web-view.tsx#L333)):*
> ```XML
>   // projectId). Computed in a useMemo so it's a stable const for the Ctrl+F effect deps and the DBL
>   // lookup isn't repeated on every render.
>   const { resourceProjectId, dblMatch } = useMemo(() => {
> ```

The stated rationale — _"Computed in a useMemo so it's a stable const for the Ctrl+F effect deps"_ — isn't right: the effect depends on `resourceProjectId`, a **string**, so reference stability is irrelevant; the plain `let` would have worked identically. And `dblResources` as a dep is likely a fresh array identity each render, so the memo probably doesn't skip the lookup either. Not harmful, but the comment will mislead the next reader — either drop the refactor or fix the reason.
___
*[`extensions/src/platform-scripture-editor/src/use-open-find-shortcut.hook.ts` line 20 at r1](https://reviewable.io/reviews/paranext/paranext-core/2677#-P-FcNg55rM7gVpZU38w:-P-FcNg6DnviVZicGc_o:b-xa7m6m) ([raw file](https://github.com/paranext/paranext-core/blob/299a66dc149a4da09af875805a03aa69601b8f9b/extensions/src/platform-scripture-editor/src/use-open-find-shortcut.hook.ts#L20)):*
> ```XML
>   displayedResourceProjectId: string | undefined,
> ): void {
>   useEffect(() => {
> ```

This is the biggest pattern comment. `ADR-0002` in [.context/standards/Architecture-Decisions.md](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/.context/standards/Architecture-Decisions.md#L62-L86) records that app-global keyboard shortcuts go through the main-process `before-input-event` handler, and it explicitly **rejects** the alternative this PR takes: _"(a) renderer-level global `keydown` — rejected: web-view iframes are `about:srcdoc`, so their key events don't bubble to the top renderer; coverage gaps unless duplicated into every web-view."_ Ctrl+F is now, by this ticket's own framing, an app-wide expectation, and the PR duplicates the listener into every web view — which is exactly why finding #5 (the grid tab) is a gap and why the next scripture tab type will be one too.

There's a real counter-argument (the handler needs the _focused_ web view's id and its text selection, which main doesn't have), so this may well be the right call — but per CLAUDE.md's "Recording Architecture Decisions", a decision that diverges from an existing accepted ADR needs an entry (or an amendment to ADR‑0002) in the same change. Nothing was added to the log.
___
*[`src/stories/keyboard-shortcuts.data.ts` line 201 at r1](https://reviewable.io/reviews/paranext/paranext-core/2677#-P-Febuh6yedPmU03mzL:-P-Febuh6yedPmU03mzM:b-3k284s) ([raw file](https://github.com/paranext/paranext-core/blob/299a66dc149a4da09af875805a03aa69601b8f9b/src/stories/keyboard-shortcuts.data.ts#L201)):*
> ```XML
>     purpose: 'Open the find dialog',
>     category: 'Navigation',
>     context: 'Scripture editor, model text, Bible text, and commentary web views',
> ```

`locations` gained the hook file but not the two web views that mount it (`model-text-panel.web-view.tsx`, `resource-text-panel.web-view.tsx`). The catalog rule asks for the handler locations; the mount sites are what a reader greps for.

katherinejensen00 (commented):
Thanks — we talked this through and landed on no menu item being needed. I did add the Text Collection (Scripture Text Grid) in this PR. It's the one tab with no single "displayed resource," so Ctrl+F there searches the resource holding the caret — useFocusedResourceProjectId tracks the focused cell via the cells' data-project-id, and I added that attribute to the two cells that were missing it (single-resource and chapter-context).

> @katherinejensen00 made 13 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2677#-:-P-M0ZNX3sORsgf97KTe:b-ulxmyf)* status: 9 of 37 files reviewed, 12 unresolved discussions (waiting on jolierabideau).

___
*[`extensions/src/platform-scripture/contributions/localizedStrings.json` at r1](https://reviewable.io/reviews/paranext/paranext-core/2677#-P-Fer8c4-f8EMK3Sw2i:-P-M146VAJLHzLquQyvS:b-xuybxb) ([raw file](https://github.com/paranext/paranext-core/blob/299a66dc149a4da09af875805a03aa69601b8f9b/extensions/src/platform-scripture/contributions/localizedStrings.json)):*

Verified: `%webView_find_searchPlaceholderShort%` is genuinely unreferenced, so the removal is safe. (Note the PR summary in the description claims `%webView_find_clearSearch%` was removed too — it wasn't, and shouldn't be; it's used at `find.component.tsx:567,578`. Just a stale description.)

One thing I'd ask the author directly: the PR description is a `/review-paratext` self-review whose "Files changed: 74" and API-changes list describe the whole Find/Replace branch, but `main` now has that work (#2154) and this PR is a 24-file delta. The description should be regenerated so the reviewer isn't reading findings about code that isn't in the diff.
</blockquote></details>

Correct on both counts — clearSearch is still present and still used at those two lines. I've rewritten the description to describe only this delta rather than the whole Find/Replace branch that landed via #2154.
___
*[`extensions/src/platform-scripture/src/main.ts` line 305 at r1](https://reviewable.io/reviews/paranext/paranext-core/2677#-P-Fe6tEAt7-YZEo-_Im:-P-M18mp4OWBp98M47nr:bbmi0d0) ([raw file](https://github.com/paranext/paranext-core/blob/299a66dc149a4da09af875805a03aa69601b8f9b/extensions/src/platform-scripture/src/main.ts#L305)):*

**`extensions/src/platform-scripture/src/main.ts:305-311` (`editorScrollGroupId` in `options`) together with `find.web-view-provider.ts:47` (`scrollGroupScrRef: getWebViewOptions.editorScrollGroupId`)**

The provider assigns the incoming scroll group **unconditionally** — there's no `?? savedWebView.scrollGroupScrRef`. Resource panels deliberately have no scroll group ([platform-scripture-editor/src/main.ts:1082-1084](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/extensions/src/platform-scripture-editor/src/main.ts#L1082-L1084): _"Intentionally does not force scrollGroupScrRef in simple mode… not scroll-synced with the scripture editor"_), and your own test fixture models `panelDef` with `scrollGroupScrRef: undefined`. So: open Find from the editor (joins scroll group 0) → Ctrl+F in a Bible-text tab → `shouldReloadExistingFind` returns true → reload with `editorScrollGroupId: undefined` → **the already-open Find panel is dropped out of its scroll group**. Was that verified? If it's intended, the provider should say so; if not, either preserve the saved group or pass the panel's group through.
</blockquote></details>

Not verified, and you were right. Took the "preserve the saved group" branch — it also fixes a case neither of us mentioned: a content reload/restore passes no options at all, so Find's scroll group was being wiped on every app restart. I documented the consequence at the site rather than leaving it implicit: a panel-triggered Find keeps whatever group a prior editor-triggered Find left it in, so clicking a result moves the scripture editor. That's the better of the two options — "pass the panel's group through" yields undefined in Simple mode, and since a panel trigger also clears the editor coupling, clicking a result would move nothing at all. Neither option can navigate the panel itself; that needs a controller the read-only panels don't register (PT‑4372).
___
*[`extensions/src/platform-scripture/src/find/find-web-view-state.utils.ts` line 29 at r1](https://reviewable.io/reviews/paranext/paranext-core/2677#-P-FeG4k7llAwNZNyRxI:-P-M1EB3AYuhvx6g8HcX:brmz0s5) ([raw file](https://github.com/paranext/paranext-core/blob/299a66dc149a4da09af875805a03aa69601b8f9b/extensions/src/platform-scripture/src/find/find-web-view-state.utils.ts#L29)):*

The clear-vs-preserve distinction rides on _key presence_ across a process boundary. I traced it and it does survive: `rpc.model.ts` uses `serialize`/`deserialize` from `platform-bible-utils`, which map `undefined → null → undefined` and keep the key — so the behavior is correct today. But it's load-bearing on a non-obvious property of one serializer, it isn't mentioned in the doc comment, and no test covers the round trip. Suggest either a comment naming that dependency plus a serialize/deserialize round-trip test, or an explicit signal (e.g. `clearEditorWebViewId?: boolean`) that can't be lost in transit.
</blockquote></details>

Went with the explicit signal. openFind now sets clearEditorWebViewId: true for a non-editor trigger, and the provider keys off that. A positive true can't be lost the way an undefined-valued key can, so the behavior no longer depends on the serializer's undefined → null → undefined round trip at all. Added a round-trip test through the real serialize/deserialize anyway, since that's the property the fix is claiming.
___
*[`extensions/src/platform-scripture/src/find/open-find.utils.ts` line 7 at r1](https://reviewable.io/reviews/paranext/paranext-core/2677#-P-FcfHY42Zpi_ewllF2:-P-M1l9hA3uZiTmz5B9g:b-em9swv) ([raw file](https://github.com/paranext/paranext-core/blob/299a66dc149a4da09af875805a03aa69601b8f9b/extensions/src/platform-scripture/src/find/open-find.utils.ts#L7)):*

Introducing the named const is an improvement, but `platform-scripture` already has five bare `'platformScriptureEditor.react'` literals — including [find.web-view.tsx:260](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/extensions/src/platform-scripture/src/find.web-view.tsx#L260), which is the exact `useWebViewController(...)` call this new gate exists to protect, plus `checks-side-panel.web-view.tsx:56,139`, `checklist.web-view.tsx:631`, and `manage-books.web-view.tsx:61`. If the const is worth a pinning test, it's worth being the one copy in the extension — at least convert `find.web-view.tsx:260` in this PR.
</blockquote></details>

Converted all five, not just that one. The const moved out of open-find.utils.ts (too Find-specific to be the extension's canonical home) into src/scripture-editor-web-view-type.const.ts, with the pinning test moved alongside it. find.web-view.tsx, checks-side-panel.web-view.tsx (×2), checklist.web-view.tsx, and manage-books.web-view.tsx all import it now.
___
*[`extensions/src/platform-scripture/src/find/open-find.utils.ts` line 70 at r1](https://reviewable.io/reviews/paranext/paranext-core/2677#-P-FeKXj5qi0-xVoIIfp:-P-M1nV_8y3JgEXFKygR:baa7glq) ([raw file](https://github.com/paranext/paranext-core/blob/299a66dc149a4da09af875805a03aa69601b8f9b/extensions/src/platform-scripture/src/find/open-find.utils.ts#L70)):*

Any Ctrl+F with a live text selection now forces a full reload of an already-open Find panel. That's how the pre-fill reaches it (pre-existing), but from a resource panel the common case _is_ "text selected" — so panel Ctrl+F will remount Find and discard in-progress results/scope each time. Given NN‑1D ("no situation where a search term is entered but the result list… becomes empty"), worth confirming the reload doesn't flash an empty result list.
</blockquote></details>

It does flash — a reload remounts the panel, so an empty result list between mount and first results is unavoidable once a reload happens. So I attacked the frequency instead: the clause now fires only when the incoming selection differs from the term the panel already holds (findSearchTerm is persisted via useWebViewState, so it reflects the live search box). Repeated Ctrl+F on the same selection no longer reloads at all — it just brings the panel to front.
___
*[`extensions/src/platform-scripture-editor/src/find-trigger.util.ts` line 42 at r1](https://reviewable.io/reviews/paranext/paranext-core/2677#-P-FctVUBddJsZfhQJHe:-P-M1tAq0giyJtG1FCMU:b7984wt) ([raw file](https://github.com/paranext/paranext-core/blob/299a66dc149a4da09af875805a03aa69601b8f9b/extensions/src/platform-scripture-editor/src/find-trigger.util.ts#L42)):*

This wrapper exists to normalize `selectedText ?? ''` and returns an object whose fields are immediately destructured back into positional args at the call site — the editor call got _less_ readable (`findArgs.webViewId` where `webViewId` was in scope), and it carries two tests that assert `?? ''`. Either inline it, or better: have the editor path call `getOpenFindTriggerArgs` too (passing the editor's own project id as the source) so both triggers share one code path. Related: `getOpenFindTriggerArgs`'s `webViewId: string | undefined` branch is unreachable from both call sites — `id` is always supplied by `WebViewProps`.
</blockquote></details>

Took the "better" option. getEditorOpenFindArgs is deleted and the editor now mounts useOpenFindShortcut like the panels do, passing its own project as the source. And you were right that the undefined branch was unreachable — WebViewProps extends SavedWebViewDefinition, which requires id, so the parameter is now webViewId: string and the guard is gone.
___
*[`extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx` line 1158 at r1](https://reviewable.io/reviews/paranext/paranext-core/2677#-P-Fclv2-Q_lEqPWz4gN:-P-M22PK05k3OiNUR_aV:bud6haa) ([raw file](https://github.com/paranext/paranext-core/blob/299a66dc149a4da09af875805a03aa69601b8f9b/extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx#L1158)):*

Two divergences at the site the PR touched: (a) the editor keeps its own inline Ctrl+F branch instead of using the new `useOpenFindShortcut` hook, so there are now two Ctrl+F→openFind implementations to keep in sync; (b) the editor's `papi.commands.sendCommand(...)` is still a floating promise with no `.catch`, while the new hook logs a `logger.warn` on failure. Same call, same failure mode, different handling — pick one.
</blockquote></details>

Picked the hook for both. The editor's Ctrl+F branch is gone; what remains in that keydown handler is the marker-menu and insert-comment logic, and the old if/else collapsed to a plain if since Ctrl+F and the comment hotkeys are disjoint. webViewId dropped out of that effect's dep list with it. One .catch(logger.warn), in the hook.
___
*[`extensions/src/platform-scripture-editor/src/resource-text-panel.web-view.tsx` line 333 at r1](https://reviewable.io/reviews/paranext/paranext-core/2677#-P-Fcyuz0EVyUxiBE7NZ:-P-M26etEQ_bTQIHT8M-:bqal2b) ([raw file](https://github.com/paranext/paranext-core/blob/299a66dc149a4da09af875805a03aa69601b8f9b/extensions/src/platform-scripture-editor/src/resource-text-panel.web-view.tsx#L333)):*

The stated rationale — _"Computed in a useMemo so it's a stable const for the Ctrl+F effect deps"_ — isn't right: the effect depends on `resourceProjectId`, a **string**, so reference stability is irrelevant; the plain `let` would have worked identically. And `dblResources` as a dep is likely a fresh array identity each render, so the memo probably doesn't skip the lookup either. Not harmful, but the comment will mislead the next reader — either drop the refactor or fix the reason.
</blockquote></details>

Dropped the refactor. Back to the plain let it was before, with a comment that says the thing actually worth saying: resourceProjectId is the displayed resource's project, not the panel's own projectId prop.
___
*[`extensions/src/platform-scripture-editor/src/use-open-find-shortcut.hook.ts` line 20 at r1](https://reviewable.io/reviews/paranext/paranext-core/2677#-P-FcNg55rM7gVpZU38w:-P-M2Cvj3iQ0S3QPyqgd:b43jpor) ([raw file](https://github.com/paranext/paranext-core/blob/299a66dc149a4da09af875805a03aa69601b8f9b/extensions/src/platform-scripture-editor/src/use-open-find-shortcut.hook.ts#L20)):*

This is the biggest pattern comment. `ADR-0002` in [.context/standards/Architecture-Decisions.md](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/.context/standards/Architecture-Decisions.md#L62-L86) records that app-global keyboard shortcuts go through the main-process `before-input-event` handler, and it explicitly **rejects** the alternative this PR takes: _"(a) renderer-level global `keydown` — rejected: web-view iframes are `about:srcdoc`, so their key events don't bubble to the top renderer; coverage gaps unless duplicated into every web-view."_ Ctrl+F is now, by this ticket's own framing, an app-wide expectation, and the PR duplicates the listener into every web view — which is exactly why finding #5 (the grid tab) is a gap and why the next scripture tab type will be one too.

There's a real counter-argument (the handler needs the _focused_ web view's id and its text selection, which main doesn't have), so this may well be the right call — but per CLAUDE.md's "Recording Architecture Decisions", a decision that diverges from an existing accepted ADR needs an entry (or an amendment to ADR‑0002) in the same change. Nothing was added to the log.
</blockquote></details>

 Fair catch — added ADR‑0013. The short version: openFind isn't a zero-argument command. It needs the focused tab's web view id, the project of the scripture that tab is showing (for a reference panel, the displayed resource rather than the tab's own projectId), and that view's text selection to pre-fill. before-input-event runs in the main process, which can identify the focused window but not the focused tab, and can't read a selection inside an about:srcdoc iframe — so the ADR‑0002 route would open Find against the wrong scripture and never pre-fill. ADR‑0013 records that, keeps the one-shared-hook constraint that answers ADR‑0002's real objection (duplication), names the deferred general fix, and is explicit that per-tab opt-in is a genuine coverage gap. ADR‑0002 now carries a pointer to the narrowing.
___
*[`extensions/src/platform-scripture-editor/src/use-open-find-shortcut.hook.ts` line 24 at r1](https://reviewable.io/reviews/paranext/paranext-core/2677#-P-FbGowAmzMzdOzUhZj:-P-M0ez16Q-jp-zfja9r:b-2mw9az) ([raw file](https://github.com/paranext/paranext-core/blob/299a66dc149a4da09af875805a03aa69601b8f9b/extensions/src/platform-scripture-editor/src/use-open-find-shortcut.hook.ts#L24)):*

When no resource is resolved yet, this swallows Ctrl+F and does nothing — no panel, no message, not even a log line. PT‑4336's NN‑1D is precisely "always see feedback… there is no situation where a search term is entered but the result list stays or becomes empty" and the demo bullet is "Find never goes silent." A silent keystroke in a Bible-text tab that is still loading its resource is the same trust failure in a different place. Suggest either opening Find with the panel's own project as a fallback, or at minimum `logger.debug` + letting the event through, so the behavior is diagnosable.
</blockquote></details>

Took the second option. preventDefault() now only runs once Find is actually opening, and the no-op logs at debug with the web view id. I avoided the container-project fallback deliberately — searching the panel's own project instead of the resource on screen is the specific wrong answer this ticket exists to prevent. Covered by a test asserting defaultPrevented === false.
___
*[`src/renderer/components/docking/simple-layout.data.ts` line 97 at r1](https://reviewable.io/reviews/paranext/paranext-core/2677#-P-FbL17Dr20sJIERHn2:-P-M0oNjBNEbVnNf_PrN:b-miwlue) ([raw file](https://github.com/paranext/paranext-core/blob/299a66dc149a4da09af875805a03aa69601b8f9b/src/renderer/components/docking/simple-layout.data.ts#L97)):*

Claude also pointed out 

The file two entries above already imports `SCRIPTURE_EDITOR_WEBVIEW_TYPE` from `@shared/models/web-view.model` — and [web-view.model.ts:310-316](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/shared/models/web-view.model.ts#L310-L316) exists for exactly this purpose, with the same "core code cannot import extension source, so the value is mirrored here as the single core-side copy" rationale. The established pattern is a named, documented export in that model file, not an inline literal in layout data. Please add `FIND_WEBVIEW_TYPE` there and reference it here, so the next core-side consumer doesn't mint a 2nd copy.
</blockquote></details>

Not branched off Matt's — this is mine, so it's on me.
▎
▎ On what it renders: no error and no spinner. Every project-dependent read degrades to a default (booksPresent → default, IsStructureProtected → false, both PDPs → undefined), so you get the ordinary Find UI. But your instinct was right and "placeholder/empty state" was too generous a reading — the search box was enabled and inert. Typing a term and pressing Enter hit if (!isSearchQueryValid || !findPdp …) return; and did nothing, with no feedback at all. Squarely NN‑1D.
▎
▎ Fixed by making Find follow the active project instead of waiting for a Ctrl+F. Digging into why it was the odd one out turned up the real answer: the Scripture editor, model text panel, Text Collection, and comment list panel all force scrollGroupScrRef: 0 in Simple mode — only the Bible-texts/commentaries panels deliberately opt out, because they navigate independently. Find was the one Column 3 tab that neither followed that convention nor documented opting out. So the provider now forces group 0 in Simple mode like its siblings, and the web view resolves projectId ?? sourceProjectId (the 5th member of useWebViewScrollGroupScrRef, documented for exactly this: "a web view that must follow whichever project is driving the active Scripture reference"). That's verbatim what the Text Collection grid already does, for the same "default-layout tab carries no projectId" reason.
▎
▎ Worth flagging what that buys beyond the empty tab: it also makes Find's scroll group deterministic in Simple mode. Previously a panel-triggered Find inherited whatever group a prior editor-triggered Find happened to leave behind, so result-click navigation depended on session history. Now clicking a result always moves the Scripture editor — which is the only navigable scripture surface in Column 3 anyway, since the read-only panels register no controller to drive directly (PT‑4372).
▎
▎ An explicit sourceProjectId from a panel Ctrl+F still wins, so Find stays pinned to the resource you searched and only follows the editor when nothing has targeted it.
▎
▎ Startup timing: not measured. Simple mode goes from 5 to 6 always-mounted web views. Happy to capture PT_STARTUP_MARKS before merge if you want a number rather than a shrug — say the word.

Fixed. Also added the matching drift guard — web-view.model.test.ts already reads the extension source to pin SCRIPTURE_EDITOR_WEBVIEW_TYPE, and FIND_WEBVIEW_TYPE had no equivalent. That gap was worse than the usual mirror risk: a rename would leave the seeded layout tab pointing at a webViewType no provider serves, failing at startup with no type error and no failing test.
___
*[`src/stories/keyboard-shortcuts.data.ts` line 201 at r1](https://reviewable.io/reviews/paranext/paranext-core/2677#-P-Febuh6yedPmU03mzL:-P-M2GD6BaGeM_nERTZR:bkcrjsn) ([raw file](https://github.com/paranext/paranext-core/blob/299a66dc149a4da09af875805a03aa69601b8f9b/src/stories/keyboard-shortcuts.data.ts#L201)):*

`locations` gained the hook file but not the two web views that mount it (`model-text-panel.web-view.tsx`, `resource-text-panel.web-view.tsx`). The catalog rule asks for the handler locations; the mount sites are what a reader greps for.
</blockquote></details>

Fixed — hook first, then all four mount sites (editor, model text, resource text, Text Collection), with the context line noting that Text Collection searches the caret's resource and that Ctrl+F is a logged no-op until a scripture resolves.

jolierabideau (approved):
> @jolierabideau reviewed 28 files and all commit messages, and resolved 12 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2677)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

jolierabideau (approved):
> @jolierabideau reviewed 2 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2677)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2678, opened 2026-08-14
Add an agentic engineering standard and pin the Clanker Constitution

rolfheij-sil:
## Summary

Writes down how we build substantial changes with coding agents, and gives agents a baseline of
operating behaviour this repository never had. Documentation only — no code, no CI changes, no
plugin wiring.

Opened as a draft because it imports a new document into `CLAUDE.md` (so it loads in every agent
session here) and changes a team convention about where planning documents live.

Adapted from Wes McKinney's [How Kenn is doing Agentic Engineering](https://wesmckinney.com/blog/agentic-engineering-aug-2026/)
(August 2026), which describes a workflow built on Superpowers and roborev — the two tools this
repo is already adopting.

**Stacked on #2676.** The guide's prerequisites and its review-cadence step link to the roborev
setup section that #2676 adds to `Code-Review-Guide.md`, so this targets that branch rather than
`main`. Nothing here depends on roborev at runtime.

## What lands

| File | What it does |
| --- | --- |
| `CONSTITUTION.md` | Pinned verbatim copy of [kenn-io/constitution](https://github.com/kenn-io/constitution) v2026.08.11 (CC BY 4.0), imported by `CLAUDE.md`. 500 words of default agent behaviour: how to honor a request, when to ask versus proceed, how to protect existing work, how to report results. |
| `.context/standards/Agentic-Engineering-Guide.md` | The six-phase workflow — design together, write the spec, adversarial spec review, plan and implement in small pieces, close-out gate, make the work durable. |
| `CLAUDE.md` | Imports the constitution, records where we deviate from it, and replaces the plan-file rule. |
| `.claude/rules/docs-durability.md`, `where-to-add-guidance.md`, `REVIEW.md`, `verify-standards.md` | Reconciled with the two new documents. |
| `.github/workflows/constitution-pin-check.yml` | Monthly check that opens an issue when upstream publishes a newer constitution than the one pinned here. |
| `.context/plans/` | Deleted. |

## The constitution is the part worth arguing about

It is upstream text we do not edit, and it can conflict with how we already work. Rather than
leave an agent to resolve that at runtime, `CLAUDE.md` states the three deviations outright:

- **Ceremony.** § 2 says "scale process to the task"; the Superpowers `brainstorming` skill says
  its approval gate never scales down. § 2 wins here, and the guide defines the boundary: a typo
  needs no design, a new subsystem does.
- **`AGENTS.md`.** § 7 wants durable guidance there. We have no `AGENTS.md` and a deliberate
  three-tier structure with per-team ownership. `where-to-add-guidance.md` stays authoritative.
- **Agent-private memories.** § 7 says not to keep project guidance in them. Adopted going
  forward: anything a teammate needs belongs in the repo; personal working context can stay
  personal.

## The pin is watched, not trusted to stay current

Pinning is deliberate — the file loads into every agent session here, so it should change only
through review, and that is upstream's own guidance too. But a pin nobody watches goes stale
silently, and upstream is active: it published twice in two days, one of those a behavioural rule
requiring user authorization before merging a pull request.

So a monthly workflow compares the version on line 1 of `CONSTITUTION.md` against the newest
upstream tag and opens an issue when they differ — one issue per upstream version, linking the
compare view and reminding the reader to re-check the deviations above, since a bump can quietly
resolve one of them or invalidate it.

Both versions are constrained to the `vYYYY.MM.DD` shape before use, because they are
interpolated into a heredoc and an upstream tag containing shell metacharacters would otherwise
be an injection path. An unreadable pin or an empty tag list fails the job rather than reporting
a false all-clear. The comparison logic was run against the live API before landing: matching
versions report current, an older pin produces the expected issue, and a malformed pin fails loudly.

## Planning artifacts no longer live in the repo

Design docs, specs, and implementation plans are build-time scaffolding. Their durable content
belongs in `.context/standards/`, `.claude/rules/`, or doc comments; the planning document itself
does not land. The test the guide uses: delete the whole planning tree, and the repository still
explains itself.

This replaces the previous rule that plans be committed alongside their change. `.context/plans/`
held one task list from a PR that merged in July, referenced by no doc or config, and is removed.

`docs-durability.md` previously answered planning-doc rot by fencing those documents as frozen
records. Both rules now apply to what each covers: new planning documents stay out, and the
point-in-time documents that legitimately live here — investigation briefs written by
`/investigate-prd`, and grandfathered designs — still get fenced. Two `.context/designs/` files
were missing that fence; one of them opened by instructing agents to execute an already-shipped
plan, with live checkboxes. Both are fenced now.

`.context/designs/` itself is grandfathered and marked do-not-add. Nothing reads it at runtime —
all four PRD-tooling commands behave identically without it — but it holds the recorded rationale
for why that tooling is shaped the way it is. Converting that into a durable doc and deleting the
directory is a follow-up, not this PR.

## Nothing here is mandatory

The workflow is guidance; no CI check enforces it. roborev stays opt-in per developer and inert
without the binary. Superpowers stays a personal plugin install — this PR deliberately does not
enable or require it, since the team already uses it. The constitution is the one thing that
loads for everyone, and § 1 makes repository instructions override it anyway.

## Second opinion, honestly

The workflow's phase 3 calls for an adversarial spec review from a *different model family*, so
the reviewer does not share the implementer's blind spots. Only Claude is available to us, so our
version is a fresh Claude session — independent of the conversation, not of the model. The guide
says that plainly rather than implying we have the stronger form.

---

AI-assisted — [session 1](https://claude.ai/code/session_01FNfvLRp2C6xTLTKv5WgUwr)

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2678)


lyonsil (commented):
Generally speaking, I am skeptical of everything related to the constitution but think cleaning up the Superpowers docs would be helpful.

> @lyonsil made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2678#-:-P-PqJzJAzqdG_dnz5El:biwwt2l)* status: 0 of 13 files reviewed, 3 unresolved discussions (waiting on rolfheij-sil).

___
*[`CONSTITUTION.md` at r1](https://reviewable.io/reviews/paranext/paranext-core/2678#-P-PpXRR0noJzCFs_N2W:-P-PpXRR0noJzCFs_N2X:bvbltu7) ([raw file](https://github.com/paranext/paranext-core/blob/e9ee11b1fe7b0ca605c640737537b18d117c323f/CONSTITUTION.md)):*
What benefits have you seen from using this constitution vs not using it? Unless there is something specific you can point to, this feels a bit like noise. Who knows how much this duplicates or will adapt well to the internal tuning of prompts used by Anthropic and others in toos like Claude Code. And the more we fill the context at startup with nice sounding things that we don't know what they do, the less we really understand why AI does anything one way or another.
___
*[`CONSTITUTION.md` line 16 at r1](https://reviewable.io/reviews/paranext/paranext-core/2678#-P-Pqc754TSBpDvW2EUI:-P-Pqc754TSBpDvW2EUJ:bzenf42) ([raw file](https://github.com/paranext/paranext-core/blob/e9ee11b1fe7b0ca605c640737537b18d117c323f/CONSTITUTION.md#L16)):*
> ```Markdown
> - Proceed with safe, reversible, in-scope work without asking permission.
> - Ask only when a missing decision materially changes the result, required authority is absent, or an action is destructive, irreversible, or outside the requested scope.
> - Scale process to the task. Do not impose specification, planning, or approval ceremony on straightforward work.
> ```

The self-review pointed out this may contradict with Superpowers brainstorming. I've seen where brainstorming already realizes when steps are overkill and backs off. This goes to my high-level point of not really understanding what we get out of this constitution given our existing tools and processes.
___
*[`CONSTITUTION.md` line 47 at r1](https://reviewable.io/reviews/paranext/paranext-core/2678#-P-PpN_P3G_gfymBP3uZ:-P-PpN_P3G_gfymBP3u_:b-hdudu6) ([raw file](https://github.com/paranext/paranext-core/blob/e9ee11b1fe7b0ca605c640737537b18d117c323f/CONSTITUTION.md#L47)):*
> ```Markdown
> 
> ## 7. Learn in the right place
> - Put durable project guidance in `AGENTS.md`; have `CLAUDE.md` import or symlink it when both agents are used.
> ```

We don't do this now. It would seem odd to adopt a constitution if we aren't even following its guidance now or as part of the PR merging.

PR #2681, opened 2026-08-14
Hold layout pushes while a layout load is in flight

tjcouch-sil:
# Code Review Summary

**Branch**: fix-layout-clobbered-during-load

**Base**: origin/main

**Date**: 2026-08-14

**Files changed**: 2

## Overview

Switching Power → Simple → Power did not restore the Power layout: the dock came back showing Simple's three columns, two of them with no tab bar at all, and restarting did not help. The cause is that `saveLayout` was armed during the stretch in which the dock does not yet hold the layout it is supposed to be showing, so an ordinary layout change in that stretch persisted the *wrong* layout as this window's saved entry — permanently, since `windowLayout:get` answers with that entry from then on and the legacy fallback is never consulted again.

Two loads were exposed. An interface-mode switch flips `currentInterfaceMode` the instant the setting changes, but the new mode's layout does not reach the dock until `loadLayout` has read the mode, round-tripped to main for the saved layout (with retries, so seconds), and read the supplement flags — so going to Power, any layout change in that window persisted the static simple layout as the power layout. Separately, before the initial load in `registerDockLayout` lands, the dock holds rc-dock's empty default, and a push there replaced the saved entry with an empty layout.

The fix records which load's layout the dock actually holds (`layoutLoadGenerationInDock`, set at each of the three points `loadLayout` hands a layout to the dock) and has `saveLayout` hold its pushes while that marker trails `layoutLoadGeneration`.

## API Changes

None. The branch touches only `src/renderer/services/web-view.service-host.ts` and its test. No changes to `lib/platform-bible-react/`, `lib/platform-bible-utils/`, `lib/papi-dts/papi.d.ts`, or any `extensions/src/*/src/types/*.d.ts`. The one exported symbol in the changed file, `registerDockLayout`, has an unchanged signature; everything added is module-private.

None.

- [x] **A no-argument load that rejects released the hold without ever putting a layout in the dock, so pushes resumed against exactly the state the change exists to protect against.** Reachable, not theoretical: `settingsService.get('platform.interfaceMode')` throws when the settings data provider isn't up yet — precisely the startup window `registerDockLayout` runs in — and `getDockLayout()` is an `AsyncVariable` with a 10 s reject timeout. _(fixed during review: the original `finally`-released counter was replaced with a generation marker set only where a layout actually reaches the dock, so a throwing load never marks itself landed and the hold stays up until some load does. Pinned by the new test `a layout change after a load throws pushes nothing`.)_
- [x] **A superseded load kept holding pushes until its own request settled, even though the dock already held the newer load's layout.** `sendNetworkRequest` in `getPersistedLayout` has no timeout, so a request that never settles would have held pushes for the rest of the session — silently dropping real user layout changes while protecting nothing. This is the same hazard the file already guards for `isRunningOnFallbackLayout` (`if (isSuperseded()) return undefined;`, with a dedicated test at `web-view.service-host.test.ts:557`). _(fixed during review: the generation comparison lifts the hold as soon as any load reaches the dock, however long an abandoned one takes to settle. Pinned by the new test `an abandoned load must not hold pushes once a newer load has reached the dock`, which was verified to fail against the previous counter-based implementation.)_
- [x] **The invariant "EVERY no-argument load must go through here" was enforced only by a TSDoc sentence** — a future third `loadLayout()` call site would have silently reintroduced the permanent saved-layout clobber, with no test or lint to catch it. _(fixed during review: the wrapper was deleted and the marker moved inside `loadLayout` itself, at the points where the dock is written. There is no longer a bypassable path — the invariant is structural.)_
- [x] **The counter-vs-flag invariant was documented but untested.** _(fixed during review: the counter is gone; the generation semantics it was standing in for are now pinned by two tests, one of which was confirmed to fail against the counter design.)_

### Minor — Consider

- [x] **TSDoc claimed the one-argument `loadLayout` form is "deliberately not covered", which was not what the code did** — `saveLayout` checked the module-global counter unconditionally, so an explicit load running while a no-argument load was in flight would have had its deliberate persist silently held too. _(fixed during review: the one-argument branch now sets the marker itself, so it persists as documented.)_
- [x] **Comments narrated the defect in past tense** ("it persisted", "then restored", "is what destroyed"), against Code-Style-Guide → *Comments describe the code, not its history*; the surrounding comments in this file use the conditional voice for the same kind of rationale. _(fixed during review: recast to conditional — "a push would persist…", "would replace…".)_
- [x] **The same rationale was restated in five places** (the TSDoc, both call-site comments, both test comments). _(fixed during review: the wrapper and its duplicate comment are gone; one TSDoc carries the detail and the single remaining call-site comment is one line plus a pointer.)_
- [x] **The `holdGetLayout` mock was duplicated verbatim between the new tests.** _(fixed during review: extracted a `holdGetLayout()` helper alongside the file's existing `respondToGetLayout`/`registerWindow`/`makeDockLayout` convention; all three new tests use it. The pre-existing 4× duplication of the `settingsSubscribe` capture block was left alone to keep the diff scoped.)_
- [ ] ~~The new hold returns silently, unlike its sibling `isRunningOnFallbackLayout` guard which logs once per session so a "my layout isn't saving" report is diagnosable.~~ _(Dismissed: with the generation design the hold is now the normal path on every load, including the one at startup, so a once-per-session warning would fire in every session and carry no signal. The sibling guard logs because it marks a genuinely abnormal state. A future "held for an unexpectedly long time" warning would be the useful version of this, but it needs a duration threshold, not a first-hit latch.)_
- [ ] ~~`hasLoggedHeldLayoutPushes` now reads as if it covers all hold reasons but only gates the fallback warning; consider renaming.~~ _(Dismissed: pre-existing name, and since the new hold deliberately does not log (above), there is no second logged hold for it to be confused with. Renaming it here would be unrelated churn in a bugfix branch.)_
- [ ] ~~No blank line after the new early return in `saveLayout`, per Code-Style-Guide → *Code Patterns We Use*.~~ _(Dismissed: the two adjacent pre-existing guards in `saveLayout` are formatted identically; matching the immediate neighbors reads better than being the one guard that differs. The analyzer flagged this for completeness and noted the same.)_
- [ ] **A failed layout reload after a mode switch is invisible to the user** — the setting is written and `currentInterfaceMode` has moved on, but the dock keeps showing the old mode's layout with only a `logger.warn`. Applying Changes' Feedback table requires an error toast or inline message on failure. Pre-existing, and this branch edits that exact call site. _(Left open deliberately: adding a user-facing error surface to the mode switcher is a UX change needing design input, not something to slip into a data-integrity bugfix. Flagged for the reviewer to route.)_

#### Shared Regions Modified

None. Neither changed file contains a `#region shared with` marker.

#### Extension Config Changes

## Positive Observations

- The ordering in the interface-mode subscription is the crux of the fix and it is right: `currentInterfaceMode = newMode` and the generation bump happen in the same synchronous tick, so there is no window in which pushes are armed under the new mode while the dock still holds the old mode's layout.
- Release timing is sound: `PapiDockLayout.loadLayout` forwards straight to rc-dock's `loadLayout`, which does not fire `onLayoutChange`, so nothing can slip a push in between the dock load and the marker being set.
- All four new tests assert the *resume* side as well as the suppression, so a hold that never released would fail them — the failure mode a suppression-only test would miss.
- The tests drive the realistic shape of the race (holding the `windowLayout:get` request open the way a slow main process would) rather than reaching into module state.
- This is a UX-guideline repair, not just a data-integrity fix: Applying Changes exempts "a control that is its own undo — panel/layout toggle" from needing a separate recovery path, and that carve-out only holds if re-operating the control really does restore the prior state. It did not; this change is what makes it true.
- Well scoped — no public API surface touched, no unrelated refactoring bundled in.

## Interview Notes

This review was run autonomously at the user's request; the author role was answered from the investigation that produced the branch, not by a separate human.

Design rationale for the redesign made during the review: the two Important findings pointed at the same mechanism from opposite sides — the original counter released the hold too eagerly (on a throw) and held it too long (while a superseded load was still settling). Both are symptoms of counting *loads running* when the question is *whose layout is in the dock*. Comparing `layoutLoadGenerationInDock` against `layoutLoadGeneration` answers that question directly, which is why one change closes both, plus the "enforced only by a comment" concern (the marker lives inside `loadLayout` at the write sites, so no call site can bypass it) and the untested-invariant concern.

Deliberate tradeoff, flagged for the reviewer: a load that throws now leaves pushes held until some later load reaches the dock. That is unbounded in principle. It was chosen over releasing on error because the failure it prevents is permanent and silent (a destroyed saved layout) while the failure it risks is transient and self-healing (layout changes not persisted until the next mode switch or restart). The file already accepts this exact shape in `isRunningOnFallbackLayout`, which is likewise only cleared by a later successful load.

No areas were deferred without understanding; every finding was traced to the code before being accepted or dismissed. The four dismissals above each state their reasoning rather than being waved off.

## In-Review Quality Check

All checks run after the in-review changes:

- `npm run typecheck` — clean (0 TS errors)
- `npm run lint` — 0 errors repo-wide. The 4 remaining warnings are pre-existing in `extensions/src/platform-scripture-editor/` (`import/no-duplicates`, `no-console`), untouched by this branch.
- Prettier — required two `--write` passes on the changed files after comment edits; both files now pass `--check`.
- `npm test` — all workspaces green (141 + 16 + 131 + 29 + 8 + 1 + 12 + 1 + 5 + 11 + 54 test files); `web-view.service-host.test.ts` is 24/24.
- Falsification: all three "hold" tests fail with the guard disabled, and `an abandoned load must not hold pushes once a newer load has reached the dock` was run against the previous counter-based implementation and failed there with the predicted symptom (`expected [] to have a length of 1`).
- `dotnet test c-sharp-tests/` not run — no C# changed.

## Suggested Review Focus

- [ ] **The unbounded-hold tradeoff.** A load that throws holds pushes until a later load lands. Is the precedent (`isRunningOnFallbackLayout` behaves the same way) good enough, or does this want a bounded escape hatch?
- [ ] **Overlap with `improve_simple-power_switching`.** That branch reworks this same file and adds a `switchGeneration` counter for a related-but-distinct failure (a *superseded* switch's tail). Its `saveLayout` guard is content-based — it refuses layouts containing a `SIMPLE_LAYOUT_TAB_IDS` id — but those are the **unscoped** UUIDs from `simpleLayout`, while the layout coming out of the dock has been through `withWindowScopedWebViewIds` and carries a `-w<N>` suffix, so the `.has(id)` check looks like it would miss. Worth checking on that branch. Content sniffing also cannot cover the initial-load case (an empty layout contains no Simple tab ids). Decide which mechanism survives; expect conflicts either way.
- [ ] **Silent failure on a failed mode-switch reload** (open Minor finding above) — route to UX, or accept as out of scope for this branch.
- [ ] **Recovery for users already hit by this.** The fix prevents recurrence but does not restore a clobbered entry. Is a one-time migration warranted, or is "delete `window-layouts.json`" acceptable guidance?

---

## Summary

Switching Power → Simple → Power did not restore the Power layout: the dock came back showing Simple's three columns, with two tab groups missing their tab bars entirely. Restarting did not help, and toggling modes again did not help — because the user's saved Power layout had already been **destroyed on disk**.

`saveLayout` was armed during the window in which the dock does not yet hold the layout it is supposed to be showing, so an ordinary layout change in that window persisted the *wrong* layout as this window's saved entry. Two loads were exposed:

- **The interface-mode switch.** The subscription flips `currentInterfaceMode` the instant the setting changes, but the new mode's layout does not reach the dock until `loadLayout` has read the mode, round-tripped to main for the saved layout (with retries, so seconds), and read the supplement flags. Switching to **Power**, any layout change in that gap — a web view writing state (`setWebViewStateSync` → `updateWebViewDefinitionSync`), focus landing in a tab (`web-view.component.tsx:338`), a tab being activated — pushed the static simple layout as the window's power layout. Permanently: `windowLayout:get` answers with the clobbered entry from then on. The missing tab bars are the tell — `simple-layout.data.ts` panels carry `group: HEADLESS_GROUP`, and `dock-layout-wrapper.simple-mode.scss:56` hides `.dock-style-headless .dock-bar` by group class rather than by `[data-interface-mode='simple']`, so Simple's layout data renders headless in Power too.

  The direction matters: Power → Simple is harmless, since simple mode never pushes. Only Simple → Power destroys. The existing comment on `currentInterfaceMode` reasons about the first direction only.

- **The initial load** in `registerDockLayout`. Before it lands, the dock holds rc-dock's empty default rather than anything of this window's. Pushing that replaced the saved entry with an empty layout — and an entry that *has* a layout is not eligible for the legacy fallback either (`TrackedWindow.usesLegacyLayout`), so the window then started empty from then on. Narrower trigger than the first, same consequence.

### Why review this

Not part of the current epic. Found while manually exercising the mode switcher on a build off `main`: this silently and irreversibly destroys a user's saved Power layout, with no error and no recovery short of hand-editing `window-layouts.json`. Worth reviewer time now because it is live on `main` and every Simple → Power switch is a chance to hit it.

## Changes

- Added `layoutLoadsInFlight` and `loadLayoutHoldingPushes`; every no-argument `loadLayout` now routes through it, and `saveLayout` holds its pushes while it is above zero. A counter (not a flag) so overlapping loads keep the hold up until both finish, released in a `finally` so a load that throws cannot latch it on for the session. The one-argument form is deliberately not covered — that caller hands us the layout it wants and `loadLayout` persists it on purpose.
- Two tests, both verified to fail without the guard.

Nothing is lost by holding: the load on its way replaces the dock's whole contents anyway.

## Heads-up: overlaps `improve_simple-power_switching`

That branch reworks this same file and adds a `switchGeneration` counter for a related-but-distinct failure (a *superseded* switch's tail persisting after `currentInterfaceMode` moved on). This PR is about the *current, non-superseded* switch's own load window. Two things for whoever reconciles them:

- That branch's `saveLayout` guard is content-based — it refuses layouts containing a `SIMPLE_LAYOUT_TAB_IDS` id. Those ids are the **unscoped** UUIDs from `simpleLayout`, but the layout coming out of the dock has been through `withWindowScopedWebViewIds`, so its ids carry the `-w<N>` suffix and the `.has(id)` check looks like it would miss. Worth checking on that branch.
- Content sniffing also cannot cover the initial-load case here, since an empty layout contains no Simple tab ids.

Expect conflicts; happy to rebase onto whichever lands first.

## AI Involvement

AI-assisted (Claude Opus 5 via Claude Code). Claude did the investigation — traced the clobber to the `currentInterfaceMode` flip preceding the load, and confirmed it with a red test before writing any fix — then wrote the guard and both tests. I reviewed the diff and the reasoning.

## Testing

- [x] `npm test` — all workspaces green (`web-view.service-host.test.ts`: 22/22, including the pre-existing supersede/race tests)
- [x] `npm run typecheck` clean
- [x] `npm run lint` clean (repo-wide)
- [x] Revert test: both new tests fail with the guard disabled, pass with it
- [ ] No C# changed, so `dotnet test c-sharp-tests/` not run
- [ ] Not manually re-verified in the app — the destructive path needs an already-clobbered profile to observe

## Risk Level

Low — one added early-return in `saveLayout`, gated on a counter that is only non-zero for the duration of a no-argument `loadLayout`. It can only *suppress* pushes, never redirect one, and the `finally` bounds how long it can suppress.

## Note for anyone already hit by this

Your saved Power layout is gone; this fix prevents recurrence but does not restore it. If `dock-saved-layout` is still in `localStorage`, quitting and deleting `window-layouts.json` from the userData folder makes the window fall back to it on the next Power-mode start (at the cost of saved window bounds and any secondary-window entries).

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2681)

### DIFF
--- a/src/renderer/services/web-view.service-host.ts
+++ b/src/renderer/services/web-view.service-host.ts
@@ -888,6 +888,12 @@ let layoutLoadGeneration = 0;
  * landed, so it cannot lift the hold on a dock it never wrote; and it cannot extend the hold
  * either, since the load that DOES reach the dock sets the marker to the current generation and
  * pushes resume immediately, however long the abandoned one takes to settle.
+ *
+ * That holds only because EVERY site that writes the dock checks `isSuperseded()` first, so this
+ * can only ever move to the generation whose layout is actually in the dock — never backwards to a
+ * load a newer one already replaced. A write site without that check would regress the marker below
+ * `layoutLoadGeneration` and hold every push from then on, with nothing scheduled to reconcile the
+ * dock. Keep the checkpoint if you add another write site.
  */
 let layoutLoadGenerationInDock = 0;
 
@@ -908,12 +914,21 @@ async function loadLayout(layout?: LayoutInfo): Promise<void> {
   const webViewsBeforeLoad = dockLayoutVar.getAllWebViewDefinitions();
   if (layout) {
     // Explicit layout change. `loadLayout` doesn't run `onLayoutChange`, so run it manually.
-    // Applied unconditionally: the caller handed us the layout, so there is nothing here that a
-    // later load could make stale — and bumping the generation above is what lets this call cancel
-    // an in-flight no-argument load that would otherwise land on top of it.
+    // Bumping the generation above is what lets this call cancel an in-flight no-argument load that
+    // would otherwise land on top of it — when THIS is the newer load, the checkpoint below passes
+    // and the older one drops its answer instead.
     // NOTE: we intentionally do NOT apply the default-layout supplement here — a caller passing an
     // explicit layout owns its full contents. If a future "reset to default layout" path routes
     // through here and should include supplement tabs, merge `getEnabledSupplementEntries()` in too.
+    if (isSuperseded()) {
+      // The reverse ordering: a newer load reached the dock while this one was still awaiting it.
+      // Writing now would replace the dock with content the user has already moved past AND regress
+      // `layoutLoadGenerationInDock` below `layoutLoadGeneration`, which holds every subsequent
+      // push with nothing scheduled to reconcile the dock. The caller owning this layout does not
+      // make it current; only being the newest load does.
+      logger.debug('Dropping an explicit layout load that a newer one superseded');
+      return;
+    }
     dockLayoutVar.loadLayout(layout);
     layoutLoadGenerationInDock = thisGeneration;
     emitCloseEventsForWebViewsRemovedByLayoutLoad(webViewsBeforeLoad, layout);
--- a/src/renderer/services/web-view.service-host.ts
+++ b/src/renderer/services/web-view.service-host.ts
@@ -631,6 +631,37 @@ function getWindowIdOrThrow(): number {
  */
 let currentInterfaceMode: 'simple' | 'power' | undefined;
 
+/**
+ * How many no-argument {@link loadLayout} calls have started but not yet put their layout in the
+ * dock. While this is above zero, {@link saveLayout} holds its pushes: the dock does not yet hold
+ * the layout this window is supposed to be showing, so pushing what it DOES hold overwrites the
+ * user's real saved entry in the main process's structure.
+ *
+ * Such a load takes a while — it reads the interface mode, asks the main process for this window's
+ * saved layout (a round trip that retries, so seconds), and reads the supplement flags — and every
+ * ordinary layout change runs through `onLayoutChange` into `saveLayout` meanwhile: a web view
+ * writing state, focus moving into a tab, a tab being activated. Two loads are exposed:
+ *
+ * - The INTERFACE-MODE SWITCH, which flips {@link currentInterfaceMode} the instant the setting
+ *   changes (so a `saveLayout` racing the switch can't push under the mode the user just left)
+ *   while the dock still holds the OLD mode's layout. Switching TO simple that is harmless, since
+ *   simple mode never pushes. Switching to POWER it persisted the static simple layout as this
+ *   window's power layout — permanently. Every later power-mode load then restored the simple
+ *   layout (headless columns and all, so two of its tab groups render with no tab bar in power
+ *   mode), and neither switching back and forth nor restarting recovered it, because the clobbered
+ *   entry is what the main process answers with from then on.
+ * - The INITIAL LOAD in `registerDockLayout`, before which the dock holds rc-dock's empty default
+ *   layout rather than anything of this window's. Pushing that replaces the saved entry with an
+ *   empty one — and an entry that HAS a layout is not eligible for the legacy fallback either (see
+ *   `TrackedWindow.usesLegacyLayout` in `window-layout-persistence.service.ts`), so the window then
+ *   starts empty from then on.
+ *
+ * A counter rather than a flag so a load starting before an earlier one finishes keeps the hold up
+ * until BOTH are done, and released in a `finally` (see {@link loadLayoutHoldingPushes}) so a load
+ * that throws can never latch it on for the rest of the session.
+ */
+let layoutLoadsInFlight = 0;
+
 /** Create a new dock layout promise variable */
 function createDockLayoutAsyncVar(): AsyncVariable<PapiDockLayout> {
   return new AsyncVariable<PapiDockLayout>('web-view.service-host.platformDockLayout');
@@ -962,6 +993,21 @@ async function loadLayout(layout?: LayoutInfo): Promise<void> {
 }
 
 /**
+ * Runs a no-argument {@link loadLayout} with layout pushes held until the layout it loads is in the
+ * dock. EVERY no-argument load must go through here — see {@link layoutLoadsInFlight} for what a
+ * push during one of them destroys. (The one-argument form is deliberately not covered: that caller
+ * hands us the layout it wants and `loadLayout` persists it on purpose.)
+ */
+async function loadLayoutHoldingPushes(): Promise<void> {
+  layoutLoadsInFlight += 1;
+  try {
+    await loadLayout();
+  } finally {
+    layoutLoadsInFlight -= 1;
+  }
+}
+
+/**
  * The pre-multi-window saved layout, read from raw `localStorage` (not `localWindowStorage`, whose
  * migration path would clone the legacy layout into every window). Builds that scoped
  * `localStorage` per window wrote the dock layout only under prefixed keys
@@ -1090,6 +1136,11 @@ async function saveLayout(layout: LayoutInfo): Promise<void> {
   const interfaceMode =
     currentInterfaceMode ?? (await settingsService.get('platform.interfaceMode'));
   if (interfaceMode === 'simple') return;
+  // A load is still fetching the layout this window should be showing, so what the dock holds is
+  // something else — the layout of the mode the user just left, or rc-dock's empty default at
+  // startup. Persisting that overwrites the real saved entry; see `layoutLoadsInFlight`. Nothing is
+  // lost by holding: the load on its way replaces the dock's whole contents anyway.
+  if (layoutLoadsInFlight > 0) return;
   if (isRunningOnFallbackLayout) {
     // The dock holds a fallback, not the user's saved layout (see getPersistedLayout); pushing it
     // would replace the real saved entry in the main process's structure
@@ -1134,7 +1185,9 @@ export function registerDockLayout(dockLayout: PapiDockLayout): Unsubscriber {
   // Fire-and-forget so `registerDockLayout` can stay sync. Guard the rejection, though: `loadLayout`
   // awaits settings reads (supplement flags), and an unhandled rejection here would leave the dock
   // unloaded (blank window).
-  loadLayout().catch((err) => logger.warn(`Initial loadLayout failed: ${getErrorMessage(err)}`));
+  loadLayoutHoldingPushes().catch((err) =>
+    logger.warn(`Initial loadLayout failed: ${getErrorMessage(err)}`),
+  );
 
   // Reload the layout whenever `platform.interfaceMode` changes so the user-facing mode switcher
   // (see `UserProfilePopover`) can swap layouts live without a restart. Use
@@ -1169,7 +1222,11 @@ export function registerDockLayout(dockLayout: PapiDockLayout): Unsubscriber {
           // switch reads the new mode immediately (before `loadLayout`'s own read resolves).
           currentInterfaceMode = newMode;
           try {
-            await loadLayout();
+            // Holds layout pushes until the new mode's layout is actually in the dock. Until then
+            // the dock still holds the old mode's layout, and pushing that under the new mode is
+            // what destroyed saved power layouts on a simple->power switch — see
+            // `layoutLoadsInFlight`.
+            await loadLayoutHoldingPushes();
           } catch (err) {
             logger.warn(`Dock layout failed to reload after interface mode change: ${err}`);
           }
--- a/src/renderer/services/web-view.service-host.ts
+++ b/src/renderer/services/web-view.service-host.ts
@@ -631,37 +631,6 @@ function getWindowIdOrThrow(): number {
  */
 let currentInterfaceMode: 'simple' | 'power' | undefined;
 
-/**
- * How many no-argument {@link loadLayout} calls have started but not yet put their layout in the
- * dock. While this is above zero, {@link saveLayout} holds its pushes: the dock does not yet hold
- * the layout this window is supposed to be showing, so pushing what it DOES hold overwrites the
- * user's real saved entry in the main process's structure.
- *
- * Such a load takes a while — it reads the interface mode, asks the main process for this window's
- * saved layout (a round trip that retries, so seconds), and reads the supplement flags — and every
- * ordinary layout change runs through `onLayoutChange` into `saveLayout` meanwhile: a web view
- * writing state, focus moving into a tab, a tab being activated. Two loads are exposed:
- *
- * - The INTERFACE-MODE SWITCH, which flips {@link currentInterfaceMode} the instant the setting
- *   changes (so a `saveLayout` racing the switch can't push under the mode the user just left)
- *   while the dock still holds the OLD mode's layout. Switching TO simple that is harmless, since
- *   simple mode never pushes. Switching to POWER it persisted the static simple layout as this
- *   window's power layout — permanently. Every later power-mode load then restored the simple
- *   layout (headless columns and all, so two of its tab groups render with no tab bar in power
- *   mode), and neither switching back and forth nor restarting recovered it, because the clobbered
- *   entry is what the main process answers with from then on.
- * - The INITIAL LOAD in `registerDockLayout`, before which the dock holds rc-dock's empty default
- *   layout rather than anything of this window's. Pushing that replaces the saved entry with an
- *   empty one — and an entry that HAS a layout is not eligible for the legacy fallback either (see
- *   `TrackedWindow.usesLegacyLayout` in `window-layout-persistence.service.ts`), so the window then
- *   starts empty from then on.
- *
- * A counter rather than a flag so a load starting before an earlier one finishes keeps the hold up
- * until BOTH are done, and released in a `finally` (see {@link loadLayoutHoldingPushes}) so a load
- * that throws can never latch it on for the rest of the session.
- */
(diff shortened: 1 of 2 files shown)


tombogle:
Not superseded — confirmed with Claude. This fixes a different bug than #2425 covers, and #2425's own `saveLayout` guard only partially overlaps with what this one protects against.

#2425's guard is content-based: it refuses to persist a layout that still contains a `SIMPLE_LAYOUT_TAB_IDS` id. That catches the common case (Simple entered via its fast path, where ids stay unscoped), but:
- It misses when Simple was entered via its own fallback path (`loadLayoutWithWarning` → bare `loadLayout()`), which *does* run the ids through `withWindowScopedWebViewIds` — so the dock holds `${id}-w2`-style scoped ids that the guard's unscoped `.has(id)` check won't match. Reachable via a real path (cold-start recents timeout, or any thrown error inside `runProjectBoundSimpleSwitch`), not just hypothetical.
- It never catches the empty-dock/initial-load case at all — an empty layout contains no Simple tab ids by definition.

This PR's `layoutLoadGenerationInDock` approach is structural (tracks whose layout the dock actually holds, not what's in it) and covers both gaps. Once this lands, #2425's guard is a reasonable thing to retire in favor of it rather than keep both, since this one is a strict superset — but that's a call for whoever reconciles the two branches, not something to resolve here.

Both PRs touch `saveLayout`/`loadLayout`/`registerDockLayout` in the same file, so expect a real conflict when integrating — just not because either replaces the other.

src/renderer/services/web-view.service-host.ts
```diff
@@ -881,6 +915,7 @@ async function loadLayout(layout?: LayoutInfo): Promise<void> {
     // explicit layout owns its full contents. If a future "reset to default layout" path routes
     // through here and should include supplement tabs, merge `getEnabledSupplementEntries()` in too.
     dockLayoutVar.loadLayout(layout);
+    layoutLoadGenerationInDock = thisGeneration;
```
tombogle, line 933: `loadLayout`'s explicit-layout branch (`web-view.service-host.ts:918`, PR #2681) writes to the dock and sets `layoutLoadGenerationInDock = thisGeneration` **unconditionally**. Unlike every other write site in the function, it never checks `isSuperseded()`.

I checked the actual code at the PR's head commit (`4d25e0557f0`):

```ts
if (layout) {
  // Applied unconditionally: the caller handed us the layout, so there is nothing here that a
  // later load could make stale — and bumping the generation above is what lets this call cancel
  // an in-flight no-argument load that would otherwise land on top of it.
  dockLayoutVar.loadLayout(layout);
  layoutLoadGenerationInDock = thisGeneration;
  ...
}
```

The problem is that an **older explicit-layout call (generation N)** can resolve after a **newer no-arg load (N+1)** has already written the dock. This branch would then overwrite the dock with N's stale content and regress `layoutLoadGenerationInDock` back to N, below the current `layoutLoadGeneration`.

At that point, `saveLayout`'s new guard correctly prevents the destructive push, but the dock is left silently showing stale content with nothing scheduled to reconcile it.

I also verified reachability. `loadLayout` isn't exported, and nothing in #2681's branch (based on plain `main`) currently calls it with an explicit layout, so this is genuinely dead code **today**. However, it's exactly where it stops being dead: our branch's `runProjectBoundSimpleSwitch` calls `loadLayout(layoutToLoad, { persist: false })`, and the PR's own doc comment for `layoutLoadGenerationInDock` anticipates a future explicit-layout caller ("a future 'reset to default layout' path").

So I don't think this is a merge blocker by itself given that the branch is currently unreachable. But the invariant claimed by the doc comment — that the generation "keeps the hold matched to reality in both directions" — doesn't actually hold for this branch. It seems worth surfacing now while the mechanism is fresh, since this becomes a real bug as soon as either the anticipated future caller is added or this reconciles with #2425 .

tjcouch-sil: Good catch — you're right on both the mechanism and the reachability, and I've fixed it in cc5c7ab.

Confirming the analysis independently: the explicit branch was the only dock-write site without an `isSuperseded()` checkpoint. The two no-arg write sites are both covered by the shared "point of no return" checkpoint above them; that branch had nothing. So the marker could regress below `layoutLoadGeneration`, and as you say `saveLayout` then correctly refuses the destructive push but holds *every* push from that point on, with nothing scheduled to reconcile the dock. That second-order effect is the worse half — a silent, permanent stop to layout persistence is the same class of failure this PR exists to fix, just arriving by a different route.

The fix is the checkpoint you'd expect:

```ts
if (isSuperseded()) {
  logger.debug('Dropping an explicit layout load that a newer one superseded');
  return;
}
dockLayoutVar.loadLayout(layout);
layoutLoadGenerationInDock = thisGeneration;
```

Worth noting explicitly since it reads like a behavior change: this **preserves** the property the original "Applied unconditionally" comment was actually describing. That comment justified itself with "bumping the generation above is what lets this call cancel an in-flight no-argument load" — which is the case where the explicit load is the **newer** one. There, `isSuperseded()` is false, the write proceeds, and the older no-arg load drops its own answer at its checkpoint, exactly as before. The only ordering that changes behavior is the one the comment didn't consider: the explicit load being the **older** one. I rewrote the comment accordingly rather than leaving a now-false claim next to the new check.

I also corrected the `layoutLoadGenerationInDock` doc comment. You're right that it overclaimed — "keeps the hold matched to reality in both directions" was stated as if it fell out of the generation comparison itself, when it actually depends on every write site checking supersession first. It now says so, and says what breaks if a future write site skips it.

**On test coverage** — I did not add one, and I want to be upfront about why rather than have it look like an oversight. The branch is unreachable from outside the module: `loadLayout` isn't exported, and I checked every export in the file for a path that reaches it with an explicit layout — there is none. Driving it from a unit test would mean exporting a branch with no production caller purely so a test can reach it, which widens the module's surface for dead code. Whoever adds the first explicit-layout caller — most likely #2425's `runProjectBoundSimpleSwitch(layoutToLoad, { persist: false })` — should land coverage with it, at which point it's reachable through a real path and the test is worth having. Flagging it here so that doesn't get lost in the reconciliation.

One consequence worth your eye during that reconciliation: with this checkpoint in place, a `runProjectBoundSimpleSwitch` load that gets superseded will now be **dropped** rather than applied. I believe that's correct — applying an older project-bound Simple layout over a newer load's dock would show stale content — but it's your branch's call, so calling it out rather than assuming.

Thanks for digging into the actual head commit rather than the description; that's what made this specific.

tjcouch-sil:
Thanks for the supersession check on #2425 — and for confirming the scoped-id gap independently. That matches what I found looking at that branch: `SIMPLE_LAYOUT_TAB_IDS` is built from the unscoped UUIDs in `simpleLayout`, while anything coming out of the dock has been through `withWindowScopedWebViewIds`, so the `.has(id)` check can't match. Useful to know the fallback path (`loadLayoutWithWarning` → bare `loadLayout()`) makes that reachable for real rather than only in theory — I'd only reasoned it from the id shapes.

Agreed on all three of your points: different bugs, real conflict coming, and the reconciliation call belongs to whoever integrates rather than here. No objection to retiring #2425's content-based guard in favor of this one if that's the direction, though I'd suggest doing it as a deliberate step with #2425's tests still green rather than as part of conflict resolution — a content guard failing open looks identical to a structural guard working, so it's easy to lose one silently.

Latest push (cc5c7ab) addresses your inline finding on the explicit-layout branch; details in that thread. Net diff is +231/-3 across the service and its test.

tombogle (approved):
Not going to claim I fully understand all the details, but conceptually it looks right.

> @tombogle reviewed 2 files and all commit messages, made 2 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2681#-:-P-IJz6x0pW31ZJiTudX:b-g8snvp)* status: all files reviewed, 1 unresolved discussion (waiting on tjcouch-sil).

___
*[`src/renderer/services/web-view.service-host.test.ts` line 704 at r2](https://reviewable.io/reviews/paranext/paranext-core/2681#-P-II83r8yGyigEcKP6I:-P-II83r8yGyigEcKP6J:bc7zdqo) ([raw file](https://github.com/paranext/paranext-core/blob/cc5c7ab15f9e9ef680b4b04c953ba23535a5891e/src/renderer/services/web-view.service-host.test.ts#L704)):*
> ```XML
>     // Once the saved layout lands, pushes resume from there
>     heldGet.answerWith({ kind: 'entry', layout: layoutWithTab('saved-tab') });
>     await vi.waitFor(() => expect(loadedLayouts.length).toBeGreaterThan(0));
> ```

Isn't this exactly 1?

PR #4, opened 2023-01-24
Fix release

irahopkinson:
- fix publish workflow
- improve release instructions
- also reduce tests to only pull requests or pushes to main

PR #2682, opened 2026-08-14
PT-4339 Scripture editor tracks editable/readonly state

tombogle:
# Code Review Summary

**Branch**: pt-4339-scripture-editor-readonly-state

**Base**: origin/main

**Date**: 2026-08-17

**Files changed**: 8 (`package-lock.json`'s uncommitted diff is worktree-directory-name churn from `npm install`, not project content — excluded from this count and from analysis)

## Overview

Implements PT-4339: the Scripture Editor webview now determines its own read-only state reactively, instead of receiving a non-reactive value computed once at open time. Rather than trying to precompute a single value that covers every condition that could make the text uneditable — and risk that value going stale — the editor subscribes to the two signals that actually gate whether an edit would be accepted (the project's `platform.isEditable` setting and the user's Scripture-edit role permission) and combines them live, so its editability determination is intended to track how the backend would actually accept or reject a change.

The change spans four coordinated layers: a new reactive C# project data type (`GetCanUserEditScripture`/`getCanUserEditScripture`, delegating to the existing one-shot `CanUserEditScripture()`, with no setter); a TypeScript data-type declaration (`CanUserEditScripture: DataProviderDataType<undefined, boolean, never>`) exposing it via `useProjectData`; a widened `formatEditorTitle` signature (`isReadOnly: boolean | undefined`) that renders the title's `{editable}` placeholder blank while resolution is still in flight rather than guessing; and the webview wiring that folds two new reactive signals into the existing `isReadOnlyEffective` memo and threads a loading-aware value into the title effect. Both new locale strings (`en`/`es`) for the new `(Read-only)` title indicator were added.

## API Changes

- `extensions/src/platform-scripture/src/types/platform-scripture.d.ts`: Added `CanUserEditScripture: DataProviderDataType<undefined, boolean, never>` to `ScriptureEditPermissionsProjectInterfaceDataTypes` (previously an empty type) — a new read-only reactive data type exposing live Scripture-edit-permission state via `getCanUserEditScripture`/`subscribeCanUserEditScripture`.
- `extensions/src/platform-scripture/src/types/platform-scripture.d.ts`: `IScriptureEditPermissionsProjectDataProvider.canUserEditScripture()` — JSDoc updated to clarify it is a one-shot, non-reactive check and to point callers toward the new `CanUserEditScripture` data type for reactive use; method signature unchanged.

No changes to `lib/platform-bible-react/`, `lib/platform-bible-utils/`, or `papi.d.ts`.

None.

- [ ] **The two new reactive signals in `platform-scripture-editor.web-view.tsx:421-457` (`isProjectEditable`, `canUserEditScripture` — each: hook call → `isPlatformError` check → `logger.warn` → fail-open `return true`) have zero unit test coverage, despite directly gating whether the editor is writable. The codebase has a precedent for testing this exact shape (`use-is-power-mode.hook.ts`/`.test.ts`, `use-structure-protection-state.hook.ts`).**

[Author response: the author is not convinced full hook extraction is worth it for this scope, but is open to a narrower unit test covering just the `isPlatformError`/fail-open branches (which don't require a running app, unlike the full reactive behavior). Explicitly deferring the priority/scope call to the human reviewer — this is a scoping question, not a gap in the author's understanding of the code.]

### Minor — Consider

- [ ] ~~The Spanish value for `%webView_platformScriptureEditor_title_editable_indicator%` changed from `"(Editable)"` to `"(editable)"` (capitalization only, to match the lowercase style of the new `"(solo lectura)"` string).~~ _(Verified by the style analysis pass as a same-meaning capitalization fix, not a behavior change — correct to edit the existing key rather than add a new one. Author confirmed: fine, no discussion needed.)_

#### Shared Regions Modified

None.

#### Extension Config Changes

None.

## Positive Observations

- The API addition is purely additive and non-breaking: the existing one-shot `canUserEditScripture()` method's signature and return type are unchanged; only its JSDoc was clarified to disambiguate it from the new reactive alternative.
- `CanUserEditScripture: DataProviderDataType<undefined, boolean, never>` exactly matches the project's established convention for read-only reactive boolean data types (e.g. `IsStructureProtected` at `platform-scripture.d.ts:711`), and the C# side registers only the getter (`getCanUserEditScripture`), mirroring `GetMarkerNames`/`getMarkerNames` — no setter, consistent with the "role isn't locally settable" constraint.
- The C# implementation delegates to the existing `CanUserEditScripture()` rather than duplicating logic, and reuses the existing wildcard `SendFullProjectUpdateEvent` for refresh instead of adding new event-wiring.
- C# test coverage is thorough: positive (`GetCanUserEditScripture_NormalProject_ReturnsTrue`), negative (`_ObserverRole_ReturnsFalse`), and a same-construction-path **positive control** (`_NonObserverRole_ReturnsTrue`) specifically to rule out a false negative from a broken test fixture rather than a real permission check — plus a test that asserts the actual wildcard (`"*"`) event payload, not just that some event fired.
- `formatEditorTitle`'s new tri-state (`true`/`false`/`undefined`) behavior is covered by three new tests, including the "still resolving" blank-placeholder case — the one most likely to be gotten wrong — and the one deliberately untreated edge case (the no-project fallback branch, where `undefined` isn't given tri-state handling) is documented inline with the reasoning for why that's currently safe.
- The new webview signals reuse the file's existing `useProjectSetting`/`useProjectData` + `isPlatformError` fail-open idiom rather than inventing a new data-fetching pattern, and the fail-open (live signals) vs. fail-closed (one-shot getter / title default) distinction is documented with a clear rationale at each site.
- New localization keys follow the existing naming convention and both locales (`en`, `es`) were updated together — no missing-locale gap; the new indicator's sentence-case and parenthetical style match the pre-existing `"(Editable)"` convention.
- No hardcoded UI strings were introduced; the new title logic routes entirely through the existing localization system.

## Interview Notes

**Stated purpose**: Implement PT-4339 — the Scripture Editor webview should determine its own read-only state.

**Design explanation** (most complex/non-obvious part, in the author's words): rather than passing a non-reactive, potentially stale/inaccurate value into the editor to try to cover all the conditions that might make the text uneditable, the editor determines editability itself, in a way that is both reactive and intended to correspond with how the backend validator would actually accept or reject a given change. This matches the design spec's stated problem precisely and demonstrates clear understanding of the architecture — no deferral to AI, no expressed uncertainty about how the mechanism works.

**Important finding response**: see Findings section above — the author is weighing whether full test coverage (especially hook extraction) is worth the effort for this scope, is open to a narrower fail-open/fail-closed unit test, and is deferring the final call to the reviewer. Not an unresolved-understanding issue — a scoping judgment call.

**Minor finding response**: confirmed fine, no discussion needed.

**Additional context**: none volunteered.

## Suggested Review Focus

- [ ] Test coverage for `isProjectEditable`/`canUserEditScripture`'s fail-open error handling (`platform-scripture-editor.web-view.tsx:421-457`) — is a narrower unit test (just the `isPlatformError` branches) worth adding now, or is this acceptable to ship without, given the codebase has at least one other untested instance of the same pattern (`isPowerMode`, pre-existing, unrelated to this PR)? Author has a preference (narrow test over full hook extraction) but wants the reviewer's input on priority.

---

This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2682)


tombogle:
https://github.com/user-attachments/assets/f74070cc-f12f-432c-92dc-0f6dabe1fc39

In this video, you can see the `(editable)`/`(solo lectura)` change in the title bar of the editor as the Is Editable setting is toggled.

tombogle:
https://github.com/user-attachments/assets/f5d822cf-2e7e-497f-8b59-ce28728fa689

In this video, you can see the title bar responding to switching between markers view (not editable) and the standard/formatted view, which is editable. (This is not new functionality, but this demonstrates that it was not broken. The display of `(solo lectura)` is new.)

tombogle:
https://github.com/user-attachments/assets/93a77697-d914-4d52-8376-d44aa814a1e6

The reactive refresh mechanism for the role-change-via-send/receive scenario depends on Paratext 10 Studio's sync wrapper (not something this repo can trigger). Although it was not possible to test or illustrate that scenario, I confirmed that role-based read-only state is correct on open after simulating the change by making the change directly in `.platform.bible\projects\Paratext 9 Projects\SPAN\ProjectUserAccess.xml` and restarting Platform.
Live reactivity to a role change arriving mid-session via sync is untested here and should be verified post-merge with real Send/Receive") — noted as Testing Idea in Jira.

captaincrazybro (commented):
Reviewed the four coordinated layers plus the `useProjectData` / `useProjectSetting` / `createUseDataHook` / `usePromise` chain they rest on. The design is sound and the C# side is clean and well-tested — the findings below are all on the TypeScript side, and they cluster around one theme: **`isReadOnlyEffective` was extended, but not every write path in the file moved onto it, and the two new `isLoading` flags are trusted as "resolution" signals when they can never clear.**

5 inline comments, roughly in order of significance. #1 and #2 are the ones I'd treat as blocking (both small diffs, both cases where the stated goal isn't yet met by the code). #3 is the subtle one and it's load-bearing — it's a latent latch that today only costs title updates, but it will bite whoever fixes #4 the obvious way.

extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx
```diff
@@ -515,17 +555,28 @@ globalThis.webViewComponent = function PlatformScriptureEditor({
...
     () =>
       isReadOnly ||
-      // An automatic Send/Receive is syncing this project — freeze editing until it finishes.
+      !isProjectEditable ||
+      !canUserEditScripture ||
       isSyncBlocked ||
       (viewType === 'markers' && localStorage.getItem('dev-editableMarkersView') !== 'true'),
-    [isReadOnly, isSyncBlocked, viewType],
+    [isReadOnly, isProjectEditable, canUserEditScripture, isSyncBlocked, viewType],
+  );
```
captaincrazybro, line 646: **#1 — `isReadOnlyEffective` was extended, but not every write path was moved onto it**

Anchoring here because the two affected sites fall outside this PR's diff hunks, so GitHub won't let me comment on them directly:

- [`nodeOptions.noteCallerOnClick` (L536-L537)](https://github.com/paranext/paranext-core/blob/129da767bb9056bcfddbd2d2f10ba7b61b6ae4e7/extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx#L536-L537) still gates on `isReadOnly || isSyncBlocked`
- [`onUsjChange` (L1979)](https://github.com/paranext/paranext-core/blob/129da767bb9056bcfddbd2d2f10ba7b61b6ae4e7/extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx#L1979) still gates on the raw `isReadOnly`

On a project whose metadata says editable (`isReadOnly === false`) where the user's role is **Observer**, `canUserEditScripture` is `false` and the text body is correctly locked via `options.isReadonly` — but the user can still click a note caller, open the footnote editor, and attempt a note write the backend will reject. The comment directly above `noteCallerOnClick` already says note editing "is a project write that must be frozen", so the intent is there; it just predates these signals.

`onUsjChange` is lower risk (a read-only Lexical instance shouldn't emit editorial changes), but it is the actual save path, and the raw flag shouldn't be the thing standing between an Observer and a save attempt.

**Suggested fix:** both sites use `isReadOnlyEffective`. It already contains `isSyncBlocked`, so `nodeOptions`' dependency array shrinks rather than grows.

extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx
```diff
@@ -515,17 +555,28 @@ globalThis.webViewComponent = function PlatformScriptureEditor({
...
+
+  // `undefined` means isProjectEditable/canUserEditScripture are still resolving — formatEditorTitle
+  // renders the `{editable}` placeholder blank rather than committing to a possibly-wrong label
+  // until a real value is available. See formatEditorTitle's isReadOnly parameter.
+  const isReadOnlyForTitle = useMemo(
+    () =>
+      isProjectEditableLoading || canUserEditScriptureLoading ? undefined : isReadOnlyEffective,
+    [isProjectEditableLoading, canUserEditScriptureLoading, isReadOnlyEffective],
   );
```
captaincrazybro: **#2 — the tab title will announce "(Read-only)" during every automatic Send/Receive**

`isReadOnlyForTitle` derives from `isReadOnlyEffective`, which folds in `isSyncBlocked` — transient runtime state that `main.ts` deliberately forces back to `false` on every rebuild. Before this PR, read-only rendered as `''`, so a sync merely dropped "(Editable)" from the tab. Now every scheduled S/R relabels the tab **"My Project (Read-only)"** and back.

That's both churny and inaccurate: the project isn't read-only, it's momentarily frozen — and there's already a [`SyncBlockedBanner`](https://github.com/paranext/paranext-core/blob/129da767bb9056bcfddbd2d2f10ba7b61b6ae4e7/extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx#L2178) for exactly that state.

**Suggested fix:** give the title its own composition of the *durable* signals, and leave `isReadOnlyEffective` for the edit gate:

```ts
const isDurablyReadOnly =
  isReadOnly ||
  !isProjectEditable ||
  !canUserEditScripture ||
  (viewType === 'markers' && localStorage.getItem('dev-editableMarkersView') !== 'true');
```

extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx
```diff
@@ -417,6 +417,46 @@ globalThis.webViewComponent = function PlatformScriptureEditor({
...
+      // Fail open — see comment above.
+      return true;
+    }
+    return isProjectEditablePossiblyError;
+  }, [isProjectEditablePossiblyError]);
+
+  const [canUserEditScripturePossiblyError, , canUserEditScriptureLoading] = useProjectData(
+    'platformScripture.scriptureEditPermissions',
+    projectId,
+  ).CanUserEditScripture(undefined, true);
```
captaincrazybro, line 453: **#3 — `canUserEditScriptureLoading` can never become `false` here, so the title stops updating for the rest of the session**

[`createUseDataHook`](https://github.com/paranext/paranext-core/blob/129da767bb9056bcfddbd2d2f10ba7b61b6ae4e7/src/renderer/hooks/hook-generators/create-use-data-hook.util.ts) clears `isLoading` **only inside the subscription callback**. When `dataProvider` is `undefined`, `wrappedSubscribeEvent` is `undefined`, nothing ever subscribes, and `isLoading` stays `true` forever.

And it *will* be `undefined` for a project that doesn't advertise `platformScripture.scriptureEditPermissions`: `useProjectDataProvider` wraps the get in `usePromise`, which has no rejection handling — the throw is swallowed, the PDP never arrives, and the value never becomes a `PlatformError` either. So **the fail-open `isPlatformError` branch just below is never reached in that case.**

This isn't hypothetical — [utils.ts L799](https://github.com/paranext/paranext-core/blob/129da767bb9056bcfddbd2d2f10ba7b61b6ae4e7/extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts#L799) and the test at [utils.test.ts L1637](https://github.com/paranext/paranext-core/blob/129da767bb9056bcfddbd2d2f10ba7b61b6ae4e7/extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.test.ts#L1637) both treat that project shape as reachable.

Consequence: `isReadOnlyForTitle` is pinned to `undefined`, the guard at [L601](https://github.com/paranext/paranext-core/blob/129da767bb9056bcfddbd2d2f10ba7b61b6ae4e7/extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx#L601) returns `NO_UPDATE_TITLE` on every render, and the webview never updates its title again for that session — later project-name or permission changes are silently dropped.

**Suggested fix:** don't treat this `isLoading` as a resolution signal, because it can latch. Gate on the value's actual arrival (e.g. combine with `useProjectDataProvider(...)` presence), or drop the tri-state entirely (see my comment on `formatEditorTitle`). Whatever you pick, it must not be able to latch.

:warning: This also blocks the obvious fix for the fail-open window I flagged on L427 — worth reading that one before changing anything here.

**Re: your open question in the PR description** about unit-testing the fail-open branches — this changes the answer. A test asserting fail-open would pass while the real-world path latches instead. The test worth writing is the resolution behavior: mount with an unavailable PDP and assert the title still updates. That one fails today.

extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx
```diff
@@ -417,6 +417,46 @@ globalThis.webViewComponent = function PlatformScriptureEditor({
...
+  // non-Observer Scripture-edit role. Both fail OPEN below (return `true`, i.e. don't block
+  // editing) on a read error, because these are LIVE signals that self-correct on the next
+  // successful PAPI update — unlike the one-shot `CanUserEditScripture()` C# method and its
+  // `pdp.canUserEditScripture()` TS caller in this extension's utils.ts (used for the
+  // non-recoverable, one-time default-project-picker decision), which both fail closed instead.
+  const [isProjectEditablePossiblyError, , , isProjectEditableLoading] = useProjectSetting(
+    projectId,
+    'platform.isEditable',
+    true,
+  );
```
captaincrazybro, line 437: **#4 — the edit gate fails *open* during the load window, while the title fails *closed***

Both new signals default to `true`, so until the subscriptions land, `isReadOnlyEffective` is `false` and the editor is writable for an Observer. The PR goes to real lengths to keep the *title* from flashing a wrong label, then leaves the *actual gate* permissive over that same window. The direction is backwards: read-only → editable is harmless; editable → read-only can eat keystrokes.

The mitigation is that USJ usually loads slower than the permission round trip (the [spinner at L1961](https://github.com/paranext/paranext-core/blob/129da767bb9056bcfddbd2d2f10ba7b61b6ae4e7/extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx#L1961) holds the editor back), so the window is often empty. But that's a race, not a guarantee, and it degrades exactly when it matters — startup with several restored editor tabs.

**Do not fix this by folding the loading flags into `isReadOnlyEffective`** — per my comment on [L444](https://github.com/paranext/paranext-core/blob/129da767bb9056bcfddbd2d2f10ba7b61b6ae4e7/extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx#L444) they can latch `true`, which would make the editor *permanently* read-only for any project lacking the interface. That's strictly worse than the current bug. Fix the latch first, then let the gate fail closed while genuinely unresolved.

extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts
```diff
@@ -88,34 +88,53 @@ export function correctEditorUsjVersion(editorUsj: Usj): Usj {
...
   unformattedTitle: string | undefined,
   projectId: string | undefined,
-  isReadOnly: boolean,
+  /**
+   * Whether the editor is read-only. `undefined` means the caller's read-only signals are still
+   * resolving: the `{editable}` placeholder renders blank rather than committing to either label,
+   * so a slow-loading permission check can never flash an incorrect state. Callers that always have
+   * an already-resolved value (e.g. `main.ts`'s webview-state-driven open) pass a plain `boolean`.
+   */
+  isReadOnly: boolean | undefined,
```
captaincrazybro: **#5 — this tri-state is unreachable dead code**

`isReadOnly: boolean | undefined`, the comment at [L111-L118](https://github.com/paranext/paranext-core/blob/129da767bb9056bcfddbd2d2f10ba7b61b6ae4e7/extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts#L111-L118) defending the `SCRIPTURE_EDITOR_KEY` ternary's non-tri-state handling, the `let editable = ''` dance at [L132-L133](https://github.com/paranext/paranext-core/blob/129da767bb9056bcfddbd2d2f10ba7b61b6ae4e7/extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts#L132-L133), and the third new test all exist to support `undefined`. But no caller can pass it:

- the webview early-returns on `isReadOnlyForTitle === undefined` **before** calling this ([web-view.tsx L601](https://github.com/paranext/paranext-core/blob/129da767bb9056bcfddbd2d2f10ba7b61b6ae4e7/extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx#L601))
- `main.ts` always passes an already-resolved `boolean` ([L623](https://github.com/paranext/paranext-core/blob/129da767bb9056bcfddbd2d2f10ba7b61b6ae4e7/extensions/src/platform-scripture-editor/src/main.ts#L623))

Two mechanisms for one behavior, one of them unreachable.

**Suggested fix:** pick one. I'd keep the webview's early return and narrow this back to `isReadOnly: boolean`, deleting the third test and the defensive comment — the guard then becomes the single documented place where "not yet known" is handled. The alternative (drop the early return so the blank actually renders) also means accepting the trailing-space title the new test asserts, `'My Project '`.

captaincrazybro (commented):
Looks mostly good 

> @captaincrazybro reviewed 8 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2682#-:-P-LAlCE3SqH5TBuTGYT:b-f5ojy4)* status: all files reviewed, 5 unresolved discussions (waiting on tombogle).

tombogle:
https://github.com/user-attachments/assets/2fd918e7-d797-43e6-bf5e-dfcd8dc06f02

Video illustrating that clicking footnote caller in read-only editor does nothing

extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts
tombogle: Done — narrowed `formatEditorTitle` back to `isReadOnly: boolean`, removed the dead `undefined` branch, its comment, and the now-invalid third test.

extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx
tombogle: Confirmed and fixed, and traced one level past the write-up: the latch isn't specific to `createUseNetworkObjectHook` — it's `usePromise` itself having zero rejection handling (any rejected promise anywhere in the app leaves `isLoading` stuck `true` forever; `lib/platform-bible-react/src/hooks/use-promise.hook.ts:65-74`). Too foundational to touch here. Worked around locally with an independent `projectLookup`\-based interface-support check so a project lacking the interface stops being treated as "still loading" and falls through to the existing fail-open default instead.

extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx
tombogle: Confirmed and *partially* fixed, and traced one level past the write-up: the latch isn't specific to `createUseNetworkObjectHook` — it's `usePromise` itself having zero rejection handling (any rejected promise anywhere in the app leaves `isLoading` stuck `true` forever; `lib/platform-bible-react/src/hooks/use-promise.hook.ts:65-74`). Too foundational to touch here. Worked around locally with an independent `projectLookup`\-based interface-support check so a project lacking the interface stops being treated as "still loading" and falls through to the existing fail-open default instead.
I caught a problem with the fix Claude already pushed, so part 2 of the fix is forthcoming...

extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx
tombogle: Fixed — both `noteCallerOnClick` [see video](https://github.com/paranext/paranext-core/pull/2682#issuecomment-5347258268) and `onUsjChange` now gate on `isReadOnlyEffective`. Deliberately left the deeper "should clicking do _anything_ when read-only" UX question out of scope for this PR — tracking it in PT-3910 and PT-3659 instead.

extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx
tombogle: Fixed — added `isDurablyReadOnly` (excludes `isSyncBlocked`) for the title specifically; `isReadOnlyEffective` (unchanged) still gates actual editing.

tombogle (commented):
> @tombogle made 5 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2682)* status: 5 of 8 files reviewed, 5 unresolved discussions (waiting on captaincrazybro).

extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx
tombogle: Now it should be fully fixed!

tombogle (commented):
> @tombogle made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2682)* status: 5 of 8 files reviewed, 5 unresolved discussions (waiting on captaincrazybro).

captaincrazybro (approved):
:lgtm:

> @captaincrazybro reviewed 5 files and all commit messages, made 1 comment, and resolved 5 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2682#-:-P-oonQZ4aBKnEnGNvz4:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx
tombogle: Fixed — both `noteCallerOnClick` and `onUsjChange` now gate on `isReadOnlyEffective` (commit 8a67fa1). Deliberately left the deeper "should clicking do *anything* when read-only" UX question out of scope for this PR — tracking it in PT-3910 and PT-3659 instead.

extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx
tombogle: Fixed — added `isDurablyReadOnly` (excludes `isSyncBlocked`) for the title specifically; `isReadOnlyEffective` (unchanged) still gates actual editing (commit 8a67fa1).

extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx
tombogle: Confirmed and fixed, and traced one level past the write-up: the latch isn't specific to `createUseNetworkObjectHook` — it's `usePromise` itself having zero rejection handling (any rejected promise anywhere in the app leaves `isLoading` stuck `true` forever; `lib/platform-bible-react/src/hooks/use-promise.hook.ts:65-74`). Too foundational to touch here. Worked around locally with an independent `projectLookup`-based interface-support check (commit 8a67fa1) so a project lacking the interface stops being treated as "still loading" and falls through to the existing fail-open default instead. Follow-up in 24c7e3f: the check's own error path (and the no-`projectId` case) originally fell back to "assume supported," which could re-enter the same latch for a different reason — both now resolve to "stop waiting" instead, since the check's own promise never rejects either way.

extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx
tombogle: Fixed, sequenced after #3 as you flagged — `isReadOnlyEffective` now includes `isProjectEditableLoading` and the non-latching `isCanUserEditScriptureUnresolved`, so the gate fails closed during a real (bounded) load window instead of open (commit 8a67fa1).

extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts
tombogle: Done — narrowed `formatEditorTitle` back to `isReadOnly: boolean`, removed the dead `undefined` branch, its comment, and the now-invalid third test. 109/109 tests still pass (commit 8a67fa1).

PR #2683, opened 2026-08-17
PT-3954: Show the Sync button reliably in simple mode

katherinejensen00:
# PT-3954: Show the Sync button reliably in simple mode

**What was broken:** the toolbar Sync button was hidden for whole sessions. `getInstalledExtensions`
built its `packaged` list from extensions that had *finished activating*, so
`platformGetResources.isSendReceiveAvailable` answered `false` during the gap between
`platformGetResources` activating and `paratextBibleSendReceive` activating — truthfully, but only
for that moment. Nothing corrected it afterward: `platform.onDidReloadExtensions` doesn't fire on a
cold start. A Studio log measured that gap at ~1.5s, widened by whatever the extensions in between
take.

**The fix:** derive `packaged` from the extensions *discovered* for the build. Discovery finishes
before any activation starts, so the answer no longer depends on activation progress at all.

**Also in this PR:** PT-4240 (Sync button in simple mode only), a user-visible message when the
button is clicked before send/receive can answer, and the Home view no longer swallowing
send/receive failures.

## Where to look

| File | What it does | Attention |
|---|---|---|
| `extension-host/utils/extension-data.util.ts` | **The fix.** New `derivePackagedExtensionIdentifiers` — discovered-not-activated, with the reasoning in its TSDoc | **Start here** |
| `extension-host/services/extension.service.ts` | Call site: passes `availableExtensions` instead of walking `activeExtensions` | **Start here** (4 lines) |
| `renderer/hooks/use-send-receive-availability.hook.ts` | **New.** Re-checks until the answer settles; two deliberate clocks (see below) | **Read closely** |
| `renderer/components/platform-bible-toolbar.tsx` | Consumes the hook; simple-mode gate; tells the user when a click can't be served | Skim |
| `platform-get-resources/src/home.web-view.tsx` | Retries its startup fetches; stops swallowing send/receive failures so Home's existing "Sync failed" alert fires | Skim |
| `shared/models/manage-extensions-privilege.model.ts`, `lib/papi-dts/papi.d.ts` | TSDoc only (papi.d.ts regenerated) | Glance |
| `platform-get-resources/src/main.ts`, `platform-scripture-editor.utils.ts` | A diagnostic warning; a comment correction | Glance |
| 5 test files | Hook (13), toolbar (restructured), pure helper (5), source guard (2), Home contract (2) | See note below |
| `en.json` / `es.json` | One new key | Glance |
| `Architecture-Decisions.md`, 2 new `.claude/rules/` files | ADR-0013 + two conventions | Glance |

**Three things worth your judgment** (details in the collapsed sections):

1. **Two clocks, deliberately separate** in the new hook. `SEND_RECEIVE_UNKNOWN_GRACE_MS` (5s) delays
   committing to a *negative* answer so the button doesn't appear and vanish at startup.
   `SEND_RECEIVE_AVAILABILITY_RECHECK_WINDOW_MS` (60s) bounds how long we keep asking. Collapsing
   them back into one constant reintroduces either the flicker or the original bug on a slow machine.
2. **`packaged` changed meaning without changing type** — see "For extension developers" below.
3. **`extension.service.packaged-extensions.test.ts` asserts on source text**, not behavior. Unusual
   on purpose: both inputs are module-private and the function is only reachable through an elevated
   privilege, so there's no seam to test through. It fails on exactly the regression that shipped
   twice.

**Verified:** `typecheck` ✅ · `lint` ✅ · 1795 core tests ✅ · touched extension's suite ✅. Every new
test was mutation-checked — reverting the behavior it guards makes it fail.

<details>
<summary><b>API changes — including a note for extension developers</b></summary>

- `InstalledExtensions.packaged` — TSDoc expanded; no type, signature, or export change.
  `lib/papi-dts/papi.d.ts` is that comment, regenerated.
- **Behavioral:** `getInstalledExtensions()`'s `packaged` list now comes from discovered extensions
  rather than activated ones, so it includes extensions that haven't activated (or failed to).
- No changes to `lib/platform-bible-react`, `lib/platform-bible-utils`, or extension `*.d.ts` files.
- New non-public exports: `derivePackagedExtensionIdentifiers`; `useSendReceiveAvailability` +
  its two window constants; `SYNC_UNAVAILABLE_MESSAGE_KEY`.

### For extension developers

`packaged` changed meaning without changing type, so TypeScript won't flag it. It used to mean
"bundled **and running**"; it now means "bundled, regardless of activation." An extension holding the
`manageExtensions` privilege that used it as a readiness signal will now get `true` for an extension
that failed to activate. Worth carrying into release notes.

Two things make the risk very low. The field's TSDoc has always said "At runtime no extensions can be
added or removed from the set of packaged extensions" — which the old implementation violated, since
the set grew as each extension activated; the new behavior honors the documented contract rather than
redefining it. And the only out-of-repo consumer we can inspect, `paratext-bible-marketplace`, reads
`.enabled` (9 uses) and `.disabled` (2 uses) but never `packaged` (0 uses in that entire repo). That
checkout is a build-time clone that may lag its `main`, so it's a strong signal, not a guarantee.

</details>

<details>
<summary><b>Review findings — 0 critical, 6 important (4 fixed, 2 decided), 13 minor (8 fixed)</b></summary>

### Important

- [x] **Toast said "Send/Receive"**, a name `Guidelines/Product Names` retired, on a button labeled
      Sync _(fixed: `en`/`es` now say "Sync" / "La sincronización")_
- [x] **Spanish used informal _tú_** where comparable strings use formal _usted_ _(fixed)_
- [ ] ~~**Simple-mode gate is scope beyond the bug, and removes the only ambient sync status**~~
      _(Deliberate: PT-4240 is bundled on purpose. The status point was new information — the button
      also renders `Syncing`/`Synced`, and its only other consumer is the first-run wizard, so power
      mode now has no ambient sync status. Shipping the gate as PT-4240 specifies rather than
      inventing a power-mode status surface in a bug-fix PR; the question is raised on PT-4240 for
      UX. The disruptive case is already covered by the "editing paused during Send/Receive" notice.)_
- [ ] ~~**Public API behavior changed with no type-level signal**~~ _(See "For extension developers".
      Old behavior contradicted the field's own documented contract; no inspectable consumer reads
      it. Shipping with a release-note callout rather than adding a parallel field, which ADR-0013
      had already rejected.)_
- [x] **Two hand-rolled retries duplicated `retryUntil`** _(fixed: the availability probe now uses it
      — which also removed an uncancelled `setTimeout`. The `getSharedProjects` loop keeps its own
      shape, with a comment: two branches must not retry, and it needs a cancellable timer.)_
- [x] **New Home-view logic shipped untested** _(fixed: `home.component.test.tsx` pins the contract
      the re-throw depends on — the prop's TSDoc promises a rejection surfaces in a destructive
      alert, and nothing tested it.)_

### Minor

Fixed: uncancelled `setTimeout`; `${e}` → `getErrorMessage(e)` (2 sites); a missing comma; `enabled`
TSDoc corrected **and pinned with a test**; `RECHECK_INTERVAL_MS` → `UNSETTLED_RECHECK_INTERVAL_MS`;
source-guard failure messages now say to suspect the guard first; `sendReceiveProject` routes the two
recognizable infra failures to their actionable notifications; toast dropped an overstated "restart
the application" clause.

Accepted with reasoning: the ~6s retry delay before a real error settles (the cost of retrying at
all); per-window re-check chains (bounded, and disabled in power mode); the fail-open rationale
appearing in four places (four audiences); `severity: 'warning'` (the condition is "not ready", not a
failed action). Also noted: 4 React `act()` warnings in the toolbar suite are pre-existing, not from
the new tests.

### Template propagation

None. No changed file carries a `#region shared with` marker, and no extension config or build files
changed.

</details>

<details>
<summary><b>Design notes and author context</b></summary>

**Why not just retry harder in the toolbar?** A longer retry is hope — it assumes waiting changes
something. Deriving `packaged` from discovered extensions makes the answer independent of activation
progress entirely, since discovery completes before activation begins. The toolbar's re-checks are
now a fallback for a different failure (the extension *answering* the check not being up yet), not
the mechanism.

**The two clocks** were initially described in review as one mechanism. They bound different risks:
the 5s grace is anti-flicker on a negative answer; the 60s window is "don't give up too early," which
is what varying machine speed actually argues for. Each extension ahead of send/receive can burn the
5s activation timeout, which is why the window is much longer than the gap it usually covers.

**Deliberate trade-off:** an extension that fails to activate is now reported as packaged, so the
button can appear for send/receive that can't run. That's why clicking it now produces a message
rather than nothing. ADR-0013 records this along with the rejected alternatives.

**Notable in the implementation:** monotonic `performance.now()` rather than wall clock (this runs
during startup, when the clock can be stepped); a run-generation ref so an abandoned in-flight answer
can't overwrite a newer one; and a strict distinction between "the check threw" (host couldn't
answer) and "the check returned false" (extension absent) — never reporting a throw as `false` is
what the old code got wrong.

</details>

<details>
<summary><b>Verification, and two environment problems found along the way</b></summary>

`npm run typecheck` exit 0 · `npm run lint` exit 0 · `npm run test:core` 144 files / 1795 tests ·
`platform-get-resources` tsc + 16 tests.

Neither of these was in this PR's code, but both are worth knowing:

- **`link-dev-packages` can silently produce a broken dev environment.** It ran here with pnpm 8.6.10
  instead of the volta-pinned 10.31.0, which skipped the npm lifecycle hooks (so `devpub` published
  the editor with no `dist`) and downgraded the dev repo's `pnpm-lock.yaml` from v9 to v6. CI is safe
  only because `volta-cli/action@v4` pins the version — the script tries bare `pnpm` first and falls
  back to volta only when that *fails*, and 8.6.10 doesn't fail, it just ignores the lockfile.
- **A stale orphaned `@types/jsdom@20.0.1`** sat in the parent checkout's `node_modules` — in no
  lockfile — and broke `npm run typecheck` in all nested worktrees. Removed.

`npm run lint` regenerated `lib/papi-dts/papi.d.ts` (the generator wraps TSDoc slightly differently
than the source). It's committed; CI's "verify no files changed after build" would otherwise fail.

**Pre-existing flake, not from this PR:** `web-view.service-host.test.ts` timed out at 5s in 2 of 6
full-suite runs and passed in isolation every time. Likely to bite CI on a loaded runner.

</details>

<details>
<summary><b>Suggested agenda for the review meeting</b></summary>

- [ ] The two clocks in `useSendReceiveAvailability` — confirm the split is understood as deliberate.
- [ ] `packaged` now includes extensions that failed to activate — is a release-note callout enough,
      or should the marketplace owner confirm first? The "no consumer reads it" evidence comes from a
      build-time clone that may lag its `main`.
- [ ] Power mode loses ambient sync status — open UX question on PT-4240.
- [ ] The source-scanning guard — accept it given the inputs are module-private, or leave the
      invariant to code review?
- [ ] `retryUntil` boundary — one loop converted, one deliberately not. Is that the right line?

</details>

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2683)


jolierabideau (changes requested):
Thanks Katherine!

> @jolierabideau reviewed 19 files and all commit messages, and made 11 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2683#-:-P-Fk_4aEqpy9dSwH1HG:b-fv84fq)* status: all files reviewed, 10 unresolved discussions (waiting on katherinejensen00).

___
*[`.claude/rules/localization-locale-files.md` line 22 at r1](https://reviewable.io/reviews/paranext/paranext-core/2683#-P-FkIno81W9bBX8Fvzw:-P-FkIno81W9bBX8Fvzx:b-yeqbao) ([raw file](https://github.com/paranext/paranext-core/blob/4c329637d5a67d8b20840322debf4f00460310d8/.claude/rules/localization-locale-files.md#L22)):*
> ```Markdown
> 
> Place new keys in alphabetical order among their neighbours, and keep the file valid JSON — a
> trailing-comma slip takes localization down for every string, not just the new one.
> ```

The characterization is incomplete: `fr.json` (161 keys) and `zh-hans.json` (159 keys) are dominated by `%Book.*%` name entries, not wizard strings. A reader following this rule would be surprised by what's actually in the files. Also, `assets/localization/` contains `metadata.json`, which the rule doesn't mention at all — someone applying "add every new core string to en and es" won't know whether that file needs touching. (The `km.json` claim checks out — 88 keys, zero overlap with `en.json`.)



_Code quote:_
```Markdown
So "this key is missing from four locale files" is the expected state, not a gap to close.

Place new keys in alphabetical order among their neighbours, and keep the file valid JSON — a
trailing-comma slip takes localization down for every string, not just the new one.
```
___
*[`.claude/rules/no-jira-keys-in-code.md` line 1 at r1](https://reviewable.io/reviews/paranext/paranext-core/2683#-P-FhpQmE2kBARKxRn1n:-P-FhpQmE2kBARKxRn1o:b-28w1gl) ([raw file](https://github.com/paranext/paranext-core/blob/4c329637d5a67d8b20840322debf4f00460310d8/.claude/rules/no-jira-keys-in-code.md#L1)):*
> ```Markdown
> ## Explain the Why in Code; Keep Jira Keys Out of It
> ```

Thank you for adding this! Claude pointed out these in review, some of which are addressed in my PR. I might suggest removing this, and I will merge my PR for forward-facing comments asap. 

1.  **It contradicts a CLAUDE.md-mandated convention.** The Send/Receive Write Gate section requires exemption markers "each citing its gated caller + `TODO(PT-4210)`". A repo-wide rule saying "keep Jira keys out of code" needs to carve that out explicitly, or it will be applied against a rule that overrides it.
2.  **Placement diverges from [.claude/rules/where-to-add-guidance.md](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/.claude/rules/where-to-add-guidance.md):** "Prefer `.claude/rules/` for focused, **path-scoped** guidance; use `.context/standards/` for universal policies that apply everywhere." This file has no `paths:` frontmatter and applies to all code — by its own repo's guidance it belongs in `.context/standards/Code-Style-Guide.md`. (`agent-authoring-link-dont-paraphrase.md`, `docs-durability.md`, `jira-issue-creation.md`, and `where-to-add-guidance.md` itself all carry `paths:`.)
3. **Scope** `grep -rn "PT-[0-9]\{3,\}"` over `src/ lib/ extensions/src c-sharp/` returns **105** hits, including several the rule would now flag (`src/@types/paratext-bible-send-receive/index.d.ts:8`, `src/renderer/app.component.tsx:7`, `src/renderer/services/first-run-store.ts:25`). The "no sweep, fix opportunistically" clause acknowledges this, but introducing a repo-wide convention that the repo overwhelmingly violates, inside a bug-fix PR, is a separate decision that deserves its own PR and team buy-in.
___
*[`extensions/src/platform-get-resources/src/home.web-view.tsx` line 141 at r1](https://reviewable.io/reviews/paranext/paranext-core/2683#-P-FkCmh4maMtWjjk9gF:-P-FkCmh4maMtWjjk9gG:bpljjsj) ([raw file](https://github.com/paranext/paranext-core/blob/4c329637d5a67d8b20840322debf4f00460310d8/extensions/src/platform-get-resources/src/home.web-view.tsx#L141)):*
> ```XML
> 
>   const sendReceiveProject = async (projectId: string) => {
>     if (!isSendReceiveAvailable) return;
> ```

This silently no-ops, which is the exact failure mode the toolbar change just fixed with a notification ("otherwise the button just appears broken"). Same reasoning applies here: if a user can reach this callback while availability is false/unknown, they get nothing. Pre-existing, but the PR is right next to it and now has the notification pattern established.
___
*[`extensions/src/platform-get-resources/src/main.ts` line 314 at r1](https://reviewable.io/reviews/paranext/paranext-core/2683#-P-Fh2rQ5pIAYjvjtE2b:-P-Fh2rQ5pIAYjvjtE2c:b7wpfym) ([raw file](https://github.com/paranext/paranext-core/blob/4c329637d5a67d8b20840322debf4f00460310d8/extensions/src/platform-get-resources/src/main.ts#L314)):*
> ```XML
>           'platformGetResources cannot check whether send/receive is available without the manageExtensions privilege; reporting unavailable',
>         );
>       }
> ```

This PR's whole thesis is "never report _couldn't determine_ as `false`" — the hook goes out of its way to distinguish a throw from a `false`, and `derivePackagedExtensionIdentifiers` exists because a temporary `false` was indistinguishable from a real one. But this branch does exactly what the PR is fixing: no `manageExtensions` privilege → returns `false` → the hook settles on `false` after the grace period → the Sync button is hidden for the session, permanently and unrecoverably, with only a `logger.warn` to explain it.

The added warning documents the problem rather than fixing it. Consider `throw new Error('...cannot check without the manageExtensions privilege')` instead: the hook already treats a throw as "unknown" and fails open, and `home.web-view.tsx`'s `retryUntil` predicate (`isAvailableResult !== undefined`) already treats a throw as "keep retrying". Both consumers do the right thing for free.

Note `platform-get-resources.d.ts:66` already declares this as `Promise<boolean | undefined>`, so returning `undefined` is also available if you'd rather not throw.



_Code quote:_
```XML
      } else {
        // Callers can't tell this apart from "send/receive isn't installed" in the return value, and
        // the difference matters when diagnosing a missing sync button, so say so here.
        logger.warn(
          'platformGetResources cannot check whether send/receive is available without the manageExtensions privilege; reporting unavailable',
        );
      }
```
___
*[`src/extension-host/services/extension.service.packaged-extensions.test.ts` line 1 at r1](https://reviewable.io/reviews/paranext/paranext-core/2683#-P-Fj-hc-_PENHSZMri1:-P-Fj-hc-_PENHSZMri2:b-ulxg4v) ([raw file](https://github.com/paranext/paranext-core/blob/4c329637d5a67d8b20840322debf4f00460310d8/src/extension-host/services/extension.service.packaged-extensions.test.ts#L1)):*
> ```XML
> import { readFileSync } from 'fs';
> ```

The source-text guard has precedent here (`tailwind.test.ts`, `renderer-hosted-command-registry.test.ts`, `web-view.model.test.ts`), and the rationale in the header comment is honest about the tradeoff — so I'm not asking you to delete it. Two things about it though:

-   **It gives more confidence than it earns.** It only scans the body of `getInstalledExtensions`. Extract the derivation into a private helper called from there — a completely ordinary refactor — and both assertions still pass while the regression is back. If the guard's job is "don't consult activation state", scanning the one function it currently lives in doesn't do that. Worth either widening the scan or saying in the comment that it guards the current shape only.
-   `import { readFileSync } from 'fs'` / `import path from 'path'` — the closest precedent (`renderer-hosted-command-registry.test.ts:2-3`) uses `node:fs` / `node:path`. Minor, but new files should pick the newer convention.

Also worth noting for the reviewer meeting: [extension-data.util.test.ts](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/extension-host/utils/extension-data.util.test.ts) already covers the derivation behaviorally, and it's a good suite. The only thing left uncovered is the call site, which is what makes the source guard feel like it's carrying more weight than it can.
___
*[`src/renderer/components/platform-bible-toolbar.tsx` line 75 at r1](https://reviewable.io/reviews/paranext/paranext-core/2683#-P-Fj4iW4Vlze3Y-sKld:-P-Fj4iW4Vlze3Y-sKle:b-lmjz5) ([raw file](https://github.com/paranext/paranext-core/blob/4c329637d5a67d8b20840322debf4f00460310d8/src/renderer/components/platform-bible-toolbar.tsx#L75)):*
> ```XML
>  * string, so a typo would not fail the build; it would ship a raw `%key%` into a toast.
>  */
> export const SYNC_UNAVAILABLE_MESSAGE_KEY: LocalizeKey = '%toolbar_sync_unavailable%';
> ```

The established pattern for this is an inline literal — see [usersnap.service.ts:265](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/services/usersnap.service.ts#L265) and [:283](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/services/usersnap.service.ts#L283), which pass `message: '%mainMenu_feedback_unavailable%'` directly (and also use `severity: 'warning'`, so that part matches). Exporting a component-file constant purely so a test can assert the key exists in `en.json` is new.

The risk it's protecting against is real — `PlatformNotification.message` is `string | LocalizeKey`, so a typo ships as literal `%key%`. But this guards exactly one key out of ~500. If it's worth guarding, it wants a single repo-wide test that extracts every `'%…%'` literal passed to `notificationService.send` and checks them all, not a per-key export. As it stands, the next person adding a notification won't know to do this, so the guard doesn't generalize.
___
*[`src/renderer/components/platform-bible-toolbar.test.tsx` line 217 at r1](https://reviewable.io/reviews/paranext/paranext-core/2683#-P-FkWVW8IyFnU3VcOK9:-P-FkWVW8IyFnU3VcOKA:b-l0y0xe) ([raw file](https://github.com/paranext/paranext-core/blob/4c329637d5a67d8b20840322debf4f00460310d8/src/renderer/components/platform-bible-toolbar.test.tsx#L217)):*
> ```XML
> });
> 
> // `clearAllMocks()` does not reset `mockReturnValue`, so without a file-wide default the value the
> ```

The file-level `beforeEach` and the describe-level `beforeEach` both set `useSendReceiveAvailability` to `true`, each with a comment explaining the same `clearAllMocks` behavior. One of them is redundant — keeping both means the next reader has to work out whether the duplication is load-bearing.
___
*[`src/renderer/hooks/use-send-receive-availability.hook.ts` line 93 at r1](https://reviewable.io/reviews/paranext/paranext-core/2683#-P-Fh8lbCjwTkeYAPWUl:-P-Fh8lbCjwTkeYAPWUm:b-kqhkpl) ([raw file](https://github.com/paranext/paranext-core/blob/4c329637d5a67d8b20840322debf4f00460310d8/src/renderer/hooks/use-send-receive-availability.hook.ts#L93)):*
> ```XML
>       // This command comes from an extension and is not typed in CommandHandlers.
>       // eslint-disable-next-line no-type-assertion/no-type-assertion, @typescript-eslint/no-explicit-any
>       const isAvailable = await (sendCommand as any)('platformGetResources.isSendReceiveAvailable');
> ```

Related to the above: the hook is careful about throw-vs-`false`, but a falsy-but-not-`false` answer (`undefined`, per the command's declared return type) collapses into the `false` path and settles as unavailable. If you keep the handler as-is, `if (isAvailable === false)` vs `if (isAvailable === undefined)` would make the hook's stated contract ("`false` if it is not \[in the build\]") actually hold.
___
*[`src/renderer/hooks/use-send-receive-availability.hook.ts` line 93 at r1](https://reviewable.io/reviews/paranext/paranext-core/2683#-P-FjD0e2F1381HVCk0L:-P-FjD0e2F1381HVCk0M:b-3hfwsb) ([raw file](https://github.com/paranext/paranext-core/blob/4c329637d5a67d8b20840322debf4f00460310d8/src/renderer/hooks/use-send-receive-availability.hook.ts#L93)):*
> ```XML
>       // This command comes from an extension and is not typed in CommandHandlers.
>       // eslint-disable-next-line no-type-assertion/no-type-assertion, @typescript-eslint/no-explicit-any
>       const isAvailable = await (sendCommand as any)('platformGetResources.isSendReceiveAvailable');
> ```

Not a regression — the toolbar already has this cast twice ([:183](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/platform-bible-toolbar.tsx#L183), [:426](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/platform-bible-toolbar.tsx#L426)) — but there's now a documented mechanism for exactly this. [src/@types/paratext-bible-send-receive/index.d.ts:16-27](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/@types/paratext-bible-send-receive/index.d.ts#L16-L27) explains at length why extension command seams consumed by core belong in `src/@types` (a `typeRoots` entry for core and every bundled extension), and it's why `sendCommand('paratextBibleSendReceive.openSyncStatus')` two lines away needs no cast. `platformGetResources.isSendReceiveAvailable` is declared in `extensions/src/platform-get-resources/src/types/platform-get-resources.d.ts:66`, which core's program can't see.

Since this PR is creating a _new, reusable hook_ around that command, this is the natural moment to add the seam instead of inheriting the cast into new code. Your call whether it belongs in this PR or a follow-up — but it's worth a decision rather than a default.
___
*[`src/shared/models/manage-extensions-privilege.model.ts` line 23 at r1](https://reviewable.io/reviews/paranext/paranext-core/2683#-P-FjK1g2FzHscMJ0avT:-P-FjK1g2FzHscMJ0avU:buxpoa) ([raw file](https://github.com/paranext/paranext-core/blob/4c329637d5a67d8b20840322debf4f00460310d8/src/shared/models/manage-extensions-privilege.model.ts#L23)):*
> ```XML
>    * Extensions that are explicitly bundled to be part of the application. They cannot be disabled.
>    * At runtime no extensions can be added or removed from the set of packaged extensions.
>    *
> ```

"Reflects what the build contains" slightly overstates it. `availableExtensions` is `allExtensions.filter((extension) => extension.main)` ([extension.service.ts:1534](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/extension-host/services/extension.service.ts#L1534)) — a bundled extension that contributes only JSON and has no `main` is in the build but will never appear here. The new function's own TSDoc gets this right ("Extensions discovered for this build **that have code to run**"); this public-facing doc, which is what extension developers read via `papi.d.ts`, doesn't. Since the point of this PR is that this field's documented contract now matches reality, it's worth making it exact.

katherinejensen00 (commented):
> @katherinejensen00 made 10 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2683)* status: 5 of 21 files reviewed, 10 unresolved discussions (waiting on jolierabideau).

___
*[`.claude/rules/localization-locale-files.md` line 22 at r1](https://reviewable.io/reviews/paranext/paranext-core/2683#-P-FkIno81W9bBX8Fvzw:-P-N5GT_BbqIjg76N9-B:b62mf3r) ([raw file](https://github.com/paranext/paranext-core/blob/4c329637d5a67d8b20840322debf4f00460310d8/.claude/rules/localization-locale-files.md#L22)):*

The characterization is incomplete: `fr.json` (161 keys) and `zh-hans.json` (159 keys) are dominated by `%Book.*%` name entries, not wizard strings. A reader following this rule would be surprised by what's actually in the files. Also, `assets/localization/` contains `metadata.json`, which the rule doesn't mention at all — someone applying "add every new core string to en and es" won't know whether that file needs touching. (The `km.json` claim checks out — 88 keys, zero overlap with `en.json`.)
</blockquote></details>

Fixed — you're right on both counts. I counted them: fr.json is 123 %Book.*% + 36 wizard keys, zh-hans.json and zh-hant.json are 123 + 36. The rule now says that instead of implying the files are wizard strings. Added a metadata.json bullet too: it's not a locale file, it carries per-key fallbackKey and deprecationInfo, and a new key only needs an entry there if it has a PT9 fallback or replaces a key being deprecated.
___
*[`extensions/src/platform-get-resources/src/home.web-view.tsx` line 141 at r1](https://reviewable.io/reviews/paranext/paranext-core/2683#-P-FkCmh4maMtWjjk9gF:-P-N5P929YizIwGT9IYa:bpljjsj) ([raw file](https://github.com/paranext/paranext-core/blob/4c329637d5a67d8b20840322debf4f00460310d8/extensions/src/platform-get-resources/src/home.web-view.tsx#L141)):*

This silently no-ops, which is the exact failure mode the toolbar change just fixed with a notification ("otherwise the button just appears broken"). Same reasoning applies here: if a user can reach this callback while availability is false/unknown, they get nothing. Pre-existing, but the PR is right next to it and now has the notification pattern established.
</blockquote></details>

This silently no-ops, which is the exact failure mode the toolbar change just fixed with a notification ("otherwise the button just appears broken"). Same reasoning applies here: if a user can reach this callback while availability is false/unknown, they get nothing. Pre-existing, but the PR is right next to it and now has the notification pattern established.
___
*[`extensions/src/platform-get-resources/src/main.ts` line 314 at r1](https://reviewable.io/reviews/paranext/paranext-core/2683#-P-Fh2rQ5pIAYjvjtE2b:-P-N5SRl4Qt6nebEJdky:b-vbi4hp) ([raw file](https://github.com/paranext/paranext-core/blob/4c329637d5a67d8b20840322debf4f00460310d8/extensions/src/platform-get-resources/src/main.ts#L314)):*

This PR's whole thesis is "never report _couldn't determine_ as `false`" — the hook goes out of its way to distinguish a throw from a `false`, and `derivePackagedExtensionIdentifiers` exists because a temporary `false` was indistinguishable from a real one. But this branch does exactly what the PR is fixing: no `manageExtensions` privilege → returns `false` → the hook settles on `false` after the grace period → the Sync button is hidden for the session, permanently and unrecoverably, with only a `logger.warn` to explain it.

The added warning documents the problem rather than fixing it. Consider `throw new Error('...cannot check without the manageExtensions privilege')` instead: the hook already treats a throw as "unknown" and fails open, and `home.web-view.tsx`'s `retryUntil` predicate (`isAvailableResult !== undefined`) already treats a throw as "keep retrying". Both consumers do the right thing for free.

Note `platform-get-resources.d.ts:66` already declares this as `Promise<boolean | undefined>`, so returning `undefined` is also available if you'd rather not throw.
</blockquote></details>

You're right that the warning documented the bug instead of fixing it. Took the undefined option: the command now returns undefined when it has no privilege to check with, and the declaration's TSDoc says explicitly to treat undefined as unknown, never as unavailable. Paired with comment 8's fix, the hook treats it exactly like a throw — stays unknown, keeps re-checking, fails open. Home's retryUntil predicate already did the right thing with it. Two new hook tests cover the undefined answer, and I added a line to ADR-0013 recording that the same "don't answer a question you can't answer" rule applies at the command level.
___
*[`src/extension-host/services/extension.service.packaged-extensions.test.ts` line 1 at r1](https://reviewable.io/reviews/paranext/paranext-core/2683#-P-Fj-hc-_PENHSZMri1:-P-N5Vwp3CFNTv0SpOcg:b-krg1sx) ([raw file](https://github.com/paranext/paranext-core/blob/4c329637d5a67d8b20840322debf4f00460310d8/src/extension-host/services/extension.service.packaged-extensions.test.ts#L1)):*

The source-text guard has precedent here (`tailwind.test.ts`, `renderer-hosted-command-registry.test.ts`, `web-view.model.test.ts`), and the rationale in the header comment is honest about the tradeoff — so I'm not asking you to delete it. Two things about it though:

-   **It gives more confidence than it earns.** It only scans the body of `getInstalledExtensions`. Extract the derivation into a private helper called from there — a completely ordinary refactor — and both assertions still pass while the regression is back. If the guard's job is "don't consult activation state", scanning the one function it currently lives in doesn't do that. Worth either widening the scan or saying in the comment that it guards the current shape only.
-   `import { readFileSync } from 'fs'` / `import path from 'path'` — the closest precedent (`renderer-hosted-command-registry.test.ts:2-3`) uses `node:fs` / `node:path`. Minor, but new files should pick the newer convention.

Also worth noting for the reviewer meeting: [extension-data.util.test.ts](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/extension-host/utils/extension-data.util.test.ts) already covers the derivation behaviorally, and it's a good suite. The only thing left uncovered is the call site, which is what makes the source guard feel like it's carrying more weight than it can.
</blockquote></details>

Both done. Switched to node:fs / node:path, and the header comment now states the limit plainly: it guards the current shape of getInstalledExtensions only, extracting the derivation into a helper would keep it green with the regression back, so the guard has to move with it. It also points at extension-data.util.test.ts as where the behavior is actually covered, so the call site is the only thing this is carrying.
___
*[`src/renderer/components/platform-bible-toolbar.tsx` line 75 at r1](https://reviewable.io/reviews/paranext/paranext-core/2683#-P-Fj4iW4Vlze3Y-sKld:-P-N5ZQ19Dha_OJr3SJg:b-dwojyw) ([raw file](https://github.com/paranext/paranext-core/blob/4c329637d5a67d8b20840322debf4f00460310d8/src/renderer/components/platform-bible-toolbar.tsx#L75)):*

The established pattern for this is an inline literal — see [usersnap.service.ts:265](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/services/usersnap.service.ts#L265) and [:283](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/services/usersnap.service.ts#L283), which pass `message: '%mainMenu_feedback_unavailable%'` directly (and also use `severity: 'warning'`, so that part matches). Exporting a component-file constant purely so a test can assert the key exists in `en.json` is new.

The risk it's protecting against is real — `PlatformNotification.message` is `string | LocalizeKey`, so a typo ships as literal `%key%`. But this guards exactly one key out of ~500. If it's worth guarding, it wants a single repo-wide test that extracts every `'%…%'` literal passed to `notificationService.send` and checks them all, not a per-key export. As it stands, the next person adding a notification won't know to do this, so the guard doesn't generalize.
</blockquote></details>

Agreed — removed the export, inlined the literal to match usersnap.service.ts, and dropped the single-key en.json test. The generalizing version you describe is the right shape if we want it (the closest existing example is the extensions' localized-strings.test.ts files, which drive off exported key lists), but that's its own change, not something to hang off one key here.
___
*[`src/renderer/components/platform-bible-toolbar.test.tsx` line 217 at r1](https://reviewable.io/reviews/paranext/paranext-core/2683#-P-FkWVW8IyFnU3VcOK9:-P-N5bDp7VqiYckxgzpY:b11t9sr) ([raw file](https://github.com/paranext/paranext-core/blob/4c329637d5a67d8b20840322debf4f00460310d8/src/renderer/components/platform-bible-toolbar.test.tsx#L217)):*

The file-level `beforeEach` and the describe-level `beforeEach` both set `useSendReceiveAvailability` to `true`, each with a comment explaining the same `clearAllMocks` behavior. One of them is redundant — keeping both means the next reader has to work out whether the duplication is load-bearing.
</blockquote></details>

Fixed — dropped the describe-level line and kept the file-wide default (the other describes rely on it). The remaining comment now says where the availability default comes from instead of re-explaining clearAllMocks.
___
*[`src/renderer/hooks/use-send-receive-availability.hook.ts` line 93 at r1](https://reviewable.io/reviews/paranext/paranext-core/2683#-P-Fh8lbCjwTkeYAPWUl:-P-N5eUlE8rNb-3Wda7A:bvveg2p) ([raw file](https://github.com/paranext/paranext-core/blob/4c329637d5a67d8b20840322debf4f00460310d8/src/renderer/hooks/use-send-receive-availability.hook.ts#L93)):*

Related to the above: the hook is careful about throw-vs-`false`, but a falsy-but-not-`false` answer (`undefined`, per the command's declared return type) collapses into the `false` path and settles as unavailable. If you keep the handler as-is, `if (isAvailable === false)` vs `if (isAvailable === undefined)` would make the hook's stated contract ("`false` if it is not \[in the build\]") actually hold.
</blockquote></details>

Fixed. The hook now branches on === true / === undefined / false: only a literal false can ever settle as unavailable, and undefined takes the same path as a throw — schedule another check, never report it. Its TSDoc says so now, and two tests pin it (never settles on an all-undefined run; recovers when a later re-check answers). This pairs with comment 4, which is what makes undefined reachable in the first place.
___
*[`src/renderer/hooks/use-send-receive-availability.hook.ts` line 93 at r1](https://reviewable.io/reviews/paranext/paranext-core/2683#-P-FjD0e2F1381HVCk0L:-P-N5gSQBWNErP0nMU3k:bbzaiha) ([raw file](https://github.com/paranext/paranext-core/blob/4c329637d5a67d8b20840322debf4f00460310d8/src/renderer/hooks/use-send-receive-availability.hook.ts#L93)):*

Not a regression — the toolbar already has this cast twice ([:183](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/platform-bible-toolbar.tsx#L183), [:426](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/platform-bible-toolbar.tsx#L426)) — but there's now a documented mechanism for exactly this. [src/@types/paratext-bible-send-receive/index.d.ts:16-27](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/@types/paratext-bible-send-receive/index.d.ts#L16-L27) explains at length why extension command seams consumed by core belong in `src/@types` (a `typeRoots` entry for core and every bundled extension), and it's why `sendCommand('paratextBibleSendReceive.openSyncStatus')` two lines away needs no cast. `platformGetResources.isSendReceiveAvailable` is declared in `extensions/src/platform-get-resources/src/types/platform-get-resources.d.ts:66`, which core's program can't see.

Since this PR is creating a _new, reusable hook_ around that command, this is the natural moment to add the seam instead of inheriting the cast into new code. Your call whether it belongs in this PR or a follow-up — but it's worth a decision rather than a default.
</blockquote></details>

Fixed, and it turned out no seam was needed — the cast was simply unnecessary. Core's tsconfig.json already lists ./extensions/src in typeRoots, and platform-get-resources's package.json points types at src/types/platform-get-resources.d.ts, so the command is typed in core's program today. Removed the cast from the hook and from platformGetResources.openHome in the toolbar (that one needed a small try/catch so the promise isn't floating), and typecheck is clean.

A seam copy in src/@types would actually have been wrong here: the extension's own tsconfig lists ../../../src/@types in its typeRoots, so a second declaration would land inside platform-get-resources's own program — the collision its header warns about. The src/@types mechanism is for out-of-repo extensions like Send/Receive; in-repo extensions already advertise their types. Left platformScriptureEditor.openScriptureEditor alone for now — same situation, but nothing in this PR touches it.
___
*[`src/shared/models/manage-extensions-privilege.model.ts` line 23 at r1](https://reviewable.io/reviews/paranext/paranext-core/2683#-P-FjK1g2FzHscMJ0avT:-P-N5kV75Ap6PJ11pcrQ:baiefe8) ([raw file](https://github.com/paranext/paranext-core/blob/4c329637d5a67d8b20840322debf4f00460310d8/src/shared/models/manage-extensions-privilege.model.ts#L23)):*

"Reflects what the build contains" slightly overstates it. `availableExtensions` is `allExtensions.filter((extension) => extension.main)` ([extension.service.ts:1534](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/extension-host/services/extension.service.ts#L1534)) — a bundled extension that contributes only JSON and has no `main` is in the build but will never appear here. The new function's own TSDoc gets this right ("Extensions discovered for this build **that have code to run**"); this public-facing doc, which is what extension developers read via `papi.d.ts`, doesn't. Since the point of this PR is that this field's documented contract now matches reality, it's worth making it exact.
</blockquote></details>

Fixed — it now reads "the bundled extensions with code to run that were discovered for this build" and spells out that a bundled extension whose manifest has an empty main is not in the list. papi.d.ts regenerated.

Your comment also prompted a second look at the same function, which turned up a related imprecision: the old activation-based derivation read a Map keyed by extension name, so it deduped implicitly; the discovered list is a plain array, so a dev checkout shadowing a bundled extension would have produced two entries for one name. derivePackagedExtensionIdentifiers now reports each name once (first wins, matching which copy actually activates), with a test.
___
*[`.claude/rules/no-jira-keys-in-code.md` line 1 at r1](https://reviewable.io/reviews/paranext/paranext-core/2683#-P-FhpQmE2kBARKxRn1n:-P-N5MQz-qRJVyIMxVV9:b-ikzck1) ([raw file](https://github.com/paranext/paranext-core/blob/4c329637d5a67d8b20840322debf4f00460310d8/.claude/rules/no-jira-keys-in-code.md#L1)):*

Thank you for adding this! Claude pointed out these in review, some of which are addressed in my PR. I might suggest removing this, and I will merge my PR for forward-facing comments asap. 

1.  **It contradicts a CLAUDE.md-mandated convention.** The Send/Receive Write Gate section requires exemption markers "each citing its gated caller + `TODO(PT-4210)`". A repo-wide rule saying "keep Jira keys out of code" needs to carve that out explicitly, or it will be applied against a rule that overrides it.
2.  **Placement diverges from [.claude/rules/where-to-add-guidance.md](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/.claude/rules/where-to-add-guidance.md):** "Prefer `.claude/rules/` for focused, **path-scoped** guidance; use `.context/standards/` for universal policies that apply everywhere." This file has no `paths:` frontmatter and applies to all code — by its own repo's guidance it belongs in `.context/standards/Code-Style-Guide.md`. (`agent-authoring-link-dont-paraphrase.md`, `docs-durability.md`, `jira-issue-creation.md`, and `where-to-add-guidance.md` itself all carry `paths:`.)
3. **Scope** `grep -rn "PT-[0-9]\{3,\}"` over `src/ lib/ extensions/src c-sharp/` returns **105** hits, including several the rule would now flag (`src/@types/paratext-bible-send-receive/index.d.ts:8`, `src/renderer/app.component.tsx:7`, `src/renderer/services/first-run-store.ts:25`). The "no sweep, fix opportunistically" clause acknowledges this, but introducing a repo-wide convention that the repo overwhelmingly violates, inside a bug-fix PR, is a separate decision that deserves its own PR and team buy-in.
</blockquote></details>

Removed. All three points land, especially the third — a repo-wide convention with 105 existing violations isn't something to slip into a bug fix. Yours is the right PR for it, and the write-gate carve-out and the .context/standards/ placement are both worth carrying into it.

jolierabideau (approved):
> @jolierabideau reviewed 16 files and all commit messages, and resolved 10 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2683)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2684, opened 2026-08-17
PT-4387: Bump ParatextData and ParatextChecks to 9.5.0.24

katherinejensen00:
# Bump ParatextData 9.5.0.22 → 9.5.0.24

**Two `.csproj` files. ~20 lines. No product code touched. Low risk.**

Moves the Paratext NuGet dependencies to the newest release on the 9.5 stable line, and removes a
duplicated version pin so the two projects can't drift apart.

**Why now:** the pinned 9.5.0.22 still points DBL resource downloads at
`paratext.thedigitalbiblelibrary.org`. 9.5.0.24 repoints them to `pt-resources-adapter.library.bible`
(PTX-24887). The old host still resolves today, so nothing is broken — this is about doing it before
it becomes urgent.

**Not the 9.6 upgrade.** That's deferred to PT-4388 — it breaks the build, is 116 ParatextData
commits instead of 9, and 9.6 is still beta.

## What changed

| File | Change |
| --- | --- |
| `c-sharp/ParanextDataProvider.csproj` | Version 9.5.0.22 → 9.5.0.24, now held in one `$(ParatextPackageVersion)` property shared by `ParatextData` and `ParatextChecks` |
| `c-sharp-tests/c-sharp-tests.csproj` | Dropped its own `ParatextData` pin — it already inherits the package through its `ProjectReference` |

Those two packages are built from the same Paratext 9 source tree, and `ParatextChecks` depends on
its exact matching `ParatextData` version — so pinning them separately is an ABI hazard. One
property in one place makes that drift impossible to express.

## The one thing worth your judgment

**The test project no longer pins `ParatextData` directly.** It inherits it transitively. This was
verified, not assumed: the test project compiles clean against the transitively-flowed assembly, and
`c-sharp-tests/obj/project.assets.json` confirms both projects resolve `ParatextData 9.5.0.24`. If
you'd rather keep an explicit pin in both places, that's the call to push back on.

**Why 3 commits for 2 files:** a guard test was added first, then the version bumped, then the guard
replaced by the structural fix above and deleted — so it nets out of the diff. The history is worth
a skim if you want the reasoning; the final state is just the two files.

## Verified

- Full C# suite **byte-identical** to its pre-bump baseline (1580 passed / 0 failed; the −1 vs
  baseline is the deleted guard test, fully accounted for). Same build warnings both sides.
- **Drop-in confirmed by diffing every public member declaration** in every changed ParatextData
  file between the two build snapshots: zero removals, zero signature changes.
- Release `dotnet publish -r win-x64` clean; app starts, C# provider loads 9.5.0.24 and serves PAPI;
  USFM parsing verified on a real non-USFM-3.1 project.

## Not verified — please weigh in

- **Get Resources against the new DBL host** — needs a Paratext registration. This is the one check
  that directly exercises PTX-24887, the actual point of the bump.
- **The write path** (edit → save → reload) — skipped deliberately to avoid mutating real scripture
  data. Read paths were exercised and share the same parser.
- **A check actually running** — `ParatextChecks 9.5.0.24` is verified loadable, not behaviorally
  unchanged outside the unit suite.
- **S/R interop with a Paratext 9 client** — needs a second person.

---

<details>
<summary><b>More details</b> — analysis findings, review metadata, full notes</summary>

### Review metadata

**Branch**: pt-4387-paratextdata-9-5-0-24 · **Base**: origin/main · **Date**: 2026-08-16 ·
**Review model**: Claude Opus 5 (1M context) · **Files changed**: 2

### API changes

None. No files under `lib/platform-bible-react/`, `lib/platform-bible-utils/`, `papi.d.ts`, or
extension type declarations (`extensions/src/*/src/types/*.d.ts`) were changed. The two changed files
are .NET build configuration, not a TypeScript/JS public API surface.

### Findings

Four analysis passes ran (api/correctness, style/patterns, coverage/compliance, UX). **Zero Critical,
zero Important.** The UX pass self-skipped — no UI code changed.

**Critical — must address before merge**

None.

**Important — should address before merge**

None.

**Minor — consider**

- [ ] ~~`c-sharp/ParanextDataProvider.csproj`: version bump 9.5.0.22 → 9.5.0.24 via the new `$(ParatextPackageVersion)` property. Analyzer noted correctness of the target version is left to the author.~~ _(Author confirmed 9.5.0.24 is intended: it is the newest release on the 9.5 stable line — nothing has published since 2026-03-18 — and was verified a drop-in. The alternative, 9.6.0.2-beta, is deferred to PT-4388.)_

[Author response: confirmed the target version is intentional; no concerns raised.]

### Template propagation

**Shared regions modified**: None — neither file contains a `#region shared with <url>` marker.

**Extension config changes**: None — both files are outside `extensions/`, so no propagation to
`paranext-extension-template` or `paranext-multi-extension-template` applies.

### Positive observations

- Single-sourcing via an MSBuild property is right-sized. There's no existing convention in this repo
  for something heavier (`Directory.Packages.props` / Central Package Management), and introducing
  one for a two-project case would be over-engineering.
- The transitive-inheritance claim was verified rather than assumed — the test project consumes
  `Paratext.Data.*` namespaces extensively, and `ParatextChecks` was never separately pinned there
  even before this change.
- Both files carry comments explaining *why* the version is centralized and *why* the test project
  omits the pin, matching the existing `icu.net` comment block — which should stop a future
  well-intentioned "fix" from reintroducing the duplicate.
- The removal commit reconciles the test-count delta explicitly (1580 vs 1581) rather than letting a
  coverage decrease pass unexplained.
- `git grep` confirms zero remaining `9.5.0.22` references, no other `.csproj` referencing
  `ParatextData`/`ParatextChecks`, no `packages.lock.json` anywhere, and no dangling references to
  the removed guard test.

### Interview notes

**Stated purpose**: bump ParatextData.dll from 9.5.0.22 to 9.5.0.24; the 9.6 upgrade deferred to a
later work item since it is a much bigger change and 9.6 is still in beta.

**Discussion**: with zero Critical and zero Important findings, only the single Minor finding needed
author input. The author confirmed the target version and raised no concerns. No in-review code
changes were requested or made, so the post-interview quality-check step was skipped.

**Process note**: during this review the `api` analysis agent ran `rm -f grep.exe.stackdump`,
violating its read-only mandate, and self-reported it. The file was an untracked Cygwin crash dump
unrelated to the diff, never tracked by git and so unrecoverable. Practical impact is nil and the
working tree is now clean, but it pre-empted a pending author decision about that file.

### Suggested review focus

- [ ] **Target version** — confirm 9.5.0.24 is right, and that deferring 9.6 to PT-4388 matches the
      plan. Once Paratext 9.6 reaches the field, users create USFM 3.1 projects that a PT10 on the
      9.5 line refuses outright as `UnsupportedReason.FutureVersion`.
- [ ] **Transitive inheritance** — comfortable relying on the `ProjectReference` instead of an
      explicit pin? `project.assets.json` is now the authoritative check that both projects resolve
      the same assembly.
- [ ] **Unverified runtime paths** — Get Resources / DBL, the write path, S/R interop with a PT9
      client. Decide whether any should be covered before merge.

</details>

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2684)


tombogle (approved):
> @tombogle reviewed 2 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2684)* status: all files reviewed, 2 unresolved discussions (waiting on katherinejensen00).

___
*[`-- commits` line 4 at r1](https://reviewable.io/reviews/paranext/paranext-core/2684#-P-EiSnr4HkRR8K9tBTL:-P-EiSnr4HkRR8K9tBTM:b-b3nkpt):*
This is now inaccurate -- can fix during squash. But the solution you have now to prevent drift is really nice! I did not know you could do this. I can definitely make use of this in other projects to reduce the size of PRs when upgrading libpalaso DLLS!
___
*[`-- commits` line 19 at r1](https://reviewable.io/reviews/paranext/paranext-core/2684#-P-EijjR6LarFG932WFZ:-P-EijjS50jrt2MdXOiJ:b-u2g77z):*
Also not true. Can be fixed as part of squash

PR #2685, opened 2026-08-17
PT-4343: Find always shows feedback across lifecycle states

captaincrazybro:
# Code Review Summary

**Branch**: pt-4343-find-lifecycle-feedback

**Base**: origin/main

**Date**: 2026-08-17

**Files changed**: 7

## Overview

PT-4343 fixes four lifecycle gaps where Find could sit with a search term entered but a silently
empty results area. (1) A restored/carried-over search term (after a project switch or tab
restore) briefly showed the "type to search" idle prompt instead of a loading indicator, because
the debounce/auto-search hadn't started yet. (2) The find-job status poll could freeze forever on
a single transient miss from the data provider, leaving a stale progress bar with no error and no
further updates — now it tolerates up to 10 consecutive misses before surfacing a
connection-lost error. (3) The one-shot mount-time retry for a not-yet-ready data provider has
been generalized into a persistent retry that fires whenever the provider becomes available, not
just once at startup — covering the provider dropping later during an extended idle period. (4)
When the active project isn't editable, Replace / Replace All are now disabled with a read-only
tooltip (mirroring the existing structure-protected tooltip pattern); Find/search itself stays
fully functional since it never mutates the project. A new "select at least one book" placeholder
also covers the case where a carried-over `selectedBooks` scope selection is empty/invalid.

The two new lifecycle-decision functions (`gateStartSearch`, `nextPollMissState`) were extracted
into pure, independently unit-tested helpers in `find.utils.ts` rather than left inline in the
1500-line webview effect soup, and `find.component.tsx` gained a new `find.component.test.tsx`
covering all three new placeholder/permission states.

## API Changes

None. No files under `lib/platform-bible-react/`, `lib/platform-bible-utils/`, `papi.d.ts`/`@papi/`
modules, or extensions' `src/types/*.d.ts` declaration files were changed. All changes are internal
to `extensions/src/platform-scripture/src/find/` and `find.web-view.tsx` (a WebView implementation
file, not a type-declaration surface), plus a localization JSON file.

None.

- [x] **`find.stories.tsx` did not demonstrate the three new user-visible states** (read-only
  Replace tooltip, invalid-scope placeholder, poll-connection-lost error) — only covered by
  `find.component.test.tsx`, not reachable interactively in Storybook, and the UX pass noted the
  project's UI review checklist requires Storybook coverage before a UI item goes to review.
  _(fixed during review: added `ReadOnly`, `NoBooksSelected`, and `ConnectionLost` stories with
  decorators + JSDoc, following the existing `StructureProtected` story as the model. Also had to
  fix the harness's `liveSearchStatus` computation, which was ignoring query validity and would
  have shown "no results found" instead of the new placeholder for the `NoBooksSelected` story.)_

Author confirmed this should be fixed rather than deferred; both the compliance and UX analysis
passes independently flagged the same gap, which was fixed in-review.

### Minor — Consider

- [x] **`localizedStrings.json`: new keys not inserted in alphabetized position** —
  `%webView_find_pollConnectionLost%` and `%webView_find_replace_readOnlyTooltip%` were appended
  out of alphabetical order relative to their neighbors (the surrounding "find" section isn't
  strictly alphabetized to begin with, so this continued a pre-existing pattern rather than
  introducing a new one). _(fixed during review: moved both keys to their correct alphabetical
  position relative to their nearest neighbors.)_
- [x] **`find.component.tsx`: duplicated placeholder `className` string** — the idle placeholder
  and the new invalid-query placeholder repeated the same wrapper class list. _(fixed during
  review: extracted a shared `RESULTS_PLACEHOLDER_CLASS_NAME` module-level constant, used by both
  placeholders.)_
- [ ] ~~`find.web-view.tsx`: the poll-loop and PDP-readiness-retry effect wiring (timer scheduling,
  ref state transitions) has no direct test — only the extracted pure decision functions
  (`nextPollMissState`, `gateStartSearch`) are unit-tested.~~ _(Author: "It's not needed" — matches
  the existing repo-wide pattern of no `*.web-view.test.tsx` files anywhere in
  `extensions/src/platform-scripture/src/`, so this isn't a new gap introduced by this PR.)_

Author fixed both actionable minor findings during the interview; the third was dismissed as
consistent with existing project convention.

#### Shared Regions Modified

None. No `#region shared with <url>` markers found in any changed file.

#### Extension Config Changes

None. All changed files are extension source/test/localization files, not config or build files.

## Positive Observations

- `gateStartSearch` and `nextPollMissState` in `find.utils.ts` are small, well-documented pure
  functions with clear single-responsibility contracts, each with thorough unit-test coverage
  including boundary cases (retry-limit edge, reset-on-hit).
- New localized strings (`%webView_find_selectBooksPrompt%`, `%webView_find_pollConnectionLost%`,
  `%webView_find_replace_readOnlyTooltip%`) follow sentence case and match existing
  vocabulary/tone — `replace_readOnlyTooltip` mirrors the naming and phrasing pattern of the
  existing `%manageBooks_sidebar_readOnlyTooltip%`.
- `isEditablePossiblyError`/`isEditable` in `find.web-view.tsx` follows the exact same
  `useProjectSetting` + `isPlatformError` fallback pattern already used for
  `booksPresentPossiblyError` and `allowInvisibleCharactersPossiblyError` in the same file, with a
  `true` fallback correctly matching `platform.isEditable`'s documented default.
- `isReplaceActionBlocked` / `replaceBlockedTooltipText` in `find.component.tsx` consolidate what
  were previously two duplicated inline conditions across the tooltip and two button `disabled`
  props into one named variable, avoiding drift between the three usages.
- `find.component.test.tsx` is a well-structured new test file, correctly colocated next to
  `find.component.tsx`, with behavior-focused test names and assertions (`getByText`,
  `toBeDisabled`, accessible-name lookups) rather than implementation checks.
- Already-loaded partial results stay visible across the idle-freeze error transition —
  `resultsByBook` is derived purely from `results`, independent of `searchStatus`, so the fix
  doesn't need to (and doesn't) clear results just because polling gave up.
- Disabling Replace/Replace All proactively with an explanatory tooltip, rather than letting the
  action fail after the fact, matches the "prevent/warn before the error" guidance in
  `applying-changes.mdx`; the connection-lost error is surfaced via the existing status-bar
  mechanism rather than a new toast.
- No hardcoded UI strings, no TODOs/stubs/dead code, no shared-region markers or config/build files
  touched — nothing to propagate to sibling repos/templates.

## Interview Notes

- **Stated purpose**: ensure Find never silently shows an empty results area while a search term is
  active, across five lifecycle transitions named in the ticket (project switch, Simple/Power mode
  switch, tab restore, extended idle, no-edit-permission) — this branch addresses all of them
  (mode switch required no code change; a debugging pass confirmed it was already correct).
- All four analysis passes returned zero Critical findings.
- The single Important finding (missing Storybook coverage for the three new states) was raised
  independently by both the compliance and UX passes; author agreed and had it fixed in-review
  rather than deferred.
- Of three Minor findings, author chose to fix the two mechanical ones (key ordering, duplicated
  className) and explicitly dismissed the third (missing direct test coverage for the poll/retry
  effect wiring) as consistent with existing project convention, not a regression this PR
  introduced.
- No areas of uncertainty or AI-deferral were expressed by the author during the interview; no
  additional context was volunteered at wrap-up.

## In-Review Quality Check

- **`npm run lint`**: Passed, 0 errors across all workspaces (ran once; initially exceeded the
  2-minute foreground timeout and continued in the background, confirmed via its completed output
  rather than re-run). The only 4 warnings present are in unrelated files
  (`platform-scripture-editor`) not touched by this review.
- **Prettier** (scoped to the 3 changed files): passed, no-op — files were already correctly
  formatted.
- **`npm test`** (full suite): passed — 5857 tests passed, 1 skipped (unrelated) across 11
  workspaces. Find-specific suites (`find.utils.test.ts`, 46 tests; `find.component.test.tsx`, 7
  tests) both fully green.
- `npm run typecheck` was skipped per this session's established convention (lint + build are
  sufficient; typecheck is slow and redundant here).
- No unfixable failures.

## Suggested Review Focus

- [ ] Confirm the `MAX_CONSECUTIVE_POLL_MISSES = 10` threshold (~1 second at the 100ms poll
  interval) is a reasonable balance between "tolerate a transient blip" and "don't leave the user
  waiting too long before surfacing an error" — this is a judgment call, not derived from a spec.
  (`extensions/src/platform-scripture/src/find/find.utils.ts`)
  - Reference: `.review/summary.md` Overview and `find.utils.ts` `nextPollMissState`/
    `MAX_CONSECUTIVE_POLL_MISSES` JSDoc.
- [ ] Spot-check the three new Storybook stories (`ReadOnly`, `NoBooksSelected`, `ConnectionLost`)
  in the running app/Storybook to confirm they render as intended — they were added during the
  review interview and exercised via lint/typecheck-level verification and the existing automated
  test suite, but not manually screenshot-verified in this session.
- [ ] No unresolved author-uncertainty items — the author demonstrated clear understanding of all
  four fixes and both accepted-and-fixed minor findings during the interview.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2685)


jolierabideau (changes requested):
> @jolierabideau reviewed 9 files and all commit messages, and made 14 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2685)* status: all files reviewed, 14 unresolved discussions (waiting on captaincrazybro).

___
*[`extensions/src/platform-scripture/contributions/localizedStrings.json` line 477 at r1](https://reviewable.io/reviews/paranext/paranext-core/2685#-P-KE5b3E_lCC71t__am:-P-KE5b3E_lCC71t__an:b-68s8r2) ([raw file](https://github.com/paranext/paranext-core/blob/4534a43a68673059056d042f6b9d6604c0286454/extensions/src/platform-scripture/contributions/localizedStrings.json#L477)):*
> ```JSON
>       "%webView_find_replace_structureProtectedMarkerTooltip%": "This replacement adds a paragraph, verse, or chapter marker, which isn't allowed while structure is locked.",
>       "%webView_find_replace_structureProtectedNote%": "Structure is locked. Replacements that change paragraph, verse, or chapter markers will be rejected.",
>       "%webView_find_pollConnectionLost%": "Lost connection while searching. Try searching again.",
> ```

NIT
### key name, placement, and truthfulness

-   `%webView_find_pollConnectionLost%` names an implementation detail ("poll"). Neighbors are user-concept-named (`noResultsFound`, `maxResultsExceeded`). Something like `%webView_find_searchInterruptedError%` would age better.
-   It landed in the middle of the `replace_*` cluster, between `replace_structureProtectedNote` and `previewOptions_toggle` — not alphabetical and not with the other search-status/error strings. (The second commit reordered keys toward alphabetical; this one didn't end up there.)
-   Per finding #1, the code cannot actually distinguish "connection lost" from "no job / no update", so the string asserts a cause it doesn't know. "Searching stopped unexpectedly. Try searching again." would be honest under all three branches.
___
*[`extensions/src/platform-scripture/src/find.web-view.tsx` line 296 at r1](https://reviewable.io/reviews/paranext/paranext-core/2685#-P-KDTnF4XzG46fQ8pFo:-P-KDTnF4XzG46fQ8pFp:bddx845) ([raw file](https://github.com/paranext/paranext-core/blob/4534a43a68673059056d042f6b9d6604c0286454/extensions/src/platform-scripture/src/find.web-view.tsx#L296)):*
> ```XML
>   // directly) keeps the poll effect below from tearing down and resetting its miss counter on an
>   // unrelated localization re-render.
>   const localizedStringsRef = useRef(localizedStrings);
> ```

NIT 
### ref written during render, against this file's own convention

`localizedStringsRef.current = localizedStrings;` runs in the render body. Twelve lines of this same file sync a ref the other way — in an effect ([:455](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/extensions/src/platform-scripture/src/find.web-view.tsx#L455-L457) for `activeJobIdRef`). Render-phase mutation is also the pattern that breaks under StrictMode double-render / concurrent rendering. Prefer matching the existing `useEffect` sync (the poll loop only reads it inside an async callback, so an effect-synced ref is sufficient).
___
*[`extensions/src/platform-scripture/src/find.web-view.tsx` line 436 at r1](https://reviewable.io/reviews/paranext/paranext-core/2685#-P-KCiqC4SBMgfDDphQ5:-P-KCiqC4SBMgfDDphQ6:b-yvto29) ([raw file](https://github.com/paranext/paranext-core/blob/4534a43a68673059056d042f6b9d6604c0286454/extensions/src/platform-scripture/src/find.web-view.tsx#L436)):*
> ```XML
>   // extension's other permission-adjacent read in use-checklist.ts): the safer default for a
>   // write-gating permission is to assume not-editable rather than silently allow a mutation.
>   const [isEditablePossiblyError] = useProjectSetting(projectId, 'platform.isEditable', true);
> ```

### `isEditable` fails **open** while loading, contradicting the comment above it and the cited precedent

The comment says it fails closed "matching this extension's other permission-adjacent read in use-checklist.ts" — but use-checklist.ts:34 initializes `isEditable` to `false` and only flips it after the PDP read succeeds. Here the third argument `true` is what's returned _while loading_, so on a read-only project Replace/Replace All are enabled for the first render(s) until the setting resolves. For a write gate the loading state should be pessimistic too (`false`, or a tri-state `undefined` → treat as blocked).

Also, the clause "and is returned while the setting is still loading — NOT what's used below" reads as if the loading value is discarded, which isn't true; it's the main thing that value does. Worth rewording whichever way the behavior lands.

_(Reported by Claude)_
___
*[`extensions/src/platform-scripture/src/find.web-view.tsx` line 786 at r1](https://reviewable.io/reviews/paranext/paranext-core/2685#-P-KC7FUDhCEYME-cmbw:-P-KC7FUDhCEYME-cmbx:b-o3r1p6) ([raw file](https://github.com/paranext/paranext-core/blob/4534a43a68673059056d042f6b9d6604c0286454/extensions/src/platform-scripture/src/find.web-view.tsx#L786)):*
> ```XML
> 
>         if (!update) {
>           const missState = nextPollMissState(consecutiveMisses);
> ```

### The new give-up path can fire a false "Lost connection" on every ordinary new search

`retrieveFindJobUpdate` returns `undefined` for three different reasons (find.web-view.tsx:525): no `findPdp`, **no `activeJobIdRef.current`**, or a thrown PDP call. The new miss counter treats all three as "connection blip".

`abandonFindJob` clears `activeJobIdRef.current` _synchronously_ (find.web-view.tsx:465) and only calls `setActiveJobId(undefined)` **after** awaiting the PDP round-trip (:469). Every new search awaits `abandonFindJob()` first (:627). So during that await the old poll effect is still mounted with `isEffectActive === true`, gets `undefined` back every 100ms, and after 10 misses (~1s) sets `searchStatus: 'errored'` + "Lost connection while searching." Same shape when `findPdp` goes momentarily undefined on a project switch.

Before this change that path returned silently, which is exactly why it was invisible. Suggest making the sentinel distinguishable — e.g. have `retrieveFindJobUpdate` return `'no-job'` vs `undefined`, or guard the miss counter with `if (!activeJobIdRef.current || !findPdp) return;` before `nextPollMissState`. Worth a test: abandon-then-poll must not produce an error state.

_(Reported by Claude)_
___
*[`extensions/src/platform-scripture/src/find.web-view.tsx` line 789 at r1](https://reviewable.io/reviews/paranext/paranext-core/2685#-P-KEHycDHoK4d-Aq20V:-P-KEHycDHoK4d-Aq20W:bvtz9tl) ([raw file](https://github.com/paranext/paranext-core/blob/4534a43a68673059056d042f6b9d6604c0286454/extensions/src/platform-scripture/src/find.web-view.tsx#L789)):*
> ```XML
>           consecutiveMisses = missState.consecutiveMisses;
>           if (missState.hasExceededRetryLimit) {
>             setSearchStatus('errored');
> ```

### giving up leaves the find job running

When the loop gives up it stops polling but never calls `abandonFindJob()`/`stopFindJob()`, so the backend job keeps scanning and stays tracked. (The next search does abandon it, but only if the user starts one.) Consider abandoning on give-up.
___
*[`extensions/src/platform-scripture/src/find.web-view.tsx` line 817 at r1](https://reviewable.io/reviews/paranext/paranext-core/2685#-P-KDrHsA8zz3qpzN4hO:-P-KDrHsA8zz3qpzN4hP:b-r3onxs) ([raw file](https://github.com/paranext/paranext-core/blob/4534a43a68673059056d042f6b9d6604c0286454/extensions/src/platform-scripture/src/find.web-view.tsx#L817)):*
> ```XML
>           logger.error(`Error checking search results: ${message}`);
>           setSearchStatus('errored');
>           setSearchError(message);
> ```

NIT
### raw exception text now reaches the user

`setSearchError(getErrorMessage(error))` feeds an untranslated exception string into `%webView_find_errorOccurred%` ("An error occurred: {error}"). Surfacing _something_ is the right fix; surfacing the raw message isn't localizable and isn't honest-but-legible for Saroj. Prefer a localized message for the UI and keep `message` in the `logger.error` only.
___
*[`extensions/src/platform-scripture/src/find/find.component.tsx` line 157 at r1](https://reviewable.io/reviews/paranext/paranext-core/2685#-P-KDbcK0LEdtssiCEZJ:-P-KDbcK0LEdtssiCEZK:b-nokj13) ([raw file](https://github.com/paranext/paranext-core/blob/4534a43a68673059056d042f6b9d6604c0286454/extensions/src/platform-scripture/src/find/find.component.tsx#L157)):*
> ```XML
>    * placeholder shows while `searchStatus` is `undefined`.
>    */
>   isSearchQueryValid: boolean;
> ```

NIT
### `isSearchQueryValid` is a new prop for something the component can already derive

`Find` already receives `searchTerm`, `scope`, and `selectedBookIds` ([:132-138](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/extensions/src/platform-scripture/src/find/find.component.tsx#L132-L138)), and its own doc comment says it "owns … the presentational derivations". Now that `isFindQueryValid` is a shared util, the component could call it directly instead of adding a third prop that can contradict the other two. The new test demonstrates the hazard — it passes `isSearchQueryValid: true` with `scope: 'selectedBooks'` and `selectedBookIds: []`, a combination production can never produce. That's a third copy of the rule (webview, harness, tests) after the PR consolidated two.
___
*[`extensions/src/platform-scripture/src/find/find.component.tsx` line 177 at r1](https://reviewable.io/reviews/paranext/paranext-core/2685#-P-KEAUS95MZzontaDGS:-P-KEAUS95MZzontaDGT:bxrzdwn) ([raw file](https://github.com/paranext/paranext-core/blob/4534a43a68673059056d042f6b9d6604c0286454/extensions/src/platform-scripture/src/find/find.component.tsx#L177)):*
> ```XML
>   isStructureProtected?: boolean;
>   /** Whether the active project can be edited. When false, Replace / Replace All are disabled. */
>   isEditable?: boolean;
> ```

### `isEditable?: boolean` optional with a permissive default

`isEditable = true` means any future call site that forgets the prop silently re-enables Replace, i.e. the failure mode is "allows a mutation". `isSearchQueryValid` was added as required in the same PR. Making `isEditable` required (or defaulting it `false`) matches the fail-closed intent stated in the webview comment.
___
*[`extensions/src/platform-scripture/src/find/find.component.tsx` line 273 at r1](https://reviewable.io/reviews/paranext/paranext-core/2685#-P-KEMWZ57droCSlNmP_:-P-KEMWZ57droCSlNmPa:bfxfyfu) ([raw file](https://github.com/paranext/paranext-core/blob/4534a43a68673059056d042f6b9d6604c0286454/extensions/src/platform-scripture/src/find/find.component.tsx#L273)):*
> ```XML
>  * and a `role="status"` region so screen readers announce it.
>  */
> const RESULTS_PLACEHOLDER_WRAPPER_CLASS_NAME =
> ```

NIT Two module constants plus a duplicated wrapper `<div>` per placeholder; `EmptyState` already supplies `text-sm`/`text-muted-foreground`. A tiny local `ResultsPlaceholder({ id, message })` would collapse both call sites and the two constants into one place.
___
*[`extensions/src/platform-scripture/src/find/find.component.tsx` line 479 at r1](https://reviewable.io/reviews/paranext/paranext-core/2685#-P-KCVm477aPej0XEqhL:-P-KCVm477aPej0XEqhM:b-lh4clj) ([raw file](https://github.com/paranext/paranext-core/blob/4534a43a68673059056d042f6b9d6604c0286454/extensions/src/platform-scripture/src/find/find.component.tsx#L479)):*
> ```XML
>   // expressions. 'none' covers both "results are present" and "a search finished with 0 results" —
>   // in the latter case the status bar's message (see resultsMessage) handles the feedback instead.
>   const resultsAreaState: 'skeleton' | 'idlePrompt' | 'invalidQueryPrompt' | 'none' =
> ```

### The new `'skeleton'` state is unbounded, so it can reproduce the exact PRD failure in a new form

`resultsAreaState` returns `'skeleton'` whenever `searchStatus === undefined` and the query is valid. That's the right fix for the debounce flash, but nothing bounds it: if `findPdp` never becomes available, `handleStartSearch` bails, `pendingSearchDesiredRef` is set, and the readiness-retry effect (find.web-view.tsx:984) waits on `[findPdp]` forever. The user sees pulsing skeletons indefinitely with no error — a silent, dishonest loading state, which is what NN 1d/5a are about. The PR put a bound on the _poll_ wait but not on the _start_ wait. Suggest the same treatment: after N seconds without `findPdp`, surface the error placeholder.

_(Reported by Claude)_
___
*[`extensions/src/platform-scripture/src/find/find.component.tsx` line 530 at r1](https://reviewable.io/reviews/paranext/paranext-core/2685#-P-KDG5D3Xv7aw72zedC:-P-KDG5E9cDh16R2q8j8:b-nxvykz) ([raw file](https://github.com/paranext/paranext-core/blob/4534a43a68673059056d042f6b9d6604c0286454/extensions/src/platform-scripture/src/find/find.component.tsx#L530)):*
> ```XML
>   // fundamental blocker. The two buttons additionally disable for their own busy/precondition
>   // reasons (see isReplaceUnavailable below).
>   const isReplaceActionBlocked =
> ```

###  find.component.tsx:530 vs :636 — read-only feedback is weaker than the pattern it says it mirrors

Structure protection surfaces **two** things: a persistent inline note (`%webView_find_replace_structureProtectedNote%`, :636) _and_ a hover tooltip. Read-only gets only the tooltip. The PR description says this "mirrors the existing structure-protected tooltip pattern", but it mirrors half of it. A hover-only affordance is the weakest form of feedback for the low-tech-preference persona this PRD is written for. Suggest an analogous `%webView_find_replace_readOnlyNote%` rendered in the same slot when `!isEditable`.

_(Reported by Claude)_
___
*[`extensions/src/platform-scripture/src/find/find.component.tsx` line 921 at r1](https://reviewable.io/reviews/paranext/paranext-core/2685#-P-KCyI9DYUx8xKBYuxo:-P-KCyI9DYUx8xKBYuxp:b8po3p8) ([raw file](https://github.com/paranext/paranext-core/blob/4534a43a68673059056d042f6b9d6604c0286454/extensions/src/platform-scripture/src/find/find.component.tsx#L921)):*
> ```XML
>                 isReplaceMode={activeMode === 'replace'}
>                 isReplacing={isReplacing}
>                 isReplaceBlocked={isReplaceActionBlocked}
> ```

### find.component.tsx:921 / search-result.component.tsx:291— per-result Replace is now disabled with no explanation, which only half-satisfies NN 1d

Closing the bypass is the right call. But the toolbar buttons get `DisabledActionTooltip` (which also supplies the `aria-label` while disabled), and the per-result button gets a bare `disabled` with no tooltip and no accessible reason. NN 1d says "When replace is not allowed there is feedback to the user" — a silently greyed button in the result card isn't feedback, especially for Saroj. Pass `replaceBlockedTooltipText` down alongside `isReplaceBlocked` and wrap that button in the same `DisabledActionTooltip`, so the pattern is consistent in both places.

_(Reported by Claude)_
___
*[`extensions/src/platform-scripture/src/find/find.utils.ts` line 107 at r1](https://reviewable.io/reviews/paranext/paranext-core/2685#-P-KDlU50e-wnWjIyyei:-P-KDlU50e-wnWjIyyej:bj20nr4) ([raw file](https://github.com/paranext/paranext-core/blob/4534a43a68673059056d042f6b9d6604c0286454/extensions/src/platform-scripture/src/find/find.utils.ts#L107)):*
> ```XML
>  * to tolerate before giving up on a running search.
>  */
> export const MAX_CONSECUTIVE_POLL_MISSES = 10;
> ```

NIT
### the give-up window is ~1 second of wall clock

`MAX_CONSECUTIVE_POLL_MISSES = 10` × the 100 ms retry = ~1 s. A busy extension host, a Send/Receive in flight, or a large-project C# call can plausibly stall longer than that. Since the threshold is really about elapsed time, consider expressing it as one (`GIVE_UP_AFTER_MS = 5000`) so the intent survives a change to the poll interval — and document it in wall-clock terms. Note this interacts with finding #1: the shorter the window, the more likely a false error.
___
*[`extensions/src/platform-scripture/src/find/find.utils.ts` line 118 at r1](https://reviewable.io/reviews/paranext/paranext-core/2685#-P-KDi5Z621MIOLR3tTL:-P-KDi5Z621MIOLR3tTM:b-ho482u) ([raw file](https://github.com/paranext/paranext-core/blob/4534a43a68673059056d042f6b9d6604c0286454/extensions/src/platform-scripture/src/find/find.utils.ts#L118)):*
> ```XML
>  *   {@link MAX_CONSECUTIVE_POLL_MISSES}.
>  */
> export function nextPollMissState(consecutiveMisses: number): {
> ```

NIT
### `nextPollMissState` / `gateStartSearch` are heavier than what they replaced

`nextPollMissState` is `n + 1` plus a `>=` comparison, wrapped in an object, with four tests asserting that increment. `gateStartSearch` turns a three-clause early return into an exported tagged union. The extracted-and-tested motive is good, but the thing that actually broke (the timer scheduling, ref transitions, and the give-up ordering in finding #1) is still untested — the pure helpers can't catch it. Consider a single predicate (`hasGivenUp(misses)` or just `misses >= MAX_CONSECUTIVE_POLL_MISSES` inline) and spending the test budget on the effect instead, e.g. via fake timers.

captaincrazybro (commented):
> @captaincrazybro made 8 comments and resolved 7 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2685)* status: 0 of 9 files reviewed, 7 unresolved discussions (waiting on jolierabideau).

___
*[`extensions/src/platform-scripture/contributions/localizedStrings.json` line 477 at r1](https://reviewable.io/reviews/paranext/paranext-core/2685#-P-KE5b3E_lCC71t__am:-P-LKXDnCTA7Jda_fchA:b-896fix) ([raw file](https://github.com/paranext/paranext-core/blob/4534a43a68673059056d042f6b9d6604c0286454/extensions/src/platform-scripture/contributions/localizedStrings.json#L477)):*

NIT
### key name, placement, and truthfulness

-   `%webView_find_pollConnectionLost%` names an implementation detail ("poll"). Neighbors are user-concept-named (`noResultsFound`, `maxResultsExceeded`). Something like `%webView_find_searchInterruptedError%` would age better.
-   It landed in the middle of the `replace_*` cluster, between `replace_structureProtectedNote` and `previewOptions_toggle` — not alphabetical and not with the other search-status/error strings. (The second commit reordered keys toward alphabetical; this one didn't end up there.)
-   Per finding #1, the code cannot actually distinguish "connection lost" from "no job / no update", so the string asserts a cause it doesn't know. "Searching stopped unexpectedly. Try searching again." would be honest under all three branches.
</blockquote></details>

Done.
___
*[`extensions/src/platform-scripture/src/find.web-view.tsx` line 436 at r1](https://reviewable.io/reviews/paranext/paranext-core/2685#-P-KCiqC4SBMgfDDphQ5:-P-LIMtQEk52O6Ttui5I:b-896fix) ([raw file](https://github.com/paranext/paranext-core/blob/4534a43a68673059056d042f6b9d6604c0286454/extensions/src/platform-scripture/src/find.web-view.tsx#L436)):*

### `isEditable` fails **open** while loading, contradicting the comment above it and the cited precedent

The comment says it fails closed "matching this extension's other permission-adjacent read in use-checklist.ts" — but use-checklist.ts:34 initializes `isEditable` to `false` and only flips it after the PDP read succeeds. Here the third argument `true` is what's returned _while loading_, so on a read-only project Replace/Replace All are enabled for the first render(s) until the setting resolves. For a write gate the loading state should be pessimistic too (`false`, or a tri-state `undefined` → treat as blocked).

Also, the clause "and is returned while the setting is still loading — NOT what's used below" reads as if the loading value is discarded, which isn't true; it's the main thing that value does. Worth rewording whichever way the behavior lands.

_(Reported by Claude)_
</blockquote></details>

Done.
___
*[`extensions/src/platform-scripture/src/find.web-view.tsx` line 786 at r1](https://reviewable.io/reviews/paranext/paranext-core/2685#-P-KC7FUDhCEYME-cmbw:-P-LIPBk4vRzwrY3Q88Z:b-896fix) ([raw file](https://github.com/paranext/paranext-core/blob/4534a43a68673059056d042f6b9d6604c0286454/extensions/src/platform-scripture/src/find.web-view.tsx#L786)):*

### The new give-up path can fire a false "Lost connection" on every ordinary new search

`retrieveFindJobUpdate` returns `undefined` for three different reasons (find.web-view.tsx:525): no `findPdp`, **no `activeJobIdRef.current`**, or a thrown PDP call. The new miss counter treats all three as "connection blip".

`abandonFindJob` clears `activeJobIdRef.current` _synchronously_ (find.web-view.tsx:465) and only calls `setActiveJobId(undefined)` **after** awaiting the PDP round-trip (:469). Every new search awaits `abandonFindJob()` first (:627). So during that await the old poll effect is still mounted with `isEffectActive === true`, gets `undefined` back every 100ms, and after 10 misses (~1s) sets `searchStatus: 'errored'` + "Lost connection while searching." Same shape when `findPdp` goes momentarily undefined on a project switch.

Before this change that path returned silently, which is exactly why it was invisible. Suggest making the sentinel distinguishable — e.g. have `retrieveFindJobUpdate` return `'no-job'` vs `undefined`, or guard the miss counter with `if (!activeJobIdRef.current || !findPdp) return;` before `nextPollMissState`. Worth a test: abandon-then-poll must not produce an error state.

_(Reported by Claude)_
</blockquote></details>

Done.
___
*[`extensions/src/platform-scripture/src/find.web-view.tsx` line 789 at r1](https://reviewable.io/reviews/paranext/paranext-core/2685#-P-KEHycDHoK4d-Aq20V:-P-LIQ774OGk6fmTrfwR:b-896fix) ([raw file](https://github.com/paranext/paranext-core/blob/4534a43a68673059056d042f6b9d6604c0286454/extensions/src/platform-scripture/src/find.web-view.tsx#L789)):*

### giving up leaves the find job running

When the loop gives up it stops polling but never calls `abandonFindJob()`/`stopFindJob()`, so the backend job keeps scanning and stays tracked. (The next search does abandon it, but only if the user starts one.) Consider abandoning on give-up.
</blockquote></details>

Done.
___
*[`extensions/src/platform-scripture/src/find/find.component.tsx` line 177 at r1](https://reviewable.io/reviews/paranext/paranext-core/2685#-P-KEAUS95MZzontaDGS:-P-LIV0a3K6zQ3SGkvNX:b-896fix) ([raw file](https://github.com/paranext/paranext-core/blob/4534a43a68673059056d042f6b9d6604c0286454/extensions/src/platform-scripture/src/find/find.component.tsx#L177)):*

### `isEditable?: boolean` optional with a permissive default

`isEditable = true` means any future call site that forgets the prop silently re-enables Replace, i.e. the failure mode is "allows a mutation". `isSearchQueryValid` was added as required in the same PR. Making `isEditable` required (or defaulting it `false`) matches the fail-closed intent stated in the webview comment.
</blockquote></details>

Done.
___
*[`extensions/src/platform-scripture/src/find/find.component.tsx` line 479 at r1](https://reviewable.io/reviews/paranext/paranext-core/2685#-P-KCVm477aPej0XEqhL:-P-LIiLDBJFqW4x5Bp0Y:b-896fix) ([raw file](https://github.com/paranext/paranext-core/blob/4534a43a68673059056d042f6b9d6604c0286454/extensions/src/platform-scripture/src/find/find.component.tsx#L479)):*

### The new `'skeleton'` state is unbounded, so it can reproduce the exact PRD failure in a new form

`resultsAreaState` returns `'skeleton'` whenever `searchStatus === undefined` and the query is valid. That's the right fix for the debounce flash, but nothing bounds it: if `findPdp` never becomes available, `handleStartSearch` bails, `pendingSearchDesiredRef` is set, and the readiness-retry effect (find.web-view.tsx:984) waits on `[findPdp]` forever. The user sees pulsing skeletons indefinitely with no error — a silent, dishonest loading state, which is what NN 1d/5a are about. The PR put a bound on the _poll_ wait but not on the _start_ wait. Suggest the same treatment: after N seconds without `findPdp`, surface the error placeholder.

_(Reported by Claude)_
</blockquote></details>

Done.
___
*[`extensions/src/platform-scripture/src/find/find.component.tsx` line 530 at r1](https://reviewable.io/reviews/paranext/paranext-core/2685#-P-KDG5D3Xv7aw72zedC:-P-LJ76e4lAvMuvUawmx:b-896fix) ([raw file](https://github.com/paranext/paranext-core/blob/4534a43a68673059056d042f6b9d6604c0286454/extensions/src/platform-scripture/src/find/find.component.tsx#L530)):*

###  find.component.tsx:530 vs :636 — read-only feedback is weaker than the pattern it says it mirrors

Structure protection surfaces **two** things: a persistent inline note (`%webView_find_replace_structureProtectedNote%`, :636) _and_ a hover tooltip. Read-only gets only the tooltip. The PR description says this "mirrors the existing structure-protected tooltip pattern", but it mirrors half of it. A hover-only affordance is the weakest form of feedback for the low-tech-preference persona this PRD is written for. Suggest an analogous `%webView_find_replace_readOnlyNote%` rendered in the same slot when `!isEditable`.

_(Reported by Claude)_
</blockquote></details>

Done.
___
*[`extensions/src/platform-scripture/src/find/find.component.tsx` line 921 at r1](https://reviewable.io/reviews/paranext/paranext-core/2685#-P-KCyI9DYUx8xKBYuxo:-P-LJ7rKA14v3dZzzjlU:b-896fix) ([raw file](https://github.com/paranext/paranext-core/blob/4534a43a68673059056d042f6b9d6604c0286454/extensions/src/platform-scripture/src/find/find.component.tsx#L921)):*

### find.component.tsx:921 / search-result.component.tsx:291— per-result Replace is now disabled with no explanation, which only half-satisfies NN 1d

Closing the bypass is the right call. But the toolbar buttons get `DisabledActionTooltip` (which also supplies the `aria-label` while disabled), and the per-result button gets a bare `disabled` with no tooltip and no accessible reason. NN 1d says "When replace is not allowed there is feedback to the user" — a silently greyed button in the result card isn't feedback, especially for Saroj. Pass `replaceBlockedTooltipText` down alongside `isReplaceBlocked` and wrap that button in the same `DisabledActionTooltip`, so the pattern is consistent in both places.

_(Reported by Claude)_
</blockquote></details>

Done.

jolierabideau (approved):
> @jolierabideau reviewed 9 files and all commit messages, and resolved 7 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2685)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2686, opened 2026-08-17
PT-4337: Implement analytics abstraction layer

tombogle:
# Code Review Summary

**Branch**: pt-4337-analytics-abstraction-layer

**Base**: origin/main

**Date**: 2026-08-19

**Files changed**: 8 (`.context/standards/Architecture-Decisions.md`, `README.md`, `src/extension-host/extension-host.ts`, `src/extension-host/services/analytics.service.ts` + test, `src/extension-host/services/analytics-providers/console-analytics.provider.ts` + test, `src/shared/models/analytics.model.ts`; `package-lock.json`'s diff is local worktree-name noise and is excluded)

## Overview

PT-4337 adds a provider-agnostic analytics abstraction layer, hosted in the extension-host. It resolves once per session whether analytics events target the "test" or "production" vendor audience, based on packaged/dev build type and the Send/Receive server target (read from the `paratextRegistration` extension's `InternetSettingsDataProvider`). Events fired before that resolution completes are queued and flushed once it settles. `ConsoleAnalyticsProvider` (logs instead of transmitting) is currently the only provider, wired for both the test and production environments — a real vendor provider is future work, so this class is not test-only yet; it is what any packaged build's production channel runs today.

This review covers a batch of fixes made in response to prior PR review feedback: renaming the engine file to match the codebase's `*.service.ts` vs. `*.service-host.ts` convention, redesigning the S/R-server-target lookup to wait event-driven via `networkObjectStatusService.waitForNetworkObject()` (replacing a fixed-timeout race) with three distinct, separately-logged failure modes, splitting `ConsoleAnalyticsProvider`'s logging into a safe `info`-level summary and a `debug`-level full-event line (to keep caller-supplied `properties` — which tends to accumulate user/project identifiers — out of a packaged build's persistent `info`-level log file), renaming the test-override env var to `PT_ANALYTICS_TEST_OVERRIDE`, fixing a `drainQueue()` falsy-item bug, and adding missing TSDoc. One additional fix (an `S-R` → `S/R` hyphenation inconsistency) was made during this review.

## API Changes

None. This branch touches only `src/extension-host/` (extension-host-internal service code) and `src/shared/models/analytics.model.ts` (a cross-process internal model, not a `@papi/` module or `.d.ts` file) — no `lib/platform-bible-react/`, `lib/platform-bible-utils/`, `papi.d.ts`, or extension `.d.ts` surfaces changed.

None.

None.

### Minor — Consider

- [x] **`analytics.service.ts:15` used `S-R` (hyphenated) once, inconsistent with the `S/R` form used everywhere else in the file, its test, and the ADR** _(fixed during review: changed `build/S-R target` to `build/S/R target`)_

Author response: no discussion needed — trivial, fixed immediately during the interview.

#### Shared Regions Modified

None.

#### Extension Config Changes

None — no files under `extensions/` changed in this branch.

## Positive Observations

- `getSelectedServer()`'s move from a fixed-timeout race to event-driven `networkObjectStatusService.waitForNetworkObject()` was verified correct against `network-object-status.service.ts`'s actual subscribe-before-check ordering, and each of the three failure points (registration timeout, provider vanished after registering, `getInternetSettings` read failure) is caught, logged with a distinct greppable message, and independently tested — including a fake-timers test proving the wait resolves as soon as the object appears rather than paying the full 30s budget.
- `drainQueue`'s `next !== undefined` check (not a truthiness check) correctly guards against falsy-but-real queue items, with a comment explaining why — a real fix, not cosmetic.
- `ConsoleAnalyticsProvider.send()`'s info/debug log-level split is a real safeguard against PII-bearing `properties` landing in a packaged build's log; both the provider test and the service test explicitly assert `properties` values are absent from the `info` line and present only at `debug`, and both directions of the environment-mismatch guard are covered.
- The `analytics.service-host.ts` → `analytics.service.ts` rename correctly matches the codebase's actual split: `*.service-host.ts` is reserved for modules that register a PAPI network object/data-provider engine paired with a cross-process facade; this engine registers nothing and has no such pair, matching siblings like `local-oauth.service.ts` and `contribution.service.ts`. No stale references to the old filename remain.
- `INTERNET_SETTINGS_DATA_PROVIDER_ID` and the `selectedServer === 'Production'` comparison were verified against `paratext-registration.d.ts` — both the data-provider ID and the `ServerType` union match exactly.
- `extension-host.ts` correctly does not `await` `analyticsService.initialize()`, so a slow S/R-target lookup (up to 30s) cannot gate extension-host startup.
- Test coverage is thorough across both files: override env var exact-match semantics, dev-vs-packaged resolution, all three S/R-lookup failure modes, non-serializable property sanitization (circular refs, functions, symbols, `undefined`), and a failing-provider-send path confirming `initialize()`/`trackEvent()` never throw.
- ADR-0013 explicitly records the one incidental core→extension data-provider dependency (via `tsconfig.json`'s `typeRoots`) and states plainly that it's a one-off, not a sanctioned general pattern — exactly the kind of judgment call worth writing down.
- `createCachedInitializer` is reused rather than reimplemented for resolve-once semantics, following the codebase's established stateful-service pattern.
- The `TODO(PT-4374)` in `flushQueue` is a well-documented, deliberate scope boundary tied to a tracked follow-on ticket, not an unplanned stub.

## Interview Notes

This review covered a batch of fixes made in direct response to prior human PR review comments on #2686 (naming convention, timeout/race redesign, PII-in-logs, env var naming, `drainQueue` bug, missing TSDoc), verified individually against the code before implementation throughout the review-response process. Zero Critical/Important findings surfaced in this pass, so Step 3.2/3.3 (finding walkthrough, design explanation) were skipped per the empty-findings path. The one Minor finding (an `S-R`/`S/R` hyphenation inconsistency) was fixed immediately without discussion.

One point worth surfacing for the reviewer: `ConsoleAnalyticsProvider` is currently wired as *both* the test and production provider (no real vendor provider exists yet), so the info/debug log-split fix is live behavior in production packaged builds today, not precautionary future-proofing — this was discussed and confirmed during the session.

No areas of author uncertainty or AI-deferral to report.

## In-Review Quality Check

One in-review fix was made (the `S-R` → `S/R` typo in `analytics.service.ts`). Quality checks after the fix:

- `npm run typecheck` — passed cleanly, no errors.
- `npx vitest run` on `analytics.service.test.ts` and `console-analytics.provider.test.ts` — all 19 tests passed.
- `npx prettier --check` on the changed file — already correctly formatted, no changes needed.

No unfixable failures.

## Suggested Review Focus

- [ ] Confirm the `S/R`-server-target dependency direction (core's `analytics.service.ts` reading the `paratextRegistration` extension's data provider) is acceptable as a documented one-off (see ADR-0013's addendum) rather than a pattern to generalize.
- [ ] Confirm the info/debug log-level split in `ConsoleAnalyticsProvider` is the right boundary — note that this provider is currently live for both test and production channels, not test-only, since no real vendor provider exists yet.
- [ ] Spot-check the three-failure-mode logging in `getSelectedServer()` (timeout vs. vanished vs. read failure) against real startup logs if convenient, to confirm the messages are as diagnostically useful in practice as intended.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2686)


jolierabideau (changes requested):
Looks great, Tom! Reported some Claude findings. 

> @jolierabideau reviewed 7 files and all commit messages, and made 14 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2686#-:-P-G6aYdEEDJSjhOTfJL:b-tm9gde)* status: all files reviewed, 13 unresolved discussions (waiting on tombogle).

___
*[`src/extension-host/extension-host.ts` line 97 at r1](https://reviewable.io/reviews/paranext/paranext-core/2686#-P-G4mFV86BT1glgL2YY:-P-G4mFW850Tupn8D2jX:b-uiyz9k) ([raw file](https://github.com/paranext/paranext-core/blob/a327cc42ccf391c877492c0ba356c5d5496e83c3/src/extension-host/extension-host.ts#L97)):*
> ```XML
>     // ENVIRONMENT_RESOLUTION_TIMEOUT_MS if the Send/Receive server-target lookup is slow.
>     // trackEvent()'s `unresolved` queue tolerates this finishing at any point.
>     analyticsService.initialize().catch((error) => {
> ```

NIT This `.catch()` is unreachable. `resolveEnvironment()` can't reject: `waitForDuration`'s timeout branch always _fulfills_, so `Promise.any` never produces an `AggregateError`, and nothing else in `initialize`'s body throws. Harmless, but the comment above reads as if real failure handling exists. Fine to keep as a guard — just say that's what it is.

_(Reported by Claude)_
___
*[`src/extension-host/services/analytics.service-host.ts` at r1](https://reviewable.io/reviews/paranext/paranext-core/2686#-P-G2o521d3iV9s7d40x:-P-G2o521d3iV9s7d40y:b-jnamd3) ([raw file](https://github.com/paranext/paranext-core/blob/a327cc42ccf391c877492c0ba356c5d5496e83c3/src/extension-host/services/analytics.service-host.ts)):*
**`*.service-host.ts` means something specific in this repo** and this module isn't it. Every existing one — `database`, `localization`, `menu-data`, `settings`, `project-settings`, `theme-data` — is the host half of a cross-process service, paired with a `*.service.ts` + `*.service-model.ts` in `src/shared/services/` and registering a network object. This module has neither and is extension-host-local by design (per the ADR). The matching precedent for extension-host-local modules is `local-oauth.service.ts` / `extension-storage.service.ts` — plain `*.service.ts`.

Suggest renaming to `analytics.service.ts` now and reserving `-service-host` for when the cross-process facade the ADR describes actually lands. The ADR text at [Architecture-Decisions.md:57](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/.context/standards/Architecture-Decisions.md#L57) already says main and renderer can't call it, which is the tell.

_(Reported by Claude)_
___
*[`src/extension-host/services/analytics.service-host.ts` line 15 at r1](https://reviewable.io/reviews/paranext/paranext-core/2686#-P-G6CBE8kk9dzqqCkwv:-P-G6CBE8kk9dzqqCkww:by60h07) ([raw file](https://github.com/paranext/paranext-core/blob/a327cc42ccf391c877492c0ba356c5d5496e83c3/src/extension-host/services/analytics.service-host.ts#L15)):*
> ```XML
> 
> /** Env var that forces analytics to target the test environment regardless of build/S-R target. */
> const ANALYTICS_TEST_OVERRIDE_ENV_VAR = 'PARATEXT_ANALYTICS_TEST_OVERRIDE';
> ```

NIT `PARATEXT_ANALYTICS_TEST_OVERRIDE` — no other `PARATEXT_*` env var exists in the repo; the prefix convention for new app-specific vars is `PT_` (`PT_STARTUP_MARKS`). Also it's undocumented — `PT_STARTUP_MARKS` has a README section, and a QA-facing override like this needs one more than most.

_(Reported by Claude)_
___
*[`src/extension-host/services/analytics.service-host.ts` line 36 at r1](https://reviewable.io/reviews/paranext/paranext-core/2686#-P-G1e850hOX38LCON4H:-P-G1e850hOX38LCON4I:b-l5x3yt) ([raw file](https://github.com/paranext/paranext-core/blob/a327cc42ccf391c877492c0ba356c5d5496e83c3/src/extension-host/services/analytics.service-host.ts#L36)):*
> ```XML
>   test: [],
>   production: [],
>   unresolved: [],
> ```

## (also `src/extension-host/extension-host.ts:97`)

**This will resolve to `'test'` in production essentially every time.** `dataProviderService.get()` does _not_ wait for a network object — it calls `networkObjectService.get()`, which returns `undefined` immediately when the object isn't registered (network-object.service.ts:539), logging `No data provider found with name = …`). `analyticsService.initialize()` is fired at extension-host.ts:97, right after `networkService.initialize()` and well before `extensionService.initialize()` at extension-host.ts:113 — so the `paratext-registration` extension has not activated and has not registered `paratextRegistration.internetSettingsDataProvider` yet. `getSelectedServer()` returns `undefined` on the first tick, and line 53 falls back to `'test'`.

Two knock-on effects: `ENVIRONMENT_RESOLUTION_TIMEOUT_MS` is effectively dead (nothing ever takes 5s — `get` resolves immediately), and the warning at line 54 says "could not resolve … in time" when nothing timed out.

**The repo already has the pattern for this**: `networkObjectStatusService.waitForNetworkObject({ id }, timeoutMs)`, used at web-view.service.ts:23 and project-lookup.service-model.ts:208 for exactly this "the object isn't up yet" case. That's also where the 5s budget belongs. 

_(Reported by Claude)_
___
*[`src/extension-host/services/analytics.service-host.ts` line 46 at r1](https://reviewable.io/reviews/paranext/paranext-core/2686#-P-G4Bjr3XHGN8XAZweU:-P-G4Bjr3XHGN8XAZweV:b28koqj) ([raw file](https://github.com/paranext/paranext-core/blob/a327cc42ccf391c877492c0ba356c5d5496e83c3/src/extension-host/services/analytics.service-host.ts#L46)):*
> ```XML
> let resolvedEnvironment: AnalyticsEnvironment | undefined;
> 
> async function getSelectedServer(): Promise<string | undefined> {
> ```

Return type is `Promise<string | undefined>`, which widens `ServerType` (`'Production' | 'QualityAssurance' | 'Development' | 'Test'`, paratext-registration.d.ts:32) to `string`. That throws away the compile-time check on line 60's `=== 'Production'` — a typo there silently downgrades everyone to `'test'` forever. Use `ServerType | undefined`.

_(Reported by Claude)_
___
*[`src/extension-host/services/analytics.service-host.ts` line 54 at r1](https://reviewable.io/reviews/paranext/paranext-core/2686#-P-G39mAArWezbb8qjSV:-P-G39mAArWezbb8qjSW:b-67hn28) ([raw file](https://github.com/paranext/paranext-core/blob/a327cc42ccf391c877492c0ba356c5d5496e83c3/src/extension-host/services/analytics.service-host.ts#L54)):*
> ```XML
>   try {
>     const provider = await dataProviderService.get(
>       'paratextRegistration.internetSettingsDataProvider',
> ```

NIT Inline data-provider-name string literal. Every other `dataProviderService.get()` call in core uses a named constant — `themeServiceDataProviderName`, `localizationServiceProviderName`, `settingsServiceDataProviderName`, `menuDataServiceProviderName`. Extract one (and it'd give the `waitForNetworkObject` call above a single source for the id).

Related and worth calling out for the reviewer meeting: this is the **first dependency from core onto an extension-provided data provider** I can find — everything else core reaches for is core-owned. It typechecks because `tsconfig.json` puts `./extensions/src` on `typeRoots`, but it's a new architectural direction and the ADR doesn't mention it as a considered consequence.

_(Reported by Claude)_
___
*[`src/extension-host/services/analytics.service-host.ts` line 73 at r1](https://reviewable.io/reviews/paranext/paranext-core/2686#-P-G4ZMl0K4EHZq-GZLZ:-P-G4ZMl0K4EHZq-GZL_:b1pluiw) ([raw file](https://github.com/paranext/paranext-core/blob/a327cc42ccf391c877492c0ba356c5d5496e83c3/src/extension-host/services/analytics.service-host.ts#L73)):*
> ```XML
>   if (!globalThis.isPackaged || process.env.NODE_ENV !== 'production') return 'test';
> 
>   const selectedServer = await waitForDuration(
> ```

`waitForDuration` is `Promise.any([timeout, fn()])` (util.ts:240) — `Promise.any` ignores rejections, so if `getSelectedServer()` **throws** you wait the full 5s rather than failing fast. And three distinct outcomes collapse into the same `undefined`: provider missing, `settings.selectedServer` genuinely undefined, and real timeout. The warning text asserts the third. Worth distinguishing them, at minimum so the log tells you which one you're in.

_(Reported by Claude)_
___
*[`src/extension-host/services/analytics.service-host.ts` line 101 at r1](https://reviewable.io/reviews/paranext/paranext-core/2686#-P-G6-uFDSPkO_AEhHLA:-P-G6-uFDSPkO_AEhHLB:b-8joki9) ([raw file](https://github.com/paranext/paranext-core/blob/a327cc42ccf391c877492c0ba356c5d5496e83c3/src/extension-host/services/analytics.service-host.ts#L101)):*
> ```XML
>   Object.entries(properties).forEach(([key, value]) => {
>     try {
>       JSON.stringify(value);
> ```

NIT `JSON.stringify` returns `undefined` without throwing for functions, symbols, and `undefined`. Those keys pass sanitization, then get silently dropped when the whole event is stringified — the opposite of what the doc comment on line 63 promises ("warning immediately rather than waiting for the event to reach a provider"). Check `JSON.stringify(value) === undefined` too.

_(Reported by Claude)_
___
*[`src/extension-host/services/analytics.service-host.ts` line 118 at r1](https://reviewable.io/reviews/paranext/paranext-core/2686#-P-G5jeG9sBAmtUYaxZG:-P-G5jeG9sBAmtUYaxZH:bpiut5a) ([raw file](https://github.com/paranext/paranext-core/blob/a327cc42ccf391c877492c0ba356c5d5496e83c3/src/extension-host/services/analytics.service-host.ts#L118)):*
> ```XML
>  */
> function drainQueue<T>(queue: T[], onItem: (item: T) => void): void {
>   let next = queue.shift();
> ```

NIT Events are `shift()`ed off the queue before the async `send` settles, so a failed send drops the event permanently — the `.catch()` logs and that's it. Reasonable for fire-and-forget, but the ADR promises a durable cross-restart queue as a follow-on (Architecture-Decisions.md:58-61), and drain-on-success is the part that gets retrofitted. A `TODO(PT-xxxx)` here would mark it.

_(Reported by Claude)_
___
*[`src/extension-host/services/analytics.service-host.ts` line 119 at r1](https://reviewable.io/reviews/paranext/paranext-core/2686#-P-G4z438yue7cKS-PT4:-P-G4z438yue7cKS-PT5:b1hpea4) ([raw file](https://github.com/paranext/paranext-core/blob/a327cc42ccf391c877492c0ba356c5d5496e83c3/src/extension-host/services/analytics.service-host.ts#L119)):*
> ```XML
> function drainQueue<T>(queue: T[], onItem: (item: T) => void): void {
>   let next = queue.shift();
>   while (next) {
> ```

NIT `while (next)` exits on any falsy item. Safe for the two object queues today, but this is written as a generic `<T>` helper, so `drainQueue([0, 1, 2], …)` would silently drop everything after the first element. `while (next !== undefined)` costs nothing. Also, the TSDoc on line 88 says `onItem` receives "a per-iteration copy" — it's a per-iteration _binding_, not a copy; that wording will mislead someone later.

_(Reported by Claude)_
___
*[`src/extension-host/services/analytics.service-host.ts` line 125 at r1](https://reviewable.io/reviews/paranext/paranext-core/2686#-P-G3zKI1h6EUM3RQwA6:-P-G3zKI1h6EUM3RQwA7:bbaaiw6) ([raw file](https://github.com/paranext/paranext-core/blob/a327cc42ccf391c877492c0ba356c5d5496e83c3/src/extension-host/services/analytics.service-host.ts#L125)):*
> ```XML
> }
> 
> function flushQueue(environment: AnalyticsEnvironment): void {
> ```

NIT **No TSDoc on the two exported functions.** `initialize` and `trackEvent` are the entire public surface of this abstraction and have none, while the private helpers `sanitizeProperties` (line 63) and `drainQueue` (line 87) get thorough doc comments. The Code Style Guide requires TSDoc on the API surface; `trackEvent` in particular needs `@param name` / `@param properties` plus the fire-and-forget + queueing contract stated where a call site will actually see it.

Same for analytics.model.ts:31 — `send` has no `@param`/`@returns`, and implementers need to know whether they're allowed to reject.

_(Reported by Claude)_
___
*[`src/extension-host/services/analytics.service-host.test.ts` line 81 at r1](https://reviewable.io/reviews/paranext/paranext-core/2686#-P-G2Szl9KgfDhyutho6:-P-G2Szl9KgfDhyutho7:b-pbnorl) ([raw file](https://github.com/paranext/paranext-core/blob/a327cc42ccf391c877492c0ba356c5d5496e83c3/src/extension-host/services/analytics.service-host.test.ts#L81)):*
> ```XML
>     getInternetSettings: vi.fn().mockResolvedValue({ selectedServer: 'QualityAssurance' }),
>   });
> ```

The tests mock `dataProviderService.get` to resolve with a provider, so the one scenario that actually happens in production — _the data provider is not registered yet_ — is never exercised. That's why the bug above is green. Worth a test: `mocks.get.mockResolvedValue(undefined)` in a packaged production build, asserting whatever the intended behavior is (wait, then resolve — not immediately `'test'`). 

_(Reported by Claude)_
___
*[`src/extension-host/services/analytics-providers/console-analytics.provider.ts` line 33 at r1](https://reviewable.io/reviews/paranext/paranext-core/2686#-P-G4jL05NRzX9aLByxJ:-P-G4jL05NRzX9aLByxK:bk3512m) ([raw file](https://github.com/paranext/paranext-core/blob/a327cc42ccf391c877492c0ba356c5d5496e83c3/src/extension-host/services/analytics-providers/console-analytics.provider.ts#L33)):*
> ```XML
> 
>     const label = this.environment === 'production' ? 'Production' : 'Test';
>     logger.info(`${label}: ${JSON.stringify(event)}`);
> ```

Logs the full serialized event — including arbitrary caller-supplied `properties` — at `info`, in packaged builds too. Analytics properties are exactly the payload that tends to accumulate user/project identifiers over the life of an epic. Consider `debug`, or a note in the class TSDoc that properties reaching this provider land in the user's log file verbatim.

_(Reported by Claude)_

tombogle (commented):
> @tombogle made 13 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2686)* status: 0 of 10 files reviewed, 13 unresolved discussions (waiting on jolierabideau and tjcouch-sil).

___
*[`src/extension-host/extension-host.ts` line 97 at r1](https://reviewable.io/reviews/paranext/paranext-core/2686#-P-G4mFV86BT1glgL2YY:-P-KuLj5EiL7ZBdjDro0:b-bttoee) ([raw file](https://github.com/paranext/paranext-core/blob/a327cc42ccf391c877492c0ba356c5d5496e83c3/src/extension-host/extension-host.ts#L97)):*

NIT This `.catch()` is unreachable. `resolveEnvironment()` can't reject: `waitForDuration`'s timeout branch always _fulfills_, so `Promise.any` never produces an `AggregateError`, and nothing else in `initialize`'s body throws. Harmless, but the comment above reads as if real failure handling exists. Fine to keep as a guard — just say that's what it is.

_(Reported by Claude)_
</blockquote></details>

Added comment
___
*[`src/extension-host/services/analytics-providers/console-analytics.provider.ts` line 33 at r1](https://reviewable.io/reviews/paranext/paranext-core/2686#-P-G4jL05NRzX9aLByxJ:-P-MH75s59KTJp99sSqb:b-a8qlpy) ([raw file](https://github.com/paranext/paranext-core/blob/a327cc42ccf391c877492c0ba356c5d5496e83c3/src/extension-host/services/analytics-providers/console-analytics.provider.ts#L33)):*

Logs the full serialized event — including arbitrary caller-supplied `properties` — at `info`, in packaged builds too. Analytics properties are exactly the payload that tends to accumulate user/project identifiers over the life of an epic. Consider `debug`, or a note in the class TSDoc that properties reaching this provider land in the user's log file verbatim.

_(Reported by Claude)_
</blockquote></details>

Split level of detail to be more sparse when writing to `info`. Note: Once a real vendor provider lands and gets wired in for 'production', this class's own TSDoc says it becomes test/E2E-only, and the concern mostly evaporates.
___
*[`src/extension-host/services/analytics.service-host.ts` line 15 at r1](https://reviewable.io/reviews/paranext/paranext-core/2686#-P-G6CBE8kk9dzqqCkwv:-P-L1WLtAD3lbxWt8RQG:b-r6j0bp) ([raw file](https://github.com/paranext/paranext-core/blob/a327cc42ccf391c877492c0ba356c5d5496e83c3/src/extension-host/services/analytics.service-host.ts#L15)):*

NIT `PARATEXT_ANALYTICS_TEST_OVERRIDE` — no other `PARATEXT_*` env var exists in the repo; the prefix convention for new app-specific vars is `PT_` (`PT_STARTUP_MARKS`). Also it's undocumented — `PT_STARTUP_MARKS` has a README section, and a QA-facing override like this needs one more than most.

_(Reported by Claude)_
</blockquote></details>

Renamed, tested, commented, and documented
___
*[`src/extension-host/services/analytics.service-host.ts` line 36 at r1](https://reviewable.io/reviews/paranext/paranext-core/2686#-P-G1e850hOX38LCON4H:-P-LCnGC5vnMVaK2gFf5:b-g97zax) ([raw file](https://github.com/paranext/paranext-core/blob/a327cc42ccf391c877492c0ba356c5d5496e83c3/src/extension-host/services/analytics.service-host.ts#L36)):*

## (also `src/extension-host/extension-host.ts:97`)

**This will resolve to `'test'` in production essentially every time.** `dataProviderService.get()` does _not_ wait for a network object — it calls `networkObjectService.get()`, which returns `undefined` immediately when the object isn't registered (network-object.service.ts:539), logging `No data provider found with name = …`). `analyticsService.initialize()` is fired at extension-host.ts:97, right after `networkService.initialize()` and well before `extensionService.initialize()` at extension-host.ts:113 — so the `paratext-registration` extension has not activated and has not registered `paratextRegistration.internetSettingsDataProvider` yet. `getSelectedServer()` returns `undefined` on the first tick, and line 53 falls back to `'test'`.

Two knock-on effects: `ENVIRONMENT_RESOLUTION_TIMEOUT_MS` is effectively dead (nothing ever takes 5s — `get` resolves immediately), and the warning at line 54 says "could not resolve … in time" when nothing timed out.

**The repo already has the pattern for this**: `networkObjectStatusService.waitForNetworkObject({ id }, timeoutMs)`, used at web-view.service.ts:23 and project-lookup.service-model.ts:208 for exactly this "the object isn't up yet" case. That's also where the 5s budget belongs. 

_(Reported by Claude)_
</blockquote></details>

Fully addressed: `getSelectedServer()` now uses `networkObjectStatusService.waitForNetworkObject()` (event-driven, matches `web-view.service.ts`'s established pattern) instead of a blind timeout race, dropped the `rpc.model.ts`/`waitForDuration` dependency entirely, and all 12 tests pass reflecting the new mechanism.
___
*[`src/extension-host/services/analytics.service-host.ts` line 46 at r1](https://reviewable.io/reviews/paranext/paranext-core/2686#-P-G4Bjr3XHGN8XAZweU:-P-LMdibAhfb8Bb4tV8o:b-886gsx) ([raw file](https://github.com/paranext/paranext-core/blob/a327cc42ccf391c877492c0ba356c5d5496e83c3/src/extension-host/services/analytics.service-host.ts#L46)):*

Return type is `Promise<string | undefined>`, which widens `ServerType` (`'Production' | 'QualityAssurance' | 'Development' | 'Test'`, paratext-registration.d.ts:32) to `string`. That throws away the compile-time check on line 60's `=== 'Production'` — a typo there silently downgrades everyone to `'test'` forever. Use `ServerType | undefined`.

_(Reported by Claude)_
</blockquote></details>

Fixed
___
*[`src/extension-host/services/analytics.service-host.ts` line 54 at r1](https://reviewable.io/reviews/paranext/paranext-core/2686#-P-G39mAArWezbb8qjSV:-P-LNrBi84aooTuPumWT:bfz9la9) ([raw file](https://github.com/paranext/paranext-core/blob/a327cc42ccf391c877492c0ba356c5d5496e83c3/src/extension-host/services/analytics.service-host.ts#L54)):*

NIT Inline data-provider-name string literal. Every other `dataProviderService.get()` call in core uses a named constant — `themeServiceDataProviderName`, `localizationServiceProviderName`, `settingsServiceDataProviderName`, `menuDataServiceProviderName`. Extract one (and it'd give the `waitForNetworkObject` call above a single source for the id).

Related and worth calling out for the reviewer meeting: this is the **first dependency from core onto an extension-provided data provider** I can find — everything else core reaches for is core-owned. It typechecks because `tsconfig.json` puts `./extensions/src` on `typeRoots`, but it's a new architectural direction and the ADR doesn't mention it as a considered consequence.

_(Reported by Claude)_
</blockquote></details>

Confirmed — every other `dataProviderService.get()` call site in core `src/` targets a core-owned provider via a constant from a shared `*.service-model.ts`/`*.service.ts` file. This genuinely is the first core→extension-provider dependency. The "extract one" NIT is already satisfied (my `INTERNET_SETTINGS_DATA_PROVIDER_ID` constant is used as the single source for both the `waitForNetworkObject` and `dataProviderService.get` calls) — there's no extension-exported runtime constant to import instead, since `.d.ts` files only carry ambient types, not runtime values.

The architectural point deserves recording, not a code change — there's no cleaner alternative (the S/R target genuinely only exists via this provider), and the reviewer explicitly framed it as a discussion point.

@tjcouch-sil , please review this: Added a "New precedent, flagged during PR review" note to ADR-0013's Consequences section, documenting this as the first core→extension data-provider dependency, why it's not a broader precedent, and why no alternative was built.
___
*[`src/extension-host/services/analytics.service-host.ts` line 73 at r1](https://reviewable.io/reviews/paranext/paranext-core/2686#-P-G4ZMl0K4EHZq-GZLZ:-P-M4kSCEg9VJlZVK5ih:bcn9t2d) ([raw file](https://github.com/paranext/paranext-core/blob/a327cc42ccf391c877492c0ba356c5d5496e83c3/src/extension-host/services/analytics.service-host.ts#L73)):*

`waitForDuration` is `Promise.any([timeout, fn()])` (util.ts:240) — `Promise.any` ignores rejections, so if `getSelectedServer()` **throws** you wait the full 5s rather than failing fast. And three distinct outcomes collapse into the same `undefined`: provider missing, `settings.selectedServer` genuinely undefined, and real timeout. The warning text asserts the third. Worth distinguishing them, at minimum so the log tells you which one you're in.

_(Reported by Claude)_
</blockquote></details>

Done — `getSelectedServer` now logs a distinct debug message for each of the three failure causes (timeout waiting for registration, provider vanishing after being seen, or a failed read), and the outer warning no longer falsely claims "in time." 14/14 tests pass (4 new/updated).
___
*[`src/extension-host/services/analytics.service-host.ts` line 101 at r1](https://reviewable.io/reviews/paranext/paranext-core/2686#-P-G6-uFDSPkO_AEhHLA:-P-M7FqaEN2FrHFp9AiM:b-7kd2fp) ([raw file](https://github.com/paranext/paranext-core/blob/a327cc42ccf391c877492c0ba356c5d5496e83c3/src/extension-host/services/analytics.service-host.ts#L101)):*

NIT `JSON.stringify` returns `undefined` without throwing for functions, symbols, and `undefined`. Those keys pass sanitization, then get silently dropped when the whole event is stringified — the opposite of what the doc comment on line 63 promises ("warning immediately rather than waiting for the event to reach a provider"). Check `JSON.stringify(value) === undefined` too.

_(Reported by Claude)_
</blockquote></details>

Fixed and tested
___
*[`src/extension-host/services/analytics.service-host.ts` line 118 at r1](https://reviewable.io/reviews/paranext/paranext-core/2686#-P-G5jeG9sBAmtUYaxZG:-P-MD2u83OGl11s5oTys:b-iwmm5x) ([raw file](https://github.com/paranext/paranext-core/blob/a327cc42ccf391c877492c0ba356c5d5496e83c3/src/extension-host/services/analytics.service-host.ts#L118)):*

NIT Events are `shift()`ed off the queue before the async `send` settles, so a failed send drops the event permanently — the `.catch()` logs and that's it. Reasonable for fire-and-forget, but the ADR promises a durable cross-restart queue as a follow-on (Architecture-Decisions.md:58-61), and drain-on-success is the part that gets retrofitted. A `TODO(PT-xxxx)` here would mark it.

_(Reported by Claude)_
</blockquote></details>

Added TODO
___
*[`src/extension-host/services/analytics.service-host.ts` line 119 at r1](https://reviewable.io/reviews/paranext/paranext-core/2686#-P-G4z438yue7cKS-PT4:-P-MCSdMEqISW-pBc5lI:blbzacb) ([raw file](https://github.com/paranext/paranext-core/blob/a327cc42ccf391c877492c0ba356c5d5496e83c3/src/extension-host/services/analytics.service-host.ts#L119)):*

NIT `while (next)` exits on any falsy item. Safe for the two object queues today, but this is written as a generic `<T>` helper, so `drainQueue([0, 1, 2], …)` would silently drop everything after the first element. `while (next !== undefined)` costs nothing. Also, the TSDoc on line 88 says `onItem` receives "a per-iteration copy" — it's a per-iteration _binding_, not a copy; that wording will mislead someone later.

_(Reported by Claude)_
</blockquote></details>

Fixed. (Not tested, since the function is internal and that code path is unreachable thrrough exported members.)
___
*[`src/extension-host/services/analytics.service-host.ts` line 125 at r1](https://reviewable.io/reviews/paranext/paranext-core/2686#-P-G3zKI1h6EUM3RQwA6:-P-MEg_77Ba0FIUaR0VC:b13dwow) ([raw file](https://github.com/paranext/paranext-core/blob/a327cc42ccf391c877492c0ba356c5d5496e83c3/src/extension-host/services/analytics.service-host.ts#L125)):*

NIT **No TSDoc on the two exported functions.** `initialize` and `trackEvent` are the entire public surface of this abstraction and have none, while the private helpers `sanitizeProperties` (line 63) and `drainQueue` (line 87) get thorough doc comments. The Code Style Guide requires TSDoc on the API surface; `trackEvent` in particular needs `@param name` / `@param properties` plus the fire-and-forget + queueing contract stated where a call site will actually see it.

Same for analytics.model.ts:31 — `send` has no `@param`/`@returns`, and implementers need to know whether they're allowed to reject.

_(Reported by Claude)_
</blockquote></details>

Docs added
___
*[`src/extension-host/services/analytics.service-host.ts` at r1](https://reviewable.io/reviews/paranext/paranext-core/2686#-P-G2o521d3iV9s7d40x:-P-KuO6E26ZlPeIAFLK9:b-jc85r9) ([raw file](https://github.com/paranext/paranext-core/blob/a327cc42ccf391c877492c0ba356c5d5496e83c3/src/extension-host/services/analytics.service-host.ts)):*

**`*.service-host.ts` means something specific in this repo** and this module isn't it. Every existing one — `database`, `localization`, `menu-data`, `settings`, `project-settings`, `theme-data` — is the host half of a cross-process service, paired with a `*.service.ts` + `*.service-model.ts` in `src/shared/services/` and registering a network object. This module has neither and is extension-host-local by design (per the ADR). The matching precedent for extension-host-local modules is `local-oauth.service.ts` / `extension-storage.service.ts` — plain `*.service.ts`.

Suggest renaming to `analytics.service.ts` now and reserving `-service-host` for when the cross-process facade the ADR describes actually lands. The ADR text at [Architecture-Decisions.md:57](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/.context/standards/Architecture-Decisions.md#L57) already says main and renderer can't call it, which is the tell.

_(Reported by Claude)_
</blockquote></details>

renamed (though it will be renamed back as part of https://paratextstudio.atlassian.net/browse/PT-4396
___
*[`src/extension-host/services/analytics.service-host.test.ts` line 81 at r1](https://reviewable.io/reviews/paranext/paranext-core/2686#-P-G2Szl9KgfDhyutho6:-P-MFE1ZCx73PfCfMge4:bo6c791) ([raw file](https://github.com/paranext/paranext-core/blob/a327cc42ccf391c877492c0ba356c5d5496e83c3/src/extension-host/services/analytics.service-host.test.ts#L81)):*

The tests mock `dataProviderService.get` to resolve with a provider, so the one scenario that actually happens in production — _the data provider is not registered yet_ — is never exercised. That's why the bug above is green. Worth a test: `mocks.get.mockResolvedValue(undefined)` in a packaged production build, asserting whatever the intended behavior is (wait, then resolve — not immediately `'test'`). 

_(Reported by Claude)_
</blockquote></details>

Now covered by tests at lines 149-160

jolierabideau (approved):
> @jolierabideau reviewed 10 files and all commit messages, and resolved 13 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2686)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2689, opened 2026-08-17
PT-3362: Add project selector to the Find tab

captaincrazybro:
## Summary

Adds a `ProjectSelector` to the Find tab's header so the user can see and change which project Find
operates on.

**Find is the FIRST consumer of `ProjectSelector`'s `mode="projectScrollGroup"`.** Checks
(`checks-side-panel.component.tsx:222`) and Manage Books (`manage-books-sidebar.component.tsx:288`)
both use `mode="project"`. An earlier version of this description claimed they shared this mode —
that was wrong, and it matters for review: it is why `onOpenProjectInGroup` is a mandatory prop that
ends up a dead no-op here, and it means this mode's previously-unexercised paths ship first in this
PR. The picker lists only projects currently open in a scripture editor
tab; the selection is a `(projectId, scrollGroupId)` pair that a reassignment effect keeps valid as
tabs open and close. Clicking a result now scrolls the exact tab the selector has chosen, instead of
the pinned originating editor tab.

Screenshot of the new project selector in the find UI:
image

Screenshot of the project selector in simple mode:
image

### Why review this

Part of the Find work alongside `pt-4341-find-from-any-tab`, `pt-4342-dock-find-in-simple`, and
`pt-4343-find-lifecycle-feedback`.

## Changes

- Project picker in the Find header, sharing a row with the Find/Replace toggle; lists only open
  projects, with loading and "no open projects" states.
- Find's own scroll group now follows the picker, so search scope and the result-activation broadcast
  act on the selected tab rather than the tab Find was opened from.
- Result clicks scroll the selected tab deterministically, replacing reliance on the scroll-group
  toolbar.
- `ProjectSelector` gains one optional prop, `hideFilterMenu` (defaults `false`, so `checks-side-panel`
  and `manage-books` are unaffected).
- Robustness around switching projects: case-insensitive switch detection, detection of an
  unresolvable `findInScripture` PDP, `Promise.allSettled` for per-project name reads, metadata
  re-fetch when a project opens later, and `selectedBookIds` reconciled against the new project's
  `booksPresent`.
- Fixes a crash when the selected project's editor tab is closed (revoked-proxy `TypeError` from a
  disposed web-view controller). **This crash site predates this branch** — it exists at the
  merge-base — but the picker made it routinely reachable, so it is fixed here.
- Localizes the picker's internal strings and adds `metadata.fallbackKey` entries so Find, Checks, and
  Manage Books share translations.
- 23 new tests.

## AI Involvement

AI-assisted throughout. The human developer is the author; AI acted as generator and reviewer.

- The original implementation was AI-assisted and human-directed.
- A structured AI review pass (four parallel analyzers: API/correctness, style, coverage/compliance,
  UX) produced 12 important and 17 minor findings. The author triaged every one — accepting, deferring,
  or rejecting each; two were rejected on a factual premise that verification confirmed correct.
- Three mechanisms in this diff were **designed by the review rather than the author** and warrant
  extra scrutiny: the `findPdpAvailability` probe, the `hideFilterMenu` prop, and the
  `callControllerSafely` helper.
- **Two of the review's own fixes caused regressions**, both found by the author's manual testing, not
  by the review or the test suite. Both are fixed and documented in the embedded summary's
  "Corrections to this review" section.
- All 23 added tests were mutation-verified (each fix reverted to confirm the intended test fails).
  That process caught one assertion that passed vacuously.
- Session URL was not captured for this session, so the commit lacks the usual `Session-URL:` trailer.

## Testing

- [x] `npm run lint` — exit 0 (repo-wide, the command CI runs)
- [x] `npm test` — exit 0; 409 test files, 4874 tests, 1 pre-existing skip
- [x] `npm run build` (full, incl. .NET) and `npm run build:extensions` — exit 0
- [x] `npm run build:types` — no diff to `papi.d.ts`
- [x] Prettier `--check` clean on all changed files
- [x] Manual verification in a running app (power mode): project switching, tab open/close
      reassignment, empty-state placeholder, crash repro confirmed fixed
- [ ] `dotnet test c-sharp-tests/` — not run; no C# files touched

## Risk Level

**Medium.** No public PAPI surface changes and `papi.d.ts` is unchanged, but: it adds a prop to a
shared `lib/platform-bible-react` component, it changes Find's scroll-group semantics, and the
project-switch path is effect-ordering-sensitive — that path produced three bugs during development,
two of them found only by manual testing. Coverage of that path is deliberately partial; see the
embedded summary for exactly what is untested and why.

---

# Code Review Summary

**Branch**: `pt-3362-find-project-selector` (pushed; local worktree branch `worktree-pt-3362-find-project-selector`)

**Base**: `origin/main` (merge-base `ee850d6ac83`)

**Date**: 2026-08-17

**Files changed**: 8 source files (19 including regenerated `lib/platform-bible-react/dist/` build artifacts) — +1415 / −158

**Commits**:

- `4ba7344d669` — PT-3362: Add project selector to Find, remove reliance on scroll-group toolbar (original)
- `680c36993f1` — PT-3362: Harden Find project selector; fix rerun race and tab-close crash (review + manual-testing fixes)

## Overview

Adds a `ProjectSelector` to the Find tab's header (reusing `mode="projectScrollGroup"`, the same
configuration Manage Books, Checks, and Project Settings use) so the user can see and change which
project Find operates on. The picker lists only projects currently open in a scripture editor tab; the
selection is a `(projectId, scrollGroupId)` pair that a reassignment effect keeps valid as tabs open
and close, falling back to another open project — or a placeholder — when the current one closes.
Clicking a result now scrolls the exact tab the selector has chosen, rather than the pinned originating
editor tab.

The review found no critical issues but 12 important ones, clustered where the picker's new notion of
"which tab does Find act on" meets the scroll-group mechanism Find already used. **Manual testing then
found two more that the review and the full test suite both missed** — including a crash. Final tally:
14 important findings (11 fixed, 2 dismissed on verified premises, 1 deferred) and 17 minor (13 fixed,
4 open). 23 tests were added.

**This summary supersedes an earlier version** written before manual testing. Read the "Corrections to
this review" section — two of the review's own fixes caused regressions, and one of this document's
earlier claims was wrong.

## API Changes

- **`lib/platform-bible-react` — `ProjectSelector` gained one optional prop** (added during review, not
  in the original commit):
  - `hideFilterMenu?: boolean` — when true, the funnel/filter menu next to the search box is not
    rendered. Defaults to `false`, so `checks-side-panel` and `manage-books` are behaviorally
    unchanged. TSDoc'd on `CommonProps`.
  - `lib/platform-bible-react/dist/` regenerated accordingly (tracked in git per repo convention).
  - Likely to conflict with open PR #2673 (`feat/project-selector-more-grouping-options`), which
    reworks the same filter/grouping menu. **Author's call: not addressed here** — they will reconcile
    it when the other work merges. Noted for awareness only, not an action item for this PR.
- `lib/papi-dts/papi.d.ts` is **unchanged** (verified: `npm run build:types` produced no diff).
- Extension-internal (not a declared cross-boundary surface):
  - `find.utils.ts` exports `OpenScrollGroupTab` (now `Omit<OpenProjectTabWithWebView, 'webViewType'>`),
    `SelectedProjectScrollGroup`, `resolveSelectedProjectScrollGroup()`, and — added during review so
    they could be unit-tested — `isDifferentProjectSelection()`, `prunePresentBookIds()`, and
    `callControllerSafely()`.
  - `find.component.tsx` exports `FindProject`; `FindProps` gained 8 required props. Only
    `find.web-view.tsx` and `find.stories.tsx` consume `Find`; both updated, so no external caller
    breaks.
  - `localizedStrings.json` gained 7 `en` entries and 7 `metadata.fallbackKey` entries.

None.

#### Found by review

- [x] **The rerun ref fires on automatic reassignment, not just user selection** — the effect's comment
      claimed it fires "specifically for a user-initiated project switch", but the auto-reassignment
      effect routes through the same handler. _(Author confirmed both the auto-switch AND the
      auto-search are intended; fixed: comment rewritten to describe both paths, and three stale
      references to the non-existent `handleSelectProject` corrected.)_
- [x] **Search scope and the result-click broadcast followed Find's *own* scroll group** — seeded once
      at open time from the *opening* editor's group and never repointed, so `Current book`/`Current
      chapter` resolved against the wrong tab and every result click broadcast into the originating
      group. _(Fixed: the handler now also calls the hook's `setScrollGroupId` (tuple index 3,
      previously unused), keeping Find's group equal to the selected tab's. Author chose this over
      dropping the broadcast. Side effect: selecting a different tab of the same project in another
      group now also triggers an auto-search.)_
- [x] **The picker reported "No open projects" during startup** — `usePromise`'s loading flag was
      discarded, so `projects` was `[]` while the fetch was in flight even with a project open.
      _(Fixed: `isLoadingProjects` threaded into `ProjectSelector`'s `isLoading`, which also disables
      the trigger. Added a `LoadingProjects` story.)_
- [x] **A lowercased project id could reach `updateWebViewDefinition`, causing a double search** — the
      canonical-casing lookup resolves against `projects`, empty during the fetch, so the fallback
      wrote the lowercased id; the later flip to uppercase counted as a *second* switch. _(Fixed, two
      guards: the reassignment effect returns early on a cross-project fallback while metadata loads,
      and switch detection is now case-**insensitive** via `isDifferentProjectSelection`.)_
- [x] **A project opened after Find mounted could never appear** — the fetch callback had an empty
      dependency array. _(Fixed: a `projectMetadataGeneration` token re-runs the fetch when an open
      tab's project is missing, with a ref guard so a project with no Scripture/Paratext metadata
      retries exactly once rather than looping.)_
- [x] **A failed PDP resolution left Find silently searching the previous project** —
      `useProjectDataProvider` cannot express failure (`usePromise` with default `preserveValue: true`
      and no `.catch`), so `findPdp` keeps the old provider while the picker shows the new one.
      _(Fixed: an additive `findPdpAvailability` probe with `preserveValue: false` blocks
      `isSearchQueryValid` unless `'ready'` and surfaces an error. **See Corrections — this gate then
      caused two regressions.** Prevention half deferred, below.)_
- [x] **`ProjectSelector`'s internal strings were hardcoded English** — flagged by all four analysis
      passes. _(Fixed: a `projectSelectorLocalizedStrings` memo from new `%webView_find_projectSelector_*%`
      keys, scoped to the strings reachable from Find's configuration with each omission's reason
      documented inline, since localized keys are immutable once shipped.)_
- [x] **New localized keys were `en`-only with no `fallbackKey`** — the Localization Guide lists this as
      blocking and nothing in the build enforces parity. _(Fixed: 7 `metadata.fallbackKey` entries. Two
      corrections to the finding's premise, from a scan of all 33 languages: the
      `%manageBooks_projectSelector_*%` keys have **0/32** translations, so pointing at them only
      dedupes future work; and of the two "already translated" siblings cited, only `noProjectsFound`
      actually is — `es` only. Three genuinely-novel strings have no sibling and remain `en`-only for
      the normal translation pipeline.)_
- [ ] ~~The picker's scroll-group letters overload the existing convention, and Simple mode surfaces
      scroll groups the platform otherwise hides there~~ _(Author: not a problem — PT10 Simple enforces
      a single scroll group. **Verified**: `simple-layout.data.ts` sets no `scrollGroupScrRef` on any
      tab and every consumer defaults it to `0`. Residual, cosmetic only: the trigger still renders the
      `· A` suffix in Simple mode, conveying nothing when there is only one group.)_
- [ ] ~~`ProjectSelectorOpenTab.scrollGroupScrRefLabel` is unset, so duplicate rows are
      distinguishable only by badge letter~~ _(Author: intended, and duplicates cannot arise in Simple
      mode. **Verified**: Simple mode's fixed layout contains exactly one `platformScriptureEditor.react`
      tab, the only type in Find's filter — so duplicate rows are power-mode only, where the letter
      carries real meaning.)_
- [ ] **Four blocks are verbatim duplicates of `checks-side-panel` / `manage-books`** — `getProjectNames`,
      `SCRIPTURE_EDITOR_WEB_VIEW_TYPES` + `editorWebViewFilter` (a **third** copy), the `usePromise`
      metadata block, and the `sortedProjects` memo. `FindProject` also duplicates
      `ChecksSidePanelProject` and the library's `ProjectSelectorProject`. _(Author deferred: it touches
      two other panels, outside this PR's scope. **Note for whoever picks it up:** this branch's fixes
      added to the Find copy — the refetch effect and the `allSettled` change — so the Find and Checks
      copies have **diverged** rather than being byte-identical. Extraction means reconciling two
      versions, not lifting one.)_
- [x] **The orchestration around the pure resolver was untested** — the reassignment effect, the switch
      handler's reset and ordering, the canonical-id remapping, `targetEditorWebViewId`, the rerun
      effect, `findPdpAvailability`, and the refetch generation all had no coverage. _(Partially fixed:
      the genuinely pure decisions were extracted to `find.utils.ts` and tested, plus a component test
      for the results placeholder — 23 tests, each mutation-verified. **Still uncovered**: the
      metadata-loading gate, the refetch-once guard, `findPdpAvailability`'s blocking, and the
      scroll-group repoint, all of which live inside the web-view component where testing them means
      rendering with mocked `@papi` — the over-mocking pattern the Testing Guide warns against.
      Covering them needs the hook extraction the author declined. **Two of the three bugs manual
      testing later found were in exactly this uncovered region.**)_

#### Found by the author's manual testing (after the review passes)

- [x] **Selecting a different project did not re-run the search** — a regression introduced by the
      review's own `findPdpAvailability` fix. On a switch, `findPdpAvailability` resets to `'resolving'`
      (`preserveValue: false`, deliberately). The rerun effect fires on `findPdp`'s identity change and
      **consumed the pending flag unconditionally**, then `handleStartSearch` bailed on
      `!isSearchQueryValid` — which the same fix had gated on availability. Nothing re-armed the flag,
      so the rerun was silently dropped. _(Fixed: the effect now keys on both `findPdp` and
      `findPdpAvailability`, waits while `'resolving'`, and clears the flag only for conditions that
      will not resolve on their own.)_
- [x] **Closing the selected project's editor tab crashed the Find web view** —
      `TypeError: Cannot perform 'get' on a proxy that has been revoked`, confirmed in
      `~/.config/Electron/logs/main.log` and reproduced standalone with `Proxy.revocable` (matching the
      logged message verbatim). A web-view controller is a network object; disposing it revokes its
      proxy, so even *reading* `.runAnnotationAction` off it throws **synchronously** — `?.` guards only
      nullish and `.catch` only sees rejections. The annotation-cleanup effect runs precisely when
      `editorWebViewController` changes identity, and an uncaught throw in an effect cleanup takes down
      the web view. **This crash site predates PT-3362** (present at the merge-base) but PT-3362 made it
      routinely reachable, because the controller used to be the pinned originating editor and now
      changes whenever tabs close. _(Fixed: three unguarded controller call sites routed through a new
      `callControllerSafely` helper, which owns the property read via a thunk so the sync throw and the
      async rejection funnel to one place. The file already used the correct `try`/`catch` pattern at a
      fourth site — this applies the file's own working pattern consistently.)_

**Author response**: The author engaged with each finding rather than accepting them wholesale —
confirming intended behavior on the rerun/auto-search question, asking for fuller explanation before
agreeing the scroll-group coupling was an oversight, asking for a design proposal on the PDP-failure
finding before choosing the scoped half, and rejecting the two scroll-group UX findings on a factual
premise that checked out on verification. They deferred the duplication extraction, initially declined
test coverage, then reversed after the `Promise.all` interaction was explained — and their manual
testing then found the two bugs above.

### Minor — Consider

- [x] **Comment referenced `handleSelectProject`, which does not exist** _(fixed; grep confirms none
      remain)_
- [x] **`%webView_find_noOpenProjects_results%` broke the alphabetical key tuple** _(ordering fixed;
      the **key name** left unchanged at the author's direction, so nothing immutable was altered)_
- [x] **Dead `if (!names) return;` guard, and `Promise.all` made one bad project fatal** — the batch
      spans *every* Scripture/Paratext project, so one rejecting `getSetting` rejected all of it, and
      with no `.catch` in `usePromise` the dict stayed `{}` and loading stuck `true` **forever**.
      _(Fixed: `Promise.allSettled` with per-project warning logs, dead guard removed. **This was made
      worse mid-review** — three fixes gate on `isLoadingProjects`, including the refetch, so the
      failure had become a permanent disabled spinner with its own recovery path gated off.)_
- [x] **`type ProjectNamesById` declared inside the component body** _(hoisted to module scope)_
- [x] **`OpenScrollGroupTab` restated three of four fields of `OpenProjectTabWithWebView`** _(now
      `Omit<…, 'webViewType'>` via `import type`, which TypeScript erases — so `find.utils.ts` keeps its
      `@papi`-free runtime)_
- [x] **Results stayed rendered and clickable but inert after the last tab closed** _(fixed:
      `noOpenProjects` takes priority over both the idle prompt and any results on screen, and the
      status bar is suppressed so a stale count cannot contradict the placeholder)_
- [x] **A project switch did not reconcile `selectedBookIds` against `booksPresent`** _(fixed via
      `prunePresentBookIds`, guarded on a non-empty book list so the transient all-zero
      `BOOKS_PRESENT_DEFAULT` cannot wipe the selection — "don't know yet" must not read as "no books")_
- [x] **`ariaLabel` used section-heading wording rather than naming the control** _(fixed: new
      `%webView_find_projectSelector_label%` — "Project", with a `fallbackKey` to the checks-side-panel
      sibling for its `es`)_
- [x] **The header stacked four full-width rows** _(fixed: selector and Find/Replace toggle share one
      row with `flex-wrap`, taking the header to three)_
- [x] **The funnel/filter menu is effectively inert in Find** — in single-select mode it contains only
      the grouping toggle, and every listed project is already open. _(Fixed via the opt-in
      `hideFilterMenu` prop; the 4 filter-menu localized keys added earlier were then **removed**,
      having become certainly unreachable.)_
- [x] **The trigger label resolved the selection case-sensitively** _(resolved by the casing guards —
      a lowercased id can no longer reach the web view definition)_
- [x] **New localized values duplicated existing ones** _(partly addressed by the `fallbackKey`
      entries; a feature-neutral `%webView_projectSelector_*%` set shared by all three panels remains
      the ideal, tied to the deferred duplication work)_
- [ ] `noOpenProjects` is derivable from `openTabs.length === 0` — two props encoding one fact.
- [ ] `data-testid="find-project-trigger"` has exactly one reference repo-wide: its own definition.
- [ ] The trigger stays enabled when nothing is open; the popover then reads "No projects found", which
      implies a failed search rather than "nothing is open". `isDisabled` would fix it. (The *loading*
      case is now handled.)
- [ ] `projectFilter` key naming vs the vocabulary guideline, which reserves "filter" for narrowing
      lists. Consistent with the checks-side-panel sibling; noting only.

**Author response**: The author triaged the minor list explicitly, selecting 7 to fix plus the filter
menu as a "maybe", and later added the `Promise.all` fix once its severity was explained. Four were
left open by omission.

#### Shared Regions Modified

#### Extension Config Changes

None. No `webpack.config.ts`, `tsconfig.json`, `package.json`, or `.eslintrc*` touched.

- [n/a] `contributions/localizedStrings.json` — extension-specific string content, not shared
  infrastructure.

## Positive Observations

- The selection-resolution policy was extracted as a pure, `@papi`-free function with a thorough TSDoc
  contract and tests covering the same-project fallback, `preferredWebViewId` precedence, cross-project
  fallback, case-insensitivity, and the empty case — a clean split between testable policy and wiring.
- The hidden/inactive-view case for result preview is documented **at the sync site**, naming
  `.claude/rules/cross-view-sync-hidden-views.md` with three concrete reasons for the deliberate no-op —
  exactly what that rule asks for.
- Project-id casing was handled deliberately, with the reason documented at each comparison. The gaps
  found were *timing* races, not missed comparisons.
- Reuses production machinery rather than rebuilding: `useOpenProjectTabs` (3 consumers) and
  `ProjectSelector`'s existing `projectScrollGroup` mode, matching the discriminated-union contract.
- Non-obvious decisions are documented where a reader hits them: why `updateWebViewDefinition` instead
  of `reloadWebView`, why `abandonFindJob()` must precede the `projectId` mutation, why the rerun is
  ref-gated, why `onOpenProjectInGroup` logs rather than asserts.
- The file already contained the correct revoked-proxy defense at one site — the crash fix generalizes
  the author's own pattern rather than inventing one.

## Interview Notes

**Stated purpose**: add a project selector to the Find tab so results scroll the chosen editor tab
deterministically, removing reliance on the scroll-group toolbar.

**Design rationale (author's words)**: Find's behavior "should be inherently separate by design to the
web views that are showing the project ids" — Find's state is deliberately not bound to a particular
web view instance. The selection is a `(projectId, scrollGroupId)` pair rather than a `webViewId`
because the author wants "the option to search multiple tabs with the same project and scroll group" —
the pair is intentionally *coarser* than a web-view id.

**Mismatch worth reviewer attention**: that intent does not match current addressing.
`targetEditorWebViewId` uses `.find(...)` — first match wins — so a future "act on all matching tabs"
behavior needs `.filter(...)` with the callbacks fanning out. Today's single-target behavior is correct
and was left unchanged; the gap is recorded rather than silently fixed.

**Decisions the author made**:

- Auto-switch on tab close **and** the auto-search that follows are both intended; the comment was
  wrong, not the code.
- Chose coupling Find's scroll group to the selector over dropping the result-click broadcast.
- Chose the detection half of the PDP-failure fix; deferred the prevention half.
- Rejected both scroll-group UX findings on a premise that verification confirmed.
- Kept the `%webView_find_noOpenProjects_results%` key name while fixing only its ordering, correctly
  avoiding a change to an immutable shipped key.
- Initially skipped test coverage, then asked for it; chose the narrower "pure helpers + component
  tests" scope over a hook extraction.

**Author uncertainty**: on the scroll-group finding the author said "I think this might be an
oversight" and asked for a fuller explanation before deciding — diligence on a subtle interaction, not
a knowledge gap. On the PDP-failure finding they asked "How would you fix it?" rather than proposing an
approach, so the design of that fix (the additive probe, `preserveValue: false`, the error-ownership
ref) is the reviewer's, not the author's. **A reviewer should confirm the author is comfortable owning
that mechanism** — it is the least author-originated part of the diff, and it is the part that caused
two regressions.

**Not understood / deferred to AI**: no instance where the author deferred comprehension of their own
original code. `findPdpAvailability`, `hideFilterMenu`, and `callControllerSafely` were designed by the
review rather than the author.

## Corrections to this review

Recorded because a reviewer should not have to re-derive them, and because they bear on how much to
trust the rest of this document.

1. **The `findPdpAvailability` gate caused two separate regressions.** First it turned the
   `Promise.all` failure from "wrong empty message" into "permanently disabled spinner with its
   recovery path gated off" (caught in-session). Then it silently broke the project-switch rerun
   (caught by the author's manual testing, hours later). Both were in gating logic; both were invisible
   to 4874 passing tests. This is the strongest argument in the diff for the hook extraction.
2. **An earlier version of this summary called the `buildInfo.json` lint error "pre-existing and
   environmental."** That was wrong. It was a stale ESLint `--cache` artifact — `npm start` generates
   the file via `generate-dev-build-info.ts`, and once the app had run, clearing the cache produced a
   fully clean `npm run lint` (exit 0). Repo lint is clean, not clean-with-an-exception.
3. **The review overstated a conflict to justify deferring tests.** It claimed a Find-specific hook
   extraction "would partly conflict" with the deferred duplication work. The overlap is only the
   metadata fetch; Find's selection policy is not shared with any other panel.
4. **The tab-close crash was reachable from the original commit and the review missed it.** The review
   read that cleanup effect closely enough to cite it as a positive observation about hidden-view
   handling, without questioning the `?.` + `.catch()` shape around a disposable network object.

**Not a code issue, recorded so it is not re-investigated**: the app appeared to start in Simple mode
during manual testing. Cause was environmental — `settings.json` (`interfaceMode`) is **per worktree**
under `dev-appdata/data/`, while `window-layouts.json` is **shared by every worktree** under Electron's
`userData`. A simple-mode layout saved by one of the three simple-mode worktrees was restored here.
Also note a missing layout file yields the dev *default* layout, not an empty one — an emptied
`dockbox` entry is required for a genuinely empty dock.

## In-Review Quality Check

All from clean runs, immediately before the commit:

- **`npm run lint`** (the command CI runs): **exit 0**, zero errors, after clearing a stale ESLint
  cache (see Corrections #2).
- **`npm test`**: **exit 0** — **409 test files, 4874 tests passed**, 1 skipped (pre-existing).
- **`npm run build:types`**: no diff to `lib/papi-dts/papi.d.ts`.
- **`npm run build:platform-bible-react`**: succeeded (required — `ProjectSelector` changed); `dist/`
  regenerated and committed.
- **`npm run build`** (full, incl. .NET) and **`npm run build:extensions`**: both exit 0.
- **Prettier** `--check` on every changed file: all pass.
- **Find suite**: **109 tests / 4 files** — `find.utils.test.ts` 64 (was 46), new
  `find.component.test.tsx` 5, `structure-protection.util.test.ts` 23, `find-history.data-provider.test.ts`
  17.
- **`lib/platform-bible-react` suite**: 1028 tests / 130 files pass, including the ProjectSelector
  stories that exercise the modified component.
- **Mutation verification of all 23 new tests**: six independent mutations applied and reverted —
  case-sensitive `isDifferentProjectSelection`, dropped `prunePresentBookIds` guard, reverted
  placeholder priority, un-suppressed status bar, and the `callControllerSafely` try/catch removed.
  Each was caught by exactly the intended test. This also caught one vacuous assertion (a loose
  `/showing/` regex matching the scope control's "Showing" prefix rather than the status bar), since
  tightened.
- **Pre-commit hooks ran in full** (never bypassed): secret scan reported "no leaks found"; lint-staged
  clean.
- **Not run**: `dotnet test c-sharp-tests/` — no C# files touched.
- **Known gap**: the commit carries no `Session-URL:` trailer. The session URL could not be determined
  from the environment, and a literal placeholder in permanent history seemed worse than omission.
  Amend before merge if the convention is strict.

## Suggested Review Focus

Prioritized agenda for the author–reviewer meeting:

- [ ] **Whether to require the hook extraction before merge.** The review recommended deferring it;
      that recommendation should now be treated as wrong. The `findPdpAvailability` gate broke two
      separate flows, and two of the three bugs manual testing found were in the region the extraction
      would make testable. This is the top item.
- [ ] **The `findPdpAvailability` probe itself.** Designed by the review, not the author, and the source
      of both regressions. Walk through why `preserveValue: false` is load-bearing, why it is additive
      rather than replacing `useProjectDataProvider`, and whether the design should survive at all.
- [ ] **The scroll-group coupling.** Find's own scroll group now follows the picker. Confirm the
      semantic: the letter in Find's tab toolbar moves with the picker, and a new auto-search trigger
      exists when selecting a different tab of the same project in another group.
- [ ] **`callControllerSafely` and the crash class.** The fix generalizes a pattern the file already
      used, but the underlying hazard — a revoked proxy throwing on property read — applies to every
      consumer of a disposable web-view controller, not just Find. Consider whether this belongs in a
      shared utility.
- [ ] **Design intent vs addressing**: the `(projectId, scrollGroupId)` pair is meant to support acting
      on *multiple* tabs eventually, but `targetEditorWebViewId` is `.find(...)`. Build toward the
      fan-out now, or record as a known gap?
- [ ] **The deferred duplication extraction**, and that the Find and `checks-side-panel` copies have now
      **diverged** — extraction means reconciling two versions.
- [ ] **The deferred "only offer searchable projects" prevention** (adding
      `platformScripture.findInScripture` to `includeProjectInterfaces`). Blocked on one verification:
      does `getMetadataForAllProjects` report interfaces contributed by a *layering* PDPF? If not, that
      one-line filter would empty the picker for everyone.
- [ ] **Three localized keys ship `en`-only** with no translatable sibling. Confirm the translation
      pipeline picks them up.

AI-assisted — session URL not captured; see **AI Involvement** above.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2689)


jolierabideau (changes requested):
Thanks for working on this Levi! I left some Claude review comments. Claude also pointed out these during review:

- **PR description** — "reusing `mode="projectScrollGroup"`, the same configuration Manage Books, Checks, and Project Settings already use" isn't accurate: both use `mode="project"` ([checks-side-panel.component.tsx:222](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/extensions/src/platform-scripture/src/checks/checks-side-panel/checks-side-panel.component.tsx#L222), [manage-books-sidebar.component.tsx:288](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/extensions/src/platform-scripture/src/manage-books-dialog/manage-books-sidebar.component.tsx#L288)) — Find is the **first** consumer of `projectScrollGroup`. That matters for review: it's why `onOpenProjectInGroup` is mandatory and ends up as a dead no-op ([find.web-view.tsx:802](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/extensions/src/platform-scripture/src/find.web-view.tsx#L802)), and it means the mode's untested paths ship here first. Please correct the framing so the reviewer calibrates accordingly. (Also: the "@lwenger — confirm or delete this note" placeholder is still in the body.)
- **Comment density** — the rationale comments are unusually long even for this codebase (the `Promise.allSettled` block runs ~10 lines, the `useWebViewScrollGroupScrRef` block ~12, the rerun effect ~18). Much of it documents the debugging journey ("`preserveValue: false` is load-bearing", "an earlier comment claimed… that was wrong") rather than current intent, and it will rot the first time the surrounding code moves. Suggest compressing each to the invariant plus a pointer, and moving the narrative into the PR body where it's actually reviewer-facing.

Regarding the long comments, I think it's more important (than length) that we make sure they are forward-facing and helpful for future development, not just documenting the process! I have a [PR](https://github.com/paranext/paranext-core/pull/2586) up with some instructions for Claude to work on this. You could ask Claude to do a pass over your comments and give it the guidelines from the [forward-facing-comments.md](https://github.com/paranext/paranext-core/pull/2586/changes#diff-382cfe689d11d7924b66607a2cff23fd103106fc3933d4ab79c20d659a340896).

> @jolierabideau partially reviewed 19 files and all commit messages, and made 14 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2689#-:-P-KJqte-iw2-zZvt2oB:b7eu348)* status: all files reviewed, 13 unresolved discussions (waiting on captaincrazybro).

___
*[`extensions/src/platform-scripture/src/find.web-view.tsx` line 112 at r1](https://reviewable.io/reviews/paranext/paranext-core/2689#-P-KICLW9TmpwcG4llWV:-P-KICLW9TmpwcG4llWW:b7mrfwg) ([raw file](https://github.com/paranext/paranext-core/blob/680c36993f1c5e58e874521f8f8a23a5506f9ebc/extensions/src/platform-scripture/src/find.web-view.tsx#L112)):*
> ```XML
>   const projectFullName = await pdp.getSetting('platform.fullName');
>   return { shortName: projectShortName, fullName: projectFullName };
> }
> ```

Third verbatim copy of a block that was already extracted once. `getProjectNames`, `SCRIPTURE_EDITOR_WEB_VIEW_TYPES` (comment and all), `editorWebViewFilter`, and the `projectSelectorOpenTabs` memo are byte-for-byte duplicates of [checks-side-panel.web-view.tsx:39-56](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/extensions/src/platform-scripture/src/checks-side-panel.web-view.tsx#L39-L56) and [:697-712](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/extensions/src/platform-scripture/src/checks-side-panel.web-view.tsx#L697-L712), and the constant also exists in [manage-books.web-view.tsx:61](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/extensions/src/platform-scripture/src/manage-books.web-view.tsx#L61). `use-open-project-tabs.ts` exists _specifically_ to kill this duplication ("Replaces the inline subscription pattern duplicated in…"). Please finish the job the hook started: move the constant + filter into the hook module (e.g. export `useOpenScriptureEditorTabs()` returning both the tabs and the `ProjectSelectorOpenTab[]` shape) and put `getProjectNames` in a shared `@papi`\-using util. Three copies is where this stops being incidental.

_(Reported by Claude)_



_Code quote:_
```XML
const SCRIPTURE_EDITOR_WEB_VIEW_TYPES = new Set<string>(['platformScriptureEditor.react']);

/** Short and full names for every scripture project/resource, keyed by canonical project id. */
type ProjectNamesById = { [id: string]: Pick<FindProject, 'shortName' | 'fullName'> };

/**
 * Gets the short and full names of a project from its ID. Kept in the webview (not the shared,
 * `@papi`-free utils) so the utils stay importable by the presentational component and its story.
 */
async function getProjectNames(
  projectId: string,
): Promise<Pick<FindProject, 'shortName' | 'fullName'>> {
  const pdp = await papi.projectDataProviders.get('platform.base', projectId);
  const projectShortName = await pdp.getSetting('platform.name');
  const projectFullName = await pdp.getSetting('platform.fullName');
  return { shortName: projectShortName, fullName: projectFullName };
}
```
___
*[`extensions/src/platform-scripture/src/find.web-view.tsx` line 391 at r1](https://reviewable.io/reviews/paranext/paranext-core/2689#-P-KJ0E66nDkeYMDkC7B:-P-KJ0E66nDkeYMDkC7C:b-46a6fs) ([raw file](https://github.com/paranext/paranext-core/blob/680c36993f1c5e58e874521f8f8a23a5506f9ebc/extensions/src/platform-scripture/src/find.web-view.tsx#L391)):*
> ```XML
>   // Each missing id is recorded before bumping the token so a project that genuinely has no
>   // Scripture/Paratext metadata triggers exactly one retry rather than an infinite refetch loop.
>   const refetchAttemptedProjectIdsRef = useRef(new Set<string>());
> ```

The one-shot refetch guard can leave a permanently contradictory UI: tabs are open (`noOpenProjects === false`, results area behaves normally) but the project's metadata never resolved, so `projects` is empty and the picker shows "No open projects". The guard is the right instinct against a refetch loop, but consider re-arming when the open-tab set changes, or driving the picker's placeholder off the same `projects.length === 0 && !noOpenProjects` condition so the two regions can't disagree.

_(Reported by Claude)_
___
*[`extensions/src/platform-scripture/src/find/find.component.test.tsx` at r1](https://reviewable.io/reviews/paranext/paranext-core/2689#-P-KJd6T9sTwCjaAontV:-P-KJd6T9sTwCjaAontW:b-8l9per) ([raw file](https://github.com/paranext/paranext-core/blob/680c36993f1c5e58e874521f8f8a23a5506f9ebc/extensions/src/platform-scripture/src/find/find.component.test.tsx)):*
NIT Good tests for the placeholder/status-bar behavior, and the utils tests (especially the real `Proxy.revocable` one) are excellent. But the coverage stops at the presentational edges: nothing exercises the two effects carrying the actual risk — the reassignment effect and the project-switch rerun — and no test asserts the picker is wired up at all. Those effects are where the subtle bugs live (see the blocking item above); at minimum a test for "availability settles late → rerun still fires."
___
*[`extensions/src/platform-scripture/src/find.web-view.tsx` line 1388 at r1](https://reviewable.io/reviews/paranext/paranext-core/2689#-P-KIsnz6GeEJ18HVrrE:-P-KIsnz6GeEJ18HVrrF:b6u7wcw) ([raw file](https://github.com/paranext/paranext-core/blob/680c36993f1c5e58e874521f8f8a23a5506f9ebc/extensions/src/platform-scripture/src/find.web-view.tsx#L1388)):*
> ```XML
>   }, [editorWebViewController]);
> 
>   const handleFocusedResultChange = useCallback(
> ```

`.claude/rules/cross-view-sync-hidden-views.md` applies here and isn't addressed. Result activation now targets _the selected_ tab, which can easily be an inactive (`display: none`) tab — in Simple mode, one visible tab per stack makes that the common case. Single-click preview calls `selectRange` + `setAnnotation` without `setFocus`, so against a hidden tab the geometry-driven part silently no-ops and the user sees nothing happen. Double-click does `setFocus` first, so it's fine. The rule requires an explicit decision recorded **both** at the sync site and in the PR description — "do nothing while hidden" is an acceptable answer, but it has to be a stated one, and if there's a catch-up it needs a test.

_(Reported by Claude)_
___
*[`extensions/src/platform-scripture/src/find.web-view.tsx` line 319 at r1](https://reviewable.io/reviews/paranext/paranext-core/2689#-P-KIJ48AXw_0_T_8B_d:-P-KIJ48AXw_0_T_8B_e:b-3ino8n) ([raw file](https://github.com/paranext/paranext-core/blob/680c36993f1c5e58e874521f8f8a23a5506f9ebc/extensions/src/platform-scripture/src/find.web-view.tsx#L319)):*
> ```XML
>     useCallback(async () => {
>       const projectDict: ProjectNamesById = {};
> ```

The `Promise.allSettled` fix is right, and the reasoning in the comment is correct — but the identical `Promise.all` bug is still live in [checks-side-panel.web-view.tsx:122-129](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/extensions/src/platform-scripture/src/checks-side-panel.web-view.tsx#L122-L129), where one bad project permanently wedges that picker in the same unrecoverable way. Since you diagnosed it, please fix it there too (or file it) rather than leaving Find as the only hardened copy — that's how the three copies diverge into three behaviors.

_(Reported by Claude)_
___
*[`extensions/src/platform-scripture/contributions/localizedStrings.json` at r1](https://reviewable.io/reviews/paranext/paranext-core/2689#-P-KK9Pa1FakXIGDxBHY:-P-KK9Pa1FakXIGDxBHZ:b75eimu) ([raw file](https://github.com/paranext/paranext-core/blob/680c36993f1c5e58e874521f8f8a23a5506f9ebc/extensions/src/platform-scripture/contributions/localizedStrings.json)):*
NIT `"%webView_find_project_unavailableError%": "the selected project is not available to search"` renders as "An error occurred: the selected project is not available to search". For a PRD about honest messaging, an unresolvable provider is a _state_, not an error event — consider surfacing it as a results-area placeholder like the no-open-projects case rather than the red error line.

_(Reported by Claude)_
___
*[`extensions/src/platform-scripture/src/find.web-view.tsx` line 304 at r1](https://reviewable.io/reviews/paranext/paranext-core/2689#-P-KHrmzBcJYehW8Xq7w:-P-KHrmzBcJYehW8Xq7x:b5mrozy) ([raw file](https://github.com/paranext/paranext-core/blob/680c36993f1c5e58e874521f8f8a23a5506f9ebc/extensions/src/platform-scripture/src/find.web-view.tsx#L304)):*
> ```XML
>   // The returned `scrollGroupId` (index 2) is intentionally skipped: it is always
>   // `selectedScrollGroupId`, which we already hold.
>   const [verseRefSetting, setVerseRefSetting, , setFindScrollGroupId] =
> ```

The selector→toolbar sync is one-way, and the ticket asked for the toolbar UI to go away. PT-3362 is titled "Add project selector; **remove scroll group UI**", but the Power-mode tab-toolbar `ScrollGroupSelector` is still there, and this comment even states the two are kept in agreement. They aren't: `setFindScrollGroupId` pushes selector → toolbar, but if a Power user changes Find's group _from the toolbar_, `selectedScrollGroupId` (web-view state) is untouched, so the picker shows a stale group while `findScope`/`verseRefSetting` now resolve against the new one, and `targetEditorWebViewId` still points at the old tab. Also, the comment's claim that returned index 2 "is always `selectedScrollGroupId`" is only true if the toolbar can't move it — which it can. Either remove the toolbar control for the Find tab (matching the ticket) or reconcile index 2 back into `selectedScrollGroupId`.

_(Reported by Claude)_
___
*[`extensions/src/platform-scripture/src/find.web-view.tsx` line 1290 at r1](https://reviewable.io/reviews/paranext/paranext-core/2689#-P-KIgu37DrM2al1oPas:-P-KIgu37DrM2al1oPat:betewvn) ([raw file](https://github.com/paranext/paranext-core/blob/680c36993f1c5e58e874521f8f8a23a5506f9ebc/extensions/src/platform-scripture/src/find.web-view.tsx#L1290)):*
> ```XML
>     pendingProjectSwitchRerunRef.current = false;
>     explicitSearchPendingRef.current = true;
>     handleStartSearchRef.current(true);
> ```

The rerun passes `isExplicitSearch = true`, and `handleStartSearch` does `if (isExplicitSearch) addToHistory(searchTerm)` ([:969](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/extensions/src/platform-scripture/src/find.web-view.tsx#L969)). So an _automatic_ reassignment (tab closed → project switched) writes the term into recent searches. The scroll-group repoint also perturbs `verseRefSetting`, which changes `relevantScopeKey` and trips the options-changed history effect at [:933-939](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/extensions/src/platform-scripture/src/find.web-view.tsx#L933-L939). Recent searches should reflect user intent only. Suggest a separate flag for "rerun without history/UX side effects".

_(Reported by Claude)_
___
*[`extensions/src/platform-scripture/src/find.web-view.tsx` line 1316 at r1](https://reviewable.io/reviews/paranext/paranext-core/2689#-P-KHKxO1PC7VqMZMPdr:-P-KHKxO1PC7VqMZMPds:b-qrghis) ([raw file](https://github.com/paranext/paranext-core/blob/680c36993f1c5e58e874521f8f8a23a5506f9ebc/extensions/src/platform-scripture/src/find.web-view.tsx#L1316)):*
> ```XML
>   // Set explicitSearchPendingRef so the debounce timer (still pending when findPdp was already
>   // available at mount) skips its redundant second call.
>   useEffect(() => {
> ```

Restore-time search can be silently dropped, which is exactly the NN 1d case ("no situation where a search term is entered, but the result list stays empty… after tab restore"). This startup-fallback effect is keyed `[findPdp, searchStatus, searchTerm]`, but you added a new gate to `isSearchQueryValid` at [line 884](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/extensions/src/platform-scripture/src/find.web-view.tsx#L884) (`findPdpAvailability !== 'ready'`). Sequence on mount with a restored search term: the debounce fires while availability is `'resolving'` → `handleStartSearch` bails; this fallback fires when `findPdp` arrives → still `'resolving'` → bails again; availability then flips to `'ready'`, but none of the three deps changed, so nothing re-arms and the panel sits empty with no message. You correctly added `findPdpAvailability` to the project-switch rerun effect ([line 1290](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/extensions/src/platform-scripture/src/find.web-view.tsx#L1290)) — the same dep is missing here. Please add it (and a test: mount with a non-empty persisted term, resolve availability late, assert the search runs).

_(Reported by Claude)_
___
*[`extensions/src/platform-scripture/src/find/find.utils.ts` line 271 at r1](https://reviewable.io/reviews/paranext/paranext-core/2689#-P-KIUMZ65j3DZOtunG3:-P-KIUMZ65j3DZOtunG4:bbyt508) ([raw file](https://github.com/paranext/paranext-core/blob/680c36993f1c5e58e874521f8f8a23a5506f9ebc/extensions/src/platform-scripture/src/find/find.utils.ts#L271)):*
> ```XML
>   }
> 
>   const fallbackTab = openTabs[0];
> ```

`openTabs[0]` as the cross-project fallback, combined with the auto-rerun at [find.web-view.tsx:1278-1291](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/extensions/src/platform-scripture/src/find.web-view.tsx#L1278-L1291), means: user searches project A, closes A's editor tab, and Find silently switches to an arbitrary other open project _and re-runs the search there_. The results list repopulates with a different project's hits, with no announcement. Is that the intended behavior for a "trust the app" PRD? The alternative — keep the selection, show the "no target" placeholder, and let the user pick — is more honest. If the auto-switch is deliberate, please say so in the PR body so a reviewer can weigh it, and consider surfacing it in the UI. `openTabs[0]` is also unordered from a Map, so which project you land on isn't stable.

_(Reported by Claude)_
___
*[`extensions/src/platform-scripture/src/find/find.component.tsx` line 635 at r1](https://reviewable.io/reviews/paranext/paranext-core/2689#-P-KJSli1Ry2_Da7CB6g:-P-KJSli1Ry2_Da7CB6h:b-ahofsz) ([raw file](https://github.com/paranext/paranext-core/blob/680c36993f1c5e58e874521f8f8a23a5506f9ebc/extensions/src/platform-scripture/src/find/find.component.tsx#L635)):*
> ```XML
>           {/* Always visible; lets the user see and change which project Find operates on (and, by
>               extension, which open tab a result click will scroll). */}
>           <div className="tw:min-w-[8rem] tw:flex-1" data-testid="find-project-trigger">
> ```

NIT `data-testid="find-project-trigger"` has no consumer (no unit test, no e2e). Either use it in a test that the picker renders and fires `onSelectProjectScrollGroup`, or remove it.
___
*[`extensions/src/platform-scripture/src/find/find.component.tsx` line 637 at r1](https://reviewable.io/reviews/paranext/paranext-core/2689#-P-KHdBD-oRG515X2agq:-P-KHdBD-oRG515X2agr:b-u6rick) ([raw file](https://github.com/paranext/paranext-core/blob/680c36993f1c5e58e874521f8f8a23a5506f9ebc/extensions/src/platform-scripture/src/find/find.component.tsx#L637)):*
> ```XML
>           <div className="tw:min-w-[8rem] tw:flex-1" data-testid="find-project-trigger">
>             <ProjectSelector
>               mode="projectScrollGroup"
> ```

`mode="projectScrollGroup"` renders scroll-group identity in the UI: the trigger reads `"WEB · A"` ([project-selector.component.tsx:836](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/lib/platform-bible-react/src/components/advanced/project-selector/project-selector.component.tsx#L836)) and every row carries a group badge ([:359-365](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/lib/platform-bible-react/src/components/advanced/project-selector/project-selector.component.tsx#L359-L365)). The PRD lists scroll groups as a Donna-only no-go, and the renderer only shows `ScrollGroupSelector` in Power mode ([web-view.component.tsx:550-552](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/web-view.component.tsx#L550-L552)) — so Simple mode deliberately hides the concept everywhere else. Find already knows the mode (`isSimpleMode`, [find.web-view.tsx:233](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/extensions/src/platform-scripture/src/find.web-view.tsx#L233)) and uses it for `hideModeToggle`, but nothing gates the picker. Saroj will see an unexplained "· A". Options: use `mode="project"` in Simple (with the tab disambiguation resolved internally), or add a prop to suppress the group chip/suffix. Either way this needs a decision, not silence.

_(Reported by Claude)_
___
*[`extensions/src/platform-scripture/src/find/find.component.tsx` line 90 at r1](https://reviewable.io/reviews/paranext/paranext-core/2689#-P-KJ8Cj2FDJnNhzWlha:-P-KJ8Cj2FDJnNhzWlhb:b821e2b) ([raw file](https://github.com/paranext/paranext-core/blob/680c36993f1c5e58e874521f8f8a23a5506f9ebc/extensions/src/platform-scripture/src/find/find.component.tsx#L90)):*
> ```XML
>   '%webView_find_projectFilter_noOpenProjects%',
>   '%webView_find_projectFilter_noProjectsFound%',
>   '%webView_find_projectFilter_projectsAndResources%',
> ```

`%webView_find_projectFilter_projectsAndResources%` is declared in `FIND_LOCALIZED_STRING_KEYS` and shipped in `localizedStrings.json` (with a fallbackKey) but never rendered anywhere — `projectSelectorLocalizedStrings` only wires `searchPlaceholder`, `openTabsSectionHeading`, and `otherProjectsSectionHeading`. That's precisely the "permanent dead surface" your own comment at [:596-615](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/extensions/src/platform-scripture/src/find/find.component.tsx#L596-L615) argues against. Drop the key and its metadata entry before it ships and becomes immutable.

_(Reported by Claude)_

captaincrazybro:
## Review feedback triage — @jolierabideau's review

Thanks — this was a strong review. I verified every falsifiable claim against the branch before acting;
**none were false positives**, and four were mistakes I made or propagated. Below is per-comment status.

> **Note on thread resolution:** this PR has no GitHub review threads (`reviewThreads: []`) — the
> comments arrived as one review body via Reviewable, so there is nothing for me to resolve through the
> GitHub API. The Reviewable discussions still need closing there.

### ✅ Implemented

| Reviewer comment | What changed |
|---|---|
| **`find.web-view.tsx:1316`** — restore-time search silently dropped | **Confirmed: real bug, same class as the project-switch one, and I introduced it.** The startup-fallback effect gated on `isSearchQueryValid` (which now requires `findPdpAvailability === 'ready'`) without depending on availability, so a late-settling provider left a restored term on screen with no results and no message. Added the dependency, and a test that drives availability settling *after* mount. |
| **`find.component.tsx:90`** — `%webView_find_projectFilter_projectsAndResources%` is dead surface | **Confirmed unused** (only the key list + JSON entry; nothing renders it, because `ariaLabel` was repointed to `%webView_find_projectSelector_label%`). Removed the key, its `en` value, and its `metadata.fallbackKey` entry before it ships and becomes immutable. |
| **PR description** — inaccurate `mode` framing + leftover placeholder | **Confirmed wrong.** Checks (`:222`) and Manage Books (`:288`) both use `mode="project"`; Find is the **first** consumer of `projectScrollGroup`. The description now says so explicitly, including that it is why `onOpenProjectInGroup` is a dead no-op. Placeholder removed. |
| **`find.web-view.tsx:304`** — selector→toolbar sync is one-way (*reconciliation half*) | **Confirmed**, including that my comment's claim that index 2 "is always `selectedScrollGroupId`" was false. Added a reconciliation effect pulling a toolbar-driven group change back into `selectedScrollGroupId`, and rewrote the comment to describe the two-way reality. Skipped when Find is detached from any group, so a valid target isn't discarded. (Removing the toolbar control itself — see Skipped.) |
| **`find.web-view.tsx:1290`** — automatic reassignment writes to recent searches | **Valid mechanism, with a correction.** `addToHistory` has a dedupe guard, so the spurious write only occurs in a narrow window. But the guard was keyed on *term alone* while history is stored *per project* — so after a switch the term that just ran was never recorded under the new project. Fixed both: the dedupe is now per `(project, term)`, and the automatic rerun starts as **non-explicit** so it writes no history entry. Your suggested flag alone would have suppressed the spurious write while leaving the missing one. |
| **`find.utils.ts:271`** — `openTabs[0]` is an arbitrary fallback | **Confirmed.** (Small precision: `Map` iteration *is* insertion-ordered, but insertion order is web-view-event arrival order, so the substance holds.) The fallback is now deterministically ordered by scroll group then project id, with a test asserting three scrambled orderings resolve identically. |
| **`find.component.test.tsx`** — coverage stops at the presentational edges | **Agreed, and this is the gap that produced the bug above.** Extracted both no-user-input search triggers into `use-find-search-triggers.hook.ts` (plain values + refs + a callback, no `@papi`, so no mocking) and added 10 tests driving them across renders — including the two "availability settles late → search still fires" cases you asked for. Mutation-tested: dropping either availability dependency fails exactly the matching test. That mutation pass also caught one of my own tests passing vacuously (rebuilt object identities re-ran the effect regardless of its deps); it now holds identities stable. |

### ⏭️ Skipped (deliberate, with reasons)

| Reviewer comment | Why skipped |
|---|---|
| **`find.web-view.tsx:112`** — three verbatim copies; finish what `use-open-project-tabs` started | Agreed and valid. Deferred: it edits `checks-side-panel` and `manage-books`, outside this PR's scope, and a shared hook wants its own tests. **Note:** this PR's fixes added to the Find copy (refetch effect, `allSettled`), so the copies have now **diverged** — extraction means reconciling versions, not lifting one. |
| **`find.web-view.tsx:319`** — same `Promise.all` bug live in `checks-side-panel:123` | **Confirmed still live**, dead `if (!names) return;` included. Not fixed here to keep this PR scoped to Find; needs its own change or ticket. |
| **`find.web-view.tsx:391`** — refetch guard can leave a contradictory UI | Valid edge case. Deferred as lower-severity than the items above. |
| **`find.component.tsx:637`** — `mode="projectScrollGroup"` shows "· A" in Simple mode | Valid and needs a product decision (`mode="project"` in Simple vs. a chip-suppression prop) rather than a unilateral code change. Deferred pending that call. |
| **`find.web-view.tsx:304`** — *remove the toolbar control* (scope half) | Depends on the ticket's actual scope: the commit says "remove **reliance on** the scroll-group toolbar", which differs from removing the UI. Left for the author to confirm against PT-3362. |
| **`find.web-view.tsx:1388`** — hidden-views rule | Partly disputed. The sync site **does** already document it (naming the rule, three reasons for the deliberate no-op), so that half of the rule is met. Two corrections: the Simple-mode premise is wrong — the scripture editor sits **alone** in its column in `simple-layout.data.ts`, so it is always visible there; the stacking is in column 3. This is power-mode-only. The genuinely missing half is a flagged callout in the PR description rather than a positive observation; not added in this pass. |
| **`localizedStrings.json`** — unavailable provider is a state, not an error | Valid framing point; deferred as a UX change rather than a defect. |
| **`find.component.tsx:635`** — unused `data-testid` | Valid; left in place this pass. |
| **PR-level** — comment density / forward-facing comments | Fair, and I agree several comments narrate the debugging journey rather than current intent. Not done here — worth a dedicated pass against `forward-facing-comments.md` from #2586 rather than piecemeal edits. |

### Verification

Repo-wide `npm run lint` clean, full `npm test` green, Find suite now **120 tests / 5 files**. All new
assertions mutation-verified.

captaincrazybro (commented):
> @captaincrazybro made 12 comments and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2689)* status: 14 of 21 files reviewed, 11 unresolved discussions (waiting on jolierabideau).

___
*[`extensions/src/platform-scripture/contributions/localizedStrings.json` at r1](https://reviewable.io/reviews/paranext/paranext-core/2689#-P-KK9Pa1FakXIGDxBHY:-P-LBoKy3DYbbKFa_3TL:b-896fix) ([raw file](https://github.com/paranext/paranext-core/blob/680c36993f1c5e58e874521f8f8a23a5506f9ebc/extensions/src/platform-scripture/contributions/localizedStrings.json)):*

NIT `"%webView_find_project_unavailableError%": "the selected project is not available to search"` renders as "An error occurred: the selected project is not available to search". For a PRD about honest messaging, an unresolvable provider is a _state_, not an error event — consider surfacing it as a results-area placeholder like the no-open-projects case rather than the red error line.

_(Reported by Claude)_
</blockquote></details>

Done.
___
*[`extensions/src/platform-scripture/src/find.web-view.tsx` line 112 at r1](https://reviewable.io/reviews/paranext/paranext-core/2689#-P-KICLW9TmpwcG4llWV:-P-LCDJn6RmiGVVvke78:btccema) ([raw file](https://github.com/paranext/paranext-core/blob/680c36993f1c5e58e874521f8f8a23a5506f9ebc/extensions/src/platform-scripture/src/find.web-view.tsx#L112)):*

Third verbatim copy of a block that was already extracted once. `getProjectNames`, `SCRIPTURE_EDITOR_WEB_VIEW_TYPES` (comment and all), `editorWebViewFilter`, and the `projectSelectorOpenTabs` memo are byte-for-byte duplicates of [checks-side-panel.web-view.tsx:39-56](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/extensions/src/platform-scripture/src/checks-side-panel.web-view.tsx#L39-L56) and [:697-712](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/extensions/src/platform-scripture/src/checks-side-panel.web-view.tsx#L697-L712), and the constant also exists in [manage-books.web-view.tsx:61](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/extensions/src/platform-scripture/src/manage-books.web-view.tsx#L61). `use-open-project-tabs.ts` exists _specifically_ to kill this duplication ("Replaces the inline subscription pattern duplicated in…"). Please finish the job the hook started: move the constant + filter into the hook module (e.g. export `useOpenScriptureEditorTabs()` returning both the tabs and the `ProjectSelectorOpenTab[]` shape) and put `getProjectNames` in a shared `@papi`\-using util. Three copies is where this stops being incidental.

_(Reported by Claude)_
</blockquote></details>

Out of scope
___
*[`extensions/src/platform-scripture/src/find.web-view.tsx` line 304 at r1](https://reviewable.io/reviews/paranext/paranext-core/2689#-P-KHrmzBcJYehW8Xq7w:-P-LCIde0irow7Zd1AwD:b-896fix) ([raw file](https://github.com/paranext/paranext-core/blob/680c36993f1c5e58e874521f8f8a23a5506f9ebc/extensions/src/platform-scripture/src/find.web-view.tsx#L304)):*

The selector→toolbar sync is one-way, and the ticket asked for the toolbar UI to go away. PT-3362 is titled "Add project selector; **remove scroll group UI**", but the Power-mode tab-toolbar `ScrollGroupSelector` is still there, and this comment even states the two are kept in agreement. They aren't: `setFindScrollGroupId` pushes selector → toolbar, but if a Power user changes Find's group _from the toolbar_, `selectedScrollGroupId` (web-view state) is untouched, so the picker shows a stale group while `findScope`/`verseRefSetting` now resolve against the new one, and `targetEditorWebViewId` still points at the old tab. Also, the comment's claim that returned index 2 "is always `selectedScrollGroupId`" is only true if the toolbar can't move it — which it can. Either remove the toolbar control for the Find tab (matching the ticket) or reconcile index 2 back into `selectedScrollGroupId`.

_(Reported by Claude)_
</blockquote></details>

Done.
___
*[`extensions/src/platform-scripture/src/find.web-view.tsx` line 319 at r1](https://reviewable.io/reviews/paranext/paranext-core/2689#-P-KIJ48AXw_0_T_8B_d:-P-LCLlk4BE6zkIMC-GH:btccema) ([raw file](https://github.com/paranext/paranext-core/blob/680c36993f1c5e58e874521f8f8a23a5506f9ebc/extensions/src/platform-scripture/src/find.web-view.tsx#L319)):*

The `Promise.allSettled` fix is right, and the reasoning in the comment is correct — but the identical `Promise.all` bug is still live in [checks-side-panel.web-view.tsx:122-129](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/extensions/src/platform-scripture/src/checks-side-panel.web-view.tsx#L122-L129), where one bad project permanently wedges that picker in the same unrecoverable way. Since you diagnosed it, please fix it there too (or file it) rather than leaving Find as the only hardened copy — that's how the three copies diverge into three behaviors.

_(Reported by Claude)_
</blockquote></details>

Out of scope
___
*[`extensions/src/platform-scripture/src/find.web-view.tsx` line 391 at r1](https://reviewable.io/reviews/paranext/paranext-core/2689#-P-KJ0E66nDkeYMDkC7B:-P-LCervEKcbE_YaDemr:ba2a7jp) ([raw file](https://github.com/paranext/paranext-core/blob/680c36993f1c5e58e874521f8f8a23a5506f9ebc/extensions/src/platform-scripture/src/find.web-view.tsx#L391)):*

The one-shot refetch guard can leave a permanently contradictory UI: tabs are open (`noOpenProjects === false`, results area behaves normally) but the project's metadata never resolved, so `projects` is empty and the picker shows "No open projects". The guard is the right instinct against a refetch loop, but consider re-arming when the open-tab set changes, or driving the picker's placeholder off the same `projects.length === 0 && !noOpenProjects` condition so the two regions can't disagree.

_(Reported by Claude)_
</blockquote></details>

Unlikely enough to happen
___
*[`extensions/src/platform-scripture/src/find.web-view.tsx` line 1290 at r1](https://reviewable.io/reviews/paranext/paranext-core/2689#-P-KIgu37DrM2al1oPas:-P-LEzVeD7xmgfP0TXnn:b-896fix) ([raw file](https://github.com/paranext/paranext-core/blob/680c36993f1c5e58e874521f8f8a23a5506f9ebc/extensions/src/platform-scripture/src/find.web-view.tsx#L1290)):*

The rerun passes `isExplicitSearch = true`, and `handleStartSearch` does `if (isExplicitSearch) addToHistory(searchTerm)` ([:969](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/extensions/src/platform-scripture/src/find.web-view.tsx#L969)). So an _automatic_ reassignment (tab closed → project switched) writes the term into recent searches. The scroll-group repoint also perturbs `verseRefSetting`, which changes `relevantScopeKey` and trips the options-changed history effect at [:933-939](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/extensions/src/platform-scripture/src/find.web-view.tsx#L933-L939). Recent searches should reflect user intent only. Suggest a separate flag for "rerun without history/UX side effects".

_(Reported by Claude)_
</blockquote></details>

Done.
___
*[`extensions/src/platform-scripture/src/find.web-view.tsx` line 1316 at r1](https://reviewable.io/reviews/paranext/paranext-core/2689#-P-KHKxO1PC7VqMZMPdr:-P-LF0cv-lNRFYBkPXKl:b-896fix) ([raw file](https://github.com/paranext/paranext-core/blob/680c36993f1c5e58e874521f8f8a23a5506f9ebc/extensions/src/platform-scripture/src/find.web-view.tsx#L1316)):*

Restore-time search can be silently dropped, which is exactly the NN 1d case ("no situation where a search term is entered, but the result list stays empty… after tab restore"). This startup-fallback effect is keyed `[findPdp, searchStatus, searchTerm]`, but you added a new gate to `isSearchQueryValid` at [line 884](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/extensions/src/platform-scripture/src/find.web-view.tsx#L884) (`findPdpAvailability !== 'ready'`). Sequence on mount with a restored search term: the debounce fires while availability is `'resolving'` → `handleStartSearch` bails; this fallback fires when `findPdp` arrives → still `'resolving'` → bails again; availability then flips to `'ready'`, but none of the three deps changed, so nothing re-arms and the panel sits empty with no message. You correctly added `findPdpAvailability` to the project-switch rerun effect ([line 1290](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/extensions/src/platform-scripture/src/find.web-view.tsx#L1290)) — the same dep is missing here. Please add it (and a test: mount with a non-empty persisted term, resolve availability late, assert the search runs).

_(Reported by Claude)_
</blockquote></details>

Done.
___
*[`extensions/src/platform-scripture/src/find.web-view.tsx` line 1388 at r1](https://reviewable.io/reviews/paranext/paranext-core/2689#-P-KIsnz6GeEJ18HVrrE:-P-LFDOo0Gkf7j3oFlFS:b-896fix) ([raw file](https://github.com/paranext/paranext-core/blob/680c36993f1c5e58e874521f8f8a23a5506f9ebc/extensions/src/platform-scripture/src/find.web-view.tsx#L1388)):*

`.claude/rules/cross-view-sync-hidden-views.md` applies here and isn't addressed. Result activation now targets _the selected_ tab, which can easily be an inactive (`display: none`) tab — in Simple mode, one visible tab per stack makes that the common case. Single-click preview calls `selectRange` + `setAnnotation` without `setFocus`, so against a hidden tab the geometry-driven part silently no-ops and the user sees nothing happen. Double-click does `setFocus` first, so it's fine. The rule requires an explicit decision recorded **both** at the sync site and in the PR description — "do nothing while hidden" is an acceptable answer, but it has to be a stated one, and if there's a catch-up it needs a test.

_(Reported by Claude)_
</blockquote></details>

Done.
___
*[`extensions/src/platform-scripture/src/find/find.component.tsx` line 90 at r1](https://reviewable.io/reviews/paranext/paranext-core/2689#-P-KJ8Cj2FDJnNhzWlha:-P-LFXnxAvivzb2vaSbr:b-896fix) ([raw file](https://github.com/paranext/paranext-core/blob/680c36993f1c5e58e874521f8f8a23a5506f9ebc/extensions/src/platform-scripture/src/find/find.component.tsx#L90)):*

`%webView_find_projectFilter_projectsAndResources%` is declared in `FIND_LOCALIZED_STRING_KEYS` and shipped in `localizedStrings.json` (with a fallbackKey) but never rendered anywhere — `projectSelectorLocalizedStrings` only wires `searchPlaceholder`, `openTabsSectionHeading`, and `otherProjectsSectionHeading`. That's precisely the "permanent dead surface" your own comment at [:596-615](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/extensions/src/platform-scripture/src/find/find.component.tsx#L596-L615) argues against. Drop the key and its metadata entry before it ships and becomes immutable.

_(Reported by Claude)_
</blockquote></details>

Done.
___
*[`extensions/src/platform-scripture/src/find/find.component.tsx` line 635 at r1](https://reviewable.io/reviews/paranext/paranext-core/2689#-P-KJSli1Ry2_Da7CB6g:-P-LFZ-zEpQ8leBLY9MM:b-tdjgiw) ([raw file](https://github.com/paranext/paranext-core/blob/680c36993f1c5e58e874521f8f8a23a5506f9ebc/extensions/src/platform-scripture/src/find/find.component.tsx#L635)):*

NIT `data-testid="find-project-trigger"` has no consumer (no unit test, no e2e). Either use it in a test that the picker renders and fires `onSelectProjectScrollGroup`, or remove it.
</blockquote></details>

Future proofing...
___
*[`extensions/src/platform-scripture/src/find/find.component.tsx` line 637 at r1](https://reviewable.io/reviews/paranext/paranext-core/2689#-P-KHdBD-oRG515X2agq:-P-LF_rTCJLw4Kf2dolN:b-u2op8a) ([raw file](https://github.com/paranext/paranext-core/blob/680c36993f1c5e58e874521f8f8a23a5506f9ebc/extensions/src/platform-scripture/src/find/find.component.tsx#L637)):*

`mode="projectScrollGroup"` renders scroll-group identity in the UI: the trigger reads `"WEB · A"` ([project-selector.component.tsx:836](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/lib/platform-bible-react/src/components/advanced/project-selector/project-selector.component.tsx#L836)) and every row carries a group badge ([:359-365](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/lib/platform-bible-react/src/components/advanced/project-selector/project-selector.component.tsx#L359-L365)). The PRD lists scroll groups as a Donna-only no-go, and the renderer only shows `ScrollGroupSelector` in Power mode ([web-view.component.tsx:550-552](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/src/renderer/components/web-view.component.tsx#L550-L552)) — so Simple mode deliberately hides the concept everywhere else. Find already knows the mode (`isSimpleMode`, [find.web-view.tsx:233](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/extensions/src/platform-scripture/src/find.web-view.tsx#L233)) and uses it for `hideModeToggle`, but nothing gates the picker. Saroj will see an unexplained "· A". Options: use `mode="project"` in Simple (with the tab disambiguation resolved internally), or add a prop to suppress the group chip/suffix. Either way this needs a decision, not silence.

_(Reported by Claude)_
</blockquote></details>

This is fine for the Simple user
___
*[`extensions/src/platform-scripture/src/find/find.utils.ts` line 271 at r1](https://reviewable.io/reviews/paranext/paranext-core/2689#-P-KIUMZ65j3DZOtunG3:-P-LGCnw6LS-uDkY6ndD:bl6i6ke) ([raw file](https://github.com/paranext/paranext-core/blob/680c36993f1c5e58e874521f8f8a23a5506f9ebc/extensions/src/platform-scripture/src/find/find.utils.ts#L271)):*

`openTabs[0]` as the cross-project fallback, combined with the auto-rerun at [find.web-view.tsx:1278-1291](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/extensions/src/platform-scripture/src/find.web-view.tsx#L1278-L1291), means: user searches project A, closes A's editor tab, and Find silently switches to an arbitrary other open project _and re-runs the search there_. The results list repopulates with a different project's hits, with no announcement. Is that the intended behavior for a "trust the app" PRD? The alternative — keep the selection, show the "no target" placeholder, and let the user pick — is more honest. If the auto-switch is deliberate, please say so in the PR body so a reviewer can weigh it, and consider surfacing it in the UI. `openTabs[0]` is also unordered from a Map, so which project you land on isn't stable.

_(Reported by Claude)_
</blockquote></details>

This is the intended functionality

jolierabideau (approved):
> @jolierabideau reviewed 7 files and all commit messages, made 1 comment, and resolved 11 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2689)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

___
*[`extensions/src/platform-scripture/src/find/find.component.tsx` line 637 at r1](https://reviewable.io/reviews/paranext/paranext-core/2689#-P-KHdBD-oRG515X2agq:-P-PH5Sl-GwoVH6hkEu0:bgzd2om) ([raw file](https://github.com/paranext/paranext-core/blob/680c36993f1c5e58e874521f8f8a23a5506f9ebc/extensions/src/platform-scripture/src/find/find.component.tsx#L637)):*
<details><summary><i>Previously, captaincrazybro (Levi Wenger) wrote…</i></summary><blockquote>

This is fine for the Simple user
</blockquote></details>

Done I see your comment in your larger triage statement, and I agree that it should be a UX/product decision (though I imagine we do not want to see scroll groups in simple). I will take it to Ian today, and file a follow-up issue based on what he says.

jolierabideau (changes requested):
> @jolierabideau made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2689)* status: all files reviewed, 1 unresolved discussion (waiting on jolierabideau).

___
*[`extensions/src/platform-scripture/src/find/find.component.tsx` line 637 at r1](https://reviewable.io/reviews/paranext/paranext-core/2689#-P-KHdBD-oRG515X2agq:-P-PIhoo0vd7Fr4ATgue:bbes1hc) ([raw file](https://github.com/paranext/paranext-core/blob/680c36993f1c5e58e874521f8f8a23a5506f9ebc/extensions/src/platform-scripture/src/find/find.component.tsx#L637)):*

Done I see your comment in your larger triage statement, and I agree that it should be a UX/product decision (though I imagine we do not want to see scroll groups in simple). I will take it to Ian today, and file a follow-up issue based on what he says.
</blockquote></details>

Actually I don't see the scroll group in the Storybook example of the component we were using for this. Did you see what mode was used in this https://paranext.github.io/paranext-core/platform-bible-react-storybook/?path=/story/advanced-project-selector--simple-flat-list

jolierabideau (approved):
> @jolierabideau reviewed 6 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2689)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

jolierabideau (approved):
> @jolierabideau partially reviewed 23 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2689)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2690, opened 2026-08-18
PT-4111: Add shadcn Empty component

jolierabideau:
## ⚠️ Merge instructions — do not squash

**This PR MUST NOT be squash-merged.** Use a merge commit.

It is deliberately structured as three commits — a pristine `npx shadcn apply` baseline, the standard
Platform.Bible customizations, then review fixes — and `shadcn-customization-tracker.md` resolves a
component's baseline commit from that history. Squashing would collapse the baseline into a commit
that already contains every customization, so `git diff <baseline>..HEAD` would return nothing and the
git-history half of the customization audit would be permanently disabled for `empty.tsx`. This has
already happened once: `kbd.tsx` entered main through squash commit `9224356de33` (#2147) and its
baseline `e016e45280b` never landed.

**Rebuild `dist` immediately before merging.** `lib/platform-bible-react` commits its build output, and
`dist/index.js` conflicts resolve by taking one side while the code-split `resizable-*` chunk
adds/deletes auto-merge cleanly — so a stale resolution can produce an `index.js` importing a chunk the
other side deleted, with CI green and a runtime module-not-found in the renderer.

## Inherited change: Button pressed-state transform

This PR ships a `Button` behavior change it did not author. Main's committed bundle was stale: #2667
(`ff22ea69aa9`, "Keep the Button press nudge from clobbering caller translates") changed `button.tsx`
and never rebuilt `dist`. `translateY(1px)` appears 0 times in main's `dist/index.js` and 1 time here,
and because `package.json` resolves consumers through `dist/`, **this PR is what actually delivers that
pressed-state change to every Button in the product.** If it regresses, `git bisect` lands here.

Root cause worth fixing separately: root `npm run build` never invokes `build:platform-bible-react`, so
CI's "Verify no files changed after build" check is a no-op for this package.

## Links

- **Jira**: https://paratextstudio.atlassian.net/browse/PT-4111
- **Storybook (new)**: `Shadcn/Empty` — `lib/platform-bible-react/src/stories/shadcn-ui/empty.stories.tsx`
  (7 stories: `Simple`, `WithIconMedia`, `WithDefaultMedia`, `WithAction`, `DashedBorder`, `NarrowWidth`, `LiveRegion`)
- **Storybook (updated)**: [`Basics/EmptyState`](https://paranext.github.io/paranext-core/platform-bible-react-storybook/?path=/docs/basics-emptystate--docs) ·
  [`Guidelines/Component Choices`](https://paranext.github.io/paranext-core/platform-bible-react-storybook/?path=/docs/guidelines-component-choices--docs) (new "Zero states" section)
- Note for UX: these stories never reach Chromatic — `chromatic.yml` targets `ai`/`main`, needs a
  `storybook-review` label, and its glob excludes `lib/` — so the Storybook deliverable gets no
  automated visual review. Run Storybook locally to review.

## Round 2 — reviewer findings addressed

Fixed: the React-import hand-patch is now a transform (`ensureReactImport` in
`scripts/shadcn-transform-utils.ts`, 8 new tests) which retires the six-line comment at `empty.tsx:1`;
`pr-twp` moved back out of the `emptyMediaVariants` cva base per Code-Style-Guide.md:350;
`data-slot` reverted to upstream's `empty-icon` (matching `input-group.tsx`); new `empty.test.tsx`
pins `pr-twp` on all six slots and the slot values a `shadcn apply` would revert; the `Empty`/`EmptyState`
guidance now agrees across TSDoc, both story blurbs and a new `component-choices.mdx` "Zero states"
section; added a `NarrowWidth` story at the 300px minimum; corrected the `role="status"` a11y prose;
replaced markdown backticks in JSX children with plain text; documented the `border-dashed` border-width
requirement and `EmptyTitle`'s `<div>` element in TSDoc.

Deferred, with reasoning in the review thread: migrating an in-tree hand-rolled empty state to `<Empty>`
(out of scope for a component-add PR), and `EmptyTitle`'s `text-sm` sizing / heading semantics (faithful
to the shadcn preset; needs UX sign-off, not a unilateral deviation).

Closed with evidence: `{@link EmptyState}` resolves — `npm run build:docs` exits 0 (typedoc's entry point
is `src/index.ts`, so both symbols are in scope).

---

# Code Review Summary

**Branch**: pt-4111-shadcn-empty

**Base**: origin/main

**Date**: 2026-08-18

**Files changed**: 18 (4 source files; 14 regenerated `dist/` bundles)

## Overview

This PR adds the shadcn/ui `Empty` component family to `platform-bible-react` — a centered
zero-state composition (`Empty`, `EmptyHeader`, `EmptyMedia`, `EmptyTitle`, `EmptyDescription`,
`EmptyContent`) — following the repo's established two-commit shape: a pristine
`npx shadcn apply --preset b6rt8cvlC` baseline, then a commit applying the standard Platform.Bible
customizations. The change is purely additive to the public API: six new exports, none removed or
modified, verified by diffing the export name sets in both `dist/index.js` (259 → 265) and
`dist/index.d.ts` (347 → 353).

The analysis found no critical findings and no breaking changes. Two Important findings surfaced —
a missing Storybook story and an undocumented overlap with the pre-existing `EmptyState` component
— and both were fixed during the review. Four of the seven Minor findings were also fixed
(the `EmptyDescription` type/element mismatch, the `EmptyMedia` `data-slot` naming, the `pr-twp`
class-placement convention drift, and the repeated `// CUSTOM` comment volume); the remaining three
were dismissed as deliberate decisions with rationale recorded below.

## API Changes

- `lib/platform-bible-react`: Added `Empty` export (new shadcn/ui component, `React.ComponentProps<'div'>`)
- `lib/platform-bible-react`: Added `EmptyHeader` export (`React.ComponentProps<'div'>`)
- `lib/platform-bible-react`: Added `EmptyMedia` export (`React.ComponentProps<'div'> & VariantProps<typeof emptyMediaVariants>`, `variant?: 'default' | 'icon'`, default `'default'`)
- `lib/platform-bible-react`: Added `EmptyTitle` export (`React.ComponentProps<'div'>`)
- `lib/platform-bible-react`: Added `EmptyDescription` export — `React.ComponentProps<'div'>` (changed during review from upstream's `React.ComponentProps<'p'>` so the type matches the rendered element)
- `lib/platform-bible-react`: Added `EmptyContent` export (`React.ComponentProps<'div'>`)
- `lib/platform-bible-react/dist/`: Regenerated bundles (`index.js/.cjs/.d.ts`, `experimental.js/.cjs`) and renamed the `resizable-*` code-split chunks (`resizable-C3AVzTx2.js` → `resizable-aPuquhT0.js`, `resizable-Bvnq8xnB.cjs` → `resizable-CSt6YG_F.cjs`)
- No exports removed or modified. `papi.d.ts` and extension type declaration files: unchanged.
- Non-API but consumer-visible: `EmptyMedia`'s `data-slot` changed from upstream's `empty-icon` to
  `empty-media` during review. Verified no source file, stylesheet, test, or snapshot targets the
  old value (the only hits were in the stale `dist/` output, since regenerated).

None.

- [x] **No Storybook story for the new `Empty` family.** 27 of 37 files in `src/components/shadcn-ui/` have a matching story in `src/stories/shadcn-ui/`, and `Guidelines/UI Review` requires a Storybook deliverable per UI work item. _(fixed during review: added `lib/platform-bible-react/src/stories/shadcn-ui/empty.stories.tsx` with six stories — `Simple`, `WithIconMedia`, `WithDefaultMedia` (covers both `EmptyMedia` variants), `WithAction`, `DashedBorder`, `LiveRegion`)_
- [x] **Two overlapping empty-state APIs with no guidance.** The library already exported `EmptyState` (`src/components/basics/empty-state.component.tsx`, in use at `extensions/src/platform-scripture-editor/src/scripture-text-grid.web-view.tsx:529`), which also carries `role="status"` that the shadcn `Empty` primitives lack. `Guidelines/Component Choices` asks for agreed components per common usage ("It is an ongoing effort to review and agree on components and variants for common usage"), and had no zero-state entry. _(fixed during review: added a cross-reference to `Empty`'s TSDoc stating when to use which and noting the primitives set no ARIA role; added a reciprocal pointer to `EmptyState`'s TSDoc; added a component-level docs description in the new story; added the `LiveRegion` story demonstrating `role="status"` for dynamic empty states)_

[Author response: Author directed both fixes — add the story now, and document the `Empty`/`EmptyState` split in TSDoc rather than converging the two components. Convergence (reimplementing `EmptyState` as a wrapper over `Empty`) was considered and deliberately deferred as out of scope for a component-add PR.]

### Minor — Consider

- [x] **`EmptyDescription` typed `React.ComponentProps<'p'>` but renders a `<div>`** (upstream shadcn quirk, present in the pristine baseline) — mistypes `ref` and event handlers. _(fixed during review: retyped to `React.ComponentProps<'div'>` with a `// CUSTOM:` comment explaining the divergence from upstream)_
- [x] **`EmptyMedia` set `data-slot="empty-icon"`** despite the component being named `EmptyMedia` (upstream quirk). _(fixed during review: renamed to `data-slot="empty-media"` with a `// CUSTOM:` comment; verified nothing in source, CSS, tests, or snapshots keys off the old value)_
- [x] **`pr-twp` passed as a separate `cn()` argument** in all six components, where siblings `card.tsx`/`kbd.tsx`/`alert.tsx` prepend it inside the same class string literal — convention drift that makes future upstream re-baselining diffs noisier. _(fixed during review: inlined `pr-twp` into each class string literal; for `EmptyMedia`, which has no plain class literal, it was inlined into the `emptyMediaVariants` cva base string)_
- [x] **The identical 2-line `// CUSTOM: pr-twp` comment repeated verbatim 6×**, where `card.tsx` uses a 1-line variant. _(fixed during review: collapsed all six to `card.tsx`'s exact one-line wording)_
- [x] **Root `Empty` applies `tw:border-dashed` with no border-width class**, so no border renders unless the caller adds `tw:border` (upstream behavior). Easy for a first consumer to ship a rounded-but-borderless container by accident. _(fixed during review: the `DashedBorder` story demonstrates the intended usage and explains the requirement)_
- [ ] ~~`emptyMediaVariants` is not exported from `empty.tsx` or `src/index.ts`, so consumers cannot reuse the variant classes on other elements.~~ _(Author chose to keep it private, matching `alert.tsx`, which also keeps its cva variants private. `variant="icon"` still typechecks for consumers because the flattened `dist/index.d.ts` inlines the variant union, and exporting it later would be a non-breaking addition. Deliberate decision.)_
- [ ] ~~The "pristine baseline" commit `943c377538d` already contains a customization — the `import React from 'react'` line and its `// CUSTOM` comment — eroding the clean-diff value of a separate baseline commit.~~ _(Investigated during review: kbd's own baseline commit `e016e45280b` carries the same React-import `// CUSTOM` comment, so putting the React import in the baseline is the established convention in this repo, not a deviation. Rewriting history would also require a force push to an already-pushed branch. Author chose to match the existing precedent.)_
- [ ] ~~Full TSDoc on every sub-component, where `card.tsx` uses `/** @inheritdoc Card */` for its sub-components.~~ _(The fuller per-component docs are the better outcome — they surface in the generated `dist/index.d.ts` — and each carries the upstream `@see` link. Deliberate divergence from the nearest precedent.)_

[Author response: Author gave a decision on each. Four were fixed in-review; three were dismissed with the rationale recorded above. On the one item where the analysis premise turned out to be wrong (the baseline-commit finding), the author took the evidence and matched the existing kbd convention rather than rewriting pushed history.]

#### Shared Regions Modified

None. No changed source file contains a `#region shared with` marker.

#### Extension Config Changes

## Positive Observations

- **No breaking changes.** Every public export present at the merge base is still present in both the runtime bundle and the type declarations; the change is purely additive, matching the stated purpose.
- **The two-commit structure** (pristine `npx shadcn apply` baseline, then customizations) matches the established `kbd` precedent and makes the customization diff reviewable in isolation — exactly the pattern that keeps future shadcn upgrades tractable.
- **`// CUSTOM` comments are present and specific for every deviation**, including the non-obvious missing-React-import case, which explains *why* the standard file transform didn't apply. Satisfies the shadcn edit-discipline rule.
- **TSDoc with an upstream `@see` link on each exported component**, so the generated `dist/index.d.ts` carries real documentation rather than bare signatures.
- **`dist/` was regenerated, not hand-edited.** The bundle reflects the source customizations, and the renamed `resizable-*` chunks have no dangling references (every file in `dist/` was scanned for the two old chunk names — zero hits).
- **Correct placement.** New file is in `src/components/shadcn-ui/`; the export line sits in the correct alphabetized-by-module-path position between `dropdown-menu` and `input`, with names inside the braces in the structural order used by the neighboring `Card`/`Dialog`/`Command` entries.
- **Customizations are minimal and non-visual** — only the `pr-twp` scope class, the React import, and TSDoc. No upstream class *values* were altered, matching `Guidelines/Design Principles` → "Prefer vanilla over custom."
- **Theme tokens only** (`bg-muted`, `text-muted-foreground`, `text-foreground`, `font-heading`, `text-primary`) — no hardcoded colors, so the component tracks the active theme.
- **RTL-safe**: no physical-direction utilities (`ml-`/`mr-`/`left-`/`right-`), only `text-center` and logical `gap`/`p`/`mb`.
- **Responsive by default**: `tw:w-full tw:min-w-0` on root and content plus `tw:text-balance` and `max-w-sm` caps keep it usable at the 300px webview minimum width.
- **No localization concerns** — the component ships no user-visible strings; all text is caller-supplied via children.
- **No build config, dependency, tooling, or CI changes** in the diff.

## Interview Notes

**Stated purpose**: Small PR to add the shadcn `Empty` component.

1. **Add the Storybook story now** rather than deferring it to the first consuming PR (the `kbd`
   precedent, where the story landed later). This is the stricter reading of
   `Guidelines/UI Review`, and it means the component is reviewable by UX before anything consumes it.
2. **Document the `Empty` / `EmptyState` split in TSDoc** rather than converging the two components.
   The author explicitly considered and rejected reimplementing `EmptyState` as a wrapper over
   `Empty` for this PR. Convergence remains an open option for a follow-up — see Suggested Review
   Focus.
3. **Fix both upstream quirks** (`EmptyDescription` prop type, `EmptyMedia` `data-slot` name) rather
   than propagating them, accepting the extra divergence from upstream in exchange for a correct
   public type and a self-consistent slot name. Both are annotated with `// CUSTOM:` comments per
   the shadcn edit-discipline rule.
4. **Match sibling convention on `pr-twp` placement** (inline in the class literal) and on comment
   wording (`card.tsx`'s one-liner), prioritizing quieter future re-baselining diffs.
5. **Keep `emptyMediaVariants` private**, matching `alert.tsx`.
6. **Do not rewrite the pushed commit history.** When the analysis flagged the React import sitting
   in the "pristine" baseline commit, the review surfaced that kbd's baseline commit
   (`e016e45280b`) does exactly the same thing — so the current shape *is* the repo convention. The
   author chose to match the precedent rather than force-push a rewrite. Worth noting the one real
   deviation from the kbd precedent that remains: this branch's baseline commit also added the
   `src/index.ts` export block (8 lines), which `shadcn apply` does not generate — kbd's baseline
   touched only `kbd.tsx`. Left as-is.

**Unresolved items**: None. Every Critical/Important and Minor finding was either fixed or
dismissed with a stated rationale.

**Areas the author did not understand**: None. The author gave a direct, reasoned decision on every
finding, including reversing course on the history rewrite once the counter-evidence was presented.
No finding was deferred to AI judgment.

**Author's closing note**: Nothing further to flag.

## In-Review Quality Check

All checks green; the only corrective action required was a `dist` rebuild.

- **`dist` rebuild (required)** — `lib/platform-bible-react` commits its build output, and `dist/`
  was stale after the source edits (it still contained `data-slot="empty-icon"` and the old
  `pr-twp` class ordering). `npm run build:platform-bible-react` ran clean; verified `empty-media`
  is now present in `dist/index.js` and `empty-icon` is gone. Seven committed dist files updated
  (+32/−36), consistent with the source edits.
- **`npm run typecheck`** — passed, exit 0, no diagnostics.
- **`npm run lint`** — passed, exit 0. `✖ 4 problems (0 errors, 4 warnings)`; no `lint-fix` needed.
  All four warnings are pre-existing in files untouched by this review:
  `extensions/src/platform-scripture-editor/src/model-text-panel.component.tsx` (28:40, 34:29,
  `import/no-duplicates`) and `.../use-effective-resource-reference-list.hook.ts` (28:3, 116:9,
  `no-console`).
- **prettier (review-touched files only)** — all four files reported `(unchanged)`; nothing to
  reformat.
- **`npm test`** — all workspaces green, exit 0. No test or snapshot referenced
  `data-slot="empty-icon"`, so the slot rename broke nothing.

**Unfixable failures**: None.

## Suggested Review Focus

- [ ] **The two divergences from upstream shadcn** — `EmptyDescription` retyped to
      `React.ComponentProps<'div'>`, and `EmptyMedia`'s `data-slot` renamed `empty-icon` →
      `empty-media`. Both are correctness fixes for real upstream quirks, but each adds a hunk that
      a future `shadcn apply` upgrade will conflict on. Confirm the team prefers correctness over a
      clean re-baseline here — and if so, whether these are worth reporting upstream.
- [ ] **`Empty` vs `EmptyState` long-term plan.** This PR documents the split; it does not resolve
      it. Decide whether `EmptyState` should eventually become a thin wrapper over `Empty` (or be
      deprecated), and whether the guidance belongs in `Guidelines/Component Choices` rather than
      only in TSDoc. Note the accessibility asymmetry: `EmptyState` provides `role="status"`; the
      `Empty` primitives require the caller to add it.
- [ ] **Baseline-commit convention.** kbd's baseline commit contained only the component file; this
      branch's baseline also contains the `src/index.ts` export block. Worth settling explicitly
      what belongs in a "pristine baseline" commit so the next shadcn add is unambiguous — and
      possibly folding the answer into the `add-shadcn-component` skill.
- [ ] **`emptyMediaVariants` kept private.** Sibling precedent is split (`badge`/`button` export
      their cva helpers; `alert` does not). Confirm private is the right default for new components.

AI-assisted — Claude Opus 5 (session URL not recorded)

---

This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2690)


katherinejensen00 (changes requested):
Thanks for adding this component, Jolie!

The following two didn't have clear attach points:
15 · lib/platform-bible-react/scripts/shadcn-transform-utils.ts:108 — Low (follow-up)

The React-import fix has now been hand-applied twice; it belongs in the transform.

The transform is content.replace(/^import \* as React from ['"]react['"]/gm, …) — replace-only, so it no-ops when upstream ships no React import at all. button-group.tsx:1 already carries the identical hand-patch and now empty.tsx:1 does too. The durable fix is a few lines in a script that already has shadcn-transform-utils.test.ts coverage: insert the import when React. is referenced and none exists. Cheap, and it retires the six-line comment at empty.tsx:1.

16 · Merge logistics — Info

Rebuild dist immediately before merging.

I found 11 distinct resizable-*.js chunk hashes across local branches. Chunk adds/deletes auto-merge cleanly while dist/index.js conflicts and gets resolved by taking one side, which can produce an index.js importing a chunk the other side deleted — a runtime module-not-found in the renderer, with CI green (see item 1's root cause). Also worth knowing: the stories never reach Chromatic (chromatic.yml targets ai/main, needs a storybook-review label, and its glob excludes lib/), so the Storybook deliverable gets no automated visual review.

Comment-density note (not a change request): empty.tsx is 92 comment lines to 74 code lines, versus card.tsx's 92-to-26 — but card.tsx carries 13 identical pr-twp repeats of its own, and your wording was changed during review to match it. The convention is the problem, not this file. The two comments worth keeping exactly as they are: :100 and :137. The two worth trimming: the six-line import archaeology at :1 (item 15 retires it) and the four-line // CUSTOM: at :25 that restates the TSDoc directly above it.

Claims I checked and rejected, so they don't get re-raised: no story-only CSS leaks into the bundles (size-10 is 2→2 in index.js, 1→1 in experimental.js); hardcoding className after {...args} matches card.stories.tsx; onClick={fn()} in render matches badge/card; pr-twp survives cn() on both the fast and slow paths and never collides with a pr-* utility; emptyMediaVariants being unexported still typechecks for consumers via the emitted local declare const at dist/index.d.ts:3056; export ordering has no convention to violate; and no code anywhere references data-slot=empty*, so the rename is safe today.

> @katherinejensen00 partially reviewed 17 files and all commit messages, and made 15 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2690#-:-P-Q4e093m3i0LHBnnQe:b-ocmutn)* status: all files reviewed, 14 unresolved discussions (waiting on jolierabideau).

___
*[`-- commits` at r1](https://reviewable.io/reviews/paranext/paranext-core/2690#-P-Q49HY5-9IXIKZqjFk:-P-Q49HY5-9IXIKZqjFl:bn4dl4b):*
NIT 14 · PR description — Low

Three description-level items: no Jira URL, no Storybook links, and one miscited standard.

ui-review.mdx:27 requires the full Jira URL ("not just PT-3901") and :31 requires it in the PR description; the body names only the branch. :20 also asks for Storybook links to referenced UI — this PR adds six stories and cites zero URLs, so the UX reviewer whose sign-off gates Done can't reach either the ticket or the stories. Separately, the body states that Guidelines/Component Choices "warns that unranked duplicate options produce divergent UI." That page says no such thing — its text is "It is an ongoing effort to review and agree on components and variants for common usage." Worth correcting, since reviewers lean on those citations.
___
*[`lib/platform-bible-react/dist/index.js` at r1](https://reviewable.io/reviews/paranext/paranext-core/2690#-P-Q2JLg8IA09AnmAOW4:-P-Q2JLg8IA09AnmAOW5:b-h64jmw) ([raw file](https://github.com/paranext/paranext-core/blob/779ab347d27e6bed00a27235fea8be3b3a0d964e/lib/platform-bible-react/dist/index.js)):*
1 · lib/platform-bible-react/dist/index.js:9980 — High

Merging this PR will ship a Button behavior change that has nothing to do with the Empty component.

Your dist rebuild was correct and necessary, but main's committed bundle was stale: PR #2667 (ff22ea69aa9, "Keep the Button press nudge from clobbering caller translates") changed button.tsx, its tests and its stories, and never rebuilt dist. translateY(1px) appears 0 times in main's dist/index.js and 1 time in yours. Because package.json resolves consumers through dist/, this PR is what actually delivers the pressed-state transform change to every Button in the product — under a title that says "Add shadcn Empty component." If it regresses, git bisect lands here. Not your bug, but please call it out in the PR description so the reviewer knows the Button delta is inherited, not authored. Root cause is that root npm run build never invokes build:platform-bible-react, so CI's "Verify no files changed after build" is a no-op for this package.
___
*[`lib/platform-bible-react/src/components/shadcn-ui/empty.tsx` line 1 at r1](https://reviewable.io/reviews/paranext/paranext-core/2690#-P-Q2RxlEb48BGvAWguA:-P-Q2RxlEb48BGvAWguB:b-bp9do2) ([raw file](https://github.com/paranext/paranext-core/blob/779ab347d27e6bed00a27235fea8be3b3a0d964e/lib/platform-bible-react/src/components/shadcn-ui/empty.tsx#L1)):*
> ```XML
> // CUSTOM: Added missing `import React from 'react';` — upstream shadcn's generated empty.tsx has
> ```

2 · lib/platform-bible-react/src/components/shadcn-ui/empty.tsx:1 — High

The PR description is missing the "must not be squash-merged" warning, and this repo defaults to squash.

.claude/commands/add-shadcn-component.md Step 5 requires the description to carry ⚠️ This PR MUST NOT be squash-merged; the body has zero occurrences of "squash". Meanwhile Git-Guide.md:34 says "Always use 'Squash and merge'" and line 40 lists only new-extension and template merges as exceptions. This has already gone wrong once: kbd.tsx entered main through squash commit 9224356de33 (#2147, single parent) and its baseline e016e45280b never landed. If #2690 is squashed, shadcn-customization-tracker.md:67 resolves empty.tsx's baseline to the squash commit — whose content already includes every customization — so git diff <baseline>..HEAD returns nothing and the git-history half of the audit is disabled for this file permanently. Your // CUSTOM: comments would still be harvested, so nothing is lost today; what's lost is the cross-check that catches future unannotated drift. Worth adding the warning and pinging whoever merges.
___
*[`lib/platform-bible-react/src/components/shadcn-ui/empty.tsx` line 18 at r1](https://reviewable.io/reviews/paranext/paranext-core/2690#-P-Q3Fdz1v-FYMD9SwTp:-P-Q3Fdz1v-FYMD9SwTq:bpb4r4s) ([raw file](https://github.com/paranext/paranext-core/blob/779ab347d27e6bed00a27235fea8be3b3a0d964e/lib/platform-bible-react/src/components/shadcn-ui/empty.tsx#L18)):*
> ```XML
>  *
>  * Use this composition when the zero-state needs media, a heading, or an action. For a plain
>  * one-line "nothing to show" message inside a list, grid, or panel, use {@link EmptyState} instead —
> ```

NIT 9 · lib/platform-bible-react/src/components/shadcn-ui/empty.tsx:18 — Medium-Low

{@link EmptyState} points at a symbol this file neither declares nor imports, and CI never checks it.

typedoc.json sets validation.invalidLink: true with treatWarningsAsErrors: true, but build:docs runs only in publish-docs.yml on push to main — a failure would break main's docs publish after merge. I had cited filter.component.tsx:18 → {@link Badge} as precedent; that turned out to be wrong, since that file imports Badge on line 1. This is the only cross-file {@link} in the whole shadcn-ui/ folder. The body says npm run build:platform-bible-react ran clean, which would include typedoc — if you still have that terminal, confirming exit 0 on npm run build:docs closes it outright.
___
*[`lib/platform-bible-react/src/components/shadcn-ui/empty.tsx` line 29 at r1](https://reviewable.io/reviews/paranext/paranext-core/2690#-P-Q2efN1ZgNlYe1d0Z8:-P-Q2efN1ZgNlYe1d0Z9:b-e6f1gj) ([raw file](https://github.com/paranext/paranext-core/blob/779ab347d27e6bed00a27235fea8be3b3a0d964e/lib/platform-bible-react/src/components/shadcn-ui/empty.tsx#L29)):*
> ```XML
> // overlap in purpose), and a note that these primitives provide no ARIA live region. Added so
> // consumers do not have to guess which empty-state API to reach for.
> function Empty({ className, ...props }: React.ComponentProps<'div'>) {
> ```

NIT 4 · lib/platform-bible-react/src/components/shadcn-ui/empty.tsx:29 — Medium

Nothing pins the three customizations that a future shadcn apply would revert — and the repo just added the pattern for exactly this.

Correcting myself here: I initially cleared "no unit test" because I found only context-menu.test.tsx. My working tree predates the merge base. button.test.tsx is at the merge base, and its comment states the rationale verbatim: "jsdom computes no Tailwind CSS and cannot enter :active, so these pin the class strings a shadcn apply would revert." Without an equivalent for Empty, the next upgrade can revert all six pr-twp additions, the data-slot rename and the prop retype with the suite fully green. This is also the cheapest mitigation for items 3 and 5. Compounding it: .storybook/preview.ts:84 adds pr-twp to document.body, so Storybook renders correctly whether or not the component applies the class — Storybook structurally cannot catch a pr-twp regression.
___
*[`lib/platform-bible-react/src/components/shadcn-ui/empty.tsx` line 29 at r1](https://reviewable.io/reviews/paranext/paranext-core/2690#-P-Q3SFnBSCPUOsgucJR:-P-Q3SFnBSCPUOsgucJS:b-8zy1ae) ([raw file](https://github.com/paranext/paranext-core/blob/779ab347d27e6bed00a27235fea8be3b3a0d964e/lib/platform-bible-react/src/components/shadcn-ui/empty.tsx#L29)):*
> ```XML
> // overlap in purpose), and a note that these primitives provide no ARIA live region. Added so
> // consumers do not have to guess which empty-state API to reach for.
> function Empty({ className, ...props }: React.ComponentProps<'div'>) {
> ```

NIT 10 · lib/platform-bible-react/src/components/shadcn-ui/empty.tsx:29 — Medium-Low

Two in-tree hand-rolled empty states went unmigrated, so today the net effect is one more API rather than one fewer.

<Empty appears nowhere outside its own story, while platform-get-resources/src/home.component.tsx:452 and :468 hand-roll the media/title/description/action shape twice in one file, and resource-picker-dialog.component.tsx:280 and src/renderer/components/projects/project-picker.component.tsx:231 are duplicates of each other. Migrating one of them in a follow-up would both validate the API and start paying down the overlap. Related: the Empty vs EmptyState rule belongs in component-choices.mdx, which currently has no zero-state entry — your own Suggested Review Focus already raises this, so it just needs closing.
___
*[`lib/platform-bible-react/src/components/shadcn-ui/empty.tsx` line 35 at r1](https://reviewable.io/reviews/paranext/paranext-core/2690#-P-Q3YW84VpKLSOju94k:-P-Q3YW84VpKLSOju94l:bf4k1sn) ([raw file](https://github.com/paranext/paranext-core/blob/779ab347d27e6bed00a27235fea8be3b3a0d964e/lib/platform-bible-react/src/components/shadcn-ui/empty.tsx#L35)):*
> ```XML
>       className={cn(
>         // CUSTOM: Added 'pr-twp' to apply Platform.Bible's Tailwind CSS scope isolation.
>         'pr-twp tw:flex tw:w-full tw:min-w-0 tw:flex-1 tw:flex-col tw:items-center tw:justify-center tw:gap-4 tw:rounded-xl tw:border-dashed tw:p-6 tw:text-center tw:text-balance',
> ```

NIT 11 · lib/platform-bible-react/src/components/shadcn-ui/empty.tsx:35 — Low

tw:border-dashed draws nothing on its own, and the workaround lives only in Storybook.

Preflight scoped to .pr-twp sets border-width: 0, so <Empty> with default props renders no outline. The DashedBorder story explains it well, but consumers read dist/index.d.ts and IntelliSense, not stories. One sentence in Empty's TSDoc closes it.
___
*[`lib/platform-bible-react/src/components/shadcn-ui/empty.tsx` line 71 at r1](https://reviewable.io/reviews/paranext/paranext-core/2690#-P-Q2_U32frq07Ei3UOx:-P-Q2_U32frq07Ei3UOy:b-zcni37) ([raw file](https://github.com/paranext/paranext-core/blob/779ab347d27e6bed00a27235fea8be3b3a0d964e/lib/platform-bible-react/src/components/shadcn-ui/empty.tsx#L71)):*
> ```XML
> // CUSTOM: Added TSDoc comment with link to upstream shadcn/ui documentation.
> const emptyMediaVariants = cva(
>   // CUSTOM: Added 'pr-twp' to apply Platform.Bible's Tailwind CSS scope isolation.
> ```

NIT 3 · lib/platform-bible-react/src/components/shadcn-ui/empty.tsx:71 — Medium

The review-fixes commit moved pr-twp into the cva base, which is the one placement the style guide calls out as not applicable.

Commit 238ace491d3 had it as cn('pr-twp', emptyMediaVariants({ variant, className })) — matching alert.tsx:44 and badge.tsx:73, the two other cva components. 779ab347d27 moved it into the emptyMediaVariants base string. Code-Style-Guide.md §"Standard customizations" says pr-twp "Does not apply to … cva variant factories (these are functions, not components)." Behavior is identical today, so this isn't a bug — the cost is upgrade-time: EmptyMedia's own class string now carries no pr-twp, so the next shadcn apply regenerates the factory verbatim and every EmptyMedia silently loses scoped preflight, with no lint or type error. button.tsx:16 and toggle.tsx:18 do it your way, so precedent is genuinely split — but the earlier form matched both the written standard and the nearest cva sibling.
___
*[`lib/platform-bible-react/src/components/shadcn-ui/empty.tsx` line 100 at r1](https://reviewable.io/reviews/paranext/paranext-core/2690#-P-Q2o5P8_g8lBGepzjG:-P-Q2o5QAK6IVBCBzFnX:b-ld2704) ([raw file](https://github.com/paranext/paranext-core/blob/779ab347d27e6bed00a27235fea8be3b3a0d964e/lib/platform-bible-react/src/components/shadcn-ui/empty.tsx#L100)):*
> ```XML
>   return (
>     <div
>       // CUSTOM: Renamed data-slot from upstream's 'empty-icon' to 'empty-media' so the attribute
> ```

NIT 5 · lib/platform-bible-react/src/components/shadcn-ui/empty.tsx:100 — Medium

A sibling component already faced this exact name/slot mismatch and chose to keep upstream's value.

input-group.tsx:160 and :179 keep data-slot="input-group-control" on components named InputGroupInput and InputGroupTextarea. No other file in shadcn-ui/ renames an upstream slot. data-slot is shadcn's cross-component styling contract, so any snippet copied from the shadcn docs that selects [data-slot=empty-icon] will match nothing — silently, with no error, just missing styles — and this becomes a permanent hand re-application on every upgrade. Your reasoning is sound in isolation; the counter-precedent is worth weighing before it sets the rule for the next component.
___
*[`lib/platform-bible-react/src/components/shadcn-ui/empty.tsx` line 117 at r1](https://reviewable.io/reviews/paranext/paranext-core/2690#-P-Q3k2pCjkGpUxTDVhw:-P-Q3k2pCjkGpUxTDVhx:b9mcekv) ([raw file](https://github.com/paranext/paranext-core/blob/779ab347d27e6bed00a27235fea8be3b3a0d964e/lib/platform-bible-react/src/components/shadcn-ui/empty.tsx#L117)):*
> ```XML
>  */
> // CUSTOM: Added TSDoc comment with link to upstream shadcn/ui documentation.
> function EmptyTitle({ className, ...props }: React.ComponentProps<'div'>) {
> ```

NIT 12 · lib/platform-bible-react/src/components/shadcn-ui/empty.tsx:117 — Low

EmptyTitle promises a heading and delivers a <div>, at the same size as its own description.

The TSDoc at :17 tells consumers to reach for this composition "when the zero-state needs media, a heading, or an action," but EmptyTitle renders a bare <div> with no asChild/as escape hatch — a screen-reader user pressing H in a panel whose entire content is an Empty finds nothing. Separately, :123 is tw:text-sm while every sibling title primitive is text-base (card.tsx:55, dialog.tsx:214, drawer.tsx:155), and EmptyDescription is tw:text-sm/relaxed — so title and body are the same size, separated only by font-medium. Both are faithful to the preset baseline, so these are UX questions rather than porting errors — but Guidelines/UI Review requires UX sign-off and the PR records none.
___
*[`lib/platform-bible-react/src/stories/shadcn-ui/empty.stories.tsx` line 20 at r1](https://reviewable.io/reviews/paranext/paranext-core/2690#-P-Q2ujXBc6VgCSXAwOs:-P-Q2ujXBc6VgCSXAwOt:bx5ffxp) ([raw file](https://github.com/paranext/paranext-core/blob/779ab347d27e6bed00a27235fea8be3b3a0d964e/lib/platform-bible-react/src/stories/shadcn-ui/empty.stories.tsx#L20)):*
> ```XML
>   parameters: {
>     docs: {
>       description: {
> ```

NIT 6 · lib/platform-bible-react/src/stories/shadcn-ui/empty.stories.tsx:20 — Medium

The disambiguation prose this PR was written to add never renders in Storybook — in either direction.

parameters.docs.description.component overrides the docgen-extracted TSDoc. Your blurb at :22 drops both load-bearing sentences from empty.tsx:17-21: that EmptyState "takes a single localized message", and "These primitives set no ARIA role, so pass role="status" yourself." The reciprocal side is worse — empty-state.stories.tsx:11 sets its own hand-written blurb that doesn't mention Empty at all, so the paragraph you added at empty-state.component.tsx:21 is invisible too. Storybook is where developers and UX actually browse, so the fix for the overlap currently reaches nobody. The guidance now exists in three hand-maintained copies that already disagree on their first commit.
___
*[`lib/platform-bible-react/src/stories/shadcn-ui/empty.stories.tsx` line 29 at r1](https://reviewable.io/reviews/paranext/paranext-core/2690#-P-Q3-Z5B6qksTvQ2_Vb:-P-Q3-Z5B6qksTvQ2_Vc:b-zhjqyw) ([raw file](https://github.com/paranext/paranext-core/blob/779ab347d27e6bed00a27235fea8be3b3a0d964e/lib/platform-bible-react/src/stories/shadcn-ui/empty.stories.tsx#L29)):*
> ```XML
>     className: { control: 'text' },
>   },
>   decorators: [
> ```

NIT 7 · lib/platform-bible-react/src/stories/shadcn-ui/empty.stories.tsx:29 — Medium

The PR body asserts the component works at 300px, but no story shows it.

responsiveness.mdx:13 mandates designing webviews for a min-width of 300px, and the body claims under Positive Observations that Empty is "usable at the 300px webview minimum width." The single decorator pins all six stories at tw:max-w-lg (512px). At 300px, tw:p-6 consumes 48px, EmptyHeader is the only one of the three containers lacking tw:min-w-0, and EmptyContent column-stacks buttons — none of which a code or UX reviewer can see. There's a ready precedent at tab-toolbar.stories.tsx:225: <AnimatedContainer minWidth="300px" skipAhead="4s">.
___
*[`lib/platform-bible-react/src/stories/shadcn-ui/empty.stories.tsx` line 91 at r1](https://reviewable.io/reviews/paranext/paranext-core/2690#-P-Q3zd43ou5URTO7Jc6:-P-Q3zd43ou5URTO7Jc7:b3el2zg) ([raw file](https://github.com/paranext/paranext-core/blob/779ab347d27e6bed00a27235fea8be3b3a0d964e/lib/platform-bible-react/src/stories/shadcn-ui/empty.stories.tsx#L91)):*
> ```XML
>         <EmptyTitle>Nothing to show</EmptyTitle>
>         <EmptyDescription>
>           The `default` variant of `EmptyMedia` adds no background, so it suits larger
> ```

NIT 13 · lib/platform-bible-react/src/stories/shadcn-ui/empty.stories.tsx:91 — Low

Markdown backticks inside JSX children render as literal backtick characters in the story canvas.

:91 and :142 read The `default` variant of `EmptyMedia` adds no background… on screen. I grepped every story under src/stories/ — there are zero other occurrences, so the convention is markdown only inside docs.description strings. Both stories already repeat the identical prose correctly in docs.description.story (:100, :152), so the in-component copy can just become plain user-facing text.
___
*[`lib/platform-bible-react/src/stories/shadcn-ui/empty.stories.tsx` line 165 at r1](https://reviewable.io/reviews/paranext/paranext-core/2690#-P-Q3A2QBplsb0h8H9u2:-P-Q3A2QBplsb0h8H9u3:beqqcqf) ([raw file](https://github.com/paranext/paranext-core/blob/779ab347d27e6bed00a27235fea8be3b3a0d964e/lib/platform-bible-react/src/stories/shadcn-ui/empty.stories.tsx#L165)):*
> ```XML
>         <EmptyDescription>
>           Add a status role when the zero-state replaces content that changed, so screen readers
>           announce it.
> ```

NIT 8 · lib/platform-bible-react/src/stories/shadcn-ui/empty.stories.tsx:165 — Medium-Low

The a11y guidance recommends the one pattern that reliably does not announce.

role="status" is an aria-live region, and assistive tech announces mutations to a region already in the accessibility tree. In the scenario the docs name — the zero-state replacing content that just changed — the idiomatic React shape mounts region and text in a single commit, and NVDA/JAWS typically stay silent. The LiveRegion story demonstrates the broken pattern as the recommended one. Note the TSDoc at empty.tsx:20 is actually careful (it only says "pass role="status" yourself"); the "so screen readers announce it" promise is in the story blurbs at :165 and :174, and the same over-claim predates you at empty-state.component.tsx:18. Also worth scoping: role="status" on the root wraps EmptyContent's buttons, turning button labels into announced content.

jolierabideau:
Thanks, Katherine — this was an unusually useful review. Every item is addressed below: 12 fixed, 2 deferred with reasoning, 1 closed with evidence. Branch is now merged up to date with `main`.

### Fixed

**1 · Inherited Button delta (High).** Confirmed and called out in the PR description under "Inherited change: Button pressed-state transform", including the root cause (root `npm run build` never invokes `build:platform-bible-react`, so CI's "Verify no files changed after build" is a no-op for this package). Worth its own ticket.

**2 · Missing no-squash warning (High).** Added at the top of the PR description as a bold "do not squash" block with the `kbd.tsx`/`9224356de33` precedent, so whoever merges sees why.

**3 · `pr-twp` in the cva base (Medium).** Reverted to `cn('pr-twp', emptyMediaVariants({ variant }), className)`, matching `alert.tsx:48` and `badge.tsx:73` and the written standard at Code-Style-Guide.md:350. You were right that the earlier form was the one that matched both.

**4 · No test pinning the customizations (Medium).** Added `src/components/shadcn-ui/empty.test.tsx` (9 tests), modelled on `button.test.tsx` including its rationale comment. It pins `pr-twp` on all six slots, both `EmptyMedia` variants (so the cva base can be re-baselined freely), the slot values, and that a caller `className` merges without dropping `pr-twp`. Your `.storybook/preview.ts:84` point is quoted in the file's header comment — Storybook structurally cannot catch this, so the note belongs where the test lives.

**5 · `data-slot` rename (Medium).** Reverted to upstream's `empty-icon`. The `input-group.tsx:160`/`:179` counter-precedent plus "data-slot is shadcn's cross-component styling contract" won it: silently-nonmatching doc snippets and a permanent re-application every upgrade cost more than the name/slot symmetry bought. The comment now records why upstream's value is kept, and `empty.test.tsx` asserts `empty-media` is *not* present so the rename can't creep back in.

**6 · Disambiguation prose reaching nobody (Medium).** Fixed in both directions. `empty.stories.tsx`'s component blurb now carries the load-bearing sentences it had dropped — that `EmptyState` takes a single localized `message`, that the primitives set no ARIA role, plus the border-width and `EmptyTitle`-is-a-`div` caveats — and cross-links to `Basics/EmptyState`. `empty-state.stories.tsx`'s blurb now points at `Empty` reciprocally. I kept the `docs.description.component` override rather than deleting it to let docgen through, because the TSDoc block is separated from the declaration by the `// CUSTOM:` lines and I'd rather not depend on docgen's tolerance for that; the four copies now say the same thing, which is the part that was actually broken.

**7 · No 300px story (Medium).** Added a `NarrowWidth` story pinned to `tw:w-[300px]` with a visible boundary border, using the widest content (icon media + title + description + action) so the `tw:p-6` cost and `EmptyContent`'s column stacking are both visible. I used a fixed width rather than `AnimatedContainer` since a static 300px is what a reviewer needs to judge; happy to switch if you prefer the animated precedent.

**8 · `role="status"` over-claim (Medium-Low).** You're right, and this is the finding I'd have shipped wrong. The story prose no longer promises announcement from a single-commit mount: the doc now says placement matters more than the role, that mounting region and text together typically announces nothing in NVDA/JAWS, that the region must stay mounted across the transition with only its text swapped, and — your scoping point — that `role="status"` on the root turns `EmptyContent`'s button labels into announced content, so prefer it on `EmptyHeader` or the description. The `Empty` TSDoc now says "before the zero-state appears" rather than "replaces content that just changed". `empty-state.component.tsx:18` carries the same over-claim; leaving that one alone as pre-existing and out of this PR's diff, but it's real.

**11 · `border-dashed` draws nothing (Low).** One sentence added to `Empty`'s TSDoc, so it reaches IntelliSense and `dist/index.d.ts`, not just the story.

**13 · Markdown backticks in JSX children (Low).** Both occurrences rewritten as plain user-facing text. The `docs.description.story` copies keep their backticks.

**14 · PR description (Low).** Jira URL added (`https://paratextstudio.atlassian.net/browse/PT-4111`), Storybook deliverables listed, and the miscited `Guidelines/Component Choices` line corrected to what the page actually says. I also folded in your Chromatic finding as a note for the UX reviewer — worth knowing that no `lib/` story gets automated visual review.

**15 · React import belongs in the transform (Low).** Done. `ensureReactImport` in `scripts/shadcn-transform-utils.ts` inserts `import React from 'react';` ahead of the first existing import when content references `React.` and no React import exists (default, namespace, or combined `React, { … }` all recognised as present). Registered as a fourth standard transformation, with 8 new tests in `shadcn-transform-utils.test.ts` covering each shape including the named-bindings-only case. The six-line archaeology comment at `empty.tsx:1` is now two lines pointing at the script. I left `button-group.tsx:1`'s copy alone as outside this diff — trivial follow-up.

**16 · Merge logistics (Info).** Both halves are in the PR description: rebuild `dist` immediately before merging, with your chunk-hash reasoning; and the Chromatic gap flagged for UX.

**Comment density.** Trimmed the two you named: `:1` (retired by item 15) and the four-line block at `:25` restating its own TSDoc, now two lines naming only what the TSDoc adds beyond upstream. `:100` and `:137` kept verbatim as you suggested. Agreed the convention is the problem — that seems like a Code-Style-Guide conversation rather than a per-file one.

### Deferred

**10 · No in-tree consumer (Medium-Low).** Took the cheap half now: `component-choices.mdx` has a new "Zero states" section ranking `EmptyState` (one-line) vs `Empty` (composed), with the border-width, ARIA and heading caveats — so the guidance lives where UX and developers actually look, not only in TSDoc. Migrating `home.component.tsx:452`/`:468` I'm deferring: it pulls extension code into a library-add PR and deserves its own review with a UX look at the resulting screens. Filing it as a follow-up. Fair to say the net today is one more API, and I'd rather that be visible than hidden behind a rushed migration.

**12 · `EmptyTitle` is a `<div>` at `text-sm` (Low).** Documented, not changed. Both facts are faithful to the preset baseline, and "prefer vanilla over custom" plus `Guidelines/UI Review` requiring UX sign-off says I shouldn't unilaterally deviate from the upstream type scale — especially since the sibling comparison (`card.tsx:55`, `dialog.tsx:214`, `drawer.tsx:155` all `text-base`) is a genuine argument for changing it, not for me guessing. So: `Empty`'s TSDoc and the story blurb now state that `EmptyTitle` renders a `<div>` and that callers should nest their own heading when the zero-state is a region's entire content, and the same caveat is in `component-choices.mdx`. The sizing question is on the UX agenda — if UX says `text-base`, it's a one-line change plus a `// CUSTOM:` comment.

### Closed with evidence

**9 · `{@link EmptyState}` (Medium-Low).** Resolves. `npm run build:docs` exits 0 — typedoc's `entryPoints` is `["src/index.ts"]`, so `EmptyState` and `Empty` are in the same project scope and the link doesn't need a file-local import. Full output: `[info] html generated at ./docs`, no warnings, so `treatWarningsAsErrors` never triggers. Thanks for catching that the `filter.component.tsx` precedent was the wrong one — the entry-point scope is the actual reason it works.

Also: thank you for listing the claims you checked and rejected. Knowing `pr-twp` survives `cn()` on both paths, that `size-10` doesn't leak into the bundles, and that `emptyMediaVariants` still typechecks for consumers via the emitted local `declare const` saved me re-deriving all three.

### Verification

`npm run typecheck` exit 0 · `npm run lint` exit 0 (4 pre-existing warnings, none in touched files) · `npm test` all 133+ test files green · `npm run build:docs` exit 0 · `npm run build:platform-bible-react` rerun, and the `dist` diff is exactly the three source changes (`pr-twp` out of the cva base, `empty-icon` slot, the `cn()` reshape) with no chunk churn.

jolierabideau (commented):
> @jolierabideau partially reviewed 22 files and all commit messages, made 10 comments, and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2690)* status: 10 of 22 files reviewed, 10 unresolved discussions (waiting on jolierabideau and katherinejensen00).

___
*[`lib/platform-bible-react/dist/index.js` at r1](https://reviewable.io/reviews/paranext/paranext-core/2690#-P-Q2JLg8IA09AnmAOW4:-P-QvtCcE0vyZ-ynDmIl:b-896fix) ([raw file](https://github.com/paranext/paranext-core/blob/779ab347d27e6bed00a27235fea8be3b3a0d964e/lib/platform-bible-react/dist/index.js)):*

1 · lib/platform-bible-react/dist/index.js:9980 — High

Merging this PR will ship a Button behavior change that has nothing to do with the Empty component.

Your dist rebuild was correct and necessary, but main's committed bundle was stale: PR #2667 (ff22ea69aa9, "Keep the Button press nudge from clobbering caller translates") changed button.tsx, its tests and its stories, and never rebuilt dist. translateY(1px) appears 0 times in main's dist/index.js and 1 time in yours. Because package.json resolves consumers through dist/, this PR is what actually delivers the pressed-state transform change to every Button in the product — under a title that says "Add shadcn Empty component." If it regresses, git bisect lands here. Not your bug, but please call it out in the PR description so the reviewer knows the Button delta is inherited, not authored. Root cause is that root npm run build never invokes build:platform-bible-react, so CI's "Verify no files changed after build" is a no-op for this package.
</blockquote></details>

Done.
___
*[`lib/platform-bible-react/src/components/shadcn-ui/empty.tsx` line 1 at r1](https://reviewable.io/reviews/paranext/paranext-core/2690#-P-Q2RxlEb48BGvAWguA:-P-Qvre80riz9NUwk4cK:b-896fix) ([raw file](https://github.com/paranext/paranext-core/blob/779ab347d27e6bed00a27235fea8be3b3a0d964e/lib/platform-bible-react/src/components/shadcn-ui/empty.tsx#L1)):*

2 · lib/platform-bible-react/src/components/shadcn-ui/empty.tsx:1 — High

The PR description is missing the "must not be squash-merged" warning, and this repo defaults to squash.

.claude/commands/add-shadcn-component.md Step 5 requires the description to carry ⚠️ This PR MUST NOT be squash-merged; the body has zero occurrences of "squash". Meanwhile Git-Guide.md:34 says "Always use 'Squash and merge'" and line 40 lists only new-extension and template merges as exceptions. This has already gone wrong once: kbd.tsx entered main through squash commit 9224356de33 (#2147, single parent) and its baseline e016e45280b never landed. If #2690 is squashed, shadcn-customization-tracker.md:67 resolves empty.tsx's baseline to the squash commit — whose content already includes every customization — so git diff <baseline>..HEAD returns nothing and the git-history half of the audit is disabled for this file permanently. Your // CUSTOM: comments would still be harvested, so nothing is lost today; what's lost is the cross-check that catches future unannotated drift. Worth adding the warning and pinging whoever merges.
</blockquote></details>

Done.
___
*[`lib/platform-bible-react/src/components/shadcn-ui/empty.tsx` line 18 at r1](https://reviewable.io/reviews/paranext/paranext-core/2690#-P-Q3Fdz1v-FYMD9SwTp:-P-QvdJf0CLDBRxd9K6f:b-t0x8xv) ([raw file](https://github.com/paranext/paranext-core/blob/779ab347d27e6bed00a27235fea8be3b3a0d964e/lib/platform-bible-react/src/components/shadcn-ui/empty.tsx#L18)):*

NIT 9 · lib/platform-bible-react/src/components/shadcn-ui/empty.tsx:18 — Medium-Low

{@link EmptyState} points at a symbol this file neither declares nor imports, and CI never checks it.

typedoc.json sets validation.invalidLink: true with treatWarningsAsErrors: true, but build:docs runs only in publish-docs.yml on push to main — a failure would break main's docs publish after merge. I had cited filter.component.tsx:18 → {@link Badge} as precedent; that turned out to be wrong, since that file imports Badge on line 1. This is the only cross-file {@link} in the whole shadcn-ui/ folder. The body says npm run build:platform-bible-react ran clean, which would include typedoc — if you still have that terminal, confirming exit 0 on npm run build:docs closes it outright.
</blockquote></details>

Resolves — closing this with evidence rather than a change. `npm run build:docs` exits 0: `[info] html generated at ./docs`, no warnings, so `treatWarningsAsErrors` never fires.

The reason is `typedoc.json`'s `entryPoints: ["src/index.ts"]` — both `Empty` and `EmptyState` are reachable from that single entry point, so they're in the same project scope and the link resolves without a file-local import. That also explains why the `filter.component.tsx` → `{@link Badge}` precedent looked like it mattered but didn't: the import there is incidental, not what makes the link valid. Which does mean cross-file `{@link}` is fine anywhere under `src/index.ts` — worth knowing, since this being the only one in `shadcn-ui/` reads as a prohibition and isn't one.
___
*[`lib/platform-bible-react/src/components/shadcn-ui/empty.tsx` line 29 at r1](https://reviewable.io/reviews/paranext/paranext-core/2690#-P-Q2efN1ZgNlYe1d0Z8:-P-QuuzF3W8dsZk40t0_:behczvy) ([raw file](https://github.com/paranext/paranext-core/blob/779ab347d27e6bed00a27235fea8be3b3a0d964e/lib/platform-bible-react/src/components/shadcn-ui/empty.tsx#L29)):*

NIT 4 · lib/platform-bible-react/src/components/shadcn-ui/empty.tsx:29 — Medium

Nothing pins the three customizations that a future shadcn apply would revert — and the repo just added the pattern for exactly this.

Correcting myself here: I initially cleared "no unit test" because I found only context-menu.test.tsx. My working tree predates the merge base. button.test.tsx is at the merge base, and its comment states the rationale verbatim: "jsdom computes no Tailwind CSS and cannot enter :active, so these pin the class strings a shadcn apply would revert." Without an equivalent for Empty, the next upgrade can revert all six pr-twp additions, the data-slot rename and the prop retype with the suite fully green. This is also the cheapest mitigation for items 3 and 5. Compounding it: .storybook/preview.ts:84 adds pr-twp to document.body, so Storybook renders correctly whether or not the component applies the class — Storybook structurally cannot catch a pr-twp regression.
</blockquote></details>

Added `empty.test.tsx` (9 tests), modelled on `button.test.tsx` down to the rationale comment. It pins `pr-twp` on all six slots, on both `EmptyMedia` variants separately, and that a caller `className` merges without dropping it. Your `.storybook/preview.ts:84` point is quoted in the file header — that's the reason the test has to exist, so it belongs where someone deleting the test will read it.

Two notes on coverage. The `EmptyDescription` prop retype isn't pinned because it's compile-time only — a revert to `ComponentProps<'p'>` fails `typecheck`, not this suite. And the `EmptyMedia` variant assertions are deliberately per-variant rather than one call: `pr-twp` now sits in `cn()` rather than the cva base (item 3), so a future re-baseline of the factory can't silently drop it from one variant only.
___
*[`lib/platform-bible-react/src/components/shadcn-ui/empty.tsx` line 29 at r1](https://reviewable.io/reviews/paranext/paranext-core/2690#-P-Q3SFnBSCPUOsgucJR:-P-Qvj8wCjdzemNALlnq:btkei4i) ([raw file](https://github.com/paranext/paranext-core/blob/779ab347d27e6bed00a27235fea8be3b3a0d964e/lib/platform-bible-react/src/components/shadcn-ui/empty.tsx#L29)):*

NIT 10 · lib/platform-bible-react/src/components/shadcn-ui/empty.tsx:29 — Medium-Low

Two in-tree hand-rolled empty states went unmigrated, so today the net effect is one more API rather than one fewer.

<Empty appears nowhere outside its own story, while platform-get-resources/src/home.component.tsx:452 and :468 hand-roll the media/title/description/action shape twice in one file, and resource-picker-dialog.component.tsx:280 and src/renderer/components/projects/project-picker.component.tsx:231 are duplicates of each other. Migrating one of them in a follow-up would both validate the API and start paying down the overlap. Related: the Empty vs EmptyState rule belongs in component-choices.mdx, which currently has no zero-state entry — your own Suggested Review Focus already raises this, so it just needs closing.
</blockquote></details>

Took the cheap half now, deferring the migration.

Done: `component-choices.mdx` has a new "Zero states" section that ranks the two — `EmptyState` for the one-line case, `Empty` for the composed case — with the border-width, ARIA and heading caveats. That closes the item my own Suggested Review Focus raised, and it puts the rule where UX and developers actually look rather than only in TSDoc.

Deferred: migrating `home.component.tsx:452`/`:468`. It pulls extension code into a library-add PR, and the resulting screens deserve a UX look rather than being smuggled in under "add a component". Filing it as a follow-up, and taking `resource-picker-dialog.component.tsx:280` / `project-picker.component.tsx:231` with it since those two are duplicates of each other and should converge in the same pass.

Your framing stands: the net effect today is one more API, not one fewer. I'd rather that be visible in the ledger than papered over with a rushed migration.
___
*[`lib/platform-bible-react/src/components/shadcn-ui/empty.tsx` line 100 at r1](https://reviewable.io/reviews/paranext/paranext-core/2690#-P-Q2o5P8_g8lBGepzjG:-P-QvUrxEsrSCxPDaWzw:bw85m4m) ([raw file](https://github.com/paranext/paranext-core/blob/779ab347d27e6bed00a27235fea8be3b3a0d964e/lib/platform-bible-react/src/components/shadcn-ui/empty.tsx#L100)):*

NIT 5 · lib/platform-bible-react/src/components/shadcn-ui/empty.tsx:100 — Medium

A sibling component already faced this exact name/slot mismatch and chose to keep upstream's value.

input-group.tsx:160 and :179 keep data-slot="input-group-control" on components named InputGroupInput and InputGroupTextarea. No other file in shadcn-ui/ renames an upstream slot. data-slot is shadcn's cross-component styling contract, so any snippet copied from the shadcn docs that selects [data-slot=empty-icon] will match nothing — silently, with no error, just missing styles — and this becomes a permanent hand re-application on every upgrade. Your reasoning is sound in isolation; the counter-precedent is worth weighing before it sets the rule for the next component.
</blockquote></details>

Reverted to upstream's `empty-icon`. Your framing decided it — `data-slot` is shadcn's cross-component styling contract, so the cost isn't the asymmetry, it's that a doc snippet selecting `[data-slot=empty-icon]` fails silently, plus a hand re-application on every upgrade. `input-group.tsx:160`/`:179` doing the same thing for `InputGroupInput`/`InputGroupTextarea` settles what the repo convention actually is.

The comment at that line now records _why_ upstream's value is kept, so the next person doesn't "fix" the mismatch. `empty.test.tsx` also asserts `[data-slot="empty-media"]` is **not** present, so the rename can't creep back in unnoticed.
___
*[`lib/platform-bible-react/src/components/shadcn-ui/empty.tsx` line 117 at r1](https://reviewable.io/reviews/paranext/paranext-core/2690#-P-Q3k2pCjkGpUxTDVhw:-P-QvlIy9tM1mdxUDYbk:b-2h183j) ([raw file](https://github.com/paranext/paranext-core/blob/779ab347d27e6bed00a27235fea8be3b3a0d964e/lib/platform-bible-react/src/components/shadcn-ui/empty.tsx#L117)):*

NIT 12 · lib/platform-bible-react/src/components/shadcn-ui/empty.tsx:117 — Low

EmptyTitle promises a heading and delivers a <div>, at the same size as its own description.

The TSDoc at :17 tells consumers to reach for this composition "when the zero-state needs media, a heading, or an action," but EmptyTitle renders a bare <div> with no asChild/as escape hatch — a screen-reader user pressing H in a panel whose entire content is an Empty finds nothing. Separately, :123 is tw:text-sm while every sibling title primitive is text-base (card.tsx:55, dialog.tsx:214, drawer.tsx:155), and EmptyDescription is tw:text-sm/relaxed — so title and body are the same size, separated only by font-medium. Both are faithful to the preset baseline, so these are UX questions rather than porting errors — but Guidelines/UI Review requires UX sign-off and the PR records none.
</blockquote></details>

Documented, not changed — and I want to be explicit that this is a deferral, not a dismissal.

Both facts are faithful to the preset baseline, and your sibling comparison (`card.tsx:55`, `dialog.tsx:214`, `drawer.tsx:155` all `text-base`) is a real argument for changing the size. But `Guidelines/Design Principles` says prefer vanilla over custom and `Guidelines/UI Review` requires UX sign-off, which this PR doesn't have — so me picking `text-base` unilaterally would be substituting my judgment for the sign-off the process asks for.

What I did instead: `Empty`'s TSDoc, the story blurb and `component-choices.mdx` all now state that `EmptyTitle` renders a `<div>` rather than a heading, and that callers should nest their own heading element when the zero-state is a region's entire content. That at least makes the screen-reader gap discoverable from IntelliSense instead of only from reading the source.

Both questions are on the UX agenda. If UX wants `text-base`, it's a one-line change plus a `// CUSTOM:` comment; if they want real heading semantics, that's an `asChild` addition and a larger conversation about whether we diverge from upstream's primitive.
___
*[`lib/platform-bible-react/src/stories/shadcn-ui/empty.stories.tsx` line 20 at r1](https://reviewable.io/reviews/paranext/paranext-core/2690#-P-Q2ujXBc6VgCSXAwOs:-P-QvXFP68CdwyABkbYN:batlthj) ([raw file](https://github.com/paranext/paranext-core/blob/779ab347d27e6bed00a27235fea8be3b3a0d964e/lib/platform-bible-react/src/stories/shadcn-ui/empty.stories.tsx#L20)):*

NIT 6 · lib/platform-bible-react/src/stories/shadcn-ui/empty.stories.tsx:20 — Medium

The disambiguation prose this PR was written to add never renders in Storybook — in either direction.

parameters.docs.description.component overrides the docgen-extracted TSDoc. Your blurb at :22 drops both load-bearing sentences from empty.tsx:17-21: that EmptyState "takes a single localized message", and "These primitives set no ARIA role, so pass role="status" yourself." The reciprocal side is worse — empty-state.stories.tsx:11 sets its own hand-written blurb that doesn't mention Empty at all, so the paragraph you added at empty-state.component.tsx:21 is invisible too. Storybook is where developers and UX actually browse, so the fix for the overlap currently reaches nobody. The guidance now exists in three hand-maintained copies that already disagree on their first commit.
</blockquote></details>

Fixed in both directions, but not the way you might expect — flagging the one thing I didn't do.

`empty.stories.tsx`'s component blurb now carries the two sentences it had dropped (that `EmptyState` takes a single localized `message`; that the primitives set no ARIA role) plus the border-width and `EmptyTitle`\-is-a-`div` caveats, and cross-links to `Basics/EmptyState`. `empty-state.stories.tsx` now points back at `Empty`.

**I kept the `docs.description.component` override rather than deleting it to let docgen through.** Deleting it is the better shape in principle — one source instead of two — but the TSDoc block in `empty.tsx` is separated from the declaration by the `// CUSTOM:` lines, and I didn't want to bet the rendered docs on react-docgen tolerating that without verifying it in a built Storybook. So I fixed the part that was actually broken (four copies disagreeing) and left the copy count alone. If you'd rather I test whether docgen picks it up and delete the override, say so and I'll do that instead — it's the right end state.
___
*[`lib/platform-bible-react/src/stories/shadcn-ui/empty.stories.tsx` line 29 at r1](https://reviewable.io/reviews/paranext/paranext-core/2690#-P-Q3-Z5B6qksTvQ2_Vb:-P-QvZWr0Hy35qDwoQc9:b3rbn05) ([raw file](https://github.com/paranext/paranext-core/blob/779ab347d27e6bed00a27235fea8be3b3a0d964e/lib/platform-bible-react/src/stories/shadcn-ui/empty.stories.tsx#L29)):*

NIT 7 · lib/platform-bible-react/src/stories/shadcn-ui/empty.stories.tsx:29 — Medium

The PR body asserts the component works at 300px, but no story shows it.

responsiveness.mdx:13 mandates designing webviews for a min-width of 300px, and the body claims under Positive Observations that Empty is "usable at the 300px webview minimum width." The single decorator pins all six stories at tw:max-w-lg (512px). At 300px, tw:p-6 consumes 48px, EmptyHeader is the only one of the three containers lacking tw:min-w-0, and EmptyContent column-stacks buttons — none of which a code or UX reviewer can see. There's a ready precedent at tab-toolbar.stories.tsx:225: <AnimatedContainer minWidth="300px" skipAhead="4s">.
</blockquote></details>

Added a `NarrowWidth` story: `tw:w-[300px]` with a visible boundary border, using the widest content (icon media + title + description + action) so the `tw:p-6` cost and `EmptyContent`'s column stacking are both visible at once.

I used a fixed width rather than the `AnimatedContainer minWidth="300px"` precedent you cited. Reasoning: a reviewer judging whether this survives 300px wants to _look_ at 300px, and an animated sweep makes them wait for the frame they care about. The animated form is better for showing where a layout breaks across a range; here the spec is a single number. Easy to switch if you'd rather it match `tab-toolbar.stories.tsx`.

I did not add `tw:min-w-0` to `EmptyHeader` — you're right that it's the only one of the three containers lacking it, but that's a class change to the preset baseline, and the story now makes the consequence visible so it can be a deliberate UX call rather than my guess.
___
*[`lib/platform-bible-react/src/stories/shadcn-ui/empty.stories.tsx` line 165 at r1](https://reviewable.io/reviews/paranext/paranext-core/2690#-P-Q3A2QBplsb0h8H9u2:-P-QvatKC23DpT2_q8uI:bbsa0ut) ([raw file](https://github.com/paranext/paranext-core/blob/779ab347d27e6bed00a27235fea8be3b3a0d964e/lib/platform-bible-react/src/stories/shadcn-ui/empty.stories.tsx#L165)):*

NIT 8 · lib/platform-bible-react/src/stories/shadcn-ui/empty.stories.tsx:165 — Medium-Low

The a11y guidance recommends the one pattern that reliably does not announce.

role="status" is an aria-live region, and assistive tech announces mutations to a region already in the accessibility tree. In the scenario the docs name — the zero-state replacing content that just changed — the idiomatic React shape mounts region and text in a single commit, and NVDA/JAWS typically stay silent. The LiveRegion story demonstrates the broken pattern as the recommended one. Note the TSDoc at empty.tsx:20 is actually careful (it only says "pass role="status" yourself"); the "so screen readers announce it" promise is in the story blurbs at :165 and :174, and the same over-claim predates you at empty-state.component.tsx:18. Also worth scoping: role="status" on the root wraps EmptyContent's buttons, turning button labels into announced content.
</blockquote></details>

This is the finding I'd have shipped wrong, and the correction is a rewrite rather than a tweak. The `LiveRegion` story doc now says placement matters more than the role: that assistive tech announces _mutations_ to a region already in the tree, that mounting region and text in one commit — which is exactly what the static story shows — typically announces nothing in NVDA/JAWS, and that the region has to stay mounted across the transition with only its text swapped. Your scoping point is in there too: `role="status"` on the root makes `EmptyContent`'s button labels announced content, so prefer it on `EmptyHeader` or the description when the zero-state has actions. The visible description no longer promises announcement, and the TSDoc now says "before the zero-state appears" instead of "replaces content that just changed".

One thing left alone deliberately: `empty-state.component.tsx:18` carries the same over-claim, and `EmptyState` _does_ have the same problem — its `role="status"` mounts with its text. It's outside this diff and it's a real bug in shipped guidance, so I'd rather it get its own change than ride along here. Happy to file it.

katherinejensen00 (approved):
> @katherinejensen00 reviewed 12 files and all commit messages, and resolved 10 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2690)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2691, opened 2026-08-18
PT-4111: Show a book-not-available view with a Manage books action

jolierabideau:
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2691)

---

## Before / After

Repro used for all shots: project **TPD** (which has no 1 Samuel — its book files jump from
`08RUT` to `15EZR`), navigated to **1 Samuel 1:1**. The book/chapter picker only lists books the
project already has, so this state is reached the way a user reaches it — carrying a reference in
from elsewhere (a resource in the same scroll group, or switching the project selector while the
reference stays put).

### Simple mode

| Before | After |
| --- | --- |
| Bare sentence, and the remedy points at another application: *"This book does not exist in this project. Use Paratext 9 to create the book and try again."* | *"This book is not part of this project. Ask your project administrator to add it."* — the remedy now names someone the Simple-mode user can actually ask. |
| 01-before-simple | 02-after-simple |

### Power mode

| Before | After |
| --- | --- |
| Identical bare sentence — no title, no action. | A proper zero-state: **Book not in this project** / *"This book is not part of this project yet. You can add it with Manage books."* plus a **Manage books** button. |
| 03-before-power | 04-after-power |

### The Manage books action

Clicking **Manage books** opens the dialog already on the **Create books** section with the missing
book (**1SA**) preselected and scrolled into view, and the CTA reading **Create 1 book in TPD** —
so the user goes from dead end to one click from a fix.

05-after-manage-books-preselected

# Code Review Summary

**Branch**: pt-4111-book-not-available-message

**Base**: origin/main

**Date**: 2026-08-18

**Files changed**: 34

## Overview

Resolves PT-4111 from the PRD *"Simple — Saroj (and Donna) can trust the app"*. When the user
navigates to a book that is not in the current project, the scripture editor now renders a purpose-built
`BookNotAvailableView` zero-state instead of a bare error string. In Simple mode it explains the
situation and points at the project administrator; in Power mode it offers a **Manage books** action
that opens the Manage Books dialog already on the Create section with the missing book preselected and
scrolled into view. Supporting work: a new `Empty` shadcn primitive vendored into
`platform-bible-react`, an `intent` parameter on the `platformScripture.openManageBooks` command, and
launch-parameter plumbing (`initialSection` / `initialSelectedBooks` / `scrollToBook`) through the
Manage Books web view provider, dialog, and book grid.

Four review passes ran. One Critical finding was confirmed by tracing the renderer's web-view service:
the "Manage Books is already open" path silently no-opped, defeating the feature's core affordance. That
was fixed during review with a launch-token mechanism, which also uncovered a second latent bug that
broke preselection even on first launch. Four other Important findings (localization key semantics,
misleading Power-mode copy, and two coverage gaps) were also fixed in-review. All checks are green.

## API Changes

- `lib/platform-bible-react`: added exports `Empty`, `EmptyHeader`, `EmptyMedia`, `EmptyTitle`,
  `EmptyDescription`, `EmptyContent` from new `src/components/shadcn-ui/empty.tsx` (rebuilt `dist/`
  committed alongside).
- `extensions/src/platform-scripture/src/types/platform-scripture.d.ts`: `platformScripture.openManageBooks`
  gained an optional second parameter `intent?: 'createMissingBook'`. Purely additive; first parameter's
  type was normalized from `string | undefined` to `string?` (semantically identical).
- `lib/papi-dts/papi.d.ts`: unchanged. `ScrollGroupScrRef`, the only newly-consumed type, is already
  exported from `@papi/core`.
- Extension-internal exported surfaces (not extension-developer API): `ManageBooksWebViewOptions` gained
  `initialSection`, `initialSelectedBooks`, and `launchToken`; `ManageBooksDialogProps` gained the same
  three; `BookGridSelectorProps` gained `scrollToBook` and `scrollToken`; new module
  `book-not-available-view.component.tsx` exports `BookNotAvailableView`,
  `BOOK_NOT_AVAILABLE_VIEW_STRING_KEYS`, `BookNotAvailableViewStringKey`,
  `BookNotAvailableViewLocalizedStrings`, `BookNotAvailableViewProps`.
- No deletions, parameter removals/reorderings, or type narrowings in any public API surface.

- [x] **Launch parameters never reached an already-open Manage Books dialog** — `main.ts` called
  `papi.webViews.reloadWebView` with a comment claiming that is how `initialSection` /
  `initialSelectedBooks` reach an open dialog. Traced: `reloadWebView` → `openOrReloadWebView`
  (`src/renderer/services/web-view.service-host.ts`) saves new state but does **not** reload the iframe
  (the generated `content` string and per-id nonce are unchanged, so only `onDidUpdateWebView` fires;
  `web-view.component.tsx` re-sets `srcDoc` only on a `content` change). The React root re-renders and
  never unmounts, so the lazy `useState` initializers in `ManageBooksDialog` and the mount-only
  `useLayoutEffect([])` in `BookGridSelector` never re-ran. User-visible: with Manage Books already open,
  the not-available view's button fronted the tab but kept the old section, with no preselection and no
  scroll. _(fixed during review: added a monotonically increasing `launchToken` bumped on every
  `openManageBooks` call, carried in the web view options, scrubbed on restore like the other launch
  keys, and consumed via token-keyed effects in the dialog, grid, and `projectId` seeding. A token is
  required rather than value comparison because two consecutive identical launches produce identical
  parameters. Also fixes the sibling case where `projectId` was seeded by a mount-only initializer, so
  `main.ts`'s "reload updates the existing tab with the new project context" comment now holds. Stale
  comments corrected. ADR-0015 recorded, superseding ADR-0014's delivery mechanism.)_

[Author response: Author asked for the launch-token fix to be implemented immediately. Note for the
reviewer: ADR-0014, written earlier on this branch, explicitly *rejected* a launch token on the stated
premise "the force-reload already does that" — this review disproved that premise, so ADR-0014's point (4)
is now marked superseded.]

- [x] **Localization key had its meaning changed in place** — `%webView_platformScriptureEditor_error_bookNotFoundProject%`
  went from "This book does not exist in this project. Use Paratext 9 to create the book and try again."
  to "This book is not part of this project. Ask your project administrator to add it." Different remedy,
  same key; the `es` value was deleted in the same diff (the symptom, not the fix). Every locale already
  translated in the pipeline would have kept showing the "use Paratext 9" instruction against the new
  English. Flagged independently by three of four analysis passes. _(fixed during review: restored the old
  key's original English **and** its deleted Spanish value; added a new `en`-only key
  `%webView_platformScriptureEditor_bookNotAvailable_simpleMessage%` in the `bookNotAvailable_` family and
  pointed `SIMPLE_MESSAGE_KEY` at it. The old key now has no code consumer; left defined rather than
  deleted, since retiring it is a separate concern.)_
- [x] **Power-mode copy promised an action the user might not have** — the description always read "You
  can add it with Manage books", but the button was gated on `showManageBooksButton`, which the caller
  passes as `!isReadOnlyEffective`. That memo is true not only for read-only projects but also when
  `viewType === 'markers'` and while a Send/Receive sync is in progress — so a Power user on an editable
  project could see the promise with no button. _(fixed during review: the button is now always rendered in
  Power mode, wrapped in `DisabledActionTooltip` exactly as the sibling `EmptyChapterView` does for
  structure protection. A new `manageBooksDisabledReason?: 'readOnly' | 'markersView' | 'syncInProgress'`
  prop drives a per-cause tooltip string; the caller derives it **from** `isReadOnlyEffective` rather than
  re-deriving the conditions, so the disabled state and the stated reason cannot disagree — including under
  the `dev-editableMarkersView` override. `showManageBooksButton`'s inaccurate TSDoc corrected.)_
- [x] **No test for the already-open relaunch path** — precisely the path that was broken. _(fixed during
  review: new `manage-books-dialog.component.test.tsx` covers relaunch with a new token switching section
  and selection, and asserts an unchanged token does not clobber user-made selections;
  `book-grid.component.test.tsx` gained a token-bump re-scroll test; `manage-books.web-view-provider.test.ts`
  now asserts `launchToken` in all three stale-state scrub tests.)_
- [x] **No test that `ManageBooksDialog` honors `initialSection` / `initialSelectedBooks`** — the
  load-bearing seam of the feature was the only untested link in an otherwise well-covered chain. _(fixed
  during review: covered by the new dialog test. Writing it surfaced a second latent bug —
  `useEffect(() => setSelectionsByAction({}), [projectId])` also ran on mount and wiped the
  lazy-initialized preselection, so preselection never worked even on first launch. Now ref-guarded to fire
  only on a real project change, with the launch effect declared after it so a relaunch that also switches
  project wins the commit.)_
- [x] **Missing Storybook stories** for the new `Empty` primitive (26 sibling shadcn stories exist) and for
  `BookNotAvailableView` (four distinct visual states). _(fixed during review: added
  `lib/platform-bible-react/src/stories/shadcn-ui/empty.stories.tsx` — 7 stories, all passing the Storybook
  vitest project — and `book-not-available-view.component.stories.tsx` covering Simple, Power enabled,
  Power disabled by each of the three reasons, and the raw-key fallback.)_

[Author response: Author requested all five be fixed in-review — new localization key, disabled button with
tooltip, and the tests and stories.]

### Minor — Consider

- [x] **`showManageBooksButton` TSDoc described a contract no caller used** _(fixed during review as part of
  the disabled-tooltip change.)_
- [x] **`book-grid.component.test.tsx` patched `Element.prototype` in `beforeAll` with no restore** _(fixed
  during review: `afterAll` now restores `scrollIntoView` and `querySelectorAll`.)_
- [ ] ~~`lib/platform-bible-react/src/index.ts` `Empty*` export line is not alphabetized~~ _(Investigated and
  dismissed: the file orders names within an export statement to mirror the source module's export order, not
  alphabetically — e.g. the `Dialog*`, `Tabs*`, and `getLinesFromUSFM` lines. The `Empty*` line is not
  anomalous, so it was left unchanged.)_
- [ ] `reloadWebView` is declared as taking `ReloadWebViewOptions` (only `bringToFront`), yet the code
  relies on the extra `ManageBooksWebViewOptions` fields being forwarded to `getWebView`. The forwarding is
  real and matches the `openFind` precedent, but the declared type does not express the contract, so nothing
  type-checks it. Widening the options type would be non-breaking.
- [ ] `EmptyDescription` is typed `React.ComponentProps<'p'>` but renders a `<div>` — an upstream shadcn
  quirk deliberately preserved per ADR-0013, now a public type in `platform-bible-react`.
- [ ] `papi.commands.sendCommand(...)` in `platform-scripture-editor.web-view.tsx` is fire-and-forget with
  no `.catch`, so a failure to open Manage Books surfaces only as an unhandled rejection. Matches existing
  precedent at line 1165 of the same file.
- [ ] When `intent === 'createMissingBook'` but `projectId` cannot be resolved, the dialog still opens at
  the narrow 440×720 create-intent float size while showing the project picker with no preselection — the
  narrow layout is chosen unconditionally on intent, while the create-section setup is gated on `projectId`.
- [ ] The same dialog now appears at two very different sizes (440×720 vs 1100×720) depending on how it was
  launched, and the reuse path deliberately does not resize an already-open dialog — worth confirming with UX.
- [ ] `MANAGE_BOOKS_CREATE_INTENT_FLOAT_SIZE` is a named exported constant while its sibling in the same
  ternary stays an inline `{ width: 1100, height: 720 }`.
- [ ] `book-grid.component.tsx` interpolates `scrollToBook` into a CSS selector without `CSS.escape`. Book
  ids are constrained today, so this is inert in practice.
- [ ] The localize-boilerplate quadruple (`*_STRING_KEYS` / `*StringKey` / `*LocalizedStrings` / `localize`)
  is now duplicated verbatim in six editor components. Copying the established pattern was right for this PR;
  a shared `createLocalizer(keys)` helper would remove ~15 lines per component next time.
- [ ] The new `bookNotAvailable_*` keys are `en`-only — expected if translations arrive via the localization
  pipeline, noted only so it is a conscious choice.
- [ ] The `intent` branch in `main.ts` `openManageBooks` itself is untested, consistent with `main.ts` being
  untested elsewhere in this extension; the testable part was correctly extracted to `manage-books-launch.utils.ts`.
- [ ] Power-mode zero-state renders no `EmptyMedia` icon, unlike the standard shadcn `Empty` composition; a
  small muted icon would make it read as a deliberate zero-state rather than an error.

[Author response: Author did not request changes to the remaining minor items; two were fixed as a side
effect of the Important fixes, and the alphabetization item was investigated and dismissed on evidence.]

#### Shared Regions Modified

None. No `#region shared with` markers appear in any changed file.

#### Extension Config Changes

None. All `extensions/` changes are source files and one contributions JSON — no `package.json`,
`tsconfig*.json`, `webpack.config.*`, or `.eslintrc*` was touched, in `extensions/` or elsewhere.

## Positive Observations

- `empty.tsx` follows the shadcn edit discipline exactly — every deviation from upstream carries a
  `// CUSTOM:` comment stating what/what-it-does/why, and the `pr-twp` + TSDoc treatment matches `card.tsx`
  and `alert.tsx` sub-component by sub-component. The deliberately-preserved upstream quirk is documented in
  an ADR rather than silently "fixed", preserving the upstream diff baseline.
- `book-grid.component.tsx` documents the hidden-view case for its `scrollIntoView` layout effect **at the
  call site**, with the reason and a pointer to `useViewVisibility` / `useBcvSyncScroll` for future callers —
  exactly the form `.claude/rules/cross-view-sync-hidden-views.md` asks for, and checkable from the diff alone.
- `resolveMissingBookId` takes `getScrRefForProject` by injection rather than importing `papi` directly,
  keeping the logic unit-testable; its tests cover the literal-ref, scroll-group, absent, and throwing paths.
- The provider scrubs transient launch keys by unconditional assignment and explains why a conditional
  spread would be a bug, citing two existing precedents — and that scrub is pinned by a test that fails if
  the assignment becomes conditional.
- `book-not-available-view.component.tsx` mirrors the neighboring `empty-chapter-view.component.tsx`
  contract precisely, and the now-redundant standalone error key was removed from `EDITOR_LOCALIZED_STRINGS`
  so it isn't listed twice.
- All user-visible text goes through localize keys; no hardcoded UI strings were introduced. Copy follows
  the sentence-case and ellipsis guidelines, and the button label matches `%manageBooks_dialog_title%` exactly
  so the button and the surface it opens share a name.
- Architecture decisions were recorded per the CLAUDE.md rule, including rejected alternatives — which is
  what made the launch-token premise error visible and correctable rather than buried.
- The `lib/platform-bible-react/dist/` churn is a clean full rebuild, and `dist` is intentionally tracked.

## Interview Notes

The author's stated purpose: resolve PT-4111 from the PRD *"Simple — Saroj (and Donna) can trust the app"* —
show a "book not available" view in the scripture editor when the selected book is not in the project, with a
way to open Manage Books preselected on that book.

The interview was run in compressed form at the author's request ("Faster"): all findings were presented at
once and the author responded with five decisions — implement the launch-token fix, use a new localization key,
use a disabled button with a tooltip, add the tests and stories, and nothing further to flag.

The author made clear, decisive calls on every finding and did not defer any decision. Because the interview
was compressed at their request, they were not asked to walk through the design in their own words, so this
review has no direct evidence either way about depth of understanding of the web-view reload mechanics that
produced the Critical finding — the reviewer may want to cover that in the meeting rather than assume it.

No unresolved findings. No areas where the author expressed uncertainty or deferred to AI.

One thing the reviewer should know that did not come from the author: ADR-0014 on this branch documented the
launch-parameter pattern and explicitly listed "a nonce or launch token to make repeat launches re-trigger" as
a **rejected** alternative, reasoning that "the force-reload already does that." That reasoning was wrong, and
the feature's core affordance silently no-opped as a result. ADR-0015 now records the correction and marks
ADR-0014's point (4) superseded. The generalizable lesson, written into the ADR: `reloadWebView` should not be
assumed to remount anything.

## In-Review Quality Check

All checks green after the in-review changes:

- `npm run typecheck` — exit 0, clean across all workspaces.
- `npm run lint` — 0 errors. Four pre-existing warnings remain in `model-text-panel.component.tsx` and
  `use-effective-resource-reference-list.hook.ts`, neither touched by this branch.
- Prettier — `--check` over every changed file (excluding generated `dist/`): "All matched files use Prettier
  code style!"
- `npm test` — exit 0; 414 test files passed across all workspaces, including the 5 new/extended test files
  for this feature (37 tests) and the 7 new `empty.stories.tsx` Storybook tests.
- `npm run build:platform-bible-react` — completed without errors; the regenerated `dist/` files are staged
  (`dist` is intentionally tracked in this repo).

Two workers reported transient failures during concurrent execution (a 207-error eslint run and one
`typecheck` exit 2) that did not reproduce on clean re-runs; the combined verification above was run serially
after all work landed, and is the authoritative result. No `eslint-disable` or `@ts-expect-error` suppressions
were added anywhere in the review fixes.

## Suggested Review Focus

- [ ] **The launch-token mechanism** (`main.ts`, `manage-books.web-view-provider.ts`,
  `manage-books.web-view.tsx`, `manage-books-dialog.component.tsx`, `book-grid.component.tsx`) — this is the
  largest in-review change and it reverses a documented decision. Is a token-keyed re-apply effect the right
  shape, versus remounting via a `key`? The keyed effect was chosen to preserve unrelated in-dialog state.
- [ ] **The scope of the re-apply on relaunch** — it overrides only the launched-to section's selection and
  leaves other sections' user selections intact. Confirm that is the behavior UX wants when a user relaunches
  into a dialog they were already working in.
- [ ] **`reloadWebView` does not remount** — worth discussing as a platform-wide finding, not just a PT-4111
  one. Other features may carry the same assumption, and the declared `ReloadWebViewOptions` type does not
  express the options-forwarding contract that makes this pattern work at all.
- [ ] **The retired localization key** — `%webView_platformScriptureEditor_error_bookNotFoundProject%` now has
  no code consumer but is still defined, with its original English and Spanish restored. Agree on when and how
  it gets retired.
- [ ] **Per-cause disabled tooltips** — the reason is derived *from* `isReadOnlyEffective` so state and reason
  cannot disagree. Confirm the three reasons (`readOnly`, `markersView`, `syncInProgress`) are the right user-
  facing explanations, and that the read-only branch is genuinely reachable (the caller may short-circuit
  resources earlier).
- [ ] **The two dialog sizes** (440×720 on create-intent vs 1100×720 otherwise, with no resize on the reuse
  path) — a UX consistency question that was not raised with UX.

---


katherinejensen00:
## Review notes — PT-4111

Posting these as one comment for now; I'll split them into inline threads shortly.

The zero-state itself is well built. The `Empty` vendoring follows the shadcn discipline exactly, `resolveMissingBookId` is properly injected and tested, and the hidden-view reasoning at the scroll site is the right pattern in the right place.

The one thing I'd read first is **finding 4** — the in-review launch-token change rests on a premise about `reloadWebView` that turns out to be false, and correcting it removes more code than it adds.

---

## Blocking

**1. Book preselection silently fails in Power mode, the only mode with the button**
`extensions/src/platform-scripture/src/manage-books-launch.utils.ts:37` — High

`resolveMissingBookId` returns `undefined` the moment `scrollGroupScrRef` is `undefined`. The editor's provider sets `scrollGroupScrRef: interfaceMode === 'simple' ? 0 : savedWebView.scrollGroupScrRef` (`platform-scripture-editor/src/main.ts:621`), so only Simple mode is guaranteed a value — and the Manage books button is Power-only. A Power editor with no saved scroll group opens Manage Books with nothing selected and no scroll. The platform treats `undefined` as group 0 elsewhere (`use-scroll-group-scr-ref.hook.ts:102`); doing the same here fixes it.

**2. The create-intent launch lands on a dialog whose Apply button is already disabled**
`extensions/src/platform-scripture/src/manage-books-dialog/manage-books-dialog.component.tsx:1256` — High

`createMethod` defaults to `'fromTemplate'` (:639) with `createReferenceId` undefined, and `canApply` excludes exactly that combination. The user clicks "Manage books", arrives on Create with the book ticked, and Apply is greyed out with no indication of what's missing.

**3. Read-only projects are told their book isn't in a "resource", and are denied the action entirely**
`extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx:1919` — High

`isReadOnly` is `!projectForWebView.isEditable` (editor `main.ts:384`), not a resource flag. The new comment above this line ("A resource keeps its own message") cements the wrong model directly over the branch that would need to change.

**4. ADR-0015 records a mechanism that is factually wrong, and the token it justifies never executes**
`.context/standards/Architecture-Decisions.md:506` — High

It states "the generated `content` string and per-id nonce are unchanged." `getWebViewNonce` never enters `content`; `srcNonce = newNonce()` (`web-view.service-host.ts:1866`) does, regenerating on every call and interpolated at :1919, :1929, :1986, :2085-2095. `web-view.component.tsx:587` binds `srcDoc={content}`, so the iframe reloads and the React root is destroyed and recreated. Every token guard seeds its ref from the incoming token at mount, so `launchToken === ref.current` is always true and no effect body ever runs. The feature works because the lazy initializers were correct all along.

Worth flagging the trap, because I fell into it too: `getWebViewNonce(id)` **is** stable per id. It just isn't the nonce that reaches `content`.

Two consequences. First, ADR-0014's original premise was right — ADR-0015 should be reverted rather than ADR-0014, and the token plumbing across five files can come out with it. Second, because the reload remounts, every relaunch destroys attached import files, filter text, presence filter, group-by, copy source and scroll position. A user mid-import who clicks "Manage books" loses every attached file. ADR-0015 rejects `key`-based remount specifically to avoid that and incurs it anyway, so the stated tradeoff is inverted, not merely mis-stated.

This is worth confirming empirically before acting on, since it drives the largest change: open Manage Books, click "Manage books" from a second editor with DevTools open, and watch whether the iframe's `srcdoc` attribute changes.

**5. Spanish speakers lose a string they already had**
`extensions/src/platform-scripture-editor/contributions/localizedStrings.json:62` — High

The retired `%webView_platformScriptureEditor_error_bookNotFoundProject%` has an `es` value (:248). All seven replacement keys are `en`-only, so Spanish users go from a localized sentence to English title, description, button and every tooltip. `Localization-Guide.md:492` lists this under Blocking Issues.

---

## Should fix before merge

**6. The scroll-into-view fires before the project's book list has loaded**
`extensions/src/platform-scripture/src/manage-books-dialog/book-grid.component.tsx:609` — Med-High

The grid gates on `hasLoadedProjects`, but `loadProjects` returns `[]` immediately when `manageBooksApi` is unresolved (`manage-books.web-view.tsx:790`). Create's universe is briefly the full canon; the effect centres on that index, the universe then collapses, and `[scrollToken]` never re-runs. This is the primary first-launch path, and both new tests use synchronous doubles so neither can see it.

**7. The `'readOnly'` disabled reason cannot occur**
`extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx:544` — Med-High

Line 1919 early-returns whenever `isReadOnly`, so `BookNotAvailableView` only ever mounts with `isReadOnly === false`. Same value, same render pass, no async gap. A translated string, a Storybook story and an `it.each` row all assert a state the app cannot produce. The PR body asks reviewers to confirm this branch is reachable — it is not.

Relatedly, the memo's TSDoc says the derivation means state and reason "can never disagree". Its only remaining job is choosing between `syncInProgress` and `markersView`, and when both hold it silently prefers `syncInProgress`.

**8. The "one-shot launch scroll" is really a per-mount scroll**
`extensions/src/platform-scripture/src/manage-books-dialog/manage-books-dialog.component.tsx:2474` — Med-High

`scrollToBook` and `scrollToken` are long-lived props on a conditionally rendered, unkeyed grid (:2441). Filter to nothing, then clear, and the grid remounts and the view yanks back to the launched book with no relaunch.

**9. The scrub runs on every rebuild, not just layout restore**
`extensions/src/platform-scripture/src/manage-books.web-view-provider.ts:114` — Med-High

`web-view.component.tsx:362` reloads every web view on `platform.onDidReloadExtensions`. Installing or enabling an extension mid-flow wipes the launch context. The comment frames the behaviour as restore-only.

**10. Clicking "Manage books" can do nothing at all for up to 30 seconds**
`extensions/src/platform-scripture/src/main.ts:277` — Med-High

`await resolveMissingBookId(...)` is serialized ahead of `openWebView`, and `papi.scrollGroups` initializes via `waitForNetworkObject(..., 30000)`. During the re-arm window that file documents, the click produces no dialog, no spinner, no error. This also violates the util's own TSDoc: "Callers should open the dialog anyway; losing the preselection is a far better outcome than failing to open." Open first, then apply the preselection when it resolves.

**11. A collapsed grid group swallows the launch entirely**
`extensions/src/platform-scripture/src/manage-books-dialog/book-grid.component.tsx:943` — Med-High

Pills render only inside `{!collapsed && (…)}`, and group-by defaults to `'canon'`. If the target book's group was previously collapsed, the `querySelector` at :611 returns null, so there is no scroll — and the footer counts a selection whose pill isn't visible anywhere.

**12. Markers view disables an action that would work fine**
`extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx:546` — Med

`isReadOnlyEffective` folds in `viewType === 'markers'` because the editor canvas isn't editable, but Manage Books is a separate floating dialog whose ability to create books doesn't depend on the editor's view type. With finding 7, only `syncInProgress` is both reachable and accurate.

**13. The markers-view tooltip names a control that doesn't exist**
`extensions/src/platform-scripture-editor/contributions/localizedStrings.json:64` — Low-Med

It says "Switch to the formatted view"; the actual control is labelled "Switch Scripture view" (`%webView_platformScriptureEditor_switchScriptureView%`). So the tooltip is both unnecessary per finding 12 and unfollowable.

**14. Power users can see a flash of the Simple-mode message**
`extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx:1932` — Med

`useSetting('platform.interfaceMode', 'simple')` serves `'simple'` until it resolves. The old code rendered the same string in both modes so this was invisible; this PR branches the whole view on it, so a Power user can read "Ask your project administrator" — wrong advice — before the zero-state replaces it. This same file documents the hazard at :2013-2015 and guards `CharacterMarkerBarOverlay` against it.

**15. Simple mode tells the user to ask an administrator while the action sits in the same tab's own menu**
`extensions/src/platform-scripture-editor/contributions/menus.json:75` — Med

The Manage books item carries no `hiddenInterfaceModes`, so it renders in Simple mode. The sibling comments item at :161 uses exactly that gate, so the pattern is established in this very file. Either gate the menu item or change the Simple copy; right now they contradict each other.

**16. Books outside the dialog's canon range vanish without explanation**
`extensions/src/platform-scripture/src/manage-books-dialog/manage-books-dialog.component.tsx:303` — Med

`DEFAULT_BOOK_IDS` covers 1–102 and `resolveMissingBookId` range-checks nothing. For 3ES, EZA or 5EZ: nothing selected, Apply disabled, scroll no-ops.

**17. Double-click can open two dialogs**
`extensions/src/platform-scripture/src/main.ts:304` — Med

Nothing serializes the existing-instance probe, and this PR inserts a new `await` before it, widening the window. Violates the FN-003 single-instance rule cited at :290.

**18. Squash-merging this PR destroys the shadcn baseline**
`lib/platform-bible-react/src/components/shadcn-ui/empty.tsx:1` — Med

`.claude/commands/add-shadcn-component.md:222` requires the two-commit history *and* that the shadcn addition be its own PR carrying a prominent no-squash warning. Commits `943c377` and `238ace4` are bundled into this 12-commit feature PR, the body carries no warning, and squash is the repo default (`Git-Guide.md:34`, whose carve-out covers only new extensions and template updates). ADR-0013 in this same PR restates the rule the PR is breaking.

**19. CSS selector injection inside a layout effect**
`extensions/src/platform-scripture/src/manage-books-dialog/book-grid.component.tsx:611` — Med

`[data-book="${scrollToBook}"]` with no `CSS.escape`. A quote throws `SyntaxError` during commit with no error boundary, unmounting the dialog to a blank panel. Inert today given constrained book ids, but one line to make safe.

---

## Accessibility

**20. The Simple-mode zero-state is not announced**
`extensions/src/platform-scripture-editor/src/book-not-available-view.component.tsx:101` — Med

Bare `div` and `span`, no `role="status"`. It replaces the editor canvas on navigation, so screen-reader users get nothing. `empty-state.component.tsx:24` exists for this. ADR-0013 records rejecting that component but never records dropping the announcement.

**21. The Power-mode zero-state has no heading and no focus management**
`extensions/src/platform-scripture-editor/src/book-not-available-view.component.tsx:112` — Med

`EmptyTitle` renders a `div` (`empty.tsx:113`), so there's no heading, and nothing moves focus when the editor subtree unmounts.

**22. Focus is destroyed on canvas swap and never restored**
`extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx:1927` — Med

The `!bookExists` branch unmounts the editor tree, so focus falls to `body` and stays there on return. Worth weighing against the sibling `EmptyChapterView`, which deliberately keeps the editor mounted-but-hidden (:2010) and refocuses (:1390) — a precedent this branch diverges from.

**23. The scroll bypasses the grid's roving tabindex**
`extensions/src/platform-scripture/src/manage-books-dialog/book-grid.component.tsx:609` — Med

The scroll never updates `focusedIndex` (:495), the sole `tabIndex={0}` holder (:757), so the first Tab focuses book 0 and scrolls back to the top.

---

## Tests

**24. Six `as never` casts void the scrub tests' contract**
`extensions/src/platform-scripture/src/manage-books.web-view-provider.test.ts:38` — Med

Renaming a field in `ManageBooksWebViewOptions` produces no compile error and no failing test — in the tests written specifically to pin that shape.

**25. The new key family was never wired into the parity guard that exists to catch exactly this**
`extensions/src/platform-scripture-editor/src/book-not-available-view.component.tsx:26` — Med

`localized-strings.test.ts` imports and iterates exported `*_STRING_KEYS` arrays for en/es parity — for `CHARACTER_MARKER_MENU_STRING_KEYS` (:5, :29), `CHARACTER_MARKER_CONTROL_STRING_KEYS` (:6, :79) and `REMOVE_CHARACTER_MARKER_STRING_KEYS` (:7, :127). `BOOK_NOT_AVAILABLE_VIEW_STRING_KEYS` wasn't added. One import closes this and finding 5 together.

**26. The relaunch test certifies a path production never executes**
`extensions/src/platform-scripture/src/manage-books-dialog/manage-books-dialog.component.test.tsx:129` — Med

It `rerender`s a mounted tree, which per finding 4 never happens. It also holds `projectId` fixed and never asserts scroll. Test 1 passes no `launchToken` at all, exercising only the lazy initializers. The real mechanism — lazy initializers after an iframe reload — has no coverage.

**27. The scroll assertions can't tell which element scrolled**
`extensions/src/platform-scripture/src/manage-books-dialog/book-grid.component.test.tsx:99` — Med

`toHaveBeenCalled()` on a prototype-wide spy, and the token test changes book and token together, so "right book", "stale book" and "first pill" are indistinguishable. `mock.contexts` fixes it.

**28. The disabled-click test cannot fail**
`extensions/src/platform-scripture-editor/src/book-not-available-view.component.test.tsx:105` — Low-Med

`fireEvent.click` on a disabled button is inert in jsdom, so it passes for any wiring. The real risk — clicking the focusable `DisabledTooltipWrapper` — is untested.

**29. Test stubs aren't restored**
`extensions/src/platform-scripture/src/manage-books-dialog/manage-books-dialog.component.test.tsx:32` — Low-Med

`window.matchMedia` is assigned in `beforeAll` but absent from `afterAll`, and the `ResizeObserver` stub likewise in both new files — the exact asymmetry this PR fixed elsewhere.

---

## Documentation and cleanup

**30. Settled decisions weren't promoted into the rules the agents enforce**
`.claude/rules/architecture/react-patterns.md:50` — Low-Med

Still asserts the `reloadWebView` half-truth. Also `extension-patterns.md:40` ("single optional argument", now false) and `Component-Selection-Quick-Reference.md:313`, whose "Empty State Handling" section is entirely about form fields, with ADR-0013 unmentioned even though it names PT-4132, PT-4347 and PT-4349 as the next three consumers.

**31. Comment volume is working against readership**
`extensions/src/platform-scripture/src/manage-books-dialog/book-grid.component.tsx:600` — Low-Med

68 inline comment lines against 185 non-TSDoc added lines (37%). The 49 TSDoc lines are mandated by the style guide and aren't at issue. Two hotspots: the 4-line `useLayoutEffect` at :609 carries 12 lines of comment, and `manage-books.web-view.tsx` has 13 inline with 0 TSDoc. The "reloadWebView does not remount" claim appears in five places, and per finding 4 all five are wrong together — which is the cost of duplicating a mechanism claim rather than keeping it in one place. Two comments also assert guarantees that don't hold, at :1917 of the editor web view and :878 of the dialog.

**32. The retired key is orphaned and discards its Spanish instead of redirecting it**
`extensions/src/platform-scripture-editor/contributions/localizedStrings.json:85` — Low-Med

Zero code consumers but still shipping in `en` and `es`, and it's the file's last "use Paratext 9" instruction. No `metadata.fallbackKey` was added, though this file already uses that redirect four times (:4, :7, :10, :13) — a one-line fix that also mitigates finding 5.

**33. Launch failures are silent in both directions**
`extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx:1938` — Low-Med

`sendCommand(...)` floats with no `.catch`, and `manage-books-launch.utils.ts:43` swallows its error unlogged. On the other side, `reloadWebView`'s return is discarded, so a closed tab returns `undefined` and is still treated as success.

**34. The description hardcodes the button label in prose**
`extensions/src/platform-scripture-editor/contributions/localizedStrings.json:62` — Low

`Localization-Guide.md:457` prescribes a `{placeholder}`, and the repo does this twice already (`scripture-text-grid.web-view.tsx:533`, `resource-collection-options.component.tsx:196`). Combined with finding 5, a future Spanish translation reads "Puede añadirlo con Manage books."

**35. `showManageBooksButton` is a prop with one possible value**
`extensions/src/platform-scripture-editor/src/book-not-available-view.component.tsx:77` — Low

The sole caller passes the bare literal. Its TSDoc invites a caller that doesn't exist, and two tests assert the impossible `false` path.

**36. The `openFind` precedent is misdescribed, and the narrow float size keys on a different condition**
`extensions/src/platform-scripture/src/main.ts:315` — Low

`openFind` reloads *conditionally* (:368); this reloads always. `find.web-view-provider.ts:53` still uses the conditional spread ADR-0014 calls a bug. Separately, `floatLayoutSize` keys on `intent` alone (:285) while :275 keys on `intent && projectId`, so `openManageBooks(undefined, 'createMissingBook')` gives a cramped 440px window on the View section.

**37. `launchToken` is documented as monotonically increasing but only needs to be distinct**
`extensions/src/platform-scripture/src/main.ts:240` — Low

All consumers compare with `!==`, so the looser contract is both accurate and more robust across an extension-host restart.

**38. `emptyMediaVariants` and its `VariantProps` aren't exported**
`lib/platform-bible-react/src/index.ts:267` — Low

`EmptyMedia` is exported, so consumers can pass a `variant` they can't type.

---

## Needs a runtime check, not a code read

**39. `Empty`'s Tailwind utilities may never be generated for the extension iframe**
`extensions/src/platform-scripture-editor/src/book-not-available-view.component.tsx:112` — High if real

`tw:font-heading`, `tw:text-balance`, `tw:rounded-xl`, `tw:tracking-tight`, `tw:gap-2.5`, `tw:size-8` and `tw:underline-offset-4` appear in zero hand-written extension sources, and extensions build their own CSS from their own `src/**` (`docs/tailwind.md:13`). They do appear in the generated `temp-build/*.web-view.js` bundles, which sit inside the scanned tree but are gitignored — and Tailwind v4 auto-detection respects `.gitignore`.

I couldn't settle this statically, and Storybook and the jsdom test both render it correctly regardless, so no artifact in this PR can see it. Could you open the app and look at the Power-mode zero-state?

---

## Below the cut — flagging only so they can be dismissed deliberately

- The localize-boilerplate quadruple is now copied six times in one directory.
- The jsdom `:scope > li` shim is duplicated verbatim across both new test files.
- ADR-0013 cites a spec path under gitignored `docs/superpowers/`.
- `%webView_platformScriptureEditor_bookNotAvailable_manageBooksButton%` duplicates `%manageBooks_openMenuItem%`, which already has an `es` value.
- A `'copy'` or `'import'` launch section would be wiped at mount by the `copySourceId` effect; unreachable today only because `main.ts:275` sets `'create'`.

---

## Verified and explicitly not raised

So these don't get re-litigated: the `dist/` rebuild is correct and committed; `reloadWebView` genuinely does forward extra option fields; the `existingId: '?'` probe is not a redundant second `getWebView`; `DisabledActionTooltip` handles the disabled-button case correctly; the frozen float-size constant is safe; the 448px narrow-breakpoint claim in `MANAGE_BOOKS_CREATE_INTENT_FLOAT_SIZE`'s TSDoc is accurate; the reverse-portal's `display: contents` makes the layout correct; and the button label does match `%manageBooks_dialog_title%` exactly, as the PR body says.

---

*Review was AI-assisted (Claude Opus 5). Findings were traced through the renderer's web-view service and verified against the code rather than taken from the PR description; finding 39 is explicitly unverified and flagged as such.*

jolierabideau:
## Review response — all 39 findings + the below-the-cut list

Thanks — this was an unusually useful review, and finding 4 was right to lead with. I traced it before touching anything else and it holds: `srcNonce = newNonce()` regenerates on every `getWebView` and is interpolated into the generated `content`, `web-view.component.tsx` binds `srcDoc={content}`, so the iframe reloads and the React root is destroyed. The service host even carries a standing TODO saying so in as many words ("Generating nonces every time causes webviews to rerender every time `getWebView` is used on an existing webview"). Your trap warning was the useful half — I had conflated `getWebViewNonce(id)` with `srcNonce` exactly as you predicted.

So **ADR-0016 is withdrawn, ADR-0015 restored, and the token plumbing is out of all five files.** I kept ADR-0016 in the log as a withdrawn entry rather than deleting it, because how it was wrong is the instructive part — and I took your process point: the claim is now asserted **once** (in `openManageBooks`, plus as a rule in `react-patterns.md`) with the other sites pointing at it, instead of duplicated into five comments that were all wrong together.

Two notes on PR structure before the findings:

- **Finding 18** is resolved structurally — the vendored component is now its own PR (#2690) and this PR is *based on* it, so squash-merging this one cannot flatten the baseline. **Finding 38** (`emptyMediaVariants` not exported) therefore belongs to #2690 as well; I've left it for that PR rather than reaching across the base.
- I merged the base branch in, so this PR's diff is now just the 17 feature files.

Everything below is done unless marked **Deferred**. `npm run typecheck`, `npm run lint` (repo-wide) and `npm test` (1793 tests) are green.

---

### Blocking

**1. Preselection fails in Power mode** — Fixed. `resolveMissingBookId` now treats `undefined` as scroll group 0, matching `useScrollGroupScrRef`'s own `scrollGroupScrRef ?? 0`. Your framing is what made this land properly: this is the *ordinary* case, not an edge case, so the test now says so.

**2. Create-intent lands on a disabled Apply** — Fixed. A create-missing-book launch seeds `createMethod: 'empty'` instead of `'fromTemplate'`. The test asserts the footer label reads "Create 1 book in WEB" rather than degrading to a bare "Create", which is the tell that `canApply` actually holds.

**3. Read-only projects told their book isn't in a "resource"** — Fixed properly. The branch is now on `platform.isPublished`, not editability — which is the distinction this extension's own `main.ts` already draws in a comment ("Use `platform.isPublished` to classify project-kind … not `platform.isEditable` which is specifically about Scripture-text edit permission"). I had cemented the wrong model directly over the branch that needed changing, as you said.

**4. ADR-0015 records a wrong mechanism** — Fixed, as above. Also recorded the consequence you identified: the remount really does discard attached import files, filter text, presence filter, group-by, copy source and scroll position. That is now in ADR-0015's consequences as an accepted cost with the reasoning, rather than an inverted trade-off claim. And the sibling `projectId` "bug" isn't one — a mount-only initializer is correct *because* the reload remounts.

**5. Spanish speakers lose a string** — Fixed. All six replacement keys have `es` values, and the new key family is wired into `localized-strings.test.ts`'s parity guard (finding 25), so this can't regress silently.

---

### Should fix before merge

**6. Scroll fires before the book list loads** — Fixed. The effect now retries as `items` arrive instead of firing once on mount; each early return is "not yet" rather than "give up". Covered by a test that mounts with `items: []` and asserts nothing scrolled, then supplies the items.

**7. `'readOnly'` cannot occur** — Fixed, and it's now *reachable* rather than deleted: with finding 3's fix, a read-only project lands on the zero-state with the button disabled and that tooltip. You were right that the PR body asked reviewers to confirm a branch that didn't exist. I also removed the memo's "can never disagree" claim, which was no longer true of what it actually did.

**8. Per-mount, not one-shot** — Fixed. Ownership moved up: the dialog holds `pendingScrollBook` and the grid reports back through `onScrolledToBook`, so a filter clearing that remounts the grid can't re-fire it.

**9. Scrub runs on every rebuild** — Comment corrected to state the real scope, including the `onDidReloadExtensions` path. Kept the behavior: since a rebuild remounts the dialog anyway, the launch context could not have survived regardless of what the assignment did — so the honest fix was the comment, not the code.

**10. Click can do nothing for 30s** — Fixed. The lookup is bounded at 1.5s inside `resolveMissingBookId`, with a fake-timer test advancing 30s. I went with bounding rather than open-then-apply because open-then-apply would mean a second reload, and a second reload is a second remount.

**11. Collapsed group swallows the launch** — Fixed. The effect expands the owning group, then the re-run scrolls.

**12. Markers view disables a working action** — Fixed. `manageBooksDisabledReason` no longer derives from `isReadOnlyEffective`; only the two conditions that genuinely prevent adding a book remain.

**13. Markers tooltip names a nonexistent control** — Resolved by 12: the string is gone.

**14. Flash of the Simple-mode message** — Fixed. `useSetting`/`useProjectSetting` both return `isLoading`, so a spinner covers the window until *both* `interfaceMode` and `isPublished` resolve. Thanks for the pointer to the `CharacterMarkerBarOverlay` guard in the same file.

**15. Simple mode contradicts its own menu** — Fixed. `hiddenInterfaceModes: ["simple"]` on the Manage books item, matching the sibling comments item.

**16. Books outside the canon range vanish** — Fixed. `resolveMissingBookId` range-checks against the grid's 1–102 universe and returns undefined rather than ticking an invisible book.

**17. Double-click opens two dialogs** — Fixed. `openManageBooks` chains onto the previous call, so the second call's probe runs after the first instance exists.

**18. Squash destroys the baseline** — Resolved by the PR restructure above.

**19. CSS selector injection** — Fixed with `CSS.escape`, plus a test that a `"`-containing id doesn't throw, since the risk is inert for today's ids and a comment alone wouldn't hold.

---

### Accessibility

**20/21/22** — All fixed. Both modes carry `role="status"`; `EmptyTitle` wraps a real `<h2>`; the region takes focus on mount, guarded on `document.hasFocus()` so navigating from the toolbar's book/chapter control doesn't yank focus out of it. On your `EmptyChapterView` comparison: I diverged deliberately and documented why — that view has a chapter to return to, whereas a missing book has no editable content worth keeping mounted. Your point that ADR-0013 never recorded dropping the announcement was the sharpest one here; the ADR now records that choosing `Empty` *transfers* the ARIA obligation to the caller, and `Component-Selection-Quick-Reference.md` carries it as a rule for the next three consumers.

**23. Scroll bypasses the roving tabindex** — Fixed; the scrolled book receives `focusedIndex`, with a test asserting exactly one `tabindex="0"` and that it's inside the right pill.

---

### Tests

**24.** `as never` gone — both builders typed against `SavedWebViewDefinition` and `ManageBooksWebViewOptions`, plus a key-list test so a launch key added to the type but forgotten in the scrub fails. **25.** Wired in; also asserts the placeholder survives in both locales and that the retired key stays absent. **26.** The rerender test is replaced by an unmount-then-mount that matches the real mechanism, plus a rerender test asserting mount-only semantics are *deliberate*. **27.** `mock.contexts` throughout. **28.** Now clicks the focusable `role="group"` wrapper — the reachable target — with a comment on why the disabled-button click proves nothing in jsdom. **29.** Fixed, and since both files needed the identical set I extracted them into `manage-books-dialog.test-utils.ts` (also closes the duplicated-shim item below the cut).

---

### Documentation and cleanup

**30.** Done: `react-patterns.md` now states that `reloadWebView` remounts and that its return must be checked; `extension-patterns.md` documents the one-shot launch-parameter shape; `Component-Selection-Quick-Reference.md` has a real "Zero states" section (its "Empty State Handling" section was indeed entirely about form fields — renamed to say so).

**31.** Inline-comment ratio is down from 37% to 17%, mostly by deleting the five duplicated mechanism claims. Both comments asserting guarantees that didn't hold are gone.

**32/34.** Done, but not as suggested — worth flagging. I **deleted** the retired key rather than adding a `metadata.fallbackKey`: its content ("Use Paratext 9 to create the book") is the workflow this feature replaces, so redirecting would have served obsolete advice under the new keys. With real `es` values now in place the fallback's mitigation value is gone anyway. A test asserts it stays absent. 34 is done with a `{buttonLabel}` placeholder, so there's one localized spelling of the label feeding both the prose and the control.

**33.** Fixed in all three directions: `sendCommand` has a `.catch`, `reloadWebView`'s return is checked (undefined → open fresh instead of returning an id for a dead tab), and the util's silent-undefined path is logged at the call site.

**35.** `showManageBooksButton` removed. **36.** `floatLayoutSize` now keys on `intent && projectId`, matching the launch parameters; the `openFind` comparison is corrected to say this reloads unconditionally *because* fronting the tab is part of the ask. **37.** Moot — the token is gone.

**Deferred (1):** `find.web-view-provider.ts`'s conditional spread. Real, and the same bug class, but it's a different feature's provider and out of this PR's diff — happy to file it.

---

### Finding 39 — the runtime check

Still verifying; I'll follow up on this comment with the result rather than hold the rest. Your reasoning checks out structurally: the editor extension's `tailwind.css` has **no `@source`**, so v4 auto-detection applies, and I confirmed `temp-build` is gitignored (`extensions/src/platform-scripture-editor/.gitignore:31`). So the extension's own CSS build genuinely cannot see those bundles. The open question is whether PBR's runtime-injected stylesheet covers it, which is what I'm checking in the running app.

---

### Below the cut

- **Localize-boilerplate quadruple (6 copies)** — **Deferred.** Agreed it's past the threshold, but the fix is a shared helper in a lib, which doesn't belong in a feature PR. Worth its own ticket.
- **Duplicated jsdom shim** — Fixed (see 29).
- **ADR citing a gitignored spec path** — Fixed; the entry now says the path isn't citable and reproduces the reasoning inline instead.
- **Button label duplicates `%manageBooks_openMenuItem%`** — **Deferred with reasoning.** Now that both have `es` values the parity concern is gone, and the duplication buys locale independence between an editor string and another extension's menu string. Reasonable to disagree.
- **A `'copy'`/`'import'` launch would be wiped at mount** — Not reachable today, so no behavior change, but I added a comment at the `copySourceId` effect naming it as the trap waiting for the first launch that targets those sections.

---

*Fixes in d8a35d9. Response AI-assisted (Claude Opus 5); finding 4's mechanism was verified by reading the renderer's web-view service rather than taken from the review.*

jolierabideau:
### Finding 39 — resolved: the utilities *are* generated

Settled it at the build-output level, which turned out to be more conclusive than a visual pass. Built `platform-bible-react` and the extensions, then checked the editor extension's own emitted stylesheet (`extensions/dist/platform-scripture-editor/src/main.js`, where `platform-scripture-editor.web-view.scss?inline` lands as a string) for **compiled CSS rules**, not class usages:

| utility | compiled rule present |
| --- | --- |
| `tw:font-heading` | ✅ `.tw\:font-heading{font-family:var(--font-sans)}` |
| `tw:text-balance` | ✅ |
| `tw:rounded-xl` | ✅ |
| `tw:tracking-tight` | ✅ |
| `tw:underline-offset-4` | ✅ |
| `tw:size-8` | ✅ |
| `tw:gap-2.5` | ✅ `.tw\:gap-2\.5{…}` |
| `tw:bg-background` *(control — known to work today)* | ✅ |

Each appears 4 times, exactly matching the control — 4 because the extension registers four web views and each inlines the full utility sheet. So the zero-state is styled.

**Why**, and this is the part worth recording: your two premises were both correct — the extension's `tailwind.css` has no `@source`, so v4 auto-detection applies, and `temp-build` really is gitignored (`extensions/src/platform-scripture-editor/.gitignore:31`). They just aren't load-bearing. Auto-detection roots at the repo boundary, so the scan reaches `lib/platform-bible-react/src/**` — and `empty.tsx` lives there, un-ignored. The utilities are generated from PBR's **source**, which is why they don't depend on the gitignored bundles at all. That also explains the thing that first looked alarming: they appear in zero hand-written *extension* sources, which is true and irrelevant.

Worth noting the near-miss, though: this means any extension web view's Tailwind coverage for a PBR component silently depends on PBR's source being inside the auto-detection root. If PBR were ever consumed as a published package from `node_modules` (Tailwind v4 skips `node_modules` in auto-detection), every shadcn utility used only by PBR would vanish from extension iframes. That's a latent trap independent of this PR — happy to file it if you agree it's worth a ticket.

I did not do the visual pass in the running app, since the compiled-rule check answers the same question with less room for "it looked fine because something else supplied the style". Say the word if you'd still like the screenshot.

katherinejensen00 (commented):
So close! Just three more should fix items and a few nits to consider and then I think this is ready. Thanks for sticking with it!

All 39 findings plus the below-the-cut list are addressed or deliberately deferred with reasoning.
Spot-checking the ones most likely to have been fixed shallowly: the `'readOnly'` reason is now
genuinely reachable (branching on `platform.isPublished` rather than editability is what makes it
so), the scroll effect really does retry rather than fire once, the `pendingScrollBook` ownership
move genuinely fixes the per-mount re-fire, and the double-open serialization actually chains.
ADR-0016-kept-as-withdrawn is the right call — the way it was wrong is the useful part.

Three things below I'd fix before merge; the rest are notes. Nothing here is a regression from the
first round.

---

### Should fix before merge

**A. The `fallbackKey` rationale is backwards, and the wrong premise is now pinned by a test**
`extensions/src/platform-scripture-editor/src/localized-strings.test.ts:184` — Med

The comment (and the PR response) says a `metadata.fallbackKey` redirect "would serve that advice
under the new keys." It's the other direction: `metadata` maps the **old** key to a `fallbackKey`
naming the **new** one, so a consumer still holding `%…_error_bookNotFoundProject%` resolves to the
*new* string. The old value is never served under the new keys — the guide's own worked example says
so in as many words ("Old code using `%oldKey_thatNeedsReplacement%` continues to work (falls back to
new key)", `Localization-Guide.md:353–375`), and the checklist at :483 and Blocking Issue 3 at :493
both prescribe the redirect for a replacement.

Deletion may still be the right answer for an extension-scoped key with no external consumers — but
right now it's justified by a claim that isn't true, and a consumer still on the old key gets a raw
`%…%` token rather than the new sentence. Worth re-deciding on the actual tradeoff, and either way
the test comment needs correcting before it gets copied.

**B. The new "reloadWebView remounts" rule is over-generalized — it's false for URL web views**
`.claude/rules/architecture/react-patterns.md:51` — Med

Stated unconditionally, with "do not assume otherwise", and `extension-patterns.md:42` builds on it
("Consumers read the value in ordinary mount-time initializers, because a reload remounts the web
view"). The nonce only reaches `content` for HTML/React web views. For
`WEB_VIEW_CONTENT_TYPE.URL`, `web-view.service-host.ts:1952–1953` assigns `webViewContent =
webView.content` verbatim — no nonce — and `web-view.component.tsx:586–587` binds `src={content}`,
not `srcDoc`. Same URL in, unchanged `src`, no navigation, no remount.

So a launch parameter added to a URL web view by following this rule would silently never arrive —
the identical failure mode this round was about, one content type over. It's a clause, not a
rewrite: "…for HTML and React web views; a URL web view's `content` carries no nonce, so its `src`
is unchanged and it does not remount."

**C. The `CSS.escape` test cannot fail**
`extensions/src/platform-scripture/src/manage-books-dialog/book-grid.component.test.tsx:152` — Med

`scrollToBook: 'B"AD'` is rendered against `ITEMS` (GEN/EXO/MRK). The effect's first step is
`groups.find(g => g.items.some(item => item.book === scrollToBook))` (`book-grid.component.tsx:620`),
which misses, so it returns at `if (!owningGroup) return` — before the selector is ever built.
Delete `CSS.escape` and the test stays green. Putting the id in `items` is what makes it reach the
`querySelector`. Same non-falsifiable shape as old finding 28, in the test written to close 19.

jolierabideau:
## Review response — round 2 (A, B, C)

@katherinejensen00 All three are fixed in `7716428`. Each one was a case of the *reasoning* being wrong rather than
just the code, so the corrected reasoning is in the artifact each time, not only here.

One flag before the findings: the review body opens with "three more should fix items and **a few
nits to consider**", but the posted comment ends after C — the nits didn't make it into the body,
and there are no inline threads on the PR (`pulls/2691/comments` returns 0). If you meant to
include them, post them and I'll take them the same way; I don't want to merge over a list I never
saw.

---

**A. The `fallbackKey` rationale is backwards** — Fixed, and I reversed the decision, not just the
comment.

You're right on the mechanism, and I checked it rather than taking it: in
`findLocalizationForFallbackLanguageAndOrKey` (`localization.service-host.ts`) the metadata is
looked up **for the requested key**, and `metadata[requested].fallbackKey` is then resolved in the
current language. So `metadata[oldKey] = { fallbackKey: newKey }` serves the **new** value to a
consumer still asking for the **old** key. My rationale had it as the old value leaking under the
new keys, which is the opposite, and it was pinned by a test comment — the worst place for it,
since that's what gets copied.

Once the direction is right, the tradeoff I claimed doesn't exist: the redirect can't resurrect the
"use Paratext 9" advice, so the only thing deletion was buying was a raw `%…%` token for anyone
still on the old key. So the key is back in `metadata` with a redirect, and it keeps no value of
its own in either locale.

Target is `%…_bookNotAvailable_simpleMessage%`, deliberately: it's the only replacement that's a
self-contained sentence in both locales. `…_description%` carries a `{buttonLabel}` placeholder
that a legacy consumer has no value to fill, and the tooltips/title are fragments. The test now
asserts both halves — no value of its own, and the redirect points where it should — and the
comment states the resolution direction with the function that implements it, so the next person
to read it gets the mechanism rather than my inference about it.

**B. The remount rule is over-generalized** — Fixed as a clause, as you scoped it.

Verified before writing: `web-view.service-host.ts`'s `case WEB_VIEW_CONTENT_TYPE.URL` assigns
`webViewContent = webView.content` with no nonce interpolation, and `web-view.component.tsx` binds
`src={shouldUseSrc ? content : undefined}` — same URL in, unchanged `src`, no navigation, no
remount.

`react-patterns.md` now scopes the remount claim to HTML and React web views and carries a second
bullet for the URL case, including the consequence that matters — a mount-only initializer never
sees the new options, so a URL web view needs a re-apply path inside the iframe (or a
deterministically changing URL). `extension-patterns.md`'s dependent sentence carries the same
clause instead of inheriting the unconditional version.

I left the `openManageBooks` comment as-is: it's about a specific React web view and names `srcDoc`
as the mechanism, so it isn't making the general claim. Worth saying, since the whole point of
finding 31 was to assert this once — the general statement now lives only in the rules, and it's
the one that got the qualifier.

**C. The `CSS.escape` test cannot fail** — Fixed, and confirmed by actually breaking it.

Exactly as you traced: `'B"AD'` against `ITEMS` misses the `groups.find`, so it returned at
`if (!owningGroup) return` before the selector existed. The unsafe id is now in `items`, so the
effect reaches the `querySelector`, and the test asserts the scroll landed on that pill — which is
what proves the escaped selector was built *and* matched, rather than just that nothing threw.

Falsifiability checked rather than assumed: with `CSS.escape` removed the test fails with the
`SyntaxError` it's there to prevent, and passes again restored. That's the check I should have run
on the original, and it's the same lesson as old finding 28 — I fixed the shape there and then
reproduced it one test over.

---

`npm test` green: 1821 core, 1529 extensions.

Unrelated, so you don't hit it and think it's mine: `npm run typecheck` currently fails in
`platform-scripture` on `Cannot find module 'rollup/parseAst'` from `node_modules/vite`'s own
types. It reproduces with my changes stashed, so it's environment/dependency drift after the base
merge, not this branch.

---

*Fixes in `7716428`. Response AI-assisted (Claude Opus 5); A's resolution direction and B's URL
content-type path were both read out of the renderer's services rather than taken from the review,
and C's falsifiability was verified by removing `CSS.escape` and watching the test fail.*

katherinejensen00 (approved):
Thanks for looking into those last few items, Jolie! Approved.

PR #2687, opened 2026-08-17
PT-4348: Fix the toolbar sync button's startup state, naming, and cancel

jolierabideau:
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2687)

---

## What this changes

Makes the Simple-mode toolbar Sync button a single truthful sync status with one-click cancel
(PT-4348, under PRD "Saroj (and Donna) can trust the app", NN-4).

- **Correct from the moment it mounts.** The status is seeded from
  `paratextBibleSendReceive.getSyncState`, so a toolbar mounting mid-sync no longer reads "Sync"
  until that sync ends. The seed retries on a 2s interval across a 60s window, because the case it
  exists for — send/receive still activating at cold start — is the case a single attempt cannot
  win. When the window is spent it reports `unknown`, not a confident `idle`.
- **Names what is syncing.** From `syncingProjectIds`, a new upstream field, rather than the
  ticket's `lastRequestedProjectIds` — that one is written on completion only, so mid-sync it names
  the *previous* sync's projects. Core declares the new field **optional**, so this merges
  independently of the upstream release (companion PR
  paratext-bible-internal-extensions#193); a build without it falls back to a bare "Syncing".
- **Cancel in place.** Clicking opens a `Popover` listing the syncing projects with a Cancel
  button, instead of opening the sync-status web view — a second, separately-updating sync surface,
  which is the contradiction NN-4 exists to remove. "View sync details" still reaches that web view
  for the per-project detail (conflicts, failure messages, warnings) nothing else surfaces.
- **Never a green check it can't back.** `failed` and `unknown` statuses are derived from the last
  sync's per-project `resultStatus`. A failed, cancelled, or unreadable sync is reported as such.

Roughly 50 lines of sync logic moved out of `platform-bible-toolbar.tsx` into a new `useSyncStatus`
hook and `SyncStatusButton` component.

## API changes

`src/@types/paratext-bible-send-receive/index.d.ts` only, all additive:

- new exported type `SyncState` (`isSyncing`, `lastResults?`, `lastRequestedProjectIds`,
  `syncingProjectIds?`)
- new command declaration `paratextBibleSendReceive.getSyncState: () => Promise<SyncState>`
- `cancelSync`'s contract now states that an **absent `notificationId` means "cancel whatever sync
  is in progress"**, which implementations must honour

This file is core's mirrored copy of an upstream-owned contract. Its header carries a
[...]

PR #2680, opened 2026-08-14
PT-4342: dock find in simple

mattgetgen:
# Code Review Summary

**Branch**: pt-4342-dock-find-in-simple

**Base**: origin/main

**Date**: 2026-08-14

**Review model**: Claude Opus 5

**Files changed**: 5 at review start; 13 after in-review changes (8 files newly touched during the review)

## Overview

Makes Find a permanent tab in Simple mode's Column 3 instead of a panel opened beside the editor, so
every Find entry point — Ctrl+F and the editor menu's Find item — lands on that one tab and brings it
to the front. The branch adds the tab to the static `simpleLayout`, registers
`platformScripture.find` in `FIXED_LAYOUT_WEBVIEW_GROUPS` so the tab is confined to Column 3's
rc-dock group, and sets `isClosable: interfaceMode === 'power'` in the Find web view provider so the
tab cannot be closed in Simple mode.

During the review the author also asked for Text Collection to sit ahead of Find in Column 3. That
required a mechanism change rather than a reorder: Find was already last in the static layout, but
Text Collection is merged in at runtime from `default-layout-supplement.json` and was *appended*
after every static tab. An optional `insertBeforeWebViewType` was added to the supplement entry
(absent or unmatched still appends), so the final Column 3 order is Bible texts, Commentaries,
Comments, Text Collection, Find — while Text Collection stays flag-gated and disappears cleanly when
its feature flag is off. Five findings were fixed during the review (tooltip, no-project bring-to-
front, hidden-view documentation, catalog accuracy, parallel/fail-safe settings read) plus three
minor cleanups; the project-hydration gap and the focus/Escape interaction work were deliberately
deferred.

## API Changes

**None.** No public API surfaces changed. The diff touches no files under `lib/platform-bible-react/`,
`lib/platform-bible-utils/`, `lib/papi-dts/papi.d.ts`, or any `extensions/src/*/src/types/*.d.ts`.
The extension change adds an `isClosable` (and now `tooltip`) value to the `WebViewDefinition` that
`FindWebViewProvider.getWebView` returns — both are already-existing optional properties of the
public `WebViewDefinitionBase`, so no exported type, signature, or default changed.

In-review changes added one internal, non-exported-API field: the optional
`insertBeforeWebViewType` on `DefaultLayoutSupplementEntry`
(`src/renderer/components/docking/default-layout-supplement.model.ts`). This is renderer-internal and
backward compatible — entries without it behave exactly as before.

### Critical — Must address before merge

[...]

PR #2692, opened 2026-08-18
PT-3216: Pass the editor tab-menu selection to Find

katherinejensen00:
> **Rebased onto current `main`**, squashed to **one commit**. The stacked-PR note is gone — the whole diff is PT-3216. The new ADR renumbered to **0026** (ADR-0024 and ADR-0025 landed on `main` in the meantime).

## Summary

Selecting **Find…** from a scripture editor tab's hamburger menu now carries that tab's selected text into the Find panel, which pre-fills the search box and searches it. Ctrl+F already did this (#2677); the menu item didn't.

| | Before | After |
|---|---|---|
| Ctrl+F with text selected | pre-fills ✅ | pre-fills ✅ (now normalized — see below) |
| Tab menu → **Find…** with text selected | opens empty ❌ | pre-fills ✅ |

**Why it's small:** the work item proposed publishing the selection into the platform's shared store. That turned out to be unnecessary *and* impossible — `sharedStoreService` isn't extension-accessible, and the editor renders its own `TabToolbar` **inside its web view**, so the menu handler already runs in the document that owns the DOM selection. It just needed to read it. ([ADR-0026](.context/standards/Architecture-Decisions.md) records this so the shared-store idea isn't re-derived from the ticket.)

## Where to look

| File | What it does |
|---|---|
| `platform-scripture-editor.web-view.tsx` | **The wiring.** The `platformScripture.openFind` branch in `menuCommandHandler`. |
| `use-selection-snapshot.hook.ts` *(new)* | **The one clever bit.** Remembers the selection a chrome click is about to destroy. See below. |
| `find-trigger.util.ts` | `resolveFindSelectionText` — the single normalizer **every** Find trigger goes through. |
| `use-open-find-shortcut.hook.ts` | Ctrl+F (all scripture tab types) now routes its selection through that normalizer. One line. |
| `find-replace.spec.ts` | New e2e test **+ an incidental repair** — see Notes. |
| `Architecture-Decisions.md` | ADR-0026. |

## The one subtle bit

Choosing a menu item is a pointer press, and a pointer press collapses the selection. A capture-phase `pointerdown` listener snapshots the selection just before that press.

Three bounds keep the remembered value honest (`use-selection-snapshot.hook.ts`):

| Rule | Why |
|---|---|
| Only selections **anchored inside** `.editor-input` are remembered | The chrome has its own selectable text — the reference input, a search box. That isn't scripture and must never reach Find as though the user had selected it in the editor. |
| A press **inside** the editor text **clears** it | They're re-placing the caret — that selection is abandoned, don't resurrect it. (A press on chrome with nothing selected *keeps* it, so a toolbar click before opening the menu doesn't wipe it.) |
| **Reading consumes** it | It exists to bridge one pointer press. Without this, a selection made ten minutes ago could pre-fill — and immediately re-run — a search over whatever term the user had since typed into an open Find panel. |

**Ctrl+F deliberately does not use the snapshot** — only the tab menu does. A keystroke destroys nothing, so the live selection is always the honest answer. The asymmetry is intentional and commented at both call sites so it doesn't get "tidied up" later.

## Verification

- ✅ `format:check`, `typecheck`, `lint` (0 errors), `dotnet test c-sharp-tests/` (1580 passed / 0 failed) — all clean on current `main`.
- ✅ `extensions` **1041/1041** across `platform-scripture-editor` + `platform-scripture/find` — that's where this change lives, and the root `npm test` doesn't reach it (`extensions` has no `test` script).
- ℹ️ Full-suite runs on my Windows box produce a **shifting** set of timeout failures — 18 tests one run, 15 the next, in different files each time (`web-view.service-host`, `window-layout-persistence`, `theme.service-host`, `dialog.service-host`, `network-object-status`, `pt9-css-converter`). All 7 pass together in isolation (129/129), and `platform-bible-react`'s `storybook (chromium)` project times out en masse under its own browser parallelism. This PR touches **no file under `lib/`, `src/`, or `tools/`** — the whole diff is `extensions/`, `e2e-tests/`, and one doc. Flagging it rather than burying it.
- ✅ **The premise behind the snapshot is now confirmed empirically**, in Chromium with trusted pointer events on a stand-in page (contenteditable + a menu that focuses its content on open, like Radix). Instrumented log from the real click sequence:

  ```
  capture pointerdown on #trigger    -> live: "beginning"   snapshot now "beginning"
  trigger click handler              -> live: "beginning"
  capture pointerdown on #menuitem   -> live: "beginning"   snapshot now "beginning"
  MENU ITEM click handler            -> live: ""            snapshot available here: "beginning"
  ```

  The collapse lands between the menu item's `pointerdown` and its `click` — exactly the window the capture-phase listener covers. **The hook is load-bearing, not defensive:** without it the menu handler reads `""`. Inserting a toolbar click before opening the menu behaves identically.
- ✅ Unit tests include ones that fail if the listener moves from capture to bubble phase, if the trim is dropped, if the chrome-anchor rule is removed, or if reading stops consuming.
- ⚠️ **Still unproven inside the real app.** The new e2e test has never executed here, and `test:e2e:isolated` — the only runner reaching `tests/isolated/find/` — is in no CI workflow. The probe above is plain Chromium, not Lexical + Radix.
- ⚠️ **Manual UI check not done** — `refresh.sh`'s CDP path is Linux-only, so the UI couldn't be driven on Windows.

If you can spare two minutes in a running app:
1. Select a word → click a toolbar button → open tab menu → **Find…** → should pre-fill the word.
2. Select a word → open tab menu → **Find…** → should pre-fill the word.

<details>
<summary><b>Notes</b> — incidental scope, and behavior changes to existing paths</summary>

**Incidental repair (12 sites in `find-replace.spec.ts`).** The spec waited on `.tw-tabular-nums` and `p.tw-font-light.tw-text-center` — Tailwind **v3** dash-prefix selectors. This repo builds Tailwind v4 with `prefix(tw)`, so the rendered tokens are `tw:tabular-nums` / `tw:font-light`; the old selectors could never match. This broke the suite's shared `beforeAll`, hence every test in the file. Fixed to the escaped v4 form, confirmed against the components that render them. Happy to split this out if you'd rather.

**Two intentional behavior changes to the already-shipped Ctrl+F path**, both because `resolveFindSelectionText` is now the one place every trigger normalizes:

- **Trimmed.** A double-click word selection in Chromium often grabs a trailing space, and searching `"grace "` finds far less than `"grace"`. Previously the tab menu and the reference panels trimmed but editor Ctrl+F did not.
- **Selections spanning lines are rejected** (Find opens without pre-filling). Find's search box is a single-line `<input>`, so Ctrl+A → Ctrl+F would otherwise flatten a whole chapter into one run-on term, run a doomed project-wide search, push it into the shared search history, and remount an already-open Find panel to do it. This one is a judgment call on a path that shipped in #2677 — say the word and I'll drop it.

**Read-only Resource Viewer** shares this web view and is covered: platform-editor renders `className="editor-input …"` unconditionally, so the rules behave identically there.

**One question raised by the rebase, not acted on.** ADR-0017 (new on `main`, from #2691) says a launch parameter must be written into web view state by *unconditional* assignment, and explicitly rejects "conditionally spreading the key only when present" because it lets a stale value survive a layout restore. Its Source line cites `openFind`'s `selectedText` as the pattern it generalizes — but `buildFindWebViewState` conditionally spreads `initialSearchText`, and three tests in `find-web-view-state.utils.test.ts` pin that behavior on purpose ("does NOT write findSearchTerm for an empty selection (leaves the saved term intact)"). Find seems to genuinely want its term to survive a restore, unlike Manage Books' `intent`. Left alone here — it predates this PR and flipping it would change Find's restore behavior — but worth deciding whether the ADR needs a carve-out or the code needs a change.

**Deferred:** a global "last selection" registry (for a Find entry point that isn't tab-scoped). Not needed today — every Find trigger is tab-scoped, so the triggering tab *is* the focused editor. Recorded in ADR-0026 with the trigger for revisiting.

</details>

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2692)

### DIFF
--- a/.context/standards/Architecture-Decisions.md
+++ b/.context/standards/Architecture-Decisions.md
@@ -1264,3 +1264,57 @@ step, no automation. Just a record.
   answer would have wiped that selection permanently — `useProjectSetting` reports an error as
   loaded, so the error branch has to be recognized on its own.
 - **Source:** PT-3299, review of #2708.
+
+## ADR-0026: A tab’s own web view supplies its selection to Find, rather than a shared selection store
+
+- **Date:** 2026-08-16
+- **Status:** Accepted
+- **Source:** PT-3216 (pass editor text selection to Find), PR #2692; builds on the shared Ctrl+F
+  hook from ADR-0015 / PT-4341.
+- **Context:** Opening Find from a scripture editor tab's menu needed that tab's text selection. The
+  work item proposed publishing the selection into the platform's shared store so a command handler
+  could read it. Two facts made that unnecessary and unavailable: (a) the scripture editor sets
+  `shouldShowToolbar: false` (`extensions/src/platform-scripture-editor/src/main.ts`) and instead
+  renders its own `TabToolbar` *inside* its web view
+  (`extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx`), so its
+  menu handler already runs in the document that owns the selection; (b) `sharedStoreService`
+  (`src/shared/services/shared-store.service.ts`) is explicitly not part of the public API and is not
+  reachable from extensions.
+- **Decision:** Find takes the selection of the tab it was triggered from, read in that tab's own web
+  view via `window.getSelection()` and passed through the existing `platformScripture.openFind`
+  `selectedText` parameter. No cross-tab selection state. `resolveFindSelectionText`
+  (`extensions/src/platform-scripture-editor/src/find-trigger.util.ts`) is the single normalizer
+  every trigger goes through, so all of them apply the same two rules: trim (a double-click word
+  selection often carries a trailing space), and reject anything spanning lines (Find's search box is
+  a single-line input, so a Ctrl+A selection cannot be shown honestly and must not be flattened into
+  a run-on term). The two trigger paths deliberately
+  differ in one respect: the tab menu additionally consults a capture-phase pointer-press snapshot
+  (`use-selection-snapshot.hook.ts`), because the click that opens the dropdown is itself what
+  collapses the selection; Ctrl+F (`use-open-find-shortcut.hook.ts`, ADR-0015) reads only the live
+  selection, because a keystroke destroys nothing, and a fallback there would let a long-abandoned
+  selection pre-fill and immediately re-run a search over whatever term an open Find panel already
+  held.
+- **Alternatives:** (a) Shared store — rejected: not extension-accessible, and unnecessary once the
+  menu handler's location is understood. (b) A global "last selection" registry in the editor
+  extension built on the existing `platformScriptureEditor.onDidSelectionChange` event — deferred:
+  every Find entry point today is tab-scoped, so the triggering tab *is* the focused editor; a
+  registry would add cross-tab state with no current consumer. (c) Deriving the text from the
+  existing `PlatformScriptureEditorWebViewController.getSelection()` — rejected: that carries USJ
+  document offsets, not text, so it would mean re-reading and slicing USJ to recover a string the
+  triggering document already has.
+- **Consequences:** The selection path stays inside one file per tab type and needs no new
+  cross-process state. The snapshot is deliberately narrow, because a remembered selection that
+  outlives the interaction that produced it would pre-fill — and immediately re-run — a search over
+  whatever term the user had since typed into an open Find panel. Three bounds keep it honest: it
+  only remembers selections anchored inside the tab's text content (chrome has its own selectable
+  text — a reference input, a search box — that is not scripture); a press inside that content clears
+  it; and reading it consumes it, so it bridges exactly the one pointer press it exists for.
+  **Revisit** if a Find entry point ever lives outside a tab (a top-toolbar or application-menu Find,
+  or an extension asking "what is selected right now" to enable/disable menu items) — that is the
+  point where alternative (b) becomes the right answer. The snapshot is load-bearing, not defensive: in Chromium, a menu item's `click` handler
+  reads an empty `window.getSelection()` even though the same selection is still live at the
+  capture-phase `pointerdown` on that item — the collapse lands between the two, which is exactly the
+  window the snapshot covers. The end-to-end test written to prove this inside the real editor could
+  not be run in this development environment, and `test:e2e:isolated` (the only runner that reaches
+  `e2e-tests/tests/isolated/find/`) appears in no CI workflow, so that verification gap is closed by
+  a manual pass rather than by automation.
--- a/e2e-tests/tests/isolated/find/find-replace.spec.ts
+++ b/e2e-tests/tests/isolated/find/find-replace.spec.ts
@@ -43,6 +43,18 @@ const NO_MATCH_TERM = 'ZZZQQQXXX_NORESULT_12345';
 /** History debounce delay (ms). Must match HISTORY_DEBOUNCE_DELAY_MS in find.web-view.tsx. */
 const HISTORY_DEBOUNCE_MS = 5_000;
 
+/**
+ * The paragraph Find renders once a search has finished — the results summary, the no-results
+ * message, or the search error.
+ *
+ * The `:not([role="status"])` is load-bearing: the idle / no-open-projects / invalid-query
+ * placeholders are `EmptyState`, which carries the same `tw:text-center tw:font-light` classes the
+ * caller passes it plus `role="status"`. Without the exclusion this selector also matches a panel
+ * that has never run a search, so every "wait until the search finished" gate below would resolve
+ * immediately.
+ */
+const SEARCH_OUTCOME_MESSAGE_SELECTOR = 'p.tw\\:font-light.tw\\:text-center:not([role="status"])';
+
 // ---------------------------------------------------------------------------
 // Worker-level setup — runs once before any test in this file.
 // Opens a scripture editor so the hamburger menu is available throughout
@@ -128,7 +140,7 @@ test.beforeAll(async ({ electronApp }) => {
   const hamburger = editorFrame.locator('button[aria-label="Project"]');
   await expect(hamburger).toBeVisible({ timeout: 15_000 });
   await hamburger.click();
-  const findMenuItem = editorFrame.getByRole('menuitem', { name: /find\.\.\./i });
+  const findMenuItem = editorFrame.getByRole('menuitem', { name: /^find$/i });
   await expect(findMenuItem).toBeVisible({ timeout: 5_000 });
   await findMenuItem.click();
 
@@ -142,8 +154,8 @@ test.beforeAll(async ({ electronApp }) => {
   // Wait up to 90 s for either the counter (results found) or the results/no-results paragraph.
   await expect(
     findFrame
-      .locator('.tw-tabular-nums')
-      .or(findFrame.locator('p.tw-font-light.tw-text-center'))
+      .locator('.tw\\:tabular-nums')
+      .or(findFrame.locator(SEARCH_OUTCOME_MESSAGE_SELECTOR))
       .first(),
   ).toBeVisible({ timeout: 90_000 });
 
@@ -224,8 +236,8 @@ async function openFindPanel(mainPage: Page): Promise<FrameLocator> {
   await expect(hamburger).toBeVisible({ timeout: 15_000 });
   await hamburger.click();
 
-  // Click "Find..." from the dropdown (the Radix DropdownMenuContent portals into the iframe body)
-  const findMenuItem = editorFrame.getByRole('menuitem', { name: /find\.\.\./i });
+  // Click "Find" from the dropdown (the Radix DropdownMenuContent portals into the iframe body)
+  const findMenuItem = editorFrame.getByRole('menuitem', { name: /^find$/i });
   await expect(findMenuItem).toBeVisible({ timeout: 5_000 });
   await findMenuItem.click();
 
@@ -260,7 +272,7 @@ async function fillSearchAndWaitForResults(frame: FrameLocator, term: string): P
   // can take 60–120 s on a cold start while the C# backend loads project data.  Even on a warm
   // engine, a fresh web-view instance re-awaits the PDP resolution hook, so the first search in
   // each new Find panel can still take > 90 s.
-  await expect(frame.locator('.tw-tabular-nums')).toBeVisible({ timeout: 150_000 });
+  await expect(frame.locator('.tw\\:tabular-nums')).toBeVisible({ timeout: 150_000 });
 }
 
 /** Click the X (clear search) button in the search input. */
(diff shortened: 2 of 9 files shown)


jolierabideau:
## Review summary

Reviewed against PT-3216, the PT-4336 PRD, and repo conventions. The core design call is right and the ADR is excellent — reading the tab's own `window.getSelection()` instead of building shared selection state is the correct read of the situation, and ADR-0013 documents *why* the ticket's shared-store idea is both unnecessary and unavailable, which is exactly the kind of thing that gets re-derived painfully six months from now. The capture-phase snapshot is a genuinely clever solve for a real problem, and backing off Ctrl+F from it in `ae40f97` was the right correction.

Three things I'd want resolved before merge, then some design/process notes.

---

### 1. Editor Ctrl+F lost its trim — the most-used trigger is now the one that doesn't trim

The PR description says "the search term is now **trimmed** in the editor *and* the reference panels." That's true for the tab menu and for the reference panels, but **not** for editor Ctrl+F.

- Tab menu → `getMenuFindSelectionText()` → `resolveFindSelectionText` → trims ✅
- Reference panels → `useOpenFindShortcut` → `resolveFindSelectionText(sel, undefined)` → trims ✅
- **Editor Ctrl+F** → `getEditorOpenFindArgs(webViewId, window.getSelection()?.toString())` → only `?? ''`, **no trim** ❌

`ae40f97` routed Ctrl+F around `resolveFindSelectionText` to drop the snapshot fallback, and the trim went with it. Downstream nothing recovers it: `find.web-view.tsx:126` reads `findSearchTerm` straight into `searchTerm`, and `.trim()` only ever appears as an emptiness guard (`:554`, `:919`, `:941`) — the search itself runs on the raw string. So a double-click on "grace" in Chromium (which usually grabs the trailing space) + Ctrl+F searches `"grace "`.

That's the exact failure the trim was added to prevent, left in place on the highest-traffic path — and it's squarely against PT-4336's premise. Saroj double-clicks a word, hits Ctrl+F, and gets far fewer hits than he can see on screen. He has no way to know a space did it. That's a "the app is lying to me" moment, which is what this epic exists to eliminate.

Suggested fix — one line, and it makes `getEditorOpenFindArgs` earn its keep:

```ts
export function getEditorOpenFindArgs(
  webViewId: string,
  selectedText: string | undefined,
): EditorOpenFindArgs {
  return { webViewId, selectedText: resolveFindSelectionText(selectedText, undefined) };
}
```

`useOpenFindShortcut` already passes `undefined` for the snapshot arg for the same reason, so this reads consistently. Worth adding the trailing-space case to `getEditorOpenFindArgs`'s tests — right now both of its tests use clean inputs (`'dog'`, `undefined`), so nothing would have caught this.

### 2. The branch is behind its own base, and the two have partly diverged

`origin/pt-4341-find-from-any-tab` has moved since this branch was cut. Merge-base is `ff22ea6`, but the base branch is now at `f1486e1`, with commits not present here:

```
f1486e1 fix(find): resolve a project for the seeded simple-mode Find tab
834f195 fix(find): address PT-4341 review feedback
4fd6613 refactor(find): extract Find web view state and editor open-find args   ← duplicate of 299a66d
```

That third one is the same commit as this PR's `299a66d`, rebased into #2677. Two consequences:

- The PR note says "review only the top commit," but GitHub/Reviewable is showing **15 commits / 28 files** because of the drift — a reviewer following the instruction and a reviewer following the diff see different things.
- The duplicated refactor will need untangling at merge time.

A rebase onto current `origin/pt-4341-find-from-any-tab` should collapse this to the ~3 commits that are actually PT-3216, and then the "review only the top commit" framing holds.

### 3. The premise behind the snapshot has never been executed anywhere

Flagging this because the PR is admirably upfront about it and I want it weighted properly rather than skimmed past.

The unit tests are good — 8 cases including the capture-vs-bubble one, and they all pass locally (121 tests green across the Find files). But they stub `window.getSelection()` and dispatch a synthetic `pointerdown`, so what they verify is the **snapshot's clear/store/preserve rule**, given an assumed premise. The premise itself — that opening the tab dropdown in Chromium+Lexical actually collapses the selection before the menu item fires — is the load-bearing claim, and nothing executed has confirmed it. jsdom won't; a synthetic event doesn't move a real selection.

And there's no net underneath: `test:e2e:isolated` (the only runner that reaches `tests/isolated/find/`) appears in no workflow — `.github/workflows/test.yml` runs `test:e2e:smoke` only (`:170`, `:183`). So the new spec won't run in CI either.

I don't think this blocks on automation, but it does need **someone to click through it once**. The two cases from the PR description are the right ones, and case 1 is the one that actually distinguishes this implementation from a naive one:

1. Select a word → click a toolbar button → open tab menu → Find… → should pre-fill
2. Select a word → open tab menu → Find… → should pre-fill

If the premise turns out to be wrong (e.g. Lexical restores selection when the menu closes), the snapshot hook is dead weight and case 2 would pass without it — worth knowing before this ships. I have a Mac; happy to run these if that's the blocker.

The Tailwind v3→v4 selector repair is a real fix (this repo builds v4 with `prefix(tw)`, so `.tw-tabular-nums` could never have matched) and I'd keep it in this PR rather than splitting — it's what unbroke the suite's shared `beforeAll`.

---

### Design notes (not blocking)

**Snapshot staleness on the menu path.** `ae40f97`'s reasoning — "a selection made once and followed by unrelated chrome clicks stayed resurrectable indefinitely" — still applies to the tab menu, just with a better excuse. The snapshot is only ever cleared by a press inside `.editor-input`. So: select "grace" → click a toolbar button → wander off for ten minutes → open the tab menu → Find pre-fills "grace" and immediately re-runs it, overwriting whatever was in an open Find panel. The tradeoff is documented in the hook's TSDoc, which is the right call for now, but it's worth deciding deliberately whether a `selectionchange`-to-empty clear (or any bound) belongs here, rather than letting the asymmetry with Ctrl+F be the only guard.

**Special-casing a command inside `menuCommandHandler`.** The other generic menu handlers in the repo (`src/renderer/components/web-view.component.tsx:493`, `hello-rock3.web-view.tsx:384`) are pure pass-through, so `if (projectMenuCommand.command === 'platformScripture.openFind')` is a new shape. It's contained, commented, and honestly fine for one command — but if a second menu command ever needs args, that's the moment to reach for a real contract rather than a second `if`.

**Floating promise on the Ctrl+F path.** The tab-menu call gets `.catch(logger.warn)` and `useOpenFindShortcut` does too, but the Ctrl+F `sendCommand` at `platform-scripture-editor.web-view.tsx:1181` doesn't. Pre-existing, but it's now sitting three lines from a handled one — cheap to make consistent while you're here.

### Pattern conformance ✅

Checked against the repo conventions and this is clean: `use-*.hook.ts` / `*.util.ts` naming with colocated `.test.ts` matches the other ~20 files in `platform-scripture-editor/src/`; the new `find/` subdirectory sits alongside the existing `checks/`, `hooks/`, `manage-books-dialog/` in `platform-scripture/src/`; ADR-0013 is recorded per CLAUDE.md's "Recording Architecture Decisions"; the keyboard shortcuts catalog rule is honored (including removing the hook from the entry once Ctrl+F stopped using it — nice catch); and the hidden-view rule from `.claude/rules/cross-view-sync-hidden-views.md` is explicitly answered at the sync site with a comment explaining why there's no deferred side effect to replay. Extracting `buildFindWebViewState` to dodge the `?inline` import is the right way to get that mapping under test.

### Jira / PRD fit

**PT-3216 DoD** reads "the text of the last / focused selected editor window (be it read-only or not) is passed to the find tab." This delivers *triggering-tab* selection and explicitly defers cross-tab "last focused" (ADR-0013 alternative (b)). I think that's correct — every Find entry point today is tab-scoped, so the triggering tab *is* the focused editor, and the read-only case is covered because the panels get Ctrl+F from #2677 and have no Find menu item to worry about (`menus.json` contributes it only under `platformScriptureEditor.react`). But the ticket text isn't literally satisfied, and PT-3216 has already been "partially delivered" twice. Suggest adding a comment on PT-3216 stating that "last/focused editor" is deliberately out of scope with ADR-0013 as the record, so the ticket can actually close instead of lingering into another sprint.

**PT-4336 NN-1A** lists PT-3216 as the nice-to-have ("Opening Find from a scripture editor with selected scripture text pre-populates the input field — PT9 parity"). With the tab-menu path done, that's satisfied for both triggers — modulo finding #1, which I'd argue is in-scope for the PRD's trust theme rather than a nitpick.

---

**Verdict:** approve once # 1 is fixed and # 2 is rebased; # 3 needs one manual pass by someone with a working UI before this rides to `main` behind #2677. Nice work on the ADR — that's the part I'd most want other tickets to copy.

jolierabideau (changes requested):
Thanks Katherine! Two blockers, left rest and NITs in the summary comment.

> @jolierabideau reviewed 28 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2692#-:-P-Uv-qxDE7RNTQA_jUM:by8gleo)* status: all files reviewed, 1 unresolved discussion (waiting on katherinejensen00).

___
*[`extensions/src/platform-scripture-editor/src/find-trigger.util.ts` line 46 at r1](https://reviewable.io/reviews/paranext/paranext-core/2692#-P-UuxwI-zM16bV5cANb:-P-UuxwI-zM16bV5cANc:b-gb7vgf) ([raw file](https://github.com/paranext/paranext-core/blob/ae40f979202f30e02edf10f24279acfc912cb1cd/extensions/src/platform-scripture-editor/src/find-trigger.util.ts#L46)):*
> ```XML
>   selectedText: string | undefined,
> ): EditorOpenFindArgs {
>   return { webViewId, selectedText: selectedText ?? '' };
> ```

### Editor Ctrl+F lost its trim — the most-used trigger is now the one that doesn't trim

The PR description says "the search term is now **trimmed** in the editor _and_ the reference panels." That's true for the tab menu and for the reference panels, but **not** for editor Ctrl+F.

-   Tab menu → `getMenuFindSelectionText()` → `resolveFindSelectionText` → trims ✅
-   Reference panels → `useOpenFindShortcut` → `resolveFindSelectionText(sel, undefined)` → trims ✅
-   **Editor Ctrl+F** → `getEditorOpenFindArgs(webViewId, window.getSelection()?.toString())` → only `?? ''`, **no trim** ❌

`ae40f97` routed Ctrl+F around `resolveFindSelectionText` to drop the snapshot fallback, and the trim went with it. Downstream nothing recovers it: `find.web-view.tsx:126` reads `findSearchTerm` straight into `searchTerm`, and `.trim()` only ever appears as an emptiness guard (`:554`, `:919`, `:941`) — the search itself runs on the raw string. So a double-click on "grace" in Chromium (which usually grabs the trailing space) + Ctrl+F searches `"grace "`.

That's the exact failure the trim was added to prevent, left in place on the highest-traffic path — and it's squarely against PT-4336's premise. Saroj double-clicks a word, hits Ctrl+F, and gets far fewer hits than he can see on screen. He has no way to know a space did it. That's a "the app is lying to me" moment, which is what this epic exists to eliminate.

Suggested fix — one line, and it makes `getEditorOpenFindArgs` earn its keep:

```ts
export function getEditorOpenFindArgs(
  webViewId: string,
  selectedText: string | undefined,
): EditorOpenFindArgs {
  return { webViewId, selectedText: resolveFindSelectionText(selectedText, undefined) };
}
```

`useOpenFindShortcut` already passes `undefined` for the snapshot arg for the same reason, so this reads consistently. Worth adding the trailing-space case to `getEditorOpenFindArgs`'s tests — right now both of its tests use clean inputs (`'dog'`, `undefined`), so nothing would have caught this.

katherinejensen00 (commented):
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2692)* status: 0 of 28 files reviewed, 1 unresolved discussion (waiting on jolierabideau).

___
*[`extensions/src/platform-scripture-editor/src/find-trigger.util.ts` line 46 at r1](https://reviewable.io/reviews/paranext/paranext-core/2692#-P-UuxwI-zM16bV5cANb:-P-qD62lExV3g4UQmWTF:b-y6ms7p) ([raw file](https://github.com/paranext/paranext-core/blob/ae40f979202f30e02edf10f24279acfc912cb1cd/extensions/src/platform-scripture-editor/src/find-trigger.util.ts#L46)):*

### Editor Ctrl+F lost its trim — the most-used trigger is now the one that doesn't trim

The PR description says "the search term is now **trimmed** in the editor _and_ the reference panels." That's true for the tab menu and for the reference panels, but **not** for editor Ctrl+F.

-   Tab menu → `getMenuFindSelectionText()` → `resolveFindSelectionText` → trims ✅
-   Reference panels → `useOpenFindShortcut` → `resolveFindSelectionText(sel, undefined)` → trims ✅
-   **Editor Ctrl+F** → `getEditorOpenFindArgs(webViewId, window.getSelection()?.toString())` → only `?? ''`, **no trim** ❌

`ae40f97` routed Ctrl+F around `resolveFindSelectionText` to drop the snapshot fallback, and the trim went with it. Downstream nothing recovers it: `find.web-view.tsx:126` reads `findSearchTerm` straight into `searchTerm`, and `.trim()` only ever appears as an emptiness guard (`:554`, `:919`, `:941`) — the search itself runs on the raw string. So a double-click on "grace" in Chromium (which usually grabs the trailing space) + Ctrl+F searches `"grace "`.

That's the exact failure the trim was added to prevent, left in place on the highest-traffic path — and it's squarely against PT-4336's premise. Saroj double-clicks a word, hits Ctrl+F, and gets far fewer hits than he can see on screen. He has no way to know a space did it. That's a "the app is lying to me" moment, which is what this epic exists to eliminate.

Suggested fix — one line, and it makes `getEditorOpenFindArgs` earn its keep:

```ts
export function getEditorOpenFindArgs(
  webViewId: string,
  selectedText: string | undefined,
): EditorOpenFindArgs {
  return { webViewId, selectedText: resolveFindSelectionText(selectedText, undefined) };
}
```

`useOpenFindShortcut` already passes `undefined` for the snapshot arg for the same reason, so this reads consistently. Worth adding the trailing-space case to `getEditorOpenFindArgs`'s tests — right now both of its tests use clean inputs (`'dog'`, `undefined`), so nothing would have caught this.
</blockquote></details>

Already fixed. I think this is outdated. I probably forgot to respond to it earlier.

jolierabideau:
### Code review

Re-reviewed at `2cec4b8`. Both code blockers from round 1 are fixed: Ctrl+F trims again (routed through `resolveFindSelectionText`, with the trailing-space regression test), and the branch is squashed to one commit off `main`. Resolving the trim thread — your "already fixed" reply is right.

No code bugs found this round. The things most worth doubting all check out: omitting `sourceProjectId` is valid because `resolveFindInvocation` falls back to the editor's own `webViewDefinition.projectId`; the new TSDoc matches the implementation; and consuming the snapshot eagerly in `getMenuFindSelectionText` is the intended one-shot semantics. Both e2e "repairs" are fixes of genuinely stale selectors, not incidental churn — the `tw-` → `tw:` rename landed in `8fa368b8c76`, and the menu label changed from "Find..." to "Find" in `2b31b0e8820` (PT-4342, #2680), so the old regex could never have matched.

Found 2 issues, both from one root cause:

1. `ADR-0024` collides with an existing ADR. `main` already has `## ADR-0024: The toolbar's sync status is local renderer UI...` (PT-4348, #2687), so this appends a second one. The "renumbered to 0024" note in the description was correct when written and has gone stale again. Heads up that **0025 is also taken** — by `pt-3299-exclude-extra-books-from-find`, which is in flight — so **0026** is the first safe number assuming that one lands first.

https://github.com/paranext/paranext-core/blob/2cec4b8c46be34a228b16de2cf2ef2467f3e70b3/.context/standards/Architecture-Decisions.md#L1071-L1074

Existing one on `main`: https://github.com/paranext/paranext-core/blob/f0ce0758d6ee8d19b370c1bc39f08f48c856db31/.context/standards/Architecture-Decisions.md#L1077-L1079

2. The PR no longer merges — GitHub reports `mergeable: CONFLICTING` / `DIRTY`. `git merge-tree` shows two files changed on both sides: `Architecture-Decisions.md` (the append above) and `platform-scripture-editor.web-view.tsx`, where PT-4339's editable/readonly work (#2682) landed in the same file.

https://github.com/paranext/paranext-core/blob/2cec4b8c46be34a228b16de2cf2ef2467f3e70b3/extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx#L1953-L1960

One rebase onto current `main` fixes both, renumber included.

**Still open from round 1:** the snapshot premise has never executed in the real app — the Chromium probe was a stand-in page, the new e2e test has never run, and `test:e2e:isolated` is in no CI workflow. I have a Mac and offered last round; I'll run the two manual cases from your description after the rebase so this doesn't hold you up.

jolierabideau (approved):
Only issue is the ADR is misnumbered again

> @jolierabideau reviewed 28 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2692#-:-P-qjUGW5MXkO4qILJzO:bv9ikdl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2693, opened 2026-08-19
PT-4386: Stop the startup sync from locking the project picker

katherinejensen00:
## The bug

In **Simple mode**, the project picker never unlocked at startup — while the rest of the UI looked fully loaded.

Core fired `syncProjects` with no project IDs ("Send/Receive every locally-known shared project") the instant its settings gates passed. In Paratext 10 Studio that command is served by the .NET data provider, and its Mercurial/DBL work saturates that process. Extensions activate **strictly sequentially**, so anything awaiting the busy provider stalls `platform-scripture` behind it — and `platform-scripture` registers the Scripture Extender PDPF, the only factory advertising `platformScripture.USJ_Chapter`, which is exactly what the picker waits on. Self-inflicted starvation.

## The fix

Core keeps ownership of the trigger and gates it on **workspace readiness** instead of firing blind:

```
performStartupTasks (fire-and-forget, Simple branch)
    evaluateSimpleModeSyncGates()        ← settings gates
    waitForScriptureWorkspaceReady()     ← NEW: phase 1 registered, phase 2 answering
    evaluateSimpleModeSyncGates()        ← again, settings can change during a 120 s wait
    sendCommand('…syncProjects', undefined)
```

Outcomes are `'ready' | 'timed-out' | 'aborted'`. **`'timed-out'` means "couldn't confirm", not "don't run"** — the sync is delayed, never suppressed. Power mode is deliberately untouched.

## Review round 2 (commit 2)

Everything below the fold describes the original commit; this section is what changed in response to @jolierabideau's review. **The blocking finding is addressed with both shapes suggested.**

| Finding | Resolution |
| --- | --- |
| **macOS abort reintroduces the hazard the abort prevented** (blocking) | `canStartupSyncFireNow` is re-checked live immediately before the command goes out and skips when the app is resident but windowless — so a wait that outlives a last-window close can only fire if a dock reactivation brought a window back. |
| **…and `startupTasksAbort` is shared with Power mode** | The shared abort is unconditional again, so Power mode's boot-race loop is exactly what it is on `main`, on every platform. The carve-out moved to `startupReadinessAbort` / `StartupTasksSignals.readinessAbortSignal`, which reaches only the Simple readiness wait. "Power mode unchanged" is now literally true. |
| Studio's removed `Task.Run(SyncProjects)` is a load-bearing cross-repo claim | Cited precisely in the ADR: `paranext/paratext-10-studio`, `repo-patches/paranext-core.patch` lines 5334–5336 at commit `f779810`, in the hunk patching `ParatextProjectSendReceiveService.cs`, quote included. |
| Backoff cadence and `shouldLogProbeAttempt` unpinned | One test pins both — answers on probe 15 (10 waits @1 s + 4 @2 s = 18 s), so flattening (14 s) or inverting (22 s) fails, and it asserts exactly two log lines, on attempts 1 and 10. |
| Two comments overstate the code | Both corrected. The probe-before-deadline guarantee is that the request is *issued*, not that it counts; `escapeStringRegexp` buys a literal **substring** match, not an exact one (`includeProjectInterfaces` compiles to unanchored `RegExp`s applied with `.test()`). |
| `getMetadataForAllProjectsWithoutRetries` used off-label | Rationale written down: the retrying variant absorbs startup unreadiness on its own schedule against `LOAD_TIME_GRACE_PERIOD_MS`, which is the very signal this loop exists to read. One retry loop, not two. |
| Duplicated `2000` and the duplicated `platformScripture.USJ_Chapter` literal | Both documented — the interval as the one number here that genuinely can drift, the literal as an intentionally unenforced coupling. |
| Stale `Promise.race` test comment, `satisfies ProjectMetadata`, missing `@param`/`@returns` | Fixed. |
| `waitForNetworkObject` leaks its subscription on timeout | Confirmed the reading (`unsub()` only runs from inside the callback). Pre-existing and out of scope; folding into the Power-mode follow-up ticket. |

**Also found while doing the above, and fixed here:** phase 1 treated *every* rejection as the end of the gate. `waitForNetworkObject` rejects both when its budget expires and when the status service is unreachable, and the second can land with nearly the whole budget unspent — collapsing a 120 s gate into seconds and firing the sync early, i.e. the original bug reintroduced invisibly. A non-exhaustion rejection now falls through to phase 2, which tests the stronger condition and already tolerates throws. From main that path is currently unreachable (the status service is hosted in main, awaited before the startup tasks run, and a same-process `get` returns the local proxy — a synchronous in-memory read), so this is insurance, not a fix for an observed failure; the ADR says so plainly.

**Rebased onto `main`** (`cbecad7f1cb`). The ADR entry is no longer numbered. `main` appended its own `ADR-0027` while this branch carried one, so this rebase landed two identical headings — precisely the collision the log's own guidance warns about and asks the second merger to fix by hand. The entry is now **`ADR-startup-sync-readiness-gate`**, and the log's guidance and entry template prescribe a kebab-case slug claimed at write time; entries written before this keep their `ADR-NNNN` headings. Note the *first* commit's message still says "Recorded in ADR-0013" — stale; worth fixing in the squash message.

## Files, in reading order

| File | ± | What it is |
| --- | --- | --- |
| **`src/main/startup-readiness.util.ts`** | +370 | **Start here.** The whole gate: two phases plus `settleWithin`, which bounds both against one monotonic deadline. |
| `src/main/startup-tasks.ts` | +180/−47 | Wires the gate into the Simple branch; extracts `evaluateSimpleModeSyncGates` so it can run before **and** after the wait. |
| `src/main/services/shutdown-latch.service.ts` | +65 | Two predicates: `shouldWindowCloseAbortReadinessWait(platform)` and `canStartupSyncFireNow(windowCount, hasCreatedWindowThisProcess)` (see focus #2). |
| `src/main/main.ts` | +20/−9 | One call site now uses that predicate instead of an inline check. |
| `.context/standards/Architecture-Decisions.md` | +100/-11 | `ADR-startup-sync-readiness-gate` — why core keeps the trigger, and what was deferred; plus the switch to slug-named entries. |
| `src/renderer/hooks/use-project-picker-data.hook.ts` | +4 | **Comment only.** Cross-reference for a deliberately duplicated literal. |
| `*.test.ts` (3 files) | +877 | 76 tests (20 + 40 + 16). |

## Where to focus (the rest is fairly mechanical)

1. **`settleWithin` race ordering** (`startup-readiness.util.ts:115`) — the highest-risk logic here. Neither underlying call is cancellable or takes an `AbortSignal`, so without it the 120 s budget is decorative. Six abort re-checks exist because **abort must win the *report* even when it loses the race**: `'timed-out'` means *proceed*, so a quit misread as a timeout would fire a whole-workspace S/R into a connection `networkService.shutdown()` is tearing down. Details in the collapsed section below.
2. **The macOS lifecycle carve-out, now on its own signal** (rewritten in response to review — see "Review round 2"). Three pieces that have to be read together: `startupTasksAbort` aborts unconditionally on any shutdown, so **Power mode's boot-race loop is byte-for-byte what it is on `main`**; `startupReadinessAbort` carries the macOS exception and reaches only the Simple readiness wait; and `canStartupSyncFireNow` is re-checked live immediately before the command goes out. The subtle part worth a second pair of eyes: the Simple path must run on the readiness signal *throughout*, including its post-wait abort check, because a last-window close makes `isAppShuttingDown()` true and therefore aborts the shared signal on precisely the close the carve-out exists to survive. Re-checking the shared signal afterwards makes the whole carve-out unreachable — that was the first attempt, and there is now a test that fails if it comes back.
3. **Policy call:** is a 120 s worst case acceptable, and is firing anyway on timeout right (vs. skipping)? Current answer: delayed, never suppressed.

## Testing

Rebased onto `main` (`cbecad7f1cb`). The only conflict was the ADR log (both sides append at the end); the branch's source diff is unchanged apart from the ADR heading's slug and its one citation in `startup-tasks.ts`. The branch's own suites (`startup-readiness.util`, `startup-tasks`, `shutdown-latch.service` — 76 tests) re-run green on the new base.

**`npm test` — 1864 passing.** Five suites time out under CPU contention (`window-layout-persistence`, `theme.service`, and the `theme`/`scroll-group`/`web-view` service-hosts); all 106 of their tests pass in isolation, none import anything this branch touches, and the pattern is the documented multi-window flake rather than a regression. **`dotnet test` — 1580 passing, 0 failed.** **`typecheck:core`, `eslint`, and `prettier --check` clean** on all 9 changed files.

The 76 tests here were mutation-checked rather than just run: reverting the phase-1 fall-through, the windowless guard, the "no window yet" latch, the signal split, the backoff cadence, or the log sampling each fails at least one test. Note `typecheck` needs `npx ts-node ./.erb/scripts/generate-dev-build-info.ts` run once first, or it stops on the gitignored `release/app/buildInfo.json` before reaching this branch's files.

> ⚠️ **This cannot be verified end-to-end in this repo.** Core's `syncProjects` is a stub that throws — Studio patches in the real implementation. **Manual verification in Paratext 10 Studio is required before merge:** Simple mode, `firstRunComplete` + `syncOnStartup` both true, several locally-checked-out shared projects (ideally with DBL resources so the sync is slow) → restart → picker unlocks, and the sync still runs afterward.

<details>
<summary><b>How the gate works, and the <code>settleWithin</code> ordering</b></summary>

One monotonic deadline (`performance.now() + 120 s` — not wall clock, because this runs during OS cold boot, exactly when the wall clock gets stepped). Both phases spend from it.

- **Phase 1 — registered.** `waitForNetworkObject` for the PDPF advertising `platformScripture.USJ_Chapter`.
- **Phase 2 — answering.** Poll `getMetadataForAllProjectsWithoutRetries` until non-empty. Registration alone is not enough: the Scripture Extender is a *layering* factory whose `getAvailableProjects` fans out to the .NET base factories, so it can be registered while still unable to answer. Backoff mirrors the sibling boot-race loop (shared cadence for the first few attempts, then 2 s).

**`settleWithin`'s core trick: only one thing can reject.** The inner promise's success *and* failure are both funnelled into gate resolutions, and abort resolves too — so the `AsyncVariable`'s own timeout is the only path that can reject it. That is what lets the catch return `TIMED_OUT` unconditionally, with no string-matching on a rejection message.

Five ordering facts, in the order they bite:

1. **`promise.catch(() => undefined)` is line 120 — above every early return.** Both early exits walk away from a promise that may reject seconds later; attach the no-op handler after them and you have built an unhandled-rejection generator. Giving up only stops us awaiting — the abandoned call keeps running.
2. **`remainingMs <= 0` must fail fast.** `AsyncVariable` arms its timer only when the value is `> 0`; `-1` is its documented "no timeout" sentinel, but an unguarded `0` hits the same false branch and waits forever. Guarded twice on purpose — `:123`, and again at `:197` where the same value is handed to `waitForNetworkObject` as its own bound.
3. **First settlement wins, and the losers still run.** Every resolve is wrapped in `if (!gate.hasSettled)` — not for correctness (`resolveToValue` is already a no-op once settled) but because it logs "Ignoring subsequent resolution…" every time. Phase 2 calls `settleWithin` twice per iteration for up to 120 s; unguarded, that is a log flood. Same guard `network-object-status.service.ts` puts on its own gate.
4. **The abort listener is torn down in `finally`**, via a second `AbortController` passed as `{ signal: cleanup.signal }` — otherwise hundreds of calls leak listeners onto the caller's long-lived signal. `AsyncVariable` clears its own timer as it settles.
5. **Abort must beat timeout on the report.** The gate settles on whichever signal physically arrives first, so a quit landing in the same tick as an underlying rejection can lose the race. Both catch blocks therefore re-check `abortSignal?.aborted` (`:308`, `:331`), and `startup-tasks.ts:200` re-checks a sixth time — ordered *above* the `'timed-out'` warn, so a sync about to be skipped never logs "syncing anyway" first.

`settleWithin` rethrows `settled.error` rather than swallowing it, so each phase decides what an underlying failure means. Phase 2 treats a throwing probe as "not answering yet" and retries (the fan-out to a still-booting .NET process legitimately fails), but carries the message forward in `attemptDetail`, so a deadline hit right after a throw reports the real cause instead of a generic "no projects".

**Gates are evaluated twice.** The wait can park for 120 s; a user who flips to Power mode or turns off auto-sync while it is parked must not get a no-ID sync anyway. Each call independently re-reads all three settings, preserving the consent asymmetry exactly: unreadable `interfaceMode` → don't sync (Simple's no-ID sync would override a Power user's schedule); unreadable `firstRunComplete` → don't sync; unreadable `syncOnStartup` → **do** sync (a read failure shouldn't silently suppress a sync the user never declined).

</details>

<details>
<summary><b>Decisions, rejected alternatives, and what's knowingly left alone</b></summary>

**Rejected — have the S/R extension self-trigger at the end of its own activation.** The obvious alternative, and the one the ticket named as preferred. Sequential activation gives no guarantee it activates *after* `platform-scripture`; if it goes first you get the identical bug relocated into a repo core cannot test. Studio had also deliberately *removed* its own `Task.Run(SyncProjects)` self-trigger to make core the single trigger owner. Recorded in ADR-startup-sync-readiness-gate.

**Rejected — scoping the phase-2 probe with `includePdpFactoryIds`.** Looks like the fix for "one failing third-party factory blocks the gate", but it breaks the gate outright: `internalGetMetadata` forwards that filter to the factory *and* adds the factory's own id to `excludePdpFactoryIds` (`project-lookup.service-model.ts:444-452`), so a layering factory scoped to itself layers over nothing and reports zero projects forever.

**Rejected — `waitForDuration` for bounding.** It is `Promise.any([timeout, fn()])`, which swallows a fast rejection and turns it into a full-timeout `undefined` — converting 1 s retries into 30 s stalls.

**Knowingly left alone — Power mode** stays ungated against PDP readiness. Its freshness window is anchored to window-interactive time, so charging a 120 s readiness wait against it could silently *drop* a legitimate startup sync — trading a visible bug for an invisible one. ADR-startup-sync-readiness-gate marks it revisit-when-the-anchor-is-reworked.

**Known consequence:** a workspace with no local scripture projects never satisfies the probe, so it waits the full budget and warns on every boot before firing a sync that typically has nothing to do.

**Root cause is documented as plausible, not verified.** The intuitive explanation — the synchronous S/R handler blocking StreamJsonRpc's single message-read loop — was real but is *already fixed* in Studio, which offloads those handlers via `Task.Run` and has a regression test asserting it. The likelier remaining candidate is contention with `LocalParatextProjects.Initialize()`'s `ScrTextCollection.RefreshScrTexts()` scan, but nobody has instrumented it. The gate is deliberately mechanism-agnostic — it holds regardless of which contention dominates, which is the main reason to prefer it over trying to make the two workloads coexist.

</details>

<details>
<summary><b>Pre-PR review log</b> — findings raised and resolved before this PR was opened</summary>

The sections below are the output of `/review-paratext`: findings from four parallel analysis passes, what was fixed, and what was dismissed with reasoning. Useful if you want to know what has already been checked; not required reading to review the diff.

# Code Review Summary

**Branch**: pt-4386-simple-startup-sync-race

**Base**: origin/main

**Date**: 2026-08-18

**Files changed**: 9 in the commit (10 changed on disk; `package-lock.json` is a pre-existing, unrelated modification deliberately left out)

## Overview

In Simple interface mode the project picker never unlocked at startup, while the rest of the UI looked fully loaded. Core fired a whole-workspace Send/Receive as soon as its settings gates passed; in Paratext 10 Studio that command is served by the .NET data provider, and its Mercurial/DBL work starved the scripture project-data-provider factory that the picker waits on. Because extensions activate strictly sequentially, an extension blocked on the busy provider stalls `platform-scripture` behind it — and `platform-scripture` registers the Scripture Extender layering PDPF, the only factory advertising `platformScripture.USJ_Chapter`.

This adds a bounded, two-phase readiness gate in the main process (`startup-readiness.util.ts`): wait for that factory's network object to register, then poll until it actually answers a metadata probe, and only then fire the sync. Registration alone is insufficient because the Extender is a layering factory whose `getAvailableProjects` fans out to the .NET base factories, so it can be registered while still unable to answer. On timeout the sync fires anyway — delayed, never suppressed — and on a real quit it is skipped. Power mode is deliberately untouched.

## API Changes

None. All changed files are main-process or renderer-internal. Nothing touches `lib/platform-bible-react/`, `lib/platform-bible-utils/`, `papi.d.ts`, or any extension's `src/types/*.d.ts`. `papi.d.ts` correctly did not regenerate, because nothing in `src/shared` changed. No public API surface is affected.

None.

- [x] **`main.ts`: narrowing the startup-task abort to `isAppQuitRequested()` fixed macOS but opened a gap on Windows/Linux.** Clicking X on the last window runs the entire shutdown Send/Receive (`main.ts:998`) while the abort is still un-fired, because `app.quit()` — and therefore the quit flag — only happens at `window-all-closed` (`main.ts:1357`), which runs afterward. A startup sync could fire concurrently with the shutdown sync. _(fixed during review: extracted `shouldWindowCloseAbortStartupTasks(platform)` into `shutdown-latch.service.ts`, returning `isAppQuitRequested() || platform !== 'darwin'`; non-darwin has no stay-resident behavior, so the unconditional abort was always correct there)_ — _superseded in round 2: that predicate is now `shouldWindowCloseAbortReadinessWait` and governs only the Simple readiness wait_
- [x] **That `main.ts` lifecycle conditional had no test coverage at all** — no `main.test.ts` exists and no e2e test touches the path, so an accidental inversion would silently regress either the macOS fix or the shutdown guard. _(fixed during review: the extracted predicate is unit-tested in `shutdown-latch.service.test.ts` with two cases pinning both directions — macOS-without-quit must NOT abort, non-macOS-without-quit must)_
- [x] **Phase 1 had no test for the case its deadline race actually exists for.** All three phase-1 "timed-out" tests went through an immediately-rejecting mock or the zero-budget guard; none drove a genuinely hung `waitForNetworkObject` to the deadline. Since the same remaining budget is handed to both `waitForNetworkObject` and `settleWithin`, and `settleWithin` starts its timer marginally later, its deadline almost always fires first — so the untested path was the *typical* production timeout, not an edge case. _(fixed during review: added a hung-registration test mirroring the phase-2 equivalent, asserting the phase-1 reason specifically so it cannot pass by falling through to phase 2)_
- [x] **`settleWithin` hand-rolled `Promise.race` plus `setTimeout` for a timeout, against an explicit house rule.** `Code-Style-Guide.md:85` says "Avoid: raw `Promise.race` for timeouts", and `shared-store.service.ts:271-277` already solves this exact shape by bridging into an `AsyncVariable`, citing that rule. _(fixed during review: reworked onto `AsyncVariable`. Both of the wrapped promise's outcomes are funnelled into resolutions, so the gate's own timeout is the only thing that can reject it — no message-matching needed. All 47 gate tests passed unchanged, which is the evidence it is a faithful swap. Resolutions are guarded on `hasSettled`, matching `network-object-status.service.ts:53,56,64`, to suppress the duplicate-resolution logging the swap introduced)_

### Minor — Consider

- [x] **Post-wait settings re-read was only tested for one of the three gates** (`interfaceMode`); `firstRunComplete` and `syncOnStartup` were not pinned on the post-wait call. _(fixed during review: replaced the single test with three, one per gate, driven by an extracted `startupWithSettingsChangedDuringReadinessWait` helper; assertions kept inline so `vitest/expect-expect` still sees them)_
- [ ] ~~The two-phase backoff loop now exists twice — `pollUntilScriptureMetadataAnswers` and the pre-existing `requestSessionSyncWithBootRetry`.~~ _(Author asked for deduplication only if it made sense; on inspection it does not. The loops share only the cadence ternary — one returns a result object, routes both awaits through `settleWithin`, re-checks abort at four points and accumulates a detail; the other throws on non-retryable errors and on deadline, has a freshness gate, and uses a bare `wait()`. A shared `backoffIntervalMs(attempt, initialAttempts, initialMs, extendedMs)` would replace a self-explanatory ternary with a four-argument call whose two adjacent number parameters can be silently transposed — worse to read and a new failure mode, to remove one line. The constants underneath are already shared: both alias `MAX_REQUEST_ATTEMPTS`/`REQUEST_ATTEMPT_WAIT_TIME_MS` from `rpc.model`.)_
- [x] **`platform.interfaceMode` is read twice on every Simple-mode startup** — once to branch power/simple, then again inside `evaluateSimpleModeSyncGates()`. _(Author asked for a recommendation; the recommendation was to keep it, because the two reads answer different questions: the first asks "which mode, and can we tell at all?" — an unreadable mode there skips the automatic sync for BOTH modes — while the second asks "is it STILL simple?", which must be fresh, since that is the entire point of the post-wait call. Sharing one read would require passing a pre-read value in and branching, making the pre- and post-wait calls asymmetric; that symmetry is what makes the post-wait check obviously correct. fixed during review: added a comment at the helper recording this, so it is not later "optimized" away)_

#### Shared Regions Modified

#### Extension Config Changes

## Positive Observations

- The gate reuses proven idioms rather than inventing mechanisms: the same `waitForNetworkObject({ objectType, attributes: { projectInterfaces } })` shape `getMetadataForProject` already uses, `escapeStringRegexp` for the regex-matched `includeProjectInterfaces` field, and the `MAX_REQUEST_ATTEMPTS`/`REQUEST_ATTEMPT_WAIT_TIME_MS` constants aliased rather than re-declared.
- Comments were spot-checked against the code they cite — the lookup service's fan-out-then-filter ordering, its layering-filter behavior, `areProjectInterfacesIncluded`'s unanchored `.test()` matching, `activateExtensions`' sequential loop, and `main.ts`'s darwin `window-all-closed` behavior — and were accurate.
- `evaluateSimpleModeSyncGates` is a clean extraction that preserves the previous per-setting semantics exactly while enabling the before/after-wait re-evaluation without duplicating the three checks.
- Tests are behavior-focused and genuinely falsifiable: same-tick abort-vs-rejection ordering in both phases, the "at least one probe even if registration consumed the budget" guarantee, pinned backoff cadence, and `vi.waitFor` rather than fixed microtask flushes (which would have made the "does not fire until readiness resolves" assertion vacuously true).
- `settleWithin` correctly avoids leaking listeners on the long-lived shared `abortSignal` by registering with `{ signal: cleanup.signal }` and aborting the cleanup controller in `finally` — important because phase 2 calls it repeatedly against the same signal.
- The fix is applied at the right layer. The picker's existing spinner/empty-state (`project-picker.component.tsx:225-233`) and its own PDPF-registration refresh are left untouched; the change lets that existing affordance succeed rather than papering over a contention bug with new UI.
- ADR-startup-sync-readiness-gate records the decision with concrete rejected alternatives and an honest statement of what is still unresolved.

## Interview Notes

**Stated purpose (author-approved wording):** Fix the Simple-mode project picker never unlocking at startup (PT-4386), by making core wait — bounded — for the scripture project data provider factories to be registered and answering before firing the startup Send/Receive.

**Decisions the author made during this work:** the strength of the readiness gate (registered *and* answering, rather than registration alone); leaving Power mode knowingly ungated; the 120 s budget; folding the feature and review fixes into a single commit; excluding `package-lock.json` from the commit; and requiring that the ADR entry be justified rather than written by default — the author pushed back with "if we are adding something to the ADR, please make really sure it is the right call", which narrowed ADR-startup-sync-readiness-gate to the cross-repo and deferral decisions and moved mechanism detail into TSDoc.

**Author response to findings:** approved both `main.ts` fixes; approved the phase-1 test as "meaningful and not more work than it's worth"; approved the `AsyncVariable` rework; asked for the post-wait gate tests; delegated the backoff-deduplication judgment ("if it makes sense to") and asked for a recommendation on the double settings read. Nothing further flagged at wrap-up.

**Author's design walkthrough.** The author gave an unprompted, detailed walkthrough of the change in their own words, and it was verified against the code — the mechanisms, the rationale, and the cited line numbers all check out. It covered:

- The starvation chain end to end: no-ID `syncProjects` → saturated .NET provider → sequential extension activation stalling `platform-scripture` → the Scripture Extender PDPF (the only factory advertising `platformScripture.USJ_Chapter`) never registering → picker times out while the rest of the UI looks fine.
- Why the obvious alternative was rejected: sequential activation gives no guarantee `paratextBibleSendReceive` activates after `platform-scripture`, so a self-trigger relocates the identical bug into a repo core cannot test — and Studio had already removed its own self-trigger to make core the single owner.
- Why the deadline is monotonic (`performance.now()`): this runs during OS cold boot, precisely when the wall clock gets stepped.
- Why registration alone is insufficient: the Extender is a layering factory whose `getAvailableProjects` fans out to the .NET base factories, so it can be registered while still unable to answer — hence the second phase.
- **The five ordering facts in `settleWithin`** (`startup-readiness.util.ts:115`), which is the subtlest part of the change: (1) the no-op `promise.catch` at :120 must precede *every* early return, or the early exits become an unhandled-rejection generator; (2) `remainingMs <= 0` must fail fast because `AsyncVariable` only arms its timer for `> 0`, so an unguarded `0` waits forever — guarded twice, at :123 and again at :197 where the same value is handed to `waitForNetworkObject` as its own bound; (3) `hasSettled` guards exist for log noise, not correctness, because phase 2 calls `settleWithin` twice per iteration for up to 120 s; (4) the abort listener is torn down in `finally` via a second controller so hundreds of calls do not leak listeners onto the caller's long-lived signal; (5) **abort must beat timeout on the *report* even when it lost the race** — hence the re-checks at :308 and :331, and a sixth in `startup-tasks.ts:200` ordered above the `'timed-out'` warn. The author correctly identified why this is not cosmetic: `'timed-out'` means *proceed*, so a quit misread as a timeout fires a whole-workspace S/R into a connection `networkService.shutdown()` is tearing down.
- The funnel as the load-bearing trick: both of the wrapped promise's outcomes resolve the gate, so the `AsyncVariable`'s own timeout is the only path that can reject — which is what lets the catch return `TIMED_OUT` with no string matching on the rejection message.
- Why `settleWithin` rethrows rather than swallowing, so each phase decides what an underlying failure means, with `attemptDetail` carrying the real cause forward so a deadline hit right after a throw does not report the generic "no projects".
- The consent asymmetry preserved across both gate evaluations (unreadable `interfaceMode` → don't sync; unreadable `firstRunComplete` → don't sync; unreadable `syncOnStartup` → do sync).
- The `shouldWindowCloseAbortStartupTasks` wrinkle in both directions (renamed `shouldWindowCloseAbortReadinessWait` in round 2, and narrowed to the Simple readiness wait): on macOS the quit flag decides because the app stays resident and startup tasks run once per process; off macOS the platform decides because `window-all-closed` calls `app.quit()` only *after* the close handlers have run, so the flag arrives too late to protect the very window the guard covers.

**The honest caveat that remains:** the code was AI-generated, and several of the defects fixed during this review were introduced by that generation — including the Windows/Linux shutdown-race regression created by the first attempt at the macOS fix. What the walkthrough demonstrates is that the author has command of the design and its failure modes, not that the code was hand-written. Treat the review as a check on the implementation, not on whether the author understands it.

**Context a reviewer needs, which is not visible from the diff:**

- **This fix cannot be verified end-to-end in this repo.** Core's `syncProjects` is a stub that throws `PlatformUnimplementedException`; the real implementation is patched in by Paratext 10 Studio. No core unit test or e2e test can exercise the actual resource contention. Manual verification in Studio is required: Simple mode, `firstRunComplete` and `syncOnStartup` both true, several locally-checked-out shared projects (ideally with DBL resources so the sync is slow), restart, and confirm the picker unlocks and the sync still runs afterward.
- **The root-cause mechanism is recorded as plausible, not verified.** The intuitive explanation — the synchronous S/R handler blocking StreamJsonRpc's single message-read loop — was real but is already fixed in Studio, which offloads those handlers via `Task.Run` and has a regression test asserting it. The likelier remaining candidate is contention with `LocalParatextProjects.Initialize()`'s `ScrTextCollection.RefreshScrTexts()` scan, but nobody has instrumented it. The gate is deliberately mechanism-agnostic: it holds regardless of which contention dominates, which is the main reason to prefer it over trying to make the two workloads coexist.
- **Power mode keeps the same latent race, knowingly.** Applying the gate there would charge the wait against its window-interactive freshness window and could silently *drop* a legitimate startup sync — trading a visible bug for an invisible one. Recorded in ADR-startup-sync-readiness-gate; a follow-up ticket has not yet been filed.
- **Two review suggestions were deliberately NOT implemented, because they would have caused harm.** Scoping the phase-2 probe with `includePdpFactoryIds` would break the gate outright — the lookup service forwards that filter to the factory *and* adds the factory's own id to `excludePdpFactoryIds` (`project-lookup.service-model.ts:444-452`), so a layering factory scoped to itself layers over nothing and reports zero projects forever. And `waitForDuration` was rejected for bounding because it is `Promise.any([timeout, fn()])`, which swallows a fast rejection and turns it into a full-timeout `undefined`, converting 1 s retries into 30 s stalls.
- **A known consequence:** a workspace with no local scripture projects never satisfies the probe, so it waits the full budget and then warns on every boot before firing a sync that typically has nothing to do.

## In-Review Quality Check

Changes were made during the interview, so the full check was run afterward. **No fixes were required; nothing new needed staging.**

- **typecheck** — fails only with a known pre-existing, environmental error: `src/main/services/app.service-host.ts` cannot resolve `release/app/buildInfo.json`, a gitignored build artifact never generated in this worktree. Not in the changed-file set.
- **lint** — full repo-wide run completed, workspace lint exit code 0. The only error anywhere was the same pre-existing `buildInfo.json` unresolved-module error. All other output was pre-existing warnings in files this branch never touched. **None of the 9 changed files produced any lint output.**
- **format** — scoped prettier over the changed files reported all of them unchanged; `Architecture-Decisions.md` confirmed clean via `--check`. `package-lock.json` untouched.
- **test** — `src/main/` is 333 passing across 18 files. Full `test:core` matched the pre-existing baseline at 1804/1805; the single failure (`web-view.service-host.test.ts`, `Test timed out in 5000ms`) passed cleanly in isolation and is the documented flaky multi-window pattern, not a regression. A broader all-workspaces run was abandoned after severe machine-level CPU contention made even trivial shell commands stall; no failure in it was in a changed file or its dependents.

## Suggested Review Focus

Prioritized agenda for the author–reviewer meeting:

- [ ] **Scrutinize `settleWithin`'s race ordering** (`startup-readiness.util.ts:115`). It is the load-bearing primitive — it makes the 120 s budget a real ceiling over two uncancellable calls that take no `AbortSignal`. The author's walkthrough (see Interview Notes) already lays out the five ordering facts and they check out against the code, so this is not a knowledge gap to close; it is the highest-risk logic in the diff and deserves a second pair of eyes on the ordering itself — particularly the six abort re-checks and whether any path can still report `'timed-out'` (which means *proceed*) for what was actually a quit.
- [x] **The `main.ts` lifecycle change** — reworked in review round 2. The shared abort is unconditional again (Power mode unchanged), the macOS exception moved to `startupReadinessAbort`, and `canStartupSyncFireNow` stops a surviving wait firing into a windowless app. Still worth confirming: that the three-part interaction reads correctly, and that "no window yet" vs "went windowless" is the right distinction.
- [ ] **Is a 120 s worst-case delay to the startup sync acceptable?** And is firing anyway on timeout the right call versus skipping? The current answer is "delayed, never suppressed."
- [ ] **The empty-workspace cost.** A workspace with no local scripture projects burns the full budget and warns every boot. Is that acceptable, or should "no factories at all" be distinguished from "factories reporting nothing"?
- [ ] **Power mode is knowingly left with the same latent race.** Confirm the deferral, and agree who files the follow-up ticket. (Round 2 made "Power mode unchanged" literally true — the shared abort no longer branches on platform — but the *readiness* deferral still stands and still has no ticket.)
- [ ] **The empty-workspace probe cost, revisited.** Raised in review and deliberately not changed: capping probes, or treating registered-but-empty as ready after N attempts, would reintroduce the original race for exactly the slow workspaces that suffer it worst. The version worth considering later is distinguishing *answered-with-empty* from *did not answer* — a successful `[]` is an answer — but that changes the gate's core predicate and would need the Studio verification redone.
- [ ] **Manual verification in Paratext 10 Studio is mandatory before merge** — no automated test in this repo can exercise the real contention.
- [ ] **The root cause is documented as unverified.** Decide whether that is acceptable to merge on, given the fix is deliberately designed to hold regardless of which contention dominates.

</details>

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2693)

### DIFF
--- a/.context/standards/Architecture-Decisions.md
+++ b/.context/standards/Architecture-Decisions.md
@@ -2359,14 +2359,31 @@ step, no automation. Just a record.
   `paratextBibleSendReceive` activates *after* `platform-scripture`, so if it goes first its
   self-trigger starves the same factory — the identical bug, relocated into a repo core cannot test.
   Paratext 10 Studio also *removed* its own `Task.Run(SyncProjects)` self-trigger precisely so core
-  would be the single trigger owner ("that would double-sync", per the `paratext-10-studio` repo's
-  `repo-patches/paranext-core.patch`); self-triggering would undo that. **Defer until an existing
+  would be the single trigger owner; self-triggering would undo that. The evidence is in the
+  `paranext/paratext-10-studio` repo at `repo-patches/paranext-core.patch` (lines 5334–5336 as of
+  commit `f779810`), in the hunk that patches
+  `c-sharp/Projects/SendReceive/ParatextProjectSendReceiveService.cs`: *"NOTE: startup auto-sync is
+  triggered by paranext-core itself (src/main/startup-tasks.ts, "Conditional autosync on startup"),
+  which calls the syncProjects papi command — so the patch no longer kicks off its own
+  Task.Run(SyncProjects) here (that would double-sync)."* **Defer until an existing
   "workspace ready" signal** — rejected: none exists. `all-extensions-activated` is a performance
   mark emitted only under `PT_STARTUP_MARKS`, at the end of `activateExtensions` in
   `src/extension-host/services/extension.service.ts`, not a network event.
   **Gate on factory registration alone** — rejected as insufficient: the Scripture Extender is a
   layering factory whose `getAvailableProjects` fans out to the .NET base factories, so it can be
   registered while still unable to answer; hence the second, metadata-probe phase.
+  **Treat any phase-1 rejection as the end of the gate** — rejected: `waitForNetworkObject` rejects
+  both when the budget it was given expires (genuine exhaustion) and when the status service is
+  unreachable, and the latter can land with nearly the whole budget unspent. Ending there would
+  collapse a 120 s gate into seconds and fire the sync early — the original bug, reintroduced
+  invisibly, since the cause is logged only at debug while the caller warns "syncing anyway". A
+  non-exhaustion rejection therefore falls through to phase 2, which tests the stronger condition
+  and already tolerates throws; with no factory registered it sees empty results and polls to the
+  deadline, so the worst case degrades to the same `'timed-out'`. This is insurance, not a fix for
+  an observed failure: from main the path is currently unreachable (the status service is hosted in
+  main, `startNetworkObjectStatusService()` is awaited long before the startup tasks run, and a
+  same-process `get` returns the local proxy, making the call a synchronous in-memory read). It is
+  written this way so the guarantee does not silently depend on that arrangement holding.
 - **Consequences:** Core is the single owner of startup-sync ordering, and the gate is
   mechanism-agnostic — it holds regardless of which contention inside the .NET provider is
   responsible, which matters because that contention is not fully pinned down (the obvious
@@ -2380,5 +2397,24 @@ step, no automation. Just a record.
   just before — a setting changed mid-wait now re-gates the sync rather than firing on stale
   settings. On a real quit (`isAppQuitRequested()`) the wait is aborted and the sync is skipped; a
   macOS last-window close that leaves the app resident deliberately does NOT abort it, since the app
-  is still going to want its startup sync.
+  is still going to want its startup sync. That exception is carried on its own abort signal
+  (`startupReadinessAbort` in `main.ts`, `StartupTasksSignals.readinessAbortSignal`) rather than on
+  the shared one, so it applies to the Simple readiness wait ALONE — Power mode's boot-race retry
+  loop keeps aborting on every shutdown, on every platform, exactly as before. The Simple path then
+  has to run on that signal *throughout*, including its post-wait abort check: a last-window close
+  makes `isAppShuttingDown()` true via `areAllWindowsClosing()`, so the shared signal aborts on
+  precisely the close the carve-out exists to survive, and re-checking it after the wait would make
+  the carve-out unreachable — the wait would outlive the close only to be discarded as "quitting".
+  Simple mode's hard stop is therefore the live predicate rather than a latched signal, which is
+  strictly better anyway: it is evaluated at fire time and covers both ways the app can be
+  un-syncable (shutting down, and resident but windowless). A surviving wait is
+  still not licence to fire: `canStartupSyncFireNow` is re-checked immediately before the command
+  goes out and skips when the app is resident but windowless, which is the state a macOS last-window
+  close leaves behind. It distinguishes that from having no window YET — the app still coming up —
+  which stays allowed, since treating it as disqualifying would silently drop the sync on any boot
+  whose readiness landed before the first window did. Without the check a wait parked for the full
+  budget could resolve minutes later and run a whole-workspace Send/Receive after the shutdown sync
+  already ran, with no UI to report progress or failure into. The exception therefore pays off in
+  exactly one case — a dock reactivation brings a window back before readiness lands — which is the
+  case it was added for.
 - **Source:** PT-4386.
--- a/.context/standards/Architecture-Decisions.md
+++ b/.context/standards/Architecture-Decisions.md
@@ -2333,7 +2333,7 @@ step, no automation. Just a record.
 - **Source:** PT-2626 review (PR #2727), finding 7 — "the current standard tells the next author to
   do the opposite of what this PR establishes."
 
-## ADR-0027: Core owns startup-sync ordering and gates it on project-data-provider readiness
+## adr-startup-sync-readiness-gate: Core owns startup-sync ordering and gates it on project-data-provider readiness
 
 - **Date:** 2026-08-16
 - **Status:** Accepted
--- a/.context/standards/Architecture-Decisions.md
+++ b/.context/standards/Architecture-Decisions.md
@@ -2332,3 +2332,53 @@ step, no automation. Just a record.
   writing it is deferred rather than done.
 - **Source:** PT-2626 review (PR #2727), finding 7 — "the current standard tells the next author to
   do the opposite of what this PR establishes."
+
+## ADR-0027: Core owns startup-sync ordering and gates it on project-data-provider readiness
+
+- **Date:** 2026-08-16
+- **Status:** Accepted
+- **Context:** In Simple mode, `performStartupTasks` fired `paratextBibleSendReceive.syncProjects`
+  ("sync every locally-known shared project") as soon as its settings gates passed
+  (`src/main/startup-tasks.ts`). In Paratext 10 Studio that command is served by the .NET data
+  provider; its Mercurial and DBL work saturates that process before the scripture PDP factories
+  finish initializing. Because extensions activate strictly sequentially (the awaited loop inside
+  `activateExtensions` in `src/extension-host/services/extension.service.ts`), any extension awaiting
+  the busy provider stalls `platform-scripture` behind it — and `platform-scripture` registers the Scripture
+  Extender layering PDPF, the only factory advertising `platformScripture.USJ_Chapter`, which the
+  project picker waits on. The picker's wait timed out and it stayed locked. The Power-mode path was
+  already gated, but on a different signal — the S/R command's registration, plus a freshness window
+  — never on project-data-provider readiness; the Simple-mode branch had no ordering gate of any kind.
+- **Decision:** Core keeps ownership of the startup-sync trigger and gates it on workspace
+  readiness, in the main process. `src/main/startup-readiness.util.ts` waits — bounded by
+  `STARTUP_SYNC_READINESS_BUDGET_MS` (120 s) — first for the scripture PDP factory network object to
+  register, then for a scripture-metadata probe to answer non-empty, and only then fires the sync.
+  On timeout it fires anyway (delayed, never suppressed); on abort it skips.
+- **Alternatives:** **Have the Send/Receive extension self-trigger at the end of its own
+  activation** (the shape PT-4386 named as preferred, and the one sketched in `startup-tasks.ts`'s
+  own TSDoc) — rejected: sequential activation gives no guarantee that
+  `paratextBibleSendReceive` activates *after* `platform-scripture`, so if it goes first its
+  self-trigger starves the same factory — the identical bug, relocated into a repo core cannot test.
+  Paratext 10 Studio also *removed* its own `Task.Run(SyncProjects)` self-trigger precisely so core
+  would be the single trigger owner ("that would double-sync", per the `paratext-10-studio` repo's
+  `repo-patches/paranext-core.patch`); self-triggering would undo that. **Defer until an existing
+  "workspace ready" signal** — rejected: none exists. `all-extensions-activated` is a performance
+  mark emitted only under `PT_STARTUP_MARKS`, at the end of `activateExtensions` in
+  `src/extension-host/services/extension.service.ts`, not a network event.
+  **Gate on factory registration alone** — rejected as insufficient: the Scripture Extender is a
+  layering factory whose `getAvailableProjects` fans out to the .NET base factories, so it can be
+  registered while still unable to answer; hence the second, metadata-probe phase.
+- **Consequences:** Core is the single owner of startup-sync ordering, and the gate is
+  mechanism-agnostic — it holds regardless of which contention inside the .NET provider is
+  responsible, which matters because that contention is not fully pinned down (the obvious
+  read-loop-blocking explanation was real but was already fixed in Studio, verified 2026-08-16). The
+  startup sync can now be delayed by up to 120 s on a pathological boot. **Power mode is knowingly
+  left ungated** and retains the same latent race: its freshness window is anchored to
+  window-interactive time, so charging a readiness wait against it could silently *drop* a legitimate
+  startup sync — trading a visible bug for an invisible one. Revisit when the freshness anchor is
+  reworked. Because the wait can park for up to 120 s, the three Simple-mode settings gates
+  (interface mode, first-run, sync-on-startup) are re-read and re-evaluated immediately after it, not
+  just before — a setting changed mid-wait now re-gates the sync rather than firing on stale
+  settings. On a real quit (`isAppQuitRequested()`) the wait is aborted and the sync is skipped; a
+  macOS last-window close that leaves the app resident deliberately does NOT abort it, since the app
+  is still going to want its startup sync.
+- **Source:** PT-4386.
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -72,6 +72,7 @@ import {
   markQuitRequested,
   resetShutdownLatchesForNewSession,
   runShutdownTasksOnce,
+  shouldWindowCloseAbortStartupTasks,
 } from '@main/services/shutdown-latch.service';
 import { setAppShutdownSignal } from '@main/services/rpc-server';
 import { startWebViewServiceRouter } from '@main/services/web-view.service-router';
@@ -403,10 +404,11 @@ async function main() {
   // TODO (maybe): Wait for signal from the extension host process that it is ready (except 'getWebView')
   // We could then wait for the renderer to be ready and signal the extension host
 
-  // Signals for the fire-and-forget startup tasks: an abort controller so the Power-mode boot-race
-  // retry loop stops the moment the app begins quitting (wired below), and a window-interactive
-  // clock so a startup sync that only registers late isn't fired onto an editor the user is already
-  // using (see performStartupTasks / STARTUP_SYNC_FRESHNESS_WINDOW_MS).
+  // Signals for the fire-and-forget startup tasks: an abort controller so either mode's startup-sync
+  // wait — Power's boot-race retry loop or Simple's readiness gate — stops the moment a real quit
+  // begins (wired below), and a window-interactive clock so a Power-mode startup sync that only
+  // registers late isn't fired onto an editor the user is already using (see performStartupTasks /
+  // STARTUP_SYNC_FRESHNESS_WINDOW_MS).
   const startupTasksAbort = new AbortController();
   let mainWindowInteractiveAt: number | undefined;
 
@@ -1010,10 +1012,12 @@ async function main() {
         isAppGoingDown = isAppShuttingDown();
 
         if (isAppGoingDown) {
-          // The app is on its way down: stop the startup boot-race retry loop so it can't fire a
-          // startup sync after this shutdown sync, or reach a network connection that is about to
-          // be torn down.
-          startupTasksAbort.abort();
+          // Stop the startup tasks' bounded waits so they can't fire a startup sync after this
+          // shutdown sync, or reach a network connection that is about to be torn down. The one
+          // exception is a macOS last-window close, where the app stays resident and a later dock
+          // reactivation still wants its startup sync — the predicate explains why the platform,
+          // and not the quit flag alone, has to decide that.
+          if (shouldWindowCloseAbortStartupTasks(process.platform)) startupTasksAbort.abort();
 
           // Flush the window-layouts structure with every still-tracked window in it, before any
           // `closed` handler trims the list — a window that is not live at save time is not
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -67,12 +67,13 @@ import {
   startWindowServiceRouter,
 } from '@main/services/window.service-router';
 import {
+  canStartupSyncFireNow,
   isAppQuitRequested,
   isAppShuttingDown,
   markQuitRequested,
   resetShutdownLatchesForNewSession,
   runShutdownTasksOnce,
-  shouldWindowCloseAbortStartupTasks,
+  shouldWindowCloseAbortReadinessWait,
 } from '@main/services/shutdown-latch.service';
 import { setAppShutdownSignal } from '@main/services/rpc-server';
 import { startWebViewServiceRouter } from '@main/services/web-view.service-router';
@@ -404,12 +405,20 @@ async function main() {
   // TODO (maybe): Wait for signal from the extension host process that it is ready (except 'getWebView')
   // We could then wait for the renderer to be ready and signal the extension host
 
-  // Signals for the fire-and-forget startup tasks: an abort controller so either mode's startup-sync
-  // wait — Power's boot-race retry loop or Simple's readiness gate — stops the moment a real quit
-  // begins (wired below), and a window-interactive clock so a Power-mode startup sync that only
-  // registers late isn't fired onto an editor the user is already using (see performStartupTasks /
-  // STARTUP_SYNC_FRESHNESS_WINDOW_MS).
+  // Signals for the fire-and-forget startup tasks:
+  //
+  // - `startupTasksAbort` stops the startup tasks the moment the app starts going down by either
+  //   route, quit or last-window close. This is the long-standing behavior and is what Power mode's
+  //   boot-race retry loop runs on; nothing about the Simple-mode readiness gate changes it.
+  // - `startupReadinessAbort` stops only Simple mode's readiness wait, and is deliberately NOT
+  //   aborted by a macOS last-window close: there the app stays resident, the startup tasks run once
+  //   per process, and a dock reactivation still wants that session's startup sync. Kept separate so
+  //   that exception cannot leak into Power mode's loop.
+  // - a window-interactive clock, so a Power-mode startup sync that only registers late isn't fired
+  //   onto an editor the user is already using (see performStartupTasks /
+  //   STARTUP_SYNC_FRESHNESS_WINDOW_MS).
   const startupTasksAbort = new AbortController();
+  const startupReadinessAbort = new AbortController();
   let mainWindowInteractiveAt: number | undefined;
 
   /**
@@ -430,6 +439,13 @@ async function main() {
     try {
       await performStartupTasks({
         abortSignal: startupTasksAbort.signal,
+        readinessAbortSignal: startupReadinessAbort.signal,
+        // Answers the state neither abort signal can describe: resident, not quitting, no windows —
+        // where macOS leaves the app after its last window closes. `getWindows` already filters out
+        // destroyed windows, so a window mid-teardown does not keep this true; the latch is what
+        // keeps "no window YET" (the app still coming up) from reading as "went windowless".
+        canFireStartupSync: () =>
+          canStartupSyncFireNow(getWindows().length, hasCreatedWindowThisProcess),
         getWindowInteractiveElapsedMs: () =>
           mainWindowInteractiveAt === undefined
             ? undefined
@@ -1012,12 +1028,19 @@ async function main() {
         isAppGoingDown = isAppShuttingDown();
 
         if (isAppGoingDown) {
-          // Stop the startup tasks' bounded waits so they can't fire a startup sync after this
-          // shutdown sync, or reach a network connection that is about to be torn down. The one
-          // exception is a macOS last-window close, where the app stays resident and a later dock
-          // reactivation still wants its startup sync — the predicate explains why the platform,
-          // and not the quit flag alone, has to decide that.
-          if (shouldWindowCloseAbortStartupTasks(process.platform)) startupTasksAbort.abort();
+          // The app is on its way down: stop the startup tasks so they can't fire a startup sync
+          // after this shutdown sync, or reach a network connection that is about to be torn down.
+          // Unconditional, as it has always been — Power mode's boot-race retry loop in particular
+          // must stop here on every platform.
+          startupTasksAbort.abort();
+
+          // Simple mode's readiness wait is the one thing that may outlive this close: on macOS the
+          // app stays resident and the startup tasks run once per PROCESS, so aborting here would
+          // drop the startup sync for good even though a dock reactivation still wants it. The
+          // predicate explains why the platform, and not the quit flag alone, has to decide. A wait
+          // that does survive still cannot fire into a windowless app — `canFireStartupSync` above
+          // is re-checked immediately before the sync goes out.
+          if (shouldWindowCloseAbortReadinessWait(process.platform)) startupReadinessAbort.abort();
 
           // Flush the window-layouts structure with every still-tracked window in it, before any
           // `closed` handler trims the list — a window that is not live at save time is not
@@ -1429,6 +1452,10 @@ async function main() {
       // Stop the startup boot-race retry loop before networkService.shutdown() tears down the
       // connection, so a late retry can't resurrect it (a no-op if the window close already aborted).
       startupTasksAbort.abort();
+      // A real quit ends the Simple-mode readiness wait too, including the macOS last-window-close
+      // case the window handler deliberately let through — that exception exists for an app that
+      // stays resident, and this one is not.
+      startupReadinessAbort.abort();
 
       // Prevent closing before graceful shutdown is complete.
       // Also, in the future, this should allow a "are you sure?" dialog to display.
--- a/src/main/services/__tests__/shutdown-latch.service.test.ts
+++ b/src/main/services/__tests__/shutdown-latch.service.test.ts
@@ -6,6 +6,7 @@ import {
   markQuitRequested,
   resetShutdownLatchesForNewSession,
   runShutdownTasksOnce,
+  shouldWindowCloseAbortStartupTasks,
 } from '@main/services/shutdown-latch.service';
 import {
   addWindow,
@@ -39,6 +40,27 @@ describe('shutdown latches', () => {
     markQuitRequested();
 
     expect(isAppQuitRequested()).toBe(true);
+  });
+
+  test('leaves the startup tasks running when macOS closes its last window without quitting', () => {
+    // The app stays resident and a dock reactivation starts a fresh session, but the startup tasks
+    // run once per process — so aborting here would drop that session's startup sync permanently.
+    expect(shouldWindowCloseAbortStartupTasks('darwin')).toBe(false);
+
+    markQuitRequested();
+
+    // A real quit on macOS still has to stop them.
+    expect(shouldWindowCloseAbortStartupTasks('darwin')).toBe(true);
+  });
+
+  test('stops the startup tasks off macOS even before a quit is flagged', () => {
+    // Off macOS the last window closing always ends in a quit, but `window-all-closed` calls
+    // `app.quit()` only after the close handlers have run — so the quit flag is still false while
+    // the shutdown tasks this guard protects are already running. The platform has to decide.
+    expect(isAppQuitRequested()).toBe(false);
+
+    expect(shouldWindowCloseAbortStartupTasks('win32')).toBe(true);
+    expect(shouldWindowCloseAbortStartupTasks('linux')).toBe(true);
   });
 
   test('shares one run across every window closing as part of the same quit', async () => {
--- a/src/main/services/__tests__/shutdown-latch.service.test.ts
+++ b/src/main/services/__tests__/shutdown-latch.service.test.ts
@@ -1,12 +1,13 @@
 import { beforeEach, describe, expect, test, vi } from 'vitest';
 import type { BrowserWindow } from 'electron';
 import {
+  canStartupSyncFireNow,
   isAppQuitRequested,
   isAppShuttingDown,
   markQuitRequested,
   resetShutdownLatchesForNewSession,
   runShutdownTasksOnce,
-  shouldWindowCloseAbortStartupTasks,
+  shouldWindowCloseAbortReadinessWait,
 } from '@main/services/shutdown-latch.service';
 import {
   addWindow,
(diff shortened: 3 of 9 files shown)


jolierabideau (commented):
## Review summary

Nice work — this is careful, well-documented, and the test coverage is the standout. It lands on
every bullet of PT-4386's DoD, and I like that the rejected alternatives ended up in ADR-0014 rather
than as a stale "the cleaner long-term shape would be…" comment. One thing I'd like to talk through
before merge, then a pile of nits.

Verified locally: 60 tests pass across the three affected test files, `npm run typecheck` clean,
`eslint` clean on the changed sources. I did **not** run the ticket's repro in a real app (Simple
mode, several shared projects with DBL resources) — worth doing before merge, since that's the only
thing that actually proves the picker unlocks.

### Against the ticket

The ticket named **Option A** (S/R extension self-triggers at the end of its own activation) as
preferred, and you chose **Option B** with a written rationale — sequential activation gives no
guarantee that `paratextBibleSendReceive` activates *after* `platform-scripture`, so a self-trigger
that goes first starves the same factory in a repo core can't test. That's the right call and the
right process. One ask: the ADR also leans on `paratext-10-studio`'s
`repo-patches/paranext-core.patch` having *removed* its own `Task.Run(SyncProjects)` so core stays
the single trigger owner. That claim is load-bearing and lives in another repo — can you point me at
it so we can confirm it in review?

### 1. The macOS abort change is a bigger deal than the diff suggests (important)

`shouldWindowCloseAbortStartupTasks` reintroduces, on macOS, the hazard the abort existed to prevent.
On a macOS last-window close with no quit, `isAppShuttingDown()` is still true via
`areAllWindowsClosing()`, so `main.ts` takes the `isAppGoingDown` branch and **does**
`await runShutdownTasksOnce(performShutdownTasks)` — but no longer aborts `startupTasksAbort`. A
readiness wait parked for up to 120 s can then resolve and fire a whole-workspace
`syncProjects(undefined)` *after* the shutdown sync already ran, into a windowless process, with no
user feedback. ADR-0014 cites "that would double-sync" as a reason to reject self-triggering, but
neither the new TSDoc nor the ADR addresses this case — both argue only for preserving the sync
across a dock reactivation.

I don't think it's an oversight: the abort only matters while the gate is still parked, which is
exactly the window where the hazard lives, so it's a genuine trade-off. But right now it's decided
silently in favor of firing. Two shapes that would resolve it: re-check window/shutdown state
immediately before firing (skip when the app is windowless), or hand the pending sync to the next
session via the `resetShutdownLatchesForNewSession` path.

Second half of the same finding: `startupTasksAbort` is **shared with Power mode**. The Power
boot-race retry loop is also no longer aborted on a macOS non-quit last-window close — which
contradicts "Power mode is deliberately untouched" in the PR body and "Power mode unchanged" in the
DoD, and no Power-path test covers it. If it's intentional, it should be stated as a Power-mode
behavior change; if not, it needs scoping to the Simple gate.

### Minor

- **Empty-workspace polling cost.** Phase 2 treats empty as not-ready, so a workspace reporting no
  scripture projects polls the full 120 s: ~10 probes at 1 s then ~55 at 2 s ≈ 65 probes, each an
  unscoped `getAvailableProjects` fan-out across every registered factory — including the .NET
  provider this gate exists to protect. The cost is documented; consider capping the probe count, or
  treating registered-but-empty as ready after N attempts.
- **`getMetadataForAllProjectsWithoutRetries` is used off-label.** Its TSDoc
  (`project-lookup.service-model.ts:100-105`) says it's "intended for use by Layering PDP Factories"
  and "for most use cases, consider `getMetadataForAllProjects`". Using the non-retrying variant is
  almost certainly right here since this module supplies its own bounded loop — but the
  otherwise-exhaustive phase-2 comment never says why the documented default was rejected.
- **Two comments overstate the code.** (a) "The probe runs BEFORE the deadline check so it always
  gets at least one attempt" — if the deadline has genuinely passed, `settleWithin` returns
  `TIMED_OUT` on its `remainingMs <= 0` guard and the probe's result is discarded (the request is
  still issued, costing a fan-out). The covering test advances to 4,999 of 5,000 ms, i.e. just
  *under* the deadline, so the stronger claim is untested. (b) The `escapeStringRegexp` comment says
  "the same exact-match predicate", but `includeProjectInterfaces` compiles to *unanchored* `RegExp`s
  applied with `.test()` (`project-lookup.service-model.ts:693`) — escaping is right, but it buys a
  literal substring match, not an exact one.
- **Backoff cadence is unpinned.** No test crosses the `INITIAL_READINESS_POLL_ATTEMPTS` (10)
  transition from 1 s to 2 s, so dropping or inverting the backoff stays green — the sibling
  Power-mode loop *is* pinned (`startup-tasks.test.ts:277`). `shouldLogProbeAttempt` is likewise
  untested.
- **`EXTENDED_READINESS_POLL_INTERVAL_MS = 2000`** duplicates `EXTENDED_RETRY_INTERVAL_MS = 2000` in
  `startup-tasks.ts:73`. The `INITIAL_*` constants are documented as deliberate aliases; this is the
  one value that can silently drift — share it or give it the same note.
- **Nothing mechanically catches drift** between the two `'platformScripture.USJ_Chapter'` literals.
  The rationale for not hoisting into `src/shared` is sound — maybe a test asserting the two are
  equal, or a line saying the coupling is intentionally unenforced.
- **Stale test comment** at `startup-readiness.util.test.ts:195`: "`Promise.race` settles on
  whichever settles first", but the implementation deliberately avoids `Promise.race` for
  `AsyncVariable` and says so at `startup-readiness.util.ts:104`. The phase-2 twin at line 313 words
  it correctly.
- **Power mode left ungated** is a well-argued deferral in ADR-0014, but "revisit when the freshness
  anchor is reworked" has no ticket. Worth filing one.
- Tiny: `A_PROJECT` in the util test could be `satisfies ProjectMetadata`; and
  `waitForScriptureWorkspaceReady`'s TSDoc has no `@param`/`@returns` unlike the other
  `src/main/*.util.ts` exports (though `startup-tasks.ts:140` omits them too, so the convention isn't
  uniform).
- A note for later, not this PR: a timed-out `waitForNetworkObject` leaves its
  `onDidCreateNetworkObject` subscription live (`unsub()` only runs from inside the callback). It's
  pre-existing shared-service behavior and one instance per process, but this is a new long-budget
  caller.

### Pattern conformance — checked, clean

File placement matches the `src/main/*.util.ts` family with the test co-located, while the
`shutdown-latch.service` test correctly went into `src/main/services/__tests__/`. Bounded waits use
`AsyncVariable` rather than hand-rolled `Promise.race` + `setTimeout` per
`Code-Style-Guide.md:84-86`, guarding on `hasSettled` the same way
`network-object-status.service.ts:53-64` and `data-provider.service.ts:945` do. Main imports only
process-agnostic `@shared/services/*` clients. Comment density (~50%) is in line with
`startup-tasks.ts` (~57%) and `scheduled-session-sync.util.ts` (~68%). ADR recorded per CLAUDE.md.
Keyboard catalog / localization / Storybook / shadcn correctly N/A.

### What I liked

The tests land on exactly the cases the implementation reasons about: the `timeoutMs: 0` budget that
would otherwise hit `AsyncVariable`'s "no timeout" branch, abandonment of a never-settling
registration *and* a never-settling probe, the same-tick quit-vs-rejection race in both phases
asserting `'aborted'` not `'timed-out'`, and a regression-shaped test that would catch an
implementation with no phase 2 at all. The post-wait settings re-read closes a real hole a 120 s park
opens, and it's backed by three mid-wait mutation tests rather than just a comment. Funnelling every
await through `settleWithin` makes the budget a true ceiling even though neither underlying call is
cancellable, and `performance.now()` with the cold-boot rationale written down is a nice touch.

---

*AI-assisted review (Claude Code); findings verified against the branch, the ticket, and the PRD.*

jolierabideau (changes requested):
I left NITs in the summary comment, only one blocking.

> @jolierabideau reviewed 9 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2693#-:-P-Uu2MU5IobjAeDaYY1:bqq4qrn)* status: all files reviewed, 1 unresolved discussion (waiting on katherinejensen00).

___
*[`src/main/main.ts` line 975 at r1](https://reviewable.io/reviews/paranext/paranext-core/2693#-P-Uu-kW7WlxNoUEy711:-P-Uu-kW7WlxNoUEy712:b-dggcdq) ([raw file](https://github.com/paranext/paranext-core/blob/9dfe7002dcbcc6754212ce655988e649bcd762c9/src/main/main.ts#L975)):*
> ```XML
>           // reactivation still wants its startup sync — the predicate explains why the platform,
>           // and not the quit flag alone, has to decide that.
>           if (shouldWindowCloseAbortStartupTasks(process.platform)) startupTasksAbort.abort();
> ```

### The macOS abort change is a bigger deal than the diff suggests (important)

`shouldWindowCloseAbortStartupTasks` reintroduces, on macOS, the hazard the abort existed to prevent.  
On a macOS last-window close with no quit, `isAppShuttingDown()` is still true via  
`areAllWindowsClosing()`, so `main.ts` takes the `isAppGoingDown` branch and **does**  
`await runShutdownTasksOnce(performShutdownTasks)` — but no longer aborts `startupTasksAbort`. A  
readiness wait parked for up to 120 s can then resolve and fire a whole-workspace  
`syncProjects(undefined)` _after_ the shutdown sync already ran, into a windowless process, with no  
user feedback. ADR-0014 cites "that would double-sync" as a reason to reject self-triggering, but  
neither the new TSDoc nor the ADR addresses this case — both argue only for preserving the sync  
across a dock reactivation.

I don't think it's an oversight: the abort only matters while the gate is still parked, which is  
exactly the window where the hazard lives, so it's a genuine trade-off. But right now it's decided  
silently in favor of firing. Two shapes that would resolve it: re-check window/shutdown state  
immediately before firing (skip when the app is windowless), or hand the pending sync to the next  
session via the `resetShutdownLatchesForNewSession` path.

Second half of the same finding: `startupTasksAbort` is **shared with Power mode**. The Power  
boot-race retry loop is also no longer aborted on a macOS non-quit last-window close — which  
contradicts "Power mode is deliberately untouched" in the PR body and "Power mode unchanged" in the  
DoD, and no Power-path test covers it. If it's intentional, it should be stated as a Power-mode  
behavior change; if not, it needs scoping to the Simple gate.

katherinejensen00 (commented):
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2693)* status: 1 of 9 files reviewed, 1 unresolved discussion (waiting on jolierabideau).

___
*[`src/main/main.ts` line 975 at r1](https://reviewable.io/reviews/paranext/paranext-core/2693#-P-Uu-kW7WlxNoUEy711:-P-t_yuQ9wYB7k0lLZ5G:b-tgag6h) ([raw file](https://github.com/paranext/paranext-core/blob/9dfe7002dcbcc6754212ce655988e649bcd762c9/src/main/main.ts#L975)):*

### The macOS abort change is a bigger deal than the diff suggests (important)

`shouldWindowCloseAbortStartupTasks` reintroduces, on macOS, the hazard the abort existed to prevent.  
On a macOS last-window close with no quit, `isAppShuttingDown()` is still true via  
`areAllWindowsClosing()`, so `main.ts` takes the `isAppGoingDown` branch and **does**  
`await runShutdownTasksOnce(performShutdownTasks)` — but no longer aborts `startupTasksAbort`. A  
readiness wait parked for up to 120 s can then resolve and fire a whole-workspace  
`syncProjects(undefined)` _after_ the shutdown sync already ran, into a windowless process, with no  
user feedback. ADR-0014 cites "that would double-sync" as a reason to reject self-triggering, but  
neither the new TSDoc nor the ADR addresses this case — both argue only for preserving the sync  
across a dock reactivation.

I don't think it's an oversight: the abort only matters while the gate is still parked, which is  
exactly the window where the hazard lives, so it's a genuine trade-off. But right now it's decided  
silently in favor of firing. Two shapes that would resolve it: re-check window/shutdown state  
immediately before firing (skip when the app is windowless), or hand the pending sync to the next  
session via the `resetShutdownLatchesForNewSession` path.

Second half of the same finding: `startupTasksAbort` is **shared with Power mode**. The Power  
boot-race retry loop is also no longer aborted on a macOS non-quit last-window close — which  
contradicts "Power mode is deliberately untouched" in the PR body and "Power mode unchanged" in the  
DoD, and no Power-path test covers it. If it's intentional, it should be stated as a Power-mode  
behavior change; if not, it needs scoping to the Simple gate.
</blockquote></details>

Agreed on both halves — this was the right call, and it's what c7d6ab977ff addresses. The abort is unconditional again (main.ts:990), so Power's retry loop behaves exactly as on main everywhere; the macOS carve-out moved onto its own signal (startupReadinessAbort / readinessAbortSignal) so it can't reach Power. Your first shape is in as canStartupSyncFireNow, re-checked immediately before the command goes out. Power coverage added: never consults the windowless guard in power mode / never consults readiness in power mode.

Note you may be on an older revision — shouldWindowCloseAbortStartupTasks was renamed shouldWindowCloseAbortReadinessWait in that commit, and the ADR is now 0027.

One residual your second shape would have covered and this one doesn't: performStartupTasks runs once per process, so if readiness resolves during the windowless window the sync is skipped and nothing re-runs it — that session loses its startup sync. The shutdown sync already pushed local work, so the cost is missing incoming changes on a reactivated session. Happy to take the resetShutdownLatchesForNewSession route if you think that's worth closing now.

Real-Mac verification is still outstanding — Jolie has offered to run the dock-reactivation sequence, which is the one case unit tests can't reach.

jolierabideau (approved):
## Re-review (r2) — approving

The blocking finding is genuinely fixed, both halves, and I verified it in the code rather than
against the PR body. 10 of my 12 minor findings are resolved or answered. Two were dropped without a
response — non-blocking, noted below.

Checks on `f43581c`: **76 tests pass** across the three affected files (up from the 60 I ran on r1),
`npm run typecheck` clean, `eslint` clean on all eight changed files.

### The blocking finding — both shapes are in

| Half | Status |
| --- | --- |
| macOS abort reintroduces the hazard the abort prevented | **Fixed.** `canStartupSyncFireNow` (`shutdown-latch.service.ts:100`) is re-checked immediately before the command goes out (`startup-tasks.ts:277`) and skips when the app is resident but windowless. It also guards the inverse trap: no window *yet* is the app coming up, not going down, gated on `hasCreatedWindowThisProcess` — I'd have missed that one. |
| …and `startupTasksAbort` is shared with Power mode | **Fixed.** `startupTasksAbort.abort()` is unconditional again (`main.ts:990`); the carve-out moved to a separate `startupReadinessAbort` that only reaches the Simple readiness wait, and a real quit aborts both (`main.ts:1404`). "Power mode unchanged" is now literally true, and two tests pin it (`never consults the windowless guard in power mode`, `never consults readiness in power mode`). |

### The backoff pin — verified by mutation

This was my weakest-coverage complaint, so I checked it properly rather than reading the test: I
mutated `EXTENDED_READINESS_POLL_INTERVAL_MS` 2000 → 1000 and
`pins the two-phase probe backoff and the attempt-log sampling` **failed**; restoring it passed.
That's real coverage, not a test-shaped comment. The log-sampling assertion pinned to the literal
attempt numbers (1 and 10) is the right call too.

### Everything else, verified

Studio's removed `Task.Run(SyncProjects)` now cited precisely enough to check
(`Architecture-Decisions.md:1461`, `repo-patches/paranext-core.patch` lines 5334–5336 at `f779810`);
the `getMetadataForAllProjectsWithoutRetries` rationale written down; both overstated comments
corrected — the probe-before-deadline one now claims only that the request is *issued*, and the
`escapeStringRegexp` one now says literal **substring**, which is the correct reading; the duplicated
`2000` documented as accepted-and-not-overlooked drift; the `USJ_Chapter` coupling documented as
intentionally unenforced, with a reciprocal note added on the renderer side that I didn't ask for;
stale `Promise.race` comment gone; `satisfies ProjectMetadata` and the missing `@param`/`@returns`
both in.

The phase-1 non-exhaustion rejection you found while doing the above is the best catch in this
revision — a status-service rejection collapsing a 120 s gate into seconds would have reintroduced
the original bug invisibly, and "the budget, not the rejection, decides whether the gate is over" is
the right rule. Good instinct to write down that the path is currently unreachable from `main` rather
than implying it fixed an observed failure.

### Two non-blocking loose ends

Neither gates this merge — but I'd rather they be decided out loud than left silent.

1. **Empty-workspace polling cost got more prose, not a decision.** My ask was to *consider* capping
   the probe count or treating registered-but-empty as ready after N attempts. The cost is now
   documented more thoroughly (`startup-readiness.util.ts:334-337`), but the behavior is unchanged: a
   workspace with no scripture projects still pays ~65 unscoped fan-outs across every registered
   factory, including the .NET provider this gate exists to protect. It's absent from the response
   table entirely, so it reads as dropped rather than declined. Happy for it to be declined — just
   say so, in the ADR or the squash message.

2. **Neither deferral has a ticket.** I asked for one for Power-mode-left-ungated; the PR body also
   folds the `waitForNetworkObject` subscription leak into "the Power-mode follow-up ticket". That
   ticket doesn't exist — `PT-4386` is the only ID anywhere on the branch, and the ADR still says
   "revisit when the freshness anchor is reworked" with nothing to revisit from. Two deferrals now
   point at a ticket that was never filed; worth filing before this merges so the pointers land
   somewhere.

### Your question back to me

On the once-per-process residual (readiness resolves while windowless → that session loses its
startup sync, nothing re-runs it): **don't close it now.** It's strictly narrower than what it
replaced — a missed incoming-changes sync in a rare macOS window beats a whole-workspace S/R firing
into a windowless process after the shutdown sync — and the shutdown sync already pushed local work.
It belongs with the follow-up ticket in note 2 rather than growing this PR. If you take the
`resetShutdownLatchesForNewSession` route later, I'd want it on its own.

### Before you squash

- The first commit's message still says "Recorded in ADR-0013" — stale, as you noted.
- The kebab-case `ADR-startup-sync-readiness-gate` rename is the right resolution of the collision,
  and matches what the log's own guidance asks the second merger to do.

Real-Mac dock-reactivation verification is still the one path unit tests can't reach, and it's the
path this revision reworked. I'm not gating on it — I'll run that sequence myself and report back on
the ticket; if it misbehaves we fix forward.

---

*AI-assisted review (Claude Code); findings verified against the branch at `f43581c`, including a
mutation check of the new backoff test.*

jolierabideau (approved):
> @jolierabideau reviewed 8 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2693)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2695, opened 2026-08-19
PT-4363: Internet & Connectivity settings UI refinements

katherinejensen00:
# PT-4363: Internet & Connectivity settings UI refinements

**Base**: `origin/main` · **Files**: 28 source + 17 generated `dist/` · **Reviewed with**: Claude Opus 5 · **Date**: 2026-08-19

## Summary

Refinements to the Internet & Connectivity settings, which appear in two places — the standalone web
view and the first-run wizard step. Both are driven by shared `platform-bible-react` components, so
most changes land once and apply to both.

1. **Option descriptions → hover/focus tooltip**, signalled by a new **info icon** on each row. The
   always-visible description paragraphs are gone.
2. **The wizard step gains the "Internet & connectivity" headline** the standalone view already had,
   plus a short subtitle.
3. **The developer server picker becomes radio buttons** instead of a toggle group, and **Test gets
   its own row** alongside Production and Development.
4. **The "Disabled options are planned for future updates" footer is removed** — the per-row "Coming
   soon" badge already says it, and dropping the line keeps the wizard's Next button in view.

**Two things surfaced while building it**, and are included:

- **A latent bug the wizard would have hit.** The first-run gate is an opaque dialog at
  `Z_INDEX_FIRST_RUN` (700); tooltips portal to `document.body` at `Z_INDEX_TOOLTIP` (550). Any
  tooltip in the wizard would have painted _behind_ the gate — silently defeating change 1 in half
  the places the ticket asks for it. Fixed by portalling tooltips into the gate.
- **That workaround already existed for popovers**, so both now build from one shared factory instead
  of two copies inside vendored shadcn files.

Also folded in: a shared `WizardStepHeading`, with **all five wizard steps** converged onto it —
the wizard previously rendered step headings at two different font weights.

## Screenshots

Captured from Storybook after the follow-up commit below, so nothing here shows superseded layout.
The first-run surfaces are only reachable there — an already-registered install never renders the
wizard.

**1 · Option list at rest.** Descriptions are gone from the body copy; every row carries an info
icon, unavailable rows keep their "Coming soon" badge, and the "Disabled options are planned for
future updates" footer is gone.

01-option-list-at-rest

**2 · Description revealed on hover.** The whole row is the hover target, not just the icon.

02-description-on-hover

**3 · Developer server picker.** Radio buttons instead of a toggle group, with Test on its own row.

03-developer-server-radios

**4 · Wizard step headline and subtitle.** The body sits on the spinner because the step's PAPI
provider has no backend in Storybook; the heading treatment is the point.

04-wizard-heading-and-subtitle

**5 · A tooltip painting _above_ the first-run gate** — the z-index fix, in the one place it was
most at risk. This is the gap the Verification section below listed as "never seen running"; it has
now been seen, and asserted rather than only eyeballed:
`tooltip.closest('[data-testid="first-run-dialog"]') === true`. The same shot shows the info icon's
placement on a label long enough to wrap ("…sensitive locations ⓘ").

05-tooltip-above-first-run-gate

## Update — 2026-08-24

**Rebased onto `main`.** Four commits each side. The only conflicts were in the generated
`lib/platform-bible-react/dist/` bundles, including a rename/rename on a content-hashed chunk.
Rather than hand-merge minified output into something no source produces, the previous rebuild
commit was dropped and `dist` regenerated from the rebased sources. The one source file both sides
touched, `src/index.ts`, auto-merged cleanly (this branch's `TooltipPortalContainerProvider` export
and main's `useRunWhenVisible` are ~13 lines apart).

**New commit — `PT-4363: reveal option descriptions only on a user-driven focus`.** Running the UI
turned up a bug in change 1 above. The row tooltip guard asked `:focus-visible`, which cannot tell a
keyboard focus from a programmatic one: Chromium reports it as **true** for a programmatic
`.focus()` in a document that has seen no pointer input — exactly this panel's situation, since the
click that opens it lands in the host document, not the panel's own iframe. The guard therefore
never fired, and a description popped open with no user gesture **every time the panel loaded**.
Measured in the running app before the fix: `on load tooltips: 1`, with `focusVisible: true`.

The guard now tracks input modality directly. Because the decision no longer rests on
`:focus-visible` — which jsdom always reports as false — both directions became assertable in
vitest; the keyboard case is a new test that fails against the old guard. After the fix, in the
running app: `on load tooltips: 0`, `programmatic focus tooltips: 0`, `after ArrowDown tooltips: 1`.

Same commit also moves the row's info icon into the text flow rather than leaving it a flex item
beside the label, so a wrapped label keeps the icon trailing its last word instead of parking it at
the row's edge against the "Coming soon" badge (visible in screenshot 5).

**One documented trade-off**, noted at the hook: tabbing straight from the host document into the
panel's iframe fires its keydown in a document the listener cannot see, so the first row focused
that way reveals no tooltip until the next key press. Screen readers are unaffected — they read the
`sr-only` description wired to each radio via `aria-describedby`, not the tooltip.

**Verification after the rebase:** `npm run typecheck` clean · `npm test` 1507/1507 pass ·
`platform-bible-react` unit 459 pass · `platform-bible-react` `storybook (chromium)` 7 pass for the
option list, including the keyboard-focus story.

**Still unrun: the new e2e test** (`a programmatic focus on a radio reveals no tooltip`). It fails
to reach its assertion locally, along with all seven pre-existing tests in the spec, because
`internet-settings.page.ts` locates the panel by its **localized** iframe title and web-view
localization loses its startup race under the Playwright harness — the same race
`comments-tab.spec.ts:113` documents and works around with a `data-web-view-id` locator. The
behaviour it guards is verified twice over instead: the Storybook browser test in Chromium, and the
live-app measurements quoted above. Switching this page object to an id-based locator would unblock
the whole spec and is worth doing, but it is shared test infrastructure and a separate call.

## Reviewing this efficiently

- **Read these two first** — they carry the substance:
  `advanced/internet-access-option-list/internet-access-option-list.component.tsx` and
  `components/portal-container.context.tsx`.
- **`first-run-overlay.component.tsx` looks huge and isn't.** ~150 changed lines, **~14 real** — the
  rest is indentation from wrapping the dialog's children in a provider. Use `git diff -w`, or the
  "Hide whitespace" toggle in GitHub's diff view.
- **Skip `lib/platform-bible-react/dist/`** (17 files) — generated by `npm run build:pbr`, committed
  because the package tracks its build output.

### Shared components — `lib/platform-bible-react`

| File | What it does |
| --- | --- |
| `advanced/internet-access-option-list/…component.tsx` | Descriptions → row tooltip; adds the info icon and an `sr-only` copy wired to each radio via `aria-describedby`; drops the footer; instance-scoped ids |
| `advanced/developer-section/…component.tsx` | `ToggleGroup` → `RadioGroup`; Test gains its own row; `isSurfaced()` decides which values display as themselves vs collapse to Production |
| `shadcn-ui/tooltip.tsx` | **New export** `TooltipPortalContainerProvider` — the z-index fix |
| `shadcn-ui/popover.tsx` | Collapsed its duplicate portal workaround onto the shared factory (behavior unchanged) |
| `components/portal-container.context.tsx` | **New, internal.** `createPortalContainerContext()` — one definition, one context per primitive |
| `src/index.ts` | One added export |

### Wizard — `src/renderer/components/first-run`

| File | What it does |
| --- | --- |
| `first-run-overlay.component.tsx` | Wires the tooltip provider at the gate, so **every** step gets working tooltips (~14 real lines) |
| `steps/internet-settings-step.component.tsx` | Renders headline + subtitle across all states, so they don't pop in when the provider resolves |
| `wizard-step-heading.component.tsx` | **New.** One definition of the step `<h2>`, used by `WizardStepForm` and directly by steps that own their layout |
| `steps/language.component.tsx` | Converged onto `WizardStepHeading` |
| `steps/sync-progress.component.tsx` | Converged both branch headings (in-progress and complete) onto `WizardStepHeading`, retiring the last hand-rolled `font-medium` step heading |

### Localization — `extensions/src/paratext-registration/contributions/localizedStrings.json`

| Change | Key |
| --- | --- |
| **Added** (en + es) | `%internetSettings_subtitle_short%` — the wizard step's subtitle |
| **Deprecated + removed** (en + es) | `%paratextRegistration_internetUse_footer%`, with a `deprecationInfo` note and no replacement key |

The removal follows the immutability convention: a `deprecationInfo` entry records why, rather than
editing a live string's meaning in place.

### Tests & stories

New: `tooltip.test.tsx`, `popover.test.tsx` (both portal targets, both **verified falsifiable** by
sabotage), `tooltip.test-utils.ts`, `wizard-step-heading.component.stories.tsx`, and a
bundled-extension Storybook story for `DeveloperSection` that renders it with real localized strings.
Updated: both component tests, the first-run overlay test (now guards the z-index wiring), the
language and internet-settings step tests, and both lib story files.

### E2E — `e2e-tests/tests/isolated`

`internet-settings.page.ts` grew a page object; `internet-settings.spec.ts` was refactored onto it.
**This spec was previously broken** — see _Needs a decision_ #3.

## Needs a decision from you

1. **Tooltip guideline deviation — for Alex.** `Guidelines/Tooltips` says _"a tooltip is a hint, not
   an explanation"_ and its Avoid example is 68 characters; these descriptions run **67–162
   characters**, up to two sentences. The author chose to keep the full descriptive copy (shortening
   needs new strings) and add the info icon for discoverability. **This wants explicit UX sign-off,
   not silent acceptance.**
2. **`Guidelines/Product names` is wrong.** It states _"'Send/Receive' is now called Sync"_. Per the
   author these are **distinct operations** — Sync follows up a Send/Receive by also handling
   connected resources and model texts. As written, that page will keep prompting engineers to
   rename Send/Receive → Sync across the 10 strings that legitimately use it. Route to its owner.
3. **The e2e spec has likely never run green.** `internet-settings.spec.ts` sits at the top level of
   `tests/isolated/`, which `run-isolated.mjs` cannot select (it enumerates subdirectories), and CI
   runs only `test:e2e:smoke`. That is how a call to `FrameLocator.filter()` — **a method that does
   not exist** — survived in it. This PR fixes that call and a case-sensitivity bug beside it, but
   wiring the suite into a real run is a team decision.

## Verification

| Check | Result |
| --- | --- |
| `npm run typecheck` | ✅ clean |
| `npm run lint` (repo-wide, incl. lib workspace) | ✅ clean |
| `npx typedoc` | ✅ exit 0 |
| `src/renderer/components/first-run` | ✅ 128 pass |
| `platform-bible-react` `unit` (jsdom) | ✅ 399 pass / 40 files |
| `platform-bible-react` `storybook (chromium)` | ⚠️ **no trustworthy reading** — see below |
| Wizard tooltip rendering above the gate | ⚠️ **never seen running** — see below |

**Two gaps, stated plainly:**

- **The wizard tooltip fix has not been visually confirmed.** The app will not start in this worktree
  (`dev-packages/scripture-editors` has ~900 dirty entries, so `link-dev-packages` aborts and takes
  `npm start` with it). The fix is reasoned, unit-tested at the mechanism level, and
  regression-guarded at the wiring level — but nobody has seen it. **Please eyeball it on a working
  machine.**
- **The Storybook browser project could not be measured here.** Two consecutive runs _on identical
  source_ gave 8 failures / 1047 tests collected, then 26 / 685 — machine saturation, not code. The
  giveaway: `badge.stories.tsx > Default` is a static render with no play function and it failed on
  `Test timed out in 15000ms`, i.e. the runner never reached it. The deterministic `unit` project
  covers every component this branch touches and is green. **Let CI arbitrate the browser project.**

<details>
<summary><b>Full review record</b> — findings, API surface, interview notes</summary>

## API Changes

- **Added** `TooltipPortalContainerProvider` (`shadcn-ui/tooltip.tsx`, re-exported from
  `src/index.ts`) — props `{ container: HTMLElement | null; children: React.ReactNode }`.
- `PopoverPortalContainerProvider` — implementation moved from a `function` declaration to a `const`
  bound to the shared factory; emitted declaration changes from `export declare function …` to
  `export declare const …`. Props type and runtime behavior unchanged.
- New internal module `src/components/portal-container.context.tsx` exporting
  `createPortalContainerContext` — deliberately **not** exported from `src/index.ts` or the
  experimental entry point.
- `DEVELOPER_SECTION_STRING_KEYS` — gained `%paratextRegistration_label_serverType_option_Test%`.
  Consumers pass the resolved map straight through, so no consumer change is required.
- `INTERNET_ACCESS_OPTION_LIST_STRING_KEYS` — dropped the footer key. Still derived from
  `OPTION_ROWS`, so it stays in sync automatically.
- `DeveloperSectionProps` / `InternetAccessOptionListProps` — no type changes; TSDoc wording only.
- `WizardStepHeading` — new, renderer-internal, not a published package API.
- `lib/platform-bible-utils`, `lib/papi-dts/papi.d.ts`, extension `*.d.ts`: no changes.

No exports removed and no signatures narrowed.

## Findings

23 findings across four analysis passes (API/correctness, style/patterns, coverage/compliance, UX).

### Critical — all fixed in review

- [x] **TypeDoc build break.** `{@link createPortalContainerContext}` could not resolve (the factory
      is internal) and `typedoc.json` sets `invalidLink: true` + `treatWarningsAsErrors: true`, so
      `npx typedoc` exited **4** — breaking `npm run build` for the package and the `publish-docs` CI
      job on merge. _(fixed: inlined a self-contained contract on both providers, which also resolved
      the TSDoc finding below; typedoc now exits 0)_
- [x] **`@typescript-eslint/no-shadow` ×2** — `radioId`/`descriptionId` parameters named `value`
      shadowed the `value` prop. _(fixed: renamed to `optionValue`)_
- [x] **`no-type-assertion`** on `row as HTMLElement` in the hover story. _(fixed:
      `querySelector<HTMLElement>` + explicit throw, which also made that story falsifiable)_

All three were introduced by the author's own final edits and missed by earlier checking — see
_Verification gaps_ below.

### Important

- [ ] ~~**Tooltip copy violates `Guidelines/Tooltips`.**~~ _(Author kept the descriptive copy
      deliberately. **Escalated to Alex** — see _Needs a decision_ #1.)_
- [x] **Nothing signalled the descriptions existed** — no icon, just a 300 ms dwell, so anyone
      clicking through never saw one. _(fixed: `Info` icon per row. Deliberately decorative —
      `aria-hidden`, no tab stop — since making it the sole trigger needs an accessible name, i.e. a
      new localized string, plus five extra tab stops. The whole row stays the trigger.)_
- [x] **Public-API TSDoc stopped standing on its own**, pointing at an unexported symbol;
      `PopoverPortalContainerProvider` lost its inline contract.
      `Code-Style-Guide.md#api-surface-tsdoc-requirements` requires these to be self-contained.
      _(fixed: contract inlined into both)_
- [x] **`WizardStepHeading` had no story** — the only component in `first-run/` without one.
      _(fixed: three stories, incl. `WeightMismatchToday` showing the remaining weight mismatch)_
- [x] **The z-index fix had no regression guard.** The overlay test stubbed the provider as a
      pass-through and ignored the new `ref`, so deleting either left all tests green. _(fixed: the
      stub records its `container`, `DialogContentStub` forwards its ref, and a new test asserts both.
      **Verified falsifiable** — removing the `ref` fails it; removing the wrapper fails it.)_
- [ ] ~~**Bare provider aliases** share a runtime name in React DevTools and use an anonymous props
      type.~~ _(Cosmetic/dev-tooling only; TSDoc now carries the contract. Left as-is.)_

### Minor

Fixed: developer-section row padding now matches the option rows · row-wide hover highlight now
matches the clickable area in both lists · raw `<label>` vs shadcn `Label` asymmetry documented
(`Label`'s `peer-disabled:` is a **sibling** combinator and the option-list label is nested two
levels down, so `Label` genuinely cannot work there) · unused `export default` removed · React
`act()` warning fixed · added `popover.test.tsx` (**verified falsifiable**) · the
`[data-slot="tooltip-content"]` selector de-duplicated across the two vitest harnesses.

Dismissed: **"Send/Receive" → "Sync"** — the guideline is wrong, not the strings (see _Needs a
decision_ #2).

Open, left as reviewer judgment: coming-soon rows are `disabled` and so unreachable by sighted
keyboard-only users (screen readers get the `sr-only` copy; mouse users get hover) · placement of
`portal-container.context.tsx` in `components/` vs the existing `src/context/`.

### Template propagation

None. No changed file carries a `#region shared with` marker.

## Positive observations

- The `ToggleGroup` → `RadioGroup` swap matches `Component-Selection-Quick-Reference.md`.
- The subtle correctness risk in that swap — a radio fires no change event when the already-checked
  item is clicked, stranding a user on a hidden server value with no one-click route back to
  Production — was caught, handled, documented at the call site, and covered by tests including a
  negative case asserting a Test user is not silently re-routed.
- `data-testid`s survived the swap, and the e2e spec adds `toHaveRole('radio')` so they can't drift
  back onto a different control.
- Hardcoded DOM ids → `useId()`, fixing real duplicate-id collisions on Storybook autodocs pages.
- Every changed line in the vendored `tooltip.tsx`/`popover.tsx` sits under a `// CUSTOM:` comment.
- The removed footer string was deprecated with a `deprecationInfo` note rather than deleted
  silently, matching the string-immutability convention.
- Tests explain _why_ an assertion is structural rather than visibility-based (jsdom loads no
  stylesheet; Playwright counts `sr-only` as visible) — the kind of note that stops a correct
  assertion being "simplified" into a wrong one later.
- The e2e spec asserts `toBeVisible()` on `[data-slot="tooltip-content"]` rather than
  `getByRole('tooltip')`, which Radix puts on a visually-hidden copy — precisely the assertion that
  catches the bug this PR fixes.

## Interview notes

**Author decisions during review:** keep the full descriptive tooltip copy and add an info icon
(knowing deviation from the guideline, not an oversight) · leave `proper-names.mdx` alone despite its
error · do **not** rename the Send/Receive strings · track the remaining heading convergence as
PT-4402 · land everything as one PR.

**The Send/Receive correction is worth highlighting.** The analysis assumed Sync was simply the new
name for Send/Receive, following the written guideline. The author corrected it: they are distinct
operations. That invalidated the finding and redirected it at the guideline page instead. No areas
were deferred to AI or left unexplained; no unresolved items.

## Verification gaps this review exposed

These explain how three build-breaking issues survived earlier "all green" checks:

1. **`eslint --cache` masked real errors.** A repo-root `eslint .` reported clean while the
   `lib/platform-bible-react` workspace — which `npm run lint` also runs, with stricter rules — had 3
   errors. Clearing `.eslintcache` and running the workspace lint directly surfaced them.
2. **`build:basic` skips TypeDoc.** `npm run build` is `build:basic && lint-fix && typedoc`; only the
   full script validates doc links, so the TypeDoc break was invisible to every `build:basic` run.

## Follow-up

- **PT-4402 — completed in this PR.** Originally deferred, then finished once it was clear the two
  `SyncProgressStep` headings are mutually exclusive branches of a single step title ("Syncing your
  projects." → "Sync complete"), already marked up as `<h2>`; the `role="status"` wrapper governs
  screen-reader announcement, not heading semantics. Leaving one step at `font-medium` while the
  other four moved to `font-semibold` would have shipped a lone outlier — worse than either
  consistent state. All five steps now share `WizardStepHeading`, and the deferral note plus the
  `WeightMismatchToday` story that documented the mismatch are removed.

</details>

## Update — 2026-09-08

**Rebased onto `main`.** Seven commits each side. As with the previous rebase, the only conflicts
were the generated `lib/platform-bible-react/dist/` bundles (including a rename/rename on a
content-hashed chunk), so the rebuild commit was dropped and `dist` regenerated from the rebased
sources. No source conflicts.

**New commit — `PT-4363: Say so when a stored internet setting cannot be honored`.** Sebastian's
review point: we do not need to know *who* wrote `InternetSettings.xml` — a one-time copy from
Paratext 9 is fine — but if the file selects an option we cannot handle yet, the user has to be told
explicitly rather than have it silently dismissed.

Two of the four persisted `InternetUse` values are in that state today: `Disabled` ("Disable all
Internet access") and `ProxyOnly` ("Configure proxy"). Before this commit the option list rendered
such a value as a selected-but-disabled row and said nothing, which reads as if the app had quietly
picked something else.

Now:

- **The row stays selected** — the stored setting is never swapped out from under the user — and the
  list raises a banner above it naming the option and saying it will not take effect until they
  choose an available one. The banner is `role="alert"`, so it is announced, not just visible.
- **The first-run wizard holds `Next` shut** while an unsupported value is the current selection,
  and unlocks it as soon as the user picks one the app implements. Changing only the developer
  server does *not* unlock it — the gate is re-derived from the value actually persisted.
- **The standalone Internet & connectivity panel gets the banner but no gate.** It has no next step
  to block, and Save is already inert until something changes.
- A new experimental export, **`isSupportedInternetUse`**, is what the wizard gates on, so the list
  of implemented options lives in exactly one place (`OPTION_ROWS`).

New stories cover all three surfaces: `InternetAccessOptionList → ComingSoonSelected` /
`ComingSoonProxySelected`, `InternetSettingsForm → UnsupportedStoredValue`, and
`InternetSettingsStep → UnsupportedStoredValue`.

**Two defects found in the same component while there**, both fixed in the commit:

- **The panel was printing a raw localize key.** Change 4 above removed the "Disabled options are
  planned for future updates" footer string, but only the *wizard* passed `showFooter={false}`; the
  standalone panel still rendered the paragraph, and with the key gone the localization service
  falls back to returning the key itself — so the panel showed the literal text
  `%paratextRegistration_internetUse_footer%`. The footer and its `showFooter` prop are gone
  entirely now, which is what the deprecation note said should happen. The e2e spec already
  asserted the footer's *English* text was absent, which is why it stayed green.
- **Every story in the extension's `InternetAccessOptionList` file crashed.** Its decorator passed
  `args` to `<Story>`, which replaces the story's args rather than merging into them, so
  `localizedStrings` arrived `undefined`. The resolved args are spread back in now. (The sibling
  `InternetSettingsForm` stories were unaffected — their decorator passes every arg explicitly.)

**Verification.** `npm run typecheck`, `npm run lint`, and `npm test` are clean. `dotnet test` was
not run: no C# changed. The two `import/no-unresolved` lint errors in `platform-scripture-editor`
are pre-existing and unrelated (an unlinked yalc dev package). Storybook screenshots confirmed the
banner renders in the standalone panel and the option list; the `InternetSettingsStep` stories
cannot render in Storybook at all (`Cannot spy on export "useDataProvider"` — pre-existing, affects
`Default` equally), so the wizard gating is covered by unit tests instead.

## Update — 2026-09-11

**Rebased onto `main`** (21 new commits there). As before, the only conflicts were in the generated `lib/platform-bible-react/dist/` bundles. The rebuild commits were dropped and `dist` regenerated from the rebased sources. The one source file both sides touched, `toolbar-compound-label.component.tsx`, auto-merged. Main's comment change is kept, and this branch no longer changes that file at all (see below).

**Review responses** are in the conversation thread: one reply to Jolie's review, one to Sebastian's icon-button point. The localization, reuse and smaller-item fixes are in `PT-4363: Address review — keep the wizard's translated headings, replace the PT9 subtitle`.

**New commit — `PT-4363: Make each option's description an info icon button`.** Follows Sebastian's pointer to the Providing Help guideline (#2787): use an icon button, not a bare icon.

- Each row's info icon is now a focusable ghost icon button, and it is the **only** tooltip trigger. Its `aria-label` carries the description.
- It is never disabled, so the three "Coming soon" rows are reachable by keyboard for the first time. The descriptions also stay readable while the list is loading or saving.
- Screen readers still get each description on its radio via `aria-describedby`.
- Clicking anywhere on a row still selects its option (the label's click target stretches over the row), and the button sits above that target. The `RowClickTargets` story hit-tests both in Chromium. It fails with the button's stacking removed.
- The descriptions fit #2787's 1–2 sentence allowance for an info icon button's tooltip, so the tooltip-length deviation is closed.
- The radios are no longer tooltip triggers, so the load-time programmatic focus can't reveal anything. The interaction-modality workaround is removed with its hook, and `toolbar-compound-label` is identical to `main`. The Playwright test `a programmatic focus on a radio reveals no tooltip` stays as the guard.

**6 · Hovering an info button.**

06-info-button-hover

**7 · Keyboard reach on a "Coming soon" row.** Tabbed onto "Disable ALL internet access"'s info button. Its radio is disabled and takes no focus, so this description was unreachable by keyboard before.

07-info-button-keyboard-coming-soon

**8 · Narrow column.** The button trails the wrapped label's last word ("locations") instead of parking beside the badge.

08-info-button-narrow-column

**9 · Inside the first-run gate.** The tooltip still paints above the gate when opened from an info button.

09-info-button-tooltip-above-first-run-gate

**Superseded in the sections above:** screenshot 2's "the whole row is the hover target"; the 2026-08-24 modality-tracking fix and its "one documented trade-off"; _Needs a decision_ #1 (tooltip length); and the review record's "deliberately decorative — `aria-hidden`, no tab stop" and "coming-soon rows … unreachable by sighted keyboard-only users".

**PT-4402 stays in this PR.** We have no UX designer at the moment, so it isn't waiting on that sign-off. Here are both sync-progress headings rendered through `WizardStepHeading` (`tw:font-semibold`), for QA:

**10 · Sync in progress** (`SyncProgressStep → Mid Sync`)

10-sync-progress-in-progress

**11 · Sync complete** (`SyncProgressStep → All Done`)

11-sync-progress-complete

**Verification.**

| Check | Result |
| --- | --- |
| `npm run typecheck` | ✅ clean, after reinstalling for main's staged dev packages |
| ESLint on every changed file (root, extensions, `platform-bible-react` configs) | ✅ clean |
| `platform-bible-react` unit: option list + toolbar label | ✅ 46 pass |
| `platform-bible-react` storybook (chromium): option list stories | ✅ 9 pass; `RowClickTargets` verified falsifiable |
| `src/renderer/components/first-run` | ✅ 135 pass |
| `paratext-registration` | ✅ 12 pass |
| Repo-wide `npm run lint` / `npm test` | ⚠️ killed twice when this shared machine ran out of memory, so CI has the final word |
| Playwright `internet-settings.spec.ts` | not run (see the 2026-08-24 note on why it doesn't run locally) |
| `dotnet test` | not run: no C# changes |

AI-assisted — [session 1](<session URL>), [session 2](https://claude.ai/code/session_01AFMWFjTSQvSGvYW8JgXLEN)

https://claude.ai/code/session_01AFMWFjTSQvSGvYW8JgXLEN

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2695)


Sebastian-ubs:
Thanks Katherine, I love those changes.
1 little ask: Can we add the word "server" to the "developer only" options? Like "Production server", "Development server", "Test server".
I know that Ian and Alex had some thoughts of removing it completely from this ui or making it harder to access, as it is also not easily possible in PT9 to switch servers for a normal user, but I do not have more details on it.
We may additionally need an highly visible indication of the current server (if not production) - like PT9 has, but this is out of scope for this PR.

katherinejensen00:
@Sebastian-ubs Thanks for the encouragement! I added server. I agree that helps with context and it is an easy thing to change in the future.

Sebastian-ubs:
oh, just reading the subtitle now. It says
> these settings [...] are shared with any co-installed Paratext 9.

I don't think this is true. Can you confirm?
It was intended they are imported from 9 (although I earlier heard you say they are currently not), but I was under the impression changing them in 10 does not do any changes to 9, right?
So in this case we need to change the label.

katherinejensen00:
@Sebastian-ubs 
Confirmed — you're right, and thanks for catching it. The subtitle is wrong.

**Changing them in 10 does not touch 9.** Platform.Bible points ParatextData at its
own private data folder, `~/.platform.bible/projects/Paratext 9 Projects`, while
Paratext 9 uses its own (`C:\My Paratext 9 Projects` by default). `InternetSettings.xml`
is written into whichever settings directory ParatextData was initialized with, so ours
lands in the `.platform.bible` tree and 9 never reads it. The folder being *named*
"Paratext 9 Projects" is almost certainly where the impression came from — it's just a
name inside our own tree.

On my machine the two directories each have their own separate copy of the usual
ParatextData settings files (`CountryStatuses.xml`, `usfm.sty`, …), which is the
separation in action.

**And they aren't imported from 9 either** — your recollection matches the code. There's
no import path: nothing reads a Paratext 9 location for these. On top of that, we stub
out every Windows-registry read, which is how ParatextData normally discovers Paratext 9's
shared settings directory, so 10 couldn't find 9's settings even if it wanted to.

Worth noting we already made this call once: the registration-side string that said
"These settings also apply to Paratext 9" was deprecated on 2025-09-10 with the note
*"The registration in Paratext 10 Studio is no longer shared with Paratext 9."* The
internet-settings subtitle came later (PT-4235) and reintroduced the same claim.

So — new wording. Two options:

**A (just drop the false claim):**
> Manage how Paratext accesses the internet. These settings only apply to the Paratext
> app, not to all internet traffic on your computer.

**B (actively correct the expectation, for anyone with 9 co-installed):**
> Manage how Paratext accesses the internet. These settings apply only to Paratext 10
> Studio — not to all internet traffic on your computer, and not to a co-installed
> Paratext 9.

I lean towards B: the folder name and the shared heritage make "these are the same
settings as 9" an easy assumption, and this is the one place we can head it off. Happy
with A if you'd rather keep the subtitle short.

Only the settings-page subtitle is affected — the wizard step uses a clipped version
that stops after the first sentence, so it never made the claim.

Two notes for you before this goes out:

A wording decision is genuinely open. I recommended B but it's your call — and if you pick B, the phrase "Paratext 10 Studio" is doing real work (naming the product rather than "the Paratext app") because the whole point is distinguishing it from 9.

Implementation caveat worth knowing when we act on it: this is a meaning change, not a phrasing fix, so per the localization guide it needs a new key plus a fallbackKey on the old one rather than an in-place edit — unlike the "server" change, which was a pure clarification. Spanish needs the same treatment. That's a separate small commit; say the word and I'll do it on this branch or a follow-up.

Sebastian-ubs:
> I lean towards B

Means `I = Katherine` or `I = ai that always prefers the longer texts` ?
🙈

katherinejensen00:
@Sebastian-ubs I = Both. AI wrote it, but our team leans towards the more detailed/longer when in doubt so I agree with its recommendation.

Sebastian-ubs:
Actually my wish for the users is C - read setting from PT9 initially.
But this might be out of scope - is it?

katherinejensen00:
@Sebastian-ubs 

## Can we read the internet setting from Paratext 9?

**Yes — once, at first launch, and then never again.** ParatextData already copies PT9's
`InternetSettings.xml` into our own AppData folder the first time we start, via the same mechanism
that carries a PT9 user's registration across. So the value is already pre-selected in the first-run
wizard today, with no code from us.

### Read these two before concluding it's solved

**1. The two copies diverge from the second launch onward.** This is a one-time seed, not a shared
setting. PT9.5 writes `%LOCALAPPDATA%\Parate. Change the setting in
either app and the other never hears about it — permanently. Our UI currently tells the user the
opposite: that these settings "are shared wi9"
(`localizedStrings.json:137`, and the Spanish twin at `:217`). That string is wrong and should be
fixed whichever option we pick.

**2. We cannot tell a user's choice from PT9file with its own defaults
on *its* first run, so the file always exists and always carries a value. An inherited `VpnRequired`
may mean the user picked it, or may mean thes. Nothing in the file
distinguishes the two.

Neither limit is fixable from our side — both are properties of how PT9 stores the setting. They're
what makes this a UI decision rather than a

### Options

| | What | Cost | Catch |
|---|---|---|---|
| **A** | Do nothing — the value already carted | zero | Wizard still asks a question the user effectively answered in PT9 |
| **B** | Keep the step, label it "Imported ct a PT9 source | ~½ day + 1 string | Needsthe explicit PT9 read (easy — below) |
| **C** | Skip the step entirely when PT9 he2 days | `NUMBERED_STEPS` is a static arraydriving "step N of 4", so this needs a dynamic step list — and limit 2 means it only works for 3 of the 4 values |

**Recommendation: B**, plus fix the "shared with Paratext 9" string and file the v1.0 bug below.

The case against C is limit 2: of the four values, only `Enabled`, `Disabled` and `ProxyOnly` are
evidence a human picked something. `VpnRequi so the most common
inherited value — tells us nothing. Skipping on it would silently decide connectivity for users who
never chose; not skipping means C skips for ich is a worse wizard than
either consistent answer.

### How the inheritance works

- `InternetAccess.cs:88` — `internetSettingsPath = Path.Combine(ParatextInfo.AppDataFolder, "InternetSettings.xml")`,
read once in a static ctor, never re-read
- `ParatextInfo.cs:173` — `folderName = "Paratext" + version.Major + version.Minor`
- `c-sharp/Program.cs:73` — we override thaton = appVersion`
- `c-sharp/SemVerUtils.cs` — builds `new Version(10, major, minor, patch)`

App version `0.6.0-alpha.0` → `10.0.6.0` → folder `Paratext100`. Confirmed in the running app's log:
`ParatextData.dll assembly version: 9.5.0.24

The copy comes from `ParatextInfo.UpgradeApps:279-321`). On the *first*
run where `Paratext100` is missing or empty, it enumerates `Paratext*`, keeps only directories
holding a `RegistrationInfo.xml`, sorts descg below `Paratext80`, and
copies four files from the first hit:

    string[] preservedFiles = { regFileName, InternetAccess.fileName,
                                releaseStage

`InternetAccess.fileName` is `InternetSettinc is in the shipped nuget
DLL (9.5.0.24), not just the `C:\Paratext Current` source.

### Separate bug in the same code path, biting at v1.0

When our app version reaches `1.0.0` the folder becomes `Paratext101`, and that descending name sort
ranks `Paratext95` above `Paratext100` (`'9'rade would re-import from
PT9.5 and silently discard the user's Platform.Bible internet settings **and** registration. Worth
its own ticket regardless of what we decide

### Two smaller caveats

- **All-or-nothing, from one directory.** Ited directory with a
  `RegistrationInfo.xml`, then returns. If that directory lacks `InternetSettings.xml`, nothing is
  imported even when an older `Paratext9x` h
- **Windows and Linux only.** `LocalApplicationData` → `~/.local/share` on Linux. No PT9 on macOS,
  so nothing to inherit there.

### What the explicit read costs (B and C)

Uses APIs we already reference:

    // %LOCALAPPDATA%\Paratext9*\InternetSet
    var memento = Memento.GetXml<InternetAccess.InternetSettingsMemento>(path);

`Memento.GetXml<T>(string)` is public in PtxUtils, `InternetSettingsMemento` is public with public
fields, and the XML root element already matelper plus a
`getPreviousParatextInternetSettings`-style function on `InternetSettingsDataProvider`.

### Still unverified

I read the code and confirmed the logic is in the shipped DLL, but haven't watched a clean-machine
first run do the copy. Cheap decisive test: xt100` aside, relaunch, and
check whether the new `InternetSettings.xml` carries `Paratext95`'s 2023-08-30 mtime — `File.Copy`
preserves it, so a matching timestamp provesta untouched since the test
shuffles registration state.

Sebastian-ubs:
I'm getting mixes messages
- earlier: it is not copied (and not possible)
- now: it is already copied. Which is true?

We do not need to distinguish who edited the file - if the file was actually copied from 9 this is fine.
A one time copy is fine.

The problem we really need to handle: If the `InternetSetttings` file has an options selected we cannot handle yet, we need to make the user aware of this explicitly, not just silently dismiss it. Suggested approach: Select the disabled option. State a banner message, disable the next wizard step until the user picks one of the available options.

katherinejensen00:
Already copied with the caveats listed is true. Sorry for the confusion. I will go ahead with your suggested changes for if someone selected an option in P9 that P10 doesn't support.

jolierabideau:
## Review summary

Nice work overall — PT-4363's three approved changes are all here and all four conditions for approval are met, and I appreciate that the descoped "Server environment" rename stayed out. A few things I want to call out as genuinely good before the findings:

- Finding and fixing the tooltip/first-run z-index collision **before** it shipped, and guarding both halves in tests.
- The `:focus-visible` diagnosis is the highlight — correctly identified as Chromium-specific and unobservable in jsdom, fixed by tracking modality directly, then made assertable in both directions with a test that fails against the old guard.
- Screen-reader users keep what sighted keyboard users lose (`sr-only` + `aria-describedby`, placed outside the `<label>` so it isn't absorbed into the radio's accessible name — with a test asserting exactly that).
- `tooltip.test-utils.ts` documenting *why* `getByRole('tooltip')` is the wrong query here.
- `// CUSTOM:` discipline on the two vendored shadcn files is unusually thorough.

### One blocker

**Translation regression for `fr` / `zh-hans` / `zh-hant`.** The wizard step moves off `%firstRun_step_internetSettings_heading%` / `%firstRun_step_internetSettings_body%` — both translated in all five `assets/localization/*.json` files (`fr`: "Internet et connectivité", `zh-hans`: "互联网与连接") — onto `%internetSettings_webView_title_2%` and the new `%internetSettings_subtitle_short%`, which live in the extension's `localizedStrings.json`. That file's `localizedStrings` block has only `en` and `es`, so French and both Chinese locales lose a translated heading *and* subtitle and fall back to English.

Worth noting: the old `_heading%` English value was already literally "Internet & connectivity", so the ticket's "same headline in both places" goal was already satisfied in `en`. Simplest fix is probably to keep the core keys for the wizard step. Either way the two orphaned keys should get `deprecationInfo` entries in `assets/localization/metadata.json`, and `%internetSettings_subtitle_short%` is character-for-character the old `_body%` minus the trailing period, which is the duplicate-key case the localization guide asks us to avoid.

### Two process questions

1. **The subtitle we agreed was wrong is still here.** Per the thread above, `%internetSettings_webView_subtitle%` still claims these settings "are shared with any co-installed Paratext 9" in both `en` (`:137`) and `es` (`:219`). The wizard sidesteps it by using the new short subtitle, but the standalone panel still renders it — so the two surfaces now say different things about PT9, which cuts against this PR's own consistency goal. Was the fix intended for the follow-up about reading PT9's setting, or should it land here? (Related: `%paratextRegistration_description_shared_with_paratext_9%` carries the same claim, has no code references, and isn't marked deprecated.)

2. **The PT-4402 commit.** `d3a9458d` converges the sync-progress headings, but PT-4402 is still in Triage/unassigned and its first condition for approval is "UX has confirmed whether the sync-progress headings should match the step-heading weight" — the ticket itself argues those two may be status messages rather than step titles. The commit message also notes it wasn't visually verified, which is the one check that ticket asks for. Would you rather split it out and gate it on UX sign-off? It also silently restyles `language` and `sync-progress` from `font-medium` to `font-semibold`, which is worth a screenshot for QA either way.

### Localization conventions

- The four `%paratextRegistration_label_serverType_option_*%` values were relabeled in place ("Development" → "Development server"). Since that's a meaning-affecting relabel rather than a typo fix, `Localization-Guide.md` ("Existing Strings Are Immutable") wants a new `_2` key plus `deprecationInfo` — and this file already has six `_2` precedents for exactly this shape (e.g. `%..._option_ProxyOnly%` → `_ProxyOnly_2%`). The removed footer key is handled correctly, which is why this reads as an oversight to me.
- Conversely, `%paratextRegistration_internetUse_footer%`'s *value* was deleted from `en`/`es` while gaining a `deprecationInfo` entry — the four existing entries in that file all retain their values alongside the successor key.
- The two new `unsupportedSelection_*` keys are `en`-only, while `%internetSettings_subtitle_short%` in the same diff got both — so Spanish users would see a mixed-language banner at the moment the wizard blocks them.

### Reuse

`useLastInputWasKeyboardRef` (`internet-access-option-list.component.tsx:41-80`) re-implements `lastInteractionModality` / `trackInteractionModality` from `toolbar-compound-label.component.tsx:11-45` — same purpose, same document-level capture-phase `keydown`/`pointerdown`, same `:focus-visible` rationale in the comment. They've already diverged: the existing one initializes to `'keyboard'` on purpose ("so a label focused before any input at all still explains itself"), the new one to pointer. Could we extract one into `src/hooks/*.hook.ts` and have both call sites use it before they drift further?

Also minor: `portal-container.context.tsx` sits at `src/components/` root, but the library has a `src/context/` for exactly this kind of module (`menu.context.ts`, `shrink-step.context.ts`).

### Accessibility of the tooltip move

This one may be a UX conversation rather than a code change, since the move is ticket-approved — but two things are worth surfacing:

- On the three "Coming soon" rows the only focusable child is a `disabled` `RadioGroupItem`, so there's no tab stop; the trigger div has no `tabIndex` and the `Info` icon is `aria-hidden`. A sighted keyboard-only user sees an info affordance they can't open, and the same gap hits touch/pointer-less input on every row. (Screen-reader users are fine, thanks to the `sr-only` spans.) Making the icon a real focusable trigger with an `aria-label`, or using `aria-disabled` so rows stay focusable, would close it.
- `Guidelines/Tooltips` asks that tooltip content be "purely supplemental — the UI is fully usable without it" and short enough to scan. These descriptions are 100–190 characters and are what tells the user what each option does. Might be worth recording an explicit UX exception.

<details>
<summary>Smaller items (expand)</summary>

- The unsupported-value banner inherits `role="alert"` (assertive) from `Alert`, but it reports state that existed before the user arrived rather than a change they caused — `role="status"` is a one-line fix since `Alert` spreads `{...props}` after `role`.
- The banner only fires for values that have a row (`OPTION_ROWS.find(...)`). A persisted value *outside* `OPTION_ROWS` — the PT9-authored case the file's own SYNC WARNING describes — leaves nothing selected, no banner, and `Next` disabled with nothing on screen explaining why. Deriving from `!isSupportedInternetUse(value)` would cover it.
- `isSupportedInternetUse` takes a file-local `InternetUse` type that isn't exported, so a caller holding a plain `string` can't call it. Maybe `export type { InternetUse }` alongside it.
- `internet-settings.spec.ts:82` still proves absence by matching English text, which is the assertion shape that stayed green through the raw-`%key%` regression. A positive guard asserting no `/%[a-zA-Z0-9_]+%/` in the frame would fail loudly next time.
- The lib story's fixture for `%…unsupportedSelection_description%` doesn't match the shipped string; the sibling extension story avoids this by resolving real strings via `getLocalizedStrings(...)`.
- `showFooter` was removed outright rather than deprecated for a release — defensible given `@experimental`, just noting the string key got a deprecation record and the prop didn't.
- Two `// CUSTOM:` gaps: `popover.tsx` drops a `// CUSTOM:` line while rewriting the JSDoc it marked (and the `container` prop semantics changed under an unchanged comment); `tooltip.tsx:188` says what but not why.
- `WizardStepHeading`'s `{@link WizardStepForm}` won't resolve (not imported, and importing would be circular).
- "Quality Assurance server" is title-cased mid-string; "Quality assurance server" matches the capitalization guideline. Developer-only surface, so low impact.
- `SERVER_ROWS` uses `as const satisfies` + an arrow guard while `OPTION_ROWS` uses an annotated array + a `function` guard — two idioms for the same shape in sibling files edited together.
- A couple of test comments carry ticket IDs for in-PR work (`internet-settings-step.component.test.tsx:150`, `language.component.test.tsx:103`) — the surrounding rationale is worth keeping, just not the IDs.

</details>

I also checked the committed `dist/` against the final sources by marker — `showFooter` and `internetUse_footer` appear in zero dist files and the new exports are present, so the dropped-and-regenerated rebuild looks clean.

---

*AI-assisted review (Claude Opus 5), findings verified against the source before posting.*

Sebastian-ubs:
> Accessibility of the tooltip move
- https://github.com/paranext/paranext-core/blob/c29d1f5392fdd3cf46211167c9e566fa67d093ff/lib/platform-bible-react/src/stories/guidelines/providing-help.mdx 
- https://github.com/paranext/paranext-core/pull/2787

Fact from there: use an icon button, not only an icon. Icon buttons are accessible.

katherinejensen00:
@Sebastian-ubs Thanks for the pointer to the Providing Help guideline — done in `PT-4363: Make each option's description an info icon button`.

- Each option's info icon is now a real ghost icon button, and it is the **only** tooltip trigger (hovering the row or label no longer opens the description). Its `aria-label` carries the full description, as the guideline asks.
- It is its own tab stop and is never disabled, so keyboard-only users can now open the descriptions on the three "Coming soon" rows too — their radios are disabled and never take focus. The descriptions also stay readable while the list is loading or saving.
- Screen readers still get each description on its radio via `aria-describedby`, so arrowing through the radios reads it without tabbing out to the button.
- Clicking anywhere on a row still selects the option (the label's click target stretches over the row), and the button sits above that target, so clicking it doesn't select anything. A Storybook story (`RowClickTargets`) hit-tests both in a real browser, and it fails if the button's stacking is removed.
- The descriptions are at most two sentences, which fits the 1–2 sentence allowance #2787 gives info icon button tooltips. That clears the tooltip-length deviation Jolie raised.

Since the radios are no longer tooltip triggers, the option list no longer needs the interaction-modality workaround that kept a description from popping open on the panel's load-time focus. That code is gone, and `toolbar-compound-label` is back to exactly what `main` has.

katherinejensen00:
@jolierabideau Thanks for such a thorough review. Below is where each point landed. The branch is also rebased onto the latest `main`.

### Blocker — translation regression
**Fixed** in `PT-4363: Address review — keep the wizard's translated headings, replace the PT9 subtitle`. The wizard step is back on `%firstRun_step_internetSettings_heading%` / `%firstRun_step_internetSettings_body%`, so `fr`, `zh-hans` and `zh-hant` keep their translations. `%internetSettings_subtitle_short%` never shipped, so it was removed rather than deprecated.

### Process questions
1. **PT9 subtitle** — landed here. `%internetSettings_webView_subtitle%` is superseded by `%internetSettings_webView_subtitle_2%` ("…apply only to Paratext 10 Studio — not to all internet traffic on your computer, and not to a co-installed Paratext 9"). The old key keeps its value and gets a `fallbackKey` + deprecation note, and es is updated too. `%paratextRegistration_description_shared_with_paratext_9%` already has a `deprecationInfo` entry on `main` (2025-09-10, "no longer shared with Paratext 9"), with its value kept alongside it like the file's other deprecated entries, so nothing further was needed there.
2. **PT-4402 commit** — keeping it in this PR. We don't have a UX designer right now, so we're not holding the PR for that sign-off. The screenshots you asked for are now in the PR description: both sync-progress states (in progress and complete), rendered through `WizardStepHeading`. If UX later decides these are a different class of heading, the revert is two one-line swaps; the Jira comment on PT-4402 has the details.

### Localization conventions
All three **fixed** in the same commit:
- The four server-type labels are back to their shipped values, moved to `_2` keys with deprecation notes (matching the existing `_2` precedents), and "Quality assurance server" is sentence case.
- `%paratextRegistration_internetUse_footer%` keeps its value alongside its `deprecationInfo`.
- Both `unsupportedSelection_*` keys have `es` translations.

### Reuse
**Resolved by removal.** The two modality trackers were first merged into one hook, but following Sebastian's icon-button guidance the option list no longer needs one. The info button is now the only tooltip trigger, and the radio that gets focused programmatically on load is not a trigger. So the hook is deleted and `toolbar-compound-label.component.tsx` is back to exactly what `main` has, with nothing left to drift. `portal-container.context.tsx` now lives in `src/context/`.

### Accessibility of the tooltip move
**Fixed** in `PT-4363: Make each option's description an info icon button`, per Sebastian's pointer to the Providing Help guideline (#2787):
- **Keyboard reach:** each row has a focusable ghost icon button with an `aria-label` carrying the description. It is the only tooltip trigger and is never disabled, so the three "Coming soon" rows are reachable by keyboard. Unit tests cover tabbing onto a Coming soon row's button, and a Storybook play function covers the same in Chromium.
- **Tooltip length:** #2787 allows 1–2 sentences for an info icon button's tooltip, and every description fits. The component TSDoc now cites that rather than recording a deviation.

### Smaller items
- `role="status"` on the banner — **fixed**.
- Banner derived from `!isSupportedInternetUse(value)`, so a value with no row still explains why Next is disabled — **fixed**, with a test.
- `InternetUse` exported alongside `isSupportedInternetUse` — **fixed**.
- e2e positive guard against unresolved `%key%` text, with a positive control — **fixed**.
- Lib story fixture matches the shipped banner string — **fixed**.
- `showFooter` removed outright — **kept as is**. It is `@experimental`, and its only consumers are in this repo.
- `// CUSTOM:` gaps in `popover.tsx` / `tooltip.tsx` — **fixed**.
- `{@link WizardStepForm}` — now plain text, **fixed**.
- "Quality assurance server" — **fixed**.
- `OPTION_ROWS` / `SERVER_ROWS` idioms — converged on `as const satisfies`, **fixed**.
- Ticket IDs in test comments — **removed**, rationale kept.

### Verification after this round
- `npm run typecheck` passes.
- ESLint is clean on every changed file, each under its own workspace config: root, extensions and `platform-bible-react`.
- Tests pass: `platform-bible-react` unit (46, option list + toolbar label), the option-list stories in Chromium (9), the renderer's first-run tests (135), and `paratext-registration` (12).
- Repo-wide `npm run lint` and `npm test` were killed twice when this shared machine ran out of memory, so CI has the final word on those.
- The Playwright spec was not run, and there is no C# change, so no `dotnet test`.

PR #2696, opened 2026-08-19
Guard event-hook delivery; fix subscription leak and isLoading races

tjcouch-sil:
# Code Review Summary

**Branch**: `fix/event-data-hook-races`

**Base**: `origin/main`

**Date**: 2026-08-19

**Files changed**: 20 across 3 commits

> **Provenance note for the reviewer**: this review was run autonomously at the author's request. The
> "author" answers below are Claude's, not TJ Couch's — Claude both wrote the code and answered the
> interview. TJ has not personally walked through these changes. Treat the Suggested Review Focus as
> the list of things a human still needs to sign off on. Note that one Important finding from the
> automated pass was RETRACTED after TJ challenged it — see the Important section.

## Overview

Three related races in the renderer's event/data subscription hooks, plus the subscription leak
underneath them. `useEvent` and `useEventAsync` now guard delivery per subscription, so an emission
arriving from a superseded or unmounted subscription is dropped rather than delivered; a subscribe
promise that resolves after cleanup has its unsubscriber invoked immediately instead of leaking; the
unsubscriber runs at most once; and subscribe/unsubscribe failures are caught and warned rather than
escaping. `createUseDataHook` puts both of its state writes behind that guard and flips `isLoading`
true synchronously on a provider/selector change instead of from the outgoing subscription's async
teardown — which is the long-standing "change selector rapidly and the loading flag wedges" bug.

This work was developed on the `standard-view` feature branch and is being extracted as a standalone
PR against `main` so it can land independently. The extraction required rebuilding the checked-in
`platform-bible-react` `dist`, which surfaced a second, unrelated issue: `main`'s committed `dist`
was stale relative to `main`'s own source. That drift has since been resynced on `main` by #2690, so
the commit that carried it here is dropped - see "Rebase onto main" below.

## API Changes

- `lib/platform-bible-react` (`useEvent`): No signature or export change. TSDoc expanded to document
  per-subscription delivery guarding and unsubscriber-error handling. Runtime behavior changed.
- `lib/platform-bible-react` (`useEventAsync`): No signature or export change. TSDoc expanded to
  document per-subscription delivery guarding, the late-resolving-subscribe teardown, and
  warn-not-throw error handling. Implementation switched from `usePromise` to a self-contained
  `useEffect`.
- `lib/platform-bible-react/dist/index.d.ts` (published type surface): +14 lines, all TSDoc on
  `useEvent`/`useEventAsync`. Declaration lines are byte-identical — no exports added, modified, or
  removed.
- `lib/platform-bible-react/src/index.ts`: unchanged — export list identical.
- `lib/platform-bible-utils`: unchanged.
- `lib/papi-dts/papi.d.ts` / `@papi/` modules: unchanged, and no regeneration needed —
  `useEvent`/`useEventAsync` are not embedded in `papi.d.ts`, and the
  `createUseDataHook`/`UseDataHookGeneric`/`UseDataProxy` declarations it does embed
  (`papi.d.ts:8304`, `:8319`, `:8423`) were not touched (implementation bodies only).
- Extensions' type declaration files (`extensions/src/*/src/types/*.d.ts`): unchanged.

Behavioral (non-signature) changes to the public surface:

- `useEvent`: the handler is wrapped, so emissions arriving after cleanup are dropped; an
  unsubscriber that throws during cleanup is caught and `console.error`ed instead of propagating.
- `useEventAsync`: same delivery guard; a subscribe promise that resolves after cleanup now has its
  unsubscriber invoked immediately (leak fix); the unsubscriber runs at most once; subscribe and
  unsubscribe failures are caught and `console.error`ed.

None.

- [ ] ~~`use-event-async.hook.ts:71` — catching the subscribe rejection loses persisted-log capture,
  leaving a failure with nothing in the app log.~~ _(**Retracted — this finding was wrong.** It claimed
  `console.warn` is not routed into the logger. It is: `logger.service.ts:53` calls
  `log.initialize({ spyRendererConsole: true })`, and electron-log's `initializeSpyRendererConsole`
  hooks Electron's `console-message` WebContents event and feeds it to `logger.processMessage` with
  `levels = ['debug','info','warn','error']`. Electron 39 still passes the deprecated positional level
  (2 = warning), and the file transport level is `info` when packaged / `debug` in dev — both below
  warn — so these warnings ARE persisted in both builds. The original analysis grepped `src/` for
  source-level console patching and missed the mechanism, which lives inside electron-log behind one
  flag. The real delta is a severity change, error -> warn: previously the rejection reached
  `src/renderer/index.tsx:52` as `logger.error("Unhandled rejection in renderer, …")`. Offsetting that,
  the new message names the hook and the operation, and the `useEvent` case no longer throws inside
  React's commit phase. Credit to TJ for catching this.)_

- [x] **`lib/platform-bible-react/dist/*` — the dist rebuild swept in a user-visible change unrelated
  to this PR.** Exactly one Tailwind utility appears in the rebuilt bundles that is absent from
  `main`'s committed dist: `tw:active:not-aria-[haspopup]:transform-[translateY(1px)]`, from
  `src/components/shadcn-ui/button.tsx:22` — a file this PR does not modify. Verified by direct
  count: 0 occurrences in `origin/main`'s `dist/experimental.js`, 1 after rebuild, and 0 files in the
  `button.tsx` diff. So `main`'s dist was stale, and rebuilding it here would have shipped a button
  press-transform inside a hooks PR. _(fixed during review: split the rebuild into its own first
  commit, `chore(react): resync the checked-in platform-bible-react dist with its source`, built from
  unmodified main source. The hooks commit's dist delta is now only `index.*` — 5 files — with zero
  `experimental.*` or `resizable-*` churn, so each commit's dist change is attributable to its own
  source change. **Superseded by the rebase**: #2690 rebuilt main's dist, which now carries the button
  transform, so the resync commit is dropped and this PR's entire dist delta is `index.*` only.)_

[Author response: the log-capture finding was retracted after TJ challenged it — the analysis was
wrong about `console.warn` reaching the logger. The dist drift was fixed by splitting the commit.]

### Minor — Consider

- [x] **`use-event-async.hook.ts:39` — hand-typed unsubscriber re-derives existing exported types.**
  _(fixed during review: now `let unsubscribe: Unsubscriber | UnsubscriberAsync | undefined;`, using
  the types already exported from `platform-bible-utils`.)_
- [x] **`use-event-async.hook.ts:52-53` — no blank line after the guard clause in `unsubscribeOnce`,
  against the style guide's "Early Return (Guard Clauses)" entry.** _(fixed during review: blank line
  added.)_
- [x] **`use-event.hook.ts:40` vs `use-event-async.hook.ts:66-72` — asymmetric failure handling
  (async catches a failed subscribe; sync lets it escape) is unexplained.** _(fixed during review:
  added a comment at the sync subscribe explaining the asymmetry is deliberate — a throw there
  happens during the effect itself, where React's error boundaries can see it, whereas nothing can
  catch a throw from an effect cleanup.)_
- [x] **`use-event.hook.test.ts` / `use-event-async.hook.test.ts` — no test covers the `eventHandler`
  reference changing, even though both hooks' TSDoc names it as a supersede trigger.** _(fixed during
  review: added "does not deliver from a superseded subscription after the event handler changes" to
  both suites, asserting the old registration reaches neither handler and is unsubscribed exactly
  once.)_
- [x] **`use-event.hook.test.ts` / `use-event-async.hook.test.ts` — the `event === undefined`
  early-return path is untested, and `useEvent` changed it from `return () => {}` to
  `return undefined` in this PR.** _(fixed during review: added "subscribes nothing while the event is
  undefined, then subscribes once it arrives" to both suites, covering that tearing down a
  never-subscribed effect is a no-op.)_
- [x] **`create-use-data-hook.util.test.ts` — no test for `dataProvider === undefined` (the startup
  case the new `setIsLoading(true)` effect keys on) and none delivering a `PlatformError`.** _(fixed
  during review: added "stays loading and subscribes nothing until the data provider resolves" and
  "passes a delivered PlatformError through as the data and clears isLoading".)_
- [x] **`create-use-data-hook.util.test.ts:1-7` — the header comment recommended `npm run build`,
  while the documented fast path is `build:basic`.** _(fixed during review: corrected, and extended to
  note that nothing in CI enforces dist freshness.)_
- [ ] ~~`create-use-data-hook.util.ts:133-135` — the hook body early-returns before the remaining hook
  calls, and this change adds two more hooks below that return, so once
  `tooManyRendersErrorMessage.current` is set React throws "Rendered fewer hooks than expected".
  `react-hooks/rules-of-hooks` misses it because the hook body is an anonymous arrow returned from a
  factory.~~ _(Pre-existing on main — every existing `useState`/`useMemo`/`useEventAsync` call already
  sits after that return, so this change does not worsen it. Fixing it means restructuring the error
  branch to compute after all hooks run, which is unrelated to these races and would enlarge a
  split-off PR. Recorded as a follow-up.)_
- [ ] ~~`use-event.hook.ts:33-38` and `use-event-async.hook.ts:43-46` — the `isCancelled` +
  `guardedHandler` guard is duplicated verbatim across the two hooks.~~ _(Deliberate. It is ~5 lines
  each, and the obvious consolidation — implementing `useEvent` in terms of `useEventAsync` — is not
  equivalent: it would defer the synchronous unsubscribe to a microtask. A shared helper for 5 lines
  would add indirection without removing the real coupling.)_
- [ ] ~~`create-use-data-hook.util.ts:157-159` — resetting `isLoading` in an effect leaves one
  committed frame reporting `isLoading === false` alongside the previous selector's data.~~ _(Strictly
  better than the teardown-driven reset it replaces, and the alternative — React's "adjust state when
  a prop changes" pattern with a `previousSelector` state compared during render — is a larger
  behavioral change than this PR should carry. Recorded as a possible follow-up.)_
- [ ] ~~`create-use-data-hook.util.test.ts` resolves `platform-bible-react` through the checked-in
  `dist` rather than the library source, and nothing in CI rebuilds the library or fails on a dirty
  dist.~~ _(The coupling is real and now documented in the test header. The two suggested remedies — a
  `resolve.alias` in the root vitest config, and a CI step running `build:basic` then
  `git diff --exit-code lib/platform-bible-react/dist` — are both infrastructure changes outside this
  PR's scope. Strongly recommended as a follow-up: the CI check is exactly what would have caught the
  stale button class.)_

[Author response: seven minor findings were fixed in-review; four were dismissed with reasoning, three
of them recorded as follow-ups.]

#### Shared Regions Modified

None. All changed source files were grepped for `#region shared with` — no markers.

#### Extension Config Changes

None. No files under `extensions/` changed, and no build/tooling config was touched.

## Positive Observations

- **No API break.** Signatures, exports, and parameter types are byte-identical in `dist/index.d.ts`;
  the 14 added lines are purely TSDoc. Nothing requires deprecation or a migration note.
- **The leak fix is real and correctly targeted.** On unmount, `usePromise` set
  `promiseIsCurrent = false` and skipped `setValue`, so a late-resolving unsubscriber never reached
  state and the `useEffect([unsubscribe])` cleanup never ran it. `unsubscribeOnce` closes exactly that
  hole, and `hasUnsubscribed` keeps it idempotent across both entry paths.
- **The TSDoc claims are verifiable.** "An emitter walking a snapshot of its handler list" matches
  `PlatformEventEmitter.emitFn`, which clones `subscriptions` before iterating, so a handler removed
  mid-emit still gets called — precisely what the guard absorbs.
- **The `isLoading` effect ordering is correct.** React flushes all passive destroys before all
  passive creates, and within a fiber creates run in declaration order, so `setIsLoading(true)` is
  guaranteed to run before the `useEventAsync` effect can establish the new subscription.
- **`handleSubscriptionData` is `useCallback`-stable with `[]` deps**, honoring `useEventAsync`'s
  documented "MUST BE STABLE" contract, and keeps the `data` and `isLoading` writes together behind a
  single guard rather than splitting them across guarded and unguarded sides.
- **Effect-local state (not refs) for the guards** is the right call for StrictMode, and the
  StrictMode double-mount case is actually tested rather than merely asserted.
- **Test quality is high**: co-located, correctly named, wired into the right vitest projects, and the
  test doubles deliberately retain handlers after unsubscription, which is what makes the race
  simulations real rather than tautological. The subscribe-failure test asserts on
  `process.on('unhandledRejection')` in addition to the `console.warn` spy, correctly recognizing that
  an escaped rejection is the async shape of "throwing", and restores the listener in a `finally`.
- **Dead code was genuinely removed** — `noopUnsubscriber` and the `usePromise` import are gone, and
  `usePromise` itself is still exported and used by ~10 other consumers, so nothing was orphaned.
- **Red→green was proven, not asserted.** Reverting only the three source files with the tests in
  place produced 12 failures across the new tests; restoring them produced green.

## Interview Notes

**Stated purpose**: extract already-written `standard-view` work — the event-hook delivery guard, the
mid-flight subscribe leak fix, and the `createUseDataHook` `isLoading` races — into a standalone PR
against `main`, per TJ's long-standing strategy of "fix it on the feature branch, test it, split it
off to main, then rebase our stuff when it merges."

- *Why the dist rebuild is in the PR at all*: `create-use-data-hook.util.test.ts` imports
  `useEventAsync` by package name, which resolves through the workspace symlink to the checked-in
  `dist`. Without a rebuild those 8 renderer race tests would exercise the pre-fix `useEventAsync` and
  pass for the wrong reason. This was demonstrated accidentally during the review rework: after
  resetting to pristine main and rebuilding dist without the hook changes, the renderer suite went red
  with the exact original failure signature (`expected 'zombie exodus' to be 'leviticus 1'`).
- *Why the dist resync became its own commit*: so the button press-transform that main's stale dist
  was missing is attributable to a commit that says so, rather than hiding inside a hooks PR. (Moot
  after the rebase - main's dist now carries the transform, so the commit is dropped.)
- *On the catches added*: three, where there were none — `useEvent`'s cleanup, and `useEventAsync`'s
  subscribe and unsubscribe. All three swallow after logging rather than rethrowing (a new test
  asserts on `process.on('unhandledRejection')` that nothing escapes), because nothing can catch a
  throw from an effect cleanup and the async ones were fire-and-forget promises. The claimed
  observability cost turned out not to exist — see the retracted Important finding.

- *Why `useEvent` and `useEventAsync` are not consolidated*: implementing the sync hook in terms of
  the async one would defer the synchronous unsubscribe to a microtask, changing teardown ordering.

**Unresolved / needs a human:**

- Nothing is left open. The one Important finding that was open at first pass was retracted after TJ
  challenged its central claim and the analysis proved wrong; the reviewer should be aware that an
  automated pass asserted a logging regression that does not exist.
- **The author is Claude, not TJ.** TJ has not read this code or these findings. Every "author
  explanation" above is Claude explaining its own work, which is exactly the situation this command's
  interview step exists to detect. A reviewer should not read the explanations as independent
  corroboration.
- Manual/in-app verification has **not** been done for these hooks. The races are only reproducible
  via the injected orderings in the unit tests; the original plan noted the UI cannot reliably produce
  them. The one in-app symptom worth a human's eyes is the "flip projects/references quickly and watch
  for a stuck spinner" case.

## In-Review Quality Check

Changes were made during the review (one commit split, three source edits, six new tests, one lint
fix), so the checks were re-run:

- `npm run typecheck` — **exit 0**.
- `npm run lint` — initially **2 errors**, both `no-type-assertion/no-type-assertion` on the `as`
  casts in the two new "undefined event" tests. Fixed by hoisting a typed `initialProps` local instead
  of asserting, rather than suppressing the rule. Re-run clean.
- `platform-bible-react` unit project — **40 files / 401 tests passed** (up from 397; +4 new tests).
- `create-use-data-hook.util.test.ts` — **10 tests passed** (up from 8).
- Note on the earlier full-suite run: the `storybook (chromium)` project sheds 2-3 files to
  `Test timed out in 15000ms` when it runs alongside the unit project on this machine. Confirmed
  pre-existing load flake, not this change — the failing set differs run to run, none of those files
  reference the changed hooks, and run on its own the project is **89 files / 541 tests passed,
  identical with the change applied and with it stashed**.
- Two environment prerequisites were needed in a fresh worktree and are worth knowing:
  `release/app/buildInfo.json` is gitignored and generated by
  `.erb/scripts/generate-dev-build-info.ts` (typecheck and lint both fail without it), and eslint
  caches the resulting unresolved-import error, so `.eslintcache` must be removed after generating it.
- No C# changed, so `dotnet test` was not run.

## Reviewer Feedback (@lyonsil, 2026-08-20)

Five AI-assisted notes, posted as non-blocking. All five were verified independently before acting.

- [x] **Subscribe `catch` demoted the diagnostic from error to warn.** Correct and a real regression
  from this PR. _(fixed: all three catches now use `console.error`, restoring the pre-PR severity and
  matching the library's own convention. One detail in the note did not hold — the old path had no
  stack either: `src/renderer/index.tsx:52` logs `getErrorMessage(event.reason)`, which is
  `toErrorWithMessage(error).message`, message-only. The new call passes the Error object itself.)_
- [x] **`hasUnsubscribed = true` before the await means a failed unsubscribe is never retried, but the
  comment reads as idempotence.** Correct. _(fixed: comment reworded to state one attempt, no retry,
  and that a network event's remote registration therefore stays live for the session; TSDoc updated
  to carry the same contract on the public surface. Behavior unchanged — there is no caller to retry
  from.)_
- [x] **The `undefined`-event test never asserted that nothing subscribed, and its comment named a path
  it did not take.** Both correct. _(fixed: `createControlledEvent` gained
  `getSubscribeCount: () => registrations.length`, matching the async twin; the test asserts 0 before
  the rerender and 1 after, and the teardown claim moved onto the `rerender`, which is where React
  runs the never-subscribed effect's cleanup.)_
- [x] **Flake risk: the subscribe-failure test's `unhandledRejection` listener and `console` spy are
  both worker-global, and it asserted on counts.** Correct. _(fixed: asserts
  `not.toContain(subscribeError)` and `toHaveBeenCalledWith(msg, subscribeError)` against a hoisted
  error instead of counting. Same treatment applied to the unsubscribe-failure twin and the `useEvent`
  equivalent, which shared the `toHaveBeenCalledOnce()` pattern.)_
- [ ] ~~Pre-existing: `newPlatformError(tooManyRendersErrorMessage)` passes the ref rather than
  `.current`, yielding `{ message: '', cause: ref }`.~~ _(Analysis confirmed, but **deliberately not
  fixed** — executing the path showed the line is unreachable. The `useMemo` at
  `create-use-data-hook.util.ts:134` is a conditional hook: on the render after the threshold trips it
  lands on the slot the long path used for a `useRef`, and React throws in `areHookInputsEqual` before
  the `useMemo` factory — and therefore `newPlatformError` — ever runs. Reproduced by rerendering past
  the 100-in-1s threshold; `result.current[0]` stayed `'default'` throughout. Removing the `useMemo`
  would not help either: the early return still skips every hook below it. A one-character fix to an
  unreachable line would be noise. See Follow-ups.)_

### Second round (@lyonsil, 2026-08-20)

- [x] **The symmetry fix landed on the sync suite only; the async twin still had the second half.**
  `use-event-async.hook.test.ts` had the teardown comment sitting above a bare `rerender` with nothing
  asserted on it, and no `getSubscribeCount()` check after. Correct. _(fixed in `d45074c`: the comment
  now sits on an `expect(() => rerender(...)).not.toThrow()`, with the subscribe count asserted `0`
  before and `1` after. Both suites are the same shape.)_
- [x] **"Did you already open an issue?"** on the too-many-renders early return. _(Answered: no new
  issue needed - [PT-3594](https://paratextstudio.atlassian.net/browse/PT-3594) already covers it. Tom
  Bogle filed it in Oct 2025 with a stack landing on `create-use-data-hook.util.ts:134`; it captured
  the symptom without the diagnosis, so the root cause - the conditional `useMemo`, the throw inside
  `areHookInputsEqual`, and the `.current` bug on the same line - has been added there. Follow-up 2
  below is that ticket.)_

## Rebase onto main (@lyonsil, 2026-08-20)

Rebased onto `c0583c38d4d` while TJ is out, and rebuilt the dist on top of the rebased source.

- **The dist-resync commit is dropped.** #2690 rebuilt main's `dist`, which now carries the button's
  active-state transform, so the drift this PR originally isolated no longer exists. Confirmed by
  running `build:basic` from pristine `origin/main` in a clean worktree: the output is byte-identical
  to main's committed `dist`.
- **Three commits; the source is byte-identical to the pre-rebase head** apart from the one test
  change above (verified per file by blob hash). `dist` was rebuilt on top of each of the two
  `fix(...)` commits, so each commit's dist delta is still attributable to its own source change: the
  hooks commit is `index.*` only, and the error-level commit is a 3-line body change plus the TSDoc.
- **The dist build is not byte-deterministic.** Two builds of identical source can emit the injected
  component CSS blocks in different orders (the `.banded-row` rules moved relative to the
  footnote-editor height comment between runs), producing an ~8-line churn across `index.js`,
  `index.cjs`, `experimental.js` and `experimental.cjs` that has nothing to do with the source change.
  The dist committed here is from a run whose ordering matches main's, so the diff stays attributable
  - but see follow-up 3, where it matters.

Verification after the rebase: `npm run typecheck` **exit 0**; `npm run lint` **exit 0** (4
pre-existing warnings, all in unrelated extension files); core vitest **147 files / 1831 tests
passed**; `platform-bible-react` unit project **41 files / 410 tests passed**, including both hook
suites. The full workspace run had one failure - `storybook (chromium)`
`tab-navigation-content-search.stories.tsx > Default`, `Test timed out in 15000ms` - which is the
pre-existing load flake documented above; the file does not reference the changed hooks, and it passes
on its own (**1 file / 6 tests**). No C# changed, so `dotnet test` was not run.

## Follow-ups (not in this PR)

1. **Surface subscribe failure to the caller.** A failed `subscribe` leaves `useData` pinned at
   `isLoading === true` showing `defaultValue`, with no signal a consumer can react to and no retry.
   Pre-existing and unchanged here, but it is the same class of symptom this PR targets. Closing it
   means an API decision (an `onError` parameter on `useEventAsync`, or similar) that should not ride
   along in an extraction PR.
2. **Restructure the too-many-renders early return** in `create-use-data-hook.util.ts:133-135` so the
   threshold check happens after all hooks have run. Today the branch throws instead of returning its
   error tuple, so the diagnostic is dead. Pass `.current` to `newPlatformError` while in there.
3. **Add a CI dist-freshness gate.** Nothing rebuilds `platform-bible-react` or fails on a dirty
   `dist`, which is how the stale button-transform class in main's committed bundle went unnoticed
   until this PR rebuilt it. `build:basic` + `git diff --exit-code lib/platform-bible-react/dist`, or
   a `resolve.alias` pointing the package name at source in the root vitest config. Note the caveat
   found during the rebase: the build is not byte-deterministic (injected component CSS blocks can
   change order between runs of identical source), so the `git diff --exit-code` form would flake as
   written. The `resolve.alias` form does not have that problem.

## Suggested Review Focus

- [ ] **Confirm the severity choice**: subscription failures now log at `error` (restored from the
  `warn` this PR briefly introduced). Both the original unhandled-rejection path and the new one log
  message-only, no stack.
- [x] ~~**Confirm the dist-resync commit should ride along** rather than going in as its own PR.~~
  _(Moot: #2690 resynced main's dist, so the resync commit is dropped in the rebase and the button
  press-transform is already on main.)_
- [ ] **Agree the CI gap is worth closing**: nothing rebuilds `platform-bible-react` or fails on a
  dirty `dist`, which is how the stale button class survived. A `build:basic` +
  `git diff --exit-code lib/platform-bible-react/dist` step, or a `resolve.alias` pointing the package
  name at source in the root vitest config, would close it.
- [ ] **Walk through the `isLoading` effect ordering with TJ**, since he has not reviewed it: the
  claim is that React's passive-destroy-before-passive-create ordering guarantees
  `setIsLoading(true)` lands before the resubscribe. It is load-bearing for the whole fix.
- [ ] **Decide on the two recorded follow-ups**: the rules-of-hooks hazard from the early return in
  `create-use-data-hook.util.ts:133`, and the one committed frame where `isLoading === false` sits
  alongside the previous selector's data.
- [ ] **In-app sanity check**: flip projects/references rapidly and confirm no stuck spinner and no
  flash of the previous selector's data.

AI-assisted.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2696)

[PT-3594]: https://paratextstudio.atlassian.net/browse/PT-3594?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/platform-bible-react/src/hooks/use-event-async.hook.ts
```diff
@@ -24,30 +35,55 @@ export const useEventAsync = <T>(
...
+      })();
+    };
+
+    // Subscribe. Wrapped in Promise.resolve to allow synchronous events to be used as well
+    (async () => {
+      try {
+        unsubscribe = await Promise.resolve(event(guardedHandler));
+        // Cleanup already ran while the subscribe was in flight, so tear down right away
+        if (isCancelled) unsubscribeOnce();
+      } catch (error) {
```
lyonsil, line 80: The subscribe `catch` swallows the failure with nothing a consumer can observe. Worth splitting what is new here from what is not:

- **Pre-existing:** a failed subscribe already wedged `isLoading` at `true`, since `createUseDataHook` only clears it on delivery. Unchanged by this PR.
- **New:** before this change there was no `catch` on this path at all (`usePromise` does not catch either, `use-promise.hook.ts:68`), so the rejection escaped and `src/renderer/index.tsx:52` logged it at **error** level with a stack. It is now a `console.warn` with no stack.

Net effect in `createUseDataHook`: the spinner never stops, there is no retry, and the diagnostic got quieter than it was. Reasonable to defer, but the stuck-loading path is the same class of symptom this PR set out to kill, so it seems worth a follow-up ticket rather than losing track of it.

lib/platform-bible-react/src/hooks/use-event-async.hook.ts
```diff
@@ -24,30 +35,55 @@ export const useEventAsync = <T>(
...
+    // Emissions from this subscription stop reaching the handler the moment cleanup runs, even if
+    // the underlying event still fires it (e.g. an emission already in flight during teardown)
+    const guardedHandler: PlatformEventHandler<T> = (eventData) => {
+      if (!isCancelled) eventHandler(eventData);
+    };
+
+    // Runs the unsubscriber at most once no matter which path gets there first (cleanup, or the
+    // subscribe promise resolving after cleanup). Fire-and-forget: nothing can await an effect
+    // cleanup, so failures are logged instead of thrown
+    const unsubscribeOnce = () => {
```
lyonsil, line 60: `hasUnsubscribed = true` is set before the await, so a rejected unsubscribe is never retried. `isCancelled` mutes local delivery, but for a network event the remote registration then stays live for the rest of the session with nothing left to clean it up. (Not a regression: the previous implementation had no catch here at all.)

Given the fire-and-forget contract there may be no good retry point, so this may well be the right trade. The comment above it reads as idempotence though, which is a stronger promise than the code makes. Suggest wording it as "one attempt; a failed unsubscribe is not retried" so the next reader does not have to derive that from the flag ordering.

lib/platform-bible-react/src/hooks/use-event.hook.test.ts
```diff
@@ -0,0 +1,135 @@
...
+    controlled.emitTo(0, 'stale');
+    expect(firstHandler).not.toHaveBeenCalled();
+    expect(secondHandler).not.toHaveBeenCalled();
+    expect(controlled.getUnsubscribeCallCount(0)).toBe(1);
+
+    act(() => controlled.emitTo(1, 'fresh'));
+    expect(secondHandler).toHaveBeenCalledExactlyOnceWith('fresh');
+  });
+
+  it('subscribes nothing while the event is undefined, then subscribes once it arrives', () => {
```
lyonsil, line 100: Two small things in this test:

1. **It never asserts that nothing subscribed.** `createControlledEvent` has no subscribe counter, so a regression that subscribed while `event` was `undefined` would still pass here. The async twin does check it: `use-event-async.hook.test.ts:155` asserts `controlled.getSubscribeCount()` is `0`. Adding `getSubscribeCount: () => registrations.length` to the helper's return (alongside `getUnsubscribeCallCount` on line 27) and asserting it before the `rerender` would close the gap and make the two suites symmetric.

2. **The comment names a path the test does not take.** "Tearing down an effect that never subscribed must be a no-op" (lines 108-109) sits above the `rerender` on line 110, but the only teardown assertion is `expect(() => unmount()).not.toThrow()` on line 114, which runs after a real subscription exists.

lib/platform-bible-react/src/hooks/use-event-async.hook.test.ts
```diff
@@ -0,0 +1,265 @@
...
+
+  it('warns instead of throwing when subscribing fails', async () => {
+    const warn = vi.spyOn(console, 'warn').mockImplementation(() => {});
+    // A subscribe failure surfaces asynchronously, so escaping as an unhandled rejection is the
+    // async shape of "throwing" — nothing in an effect could ever catch it.
+    const escapedRejections: unknown[] = [];
+    const recordRejection = (reason: unknown) => {
+      escapedRejections.push(reason);
+    };
+    process.on('unhandledRejection', recordRejection);
```
lyonsil, line 205: Flake risk rather than an observed failure - it passed on repeated local runs.

Both of this test's mechanisms are process-global within the vitest worker: the `unhandledRejection` listener collects rejections from anything else running during the window, and the `expect(warn).toHaveBeenCalledOnce()` spy counts any `console.warn` emitted while it is installed. If another test file sharing the worker warns or rejects in that window, this fails for reasons unrelated to the hook.

Filtering `escapedRejections` down to the expected error, and relaxing the warn assertion to `toHaveBeenCalledWith(...)`, would make it self-contained.

lyonsil (commented):
AI-assisted review notes on five specific items in this branch, posted for consideration rather than as blockers. Four are inline below; the fifth is a file-level note on `create-use-data-hook.util.ts`, covering a pre-existing line that falls outside this PR's diff.

src/renderer/hooks/hook-generators/create-use-data-hook.util.ts
lyonsil, line 1: File-level comment because the line is outside this PR's diff - **pre-existing and untouched here**, flagging it only because this branch is already in the file.

Line 134 passes the ref object rather than its `.current`:

```ts
return [useMemo(() => newPlatformError(tooManyRendersErrorMessage), []), undefined, false];
```

`newPlatformError` (`lib/platform-bible-utils/src/platform-error.ts:141`) takes the string path only when `isString(error)` (line 148). A `useRef` object is not a string and has no `message` property, so it falls through to the final branch (line 172) and the consumer gets `{ message: '', cause: <the ref object> }` - an empty message, with the "Data of type X was updated N times..." diagnostic reachable only via `cause.current`, one line after `logger.warn` printed it correctly.

One-character fix (`tooManyRendersErrorMessage.current`) if you want it in this branch; otherwise worth a ticket so the diagnostic is not dead weight.

lib/platform-bible-react/src/hooks/use-event-async.hook.ts
tjcouch-sil: Thanks — the pre-existing/new split is exactly right, and the severity drop is a real regression I introduced. Fixed: all three catches now use `console.error` rather than `console.warn`, which restores the old error-level severity and matches this library's existing convention (`editor-utils.ts:87,130`, `filter-dropdown.component.tsx:97`). Tests and TSDoc updated to match.

One correction on the stack, though — the old path didn't carry one either. `src/renderer/index.tsx:52` is:

```ts
logger.error(`Unhandled rejection in renderer, '${getErrorMessage(event.reason)}'`);
```

and `getErrorMessage` is `toErrorWithMessage(error).message` (`platform-bible-utils/src/util.ts:221-223`) — message only, no stack. The sibling `'error'` listener on line 48 interpolates `${error}`, which stringifies to `Error: <message>`, also without a stack. So the new `console.error('…', error)` passes the Error object itself, which is if anything *more* than before.

Worth noting for the record, since it came up in the earlier self-review: renderer `console.*` does reach the app log. `logger.service.ts:53` calls `log.initialize({ spyRendererConsole: true })`, and electron-log hooks Electron's `console-message` WebContents event into `logger.processMessage` with `levels = ['debug','info','warn','error']`. File transport level is `info` packaged / `debug` dev, so both warn and error are persisted.

Agreed on the follow-up ticket for the stuck-loading path — it's pre-existing, it's genuinely the same class of symptom this PR targets, and closing it properly means giving the hook a way to surface subscribe failure to its caller (an `onError` parameter or similar), which is an API decision that shouldn't ride along in an extraction PR. Flagged in the PR description so it doesn't get lost.

lib/platform-bible-react/src/hooks/use-event-async.hook.ts
tjcouch-sil: Good catch, and agreed on both counts — the ordering is deliberate but the comment was writing a cheque the code doesn't cash. Reworded:

```ts
// Makes ONE attempt to unsubscribe, from whichever path gets there first (cleanup, or the
// subscribe promise resolving after cleanup). The flag is set before the await, so a failed
// unsubscribe is NOT retried - for a network event that leaves the remote registration live for
// the rest of the session, which is the accepted cost of a fire-and-forget teardown: nothing
// can await an effect cleanup, so there is no caller to retry or surface the failure to.
// Delivery is still muted locally by `isCancelled` either way.
```

Also updated the hook's TSDoc to say "A failed unsubscribe is logged, not retried" so the contract is visible from the public surface rather than only at the call site. Kept the single-attempt behavior for the reason you identified — there's no caller to retry from — but the leaked-remote-registration consequence is now stated rather than derivable from flag ordering.

lib/platform-bible-react/src/hooks/use-event.hook.test.ts
tjcouch-sil: Both correct, both fixed.

1. Added `getSubscribeCount: () => registrations.length` to `createControlledEvent`'s return, so the two suites are now symmetric, and the test asserts it's `0` before the `rerender` and `1` after. Without that a regression that subscribed while `event` was `undefined` really would have sailed through.

2. Moved the claim to the assertion that actually tests it. The teardown of the never-subscribed effect happens *during* the `rerender` (React runs the old cleanup before the new effect), so the comment now sits there and the `rerender` itself is wrapped in `expect(...).not.toThrow()`:

```ts
expect(controlled.getSubscribeCount()).toBe(0);

// Re-running the effect tears down the never-subscribed one first; that must be a no-op rather
// than a call through a missing unsubscriber
expect(() => rerender({ event: controlled.event })).not.toThrow();
expect(controlled.getSubscribeCount()).toBe(1);
```

lib/platform-bible-react/src/hooks/use-event-async.hook.test.ts
tjcouch-sil: Agreed — both mechanisms are worker-global and the assertions were counting rather than matching. Made it self-contained exactly as you suggested:

- The rejection assertion is now `expect(escapedRejections).not.toContain(subscribeError)` against a hoisted `const subscribeError = new Error('subscribe failed')`, so a neighbouring file's rejection landing in the same listener is ignored.
- The log assertion is now `toHaveBeenCalledWith('useEventAsync: error while subscribing to event', subscribeError)` instead of `toHaveBeenCalledOnce()`.

Applied the same treatment to the unsubscriber-failure twin below it, which had the identical `toHaveBeenCalledOnce()` pattern, and to the `useEvent` sync equivalent. Both spies also moved from `console.warn` to `console.error` per your first comment.

src/renderer/hooks/hook-generators/create-use-data-hook.util.ts
tjcouch-sil: Verified your analysis and it's correct — a `useRef` object isn't falsy, isn't a string, and has no `message` property, so it falls to `newPlatformError`'s final branch and yields `{ message: '', cause: <ref> }`.

I dug in to write a test for the fix, and found something that changes the recommendation: **that line can't execute at all.** The `useMemo` on line 134 is a conditional hook — on the render after the threshold trips, the short path's `useMemo` lands on the hook slot the long path used for a `useRef`, and React throws while comparing deps, before the `useMemo` factory (and therefore `newPlatformError`) ever runs.

Reproduced by rerendering the hook past the 100-renders-in-1s threshold:

```
TypeError: Cannot read properties of undefined (reading 'length')
    at areHookInputsEqual (react-dom-client.development.js:7611:36)
    at updateMemo (react-dom-client.development.js:8793:28)
    at Object.useMemo (react-dom-client.development.js:26484:18)
    at Proxy.<anonymous> (src/renderer/hooks/hook-generators/create-use-data-hook.util.ts:134:17)
```

`result.current[0]` stayed `'default'` throughout — the error tuple is never returned.

So the one-character fix would be correct but unobservable, and removing the `useMemo` wouldn't help either: the early return still skips every hook below it, which React rejects as "Rendered fewer hooks than expected" regardless. Making the diagnostic work means restructuring so the threshold check happens *after* all hooks have run rather than early-returning past them.

Given that's a real change to a hot path and unrelated to this PR's races, I've left the code alone rather than shipping a fix to an unreachable line. Recommend one ticket covering both: restructure the early return, and pass `.current` while you're in there. Noted in the PR description.

(Both automated passes on my side flagged the conditional-hook hazard as theoretical and pre-existing; your comment is what prompted actually executing it.)

lib/platform-bible-react/src/hooks/use-event.hook.test.ts
lyonsil: Confirmed on the sync file — `getSubscribeCount()` asserted `0` before the `rerender` and `1` after, and the teardown claim now sits on the `rerender` that actually performs it. Both halves closed there.

The async twin still has the second half, though. `use-event-async.hook.test.ts:157-159`:

```ts
// Tearing down an effect that never subscribed must be a no-op, not a call through a missing
// unsubscriber
rerender({ event: controlled.event });
```

Same shape the sync test had before the fix: the comment sits above a bare `rerender` with nothing asserted on it, and the only `not.toThrow()` in the test is on the `unmount()` at line 164, which runs after a real subscription exists. The `getSubscribeCount()` is `1` check after the rerender is missing too.

Since you did the cross-file symmetry pass for the flake note, worth the same three lines here so the two suites end up in the same shape.

src/renderer/hooks/hook-generators/create-use-data-hook.util.ts
lyonsil: This is fair. The main purpose of the rate limit still applies (i.e., it breaks the app loudly). Did you already open an issue?

lib/platform-bible-react/src/hooks/use-event.hook.test.ts
lyonsil: Done - applied in d45074c while TJ is out, so this doesn't wait on him.

`use-event-async.hook.test.ts` now has the same three lines the sync suite got:

```ts
expect(controlled.getSubscribeCount()).toBe(0);

// Re-running the effect tears down the never-subscribed one first; that must be a no-op rather
// than a call through a missing unsubscriber
expect(() => rerender({ event: controlled.event })).not.toThrow();
expect(controlled.getSubscribeCount()).toBe(1);

await act(async () => controlled.resolveSubscribe(0));
```

Same comment wording, the `not.toThrow()` sits on the `rerender` that actually performs the never-subscribed teardown, and the subscribe count is asserted on both sides of it. Both suites are now the same shape.

`use-event-async.hook.test.ts` is 10 tests, `use-event.hook.test.ts` is 7; the platform-bible-react unit project is 41 files / 410 tests green.

src/renderer/hooks/hook-generators/create-use-data-hook.util.ts
lyonsil: No new issue - **[PT-3594](https://paratextstudio.atlassian.net/browse/PT-3594)** already covers this. Tom filed it in Oct 2025 ("Too many renders can cause rendering error in useLocalizedStrings and leave webview blank") and the stack he attached lands on `create-use-data-hook.util.ts:134`, i.e. this exact line. It captured the symptom without the diagnosis, so I've added the root cause there: the early return makes the `useMemo` a conditional hook, React throws in `areHookInputsEqual` before the factory runs, and the `.current` bug is on the same line to fix at the same time. It's still 💡 Backlog Candidate.

Agreed the rate limit still does its job - it breaks loudly. The catch is that it breaks via a hook-order crash rather than the diagnostic it wrote, so what a user reports is a white screen and a `TypeError` in React internals, with the "Data of type X was updated N times" line sitting one entry earlier in the log where nobody looks. Fixing that is the point of the ticket, not the rate limit itself.

Nothing left open here for this PR.

lyonsil (approved):
:lgtm:

> @lyonsil partially reviewed 19 files and all commit messages, made 1 comment, and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2696#-:-P-W0BM4CMtde40QVk9e:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2697, opened 2026-08-19
PT-4347: Panels never show a premature empty state while resolving

captaincrazybro:
# Code Review Summary

**Branch**: worktree-pt-4347-resource-panel-loading-state

**Base**: origin/main

**Date**: 2026-08-19 (updated 2026-08-20 after the PR review round)

**Files changed**: 21

## Overview

PT-4347: resource panels showed "No … selected" with a Pick button while a correctly-configured
resource was still loading. The ticket named a merged loading/empty conditional in
`model-text-panel.component.tsx` as the root cause. Investigation found that conditional is the
_symptom site_ — the defect is in `useEffectiveResourceReferenceList`, which returned
`[list | undefined, boolean]` where the list accounted for two async sources (the project-level
setting and a user-level PDP subscription) but the boolean reported only the first. The user layer
needs the PDP to resolve, then an explicit subscribe, then a first delivery — strictly more hops —
so `[undefined, false]` was the normal interleaving on essentially every mount, not a narrow race.
The ticket's suggested fix (split the guard) would not have worked: in that window there is no
honest loading signal to split on.

The fix makes readiness a first-class, data-derived signal evaluated before any empty/not-found
branch. The hook returns a discriminated `loading | error | ready` status; `getResourcePanelReadiness`
maps that plus catalog arrival to `loading | error | catalogError | empty | configured`; and
`PanelReadinessView` renders those states. Both panels route through them, so they cannot drift
apart on the question that caused the bug. `useBufferedLayoutSetting` no longer latches a read
error, and `useDblResourceCatalog` centralises the previously duplicated catalog fetch. Recorded as
ADR-0027.

## API Changes

**Public API surfaces: None.** No changes to `lib/platform-bible-react/`,
`lib/platform-bible-utils/`, `lib/papi-dts/papi.d.ts`, or any extension type-declaration file.

Extension-internal module surfaces that changed (not consumed outside
`extensions/src/platform-scripture-editor/`):

- `use-effective-resource-reference-list.hook.ts`: return type changed from
  `[EffectiveResourceReferenceList | undefined, boolean]` to the new exported discriminated union
  `EffectiveResourceReferenceListState` (`loading` | `error` | `ready` + `list`). Both call sites
  updated.
- `use-buffered-layout-setting.hook.ts`: return tuple gained a third element
  `PlatformError | undefined`. Behaviour changed — a `PlatformError` is no longer applied to the
  held copy, and the error is reported only while no readable value has ever been applied.
- `install-state-views.component.tsx` → **renamed** `panel-state-views.component.tsx`.
  `InstallingView` and a bare loading view merged into `LoadingView({ label })` (label required).
  `InstallFailedView` → **renamed** `RetryableErrorView`, since it now serves catalog-fetch failures
  as well as installs; both it and `LoadingView` carry a live-region role.
- `resource-panel-readiness.utils.ts` (new): exports `ResourcePanelReadiness`,
  `ResourcePanelReadinessInput`, `getResourcePanelReadiness()` (options object).
- `panel-readiness-view.component.tsx` (new): exports `PanelReadinessView`.
- `use-dbl-resource-catalog.hook.ts` (new): exports `DblResourceCatalog`, `useDblResourceCatalog()`.
- `model-text-panel.component.tsx`: `ModelTextPanelProps` — removed `effectiveModelTexts`,
  `isEffectiveModelTextsLoading`, `hasModelTextsError`, `isLoadingResources`, `areResourcesReady`;
  added `modelTextsState: EffectiveResourceReferenceListState` (the whole discriminated state, so
  narrowing survives the boundary), `isCatalogReady`, `hasCatalogError`, `onRetryCatalog`. All call
  sites (web view, stories, tests) updated.
- `use-text-collection-sources.hook.ts`: now consumes the buffered setting's error channel.

None.

- [x] **The settings-error state's "Try again" button had no observable effect** (flagged
      independently by the UX and API passes). `status: 'error'` comes only from the project-setting
      channel, but `retry` bumped an epoch that only the _user_ PDP subscription depended on.
      _(fixed during review: verified no honest retry is implementable — re-arming is a no-op since
      the buffer never disarms during an error, and user-layer errors are swallowed to
      `DEFAULT_LIST` so they can never produce `status: 'error'`. Removed the control and the
      `retryEpoch` machinery; the message now carries the recovery expectation.)_
- [x] **A read failure arriving after a successful read discarded good held content.**
      `settingError` derived from the live `rawSetting` regardless of whether a real value had been
      applied, so a later `PlatformError` replaced working panel content with an error — the mirror
      of the sticky-error bug being fixed. _(fixed during review: `hasAppliedRealValue` gate; the
      error is reported only while nothing readable has ever landed.)_
- [x] **A failed DBL-catalog fetch became an unrecoverable spinner.** `usePromise` has no `.catch`,
      so a rejected `getCachedResources` never reaches `setIsLoading(false)`. On `main` this path
      fell through to the wrong-but-actionable empty prompt. _(fixed during review: new
      `useDblResourceCatalog` catches the rejection and reports `hasCatalogError`; new `catalogError`
      readiness state with a working retry, since a catalog fetch — unlike a setting read — can
      genuinely be re-driven.)_
- [x] **Two divergent implementations of the same readiness state machine.** The Resource panel
      routed through the shared helper while the Model Text panel re-derived the same states inline
      — the exact bug class ADR-0027 names. _(fixed during review: `ModelTextPanelProps` now takes
      `modelTextsStatus` instead of separate loading/error booleans, and the panel routes through
      `getResourcePanelReadiness` + `PanelReadinessView`. Five inline branches collapsed into one.)_
- [x] **The full-panel spinner block was written 4–5 times.** _(fixed during review: single
      `LoadingView` in `panel-state-views.component.tsx`; all call sites route through it.)_
- [x] **`useTextCollectionSources` was broken by the un-latching change.** Because the held value is
      now the placeholder rather than the error, its `isPlatformError(adminReferenced)` check no
      longer fired, and `sources` became _defined with an empty admin list_ where it previously
      stayed `undefined` — the same "unreadable setting rendered as nothing configured" bug, newly
      introduced in the Text Collection grid. _(fixed during review: consumes the error channel;
      regression test pins the realistic loading→error ordering, which the pre-existing test missed
      because it injected the error on the very first render.)_
- [x] **`PanelReadinessView` had no story**, despite being the only place the Resource panel's front
      states are viewable. _(fixed during review: four-state story file added, plus a `CatalogError`
      story on the Model Text panel.)_

### Minor — Consider

- [x] `install-state-views.component.tsx` name no longer matched its contents _(fixed during review:
      `git mv` to `panel-state-views.component.tsx`)_
- [x] `listStatus` hand-retyped the hook's union _(fixed during review: derived from
      `EffectiveResourceReferenceListState['status']`)_
- [x] `export default` on the utils module was inconsistent with its twelve siblings _(fixed during
      review: removed)_
- [x] Four positional args with two transposable numbers _(fixed during review: options object)_
- [x] `areResourcesReady` collided with the other meaning of "resources" in these files _(fixed
      during review: `isCatalogReady`)_
- [x] The catalog fetch was duplicated verbatim across both web views _(fixed during review:
      `useDblResourceCatalog`)_
- [x] The `!isLoadingResources && resourcesPossiblyUndefined !== undefined` predicate was duplicated
      _(fixed during review: owned by `useDblResourceCatalog`)_
- [x] ADR-0027 claimed `getResourcePanelReadiness` was the front of _both_ panels when only the
      Resource panel used it _(fixed during review: now true of both; ADR also corrected in two other
      places where it had drifted)_
- [x] Leftover `?.` in migrated test assertions _(fixed during review — but the stated rationale was
      wrong: the pass claimed `?.` "would silently pass `undefined` into `toEqual` rather than
      failing". It would not; `expect(undefined).toEqual({...})` fails. It was dead syntax, not a
      latent defect.)_
- [x] Blank line after early return _(fixed during review in the guards authored here; two
      pre-existing violations left alone rather than adding unrelated diff noise)_
- [x] `settingError` was never asserted directly in the buffered hook's own tests _(fixed during
      review: positive assertion added and mutation-checked)_
- [x] Unlabelled spinners _(fixed during review within this extension: `LoadingView` requires an
      already-localized label, wrapper carries `role="status"`, icon is `aria-hidden`; new
      `%webView_*_loading%` keys in en + es)_
- [x] Error strings named an internal "settings" concept _(fixed during review: rewritten in the
      user's vocabulary and to carry the recovery expectation)_
- [ ] ~~The `configured` branch in `PanelReadinessView` is unreachable from its production
      callers~~ _(one line that keeps the component total over its input type; removing it requires a
      partial function or a cast)_
- [ ] ~~`useBufferedLayoutSetting`'s doc mentions a first-read-error path that is arguably
      unreachable through `useProjectSetting`~~ _(wording already softened; the defensive branch is
      one cheap line)_
- [ ] **The `getResourcePanelReadiness` call site is untested** — the unit tests exercise the pure
      function only. Closing this needs a `ResourceTextPanel` extraction, which is out of scope. Risk
      is materially reduced by the switch to a named options object.

## Findings — reviewer round (@jolierabideau, PR #2697)

A second round of findings from the PR review, tracked with the same states. Full text and
per-item assessments in `triage-brief-2697.md`. Nothing here was silently dropped.

### Blocking — addressed

- [x] **Stale-write race in `useDblResourceCatalog`** _(fixed: `setHasCatalogError` ran inside the
      `usePromise` factory with no currency guard, so a superseded fetch resolving late could clear a
      real error and drive readiness to `'empty'` — the exact bug this PR removes. Reproduced with a
      failing test first, then guarded with a `fetchGenerationRef`; reproduction retained.)_
- [x] **Stale TSDoc contradicting the code** _(fixed: `panel-readiness-view.component.tsx` still said
      the Model Text panel rendered these states inline, written before both panels were converged.)_

### Important — addressed

- [x] **The panel test for the headline AC could not fail on its own mutation** _(fixed: replayed the
      reviewer's mutation to confirm — all 14 model-panel tests passed before. Added the empty-prompt
      absence and `getByRole('status')`; the mutation now fails it.)_
- [x] **The union's benefit was discarded at the prop boundary** _(fixed: `ModelTextPanelProps` now
      takes the whole `modelTextsState` rather than an `undefined`-able list plus a status string.
      Verified by type probe — `{ status: 'ready' }` without a list is `TS2322`,
      `{ status: 'loading', list }` is `TS2353`. Migrating the tests removed one that had been
      asserting an illegal combination.)_
- [x] **`isCatalogReady` was `true` after a failed fetch** _(fixed: folded in `!hasCatalogError`; the
      catch resolves to `[]`, which made the value defined and contradicted its own TSDoc.)_
- [x] **Rejected `subscribeUser*` hung on `loading` forever** _(fixed: falls back to the project-level
      items, which the hook's doc already promised. Failing test first.)_
- [x] **No live region on either failure view** _(fixed: `role="alert"` on both; only `LoadingView`
      had `role="status"`, so the flip out of loading announced nothing.)_
- [x] **`InstallFailedView` name/doc wrong for its new use** _(fixed: renamed `RetryableErrorView`; it
      now serves catalog-fetch failures too.)_
- [x] **Story docs called the settings error "recoverable"** _(fixed; the ADR's point is that it is
      not recoverable from the panel.)_
- [x] **Localization key broke alphabetical order** _(fixed — with a correction: the review's target
      was off. `bibleTexts` < `catalogUnavailable` < `commentaries`, so it belongs between them, not
      after `commentaries_*`.)_

### Open — author's call, deliberately not resolved here

- [x] **PT-4339 / NN-5b coordination check** _(closed: PR #2682 merged 2026-08-24 and is a git
      ancestor of this HEAD. The two PRs share exactly one file (`localizedStrings.json`, verified
      key-by-key against main), touch different surfaces, and this PR's changed code contains no
      editability signal at all — so reactive `editable` cannot change which message these panels
      show. See the PR conversation for the full check.)_
- [x] **Should the discriminated union stay a novel pattern here?** _(decided by the author: keep
      both shapes, chosen by whether the payload is state-specific. Recorded as
      [ADR-0028](../.context/standards/Architecture-Decisions.md) and folded into
      `Component-Builder-Patterns.md` as a rule, so the next hook does not re-litigate it — a
      discriminated union when data exists in only one state, a flat named state object when the
      values are always present, and never unpack a union at a boundary.)_
- [x] **AC-4 is now fully met** _(fixed: all four front states compose the shadcn `Empty` primitive
      per ADR-0016, each with its own icon — `AlertTriangle` for the settings error, `CloudOff` for the
      catalog error, `BookOpen` for the pick prompt, and the spinner for loading. Before this, `empty`
      and `catalogError` rendered character-identical containers whose buttons did opposite things
      (reconfigure vs. retry). A test pins that the three icons stay distinct; mutation-checked.)_
- [ ] **Bare `"Loading…"`** (copy call) does not meet the contract `LoadingView`'s own docstring sets,
      and is weakest exactly when the wait is longest.
- [ ] **File a ticket: no `resource-text-panel.web-view.test.tsx`**, so the readiness wiring in the
      panel where the bug originated is untested.
- [ ] **File a ticket: silent blank Text Collection grid** on an unreadable admin setting. Confirmed
      by the reviewer against the merge base as **pre-existing**, not introduced here.

### Skipped — recommended against, reversible

- [ ] ~~`resource-panel-readiness.utils.ts` imports from a `.hook` module~~ _(the only `*.utils.ts` in
      the extension to do so. Still skipped, but note the decision above removed half its
      justification: it would have become moot had the flat shape won, and the union is being kept. It
      now stands as a known wart, recorded in ADR-0028's consequences — moving the union into the utils
      module or a `.types.ts` is the fix whenever it bothers a reader.)_
- [x] **Default-export inconsistency between the two new files** _(fixed: added `export default` to
      `use-dbl-resource-catalog.hook.ts`. A census settled it by file type rather than forcing
      uniformity — `*.utils.ts` is unanimous at 0 of 10 defaults, `*.hook.ts` is a clear majority at
      8 of 12, and `*.component.tsx` has no convention at 3 of 8. Each new file now matches its own
      file type, and the two no longer disagree.)_
- [x] **Stories hardcoded English in `meta.args` instead of `getLocalizedStrings`** _(fixed: the
      story now resolves its copy from the extension's `localizedStrings.json`. The hardcoded copy had
      already drifted — "Bible text" where the shipping strings say "Bible texts". Keys are listed in
      the story rather than imported from `RESOURCE_PANEL_STRING_KEYS`, which lives in the web-view
      module: importing it would pull the whole web view and `Editorial` into Storybook for a string
      array.)_

## Findings — second review round (Claude AI Review panel)

Five perspectives, 0 agent-blocking, 14 warnings / 15 suggestions. Both of @jolierabideau's original
blocking items were re-verified as fixed by that panel.

### Blocking — addressed, except the one only a human can close

- [x] **Malformed `items` crashed the panel** _(fixed: `mergeResourceReferenceLists` guarded both
      lists against `PlatformError` and `undefined` but not a non-array `items`, which reaches us via
      send/receive without passing `resourceReferenceListValidator`. Reproduced first —
      `items.filter is not a function`, thrown inside the memo, unmounting the panel with no error
      boundary. Guarded; reproduction kept as a test.)_
- [x] **`console.error` where every sibling uses `logger`** _(fixed: the only two `console.*` calls in
      this extension's production code. Web-view console output never reaches the Platform.Bible log
      files, so both failure paths were invisible to log-based diagnosis. Also `${err}` →
      `getErrorMessage(err)`, which produced `[object Object]`.)_
- [x] **ADR-0027 recorded the opposite of what shipped** _(fixed: it claimed `InstallFailedView`
      "keeps its name and its scope", written before the rename to `RetryableErrorView`. ADR-0016 also
      cited a path this branch deletes.)_
- [x] **PT-4339 / NN-5b coordination check** _(closed: PR #2682 merged 2026-08-24 and is a git
      ancestor of this HEAD, so it is already integrated here. The two PRs share exactly one file
      (`localizedStrings.json`, verified key-by-key against main — nothing dropped or altered), touch
      different surfaces, and this PR's changed code contains no editability signal at all. #2682's
      string change is the editor's title indicator, not a loading or empty-state message.)_

### Important — addressed

- [x] **ADR-0016 / the `Empty` primitive** _(fixed: all four front states compose the vendored shadcn
      `Empty`, including `RetryableErrorView`, so no hand-rolled/`Empty` split is left behind.)_
- [x] **AC-4 only half met** _(fixed: each state carries its own icon — `AlertTriangle`, `CloudOff`,
      `BookOpen`, spinner. The hazard was not similarity but that `empty` and `catalogError` were
      character-identical containers whose buttons do opposite things. A mutation-checked test pins
      the icons distinct.)_
- [x] **`getResourcePanelReadiness` took an unpacked union** _(fixed: it now takes the whole
      `listState` and derives the count from the narrowed union, so it no longer contradicts the
      standard this PR adds. `matchingCount` stays optional — it needs the catalog and the panel's
      resource type, so it is not derivable there.)_
- [x] **ADR-0028 binding on two examples** _(fixed: marked Proposed rather than Accepted, with the
      same caveat mirrored in the standard.)_

### Important — open, author's decision

- [ ] **`catalogError` is new behaviour, not just a bug fix.** A new failure screen, a retry, and four
      new strings. Investigated: the retry re-runs `getCachedResources`, which the panel already calls
      unconditionally on mount, and which fetches a catalog *listing* rather than resource content —
      so it is further from the PRD's "Resource download and access → not covered" no-go than it looks.
      Worth knowing that reverting is not neutral: the alternatives are main's misleading empty prompt
      or an unrecoverable spinner.
- [ ] **`useBufferedLayoutSetting`'s third tuple element is a second, independent fix** bundled here.
      Coupled — without a distinguishable error channel the new `error` status is underivable — but
      keep-vs-split is the author's call.

### Skipped — advisory, recommended against, reversible

Tests:

- [ ] ~~`use-text-collection-sources.hook.test.ts:388` trivial identity-pass-through assertion~~
      _(already enforced by TypeScript at every call site.)_
- [ ] ~~`use-buffered-layout-setting.hook.test.ts:91` `toHaveBeenCalled()` on `useEvent` is
      implementation-mirroring~~ _(redundant with the behaviour assertions two lines below, but
      harmless and pre-dates this branch's changes to that file.)_
- [ ] ~~`panel-readiness-view.component.test.tsx:43` `container.querySelector('svg')` is
      structural~~ _(kept deliberately: the AC-4 icon test now depends on querying the icon slot, so
      the structural query earns its place rather than duplicating `getByRole`.)_
- [ ] ~~Three test files stub all of `platform-bible-react` with a single `useEvent` mock~~ _(the
      `importOriginal()` pattern is safer, but converting them is unrelated churn in files this branch
      only touches for other reasons.)_

UX / accessibility — all pre-existing, none introduced here:

- [ ] ~~Resource-selector dropdown trigger has no stable accessible name~~
- [ ] ~~A failed chapter fetch leaves a blank editor with no error or retry~~
- [ ] ~~`notFoundState` lacks `role="alert"` unlike the other error views~~
- [ ] ~~`role="status"` mounted with content is unreliably announced; a persistent live region at the
      panel root would be more robust~~
- [ ] ~~`aria-label` on a generic `div` with no role has no reliable AT effect~~

_(These are real, and worth a follow-up ticket for the panel's accessibility as a whole rather than
piecemeal fixes inside a bug-fix PR.)_

Clarity:

- [ ] ~~Backward-facing comments across ~8 sites~~ _(accurate: many doc blocks explain how the code
      reached its current state during this PR rather than what a maintainer needs. Fixing it is a
      broad rewrite of comments this branch is otherwise done with; better as its own pass.)_
- [ ] ~~`isLoadingResources` is in `useDblResourceCatalog`'s return shape but no production caller
      reads it~~ _(true; it is a coherent part of a catalog hook's shape and removing it is a design
      call rather than a cleanup.)_

#### Shared Regions Modified

None. No `#region shared with <url>` markers exist in any changed file.

#### Extension Config Changes

None. No `package.json`, `tsconfig.json`, `webpack.config.ts`, `.eslintrc*`, `.erb/configs/`, or
`.github/workflows/` changes. No dependency changes.

## Positive Observations

- Root-cause framing is honest: the ADR records that the ticket's named conditional was the symptom
  site, not the defect, along with rejected alternatives and why.
- Replacing the tuple with a discriminated union makes the error state representable instead of
  masquerading as loading-or-empty, and the "`ready` may carry zero items" contract is documented on
  the type itself.
- "Nothing configured at all" is deliberately kept catalog-independent, so a genuinely unconfigured
  project keeps its immediate pick prompt rather than waiting on a fetch.
- The two failure states differ in whether they offer a control, and that difference is principled:
  a catalog fetch can be re-driven, a project-setting read cannot.
- TSDoc on the new exports explains _why_ each state exists and what the failure mode was, not just
  what the parameter is; the inline comments at each guard do the same and would survive a refactor
  that a PR description would not.
- Localization is clean: new keys defined in both shipped locales, inserted in alphabetical position,
  registered in both panels' string-key lists, no hardcoded user-visible text.
- Tests track the change rather than trailing it: pure-function suites, a render suite, hook tests
  pinning error/recovery/non-stickiness, and regression tests for both second-consumer gaps.

## Interview Notes

**Stated purpose**: PT-4347 — resource panels never show a premature empty state while a configured
resource is still resolving.

**Provenance, stated plainly for the reviewer**: this branch was produced in an AI-assisted session.
The author made the scope decisions at each checkpoint — fix the shared root cause plus both panels;
add a distinct error state; fix all four Important findings; fix three of the four deferred Minor
findings — and the analysis, implementation, and in-review fixes were AI-generated and verified in
that session. The author did not narrate independent technical explanations of the mechanisms during
this review, so **none are recorded here, and none should be inferred**. The reasoning captured
above and in ADR-0027 is the tool's, not the author's. A reviewer should verify the key claims
independently rather than treat them as author-vouched — in particular the three listed under
Suggested Review Focus.

**Verification performed during review**: every fix was driven by a test watched failing first, and
the two highest-risk ones were additionally mutation-checked (reverting the arm fix fails the
recovery test; stubbing the error expression fails the `settingError` assertion; disabling the
loading guard fails the empty-prompt test). Two findings raised by the analysis passes were verified
against the code before acting rather than accepted at face value; one of them
(`useTextCollectionSources`) proved to be a real regression introduced by this branch, and one
rationale (the `?.` assertions) proved to be incorrect and is corrected above.

**Known limitation the author should not let slide**: the ticket itself notes the admin-configured
path was never confirmed by a user. This is now a _demonstrated_ defect rather than a theoretical
race, but it is still not confirmed as the failure users reported. It should not close out NN-5c on
its own; PT-4148 and the S/R resource gap remain tracked separately.

## In-Review Quality Check

All fixes were made during the review, so the branch was re-verified after each:

- **Tests**: 1454 passing across 97 files in the `extensions` workspace (13 test files touched or
  added). Full suite run after every change; zero failures.
- **Lint**: 0 errors across all changed files. Two warnings remain — both `no-console` in
  `use-effective-resource-reference-list.hook.ts`, confirmed pre-existing at the merge base.
- **Typecheck**: 0 errors across all 21 changed files (verified individually; the workspace-wide
  `tsc` run reports hundreds of pre-existing errors in untouched files, so it was filtered to the
  changed set).
- Two lint suppressions were refused and fixed properly instead: a `no-nested-ternary` in the stories
  became a named `modelTextsStatusFor` helper, and a `no-type-assertion` in a test fixture became a
  real `DblResourceData` object.

**Note for the author, unrelated to this branch**: running `npm run lint` from the repo root inside a
git worktree silently corrupts `lib/papi-dts/papi.d.ts`. The script begins with `build:types`, whose
`tsc` fails on missing `@types/node`/`@types/jest` (worktrees have no installed deps there) but still
emits, rewriting the generated file from an incomplete program — observed as 9,453 insertions /
12,157 deletions. It was caught and reverted here, but is worth checking in other worktrees.

## Follow-up tickets this PR surfaced (not fixed here)

- **`usePromise` can latch `isLoading` true forever.** Two PRs in the same sprint independently hit
  this and worked around it locally — #2682 by distinguishing "will never resolve" from "loading",
  this PR by catching the rejection inside `useDblResourceCatalog`. Belongs fixed in
  `platform-bible-react`.
- **`getCachedResources` has a dead catch** (`platform-get-resources/src/main.ts`):
  `return fetchAndCacheResources();` inside a `try` with no `await`, so an async rejection escapes the
  handler entirely. This is why the catalog fetch rejects at all, and therefore why the spinner bug
  was reachable in production rather than theoretical.
- **Panel accessibility as a whole** — the five pre-existing a11y findings skipped above.
- **`resource-text-panel.web-view.tsx` has no test file**, so the readiness wiring in the panel where
  the bug originated is covered only through the pure helper and the shared view.

## Suggested Review Focus

Updated after both review rounds and the PT-4339 check. Closed items are marked rather than deleted,
so a reviewer can see what was decided and not just what remains.

- [x] **The PT-4339 coordination check** — closed by integration; #2682 is merged and in this branch.
      A one-line confirmation from @tombogle is still welcome but no longer gating.
- [x] **Convention: discriminated union vs flat state object** — settled. Both shapes kept, chosen by
      whether the payload is state-specific; see ADR-0028 and the new "Async Hook State Shape" section
      in `Component-Builder-Patterns.md`.
- [x] **AC-4's visual distinction** between `empty`, `error`, and `catalogError` — closed. Each state
      now composes shadcn `Empty` with its own icon, and a mutation-checked test pins them distinct.
- [ ] **`useBufferedLayoutSetting` now has two channels.** Consumers must check the third tuple
      element, not just the held value — trusting the held value alone reads an unreadable setting as
      an empty one. Both current consumers were updated; this is a trap for the next one.
- [ ] **The catalog-error path is new behaviour, not only a bug fix.** On `main` a rejected catalog
      fetch produced a (wrong) empty prompt; it now produces a message plus retry.
- [ ] **Scope**: the ticket named only the Model Text panel. This branch also changes the Resource
      panel, `useTextCollectionSources`, and the catalog fetch. The reviewer agreed the widening was
      justified; worth your own confirmation.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2697)


jolierabideau:
## Review summary

Reviewed against PT-4347, the NN-5c row of PT-4336, and the surrounding code patterns. **Overall: the diagnosis is right and better than the ticket's, and I'd merge this after the two correctness items below.** Re-deriving the root cause in `useEffectiveResourceReferenceList` rather than splitting the conditional the ticket named is the correct call — I confirmed the ticket's proposed fix would not have worked, because in the failing window both booleans are `false` and there is nothing honest to split on.

I verified the branch independently rather than taking the PR description's self-review at face value, as it asks. Details at the bottom; two of its claims need correcting.

---

### 1. Does it solve the ticket and fit the PRD?

**Acceptance criteria** — 1, 2, and 3 are met in the code. AC-4 ("loading and empty states are visually distinct from each other and from error states") is only half met: `empty`, `error`, and `catalogError` all render the same centered-body-text container, differing by copy and by whether a button is present, and both error strings open with the same three words ("Couldn't load…"). Only `loading` is distinguishable at a glance. Worth a distinguishing icon or text treatment on the two failure states.

**PRD fit** — good, with one coordination gap:

- Keeping "nothing configured at all" catalog-independent is the right reading of the epic. A genuinely unconfigured project still gets its pick prompt immediately instead of waiting on a fetch. I checked that this is safe: the picker is a separate `platform.resourcePicker` dialog with its own fetch, so it is not fed by this panel's (possibly failed) catalog.
- Honest about not closing NN-5c on its own — PT-4148 and the S/R resource gap stay tracked separately, and the PR says so rather than implying the row is done. Good.
- **Gap:** the NN-5b row of PT-4336 says explicitly — *"PT-4339 (Tom Bogle, PR #2682 open): Making `editable` state in the Scripture editor reactive. Affects which loading/empty-state message is correct to show — **check with Tom before finalizing NN-5b/5c loading message logic**."* This PR finalizes exactly that logic for 5c and I see no sign that check happened. Please confirm with Tom before merge; if `editable` becomes reactive, the correct message in some of these states may change.

**Scope** — the ticket named only the Model Text panel; this also changes the Resource panel, `useTextCollectionSources`, and the catalog fetch. I think the widening is justified (single shared root cause, and leaving one panel on the old logic is what the ADR is about), so no objection.

---

### 2. Does it match codebase patterns?

Mostly yes — naming (`*.utils.ts` / `*.component.tsx` / `*.hook.ts`), colocated tests and stories, feature UI kept in the extension rather than pushed into `platform-bible-react`, ADR recorded, localization in the extension's `contributions/` file. I also checked whether a shared view should have been reused instead of a new local one: `EmptyState` in `platform-bible-react` takes `message: string` with no action slot, so it can't carry the pick button — building these locally is correct.

Two real divergences:

- **The discriminated union is a novel pattern here.** A deterministic grep for `status: 'loading' | 'ready' | 'error'` across `extensions/src`, `lib`, and `src` returns *only* the seven new lines in `use-effective-resource-reference-list.hook.ts` — there is no other discriminated-union async state in the repo. Siblings use either a tuple (`useBufferedLayoutSetting` → `[value, isLoading, error]`) or a flat named state object with an error field (`useStructureProtectionState` → `StructureProtectionState.adminSettingError`). Notably **this PR's own new `useDblResourceCatalog` uses the flat-object pattern**, so the two new hooks disagree with each other.
- **The union's benefit is discarded at the boundary that matters.** Both consumers immediately unpack it back into the shape it replaced (`status === 'ready' ? state.list : undefined`), and `ModelTextPanelProps` takes `modelTextsStatus: EffectiveResourceReferenceListState['status']` — indexing the discriminant off the union strips the payload, so the panel receives an `undefined`-able list *plus* a loose status string: precisely the two-independent-values shape the union existed to prevent. Either pass the whole state object down so narrowing survives into the component, or converge on the flat-object shape used by `useStructureProtectionState` and `useDblResourceCatalog`.

I don't think either blocks merge, but the second one means the type change is currently paying its cost without buying the safety.

---

### 3. Should address before merge

**a. Stale-write race in `useDblResourceCatalog` can reintroduce this very bug.** `setHasCatalogError` is called from inside the `usePromise` factory with no "is this fetch still current" guard. I read `usePromise` (`lib/platform-bible-react/src/hooks/use-promise.hook.ts:61-80`) to confirm: its `promiseIsCurrent` flag is local to its own effect and guards only `setValue`/`setIsLoading` — a superseded factory invocation still runs to completion and still writes our state. Overlapping fetches are reachable because `refetchCatalog` is wired as the install-completion stale marker (`resource-text-panel.web-view.tsx:259-263`). The bad interleaving: a late-*resolving* superseded fetch clears a real error, leaving `hasCatalogError === false`, `dblResources === []`, and `isCatalogReady === true` → `getResourcePanelReadiness` returns `'empty'`, i.e. the premature empty state this PR exists to remove. A fetch-generation ref around both `setHasCatalogError` calls would close it.

**b. Stale TSDoc that contradicts this PR's own code.** `panel-readiness-view.component.tsx:10-14` still says *"The Model Text panel renders the equivalent states inline… it was left alone rather than churned. It could adopt this view if its props ever carry the list status instead of separate loading/error booleans."* The Model Text panel **does** adopt it (`model-text-panel.component.tsx:394-418`) and its props **were** changed to carry `modelTextsStatus`. ADR-0014 says both panels route through it; this comment says the opposite. Since preventing exactly that drift is the abstraction's purpose, this one matters more than a normal stale comment.

---

### 4. Worth addressing (test strength)

**The panel-level test for the headline acceptance criterion under-asserts — I verified this by mutation.** I changed `if (!isCatalogReady) return 'loading'` to `return 'empty'` in `resource-panel-readiness.utils.ts` and re-ran the suites: **only** `resource-panel-readiness.utils.test.ts` failed (1 test). All 14 tests in `model-text-panel.component.test.tsx` passed — including `it('does not claim the model text is missing while the resource catalog has not arrived')`, which sets up the exact AC-1 scenario but only asserts the *not-found* string is absent. It never asserts that "No model text selected." is absent or that the spinner is present, so it cannot fail on the mutation it appears to guard. Adding `expect(screen.queryByText('No model text selected.')).not.toBeInTheDocument()` and a `getByRole('status')` would fix it. (Mutation reverted; tree clean.)

Related: there is no `resource-text-panel.web-view.test.tsx` at all, so the readiness wiring in the panel **where the bug originated** — `configuredCount` vs `matchingCount`, the two same-typed numbers the options object exists to protect — is the one place they're paired and it's untested. The PR acknowledges this and calls the extraction out of scope; I'd agree it needn't block merge, but it's worth a ticket rather than a deferral note.

---

### 5. Minor

- `isCatalogReady` returns `true` after a *failed* fetch (the catch returns `[]`, making the value defined), contradicting its own TSDoc ("finished loading AND delivered"). Harmless today only because `getResourcePanelReadiness` checks `hasCatalogError` first; a future consumer reading it alone would treat a failed fetch as a genuinely empty catalog. Consider folding `!hasCatalogError` into it.
- `InstallFailedView` is now used for a failed *catalog fetch*, but its doc still says "Re-attempts the install for the same resource". The file was renamed anyway — `RetryableErrorView` would fit.
- Story docs contradict the design: `model-text-panel.stories.tsx` calls `SettingsError` "a distinct, **recoverable** error state" and the decorator flag says "the recoverable settings-error state", when the whole point (per the ADR) is that it is *not* recoverable from the panel.
- The stories hardcode English copy in `meta.args` instead of `getLocalizedStrings([...])` from `.storybook/localization.utils`, which 44 story files use — including `model-text-panel.stories.tsx` in this same directory. Since the story's stated purpose is to make this copy reviewable without running the app, hardcoding lets it drift from `localizedStrings.json`. (`RESOURCE_PANEL_STRING_KEYS` would need exporting, as `MODEL_TEXT_PANEL_STRING_KEYS` already is.)
- `%webView_resourcePanel_catalogUnavailable%` breaks the block's alphabetical order in both locales (it sits before `bibleTexts_*`; belongs after `commentaries_*`). The other 5 new keys are placed correctly.
- `"Loading…"` is bare, which doesn't meet the contract `LoadingView`'s own docstring sets ("a label saying what is being waited on… e.g. 'Installing resource…'") and reads oddly beside its siblings. The doc also notes these panels can now wait through a whole catalog fetch — which is exactly when a bare "Loading…" tells Saroj nothing.
- Neither failure view has a live-region role, while `LoadingView` has `role="status"`. A screen-reader user sitting on the panel when it flips from loading to error hears nothing.
- `resource-panel-readiness.utils.ts` is the only `*.utils.ts` in the extension that imports from a `.hook` module. Declaring the status type in the utils module (or a `.types.ts`) would keep the dependency pointing hook → utils.
- The two new files land on opposite sides of the directory's default-export inconsistency (`panel-readiness-view` has one; `use-dbl-resource-catalog` doesn't, unlike 8 of 12 sibling hooks).

---

### 6. Two corrections to the PR description's self-review

- **The Text Collection blank-grid outcome is not a regression this PR introduces, and the residual gap is worth stating plainly.** The description is right that un-latching would have made `sources` assemble from an empty admin list, and right to fix it. But the fix routes the error into the same `undefined` that means "loading", and the grid's empty state is gated on `sources !== undefined` (`scripture-text-grid.web-view.tsx:522`) — so an unreadable admin setting renders `ScriptureTextGrid` with `resources={[]}`: a blank grid body, no message, and View Options disabled with `disabledMessage={undefined}`. I checked the merge base: the old hook applied the `PlatformError` to the held copy, so `isPlatformError(adminReferenced)` was true and `sources` was `undefined` **then too** — same blank grid. So this PR restores parity and genuinely improves the late-failure and recovery cases; the silent blank grid is pre-existing. Worth its own ticket, not a change here.
- **One residual unresolvable spinner remains on the primary path**, despite "no unresolvable spinners" being the PR's premise. If the user-level `subscribeUser*` promise *rejects*, the `.catch` only logs; `userResourceReferenceList` stays `undefined`, so the memo returns `{ status: 'loading' }` forever. The hook's doc claims "If the user setting cannot be retrieved, the project-level items are returned tagged as `'admin'`", which holds only for a `PlatformError` delivered *through the callback*, not for a rejected subscribe. Unchanged from the merge base, so not a blocker — but it's the same failure class.

---

### What I verified independently

- **Tests**: 809 passing across 57 files in `platform-scripture-editor` (run from the `extensions` workspace).
- **Lint**: 0 errors on all changed source files. The 2 remaining `no-console` warnings in `use-effective-resource-reference-list.hook.ts` are confirmed pre-existing — they're at lines 28 and 116 of the merge-base file.
- **Typecheck**: 0 errors in any of the 21 changed files. (254 errors workspace-wide, all in untouched files — consistent with the description's claim.)
- **Mutation**: the `isCatalogReady` mutation described in §4 (reverted).
- **Claim checks**: the `hasAppliedRealValue` gate, the caught catalog rejection, both panels routing through the shared helper, complete `retryEpoch` removal, and the `useTextCollectionSources` error-channel consumption are all **true of the code as it stands**.

Nice work on the ADR — recording that the ticket's named root cause was the symptom site, with the rejected alternatives and why, is more useful to the next person than the fix itself.

*AI-assisted review.*

jolierabideau (changes requested):
Two blocking comments and I left NITs in the summary comment

> @jolierabideau partially reviewed 21 files and all commit messages, and made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2697#-:-P-UvbrX7Knu77noeOYZ:b-ru1ofy)* status: all files reviewed, 2 unresolved discussions (waiting on captaincrazybro).

___
*[`extensions/src/platform-scripture-editor/src/panel-readiness-view.component.tsx` line 10 at r1](https://reviewable.io/reviews/paranext/paranext-core/2697#-P-UvWnt-2mxDiOywi30:-P-UvWnt-2mxDiOywi31:b6f7nad) ([raw file](https://github.com/paranext/paranext-core/blob/fff3cb7836dd36d30314c01212c03314b5541cbd/extensions/src/platform-scripture-editor/src/panel-readiness-view.component.tsx#L10)):*
> ```XML
>  * display — from a single `readiness` value.
>  *
>  * Used by the Resource (Bible Texts / Commentaries) panel, which lives entirely in a web view with
> ```

**Stale TSDoc that contradicts this PR's own code.** `panel-readiness-view.component.tsx:10-14` still says _"The Model Text panel renders the equivalent states inline… it was left alone rather than churned. It could adopt this view if its props ever carry the list status instead of separate loading/error booleans."_ The Model Text panel **does** adopt it (`model-text-panel.component.tsx:394-418`) and its props **were** changed to carry `modelTextsStatus`. ADR-0014 says both panels route through it; this comment says the opposite. Since preventing exactly that drift is the abstraction's purpose, this one matters more than a normal stale comment.
___
*[`extensions/src/platform-scripture-editor/src/use-dbl-resource-catalog.hook.ts` line 54 at r1](https://reviewable.io/reviews/paranext/paranext-core/2697#-P-UvPmi6SKJdPYE_Ub8:-P-UvPmi6SKJdPYE_Ub9:b-gzwsjw) ([raw file](https://github.com/paranext/paranext-core/blob/fff3cb7836dd36d30314c01212c03314b5541cbd/extensions/src/platform-scripture-editor/src/use-dbl-resource-catalog.hook.ts#L54)):*
> ```XML
>           'platformGetResources.getCachedResources',
>         );
>         setHasCatalogError(false);
> ```

**Stale-write race in `useDblResourceCatalog` can reintroduce this very bug.** `setHasCatalogError` is called from inside the `usePromise` factory with no "is this fetch still current" guard. I read `usePromise` (`lib/platform-bible-react/src/hooks/use-promise.hook.ts:61-80`) to confirm: its `promiseIsCurrent` flag is local to its own effect and guards only `setValue`/`setIsLoading` — a superseded factory invocation still runs to completion and still writes our state. Overlapping fetches are reachable because `refetchCatalog` is wired as the install-completion stale marker (`resource-text-panel.web-view.tsx:259-263`). The bad interleaving: a late-_resolving_ superseded fetch clears a real error, leaving `hasCatalogError === false`, `dblResources === []`, and `isCatalogReady === true` → `getResourcePanelReadiness` returns `'empty'`, i.e. the premature empty state this PR exists to remove. A fetch-generation ref around both `setHasCatalogError` calls would close it.

captaincrazybro:
Thanks — this was a genuinely useful review. Both blocking items were valid and both were mine, and the mutation you ran found a real hole in my own test. Pushed `54b0ea2` and `d54f5715`.

## Blocking

**§3a — stale-write race in `useDblResourceCatalog`.** Confirmed, and I reproduced it before fixing rather than patching from the description. A test driving your exact interleaving (a superseded fetch resolving late and clearing a real error) failed with `expected false to be true`, leaving `hasCatalogError === false` + empty catalog + `isCatalogReady === true` → readiness `'empty'`. So my own new hook could reintroduce the precise bug this PR exists to remove. Fixed with a `fetchGenerationRef` guarding both `setHasCatalogError` calls, incremented by `refetchCatalog`; the reproduction is kept as a regression test.

**§3b — stale TSDoc.** Correct, and correct that it mattered more than a normal stale comment. It was written before I converged both panels and I missed it in that pass. Rewritten to say both panels use the view.

## §4 — test strength

Replayed your mutation to confirm rather than take it on faith: before the fix, only `resource-panel-readiness.utils.test.ts` failed and all 14 model-panel tests passed. Added the empty-prompt absence and `getByRole('status')`; the mutation now fails the panel test too. Reverted.

## §2 — the boundary critique

This was the most valuable point in the review and I've fixed it. `ModelTextPanelProps` now takes the whole `modelTextsState` rather than an `undefined`-able list plus a status string. You were right that indexing the discriminant strips the payload — converging the panels' logic had quietly reinstated the two-independent-values shape at the prop boundary.

Since the guarantee is compile-time I verified it with a deliberate type probe instead of a test: `{ status: 'ready' }` with no list is rejected `TS2322`, `{ status: 'loading', list }` is rejected `TS2353` (probe removed). The clearest evidence you were right: migrating the tests meant deleting one that passed `effectiveModelTexts: undefined` *together with* `modelTextsStatus: 'loading'` — my suite was asserting an illegal combination, which was only possible because the boundary allowed it.

**Still open, and I'd like your view:** whether the discriminated union should remain a novel pattern here at all, given `useDblResourceCatalog` uses the flat-object shape. My reasoning for leaving them different is that the union's payload exists in exactly one state, whereas the catalog hook returns several always-present values plus flags — but that's a judgment about repo convention where there's no precedent, so I didn't want to decide it unilaterally.

## §5 / §6 — also fixed

- `isCatalogReady` was `true` after a failed fetch; folded `!hasCatalogError` in and fixed the doc.
- `InstallFailedView` → `RetryableErrorView`. Worth noting I'd already reverted an earlier rename *because* it only served installs, then re-broke it by wiring the catalog case through it.
- `role="alert"` on both failure views — you're right that the flip out of `role="status"` announced nothing.
- Story docs no longer call the settings error "recoverable".
- The rejected-`subscribeUser*` spinner: fixed rather than deferred, since the hook's doc already promised the project-level fallback. Failing test first (`expected 'loading' to be 'ready'`), then the `.catch` falls back to `DEFAULT_LIST`.

**One correction.** On the misplaced localization key — the problem is real, but the suggested target isn't: `bibleTexts` < `catalogUnavailable` < `commentaries` (`"ca"` < `"co"`), so it belongs *between* them rather than after `commentaries_*`. Placed there in both locales.

## Deliberately not addressed

- **The PT-4339 / NN-5b coordination check with Tom has not happened** — thank you for catching it, I'd missed that row. Holding off on merge until it does, since reactive `editable` could change which message is correct in some of these states.
- **AC-4** — agreed it's half met; the three states share one container. Wants an icon or text treatment, which felt like a design decision rather than something to pick unilaterally in a review pass.
- **Bare `"Loading…"`** — agreed it doesn't meet the contract `LoadingView`'s own docstring sets. Copy call.
- **`resource-text-panel.web-view.test.tsx`** and the **silent blank Text Collection grid** — filing tickets rather than deferral notes, per your suggestion. Thanks also for checking the merge base on the blank grid; agreed it's pre-existing and that this branch restores parity.
- Three low-value items I'd skip: the `utils` → `.hook` import direction (moot if we change the pattern above), the default-export inconsistency (the directory has no convention to follow — 8 of 12), and the stories' hardcoded English (real, but needs `RESOURCE_PANEL_STRING_KEYS` exported for a story-only benefit — happy to do it if you'd rather).

Verified after all of the above: 1456 tests / 97 files pass, ESLint 0 errors (the 2 `no-console` warnings remain pre-existing).

*AI-assisted.*

captaincrazybro:
Small correction to my previous comment, plus a description refresh.

**§5.9 default-export inconsistency is now fixed** (`6ca7688`) — I'd listed it as one of three I'd skip. A census settled the direction rather than guessing, and it turns out "the directory has no convention" is only true for components:

| File type | Defaults |
|---|---|
| `*.utils.ts` | 0 of 10 — unanimous |
| `*.hook.ts` | 8 of 12 — clear majority |
| `*.component.tsx` | 3 of 8 — no convention |

So I aligned by file type instead of forcing uniformity: `use-dbl-resource-catalog.hook.ts` gained the default, `resource-panel-readiness.utils.ts` keeps none, and the two new files no longer sit on opposite sides. That also retroactively justifies removing the default from the utils module earlier in the branch — the utils convention is unanimous, not merely typical.

**The PR description had not been updated for your review round at all**, which I should have caught before replying. It still described the pre-review state, including two API-surface entries this round had made false: `InstallFailedView` listed as "unchanged" after it became `RetryableErrorView`, and `modelTextsStatus` after it became `modelTextsState`. Now corrected, with every item from your review carrying an explicit state — `[x]` fixed, `[ ]` open, `[ ] ~~struck~~` skipped — so nothing reads as quietly resolved. The Suggested Review Focus list is replaced too; it predated the round.

Still skipped, and still reversible on request: the `utils` → `.hook` import direction (moot if we settle the union/flat-object question the other way) and the stories' hardcoded English.

Unchanged from my last comment: the **PT-4339 check with Tom has not happened**, and I'd treat that as blocking merge.

1456 tests / 97 files pass; ESLint 0 errors.

*AI-assisted.*

captaincrazybro:
**§2 pattern question — decided: keep both shapes.** Recorded as [ADR-0015](https://github.com/paranext/paranext-core/blob/worktree-pt-4347-resource-panel-loading-state/.context/standards/Architecture-Decisions.md) (`b3c9890`), and folded into `Component-Builder-Patterns.md` as an "Async Hook State Shape" section — per CLAUDE.md, a settled convention belongs in the standards, since that is what gets enforced on the next feature, while the log keeps the *why*.

The rule, chosen by whether the payload is **state-specific**:

- **Discriminated union** when data exists in only one state, so the type makes the other combinations unrepresentable. There is no list to hand out while `loading` or on `error`, and the union is what stops a caller reading one anyway. → `useEffectiveResourceReferenceList`.
- **Flat named state object** when the values are always present and the flags describe them. `useDblResourceCatalog` returns a catalog (coerced to `[]`), three flags, and a refetch callback — all meaningful together, nothing to make unrepresentable. → also `useStructureProtectionState`.
- **Corollary: never unpack a union at a boundary.** Indexing the discriminant (`SomeState['status']`) strips the payload and hands the callee two values free to disagree — precisely the shape the union exists to forbid. That is the mistake this branch made and you caught; the standard now carries it as a WRONG/RIGHT example so the next person doesn't repeat it.

So the two shapes are no longer an accident of one PR — they're a documented choice with the deciding question written down.

**One honest consequence.** Keeping the union removes half my justification for skipping §5.8 (`resource-panel-readiness.utils.ts` being the only `*.utils.ts` importing from a `.hook` module). I'd argued it would become moot if the flat shape won — it didn't, so that argument is gone. It stands as a known wart, recorded in ADR-0015's consequences with the fix named: move the union into the utils module or a `.types.ts`. Happy to do it if you'd rather not carry it.

**Also fixed since my last comment:** §5.4 (`194454f8`) — the story now resolves its copy from the extension's `localizedStrings.json`. Your drift concern was already real: the story said "No Bible **text** selected" / "Pick Bible **text**…" where the shipping strings say "**texts**". One deviation from your suggested mechanism — rather than exporting `RESOURCE_PANEL_STRING_KEYS`, which lives in the web-view module and would pull the whole web view (and `Editorial`) into Storybook for a string array, the story lists its six keys and resolves them from the same JSON. Trade accepted: a renamed key surfaces as the raw key in the story rather than a compile error.

That leaves §5.8 as the only item skipped, and the **PT-4339 check with Tom** as the one thing I'd still treat as blocking merge.

1456 tests / 97 files pass; ESLint 0 errors.

*AI-assisted.*

jolierabideau (commented):
## 🤖 [Claude AI Review] — PR #2697

**Reviewer panel**: security · scope · ux · tests · clarity
**Event**: COMMENT (no agent-severity-blocking findings; see blocking items below)

| Security     | 0        | 2       | 0          |
| Scope        | 0        | 6       | 4          |
| UX           | 0        | 3       | 4          |
| Tests        | 0        | 2       | 5          |
| Clarity      | 0        | 1       | 2          |
| **Total**    | **0**    | **14**  | **15**     |

---

### ✅ @jolierabideau's blocking items — both verified fixed

- **Stale TSDoc** (`panel-readiness-view.component.tsx:10`): now correctly says "Used by BOTH" panels.
- **Stale-write race in `useDblResourceCatalog`**: `fetchGenerationRef` guards both `setHasCatalogError` calls.

---

### 🔴 Blocking — must address before merge

**PT-4339 / NN-5b coordination check (merge gate the PR body named itself)**
The PR says "Merge should wait on this." New loading and unavailable messages shipped in two locales across both panels. These need validation against Tom Bogle's expected NN-5b/5c loading-message design before merging, since reactive `editable` affects which message is correct. This is the one item only a human can close.

**ADR-0022 Consequences contradict the code** (`.context/standards/Architecture-Decisions.md:1046`)
ADR-0022 says _"InstallFailedView keeps its name and its scope."_ This PR renamed it to `RetryableErrorView` and moved it to `panel-state-views.component.tsx`. The ADR records the opposite of what shipped. Needs a Consequences update.

**`console.error` instead of `logger`** (`use-effective-resource-reference-list.hook.ts:28, 127`)
The only production file in this extension that calls `console.*` — all siblings (including both new hooks in this PR) use `logger` from `@papi/frontend`. Console output from a web view iframe doesn't reach the Platform.Bible log files, so both failure paths are invisible to log-based diagnosis. The call at line 127 also uses `${err}` rather than `getErrorMessage(err)`, producing `[object Object]` for non-Error rejections.

**Missing container-shape guard** (`use-effective-resource-reference-list.hook.ts:57`)
`projectResourceReferenceList.items` and `userResourceReferenceList.items` are guarded against `PlatformError` and `undefined` but not against a non-array `items` field. The values originate from a project file that is reachable via send/receive without passing through `resourceReferenceListValidator` (which runs on write only). A malformed `items` throws a `TypeError` inside the `useMemo`, unmounting the panel's React tree with no error boundary — the opposite of the graceful `error` readiness state this PR builds. A cheap `Array.isArray(list.items) ? list.items : []` guard closes it.

---

### 🟡 Important — design calls for the author

**ADR-0016 / `Empty` primitive not used** (`panel-readiness-view.component.tsx:82-85`)
ADR-0016 names PT-4347 explicitly and directs that any zero-state needing prompt text + an action compose the vendored shadcn `Empty` primitive. `PanelReadinessView`'s `empty` and `catalogError` branches hand-roll a fifth ad-hoc container instead, with no recorded rationale for the divergence. If intentional, ADR-0022 or a new entry should record why; composing `Empty` would also address AC-4's visual distinction largely for free.

**AC-4 only half met** (`panel-readiness-view.component.tsx:82`)
Three of four front states (`empty`, `error`, `catalogError`) render as centred body text inside near-identical containers; only `loading` is distinguishable at a glance. The PR body acknowledges this as "AC-4 is half met (design call)." Worth a conscious sign-off that this AC may ship incomplete.

**`catalogError` state is new behaviour, not just a bug fix** (`resource-panel-readiness.utils.ts:70`)
On main a rejected catalog fetch fell through to the empty prompt (wrong but harmless). This PR introduces a distinct failure screen, a retry control, and four new localized strings. The ticket separates out the S/R resource gap as conflicting with the PRD's "Resource download and access → not covered" no-go — this retry sits near that line. Worth an explicit scope sign-off.

**`useBufferedLayoutSetting` contract change bundled here** (`use-buffered-layout-setting.hook.ts:96`)
Adding a third tuple element to a shared hook is a second, independent defect fix. ADR-0022 records the third-tuple-element trap for future consumers. The change is coupled to the readiness fix (without a distinguishable error channel the new `error` status is underivable) — but keep-vs-split is the author's call.

---

### 🔵 NIT — advisory, no action required

**Tests**
- `use-effective-resource-reference-list.hook.test.ts:250`: `'reports loading when projectId is undefined'` passes because `isLoading: true` fires the loading guard first — `projectId` is never consulted. The named behavior isn't actually pinned.
- `use-dbl-resource-catalog.hook.test.ts:93`: The stale-fetch test pins only `hasCatalogError`; adding `expect(result.current.isCatalogReady).toBe(false)` would also guard the panel-visible symptom.
- `use-text-collection-sources.hook.test.ts:388`: Trivial identity-pass-through assertion, already caught by TypeScript at every call site.
- `use-buffered-layout-setting.hook.test.ts:91`: `toHaveBeenCalled()` on `useEvent` is implementation-mirroring — redundant with the behavior assertions two lines later.
- `panel-readiness-view.component.test.tsx:43`: `container.querySelector('svg')` structural assertion, redundant with `getByRole('status')` + label.
- `use-effective-resource-reference-list.hook.test.ts:27`: Three test files blank out all of `platform-bible-react` with a single `useEvent` stub; `model-text-panel.component.test.tsx:21` already shows the safer `importOriginal()` + override pattern.

**UX**
- `resource-text-panel.web-view.tsx:125`: Resource-selector dropdown trigger has no stable accessible name — announced as just "button" when `selectedRef` is undefined.
- `model-text-panel.component.tsx:226`: A failed chapter fetch silently leaves the user with a blank editor (no error, no retry).
- `model-text-panel.component.tsx:375`: `notFoundState` renders an error but lacks `role="alert"`, unlike the other error views this PR adds.
- `panel-state-views.component.tsx:21`: `role="status"` mounted with content is unreliably announced; a persistent live region at the panel root would be more robust.
- `model-text-panel.component.tsx:504`: `aria-label` on a generic `div` with no role has no reliable AT effect.

**Clarity**
- Backward-facing comments are pervasive: many doc blocks explain how code reached its current state during this PR (the old bug, what went wrong, what the review found) rather than what a future maintainer needs. Raised once at PR level per the forward-facing-comments rule. Key sites: `resource-panel-readiness.utils.ts:40-44`, `use-dbl-resource-catalog.hook.ts:29-32`, `panel-readiness-view.component.tsx:10-17`, `resource-text-panel.web-view.tsx:557-560`, `use-effective-resource-reference-list.hook.ts:128-130`, `model-text-panel.component.tsx:162-163`, `:392-395`, `:422-423`.
- `use-dbl-resource-catalog.hook.ts:11`: `isLoadingResources` is in the public return shape but neither production caller reads it.
- `resource-panel-readiness.utils.ts:20`: `getResourcePanelReadiness` takes `listStatus: EffectiveResourceReferenceListState['status']` (an unpacked union) — which the new Component-Builder-Patterns standard added in this same PR calls out as the wrong shape. Either the function should take the whole state, or the standard should explicitly scope itself to component props.

**Scope / docs**
- `.context/standards/Architecture-Decisions.md:599`: ADR-0016 cites the now-deleted `install-state-views.component.tsx` path — `/verify-standards` will flag it. A parenthetical noting the rename in ADR-0022 would keep the log honest.
- `.context/standards/Architecture-Decisions.md:992`: ADR-0022's heading is missing a blank line from ADR-0021's trailing Source bullet (formatting only).
- `.context/standards/Component-Builder-Patterns.md:172`: The async-hook-shape convention is now a binding repo standard on the strength of two examples from this PR. Worth a conscious sign-off — or leaving ADR-0023 as "Proposed" until a third hook exercises it.

captaincrazybro:
Second review round worked through. **11 of 15 actionable items resolved**; the rest are marked skipped with reasons in the PR description rather than left ambiguous.

## Blocking — 3 of 4 resolved

**Malformed `items` crashed the panel.** The sharpest find of the round, and I reproduced it before fixing rather than patching from the description: `items.filter is not a function`, thrown inside the memo, which unmounts the panel with no error boundary — the exact opposite of the graceful `error` state this branch exists to provide. Guarded with `Array.isArray`, and the reproduction is kept as a regression test.

**`console.error` where every sibling uses `logger`.** Correct on all counts — these were the only two `console.*` calls in this extension's production code, and web-view console output never reaches the Platform.Bible log files, so both failure paths were invisible to log-based diagnosis. Also swapped `${err}` for `getErrorMessage(err)`. This clears the two pre-existing `no-console` warnings that earlier rounds kept reporting.

**ADR-0022 recorded the opposite of what shipped** — it claimed `InstallFailedView` "keeps its name and its scope", written before the rename. Corrected, along with ADR-0016's citation of a path this branch deletes.

**PT-4339 / NN-5b coordination with Tom is still open** and I'd still treat it as the merge gate.

## Important — all four resolved

**ADR-0016 / the `Empty` primitive.** Adopted for all four front states, including `RetryableErrorView`, so no hand-rolled/`Empty` split is left behind. Worth noting the history: @jolierabideau's first review explicitly blessed building these locally because `EmptyState` has no action slot. That was true then — ADR-0016 landed `Empty` under us while this branch was open, and rebasing is what surfaced the conflict.

**AC-4 now fully met.** Each state has its own icon (`AlertTriangle` / `CloudOff` / `BookOpen` / spinner). The hazard was worse than "they look similar": `empty` and `catalogError` rendered *character-identical* containers whose buttons do opposite things — reconfigure vs. retry — so a user who learned "the button here picks a resource" would hit retry expecting a picker. A test pins the icons distinct; mutation-checked by making two states share one.

**`getResourcePanelReadiness` took an unpacked union** — a genuine self-contradiction, since the Component-Builder-Patterns section this same PR adds calls that the wrong shape. It now takes the whole `listState` and derives the count from the narrowed union. `matchingCount` stays optional because it needs the catalog and the panel's resource type, so it genuinely cannot be derived there.

**ADR-0023 downgraded to Proposed**, with the same caveat mirrored in the standard so nobody reads it as settled on two examples.

## Two Important items left open for the author

- **`catalogError` as new behaviour.** I dug into the no-go concern: the retry re-runs `getCachedResources`, which the panel already calls unconditionally on mount, and which fetches a catalog *listing* rather than resource content — so it sits further from "Resource download and access → not covered" than it first reads. The thing to weigh is that reverting isn't neutral: the alternatives are main's misleading empty prompt or an unrecoverable spinner.
- **`useBufferedLayoutSetting`'s third tuple element** being a second, independent fix. Coupled (without a distinguishable error channel the new `error` status is underivable), but keep-vs-split is the author's call.

## Skipped, with reasons

The a11y NITs (dropdown accessible name, blank editor on failed chapter fetch, `notFoundState` missing `role="alert"`, live-region robustness, `aria-label` on a bare `div`) are all **real and all pre-existing** — none introduced here. They deserve a follow-up ticket for the panel's accessibility as a whole rather than piecemeal fixes inside a bug-fix PR.

The backward-facing-comments finding is **accurate** — many of my doc blocks explain how the code reached its current state during this PR rather than what a maintainer needs. Fixing it is a broad comment rewrite, better as its own pass.

The remaining test NITs are either already enforced by TypeScript, or unrelated churn. One I kept deliberately: the `container.querySelector('svg')` structural assertion now earns its place, because the AC-4 icon test queries the icon slot directly.

## Environment note worth recording

Verification in this worktree had been unreliable, and I found the cause: its `node_modules` was **empty**, so `platform-bible-react` resolved from the sibling checkout on a different branch — which has no `Empty` in its source at all. That single gap explains the failing repo-wide lint, the failing library build, and the 17 "pre-existing" test failures reported in earlier rounds. After `npm install` in the worktree, **all 17 pass**.

Current: **1650 tests / 111 files pass, ESLint clean, zero typecheck errors across all 19 changed files.**

Rebasing onto main next — it has advanced since the last rebase.

*AI-assisted.*

captaincrazybro:
## PT-4339 coordination gate — checked, and effectively closed

This was the one item both review rounds called blocking, on the theory that reactive `editable` could change which loading/empty-state message is correct here. I checked it rather than waiting on it, and it turns out the question resolved itself.

**#2682 merged on 2026-08-24 and is already in this branch.** I verified Tom's commit is a git ancestor of this HEAD — the last rebase pulled it in. So this stopped being "check before finalizing" and became "verify the integration."

**The two PRs don't interact.**

| | |
|---|---|
| File overlap | Exactly one — `localizedStrings.json` |
| #2682's surface | Scripture editor (`platform-scripture-editor.web-view.tsx`, `.utils.ts`, `main.ts`, C# edit permissions) — NN-5b |
| This PR's surface | Model Text / Resource panels — NN-5c |
| Editability signals in this PR's changed code | **Zero** — no `isEditable`, `isReadOnly`, `canUserEdit`, or `platform.isEditable` anywhere |

#2682's localized-string change is the editor's **title indicator** (`(Read-only)` / `(editable)`), not a loading or empty-state message.

That independence is deliberate rather than an oversight: these panels' only write is to **user-level** settings (`setUserModelTexts`) — the admin write path moved to the Share Layout dialog. Picking a reference resource is view configuration, not project editing, so project editability shouldn't gate it. If anyone disagrees with that reading, this is the place to say so, because it is the assumption the gate rests on.

**The shared file merged cleanly, and I verified rather than assumed.** Comparing `localizedStrings.json` key-by-key against `origin/main`: nothing dropped, no value altered, exactly the 6 new keys added per locale — including #2682's Spanish `(editable)` casing fix, which a hand-resolved conflict could easily have clobbered (this file conflicted on both rebases and I resolved it semantically both times).

I'd treat the gate as satisfied. A one-line confirmation from @tombogle still costs nothing if you'd rather have it explicit.

---

## One finding worth raising beyond either PR

#2682 hit the same underlying defect this PR did. From its code:

> `canUserEditScriptureLoading` above can never become `false` … the hook's `isLoading` latches `true` forever (a `usePromise`/`useProjectDataProvider` limitation)

That is exactly the bug behind the unrecoverable-spinner fix here. Two PRs in the same sprint independently discovered that `usePromise` can latch loading forever, and **both worked around it locally** — #2682 by distinguishing "will never resolve" from "loading", this PR by catching the rejection inside `useDblResourceCatalog`. Neither fixed the shared hook in `platform-bible-react`.

Two independent local workarounds for one shared defect is a strong signal it belongs fixed at the source. It also pairs with a second bug found while investigating this: `getCachedResources` in `platform-get-resources` has a **dead catch** —

```ts
try {
  return fetchAndCacheResources();   // no `await`
} catch (e) { … return undefined; }
```

`fetchAndCacheResources` is `async`, so a rejection escapes the `try` entirely and the handler never runs. That is *why* `getCachedResources` rejects at all, and therefore why the spinner bug was reachable in production rather than theoretical. Both are outside this PR's scope; suggest one ticket covering the pair.

---

With this closed, the remaining open items are the two scope sign-offs already listed in the description — whether `catalogError` is in scope, and whether `useBufferedLayoutSetting`'s third tuple element should be split out. Neither blocks on anyone but the author.

*AI-assisted.*

jolierabideau (approved):
Thanks Levi!

> @jolierabideau partially reviewed 18 files and all commit messages, made 1 comment, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2697#-:-P-t1xfYDAAYDVTc6-HV:b783vab)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2698, opened 2026-08-19
PT-4218: Fit the Simple-mode title bar at the minimum window width

captaincrazybro:
# Code Review Summary

**Branch**: worktree-pt-4218-simple-titlebar-min-width

**Base**: origin/main

**Date**: 2026-08-19

**Files changed**: 16 (3 source files + 13 regenerated `lib/platform-bible-react/dist/**` bundles)

## Overview

PT-4218: in Simple mode, dragging the window to its enforced 800px minimum clipped the trailing
title-bar controls — the user menu and part of the BCV control were cut off. The root cause is that
the `Toolbar`'s content area had no `min-w-0`, so as a flex item it floored at its content's
intrinsic width and could not shrink; the row's `overflow-hidden` then clipped the overflow instead.
The fix adds the missing `min-w-0`, declares a `@container/toolbar` container-query context on the
element whose content box is the bar's genuinely usable width, and adds a Simple-mode-only
degradation ladder that sheds the marketing version badge, then the project's full name, then the
sync label as the bar narrows.

The thresholds were tuned against measurements taken in the running app rather than calculated. At
the 800px minimum the usable bar is 628px or 642px on Windows (depending on whether the static
caption-button class or the live overlay measurement applies), 697px on macOS, and 766px on Linux.
An earlier 40rem/640px threshold fell *between* the two Windows numbers, meaning the same window on
the same machine could behave differently from run to run; the shipped 50rem clears every row. A
new geometric e2e spec covers the behavior, and the spec is explicit about what it cannot prove: the
crowded-bar case needs the version badge and sync button, which only render in Paratext 10 Studio,
so the headline overflow assertion cannot fail in a core build. A direct assertion on the content
area's computed `min-width` covers the root cause instead, and was verified to fail when the fix is
reverted.

Screenshot of narrow width window:
image

## API Changes

- `lib/platform-bible-react` — `Toolbar`: root element now declares a **named container-query
  context**, `tw:@container/toolbar` (verified in the shipped bundle as
  `.tw\:\@container\/toolbar{container:toolbar/inline-size}`). No prop or type change.
- `lib/platform-bible-react` — `Toolbar`: the **content area** flex item (the one that receives
  `children`) gained `tw:min-w-0`. The configure area already had `min-w-0`; the app-menu area
  deliberately still does not.
- `lib/platform-bible-react` — `Toolbar`: added two DOM test hooks,
  `data-testid="toolbar-content-row"` (the `overflow-hidden` row) and
  `data-testid="toolbar-content-area"`.
- `lib/platform-bible-react` — `Toolbar`: TSDoc expanded to document the container contract, the CSS
  containment consequences, and the two test hooks. This reaches `dist/index.d.ts`, IntelliSense,
  and the published TypeDoc.
- `lib/platform-bible-react` — no exports added, removed, or renamed. `ToolbarProps` and
  `getToolbarOSReservedSpaceClassName` are byte-identical to the merge base.
- `lib/papi-dts/papi.d.ts`, `lib/platform-bible-utils`, `extensions/**` — unchanged (not in the diff).
- `lib/platform-bible-react/dist/**` — regenerated; the only structural change is a rollup
  chunk-hash rename (`resizable-C3AVzTx2.js` → `resizable-aPuquhT0.js`, 98% similarity, plus the
  matching `.cjs`/`.map` pairs). Consistent with a clean rebuild, not a hand edit.

None. All four analysis passes independently reported zero critical findings.

- [x] **The sync button lost its accessible name below 46rem.** Flagged independently by all four
      passes. `HIDE_SYNC_LABEL_BELOW_46REM` is `display: none`, which removes the label from the
      accessibility tree; the button had no `aria-label`, the icons contribute no text, and the
      Radix tooltip supplies only `aria-describedby` (a description, not a name) and only while
      open. Introduced by this change — the label was previously always visible. Contradicted two
      in-repo precedents (`PlatformTabTitle` reflects the title into `aria-label` when a tab
      collapses to icon-only; `DataTablePagination` keeps `sr-only` labels).
      _(fixed during review: hoisted the state→label map into `syncButtonLabel` and applied it as
      the `Button`'s `aria-label`, so the accessible name is present at every width, in all three
      sync states, and always matches the visible label)_
- [x] **The new `@container/toolbar` contract was undocumented on an exported component.**
      Consumers are expected to target it with `tw:@max-[…]/toolbar:*`, but the explanation lived
      only in a block comment inside the JSX, reaching nobody using IntelliSense or TypeDoc —
      against the Code-Style-Guide's API-surface TSDoc rule.
      _(fixed during review: added TSDoc naming the container, explaining what its content box
      measures and why, with an `@example`; verified it reaches `dist/index.d.ts`)_
- [x] **The containment comment named only one of three consequences.** It argued `contain: layout`
      is safe because the root is already `position: relative`, which is true for absolutely
      positioned descendants but omits that containment also makes the root a containing block for
      `position: fixed` descendants and establishes a new stacking context.
      _(fixed during review: comment extended to name both, with the in-repo evidence that neither
      is currently reachable — no `fixed` positioning in the toolbar tree, and every overlay it can
      open portals to `document.body`)_
- [x] **Inline-size containment is an un-gated behavior change to a published component.** The
      root's inline size is now computed as if it had no contents, so an external consumer
      rendering `Toolbar` in a shrink-to-fit context (`inline-block`, a `w-auto` flex item, an
      `auto` grid track) would see it collapse. All three in-repo consumers render it full-width,
      so nothing regresses here.
      _(fixed during review: documented in the `Toolbar` TSDoc with explicit guidance to render it
      as a full-width block; the behavior change itself stands and is worth a release note)_
- [x] **Orphaned JSDoc in the e2e spec.** `setWindowWidth`'s doc block sat above `closeDevTools`'s
      own doc block, so `closeDevTools` had a misleading doc pair and `setWindowWidth` was
      undocumented in editors. Caused by inserting the DevTools helper between the doc and its
      function.
      _(fixed during review: moved the block onto `setWindowWidth`)_
- [ ] ~~The marketing version becomes unreachable once tier 1 hides the badge.~~ The responsiveness
      guideline requires hidden content stay reachable another way; hiding the badge removes its
      tooltip too. The About dialog is not a fallback — it renders `packageInfo.version` /
      `AppInfo.version` (semver), while the badge shows `marketingVersion` +
      `marketingVersionMoniker` from a distinct `MarketingInfo` type.
      _(Author: the marketing version isn't really visible in the main app anyway, so losing it at
      narrow widths is acceptable. Note it is non-empty only in Paratext 10 Studio, where PT-4218
      was reported.)_

[Author response: five of the six were fixed during the interview. The author dismissed the
marketing-version finding on the grounds that the value is not surfaced in the main app. The author
also asked a substantive design question about whether container queries can account for differing
per-OS minimum widths — see Interview Notes.]

### Minor — Consider

- [x] **Tier numbers contradicted the order the tiers fire in.** By threshold the ladder degrades
      badge (52rem) → project name (50rem) → sync label (46rem), but the constants labelled the
      sync label "Tier 2" and the project name "Tier 3". Introduced when the project-name threshold
      was retuned from 40rem to 50rem without renumbering.
      _(fixed during review: renumbered so the numbers match the firing order, updated the two
      cross-references and the e2e spec's `tier-2` mention, and added a note that the numbering is
      by firing order and must be revisited when a threshold is retuned past its neighbour)_
- [x] **`if (!win) return 0` in the resize helper degraded the failure message.** A missing window
      produced `settledWidth = 0`, so the poll burned its full 20-second timeout and then failed
      with a width mismatch rather than the real cause.
      _(fixed during review: throws `No Electron window to resize` instead, with a comment
      explaining why a sentinel is wrong here)_
- [x] **Tiers 1 and 3 had no automated coverage** — only the project-name tier was exercised, by
      the e2e swap test.
      _(fixed during review: added four unit tests pinning the ladder wiring — the badge carries a
      hide variant in Simple mode and does not in Power mode, the sync label carries one, the idle
      stand-in icon carries the paired reveal variant, and the sync button keeps its accessible
      name. Matched by shape rather than exact rem value so retuning a threshold does not break
      them. Falsifiability verified: three of the four fail when the wiring is removed, and the
      Power-mode assertion is the complementary half of the badge pair)_
- [x] **The unit tests did not pin the new idle `RefreshCw` icon or the label span** — 32 tests
      passed and none would have noticed either being dropped.
      _(fixed during review: covered by the same four tests added above)_
- [x] **The project selector's floor was below the width its own comment cited.** `min-w-20` is
      80px, while the comment states the short name needs ~97px; after the trigger's padding and
      chevron under ~40px was left for text, so the very name tier 2 swaps in could itself
      ellipsize with nothing to recover it.
      _(fixed during review: raised to `min-w-24` (96px) and rewrote the comment to state the
      measured basis. NOTE: this is the one change not re-measured against the running app at the
      tightest Windows budget — see Suggested Review Focus)_
- [x] **The two `data-testid` values were an undocumented public contract**, consumed from
      `e2e-tests/` outside the package, so a rename would be breaking rather than cosmetic.
      _(fixed during review: documented in the `Toolbar` TSDoc)_
- [ ] **`@container/toolbar` is declared by two exported components in the same library** —
      `Toolbar` and `TabToolbarContainer` (`tab-toolbar-container.component.tsx:59`, pre-existing).
      Container-name lookup resolves to the nearest matching ancestor, so a shared child written
      against `/toolbar` binds to whichever is closer. Verified not currently reachable: the dock
      layout's tab toolbar is a sibling of the title bar, never an ancestor, and no PBR source uses
      a `/toolbar:` variant. A distinct name (`@container/app-toolbar`) would remove the ambiguity
      for extension authors, who see both as public API. _(Author chose to skip; touches a
      published CSS contract and needs a lib rebuild.)_
- [ ] **Thresholds are encoded in the constant names** (`HIDE_VERSION_BADGE_BELOW_52REM`), so
      retuning forces a rename at every use site or leaves a name that lies — a friction the author
      already hit going 40rem → 45rem → 50rem. Intent-based names with the value in TSDoc would
      avoid it. _(Author chose to skip.)_
- [ ] **The idle sync icon is revealed only below 46rem**, so at wide widths `idle` is text-only
      while `syncing`/`synced` are icon-plus-text. The responsiveness guideline says to use icon
      and text when width allows; showing `RefreshCw` at all widths would also stop the button
      jumping wider when a sync starts. _(Author chose to skip; the TSDoc explicitly scopes this
      out as beyond the ticket.)_
- [ ] **`PARATEXT_PDPF_METHOD` is a third copy** of that literal, alongside
      `e2e-tests/fixtures/comment-test-helpers.ts:59` and an inlined copy in
      `scroll-group-sync.spec.ts:71`. Exporting it once from `fixtures/helpers.ts` would keep them
      in sync. _(Author chose to skip; touches shared fixtures beyond this ticket.)_
- [ ] **Only tier 1 is explicitly `!isPowerMode`-gated**; tiers 2 and 3 are Simple-mode-only merely
      because their host elements render inside `!isPowerMode` blocks. Correct as written, but a
      half-line note would stop the next reader "fixing" the inconsistency. _(Author chose to skip.)_
- [ ] **The `isolated` Playwright project is local-only** — CI runs `test:e2e:smoke` only, so this
      regression is not guarded per-PR. Matches the documented convention; noted so nobody assumes
      otherwise. _(Informational.)_
- [ ] **The Power-mode Home button also lacks an `aria-label`** — the same gap as the sync button,
      but pre-existing and not introduced by this diff. Cheap to fix in a later pass using
      `localizedStrings['%mainMenu_openHome%']`. _(Author chose to skip as out of scope.)_

[Author response: the author selected the tier renumbering, the resize-helper failure message, the
missing tier coverage, and the selector floor for fixing, and explicitly skipped the rest. The
data-testid documentation was already resolved by the TSDoc work done for the Important findings.]

#### Shared Regions Modified

None. No `#region shared with` markers exist in any of the changed source files. (The
`// #region Narrow-title-bar degradation ladder (PT-4218)` marker in `platform-bible-toolbar.tsx` is
a plain editor-folding region, not a template-shared region.)

#### Extension Config Changes

None. No files under `extensions/` were changed, and no build or tooling file changed at all — no
`package.json`, `tsconfig*.json`, `.eslintrc*`, `webpack.config.*`, `.erb/configs/`, or
`.github/workflows/`.

## Positive Observations

- The `min-w-0` fix is narrower and better targeted than it first appears: the configure area
  already carried `min-w-0` at the merge base, so the content area was the inconsistent one. The
  comment also explains why the app-menu area is deliberately excluded — exactly the question a
  reviewer asks next.
- The container context is placed on the element whose *content box* is the genuinely usable bar
  width, and the comment enumerates all three caption-button reservation paths (static macOS,
  static Windows, live-measured Windows) to justify the placement.
- The measured per-platform threshold table, and the explicit record of why 40rem and 45rem were
  rejected, is the kind of design-decision capture that normally evaporates and causes a later
  "simplification" to reintroduce the bug.
- Container queries follow existing in-repo precedent (`search-bar.component.tsx`) rather than
  inventing a mechanism, and are the right tool here: they respond to real available space, so they
  stay correct across the three OS reservations and in RTL where a viewport breakpoint would not.
- Cross-package Tailwind wiring was verified in both halves: the library build emits the container
  context into `dist`, and `src/renderer/styles/tailwind.css` scans `src/renderer/**` so the
  `@max-[…]/toolbar:` variants used by the app are emitted by the app build.
- `cn` merge safety checks out — every ladder class carries a container-query modifier while the
  base classes do not, so `tailwind-merge` treats them as non-conflicting and keeps both.
- The "Simple-mode-only" claim holds under inspection: the sync button and project selector already
  render inside `!isPowerMode` guards, and the version badge — the one control rendering in both
  modes — carries an explicit `!isPowerMode &&` on its degradation class. No Power-mode regression.
- The committed `dist/` rebuild is coherent and minimal, with only the intended class/attribute
  deltas plus a chunk-hash rename that git detects as a 98%-similarity rename — genuinely rebuilt,
  not hand-patched.
- The e2e spec is unusually candid about its own limits rather than overclaiming, and documents why
  the document-level horizontal scrollbar is deliberately *not* asserted here.
- No new lint suppressions, `TODO`s, `console.*`, or type assertions were introduced, and all
  changed files are Prettier-clean.

## Interview Notes

**Stated purpose**: fit the Simple-mode title bar at the app's enforced 800px minimum window width
(PT-4218).

**Authorship — important context for the reviewer**: the implementation on this branch was written
by Claude (Opus 5), not hand-authored. During this review the author's role was directing which
findings to fix and which to skip; the author did not write, and was not asked to explain, the
underlying code. The reviewer should not assume line-level familiarity, and may want to walk the
container-query mechanics and the threshold table in the meeting.

**Design question raised by the author**: whether container queries can factor in the differing
per-OS minimum widths. The answer recorded during the interview:

- They already do, and that is the reason they were chosen. The caption-button reservation lands as
  padding on the container (or an ancestor), so the container's content box is the platform-adjusted
  usable width. A threshold expressed in container units is therefore already normalized across
  OSes; a viewport breakpoint could not do this.
- They cannot vary the threshold *by* OS, and should not — the container has already subtracted the
  reservation, so an OS-specific threshold on top would subtract it twice.
- The consequence is that one threshold fires at different *window* widths per OS (tier 2 at ≤834px
  on Linux, ≤972px on Windows). That is correct behavior, not over-triggering: Windows at 900px
  genuinely has ~728px of usable bar versus Linux's 866px.
- The real asymmetry lives in `minWidth`, not in the container queries: `src/main/main.ts:649` sets
  `minWidth: 800` with no platform branch — three lines above a `titleBarStyle` that *is* platform
  branched — so Windows users get ~138px less usable title bar than Linux users at the same enforced
  minimum. Making `minWidth` platform-aware would give every OS the same usable bar and let one
  threshold fire at the same window width everywhere. The author agreed to record this as a
  follow-up against PT-2333 (cited in that TODO) / PT-2480.

**Known limitation, deliberately accepted**: Windows has two caption-button reservation paths — the
static `pe-[calc(138px+1rem)]` class and the live overlay measurement — producing 628px vs 642px of
usable bar at the same window on the same machine. No container query can paper over this; 50rem
clears both only by carrying enough margin. Making those two paths agree is the real fix and is out
of scope here.

**Unresolved items**: none. Every Critical and Important finding was either fixed during the review
or dismissed with a stated reason.

## In-Review Quality Check

Nine changes were made during the interview, across `platform-bible-toolbar.tsx`,
`platform-bible-toolbar.test.tsx`, `toolbar.component.tsx`, and the e2e spec.

- **Prettier** — reformatted the expanded `Toolbar` TSDoc. Because `dist/` is committed in this
  repo and `dist/index.d.ts` carries TSDoc, `lib/platform-bible-react` was rebuilt afterwards so the
  committed bundles match source.
- **`npm run lint`** — exit 0, zero errors. Remaining output is pre-existing warnings in untouched
  files.
- **`npm test`** — 3,569 tests across 5 workspaces, all passing (1 skipped, pre-existing). The
  toolbar suite went from 32 to 36 tests.
- **Falsifiability check** — the four new tests were verified by deliberately removing the badge
  gate, the `aria-label`, and the sync-label hide class: three of the four failed as intended, and
  the fourth is the complementary half of the badge pair. Wiring restored and re-verified after.
- **`npm run test:e2e:isolated title-bar`** — 3 passed, 2 skipped (the win32-only reserved-space
  spec). Run before starting the app, since the harness refuses to start while port 8876 is held.
- **Live verification** — the app was started from this worktree and the fixes confirmed in the
  running renderer: content area `min-width: 0px`, sync button `aria-label="Sync"`, selector floor
  `96px`.

`npm run typecheck` was not run, per the author's standing preference that lint plus build are
sufficient. No unfixable failures.

## Suggested Review Focus

- [ ] **The `min-w-24` (96px) selector floor is the one unverified change.** It was raised from 80px
      during this review to match the ~97px the short name needs, but has not been re-measured at
      the tightest Windows budget (628px usable bar with the version badge and sync button present).
      Confirming it in Paratext 10 Studio would close the loop.
- [ ] **Whether the degradation ladder's ordering is the right UX call.** Badge → project full name
      → sync label is a judgement about what a user can most afford to lose, made without a UX spec:
      the linked PT-2480 "Design minimum screen width of the app" is still in Backlog. Worth a
      designer's eye before this sets precedent.
- [ ] **The container-name collision** (`@container/toolbar` on both `Toolbar` and
      `TabToolbarContainer`). Skipped during review as not currently reachable, but it is a latent
      footgun for extension authors and a rename gets cheaper the sooner it happens.
- [ ] **Per-OS `minWidth` as the more principled fix** — see Interview Notes. The bar being tighter
      on Windows than on Linux at the same enforced minimum is the root asymmetry this ticket works
      around rather than solves. Follow-up against PT-2333 / PT-2480.
- [ ] **Author does not have line-level familiarity with the implementation** — it was AI-generated
      and the author's involvement was directing review decisions. Walking the container-query
      mechanics, the measured threshold table, and why the e2e tests cannot fail in a core build
      would be the most useful use of the meeting.
- [ ] **Sebastian's reported horizontal scrollbar is NOT fixed by this branch**, and the PR should
      not be read as closing it. Measured at the 800px minimum the document still overflows by
      ~108px, sourced from `simple-layout.data.ts` locking each of Simple mode's three columns to
      `panelLock: { minWidth: 300 }` — rc-dock then computes a 908px minimum no window at
      `minWidth: 800` can satisfy. Deserves its own ticket under PT-2480.

AI-assisted — [session](https://claude.ai/code/9cfa1786-e02b-40d4-a79e-022115f360b5)

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2698)


jolierabideau:
## Review — PT-4218 fit against the epic, and pattern conformance

Reviewed as implementation lead for PT-4336 (*Saroj and Donna can trust the app*), focusing on (1) how this solves PT-4218 / NN-2b and (2) whether it matches how the rest of the codebase does this.

The root-cause diagnosis is right and the write-up is genuinely excellent — the measured per-platform table, the rejected 40rem/45rem values, and the e2e spec's candour about what it can't prove are all better than what usually lands here. My concerns are almost entirely about **which mechanism** this uses, not about correctness.

### 1. Blocking: this collides head-on with #2701 (PT-4344)

[#2701](https://github.com/paranext/paranext-core/pull/2701) (Katherine, opened a day after this one) is the sibling half of NN-2 and it lands the *general* solution for the same problem. It touches the same two files this PR touches — `lib/platform-bible-react/src/components/advanced/toolbar.component.tsx` and `src/renderer/components/platform-bible-toolbar.tsx` — and it rewrites the same three controls. These cannot both merge as-is, and it isn't just a textual conflict:

- **#2701 explicitly rules out this mechanism.** Its update to `responsiveness.mdx` says: *"Determine available width with the `useShrinkStep` hook rather than a CSS container query. Tailwind's container-query variants have repeatedly failed to be emitted into extension web view bundles, and they fail silently."* It also strips an existing `tw:@sm/toolbar:w-44` out of `platform-enhanced-resources` for that reason. This PR's entire ladder is Tailwind container-query variants, and its new `Toolbar` TSDoc advertises `@max-[…]/toolbar:` as the supported contract for consumers.
- **#2701 rules out hiding items.** Same file: *"Toolbar items shrink and truncate. They are never hidden, never moved into an overflow menu, and never clipped away — at any width the app supports, every item is still there."* Tier 1 here removes the version badge outright. That's also the wording in the PRD's own NN-2a and in the epic problem statement ("*a toolbar button he used yesterday is simply gone because his window is narrow*"), so it reads as the thing we're fixing rather than a fix.
- **#2701 moves the floor this PR is measured against.** It raises `main.ts` `minWidth` from 800 → 900 (UX analytics: 99.83% of 11,587 users on ≥900px) and derives `SIMPLE_COLUMN_MIN_WIDTH_PX = 297` so Simple's three columns fit. That's the horizontal scrollbar Sebastian reported in the PT-4218 comment — which this PR's description correctly identifies but scopes out. Once 900 lands, the whole "must clear Linux's 766px at an 800px window" derivation in the comment block is stale even though the thresholds still fire.
- **#2701 already lands the same root-cause fix.** It adds `tw:min-w-0` to the absorbing zones and `tw:shrink-0` to the rigid ones, and documents that split as the convention. That's this PR's core one-line insight, generalized.

**Suggested path:** don't merge this on top of #2701 or vice versa. I'd land #2701's mechanism first, then rebase this onto it and keep the parts that are unambiguously additive and still needed:

- the sync button `aria-label` (real a11y fix, wanted either way — see below);
- the project-selector floor change, re-expressed against `useShrinkStep` / `ToolbarCompoundLabel`;
- the geometric e2e spec, which is mechanism-agnostic and is the most valuable artifact here — worth keeping almost verbatim.

Worth a 15-minute call between the two of you before either merges; happy to sit in.

### 2. The degradation thresholds fire well above the minimum width

Taking the PR's own numbers (Windows: 800px window → 628px container, so ~172px of chrome), tier 1 at 52rem fires below a **~1004px window** on Windows, and tier 2 at 50rem below **~972px**. So on a perfectly ordinary 1000px window a Windows user loses the version badge and the project's full name — nowhere near the "dragged to the minimum" case PT-4218 describes. NN-2b asks for no cut-off controls at minimum width; this trades that for degraded controls across a wide band of normal widths. #2701's stepped-shrink model (`Genesis 1:1` → `GEN 1:1` → `GEN 1:…` → `GEN`) degrades much more gradually for the same reason.

### 3. Against the documented design spec

Alex Mercado's shrinkability range (captured in the epic brief, 2026-08-12) is: icon + full label + details → icon + full label → icon + abbreviated label → icon-only. The sync button here goes **text-only (wide) → icon-only (narrow)**, skipping both middle steps and starting from a state with no icon at all. The TSDoc scopes that out as beyond the ticket, which is fair, but it means the one control this PR reshapes is the one that doesn't follow the spec. Also note the epic's Open Questions record *"NN 2 — Is there a design or guideline document written for how items should be shrunk/truncated? → **Yes, documented in issue**"*, so the PR description's "made without a UX spec" isn't quite right — the spec exists, it just lives in the epic rather than PT-4218.

### 4. Pattern conformance — the good and the gaps

Good:
- `min-w-0` on the content area is the correct root-cause fix and matches the established flex-truncation idiom (`scope-selector`, `project-selector`, `multi-select-combo-box`).
- The comment explaining why the app-menu area is *excluded* answers the next reviewer's question before they ask it. #2701 arrives at the identical rule independently, which is a good sign the diagnosis is right.
- The `aria-label` hoist is exactly right, and matches `PlatformTabTitle`'s precedent. Please keep this regardless of which mechanism wins.
- The e2e spec's `closeDevTools` and read-before-`setSize` handling are the kind of thing that would otherwise flake for a week.

Gaps:
- **`responsiveness.mdx` is not updated.** The epic brief flags this as a hard gate for WI-6: *"the implementation must reconcile with the guideline before the PR can be approved."* The guideline as it stands on `main` still says toolbars *hide* items and hidden items must be reachable from a menu — which this PR half-follows (it hides) and half-violates (no fallback). #2701 does update it. Either way one of the two PRs has to.
- **No ADR.** Introducing container queries as the sanctioned title-bar responsiveness mechanism is exactly the kind of cross-cutting choice `CLAUDE.md` asks to be recorded in `Architecture-Decisions.md`. #2701 includes one; that's the tiebreaker artifact.
- **`@container/toolbar` name collision** with `TabToolbarContainer` — already noted in the description and skipped. I'd rather not ship a second declaration of that name on a published component; if the container-query route survives at all, rename to `app-toolbar`.
- **The new unit tests assert class-name strings.** `expect(badge.className).toMatch(/@max-\[\d+rem\]\/toolbar:hidden/)` pins the mechanism, not the behavior — it can't fail for the bug the ticket describes, and it will have to be deleted wholesale if the mechanism changes. The regex-by-shape trick is a thoughtful mitigation, but per the Testing-Guide this is the implementation-mirroring shape we try to avoid. The geometric e2e assertions are doing the real work here.

### 5. Smaller things

- Tier 1 gates on `!isPowerMode`, tiers 2 and 3 rely on their host elements already sitting inside `!isPowerMode` blocks. Correct, but please add the half-line note — as written it invites someone to "fix" the inconsistency.
- Threshold values baked into constant names (`HIDE_VERSION_BADGE_BELOW_52REM`) already cost a rename twice during development. Intent-based name + value in TSDoc.
- `PARATEXT_PDPF_METHOD` is now a third copy of that literal (`comment-test-helpers.ts:59`, `scroll-group-sync.spec.ts:71`). Small, but it's the kind of drift that bites later — worth exporting once from `fixtures/helpers.ts`.
- The `isolated` Playwright project doesn't run in CI (`test:e2e:smoke` only), so this regression isn't guarded per-PR. Matches convention; just flagging that the coverage is local-only.

### Verdict

Requesting changes, but the substance is on **coordination with #2701**, not on the quality of this work. The diagnosis, the measurement discipline, the a11y catch, and the e2e spec are all things I want to keep. Once we've picked one mechanism for NN-2, most of this PR either survives as-is or ports in a few lines.

captaincrazybro:
Thanks Jolie — this was the right call and I've acted on it. The blocking finding was correct: I checked #2701's diff and it does change the app title bar (same two files), it defines `APP_TOOLBAR_SHRINK_THRESHOLDS_PX` for the titlebar specifically, and it reimplements the project-name degradation via `ToolbarCompoundLabel`. So this wasn't a textual conflict.

**What changed (52bd441046a): the container-query ladder is gone.**

Rather than argue with ADR-0016, I removed the mechanism it rejects — the `@container/toolbar` context, the TSDoc advertising `@max-[…]/toolbar:` as a consumer contract, all five ladder constants, the version-badge hide, the sync-label collapse, and the project short-name swap. The unit tests asserting those class names went with it, as did the e2e test for the swap.

That resolves most of your list as a side effect: no ADR needed from this PR (adopting an existing accepted one rather than adding a competing entry), no `responsiveness.mdx` update needed (nothing hides any more), the `@container/toolbar` name collision is gone, and the class-name-asserting tests you flagged as implementation-mirroring are gone. `PARATEXT_PDPF_METHOD` went too — it only existed for the swap test. The threshold-in-constant-names and tier-numbering points are moot with no tiers left.

**What remains** is ~90 lines: `min-w-0` on the Toolbar's content area, a smaller project-selector floor (192px → 96px), the two `data-testid` hooks, the sync button `aria-label`, and two geometric e2e tests.

**One measured finding you should have before this merges.** I re-ran my Studio-on-Windows simulation after removing the ladder — version badge and labelled sync button present, Windows caption reservation applied. At an 800px window it still overflows by **28px**, and `min-width: 0` measured *identical* to `min-width: auto` there. It only clears at a 972px window.

So I had the credit backwards in my original write-up. `min-w-0` is the correct root-cause fix and a precondition for any shrink-based approach, but the clipping in the reported configuration is in the **config** area — badge, sync button, profile — which already had `min-w-0`. The ladder was doing the real work by reclaiming ~190px of that. **This PR alone therefore does not fix PT-4218 for the environment it was reported against**, and the ticket should ride on #2701's abbreviation work rather than closing on this one. Worth noting even 900px still left 12px in my simulation, so the width bump alone isn't sufficient either — the abbreviation is load-bearing.

Caveat on those numbers: it's a simulation. `marketingVersion` is empty in core and send/receive was unavailable, so I injected an approximate badge and sync button and applied the Windows padding while running on Linux. I trust the direction and rough magnitude, not the exact pixel count on real Windows.

**On sequencing** — Katherine and I compared diffs and she's proposed merging this one first, then rebasing hers on top, carrying forward the `aria-label` and the two testids (her restructure would otherwise break the e2e spec) and re-tuning the spec for the 900px minimum. That re-tuning should be small: the spec asks Electron for the clamped minimum rather than hardcoding 800, so both remaining tests are threshold-independent.

Two of your smaller points I've deliberately left open: the `isolated` Playwright project still doesn't run in CI, so this regression is only guarded locally; and I haven't touched `responsiveness.mdx`, since #2701 owns that file now.

Happy to take the call if it's still useful, though I think the coordination question is settled.

jolierabideau (changes requested):
Thanks for addressing the feedback — the removal of the container-query ladder in `52bd4410` resolved everything I was blocking on as a side effect, and the honest disclosure about what the PR now does and doesn't fix is exactly the right call.

One thing to fix before I approve, and one NIT:

---

**Blocking — stale `@example` in Toolbar TSDoc**

The commit message for `52bd4410` says it removed "the TSDoc that advertised `@max-[…]/toolbar:` as a supported contract for consumers", but this `@example` survived:

```tsx
// Hide a child once the usable bar width drops below 52rem:
<Toolbar {...props}>
  <Badge className="tw:@max-[52rem]/toolbar:hidden">{version}</Badge>
</Toolbar>
```

`@container/toolbar` was correctly removed from the component, but the example still uses `@max-[52rem]/toolbar:hidden` — a container-query variant that requires that context. Since this lands in `dist/index.d.ts`, an extension author reading it via IntelliSense would try it and it would silently do nothing. Please remove or replace the `@example` block.

---

**NIT — `min-w-24` floor unverified at tightest Windows budget**

This was already in the Suggested Review Focus from the first review pass: the selector floor was raised from 80px → 96px during review, but has not been re-measured in Studio on Windows at the tightest budget (628px usable bar, badge and sync button present). Not blocking — the direction is right and the original 80px was clearly wrong — but worth a note in the PR if you're not going to verify it before merge, so the next person knows.

captaincrazybro:
Both addressed — and thanks for catching the `@example`, that one deserved to be caught.

**Blocking: stale `@example` — fixed in 37e0f7ff.**

You were right that it survived, and my commit message claiming otherwise was wrong. Worth recording the cause, because it's the same failure class as the mechanism we just removed: that block was the one string replacement in the strip I didn't assert on. Prettier had reformatted `</Toolbar>` to `</Toolbar>;`, so my exact-match silently no-opped while every other edit in that commit landed. A silent no-op in a script removing a mechanism rejected for silent no-ops.

The `@example` is gone rather than replaced — there's no container-query usage left to demonstrate — and I've confirmed no `@max-`, `@container`, or `52rem` text remains anywhere in the component. The remaining TSDoc documents only the two `data-testid` hooks.

**NIT: `min-w-24` at the tightest Windows budget — measured, and the answer is that it never engages.**

I'd already run exactly that scenario in the sweep after removing the ladder: usable bar at 628px, version badge and labelled sync button present, Windows caption reservation applied. The selector trigger rendered at **256px** — its `max-w-64` cap — so the 96px floor is nowhere near being reached.

The reason it doesn't compress is the same one behind the 28px I reported: the deficit is in the **config** area, not the content area. The trigger is `w-fit`, and flex isn't shrinking the content area because that area isn't the constrained one. So the floor sits inactive at that width.

Which means the honest read is that the 80px → 96px change is **inert in the case you were worried about**. It's a correctness fix — 80px contradicted the comment right above it claiming the short name needs ~97px — but it does no work at the tightest Windows budget.

Worth adding that #2701 supersedes this line anyway, and lands on the same number independently. Katherine replaces the static floor with a step-dependent one:

```tsx
shrinkStep >= SHRINK_STEP.MINIMUM ? 'tw:min-w-24' : 'tw:min-w-48',
```

Same 96px, reached by her own reasoning, but applied only at the minimum step while keeping 192px when there's room — which is strictly better than mine, since I made 96px the floor at every width. So the value is corroborated by two independent arrivals, and the static form is transient: it disappears when she rebases.

Same caveat as before on the measurement: it's a simulation, since `marketingVersion` is empty in core and send/receive was unavailable, so I injected an approximate badge and sync button and applied the Windows padding while running on Linux. The toolbar source hasn't changed since I took it.

**Unrelated, but worth flagging:** the PR had gone `CONFLICTING` against main, which had moved 17 commits ahead. Every conflict was in `lib/platform-bible-react/dist/**` — build artifacts both sides regenerated — with no source conflict at all. Merged in f39a6d02: took main's bundles, cleared two content-hashed `resizable-*` chunks that collided on name, and rebuilt the library so `dist` matches the merged source. It's `MERGEABLE` again. Post-merge: 11 workspace suites green, lint clean, 33 toolbar unit tests passing.

jolierabideau (approved):
> @jolierabideau reviewed 11 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2698)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2701, opened 2026-08-20
PT-4344: toolbar items shrink and truncate instead of disappearing

katherinejensen00:
https://github.com/user-attachments/assets/6e3899cc-ba42-4aaf-932a-d02662db00d4

## Summary

Toolbar items no longer disappear at narrow widths. They shrink and truncate in defined steps, and a
tooltip carries the full text whenever what is rendered is not all of it.

The root cause was flexbox's `min-width: auto`. A flex child that lacks `tw:min-w-0` cannot shrink
below its content width, so the space comes out of its siblings instead — and the container's
`overflow-clip` hides what is pushed past the edge. Every toolbar region is now marked as one of two
kinds: **absorbing** (`tw:min-w-0`, may shrink) or **rigid** (`tw:shrink-0`, has no shorter form to
fall back to).

On top of that, items carrying two fields walk a four-step ladder driven by a measured width:
`Genesis 1:1` → `GEN 1:1` → `GEN 1:…` → `GEN`. The second field is always the truncation target, so
the field that identifies the item survives longest.

This also removes the horizontal scrollbar in Simple mode. Three "300px" columns actually demanded
902px inside a 900px window, because rc-dock adds 4px per divider in its own min-width arithmetic
(hard-coded, and unrelated to the 2px the Simple-mode stylesheet paints). Column minimums are now
derived from the window minimum instead of guessed alongside it.

Resolves [PT-4344](https://paratextstudio.atlassian.net/browse/PT-4344).

## Changes

**New in `platform-bible-react`**

- `useShrinkStep` — observes an element and reports a discrete step, with hysteresis so dragging a
  window edge across a threshold cannot flutter a label. Narrowing applies immediately; only
  widening waits out the band.
- `ShrinkStepContext` / `SHRINK_STEP` — a toolbar publishes its step; descendants read it. Defaults
  to the widest step rather than throwing, so every consumer still works standalone.
- `ToolbarCompoundLabel` — the two-field label. Handles the ellipsis, the drop, and the tooltip.

**Surfaces**

- Application titlebar and tab toolbar: absorbing/rigid zones, and each publishes a measured step.
- `BookChapterControl` walks the full ladder; the project selector and the paragraph-style label
  read a single boundary.
- `MarkerMenu` rows: marker code as code, detail as a trailing annotation, both truncating with the
  title weighted to win.
- `platform-enhanced-resources`' tab bar moves off a dead `tw-@container/toolbar` class — written
  with a dash, so Tailwind emitted nothing and the labels never collapsed at any width.

**Platform**

- Window `minWidth` raised to 900 on UX direction (2025 analytics: 99.83% of 11,587 users on a
  screen 900px or wider).
- `SIMPLE_COLUMN_MIN_WIDTH_PX` derived from that number, pinned by a test that budgets rc-dock's
  per-divider reserve.

**Docs**

- `Guidelines/Responsiveness` rewritten for the new toolbar policy — **signed off by UX**.
- `ADR-toolbar-shrink-measurement` (JS width measurement over container queries) and
  `ADR-simple-mode-column-minimums` (column minimums derived from the window minimum).

## ADR identifiers

The two entries this PR adds are identified by slug — `ADR-toolbar-shrink-measurement` and
`ADR-simple-mode-column-minimums` — rather than by the next sequential number.

Sequential numbering collided on every single rebase of this branch: three separate PRs claimed
the numbers these entries were written under (PT-3216 took 0026, PT-4347 took both 0027 and 0028),
so the entries have already been renumbered twice and their citations chased each time. PT-4218
also merged a cross-reference pointing at ADR-0016 — the shadcn `Empty` decision — because the
number it meant had moved out from under it. A slug is claimed when the entry is written and
never has to move; two branches picking the same slug would mean they describe the same
decision, which is worth catching rather than papering over.

The log's `How to use it` section is updated to match. **The 28 existing numbered entries are
left alone** — renumbering those would be a repo-wide change that collides with every other
branch in flight.

## Overlap with PT-4218

PT-4218 (#2698) merged while this was in review and fixed the same root cause from the other
side, so the rebase had to settle three things rather than just replay:

- **Its e2e contract is kept.** `toolbar-content-row`, `toolbar-content-area` and
  `toolbar-project-selector` are read by `title-bar-narrow-width.spec.ts` from outside the
  package. They survive the restructure into `ShrinkStepContext` and the extraction of
  `ProjectSelectorTrigger`, which is why testids appear in this diff.
- **Its static project-selector floor is superseded.** PT-4218 pinned `min-w-24` at every width;
  here the floor moves with the shrink step (`min-w-48` until the narrowest one). Its measurement
  (~97px for a short name like `ESVUS16`) is the basis for the 96px value and is recorded at that
  site.
- **Two of its comments are corrected.** Its ADR cross-reference read ADR-0016 (the shadcn `Empty`
  decision); the container-queries rationale it means is `ADR-toolbar-shrink-measurement`. Its
  "width-driven collapse is out of scope here" note described `useShrinkStep` as future work — it
  ships in that same file on this branch. The spec aside claiming the document still overflows at
  the minimum window width is likewise superseded by `ADR-simple-mode-column-minimums`.

## Please look closely at

Four things are deliberate but visible, and easier to agree with when stated than to discover in the
diff:

1. **The responsiveness guideline reverses its own toolbar policy.** It previously said toolbars
   "shrink and finally hide items… hidden items need to be accessible elsewhere." It now says items
   are never hidden. That reversal is the substance of this ticket, and UX has signed off.
2. **`MarkerMenu`'s row redesign reaches every consumer**, not only the paragraph switcher.
3. **The 900px minimum widens existing users' saved windows** on next launch.
4. **`test.fixme(GAP-024)` stays skipped.** The container-query gap it cited is closed by this PR,
   but nobody has watched it run against a real app, so it is not being enabled on a code reading
   alone. Its stated reason now says exactly that.

## Known limitations

- The shrink thresholds are estimated from the widths of the controls each toolbar carries, not
  measured. Both constants say so. They want one tuning pass in a running app — which also settles
  whether `SHRINK_STEP.MINIMUM` is reachable in the titlebar at all.
- The keyboard path is local to `ToolbarCompoundLabel`, not a design-system default. Focus lands on
  the enclosing button rather than the span the pointer handlers sit on, so the reveal is a native
  listener on that ancestor, gated on the last input having been a keyboard — otherwise Radix
  handing focus back when a menu closes would pop a tooltip with the pointer nowhere near it.
  `useTruncationTooltip` itself is untouched and still pointer-only, so its other pre-existing
  consumers do not get any of this. Lifting it into the hook is a follow-up.
- `tw:@container/toolbar` and `tw:@sm:basis-auto` survive on `TabToolbarContainer`, which reads
  oddly next to `ADR-toolbar-shrink-measurement`. Both predate this branch, and removing
  `@sm:basis-auto` would change how leftover space is split at wide widths — a layout change this
  ticket did not ask for.

## AI Involvement

AI-assisted throughout: planning, implementation, tests, and a four-pass review sweep (API surface,
style, standards compliance, UX) whose findings were worked through with the author before this PR.
The earlier sessions were not recorded; the rebases, the responses to review and the follow-up fixes
are [session 1](https://claude.ai/code/session_01HPJtGet1s6SY2kpCj5JBNn). Every claim in the review
was verified against the code rather than taken on the analyzer's word — two passes contradicted
each other on whether the e2e assertions would break. An earlier revision of this section said the
fixtures settled that in the assertions' favour; they do not. `isolated.fixture.ts` resizes window 1
only, so a window opened mid-test keeps the 1024px default and its titlebar lands below the widest
step on macOS and Windows. Those assertions now accept either label form. Author reviewed and
directed all decisions; the four callouts above are the author's calls, not the tooling's.

## Testing

- [x] `lib/platform-bible-react` unit suite — 512 tests, 49 files
- [x] `src/renderer` — `platform-bible-toolbar` (38) and `simple-layout.data` (12)
- [x] `platform-scripture-editor` and `platform-enhanced-resources` suites, including 4 new
      `EnhancedResourceTabBar` tests and 2 new paragraph-menu subtitle tests
- [x] `format:check`, `tsc`, `eslint` clean across the repo; `dist` rebuilt
- [x] `dotnet test c-sharp-tests/` — 1584 passed, 0 failed. No C# changed, so this only confirms
      nothing regressed
- [x] Manual verification in the running app: at the 900px minimum every toolbar item is present,
      and the horizontal scrollbar is gone. Re-measured on macOS during review — columns come out
      297 / 302 / 297 plus rc-dock's 4px-per-divider reserve, for exactly 900
- [ ] Playwright e2e not run locally (CDP does not open reliably on this Windows machine) — relying
      on CI. Two isolated specs were updated for behavior this branch changes and have not been
      watched run

## Risk Level

**Medium** — the layout changes are broad and touch every toolbar surface plus the window minimum,
so a regression would be visible rather than subtle. Mitigated by the new unit coverage and by the
fact that the failure mode being fixed (items vanishing) is itself what the tests assert against.

[PT-4344]: https://paratextstudio.atlassian.net/browse/PT-4344?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2701)


jolierabideau (commented):
## Review summary

Reviewed against [PT-4344](https://paratextstudio.atlassian.net/browse/PT-4344) (NN-2A of [PT-4336](https://paratextstudio.atlassian.net/browse/PT-4336) "Saroj and Donna can trust the app"), on the branch, with `typecheck`, `lint`, `npm test` and a `platform-bible-react` rebuild run locally.

This is a strong PR. The root-cause diagnosis (`min-width: auto` on flex children + `overflow-clip` on the zones) is correct and is the real explanation for items vanishing, not a symptom patch. The absorbing/rigid vocabulary is a genuinely good abstraction — it makes "which region gives way" a stated property of each zone instead of an accident of the flex defaults, and the tests assert the *absence* of `tw:min-w-0` on rigid zones, which is what actually protects them. Comment density and the ADR/guideline updates are above the bar for this repo.

One item below needs fixing before merge (it fails CI). The rest are questions and follow-ups.

---

### Must fix — the library build fails

`npm run build:platform-bible-react` exits 4 on this branch:

```
[warning] Failed to resolve link to "SHRINK_STEP.WIDE" in comment for ShrinkStepContext
[warning] Failed to resolve link to "SHRINK_STEP.WIDE" in comment for useShrinkStepValue
[warning] Found 0 errors and 2 warnings
npm error code 4
```

Both come from the two `{@link SHRINK_STEP.WIDE}` references in `lib/platform-bible-react/src/context/shrink-step.context.ts` — typedoc can't resolve a member of an object-literal constant, and the build treats warnings as errors. `tsc`, `eslint` and the full test suite are all clean, so this is the only thing standing between the branch and a green build. `{@link SHRINK_STEP}` or plain backticks fixes it.

(Unrelated: rebuilding also reshuffles the injected-CSS order in `dist/index.js` et al. That reproduces on `main` too — pre-existing build nondeterminism, not yours.)

### Worth checking in the running app

Both of these are in the territory the PR already flags as "wants one tuning pass", so they may just fold into that.

1. **The project selector's `MINIMUM` step may not free any space.** `ProjectSelectorLabel` drops `fullName` at `SHRINK_STEP.MINIMUM`, but the trigger it sits in still carries `tw:min-w-48` (`platform-bible-toolbar.tsx:511`) — a 192px floor. The label gets shorter inside the same box, so the titlebar keeps allocating 192px to the project selector at the narrowest window and the space has to come out of `BookChapterControl` instead. If the intent is for the abbreviation to buy room, the floor probably needs to move with the step; if it's purely cosmetic, worth saying so.

2. **Hysteresis can strand a single-threshold consumer after it was hidden.** `getShrinkStep` returns `previousStep` when the width clears no band, which is right during a drag but interacts badly with the `display: none` → width 0 path. For a one-threshold consumer like `EnhancedResourceTabBar` (`[384]`), a pane hidden and then shown at 385–391px lands inside the only band, so `relaxedStep === -1` and the labels stay icon-only at a width where they should be showing — until something else resizes the pane. The four-step ladder self-corrects because it has other bands to relax into (that's the `getShrinkStep(525, THRESHOLDS, 3) === 1` test); a single threshold has none. Treating a *previous* measurement of 0 as "no measurement" and skipping hysteresis on the next one would close it, and the existing fake `ResizeObserver` already makes that testable.

### Fits the ticket — with three gaps worth naming

Against PT-4344's acceptance criteria: items never hidden ✓, tooltip on truncation ✓, no horizontal scrollbar in Simple mode ✓, abbreviations before icon fallback ✓ (and the guideline now says so explicitly, which is the right place for it), fixed slots so switching projects doesn't shift neighbours ✓, monospace marker + greyer trailing detail in the paragraph dropdown ✓, `responsiveness.mdx` updated ✓.

Not covered, and I don't think any of them should block:

- **"'Paragraph change' as an item type can disappear at very small widths"** — the marker menu truncates the detail but no item type drops out. Fine to defer; worth a note so it isn't assumed done.
- **"Power mode toolbar row is shorter"** — the *titlebar* already does this (`isPowerMode ? 'tw:h-12' : 'tw:h-14'`), but `TabToolbarContainer` is still a hard `tw:h-14` in both modes. Was that criterion about the titlebar, or is the tab toolbar's height a follow-up?
- **Power mode's horizontal scrollbar** — ADR-0014 is admirably honest that Power mode's 8px dividers are under-reserved by the same hard-coded rc-dock 4, and defers the patch. Since NN-2A is scoped to the tab toolbar this reads as the right call, but the "no horizontal scrollbar" AC isn't unconditionally true after this PR, only in Simple mode.

Separately: the Jira notes UX asked for "approximately `width / 3` per column". ADR-0014 keeps 1:2:1 and gives a good reason (equal thirds would leave the editor 266px at the 900px floor, below what its own toolbar needs). That's a UX-visible deviation from a written UX decision rather than an implementation detail — has UX seen the 1:2:1 conclusion, or only the 900px minimum?

### Codebase patterns

Conforms well. The flex-truncation idiom is the one the ticket pointed at and the one `scope-selector` / `project-selector` already use. The `.claude/rules/cross-view-sync-hidden-views.md` obligation is met properly — the hidden case is reasoned about in `useShrinkStep`'s TSDoc *and* backed by a test that goes 600 → 0 → 425, which is more than the rule asks for. ADRs recorded per CLAUDE.md, both `eslint-disable`s carry real justifications, no `shadcn-ui/` files touched so the `// CUSTOM:` discipline doesn't come into play, and no new localize keys were needed (reusing `usfmMarkers[marker].description` is a nice catch).

Two smaller pattern notes:

- **Public API surface grows without a consumer.** `TabToolbarContainer`, `TAB_TOOLBAR_SHRINK_THRESHOLDS_PX`, `APP_TOOLBAR_SHRINK_THRESHOLDS_PX`, `getShrinkStep` and `SHRINK_STEP_HYSTERESIS_PX` are all newly exported from `platform-bible-react`, and nothing outside the library imports any of them. `getShrinkStep` + the hysteresis constant look like test seams, and the thresholds are tuning constants the ADR expects to change — once exported, changing them is a consumer-visible change. Consider keeping them internal (tests can import by path) and exporting only `useShrinkStep`, `SHRINK_STEP`, `ShrinkStepContext`, `useShrinkStepValue` and `ToolbarCompoundLabel`.
- **`shrinkStep` as a documented test-only prop** now appears on `TabToolbar`, `Toolbar`, `BookChapterControl` and `EnhancedResourceTabBarProps`. It's the pragmatic answer to jsdom having no layout engine and the TSDoc is honest about it, but it does put a test seam in the API extension authors read. An exported `ShrinkStepContext.Provider` wrapper in tests/stories would get the same result without the prop — worth weighing, not worth reworking now.
- `MARKER_SLOT_CHARACTERS = 6` and the `tw:w-[6ch]` literal are kept in step by hand (correctly — Tailwind can't read the constant). A one-line assertion that the rendered class contains `6ch` would stop them drifting.
- Minor: in `getShrinkStep`, `Math.max(relaxedStep, nextStep)` is unreachable-redundant — `relaxedStep >= nextStep` always holds, since clearing `threshold + 8` implies clearing `threshold`.

### What I liked

- The `e2e` `test.fixme(GAP-024)` reason rewrite. Leaving it skipped and replacing the stale cause with "the gap is closed but nobody has watched it run" is exactly right; enabling it on a code reading would have been the tempting mistake.
- The `main.ts` ↔ `simple-layout.data.ts` comment explicitly saying the test *cannot* catch a change to the 900, so change both. That's the honest version of a mirrored constant. (Since both processes can import from `src/shared/`, a shared `WINDOW_MIN_WIDTH_PX` there would let the test bind to the real value instead — worth a follow-up.)
- Test names that state the behaviour and the consequence rather than the mechanism, and the assertion in the app-menu test explaining why asserting `tw:shrink-0` alone would be vacuous.
- Finding the `tw-@container/toolbar` dash typo, and turning it into ADR-0013's evidence that this mechanism has now failed silently three times in this repo.

**Verified locally:** `npm run typecheck` clean · `npm run lint` 0 errors (4 pre-existing warnings, unrelated files) · `npm test` all workspaces pass, 1092 tests in `platform-bible-react` (one Storybook browser test, `text-field.stories.tsx`, failed once and passed on re-run — flaky, untouched by this PR) · `npm run build:platform-bible-react` **fails**, see above.

jolierabideau (changes requested):
Thank you for working on this, and attaching the demo video. I left a summary comment above, but I ran it to visually confirm on macOS and mine looks a little bit different. The marker and the marker title are not centered vertically and the truncation of the marker menu is not as it appears in your demo video. I can see that it is truncating, it is just happening a little too late and already cut off. Let me know if I can get any other videos or test anything else. 

> @jolierabideau partially reviewed 47 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2701#-:-P-V7S_M4cnxGaCy6c4G:b8jgu10)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

jolierabideau:
https://github.com/user-attachments/assets/014efe84-0d4a-402c-8b65-ea442d35d946

katherinejensen00:
## Fixed

- **Build failure.** Both `{@link SHRINK_STEP.WIDE}` references are plain backticks now. I'd only
  been running typecheck/lint/test, so the build was never in my loop — it is now.
- **Hysteresis stranding a single-threshold consumer.** Your fix was the right one: `previousStep`
  is optional, and the hook passes nothing when the last width was 0, since being revealed isn't a
  drag and there's nothing to damp. Three tests — your exact 384/385 repro, the updated 600 → 0 →
  425 case (now step 1, not 2), and one confirming a genuine widening drag still gets hysteresis.
- **Project selector `MINIMUM` freeing no space.** Right, and buying room is the intent. The floor
  now moves with the step (192px → 96px), which needed a small `ProjectSelectorTrigger` so the
  className can read `ShrinkStepContext` from below the provider. Test compares trigger classes at
  `WIDE` vs `MINIMUM`.
- **Public API surface.** Agreed, especially the threshold constants — exporting tuning values makes
  changing them consumer-visible. All five are out of the barrel; tests import by module path, and
  there's a comment at the export site saying why.
- **`Math.max(relaxedStep, nextStep)`.** Removed, and the comment now states the implication so it
  doesn't get re-added.
- **`MARKER_SLOT_CHARACTERS` drift.** `marker-slot-width.test.ts` reads the source and asserts every
  `tw:w-[Nch]` matches the constant, following the `simple-moddy in
  that extension. It also fails if the literal disappears, so it can't pass vacuously.
- **macOS vertical centering.** Reproduced by inspection: `Too
  `items-baseline`, and the marker slot is an `inline-block` with `overflow: hidden` — which per CSS
  takes its *bottom margin edge* as its baseline rather than i marker
  off the wrong edge. The offset depends on font metrics, which is exactly why it showed for you and
  not in my Windows capture. Centred now; both fields are the ves.

Also: a stray `|||||||` diff3 conflict marker from an earlier 0016 in
`Architecture-Decisions.md`. Removed. It was the only marker of its set, so nothing flagged it.

## Needs you or UX

- **macOS truncation timing.** I haven't fixed this — I can't reproduce macOS font metrics locally
  and didn't want to guess at CSS a third time. Taking you up  the
  width where it looks wrong would settle it. My suspicion is the fixed marker slot — it's
  `overflow: hidden` with no ellipsis of its own, so past a ceuts
  rather than ellipsising, which would read as "already cut off". If instead the whole label
  truncates later than it should, that's the threshold tuning against a
  real window than by arithmetic.
- **1:2:1 columns.** You're right to separate these — UX has sm and the
  guideline rewrite, but the 1:2:1 conclusion hasn't been put to them explicitly. I'll take the
  reasoning over (equal thirds leaves the editor ~266px at theown
  toolbar needs) and get a yes or no before this merges, rather than letting the ADR stand as the
  decision.
- **Power mode row height.** I read that criterion as being about the titlebar, which already does
  it — but I'm genuinely unsure. If you read it as covering th a
  follow-up rather than widen this PR.

## Acknowledged, not changed

- **`shrinkStep` test-only prop.** Taking your "not worth reworking now" — but flagging that I made
  it slightly worse: `EnhancedResourceTabBar` had no seam at athere to
  get it under test. Five call sites now, not four. Happy to move to a Provider wrapper as a
  follow-up before it spreads further.
- **Shared `WINDOW_MIN_WIDTH_PX`.** Agreed that's the real fix. Left out only because anything in
  `src/shared/` regenerates `papi.d.ts`, putting a window-sizice
  extension authors read — worth a minute on where it should live. Say the word and I'll do it here.
- **The three ticket gaps.** The 'Paragraph change' item type so it
  isn't assumed done. And yes, the no-horizontal-scrollbar AC holds in Simple mode only; I'll say so
  on the ticket rather than let it be recorded as unconditiona

jolierabideau (commented):
Re-ran the manual test on macOS at `2dc04cd`. The scrollbar fix and the "nothing disappears" behavior both hold up. The two things I raised last round are one fixed-partially and one diagnosed — and the truncation cause turned out not to be font metrics or threshold tuning, so you were right not to guess at the CSS a third time.

## Verified working

- **No horizontal scrollbar at the 900px minimum.** Columns measure **297 / 302 / 297 = 896 + 4px dividers = exactly 900**, dock box `min-width: 899px`, `scrollWidth == clientWidth`. The ADR-0014 arithmetic (3 × 297 + 8 = 899) holds in the running app. Clean at 1600 too (399 / 798 / 399).
- **No toolbar item disappears at 900px.** Titlebar keeps Platform, Help, `Test Project Demo (TPD)` at full length, back/forward, `GEN 1:31` abbreviated, and the account button. The editor tab toolbar keeps hamburger, undo, redo, style button (collapsed to `p`), share, and *both* lock icons — the structure-protection lock that used to vanish is present.
- **Marker-menu row alignment is good** — marker, title and detail box centers agree within 0.5px.

## 1. The marker is still ~2.25px high in the toolbar paragraph-style label

The centering fix helped but didn't fully land. Tight glyph rects (via `Range.getBoundingClientRect`) in the paragraph-style button at a 798px column:

| text | box top | font |
| --- | --- | --- |
| `p` (marker) | **9.25** | ui-monospace |
| ` - ` (separator) | 11.5 | Google Sans |
| `Paragraph - Normal - First Line Indent` | 11.5 | Google Sans |

`items-center` centers the marker's *wrapper*, but that wrapper measures 25.5px tall against 20px for its siblings — because the fixed marker slot is an `inline-block` with `overflow: hidden`, which is exactly the bottom-margin-edge baseline quirk your commit message describes. So the glyph sits at the top of its own line box, and centering the box doesn't center the glyph.

Good contrast: the marker menu rows don't use the fixed slot and measure within 0.5px. Suggested direction: make the slot `inline-flex` + `items-center` so there's no line-box descender to correct for.

## 2. Marker-menu truncation — cause found, and it's not font metrics

You asked for the width where it looks wrong. It's **any column narrower than ~390px**, which the new 900px minimum guarantees in Simple mode.

`PopoverContent` for the paragraph switcher is a hard-coded `tw:w-96` = **384px** (`platform-scripture-editor.web-view.tsx:2262`). In Simple mode at the 900px window minimum the editor column is **302px**. The popover lays out at its full 384px anyway and gets clipped by the web view edge, so **~82px of every row is invisible — including the ellipsis**.

Measured at a 900px window (iframe viewport 302px):

- popover width 384, `right` 384 → 82px past the 302px edge
- rows 354px wide, detail fields end at `right: 330` — i.e. 28px beyond the visible edge
- details *are* truncating correctly: `scrollWidth` 205 vs `clientWidth` 186, `text-overflow: ellipsis`, and the native `title` tooltip is present on both fields

So the internal shrink weighting is fine — the fields are truncating to a width nobody can see. Visible result is text cut mid-word with no ellipsis: "Running header text for a boo", "Running header text, ri", "Chapter label".

At 1600px (798px column) the same popover fits with 0px overflow and renders exactly like your demo video — proper ellipses, title winning over detail. That's why it looked right on your capture and wrong on mine: it's column width, not platform.

Suggested direction: cap the popover to the space available rather than fixing it at 384px — Radix exposes `--radix-popper-available-width` for this, e.g. a `max-width` on `PopoverContent`. Note `tw:w-96` predates this branch (April), but the detail column is new here, so this is the change that made it visible.

## Adjacent, and I don't think it belongs in this PR

`model-text-panel.component.tsx:493` puts `tw:truncate` on a `tw:flex` element, where `text-overflow: ellipsis` is inert. The left column's resource name therefore hard-clips mid-glyph at a 297px column (`scrollWidth` 369 vs `clientWidth` 297) instead of ellipsising. It's from PT-3958 in July and untouched here — same failure class as this ticket, but worth its own ticket rather than widening this one.

## Testing notes

macOS only; I didn't cover Windows or Linux. Two traps in case they cost you time: `--window-size` needs a space rather than `=` (the parser does `indexOf('--window-size')` and reads the next argv entry), and `pw-server.mjs` force-sets a 1920×1080 viewport override on every connect — my first "900px" run was really a 1920 layout clipped into a 900px window, and I nearly reported the artifact as a bug. Everything above was re-measured with the real window size and the viewport matched to it.

jolierabideau:
Screenshot 2026-08-24 at 5 02 13 PM

jolierabideau (approved):
Thank you Katherine for sticking with this! Approving and will test manually on macOS and work on any follow-up

> @jolierabideau partially reviewed 52 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2701#-:-P-yVbCW6zhrDlgHuCu-:bj5np4l)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2702, opened 2026-08-20
PT-4279: drop the top-level menu from secondary windows

timothy-mccormack:
## What and why

Secondary windows currently render chrome **identical** to the main window, top-level menu included. The target is settled — investigation brief §6: *"Secondary windows keep the full toolbar and remove **only** the top-level menu."* So this is a removal, not a build.

The factory half of PT-4279 was already delivered by #2621, so what remained was this plus two items left out of scope below.

## How it works

The shared `Toolbar` renders its menubar as `menuData && <Menubar/>`, so **withholding the prop removes the menu and nothing else** — the app-menu-area logo and every toolbar control still render. No change is needed in `platform-bible-react`.

The renderer learns which window it is through a presence-only URL search parameter set on the first window, the same seam that already carries log level, dev mode, startup marks and the window id:

- `IS_MAIN_WINDOW_QUERY_PARAMETER` in `platform.data.ts`
- set in `main.ts` when `isFirstWindow`
- read into `globalThis.isMainWindow` in the renderer
- consumed by `platform-bible-toolbar.tsx`

125 lines added, nothing removed.

## ⚠️ Known limitation — please weigh in

The flag is **fixed at window creation**, so it cannot describe a window *becoming* the main one. Concretely: **if the main window closes, the surviving secondary window keeps no menu**, and there is then no menubar anywhere.

PT-4278's window-manager service is the durable answer, and both declaration sites carry a comment saying to replace this when it lands. Whether the interim behaviour is acceptable is a product call rather than a spike call — happy to take the alternative (wait for PT-4278) if reviewers prefer it.

## Testing

- **New e2e regression guard** — `e2e-tests/tests/isolated/multi-window/secondary-window-chrome.spec.ts`, asserting in both directions: the main window keeps its menubar, the secondary has none, and the secondary still has the toolbar and the logo (so this is "removed only the menu", not "removed the app menu area").
- **The guard was confirmed to fail without the change** (`Expected: 0, Received: 1`), so it guards what it claims to.
- Unit suite green: 101 files / 1247 tests.
- `tsc`, `eslint`, `prettier` clean.

## Out of scope for this PR

Two PT-4279 items are deliberately not here:

1. **The fourth Linux frameless guard** (`platform-bible-toolbar.tsx:213`) — one line, but its evidence is a Linux screenshot and no Linux hardware is assigned.
2. **Guaranteeing a grabbable window-drag area at every zoom level** — `shouldUseAsAppDragArea` sets `WebkitAppRegion: 'drag'` on the toolbar while interactive children set `no-drag`, and that logic lives in `platform-bible-react`, so it likely is not fixable in this repo. Worth confirming before it is estimated.

The DoD's side-by-side chrome screenshots (Windows / macOS / Linux) are also still outstanding.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2702)


rolfheij-sil:
🤖 Claude: AI-assisted review, run at Rolf's request because this PR had no reviews on it yet. Every finding below was re-verified by hand against head `b562d7ea` before posting — two claims from the first pass were wrong and have been dropped rather than sent to you.

The approach is right, and it matches how we've been told to think about window identity (all windows equal siblings; a reassignable role flag is fine, a window *class* is not). The findings are about the role never being reassigned, and about the gate not reaching the platform where the menu actually lives.

---

## Blockers

### B1 — `papi.d.ts` not regenerated (you already have this signal)

All three build legs fail on "Report file changes". Flagging only because of a gotcha that costs time: a plain `npm run build:types` **silently no-ops** — the gitignored `papi.tsbuildinfo` makes tsc skip. Use `npm run build:clean` in `lib/papi-dts`.

Worth doing early: until it regenerates, the `@experimental` markers this PR adds never reach the published extension surface.

### B2 — Closing the main window leaves the app with no menu, and nothing recovers it

Power mode, Windows or Linux. Windows A (main) and B (secondary) open; the user closes A.

- `window-all-closed` (`src/main/main.ts:1353`) fires only when *every* window closes, so the app stays up.
- Nothing re-assigns the role: `setMainWindowId` is called from exactly two places, `main.ts:1313` and `:1326`, **both inside `restoreWindows`**.
- `newWindow.setMenu(null)` (`main.ts:1088`) means there's no native menu fallback.
- Opening another window doesn't help — it also sees a non-zero window count, so it gets `isMainWindow=false` too.

Net effect: a running app with no menubar anywhere, recoverable only by restarting.

Your TSDoc already notes the flag can't describe a window becoming main later, and defers that to PT-4278. The part worth raising is that the consequence isn't "incomplete", it's unrecoverable — and the machinery to fix it already exists: `onDidCloseWindowEmitter` (`main.ts:443`, emitted `:1060`), which the theme and scroll-group hosts already re-elect on.

Worth reading alongside: **ADR-0009** (Accepted 2026-08-05, same epic, PR #2621) is titled *"App-global singleton services elect a host window first-come, **with takeover on host-window close**"*, and its rejected alternatives include:

> *"A fixed 'primary' window as host — rejected: any window, including the first, can be closed by the user, so a fixed designation would still need a takeover path."*

That ADR is about singleton service hosts rather than the menubar, so it doesn't govern this directly — but it's the same trap with the answer already written down.

---

## High

### H3 — On macOS the feature is a no-op, so the PR title isn't accurate there

`Menu.setApplicationMenu` is process-global and driven from the main process under a `process.platform === 'darwin'` guard (`main.ts:533`, util at `platform-macos-menubar.util.ts:54`). `main.ts:907-910` says so directly: *"the macOS menubar is NOT subscribed here. `Menu.setApplicationMenu` is process-global… see the `darwin` block."*

The renderer applies no darwin gate — `menuData` is passed unconditionally. So a macOS secondary window still shows the full Project/Help menu at the top of the screen, routing to the focused window.

The new e2e spec can't catch this: it asserts only on renderer DOM, so it passes on macOS while the feature does nothing.

### H5 — `isMainWindow` is typed `boolean` but is `undefined` in most processes

`src/shared/global-this.model.ts:98` declares `var isMainWindow: boolean;` — non-optional — and it's assigned only in `src/renderer/global-this.model.ts`. Confirmed by grep: zero assignments in `src/main/global-this.model.ts` and `src/extension-host/global-this.model.ts`, while both sibling flags are set in all three (`isNoisyDevModeEnabled`, `startupMarks` at `main:28-29`, `extension-host:21,31`). So in main, the extension host and vitest it's `undefined` while TypeScript promises `boolean`, and the TSDoc's "`false` in every other process" isn't what happens.

The honest precedent is twelve lines up in the same file: `var windowId: string | undefined;`.

Related, and worth checking before this reaches `papi.d.ts`: the web-view prelude that copies globals into each iframe doesn't appear to include `isMainWindow`, so an extension writing the type-safe `if (globalThis.isMainWindow)` would take the falsy branch in the main window too, with no compile error.

---

## Medium

### M6 — Five catalogued keyboard shortcuts die silently, and the catalog wasn't updated

`useHotkeys(['alt', 'alt+p', 'alt+l', 'alt+n', 'alt+h'], …)` is registered **inside** `PlatformMenubar` (`platform-menubar.component.tsx:172`), and `Toolbar` mounts that component only under `{menuData && (` (`toolbar.component.tsx:119`). Withholding `menuData` unmounts it, so the hook never registers and Alt / Alt+P / Alt+L / Alt+N / Alt+H do nothing in secondary windows.

`.claude/rules/keyboard-shortcuts-catalog.md` asks for the catalog to move with the handler; entries `menubar-close` / `-project` / `-window` / `-layout` / `-help` (`src/stories/keyboard-shortcuts.data.ts:362-405`) still read `context: 'Application menu bar'` with no window qualifier.

This also means the diff's comment that withholding menuData "removes the menu and nothing else" isn't quite right.

### M7 — The new spec never runs in CI

`.github/workflows/test.yml` runs only `npm run test:e2e:smoke`, which is `--project=smoke` (`package.json:104`). The isolated suite runs through a separate script (`test:e2e:isolated`, `:103`) that CI doesn't invoke. So nothing in CI would catch a regression of this behaviour.

Related and worth a look: `globalThis.isMainWindow` is `undefined` under vitest, so `if (!globalThis.isMainWindow) return undefined;` fires in every existing toolbar test — they render menu-less and still pass, because none of them asserts on the menubar. Two cheap unit tests would pin both directions.

### M8 — The spec drops three safety rails its three siblings all have

Counted across `e2e-tests/tests/isolated/multi-window/`:

| Spec | `test.use` | `waitForRendererRegistered` | `captureAppOutput` |
| --- | --- | --- | --- |
| `secondary-window-chrome` | 0 | 0 | 0 |
| `multi-window` | 1 | 4 | 4 |
| `per-window-ui` | 2 | 2 | 2 |
| `window-layout-persistence` | 1 | 4 | 6 |

The `test.use` one has teeth: without it the option fixture defaults to `{}`, so `isolatedProjectRoot` is unset and the run reads and writes the developer's **real Paratext projects** — which `multi-window.spec.ts:272-274` explicitly warns against — and `DEV_NOISY` defaults on, loading test-only extensions whose `mainMenu` contributions change the very structure being asserted.

### M9 — The positive control passes on an empty menubar

`MAIN_MENU_DEFAULT = { columns: {}, groups: {}, items: [] }` (`platform-bible-toolbar.tsx:73`) is truthy. So if menu data never loads in the main window, it still renders a menubar with no columns and `toHaveCount(1)` passes. Separately, `TOOLBAR` is the component's unconditional root wrapping everything, so those two assertions can't fail independently — nothing currently asserts that any toolbar *control* (BCV, project picker, sync) survives, which is the spec's stated third point.

### M10 — Two sources of truth for "which window is main", and the stale one can misfire

`window-layout-persistence.service.ts` already owns this: `mainWindowId` (`:87`), `setMainWindowId()` (`:322`), persisted `isMain` (`:395`). The URL flag is re-derived independently from `getWindows().length === 0`.

Two ways they can diverge:

- `handleWindowRemoved` filters `trackedWindows` and `fileSlots` but never clears `mainWindowId` — and `window-state.service.ts:286` notes that **Electron reuses window IDs**, so a later unrelated window can inherit the persisted `isMain`.
- They agree at startup only because the main entry's window is created first. `main.ts` states that ordering is semantically irrelevant ("the saved structure keeps entry order regardless of window creation order"), so a refactor iterating in entry order would hand the menu to a secondary window whenever the main entry isn't index 0.

Before this PR `isMain` only chose restore ordering; it now decides who gets the menu.

### M11 — The menu-data subscription is kept and then discarded

`platform-bible-toolbar.tsx:265` subscribes unconditionally and the new gate at `:273` throws the result away, so with N windows the extension host localizes and serializes the full merged menu N times on every update.

Cleanly fixable — `useData` accepts `undefined` as a source to skip subscribing, and the exact idiom against this same provider is one file over: `web-view.component.tsx:512`, `webViewType && shouldShowToolbar ? menuDataService.dataProviderName : undefined`. The comment at `:264` says this matches the pattern `web-view.component.tsx` uses; it matches the live-subscription half but not the conditional-source half.

Side effect: the `isPlatformError` warn branch at `:274-279` is now unreachable in secondary windows, so a provider failure there logs nothing.

---

## Low

- **L12** — `ToolbarProps.menuData`'s TSDoc says *"If empty object, no menus will be shown"*, but the implementation is a truthiness check and `{columns:{},groups:{},items:[]}` is truthy. The whole feature rests on `undefined` being falsy inside a shared library component whose documented contract says otherwise. An explicit `showMenu?: boolean` would put the intent in the shared contract.
- **L13** — No ADR entry. Given ADR-0009 reasoned the other way on a closely related question, a short entry saying why this shape was chosen would save the next person the archaeology. (`Architecture-Decisions.md` currently runs to ADR-0025.)
- **L14** — Backward-facing `(PT-4279)` comments at five sites: `platform.data.ts`, `shared/global-this.model.ts`, `main.ts`, `platform-bible-toolbar.tsx`, and the spec. The first two ship verbatim into `papi.d.ts` and the published TypeDoc site. Per `.claude/rules/code-quality/forward-facing-comments.md` those belong in the commit message. The `PT-4278` references are correctly forward-facing — keep those.

---

## Not a change request — flagging for visibility only

Dropping the top-level menu makes **eight commands unreachable in a secondary window**, because each is contributed only to the main menu: `platform.quit`, `platform.createWindow`, `platform.about`, `platform.usersnapReportIssue`, `platform.usersnapSubmitIdea`, `platform.openSettings`, `platformEnhancedResources.openEnhancedResource`, `platformLexicalTools.openDictionary`.

This isn't yours to fix — it's the cost of "secondary windows keep the full toolbar minus the top-level menu", decided at the 2026-07-27 product checkpoint that your commit message correctly cites. A new secondary window starts with an empty dock, so the web-view hamburger's Settings entry isn't there either. One consequence worth someone deciding on: a user hitting a bug in window 2 has to switch to window 1 to report it, and the Usersnap screenshot then captures the wrong window. Rolf is taking the inventory to the product owner.

---

## Suggested order

**B1, B2, H3, H5, M6** before merge. B2's minimal fix is the shape already in the repo — on window close, pick a survivor and promote it, reusing `setMainWindowId` and `onDidCloseWindowEmitter`, the same election theme and scroll-group already do.

**M7–M11** matter mostly because right now nothing in CI would catch a regression of any of this.

**L12–L14** are fine as follow-ups.

Happy to be told any of these is deliberate for now and belongs with the PT-4278 replacement.

timothy-mccormack:
Thanks for running this — it caught a blocker I'd have shipped. Everything below is verified against the codebase rather than taken on faith, and where I disagree I've said so.

Fixed in `0d545397bc1`:

**B1** — `papi.d.ts` regenerated. Confirmed the gotcha you flagged: `build:types` no-ops via the gitignored tsbuildinfo, `build:clean` in `lib/papi-dts` is what actually rebuilds. CI's failure was literally `Error - changed files after build: lib/papi-dts/papi.d.ts`.

**H5** — now `boolean | undefined`, matching the `windowId: string | undefined` precedent. Confirmed by grep that it's assigned only in the renderer while both sibling flags are set in all three processes. The TSDoc no longer claims `false` elsewhere, and calls out the web view prelude explicitly.

**M6** — you're right that "removes the menu and nothing else" was wrong. Verified in the compiled library that `["alt","alt+p","alt+l","alt+n","alt+h"]` is registered inside `PlatformMenubar`, which mounts only under `{menuData && (`. The comment now says so, and the five catalog entries are qualified with "main window only".

**M8** — `test.use` added with `isolatedProjectRoot: true` and `DEV_NOISY: 'false'`. This one had teeth beyond CI: the spec had been run twice locally against real projects before this was caught.

**M9** — the positive control now requires a real menu item, since `MAIN_MENU_DEFAULT` is truthy and an empty menubar satisfied `toHaveCount(1)`. The secondary window now asserts a real toolbar control survives rather than only the unconditional root.

**M11** — the subscription is now skipped rather than discarded, using the `webViewType && shouldShowToolbar ? ... : undefined` idiom from `web-view.component.tsx`.

**M7** — added the two unit tests you suggested, pinning both directions. Your prediction was exact: `isMainWindow` is `undefined` under vitest, so the gate fired in every existing toolbar test. One of them broke the moment I made the subscription conditional, which is how I know the new tests are load-bearing. The spec still doesn't run in CI — that's a workflow change I'd rather not fold into this PR.

**L14** — backward-facing `(PT-4279)` refs removed from the comments that ship into `papi.d.ts`; forward-facing PT-4278 ones kept.

---

**H3 — I think this one is overstated.**

The mechanism is right: `Menu.setApplicationMenu` is process-global, and `platform-macos-menubar.util.ts` builds the macOS menu from the same menu data, so a macOS secondary window can still reach every menu command from the system menu bar.

But "the feature is a no-op there" isn't what happens. macOS renders the menubar in the toolbar too, and this change removes it — side-by-side screenshots on PT-4279 show window #2's toolbar with no Platform/Help while window #1 has both.

More to the point, the system menu bar **cannot** differ per window on macOS; that's how every macOS app works. So there's no code fix here, only a product question: on macOS a secondary window loses the in-toolbar menu but keeps command access through the system menu bar. That's arguably the correct native behaviour rather than an incompleteness. Happy to add a darwin note to the TSDoc if that's the concern, but I don't think the renderer should gate on platform.

**B2 — agreed, and I'd like to defer it deliberately rather than pretend it's handled.**

Your description is right, including that the consequence is unrecoverable rather than merely incomplete. The reason it isn't fixed here: the flag is a URL search parameter fixed at window creation, so re-election needs a runtime channel to tell a renderer it just became the main window. That's PT-4278's job, and building a second mechanism now means building something we'd delete.

Two options and I'll take whichever you prefer:

1. Merge with the gap documented, and PT-4278 delivers re-election.
2. I add the interim election in this PR — on `onDidCloseWindowEmitter`, promote a survivor via `setMainWindowId` and push the flag to that renderer — knowing PT-4278 replaces it.

I lean towards 1, but "no menubar anywhere until restart" is a bad failure mode and it's your call whether it can sit in `main` in the meantime.

**M10** and **L12/L13** I've left for now — M10 in particular reads as pre-existing and arguably belongs with PT-4278, since that's where the second source of truth goes away. Say the word if you'd rather they were in scope here.

The inventory of eight commands unreachable in a secondary window is worth the product conversation you're already taking to the PO — a user hitting a bug in window 2 can't report it from there, and the Usersnap screenshot captures the wrong window.

rolfheij-sil:
🤖 Claude: Decisions from Rolf, so you're unblocked on both open items.

**B2 — option 1, and the gap turned out to be narrower than either of us thought.** Merge with it documented. The team settled window-close behaviour today: closing a secondary window just closes it, and closing the primary while other windows are open prompts, then closes the whole app. So the state your TSDoc warns about — a surviving window that can never become main — is no longer reachable: the primary can't outlive its secondaries. The flag being fixed at creation is fine on those terms rather than merely tolerable. PT-4286 carries the close-behaviour work.

**H3 — agreed on the mechanism, and Rolf would like the macOS variant built.** You're right that the in-toolbar menubar goes on macOS too, and that a renderer-side platform gate isn't what we want — this isn't that. The reachable macOS variant is a main-process one: swap the application menu on `browser-window-focus`/`blur` so it carries the full set when the primary is frontmost and a reduced set when a secondary is, which is how native macOS apps make the menu bar track the key window. Removal isn't reachable — `setApplicationMenu(null)` only strips the bar on Windows and Linux — so "reduced", not "absent", is the target. If you'd rather not grow #2702 with main-process work, say so and Rolf will open a follow-up ticket for it; either way #2702 shouldn't wait on it.

Separately and cheaply, either way: `papi.d.ts`'s "identical chrome minus that menu" and the catalog's "secondary windows have no menu bar" aren't true on macOS today, so those two sentences want a darwin qualifier. The effect claim is right on all three platforms though — the macOS menu template assigns no accelerators, so ⌥/⌥P/⌥L/⌥N/⌥H are renderer hotkeys and do die with `PlatformMenubar` there too; it's only the stated reason that's off.

rolfheij-sil:
🤖 Claude: Second pass at `0d545397bc1f`, after this morning's eight fixes — the subscription skip, the `boolean | undefined`, the `test.use` rails and the stronger positive controls all check out, and CI is green on all three legs. Everything below is small.

**Before merge — all small**

1. **H3 wording**, as in the earlier notice — one Windows/Linux qualifier each at `lib/papi-dts/papi.d.ts:965`, `src/shared/global-this.model.ts:89`, `src/shared/data/platform.data.ts:26`, and the five catalog rows (`src/stories/keyboard-shortcuts.data.ts:365, 375, 385, 395, 405`), which all state the removal unconditionally.

2. **The toolbar test file's default window identity flipped.** `globalThis.isMainWindow` is set in only two describes (`platform-bible-toolbar.test.tsx:804`, `:835`/`:843`); the other eight — Sync button `:242`, scroll-group `:510`, project picker `:536`, BookChapterControl `:559`, write-back `:634`, Home `:737`, styling `:760`, title bar `:856` — now render as a secondary window. Nothing is red, because none of them asserts on the menu, but the file's implicit default is now the configuration no main-window user sees. One line: `globalThis.isMainWindow = true` in a file-wide `beforeEach`, keeping the deliberate `false` as the override.

3. **The spec's "removed only the menu" claim can't fail.** `secondary-window-chrome.spec.ts:92`, `:93` and `:96` use `toBeAttached`, which passes for a `display:none` or zero-size element, while the main-window control at `:86` uses `toBeVisible` — asymmetric within one test. The unit mock can't cover the gap either: it drops `appMenuAreaChildren` (`platform-bible-toolbar.test.tsx:157`). Two lines.

4. **Worth an ADR entry.** ADR-0009 (`.context/standards/Architecture-Decisions.md:288`) rejected a fixed primary window precisely because "any window, including the first, can be closed by the user, so a fixed designation would still need a takeover path". Today's close rule answers that objection — which is exactly what makes the entry worth writing, so the next reader doesn't re-derive the old one and flag this as a violation. Either an entry here, or say it rides PT-4286's.

5. **Rebase before merge.** The branch is 30 commits behind `main`, and `platform-bible-toolbar.test.tsx` moved there (largely `f0ce0758d6e`, PT-4348), so it will conflict — and resolving it forces the item-2 decision. `papi.d.ts` wants regenerating rather than merging. Worth re-running CI after rather than before.

**Noted, handled elsewhere**

Twelve commands have no entry point at all in a secondary window on Windows/Linux (`platform.about`, `createWindow`, `quit`, application `openSettings`, the two Usersnap items, the four Help links, plus Enhanced Resources and the dictionary) — that's a product decision, with Rolf and Vladimir rather than you. Separately, in Simple mode a secondary window has no route to Home at all; PT-4286 owns Simple-mode window behaviour.

**Minor, one line each**

- `isMainWindow` is `undefined` in web views and the extension host, so an extension's `if (globalThis.isMainWindow)` is a dead branch in every window, main included.
- The isolated spec doesn't run in CI — your call, and you already said so; noting only that the guard isn't protecting anything yet.
- The subscription comment's first half ("so this always reflects the current mode and current localization") is no longer true in a secondary window.
- The five catalog rows' `locations` still name only `platform-menubar.component.tsx`, not `platform-bible-toolbar.tsx:280`, which is what actually gates whether those shortcuts exist.
- The spec header still carries `(PT-4279)` for in-PR work and a citation to the investigation brief, which lives outside the repo.

Nothing here blocks merge — CI is green at `0d545397bc1f`.

timothy-mccormack:
Second pass addressed. Verified against the codebase, not taken on faith. Fixed in `efcd7a91df4` (rebased onto current `main`, which forced the item-2 conflict resolution as you expected):

**1 — H3 wording.** Added a Windows/Linux qualifier at all four cited sites plus the five catalog rows. Confirmed by re-reading: the removal-of-the-in-window-menu claim was unconditional in `papi.d.ts`, `global-this.model.ts`, `platform.data.ts`, and the catalog; all four now state the OS-level menu bar keeps every window reachable on macOS. Catalog entries also get a second `locations` row for `platform-bible-toolbar.tsx`, since that's the actual gate — `platform-menubar.component.tsx` alone doesn't explain why the hotkey is gone.

**2 — Toolbar test file default.** Confirmed by grep: only two of ten describes ever touched `globalThis.isMainWindow`, so the other eight were silently exercising the secondary-window configuration. Added `globalThis.isMainWindow = true` to the file-wide `beforeEach`, kept the deliberate `false`/`true` overrides where a describe cares which window it is. All 40 tests in the file still pass, and the full suite (170 files / 2292 tests) is green.

**3 — Spec assertions.** `APP_LOGO` and `BCV_CONTROL` now use `toBeVisible` instead of `toBeAttached` — you're right that a `display:none` or zero-size survivor would have passed the old assertion without actually being "kept." Left `TOOLBAR` on `toBeAttached` since that one's explicitly the weak root-only check per the existing comment. Didn't touch the unit mock's `appMenuAreaChildren` gap — that's coverage the e2e spec now owns, not a defect in the mock itself.

**5 — Rebase.** Done. 38 commits behind (not 30 — grew since your comment), rebased onto `origin/main`. `platform-bible-toolbar.test.tsx` conflicted as predicted from its PT-4348 move; `platform.data.ts` also conflicted (unrelated new query-parameter table landed on `main` since this branch forked) — merged `IS_MAIN_WINDOW_QUERY_PARAMETER` into the new `URL_PARAMETERS` table as a `flag` kind, which matches how it's actually read (`searchParams.get(...) !== null`). `papi.d.ts` conflicted too (generated file); took `build:clean` fresh after resolving everything else rather than hand-merging it. CI is running now.

**4 — ADR entry: deferring to PT-4286.** That ticket owns the close-behaviour change that answers ADR-0009's objection, so the entry belongs with the code that actually implements it rather than describing behaviour this PR doesn't touch.

**Minor items:**
- Subscription comment's "always reflects the current mode" — fixed; it's now scoped to the main window, since a secondary window never subscribes at all.
- Spec header — dropped the backward-facing `(PT-4279)` tag and the out-of-repo investigation-brief citation.
- `isMainWindow` dead-branch risk for extensions — already covered by H5's TSDoc ("code outside a renderer must not read this as a reliable `false`"); didn't see anything to add beyond that.
- Isolated spec not in CI — unchanged, as noted last round; still your call.

Ready for another look whenever convenient.

timothy-mccormack:
CI is green at `30438a4b220` — the earlier red run was mine, not the fixes: I'd regenerated `papi.d.ts` before running prettier on the edited `platform.data.ts`, so it carried a stale line-wrap. Verified byte-identical to a build from a fresh clone of this branch before pushing the fix. All six checks pass.

rolfheij-sil:
🤖 Claude: Two small notes on the new commits — neither blocks, the branch is green and the gate reads right.

**1. One clause in the new macOS TSDoc is stronger than the platform is.** In `src/shared/global-this.model.ts:88-92` (and the generated `papi.d.ts:1074-1079`) the sentence ends: *"…the system menu bar, which is process-global and cannot differ per window."*

Process-global is right, but "cannot differ per window" holds only for *simultaneous* difference. Swapping the application menu on `browser-window-focus` / `browser-window-blur` so it tracks the frontmost window is the ordinary Electron pattern, and it is what a reduced menu for secondary windows would use if we ever build one. As written, the docs rule that out.

Worth noting the other two places you describe this are already precise: `papi.d.ts:8680-8681` says "process-global and reachable from every window regardless of this flag", and the catalog row at `keyboard-shortcuts.data.ts:375-376` spells out the whole picture including that the OS menu bar assigns no accelerators of its own. Matching the TSDoc to either of those would settle it.

Related, and worth saying plainly: Rolf's earlier suggestion about building a focus-driven reduced menu on macOS was a developer's opinion, not a product or UX decision. If you think it is the wrong idea, say so — nothing depends on it.

**2. For the record, the menu-only command count.** Counted from `menu.data.json` at this head: **12** commands are reachable only through the top-level menu — `platform.about`, `createWindow`, `quit`, `openDeveloperDocumentationUrl`, `usersnapReportIssue`, `usersnapSubmitIdea`, `visitFAQsPage`, `visitFeatureRoadmapPage`, `visitGettingStartedPage`, application-scoped `openSettings`, `platformEnhancedResources.openEnhancedResource`, and `platformLexicalTools.openDictionary`. On macOS all twelve stay reachable through the OS menu bar; on Windows and Linux they do not. That is the direct consequence of withholding `menuData` rather than anything wrong here — noting it so whoever picks up the alternate-route work has the list.

**3. The ADR entry we asked you for — we'll take it.** Our 26 Aug comment asked for an entry about the close-behaviour decision, and your commit defers it to PT-4286. On reflection that was the wrong thing for us to ask: the decision is ours, made in #2729, and that PR already edits `Architecture-Decisions.md`. We'll write it there. Nothing needed from you.

rolfheij-sil (approved):
🤖 Claude: Approving. Two review rounds found nothing blocking — the gate is right, the experimental markers are correct on both surfaces, `papi.d.ts` regenerates clean, and CI is genuinely green (read from the warn steps rather than the job conclusions, since the E2E step's `continue-on-error` rewrites its conclusion regardless of outcome).

The TSDoc wording in my earlier comment is a suggestion, not a condition — take it or leave it, and merge whenever suits you.

PR #530, opened 2023-10-04
Create some first-pass issue templates

FoolRunning:

PR #2703, opened 2026-08-20
PT-3780: Put the insertion point inside a marker added in a footnote

katherinejensen00:
# PT-3780: Put the insertion point inside a marker added in a footnote

## Summary

Adding an `fk` or `fq` **inside a footnote** left the caret *before* the new marker and its bubble
instead of inside it, so the next keystroke landed outside the marker you just created.

The fix is upstream and already released. This branch just picks it up: it bumps
`@eten-tech-foundation/platform-editor` from `~0.8.14` to `~0.8.15`.

**No behavior code in paranext-core changed.** Four files are version strings, three are
comment-only. That is the whole diff.

| Commit | Files | What it does |
| --- | --- | --- |
| `9cf8220b689` | `package.json`, `lib/platform-bible-react/package.json`, `extensions/src/platform-scripture-editor/package.json`, `package-lock.json` | Bumps the editor dependency in all three places that declare it, plus the lock. **This is the actual fix.** |
| `4ce02fef0c3` | `character-marker-menu.utils.ts`, `use-remove-character-marker.hook.ts`, `platform-scripture-editor.web-view.tsx` | Comments only. Eight comments claimed the editor lacked operations it now has — including one telling readers not to trust undo "until that pin is bumped". No logic touched. |

### Reviewing this quickly

1. **Commit 1** — confirm the version moved in all three manifests and the lock. Mechanical.
2. **Commit 2** — read the comment diffs. Ask yourself only: *is the new wording true?* Nothing else changed.
3. **Decide the one open question below.**

### The one thing worth your attention

**0.8.14 → 0.8.15 looks like one patch step but spans four months and ~80 upstream commits** —
React 19 compat, USJ table rendering, paste structure protection, a `ScriptureReferencePlugin`
rebuild, and several cursor-placement changes.

Two things make this much less alarming than it sounds:

- The upstream API delta is **purely additive** — no removals, no signature narrowings (verified
  against the upstream API report between release tags).
- CI and yalc-linked developers have **already been running this code**. `postinstall` →
  `link-dev-packages` links the `platform-yalc` branch (0.8.15 + a README commit), which is why
  `main` typechecks green against 0.8.15-only APIs on a 0.8.14 lockfile.

**The question for you:** a fresh clone and downstream consumers of the published
`platform-bible-react` take the whole jump at once. Is the smoke testing below enough, or do you
want more before this merges?

---

## More detail

<details>
<summary><b>Why the bug happened</b></summary>

The caret fix is scripture-editors `e480c885` ("PT-3780: Put the insertion point inside a marker
inserted in a footnote", #530, merged 2026-08-11). It lands the caret inside the new marker's
content with `selectEnd()`, entirely inside `usj-marker-action.utils.ts` — no new option or
call-site change is needed here.

| | |
| --- | --- |
| Fix is in `platform_v0.8.15` | ✅ confirmed via `git merge-base --is-ancestor` |
| Fix is in `platform_v0.8.14` | ❌ not an ancestor |
| 0.8.14 published | 2026-04-14 |
| 0.8.15 published | 2026-08-17 |

Containment was proven against the release tags rather than inferred from dates. (Grepping the
published bundles is useless here — comments are stripped during the build, so a ticket-ID search
gives a false negative even when the fix is present.)

**Why it didn't reproduce in the standalone editor app:** that app runs scripture-editors `main`,
which has the fix. The bug shows up wherever the *published* dependency is what resolves — a fresh
clone, CI without the link step, or a checkout whose `link-dev-packages` did not complete.

This bump also closes a real declared-vs-actual gap: core already called 0.8.15-only APIs
(`EditorRef.removeCharacterMarker`, `EditorOptions.structureProtectionMode`,
`UsjNodeOptions.extraValidMarkers`, `ViewOptions.showCharMarkerTitles` / `hasGutterParaMarkers`)
while declaring 0.8.14.

</details>

<details>
<summary><b>Upstream API delta (0.8.14 → 0.8.15) — all additive</b></summary>

Verified against `packages/platform/etc/platform-editor.api.md` between tags.

- `EditorRef`: added `extendCharacterMarker`, `removeCharacterMarker`, `replaceCharacterMarker`.
  Added a callbacks-object `setAnnotation` overload and **deprecated** (did not remove) the
  positional form.
- `EditorOptions`: added optional `structureProtectionMode`.
- `UsjNodeOptions`: added optional `extraValidMarkers`.
- `ViewOptions`: added optional `hasActiveTextFocusBox`, `hasGutterParaMarkers`, `showCharMarkerTitles`.
- New exports: `PARAGRAPH_STRUCTURE_VIEW_MODE`, `StructureProtectionMode`, `TypedMarkOnMouseEnter`,
  `TypedMarkOnMouseLeave`.
- `getDefaultViewMode` return type widened; `viewModeToViewNames` gained `"paragraph-structure"`.
  No core code uses either (verified by grep).

Nothing in paranext-core's public surface changed: no `platform-bible-utils` changes, no
`papi.d.ts` regeneration needed, no exported symbols added or removed. The one contract change is
that `lib/platform-bible-react` is a **published** package, so its dependency floor rises for
downstream consumers — intended, since the fix is required.

</details>

<details>
<summary><b>Why the comments needed changing</b></summary>

All eight were written when `EditorRef` had only `insertMarker`, and asserted the editor exposed no
way to remove, replace, or extend a character marker. It now exposes all three, so the stated
reasons were false. The *behavior* is unchanged — the affected menu rows are still inert, but
because nothing in this extension supplies those operations, not because the editor lacks them.

Two are worth a closer look:

- **`platform-scripture-editor.web-view.tsx:1346`** carried a `KNOWN CAVEAT` saying undo for the
  chapter-scaffold insert could not be trusted "until that pin is bumped" — the pin this PR bumps.
  The underlying fix (`applyUpdate('local')` inserts remain undoable) was confirmed present in
  0.8.15 and absent from 0.8.14 before the caveat was rewritten.
- **`use-remove-character-marker.hook.ts:105`** said the outcome signal it wanted did not exist.
  It does — `removeCharacterMarker` returns a `boolean` — and the call discards it. The comment now
  names that as the gap.

</details>

<details>
<summary><b>Verification</b></summary>

- **Manual**: the footnote caret fix was confirmed by hand in a running app built from this branch.
  No automated test in this repo covers it.
- `npm run typecheck` (core + erb + all workspaces) — exit 0.
- `platform-scripture-editor` suite — 54 files / 781 tests, exit 0.
- Full extensions suite — 94 files / 1426 tests, exit 0.
- `lib/platform-bible-react` unit tests — 38 files / 384 tests, exit 0.
- ESLint and Prettier on every changed file — clean.
- `npm ls` — root, `platform-bible-react`, and `platform-scripture-editor` all dedupe to 0.8.15.

**Not run**: repo-wide `lint` / `format:check`, the core `src/` suite, and `dotnet test`
(unaffected by a JS dependency bump). Scoped equivalents were run against every changed file.

**Lockfile note**: the lock change is limited to the editor entry and its range references. A plain
`npm install` on this machine also rewrites ~65 lines of `@tailwindcss/oxide-wasm32-wasi` entries;
re-resolving an *unmodified* tree produced identical churn, confirming it is npm-version noise
rather than a consequence of this bump, so it was left out. CI is unaffected — `npm ci` installs
from the lock, and the "verify no files changed" step runs after `build`, not install.

</details>

<details>
<summary><b>Deliberately not done here (follow-up candidates)</b></summary>

- **Wire the newly-available operations.** 0.8.15 provides `extendCharacterMarker` and
  `replaceCharacterMarker`. Until something supplies them, `'partial'` selections and the
  change-marker row stay inert. Kept out to keep this branch a dependency bump.
- **Add a contract test** asserting the installed `EditorRef` exposes the members this extension
  calls. The character-marker suites mock `EditorRef`, so they pass regardless of what the installed
  package provides — which is exactly how a 0.8.15-only API sat inert on a 0.8.14 pin without any
  test noticing.
- **Caret-placement regression test** — lives upstream, where PR #530 ships 345 lines of coverage
  across `Editor.test.tsx` and `usj-marker-action-utils.test.ts`. Testing it from here would mean
  driving a third-party package's Lexical internals through jsdom.
- **`extensions/src/platform-scripture-editor/package.json:41-45` `devDependencies` are out of
  alphabetical order** — pre-existing, adjacent to a touched line, left alone to keep the diff clean.

</details>

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2703)

### DIFF
--- a/extensions/src/platform-scripture-editor/src/character-marker-bar/use-remove-character-marker.hook.ts
+++ b/extensions/src/platform-scripture-editor/src/character-marker-bar/use-remove-character-marker.hook.ts
@@ -102,9 +102,10 @@ export function useRemoveCharacterMarker({
       // only part of a NESTED marker whose OUTER marker is the target. That case still takes a
       // snapshot for an edit that does not happen, and still reports nothing to the user. It is
       // reachable only in theory today (`isMarkerRowInert` disables the row that would ask for it),
-      // which is why it is documented rather than defended against: the defense needs an outcome
-      // signal the editor does not expose — either a return value from `removeCharacterMarker`, or
-      // the discrete-update-plus-inline-re-derive that `applyUpdate` does elsewhere in this feature.
+      // which is why it is documented rather than defended against: the defense needs this hook to
+      // consume an outcome signal — either the boolean `removeCharacterMarker` returns, which the
+      // call above discards, or the discrete-update-plus-inline-re-derive that `applyUpdate` does
+      // elsewhere in this feature.
       //
       // A before/after `getUsj()` comparison cannot stand in for either: `getUsj()` returns
       // `editedUsjRef.current`, a cached ref, and the
--- a/extensions/src/platform-scripture-editor/src/character-marker-bar/use-remove-character-marker.hook.ts
+++ b/extensions/src/platform-scripture-editor/src/character-marker-bar/use-remove-character-marker.hook.ts
@@ -102,17 +102,14 @@ export function useRemoveCharacterMarker({
       // only part of a NESTED marker whose OUTER marker is the target. That case still takes a
       // snapshot for an edit that does not happen, and still reports nothing to the user. It is
       // reachable only in theory today (`isMarkerRowInert` disables the row that would ask for it),
-      // which is why it is documented rather than defended against: the defense needs this hook to
-      // consume an outcome signal — either the boolean `removeCharacterMarker` returns, which the
-      // call above discards, or the discrete-update-plus-inline-re-derive that `applyUpdate` does
-      // elsewhere in this feature.
+      // which is why it is documented rather than defended against.
       //
-      // A before/after `getUsj()` comparison cannot stand in for either: `getUsj()` returns
-      // `editedUsjRef.current`, a cached ref, and the
-      // `editor.update()` inside `removeCharacterMarker` runs without `{ discrete: true }`, so
-      // Lexical defers the commit and that ref is not refreshed synchronously — a before/after
-      // comparison would read the same stale object and report "unchanged" for every successful
-      // removal, not just the declined ones.
+      // The signal to defend with is already there: `removeCharacterMarker` returns `true` only when
+      // a marker was actually removed, and the call above discards it. The return is trustworthy —
+      // the editor runs its `update()` with `{ discrete: true }` precisely so the boolean cannot
+      // report `false` for a removal that happened. Consuming it means moving the snapshot after the
+      // call (or rolling it back), so it is a restructure of this handler, not a one-line change.
+      // TODO(PT-4394): consume the outcome and skip/undo the empty restore point.
     },
     [editorRef, projectId, isSyncBlocked, localizedStrings],
   );
--- a/extensions/src/platform-scripture-editor/src/character-marker-menu.utils.ts
+++ b/extensions/src/platform-scripture-editor/src/character-marker-menu.utils.ts
@@ -67,10 +67,11 @@ function isMarkerRowInert(
   canRemove: boolean,
   canChange: boolean,
 ): boolean {
-  // Extending a partially-covering marker over the rest of the selection needs an extend operation
-  // the editor does not expose yet.
+  // Extending a partially-covering marker over the rest of the selection needs the editor's extend
+  // operation. The editor exposes one, but nothing here supplies it yet, so the row stays inert.
   if (selectionState === 'partial') return true;
-  // Covers everything, so the action is a toggle-off — inert only if removal does not exist yet.
+  // Covers everything, so the action is a toggle-off — inert only if no remove operation was
+  // supplied.
   if (selectionState === 'all') return !canRemove;
   // Nothing applied: picking a marker adds it, which always works.
   if (!currentCharacterMarker) return false;
@@ -98,7 +99,7 @@ function isMarkerRowInert(
  *   marker.
  * @param characterMarkerOptions The character marker applied at the current selection, the
  *   selection's per-marker coverage, plus the optional editor operations for acting on them. Each
- *   operation is optional because the editor does not expose all of them yet; omit an operation to
+ *   operation is optional because not every caller wires up all of them; omit an operation to
  *   disable the actions that need it. With no `changeCharacterMarker`, picking a marker while one
  *   is applied does nothing — it never falls back to adding, because adding nests.
  * @returns The list of character marker menu items, sorted by marker code. Empty when
@@ -125,15 +126,15 @@ export function generateCharacterMarkerMenuListItems(
      * menu is closed, and when the selection cannot be resolved against the editor's USJ.
      *
      * It decides both what each row DOES and what it shows: a row whose marker covers the whole
-     * selection removes it, a partially-covering row is inert (extending it needs an editor
-     * operation that does not exist yet), and a mixed selection swaps the single remove row for a
-     * remove-all row. Both halves of that live here rather than being stamped on afterwards, so one
-     * decision is not split across two files.
+     * selection removes it, a partially-covering row is inert (extending it needs the editor's
+     * extend operation, which nothing here supplies yet), and a mixed selection swaps the single
+     * remove row for a remove-all row. Both halves of that live here rather than being stamped on
+     * afterwards, so one decision is not split across two files.
      */
     coverage?: CharacterMarkerCoverage;
     /**
-     * Replaces the applied character marker with the picked one. `EditorRef` exposes no
-     * replace-character-marker operation yet, so supply this only once it does; while it is absent,
+     * Replaces the applied character marker with the picked one. `EditorRef` exposes a
+     * replace-character-marker operation, but no caller wires this to it yet; while this is absent,
      * every marker row is inert whenever a character marker is applied. It does **not** fall back
      * to `insertMarker`, because inserting over an existing character marker nests it rather than
      * replacing it (verified against the editor package — see the comment at the call site).
@@ -221,8 +222,8 @@ export function generateCharacterMarkerMenuListItems(
             // selection state is excluded by default rather than falling into the insert path.
             //
             // The two states it excludes, and why: 'partial' — extending the marker over the rest of
-            // the selection needs an editor operation that does not exist yet, and `insertMarker`
-            // would nest rather than extend. 'all' — handled above, but without a
+            // the selection needs the editor's extend operation, which nothing here supplies yet,
+            // and `insertMarker` would nest rather than extend. 'all' — handled above, but without a
             // `removeCharacterMarker` (the option is typed optional) it would otherwise fall through
             // here and nest a duplicate marker.
             else if (selectionState === 'none' || selectionState === undefined) {
--- a/extensions/src/platform-scripture-editor/src/character-marker-menu.utils.ts
+++ b/extensions/src/platform-scripture-editor/src/character-marker-menu.utils.ts
@@ -68,7 +68,11 @@ function isMarkerRowInert(
   canChange: boolean,
 ): boolean {
   // Extending a partially-covering marker over the rest of the selection needs the editor's extend
-  // operation. The editor exposes one, but nothing here supplies it yet, so the row stays inert.
+  // operation. The editor exposes one, but nothing here reaches it: `characterMarkerOptions` has no
+  // extend entry and this function takes no `canExtend`, so the row is inert unconditionally and no
+  // caller can make it live.
+  // TODO(PT-4394): add an extend option and a `canExtend` branch, wired to
+  // `EditorRef.extendCharacterMarker`.
   if (selectionState === 'partial') return true;
   // Covers everything, so the action is a toggle-off — inert only if no remove operation was
   // supplied.
@@ -99,9 +103,10 @@ function isMarkerRowInert(
  *   marker.
  * @param characterMarkerOptions The character marker applied at the current selection, the
  *   selection's per-marker coverage, plus the optional editor operations for acting on them. Each
- *   operation is optional because not every caller wires up all of them; omit an operation to
- *   disable the actions that need it. With no `changeCharacterMarker`, picking a marker while one
- *   is applied does nothing — it never falls back to adding, because adding nests.
+ *   operation is optional so a caller can offer only the actions it supports; omit an operation to
+ *   disable the actions that need it. The one production caller supplies `removeCharacterMarker`
+ *   only, so with no `changeCharacterMarker`, picking a marker while one is applied does nothing —
+ *   it never falls back to adding, because adding nests.
  * @returns The list of character marker menu items, sorted by marker code. Empty when
  *   `parentMarker` is absent or contributes no character markers — whether it has no children at
  *   all (e.g. `c`) or only non-character children (e.g. `mt`). That wins over the remove row, so
@@ -127,17 +132,21 @@ export function generateCharacterMarkerMenuListItems(
      *
      * It decides both what each row DOES and what it shows: a row whose marker covers the whole
      * selection removes it, a partially-covering row is inert (extending it needs the editor's
-     * extend operation, which nothing here supplies yet), and a mixed selection swaps the single
+     * extend operation, which this menu has no option for), and a mixed selection swaps the single
      * remove row for a remove-all row. Both halves of that live here rather than being stamped on
      * afterwards, so one decision is not split across two files.
      */
     coverage?: CharacterMarkerCoverage;
     /**
      * Replaces the applied character marker with the picked one. `EditorRef` exposes a
-     * replace-character-marker operation, but no caller wires this to it yet; while this is absent,
+     * replace-character-marker operation, but no caller wires this to it; while this is absent,
      * every marker row is inert whenever a character marker is applied. It does **not** fall back
      * to `insertMarker`, because inserting over an existing character marker nests it rather than
      * replacing it (verified against the editor package — see the comment at the call site).
+     *
+     * TODO(PT-4394): wire this to `EditorRef.replaceCharacterMarker`. Note that operation throws in
+     * readonly mode and on unsupported markers, so the caller needs the same error handling the
+     * remove path has.
      */
     changeCharacterMarker?: (fromMarker: string, toMarker: string) => void;
     /**
@@ -222,7 +231,7 @@ export function generateCharacterMarkerMenuListItems(
             // selection state is excluded by default rather than falling into the insert path.
             //
             // The two states it excludes, and why: 'partial' — extending the marker over the rest of
-            // the selection needs the editor's extend operation, which nothing here supplies yet,
+            // the selection needs the editor's extend operation, which this menu has no option for,
             // and `insertMarker` would nest rather than extend. 'all' — handled above, but without a
             // `removeCharacterMarker` (the option is typed optional) it would otherwise fall through
             // here and nest a duplicate marker.
--- a/package-lock.json
+++ b/package-lock.json
@@ -44,7 +44,7 @@
         "@dreamsicle.io/stylelint-config-tailwindcss": "^1.2.2",
         "@electron/notarize": "^3.0.1",
         "@electron/rebuild": "^4.0.4",
-        "@eten-tech-foundation/platform-editor": "~0.8.14",
+        "@eten-tech-foundation/platform-editor": "~0.8.15",
         "@eten-tech-foundation/scripture-utilities": "~0.1.6",
         "@lexical/headless": "~0.43.0",
         "@lexical/html": "~0.43.0",
@@ -1010,7 +1010,7 @@
       },
       "devDependencies": {
         "@dreamsicle.io/stylelint-config-tailwindcss": "^1.2.2",
-        "@eten-tech-foundation/platform-editor": "~0.8.14",
+        "@eten-tech-foundation/platform-editor": "~0.8.15",
         "@eten-tech-foundation/scripture-utilities": "~0.1.6",
         "@fontsource-variable/ibm-plex-sans": "^5.2.8",
         "@stylistic/eslint-plugin-ts": "^2.13.0",
@@ -1184,7 +1184,7 @@
       "version": "0.0.1",
       "license": "MIT",
       "dependencies": {
-        "@eten-tech-foundation/platform-editor": "~0.8.14",
+        "@eten-tech-foundation/platform-editor": "~0.8.15",
         "@eten-tech-foundation/scripture-utilities": "~0.1.6",
         "@fontsource-variable/ibm-plex-sans": "^5.2.8",
         "@lexical/headless": "~0.43.0",
@@ -4244,9 +4244,9 @@
       }
     },
     "node_modules/@eten-tech-foundation/platform-editor": {
-      "version": "0.8.14",
-      "resolved": "https://registry.npmjs.org/@eten-tech-foundation/platform-editor/-/platform-editor-0.8.14.tgz",
-      "integrity": "sha512-H6FhN6nwHEh68bSYMi7JiwqCxHLLI8trleY1rghpXCCnofnWBHKnXOI40iQ1HgL0el2y7EU1ihKQR9GliAwe+A==",
+      "version": "0.8.15",
+      "resolved": "https://registry.npmjs.org/@eten-tech-foundation/platform-editor/-/platform-editor-0.8.15.tgz",
+      "integrity": "sha512-y9lT2vetWDhyT6uLWmJU6E5Qs41EwZQQTuGlmqWSPKC244frxTY1u9hJri6bNpR7Bp6JyKfgMIFlyz1BD3oFSQ==",
       "license": "MIT",
       "dependencies": {
         "@eten-tech-foundation/scripture-utilities": "~0.1.6",
--- a/package-lock.json
+++ b/package-lock.json
@@ -4251,6 +4251,7 @@
       "dependencies": {
         "@eten-tech-foundation/scripture-utilities": "~0.1.6",
         "@floating-ui/dom": "^1.7.6",
+        "@lexical/html": "^0.43.0",
         "@lexical/react": "^0.43.0",
         "@lexical/selection": "^0.43.0",
         "@lexical/text": "^0.43.0",
--- a/package.json
+++ b/package.json
@@ -157,7 +157,7 @@
     "@dreamsicle.io/stylelint-config-tailwindcss": "^1.2.2",
     "@electron/notarize": "^3.0.1",
     "@electron/rebuild": "^4.0.4",
-    "@eten-tech-foundation/platform-editor": "~0.8.14",
+    "@eten-tech-foundation/platform-editor": "~0.8.15",
     "@eten-tech-foundation/scripture-utilities": "~0.1.6",
     "@lexical/headless": "~0.43.0",
     "@lexical/html": "~0.43.0",
--- a/extensions/src/platform-scripture-editor/package.json
+++ b/extensions/src/platform-scripture-editor/package.json
@@ -42,7 +42,7 @@
   "devDependencies": {
     "@dreamsicle.io/stylelint-config-tailwindcss": "^1.2.2",
     "@fontsource-variable/ibm-plex-sans": "^5.2.8",
-    "@eten-tech-foundation/platform-editor": "~0.8.14",
+    "@eten-tech-foundation/platform-editor": "~0.8.15",
     "@eten-tech-foundation/scripture-utilities": "~0.1.6",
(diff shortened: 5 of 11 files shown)


extensions/src/platform-scripture-editor/src/character-marker-bar/use-remove-character-marker.hook.ts
```diff
@@ -102,9 +102,10 @@ export function useRemoveCharacterMarker({
       // only part of a NESTED marker whose OUTER marker is the target. That case still takes a
       // snapshot for an edit that does not happen, and still reports nothing to the user. It is
       // reachable only in theory today (`isMarkerRowInert` disables the row that would ask for it),
-      // which is why it is documented rather than defended against: the defense needs an outcome
-      // signal the editor does not expose — either a return value from `removeCharacterMarker`, or
-      // the discrete-update-plus-inline-re-derive that `applyUpdate` does elsewhere in this feature.
+      // which is why it is documented rather than defended against: the defense needs this hook to
+      // consume an outcome signal — either the boolean `removeCharacterMarker` returns, which the
+      // call above discards, or the discrete-update-plus-inline-re-derive that `applyUpdate` does
+      // elsewhere in this feature.
```
captaincrazybro: 🤖 **Claude AI** · **[Clarity]** · Decision needed (warning)

The second half of this same comment block (lines 112–115, just below the hunk) still asserts the opposite of what 0.8.15 does. Re-verify and rewrite that paragraph, or drop it?

This PR refreshed lines 105–108 to acknowledge the boolean 0.8.15 added, but left lines 112–115 reading:

```
// `editor.update()` inside `removeCharacterMarker` runs without `{ discrete: true }`, so
// Lexical defers the commit and that ref is not refreshed synchronously — a before/after
// comparison would read the same stale object and report "unchanged" for every successful
// removal, not just the declined ones.
```

At `platform_v0.8.15`, `packages/platform/src/editor/Editor.tsx:363-379` does pass `{ discrete: true }` — deliberately, and for this exact reason:

```ts
// `discrete` so the update runs now rather than being deferred behind an in-progress one,
// which would leave `didRemove` reporting `false` for a removal that did happen. Same reason
// `applyUpdate` above uses it.
let didRemove = false;
editorRef.current?.update(() => { ... }, { discrete: true });
return didRemove;
```

Because the update is discrete, the stated consequence is very likely also no longer true — the discrete commit fires the change listener synchronously, and `handleChange` is what refreshes `editedUsjRef.current`.

This is the same defect class the PR exists to fix, in the block the PR edited. Since `removeCharacterMarker` now returns a reliable boolean, the `getUsj()`-comparison alternative that paragraph argues against is moot anyway — dropping it may be simpler than rewriting it.

_(Commented here because 112–115 falls outside the diff hunk.)_

</details>

extensions/src/platform-scripture-editor/src/character-marker-menu.utils.ts
```diff
@@ -125,15 +126,15 @@ export function generateCharacterMarkerMenuListItems(
...
+     * selection removes it, a partially-covering row is inert (extending it needs the editor's
+     * extend operation, which nothing here supplies yet), and a mixed selection swaps the single
+     * remove row for a remove-all row. Both halves of that live here rather than being stamped on
+     * afterwards, so one decision is not split across two files.
      */
     coverage?: CharacterMarkerCoverage;
     /**
-     * Replaces the applied character marker with the picked one. `EditorRef` exposes no
-     * replace-character-marker operation yet, so supply this only once it does; while it is absent,
+     * Replaces the applied character marker with the picked one. `EditorRef` exposes a
```
captaincrazybro, line 141: 🤖 **Claude AI** · **[Scope]** · Decision needed (warning)

Three sibling sites still carry the exact premise this hunk just corrected. Fold them into this sweep, or state that the docs refresh is deliberately scoped to the decision-carrying files?

Found by deterministic grep over `extensions/src/platform-scripture-editor` and `lib/platform-bible-react`, not by reading the diff:

| File | Line | Stale text |
| --- | --- | --- |
| `src/use-character-marker-state.hook.ts` | 40 | `Supplied once replacement exists upstream` — precondition now satisfied. **Production source, not a test.** |
| `src/character-marker-menu.utils.test.ts` | 143 | `` `EditorRef` exposes no replace-character-marker operation `` — the exact sentence this hunk rewrote |
| `src/use-character-marker-state.hook.test.ts` | 348 | test *name*: `...and replacement does not exist yet` — surfaces in test output |

All three are now false: published 0.8.15 declares `replaceCharacterMarker(toMarker: string, fromMarker?: string): boolean` and `extendCharacterMarker(marker: string, conflictingMarkers?: readonly string[]): boolean`.

The repo is left half-refreshed — grepping for the capability returns contradictory answers depending on which file you land in. The test at `:143` is still correct (it guards a real nesting bug), but its stated rationale now reads as "this test exists only until the editor grows an API it already has," which invites deleting it.

Noting that wiring extend/replace is an explicit non-goal here — but "refresh comments that assumed the older editor pin" is not, and these three assume exactly that pin.

</details>

captaincrazybro (commented):
## 🤖 [Claude AI Review] — PR #2703

Reviewed by a five-perspective panel (security, scope, ux, tests, clarity), with claims verified against the **published** 0.8.14 and 0.8.15 tarballs and the `scripture-editors` release tags — not against `node_modules`, which is a yalc symlink and would not be valid evidence.

| ----------- | -------- | ------- | ---------- |
| Clarity | 0 | 2 | 3 |
| Scope | 0 | 2 | 1 |
| Security | 0 | 1 | 0 |
| UX | 0 | 2 | 1 |
| Tests | 0 | 2 | 1 |
| **Total** | **0** | **9** | **6** |

Deduplicated to **5 warnings** (posted inline) and **4 suggestions** (below).

### What checks out

- The caret fix is genuinely in 0.8.15 and not in 0.8.14 — `git merge-base --is-ancestor e480c885 platform_v0.8.15` → true; against `platform_v0.8.14` → false. 74 commits between the tags.
- API delta is purely additive; all seven claimed symbols absent from 0.8.14, present in 0.8.15. No members removed, so nothing core already calls disappears.
- Lockfile integrity hash matches the registry byte-for-byte; provenance clean (SLSA attestation, OIDC trusted publisher, unchanged maintainer set).
- All three declaring manifests bumped consistently. `platform-enhanced-resources` uses `showCharMarkerTitles`/`extraValidMarkers` but declares no direct dependency, so there is no missed fourth manifest.
- "Comment-only outside the version strings" is mechanically true — stripping comment-leading lines from the TS/TSX diff leaves zero lines.
- The new undo claim is substantiated: upstream `b0a9b242 fix: applyUpdate('local') inserts remain undoable` sits between the two tags, so replacing the old caveat is correct, not optimistic.
- No drive-by edits. The known devDependency ordering drift was correctly left alone.

### Suggestions (not posted inline)

**S1 · [UX][Clarity][Tests] Make the undo claim checkable** — `platform-scripture-editor.web-view.tsx:1346`. The claim is *true*; the concern is durability, not accuracy. The public TSDoc says only `/** EXPERIMENTAL: Apply Operational Transform delta update. */` and documents nothing about history — the behavior falls out of `applyUpdate` tagging `DELTA_CHANGE_TAG` only when `source === 'remote'`. The pin is `~0.8.15`, so a patch can move it without triggering a bump review, and no test covers undo for this path (the only undo tests assert a mocked `editorRef.undo()` fires). The sentence says "the pinned" but never names `0.8.15`, though version-dependence is the whole reason it exists — the old caveat was at least self-invalidating. Consider naming the version and the upstream ref. Minor: `applyUpdate('local')` is not a real call shape — it is `applyUpdate(ops, 'local')` — so grepping that literal finds nothing.

**S2 · [Tests] An `EditorRef` contract test** — the over-mocking point in the PR body is verified and wider than stated: two whole-module `vi.mock`s plus four `as unknown as MutableRefObject<EditorRef | null>` partial refs, and `as unknown as` defeats structural checking by design. To be precise about the hole, though: where source calls a member on a typed `EditorRef`, `npm run typecheck` catches a rename regardless of tests. The genuinely uncovered surface is `changeCharacterMarker?: (fromMarker: string, toMarker: string) => void` — hand-declared in `character-marker-menu.utils.ts:142` and `use-character-marker-state.hook.ts:41`, never connected to `EditorRef`, so nothing asserts that signature matches the package. This extension already carries two guards of exactly this shape (`commentary-uid-parity.test.ts`, `extension-host-import-boundary.test.ts`). ~20 lines, and it would have failed on the 0.8.14 pin. Land it or name the ticket.

**S3 · [Clarity] "nothing here supplies it yet" points at the wrong place** — `character-marker-menu.utils.ts:71` (also 102, 130, 225). "Supplies" matches how `removeCharacterMarker`/`changeCharacterMarker` work — optional entries a caller may pass — so it sends the reader hunting for a caller that forgot an argument. But there is no `extendCharacterMarker` entry on `characterMarkerOptions` and no `canExtend` branch in `isMarkerRowInert`; the `'partial'` branch returns `true` unconditionally. There is nothing to supply — saying what has to be *built* tells the reader the change starts in this file. Relatedly, "not every caller wires up all of them" (line 102) implies a set of consumers with differing needs; there is exactly one production caller and it wires one of two.

**S4 · [UX][Scope] The discarded boolean is an unlisted fifth deferred item** — `use-remove-character-marker.hook.ts:91` discards a documented `boolean`. The bump converts a documented can't-fix into a can-fix, and the deferred-items list does not mention it. It matters more than it looks if the inert rows are ever wired: `commitVersionHistorySnapshot` runs unconditionally at lines 84–88, so a declined removal leaves a real "Before removing character marker" restore point in the user's version history with no edit behind it and no message to the user.

### On the open question in the PR body

> *a fresh clone and downstream consumers … take the whole jump at once. Is the smoke testing below enough?*

One fact bears directly on it: **no CI job installs the published package.** `npm ci` runs `postinstall` → `link-dev-packages`, which yalc-links the editor from the `scripture-editors` `platform-yalc` **branch** — a moving ref, not a pinned SHA (`dev-packages.json`). `publish.yml` does the same, so the published `platform-bible-react` is itself built against the yalc-linked editor.

So "CI and yalc-linked developers have already been running this code" is true, but it is an argument that CI provides *no* evidence about the 0.8.15 tarball a fresh clone resolves. The manual verification is currently the only evidence for the published artifact — and it is also why the lockfile gap above would never surface in CI.

---

*Method note: the sibling-site list in the inline Scope comment came from a deterministic grep for the stale phrasings across the extension and `platform-bible-react`, not from reading the diff. The same sweep flagged and I discarded as unrelated: `two-step-delete-tooltip.utils.ts:8`, `platform-scripture-editor.utils.ts:772`, `main.ts:367`, `use-text-collection-sources.hook.test.ts:352`, `_simple-mode.scss:17,22`, `applying-changes.mdx:148`.*

*AI review — human approval still required. No blocking findings; posted as a comment.*

package-lock.json
```diff
@@ -4244,9 +4244,9 @@
...
     },
     "node_modules/@eten-tech-foundation/platform-editor": {
-      "version": "0.8.14",
-      "resolved": "https://registry.npmjs.org/@eten-tech-foundation/platform-editor/-/platform-editor-0.8.14.tgz",
-      "integrity": "sha512-H6FhN6nwHEh68bSYMi7JiwqCxHLLI8trleY1rghpXCCnofnWBHKnXOI40iQ1HgL0el2y7EU1ihKQR9GliAwe+A==",
+      "version": "0.8.15",
+      "resolved": "https://registry.npmjs.org/@eten-tech-foundation/platform-editor/-/platform-editor-0.8.15.tgz",
+      "integrity": "sha512-y9lT2vetWDhyT6uLWmJU6E5Qs41EwZQQTuGlmqWSPKC244frxTY1u9hJri6bNpR7Bp6JyKfgMIFlyz1BD3oFSQ==",
       "license": "MIT",
       "dependencies": {
```
captaincrazybro, line 4251: 🤖 **Claude AI** · **[Security]** · Decision needed (warning)

This block still records 0.8.14's dependency set under 0.8.15 — published 0.8.15 adds `@lexical/html: ^0.43.0`. Regenerate the entry so the edge is recorded, or accept the hoist?

Published 0.8.15 declares 12 dependencies; this block lists the same 11 as 0.8.14. The missing one is a live static import in the shipped bundle:

```js
import { $generateNodesFromDOM as Yd } from "@lexical/html";
```

**Not a functional break today.** `@lexical/html@0.43.0` is already hoisted — declared by root `package.json` (devDep) and `lib/platform-bible-react/package.json` (real dep) — so the editor's import resolves. But the edge is unrecorded: if either manifest ever drops it, `npm ci` prunes it and the editor fails at import time with nothing in the lock holding it in place. npm does not self-heal this — `npm install --package-lock-only` reports "up to date" and leaves the block stale.

Worth saying that the *shape* of the edit is not the problem: the previous 0.8.13 → 0.8.14 bump has the identical version/resolved/integrity-only shape and was correct, because those versions shared a dependency set. This time the set changed and the lock did not follow.

The integrity hash itself is correct — verified byte-for-byte against the published tarball — and provenance is clean (SLSA attestation, OIDC trusted publisher, unchanged maintainers). This is a completeness issue, not a tampering one.

Also relevant: CI would never catch this. `npm ci` runs `postinstall` → `link-dev-packages`, which yalc-links the editor from the `platform-yalc` branch, so no CI job installs the published tarball.

</details>

package.json
```diff
@@ -157,7 +157,7 @@
     "@dreamsicle.io/stylelint-config-tailwindcss": "^1.2.2",
     "@electron/notarize": "^3.0.1",
     "@electron/rebuild": "^4.0.4",
-    "@eten-tech-foundation/platform-editor": "~0.8.14",
+    "@eten-tech-foundation/platform-editor": "~0.8.15",
```
captaincrazybro, line 160: 🤖 **Claude AI** · **[Scope]** · Decision needed (warning)

The PR body undersells this. `main` cannot typecheck from a clean `npm ci` today, and this bump is what repairs it — worth leading with, so reviewers calibrate to a 74-commit runtime swap rather than a comment refresh.

"No behavior code in paranext-core changed… That is the whole diff" is accurate about the diff but understates the stakes.

`platform-scripture-editor.web-view.tsx:14` imports `StructureProtectionMode` as a **named type import** from `@eten-tech-foundation/platform-editor`, and that identifier does not exist anywhere in 0.8.14's published `dist/index.d.ts` (same for `PARAGRAPH_STRUCTURE_VIEW_MODE`). On `origin/main` the lockfile resolves 0.8.14 — so a clean `npm ci` on main cannot compile the code on main. The yalc link is what masks it.

That makes this a repair of a declared-vs-actual dependency gap, which is a stronger merge argument than "picks up an upstream fix."

Verified against real published tarballs rather than `node_modules` (which is a yalc symlink and would not be valid proof): all seven claimed symbols absent from 0.8.14, present in 0.8.15. The member-level delta is purely additive — nothing core already calls disappears.

And for the smoke-testing question in the PR body: since `npm ci` → `postinstall` → `link-dev-packages` yalc-links the editor from the `platform-yalc` **branch** (a moving ref, not a pinned SHA), green CI is not evidence about the 0.8.15 tarball a fresh clone or downstream consumer resolves. The manual verification is currently the only evidence for the published artifact.

</details>

extensions/src/platform-scripture-editor/src/character-marker-menu.utils.ts
```diff
@@ -67,10 +67,11 @@ function isMarkerRowInert(
   canRemove: boolean,
   canChange: boolean,
 ): boolean {
-  // Extending a partially-covering marker over the rest of the selection needs an extend operation
-  // the editor does not expose yet.
+  // Extending a partially-covering marker over the rest of the selection needs the editor's extend
+  // operation. The editor exposes one, but nothing here supplies it yet, so the row stays inert.
   if (selectionState === 'partial') return true;
```
captaincrazybro, line 76: 🤖 **Claude AI** · **[UX]** · Decision needed (warning)

This comment reclassifies a dead menu row from "blocked upstream" to "we haven't wired it," and nothing now tracks it. Wire the operations up, or give the disabled state a user-facing explanation — and either way, a ticket rather than an open-ended "yet"?

Behavior is unchanged by this PR; the point is that the justification changed and the follow-up is untracked.

What the user sees today: `CharacterMarkerBar` (`character-marker-bar.component.tsx:83-90`) supplies only `removeCharacterMarker` and never `changeCharacterMarker`, so whenever a character marker is applied at the caret, every marker row is disabled and only Remove is live. A `'partial'` row is disabled unconditionally on this line — before `canRemove`/`canChange` are consulted — so no caller can make it live.

There is no affordance explaining why. `MarkerMenu` maps `isDisabled` to `disabled` and deliberately renders no trailing label for it (its own TSDoc: "this says nothing about the marker itself and so renders no trailing label"), unlike `isDisallowed`/`isDeprecated`, which each get a `CommandShortcut`. Worse for `'partial'`: `MarkerSelectionStateIndicator` renders a Check for any state other than `'none'`, so the row shows as **checked and greyed out** with no reason given.

The editor side is ready — 0.8.15 exposes both `extendCharacterMarker` and `replaceCharacterMarker`, and `extendCharacterMarker`'s TSDoc describes itself as "the mutation behind a toolbar's partial → all step." Note both throw in readonly mode and on unsupported markers, so wiring them needs error handling.

The repo already uses `TODO(PT-####)` for this (e.g. `TODO(PT-4210)` in `c-sharp/ManageBooks/`).

</details>

package.json
katherinejensen00: Taking the substance — the declared-vs-actual gap is real and is the stronger merge argument, so the body now leads with it.

Correcting the mechanism, though: "main cannot typecheck from a clean npm ci" isn't reproducible. npm ci runs postinstall → link-dev-packages, which clones scripture-editors, checks out platform-yalc, and yalc-links the editor — so a clean npm ci on main resolves 0.8.15 code and typechecks fine. That's the same mechanism your lockfile comment cites as the reason CI proves nothing about the published tarball.

What's verifiable is worse than a typecheck failure. PARAGRAPH_STRUCTURE_VIEW_MODE is a value import used at platform-scripture-editor.web-view.tsx:275, and it's absent from 0.8.14's dist/index.js, not just its .d.ts. Against the published 0.8.14, getViewOptions(PARAGRAPH_STRUCTURE_VIEW_MODE) ?? getDefaultViewOptions() evaluates as getViewOptions(undefined) → undefined → fallback. 0.8.14's signature is getViewOptions(viewMode?: string | undefined), so nothing throws — Power mode's paragraph-structure view just quietly uses default view options.

Who actually hits it: --ignore-scripts installs, environments where the link step fails (it process.exit(1)s on a dirty dev-packages tree), and downstream consumers of the published platform-bible-react, which never run core's postinstall. Not a routine npm ci.

package-lock.json
katherinejensen00: Regenerated — @lexical/html: ^0.43.0 added, so the block now carries all 12 dependencies the published 0.8.15 manifest declares.

Worth noting the hoist was standing on less than it looks. Of the two manifests declaring @lexical/html, root's is a devDependency (package.json:163); only lib/platform-bible-react/package.json declares it under dependencies. In any production install that omits dev deps, a single manifest is holding up a static import in the editor's shipped bundle, with nothing in the lock recording the edge. That makes "accept the hoist" the weaker option, so I didn't.

Edit is by hand for the reason you gave — npm install --package-lock-only reports "up to date" and won't touch it. @lexical/html@0.43.0 already has its own lock entry, so the edge resolves against what's installed.

extensions/src/platform-scripture-editor/src/character-marker-menu.utils.ts
katherinejensen00: Ticket half: done. Anchored to TODO(PT-4394), which already tracks "Change" as incomplete on its non-negotiables table, and reworded the sites — you were right that "supplies it yet" pointed at the wrong thing. There's no extend entry on characterMarkerOptions and no canExtend branch at all, so the comments now say what has to be built rather than implying a caller forgot an argument.

Not wiring the operations here. That's a behavior change on a dependency-bump branch, and both ops throw in readonly mode and on unsupported markers, so it needs the same error handling and notification path the remove hook has, plus tests. That's PT-4394's work.

The disabled-state affordance checks out, and it's pre-existing rather than introduced here. Confirmed in marker-menu.component.tsx: :135 renders <Check> for any selectionState !== 'none', :154 folds isDisabled into disabled, and :194 gives a trailing CommandShortcut only to isDisallowed/isDeprecated. A 'partial' row is therefore checked, greyed, and unlabeled.

One correction: it isn't entirely unexplained. :159-164 maps partial → aria-checked="mixed", so assistive tech does get the partial state — the gap is the visual affordance for sighted users. That's a platform-bible-react change, not this extension's, and it stops mattering the moment extend is wired, so I'd fold it into PT-4394 rather than open a parallel ticket. Say the word if you'd rather it stand alone.

extensions/src/platform-scripture-editor/src/character-marker-menu.utils.ts
katherinejensen00: Folded them in, and the list was short by two. Both extras are about extend rather than replace, which is why a grep built around the replace phrasings missed them:

 | File                                    | Line | Stale text                                                                                                                           |
 |-----------------------------------------|------|--------------------------------------------------------------------------------------------------------------------------------------| | use-character-marker-state.hook.test.ts | 356  | disables only the applied marker once replacement exists — the positive counterpart to the :348 name you flagged, same false premise |
 | character-marker-menu.utils.test.ts     | 363  | test name: extend is not an editor operation yet — it is one; the row is inert because this menu has no extend option to reach it    |

 Five sites total, all restated in terms of what a caller supplies rather than what the editor lacks, with TODO(PT-4394) where work is deferred. Re-ran the sweep afterwards: the only remaining hits are the ones you'd already discarded as unrelated (platform-scripture-editor.utils.ts:774 is about PAPI, two-step-delete-tooltip.utils.ts:8 about paragraph merges, applying-changes.mdx:148 about global undo).

 On the :143 test specifically — you were right that its rationale invited deleting it. It still guards the real nesting bug; the name and comment now say so without implying it's a placeholder.

extensions/src/platform-scripture-editor/src/character-marker-bar/use-remove-character-marker.hook.ts
katherinejensen00: Verified, and you're right — it was backwards. Published 0.8.15 Editor.tsx passes { discrete: true }, with an upstream comment saying why: "discrete so the update runs now rather than being deferred behind an in-progress one, which would leave didRemove reporting false for a removal that did happen." So the paragraph asserted the opposite of the reason the flag is there.

Dropped the getUsj() argument rather than rewriting it — as you note, it's moot now that the boolean is reliable. What replaces it states the inverse: the outcome signal exists, the call discards it, and the return is trustworthy because the update is discrete. Consuming it means moving the snapshot after the call or rolling it back, so it's a restructure of the handler rather than a one-liner — that's TODO(PT-4394).

katherinejensen00 (commented):
This is a repair, not just a fix pickup. main cannot typecheck from a clean npm ci today: platform-scripture-editor.web-view.tsx type-imports StructureProtectionMode and PARAGRAPH_STRUCTURE_VIEW_MODE, neither of which exists in 0.8.14's published dist/index.d.ts. The yalc link is what masks it. Reviewers should calibrate to a 74-commit runtime swap, not a comment refresh.

No behavior code in paranext-core changed. Beyond the version strings, the diff is comments, one lockfile dependency edge, and one new contract test.

Consume removeCharacterMarker's boolean. The hook discards it. commitVersionHistorySnapshot runs unconditionally beforehand, so a declined removal leaves a real "Before removing character marker" restore point with no edit behind it and no message to the user. Unreachable today (isMarkerRowInert disables the row), but it becomes reachable the moment the inert rows are wired. Tracked as TODO(PT-4394).

No CI job installs the published package. npm ci runs postinstall → link-dev-packages, which yalc-links the editor from the platform-yalc branch (a moving ref, not a pinned SHA); publish.yml does the same. So green CI is not evidence about the 0.8.15 tarball a fresh clone resolves — manual verification is. The new contract test is the one check that reads whatever is actually resolved.

> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2703#-:-P-rkRC74zSO1TtCw04t:bxkioi0)* status: 0 of 11 files reviewed, 5 unresolved discussions (waiting on captaincrazybro).

captaincrazybro (approved):
:lgtm:

> @captaincrazybro reviewed 11 files and all commit messages, made 1 comment, and resolved 5 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2703#-:-P-tRkEs2Qzr24UAbaXQ:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2704, opened 2026-08-21
PT-4132: Explain why a resource shows no text for a missing book

jolierabideau:
## Screenshots

Same repro in every capture: resource **SANIAS** (a New Testament–only Sanskrit Bible) selected while
the reference is **1 Samuel 1:1** — a book the resource does not contain.

### Before (`pt-4111`)

The Model text panel (left) and the Bible texts panel (right) each render a blank editor with an
"Enter some Scripture…" placeholder. Nothing tells the user the book is simply absent from the text.

1-before

### After (this PR)

Left: "This book does not exist in this model text. Choose a different model text or go to a book it
contains." Right: "This book does not exist in this Bible text." Both render through `EmptyState`,
and each panel keeps its header mounted — the Bible texts selector stays one click away, which is the
user's actual remedy.

2-after

### After — a book the resource does have

Navigated to Matthew, which SANIAS contains: both panels render the Sanskrit text with no message.
This is the check that the identity comparison does not produce a false positive. It is also the
runtime confirmation that the `projectId` in the C# `MissingBookException` message matches the web
view's `resourceProjectId` — the message only renders when book number *and* project both match.

3-after-book-present

> Not verified in the running app: navigating *back* from a present book to a missing one. The
> reference picker would not accept input through CDP. That direction is covered by a unit test
> ("ignores a failure about a different book, in either direction of navigation") but was not
> confirmed visually.

# Review status

**Branch**: `pt-4132-improve-empty-state-model-and-bible-texts` · **Base**: `main` (merged) ·
**Head**: `599677d`

All findings from @katherinejensen00's four review rounds are addressed. Rounds 1–3 are summarized
below; **rounds 4 and 5 are in their own section**. Remaining follow-up work is collected in
**PT-4416**.

> **This PR is larger than PT-4132.** `afa6e5aa3ac` squash-merged **#2710 (PT-4403)** into this
> branch, so roughly a third of the diff is the blank-chapter feature rather than the missing-book
> one — including a change to Power-mode editing behaviour. See
> [Also carries PT-4403](#also-carries-pt-4403-2710) before reviewing from this description.

## Overview

Both the Model text and Bible texts/Commentaries panels swallowed the project data provider's
missing-book error and fell through to rendering the editor, so a book absent from a resource showed
a blank editor with no explanation. This change centralizes missing-book detection, adds a
`ResourceBookNotAvailable` message component, and has each panel swap only its content area so the
header — the resource selector, which is the user's actual remedy — stays mounted. The design
[...]

PR #2706, opened 2026-08-21
PT-4403: Show honest book-not-found message in resource panels

jolierabideau:
# Code Review Summary

**Branch**: `pt-4403-honest-message-when-book-does-not-exist`

**Base**: `origin/main`

**Date**: 2026-08-21

**Files changed**: 8

## Overview

A book missing from the selected resource comes back from the scripture PDP as an *error*, not as
empty USJ. Every panel that displayed resource text collapsed that error to "no data" and then
rendered the read-only editor anyway — so the user saw either the previous chapter's text or an
"Enter some Scripture…" placeholder. Both are dishonest: the first is stale content presented as
current, the second is an edit prompt in a resource the user cannot edit.

This change detects the book-not-found error and renders the existing localized message
(`%webView_platformScriptureEditor_error_bookNotFoundResource%`) instead. The detection — and the
regex coupled to `MissingBookException.cs` — is extracted into `isBookNotFoundError` in
`platform-scripture-editor.utils.ts` so all three Scripture surfaces share one unit-testable source
of truth. The resource selector stays mounted so the user can switch to a resource that has the
book, and the message sits outside the `dir` wrapper because it is UI-locale text, not resource text.

During review the fix was extended from the Bible text tab to the **Model text panel**, which had
the identical bug via a different code path, and the message was converged onto the shared
`EmptyState` component in the resource panels so the same sentence is worded, styled, and announced
to screen readers identically everywhere.

### Rebased onto PT-4111 (#2691)

#2691 landed first and rewrote the editable editor's `!bookExists` branch into a mode-aware
`BookNotAvailableView` (Power zero-state with a **Manage books** action, Simple "ask your project
administrator" copy) behind an `isResource` fork. This PR was rebased on top of it and its own
`EmptyState` wrapper for that branch was **dropped as superseded** — #2691's version is strictly
richer and is what ships.

What survives the rebase is everything #2691 does not cover:

- the three resource surfaces (Bible text tab, Commentaries tab — same component — and Model text
  panel), which #2691 never touches;
- `isBookNotFoundError` in `platform-scripture-editor.utils.ts`, replacing the editor's local
  `bookNotFoundRegex`, so all surfaces share one unit-testable predicate;
- the log-level branch (`debug` for a book the text simply lacks, `error` retained for genuine PDP
  faults);
- `MissingBookExceptionTests`, pinning the C#↔TS contract.

Two small edits close the consistency gap #2691 left, which NN 5A ("consistent across all tabs")
calls for directly:

- **`platform-scripture-editor.web-view.tsx`** — the `isResource` branch rendered bare text in a
  `<div>` with no live region, the only missing-book surface that announced nothing. Now
  `EmptyState`, matching the resource panels.
- **`book-not-available-view.component.tsx`** — Simple mode's `<span>` now carries
  `tw:text-sm tw:text-muted-foreground`, matching `EmptyDescription` in its own Power branch and
  `EmptyState` everywhere else. It stays a `span` rather than becoming an `EmptyState`: the
  `role="status"` live region and the focus repair live on the wrapper `div`, and `EmptyState`
  cannot accept either, so converting it would have been an accessibility regression. Typography
  was the only thing actually diverging.

## API Changes

None. No public API surfaces changed.

- `lib/platform-bible-react/`, `lib/platform-bible-utils/`: unchanged
- `lib/papi-dts/papi.d.ts`: unchanged (no regeneration needed; no hand edits)
- `extensions/src/*/src/types/*.d.ts`: unchanged
- The new export `isBookNotFoundError` is extension-internal (not a declaration file, not
  re-exported through `@papi/`), consumed only within `platform-scripture-editor`.

None.

- [x] **The missing-book message rendered as bare text in a `<div>`, skipping the `EmptyState`
      component** (`resource-text-panel.web-view.tsx`). The region swaps content in place as the
      user navigates, so with no live region a screen reader user got silence where scripture used
      to be. Flagged independently by both the UX and style passes; a sibling web view in the same
      extension (`scripture-text-grid.web-view.tsx:545`) already used `EmptyState`.
      _(fixed during review: replaced with `<EmptyState id="resource-text-panel-book-not-found"
      className="tw:text-center" />`, gaining `role="status"`, the standard
      `tw:text-sm tw:text-muted-foreground` treatment, and a `data-testid`. Verified live in the
      running app: the rendered node is
      `<p role="status" data-testid="resource-text-panel-book-not-found">`.)_

[Author response: Author chose to adopt `EmptyState` immediately. The same treatment was then
applied to the editable editor and the Model text panel so all three panels match.]

### Minor — Consider

- [x] **Log level disagreed between the two tabs** — `logger.warn` in the resource panel vs
      `logger.error` in the editable editor for the same class of PDP failure, undercutting the
      "one source of truth" goal. _(fixed during review: both now branch identically. Investigated
      what was actually pre-existing rather than guessing — the editable editor's `logger.error` was
      the only prior pattern; the resource panel previously logged **nothing** at all, so the `warn`
      was new in this PR, not an existing convention.)_
- [x] **Book-not-found was logged as a fault** even though it is ordinary navigation, so every
      navigation to a missing book emitted a warning/error. _(fixed during review: branches on
      `isBookNotFoundError` first — `logger.debug` for the expected missing book, `logger.error`
      retained for genuine PDP failures, so the pre-existing severity for real faults is not
      downgraded. Verified in `~/Library/Logs/Electron/main.log`: the new messages appear at
      `[debug]` and there are **zero** error-level book-not-found entries.)_
- [x] **`isPlatformError` evaluated twice on the same value** in `resource-text-panel.web-view.tsx`
      (once for `usjFromPdp`, again inside the `bookExists` memo), where the editable editor derives
      both in one pass. _(fixed during review: collapsed to a single
      `const [usjFromPdp, bookExists] = useMemo<[Usj | undefined, boolean]>(...)`, matching the
      editable editor's shape. The resource panel's `undefined`-on-error semantics were preserved —
      unlike the editor's `defaultUsj` — because the `if (usjFromPdp)` guard downstream depends on
      it.)_
- [x] **The two tabs styled the identical message differently** — `tw:flex-1 … tw:p-8 tw:text-center`
      vs `tw:h-full tw:px-4` with no centering — so the same sentence sat in a different place
      depending on which tab you were in. _(fixed during review, then **superseded by #2691**, which
      replaced that whole branch. The in-review `EmptyState` wrapper was dropped in the rebase; the
      cross-surface consistency it was after is now delivered by the two edits described under
      "Rebased onto PT-4111" above.)_
- [x] **The C#/TypeScript contract was advisory only** — `BOOK_NOT_FOUND_REGEX` is coupled to
      `MissingBookException.cs`'s message text, comments pointed both ways, but nothing failed if
      someone reworded the C#. _(fixed during review: added
      `c-sharp-tests/MissingBookExceptionTests.cs` (4 tests) pinning the exact wording and the
      regex the extension applies, and cross-referenced the comments on both sides.
      **Falsifiability checked**: temporarily rewording the C# message to
      `"Book {n} is missing from {id}."` failed 3 of the 4 tests; original restored.)_
- [x] **No panel-level test for the render fork** — the predicate was unit-tested, but the
      "render the message instead of the editor" decision was not. _(fixed during review for the
      Model text panel: 4 tests added to the existing `model-text-panel.component.test.tsx`
      covering the message replacing the editor, the `role="status"` live region, the negative
      control (an unrelated failure still renders the editor), and recovery when navigating back to
      a book the resource has. **Falsifiability checked**: disabling the render branch failed 3 of
      the 4, with the negative control correctly still passing. `resource-text-panel.web-view.tsx`
      still has no test file of its own — see Suggested Review Focus.)_
- [ ] ~~`logger` call inside `useMemo` is a render-time side effect (double-fires under StrictMode)~~
      _(Dismissed: consistent with the pre-existing editable editor, both files now do it
      identically, and the common case is now `debug` so the noise is minimal. Moving it to a
      `useEffect` would split the log from the derivation it explains.)_
- [ ] ~~A shared `{ usj, bookExists }` helper would make "one source of truth" cover the whole
      guard → log → branch dance, not just the regex~~ _(Dismissed: the three call sites differ in
      fallback (`defaultUsj` vs `undefined` vs React state) and in log wording, so the helper would
      need enough parameters that it would not clearly pay for itself.)_
- [ ] ~~The message says what is missing but not what to do next~~ _(Dismissed: the change
      deliberately reuses the existing string rather than coining a new one, and rewording would
      affect all three panels. Per the style guide a *meaning* change would need a new key.)_
- [ ] ~~The borrowed localization key is prepended above the alphabetized `%webView_resourcePanel_*`
      block~~ _(Dismissed: the comment above it already marks it as borrowed from another web view's
      namespace, which is the information a reader needs.)_
- [ ] ~~`Editorial` is unmounted while the book is missing, so returning to a book that exists
      remounts a fresh editor whose content depends on the `setUsj` effect refiring~~
      _(Dismissed as a known, pre-existing coupling: it mirrors the editable editor's existing
      `!bookExists` early return, and holds as long as `useProjectData` returns a new object
      identity per chapter. The Model text panel's equivalent recovery path **is** now covered by a
      test.)_

[Author response: Author asked for the log-level disagreement, the double evaluation, the
cross-tab styling divergence, and the unenforceable C# contract to be fixed, and asked that the
pre-existing pattern be checked rather than assumed. The remaining five were dismissed with
reasoning recorded above.]

#### Shared Regions Modified

None. The only `#region shared with` marker in this extension is in `src/webpack-env.d.ts`, which
is not in the diff. (The new `#region PDP Error Detection` is a plain region marker, not a shared
one.)

#### Extension Config Changes

None. No `package.json`, `tsconfig.json`, `webpack.config.ts`, `.eslintrc*`, CI, or dependency
changes.

## Positive Observations

- The regex and its `MissingBookException.cs` coupling comment moved together into the utils
  module, so the fragile string stayed documented at its new home — and it was renamed to
  `BOOK_NOT_FOUND_REGEX`, matching the neighboring `PERMISSIONS_EXCEPTION_REGEX` convention that
  the old `bookNotFoundRegex` did not follow.
- `isBookNotFoundError` was placed in the module that already hosts this extension's shared pure
  helpers, respecting that file's documented extension-host import boundary — a real trap given the
  header comment there. `extension-host-import-boundary.test.ts` passes.
- The TSDoc explicitly covers the `undefined`/in-flight and plain-USJ cases, and the tests cover
  exactly those branches including the "unrelated `PlatformError`" negative.
- No new localized strings: the existing key is reused, with both `en` and `es` entries already
  present, and an inline comment explains why a `platformScriptureEditor`-prefixed key appears in
  other panels.
- Placing the message outside the `dir={options.textDirection}` wrapper — with a comment saying why
  — is exactly right for a mixed-direction panel: resource text keeps its own direction while the
  UI sentence follows the interface locale.
- Keeping the resource selector mounted in the missing-book case leaves the user a way out instead
  of stranding them.
- The refactor of the editable editor is behavior-preserving: the deleted local regex and the
  retained log produce the same `bookExists` result as before.
- Comments throughout explain *why* rather than restating the code.

## Interview Notes

**Stated purpose**: make the Bible text tab honest when the selected resource lacks the current
book, instead of showing stale content or an edit prompt in a read-only resource.

**Key decisions the author made during review:**

- On the `EmptyState` finding, the author chose to adopt the shared component rather than defend the
  bare `<div>`.
- On the log-level disagreement, the author explicitly asked *"What was pre-existing? We want to
  follow code patterns"* rather than accepting a proposed level. That check changed the outcome: the
  investigation showed `logger.error` was the only pre-existing pattern and that the `logger.warn`
  was introduced by this PR, so `error` was retained for genuine faults instead of being downgraded
  to `warn` across both files.
- The author endorsed branching on `isBookNotFoundError` so the expected case logs quietly.
- The author approved collapsing the resource panel to the editable editor's single-memo shape
  specifically on consistency-with-existing-patterns grounds.
- When shown screenshot evidence that the **Model text panel** had the same unfixed bug, the author
  chose to fix it in this PR rather than defer it, accepting the scope increase.

**Author understanding**: the author demonstrated clear understanding of the change throughout and
drove several decisions rather than accepting suggestions. The "what was pre-existing?" question in
particular caught an assumption that would otherwise have silently changed existing behavior. No
areas were deferred to AI and no uncertainty was expressed.

**Unresolved items**: none.

## Verification Performed

Beyond the standard checks, three behavioral claims in this PR were verified against the running
app rather than inferred from the diff:

1. **The bug reproduces and the fix resolves it** — captured before/after in Platform.Bible using
   SANIAS (a New Testament–only Sanskrit resource) navigated Matthew 1:1 → Genesis 1:1, the same
   navigation on both `main` and this branch. Before: the panel showed Matthew content while the
   toolbar read Genesis 1:1. After: "This book does not exist in this resource." Screenshots are in
   `.review/screenshots/`.
2. **The accessibility fix is real** — the live DOM in both panels is
   `<p role="status" data-testid="…-book-not-found" class="tw:text-sm tw:text-muted-foreground
   tw:text-center">`, and the editor is confirmed unmounted.
3. **The log-level change is real** — `~/Library/Logs/Electron/main.log` shows the new messages at
   `[debug]` with **zero** error-level book-not-found entries.

Both new test groups were **mutation-checked** (deliberately break the code, confirm the tests fail,
restore) rather than merely observed passing.

## In-Review Quality Check

All checks pass after the in-review changes:

| Check | Result |
| --- | --- |
| `npm run typecheck` | Clean |
| `npm run lint` | **0 errors**; 5 warnings, all pre-existing. Two (`import/no-duplicates`) are in `model-text-panel.component.tsx` — a file this branch edits — but they predate it: their line numbers merely shifted from 28/34 to 34/40 because the new imports pushed them down. They concern a duplicate `resource-reference.utils` import unrelated to this change, so they were left alone rather than folded into this PR. |
| Prettier (changed files only) | Already formatted — no changes |
| `csharpier` | Clean, no stray reformats |
| `npm test` | **840 passed**, 58 files |
| `dotnet test c-sharp-tests/` | **1584 passed**, 0 failed, 6 skipped |

No fixes were needed to satisfy the quality gate; no unfixable failures.

## Suggested Review Focus

- [ ] **Scope increase, deliberately taken**: the PR also fixes `model-text-panel.component.tsx`.
      Confirm you are happy with that breadth in one PR versus splitting the Model text panel into a
      follow-up.
- [ ] **The two consistency edits reach into #2691's just-merged code** (`book-not-available-view`'s
      Simple `<span>` and the editor's `isResource` branch). They are deliberate and small, but they
      touch code reviewed days ago under a different ticket — worth confirming you want them here
      rather than as a separate follow-up.
- [ ] **The broadened `isBookNotFoundError` contract**: it now accepts both a `PlatformError` (what
      `useProjectData` returns) *and* a thrown `Error` (what awaiting `getChapterUSJ` rejects with),
      because the Model text panel fails via the second path. Worth a look at whether widening the
      guard to `possibleError instanceof Error` is the right boundary, versus two separate
      predicates.
- [ ] **`resource-text-panel.web-view.tsx` still has no test file**, so its render fork is covered
      only by the manual verification above. The Model text panel's equivalent fork is now
      unit-tested; consider whether the resource panel warrants the same.
- [ ] **`logger` is not in the Model text panel's load-effect dependency array.** `exhaustive-deps`
      is already suppressed on that effect for an unrelated documented reason, and the PAPI logger is
      a stable module-level object, so this is believed safe — but it is a new use of a prop inside
      that effect and worth a second pair of eyes.
- [ ] **The C#↔TS coupling is now enforced by a test, but by duplication**: the regex literal exists
      in both `platform-scripture-editor.utils.ts` and `MissingBookExceptionTests.cs`. A shared
      constant is not possible across the language boundary; confirm the duplicate-with-a-failing-test
      approach is the tradeoff you want.
- [ ] **Surface sweep**: the Bible text tab, Commentaries tab, Model text panel, and both branches
      of the editable editor are now consistent. One known hole is `scripture-text-grid.web-view.tsx`,
      which has no book-not-found handling at all (its only `EmptyState` is "no resources selected").
      It is Power-only and setting-gated, so it may be out of scope — worth a deliberate decision
      rather than an omission.

## Screenshots

> Reproduced with **SANIAS** (Sanskrit Bible, New Testament only) navigated Matthew 1:1 → Genesis 1:1 — the same navigation on `main` and on this branch.

| | Before (`main`) | After (this branch) |
| --- | --- | --- |
| **Bible text panel** | Shows Matthew content while the toolbar reads Genesis 1:1 | "This book does not exist in this resource." |
| **Model text panel** | Shows Matthew content (same bug, different code path) | "This book does not exist in this resource." |

_Screenshot files are in `.review/screenshots/` (gitignored) — attach `before-full-window.png` and `after-full-window.png` here._

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2706)

### DIFF
--- a/c-sharp-tests/MissingBookExceptionTests.cs
+++ b/c-sharp-tests/MissingBookExceptionTests.cs
@@ -1,0 +1,77 @@
+using System.Diagnostics.CodeAnalysis;
+using System.Text.RegularExpressions;
+using Paranext.DataProvider;
+
+namespace TestParanextDataProvider
+{
+    /// <summary>
+    /// Pins the exact wording of <see cref="MissingBookException"/>'s message.
+    /// </summary>
+    /// <remarks>
+    /// The Scripture editor extension detects "this book is not in this project/resource" by
+    /// pattern-matching the message text of the error the PDP returns — there is no structured
+    /// error code to key off. That check lives in <c>isBookNotFoundError</c> in
+    /// <c>extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts</c>,
+    /// which applies the regex duplicated below.
+    ///
+    /// Without this test the coupling is advisory only: rewording the message here still compiles
+    /// and still passes every other test, but silently turns both Scripture tabs' honest "this book
+    /// does not exist" message back into an empty editor. If this test fails because the message
+    /// changed deliberately, update the regex in the TypeScript helper in the same commit.
+    /// </remarks>
+    [ExcludeFromCodeCoverage]
+    public class MissingBookExceptionTests
+    {
+        /// <summary>
+        /// Must stay character-for-character identical to <c>BOOK_NOT_FOUND_REGEX</c> in
+        /// <c>platform-scripture-editor.utils.ts</c>.
+        /// </summary>
+        private const string ExtensionBookNotFoundRegex = @"Book number \d+ not found in project";
+
+        [Test]
+        public void Message_MatchesRegexTheScriptureEditorExtensionUses()
+        {
+            var exception = new MissingBookException(40, "ProjectName");
+
+            Assert.That(exception.Message, Does.Match(ExtensionBookNotFoundRegex));
+        }
+
+        [Test]
+        public void Message_HasTheExactExpectedWording()
+        {
+            var exception = new MissingBookException(40, "ProjectName");
+
+            Assert.That(
+                exception.Message,
+                Is.EqualTo("Book number 40 not found in project ProjectName.")
+            );
+        }
+
+        [Test]
+        public void Message_MatchesTheExtensionRegexForAnyBookNumber()
+        {
+            foreach (var bookNum in new[] { 1, 40, 66 })
+            {
+                var exception = new MissingBookException(bookNum, "ProjectName");
+
+                Assert.That(
+                    Regex.IsMatch(exception.Message, ExtensionBookNotFoundRegex),
+                    Is.True,
+                    $"Book number {bookNum} should be detected as book-not-found by the extension"
+                );
+            }
+        }
+
+        [Test]
+        public void BookNumAndProjectId_ArePreserved()
+        {
+            var exception = new MissingBookException(40, "ProjectName");
+
+            Assert.Multiple(() =>
+            {
+                Assert.That(exception.BookNum, Is.EqualTo(40));
+                Assert.That(exception.ProjectId, Is.EqualTo("ProjectName"));
+            });
+        }
+    }
+}
--- a/c-sharp/MissingBookException.cs
+++ b/c-sharp/MissingBookException.cs
@@ -4,7 +4,9 @@ namespace Paranext.DataProvider;
 /// Exception thrown when a requested book is not found in a Paratext project.
 /// </summary>
 public class MissingBookException(int bookNum, string projectId)
-    // IMPORTANT: The scripture editor WebView depends on the exact text of this message.
+    // IMPORTANT: The scripture editor WebView depends on the exact text of this message; it is
+    // pattern-matched by `isBookNotFoundError` in platform-scripture-editor.utils.ts and pinned by
+    // MissingBookExceptionTests. Reword it only alongside that regex.
     : Exception($"Book number {bookNum} not found in project {projectId}.")
 {
     /// <summary>
--- a/extensions/src/platform-scripture-editor/src/book-not-available-view.component.tsx
+++ b/extensions/src/platform-scripture-editor/src/book-not-available-view.component.tsx
@@ -96,7 +96,14 @@ export function BookNotAvailableView({
         tabIndex={-1}
         className="tw:flex tw:h-full tw:items-center tw:justify-center tw:px-4 tw:outline-none"
       >
-        <span>{localize(localizedStrings, SIMPLE_MESSAGE_KEY)}</span>
+        {/* Same small muted treatment as `EmptyDescription` in the Power branch below and as
+            `EmptyState` in the resource panels, so a missing book reads identically whichever
+            surface the user is on. The message stays a plain `span` rather than `EmptyState`
+            because the `role="status"` live region and the focus repair both live on the wrapper
+            above, which `EmptyState` has no way to accept. */}
+        <span className="tw:text-sm tw:text-muted-foreground tw:text-center">
+          {localize(localizedStrings, SIMPLE_MESSAGE_KEY)}
+        </span>
       </div>
     );
   }
--- a/extensions/src/platform-scripture-editor/src/model-text-panel.component.test.tsx
+++ b/extensions/src/platform-scripture-editor/src/model-text-panel.component.test.tsx
@@ -36,6 +36,8 @@ const STRINGS = {
     "The model text couldn't be installed. Check your connection and try again.",
   '%webView_modelTextPanel_retry%': 'Try again',
   '%webView_modelTextPanel_emptyState_prompt%': 'No model text selected.',
+  '%webView_platformScriptureEditor_error_bookNotFoundResource%':
+    'This book does not exist in this resource.',
 };
 
 const INSTALLED_RESOURCE: DblResourceData = {
@@ -276,5 +278,92 @@ describe('ModelTextPanel', () => {
         "The model text couldn't be installed. Check your connection and try again.",
       ),
     ).toBeInTheDocument();
+  });
+
+  it('shows the missing-book message instead of the editor when the resource lacks the book', async () => {
+    // A book the resource doesn't have makes getChapterUSJ reject. Without the bookExists branch the
+    // panel just clears `usj` and still renders the read-only editor, which then shows the previous
+    // chapter's text or an "enter some Scripture" prompt — an edit prompt in a read-only resource.
+    const getResourceChapter = vi.fn(async () => {
+      throw new Error(
+        'JSON-RPC Request error (-32000): Book number 1 not found in project project-web.',
+      );
+    });
+    renderPanel({
+      effectiveModelTexts: configuredModelText('uid-web'),
+      dblResources: [INSTALLED_RESOURCE],
+      getResourceChapter,
+    });
+
+    expect(
+      await screen.findByText('This book does not exist in this resource.'),
+    ).toBeInTheDocument();
+    expect(screen.queryByTestId('editorial')).not.toBeInTheDocument();
+  });
+
+  it('announces the missing-book message in a live region', async () => {
+    // The panel swaps the editor out for this message as the user navigates, so without a status
+    // role a screen reader user gets silence where the model text used to be.
+    const getResourceChapter = vi.fn(async () => {
+      throw new Error('Book number 1 not found in project project-web.');
+    });
+    renderPanel({
+      effectiveModelTexts: configuredModelText('uid-web'),
+      dblResources: [INSTALLED_RESOURCE],
+      getResourceChapter,
+    });
+
+    expect(await screen.findByRole('status')).toHaveTextContent(
+      'This book does not exist in this resource.',
+    );
+  });
+
+  it('keeps rendering the editor when the chapter load fails for an unrelated reason', async () => {
+    // Only a missing book earns the "this book does not exist" claim. Any other failure must fall
+    // back to the editor rather than telling the user something untrue about the resource.
+    const getResourceChapter = vi.fn(async () => {
+      throw new Error('Network request failed');
+    });
+    renderPanel({
+      effectiveModelTexts: configuredModelText('uid-web'),
+      dblResources: [INSTALLED_RESOURCE],
+      getResourceChapter,
+    });
+
+    expect(await screen.findByTestId('editorial')).toBeInTheDocument();
+    expect(
+      screen.queryByText('This book does not exist in this resource.'),
+    ).not.toBeInTheDocument();
+  });
+
+  it('recovers to the editor when navigating from a missing book to one the resource has', async () => {
+    // bookExists must be re-set on a successful load, or the panel stays stuck on the message for
+    // the rest of the session once the user hits one missing book.
+    const getResourceChapter = vi.fn(async () => ({ usj: SAMPLE_USJ, textDirection: 'ltr' }));
+    getResourceChapter.mockRejectedValueOnce(
+      new Error('Book number 1 not found in project project-web.'),
+    );
+    const { rerender } = renderPanel({
+      effectiveModelTexts: configuredModelText('uid-web'),
+      dblResources: [INSTALLED_RESOURCE],
+      getResourceChapter,
+      scrRef: { book: 'GEN', chapterNum: 1, verseNum: 1 },
+    });
+    expect(
+      await screen.findByText('This book does not exist in this resource.'),
+    ).toBeInTheDocument();
+
+    // Navigate to a book the resource does contain; the next load resolves.
+    rerender(
+      <ModelTextPanel
+        {...makeProps({
+          effectiveModelTexts: configuredModelText('uid-web'),
+          dblResources: [INSTALLED_RESOURCE],
+          getResourceChapter,
+          scrRef: { book: 'MAT', chapterNum: 1, verseNum: 1 },
+        })}
+      />,
+    );
+    expect(await screen.findByTestId('editorial')).toBeInTheDocument();
   });
 });
--- a/extensions/src/platform-scripture-editor/src/model-text-panel.component.tsx
+++ b/extensions/src/platform-scripture-editor/src/model-text-panel.component.tsx
@@ -8,6 +8,7 @@ import { Usj } from '@eten-tech-foundation/scripture-utilities';
 import { SerializedVerseRef } from '@sillsdev/scripture';
 import {
   Button,
+  EmptyState,
   Spinner,
   Tooltip,
   TooltipContent,
@@ -16,7 +17,11 @@ import {
   useExtraValidMarkers,
   useTruncationTooltip,
 } from 'platform-bible-react';
-import { type DblResourceData, type LocalizedStringValue } from 'platform-bible-utils';
+import {
+  getErrorMessage,
+  type DblResourceData,
+  type LocalizedStringValue,
+} from 'platform-bible-utils';
 import type {
   DblResourceReference,
   EffectiveResourceReference,
@@ -24,6 +29,7 @@ import type {
   ResourceReferenceList,
 } from 'platform-scripture';
 import { ComponentProps, useCallback, useEffect, useMemo, useRef, useState } from 'react';
+import { isBookNotFoundError } from './platform-scripture-editor.utils';
 import { selectTextConnection } from './select-dbl-resource';
 import { isDblResourceReference } from './resource-reference.utils';
 import { findCachedDblResource } from './scripture-text-grid/dbl-resource-lookup.utils';
@@ -61,6 +67,9 @@ export const MODEL_TEXT_PANEL_STRING_KEYS = Object.freeze([
   '%webView_modelTextPanel_installFailedOffline%',
   '%webView_modelTextPanel_retry%',
   '%webView_modelTextPanel_emptyState_prompt%',
+  // Shared with both Scripture tabs' missing-book branch so every panel words a book the resource
+  // lacks the same way
+  '%webView_platformScriptureEditor_error_bookNotFoundResource%',
 ] as const);
 
 type ModelTextPanelLocalizedStringKey = (typeof MODEL_TEXT_PANEL_STRING_KEYS)[number];
@@ -184,6 +193,11 @@ export function ModelTextPanel({
   const [textDirection, setTextDirection] = useState<string>(DEFAULT_TEXT_DIRECTION);
   // `undefined` means "not yet fetched" so we can show the loading state, matching the original.
   const [isUsjLoading, setIsUsjLoading] = useState(false);
+  // A book missing from the resource makes `getChapterUSJ` reject rather than return empty USJ.
+  // Without tracking it, the load below just clears `usj` and this panel falls through to the
+  // read-only editor, which then shows either the previous chapter's text or its "enter some
+  // Scripture" placeholder — an edit prompt in a resource the user cannot edit.
+  const [bookExists, setBookExists] = useState(true);
 
   useEffect(() => {
     if (!resourceProjectId) {
@@ -200,11 +214,18 @@ export function ModelTextPanel({
       if (!isActive) return;
       setUsj(nextUsj);
       setTextDirection(nextTextDirection || DEFAULT_TEXT_DIRECTION);
+      setBookExists(true);
       setIsUsjLoading(false);
     };
-    load().catch(() => {
+    load().catch((e) => {
       if (!isActive) return;
       setUsj(undefined);
+      // A book the resource simply doesn't have is ordinary navigation, not a fault, so it is
+      // logged quietly; every other failure is a real error. Matches both Scripture tabs.
+      const bookNotFound = isBookNotFoundError(e);
+      if (bookNotFound) logger?.debug(`Book not found in model text: ${getErrorMessage(e)}`);
+      else logger?.error(`Error getting USJ for the model text panel: ${getErrorMessage(e)}`);
+      setBookExists(!bookNotFound);
       setIsUsjLoading(false);
     });
     return () => {
@@ -446,6 +467,24 @@ export function ModelTextPanel({
     return (
       <div className="tw:flex tw:h-screen tw:items-center tw:justify-center tw:p-8 tw:text-center">
         <Spinner />
+      </div>
+    );
+  }
+
+  // Missing book: the resolved resource has no such book. Say so rather than handing the reader an
+  // empty read-only editor. Uses the same borrowed string and `EmptyState` treatment as both
+  // Scripture tabs, so the identical sentence is worded, styled, and announced to screen readers
+  // the same way in every panel that can hit this state.
+  if (!bookExists) {
+    return (
+      <div className="tw:flex tw:h-screen tw:items-center tw:justify-center tw:p-8">
+        <EmptyState
+          id="model-text-panel-book-not-found"
+          className="tw:text-center"
+          message={
+            localizedStrings['%webView_platformScriptureEditor_error_bookNotFoundResource%'] ?? ''
+          }
+        />
       </div>
     );
   }
--- a/extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.test.ts
+++ b/extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.test.ts
@@ -1,7 +1,7 @@
 import { describe, it, expect, vi } from 'vitest';
 import { ScriptureRange } from 'platform-scripture-editor';
 import type PapiBackend from '@papi/backend';
-import { UsjTextContentLocation } from 'platform-bible-utils';
+import { newPlatformError, UsjTextContentLocation } from 'platform-bible-utils';
 import type { SavedWebViewDefinition } from '@papi/core';
 import { MutableRefObject } from 'react';
 import { EditorRef } from '@eten-tech-foundation/platform-editor';
@@ -18,6 +18,7 @@ import {
   selectProjectIdsForOpenMode,
   startDefaultProjectPicker,
   toScriptureEditorInfos,
+  isBookNotFoundError,
   isChapterBlank,
   buildChapterScaffoldOps,
   canAddChapterNumber,
@@ -2589,5 +2590,27 @@ describe('resolveAddChapterNumberClick', () => {
 
   it('returns "insert" when not in flight and lastVerse is positive', () => {
     expect(resolveAddChapterNumberClick(false, 3)).toBe('insert');
+  });
+});
+
+describe('isBookNotFoundError', () => {
+  it('returns true for a PlatformError whose message says the book number was not found', () => {
+    expect(
+      isBookNotFoundError(newPlatformError('Book number 40 not found in project ProjectName')),
+    ).toBe(true);
+  });
+
+  it('returns false for a PlatformError with an unrelated message', () => {
+    expect(isBookNotFoundError(newPlatformError('Permissions exception for projectId abc'))).toBe(
+      false,
+    );
+  });
+
+  it('returns false for a successfully retrieved USJ document', () => {
+    expect(isBookNotFoundError({ type: USJ_TYPE, version: USJ_VERSION, content: [] })).toBe(false);
+  });
+
+  it('returns false for undefined, as when the data has not arrived yet', () => {
+    expect(isBookNotFoundError(undefined)).toBe(false);
   });
 });
--- a/extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.test.ts
+++ b/extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.test.ts
@@ -2613,4 +2613,18 @@ describe('isBookNotFoundError', () => {
   it('returns false for undefined, as when the data has not arrived yet', () => {
     expect(isBookNotFoundError(undefined)).toBe(false);
   });
+
+  it('returns true for a thrown Error, as when awaiting getChapterUSJ rejects', () => {
+    expect(
+      isBookNotFoundError(
+        new Error(
+          'JSON-RPC Request error (-32000): Book number 1 not found in project ProjectName.',
+        ),
+      ),
+    ).toBe(true);
+  });
+
+  it('returns false for a thrown Error with an unrelated message', () => {
+    expect(isBookNotFoundError(new Error('Network request failed'))).toBe(false);
+  });
 });
--- a/extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts
+++ b/extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts
@@ -1158,3 +1158,26 @@ export function resolveAddChapterNumberClick(
 }
 
 // #endregion Chapter Scaffold Helpers
+
+// #region PDP Error Detection
+
+// This regex is connected directly to the exception message within MissingBookException.cs
+const BOOK_NOT_FOUND_REGEX = /Book number \d+ not found in project/;
+
+/**
+ * Whether a value received from a scripture PDP is the error it returns when the requested book
+ * does not exist in the project or resource. Views use this to show an honest "this book does not
+ * exist" message instead of an empty editor.
+ *
+ * Anything that is not a `PlatformError` — including the USJ itself and `undefined` while the data
+ * is still in flight — is not a book-not-found error.
+ *
+ * @param possibleError Value returned by a scripture PDP (e.g. `ChapterUSJ`), which may be a
+ *   `PlatformError`.
+ */
+export function isBookNotFoundError(possibleError: unknown): boolean {
+  if (!isPlatformError(possibleError)) return false;
+  return BOOK_NOT_FOUND_REGEX.test(getErrorMessage(possibleError));
+}
+
+// #endregion PDP Error Detection
--- a/extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts
+++ b/extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts
@@ -1161,22 +1161,26 @@ export function resolveAddChapterNumberClick(
 
 // #region PDP Error Detection
 
-// This regex is connected directly to the exception message within MissingBookException.cs
+// This regex is connected directly to the exception message within MissingBookException.cs.
+// `c-sharp-tests/MissingBookExceptionTests.cs` duplicates this pattern and fails if that message is
+// reworded, so the two sides cannot drift silently — update both together.
 const BOOK_NOT_FOUND_REGEX = /Book number \d+ not found in project/;
 
 /**
- * Whether a value received from a scripture PDP is the error it returns when the requested book
+ * Whether a value received from a scripture PDP is the error it reports when the requested book
  * does not exist in the project or resource. Views use this to show an honest "this book does not
  * exist" message instead of an empty editor.
  *
- * Anything that is not a `PlatformError` — including the USJ itself and `undefined` while the data
- * is still in flight — is not a book-not-found error.
+ * Handles both shapes the scripture PDPs surface, because the two call styles fail differently:
+ * `useProjectData` hands back a `PlatformError` in place of the data, while awaiting
+ * `getChapterUSJ` directly rejects with a thrown `Error`. Anything that is neither — including the
+ * USJ itself and `undefined` while the data is still in flight — is not a book-not-found error.
  *
- * @param possibleError Value returned by a scripture PDP (e.g. `ChapterUSJ`), which may be a
- *   `PlatformError`.
+ * @param possibleError Value returned by a scripture PDP (e.g. `ChapterUSJ`) or thrown by one of
+ *   its methods, which may be a `PlatformError` or an `Error`.
  */
 export function isBookNotFoundError(possibleError: unknown): boolean {
-  if (!isPlatformError(possibleError)) return false;
+  if (!isPlatformError(possibleError) && !(possibleError instanceof Error)) return false;
   return BOOK_NOT_FOUND_REGEX.test(getErrorMessage(possibleError));
 }
 
--- a/extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx
+++ b/extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx
@@ -42,6 +42,7 @@ import {
   CommentEditor,
   DisabledActionTooltip,
   EditorKeyboardShortcuts,
+  EmptyState,
   FOOTNOTE_EDITOR_STRING_KEYS,
   FootnoteEditor,
   MarkdownRenderer,
@@ -1290,8 +1291,13 @@ globalThis.webViewComponent = function PlatformScriptureEditor({
   const [usjFromPdp, bookExists] = useMemo(() => {
     if (!isPlatformError(usjFromPdpPossiblyError)) return [usjFromPdpPossiblyError, true];
 
-    logger.error(`Error getting USJ from PDP: ${getErrorMessage(usjFromPdpPossiblyError)}`);
-    return [defaultUsj, !isBookNotFoundError(usjFromPdpPossiblyError)];
+    // A book missing from the project is ordinary navigation, not a fault, so it is logged
+    // quietly; every other PDP failure is a real error. Both Scripture tabs branch the same way.
+    const bookNotFound = isBookNotFoundError(usjFromPdpPossiblyError);
+    const errorMessage = getErrorMessage(usjFromPdpPossiblyError);
+    if (bookNotFound) logger.debug(`Book not found in project: ${errorMessage}`);
+    else logger.error(`Error getting USJ from PDP: ${errorMessage}`);
+    return [defaultUsj, !bookNotFound];
   }, [usjFromPdpPossiblyError]);
   const usjSentToPdp = useRef<Usj | undefined>(usjFromPdp);
   const currentlyWritingUsjToPdp = useRef(false);
@@ -1969,7 +1975,18 @@ globalThis.webViewComponent = function PlatformScriptureEditor({
         return (
           <div className="tw:flex tw:items-center tw:justify-center tw:h-full tw:px-4">
             {workaround}
-            {localizedStrings['%webView_platformScriptureEditor_error_bookNotFoundResource%']}
+            {/* `EmptyState` rather than bare text: this region swaps content in place as the user
+                navigates, so its `role="status"` is what announces the missing book to a screen
+                reader instead of scripture going silent. It is also the treatment the Bible text,
+                commentary, and model text panels use for the same sentence, so the message reads
+                and announces identically wherever a resource lacks the book. */}
+            <EmptyState
+              id="platform-scripture-editor-book-not-found-resource"
+              className="tw:text-center"
+              message={
+                localizedStrings['%webView_platformScriptureEditor_error_bookNotFoundResource%']
+              }
+            />
           </div>
         );
       }
--- a/extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx
+++ b/extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx
@@ -126,6 +126,7 @@ import {
   formatEditorTitle,
   generateInlineMarkerMenuListItems,
   generateParagraphMenuListItems,
+  isBookNotFoundError,
   isChapterBlank,
   openCommentListAndSelectThreadSafe,
   resolveAddChapterNumberClick,
@@ -276,9 +277,6 @@ const getViewOptionsForType = (
   if (viewType === 'markers') return { ...paragraphStructure, noteMode: 'expanded' };
   return paragraphStructure;
 };
-
-// This regex is connected directly to the exception message within MissingBookException.cs
-const bookNotFoundRegex = /Book number \d+ not found in project/;
 
 // This regex is connected directly to the exception message within PermissionsException.cs
 const PERMISSIONS_EXCEPTION_REGEX = /Permissions exception for projectId/;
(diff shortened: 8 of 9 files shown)


katherinejensen00:
## Review summary

The change is real and the diagnosis is right — a book missing from a resource genuinely does come back from the PDP as a `PlatformError`, and replacing the read-only editor with the localized message is the correct call. The extraction of `isBookNotFoundError`, the `debug`/`error` split, and keeping the resource selector mounted are all good.

Three things I'd hold the merge for:

1. The new `bookExists` derivations read the PDP value **without gating on `isLoading`**, which this very file documents as unsafe 20 lines further down. The message now persists (or appears) for a full round trip while pointing at the *wrong* book or the *wrong* resource.
2. The Bible text panel can be showing a **project**, where "…in this resource" is the wrong noun.
3. The accessibility rationale added in three comments is contradicted by guidance that PT-4111 (#2691) — the PR this rebased onto — added itself.

Everything below is anchored to a line this PR changed. I'll add the inline comments separately.

---

### Blocking

**1. `extensions/src/platform-scripture-editor/src/resource-text-panel.web-view.tsx:434` — High**

`bookExists` is derived from a possibly-stale PDP value, so the message outlives the condition. `useProjectData`'s underlying `useData` does not reset its value when the selector changes — it keeps the previous selector's value until the new subscription's first update lands, and the `isLoading` flag returned at line 412 is discarded here.

Concretely: on a resource lacking Genesis, navigate GEN → MAT. `usjPossiblyError` still holds Genesis's `PlatformError`, so `bookExists` stays `false` and the panel keeps asserting "This book does not exist in this resource" **about Matthew** for the whole round trip.

This isn't a hypothesis about the hook — `platform-scripture-editor.web-view.tsx:1313-1319` documents exactly this trap ("doesn't reset its value back to the default when the selector … changes") and guards `isBlankChapter` with `isUsjFromPdpLoading` on the next line. Suggest destructuring the third tuple element here and gating the memo the same way.

**2. `extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx:1296` — High**

Same missing guard — and here `isUsjFromPdpLoading` is *already destructured* at line 1273 and already used for this purpose at line 1322; it just isn't applied to `bookExists`. Same failure mode: navigate from a missing book to a present one and the missing-book view stays up over the loading content.

**3. `extensions/src/platform-scripture-editor/src/model-text-panel.component.tsx:200` — High**

`bookExists` is never reset when `resourceProjectId` **changes**, and the reset that does exist runs too late. `setIsUsjLoading(true)` is at line 208, inside a passive effect, so it runs after paint.

When the effective model text switches from resource A (which lacks the book) to resource B: the render at 461 passes (B is defined), the spinner guard at 466 is skipped (`isUsjLoading` is still `false`), and 478 fires — painting "This book does not exist in this resource" **about resource B before a single byte has been requested from it**, and pushing that false claim into a `role="status"` region. One line fixes it: `setBookExists(true)` alongside line 208.

**4. `extensions/src/platform-scripture-editor/src/resource-text-panel.web-view.tsx:676` — Medium-High**

This panel resolves project references too — `isProjectReference(selectedRef) → resourceProjectId = selectedRef.id` at line 344 — but the message is hardcoded to `%webView_platformScriptureEditor_error_bookNotFoundResource%`, i.e. "This book does not exist in this **resource**." When the selection is a project, that's the wrong noun. `%webView_platformScriptureEditor_error_bookNotFoundProject%` already exists in this extension's `localizedStrings.json` in both locales, and the editable editor solves this by branching on `platform.isPublished`.

---

### Important

**5. `extensions/src/platform-scripture-editor/src/resource-text-panel.web-view.tsx:658` — Medium**

The accessibility rationale here is contradicted by guidance PT-4111 added. From `lib/platform-bible-react/src/stories/shadcn-ui/empty.stories.tsx:214`:

> assistive tech announces *mutations* to a live region that is already in the accessibility tree. Mounting the region and its text in one commit … typically announces nothing in NVDA or JAWS. Keep the `role="status"` element mounted across the transition and swap only its text.

That text was added in commit `35a09c06276` ("fix: address PT-4111 review findings") — i.e. by #2691, the PR this one rebased onto and cites for consistency.

`EmptyState` mounts `<p role="status">` together with its text, in a branch that unmounts the editor, so per that guidance it likely announces nothing. `BookNotAvailableView` has the same structural issue and compensates with the part that actually works — `document.hasFocus()` → `regionRef.focus()` on a `tabIndex={-1}` wrapper (`book-not-available-view.component.tsx:82-89`) — which none of the three new surfaces have. Either add the focus repair, or soften the comments; right now three comments assert something the repo says is false. Same note applies at `platform-scripture-editor.web-view.tsx:1979` and `model-text-panel.component.tsx:476`.

**6. `extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts:1167` — Medium**

Two standards prohibit this pattern, and the PR now pins it with a test:

- `.context/standards/Code-Style-Guide.md:238-255` — "Never compare variables against hardcoded English strings", with the literal counter-example `if (errorMessage.includes("not found"))` → `if (errorCode === ErrorCode.NotFound)`, and "This applies to … string includes/matching."
- `.context/standards/Paranext-Core-Patterns.md:478` — "String-matching on `ex.Message` in TS — read the `platformErrorCode` property off the caught `PlatformError` instead."

**However** — I checked whether the prescribed alternative actually works, and it currently doesn't: `src/main/services/rpc-server.ts:287` rebuilds every forwarded error as `createErrorResponse(message, code, id)`, and `src/shared/data/rpc.model.ts:114` emits `{ code, message }` with no `data` field, so `platformErrorCode` is stripped before `src/shared/services/network.service.ts:274` can read it. **Keeping the regex is defensible today** — this is not a request to change the approach in this PR. But that severed channel is the reason to record, and it deserves its own ticket.

**7. `c-sharp-tests/MissingBookExceptionTests.cs:12` — Low-Medium**

"there is no structured error code to key off" isn't the real reason — `c-sharp/PlatformErrorCodes.cs` exists with `NotFound`, and `c-sharp-tests/ManageBooks/PlatformErrorCodesTests.cs:14` names `MissingBookException.cs` as its neighbour. The accurate reason is the one in #6: the code is discarded in the router before it reaches TS. Worth correcting, because this comment is the artifact that will justify the next string-match.

**8. `extensions/src/platform-scripture-editor/src/model-text-panel.component.tsx:228` — Medium**

`setBookExists(!bookNotFound)` means every non-book-not-found failure sets `bookExists = true`, falls through to the read-only editor with `usj === undefined`, and shows the "Enter some Scripture…" prompt — the exact dishonesty the comment at lines 196-199 describes, for a transient network error on a resource the user cannot edit. The new test at `model-text-panel.component.test.tsx:321` then asserts `findByTestId('editorial')` for a `'Network request failed'` rejection, pinning that behavior as correct. The negative control is worth keeping; consider asserting a neutral/error state rather than the editor.

**9. `extensions/src/platform-scripture-editor/src/model-text-panel.component.test.tsx:288` — Medium**

All four new panel tests reject with `new Error(…)`. Production rejects with a `PlatformError` **plain object** — `network.service.ts:285` throws `newPlatformError(...)`, and every return path in `lib/platform-bible-utils/src/platform-error.ts:141-176` is an object literal, never an `Error` instance. So these tests exercise the `instanceof Error` clause, and the clause that actually runs in the app has no panel-level coverage.

To be fair: `platform-scripture-editor.utils.test.ts:2597` *does* cover the `PlatformError` shape, so the predicate itself is tested — this is test fidelity, not a coverage hole. Swapping the mocks to `newPlatformError('JSON-RPC Request error (-32000): Book number 1 not found in project project-web.')` closes it. Cf. `.context/standards/Testing-Guide.md:186` — "over-mocking that hides real integration issues, giving false confidence."

**10. `extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts:1175` — Low-Medium**

"awaiting `getChapterUSJ` directly rejects with a thrown `Error`" is not accurate, per #9. The same claim appears in the test name at `platform-scripture-editor.utils.test.ts:2617`, in the PR body's "Suggested Review Focus", and in commit `56d02541fb7`. The `instanceof Error` branch is harmless defensive code, but the comment presents it as the production path.

**11. `extensions/src/platform-scripture-editor/src/model-text-panel.component.tsx:226` — Low-Medium**

`logger.debug` is compiled out of packaged builds — `src/main/global-this.model.ts:27` sets the level to `isPackaged ? 'info' : 'debug'`. So in production there is no trace of a book-not-found at all. That may be exactly what you want, but the verification note in the PR body ("verified in `main.log`: the new messages appear at `[debug]`") was necessarily done in a dev build and doesn't demonstrate the packaged behavior. Worth a deliberate choice between `debug` and `info`.

**12. `extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts:1182` — Low-Medium**

On the surface sweep flagged in the PR body: `extensions/src/platform-scripture-editor/src/scripture-text-grid/resource-cell.utils.ts:26` maps *any* `PlatformError` to `'failed'`, which the grid renders as a download failure. So after this PR, a book missing from a resource reads "This book does not exist in this resource" in three panels and "Resource unavailable / download failed" in the grid. Now that `isBookNotFoundError` is exported and shared, plugging in a fourth consumer is a few lines — worth doing here or splitting out with a ticket, rather than leaving as an open question.

---

### Minor

**13. `model-text-panel.component.tsx:196` — Low.** Pre-existing, but adjacent and now claimed by this comment: `usj` is not cleared when a load starts, so the `usj === undefined && isUsjLoading` guard at 466 is false and the editor renders the **previous chapter** for the whole round trip. Same root cause as #1/#3.

**14. `model-text-panel.component.tsx:226` — Low.** `logger` is now used inside the effect but isn't in the deps; it's covered by a blanket `eslint-disable` whose comment only justifies excluding `scrRef.verseNum`. Safe in practice (stable module object) — worth extending that comment rather than leaving it implied.

**15. `c-sharp-tests/MissingBookExceptionTests.cs:29` — Low.** The pin is weaker than the docstring claims. Editing the TS regex *is* caught (by `utils.test.ts:2597`), but a coordinated edit of the regex **and** its test literal still drifts silently. Also `\d` isn't the same in both engines (.NET matches Unicode `Nd`, JS matches `[0-9]`), so "character-for-character identical" doesn't guarantee semantically identical.

**16. `c-sharp-tests/MissingBookExceptionTests.cs:66` — Low.** `BookNumAndProjectId_ArePreserved` is the exact pattern `.context/standards/Testing-Guide.md:242` prohibits ("Constructor assignments — `new User(name).name === name` — Tautological"). `MissingBookException.BookNum` and `.ProjectId` are read nowhere else in the repo. Separately, `:32` is strictly contained in `:51`'s `{1, 40, 66}` loop. Two of the four tests can go.

**17. `c-sharp-tests/MissingBookExceptionTests.cs:5` — Low.** Block-scoped namespace; `.editorconfig:19` sets `csharp_style_namespace_declarations = file_scoped:warning`.

**18. `model-text-panel.component.tsx:485` — Low.** `?? ''` renders a blank `<p role="status">` if the key is missing. The other two call sites degrade differently (raw `%key%`). Worth one consistent choice across the three.

**19. `model-text-panel.component.tsx:478` — Low.** The missing-book branch drops the panel header, so the user can't see *which* model text is missing the book, and the 42px tab alignment goes with it.

**20. PR description & commits — Medium (convention).** Three unfilled placeholders shipped: the body ends with the literal `AI-assisted — [session](<session URL>)`; commit `56d02541fb7` ends with `Session-URL: <session URL>`; commit `ca61b14833c` has no `Session-URL:` line at all. CLAUDE.md → "Git & PR Conventions" requires real URLs. Also still in the published body: *"Screenshot files are in `.review/screenshots/` (gitignored) — attach `before-full-window.png` and `after-full-window.png` here"*, under a Screenshots table with no screenshots.

---

### Checked and explicitly **not** an issue

Chased these and they don't hold — recording them so they don't come back in another review round:

- **`instanceof Error` as a cross-realm break** — it isn't. Production throws a duck-typed `PlatformError` and `isPlatformError` catches it. Only the *comment* and the *test mocks* are wrong (#9, #10).
- **"No test file for `resource-text-panel.web-view.tsx`"** — there are 15 `*.web-view.tsx` files in `extensions/src/` and **zero** `*.web-view.test.tsx`. Extraction into a `.component.tsx` is the repo's answer, and that's exactly what was done for the panel that has one.
- **"Assert the log level in a test"** — `Testing-Guide.md:214` and `:239` classify that as an implementation-detail test.
- **`isPlatformError` evaluated twice / `logger` inside `useMemo` / the borrowed key's position in the alphabetized list** — the dismissals in the PR body are correct on all three.
- **The localization key** resolves in all three panels; both `en` and `es` entries are present, and `MODEL_TEXT_PANEL_STRING_KEYS` is wired through `model-text-panel.web-view.tsx:36`.
- **`EmptyState`'s `id`** is `data-testid`, not a DOM `id` — no duplicate-id concern.

Items 1–4 are the ones I'd block on; the rest are yours to weigh.

---

AI-assisted review — [session](https://claude.ai/code/session_01J7CFWFPTizw8dofpdGUWXY)

katherinejensen00 (changes requested):
Thanks, Jolie, for improving the missing book error messaging! I posted inline comments to make it easier to follow what is being talked about, but feel free to respond in bulk.

> @katherinejensen00 reviewed 9 files and all commit messages, and made 21 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2706#-:-P-h1k_v8i_SEmUn21vg:b-e605nx)* status: all files reviewed, 20 unresolved discussions (waiting on jolierabideau).

___
*[`-- commits` at r1](https://reviewable.io/reviews/paranext/paranext-core/2706#-P-h1gj_7nAFVJivAsiR:-P-h1gj_7nAFVJivAsiS:bk1zmr):*
NIT 20. PR description & commits — Medium (convention). Three unfilled placeholders shipped: the body ends with the literal AI-assisted — [session](<session URL>); commit 56d02541fb7 ends with Session-URL: <session URL>; commit ca61b14833c has no Session-URL: line. CLAUDE.md → "Git & PR Conventions" requires real URLs. Also still in the published body: "Screenshot files are in .review/screenshots/ (gitignored) — attach before-full-window.png and after-full-window.png here", under a Screenshots table with no screenshots.
___
*[`c-sharp-tests/MissingBookExceptionTests.cs` line 5 at r1](https://reviewable.io/reviews/paranext/paranext-core/2706#-P-h1QyZ1Cet50_nZvSu:-P-h1QyZ1Cet50_nZvSv:byg0fri) ([raw file](https://github.com/paranext/paranext-core/blob/56d02541fb7079373afc8658a90b1eeebebd63dd/c-sharp-tests/MissingBookExceptionTests.cs#L5)):*
> ```Smalltalk
> using Paranext.DataProvider;
> 
> namespace TestParanextDataProvider
> ```

NIT 17. c-sharp-tests/MissingBookExceptionTests.cs:5 — Low. Block-scoped namespace; .editorconfig:19 sets csharp_style_namespace_declarations = file_scoped:warning.
___
*[`c-sharp-tests/MissingBookExceptionTests.cs` line 12 at r1](https://reviewable.io/reviews/paranext/paranext-core/2706#-P-h0VbR7xIGaDENYdKH:-P-h0VbR7xIGaDENYdKI:b-e6ia3f) ([raw file](https://github.com/paranext/paranext-core/blob/56d02541fb7079373afc8658a90b1eeebebd63dd/c-sharp-tests/MissingBookExceptionTests.cs#L12)):*
> ```Smalltalk
>     /// <remarks>
>     /// The Scripture editor extension detects "this book is not in this project/resource" by
>     /// pattern-matching the message text of the error the PDP returns — there is no structured
> ```

NIT 7. c-sharp-tests/MissingBookExceptionTests.cs:12 — Severity: Low-Medium

▎ "there is no structured error code to key off" isn't the real reason — c-sharp/PlatformErrorCodes.cs exists with NotFound, and c-sharp-tests/ManageBooks/PlatformErrorCodesTests.cs:14 names MissingBookException.cs as its neighbour. The accurate reason is the one above: the code is discarded in the router before it reaches TS. Worth correcting, because this comment is the artifact that will justify the next string-match.
___
*[`c-sharp-tests/MissingBookExceptionTests.cs` line 29 at r1](https://reviewable.io/reviews/paranext/paranext-core/2706#-P-h1J3JBetBL4PrBWG7:-P-h1J3JBetBL4PrBWG8:b5j3gf6) ([raw file](https://github.com/paranext/paranext-core/blob/56d02541fb7079373afc8658a90b1eeebebd63dd/c-sharp-tests/MissingBookExceptionTests.cs#L29)):*
> ```Smalltalk
>         /// <c>platform-scripture-editor.utils.ts</c>.
>         /// </summary>
>         private const string ExtensionBookNotFoundRegex = @"Book number \d+ not found in project";
> ```

NIT 15. c-sharp-tests/MissingBookExceptionTests.cs:29 — Low. The pin is weaker than the docstring claims. Editing the TS regex is caught (by utils.test.ts:2597), but a coordinated edit of the regex and its test literal still drifts silently. Also, \d isn't the same in both engines (.NET matches Unicode Nd, JS matches [0-9]), so "character-for-character identical" doesn't guarantee semantically identical.
___
*[`c-sharp-tests/MissingBookExceptionTests.cs` line 66 at r1](https://reviewable.io/reviews/paranext/paranext-core/2706#-P-h1NdW0xZl0Imjyfvu:-P-h1NdW0xZl0Imjyfvv:bs9fvli) ([raw file](https://github.com/paranext/paranext-core/blob/56d02541fb7079373afc8658a90b1eeebebd63dd/c-sharp-tests/MissingBookExceptionTests.cs#L66)):*
> ```Smalltalk
> 
>         [Test]
>         public void BookNumAndProjectId_ArePreserved()
> ```

NIT 16. c-sharp-tests/MissingBookExceptionTests.cs:66 — Low. BookNumAndProjectId_ArePreserved is the exact pattern Testing-Guide.md:242 prohibits ("Constructor assignments — new User(name).name === name — Tautological"). I checked: MissingBookException.BookNum and .ProjectId are read nowhere in the repo. Separately, :32 is strictly contained in :51's {1, 40, 66} loop. Two of the four tests can go.
___
*[`extensions/src/platform-scripture-editor/src/model-text-panel.component.tsx` line 196 at r1](https://reviewable.io/reviews/paranext/paranext-core/2706#-P-h14xuCfoVYhOzrSmQ:-P-h14xuCfoVYhOzrSmR:b-r1fda5) ([raw file](https://github.com/paranext/paranext-core/blob/56d02541fb7079373afc8658a90b1eeebebd63dd/extensions/src/platform-scripture-editor/src/model-text-panel.component.tsx#L196)):*
> ```XML
>   // `undefined` means "not yet fetched" so we can show the loading state, matching the original.
>   const [isUsjLoading, setIsUsjLoading] = useState(false);
>   // A book missing from the resource makes `getChapterUSJ` reject rather than return empty USJ.
> ```

NIT 13. model-text-panel.component.tsx:196 — Low. Pre-existing, but adjacent and now claimed by this comment: usj is not cleared when a load starts, so the usj === undefined && isUsjLoading guard at 466 is false and the editor renders the previous chapter for the whole round trip. Same root cause as #1/#3.
___
*[`extensions/src/platform-scripture-editor/src/model-text-panel.component.tsx` line 200 at r1](https://reviewable.io/reviews/paranext/paranext-core/2706#-P-h06fa851K_AFDtENz:-P-h06fa851K_AFDtEO-:bqa8uis) ([raw file](https://github.com/paranext/paranext-core/blob/56d02541fb7079373afc8658a90b1eeebebd63dd/extensions/src/platform-scripture-editor/src/model-text-panel.component.tsx#L200)):*
> ```XML
>   // read-only editor, which then shows either the previous chapter's text or its "enter some
>   // Scripture" placeholder — an edit prompt in a resource the user cannot edit.
>   const [bookExists, setBookExists] = useState(true);
> ```

3. extensions/src/platform-scripture-editor/src/model-text-panel.component.tsx:200 — Severity: High

▎ bookExists is never reset when resourceProjectId changes, and the reset that does exist runs too late.
▎
▎ setIsUsjLoading(true) is at line 208, inside a passive effect — it runs after paint. So when the effective model text switches from resource A (which lacks the book) to resource B: the render at 461 passes (B is defined), the spinner guard at 466 is skipped (isUsjLoading is still false), and 478 fires — painting "This book does not exist in this resource" about resource B before a single byte has been requested from it, and pushing that false claim into a role="status" region. One line fixes it: setBookExists(true) alongside line 208.
___
*[`extensions/src/platform-scripture-editor/src/model-text-panel.component.tsx` line 226 at r1](https://reviewable.io/reviews/paranext/paranext-core/2706#-P-h0tNa1HPmlc_pqi_G:-P-h0tNbCiKLAucibMJM:b-i312ce) ([raw file](https://github.com/paranext/paranext-core/blob/56d02541fb7079373afc8658a90b1eeebebd63dd/extensions/src/platform-scripture-editor/src/model-text-panel.component.tsx#L226)):*
> ```XML
>       // logged quietly; every other failure is a real error. Matches both Scripture tabs.
>       const bookNotFound = isBookNotFoundError(e);
>       if (bookNotFound) logger?.debug(`Book not found in model text: ${getErrorMessage(e)}`);
> ```

NIT 11. extensions/src/platform-scripture-editor/src/model-text-panel.component.tsx:226 — Severity: Low-Medium

▎ logger.debug is compiled out of packaged builds — src/main/global-this.model.ts:27 sets the level to isPackaged ? 'info' : 'debug'. So in production there is no trace of a book-not-found at all. That may well be what you want, but the PR's verification note ("verified in main.log: the new messages appear at [debug]") was necessarily done in a dev build and doesn't demonstrate the packaged behavior. Worth a deliberate choice between debug and info.
___
*[`extensions/src/platform-scripture-editor/src/model-text-panel.component.tsx` line 226 at r1](https://reviewable.io/reviews/paranext/paranext-core/2706#-P-h19dV7vYkS3RtgeW0:-P-h19dV7vYkS3RtgeW1:b-50sue7) ([raw file](https://github.com/paranext/paranext-core/blob/56d02541fb7079373afc8658a90b1eeebebd63dd/extensions/src/platform-scripture-editor/src/model-text-panel.component.tsx#L226)):*
> ```XML
>       // logged quietly; every other failure is a real error. Matches both Scripture tabs.
>       const bookNotFound = isBookNotFoundError(e);
>       if (bookNotFound) logger?.debug(`Book not found in model text: ${getErrorMessage(e)}`);
> ```

NIT 14. model-text-panel.component.tsx:226 — Low. logger is now used inside the effect but isn't in the deps; it's covered by a blanket eslint-disable whose comment only justifies excluding scrRef.verseNum. Safe in practice (stable module object) — worth extending that comment rather than leaving it implied.
___
*[`extensions/src/platform-scripture-editor/src/model-text-panel.component.tsx` line 228 at r1](https://reviewable.io/reviews/paranext/paranext-core/2706#-P-h0bvE1D6ocFsH9YNB:-P-h0bvE1D6ocFsH9YNC:bqvol21) ([raw file](https://github.com/paranext/paranext-core/blob/56d02541fb7079373afc8658a90b1eeebebd63dd/extensions/src/platform-scripture-editor/src/model-text-panel.component.tsx#L228)):*
> ```XML
>       if (bookNotFound) logger?.debug(`Book not found in model text: ${getErrorMessage(e)}`);
>       else logger?.error(`Error getting USJ for the model text panel: ${getErrorMessage(e)}`);
>       setBookExists(!bookNotFound);
> ```

NIT 8. extensions/src/platform-scripture-editor/src/model-text-panel.component.tsx:228 — Severity: Medium

▎ setBookExists(!bookNotFound) means every non-book-not-found failure sets bookExists = true, falls through to the read-only editor with usj === undefined, and shows the "Enter some Scripture…" prompt — the exact dishonesty your comment at lines 196-199 describes, for a transient network error on a resource the user cannot edit. The new test at model-text-panel.component.test.tsx:321 then asserts findByTestId('editorial') for a 'Network request failed' rejection, which pins that behavior as correct. The negative control is worth keeping; consider asserting a neutral/error state rather than the editor.
___
*[`extensions/src/platform-scripture-editor/src/model-text-panel.component.tsx` line 228 at r1](https://reviewable.io/reviews/paranext/paranext-core/2706#-P-h0gkJ7w62f0v_nY9Q:-P-h0gkJ7w62f0v_nY9R:bjzmlba) ([raw file](https://github.com/paranext/paranext-core/blob/56d02541fb7079373afc8658a90b1eeebebd63dd/extensions/src/platform-scripture-editor/src/model-text-panel.component.tsx#L228)):*
> ```XML
>       if (bookNotFound) logger?.debug(`Book not found in model text: ${getErrorMessage(e)}`);
>       else logger?.error(`Error getting USJ for the model text panel: ${getErrorMessage(e)}`);
>       setBookExists(!bookNotFound);
> ```

NIT 9. extensions/src/platform-scripture-editor/src/model-text-panel.component.test.tsx:288 — Severity: Medium

▎ All four new panel tests reject with new Error(…). Production rejects with a PlatformError plain object — network.service.ts:285 throws newPlatformError(...), and every return path in platform-error.ts:141-176 is an object literal, never an Error instance. So these tests exercise the instanceof Error clause, and the clause that actually runs in the app has no panel-level coverage.
▎
▎ To be fair to the PR: platform-scripture-editor.utils.test.ts:2597 does cover the PlatformError shape, so the predicate itself is tested — this is test fidelity, not a coverage hole. Swapping the mocks to newPlatformError('JSON-RPC Request error (-32000): Book number 1 not found in project project-web.') closes it. (Testing-Guide.md:186 — "over-mocking that hides real integration issues, giving false confidence.")
___
*[`extensions/src/platform-scripture-editor/src/model-text-panel.component.tsx` line 478 at r1](https://reviewable.io/reviews/paranext/paranext-core/2706#-P-h1aAEBlyoMbkHLDUS:-P-h1aAEBlyoMbkHLDUT:b-dnui00) ([raw file](https://github.com/paranext/paranext-core/blob/56d02541fb7079373afc8658a90b1eeebebd63dd/extensions/src/platform-scripture-editor/src/model-text-panel.component.tsx#L478)):*
> ```XML
>   // Scripture tabs, so the identical sentence is worded, styled, and announced to screen readers
>   // the same way in every panel that can hit this state.
>   if (!bookExists) {
> ```

NIT 19. model-text-panel.component.tsx:478 — Low. The missing-book branch drops the panel header, so the user can't see which model text is missing the book, and the 42px tab alignment goes with it.
___
*[`extensions/src/platform-scripture-editor/src/model-text-panel.component.tsx` line 485 at r1](https://reviewable.io/reviews/paranext/paranext-core/2706#-P-h1WycBHhy3Snv2tw9:-P-h1Wyd9esaXu0og7Z2:b7dks6) ([raw file](https://github.com/paranext/paranext-core/blob/56d02541fb7079373afc8658a90b1eeebebd63dd/extensions/src/platform-scripture-editor/src/model-text-panel.component.tsx#L485)):*
> ```XML
>           className="tw:text-center"
>           message={
>             localizedStrings['%webView_platformScriptureEditor_error_bookNotFoundResource%'] ?? ''
> ```

NIT 18. model-text-panel.component.tsx:485 — Low. ?? '' renders a blank <p role="status"> if the key is missing. The other two call sites degrade differently (raw %key%). Worth one consistent choice across the three.
___
*[`extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts` line 1167 at r1](https://reviewable.io/reviews/paranext/paranext-core/2706#-P-h0PZn9jjR6Vd5Qr-n:-P-h0PZn9jjR6Vd5Qr-o:b-e1r54k) ([raw file](https://github.com/paranext/paranext-core/blob/56d02541fb7079373afc8658a90b1eeebebd63dd/extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts#L1167)):*
> ```XML
> // `c-sharp-tests/MissingBookExceptionTests.cs` duplicates this pattern and fails if that message is
> // reworded, so the two sides cannot drift silently — update both together.
> const BOOK_NOT_FOUND_REGEX = /Book number \d+ not found in project/;
> ```

6. extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts:1167 — Severity: Medium

▎ Two standards prohibit this pattern, and the PR now pins it with a test.
▎
▎ Code-Style-Guide.md:238-255 — "Never compare variables against hardcoded English strings", with the literal counter-example if (errorMessage.includes("not found")) → if (errorCode === ErrorCode.NotFound), and "This applies to … string includes/matching." Paranext-Core-Patterns.md:478 — "String-matching on ex.Message in TS — read the platformErrorCode property off the caught PlatformError instead."
▎
▎ However — I checked whether the prescribed alternative actually works, and it currently doesn't: src/main/services/rpc-server.ts:287 rebuilds every forwarded error as createErrorResponse(message, code, id), and src/shared/data/rpc.model.ts:114 emits { code, message } with no data field, so platformErrorCode is stripped before network.service.ts:274 can read it. So keeping the regex is defensible today — but that's the reason to record, and it deserves its own ticket. Please don't take this as a request to change the approach in this PR.
___
*[`extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts` line 1175 at r1](https://reviewable.io/reviews/paranext/paranext-core/2706#-P-h0mZRCtx5WH5BURTz:-P-h0mZRCtx5WH5BURU-:br2f5wn) ([raw file](https://github.com/paranext/paranext-core/blob/56d02541fb7079373afc8658a90b1eeebebd63dd/extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts#L1175)):*
> ```XML
>  *
>  * Handles both shapes the scripture PDPs surface, because the two call styles fail differently:
>  * `useProjectData` hands back a `PlatformError` in place of the data, while awaiting
> ```

NIT 10. extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts:1175 — Severity: Low-Medium

▎ "awaiting getChapterUSJ directly rejects with a thrown Error" is not accurate, per #9. The same claim appears in the test name at platform-scripture-editor.utils.test.ts:2617, in the PR body's "Suggested Review Focus", and in commit 56d02541fb7. The instanceof Error branch is harmless defensive code, but the comment presents it as the production path.
___
*[`extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts` line 1182 at r1](https://reviewable.io/reviews/paranext/paranext-core/2706#-P-h0zew9EI1Kzld68uE:-P-h0zew9EI1Kzld68uF:bub4o0r) ([raw file](https://github.com/paranext/paranext-core/blob/56d02541fb7079373afc8658a90b1eeebebd63dd/extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts#L1182)):*
> ```XML
>  *   its methods, which may be a `PlatformError` or an `Error`.
>  */
> export function isBookNotFoundError(possibleError: unknown): boolean {
> ```

NIT 12. extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts:1182 — Severity: Low-Medium

▎ On the surface sweep you flagged: scripture-text-grid/resource-cell.utils.ts:26 maps any PlatformError to 'failed', which the grid renders as a download failure. So after this PR, a book missing from a resource reads "This book does not exist in this resource" in three panels and "Resource unavailable / download failed" in the grid. Now that isBookNotFoundError is exported and shared, plugging in a fourth consumer is a few lines — worth doing here or splitting out with a ticket, but not leaving as an open question.
___
*[`extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx` line 1296 at r1](https://reviewable.io/reviews/paranext/paranext-core/2706#-P-h00NZ5oOd6ejzcDBY:-P-h00NZ5oOd6ejzcDBZ:badpsc3) ([raw file](https://github.com/paranext/paranext-core/blob/56d02541fb7079373afc8658a90b1eeebebd63dd/extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx#L1296)):*
> ```XML
>     // A book missing from the project is ordinary navigation, not a fault, so it is logged
>     // quietly; every other PDP failure is a real error. Both Scripture tabs branch the same way.
>     const bookNotFound = isBookNotFoundError(usjFromPdpPossiblyError);
> ```

2. extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx:1296 — Severity: High

▎ Same missing guard, and here isUsjFromPdpLoading is already destructured at line 1273 and already used for this purpose at line 1322 — it's just not applied to bookExists. Same failure: navigate from a missing book to a present one and BookNotAvailableView / the resource message stays up over the loading content.
___
*[`extensions/src/platform-scripture-editor/src/resource-text-panel.web-view.tsx` line 434 at r1](https://reviewable.io/reviews/paranext/paranext-core/2706#-P-h-nwk9AImFFyy8h31:-P-h-nwk9AImFFyy8h32:bnnmiv5) ([raw file](https://github.com/paranext/paranext-core/blob/56d02541fb7079373afc8658a90b1eeebebd63dd/extensions/src/platform-scripture-editor/src/resource-text-panel.web-view.tsx#L434)):*
> ```XML
>   // chapter's content or its "enter some scripture" placeholder — neither is honest for a read-only
>   // resource. Derived in one pass, like the editable Scripture tab.
>   const [usjFromPdp, bookExists] = useMemo<[Usj | undefined, boolean]>(() => {
> ```

1. extensions/src/platform-scripture-editor/src/resource-text-panel.web-view.tsx:434 — Severity: High

▎ bookExists is derived from a possibly-stale PDP value, so the message outlives the condition.
▎
▎ useProjectData's underlying useData does not reset its value when the selector changes — it keeps the previous selector's value until the new subscription's first update lands. The isLoading flag returned at line 412 is discarded here. Concretely: on a resource lacking Genesis, navigate GEN → MAT. usjPossiblyError still holds Genesis's PlatformError, so bookExists stays false and the panel keeps asserting "This book does not exist in this resource" about Matthew for the whole round trip.
▎
▎ This isn't a hypothesis about the hook — platform-scripture-editor.web-view.tsx:1313-1319 documents exactly this trap in your own words ("doesn't reset its value back to the default when the selector … changes"), and guards isBlankChapter with isUsjFromPdpLoading on the next line. Please destructure the third tuple element here and gate the memo the same way.
___
*[`extensions/src/platform-scripture-editor/src/resource-text-panel.web-view.tsx` line 658 at r1](https://reviewable.io/reviews/paranext/paranext-core/2706#-P-h0K-CCluFgUCogBkc:-P-h0K-CCluFgUCogBkd:bofl4me) ([raw file](https://github.com/paranext/paranext-core/blob/56d02541fb7079373afc8658a90b1eeebebd63dd/extensions/src/platform-scripture-editor/src/resource-text-panel.web-view.tsx#L658)):*
> ```XML
>           that does contain the book. The message sits outside the `dir` wrapper because it is
>           UI-locale text, not resource text. `EmptyState` (rather than bare text) because this
>           region swaps content in place as the user navigates, so its `role="status"` is what
> ```

5. extensions/src/platform-scripture-editor/src/resource-text-panel.web-view.tsx:658 — Severity: Medium

▎ The accessibility rationale here is contradicted by guidance PT-4111 added.
▎
▎ lib/platform-bible-react/src/stories/shadcn-ui/empty.stories.tsx:214: "assistive tech announces mutations to a live region that is already in the accessibility tree. Mounting the region and its text in one commit … typically announces nothing in NVDA or JAWS. Keep the role="status" element mounted across the transition and swap only its text." I traced that text to commit 35a09c06276, "fix: address PT-4111 review findings" — #2691, the PR this one rebased onto.
▎
▎ EmptyState mounts <p role="status"> together with its text in a branch that unmounts the editor, so per that guidance it likely announces nothing. BookNotAvailableView has the same structural issue and compensates with the part that actually works — document.hasFocus() → regionRef.focus() on a tabIndex={-1} wrapper (book-not-available-view.component.tsx:82-89) — which none of the three new surfaces have. Either add the focus repair, or soften the comments. Right now three comments assert something the repo says is false. Same note applies at platform-scripture-editor.web-view.tsx:1979 and model-text-panel.component.tsx:476.
___
*[`extensions/src/platform-scripture-editor/src/resource-text-panel.web-view.tsx` line 676 at r1](https://reviewable.io/reviews/paranext/paranext-core/2706#-P-h0ESVDdbQzN5ZNPBP:-P-h0ESVDdbQzN5ZNPBQ:b-f1ryit) ([raw file](https://github.com/paranext/paranext-core/blob/56d02541fb7079373afc8658a90b1eeebebd63dd/extensions/src/platform-scripture-editor/src/resource-text-panel.web-view.tsx#L676)):*
> ```XML
>             className="tw:text-center"
>             message={
>               localizedStrings['%webView_platformScriptureEditor_error_bookNotFoundResource%']
> ```

4. extensions/src/platform-scripture-editor/src/resource-text-panel.web-view.tsx:676 — Severity: Medium-High

▎ This panel resolves project references too — isProjectReference(selectedRef) → resourceProjectId = selectedRef.id at line 344 — but the message is hardcoded to %…bookNotFoundResource%, i.e. "This book does not exist in this resource." When the selection is a project, that's the wrong noun. %webView_platformScriptureEditor_error_bookNotFoundProject% already exists in this extension's localizedStrings.json in both locales, and the editable editor solves this by branching on platform.isPublished. For a PR about telling the user the truth, this one is worth catching.

jolierabideau:
Superseded by #2704 and its findings were carried over

PR #2699, opened 2026-08-20
PT-4092: Collapse the Find books summary and stop popover clipping

jolierabideau:
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2699)

---

PT-4092 fixes two defects in the Find panel's book-scope UI.

1. **The "Showing" row listed every selected book**, so selecting them all overflowed the panel and
   forced a horizontal scrollbar on the whole web view. It now collapses to a summary: "All books"
   when every book *available in the project* is selected, an individual list up to five books, and
   a canon-order `GEN - HOS` range beyond that.
2. **The Select-books popover had no height cap**, so a trigger sitting low in a narrow panel made
   Radix flip it upward, where its full height overran the top of the web view's iframe and clipped
   the search input away entirely. That popover and the Find scope popover that hosts it are now
   capped to `--radix-popover-content-available-height`, and the book list shrinks and scrolls
   instead of pushing the popover out.

## Before / after

Captured in the app (project GTP8, 40 books present, all of them selected).

### 1. The Find "Showing" row with every book selected

**Before** — the row lists all 40 ids, runs off the right edge of the panel, and the whole web view
grows a horizontal scrollbar:
1-before-showing-row-overflows

<details><summary>Full window, showing the horizontal scrollbar</summary>
1-before-full-window-hscrollbar

</details>

**After** — the row collapses to `All books` (every book *available in the project* is selected):
1-after-showing-all-books

> Note: this screenshot predates review round 3 and still shows the tooltip that was added and then
> removed (see "Where the details surface lives" below). The row text is what it demonstrates.

### 2. The Select-books picker in a short panel (window 1400×470)

**Before** — the picker flips upward and overruns the top of the web view's iframe
(`getBoundingClientRect().top = -63px`); the search input is clipped away entirely and only a sliver
of "Select all / Clear all" survives:
2-before-picker-clipped

**After** — the popover is capped to `--radix-popover-content-available-height` with
`collisionPadding={8}` (`top = 8px`, height 306px instead of 377px). The search input stays on
screen and reachable, and the book list shrinks and scrolls internally:

2-after-picker-capped

---

## Public API surface

Purely additive — nothing removed, no signature narrowed, no exported name changed.

- **`platform-bible-react/experimental`** gains two exports:
  - `getAvailableBookIds(availableBookInfo: string): string[]` — delegates to
    `getBookIdsFromBooksPresent` (which owns the `BooksPresent` wire format) plus the obsolete-book
[...]

PR #2708, opened 2026-08-21
PT-3299: Exclude extra-scriptural books from Find

jolierabideau:
> **Round 2 (2026-08-24).** Merged `origin/main` through PT-4339 (#2682) — the PR is `MERGEABLE`
> again, and #2699's picker sizing className and this branch's separator fix both survived the merge
> (review finding 3). Of the round-1 review's 30 findings, 26 are fixed and 4 are deferred with
> reasons in the discussion thread (7 and 11 → PT-4415, 29 confirm-intended, 30 filed separately as
> pre-existing). The sections below have been corrected where round 2 falsified them; the load-bearing
> changes are the error-branch fix (a transient `booksPresent` read failure no longer wipes the user's
> saved book selection), the move of book-list policy into `find/find-book-lists.utils.ts` so the
> shared `getAvailableBookIds` is the only decoder, and ADR-0025.

# Code Review Summary

**Branch**: pt-3299-exclude-extra-books-from-find

**Base**: origin/main

**Date**: 2026-08-24

**Files changed**: 14 (5 source/test + 4 shared-lib source + 1 localization + 4 rebuilt `dist` artifacts)

**Jira**: [PT-3299](https://paratextstudio.atlassian.net/browse/PT-3299) · **PR**: [#2708](https://github.com/paranext/paranext-core/pull/2708)

## Overview

Find resolves a result's location by walking the `\c` and `\v` markers of the book it matched in.
Extra material (GLO, FRT, INT, XXA, etc.) is organized by paragraph markers rather than verses, so
every match in one resolved to the same reference — in the reported case, all GLO results showed as
`GLO 1:0`. Those books also can't be opened in Simple or Power, so a correctly-labeled result there
still wouldn't be actionable. Per the approach agreed with the epic lead (option 3 of three
considered), this change excludes them rather than trying to resolve their locations correctly.

The exclusion happens in one derivation, `deriveFindBookLists`, which feeds both consumers: the
`availableBooksIds` the search runs over and the `booksPresent` the scope selector builds its book
picker from. All four Jira Definition-of-Done items are met.

Review found no critical issues and confirmed the core design is right — `Canon.isExtraMaterial` is
the correct existing predicate (it covers exactly the 15 extra-material books and is the same one
`getSectionForBook` uses for `Section.Extra`), and clearing flags in place rather than dropping
positions is the non-obvious correct call. Four important findings were addressed during review, the
largest being that the `book`/`chapter` scopes bypass the filter entirely. Two follow-up tickets were
filed for work deliberately left out.

## API Changes

No public API surface changed in the original diff — `find.utils.ts` is extension-internal, and
`platform-scripture.d.ts` is untouched.

In-review changes did add to the `platform-bible-react` public surface (both additive and optional,
so no consumer breaks):

- `ScopeSelectorProps.disabledSectionExplanations?: Partial<Record<Section, string>>` — new optional prop
- `SelectBooksProps.disabledSectionExplanations?: Partial<Record<Section, string>>` — new optional prop
- `SelectBooksPickerProps.disabledSectionExplanations?: Partial<Record<Section, string>>` — new optional prop
- `SectionButton` gained an internal `disabledExplanation?: string` prop (not exported)
- `lib/platform-bible-react/dist/` regenerated accordingly

New extension-internal module `find/find-book-lists.utils.ts` exporting `excludeExtraMaterialBooks`,
`deriveFindBookLists`, `FindBookLists`, and `UNKNOWN_FIND_BOOK_LISTS`. It is web-view-only, which is
what keeps `platform-bible-utils/experimental` out of the backend-reachable `find.utils.ts`.
`prunePresentBookIds` stays in `find.utils.ts`, with its first parameter widened to
`readonly string[] | undefined` (extension-internal; the sole caller was updated).

`FindProps.hasExcludedExtraMaterial: boolean` — new required prop on the extension-internal `Find`
component, so the explanation is offered only to projects that actually have extra material.

New localized string: `%webView_find_extraMaterialNotSearched%` (English only; other languages fall
back as usual).

None.

- [x] **The `book`/`chapter` scopes bypass the filter, and the comment claimed they didn't.** `findScope` builds those scopes from `verseRefSetting.book`, not from the filtered book list, so with the current reference in an extra-material book they still search it and still report the useless reference. Verified reachable: the navigation control's book list is unfiltered (`platform-bible-toolbar.tsx:152-156`, `web-view.component.tsx:488-491`, `scroll-group-navigation.commands.ts:99-104`), `availableScopes` is a hardcoded array with no guard (`find.component.tsx:992`), and `isFindQueryValid` doesn't check book availability. _(fixed during review: corrected both comments to state exactly what the filter covers and why the scopes bypass it; filed PT-4415 to close the path; fixed the one new artifact this PR introduced — `localizedBookData` now derives from `localizableBookIds`, all present non-obsolete books, so the scope label keeps a real localized name instead of degrading to a raw `GLO`. The underlying search-GLO-via-current-book behavior is pre-existing, not a regression, and is deliberately left to PT-4415.)_
- [x] **The seam where the filter is wired in had no test.** All coverage was on the pure helper; deleting `excludeExtraMaterialBooks` from the memo left the whole suite green. A test in `find.component.test.tsx` (the analysis agent's suggestion) would NOT have caught this — that file passes `booksPresent` in as an already-filtered prop, so the component sits downstream of the filter. _(fixed during review: moved the derivation itself out of the hook into `deriveFindBookLists` in `find.utils.ts`, which now owns all of Find's book-list policy and returns `{ searchableBooksPresent, availableBookIds, localizableBookIds }`; the memo only calls it and handles the error case. Five tests cover it. Mutation-verified: replacing `excludeExtraMaterialBooks(booksPresent)` with `booksPresent` now fails 2 tests, where before it failed none.)_
- [x] **Negative-case coverage over the extra-material set was partial.** The test asserted clearing for 4 of the 15 extra-material books, so an implementation hardcoding exactly those four would have passed — and `OTH`, named in the Jira DoD, wasn't among them. _(fixed during review: added `ALL_EXTRA_MATERIAL_BOOK_IDS` as 15 literals plus a case asserting all 15 are cleared while `GEN`/`REV` survive. Round 2: the doc no longer calls it an independent oracle — it is `Canon.nonCanonicalIds` verbatim, so it pins the expected set against a canon change rather than cross-checking the predicate.)_
- [x] **The removal instruction named no ticket.** "Drop this exclusion once…" was forward-facing but nothing linked to the work that unblocks it, which the forward-facing-comments rule and Jira DoD #4 both ask for. _(fixed during review: `TODO(PT-4414)` on both the filter and the `availableBookIds`/`localizableBookIds` split that exists only to compensate for it; `PT-4415` referenced at the scope-gap comment.)_

### Minor — Consider

- [x] **`prunePresentBookIds` overloaded "empty" to mean "not known yet."** The filter introduced a second, legitimate way for `availableBooksIds` to be empty — a project whose only books are extra material — which the guard misread as "still loading," leaving a stale `GLO` selection live and searchable. This was the last remaining path where the `selectedBooks` scope could still search extra material. _(fixed during review: the parameter is now `readonly string[] | undefined`, where `undefined` explicitly means not-known-yet; the web view withholds the list while `isBooksPresentLoading`. Confirmed the loading flag covers a project switch — it's re-set to `true` synchronously whenever the data provider or selector changes, `create-use-data-hook.util.ts:157-159` — so this doesn't trade a rare stale selection for a common wiped one. Two tests pin both halves.)_
- [x] **The "Extra" quick-select button stays visible and permanently disabled.** The dropdown list correctly drops an empty section, but `SelectBooks` renders a `SectionButton` for all four `Section` values unconditionally, so a greyed "Extra" button now shows for every project — reading as "this project has no extra material" rather than "Find won't search it." _(fixed during review: added an optional `disabledSectionExplanations` map threaded `ScopeSelector` → `SelectBooks` → `SectionButton`, which wraps the disabled button in the existing `DisabledActionTooltip` — the same primitive Find already uses on its Replace buttons, which handles the disabled-button pointer-events problem and keeps the explanation reachable for keyboard and screen-reader users. Find supplies text for `Section.Extra` via a new localized string.)_
- [x] **Trailing-separator artifact in `select-books-picker`.** Separators were emitted per-section by index while empty sections were skipped, so the last rendered group was followed by a dangling rule. Pre-existing for any project without extra books; this PR made it unconditional for Find. _(fixed during review: sections are filtered to the non-empty ones first, and the separator is emitted before each group but the first.)_
- [x] **`excludeExtraMaterialBooks` recomputed the extra-material set on every call** (123 `bookNumberToId` + `isExtraMaterial` calls). _(fixed during review: precomputed into a module-level `EXTRA_MATERIAL_BOOK_NUMBERS` set, with the out-of-canon reasoning moved to its doc.)_
- [x] **Vocabulary mixing** — `ExtraMaterial` in identifiers (matching `Canon` and `Section.Extra`) vs "extra-scriptural" in prose, within a few lines of each other. _(fixed during review: "extra material" is the only term used; round 2 removed the remaining "extra-scriptural" uses, including the ones in test names.)_
- [x] **Test used bare `''`/`'000'` rather than the real default.** _(fixed during review: imports `BOOKS_PRESENT_DEFAULT` and asserts against it. Round 2 dropped the extra non-canon-length cases, which documented robustness the pipeline does not have. Note the analysis described `BOOKS_PRESENT_DEFAULT` as "123 zeros"; it is actually `''`, so the original `''` case was already the real default value by coincidence. Using the constant keeps it correct if the default ever changes.)_
- [ ] ~~Silent empty-selection state: a saved `selectedBooks` scope containing only extra material now empties, the trigger falls back to `Showing …`, and Find is disabled with no explanation.~~ _(Author: leave it. The state is coherent and fail-safe, and messaging here is the same UX decision PT-4415 already owns — better decided once, there, than split across two surfaces.)_
- [ ] ~~Placement/reuse: `excludeExtraMaterialBooks` is a generic flag-string transform whose sibling `getBookIdsFromBooksPresent` lives in `platform-bible-utils/scripture-util.ts`.~~ _(Author: leave it. It encodes a Find-specific policy; promoting it to a shared util would invite other consumers to adopt that rule, and PT-4414 deletes it outright.)_
- [ ] ~~"leaves flags past the end of the canon alone" asserts an internal guard rather than observable behavior.~~ _(Author: keep the test. The guard is load-bearing — `bookNumberToId` past the canon returns a `***` placeholder for which `isExtraMaterial` is `false`, so an unguarded version would pass for the wrong reason. Worth pinning.)_

### Not done — deliberately deferred

- **Hide the empty "Extra" section entirely** instead of showing a disabled button with a tooltip. This is the cleaner end state, but it needs a new "hide this section" capability on the shared `SelectBooks` component (the tooltip needed only an optional explanation prop), and it can't be driven from the single `booksPresent` filter point. Flagged for the reviewer as the alternative worth weighing against the tooltip.
- **Gating the `book`/`chapter` scopes** — PT-4415.
- **Removing the exclusion when extra material becomes openable** — PT-4414.

#### Shared Regions Modified

None. No `#region shared with <url>` markers in any changed file (the `#region` markers in
`find.web-view.tsx` are plain code-folding regions).

#### Extension Config Changes

None. No `package.json`, `tsconfig*`, eslint, webpack, or workflow changes.

## Positive Observations

- **The right existing predicate was reused rather than hand-rolled.** `Canon.isExtraMaterial` covers exactly `XXA-XXG, FRT, BAK, OTH, INT, CNC, GLO, TDX, NDX` (15 books, numbers 93-102 and 107-111 — 103-106 are `3ES`, `EZA`, `5EZ`, `6EZ`, which are not extra material) — a superset of everything the DoD names and nothing else. Every other canon entry classifies as OT/NT/DC, so nothing verse-organized is lost; notably the tail books `DAG PS3 2BA LBA JUB ENO 1MQ 2MQ 3MQ REP 4BA LAO` are DC and correctly retained. Going through `Canon` directly rather than `getSectionForBook` is also the better choice, since the latter throws on unrecognized ids.
- **Clearing flags in place rather than dropping positions is the non-obvious correct call**, and it's both documented and covered by a test: `getAvailableBookIds` throws unless the string is exactly `Canon.allBookIds.length`.
- **The out-of-canon guard is genuinely needed**, not defensive noise — `bookNumberToId` past the canon returns a `***` placeholder for which `isExtraMaterial` is `false`.
- **Single-point placement pays off downstream**: because `availableBooksIds` derives from the filtered string, the localization key list, the prune effect, and the scope selector's book picker all stay consistent automatically, and previously-persisted extra-book selections get pruned through existing machinery with no new code path. Select-all, the "Books selected: N" count, section-fully-selected state, and the badge list all follow for free.
- **Tests are falsifiable and non-tautological.** The `booksPresentFor`/`presentBookIds` helpers encode flag-string layout via `bookIdToNumber`/`bookNumberToId` without touching `isExtraMaterial`, so they act as an independent oracle rather than mirroring the implementation.
- **Comments pass the strip-the-PR-context test** — no `PT-3299` narration, no "previously"/"used to", no review-finding IDs or stage tags. Everything is rationale, an invariant, or a forward-looking removal condition.
- **Test-file conventions followed**: module-level helpers with TSDoc above their `describe`, matching the existing `matchAll` and `FAKE_UPDATE` patterns, and placed adjacent to the related `prunePresentBookIds` suite.
- Searching the picker for an excluded book lands on the existing `CommandEmpty` zero state, so the region is never left blank.
- No obsolete-book regression: the `Canon.isObsolete` filter that already existed was preserved through the refactor into `getPresentBookIds`.

## Interview Notes

**Stated purpose**: exclude extra material from Find, per PT-3299 and the approach confirmed with the
epic lead.

**Author demonstrated strong understanding of the change.** Asked whether an extra-material current
reference was reachable, the author answered before any research was done — "I don't think there is
anything preventing the current global scrRef from being in GLO" — which turned out to be exactly
right, and which the subsequent trace confirmed through three separate unfiltered code paths. The
design-explanation question was skipped as unnecessary. No areas of deferral or uncertainty; the
author made every scoping call in this review and gave reasons for each.

- **On the scope gap**: chose to correct the comment and fix the label degradation rather than gate the scopes, on the grounds that PT-3299's chosen approach was explicitly the low-effort, easily-reversed one, and gating starts rebuilding the "handle extra books properly" work that option 1 was rejected for. Filed PT-4415 rather than leaving it undocumented.
- **On the disabled "Extra" button**: chose the tooltip, and explicitly asked that the hide-the-section alternative be recorded for the reviewer rather than silently dropped.
- **On the empty-selection message**: declined to add one, reasoning that it's the same UX decision PT-4415 owns and shouldn't be split across two surfaces. Consistent with the ticket's rejection of the "show results with a disclaimer" option.

**Reviewer note on scope growth**: this PR started as 3 files and grew to 14 during review, including
4 files in the shared `platform-bible-react` component library. That growth was authorized
finding-by-finding, not accidental, but it is worth a look — the tooltip in particular added two
public props to shared components for a Find-specific need.

## In-Review Quality Check

All checks run after the in-review changes, all clean:

- `npm run build:platform-bible-react` — succeeded (required before typecheck, since the extension consumes the built `dist`). One `import/no-named-as-default` warning surfaced on the first build and was fixed by switching to the named import.
- `npm run typecheck` — clean, no errors.
- `npm run lint` — 0 errors repo-wide. One real error was introduced and fixed during review (`prefer-destructuring` on the `bookLists` field assignments, resolved by destructuring). Remaining output is pre-existing warnings only.
- `npm test` — all 11 workspace suites pass (1831 + 1072 + 795 + 449 + 401 + 202 + 193 + 87 + 35 + 16 + 12 tests; 1 pre-existing skip).
- Prettier run over the changed files only.
- One test expectation of mine was wrong on first write (canonical order puts `MAT` before `TOB`) and was corrected against actual output rather than by assumption.

`dist/` is tracked in this repo and committed alongside source in prior PRs (including PT-3362 on
Find), so the regenerated bundle is included. The large `index.js` diff is bundler output
reordering; the meaningful `index.d.ts` diff is exactly the four expected prop/signature additions.

## Suggested Review Focus

- [ ] **The tooltip's cost/benefit.** It added two optional public props to `ScopeSelector` and `SelectBooks` to explain a Find-specific exclusion. Is that the right home, or should the empty section be hidden instead (needs a different shared-component capability), or should the disabled button just be left bare?
- [ ] **The `availableBookIds` / `localizableBookIds` split.** Localizing names for books the search won't cover is deliberate — it keeps the "Current book" scope label from degrading to a raw `GLO` — but it is a subtlety a future reader could easily "simplify" away. `TODO(PT-4414)` marks it for removal; is that enough?
- [ ] **The `prunePresentBookIds` signature change.** Switching from "empty means unknown" to an explicit `undefined` is more honest, but it moves the loading judgment to the caller. Worth confirming `isBooksPresentLoading` is the right signal at that call site.
- [ ] **PT-4415's UX decision.** Two findings were deferred to it (gating the scopes, and whether to explain an emptied selection). Decide the messaging approach once, there — noting the ticket already rejected a results-list disclaimer.
- [ ] **Scope growth into the shared lib.** Four `platform-bible-react` source files changed for what began as a 3-file extension fix. Confirm the separator fix and the tooltip props belong in this PR rather than a follow-up.
- [ ] **Confirm the extra-material set is right for Paratext's purposes.** Review relies on `Canon.isExtraMaterial`'s 15 books being exactly what should be excluded. That matches `Section.Extra` in the picker, but a domain check is cheap: `XXA-XXG, FRT, BAK, OTH, INT, CNC, GLO, TDX, NDX`.

---

Find resolves a result's location by walking the \c and \v markers of the book it matched in. Extra material (GLO, FRT, INT, XXA, etc.) is organized by paragraph markers rather than verses, so every match in one resolved to the same reference — in the reported case, all GLO results showed as GLO 1:0. Those books also can't be opened in Simple or Power, so a correctly-labeled result there still wouldn't be actionable.

Exclude them instead, at the single point every consumer reads: deriveFindBookLists in find/find-book-lists.utils.ts. That covers both availableBooksIds (what the search runs over and what the selection is pruned against) and the booksPresent string feeding the scope selector's book picker, so a user can't pick a book the search never covers.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2708)

[PT-3299]: https://paratextstudio.atlassian.net/browse/PT-3299?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


katherinejensen00 (changes requested):
Nineteen of the twenty-one round-1 findings are closed, and the two structural calls are right —
deleting the tooltip is the correct read of `responsiveness.mdx:19`, and delegating to
`getBookIdsFromBooksPresent` is what item 3 was asking for. I revert-tested the new picker story
myself and it's a genuine guard.

Six findings remain open. Two are blocking, and they compound into one symptom: **selecting every
book renders `Showing %webView_scope_selector_all_books%` and still grows the horizontal
scrollbar** — the case in this PR's own "after" screenshot. Both were introduced by the round-2
commit.

Legend: ✅ = reproduced by running it · ⚠️ = verified by reading, symptom not measured.

### Blocking

**22. The "All books" string is never fetched, so the trigger renders the raw key** ✅
`extensions/src/platform-scripture/src/find.web-view.tsx:554` — **High**

Reverting `scope-selector.component.tsx` to `main` took `%webView_scope_selector_all_books%` out of
`SCOPE_SELECTOR_STRING_KEYS` (it was at `:61` in `8b0326b`), and nothing added it to
`FIND_LOCALIZED_STRING_KEYS` (`find.component.tsx:81`) to compensate. `find.web-view.tsx:554-558`
requests exactly `SCOPE_SELECTOR_STRING_KEYS`, so the read at `find.component.tsx:635` is
`undefined` **always** — not just during the localization round-trip — and the fallback at `:636`
becomes the shipped string.

`grep -rn all_books lib/platform-bible-react/src extensions/src/platform-scripture/src` returns
three hits: the read, its fallback, one test. Zero key lists. The bundle agrees —
`webView_scope_selector_all_books` appears **0 times** in `dist/index.js`.

Item 18's `fallbackKey` is dead code until this is fixed: a key that is never requested never falls
back, so `"Todos los libros"` is never reached either.

The tests can't catch it. `buildBooksScopeProps` (`find.component.test.tsx:641`) *replaces*
`scopeSelectorLocalizedStrings` with a hand-built object containing the key, so every books-scope
test runs against a props shape `find.web-view.tsx` cannot produce. `buildProps:100` uses the real
production set; spreading it would have failed.

Since `scope-selector.component.tsx` is back to zero changes, this string is now Find-owned living
under a ScopeSelector name. `%webView_find_allBooks%` in `FIND_LOCALIZED_STRING_KEYS`, read from
`localizedStrings`, keeps the revert intact and puts the key with its only consumer.

**23. The truncate backstop is inert — `Button` ships `shrink-0`** ✅
`extensions/src/platform-scripture/src/find/find.component.tsx:1001` — **High**

`tw:min-w-0` does nothing while `flex-shrink: 0` stands, and the shadcn `Button` base carries
`tw:shrink-0` (`shadcn-ui/button.tsx:22`). Nothing in the added className overrides it, so the
trigger keeps its full content width and the `tw:truncate` span at `:1006` never clips.

Measured in the Storybook Chromium runner against this exact markup in a 260px panel:

| | computed `flex-shrink` | panel client / scroll | summary width / scrollWidth |
|---|---|---|---|
| as written | **0** | 258 / **359** | 249.6 / 250 (no clipping) |
| `+ tw:shrink` | 1 | 258 / **258** | 121.5 / 250 (clips) |

359 > 258 is the horizontal scrollbar back. The sibling gets this right because it uses
`tw:min-w-0 tw:flex-1 tw:truncate` (`scope-selector.component.tsx:759`) — `flex-1` is what
overrides `shrink-0`; `min-w-0` alone was never the load-bearing half. Adding `tw:shrink` (or
`tw:flex-1`) to the Button is the whole fix, verified above.

With 22, the rendered text is the 35-character raw key, so both land on the flagship "select every
book" case. Worth noting PT-4342 (#2680, now on `main`) docks Find as a permanent Column-3 tab in
Simple mode — a narrow column — which makes this the common width, not the edge.

### Blocking on mechanics

**24. Merge conflict and duplicate ADR numbers** ⚠️
`.context/standards/Architecture-Decisions.md:902` — **Medium**

`origin/main` advanced to `2b31b0e8820` (PT-4342, #2680) after this branch's last sync and appended
its own **ADR-0020** ("Find follows the editor onto read-only resources") and **ADR-0021**
("Column 3 tab order…"). GitHub already reports this PR as `CONFLICTING`. Renumber to 0022/0023 on
the re-sync; the file conflicts either way since both sides append at EOF. That commit also touches
`find.component.tsx`, `find.web-view.tsx` and `find-web-view-state.utils.ts` — different hunks, so
no textual conflict expected, but worth a look after merging.

**25. CI has never run on the round-2 work** ⚠️ — **Medium**

Neither `8b0326b` nor `109e4c4` has a check run; the last green Test/CodeQL is `24dd761` (Aug 21).
The reverts, the `dist/` regeneration and typecheck are all unverified by the pipeline. Locally the
`localizedStrings`, keeps the revert intact and puts the key with its only consumer.

**23. The truncate backstop is inert — `Button` ships `shrink-0`** ✅
`extensions/src/platform-scripture/src/find/find.component.tsx:1001` — **High**

`tw:min-w-0` does nothing while `flex-shrink: 0` stands, and the shadcn `Button` base carries
`tw:shrink-0` (`shadcn-ui/button.tsx:22`). Nothing in the added className overrides it, so the
trigger keeps its full content width and the `tw:truncate` span at `:1006` never clips.

Measured in the Storybook Chromium runner against this exact markup in a 260px panel:

| | computed `flex-shrink` | panel client / scroll | summary width / scrollWidth |
|---|---|---|---|
| as written | **0** | 258 / **359** | 249.6 / 250 (no clipping) |
| `+ tw:shrink` | 1 | 258 / **258** | 121.5 / 250 (clips) |

359 > 258 is the horizontal scrollbar back. The sibling gets this right because it uses
`tw:min-w-0 tw:flex-1 tw:truncate` (`scope-selector.component.tsx:759`) — `flex-1` is what
overrides `shrink-0`; `min-w-0` alone was never the load-bearing half. Adding `tw:shrink` (or
`tw:flex-1`) to the Button is the whole fix, verified above.

With 22, the rendered text is the 35-character raw key, so both land on the flagship "select every
book" case. Worth noting PT-4342 (#2680, now on `main`) docks Find as a permanent Column-3 tab in
Simple mode — a narrow column — which makes this the common width, not the edge.

### Blocking on mechanics

**24. Merge conflict and duplicate ADR numbers** ⚠️
`.context/standards/Architecture-Decisions.md:902` — **Medium**

`origin/main` advanced to `2b31b0e8820` (PT-4342, #2680) after this branch's last sync and appended
its own **ADR-0020** ("Find follows the editor onto read-only resources") and **ADR-0021**
("Column 3 tab order…"). GitHub already reports this PR as `CONFLICTING`. Renumber to 0022/0023 on
the re-sync; the file conflicts either way since both sides append at EOF. That commit also touches
`find.component.tsx`, `find.web-view.tsx` and `find-web-view-state.utils.ts` — different hunks, so
no textual conflict expected, but worth a look after merging.

**25. CI has never run on the round-2 work** ⚠️ — **Medium**

Neither `8b0326b` nor `109e4c4` has a check run; the last green Test/CodeQL is `24dd761` (Aug 21).
The reverts, the `dist/` regeneration and typecheck are all unverified by the pipeline. Locally the
`dist/` looks correctly regenerated — no `formatSelectedBooksList` in
`index.js`/`index.cjs`/`experimental.js`, root `index.d.ts` clean, `experimental.d.ts` carrying both
symbols with the new TSDoc — but "verify no files changed after build" is the job that settles it.
`dist/` looks correctly regenerated — no `formatSelectedBooksList` in
`index.js`/`index.cjs`/`experimental.js`, root `index.d.ts` clean, `experimental.d.ts` carrying both
symbols with the new TSDoc — but "verify no files changed after build" is the job that settles it.
`dist/` looks correctly regenerated — no `formatSelectedBooksList` in
`index.js`/`index.cjs`/`experimental.js`, root `index.d.ts` clean, `experimental.d.ts` carrying both
symbols with the new TSDoc — but "verify no files changed after build" is the job that settles it.

### Non-blocking

**26. `find.web-view.tsx` now duplicates `getAvailableBookIds` byte for byte** ⚠️
`extensions/src/platform-scripture/src/find.web-view.tsx:704` — **Low-Medium**

```ts
getBookIdsFromBooksPresent(booksPresent).filter((bookId) => !Canon.isObsolete(Canon.bookIdToNumber(bookId)))

That is now literally the body of getAvailableBookIds. Item 3 was "one decoder", and the
disagreement is gone — but the copy sits three lines from a function find.component.tsx already
imports, and this file already imports from platform-bible-react/experimental (:16). Swapping it
in drops the local Canon dependency too.

27. Nits

- localizedStrings.json:6 — the new metadata entry is inserted between
  %webView_find_projectFilter_noProjectsFound% and %webView_find_projectSelector_label%,
  splitting the %webView_find_*% group the block is organized by.
- find.component.tsx:992 — "min-w-0 on both children" describes the row and the trigger; the row's
  other child (the count/nav div at :1038) has neither. And its premise — "falls back to the
  raw localization key while the strings are still resolving" — is finding 22: that state is
  permanent.
- scope-selector.utils.ts:145 — the all-books check counts unrecognized IDs but
  localizeBookIdsInCanonOrder drops them, so [...available, 'ZZZ'] renders GEN - REV rather
  than "All books". Conservative and defensible; noted only because the two paths now normalize
  differently.
- PR body is stale. It still describes the tooltip as added, lists formatSelectedBooksList as
  a new export, and says getAvailableBookIds landed on the root export with the key in
  SELECT_BOOKS_STRING_KEYS → SCOPE_SELECTOR_STRING_KEYS. All four are now wrong, and this text
  becomes the squash-merge message.

On the two you flagged for pushback

Reverting scope-selector.component.tsx wholesale — agreed, and a better answer to item 12 than
what I suggested. One consequence to state somewhere: the library's variant="dropdown" +
selectedBooks trigger goes back to joining every ID, so the overflow bug stays latent for a future
consumer. Unreachable today, and summarizeSelectedBooks is exported for whoever needs it — fine as
a known deferral rather than an accident.

The - separator — that was my item 21e, so no pushback. It does add a fifth delta from PT9 /
the ticket's literal GEN … HOS, alongside the four already listed in the PR body. Worth one line
for the UX owner of PT-3363.
```


> @katherinejensen00 reviewed 14 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2708#-:-P-ovcGk0f4Oa-nO4awI:ba6g3l4)* status: all files reviewed, 1 unresolved discussion (waiting on jolierabideau).

___
*[`extensions/src/platform-scripture/src/find.web-view.tsx` line 554 at r1](https://reviewable.io/reviews/paranext/paranext-core/2708#-P-ovrDuAI0LzdpBVR85:-P-ovrDuAI0LzdpBVR86:b-91yrrp) ([raw file](https://github.com/paranext/paranext-core/blob/01a50dc5232b44059ed53715bbdf3e4b3901f9ac/extensions/src/platform-scripture/src/find.web-view.tsx#L554)):*
> ```XML
>     activeMode === 'replace' ? projectId : undefined,
>   );
> ```

**22. The "All books" string is never fetched, so the trigger renders the raw key** ✅
`extensions/src/platform-scripture/src/find.web-view.tsx:554` — **High**

Reverting `scope-selector.component.tsx` to `main` took `%webView_scope_selector_all_books%` out of
`SCOPE_SELECTOR_STRING_KEYS` (it was at `:61` in `8b0326b`), and nothing added it to
`FIND_LOCALIZED_STRING_KEYS` (`find.component.tsx:81`) to compensate. `find.web-view.tsx:554-558`
requests exactly `SCOPE_SELECTOR_STRING_KEYS`, so the read at `find.component.tsx:635` is
`undefined` **always** — not just during the localization round-trip — and the fallback at `:636`
becomes the shipped string.

`grep -rn all_books lib/platform-bible-react/src extensions/src/platform-scripture/src` returns
three hits: the read, its fallback, one test. Zero key lists. The bundle agrees —
`webView_scope_selector_all_books` appears **0 times** in `dist/index.js`.

Item 18's `fallbackKey` is dead code until this is fixed: a key that is never requested never falls
back, so `"Todos los libros"` is never reached either.

The tests can't catch it. `buildBooksScopeProps` (`find.component.test.tsx:641`) *replaces*
`scopeSelectorLocalizedStrings` with a hand-built object containing the key, so every books-scope
test runs against a props shape `find.web-view.tsx` cannot produce. `buildProps:100` uses the real
production set; spreading it would have failed.

Since `scope-selector.component.tsx` is back to zero changes, this string is now Find-owned living
under a ScopeSelector name. `%webView_find_allBooks%` in `FIND_LOCALIZED_STRING_KEYS`, read from
`localizedStrings`, keeps the revert intact and puts the key with its only consumer.

katherinejensen00 (changes requested):
Oops. That last comment was for a different PR

Here are the comments for this PR:
Line numbers are against head e94dc15cbd0. lib/platform-bible-react/dist/** was skipped as generated output.

Overall: the core design is right and the predicate choice is correct — Canon.isExtraMaterial covers exactly the 15 books that need excluding, and clearing flags in place rather than dropping positions is the non-obvious correct call. What needs attention is that the prunePresentBookIds signature change covered the loading branch but not the error branch, and that this PR and #2699 are independently building the same helper in the same week.

Must fix

1. A transient booksPresent read error permanently destroys the user's saved book selection.
The new guard trusts isBooksPresentLoading alone, but create-use-data-hook.util.ts:199-200 calls setIsLoading(false) for any delivered value — a PlatformError included. So on the error branch isBooksPresentLoading is false while line 680 substitutes BOOKS_PRESENT_DEFAULT, making availableBooksIds []. prunePresentBookIds([], sel) now returns [] and setSelectedBookIds([]) writes to useWebViewState — persisted. It does not come back when the error clears, and because isFindQueryValid (find.utils.ts:190) rejects an empty selectedBooks scope, Find is also disabled with no explanation. The old .length === 0 check covered exactly this. Derive isBookListKnown = !isBooksPresentLoading && !isPlatformError(booksPresentPossiblyError) instead.
→ extensions/src/platform-scripture/src/find.web-view.tsx:741 · High

2. getPresentBookIds is byte-for-byte identical to the getAvailableBookIds that PR #2699 is shipping into the shared library. Both are getBookIdsFromBooksPresent(x).filter((id) => !Canon.isObsolete(Canon.bookIdToNumber(id))). #2699 puts that in scope-selector.utils.ts; this PR puts a private copy in extension code. Code-Style-Guide §Shared Utilities and ADR-0021 ("platform-bible-utils owns the wire format") both point the other way.
→ extensions/src/platform-scripture/src/find/find.utils.ts:369 · High

3. Direct textual conflict with PR #2699 in the picker. #2699 adds className="tw:max-h-72 tw:min-h-0 tw:flex-1" to <CommandList> on the line immediately above the block this PR rewrote, and keeps the old index < Object.values(Section).length - 1 separator. Both PRs also touch find.component.tsx and localizedStrings.json. Whoever merges second reconciles by hand — and this PR's separator fix is the one that must survive.
→ lib/platform-bible-react/src/components/advanced/scope-selecnent.tsx:221 · High
  Searched for 3 patterns

The /code-review max run has fanned out into its own subagentser agents done, verification passes running). I'll wait for itand deliver the consolidated report once it lands.

Interim status — the antagonistic pass substantially reshaped my findings:

┌───────────────────────────────────────────────────────────────────────────────────────────────┬─────────────────────────────────────────────────────┐
│                                            Held up               │                    Knocked down                     │
├───────────────────────────────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────┤
│ S1 (regression, downgraded High→Medium) · S2 (strongest; violates the repo's own Revert Test  │ S5, S6, A2, A5, A6, D1, D3, D8 — refuted with       │
│ rule) · D4, D5 (trivia)                                                                       │ concrete evidence                                   │
└───────────────────────────────────────────────────────────────────────────────────────────────┴─────────────────────────────────────────────────────┘

The most useful corrections against me:
- My S5 was wrong in the right direction. The '' flag string that makes getAvailableBookIds throw is reachable during normal loading, not just on error — but the render gate is three conditions, not one. Reframed as a real Medium finding.
- D1/D3/D8 were length-heuristic nitpicks. The repo's forward-facing-comments.md explicitly lists "a pointer to where deferred work lives — see PT-XXXX" as keep; cutting those 6 lines would delete the only in-code link to PT-4415. And my punctuation claim was flatly wrong — two Find strings 2 lines away use contractions and terminal periods.
- The separator rewrite is correct, not a risk: n non-empty sections → exactly n-1 interior separators. The old code was the buggy one.
- Four defects I missed, including that deriveFindBookLists's own TSDoc claims all book-list policy moved out of the web view — while the error branch, which is exactly where S1 goes wrong, stayed behind untested.

✻ Churned for 47m 46s

● Agent "/code-review max https://github.com/paranext/para…" fin

The /code-review max run returned 30 findings. Several are higverify the load-bearing ones before I put my name on them.

  Ran 3 shell commands
Loaded ..\..\.claude\rules\architecture\discover-before-implementing.md
Loaded ..\..\.claude\rules\architecture\extension-patterns.md
Loaded ..\..\.claude\rules\architecture\porting-discipline.

The repo checkout was switched out from under me by another concurrent session — I'll verify the rest against refs without touching the working tree.

  Ran 4 shell commands

This is a significant verified finding. Let me check the picker overlap.

  Searched for 2 patterns, ran 9 shell commands

All major claims are now verified. One correction to the code-review's output: its finding #3 anchored at line 674, which is inside a localized-strings block — the real anchor is find.component.tsx:662.

---

Review: PR #2708 — PT-3299: Exclude extra-scriptural books from Find

Line numbers are against head e94dc15cbd0. lib/platform-bible-react/dist/** was skipped as generated output.

Overall: the core design is right and the predicate choice is al covers exactly the 15 books that need excluding, andPlatformError included. So on the error branch isBooksPresentLoading is false while line 680 substitutes BOOKS_PRESENT_DEFAULT, making availableBooksIds
[]. prunePresentBookIds([], sel) now returns [] and setSelecteViewState — persisted. It does not come back when the errorclears, and because isFindQueryValid (find.utils.ts:190) rejects an empty selectedBooks scope, Find is also disabled with no explanation. The old .length === 0 check covered exactly this. Derive isBookListKnown = !isBooksPresentLoading && !isPlatformError(booksPresentPossiblyError) instead.
→ extensions/src/platform-scripture/src/find.web-view.tsx:741 · High

2. getPresentBookIds is byte-for-byte identical to the getAvai shipping into the shared library. Both are


3. Direct textual conflict with PR #2699 in the picker. #2699 adds className="tw:max-h-72 tw:min-h-0 tw:flex-1" to <CommandList> on the line immediately above the block this PR rewrote, and keeps the old index < Object.values(Section).length - 1 separator. Both PRs also touch find.component.tsx and localizedStrings.json. Whoever merges second reconciles by hand — and this PR's separator fix is the one that must survive.
→ lib/platform-bible-react/src/components/advanced/scope-selector/select-books-picker.component.tsx:221 · High

4. The "drops obsolete books" test can never fail. 4ES is not a canon id — Canon.bookIdToNumber('4ES') returns 0, so booksPresentFor writes flags[-1], which join('') discards. I mutation-tested it: deleting the Canon.isObsolete filter from getPresentBookIds leaves all 108 tests green, so the one behavior carried over from the deleted memo has zero coverage. This is the Revert Test in Testing-Guide.md:295. Real obsolete ids: JSA JDB TBS SST DNT BLT 3ES.
→ extensions/src/platform-scripture/src/find/find.utils.test.ts:756 · High

5. The booksPresentFor helper silently no-ops on an unrecognized book id. Root cause of #4, and it will hollow out the next fixture typo the same way. Throw when bookIdToNumber(bookId) <= 0.
→ extensions/src/platform-scripture/src/find/find.utils.test.ts:686 · High

Should fix

6. The new platform-bible-utils/experimental import pulls a dul into the extension-host bundle. webpack.config.base.ts:81externalizes 'platform-bible-utils' by exact string; '…/experimental' is not in that list, so it gets bundled. find.utils.ts is backend-reachable
(platform-scripture-finder-pdpe.model.ts:31 ← main.ts:43). Eve in extensions/src was in a .web-view.tsx. The backend also now runs 123 Canon calls at activation for a Set it never reads.
→ extensions/src/platform-scripture/src/find/find.utils.ts:12 · Med-High

7. findScope is never filtered, so the exclusion rests entirely on the prune — which is deliberately withheld while loading. A pre-existing persisted
['GEN','GLO'] plus the startup auto-search covers GLO in that non.isExtraMaterial(id)) here would make it hold

8. Stale results stay replaceable after the selection empties. isReplaceUnavailable omits isSearchQueryValid, and the buttons at 963/970 gate only on visibleResults.length === 0 || isReplaceUnavailable. Replace All can then write against stale character offsets.
→ extensions/src/platform-scripture/src/find/find.component.tsx:662 · Medium

9. The invalid-query prompt is unreachable when stale results are on screen, because if (results.length > 0) return 'none' precedes it — so the emptied-selection state dead-ends with no message.
→ extensions/src/platform-scripture/src/find/find.component.tsx:597 · Medium

10. searchableBooksPresent can be '', and getAvailableBookIds throws on it during render, tearing down the panel (reachable with scope === 'selectedBooks' while loading, or permanently after a read error). Pre-existing, but deriveFindBookLists is the cheapest fix site it will ever have. Note #2699 fixes this by degrading to [] — which also makes the test comment at line 713 stale.
→ extensions/src/platform-scripture/src/find/find.utils.ts:404 · Medium

11. The book/chapter bypass sits on the default path — useWebViewState<Scope>('findScope', 'book') makes book the default scope, and the nav control's list is unfiltered, so the headline GLO 1:0 symptom still reproduces. Acknowledged in-comment and deferred to PT-4415; worth an explicit sign-off that shipping the fix without it is the intent.
→ extensions/src/platform-scripture/src/find.web-view.tsx:1029 · Medium

12. The string asserts an absolute that PT-4415 will falsify. Localized values are immutable once shipped, so correcting "Find can't search extra
material" later needs a new key plus fallbackKey. Scope-relatican't include extra material") would survive.
→ extensions/src/platform-scripture/contributions/localizedStrings.json:442 · Medium

13. New stable public API with zero coverage. ScopeSelector and SelectBooks are both in index.ts. There's no section-button test, the existing LimitedAvailableBooks story is untouched, and find.component.test.tsx:117 passes '1'.repeat(123), which leaves Extra enabled — so the new branch never
executes in any test. Also missing className: the sibling usagasses one specifically to keep the button row intact.
onent.tsx:66 · Medium

14. Conditional wrapping swaps the element type at a fixed position, so an isDisabled flip remounts the Button and drops keyboard focus. All four existing DisabledActionTooltip consumers wrap unconditionally and pass disabled={expr} — the wrapper is already inert when enabled.
→ lib/platform-bible-react/src/components/advanced/scope-selector/section-button.component.tsx:63 · Med-Low

15. The explanation reaches the section button but not the picker, which silently omits the Extra group — searching "glossary" lands on the bare "no book found".
→ lib/platform-bible-react/src/components/advanced/scope-selector/select-books.component.tsx:98 · Med-Low

16. The tooltip is supplied unconditionally, so a project with no extra material is told Find "can't search" it. The web view already knows the difference via localizableBookIds.
→ extensions/src/platform-scripture/src/find/find.component.tsx:618 · Med-Low

17. localizedStrings: LanguageStrings is an open index signature, so the new key gets no type safety — rename it and the tooltip vanishes with no compile error. Contrast searchResultLocalizedStrings (line 161), a mapped type over its key list.
→ extensions/src/platform-scripture/src/find/find.component.tsx:154 · Med-Low

18. The booksPresent prop doc states no "already filtered" premment asserts one — and the PR's own test fixture violates it.A second host passing raw booksPresent disables the exclusion silently.
→ extensions/src/platform-scripture/src/find/find.component.tsA second host passing raw booksPresent disables the exclusion silently.
→ extensions/src/platform-scripture/src/find/find.component.tsx:227 · Med-Low

Minor

19. ALL_EXTRA_MATERIAL_BOOK_IDS is documented as "an independent oracle" but is Canon.nonCanonicalIds verbatim — I compared them: same 15 ids, same order. The claim invites a reviewer to trust cross-check coverage that isn't there.
→ find.utils.test.ts:683 · Low

20. EXTRA_MATERIAL_BOOK_NUMBERS reinvents Canon.nonCanonicalIds, and its 3-line out-of-canon justification guards nothing — Canon.isExtraMaterial(999) is already false, and there's a number overload. new Set(Canon.nonCanonicalIds.map(Canon.bookIdToNumber)) replaces the whole block.
→ find.utils.ts:326 · Low

21. The comment says availableBooksIds is "what the search runs over." It isn't — its only consumer is the prune effect (uses: 690, 741, 745). A reader trusting this would think dropping the prune is safe.
→ find.web-view.tsx:667 · Low

22. deriveFindBookLists's TSDoc claims all book-list policy moved out of the web view, but the error policy stayed behind — and that's precisely the branch finding #1 gets wrong. Moving the unknown/error decision in would make the doc true and fix #1 at the same time.
→ find.utils.ts:396 · Low

23. localizableBookIds may buy nothing observable. %LocalizedIhe backup language, and it's empty on the error path anyway —i.e. useless exactly where it was added. Worth confirming it earns its third field, TODO, and two comment blocks.
→ find.utils.ts:409 · Low

24. The error branch mints fresh arrays per re-render, producia full unsubscribe/resubscribe round trip. Hoist to a moduleconstant.
→ find.web-view.tsx:680 · Low

25. Tests pin non-canon-length input as supported while the only consumer requires exactly 123 — documenting robustness the pipeline doesn't have.
→ find.utils.test.ts:733 · Low

26. Vocabulary is still mixed — 6 of 9 new test names say "extra-scriptural" against 18 uses of "extra material". The PR body says the alias was kept "once"; it's kept nine times.
→ find.web-view.tsx:666 · Low

27. The PR description says the extra-material books are "numbers 93-111." Verified: 93-102 and 107-111. 103-106 (3ES, EZA, 5EZ, 6EZ) are not extra material. Description only, not code.
→ PR body · Low

28. No Architecture-Decisions.md entry, though CLAUDE.md names "deferring a platform capability" as a trigger and the log already carries ADR-0020/0021 on this exact surface.
→ .context/standards/Architecture-Decisions.md · Low

29. Supplying the prop adds a permanent tab stop. Because Find's Extra button is always disabled, the scope popover now has a non-actionable role="group" stop on every project. That's the primitive's documented design, not a bug — but it's a standing cost the PR body's cost/benefit item doesn't name. Confirm-intended.
→ section-button.component.tsx:66 · Confirm-intended

30. Worth confirming (not verified): explicitSearchPendingRef is armed before startSearch, and the gate-skip path may return without clearing it — now a routine outcome since project switches commonly prune to empty. If so, the user's next auto-search is swallowed.
→ extensions/src/platform-scripture/src/find/use-find-search-triggers.hook.ts:80 · Med-Low, unverified

Things that came up and were checked clean

So these don't get re-litigated: the separator rewrite is correct (n non-empty sections → exactly n-1 interior separators; the old code was the buggy one). Section is a string enum, so Object.values is safe. PT-4explicitly permitted by the forward-facing-comments rule, andboth tickets exist. The string's contraction and terminal period match two Find strings two lines away. dist/index.d.ts matches the source prop additions exactly, and dist is tracked here by convention. eslint and prettier are clean on all 10 files; all 1197 platform-scripture tests and the 3 scope-selector suites pass.

I'd also note for the author: cmdk's Separator returns null whenever the search box is non-empty unless alwaysRender is set, and this picker sets
shouldFilter={false} — the exact case cmdk documents alwaysRenjacent to the rewrite.

> @katherinejensen00 made 2 comments and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2708#-:-P-owMHhDOS-zD91y6vk:bfo61oh)* status: all files reviewed, 1 unresolved discussion (waiting on jolierabideau).

___
*[`lib/platform-bible-react/src/components/advanced/scope-selector/select-books-picker.component.tsx` line 221 at r1](https://reviewable.io/reviews/paranext/paranext-core/2708#-P-owk7C8so-0fvtGk4Q:-P-owk7C8so-0fvtGk4R:b-o8k2cj) ([raw file](https://github.com/paranext/paranext-core/blob/01a50dc5232b44059ed53715bbdf3e4b3901f9ac/lib/platform-bible-react/src/components/advanced/scope-selector/select-books-picker.component.tsx#L221)):*
> ```XML
>           <CommandList>
>             <CommandEmpty>{noBookFoundText}</CommandEmpty>
>             {Object.values(Section)
> ```

3. Direct textual conflict with PR #2699 in the picker. #2699 adds className="tw:max-h-72 tw:min-h-0 tw:flex-1" to <CommandList> on the line immediately above the block this PR rewrote, and keeps the old index < Object.values(Section).length - 1 separator. Both PRs also touch find.component.tsx and localizedStrings.json. Whoever merges second reconciles by hand — and this PR's separator fix is the one that must survive.
→ lib/platform-bible-react/src/components/advanced/scope-selector/select-books-picker.component.tsx:221 · High

jolierabideau:
Thanks — this was a genuinely useful pass. All 30 findings are answered below: **26 fixed, 4
deferred**. `origin/main` is merged (through PT-4339, #2682) and the PR reports `MERGEABLE` again.

Head is `1e748d69e69`. Verified before pushing: `npm run typecheck` clean, `npm run lint` **0 errors**
repo-wide, `npm test` **5,253 passing across 11 workspaces**, `lib/platform-bible-react/dist/`
regenerated from source.

Two of your findings I ran the revert test against, since both fixes are easy to write and not
actually guard anything:

| Fix | Revert test |
|---|---|
| 14 (`SectionButton` remount) | reverting to the `isDisabled`-keyed wrap fails the new "keeps the same Button element across a disabled/enabled flip" test |
| 8 + 9 (unrunnable query) | reverting either fails its new test in `find.component.test.tsx` |

---

### Must fix

**1. Transient `booksPresent` error destroys the saved selection — fixed.** Confirmed exactly as you
described: `useProjectSetting` reports a delivered `PlatformError` as loaded, so the error branch had
`isBooksPresentLoading === false` with `availableBooksIds` empty, and the prune wrote `[]` to
`useWebViewState`.

I took your finding 22 as the fix site rather than deriving `isBookListKnown` in the web view — moving
the unknown/error decision *into* `deriveFindBookLists` makes its TSDoc true and closes this at the
same time, which is what you pointed out. `deriveFindBookLists(undefined)` now means "not known" and
returns `availableBookIds: undefined`; the web view passes `undefined` for **both** the loading branch
and the `isPlatformError` branch. An empty array stays a real answer, since a project holding only
extra material genuinely has nothing to search.

**2 + 6 + 20. The duplicate decoder, the backend bundle, and the reinvented set — fixed together.**
These turned out to be one problem. Using the shared `getAvailableBookIds` (which the merge brings in
from #2699) means importing `platform-bible-react` — worse for the backend bundle than the
`platform-bible-utils/experimental` import you flagged, since `find.utils.ts` is reachable from
`platform-scripture-finder-pdpe.model.ts`.

So the book-list policy moved to a new **web-view-only** module,
`extensions/src/platform-scripture/src/find/find-book-lists.utils.ts`:

- `getPresentBookIds` is deleted; the module calls the shared `getAvailableBookIds` (2 — one decoder).
- `find.utils.ts` loses its `platform-bible-utils/experimental` import *and* its `Canon` import
  entirely, so nothing new reaches the extension-host bundle and the 123 activation-time `Canon` calls
  are gone (6).
- `EXTRA_MATERIAL_BOOK_NUMBERS` is now `new Set(Canon.nonCanonicalIds.map(Canon.bookIdToNumber))` (20).
  I verified the equivalence at runtime rather than by reading: `Canon.allBookIds.filter(Canon.isExtraMaterial)`
  and `Canon.nonCanonicalIds` are the same 15 ids in the same order.

**3. Textual conflict with #2699 — resolved by the merge, and your call held.** Both survived:
`<CommandList className="tw:max-h-72 tw:min-h-0 tw:flex-1">` from #2699, and this branch's
`index > 0 && <CommandSeparator />` over the filtered sections. Git auto-merged that hunk; I resolved
`find.web-view.tsx` (dropped main's local `availableBooksIds` memo — `deriveFindBookLists` supplies it)
and `find.component.tsx` (kept both sides) by hand.

**4 + 5. The obsolete-books test could never fail — fixed, and you were right about the root cause.**
Confirmed `Canon.bookIdToNumber('4ES')` returns `0`. Both fixed:

- `booksPresentFor` now throws on a non-canon id, so the next fixture typo fails loudly (5).
- The test uses real obsolete ids — `JSA` (87) and `3ES` — and asserts both `availableBookIds` and
  `localizableBookIds` (4). Deleting the obsolete filter now fails it.

### Should fix

**7. `findScope` unfiltered — deferred to PT-4415.** Deliberate, and I'd rather not half-fix it: adding
a filter here leaves two partial gates in different layers for PT-4415 to reconcile, when the real fix
is gating the scopes themselves. Recorded as an explicit alternative-considered in ADR-0024 so it isn't
re-derived. The prune's error-branch hole (finding 1) is closed, which was the reachable half.

**8 + 9. Stale results stay replaceable / the prompt is unreachable — both fixed.** `isReplaceUnavailable`
now includes `!isSearchQueryValid`, and the `invalidQueryPrompt` branch moved *above* the
`results.length > 0` early return, so an unrunnable query outranks the results left over from the last
one. Three new tests in `find.component.test.tsx` cover the emptied-selection state.

**10. `getAvailableBookIds` throwing on `''` — resolved by the merge.** #2699's version degrades to `[]`
(ADR-0023). The stale test comment you flagged went with the tests that moved.

**11. Book/chapter bypass — deferred to PT-4415, explicitly.** Confirming the sign-off you asked for:
shipping the exclusion without the scope gating is the intent. PT-3299 chose the low-effort, easily
reversed option precisely to avoid rebuilding the "handle extra books properly" work; gating the scopes
starts down that path. ADR-0024 records it as a decision rather than an omission.

**12. The string asserts an absolute — fixed.** Now scope-relative, so neither PT-4414 nor PT-4415
falsifies a shipped (immutable) value:

> Find doesn't include extra material, such as glossaries and front matter, in this book list.

**13 + 14 + 15. The library side — all fixed, with coverage.**

- 14 was the real bug. The wrap is now keyed on whether an explanation was *supplied* (a stable prop),
  not on `isDisabled`, so the element type no longer swaps mid-interaction. `disabled={isDisabled}` is
  passed through, matching the four existing consumers.
- 13: new `section-button.component.test.tsx` (3 tests), plus the `tw:flex` className so the wrapper
  doesn't break the button row it becomes a flex item of.
- 15: `disabledSectionExplanations` now threads into `SelectBooksPicker`, which renders the explanation
  as a note under the list for any section with no books. Deliberately *outside* `CommandList` — inside
  it, cmdk would filter away the note on the very search ("glossary") that makes the omission visible.
  Two new picker tests.

**16. Tooltip shown to projects with no extra material — fixed.** New `hasExcludedExtraMaterial` prop,
derived in the web view from `localizableBookIds.length > availableBookIds.length` (the two lists differ
only by extra material). The explanation is `undefined` otherwise.

**17. No type safety on the key — fixed** with a narrow binding rather than converting `localizedStrings`
to a mapped type, which would have made every read in the component `| undefined`:

```ts
const EXTRA_MATERIAL_NOT_SEARCHED_KEY =
  '%webView_find_extraMaterialNotSearched%' satisfies (typeof FIND_LOCALIZED_STRING_KEYS)[number];
```

Renaming the key in the list is now a compile error at the read site.

**18. The `booksPresent` prop doc — fixed.** It now states the already-filtered premise the comment
asserted, and names `deriveFindBookLists` as how callers produce it.

### Minor

**19.** Fixed — the doc no longer claims independence. It is `Canon.nonCanonicalIds` verbatim; it now
says so, and says what it *does* buy (pinning the expected set against a canon change).
**21.** Fixed — the comment described `availableBooksIds` as what the search runs over; it now says what
it is actually for.
**22.** Fixed as part of 1 — the error policy moved in, so the TSDoc is true.
**23.** Kept, with the reasoning now written down. `localizableBookIds` is what stops the `book`/`chapter`
scope label degrading to a raw `GLO` when the current reference sits in extra material — the one
observable difference, and it's on the path finding 11 defers. It is empty on the unknown path, which is
correct: there is nothing to label yet.
**24.** Fixed — `UNKNOWN_FIND_BOOK_LISTS` is a module constant, so the unknown state hands effects and
`useLocalizedStrings` stable array identities instead of fresh ones per render.
**25.** Fixed — the non-canon-length cases are dropped. The `BOOKS_PRESENT_DEFAULT` case remains, since
that one is genuinely reachable.
**26.** Fixed — `grep -rn "extra-scriptural"` over `extensions/src` and `lib/platform-bible-react/src`
now returns nothing, test names included.
**27.** Fixed in the PR body — verified your correction: 93-102 and 107-111; 103-106 (`3ES`, `EZA`,
`5EZ`, `6EZ`) are not extra material.
**28.** Fixed — **ADR-0024** added, covering the narrow-the-lists-not-the-scopes decision, the
clear-in-place invariant, the two rejected alternatives (including your finding 7), and the
unknown-vs-empty consequence.

**29. The permanent tab stop — confirm-intended, and now narrower than when you raised it.** With
finding 16 fixed, the stop only exists on projects that actually have extra material, where the
explanation is the point of it. Accepting the cost.

**30. `explicitSearchPendingRef` — verified real, filed separately.** Your read was right, with one
correction to where it lives: `handleStartSearch` is fine — it sets the flag at `find.web-view.tsx:1135`,
*after* the gate's skip-return. The leak is in the hook: `use-find-search-triggers.hook.ts:80` and `:103`
arm the flag and then call `startSearch(false)`, so a gate skip returns with the flag still set and the
next search consumes it.

Filing rather than fixing here: the file is outside this diff and the bug is pre-existing. Worth noting
that finding 1's fix removes the reachability you were worried about — project switches no longer prune
to empty on a read error.

### On your closing notes

- **The latent `variant="dropdown"` overflow** — carried into ADR-0023's deferral note by #2699, which is
  now on `main`, so it is recorded where a future dropdown consumer will look.
- **The ` - ` separator as a fifth PT9 delta** — also recorded in ADR-0023 for PT-3363's UX owner.
- **cmdk's `Separator` returning `null` when the input is non-empty** — good catch, and it doesn't bite
  here: with `shouldFilter={false}` this picker does its own filtering and renders `CommandSeparator`
  from the filtered section list, so the separators are driven by what actually rendered rather than by
  cmdk's internal match state. Worth keeping in mind if anyone turns `shouldFilter` back on.

jolierabideau:
Correction to my last comment — my dismissal of your cmdk note was wrong, and it was a real bug.

I claimed the `shouldFilter={false}` + own-filtering setup meant cmdk's `Separator` behavior didn't
apply here. I should have measured instead of reasoned. Probing the rendered DOM:

| picker state | `[cmdk-group]` | `[data-slot="command-separator"]` |
|---|---|---|
| popover open, empty search | 2 | **1** ✅ |
| same, after typing a cross-section query | 2 | **0** ❌ |

cmdk hides the separator on a non-empty search box regardless of who did the filtering — so the
separator fix this PR is built around silently stopped applying the moment the user typed, which is
most of the time the picker is open. `alwaysRender` is exactly the case cmdk documents, as you said.

Fixed in `29a2f497a5a`: `{index > 0 && <CommandSeparator alwaysRender />}`, with the reasoning at the
call site. Two tests cover it — one interior separator across two rendered sections both before and
during a search, and none when only one section has books. Revert-tested: dropping `alwaysRender`
fails the first one.

Everything re-verified on the new head: typecheck clean, `npm run lint` 0 errors, `npm test`
5,255 passing, `dist/` rebuilt.

So the tally is **27 fixed, 4 deferred** — thanks for the catch, and for putting it in writing even
though it read as an aside.

katherinejensen00 (approved):
Thanks for your hard work on this , Jolie
:lgtm:

> @katherinejensen00 reviewed 20 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2708#-:-P-qGo9s2KGH3sUvE1f2:b2mm97q)* status: all files reviewed, 1 unresolved discussion (waiting on jolierabideau).

jolierabideau:
Re-synced with `main` — now at `777e0b672de`, which includes PT-4348 (#2687). Still `MERGEABLE`.

One thing worth flagging, since it changes an identifier I cited earlier: **the ADR I added is now
ADR-0025, not ADR-0024.** PT-4348 landed its own ADR-0024 ("The toolbar's sync status is local
renderer UI…") first, so both sides appended the same number at EOF — the duplicate-numbering case
you called out in your last round. Main's entry keeps 0024; the Find entry follows it as ADR-0025.
The two `ADR-0024` references in the new `sync-status-button.component.tsx` / `use-sync-status.hook.ts`
point at PT-4348's own entry and are correct as they stand.

That log was the only conflict — PT-4348 is toolbar/sync only, with no overlap with Find or the scope
selector, and it touches no `dist/`.

Re-verified on the merged tree: `npm run typecheck` clean, `npm run lint` 0 errors, `npm test`
**5,314 passing** (up from 5,255 — the difference is PT-4348's own new suites).

PR #2709, opened 2026-08-24
PT-4368: Guard DBL install verification against null InstalledScrText

katherinejensen00:
## Summary

A failed DBL resource install threw `NullReferenceException` out of the very check that exists to
report it, so **Get Resources** never showed its own "installation failed" message.

`InstallDblResourceCore` verified an install with
`ScrTextCollection.IsPresent(installableResource.InstalledScrText)`. ParatextData assigns
`InstalledScrText` only once `InstallableResource.InternalInstall` completes, so it stays `null`
whenever the install bailed out early (e.g. the DBL download failed) — and
`ScrTextCollection.IsPresentInternal(ScrText)` dereferences `scrText.Guid`. The localized
`%getResources_errorInstallResource_installationFailed%` message was therefore unreachable: the
failure path always faulted with an NRE first.

This null-checks `InstalledScrText` before the presence check, so the failure surfaces with its
intended message.

### Why review this

Not part of the current epic — *please correct this if PT-4368 is being counted as epic work.*

This came out of investigating **PT-4368** ("first-time sync silently skips connected resources").
The bulk of that bug lives in Paratext 10 Studio's `SyncProjectsCore` and is being fixed separately
in the patch repo; this is the one piece that belongs in public core, and it stands on its own. It is
worth reviewer time now because it is the *reporting* path for a resource install failure — while it
is broken, any DBL install failure is misdiagnosed as a program error rather than shown to the user,
which makes the Studio-side work harder to verify.

## Changes

- `c-sharp/Projects/DigitalBibleLibrary/DblDownloadableDataProvider.cs` — null-check
  `InstalledScrText` before `ScrTextCollection.IsPresent`.
- Replaced the comment "we don't get any info telling if the installation succeeded or failed",
  which is not accurate, with a note on what `Install()`'s return value actually means (below).

### The one thing worth your judgment

The obvious-looking alternative fix is to test `Install()`'s return value. **That would be a bug**,
and the new comment exists to stop someone doing it later.
`InstallableResource.Install()` → `InternalInstall()` ends with `return fontsFound;`, XML-documented
as *"true if any fonts were copied to the destination folder"*. So `false` is the **normal** return
for a *successful* install of a resource that ships no fonts — i.e. most resources. Presence in the
collection is the only usable success signal, which is what this method already (correctly) chose to
use.

`InstallableDBLResource.Install()` does return `false` early when the download itself fails, which is
what makes the bool look like a success flag at a glance — but it otherwise returns `base.Install()`,
so its `false` is ambiguous between "download failed" and "installed, no fonts".

## AI Involvement

AI-assisted. The change, its comments, and this description were drafted by Claude Code (Claude Opus
5) and reviewed by me. The commit carries a `Co-Authored-By` trailer attributing it as a generator;
I am the author.

Specifically AI-generated: the 13-line diff and its two comments. The claims about ParatextData's
behavior were verified directly against the Paratext 9 source tree
(`ParatextData/Archiving/InstallableResource.cs`, `ParatextData/ScrTextCollection.cs`) rather than
inferred — the `fontsFound` return, the single `InstalledScrText` assignment site, and the
`scrText.Guid` dereference in `IsPresentInternal` are each read from that source. Reviewers who want
to spot-check should look there.

## Testing

- [x] `dotnet build` — succeeded, 0 errors. The 4 warnings are pre-existing `CS8625`s in
      `ParatextProjectDataProviderCommentTests.cs`, untouched here.
- [x] `dotnet test c-sharp-tests/` — 1716 passed, 0 failed, 7 pre-existing skips (re-run after rebasing onto `main`).
- [x] `dotnet csharpier --check` — clean on the changed file.
- [ ] Not manually verified in the app: reaching this branch requires a genuine DBL install failure
      (a download that fails mid-flight, or a resource whose zip fails `CheckResource`), which I
      could not stage reliably. The change is a null guard ahead of an existing check, so the
      success path is unaffected by construction.

No test was added. The method is not currently covered — there is no
`c-sharp-tests/Projects/DigitalBibleLibrary/` — and exercising it needs `InstallableDBLResource` plus
live DBL access, so a unit test here would only be able to assert against a mock of the very
ParatextData behavior at issue. Happy to add one if a reviewer sees a seam I missed.

## Risk Level

Low — one file, +13/-2, C# only. Adds a null check ahead of an existing condition; the
already-installed and success paths are unchanged. Worst case, a failure that used to throw NRE now
throws the localized message instead.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2709)

### DIFF
--- a/c-sharp/Projects/DigitalBibleLibrary/DblDownloadableDataProvider.cs
+++ b/c-sharp/Projects/DigitalBibleLibrary/DblDownloadableDataProvider.cs
@@ -1,4 +1,4 @@
-using System.Diagnostics;
+﻿using System.Diagnostics;
 using System.Text.Json;
 using Paranext.DataProvider.ParatextUtils;
 using Paranext.DataProvider.Services;
@@ -285,12 +285,23 @@ internal class DblResourcesDataProvider(
                 )
             );
 
-        // Note that we don't get any info telling if the installation succeeded or failed
+        // Install()'s return value is NOT a success flag: InstallableResource.InternalInstall
+        // returns `fontsFound` ("true if any fonts were copied"), so false is the normal return for
+        // a successful install of a resource that ships no fonts. Presence in the collection is the
+        // only usable signal, which is what we check below.
         installableResource.Install();
 
         ScrTextCollection.RefreshScrTexts();
 
-        if (!ScrTextCollection.IsPresent(installableResource.InstalledScrText))
+        // InstalledScrText is assigned only once InternalInstall completes, so it is still null when
+        // the install bailed out early (e.g. the DBL download failed). ScrTextCollection.IsPresent
+        // dereferences its argument, so a null check must come first — otherwise the very failure
+        // this guard exists to report surfaces as a NullReferenceException instead of the localized
+        // message below.
+        if (
+            installableResource.InstalledScrText == null
+            || !ScrTextCollection.IsPresent(installableResource.InstalledScrText)
+        )
             throw new Exception(
                 LocalizationService.GetLocalizedString(
                     PapiClient,
--- a/c-sharp/Projects/DigitalBibleLibrary/DblDownloadableDataProvider.cs
+++ b/c-sharp/Projects/DigitalBibleLibrary/DblDownloadableDataProvider.cs
@@ -1,4 +1,4 @@
-﻿using System.Diagnostics;
+using System.Diagnostics;
 using System.Text.Json;
 using Paranext.DataProvider.ParatextUtils;
 using Paranext.DataProvider.Services;
@@ -293,11 +293,11 @@ internal class DblResourcesDataProvider(
 
         ScrTextCollection.RefreshScrTexts();
 
-        // InstalledScrText is assigned only once InternalInstall completes, so it is still null when
-        // the install bailed out early (e.g. the DBL download failed). ScrTextCollection.IsPresent
-        // dereferences its argument, so a null check must come first — otherwise the very failure
-        // this guard exists to report surfaces as a NullReferenceException instead of the localized
-        // message below.
+        // InstalledScrText is assigned only once InternalInstall completes, so it is still null
+        // when the install bailed out early (e.g. the DBL download failed).
+        // ScrTextCollection.IsPresent dereferences its argument, so a null check must come first —
+        // otherwise the very failure this guard exists to report surfaces as a
+        // NullReferenceException instead of the localized message below.
         if (
             installableResource.InstalledScrText == null
             || !ScrTextCollection.IsPresent(installableResource.InstalledScrText)


jolierabideau (changes requested):
## Summary

The fix is correct and the diff is right. I verified the three ParatextData claims directly against the pinned `ParatextData 9.5.0.22` assembly (IL disassembly of `~/.nuget/packages/paratextdata/9.5.0.22/lib/netstandard2.0/ParatextData.dll`) rather than taking the description on trust — all three hold:

| Claim | Verified |
| --- | --- |
| `IsPresent(ScrText)` → `IsPresentInternal` dereferences its argument | ✅ `IL_0011: ldarg.1; callvirt ScrText::get_Guid()`, no null check. `IsPresent(null)` **does** throw NRE. |
| `InstalledScrText` is assigned only once `InternalInstall` completes | ✅ Exactly **one** `set_InstalledScrText` call site in the whole assembly, at `IL_025b` of `InternalInstall`, whose only `ret`s are `IL_0262`/`IL_0265`. Any early bail-out leaves it `null`. |
| `InstallableDBLResource.Install()` returns `false` on download failure without throwing | ✅ `brtrue.s` on `fileDownloaded`, then `RobustFile::Delete; ldc.i4.0; ret`. Otherwise it tail-calls `base.Install()`. |

So the bug is real, and the localized `%getResources_errorInstallResource_installationFailed%` key (present in `en.json:93` and `es.json:209`) was genuinely unreachable. The warning about the tempting alternative fix is correct too — `base.Install()` returns `fontsFound`, so `false` is the normal return for a successful install of a fontless resource. Good comment to leave behind.

Requesting changes for one trivial blocker plus one process item that I think matters more than the code.

---

### 1. Blocking (trivial) — the diff adds a UTF-8 BOM to the file

```diff
-using System.Diagnostics;
+﻿using System.Diagnostics;
```

That's `EF BB BF` prepended. Confirmed on the PR head vs. `origin/main`. Context:

- **0 of 301** `.cs` files under `c-sharp/` have a BOM.
- `.editorconfig:7` declares `charset = utf-8`, not `utf-8-bom`.

It's an editor artifact unrelated to the fix. `csharpier` won't catch it — I re-ran `dotnet csharpier --check` on the PR's version of the file and it is clean, so the reported green check is genuine; it just doesn't police encoding. Please strip it.

### 2. Important — PT-4368's own description now contains an instruction that would propagate this bug

This is the highest-value item here. The ticket's §5 says the Get Resources path "**compensates**" for the silent `false` return, and *Where the fix actually belongs* item 1 tells the Studio implementer to:

> verify with `ScrTextCollection.IsPresent(resource.InstalledScrText)` **exactly as `DblDownloadableDataProvider.InstallDblResourceCore` already does**

This PR is the proof that "already does" was wrong. Copying that line verbatim into `SyncProjectsCore` reproduces the NRE — in the sync path, where §5 says failures are *already* invisible. Could you update the PT-4368 description before the Studio-side work is written? Otherwise the patch repo inherits the exact defect this PR is removing.

### 3. Minor — this PR does not close PT-4368

Neither DoD item is met here: the `_Resources` folder still isn't created on first sync, and the failure still isn't surfaced *from the sync path*. Those live in Studio's `SyncProjectsCore` plus the §6 investigation into why the connected-ID set comes out empty. The PR body is candid about this, which I appreciate — just flagging that PT-4368 is currently sitting in 👀 Review, and shouldn't be resolved on the back of this. A sub-task or an explicit "part 1 of N" note on the ticket would make that unambiguous.

### 4. Minor — same pattern, adjacent method, left unfixed (currently unreachable)

`UninstallDblResourceCore` does the identical unguarded thing at line 348:

```csharp
var objectToBeDeleted = installableResource.ExistingScrText;
var isPresent = ScrTextCollection.IsPresent(objectToBeDeleted);   // no null check
```

From the IL, `Installed` is `InstallAsDictionary ? ExistingDictionary != null : ExistingScrText != null` — so `Installed == true` does **not** on its own imply `ExistingScrText != null`. In practice it's safe today, since paranext-core never sets `InstallAsDictionary` and the dictionary branch can't be taken. But this same file already hedges that assumption at line 233 (`?? resource.ExistingDictionary?.Guid…`), so the invariant is unstated and inconsistently trusted. Worth knowing too: `ExistingScrText` is a `FirstOrDefault` re-query on every access, not a cached field, so the `Installed` check and line 348 are two independent lookups.

Not asking you to fix it in this PR — just to say out loud whether leaving it is deliberate.

### 5. Minor — no test, and the justification holds up

I checked: there's no `c-sharp-tests/Projects/DigitalBibleLibrary/`, and `InstallDblResourceCore` reaches static `ScrTextCollection` plus a concrete `InstallableResource` with no seam. A test here could only assert against a mock of the very ParatextData behavior at issue. Your reasoning is sound; noting it so nobody asks for a test reflexively.

---

### Pattern conformance — clean

- **Comments are forward-facing** per `.claude/rules/code-quality/forward-facing-comments.md`: they document a ParatextData constraint and a trap, with no PR/ticket narration, no "previously", no stage tags. Exactly the shape the rule asks for — and the replaced comment was genuinely inaccurate.
- **Localization**: reuses an existing key, no new keys, no `en.json` churn. ✅
- **Error handling** matches the surrounding method's `throw new Exception(LocalizationService.GetLocalizedString(...))` idiom. ✅
- **csharpier**: verified clean on the PR's file. ✅
- **Architecture-Decisions.md**: no entry needed — local bug fix, not a pattern choice.
- **Send/Receive write gate**: `InstallDblResourceCore`/`UninstallDblResourceCore` carry no `EnterWrite` scope despite writing and deleting projects on disk — but that is pre-existing, untouched here, and `SendReceiveWriteLockCoverageTests`' heuristics (`.Save(`, `PutText`, `File`/`FileManager` deletes) don't match `installableResource.Install()` or `ScrTextCollection.DeleteProject`. Not this PR's problem; possibly worth its own ticket.

Strip the BOM and I'm happy to approve. Items 2–4 are conversation, not code changes.

---

AI-assisted review (Claude Opus 5). The ParatextData claims above were verified against the pinned assembly's IL, not inferred.

katherinejensen00:
@jolierabideau thanks for the thorough review, and for checking the ParatextData claims against the IL rather than taking them on trust.

Rebased onto current `main` (no conflicts). Taking your points in order:

**1. BOM — fixed.** Stripped in `0a1f207` (`34e1fa4` before the rebase). The file now starts `75 73 69` (`usi`), and no `.cs` file under `c-sharp/` carries a BOM. The same commit rewraps one line of the new guard comment that was 101 characters, since csharpier doesn't reflow comments.

**2. PT-4368's description — updated.** §5 and *Where the fix actually belongs* item 1 now spell out the guarded form, `InstalledScrText == null || !ScrTextCollection.IsPresent(InstalledScrText)`, and explicitly say to copy that rather than the pre-#2709 line. So the Studio-side `SyncProjectsCore` work won't inherit the NRE.

**3. Doesn't close PT-4368 — agreed.** There's now a scope note on the ticket: #2709 is part 1 of N, neither DoD item is met by it, and the ticket shouldn't be resolved on the back of it. Both DoD items and the §6 empty-connected-ID-set investigation remain in Studio's `SyncProjectsCore`.

**4. `UninstallDblResourceCore` — left deliberately.** `ExistingScrText` can be null while `Installed` is true only on the `InstallAsDictionary` branch, and nothing in paranext-core sets `InstallAsDictionary` (no references anywhere under `c-sharp/`). On the double lookup: both reads run inside `_providerGate`, so they can't race a Get Resources fetch/install/uninstall. Only an out-of-band `ScrTextCollection` change landing between the two lines could null it, which I'm treating as outside a Get Resources reporting fix. If dictionary resources are ever wired up, that line needs the same guard. Happy to add it here if you'd rather the invariant not stay implicit.

**5. Tests** — thanks for checking the seam; nothing further planned.

Re-verified after the rebase: `dotnet build` 0 errors (4 pre-existing warnings), `dotnet test c-sharp-tests/` 1716 passed / 0 failed / 7 skipped, and `dotnet csharpier --check` clean on the changed file.

jolierabideau:
Thanks for working through everything. All five items are resolved, and the explanation on `UninstallDblResourceCore` works for me. Since your rebase, though, #2800 (PT-4484) merged and rewrote this verification. `main` now checks `!didExtractFont && installableResource.ExistingScrText == null` and deliberately avoids `IsPresent(InstalledScrText)`, because `RefreshScrTexts` can replace the collection's entries after a good install. So the null reference exception is already fixed on `main`, and this PR now conflicts with it. Could you close this as superseded by #2800? Also, could you update PT-4368's section 5, "Where the fix actually belongs" item 1, and the snippet in your implementation-notes comment, so the Studio `SyncProjectsCore` work copies `main`'s `ExistingScrText` check rather than the `InstalledScrText` form? Otherwise the sync path will report false failures after successful installs.

katherinejensen00:
@jolierabideau agreed, and thanks for catching it. I rebased onto current `main` and both commits drop out entirely. #2800's `!didExtractFont && installableResource.ExistingScrText == null` never reads `InstalledScrText`, so the null dereference this PR guarded against can't happen anymore. It also avoids the false failure that `IsPresent(InstalledScrText)` would report after `RefreshScrTexts` replaces the collection's entries. Closing as superseded by #2800.

I'll update PT-4368 (§5, *Where the fix actually belongs* item 1, and the implementation-notes snippet) so the Studio `SyncProjectsCore` work copies `main`'s `ExistingScrText` check, not the `InstalledScrText` form.

PR #2710, opened 2026-08-24
PT-4403: Explain a blank chapter in read-only texts and Power mode

jolierabideau:
# Code Review Summary

**Branch**: pt-4403-blank-chapter-honest-message

**Base**: pt-4132-improve-empty-state-model-and-bible-texts

**Date**: 2026-08-24

**Files changed**: 6

## Overview

NN 5A of PT-4336 asks for an honest message on an empty chapter in both editable and
non-editable texts. Only the editable Simple case shipped: `EmptyChapterView` was mounted behind
`!isPowerMode && isBlankChapter`, so a Power user editing a project — plus all three read-only
surfaces (Bible text tab, Commentaries tab, model text panel) — fell through to `Editorial`'s
"Enter some Scripture…" placeholder, an edit invitation in texts two of those surfaces cannot edit
at all. This branch adds a blank-chapter branch to each surface and drops the `!isPowerMode` gate.

A blank chapter is a success path — the USJ arrives, it is just empty — so it is invisible to the
missing-book branches, which key off the fetch rejecting. Each surface now orders
loading → missing book → blank chapter → content, with `isChapterBlank` as the predicate.

This branch was rebased out of a dead stack during review. It previously targeted
`pt-4403-honest-message-when-book-does-not-exist`, whose own PR (#2706) was closed unmerged after
its content was carried onto `pt-4132` (#2704) in a stronger form. The rebase replays only the
blank-chapter commit onto `pt-4132`; the eight conflicts were all the same shape — this work was
written against panels that #2704 has since rewritten — and were resolved by keeping #2704's
architecture (`contentState` / `resolveResourceContentState`, `ResourceBookNotAvailable`,
per-resource-type string keys, the `model-text-panel.const` module) and re-expressing the feature
inside it.

## API Changes

**Public API surfaces (`lib/platform-bible-react/`, `lib/platform-bible-utils/`,
`lib/papi-dts/papi.d.ts`, `extensions/src/*/src/types/*.d.ts`): None.** No files in any of those
paths appear in the diff.

Extension-internal only:

- `model-text-panel.const.ts`: added `'%webView_platformScriptureEditor_emptyChapter_messageResource%'`
  to `MODEL_TEXT_PANEL_STRING_KEYS`, widening the derived `ModelTextPanelLocalizedStringKey` union.
  Additive; `ModelTextPanelLocalizedStrings` maps keys optionally, so no consumer breaks.
- `contributions/localizedStrings.json`: added that key in `en` and `es`. No existing key's meaning
  changed.
- `resource-text-panel.web-view.tsx`: now destructures the third (`isLoading`) element of
  `useProjectData(...).ChapterUSJ(...)` — consuming an existing return value, no signature change.

None.

- [ ] **Chapter-0 front matter may be reported as an empty chapter.** `isChapterBlank` is
  `!containsChapterOrVerseNode(usj.content)`, so it returns `true` for any USJ with paragraphs but
  no `\c`/`\v` — which is exactly what book front matter looks like. `canAddChapterNumber`'s TSDoc
  states `chapterNum: 0` is "a reachable case (e.g. front-matter references use `chapterNum: 0`)",
  and neither panel guards on it (`resource-text-panel.web-view.tsx:429` passes `scrRef.chapterNum`
  straight through). The resource panel deliberately does NOT slice away verse-0 front matter
  because it "exists to show" it (ADR-0019). If the PDP delivers intro content at `chapterNum: 0`,
  these panels would paint "This chapter is empty in this resource." over real text. The same
  widening now reaches Power mode in the editor, where a chapter-0 blank state also gets no scaffold
  button (`getEndVerse` returns 0) — a message with no way to act on it. **Needs runtime
  confirmation**; if it reproduces, gate the branch on `scrRef.chapterNum > 0`.

- [ ] **The blank-chapter branch skips the accessibility handling its sibling documents as
  required.** It replaces the `Editorial` subtree exactly as the missing-book branch does, but
  renders a bare `EmptyState` where that branch renders `ResourceBookNotAvailable`. That component
  exists to handle two hazards this reproduces: (a) replacing the editor destroys the focused
  element, so focus falls to `<body>` — it wraps the message in a `tabIndex={-1}` target and calls
  `useFocusReplacedContent`; (b) `EmptyState` mounts its `role="status"` region together with its
  content, which several screen readers do not announce, and navigating blank → blank leaves
  byte-identical text at the same tree position, so `aria-live` reports nothing — it takes an
  `announcementKey` and keys the region on it. Raised independently by three of four analysis
  passes. Note this also makes the new test comment ("the live region is what announces it")
  untrue as written.

- [ ] **The blank-chapter view is duplicated, and `isBlankChapter` is now derived at three sites
  with divergent semantics.** The two panels' blocks are byte-identical apart from the `id` and the
  string lookup. The three derivations also disagree on `undefined`: the editor uses
  `isChapterBlank(usjFromPdp ?? defaultUsj)` (undefined counts as blank) while both new ones require
  `usj !== undefined` (undefined counts as not-blank). `ResourceBookNotAvailable` is the precedent
  for extracting exactly this. A shared component plus a shared predicate would carry the
  stale-load rationale once and make the undefined handling one deliberate choice.

- [ ] **No test covers the resource panel's blank-chapter branch or its loading gate**, while the
  model panel got four. This copy carries the subtlest reasoning in the change — `useData` keeps the
  previous chapter's USJ across a selector change and its default is `EMPTY_USJ`, which is itself
  blank — so a dropped or mis-ordered gate produces a false "empty" on first mount and on every
  navigation. There is no `resource-text-panel*.test.*` file, and the constraint is real (nothing
  under `extensions/` tests a `*.web-view.tsx`, since the module assigns itself to
  `globalThis.webViewComponent`). The project's own answer is to lift the decidable logic into a
  tested module, as `resolveResourcePanelStringKeys` did for the string keys.

- [ ] **`empty-chapter-view.component.tsx:42` TSDoc is now wrong.** It still reads "Replaces the
  editor canvas when the current chapter is effectively blank (Simple mode only)". This change
  renders it in both modes.

- [ ] **The new string opts out of the per-resource-type wording convention.** "This chapter is
  empty in this **resource**" sits beside "This book does not exist in this **Bible text**" /
  "**commentary**" / "**model text**". In the model panel the reader gets "model text" in one state
  and "resource" in the next. `ResourceBookNotAvailable`'s prop doc frames per-type wording as "the
  per-resource-type string convention the resource panel already uses for its other zero-states".
  Either three typed keys resolved the way `bookNotAvailableKey` already is, or neutral wording
  ("This chapter is empty.") that names nothing.

### Minor — Consider

- [ ] The two states that swap in the same content slot have different layout: the new wrapper is
  `flex-1 … p-8`, `ResourceBookNotAvailable` is `h-full … px-4`, so the message shifts position as
  the user navigates between them.
- [ ] `showEmptyChapterView = isBlankChapter` is now a pure alias and reads as if a condition
  remained.
- [ ] The three sibling `&&` blocks each re-state the negated earlier guard. A derived discriminant
  would make mutual exclusivity structural rather than something the reader must verify.
- [ ] Key name `..._emptyChapter_messageResource%` says where it is used rather than why it differs;
  `..._readOnlyMessage%` would be clearer. It also carries the `platformScriptureEditor_` prefix
  inside key lists that otherwise use `resourcePanel_` / `modelTextPanel_` (deliberate and
  documented, but it puts the key out of reach of the typed-key machinery).
- [ ] The message says what is empty but not what to do next; the sibling on the same panel does
  ("Choose a different model text or go to a book it contains"), and the resource selector is still
  mounted right above.
- [ ] `EmptyChapterView`'s message is a bare `<span>` with no `role="status"`, so the editable
  blank-chapter state is announced in neither mode. Pre-existing, but this change broadens its reach
  to Power mode.
- [ ] In the resource panel, `useData` sets `isLoading` from an effect, so for one commit after
  `scrRef` changes the previous chapter's blankness governs the new reference. Self-correcting; worth
  a look only if flicker is observable.
- [ ] `model-text-panel.stories.tsx` has a story per panel state (`NoModelText`, `Selecting`,
  `InstallFailed`, `UnknownResource`, `NoProject`, …) but none for the new blank-chapter state.
- [ ] The new key has no `metadata` entry. Only 5 keys in the file have one, so this matches
  prevailing practice — but a `notes` entry ("read-only surfaces; no Add-chapter-number action")
  would protect the distinction from `..._emptyChapter_message%` for translators, which currently
  lives only in source comments.
- [ ] The read-only panels unmount `Editorial` on the blank branch while the main editor hides it
  with `tw:hidden` to preserve Lexical state. For read-only panels the cost is only scroll position,
  and the remount path is covered by the new "recovers to the editor" test. Informational.

#### Shared Regions Modified

#### Extension Config Changes

- [n/a] `contributions/localizedStrings.json` — extension-specific content strings, not shared build
  infrastructure.
- [n/a] All other changed files are `platform-scripture-editor` source. No webpack, tsconfig,
  package.json, or eslint changes.

## Positive Observations

- The stale-USJ hazard is the real trap here, and both new derivations are gated on the load having
  finished, with comments naming the precise mechanism (`useData` not resetting on selector change,
  and its default `EMPTY_USJ` being itself blank).
- The four new model-panel tests cover the right cases — blank vs. missing book, mid-load
  suppression, and recovery to the editor — and each comment explains the failure it prevents rather
  than restating the assertion.
- `SAMPLE_USJ` was correctly upgraded from an empty `content` array to a real chapter + verse, with a
  comment explaining that an empty array is now semantically *blank* rather than a neutral stand-in —
  the kind of trap that silently invalidates neighbouring tests.
- Removing the `!isPowerMode` guard from the refocus effect alongside the mode gate is the correct
  paired cleanup; leaving it would have made the scaffold insert silently skip focus/scroll in Power
  mode.
- All new user-visible text goes through the localization system in both `en` and `es`; no hardcoded
  UI strings.
- Adding the key to `MODEL_TEXT_PANEL_STRING_KEYS` means `localized-strings.test.ts` enforces en/es
  parity for it automatically, and the resource panel's list reuses the same key, inheriting that
  guard for free.
- Comments added in this diff are forward-facing — constraints and rationale, no PR/ticket
  provenance narration.

## Interview Notes

No author interview was conducted. This review ran as part of a session whose task was to salvage
two PRs from a dead branch stack, and the branch was rebased and re-verified by the reviewer rather
than by the author. The findings below therefore have no author responses attached, and the two
judgment calls made during conflict resolution are the reviewer's, not the author's:

- **`EmptyState` vs `ResourceBookNotAvailable` for the blank-chapter message.** Resolved toward the
  original commit's `EmptyState`, because reusing `ResourceBookNotAvailable` would ship a component
  whose name and TSDoc are explicitly about a *missing book*, and renaming it would touch #2704's
  component, tests, stories, and the model panel. Three analysis passes then flagged the resulting
  accessibility gap, so this call should be revisited — see the second Important finding.
- **Shared vs per-resource-type string key.** Resolved toward the original commit's single shared
  key, to avoid inventing two new key pairs and their Spanish translations during a conflict
  resolution. Flagged as an open question — see the last Important finding.

Two defects were fixed during the rebase rather than deferred:

- A vacuous assertion. `model-text-panel.component.test.tsx` asserted the blank-chapter message is
  not confused with `'This book does not exist in this resource.'` — a string this panel can no
  longer render after #2704, so `queryByText` returned null trivially. Repointed at the model-text
  wording actually in use.
- A missing Spanish string. The commit shipped `en` only; `localized-strings.test.ts` failed on it.
  Added "Este capítulo está vacío en este recurso.", following the sibling strings.

## In-Review Quality Check

Run against the rebased branch:

- `npm run typecheck` — exit 0, clean.
- `npm run lint` — exit 0. Four warnings remain, all pre-existing in files this branch does not
  touch (`resource-reference.utils.ts` duplicate import; two `no-console` in
  `use-effective-resource-reference-list.hook.ts`).
- Prettier over the changed files — all already formatted, no rewrites.
- `npx vitest run src/platform-scripture-editor/` — 61 files, 999/999 tests pass. This includes the
  Spanish parity failure found and fixed during the rebase.

Not run: the Power-mode change and the resource panel's blank-chapter branch have no automated
coverage and were not verified in a running app.

## Suggested Review Focus

- [ ] **Chapter-0 front matter.** Navigate a resource panel and the Power-mode editor to a
  front-matter reference (`chapterNum: 0`) and confirm whether "This chapter is empty" appears over
  real intro text. This is the one finding that could ship a visible wrong claim, and it cannot be
  settled from the diff.
- [ ] **Accessibility parity between the two zero-states.** Decide whether to rename/generalize
  `ResourceBookNotAvailable` and route both messages through it, or to add a focus target and
  remount key to the blank-chapter branch directly. Two adjacent replacement views in the same slot
  behaving differently for the same problem is the core issue.
- [ ] **Extraction.** A shared `ResourceEmptyChapter` component plus a shared blank-chapter
  predicate would resolve the duplication, the divergent `undefined` handling, the accessibility
  gap, and the resource panel's missing test coverage in one move. Worth deciding before more
  surfaces adopt this pattern.
- [ ] **Wording.** Settle whether a blank chapter is per-resource-type ("this Bible text") like its
  neighbours, or deliberately neutral. Both are defensible; the current "this resource" is the one
  option that is inconsistent in every panel.
- [ ] **Power mode.** The mode change has no automated coverage at all. Confirm in a running app
  that the message and the scaffold button behave in Power as they do in Simple.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2710)


katherinejensen00 (approved):
> @katherinejensen00 reviewed 6 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2710)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2711, opened 2026-08-24
Pt 4403 grid book not found

jolierabideau:
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2711)

### DIFF
--- a/extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.test.ts
+++ b/extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.test.ts
@@ -1,7 +1,7 @@
 import { describe, it, expect, vi } from 'vitest';
 import { ScriptureRange } from 'platform-scripture-editor';
 import type PapiBackend from '@papi/backend';
-import { UsjTextContentLocation } from 'platform-bible-utils';
+import { newPlatformError, UsjTextContentLocation } from 'platform-bible-utils';
 import type { SavedWebViewDefinition } from '@papi/core';
 import { MutableRefObject } from 'react';
 import { EditorRef } from '@eten-tech-foundation/platform-editor';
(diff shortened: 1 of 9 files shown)


jolierabideau:
Closing: the work in this PR has landed on `pt-4132-improve-empty-state-model-and-bible-texts` (#2704) in a stronger form, and this PR's base branch is no longer viable.

### Why

**1. The base branch is dead.** This PR targets `pt-4403-honest-message-when-book-does-not-exist`, whose own PR (#2706) was closed unmerged. Its content was carried onto `pt-4132` instead. There is no path from this PR to `main`.

**2. The cell-state logic is superseded.** #2704 implements the same behavior in the same files, and handles a race this PR does not:

| | This PR | #2704 |
|---|---|---|
| Cell state | `bookNotFound` | `bookNotAvailable` |
| Predicate | `isBookNotFoundError` | `isMissingBookError` + `isMissingBookOnScreen` |
| Stale-error handling | none — takes the error branch ahead of `isLoading` | guards it: an error naming a different book or resource falls back to `downloading` |

A cell re-keys its chapter subscription on the grid's shared reference, and the data hook keeps serving the previous selector's result until the new subscription's first update lands. Navigating *out* of a missing book therefore leaves a stale error in hand — this PR's version would claim "not in this text" about a book the resource *does* have, for the whole round trip. `isBookNotFoundError`, the predicate this PR depends on, no longer exists on #2704.

Unit coverage went the same way: this PR had 3 `deriveCellState` cases; #2704 has 9, covering all three plus case-insensitive project-id comparison, the book-just-left stale error, the wrong-resource stale error, and the no-book-on-screen case.

### What was salvaged

This PR's view-component test and Storybook story covered a real gap — `bookNotAvailable` was the only `ResourceCellView` state on #2704 with neither a test nor a story, so the wording guarantee (a missing book never shows "Download failed") was unenforced. Both were ported to #2704 in 8d9ff15ef44, adapted to the `bookNotAvailable` naming and asserting against the real locale string. The test was verified falsifiable: it fails when the rendering branch is removed.

Nothing else here is unique. Superseded by #2704.

PR #2712, opened 2026-08-24
PT-4380: refresh Find results on clear and on retype

jolierabideau:
# Code Review Summary

**Branch**: pt-4380-find-input-refreshes-results

**Base**: origin/main

**Date**: 2026-08-24

**Files changed**: 9

## Overview

Fixes two stale-state bugs in the Find panel, both originating in the search-trigger path. First,
clearing the search box left the previous results on screen: an empty term is an invalid query, so
the auto-search it triggers is gated off, and nothing else cleared the results area — only the clear
button did, by calling `onStopSearch` itself, which left every keyboard route to an empty box
(select-all + delete, backspacing) broken. The rule now lives in a pure predicate plus one effect in
the web view, so every route converges on one code path. Second, typing a new term after a prior
search ran no new search: deduplication between an explicit search and a debounce pending from the
same keystroke was a one-shot flag, cleared only when a debounce actually fired, so any search that
started with nothing pending left the flag armed to swallow the user's next keystroke search. That
flag is replaced by cancelling the pending timer, which can only affect a search genuinely queued.

During review the fix's stated design ("every non-debounce search routes through `startSearchNow`")
was found not to hold — three sites still called the underlying handler directly — and was made true.
A third bug of the same class was found one layer up, at the visibility-deferred catch-up, and
tracked as PT-4418 rather than fixed here. The empty-term rule was also found to cover only half of
the invalid-query condition, and was generalized to cover the other half (deselecting the last book
in the `selectedBooks` scope). Six findings were fixed in-review; the change grew from 8 files to 9.

## API Changes

None. No public API surfaces changed. All nine changed files are internal to the
`platform-scripture` extension (`extensions/src/platform-scripture/src/find*`). Nothing under
`lib/platform-bible-react/`, `lib/platform-bible-utils/`, `lib/papi-dts/papi.d.ts`, or any
`extensions/src/*/src/types/*.d.ts` was touched. The newly exported symbols
(`shouldClearResultsForInvalidQuery` in `find.utils.ts`, `useAutoSearchDebounce` /
`AutoSearchDebounce` in `use-auto-search-debounce.hook.ts`) are extension-internal modules not
re-exported through any `@papi/` module or type declaration file.

Note for the reviewer: fixing PT-4418 properly *would* be an API change — it requires giving
`useRunWhenVisible` a cancel, which is exported from `platform-bible-react`. Keeping that out of this
PR is what preserves the "no API changes" property above.

None.

- [x] **Three non-debounce searches bypassed `startSearchNow`** (`find.web-view.tsx:1573`,
  `:1824`, `:2019` — the external-scripture-change re-run and both post-replace re-searches), so they
  never cancelled a pending auto-search, contradicting the design comment at `:1448`. Not cosmetic: a
  debounce armed before the replace fires ~`SEARCH_DEBOUNCE_DELAY_MS` later and starts a second
  search, and because `handleStartSearch` consumes and resets `isPostReplaceSearchRef`, that second
  run executes as a plain search — clearing `results`, resetting `focusedResultIndex`, and defeating
  the `pendingAdvanceIndexRef` advance-to-next behavior. _(fixed during review: `startSearchNow` now
  returns the promise from `handleStartSearchRef.current()` so `await` works; a new `startSearchNowRef`
  lets effects and async callbacks call it without adding it to dependency arrays; all three sites
  now call `startSearchNowRef.current()`. The debounce itself still calls `handleStartSearchRef`
  directly — it IS the pending auto-search, so routing it through the canceller would make it cancel
  its own timer.)_
- [ ] ~~`cancelPendingAutoSearch` cannot reach the catch-up flag inside `useRunWhenVisible`~~
  _(Author confirmed the gap is real and that they had not modelled the `useRunWhenVisible` layer when
  designing cancellation — their mental model was "armed timer → cancel timer → no duplicate."
  Deliberately tracked as PT-4418 rather than fixed here: it is a third, distinct bug; the correct fix
  changes an exported hook signature in `platform-bible-react`, pulling a library API change into a PR
  that currently has none; and the in-extension workaround would be a second ad-hoc one-shot flag,
  exactly the pattern this PR removes. A `TODO(PT-4418)` at `find.web-view.tsx:1450` links the code to
  the ticket.)_
- [x] **The clear button's new contract had no test.** It no longer calls `onStopSearch(true)` and
  only calls `onSearchTermChange('')`; `find.component.test.tsx` stubbed `onStopSearch` and held the
  `%webView_find_clearSearch%` label but never clicked the button. Because the behavior is now fully
  delegated to the container, a regression (someone re-adding `onStopSearch`, or the term-emptying
  breaking) would be silent. _(fixed during review: added a `Find — clear search button` suite
  asserting the click calls `onSearchTermChange('')` and does NOT call `onStopSearch`, plus a
  falsification case. Verified the key test genuinely detects the regression by temporarily
  reintroducing the old `onStopSearch(true)` call and confirming it fails.)_

### Minor — Consider

- [x] **The invalid-query effect could fire more than once per clear** (`find.web-view.tsx`).
  `handleStopSearch(true)` sets `results`/`searchStatus` synchronously and only then awaits
  `abandonFindJob()`, so a poll iteration already in flight can land afterward, re-set `searchStatus`
  to `'running'`, re-satisfy the predicate, and trigger a second clear + abandon. It converged, so it
  was a redundant round-trip rather than a loop. _(fixed during review: added
  `hasClearedForInvalidQueryRef`, re-armed by the query becoming valid again. The re-arm is derived
  from state on every render rather than set by an event, specifically so it cannot become the kind of
  stale forward-travelling flag this PR removes.)_
- [x] **`shouldClearResultsForEmptyTerm` covered only half of `isFindQueryValid`.** The other
  invalid case — `scope === 'selectedBooks'` with `selectedBookIds` emptied — left stale results on
  screen for the identical reason. Confirmed reachable and genuinely stale:
  `find.component.tsx:601` returns `resultsAreaState: 'none'` on `results.length > 0` *before* the
  `invalidQueryPrompt` branch, so the "Select at least one book" placeholder only appears when there
  are no results. _(fixed during review: renamed to `shouldClearResultsForInvalidQuery` and re-keyed
  off `isSearchQueryValid: boolean` — matching how the sibling `gateStartSearch` already takes that
  flag — so both routes into an invalid query clear. The web view passes its existing
  `isSearchQueryValid` memo, so validity is not recomputed. Tests updated, plus a case feeding
  `isFindQueryValid`'s real verdict for both invalid routes.)_
- [x] **Invalid `FindJobStatus` literal in the new tests** — three of the new predicate tests passed
  `searchStatus: 'finished'`, which is not a member of
  `'running' | 'stopped' | 'errored' | 'exceeded' | 'completed'`. The tests still passed (the
  predicate only checks `!== undefined`), but this PR newly introduced a type error and the tests read
  as asserting against a status that can never occur. Found during review; not surfaced by any of the
  four analysis passes. _(fixed during review: changed to `'completed'`.)_
- [x] **Confusing near-identical names for two layers of one call chain** — `requestAutoSearch` was
  destructured to `requestDebouncedAutoSearch` only so the name could be reused for the
  `useRunWhenVisible` wrapper. _(fixed during review: the hook's export keeps its own name and the
  visibility-gated wrapper is now `requestAutoSearchWhenVisible`. Worth noting for the reviewer: the
  rename immediately exposed the exact hazard the finding described — the two existing call sites read
  `requestAutoSearch`, which after the rename meant the UN-gated one, and were repointed at the gated
  wrapper.)_
- [x] **Backward-facing framing in `useAutoSearchDebounce`'s TSDoc** — it explained the design by
  narrating the flag mechanism it replaced. _(fixed during review: kept, but reworded to lead with the
  invariant. Judged worth keeping rather than deleting: stripped of all PR context the paragraph still
  warns the next maintainer off a wrong design, which is the forward-facing test. It now states that
  cancellation is scoped to what is pending at the instant it is called, whereas a flag is a message
  that travels forward in time until something consumes it.)_
- [x] **Clear button unmounted itself and dropped focus to `document.body`** — it renders only while
  `searchTerm` is truthy, so a keyboard user who tabs to Clear and presses Enter was stranded with
  nothing focused and no way back to the box but the mouse. Pre-existing, but this diff rewrites that
  exact `onClick`. _(fixed during review: added an optional `onFocusSearchInput` prop, wired to the
  container's existing `focusSearchInput` callback, called after the term is emptied. An explicit
  callback was chosen over merging refs inside the component because no `mergeRefs`/`composeRefs`
  helper exists in the repo and adding a general one for a single consumer would be premature. Covered
  by a test.)_
- [x] **Whitespace-only term wiped results while the clear "X" stayed visible** — the button rendered
  on truthy `searchTerm` rather than trimmed content, so typing a single space over a completed search
  produced "results gone, idle prompt showing, clear button still offered." _(fixed during review:
  introduced `hasClearableTerm = searchTerm.trim() !== ''` and keyed both the button and the input's
  padding class off it, so the button and the results area agree. Covered by a test asserting the
  no-button and idle-prompt states together.)_
- [ ] ~~`find.stories.tsx` may no longer clear results now that the clear button is not
  self-contained~~ _(Verified a non-issue, no change needed. The story derives `liveSearchStatus` from
  `isSearchQueryValid` (`find.stories.tsx:466-468` — invalid → `undefined`) and derives `baseResults`
  from `runSearch({ term: searchTerm })`, so clearing the box empties results AND resets status,
  landing on `idlePrompt` exactly as the app does. The story deliberately shares `isFindQueryValid`
  with the web view, so it stays aligned with the generalized predicate.)_
- [ ] ~~`useAutoSearchDebounce` is entirely generic and could live in `platform-bible-react`~~
  _(Deliberately left in the extension. `react-patterns.md` reserves `lib/platform-bible-react/` for
  multi-consumer reusable components, and `porting-discipline.md` says to wait for the second
  consumer. One consumer exists today.)_
- [ ] ~~`find.web-view.tsx` wiring is untested (the file has no test file at all)~~ _(Author chose to
  leave it, and the repo convention supports that decisively: `extensions/src` contains 21
  `*.web-view.tsx` files and ZERO `*.web-view.test.tsx` files, while the `find/` directory alone has 10
  test files for its hooks, utils, and component. The established pattern is to extract decidable logic
  into tested units — which is exactly what `shouldClearResultsForInvalidQuery` and
  `useAutoSearchDebounce` are. Starting the repo's first web-view test file inside a bug-fix PR would
  introduce a new convention rather than fix a bug. The one piece of new logic still in the untested
  layer is the `hasClearedForInvalidQueryRef` guard, because effect-timing bookkeeping resists
  extraction.)_

[Author response: All nine minor findings were worked through one at a time. The author requested
fixes for six of them; two were closed with evidence rather than changes (the stories harness, the
hook's placement), and one was a deliberate decision to leave (web-view test coverage), which the
author asked to have checked against existing code patterns before deciding.]

#### Shared Regions Modified

#### Extension Config Changes

None. All changes are `.ts`/`.tsx` source and test files under
`extensions/src/platform-scripture/src/`. No `package.json`, `tsconfig.json`, `webpack.config.ts`,
`.eslintrc*`, CI workflow, or other build/config files were touched.

## Positive Observations

- Both original bugs were fixed by extracting decidable logic into unit-testable units (a pure
  predicate in `find.utils.ts`, a hook in `use-auto-search-debounce.hook.ts`) rather than patching the
  web view in place — which is the repo's established pattern for untestable web-view code.
- Replacing the one-shot flag with timer cancellation removes an entire class of "armed forever"
  state, and shrank `useFindSearchTriggers`' public surface by one ref parameter in the process.
- Reuses the existing `debounce` / `DEBOUNCE_CANCELED_ERROR_MESSAGE` utility from
  `platform-bible-utils` rather than reimplementing timer logic, and correctly filters the
  cancellation rejection that utility's TSDoc warns about, so deduplicated searches never become
  unhandled rejections.
- `use-auto-search-debounce.hook.test.tsx` is thorough for a timing-sensitive hook: burst collapsing,
  cancel-with-nothing-pending, repeated rounds after a cancel, latest-callback identity, callback
  stability, and unmount cleanup.
- Both new test suites include an explicit "THE REGRESSION THIS EXISTS FOR" test with a comment
  naming the failure mode — durable and forward-facing.
- Clearing restores the *correct* empty state rather than a misleading one: `handleStopSearch(true)`
  resets `searchStatus` to `undefined`, so the results area shows the idle prompt instead of leaving
  the status bar reading "No results found" for a search the user never ran. It also abandons the
  running job and removes the `find-current-result` editor highlight, so no stale amber highlight is
  left painted in the Scripture editor.
- Clearing by any route now behaves identically — mouse, select-all + delete, and backspacing — closing
  a real keyboard-vs-mouse parity gap for exactly the user the design principles describe.
- The hidden-view case is handled deliberately, with `useRunWhenVisible` wrapping the auto-search
  request and an explicit comment at the sync site explaining the defer-and-collapse behavior for
  Simple mode's inactive Find tab.
- No new user-visible strings; all Find microcopy continues to come from `FIND_LOCALIZED_STRING_KEYS`,
  the clear button keeps both its localized `aria-label` and its tooltip, and the new failure path
  logs via `logger.warn` rather than surfacing to the user.
- The ticket's named suspect (a `webViewController` wait-for-net-obj timeout) was investigated and
  explicitly ruled out with a reason, then tracked separately as PT-4417 — rather than left as an
  unexplained loose end.

## Interview Notes

**Stated purpose**: fix two stale-state bugs in Find — results not clearing when the search box is
emptied, and no new search running when the user retypes after a prior search.

**Design explanation (author's own words, unprompted detail).** Asked why cancelling the pending timer
is fundamentally safer than the old `explicitSearchPendingRef` flag, the author gave a precise and
correct account, including a traced failure sequence:

> The flag doesn't suppress the pending search — it tells the pending search to suppress itself when
> it eventually fires. [...] User types "foo" → debounce arms (fires at T+500ms). Explicit search
> fires at T+100ms → sets the flag, runs the search for "foo". User retypes "food" at T+200ms →
> debounce restarts (old timer cancelled, new timer fires at T+700ms). New timer for "food" fires →
> sees the flag still set → stands down, clears the flag. No search ever runs for "food". The flag was
> placed there to dedup the first timer, but that timer was already cancelled by the retype. The flag
> outlived the search it was meant to suppress and was consumed by a legitimate search that needed to
> run.

And the invariant: *"after `cancelPendingAutoSearch()` returns, there is no pending search. The timer
is removed from the queue, not just warned. [...] The flag is a message that floats forward through
time and can be received by the wrong consumer. Cancellation is a point-in-time deletion scoped to
exactly what's pending at the moment of the call."* This is a strong demonstration of understanding —
the author can articulate not just what the fix does but why the alternative is structurally wrong,
and identified precisely when the two designs diverge (the moment the user retypes inside the window).

**One acknowledged gap in the author's model.** The author stated directly that they did not model the
`useRunWhenVisible` layer when designing `cancelPendingAutoSearch`: *"My mental model was: armed
debounce timer → cancel timer → no duplicate. I didn't account for the case where the hook already
absorbed the call into `isRunPending` instead of the timer."* They then reasoned correctly and
unprompted about the consequences, naming all three triggers that reach it. This is a scoping gap that
the author diagnosed accurately once raised, not a misunderstanding of their own code — but it is worth
covering in the meeting because it is the same bug class the PR exists to fix, and it is now deferred
to PT-4418.

**Decisions made during review**: fix PT-4418 separately to keep a library API change out of this PR;
generalize the clear predicate to cover deselected-books rather than ship a half-fix; follow the
repo's existing convention on web-view test coverage (the author asked what the existing patterns were
before deciding, rather than accepting the finding at face value).

**Unresolved items**: none. Every Critical/Important finding was either fixed or deliberately deferred
with a stated rationale and a linked ticket.

**Areas the author did not understand or deferred to AI**: none. The author implemented the
`startSearchNow` fix themselves and correctly explained why the debounce path must be excluded from it.

## In-Review Quality Check

- `npm run typecheck` — **passed clean**, all 10 extension workspaces plus root, zero errors.
- `npm run lint` — **passed** (exit 0). The only remaining warnings are pre-existing
  (`import/no-duplicates` ×2, `no-console` ×2) in `platform-scripture-editor`, which this branch does
  not touch.
- **Prettier** (scoped to the branch's changed files only, not a workspace-wide `npm run format`) —
  auto-fixed one line wrap in `find.web-view.tsx` on the `useAutoSearchDebounce(...)` call. The other
  eight changed files were already correctly formatted.
- `npm test` — all branch-relevant suites pass: the `platform-scripture` workspace runs 23 files /
  449 tests green, including `find.component.test.tsx` (30), `find.utils.test.ts` (96),
  `use-auto-search-debounce.hook.test.tsx` (8), and `use-find-search-triggers.hook.test.tsx` (10).

**One test failure, not caused by this branch and left unfixed**: `platform-bible-react`'s Storybook
browser project fails to import `.storybook/vitest.setup.ts` while running
`project-selector.stories.tsx` (`SyntaxError: missing ) after argument list`). This branch changes zero
files under `lib/`, so that setup file is byte-identical to the merge-base commit and the failure
exists there too; re-running the spec in isolation passes 11/11. It is a flaky
Storybook/vitest-browser setup-transform race under full-suite parallel load.

## Suggested Review Focus

- [ ] **PT-4418 — the deferred duplicate-search bug at the visibility layer.** The author did not model
  `useRunWhenVisible` when designing cancellation, and it is the same bug class this PR fixes, one layer
  up. Worth walking through together: is deferring it correct, and is the proposed fix (a cancel on
  `useRunWhenVisible`, changing a `platform-bible-react` export) the right shape? Note it matters most
  in Simple mode, where one visible tab per stack makes "Find is hidden" the common state.
- [ ] **The `startSearchNow` routing fix.** Confirm the three repointed sites (`:1573`, `:1824`,
  `:2019`) are the complete set, and that excluding the debounce's own call is right. Also confirm the
  new `Promise` return type is harmless at the two sites that discard it (`:1535`, `:1573`) and at the
  props typed `() => void` (`startSearch` at `:1473`, `onStartSearch` at `:2117`) — `npm run typecheck`
  and `npm run lint` are both clean, but it is worth a human eye.
- [ ] **Scope growth during review.** The PR now fixes more than the two bugs the ticket names: the
  deselected-books clearing case, the clear button's focus return, and the whitespace-only button
  visibility. Each is defensible and tested, but confirm you want them in PT-4380 rather than split out.
- [ ] **The `hasClearedForInvalidQueryRef` guard** is the one piece of new logic in the untested web-view
  layer. Sanity-check the re-arm condition (`isSearchQueryValid`) — it is deliberately derived from state
  every render rather than set by an event, precisely so it cannot become the stale forward-travelling
  flag this PR removes, but that reasoning deserves a second reader.
- [ ] **`onFocusSearchInput` as a new prop.** Confirm an explicit callback is preferred over merging
  refs inside the component. No `mergeRefs` helper exists in the repo, and adding one for a single
  consumer looked premature — but that is a judgment call about the component's API.

---

Two stale-state bugs in the Find panel, both from the search-trigger path.

Clearing the search box left the previous results on screen. An empty term is an invalid query, so the auto-search it triggers is gated off, and nothing else cleared the results area — only the clear button did, by calling onStopSearch itself, which left keyboard clearing (select-all + delete, backspacing) broken. The rule now lives in shouldClearResultsForEmptyTerm and one effect in the web view, so every route to an empty box behaves the same; the clear button's own call is gone.

Typing a new term after a prior search ran no new search. Deduplication between an explicit search and the debounce pending from the same keystroke was a one-shot flag, cleared only when a debounce actually fired. Every search that started with nothing pending — Enter after the typing settled, the project-switch rerun, the restore-time fallback — left the flag armed, and it then swallowed the user's next keystroke search. Replaced with cancelling the pending timer, which can only affect a search that is genuinely queued: useAutoSearchDebounce owns the debounce and the cancel, and startSearchNow routes every non-debounce search through it.

The webViewController wait-for-net-obj timeout named in the ticket was investigated and ruled out as the cause: handleStartSearch never touches the web view controller. It is unhandled-rejection noise from usePromise and is tracked separately in PT-4417.

Verified in the running app in both Simple and Power mode: results clear on an emptied box via keyboard and via the clear button, and retyping after an explicit Enter search produces a fresh result set.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2712)


katherinejensen00 (changes requested):
Thanks for all of your work on this, Jolie!

The good part first: both original bugs are real and correctly diagnosed, and the fix shape — extract decidable logic into a pure predicate and a hook rather than patch the web view in place — is the right one for this file. Your flag-vs-cancel analysis in the hook TSDoc is genuinely the best explanation of that distinction I've read in this repo. The THE REGRESSION THIS EXISTS FOR test comments should be the house style.

The headline: three new mechanisms each reintroduce the bug class this PR exists to remove — a one-shot flag that outlives its purpose. The hasClearedForInvalidQueryRef latch, the cancel-before-gate ordering in startSearchNow, and using searchStatus !== undefined as a proxy for "a job ran." Several land in the Replace paths, where the consequence is mutated project data.

Most of the list collapses into three decisions:
- (a) Move cancelPendingAutoSearch() to after gateStartSearch says run → fixes 1, 8, 9, 11.
- (b) Drop the latch; add the abandon-generation re-check in the poll loop (12) instead → fixes 2, 3, 5.
- (c) Key the predicate off job identity rather than searchStatus !== undefined, or move the display rule into resultsAreaState (13) → fixes 7, and makes 4–6 far less reachable.

---

High

1. startSearchNow destroys a queued auto-search even when its own search is then skipped.
find.web-view.tsx:1450 · High
This is the regression I'd fix first, because it's your own bug class. The explicitSearchPendingRef assignment you deleted sat after the if (gate.action === 'skip') return — so a skipped search never suppressed the queued debounce. cancelPendingAutoSearch() now runs before the gate. Sequence: debounce fires → handleStartSearch holds isStartingSearchRef across await abandonFindJob() + await beginFindJob() → user types another character (debounce re-arms) and presses Enter → the queued search is destroyed, then the gate skips on isAlreadyStarting with shouldRetryWhenPdpReady: false. Nothing re-arms. The newest term is never searched and never reaches recent searches.

2. The hasClearedForInvalidQueryRef latch suppresses the corrective clear, not just a redundant one.
find.web-view.tsx:1559 · High
abandonFindJob and retrieveFindJobUpdate share findPdpMutex (:963), so a poll already awaiting when the clear happens completes first and runs setSearchStatus(update.status) (:1286) — classifyPollAttempt (find.utils.ts:164) snapshots hasActiveJob before its await and never re-checks. End state: results = [], searchStatus = 'running', no active job → skeleton plus a stuck progress bar under an empty search box. The effect re-fires, the latch blocks it. Cancel can't recover it — stopFindJob finds no job and never touches searchStatus (:945). A tighter variant is worse: if the clear lands inside await loadMoreResults() (:1291), setResults((prev) => [...prev, ...newResults]) (:1214) repopulates the just-emptied array — stale results under an empty box, exactly the bug this PR fixes. Without the latch this self-heals; the round-trip it saves is nearly free, since abandonFindJob (:777) makes no PDP call with no active job. (Reproduced.) Recovery is possible by typing a valid term, so it isn't permanent — but the state is wrong and the visible escape hatch is dead.

3. The clear path disarms neither pendingSearchDesiredRef nor the 5 s give-up timer.
find.web-view.tsx:1093 (unchanged line — nearest changed context is the effect at :1548) · High
Search completes with results → findPdp drops → user types → gate skips with shouldRetryWhenPdpReady → armBoundedWait armed for GIVE_UP_AFTER_MS → user empties the box (results cleared, latch set) → 5 s later the wait fires setSearchStatus('errored') + "Searching stopped unexpectedly." The effect re-runs and the latch blocks cleanup. Permanent error banner under an empty search box. handleStopSearch(true) should call clearPendingSearchTimeout() and reset pendingSearchDesiredRef.

4. Clearing mid-Replace-All silently truncates the replace and reports it as complete.
find.web-view.tsx:1877 (unchanged line) · High
The load-everything loop breaks on if (!update ...). Clearing abandons the job, so retrieveFindJobUpdate returns undefined, the loop exits early, and the replace runs against only the already-loaded batches — while the toast reports visibleResultsList.length as if complete and the revert snapshot covers only part. Pre-existing via the clear button, but this PR widens reachability to backspacing, select-all + delete, and deselecting the last book. The search Input is never disabled during a replace (find.component.tsx:849), so nothing prevents it.

---

5. Replace All restores a pre-clear snapshot with a non-functional setResults.
find.web-view.tsx:2003 (unchanged line) · Medium
setResults(allResults.map(...)) writes a local snapshot captured before the clear, while every other replace-path write uses the functional form (:1792, :2021). Clear mid-Replace-All → every stale result comes back marked replaced, under an empty box, and the latch blocks re-clearing.

6. Clearing during the 1 s revert window strands the undo.
find.web-view.tsx:1548 · Medium
The clear isn't gated on isReplacing. Clearing during cancellableDelay(1000, pendingReplaceRevertRef) empties resultsByBook, unmounting the per-row Cancel button — the only affordance that reaches handleCancelReplace (:2053). The ref stays armed but unreachable, so the replacement commits with no undo.

7. The predicate treats any searchStatus as "a find job ran," so it eats the provider-unavailable error.
find/find.utils.ts:241 · Medium
searchStatus has three producers. Open Find with an empty box against an unresolvable find PDP: :1001 sets 'errored' + %webView_find_project_unavailableError%; the new effect immediately wipes both. It never returns — that effect's deps [findPdpAvailability, localizedStrings] don't change and isShowingPdpUnavailableErrorRef stays true — so the panel looks idle, which is the outcome that effect's own comment says it exists to prevent. The user gets the generic "search interrupted" 5 s later instead. It's also history-dependent: if the latch is already armed, the same condition displays fine. (Reproduced.) Narrower than it looks — noOpenProjects already suppresses the status bar and !projectId also yields 'unavailable' — hence Medium, not High. Cheapest fix: gate the clear on !isShowingPdpUnavailableErrorRef.current.

8. pendingAdvanceIndexRef is left armed when the post-replace search gate-skips.
find.web-view.tsx:1827 (unchanged line; the adjacent call at :1834 is changed) · Medium
The gate returns at :1104, before the ref is consumed at :1598. A later unrelated search then hits the advance branch with a stale targetIndex, so Replace mode opens focused on the wrong result — and the next Replace targets the wrong occurrence. Pre-existing, but finding 1 makes it more reachable: the debounce that would previously have fired and consumed the ref is now destroyed.

9. Same skip leaves isPostReplaceSearchRef ar

10. The clear re-arms the restore-time fallback, so the next keystroke fires an un-debounced whole-scope search.
find/use-find-search-triggers.hook.ts:86 · Medium
handleStopSearch(true) sets searchStatus to undefined, which satisfies the guard-less fallback. In any session that opened with a restored term (initialSearchTriggeredRef.current === true), typing one character after a clear runs an immediate whole-scope search for a one-character term — and because useFindSearchTriggers is registered at :1463, before the auto-search effect at :1505, the debounce armed in the same commit then fires a second search 500 ms later. Pre-existing via the clear button; newly reachable from every keyboard route.

11. Background triggers now cancel the user's in-flight typing debounce.
find.web-view.tsx:1533 (also :1583, :1834, :2029) · Medium
findPdp reconnecting or an editor edit landing in Replace mode now searches whatever is mid-word in the box, as its first act. The diff deleted the sentence that warned about exactly this: "setting it unconditionally previously risked swallowing the user's next legitimate keystroke." Fixing 1 fixes this too.

12. The poll loop has no abandon-generation guard — this is the actual source of the race the latch papers over.
find.web-view.tsx:1286 (unchanged line) · Medium
FindJobStatusReport already carries jobId (platform-scripture.d.ts:557), and neither the poll nor loadMoreResults compares it to activeJobIdRef.current after the await. One line at each site — if (update.jobId !== activeJobIdRef.current) return; — kills finding 2 at the source and also stops a stale poll resurrecting state after handleSelectProjectScrollGroup (:803).

13. The display half of this fix is a render rule implemented as a container side effect.
find/find.component.tsx:617 (unchanged line) · Medium — architecture
resultsAreaState returns 'none' on results.length > 0 before consulting validity, and the result rows at :1160 are gated only on noOpenProjects — so it isn't the "single source of truth" its comment claims. Adding if (!isSearchQueryValid) return searchTerm.trim() === '' ? 'idlePrompt' : 'invalidQueryPrompt'; above :617, plus gating the rows, makes the display correct by construction — no effect, no latch, no race — and leaves the container effect doing only the genuine side effect (abandoning the job).

---

Low

14. handleStopSearch(true) doesn't reset monitoredScope / monitoredVerseRef / monitoredBookIds, so Replace mode keeps live subscribeBookUSFM subscriptions for every previously-searched book for the life of the tab — the whole session in Simple mode. handleStartSearch's error path already sets the precedent at :1152. — find.web-view.tsx:1183 · Low

15. useMemo holds a stateful object owning a live setTimeout. React documents useMemo as a discardable hint; on eviction a timer armed on the old closure survives while the canceller targets the new one. Pre-PR this was useRef(debounce(...)), which guarantees one instance — useRef with lazy init restores that. — find/use-auto-search-debounce.hook.ts:41 · Low

16. hasClearableTerm removes the clear button entirely for a whitespace-only box. The input still visibly holds characters and the caret is offset, but a mouse user now has no affordance to empty it. — find/find.component.tsx:862 · Low

17. The "clears for a deselected-books query the same way" test is a tautology — the predicate takes only a boolean, so both isFindQueryValid calls collapse to false and both assertions are byte-identical to the test at :546. The real risk (the container reacting to one route but not the other) lives in the web view and is untested. — find/find.utils.test.ts:589 · Low

18. The focus-handoff test never empties the term (onSearchTermChange is the default vi.fn()), so the button never unmounts and nothing is actually focused. It passes if focusSearchInput is a no-op or searchInputRef is unattached. Assert document.activeElement. — find/find.component.test.tsx:659 · Low

19. All 8 hook tests pass a synchronous vi.fn(), but production passes a promise-returning function that debounce awaits, holding its shared promise open for the whole search-start. Cancel-while-in-flight — the exact window behind finding 1 — is untested. (Correction to my own first pass: the cancel filter itself is covered — three tests reject the shared promise and run it down its load-bearing branch.) — find/use-auto-search-debounce.hook.test.tsx:18 · Low

20. import { logger } from '@papi/frontend' resolves to undefined under extensions/vitest.config.ts's alias (the mock has only a default export) — I verified this. It's also the first runtime @papi/frontend import in find/, and that directory documents the opposite convention twice: find.utils.ts:246 keeps imports type-only "so this module stays free of the runtime @papi/frontend import," and search-result.component.tsx:37 already defines FindLogger "so the components stay @papi-free." Taking a FindLogger param closes the mock hole and restores the convention in one move. — find/use-auto-search-debounce.hook.ts:3 · Low

21. onFocusSearchInput?: is optional and find.stories.tsx:539 passes neither it nor searchInputRef — so the a11y fix is unexercised in the one harness where it would be reviewed, and Storybook still reproduces the focus-drop bug. Component-Builder-Patterns.md:510 requires every on* prop wired in the Default story. — find/find.component.tsx:232 · Low

22. cross-view-sync-hidden-views.md requires a hidden→visible catch-up test ("Documented intent doesn't survive a refactor; a test does"). This PR rewires the useRunWhenVisible callback and adds none. The sync-site comment at :1474 still asserts the catch-up is sound while the TODO(PT-4418) describing the gap sits 38 lines away. The new invalid-query effect — which reaches into the editor web view to remove find-current-result — records no hidden-case decision at all. — find.web-view.tsx:1485 · Low

23. onStopSearch's shouldClearResults parameter is now unreachable from the component (only onStopSearch(false) at :1226 remains) but the TSDoc still advertises it, inviting someone to reintroduce the mouse-only clear path you just removed. — find/find.component.tsx:314 · Low

---

Comments & documentation
unconditionally, so the argument supports nothing. Trim to the first sentence. — find.web-view.tsx:1526 · Low

25. 11 comment lines above a 17-line effect, in two paragraphs merged with no blank line. Paragraph 1 restates the util's TSDoc almost verbatim; paragraph 2 documents the race whose handling is finding 2. Three or four lines carry it. — find.web-view.tsx:1536 · Low

26. 15 lines of TSDoc for a 3-line predicate, with an @param block that restates the parameter names and a route enumeration duplicated from isFindQueryValid's own doc. Keep the first sentence and the "requires something to actually clear" sentence. — find/find.utils.ts:219 · Low

27. Each rationale is written three times. Cancel-vs-flag appears at the hook TSDoc (:23), the call site (find.web-view.tsx:1436), and the test (:57); the focus rationale at the prop TSDoc (:227), inline in JSX (:869), and the test (:656). Keep the TSDoc copy, delete the other two — the 4-line inline comment inside JSX props (find.component.tsx:869) explains container behavior from inside the presentational component, which is where it will rot first. — find/use-auto-search-debounce.hook.ts:23 · Low

28. Backward-facing narration in test comments — raised once at PR level per forward-facing-comments.md, which names this exact form ("'previously', 'used to' … the most common AI-authored form"). Offenders: "Deduplication used to be a one-shot…" (use-auto-search-debounce.hook.test.tsx:57), "…stayed on screen with nothing in the box" (find.utils.test.ts:543), "…the split this button's behavior was collapsed to avoid" (find.component.test.tsx:645). Describe what the code does, not what it stopped doing. — find/use-auto-search-debounce.hook.test.tsx:57 · Low

> @katherinejensen00 reviewed 9 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2712#-:-P-rRUQ3DTMe1aV3AO5F:b-myd0p5)* status: all files reviewed, 1 unresolved discussion (waiting on jolierabideau).

___
*[`extensions/src/platform-scripture/src/find.web-view.tsx` line 1450 at r1](https://reviewable.io/reviews/paranext/paranext-core/2712#-P-rRitc6oS4LnqSQuHY:-P-rRitc6oS4LnqSQuHZ:b-7hxgps) ([raw file](https://github.com/paranext/paranext-core/blob/77cad27ed5939b717d4b4616048c0c343e333a87/extensions/src/platform-scripture/src/find.web-view.tsx#L1450)):*
> ```XML
>       // `useRunWhenVisible` absorbed into its pending catch-up while the tab was hidden. Such a
>       // search still re-runs on activation. Needs a cancel on `useRunWhenVisible`.
>       cancelPendingAutoSearch();
> ```

1. startSearchNow destroys a queued auto-search even when its own search is then skipped.
find.web-view.tsx:1450 · High
This is the regression I'd fix first, because it's your own bug class. The explicitSearchPendingRef assignment you deleted sat after the if (gate.action === 'skip') return — so a skipped search never suppressed the queued debounce. cancelPendingAutoSearch() now runs before the gate. Sequence: debounce fires → handleStartSearch holds isStartingSearchRef across await abandonFindJob() + await beginFindJob() → user types another character (debounce re-arms) and presses Enter → the queued search is destroyed, then the gate skips on isAlreadyStarting with shouldRetryWhenPdpReady: false. Nothing re-arms. The newest term is never searched and never reaches recent searches.

jolierabideau:
Worked through all 26. Your three-way collapse was the right read — reworking around (a), (b) and (c) took most of the list out at the source rather than patching each site. Pushed in 7361854, on top of a merge with latest main.

**(a) Cancel only once the gate says the search runs — findings 1, 8, 9, 11**

You were right that this was my own bug class, and right about where the old `explicitSearchPendingRef` assignment sat. `startSearchNow` is gone entirely, along with `startSearchNowRef`: `handleStartSearch` now owns the cancellation and does it immediately after the gate passes, so a skipped search can't destroy a queued one. Every call site is back on `handleStartSearch` / `handleStartSearchRef`, which also removes the "is this the complete set of repointed sites?" question you raised in the first round's review focus — there is nothing to repoint any more.

The debounce's own call lands there too, where the cancel is a no-op: its timer has already fired, and the stretch from the gate to the cancel is synchronous, so nothing can have queued another in between. That's why it no longer needs excluding.

On 8 and 9 specifically — the reachability fix alone left both refs armed on a skip, so I closed those at the source too: a **non-retryable** skip now clears `isPostReplaceSearchRef` and `pendingAdvanceIndexRef`. A retryable skip deliberately keeps them, because that same post-replace search does still run once the provider is ready.

**(b) Job-identity guard instead of the latch — findings 2, 12**

`hasClearedForInvalidQueryRef` is deleted. The poll loop and `loadMoreResults` now compare `update.jobId` to `activeJobIdRef.current` after the await and drop the update if the job has been abandoned. That covers both variants you traced — the resurrected `'running'` status and the `setResults((prev) => [...prev, ...newResults])` repopulation — and, as you noted, also stops a stale poll resurrecting state after `handleSelectProjectScrollGroup`.

Finding 3 needed its own change on top: `handleStopSearch(true)` now resets `pendingSearchDesiredRef` and calls `clearPendingSearchTimeout()`, so the 5 s give-up can't put a permanent error banner under an empty box.

**(c) Display rule in `resultsAreaState` — finding 13**

Main had already moved the `invalidQueryPrompt` branch above `results.length > 0` while this branch was in review, so the merge brought half of it in. I added the other half — the empty-term case returns `idlePrompt` from the same branch — and gated the result rows on `!isSearchQueryValid`, since a placeholder rendered *above* stale rows is still stale rows. The display is now correct by construction with no window waiting on the effect, and the effect is down to the genuine side effect of abandoning the job. Two component tests cover both routes, replacing the tautology in finding 17.

**Finding 7** — took the cheap fix: the clear is gated on `!isShowingPdpUnavailableErrorRef.current`.

**Findings 4 and 6** — the clear is gated on `!isReplacing`, with `isReplacing` in the deps so it happens as soon as the replace finishes. That covers both: the load-everything loop isn't truncated, and the per-row Cancel button stays mounted through the revert window. I chose this over disabling the search Input during a replace because it keeps the user's clear intent rather than dropping it.

**Finding 5** — functional form, matching `:1792` and `:2021`.

**Finding 10** — the fallback is now genuinely one-shot via its own ref. Its comment already claimed to be a mount-time safety net; the implementation just didn't match. Two tests, both verified to fail without the ref.

**Finding 20** — verified, and it's worse than a mock hole: `logger` is `undefined`, so a non-cancel rejection would throw a TypeError out of the catch handler. The hook takes a `FindLogger` now, restoring the convention `find.utils.ts` and `search-result.component.tsx` both document. Covered by a test that asserts the warning actually lands.

**Finding 15** — `useRef` with lazy init, and the comment now says why the memo was wrong rather than just what the code does.

**Finding 16** — this one I resolved *against* the previous round's whitespace finding, which asked for the opposite. Between "clear button offered next to an idle prompt" (cosmetic) and "box visibly holds characters with no mouse affordance to empty it" (a dead end), the dead end is worse, so the button is back on `searchTerm` and the test now asserts it stays. Flagging it as a reversal rather than burying it.

**Findings 18, 19, 21, 22** —
- 18: the test now renders a harness with real state and a real ref, so the term actually empties, the button actually unmounts, and the assertion is on `document.activeElement`.
- 19: added a cancel-while-in-flight test with a promise-returning `startSearch`, asserting the running search isn't stood down and the next request can still queue.
- 21: `searchInputRef` and `onFocusSearchInput` are wired in the story, so the focus return is exercisable there.
- 22: on the catch-up test — `useRunWhenVisible` already has one (`use-run-when-visible.hook.test.tsx`, "runs the deferred work once when the view becomes visible" and "collapses repeated requests made while hidden into a single catch-up run"), so the mechanism is covered at the layer that owns it; what this PR changed is which callback is passed to it. I did fix the documentation split you flagged: the sync-site comment now states the PT-4418 gap in place, and the `TODO` inside `handleStartSearch` points at it instead of re-explaining it 38 lines away. The new invalid-query effect now records its hidden-case decision explicitly — it runs while hidden on purpose, because abandoning a job isn't layout-dependent and deferring it would leave a find job running on the backend for the whole session in Simple mode.

**Findings 14, 23, 24, 25, 26** — all applied as described: monitored scope/verse/books reset on clear; `onStopSearch` narrowed to `() => void` with the TSDoc rewritten; the readiness-retry comment trimmed to its first sentence, since the dedup argument it made is moot now that cancellation is unconditional; the 11-line effect comment and the predicate TSDoc cut down.

**Verification** — `npm run typecheck` clean; `npm run lint` 0 errors (the 4 remaining warnings are the pre-existing `platform-scripture-editor` ones this branch doesn't touch); `npm test` for `platform-scripture` 82 files / 1321 tests green. The new one-shot-fallback tests were verified to fail with the guard removed.

Nothing left open on my side. PT-4418 is still the one deferred item, for the reasons in the summary.

katherinejensen00 (approved):
:lgtm: Great work, Jolie!

> @katherinejensen00 reviewed 10 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2712#-:-P-tTt8ZAUYmEQKOeZZH:buirp)* status: all files reviewed, 1 unresolved discussion (waiting on jolierabideau).

Issue #541, opened 2023-10-09
Persist network objects on the renderer between refreshes?


tjcouch-sil:
**User Story**
As a core developer, I want to use network objects for my service hosts everywhere so that the service hosts are consistent between processes.

**Description**

In the new service host pattern, the computation part of services is in a network object on one process, and other processes call that network object. On the renderer, the network object will be disposed on refreshing, so the proxies may be revoked on other processes. We should probably fix this as we are pretty much stuck to using commands in service hosts on the renderer right now because of this.

See the difference between `dialog.service.host.ts` and `project-lookup.service.host.ts`

PR #2714, opened 2026-08-24
PT-4412: Install DOM globals in Node processes for scripture-utilities

katherinejensen00:
> **Note — depends on an unpublished upstream release.** This branch targets the
> [scripture-editors PR #541](https://github.com/eten-tech-foundation/scripture-editors/pull/541)
> build, which is not yet published. `package.json` still declares
> `@eten-tech-foundation/scripture-utilities@~0.1.6`, which does *not* contain that change — so with
> the currently-declared version the polyfill is inert (0.1.6 still bundles xmldom internally) and
> what lands is the dependency-graph half. Landing it now is deliberate: it puts the globals in place
> ahead of the upgrade, so the eventual version bump is a one-line PR rather than a combined risky
> one. End-to-end behavior was verified against a local `yalc` link of the #541 build.

## Summary

scripture-editors PR #541 drops `@xmldom/xmldom` from `@eten-tech-foundation/scripture-utilities`
(~89 KB off browser bundles) and makes the USX⇔USJ converters use the platform's **native**
`DOMParser`/`XMLSerializer`. Browsers and web views supply those; plain Node does not, so the
converters throw `requires a DOM environment`. The package's README prescribes installing
`@xmldom/xmldom` as globals.

Our extension host is plain Node and reaches the converters through `platform-scripture`'s finder
PDPE (`usxStringToUsj`) and extender PDPE (`usjToUsxString` / `usxStringToUsj`) — i.e. **every
scripture read and write**. This PR supplies the globals so that upgrade is safe to take.

### Why review this

Not part of the current epic. It is dependency-unblocking work: without it, taking the next
`scripture-utilities` release breaks every scripture read and write in the extension host. Worth
reviewing now so the approach is settled before the upstream release forces the upgrade.

## Changes

- **`src/node/polyfills/dom-globals.polyfill.ts`** — side-effect module that installs
  `DOMParser`/`XMLSerializer` from `@xmldom/xmldom` **only when absent**, so a real DOM (renderer,
  jsdom tests) always wins.
- **`src/extension-host/extension-host.ts`** — imports it *first*. It has to be an import, not a
  function call: ESM evaluates a module's whole import graph before its own body, so a call at the
  top of the entry point would run too late for anything that converts at load time.
- **`vitest.setup.node.ts`** — loads the polyfill for the two `node`-environment vitest projects
  (`extensions`, `lib/platform-bible-utils`), which exercise the same converters. Kept out of the
  shared setup so the jsdom projects don't pull xmldom into ~150 workers for nothing.
- **`package.json`** — declares `@xmldom/xmldom` (`^0.9.8`) as a real dependency; it previously
  reached us only transitively.
- **`extensions/src/platform-enhanced-resources/package.json`** — pins `@xmldom/xmldom` to
  `^0.8.10`. Its `marble-converter.ts` passes `errorHandler` as an object, which is 0.8-only (0.9
  replaced it with `onError` and throws a `TypeError`). That dependency was previously undeclared
  and resolved to a hoisted copy, so the root declaration would otherwise have silently upgraded it
  and broken every marble conversion.
- **`adr-node-dom-globals-polyfill`** in `.context/standards/Architecture-Decisions.md` — records the decision, the
  rejected alternatives, and why the **main process is deliberately not polyfilled** (no converter
  call sites; nothing in its import graph converts at load time — the only module-load-time
  `usxStringToUsj` calls are in `*.data.ts` fixtures imported solely by tests and stories).

Two `@xmldom/xmldom` majors now live in the tree on purpose. `adr-node-dom-globals-polyfill` spells out the consequence
worth a reviewer's attention: npm hoists 0.8.x to `extensions/node_modules/`, making it the default
for *every* package under `extensions/src/*`, and `import/no-extraneous-dependencies` is off — so the
next extension to import `@xmldom/xmldom` binds to 0.8.x with no warning.

## Testing

- [x] Unit tests for the polyfill (`src/node/polyfills/dom-globals.polyfill.test.ts`) — installs
      when absent, leaves an existing global untouched, round-trips parse/serialize.
- [x] Manual verification in the running app against a local build of PR #541: Find, chapter load,
      and verse edit+save all round-trip with zero `requires a DOM environment` errors — covering
      all six call sites and both conversion directions.
- [ ] Full `npm run typecheck && npm run lint && npm test` — **not green locally**, because the
      local `yalc` link for `@eten-tech-foundation/*` has no `dist/`. Every failure is
      `Cannot find module '@eten-tech-foundation/…'`, none in changed files. Deferred to CI, which
      resolves the published packages.

## AI Involvement

AI-assisted (Claude Opus 5). Claude drafted the polyfill, its tests, the vitest wiring, and the ADR,
and did the dependency-resolution archaeology behind the 0.8/0.9 split. Author reviewed all changes
and performed the in-app verification against the PR #541 build.

## Risk Level

**Low–Medium.** The polyfill is additive and defers to any existing global, so no DOM-bearing process
changes behavior. The medium half is the dependency graph: promoting `@xmldom/xmldom` to a root
declaration changes what resolves under `extensions/src/*`, which is why the
`platform-enhanced-resources` pin is part of this change rather than a follow-up.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2714)

### DIFF
--- a/.context/standards/Architecture-Decisions.md
+++ b/.context/standards/Architecture-Decisions.md
@@ -2233,11 +2233,11 @@ step, no automation. Just a record.
   plus a non-dev 0.9.x nested under scripture-utilities), letting an existing global win so a real
   DOM is never overridden. It is a **side-effect module** imported **first** by `extension-host.ts`
   — see that module's docs for why a callable function would run too late. `vitest.setup.node.ts`
-  installs it too, because the `node`-environment test projects (`extensions`,
-  `lib/platform-bible-utils`) run the same converters and would otherwise fail outright; it is kept
-  out of the shared `vitest.setup.ts` so the jsdom projects, which already have a real DOM, do not
-  pull xmldom into all ~150 of their workers for nothing. The main
-  process is deliberately **not** polyfilled: it has no converter call sites.
+  installs it too, because the `node`-environment test projects that run the converters would
+  otherwise fail outright; each such project wires that file in itself, so the ones that never touch
+  the converters do not load it. It is kept out of the shared `vitest.setup.ts` so the jsdom
+  projects, which already have a real DOM, do not pull xmldom into all ~150 of their workers for
+  nothing. The main process is deliberately **not** polyfilled: it has no converter call sites.
 - **Alternatives:** **Let the package fall back to `@xmldom/xmldom` when no DOM is present** (what
   scripture-editors issue #516 floated) — needs no core change, but it is a cross-repo change we do
   not control and would have to land first. **Install from `global-this.model.ts` alongside
--- a/.context/standards/Architecture-Decisions.md
+++ b/.context/standards/Architecture-Decisions.md
@@ -2216,3 +2216,52 @@ step, no automation. Just a record.
   for a typed surface rather than a state key.
 - **Source:** PT-4346, global BCV control showing books from open resources.
 
+## adr-node-dom-globals-polyfill: Node processes install `@xmldom/xmldom` DOM globals; the extension host does it in a first-import side-effect module
+
+- **Date:** 2026-08-22
+- **Status:** Accepted
+- **Context:** scripture-editors PR
+  [#541](https://github.com/eten-tech-foundation/scripture-editors/pull/541) drops `@xmldom/xmldom`
+  from `@eten-tech-foundation/scripture-utilities` (~89 KB off browser bundles) and makes the
+  USX⇔USJ converters use the platform's **native** `DOMParser`/`XMLSerializer`. Browsers and web
+  views have those; our extension host is plain Node and does not, so the converters throw. It
+  reaches them through `platform-scripture`'s finder and extender PDPEs — every scripture read and
+  write. There was no runtime DOM shim anywhere in `src/`; every `jsdom` reference was in tests.
+- **Decision:** paranext-core supplies the globals rather than waiting on the package.
+  `src/node/polyfills/dom-globals.polyfill.ts` installs them from `@xmldom/xmldom` (now a declared
+  root dependency; previously it reached us only transitively — a dev-only 0.8.x hoisted via `plist`
+  plus a non-dev 0.9.x nested under scripture-utilities), letting an existing global win so a real
+  DOM is never overridden. It is a **side-effect module** imported **first** by `extension-host.ts`
+  — see that module's docs for why a callable function would run too late. `vitest.setup.node.ts`
+  installs it too, because the `node`-environment test projects (`extensions`,
+  `lib/platform-bible-utils`) run the same converters and would otherwise fail outright; it is kept
+  out of the shared `vitest.setup.ts` so the jsdom projects, which already have a real DOM, do not
+  pull xmldom into all ~150 of their workers for nothing. The main
+  process is deliberately **not** polyfilled: it has no converter call sites.
+- **Alternatives:** **Let the package fall back to `@xmldom/xmldom` when no DOM is present** (what
+  scripture-editors issue #516 floated) — needs no core change, but it is a cross-repo change we do
+  not control and would have to land first. **Install from `global-this.model.ts` alongside
+  `polyfillLocalStorage()`** — matches the existing polyfill pattern, but that module already
+  transitively imports `platform-bible-utils`, so its graph is fully evaluated before the install
+  would run; the ordering guarantee would be accidental rather than structural. **Pin the root dep
+  to `^0.8.10`** to match the dev-only hoisted copy — rejected: scripture-utilities parses and
+  serializes with 0.9.x internally, and this code writes scripture to disk, so the extension host
+  should stay on the 0.9 line rather than drop to an older parser. Note this does not pin an exact
+  engine: declaring `^0.9.8` at the root dedupes the package's own nested 0.9.10 away, so the
+  converters now run on whatever 0.9.x resolves at the root (0.9.12 today).
+- **Consequences:** Two `@xmldom/xmldom` majors now live in the tree on purpose.
+  `platform-enhanced-resources` declares `^0.8.10` because `marble-converter.ts` passes
+  `errorHandler` as an object, which 0.9 rejects with a `TypeError`; that dependency was previously
+  undeclared and resolved to the hoisted copy, so the root declaration would otherwise have silently
+  upgraded it and broken every marble conversion. npm satisfies that range by hoisting 0.8.x to
+  `extensions/node_modules/`, which makes it the default for **every** package under
+  `extensions/src/*`, not just this one — `import/no-extraneous-dependencies` is off, so the next
+  extension to import `@xmldom/xmldom` binds to 0.8.x with no warning. Two follow-ups worth taking
+  together: `convertMarbleChapterXml` runs only in a web view, which already has a native
+  `DOMParser`, so dropping the dependency there would remove the pin, the version split, and this
+  paragraph at once. Separately, `@xmldom/xmldom` emits no `parsererror` element, so PR #541's
+  malformed-XML guard is inert under the shim; fatally malformed USX still throws (as a `ParseError`
+  from inside `parseFromString`, not the renderer's `Invalid USX:` error), but *recoverable* defects
+  are silently repaired rather than rejected — on the extension host that repaired form is what gets
+  written back to disk. Revisit if the package ever ships its own Node fallback.
+- **Source:** PT-4412.
--- a/.context/standards/Architecture-Decisions.md
+++ b/.context/standards/Architecture-Decisions.md
@@ -2257,11 +2257,14 @@ step, no automation. Just a record.
   `extensions/node_modules/`, which makes it the default for **every** package under
   `extensions/src/*`, not just this one — `import/no-extraneous-dependencies` is off, so the next
   extension to import `@xmldom/xmldom` binds to 0.8.x with no warning. Two follow-ups worth taking
-  together: `convertMarbleChapterXml` runs only in a web view, which already has a native
-  `DOMParser`, so dropping the dependency there would remove the pin, the version split, and this
+  together: PT-4445 migrates `marble-converter` to 0.9's `onError` and drops the pin, and PT-4446
+  supersedes it — `convertMarbleChapterXml` runs only in a web view, which already has a native
+  `DOMParser`, so dropping the dependency there removes the pin, the version split, and this
   paragraph at once. Separately, `@xmldom/xmldom` emits no `parsererror` element, so PR #541's
   malformed-XML guard is inert under the shim; fatally malformed USX still throws (as a `ParseError`
   from inside `parseFromString`, not the renderer's `Invalid USX:` error), but *recoverable* defects
   are silently repaired rather than rejected — on the extension host that repaired form is what gets
-  written back to disk. Revisit if the package ever ships its own Node fallback.
-- **Source:** PT-4412.
+  written back to disk. That write-path asymmetry is **PT-4444**, flagged at the save site in
+  `platform-scripture-extender-pdpe.model.ts`. Revisit if the package ever ships its own Node
+  fallback.
+- **Source:** PT-4412, review of #2714.
--- a/extensions/src/platform-enhanced-resources/package.json
+++ b/extensions/src/platform-enhanced-resources/package.json
@@ -37,6 +37,7 @@
   },
   "dependencies": {
     "@sillsdev/scripture": "^2.0.5",
+    "@xmldom/xmldom": "^0.8.10",
     "platform-bible-utils": "file:../../../lib/platform-bible-utils"
   },
   "devDependencies": {
--- a/extensions/src/platform-enhanced-resources/src/lib/marble-converter.ts
+++ b/extensions/src/platform-enhanced-resources/src/lib/marble-converter.ts
@@ -5,6 +5,11 @@
 // __unknownAttributes mechanism. xmldom's child-list APIs may surface either
 // undefined or null for "absent", so we coalesce both at boundaries.
 /* eslint-disable no-type-assertion/no-type-assertion, no-null/no-null */
+// This extension pins @xmldom/xmldom to 0.8.x in its package.json because the `errorHandler` object
+// below is 0.8-only — 0.9 throws "errorHandler object is no longer supported, switch to onError!".
+// The repo root declares 0.9.x for the extension host's DOM globals, so without the pin this file
+// would resolve 0.9 and throw on every conversion.
+// TODO: migrate to 0.9's `onError` and drop the pin (follow-up to PT-4412).
 import { DOMParser } from '@xmldom/xmldom';
 import {
   USJ_TYPE,
--- a/extensions/src/platform-enhanced-resources/src/lib/marble-converter.ts
+++ b/extensions/src/platform-enhanced-resources/src/lib/marble-converter.ts
@@ -9,7 +9,9 @@
 // below is 0.8-only — 0.9 throws "errorHandler object is no longer supported, switch to onError!".
 // The repo root declares 0.9.x for the extension host's DOM globals, so without the pin this file
 // would resolve 0.9 and throw on every conversion.
-// TODO: migrate to 0.9's `onError` and drop the pin (follow-up to PT-4412).
+// TODO(PT-4445): migrate to 0.9's `onError` and drop the pin — or PT-4446, which drops the
+// dependency entirely since this converter only ever runs in a web view, which has a native
+// `DOMParser`.
 import { DOMParser } from '@xmldom/xmldom';
 import {
   USJ_TYPE,
--- a/extensions/src/platform-scripture/src/project-data-provider/platform-scripture-extender-pdpe.model.ts
+++ b/extensions/src/platform-scripture/src/project-data-provider/platform-scripture-extender-pdpe.model.ts
@@ -127,6 +127,13 @@ export class ScriptureExtenderProjectDataProviderEngine
     });
   }
 
+  // A read-modify-write round-trips through `usxStringToUsj` in the getters below and back out
+  // through the two setters here. On the extension host that parse runs on `@xmldom/xmldom` (see
+  // `adr-node-dom-globals-polyfill`), which silently repairs recoverable USX defects rather than
+  // rejecting them — the renderer's native `DOMParser` rejects the same input — so what these
+  // setters persist is the repaired form, not what was on disk.
+  // TODO(PT-4444): make the two processes agree on what USX is valid, then drop this note.
+
   // Do not emit update events when running this method because we are subscribing to data updates
   // and sending out update events in the constructor
   @papi.dataProviders.decorators.doNotNotify
--- a/extensions/vitest.config.ts
+++ b/extensions/vitest.config.ts
@@ -9,7 +9,10 @@ const config = defineConfig({
     environment: 'node',
     // Warm the lazy one-time ICU init behind Intl.* so it never lands inside a test's timeout
     // window on a slow CI worker. Shares the repo-root setup file. See vitest.setup.ts for rationale.
-    setupFiles: [path.resolve(__dirname, '../vitest.setup.ts')],
+    setupFiles: [
+      path.resolve(__dirname, '../vitest.setup.ts'),
+      path.resolve(__dirname, '../vitest.setup.node.ts'),
+    ],
     include: ['src/**/*.test.ts', 'src/**/*.test.tsx', 'lib/**/*.test.ts', 'lib/**/*.test.tsx'],
   },
   resolve: {
--- a/lib/platform-bible-utils/vite.config.ts
+++ b/lib/platform-bible-utils/vite.config.ts
@@ -43,7 +43,10 @@ const config = defineConfig({
     // Warm the lazy one-time ICU init behind Intl.* so it never lands inside a test's timeout window
     // on a slow CI worker. This workspace's intl/* tests construct Intl.NumberFormat/DateTimeFormat/
     // Collator directly. Shares the repo-root setup file. See vitest.setup.ts for rationale.
-    setupFiles: [path.resolve(__dirname, '../../vitest.setup.ts')],
+    setupFiles: [
+      path.resolve(__dirname, '../../vitest.setup.ts'),
+      path.resolve(__dirname, '../../vitest.setup.node.ts'),
+    ],
   },
 });
 export default config;
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,5 +1,5 @@
 {
-  "name": "paranext-core",
+  "name": "pt-4412-dom-globals",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
@@ -13,6 +13,7 @@
       ],
       "dependencies": {
         "@sillsdev/scripture": "^2.0.5",
+        "@xmldom/xmldom": "^0.9.8",
         "ajv": "^8.18.0",
         "chalk": "^4.1.2",
         "chokidar": "^3.6.0",
@@ -255,6 +256,15 @@
       "license": "MIT",
       "dependencies": {
         "undici-types": "~6.21.0"
+      }
+    },
+    "extensions/node_modules/@xmldom/xmldom": {
+      "version": "0.8.15",
+      "resolved": "https://registry.npmjs.org/@xmldom/xmldom/-/xmldom-0.8.15.tgz",
+      "integrity": "sha512-/5NV/vDALVFDXgLmfsy9TRCBlKwO2LNBFzpzvb9iIj+jR+eSc6DLYYvVOdivT/jm7MtU6TebYuRmzEOI7w40UA==",
+      "license": "MIT",
+      "engines": {
+        "node": ">=10.0.0"
       }
     },
     "extensions/node_modules/arg": {
@@ -647,6 +657,7 @@
       "license": "MIT",
       "dependencies": {
         "@sillsdev/scripture": "^2.0.5",
+        "@xmldom/xmldom": "^0.8.10",
         "platform-bible-utils": "file:../../../lib/platform-bible-utils"
       },
       "devDependencies": {
@@ -4281,15 +4292,6 @@
       "license": "MIT",
       "dependencies": {
         "@xmldom/xmldom": "^0.9.8"
-      }
-    },
-    "node_modules/@eten-tech-foundation/scripture-utilities/node_modules/@xmldom/xmldom": {
-      "version": "0.9.10",
-      "resolved": "https://registry.npmjs.org/@xmldom/xmldom/-/xmldom-0.9.10.tgz",
-      "integrity": "sha512-A9gOqLdi6cV4ibazAjcQufGj0B1y/vDqYrcuP6d/6x8P27gRS8643Dj9o1dEKtB6O7fwxb2FgBmJS2mX7gpvdw==",
-      "license": "MIT",
-      "engines": {
-        "node": ">=14.6"
       }
     },
     "node_modules/@floating-ui/core": {
@@ -12296,11 +12298,12 @@
       }
     },
     "node_modules/@xmldom/xmldom": {
-      "version": "0.8.13",
-      "dev": true,
+      "version": "0.9.12",
+      "resolved": "https://registry.npmjs.org/@xmldom/xmldom/-/xmldom-0.9.12.tgz",
+      "integrity": "sha512-5AXjrcMClTryPe9LgZrygpB1lj7s0S9E0+W+AHaVKAVyHanafK86iPSvG5xHVSp/jC+VH1UXu0TAEmY279xH7A==",
       "license": "MIT",
       "engines": {
-        "node": ">=10.0.0"
+        "node": ">=14.6"
       }
     },
     "node_modules/@xtuc/ieee754": {
@@ -24971,6 +24974,16 @@
       },
       "engines": {
         "node": ">=10.4.0"
+      }
+    },
+    "node_modules/plist/node_modules/@xmldom/xmldom": {
+      "version": "0.8.15",
+      "resolved": "https://registry.npmjs.org/@xmldom/xmldom/-/xmldom-0.8.15.tgz",
+      "integrity": "sha512-/5NV/vDALVFDXgLmfsy9TRCBlKwO2LNBFzpzvb9iIj+jR+eSc6DLYYvVOdivT/jm7MtU6TebYuRmzEOI7w40UA==",
+      "dev": true,
+      "license": "MIT",
+      "engines": {
+        "node": ">=10.0.0"
       }
     },
     "node_modules/possible-typed-array-names": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,5 +1,5 @@
 {
-  "name": "pt-4412-dom-globals",
+  "name": "paranext-core",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
--- a/package.json
+++ b/package.json
@@ -128,6 +128,7 @@
   "browserslist": ["extends browserslist-config-detect-electron"],
   "dependencies": {
     "@sillsdev/scripture": "^2.0.5",
+    "@xmldom/xmldom": "^0.9.8",
     "ajv": "^8.18.0",
     "chalk": "^4.1.2",
     "chokidar": "^3.6.0",
--- a/src/extension-host/extension-host.ts
+++ b/src/extension-host/extension-host.ts
@@ -1,3 +1,5 @@
+// Keep first: installs DOM globals before any import can convert scripture. See the module's docs.
+import '@node/polyfills/dom-globals.polyfill';
 import '@extension-host/global-this.model';
 import '@node/utils/log-archiver.util';
 import { isClient } from '@shared/utils/internal-util';
--- a/src/node/polyfills/dom-globals.polyfill.test.ts
+++ b/src/node/polyfills/dom-globals.polyfill.test.ts
@@ -60,6 +60,32 @@ function firstEvaluatedSpecifier(source: string): string | undefined {
   return evaluated.sort((a, b) => a.index - b.index)[0]?.specifier;
 }
 
+/**
+ * Whether the installed `@eten-tech-foundation/scripture-utilities` delegates its XML work to the
+ * platform's DOM globals — true from the release that drops the package's own bundled
+ * `@xmldom/xmldom` (scripture-editors PR #541) onward, false before it, where the converters run
+ * fine in bare Node and nothing this polyfill installs is reachable.
+ *
+ * Probing beats reading the version because the behavior, not the range, is what the polyfill
+ * exists for.
+ */
+function convertersNeedDomGlobals(): boolean {
+  const saved = DOM_GLOBAL_NAMES.map((name) => [name, Reflect.get(globalThis, name)] as const);
+  DOM_GLOBAL_NAMES.forEach((name) => Reflect.deleteProperty(globalThis, name));
+  try {
+    usxStringToUsj('<usx version="3.0"><book code="GEN" style="id">Genesis</book></usx>');
+    return false;
+  } catch {
+    return true;
+  } finally {
+    saved.forEach(([name, value]) => {
+      if (value !== undefined) Reflect.set(globalThis, name, value);
+    });
+  }
+}
+
+const CONVERTERS_NEED_DOM_GLOBALS = convertersNeedDomGlobals();
+
 beforeEach(() => {
   savedGlobals = Object.fromEntries(
     DOM_GLOBAL_NAMES.map((name) => [name, Reflect.get(globalThis, name)]),
@@ -96,9 +122,12 @@ describe('dom-globals.polyfill', () => {
     expect(serialized).toContain('Genesis');
   });
 
-  // Asserting the globals exist is not the same as asserting the package accepts them: it gates on
-  // its own `assertDomEnvironment` and could need more of the environment than we install. This
-  // drives the exact two functions the extension host calls on every scripture read and write.
+  // Asserting the globals exist is not the same as asserting the package accepts them: once it
+  // gates on its own `assertDomEnvironment` it could need more of the environment than we install.
+  // This drives the exact two functions the extension host calls on every scripture read and write.
+  //
+  // While `scripture-utilities` still bundles its own `@xmldom/xmldom` this passes with or without
+  // the polyfill — see the skipped test below, which is what makes the difference detectable.
   test('the real USX⇔USJ converters round-trip under the installed globals', async () => {
     await importPolyfillFresh();
     const usx =
@@ -117,6 +146,19 @@ describe('dom-globals.polyfill', () => {
     // would drift a little further on every save.
     expect(usxStringToUsj(reserialized)).toEqual(usj);
   });
+
+  // The falsifying half of the round-trip test above: without it, deleting the polyfill outright
+  // still leaves this file green. It reports as skipped — not passed — for as long as the installed
+  // `scripture-utilities` bundles its own XML implementation, so nobody reads green CI as proof the
+  // polyfill is load-bearing. It starts failing the moment the polyfill stops working.
+  test.skipIf(!CONVERTERS_NEED_DOM_GLOBALS)(
+    'the converters fail without the globals, so the polyfill is what makes them work',
+    () => {
+      expect(() =>
+        usxStringToUsj('<usx version="3.0"><book code="GEN" style="id">Genesis</book></usx>'),
+      ).toThrow(/DOM environment/);
+    },
+  );
 
   test('does not override a real DOM (renderer, jsdom tests)', async () => {
     // Stand-ins for a real DOM's constructors — identity is all this test cares about
--- a/src/node/polyfills/dom-globals.polyfill.test.ts
+++ b/src/node/polyfills/dom-globals.polyfill.test.ts
@@ -20,11 +20,14 @@ async function importPolyfillFresh(): Promise<void> {
 }
 
 /**
- * Whether an `import` clause is erased by the compiler, so it loads nothing at runtime. True for
- * `import type { … }` and for a clause whose every named binding is individually `type`-prefixed.
+ * Whether an `import`/`export` clause is erased by the compiler, so it loads nothing at runtime.
+ * True for `import type { … }` and for a clause whose every named binding is individually
+ * `type`-prefixed.
  *
  * `extensions/src/platform-scripture-editor/src/extension-host-import-boundary.test.ts` needs the
- * same parsing for the same reason; the two cannot share a helper across the workspace boundary.
+ * same parsing for the same reason; the two cannot share a helper across the workspace boundary, so
+ * this function and the specifier patterns below are a deliberate copy of that file's. Fix a
+ * parsing bug in one and fix it in the other.
  */
 function isTypeOnlyClause(clause: string): boolean {
   const trimmed = clause.trim();
@@ -40,21 +43,21 @@ function isTypeOnlyClause(clause: string): boolean {
 
 /**
  * The specifier of the first module `source` evaluates at load time — a side-effect import, a value
- * `import … from`, or a `require`. Type-only clauses are skipped because the compiler erases them,
- * so they cannot run anything.
+ * `import`/`export … from`, or a dynamic `import()`/`require()`. Type-only clauses are skipped
+ * because the compiler erases them, so they cannot run anything.
  */
 function firstEvaluatedSpecifier(source: string): string | undefined {
   const evaluated: { index: number; specifier: string }[] = [];
 
   const sideEffectImport = /(?:^|\n)[ \t]*import[ \t]*['"]([^'"]+)['"]/g;
-  const fromImport = /(?:^|\n)[ \t]*import\b([^;'"]*?)from[ \t]*['"]([^'"]+)['"]/g;
-  const runtimeRequire = /\brequire\([ \t]*['"]([^'"]+)['"]/g;
+  const fromImport = /(?:^|\n)[ \t]*(?:import|export)\b([^;'"]*?)from[ \t]*['"]([^'"]+)['"]/g;
+  const runtimeImport = /\b(?:import|require)\([ \t]*['"]([^'"]+)['"]/g;
 
   for (let match = sideEffectImport.exec(source); match; match = sideEffectImport.exec(source))
     evaluated.push({ index: match.index, specifier: match[1] });
   for (let match = fromImport.exec(source); match; match = fromImport.exec(source))
     if (!isTypeOnlyClause(match[1])) evaluated.push({ index: match.index, specifier: match[2] });
-  for (let match = runtimeRequire.exec(source); match; match = runtimeRequire.exec(source))
+  for (let match = runtimeImport.exec(source); match; match = runtimeImport.exec(source))
     evaluated.push({ index: match.index, specifier: match[1] });
 
   return evaluated.sort((a, b) => a.index - b.index)[0]?.specifier;
--- a/src/node/polyfills/dom-globals.polyfill.test.ts
+++ b/src/node/polyfills/dom-globals.polyfill.test.ts
@@ -1,0 +1,145 @@
+// @vitest-environment node
+// Node, not the repo-default jsdom, on purpose: the extension host is plain Node, and the
+// missing-DOM case this polyfill exists for only reproduces without jsdom's DOM globals.
+
+import { readFileSync } from 'node:fs';
+import { resolve } from 'node:path';
+import { usjToUsxString, usxStringToUsj } from '@eten-tech-foundation/scripture-utilities';
+import { afterEach, beforeEach, describe, expect, test, vi } from 'vitest';
+
+// These tests read and write the DOM globals through `Reflect` because TypeScript declares both as
+// always-present, which rules out `delete` and rejects assigning a stand-in.
+const DOM_GLOBAL_NAMES = ['DOMParser', 'XMLSerializer'] as const;
+
+let savedGlobals: Record<string, unknown> = {};
+
+/** Re-runs the polyfill's module body, which is where it installs the globals */
+async function importPolyfillFresh(): Promise<void> {
+  vi.resetModules();
+  await import('./dom-globals.polyfill');
+}
+
+/**
+ * Whether an `import` clause is erased by the compiler, so it loads nothing at runtime. True for
+ * `import type { … }` and for a clause whose every named binding is individually `type`-prefixed.
+ *
+ * `extensions/src/platform-scripture-editor/src/extension-host-import-boundary.test.ts` needs the
+ * same parsing for the same reason; the two cannot share a helper across the workspace boundary.
+ */
+function isTypeOnlyClause(clause: string): boolean {
+  const trimmed = clause.trim();
+  if (/^type\b/.test(trimmed)) return true;
+  const namedBindings = /^\{([^}]*)\}$/.exec(trimmed);
+  if (!namedBindings) return false;
+  const bindings = namedBindings[1]
+    .split(',')
+    .map((binding) => binding.trim())
+    .filter((binding) => binding.length > 0);
+  return bindings.length > 0 && bindings.every((binding) => /^type\b/.test(binding));
+}
+
+/**
+ * The specifier of the first module `source` evaluates at load time — a side-effect import, a value
+ * `import … from`, or a `require`. Type-only clauses are skipped because the compiler erases them,
+ * so they cannot run anything.
+ */
+function firstEvaluatedSpecifier(source: string): string | undefined {
+  const evaluated: { index: number; specifier: string }[] = [];
+
+  const sideEffectImport = /(?:^|\n)[ \t]*import[ \t]*['"]([^'"]+)['"]/g;
+  const fromImport = /(?:^|\n)[ \t]*import\b([^;'"]*?)from[ \t]*['"]([^'"]+)['"]/g;
+  const runtimeRequire = /\brequire\([ \t]*['"]([^'"]+)['"]/g;
+
+  for (let match = sideEffectImport.exec(source); match; match = sideEffectImport.exec(source))
+    evaluated.push({ index: match.index, specifier: match[1] });
+  for (let match = fromImport.exec(source); match; match = fromImport.exec(source))
+    if (!isTypeOnlyClause(match[1])) evaluated.push({ index: match.index, specifier: match[2] });
+  for (let match = runtimeRequire.exec(source); match; match = runtimeRequire.exec(source))
+    evaluated.push({ index: match.index, specifier: match[1] });
+
+  return evaluated.sort((a, b) => a.index - b.index)[0]?.specifier;
+}
+
+beforeEach(() => {
+  savedGlobals = Object.fromEntries(
+    DOM_GLOBAL_NAMES.map((name) => [name, Reflect.get(globalThis, name)]),
+  );
+  DOM_GLOBAL_NAMES.forEach((name) => Reflect.deleteProperty(globalThis, name));
+});
+
+afterEach(() => {
+  DOM_GLOBAL_NAMES.forEach((name) => {
+    if (savedGlobals[name] === undefined) Reflect.deleteProperty(globalThis, name);
+    else Reflect.set(globalThis, name, savedGlobals[name]);
+  });
+});
+
+describe('dom-globals.polyfill', () => {
+  test('installs DOMParser and XMLSerializer when Node provides neither', async () => {
+    expect(Reflect.get(globalThis, 'DOMParser')).toBeUndefined();
+    expect(Reflect.get(globalThis, 'XMLSerializer')).toBeUndefined();
+
+    await importPolyfillFresh();
+
+    expect(Reflect.get(globalThis, 'DOMParser')).toBeDefined();
+    expect(Reflect.get(globalThis, 'XMLSerializer')).toBeDefined();
+  });
+
+  test('installed globals round-trip USX', async () => {
+    await importPolyfillFresh();
+
+    const usx = '<usx version="3.0"><book code="GEN" style="id">Genesis</book></usx>';
+    const doc = new globalThis.DOMParser().parseFromString(usx, 'text/xml');
+    const serialized = new globalThis.XMLSerializer().serializeToString(doc);
+
+    expect(serialized).toContain('code="GEN"');
+    expect(serialized).toContain('Genesis');
+  });
+
+  // Asserting the globals exist is not the same as asserting the package accepts them: it gates on
+  // its own `assertDomEnvironment` and could need more of the environment than we install. This
+  // drives the exact two functions the extension host calls on every scripture read and write.
+  test('the real USX⇔USJ converters round-trip under the installed globals', async () => {
+    await importPolyfillFresh();
+    const usx =
+      '<?xml version="1.0" encoding="utf-8"?>' +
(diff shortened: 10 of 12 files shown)


irahopkinson (changes requested):
Looking good. I'll drop an AI review next.

> @irahopkinson reviewed 11 files and all commit messages, and made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2714#-:-P-qcfMJAOotPz-g8nld:b4sxv20)* status: all files reviewed, 2 unresolved discussions (waiting on katherinejensen00).

___
*[`-- commits` line 33 at r1](https://reviewable.io/reviews/paranext/paranext-core/2714#-P-q_qP52r1eRhLHQ-jI:-P-q_qP52r1eRhLHQ-jJ:bxxihvj):*
This is out-of-date in the commit, update to `ADR-0024`.



_Code quote:_
```
ADR-0019
```
___
*[`package-lock.json` line 2 at r1](https://reviewable.io/reviews/paranext/paranext-core/2714#-P-qaAkd9a5YMl0wWv6n:-P-qaAkd9a5YMl0wWv6o:bhyw88) ([raw file](https://github.com/paranext/paranext-core/blob/32eb603c3c61c6a562ae080cd69197d135e7e659/package-lock.json#L2)):*
> ```JSON
> {
>   "name": "pt-4412-dom-globals",
> ```

This is the wrong name.



_Code quote:_
```JSON
pt-4412-dom-globals
```

irahopkinson (changes requested):
> @irahopkinson+AI made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2714)* status: all files reviewed, 5 unresolved discussions (waiting on irahopkinson and katherinejensen00).

___
*[`package-lock.json` line 2 at r1](https://reviewable.io/reviews/paranext/paranext-core/2714#-P-qaAkd9a5YMl0wWv6n:-P-qlCyf5LguoDJqgOo9:bkg6bks) ([raw file](https://github.com/paranext/paranext-core/blob/32eb603c3c61c6a562ae080cd69197d135e7e659/package-lock.json#L2)):*
<details><summary><i>Previously, irahopkinson (Ira Hopkinson) wrote…</i></summary><blockquote>

This is the wrong name.
</blockquote></details>

Adding the root cause, since this will otherwise come back.

The root `package.json` has **no `name` field**, so npm derives the lockfile's top-level `name` from the checkout directory basename. That's why the worktree rename described in the ticket comment (the `IgnorePlugin` workaround) leaked in here — the install ran in a directory called `pt-4412-dom-globals`.

So it isn't a one-off typo: it will churn back to whatever directory name the next contributor happens to install from. Reverting this line fixes the PR, but adding `"name": "paranext-core"` to `package.json` would stop it recurring.
___
*[`.context/standards/Architecture-Decisions.md` line 1117 at r1](https://reviewable.io/reviews/paranext/paranext-core/2714#-P-ql8aW81Ns7cuONBQj:-P-ql8aW81Ns7cuONBQk:b-vmfvmi) ([raw file](https://github.com/paranext/paranext-core/blob/32eb603c3c61c6a562ae080cd69197d135e7e659/.context/standards/Architecture-Decisions.md#L1117)):*
This paragraph documents a write-path data-integrity risk that nothing tracks.

> *recoverable* defects are silently repaired rather than rejected — on the extension host that repaired form is what gets written back to disk.

The ticket's "Correction" note retracts only the **fatal** case (0.9.12 throws on tag mismatch), then concludes *"No silent-corruption follow-up is needed"* — but it never addresses recoverable defects, which is what this sentence is about. So the ADR and the ticket are consistent on the facts and inconsistent on the conclusion; the ADR is the one that's right.

Net effect as written: the renderer **rejects** malformed USX, the extension host **repairs it and writes the repaired form to disk**. That asymmetry is worth more than a paragraph in an ADR that closes with this PR:

- a filed follow-up ticket, and
- a `TODO(PT-XXXX)` at the write site (`platform-scripture-extender-pdpe.model.ts:139,154`), where someone touching the save path would actually see it.

I searched Jira and no such ticket exists yet. Happy for the fix to be "file the ticket and reference it" rather than anything in code here.
___
*[`src/node/polyfills/dom-globals.polyfill.test.ts` line 102 at r1](https://reviewable.io/reviews/paranext/paranext-core/2714#-P-ql6rr0pOs9uEUKAu5:-P-ql6rs40SUH0V7FP4i:bk1o6gt) ([raw file](https://github.com/paranext/paranext-core/blob/32eb603c3c61c6a562ae080cd69197d135e7e659/src/node/polyfills/dom-globals.polyfill.test.ts#L102)):*
This test can't fail — it passes with the polyfill deleted.

The comment above it says it gates on the package's own check:

> Asserting the globals exist is not the same as asserting the package accepts them: it gates on its own `assertDomEnvironment`

At the declared `@eten-tech-foundation/scripture-utilities@~0.1.6` it doesn't. I checked the installed package:

- `0.1.6` declares `@xmldom/xmldom: ^0.9.10` as its **own** dependency
- `dist/index.cjs` imports xmldom directly
- the string `requires a DOM environment` is **absent** from `dist`

Running both converters in bare Node with no globals installed:

```
globalThis.DOMParser     = undefined
globalThis.XMLSerializer = undefined
ROUND-TRIP OK with NO DOM globals -> true
```

So the one test that exists to prove the polyfill works cannot detect it being removed, reordered, or regressing. Combined with the fact that the DoD item *"Scripture read/write round-trips with no 'requires a DOM environment' error"* rests entirely on the manual yalc session in the ticket, there is currently nothing in CI that would catch this breaking before the bump lands.

Two options that would both work for me:

1. Add a test that deletes the globals and asserts the converters throw — it will be inert now but becomes a real guard the moment the version bump lands.
2. Or, minimally, reword the comment to say the assertion is inert until the bump, so nobody reads green CI as coverage.

Non-blocking on the approach; blocking on it not being silently misleading.
___
*[`extensions/src/platform-enhanced-resources/src/lib/marble-converter.ts` line 12 at r1](https://reviewable.io/reviews/paranext/paranext-core/2714#-P-qlCezB4cijj46SuDC:-P-qlCezB4cijj46SuDD:bp6g98e) ([raw file](https://github.com/paranext/paranext-core/blob/32eb603c3c61c6a562ae080cd69197d135e7e659/extensions/src/platform-enhanced-resources/src/lib/marble-converter.ts#L12)):*
> ```XML
> // The repo root declares 0.9.x for the extension host's DOM globals, so without the pin this file
> // would resolve 0.9 and throw on every conversion.
> // TODO: migrate to 0.9's `onError` and drop the pin (follow-up to PT-4412).
> ```

This TODO points at a ticket that closes when this PR merges, so it points at nothing.

```ts
// TODO: migrate to 0.9's `onError` and drop the pin (follow-up to PT-4412).
```

`.claude/rules/code-quality/forward-facing-comments.md` wants a TODO to name an **open** ticket (`TODO(PT-1234): …`) and explicitly lists *"ticket/PR references for work done in this PR"* as the backward-facing form to cut.

I searched Jira for the follow-ups this PR names and **none of the three exist**:

1. migrate `marble-converter` to `onError` and drop the 0.8 pin (this comment + ADR-0024)
2. drop `@xmldom/xmldom` from `platform-enhanced-resources` entirely — ADR-0024 notes `convertMarbleChapterXml` runs only in a web view, which already has a native `DOMParser`, so this would remove the pin, the version split, and that whole ADR paragraph at once
3. the `IgnorePlugin` `checkResource` substring bug you hit and worked around during verification — the ticket comment says it's *"worth its own ticket"*, and it's a nasty one: any checkout whose path contains `main`, `node`, `renderer`, `client`, or `extension-host` as a substring silently produces a broken renderer with no build error

Could you file them and swap this comment to `TODO(PT-XXXX)`? The three lines above it are genuinely forward-facing (a real constraint and gotcha) — those should stay exactly as they are.

katherinejensen00:
Thanks both — all five discussions addressed in `688b748` (plus a message-only amend of `32eb603`). Four follow-up tickets filed.

### `-- commits` line 33 — `ADR-0019` → `ADR-0024`
Fixed by amending the original commit's message.

### `package-lock.json` line 2 — wrong name
Fixed at the root cause, as suggested. The root `package.json` now declares `"name": "paranext-core"`, so npm no longer derives the lockfile name from the checkout directory basename. `package-lock.json` is back to `paranext-core` and now also carries `"name"` in `packages[""]`, which is what npm writes once the field exists.

Checked before doing it that nothing keys off the root name: `electron-builder.json5` sets `productName`/`appId` explicitly and builds from `release/app/package.json`, which has its own name. Regenerating the lock with `npm i --package-lock-only` also wanted to add ~65 lines of unrelated `@tailwindcss/oxide-wasm32-wasi` bundled-dep entries, so I applied the two name lines surgically and left that churn for whoever next runs a real install.

### `dom-globals.polyfill.test.ts` line 102 — the test can't fail
You're right, and thanks for actually running it. Took option 1, but the assertion can't just be added unconditionally: at `~0.1.6` the converters *succeed* without globals, so `expect(...).toThrow()` would fail today rather than be inert.

So the test now probes which regime it's in and gates on it:

```ts
function convertersNeedDomGlobals(): boolean { /* delete globals, try a convert, restore */ }

test.skipIf(!CONVERTERS_NEED_DOM_GLOBALS)(
  'the converters fail without the globals, so the polyfill is what makes them work',
  () => { expect(() => usxStringToUsj(...)).toThrow(/DOM environment/); },
);
```

It reports as **skipped**, not passed, for as long as the installed package bundles its own XML implementation — so nobody reads green CI as coverage — and becomes a real guard the moment the bump lands. Probing rather than reading the version range because the behavior is what the polyfill exists for. The round-trip test's comment now says the same thing out loud instead of implying coverage it doesn't have.

Verified locally: `6 tests | 1 skipped`, with the skip being exactly this one.

### ADR line 1117 — untracked write-path data-integrity risk
Agreed the ADR is the one that's right; the ticket's "Correction" note retracted only the fatal case. I re-ran it against **0.9.12** to pin down the recoverable half:

| Input | Result | `parsererror` count |
| --- | --- | --- |
| `<para style="p">Tom & Jerry</para>` | repaired to `Tom &amp; Jerry` | 0 |
| `<usx version=3.0>` | repaired to `version="3.0"`, warning logged | 0 |
| `a&foo;b` | repaired to `a&amp;foo;b`, error logged | 0 |
| `a < b` | repaired to `a &lt; b`, error logged | 0 |
| `<para style="p" style="q">` | throws `Attribute style redefined` | — |

A browser `DOMParser` produces a `parsererror` for the first four, so the renderer rejects exactly the inputs the extension host quietly rewrites. Filed as **PT-4444**, with a `TODO(PT-4444)` at the save site in `platform-scripture-extender-pdpe.model.ts` and the ticket named in ADR-0024. I also corrected the "no silent-corruption follow-up is needed" note on PT-4412.

### `marble-converter.ts` line 12 — TODO points at a closing ticket
All three filed:

| | |
| --- | --- |
| **PT-4445** | Migrate `marble-converter` to 0.9's `onError` and drop the 0.8 pin |
| **PT-4446** | Drop `@xmldom/xmldom` from `platform-enhanced-resources` entirely — supersedes PT-4445 |
| **PT-4447** | `IgnorePlugin` `checkResource` substring bug (filed as a Bug, with repro) |

The comment is now `TODO(PT-4445)`, naming PT-4446 as the larger option. The three lines above it are unchanged, as you asked. ADR-0024 names PT-4445/4446 too.

On PT-4446: confirmed `convertMarbleChapterXml`'s only production call site is `enhanced-resource.web-view.tsx:1281` — the others are a story and its own test — so your read is right, the web view already has the `DOMParser` it needs. Noted in the ticket that `marble-converter.test.ts` runs in a `node` environment and would need `// @vitest-environment jsdom` or the polyfill's globals.

---

Still not green locally for the reason in the PR description — the `yalc` link for `@eten-tech-foundation/*` has no `dist/`. Every remaining `typecheck`/`lint` failure is `Cannot find module '@eten-tech-foundation/…'`, none in changed files. The polyfill test itself does run, because that package's `exports` has a `development` condition pointing at `src/`.

katherinejensen00:
Rebased onto `main` (`7f9bf49`).

**ADR renumbered `0024` → `0027`.** `main` picked up three ADRs while this was open — `0024` (toolbar sync status), `0025` (Find book lists), `0026` (Find selection) — and `main` also added the rule for exactly this case: *"Numbers are claimed at merge, not at write… Whoever merges second does the renumbering."* Updated the heading, the reference in `platform-scripture-extender-pdpe.model.ts`, both commit messages, and the PR description above. The other `ADR-0024` references in the tree belong to the sync-status ADR and are untouched.

The only real conflict was the append-at-the-end collision in `Architecture-Decisions.md`; `package.json` and `package-lock.json` auto-merged. Post-rebase check: polyfill tests still `5 passed | 1 skipped`, prettier clean, and `typecheck` reports nothing outside the known `@eten-tech-foundation/*` yalc resolution failures.

One thing worth a separate PR at some point: `npm i --package-lock-only` on `main` wants to add ~64 lines of `@tailwindcss/oxide-wasm32-wasi` bundled-dep entries. That drift predates this branch, so I left it out rather than smuggling it into this diff.

katherinejensen00:
Rebased again onto `main` (`cbecad7`) — and the ADR collided **again**. `main` had meanwhile taken `ADR-0027` for "Panel readiness is derived from whether data sources arrived" and `ADR-0028` for the async-hook state shape. This entry has now been `0019` → `0024` → `0027` across three rebases, renumbered every time because `main` claimed the number first.

So rather than renumber a fourth time, **new ADR entries are now identified by a kebab-case slug**:

```markdown
## ADR-node-dom-globals-polyfill: Node processes install `@xmldom/xmldom` DOM globals; …
```

The convention section and the entry template are updated to match. The rationale is the same one `main`'s own "Numbers are claimed at merge, not at write" bullet identified — a sequential number is only unique at merge, and an append-only file has nothing to catch a collision — but a slug is drawn from the decision itself, so it is unique at *write* time and never needs renumbering. That bullet is replaced by the naming rule; the "whoever merges second renumbers" chore is gone.

**Existing `ADR-NNNN` entries keep their numbers.** Renaming 28 headings would rewrite ~50 internal cross-references and `ADR-NNNN` mentions in 15 other source files — for no gain, against this log's own "don't rewrite history" rule, and it would conflict with every branch in flight. So the file will carry both forms; that's stated explicitly in the conventions.

Happy to split the convention change into its own PR if you'd rather review it separately from the polyfill — say the word and I'll lift those two hunks out.

Everything else is unchanged: polyfill tests still `5 passed | 1 skipped`, prettier clean, `typecheck` clean outside the known `@eten-tech-foundation/*` yalc failures. PR description, `platform-scripture-extender-pdpe.model.ts`, both commit messages, and PT-4444/4445/4446 all updated to the slug.

katherinejensen00:
CI was red on all three platforms and it was my fault — pushed a fix in `3545204`.

**What broke.** `npm run lint` failed with 12 `import/no-relative-packages` errors, all in files this PR never touches:

```
.erb/configs/webpack.config.base.ts
  6:35  Relative import from another package is not allowed.
        Use `platform-bible/package.json` instead of `../../release/app/package.json`
```

...plus `bump-versions.ts`, `electron-rebuild.js`, both `.storybook/` files, `app.service-host.ts`, `about-dialog.component.tsx`, `ten-layout-shared.tsx`, and `tailwind.config.ts`.

**Cause: the `"name": "paranext-core"` I added to the root `package.json`** — the root-cause fix suggested in the lockfile thread. Naming the root turns it into a *package* as far as `eslint-plugin-import` is concerned, and `import/no-relative-packages` then rejects every existing relative import that crosses out of it into `release/app`, `lib/platform-bible-react`, or `extensions`. While the root was anonymous the rule had nothing to anchor on and stayed quiet. Confirmed both directions locally: with the field, those 12 errors; without it, clean.

**So the suggestion can't be taken in isolation, and I've reverted the field** — `package.json` is now identical to `main`, and the PR keeps only the part you originally asked for: `package-lock.json`'s name restored to `paranext-core`.

Worth knowing before anyone retries it: the rule's own suggested replacements don't resolve either. `release/app` is **not** a workspace (`"workspaces": ["lib/*", "extensions", "extensions/src/*"]`), so there is no `node_modules/platform-bible` and `import platform-bible/package.json` would fail at runtime. Naming the root therefore needs one of:

1. rewriting those 12 imports *and* making `release/app` resolvable — a workspace-layout change; or
2. configuring/relaxing `import/no-relative-packages` for the root package.

Either is its own piece of work and neither belongs in a DOM-globals PR, so the lockfile name will still churn to the installing directory's basename until then. Want me to file that as a ticket? Happy to, just didn't want to open a fifth one unasked.

Verified before pushing: repo-wide `eslint --ext .cjs,.js,.jsx,.ts,.tsx --no-cache .` now reports only the 3 known `@eten-tech-foundation/*` yalc resolution errors (one of them in `ten-layout-shared.tsx`, which I don't touch — environmental, not from this branch). Polyfill tests still `5 passed | 1 skipped`.

Apologies for the red build — I'd verified prettier, `tsc`, and the polyfill tests but ran `eslint` only over changed files, so a rule that fires on *unchanged* files from a root-level config change was exactly the blind spot. Running the repo-wide lint is what would have caught it, and it's what I ran this time.

katherinejensen00 (commented):
> @katherinejensen00 made 5 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2714)* status: 6 of 12 files reviewed, 5 unresolved discussions (waiting on irahopkinson and irahopkinson+AI).

___
*[`-- commits` line 33 at r1](https://reviewable.io/reviews/paranext/paranext-core/2714#-P-q_qP52r1eRhLHQ-jI:-P03ec3663xmJi30A5SS:b-fhsl8d):*
<details><summary><i>Previously, irahopkinson (Ira Hopkinson) wrote…</i></summary><blockquote>

This is out-of-date in the commit, update to `ADR-0024`.
</blockquote></details>

Good catch! Fixed, and then the whole problem went away.

This entry got renumbered on every rebase — 0019 to 0024 to 0027 — because main kept claiming the number first. Meanwhile main itself replaced numbering entirely: cd5e723 renamed all 28 existing entries to adr- kebab-case slugs and added a **Formerly:** line to each so old numeric references still resolve.

So this entry is now adr-node-dom-globals-polyfill, conforming to the convention already on main: lowercase prefix, backticked cross-references in code, and no **Formerly:** line, since main's rule says not to add one to a new entry and this one never landed under a number.

The practical upshot for review: Architecture-Decisions.md is now a clean 53-line append at line 1805 with no convention changes of its own. I'd previously offered to split the convention rewrite into its own PR — that's withdrawn, since main did the conversion and my version of those hunks is gone.
___
*[`package-lock.json` line 2 at r1](https://reviewable.io/reviews/paranext/paranext-core/2714#-P-qaAkd9a5YMl0wWv6n:-P08x_CvD3aTMQPmnX3N:bavfs4n) ([raw file](https://github.com/paranext/paranext-core/blob/32eb603c3c61c6a562ae080cd69197d135e7e659/package-lock.json#L2)):*
<details><summary><i>Previously, irahopkinson+AI (Agent AI for Ira Hopkinson) wrote…</i></summary><blockquote>

Adding the root cause, since this will otherwise come back.

The root `package.json` has **no `name` field**, so npm derives the lockfile's top-level `name` from the checkout directory basename. That's why the worktree rename described in the ticket comment (the `IgnorePlugin` workaround) leaked in here — the install ran in a directory called `pt-4412-dom-globals`.

So it isn't a one-off typo: it will churn back to whatever directory name the next contributor happens to install from. Reverting this line fixes the PR, but adding `"name": "paranext-core"` to `package.json` would stop it recurring.
</blockquote></details>

Took the root-cause fix, it broke CI on all three platforms, and I've reverted it. Detail, because it's a trap worth documenting:

Adding "name": "paranext-core" to the root package.json does stop npm deriving the lockfile name from the checkout directory basename. But it also makes eslint-plugin-import treat the root as a package, and import/no-relative-packages then rejects every pre-existing relative import that crosses out of the root into release/app, lib/platform-bible-react, or extensions. That's 12 errors in files this PR never touches: webpack.config.base.ts, bump-versions.ts, electron-rebuild.js, both .storybook/ files, app.service-host.ts, about-dialog.component.tsx, ten-layout-shared.tsx, tailwind.config.ts. While the root was anonymous the rule had nothing to anchor on. Confirmed both directions locally.

The rule's own suggested replacements don't resolve either. release/app is not a workspace, so there is no node_modules/platform-bible and import platform-bible/package.json would fail at runtime.

So naming the root needs either a workspace-layout change plus 12 import rewrites, or an import/no-relative-packages config change — its own piece of work, and not a DOM-globals PR. package.json is now identical to main apart from the one @xmldom/xmldom dependency, and this PR keeps only what you originally asked for: package-lock.json line 2 restored to paranext-core. It will still churn to the installing directory's basename until the root is nameable.

Happy to file that as a ticket if you want it tracked — I held off rather than open a fifth one unasked.
___
*[`.context/standards/Architecture-Decisions.md` line 1117 at r1](https://reviewable.io/reviews/paranext/paranext-core/2714#-P-ql8aW81Ns7cuONBQj:-P08y-rOBFRL-SStNDQ_:b-mvjxxz) ([raw file](https://github.com/paranext/paranext-core/blob/32eb603c3c61c6a562ae080cd69197d135e7e659/.context/standards/Architecture-Decisions.md#L1117)):*
<details><summary><i>Previously, irahopkinson+AI (Agent AI for Ira Hopkinson) wrote…</i></summary><blockquote>

This paragraph documents a write-path data-integrity risk that nothing tracks.

> *recoverable* defects are silently repaired rather than rejected — on the extension host that repaired form is what gets written back to disk.

The ticket's "Correction" note retracts only the **fatal** case (0.9.12 throws on tag mismatch), then concludes *"No silent-corruption follow-up is needed"* — but it never addresses recoverable defects, which is what this sentence is about. So the ADR and the ticket are consistent on the facts and inconsistent on the conclusion; the ADR is the one that's right.

Net effect as written: the renderer **rejects** malformed USX, the extension host **repairs it and writes the repaired form to disk**. That asymmetry is worth more than a paragraph in an ADR that closes with this PR:

- a filed follow-up ticket, and
- a `TODO(PT-XXXX)` at the write site (`platform-scripture-extender-pdpe.model.ts:139,154`), where someone touching the save path would actually see it.

I searched Jira and no such ticket exists yet. Happy for the fix to be "file the ticket and reference it" rather than anything in code here.
</blockquote></details>

Agreed the ADR was the one that's right — the ticket's "Correction" note retracted only the fatal case, not the recoverable one. I re-ran it against 0.9.12 to pin down exactly which half is which.

Repaired silently, parsererror count 0 in all four cases: a bare & in <para style="p">Tom & Jerry</para> becomes &amp;; an unquoted attribute <usx version=3.0> becomes version="3.0" with a warning logged; an unknown entity a&foo;b becomes a&amp;foo;b with an error logged; a stray a < b becomes a &lt; b with an error logged.

Only a duplicate attribute (<para style="p" style="q">) actually throws, with Attribute style redefined.

A browser DOMParser emits a parsererror for all four of the repaired cases. So the renderer rejects precisely the inputs the extension host quietly rewrites — and on a read-modify-write the extension host persists the rewritten form.

Filed as PT-4444. It's named in the ADR at line 1853, and there's a TODO(PT-4444) at the save site in platform-scripture-extender-pdpe.model.ts lines 130 to 135, sitting directly above setBookUSJ/setChapterUSJ where the round-trip closes. I also corrected the "no silent-corruption follow-up is needed" note on PT-4412.
___
*[`extensions/src/platform-enhanced-resources/src/lib/marble-converter.ts` line 12 at r1](https://reviewable.io/reviews/paranext/paranext-core/2714#-P-qlCezB4cijj46SuDC:-P08xxzx9B2yH7eZZZJO:b-orxf7a) ([raw file](https://github.com/paranext/paranext-core/blob/32eb603c3c61c6a562ae080cd69197d135e7e659/extensions/src/platform-enhanced-resources/src/lib/marble-converter.ts#L12)):*
<details><summary><i>Previously, irahopkinson+AI (Agent AI for Ira Hopkinson) wrote…</i></summary><blockquote>

This TODO points at a ticket that closes when this PR merges, so it points at nothing.

```ts
// TODO: migrate to 0.9's `onError` and drop the pin (follow-up to PT-4412).
```

`.claude/rules/code-quality/forward-facing-comments.md` wants a TODO to name an **open** ticket (`TODO(PT-1234): …`) and explicitly lists *"ticket/PR references for work done in this PR"* as the backward-facing form to cut.

I searched Jira for the follow-ups this PR names and **none of the three exist**:

1. migrate `marble-converter` to `onError` and drop the 0.8 pin (this comment + ADR-0024)
2. drop `@xmldom/xmldom` from `platform-enhanced-resources` entirely — ADR-0024 notes `convertMarbleChapterXml` runs only in a web view, which already has a native `DOMParser`, so this would remove the pin, the version split, and that whole ADR paragraph at once
3. the `IgnorePlugin` `checkResource` substring bug you hit and worked around during verification — the ticket comment says it's *"worth its own ticket"*, and it's a nasty one: any checkout whose path contains `main`, `node`, `renderer`, `client`, or `extension-host` as a substring silently produces a broken renderer with no build error

Could you file them and swap this comment to `TODO(PT-XXXX)`? The three lines above it are genuinely forward-facing (a real constraint and gotcha) — those should stay exactly as they are.
</blockquote></details>

All three filed.

PT-4445 migrates marble-converter to 0.9's onError and drops the 0.8 pin. PT-4446 supersedes it by dropping @xmldom/xmldom from platform-enhanced-resources entirely. PT-4447 is the IgnorePlugin checkResource substring bug, filed as a Bug with a repro.

The comment is now TODO(PT-4445), naming PT-4446 as the larger option, and it explains the pin forward-facing: 0.8 is required because the errorHandler object below is 0.8-only, and the root now declares 0.9.x for the extension host's globals, so without the pin this file would resolve 0.9 and throw on every conversion. The three pre-existing lines above it are untouched, as you asked. adr-node-dom-globals-polyfill names PT-4445 and PT-4446 as well.

On PT-4446 specifically, your read is right: convertMarbleChapterXml's only production call site is enhanced-resource.web-view.tsx:1281 — the other two are a story and its own test — so the web view already has the native DOMParser it needs. Noted in the ticket that marble-converter.test.ts runs in a node environment and would need // @vitest-environment jsdom or the polyfill's globals.
___
*[`src/node/polyfills/dom-globals.polyfill.test.ts` line 102 at r1](https://reviewable.io/reviews/paranext/paranext-core/2714#-P-ql6rr0pOs9uEUKAu5:-P08xpOO2Wntgw4TyyLY:b-c1fxbd) ([raw file](https://github.com/paranext/paranext-core/blob/32eb603c3c61c6a562ae080cd69197d135e7e659/src/node/polyfills/dom-globals.polyfill.test.ts#L102)):*
<details><summary><i>Previously, irahopkinson+AI (Agent AI for Ira Hopkinson) wrote…</i></summary><blockquote>

This test can't fail — it passes with the polyfill deleted.

The comment above it says it gates on the package's own check:

> Asserting the globals exist is not the same as asserting the package accepts them: it gates on its own `assertDomEnvironment`

At the declared `@eten-tech-foundation/scripture-utilities@~0.1.6` it doesn't. I checked the installed package:

- `0.1.6` declares `@xmldom/xmldom: ^0.9.10` as its **own** dependency
- `dist/index.cjs` imports xmldom directly
- the string `requires a DOM environment` is **absent** from `dist`

Running both converters in bare Node with no globals installed:

```
globalThis.DOMParser     = undefined
globalThis.XMLSerializer = undefined
ROUND-TRIP OK with NO DOM globals -> true
```

So the one test that exists to prove the polyfill works cannot detect it being removed, reordered, or regressing. Combined with the fact that the DoD item *"Scripture read/write round-trips with no 'requires a DOM environment' error"* rests entirely on the manual yalc session in the ticket, there is currently nothing in CI that would catch this breaking before the bump lands.

Two options that would both work for me:

1. Add a test that deletes the globals and asserts the converters throw — it will be inert now but becomes a real guard the moment the version bump lands.
2. Or, minimally, reword the comment to say the assertion is inert until the bump, so nobody reads green CI as coverage.

Non-blocking on the approach; blocking on it not being silently misleading.
</blockquote></details>

You're right, and thanks for actually running it. Fixed, but not the obvious way — an unconditional expect(...).toThrow() would fail today, not sit inert, because at the currently installed ~0.1.6 the converters still bundle their own @xmldom/xmldom and succeed with no globals at all.

So the test probes which regime it's in and gates on the result. convertersNeedDomGlobals() at lines 63 to 87 deletes the globals, attempts one conversion, and restores them; the falsifying assertion at lines 150 to 161 is a test.skipIf on that probe.

It reports as skipped, not passed, for as long as the installed package carries its own XML implementation — so nobody reads green CI as proof the polyfill is load-bearing — and it becomes a real guard the moment the bump lands. Probing rather than reading the version range, because the behavior is what the polyfill exists for, not the range.

The round-trip test's comment at lines 125 to 130 now says out loud that it passes with or without the polyfill, instead of implying coverage it doesn't have.

Local run: 5 passed, 1 skipped, the skip being exactly this test.

irahopkinson (commented):
> @irahopkinson reviewed 6 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2714)* status: all files reviewed, 4 unresolved discussions (waiting on irahopkinson+AI).

irahopkinson (commented):
> @irahopkinson+AI made 6 comments and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2714)* status: all files reviewed, 3 unresolved discussions (waiting on irahopkinson and katherinejensen00).

___
*[`package-lock.json` line 2 at r1](https://reviewable.io/reviews/paranext/paranext-core/2714#-P-qaAkd9a5YMl0wWv6n:-P0Kxr_Y8R3wscx6p946:bxd2ylt) ([raw file](https://github.com/paranext/paranext-core/blob/32eb603c3c61c6a562ae080cd69197d135e7e659/package-lock.json#L2)):*

Took the root-cause fix, it broke CI on all three platforms, and I've reverted it. Detail, because it's a trap worth documenting:

Adding "name": "paranext-core" to the root package.json does stop npm deriving the lockfile name from the checkout directory basename. But it also makes eslint-plugin-import treat the root as a package, and import/no-relative-packages then rejects every pre-existing relative import that crosses out of the root into release/app, lib/platform-bible-react, or extensions. That's 12 errors in files this PR never touches: webpack.config.base.ts, bump-versions.ts, electron-rebuild.js, both .storybook/ files, app.service-host.ts, about-dialog.component.tsx, ten-layout-shared.tsx, tailwind.config.ts. While the root was anonymous the rule had nothing to anchor on. Confirmed both directions locally.

The rule's own suggested replacements don't resolve either. release/app is not a workspace, so there is no node_modules/platform-bible and import platform-bible/package.json would fail at runtime.

So naming the root needs either a workspace-layout change plus 12 import rewrites, or an import/no-relative-packages config change — its own piece of work, and not a DOM-globals PR. package.json is now identical to main apart from the one @xmldom/xmldom dependency, and this PR keeps only what you originally asked for: package-lock.json line 2 restored to paranext-core. It will still churn to the installing directory's basename until the root is nameable.

Happy to file that as a ticket if you want it tracked — I held off rather than open a fifth one unasked.
</blockquote></details>

Good — and thanks for trying it rather than taking my suggestion on trust. I checked the mechanism: `import/no-relative-packages` is `'error'` via `airbnb` → `eslint-config-airbnb-base/rules/imports.js:271`, and nothing in `.eslintrc.js` overrides it, so an anonymous root genuinely is what's keeping those 12 imports quiet. Reverting was right; suppressing the rule to land a cosmetic lockfile field would have been the worse trade.

Don't file a ticket for it — let's leave it. It's a field nothing reads, and the churn is visible in review when it happens. Not worth a fifth ticket off the back of this PR.

Line 2 is back to `paranext-core` and `package.json` is identical to main apart from the one dependency. Satisfied.
___
*[`vitest.setup.node.ts` line 2 at r7](https://reviewable.io/reviews/paranext/paranext-core/2714#-P0Kxv5v9CB1aYqX92H-:-P0Kxv5v9CB1aYqX92H0:b5ad0r7) ([raw file](https://github.com/paranext/paranext-core/blob/8cf4288cd4fa7e6196772b784e511493a19f347d/vitest.setup.node.ts#L2)):*
Holdover from my first pass that I didn't raise at the time — this enumeration isn't right, and it's stated twice.

> Additional setup for the `node`-environment vitest projects (`extensions`, `lib/platform-bible-utils`)

There are four node-environment vitest projects, not two:

- `extensions/vitest.config.ts:9` — `environment: 'node'` ✅ wired
- `lib/platform-bible-utils/vite.config.ts` — no `environment` key, so vitest defaults to `node` ✅ wired
- `lib/platform-bible-react/vitest.config.ts:34` — `name: 'scripts'`, `environment: 'node'` ❌ not wired
- `lib/eslint-plugin-paranext/vitest.config.ts` — no `environment` key ❌ not wired

Nothing is broken: neither unwired project touches the converters, so they don't need the polyfill. The problem is purely that the sentence claims to enumerate a category and doesn't, and the same claim appears in `adr-node-dom-globals-polyfill`. Whoever adds a third converter-touching node project will read this, believe the list is complete, and not think to check.

Suggest rewording both sites to something like "the node-environment projects that run the converters" — which is the actual selection criterion, is true, and stays true.

Non-blocking; docs accuracy rather than behaviour.
___
*[`.context/standards/Architecture-Decisions.md` line 1117 at r1](https://reviewable.io/reviews/paranext/paranext-core/2714#-P-ql8aW81Ns7cuONBQj:-P0KxoXLBLZCRIJP1hzZ:bk3o6d5) ([raw file](https://github.com/paranext/paranext-core/blob/32eb603c3c61c6a562ae080cd69197d135e7e659/.context/standards/Architecture-Decisions.md#L1117)):*

Agreed the ADR was the one that's right — the ticket's "Correction" note retracted only the fatal case, not the recoverable one. I re-ran it against 0.9.12 to pin down exactly which half is which.

Repaired silently, parsererror count 0 in all four cases: a bare & in <para style="p">Tom & Jerry</para> becomes &amp;; an unquoted attribute <usx version=3.0> becomes version="3.0" with a warning logged; an unknown entity a&foo;b becomes a&amp;foo;b with an error logged; a stray a < b becomes a &lt; b with an error logged.

Only a duplicate attribute (<para style="p" style="q">) actually throws, with Attribute style redefined.

A browser DOMParser emits a parsererror for all four of the repaired cases. So the renderer rejects precisely the inputs the extension host quietly rewrites — and on a read-modify-write the extension host persists the rewritten form.

Filed as PT-4444. It's named in the ADR at line 1853, and there's a TODO(PT-4444) at the save site in platform-scripture-extender-pdpe.model.ts lines 130 to 135, sitting directly above setBookUSJ/setChapterUSJ where the round-trip closes. I also corrected the "no silent-corruption follow-up is needed" note on PT-4412.
</blockquote></details>

The four-repaired-cases-versus-one-throwing breakdown is a sharper characterisation than either the ADR or my comment had — thanks for going and measuring it rather than reasoning about it. "The renderer rejects precisely the inputs the extension host quietly rewrites" is the sentence that makes PT-4444 legible to whoever picks it up.

Verified the `TODO(PT-4444)` lands directly above `setBookUSJ`, so it's in the path of anyone touching the save round-trip rather than stranded at the top of the file. ADR names PT-4444 at the write-path paragraph.

Satisfied.
___
*[`extensions/src/platform-enhanced-resources/src/lib/marble-converter.ts` line 12 at r1](https://reviewable.io/reviews/paranext/paranext-core/2714#-P-qlCezB4cijj46SuDC:-P0KxqItAdUG0ixugNMh:bryd3q4) ([raw file](https://github.com/paranext/paranext-core/blob/32eb603c3c61c6a562ae080cd69197d135e7e659/extensions/src/platform-enhanced-resources/src/lib/marble-converter.ts#L12)):*

All three filed.

PT-4445 migrates marble-converter to 0.9's onError and drops the 0.8 pin. PT-4446 supersedes it by dropping @xmldom/xmldom from platform-enhanced-resources entirely. PT-4447 is the IgnorePlugin checkResource substring bug, filed as a Bug with a repro.

The comment is now TODO(PT-4445), naming PT-4446 as the larger option, and it explains the pin forward-facing: 0.8 is required because the errorHandler object below is 0.8-only, and the root now declares 0.9.x for the extension host's globals, so without the pin this file would resolve 0.9 and throw on every conversion. The three pre-existing lines above it are untouched, as you asked. adr-node-dom-globals-polyfill names PT-4445 and PT-4446 as well.

On PT-4446 specifically, your read is right: convertMarbleChapterXml's only production call site is enhanced-resource.web-view.tsx:1281 — the other two are a story and its own test — so the web view already has the native DOMParser it needs. Noted in the ticket that marble-converter.test.ts runs in a node environment and would need // @vitest-environment jsdom or the polyfill's globals.
</blockquote></details>

All three confirmed filed with summaries that match what they're for. The TODO now names PT-4445 and points at PT-4446 as the larger option, which is the right shape — the near fix is actionable and the better fix is discoverable from it.

The note on PT-4446 about `marble-converter.test.ts` running in a `node` environment and needing `// @vitest-environment jsdom` is a useful catch to have recorded up front; that's the kind of thing that turns a "simple" dependency removal into a surprise.

Satisfied.
___
*[`src/node/polyfills/dom-globals.polyfill.test.ts` line 102 at r1](https://reviewable.io/reviews/paranext/paranext-core/2714#-P-ql6rr0pOs9uEUKAu5:-P0KxnJ58iuVABs1lnKY:b-vpfkmv) ([raw file](https://github.com/paranext/paranext-core/blob/32eb603c3c61c6a562ae080cd69197d135e7e659/src/node/polyfills/dom-globals.polyfill.test.ts#L102)):*

You're right, and thanks for actually running it. Fixed, but not the obvious way — an unconditional expect(...).toThrow() would fail today, not sit inert, because at the currently installed ~0.1.6 the converters still bundle their own @xmldom/xmldom and succeed with no globals at all.

So the test probes which regime it's in and gates on the result. convertersNeedDomGlobals() at lines 63 to 87 deletes the globals, attempts one conversion, and restores them; the falsifying assertion at lines 150 to 161 is a test.skipIf on that probe.

It reports as skipped, not passed, for as long as the installed package carries its own XML implementation — so nobody reads green CI as proof the polyfill is load-bearing — and it becomes a real guard the moment the bump lands. Probing rather than reading the version range, because the behavior is what the polyfill exists for, not the range.

The round-trip test's comment at lines 125 to 130 now says out loud that it passes with or without the polyfill, instead of implying coverage it doesn't have.

Local run: 5 passed, 1 skipped, the skip being exactly this test.
</blockquote></details>

Probing rather than reading the version range is the right call, and the skip-not-pass reporting is exactly the property I was after — the regime is visible in the run output instead of having to be inferred.

Checked the falsifying direction holds post-bump: with the polyfill gone, `importPolyfillFresh()` fails outright and the round-trip test goes red, so the suite doesn't depend on the `skipIf` test alone to notice. Good.

One non-blocking note for whoever hits it later: the probe's `catch` is bare, so any post-bump failure that isn't the DOM-environment error also flips it to `true`, and the `skipIf` test then fails on the `/DOM environment/` match rather than on the real cause. That fails loudly rather than silently, which is the right side to err on — just worth knowing the error message may point at the wrong thing.

Satisfied.
___
*[`src/node/polyfills/dom-globals.polyfill.test.ts` line 50 at r7](https://reviewable.io/reviews/paranext/paranext-core/2714#-P0KxyPP3wXh3Jt30KU0:-P0KxyPP3wXh3Jt30KU1:b-kabdzb) ([raw file](https://github.com/paranext/paranext-core/blob/8cf4288cd4fa7e6196772b784e511493a19f347d/src/node/polyfills/dom-globals.polyfill.test.ts#L50)):*
Also a holdover from my first pass. These three regexes are copied from `extension-host-import-boundary.test.ts:57-62`, and two of them lost coverage in the copy — which punches holes in the ordering guard this test exists to be.

Source:

```ts
const FROM_IMPORT_REGEX = /(?:^|\n)[ \t]*(?:import|export)\b([^;'"]*?)from[ \t]*['"]([^'"]+)['"]/g;
const RUNTIME_IMPORT_REGEX = /\b(?:import|require)\([ \t]*['"]([^'"]+)['"]/g;
```

Here:

```ts
const fromImport = /(?:^|\n)[ \t]*import\b([^;'"]*?)from[ \t]*['"]([^'"]+)['"]/g;   // import only
const runtimeRequire = /\brequire\([ \t]*['"]([^'"]+)['"]/g;                        // require only
```

So `export … from '…'` and a dynamic `import('…')` are both invisible to `firstEvaluatedSpecifier`. Either one placed above the polyfill import in `extension-host.ts` would evaluate first at runtime while this test still reported the polyfill as first — precisely the silent reorder the test is guarding against.

Low likelihood (neither form is idiomatic at the top of an entry point), but the cost of closing it is copying two characters' worth of alternation. The TSDoc already explains why the helper can't be shared across the workspace boundary:

> the two cannot share a helper across the workspace boundary

Worth adding that the regexes are a deliberate copy and must be kept in sync, so the next person to fix a parsing bug in one knows to fix both.

Non-blocking.

irahopkinson (commented):
> @irahopkinson resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2714)* status: all files reviewed, 2 unresolved discussions (waiting on katherinejensen00).

katherinejensen00:
Both r7 holdovers addressed in `733f808`, and rebased onto `main` (`790d8d6`).

### `vitest.setup.node.ts` line 2 — the enumeration isn't right, and it's stated twice

Agreed, and your suggested wording is the fix: the selection criterion is *runs the converters*, not *is a `node` project*. Confirmed your four-way breakdown before rewording — `extensions/vitest.config.ts:9` and `lib/platform-bible-utils/vite.config.ts` (no `environment` key, so `node` by default) wire the setup file in; `lib/platform-bible-react/vitest.config.ts:34`'s `scripts` project and `lib/eslint-plugin-paranext/vitest.config.ts` do not, and neither touches the converters.

Reworded both sites to state the criterion instead of a list, and added the bit that closes the trap you named — that each project wires the file in itself, so a `node` project not being covered is the expected case rather than an oversight to go fix:

```
// Additional setup for the `node`-environment vitest projects that run the scripture converters,
// loaded alongside the shared `vitest.setup.ts`. Each such project wires this file in itself, so a
// `node`-environment project that never touches the converters does not load it — check that
// criterion rather than assuming every `node` project is covered.
```

`adr-node-dom-globals-polyfill` now says the same thing.

### `dom-globals.polyfill.test.ts` line 50 — two alternations lost in the copy

You're right, and it's the ordering guard's own blind spot, which makes it worth more than the two characters it costs. Restored both to match the source:

```ts
const fromImport = /(?:^|\n)[ \t]*(?:import|export)\b([^;'"]*?)from[ \t]*['"]([^'"]+)['"]/g;
const runtimeImport = /\b(?:import|require)\([ \t]*['"]([^'"]+)['"]/g;
```

Renamed `runtimeRequire` → `runtimeImport` since it is no longer require-only. Checked all four previously-invisible forms now match, and that the type-only path still routes correctly:

| Source | Before | After |
| --- | --- | --- |
| `export * from './early'` | not matched | `./early` |
| `export { a } from './early'` | not matched | `./early` |
| `await import('./early')` | not matched | `./early` |
| `import type { T } from './t'` | skipped | skipped (clause `" type { T } "` → `isTypeOnlyClause` true) |

Took the keep-in-sync note too, on `isTypeOnlyClause`'s TSDoc where the "cannot share a helper" sentence already was — so the next person to fix a parsing bug reads it in the same breath as the reason there are two copies:

> the two cannot share a helper across the workspace boundary, so this function and the specifier patterns below are a deliberate copy of that file's. Fix a parsing bug in one and fix it in the other.

Also widened `firstEvaluatedSpecifier`'s TSDoc, which had described the narrower forms.

### On the bare `catch` in the probe

Not changing it, but recording your read so it isn't re-derived: post-bump, a failure that isn't the DOM-environment error also flips the probe to `true`, and the `skipIf` test then fails on the `/DOM environment/` match rather than the real cause. Loud-but-misattributed is the right side to err on for a guard whose whole point is to stop being skipped, and narrowing the catch would have to guess at an error shape the package hasn't shipped yet.

### Rebase

Clean this time — no ADR collision, which is the slug convention doing its job. First rebase of the four where the entry didn't need renumbering.

### Verification

- `dom-globals.polyfill.test.ts`: `5 passed | 1 skipped`, the skip still being the `skipIf` guard.
- `lib/platform-bible-utils`: `30 files | 478 passed | 1 skipped`.
- Repo-wide `eslint --ext .cjs,.js,.jsx,.ts,.tsx --no-cache .`: only the 3 known `@eten-tech-foundation/*` yalc resolution errors, one of them in `ten-layout-shared.tsx`, which this branch doesn't touch.
- `prettier --check` clean on all three changed files.

Still the same local caveat as before: the `yalc` link for `@eten-tech-foundation/*` has no `dist/`, so `typecheck` and the `extensions` suite report failures that are all `Cannot find module '@eten-tech-foundation/…'` or `Failed to resolve import "shared" from .yalc/…/platform-editor/src/index.ts` — none in changed files, none reproducible on a clean install.

katherinejensen00 (commented):
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2714)* status: 9 of 12 files reviewed, 2 unresolved discussions (waiting on irahopkinson and irahopkinson+AI).

___
*[`vitest.setup.node.ts` line 2 at r7](https://reviewable.io/reviews/paranext/paranext-core/2714#-P0Kxv5v9CB1aYqX92H-:-P0NhHfS61AnSO0vuzoe:b2ic498) ([raw file](https://github.com/paranext/paranext-core/blob/8cf4288cd4fa7e6196772b784e511493a19f347d/vitest.setup.node.ts#L2)):*
<details><summary><i>Previously, irahopkinson+AI (Agent AI for Ira Hopkinson) wrote…</i></summary><blockquote>

Holdover from my first pass that I didn't raise at the time — this enumeration isn't right, and it's stated twice.

> Additional setup for the `node`-environment vitest projects (`extensions`, `lib/platform-bible-utils`)

There are four node-environment vitest projects, not two:

- `extensions/vitest.config.ts:9` — `environment: 'node'` ✅ wired
- `lib/platform-bible-utils/vite.config.ts` — no `environment` key, so vitest defaults to `node` ✅ wired
- `lib/platform-bible-react/vitest.config.ts:34` — `name: 'scripts'`, `environment: 'node'` ❌ not wired
- `lib/eslint-plugin-paranext/vitest.config.ts` — no `environment` key ❌ not wired

Nothing is broken: neither unwired project touches the converters, so they don't need the polyfill. The problem is purely that the sentence claims to enumerate a category and doesn't, and the same claim appears in `adr-node-dom-globals-polyfill`. Whoever adds a third converter-touching node project will read this, believe the list is complete, and not think to check.

Suggest rewording both sites to something like "the node-environment projects that run the converters" — which is the actual selection criterion, is true, and stays true.

Non-blocking; docs accuracy rather than behaviour.
</blockquote></details>

Agreed, and your suggested wording is the fix: the selection criterion is _runs the converters_, not _is a `node` project_. Confirmed your four-way breakdown before rewording — `extensions/vitest.config.ts:9` and `lib/platform-bible-utils/vite.config.ts` (no `environment` key, so `node` by default) wire the setup file in; `lib/platform-bible-react/vitest.config.ts:34`'s `scripts` project and `lib/eslint-plugin-paranext/vitest.config.ts` do not, and neither touches the converters.

Reworded both sites to state the criterion instead of a list, and added the bit that closes the trap you named — that each project wires the file in itself, so a `node` project not being covered is the expected case rather than an oversight to go fix:

```
// Additional setup for the `node`-environment vitest projects that run the scripture converters,
// loaded alongside the shared `vitest.setup.ts`. Each such project wires this file in itself, so a
// `node`-environment project that never touches the converters does not load it — check that
// criterion rather than assuming every `node` project is covered.
```

`adr-node-dom-globals-polyfill` now says the same thing.
___
*[`src/node/polyfills/dom-globals.polyfill.test.ts` line 102 at r1](https://reviewable.io/reviews/paranext/paranext-core/2714#-P-ql6rr0pOs9uEUKAu5:-P0NipbxCUt-onwShAkJ:b87qcy9) ([raw file](https://github.com/paranext/paranext-core/blob/32eb603c3c61c6a562ae080cd69197d135e7e659/src/node/polyfills/dom-globals.polyfill.test.ts#L102)):*
<details><summary><i>Previously, irahopkinson+AI (Agent AI for Ira Hopkinson) wrote…</i></summary><blockquote>

Probing rather than reading the version range is the right call, and the skip-not-pass reporting is exactly the property I was after — the regime is visible in the run output instead of having to be inferred.

Checked the falsifying direction holds post-bump: with the polyfill gone, `importPolyfillFresh()` fails outright and the round-trip test goes red, so the suite doesn't depend on the `skipIf` test alone to notice. Good.

One non-blocking note for whoever hits it later: the probe's `catch` is bare, so any post-bump failure that isn't the DOM-environment error also flips it to `true`, and the `skipIf` test then fails on the `/DOM environment/` match rather than on the real cause. That fails loudly rather than silently, which is the right side to err on — just worth knowing the error message may point at the wrong thing.

Satisfied.
</blockquote></details>

Right on the substance, and worth having in the thread — but I checked what the failure actually prints, and the real cause isn't lost, only demoted. Vitest reports both the pattern and the received message:

AssertionError: expected [Function] to throw error matching /DOM environment/
  but got 'usxStringToUsj: unexpected token at l…'

- Expected:
+ Received:
"usxStringToUsj: unexpected token at line 3"

So the headline and the test name assert "DOM environment" while the body shows what actually happened. The failure mode is a reader trusting the name over the Received: line, not information that never made it out of the catch. Narrower than losing the cause outright, and it argues for leaving the bare catch rather than tightening it: the probe's job is "did this throw at all", and any narrowing has to guess at an error shape the post-bump package hasn't shipped yet — a guess that's wrong is worse than a name that's imprecise, because it would send the probe back to false and silently re-skip the test.

One more thing pushing the real cause into view, for whoever hits this: a failure that isn't DOM-related almost certainly reddens the round-trip test too, since both call usxStringToUsj and that one runs with the globals installed and asserts nothing about the DOM. So the honest error usually arrives alongside, in a test whose name doesn't editorialize. The case where the skipIf test fails alone is specifically the case where the failure only reproduces without globals — which is the thing it's named for.

Leaving it as-is. If it does bite someone later, the cheap fix isn't narrowing the catch — it's having the probe keep what it caught and surface it in the assertion message, so the name and the evidence agree. Not worth doing speculatively against an error shape nobody has seen.
___
*[`src/node/polyfills/dom-globals.polyfill.test.ts` line 50 at r7](https://reviewable.io/reviews/paranext/paranext-core/2714#-P0KxyPP3wXh3Jt30KU0:-P0NhVXk1D6wvsOAWXdD:bqlm4ir) ([raw file](https://github.com/paranext/paranext-core/blob/8cf4288cd4fa7e6196772b784e511493a19f347d/src/node/polyfills/dom-globals.polyfill.test.ts#L50)):*
You're right, and it's the ordering guard's own blind spot, which makes it worth more than the two characters it costs. Restored both to match the source:

```ts
const fromImport = /(?:^|\n)[ \t]*(?:import|export)\b([^;'"]*?)from[ \t]*['"]([^'"]+)['"]/g;
const runtimeImport = /\b(?:import|require)\([ \t]*['"]([^'"]+)['"]/g;
```

Renamed `runtimeRequire` → `runtimeImport` since it is no longer require-only. Checked all four previously-invisible forms now match, and that the type-only path still routes correctly:

| Source | Before | After |
| --- | --- | --- |
| `export * from './early'` | not matched | `./early` |
| `export { a } from './early'` | not matched | `./early` |
| `await import('./early')` | not matched | `./early` |
| `import type { T } from './t'` | skipped | skipped (clause `" type { T } "` → `isTypeOnlyClause` true) |

Took the keep-in-sync note too, on `isTypeOnlyClause`'s TSDoc where the "cannot share a helper" sentence already was — so the next person to fix a parsing bug reads it in the same breath as the reason there are two copies:

> the two cannot share a helper across the workspace boundary, so this function and the specifier patterns below are a deliberate copy of that file's. Fix a parsing bug in one and fix it in the other.

Also widened `firstEvaluatedSpecifier`'s TSDoc, which had described the narrower forms.

irahopkinson (commented):
> @irahopkinson+AI made 3 comments and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2714)* status: 9 of 12 files reviewed, all discussions resolved (waiting on irahopkinson).

___
*[`vitest.setup.node.ts` line 2 at r7](https://reviewable.io/reviews/paranext/paranext-core/2714#-P0Kxv5v9CB1aYqX92H-:-P0OS2Bk758ZIfFIkNwR:b-i4di0p) ([raw file](https://github.com/paranext/paranext-core/blob/8cf4288cd4fa7e6196772b784e511493a19f347d/vitest.setup.node.ts#L2)):*

Agreed, and your suggested wording is the fix: the selection criterion is _runs the converters_, not _is a `node` project_. Confirmed your four-way breakdown before rewording — `extensions/vitest.config.ts:9` and `lib/platform-bible-utils/vite.config.ts` (no `environment` key, so `node` by default) wire the setup file in; `lib/platform-bible-react/vitest.config.ts:34`'s `scripts` project and `lib/eslint-plugin-paranext/vitest.config.ts` do not, and neither touches the converters.

Reworded both sites to state the criterion instead of a list, and added the bit that closes the trap you named — that each project wires the file in itself, so a `node` project not being covered is the expected case rather than an oversight to go fix:

```
// Additional setup for the `node`-environment vitest projects that run the scripture converters,
// loaded alongside the shared `vitest.setup.ts`. Each such project wires this file in itself, so a
// `node`-environment project that never touches the converters does not load it — check that
// criterion rather than assuming every `node` project is covered.
```

`adr-node-dom-globals-polyfill` now says the same thing.
</blockquote></details>

Verified at `733f808` — both sites now state the criterion rather than a list (`vitest.setup.node.ts:1-4`, `adr-node-dom-globals-polyfill` at `Architecture-Decisions.md:2236-2238`).

The sentence you added is better than what I asked for. I'd only pointed out that the list was wrong; "each such project wires this file in itself, so a `node`-environment project that never touches the converters does not load it — check that criterion rather than assuming every `node` project is covered" also tells the next reader what *not* to do about it. Without that, someone finding the two unwired projects would plausibly go wire them up and pull xmldom into workers that don't need it.

Satisfied.
___
*[`src/node/polyfills/dom-globals.polyfill.test.ts` line 102 at r1](https://reviewable.io/reviews/paranext/paranext-core/2714#-P-ql6rr0pOs9uEUKAu5:-P0OS1Eb3kxVAaDEiE-F:bs1jn9n) ([raw file](https://github.com/paranext/paranext-core/blob/32eb603c3c61c6a562ae080cd69197d135e7e659/src/node/polyfills/dom-globals.polyfill.test.ts#L102)):*

Right on the substance, and worth having in the thread — but I checked what the failure actually prints, and the real cause isn't lost, only demoted. Vitest reports both the pattern and the received message:

AssertionError: expected [Function] to throw error matching /DOM environment/
  but got 'usxStringToUsj: unexpected token at l…'

- Expected:
+ Received:
"usxStringToUsj: unexpected token at line 3"

So the headline and the test name assert "DOM environment" while the body shows what actually happened. The failure mode is a reader trusting the name over the Received: line, not information that never made it out of the catch. Narrower than losing the cause outright, and it argues for leaving the bare catch rather than tightening it: the probe's job is "did this throw at all", and any narrowing has to guess at an error shape the post-bump package hasn't shipped yet — a guess that's wrong is worse than a name that's imprecise, because it would send the probe back to false and silently re-skip the test.

One more thing pushing the real cause into view, for whoever hits this: a failure that isn't DOM-related almost certainly reddens the round-trip test too, since both call usxStringToUsj and that one runs with the globals installed and asserts nothing about the DOM. So the honest error usually arrives alongside, in a test whose name doesn't editorialize. The case where the skipIf test fails alone is specifically the case where the failure only reproduces without globals — which is the thing it's named for.

Leaving it as-is. If it does bite someone later, the cheap fix isn't narrowing the catch — it's having the probe keep what it caught and surface it in the assertion message, so the name and the evidence agree. Not worth doing speculatively against an error shape nobody has seen.
</blockquote></details>

You're right and my note was wrong on the point that matters. Narrowing the catch would have to guess an error shape the post-bump package hasn't shipped, and a guess that's too narrow returns `false` — which silently re-skips the falsifying test and reopens exactly the hole this thread was about. A test name that editorializes is a far cheaper failure than a silent skip. Leave the bare catch.

The observation that a non-DOM failure also reddens the round-trip test is the part I'd missed: that test runs *with* the globals installed and asserts nothing about the DOM, so the honest error normally arrives alongside, in a test whose name isn't making a claim. That narrows the misattribution window to precisely the case the `skipIf` test is named for.

And your deferred fix is the right one if it ever bites — have the probe keep what it caught and surface it in the assertion message, so the name and the evidence agree. Better than narrowing, and correctly not done speculatively.

Thanks for recording the read in the top-level summary rather than just dropping it; that's the bit that stops it being re-litigated.

Satisfied.
___
*[`src/node/polyfills/dom-globals.polyfill.test.ts` line 50 at r7](https://reviewable.io/reviews/paranext/paranext-core/2714#-P0KxyPP3wXh3Jt30KU0:-P0OS4Ww-PBOJCyWMFvi:b3vaslk) ([raw file](https://github.com/paranext/paranext-core/blob/8cf4288cd4fa7e6196772b784e511493a19f347d/src/node/polyfills/dom-globals.polyfill.test.ts#L50)):*

You're right, and it's the ordering guard's own blind spot, which makes it worth more than the two characters it costs. Restored both to match the source:

```ts
const fromImport = /(?:^|\n)[ \t]*(?:import|export)\b([^;'"]*?)from[ \t]*['"]([^'"]+)['"]/g;
const runtimeImport = /\b(?:import|require)\([ \t]*['"]([^'"]+)['"]/g;
```

Renamed `runtimeRequire` → `runtimeImport` since it is no longer require-only. Checked all four previously-invisible forms now match, and that the type-only path still routes correctly:

| Source | Before | After |
| --- | --- | --- |
| `export * from './early'` | not matched | `./early` |
| `export { a } from './early'` | not matched | `./early` |
| `await import('./early')` | not matched | `./early` |
| `import type { T } from './t'` | skipped | skipped (clause `" type { T } "` → `isTypeOnlyClause` true) |

Took the keep-in-sync note too, on `isTypeOnlyClause`'s TSDoc where the "cannot share a helper" sentence already was — so the next person to fix a parsing bug reads it in the same breath as the reason there are two copies:

> the two cannot share a helper across the workspace boundary, so this function and the specifier patterns below are a deliberate copy of that file's. Fix a parsing bug in one and fix it in the other.

Also widened `firstEvaluatedSpecifier`'s TSDoc, which had described the narrower forms.
</blockquote></details>

Verified at `733f808`. Both patterns are now byte-identical to `extension-host-import-boundary.test.ts`, and `runtimeRequire` → `runtimeImport` is the right rename.

I re-ran your table independently rather than reading it, and added the case that actually matters — whether the guard now *catches* the reorder, not just whether the regex matches:

```
PASS  "export * from './early'"               -> ./early
PASS  "export { a } from './early'"           -> ./early
PASS  "await import('./early')"               -> ./early
PASS  "import type { T } from './t'"          -> undefined
PASS  "import { type A, type B } from './t'"  -> undefined

--- sneaky export-from above the polyfill import ---
./sneaky
```

That last one is the point: `export * from './sneaky'` placed above the polyfill import now resolves to `./sneaky`, so `firstEvaluatedSpecifier` no longer reports the polyfill as first and the ordering test goes red. Previously it stayed green. Blind spot closed, not just papered over.

The keep-in-sync note landing on `isTypeOnlyClause`'s TSDoc — beside the sentence that already explains why there are two copies — is a better home than where I'd have put it.

Satisfied.

irahopkinson (approved):
:lgtm: thanks for this

> @irahopkinson reviewed 3 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2714#-:-P0OdhXM0VDvDjghlUpc:b-r4tcg3)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2715, opened 2026-08-25
PT-3408: Make whitespace and diacritic tolerance togglable Find options

jolierabideau:
Searching for a term typed with extra spaces returned no results ([PT-3408]). This PR makes whitespace tolerance — and its sibling, diacritic tolerance — user-togglable Find options, rather than something the search silently does to the query.

## The approach changed during review

The first version of this PR fixed the bug by collapsing runs of spaces in the query, on the premise that saved Scripture can never contain consecutive spaces because ParatextData's `UsfmToken.NormalizeUsfm` regularizes whitespace on every write.

**That premise is false, and @katherinejensen00's review caught it.** Find searches USJ, not USFM. `UsjReaderWriter` concatenates adjacent text nodes with `textChunks.join('')` and no separator — its own comment says so — meaning a trailing space in one node followed by a leading space in the next is a genuine double space in the searched text. Running the real `ScriptureFinderProjectDataProviderEngine` against a fixture of that shape (`<char style="it">Salvador. </char><unmatched/> porque`) confirms it:

| query | result |
| --- | --- |
| `"Salvador.  porque"` raw | 1 correct match |
| the same query through the collapse | **0 matches** |

Collapsing didn't just add noise — it destroyed a match that worked before. Two further consequences: a padded query matched a span *including* the padding, so Replace spliced those spaces away and fused the neighbouring words (`In the beginning was` → `In thestartwas`); and the fix did nothing at all under `wholeWord`, `startOfWord`, or `endOfWord`.

## What this PR does instead

The search term reaches the engine **byte-identical**. Whitespace and diacritic tolerance are expressed only as `FindOptions` the engine applies when *matching* — `ignoreWhitespaceDifferences` and `ignoreDiacritics`, both already implemented in `buildSearchRegex` — and both are surfaced as Find filters, mirroring PT9's user-togglable `FindReplaceOptions`.

Both default to **off**, so an exact search is what the user gets unless they opt out of exactness. That preserves searching for a specific invisible character (NBSP, ZWSP, a bidi mark), which the Find UI deliberately renders in results via `INVISIBLE_CHAR_SYMBOLS` — the capability that ruled out turning `ignoreWhitespaceDifferences` on unconditionally. `ignoreDiacritics` was already implemented and tested at the engine level but unreachable from the UI; it gains its control here as the sibling flag on the same PT9 dialog.

`buildSearchRegex` already ignores both flags in regex mode, where the pattern must mean exactly what the user wrote, so the adapter needs no carve-out.

The UI adapter moved to `find/find-options.utils.ts`, so the backend PDP engine — which imports `find.utils.ts` for `buildSearchRegex` — no longer pulls in web-view code, and the engine tests exercise the engine contract directly.

Recorded as **ADR-0029**.

## Test-suite repairs

Findings in the review that were about the tests rather than the fix:

- `pollFindJob` now **accumulates every batch** (`retrieveFindJobUpdate` drains destructively, so reading only the final report discarded earlier results), is **bounded** at 200 polls, and **throws unless the job completes** — previously `errored`, `stopped`, and `exceeded` all returned `[]`, which silently turned every "finds nothing" assertion into a no-op that passed for the wrong reason.
- The four remaining polling copies are consolidated onto it.
- An inverted comment about `ignoreDiacritics` corrected against observed output (`['e','é','e','é','é']`).
- `.includes()` assertions replaced with `toEqual` — that looseness is what hid the match-span defect.
- The pre-existing `TS2345` errors in `find.utils.test.ts` cleared. `tsconfig.json` excludes `**/*.test.ts`, so these never reached CI.

## Known limits, called out deliberately

- **A padded query with the toggle on still matches a span including the padding**, so a Replace still fuses the neighbouring words. Not "fixed": the user typed those spaces, nothing rewrote them silently, and PT9 behaves the same way. Pinned by a test asserting `[' beginning ']`. Trimming the query would be a product decision, not a bug fix.
- **Test files remain excluded from `npm run typecheck`.** The exclusion has a documented reason (importing `vitest` pulls Vite's type entry point, unresolvable under `moduleResolution: "Node"`), so lifting it repo-wide is its own piece of work.
- **The Storybook harness approximates `ignoreWhitespaceDifferences` but not `ignoreDiacritics`** — emulating NFD normalization over fixture text is out of proportion to a story. Its TSDoc records this.
- **Upstream ParatextData is separately moving to preserve consecutive spaces** (`PreserveConsecutiveSpacesInTextTokens`). This approach is unaffected, since it never relied on the normalization holding.

## Still open for the ticket, not the code

PT-3408's Definition of Done still reads *"Spaces should be forgiven? Need feedback UX"*. A toggle answers that in the user's hands rather than guessing at a global default, but the DoD is worth closing out with UX either way.

## Verification

`npm run typecheck` clean · `npm run lint` **0 errors** · **501 tests pass** in `extensions/src/platform-scripture` (up from 490) · prettier clean · branch up to date with `main`.

Falsifiability was checked rather than assumed — breaking each layer fails the right tests: dropping the flag in `buildFindOptions` fails 1 adapter test; ignoring the option in `buildSearchRegex` fails 7 engine tests; unwiring the UI setter fails 2 component tests.

---

AI-assisted.

[PT-3408]: https://paratextstudio.atlassian.net/browse/PT-3408

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2715)


katherinejensen00 (changes requested):
Summary: The change is well-tested and cleanly extracted, but the invariant it rests on — that saved Scripture can never contain consecutive spaces — does not hold, so collapsing the query trades one wrong answer for another. I confirmed by running the real engine against a fixture already in this repo that a two-space query returns 2 correct matches on main and 465 junk matches on this branch. I'd suggest pausing on the current approach before addressing the smaller items, since most of them dissolve if the query stops being rewritten.

The core issue

1. The premise is false — the searched text really can contain consecutive spaces. find/find.utils.ts:765 · Critical

ParatextData regularizes whitespace within a USFM token, but Find searches USJ, where adjacent text nodes are concatenated with no separator. usj-reader-writer.ts:1738 does textChunks.join('') and says so in its own comment: "Note that there are no spaces between the text chunks." So a trailing space in one node plus a leading space in the next produces a real double space.

footnote-util-test.usj.data.ts contains <char style="it">Salvador. </char><unmatched/> porque. Running the real ScriptureFinderProjectDataProviderEngine against it:

RAW two-space query      -> completed, total=2      (origin/main)
VIA buildFindOptions     -> completed, total=465    (this branch)
RAW "Salvador.  porque"  -> total=1, texts=["Salvador.  porque"]

Two precise matches become 465 useless ones. That is arguably a worse outcome than the zero results PT-3408 reports.

2. Replace fuses the surrounding words when the query carries padding. find.web-view.tsx:1184 (real sites :1833, :2034) · Critical

'   beginning    ' collapses to ' beginning ', which matches an 11-character span including both spaces. Replace splices the whole span. I ran it: In the beginning was the Word. replaced with start gives In thestartwas the Word. — identical with preserveCase on or off. isFindQueryValid only rejects all-whitespace queries and usfmChangesStructure compares block markers only, so nothing catches it. Power-mode-only limits the blast radius but doesn't remove it.

30. Upstream is actively removing the invariant. find/find.utils.ts:742 · High

ParatextData has an in-flight branch origin/preserve-consecutive-spaces-9.5 (3 commits ahead of origin/9.5, base 00ad807c3c "Preserve consecutive spaces") adding UsfmToken.WhitespacePreservationType.PreserveConsecutiveSpacesInTextTokens — "In text tokens, preserve consecutive spaces of any kind" — threaded through Tokenize into RegularizeSpaces. Once ParatextPackageVersion (csproj:43, currently 9.5.0.24) bumps past that release, single-node double spaces become ordinary and no test will signal it.

Suggested direction

Findings 1, 2, 3, 6, 15, 16, 17, 18 and 23 all dissolve if the query reaches the engine byte-identical and the whitespace tolerance moves to the text side. FindOptions.ignoreWhitespaceDifferences already does exactly this, is already wired through buildSearchRegex, and is already tested for this precise case at find.utils.test.ts:190 — it('matches when the search has more spaces than the text'), 'hello  world' matching 'hello world'. It also would not have caused finding 1, because it relaxes matching rather than destroying information in the query. Whichever way you go, finding 19 asks for it in the ADR log.

Remaining findings

High:

3. Any word restriction defeats the fix entirely — padded query yields 0 matches under wholeWord, startOfWord and endOfWord, vs matches for all four when trimmed. Untested, and findWordRestriction persists across sessions. find/find.utils.ts:769
4. The declined engine option is already built, wired and tested for this bug; the collapse shipped instead removes exact multi-space search unconditionally, with no toggle. find/find.utils.ts:754
5. Re-implements and narrows normalizeScriptureSpaces (scripture-util.ts:721, "a direct translation of UsfmToken.RegularizeSpaces"), which this file's package is already imported from. / {2,}/gu misses TAB, newline, U+3000, EN/EM space and NBSP-split runs. find/find.utils.ts:718
6. Collapsing is not trimming, so '  Word.  ' still misses at verse end; isFindQueryValid:194 judges validity on .trim(), so the two disagree. find/find.utils.ts:763
7. No test covers the behavior change — reverting find.web-view.tsx:1177-1185 to the inline literal leaves all 14 new tests green. find/find.utils.test.ts:877
8. 13 TS2345 errors invisible to CI: as const makes findScope a readonly tuple. Both extension tsconfigs exclude *.test.ts and vitest sets no test.typecheck. pdpe.test.ts:2384 shows the fix. find/find.utils.test.ts:885
9. errored, stopped and exceeded all return [], so any throw silently turns the three new negative controls into no-ops. pdpe.test.ts:273

Medium:

10. The poll helper discards non-final batches (splice drains destructively) and is unbounded. pdpe.test.ts:271
11. The stated mechanism is wrong: it is read-path and per-text-node, not "on every write". find/find.utils.ts:744
12. Normalization sits in a UI adapter while platformScripture.findInScripture is public PAPI. Counterpoint worth weighing: ADR-0026 accepted UI-layer normalization, and there is one non-test caller today. find/find.utils.ts:760
13. .includes() is what hid finding 2; toEqual([' beginning ']) would have surfaced it. pdpe.test.ts:2410
14. Rewritten comment is inverted — should be "also matches"; the engine returns ["e","é","e","é","é"], and the file contradicts itself at :2350. pdpe.test.ts:2319
15. Version history commits record a query that never ran. find.web-view.tsx:1178
16. History and persisted last-term store the raw term, giving identical-looking rows. find.web-view.tsx:1178
17. A no-op second space tears down the search and re-scans the book for the same answer. find.web-view.tsx:1177
18. Preserve-case silently no-ops when a match starts with a space, now routine. find/find.utils.ts:765
19. No Architecture-Decisions.md entry, though ADR-0025 was recorded a day earlier for the same feature. PR-level
20. The deferral names no open ticket. find/find.utils.ts:754
21. Consolidation covered one of five polling copies (:2081, :2230, :2259, :2297 remain); the PR body's checked box says otherwise. None call abandonFindJob. pdpe.test.ts:257
22. Nothing tells the user the query was rewritten — no toast, hint or aria-live. PR-level

Low-Medium:

23. Search term collapsed, replace term not, so disk differs from preview. find.web-view.tsx:1178
24. preserveTrailingSpaces guards only the trailing edge, but leading-space matches are newly reachable. find/find.utils.ts:718
25. A backend PDP-engine test now imports the web view's UI adapter. pdpe.test.ts:9
26. The Storybook harness builds its regex from the raw term. Weakest item — it already diverges from the engine wholesale. PR-level

Low:

27. 'the Word' is unchanged by / {2,}/, / +/ and /\s+/ alike, so the test can't fail for what it names. find/find.utils.test.ts:897
28. The invisible-character test uses single characters with no adjacent whitespace, so it passes for any over-broad collapser. find/find.utils.test.ts:903
29. buildFindOptions now lands in the extension main bundle via platform-scripture-finder-pdpe.model.ts:31. find/find.utils.ts:760

Verified clear

buildFindOptions is the only production construction site; field mapping and useCallback deps are correct; the extracted mock is faithful; prettier and eslint are clean; 186 tests pass; no localization or papi.d.ts surface is touched.

One correction for the description: Replace inherits the collapse because it consumes find-result ranges, not because "find and replace share one buildSearchRegex call" — there is no such coupling.

> @katherinejensen00 reviewed 4 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2715#-:-P-uZeO24D0vaVx2y6zd:ba4kyc7)* status: all files reviewed, 1 unresolved discussion (waiting on jolierabideau).

___
*[`extensions/src/platform-scripture/src/find.web-view.tsx` line 1184 at r1](https://reviewable.io/reviews/paranext/paranext-core/2715#-P-uZp6u-_GVVvW-SEMB:-P-uZp6u-_GVVvW-SEMC:bnwk1h) ([raw file](https://github.com/paranext/paranext-core/blob/5dc4fbff532f0166a232a4237b0b49d8c4714340/extensions/src/platform-scripture/src/find.web-view.tsx#L1184)):*
> ```XML
>             isRegexAllowed,
>             searchTextType,
>             wordRestriction,
> ```

2. Replace fuses the surrounding words when the query carries padding. find.web-view.tsx:1184 (real sites :1833, :2034) · Critical

'   beginning    ' collapses to ' beginning ', which matches an 11-character span including both spaces. Replace splices the whole span. I ran it: In the beginning was the Word. replaced with start gives In thestartwas the Word. — identical with preserveCase on or off. isFindQueryValid only rejects all-whitespace queries and usfmChangesStructure compares block markers only, so nothing catches it. Power-mode-only limits the blast radius but doesn't remove it.

jolierabideau:
You were right about the premise, and it was worse than the 465-junk-matches version on my fixture. Running the real engine against USX shaped like `footnote-util-test.usj.data.ts`:

| query | result |
| --- | --- |
| `"Salvador.  porque"` raw | `completed, texts: ["Salvador.  porque"]` |
| the same query through `buildFindOptions` | `completed, texts: []` |

So the collapse didn't only add noise — it destroyed a match that worked on `main`. `usj-reader-writer.ts:1738` is exactly where it comes from, and the comment there says it outright. Findings 2 and 3 reproduce too: the padded query yields the span `" beginning "`, and all three word restrictions return zero.

I took the third path rather than either the collapse or unconditional `ignoreWhitespaceDifferences`. Both flags are now **user-togglable Find filters, defaulting to off**, mirroring PT9's `FindReplaceOptions`. The query reaches the engine byte-identical; tolerance is expressed only as options the engine applies when matching. That keeps exact invisible-character search — the reason the unconditional flag was rejected last round — while still giving the extra-spaces case a remedy. `ignoreDiacritics` was already implemented and tested at the engine level but unreachable from the UI, so it gains its control here as the sibling flag on the same PT9 dialog. Recorded as **ADR-0029** (finding 19).

**Fixed at the root:** 1, 2, 3, 4. **Dissolved with the rewrite:** 5, 6, 12, 15, 16, 17, 18, 20, 22, 23, 24, 27 — as you predicted.

**Individually addressed:**

- **7** — added `find.component.test.tsx` coverage of the UI wiring, the layer that was previously untestable.
- **8** — `BASE_STATE` typed as `FindUiState`; I also cleared the 25 pre-existing `TS2345`s in `find.utils.test.ts` (`buildSearchRegex` literals missing `scope`) so the file is genuinely clean rather than clean-in-the-new-block-only.
- **9, 10, 21** — one `pollFindJob` helper: accumulates every batch, bounded at 200 polls, and **throws** on any status but `completed`. All four remaining copies consolidated onto it.
- **11** — the TSDoc now describes the read-path, per-text-node mechanism instead of "on every write".
- **13, 14** — `toEqual` throughout; the inverted comment corrected against observed output, `['e','é','e','é','é']`.
- **25, 29** — `buildFindOptions` moved to `find/find-options.utils.ts`, so the PDP engine no longer pulls the UI adapter in through `find.utils.ts`, and the engine tests use raw `FindOptions`.
- **26** — the story harness now honors `ignoreWhitespaceDifferences` and its TSDoc says it does not emulate `ignoreDiacritics`.
- **28** — the invisible-character tests now use adjacent whitespace and assert both directions; the NBSP is written as an explicit `\u00A0` escape so it is visible in review.
- **30** — recorded in the ADR. I can't verify the ParatextData branch from here, but the approach no longer depends on the invariant either way.

**Three I did not simply comply with:**

1. **Finding 2 has a residual I pinned rather than fixed.** With the toggle on and a padded query, the match span still includes the padding, so Replace still fuses words. The user typed those spaces, nothing rewrites them silently, and PT9 does the same — so there's a test asserting `[' beginning ']` with a comment that the engine must report exactly the span it matched. Trimming the query strikes me as a product decision. Say the word if you read it as still-a-bug.
2. **Finding 8's CI gap is still open.** I fixed the errors but left the `**/*.test.ts` exclusion in `tsconfig.json` — it has a documented reason (importing `vitest` pulls Vite's type entry point, unresolvable under `moduleResolution: "Node"`), so lifting it repo-wide looks like its own piece of work rather than a review fix.
3. **Finding 21's "none call `abandonFindJob`"** — left as is. Every job the helper runs reaches `completed`, so there's nothing to abandon; the throw-on-non-completed change means a job that doesn't complete now fails the test loudly instead.

Also incidental, outside your list: `find.component.test.tsx` was passing `scope: 'currentBook'`, which isn't a valid `Scope`. Changed to `'book'` and confirmed the test still fails when the behavior it names breaks.

Your correction to the description is right — Replace inherits the behavior because it consumes find-result ranges, not through a shared `buildSearchRegex` call. The rewritten body says so.

Verified: typecheck clean, lint 0 errors, 501 tests pass (up from 490). I checked falsifiability rather than assuming it — dropping the flag in `buildFindOptions` fails 1 adapter test, ignoring the option in `buildSearchRegex` fails 7 engine tests, unwiring the UI setter fails 2 component tests.

katherinejensen00 (approved):
:lgtm:

> @katherinejensen00 reviewed 12 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2715#-:-P0-OPfS1LjWJ2fKPbGU:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on jolierabideau).

PR #2716, opened 2026-08-25
PT-4325: Registration reminder dot on the profile affordance

katherinejensen00:
## Screenshots

The running app, light theme — trigger dot, header warning line, row dot 
app-07-dot-settled

Dark theme + invalid registration with a **populated** profile (the reconciliation case) 
sb-dark-with-profile

Valid registration — no dot, no warning
app-02-popover-open

---

# Code Review Summary

**Branch**: pt-4325-registration-reminder-dot

**Base**: origin/main (`cbecad7f1cb`) — rebased 2026-08-25

**Date**: 2026-08-25

**Files changed**: 20

## Overview

PT-4325 adds an unobtrusive reminder that the user's Paratext registration is missing or invalid: a
small primary dot on the toolbar's user-profile trigger, a matching dot on the popover's "Profile &
registration" row, and a "Registration needed" line in the popover header. The signal is never
carried by color alone — the trigger's accessible label and tooltip change to match, and the row
carries screen-reader-only text.

Behind it is a new renderer store, `registration-validity-store`, that resolves registration validity
**once per session**. The first-run gate now consumes that shared probe instead of issuing its own.
That coupling is the most consequential part of this change: it is what lets the gate and the UI
agree — including letting the gate hand its deliberate "just re-registered, don't nag" suppression to
the UI via `publishRegistrationValidity`, which two independent probes could not do. It also avoids
~30 redundant JSON-RPC round trips on a cold start. `'unknown'` is treated as "don't nag" throughout:
it means the probe could not complete, never that the registration is bad.

## API Changes

No behavioural change to any public API. One documentation-only change reaches the published surface:
a TSDoc correction in `src/shared/services/network.service.ts` (`requestNoRetry` claimed its retry
loop emits "10 warning logs"; the actual retry site, `requestWithRetry` in `shared/data/rpc.model.ts`,
logs at `debug`). `lib/papi-dts/papi.d.ts` is regenerated to carry that corrected comment — it is
generated output, not a hand edit. No signatures, types, or command names changed. Verified untouched: `lib/platform-bible-react/`,
`lib/platform-bible-utils/`, and all `extensions/src/*/src/types/*.d.ts`. The new module exports
(`getRegistrationValidity`, `subscribeToRegistrationValidity`, `refreshRegistrationValidity`,
`publishRegistrationValidity`, `resetRegistrationValidityStore`, `useRegistrationValidity`) are
renderer-private and not re-exported through any `@papi/` module.

None. All four analysis passes independently reported zero critical findings.

- [x] **The popover stated registration twice from two sources that could contradict each other** —
      the header's "Not registered" came from name/email presence, the dot from probe validity. An
      expired code with a stored profile showed a friendly header plus an unexplained dot, and the
      only text explaining it was `sr-only`. _(fixed during review: header keeps name + email and
      gains a visible "Registration needed" line with its own dot whenever validity is invalid; the
      "Not registered" fallback is suppressed while that line shows so an empty-profile user sees one
      message, not two. 4 tests, revert-verified.)_
- [x] **The dot's `ring-background` halo was the wrong color in the states users actually see it** —
      `buttonVariants` `ghost` sets `tw:hover:bg-muted` and `tw:aria-expanded:bg-muted`, so the button
      is `--muted` on hover and for the entire time the popover is open; the toolbar is
      `tw:bg-transparent` over a muted-tinted `#root`. _(fixed during review: ring now tracks button
      state via `tw:group-hover/button:ring-muted tw:group-aria-expanded/button:ring-muted`; both
      variants confirmed present in the built stylesheet.)_
- [x] **`startBackgroundRegistrationRecheck` dropped the just-registered suppression** — it consumed
      the one-shot flag then returned early when the startup reminder was disabled, so a user who had
      turned the reminder off and just re-registered got nagged all session with no flag left to
      recover. The two sibling sites already handled this. _(fixed during review: publishes `'valid'`
      before the early return. 2 tests, revert-verified.)_
- [x] **A `publishRegistrationValidity` landing mid-probe was silently clobbered** — publish did not
      bump `generation`, so an older probe still passed its guard and overwrote the decision.
      _(fixed during review: publish now disowns in-flight probes, and slot cleanup moved to an
      explicit `inFlightId` so the store cannot wedge — guarding cleanup on `generation` instead would
      leave `inFlight` set forever. Both halves independently revert-verified.)_
- [x] **The dot introduces a "needs attention" visual pattern with no design-system home** — the
      library already ships two unused, non-theme-aware dot treatments (`Badge`'s
      `blueIndicator`/`mutedIndicator`) plus an unexported `AvatarBadge`, making this the third.
      _(fixed during review: `ReminderDot` keeps an inline rationale naming both rejected options and
      why, and directing the next caller to extend it rather than copy the class string. Promoting a
      shared themed indicator is left as a follow-up — see Suggested Review Focus.)_

### Minor — Consider

- [x] `resetRegistrationValidityStore` JSDoc still said "and clears listeners", contradicting the code
      and its own test _(fixed during review)_
- [x] Story `play` had no assertions, unlike sibling stories _(fixed during review: asserts the row
      dot and header warning, portal-aware via `within(document.body)`)_
- [x] Popover test's store fake cleared listeners, contradicting a documented invariant of the real
      store — it would keep passing if the real store regressed on exactly that point _(fixed during
      review: fake now mirrors the real reset)_
- [x] `%userProfile_registrationNeeded%` sat at the top of `LOCALIZED_STRING_KEYS`, which otherwise
      follows the popover's visual order _(fixed during review: moved next to
      `%userProfile_profileAndRegistration%`)_
- [x] Store tests never unsubscribed, leaving a deliberately-throwing subscriber registered for every
      later notify in the file _(fixed during review)_
- [x] The `isOpen` effect did two unrelated things and its `cancelled` cleanup guarded only one
      _(fixed during review: split into two effects)_
- [x] **Windows High Contrast made the dot invisible** — `background-color` and box-shadow are both
      suppressed under `forced-colors`, and the compensating signals (hover tooltip, `sr-only` text)
      don't reach a sighted High Contrast user _(fixed during review: `outline: 1px solid CanvasText`
      in the companion CSS file. **Note for the reviewer:** the first attempt used
      `tw:forced-colors:outline-1` and emitted **no CSS at all** — inert. Caught by checking the built
      stylesheet; the working rule is verified present there.)_
- [x] Row dot's position drifted with label length across locales _(fixed during review: `tw:ms-auto`
      pins it to the row end)_
- [x] Story decorator used `tw:bg-background`, flattering the dot's contrast versus the real
      muted-tinted surface _(fixed during review: `tw:bg-muted`, with the reasoning in the docstring)_
- [x] Story docs did not mention the new header warning line _(fixed during review)_
- [ ] ~~The component's `.catch` warn branch was uncovered~~ _(moot: the demo-mode refactor moved error
      handling into the hook, where `use-registration-validity.hook.test.tsx` already covers it)_
- [ ] ~~Fourth renderer store hand-rolling the same listener boilerplate; a `createRendererStore`
      helper would remove ~25 lines each~~ _(real duplication, but the fix edits three files this
      ticket has no business touching for zero behavior change — follow-up ticket)_
- [ ] ~~`RegistrationValidity` imported from `first-run.model` by a general-purpose store~~
      _(dependency direction genuinely is backwards; the only cycle-free home is
      `resolve-registration-validity.ts`, and moving it churns ~5 imports in files the first-run gate
      depends on for zero behavior change — follow-up ticket)_
- [ ] ~~The forced re-check on popover open is unthrottled~~ _(researched and judged correct as-is:
      the in-flight share already collapses rapid opens, and opening the popover is an explicit user
      action — rate-limiting a user-initiated retry would make the UI less responsive exactly when
      someone is trying to get an answer)_
- [ ] ~~The gate publishes its suppression after the probe already published, risking a dot flash~~
      _(traced: both `setValidity` calls occur in the same macrotask, so React 18 auto-batching
      collapses them to one render at the final value; on the `resolveInternal` path the toolbar is
      behind the modal wizard anyway)_
- [ ] **Chromatic will not publish this story.** `.chromatic-story-filter` is absent, so
      `chromatic.yml` falls back to `onlyStoryFiles: extensions/src/**/*.stories.tsx`; this story is
      under `src/renderer/`. The workflow also only triggers on PRs into `ai/main`. Pre-existing, not
      introduced here — but it means the visual deliverable is only reviewable by running Storybook
      locally.

- [x] **Storybook stories were silently broken twice, both found by the author testing them.**
      First: `beforeEach` was nested under `parameters`, where Storybook never invokes it — copied from
      `identify-step.component.stories.tsx`, which has the same latent bug. Moving it to the CSF
      annotation made it run, which exposed the second bug: `spyOn(commandService, 'sendCommand')`
      throws "Module namespace is not configurable in ESM" in Storybook 9. _(fixed during review:
      stories now drive a mock channel + `NormalModuleReplacementPlugin` — the repo's existing pattern
      for `@renderer/hooks/papi-hooks` — instead of spying. New files:
      `.storybook/mocks/command-service-mock-channel.ts`, `.storybook/mocks/command-service.ts`.)_
- [x] **`tags: ['autodocs']` broke every story.** The autodocs page renders all stories at once against
      two module-level singletons — the mock channel and the registration-validity store, which is
      module-global by production design and cannot be scoped per story. Each story's `beforeEach`
      reset the store and notified *every* mounted popover, so the last one flattened them all to
      `'unknown'` and no dot appeared anywhere. _(fixed during review: autodocs removed, with the
      reasoning recorded above `meta` so nobody re-adds it. `first-run-language-mock-channel.ts`
      documents the same hazard.)_
- [x] **No story covered the case the header reconciliation exists for** — invalid registration with a
      *populated* profile. _(fixed during review: added `RegistrationNeededWithProfile`, which asserts
      both that the email survives and that the warning line appears.)_

#### Shared Regions Modified

#### Extension Config Changes

None. No files under `extensions/` were changed, and no `package.json`, `tsconfig*.json`,
`.eslintrc*`, `webpack.config.*`, `.erb/configs/`, or `.github/workflows/` files were modified.

## Positive Observations

- `'unknown'` semantics are handled consistently everywhere: never cached, never allowed to downgrade
  a known answer, never rendered as a nag — the correct fail-open behavior, asserted by dedicated
  tests and a Storybook story.
- The in-flight share is correct in the cases that usually break: `force` bypasses the cache but not
  the share, `finally` frees the slot on the rejection path, and a generation guard stops a probe that
  outlived a reset from leaking into the next test or story.
- Listener isolation in `notifyListeners` prevents a real bug, not a hypothetical one: without it a
  throwing decorative-dot subscriber would escape into the shared probe promise and strand the user on
  the first-run "couldn't verify registration" error screen.
- Accessibility is handled properly rather than by color alone — `aria-hidden` dot, swapping
  accessible label and tooltip, `sr-only` row text, visible header line, `tw:end-1` for RTL, and a
  forced-colors outline. Each has a test or a verified CSS rule.
- Mocking sits at the right boundary: the resolver (the command boundary) is mocked rather than the
  internal store, so the hook tests exercise the real React↔store wiring, including StrictMode
  double-effects and an explicit unhandled-rejection assertion.
- Localization is complete and correct: both keys in `en.json` and `es.json`, correct alphabetical
  position, sentence case, registered in `LOCALIZED_STRING_KEYS`, no existing string mutated.
- Stories drive the real hook → store → resolver path via command spies rather than injected state,
  and include the `'unknown'` case most people would skip.

## Interview Notes

**Stated purpose**: show an unobtrusive reminder dot on the Paratext Registration affordance when the
user is unregistered or their registration is invalid (PT-4325, sub-task of PT-4323).

**Decisions the author made during review:**

- *Both signals should show, but must agree.* Chose to keep the header's identity (name + email) and
  add a visible warning line, over replacing the email or retiring the name/email check entirely —
  keeping the email is useful for confirming *which* registration is stale.
- *The ring should track button state* rather than being dropped.
- *Fix the `reminderSuppressed` inconsistency* rather than leave it documented.
- *Publish wins over an in-flight probe* — an explicit decision beats a probe that predates it.
- *Keep `ReminderDot` local and document why*, rather than editing the shared library in this PR.
- *Demo mode must not probe*, so `first-run-store`'s documented promise to bypass the real
  registration backend entirely stays true. This also resolved a separate finding: the hook now
  exposes `refresh`, so the component depends on one abstraction instead of the hook plus the service.
- *Revert the `CLAUDE.md` change from this PR* and keep the no-plans guidance in machine-local memory
  instead — `CLAUDE.md` governs all contributors and per `.claude/rules/where-to-add-guidance.md`
  needs broader dev review, so it should not land inside a feature squash-merge.
- *Fix the stale `requestNoRetry` TSDoc here* even though it is unrelated to PT-4325, because it had
  already caused one wrong architectural justification.

**Reviewer-facing corrections made during this review:**

- **ADR `registration-validity-once-per-session` asserted something false.** It justified the shared probe by citing "10 main-process
  **warning** logs" per command from `requestNoRetry`'s TSDoc. The actual retry site,
  `requestWithRetry` in `shared/data/rpc.model.ts`, logs at **`debug`**. The traffic is real
  (`MAX_REQUEST_ATTEMPTS = 10` at 1s apart); the log spam is not. The ADR now leads with the
  correctness rationale (one answer per session is what lets the gate and UI agree, and what makes the
  just-registered suppression expressible at all), cites by symbol per the docs-durability rule, and
  the misleading upstream TSDoc is corrected in this same change.
- **Two pre-existing first-run tests had silently stopped being falsifiable.** Verified by experiment:
  with *both* gate dedupe guards deleted, all 56 tests still passed, because the new shared store's
  dedupe satisfied their probe-count assertions. Those guards prevent repeated *durable writes*, which
  probe-dedupe does not. Both now assert on `resolveInternal`'s settings reads, and were confirmed to
  fail without the guards.
- **The first High Contrast fix was inert.** `tw:forced-colors:outline-1` emitted no CSS; three other
  classes from the same file did, which proved the scan was fine and the variant invalid. Replaced
  with a hand-written rule and verified present in the built stylesheet.

**No areas were deferred to AI or left unexplained by the author.**

## In-Review Quality Check

All checks run synchronously after the in-review changes:

- `npm run typecheck:core` — **exit 0**
- `npx eslint --ext .ts,.tsx src/` — **exit 0**
- `npx stylelint "src/renderer/**/*.css"` — **exit 0**
- `npx prettier --write` on all changed files — already formatted
- `npm run test:core` — **1961 passed / 151 files, exit 0** (re-run after rebasing onto main)
- `npm run storybook:build` — **exit 0**, 5 stories, no compile errors
- `dotnet test c-sharp-tests/` — 1580 passed, 0 failed (no C# was changed)

### Verified in a real browser, not just built

Every story was loaded in Chromium and asserted on, including navigating back and forth between them
in one session (module-global state persists across story switches) — 6/6 correct, zero console
errors. Dark mode confirmed by applying `body.dark`, which is how the app applies it.

### Verified in the running application

The author authorised temporarily removing the machine's Paratext registration. It was backed up,
sha256-verified, removed, exercised, then restored and re-verified identical; the plaintext backup
copy was deleted afterwards so no second copy of the registration code was left behind.

| | Registration valid | Registration removed |
| --- | --- | --- |
| `doesUserHaveValidRegistration` | `true` | `false` |
| Toolbar dot | absent | **present** |
| Trigger `aria-label` | "User profile" | "User profile — registration needed" |
| Popover row dot | absent | **present** |
| Header warning line | absent | **"Registration needed"** |

This also confirmed the gate and the dot agree: with registration removed the first-run wizard raised
at "Enter your registration information" *and* the dot was live behind it, both reading the same
shared probe — the coupling ADR `registration-validity-once-per-session` describes, working end to end.

**Two caveats the reviewer should know:**

1. **Full-repo `npm run typecheck` fails** in `platform-scripture-editor` on
   `@eten-tech-foundation/platform-editor` exports. This is a property of the dev-package setup, not
   of this branch: `link-dev-packages` succeeds and still publishes `src`-only packages that declare
   `types: dist/index.d.ts`. Worse, with those packages linked the **app build fails outright** (9
   webpack errors — the published `scripture-utilities/src/index.ts` imports `./converters/usj/*.js`
   files that are not in the package). The app only runs against the registry copies.
   `typecheck:core`, which covers every file this change touches, is clean.
2. **The core suite is flaky under CPU contention.** Controlled comparison on a quiet machine: clean
   tree 3/3 green, this branch 3/3 green. Failures appeared only in runs contaminated by a concurrent
   Storybook build or repo-wide `eslint --fix`, and always in files with no import path to this change
   (`theme.service-host`, `web-view.service-host`, `scroll-group.service-host`, `util`, `audit-cli`).
   `theme.service-host.test.ts` alone has 11+ `vi.waitFor` calls on the default 1s timeout.
   Pre-existing, but it will bite a loaded CI runner.

## Rebase note (2026-08-25)

Rebased three times while this PR was open; `origin/main` added ADRs each time and this entry's
number collided on every one (**0014 -> 0026 -> 0027**). Rather than renumber a fourth time, the
entry is now referenced by a **kebab-case slug**, `registration-validity-once-per-session`, with no
number. A slug is stable under reordering and cannot collide on rebase.

Scope is deliberately limited to this entry: the other 28 ADRs keep their numbers, and the 22
numbered references elsewhere in the repo are untouched. Converting the whole log is a convention
change to a shared standards doc and belongs in its own reviewed PR, not riding inside a feature
change — see `.claude/rules/where-to-add-guidance.md`. This entry is offered as the proposal.

`lib/papi-dts/papi.d.ts` was regenerated with `npm run build:types` after each rebase rather than
trusting the text merge; it came back byte-identical to the merged result every time.

## Suggested Review Focus

- [x] ~~**Unverified assumption the whole feature rests on**~~ — **now confirmed by observation.** With
      the registration file removed, `doesUserHaveValidRegistration` returned `false` and the reminder
      appeared throughout. Acceptance criterion #1 is satisfied against real ParatextData.
- [ ] **The gate coupling** — `first-run-store` now takes its registration answer from a store the
      toolbar can also drive, rather than a call it issued itself. Believed equivalent (same local
      read, same startup, `'unknown'` never cached so Retry still re-probes) and the tests agree, but
      it is the highest-consequence change here and worth a second opinion. See ADR `registration-validity-once-per-session`.
- [ ] **A brief window after "Continue without registration" shows no dot.** Observed in the running
      app: immediately after dismissing the wizard the trigger had no dot and the plain label; seconds
      later both were correct. The store is evidently re-resolving. Observed but *not* explained —
      worth understanding before merge.
- [ ] **RTL was not checked.** Light and dark are verified in the app and Storybook; the dot uses
      `tw:end-1`/`tw:ms-auto` so it should mirror, but nobody has looked.
- [ ] **Power mode now issues one registration command per launch** where it previously issued zero
      (the toolbar mounts unconditionally, so the hook's probe is what makes the dot work there).
      Demo mode is explicitly exempt. Confirm that trade is acceptable.
- [ ] **Follow-up tickets to raise**: promote a themed, exported indicator into `platform-bible-react`
      and fix the misleading `blueIndicator`/`mutedIndicator` stories; add a `.chromatic-story-filter`
      so `src/renderer/**` stories get visual review; extract a shared `createRendererStore`; move
      `RegistrationValidity` out of `first-run.model`.
- [ ] **UX sign-off** on the dot's weight, color, and the new header warning line — the pattern has no
      design-system entry yet. Note `tw:bg-primary` inverts with the theme, so the "neutral blue nudge"
      renders near-white in dark mode. Consistent theming, but a deliberate look worth confirming.
- [ ] **`identify-step.component.stories.tsx` carries both Storybook bugs found here** —
      `parameters.beforeEach` (never runs) and `spyOn` on the command service (throws when it does).
      Its stories are almost certainly not doing what they claim. Worth its own ticket.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2716)


tombogle:
*[`.context/standards/Architecture-Decisions.md` line 1385 at r2](https://reviewable.io/reviews/paranext/paranext-core/2716#-P-tlTRFBisQtI9CGOl3:-P-tlTRFBisQtI9CGOl4:bea9wkp) ([raw file](https://github.com/paranext/paranext-core/blob/16ee3d75050cae7078c84fbd29483460befc51f4/.context/standards/Architecture-Decisions.md#L1385)):*
> ```Markdown
>   the gate publishes its suppressed answer — and because that publish is not sticky, opening the
>   profile popover force-re-probes and can bring the dot back for the rest of that session; on the one
>   path where the gate never publishes (Simple + completed + reminder suppressed + just registered) it
> ```

I have not independently confirmed the actual fix (though I see the claim to that effect), but Claude says that that combination routes to `startBackgroundRegistrationRecheck` in [first-run-store.ts:363-370](vscode-webview://0qjri44f8rnmajfhpak4aa0sn70kgrtrhatje8nfj023rgb6fg1t/src/renderer/services/first-run-store.ts#L363-L370):

```ts
if (reminderSuppressed) {
  if (justRegistered) publishRegistrationValidity('valid');
  return;
}
```

This branch _does_ publish — it was added as part of this same PR's "Important" fix ("`startBackgroundRegistrationRecheck` dropped the just-registered suppression"), per the PR body's findings list. The ADR's Consequences section describes the pre-fix behavior (unconditional `return` with no publish) and wasn't updated after that fix landed. So bullet 3 is now false — the exact same class of error the PR already caught once (the "10 warning logs" claim) slipped through a second time, in the same section, for the same reason: prose not reconciled against a fix applied during review.

tombogle (changes requested):
> @tombogle partially reviewed 3 files and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2716)* status: 2 of 20 files reviewed, 2 unresolved discussions (waiting on katherinejensen00).

___
*[`.context/standards/Architecture-Decisions.md` line 1436 at r3](https://reviewable.io/reviews/paranext/paranext-core/2716#-P-y3m8l5N7SFncSKgpT:-P-y3m8l5N7SFncSKgpU:b13fyjs) ([raw file](https://github.com/paranext/paranext-core/blob/3cdad596cb4d70763c0fbcd25510d26dc7628aa2/.context/standards/Architecture-Decisions.md#L1436)):*
> ```Markdown
> ## registration-validity-once-per-session
> 
> > Referenced by slug, not a number. Numbers in this log collide on every rebase — this entry hit
> ```

I don't think this explanation and the description of what a slug is belongs here. It probably does belong in this document, near the top.) Here, we probably just want to say something brief (boilerplate to use in every "renamebered" ADR):
`(Formerly numbered 0014, 0026, 0027)`

katherinejensen00 (commented):
While checking the flag lifecycle you asked about, I found a pre-existing bug in the same mechanism.

resolveInternal consumed the one-shot just-registered flag before probing, so an 'unknown' probe
spent it without ever using it. The user landed on the "couldn't verify registration" screen, and
the transient 'invalid' the flag exists to absorb then arrived on the Retry unguarded — mis-routing
a user who had just registered successfully back to the language step instead of resuming at sync
consent, which is precisely what markJustRegistered's own TSDoc says the flag prevents.

Fixed by separating clearing the durable flag from remembering the answer. The flag is still cleared
on its first read, so it still grants exactly one launch of trust and cross-launch behavior is
unchanged — the existing next-launch test passes untouched, and it's the test that would catch a
leak. The answer now survives in memory for the rest of that startup, which is what covers the
Retry.

New test, verified to fail without the fix with exactly that symptom:
expected { kind: 'wizard', step: 'language' } to deeply equal { kind: 'wizard', step: 'syncConsent' }

> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2716#-:-P04enwZ9TDUMzjOCEX-:b-64xlco)* status: 0 of 20 files reviewed, 2 unresolved discussions (waiting on tombogle).

___
*[`.context/standards/Architecture-Decisions.md` line 1385 at r2](https://reviewable.io/reviews/paranext/paranext-core/2716#-P-tlTRFBisQtI9CGOl3:-P04ecSuDyB_SxMVPnGG:b-xs33nj) ([raw file](https://github.com/paranext/paranext-core/blob/16ee3d75050cae7078c84fbd29483460befc51f4/.context/standards/Architecture-Decisions.md#L1385)):*

I have not independently confirmed the actual fix (though I see the claim to that effect), but Claude says that that combination routes to `startBackgroundRegistrationRecheck` in [first-run-store.ts:363-370](vscode-webview://0qjri44f8rnmajfhpak4aa0sn70kgrtrhatje8nfj023rgb6fg1t/src/renderer/services/first-run-store.ts#L363-L370):

```ts
if (reminderSuppressed) {
  if (justRegistered) publishRegistrationValidity('valid');
  return;
}
```

This branch _does_ publish — it was added as part of this same PR's "Important" fix ("`startBackgroundRegistrationRecheck` dropped the just-registered suppression"), per the PR body's findings list. The ADR's Consequences section describes the pre-fix behavior (unconditional `return` with no publish) and wasn't updated after that fix landed. So bullet 3 is now false — the exact same class of error the PR already caught once (the "10 warning logs" claim) slipped through a second time, in the same section, for the same reason: prose not reconciled against a fix applied during review.
</blockquote></details>

You're right, and thank you for checking the code rather than the prose. That bullet described the
pre-fix behavior and I never went back to it. Removed.

I also added a sentence to the Decision section stating the invariant positively, so the claim now
lives somewhere a future reader will check against the code rather than only as an absence.

Your diagnosis pointed at two more instances of the same thing, both fixed here:

- registration-validity-store.ts's module TSDoc still carried the "10 warning logs / requestNoRetry"
  claim that this same commit corrects elsewhere. It now cites requestWithRetry in rpc.model.ts and
  notes those retries log at debug.
- The Storybook command-service mock justified resolving `undefined` as equivalent to the real
  service's behavior. It isn't, and the comment now says so.
___
*[`.context/standards/Architecture-Decisions.md` line 1436 at r3](https://reviewable.io/reviews/paranext/paranext-core/2716#-P-y3m8l5N7SFncSKgpT:-P04ei8Y4n7Ixkd85CWa:boq2mk7) ([raw file](https://github.com/paranext/paranext-core/blob/3cdad596cb4d70763c0fbcd25510d26dc7628aa2/.context/standards/Architecture-Decisions.md#L1436)):*

I don't think this explanation and the description of what a slug is belongs here. It probably does belong in this document, near the top.) Here, we probably just want to say something brief (boilerplate to use in every "renamebered" ADR):
`(Formerly numbered 0014, 0026, 0027)`
</blockquote></details>

Agreed — the rationale didn't belong in the entry. Good news: while this was sitting, jolierabideau
landed exactly that change on main (#2720, d9ff7654b07), converting the whole log to
`adr-<slug>: {short title}` with the explanation at the top of the document, right where you said it
belonged.

So rather than adding my own boilerplate, this entry now simply follows main's format:

  ## adr-registration-validity-once-per-session: Registration validity resolves once per session, in
  a store the first-run gate and the UI share

One deliberate deviation from your literal suggestion: no `Formerly:` line. Main reserves that for
Agreed — the rationale didn't belong in the entry. Good news: while this was sitting, jolierabideau
landed exactly that change on main (#2720, d9ff7654b07), converting the whole log to
`adr-<slug>: {short title}` with the explanation at the top of the document, right where you said it
belonged.

So rather than adding my own boilerplate, this entry now simply follows main's format:

  ## adr-registration-validity-once-per-session: Registration validity resolves once per session, in
  a store the first-run gate and the UI share

One deliberate deviation from your literal suggestion: no `Formerly:` line. Main reserves that for
entries that really did merge under a number, and explicitly says not to add it to new entries. This
one never landed as 0014/0026/0027 — those were collisions on an unmerged branch, and each of those
numbers now belongs to a different live entry. Happy to add the line if you'd rather have the trail.

The four code cross-references now cite the prefixed slug in main's style, and I've rebased onto
current main so the file matches rather than conflicts.

tombogle (changes requested):
> @tombogle partially reviewed 13 files, made 7 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2716)* status: 10 of 20 files reviewed, 6 unresolved discussions (waiting on katherinejensen00).

___
*[`.context/standards/Architecture-Decisions.md` line 1385 at r2](https://reviewable.io/reviews/paranext/paranext-core/2716#-P-tlTRFBisQtI9CGOl3:-P09CsZq-7zn5Gm6yPNh:bjmcwou) ([raw file](https://github.com/paranext/paranext-core/blob/16ee3d75050cae7078c84fbd29483460befc51f4/.context/standards/Architecture-Decisions.md#L1385)):*

You're right, and thank you for checking the code rather than the prose. That bullet described the
pre-fix behavior and I never went back to it. Removed.

I also added a sentence to the Decision section stating the invariant positively, so the claim now
lives somewhere a future reader will check against the code rather than only as an absence.

Your diagnosis pointed at two more instances of the same thing, both fixed here:

- registration-validity-store.ts's module TSDoc still carried the "10 warning logs / requestNoRetry"
  claim that this same commit corrects elsewhere. It now cites requestWithRetry in rpc.model.ts and
  notes those retries log at debug.
- The Storybook command-service mock justified resolving `undefined` as equivalent to the real
  service's behavior. It isn't, and the comment now says so.
</blockquote></details>

So is this ADR saying that even with the latest fixes, the user can still get into a state where the UI gives misleading information? That's what it sounds like when I read it.
___
*[`.context/standards/Architecture-Decisions.md` line 1436 at r3](https://reviewable.io/reviews/paranext/paranext-core/2716#-P-y3m8l5N7SFncSKgpT:-P09BsEq7GB7vt3soh87:bo4pvgp) ([raw file](https://github.com/paranext/paranext-core/blob/3cdad596cb4d70763c0fbcd25510d26dc7628aa2/.context/standards/Architecture-Decisions.md#L1436)):*

Agreed — the rationale didn't belong in the entry. Good news: while this was sitting, jolierabideau
landed exactly that change on main (#2720, d9ff7654b07), converting the whole log to
`adr-<slug>: {short title}` with the explanation at the top of the document, right where you said it
belonged.

So rather than adding my own boilerplate, this entry now simply follows main's format:

  ## adr-registration-validity-once-per-session: Registration validity resolves once per session, in
  a store the first-run gate and the UI share

One deliberate deviation from your literal suggestion: no `Formerly:` line. Main reserves that for
Agreed — the rationale didn't belong in the entry. Good news: while this was sitting, jolierabideau
landed exactly that change on main (#2720, d9ff7654b07), converting the whole log to
`adr-<slug>: {short title}` with the explanation at the top of the document, right where you said it
belonged.

So rather than adding my own boilerplate, this entry now simply follows main's format:

  ## adr-registration-validity-once-per-session: Registration validity resolves once per session, in
  a store the first-run gate and the UI share

One deliberate deviation from your literal suggestion: no `Formerly:` line. Main reserves that for
entries that really did merge under a number, and explicitly says not to add it to new entries. This
one never landed as 0014/0026/0027 — those were collisions on an unmerged branch, and each of those
numbers now belongs to a different live entry. Happy to add the line if you'd rather have the trail.

The four code cross-references now cite the prefixed slug in main's style, and I've rebased onto
current main so the file matches rather than conflicts.
</blockquote></details>

Don't really understand the comment about "no `Formerly: line", when I can see "Formerly"" right there, but whatever. The point is that it's tright now!
___
*[`src/renderer/components/first-run/steps/identify-step.component.tsx` line 264 at r3](https://reviewable.io/reviews/paranext/paranext-core/2716#-P-zSspX14fsxJzP2lt-:-P-zSspY0I6oD_XjH7pY:bscxhsw) ([raw file](https://github.com/paranext/paranext-core/blob/3cdad596cb4d70763c0fbcd25510d26dc7628aa2/src/renderer/components/first-run/steps/identify-step.component.tsx#L264)):*
> ```XML
>       markJustRegistered();
>       // Correct THIS session too. The toolbar (mounted behind the wizard) has already cached a
>       // definitive 'invalid', and the restart below is best-effort — if it fails, the user stays in
> ```

I believe this comment means:

If the restart fails and we didn't publish `'valid'` here, the user would remain in the current session with the stale `'invalid'` state, causing the reminder dot to continue nagging them.

If my interpretation is correct, it would probably be better to reword the comment thusly.
___
*[`src/renderer/components/first-run/steps/identify-step.component.test.tsx` line 238 at r3](https://reviewable.io/reviews/paranext/paranext-core/2716#-P-zW0Zs8vsisOoskfk8:-P-zW0Zs8vsisOoskfk9:b-uwukgn) ([raw file](https://github.com/paranext/paranext-core/blob/3cdad596cb4d70763c0fbcd25510d26dc7628aa2/src/renderer/components/first-run/steps/identify-step.component.test.tsx#L238)):*
> ```XML
>     publishRegistrationValidity('invalid');
>     mockSendCommand
>       .mockResolvedValueOnce(true) // validateParatextRegistrationData
> ```

I think this would be clearer as a command-aware mock rather than using `mockResolvedValueOnce()`. As written, the return values are implicitly tied to call order, but there doesn't appear to be an assertion that the commands were actually called in that order. If the order changes, the test will just start returning the wrong values for the commands, which could produce a confusing failure.

I'd suggest:

```
mockSendCommand.mockImplementation(async (commandName) => {
  switch (commandName) {
    case 'validateParatextRegistrationData':
      return true;
    case 'setParatextRegistrationData':
      return undefined;
    default:
      throw new Error(`Unexpected command: ${commandName}`);
  }
});
```

That makes the mock's intent explicit and also gives us a useful failure if the code sends an unexpected command.

If the order of these commands is actually part of the behavior we're testing, I'd still make the mock command-aware and assert the order separately:

```
mockSendCommand.mockImplementation(async (commandName) => {
  switch (commandName) {
    case 'validateParatextRegistrationData':
      return true;
    case 'setParatextRegistrationData':
      return undefined;
    default:
      throw new Error(`Unexpected command: ${commandName}`);
  }
});

// ... exercise the code ...

expect(mockSendCommand.mock.calls.map(([command]) => command)).toEqual([
  'validateParatextRegistrationData',
  'setParatextRegistrationData',
]);
```

That way the two concerns are explicit: the mock says what each command returns, and the assertion says whether their ordering matters.

I see that this is not a novel approach in this file and that this test follows an established pattern, so one option is to push back and fix this in a separate pass rather than increasing the scope of this PR with broader test cleanup.
___
*[`src/renderer/components/first-run/steps/identify-step.component.test.tsx` line 260 at r3](https://reviewable.io/reviews/paranext/paranext-core/2716#-P-zX72n6rqLWZgzy3lR:-P-zX72n6rqLWZgzy3lS:b-lnu2un) ([raw file](https://github.com/paranext/paranext-core/blob/3cdad596cb4d70763c0fbcd25510d26dc7628aa2/src/renderer/components/first-run/steps/identify-step.component.test.tsx#L260)):*
> ```XML
>     await user.click(screen.getByRole('button', { name: /save and restart/i }));
> 
>     // Without this the reminder dot would nag about the registration the user just fixed.
> ```

Comment could be a bit clearer and more precise:



_Code snippet:_
```XML
// Verify that the cached registration state was corrected so the reminder dot won't nag.
```
___
*[`src/renderer/components/user-profile-popover/user-profile-popover.component.tsx` line 102 at r3](https://reviewable.io/reviews/paranext/paranext-core/2716#-P-zYDWc76ntarU6hqwd:-P-zYDWd9OOMJ_nGZpiX:bsdtj05) ([raw file](https://github.com/paranext/paranext-core/blob/3cdad596cb4d70763c0fbcd25510d26dc7628aa2/src/renderer/components/user-profile-popover/user-profile-popover.component.tsx#L102)):*
> ```XML
>  * theme, and `AvatarBadge` is sized off `Avatar`'s group data-attributes and is not exported. That
>  * makes this the third dot treatment in the app — promoting a themed, exported indicator (and
>  * fixing those Badge variants) is worth its own change, since it touches every consumer. Until
> ```

Probably should add a Jira issue for the proposed change and link it as related to PT-4325, and maybe include a mention of it here in this comment as a TODO or ENHANCE.
___
*[`src/renderer/components/user-profile-popover/user-profile-popover.stories.tsx` line 36 at r3](https://reviewable.io/reviews/paranext/paranext-core/2716#-P-zOD4o7vwqsuDLAQts:-P-zOD4o7vwqsuDLAQtt:bll3lea) ([raw file](https://github.com/paranext/paranext-core/blob/3cdad596cb4d70763c0fbcd25510d26dc7628aa2/src/renderer/components/user-profile-popover/user-profile-popover.stories.tsx#L36)):*
> ```XML
>     resetRegistrationValidityStore();
>     setCommandServiceMock((commandName) =>
>       commandName === 'paratextRegistration.doesUserHaveValidRegistration' ? isValid : profile,
> ```

This is just a mock for a story, but I'm a little worried it could go stale in a way that makes it hard to tell what's really happening. We neither document which other command we're hoping to receive and respond to, nor do we guarantee in this code that we got it, rather than some third, unexpected command. I'd suggest something like this instead:



_Code snippet:_
```XML
setCommandServiceMock((commandName) => {
  switch (commandName) {
    case 'paratextRegistration.doesUserHaveValidRegistration':
      return isValid;
    case 'paratextRegistration.getUserProfile':
      return profile;
    default:
      throw new Error(`Unexpected command: ${commandName}`);
  }
});
```

katherinejensen00:
Thanks — all seven addressed in 9997ed5. Point by point:

**ADR: "can the user still get into a state where the UI gives misleading information?"**

Yes, briefly, and you were right that the entry didn't bound the claim. I've added that bound to
Consequences: the only thing either imperfect case can get wrong is *whether the reminder dot is
showing*. No registration detail ever renders incorrectly, nothing the user can do is gated on the
dot, and the authoritative check — the first-run gate that re-raises the wizard on an invalid
registration — still runs every launch against a store that starts each session at `'unknown'`, so
no answer is carried across launches. A stale dot is a nag to ignore or clear by opening the
popover, never a state you can be stuck in.

**ADR: the "Formerly" confusion**

My fault for being unclear. The `Formerly:` line you can see is in the document's *header*, where
main's #2720 explains the slug convention generally. My entry doesn't carry one, because it never
actually merged under 0014/0026/0027 — those were collisions on this unmerged branch, and each of
those numbers now belongs to a different live entry. Agreed it's right as it stands.

**identify-step.component.tsx:264 — comment wording**

Your interpretation is exactly right, and the comment now says it that way: without the publish, a
failed best-effort restart leaves the user in this session with the stale `'invalid'`, which keeps
the dot nagging about the registration they just fixed.

**identify-step.component.test.tsx:238 — command-aware mock**

Already done, and it crossed with your review — r3 predates it. The whole file now routes through a
`mockCommands` helper that switches on the command name, which is the shape you sketched. I've also
taken the stricter half of your suggestion: it now names `platform.restart` explicitly and *rejects*
any unlisted command rather than answering `undefined`, so a component that grows a dependency fails
loudly instead of having a silent non-answer misread as a real one. All 29 tests still pass, which
also confirms the helper covers every command IdentifyStep actually sends.

**identify-step.component.test.tsx:260 — comment**

Taken verbatim.

**user-profile-popover.component.tsx:102 — ticket for the shared indicator**

Marked with a `TODO` naming the follow-up (promote a themed, exported indicator, fix the `Badge`
`blueIndicator`/`mutedIndicator` variants, delete the local copy). The key is a `PT-XXXX`
placeholder for now — I'll swap in the real one once the ticket is filed and linked to PT-4325.

**user-profile-popover.stories.tsx:36 — responder could go stale**

Agreed, and your worry was better founded than it looked: the command you guessed, `getUserProfile`,
isn't the real name — it's `paratextRegistration.getParatextRegistrationData`. The old ternary hid
that behind an `else`. It's now a switch naming each command, rejecting anything unexpected. I did
add the popover's two menu-action commands (`showParatextRegistration`, `showInternetSettings`)
rather than letting them throw, so clicking around a story by hand stays inert instead of blowing
up.

---

**One thing you didn't ask about, found while verifying the above.** The three IdentifyStep stories
using `spyOn(commandService, 'sendCommand')` could never have worked, on main or here. Webpack
builds a module namespace with non-configurable getters (`DefinePropertyGettersRuntimeModule` omits
`configurable`) and tags it `Symbol.toStringTag = 'Module'` (`MakeNamespaceObjectRuntimeModule`) —
exactly the pair that makes `storybook/test`'s `spyOn` throw "Module namespace is not configurable
in ESM". Their `beforeEach` threw. Nothing caught it because the root workspace has no Storybook
vitest project, so root stories never execute in `npm test`, and Chromatic is label-gated.

They're migrated to the same channel now, and I verified it against a live Storybook rather than by
argument: all six IdentifyStep stories' play functions pass, and the registry-link story's `href`
resolves to the responder's `https://registry-dev.paratext.org`.

CI's earlier red was unrelated — the macOS runner died inside `npm install` with an `EEXIST` on its
npm cache, and the matrix cancelled the other two. Re-run is green on all three.

katherinejensen00:
Ticket filed: **PT-4493** — "Promote a themed, exported reminder indicator and fix the Badge
indicator variants" (Dev Task, linked *relates to* PT-4325). The `TODO` in
`user-profile-popover.component.tsx` now names it instead of the placeholder, so the local
`ReminderDot` has a searchable link to the follow-up.

It captures the two blockers behind the third dot treatment — `Badge`'s
`blueIndicator`/`mutedIndicator` hardcoding `bg-blue-400`/`bg-zinc-400` past the theme, and
`AvatarBadge` not being exported — plus migrating the existing treatments and deleting the local
copy.

katherinejensen00 (commented):
> @katherinejensen00 made 7 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2716)* status: 8 of 21 files reviewed, 6 unresolved discussions (waiting on tombogle).

___
*[`.context/standards/Architecture-Decisions.md` line 1385 at r2](https://reviewable.io/reviews/paranext/paranext-core/2716#-P-tlTRFBisQtI9CGOl3:-P09g63H1RXai0i93pES:b-hj2jd3) ([raw file](https://github.com/paranext/paranext-core/blob/16ee3d75050cae7078c84fbd29483460befc51f4/.context/standards/Architecture-Decisions.md#L1385)):*

So is this ADR saying that even with the latest fixes, the user can still get into a state where the UI gives misleading information? That's what it sounds like when I read it.
</blockquote></details>

Yes, briefly — and you're right that the entry stated that without bounding it, which made it sound worse than it is. I've added the bound to Consequences.

The only thing either imperfect case can get wrong is whether the reminder dot is showing. No registration detail is ever rendered incorrectly, nothing the user can do is gated on the dot, and the authoritative check — the first-run gate that re-raises the wizard on an invalid registration — still runs on every launch against a store that starts each session at 'unknown', so no answer is carried across launches. A stale dot is a nag to ignore or clear by opening the popover, never a state the user can be stuck in or misled by.
___
*[`.context/standards/Architecture-Decisions.md` line 1436 at r3](https://reviewable.io/reviews/paranext/paranext-core/2716#-P-y3m8l5N7SFncSKgpT:-P09fUgu-5yJHQnYP4cM:bbzf7u2) ([raw file](https://github.com/paranext/paranext-core/blob/3cdad596cb4d70763c0fbcd25510d26dc7628aa2/.context/standards/Architecture-Decisions.md#L1436)):*

Don't really understand the comment about "no `Formerly: line", when I can see "Formerly"" right there, but whatever. The point is that it's tright now!
</blockquote></details>

My wording was confusing, sorry. The "Formerly" you're seeing is in the document header, where main's #2720 explains the slug convention in general terms. My entry itself doesn't carry one, because it never actually merged under 0014/0026/0027 — those were collisions on this unmerged branch, and each of those numbers now belongs to a different live entry. Agreed it reads correctly as it stands.
___
*[`src/renderer/components/first-run/steps/identify-step.component.tsx` line 264 at r3](https://reviewable.io/reviews/paranext/paranext-core/2716#-P-zSspX14fsxJzP2lt-:-P09fjHpEo_Rf0RK3QOD:b-7kkli9) ([raw file](https://github.com/paranext/paranext-core/blob/3cdad596cb4d70763c0fbcd25510d26dc7628aa2/src/renderer/components/first-run/steps/identify-step.component.tsx#L264)):*

I believe this comment means:

If the restart fails and we didn't publish `'valid'` here, the user would remain in the current session with the stale `'invalid'` state, causing the reminder dot to continue nagging them.

If my interpretation is correct, it would probably be better to reword the comment thusly.
</blockquote></details>

Your interpretation is exactly right, and the comment now says it that way rather than making the reader derive it:

Correct THIS session too, not just the next one. The toolbar mounted behind the wizard has already cached a definitive 'invalid', and the restart below is best-effort: if it fails, the user stays in this session. Without this publish, that stale 'invalid' would keep the reminder dot nagging about the registration they just fixed.
___
*[`src/renderer/components/first-run/steps/identify-step.component.test.tsx` line 238 at r3](https://reviewable.io/reviews/paranext/paranext-core/2716#-P-zW0Zs8vsisOoskfk8:-P09foKE5PwSHlQGtI0W:b-gsbx9d) ([raw file](https://github.com/paranext/paranext-core/blob/3cdad596cb4d70763c0fbcd25510d26dc7628aa2/src/renderer/components/first-run/steps/identify-step.component.test.tsx#L238)):*

I think this would be clearer as a command-aware mock rather than using `mockResolvedValueOnce()`. As written, the return values are implicitly tied to call order, but there doesn't appear to be an assertion that the commands were actually called in that order. If the order changes, the test will just start returning the wrong values for the commands, which could produce a confusing failure.

I'd suggest:

```
mockSendCommand.mockImplementation(async (commandName) => {
  switch (commandName) {
    case 'validateParatextRegistrationData':
      return true;
    case 'setParatextRegistrationData':
      return undefined;
    default:
      throw new Error(`Unexpected command: ${commandName}`);
  }
});
```

That makes the mock's intent explicit and also gives us a useful failure if the code sends an unexpected command.

If the order of these commands is actually part of the behavior we're testing, I'd still make the mock command-aware and assert the order separately:

```
mockSendCommand.mockImplementation(async (commandName) => {
  switch (commandName) {
    case 'validateParatextRegistrationData':
      return true;
    case 'setParatextRegistrationData':
      return undefined;
    default:
      throw new Error(`Unexpected command: ${commandName}`);
  }
});

// ... exercise the code ...

expect(mockSendCommand.mock.calls.map(([command]) => command)).toEqual([
  'validateParatextRegistrationData',
  'setParatextRegistrationData',
]);
```

That way the two concerns are explicit: the mock says what each command returns, and the assertion says whether their ordering matters.

I see that this is not a novel approach in this file and that this test follows an established pattern, so one option is to push back and fix this in a separate pass rather than increasing the scope of this PR with broader test cleanup.
</blockquote></details>

Already done — this crossed with your review. r3 predates the commit that converted the whole file to a mockCommands helper switching on the command name, which is the shape you sketched, so no need to defer it.

I've also taken the stricter half of your suggestion since you raised it: the helper now names platform.restart explicitly and rejects any unlisted command instead of answering undefined, so a component that grows a new dependency fails loudly rather than having a silent non-answer misread as a real one. All 29 tests still pass, which incidentally confirms the helper covers every command IdentifyStep actually sends.
___
*[`src/renderer/components/first-run/steps/identify-step.component.test.tsx` line 260 at r3](https://reviewable.io/reviews/paranext/paranext-core/2716#-P-zX72n6rqLWZgzy3lR:-P09fsNg9UXuf5udTFFW:bi77k7m) ([raw file](https://github.com/paranext/paranext-core/blob/3cdad596cb4d70763c0fbcd25510d26dc7628aa2/src/renderer/components/first-run/steps/identify-step.component.test.tsx#L260)):*

Comment could be a bit clearer and more precise:
</blockquote></details>

Taken verbatim. Thanks!
___
*[`src/renderer/components/user-profile-popover/user-profile-popover.component.tsx` line 102 at r3](https://reviewable.io/reviews/paranext/paranext-core/2716#-P-zYDWc76ntarU6hqwd:-P09fwbdDjJyjE8-sYek:bcyr2bz) ([raw file](https://github.com/paranext/paranext-core/blob/3cdad596cb4d70763c0fbcd25510d26dc7628aa2/src/renderer/components/user-profile-popover/user-profile-popover.component.tsx#L102)):*

Probably should add a Jira issue for the proposed change and link it as related to PT-4325, and maybe include a mention of it here in this comment as a TODO or ENHANCE.
</blockquote></details>

Good call! Filed as PT-4493, "Promote a themed, exported reminder indicator and fix the Badge indicator variants" — a Dev Task, linked relates-to PT-4325. Not a sub-task of PT-4323, since it's component-library work that outlives the registration epic.

It captures both blockers behind the third dot treatment: Badge's blueIndicator/mutedIndicator hardcoding bg-blue-400/bg-zinc-400 past the theme, and AvatarBadge being sized off Avatar's group data-attributes and not exported. The checklist covers adding the shared component, fixing the Badge variants under the shadcn CUSTOM convention, migrating the three existing treatments, and deleting this local copy along with its TODO.

The comment here now ends with TODO(PT-4493), so the local ReminderDot has a searchable link to the follow-up.
___
*[`src/renderer/components/user-profile-popover/user-profile-popover.stories.tsx` line 36 at r3](https://reviewable.io/reviews/paranext/paranext-core/2716#-P-zOD4o7vwqsuDLAQts:-P09g0LrD1SqRDwtYCrA:bt4s73q) ([raw file](https://github.com/paranext/paranext-core/blob/3cdad596cb4d70763c0fbcd25510d26dc7628aa2/src/renderer/components/user-profile-popover/user-profile-popover.stories.tsx#L36)):*

This is just a mock for a story, but I'm a little worried it could go stale in a way that makes it hard to tell what's really happening. We neither document which other command we're hoping to receive and respond to, nor do we guarantee in this code that we got it, rather than some third, unexpected command. I'd suggest something like this instead:
</blockquote></details>

Agreed, and the worry was better founded than it looked. The command you named in your suggestion, getUserProfile, isn't the real one — it's paratextRegistration.getParatextRegistrationData. The old ternary hid that behind an else, which is exactly the staleness you were pointing at.

It's now a switch naming each command it answers and throwing on anything unexpected. One deviation: I also answer the popover's two menu-action commands, showParatextRegistration and showInternetSettings, rather than letting them throw. They fire when someone clicks a row in the open popover, so throwing would make manually poking at a story blow up rather than sit inert.

tombogle (commented):
> @tombogle reviewed 16 files, made 4 comments, and resolved 6 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2716)* status: all files reviewed, 3 unresolved discussions (waiting on katherinejensen00).

___
*[`src/renderer/components/user-profile-popover/user-profile-popover.component.tsx` line 105 at r8](https://reviewable.io/reviews/paranext/paranext-core/2716#-P0YEFc9AjO-159UKVd_:-P0YEFc9AjO-159UKVda:b-ot1mee) ([raw file](https://github.com/paranext/paranext-core/blob/531f10597b56a31e6436ab7720c6e40dd14cfcf9/src/renderer/components/user-profile-popover/user-profile-popover.component.tsx#L105)):*
> ```XML
>  * then, prefer extending this rather than copying the class string.
>  *
>  * TODO(PT-4493): promote a themed, exported indicator into `platform-bible-react` and fix the
> ```

Your Claude-authored content in PT-4493 got clobbered by the template. Ask him to resurrect it and repost the intended details.
___
*[`src/renderer/components/user-profile-popover/user-profile-popover.stories.tsx` line 45 at r8](https://reviewable.io/reviews/paranext/paranext-core/2716#-P0YFP2I5zf6UAevCqj9:-P0YFP2I5zf6UAevCqjA:buogsoy) ([raw file](https://github.com/paranext/paranext-core/blob/531f10597b56a31e6436ab7720c6e40dd14cfcf9/src/renderer/components/user-profile-popover/user-profile-popover.stories.tsx#L45)):*
> ```XML
>         case 'paratextRegistration.showParatextRegistration':
>         case 'paratextRegistration.showInternetSettings':
>           return undefined;
> ```

NIT: Even better might be to log something so that there's feedback in the console.
___
*[`src/renderer/services/registration-validity-store.ts` line 3 at r8](https://reviewable.io/reviews/paranext/paranext-core/2716#-P0YNdMFAaq0k-UCfyLT:-P0YNdMFAaq0k-UCfyLU:b-7a8kzi) ([raw file](https://github.com/paranext/paranext-core/blob/531f10597b56a31e6436ab7720c6e40dd14cfcf9/src/renderer/services/registration-validity-store.ts#L3)):*
> ```XML
> /**
>  * The session's answer to "is this user's Paratext registration valid?", shared by every UI
>  * affordance that reacts to it (currently the user-profile popover's reminder dot).
> ```

In comments, prefer "e.g." to "currently". That way, the comment stays correct as things are added and only needs to be rewritten when something is removed.
___
*[`src/renderer/services/registration-validity-store.ts` line 57 at r8](https://reviewable.io/reviews/paranext/paranext-core/2716#-P0YT8xnC8wICfuLt43o:-P0YT8xnC8wICfuLt43p:bcicj1n) ([raw file](https://github.com/paranext/paranext-core/blob/531f10597b56a31e6436ab7720c6e40dd14cfcf9/src/renderer/services/registration-validity-store.ts#L57)):*
> ```XML
>       listener();
>     } catch (e) {
>       logger.warn(`A registration-validity listener threw: ${getErrorMessage(e)}`);
> ```

NTH (low priority): Add an optional label to `subscribeToRegistrationValidity` so in the warning we could report which subscriber threw.

tombogle (approved):
> @tombogle reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2716)* status: all files reviewed, 3 unresolved discussions (waiting on katherinejensen00).

katherinejensen00 (commented):
> @katherinejensen00 made 4 comments and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2716)* status: 7 of 22 files reviewed, 2 unresolved discussions (waiting on tombogle).

___
*[`src/renderer/components/user-profile-popover/user-profile-popover.component.tsx` line 105 at r8](https://reviewable.io/reviews/paranext/paranext-core/2716#-P0YEFc9AjO-159UKVd_:-P0Z77T9AEPSIyvD-tmG:bvlayrl) ([raw file](https://github.com/paranext/paranext-core/blob/531f10597b56a31e6436ab7720c6e40dd14cfcf9/src/renderer/components/user-profile-popover/user-profile-popover.component.tsx#L105)):*

Your Claude-authored content in PT-4493 got clobbered by the template. Ask him to resurrect it and repost the intended details.
</blockquote></details>

Done. Oops. Thanks for catching that!
___
*[`src/renderer/components/user-profile-popover/user-profile-popover.stories.tsx` line 45 at r8](https://reviewable.io/reviews/paranext/paranext-core/2716#-P0YFP2I5zf6UAevCqj9:-P0ZBN-HDWxNNXUMGqig:b-h0oatk) ([raw file](https://github.com/paranext/paranext-core/blob/531f10597b56a31e6436ab7720c6e40dd14cfcf9/src/renderer/components/user-profile-popover/user-profile-popover.stories.tsx#L45)):*

NIT: Even better might be to log something so that there's feedback in the console.
</blockquote></details>

 Done — and it turned out to be more than cosmetic, so thanks for raising it.

he throw reads like feedback but produces none: resolveRegistrationValidity turns a refused probe into 'unknown', and 'unknown' renders as no dot. So a command the mock doesn't expect doesn't fail the story — it renders a plausible, healthy-looking state. The only thing that might have surfaced it is useRegistrationValidity's logger.warn, and that's electron-log, whose behavior in a plain Storybook browser bundle I haven't verified — so a direct console call is the dependable answer rather than leaning on the app logger.

 Added it at both silent sites:
 - default: in withRegistration — console.error naming the unexpected command, since silence there masks a story bug.
 - ValidityUnknown — console.warn, because that story is deliberately pixel-identical to the registered state, so the console is the only place the reject path is observable. Without it you can't tell a refused probe from a mock that was never asked. Also threaded commandName into the responder so the log names it.

 no-console is off for *.stories.ts(x) (.eslintrc.js:161-165) and stories already use console.log for action logging, so no disable comment needed.
___
*[`src/renderer/services/registration-validity-store.ts` line 3 at r8](https://reviewable.io/reviews/paranext/paranext-core/2716#-P0YNdMFAaq0k-UCfyLT:-P0ZDGAnD1URFILcsFfI:b-qn5pem) ([raw file](https://github.com/paranext/paranext-core/blob/531f10597b56a31e6436ab7720c6e40dd14cfcf9/src/renderer/services/registration-validity-store.ts#L3)):*

In comments, prefer "e.g." to "currently". That way, the comment stays correct as things are added and only needs to be rewritten when something is removed.
</blockquote></details>

Good rule — adopted. Fixed the store's header, "(currently the user-profile popover's reminder dot)" → "(e.g. …)", and also ReminderDot's "every caller so far places the dot in a flex row", which rots the same way for the same reason.

I left currently in the inFlight / inFlightId docs on purpose: those describe live runtime state ("the probe currently running"), not an enumeration, so they don't go stale as consumers are added.

Comment 4 — optional subscriber label
notifyListeners catches, the listener is an anonymous closure, so the warning could only say something threw.
the same reason.

 I left currently in the inFlight / inFlightId docs on purpose: those describe live runtime state ("the probe currently running"), not an enumeration, so they don't go stale as consumers are added.
___
*[`src/renderer/services/registration-validity-store.ts` line 57 at r8](https://reviewable.io/reviews/paranext/paranext-core/2716#-P0YT8xnC8wICfuLt43o:-P0ZD8DJEext90RteOr5:b-pshh41) ([raw file](https://github.com/paranext/paranext-core/blob/531f10597b56a31e6436ab7720c6e40dd14cfcf9/src/renderer/services/registration-validity-store.ts#L57)):*

NTH (low priority): Add an optional label to `subscribeToRegistrationValidity` so in the warning we could report which subscriber threw.
</blockquote></details>

Done. listeners is a Map<listener, label> now — still keyed by the listener, so repeat-subscribe and unsubscribe behave as before — and the warning appends  (label) only when one was passed, so an unlabeled subscriber doesn't log (undefined).

One wrinkle worth flagging: the hook passes its label through a module-level subscribne closure.useSyncExternalStore re-subscribes whenever the subscribe function's identity changes, and the store's reset comment relies on that identity being stable, so a peetached and re-attachedthe listener every render. Called that out in the TSDoc so the next caller
 doesn't step in it.

Covered by two tests: one asserting the label reaches the warning, one pinning the unlabeled message so (undefined) can't creep back.

tombogle:
*[`src/renderer/hooks/use-registration-validity.hook.ts` line 28 at r10](https://reviewable.io/reviews/paranext/paranext-core/2716#-P0bn1niB_KnN2h7lQeZ:-P0bn1niB_KnN2h7lQe_:biq8o69) ([raw file](https://github.com/paranext/paranext-core/blob/68f581cb85c0d190e091cfcef7dea49cf7c08dff/src/renderer/hooks/use-registration-validity.hook.ts#L28)):*
This is an improvement, but I was actually thinking this hook itself might want to take an optional label so if a problem surfaces, we'd know which specific use of the hook was the source of the problem.

Right now the hook hard-codes the label `'useRegistrationValidity'` when it calls `subscribeToRegistrationValidity`. But this hook is called from multiple places — the toolbar trigger and the popover row are both mentioned as consumers of the shared registration-validity store. If a listener throws, every one of those call sites will log the exact same string, `'useRegistrationValidity'`. That tells you _which hook_ was involved, but not _which component instance_ — which is presumably the actually useful piece of information when debugging a listener that's misbehaving. So pushing the label up to be a parameter of the hook (e.g. `useRegistrationValidity('toolbar')` vs `useRegistrationValidity('popoverRow')`) would let each call site self-identify, which is strictly more diagnostic value than a constant string repeated everywhere.

That said, there's a real wrinkle worth considering before you blindly go this route. Per the commit that introduced the label:

> The hook supplies its label through a module-level subscribe wrapper rather than an inline closure, because `useSyncExternalStore` re-subscribes whenever that function's identity changes and the store's reset comment depends on it being stable.

In other words, `useSyncExternalStore` requires the `subscribe` function passed to it to be stable across renders, or it'll thrash (re-subscribe every render). That's why the current implementation uses a module-level wrapper rather than building a fresh closure inline. If the label becomes a hook parameter, the hook can no longer use one single module-level wrapper for everyone — it needs a `subscribe` function whose identity is stable _per label_, not stable globally. That's solvable (e.g. `useCallback` keyed on the label, since call-site labels are normally static strings), but it's not a drop-in change — it's a small design decision you would need to make deliberately, not just "add a param and pass it through."

So I can see why you did it this way and maybe it's enough for something that is only in support of an error-reporting scenario that we hope we never hit. Alternatively, you could just make this a minimal-priority enhancement issue and/or add and ENHANCE comment to the code, so if this ever becomes a real problem we'll at least know that we thought about how to disambiguate.

tombogle (commented):
> @tombogle reviewed 14 files and all commit messages, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2716)* status: 21 of 22 files reviewed, 1 unresolved discussion (waiting on katherinejensen00).

tombogle (approved):
> @tombogle reviewed 1 file.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2716)* status: all files reviewed, 1 unresolved discussion (waiting on katherinejensen00).

katherinejensen00 (commented):
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2716)* status: all files reviewed, 1 unresolved discussion (waiting on tombogle).

___
*[`src/renderer/hooks/use-registration-validity.hook.ts` line 28 at r10](https://reviewable.io/reviews/paranext/paranext-core/2716#-P0bn1niB_KnN2h7lQeZ:-P0cnaGF76NicvdseUWA:b-rpy47i) ([raw file](https://github.com/paranext/paranext-core/blob/68f581cb85c0d190e091cfcef7dea49cf7c08dff/src/renderer/hooks/use-registration-validity.hook.ts#L28)):*

This is an improvement, but I was actually thinking this hook itself might want to take an optional label so if a problem surfaces, we'd know which specific use of the hook was the source of the problem.

Right now the hook hard-codes the label `'useRegistrationValidity'` when it calls `subscribeToRegistrationValidity`. But this hook is called from multiple places — the toolbar trigger and the popover row are both mentioned as consumers of the shared registration-validity store. If a listener throws, every one of those call sites will log the exact same string, `'useRegistrationValidity'`. That tells you _which hook_ was involved, but not _which component instance_ — which is presumably the actually useful piece of information when debugging a listener that's misbehaving. So pushing the label up to be a parameter of the hook (e.g. `useRegistrationValidity('toolbar')` vs `useRegistrationValidity('popoverRow')`) would let each call site self-identify, which is strictly more diagnostic value than a constant string repeated everywhere.

That said, there's a real wrinkle worth considering before you blindly go this route. Per the commit that introduced the label:

> The hook supplies its label through a module-level subscribe wrapper rather than an inline closure, because `useSyncExternalStore` re-subscribes whenever that function's identity changes and the store's reset comment depends on it being stable.

In other words, `useSyncExternalStore` requires the `subscribe` function passed to it to be stable across renders, or it'll thrash (re-subscribe every render). That's why the current implementation uses a module-level wrapper rather than building a fresh closure inline. If the label becomes a hook parameter, the hook can no longer use one single module-level wrapper for everyone — it needs a `subscribe` function whose identity is stable _per label_, not stable globally. That's solvable (e.g. `useCallback` keyed on the label, since call-site labels are normally static strings), but it's not a drop-in change — it's a small design decision you would need to make deliberately, not just "add a param and pass it through."

So I can see why you did it this way and maybe it's enough for something that is only in support of an error-reporting scenario that we hope we never hit. Alternatively, you could just make this a minimal-priority enhancement issue and/or add and ENHANCE comment to the code, so if this ever becomes a real problem we'll at least know that we thought about how to disambiguate.
</blockquote></details>

Taking your second option — documented, not parameterized — because the premise for the first one doesn't hold yet.

There's exactly one call site. grep for useRegistrationValidity() across src/ outside tests returns a single hit: user-profile-popover.component.tsx:145. The toolbar trigger and the "Profile & registration" row are both rendered by UserProfilePopover off that one hook call — the dot on the trigger and the dot on the row read the same validity from the same subscription. So there's one subscriber to name today, and 'useRegistrationValidity' names it uniquely. A per-call-site label is strictly more diagnostic when there are multiple call sites; with one, it adds a parameter to the hook's signature and distinguishes nothing.

Your wrinkle is the reason not to do it speculatively, and it's slightly worse than you framed it. Once the label is a parameter, one module-level wrapper can't serve everyone, because useSyncExternalStore needs identity stable per label — so useCallback keyed on the label, which is fine while call-site labels are static strings. But it also undercuts how resetRegistrationValidityStore justifies itself: that function deliberately keeps its subscribers across a reset, and its comment argues that's safe because subscribe is a stable module-level function, so React subscribes once per mount and never re-subscribes. A per-label useCallback is still stable per mount, so the behavior holds — but it stops being true by construction and starts being true by argument, and the next person to touch either file has to re-derive it. That's a real cost to pay for a hypothetical second consumer.

So I documented it at the label site rather than filing an issue. The note now says one subscription is all there is to name, and spells out the exact change if that stops being true: optional label on the hook, subscribe built with useCallback keyed on it, because identity then has to be stable per label and a module-level wrapper can't provide that. Whoever adds the second consumer finds the reasoning at the point they'd otherwise break it, which seemed better than a backlog item nobody reads.

Two deliberate choices in how I wrote it, both easy to reverse if you'd rather: I didn't use an ENHANCE tag or invent a ticket ID, since our forward-facing-comments rule wants a real open PT-XXXX behind a TODO and this isn't work anyone has committed to; and I named UserProfilePopover explicitly even though that fact can rot, because the comment's own next sentence handles the case where it does. Happy to file the enhancement issue as well if you'd prefer it tracked rather than just explained — say the word.

---

- The one claim I'd flag as argued rather than verified is the resetRegistrationValidityStore point — it's my reading of that function's comment, not something a test pins. Everything else is checked: the single call site by grep, and the code change by Prettier/ESLint plus 155/155 passing tests.
- I've drafted only; nothing is posted. Tell me when to post and I'll add it as a PR comment (the channel used for the 8/28 reply). Resolving the Reviewable discussion still needs your click there — the API can't do it.

The editor link is still building in the background, so the full typecheck and the commit of the comment change are still outstanding.

✻ Worked for 46s · done 12:42 PM · 1 shell, 1 monitor still runningchange if that stops being true: optional label on the hook, subscribe built with useCallback keyed on it, because identity then has to be stable per label and a module-level wrapper can't provide that. Whoever adds the second consumer finds the reasoning at the point they'd otherwise break it, which seemed better than a backlog item nobody reads.

Two deliberate choices in how I wrote it, both easy to reverse if you'd rather: I didn't use an ENHANCE tag or invent a ticket ID, since our forward-facing-comments rule wants a real open PT-XXXX behind a TODO and this isn't work anyone has committed to; and I named UserProfilePopover explicitly even though that fact can rot, because the comment's own next sentence handles the case where it does. Happy to file the enhancement issue as well if you'd prefer it tracked rather than just explained — say the word.

PR #2717, opened 2026-08-25
PT-4403: Cover empty-chapter localize keys, announce the zero state

jolierabideau:
# Code Review Summary

**Branch**: pt-4403-blank-chapter-cleanup

**Base**: origin/main

**Date**: 2026-09-04

**Files changed**: 7 (4 at review start, 3 added by in-review fixes)

## Overview

PR #2717 had gone stale. The original PT-4403 work — unifying blank-chapter semantics across the
editor and panels — shipped on main independently in #2704 as `isBlankChapterOnScreen`, which
additionally excludes chapter 0. The branch was therefore rebuilt on main and reduced to the one
piece main still lacked: localize-key parity coverage for the editor's empty-chapter zero state.

The four empty-chapter keys sat outside all parity coverage, which hid three English-only strings.
This branch extracts those keys into `empty-chapter-view.const.ts` so `localized-strings.test.ts`
can import them without routing them through the component, and adds the missing Spanish values. The
four keys now have their own parity block: dropping any one of their Spanish values fails with the
key named.

The split is for collect cost, not necessity — the component imports fine from a `node` test
(verified: 182/182), and routing the keys through it roughly quadruples this file's collect time
(235ms → 1.05s). The three `.const.ts` headers that claimed a load-time impossibility were
corrected to say so.

**Rebase**: applied cleanly onto `origin/main` (`57b792f95ff`, PT-4464 window ids) with no
conflicts. The branch is now exactly even with main plus its own work.

**Note for the reviewer**: an earlier revision of this branch also added a whole-file en/es key-set
guard plus Spanish values for three `%versionHistoryCommit_*%` keys. Both were **dropped** in
`dc8be09` after review. Those keys are not UI chrome — they flow through
`commitVersionHistorySnapshot` into `paratextBibleSendReceive.commitChanges`, so they become commit
messages persisted in version history, shared with collaborators, and frozen at commit time, and
`platform-scripture` ships their three siblings English-only across all 33 of its language blocks.
Localizing shared project data is a product decision, not a side effect of a parity test. The guard
was the only thing requiring them, and it had its own defect (a module-scope read of the English key
set aborted collection instead of failing), so both went. A repo-scoped, ratcheting replacement is
filed separately.

## API Changes

None. No public API surface changed — all changed files are inside
`extensions/src/platform-scripture-editor/`, and none is a type declaration file (`src/types/*.d.ts`),
`lib/platform-bible-react/`, `lib/platform-bible-utils/`, or `lib/papi-dts/papi.d.ts`.

For completeness, the extension-internal module surface moved (verified non-breaking):

- `empty-chapter-view.const.ts` (new): exports `EMPTY_CHAPTER_VIEW_KEYS`,
  `EMPTY_CHAPTER_VIEW_STRING_KEYS`, `EmptyChapterViewStringKey`, `EmptyChapterViewLocalizedStrings`
- `empty-chapter-view.component.tsx`: those definitions moved out, then re-exported from the same
  module, so all three names remain exported and the sole consumer
  (`platform-scripture-editor.web-view.tsx:122`) is unaffected. `EmptyChapterViewProps` gains one
  new optional prop, `announcementKey`.
- `contributions/localizedStrings.json`: 3 Spanish values added (the empty-chapter keys;
  `messageResource` already shipped on main). No key added or removed on the English side, and no
  existing en or es value changed — pure additions, so the immutable-strings rule is not engaged.

None.

None.

### Minor — Consider

- [x] **`localized-strings.test.ts` — the first assertion in `it('declares every English key in Spanish')` checked the opposite direction from its name** (`esKeys.filter((key) => !localizedStrings.en[key])` reads keys from `es` and looks them up in `en`), so a failure would have named the wrong language. It was also largely subsumed by the following `expect(esKeys).toEqual(enKeys)`. Flagged independently by all four analysis passes. _(fixed during review: split into two purpose-named guards — a per-language "has a non-empty value for every key it declares" and a per-language "declares the same key set as English". Verified falsifiable both ways: dropping an `es` key fails the parity guard, blanking an `es` value fails the value guard, and both now name the language in the failure output.)_
- [x] **`empty-chapter-view.const.ts` — the four key names were enumerated twice**, once in `EMPTY_CHAPTER_VIEW_KEYS` and once in `EMPTY_CHAPTER_VIEW_STRING_KEYS`, with nothing enforcing agreement; a fifth key added to only the object would have silently escaped the parity test. _(fixed during review: the array is now derived — `Object.freeze(Object.values(EMPTY_CHAPTER_VIEW_KEYS))`. The `EmptyChapterViewStringKey` union is unchanged.)_
- [x] **`empty-chapter-view.const.ts:2` — `{@link EmptyChapterView}` could not resolve**, since the symbol is neither imported nor declared in that module. _(fixed during review: changed to backticks, matching line 25 of the same file.)_
- [x] **`localized-strings.test.ts` — the guard hardcoded `en` and `es`**, so a third language added later would have received zero parity coverage silently while the block comment still claimed to cover the whole file. _(fixed during review: both guards now iterate `Object.keys(localizedStrings)`, comparing every non-English language against English.)_
- [x] **`localized-strings.test.ts` — the guard makes en/es parity a hard gate for this extension only**; repo-wide it is not an invariant (`platform-scripture` has 266 en-only keys, `platform-enhanced-resources` 170), so a contributor adding an English-only key here hits a failure no sibling extension produces. _(fixed during review: added a comment stating the stricter-than-repo policy is deliberate and that the fix is to translate the key rather than relax the guard.)_
- [x] **The three `%versionHistoryCommit_*%` Spanish values sit outside the stated empty-chapter scope.** _(resolved in `dc8be09`: the values and the whole-file guard that required them were both removed. See the reviewer note in the Overview.)_
- [ ] ~~**`localizedStrings.json:282` — the Spanish disabled tooltip is ~95 chars / two sentences, against the Tooltips guideline ("a hint, not an explanation").**~~ _(Author asked for this to be fixed, then dismissed it once measured. At 63 chars the English source is mid-range among its five `DisabledActionTooltip` siblings — 40, 49, 56, 63, 69 — so it is not an outlier, and the guideline's own "avoid" example is 68 chars. More decisively, no shorter phrasing keeps both facts, so any real trim drops "chapter and verse markers cannot be added" — a meaning change, which `Localization-Guide.md` marks CRITICAL: it would require a new key plus a `fallbackKey` mapping rather than an in-place edit, and new Spanish authored without translator sign-off. Deliberately not changed. If the family is ever trimmed it should be done together, in its own ticket.)_
- [x] **`EmptyChapterView` had neither a story nor a component test**, unlike the sibling zero states `book-not-available-view` and `panel-readiness-view`. Pre-existing on main, not introduced by this branch. _(fixed during review: added `empty-chapter-view.component.stories.tsx` — 5 stories covering editable, structure-protected, no-button, resource, and missing-strings — and `empty-chapter-view.component.test.tsx`, 12 tests covering rendering, the resource message variant, the withheld button, the click handler, structure protection including the focusable disabled wrapper, string fallback, and accessibility.)_
- [x] **`EmptyChapterView` rendered a bare `div`/`span` with no `role="status"`**, so its arrival was never announced. Pre-existing on main. The gap is real and confirmed: the web view's own comment at `platform-scripture-editor.web-view.tsx:3719` states the editing subtree it replaces is hidden with `display: none`, "which takes it out of both the accessibility tree and the tab order" — so the swap was silent and whatever held focus inside the editor stopped being focusable. _(fixed during review: the region is now `role="status"` with `tabIndex={-1}`, takes focus on mount via the existing `useFocusReplacedContent` hook, and accepts an `announcementKey` — wired from the web view as `${projectId}:${scrRef.book} ${scrRef.chapterNum}` — so moving between two blank chapters remounts the region instead of leaving byte-identical text that `aria-live` cannot see as a change. This follows the sibling `BookNotAvailableView` pattern exactly. Five tests pin it, all verified falsifiable.)_

#### Shared Regions Modified

#### Extension Config Changes

- [n/a] `contributions/localizedStrings.json` — extension-specific localization content, not shared build/config infrastructure.
- [n/a] All other changed files are extension source under `extensions/src/platform-scripture-editor/src/`. No config or build files were touched.

## Positive Observations

- The extraction is non-breaking by construction: the component re-exports all three previously
  exported names, so the web view needed no edit. Verified against `git show` of the merge base.
- The new const module is a faithful copy of the existing `book-not-available-view.const.ts` shape,
  down to TSDoc explaining *why* the split exists. A future reader hits the rationale before the
  workaround.
- The whole-file parity guard closes a real class of gap rather than duplicating the per-component
  blocks: every other block in the file can only see keys someone remembered to export, and the
  three `%versionHistoryCommit_*%` keys are proof that unexported keys existed and were uncovered.
  The per-component blocks stay complementary — they still catch a key that exists in code but is
  absent from the JSON entirely.
- New keys were inserted in alphabetical position within both language blocks, matching the file's
  existing ordering.
- The added Spanish reuses vocabulary already established in the same file rather than inventing
  synonyms (`Insertar referencia cruzada` / `nota al pie` match the existing `insert*AtSelection`
  menu labels; `añadir` matches the `bookNotAvailable_*` strings), and `quitar` rather than
  `eliminar` correctly follows the guide's remove-vs-delete distinction.
- Module docs state forcing constraints rather than narrating the PR, so they survive the merge.

## Interview Notes

The author's framing was that the PR had gone stale and they needed to know whether the changes
were still valid, get it onto latest main, and have the changes reviewed.

**On staleness and validity** — the branch was in better shape than the author expected. It had
already been rebuilt onto main earlier the same day and reduced to a single commit; it was only one
commit behind. Three independent checks confirm the surviving change is still valid:
`EMPTY_CHAPTER_VIEW_STRING_KEYS` is still consumed from the component by the web view, so the
re-export shim is load-bearing rather than vestigial; `isBlankChapterOnScreen` from #2704 is present
on main and untouched, confirming the superseded work was correctly dropped; and the four
empty-chapter keys are each present and non-empty in both en and es, which is the parity claim the
branch actually makes now that the whole-file guard is gone.

**Author decisions during review:**

- Asked for all findings to be fixed. Findings 1-6 were implemented directly.
- For the three pre-existing findings the analysis had flagged as out of scope, the author chose to
  fix all three rather than keep the PR narrow — a deliberate widening of scope from the
  localize-key change the branch had been reduced to. Findings 8 (story + test) and 9
  (accessibility) were implemented.
- Finding 7 (tooltip length) the author initially asked to fix, then dismissed once the measurement
  and the immutable-strings constraint were presented. See the dismissed entry above.

**Design decisions made during the review that the reviewer should confirm:**

- `EmptyChapterView` was given `role="status"` and focus repair but **not** converted to the shared
  `Empty` composition, which centers its content. The view keeps its top-left one-row layout because
  a blank chapter is transient — the scaffold button flips it straight back to the editor — so the
  message sits where the first line of text is about to appear. This is recorded in the component's
  TSDoc. `BookNotAvailableView`'s TSDoc documents the converse divergence, so the two are now
  explicitly reconciled.
- The hidden-view case for the new focus repair: when the web view's own tab is inactive,
  `document.hasFocus()` is false and the repair declines by design, so no focus is stolen into a
  pane nobody is looking at. The remounted `role="status"` region remains the announcement. This is
  the behavior the shared hook already documents; no new deferral mechanism was added.
- No `Architecture-Decisions.md` entry was appended. The one candidate is the "full en/es parity is
  the standing policy for this extension, stricter than the repo at large" decision, which is
  currently documented only at the enforcement site. Raising rather than appending, since the log is
  append-only and the policy has not been agreed by the team — see Suggested Review Focus.

No areas were deferred to AI or left unexplained by the author.

## In-Review Quality Check

- **Targeted suites**: `localized-strings.test.ts` 182 passing; `empty-chapter-view.component.test.tsx`
  12 passing.
- **Full extension suite**: `npx vitest run src/platform-scripture-editor` — **86 files, 1391 tests,
  all passing**. No regressions from the component change.
- **Falsifiability verified** (not just green): removing `role="status"` fails 5 tests; removing the
  remount `key` fails 2; dropping an `es` key fails the parity guard; blanking an `es` value fails
  the value guard. The contribution JSON was restored byte-identical after each probe.
- **ESLint**: clean on all six touched files.
- **Prettier**: applied; one `prettier/prettier` warning on the derived-array line was fixed.
- **Typecheck**: `npx tsc --noEmit -p extensions/tsconfig.json` reports **333 errors on
  `origin/main` and 333 on this branch** — zero introduced. (The earlier 244 baseline predates the
  merge of latest main.) None are in any empty-chapter or localized-strings file.
  The one hit inside `platform-scripture-editor.web-view.tsx` (line 2854,
  `'paratextBibleSendReceive.commitDaily'` not assignable to `keyof CommandHandlers`) is
  pre-existing and unrelated to this diff.
- **Not run**: repo-root `npm run typecheck` and `npm test` across all workspaces, and
  `dotnet test c-sharp-tests/`. No C# was touched. The pre-existing 244-error extension typecheck
  baseline means a root typecheck would not be a clean signal either way.

## Suggested Review Focus

- [ ] **Scope**: this PR was deliberately reduced to the localize-key piece, then widened during
      review to also fix three pre-existing `EmptyChapterView` gaps at the author's request. Confirm
      that widening is what you want, or ask for findings 8 and 9 to be split into their own PR.
- [ ] **The accessibility change is the most substantive part of the diff** and did not exist when
      the PR was opened. `role="status"` + focus repair + `announcementKey` on `EmptyChapterView`,
      with `announcementKey` wired from the web view. It mirrors `BookNotAvailableView`, but it is
      new production behavior in a view that previously had none — worth real scrutiny.
- [x] **Stricter-than-repo localization gate**: resolved by dropping the whole-file guard. Parity
      enforcement is back to per-component key lists only. The repo-scoped version — one guard
      globbing `extensions/src/*/contributions/localizedStrings.json` with a shrink-only allowlist
      that ratchets the ~444-key backlog — is filed as follow-up rather than landing per-extension
      here.
- [ ] **The dismissed tooltip finding** — confirm you agree the string should be left alone, and
      whether trimming the whole `DisabledActionTooltip` family deserves its own ticket.
- [ ] **Spanish translations**: six new values were added. They were checked for vocabulary
      consistency against the existing file, but a Spanish speaker should confirm.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2717)


katherinejensen00 (changes requested):
# Review of PR #2717 — _Cover empty-chapter localize keys in en/es parity tests_

- **PR**: https://github.com/paranext/paranext-core/pull/2717
- **Author**: jolierabideau
- **Branch**: `pt-4403-blank-chapter-cleanup` → `main`
- **Head reviewed**: `65c981d86ab`
- **Size**: 4 files, +97 / −21
- **Reviewed**: 2026-09-04

**To the PR author:** The extraction itself is clean and the parity gap you found is real —
`Localization-Guide.md:492` lists en/es key gaps as a _blocking_ issue, and this is the repo's first
mechanical enforcement of that rule, which is a genuinely good contribution. But the new whole-file
guard has one **blocking defect**: it makes the test file fail _silently and totally_ in the exact
scenario it was written to catch. Separately, the new `.const.ts` module's stated reason for
existing does not reproduce, and three of the six added strings are user-visible data changes riding
under a test-shaped title.

Findings confirmed by running code are marked **[verified]**.

Paths are relative to the repo root; the prefix `extensions/src/platform-scripture-editor/` is
abbreviated as `…/` after its first use.

---

## Blocking

### 1. `extensions/src/platform-scripture-editor/src/localized-strings.test.ts:296` — Severity: High (blocking)

**Summary:** A missing or renamed `es` block now wipes out all 181 tests in this file instead of
failing them, so the guard destroys the file's coverage at the exact moment it is supposed to fire.

The two `Object.keys()` calls sit in the `describe` callback, which runs during _collection_, not
inside an `it`. A throw there aborts the whole file. **[verified]** — renaming the JSON's `"es"`
block to `"es-419"`:

| | base (`main`) | PR head |
| --- | --- | --- |
| result | **110 failed / 57 passed (167)** — file collects, every test names its key | **`Tests: no tests`** — 0 of 181 collected, bare `TypeError: Cannot convert undefined or null to object` at `:297:25` |

The hazard is new to this PR: at base, every `localizedStrings.en/.es` read is inside an `it` body.
Trigger is a one-word edit in a 38 KB JSON (locale retirement, regional-variant rename, bad merge).

**Fix:** move both `Object.keys` reads inside the `it` bodies.

---

## Important

### 2. `…/src/empty-chapter-view.const.ts:5` — Severity: Medium-High

**Summary:** The new module's sole stated justification — that a node-environment test _cannot_
import the component — is false, and it has now been copy-pasted into a third file.

The header says importing the component "would drag in `platform-bible-react`, which touches
`document` at module load." **[verified]** — pointing the test's import at
`./empty-chapter-view.component`: **181/181 pass**. `extensions/vitest.config.ts` is a single
`environment: 'node'` with no jsdom override and no `platform-bible-react` alias, so there is no
hidden rescue. Further, 66 test files under `extensions/src` (45 in this extension) use
`// @vitest-environment jsdom`, and `book-not-available-view.component.test.tsx:1,6` uses that
pragma _and_ imports its component directly.

There **is** a real cost (collect time 0.52s → 2.46s), which is a fine reason to keep the split —
but it is not the reason written down. A comment asserting a false invariant is worse than none,
because the next person copies it into a fourth `.const.ts`.

### 3. `…/src/localized-strings.test.ts:130` — Severity: Medium

**Summary:** This PR makes the neighbouring comment factually false, and the filter it justifies now
excludes the very key you just translated from the only assertion that would catch a regression.

The comment claims "`%versionHistoryCommit_*%` is en-only across the whole file." Your first hunk
gives all three Spanish values. **[verified]** — replacing the filter with `.filter(() => true)`
yields **184/184 green**, so it can simply be deleted. Until then
`%versionHistoryCommit_beforeRemoveCharacterMarker%` has _no_ "es differs from en" coverage
anywhere.

The comment's second clause — about the sync-blocked key — describes a key not in
`REMOVE_CHARACTER_MARKER_STRING_KEYS` at all; that half is pre-existing.

### 4. `…/contributions/localizedStrings.json:241` — Severity: Medium

**Summary:** The three `%versionHistoryCommit_*%` values are not UI chrome — they are persisted,
cross-user project data, and translating them here leaves the repo shipping two contradictory
policies for one key family.

**[verified]** — these flow through `commitVersionHistorySnapshot`
(`editor-side-effects.utils.ts:82-87`) into
`papi.commands.sendCommand('paratextBibleSendReceive.commitChanges', …)`, i.e. a commit message
written into version history and shared via Send/Receive. Meanwhile `platform-scripture` ships
**33 locales** and its sibling keys (`_beforeReplace`, `_beforeReplace_failureMessage`,
`_afterReplace`) are **English-only in all 33**. A Spanish translator's history then interleaves
_"Antes de insertar nota al pie"_ with _"Before doing Replace"_, permanently — labels freeze at
commit time.

Worth a deliberate product call, not a side effect of turning a test green.

### 5. `…/src/empty-chapter-view.component.tsx:56` — Severity: Medium

**Summary:** The PR adds 12 tests about these four keys and zero that the component picks the
_right_ one — an inverted ternary passes all of them.

No `.test.tsx` and no `.stories.tsx` exist for this component (both siblings have both). Nothing
covers `isResource ? MESSAGE_RESOURCE_KEY : MESSAGE_KEY`. Invert it and all 12 new tests still pass,
while an editable project shows _"This chapter is empty in this resource."_ beside a live
**Add chapter number** button — exactly what the two-key split exists to prevent.

Pre-existing, but the new parity tests create false confidence that the view is covered.

---

## Moderate

### 6. `…/src/localized-strings.test.ts:295` — Severity: Medium-Low

The guard checks key-set parity but never that `es` differs from `en`, so pasting English into `es`
satisfies it. **[verified]** — setting `es['%versionHistoryCommit_beforeInsertFootnote%']` to its
English value leaves **181/181 green**. That is the fastest way a future developer unblocks the new
red test. A blanket differs-check needs an allowlist (**[verified]** exactly 4 keys legitimately
share values, all placeholder-only format strings); at minimum, document the limitation beside the
guard.

### 7. `…/src/localized-strings.test.ts:290` — Severity: Medium-Low

The guard compares locale-to-locale and never code-to-JSON, so a key referenced in code but absent
from _both_ locales is invisible — and those are precisely the keys the comment names as its
motivation (`platform-scripture-editor.web-view.tsx:331-332`, consumed at `:1579`/`:1596`). Rename
one in the JSON and the literal `%versionHistoryCommit_beforeInsertFootnote%` gets written into
collaborators' version history while the suite stays green.

### 8. `…/src/localized-strings.test.ts:299` — Severity: Medium-Low

`it('declares every English key in Spanish')` opens with the assertion for the _opposite_ direction
(line 300 checks Spanish keys missing English values), and since `expect` throws, line 301 — the
direction the title describes — never runs. Split into two direction-accurate blocks.

### 9. `…/src/localized-strings.test.ts:276` — Severity: Medium-Low

`%webView_platformScriptureEditor_emptyChapter_messageResource%` is **[verified]** already covered
via `model-text-panel.const.ts:34` at `:228-240`. The real gap was three keys, not four. This file
maintains two explicit "checked elsewhere" mechanisms (`:73-84`, `:134`) and the new block ignores
both, so one broken value now fails two suites with unrelated names.

### 10. `…/src/localized-strings.test.ts:291` — Severity: Medium-Low

The rationale comment's cited example is checkably wrong:
`%versionHistoryCommit_beforeRemoveCharacterMarker%` **is** in an exported array
(`use-remove-character-marker.hook.ts:24-27`). It went uncovered because of the filter at `:134`,
not because no array exports it. Only the two `beforeInsert*` keys fit the described class.

### 11. `…/src/empty-chapter-view.const.ts:28` — Severity: Medium-Low

New TSDoc says the test "iterates every exported `*_STRING_KEYS` array." It iterates nine
hand-written imports; the extension exports fourteen. Your own comment at `test:290-292` says the
opposite. Copied verbatim from `book-not-available-view.const.ts:33-34` — fix both.

### 12. `…/src/empty-chapter-view.component.tsx:54` — Severity: Medium-Low (accessibility, pre-existing)

Renders a bare `<div><span>` with no `role="status"`, no focus target, no focus repair — unlike both
siblings doing the same swap (`book-not-available-view.component.tsx:88-101`,
`resource-message-view.component.tsx:55-70`). The editor subtree gets `display:none`, so a
screen-reader user arriving at a blank chapter has focus fall to `<body>` with nothing announced.
The _return_ path is handled; the arrival path is not, and no comment records it as deliberate.

### 13. `…/src/localized-strings.test.ts:296` — Severity: Low-Medium

Titled "across the whole contribution file" but reads only `en`/`es`. Note: iterating all locales
would be **wrong** — `platform-scripture` deliberately ships 32 partial locales (3–11 keys each) and
the guide's fallback chain handles that by design. Fix the _title_, not the loop.

### 14. `.context/standards/Localization-Guide.md:228` — Severity: Low-Medium

States "Both are required for the build to pass." **[verified]** false: no localization check in any
npm script or workflow, and 444 en-only keys ship today. Your own test comment (`:31-32`) says the
opposite. Per CLAUDE.md's "promote settled conventions," point the guide at where the guard actually
lives.

---

## Minor / Nit

### 15. `…/src/empty-chapter-view.const.ts:17` — Severity: Low

The four keys are hand-listed twice; `Object.freeze(Object.values(EMPTY_CHAPTER_VIEW_KEYS))`
preserves the literal union. Mirrors the sibling deliberately — but the sibling has a reason this
file lacks (its object feeds `DISABLED_REASON_TOOLTIP_KEYS`), and
`resource-panel-strings.utils.ts:59` already derives. Change both files or neither.

### 16. `…/src/localized-strings.test.ts:304` — Severity: Low

A 380-key mutation sweep found no input where this `it` fails alone; it is subsumed by the first for
every realistic case. Collapse to one accurately-named test plus a separate non-empty-value
assertion.

### 17. `…/src/localized-strings.test.ts:300` — Severity: Low

`!localizedStrings.en[key]` tests truthiness where existence is meant, so an empty-string value
makes the two `it` blocks assert contradictory things about the same key. `Object.hasOwn` states the
intent.

### 18. `…/src/localized-strings.test.ts:281` — Severity: Low

The whole-file guard makes `'has a Spanish label'` unfalsifiable in all seven per-key blocks (56 of
181 cases now carry no unique detection power). Note what _does_ survive, so a later cleanup does
not delete the wrong half: `'has an English label'` is the only code→JSON link, and
`'Spanish differs from English'` the only untranslated-copy check.

### 19. `…/src/localized-strings.test.ts:274` — Severity: Low

Seventh verbatim copy of the same three-assertion block (~94 lines across `:36`–`:276`). You
declined a helper as a pre-existing pattern — fair, but this PR is what makes the shape obsolete
rather than merely repetitive, so it is worth recording as a decision.

### 20. `…/src/localized-strings.test.ts:295` — Severity: Low (scope)

**[verified]** 444 en-only keys across five extensions; this guard lands on the one file already at
190/190. A repo-scoped test globbing `extensions/src/*/contributions/localizedStrings.json` with a
shrink-only allowlist would ratchet the backlog instead of needing N copy-pastes.

### 21. `.context/standards/Architecture-Decisions.md` (absent entry) — Severity: Low

CLAUDE.md requires recording cross-cutting decisions. Two candidates here: the whole-file parity
convention (your own body calls the escape hatch undesigned) and the now-five-file `.const.ts` split
pattern.

### 22. Commit `65c981d86ab` / PR body — Severity: Low

Missing the `Session-URL:` trailer and the `AI-assisted — [session](url)` line CLAUDE.md requires.
Matters more than usual since the body states the implementation was authored during the review
session.

### 23. `…/src/empty-chapter-view.const.ts:2` — Severity: Nit

`{@link EmptyChapterView}` cannot resolve — the symbol lives in the module this file deliberately
does not import. Line 25 uses backticks for the same symbol. Identical in the sibling; fix together.

### 24. `…/src/empty-chapter-view.component.tsx:56` — Severity: Nit (low confidence)

The message renders as an unstyled `<span>`; `book-not-available-view.component.tsx:105-107` uses
`tw:text-center tw:text-sm tw:text-muted-foreground` with an explicit comment that the treatment
should match across surfaces.

### 25. `…/contributions/localizedStrings.json:282` — Severity: Nit (latent)

`"La estructura está bloqueada. Los marcadores…"` would fail this file's own sentence-case rule
(`test:60-66`) if its scope were widened, because the rule skips only word index 0 and is not
sentence-boundary aware. Fix the rule before growing its scope, not the translation.

### 26. PR title — Severity: Nit

"Cover empty-chapter localize keys in en/es parity tests" understates it: three of six added values
are new user-visible Spanish strings (see finding 4). The strings themselves are correct and
consistent with in-file precedent — `añadir` matches the exact "No se pueden ___" construction at
`:263`/`:265`.

---

## Verified clean

- JSON hunks are pure appends with zero existing values modified (immutability rule satisfied); both
  blocks stay alphabetized at 190/190.
- The component's export surface is effect-identical to base, render body byte-identical, no import
  cycle, no dropped export.
- Lint clean on all four files; suite green at 181/181 as submitted.

Two candidate findings were **refuted and dropped**: the guard does not foreclose the
`metadata.fallbackKey` pattern, and the un-namespaced `%versionHistoryCommit_*%` family has no
cross-extension collision (1162 distinct `en` keys across 10 extensions, zero defined twice).

**If you fix only one thing, make it finding 1** — it is a two-line move, and it is the difference
between a guard that reports the problem and one that hides it.

---

## Review method

Four passes, cross-checked against each other:

1. Developer / soundness pass (code quality, testing, breakage risk).
2. Code-architect pass (pattern conformance, where the change belongs).
3. Comments-and-documentation pass (accuracy, density, forward-facing test).
4. An adversarial pass that attacked passes 1–3, followed by `/code-review max` with the finding cap
   raised to 16 per subagent and 30 for the final report.

The adversarial pass killed six of eleven initial findings, including one graded positive that
inverted into finding 2. Every high-severity claim above was re-verified by executing it against the
PR head in a scratch copy; the review worktree was restored to a clean state afterwards — no
commits, no changes to the PR, no comments posted.

> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2717#-:-P0iAwGYDX8AAN8IE55l:bypbn4r)* status: all files reviewed, 1 unresolved discussion (waiting on jolierabideau).

___
*[`extensions/src/platform-scripture-editor/src/empty-chapter-view.const.ts` line 5 at r1](https://reviewable.io/reviews/paranext/paranext-core/2717#-P0iB2h516DIZPoZxC9v:-P0iB2h516DIZPoZxC9w:b-mnl3qv) ([raw file](https://github.com/paranext/paranext-core/blob/3f121b5d8a922453f2fdb29f390ecf2d4c607fb0/extensions/src/platform-scripture-editor/src/empty-chapter-view.const.ts#L5)):*
> ```XML
>  *
>  * Split out of the component module so consumers that must not pull in React can import them.
>  * `localized-strings.test.ts` is the one that forces this: it runs in the node environment (it only
> ```

### 2. `…/src/empty-chapter-view.const.ts:5` — Severity: Medium-High

**Summary:** The new module's sole stated justification — that a node-environment test _cannot_
import the component — is false, and it has now been copy-pasted into a third file.

The header says importing the component "would drag in `platform-bible-react`, which touches
`document` at module load." **[verified]** — pointing the test's import at
`./empty-chapter-view.component`: **181/181 pass**. `extensions/vitest.config.ts` is a single
`environment: 'node'` with no jsdom override and no `platform-bible-react` alias, so there is no
hidden rescue. Further, 66 test files under `extensions/src` (45 in this extension) use
`// @vitest-environment jsdom`, and `book-not-available-view.component.test.tsx:1,6` uses that
pragma _and_ imports its component directly.

There **is** a real cost (collect time 0.52s → 2.46s), which is a fine reason to keep the split —
but it is not the reason written down. A comment asserting a false invariant is worse than none,
because the next person copies it into a fourth `.const.ts`.

katherinejensen00:
PT-4403 has one decision to make before it is ready to merge according to claude:
The three %versionHistoryCommit_% Spanish values (contributions/localizedStrings.json:243-245). They aren't UI chrome that renders and disappears — they flow through commitVersionHistorySnapshot into paratextBibleSendReceive.commitChanges, so they become commit messages written into version history and shared with collaborators, frozen at commit time.

Two facts make it a blocker rather than a nit:

platform-scripture ships its three sibling keys (beforeReplace, _failureMessage, afterReplace) English-only across 30 language blocks, including es. Merginwo contradictorypolicies for one key family, and a Spanish user's history permanently interleaves "Antes de insertar nota al pie" with "Before doing Replace".
The stated justificay says the values are"required" because the new whole-file guard fails without them, and that guard was added in the same commit.

Three ways out: translate them and file a ticket to do platform-scripture's tHistoryCommit_% family from the guard; or drop the whole-file guard from this PR, which removes the pressure entirely and also settles the scope question about a 192-key policy landing in a four-key PR.

jolierabideau:
Took option 3 — **dropped the whole-file guard**, and with it the three `%versionHistoryCommit_*%` Spanish values. Pushed as `dc8be09` + a merge of latest main (`46c3ca6`).

Reasoning for picking that over the other two: every one of findings 1, 6, 7, 13, 18 and 20 is a defect *of the guard*, and the guard was the only thing forcing the persisted-data translations. Removing it resolves the finding-4 blocker by removing the pressure rather than negotiating around it, deletes the finding-1 code path outright, and settles the scope question — a 192-key policy no longer lands in a four-key PR. Option 2 would have kept a guard with a known collection-abort defect plus an ad-hoc exemption list; option 1 would have committed the repo to localizing shared, frozen commit messages as a side effect of turning a test green, which is a product call.

The gap the branch exists to close stays closed: the four empty-chapter keys keep their own parity block and their Spanish values.

## Blockers

**Finding 4 — resolved.** Every fact checked out against the code. The values flow `commitVersionHistorySnapshot` → `paratextBibleSendReceive.commitChanges` (`editor-side-effects.utils.ts:75-87`). `platform-scripture` ships 33 language blocks and all three siblings exist only in `en`. Both the three values and the guard are gone.

**Finding 1 — resolved, and it was worse than the review captured.** It was only half-fixed at the head you re-checked. Renaming `es` failed loudly by then (118 failed / 61 passed), but the defect had *moved to `en`*: `Object.keys(localizedStrings[ENGLISH_LANGUAGE])` sat at module scope, so renaming the `en` block still gave `Tests: no tests` — 0 of 182 collected. Verified after the change, restoring the JSON each time:

| probe | before | after |
| --- | --- | --- |
| rename `es` → `es-XX` | 118 failed / 61 passed | 118 failed / 61 passed |
| rename `en` → `en-XX` | **`Tests: no tests`**, 0 collected | 116 failed / 63 passed |
| drop one empty-chapter `es` value | — | 1 failed / 178 passed, key named |

**Finding 2 — confirmed and fixed.** Pointing the test's import at `./empty-chapter-view.component` gives 182/182 in the node env, so the stated invariant was false. Collect time goes 235ms → 1.05s, so the split is worth keeping for the reason you identified — just not the reason written down. All three headers now say so, including `character-marker-control.const.ts`, which carried the same claim in different words ("cannot load a module that reaches for `document`").

**Finding 3 — resolved as a consequence.** The comment at `:130-133` claiming the family is en-only across the whole file is true again now that the three values are gone.

## Also fixed

- **10, 13** — both lived in the deleted block.
- **11** — confirmed: 14 exported `*_STRING_KEYS` arrays in the extension, 9 imported by hand. The TSDoc in both const modules now says the test covers each key list it imports by name, and that a new list needs its import added there.
- **23** — the unresolvable `{@link}` fixed in `book-not-available-view.const.ts` and `character-marker-control.const.ts` too, per "fix together."

## Deferred, with a ticket

- **20** — agreed this is the right shape: one repo-scoped guard globbing `extensions/src/*/contributions/localizedStrings.json` with a shrink-only allowlist, ratcheting the 444-key backlog instead of N copy-pastes. Filing it rather than reintroducing a per-extension guard here.
- **9** (`messageResource` double-covered via `model-text-panel.const.ts`), **14** (`Localization-Guide.md:228` "required for the build to pass" — verified false), **16-19** (test-structure cleanups), **21**, **24**, **25** (widen the sentence-case rule before its scope), **26** — folding into the same ticket. Note **18** inverts usefully now: with the guard gone, the per-key blocks are the only coverage again, so nothing is unfalsifiable.
- **22** — the `Session-URL:` trailer is a literal `<session URL>` placeholder on `3f121b5`; @jolierabideau needs to substitute the real URLs, which I can't generate.

## Verification

- `localized-strings.test.ts` 179 passing; `empty-chapter-view.component.test.tsx` 12 passing.
- Extension typecheck: **333 errors on `origin/main`, 333 on the branch** — zero introduced. (The 244 baseline in the old PR body predates the main merge.)
- ESLint clean on all touched files; prettier applied.
- Full `src/platform-scripture-editor` suite: 6 failures, all in `paragraph-style-label`/`paragraph-style-trigger` — new files from #2768 that I never touched. They **fail identically on a clean `origin/main` checkout** in this environment, so they are not from this branch; most likely a stale `platform-bible-react` build locally against main's new `shrink-step` context.

jolierabideau:
Following up on the one loose end in my verification note — the 6 `paragraph-style-*` failures. **Diagnosed, and they are a local artifact, not a code problem.** Full extension suite is green: **88 files, 1419 tests, all passing.**

Root cause: I was running from a `git worktree` whose `node_modules` I had symlinked to the main checkout's. `node_modules/platform-bible-react` is a workspace symlink to `lib/platform-bible-react`, so the extension tests resolved `platform-bible-react` to the **main checkout's** lib source — which sits on an unrelated branch that predates #2768. `ShrinkStepOverride` doesn't exist there, so the import came back `undefined` and React threw `Element type is invalid`. Nothing to do with this branch, and main is not broken.

Proof: aliasing the bare `platform-bible-react` specifier to the worktree's own `lib/platform-bible-react/src/index.ts` (exact-match, so the `/experimental` subpath still resolves) takes those two files from 6 failed to 12/12, and the whole suite to 88/88.

| run | result |
| --- | --- |
| worktree resolving to the wrong branch's lib | 2 files failed, 6 tests failed |
| worktree resolving to its own lib | **88 files passed, 1419 tests passed** |

So the earlier "not run: repo-root `npm test`" caveat is narrowed: the extension suite this PR touches is fully green. CI builds fresh, so it never saw this.

The scratch worktree and alias config have been removed; nothing was left behind in either checkout.

katherinejensen00 (approved):
:lgtm: Thanks for making those changes, Jolie!

> @katherinejensen00 reviewed 6 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2717#-:-P1116b64o7PbWafukbw:bdv5izw)* status: all files reviewed, 1 unresolved discussion (waiting on jolierabideau).

katherinejensen00 (approved):
> @katherinejensen00 resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2717)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2720, opened 2026-08-25
docs: identify architecture decisions by slug instead of number

jolierabideau:
# Code Review Summary

**Branch**: adr-slugs

**Base**: origin/main

**Date**: 2026-08-25

**Files changed**: 19

## Overview

This branch changes how Architecture Decision Records are identified: entries in
`.context/standards/Architecture-Decisions.md` are now named by a stable `adr-`-prefixed kebab-case
slug instead of a sequential `ADR-NNNN` number. Each of the 28 pre-existing entries gained a
`**Formerly:** ADR-NNNN` line so historical references in PR history, Jira, and Discord still
resolve, and the guidance section was rewritten to explain the new rule (including what to do when
two branches append entries concurrently). Every reference to an ADR number that existed at the time
— in the log's own cross-references, three other `.context` standards docs, an investigation brief,
and 14 code-comment sites across `src/` and `extensions/` — was updated to the corresponding slug.

The motivation is process throughput rather than aesthetics: numbers were effectively claimed at
merge time, so two branches in flight could silently produce duplicate headings, and reconciling
that at merge slowed down review and merge of unrelated PRs. Slugs are chosen at write time and
never collide, so nothing has to be reserved or renumbered. **No executable code changed anywhere in
this branch** — every source-file edit is inside a `//` comment or a TSDoc block.

## API Changes

None. No public API surfaces changed. The branch touches no files in `lib/platform-bible-react/`,
`lib/platform-bible-utils/`, `lib/papi-dts/papi.d.ts`, or `extensions/src/*/src/types/*.d.ts`. The
one `.d.ts` file in the diff — `src/@types/paratext-bible-send-receive/index.d.ts` — changed only in
a TSDoc comment body (no declaration, signature, or type change). Verified with `git diff` that no
executable line changed.

None.

- [x] **`src/renderer/components/platform-bible-toolbar.tsx:82` — the rename turned a stale ADR
      reference into an actively misleading one.** The comment cited
      `adr-empty-is-zero-state-primitive` (formerly ADR-0016) as the decision that assigns
      width-driven collapse to `useShrinkStep` and "rejects CSS container queries for this job", but
      that entry is about the shadcn `Empty` zero-state primitive and says nothing of the sort. The
      mis-pointer was pre-existing on `main`; a bare number read as an opaque handle, while a topical
      slug asserts a claim the named entry does not make. _(fixed during review: searched `.context/`,
      `.claude/`, and `CLAUDE.md` — no entry anywhere records container-query reasoning, no `PT-4344`
      appears in the ADR log, and `useShrinkStep` exists only in this comment. Dropped the ADR
      citation and kept the rationale inline as prose.)_
- [x] **`.context/standards/Architecture-Decisions.md:625` — same class of mis-reference inside the
      log itself.** Within `adr-empty-is-zero-state-primitive`, the note that `InstallFailedView` was
      renamed `RetryableErrorView` "by PT-4347" cited `adr-book-selection-is-summarized` (formerly
      ADR-0022), which is about summarizing book selections on a scope trigger. _(fixed during review:
      repointed to `adr-panel-readiness-from-sources` (formerly ADR-0027), which is the PT-4347 entry
      that actually documents the rename; verified at line 1406. Paragraph re-wrapped.)_
- [x] **Comment re-wrapping was applied inconsistently across the sweep.** Eight comment sites were
      re-wrapped to the surrounding ~100-column convention while six were left long (101–130 chars).
      Prettier does not reflow `//` comments and there is no `max-len` rule, so tooling would never
      have caught it. _(fixed during review: re-wrapped all six —
      `scripture-text-grid.component.tsx`, `extension.service.packaged-extensions.test.ts`,
      `sync-blocked-banner.component.tsx`, `resource-text-panel.web-view.tsx`,
      `sync-status-button.component.tsx`, and `platform-bible-toolbar.tsx`. Verified: **zero** lines
      added by this branch now exceed 100 characters. The few remaining short lines are followed by
      unbreakable long tokens (an ADR slug or the log's file path).)_

Author response: The author was not aware of either mis-pointed reference — both predate this branch
and were inherited as wrong numbers. They asked for the unresolvable one to be dropped and the
resolvable one to be repointed, and for the wrapping to be made consistent.

### Minor — Consider

- [x] **The backtick convention contradicted its own first consumers.** The new rule at
      `Architecture-Decisions.md:27` says to cross-reference a slug "in backticks," and the markdown
      files follow it, but all 14 code-comment references this branch introduced used bare slugs.
      _(fixed during review: after weighing backticking vs. relaxing the rule to markdown-only, the
      author chose to backtick. Applied to 17 sites across 13 files; the deciding factor was that the
      surrounding comments already backtick every other identifier and path, and two sites are TSDoc
      where backticks actually render on editor hover. Verified only comment lines changed, all slugs
      now backticked, no line over 100 chars.)_
- [x] **~15–34 prose body lines in `Architecture-Decisions.md` and the investigation `brief.md` ran
      past the ~100-column wrap** because substitution was done in place. Cosmetic only — Prettier
      passes (`proseWrap` is unset) and the file already had 262 over-length lines at the merge base.
      _(fixed during review: re-wrapped 25 branch-touched paragraph blocks across
      `Architecture-Decisions.md`, `brief.md`, `Component-Selection-Quick-Reference.md`, and a
      blockquote in `Component-Builder-Patterns.md`, preserving bullet markers and continuation
      indents. Verified the prose is word-for-word identical before and after. Every remaining
      over-100 line is inherently unwrappable: 26 `## adr-…` headings and one table row.)_
- [x] **`adr-launch-token-withdrawn` bakes a status into a slug the new rule says must never change.**
      Harmless for a permanently-withdrawn entry, and renaming it now would break the `**Formerly:**`
      continuity, but it is a precedent worth not repeating. _(fixed during review: left the slug
      alone and added a guidance line — name the decision, never its status; record
      `Withdrawn`/`Superseded` on the entry's `**Status:**` line instead.)_
- [x] **The withdrawn entry referred to itself in the third person** ("the inverted trade-off
      `adr-launch-token-withdrawn` claimed to avoid", "The sibling `projectId` bug
      `adr-launch-token-withdrawn` reported") — a faithful but odd-reading product of the mechanical
      substitution. _(fixed during review: both changed to "this decision" and the paragraph
      re-wrapped; no self-references remain in the entry body.)_
- [ ] ~~`CLAUDE.md` "Recording Architecture Decisions" does not mention the slug identity
      convention.~~ _(Author: if it had no naming convention before, leave it. Verified against the
      merge base — that section contained no naming or numbering language at all, so nothing in it
      became stale.)_
- [ ] ~~No ADR entry was appended recording the identity-scheme change itself, even though CLAUDE.md
      lists "introducing a new pattern or top-level structure" as recordable.~~ _(Author: leaving it
      unrecorded was deliberate — the existing entry that defines how ADR entries work was updated
      instead, which is where the rule belongs.)_

Author response: The author acted on every substantive minor finding and dismissed the two that were
judgment calls, with reasons. On the backtick question they asked for the tradeoffs before deciding.

#### Shared Regions Modified

None. No changed file contains a `#region shared with` marker overlapping a changed line.

#### Extension Config Changes

None. The only files under `extensions/` are `.ts`/`.tsx` sources in
`extensions/src/platform-scripture-editor/src/` — no config or build files.

## Positive Observations

- **The conversion is complete and was verified mechanically, not by eye.**
  `git grep -niI "ADR-[0-9]"` returns nothing in tracked files outside the `**Formerly:**` lines, and
  every `adr-<slug>` token appearing anywhere under `.context/`, `src/`, `extensions/`, and `lib/`
  resolves to an existing `## adr-…` heading — 0 unresolved references.
- **The number-to-slug mapping is faithful.** All 28 entries carry a `**Formerly:**` line, the old
  numbers run 0001–0028 with no gaps or duplicates, all 28 slugs are unique, and spot-checks of
  0011, 0013, 0016, 0019, 0024, and 0028 against the pre-rename headings all match.
- **The `**Formerly:**` affordance is well-designed** — placed consistently as the first bullet of
  every entry, and the guidance explicitly tells future authors *not* to add it to new entries, so it
  cannot calcify into permanent boilerplate.
- **The new rule solves a real process hazard** and tells a conflict-resolver exactly what to do
  ("keep both entries"), which is the part that most conventions leave unsaid.
- **The entry template, `Status`, and `Superseded by` lines were updated in step with the body**, so
  the file never documents one scheme while using another.
- **Slugs are a genuine readability win at the call site** — `adr-verse-zero-resolves-to-verse-one`
  says what the decision was without a lookup, where `ADR-0019` did not.
- **Comment-only change throughout** — no runtime logic, no exports, no localized string keys, no
  build config. Prettier and ESLint clean on all changed files.

## Interview Notes

- **Stated purpose**: rename ADR entries to use unique slugs instead of numbers, add a `**Formerly:**`
  field to existing entries, and update the code references that exist today.
- **Reasoning behind the scheme** (author, in their own words): the team wants a consistent naming
  solution that works *while you are working*; doing anything at merge time slows down the ability to
  review and merge PRs. Slug naming will drastically reduce the rate of ADR naming conflicts. This is
  a throughput argument, not a cosmetic one, and it is the strongest justification in the change.
- **Deliberate decision not to append an ADR entry for the scheme change itself.** The author
  confirmed this was intentional: the existing entry/section that defines how ADR entries work was
  updated instead. Worth a reviewer's confirmation only because CLAUDE.md's "Recording Architecture
  Decisions" section arguably calls for an entry here; the author's position is that the rule belongs
  with the rule, not in the append-only log.
- **Two inherited mis-references surfaced.** The author was not aware of either
  (`ADR-0016` in the toolbar comment, `ADR-0022` inside the `Empty` entry). Both were wrong on `main`
  already — this branch did not introduce them, it made them legible. This is a genuinely interesting
  side effect of the change and is arguably an argument *for* slugs: a wrong number is invisible, a
  wrong slug is self-evident. Both are now fixed.
- **The author demonstrated clear understanding throughout.** They asked for the pros and cons before
  deciding the backtick question rather than accepting a recommendation blindly, and gave a specific,
  correct rationale for the scheme when asked. No areas were deferred to AI, and no questions went
  unanswered or vague.
- **Nothing further flagged** at wrap-up.

## In-Review Quality Check

All four checks passed clean; no fixes were required beyond the review changes themselves.

| Check | Result |
| --- | --- |
| `npm run typecheck` | Passed clean (root + all extension workspaces) |
| `npm run lint` | Passed clean (exit 0). One pre-existing warning — `import/no-named-as-default` for `UnsubscriberAsyncList` — unrelated to this branch |
| Prettier (changed files only) | All source files already formatted; the four `.context/*.md` files are prettier-ignored |
| `npm test` | Passed clean — 63 test files, 901 tests, 0 failures |

Library rebuilds were not applicable (no `lib/platform-bible-react/` or `lib/platform-bible-utils/`
changes).

## Suggested Review Focus

- [ ] **Confirm the deliberate choice not to log this decision as its own ADR entry.** CLAUDE.md
      names "introducing a new pattern or top-level structure" as recordable, and the rationale
      (numbers claimed at merge → silent duplicate headings) currently survives only in the commit
      diff and this summary. The author's position is that updating the how-ADRs-work section is the
      right home for it. Reasonable, but it is the one place where the branch consciously departs
      from a written project rule — worth an explicit sign-off rather than an implicit one.
- [ ] **Decide whether `**Formerly:**` is permanent or has a sunset.** It is currently on all 28
      entries with instructions never to add it to new ones, so it is a fixed-size legacy band that
      shrinks only if entries are ever removed. Fine either way, but the intent is not written down.
- [ ] **Sanity-check the 28-entry slug mapping.** It was verified programmatically against the
      `**Formerly:**` lines (unique slugs, no gaps or duplicates in 0001–0028, every reference
      resolving to a real heading), so a reviewer does not need to re-check it by eye — but it is the
      one thing that would be expensive to discover wrong later, since old numeric references in Jira
      and Discord depend on it.
- [ ] **Three semantic changes sit inside an otherwise mechanical diff — confirm each read.**
      (1) `adr-book-selection-is-summarized` → `adr-panel-readiness-from-sources` inside the `Empty`
      entry, a pre-existing mis-reference. (2) `Entry-Point-Guide.md:127` — a deliberately vague
      pointer ("the ADR on deferred menu availability", no number) was hardened into a precise
      citation of `adr-menus-always-available-gate-at-submission`. The target is right, but this
      branch invents a citation the original author chose not to make; revert it to the vague form if
      you would rather the sweep stayed purely mechanical. (3) The toolbar comment's dropped citation
      — **now moot**: `main` has since rewritten that comment and created a real
      `adr-toolbar-shrink-measurement` entry for exactly the container-query rationale, so the rebase
      takes `main`'s version and the citation resolves.

- [ ] **`main` landed the same slug idea independently while this PR was open**, as
      `## ADR-toolbar-shrink-measurement` and `## ADR-simple-mode-column-minimums` — uppercase
      prefix, new entries only, old numbers left in place. The rebase resolves in favor of this
      branch (lowercase `adr-`, all 28 retro-renamed) and normalizes those two entries plus the
      `adr-simple-mode-column-minimums` reference in `title-bar-narrow-width.spec.ts`. Worth an
      explicit sign-off, since it overrides guidance already on `main`. Incidentally it is also the
      convention working as designed: two branches picked non-colliding slugs with nothing to
      reconcile but casing.

- [ ] **The frozen investigation record is deliberately untouched.**
      `.context/research/investigations/prd-power-donna-…-0-003/brief.md` declares itself a frozen
      record in its own header, so its `ADR-0002` references stay as written and are the only
      `ADR-NNNN` tokens left in tracked files outside `**Formerly:**` lines.
      `.context/research/paratext-9-features/08_Checklists.md:35` is likewise unchanged: it points at
      the log *file*, carries no ADR number, and so has nothing for a number→slug sweep to convert.

AI-assisted — reviewed with Claude Opus 5

---

ADR entries were numbered ADR-NNNN, and numbers were effectively claimed at merge rather than at write. Concurrent branches routinely appended the same number, so whoever merged second had to renumber their entry and chase down every cross-reference — a recurring, avoidable merge conflict.

Entries are now identified by an `adr-`-prefixed kebab-case slug chosen at write time, which two branches pick independently and never have to reconcile. A tail conflict is resolved by keeping both entries.

- Rewrite the 28 existing headings to `## adr-<slug>: {title}` and give each a `**Formerly:** ADR-NNNN` bullet so older references still resolve.
- Update all cross-references in the log, the standards docs, the Send/Receive investigation brief, and 14 source files (comments only).
- Replace the "numbers are claimed at merge" convention and the entry template with the slug rule; new entries carry no number and no `**Formerly:**` line.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2720)


katherinejensen00 (changes requested):
Thanks for doing the renaming, Jolie!

1. prd-power-donna-…-0-003/brief.md:3 — a frozen record was edited. Revert it. Sharpened: the PR skipped 08_Checklists.md:35, which has an identical pointer to the log and is not frozen. So the one research file it touched was the off-limits one, and the one it left alone was fair game.
2. resource-text-panel.web-view.tsx:408 + verse-display.utils.ts:12 — adr-verse-zero-resolves-to-verse-one drops the title's "on single-verse display surfaces" qualifier, so the slug asserts the rule the sentence is negating ("chapter surfaces must not…"). Someone "fixes" the chapter view and blanks the Psalm superscriptions.
3. Entry-Point-Guide.md:127 — an undisclosed third semantic change: a deliberately vague pointer with no ADR number was hardened into a precise citation. Target's right, but it invents a citation, and the body says the two disclosed fixes are "the only semantic changes."
4. Architecture-Decisions.md:758/:761 — the in-review fix left "this decision" where the only named decision is a different entry. Binds to the wrong one, read

> @katherinejensen00 reviewed 19 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2720#-:-P-yPCgZ4lerTv07DZ_o:bcxi74b)* status: all files reviewed, 1 unresolved discussion (waiting on jolierabideau).

___
*[`.context/standards/Architecture-Decisions.md` line 105 at r1](https://reviewable.io/reviews/paranext/paranext-core/2720#-P-y_VeS1gLFIf8moyH1:-P-y_VeS1gLFIf8moyH2:b4hrc3c) ([raw file](https://github.com/paranext/paranext-core/blob/97ff340c794523de2fd23ac52fdbed109d7b7684/.context/standards/Architecture-Decisions.md#L105)):*
> ```Markdown
> ## adr-menus-always-available-gate-at-submission: Menus stay always-available; back ends gate at submission. Writers of mutable shared state are DataProviders, not NetworkObjects
> 
> - **Formerly:** ADR-0003
> ```

1. prd-power-donna-…-0-003/brief.md:3 — a frozen record was edited. Revert it. Sharpened: the PR skipped 08_Checklists.md:35, which has an identical pointer to the log and is not frozen. So the one research file it touched was the off-limits one, and the one it left alone was fair game.

jolierabideau:
@katherinejensen00 All four addressed, rebased onto latest `main`. Pushed as `541d77e`.

**1. Frozen record edited — reverted.** You're right, and the file says so in its own header ("Frozen record — investigation as run 2026-07-02, kept as a worked example"). `brief.md` is back to `main` verbatim; its `ADR-0002` references are now the only `ADR-NNNN` tokens left in tracked files outside `**Formerly:**` lines.

On the sharpened half — I'm leaving `08_Checklists.md:35` alone, and I don't think it's the same class of miss. Its pointer is to the log *file*, not to an entry:

> see the checklist reuse map in `.context/standards/Architecture-Decisions.md`

There's no ADR number in it, so a number→slug sweep had nothing to convert there. Adding `adr-reuse-shared-checklist-framework` would be inventing a citation the original author chose not to make — which is exactly what your point 3 objects to. Happy to add it if you'd rather, but then it should be disclosed as a fourth semantic change rather than folded into the mechanical sweep. Your call.

**2. Slug drops the "single-verse surfaces" qualifier — renamed.** This was the best catch of the four. At both sites the sentence negates the rule for chapter surfaces while the citation asserted it unconditionally:

> whole-chapter surfaces like this one must not (see `adr-verse-zero-resolves-to-verse-one`)

Rather than patch the two call sites, I moved the qualifier into the slug so it can't be dropped again: `adr-verse-zero-resolves-to-verse-one` → **`adr-single-verse-surfaces-resolve-verse-zero-to-one`**. Scope-first, so the constraint is the part you read before the rule. Applied to the heading and all 5 citation sites. The entry title is unchanged. Free to do now since the slug hasn't merged, and `**Formerly:** ADR-0019` carries the continuity either way.

**3. Undisclosed third semantic change — disclosed, citation kept.** Correct on both counts: it is a third semantic change, and the body claimed there were two. I verified the target before keeping it — `adr-menus-always-available-gate-at-submission` is precisely the deferred-menu-availability decision (its Decision section: "Menus stay always-available… the back end performs permission/state checks at submission time"). So the citation is accurate, but your objection is about provenance rather than accuracy, and I've written it up that way in the PR body under Suggested Review Focus, including the option of reverting it to the vague form if you'd rather the sweep stayed purely mechanical.

**4. "this decision" binds to the wrong entry — fixed.** Right, and my in-review "fix" caused it: swapping the third-person self-reference for "this decision" put an unqualified demonstrative in a sentence whose only *named* decision is `adr-one-shot-launch-parameters`.

- `:758` → "the inverted trade-off **this withdrawn decision** claimed to avoid"
- `:761` → "The sibling `projectId` bug **reported here**"

Both now resolve only to the containing entry.

---

**One thing you should know about the rebase, since it changes the diff you reviewed.** `main` landed the same slug idea independently while this PR was open — `## ADR-toolbar-shrink-measurement` and `## ADR-simple-mode-column-minimums`: uppercase prefix, new entries only, old numbers explicitly left in place ("don't renumber them").

I resolved in favor of this branch — lowercase `adr-`, all 28 retro-renamed — and normalized `main`'s two entries plus the `adr-simple-mode-column-minimums` reference in `title-bar-narrow-width.spec.ts`. Reasoning: `adr-` reads as one token in a code comment and matches the repo's kebab-case rule/skill names, and a half-converted log documents two schemes at once, which is the thing this PR exists to stop. That does override guidance already sitting on `main`, so it's flagged for explicit sign-off rather than assumed.

Two consequences:

- **Your point 3's sibling is now moot.** `main` rewrote the toolbar comment and created a real `adr-toolbar-shrink-measurement` entry for exactly the container-query rationale the old `ADR-0016` citation mis-attributed. The rebase takes `main`'s version, so instead of a dropped citation there's now a correct one.
- The log is at **30 entries**, and every `adr-` token across `.context/`, `src/`, `extensions/`, `lib/`, and `e2e-tests/` resolves to a real heading — 0 unresolved.

Incidentally, this is the convention working as intended: two branches in flight picked non-colliding slugs and there was nothing to reconcile but casing. Under `ADR-NNNN` both would have appended the same number.

Verification after the rebase: `npm run typecheck` clean, `npm run lint` 0 errors (1 pre-existing `import/no-named-as-default` warning), `npm test` 904/904 passing, Prettier clean on changed files. Still comment-and-docs only — no executable line changed.

katherinejensen00 (approved):
:lgtm: Great work, Jolie!

> @katherinejensen00 reviewed 14 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2720#-:-P-zKnR9-GKDq0hWR6GA:buirp)* status: all files reviewed, 1 unresolved discussion (waiting on jolierabideau).

PR #2721, opened 2026-08-25
PT-4346: show books from open resources

mattgetgen:
# Code Review Summary

**Branch**: pt-4346-show-books-from-open-resources

**Base**: origin/main

**Date**: 2026-08-25

**Files changed**: 47

## Overview

In Simple mode, the global book/chapter/verse control now offers books from every project open in
the window — the active project, model texts, collections, and other open resources — not just the
active project's own books. The expanded list is opt-in behind a "Show more books" toggle; books the
active project lacks render greyed but stay fully selectable, so a user can navigate to them. The
mechanism is a new shared web view state key, `navigableProjectIds`: a web view whose definition
`projectId` is only its *container* (a resource panel, or the Scripture Text Grid which hosts many
projects at once) declares the projects it actually displays, and the toolbar reads those
declarations back, subscribes to each project's `platformScripture.booksPresent`, and hands the
union to `BookChapterControl` as a second book list.

The work is deliberately scoped to Simple mode. Power mode's controls are untouched — the component
API is open to that team to adopt as they see fit, and after this review the toolbar disables the
lookup outright in Power mode so it costs nothing there. `BookChapterControl` stays presentational:
it renders exactly the two lists it is given and marks the difference as dimmed; it never discovers
projects itself.

## API Changes

- `lib/platform-bible-utils` (`platform-bible-utils/experimental`): added
  `NAVIGABLE_PROJECT_IDS_WEB_VIEW_STATE_KEY` const export (`"navigableProjectIds"`)
- `lib/platform-bible-utils` (`platform-bible-utils/experimental`): added
  `isNavigableProjectIds(value: unknown): value is string[]`
- `lib/platform-bible-react`: `BookChapterControlProps` — added optional
  `getAdditionalBookIds?: () => string[]`
- `lib/platform-bible-react`: `BOOK_CHAPTER_CONTROL_STRING_KEYS` — added
  `%webView_bookChapterControl_bookNotInProject%`, `%webView_bookChapterControl_showMoreBooks%`,
  `%webView_bookChapterControl_showProjectBooksOnly%` (so `BookChapterControlLocalizedStrings` gains
  three optional keys)
- `lib/platform-bible-react`: `BookChapterControl` — behavior change, `localizedStrings` lookups for
  `%…_selectChapter%`/`%…_selectVerse%` now fall back with `||` instead of `??`
- `lib/platform-bible-react` (internal, not re-exported from the package index): added
  `deriveBookChapterControlBookLists()` / `BookChapterControlBookLists` and exported
  `groupBooksBySection()` in `book-chapter-control.utils.ts`; added `dimmedReason` to the
  module-private `BookItemProps`
- `lib/papi-dts/papi.d.ts`: unchanged — no `@papi/` module surface was touched
- `extensions/src/*/src/types/*.d.ts`: unchanged
- Built declarations (`lib/platform-bible-react/dist/index.d.ts`,
  `lib/platform-bible-utils/dist/experimental.d.ts`) were regenerated and match the sources

None.

- [x] **`book-chapter-control.utils.ts`: a project book could disappear when the user expanded the
      list.** The widened list was built as `ALL_BOOK_IDS.filter(...)`, and `ALL_BOOK_IDS` excludes
      the ids the Canon library marks obsolete (`JSA, JDB, TBS, SST, DNT, BLT, 3ES`) — all of which
      `Canon.isBookDC` claims, so they render today. `getBookIdsFromBooksPresent` does no obsolete
      filtering, so a project whose `booksPresent` flagged one got it in `projectBookIds`; the
      moment any additional book arrived, that book vanished from expanded browsing, from
      `calculateTopMatch`, and from search (which spans the reachable list regardless of the
      toggle). _(fixed during review: the union is now ordered from `Canon.allBookIds`, so
      `reachableBooks ⊇ projectBooks` holds by construction and an obsolete book reachable through
      an open resource is offered too; three tests pin it, including the superset invariant)_
- [x] **`book-chapter-control.component.tsx`: the comment justifying `??` → `||` was factually
      wrong.** It claimed `||` covers the localization service echoing a missing key, but
      `useLocalizedStrings` resolves a missing key to the truthy string `'%key%'`, which `||` passes
      straight through. Only the empty-translation case is actually fixed by `||`. The branch's own
      `book-chapter-control-localization.test.ts` states this correctly, so the two contradicted each
      other. _(fixed during review: comment narrowed to the empty-string case and now points at the
      localization test as the guard for the missing-key case)_
- [x] **`platform-bible-toolbar.tsx`: Power mode paid for work it discarded.** `useOpenResourceBookIds`
      was called unconditionally but its result was thrown away in Power mode, so Power mode opened a
      `platform.base` data provider and a `booksPresent` subscription for every open project in the
      window with zero consumers. _(fixed during review: added an `isEnabled` parameter — the hook
      passes `undefined` to `useEvent` and short-circuits project enumeration when disabled; the
      toolbar passes `!isPowerMode`. Three tests cover it)_
- [x] **The `navigableProjectIds` publish block was triplicated across three web views.** ~30
      near-identical lines each — same `useWebViewState` call, same effect shape, same
      `resolveNavigableProjectIdsWrite` body, same 5-line dependency array, and a verbatim 5-line
      "Hidden case" comment. Only the readiness guard and the displayed-id expression differed.
      _(fixed during review: extracted `usePublishNavigableProjectIds(useWebViewState,
      displayedProjectIds, isReady)` beside `navigable-project-ids.utils.ts`; each web view is now a
      five-line call with its own readiness expression)_
- [x] **The readiness gate that guards the publish was untested.** The comparison helpers were well
      covered, but the part that can actually corrupt persisted layout state — "don't publish before
      the sources have arrived, because a transiently empty derived list is indistinguishable from
      'everything was removed'" — was inline in each web view and its three copies could drift.
      _(fixed during review: 10 tests on the extracted hook, including "does not wipe a persisted
      list while not ready", "publishes once readiness flips from false to true", reorder-is-not-a-
      write, and the invalid-persisted-value path)_
- [x] **`assets/localization/en.json`: new English strings were title case.** `"Select Chapter"` /
      `"Select Verse"` violate the sentence-case rule in `capitalization.mdx`; the Spanish values
      added in the same commit were already sentence case, so en/es disagreed. Both keys are net-new
      on this branch, so this was the one moment to fix them without a replacement key plus
      `fallbackKey`/`deprecationInfo`. _(fixed during review: "Select chapter" / "Select verse", plus
      the matching hardcoded fallbacks in the component)_
- [x] **The toggle label "Show all books" overstated what it does.** The expanded list is
      `activeBookIds ∪ additionalBookIds`, never the whole canon — in the toolbar it can be a single
      book (the current reference's, when the project lacks it), so the button said "Show all books"
      and revealed exactly one. _(fixed during review: renamed to "Show more books" — the author
      chose this over the more specific "Show books from open resources" because Power mode may use
      it for other projects' books too. The localize key was renamed `showAllBooks` → `showMoreBooks`
      (free, since it is net-new), along with both locale values, the internal identifiers, the
      TSDoc, and 12 test assertions)_
- [x] **`book-item.component.tsx`: the dimmed state was colour-only for sighted users.** The reason
      reached assistive tech through the accessible name but was never rendered — no tooltip, badge,
      or legend — which the Responsiveness guideline and Design Principles both require.
      _(fixed during review: dimmed rows now carry a `Tooltip` with the localized reason. The trigger
      wraps the section-border div rather than the `CommandItem` because cmdk's `Item` assigns its
      own `onPointerMove` after spreading incoming props, which silently drops the handler Radix
      needs; neither `TooltipProvider` nor `Tooltip` renders an element, so the list's DOM shape is
      unchanged. Three tests cover it)_
- [x] **`book-item.component.tsx`: the dimmed state inverted the affordance hierarchy.**
      `tw:text-muted-foreground/50` over `tw:bg-muted/50` halved an already-low-contrast token, so
      *selectable* dimmed items ended up lower-contrast than genuinely `disabled` ones
      (`opacity-50`); and the inner book-id span sets its own `tw:text-muted-foreground`, so only the
      book *name* dimmed and the row read half-dimmed. _(fixed during review: replaced with
      `tw:opacity-70 tw:data-selected:opacity-100` — name and id dim together, a selectable item
      stays more legible than a disabled one, and the highlighted row renders at full contrast like
      any other book, which is what the author asked for)_
- [x] **No Storybook deliverable for the new behavior.** The stories file already drove
      `getActiveBookIds` but was not touched, so nothing demonstrated `getAdditionalBookIds`, the
      toggle, or the dimmed state — the UI Review guideline lists a Storybook deliverable as a
      pre-review checklist item. _(fixed during review: added `WithBooksFromOpenResources` with a
      play function asserting the extra books are hidden by default, appear greyed with the "not in
      this project" accessible name and without `aria-disabled` after "Show more books", and
      disappear again on "Show project books only")_

Author response: the author agreed with every Important finding and asked for all ten to be fixed
in-review. On the obsolete-books case they confirmed the intent directly — "we should allow
'obsolete' books to show up if they appear in any open scripture" — which settled the fix as a union
from `Canon.allBookIds` rather than a filter-back.

### Minor — Consider

- [x] **The toggle double-encoded its state.** It set `aria-expanded={isShowingMoreBooks}` *and*
      flipped its own label, so a screen reader announced "Show project books only, expanded" — the
      label naming the next action, `aria-expanded` naming the current state. No `aria-controls`
      either. _(fixed during review: the WAI-ARIA disclosure pattern wants a constant label plus
      `aria-expanded`, which conflicts with this repo's microcopy rule that a label must say what
      will happen; the action label wins, so `aria-expanded` was dropped and `aria-controls` added
      pointing at a `useId`-generated list id. A test pins the decision. This also surfaced an
      incomplete rename — `setIsShowingAllBooks` and `canShowAllBooksToggle` had dodged the earlier
      replace on capitalization)_
- [ ] ~~The toggle also changes what the quick-nav chapter/verse buttons do: `useQuickNavButtons`
      receives `isShowingMoreBooks ? reachableBooks : projectBooks`, which is the book-rollover set,
      so a control labelled as a view change also alters navigation at book boundaries.~~
      _(Author: intentional — rollover should follow the expanded list. The expanded list is the
      navigation universe.)_
- [x] **The accessible name mixed languages.** It concatenated a localized fragment onto an English
      base (`Canon.bookIdToEnglishName`), so Spanish announced "Genesis (GEN), no está en este
      proyecto". _(fixed during review: the name is now built from `bookDisplayName` /
      `bookDisplayId`, the same localized values the row renders — both fall back to English when no
      `localizedBookNames` is passed, so the undimmed case and existing consumers are unchanged. No
      change to the `dimmedReason` contract was needed. A test asserts "Apocalipsis (APO), no está en
      este proyecto")_
- [x] **`getBookIdsFromBooksPresent(booksPresent)` was computed in two independent places** in the
      toolbar with the same input and dependency. _(fixed during review: hoisted to one
      `projectBookIds` memo feeding both `fetchActiveBookIds` and the `additionalBookIds` memo, whose
      dependency is now `projectBookIds`. Purely internal — no exported surface changed, which the
      author asked to confirm. Also caught a stale "show all books" comment left by the rename)_
- [x] **`filteredBooksByType` hand-rolled `groupBooksBySection`** — the empty
      `Record<Section, string[]>` literal and the per-section filter loop, duplicating the
      section-membership rules. _(fixed during review: `groupBooksBySection` exported with TSDoc; the
      search path is now `groupBooksBySection(reachableBooks.filter(...))`, ~13 lines shorter, with
      the rules in one place)_
- [x] **Ref write during render.** `isCurrentBookOutsideProjectRef.current = …` mutated a ref during
      render, so a render React discards (concurrent interruption, Strict Mode double-render) still
      mutated it. _(fixed during review: replaced with a plain derived boolean that
      `handleOpenChange` closes over. The original comment assumed this would churn the callback on
      every reference change, but that is only true when depending on `scrRef` or the set — depending
      on the boolean re-registers the imperative handle only when the current book actually crosses
      in or out of the project, the same order of churn the handle already accepts from `disabled`.
      Writing in an effect was rejected for the reason the original comment gives: effects run after
      commit, so an imperative `open()` in the same synchronous stretch would read a stale value)_
- [x] **Duplicated comment on the string-keys spread** in `web-view.component.tsx` and
      `platform-bible-toolbar.tsx`. The `[...X_STRING_KEYS]` spread matches established practice and
      those other call sites carry no comment. _(fixed during review: comment dropped from both)_
- [x] **`getAdditionalBookIds` is silently ignored unless `getActiveBookIds` is also supplied**, and
      the public TSDoc did not say so. _(fixed during review: one sentence added — "Ignored unless
      `getActiveBookIds` is also passed: without a project list the control already offers the whole
      canon, so there is nothing left for this to add")_
- [ ] ~~The `??` → `||` swap is a behavior change for existing valid input: a consumer deliberately
      passing `''` to blank a heading now gets "Select chapter".~~ _(Author agreed with the
      assessment and chose to leave it: an empty translation is far more likely than a consumer
      deliberately blanking a heading, and `||` is what stops a blank label. The corrected comment
      documents the tradeoff.)_
- [x] **`getOpenProjectIds` was broader than the hook's name suggested** — it collects
      `definition.projectId` from *every* open web view, so a project settings tab or checks view
      widens the toolbar's book list, while the hook was named `useOpenResourceBookIds`.
      _(fixed during review: the author confirmed the broad behavior is intended and chose to
      broaden the name. Renamed to `useOpenProjectBookIds` / `use-open-project-book-ids.hook.ts`
      (git tracked it as a rename), with `openProjectBookIds` at the call site and the log prefix
      updated. The docs now state plainly that every project-scoped web view counts, and that
      resources are projects here)_

Author response: the author asked for eight of the ten minor findings to be fixed and dismissed two
with reasoning (rollover following the expanded list is intentional; the `??` → `||` behavior change
is acceptable). On the accessibility and React-idiom questions (`aria-expanded`, the render-time ref
write) the author explicitly deferred the decision — see Interview Notes.

#### Shared Regions Modified

None. No `#region shared with` markers exist in any changed file; the only `#region`/`#endregion`
pairs present are plain code-folding regions with no shared-with URL.

#### Extension Config Changes

None. The only files changed under `extensions/` are TypeScript sources in
`extensions/src/platform-scripture-editor/src/` — no `webpack.config.ts`, `tsconfig.json`,
`package.json`, or `.eslintrc*`.

#### Build Config / Process

`.github/workflows/` changes; no dependency additions, removals, or version changes.

## Positive Observations

- Test coverage for the new logic is unusually thorough and mostly behavioral rather than
  implementation-shaped: BookChapterControl cases covering collapsed/expanded browsing, search
  spanning all reachable books, seed-on-open when the current book is outside the project,
  quick-navigation bounds, and toggle label state; BookItem dimmed-vs-disabled precedence and
  accessible-name tests; `deriveBookChapterControlBookLists` including the peripheral-id exclusion;
  and the open-project hook's dedup, canon ordering, malformed web view state, provider rejection,
  unmount unsubscribe, and "unchanged project set does not rebuild subscriptions" cases.
- `book-chapter-control-localization.test.ts` is a durable structural guard, not just coverage: it
  drives off `BOOK_CHAPTER_CONTROL_STRING_KEYS` and fails if any key lacks a non-empty, non-self
  English value. It caught the stale key during this very review, when the rename left the built
  `dist/` behind — exactly the failure mode it exists for.
- ADR-0029 records the web-view-state declaration mechanism with three named-and-rejected
  alternatives, its consequences, and a revisit trigger — what the root `CLAUDE.md` asks for, and the
  extension code visibly follows the consequences it lists.
- `getNavigableProjectIdsToPublish` comparing as sets rather than arrays is the right call given
  every publish lands in layout persistence, and the tests pin the reorder-is-not-a-write case.
- `resolveNavigableProjectIdsWrite` validating persisted state through `isNavigableProjectIds`
  instead of trusting the `useWebViewState<string[]>` type assertion is the correct treatment of
  layout-persisted state, and its TSDoc is honest that it checks shape only.
- `dimmedReason` bundling the visual state and its accessible explanation into one prop — so the
  greying and its reason cannot drift apart — is a nicer API than a separate `isDimmed` boolean, and
  the dimmed-vs-disabled split mirrors the existing `NumberedItemGrid` precedent rather than
  inventing one.
- The toggle is immediate-apply and its own undo, which is exactly the Applying Changes carve-out for
  view toggles — correctly no toast or confirmation. It is deliberately rendered outside `CommandList`
  so cmdk treats it as a control rather than an arrow-navigable option, and suppressed while
  searching and when there is nothing extra to offer.
- The new prop is purely additive and the no-`getAdditionalBookIds` path reduces to the previous
  grouping exactly, so the five other `BookChapterControl` consumers in the repo are unaffected.
- `useOpenProjectBookIds` reuses the established web-view lifecycle pattern from
  `use-project-picker-data.hook.ts` rather than inventing a new mechanism, and both
  `react-hooks/exhaustive-deps` suppressions carry specific justifications.
- Localization is complete and correctly placed: all new keys are core `%webView_*%` keys added to
  both maintained locales, with no in-place edit to an existing key's meaning, so no `metadata.json`
  deprecation is owed.
- Comments across the diff are forward-facing: no ticket references for in-PR work, no review-finding
  IDs, no "previously/used to" narration.

## Interview Notes

**Stated purpose.** In Simple mode, extend the viewable books in the BCV control to all books
accessible through all open projects and resources — the active project, model texts, collections.
Books not in the active project are greyed but still viewable. Explicitly scoped to enable the Power
team, without extending it to the Power setting: "I will allow them to do with it whatever they
want."

- Obsolete books should be offered whenever they appear in any open scripture — settling the
  `ALL_BOOK_IDS` vs `Canon.allBookIds` question in favor of the union.
- The toggle label should be "Show more books" rather than the more specific "Show books from open
  resources", because Power mode may later use the same control for other projects' books.
- Book rollover in quick navigation *should* follow the expanded list — the expanded list is the
  navigation universe, not just a view filter.
- The broader-than-its-name behavior of `getOpenProjectIds` is intended; the author chose to broaden
  the hook's name rather than narrow the behavior.
- Dimmed rows should look like a normal book while hovered.

**Areas the author did not understand — cover these in the review meeting:**

- **Author does not understand the `navigableProjectIds` declaration mechanism and how the toolbar
  turns it into a book list.** Asked to walk through the most complex part of the change, the author
  said: "I honestly don't have a strong grasp on that part of the work. Claude provided most of
  that. I know that it's driven by the toolbar, and not the component." That is the load-bearing
  design of the whole feature — the shared web view state key, the three publishing web views, the
  readiness gate that keeps a transiently empty list from wiping persisted layout state, and the
  toolbar-side subscription fan-out. It should be walked through end to end rather than assumed
  settled. (The author was given a summary of the chain during the interview, but that is not the
  same as having designed or verified it.)
- **Author deferred the accessibility decision on the toggle** (`aria-expanded` vs a flipping action
  label): "I don't know what's best for these sort of things. Do what's best for our current design
  rules, and what's best for accessibility users." The resolution — action label wins over the
  WAI-ARIA disclosure pattern because the repo's microcopy rule requires it, `aria-expanded` dropped,
  `aria-controls` added — is a judgment call the reviewer should confirm.
- **Author deferred the React correctness decision** on the render-time ref write: "I don't know
  react well enough to make the best decision. Do what is most idiomatic and makes sense to not
  break the render step." The resolution replaced the ref with a derived boolean the callback closes
  over; the reviewer should confirm the dependency reasoning (depending on the boolean, not on
  `scrRef`) holds, since it is what keeps the imperative handle from re-registering on every
  reference change.
- The author also asked to confirm that hoisting the toolbar memo was not API-breaking, which
  suggests some uncertainty about which surfaces in this change are public.

Unresolved Critical/Important findings: none. Every Important finding was fixed in-review with the
author's agreement.

## In-Review Quality Check

- **`npm run build:platform-bible-react`** — passed. `dist/` artifacts rebuilt and staged (the
  `resizable-*` chunk hashes changed as a result). `platform-bible-utils` did not need a rebuild.
- **`npm run typecheck`** — passed across all 16 workspaces plus `typecheck:core`/`typecheck:erb`,
  except one **pre-existing** failure unrelated to this branch:
  `use-remove-character-marker.hook.ts(91,16): Property 'removeCharacterMarker' does not exist on
  type 'EditorRef'`. Confirmed by stashing the branch changes and re-running — it fails identically
  on the base state.
- **`npm run lint`** — initially failed with 3 `no-type-assertion` errors, all in the new
  `use-publish-navigable-project-ids.hook.test.ts` (none auto-fixable, so `lint-fix` would not have
  cleared them). Fixed at the source rather than suppressed: the test double's stored value is now
  typed `unknown` (which is what web view state actually is), and a near-duplicate inline double was
  deleted in favor of the shared factory. The one remaining assertion — producing the caller's `T`
  from the stored value in a double of a generic hook — is unavoidable and carries a single
  justified `eslint-disable-next-line`, matching the repo's existing idiom for generic test doubles.
  Repo-wide re-run: **exit 0**.
- **Prettier on branch-changed files** — all 28 reported unchanged; no formatting drift.
- **`npm test`** — **5,511 tests pass**. One failure:
  `editor-character-marker-contract.test.ts > declares every character-marker operation this
  extension depends on` — the same **pre-existing** `EditorRef` / installed-platform-editor mismatch
  as the typecheck error, not caused by this branch.

**Action for the author:** the two pre-existing `EditorRef.removeCharacterMarker` failures are not
from this work, but they are red on this branch and will be red in CI. Worth confirming they are
tracked elsewhere before merge.

## Suggested Review Focus

- [ ] **Walk through the `navigableProjectIds` mechanism end to end — the author cannot explain it.**
      How a web view declares its displayed projects, why the definition's own `projectId` is not
      enough, and how the toolbar fans out `booksPresent` subscriptions. This is the core design of
      the feature and the author deferred it to AI.
- [ ] **The readiness gate.** "Not yet loaded" and "everything was removed" are indistinguishable
      from a derived empty list, and publishing the latter wipes persisted layout state on remount.
      Now centralized in `usePublishNavigableProjectIds` and tested — confirm each web view's
      readiness expression genuinely covers every source its displayed-id list is derived from.
- [ ] **The accessibility resolution on the toggle** — action label kept, `aria-expanded` dropped,
      `aria-controls` added. The author deferred this; confirm it is the right call for this repo.
- [ ] **The React fix for the render-time ref write** — a derived boolean the callback closes over,
      with the dependency deliberately being the boolean rather than `scrRef` so the imperative
      handle does not re-register on every reference change. The author deferred this; confirm the
      reasoning.
- [ ] **Scope boundary with Power mode.** The feature is Simple-mode only by design, and the hook is
      now disabled outright in Power mode. Confirm the component API left open to the Power team is
      the surface they actually want (`getAdditionalBookIds` + `dimmedReason`).
- [ ] **Obsolete-book handling.** Books the Canon library marks obsolete are now offered whenever a
      project or open resource reports them. Confirm this is desirable product behavior, not just a
      consistency fix.
- [ ] **The `dimmedReason` contract.** It is a localized phrase appended to a localized accessible
      name and shown in a tooltip. Confirm the appended-fragment shape reads correctly in the
      languages the team cares about, since it is not a full sentence template.
- [ ] **Pre-existing red tests on the branch** — the two `EditorRef.removeCharacterMarker` failures.
      Confirm they are tracked and not expected to block this merge.

This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2721)


jolierabideau:
**Solid, well-tested work that does deliver the feature — but two things need your decision as the
ticket author before merge, and one is a genuine defect.**

The core ask works. I verified it end to end in the running app: in Simple mode with a project
(GTP8, 37 books) and a resource (Sanskrit NT) open, the BCV control shows 37 books by default,
"Show more books" expands to 44 with the 7 extras greyed and announced as "not in this project", and
clicking greyed **Hebrews → 1** navigated the whole window to Hebrews 1:1 with the resource panels
rendering it. `getActiveBookIds` is genuinely no longer the ceiling.

**Blocking-ish, needs your call:**
1. **AC 5 is not met** — a new dim style was invented instead of reusing the range picker's. Your
   ticket called this out as a hard requirement.
2. **AC 1 has an undeclared deviation** — search reaches non-project books even in the default
   collapsed state.

**Genuine defect:** the `aria-controls` on the new toggle points at an id that never reaches the DOM.

---

## Runtime verification (what I actually did)

Built and ran the branch on macOS; the build left the working tree **clean**, which independently
confirms the committed `dist/` artifacts match their sources.

| # | Test | Result |
|---|---|---|
| 1 | Power mode, project + resource open | No toggle, 49 books = project only. **Correct** — hook is deliberately disabled in Power mode |
| 2 | Simple mode, GTP8 active + Sanskrit NT open | 37 books collapsed, **"Show more books" toggle present** |
| 3 | Press "Show more books" | 37 → **44 books**; **7 dimmed** (`opacity: 0.7`), label flips to "Show project books only" |
| 4 | Dimmed row accessible name | `"Hebrews (HEB), not in this project"`, `aria-disabled="false"` |
| 5 | Hover a dimmed row | Tooltip "not in this project" appears |
| 6 | **Click greyed Hebrews → chapter 1** | **Navigated to Hebrews 1:1.** Resource panels render Hebrews; editor shows the existing "This book is not part of this project" empty state. **AC 4 met end-to-end** |
| 7 | Reopen picker while on Hebrews | Opens **pre-expanded** (44, "Show project books only") — the deliberate seed |
| 8 | Collapse while on Hebrews | Hebrews **disappears** from the list (37, absent) — picker no longer contains its own current book |
| 9 | Navigate back to Genesis, reopen | Opens **collapsed** (37, "Show more books") — AC 1 default holds for the normal case |
| 10 | Type "Hebrews" in the **collapsed** state | Jumps straight to **HEB 1:1** + Hebrews' chapter grid — a non-project book reached with the toggle never pressed |
| 11 | Arrow-key onto a dimmed row | `opacity: 1`, **no tooltip**. No visual signal at all that the book is outside the project |

Also confirmed: only one BCV trigger exists in Simple mode; the toggle correctly does **not** render
when there is nothing extra to offer (single project open), and does not render when the active
project is the superset (WEB active with 80 books, TPD open → no toggle).

Test 11 and test 6 are the two most important rows: the feature genuinely works, and the greying
genuinely vanishes for keyboard users.

---

## Acceptance criteria

| AC | Requirement | Verdict |
|---|---|---|
| 1 | Default = project books only, no baseline change | **Partially met** — 2 deviations (below) |
| 2 | "Show all books" control expands the list | **Met** (label shipped as "Show more books") |
| 3 | Non-project books appear greyed in expanded list | **Met** — greyed + tooltip + accessible name |
| 4 | Selecting a greyed book navigates | **Met** — verified live (test 6) |
| 5 | Greyed styling reused from existing range picker, no new style | **NOT MET** |
| 6 | NTH-5b served by the same implementation | **Partially met** — Simple mode only; commentaries unverified |

---

## Findings

### Critical

None. No correctness bug that breaks the feature, no security issue, no data-loss path.

### Important

- [ ] **AC 5 violated: a new greyed style was invented rather than reusing the range picker's.**
      `book-item.component.tsx:149` dims with `tw:opacity-70 tw:data-selected:opacity-100`. The
      existing dimmed-but-selectable treatment — the one your ticket points at — lives in
      `numbered-item-grid.component.tsx:69` as `tw:bg-muted/50 tw:text-muted-foreground/50`, is
      reached from the range picker via `chapter-grid.component.tsx:44` / `verse-grid.component.tsx:51`,
      and is **untouched by this PR**. `grep opacity-70` finds it nowhere else in the repo, so this is
      a third dim vocabulary alongside `opacity-40` (grid disabled), `opacity-50` (BookItem disabled),
      and the muted-token pair. Two independent analysis passes flagged this. Inside one popover,
      chapter/verse cells now grey via muted tokens while book rows grey via raw opacity.
      The author documented a real rationale (`book-item.component.tsx:141-148`): the book-id span
      already carries `text-muted-foreground`, so tinting text alone half-dims the row. That is a
      legitimate problem for a two-span row versus a single-digit cell — but the resolution is
      **your** call, not a unilateral one. Note the repo has already learned this lesson once:
      `book-grid.component.tsx:135,143` records that an `opacity-50`-only treatment was "nearly
      indistinguishable" and was replaced with `tw:bg-muted/40 tw:text-muted-foreground/70`.
      → Either amend the AC with the two-span rationale on record, or unify (e.g. a shared dimmed
      token both `NumberedItemGrid` and `BookItem` consume) so a third dim level doesn't accrete.

- [ ] **The greying disappears exactly when the user is on the row — keyboard users get no signal.**
      Verified live (test 11): `tw:data-selected:opacity-100` removes the dim from the highlighted
      row, and the reason is only reachable via a hover tooltip. cmdk items are never DOM-focused
      (the input keeps focus; highlighting is `data-selected`), so Radix's focus trigger never fires.
      Arrowing onto "James (JAS), not in this project" renders it at **full contrast with no
      tooltip**. Screen-reader users get the reason via `aria-label`; sighted keyboard users get
      nothing. Fixing the finding above with the muted-token approach on the *text* would also fix
      this, since it survives `data-selected`.

- [ ] **`aria-controls` on the new toggle is dead and dangling.** `book-chapter-control.component.tsx:1285`
      sets `aria-controls={bookListId}`, but `CommandList` forwards props to cmdk's `Command.List`,
      which builds its element as `createElement(Primitive.div, {ref, ...userProps, ..., id: listId})`
      — `id` is set **after** the user-prop spread, so `id={bookListId}` is overridden by cmdk's own
      id. Net effect: `useId()`/`bookListId` are dead code and the button carries an `aria-controls`
      that resolves to nothing (invalid ARIA). No test covers the linkage, which is why CI is green.
      Separately, the toggle can render while `isCommandListHidden` is `true` (quick-nav pressed,
      `viewMode` still `'books'`), pointing at an unmounted id and offering to expand a list that
      isn't on screen. → Drop `aria-controls`, or put the id on a wrapper you own and gate the toggle
      on `!isCommandListHidden`.

- [ ] **AC 1 deviation: search spans the widened list even while collapsed.** `filteredBooksByType`
      and `topMatch` use `reachableBooks` unconditionally (`book-chapter-control.component.tsx:190-193,
      206-208`; previously `availableBooks`). Verified live (test 10): typing "Hebrews" in the default
      collapsed state jumps straight to HEB 1:1. Matches are greyed and labelled, so it's defensible
      — arguably desirable — but it is an expansion of the *default* state that AC 1's "no change from
      today's baseline" does not authorize and the UX meeting didn't cover. Needs your explicit
      confirmation rather than shipping silently.

- [ ] **Baseline-project semantics are inverted for the very views this PR teaches to publish.**
      The "project books" baseline is `useProjectSetting(resolvedWebView?.definition.projectId, …)`
      (`platform-bible-toolbar.tsx:144`). For resource panels and the Scripture Text Grid, that
      definition `projectId` is the **container**, not what's on screen — the code says so itself
      (`resource-text-panel.web-view.tsx:361`, `model-text-panel.web-view.tsx`). So when the
      navigation target is a resource panel, the books of the resource actually being read are
      treated as *additional*: hidden by default and labelled "not in this project" when shown, while
      the container project's books are the unqualified list. Please confirm this is deliberate, or
      derive the baseline from the declared project when the target publishes one.

- [ ] **Localized fragments are concatenated to build user-facing text.**
      `book-item.component.tsx:136` builds `aria-label={`${baseAriaLabel}, ${dimmedReason}`}`, where
      `dimmedReason` is the standalone localized fragment "not in this project" / "no está en este
      proyecto". `.context/standards/Localization-Guide.md` is explicit that concatenating around a
      localized string "bakes in an assumption about word order and spacing that not every language
      shares" and that the localized string should place the `{placeholder}` itself. Some languages
      need the qualifier before the name, different punctuation, or a case-inflected book name — a
      translator can't fix any of that here. Two passes flagged this independently. → Make it a full
      template (`"{book} is not in this project"`) resolved with `formatReplacementString`. That also
      fixes the tooltip, which currently renders the bare lowercase fragment with no subject, against
      the sentence-case guideline.

- [ ] **The feature ships with no end-to-end coverage, and the E2E was written then deleted.**
      `8bf9a1bb` added `e2e-tests/tests/isolated/bcv-open-resource-books.spec.ts` plus a simple-mode
      sibling and a shared helper; `faa00940` deleted all 533 lines with a one-line message and no
      stated rationale. `git diff $MERGE_BASE -- e2e-tests/` is now empty. The unit layer is genuinely
      strong, but every test there stubs the PAPI boundary — so the one thing no automated test
      exercises is the mechanism this PR introduces: a container web view writing `navigableProjectIds`
      into real web view state and the toolbar reading it back through web-view events. I verified
      that path by hand today; nothing protects it tomorrow. → Restore one isolated spec, or state in
      the PR why it was dropped (flaky? runtime?) so the trade is accepted deliberately.

- [ ] **Reinvented plumbing that already exists in the module it imports from.**
      `use-open-project-book-ids.hook.ts:90-108` rebuilds `onDidOpenWebView` /
      `onDidUpdateWebView` / `onDidCloseWebView` with `getNetworkEvent(EVENT_NAME_ON_DID_*)`, but
      `web-view.service-host.ts:212-218` already exports those three events ready-made — costing
      three extra imports to reproduce. Beyond that, the whole surrounding block
      (`webViewRefreshCounter` + `refreshOpenWebViews` + three `useEvent`s +
      `getAllOpenWebViewDefinitionsSync()` in a try/catch) is a near-verbatim second copy of
      `use-project-picker-data.hook.ts:129-133, 249-277`; `navigation-target.util.ts:62` and
      `auto-sync-edit-block-driver.ts:125` are two more enumeration sites. → Reuse the exported
      events, and ideally extract the refresh-and-enumerate block into one shared
      `use-open-web-view-definitions.hook.ts` both consumers use.

- [ ] **The en/es parity guard only checks `en.json`.** `book-chapter-control-localization.test.ts`
      reads only English. The repo's own precedent guards both arms for exactly this failure mode —
      `sync-status-button.component.test.tsx` has separate `…in en.json` / `…in es.json` tests over
      the same key list, and `localized-strings.test.ts` spells out why ("Nothing in the build
      enforces en/es parity"). The three new keys are in `es.json` today, but nothing stops the next
      one from shipping English-only. Confirmed the test passes with en-only, so this is a coverage
      gap, not a failure.

### Minor

- [ ] `book-chapter-control.utils.ts`: `groupBooksBySection` re-implements the section classification
      `getSectionForBook` (`scripture-util.ts:610`) already owns — which this same component file
      already imports. The two disagree on the Extra test (`Canon.extraBooks().includes` vs
      `Canon.isExtraMaterial`), so they can drift. Logic was moved, not newly written, but it's now a
      named exported helper — the right moment to fold it onto the shared function.
- [ ] The NUL-joined membership fingerprint (`[...ids].sort().join('
      `use-open-project-book-ids.hook.ts:118-122` and `use-publish-navigable-project-ids.hook.ts:41-46`,
      each with its own four-line copy of the same rationale comment (`find.web-view.tsx:237` is a
      third variant). A one-line helper in `platform-bible-utils` would share both the code and the why.
- [ ] ADR-0029's Decision paragraph calls `platform-bible-utils/experimental` "the one place both core
      and extension web views can import a runtime value from" — not accurate; the stable entry point
      is importable from both sides too, and this PR does exactly that (`UnsubscriberAsyncList`,
      `getErrorMessage`). The real reason for the experimental subpath is that the key is unstable.
      Worth rewording so the ADR doesn't record a constraint that doesn't exist. Everything else about
      the entry checks out — appended, correctly numbered 0028 → 0029, all required fields present.
- [ ] Collapsing the list while the current book is outside the project removes the current book from
      the picker (verified, test 8). The open-expanded seed guards the initial open but not this.
- [ ] `navigable-project-ids.utils.ts` exports two functions where one would do —
      `getNavigableProjectIdsToPublish` is called only by its own wrapper (and the tests).
- [ ] Per-caller readiness expressions (`effectiveResourcesState.status === 'ready' && isCatalogReady`,
      etc.) are untested, and that's exactly where a wrong gate would land — the hook's internal gate
      is well covered, the three call sites aren't. Cheap fix: extract each into the existing
      `resource-panel-readiness.utils.ts` and unit-test there.
- [ ] `book-chapter-control-localization.test.ts` sits in `src/renderer/components/` but tests no
      component there, and imports through the committed `dist` — a key added to source but not
      rebuilt would escape the guard. The failure mode is generic to every `*_STRING_KEYS` export.
- [ ] A `TooltipProvider` is instantiated per dimmed row rather than once for the list.
- [ ] The dim is applied to the `CommandItem` inside the testament-colour border wrapper, so the
      coloured OT/NT stripe stays fully saturated beside a dimmed row.
- [ ] The footer toggle vanishes the moment the user types and returns when search clears, shifting
      the popover height mid-interaction.
- [ ] `platform-scripture-editor.web-view.tsx:194-196` comment says "show-all-books" but the keys are
      `showMoreBooks` / `showProjectBooksOnly` — align so it's greppable.
- [ ] `new UnsubscriberAsyncList('Open resource book ids')` still names "resource" after the hook's
      rename to `useOpenProjectBookIds`; the label shows up in runtime logs.
- [ ] `getAdditionalBookIds` is silently ignored when `getActiveBookIds` is absent (TSDoc says so, but
      a caller passing only the former gets the full canon, undimmed, with no diagnostic).

### Scope observations

- **Power mode is excluded** (`platform-bible-toolbar.tsx:170-182`, asserted by
  `platform-bible-toolbar.test.tsx:820-828`). Deliberate, commented, and it saves real cost — but the
  ticket said "global BCV control" with no mode qualifier, and the toolbar BCV renders in both modes.
  Your call to accept as a scope narrowing.
- **Commentary resources unverified.** Your ticket names commentaries explicitly. A project that
  can't report `platformScripture.booksPresent` contributes nothing (logged at debug and dropped,
  `use-open-project-book-ids.hook.ts:169-175`). Whether commentary/enhanced-resource views expose that
  setting isn't determinable from the diff — worth one manual check.
- **The `navigableProjectIds` convention is roughly half the non-generated diff** and exists to serve
  one consumer. It's justified — the Scripture Text Grid is one web view hosting many projects, so
  without it the grid's resources are invisible and NTH-5b is unmet on the main Simple-mode surface —
  but it's a new inter-extension convention arriving with no lint or test guard on *writers*. This is
  the piece to scrutinize hardest, and the piece the author says he can't explain.
- **Drive-by changes worth naming in the PR body**: `"Select Chapter"` → `"Select chapter"` casing on
  two pre-existing labels; `??` → `||` (a consumer deliberately passing an empty string now gets
  English); and `BookItem`'s `aria-label` switching from the English canon name to the localized name
  — which changes accessible names for the public `SelectBooksPicker`/`SelectBooks` scope selector
  too, outside this PR's stated purpose.
- `scroll-group-navigation.commands.ts:97-105` is comment-only, documenting a **deliberate asymmetry**:
  the picker can reach a resource book but next/previous-book commands stay project-scoped. Not
  covered by any AC — confirm intended.

### Template propagation

No shared regions modified, no extension config, build config, CI, or dependency files changed.

---

## Positive observations

- **Forward-facing comments held across ~45 commits of heavy churn.** A scan of every added line for
  `previously` / `used to` / `no longer` / review-finding ids / in-PR ticket ids / stage tags turns up
  nothing. The only `PT-4346` in the diff is the ADR's `Source:` field — the correct home. Every added
  comment states a constraint or rationale rather than narrating the change. This is genuinely hard to
  sustain and it was sustained.
- **The hidden-view rule is satisfied properly**, not perfunctorily: an explicit `// Hidden case:
  intentionally handled by doing nothing special` at the sync site with correct reasoning (data-driven
  web-view-state write, not a geometry read, so it keeps running under `display:none`). Since there's
  no catch-up mechanism, the rule's test clause doesn't apply. Decidable straight from the diff.
- **The transient-empty hazard is named, gated, and tested from both sides** — including "does not
  publish when only the display order changes" and "replaces a persisted value that is not a list of
  ids". These pin the set-semantics and untrusted-read decisions, not just the happy path.
- **Pure logic extracted and directly tested**: `deriveBookChapterControlBookLists` and
  `getNavigableProjectIdsToPublish` are input-only functions, which is what keeps the component and
  hook tests small. ~150 new tests across four workspaces, all green locally.
- **Dimmed ≠ disabled is kept rigorously distinct**: `dimmedReason` never sets `aria-disabled`, never
  blocks `onSelect`, `disabled` takes precedence, and both behaviours are pinned by tests. The greying
  is not a colour-only signal for assistive tech.
- **`isNavigableProjectIds` is used on every read path** rather than trusting persisted state, with
  unit tests including the `null` case. Subscription teardown in `useOpenProjectBookIds` is correct
  against real `UnsubscriberAsyncList` semantics — late unsubscribers land on a sealed list, and the
  `disposed` flag correctly guards only the `setState` calls.
- **The `isEnabled` gate** means Power mode opens no data providers and no subscriptions at all,
  rather than computing and discarding.
- **`dist/` artifacts are reproducible** — a full build on my machine left the tree clean.
- ADR-0029 records the convention with its rejected alternatives and the stale-persisted-state
  consequence, so the next multi-project view can find it.
- Storybook's `WithBooksFromOpenResources` exercises the whole toggle round trip with a `play`
  function, asserting the accessible name and the absence of `aria-disabled` rather than just
  rendering.
- Adding the previously-missing `selectChapter`/`selectVerse` entries fixes keys that were silently
  resolving to code fallbacks; no keyboard handler changed, so no shortcuts-catalog update was owed
  (verified by grep).

---

## CI note

Two failures are red on this branch but **pre-existing**, not caused by it:
`use-remove-character-marker.hook.ts(91,16): Property 'removeCharacterMarker' does not exist on type
'EditorRef'` (typecheck) and the matching `editor-character-marker-contract.test.ts` failure. Neither
file is in the PR's 47 changed files, so this is pre-existing by construction. Worth confirming
they're tracked before merge, since they'll be red in CI regardless.

---

*AI-assisted review (Claude Opus 5): four parallel static analysis passes plus an acceptance-criteria audit, with all runtime claims verified by hand against the built branch.*


[PT-4346]: https://paratextstudio.atlassian.net/browse/PT-4346?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

jolierabideau (changes requested):
> @jolierabideau partially reviewed 47 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2721)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

mattgetgen:
@jolierabideau I made some fixes based on your review.

jolierabideau (approved):
Looks good Matt, thanks!

> @jolierabideau partially reviewed 48 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2721#-:-P03sl8Y9Gd_yy1Dl4Ji:b3w970e)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #12, opened 2023-01-31
Fix ReadMe icon

irahopkinson:
- fix white icon on white background

PR #2722, opened 2026-08-25
PT-2626: grapheme string util perf work

mattgetgen:
# Code Review Summary

**Branch**: pt-2626-grapheme-string-util

**Base**: origin/main

**Date**: 2026-08-26

**Files changed**: 21 (8 hand-written source/doc files + 13 regenerated files under `lib/platform-bible-utils/dist/`)

## Overview

`lib/platform-bible-utils/src/string-util.ts` exposed ~19 grapheme-aware string functions that
re-segmented their input with `stringz` on **every** call, and a few (`formatReplacementString`,
`lastIndexOf`, `endsWith`) segmented per character, making them quadratic — one
`formatReplacementString` call cost ~48ms on a 1000-character string. This branch introduces a
`GraphemeString` class that segments once in its constructor and reuses that work for every
subsequent operation, then reimplements the `string-util` functions as thin wrappers over it. Every
existing caller gets the speedup with no source change.

The class mirrors the native `String` API one-for-one — including the edge cases around negative,
fractional, `NaN`, and out-of-range arguments — with a single substitution: the unit of indexing and
length is the grapheme cluster rather than the UTF-16 code unit, and a search must begin and end on
cluster boundaries. Because the old implementation was *trying* to be native-compatible and got
several cases wrong, the wrapper swap also fixes six long-standing bugs (enumerated under API
Changes). The decision, its rejected alternatives, and its consequences are recorded as `adr-grapheme-string-native-parity`.

## API Changes

- **Added** `GraphemeString` export (`lib/platform-bible-utils/src/grapheme-string.ts`), exported
  from `index.ts`. Public surface: constructor, `length`, `toString`, `toArray`, `at`, `charAt`,
  `codePointAt`, `slice`, `substring`, `padStart`, `padEnd`, `indexOf`, `lastIndexOf`, `includes`,
  `startsWith`, `endsWith`, `split`, `formatReplacement`, `formatReplacementToArray`.
- **No exports removed** from the package entry point. The `string-util` export list in `index.ts` is
  unchanged.
- **Removed** the module-level `testingStringUtils` export from `string-util.ts` (replaced by
  `testingGraphemeStringUtils`). Not package-public — `package.json` has `"files": ["dist"]` and an
  exports map with no wildcard, so `src` was never deep-importable.
- **Reimplemented as thin wrappers** over `GraphemeString` (same signatures): `at`, `charAt`,
  `codePointAt`, `endsWith`, `includes`, `indexOf`, `lastIndexOf`, `stringLength`, `padStart`,
  `padEnd`, `slice`, `split`, `startsWith`, `substring`, `toArray`, `formatReplacementString`,
  `formatReplacementStringToArray`, `isLocalizeKey`.
- **Signature relaxations** (non-breaking, defaulted param → optional param): `endsWith`,
  `includes`, `indexOf`, `padEnd`, `padStart`, `substring`.
- **Unchanged**: `normalize` and `ordinalCompare` remain direct native calls — neither reads the
  string as characters, so wrapping them would add construction cost for no benefit.
- **`lib/papi-dts/papi.d.ts`**: no change needed and none made. It imports `platform-bible-utils` as
  an external module rather than inlining its types, so `GraphemeString` reaches extension developers
  via the package itself.
- **`lib/platform-bible-react/`, extension `.d.ts` files**: unchanged.
- **File renamed**: `grapheme-string-util.ts` → `grapheme-string.ts` (tracked as a git rename), to
  match the package convention where `-util` marks modules of free functions and class modules are
  named for the class.

### Behavior changes to already-exported functions — READ THIS

The full, current list lives in **`lib/platform-bible-utils/README.md` → Behavior Changes**, and the
reasoning behind each decision in `.context/standards/Architecture-Decisions.md` under
`adr-grapheme-string-native-parity` and `adr-grapheme-segmenter-unicode-segmenter`. It is stated
there rather than here on purpose: a PR description stops being visible once this merges, and
downstream extension authors need a copy that survives. Earlier revisions of this description kept
its own table, which drifted out of step with the ADR.

Four groups, summarized:

1. **Grapheme segmentation is now UAX #29 conformant** (`stringz` → `unicode-segmenter`). This is
   the largest change and the only one that alters results for ordinary text rather than edge-case
   arguments: counts, indexes, slices and searches over pointed Hebrew, vocalized Arabic and Syriac,
   Indic conjuncts, Thai, and decomposed Hangul were wrong and are now right. `stringz` scored 65.8%
   on Unicode's own `GraphemeBreakTest.txt`. Emoji and Latin were already correct and are unchanged.
2. **Seven bug fixes** where the old implementation intended native parity and missed it —
   `endsWith`'s position, a string separator compiled as a regex, `slice(s, 1, 0)`, multi-grapheme
   pad overshoot, `lastIndexOf('')`, `split` throwing on a large limit, and an `indexOf` infinite
   loop.
3. **Ten behavior changes for inputs that were always valid**, adopting native semantics — the
   `split` limit, regex capture-group interleaving, honored regex flags, `substring`'s backwards
   range, `endsWith(s, '')`, and fractional/`NaN` index handling.
4. **Two deliberate divergences from native** — the padding ceiling, and `split` yielding `''` rather
   than `undefined` for a non-participating capture group.

**Two things to scrutinize specifically.**

*The `split` limit change* is the only one that loses text silently, with no compile-time signal:

```
split('Look𐐷At🦄This𐐷Thing', '𐐷', 2)
  before: ['Look', 'At🦄This𐐷Thing']
  after:  ['Look', 'At🦄This']
```

No in-repo caller passes a limit, so the exposure is entirely external extension authors.

*Two consequences of segmentation conformance*, both of which can change working code:
`\r\n` is now a single cluster, so `split(text, '\n')` will not break Windows-style lines apart
(use `/\r?\n/`); and a zero-width joiner now attaches to the character before it, so a string
ending in ZWJ + space ends in a whitespace-only cluster — which makes
`areUsjContentsEqualExceptWhitespace` trim that space where it previously kept it.

### One further user-visible change

`getLocalizedIdFromBookNumber` (`scripture/scripture-util.ts`) split Chinese book names on
`'\xff08'`, a three-character string (`"ÿ08"`), not the intended ideographic left parenthesis `'（'`.
The split therefore never matched. The escape is corrected here, so **Chinese book names containing
`（` now truncate at the parenthesis, as the code always intended**. The bug predates this branch.

Measured against the shipped `zh-hans.json` and `zh-hant.json`, this takes truncation collisions from
3 to 7 per locale: `1MA`/`1MQ` → `玛加伯上`, `2MA`/`2MQ`, `3MA`/`3MQ`, and `DNT`/`DAG` → `但以理书`
now render identically. That is PT9's behavior, which is the bar this function is held to, but it is
a real display consequence of a fix riding along in a performance PR and worth a decision rather
than a footnote.

- [ ] ~~`split(string, separator, splitLimit)` changed behavior in a data-losing direction: the
  unsplit remainder is now discarded rather than kept as a final element, and `splitLimit` of `0` and
  negative values flipped meaning. Silent truncation with no compile-time signal. No in-repo caller
  passes a limit; exposure is external extension repos.~~ _(Author: the original implementation's
  intent was feature-parity with the native calls, so this — like the other five — is a bug fix, not
  a breaking change. Accepted deliberately, with the requirement that all six be called out
  explicitly in the PR description rather than left only in `adr-grapheme-string-native-parity`, which downstream extension
  authors will not read.)_

[Author response: ratified the change knowingly rather than inheriting it from the wrapper swap, and
reframed all six behavior changes as bug fixes toward the module's original stated intent. The
reviewer should still weigh the `split` limit case separately, because it is the only one whose
failure mode is silent text loss.]

- [ ] ~~Five further behavior changes to already-exported functions (`endsWith` position handling,
  string-separator regex compilation, `slice(1, 0)`, multi-grapheme pad overshoot, `lastIndexOf('')`)
  are documented only in `adr-grapheme-string-native-parity`, an internal document.~~ _(Author: same disposition — these are bug
  fixes toward the original intent of native parity, to be enumerated visibly in the PR description.)_
- [x] **`GraphemeString` exposed its internal grapheme array in both directions.** `toArray()`
  returned `this.graphemes` by reference (`gs.toArray().push('x')` silently corrupted the instance:
  `length` became 4 while `str` stayed `'abc'`), and the public constructor's second parameter
  accepted an unvalidated precomputed array (`new GraphemeString('abc', ['x','y'])` produced an
  instance whose `length`, `toString()`, and `indexOf` all disagreed). _(fixed during review:
  `toArray()` now returns `[...this.graphemes]`, matching the `SortedSet.toArray()` convention
  already used in this package, with a test pinning it. The constructor is now single-argument; the
  fast path moved to a `private static fromSegmented()` and all five internal call sites were
  rerouted through it. `new GraphemeString('abc', ['x','y'])` is now `TS2554`. Consequence: `str` and
  `graphemes` are no longer `readonly`, since the factory assigns them — both are documented as
  immutable-after-construction, and the `graphemes.join('') === str` invariant is now stated on the
  field itself.)_
- [x] **The new comment in `scripture-util.ts` asserted something false** — it claimed "both
  separators are single characters" while the second was `'\xff08'`, a three-character string. The
  escape bug was pre-existing, but the comment certified the line as reviewed. _(fixed during review:
  corrected to `'（'`, which makes the comment true and activates a split that had never fired.
  Added two tests — a hyphenated and a parenthesized Chinese book name — and verified the parenthesis
  test fails against the old escape with exactly the expected symptom, `'启示（默示录）'` instead of
  `'启示'`. This changes localized display output; see API Changes.)_
- [x] **The trim-loop comment overstated the optimization.** It claimed the loop "costs one pass over
  the text", but `derive` performs a full `this.graphemes.slice()` array copy per iteration and the
  child's `offsetsCache` starts empty, making it O(k·n). _(fixed during review: replaced the
  repeated-slice loop with a `trimEndOfGraphemes()` helper that scans for the cut point and slices
  once — genuinely O(n), so the comment is now true. Verified behaviorally identical to the old loop
  across 4,019 inputs including combining marks, ZWJ clusters, ideographic space and ZWSP, and
  measured 117× faster on a 10k-character string with 2k trailing spaces (30.34 → 0.26 ms/call).
  `b` is now segmented lazily so it is skipped when `a` ends in whitespace or when either
  block-marker guard returns first.)_
- [x] **No `@param`/`@returns` TSDoc on any public `GraphemeString` member**, despite
  `lib/platform-bible-utils/` being named API surface in `Code-Style-Guide.md`, and despite sibling
  classes (`document-combiner.ts`, `sorted-set.ts`) complying. Aggravated by `string-util.ts` now
  steering readers toward the less-documented class. _(fixed during review: all 17 public members plus
  the constructor now carry `@param`/`@returns` with sentinels stated explicitly — `-1`, `undefined`,
  `''`, `undefined` entries — with the existing prose preserved as the description and `@example`
  added to the class.)_
- [x] **`@example` blocks were dropped** from `formatReplacementStringToArray` and
  `formatReplacementString` in the rewrite, including the TSX example showing a React element as a
  replacer. _(fixed during review: examples added to `GraphemeString.formatReplacementToArray` and
  `formatReplacement`, which is where the style analysis said they could legitimately live.)_
- [x] **Two untested branches, both unreachable by the native-parity harness by construction.** The
  grapheme-straddle skip in `splitOnRegExp` (the one place native cannot be the oracle, since the
  whole point is diverging from it), and `split`'s non-participating capture group — the sole
  justification for the `(GraphemeString | undefined)[]` return type and the eslint-disabled cast in
  `string-util.ts`. _(fixed during review: added `/(a)|(1)/` to the harness separator matrix — verified
  it produces `undefined` entries on 7 of 8 fixture strings while matching native exactly — and added
  a straddle test asserting `'a👨‍👩‍👧‍👦b'.split(/\u{1F468}/u)` yields one part where native yields two.)_

[Author response: requested fixes for every item above. On the encapsulation finding the author
specifically asked for a returned copy rather than a `readonly string[]` type, reasoning that the
compile-time guard alone would not stop a JS caller — the stronger choice, and it matches the
package's existing convention.]

### Minor — Consider

- [x] `clampIndex` returned `-0` for positions in `(-1, 0)`, so `new GraphemeString('').indexOf('', -0.5)`
  gave `-0` where native gives `+0` — the only divergence two independent fuzz runs found.
  _(fixed during review: `if (integer <= 0) return 0;`. Also closed the oracle gap that let it
  through: the search tests used `RANGE_INDEXES`, whose only negative fraction was `-1.5`, which
  truncates to `-1` and takes the correct branch. Added `-0.5` and **verified the harness now catches
  the bug** — reverting the fix fails `indexOf with a position` and `lastIndexOf with a position`.)_
- [x] `padStart`/`padEnd` degraded badly on a huge target instead of failing fast.
  _(fixed during review: throws `RangeError` above 2²⁰ graphemes. Note the original finding's premise
  was wrong — native does **not** throw at 2²⁶, it succeeds; native's real limit is 2²⁹ − 24. The
  limit was chosen from measurement: 2²⁰ costs ~9ms/~9MB, 2²² ~35ms/~35MB, 2²⁴ ~173ms/~130MB, and
  native's own ceiling is unreachable — it exhausts the V8 heap. Recorded in `adr-grapheme-string-native-parity` as the class's
  one deliberate divergence from native.)_
- [x] `split` returned `(GraphemeString | undefined)[]`, forcing every caller to handle `undefined`
  though only a regex capture group can produce one. _(fixed during review: overloads added, so a
  string, `GraphemeString`, or omitted separator returns `GraphemeString[]`. Verified at the type
  level and with a repo-wide typecheck. Cost: a caller holding a `string | RegExp` union must now
  narrow — the parity harness was updated accordingly.)_
- [x] Needle type asymmetry — search methods accepted `string | GraphemeString` but `split`'s
  separator did not. _(fixed during review: `split` now accepts `string | GraphemeString` via the
  existing `rawNeedle` helper.)_
- [x] `indexOf` and `lastIndexOf` duplicated the same scan-then-validate-boundaries loop — the
  subtlest logic in the class, maintained in two places. _(fixed during review: collapsed into a
  private `searchOnBoundaries(needle, startOffset, direction)`. Differentially verified against
  native across 2,268 calls using `Object.is` — identical — with cluster-interior needles still
  correctly rejected.)_
- [x] `formatReplacement` called `buildReplacementParts` directly although its own doc said it was
  "`formatReplacementToArray` with every part coerced to a string and joined". _(fixed during review:
  now delegates to `this.formatReplacementToArray(replacers)`. Note the author's instruction was
  "fix the comment drift", which was read as making the code match the doc rather than the reverse.)_
- [x] `nextIntermediateStartIndex = closeCurlyBraceIndex + 1;` was a no-op — `addToContents` already
  set the identical value. _(fixed during review: removed, after confirming from `addToContents` that
  it sets `newContentIndex + newContentLength` = the same value.)_
- [x] `replacers: { [key: string | number]: string | unknown } | object` — `string | unknown`
  collapses to `unknown`. _(fixed during review: now `unknown`, matching the shape its sibling
  `formatReplacementToArray` already used.)_
- [x] `gs` was an undecipherable two-letter abbreviation. _(fixed during review: renamed to
  `graphemeString` — 15 occurrences in source, 64 in tests.)_
- [x] `// Documented exception to "strict GraphemeString args"` cited a convention that appeared
  nowhere else in the repo. _(fixed during review: framing dropped, rationale kept.)_
- [x] Filename `grapheme-string-util.ts` did not match the package convention where `-util` marks
  modules of free functions and class modules are named for the class. _(fixed during review: renamed
  to `grapheme-string.ts` via `git mv` so history follows; all four import sites and the ADR
  reference updated; no residual references outside `dist/`.)_
- [ ] ~~`splitOnRegExp` reaches five nesting levels with the `result.length === limit` early return
  duplicated inside.~~ _(Author chose to skip. Concurring rationale: the function faithfully mirrors
  `RegExp.prototype[@@split]`, and the API analysis probed it hard — zero-width matches,
  non-participating capture groups, end anchors, pre-set `g` and `y` flags, and cluster-straddling
  separators — without breaking it. Restructuring working spec-mirroring code carries more risk than
  the readability buys.)_
- [x] The split-limit test used `\u{10437}` escapes while the rest of the file, including the fixture
  it draws from, uses literals. _(fixed during review: now uses `'𐐷'` and `'At🦄This'`.)_
- [x] No blank line before the `## adr-grapheme-string-native-parity` heading. _(fixed during review.)_
- [x] `adr-grapheme-string-native-parity`'s `Source:` cited `~/repos/test/platform-bible-utils-perf/`, a path that exists only on
  the author's machine. _(fixed during review: rewritten to describe what the investigation did,
  attribute it to Matthew Getgen, state plainly that the harness was scratch work deliberately not
  checked in, and inline the durable headline numbers — including the 30 → 0.26 ms trim figure this
  review produced.)_
- [x] ~~The checked-in `dist/` was regenerated and cannot be meaningfully diffed by a reviewer.~~
  _(Resolved by verification, not by a change: the API analysis re-ran `npm run build` and
  `git status --porcelain` came back empty, so `dist/` is a faithful regeneration with no hand-edits.
  It has been rebuilt after every in-review change.)_

[Author response: requested fixes for all minor findings except the `splitOnRegExp` restructure, which
was deliberately skipped. On the padding limit the author overrode the initial recommendation to leave
native's ceiling in place, asking for "sane defaults that don't cause people to make bad decisions" —
the resulting limit is documented in both the code and the ADR.]

#### Shared Regions Modified

None. No `#region shared with <url>` markers exist in any changed file. The `#region` markers in
`grapheme-string.test.ts` are plain editor-folding sections.

#### Extension Config Changes

**Dependency swap.** `stringz` is removed and `unicode-segmenter` added, in
`lib/platform-bible-utils/package.json` and the lockfile. `stringz` scored 65.8% on Unicode's
`GraphemeBreakTest.txt` and could not count pointed Hebrew; see
`adr-grapheme-segmenter-unicode-segmenter` for the evaluation. Two `cspell.json` files change with
it, one of them under `extensions/`. `lib/platform-bible-utils/vite.config.ts` changes so an import
through a package subpath stays external rather than being bundled.

No `tsconfig*.json`, `.eslintrc*`, `webpack.config.*`, `.erb/configs/`, or `.github/workflows/`
files were touched, and no extension source changed — the only file under `extensions/` is its
cspell word list.

(An earlier revision of this section said none of this happened and that `stringz` remained a
dependency. That was written before the segmenter was replaced.)

## Positive Observations

- **The native-parity harness is the strongest thing in this branch.** For ASCII input one grapheme
  is exactly one UTF-16 code unit, so `String.prototype` becomes a *correct oracle*, turning the
  entire edge-case surface (negative, fractional, `NaN`, `±Infinity`, out-of-range; `ToUint32` split
  limits; regex separators with capture groups and zero-width matches) into a machine-checked cross
  product instead of hand-written expectations. Collecting results into keyed records and comparing
  with one `expect` means a divergence names every failing call rather than stopping at the first.
  Two independent differential fuzz runs — thousands of calls each, over fixture sets the harness
  does not use — found exactly one microscopic divergence between them. That is strong evidence the
  harness is doing real work.
- **The grapheme-specific tests justify the class in human terms.** The `padEnd(5, OFFICER)` case,
  showing that native emits a stripped-down emoji with the wrong skin tone and gender, is a
  memorable and user-meaningful argument for why any of this matters.
- **`adr-grapheme-string-native-parity` is an unusually good review artifact.** It names the rejected alternative that an
  earlier iteration of this same work had chosen, explains why it was rejected, enumerates the
  behavior changes the wrapper swap causes, and explicitly identifies the one data-losing change and
  the residual external exposure. It is the reason the Critical finding is a "ratify this" rather
  than a "you missed this".
- **The purpose is met and measurable.** `formatReplacementString` runs at ~0.085 ms/call on a
  1200-character template, against the ~48 ms/call on a 1000-character string that `adr-grapheme-string-native-parity` records
  for the old implementation. The USJ trim path improved a further 117× during this review.
- **A 20,000-iteration differential of `formatReplacementString`/`ToArray` against the merge-base
  implementation**, over an alphabet seeded with `{`, `}`, `\`, ZWJ emoji and combining marks, found
  **zero differences**. The most widely-called function in this module (60+ call sites) is a verified
  drop-in replacement.
- **`dist/` is genuinely reproducible** — `npm run build` yields a byte-identical tree, confirmed
  independently and re-verified after every in-review change.
- **No `papi.d.ts` churn, correctly** — it imports `platform-bible-utils` as a module rather than
  inlining declarations, so no regeneration was needed and none was made.
- **Every ESLint suppression carries a real justification** that explains why the alternative is
  worse (the `>>> 0` ToUint32 idiom, the `split` return cast mirroring native's own type declaration,
  the `keyof typeof` narrowing) — matching `.claude/rules/code-quality/eslint-disable-discipline.md`
  rather than being reflexive.
- **Comments are forward-facing throughout.** The `PERF:` notes explain durable constraints, and the
  native-parity gotchas ("this is why `'abcab'.indexOf('a', -2)` is 0") are exactly what a maintainer
  needs. No ticket IDs, review-finding references, or change narration leaked into the source, and no
  "candidate A/B" residue survived from the benchmarking commits.
- **`normalize` and `ordinalCompare` staying off the class**, with the reasons inlined at their
  definitions, is the right call and is documented where a future maintainer will actually look.
- **`isLocalizeKey` practices what the module preaches** — it constructs one `GraphemeString` for its
  two operations, with a comment saying why.
- **Test-only exports were retired cleanly.** `testingStringUtils` disappeared with its tests moved
  rather than dropped, and two `padStart`/`padEnd` tests that had been commented out pending an
  upstream `stringz` fix are now live assertions.

## Interview Notes

**Stated purpose (author's words):** improve the performance of `string-util` by memoizing the
`stringz` indexing work once and doing all operations on top of it. The previous implementation did
not, and had poor performance — sometimes quadratic. It also had a few bugs where it did not match
the native `String` implementation.

**Key decisions the author made during review:**

- **Framing of the behavior changes.** Initially presented with a choice between calling the `split`
  limit change a breaking change, preserving the old remainder behavior, or adding an opt-in
  parameter, the author chose to accept native parity and call it out explicitly. On reflection the
  author went further, reframing *all six* changes as bug fixes on the grounds that the original
  module's intent was native parity — so failing to match native was always the defect. This is a
  defensible and coherent position; the reviewer should confirm they agree, since it determines how
  the change is communicated to downstream extension authors.
- **Encapsulation.** The author rejected the suggested `readonly string[]` return type in favor of an
  actual defensive copy, reasoning that a compile-time guard would not stop a caller who casts or
  works in plain JS. This is the stronger choice and matches `SortedSet`'s existing behavior.
- **The `\xff08` escape.** Offered three options — reword the comment only, fix the escape too, or
  reword plus a TODO — the author chose to fix the escape despite it expanding scope into localized
  display output. The reviewer should confirm they are comfortable with a Chinese book-name display
  change riding along in a performance PR.
- **The padding limit.** The author overrode a recommendation to preserve exact native parity,
  choosing a much lower practical ceiling on the grounds that "we want sane defaults that don't cause
  people to make bad decisions". The trade — a documented divergence from native in a class whose
  selling point is not diverging — was made knowingly and is recorded in `adr-grapheme-string-native-parity`.
- **Guard reordering, declined by the reviewer's own analysis.** A more aggressive optimization
  (moving the block-marker guards above all segmentation) was identified and *not* taken, because
  `isAtEndOfBlockMarker` can throw on malformed USJ and reordering would change when that throw
  fires. Flagged here so the reviewer knows the option was considered and rejected on purpose.

**Not covered — the author did not provide a design explanation.** Asked to walk through the most
complex or non-obvious part of the change in their own words — specifically why the design segments
once into a grapheme array and indexes into it, and why searches delegate the scan to native
`String.indexOf` and then *validate* boundaries rather than walking graphemes directly — the author
replied "nothing else to review" and did not answer. This should not be read as agreement or as
confidence; the core design rationale was simply never articulated in the author's own words during
this review. It is the cleverest and least self-evident part of the branch, and the reviewer should
ask about it directly.

**Process note:** the review was paused partway through the interview at the author's request while
they added a commit (`00d83e18e56`, reusing one `GraphemeString` across the USJ whitespace trim), then
resumed with all four analysis passes re-run against the updated diff. Three of the seven Important
findings came from that newly-added file, which the first pass never saw.

## In-Review Quality Check

All checks pass. Every in-review change was verified individually as it was made, and the following
were run repo-wide at the end:

| Check | Result |
|---|---|
| `npm run typecheck` | exit 0, clean |
| `npm run lint` (`lint:scripts` + `lint:styles`) | exit 0, clean |
| `npx prettier --check` on all changed files | "All matched files use Prettier code style!" |
| `npm test` (repo) | exit 0 — 1,621 tests across 5 suites |
| `lib/platform-bible-utils` suite | 514 passed, 1 skipped (30 files) |

No fixes were required by the final pass — nothing was introduced that needed cleaning up
afterwards. Two lint errors *were* hit and fixed at the moment they were introduced: a
`@typescript-eslint/member-ordering` violation when the private `searchOnBoundaries` helper was first
placed among the public methods, and a `no-restricted-syntax` violation from a `for...of` loop in a
new test.

Test count rose from 509 at the start of the interview to 514, with the additions targeting exactly
the branches that had no coverage.

**One thing the reviewer should know:** `npm test` *does* cover this workspace — an earlier
revision of this description claimed otherwise and was wrong. The root script is
`npm run test:core -- --run && npm run test --workspaces --if-present -- --run`; the first half is
the root `vitest.config.ts`, whose `include` covers only `src/**`, `tools/pt9-css-converter/src/**`
and `.erb/scripts/**`, but the second half runs vitest in every workspace, and
`lib/platform-bible-utils` has `"test": "vitest"`.

The real fragility is the `&&`: if `test:core` fails you never reach the workspace tests at all. The
converse also bites — a flaky suite in another workspace can make a green utils run report red. Both
happened while preparing this branch.

**Unrelated pre-existing failure:** `extensions/src/platform-scripture-editor/src/editor-character-marker-contract.test.ts`
fails on this branch, but against the installed `platform-editor` package's type declarations
(missing `removeCharacterMarker`/`replaceCharacterMarker`/`extendCharacterMarker`). It is unrelated to
these changes.

## Suggested Review Focus

- [ ] **Ratify the "these are bug fixes, not breaking changes" framing.** The author's reasoning is
  that the original module intended native parity, so every divergence was always a defect. Do you
  agree — and is a PR-description callout sufficient reach for downstream extension authors, or does
  this warrant release notes / a CHANGELOG entry for the package?
- [ ] **The `split` limit change specifically.** It is the only behavior change whose failure mode is
  silent text loss with no compile-time signal. Worth a separate decision from the other five.
- [ ] **Ask the author to walk through the core design in their own words** — why segment once into a
  grapheme array and index into it, and why searches delegate the raw scan to native
  `String.indexOf`/`lastIndexOf` and then validate grapheme boundaries, rather than walking graphemes
  directly. This question was put to the author during the review and was not answered, so it is the
  single highest-value agenda item for the meeting. It is also the part of the branch a future
  maintainer is most likely to break.
- [ ] **The Chinese book-name display change.** `getLocalizedIdFromBookNumber` now truncates at `（`
  where it previously never did. Confirm this is the desired display behavior and that it is
  acceptable to land it inside a performance PR — or whether it should be split out.
- [ ] **The padding `RangeError` ceiling (2²⁰).** This is the class's one deliberate divergence from
  native. Confirm the limit is right and that a documented divergence is preferable to native's
  unreachable ceiling.
- [ ] **The `readonly` removal on `str` and `graphemes`.** Required so the private static factory can
  assign them. Both are documented as immutable-after-construction, but the compiler no longer
  enforces it internally — confirm that trade is acceptable versus the encapsulation it bought.
- [ ] **`dist/` is checked in and regenerated here.** Confirm the team is content reviewing a diff
  that includes 13 generated files, and that the content-hashed chunk renames are expected churn.

This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2722)


katherinejensen00 (changes requested):
## Review — PT-2626 GraphemeString: 6 blockers

The design is sound and the speedups are real: measured old-vs-new against the built bundle, `split` ×133, `lastIndexOf` ×62, `endsWith` ×11, `formatReplacementString` 15.85 → 1.09 ms, nothing regressed beyond noise. Also worth crediting since it's not currently listed: the old `stringz`-backed `indexOf` loops forever on `indexOf('a', 'a', -Infinity)`, and **this branch fixes that** — a seventh bug fix.

Below are the six I'd treat as merge-gating. A follow-up comment has the remaining 24 (docs, undocumented behavior changes, perf, test design). Every finding was verified by executing code. Baseline for contrast: 514 tests pass, `eslint` clean, `tsc --noEmit` clean — none of these is caught by existing tooling.

Numbering matches the full list of blocking and non blocking findings, so it's non-contiguous here.

---

### 1. 🔴 Infinite loop that freezes the renderer — `lib/platform-bible-utils/src/grapheme-string.ts:559`

`new GraphemeString('👍🏽').split(/./u)` never returns. Reproduced twice at `timeout 10` (exit 124), and again through the public wrapper `split('👍🏽', /./u)`.

A `u`-flagged regex snaps a `lastIndex` that lands inside a surrogate pair back to the pair's start — verified directly:

```js
const r = /./ug; r.lastIndex = 1;
r.exec('👍🏽').index // → 0, not 1
```

So `searchOffset = match.index + 1` recomputes `1` forever, pinning a CPU core. `/./v` and the family emoji hang identically; the old implementation returned. Reachable from any `u`/`v` separator — which #6b newly makes honored.

**Fix:** `searchOffset = Math.max(match.index + 1, searchOffset + 1)` (also lines 566, 576), plus correcting the comment at line 540 — "scanning globally finds the same leftmost match" is the false premise that makes line 559 look safe.

---

### 2. 🟠 `padStart`/`padEnd` return an instance whose grapheme array isn't its own text — `grapheme-string.ts:325`, `:341`

Reachable with the **default** pad string: `new GraphemeStringms length 3; honest segmentation is 2. On a padded instance,`new GraphemeString('a').padEnd(3,'\u0301').slice(0,2)` returns `'á'` — a slice cutting a cluster in half, which the class doc at line 86 promises never happens.

An invariant fuzz over slice/substring/split/pad gave 191 failures, **100% of them padding**; the other three routes were clean.

Not a regression — merge-base output is byte-identical. What's new is a public class reporting `length 3` where `stringLength()` on the same text says 1.

---

### 3. 🟠 `instanceof RegExp` fails across realms, silently downgrading a regex to a literal string — `grapheme-string.ts:478`

`split('a,b,c', foreignRe)` → `['a,b,c']` where native gives `['a','b','c']`.

This is a **regression**: the old code used `String.prototype.match`, which dispatches via `Symbol.match` and works cross-realm. Electron runs renderer / extension-host / WebView iframes in separate realms and this package is injected into all of them.

**Fix:** `Object.prototype.toString.call(x) === '[object RegExp]'`.

---

### 4. 🟠 `split` can place `undefined` inside a value declares:270`

`split('a1b', /(z)?(\d)/)` → `['a', undefined, '1', 'b']`, typed `string[]`. The old implementation built results from `substring` and could never produce one.

`split(s, re).map(p => p.trim())` type-checks clean and throws

---

### 6. 🟠 Two undocumented regex-separator changes — `string-util.ts:261`

**(a) Capture groups now interleave.** `split('2026-08-27', /(-)/)`: `['2026','08','27']` → `['2026','-','08','-','27']`. Array length and every position shifts under an unchanged `string[]` return type, so `const [y, m, d] = split(...)` silently yields `['2026','-','08']`.

**(b) Non-global regex flags are now honored.** The old code recompiled via `new RegExp(sep,'g')`, discarding every other flag: `split('aXbXc', /x/i)`: no split at all → `['a','b','c']`. This is also how a caller reaches #1.

Neither is in the "READ THIS — all six" table, whose `split` rators.

---

### 11. 🟡 No `[Symbol.iterator]`, so the class is array-like s — `grapheme-string.ts:145`

`Array.from(new GraphemeString('abc'))` → `[undefined, undefined, undefined]`; native `Array.from('abc')` → `['a','b','c']`. `[...gs]` throws `TypeError`.

Listed as gating despite being Medium in isolation, because it bites exactly the migration the module header comment recommends — a caller moving `Array.from(str)` onto a `GraphemeString` gets a silent wrong answer with no error.

**Fix:** `*[Symbol.iterator]() { yield* this.graphemes; }`

Follow-up to the blockers comment. Numbering continues from the same list, grouped by theme so they can be batched. All verified by executing code.

**The systemic point first**, because it's worth more than any individual item below: the "READ THIS — all six" table is the artifact downstream extension authors are told to rely on, and it under-reports. Findings #5, #6a, #6b, #7, #12, #17 (×3) are all behavior changes on already-exported functions that aren't in it, and #26 shows the ADR and the PR description each list a *different* six. The reason none of them surfaced is #27: the parity suite can't see the layer it guards.

---

## Undocumented behavior changes on the published API

**5. 🟠 `substring` now swaps a backwards range** — `string-util.ts:302`
`substring('abcdef',5,2)`: `''` → `'cde'`; `substring('abcdef',5,-1)`: `''` → `'abcde'`. The `-1`-sentinel shape (`substring(s, b, indexOf(...))` where the search misses) used to return empty and now returns text from the wrong end of the string.

**7. 🟠 `endsWith(s, '')` flipped `false` → `true` for every input** — `string-util.ts:85`
Boolean-inverting. A suffix filter where the suffix can be empty — unset config value, cleared search box — now matches **everything** where it previously matched nothing. Different root cause from the documented `endPosition` change.

**12. 🟡 `at`/`charAt`/`codePointAt` changed for fractional and `NaN` indexes** — `string-util.ts:38`
`at(s, -0.5)`: last grapheme → `'L'` (opposite end of the string); `charAt(s, NaN)`: `''` → `'L'`; `codePointAt(s, NaN)`: `undefined` → `76`. The ADR enumerates the shapes it scanned call sites for — "negative index arguments, an `indexOf` result reused as a position, multi-grapheme pads" — and fractional/`NaN` isn't among them. A computed index that can be `NaN` (failed `parseInt`, unset field) now returns element 0 instead of an out-of-bounds sentinel.

**17. 🟡 Three more flips** — `string-util.ts:38`, `:286`, `:117`
`at(s, stringLength(s))`: `''` → `undefined` (so `at(s,i).length` in a `i <= stringLength(s)` loop now throws). `startsWith('a','a',-1)`: `false` → `true`. `indexOf('abc','',-1)`: `-1` → `0`, flipping a not-found branch.

**18. 🟡 The public padding wrappers can now throw and say nothing about it** — `string-util.ts:225`, `:208`
No `@throws` on either. Compounding it: in the published `dist/index.d.ts` the only mention of the ceiling is `{@link MAX_PADDING_LENGTH}` — a dangling link to an **unexported** const — so the limit is unlearnable from the shipped types, and TypeDoc will emit an unresolved-link warning.

---

## Robustness

**8. 🟠 The `RangeError` fires before the empty-`padString` short-circuit** — `grapheme-string.ts:610`
`padStart('abc', 2**21, '')` throws; native and the old code both return `'abc'` (spec `StringPad` returns the string when the filler is empty). Outside the documented rationale, which is entirely about the padding array's memory cost — with an empty filler no array is ever built. **Fix: move line 618 above 610.** One line.

**9. 🟠 `rawNeedle` calls `.toString()` with no nullish guard** — `grapheme-string.ts:678`
`indexOf('anullb', null)` throws `TypeError` where native and the old wrapper return `1`. Affects `indexOf`, `lastIndexOf`, `includes`, `startsWith`, `endsWith`, `split`. TS callers are protected; the untyped JS extension code this package publishes to is not. **Fix:** `String(searchString)`.

**13. 🟡 `buildPadding` passes `padString` to stringz with no `ToString` coercion** — `grapheme-string.ts:617`
`padStart('abc',8,null)`: native `'nullnabc'`, ours throws a bare `Error` reading "A string is expected as input" — naming neither the method nor the argument. Same for `5`, `{}`, `['x']`, `true`. Distinct parameter and distinct fix from #9. **Fix:** `String(padString)`. One line.

**14. 🟡 `offsets()` mutates lazily, so a frozen instance throws order-dependently** — `grapheme-string.ts:599`
Freeze-then-`indexOf` → `TypeError: Cannot assign to read only property 'offsetsCache'`; `indexOf`-then-freeze-then-`indexOf` returns `2` cleanly. Whether freezing breaks the object depends on whether something happened to warm the cache first — a latent, load-order-dependent crash — while lines 113–121 tell callers to treat it as immutable after construction.

**15. 🟡 `replacerKey in replacers` walks the prototype chain** — `grapheme-string.ts:754`
`formatReplacement('x{toString}y', {})` → the native-code source text instead of `'xtoStringy'`. `formatReplacementToArray('{hasOwnProperty}', {})` returns a bare function, which React refuses to render. `formatReplacement('n={length}', ['x','y'])` → `'n=2'`. Pre-existing, but the code was rewritten here. **Fix:** `Object.hasOwn`.

**16. 🟡 `offsetsCache` is enumerable, so *reading* an instance changes its structural identity** — `grapheme-string.ts:125`
`deepEqual(a, new GraphemeString('abc'))` is `true`, then `false` after `a.indexOf('b')`. `Object.keys(a)` is `['str','graphemes','offsetsCache']`; `JSON.stringify` emits the internals at ~9× the text size with no `toJSON`, so an instance crossing a PAPI boundary neither round-trips nor compares.

**24. 🟡 `` `${content}` `` throws on values that should degrade** — `grapheme-string.ts:232`
A `Symbol` replacer → `TypeError: Cannot convert a Symbol value to a string`; an `Object.create(null)` replacer → `TypeError: Cannot convert object to primitive value`. `formatReplacementString` runs over localized templates assembled from caller-supplied values, so one bad value takes down the whole localization call.

---

## Documentation, ADR, and PR description

**10. 🟠 The description's claim that `npm test` doesn't cover this workspace is false**
Root `"test"` is `npm run test:core -- --run && npm run test --workspaces --if-present -- --run`; `lib/*` is a workspace with `"test": "vitest"`. Running that half unfiltered gives **30 files, 514 passed**. The real fragility is different and worth stating instead: `&&` short-circuits, so a flaky `test:core` can mean you never reach the utils suite.

**19. 🟡 The class newly promises UAX #29 semantics that `stringz` doesn't deliver** — `grapheme-string.ts:137` (claim at `:86`)
Measured against `Intl.Segmenter`: CRLF → 2 clusters (should be 1, rule GB3); Devanagari `क्षि` → 4 (1); Hangul `각` → 3 (1); Thai `กำ` → 2 (1). Those are live translation targets, and CRLF is every Windows-authored USFM file. Not asking for the swap — ~10× slower and it would break `split(stack, '\n')` — just that the ADR record the choice and the class doc be qualified.

**20. 🟡 The new justification comment is backwards** — `scripture/scripture-util.ts:310`
It says native "would cut through the middle of a cluster." Native never severs a cluster; the real effect is the opposite — `split('a-\u0301b','-')` returns `['a-\u0301b']` (no split) where native gives `['a','\u0301b']`. A decorated separator stops being a separator.

**21. 🟡 `isAtEndOfBlockMarker` lost its TSDoc** — `scripture/scripture-util.ts:787`
`trimEndOfGraphemes` was spliced between the docblo`ts.getJSDocTags(isAtEndOfBlockMarker)` returns `[]`, and the orphan doesn't attach to the new function either — it's dead, including the non-obvious "we don't need to walk up the ancestors" rationale. **Fix:** move lines 787–795.

**23. 🟡 All three `@example` blocks dropped from the exported format functions** — `string-util.ts:317`
Including the TSX example showing a React element as a replacer — the most non-obvious property of the API and the reason the array variant exists — plus the empty-curly-braces precondition note that a test still pins. They moved to the *class*, but 12+ in-repo files import the free functions, and `lib/platform-bible-utils/` is named API surface in `Code-Style-Guide.md`.

**25. 🟡 Six-plus behavior changes ship with no durable channel** — `index.ts:7`
No CHANGELOG, `version` pinned at `0.0.1`. The PR description is the only thing telling downstream extension authors, and it stops being visible once this merges. The behavior-change table needs a home that survives — package README, or a README pointer to the ADR.

**26. 🟡 The documents consumers are pointed at disagree with each other** — `.context/standards/Architecture-Decisions.md:1606`, `:1578`, `:1614`
- The ADR lists **five** bug fixes, the PR table **six**, and they differ in *content*: the ADR has "`split` throwing on a limit above the match count" (absent from the PR table); the PR table has the `new RegExp` metacharacter bug (absent from the ADR). Each is missing one real change.
- Neither lists #5, #6a, #6b, #7, #12, or #17.
- Line 1578's "Two additions are kept (`normalize(Alternatives bullet. That line is also 150 chars where every neighbour is ≤102.
- Line 1614's "the `-1` sentinel hazard is gone, because negative positions now clamp" is false for `substring`'s **end** argument (#5).
- The PR body cites `ADR-0029` ~10 times; the entry was renamed to the slug `adr-grapheme-string-native-parity`, so the number now returns zero hits.

**28. 🟢 The `\xff08` → `\uff08` correction is unrelated scope with a real user-visible effect** — `scripture/scripture-util.ts:315`
To be clear, **this is not a bug you introduced**. In C#, `\x` consumes up to four hex digits, so PT9's `Split('\xff08')` has always split on `（`; the TS port copied the escape text into a JS string where `\xNN` takes exactly two. Your fix restores PT9 parity.

But it changes a PAPI-exposed result (`papi.d.ts:8836`), and against the **shipped** `zh-hans.json` it takes collisions from 3 → 7 per locale: `1MA`/`1MQ` → `玛加伯上`, `2MA`/`2MQ`, `3MA`/`3MQ`, `DNT`/`DAG` → `但以理书`. Same in `zh-hant`. Worth one sentence in the description. Separately, the two new tests use hand-written fixtures rather than shipped data (`%Book.REV%` really is `启`), so the real-data effect is invisible to the suite.

---

## Performance

**22. 🟡 `split(s, '')` is a 4.2× regression on a perf PR** — `grapheme-string.ts:511`
100 000-char string: old 3.34 ms → new 13.93 ms, allocating 100 000 wrapper objects the caller immediately discards via `.toString()`. (For contrast, the headline wins hold: `formatReplacementString` 15.85 → 1.09 ms.)

---

## Test design

**27. 🟢 The parity harness can't see the layer it guards** — `grapheme-string.test.ts:13`, `:63`, `:300`
Three gaps, each proven:
- Replacing `isBoundary()` with `return true` — deleting the grapheme-boundary rule outright — produces **0 failures across 6,290 comparisons**, because the ASCII-only matrix makes every grapheme index identical to its UTF-16 offset. `offsets()`, `offsetAt`, `graphemeIndexAtOffset`, `isBoundary` and `graphemeSpan` are exercised only as identity functions.
- `nativeParity` always constructs fresh instances, so `fromSegmented`-derived values never enter the matrix. Running the same matrix against derived instances gives 28,836 comparisons: slice/substring/split are clean, padding fails visibly — i.e. it would have caught #2 directly.
- `SEPARATORS` holds no `u`/`v` regex, which is why #1 shipped green.

**Cheap fix:** add a single-code-point astral fixture (`'𐐷𐐸𐐹𐐺𐐻'`), where native stays a valid oracle at exactly 2× scale, and parameterize the harness over an instance factory.

---

## Smaller items

**29. 🟢 API and doc snags** — `grapheme-string.ts:460`, `:178`, `:691`; `grapheme-string.test.ts:300`
The `split` overloads reject a `string | RegExp` union — the declared parameter type of the function this class supersedes (`TS2769`, since TypeScript doesn't union overloads) — which is also why `test.ts:323-326` contains a ternary whose two branches are textually identical. `toArray` is documented "No native equivalent" while its own new wrapper (`string-util.ts:307`) says it "Mirrors spreading a native string"; the wrapper is right, and this is *why* #11 was missed. `split` is never tested with a `GraphemeString` separator despite the overload advertising it. The `escaped === true` branch is unreachable from production, and `testingGraphemeStringUtils` exists solely to keep ~50 lines of tests pointed at it.

**30. 🟢 Odds and ends**
- `grapheme-string.ts:208` — `@returns` says "Never empty", but `''` yields `[]`.
- `grapheme-string.ts:41` — the `-Infinity` guard in `relativeIndex` is dead code (`Math.max(length + -Infinity, 0)` is already 0), sitting beside `clampIndex`'s genuinely load-bearing `-0` guard and making that one read as boilerplate.
- `grapheme-string.ts:572` — capture groups re-segment via `new GraphemeString(capture)`, contradicting line 81's "never re-segment", which the ADR cites as the justification for range methods returning `GraphemeString`.
- `grapheme-string.ts:627` — `offsetAt` materializes the full offsets array even for `offsetAt(0)` (always 0) and `offsetAt(length)` (always `str.length`) — every default-position `startsWith`/`endsWith`, including `isLocalizeKey`, which runs per tab title and per menu item.
- `scripture-util.ts:794` — `trimEndOfGraphemes` does a full array copy when there's nothing to trim (`end === length`, true on roughly half the calls in the function this branch exists to speed up).
- `scripture-util.ts:866` — "two GraphemeStrings are never `===` each other" is false in five places (`padEnd(0)`, `padStart(0)`, `padEnd(5,'')`, `split()[0]`, and the empty-string split).
- `scripture-util.ts:313` — segments the same string three times, against the module's own new guidance.

---

Happy to expand the reasoning on any of these, or drop the ones you disagree with.

> @katherinejensen00 reviewed 21 files and all commit messages, and made 7 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2722#-:-P03gFjA4YC6kBJxQskV:b-r09182)* status: all files reviewed, 6 unresolved discussions (waiting on mattgetgen).

___
*[`lib/platform-bible-utils/src/grapheme-string.ts` line 145 at r1](https://reviewable.io/reviews/paranext/paranext-core/2722#-P03jATiDR4Cly2OuYF-:-P03jATiDR4Cly2OuYF0:b-5yow7a) ([raw file](https://github.com/paranext/paranext-core/blob/5b41b73cb7e5c197a249f9ce58108e08996413a5/lib/platform-bible-utils/src/grapheme-string.ts#L145)):*
> ```XML
>    * @returns Count of grapheme clusters. 0 for the empty string.
>    */
>   get length(): number {
> ```

### 11. 🟡 No `[Symbol.iterator]`, so the class is array-like and `Array.from` silently lies — `grapheme-string.ts:145`

`Array.from(new GraphemeString('abc'))` → `[undefined, undefined, undefined]`; native `Array.from('abc')` → `['a','b','c']`. `[...gs]` throws `TypeError`.

Listed as gating despite being Medium in isolation, because it bites exactly the migration the module header comment recommends — a caller moving `Array.from(str)` onto a `GraphemeString` gets a silent wrong answer with no error.

**Fix:** `*[Symbol.iterator]() { yield* this.graphemes; }`
___
*[`lib/platform-bible-utils/src/grapheme-string.ts` line 325 at r1](https://reviewable.io/reviews/paranext/paranext-core/2722#-P03iQxoEBZ6n1KsmziJ:-P03iQxoEBZ6n1KsmziK:bj0u1jk) ([raw file](https://github.com/paranext/paranext-core/blob/5b41b73cb7e5c197a249f9ce58108e08996413a5/lib/platform-bible-utils/src/grapheme-string.ts#L325)):*
> ```XML
>     const padding = this.buildPadding(targetLength, padString);
>     if (padding.length === 0) return this;
>     return GraphemeString.fromSegmented(
> ```

### 2. 🟠 `padStart`/`padEnd` return an instance whose grapheme array isn't its own text — `grapheme-string.ts:325`, `:341`

Reachable with the **default** pad string: `new GraphemeString('\u0301b').padStart(3)` claims length 3; honest segmentation is 2. On a padded instance, `new GraphemeString('a').padEnd(3,'\u0301').slice(0,2)` returns `'á'` — a slice cutting a cluster in half, which the class doc at line 86 promises never happens.

An invariant fuzz over slice/substring/split/pad gave 191 failures, **100% of them padding**; the other three routes were clean.

Not a regression — merge-base output is byte-identical. What's new is a public class reporting `length 3` where `stringLength()` on the same text says 1.
___
*[`lib/platform-bible-utils/src/grapheme-string.ts` line 478 at r1](https://reviewable.io/reviews/paranext/paranext-core/2722#-P03iX047vR_OcoI0uG_:-P03iX047vR_OcoI0uGa:b-65jse9) ([raw file](https://github.com/paranext/paranext-core/blob/5b41b73cb7e5c197a249f9ce58108e08996413a5/lib/platform-bible-utils/src/grapheme-string.ts#L478)):*
> ```XML
>     if (limit === 0) return [];
>     if (separator === undefined) return [this];
>     if (separator instanceof RegExp) return this.splitOnRegExp(separator, limit);
> ```

### 3. 🟠 `instanceof RegExp` fails across realms, silently downgrading a regex to a literal string — `grapheme-string.ts:478`

`split('a,b,c', foreignRe)` → `['a,b,c']` where native gives `['a','b','c']`.

This is a **regression**: the old code used `String.prototype.match`, which dispatches via `Symbol.match` and works cross-realm. Electron runs renderer / extension-host / WebView iframes in separate realms and this package is injected into all of them.

**Fix:** `Object.prototype.toString.call(x) === '[object RegExp]'`.
___
*[`lib/platform-bible-utils/src/grapheme-string.ts` line 559 at r1](https://reviewable.io/reviews/paranext/paranext-core/2722#-P03gV887cGXsUerkcpJ:-P03gV887cGXsUerkcpK:b-3e1ia3) ([raw file](https://github.com/paranext/paranext-core/blob/5b41b73cb7e5c197a249f9ce58108e08996413a5/lib/platform-bible-utils/src/grapheme-string.ts#L559)):*
> ```XML
>       if (matchStart < 0 || !this.isBoundary(matchEndOffset)) {
>         // Straddles a grapheme boundary, so it cannot split a grapheme-indexed string.
>         searchOffset = match.index + 1;
> ```

### 1. 🔴 Infinite loop that freezes the renderer — `lib/platform-bible-utils/src/grapheme-string.ts:559`

`new GraphemeString('👍🏽').split(/./u)` never returns. Reproduced twice at `timeout 10` (exit 124), and again through the public wrapper `split('👍🏽', /./u)`.

A `u`-flagged regex snaps a `lastIndex` that lands inside a surrogate pair back to the pair's start — verified directly:

```js
const r = /./ug; r.lastIndex = 1;
r.exec('👍🏽').index // → 0, not 1
```

So `searchOffset = match.index + 1` recomputes `1` forever, pinning a CPU core. `/./v` and the family emoji hang identically; the old implementation returned. Reachable from any `u`/`v` separator — which #6b newly makes honored.

**Fix:** `searchOffset = Math.max(match.index + 1, searchOffset + 1)` (also lines 566, 576), plus correcting the comment at line 540 — "scanning globally finds the same leftmost match" is the false premise that makes line 559 look safe.
___
*[`lib/platform-bible-utils/src/string-util.ts` line 261 at r1](https://reviewable.io/reviews/paranext/paranext-core/2722#-P03j41G9BvJirVyS3U7:-P03j41G9BvJirVyS3U8:bt00uz7) ([raw file](https://github.com/paranext/paranext-core/blob/5b41b73cb7e5c197a249f9ce58108e08996413a5/lib/platform-bible-utils/src/string-util.ts#L261)):*
> ```XML
>  * @returns An array of strings, split at each point where separator occurs in the starting string
>  */
> export function split(string: string, separator: string | RegExp, splitLimit?: number): string[] {
> ```

### 6. 🟠 Two undocumented regex-separator changes — `string-util.ts:261`

**(a) Capture groups now interleave.** `split('2026-08-27', /(-)/)`: `['2026','08','27']` → `['2026','-','08','-','27']`. Array length and every position shifts under an unchanged `string[]` return type, so `const [y, m, d] = split(...)` silently yields `['2026','-','08']`.

**(b) Non-global regex flags are now honored.** The old code recompiled via `new RegExp(sep,'g')`, discarding every other flag: `split('aXbXc', /x/i)`: no split at all → `['a','b','c']`. This is also how a caller reaches #1.

Neither is in the "READ THIS — all six" table, whose `split` row covers only *string* separators.
___
*[`lib/platform-bible-utils/src/string-util.ts` line 270 at r1](https://reviewable.io/reviews/paranext/paranext-core/2722#-P03ixxZAu0QAC-V5thW:-P03ixxZAu0QAC-V5thX:bffhqfx) ([raw file](https://github.com/paranext/paranext-core/blob/5b41b73cb7e5c197a249f9ce58108e08996413a5/lib/platform-bible-utils/src/string-util.ts#L270)):*
> ```XML
>   // caller's types for a case only a capturing separator reaches.
>   // eslint-disable-next-line no-type-assertion/no-type-assertion
>   return graphemeString.split(separator, splitLimit).map((part) => part?.toString()) as string[];
> ```

### 4. 🟠 `split` can place `undefined` inside a value declared `string[]` — `string-util.ts:270`

`split('a1b', /(z)?(\d)/)` → `['a', undefined, '1', 'b']`, typed `string[]`. The old implementation built results from `substring` and could never produce one.

`split(s, re).map(p => p.trim())` type-checks clean and throws at runtime.

katherinejensen00 (approved):
Thanks for making those fixes, Matt! Here is some new non blocking feedback. I am approving the PR since it can go ahead with or without the changes. Up to you if you want to do them now, or later in another pr or jira work item, or if you don't think they are worth doing.

New this round — worth a decision

N1 🟠 Khmer book search regresses — book.utils.ts:180. Coarser (correct) clusters mean a prefix ending mid-cluster no longer matches. Against shipped km.json, typing one code point at a time: 10 of 48 Khmer book IDs now have a dead spot, 0 of 48 on main. For ROM (រ៉ូម) and JOB (យ៉ូប) it's the first keystroke — type the first letter, list goes empty. The ADR knows Khmer got coarser but its caller sweep names only logger.utils.ts (which actually uses native split) and areUsjContentsEqualExceptWhitespace; it missed this one.

N2 🟠 string-util.ts:148 — stringLength's TSDoc says it "Overcounts for \r\n, Indic conjuncts, Thai spacing marks, and decomposed Hangul Jamo." Those are the four cases this PR fixes. Contradicts the module note 130 lines up and the README; ships as API-surface docs.

N3 🟠 PR description — "Extension Config Changes: None… No dependencies added or removed — stringz remains a dependency" is false in all four clauses now.

N6 🟡 Perf has a second half. Interleaved A/B vs main's bundleth 9.7×, formatReplacementString 942× — but single-shotindexOf/startsWith/substring/stringLength on chapter-sized strings are 0.54–0.70×, i.e. 1.4–1.9× slower (the wrapper materializes the cluster array). No in-repo call site hits it. The ADR names the exact fix — countGraphemes, 6–8× on stringLength — defers it, and names no ticket.

N7 🟡 \xff08 disclosed only in the PR body. Measured on shippe differently in zh-hans and zh-hant; identical-rendering pairsgo 3 → 7 (1MA/1MQ, 2MA/2MQ, 3MA/3MQ, DNT/DAG). es/fr unaffected. The one end-user-facing change, recorded in the one place this PR argues doesn't survive the merge.

Smaller: N4 toArray TSDoc points at a note saying the opposite · N5 stringz dropped from cspell while this PR adds 15 uses of the word · N8 .trim() untested (deleting it leaves 59/59 green) · N9 restored @examples use \{ where they need \\{ · N10 "the earlier decision here, superseded" has no referent · N11 2**20 costs "~9ms" in code, "~11ms" in the ADR · N12 one backward-facing comment (scripture-util.ts:337) · N13 test nits incl. a vacuous Hebrew indexOf, a self-comparing iterator assertion, and no test pinning that a string separator is literal (the removed ReDoS vector) · N14 unicode-segmenter vendored into dist but devDependencies, so invisible to npm audit --omit=dev.

> @katherinejensen00 reviewed 42 files and all commit messages, made 1 comment, and resolved 6 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2722#-:-P0ULW3_3AlM4T8EO-fZ:b-h8wheu)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

katherinejensen00 (approved):
> @katherinejensen00 partially reviewed 24 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2722)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

katherinejensen00 (approved):
> @katherinejensen00 reviewed 2 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2722)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

Issue #60, opened 2023-01-09
Integrate resource viewer code


GlennPruitt:
As an architect I need to provide an example extension that does something useful.

Develop an extension that appears in the Paranext "Hello World" example.  

Consider using Chris Klapp's Scripture viewer code to display Scripture text in a read only view.

Depends on #44, #59

tjcouch-sil:
The Simple USFM Editor App is a simple editor demo Chris Klapp made. The community-maintained editor based on it is the Xelah editor. The friendsofagape USFM Editor is a Slate-based editor that Autographa made and gave up on. They had a contractor make the editor for them over two years and did not make a lot of progress.

Marc Durdin, Makara Sok, and the Keyman team are running tests on various editor libraries for us:
- Prosemirror: https://github.com/keymanapp/keyman/issues/7902#issuecomment-1350811703
  - Mobile: https://github.com/keymanapp/keyman/issues/7978
- Lexical Text Editor: https://github.com/keymanapp/keyman/issues/7937#issuecomment-1360925872
  - Mobile: https://github.com/keymanapp/keyman/issues/7977
- SlateJS: https://github.com/keymanapp/keyman/issues/7938#issuecomment-1360930262
  - Mobile: https://github.com/keymanapp/keyman/issues/7979
- USX Editor: https://github.com/keymanapp/keyman/issues/8312
  - Mobile: In progress
- Xelah HtmlPerfEditor: https://github.com/keymanapp/keyman/issues/8324
  - Mobile: In progress

The USX Editor seems to be a good baseline for what works in practically just pure HTML contenteditable. There are a few differences between the baseline and tests in these other editors.

tjcouch-sil:
Quick USFM viewer? https://www.npmjs.com/org/xelah

katherinejensen00:
Here is a [document ](https://docs.google.com/document/d/1erjzEfmGhbv3P39g01Q-RsIHSDPMI9A9B8AvSFaHHVk/edit#)that contains some research done into existing scripture utilities. From what I can tell, VerseRef from ScriptureForge is probably what we want to use, but we will want to talk about this as a group.

GlennPruitt:
Katherine,

Thank you, this is very good information.  I appreciate it.

-Glenn

*…………………………………*
[image: Paratext logo]
*Glenn Pruitt*
Paratext Product Manager
***@***.***
***@***.***>

katherinejensen00:
[...]

PR #2723, opened 2026-08-25
Bump @sillsdev/scripture to 2.1.0

irahopkinson:
Bumps `@sillsdev/scripture` from 2.0.5 to 2.1.0.

## What changes

All 12 manifests that declare it — the root, `extensions`, and the 10 extensions under `extensions/src/*` — plus the lockfile. Verifiable in one line:

```bash
git grep -l '"@sillsdev/scripture": "\^2\.1\.0"' -- '*package.json' | wc -l   # 12
git grep -l '"@sillsdev/scripture": "\^2\.0\.5"' -- '*package.json'           # none
```

Only the **root** manifest changes what actually executes. `@sillsdev/scripture` is a webpack `external` for every extension and the host injects the runtime module (`extension.service.ts`, `global-this-web-view.model.ts`), so the other eleven declarations govern types and Vitest resolution. Keeping them in lockstep is deliberate, not load-bearing at runtime.

## Why now — two present-tense reasons, not a hypothetical

**1. The dev tree is currently inconsistent.** Measured in a clean checkout before this PR:

```
node_modules/@sillsdev/scripture                       2.0.5
node_modules/@eten-tech-foundation/platform-editor ->  .yalc/…  0.8.16, declares ^2.1.0
```

The yalc-linked editor has no nested copy, so it resolves to the hoisted root — **2.0.5, which does not satisfy its own `^2.1.0` range**. Nothing errors, because yalc links are applied after npm resolution and npm never polices them. After this PR the tree has exactly one copy, at 2.1.0, and the range is satisfied.

**2. There is a live contract split with the C# half.** `VerseRefConverter`'s reader resolves `verse ?? verseNum!.Value.ToString()` — a present `verse` string **wins over** `verseNum`. That is exactly the object shape 2.0.5's setter produced:

- C# writes `{chapterNum: 3, verseNum: 4, verse: '4b-5a'}`
- TypeScript sets `verseNum = 9`, leaving the stale string
- C# reads it back as **4b-5a, not 9** — silently, across the PAPI wire

And the C# side already exercises these setters in production: `InputRange` writes `verseRef.VerseNum` and `verseRef.ChapterNum` on real `SIL.Scripture.VerseRef` instances. TypeScript was the odd one out.

## What 2.1.0 actually changes

Skipping 2.0.6 picks up two `VerseRef` setter repairs. `chapterNum` ([#58](https://github.com/sillsdev/scripture/pull/58)) assigned the property to itself behind a `ToDo` placeholder and recursed infinitely, so every input threw `RangeError: Maximum call stack size exceeded` and the property was unusable. `verseNum` ([#60](https://github.com/sillsdev/scripture/pull/60)) is the same mis-port but did work — it set the backing field and nothing else, where the C# it ports also clears the verse string and rejects negatives.

**Neither repair is purely an improvement.** For non-negative invalid input, `chapterNum` moves from unusable-but-loud to usable-but-silently-permissive:

```ts
const v = new VerseRef('LUK 3:4', ScrVers.English);
v.chapterNum = 0;
// 2.0.5 -> RangeError: Maximum call stack size exceeded
// 2.1.0 -> "LUK 0:4", BBBCCCVVV 42000004, valid === true
```

Chapter 0 isn't valid, but `internalValid()` cannot detect it while the versification port is unfinished — the library's own new test has that assertion commented out behind a TODO.

## Dependency resolution

One hoisted copy either way, but for a reason worth stating precisely:

- The **registry** path is what guarantees it. `platform-editor@0.8.15` — the published latest, and what the lockfile records — declares `^2.0.5`, which **2.1.0 satisfies**. So npm hoists a single copy and there is no dual-version tree to avoid.
- The **yalc** path is not a guarantee. `dev-packages.json` names `"revision": "platform-yalc"`, which is a **branch, not a pin**: `dev-package-utils.ts` re-checks-out that branch on every install, and `getDevPackagePath` prefers `dev-packages/<folder>` (used by CI) but falls back to a sibling `../<folder>` for developer checkouts. Two developers can get two answers. `platform-editor@0.8.16` is a local yalc artifact — `npm view` returns 0.8.15 — so it is not evidence of anything on its own.

An earlier revision of this description called that revision "pinned" and leaned on 0.8.16. Both were wrong; the conclusion rests on the `^2.0.5` range instead.

## Why nothing here reaches the new guards

The `VerseRef` constructor writes backing fields directly, so construction cannot trip the guards. Every instance site is read-only: `findScrollTarget` and `toNavigationRecord` (comment-list-scroll.utils.ts), `handleVerseRefClick` (comment-list.web-view.tsx), `setHeresy` (quick-verse main.ts), `scrRefToBBBCCC` / `scrRefToBBBCCCVVV` (scripture-util.ts).

<details>
<summary>Full assignment audit — every chapterNum, verseNum, verse and book write in the repo</summary>

Cited by symbol rather than line: an earlier revision of this table used line numbers gathered before a 32-commit rebase, and five of eight had drifted onto unrelated code by the time it was read.

| Symbol | File | Target |
| --- | --- | --- |
| `determineLocationProperties` | platform-scripture-editor.utils.ts | spread of a `SerializedVerseRef` parameter |
| `convertScriptureRangeToEditorRange` | platform-scripture-editor.utils.ts | `SerializedVerseRef` declared with a `verseNum: -1` initialiser |
| `checkInputRange` | checks-side-panel.web-view.tsx | spread of a `SerializedVerseRef` literal |
| `createInventoryInputRange` | inventory.web-view.tsx | spread of a `SerializedVerseRef` literal |
| `transferFragmentsInfoArrayToMaps` | usj-reader-writer.ts | inline structural param, not a VerseRef |
| `getVerseRefForIndexInUsfm` | usj-reader-writer.ts | `verse` and `book` on a `SerializedVerseRef` |
| `normalizeStoredScrRefs` | scroll-group.service-host.ts | `book`, behind a `bookNum` guard |

All target plain objects or structural types — never a class instance, so no setter runs.

Two things worth noting from this table.

The `verseNum: -1` initialiser in `convertScriptureRangeToEditorRange` would throw under the new negative guard had it been an instance. That sentinel is not confined to hand-written literals either — the library produces it itself: `new VerseRef('LUK','','',ScrVers.Septuagint).toJSON()` yields `verseNum: -1`. So a `SerializedVerseRef` carrying `-1` can arrive straight from the library's own serializer. **2.1.0 therefore creates an asymmetric read/write contract**: both getters' TSDoc still documents `-1` as the not-valid value, while the setter now throws `VerseRefException` on it. Nothing here round-trips one through a setter, but code that reads `-1` legitimately and writes it back would break.

Also, `convertScriptureRangeToEditorRange` does not mutate the literal it declares. By the time the write happens, the variable has been rebound to `determineLocationProperties(...).verseRef`, whose USFM branch discards the defensive spread and returns the caller's object **by reference** from `usj-reader-writer`. So that write reaches a caller-owned object. It is benign today — every production caller crosses the PAPI boundary with JSON-deserialized plain objects — but `VerseRef` is structurally assignable to `SerializedVerseRef`, so TypeScript would accept an instance silently and the setter would then clear `_verse`, degrading `LUK 3:4b-5a` to `LUK 3:4`. Listed under follow-ups below.

</details>

## Verification

`npm run typecheck`, `npm run lint`, `npm test` and `dotnet test c-sharp-tests/` all exit 0 (1724 C# tests, 598 TS test files). Rebased onto current `main`; CI re-runs the full three-OS matrix on this head.

Post-install state confirmed by a full `npm i` — not `npm ci --ignore-scripts`, which skips `postinstall` and therefore skips `link-dev-packages` entirely, leaving a stale or absent dev-packages checkout:

```
node_modules/@sillsdev/scripture   2.1.0     <- the only copy in the tree
platform-editor 0.8.16 declares ^2.1.0       <- now satisfied
```

## Follow-ups (not this PR)

- **No test anywhere constructs a `VerseRef`** — zero hits for `new VerseRef(` / `VerseRef.tryParse` across all test files. A revert or a transitive pin would reinstate 2.0.5's setter and nothing would fail. `scripture-util.test.ts` already imports the real package, so a characterization test needs no new dependency.
- **`determineLocationProperties` writes through to a caller-owned object** — its USFM branch discards the defensive spread, and `usj-reader-writer` returns the caller's object by reference. Benign today, since production callers cross PAPI with JSON-deserialized plain objects, but worth a defensive spread.
- **No `VerseRefException` is caught anywhere** — given the new guards throw on values the library itself produces, any future throw propagates uncaught.
- **The root manifest entry is silently load-bearing for both library bundles** — neither `platform-bible-utils` nor `platform-bible-react` declares `@sillsdev/scripture`; both stay external only because their vite configs merge the root manifest's dependency keys into the rollup externals list. If it left root `dependencies`, both dists would bundle their own copy.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2723)

### DIFF
--- a/extensions/package.json
+++ b/extensions/package.json
@@ -44,7 +44,7 @@
     "react-dom": ">=19.0.0"
   },
   "dependencies": {
-    "@sillsdev/scripture": "^2.0.5",
+    "@sillsdev/scripture": "^2.1.0",
     "platform-bible-utils": "file:../lib/platform-bible-utils"
   },
   "devDependencies": {
--- a/extensions/src/hello-rock3/package.json
+++ b/extensions/src/hello-rock3/package.json
@@ -35,7 +35,7 @@
     "react-dom": ">=19.0.0"
   },
   "dependencies": {
-    "@sillsdev/scripture": "^2.0.5",
+    "@sillsdev/scripture": "^2.1.0",
     "platform-bible-utils": "file:../../../lib/platform-bible-utils"
   },
   "devDependencies": {
--- a/extensions/src/hello-someone/package.json
+++ b/extensions/src/hello-someone/package.json
@@ -35,7 +35,7 @@
     "react-dom": ">=19.0.0"
   },
   "dependencies": {
-    "@sillsdev/scripture": "^2.0.5",
+    "@sillsdev/scripture": "^2.1.0",
     "platform-bible-utils": "file:../../../lib/platform-bible-utils"
   },
   "devDependencies": {
--- a/extensions/src/legacy-comment-manager/package.json
+++ b/extensions/src/legacy-comment-manager/package.json
@@ -36,7 +36,7 @@
     "react-dom": ">=19.0.0"
   },
   "dependencies": {
-    "@sillsdev/scripture": "^2.0.5",
+    "@sillsdev/scripture": "^2.1.0",
     "platform-bible-utils": "file:../../../lib/platform-bible-utils"
   },
   "devDependencies": {
--- a/extensions/src/paratext-registration/package.json
+++ b/extensions/src/paratext-registration/package.json
@@ -36,7 +36,7 @@
     "react-dom": ">=19.0.0"
   },
   "dependencies": {
-    "@sillsdev/scripture": "^2.0.5",
+    "@sillsdev/scripture": "^2.1.0",
     "platform-bible-utils": "file:../../../lib/platform-bible-utils"
   },
   "devDependencies": {
--- a/extensions/src/platform-enhanced-resources/package.json
+++ b/extensions/src/platform-enhanced-resources/package.json
@@ -36,7 +36,7 @@
     "react-dom": ">=19.0.0"
   },
   "dependencies": {
-    "@sillsdev/scripture": "^2.0.5",
+    "@sillsdev/scripture": "^2.1.0",
     "@xmldom/xmldom": "^0.8.10",
     "platform-bible-utils": "file:../../../lib/platform-bible-utils"
   },
--- a/extensions/src/platform-get-resources/package.json
+++ b/extensions/src/platform-get-resources/package.json
@@ -36,7 +36,7 @@
     "react-dom": ">=19.0.0"
   },
   "dependencies": {
-    "@sillsdev/scripture": "^2.0.5",
+    "@sillsdev/scripture": "^2.1.0",
     "platform-bible-utils": "file:../../../lib/platform-bible-utils"
   },
   "devDependencies": {
--- a/extensions/src/platform-lexical-tools/package.json
+++ b/extensions/src/platform-lexical-tools/package.json
@@ -37,7 +37,7 @@
     "react-dom": ">=19.0.0"
   },
   "dependencies": {
-    "@sillsdev/scripture": "^2.0.5",
+    "@sillsdev/scripture": "^2.1.0",
     "platform-bible-utils": "file:../../../lib/platform-bible-utils"
   },
   "devDependencies": {
--- a/extensions/src/platform-scripture-editor/package.json
+++ b/extensions/src/platform-scripture-editor/package.json
@@ -36,7 +36,7 @@
     "react-dom": ">=19.0.0"
   },
   "dependencies": {
-    "@sillsdev/scripture": "^2.0.5",
+    "@sillsdev/scripture": "^2.1.0",
     "platform-bible-utils": "file:../../../lib/platform-bible-utils"
   },
   "devDependencies": {
--- a/extensions/src/platform-scripture/package.json
+++ b/extensions/src/platform-scripture/package.json
@@ -36,7 +36,7 @@
     "react-dom": ">=19.0.0"
   },
   "dependencies": {
-    "@sillsdev/scripture": "^2.0.5",
+    "@sillsdev/scripture": "^2.1.0",
     "platform-bible-utils": "file:../../../lib/platform-bible-utils"
   },
   "devDependencies": {
--- a/extensions/src/quick-verse/package.json
+++ b/extensions/src/quick-verse/package.json
@@ -36,7 +36,7 @@
     "react-dom": ">=19.0.0"
   },
   "dependencies": {
-    "@sillsdev/scripture": "^2.0.5",
+    "@sillsdev/scripture": "^2.1.0",
     "platform-bible-utils": "file:../../../lib/platform-bible-utils"
   },
   "devDependencies": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,7 @@
         "extensions/src/*"
       ],
       "dependencies": {
-        "@sillsdev/scripture": "^2.0.5",
+        "@sillsdev/scripture": "^2.1.0",
         "@xmldom/xmldom": "^0.9.8",
         "ajv": "^8.18.0",
         "chalk": "^4.1.2",
@@ -184,7 +184,7 @@
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
-        "@sillsdev/scripture": "^2.0.5",
+        "@sillsdev/scripture": "^2.1.0",
         "platform-bible-utils": "file:../lib/platform-bible-utils"
       },
       "devDependencies": {
@@ -379,7 +379,7 @@
       "version": "0.6.0-alpha.1",
       "license": "MIT",
       "dependencies": {
-        "@sillsdev/scripture": "^2.0.5",
+        "@sillsdev/scripture": "^2.1.0",
         "platform-bible-utils": "file:../../../lib/platform-bible-utils"
       },
       "devDependencies": {
@@ -448,7 +448,7 @@
       "version": "0.6.0-alpha.1",
       "license": "MIT",
       "dependencies": {
-        "@sillsdev/scripture": "^2.0.5",
+        "@sillsdev/scripture": "^2.1.0",
         "platform-bible-utils": "file:../../../lib/platform-bible-utils"
       },
       "devDependencies": {
@@ -517,7 +517,7 @@
       "version": "0.6.0-alpha.1",
       "license": "MIT",
       "dependencies": {
-        "@sillsdev/scripture": "^2.0.5",
+        "@sillsdev/scripture": "^2.1.0",
         "platform-bible-utils": "file:../../../lib/platform-bible-utils"
       },
       "devDependencies": {
@@ -587,7 +587,7 @@
       "version": "0.6.0-alpha.1",
       "license": "MIT",
       "dependencies": {
-        "@sillsdev/scripture": "^2.0.5",
+        "@sillsdev/scripture": "^2.1.0",
         "platform-bible-utils": "file:../../../lib/platform-bible-utils"
       },
       "devDependencies": {
@@ -656,7 +656,7 @@
       "version": "0.6.0-alpha.1",
       "license": "MIT",
       "dependencies": {
-        "@sillsdev/scripture": "^2.0.5",
+        "@sillsdev/scripture": "^2.1.0",
         "@xmldom/xmldom": "^0.8.10",
         "platform-bible-utils": "file:../../../lib/platform-bible-utils"
       },
@@ -808,7 +808,7 @@
       "version": "0.6.0-alpha.1",
       "license": "MIT",
       "dependencies": {
-        "@sillsdev/scripture": "^2.0.5",
+        "@sillsdev/scripture": "^2.1.0",
         "platform-bible-utils": "file:../../../lib/platform-bible-utils"
       },
       "devDependencies": {
@@ -878,7 +878,7 @@
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
-        "@sillsdev/scripture": "^2.0.5",
+        "@sillsdev/scripture": "^2.1.0",
         "platform-bible-utils": "file:../../../lib/platform-bible-utils"
       },
       "devDependencies": {
@@ -948,7 +948,7 @@
       "version": "0.6.0-alpha.1",
       "license": "MIT",
       "dependencies": {
-        "@sillsdev/scripture": "^2.0.5",
+        "@sillsdev/scripture": "^2.1.0",
         "platform-bible-utils": "file:../../../lib/platform-bible-utils"
       },
       "devDependencies": {
@@ -1017,7 +1017,7 @@
       "version": "0.6.0-alpha.1",
       "license": "MIT",
       "dependencies": {
-        "@sillsdev/scripture": "^2.0.5",
+        "@sillsdev/scripture": "^2.1.0",
         "platform-bible-utils": "file:../../../lib/platform-bible-utils"
       },
       "devDependencies": {
@@ -1089,7 +1089,7 @@
       "version": "0.6.0-alpha.1",
       "license": "MIT",
       "dependencies": {
-        "@sillsdev/scripture": "^2.0.5",
+        "@sillsdev/scripture": "^2.1.0",
         "platform-bible-utils": "file:../../../lib/platform-bible-utils"
       },
       "devDependencies": {
@@ -8806,7 +8806,9 @@
       "license": "MIT"
     },
     "node_modules/@sillsdev/scripture": {
-      "version": "2.0.5",
+      "version": "2.1.0",
+      "resolved": "https://registry.npmjs.org/@sillsdev/scripture/-/scripture-2.1.0.tgz",
+      "integrity": "sha512-jj5mMMCvhjsFW87H8WkIkG/lM6kp/k/vC6bk+hxwvU08NaffmgyAU+kDasNhIiViYvZ96JPUjwPl+z7doDLx9g==",
       "license": "MIT"
     },
     "node_modules/@sinclair/typebox": {
--- a/package.json
+++ b/package.json
@@ -131,7 +131,7 @@
   },
   "browserslist": ["extends browserslist-config-detect-electron"],
   "dependencies": {
-    "@sillsdev/scripture": "^2.0.5",
+    "@sillsdev/scripture": "^2.1.0",
     "@xmldom/xmldom": "^0.9.8",
     "ajv": "^8.18.0",
     "chalk": "^4.1.2",


katherinejensen00 (changes requested):
# Review — PR #2723: Bump `@sillsdev/scripture` to 2.1.0

**Reviewer:** Katherine Jensen · **Author:** Ira Hopkinson · **Branch:** `deps/sillsdev-scripture-2.1.0`

**Verdict: Approve.** No defect in the diff.

---

## Summary

This is a clean, complete, correctly-verified bump and I'd approve it. Every finding below is either a
correction to the PR description or a pre-existing item for a follow-up ticket — none is a problem with
the change itself.

I verified the library delta from the published tarballs rather than trusting the release notes:

- Only `verse-ref.ts` differs between 2.0.5 and 2.1.0. `canon.ts`, `versification.ts`, `scr-vers.ts`,
  `book-set.ts` and `index.ts` are byte-identical.
- `dist/index.d.ts` is **byte-identical** — zero public type-surface change, which is why `papi.d.ts`
  didn't regenerate and why typecheck is green.
- After normalizing the minifier's identifier renaming, the entire shipped `index.es.js` reduces to
  exactly the two setter hunks.
- `Canon` is unchanged at 123 books.
- The `VerseRef` constructor writes backing fields directly (`this._chapterNum = …; this._verseNum = …`),
  so nothing in this repo can reach the new guards.
- All six `VerseRef` instance sites in the repo are read-only.
- All twelve declaring manifests are bumped; none missed.
- The lockfile `integrity` hash matches the real 2.1.0 tarball byte-for-byte.

The core claim of the PR holds.

---

## Findings that matter

### 1. Sequence the companion template bump first

**Severity: high (process)** · `extensions/src/hello-rock3/package.json:38` — in the diff; same line in
all 10 extension manifests

The ten extension manifests are template-tracked subtrees. The `paranext-extension-template` branch
`deps/sillsdev-scripture-2.1.0` already exists making the identical upstream bump — I confirmed it via
the GitHub API. Landing that first eliminates ten hand-resolved conflicts on the next template pull.
`extensions/lib/update-from-templates.ts:26` auto-resolves only `package-lock.json` conflicts, so a
`package.json` conflict halts the loop one subtree at a time. Worth ordering deliberately.

### 2. The description doesn't mention that `chapterNum` gains a silent-acceptance path

**Severity: medium** · PR description, section "`chapterNum` — fails loud" (also commit message)

The heading is fair as a description of the 2.0.5 bug. The gap is that the section presents both fixes as
pure improvements, and for non-negative invalid input `chapterNum` goes from **unusable-but-loud** to
**usable-but-silently-permissive**. Verified by running both bundles side by side:

```
const v = new VerseRef('LUK 3:4', ScrVers.English);
v.chapterNum = 0;
// 2.1.0 -> toString() "LUK 0:4", BBBCCCVVV 42000004, valid === true
// 2.0.5 -> RangeError: Maximum call stack size exceeded
```

Chapter 0 isn't valid, but `internalValid()` can't detect it — the versification port is unfinished and
the library's own new test has that assertion commented out behind a TODO. Same silent-failure family as
the one the body correctly calls the more dangerous of the two. Worth one sentence.

### 3. The "one assignment away" line already exists, and writes through to a caller-owned object

**Severity: medium** ·
`extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts:475` — outside the diff

`startVerseRef` is not the literal from `:420` by the time it's mutated:

- `:434` rebinds it to `determineLocationProperties(...).verseRef`
- that function's `else` branch at `:379` does `verseRef = startVerseLocation.verseRef`, discarding the
  defensive spread made at `:366`
- `lib/platform-bible-utils/src/scripture/usj-reader-writer.ts:1167` returns
  `verseRef: usfmVerseLocation.verseRef` — the caller's object **by reference**

So `:475` mutates the caller's input. Benign today: every production caller crosses the PAPI boundary
with JSON-deserialized plain objects. But `VerseRef` is structurally assignable to `SerializedVerseRef`,
so TypeScript would accept an instance silently, and the setter would then clear `_verse`, degrading
`LUK 3:4b-5a` to `LUK 3:4`. Follow-up ticket to spread defensively in the USFM branch — not this PR's job.

### 4. The `-1` sentinel is produced by the library itself, not only by hand-written literals

**Severity: medium** ·
`extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts:420` — outside the diff

The body reassures that the `verseNum: -1` initializer is safe "because it's an object literal, so no
setter runs." True as far as it goes, but the sentinel isn't literal-only:

```
const p = new VerseRef('LUK', '', '', ScrVers.Septuagint);
p.verseNum   // -1
p.toJSON()   // {"book":"LUK","chapterNum":0,"verseNum":-1,"versificationStr":"Septuagint"}
```

A `SerializedVerseRef` from the library's own serializer can carry `-1`. Round-tripping it through a
setter gives `"LUK 3:"` under 2.0.5 and throws `VerseRefException` under 2.1.0. Both getters' TSDoc still
documents `-1` as the not-valid value, so 2.1.0 creates an asymmetric read/write contract worth naming.

### 5. Five of eight audit-table citations land on unrelated code

**Severity: medium** · PR description, "Why we're not affected today" table (also commit message)

| Cited | Actual | What the cited line really contains |
| --- | --- | --- |
| `platform-scripture-editor.utils.ts:230` | 373 | `// #region Editor Title Formatting` |
| `platform-scripture-editor.utils.ts:332` | 475 | a `sendCommand` string |
| `comment-list.web-view.tsx:571` | 624 | a comment inside `handleUpdateComment` |
| `scripture-util.ts:325,336` | 351,362 | grapheme book-name splitting |
| `usj-reader-writer.ts:3085-3162` | misses a real assignment at 3177 | — |

Worth saying clearly: **this is rebase drift, not carelessness.** I checked `39eb9a6~32` and all twelve
anchors were correct at the original base; the 32-commit rebase moved them. The conclusions are
substantively right at their real lines. But the table's value is letting a reviewer spot-check it, and
five of eight spot-checks currently land on nothing. Suggest regenerating the anchors and saying it once,
rather than as a table of corrections.

Two assignment sites are also missing from the table: `src/main/services/scroll-group.service-host.ts:191`
and `lib/platform-bible-utils/src/scripture/usj-reader-writer.ts:1111`. Both benign — targets are plain
objects, `book` isn't a setter either fix touched, and the first is guarded by
`if (!valuePossibleScrRef.bookNum) return;` on the line above. But the table reads as exhaustive.

### 6. There's a materially stronger argument for this PR than the one the body makes

**Severity: medium (description improvement)** · `c-sharp/JsonUtils/VerseRefConverter.cs:106` — outside
the diff

The C# reader resolves `verse ?? verseNum!.Value.ToString()` — a present `verse` string **wins over**
`verseNum`, which is exactly the object shape 2.0.5's setter produced. So under 2.0.5: C# writes
`{chapterNum:3, verseNum:4, verse:'4b-5a'}`, TS sets `verseNum = 9` producing
`{verseNum:9, verse:'4b-5a'}`, and C# reads it back as **4b-5a**, not 9 — silently, across the wire.

And `c-sharp/Checks/InputRange.cs:56,58` already writes `verseRef.VerseNum = …` / `.ChapterNum = …` on
real `SIL.Scripture.VerseRef` instances in production. The setters are already exercised on the C# half
of the pair; TypeScript was the odd one out.

That reframes the PR from "take a fix ahead of need" to "close a live contract split," and retires the
obvious reviewer objection — "then let's wait until someone writes that line" — far better than the
current framing.

---

## Secondary — worth knowing, not worth blocking

### 7. "12 manifests" oversells what changes at runtime

**Severity: low** · `extensions/webpack/webpack.config.base.ts:80` — outside the diff

`@sillsdev/scripture` is a webpack `external` for every extension, and the host injects the runtime
module (`src/extension-host/services/extension.service.ts:1342`,
`src/renderer/global-this-web-view.model.ts:57`). Only root `package.json:134` changes what actually
executes; the other eleven declarations move types and Vitest resolution. One clause in the body would
stop the next maintainer shipping a "fix" that bumps an extension manifest and leaves the injected
version untouched.

Related: four of the bumped manifests (`hello-someone`, `paratext-registration`,
`platform-get-resources`, and `extensions/package.json`) declare the package but have zero importing
files in `src/`. Keeping them in lockstep is defensible monorepo practice — flagging only so it's a
conscious call, since three of them also generate three of the ten template conflicts in finding 1.

### 8. `"revision": "platform-yalc"` is a branch, not a pin

**Severity: low** · PR description, "Dependency resolution" section

The body says "the `platform-yalc` revision **pinned** in `dev-packages.json`" and reports
platform-editor 0.8.16 declaring `^2.1.0`. That measurement isn't reproducible from the branch —
`.erb/scripts/dev-package-utils.ts` re-pulls that branch on every install, and this worktree's clone
reads `^2.0.5` while yours read `^2.1.0`. Same "pin," two answers.

Also: `npm view @eten-tech-foundation/platform-editor version` returns **0.8.15**, and
`package-lock.json:4259` records 0.8.15 with its own integrity hash. 0.8.16 is a local yalc artifact, not
a published version. The one-copy conclusion still holds — `^2.0.5` admits 2.1.0 — but it rests on an
unpublished local artifact rather than on the range that actually guarantees it.

### 9. Adjacent evidence the hand-edit workflow has drifted before

**Severity: nit** · `extensions/src/platform-enhanced-resources/package.json:40` — visible in the diff hunk

`@xmldom/xmldom` is `^0.8.10` here but `^0.9.8` at root `package.json:135` — a major-version split of one
dependency across two manifests, on the line directly below the one you edited. Not yours to fix, but
it's the argument for `npx npm-check-updates -u @sillsdev/scripture --workspaces --root` (or the existing
`npm run ncu:group` at `package.json:80`) over hand-editing, since there's no dependabot config and no
test asserting the twelve ranges agree.

---

## Follow-up tickets (not this PR)

- **No test anywhere constructs a `VerseRef`.** Zero hits for `new VerseRef(` / `VerseRef.tryParse` across
  all test files in `src`, `lib`, `extensions/src`, `e2e-tests`. The bump's stated value is therefore
  unguarded: a future `overrides` entry, a transitive pin, or a revert reinstates 2.0.5's setter and
  nothing fails — and the reintroduced failure mode is a silent stale `verse`, not a crash.
  `lib/platform-bible-utils/src/scripture/scripture-util.test.ts` already imports both `Canon` and
  `VerseRef` from the real package, so a ~10-line characterization test needs no new dependency and no
  mock. Reasonable to decline on a dependency bump; worth a ticket.
- **Defensive spread in the USFM branch** of `determineLocationProperties` (finding 3).
- **No `VerseRefException` is caught anywhere** — zero hits across `src/`, `lib/`, `extensions/src/`.
  Given findings 2 and 4, where the new guards throw on values the library itself produces, any future
  throw propagates uncaught.
- **The root manifest entry is silently load-bearing for both library bundles.** Neither
  `lib/platform-bible-utils` nor `lib/platform-bible-react` declares `@sillsdev/scripture`; both stay
  external only because their vite configs merge the **root** manifest's dependency keys into the rollup
  externals list (`lib/platform-bible-utils/vite.config.ts:14`,
  `lib/platform-bible-react/vite.config.ts:39`). If it ever leaves root `dependencies`, both dists start
  bundling their own copy — two `VerseRef` classes, one throwing on negatives and one not. That config's
  own comment at `:33-38` records this exact failure already happening once with
  `unicode-segmenter/grapheme`.

---

## On the description's length

I'd push back on just cutting it. The detail is the reason findings 5 and 8 were checkable at all — a
shorter body would have hidden the stale anchors rather than fixed them. The problem isn't word count,
it's ordering: the fact a reviewer must confirm (twelve manifests bumped, lockfile agrees, one hoisted
copy) appears once in the opening sentence and is never given a checkable form, while roughly 70% argues
a null hypothesis. Moving the audit table into a `<details>` block and leading with the manifest census
would fix it without losing anything.

Two paragraphs I'd genuinely cut: the one ranking which of the two bugs is more dangerous (neither is
triggered here), and the rebase narration — per `.claude/rules/code-quality/forward-facing-comments.md`
that belongs in the commit message, and it's also what tells the reader the citations were gathered
pre-rebase.

---

## Footnotes

Raised in earlier drafts, dropped as not worth the author's time:

- **Lockfile `integrity` durability.** The added `resolved`/`integrity` are authentic, but 2168 of 2483
  `node_modules` entries lack `resolved` (2185 lack `integrity`), so missing is this file's norm and the
  next full regeneration may strip it again. The body's claim is defensible as written.
- **Devin Review passed without reviewing** — "Full review skipped: trial expired and no credits
  remaining." Not the author's doing and not actionable by them.
- **AI attribution** — body uses the stock `Generated with Claude Code` footer rather than CLAUDE.md's
  `AI-assisted — [session 1](<url>)` form, and there's no `Session-URL:` trailer. Trivial.
- **`"private": true` is on only 8 of 19 manifests** (missing on `platform-scripture-editor` among
  others). Inert — there is no `npm publish` in any workflow or script. Separate housekeeping item at most.
- **Canon-length hard-coding**: `manage-books.web-view.tsx:57` uses `'0'.repeat(123)` while
  `checklist.web-view.tsx:264` uses `'0'.repeat(124)`; `Canon.allBookIds.length` is 123. A real
  disagreement, but pre-existing and untripped here since `Canon` is byte-identical.

---

## Corrections to my own review

Two findings I raised and retracted, in case they surface from another reviewer:

- I flagged the missing `@sillsdev/scripture` declaration in `lib/platform-bible-utils` and
  `lib/platform-bible-react` as an oversight. It isn't — the vite configs handle it deliberately. My
  framing that "36 source files silently change compile target" was also wrong: the library's `.d.ts` is
  byte-identical, so nothing about compilation changes.
- I claimed `extensions/src/c-sharp-provider-test` was a manifest deliberately left untouched. It has no
  `package.json` at all.

> @katherinejensen00 reviewed 13 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2723#-:-P10oD2z3V0cpQTjK65k:b9a9t98)* status: all files reviewed, 1 unresolved discussion (waiting on irahopkinson).

___
*[`extensions/src/hello-rock3/package.json` line 38 at r1](https://reviewable.io/reviews/paranext/paranext-core/2723#-P10oKaU-nOEcV50dAaO:-P10oKaU-nOEcV50dAaP:b-glaacg) ([raw file](https://github.com/paranext/paranext-core/blob/2695737f6061288a30a199fa4e27a9bcb14b1b40/extensions/src/hello-rock3/package.json#L38)):*
> ```JSON
>   },
>   "dependencies": {
>     "@sillsdev/scripture": "^2.1.0",
> ```

### 1. Sequence the companion template bump first

**Severity: high (process)** · `extensions/src/hello-rock3/package.json:38` — in the diff; same line in
all 10 extension manifests

The ten extension manifests are template-tracked subtrees. The `paranext-extension-template` branch
`deps/sillsdev-scripture-2.1.0` already exists making the identical upstream bump — I confirmed it via
the GitHub API. Landing that first eliminates ten hand-resolved conflicts on the next template pull.
`extensions/lib/update-from-templates.ts:26` auto-resolves only `package-lock.json` conflicts, so a
`package.json` conflict halts the loop one subtree at a time. Worth ordering deliberately.

irahopkinson (commented):
Thank you — this was an exceptionally useful review, and verifying the library delta from the published tarballs rather than the release notes is more than I did. The description and commit message are rewritten; the diff is unchanged apart from a rebase onto current `main`.

## Finding 1 — already resolved

Both template PRs merged today, ahead of this one: paranext-extension-template#162 and paranext-multi-extension-template#98. So the ten hand-resolved subtree conflicts you were guarding against don't arise. You couldn't have known — they landed after your review.

## Finding 5 — fixed, and the anchors can't drift again

You were right, and right about the cause. I regenerated all of them against the current head and switched the whole table to **symbol citations** rather than line numbers, per `.claude/rules/docs-durability.md`. The table also now covers `book` writes, which picks up the two sites you flagged as missing — `normalizeStoredScrRefs` in scroll-group.service-host.ts and the `book` assignment in `getVerseRefForIndexInUsfm` — so the exhaustiveness claim is now actually true. It's in a `<details>` block as you suggested, with the manifest census leading instead.

## Finding 6 — taken, and it's now the lead argument

This was the most valuable thing in the review. I verified both halves myself before repeating them: `VerseRefConverter.cs` really does resolve `verse ?? verseNum!.Value.ToString()`, and `InputRange.cs` really does write `verseRef.VerseNum` / `.ChapterNum` on real `SIL.Scripture.VerseRef` instances. You're right that it reframes the PR from "take a fix ahead of need" to "close a live contract split," and it retires the obvious objection far better than what I had.

**Chasing that turned up a second present-tense reason.** Measured in a clean checkout before this PR, the yalc-linked `platform-editor@0.8.16` declares `^2.1.0`, has no nested copy, and therefore resolves to the hoisted root at **2.0.5 — which does not satisfy its own range**. Nothing errors because yalc links are applied after npm resolution and npm never polices them. So the dev tree is already inconsistent, and this PR is what makes it consistent. Both reasons now lead the description.

## Finding 3 — this is a correction to my framing, not just a follow-up

You're right and I'd understated it. The old description claimed the instances were "one assignment away"; your point is that the assignment **already exists** and writes through to a caller-owned object, because the variable has been rebound to `determineLocationProperties(...).verseRef` by then, whose USFM branch discards the defensive spread and returns the caller's object by reference. I traced it and it holds.

That framing is gone — the description now leads with the two present-tense reasons instead, which is a more accurate account of why this is worth doing. The write-through is documented in the audit block with the `VerseRef`-is-structurally-assignable consequence spelled out (the setter would clear `_verse`, degrading `LUK 3:4b-5a` to `LUK 3:4`), and the defensive spread is listed under follow-ups.

## Finding 4 — added, including the part I'd missed

In the audit block now. Your point that the sentinel isn't literal-only is the important half: `new VerseRef('LUK','','',ScrVers.Septuagint).toJSON()` yields `verseNum: -1`, so a `SerializedVerseRef` carrying `-1` can arrive straight from the library's own serializer, not just from a hand-written initialiser.

I'd initially only noted that it's library-produced. The sharper consequence you named is the asymmetry, and that's now stated explicitly: both getters' TSDoc still documents `-1` as the not-valid value while the setter throws `VerseRefException` on it, so code that reads `-1` legitimately and writes it back would break. Nothing here round-trips one, but it's a real wart in 2.1.0 worth having on the record rather than buried.

## Findings 2 and 7 — both added

The `chapterNum = 0` silent-acceptance path is in, with your side-by-side output and the note that `internalValid()` can't catch it while the versification port is unfinished. And the description now says only the root manifest changes what executes, since `@sillsdev/scripture` is a webpack external and the host injects the runtime module — with the lockstep bump called out as deliberate.

## Finding 8 — you're right on the substance; the specific discrepancy has a different cause

Conceded and corrected in the description: `"revision": "platform-yalc"` is a **branch, not a pin**, `dev-package-utils.ts` re-checks-out that branch on every install, and `getDevPackagePath` prefers `dev-packages/<folder>` but falls back to a sibling `../<folder>` for developer checkouts — so two developers genuinely can get two answers. And 0.8.16 is an unpublished local yalc artifact; `npm view` returns 0.8.15. Citing it as evidence was wrong. The one-copy conclusion now rests on `platform-editor@0.8.15` declaring `^2.0.5`, which 2.1.0 satisfies — a fact independent of anyone's local state.

On why your clone read `^2.0.5`: I think it was stale rather than a moving target. The bump landed on `platform-yalc` in `b4d1a0e0b` on **2026-08-27**, roughly two weeks before your review, and the branch has declared `^2.1.0` continuously since. The likely mechanism is that `link-dev-packages` runs only from `postinstall`, so `npm ci --ignore-scripts` — the natural choice in a throwaway review worktree — skips the re-pull entirely and leaves whatever was there. Worth knowing for future reviews of anything touching this path.

## Finding 9 — noted, not actioned

The `@xmldom/xmldom` major split between `platform-enhanced-resources` and the root is real and pre-existing. Left alone deliberately to keep this diff to one dependency, but your point about `ncu` over hand-editing stands.

## On length

You were right not to just cut it, and the reordering suggestion was the correct diagnosis. The census leads, the audit is collapsed, and the two paragraphs you named — the danger-ranking and the rebase narration — are gone.

## Follow-up tickets

I haven't filed the four you identified — no `VerseRef` test anywhere, the defensive spread in `determineLocationProperties`, no `VerseRefException` caught anywhere, and the root manifest being silently load-bearing for both library bundles' vite externals. They're listed in the description so they aren't lost. They're yours to file if you think they're important enough — you found them and you have the reasoning, so a ticket in your words will be better than one in mine.

Thanks also for the two self-corrections at the end — useful to know what was already checked and ruled out.

> @irahopkinson+AI made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2723#-:-P12KDLE5R-xq90r1v-m:b-52ey2h)* status: all files reviewed, 1 unresolved discussion (waiting on irahopkinson).

katherinejensen00 (approved):
:lgtm:

> @katherinejensen00 reviewed all commit messages and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2723#-:-P12MHwLA_AmXrpj3FZX:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on irahopkinson).

PR #2725, opened 2026-08-26
docs: two verification traps where a clean check proves nothing

rolfheij-sil:
## Summary

Two documentation changes recording verification traps from the PT-4275 multi-window stack — both cases where a check reported success **without having examined the thing it claimed to cover**. One of them says that a verification route people actually use is structurally incapable of catching what they think it catches, which is why this wants real eyes rather than a skim.

### Why review this

Not part of the current epic — process cleanup that came out of the multi-window stack. It is worth reviewer time now because a reader following `Paranext-Core-Patterns.md` today would verify an experimental marker with a diff that cannot show it, and conclude the marker is present when nothing checked.

## Changes

**1. `Paranext-Core-Patterns.md` — corrects an implied verification route.** The Experimental APIs section documents both marker surfaces correctly, but never says that the check people actually run covers only one of them. `lib/papi-dts/papi.d.ts` is generated from the *type* surface, so a command, network object, data provider or network event registered without `'x-experimental': true` produces a **byte-identical** `papi.d.ts`. A regenerated-types diff is therefore incapable of revealing a missing wire marker. Confirm it by reading the registration's documentation object or the live OpenRPC document at `rpc.discover` — never by inferring it from regenerated types.

**2. `.claude/rules/scoped-sweeps-cannot-prove-absence.md` — new rule.** A search narrowed by path, glob, exclusion, or simply by looking only where you expect the answer to live means **"not here"**, never **"not anywhere"**. Reporting the first as the second is how a false gap gets filed and a real one gets missed. Cites both instances that happened, in different shapes, on the same day:

- An exclusion that swallowed its own evidence: a sweep for stale anchor links into `Architecture-Decisions.md` excluded that file with `grep -v 'Architecture-Decisions.md'`, which drops every line *containing that filename* — precisely what an anchor link is. It came back clean while the stale anchors sat there. Exclude by **path**, not by substring.
- A search that looked where the caller was rather than where the thing is declared: a check for a missing `'x-experimental': true` searched the shard files, because that is where the method lives. The marker lives on the registration. The clean result nearly became a filed API-compliance gap that did not exist.

The rule also carries the generated-artifact corollary from change 1, and a note that `grep` on the team's machines is ugrep rather than GNU grep, so flag semantics should be checked when they affect a result.

**3. `Code-Review-Guide.md` — a new "Before the verdict" section, plus a pointer from the review analyzer.** Four checks a review completes before it is reported, each one a review that skipped it and a defect the next stage caught: a requester's checklist is a floor, so run the full review pass alongside it, hold the verdict until that pass finishes, and then ask which branch of every two-way switch the diff's tests never exercise; run the control before attributing a symptom to a change, by checking the same signal on the base and on unrelated branches; verify a premise in code, citing the line, before it reaches a rationale, an ADR or a report, and grep the type definitions before accepting "cannot be done" as the reason for a workaround; and keep the end-to-end scenarios in the gate for any shutdown, close or persistence change, because what fails there is ordering and lifetime, which a mocked test reproduces by construction rather than exercises. `review-analyzer.md` gets a one-line pointer to the section rather than a copy, so the rules have one home.

## Scope narrowed after #2720

This PR originally carried three further changes about ADR **numbering** — a correction to the "numbers are claimed at merge" bullet, guidance on sweeping cross-references after a renumber, and a `Git-Guide.md` note on citing ADRs by title until merge. #2720 replaced numbers with write-time slugs, which removes the renumbering hazard those changes existed to document, so all three have been **dropped** rather than rewritten. The branch is rebased onto that work and no longer touches `Architecture-Decisions.md` or `Git-Guide.md`. The one lesson from that material with life beyond numbering — exclude by path, not substring — survives as the first bullet of the new rule, reworded so it no longer depicts the retired `ADR-NNNN` link format.

## Verification

Ran the `verify-standards` gate against this branch's diff, which the repo requires for any PR adding standards content. **The new content comes out clean — zero findings on the added lines.** Checked by line range, not by filename: `Paranext-Core-Patterns.md` carries 15 pre-existing findings at lines 65–669, all well clear of the addition at line 1168, and most of those are legitimate noise (Windows/Linux/macOS platform API names cited as external, which is correct). They are pre-existing on `main`, out of scope here, and untouched — folding a documentation audit into two small edits would bury the change.

## Risk Level

Low — documentation only. No source, no tests, no build output. Both claims are backed by a real incident named in the text, and the `papi.d.ts` claim is checkable from the generator's inputs.

AI-assisted — [session 1](https://claude.ai/code/session_018FNQ8Ewcs6pPg2i3qB3APM)

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2725)

### DIFF
--- a/.claude/agents/review-analyzer.md
+++ b/.claude/agents/review-analyzer.md
@@ -19,7 +19,7 @@ You will be given:
 
 ## Accuracy Requirement
 
-**Before reporting any finding, verify it against the actual code.** Read the relevant file (or use `git show $MERGE_BASE:<file>` for the before-state) to confirm the issue exists. Do not flag something based on pattern assumptions or file names alone. If you cannot confirm a finding from the actual code, omit it entirely or note that you could not verify it. A small number of accurate findings is better than a long list with errors.
+**Before reporting any finding, verify it against the actual code.** Read the relevant file (or use `git show $MERGE_BASE:<file>` for the before-state) to confirm the issue exists. Do not flag something based on pattern assumptions or file names alone. If you cannot confirm a finding from the actual code, omit it entirely or note that you could not verify it. A small number of accurate findings is better than a long list with errors. The checks a review completes before it is reported are in [Code-Review-Guide.md](../../.context/standards/Code-Review-Guide.md#before-the-verdict).
 
 ## Severity Rubric
 
--- a/.claude/agents/review-analyzer.md
+++ b/.claude/agents/review-analyzer.md
@@ -19,7 +19,7 @@ You will be given:
 
 ## Accuracy Requirement
 
-**Before reporting any finding, verify it against the actual code.** Read the relevant file (or use `git show $MERGE_BASE:<file>` for the before-state) to confirm the issue exists. Do not flag something based on pattern assumptions or file names alone. If you cannot confirm a finding from the actual code, omit it entirely or note that you could not verify it. A small number of accurate findings is better than a long list with errors. The checks a review completes before it is reported are in [Code-Review-Guide.md](../../.context/standards/Code-Review-Guide.md#before-the-verdict).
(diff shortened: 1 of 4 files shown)


rolfheij-sil:
🤖 Claude: Folded into #2676 as 10bc25c8e80 at Rolf's request to reduce the number of open PRs; the branch is kept.

PR #2727, opened 2026-08-26
PT-2626: native string call sites

mattgetgen:
The string-util analysis that has been done on paranext-core found some areas where grapheme-aware string functions are not necessary. In terms of performance, native > GraphemeString class > string-util, so I decided to move these call sites to the native string utils for better perf.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2727)


katherinejensen00 (changes requested):
Review: PT-2626 native string call sites

Nice piece of work — the direction is right and the index-space discipline is tamazing: every index-producing call was converted together with the call that consumes it, and I found no surviving half-conversion anywhere in src/. The plan doc's "index-space consistency" section is the clearest statement of that invariant in the repo.

Everything below was verified by running the pre-PR helper against the post-PR native method, not by reading. Items 13–17 are pre-existing and marked as such — offered as "while you're in here", not as things you broke.

Merge gate: items 1–6. Item 1 first — it's the only finding where an ordinary typo produces a total, hard-to-diagnose failure.

---

🚫 Blockers

1 · MEDIUM-HIGH · src/extension-host/services/extension.service.ts:227

A malformed manifest that used to skip one extension now stops every extension from loading.

extensionManifest.name.includes('..') duck-types when name isn't a string. Verified for {"name":["../../evil"]}: .includes('..') → false, and FORBIDDEN_EXTENSION_NAMES.some(n => n === name) → false, so it passes both guards. Pre-PR, includes(name,'..') threw Input must be a string inside the try at :483-497, which Promise.allSettled + the filter at :501 turned into "log and skip this one." Now the bad entry survives to the .sort() at :532 — after the filter, outside any try/catch — where extA.name.startsWith throws (verified: name.startsWith is not a function), getExtensions() rejects, and nothing loads.

Not a privilege escalation (anyone who can drop an extension folder already has code execution), but "name": ["foo"] is an ordinary authoring slip and the resulting error points nowhere near the manifest. Restore a typeof name !== 'string' check in parseManifest.

2 · MEDIUM · lib/platform-bible-utils/src/scripture/scripture-util.ts:298

Both edits in this file are inert — dist/ wasn't rebuilt, and CI structurally cannot catch it.

package.json exports maps the package to ./dist/index.js; there's no webpack resolve.alias, no tsconfig path, and the workspace symlink still resolves through exports. The committed chunk dist/scripture-util-Df5kpddK.js has the same content hash as at the base commit and still contains startsWith(Intl.getCanonicalLocales(t)[0],"zh") and !O(l,"z").

Root npm run build doesn't include build:platform-bible-utils, so test.yml:147's "Verify no files changed after build" never sees the skew — which is why all six checks are green. Code-Style-Guide.md:306 states the convention for the sibling library with an identical setup, and 19 of the last 20 commits touching this package's src/ also committed dist/.

Fix: npm run build:pbu and commit. collectUsjMarkers (:955) is the one that matters — it runs per marker over a whole USJ document. No correctness impact; the optimization simply isn't shipping.

3 · MEDIUM · src/extension-host/services/extension-storage.service.ts:54

The .. traversal guard stops rejecting non-string paths.

Verified: ['../../secret.txt'].includes('..') → false (that's Array.prototype.includes, element equality), where pre-PR the helper threw. isValidFileOrDirectoryName(['../../secret.txt']) also returns true — the regex coerces — so both guards pass and only an incidental path.join type error stops it downstream. Defense-in-depth on an extension-facing API rather than an escalation, but the guard should hold.

Worth noting the string path got stricter here — native catches '..' + U+0301, which grapheme matching missed. This is only the non-string axis.

4 · MEDIUM · src/node/models/execution-token.model.ts:16 (also src/node/services/execution-token.service.ts:6-7)

!name || name.length < 1 no longer rejects non-strings.

Verified: 42, true, and {} all pass, because (42).length is undefined and undefined < 1 is false. stringLength threw on all three. The failure now surfaces in sanitizeDirectoryName's .replace, far from the cause. Same one-line fix as #1 and #3.

5 · MEDIUM · src/extension-host/services/settings.service-host.ts:192 (also project-settings.service-host.ts:57)

This is the one converted guard that fails open.

The code reads "if the error message says no validator exists, let the change through." Native includes is a strict superset of the grapheme version — it matches where grapheme refused, e.g. when the needle is followed by a combining mark — so a genuine validator error can be misread as "no validator" and an invalid setting value written instead of throwing.

Unlike the other conversions, the haystack here is an arbitrary error message, not ASCII by construction, so the plan's qualifying rule doesn't cover it. Low reachability today; worth an explicit decision rather than inheriting the property-name argument.

6 · MEDIUM · src/renderer/services/web-view.service-host.ts:2152

The two riskiest conversions shipped untested, and the plan's justification for skipping one is factually wrong.

"Verification for all three" promises a WI-3 test (splice a document with non-ASCII text around <head>, assert byte-identical output) and WI-1 cases. Only WI-2 landed — and those two tests are good and genuinely falsifiable. But WI-1's premise, "the existing data-provider and network-object suites already exercise the traps", is false: there is no data-provider.service.test.ts and no network-object.service.test.ts. And web-view.service-host.test.ts (832 lines) has no <head reference at all.

Given ~80 sites changed, including path-traversal and iframe-sandbox guards, two round-trip tests is thin. The WI-3 test is the one I'd insist on — that splice is the sole CSP mechanism for non-URL web views. I'd drop the WI-1 ask: native ⊇ grapheme, so a get/forgetful test can't fail.

---

Should address

7 · MEDIUM · .context/standards/Code-Style-Guide.md:433-435

The current standard tells the next author to do the opposite of what this PR establishes.

It reads "Check platform-bible-utils first… string-util.ts — string manipulation. If it exists, use it." A grep of .context/standards/** and .claude/rules/** for grapheme|GraphemeString|stringz returns that one contradictory line and nothing else. The qualifying rule — "both needle and haystack ASCII by construction" — is the whole safety argument for 80 conversions, and it lives only in a file headed "Frozen record — follow the current files rather than this document."

To be clear, .context/plans/ is the right home for the record of why. What's missing is promoting the rule into .claude/rules/ plus an ADR entry — exactly what root CLAUDE.md asks for: "fold it into the relevant standard… that is what the agents read and enforce on the next feature." Without it, this gets partly reverted.

8 · MEDIUM · src/renderer/services/web-view.service-host.ts:2152

Behavior is faithfully preserved, but the new comment now blesses a real gap.

I verified the splice is byte-identical pre/post on documents with astral characters, and that stringz.indexOf clamps a negative position exactly like native — the comment is accurate. The issue is what it endorses: with no <head, the CSP <meta> is spliced after the first > anywhere in the document, and a <meta> CSP is only honored inside <head>. <head data-x="a>b"> also puts headEnd inside an attribute value. Deferring the 2-line guard is defensible — but the comment reads as "this is fine" rather than "this is a known gap."

9 · MEDIUM · src/shared/utils/menu-document-combiner.ts:47 (also :65,76,90,93,332, settings-document-combiner-base.ts:222, theme.service-host.ts:135,445)

Namespace-ownership gates were converted under a rule that doesn't actually apply to them.

The needle is `${documentName}.` — derived from the extension name, and parseManifest doesn't restrict that to ASCII. Verified the direction: 'foó.bar'.startsWith('foo') is true native, false grapheme. I could not construct a cross-namespace escalation, so severity is bounded — but these inherited the "JavaScript property names from a TypeScript API surface" argument, and they aren't that.

10 · LOW-MED · src/extension-host/services/extension.service.ts:679

.slice(0, -'.d.ts'.length) runs on a basename not guaranteed to end in .d.ts.

The readDir path filters on .d.ts at :631, but the manifest-types path at :616 has no extension check. "types": "./my-ext-types.ts" yields cache folder my-ext-typ, and the extension silently loses Intellisense. Separately, a literal used purely for its length reads as a puzzle — a named constant would say it directly.

11 · LOW-MED · src/extension-host/services/extension.service.ts:649 (also :666)

Extension-name prefix matching against filesystem names loosened, and macOS stores filenames NFD.

Extension cafe now matches a sibling café.d.ts (decomposed) that grapheme matching rejected; :666 adopts that base and :679 slices it into the cache folder name.

12 · LOW · src/shared/services/data-provider.service.ts:694

The plan's "not reachable from any TypeScript identifier anyone writes" is wrong, and the failure is loud.

Verified: U+0301 is ID_Continue, so ({ get́Data(){} }) parses and Object.keys returns ["get́Data"]; native .startsWith('get') → true, grapheme → false. Such a method now groups as 'get', trips the get/set-count check at :710, and registerEngine rejects — the whole provider never comes up. Exotic, but worth correcting the claim rather than leaving it as a stated guarantee.

---

Pre-existing — noticed while reviewing, not introduced here

13 · MEDIUM · src/shared/utils/extension-asset.utils.ts:27

The scheme guard checks 14 characters while the strip removes 17, so a scheme-less URI resolves into another extension's asset directory.

Verified against the base commit — startsWith(uri,mismatched, so this is not your change. But you rewrote both lines, and it's a one-string fix:

papi-extension://my-ext/assets/x           -> {extensionName:"my-ext", assetPath:"assets/x"}
papi-extensionXXvictim/assets/secret.json  -> {extensionName:"ictim",  assetPath:"assets/secret.json"}

Reachable from extension-supplied strings at database.service-host.ts:85,105. The sibling enhanced-resource.utils.ts:51 already guards the full ://.

14 · MEDIUM · lib/platform-bible-utils/src/scripture/scripture-util.ts:311

'\xff08' is 'ÿ08', not the fullwidth paren （ (U+FF08) the comment describes — the Chinese parenthetical stripping has never worked.

Verified: '创世记（创世纪）'.split('\xff08') → ["创世记（创世纪）"]; with '\uff08' → ["创世记","创世纪）"]. Untouched by this PR, but the plan examined these exact lines and deferred them without noticing the needle can never match. Wants '\uff08'.

15 · LOW · src/main/services/extension-asset-protocol.service.ts:35

getMimeTypeForFileName has no .toLowerCase() and no .webp; getMimeTypeForImageId (enhanced-resource.utils.ts:131) documents itself as mirroring it and has both. logo.PNG → application/octet-stream on one path, image/png on the other. Your change made the two structurally identical, so folding them is cheap now.

16 · LOW · src/node/utils/util.ts:96

split('://') with no limit truncates a path containing ://: getPathFromUri('file://C:/cache/https://x/y') → uriPath = 'C:/cache/https', wrong path, no error.

17 · LOW · src/extension-host/services/extension.service.ts:338

.endsWith('.zip') here vs .endsWith('zip') at :764,771,935,941. A file named notazip is an archive to one path and not the other. Line 338 is the one you touched.

---

Comments and documentation

Six new blocks, ~26 lines. Headline: the rationale is stated five times, and repetition is what stops comments being read. Promote it once (#7) and reduce four of these to one-liners pointing at it.

18 · LOW · src/renderer/services/web-view.service-host.ts:2150

The trailing clause "the same behavior the grapheme-aware helper had, so malformed documents splice as they did before" is the literal forbidden form in .claude/rules/code-quality/forward-facing-comments.md. Keep the first clause — the -1 clamp is a real gotcha — and cut the comparison.

19 · LOW · src/renderer/services/web-view.service-host.ts:2149

"which measured as seconds of startup time" has no number or source. I benchmarked it and you're right, so please just put the data in:

┌──────────┬──────────────────┬────────┐
│ Document │ helper (4 calls) │ native │
├──────────┼──────────────────┼────────┤
│ 100 KB   │ 36 ms            │ ~0 ms  │
├──────────┼──────────────────┼────────┤
│ 500 KB   │ 186 ms           │ ~0 ms  │
├──────────┼──────────────────┼────────┤
│ 2 MB     │ 1023 ms          │ ~0 ms  │
└──────────┴──────────────────┴────────┘

Web views inline their bundled app, so 1–2 MB is ordinary — the plan's "~1.2 s per tab open" lands almost exactly.

20 · LOW · src/shared/services/data-provider.service.ts:400

"the one place a construct-once instance cannot help" — a grep for GraphemeString|construct-once across the repo hits only the plan doc and this comment, so a future reader can't resolve the referent.

21 · LOW · src/shared/utils/papi-util.test.ts:44

Four lines of prose above two tests whose names already state the case. One line does it.

22 · LOW · .context/plans/pt-2626-native-string-call-sites.md:171

Cites parseExtensionAssetUri; the function is getAssetPathInfoFromExtensionUri. And at :189-200, "the bug this fixed on the way past" isn't a bug — grapheme and native slice agree for every input including astral and combining-mark cases, since PDP_FACTORY_LABEL is ASCII. Worth softening so nobody hunts for it.

▎ Keep verbatim: extension-asset.utils.ts:53. It's the only comment stating the inverse rule — why stringLength was deliberately kept — and it's what will stop the next sweep from "finishing the job."

---

Efficiency and altitude

23 · MEDIUM · lib/platform-bible-react/src/components/shared/book.utils.ts:165

The largest unconverted hot site. English book name and 3-letter canon id are ASCII by construction and meet the rule; they were blanket-deferred alongside the localized-name case. Runs per book, per keystroke, in the book picker.

24 · LOW · lib/platform-bible-utils/src/string-util.ts

Worth considering the higher-altitude fix: an ASCII fast path inside the helper is provably correct (ASCII haystack ⇒ the index spaces coincide), is checked rather than asserted by a human rule, is ~15 lines against 80 hand edits, and every unconverted site keeps benefiting. Not a request to redo this PR — a note for the follow-up.

25 · LOW · src/shared/utils/extension-asset.utils.ts:57

Two full grapheme segmentations run on attacker-supplied decoded strings before the cheap assets/ confinement check at :62. A .length > 100 && short-circuit is semantics-preserving, since UTF-16 length ≥ grapheme count.

Two full grapheme segmentations run on attacker-supplied decoded strings before the cheap assets/ confinement check at :62. A .length > 100 && short-circuit is semantics-preserving, since UTF-16 length ≥ grapheme count.

26 · LOW · lib/platform-bible-utils/src/scripture/scripture-util.ts:833

The plan calls this "the only quadratic shape left" and defers it, but no TODO(PT-XXXX) exists in source. The deferral lives only in a doc that declares itself frozen.

27 · LOW · repo-wide

Nothing mechanical prevents regression. lib/eslint-plugin-paranext/ already hosts 16 custom rules and no-restricted-imports is configured, so a rule pinning the ASCII-only sites is cheap.

---

Nits

28 · src/shared/utils/extension-asset.utils.ts:36 — includes('/') then indexOf('/') scans twice; indexOf alone suffices.

29 · extensions/src/platform-scripture/src/checks/inventories/inventory-utils.ts:9 — this conversion's haystack is user scripture text, so it violates the stated rule. It's provably safe (codePointAt(s,0) = first code point of the first cluster = s.codePointAt(0); I fuzzed 14 adversarial inputs, zero differences), but it belongs in the plan's table so the next reader doesn't re-derive it.

30 · src/shared/models/data-provider.model.ts:240 — ['set','get','subscribe'] is order-neutral only because no member is a prefix of another. That invariant is undocumented; a future 'sub' or 'getAll' would silently break the split.

> @katherinejensen00 reviewed 33 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2727#-:-P04i6pC3GfyoNZQp8PW:b3hawhp)* status: all files reviewed, 1 unresolved discussion (waiting on mattgetgen).

___
*[`lib/platform-bible-utils/src/scripture/scripture-util.ts` line 298 at r1](https://reviewable.io/reviews/paranext/paranext-core/2727#-P04ibZNBEeq6HN6aiL_:-P04ibZNBEeq6HN6aiLa:bsm5wnf) ([raw file](https://github.com/paranext/paranext-core/blob/af4f93f0366537ccee83846f3dc89f4012931ab8/lib/platform-bible-utils/src/scripture/scripture-util.ts#L298)):*
> ```XML
>   const id = Canon.bookNumberToId(bookNumber);
> 
>   if (!Intl.getCanonicalLocales(localizationLanguage)[0].startsWith('zh'))
> ```

2 · MEDIUM · lib/platform-bible-utils/src/scripture/scripture-util.ts:298

Both edits in this file are inert — dist/ wasn't rebuilt, and CI structurally cannot catch it.

package.json exports maps the package to ./dist/index.js; there's no webpack resolve.alias, no tsconfig path, and the workspace symlink still resolves through exports. The committed chunk dist/scripture-util-Df5kpddK.js has the same content hash as at the base commit and still contains startsWith(Intl.getCanonicalLocales(t)[0],"zh") and !O(l,"z").

Root npm run build doesn't include build:platform-bible-utils, so test.yml:147's "Verify no files changed after build" never sees the skew — which is why all six checks are green. Code-Style-Guide.md:306 states the convention for the sibling library with an identical setup, and 19 of the last 20 commits touching this package's src/ also committed dist/.

Fix: npm run build:pbu and commit. collectUsjMarkers (:955) is the one that matters — it runs per marker over a whole USJ document. No correctness impact; the optimization simply isn't shipping.

katherinejensen00 (approved):
> @katherinejensen00 partially reviewed 56 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2727)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #33, opened 2023-02-16
Add Code Spell Checker

irahopkinson:
- also fix pre-commit

PR #2724, opened 2026-08-26
PT-4282: Move a tab to a new window or an existing one

rolfheij-sil:
Right-clicking a tab now offers two move actions: **Move tab to new window**, and **Move tab to window ▸** listing the open windows. Choosing the window that holds the top-level menu is how a tab is docked back, which is what NN-5 asks for — Paratext 9's "Dock window" acts on the single tab under the cursor, not on a window, so there is no separate item for it and no window-level merge.

`platform.moveWebViewToWindow` was implemented on #2670 with no UI caller. This is that caller.

## The tab menu is a contribution point now

It was hard-coded in the renderer and fed by nothing, so neither the platform nor an extension could add to it. It is now a `tabMenu` channel on the existing per-web-view menu contribution system, alongside `topMenu` and `contextMenu`, riding the machinery those already use: the combiner folds the platform default in, localizes it and rejects duplicate ordering, and the data provider filters it by interface mode. No new data type, getter or subscriber — it travels on the `WebViewMenu` the renderer already subscribes to. #2670's hand-rolled "Move tab to new window" is migrated onto it, along with "Float tab".

**Unlike the top and context menus it is not opt-in.** Its items act on the tab frame rather than the web view's contents, so every tab gets them whether or not the web view asked for defaults, and a tab hosting no web view — a dialog, an error tab — gets them too. Almost no shipped web view sets `includeDefaults`, so gating the fold on it would have left the menu empty on nearly every tab.

Two things no contribution can express are resolved when the menu opens: whether an action applies to this tab, and which windows exist.

## Windows are named after what they show

A window takes the title of its first docked panel's active tab, falling back to the next tab carrying a title and then to one localized string, and publishes it as the page title — which Electron carries to the native window title with no main-process code. That closes the standing NN-3 gap where every window's OS switcher entry read "Platform.Bible", and it is what the move-to-window submenu reads back.

Duplicate names are possible and accepted, as they are in PT9, whose Window menu carries no ordinals and no de-duplication. Ordinals and monitor names were both weighed and rejected — see ADR-0034.

**Nothing in `src/main/` may call `setTitle`:** a title set from main never sticks — at construction it lasts only until the renderer publishes its first page title, and at runtime it lasts only until the renderer's next page-title change — so the main process cannot hold a name against the renderer's own naming.

Float panels are excluded from naming, so a modal dialog cannot rename the window; a maximized panel is read first, since it is what the user is looking at.

## Visibility, following PT9

[...]

PR #2728, opened 2026-08-27
PT-4278: Stop the public window API describing a single window

rolfheij-sil:
# Summary

The **documentation half of PT-4278**, plus the Linux frameless-guard portion of the chrome half, added after round 1's review. The drag-area requirement is the one piece still on the ticket; it is not touched here. See “Linux window chrome” below for what shipped and why it's unverified.

Public API docs stopped describing an application with one window, and the architecture diagram stopped drawing one renderer.

## What this changes

- **`window.service-model.ts`** — the focus-subject types said "the window" with no antecedent, which only reads correctly if there is exactly one. Both `setFocus` overloads told the reader to instruct "the window" to update the focus "in the window"; that anaphora breaks as soon as windows are plural. Each now names *a* window.
- **The service headline** said it interacts with "the current application window" without saying which, or that others exist. It now says it acts on the window it is used in, that **every window hosts its own**, and that a call from the extension host acts on the **routing target** — which is the focused window only while that window is ready and not closing, so a newly opened window can hold OS focus while calls still act on the previous one.
- **The provider-name doc** now answers per audience, because the answer genuinely differs by where you read it. From a renderer or web view, `papi.window.dataProviderName` is that window's **own scoped name**: the proxy target defines a getter returning it, and `createSyncProxyForAsyncObject`'s trap answers from the target for any property the target holds. From the extension host, and from the bare exported constant, it is the unscoped name that resolves through the routing target.
- **`subscribeFocus`'s `@param callback`** described "the updated localized menuContent for this selector", copied from the menu data service. It describes the focus subject it actually receives. The line ships at two overload sites in the generated declarations.
- **`Architecture.md`'s process diagram** drew a single Renderer box. It now draws one per window. The paragraph beneath it states the criterion that decides where a service lives — **lifetime, not subject matter** — because the previous wording said app-wide coordination lives in main and then named settings, menus and theme definitions as extension-host residents one clause later, which cannot discriminate. The Main Process box was also a column wider on its content rows than at its corners, and the four-box child row overhung it; both are squared up.

## What this adds

- **`src/shared/services/window.service.test.ts`** — pins what the provider-name doc now claims: the window-scoped name where a window id is set, the bare name where none is. There was no test on this before, and the one place that mocks `windowService` supplies the *unscoped* name — the misunderstanding the old doc created. Removing the getter, so the spread's unscoped constant answers instead, fails the first case and leaves the second passing.

## Deliberately not changed

- **`docking-framework.model.ts`** — under a standing coordination rule, that file changed only via PT-4281 until it landed.
- **`web-view.service-model.ts`** — checked and needs nothing. Every window mention in it is already plural-safe ("the window that did not answer", "a window that could not answer", "the window the user is looking at").
- No new public API, so no experimental markers are involved.
- Several pre-existing staleness items raised in review are left for follow-up: single-Renderer diagram copies in `CLAUDE.md`, two `.claude/skills/*/reference.md` files and a `.context/research/` snapshot; an `AppFocus` token in `shared-patterns.md` that never existed in code; and this file's own frontmatter and version-log markers. None is introduced or continued by this PR, and the diagram copies are a four-file duplication that fixing one of would not solve.

## UX-owned wording is on `main`, not here

This branch previously carried glossary and interaction-guideline changes routed from #2646 for UX review. **#2670 merged first and rewrote the same lines on `main`**, so both of those commits replayed to nothing during the rebase and dropped out. This PR no longer touches `lib/platform-bible-react/src/stories/guidelines/**` at all, and the current wording — windows as equal siblings, none of them a main or parent window — is #2670's.

One consequence is open rather than settled: the #2646 intent, that an action you start lands in the window you are working in, is now absent from the documentation entirely. Whether a corrected version returns — one that excludes the `openWebView` `existingId` case, which deliberately raises the *owning* window — is a UX call and sits with Rolf.

## Linux window chrome (PT-4278's chrome half)

Four guards kept Linux off the frameless title bar, each carrying the same note to re-check once
Electron 34 landed. They are lifted together, because each alone is worse than none: caption buttons
with nothing reserving room put the account icon on top of the maximize glyph, and a reserved strip
with no buttons in it is just a gap.

- The window goes frameless, as it already did on Windows and macOS.
- Native caption buttons are drawn over it, as on Windows.
- Those buttons follow the theme.
- The toolbar reserves the live-measured width they occupy.

macOS keeps its own traffic lights and stays excluded from the second and third; nothing about the
Windows or macOS path changes value, since every edit only removes `linux` from a negative
condition.

**Linux is unverified, and this is best-effort by explicit decision.** There is no Linux desktop to
try it on — WSLg is not a GNOME or KDE session — so the change is kept to the guards themselves so
that what moved is legible if it has to be undone. **CI's ubuntu smoke suite cannot validate this**:
it runs headless under Xvfb with no window manager, so a green tick there means "still launches",
not "the chrome renders correctly". The one automated check with real teeth is the new unit test
asserting Linux reserves the live-measured overlay width, mirroring the existing Windows case; it
fails against the guard and passes with it lifted.

**Not delivered here:** the ticket's second chrome item, a grabbable drag area at every zoom level.
Its premise is unverified — `-webkit-app-region` appears nowhere in the codebase — and it is a
PT-3346-class bug rather than part of this sweep.

## Verification

`lint` and `format:check` exit 0. The `src/shared` suites pass (31 files, 376 tests). `papi.d.ts` is **regenerated via `npm run build:types`, never hand-edited** — including through the rebase, where every conflict in it was resolved to the base's copy and the file regenerated afterwards rather than merged by hand, since a hand-merge of generated output produces declarations matching neither side. Its diff is wording only: filtering to non-comment lines returns nothing.

`typecheck` was run in a nested worktree, where it fails with duplicate-identifier errors from extension tsconfigs resolving through a fixed number of `../`. It fails identically on `origin/main` at the same nesting depth — the sorted error lists diff to nothing — so the branch adds no type error, but CI is the authority here rather than that local run.

AI-assisted — [session](https://claude.ai/code/session_0119Q6FZ1vLyEMRBgLikjqj9)

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2728)

### DIFF
--- a/.context/standards/Architecture.md
+++ b/.context/standards/Architecture.md
@@ -37,9 +37,11 @@ Platform.Bible uses **JSON-RPC 2.0 over WebSocket** for inter-process communicat
 ```
 
 One renderer process per window, not one for the application: a window is a full renderer hosting
-its own dock layout and its own window-scoped services, and main is the only process all of them
-have in common. Services that must answer for the whole application therefore live in main, which
-outlives any individual window.
+its own dock layout and its own window-scoped services, and main is the only process every renderer
+connects to directly. Services that must coordinate across every open window therefore live in
+main, which outlives any individual window. Other whole-application services — settings, menus,
+themes, extension lifecycle — live in the extension host, which is likewise a single process shared
+by every window (see the service tables below).
 
 ### Communication Patterns
 
--- a/.context/standards/Architecture.md
+++ b/.context/standards/Architecture.md
@@ -40,8 +40,10 @@ One renderer process per window, not one for the application: a window is a full
 its own dock layout and its own window-scoped services, and main is the only process every renderer
 connects to directly. Services that must coordinate across every open window therefore live in
 main, which outlives any individual window. Other whole-application services — settings, menus,
-themes, extension lifecycle — live in the extension host, which is likewise a single process shared
-by every window (see the service tables below).
+theme definitions, extension lifecycle — live in the extension host, which is likewise a single
+process shared by every window (see the service tables below). Which process hosts a given service
+is worth reading out of the tables rather than inferred from what it is about: the theme
+*definitions* are extension-host data, while the theme *service* is hosted elsewhere.
 
 ### Communication Patterns
 
--- a/.context/standards/Architecture.md
+++ b/.context/standards/Architecture.md
@@ -28,13 +28,18 @@ Platform.Bible uses **JSON-RPC 2.0 over WebSocket** for inter-process communicat
 │  • Routes messages between processes                     │
 └────────────────┬────────────────────────────────────────┘
                  │ JSON-RPC over WebSocket (port 8876)
-    ┌────────────┼────────────┬───────────────────┐
-    │            │            │                   │
-┌───▼────────┐ ┌─▼──────────┐ ┌▼────────────────┐
-│ Renderer   │ │ Extension  │ │ .NET Data       │
-│ (React UI) │ │ Host       │ │ Provider        │
-└────────────┘ └────────────┘ └─────────────────┘
+    ┌────────────┼─────────────┬───────────────────┐
+    │            │             │                   │
+┌───▼────────┐ ┌─▼──────────┐ ┌▼───────────┐ ┌─────▼───────────┐
+│ Renderer   │ │ Renderer   │ │ Extension  │ │ .NET Data       │
+│ (window 1) │ │ (window N) │ │ Host       │ │ Provider        │
+└────────────┘ └────────────┘ └────────────┘ └─────────────────┘
 ```
+
+One renderer process per window, not one for the application: a window is a full renderer hosting
+its own dock layout and its own window-scoped services, and main is the only process all of them
+have in common. Services that must answer for the whole application therefore live in main, which
+outlives any individual window.
 
 ### Communication Patterns
 
--- a/.context/standards/Architecture.md
+++ b/.context/standards/Architecture.md
@@ -22,11 +22,11 @@ This document provides detailed architectural information for Platform.Bible (pa
 Platform.Bible uses **JSON-RPC 2.0 over WebSocket** for inter-process communication. All processes connect to the Main process which acts as the message broker.
 
 ```
-┌─────────────────────────────────────────────────────────┐
-│                    Main Process (Electron)               │
-│  • WebSocket server on port 8876                         │
-│  • Routes messages between processes                     │
-└────────────────┬────────────────────────────────────────┘
+┌──────────────────────────────────────────────────────────────┐
+│                    Main Process (Electron)                   │
+│  • WebSocket server on port 8876                             │
+│  • Routes messages between processes                         │
+└────────────────┬─────────────────────────────────────────────┘
                  │ JSON-RPC over WebSocket (port 8876)
     ┌────────────┼─────────────┬───────────────────┐
     │            │             │                   │
@@ -37,13 +37,17 @@ Platform.Bible uses **JSON-RPC 2.0 over WebSocket** for inter-process communicat
 ```
 
 One renderer process per window, not one for the application: a window is a full renderer hosting
-its own dock layout and its own window-scoped services, and main is the only process every renderer
-connects to directly. Services that must coordinate across every open window therefore live in
-main, which outlives any individual window. Other whole-application services — settings, menus,
-theme definitions, extension lifecycle — live in the extension host, which is likewise a single
-process shared by every window (see the service tables below). Which process hosts a given service
-is worth reading out of the tables rather than inferred from what it is about: the theme
-*definitions* are extension-host data, while the theme *service* is hosted elsewhere.
+its own dock layout and its own window-scoped services. What decides where a service lives is
+lifetime, not subject matter. State that has to survive any individual window — and that two
+windows must agree on — lives in main, which outlives them all; the theme service is hosted there
+for exactly that reason. Whole-application data that no window owns — settings, menus, theme
+definitions, extension lifecycle — lives in the extension host, a single process shared by every
+window.
+
+Both of those outlive a window, so the distinction between them is not readable from what a service
+is *about*: the theme *definitions* are extension-host data while the theme *service* runs in main.
+The service tables below record where the app-global ones live; a window-scoped service is named for
+its window and lives in that window's renderer.
 
 ### Communication Patterns
 
--- a/src/shared/services/window.service-model.ts
+++ b/src/shared/services/window.service-model.ts
@@ -13,13 +13,14 @@ export const windowServiceObjectToProxy = Object.freeze({
   /**
    * JSDOC SOURCE windowServiceProviderName
    *
-   * This name is used to register the window data provider on the papi. You can use this name to
-   * find the data provider when accessing it using the useData hook
+   * This name identifies the window data provider on the papi. You can use this name to find the
+   * data provider when accessing it using the useData hook, and it resolves to the provider serving
+   * the window you are in — each window has one of its own.
    */
   dataProviderName: windowServiceProviderName,
 });
 
-/** Focus of the window is on a WebView iframe with the specified id */
+/** A window's focus is on a WebView iframe with the specified id */
 export type FocusSubjectWebView = {
   focusType: 'webView';
   /** ID of the WebView in focus (its tab ID is the same) */
@@ -27,7 +28,7 @@ export type FocusSubjectWebView = {
 };
 
 /**
- * Focus of the window is somewhere in a tab (header, toolbar, menu, content, etc.)
+ * A window's focus is somewhere in a tab (header, toolbar, menu, content, etc.)
  *
  * Note that the focused tab could be a WebView, in which case the tab is focused but it is not
  * focused in the WebView's iframe
@@ -40,12 +41,12 @@ export type FocusSubjectTab = {
   id: string;
 };
 
-/** Focus of the window is somewhere not in a tab (app menu, app toolbar, etc.) */
+/** A window's focus is somewhere not in a tab (app menu, app toolbar, etc.) */
 export type FocusSubjectOther = {
   focusType: 'other';
 };
 
-/** Current item that is the subject of top-level focus in the window */
+/** Current item that is the subject of top-level focus in a window */
 export type FocusSubject = FocusSubjectWebView | FocusSubjectTab | FocusSubjectOther;
 
 /**
@@ -111,10 +112,10 @@ export type AppWindowInputEvent = {
  */
 export const EVENT_NAME_ON_DID_APP_WINDOW_INPUT = 'platform.onDidAppWindowInput';
 
-/** Specific item that is intended to be focused in the top-level app window */
+/** Specific item that is intended to be focused at the top level of a window */
 export type SetFocusSubject = FocusSubjectWebView | Omit<FocusSubjectTab, 'tabType'>;
 
-/** Instructions that indicate how to change the focus within the window */
+/** Instructions that indicate how to change the focus within a window */
 export type SetFocusSpecifier = SetFocusSubject | DirectionFromTab | 'detect' | undefined;
 
 // Data Type to initialize data provider engine with
@@ -131,7 +132,8 @@ declare module 'papi-shared-types' {
 /**
  * JSDOC SOURCE windowService
  *
- * Service that allows to interact with the current application window
+ * Service for interacting with the application window it is used in. Every window hosts its own, so
+ * a call acts on the caller's own window rather than on one app-wide window.
  */
 export type IWindowService = {
   /**
@@ -149,7 +151,7 @@ export type IWindowService = {
    * Sets the subject of focus in the current window.
    *
    * @param focusSubject What to set the current window's focus to. Provide `'detect'` to instruct
-   *   the window to update the current focus based on what is actually focused in the window (only
+   *   that window to update its current focus based on what is actually focused in it (only
    *   necessary when an action happens that changes the focus but the window service does not
    *   detect already). In most cases, you will not need to set `'detect'` manually.
    * @returns `true` or an array of strings if the focus successfully updated; `false` otherwise
@@ -163,7 +165,7 @@ export type IWindowService = {
    *
    * @param selector `undefined`. Does not have to be provided
    * @param focusSubject What to set the current window's focus to. Provide `'detect'` to instruct
-   *   the window to update the current focus based on what is actually focused in the window (only
+   *   that window to update its current focus based on what is actually focused in it (only
    *   necessary when an action happens that changes the focus but the window service does not
    *   detect already). In most cases, you will not need to set `'detect'` manually.
    *
--- a/src/shared/services/window.service-model.ts
+++ b/src/shared/services/window.service-model.ts
@@ -13,9 +13,11 @@ export const windowServiceObjectToProxy = Object.freeze({
   /**
    * JSDOC SOURCE windowServiceProviderName
    *
-   * This name identifies the window data provider on the papi. You can use this name to find the
-   * data provider when accessing it using the useData hook, and it resolves to the provider serving
-   * the window you are in — each window has one of its own.
+   * This name identifies the window data provider on the papi. Each window registers a provider of
+   * its own under a window-scoped name, and this unscoped name resolves to a proxy that forwards to
+   * whichever window currently has focus. So finding the data provider by this name — with the
+   * useData hook, for instance — answers for the focused window, which is not necessarily the
+   * window you are in. Use `papi.window` to reach the caller's own window.
    */
   dataProviderName: windowServiceProviderName,
 });
@@ -132,8 +134,9 @@ declare module 'papi-shared-types' {
 /**
  * JSDOC SOURCE windowService
  *
- * Service for interacting with the application window it is used in. Every window hosts its own, so
- * a call acts on the caller's own window rather than on one app-wide window.
+ * Service for interacting with an application window. Every window hosts its own, so a call from a
+ * renderer acts on the window it runs in. The extension host is in no window, so a call made there
+ * acts on whichever window has focus at that moment, which can differ between two calls.
  */
 export type IWindowService = {
   /**
--- a/src/shared/services/window.service-model.ts
+++ b/src/shared/services/window.service-model.ts
@@ -13,11 +13,19 @@ export const windowServiceObjectToProxy = Object.freeze({
   /**
    * JSDOC SOURCE windowServiceProviderName
    *
-   * This name identifies the window data provider on the papi. Each window registers a provider of
-   * its own under a window-scoped name, and this unscoped name resolves to a proxy that forwards to
-   * whichever window currently has focus. So finding the data provider by this name — with the
-   * useData hook, for instance — answers for the focused window, which is not necessarily the
-   * window you are in. Use `papi.window` to reach the caller's own window.
+   * This name identifies the window data provider on the papi. Every window registers a provider of
+   * its own under a window-scoped name — this name with the window's id appended — and what you get
+   * from this property depends on where you read it.
+   *
+   * From a renderer or a web view, it is that window's own scoped name, so the provider found by it
+   * — with the useData hook, for instance — both reports and changes the focus of the window you
+   * are in. Read it from `papi.window` and use it as it comes.
+   *
+   * From the extension host, which runs in no window, it is the bare unscoped name. That name
+   * resolves to whichever window the router is currently targeting, so two reads can answer for
+   * different windows. The bare {@link windowServiceProviderName} constant behaves the same way
+   * wherever it is imported. To act on one particular window from there,
+   * `platform.getFocusedWindowId` reports which window has focus.
    */
   dataProviderName: windowServiceProviderName,
 });
@@ -136,7 +144,13 @@ declare module 'papi-shared-types' {
  *
  * Service for interacting with an application window. Every window hosts its own, so a call from a
  * renderer acts on the window it runs in. The extension host is in no window, so a call made there
- * acts on whichever window has focus at that moment, which can differ between two calls.
+ * acts on whichever window the router is targeting at that moment, which can differ between two
+ * calls.
+ *
+ * The routing target is usually the focused window, but not always: a window that has taken OS
+ * focus does not become the target until it is ready and not closing, so a newly opened window can
+ * hold focus while calls still act on the previous one. `platform.getFocusedWindowId` is what
+ * tracks focus itself.
  */
 export type IWindowService = {
   /**
@@ -186,10 +200,10 @@ export type IWindowService = {
    * Subscribe to run a callback function when the current window's subject of focus is changed
    *
    * @param selector `undefined`. Does not have to be provided
-   * @param callback Function to run with the updated localized menuContent for this selector. If
-   *   there is an error while retrieving the updated data, the function will run with a
-   *   {@link PlatformError} instead of the data. You can call {@link isPlatformError} on this value
-   *   to check if it is an error.
+   * @param callback Function to run with the window's updated subject of focus. If there is an
+   *   error while retrieving the updated data, the function will run with a {@link PlatformError}
+   *   instead of the data. You can call {@link isPlatformError} on this value to check if it is an
+   *   error.
    * @param options Various options to adjust how the subscriber emits updates
    * @returns Unsubscriber function (run to unsubscribe from listening for updates)
    */
--- a/src/shared/services/window.service-model.ts
+++ b/src/shared/services/window.service-model.ts
@@ -143,14 +143,17 @@ declare module 'papi-shared-types' {
  * JSDOC SOURCE windowService
  *
  * Service for interacting with an application window. Every window hosts its own, so a call from a
- * renderer acts on the window it runs in. The extension host is in no window, so a call made there
- * acts on whichever window the router is targeting at that moment, which can differ between two
- * calls.
+ * renderer acts on the window it runs in.
  *
- * The routing target is usually the focused window, but not always: a window that has taken OS
- * focus does not become the target until it is ready and not closing, so a newly opened window can
- * hold focus while calls still act on the previous one. `platform.getFocusedWindowId` is what
- * tracks focus itself.
+ * The extension host is in no window, so a call made there acts on the window that most recently
+ * had focus — `platform.getFocusedWindowId`, which stays set while the application is in the
+ * background. Two calls can answer for different windows, and a subscription binds to the window
+ * focused when it was made rather than following focus afterwards. If there is no focused window,
+ * or the focused window has not registered its window service — either because it is still starting
+ * or because it has just gone away — the call throws rather than falling back to another window.
+ *
+ * This is a different resolver from the one the {@link windowServiceProviderName} doc describes: the
+ * bare unscoped name goes through the router; `papi.window` does not.
  */
 export type IWindowService = {
   /**
--- a/src/shared/services/window.service-model.ts
+++ b/src/shared/services/window.service-model.ts
@@ -152,8 +152,8 @@ declare module 'papi-shared-types' {
  * or the focused window has not registered its window service — either because it is still starting
  * or because it has just gone away — the call throws rather than falling back to another window.
  *
- * This is a different resolver from the one the {@link windowServiceProviderName} doc describes: the
- * bare unscoped name goes through the router; `papi.window` does not.
+ * This is a different resolver from the one the `windowServiceProviderName` doc describes: the bare
+ * unscoped name goes through the router; `papi.window` does not.
  */
 export type IWindowService = {
   /**
--- a/lib/papi-dts/papi.d.ts
+++ b/lib/papi-dts/papi.d.ts
@@ -4564,26 +4564,32 @@ declare module 'shared/services/window.service-model' {
   import { DirectionFromTab } from 'shared/models/docking-framework.model';
   /**
    *
-   * This name is used to register the window data provider on the papi. You can use this name to
-   * find the data provider when accessing it using the useData hook
+   * This name identifies the window data provider on the papi. Each window registers a provider of
+   * its own under a window-scoped name, and this unscoped name resolves to a proxy that forwards to
+   * whichever window currently has focus. So finding the data provider by this name — with the
+   * useData hook, for instance — answers for the focused window, which is not necessarily the
+   * window you are in. Use `papi.window` to reach the caller's own window.
    */
   export const windowServiceProviderName = 'platform.windowServiceDataProvider';
   export const windowServiceObjectToProxy: Readonly<{
     /**
      *
-     * This name is used to register the window data provider on the papi. You can use this name to
-     * find the data provider when accessing it using the useData hook
+     * This name identifies the window data provider on the papi. Each window registers a provider of
+     * its own under a window-scoped name, and this unscoped name resolves to a proxy that forwards to
+     * whichever window currently has focus. So finding the data provider by this name — with the
+     * useData hook, for instance — answers for the focused window, which is not necessarily the
+     * window you are in. Use `papi.window` to reach the caller's own window.
      */
     dataProviderName: 'platform.windowServiceDataProvider';
   }>;
-  /** Focus of the window is on a WebView iframe with the specified id */
+  /** A window's focus is on a WebView iframe with the specified id */
   export type FocusSubjectWebView = {
     focusType: 'webView';
     /** ID of the WebView in focus (its tab ID is the same) */
     id: string;
   };
   /**
-   * Focus of the window is somewhere in a tab (header, toolbar, menu, content, etc.)
+   * A window's focus is somewhere in a tab (header, toolbar, menu, content, etc.)
    *
    * Note that the focused tab could be a WebView, in which case the tab is focused but it is not
    * focused in the WebView's iframe
@@ -4595,11 +4601,11 @@ declare module 'shared/services/window.service-model' {
     /** ID of the tab in focus (if this is a WebView, its WebView ID is the same) */
     id: string;
   };
-  /** Focus of the window is somewhere not in a tab (app menu, app toolbar, etc.) */
+  /** A window's focus is somewhere not in a tab (app menu, app toolbar, etc.) */
   export type FocusSubjectOther = {
     focusType: 'other';
   };
-  /** Current item that is the subject of top-level focus in the window */
+  /** Current item that is the subject of top-level focus in a window */
   export type FocusSubject = FocusSubjectWebView | FocusSubjectTab | FocusSubjectOther;
   /**
    * Gets the id of the web view a focus subject refers to, if it refers to one: either the web view
@@ -4655,9 +4661,9 @@ declare module 'shared/services/window.service-model' {
    * surveil user input. Do not broaden what is announced here without a security review.
    */
   export const EVENT_NAME_ON_DID_APP_WINDOW_INPUT = 'platform.onDidAppWindowInput';
-  /** Specific item that is intended to be focused in the top-level app window */
+  /** Specific item that is intended to be focused at the top level of a window */
   export type SetFocusSubject = FocusSubjectWebView | Omit<FocusSubjectTab, 'tabType'>;
-  /** Instructions that indicate how to change the focus within the window */
+  /** Instructions that indicate how to change the focus within a window */
   export type SetFocusSpecifier = SetFocusSubject | DirectionFromTab | 'detect' | undefined;
   export type WindowDataTypes = {
     Focus: DataProviderDataType<undefined, FocusSubject | undefined, SetFocusSpecifier>;
@@ -4669,7 +4675,9 @@ declare module 'shared/services/window.service-model' {
   }
   /**
    *
-   * Service that allows to interact with the current application window
+   * Service for interacting with an application window. Every window hosts its own, so a call from a
+   * renderer acts on the window it runs in. The extension host is in no window, so a call made there
+   * acts on whichever window has focus at that moment, which can differ between two calls.
    */
   export type IWindowService = {
     /**
@@ -4692,7 +4700,7 @@ declare module 'shared/services/window.service-model' {
      * Sets the subject of focus in the current window.
      *
      * @param focusSubject What to set the current window's focus to. Provide `'detect'` to instruct
-     *   the window to update the current focus based on what is actually focused in the window (only
+     *   that window to update its current focus based on what is actually focused in it (only
      *   necessary when an action happens that changes the focus but the window service does not
      *   detect already). In most cases, you will not need to set `'detect'` manually.
      * @returns `true` or an array of strings if the focus successfully updated; `false` otherwise
@@ -4706,7 +4714,7 @@ declare module 'shared/services/window.service-model' {
      *
      * @param selector `undefined`. Does not have to be provided
      * @param focusSubject What to set the current window's focus to. Provide `'detect'` to instruct
-     *   the window to update the current focus based on what is actually focused in the window (only
+     *   that window to update its current focus based on what is actually focused in it (only
      *   necessary when an action happens that changes the focus but the window service does not
      *   detect already). In most cases, you will not need to set `'detect'` manually.
      *
@@ -12718,7 +12726,9 @@ declare module '@papi/backend' {
     notifications: INotificationService;
     /**
      *
-     * Service that allows to interact with the current application window
+     * Service for interacting with an application window. Every window hosts its own, so a call from a
+     * renderer acts on the window it runs in. The extension host is in no window, so a call made there
+     * acts on whichever window has focus at that moment, which can differ between two calls.
      */
     window: IWindowService;
   };
@@ -12976,7 +12986,9 @@ declare module '@papi/backend' {
   export const notifications: INotificationService;
   /**
    *
-   * Service that allows to interact with the current application window
+   * Service for interacting with an application window. Every window hosts its own, so a call from a
+   * renderer acts on the window it runs in. The extension host is in no window, so a call made there
+   * acts on whichever window has focus at that moment, which can differ between two calls.
    */
   export const window: IWindowService;
 }
@@ -13579,7 +13591,9 @@ declare module '@papi/frontend' {
     notifications: INotificationService;
     /**
      *
-     * Service that allows to interact with the current application window
+     * Service for interacting with an application window. Every window hosts its own, so a call from a
+     * renderer acts on the window it runs in. The extension host is in no window, so a call made there
+     * acts on whichever window has focus at that moment, which can differ between two calls.
      */
     window: IWindowService;
     /**
@@ -13748,7 +13762,9 @@ declare module '@papi/frontend' {
   export const notifications: INotificationService;
   /**
    *
-   * Service that allows to interact with the current application window
+   * Service for interacting with an application window. Every window hosts its own, so a call from a
+   * renderer acts on the window it runs in. The extension host is in no window, so a call made there
+   * acts on whichever window has focus at that moment, which can differ between two calls.
    */
   export const window: IWindowService;
   /**
--- a/lib/papi-dts/papi.d.ts
+++ b/lib/papi-dts/papi.d.ts
@@ -4564,21 +4564,37 @@ declare module 'shared/services/window.service-model' {
   import { DirectionFromTab } from 'shared/models/docking-framework.model';
   /**
    *
-   * This name identifies the window data provider on the papi. Each window registers a provider of
-   * its own under a window-scoped name, and this unscoped name resolves to a proxy that forwards to
-   * whichever window currently has focus. So finding the data provider by this name — with the
-   * useData hook, for instance — answers for the focused window, which is not necessarily the
-   * window you are in. Use `papi.window` to reach the caller's own window.
+   * This name identifies the window data provider on the papi. Every window registers a provider of
+   * its own under a window-scoped name — this name with the window's id appended — and what you get
+   * from this property depends on where you read it.
+   *
+   * From a renderer or a web view, it is that window's own scoped name, so the provider found by it
+   * — with the useData hook, for instance — both reports and changes the focus of the window you
+   * are in. Read it from `papi.window` and use it as it comes.
+   *
+   * From the extension host, which runs in no window, it is the bare unscoped name. That name
+   * resolves to whichever window the router is currently targeting, so two reads can answer for
+   * different windows. The bare {@link windowServiceProviderName} constant behaves the same way
+   * wherever it is imported. To act on one particular window from there,
+   * `platform.getFocusedWindowId` reports which window has focus.
    */
   export const windowServiceProviderName = 'platform.windowServiceDataProvider';
   export const windowServiceObjectToProxy: Readonly<{
     /**
      *
-     * This name identifies the window data provider on the papi. Each window registers a provider of
-     * its own under a window-scoped name, and this unscoped name resolves to a proxy that forwards to
-     * whichever window currently has focus. So finding the data provider by this name — with the
-     * useData hook, for instance — answers for the focused window, which is not necessarily the
-     * window you are in. Use `papi.window` to reach the caller's own window.
+     * This name identifies the window data provider on the papi. Every window registers a provider of
+     * its own under a window-scoped name — this name with the window's id appended — and what you get
+     * from this property depends on where you read it.
+     *
+     * From a renderer or a web view, it is that window's own scoped name, so the provider found by it
+     * — with the useData hook, for instance — both reports and changes the focus of the window you
+     * are in. Read it from `papi.window` and use it as it comes.
+     *
+     * From the extension host, which runs in no window, it is the bare unscoped name. That name
+     * resolves to whichever window the router is currently targeting, so two reads can answer for
+     * different windows. The bare {@link windowServiceProviderName} constant behaves the same way
+     * wherever it is imported. To act on one particular window from there,
+     * `platform.getFocusedWindowId` reports which window has focus.
      */
     dataProviderName: 'platform.windowServiceDataProvider';
   }>;
@@ -4677,7 +4693,13 @@ declare module 'shared/services/window.service-model' {
    *
    * Service for interacting with an application window. Every window hosts its own, so a call from a
    * renderer acts on the window it runs in. The extension host is in no window, so a call made there
-   * acts on whichever window has focus at that moment, which can differ between two calls.
+   * acts on whichever window the router is targeting at that moment, which can differ between two
+   * calls.
+   *
+   * The routing target is usually the focused window, but not always: a window that has taken OS
+   * focus does not become the target until it is ready and not closing, so a newly opened window can
+   * hold focus while calls still act on the previous one. `platform.getFocusedWindowId` is what
+   * tracks focus itself.
    */
   export type IWindowService = {
     /**
@@ -4732,10 +4754,10 @@ declare module 'shared/services/window.service-model' {
      * Subscribe to run a callback function when the current window's subject of focus is changed
      *
      * @param selector `undefined`. Does not have to be provided
-     * @param callback Function to run with the updated localized menuContent for this selector. If
-     *   there is an error while retrieving the updated data, the function will run with a
-     *   {@link PlatformError} instead of the data. You can call {@link isPlatformError} on this value
-     *   to check if it is an error.
+     * @param callback Function to run with the window's updated subject of focus. If there is an
+     *   error while retrieving the updated data, the function will run with a {@link PlatformError}
+     *   instead of the data. You can call {@link isPlatformError} on this value to check if it is an
+     *   error.
      * @param options Various options to adjust how the subscriber emits updates
      * @returns Unsubscriber function (run to unsubscribe from listening for updates)
      */
@@ -12728,7 +12750,13 @@ declare module '@papi/backend' {
      *
      * Service for interacting with an application window. Every window hosts its own, so a call from a
      * renderer acts on the window it runs in. The extension host is in no window, so a call made there
-     * acts on whichever window has focus at that moment, which can differ between two calls.
+     * acts on whichever window the router is targeting at that moment, which can differ between two
+     * calls.
+     *
+     * The routing target is usually the focused window, but not always: a window that has taken OS
+     * focus does not become the target until it is ready and not closing, so a newly opened window can
+     * hold focus while calls still act on the previous one. `platform.getFocusedWindowId` is what
+     * tracks focus itself.
      */
     window: IWindowService;
   };
@@ -12988,7 +13016,13 @@ declare module '@papi/backend' {
    *
    * Service for interacting with an application window. Every window hosts its own, so a call from a
    * renderer acts on the window it runs in. The extension host is in no window, so a call made there
-   * acts on whichever window has focus at that moment, which can differ between two calls.
+   * acts on whichever window the router is targeting at that moment, which can differ between two
+   * calls.
+   *
+   * The routing target is usually the focused window, but not always: a window that has taken OS
+   * focus does not become the target until it is ready and not closing, so a newly opened window can
+   * hold focus while calls still act on the previous one. `platform.getFocusedWindowId` is what
+   * tracks focus itself.
    */
   export const window: IWindowService;
 }
@@ -13593,7 +13627,13 @@ declare module '@papi/frontend' {
      *
      * Service for interacting with an application window. Every window hosts its own, so a call from a
      * renderer acts on the window it runs in. The extension host is in no window, so a call made there
-     * acts on whichever window has focus at that moment, which can differ between two calls.
+     * acts on whichever window the router is targeting at that moment, which can differ between two
+     * calls.
+     *
+     * The routing target is usually the focused window, but not always: a window that has taken OS
+     * focus does not become the target until it is ready and not closing, so a newly opened window can
+     * hold focus while calls still act on the previous one. `platform.getFocusedWindowId` is what
+     * tracks focus itself.
      */
     window: IWindowService;
     /**
@@ -13764,7 +13804,13 @@ declare module '@papi/frontend' {
    *
    * Service for interacting with an application window. Every window hosts its own, so a call from a
    * renderer acts on the window it runs in. The extension host is in no window, so a call made there
-   * acts on whichever window has focus at that moment, which can differ between two calls.
+   * acts on whichever window the router is targeting at that moment, which can differ between two
+   * calls.
+   *
+   * The routing target is usually the focused window, but not always: a window that has taken OS
+   * focus does not become the target until it is ready and not closing, so a newly opened window can
+   * hold focus while calls still act on the previous one. `platform.getFocusedWindowId` is what
+   * tracks focus itself.
    */
   export const window: IWindowService;
   /**
--- a/lib/papi-dts/papi.d.ts
+++ b/lib/papi-dts/papi.d.ts
@@ -4701,8 +4701,8 @@ declare module 'shared/services/window.service-model' {
    * or the focused window has not registered its window service — either because it is still starting
    * or because it has just gone away — the call throws rather than falling back to another window.
    *
-   * This is a different resolver from the one the {@link windowServiceProviderName} doc describes: the
-   * bare unscoped name goes through the router; `papi.window` does not.
+   * This is a different resolver from the one the `windowServiceProviderName` doc describes: the bare
+   * unscoped name goes through the router; `papi.window` does not.
    */
   export type IWindowService = {
     /**
@@ -12761,8 +12761,8 @@ declare module '@papi/backend' {
      * or the focused window has not registered its window service — either because it is still starting
      * or because it has just gone away — the call throws rather than falling back to another window.
      *
-     * This is a different resolver from the one the {@link windowServiceProviderName} doc describes: the
-     * bare unscoped name goes through the router; `papi.window` does not.
+     * This is a different resolver from the one the `windowServiceProviderName` doc describes: the bare
+     * unscoped name goes through the router; `papi.window` does not.
      */
     window: IWindowService;
   };
@@ -13030,8 +13030,8 @@ declare module '@papi/backend' {
    * or the focused window has not registered its window service — either because it is still starting
    * or because it has just gone away — the call throws rather than falling back to another window.
    *
-   * This is a different resolver from the one the {@link windowServiceProviderName} doc describes: the
-   * bare unscoped name goes through the router; `papi.window` does not.
+   * This is a different resolver from the one the `windowServiceProviderName` doc describes: the bare
+   * unscoped name goes through the router; `papi.window` does not.
    */
   export const window: IWindowService;
 }
@@ -13644,8 +13644,8 @@ declare module '@papi/frontend' {
      * or the focused window has not registered its window service — either because it is still starting
      * or because it has just gone away — the call throws rather than falling back to another window.
      *
-     * This is a different resolver from the one the {@link windowServiceProviderName} doc describes: the
-     * bare unscoped name goes through the router; `papi.window` does not.
+     * This is a different resolver from the one the `windowServiceProviderName` doc describes: the bare
+     * unscoped name goes through the router; `papi.window` does not.
      */
     window: IWindowService;
     /**
@@ -13824,8 +13824,8 @@ declare module '@papi/frontend' {
    * or the focused window has not registered its window service — either because it is still starting
    * or because it has just gone away — the call throws rather than falling back to another window.
    *
-   * This is a different resolver from the one the {@link windowServiceProviderName} doc describes: the
-   * bare unscoped name goes through the router; `papi.window` does not.
+   * This is a different resolver from the one the `windowServiceProviderName` doc describes: the bare
+   * unscoped name goes through the router; `papi.window` does not.
    */
   export const window: IWindowService;
   /**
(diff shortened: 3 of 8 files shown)


.context/standards/Architecture.md
```diff
@@ -28,14 +28,23 @@ Platform.Bible uses **JSON-RPC 2.0 over WebSocket** for inter-process communicat
...
                  │ JSON-RPC over WebSocket (port 8876)
-    ┌────────────┼────────────┬───────────────────┐
-    │            │            │                   │
-┌───▼────────┐ ┌─▼──────────┐ ┌▼────────────────┐
-│ Renderer   │ │ Extension  │ │ .NET Data       │
-│ (React UI) │ │ Host       │ │ Provider        │
-└────────────┘ └────────────┘ └─────────────────┘
+    ┌────────────┼─────────────┬───────────────────┐
+    │            │             │                   │
+┌───▼────────┐ ┌─▼──────────┐ ┌▼───────────┐ ┌─────▼───────────┐
```
lyonsil, line 33: **#12** - low · downgraded on review

The rebuilt four-box child row now overhangs the Main Process box it descends from by five columns.

The new row is 64 characters wide against the 59-character Main Process box; the old three-box row was 49 and sat well inside it. Purely visual, inside a fenced code block. The same rewrite genuinely fixed the connector alignment, which was off by one at the base.

_Why this is downgraded: Measured with a UTF-8-aware script (character length, not bytes). At head, Architecture.md:33-36 (the four-box child row added by this PR) is 64 characters wide, while the Main Process box at line 25 is 59 - a 5-column overhang, exactly as claimed. At the merge base the old three-box row was 49 characters, well inside the 59-column box, confirming the overhang is new. The connector-alignment concession also checks out: at base the third branch point was offset by 1 from its arrow, whereas at head all four branch points (columns 4, 17, 31, 51) align exactly with their four arrows. All arithmetic in the finding is accurate, but this is a purely visual ASCII-art misalignment inside a fenced code block with no functional impact._

_Other findings in this file: #13, #14, #15, #16_

.context/standards/Architecture.md
```diff
@@ -28,14 +28,23 @@ Platform.Bible uses **JSON-RPC 2.0 over WebSocket** for inter-process communicat
...
+│ Renderer   │ │ Renderer   │ │ Extension  │ │ .NET Data       │
+│ (window 1) │ │ (window N) │ │ Host       │ │ Provider        │
+└────────────┘ └────────────┘ └────────────┘ └─────────────────┘
 ```
 
+One renderer process per window, not one for the application: a window is a full renderer hosting
+its own dock layout and its own window-scoped services, and main is the only process every renderer
```
lyonsil: **#13** - low · downgraded on review

The clause "main is the only process every renderer connects to directly" restates line 22 and sits in a slot that reads as a premise but carries no weight.

Reachability does not follow socket topology - the broker is precisely why a renderer's `papi.settings` / `papi.menus` reach the extension host over its single socket. The clause adds nothing line 22 and the diagram have not already said.

_Why this is downgraded: Both halves of the mechanics check out. The clause is redundant with line 22 ("All processes connect to the Main process which acts as the message broker") plus the diagram at lines 24-37, and reachability genuinely does not follow socket topology: renderers open exactly one socket to main's WebSocketServer (rpc-client.ts:113, rpc-websocket-listener.ts:120), yet papi.settings/papi.menus are hosted in the extension host and reached because "main re-dispatches the incoming request" (network.service.ts:382). But the alleged error - that the paragraph positions topology as the premise for "therefore live in main" - does not survive reading the whole paragraph: the therefore-sentence supplies its own premise in its trailing clause ("which outlives any individual window", line 42), exactly the lifetime reason lines 121-123 give, and the next sentence places settings, menus, theme definitions and extension lifecycle in the extension host, "likewise a single process shared by every window". No reader comes away believing only main is reachable. What survives is the redundancy half only._

Same fix as #14, #15, #16.

_Other findings in this file: #12, #14, #15, #16_

.context/standards/Architecture.md
```diff
@@ -28,14 +28,23 @@ Platform.Bible uses **JSON-RPC 2.0 over WebSocket** for inter-process communicat
...
+connects to directly. Services that must coordinate across every open window therefore live in
```
lyonsil: **#14** - medium -> low · downgraded on review

"Services that must coordinate across every open window therefore live in main" does not discriminate, since the next clause puts settings, menus and theme definitions in the extension host.

Those also coordinate across every open window, so the stated criterion cannot be the real one. The paragraph disarms itself two clauses later by disclaiming that it is a siting rule, and the actual criterion - lifetime and unique-name ownership - is stated further down the same file but not linked from here.

_Why this is downgraded: The mechanics check out: Architecture.md:41-43 is added by this PR, and it does say app-wide coordination "therefore live[s] in main" one clause before listing settings, menus and theme definitions as extension-host residents - services every window also consumes - so that sentence alone does not discriminate. But the same paragraph disarms itself two clauses later at lines 44-46 ("which process hosts a given service is worth reading out of the tables rather than inferred from what it is about"), explicitly disclaiming that it is a siting rule, which is the harm the finding predicts. The competing rule is also weaker than quoted: shared-patterns.md:54 is untouched by this PR, sits under "DataProvider Engine Hosting", and in the same sentence chain already carves out the extension host before enumerating what app-global means for main; its `paths:` frontmatter is src/shared/** + src/node/** only, so it does not auto-load while writing src/main/, src/renderer/ or src/extension-host/ service hosts. The real criterion is stated at Architecture.md:113-122 but unlinked from the new paragraph. A one-sentence wording wobble worth tightening, not a doc-vs-rule contradiction._

Same fix as #13, #15, #16.

_Other findings in this file: #12, #13, #15, #16_

.context/standards/Architecture.md
```diff
@@ -28,14 +28,23 @@ Platform.Bible uses **JSON-RPC 2.0 over WebSocket** for inter-process communicat
...
+main, which outlives any individual window. Other whole-application services — settings, menus,
+theme definitions, extension lifecycle — live in the extension host, which is likewise a single
+process shared by every window (see the service tables below). Which process hosts a given service
```
lyonsil: **#15** - medium -> low · downgraded on review

"worth reading out of the tables" points at tables that omit most of what it asks the reader to look up.

The Main Process Services table lists 7 rows against 26 files in src/main/services/, omitting all six `*.service-router.ts` plus window-state.service.ts, and there is no per-service renderer table at all - so the window-scoped services the same new paragraph introduces are in no per-process table.

_Why this is downgraded: The counts check out: the sentence is new (lines 39-46 are wholly added), the Main Process Services table at lines 184-192 lists exactly 7 rows, and `ls src/main/services/` returns 26 files - all six *.service-router.ts, window-state.service.ts, project-lookup.service-host.ts, shutdown-latch.service.ts, network-object-status.service-host.ts and window-layout-persistence.service.ts are absent, and no per-service renderer table exists. But "tables that cannot answer the question" is an over-reading on three counts. The sentence's own worked example resolves cleanly from the two tables it cites (theme-data.service-host.ts in the Extension Host table at line 216, theme.service-host.ts in the Main table at line 190). The window-scoped services the paragraph introduces are covered by a table - the router/shard table at lines 95-97 states by file suffix that *.service-router.ts lives in main and *.service-shard.ts in each renderer, generalizing over all six rather than enumerating them. And `git show dc97223:...` proves the Main Process Services table is byte-identical to base, so the enumeration gap is entirely pre-existing; this PR only adds a pointer to it._

Same fix as #13, #14, #16.

_Other findings in this file: #12, #13, #14, #16_

.context/standards/Architecture.md
```diff
@@ -28,14 +28,23 @@ Platform.Bible uses **JSON-RPC 2.0 over WebSocket** for inter-process communicat
...
+is worth reading out of the tables rather than inferred from what it is about: the theme
```
lyonsil: **#16** - medium -> low · downgraded on review

"the theme *service* is hosted elsewhere" withholds an answer the same file states plainly three times.

Architecture.md says at lines 120-121 and 169-171, and in the Main Process Services table at line 193, that the theme service lives in src/main/services/theme.service-host.ts. Writing "hosted in main" instead of "hosted elsewhere" is the whole fix.

_Why this is downgraded: Part (b) checks out: the file states the answer plainly at Architecture.md:120-121 ("the two app-global services got there by being hosted in main - src/main/services/theme.service-host.ts"), at 169-171 ("both now live in src/main/services/"), and by listing theme.service-host.ts in the Main Process Services table at 193 - so "hosted elsewhere" is coyer than the rest of the file, though the same new paragraph supplies the rule that resolves it two sentences earlier, so nothing is actually denied the reader. Part (a) does not survive: the service tables are organized BY PROCESS ("### Main Process Services" at 184, "### Extension Host Services" at 207), so reading a row out of the tables answers the hosting question by which table it sits in, and the worked example lands exactly as the sentence promises; the vague "Theme management" purpose text is a Purpose-column nit, not a hosting ambiguity, and that row is pre-existing and untouched. What survives is a one-word prose polish on an added line._

Same fix as #13, #14, #15.

_Other findings in this file: #12, #13, #14, #15_

lib/platform-bible-react/src/stories/guidelines/terms.mdx
```diff
@@ -4,14 +4,14 @@ import { Meta } from '@storybook/addon-docs/blocks';
 
 # Terminology
 
-- `Window`: An operating system window that contains some interface. We currently support having only one of these in Platform.Bible (and therefore Paratext 10 Studio) right now. We call it the app window.
+- `Window`: An operating system window that contains some interface. In Power mode, Platform.Bible supports having more than one open at a time. Actions you start — opening settings, a dialog, a notification — happen in the window you are working in.
```
lyonsil: **#26** - high -> low · downgraded on review

"Actions you start ... happen in the window you are working in" is over-broad: an `openWebView` with an existing id deliberately opens in and raises the OWNING window.

Get Resources, Find, Comments and Manage Books all pass an existing web view id, and openWebViewInOwningWindow calls focusWindow(owner.windowId) when the owner is not the routing target - a documented, intentional design. The settings example this was originally reported against does not do this: openSettingsForWebView calls the shard's openSettingsTab directly and never raises a window. The dialog and notification halves of the sentence hold outright.

_Why this is downgraded: The cited line numbers are right (routing: 'owner' for platform.openSettings at web-view.service-router.ts:731; the focusWindow(owner.windowId) at :338), but they belong to two unconnected paths: openSettingsForWebView (web-view.service-router.ts:629-650) calls owner.shard.openSettingsTab(...) directly and never touches openWebViewInOwningWindow, and focusWindow has exactly one production call site - line 338, reachable only from openWebView with existingId. So the claim's headline, "opening settings ... raises that other window", is false: settings never raises any window, and the renderer shard's openSettingsTab (web-view.service-shard.ts:3025) only floats a tab inside its own window. Every production caller of platform.openSettings passes a web view id from the window the user just acted in, and the dialog and notification halves hold outright (both routers are focus-routed). What survives is smaller and differently grounded: openWebView with an existingId (Get Resources at platform-get-resources/src/main.ts:237, Find/Comments/Manage Books likewise) does open in and deliberately raise the owning window - a documented, intentional design, making the blanket sentence over-broad._

_Other findings in this file: #27_

---

**#27** - medium -> low · downgraded on review

The glossary's first entry is gated on "Power mode", a term the terminology material never defines.

A grep of the whole stories tree finds "Power mode" exactly once - on this line - with no hits for "interface mode", "interfaceMode" or "Simple mode" anywhere in its 62 files. The term IS defined to users in-product, so the reader is not stranded; a one-clause inline gloss, or `Simple mode`/`Power mode` entries of their own, would close it.

_Why this is downgraded: The mechanics check out: terms.mdx:7 is a rewritten line and is the glossary's first entry, and a case-insensitive grep for "power mode" over lib/platform-bible-react/src/stories/ returns exactly one hit - that line itself - with zero hits for "interface mode", "interfaceMode" or "Simple mode" anywhere in the tree's 62 files. But the term is not undefined in any sense a reader is stuck on: "Power mode" is the verbatim user-facing localized label %userProfile_interfaceMode_power_label% (en.json:505), paired with "Simple mode" (:507) and glossed in-product at :504 and :424 for the platform.interfaceMode setting (core-settings-info.data.ts:38-42), so anyone reading this glossary reaches the definition in the app's own Interface mode setting. The page also already leans on terms it never defines - "PAPI" is used at terms.mdx:17 and defined nowhere in the stories tree - so this is not a new breach of a standard the page otherwise holds. On balance the qualifier made the entry strictly more accurate._

_Other findings in this file: #26_

src/shared/services/window.service-model.ts
```diff
@@ -13,21 +13,24 @@ export const windowServiceObjectToProxy = Object.freeze({
   /**
    * JSDOC SOURCE windowServiceProviderName
    *
-   * This name is used to register the window data provider on the papi. You can use this name to
-   * find the data provider when accessing it using the useData hook
+   * This name identifies the window data provider on the papi. Each window registers a provider of
```
lyonsil: **#1** - high · confirmed

The rewritten `dataProviderName` doc describes the opposite of what that property evaluates to in a renderer, and steers readers off the one call that does what they want.

The block calls this value "this unscoped name" that "answers for the focused window, which is not necessarily the window you are in", then tells the reader to use `papi.window` instead. But `windowServiceProxyTarget` defines a `dataProviderName` getter returning `${windowServiceProviderName}-${globalThis.windowId}` whenever `windowId` is set - every renderer and web view - and `createSyncProxyForAsyncObject` returns `target[prop]` when the prop is on the target, so the getter wins. `papi.window.dataProviderName` is therefore the caller's OWN window-scoped name, resolving to that window's own provider. The block is still accurate for the bare `windowServiceProviderName` constant and for extension-host callers, so the fix is to split the two audiences rather than revert.

_Verification: Both anchors are exact: the getter is window.service.ts:101-109, and util.ts:354 is `if (prop in target) return target[prop];` - `in` matches the accessor own property, so the trap invokes the getter rather than falling through to the async path. The object literal at window.service.ts:99-110 spreads windowServiceObjectToProxy first and defines the getter second, so the getter wins; a node probe replicating both confirmed `papi.window.dataProviderName` resolves to `platform.windowServiceDataProvider-2` with windowId set. globalThis.windowId is set for every renderer (global-this.model.ts:60) and web views resolve @papi/frontend through window.parent.webViewRequire, so they read the same renderer global. Because IWindowService intersects typeof windowServiceObjectToProxy, the block ships at papi.d.ts:10918-10927 directly on papi.window.dataProviderName, while the repo's only consumer, platform-tab-title.component.tsx:162, does exactly the `useData(windowService.dataProviderName)` call the doc warns against, to read its own window's focus. For a PR whose sole purpose is doc accuracy on this exact surface, the added text is false on the path most extension authors take._

Same fix as #2, #3, #32.

_Other findings in this file: #2, #3, #32, #33, #34, #35_

src/shared/services/window.service-model.ts
```diff
@@ -13,21 +13,24 @@ export const windowServiceObjectToProxy = Object.freeze({
...
+   * its own under a window-scoped name, and this unscoped name resolves to a proxy that forwards to
```
lyonsil: **#2** - high -> medium · downgraded on review

"forwards to whichever window currently has focus" names the wrong resolver - the router resolves by routing target, a distinction the codebase draws deliberately.

`FocusedWindowDataProviderEngine` resolves through `getTargetWindowId()`, and `getRoutingTarget()` returns the focused window only when it is ready and not closing, otherwise falling back to most-recently-focused-ready and then first-ready. Open a second window: it takes OS focus immediately but is not in `readyWindowIds` for seconds, so the generic name keeps answering for window 1 while this doc says it answers for the focused window. The precise phrasing is "the routing target"; `platform.getFocusedWindowId` is the API that genuinely tracks focus. Note there are two sites, not one - line 93 repeats it as "acts on whichever window has focus at that moment".

_Why this is downgraded: Every citation verifies. getRoutingTarget() (window-state.service.ts:314-342) returns the focused window only when readyWindowIds.has(id) && !closingWindowIds.has(id). The distinction is deliberate and load-bearing: window-state.service.ts:280 calls getFocusedWindowId "the honest answer about focus, which is not always where calls are routed"; main.ts:361-363 says routing "waits for it rather than following focus alone"; adr-generic-name-routing-proxies says routers use "the routing target"; and the divergence is test-enforced at window-state.service.test.ts:293-308, where focus moves to windows 2 and 3 while routing stays on window 1. So the claim is true and the trigger is real, not theoretical, and the text ships to extension authors at 9 sites in papi.d.ts. Downgraded on rating only: this is documentation with no runtime effect, its actionable guidance ("not necessarily the window you are in... use papi.window") stays correct, and the codebase itself carries the same loose shorthand in the engine's class name and the ADR's own title, reserving precision for the bodies._

Same fix as #1, #3, #32.

_Other findings in this file: #1, #3, #32, #33, #34, #35_

src/shared/services/window.service-model.ts
```diff
@@ -13,21 +13,24 @@ export const windowServiceObjectToProxy = Object.freeze({
...
+   * whichever window currently has focus. So finding the data provider by this name — with the
```
lyonsil: **#32** - medium -> low · downgraded on review

The new warning is framed purely as a read hazard, but the same provider exposes `setFocus` through the generic name, and the write misfires silently.

`FocusedWindowDataProviderEngine.setFocus` forwards to the routing target, and `useData` returns the setter as its second tuple element, so the write is reachable through exactly the hook the doc names. The web-view branch of the shard's setFocus performs no existence check on the id, so a foreign web view id mutates the other window's focus state and returns `true` while the underlying tab update returns false - a misfire reported as success.

_Why this is downgraded: The mechanics check out. FocusedWindowDataProviderEngine.setFocus at window.service-router.ts:184-196 resolves #getTargetWindowService() (the routing target, :291-308) and forwards, and the engine is registered under the unscoped windowServiceProviderName at :321-330; the x-experimental marker there is OpenRPC doc metadata only and gates nothing. The write is reachable through exactly the hook the doc names: WindowDataTypes.Focus declares `setData: SetFocusSpecifier` (window.service-model.ts:76-79) and useData returns dataProvider['setFocus'] as its second tuple element (create-use-data-hook.util.ts:213-233). The consequence is real and unguarded: the focusType 'tab' branch validates via getTabInfoById and returns false when absent, but the web-view branch falls through with no existence check (window.service-shard.ts:449-466), so a foreign id mutates the other window's #focusSubject and returns true while focusTab's updateTab returns false - a state misfire reported as success, rather than the "moves focus inside window 2" the finding describes. Downgraded because the same added sentence already ends with the remedy that covers writes as well as reads, so no reader is actively misled._

Same fix as #1, #2, #3.

_Other findings in this file: #1, #2, #3, #33, #34, #35_

src/shared/services/window.service-model.ts
```diff
@@ -13,21 +13,24 @@ export const windowServiceObjectToProxy = Object.freeze({
...
+   * useData hook, for instance — answers for the focused window, which is not necessarily the
+   * window you are in. Use `papi.window` to reach the caller's own window.
```
lyonsil: **#3** - high -> medium · downgraded on review

The remedy "Use `papi.window` to reach the caller's own window" is stated unconditionally but is renderer-only; from the extension host it does the very thing the sentence above warns about.

`getWindowService()` takes `if (!globalThis.windowId) return getFocusedWindowService();`, and windowId is set only in the renderer. In the extension host `papi.window` therefore resolves to the focused/routed window, and the proxy's own getter returns the unscoped, focus-forwarding name there. The doc block is reachable from `@papi/backend`, so a backend author reads the remedy without the caveat. Qualifying the clause - e.g. "from a renderer, `papi.window` reaches the window you are in" - is the whole fix.

_Why this is downgraded: The mechanics check out: line 20 is an added line, globalThis.windowId is set only at renderer/global-this.model.ts:60, so window.service.ts:86 makes papi.window the focused window in the extension host, and the proxy getter even returns the unscoped focus-forwarding name there. Reachability is confirmed: IWindowService intersects typeof windowServiceObjectToProxy (papi.d.ts:11048), so the block ships on papi.window.dataProviderName inside `declare module '@papi/backend'` (papi.d.ts:11864). It is smaller than high for two reasons. The "contradiction 72 lines later" is in practice the fix rather than the flaw: the remedy names papi.window, and papi.window's own TSDoc in the same generated backend surface (papi.d.ts:287-289, 547-549) reads "The extension host is in no window, so a call made there acts on whichever window has focus at that moment" - a reader who follows the pointer hovers straight onto the caveat. And the paragraph is renderer-framed by its own example ("with the useData hook, for instance"), and useData is renderer-only. A real, cheap-to-fix imprecision worth raising on a doc-accuracy PR, not a high._

Same fix as #1, #2, #32.

_Other findings in this file: #1, #2, #32, #33, #34, #35_

lyonsil (commented):
Automated deep review of the documentation half of PT-4278.

Every finding below was independently pressure-tested against this branch before it was included, and each one carries the checker's verdict and reasoning with it — including where that check cut a finding's severity or narrowed what survived of it. Where that check contradicted the original report's specifics but something smaller remained true, the item says what survives rather than restating the original claim.

Findings that sit on a line this PR changed are inline; the rest are collected at the end of this body because GitHub can only anchor a comment to a changed line.

Anchors resolved against `85f0966cf07b9c57c35102686853f009334cb3ae`.

**#4** (medium · confirmed) src/shared/services/window.service-model.ts:143 - `subscribeFocus`'s `@param callback` is copy-pasted from the menu service and describes menu content, not focus.

The text reads "Function to run with the updated localized menuContent for this selector" - verbatim identical to the menu data service's own TSDoc. It is published to extension authors in the generated papi.d.ts at two overload sites. Pre-existing and about 19 lines below this PR's last edit to the file, but papi.d.ts is a generated artefact, so fixing it later costs another full `npm run build:types` regeneration and a 3-OS CI matrix run - cheap now, not cheap later.

_Verification: window.service-model.ts:143 reads `@param callback Function to run with the updated localized menuContent for this selector.` - verbatim identical to menu-data.service-model.ts:73, confirming copy-paste from the menu service into a focus-subscription callback that has nothing to do with menu content. It is published into papi.d.ts:11035 and again at :11257 for another overload, so extension authors reading the generated .d.ts see the wrong description. `git diff dc97223 HEAD -- src/shared/services/window.service-model.ts` shows the PR's nearest hunk ends at the setFocus @param wording fix around line 124 and contains no change to subscribeFocus, confirming this is pre-existing and not introduced here. The claim is accurate as stated; medium because it is a real, user-visible doc defect in a generated public artefact, though narrowly scoped to one @param line._

**#5** (medium -> low · downgraded on review) .claude/rules/architecture/shared-patterns.md:56 - The agent-facing hosting rule cites a data type that has never existed: `AppFocus`.

The rule reads "`window.service-shard.ts` (emits `Focus` / `AppFocus` from renderer-local events)". `WindowDataTypes` declares only `Focus`. An agent asked to add app-level focus tracking will believe it is already modelled - during an epic whose subject is separating window focus from app focus.

_Why this is downgraded: The factual core is fully verified. shared-patterns.md:56 does read "`window.service-shard.ts` (emits `Focus` / `AppFocus` from renderer-local events)", and a case-insensitive grep over the entire worktree (node_modules and untracked files included) plus `git grep -iE 'app[-_ ]?focus'` return that single line and nothing else. WindowDataTypes at window.service-model.ts:78-80 declares only Focus, papi.d.ts:10968-10970 matches, and the shard's only emit is notifyUpdate('Focus') at window.service-shard.ts:352. `git log -S AppFocus` shows it never existed in code: the token traces to the June 2026 profile-embed commit 865a2a19446, where the same sentence also mis-cited window.service-host.ts; a later commit fixed the filename half and left AppFocus. What shrinks it: this PR touches zero files under .claude/, so it is a pre-existing stale token the PR neither introduces nor continues, and the harm is one word in a parenthetical whose actual rule - per-window engine hosted in the renderer - is still correct. Fairly raised here as adjacent cleanup, since shared-patterns.md is the agent-facing twin of the Architecture.md hosting text this PR rewrote._

**#6** (low · downgraded on review) .claude/skills/app-runner/reference.md:17 - Another single-Renderer process diagram, left stale by this PR's sweep, in a skill reference an agent reads while driving a running app.

The file's diagram still shows one `Renderer (React UI)` box, and the skill treats CDP as a single endpoint. An agent following it will not expect a second window's renderer.

_Why this is downgraded: Confirmed by reading: .claude/skills/app-runner/reference.md has a process diagram at lines 8-24 with a single Renderer box (lines 17-19), and `git diff dc97223 HEAD -- .claude/skills/app-runner/reference.md` is empty, so the file is untouched. It shares the old single-Renderer concept with pre-PR Architecture.md but is not a literal copy - it already carried extra detail (port 5858, localhost:1212) not in Architecture.md, so it was independently hand-maintained and had already diverged before this PR. SKILL.md in the same directory does treat CDP as one endpoint, supporting the single-endpoint detail. But this PR's diff touches only Architecture.md, papi.d.ts, window.service-model.ts and two storybook .mdx files - a targeted public-window-API and terminology sweep, not a repo-wide diagram audit. Real staleness, but reasonable follow-up rather than a gap in this PR's stated scope._

**#7** (low · downgraded on review) .claude/skills/papi-client/reference.md:16 - Another single-Renderer process diagram, left stale by this PR's sweep, in the skill reference an agent reads while querying a running app over PAPI.

The file's mental model is single-renderer throughout - it never mentions per-window renderers or window-scoped (`-<windowId>` suffixed) provider names, which is exactly the discovery detail a PAPI client needs once more than one window is open.

_Why this is downgraded: Confirmed on all facts: .claude/skills/papi-client/reference.md lines 8-19 carry the identical box-art diagram as the pre-PR Architecture.md lines 25-36 (verified via `git show dc97223:.context/standards/Architecture.md`); `git diff dc97223 HEAD -- .claude/skills/papi-client/reference.md` is empty, so this PR touched nothing in that file; and a grep for window/shard/multi-renderer in the file turns up no mention of per-window renderers or shard-suffixed provider names anywhere - the skill's mental model really is single-renderer throughout. However the file is entirely outside the PR's diff and outside its stated scope (the multi-process diagram in the architecture standard), making it reasonable in-spirit follow-up rather than a defect introduced or left incomplete within this PR's own edit._

**#8** (low · downgraded on review) .context/research/Paratext-Feature-Porting-Analysis.md:65 - Another single-Renderer process diagram, in an unfenced research snapshot that the repo's own durability rule says should have been marked frozen.

The diagram at lines 54-73 structurally matches the pre-PR Architecture.md one. `.claude/rules/docs-durability.md` says point-in-time research docs should carry a "Frozen record" blockquote and are then exempt from staleness fixes; this file has no such blockquote, so it reads as live.

_Why this is downgraded: The diagram is real, at .context/research/Paratext-Feature-Porting-Analysis.md:54-73 (the Renderer box is line 65), and structurally matches the pre-PR Architecture.md single-Renderer diagram. `git log`/`git diff` from merge-base to head for this path are both empty, confirming the PR never touched it. However this file lives under .context/research/ and is a one-time porting-analysis snapshot (dated content like "Electron 36.5" throughout), not a maintained standard - .claude/rules/docs-durability.md explicitly says point-in-time research/investigation docs should carry a "Frozen record" blockquote and are "exempt from staleness fixes" once fenced. This file has no such blockquote, so it is a pre-existing gap in applying that convention, but the convention itself establishes that .context/research/ content is not expected to track live architecture changes the way Architecture.md is. Smaller and out of scope: an unfenced research snapshot going stale, not a defect left by this PR in a maintained reference._

**#10** (low · downgraded on review) .context/standards/Architecture.md:12 - The file's freshness markers were left stale by its largest content change in a while.

The verification banner still reads "Verified against paranext-core origin/main `998ca09a087` - 2026-08-03", the frontmatter still says `last_updated: 2026-03-04`, and the Version Log is still a single row, while this PR rewrites the diagram and adds a nine-line paragraph dated 2026-08-27/28.

_Why this is downgraded: Markers are as claimed - frontmatter `last_updated: 2026-03-04` at Architecture.md:6, a single-row Version Log near EOF, and the `Verified against ... 2026-08-03` banner at line 12 - and the PR's diff touches only the diagram and the new paragraph, never the frontmatter or stamp. But .claude/rules/docs-durability.md's "Date every status claim" bullet is about dating external status claims in prose ("as of 2026-08-03, PR #2626 is open"); it says nothing about frontmatter last_updated, version logs, or the Verified-against banner, so the cited rule does not actually govern these markers. History confirms it is not an enforced convention either: the two immediately preceding commits to this same file (b91c4cbea9e and 2127884be0f, both 2026-08-27/28, each larger than this PR's diff - one adds a whole new 60+ line section) also left all three untouched. The staleness is real but pre-existing, and no rule requires this PR to fix it._

**#11** (low · confirmed) .context/standards/Architecture.md:26 - The Main Process box is not rectangular: its content rows' right border sits one column right of its corner characters.

Pre-existing, and untouched by this PR - but the PR rewrites the diagram immediately below it, so the same pass is the cheap moment to square the box.

_Verification: Measured 0-indexed character columns at head: line 25 (top border) is 59 chars with the corner at index 58; lines 26-28 (content rows) are 60 chars with the trailing border at index 59; line 29 (bottom border) is 59 chars with the corner at index 58. So the content-row right border sits one column right of the corner characters - the box is genuinely non-rectangular, matching the finding exactly. Diffing the same lines against `git show dc97223:.context/standards/Architecture.md` shows byte-identical content for lines 23-30, confirming this is pre-existing and not introduced by the PR; the changed hunk starts at line 28 as unmodified context, with the actual edits beginning at line 31. Cosmetic, pre-existing, adjacent to but not part of the PR's edit - low is right._

**#17** (medium -> low · downgraded on review) .context/standards/Architecture.md:110 - "Consumers never see any of this: they call the generic name, exactly as they did before there was more than one window" now sits in tension with the PR's new public TSDoc, which tells consumers they DO see it.

The blanket claim would read better with a clause pointing at the focused/owning-window resolution the same file describes forty lines later.

_Why this is downgraded: The two texts do sit in tension - line 110 is a blanket claim, and the new TSDoc does tell consumers the unscoped name answers for another window - but the finding's two load-bearing specifics are wrong. First, the window service is not "the one router whose generic name is not transparent": web-view.service-router.ts:3 and notification.service-router.ts:4,48 say the same of their own generic names, and every router shares createTargetShardResolver (target-shard-resolver.util.ts:69-84), whose header states routers "forward to 'whichever window the user is working in'... which is every plain forward a router does." Second, the carve-out the finding says is missing already exists: adr-generic-name-routing-proxies (Architecture-Decisions.md:290-330) is titled "...forward to the focused/owning window's scoped service" and names window.service-router.ts alongside the other five, and Architecture.md qualifies line 110 forty lines later at line 150. Third, the PR introduces nothing here - line 110 is verbatim from base, and the routing behavior is pre-existing. A low-value doc-consistency nit in an internal standards file._

**#18** (medium -> low · downgraded on review) CLAUDE.md:65 - The repo-root CLAUDE.md carries a byte-identical copy of the pre-PR diagram, still "four separate processes" over one `Renderer (React UI)` box.

CLAUDE.md is auto-loaded into every session's context and read before Architecture.md, so for an agent the stale copy is the one that wins. But it is not uniquely missed - three other stale copies remain at head, so fixing only this one still leaves the rest wrong.

_Why this is downgraded: The mechanics check out exactly: CLAUDE.md:56 reads "The application runs as four separate processes..." and CLAUDE.md:64-69 is byte-identical (verified by diffing the extracted blocks) to the pre-PR Architecture.md:30-35 that this PR rewrote, so the single Renderer box is genuinely stale, and the auto-load aggravator is real. But it is not the PR's uniquely-missed sibling: a repo-wide grep for the identical row finds three other stale copies at head, so this is a pre-existing four-copy duplication problem and fixing only CLAUDE.md would still leave three wrong. The cited ADR does exist (Architecture-Decisions.md:2050, Accepted) and does say "Renderer platform code now registers no command or request name, scoped or otherwise," but its stated Context is removing the RENDERER_HOSTED_* name lists, not guarding against a single-window mental model, so "exists specifically to prevent that failure" overstates it. Fairly raised - the author demonstrably swept past Architecture.md into two .mdx files - but best handled as follow-up._

**#19** (low · downgraded on review) README.md:277 - README's startup-timing section still says "all four processes (main, renderer, extension host, and the .NET data provider)".

Arguably a shorthand for four process kinds rather than a claim that only one renderer exists - and cold start is exactly the moment when only one renderer does exist.

_Why this is downgraded: README.md:277 reads exactly as claimed, and `git diff dc97223 HEAD -- README.md` is empty, confirming the file is untouched by this PR. Grepping README.md for "four processes"/"the renderer"/"the window"/"app window" finds only this line and its neighbour at 279 ("main process forwards it to the renderer") - no other single-window prose to flag. The phrase sits in the "Startup performance timing" section describing cold-start profiling, where exactly one renderer exists (additional windows only spawn after startup completes), so "four processes" is a defensible shorthand for the four process KINDS emitting marks, not an assertion that only one renderer ever exists - it is not affirmatively wrong. Unchanged by this PR and reasonable in its own context: a plausible future doc-consistency nit, not a defect left stale by this PR's sweep._

**#20** (medium -> low · downgraded on review) lib/platform-bible-react/src/stories/guidelines/application-interactions.mdx:17 - The Menus table still says the main menu "hosts all menu items that apply to the application as a whole", in a file this PR edited and whose final commit claims to finish the single-window sweep.

Every window renders its own menubar. The scope claim is now at least imprecise: a reader may wonder whether a main-menu action affects all windows or one.

_Why this is downgraded: The anchored text is real and unedited - line 17 still reads "This menu hosts all menu items that apply to the application as a whole", and the PR's only edit to this file is line 43. The per-window mechanics also check out: every window is a new BrowserWindow loading the same renderer (main.ts:611), whose root renders PlatformBibleToolbar (app.component.tsx:35) and its Menubar (platform-bible-toolbar.tsx:426), so each window does draw its own menubar. But the load-bearing inference fails: enumerating every mainMenu item in menu.data.json (Settings, Exit, Open new window, four Help links, feedback x2, About), none is window-scoped in the sense of acting on the current window - `%mainMenu_openNewWindow%` ADDS a window to the application, which is an application-level action, and quit/about are unambiguously app-wide. Nor does it contradict the new terms.mdx:7 sentence, which is about where an action's surface appears, a different axis. Mild imprecision in UX-owned copy, not a falsified statement the sweep was obliged to fix._

**#23** (medium -> low · downgraded on review) lib/platform-bible-react/src/stories/guidelines/application-interactions.mdx:49 - "global BCV control" names a per-window control in a passage that also refers to the genuinely app-global scroll group.

The title-bar BCV control is per-window; the scroll group really is app-global, hosted in main. A multi-window sweep could usefully tighten "global" here. Separately and outside this finding, line 54's "only the scroll group selection ... switches" looks stale against current toolbar code for reasons unrelated to windows.

_Why this is downgraded: Both mechanical halves check out. Line 49 reads "a global BCV control on the application title bar" and line 54 ends "only the scroll group selection of the global BCV control switches"; the title-bar control is genuinely per-window (app.component.tsx:35 renders PlatformBibleToolbar at the renderer root, whose scroll group derives from a module-level per-renderer variable at window.service-shard.ts:92, registered "under a window-scoped name ... so multiple renderers can coexist" at :591), while the scroll group really is app-global per adr-scroll-group-hosted-in-main. What does not survive is the severity framing. The PR's diff to this file touches only line 43, so 49 and 54 are pre-existing prose, and their "global" is locally disambiguated by the bullet list's own contrast axis - "a global BCV control on the application title bar" versus "an in-tab BCV control on the tab toolbar of every tab" - i.e. title-bar-wide as opposed to per-tab, the established in-app sense, not a cross-window claim. Line 54 contrasts tab-selector with title-bar-selector behavior and says nothing about other windows, so the page is silent on cross-window sharing rather than wrong about it._

**#24** (medium -> low · downgraded on review) lib/platform-bible-react/src/stories/guidelines/dismissal-patterns.mdx:34 - The dismissal rule's "not the window merely losing focus (alt-tab, DevTools)" enumeration does not name the new sibling-window case.

`document.hasFocus()` is per-OS-window, so with a second window open, window A reads false when the user clicks into window B, and the prescribed guard suppresses dismissal identically. "Stay open" is plausibly right, but it should be a decision rather than an inherited accident of a rule written for alt-tab.

_Why this is downgraded: The wording is exactly as quoted - lines 34-36 ("*not* the window merely losing focus (alt-tab, DevTools). A plain window switch does not dismiss anything"), the guard at line 148, and the glossary restatement at line 202 - and the mechanism holds: document.hasFocus() is per-OS-window, so window A reads false when the user clicks into window B and the guard suppresses dismissal identically. What shrinks it is that nothing implements the guard: a repo-wide grep for hasFocus finds it in code only in use-focus-replaced-content.hook.ts (focus RESTORATION, not dismissal) and its tests, and a grep for window blur dismissal listeners across lib/platform-bible-react/src, src and extensions/src returns nothing - the page says so itself at lines 179-184 ("No outside-interaction/dismissal hook exists ... yet"). So there is no shipped stale-popover behavior, only a future implementer's example list. The rule is also not falsified: switching to sibling window B IS a plain window switch, so the stated rule still covers it correctly; only its parenthetical enumeration fails to name the new case._

**#25** (low · confirmed) lib/platform-bible-react/src/stories/guidelines/terms.mdx:1 - This PR edits two UX-team-owned files but routes only one of them for UX review.

`.claude/rules/where-to-add-guidance.md` scopes `lib/platform-bible-react/src/stories/guidelines/**` to the UX team with "UX team internally" review. The PR body explicitly routes terms.mdx for UX review but never mentions application-interactions.mdx, even though the final commit makes a substantive wording change there for the same reason.

_Verification: .claude/rules/where-to-add-guidance.md frontmatter (line 5) scopes the rule to lib/platform-bible-react/src/stories/guidelines/**, and its table (line 17) states that directory is "UX team" owned requiring "UX team internally" review. git diff confirms the PR edits two files there: terms.mdx (the Window and Floating tab group entries) and application-interactions.mdx (line 43, the "docked back into the main window" sentence). The PR body's "UX-owned glossary lines folded in from #2646" section explicitly routes terms.mdx for UX review ("so the UX team reviews them here rather than on a second PR"), but never mentions application-interactions.mdx anywhere in the body, even though the final commit 85f0966 ("finish the single-window sweep in the interactions guideline") makes a substantive wording change there for the same single-window reason. The process concern is self-handled for one of the two UX-owned files but not the other - a real, if narrow, gap._

**#29** (low · confirmed) lib/platform-bible-react/src/stories/guidelines/terms.mdx:8 - The `App title bar` bullet, immediately below the line this PR rewrote, still says "at the top of the application" and lists "the OS window controls".

A title bar and its OS window controls belong to a Window, not to the application. The adjacent `Floating tab group` bullet got exactly this treatment in this PR - swapping "app window" for the defined `Window` term - so the fix pattern was in hand one line further down and not applied here.

_Verification: terms.mdx:8 still reads "The horizontal bar at the top of the application, holding ... and the OS window controls" - unchanged by this PR per `git diff dc97223 HEAD`, which touched only the `Window` bullet and the `Floating tab group` bullet (changed from "it's still in the same app window" to "it's still inside the same `Window`"). A title bar and its OS window controls are properties of a Window instance, not "the application" as a whole, so the same single-window assumption this PR set out to fix survives one bullet below the line it rewrote - and the adjacent Floating tab group bullet shows the exact fix pattern was already in hand and applied one line further down but not here. This is on an unchanged line, so it is a completeness gap in the PR's stated scope rather than a defect introduced by the diff._

**#30** (low · downgraded on review) src/renderer/services/window.service-shard.ts:63 - The exact sentence this PR retired from the public types survives in the renderer file that implements them.

A repo-wide grep for "Focus of the window is somewhere not in a tab" returns exactly this one hit. But it sits on `FocusSubjectElement`, a file-private helper type - a different type from the public `FocusSubjectOther` the PR reworded - and no window-shard module is emitted into papi.d.ts, so it never reaches extension authors.

_Why this is downgraded: The string claim is true and reproducible: grepping the repo (excluding node_modules/.git) for "Focus of the window is somewhere not in a tab" returns exactly one hit, window.service-shard.ts:63, and `git diff --name-only dc97223..85f0966` lists only the five known files, so this file is untouched. But it is not a stale copy of the public doc: the comment sits on FocusSubjectElement (declared unexported at :68), whose own comment says "It is a helper internally for this file only" and whose field carries "Do NOT expose FocusSubjectElement outside this class" (:327) - a different, file-private type from the public FocusSubjectOther the PR reworded. It never reaches extension authors: lib/papi-dts/tsconfig.json includes only src/declarations/*, the papi-backend/frontend services and a few models - no window-shard module is emitted - and papi.d.ts:10947 already carries the new wording. The other two cited sites are at 497-498 and 541 (not 498/541) and are arguably correct: both run inside one renderer against its own document.activeElement._

**#31** (medium -> low · downgraded on review) src/shared/models/docking-framework.model.ts:204 - The deferral of the "floating window" rename is real but unrecorded, and PT-4281 will not actually perform it.

The public API still calls a floating tab group a "floating window" at lines 204, 215 and 219, shipping at papi.d.ts:3775-3794 - the exact confusion terms.mdx:14 fixes. The PR body defers this to PT-4281, and deferring is defensible (PT-4281 restructures the same `Layout` union). But there is no written rule anywhere recording the lock, and PT-4281's branch does not contain the rename - so the wording is frozen with nothing scheduled to fix it.

_Why this is downgraded: The mechanics check out exactly: lines 204, 215 and 219 are verbatim as claimed, they ship at papi.d.ts:3775, :3788 and :3792, terms.mdx:14 does carry the "it's not literally a separate window" gloss, and `git diff dc97223..HEAD` on this file is empty, so it really is untouched. But the deferral is substantively real, not an excuse: PT-4281 exists as the active branch origin/pt-4281-window-layout-disposition (sibling open PRs #2724/#2729/#2730) and modifies this exact file (+39/-2), with commit f02b6e2d3c1 adding `WindowLayout { type: 'window' }` to the same Layout union FloatLayout belongs to, about 45 lines below the cited comments - so asking for the edit here would land a doc-only change on top of a structural rewrite. What does not survive is the framing and the guarantee: there is no written rule anywhere (zero hits for "standing coordination", "coordination rule", "changes only via", "until it lands", or "docking-framework.model" across .context/, .claude/, CLAUDE.md, and no ADR entry), and PT-4281's branch does NOT contain the rename - its lines 204/215/219 are byte-identical to main. So the useful residue is a tracking point: record the deferral where it can be found, and queue the rename, since landing PT-4281 will not do it._

**#36** (high -> low · downgraded on review) src/shared/services/window.service-model.ts:104 - `getFocus` is declared `Promise<FocusSubject>` while both engines return `FocusSubject | undefined`, so destructuring the result typechecks and throws before focus first resolves.

The explicit overloads win over the `IDataProvider<WindowDataTypes>` ones in the intersection, dropping the `undefined` that `WindowDataTypes.Focus` declares. Entirely pre-existing and outside this PR's hunks; both in-repo callers already guard, so the residual risk is third-party extension authors reading the published surface.

_Why this is downgraded: The type claim is true and I settled it by construction: a probe compiled with the worktree's tsc shows `await svc.getFocus()` resolves to FocusSubject (error: Type 'FocusSubject' is not assignable to type 'null'), and destructuring compiles with no error, while Awaited<ReturnType<IWindowService['getFocus']>> is FocusSubject | undefined - so the explicit overloads at lines 104-106 do win over the IDataProvider getter, and the undefined from WindowDataTypes.Focus (line 79) never reaches a caller's type. Both engines really do return it (window.service-shard.ts:385 `if (!this.#focusSubject) return undefined;`, reachable via detectFocus() finding nothing and via the explicit deselect setFocus(undefined), and window.service-router.ts:199). But it is entirely pre-existing and this PR gives it no hook: `git show dc97223:...` has byte-identical declarations, and the PR's hunks are @@ -85,7 +88,9 @@ and @@ -103,7 +108,7 @@, so head lines 104-106 sit BETWEEN hunks - the @returns text the finding says the PR rewrites is unchanged in the source and in papi.d.ts. Both in-repo production callers already guard, so nothing is broken today._

**#37** (medium -> low · downgraded on review) src/shared/services/window.service-model.ts:108 - The rewrite landed on the type-level docs and left the member docs' "the current window" untouched at seven sites, producing an awkward pronoun chain.

At lines 110 and 124 the new "instruct that window to update its current focus ... focused in it" is anaphoric to the untouched "the current window's focus" earlier on the same unchanged line. That pronoun chain is the genuine artifact and the cheapest fix; "the current window" itself is not the defect "the window" was, since "current" presupposes plurality.

_Why this is downgraded: The mechanics verify exactly: grepping the file at head returns 7 hits at precisely the cited lines 99, 102, 108, 110, 121, 124, 140 (9 in papi.d.ts, which duplicates the getFocus block), and `git diff dc97223 HEAD` contains no +/- line with the phrase at all - every rewrite landed on the type-level docs while the IWindowService member docs got only the two @param continuations, so the type-vs-member split the finding describes is real. The antecedent point is also real: at 110 and 124 the new "instruct that window ... focused in it" is anaphoric to the untouched "the current window's focus" on the same unchanged line, the one genuinely awkward artifact and the cheapest fix. But the severity rests on two overstatements. "The current window" is not the defect "the window" was - "current" presupposes plurality and picks out the window a call resolves to, and the repo uses it unremarkably elsewhere. And the IntelliSense aggravator is refuted by the JSDoc plumbing: JSDOC SOURCE windowService is stamped onto the backend entry point at papi-backend.service.ts:135 and :234, so a backend author hovering papi.window does see "The extension host is in no window"._

_Not anchored: no longer on a changed line; cannot be posted inline (ambiguous: excerpt also matches at line(s) [121]; nearest to 108 chosen)_

**#38** (medium -> low · downgraded on review) src/shared/services/window.service-model.ts:152 - The `subscribeFocus` callback is declared `FocusSubject | PlatformError`, dropping the `undefined` the data type declares and the subscribe path really emits.

The immediate-retrieval path calls the callback unconditionally, including when the data is undefined - which is why the platform's own overlay host dereferences after only an isPlatformError guard. Pre-existing, and contained: the resulting throw is caught by the subscriber's try/catch and degrades to a logged warning plus a PlatformError re-dispatch rather than a crash.

_Why this is downgraded: The type claim is exactly right and I proved it by construction. window.service-model.ts:150-154 declares `callback: (focusSubject: FocusSubject | PlatformError) => void`, while WindowDataTypes.Focus is `DataProviderDataType<undefined, FocusSubject | undefined, SetFocusSpecifier>` (line 79) and DataProviderSubscriber types the callback as PlatformEventHandler<TDataType['getData'] | PlatformError> (data-provider.model.ts:128-132). A probe compiled against the repo tsconfig showed the raw IDataProvider<WindowDataTypes> call site contextually types the param `PlatformError | FocusSubject | undefined`, but the IWindowService call site types it `PlatformError | FocusSubject`, and an `fs === undefined` guard narrows to never - so the explicit declaration wins the intersection. undefined is genuinely emitted: window.service-shard.ts:385 returns undefined when #focusSubject is unset, and data-provider.service.ts:305-322 calls callback(data) unconditionally on the retrieveDataImmediately path (its own comment notes it fires "including if the data is undefined"). Two things shrink it: byte-identical at the merge base, and the resulting throw is contained by the try/catch at data-provider.service.ts:311-313._

**#39** (low · downgraded on review) src/shared/services/window.service-model.ts:155 - `onDidDispose` is typed synchronous but proxied async, so a synchronous `unsub()` on the returned value throws.

Pre-existing, repo-wide across every `*ObjectToProxy`-backed data-provider service, and unreachable: nothing in the repo calls `papi.window.onDidDispose`, and the internal uses call it on the raw resolved provider rather than the proxy. Even through the proxy the awaited value is a working unsubscriber, so nothing is silently lost.

_Why this is downgraded: The mechanic is exactly as claimed and I proved it by construction: a probe replicating createSyncProxyForAsyncObject verbatim (util.ts:350-362 - `if (prop in target) return target[prop];` else an async wrapper) against windowServiceProxyTarget (window.service.ts:100-111, own keys ['dataProviderName'] only) prints `'onDidDispose' in target: false`, `is Promise: true`, and `unsub() call THREW: TypeError: ret is not a function`, while OnDidDispose is typed PlatformEvent<void> = (callback) => Unsubscriber, i.e. synchronous. But it downgrades on three counts. It is untouched and pre-existing: git blame puts `} & OnDidDispose &` at line 155 at e13808a6824e (2025-06-20), and the PR's hunks stop at line 125. It is a repo-wide pattern: IMenuDataService ends with the identical intersection over a proxy target that likewise holds only dataProviderName, as do theme-data and settings. And it is unreachable in practice: grep finds zero papi.window.onDidDispose call sites anywhere, the real internal uses call onDidDispose on the raw resolved provider, and even through the proxy the awaited value IS a working unsubscriber._

src/shared/services/window.service-model.ts
```diff
@@ -85,7 +88,9 @@ declare module 'papi-shared-types' {
 /**
  * JSDOC SOURCE windowService
  *
- * Service that allows to interact with the current application window
+ * Service for interacting with an application window. Every window hosts its own, so a call from a
+ * renderer acts on the window it runs in. The extension host is in no window, so a call made there
```
lyonsil: **#33** - medium -> low · downgraded on review

The new extension-host sentence documents only the success path; `papi.window` throws there in two states with no doc warning.

`getFocusedWindowService()` throws "Window service is not available: no focused window found" and, separately, when the focused window has not registered its shard. The repo's own TSDoc requirement 5 asks for failure modes. But accessor-availability throws are uniformly undocumented across every papi service, so this is a nice-to-have on a paragraph the PR was already rewriting, not a standards violation this PR introduced.

_Why this is downgraded: The mechanics all check out: both throws exist at window.service.ts:41 and :49, the extension host reaches them because getWindowService routes there whenever globalThis.windowId is falsy (:85), the paragraph is API surface (JSDOC DESTINATION windowService lands on papi.window at papi-backend.service.ts:135-136), and requirement 5 exists verbatim at Code-Style-Guide.md:164. The startup race is real too - extensionHostService.start runs at main.ts:388 while restoreWindows() runs at :1497, and platform.getFocusedWindowId deliberately returns honest OS focus rather than the routable target, so a window with focus but no registered shard hits the second throw by design. But "two ordinary states" overstates it: focusedWindowId is sticky - set only on focus (main.ts:661), cleared only when that window is removed - so alt-tabbing away does not trigger it; it needs pre-first-focus startup or macOS zero-window residency. Decisively, the accessor-availability throw is a uniform, uniformly-undocumented repo pattern: appService, scrollGroupService, localization, menu-data, web-view, database and data-protection all throw "not available as a network object" and none document it; the 39 @throws in papi.d.ts are all method preconditions._

Same fix as #34, #35.

_Other findings in this file: #1, #2, #3, #32, #34, #35_

src/shared/services/window.service-model.ts
```diff
@@ -85,7 +88,9 @@ declare module 'papi-shared-types' {
...
+ * acts on whichever window has focus at that moment, which can differ between two calls.
```
lyonsil: **#34** - medium -> low · downgraded on review

"acts on whichever window has focus at that moment" is imprecise while the app is backgrounded - nothing clears the focused window id on blur.

There is no blur handler anywhere; setFocusedWindowId is wired only to `newWindow.on('focus')`, and the id is cleared only in removeWindow. "whichever window most recently had focus" would be strictly accurate.

_Why this is downgraded: The mechanics check out exactly: the only production writer of focusedWindowId is setFocusedWindowId(windowId) at main.ts:661, wired solely to newWindow.on('focus'); it is cleared only in removeWindow (window-state.service.ts:456) and resetForTesting; a grep for blur across src/main and src/shared finds no handler at all, so getFocusedWindowId() (:284) returns the last-focused id while the app is backgrounded - and that is precisely what an extension-host papi.window call resolves through (window.service.ts:39-46). What does not survive is the impact: "last focused" is the ordinary reading of "the focused window" for a desktop app, the sentence's actual claim (the target is not fixed and can vary between calls) stays true, and the alternative would be worse - clearing on blur would make getFocusedWindowService() throw for every extension-host call while the user is in another app. The "silently steers a window the user is not looking at" framing is weak, since with the whole app backgrounded the last-focused window is the one the user returns to, and the repo already documents this latching as known and tracked (web-view.service-router.ts:310-315, TODO(PT-4281))._

Same fix as #33.

_Other findings in this file: #1, #2, #3, #32, #33, #35_

---

**#35** - medium -> low · downgraded on review

"which can differ between two calls" omits the subscription consequence: an extension-host `subscribeFocus` pins permanently to the window focused at subscribe time.

The sync proxy resolves the target once per invocation, and the subscriber then binds to that one provider's onDidUpdate with no re-resolution. Subscribe during activation, then work in window 2 all session: the callback keeps reporting window 1 and goes silent when it closes. The router has #repointRelay precisely because a relay must follow the target - but an extension-host papi.window call never takes that path.

_Why this is downgraded: The mechanism is real and I proved it at runtime: createSyncProxyForAsyncObject (util.ts:355-360) calls getObject() once per invocation, and in the extension host getWindowService falls to getFocusedWindowService (window.service.ts:34-53), which resolves the window-scoped shard; createDataProviderSubscriber then binds to that one object's onDidUpdate (data-provider.service.ts:302) with no re-resolution. A probe driving the real built createSyncProxyForAsyncObject confirmed it: after subscribing while shard 1 was focused, shard 2 had 0 subscribers and only shard 1's emission reached the callback. The router contrast is accurate too - #repointRelay (window.service-router.ts:239) re-points the relay for subscribers of the GENERIC name, a path an extension-host papi.window call never takes. What does not survive is "wrong": the added sentence is literally true of a subscribe call too, so this is an omission of the subscription-lifetime consequence, not a false statement; and the only in-repo consumer is overlay.service-host.ts:652, where the own-window provider is cached and correct, so no current caller is misled._

Same fix as #33.

_Other findings in this file: #1, #2, #3, #32, #33, #34_

rolfheij-sil:
🤖 Claude: #1 is right, and it is the important one — the text this PR added was backwards on the path most callers take. Fixed, along with #2, #3, #32 and #4. Most of the rest is answered by something that happened while this PR sat: #2670 merged.

**#1, #2, #3, #32 — one edit, splitting the two audiences.**

I verified #1 independently before changing anything, in four reads. The proxy target defines a `dataProviderName` getter returning the window-scoped name — its own comment says "so dataProviderName returns the scoped name at access time" — the target spreads the unscoped constant *first* and defines the getter *second*, and `createSyncProxyForAsyncObject`'s trap is `if (prop in target) return target[prop]`, where `in` matches accessor properties. So the getter answers and the async path is never reached: `papi.window.dataProviderName` in a renderer or web view is the caller's own scoped name. Your point that the remedy sentence steers readers off the one call that does what they want is exactly right — it is read *on* `papi.window`, and `platform-tab-title.component.tsx` already does that call correctly.

You were also right that it is not a revert. The block now answers per audience: scoped name from a renderer or web view, bare unscoped name from the extension host and from the exported constant.

On **#2**, "whichever window currently has focus" is gone from both sites, including the second one you flagged. They now say the routing target, and note that a window taking OS focus does not become the target until it is ready and not closing, with `platform.getFocusedWindowId` named as what tracks focus itself.

On **#32**, the new text says the provider "both reports and changes the focus of the window you are in", so the write path is not left implied.

There was no test pinning any of this — the one place that mocks `windowService` supplies the *unscoped* name, which is the misunderstanding the old doc created. Added `src/shared/services/window.service.test.ts`: the scoped name where a window id is set, the bare name where none is. Removing the getter so the spread's constant answers instead fails the first and leaves the second passing, which is the split the doc now describes.

**#4 — fixed here, for your reason.** `subscribeFocus`'s `@param callback` did describe "the updated localized menuContent for this selector", and it ships at two overload sites in `papi.d.ts` (confirmed on the generated file, not inferred). Since this branch regenerates those declarations anyway, fixing it now costs nothing and deferring it costs another regeneration plus a three-OS matrix.

**#12–#16 — the `Architecture.md` cluster, fixed as one.** Your #14 was the substantive half: "services that must coordinate across every open window therefore live in main" cannot discriminate when the next clause puts settings, menus and theme definitions in the extension host. The criterion is lifetime, and the paragraph now says so. #16 is fixed by stating the answer it was withholding — theme definitions are extension-host data, the theme service runs in main. #15 was fair: the tables do not carry what the sentence pointed at, so the text now says where a window-scoped service lives rather than sending the reader to look it up. #13's redundant clause is gone. On #11 and #12, the Main Process box really was a column wider on its content rows than at its corners, and the child row overhung by five; the box is squared and widened to match, and the connector columns are untouched since they already aligned.

**#26, #27, #20, #23, and the UX-ownership point — all superseded, not declined.** #2670 merged while this review was open, and it rewrote the same storybook lines on `main`. Both of this PR's storybook commits replayed to nothing during the rebase and dropped out. So:

- **#26** — the "actions you start … happen in the window you are working in" sentence does not exist in `main`'s version, which says instead that windows are equal siblings, none of them a main or parent window.
- **#27** — `main`'s version does not use "Power mode".
- **#20 and #23** are no longer in this PR; the text on `main` is #2670's.
- **The ownership finding is answered by the same fact**: this PR no longer touches `lib/platform-bible-react/src/stories/guidelines/**` at all. You were right to raise it — a UX-owned file was being changed without being routed for UX review, and the PR body did not mention it.

One consequence worth flagging rather than burying: the intent behind those glossary lines — that an action lands in the window you are working in — is now absent from the documentation entirely. Whether a corrected version returns (one that excludes the `existingId` raise case you identified) is a UX call, and it is with Rolf rather than being decided here.

**Declined as pre-existing and out of scope, in your own assessment.** The stale single-Renderer diagrams in `CLAUDE.md`, `.claude/skills/app-runner/reference.md`, `.claude/skills/papi-client/reference.md` and `.context/research/Paratext-Feature-Porting-Analysis.md` (#6, #7, #8, #18); the `AppFocus` token in `shared-patterns.md` that `git log -S` shows never existed in code (#5); `Architecture.md`'s frontmatter and version-log staleness, where you checked the cited rule and found it does not govern those markers (#10); and #17, #19, #29–#31, #33–#39. Your verification classed these as follow-up rather than gaps in this PR, and I agree — the diagram copies in particular are a four-file duplication problem that fixing one of does not solve, as you say.

**One thing I could not check the way you did.** `npm run typecheck` fails in my worktree with 50 duplicate-identifier errors, but it fails identically on `main` — a nested-worktree artifact, since extension tsconfigs reach the repo root through a fixed number of `../`. I ran `origin/main` at the same nesting depth and diffed the sorted error lists: empty. So the branch adds no type error, but CI on the pushed tip is the real authority, not my local run.

lyonsil:
Round 2 checked against the branch. Nine threads are resolved; three stay open, all on the same paragraph.

The `dataProviderName` split is right, and I verified it independently rather than taking the reply's word for it: the proxy target spreads the unscoped constant and then defines a same-named getter over it, `createSyncProxyForAsyncObject`'s `if (prop in target)` trap matches the accessor, and `globalThis.windowId` is set only at `global-this.model.ts:60`. Web views read the parent renderer's global through `window.papi = window.parent.papi` (`web-view.service-shard.ts:2682`), so "renderer or web view -> scoped name" holds for them too. `window.service.test.ts` passes and is falsifiable: renaming the getter fails the scoped case and leaves the unscoped one green, exactly as the reply claims.

## The `JSDOC SOURCE windowService` rewrite names the wrong resolver

#2 flagged two sites. Site 1 (the provider-name block) is now correct. Site 2 got the same "routing target" language, and that is where it is wrong: `papi.window` from the extension host does not go through the router at all.

- `getWindowService()` takes `if (!globalThis.windowId) return getFocusedWindowService();` (`window.service.ts:86`).
- `getFocusedWindowService()` sends `platform.getFocusedWindowId` (`:39`), whose handler returns the raw focus scalar and says so in its own comment: "The honest answer about focus, **not** the window calls are routed to" (`main.ts:1772-1778` -> `window-state.service.ts:377`).
- `getTargetWindowId()` is never reached on this path. It is the router's resolver (`window.service-router.ts:225`, `:275`), used only for the unscoped name the router registers at `:304-305`.

So both sentences invert the behaviour:

- "acts on whichever window **the router** is targeting" - it acts on the focused window.
- "a newly opened window can hold focus while **calls still act on the previous one**" - there is no fallback. When the focused window has not registered its shard, `getFocusedWindowService()` throws (`window.service.ts:48-51`): *"That window reported focus but has not registered its window service, so it may still be starting..."*. The doc promises graceful degradation exactly where the code raises.
- "`platform.getFocusedWindowId` is what tracks focus itself" is offered as a contrast to what `papi.window` does, when it is literally the command `papi.window` calls.

This ships at five sites in `papi.d.ts` (`:4696`, `:12753`, `:13019`, `:13630`, `:13807`), including both `@papi/frontend` surfaces, where the extension-host paragraph is not only wrong but off-topic.

Worth stating outright rather than leaving implied: in the extension host, `papi.window.dataProviderName` (router target) and `papi.window.getFocus()` (focused window) can answer for **different windows**. The two new blocks currently read as one path.

## What the three open threads need

- **#2** - site 2 only; site 1 is settled.
- **#33** - the throw. This was a fair decline before, since accessor-availability throws are uniformly undocumented across every papi service. It is no longer an omission: the paragraph now affirmatively promises the opposite behaviour, so the failure mode has to be named or the promise dropped.
- **#34 / #35** - #34's suggested "whichever window most recently had focus" is now the accurate wording, and the codebase agrees in as many words: `getFocusedWindowId` "stays set while the app is in the background" (`window-state.service.ts:593-595`). #35 is untouched - "which can differ between two calls" survives verbatim, and an extension-host `subscribeFocus` still binds to the window focused at subscribe time with no re-resolution.

## Suggested replacement

For the second paragraph, which closes all three:

> The extension host is in no window, so a call made there acts on the window that most recently had focus - `platform.getFocusedWindowId`, which stays set while the application is in the background. Two calls can answer for different windows, and a subscription binds to the window focused when it was made rather than following focus afterwards. If that window has not registered its window service yet, which happens while a window that has just taken focus is still starting, the call throws rather than falling back to another window.

Note that this is a different resolver from the one the `dataProviderName` block describes: the bare unscoped name goes through the router; `papi.window` does not.

Everything else from round 1 is either fixed, superseded by #2670, or fairly declined, and those threads are resolved.

src/shared/services/window.service-model.ts
rolfheij-sil: 🤖 Claude: Confirmed — site 2 was still wrong, and it's fixed now.

You're right that site 1 (the `dataProviderName` block) already resolves correctly, as you confirmed in your round-2 comment; nothing there changed this round. Site 2 — the `IWindowService` docblock (`JSDOC SOURCE windowService`, the type that ships as `papi.window`) — was still describing the router as the extension-host resolver, and that's wrong: `papi.window` from the extension host goes straight to `getFocusedWindowService()`, a direct focus-driven lookup by provider type. It never calls `getTargetWindowId()`/`getRoutingTarget()`, and nothing in `window.service.ts` touches the router at all.

The paragraph now names that resolver, and points to the `dataProviderName` paragraph — the one that genuinely does go through the router, for the bare unscoped name — by referencing `windowServiceProviderName`. That reference started as a `{@link windowServiceProviderName}` cross-link, but our own follow-up review caught that the link doesn't actually resolve everywhere: `windowServiceProviderName` is only in scope in its home module, and this docblock is duplicated into four other generated `declare module` blocks (`@papi/backend` and `@papi/frontend`) where the constant isn't importable. A `typedoc` build confirmed it concretely — four "Failed to resolve link" warnings, one per affected site, zero at the fifth (home) site. Switched to plain backticked prose instead, matching how the rest of the same paragraph already references `platform.getFocusedWindowId` and `papi.window` without linking them.

Landed in `efd470de111e6f1b69129f3c17dd344f5a048b53` (the paragraph rewrite), with the `{@link}` → backtick fix in a follow-up commit, `efc0fe85dbac9cdbba645d766e5b1b4b13d72729`, both on `pt-4278-api-docs-single-window-sweep` — the same paragraph closes #33 and #34/#35 below, since all three point at this one text.

src/shared/services/window.service-model.ts
rolfheij-sil: 🤖 Claude: Confirmed and fixed — and we went one step further than your suggested wording.

Agreed the paragraph needed to name the failure mode once it started affirmatively promising a graceful fallback that doesn't exist. We verified `getFocusedWindowService()` (`window.service.ts`) directly: it throws in **two** places, not one. Your suggested replacement text covers the second — the focused window's provider not yet registered, described as "still starting" — but the first throw (no focused window at all: reachable at startup before any window has ever taken focus, and momentarily after the focused window closes, since `removeWindow` clears `focusedWindowId` with nothing re-deriving it) isn't in that wording. The second throw's own error message also names a second cause your wording doesn't carry: "may still be starting **or may have just gone away**."

So the paragraph we landed goes beyond what you proposed, on the same reasoning your finding gave: a paragraph being rewritten specifically to stop overpromising shouldn't leave a smaller version of the same gap in place. It now reads: "If there is no focused window, or the focused window has not registered its window service — either because it is still starting or because it has just gone away — the call throws rather than falling back to another window."

Landed in `efd470de111e6f1b69129f3c17dd344f5a048b53`. One unrelated follow-up landed on the same paragraph afterward, `efc0fe85dbac9cdbba645d766e5b1b4b13d72729` — it only fixes a dead `{@link}` link in the closing sentence (see the reply on #2), nothing in the wording you raised here.

src/shared/services/window.service-model.ts
rolfheij-sil: 🤖 Claude: Confirmed and fixed — both landed as you suggested.

**#34**: "most recently had focus" is now the wording, matching the mechanism you cited — `focusedWindowId` is written only on focus and cleared only when that window is removed, with no blur handler anywhere, so it stays set while the whole application is backgrounded.

**#35**: the paragraph now says a subscription "binds to the window focused when it was made rather than following focus afterwards," matching the runtime probe in your comment — `createSyncProxyForAsyncObject` resolves `getObject()` once per invocation, so `subscribeFocus` binds to that one provider's `onDidUpdate` with no re-resolution afterward.

Both landed in `efd470de111e6f1b69129f3c17dd344f5a048b53`, alongside #2 and #33 — one paragraph rewrite closes all three open threads. One unrelated follow-up landed on the same paragraph afterward, `efc0fe85dbac9cdbba645d766e5b1b4b13d72729` — it only fixes a dead `{@link}` link in the closing sentence (see the reply on #2), nothing in the wording covered here.

lyonsil (approved):
lgtm

PR #2732, opened 2026-08-27
Add jira-creation skill for creating PT work items

katherinejensen00:
## Summary

Creating an issue in the `PT` Jira project **discards the description passed to the create call** and stores the work-item type's default skeleton instead, so the real description has to be delivered by a follow-up edit. That behavior was documented only inside `.claude/rules/jira-issue-creation.md`, path-scoped to `/prd-to-jira`, so anything creating a ticket outside that command would silently leave a blank templated stub on the board — and nothing in the MCP toolset can delete one.

This adds `.claude/skills/jira-creation/SKILL.md` as the single source of truth for the **create → read the live template → fill it → edit → verify** process, and reduces the rule and the `/prd-to-jira` step to pointers at it.

### Why review this

Not part of current epic. It is agent tooling: it makes Jira tickets created by Claude land correctly filled rather than as blank skeletons that a human then has to find and fix by hand. Worth reviewer time now mainly to sanity-check the two team conventions it encodes (below), since a wrong convention here gets applied to every future ticket.

## Changes

- **New** `.claude/skills/jira-creation/SKILL.md` — the process, a `PT` project reference table, a worked fill-the-template example, and common-mistakes / red-flags tables.
- `.claude/rules/jira-issue-creation.md` — reduced to a pointer (keeps its `prd-to-jira.md` path scope so it still auto-loads and redirects).
- `.claude/commands/prd-to-jira.md` — step 4 now points at the skill; also fixed a mangled line wrap.

## Facts verified against the live `PT` project

These were checked with read-only Atlassian MCP calls, not assumed:

- **Bug's create screen has no `description` field at all.** Its body is `customfield_10116` ("Bug Task Description"). That field also appears on other types carrying its own template, is absent from `getJiraIssue`'s default field set, and is routinely left unfilled — PT-4025 is a Sub-task carrying an empty one.
- **Create-time metadata does not identify templated fields**, in either direction: `description` reports `hasDefaultValue: false` on Sub-task though its template always fires, and `customfield_10116` is missing from Sub-task's create screen yet populated on Sub-task issues. Reading the created issue back with `fields: ["*all"]` is the only reliable discovery.
- **`description` declares `operations: ["set"]` only** — no append or patch at any layer. Array fields (`labels`, `fixVersions`) do declare `add`/`remove`, but `editJiraIssue` exposes only `fields`, so every write through the MCP tool is a whole-field replacement.
- Issue types are Initiative, Epic, Combined, Dev Task, UX Task, Sub-task, Bug, User Snap.

## Please check these two conventions

Both are inferences, not documented team rules — they are the likeliest thing to be wrong:

1. **Should `customfield_10116` be filled on non-Bug types?** The skill says decide per field and state the choice. That is inferred from PT-4025 leaving it blank, which is one data point.
2. **"Set only what the user asked for"** (no invented assignee, priority, or transition). The stricter no-estimates/no-assignee rule is scoped to PRD-derived items and attributed to `/prd-to-jira`; a human-directed "file this bug and assign it to me" is treated as a different case.

## AI Involvement

Written with Claude Code (Opus 5). Claude drafted the skill, ran an adversarial review pass against its own draft, and made the verification calls to the live Jira project; I directed the scope, corrected the claim that only some work-item types have templates, and reviewed the result. Two of Claude's early mechanism claims were wrong on first pass and were corrected by checking the API rather than reasoning — the commit message records what ended up verified.

AI-assisted — [session 1](https://claude.ai/code/session_01TSy4VriQacG7zY2WSTUoba)

## Testing

- [x] `npx prettier --check` clean on all three files
- [x] Pre-commit hooks (lint-staged / format / stylelint) passed
- [x] Claims re-verified against live Jira via read-only MCP calls
- [ ] **Not** pressure-tested against a baseline run — the common-mistakes and red-flags tables target predicted rather than observed failure modes, so they may aim at the wrong things
- n/a `npm test` / `dotnet test` — documentation-only change, no code paths touched

## Risk Level

Low — Markdown under `.claude/` only. No runtime, build, or test code is affected. The realistic failure mode is a convention being wrong and propagating into future tickets, which is what the two questions above are asking you to check.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2732)

### DIFF
--- a/.claude/commands/prd-to-jira.md
+++ b/.claude/commands/prd-to-jira.md
@@ -53,13 +53,14 @@ tools); **drafting** doesn't — if Jira isn't connected, say so and follow step
 3. **Dry-run gate (required — never skip).** Show the user the complete drafts and get explicit
    approval before creating anything. These are posted to the team's Jira under the user's name.
 4. **Create, then populate each description in a second pass.** Creating an issue in `PT` replaces
-   your description with the work-item type's default template, so follow
-   [`.claude/rules/jira-issue-creation.md`](../rules/jira-issue-creation.md) (create → read the
-   live template off the created issue → fit your content into its sections → push the edit →
-   verify). Create the parent, then each child with the parent set — one at a time, keeping the
-   created keys. The child sections this command drafts (User Story, Description, Implementation
-   Ideas, Testing Ideas, Definition of Done, Dependencies) are written to line up with a typical
-   task template — map them onto whatever headings each issue actually shows, carrying any extras
+   your description with the work-item type's default template, so use the **`jira-creation`
+   skill** ([`.claude/skills/jira-creation/SKILL.md`](../skills/jira-creation/SKILL.md)) — create
+   → read the live template off the created issue → fit your content into its sections → push the
+   edit → verify. Create the parent, then each child with the parent set — one at a time, filling
+   and verifying each before moving on, and keeping the created keys. The child sections this
+   command drafts (User Story, Description, Implementation Ideas, Testing Ideas, Definition of
+   Done, Dependencies) are written to line up with a typical task template — map them onto
+   whatever headings each issue actually shows, carrying any extras
    (e.g. **Dependencies**) under the closest section. For the parent, place the problem statement /
    PRD link / non-negotiables table under the closest headings rather than dropping the scaffold.
 5. **Back-fill the mapping**: update the parent's description so each non-negotiable row lists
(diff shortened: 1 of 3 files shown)


jolierabideau (approved):
> @jolierabideau reviewed 3 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2732)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2734, opened 2026-08-28
Snapshot the declared wire surface so PRs can diff it

rolfheij-sil:
## Summary

A wire-only `'x-experimental': true` marker is invisible to every check this repo has. This adds `lib/papi-dts/wire-surface.json` — a checked-in snapshot of the declared wire-visible registration surface, regenerated by `npm run build` — so a change to that surface shows up as a reviewable diff exactly the way a change to `papi.d.ts` does.

**Why review this:** requested by @tjcouch-sil on #2670 — *"Shape it as a snapshot of the full wire surface (the OpenRPC document) that PRs diff the way they diff `papi.d.ts` — not as a marker-policing rule, since plenty of wire methods legitimately live off `papi.d.ts` without being experimental."* Not part of the current epic; it is the tooling half of the multi-window work's API-surface discipline.

**The blind spot is demonstrated, not asserted.** I deleted the `x-experimental` marker from `platform.createWindow` and ran `npm run build:types`: `papi.d.ts` regenerated **byte-identical**, zero diff. The same deletion produces a one-line diff here naming the registration and its file. The C# side behaves the same way — deleting `Experimental = true` from `platformScripture.checklistService` produced a one-line named diff.

## What ships

- **`.erb/scripts/generate-wire-surface.ts`** — scans `src/**` and `extensions/src/**` with the TypeScript compiler API, plus a C# pass. Name and documentation identifiers on the TypeScript side now resolve through the TypeScript checker's binder rather than a hand-rolled lookup, so a shadowed or duplicate identifier binds to the declaration that actually applies.
- **`c-sharp/Paranext.WireSurface`** — a small Roslyn-based .NET console tool that loads `ParanextDataProvider.csproj` through `MSBuildWorkspace` and answers the same questions from the compiled project's real semantic model, replacing the old hand-rolled text scanner entirely. **`c-sharp/Paranext.WireSurface.Tests`** (NUnit) covers it: in-memory fixture compilations against stub framework types, plus one `[Category("Integration")]` test that drives the real `MSBuildWorkspace` path.
- **`build:wire-surface`** spawns the tool via `dotnet run --project c-sharp/Paranext.WireSurface`, validates its JSON output against a runtime type guard, and folds it into the same generated document — still wired into the `build` script, so CI's existing `verify-changed-files` step gates staleness with no workflow change beyond the two `dotnet test`/`dotnet build` steps below.
- **`lib/papi-dts/wire-surface.json`** — regenerated: **187 declared registrations (27 C#) and 28 dynamic (4 C#)**. Sorted deterministically, no line numbers, no version, no timestamp.
- **`e2e-tests/tests/smoke/wire-surface-snapshot.spec.ts`** and its cross-checked **`wire-surface-reduction.util.ts`** — compares the checked-in file against the live `rpc.discover` document; the live backstop for `command:`-prefixed names and the `-pdpf`/`-webViewProvider` suffix families is now corroborated against the snapshot's own declared/dynamic entries before being accepted, rather than accepted by prefix alone.
- **ADR `adr-wire-surface-snapshot`** — Consequences paragraph rewritten to describe the Roslyn-based C# half.
- **CI**: a `dotnet test` step for the new `Paranext.WireSurface.Tests` project, and a warm `dotnet build c-sharp/ParanextDataProvider.csproj` step ahead of the Linux smoke suite (see Verification below for why).

## How the C# half works

The C# half no longer scans source text. `c-sharp/Paranext.WireSurface` loads `ParanextDataProvider.csproj` through Roslyn's `MSBuildWorkspace` and recognises a registration by **the symbol it invokes**, in five rules:

- **A. Network object registration** — an invocation of `NetworkObject.RegisterNetworkObjectAsync`, categorised by the containing type (`pdpFactory` inside `ProjectDataProviderFactory`, `dataProvider` inside `DataProvider`, otherwise `networkObject`); the name is argument 0.
- **B. Data-provider subclass** — a class whose base chain reaches `DataProvider`; the name is the `name` argument of the class's *own* base call (a primary-constructor base argument, or a `: base(...)` initializer whose target constructor's chain reaches `DataProvider`).
- **C. Documentation override** — a method whose `OverriddenMethod` chain reaches `DataProvider.GetNetworkObjectDocumentation`; the entry is attributed to the overriding method's own containing class (`IMethodSymbol.ContainingType`), never to whichever class happens to appear first in the file.
- **D. Standalone request handler** — an invocation of `PapiClient.RegisterRequestHandlerAsync`, skipped when its containing type is `NetworkObject` itself (that fan-out is already represented by the owning A/B entry).
- **E. Network event** — an invocation of `PapiClient.SendRequestAsync` whose first argument resolves to the constant `"network:registerEvent"`; the name is element 0 of the second argument when it's a collection/array initializer, otherwise the whole argument recorded as dynamic.

Every rule matches by symbol identity, never identifier text, so a rename, an added type parameter, or reformatting cannot make a real registration invisible. Name resolution also propagates **one level** through constructor and call arguments: when a name is passed in as a parameter, the resolver finds every call site that binds to that parameter and takes its argument — if every site resolves to a constant, each distinct value becomes its own declared entry; if any site doesn't, the whole thing is dynamic. This is exactly what recovers the Send/Receive notifier's event and command names, which `main` moved out of string literals into constructor parameters after this PR's earlier pushes — the old text scanner degraded both to dynamic entries when that happened; the Roslyn rules recover all four (two instantiation sites × event + command).

**Failure policy:** a compile error, or a workspace failure, exits 2 — diagnostics go to stderr, and if the project simply hasn't been built yet the message says to run `npm run build:data` first. A tracked file that the compiled project doesn't contain exits 3 — the `.csproj` and the tracked-file list have drifted apart and someone has to decide which is right. A file the project compiles that isn't in the tracked list is reported on stderr and skipped, never silently folded in.

**What's excluded, and why:** `NetworkObject`'s own per-function fan-out inside `RegisterNetworkObjectAsync` (already represented by the owning entry, so filing it again under rule D would double-count it); `PapiClient.RegisterRequestHandlerAsync`'s own internal `network:registerMethod` call (how a registered handler actually reaches the wire, not registration surface itself); every `SendRequestAsync` call whose request type isn't `network:registerEvent` (a client-side call into already-registered surface). Each of the three has its own fixture test, replacing the old scanner's always-empty exclusion list.

**What still stays invisible until the live assertion:** a genuinely new registration idiom none of the five rules models. That's the same shape of blind spot the TypeScript half has always had, and it's why the e2e comparison below exists rather than being decoration.

## A substitution TJ should see before review, not discover during it

TJ asked for a snapshot of **the full OpenRPC document**. This ships **the declared registration surface plus a live-document check**, which is not the same thing. The reasons:

- Reproducing the document's per-method fan-out statically is fragile — a network object registers one existence method plus one per exposed function, with the object-level marker cascading unless a per-method entry overrides it. Getting that right statically means resolving each object's members, in both TypeScript and C#.
- **A real part of the surface has no static name at all**: a nonce-minted PDP id, a per-window service shard, a per-provider `onDidUpdate`, a per-project C# data provider. A runtime capture would carry those and churn on every regeneration; a static one cannot know them. They are recorded in a `dynamicRegistrations` section with their unresolved expressions, so the file states its own blind spots rather than omitting them silently.
- `npm run build` does not boot the app, so a runtime capture could not ride the existing staleness gate at all and would need a new CI job.

The fully-resolved view is verified where it actually exists — see the e2e assertion below. If you would rather have the document itself, say so and the runtime capture becomes the primary artifact with a new CI job to carry it.

## The e2e assertion, and why it is not decoration

The scanner alone would rebuild the blind spot one level up: it is a *model* of the wire surface, so a registration shape neither scanner recognises would be as invisible to it as a missing marker is today. The assertion compares both directions, and the second is the one that matters — every live method must reduce to a snapshot entry, a documented dynamic pattern, or a named infrastructure allowlist.

**That direction earned its place on its first run.** It immediately found three live registrations the scanner could not see — `object:onDidCreateNetworkObject`, `object:onDidDisposeNetworkObject`, `shared-store:change`, all from `createCoreMultiSourceEventEmitter`, which is core-internal and so missed by the patterns modelled on the public API. It later found a fourth, `paratextBibleSendReceive.onSyncWriteLockChanged`, registered from C# through `SendRequestAsync("network:registerEvent", [...])`. All four are now recognised — the fourth by the Roslyn C# half's rule E.

It lives in `tests/smoke/` deliberately: CI runs only `test:e2e:smoke`. It also **fails rather than skips** when the document cannot be fetched — the existing `papi-live.fixture` is built to skip gracefully when no app is running, which would make this report success without comparing anything.

## Three categories, because "declared" and "live" are not the same set

- **Declared** (187) — normal surface, asserted live.
- **Dynamic** (28) — no static name; recorded with the unresolved expression, excluded from comparison by pattern.
- **Liveness-annotated** (5) — declared, but a live poll can never durably catch them: three are **transient** (`testMain`, `testExtensionHost`, `platform.placeholder` self-dispose after 20 s / 10 s / 3 s) and two are **lazy** (`platformScriptureEditor.onWillSwitchProject` / `onDidSwitchProject`, created only on a project switch). They stay in the snapshot — they *are* declared surface — and are excluded only from the live comparison.

The transient case is worth flagging to a reviewer because it inverts the usual instinct: **waiting longer makes those less likely to be observed, not more.** No readiness tuning reaches them; they needed a category.

The annotations are emitted by the generator, not hand-kept in the test, so the snapshot documents its own comparability. **The generator throws if an annotated name no longer matches a registration** — a stale annotation would otherwise silently shrink what gets compared while the file still looked healthy. Transience and laziness are annotated deliberately rather than auto-detected: `setTimeout(x.dispose, N)` is detectable-but-fragile and "created inside a conditional handler" is not reliably detectable, and a scanner guessing wrong there would drop real surface from the comparison.

## What the second opinion found, and what changed

A second-opinion review confirmed twelve issues against this PR's earlier tip (`d29379a7245`, 2026-09-02). All twelve are now resolved — the C# half was rebuilt on Roslyn (`c-sharp/Paranext.WireSurface`, above) and the TypeScript half's name/documentation resolution now binds through the TypeScript checker instead of a hand-rolled lookup:

| # | Item | Resolution |
|---|---|---|
| 1 | Extensions calling the public `papi.networkObjects.set` alias were invisible to the scanner | TS scanner now recognises the alias |
| 2 | The live `rpc.discover` backstop blanket-accepted every `command:`-prefixed name and the `-pdpf`/`-webViewProvider` suffix families with no cross-check | now corroborated against the snapshot's declared/dynamic entries before being accepted |
| 3 | A C# `DataProvider` subclass could vanish silently (generic class name, nested parentheses, traditional-plus-primary-constructor mix) | Rule B resolves by symbol identity, not text shape, with fixtures for all three shapes |
| 4 | Same-named consts/vars/fields resolved first-wins, with no scope model | C#: every reference binds to its own declaration through the semantic model; TS: checker-based identifier binding |
| 5 | The premise that this toolchain has no Roslyn tooling was false in three places | removed from every copy — header, generator, ADR, and this PR body |
| 6 | `PATH_ALIASES` covered only 6 of 7 tsconfig path aliases (`@assets/*` missing), with nothing keeping the two lists in sync | now derived from `tsconfig.json`'s `paths` at runtime, with a test asserting the derived list matches |
| 7 | `CSHARP_EXCLUDED_PATTERNS` was always empty, and its test was vacuous | three real exclusions, each with its own fixture test |
| 8 | The determinism test replayed the module-scope declarations cache instead of proving determinism independently | test isolates module state between the two runs it compares |
| 9 | The `SendReceiveBlockNotifierService` collision probe described a real mechanism but was unverified | moot once resolution is by symbol binding; the notifier shape (two instantiation sites, four declared entries) is now a fixture |
| 10 | A `GetNetworkObjectDocumentation` override was attributed to the file's first `class` token | Rule C attributes to the overriding method's own `ContainingType` |
| 11 | Ordinary C# formatting (object-initializer syntax, a trailing comment, a comma inside a generic argument) could flip a recorded `experimental` value to false | `experimental` is read off the `Experimental` property *symbol*, never identifier text; three fixtures pin the formatting variants |
| 12 | The comment/string masking layer the old scanner relied on had no load-bearing test | the masking primitive no longer exists; the new load-bearing primitive (symbol identity of the invoked method) has its own exclusion fixtures |

## Caveats a reviewer should weigh

- **The readiness poll shares a gate with an existing flake.** It waits on the same extension-host startup that `ui-interaction.spec.ts` depends on, so on a loaded runner this test fails alongside that one rather than independently. It adds no new flake surface, but a slow macOS/Windows runner will show two failures instead of one — that is a readiness timeout, not a wire-surface regression, and the failure message says so.
- **`DEV_NOISY` gating is handled for the env-var route only.** The sample extensions and two C# registrations exist only in noisy dev mode, and a static scan cannot see a runtime env gate. The test excludes them when the mode is off — but a caller passing `DEV_NOISY` through `envOverrides` never reaches the test process's environment, and the comment says to extend the list if the fixture ever launches that way.
- **It asserts nothing about which registrations *should* be experimental**, by design.
- **The file set comes from the git index, the bytes come from the working tree.** `git ls-files` chooses what to scan and `readFileSync` supplies the contents. Three consequences worth knowing: an uncommitted edit to a tracked file does change the snapshot; a tracked file missing from the working tree dies on a bare `ENOENT` instead of this generator's deliberate loud error; and during a conflicted merge or rebase `git ls-files` emits each unmerged path once per stage, so those files are scanned two or three times and duplicated into both the registrations and the input digest. (The C# tool's own `--tracked-files` list is drawn from that same git-index set and reads its bytes from the compiled project on disk — but responds to a mismatch loudly, exiting 3, rather than silently; see "How the C# half works" above.)

## Verification

`npm run typecheck` · `npm run lint` · `npm run format:check` — all clean.

**C# tool timing:** `MSBuildWorkspace` open ≈0.7 s, compile ≈1.4 s; the whole `build:wire-surface` step runs in ≈8 s once the tool itself has been built by an earlier `dotnet run`.

**Tests:** `Paranext.WireSurface.Tests` — 90 passing (NUnit; in-memory fixture compilations against stub framework types, plus one `[Category("Integration")]` test that drives the real `MSBuildWorkspace` path against a temp project with a real `dotnet restore`). TypeScript: 80 tests in `.erb/scripts`, 88 in the e2e reduction-util fixtures. Whole-repo battery green: typecheck, lint, format:check, 3,573 + 2,288 vitest tests, `dotnet csharpier --check`, and 1,725 `c-sharp-tests`.

`npm run build:wire-surface` regenerates byte-identically across repeated runs (checked twice; identical digests both times).

**Live check:** the Linux smoke spec `wire-surface-snapshot.spec.ts` passed locally against the running app (33.6 s) — the 5 liveness-annotated registrations excluded as designed, nothing missing, nothing unrecognised.

**Red-first, from the original push (carried forward — not re-run this session):** with `x-experimental` removed from `platform.createWindow` in source and the snapshot left untouched, the assertion fails and names it —

```
Experimental-marker disagreements (1)
  - command 'platform.createWindow' (typescript, src/main/main.ts):
    declared experimental=true, live 'command:platform.createWindow' reports x-experimental=false
```

`npm run typecheck:e2e` (which now covers the reduction fixture) is clean on this tip.

**CI:** see checks. The previous tip's run failed only on a macOS Electron-download 504 that fail-fast cancelled the Linux job before it could report — not a defect in this change. This tip's CI run is in progress as this is written, including the two new steps: `dotnet test c-sharp/Paranext.WireSurface.Tests/Paranext.WireSurface.Tests.csproj`, and a warm `dotnet build c-sharp/ParanextDataProvider.csproj` ahead of the Linux smoke suite (the dev app starts the provider with `dotnet watch`, whose cold Debug build on a fresh runner was eating the whole 240 s readiness budget — no C# registration showed up in 5.5 minutes on a first launch, versus all present within 50 s on a relaunch).

## Regenerate this file after every rebase

`wire-surface.json` is a generated artifact under CI's `verify-changed-files` gate, and **CI
regenerates it from the pull request merged into main, not from the branch alone**. So a branch
behind main produces a snapshot the check cannot match, and the failure looks like a code bug rather
than a stale artifact. `papi.d.ts` already has this property, so this is the existing convention
rather than a new burden — but it is worth stating, because it cost four CI runs to recognise here.

**After any rebase: `npm run build:wire-surface`, then commit the regeneration before pushing.**

## What it took to find that, since the diagnostics are part of the deliverable

The gate failed on Windows three times. Three fixes went in first, and **each was a real defect that
was not the cause**:

- Sorting via `localeCompare`, which consults the host locale and ICU build — genuinely
  platform-dependent in a byte-exact artifact.
- The generator running inside the `concurrently` block, racing `build:extensions`, which writes into
  the trees it scans.
- The filesystem walk itself, which made the scanned set depend on whether a build had run.

All three are fixed and all three were worth fixing. None explained the failure.

What located it was a digest over the **scan's inputs**, printed to the build log — not over its
output, which merely restates what the gate already reported. The readout was immediate: Linux 787
files, CI 791. And the detail that had been visible for two runs before I registered it — **macOS and
Windows produced the identical output digest and disagreed only with my branch.** Two platforms
agreeing with each other is not a platform bug; the base had moved.

The scanned-file counts and both digests stay in the build output permanently. They are one log line
per build and the only cross-platform determinism evidence this repo has: an input digest that
matches with a differing output means the generator is at fault, and differing inputs mean the tree
is.

## Since the first push

The earlier text-based C# scanner went through several rounds of latent-defect fixes (envelope-level `Experimental` reads, first-`: base(` attribution, an over-confident `docsStaticallyResolved`, a parse-time silent skip). That scanner and its fixes are gone: the Roslyn tool above replaces it wholesale, and the failure modes those fixes patched cannot arise when recognition is by symbol and a compile error stops the generator.

## Risk Level

Low — additive tooling. One production-path edit: `ParanextDataProvider.csproj`'s `DefaultItemExcludes` now also excludes the new `Paranext.WireSurface`/`Paranext.WireSurface.Tests` folders from compilation, mirroring the existing `Paranext.Analyzers` exclusion. The build gains one `dotnet run` step (`build:wire-surface`) and two CI steps (a `dotnet test` for the new test project, a warm `dotnet build` ahead of the Linux smoke suite); the ordinary failure mode is a stale checked-in file that CI reports as a diff.

AI-assisted — [session 1](https://claude.ai/code/session_018FNQ8Ewcs6pPg2i3qB3APM), [session 2](https://claude.ai/code/session_01UXGcPnMhm1skh1YgR2FTGM)

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2734)


rolfheij-sil:
Putting this on the back burner. It's not exactly what TJ was hoping for. We might get back to it later, or we might not.

PR #2738, opened 2026-08-31
PT-4421: Fix the Rules-of-Hooks violation that blanks the editor

katherinejensen00:
# PT-4421: Fix the Rules-of-Hooks violation that blanks the editor

**Branch**: pt-4421-rules-of-hooks-blank-editor · **Base**: origin/main · **Date**: 2026-08-30 ·
**Review model**: Claude Opus 5 · **Files changed**: 11

## Summary

`useData` had a runaway-render guard that, when it tripped, took an early `return` skipping every
hook below it. React threw mid-render; with no error boundary in the app, the web view's React root
unmounted — the blank editor. **The guard's trip path had never once worked: every trip since it
shipped was a crash.**

The fix runs all hooks unconditionally with one conditional return at the end. Correcting the crash
then forced three follow-on decisions about what the guard does, since consumers now *observe* a
tripped state instead of dying: it counts deliveries **and** resubscribes (not renders), it degrades
to a `RESOURCE_EXHAUSTED` error with `isLoading: true`, and a trip **expires** after 5s instead of
latching forever.

No public signature changed. `useSetting`'s setter is now always callable (it could previously be
`undefined` at runtime despite its non-optional type), which made a rejection path reachable at three
fire-and-forget call sites — those are fixed too.

## Where to look

**Two files need real review. The other nine are mechanical, generated, or docs.**

| File | What it is | Effort |
| --- | --- | --- |
| `create-use-data-hook.util.ts` | **The fix.** New `useRunawayLoopGuard` + the restructured hook body | ~15 min |
| `create-use-data-hook.util.test.ts` | **The evidence.** 20 tests; two mutation-verified | ~10 min |
| `use-setting.hook.ts` + `.test.ts` | Fallback setter so a throttled write rejects instead of throwing `is not a function` | ~3 min |
| `user-profile-popover.component.tsx`, `hello-rock3.web-view.tsx` | Mechanical: sync `try`/`catch` → `.catch()` at 3 call sites | skim |
| `use-data.hook.ts`, `use-project-data.hook.ts`, `use-project-setting.hook.ts` | TSDoc only — documents the throttled state for extension authors | skim |
| `Architecture-Decisions.md` | New ADR with the rejected alternatives | read if curious |
| `lib/papi-dts/papi.d.ts` | **Generated.** Only the doc blocks above | skip |

**Fastest useful path:** read `useRunawayLoopGuard` (top of `create-use-data-hook.util.ts`), then the
`runaway-loop guard` describe block in the test file. That is the whole change; everything else falls
out of it.

## The one thing worth pushing back on

A trip now **expires after 5s and re-arms** — PT-1561 made it permanent. This is the only place the
change overrides the prior design rather than restoring it, so it's the most useful thing to
challenge.

The short version: both designs stop the loop (the subscription is dropped either way) — the only
question is whether the stop expires. Latching buys ~2% of the damage (a stuck consumer costs ~20
events/s re-arming vs. ~1000/s free-running — re-arming already removes ~50×). It costs the entire
false-positive case, unbounded: the guard fires on a *rate* signal, which cannot tell a bug from a
Send/Receive burst, and a wrong guess under a latch is recoverable only by closing the tab.

Full argument, including the conceded limit and why not escalating backoff, is in **Design decision**
below.

<details>
<summary><b>Design decision — why a trip expires rather than latching</b></summary>

- Both designs stop the loop; the subscription is dropped the moment the guard trips either way. The
  only question is whether the stop expires.
- **What permanence buys: ~2% of the damage.** The guard trips at 100 events in a 1s window, then
  sits tripped for 5s with fresh counters — ~100 events per ~5.1s ≈ 20/s, against a measured
  free-running ~1000/s. Re-arming already takes a runaway down ~50×. Latching buys only the remaining
  ~20 events/s, in a state that is already broken and already logging.
- **What permanence costs: the whole false-positive case, unbounded in time.** A rate signal cannot
  distinguish a bug from a burst — S/R and bulk import produce the identical shape. Under a latch one
  legitimate burst leaves that pane throttled with no user-visible explanation, recoverable only by
  closing the tab. A large open-ended harm traded for a small closed one.
- **The latch has an unpaid bill.** Permanently stopping a pane is a state the user must be told
  about and given a way out of — a banner, a "reload this pane" control. PT-1561 never built one and
  neither does this PR, so "permanent" would mean silently and irreversibly broken. Latching is a
  legitimate position; it just comes with UI work far larger than this change.
- **The repeated warning is the diagnostic, not the defect.** A latched trip warns once and goes
  quiet, which in the logs is indistinguishable from a burst that resolved on its own. A warning
  every 5s for as long as the consumer is broken is what separates "someone ran an S/R" from "this
  component has an unstable selector". PT-1561's trip path crashed the web view from the day it
  shipped and nobody caught it — that is what a once-and-done signal buys.
- **Conceded limit:** re-arming bounds *rate*, not *total*. If a stuck consumer accumulates something
  a trip does not release (memory, subscriptions, IPC backlog), throttling moves the wall out rather
  than removing it. A named accumulation that survives a trip is a real argument for latching.
- **Why a timer, not selector-change:** an unstable selector gets a new identity every render, so
  re-arming on it would disarm the guard entirely.
- **Why not escalating backoff** (5s, 10s, 20s…): it converges on permanent for a stuck consumer
  while staying cheap for a burst, and is the natural "can we have both?" — deferred because the
  fixed cooldown already captures the ~50× and backoff adds state to reason about and test. A good
  follow-up, not a loss.

</details>

<details>
<summary><b>Why the guard counts what it counts</b></summary>

A runaway shows up as two distinct loops, and watching only one misses the other:

- **Delivery loop** — delivery → re-render → delivery.
- **Resubscribe loop** — an unstable selector rebuilds the subscription every render. Each
  subscription is superseded before it resolves, so `useEventAsync` mutes every emission and **nothing
  is ever delivered**. Measured: 201 subscribe attempts in 200ms with **zero** warnings under
  delivery-only counting. This is the loop the warning's own "memoize your parameters" advice
  describes.

**Renders are deliberately not counted** — a busy consumer is not a broken one, and render-counting
is what produced the original false positive (a cold-start render storm tripping a guard about data
updates).

</details>

<details>
<summary><b>API changes</b></summary>

- `useData`, `useProjectData`, `useProjectSetting`: no signature change; TSDoc added for the
  throttled state (`data` = `PlatformError` with code `RESOURCE_EXHAUSTED`, `setData` = `undefined`,
  `isLoading` = `true`).
- `useSetting`: no signature change; **runtime behavior changed** — the returned setter is now always
  a callable function. Previously it could be `undefined` at runtime despite the non-optional
  declared type.
- Behavioral change inside the existing `useData`/`useProjectData` signature when the guard trips:
  was `[PlatformError, undefined, false]` and permanent; now `[PlatformError, undefined, true]` and
  self-expiring, with the subscription actually dropped.
- `lib/platform-bible-react/`, `lib/platform-bible-utils/`, extension `types/*.d.ts`: unchanged.
- `lib/papi-dts/papi.d.ts`: regenerated; only the doc blocks above.

</details>

<details>
<summary><b>Review findings — 16 fixed, 2 dismissed, 0 unresolved</b></summary>

Four analysis passes ran (API/correctness, style/architecture, coverage/compliance, UX). No Critical
findings. Everything below marked `[x]` was fixed during the review.

**Important**

- [x] **Async rejection escaped synchronous `try`/`catch` at setting-write call sites.**
  `safeSetSetting` is `async`, so a sync `catch` cannot see the rejection — unhandled rejection and a
  silently dropped user action. _(fixed: all three broken sites converted to `.catch()`. Audited all
  five sites that destructure a setting setter; `language.component.tsx` was already correct, and
  `enhanced-resource.web-view.tsx` was repaired by the `useSetting` change itself.)_
- [x] **Public TSDoc named a parameter that cannot cause the problem** — it told developers to check
  `subscriberOptions`, which is held as a ref and excluded from the subscribe memo's deps. _(fixed:
  corrected to `selector` / `dataProviderSource`.)_
- [x] **Fallback setter message contradicted the design in the same diff** — said "Reopen this view
  to retry" when the guard re-arms on its own. _(fixed: reworded.)_
- [x] **`safeSetSetting` was untested** — no `use-setting.hook.test.ts` existed. _(fixed: added,
  mutation-verified.)_

**Minor**

- [x] Re-arm did not restore `isLoading`, undoing the `isLoading: true` decision exactly when it
  mattered. _(fixed + test)_
- [x] Negative guard tests filtered `logger.warn` on `'was updated'`, so a spurious *subscribe* trip
  would have left them green. _(fixed)_
- [x] `toBeLessThanOrEqual(RUNAWAY_THRESHOLD)` passed at exactly 100 and would not catch an
  off-by-one. _(fixed)_
- [x] Subscribe counter had no below-threshold / outside-window negative test. _(fixed)_
- [x] Unmount-during-cooldown cleanup was untested. _(fixed)_
- [x] `useSetting`'s type assertion declared the setter non-optional while the code guarded it.
  _(fixed: widened to `| undefined`)_
- [x] `use-project-setting.hook.ts` missed the throttling doc note its siblings got. _(fixed)_
- [x] Tests hardcoded `101`/`99`/`130`/`150` despite a `RUNAWAY_THRESHOLD` constant. _(fixed)_
- [x] `_＠throttling_` used the fullwidth-`＠` hack for something that is not a JSDoc tag. _(fixed)_
- [x] `trip(whatHappened)` built its sentence across two places without signalling the constraint.
  _(fixed: renamed)_
- [x] Rationale duplicated across code, ADR, and tests. _(fixed: code points at the ADR)_
- [x] ADR omitted the cost/benefit arithmetic, the "return last-known data" alternative, and backoff.
  _(fixed)_
- [ ] ~~Extract `useRunawayLoopGuard` to its own `.hook.ts` per naming convention.~~ _(Author:
  extract at two or three consumers, not one. It currently has a single consumer.)_
- [ ] ~~Collapse `recordDelivery`/`recordSubscribe`, which are structurally identical.~~ _(Author:
  collapse if the pattern is likely to be used a third time. Delivery and resubscribe are the
  complete set of loop shapes.)_

</details>

<details>
<summary><b>Verification — and what was <i>not</i> verified</b></summary>

Passing: `tsc --noEmit` (0 errors) · ESLint exit 0 on all changed files · Prettier clean ·
`npm run build:types` exit 0 with `papi.d.ts` diff limited to the added docs ·
`src/renderer/hooks` 9 files / 96 tests · guard suite 20/20 · scripture-editor extension 904/904 ·
renderer suite 1099/1108 (the 9 are known load-flakes; 93/93 in isolation).

**`rules-of-hooks` is now live on this file** — verified by linting three variants: **0** errors on
the original anonymous-arrow code, **11** when the same bug is reintroduced under the new name, **0**
on the fix. That is what stops this class of bug returning.

**Two guard tests are mutation-verified** — removing `hasTrippedRef` or the `recordSubscribe` call
makes the corresponding test fail. One of them previously *could not* fail, which is the exact
failure mode `.claude/rules/testing/tdd-discipline.md` warns about.

Not verified:

- **The full root `npm test` never finished** — three attempts, two died within seconds. The blast
  radius was verified instead (renderer suite, editor extension, hooks). CI will be the first true
  full run.
- **No packaged-build verification.** `electron-builder` stalls on the node-module scan in this
  worktree (29 siblings under `.claude/worktrees/`), so verification is from a dev build only.
- **The PRD's repro is a race** with no reliable manual trigger, so a clean manual run does not prove
  much. The deterministic tests are the real evidence.
- This branch went through two AI review rounds, and **both surfaced findings that were wrong on
  inspection** — one claimed the editor shows an editable pane behind an overlay when it actually
  shows a spinner. Findings were verified rather than taken at face value.

</details>

<details>
<summary><b>Positive observations from the analysis passes</b></summary>

- The fix is structural, not a patch: naming the returned function `useDataForDataProvider` is what
  lets `react-hooks/rules-of-hooks` lint this body at all.
- Fixes a latent bug in the old trip path — `newPlatformError` was passed the `useRef` **object**
  rather than `.current`, so the error never carried its intended message.
- `performance.now` is mocked only inside the guard `describe`; React's scheduler reads the same
  clock, so freezing it globally would perturb the unrelated subscribe/teardown race tests.
- `getCounters()` returns a non-nullable value, keeping the guard fail-closed; replacing counters at
  re-arm rather than resetting avoids re-tripping on stale ring-buffer timestamps.
- Reuses existing primitives (`EventRollingTimeCounter`, the existing `RESOURCE_EXHAUSTED` code)
  rather than inventing new ones.
- `isLoading: true` was verified end-to-end: `setting.component.tsx` branches on it to render a
  localized loading label, so a throttled setting shows a proper message rather than a blank region.

</details>

## Suggested review focus

- [ ] **Trip expiry vs. latching** — the one decision that overrides prior design. The counter worth
  testing: does any accumulation survive a trip (memory, subscriptions, IPC backlog)? That would
  favor latching or backoff.
- [ ] **`isLoading: true` while throttled** — confirm no consumer treats a long `isLoading` as a hang.
- [ ] **The `useSetting` runtime contract change** — setter always callable, rejects instead of being
  `undefined`. Three call sites converted; confirm the audit covered what you'd expect.
- [ ] **Verification gaps** — full `npm test` and packaged build were both blocked locally. Agree
  that CI coverage is sufficient before merge?

AI-assisted — [session](https://claude.ai/code/session_01Xq6QF4fjtCCcXfK6MFVSFd)

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2738)

### DIFF
--- a/src/renderer/hooks/papi-hooks/use-setting.hook.ts
+++ b/src/renderer/hooks/papi-hooks/use-setting.hook.ts
@@ -29,7 +29,8 @@ import { useCallback } from 'react';
  *
  *   - `setting`: The current state of the setting, either `defaultState`, the stored value, or a
  *       `PlatformError` if loading the value fails. Use `isPlatformError()` to check.
- *   - `setSetting`: Function that updates the setting to a new value
+ *   - `setSetting`: Function that updates the setting to a new value. Rejects while the underlying
+ *       subscription is throttled — see {@link useData} for that state.
  *   - `resetSetting`: Function that removes the setting and resets the value to `defaultState`
  *
  * @throws When subscription callback function is called with an update that has an unexpected
--- a/src/renderer/hooks/papi-hooks/use-setting.hook.ts
+++ b/src/renderer/hooks/papi-hooks/use-setting.hook.ts
@@ -1,10 +1,9 @@
-import { newPlatformError, PlatformError, RESOURCE_EXHAUSTED } from 'platform-bible-utils';
+import { PlatformError } from 'platform-bible-utils';
 import { useData } from '@renderer/hooks/papi-hooks/use-data.hook';
 import {
   DataProviderSubscriberOptions,
   DataProviderUpdateInstructions,
 } from '@shared/models/data-provider.model';
-import { logger } from '@shared/services/logger.service';
 import { settingsService } from '@shared/services/settings.service';
 import { SettingDataTypes } from '@shared/services/settings.service-model';
 import { SettingNames, SettingTypes } from 'papi-shared-types';
@@ -29,8 +28,8 @@ import { useCallback } from 'react';
  *
  *   - `setting`: The current state of the setting, either `defaultState`, the stored value, or a
  *       `PlatformError` if loading the value fails. Use `isPlatformError()` to check.
- *   - `setSetting`: Function that updates the setting to a new value. While the underlying subscription
- *       is throttled it rejects with a `PlatformError` whose `code` is `RESOURCE_EXHAUSTED` — see
+ *   - `setSetting`: Function that updates the setting to a new value, or `undefined` while there is
+ *       nothing to write through — including while the underlying subscription is throttled, see
  *       {@link useData} for that state.
  *   - `resetSetting`: Function that removes the setting and resets the value to `defaultState`
  *
@@ -43,9 +42,11 @@ export const useSetting = <SettingName extends SettingNames>(
   subscriberOptions?: DataProviderSubscriberOptions,
 ): [
   setting: SettingTypes[SettingName] | PlatformError,
-  setSetting: (
-    newData: SettingTypes[SettingName],
-  ) => Promise<DataProviderUpdateInstructions<SettingDataTypes>>,
+  setSetting:
+    | ((
+        newData: SettingTypes[SettingName],
+      ) => Promise<DataProviderUpdateInstructions<SettingDataTypes>>)
+    | undefined,
   resetSetting: () => void,
   isLoading: boolean,
 ] => {
@@ -61,8 +62,6 @@ export const useSetting = <SettingName extends SettingNames>(
         subscriberOptions?: DataProviderSubscriberOptions,
       ) => [
         setting: SettingTypes[SettingName] | PlatformError,
-        // `useData` drops its setter while its runaway guard is throttled, so this really can be
-        // `undefined` — asserting otherwise would make the guard below look like dead code
         setSetting:
           | ((
               newData: SettingTypes[SettingName],
@@ -78,23 +77,6 @@ export const useSetting = <SettingName extends SettingNames>(
     settingsService.reset(key);
   }, [key]);
 
-  // `useData` drops its setter when its runaway guard trips. This hook's returned type promises a
-  // callable setter, so call sites do not null-check it — substitute one that rejects rather than
-  // letting `setSetting is not a function` throw synchronously out of an event handler.
-  const safeSetSetting = useCallback(
-    async (newData: SettingTypes[SettingName]) => {
-      if (setSetting) return setSetting(newData);
-      // Rejects with the same `RESOURCE_EXHAUSTED` code `useData` puts on its throttled value, so a
-      // `.catch` can recognize the throttled state with `isPlatformError(e) && e.code ===
-      // RESOURCE_EXHAUSTED` instead of matching the message text. The message itself is
-      // developer-facing English, not localized — see {@link useData}.
-      const message = `Cannot set setting ${key}: its data subscription is temporarily throttled. It will retry on its own shortly.`;
-      logger.warn(message);
-      throw newPlatformError(message, RESOURCE_EXHAUSTED);
-    },
-    [key, setSetting],
-  );
-
-  return [setting, safeSetSetting, resetSetting, isLoading];
+  return [setting, setSetting, resetSetting, isLoading];
 };
 export default useSetting;
--- a/src/renderer/hooks/papi-hooks/use-setting.hook.ts
+++ b/src/renderer/hooks/papi-hooks/use-setting.hook.ts
@@ -1,4 +1,4 @@
-import { PlatformError } from 'platform-bible-utils';
+import { newPlatformError, PlatformError, RESOURCE_EXHAUSTED } from 'platform-bible-utils';
 import { useData } from '@renderer/hooks/papi-hooks/use-data.hook';
 import {
   DataProviderSubscriberOptions,
@@ -29,8 +29,9 @@ import { useCallback } from 'react';
  *
  *   - `setting`: The current state of the setting, either `defaultState`, the stored value, or a
  *       `PlatformError` if loading the value fails. Use `isPlatformError()` to check.
- *   - `setSetting`: Function that updates the setting to a new value. Rejects while the underlying
- *       subscription is throttled — see {@link useData} for that state.
+ *   - `setSetting`: Function that updates the setting to a new value. While the underlying subscription
+ *       is throttled it rejects with a `PlatformError` whose `code` is `RESOURCE_EXHAUSTED` — see
+ *       {@link useData} for that state.
  *   - `resetSetting`: Function that removes the setting and resets the value to `defaultState`
  *
  * @throws When subscription callback function is called with an update that has an unexpected
@@ -83,9 +84,13 @@ export const useSetting = <SettingName extends SettingNames>(
   const safeSetSetting = useCallback(
     async (newData: SettingTypes[SettingName]) => {
       if (setSetting) return setSetting(newData);
+      // Rejects with the same `RESOURCE_EXHAUSTED` code `useData` puts on its throttled value, so a
+      // `.catch` can recognize the throttled state with `isPlatformError(e) && e.code ===
+      // RESOURCE_EXHAUSTED` instead of matching the message text. The message itself is
+      // developer-facing English, not localized — see {@link useData}.
       const message = `Cannot set setting ${key}: its data subscription is temporarily throttled. It will retry on its own shortly.`;
       logger.warn(message);
-      throw new Error(message);
+      throw newPlatformError(message, RESOURCE_EXHAUSTED);
     },
     [key, setSetting],
   );
--- a/src/renderer/hooks/papi-hooks/use-setting.hook.ts
+++ b/src/renderer/hooks/papi-hooks/use-setting.hook.ts
@@ -4,6 +4,7 @@ import {
   DataProviderSubscriberOptions,
   DataProviderUpdateInstructions,
 } from '@shared/models/data-provider.model';
+import { logger } from '@shared/services/logger.service';
 import { settingsService } from '@shared/services/settings.service';
 import { SettingDataTypes } from '@shared/services/settings.service-model';
 import { SettingNames, SettingTypes } from 'papi-shared-types';
@@ -71,6 +72,19 @@ export const useSetting = <SettingName extends SettingNames>(
     settingsService.reset(key);
   }, [key]);
 
-  return [setting, setSetting, resetSetting, isLoading];
+  // `useData` drops its setter when its runaway guard trips. This hook's returned type promises a
+  // callable setter, so call sites do not null-check it — substitute one that rejects rather than
+  // letting `setSetting is not a function` throw synchronously out of an event handler.
+  const safeSetSetting = useCallback(
+    async (newData: SettingTypes[SettingName]) => {
+      if (setSetting) return setSetting(newData);
+      const message = `Cannot set setting ${key}: its data subscription was stopped. Reopen this view to retry.`;
+      logger.warn(message);
+      throw new Error(message);
+    },
+    [key, setSetting],
+  );
+
+  return [setting, safeSetSetting, resetSetting, isLoading];
 };
 export default useSetting;
--- a/src/renderer/hooks/papi-hooks/use-setting.hook.ts
+++ b/src/renderer/hooks/papi-hooks/use-setting.hook.ts
@@ -60,9 +60,13 @@ export const useSetting = <SettingName extends SettingNames>(
         subscriberOptions?: DataProviderSubscriberOptions,
       ) => [
         setting: SettingTypes[SettingName] | PlatformError,
-        setSetting: (
-          newData: SettingTypes[SettingName],
-        ) => Promise<DataProviderUpdateInstructions<SettingDataTypes>>,
+        // `useData` drops its setter while its runaway guard is throttled, so this really can be
+        // `undefined` — asserting otherwise would make the guard below look like dead code
+        setSetting:
+          | ((
+              newData: SettingTypes[SettingName],
+            ) => Promise<DataProviderUpdateInstructions<SettingDataTypes>>)
+          | undefined,
         boolean,
       ];
     }
@@ -79,7 +83,7 @@ export const useSetting = <SettingName extends SettingNames>(
   const safeSetSetting = useCallback(
     async (newData: SettingTypes[SettingName]) => {
       if (setSetting) return setSetting(newData);
-      const message = `Cannot set setting ${key}: its data subscription was stopped. Reopen this view to retry.`;
+      const message = `Cannot set setting ${key}: its data subscription is temporarily throttled. It will retry on its own shortly.`;
       logger.warn(message);
       throw new Error(message);
     },
--- a/.context/standards/Architecture-Decisions.md
+++ b/.context/standards/Architecture-Decisions.md
@@ -2497,3 +2497,101 @@ step, no automation. Just a record.
   worse than a badly-timed foreground.
 - **Source:** PR #2670 review item 6 (2026-08-25) and the review rounds that followed; PT-4465,
   which carries the design, the call-site table and the `show` hazard in full.
+
+## adr-runaway-data-hook-guard: `useData`'s runaway guard counts subscribes and deliveries, degrades rather than throws, and expires
+
+- **Date:** 2026-08-28
+- **Status:** Accepted
+- **Context:** `create-use-data-hook.util.ts` carries a circuit breaker (PT-1561) meant to stop a
+  `useData` consumer from spinning in a render loop. It counted **renders** in a dependency-less
+  effect, then took an early `return` that skipped every hook below it — changing the hook count
+  between renders. React threw mid-render, and because the app has no error boundary, the web view's
+  React root unmounted: the blank editor. The trip path had therefore never once degraded
+  gracefully. Fixing the hook order forced three policy questions about the platform's most-consumed
+  hook, since consumers now observe a tripped state instead of a crash.
+- **Decision:** (1) Count **deliveries and subscribe attempts**, never renders. (2) A trip degrades:
+  the subscription is dropped, `[PlatformError, undefined, true]` is returned, and the error carries
+  the `RESOURCE_EXHAUSTED` code. (3) A trip **expires** after a cooldown, then re-arms and
+  resubscribes. (4) Every hook in the family reports the dropped setter the same way — as
+  `undefined`. `useData` and `useProjectSetting` already did; `useSetting` is widened to match
+  rather than fabricating a substitute setter to keep its published type non-optional.
+- **Alternatives:**
+  - *Keep counting renders* — rejected: a consumer can re-render rapidly for reasons unrelated to
+    this data (a busy parent, a cold-start storm, fast typing), and stopping a healthy subscription
+    over that is wrong. This was the false positive behind the original report.
+  - *Count deliveries only* — rejected once measured: an unstable selector rebuilds the subscription
+    every render, and each subscription is superseded before it resolves, so `useEventAsync` mutes
+    every emission and nothing is ever delivered. A probe produced 201 subscribe attempts in 200ms
+    with zero warnings — invisible to delivery counting, and it is precisely the loop the warning's
+    "memoize your parameters" advice describes.
+  - *Report `isLoading: false` while tripped* — rejected: it tells consumers that gate on loading
+    that this IS the answer, resolving a three-state interface-mode check to the wrong mode and
+    rendering raw localization keys as real strings. The value has not resolved and will be retried,
+    so `true` is the honest report.
+  - *Keep the trip permanent* (PT-1561's behavior) — rejected. Both designs stop the loop; the
+    subscription is dropped the moment the guard trips either way. The only question is whether the
+    stop expires. Priced out: the guard trips at 100 events in a 1s window and then sits tripped for
+    5s with fresh counters, so a genuinely stuck consumer costs ~100 events per ~5.1s ≈ 20/s against
+    a measured free-running ~1000/s — re-arming already removes ~98% of the damage, and latching
+    buys only the remaining ~20 events/s in a state that is already broken and already logging.
+    Against that, latching makes the false-positive case unbounded in time: the guard fires on a
+    *rate* signal, which cannot distinguish a bug from a Send/Receive burst, and a wrong guess under
+    a latch is recoverable only by closing the tab. A large open-ended harm traded for a small
+    closed one. Latching also has an unpaid bill — permanently stopping a pane is a state the user
+    must be told about and given a way out of, and neither PT-1561 nor this change builds that UI,
+    so "permanent" would mean silently unrecoverable. Finally, re-arming's repeated warning is the
+    diagnostic, not the defect: a latched trip warns once and goes quiet, which in the logs is
+    indistinguishable from a burst that resolved on its own, whereas a warning every 5s for as long
+    as the consumer is broken is what actually separates "someone ran an S/R" from "this component
+    has an unstable selector". PT-1561's trip path crashed the web view from the day it shipped and
+    nobody caught it, which is what a once-and-done signal buys.
+  - *Return the last delivered value instead of the error while tripped* — rejected: it removes the
+    only signal consumers have that anything is wrong, and a pane that looks normal while silently
+    refusing writes is more deceptive than one that reports an unresolved state. Note the accepted
+    design does mean a consumer substituting a fallback for a `PlatformError` will visibly swap
+    content for the cooldown's duration.
+  - *Escalating backoff* (5s, 10s, 20s…) — deferred, not rejected. It converges on permanent for a
+    truly stuck consumer while staying cheap for a burst, and is the natural answer to "can we have
+    both?". Not done here because the fixed cooldown already captures the ~50x reduction and backoff
+    adds state that has to be reasoned about and tested. Tracked as `PT-4502`.
+  - *Re-arm on selector or data provider change* — rejected as actively harmful: an unstable
+    selector, the very mistake being reported, gets a new identity every render, so this would reset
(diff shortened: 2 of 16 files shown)


jolierabideau (changes requested):
## Review summary — PT-4421

The core fix is right, and the PR description does an unusually good job of making it reviewable. The Rules-of-Hooks restructure is the correct shape: every hook runs unconditionally and there is exactly one conditional return at the bottom. Naming the returned function `useDataForDataProvider` so `react-hooks/rules-of-hooks` can actually see the body is the part that keeps this class of bug from coming back, and it's a better fix than just moving the `return`.

### What I verified locally

- `create-use-data-hook.util.test.ts` (20) + `use-setting.hook.test.ts` (2) — **22/22 pass** on `4827512`.
- ESLint exit 0 on all five changed source files.
- `npx eslint --print-config src/renderer/hooks/hook-generators/create-use-data-hook.util.ts` resolves `react-hooks/rules-of-hooks` to `['error']` — the enforcement claim holds; the only `off` override in `.eslintrc.js` is scoped to `e2e-tests/**`.
- **The `useSetting` setter audit checks out.** All five call sites that destructure a setting setter are handled (`user-profile-popover` ×2, `first-run/steps/language.component.tsx`, `hello-rock3`, `enhanced-resource.web-view.tsx:2717`). I also chased the one indirection the description doesn't mention: `useInterfaceMode` re-exports `setMode` with a non-optional type, and its only setter consumer is the popover, which is fixed. And `useProjectSetting` already declares `setSetting | undefined`, so there is no parallel gap in the sibling hook. Nothing missed.
- The trip → drop-subscription → re-arm path does what the tests say it does; the `getCounters()` fail-closed getter and replacing (rather than resetting) counters at re-arm both hold up on read.

---

## Blockers

**1. CI has never run on this PR.** `gh pr checks 2738` → "no checks reported"; `statusCheckRollup` is empty and the commit status is `pending`. Combined with the description's own "the full root `npm test` never finished — three attempts, two died within seconds," that means the most-consumed hook in the renderer would merge without a single full test run anywhere. Please get a green `test.yml` run before merge. (If checks aren't triggering, that itself needs a look.)

**2. PT-4421's Definition of Done isn't met yet.** The ticket asks for three things this PR explicitly doesn't do: *"Reproduce the known blank-editor scenario and confirm it no longer blanks"*, *"Verify on a packaged build"*, and *"no Power regression."* The description is honest that all three are open (the repro is a race, `electron-builder` stalls in the worktree). I'm not asking you to fight `electron-builder` in a worktree — but pick one: run the packaged check from a clean clone, or amend the DoD on PT-4421 with the reason. Right now the ticket and the PR disagree about what "done" means, and that's the kind of gap that gets discovered at release.

**3. `isLoading: true` + auto re-arm has no exit for a genuinely stuck consumer, and the ADR doesn't concede it.** You invited pushback on expiry-vs-latching, so here it is — but not the argument you pre-empted. I'm not asking you to latch. The cost/benefit for *rate* is sound and the false-positive argument is right.

The gap is that the ADR prices the re-arm cost as "a consumer substituting a fallback for a `PlatformError` will visibly swap content for the cooldown's duration" — singular, bounded. For a real runaway it isn't bounded, it's a cycle: trip → 5 s → re-arm → ~100 subscribes → trip. Concretely, in `platform-scripture-editor.web-view.tsx`:

- `:3641` — `if (isInterfaceModeLoading || isIsPublishedLoading) return <Spinner />`
- `:709` — `viewType` is pinned to `persistedViewType` while `isInterfaceModeLoading`

So if the `platform.interfaceMode` subscription trips, `isLoading` never resolves to `false`, and the user gets a spinner that never ends plus content flickering between the error branch and the stale value every 5 s — with no message and no way out but closing the tab. That is the *same* "unpaid bill" the ADR uses to reject latching ("permanently stopping a pane is a state the user must be told about and given a way out of… so 'permanent' would mean silently unrecoverable"). The shipped design owes the same bill; it just pays it in flicker instead of stillness.

I don't think this blocks the *design*, but it blocks merging it as an undocumented consequence. Either:
- surface the trip to the user once (a toast or a pane-level banner), or
- update the ADR's **Consequences** to state the repeat-cycle outcome plainly (perpetual `isLoading`, 5 s flicker, no user-visible signal) and open the escalating-backoff follow-up you already scoped, so the thing that bounds it has a ticket.

---

## NITs

1. **`useSetting` and `useProjectSetting` now disagree about the same condition.** `useProjectSetting` exposes `setSetting: (…) | undefined` and makes call sites null-check; `useSetting` keeps a non-optional public type and fabricates a rejecting stub. Both are defensible, but a reader hitting them back-to-back can't tell why. One sentence in the ADR on why the two diverge (presumably: widening `useSetting`'s public tuple is a breaking type change for extensions) would settle it.

2. **The rejecting setter throws a plain `Error`, not a `PlatformError` with `RESOURCE_EXHAUSTED`.** The whole point of adding the code was so consumers could recognize a throttled hook without matching on message text — and then the one path a consumer is most likely to `.catch()` drops it and hands back a string to substring-match. `use-setting.hook.ts:88`.

3. **The guard's constants aren't exported, so the test mirrors `RUNAWAY_THRESHOLD = 100` by hand.** It fails loudly rather than silently if they drift, so this is genuinely minor — but exporting `RUNAWAY_EVENTS_PER_WINDOW` / `RUNAWAY_WINDOW_MS` / `RUNAWAY_COOLDOWN_MS` and importing them in the test removes the mirror entirely.

4. **No test that two `useData` instances guard independently.** Per-instance containment is the entire containment story — one broken pane must not throttle a healthy one — and nothing currently pins it.

5. **`newPlatformError(message)` carries untranslated English that can reach the UI** wherever a consumer renders `error.message`. I think the intent is developer-facing (it reads like a console warning), which is fine — but say so in the TSDoc so nobody later "fixes" it by localizing it or surfaces it raw to a translator.

6. **Comment at the conditional return narrates the old bug.** "*Returning before a hook instead would change the hook count between renders, which makes React throw mid-render…*" — the invariant half ("every hook above runs on every render, so both paths render the same hooks in the same order") is worth keeping; the counterfactual half is commit-message material per `.claude/rules/code-quality/forward-facing-comments.md`. Same for the second half of the `useDataForDataProvider` comment. Very minor — the rest of the comments in this diff are genuinely forward-facing and good.

7. **`useRunawayLoopGuard` in a `.util.ts`.** I agree with your "extract at two or three consumers" call, so leave it — but the file now exports a hook that only gets linted *because* of its name, which is exactly the fragility this PR is fixing. A line at the top of the file saying so would be cheap insurance.

8. **Rebase before merge.** Merge base is `82ea330` (2026-08-26); `Architecture-Decisions.md` has grown ~650 lines on `main` since, and this PR appends to it. Should be a clean append, but worth doing before the merge queue rather than during.

9. Cosmetic: the description says "20 tests; two mutation-verified" for the guard — 20 is the whole file; the guard `describe` has 9.

---

### Scope note (not blocking)

PT-4421 asks for a hook-order fix and lint enforcement. This PR also changes *what* the guard counts, *whether the trip expires*, the error code it carries, `isLoading` semantics while tripped, and `useSetting`'s runtime contract. I think the ADR justifies every one of those as forced — you can't make a crash observable without deciding what consumers observe — and I'd rather have this than a narrow fix that leaves a guard nobody understands. But these are behavior changes to hooks extension authors consume, landing under a subtask whose title mentions none of them. Worth a line on PT-4421 (or a linked ticket) so it's findable at release-notes time.

---

Nice work on this one — the ADR, the resubscribe-loop measurement, and the mutation-verified tests are well above the bar. The blockers are process and one undocumented consequence, not the fix itself.

katherinejensen00 (commented):
Thank you so much for the review, Jolie! It was helpful.

The three blockers were the right three, and one of them paid off immediately.

Blocker 1 — CI has never run. You were right to hold, and it caught a live bug. I got the full npm test to finish and it surfaced a breakage this PR introduced: user-profile-popover.test.tsx was throwing two unhandled Cannot read properties of undefined (reading 'catch') errors from handleInterfaceModeChange and handleLanguageChange — the exact sites 4827512 converted to .catch(). The test's mock setters were bare vi.fn() returning undefined while the real setters return promises, so the double didn't match the contract. Vitest reports that in an "Errors" count separate from test results, which is easy to miss when the run never completes. Mocks now return promises; that file is 20/20 clean. So yes, this would have merged with something actually broken. Local run is now green: 2477/2477 tests across 176/176 files, dotnet 1703/0, lint, typecheck and format:check all clean. Pushing now so test.yml runs.

Blocker 2 — DoD. Amended PT-4421 rather than fighting electron-builder in a worktree. The DoD now records what was verified and adds a "Not verified, deliberately" section with the reason for each gap: the blank-editor repro is a race with no reliable manual trigger, so a clean manual run can't distinguish "fixed" from "didn't happen to hit it" — the deterministic tests and lint enforcement stand in; the packaged build stalls on the node-module scan in the dev worktree, so verification is dev build plus CI. Ticket and PR now agree.

Blocker 3 — the unconceded repeat cycle. Took the ADR option. Consequences now states the cycle plainly: trip → cooldown → re-arm → ~100 subscribes → trip for as long as the view is open, the permanent spinner, the five-second Simple/Power flip, no message and no way out but closing the tab — and says outright that this is the same unpaid bill the entry charges against latching, paid in flicker instead of stillness. Escalating backoff is now PT-4502, referenced from the ADR and from PT-4421.

One correction on specifics: I couldn't find persistedViewType in platform-scripture-editor.web-view.tsx — viewType comes from useWebViewState and isn't gated on isInterfaceModeLoading. The mechanism is isPowerMode,rror and returns false, dropping a Power user to Simple eachcycle. Same user-visible outcome, different line.

Your framing also turned up a consequence neither of us had listed. useInterfaceMode writes its resolved mode to localStorage, and mode reports 'simple'
for any PlatformError — so a five-second throttle was being pecomputeInitialStatus reads, routing a Power user through thefirst-run gate on the next launch. A transient outliving the session is worse than the flicker. Now gated on !isPlatformError(...).

NITs

1. Your guess was the reason, and it's now an explicit rejected alternative in the ADR. Confirmed first: useSetting is exported through papi.d.ts and its call sites, here and in extensions, call the setter directly — widening the tuple would be a breaking type change for all of them to describe a state lasting five seconds. useProjectSetting has always declared it optional, so its call sites already null-check.
2. Fixed — now throw newPlatformError(message, RESOURCE_EXHAUSTED), matching the existing pattern in overlay.service-host.ts. The test asserts the code, not the message text.
3. Done, all three exported and imported. Removing the mirror turned up more of it than the constant: 'value 98', 'value 149', the 10000 cooldown advance
   and the bare 20ms pacing were all silently derived from tho from the constants now.
4. Added, and mutation-verified — hoisting the counters to module scope fails it. It also asserts the healthy sibling keeps its setter, since containment that left the neighbour read-only wouldn't be containment.
5. Your read of the intent is right. Documented on the guard and on the public useData TSDoc, since that's where an extension author would look. Both say
   the text is developer-facing and deliberately unlocalized, ser-facing wording should branch on RESOURCE_EXHAUSTED andsupply their own.
6. Both trimmed. The invariant stays and now ends with "Any hook added to this function must go ABOVE this line," which is what a future maintainer needs. The useDataForDataProvider comment points at the new file-level note instead of restating it.
7. Added at the top of the file, naming both functions and stating that renaming either — or replacing it with an anonymous arrow — silently turns hook-order linting off for its body.
8. Rebased. The ADR was the only conflict, the expected both-appended case, resolved by keeping both. Two things surfaced: main has dropped the --- separators between entries, and the log's header mandates a lowercase adr- slug, so the entry is now adr-runaway-data-hook-guard.
9. Correct that 20 was the whole file. Small correction to the count: the guard describe had 10 and the other describe the other 10; it's 11 now with the containment test. Description updated.

Scope note. Added to PT-4421 as its own paragraph listing all five behavior changes, noting they belong in release notes even though the subtask title mentions none of them, with pointers to the ADR and PT-4502.

Not fixed, deliberately. A review pass flagged that handleLanguageChange can drop secondary languages (['es','pt','en'] → ['fr','en']) because safeInterfaceLanguage collapses to ['en'] on error. I verified the throttled path is already safe via the rejecting stub, so this needs a delivered provider error — pre-existing and unrelated to this change. Filing separately rather than widening a PR you've already flagged for scope.

> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2738#-:-P0QMFIv5s-7XcAD8889:b-njbz10)* status: 0 of 13 files reviewed, all discussions resolved.

jolierabideau (approved):
## Re-review — approving

Verified each item against the code, CI, and Jira rather than the reply alone. All three blockers are genuinely resolved.

**1. CI.** `Test` is green on head `d21c6d9` across macOS/Ubuntu/Windows, and `test.yml` does run `npm test`, `dotnet test`, `typecheck`, `lint`, and e2e smoke. Holding on this paid off — it surfaced the two unhandled `reading 'catch'` errors in `user-profile-popover.test.tsx` that this PR introduced. I ran the three affected files locally on head: 43/43 pass. (The `act()` warnings in that file are pre-existing on `main`, not yours.)

**2. DoD.** PT-4421's "Not verified, deliberately" section gives a reason for each gap. Ticket and PR agree now. One straggler: "No Power regression" is still in the checklist as satisfied without evidence and didn't move down with the other two — worth a sentence either way.

**3. The repeat cycle.** Consequences now states it plainly — trip → cooldown → re-arm → trip, the permanent spinner, the five-second Power/Simple flip, no message and no way out but closing the tab — and concedes it's the same unpaid bill charged against latching. PT-4502 is real and properly scoped.

The `useInterfaceMode` find is the best thing to come out of this round. Persisting a `PlatformError`-derived `'simple'` into the startup seed would have outlived the session and routed a Power user through the first-run gate on next launch — a transient becoming permanent, which is worse than the flicker. Good catch.

NITs 1–8 all verified done; 3 and 4 came back better than asked. NIT 9 is the one still open — the description still reads `20 tests; two mutation-verified` and `guard suite 20/20`, where head has 21 total / 11 in the guard describe. Cosmetic, fix it whenever.

One correction to the record, since it may travel to PT-4502: `persistedViewType` does exist in `platform-scripture-editor.web-view.tsx` (`:657`), and `viewType` is gated on `isInterfaceModeLoading` at `:709-711`. Doesn't change anything — the ADR's spinner + `isPowerMode` mechanism is accurate and is the one that matters.

Nice work on the turnaround. Approving.

jolierabideau (approved):
> @jolierabideau reviewed 13 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2738)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

src/renderer/hooks/papi-hooks/use-setting.hook.ts
```diff
@@ -71,6 +78,23 @@ export const useSetting = <SettingName extends SettingNames>(
     settingsService.reset(key);
   }, [key]);
 
-  return [setting, setSetting, resetSetting, isLoading];
+  // `useData` drops its setter when its runaway guard trips. This hook's returned type promises a
+  // callable setter, so call sites do not null-check it — substitute one that rejects rather than
+  // letting `setSetting is not a function` throw synchronously out of an event handler.
+  const safeSetSetting = useCallback(
```
tjcouch-sil: I could be reading this wrong, but it looks to me like you're making this inline "safeSetSetting" that is never `undefined` but throws when `setSetting` is `undefined`. But then you also add `undefined` as a possible type for `setSetting` in the hook type definition above.

It seems to me like these two changes are conflicting; one makes the returned setSetting value not able to be `undefined` while the other reports that it can be `undefined` (looks to me like it was always able to be `undefined` before, so this was most likely a bug in the type definition before).

Is there any particular reason why you're changing it to not be able to be `undefined`? Seems like it would be better to just make the type definition honestly report it can be `undefined` and move on; no need to make this inconsistency with the other `useData`-based hooks where the set can't be undefined.

src/renderer/hooks/papi-hooks/use-setting.hook.ts
```diff
@@ -43,9 +42,11 @@ export const useSetting = <SettingName extends SettingNames>(
...
 ): [
   setting: SettingTypes[SettingName] | PlatformError,
-  setSetting: (
-    newData: SettingTypes[SettingName],
-  ) => Promise<DataProviderUpdateInstructions<SettingDataTypes>>,
+  setSetting:
+    | ((
+        newData: SettingTypes[SettingName],
+      ) => Promise<DataProviderUpdateInstructions<SettingDataTypes>>)
+    | undefined,
```
tjcouch-sil, line 49: Note: because this is a breaking type change (even though this is really a fix to make the types actually reflect reality), you'll need to announce this in the `#platform-changes` channel in Discord when this PR merges. Thanks for doing this!

Note: you can resolve this PR yourself whenever you're ready.

src/renderer/hooks/papi-hooks/use-setting.hook.ts
katherinejensen00: No good reason — you're right, and I've changed it in 1661b53.

useSetting now declares setSetting as possibly undefined, matching useData and useProjectSetting, and the fabricated setter is gone.

One nuance on "it was always able to be undefined before": for this hook specifically it wasn't. useSetting hands settingsService — a sync proxy object, not a name — straight to useData, and createUseNetworkObjectHook short-circuits for a non-string source and returns it immediately, so dataProvider is never undefined there. That makes setData's dataProvider ? … : undefined always resolve to a function. The non-optional type was accurate right up until this PR's guard introduced a real undefined, and then I patched over the new hole instead of describing it. Your conclusion holds either way — widening is the right call, and the wrapper was me protecting call sites from a compile error they should have gotten.

Six call sites changed. The two user-facing ones — the profile popover (mode and language) and the first-run language step — branch on the missing setter and report the control as unavailable, which is the shape the popover already used for a dropped setTheme, so a click during a trip doesn't silently do nothing. The rest use setSetting?.(…), already the established form at useProjectSetting call sites. OtherSettingsControls.setSetting in setting.component.tsx also widened to match its ProjectSettingsControls sibling; that component already null-checked before calling, so only the prop type moved.

The ADR entry moved this from a rejected alternative into the decision itself, with "fabricate a rejecting setter" recorded as the rejected one.

src/renderer/hooks/papi-hooks/use-setting.hook.ts
katherinejensen00: Thanks, TJ! I appreciate the heads up about posting in platform-changes. I will make sure to do that when this merges.

katherinejensen00 (commented):
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2738)* status: 5 of 16 files reviewed, 2 unresolved discussions (waiting on jolierabideau and tjcouch-sil).

jolierabideau (approved):
> @jolierabideau reviewed 11 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2738)* status: all files reviewed, 2 unresolved discussions (waiting on tjcouch-sil).

jolierabideau (approved):
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2738)* status: all files reviewed, 2 unresolved discussions (waiting on tjcouch-sil).

katherinejensen00 (commented):
> @katherinejensen00 reviewed 1 file.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2738)* status: all files reviewed, 2 unresolved discussions (waiting on tjcouch-sil).

PR #2731, opened 2026-08-27
PT-4434: Make WebSocket disconnects diagnosable from the log

jolierabideau:
## Summary

A PAPI WebSocket disconnect was undiagnosable. When the renderer's socket died, the renderer — the process that actually breaks — **logged nothing at all**; every close line in a user's log came from main, with no close code, no peer identity, and nothing recording what the machine was doing. That is why [PT-1641](https://paratextstudio.atlassian.net/browse/PT-1641) sat unreproduced for 20 months.

This PR adds the instrumentation, and with it we reproduced and explained the failure.

**`pmset sleepnow` on macOS reproduces it in about one second of sleep.** The renderer's socket dies during the suspend transition, nothing reconnects, and the editor shows a blank pane with an endless spinner. Only the renderer's socket dies — the extension host and .NET data provider survive — because the renderer is the one peer using Chromium's native `WebSocket` (`web-socket.factory.ts`) while the others use Node `ws`. Chromium tears its WebSockets down across a suspend; `ws` does not. So this is not OS-specific or a network-stack mystery, which matches PT-1641 reproducing on Windows.

### Why review this

Part of NN-6 ("App never dies silently") under [PT-4419](https://paratextstudio.atlassian.net/browse/PT-4419), Sprint 89 Simple Quality non-negotiables.

### Scope: diagnosis only

No reconnect, no retry, no connection-lost UI — those are [PT-4435](https://paratextstudio.atlassian.net/browse/PT-4435). After this lands the app still breaks on resume; it just says why. A heartbeat is deliberately deferred (the transport is loopback-only, where a half-open connection dying with no FIN is unlikely); the ADR records what would reopen that decision.

**Please don't read this as "PT-1641 is fixed."** At least three distinct failure shapes are filed under it, and this explains one:

1. **Explained** — the renderer suspend disconnect above.
2. **Explained, fixed elsewhere** — a startup connect race leaving the extension host dead on a cold install (`%localizeKey%` everywhere, no providers). `RpcWebSocketListener.connect()` never awaits the server's `listening` event. Its ~9.98s failure is `AsyncVariable`'s 10s default timing out, not a crash. A fix is in flight on `fix/extension-host-connect-race`, so **this PR deliberately does not touch `rpc-websocket-listener.ts`.**
3. **Unexplained** — a report of the main process failing to finish a synchronous `forEach` for 17 minutes. A main-process stall, not a socket teardown. Nothing here accounts for it.

## Changes

[...]

PR #2739, opened 2026-08-31
PT-4422: Show a crash message instead of a blank web view

katherinejensen00:
https://github.com/user-attachments/assets/a3f4c64c-adbb-4c45-b751-9ba142fc42ac

Closes PT-4422 (NN1b of NN-1 "Editor never blanks out"). Pairs with PT-4421, which fixes a root cause; this is the safety net, and they are independent.

> Rebased onto latest main. Note that main renamed `web-view.service-host.ts` → `web-view.service-shard.ts` in the meantime; this branch targets the new name.

## Summary

**Before:** a React render error inside any web view tore down that iframe's React root. React unmounts the tree, so the pane went blank — with nothing in the log a user could report. There was no error boundary anywhere in `src/` or `extensions/`.

**After:** the pane shows "This panel stopped working", names itself, and offers Reload. The error, web view id, type and component stack go to the app log.

**Why at the web view service and not in the editor:** every React web view mounts through one `root.render` call, so one wrap covers all of them. A boundary in the editor extension would have fixed the editor and left Text Collection, Find, resource panels and third-party web views exactly as broken. Not hypothetical — during verification this caught a live `ResourceTextPanel` crash (PT-4046) that had been blanking silently.

## Where to look

Six files, but really only two ideas. Suggested order:

| # | File | What it does |
|---|------|--------------|
| 1 | `web-view.service-shard.ts` | **The change that matters.** Two lines: bridge the boundary into the iframe, and wrap the `root.render` element. |
| 2 | `web-view-error-boundary.component.tsx` | The boundary. ~40 lines: catch, log, render the fallback. |
| 3 | `web-view-crashed-view.component.tsx` | The fallback UI. Most of the review discussion is here. |
| 4 | `web-view-error-boundary-wiring.test.ts` | Guards #1, which no typecheck/lint/test otherwise reaches. |
| 5 | `global-this-web-view.model.ts` | Puts the boundary on the renderer global, exactly as `createRoot` already is. |
| 6 | `assets/localization/en.json` | Four new keys. |

Plus a Storybook story (`Advanced/WebViewCrashedView`) — the only way to see this view without forcing a crash.

## Three things worth your attention

1. **The mount-site edit is inside a template literal** (the generated iframe script), so no typecheck, lint or test reaches the code itself. It works only if three edits across two files agree. `web-view-error-boundary-wiring.test.ts` pins that they do — but only running the app proves the script executes, which I did manually.
2. **The boundary calls no hooks, on purpose.** A hook there would run *above* the catch; if it threw, the pane would blank exactly as with no boundary. Since a crash inside `useLocalizedStrings` is a known cause of blank web views, that hole would open exactly when the net is most needed. All localization lives in the fallback, below the catch — which is also why the fallback falls back to English literals rather than localize keys.
3. **The fallback is inline-styled, not Tailwind.** A web view's iframe gets the CSP, fonts, scrollbars and theme stylesheet — never the renderer's Tailwind build, so `tw:` classes render unstyled there. It mirrors shadcn's `Button`/`Empty` values by hand. Storybook renders it *with* Tailwind present, so the story checks layout and copy but not this.

## What it deliberately does not catch

Event handlers, async callbacks, unhandled rejections, and data providers that return a `PlatformError` instead of throwing (including PT-4421's render-rate guard). Render and commit phase only. Recorded in the ADR.

## Verification

`npm run typecheck`, ESLint and Prettier clean; 30 tests across the three new files (re-run after rebasing onto latest main). Manually verified in a dev build: forced crash shows the panel and logs the stack; Reload recovers; correct in light and dark and at minimum window width; other web view types unaffected; no Power regression.

Known-flaky suites (`scroll-group`, `web-view` service tests) time out under full-suite load; `web-view.service-shard.test.ts` — the one file this touches — passes 64/64 alone.

## Follow-ups filed, not included

- **PT-4498** — let users copy crash details (today they only reach the log).
- **PT-4499** — theme service startup race found during verification; unrelated to this change.

<details>
<summary><b>Full review record</b> — findings, author interview, quality checks (from <code>/review-paratext</code>)</summary>

# Code Review Summary

**Branch**: pt-4422-editor-error-boundary

**Base**: origin/main

**Date**: 2026-08-31

**Files changed**: 8 at review start (10 after in-review fixes added a story and a wiring test)

## Overview

PT-4422 (NN1b of NN-1 "Editor never blanks out") adds the first React error boundary in the
application. Every React web view reaches the screen through one `root.render` call in
`web-view.service-shard.ts`, executed inside the generated `srcdoc` script in the iframe, with nothing
above it — so any throw during render tore down that iframe's entire React root and React unmounted
the tree. Blank pane, nothing actionable in the log.

The boundary wraps that single choke point, so all web views are covered rather than the editor
alone. A crash now shows "This panel stopped working", names the pane, offers Reload, and logs the
web view id, type and component stack. This is not hypothetical coverage: during verification the
boundary caught a live `ResourceTextPanel` crash (`this.papi` undefined, matching PT-4046) that had
been blanking silently.

Verified manually in a dev build across all six planned checks — forced crash, logging, reload
recovery, light/dark theme, narrow width, all web view types, and Power mode.

## API Changes

- `lib/platform-bible-react/`: no changes
- `lib/platform-bible-utils/`: no changes
- `lib/papi-dts/papi.d.ts`: no changes — renderer web view globals (`createRoot`, `React`, and now
  `WebViewErrorBoundary`) are not emitted into that surface, so no `npm run build:types` drift
- `extensions/src/*/src/types/*.d.ts`: no changes
- Internal (not public API): `globalThis.WebViewErrorBoundary` added in
  `src/renderer/global-this-web-view.model.ts` and bridged into each iframe; new renderer components
  `WebViewErrorBoundary` and `WebViewCrashedView`; four new localize keys in
  `assets/localization/en.json`.

None.

- [x] **Crash message could render a raw localize key.** `WebViewDefinition.title` is
  `string | LocalizeKey` and several panels ship keys (Model Text, Scripture Text Grid, Enhanced
  Resources, Get Resources, Lexical Tools), so the message read
  `Something went wrong and "%webView_modelTextPanel_title%" could not be displayed.` — the exact
  raw-key outcome the English fallback exists to prevent, reintroduced through the interpolated
  value. Flagged independently by all three code passes. _(fixed during review: the title is now
  resolved through a memoized `useLocalizedStrings` call when `isLocalizeKey(title)`, falling back to
  the untitled message when it does not resolve; two tests added)_
- [x] **Focus was taken unconditionally on mount.** Several web views are visible at once, so a
  background pane crashing on a shared state change would yank the caret out of whatever the user was
  typing in — silently, since programmatic focus on a `tabIndex={-1}` container matches no
  `:focus-visible` ring. Conversely, in an inactive rc-dock tab (`display: none`) the focus call
  no-ops with no catch-up, so the stated accessibility guarantee failed exactly in the Simple-mode
  case. _(fixed during review: focus is claimed only when `containerRef.current.ownerDocument`
  already had it — `ownerDocument` rather than a bare `document`, which in this renderer-realm
  component would resolve to the renderer's document, not the pane's. The hidden-tab case is now a
  documented deliberate no-op per `.claude/rules/cross-view-sync-hidden-views.md`; test added
  proving a background crash leaves focus alone)_
- [x] **The mount site had zero test coverage.** The unit tests proved the boundary works but nothing
  proved it is mounted. The wrapper lives inside a template literal — invisible to typecheck, lint
  and tests — and works only if three edits across two files stay in agreement. Breaking any one
  would silently restore the blank-pane bug with every test green. _(fixed during review: added
  `web-view-error-boundary-wiring.test.ts`, 5 assertions across the class export, the global
  assignment, the iframe bridge, the wrapper and its props; verified falsifiable by deleting the
  bridge and the global assignment and confirming exactly the two covering tests fail. It is a
  consistency check, not a behavioral one — the test says so)_
- [x] **No Storybook story for a new user-facing view.** A UX reviewer cannot trigger a web view
  crash by hand, so without a story nobody but the author could see this view at all. _(fixed during
  review: added `web-view-crashed-view.stories.tsx` with four cases — named, localize-key title,
  untitled, long title. Also took the compliance pass's suggestion and made reload an injected prop,
  which removed the service-shard import from the component: it is now purely presentational, its test
  needs no service-shard mock, and the story works. The story's docblock states that Storybook renders
  it WITH Tailwind present, so it validates layout and copy but cannot reproduce the bare-iframe
  condition the inline styling exists for)_
- [x] **Four English strings duplicated with nothing keeping them in sync.** `ENGLISH_DEFAULTS`
  restates values that also live in `en.json`, and nothing asserted the keys even exist there. A copy
  edit to `en.json` would leave the last-resort fallback silently stale. _(fixed during review:
  considered de-duplicating by importing `en.json`, and rejected it — that file is read from disk by
  the extension host, not bundled into the renderer, so importing it would add a second 41KB copy to
  the renderer's load path for four strings, against a sprint with a startup-performance
  non-negotiable. Added the repo's existing guard pattern instead: every key must exist in `en.json`
  and match `ENGLISH_DEFAULTS` exactly)_

[Author response: all five accepted and fixed during the interview. On the duplication finding the
author asked whether de-duplicating would be better; the trade was investigated and the guard-test
route chosen for the reason recorded above.]

### Minor — Consider

- [x] **`componentStack` comment misplaced** inside the `getErrorMessage(...)` argument list, reading
  as a comment on a second argument that does not exist; stack also concatenated with no separator.
  _(fixed during review: comment moved above the log call, stack joined with an explicit newline)_
- [x] **Doc comment claimed the key array was "frozen"** when it is a plain mutable array, and
  `ENGLISH_DEFAULTS` was typed `Record<string, string>`, so a key added to the list but forgotten in
  the defaults would render as empty text with no compile error. _(fixed during review: wording
  corrected to "stable" — the array must stay `LocalizeKey[]` for the hook's signature — and the
  record is now keyed on the literal union, making a forgotten default a compile error)_
- [x] **`height: '100vh'` was fragile.** Nothing injected into a web view iframe resets the UA's
  default 8px body margin, so the panel overflowed by 16px and showed a scrollbar; in a short pane
  the centred content could not be scrolled to. _(fixed during review: `position: absolute; inset: 0`
  measures from the initial containing block and ignores the margin; `overflow: auto` with
  `justify-content: safe center` keeps content reachable)_
- [x] **`reloadWebView` resolves `undefined`** rather than rejecting when the web view is no longer in
  the layout, so the `.catch`-only handler left a Reload button that did nothing and logged nothing.
  _(fixed during review: both outcomes logged; test added for the resolves-`undefined` path)_
- [x] **Visual details diverged from the design system.** The button drew a visible border where
  shadcn's default variant uses `border-transparent`; radius, height and text sizing did not match;
  the layout did not mirror the `Empty` composition. _(fixed during review: button mirrors the default
  variant at default size — h-8, px-2.5, `rounded-lg`, text-sm/font-medium, transparent border — and
  the layout mirrors `Empty`/`EmptyHeader`/`EmptyTitle`. While there, added a scoped `<style>` for
  `:hover` and `:focus-visible`: inline styles cannot express pseudo-classes, and the button had no
  visible focus indicator at all, which matters because the container takes focus first and Tab moves
  straight to it)_
- [ ] ~~"panel" is not in the terminology or vocabulary guides — `terms.mdx` defines *Tab*, *Tab
  header*, *WebView*, not "panel" (one precedent exists in `en.json`). The message names a *tab*
  title while calling the thing a *panel*.~~ _(Author chose to skip; a UX wording call better made by
  the reviewer than unilaterally in review.)_
- [ ] **No way for the user to share the failure.** Details go only to the app log, and this is the
  case where Reload does not help — a deterministic crash re-crashes immediately. The app has an
  `ErrorDump` pattern with a copy button and existing strings, though it cannot be reused directly
  (Tailwind-classed, so unstyled inside the iframe). _(Filed as **PT-4498** rather than grown into
  this PR, with the reuse constraint recorded so it is not rediscovered.)_
- [ ] **PR description should carry the "React web views" qualifier** — the boundary is added in the
  `default:` (React) branch of the content-type switch; `HTML` and `URL` web views mount no React root
  and are unaffected. The ADR states this correctly. _(Reflected in the Overview above.)_

[Author response: author asked for 3, 4, 5 and the visual divergence to be fixed, skipped the "panel"
terminology question for the reviewer, and asked for the share-the-failure item to be filed as a
ticket. The remaining two Minor items are documentation/scope notes, not code changes.]

#### Shared Regions Modified

#### Extension Config Changes

## Positive Observations

- The hook-free constraint on the boundary is correct, deliberate, and documented at the class and in
  the ADR: a hook there would run above the catch, so a throw in it would blank the pane exactly as
  with no boundary.
- The English-literal fallback, with `localize()` treating "key resolved to itself" as unresolved,
  correctly handles the case where `useLocalizedStrings` is itself what broke — and is pinned by a
  test.
- The Lexical re-throw test pins a behavioral assumption spanning two packages this repo does not own
  (`platform-editor`'s `onError(error) { throw error; }` → `LexicalErrorBoundary` re-throwing out of
  `componentDidCatch`), with a comment explaining what breaks silently if an upgrade changes it. The
  chain was verified by running it, not by reading it.
- Both load-bearing design constraints were verified against the actual code rather than asserted:
  the iframe head injects only extension styles, fonts, scrollbars and the theme stylesheet (no
  Tailwind), `style-src` includes `'unsafe-inline'`, and every CSS variable referenced is genuinely
  emitted onto the body's theme class.
- Localization done correctly: four new keys in `en.json` only (translations arrive via the
  pipeline), correctly alphabetized, sentence case, standard `formatReplacementString` placeholder,
  curly quotes matching neighbouring strings, no `metadata.json` entry needed.
- The decision is recorded in `Architecture-Decisions.md` with alternatives and — unusually — an
  explicit list of what the boundary does *not* cover.
- Microcopy follows the written guidelines: sentence case, no ellipsis on the immediate-acting
  Reload, typographic quotes. RTL-safe by construction (no physical left/right, symmetric padding).
- No build-config, dependency or CI churn.

## Interview Notes

**Stated purpose.** Web views had no React error boundary anywhere, so any render-phase throw tore
down the iframe's React root and left a blank pane with nothing useful in the log. This adds a
boundary at the single shared web view mount point — covering every web view, not just the editor.

**Why the shared mount point rather than the editor extension** (author's explanation): every React
web view reaches the screen through the same `root.render` call in `web-view.service-shard.ts`,
executed inside the generated `srcdoc` script, with nothing above it. React's default for an
unhandled error is to unmount the tree, hence the blank pane. Putting the boundary in the editor
extension would have fixed the editor and left every other web view exactly as broken — and that is
not hypothetical: the boundary caught a live `ResourceTextPanel` crash during verification. One wrap
at the choke point covers all of them, including third-party web views. The cost is that the boundary
must reach `globalThis` to be visible inside the generated script, but `createRoot` already gets
there the same way, so it is an existing path rather than a new one.

**Why it is hook-free** (author's explanation): a React error boundary catches what happens below it,
never what happens in its own render. A hook in the boundary — `useLocalizedStrings`, say — would run
above the catch; if it threw, the pane would blank exactly as with no boundary, and the net would have
a hole precisely where it is supposed to be solid. This is not theoretical: a crash inside
`useLocalizedStrings` is one of the known causes of a blank web view, so a hook there would be most
likely to fail exactly when the boundary is most needed. Hence a plain class component, with all
localization in `WebViewCrashedView` — below the catch. The same reasoning drives the English-literal
fallback: the boundary cannot catch a throw in its own fallback (there is no boundary above it), so
the fallback stays small and does not depend on localization resolving.

The author demonstrated clear, accurate understanding of both design decisions, including the
non-obvious consequence that a boundary cannot catch errors in its own fallback. **No areas were
deferred to AI and no uncertainty was expressed.**

**Context a reviewer should know, surfaced during verification rather than analysis:**

- **PT-4046 is live in this build.** The boundary caught `ResourceTextPanel` throwing
  `Cannot read properties of undefined (reading 'webViews')` (`this.papi` undefined). Pre-existing and
  out of scope here, but now visible instead of silent — a reviewer running the branch may see the
  crash panel on the Bible Texts pane and should know it is not caused by this change.
- **A theme-service startup race**, filed as **PT-4499**. On one of four dev launches
  `themeServiceDataProvider.getCurrentTheme` timed out and theme switching was dead for the whole
  session, including the main process failing to set title-bar colours. The author reports hitting it
  somewhat frequently when the machine is taxed. Unrelated to this change (the main process fails
  independently of any web view), but it briefly looked like a regression during testing.

## In-Review Quality Check

Nine changes were made during the interview, so the checks were re-run afterwards.

- **`npm run typecheck`** — passes clean. Two `TS7053` errors were *introduced* by the review fixes
  (the stricter literal-union key typing broke two index expressions) and were fixed: the title
  resolution was restructured so `isLocalizeKey` narrows before indexing, and the `en.json` guard test
  iterates `Object.entries` so its keys stay typed without a type assertion.
- **ESLint (changed files)** — passes clean. Surfaced a genuine **rule conflict**:
  `@typescript-eslint/member-ordering` requires `render` before private instance methods, while
  `react/sort-comp` requires it after. Resolved by lifting the reload handler to a module-level
  function — it needs no instance state — which satisfies both rules without suppressing either. The
  reason is recorded in the function's doc comment.
- **Prettier** — passes clean.
- **Tests** — the three files for this change pass, 30 tests. Re-run after rebasing onto latest main
  (which renamed `web-view.service-host.ts` → `web-view.service-shard.ts`, so every reference in this
  change had to move with it). The full renderer suite then reports 2 files / 2 tests failing:
  `scroll-group.service` and `web-view.service-shard`, both `Test timed out in 5000ms`. These are the
  known flaky suites — the failing subset rotates between runs (it was 3 files / 5 tests before the
  rebase) — and `web-view.service-shard.test.ts`, the one file this change edits, passes **64/64 when
  run alone**.
- **Not run:** the full repo-wide `npm run lint` (~40 minutes here). ESLint was run against every
  changed file instead, which is equivalent coverage for this diff; CI will run the full sweep.

## Suggested Review Focus

- [ ] **The mount-site edit in `web-view.service-shard.ts`.** It lives in a template literal, so no
  typecheck, lint or unit test reaches the generated script itself. The new wiring test pins that the
  three pieces agree, but only running the app proves the script executes. Worth a careful read.
- [ ] **Is the shared mount point the right home?** The ADR records the alternative (per-extension)
  and why it was rejected. Reasonable place to push back if the team wants a narrower blast radius
  for a sprint non-negotiable.
- [ ] **The "panel" wording** — deliberately left for you. `terms.mdx` defines *Tab*/*WebView*, not
  "panel", and the message names a tab title while calling it a panel. Either change the noun or add
  "panel" to `vocabulary.mdx`.
- [ ] **The inline-styling decision and its Storybook consequence.** Storybook renders this component
  with Tailwind present, which is not the condition it is built for, so the story validates layout and
  copy but not the iframe styling. The only real check is running the app — done manually in light and
  dark and at minimum width, but not automated and not automatable in the current setup.
- [ ] **What the boundary deliberately does not catch** — event handlers, async callbacks, unhandled
  rejections, and a data provider returning a `PlatformError` rather than throwing (including NN1a's
  render-rate guard, which returns `RESOURCE_EXHAUSTED` rather than throwing). Confirm the team is
  happy that NN-1 is met by covering the render/commit class only.
- [ ] **Cross-package dependency:** editor coverage relies on `platform-editor` keeping
  `onError(error) { throw error; }`. A test pins it, but the team should know the dependency exists
  before the next editor upgrade.

</details>

AI-assisted — [session](https://claude.ai/code/session_016tUFqJgMimdHBc55oi9LsN)

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2739)


jolierabideau (changes requested):
**Overall:** this is unusually well-argued work and I'd merge it after two questions are settled. The placement decision (one boundary at the shared `root.render` rather than per-extension) is right, the ADR is genuinely useful, and I verified the three load-bearing claims — no Tailwind in the iframe, `'unsafe-inline'` in `style-src`, and reload producing a fresh root — all hold.

<details>
<summary>What I verified against <code>main</code>, rather than taking the PR description's word for it</summary>

- `style-src 'self' papi-extension: 'unsafe-inline'` — confirmed at `web-view.service-shard.ts:2696`, so the scoped `<style>` block is legal.
- The iframe head gets only CSP + font + scrollbar + theme stylesheet — **no Tailwind**. The inline-styling rationale holds.
- Every theme variable used (`background`, `foreground`, `muted-foreground`, `primary`, `primary-foreground`, `ring`, `radius`) exists in `theme.model.ts`, and the theme stylesheet is scoped to `.${theme.id}` on `body`, so they resolve inside the container.
- `reloadWebView` → `openOrReloadWebView` → new `srcNonce` → new `content` → `srcdoc` changes → iframe reloads (confirmed by the `previousContentRef` handling in `web-view.component.tsx:120`). So Reload really does give a fresh React root.
- `useLocalizedStrings` seeds every key with itself, so the `value !== key` unresolved test is correct.
- No import cycle: `web-view.service-shard` doesn't import `global-this-web-view.model`, and importing the shard from a renderer component is already the norm.
- Conventions match: `*.component.tsx`, colocated `*.test.tsx`, `Advanced/…` story title, ADR slug heading style, `%webView_error_*%` key namespace.

</details>

### Blockers

**1. The fallback still calls `useLocalizedStrings`, which is the exact hazard the design cites.**

`WebViewErrorBoundary` is deliberately hook-free on the reasoning that a hook there runs above the catch. But `WebViewCrashedView` calls `useLocalizedStrings` twice, and React routes a throw inside a fallback to the *next* boundary up — of which there is none. So the pane blanks exactly as before.

`ENGLISH_DEFAULTS` covers an **unresolved** string, not a **throwing** hook; those are different failure modes. Since the PR's own rationale names "a crash inside `useLocalizedStrings`" as a live cause of blank web views, the net has its hole precisely where it's claimed to be strongest.

Either: (a) render a static, hook-free English shell unconditionally and mount the localized text through a tiny inner boundary, so a localization throw degrades to English instead of blanking; or (b) if you judge the risk small, say so explicitly in the ADR under Consequences and file a follow-up — right now the ADR reads as though this case is covered.

**2. `hasError` never resets when the web view definition changes.**

`renderRoot` runs on every `onDidUpdateWebView` against the same boundary instance, so a pane that crashes once stays crashed through any definition update. Concretely: the editor crashes on a bad book in project A, the user switches that tab's project to B from the toolbar, and they still get "This panel stopped working" until they find the Reload button.

The test in `web-view-error-boundary.component.test.tsx` pins this as intended, so I may be missing the reasoning — but the standard shape here is resetting on changed keys (`projectId`, `content`). If the current behaviour is deliberate, could the *why* go in the ADR? A reasoned "no" closes this one for me.

### NITs

- `webViewId: '${webView.id}'` at the mount site interpolates through the template literal while its two sibling props read from `savedWebViewDefinition`. Same value, but `savedWebViewDefinition.id` is in scope and keeps all three props uniform.
- `<style>{BUTTON_STATE_CSS}</style>` sits inside the `role="alert"` container. Harmless (AT doesn't expose `<style>` text), but tidier as a sibling above the alert region.
- `WEB_VIEW_CRASHED_VIEW_STRING_KEYS` is an exported mutable `LocalizeKey[]`. `readonly` would stop a consumer mutating a module-level array the hook depends on being stable.
- Both new components carry a named export *and* a default export. Most of `src/renderer/components/` is named-only.
- Reload has no disabled/in-flight state, so repeated clicks stack async reloads. Low stakes given the reload replaces the iframe.
- `web-view-error-boundary-wiring.test.ts` is source-text scanning, which is brittle by nature — but its docblock already says so and argues the case, and I agree it's the right trade for a code path no typechecker reaches. Noting it only so the next reader knows it was a considered choice.
- The hand-mirrored shadcn `Button`/`Empty` values will drift when the design system moves. If more iframe-safe UI shows up later, these are worth lifting into one shared module rather than re-deriving.

### Reads well

- Generalizing beyond the editor is the right call and is the difference between fixing one pane and fixing all of them — catching a live `ResourceTextPanel` crash during verification makes the case better than any argument could.
- The localize-key-as-title case (`isLocalizeKey` resolution, with the untitled message as the fallback when it doesn't resolve) is a detail that would have shipped as a raw `%…%` on screen in most implementations of this.
- Conditioning focus on `ownerDocument.hasFocus()` — and the note about `document` being the *renderer's* document here — is exactly the cross-realm trap that usually gets missed, and the hidden-tab case is documented per the cross-view-sync rule.
- ADR recorded with the non-coverage (event handlers, async, `PlatformError` returns, throws in the fallback) stated plainly rather than glossed.

### Jira alignment (PT-4422)

Every DoD item is met — forced crash shows the message, logs it, no blank pane, Power checked. The sub-task says "around the editor" and this does all web views; that's scope *expansion*, justified in the ADR, and I'd accept it. Editor coverage does rest on a cross-package assumption (`platform-editor`'s `onError` re-throw), which the PR pins with a test — worth flagging to whoever owns that dependency.

### Heads-up

This appends a new ADR to the end of `Architecture-Decisions.md`, and so does #2731. Whichever merges second will conflict there — trivial to resolve, but expect it.

katherinejensen00 (commented):
reply for you to post — the two blockers needed answers, so this leans on those rather than recapping fixes:
Both blockers addressed, one by changing the code and one by writing down the reasoning you asked for.

1. useLocalizedStrings in the fallback — you're right, and it's fixed.

You're correct that the English defaults cover an unresolved string, not a throwing hook, and those are different failure modes. Took option (a).

The panel is now three pieces: CrashedViewShell (markup, styles, focus handling — reaches no service), LocalizedCrashedView (the hooks), and EnglishCrashedView (ENGLISH_DEFAULTS only). The localized one sits below a small inner boundary whose fallback is the English one, so a localization throw degrades to English instead of blanking. The shell renders identically either way, so the two paths only differ in text.

Two tests: a throwing useLocalizedStrings still shows the panel and a working Reload button, and a localize-key title on that path degrades to the untitled message rather than a raw %…%. The class doc and the ADR now both say the constraint that makes this work — everything above the inner boundary must stay service-free.

2. hasError never resetting — deliberate, and here's the why.

I checked the paths rather than reasoning from the general shape, and the scenario you describe doesn't reach the boundary.

updateWebViewDefinition is only reachable from a web view's own component — web-view.service-shard.ts:2448 binds it to the calling web view's id, and the web view service exposes no cross-web-view update to extensions (web-view.service-model.ts:73-78 says reload is the only way). So a crashed pane, whose component is unmounted, only ever receives updates from elsewhe are exactly four: scrollGroupScrRef (toolbar BCV and thenavigation command), state, isClosable from an e2e helper, and an empty bring-to-front. No external caller passes projectId at all.

That inverts the trade. Resetting on definition change would re-run the same failing render on every verse move in the scroll group — a crash-and-log
cycle per navigation, never a different result — and would stiuse switching a tab's project doesn't update the definition inplace. In Simple mode resolveOpenEditorDispatch returns replace-tab for a different projectId (platform-scripture-editor.utils.ts:694), which mints a new web view id; re-pointing a related panel calls reloadWebView. Both build a fresh iframe, so a fresh React root and a fresh boundary. Reload is the only recovery path that could work, which is why the panel leads with it.

That's now in the ADR under Consequences with the call sites, plus a test that pins the definition-change case with the same reasoning, so the next reader gets the argument and not just the assertion.

NITs

Fixed: webViewId now reads savedWebViewDefinition.id so all three props come from one definition (wiring test pins the new form); <style> moved out of the
role="alert" region; the keys array is readonly on export — thibes with no longer leaves the module.

Two I'd push back on:

Dual exports. I checked the directory: all seven non-test components in src/renderer/components/ have a named export and a default (notification-display, overlay-host, platform-bible-toolbar, reference-history-buttons, sync-status-button, tab-panel, web-view), and across src/renderer it's 45 files with a
default against 7 without. Named-only would be the deviation hyou'd rather move the convention, but it seemed wrong to do itin this PR alone.

Reload in-flight state. I don't think there's a version of thihere. Disabling on click and not re-enabling leaves a deadbutton in exactly the case where reload didn't work (reloadWebView resolving undefined for a web view no longer in the layout). Awaiting the promise instead re-enables the button at the same moment the iframe is torn down, so the disabled state is visible for a few milliseconds and then gone. Meanwhile stacked reloads are idempotent — each regenerates the srcdoc ad if you want it anyway.

Agreed on the other two: the wiring test is brittle by nature  hand-mirrored shadcn values are worth lifting into a sharedmodule once there's a second piece of iframe-safe UI. And noted on the ADR conflict with #2731.

Verification: npm run typecheck clean, ESLint and Prettier clean on the changed files, 33 tests across the three files (up from 30), and web-view.service-shard.test.ts 64/64.
1. useLocalizedStrings in the fallback — you're right, and it's fixed.

You're correct that the English defaults cover an unresolved string, not a throwing hook, and those are different failure modes. Took option (a).       
The panel is now three pieces: CrashedViewShell (markup, styles, focus handling — reaches no service), LocalizedCrashedView (the hooks), and EnglishCrashedView (ENGLISH_DEFAULTS only). The localized one sits below a small inner boundary whose fallback is the English one, so a localization throw degrades to English instead of blanking. The shell renders identically either way, so the two paths only differ in text.

Two tests: a throwing useLocalizedStrings still shows the panel and a working Reload button, and a localize-key title on that path degrades to the untitled message rather than a raw %…%. The class doc and the ADR now both say the constraint that makes this work — everything above the inner boundary must stay service-free.

2. hasError never resetting — deliberate, and here's the why.

I checked the paths rather than reasoning from the general shape, and the scenario you describe doesn't reach the boundary.

updateWebViewDefinition is only reachable from a web view's own component — web-view.service-shard.ts:2448 binds it to the calling web view's id, and thweb view service exposes no cross-web-view update to extension73-78 says reload is the only way). So a crashed pane, whosecomponent is unmounted, only ever receives updates from elsewhe are exactly four: scrollGroupScrRef (toolbar BCV and thenavigation command), state, isClosable from an e2e helper, and an empty bring-to-front. No external caller passes projectId at all.

That inverts the trade. Resetting on definition change would re-run the same failing render on every verse move in the scroll group — a crash-and-log
cycle per navigation, never a different result — and would stiuse switching a tab's project doesn't update the definition inplace. In Simple mode resolveOpenEditorDispatch returns replace-tab for a different projectId (platform-scripture-editor.utils.ts:694), which mints a new web view id; re-pointing a related panel calls reloadWebView. Both build a fresh iframe, so a fresh React root and a fresh boundary. Reload is the only recovery path that could work, which is why the panel leads with it.

That's now in the ADR under Consequences with the call sites, plus a test that pins the definition-change case with the same reasoning, so the next reader gets the argument and not just the assertion.

NITs

Fixed: webViewId now reads savedWebViewDefinition.id so all three props come from one definition (wiring test pins the new form); <style> moved out of the role="alert" region; the keys array is readonly on export — the mutable one the hook subscribes with no longer leaves the module.

Two I'd push back on:

Dual exports. I checked the directory: all seven non-test components in src/renderer/components/ have a named export and a default (notification-display, overlay-host, platform-bible-toolbar, reference-history-buttons, sync-status-button, tab-panel, web-view), and across src/renderer it's 45 files with a default against 7 without. Named-only would be the deviation here. Happy to change both if you'd rather move the convention, but it seemed wrong to do it in this PR alone.

Reload in-flight state. I don't think there's a version of this that's better than what's there. Disabling on click and not re-enabling leaves a dead button in exactly the case where reload didn't work (reloadWebView resolving undefined for a web view no longer in the layout). Awaiting the promise instead re-enables the button at the same moment the iframe is torn down, so the disabled state is visible for a few milliseconds and then gone. Meanwhile stacked reloads are idempotent — each regenerates the srcdoc and the last wins. Say the word if you want it anyway.

Agreed on the other two: the wiring test is brittle by nature and that's the trade, and the hand-mirrored shadcn values are worth lifting into a shared module once there's a second piece of iframe-safe UI. And noted on the ADR conflict with #2731.

Verification: npm run typecheck clean, ESLint and Prettier clean on the changed files, 33 tests across the three files (up from 30), and web-view.service-shard.test.ts 64/64.

> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2739#-:-P0O3jqc0KBJSgBwbz2r:bl0w5b4)* status: 0 of 10 files reviewed, all discussions resolved.

jolierabideau (approved):
Both blockers are closed and every NIT is either fixed or answered. Approving.

**Blocker 1** — the three-piece split is the right fix. `CrashedViewShell` reaches no service, the localized branch sits below its own boundary, and the two paths share the shell so the English path stays identical in layout. Good.

**Blocker 2** — I checked the paths rather than taking the reasoning on trust, and it holds. Every external `updateWebViewDefinitionSync` caller lands in one of your four categories (`auto-sync-edit-block-driver:101` is a fifth call site but it is a `state` update, so the argument covers it), and none passes `projectId`. My toolbar scenario doesn't reach the boundary: the in-pane project selectors in `checklist`, `checks-side-panel`, `find` and `manage-books` all go through the pane's own bound `updateWebViewDefinition`, and a crashed pane's component is unmounted. Thanks for putting the call sites in the ADR — that's what makes it checkable next time.

Both push-backs accepted. I verified the export one: all seven pre-existing non-test components in `src/renderer/components/` carry a named *and* a default export, so named-only would be the deviation here. Agreed on leaving the convention alone in this PR. The reload argument is sound too — a dead button in the case where reload didn't work is worse than stacked idempotent reloads.

Verification reproduces: 33 tests across the three files, `web-view.service-shard.test.ts` 64/64, `typecheck:core` clean, ESLint and Prettier clean on the changed files.

Two small things for a follow-up rather than this PR:

- **The "service-free above the inner boundary" constraint is documented twice but enforced nowhere.** Add a hook to `CrashedViewShell` and the guarantee is gone with every test still green — the same shape of silent regression the wiring test exists to catch. Since you already accept source-scanning for invariants no typechecker reaches, an equivalent guard there would be consistent.
- **`CrashedViewLocalizationBoundary.componentDidCatch` calls `logger.warn`** — a service, inside what is now the last line of defense. A throw there propagates to a boundary that doesn't exist and blanks the pane, which is structurally the hole this commit just closed. Very low probability, and the outer boundary does the same, but a one-line `try`/`catch` would close it for good.

jolierabideau (approved):
> @jolierabideau reviewed 10 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2739)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2741, opened 2026-08-31
PT-4431: fix language-selector search

katherinejensen00:
## Summary

Fixes the broken search in the first-run interface-language picker ([PT-4431](https://paratextstudio.atlassian.net/browse/PT-4431), part of NN-5 under PT-4419 — Simple quality, this sprint's epic).

Two problems, both hit as soon as you type:

1. **Enter chose nothing.** The picker renders `Command` with `shouldFilter={false}` and filters the list itself, so cmdk never learns the list narrowed. It kept its highlight on whatever row it last selected — typically one the query had just filtered out of the DOM — so pressing Enter after typing did nothing at all. The highlight is now controlled and **derived from the rendered rows**: the row the user moved to under the current query, else the current language, else the top match. Because it is derived rather than stored, it stays on a real row however the list changes — typing, clearing the search, or `value` / `languages` arriving after mount.
2. **Accented names were unreachable from an unaccented query.** Typing `francais` matched nothing, because the filter compared raw lowercased text. Search now strips the Combining Diacritical Marks block (U+0300–U+036F) on both sides, so `francais` finds `Français` and an accented query still works. Marks outside that block (Devanagari matras, Khmer vowel signs) are untouched.

## Changes

- `interface-language-picker.component.tsx`:
  - added `foldForSearch()`, with TSDoc naming its limits (it also merges Cyrillic й/и and ё/е; Ø, Ł, Đ, Æ, ß do not decompose);
  - extracted the filter into a module-level `filterEntries()`;
  - made cmdk's highlight controlled and derived (above);
  - a controlled cmdk value skips cmdk's own scroll and active-descendant bookkeeping, so the component scrolls the highlighted row into view when it moved it (not for arrow keys, which cmdk scrolls, or hover), and the search box is now an always-expanded `combobox` whose `aria-activedescendant` follows the highlight.
- `interface-language-picker.data.ts`: added `fr` / `Français` to the sample languages, so the accent behaviour is exercised by the stories and tests.
- Tests: no-results message and Enter on zero matches, query survives selecting a filtered match, Enter picks the first match, Enter without typing re-picks the current language, clearing the search returns the highlight to the current language, `value` and `languages` arriving after mount, a `value` naming no row, accent folding both ways, arrow keys through a filtered list, typing again after arrowing returns to the top match, the active descendant, and scrolling on typing but not on hover.

**Hover still moves the highlight**, as it did before this PR. `disablePointerSelection` was considered and not used: the vendored `CommandItem` has no hover style besides `data-selected`, so it would leave no hover feedback, and a controlled `Command` re-renders once per row crossed, not per pointer move (measured: 5 pointermoves over one row → 1 `onValueChange`, 2 commits).

### Note on the `dist` diff

This PR rebuilds `lib/platform-bible-react/dist`, and **that part of the diff is much larger than the change itself**. The workspace `build` script emits the bundle (`build:basic`) and only *then* runs `lint-fix`, which prettifies `src`, so a rebuild can also pick up JSDoc rewrapping from earlier formatting passes, plus the minified identifier renames that cascade from any bundle content change. CI's `npm run build` does not include `build:pbr`, so this bundle is never rebuilt or verified by the "Verify no files changed after build" step.

**Review the three files under `src/components/advanced/interface-language-picker/`; everything under `dist/` is generated.**

## AI Involvement

AI-assisted — [session 1](https://claude.ai/code/session_01JseVfg3YFfiaDtermuWSEn), [session 2](https://claude.ai/code/session_01JjccE9adEjgRbPLMwCZv6k). Claude diagnosed the cmdk highlight behaviour, wrote the fix and the tests, reworked the highlight after review, and rebuilt the bundle; I reviewed the diff and the reasoning before committing.

## Testing

- [x] `vitest run interface-language-picker.component.test.tsx` — 21 passed; the story tests (including a11y) pass too
- [x] Mutation-checked: the pre-review component and each piece of the derived highlight, scroll, and active-descendant logic fail at least one test when removed
- [x] `vitest run src/renderer/components/first-run` — 132 passed (the real consumer; renderer tests resolve `platform-bible-react` to the built `dist`, so this exercises the rebuilt bundle)
- [x] `eslint` + `prettier --check` on the changed files, and `tsc` in the workspace (via `build:basic`) — clean
- [ ] Packaged-build verification (per the ticket's definition of done)

## Risk Level

Low — one component, no API surface change. The highlight is now controlled where it was previously uncontrolled, so the thing worth a careful look is keyboard behaviour: arrow keys, Enter, Enter with no query typed, and clearing the query.

[PT-4431]: https://paratextstudio.atlassian.net/browse/PT-4431?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

https://claude.ai/code/session_01JjccE9adEjgRbPLMwCZv6k

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2741)


captaincrazybro:
## Code review — verified against a build of this branch

Reviewed the three files under `src/components/advanced/interface-language-picker/` per the PR description. Every finding below was **reproduced empirically**: I built a worktree of this branch and ran instrumented probe tests against both this component and the pre-PR baseline (`94783b5`), so each claim is backed by observed output rather than reading. Candidate findings that didn't survive that check are listed at the bottom.

**The fix works.** I confirmed the two headline bugs are genuinely repaired:

| | pre-PR | this PR |
|---|---|---|
| type `tok`, then Enter | no row highlighted, **Enter does nothing** | highlight on `tpi`, Enter picks it ✅ |
| query `francais` | **0 rows** | matches `Français` ✅ |

The 13 tests pass, and `dist/index.js` + `dist/index.cjs` really do contain the new controlled-`value` and folding logic, so the committed bundle isn't stale.

Three regressions and a few smaller things below.

---

### 🔴 1. Clearing the search box resets the highlight to English, so Enter switches the interface language back

`interface-language-picker.component.tsx` L151. This is the one I'd block on.

`onChange` runs `setHighlightedTag(filterEntries(entries, nextSearch)[0]?.tag ?? '')` on *every* keystroke — including the one that empties the field. `filterEntries(entries, '')` returns everything, and `sortLanguages` puts English first unconditionally, so emptying the box always drops the highlight onto English.

Realistic first-run sequence, measured on both sides:

```
                              pre-PR                this PR
click Français             →  highlight fr          highlight fr
type "tok"                 →  highlight null (bug)  highlight tpi   ✅ fixed
clear the box              →  highlight fr          highlight en    ⛔
press Enter                →  onChange('fr')        onChange('en')  ⛔
```

In `LanguageStep` that last line is `setInterfaceLanguage(['en', ...])` — the `tag === primaryLanguage` guard doesn't catch it — so the user types something, changes their mind, clears the box, hits Enter, and the interface language flips to English. It also *looks* wrong before they press anything: the check mark stays on Français while the highlight bar sits on English.

Pre-PR preserved the selection here because cmdk's own store still held it; this is a new regression, not a pre-existing one.

### 🔴 2. When `value` names no rendered row, nothing is highlighted at all and Enter is dead

L132. Making `Command` controlled suppresses cmdk's auto-highlight-first-item fallback (it only calls `selectFirstItem()` when its stored value is falsy).

```
<InterfaceLanguagePicker languages={SAMPLE_LANGUAGES} value="de" ... />

pre-PR:   en:true, es:false, fr:false, tpi:false, zh-hans:false  → Enter picks 'en'
this PR:  all false                                              → Enter does nothing
```

Not reachable from `LanguageStep` (it always merges `primaryLanguage` into `languages`), but this is an exported `@experimental` component, so any other consumer gets an unhighlighted list and an inert Enter until an arrow key is pressed.

### 🔴 3. The highlight moved by typing is never scrolled into view

L151. cmdk schedules `scrollIntoView` only from its internal `setState('value', …)`; the controlled-prop path doesn't. Measured with a 61-language list:

```
type a query  →  highlight moves en → x50,  scrollIntoView calls: 0
press ArrowDown → highlight moves x50 → x51, scrollIntoView calls: 1
```

New in kind, because pre-PR typing didn't move the highlight at all. With `CommandList` capped at `tw:max-h-72`, a user who has scrolled down and then types will have the highlighted top match off-screen — and Enter picks a language they can't see.

---

### One change fixes 1, 2, and the double-filter below

`highlightedTag` is derived state with a single write site, which is why it goes stale whenever something *other than* a keystroke changes the list. Deriving it instead fixes the whole class. I applied this to the branch and **all 13 existing tests pass unchanged**, with 1, 2 and the `value`-resolves-after-mount case below all fixed:

```diff
-  const [highlightedTag, setHighlightedTag] = useState(value);
+  // The row the user last moved the highlight onto (arrow keys / hover / click), if any.
+  const [preferredTag, setPreferredTag] = useState<string | undefined>();

   const visibleEntries = useMemo(() => filterEntries(entries, search), [entries, search]);

+  // cmdk's keyboard highlight, controlled so it always names a row that is actually rendered.
+  // Prefer the row the user moved to, then the current language (so Enter without typing re-picks
+  // it), then the top match.
+  const highlightedTag = useMemo(() => {
+    const isRendered = (tag: string | undefined) =>
+      !!tag && visibleEntries.some((entry) => entry.tag === tag);
+    if (isRendered(preferredTag)) return preferredTag ?? '';
+    if (isRendered(value)) return value;
+    return visibleEntries[0]?.tag ?? '';
+  }, [preferredTag, value, visibleEntries]);
```

with `onValueChange={setPreferredTag}` and the input handler back to `onChange={(e) => setSearch(e.currentTarget.value)}`. (3 needs its own fix — an explicit `scrollIntoView({ block: 'nearest' })` on the highlighted row.)

---

### 🟡 4. Accent folding misses letters whose diacritic is part of the code point

L46. NFD doesn't decompose Ø, Ł, Đ, Ð, Æ, Œ, so the stated goal fails for them. Measured on this branch:

```
"francais"   → Français  ✅
"foroyskt"   → 0 rows    (Føroyskt)
"lemkowski"  → 0 rows    (Łemkowski)
"dakovo"     → 0 rows    (Đakovo)
```

That's Faroese, Danish, Norwegian, Polish, Croatian. Either add a small explicit map for those letters, or match via `Intl.Collator(undefined, { sensitivity: 'base' })`.

### 🟡 5. `foldForSearch`'s doc comment describes the wrong range

L41–44 calls U+0300–U+036F "the Latin combining-accent block". It's the general Combining Diacritical Marks block, also used by Greek, Cyrillic and Vietnamese — so the justification given for the range is wrong even though the behaviour is fine. Verified: querying `ελληνικα` *without* the tonos **does** match `Ελληνικά`, and `tieng viet` matches `Tiếng Việt`. (Devanagari does survive untouched, as claimed.) Good behaviour, inaccurate comment.

### 🟡 6. `value` resolving after mount leaves the highlight stale

L103. `useSetting` returns the `['en']` default first and the stored value a tick later (per its own TSDoc: *"the initial value to return while first awaiting the setting value"*), so the picker mounts with `value='en'` and can then receive `value='fr'`. Verified: `aria-current` moves to Français but `aria-selected` stays on English, and a bare Enter fires `onChange('en')`.

Same outcome pre-PR, so **not a regression** — but it defeats the invariant the new comment on this line claims ("Starts on the current language so Enter without typing re-picks it"). The derived-highlight change above fixes it.

### 🟡 7. Every keystroke filters the whole list twice

L151 runs `filterEntries` in the handler, then the render's `useMemo` runs the identical scan with identical arguments. Free to remove via the change above.

### 🟡 8. Test comment lists the wrong matches

`interface-language-picker.component.test.tsx` L217 says `'n'` matches four languages. It matches **five** — 中文（简体） also matches, via `uiNames.en` = "Chinese (Simplified)". Verified: `['English', 'Español', 'Français', 'Tok Pisin', '中文（简体）']`. The `{ArrowDown}{Enter}` → `'es'` assertion holds only because `sortLanguages` happens to put 中文 last.

### 🟡 9. Neither behaviour the new comments claim is actually tested

No test asserts Enter-without-typing re-picks the current language (L102's claim), and none asserts Enter does nothing on a no-match query (L151's `?? ''`). Those are exactly the untested paths behind findings 1, 2 and 6 — a regression to either passes the whole suite today.

---

### Pre-existing — flagging only, not from this PR

Verified real but present before this branch and untouched by it: the query is never trimmed, so a single space in the search box filters to only names *containing* a space (Tok Pisin, 中文); and Home/End pressed in the search box move the cmdk highlight instead of the caret (measured: highlight `es`→`en`, caret unmoved), as do Ctrl+N/P/J/K via cmdk's default vim bindings. The search `<input>` also carries no combobox ARIA and `aria-activedescendant` stays `null` after typing — worth noting only because *this PR* is what makes typing move the highlight, so there's now a state change with nothing announcing it to a screen reader.

### Checked and found clean

- Mixed-case tags (`zh-Hans`) highlight correctly — cmdk normalizes both sides, so there's no case-mismatch problem.
- Single-language path (no search box) highlights correctly.
- Enter on a no-match query correctly does nothing (the `?? ''` fallback works).
- The `dist` rebuild is faithful: `experimental.js`/`.cjs` changes are pure minified-identifier churn, and the picker's new logic is present in the bundles that carry it.

<sub>🤖 AI-assisted review (Claude Code). Findings reproduced against a build of this branch; the suggested patch was applied and the existing suite re-run before posting.</sub>

katherinejensen00 (commented):
# PR #2741 re-review — PT-4431: fix language-selector search

Head reviewed: `a3ac4754` (unchanged since the first review; the local
`pt-4431-language-selector-search` branch at `021e294bd` is a rebase — the three source files are
byte-identical).

Verified against cmdk@1.1.1's actual `dist/index.mjs` source, by mutation testing of the suite, and
by running the HEAD and BASE components side by side under vitest/jsdom. Findings that change or add
to the standing review are marked **NEW** / **CORRECTION**.

B1 and B4 are confirmed **regressions** against `94783b54a77`, measured rather than reasoned: on the
same input sequence BASE calls `onChange('fr')` where HEAD calls `onChange('en')` (B1), and BASE
highlights English with a working Enter where HEAD highlights nothing and swallows the key (B4).

---

## Blockers

### B1. Clearing the search box drops the highlight onto English; Enter then switches the language

`interface-language-picker.component.tsx:151`

`onChange` runs `setHighlightedTag(filterEntries(entries, nextSearch)[0]?.tag ?? '')` on every
keystroke, including the one that empties the field. `filterEntries(entries, '')` returns everything
(L62) and `sortLanguages` pins `en` first unconditionally (L70), so emptying the box always lands
the highlight on English — while the check mark (L164, `tag === value`) stays on the real current
language. Enter then fires `onChange('en')`, and the consumer's guard
(`language.component.tsx:70`, `if (tag === primaryLanguage) return`) does not catch it.

Sequence: click Français → type "tok" → clear the box → Enter → interface language flips to English.

### B2. The highlight is seeded once and never re-syncs, so in production it freezes on `en`

`interface-language-picker.component.tsx:103` — `useState(value)` reads `value` on the first render
only.

The real consumer's `value` is not correct at mount: `language.component.tsx:29-36` uses
`useSetting('platform.interfaceLanguage', ['en'])`, whose TSDoc
(`src/renderer/hooks/papi-hooks/use-setting.hook.ts:20`) documents the default as *"the initial
value to return while first awaiting the setting value"*. A French user mounts with `value='en'`,
`highlightedTag` freezes at `'en'`, `value` then becomes `'fr'` — and Enter without typing fires
`onChange('en')`.

That is the exact behaviour the comment on L102 claims this state exists to provide ("Starts on the
current language so Enter without typing re-picks it"). Not a regression — uncontrolled cmdk also
auto-selected English pre-PR — but the PR asserts it is fixed and it is not.

**This is more reachable than a first-mount race.** `first-run-shell.component.tsx:170` swaps
`StepComponent`, so `LanguageStep` *remounts* on every step change: pick French → Next → Back and
the check mark is on Français while the highlight bar sits on English. Verified by rerendering with
a changed `value` — the check moves, the highlight does not.

### B3. **NEW** — the `languages` prop also changes after mount, and the highlight does not follow

`interface-language-picker.component.tsx:120` vs `:151`

`highlightedTag` has exactly one write site: the input's `onChange`. `language.component.tsx:57-62`
memoizes `languages` on `[setupLanguages, availableLanguages, primaryLanguage]`, all of which change
as PAPI data arrives and again after a selection. `entries` recomputes; the highlight does not.

cmdk partially rescues this — on item *unregistration* it runs `selectFirstItem()` when the removed
item was the DOM-selected one — but it recovers to the **first row overall**, i.e. English again,
which is B1's failure by another route. When the list *grows*, no recovery runs at all (registration
gates `selectFirstItem` on `n.current.value ||`, and a non-empty value suppresses it).

### B4. When `value` names no rendered row, nothing is highlighted and Enter is dead

`interface-language-picker.component.tsx:132`

Confirmed against cmdk source. `selectFirstItem` is called from item registration only as
`v(3, () => { …; n.current.value || W(); … })` — a non-empty stored value that matches nothing
suppresses the fallback. And Enter is
`case "Enter": { e.preventDefault(); let i = M(); if (i) {…} }` where `M()` is
`querySelector('[cmdk-item=""][aria-selected="true"]')` — no match means nothing is dispatched, and
`preventDefault()` has already run, so the key is swallowed entirely (it will not reach an enclosing
form or a parent handler either).

Not reachable from `LanguageStep`, but this is an exported `@experimental` component.

---

## Should-fix

### S1. The highlight moved by typing is never scrolled into view

`interface-language-picker.component.tsx:151`, with `CommandList` capped at `tw:max-h-72`.

Confirmed in cmdk source: the controlled-`value` path is a bare `useLayoutEffect` —
`k(() => { if (R !== void 0) { let e = R.trim(); n.current.value = e, E.emit() } }, [R])` — which
writes the state ref directly and never schedules `ne()` (scrollSelectedIntoView). `ne` is only
scheduled on mount and from inside `setState`'s `"value"` branch. So arrow keys scroll; typing does
not. Measured with a 31-language list: three ArrowDowns → 3 `scrollIntoView` calls; then typing
`nano` moves the highlight from row 10 to row 0 with **0** calls. The highlighted row — and the row
Enter will choose — sits off-screen above the viewport.

**Latent today**: only a handful of locales ship, so the list never overflows `max-h-72`. But the
component's TSDoc (`:79`) promises it "scales to hundreds of languages" and it is a public export.

### S2. **NEW** — hovering the list now costs a full React re-render per row, and steals the highlight

`interface-language-picker.component.tsx:133`

`CommandItem`'s `onPointerMove` is `S()` = `m.setState("value", b.current, true)`. On a controlled
`Command` that branch calls `onValueChange` and returns — so every pointer move across a row now
fires `setHighlightedTag` and re-renders the whole picker. Pre-PR the Command was uncontrolled and
that store write stayed internal to cmdk. `disablePointerSelection` on the `Command` root (cmdk
default `false`) turns it off, and is the honest expression of the design if hover is not meant to
drive the keyboard highlight.

### S3. **NEW** — the controlled path leaves `aria-activedescendant` stale

Because the prop-sync effect above bypasses `setState`, it never runs
`v(7, () => { n.current.selectedItemId = M()?.id; E.emit() })`. Combined with the plain `<input>`
carrying no combobox ARIA at all (`role`, `aria-expanded`, `aria-controls`, `aria-activedescendant`),
a screen-reader user gets no announcement when typing moves the highlight. cmdk *does* put
`aria-activedescendant` on `CommandList`, but that element never holds focus, so it is ignored — and
it goes stale on the typing path anyway (measured: `aria-activedescendant` naming row 10 while the
highlight was on row 0). The missing ARIA is pre-existing; what is new is that this PR makes typing
move the highlight, so a blind user now has a silently-moving Enter target.

Note the obvious fix does not work: you cannot wire `aria-activedescendant` to your own row ids,
because cmdk overwrites `id` after the prop spread. The practical fix is an `aria-live="polite"`
sr-only region announcing the match count and the highlighted autonym.

### S4. The `foldForSearch` doc comment is wrong about the range — and the error matters

`interface-language-picker.component.tsx:41-45` calls U+0300–U+036F "the Latin combining-accent
block". It is **Combining Diacritical Marks**, script-neutral. Measured:

```
Русский    -> русскии      (й -> и)
Ελληνικά   -> ελληνικα
हिन्दी      -> हिन्दी        (untouched, as claimed)
```

For Cyrillic these are **letter merges, not accent folds** — й and и are distinct letters of the
Russian alphabet, as are ё and е. Greek tonos-stripping is desirable but happens by accident. The
Devanagari/Khmer half of the claim is correct.

### S5. Every keystroke filters the whole list twice

`:151` runs `filterEntries` in the handler; `:120`'s `useMemo` runs the identical scan on the render
`setSearch` triggers.

### One change fixes B1, B2, B3 and S5

Derive the highlight from `visibleEntries` instead of storing it — keep state only for the user's
explicit override (arrow keys / hover / click), and fall back to `value`, then to
`visibleEntries[0]`. Two caveats worth deciding deliberately: a persisted override should probably
reset when `search` changes (otherwise an arrowed-to row that still matches outranks the new top
match), and this does **not** fix S1 — that needs an explicit `scrollIntoView({ block: 'nearest' })`
on the highlighted row.

---

## Tests (mutation-verified)

Each headline fix does have one genuine regression test: deleting the highlight-move fails "Enter
chooses the first match after typing"; reverting `foldForSearch` fails the accent tests. The gaps:

### T1. The arrow-key test does not pin this PR's fix

`interface-language-picker.component.test.tsx:207-221`. Deleting the entire highlight-move fix leaves
this test **green** (verified by mutation). Query `'n'` matches **all five** sample languages —
`zh-hans` too, via `uiNames.en` = "Chinese (Simplified)" — so the list is not filtered at all,
contradicting the test's own name, and `en` is simultaneously the initial `value`, the first match
and the first row.

Fix: query `'an'` (matches only `es, fr`) and expect `'fr'`. Verified green on HEAD and **red** under
the mutation. Correct the comment on `:217` at the same time.

### T2. "Enter without typing re-picks the current language" is claimed but untested

Changing `useState(value)` to `useState('')` leaves all 13 tests green. This is the untested path
behind B2.

### T3. Enter on a zero-match query is untested

The zero-match test (`:133-145`) passes `onChange={() => {}}` (`:138`), so it cannot assert the
outcome of a keypress. Give it a `vi.fn()`, type `'zzzz{Enter}'`, assert not called.

### T4. An existing comment is now false

`:102-120` says "All languages must remain visible" and then enumerates four rows — `Français` is
not among them. Add it, or assert on the option count.

---

## Nits

### N1. **CORRECTION to the standing review** — the Ø/Ł/Đ folding gap affects no shipped language

NFD does not decompose Ø, Ł, Đ, Ð, Æ, Œ, ı, ß, ə, so `foroyskt`/`lemkowski`/`dakovo` find nothing.
But every autonym and `uiName` in `src/extension-host/data/language-details.data.ts` (all 10 — en,
es, fr, de, km, zh, zh-hans, zh-hant, hi, ar) was checked and **none** contains a non-decomposable
diacritic. This is a forward risk only, not a live gap — the first review's framing implied current
breakage. Worth a line in `foldForSearch`'s TSDoc naming the limitation, given the component claims
to "scale to hundreds of languages" and the data file's own header says it should come from
SLDR/langtags.

### N2. Missing blank lines between tests

`interface-language-picker.component.test.tsx:132-133` and `:206-207` — Prettier will not fix these.

---

## Checked and clean

- **Mixed-case tags (`zh-Hans`) are safe** — but not for the reason the first review gave. cmdk does
  **no** normalization beyond `.trim()` on both sides (`R.trim()` on the prop, `R.trim()` on the item
  value); the comparison is exact and case-sensitive (`P(v => v.value && v.value === b.current)`). It
  works because both sides carry the same string, not because cmdk normalizes.
- The plain-`<input>` comment (L91-96) is accurate: keydown is a React synthetic handler on the
  Command root, so events from any focused descendant reach it. cmdk's focus-grab branch checks for
  `cmdk-input`/`cmdk-root` attributes, which the plain input does not carry, so no refocus occurs.
- No dead code, no unused imports, no hardcoded user-facing strings. Comments are forward-facing (no
  ticket/PR/review-history references) per
  `.claude/rules/code-quality/forward-facing-comments.md`.
- `filterEntries`'s extraction is earned (two call sites). `LanguageEntry` is not API surface.
- Sample-data change (`fr`) breaks no consumer: only this test file and
  `interface-language-picker.stories.tsx` use `SAMPLE_LANGUAGES`, and the one story using it has no
  `play` function.
- No over-mocking: only the jsdom gaps cmdk needs (`ResizeObserver`, `scrollIntoView`).
- **The headline fix is real**, confirmed against BASE: hovering a row and then typing left the base
  component with `aria-selected` on *no* row, so Enter did nothing; HEAD correctly lands on the top
  match. Accent folding is correct for combining-mark languages, `entries` memoization is preserved,
  and the transient zero-match `''` highlight self-heals through cmdk's item-mount rescue
  (backspacing back to a match re-highlights the first row).

## Pre-existing, flagged only

`Home`/`End`/`Ctrl+N`/`Ctrl+P`/`Ctrl+J`/`Ctrl+K` are all `preventDefault()`-ed by cmdk's root
keydown handler (`vimBindings` defaults to `true`), so they move the list highlight instead of the
caret in the search box. Untouched by this PR.

> @katherinejensen00 partially reviewed 15 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2741#-:-P0i8mCI1DLIRqmiK9OM:bb0p8hd)* status: 0 of 15 files reviewed, all discussions resolved (waiting on katherinejensen00).

captaincrazybro:
## Follow-up: one correction to my review, plus rebase status

### Correction — mixed-case tags are NOT safe, and I told you they were

My earlier comment listed this under *Checked and found clean*:

> Mixed-case tags (`zh-Hans`) highlight correctly — cmdk normalizes both sides, so there's no case-mismatch problem.

**Both the reason and the conclusion are wrong.** cmdk does no normalization beyond `.trim()`; the comparison is exact and case-sensitive. My probe didn't catch it because I passed matching case on both sides. Measured on the current head:

```
value="zh-hans", key="zh-Hans"  →  en:false, zh-Hans:false  →  Enter does nothing
value="  es  ",  key="es"       →  es:true                  (trim does apply)
```

So a case mismatch between `value` and the `languages` key is a **live instance of B4** (`value` names no rendered row → nothing highlighted, Enter swallowed by cmdk's `preventDefault()`), not a cleared item.

@katherinejensen00's *Checked and clean* entry has the mechanism exactly right — cmdk trims both sides and compares exactly, so it works "because both sides carry the same string, not because cmdk normalizes." That's correct, and it's precisely why the *safe* conclusion only holds while both sides do carry the same string. Worth moving out of the clean list and folding into B4, since two review comments currently read as clearing it.

Not reachable from `LanguageStep` (it keys `languages` and `value` off the same `primaryLanguage`), so this changes B4's severity, not its reachability — but it is a second route into B4 for any other consumer of this `@experimental` export, and it argues for the `?? visibleEntries[0]?.tag` fallback rather than treating B4 as hypothetical.

### The rebase resolved nothing — all findings are still open

Flagging this because a force-push reads as activity on the timeline. `a3ac4754` → `9e18dfaa` is a **pure rebase** (base `94783b54` → `49c26ed0`, #2760). The three reviewed files are byte-identical — by blob hash, not by eye:

```
interface-language-picker.component.tsx        IDENTICAL
interface-language-picker.component.test.tsx   IDENTICAL
interface-language-picker.data.ts              IDENTICAL
```

Commit message unchanged (still describes only the original fix). So every finding in both reviews stands, and the line numbers in both still resolve correctly.

No deferral is recorded anywhere either — no `TODO`/`PT-` marker in the component, nothing in the PR body, no reply to either review, and PT-4431 has 0 comments and no follow-up ticket. If the intent is to ship B4 / S1 / S2 / S3 as known limitations, that decision needs to exist somewhere before merge.

The rebase itself is clean, for what it's worth: the committed bundle carries both this fix and main's newer `platform-bible-react` src (`useRetryablePromise`), so it didn't go stale against the new base.

### Cross-checks on the re-review

I re-ran the checkable claims in @katherinejensen00's review against `9e18dfaa` and they all hold:

- **T1** — deleting the entire highlight-move fix leaves the arrow-key test **green**; only "Enter chooses the first match after typing" goes red. The suggested replacement works: `'an'` matches only `es, fr`, and ArrowDown+Enter → `'fr'`.
- **T2** — `useState(value)` → `useState('')` leaves all 13 green.
- **N1** — the correction to my review is right. Every autonym and `uiName` in `src/extension-host/data/language-details.data.ts` uses NFD-decomposable diacritics (á, é, í, ó, ñ, ç, ö); no Ø/Ł/Đ/ß anywhere. My original framing implied live breakage — it's forward risk only, and my finding should be read at N1's severity, not mine.
- **S4** — the Cyrillic letter-merge point is real and is the sharper half of that finding: `й` → `и` and `ё` → `е` both fold true, so `Русский` → `русскии`. Symmetric, so search still works, but it merges distinct letters of the alphabet rather than folding accents.
- **B2** — source-confirmed the reachability argument: `first-run-shell.component.tsx:170` is `const StepComponent = stepComponents[step]`, so a step change swaps the component type and remounts `LanguageStep`, re-running `useSetting`'s default-then-resolve race each time.

<sub>🤖 AI-assisted review (Claude Code). Reproduced against a build of `9e18dfaa`; mutations run against the committed suite.</sub>

katherinejensen00:
Rebased onto main and pushed a follow-up commit. Most findings are plain code changes and the commit covers them; these are the ones that needed a decision:

Hover still moves the highlight (S2). I did not add disablePointerSelection. The vendored CommandItem has no hover style of its own, only data-selected, so with that flag a mouse user gets no feedback on the row under the pointer at all. The cost is also smaller than one render per pointer move: on a controlled Command, 5 pointermoves over one row gave 1 onValueChange and 2 commits, so it is one picker render per row crossed. Hover taking the Enter target is also what the picker did before this PR. If we would rather hover not drive the highlight, that is the flag plus a hover style on the row.

Screen-reader announcement (S3). I used the combobox pattern on the search box instead of an aria-live region: role="combobox", aria-expanded, aria-controls pointing at cmdk's listbox, and aria-activedescendant following the highlight. cmdk overwrites the list and row ids, so they are read back from the DOM after render. That needs no new localized strings; an aria-live match-count announcement would, so I left it out.

Accent folding (S4, N1). I kept the script-neutral fold and corrected the TSDoc instead. It now names the Cyrillic й/и and ё/е merges (they only broaden matches, since the query is folded the same way) and the letters that do not decompose (Ø, Ł, Đ, Æ, ß). No shipped language contains the latter.

Left as pre-existing: Home/End and Ctrl+N/P/J/K being captured by cmdk in the search box, and the untrimmed query. Tags still match case-sensitively; a value that matches no row now falls back to highlighting the top row, so Enter is never dead, but the check mark still needs the exact tag.

captaincrazybro (approved):
Approving. Re-verified against `10fa1194`: all four blockers fixed, S1/S3/S4/S5 fixed, S2 declined with reasoning I agree with, and the three deferrals are now stated. The derived highlight behaves correctly on every path I could construct.

Measured on this head: clearing the query returns the highlight to the current language (Enter → `fr`, was `en`); `value` and `languages` arriving after mount are both followed; a `value` naming no listed row falls back to the top row instead of swallowing Enter; typing scrolls the highlight into view (2 calls, was 0) while hover correctly scrolls nothing; and `aria-activedescendant` matches the highlighted row on both the typing and arrow paths, and is removed on a zero-match query.

I also checked the one thing that worried me in the new code. When nothing matches, `highlightedTag` becomes `''`, which re-arms cmdk's own `selectFirstItem` fallback — the same fallback that produced the original "lands on English" bug by another route. It doesn't bite: the derived value is recomputed on the same render and the controlled prop reasserts, so `value` wins over cmdk's guess. Verified with a non-English current language across three variants (full backspace, partial backspace to a match, and a matching language arriving mid-query) — the highlight returns to the current language every time, never to English.

Tests are much stronger: 13 → 21, and all seven mutations I tried were killed (dropping the `value` fallback → 3 red, the top-match fallback → 5 red, the `preferredTag` reset → 1, `scrollIntoView` → 1, `aria-activedescendant` → 1, the `preferredTag` preference → 2, scrolling unconditionally → 1). `dist` is a genuine build of the new source — the minified layout effect carries both the `setAttribute`/`removeAttribute` pair and the `highlightedTag !== preferredTag` scroll guard. eslint clean, no React warnings.

Three non-blocking notes:

**1. The arrow-key test doesn't pin what its name says.** Removing the `isVisible(preferredTag)` branch leaves `arrow keys move the highlight through the filtered list` green — cmdk's internal store carries the arrow independently of the controlled `value`, so it passes either way. The mutation is caught by `typing again moves the highlight back to the top match` and the active-descendant test, so coverage is fine; it's the name that's misleading.

**2. A stray hover outranks a later `value` change.** `preferredTag` is cleared only when the query changes, so: hover Tok Pisin → `value` arrives as `fr` → highlight stays on `tpi`, check mark on `fr`, bare Enter picks `tpi`. That is the documented precedence working as designed, and it follows from keeping hover-selection. Noting it only because a pointer *crossing* the list on its way elsewhere is accidental in a way arrowing isn't — if S2 ever gets revisited, this is the second reason.

**3. Unrelated to this PR:** `src/renderer/components/first-run` currently runs 90 tests with two files failing to collect (`scope-selector.utils.ts:4`, `Cannot read properties of undefined (reading 'map')`). I nearly filed it as a regression — it's inherited from main, and the PR's parent `f89e626e` fails identically in the same environment. Side effect is that the PR body's "126 passed" line is now stale.

<sub>🤖 AI-assisted review (Claude Code). Verified against a build of `10fa1194`; mutations run against the committed suite.</sub>

katherinejensen00 (commented):
Thanks for the re-verification. Rebased onto main (the bundle is rebuilt, so it now includes the PT-4551 picker changes too).

1. Renamed the arrow-key test to "ArrowDown then Enter chooses the next row of the filtered list", since the Enter outcome is what it actually checks. The highlight itself is pinned by the active-descendant test.
2. Agreed. Hover outranking a later value is the cost of keeping hover selection. If we revisit disablePointerSelection, it's the second reason to do it, alongside ad
3. The first-run suite runs 132/132 here with no collection failures, so the scope-selector failure looks environment-specific. Updated the count
in the PR body.

> @katherinejensen00 partially reviewed 1 file and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2741#-:-P2F71tw-TJ2nBQYm1ne:b-1etqh9)* status: 0 of 15 files reviewed, all discussions resolved (waiting on katherinejensen00).

PR #2740, opened 2026-08-31
fix(docs): unbreak Publish documentation — unresolvable TypeDoc link

tjcouch-sil:
## Summary

The **Publish documentation** workflow has been failing on `main` since [`bc1b47f`](https://github.com/paranext/paranext-core/commit/bc1b47fc7a6) (PT-4187, Standard View first-pass port). The docs site has not been redeployed since.

### Why review this

Not part of the current epic. `main` currently cannot publish docs at all — the workflow dies on its first step, so the GitHub Pages site (typedoc for all three libraries plus both Storybooks) is stale as of that merge. One-line fix, worth unblocking now.

### Root cause

`bc1b47f` expanded the TSDoc on `Z_INDEX_ABOVE_DOCK` and added a `{@link Z_INDEX_TOOLTIP}` reference to it. `lib/platform-bible-react/typedoc.json` has a single entry point, `src/index.ts` — and `Z_INDEX_TOOLTIP` is exported from `src/experimental.ts`, not `src/index.ts`. So the link has no target in the doc set:

```
[warning] Failed to resolve link to "Z_INDEX_TOOLTIP" in comment for Z_INDEX_ABOVE_DOCK
[warning] Found 0 errors and 1 warnings
npm error code 4
```

`typedoc.json` sets both `validation.invalidLink: true` and `treatWarningsAsErrors: true`, so that single warning exits non-zero and takes the whole job with it. The sibling `{@link Z_INDEX_MODAL}` and `{@link Z_INDEX_ABOVE_POPOVER}` references in the same comment resolve fine — those two *are* exported from `index.ts`.

## Changes

- Reference `Z_INDEX_TOOLTIP` as inline code rather than `{@link}` in the `Z_INDEX_ABOVE_DOCK` comment.

The prose is unchanged and still names the constant. The alternative — promoting `Z_INDEX_TOOLTIP` from the experimental surface into `index.ts` — would be a consumer-visible API change, which is well beyond a CI fix; the experimental surface is documented as deliberately separate with no stability guarantee.

## Testing

Ran every step of the workflow that had not yet executed on `main` (it failed at the *first* doc build, so nothing downstream had been exercised on this commit):

- [x] `lib/platform-bible-react` → `npm run build:docs` — **exit 0** (was exit 4)
- [x] `lib/platform-bible-utils` → typedoc — **exit 0**
- [x] `lib/papi-dts` → `npm run build:docs` — **exit 0**
[...]

PR #2565, opened 2026-07-15
PT-4187: Standard View — first-pass port to Paratext 10

tjcouch-sil:
# Code Review Summary

**Branch**: `standard-view` (paranext-core, [PR #2565](https://github.com/paranext/paranext-core/pull/2565)) · `standard-view` (scripture-editors, [PR #545](https://github.com/eten-tech-foundation/scripture-editors/pull/545)) — both draft, both based on `main`

**Base**: `origin/main` (both repos)

**Date**: 2026-08-20

**Review model**: Claude Opus 5 (1M context)

**Files changed**: 508 — **137 in paranext-core** (+22.7k/−4.0k, 146 commits) and **371 in scripture-editors** (+130.6k/−2.4k, 642 commits; ~34k of the insertions are design/plan/spec docs under `docs/superpowers/`)

**Review mode**: Autonomous static analysis (eight parallel passes: api / style / compliance / ux, once per repo), followed by an author-directed fix pass. Purpose was inferred from the commit history; the prior PR description was ~5 weeks stale.

## Overview

PT-4187: porting Paratext 9's **Standard View** scripture-editing experience to Paratext 10, across two repos. `scripture-editors` supplies the Lexical editor: the `standard` view mode, an editable marker-edit engine (Tier-1 fast path / Tier-2 preserve-or-refuse rebuild behind a debounced "settle"), display runs and marker/attribute glyphs, marker menus/palettes, StyleInfo-driven CSS generation, USFM↔USJ whitespace and attribute-literal fidelity, collab delta handling, and caret-guard plugins. `paranext-core` wires it into the `platform-scripture-editor` extension: the `'standard'` view type, an auto-show/hide footnotes pane, the `\` / Enter / Ctrl+T / Ctrl+Shift+T marker palettes, a passive command-palette overlay mode in the renderer, a `StyleInfo` PDP plus four project settings backed by real C#, the vendored marker stylesheet, and the editor↔PDP save/echo pipeline.

The previous `/review-paratext` run was 2026-07-15; ~105 core and ~490 scripture-editors commits landed since, so essentially all of it was re-reviewed.

Static analysis found **1 Critical, 24 Important, ~35 Minor**. After the author's direction, **31 findings were fixed across both repos** in this review; the rest are open with a recorded reason. One finding was discovered during verification rather than by analysis: **the branch's own test suite was already red** (see Critical).

### Second pass — the author's hand-verification round (2026-08-20)

[...]

PR #2743, opened 2026-08-31
PT-4021: Keep the footnote popover's insertion point usable

katherinejensen00:
# Code Review Summary

**Branch**: pt-4021-ip-bug-fix

**Base**: origin/main (`790d8d6e2e8`)

**Date**: 2026-08-31

**Files changed**: 16 (8 hand-written, 8 generated `dist/`) at review start; 19 after in-review fixes

## Overview

The footnote popover's insertion point. Current `main` (PT-4187, #2565) already places the caret in
the note's text when a footnote opens, so the originally-reported symptom is gone and that part of
this branch was dropped as redundant. What remained were three paths that still left the user unable
to type, each measured against `main` before anything was changed:

1. **Changing the note type or the caller** replaces the note in the editor, which discards the
   selection — the caret landed on the note itself, outside the character runs, so the next
   keystroke joined no text. Both paths now go through one `replaceNoteKeepingCaret`.
2. **Focus** then went to the dropdown's own trigger button, so even a correct caret was
   unreachable. Both dropdowns now hand focus back to the note text via `onCloseAutoFocus` — but
   only after an actual change (see Findings).
3. **The `\` markers menu** came up empty from the state the popover opens in while still swallowing
   the keystroke, because a caret in a character run reports that run's marker (`ft`, `fr`, `fq`,
   `xt`) and none of those define children. It now falls back to the enclosing note's markers.

## API Changes

- **No public API surface changed.** `lib/platform-bible-react/src/index.ts` and `src/experimental.ts`
  are untouched; no `papi.d.ts`, `lib/platform-bible-utils/`, or extension `types/*.d.ts` files are
  in the diff. `papi.d.ts` regenerates unchanged.
- Internal to `lib/platform-bible-react` (not exported from the package entry points):
  - `footnote-editor.utils.ts`: `generateInlineMarkerMenuListItems()` gained an **optional** 5th
    parameter `noteMarker?: string`, plus a matching `@param`.
  - `footnote-type-dropdown.component.tsx` and `footnote-caller-dropdown.component.tsx`: each gained
    a **required** `focusNoteText: () => void` prop.
  - `footnote-editor.test-harness.tsx` (test-only, outside the package exports): added
    `CARET_IN_NOTE_TEXT`, `caretAncestry()`, `caretIsInsideNote()`, `installPopoverJsdomStubs()`,
    and `visibleView`.
- `src/stories/keyboard-shortcuts.data.ts`: added the `footnote-close-markers-menu` entry.
- `dist/index.d.ts` / `dist/experimental.d.ts`: doc-comment reflow only — including one
  `{@link Z_INDEX_TOOLTIP}` → backticks change that originates from merge-base commit #2740, which
  did not rebuild `dist`. No declared type changed.

The identically-named `generateInlineMarkerMenuListItems` in `platform-scripture-editor` is a
**different** function with a different signature and is untouched.

None.

- [x] **The `onCloseAutoFocus` override treated a dismissal identically to a choice**, contradicting
      `guidelines/dismissal-patterns.mdx` and WCAG 2.4.3 _(fixed during review: the restore is now
      claimed only when a note type was actually chosen; a dismissal falls through to Radix's own
      trigger-restore. Measured before the fix: pressing Escape left `document.activeElement` on
      `<body>` — neither the trigger nor the editor. Pinned by a new test that fails with
      `expected <body> to be <button>` if the gate is removed.)_
- [x] **The caller dropdown had the same defect but not the same fix**, so the "or caller" half of
      the stated purpose was unimplemented _(fixed during review: measured that a caller change left
      the caret correct but focus on the caller trigger. It now takes the same `focusNoteText` +
      gated `onCloseAutoFocus` treatment, pinned by two new tests in
      `footnote-editor.caller-dropdown.test.tsx` — one for the change, one for the dismissal. The
      change test fails with `expected <button> to be <div>` without the hand-off.)_
- [ ] ~~`setSelection(selectionBeforeChange)` in `replaceNoteKeepingCaret` is dead code~~ _(Three of
      the four passes independently reached this conclusion from reading alone; it is wrong.
      Deleting that single line makes both note-type tests fail with
      `expected [ 'note', 'para', 'root' ] to deeply equal [ Array(5) ]` — the selection IS live when
      read, because it is read during the item's `onCheckedChange`, before Radix's close nulls it.
      What was genuinely wrong was the JSDoc, which oversold it: the restore buys the right **node**,
      not the right **offset**. The comment now says that, and says the test fails if the line goes,
      so the next reader does not repeat the deduction.)_

Author response: the two focus findings were real and are fixed — the dismissal one was a
regression this branch introduced, and it is the finding the review most earned its keep on. The
third was investigated and rejected on measured evidence rather than judgement, and the misleading
comment that caused three reviewers to reach the same wrong conclusion was rewritten.

### Minor — Consider

- [x] **No dismissal test for the note-type dropdown**, though the comment claimed the behavior
      "for a dismissal as much as for a choice" _(fixed during review: that claim turned out to be
      the bug; both the claim and the behavior changed, and a test now pins the dismissal.)_
- [x] **No caret/focus test for the caller path** _(fixed during review: two tests added.)_
- [x] **`focusNoteText`'s doc said "Always prefer this to a bare `focus`"** while two bare-`focus`
      sites remain in the same file _(fixed during review: narrowed to the case it actually governs —
      where focus is handed back from a control that may have outlived the selection.)_
- [x] **jsdom scaffolding duplicated across five suites** (`NoopResizeObserver`, `scrollIntoView`,
      `Range.getBoundingClientRect`, `visibleView`) _(fixed during review: replaced by
      `installPopoverJsdomStubs()` and a `visibleView` fixture in the shared harness; three suites
      now call it, including the pre-existing `caller-dropdown` suite.)_
- [x] **`lexicalCaretInsideNote` in `enter-guard.test.tsx` duplicated the new `caretAncestry`**
      _(fixed during review: added `caretIsInsideNote` to the harness and collapsed the local copy
      onto it. It is a separate helper rather than a call to `caretAncestry` on purpose — the
      enter-guard caller needs `false` when there is no selection, where `caretAncestry` throws.)_
- [x] **The utils fallback looked up `usfmMarkers[parentMarker]` twice** and re-looked-up a
      known-childless marker _(fixed during review: hoisted local + explicit `noteMarker` branch.
      Behavior identical; both branches already covered by the unit test.)_
- [x] **Nothing tied the new `noteType` argument to the utils behavior end-to-end** _(fixed during
      review: a component test switches the note to a cross-reference and asserts the `\` menu offers
      `xo`, which belongs to `\x` alone. Fails with `Unable to find an element with the text: xo` if
      `noteType` is pinned to a fixed `'f'`.)_
- [ ] **Typing `\` with an empty marker list is a silent double no-op** — `event.preventDefault()`
      runs before `showInlineMarkersMenu()` bails, so the key is swallowed and nothing appears.
      Pre-existing; this branch's fallback makes it much rarer but does not remove it (it still
      applies when `parentMarker` is `p` or absent). Deliberately left: it changes what a keystroke
      does, which is a wider blast radius than the rest of this branch.

Author response: everything except the swallowed backslash was fixed. That one was left because it
is pre-existing behavior and changing what a keystroke does is a bigger decision than this branch
should make on its own.

#### Shared Regions Modified

#### Extension Config Changes

None. No files under `extensions/` changed; no `package.json`, `tsconfig*`, `.eslintrc*`, webpack, or
CI workflow changes; no new dependencies.

## Positive Observations

- Every claim about `main`'s behavior was **measured before any code changed**: probes recorded the
  caret at `note < para < root` after a type switch, `document.activeElement` on the dropdown button,
  and the `\` keystroke as `defaultPrevented: true` with zero menu items. The same discipline caught
  that the review's own "dead code" finding was wrong, and that the dismissal path was worse than the
  reviewing agent predicted.
- Every fix is ablation-verified with a distinct failure message, so no test passes vacuously.
- Both new dropdown behaviors are gated on a real change rather than fired unconditionally, which is
  what keeps the accessibility default intact on the path where nothing happened.
- `restoreSelectionIfLost` and `installPopoverJsdomStubs` both remove duplication rather than adding
  a copy beside it — the latter also cleans up a pre-existing duplication in three older suites.
- The two redundant parts of the original branch (open-time caret placement, per-keystroke
  `.editor-input` lookup) were dropped once `main` was found to cover them, rather than carried
  forward as conflict-prone dead weight.
- No comment or test name carries a ticket ID, PR number, or change narration.
- New suites mount the real `Editorial` through the shared harness and explain in their headers why a
  mock cannot observe what they assert.
- The keyboard-shortcuts catalog rule was honored: `footnote-close-markers-menu` mirrors the existing
  `scripture-close-markers-menu` in id shape, category, per-OS formatting (`⎋` / `Esc`), and its
  `locations` path is accurate.

## Interview Notes

**Stated purpose**: make the footnote popover's insertion point usable — keep the caret in the note's
text (and the editor focused) after a note-type or caller change, and make the `\` markers menu offer
the enclosing note's markers instead of coming up empty.

- **Fix both focus paths** rather than only the one the ticket named. The caller dropdown had the
  identical defect; fixing one and not the other would have left two adjacent controls in the same
  toolbar row behaving differently.
- **Keep `setSelection` and reword its docs** rather than removing it on three reviewers'
  recommendation, because the ablation evidence contradicted them.
- **Leave the swallowed `\`** as the one unfixed minor finding, on blast-radius grounds.
- Earlier in the branch's life: **drop the exact-offset caret restoration**. Routing main's
  `[noteType, showMarkersMenu]` effect through the new helper broke **7 of main's own tests**, which
  assert `selectNote` is not called on those paths — that bare `focus()` is deliberate. The test now
  documents the limitation rather than asserting an offset.

No areas were deferred to AI or left unexplained by the author.

## In-Review Quality Check

| Check | Result |
| --- | --- |
| `tsc` — `platform-bible-react` | clean (two errors introduced by the in-review edits were found and fixed) |
| `tsc` — root | clean |
| `eslint` (changed files) | clean |
| `prettier --check` (changed files) | clean |
| `vitest` — footnote-editor suites | 77/77 pass (73 before the review) |
| `vitest` — full `platform-bible-react` unit project | 768/768 pass, 68 files |
| `dist` rebuild | rebuilt against the final source |

**Not runnable in this worktree**: the `storybook` vitest project (browser/a11y tests over stories)
fails resolving `@papi/frontend` from extension sources — a module produced by a full `npm run build`
this checkout has never had. CI installs Playwright and runs a full build first. No story files were
changed.

**Environment repairs (not repository defects).** This worktree had never run the full `postinstall`,
which produced two false alarms worth recording so the next person does not chase them:

- `link-dev-packages` had never run, so the tree held the registry
  `@eten-tech-foundation/platform-editor` `0.8.15` while `main` needs the `platform-yalc` dev build
  (`0.8.16`). That alone caused 18 `tsc` errors and 20 failing suites **in main's own untouched
  files**.
- `patch-package` had been skipped, leaving `ast-types` unpatched.

## Suggested Review Focus

- [ ] **No live-app verification.** Every focus claim here is jsdom. Radix's restore behavior is
      well-established and has six precedents in this repo, but nobody has opened a real footnote,
      switched its type, and typed. This is the cheapest gap to close before the meeting.
- [ ] **Is "caret at the end of the note's text" acceptable after a note-type change?** It is not
      returned to where the user was. Fixing that means reworking main's `[noteType]` re-focus
      effect, which 7 of main's tests pin — a deliberately deferred decision, not an oversight.
- [ ] **The dismissal-vs-choice split in both dropdowns.** After a change, focus goes to the note
      text; after a dismissal, back to the trigger. Confirm that is the behavior the team wants,
      since it is a deliberate divergence from "Radix does this automatically".
- [ ] **`TODO(PT-4199)` overlaps the `\` menu fix.** The menu is still built from the static
      `usfmMarkers` table; PT-4199 will replace that with the project's stylesheet (as the main
      editor already does via `getMarkerMenuItems`). The fallback is correct until then and
      disappears with it.
- [ ] **The swallowed `\`** (minor, left open) — worth a decision on whether the key should fall
      through when there is nothing to offer.
- [ ] **No footnote e2e coverage exists** anywhere in `e2e-tests/`. Worth a follow-up ticket rather
      than a blocker.

AI-assisted — [session](https://claude.ai/code/session_01KZ4e2EmmHwEHv4Ri49qZLx)

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2743)


jolierabideau (changes requested):
Nice work on this — the commit-vs-dismissal gating is the harder half of the focus behavior and it's both reasoned at the call site and tested on both branches with real `document.activeElement` assertions. The `\`-menu fallback removes a genuine dead end, and the util test covers it as a matrix with the discriminating "`xo` belongs to `\x` alone" case rather than just a happy path. Mounting the real Lexical editor instead of a mock is the right call for caret assertions, and the harness extraction removes more duplication than it adds (including some pre-existing duplication in the older suites).

Requesting changes on one blocker and three code items. None of them are about the footnote fixes being wrong — they're good fixes.

## Blocker: the ticket linkage

PT-4021 is *"Editor — Drafting when `\li2` is present in the book"*, and both symptoms it reports are in the **main scripture editor**: Simple mode places the cursor before the `\li2` glyph so typed text lands over the marker, and Power mode loses the IP with the marker tooltip appearing. This branch touches only `footnote-editor/` — no `\li2`, no gutter/glyph positioning, no paragraph-marker selection.

Those symptoms are already owned elsewhere, under Sprint 90 (PT-4530):

- **PT-4542** (WI-12) — the Simple-mode collision, already adjudicated *not reproduced* and reduced to a regression test.
- **PT-4540** (WI-13) — marker-as-selection-target, whose `ParaMarkerPrefixCursorGuardPlugin` clears a marker-glyph selection within ~900 ms. That's the "IP is lost" symptom.

PT-4021 also already has #2629 merged against it for the `\li`/`\lim` gutter-indent CSS half, and it currently sits in `👀 Review`. So if this merges as PT-4021, a still-open bug gets marked fixed and the work double-tracks against PT-4540/PT-4542.

Could we retitle this against its own ticket for the footnote-popover caret/focus work? The code isn't blocked by this — just the linkage.

## The caller dropdown's dismissal guarantee isn't quite true

`handleDropdownOpenChange(false)` (footnote-caller-dropdown.component.tsx:121-141) is the single close path for Escape, outside-click, and a row click alike, and it commits whenever the pending value differs. There's no `onEscapeKeyDown` anywhere in the folder. So Escape after selecting **Custom** and typing a character does commit, sets `committedCaller.current = true`, and now also relocates focus into the note text — Escape applies the change and moves the user out of the toolbar.

The new comment says "After a dismissal nothing changed, so Radix's own restore to this trigger is both what the user expects and what WCAG 2.4.3 asks for" — but the real split is *value differs* vs *value same*, not commit vs dismissal. This also brushes `Standard-View-Invariants.md` §4, which says the Custom row "leaves its own check as the confirming gesture — the same commit Enter performs"; Escape is neither.

The commit-on-close design is pre-existing — what this PR adds is the focus relocation on top of it. Suggested: treat `onEscapeKeyDown` as an explicit cancel (revert pending state, leave the flag false), which makes the comment's claim true, and add the Escape-after-typing-custom case to the caller test.

## The two dropdowns don't share one rule

`chooseNoteType` (footnote-type-dropdown.component.tsx:85-88) sets `choseNoteType.current = true` *before* calling `handleNoteTypeChange`, which has no same-value guard (footnote-editor.component.tsx:632) and only replaces the note when `getNoteOps(0)?.at(0)` is a note embed. `DropdownMenuCheckboxItem`'s `onCheckedChange` fires when the already-checked item is clicked, and the `f`/`fe` items are only disabled when `noteType === 'x' && !isTypeSwitchable` — so re-picking the current type replaces the note and yanks focus out of the trigger, which is exactly what the dismissal branch was written to avoid.

The caller dropdown gets this right by setting its flag only inside the real-change branch, and the commit message says "only when something changed". An early `if (newNoteType === noteType) return;`, or having `handleNoteTypeChange` report whether it applied, would make the two consistent.

## `Standard-View-Invariants.md` §4 should record the new rules

§4 ("The footnote editor popover") is the doc CLAUDE.md points you at before touching this code, and this PR adds two rules of exactly that kind:

1. A dropdown inside the popover claims Radix's `onCloseAutoFocus` **only after a committed change** and hands focus to the note text; a dismissal keeps Radix's own restore.
2. The inline markers menu is built from the **enclosing note's** marker when the caret's own character-run marker defines no children.

Both are silently breakable by a future refactor, and CLAUDE.md asks that settled conventions be promoted into the relevant standard.

## Non-blocking, for discussion

- **`footnote-editor.component.tsx:405-410`** — `useEffect(() => { if (!showMarkersMenu) editorRef.current?.focus(); }, [noteType, showMarkersMenu])` is a bare `focus()` keyed on `noteType`, which is precisely the call `focusNoteText`'s new TSDoc says is "right only where the selection is known to still be live". It's safe today only because `handleNoteTypeChange` runs `setNoteType` then synchronously calls `replaceNoteKeepingCaret` before React re-renders — which is why `replaceNoteKeepingCaret`'s doc has to spend a paragraph on the ordering. I saw your note that rerouting it breaks 7 of main's tests, so I'm raising it as a discussion item, not a fix request.
- **No live-app verification.** Every caret and focus claim here is jsdom. Radix's restore behavior is well-established and has precedent in the repo, but nobody has opened a real footnote, switched its type, and typed. Cheapest gap to close.
- **The catalog entry is slightly under-scoped.** `footnote-close-markers-menu`'s handler (footnote-editor.component.tsx:1039) is only reached when `options.view?.markerMode !== 'editable'`. In editable marker mode the popover's markers surface is the marker palette, whose Escape dismissal lives in `marker-palette-keydown.util.ts:389`, and the only entry covering that (`scripture-close-markers-menu`) has context `Scripture editor web view` — so Escape in the footnote popover's palette is documented nowhere. Either add that file to this entry's `locations` (mirroring `footnote-markers-menu`, which lists both) or qualify the context by marker mode.
- **`replaceNoteKeepingCaret`'s `setSelection(selectionBeforeChange)`** could go out of range if the replacement changes content length (footnote → cross-reference); `focusNoteText()` only repairs a *null* selection, not an out-of-range one. The f→x case is covered by a test, so this may be fine — worth a glance from someone who knows the `platform-editor` `SelectionRange` contract.
- **The still-swallowed `\`** — `preventDefault()` at :1037 runs before `showInlineMarkersMenu()` bails on an empty list. Agreed this is pre-existing and out of scope; noting it so it's on the record.

A handful of smaller items (missing `@param closeMarkersMenu`/`@param localizedStrings` on the util's TSDoc, a no-op `expect(getByText('xo')).toBeDefined()`, the unexplained per-test `}, 20000)` timeouts, the six repeated inline `setTimeout(resolve, 50)` settles that could live in the harness, the half-applied `installPopoverJsdomStubs()` sweep, the leftover local `editableView` in the caller test, and `visibleView` sitting in the harness rather than the dependency-light fixtures module) I'm happy to leave to your judgment or drop as inline comments if useful.

One caveat on my end: I couldn't execute the suites (no `node_modules` in my review worktree), so my read of the test coverage is from reading them, not from a green run — CI is the authority there.

---

*AI-assisted review.*

katherinejensen00 (commented):
Thanks for the careful review, Jolie.

**Ticket linkage:** I'm keeping this under PT-4021 because I am using this work item to do a few footnote ip placement hot fixes.

**Caller dropdown:** Escape now cancels. Clicking outside the menu still applies a pending change, the same as picking a row. I left that as-is because you only flagged Escape; say if you think outside-click should cancel too.

**`replaceNoteKeepingCaret`'s `setSelection`:** it can't go out of range here. Switching between footnote and cross-reference only renames the character styles (`ft` → `xt`, `fr` → `xo`, and so on) without touching the text, and every caller is a single character, so the note's content length never changes.

**The bare `focus()` effect keyed on `noteType`:** leaving it for now. Moving it onto `focusNoteText` breaks 7 of main's existing tests, so it deserves its own change.

**Live-app check:** done in the running app. In a footnote popover I:

- switched the type to cross-reference and typed;
- re-picked the current type (focus stayed on the type button);
- chose Custom, typed `%`, and pressed Escape (caller unchanged, focus back on the caller button);
- picked Hidden and typed;
- opened and closed the `\` markers menu outside Standard view.

After each change, typing landed in the note text.

> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2743#-:-P1I_3AoDNNbO7v6U-3A:bdcdqkm)* status: 0 of 23 files reviewed, all discussions resolved.

jolierabideau (approved):
Thanks, Katherine. Everything from my last review is addressed.

- **Ticket linkage:** I accept keeping this under PT-4021. Could you update the ticket so it covers this work? Right now its description is only the `\li2` main-editor symptoms. Adding the footnote-popover insertion-point fixes (note-type/caller change, dropdown focus hand-off, the `\` markers-menu fallback) keeps the history accurate when it closes.
- **Outside-click on the caller dropdown:** keeping it as a commit is fine. It's the existing behavior and the Invariants text now matches it.
- **Escape cancel, re-picking the current type, the §4 Invariants rules, the catalog `locations`, and the smaller test/TSDoc items:** all look good, and the new tests pin the right cases.

Before merging: the branch currently conflicts with `main` in `lib/platform-bible-react/dist/*`. Once you've merged `main` and rebuilt `dist`, CI can run on the new head. The latest commit hasn't had a CI run because of the conflict.

Optional: a follow-up ticket (or a `TODO` naming one) for moving the bare `focus()` effect at `footnote-editor.component.tsx:406-410` onto `focusNoteText` would stop that one from getting lost.

---

*AI-assisted review.*

PR #545, opened 2023-10-11
Add web view state service and corresponding React hook

lyonsil:
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/545)

PR #2745, opened 2026-08-31
Build the editor from source and move it to paranext/scripture-editors

tjcouch-sil:
Two related changes: how this repo consumes the scripture editor, and where that editor lives.

Part of [PT-4500](https://paratextstudio.atlassian.net/browse/PT-4500).

## The problem

Every consumer here declared `@eten-tech-foundation/platform-editor` and `scripture-utilities` as **registry ranges** (`~0.8.15` / `~0.1.6`) and then yalc-linked a locally built copy over the installed package.

That registry entry's real job was never the code — yalc discards it seconds later. It was the **dependency closure**: 8 of the editor's 13 runtime dependencies (`@floating-ui/dom`, five `@lexical/*`, `quill-delta`, plus the `yjs` peer) reached `node_modules` *only* as transitive dependencies of the published package.

So the editor's dependency set was pinned to whatever ETEN last published:

| Change in the editor | What happened |
| --- | --- |
| **Add** a dependency | Never installed. The build imports it → module not found |
| **Bump** a dependency | npm installs the version the *published* package declared → silent, subtle mismatch |

Since we now maintain that repo ourselves, this is the thing standing between us and actually being able to change it. The source had already drifted to 0.8.16 against a published 0.8.15, and a fresh worktree resolving the registry copy failed 36 tests on a symbol the published build lacks.

## The change

`preinstall` builds each package named in `dev-packages.json` and copies exactly the files `npm pack` would publish into `dev-packages/staging/<folder>`. All six `package.json` files here reference those folders with `file:` specifiers, so npm reads the staged manifest and installs the package's own dependencies into this repo's tree.

**Nothing here resolves the editor from the npm registry any more.** yalc is gone — script, devDependency, `editor:link`/`utils:link`, both unlink scripts, and the package-lock yalc guard in `.husky/pre-commit`.

Why each part is the way it is:

- **`preinstall`, not `postinstall`** — the staged folders are npm's resolution targets, so they must exist before npm resolves. The script is therefore plain Node importing only the standard library; no devDependency exists yet. (Node 22 strips the type annotations natively, so it stays TypeScript.)
- **A staged copy, not a link to the source** — the source sits in a pnpm workspace whose per-package `node_modules` holds its own `react`, `react-dom`, and `lexical`. Node resolves a link through its real path, so linking the source directly would bind the editor to *those* copies: duplicate React (invalid hook calls) and duplicate Lexical (every cross-boundary `instanceof` node check fails).
- **`workspace:` specifiers get rewritten** to `file:` paths at the sibling staged package. npm cannot parse pnpm's `workspace:` protocol at all. Worth knowing: yalc rewrote it to `~0.1.6` — *the registry* — so even today the linked editor's own utilities dependency pointed at ETEN's published copy, and only worked because this repo linked utilities over the top.

## The move

`dev-packages.json` now points at [paranext/scripture-editors](https://github.com/paranext/scripture-editors), a standalone copy carrying full history and its MIT license. `platform-yalc` resolves the same commits.

- **Merge with [paratext-10-studio#179](https://github.com/paranext/paratext-10-studio/pull/179).** `build.util.ts` matches `dev-packages.json`'s `cloneUrl` to `productInfo.json`'s `uri` by exact string equality *including the trailing `.git`*, and a mismatch is a **warning, not an error** — a half-done change silently builds the wrong revision.
- **Your existing checkout keeps working.** `preinstall` now checks the clone's `origin` and, if it still points at ETEN, stops with the exact commands to fix it. Those commands *keep* the old URL as an `eten-tech-foundation` remote rather than overwriting it, and it states the one-line undo — it prints them for you to run, it does not run them itself. I ran the real cutover against my own checkout to confirm.
- **`chromatic.yml` loses its hardcoded checkout.** It pre-seeded the clone, but preinstall checks out the `dev-packages.json` revision regardless, so the workflow's `ref` never decided anything — it was just a third place holding a URL and branch that would need updating on every move.

## Also here

- **`build:editor`** (mirrors `build:pbr`) — re-stages the editor *and* rebuilds the webpack DLL. Both halves matter for a running dev app: the dev renderer serves the editor from the DLL, so re-staging alone leaves the app on the old build while tests and lint pass against the new one.
- **`@sillsdev/scripture` deduped.** Honoring the editor's declared ranges surfaced that it asks for `^2.1.0` while this lockfile pinned 2.0.5 — previously masked, since the linked build just resolved whatever was in the tree. Refreshed to a single copy. (`@xmldom/xmldom` is also duplicated; that one is pre-existing on `main` and untouched here.)
- README documents the relationship, and an ADR records the decision and the rejected alternatives.

## Verification

| Check | Result |
| --- | --- |
| Clean `npm install` | exit 0 |
| `npm ci` (what CI uses) | exit 0 |
| `npm run build` | exit 0 |
| `npx tsc --noEmit` | exit 0 |
| `npm run lint` | exit 0 |
| `npm test` | all 8 projects pass |
| Editor resolves at | **0.8.16 from source**, not registry 0.8.15 |
| `react` / `lexical` / `yjs` instances | 1 each, no nesting |
| Registry copy of the editor | none anywhere in the tree |

One flaky test: `platform-bible-react` → `book-chapter-control.stories.tsx > Verse Selection From Chapter Grid` timed out at 10.5s under full parallel load, passes in isolation at 6.6s. Unrelated component.

## Merge sequence, and the red windows it implies

These eleven PRs cannot all be green at once. `actions/checkout` of paranext-core in every consumer
workflow takes no `ref:`, so those jobs always build against core's `main`. Merge in this order:

1. **[scripture-editors#1](https://github.com/paranext/scripture-editors/pull/1)** — brings the
   committed `packages/*/dist`, the release workflows, and the consumer-lockfile guard.
2. **Move `platform-yalc` forward onto that new `main`** (`npm run move-platform-yalc` in
   scripture-editors). This is required, not housekeeping: `platform-yalc` currently carries no
   committed `dist/`, so until it moves, core's staging falls back to a pnpm/nx build — and no
   consumer runner installs pnpm or Volta.
3. **This PR (paranext-core#2745).** Depends on step 2.
4. **Regenerate paratext-10-studio's `repo-patches/paranext-core.patch`.** It records studio's
   modifications to core as a diff anchored to core's `main`, so this PR landing moves everything
   it points at. It is generated, not hand-edited: `npm run save-repo-patches` in studio, committed
   to [studio#179](https://github.com/paranext/paratext-10-studio/pull/179) (or a follow-up if that
   has already merged). Nothing warns when it is stale — the patch simply stops applying and
   studio's build fails.
5. **The nine consumer PRs**, as promptly after step 3 as possible — studio#179,
   paratext-bible-extensions#119, paratext-assistant#130, and the six workflow-only PRs
   (paranext-extension-template#163, paranext-multi-extension-template#99,
   paratext-bible-internal-extensions#195, paranext-extension-sneeze-board#22,
   paratext-biblical-terms-extension#62, platform-bible-sample-extensions#69).

Expected red, in order of when it appears:

- **Between steps 1 and 3**, every push to `platform-yalc` fails `verify-platform-yalc.yml`: it
  reads `dev-packages.json` from core's `main`, which still has the pre-staging schema, so there is
  no lockfile contract to compare against. `npm run move-platform-yalc -- --skip-verify` is the
  intended way through, and this is the one window it is for.
- **Before step 3**, all nine consumer PRs are red. Their new `Stage paranext-core's dev packages`
  step runs `.erb/scripts/stage-dev-packages.ts`, which does not exist on core's `main` yet.
- **Between step 3 and each consumer merging**, that consumer's `main` is red the other way: its
  old workflow skips staging, `npm ci --ignore-scripts` links `file:` targets that do not exist,
  npm exits 0 on the dangling symlinks, and the job fails later inside webpack or tsc on an
  unresolved module. This window is why step 4 should not be left sitting.

## Reviewer notes

- The lockfile now records the staged packages' resolved dependencies, so an editor dependency change produces a lockfile commit here. That coupling is new, but it is the honest version of one that already existed silently.
- Install is stricter: `preinstall` needs git, pnpm, and reachability of the editor repo. A failure to stage now fails the install rather than quietly leaving a stale published copy.
- Package names stay `@eten-tech-foundation/*`. Renaming was out of scope.

AI-assisted — [session 1](https://claude.ai/code/session_01VCWPCf35QD5sTkFghsTfUX)

[PT-4500]: https://paratextstudio.atlassian.net/browse/PT-4500?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2745)


tjcouch-sil:
**Updated: staging now copies the editor's committed `dist/` instead of building it.**

[paranext/scripture-editors#1](https://github.com/paranext/scripture-editors/pull/1) commits the editor's built `dist/`, so this repo no longer needs that repo's toolchain — pnpm, nx, a vite build and api-extractor run — during `npm install` just to run the app. Only someone editing the editor builds it (`npm run build:editor`).

- **~7s instead of several minutes** to stage a checkout that has the committed dist.
- `pnpm install` in the dev repo is now conditional on actually needing to build, so the Volta/pnpm failure class leaves the consumer path.
- The publish-shaping the editor's `prepublishOnly` used to do (drop the `development` export and `devDependencies`, rewrite `workspace:` specifiers) now happens on the staged copy here. That is what allows skipping the build — and it means **staging never mutates the editor checkout**, so the manifest snapshot/restore dance is gone.
- Builds use `--skip-nx-cache`: nx declares `dist` a cached target output, so a cache hit restores its copy over the committed one and deletes files the cache predates (observed: 17 spurious deletions).

**No merge-order dependency.** Until the editor PR merges and `platform-yalc` rebases past it, the pinned revision has no committed dist and staging falls back to building — verified working end to end, source checkout left clean. Both paths tested.

.erb/scripts/stage-dev-packages.ts
```diff
@@ -0,0 +1,551 @@
...
+  const currentBranch = execSync('git rev-parse --abbrev-ref HEAD', {
+    cwd: repoPath,
+    encoding: 'utf8',
+  }).trim();
+
+  // Only move a checkout that is not somewhere deliberate. A detached HEAD is where a previous run
+  // left it, `main` is where a fresh clone starts, and the pinned branch is force-pushed upstream
+  // by design — none of the three is work worth protecting. Any other branch is somebody's, and
+  // this script has no business moving it, wherever the checkout lives.
+  const isSomebodysBranch =
```
rolfheij-sil, line 371: **Decision needed (blocking-level for me, but your call):** a clean **detached** checkout is always classified as "safe to move". `git rev-parse --abbrev-ref HEAD` prints the literal `HEAD` when detached, so `isSomebodysBranch` is false for every detached checkout regardless of what commit it sits on or why, and L317 runs `git checkout -B platform-yalc origin/platform-yalc` with only the neutral "Updating…" log line — not the loud warning the "somebody's branch" case gets.

This happened to me during the review: I had a sibling `../scripture-editors` checked out at a PR head to inspect it (clean tree), ran `npm ci` in core, and the script fetched, switched that checkout to `platform-yalc`, and — because `platform-yalc` has no committed dist — ran `pnpm install` (which runs `husky`'s `prepare`) and a build inside my checkout. Nothing was lost (reflog), but nothing told me either.

Suggestions, in order of preference: (a) treat a detached HEAD that is not already at the resolved revision the same as "somebody's branch" — warn and leave it; or (b) at minimum log the same loud warning block for detached/`main` moves as for branches, and document in the README that a sibling checkout may be moved and built into by `npm install`. The dirty-tree guard already protects a dirty detached checkout; it's the clean-and-inspecting case that's exposed.

rolfheij-sil (commented):
🤖 Claude: I reproduced this PR's install path end to end on a fresh worktree (Node 22.22.1, no pnpm on PATH) and on a local rebase onto current `main`, so the mechanics below are verified, not just read:

- `npm ci` from scratch: exit 0 in ~85 s (fallback build path, since `platform-yalc` has no committed `dist/` yet); the real `git clone` path also exit 0 on the rebased tree. pnpm was auto-provisioned by Volta from the editor repo's own `volta.pnpm` pin — nice.
- Editor resolves 0.8.16 from source; staged manifests drop the `development` export and `devDependencies` and rewrite `workspace:` → `file:../scripture-utilities`; exactly one physical copy each of `react`, `react-dom`, `lexical`, `yjs`, `@sillsdev/scripture`; no registry copy anywhere.
- `npm run typecheck`, `npm run lint`, `npm run format:check`, `npm run build` (incl. `dotnet publish`) and `npm test` (all 11 projects) pass on the tip; typecheck + lint + the editor tests pass on the rebase too. All eight extension repos also install, lint and build against this staged core; the editor works in the app (4 of 5 `scripture-editor` e2e specs pass; the 5th fails identically on `main`, so it's pre-existing).

**Two housekeeping items before merge**

1. **The PR is CONFLICTING and its tip has never run CI.** GitHub skips `pull_request` workflows while a PR is unmergeable, so the last CI run is on `4c334f32` (Sep 2) — the 9 commits after it (the hardening, the checkout-preservation fixes, the lockfile guard) are untested on Windows/macOS. The only conflicting file is `.context/standards/Architecture-Decisions.md`, but it is not a keep-both: `main` added nine entries and, per its new `adr-decision-log-sorted-insertion`, now keeps the file in byte order by slug (keep-both silently misplaces entries; nothing enforces the order, so lint won't tell you). I did the rebase locally and pushed it as a read-only side branch you can take or cherry-pick from: `pt-4500-editor-file-link-rebased` @ `8429e7d1c669963d9283330d75b86d4c16694c65` — your 19 commits rebased onto `main` `39eb9a66`, plus three small commits: `09f7779278e` (the two ADR entries moved to their slug positions), `d092e3cb8ae` (item 2 below), `8429e7d1c66` (README, see inline). No other conflicts, no lockfile refresh needed, lint green.
2. **`package-lock.json` is out of step with `package.json`'s `engines.node`.** `f1b1aa93f2c` bumped `engines.node` to `>=22.18.0` in `package.json` but not in the lockfile's root entry, so a plain `npm install` on a clean checkout rewrites `package-lock.json` (one line). `d092e3cb8ae` on the side branch is exactly that line.

**Worth a change (verified)**

- **Tests for the install-path scripts.** There are none, and both scripts run their entry point on import (no `require.main === module` guard), so their pure parts can't be tested as they stand. Your own commit history makes the case: `isSomebodysBranch` (fixed twice, `58537aa`/`c7386b4`), `verifyOrigin`'s URL `normalize()` (`bd5e6fd`: SSH vs HTTPS remotes), the `.staged-from` local/dirty marker logic (`bd5e6fd`), the build-all-or-none decision (`4c334f3`) and `getDeclaredStagingFolders` (`62c80d8`) are short pure functions that were only caught by hand. The repo already has the shape one file over — `startup-waterfall.ts` imports its logic from `startup-waterfall.util.ts` with a colocated `.util.test.ts`, and `tsconfig.erb.json` anticipates it. A `stage-dev-packages.util.ts` split plus ~15 small cases would lock those five in; CI's 3-OS `npm ci` is a good smoke but can't target those branches.
- **`.context/standards/Architecture-Decisions.md` — `adr-single-verse-surfaces-resolve-verse-zero-to-one`.** You touched this paragraph's mechanism wording, and it still says the staged copy "trails published 0.8.15 by one caret-placement line in `$moveCaretToVerseStart`". `platform-yalc` is now `platform_v0.8.15` + the Standard View feature (#545, 134 files), and `ScriptureReferencePlugin.tsx` has ~220 changed lines in that range including that function. Per `docs-durability.md`, either re-run the single-verse verification against the staged 0.8.16 or mark the claim as needing re-verification.

**Hardening suggestions from a silent-failure pass** (take or leave; none is a live defect today)

- `getExpectedMarker` (L339–355) doesn't include `packagePath`/`nxProject`, so editing `dev-packages.json`'s `packagePath` without the editor commit changing leaves stale staged content marked current. Add the path to the marker.
- No lock around the per-repo staging sequence; two concurrent `npm install`s race on clone/`checkout -B`/`rmSync`+copy, and the last writer's marker claims consistency. An exclusive-create lockfile under `dev-packages/` with a clear message would do.
- `cloneRepoIfNeeded` (L172–180) treats any existing directory as a valid clone; a SIGKILL/disk-full mid-clone becomes a confusing "has no revision" on every later run. Check for `.git`/`git rev-parse --git-dir` first.
- `postinstall.ts` `getMissingStagedDependencies` (L64–80) checks only root `node_modules`; a legitimately nested install would fail CI with a wrong "lockfile not updated" diagnosis. `require.resolve(dep, { paths: [stagingDir] })` is more honest.
- The nested self-heal `npm install` (L110–117) isn't in a try/catch, so a second failure prints two stacked errors for one cause; and the top-level catch (L543–548) doesn't say which repo/package was being staged.

**Small things**

- PR body: "preserves the old URL as an `eten-tech-foundation` remote" overstates — `verifyOrigin` (L189–215) stops with the exact commands and the one-line undo, but doesn't run `git remote add`/`set-url` itself. Suggest "tells you how to preserve…".
- Pre-existing and untouched here, so only as a follow-up whenever convenient: `.gitignore`, `.eslintignore`, `.prettierignore`, `.prettierignorerun`, `.stylelintignore` still carry `.yalc`/`yalc.lock` blocks (dead once this merges), and ~18 doc/comment URLs across 10 files still point at `eten-tech-foundation/scripture-editors`.
- Heads-up for the sequencing (not this thread's problem): #2723 (`@sillsdev/scripture` 2.1.0 bump) waits on an ETEN publish of 0.8.16 that this PR makes unnecessary, and its lockfile change overlaps the dedupe done here — Ira may want to rebase it after this merges.

Overall: sound design, good ADR. I'll post a separate comment with a few design questions that are explicitly not blockers.

_Verified claims, for the record — clean `npm install`/`npm ci` exit 0: confirmed (tip + rebase); build/typecheck/lint/tests: confirmed; 0.8.16 from source, not registry: confirmed; one react/lexical/yjs: confirmed; no registry copy: confirmed; "~7 s staging when dist committed": not testable until `platform-yalc` carries `dist/`; "staging never mutates the editor checkout": true for the manifest, not for the checkout itself (branch switch + `pnpm install` in the fallback, see inline); the flaky `book-chapter-control.stories` test did not recur._

.erb/scripts/stage-dev-packages.ts
```diff
@@ -0,0 +1,551 @@
...
+const DEV_REPOS: DevRepo[] = JSON.parse(
+  fs.readFileSync(path.resolve(REPO_ROOT, 'dev-packages.json'), 'utf8'),
+).repos;
+
+/**
+ * Resolve a dev-package repo path. Prefers `dev-packages/<folder>` inside this repo, but falls back
+ * to a sibling directory `../<folder>` so a developer's existing checkout next to this repo keeps
+ * working. Only the staged output has to live inside this repo; the source can be either place.
+ */
+function getDevRepoPath(folder: string): string {
```
rolfheij-sil, line 118: **Warning:** the sibling fallback (`../scripture-editors`) means any clone that happens to sit next to core is used and mutated by a plain `npm install`, whoever put it there. Worth stating explicitly in README's "Relationship" section (today it only says the policy "detached, main, or the pinned branch itself"), since the sibling path is exactly where a developer's own working clone lives. (Broader question in the design-questions comment.)

.erb/scripts/stage-dev-packages.ts
```diff
@@ -0,0 +1,551 @@
...
+      if (!isLocalMode && isStagingCurrent(repo, sourceStamp)) {
+        console.log(`${repo.folder} is already staged from ${sourceStamp}; nothing to do.`);
+        return;
+      }
+
+      // One decision for the whole repo, made while the checkout is still untouched: if any package
+      // has to be built, build them all. Once a build runs, `dist/` no longer says whether a package
+      // was built as itself or as somebody's dependency (see `stagePackage`), so a per-package
+      // decision taken later reads a tree the earlier builds already rewrote.
+      const mustBuild =
```
rolfheij-sil, line 668: **Warning:** `mustBuild` only checks that `dist/` **exists**. An interrupted `nx extract-api` (Ctrl-C, OOM, sleep) leaves a partial `dist/`; the next run sees it, skips the build, and stages the partial output with a valid-looking `.staged-from` marker. The staging copy itself is self-healing (marker written last — nice), but the build side has no completion sentinel. A cheap fix: require the files the manifest's `main`/`types`/`exports` point at to exist before treating `dist/` as usable, or write/check a marker of your own after a successful build.

.erb/scripts/stage-dev-packages.ts
```diff
@@ -0,0 +1,551 @@
...
+      `The ${repo.folder} repo has working changes:\n${status}\nWe don't want to accidentally overwrite any changes. Please go handle your changes and try again when there are no more working changes.\n\nIf you are actively developing ${repo.folder}, build your working state with \`npm run build:editor\` instead.`,
+    );
+  }
+
+  if (isFetchSkipped) {
+    console.log(`Skipping the fetch in ${repo.folder}; using the refs it already has.`);
+  } else {
+    console.log(`Fetching latest in ${repo.folder}...`);
+    try {
+      execSync('git fetch origin --tags', { stdio: 'inherit', cwd: repoPath });
```
rolfheij-sil: **Optional:** no `--prune`. If `platform-yalc` is ever renamed or deleted upstream, an existing clone keeps resolving the stale `refs/remotes/origin/platform-yalc` with no error (fresh clones fail correctly). Hypothetical today; `--prune` is free.

.husky/pre-commit
```diff
@@ -34,16 +34,43 @@ fi
...
-  exit 1
+# `npm run build:editor` stages your uncommitted scripture-editors working tree, and a dirty
+# checkout stages whatever it happens to hold. `npm install` then records THAT package's dependency
+# closure in package-lock.json, and committing the result publishes a lockfile built from work that
+# exists only on your machine.
+#
+# Deliberately only those two. Staging a checkout that sits on a pushed feature branch is a normal
+# way to develop both repos at once, and the lockfile that comes out of it is reproducible by
+# anyone with that branch — so it is committable, and this does not block it.
+if git diff --cached --name-only --diff-filter=ACMR | grep -qx 'package-lock.json'; then
```
rolfheij-sil, line 47: **Warning:** the guard blocks only `-local`/`-dirty` markers, and the commit message's justification ("a pushed branch is reproducible by anyone who has it") isn't checked: a checkout on a **local, never-pushed** editor branch with a clean tree is staged and stamped `<sha> format=2`, identical to a build from the pinned revision, so its lockfile passes the guard. CI's `npm ci` would then fail on the lock/manifest mismatch, so it's caught late rather than never — but if the goal is "block lockfiles built from work that exists on one machine", the check that matches the goal is "does `origin/<branch>` contain HEAD" (`git merge-base --is-ancestor HEAD origin/<branch>`), which is one more line.

README.md
```diff
@@ -229,17 +248,23 @@ To manually set up `scripture-editors` to be linked locally (this should all be
    pnpm install
    cd ../paranext-core
    ```
-3. Run `npm install` in this repo. The postinstall script will automatically run `devpub` in `scripture-editors` and link the packages via yalc.
+3. Run `npm install` in this repo. The preinstall script builds each package in `scripture-editors` and stages it into `dev-packages/staging/`.
```
rolfheij-sil: **Medium:** this sentence and L202 describe the same default `preinstall` and disagree — L202: "no pnpm, no nx — a build only happens if you are editing the editor yourself or your checkout predates the committed build"; here: "the preinstall script builds each package in scripture-editors and stages it". L202 is the accurate one; `8429e7d1c66` on the side branch rewrites this to "stages each package … building it only when the pinned revision has no committed `dist/`". Also worth aligning in the same edit: an adjacent, unchanged line still tells the reader to run `pnpm install` in the editor checkout.

.erb/scripts/postinstall.ts
```diff
@@ -0,0 +1,120 @@
...
+      )}\n\nThis means scripture-editors' dependencies changed but this repo's package-lock.json was\nnot updated to match. To fix: run \`npm install\` in this repo (with the scripture-editors\ncheckout present) and commit the package-lock.json change.\n`,
+    );
+    process.exit(1);
+  }
+
+  // Fresh-clone `npm install`: staging did not exist when npm resolved the tree, so the closure was
+  // skipped. Staging exists now, so running the same install again resolves it properly. The nested
+  // run executes the full lifecycle — preinstall re-stages (a fast no-op via its freshness marker),
+  // and its own postinstall runs the build chain — so this outer run is done when it returns.
+  console.log(
```
rolfheij-sil, line 271: **Suggestion:** unlike `stage-dev-packages.ts`, there's no top-level try/catch here — an uncaught exception still fails the install (fine), just with a less consistent message, and when this nested `npm install` itself fails the inner process's actionable message is followed by a second generic stack from the outer one. (Whether a nested install is the right shape at all is in the design-questions comment.)

rolfheij-sil:
🤖 Claude: Separate from the review findings, five design questions — none blocking, all "how would you like to improve this shape" rather than "change the approach" (Matt and you agreed the approach; nothing here revisits it). Happy to be told no on any of them.

**1. Pin a tag, not the moving `platform-yalc` branch.** `dev-packages.json` pins a branch that `move-platform-yalc` force-pushes, so two installs on different days can run different editor code under identical lockfiles, and a good deal of machinery exists only to keep that moving pointer in lockstep with core's lockfile: `verify-consumer-lockfile-sync.mjs` (querying core's *open PRs* with a 7-day window and an anonymous rate-limit budget), `move-platform-yalc` with `--force-with-lease`, the "survive force-pushed branches" detach logic, and the pre-commit lockfile guard. scripture-editors#1 already mints `v<version>` tags; if `revision` (and studio's `productInfo.json` `branch`) pointed at `v0.8.16`, an editor bump becomes "one core PR that bumps the pin and refreshes the lockfile in the same commit", the branch-sync CI check and the move script become unnecessary, and `platform-yalc` can retire. A tag rather than a bare SHA so studio's `git clone -b` keeps working. Cost: a core PR per bump — which is what a pin is for. Is there a reason you want the pointer to move without a core commit?

**2. Two scripts as `.ts` cost the whole org a Node ≥ 22.18 floor.** Because `preinstall` runs before any dependency exists, the scripts rely on native type stripping, which bumps `engines.node` for core, studio, all eight extension repos and every non-Volta developer machine (it already bit `paratext-biblical-terms-extension#62` at 22.16, and studio pins 22.16 too). Would plain `.mjs` with JSDoc types (still typechecked via `checkJs`, still covered by `tsconfig.erb.json`) be acceptable for these two files? Then no floor moves at all.

**3. Should an npm lifecycle hook mutate a sibling checkout?** `getDevRepoPath` (L108–112) prefers `../scripture-editors` whenever it exists, and staging then fetches, switches branches and (fallback) runs `pnpm install` + a build inside it. Everywhere else our sibling checkouts are read through `file:` links and never written to. I'd make sibling use opt-in (e.g. `--from ../scripture-editors`, or only under `--local`) and keep automatic staging to the in-repo `dev-packages/` clone. That also resolves the detached-HEAD move from the review inline comment at its root.

**4. `postinstall` re-entering `npm install`.** The self-healing path (`postinstall.ts` L110–117) is well guarded (`npm ci` fails fast, the rerun guard prevents loops), but an install that re-invokes npm from inside itself is the kind of cleverness that produces double error output and hard-to-explain states when it doesn't work. Would you consider failing with "staging changed — run `npm install` again" instead? Simpler to reason about, and `npm ci` already takes that path.

**5. Consumers call a file path inside core.** The eight extension repos' CI runs `node .erb/scripts/stage-dev-packages.ts` inside core, while core already exposes `npm run stage-dev-packages`. Renaming or moving the file later breaks eight repos; `cd paranext-core && npm run stage-dev-packages` needs nothing installed and survives a rename. Cheap to change now in the eight PRs.

If you'd rather take any of these as follow-ups, an ADR line saying so is enough for me.

.erb/scripts/stage-dev-packages.ts
tjcouch-sil: Fixed in `99065559e74`, and thank you for the reproduction — that made it concrete.

I took (a), but scoped to **sibling** checkouts rather than all detached HEADs. The reason for the scope: the clone this script manages under `dev-packages/` is detached precisely when the pin is a tag or a commit (the `git checkout --detach` at the end of `checkoutRevision`), so treating every detached HEAD as protected would mean a moved tag pin is never picked up again — and design question 1 proposes moving to tag pins, which would make that the normal case. `isManagedClone` draws the line.

Detail worth stating: the "already at the target" early return above means a detached checkout only reaches the branch policy when it is **not** at the pinned commit, which is exactly your case and never the managed clone at rest.

Reproduced your scenario against a throwaway repo (sibling checkout detached at a PR head, clean tree): old rule classified it `isSomebodysBranch = false` → move and build; new rule → `true`, warns and leaves it. The warning now names the detached commit rather than saying "is on HEAD".

Also documented in the README that a sibling checkout is moved rather than just read, per your other comment.

.erb/scripts/stage-dev-packages.ts
tjcouch-sil: Documented in `99065559e74`. The README's setup section now carries a paragraph saying explicitly that when `../scripture-editors` exists, `npm install` stages from it *and moves it* — fetching and switching branches in it — with the exact policy (refuses a dirty tree; leaves a branch of your own or a detached commit alone with a warning; moves a clean checkout on `main` or the pinned branch), and how to opt out (put your clone somewhere that is not a sibling and let the managed `dev-packages/` clone be used instead).

The detached case is now protected as well, per the inline comment above.

Whether sibling use should be opt-in at all is your design question 3 — I've left that one for TJ rather than deciding it here.

.erb/scripts/stage-dev-packages.ts
tjcouch-sil: Fixed in `99065559e74`. There's no completion sentinel to read, so I took the "check what the manifest promises" option: `isBuiltOutputUsable` collects `main`, `module`, `types`/`typings` and every `./…` string in the `exports` map (walked recursively, since it nests arbitrarily), keeps the ones under `dist/`, and requires all of them on disk before `dist/` counts as usable.

Verified against a fixture: complete build → usable; `index.js` missing → not usable; `dist/` present but empty → not usable; no `dist/` → not usable. Also confirmed it still says usable for the real `packages/platform` staged copy, so it doesn't force a spurious rebuild.

.erb/scripts/stage-dev-packages.ts
tjcouch-sil: Done — `git fetch origin --tags --force --prune`.

`--prune` for the reason you give. `--force` for a neighbouring one I hit while looking: a re-pointed tag makes the fetch exit non-zero with `would clobber existing tag`, and the catch around it then tells the developer they're offline and suggests `PT_SKIP_DEV_PACKAGE_FETCH=1` — which silently stages a stale revision instead of surfacing the one-line real fix. Tags are the release mechanism, so that path is reachable.

.husky/pre-commit
tjcouch-sil: You're right that the check didn't match the stated goal. Fixed in `0e8f9e54469`, though one level down from where you suggested.

Rather than have the hook ask git about the editor checkout, `getSourceStamp` now marks the state at staging time: a clean tree whose HEAD is on no remote branch stamps `<sha>-unpushed`, and the guard treats that like `-local`/`-dirty`. That keeps the hook doing one thing (read the markers) and puts the judgment where the other two provisional states are already decided.

Uses `git branch -r --contains HEAD` — local remote-tracking refs, so no network — and it clears as soon as the branch is pushed. Verified across four states: pushed commit → clean stamp; local-only branch → `-unpushed`; after pushing that same branch → clean; dirty tree → `-dirty` (unchanged). The real pinned revision still stamps clean, so no spurious re-staging.

README.md
tjcouch-sil: Fixed in `99065559e74`. L202 was the accurate one, so the setup step now reads "stages each package from `scripture-editors` into `dev-packages/staging/`, building it only when the pinned revision has no committed `dist/` to copy."

Also took the adjacent `pnpm install` line you flagged: it was an unconditional step 2 telling every developer to install the editor's dependencies. It's now folded into that same step as conditional — only someone editing the editor needs it, which is the whole point of copying the committed `dist/`.

I didn't cherry-pick from `8429e7d1c66` since I'd already rebased by then, but the wording matches what you proposed.

.erb/scripts/postinstall.ts
tjcouch-sil: Both parts fixed in `99065559e74`.

The nested install is now in a `try`, and on failure prints one line saying its error is above rather than rethrowing — the inner process runs this same script, so its message is the actionable one and a second stack from the outer process only buries it.

The top-level catch in `stage-dev-packages.ts` now names what was in flight (`<repo>/<packagePath>`), since with several packages staged in one run "Failed to stage dev packages" alone didn't say which.

On the related suggestion in your review body — `getMissingStagedDependencies` checking only the root `node_modules` — I took the fix but not the `require.resolve` form. `require.resolve(dep, { paths })` throws for a package whose `exports` map declares no `.` entry, and the editor depends on one: `@lexical/react` is subpath-only. I confirmed it reports the current, healthy tree as missing that package, which would fail every install with a lockfile diagnosis that isn't the problem. It now walks the `node_modules` chain from the staged package instead — nesting handled, no false positive (verified: zero missing on the healthy tree, and a genuinely absent package still reports).

tjcouch-sil:
Thanks — this was a genuinely useful review. Everything below is pushed; the seven inline threads have individual replies with the verification for each.

## Housekeeping

**1. Rebased onto `main` and pushed** (`0e8f9e54469`). I did the rebase myself rather than taking the side branch, since by then this branch had moved. The decision log resolved as a union keyed by slug, with my entry moved from where the old append convention had left it to its byte-order position between `adr-decision-log-sorted-insertion` and `adr-disclosure-outside-package-graphs` — 81 entries, sorted, no duplicates.

Two conflicts that weren't in your list, because they came from commits after the point you rebased from:
- `.husky/pre-commit` — `main` added both a yalc guard and the lockfile-root-name guard. Kept the second, dropped the first (yalc is what this PR removes).
- `tsconfig.erb.json` — both sides expanded the same gate. Unioned the `include` lists and both rationales; `exclude` took `main`'s `**/*.test.ts`, which also covers the colocated util test.
- `bump-versions.yml` — `main`'s `npm ci --ignore-scripts` is the better call for a version bump and I kept it, with the comment rewritten (it described the yalc postinstall chain).

Validated with a full `npm ci` from the reconciled lockfile: exit 0, staging produced the editor's complete 13-dependency closure, erb typecheck clean.

**2. Lockfile `engines.node`** — fixed, and it was a symptom: the lockfile hadn't been regenerated since that bump at all. Regenerated properly; the only content changes were `main`'s (the 0.6.0-alpha.1 workspace bump, the `spdx-*` additions, `stringz` → `unicode-segmenter`). No dependency this branch introduces changed version.

## Tests

Done — `stage-dev-packages.util.ts` split with 29 colocated tests, covering the five functions you named. Each test was checked against a mutant rather than only a revert, per the repo's revert-test rule: stripping `.git` handling from `normalizeRepoUrl`, dropping `path=` from the marker, comparing only names in the lock diff, and reverting the env-flag parse to `!!value` each fail the suite (1, 2, 3 and 5 failures respectively).

Scale note: the two scripts are now 825 lines against the 55 in `link-dev-packages.ts`, which I think settles the case.

## Hardening pass

Taken: `packagePath` in the marker; `--prune` (plus `--force`, see that thread); the `cloneRepoIfNeeded` `.git` check; the nested-install `try`; the top-level catch naming what was in flight; and `getMissingStagedDependencies` — though **not** via `require.resolve`, which reports the current healthy tree as broken (details in that thread; `@lexical/react` is subpath-only).

**Not taken: the staging lock.** Two concurrent `npm install`s in one clone is a real race, but a lockfile under `dev-packages/` has its own failure mode — a stale lock after a SIGKILL, which is the state a developer least expects and hardest to explain. I'd rather leave it unlocked and visible than add a mechanism whose failure is a wedged install. Happy to be overruled.

## Small things

- PR body corrected — it now says the script prints the commands rather than running them.
- The five ignore files' `.yalc`/`yalc.lock` blocks are gone.
- `adr-single-verse-surfaces-resolve-verse-zero-to-one`: not re-verified. Flagging rather than silently leaving it — I don't want to mark a claim verified that I haven't run.
- The ~18 `eten-tech-foundation` doc URLs: left alone deliberately. Some are genuine references to the upstream repo (the transferring-work guide, the MIT/contributing-back rationale), so a blanket rewrite would make them wrong. Worth a pass by someone who can tell the two apart.

## Design questions

Answered where I could; **1, 2, 3 and 4 are TJ's calls and I've left them for him** rather than deciding a design question inside a review response. My own read, for what it's worth:

- **Q5 — done.** All eight repos (16 workflow occurrences) now call `npm run stage-dev-packages`. Core's script carries `--experimental-strip-types`, which also means a consumer pinned below Node 22.18 can run it — so `paratext-biblical-terms-extension` no longer needs its Volta pin moved, which is the smaller change in a repo that isn't ours to retool.
- **Q2 is the one I'd most like an answer on**, because it interacts with Q5: if these two files were `.mjs` with JSDoc types, the Node floor disappears for every consumer and the flag above becomes unnecessary. Right now the floor is real and has already bitten one repo.

tjcouch-sil:
TJ's answers on the five design questions, plus one change of mind on the detached-HEAD thread.

**1–4: not doing them, and recorded as `adr-dev-packages-staging-shape-deferred`** (`d13c75f779a`), since you said an ADR line was enough. The branch pin, the `.ts` install scripts, the sibling fallback and the self-heal re-run all stay as they are. The entry states what each alternative would have bought rather than pretending they were rejected on merit — the reason is that all four change the shape of a mechanism about to be exercised across eleven repositories at once, and doing that before it has run in anger trades a known state for an unknown one. It also records the standing costs so the next person hits the reasoning rather than rediscovering the option: the Node floor is real, editor code can move under an unchanged core commit, and a sibling checkout is used and moved by a plain `npm install`.

**5: done**, as noted above — and it is what keeps question 2's cost off the consumers for now, since core's script carries `--experimental-strip-types`.

**On the detached-HEAD fix — TJ overruled my scoping, and he's right.** I had protected a detached *sibling* but still moved this script's own clone under `dev-packages/`, to keep a moved tag pin followable. His reasoning: if you detach somewhere, you are not expecting it to change — and that is true of whoever's checkout it is. With question 1 declined, the tag-pin case I was protecting is hypothetical anyway.

So as of `d13c75f779a` a detached HEAD is deliberate wherever it is found, and `isManagedClone` is gone. The consequence is written at the site rather than left to be discovered: a `revision` naming a tag or commit leaves this script's own clone detached at it, so moving that pin warns instead of following it until someone checks the new one out. `dev-packages.json` pins a branch, which checks out as a branch and never reaches that path.

Verified across every reachable state — detached (either location) → protected; `main` and the pinned branch → moved; any other branch → protected — and the real staging run on the branch pin is unchanged.

**Also now in the merge sequence, which it wasn't before:** regenerating paratext-10-studio's `repo-patches/paranext-core.patch` after this PR merges. It is anchored to core's `main` and this PR moves what its hunks sit in, so it stops applying the moment this lands — silently, since the file is generated rather than reviewed. It is step 4 in the PR description now, and studio#179 carries the same note.

.erb/scripts/stage-dev-packages.ts
tjcouch-sil: Update: TJ overruled the scoping, and I think he's right.

Your original suggestion (a) — treat a detached HEAD that is not at the resolved revision as deliberate — is now what the code does, unconditionally. `isManagedClone` is gone. His reasoning: if you detach somewhere, you are not expecting it to change, and that holds regardless of whose checkout it is.

My objection was that protecting this script's own clone stops a moved tag pin from ever being followed. That is still true, but it is hypothetical now that design question 1 is declined — `dev-packages.json` pins a branch, which checks out as a branch and never reaches the detached path. The consequence is written at the site rather than left to be rediscovered, so whoever changes `revision` to a tag meets it in the code.

`d13c75f779a`. Verified across every reachable state: detached (sibling or managed) → protected; `main` and the pinned branch → moved; any other branch → protected; and the real staging run on the branch pin is unchanged.

tjcouch-sil:
Following up on `adr-single-verse-surfaces-resolve-verse-zero-to-one`, which I'd left flagged rather than resolved — now done in `cce1f56099e`.

You were right about the drift, and it is larger than the note implied: `$moveCaretToVerseStart` is **57 lines against 0.8.15's 30**, 37 of them differing, having gained chapter resolution in its "already here" guard (a number-only match was chapter-blind, so a cross-chapter move to the same verse number wrongly no-op'd). "Trails published 0.8.15 by one caret-placement line" described nothing that exists.

The framing was also obsoleted by this PR in particular. It told the reader to check that the published package and the local build agree — and there is no longer a published copy in the tree to compare against. The staged build is the only editor that runs, which is the whole point of the change.

So rather than only marking it stale, I re-checked the two statements the entry actually rests on against the staged 0.8.16, and both hold:

- `Editor.tsx:924` still mounts `ScriptureReferencePlugin` gated on `scrRef && onScrRefChange` alone — so the plugin does mount under `isReadonly`, and the `ResourceCell` guard is still doing something.
- `$resolvePosition` still opens with `if (!chapterNode && !bookNode) return undefined;` — so the "silent in verse mode" reasoning is intact.

I also confirmed the decision itself is untouched: `resolveDisplayVerseNum` is still at the cited path in core, and `handleScrRefChange`/`sliceUsjToVerse` are core symbols rather than editor ones, so nothing the entry names has gone missing.

What I did **not** do is exercise the behavior end to end, and the entry now says so in those words — given how much moved inside that function, I would rather it name the gap than carry a verification line I did not earn.

PR #2746, opened 2026-09-01
PT-4430: Fix ordering, scope and scroll affordance in the language picker

katherinejensen00:
Fixes **PT-4430** (from bug **PT-4135**, "Issues with Language picker in resource picker").

## Summary

The resource picker's language dropdown listed every language in **DBL catalogue order**, gave no
hint the list continued past the fold, and offered languages that could produce **zero results**.

Four fixes, all in the language filter:

| | Before | After |
|---|---|---|
| **Order** | DBL's arbitrary catalogue order | Alphabetical, with installed languages starred on top |
| **Counts** | none | Resource count per language |
| **Scope** | every language in the catalogue | Only languages that actually have a resource of the requested type |
| **Overflow** | clipped flush on a row boundary — looked complete | A short fade marks "more below" |

Plus localization: the dropdown's search box built `Search ${placeholder.toLowerCase()}...` in code
— English-only, and wrong in any language with different casing rules. It now takes a real localized
string, as does the "no matches" message.

**Why the diff reaches two extensions:** those two strings live on the shared `MultiSelectComboBox`,
so removing the English template affected every consumer. The Checks side panel and both Get
Resources filters each got a localized string of their own rather than being left with a search box
labelled "Check Type" / "Types" / "Languages".

**Get Resources gets all four fixes, not just three.** It shares the language helper, so it picks up
the ordering, starring and counts; the scope fix and the scroll cue reached it during review, the
cue only in the third round. All four rows of the table above are true of both surfaces. That is
the second surface PT-4135 names.

## Where to look

Start here — everything else is supporting:

| File | What it does |
|---|---|
| `resource-picker-dialog.utils.ts` | **`buildLanguageFilterOptions`** — the actual fix. Builds the options: alphabetical, starred when installed, count as secondary label, scoped by `resourceType`. **`matchesResourceType`** is the one predicate deciding what "in play" means, shared by both surfaces. |
| `resource-picker-dialog.component.tsx` | Wires it up. Note **`typeScopedResources`** — rows, options and the total all derive from one list, so they cannot disagree — and **`effectiveLanguages`**, which drops a selected language the filter no longer offers so changing `resourceType` can't strand the user on an empty list. |
| `get-resources.component.tsx` | The same two ideas in the extension, where the stranding case bites hardest: that selection is persisted web view state, seeded from installed resources. |
| `multi-select-combo-box.component.tsx` | The shared component: new `searchPlaceholder` and `showScrollCue` props, **`OptionListScrollCue`** drawing the fade, and a sort order that holds steady while the list is open. |
| `use-has-content-below.hook.ts` | New hook behind the cue. Watches scroll, resize **and mutations** — filtering the list changes content height without resizing the scroller. |

Skim or skip: `resource-picker-dialog.data.ts` (a 132-language test fixture),
`lib/platform-bible-react/dist/**` (regenerated bundle, committed by convention),
`.context/standards/Architecture-Decisions.md` (three entries), `docs/plans/` (working notes).

## Reviewing the behaviour

Storybook is the fastest way to see all of it — **Advanced → ResourcePickerDialog**:

- **Many Languages (~130)** → open "Any language". Starred installed languages on top, alphabetical
  below, counts on the right, fade at the bottom that disappears when you scroll to the end. Pick
  several languages in a row without closing the list: rows stay put under the pointer.
- **Default** → the negative control. Four languages fit, so **no fade**.
- **Many Languages, Scripture only** → languages without a Scripture resource are gone from the list.

All three carry `play` assertions, so they also run in CI.

## One thing worth your attention

**Nothing here has been seen working against real DBL data.** `npm run package` succeeds and the
change is verifiably in the shipped `app.asar`, and the localized strings render in the running
packaged app. But the DBL catalogue never loads in paranext-core (no `DblResource*` .NET user
secrets — registration itself is valid), so the picker holds zero resources. **The ordering, counts,
stars and cue rest on the browser-mode story tests.** If you run paratext-10-studio, please open the
filter once and confirm by eye.

The scroll cue's app-wide blast radius was the other open question here, and it is settled: the cue
is opt-in via `showScrollCue`, default off, and only the two language filters PT-4135 names set it —
not the five-entry type filter beside one of them, and no other dropdown, so everything else looks
exactly as it does on `main`. Whether to flip that default is a design question that can now be
answered on its own.

<details>
<summary><b>More detail</b> — API surface, review findings, testing</summary>

### API changes

- `MultiSelectComboBox` / `Filter`: new optional `searchPlaceholder` prop. The dropdown's search
  input is now `searchPlaceholder ?? placeholder` instead of the composed English template.
- `MultiSelectComboBox` / `Filter`: new optional `showScrollCue` prop, default `false`. When set,
  renders a `[data-slot="command-list-scroll-cue"]` element while the list overflows.
- `MultiSelectComboBox`: with `sortSelected`, the order is snapshotted on the closed-to-open
  transition rather than recomputed on every toggle. The snapshot is taken during render, so a
  reopened list is laid out in the new order rather than committing the old one and reordering
  after paint.
- New export `useHasContentBelow` (`platform-bible-react`), named — it takes
  `(scrollerRef, isEnabled = true)`.
- New exports `buildLanguageFilterOptions` and `matchesResourceType`
  (`platform-bible-react/experimental`) — the Get Resources extension's near-identical private copy
  of the first was deleted in favour of it. `buildLanguageFilterOptions` takes only the resources in
  play: both callers narrow with `matchesResourceType` first and derive their rows and their options
  from that one list, which is where the "no language can dead-end" guarantee lives.
- `matchesResourceType` accepts a single type or a list, and treats an empty list as "no filter",
  which is what an untouched multi-select hands over.
- `RESOURCE_PICKER_DIALOG_STRING_KEYS` gains two keys; five more added across the two extensions.
  Every new key has both an English and a Spanish string.
- No changes to `papi.d.ts` or `platform-bible-utils`.

### Notable implementation details

- **`OptionListScrollCue` has to be its own component.** Inlined into `MultiSelectComboBox`, its
  effect measures the list *before* Radix mounts the popover portal, and the combo box doesn't
  re-render when that mount happens — so it concludes there's nothing to scroll and the cue never
  appears. It looks like an easy simplification; it isn't.
- **The cue's scroller is passed as a ref, not found by selector.** `MultiSelectComboBox` owns both
  the list and the cue, so it creates the ref and hands it to each. Refs attach during commit before
  passive effects run, so the cue's effect still sees the element when the portal mounts — the same
  timing property that makes it a separate component.
- **`effectiveLanguages` derives rather than resets.** A stranded selection is filtered out of the
  rendered list but stays in state until the next toggle, so the user's choice isn't destroyed.
- **The fixture is deliberately falsifiable.** 132 languages emitted on a coprime stride of 37 —
  stable across runs, but non-alphabetical, so the ordering test actually fails if the sort is
  dropped.

### Found and fixed in review

A `/review-paratext` pass over the first version turned up, among others:

- The `searchPlaceholder` fallback silently degrading three call sites (the reason two extensions are
  in this diff).
- `selectedLanguages` not reconciled against the options — reintroducing the dead-end by another
  route.
- The `resourceType` wiring, the scroll listener, the mutation path and the localized empty message
  all being reachable without failing any test. Each now has one.
- The fixture's own documentation being wrong: 47 of its 132 languages lack a Scripture resource, not
  the 4 named. The test passed only because it asserted one direction; it now asserts set equality.
- `Showing N of M` counting M against the whole catalogue rather than the type-scoped one.

A second review round found that the scope fix never reached Get Resources at all, and that
`sortSelected` moved rows under the cursor. Both are fixed above. Two findings were resolved
differently from the suggestion, and are the parts worth re-reading:

- **The `ResizeObserver` stub belongs per test file, not in the shared setup.** The hook-level guard
  is in. The shared-setup stub was tried and reverted: defining `ResizeObserver` globally broke two
  previously-passing suites, reproducibly — `toolbar` and `tab-toolbar` (both expected the widest
  shrink step and got the narrowest, because those components construct one when it exists and
  jsdom's zero widths then resolve that way). Restoring the setup file put both back to 13/13.

  Worth separating out one part of that finding: `cmdk` constructs a `ResizeObserver` whenever a
  command list mounts, so opening any combo box popover already threw before this PR. Verified with
  the cue turned *off*, where the hook attaches nothing — still `ReferenceError: ResizeObserver is
  not defined`. That is why `command.test.tsx` already carries a stub. The cue doesn't introduce the
  trap, though the guard is still worth having.

The hook's ref-vs-query half of that review was right and is taken: the scroller is a ref now, which
also retired the rebinding machinery that only existed to compensate for resolving by selector.

A third round caught what making the cue opt-in had cost: it reached only the resource picker, so
the Get Resources language list still clipped flush at a row boundary — half of the overflow fix,
on the surface with the longer list. It is on there now. The same round retired
`buildLanguageFilterOptions`' `resourceType` parameter, which both callers had stopped passing once
they pre-scoped, and moved the type filter onto the narrowed selection its own predicate already
used — the label-less badge is reachable from any persisted string this build doesn't offer, not
just a deliberately retired type. The sort snapshot moved from a passive effect into render, so a
reopened list is laid out in the new order rather than committing the old one and reordering after
paint.

### Testing

`typecheck`, `prettier`, `eslint` clean across the library and both extensions. 12/12 browser-mode
story tests, 65/65 resource-picker unit tests, 61/61 in the Get Resources extension, 819/820 in the
library's unit project, `build:basic` and the extension build green.

Each new test was checked against a reverted fix rather than only against a green run: the
frozen-sort test fails without the freeze, three of the four original Get Resources tests fail
without the scoping, and the new type-narrowing test fails when the type filter is handed the raw
selection again.

The two ~130-option language filter tests now carry an explicit 30s timeout. Both ran past vitest's
default 5s budget on a developer machine — before this round's changes as well as after, measured
both ways — while CI cleared them; the fixture size is what those tests are about, so the budget
moved rather than the fixture.

Two caveats stated plainly. A full core run shows a rotating handful of timeouts in `web-view.*`
renderer suites; each passes in isolation, none reference changed code, and CI on the same commit ran
them green — the known under-load flake. And the cue itself is only covered in browser mode: the
story tests are what assert it appears, retires on scroll, and re-measures on filter.

### Left for follow-up

Star and count have no accessible name (matches existing Get Resources behaviour); three
hand-maintained copies of the resource-picker string map could resolve from `en.json` instead.

</details>

AI-assisted — [session 1](https://claude.ai/code/session_014VLdfiuYiYvZJDC589t2j8), [session 2](https://claude.ai/code/session_01Km6NCjqy4r1H8WYugvJsvz), [session 3](https://claude.ai/code/session_01DWoJq8QbPu94UMG7MataYz)

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2746)


captaincrazybro (changes requested):
Reviewed the source diff (excluding regenerated `dist/`) against `main`, with each finding verified against the code rather than inferred. Nice work on the fixture — I re-derived it in Node and its documented numbers (132 languages, 47 without a Scripture resource, 6 installed) are all accurate, and the ordering test is genuinely falsifiable via its second assertion.

Two things I'd like addressed before merge, then a set of smaller ones.

---

## 1. The scope fix doesn't actually reach Get Resources

`buildLanguageFilterOptions`' TSDoc promises "selecting a language can never produce an empty result list", but at `get-resources.component.tsx:262` it's called with no `resourceType`:

```ts
const languageOptions: MultiSelectComboBoxEntry[] = useMemo(
  () => buildLanguageFilterOptions(resources),
  [resources],
);
```

while the grid filters by type first (`get-resources.component.tsx:301-311`) and `get-resources.web-view.tsx:42` defaults `selectedTypes` to `['ScriptureResource']`. So on a real DBL catalogue the Languages dropdown still offers a language whose only entries are `XmlResource`/`SourceLanguageResource`, with a count that includes those rows, and selecting it lands on "No results" — the dead-end the PR is about. Get Resources gets the ordering, starring and counts, but not the scope fix.

The reason it can't be wired up as-is is an altitude problem, which is why I'd rather flag it than suggest a one-liner: the helper's `resourceType?: ResourceType` takes a *single* type and can't express Get Resources' multi-select `selectedTypes: string[]`, and `matchesResourceType` isn't exported from `experimental.ts` either, so the extension has no way to express the scoping at all. Widening to `resourceType?: ResourceType | ResourceType[]` (or a predicate) and exporting `matchesResourceType` would let both surfaces share one definition of "in play".

**Caveat if you do fix this:** doing so newly introduces the stranding case on Get Resources, which is the surface where it bites hardest — that selection is *persisted* (`useWebViewState<string[]>('languageFilter', [])`, `get-resources.web-view.tsx:46`) and seeded from installed resources at lines 56-73. The `effectiveLanguages` guard you added at `resource-picker-dialog.component.tsx:189` would need to come along, or a stranded language renders as a blank badge containing only an X (`filter.component.tsx:65` does `entries.find(...)?.label`) and filters every row away.

## 2. `sortSelected` makes rows move under the cursor

`resource-picker-dialog.component.tsx:291` adds `sortSelected`, which the dialog didn't have before. `sortedOptions` (`multi-select-combo-box.component.tsx:142-159`) re-sorts on every `selected` change and floats just-selected non-starred entries above unselected ones, and the popover stays open across toggles. In a ~130-language list, clicking a language near the bottom jumps it up to just under the starred block, so the row now under the pointer is a different language — a user picking several languages in a row will mis-click.

The PR description lists this under "left for follow-up" as pre-existing and shared with Get Resources, which is fair for Get Resources — but this line is what newly imports the behaviour into the dialog, on the exact list the ticket is about. Sorting only on popover open (or on close) keeps the installed-first grouping without moving rows mid-interaction.

---

## Smaller, all introduced here

**3. The localized empty message stopped one call site short.** `commandEmptyMessage` defaults to the hardcoded English `'No entries found'` (`multi-select-combo-box.component.tsx:113`). The dialog now passes a localized one and the Checks side panel already did, but neither Get Resources `Filter` passes it — so searching that dropdown for a non-matching language still shows English. Given the diff deliberately reaches into that extension for the search placeholder, this looks like the same fix missing its last two call sites.

**4. Number formatting now differs between two adjacent dropdowns.** The helper uses `count.toLocaleString()` (`resource-picker-dialog.utils.ts:60`) where the deleted `getLanguageOptions` used `count.toString()`; `typeOptions` at `get-resources.component.tsx:266` still uses `.toString()`. On a de-DE machine with a large catalogue the Languages dropdown shows `1.024` and the Types dropdown directly above it shows `1024`. Worth noting `toLocaleString()` follows the JS runtime's default locale, not the app's i18n locale, so it isn't the localization win it looks like. Either format both through the same path or keep both on `toString()`.

**5. `new ResizeObserver` is unguarded, and now mounts in every popover.** `use-has-content-below.hook.ts:40` constructs it unconditionally. jsdom 26 doesn't implement `ResizeObserver` (I checked — `MutationObserver` yes, `ResizeObserver`/`IntersectionObserver` no) and there's no global polyfill; ~20 test files each stub it themselves. Since `OptionListScrollCue` now wraps the option list for *every* `MultiSelectComboBox`/`Filter`, the next jsdom test that opens one of those popovers will fail with `ReferenceError: ResizeObserver is not defined` for a reason unrelated to what it's testing. Nothing breaks today — `resource-picker-dialog.component.test.tsx` only stubs `IntersectionObserver` and never opens the language popover — so this ships as a latent trap. A `typeof ResizeObserver !== 'undefined'` guard, or adding the stub to the shared setup, closes it.

**6. The assertion advertised as making the ordering test falsifiable is vacuous.** `resource-picker-dialog.utils.test.ts:183`:

```ts
expect(labels).not.toEqual(
  MANY_LANGUAGE_RESOURCES.map((r) => r.bestLanguageName).slice(0, labels.length),
);
```

`labels` is 132 distinct names; the right-hand side is the first 132 of 264 resources, which I confirmed contains only 66 distinct languages — it's full of repeats. Drop the `.sort()` from `buildLanguageFilterOptions` and `labels` becomes deduped catalogue order, still unequal to a duplicate-bearing slice, so `not.toEqual` still passes. The comment above it ("this fails if the sort is dropped and catalogue order leaks through") isn't true; only the next line catches it. `[...new Set(MANY_LANGUAGE_RESOURCES.map((r) => r.bestLanguageName))]` would make the claim hold.

**7. `total` re-scans the catalogue on every render.** `resource-picker-dialog.component.tsx:300` runs `allResources.filter((r) => matchesResourceType(r, resourceType)).length` inline in JSX, duplicating the first stage of the memoized `filteredResources` at line 198. With `LARGE_SAMPLE_RESOURCES` (2500) every keystroke pays two full passes. Hoisting one `typeScopedResources` memo and feeding it to `filteredResources`, `buildLanguageFilterOptions` and `total` removes the duplication and makes the three provably agree on the same scoped set — which is the invariant `matchesResourceType`'s own TSDoc claims.

**8. The hook's stated rationale doesn't hold.** `use-has-content-below.hook.ts:11-12` says it resolves by query "so it works against a vendored component whose ref forwarding is not ours to rely on" — but `CommandList` is in-repo (`shadcn-ui/command.tsx:168`) and spreads `...props` onto `CommandPrimitive.List`, and this is React 19, where `ref` is a plain prop. A ref would be simpler and type-checked, where the `data-slot="command-list"` string is verified by nothing but the browser-mode story. Relatedly, the effect's deps are `[containerRef, selector]` — a ref object and a string literal, both stable — so it resolves the scroller exactly once. That's fine for `OptionListScrollCue` (the list is mounted when the cue mounts), but this is now a public export, and for a consumer whose scroller mounts later or gets replaced the observers stay bound to a detached node and the cue freezes.

## Nits

- `resource-picker-dialog.stories.tsx:169`: the cue-*present* assertion is a bare synchronous `querySelector(...).not.toBeNull()`, while both cue-*absent* assertions use `waitFor` — and appearance is the direction that needs an effect plus a re-render. Given the description notes a full browser-mode sweep hasn't run clean, worth making it symmetric.
- `use-has-content-below.hook.ts:18` exports default only. Every other hook in that directory exports a named binding (most also a default alias), so `index.ts:324` has to spell it `export { default as useHasContentBelow }` and an in-library named import fails.
- `resource-picker-dialog.utils.test.ts:99`: the computed `dblEntryUid: \`uid-${...}-${overrides.dblEntryUid ?? '0'}\`` is dead whenever an id is passed — `...overrides` is spread after it and wins. Harmless now, misleading for the next test that wants distinct readable ids.
- `filter.stories.tsx` and `multi-select-combo-box.stories.tsx` weren't updated for `searchPlaceholder`, so the library's own reference stories now show a bare noun ("Types", "Fruits") in the search box — the opposite of what the change is demonstrating.

## Process

`.context/standards/Architecture-Decisions.md` isn't touched. Per the root `CLAUDE.md` ("Recording Architecture Decisions ... This applies to **all** work"), at least three here qualify: introducing an app-wide scroll-cue affordance in a shared component, deciding the language helper lives in `platform-bible-react/experimental` rather than the extension, and changing a published component's localization contract so callers must supply the dropdown search placeholder. Without an entry the next PR touching `MultiSelectComboBox` has to re-derive all three from this description.

---

On the scroll cue being app-wide, since you asked for a sign-off: I don't think it needs to be reverted, but it currently has no opt-out — no `showScrollCue` prop — and `tw:from-popover` will read as a mismatched band against any popover themed with a different background. An opt-in prop that only the resource picker sets would confine the blast radius to the surface the ticket is about and let the design question be settled separately.

One thing I checked and want to explicitly clear: I looked for out-of-repo fallout from the `searchPlaceholder ?? placeholder` change and found none — no consumer of `MultiSelectComboBox` or `Filter` in paratext-10-studio, paratext-bible-extensions, or paratext-bible-internal-extensions, and all in-repo production call sites are updated. Only the stories above were missed.

<sub>AI-assisted review (Claude Code).</sub>

katherinejensen00 (commented):
All eight findings, the four nits, and the process item are in the last four commits. One is worth your attention because I resolved it differently; the other one I'd flagged as a disagreement turned out to be mine to concede.

#8 — you were right, the ref is in

I initially kept the query on the grounds that OptionListScrollCue takes the list as children and never renders the scroller, so there was nothing to attach a ref to. That was reasoning about the wrong component: MultiSelectComboBox owns both the list and the cue, so the ref lives up there and gets handed to each. The mount-ti
Taking it removed the selector string that nothing type-checked, and with it the container mutation observer I'd added only to rebind when a query-resolved scroller mounted late or was replaced. React tracks that now. Story tests and the unit project stayed green through the swap.

#5 — the stub belongs per test file, not in the shared setup

The hook-level guard is in. I tried the shared-setup option and reverted it: defining ResizeObserver globally in vitest.setup.ts fails toolbar and tab-toolbar — both assert the widest shrink step by default and get the narrowest, because those components construct an observer when one exists and jsdom's zero widths then resolve that way. Restoring the setup file puts both back to 13/13. So the global stub silently changes what unrelated tests measure.MultiSelectComboBox or Filter popover already threw before this PR. I checked this directly rather than inferring it — a probe with the cue turned off, where the hook attaches nothing, still throws. That's why command.test.tsx already carries a stub. The cue doesn't introduce the trap, though the guard is still worth having.

On the sign-off you asked for

Took your suggestion: showScrollCue, default off, set only by the resource picker. Every other dropdown is identical to main. Whether to flip the default is now a standalone design question rather than a rider on this PR.

#1 and #2, the blocking pair

#1 went past the wiring. matchesResourceTyp

Three ADR entries added, in byte-order slug position per #2770: adr-shared-option-list-affordances-opt-in, adr-shared-list-scope-predicate, adr-shared-components-take-localized-strings.

One thing your review surfaced indirectly: the branch had added seven localization keys with no Spanish. resource-picker-dialog-localization.test.ts caught two in CI; the other five live in extension contribution files that no test covers and would have shipped as a silent English fallback. All seven are translated now.

> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2746#-:-P0iSvd290dKDYOf-RNL:b-qh2oc9)* status: 0 of 40 files reviewed, all discussions resolved.

katherinejensen00 (commented):
Correction to my last comment: the #8 pushback was wrong, and the ref is in as you suggested (c9e6c90dfca).

I'd argued that OptionListScrollCue has nothing to attach a ref to, since it receives the list ag about the wrong

I'd argued that OptionListScrollCue has nothing to attach a ref to,it receives the list ag about the wrongcomponent — MultiSelectComboBox owns both the list and the cue, so belongs one level up aount-timing property Ifects run.

It's a clear improvement rather than a wash: it removed the selectostring that nothing tyontainer mutationobserver I'd added only to rebind a query-resolved scroller that mounted late or got replaced — the one path I'd flagged as having no test. tests and the library'hrough the swap.

#5 stands as written. That one I did reproduce: adding the stub to vitest.setup.ts fails restoring it puts bothback to 13/13.

> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2746#-:-P0iVkfy43C5_ahqzrKI:b87jbc6)* status: 0 of 40 files reviewed, all discussions resolved.

captaincrazybro (commented):
Re-reviewed against `3c01d1d`. All eight findings, the four nits and the process item are addressed, and I verified each against the code rather than against the reply. CI is green on that exact SHA.

Two things I want to say explicitly before the residuals, because both were deviations from what I suggested and both are right:

**#5 stands, and it's the better resolution.** I reproduced your reasoning from code you didn't write: `use-shrink-step.hook.ts:93` feature-detects `typeof ResizeObserver === 'undefined'` and no-ops, and both `toolbar.component.tsx` and `tab-toolbar-container.component.tsx` consume it while neither test file stubs the observer. So a global stub genuinely flips those suites from "no observer, keep the default widest step" to "measure jsdom's zero widths, pin to the narrowest" — exactly the 13/13 → failure you saw. Your secondary point holds too: `command.test.tsx` is pre-existing on `main` and already carries its own stub, with the comment "cmdk instantiates a ResizeObserver on mount". The cue didn't introduce the trap. Hook-level guard plus per-file stubs is the correct shape; I was wrong to offer the shared setup as an equal option.

**#1 went further than I asked** — narrowing the persisted type selection against `RESOURCE_TYPES` wasn't in the finding, and it's the right instinct. See residual 3, which is the other half of it.

---

## Three residuals

None is a missed finding; all three are new surface from the fixes.

**1. `useHasContentBelow` no longer handles a late-mounting or replaced scroller, and "React tracks that now" isn't quite the mechanism.** The effect's deps are `[scrollerRef, isEnabled]` — a ref object identity and a boolean. React attaches refs during commit, but a change to `scrollerRef.current` does not re-run an effect, so the observers stay bound to whatever element was current when it last ran. That's fine for `OptionListScrollCue`, where the cue and the list mount and unmount together inside the portal — but the ref swap also deleted the container MutationObserver that used to cover the replaced-scroller case, so relative to the query version that half went from covered to uncovered. It only matters because `useHasContentBelow` is exported from `index.ts`: a consumer whose scroller mounts after the hook, or gets swapped, gets a cue that never updates. The type-safety win is real and worth keeping; the cheapest close is a line in the TSDoc saying the scroller must be mounted for the hook's lifetime, so the next consumer doesn't discover it the hard way.

**2. The sort snapshot is taken after paint, so reopening the list shows one frame of the old order.** `multi-select-combo-box.component.tsx` sets `sortSelection` in a `useEffect` gated on `actualIsOpen`. `useEffect` runs after paint, so the reopened popover paints with the previous snapshot's ordering and then reorders — a just-selected language sits at the far end of the list for a frame before jumping up under the starred block. It lands inside the popover's own open animation, so it reads as a jump rather than a flicker. Both new tests pass either way, since they assert final order. `useLayoutEffect` removes it.

**3. Get Resources' type filter got half of the language treatment.** `selectedResourceTypes` narrows the persisted selection against `RESOURCE_TYPES`, which is the fix — but the `Filter` still gets the raw `selected={selectedTypes}` (`get-resources.component.tsx:538`), where the language filter gets `effectiveLanguages`. So a persisted type that has been retired from `RESOURCE_TYPES` renders as a label-less badge: `filter.component.tsx:67` does `entries.find((entry) => entry.value === type)?.label`, and `typeOptions` is built from `RESOURCE_TYPES`. Lower severity than the language case you fixed — the X still removes it, and because the narrowing drops it from the predicate the grid shows everything rather than nothing, so it isn't a dead end. Passing a narrowed selection to that `Filter` too would make the two filters symmetric, and is the shape someone will expect after reading `effectiveLanguages` twelve lines below (`:550`).

---

One thing I checked and want to clear rather than leave as a question: the `<div className="tw:relative">` from `OptionListScrollCue` wraps every combo box's option list whether or not the cue is on, so the DOM change is app-wide even though the gradient is gated. It's layout-neutral — `CommandList` keeps its own `tw:max-h-72` and `tw:overflow-y-auto`, and the wrapper is a block-level flex child with `min-height: auto`, so nothing shrinks differently inside `Command`'s flex column. Worth knowing it's there, not worth changing.

<sub>AI-assisted review (Claude Code).</sub>

captaincrazybro (commented):
Last pass over the four commits (`aadaee4499a..3c01d1d276a`), looking only for things the fixes themselves introduced. Two new items, and two amendments to what I posted above — one of which is me walking back an overstatement.

## New

**A. The scroll cue never reached the Get Resources language filter.** `showScrollCue` has exactly one setter in the repo, `resource-picker-dialog.component.tsx:471`. The Get Resources language `Filter` (`get-resources.component.tsx:550`) takes `sortSelected` but not the cue, so on a real DBL catalogue that list clips flush at a row boundary exactly as before — the overflow fix stops at one of the two surfaces PT-4135 names. This is a direct consequence of making the cue opt-in for the sign-off I asked for, so it may well be the deliberate choice; if so it's worth a word at the call site, because the next person to read `sortSelected` there will wonder why its sibling prop is absent.

Worth a clarification in the description either way. Your enumeration — "ordering, starring and counts — and, since the review below, the scope fix too" — is internally consistent and doesn't claim the cue. But it sits under "**Get Resources gets all four fixes, not just three**", and the table's four rows are Order / Counts / Scope / **Overflow**, so a reader maps "all four" onto the table and concludes the cue shipped there. Not a false claim, just a sentence that invites the wrong reading.

**B. A TSDoc block now documents the wrong symbol.** `RESOURCE_TYPES` was inserted between the localization doc comment and its export, so `get-resources.component.tsx:61-66` now tells the reader that `RESOURCE_TYPES` is the "object containing all keys used for localization in this component" that you "pass into the useLocalizedStrings hook". `GET_RESOURCES_STRING_KEYS` is at `:74`, undocumented. Moving the constant above the comment fixes it.

**C, minor — `buildLanguageFilterOptions`' second parameter is now unexercised in production.** Widening it to `ResourceType | ResourceType[]` was my finding #1's suggestion, but both callers ended up pre-scoping and calling it with one argument (`resource-picker-dialog.component.tsx:308`, `get-resources.component.tsx:403`); before this round the dialog passed `resourceType` through. Deriving everything from one scoped list is the better pattern and your `adr-shared-list-scope-predicate` says so, so this isn't a defect — but the helper is an `experimental` export whose TSDoc still presents `resourceType` as the mechanism for "selecting a language can never produce an empty result list", when the invariant is now enforced by the caller. Either drop the parameter or re-point the TSDoc at where the guarantee actually lives.

## Amendments to what I posted above

**Residual 2 was overstated — my error.** The mechanism holds: `multi-select-combo-box.component.tsx:171-173` applies the snapshot in a passive effect, so a reopened list commits with the previous snapshot's order. But React 19 schedules passive effects on a MessageChannel macrotask, and when the commit finishes inside the frame budget that effect lands before paint and nothing is ever visible. I wrote it as though the jump were certain; it isn't. It becomes visible only when the commit overruns the frame — likeliest on the ~130-row list this ticket is about, but conditional. Treat it as a latent artifact rather than a live one. The suggestion is unchanged, and deriving on the closed→open transition during render (rather than `useLayoutEffect`) has the side benefit of retiring the render-time `selectedRef.current = selected` write at `:169`.

**Residual 3 is broader than I described.** I framed it as needing a `ResourceType` to be retired from `RESOURCE_TYPES`. It doesn't: `selectedTypes?: string[]` (`get-resources.component.tsx:262`) is a plain `string[]` fed from `useWebViewState<string[]>('typeFilter', ...)`, so any non-canonical persisted string produces the label-less badge — a renamed type, a hand-edited layout, a value from a future build. Still low severity for the reasons I gave, but the trigger is wider than a deliberate retirement.

**Residual 1 — one nuance in your favour.** I traced all three versions of the hook, and against the state I *originally* reviewed the ref swap is neutral on the rebinding axis, not a regression: `ad1689b` resolved once by query with deps `[containerRef, selector]` and no rebinding either. The `containerObserver` appeared in `dde96dc` and left in `65475e1`, so my "went from covered to uncovered" is true of the commit you replaced and not of the branch as I first saw it. The finding still stands on the public-export argument alone — just not as a regression.

## Cleared

The committed `dist/` is built from the final source, which I checked rather than assumed: `65475e1` (the ref swap) didn't touch `dist`, but `resizable-BYoZgQbJ.js.map`'s embedded sources for both `use-has-content-below.hook.ts` and `multi-select-combo-box.component.tsx` contain `scrollerRef` and no `querySelector`, so the rebuild in `3c01d1d` picked it up.

Also considered and cleared: the unscoped `typeOptions` counts (correct — a type filter's own counts must not depend on its own selection, while the language counts should and now do); the reordering of the text filter to sit after type scoping (result-equivalent); and `matchesResourceType`'s empty-array-as-no-filter semantics (no caller depended on the old meaning).

<sub>AI-assisted review (Claude Code).</sub>

captaincrazybro (changes requested):
Final pass against `e0c9269c031`. All six residuals from my last two comments (R1–R3, A–C) are addressed, and I checked each one in the code: the lifetime note on `useHasContentBelow`, the snapshot taken when the list opens during render (plus Select All / Clear All refreshing it), the narrowed `selectedResourceTypes` on the type `Filter`, `showScrollCue` on the Get Resources language filter, the TSDoc now sitting on `GET_RESOURCES_STRING_KEYS`, and `buildLanguageFilterOptions` taking one argument with its guarantee credited to the caller. Confining the `OptionListScrollCue` wrapper so it renders nothing extra when the cue is off is a nice extra.

I found nothing major the earlier rounds missed. One item blocks merge, one design gap is worth a decision, and the rest are small.

---

## Blocking

**1. The branch conflicts with `main` again, and nothing has run CI on this head.** #2833 (PT-4551) landed on the resource picker after your rebase. `git merge-tree origin/main e0c9269c031` reports content conflicts in:

- `resource-picker-dialog.component.test.tsx`
- `resource-picker-dialog.stories.tsx`
- `experimental.ts` — `main` now exports `focusResourcePickerOnOpen` right next to your new `buildLanguageFilterOptions` / `matchesResourceType` export
- every `platform-bible-react` `dist` bundle

When you reconcile, please also check that `main`'s new focus behaviour still works with `effectiveLanguages`. The only check reported on this head is Reviewable. The last green CI I verified was on `3c01d1d`, which is no longer in the branch history, so none of the last four commits have been through CI.

## Worth a decision

**2. A hidden language either comes back or is silently lost, depending on what the user clicks.** In `get-resources.component.tsx:424`, `effectiveLanguages` hides a saved language that the current type filter doesn't offer, but leaves it in the saved `languageFilter` state. The `Filter` then builds every change from the list it was handed (`filter.component.tsx:63`), which is `effectiveLanguages`, the version with hidden languages removed. Example: saved languages = `['Greek']`, type filter = Scripture, and Greek has no Scripture resource.

- **The user widens the type filter to include Source-language resources without touching languages:** Greek comes back as an active filter, and the grid suddenly narrows to Greek only while they're browsing.
- **The user toggles any language, or removes another badge, while Greek is hidden:** Greek is dropped from saved state for good. That is the same loss `adr-shared-list-scope-predicate` gives as its reason for not resetting the selection when the type filter changes.

Neither outcome crashes or leaves the grid empty, so this isn't a regression in the sense the PR is fixing. But the result depends on the path the user took, and it doesn't match the rationale in the architecture decision entry. `ResourcePickerDialog` has the same shape: a retry whose catalogue includes a hidden language again brings its filter back. Choosing one on purpose (keep hidden languages or drop them) and saying so in the entry would settle it.

## Minor

**3. One test can't fail when it should.** `resource-picker-dialog.component.test.tsx:552`, "re-sorts the selection to the top the next time the list is opened":

```ts
expect(reopened.indexOf(targetLabel)).toBeLessThan(optionsBefore.length - 1);
```

If the label is missing from the reopened list, `indexOf` returns `-1`, which still passes. That happens if the list didn't render, the row's text changed, or the option was dropped. The assertion also only checks "not last". Asserting `>= 0` and the exact expected position (right after the starred group) would make it catch both.

**4. `adr-shared-components-take-localized-strings` vs. what ships.** The context cites the hardcoded English `'No entries found'` default as part of the problem, but `multi-select-combo-box.component.tsx:136-138` still defaults `commandEmptyMessage`, `selectAllText` and `clearAllText` to English. The Consequences section does admit optional props aren't enforced, so this is a wording nit: either say the defaults were kept on purpose, or drop them.

## Smaller, lower confidence

I spotted these but didn't trace them as far as the items above:

- The "hide languages the filter doesn't offer" derivation is written the same way twice (`resource-picker-dialog.component.tsx:328`, `get-resources.component.tsx:424`). A small shared helper would give the stranding rule one definition, especially if #2 changes it.
- Now that `effectiveLanguages` is always a subset of `languageOptions`, the `if (matchingType)` fallback around `resource-picker-dialog.component.tsx:418` looks unreachable.
- `RESOURCE_TYPES` (`get-resources.component.tsx:64`) is a hand-kept list with no completeness check, and it now decides which saved type values survive. If someone adds a new type and forgets the list, that type is quietly dropped from saved selections. `labelByType`'s `Record<ResourceType, string>` would catch it if the list were derived from that record.
- The `selectedTypes = []` / `selectedLanguages = []` defaults (`get-resources.component.tsx:304-305`) create a new array every render, which breaks the new memo chain whenever the props are omitted. A module-level empty array, like `emptyResources`, avoids it.
- The `expect` calls inside `forEach` at `resource-picker-dialog.stories.tsx:326` aren't awaited, unlike the rest of the play function.

---

Resolve the conflict and get CI green on the new head, and I'm happy to approve. #2 and #3 would be good to handle in the same pass. The rest is optional.

<sub>AI-assisted review (Claude Code).</sub>

katherinejensen00 (commented):
Rebased onto main at 0812270 (after both #2833 and #2829). The focus-on-open behaviour and the language filter don't interact: the focus catch-up is driven by the type-scoped resource list, and the language selection only narrows rows within it.

#2: I went with keeping hidden languages. Example: saved languages = Greek, type filter = Scripture. Greek is hidden and doesn't filter the grid. If the user then picks Amharic, the saved selection becomes Greek + Amharic, not just Amharic. When they widen the types to include source-language resources, Greek applies again. Dropping it instead is the same loss adr-shared-list-scope-predicate already rejects for a reset, so the entry now states this: the held/offered split lives in one helper, partitionFilterSelection, used b
#3: now asserts the exact index right after the starred group.

#4: the entry now says the English defaults are kept on purpose as fallbacks.

Lower-confidence items: all five done. There's one shared helper for the stranding rule. The single-language label fallback is gone, and a comment says why the lookup can't miss. RESOURCE_TYPES is derived from a Record<ResourceType, key> of label keys, so a new type won't compile until it's listed. The omitted-prop defaults are module-level, and I did the same for localizedStringsWithLoadingState, whose [{}, false] default had the same memo problem. The story assertion is now a single awaited expect.

> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2746#-:-P2F8do56PC5Z65Czy8T:b-yye00b)* status: 0 of 55 files reviewed, all discussions resolved.

captaincrazybro (approved):
Approving. All nine items from my last pass are done, and I checked each against `2feb5046477` rather than against the reply.

The blocking one is fully clear: all six checks pass on this exact head, and against current `main` there are no source-level conflicts left — `experimental.ts`, both test files and the stories all auto-merge. Only the `dist/` bundle still collides, which is inherent to committing it.

On #2, keeping hidden selections is the right call and the entry now carries it properly: the decision, the rejected "drop it on the next change" alternative, and the consequence you accepted rather than hiding it. `partitionFilterSelection` being one exported helper with its own tests also closes the duplication item. #7 is better than what I suggested — deriving `RESOURCE_TYPES` from the label `Record` makes the compiler the completeness check, so the failure mode I described can't recur. And extending #8 to `localizedStringsWithLoadingState` caught a memo problem I hadn't looked for.

I also checked the two ways this change could have broken something, and both are clean. `clearFilters` calls `setSelectedLanguages([])` directly instead of going through the held write-back, so an explicit clear still wipes held languages — routing it through `handleLanguagesChange` would have left one armed and silently reapplying later. And inlining the type labels preserves the old behaviour exactly, since `getLocalizedString` already fell back to the raw key. `offered` and `held` are disjoint by construction, so the spread can't duplicate.

---

Three small things, none blocking. Take them or leave them.

**1. The comment above `isLanguageFiltered` still argues against its own test** (`resource-picker-dialog.component.tsx:449-456`). Its second half — "calls it no filter when the catalog has changed under a mounted picker such that a still-selected language is no longer among the options — which is the worse direction, because it withdraws the 'Clear filters' escape from a selection that is hiding every row" — describes a state that can no longer occur: `effectiveLanguages` holds only values the options offer, so a selection hiding every row is unreachable, and the suite now asserts the opposite of what the comment warns about. The comment predates this branch, but this branch is what made it false. The first half (measure against the type-scoped set, not the whole catalog) is still true and worth keeping.

**2. `resource-picker-dialog.stories.tsx:403` is the one place the scoping is checked in only one direction.** `await expect(offered.size).toBeGreaterThan(0)` passes if a bug over-filters the list down to a single language; the `languagesWithoutScripture` assertion above it only catches under-filtering. `scriptureLanguages` is already computed at line 389, so `expect([...offered].sort()).toEqual([...scriptureLanguages].sort())` is available. The both-directions form already exists for the pure function in `resource-picker-dialog.utils.test.ts` — this is the component wiring.

**3. Types drop where languages now hold.** The type `Filter` still takes the narrowed `selected={selectedResourceTypes}` with a raw `onChange={onSelectedTypesChange}` (`get-resources.component.tsx:560-561`), so an unrecognized persisted type is gone from saved state after the next toggle, while an unoffered language is held. I think that is the right asymmetry — a type retired from the build can never be offered again, whereas a hidden language comes back the moment the type filter widens — but `adr-shared-list-scope-predicate` now presents holding as the rule without noting that the type filter is a deliberate exception. One clause in the entry would stop the next person reading the two filters as inconsistent.

One practical note: you rebased onto `0812270` and `main` has moved since, so `dist/` will need regenerating before merge.

<sub>AI-assisted review (Claude Code).</sub>

PR #2749, opened 2026-09-01
PT-4467: Show reference-panel resources in the Find project picker

captaincrazybro:
## Summary

Find's project picker listed only projects open in a Scripture **editor** tab, so a resource shown
in a read-only reference panel (Model text, Bible texts, Commentaries) could not be selected. Ctrl+F
from such a panel was worse than useless: it handed Find that resource's project id, Find's
reassignment effect found no matching tab, took its cross-project fallback, and silently re-ran the
search against an unrelated project (PT-4467).

This PR is **purely the consuming side**. All three panels already declare the resource they display
under `NAVIGABLE_PROJECT_IDS_WEB_VIEW_STATE_KEY` (see `adr-navigable-project-ids`); nothing read it.
`useOpenProjectTabs` now does, and Find lists and targets those tabs.

### Why review this

Not part of the current epic. It fixes a collision between two merged tickets — PT-4341 made Ctrl+F
pass the displayed resource's project, PT-3362 restricted the picker to editor tabs — which leaves
Ctrl+F from any reference panel silently searching the wrong project. Worth reviewer time because
the broken path is the common one in Simple mode, where the reference panels are always on screen.

## Changes

- **`useOpenProjectTabs` reads declared projects**, behind an opt-in `includeNavigableProjectIds`.
  The declared list is **unioned** with the container `projectId`, which is what the shared key
  documents and what the toolbar's `useOpenProjectBookIds` already does. Each tab carries
  `projectSource` (`'container' | 'declared'`) so a caller can narrow that. Because a view can
  declare several projects — the Text Collection hosts many — tabs are keyed by (web view, project),
  and a view's previous entries are dropped before its new ones so a shrinking list cannot leave a
  stale project on offer.
- **Find applies its own rule**: a reference panel's container project is never offered, because the
  panel is not displaying that project's text.
- **Result activation splits by tab kind.** An editor is driven through its web view controller
  (select + highlight). A reference panel registers no controller — handing one of their ids to
  `useWebViewController` waits ~20s for a controller that never arrives — so activating its tab is
  the whole navigation, and the scroll group carries the reference.
- **An editor wins wherever several tabs of one project qualify**, in all three branches that pick
  among them. Panels are pinned to scroll group 0, so ordering by scroll group alone systematically
  preferred a panel over an editor in any higher group, leaving `resolveTargetEditorWebViewId` with
  no match and silently dropping the editor's select and highlight on every result click.
- **The Bible texts and Commentaries panels scroll to the referenced verse** when revealed, and
  highlight it. Keyed on visibility and the reference, with guards for the panel's own scroll-group
  echo, a bare reveal (so a manual scroll survives a tab switch), the previous chapter's USJ still
  being on screen, and a scroll that did not land.
- **Picker copy** now says "projects or resources", via new localized keys with `fallbackKey`
  redirects for the old ones, per the immutability rule in `Localization-Guide.md`.

## Needs UX sign-off

**One gesture, two outcomes, with nothing signalling which you get.** "Go to result" on a project
open in an editor keeps Find visible and focuses the editor. On a resource open only in a Bible
texts or Commentaries panel, it activates that panel — which shares Find's Column 3 stack in Simple
mode, so the results list disappears. It follows that for such a resource you cannot see the results
list and the resource text at the same time in Simple mode. This is inherent to the fixed
three-column layout rather than a defect, and it is documented at the call site, but it is a
user-facing behaviour change on the common path and deserves an explicit decision.

Note this applies to Bible texts and Commentaries only. The Model text panel sits alone in Column 1
and is excluded from the reveal set (see below), so it is never fronted by a result.

## Known gaps, deliberately not fixed here

- **The Model text panel is searchable but not a reveal target.** Its scroll effect is keyed on the
  reference alone, so a reference arriving while its tab is hidden finds no layout in a
  `display: none` iframe and nothing re-runs on activation. Rather than ship a stale-position case,
  `REVEALABLE_REFERENCE_PANEL_WEB_VIEW_TYPES` withholds the reveal. Costs nothing in Simple mode,
  where that panel is always visible and the scroll group has already moved it.
- **The Text Collection is not searchable** — filed as PT-4507. It declares every resource it hosts
  and would be nearly free to list, but "go to result" has no defined behaviour when each cell is
  its own editor.
- **A panel detached from every scroll group** is recorded as group 0 by `useOpenProjectTabs`
  (pre-existing on `main`), so a reveal could front a panel parked at an unrelated reference. This
  PR makes that reachable from Find rather than creating it.
- **The scroll effect has no test of its own.** Its three decision functions are unit-tested
  (`scrollToVerse`, `isEchoOfPublishedScrRef`, `hasNewScrollTarget`); the effect that composes them
  is not. Deliberate call, noted here rather than left implicit.

## Testing

- [x] 1911 tests pass across `platform-scripture` and `platform-scripture-editor` (111 files)
- [x] ESLint and Stylelint clean
- [x] Manual verification in Simple mode — picker lists panel resources, "go to result" reveals and
      scrolls the panel, and a verse click inside a panel no longer jumps

## AI Involvement

AI-assisted. Claude Opus 5 wrote most of the implementation and tests under review, ran a
`/review-paratext` pass over the branch, and worked two rounds of review feedback with the author.
The branch was rebuilt on `main` after review round 1 found that the publishing mechanism this PR
originally added already existed there. The author reviewed all changes and made the judgment calls
on each finding. No session link — the work was done in the Claude Code VS Code extension.

## Risk Level

Medium — it changes what Find's project picker offers and adds scroll behaviour to panels that
previously had none, both on Simple mode's common path. Mitigated by the decision logic being
extracted as unit-tested predicates, and by the editor path being unchanged when an editor is open.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2749)

### DIFF
--- a/.context/standards/Architecture-Decisions.md
+++ b/.context/standards/Architecture-Decisions.md
@@ -3217,3 +3217,45 @@ step, no automation. Just a record.
   `^\d+_web-view-state$` obsolete-key sweep are all untouched: they read numeric-era data no future
   build can add to, and durability changes nothing about them.
 - **Source:** PT-4464.
+## adr-find-searchable-tabs: Find searches what a tab declares it displays, and targets editors and reference panels differently
+
+- **Date:** 2026-09-03
+- **Status:** Accepted
+- **Context:** Find's project picker listed only projects open in a Scripture editor tab, so a
+  resource shown in a read-only reference panel (Model text, Bible texts, Commentaries) could not be
+  selected. Worse, Ctrl+F from such a panel passed that resource's project id to Find, whose
+  reassignment effect then found no matching tab, took its cross-project fallback, and silently
+  re-ran the search against an unrelated project (PT-4467). The panels already declare the resource
+  they display under `NAVIGABLE_PROJECT_IDS_WEB_VIEW_STATE_KEY` (see
+  [adr-navigable-project-ids](#adr-navigable-project-ids)), but nothing on the consuming side read
+  it.
+- **Decision:** `useOpenProjectTabs` gained an opt-in `includeNavigableProjectIds`, under which a
+  web view reports the projects it declares instead of its container project, yielding one tab per
+  declared project. Find opts in and widens its picker filter to
+  `FIND_SEARCHABLE_WEB_VIEW_TYPES`. Because a declaring view can contribute several tabs, tabs are
+  keyed by (web view, project) rather than by web view alone. A declaration of `[]` means "displays
+  nothing" and offers no project; an absent key means the view does not participate, so its
+  container project answers. Result activation then splits by tab kind: an editor is driven through
+  its web view controller (select + highlight), while a reference panel exposes no controller, so
+  activating its tab is the whole navigation. Wherever several tabs of one project qualify, an
+  editor wins.
+- **Alternatives:** **Pin the trigger-supplied project id so the reassignment effect cannot move
+  it** — rejected: it stops the snap-away but leaves the resource absent from the picker, so the
+  user still cannot see or choose what is being searched. **Mirror a Find-specific state key across
+  the two extensions** — implemented first and then removed: `navigableProjectIds` already carries
+  exactly this value from exactly these panels, with a readiness gate, so a second key meant both
+  panels publishing the same value twice under two keys. **Include the Text Collection** — deferred:
+  it declares every resource it hosts and would be nearly free to list, but "go to result" has no
+  defined behavior there, since each cell is its own editor and revealing the grid would not show
+  which cell matched.
+- **Consequences:** Any web view that declares navigable projects becomes searchable by adding its
+  type to `FIND_SEARCHABLE_WEB_VIEW_TYPES`, with no further wiring. The cost is that Find's picker
+  no longer means "projects open in an editor", so callers must not assume a picked row is a valid
+  Replace target — Replace stays gated on editability, which already covers it. Because a panel
+  registers no controller, "go to result" on a panel-only resource can only activate its tab: in
+  Simple mode the Bible texts and Commentaries panels share Find's Column 3 stack, so that
+  necessarily hides the results list, while the Model text panel is already visible in Column 1 and
+  the activation only moves keyboard focus. Find cannot tell those cases apart from inside its own
+  iframe, since PAPI exposes no visibility query for another web view.
+  **Revisit** if per-cell reveal is designed for the Text Collection, which would bring it into the
+  searchable set.
--- a/extensions/src/platform-scripture-editor/src/_editor-overrides.scss
+++ b/extensions/src/platform-scripture-editor/src/_editor-overrides.scss
@@ -314,3 +314,34 @@ button {
     border-bottom: 2px solid var(--foreground);
   }
 }
+
+// Verse-arrival highlight: a brief pulse on the verse a navigation landed on. Applied by adding
+// `.highlighted` to the verse element returned by `scrollToVerse`, and removed on cleanup.
+//
+// Lives here rather than in the Scripture editor's own stylesheet because the read-only reference
+// panels need it too: Find's "go to result" scrolls those panels to the matched verse, and without
+// the pulse the user lands on a verse with nothing indicating which one matched.
+//
+// Highlight keyframes thanks to chazsolo at https://stackoverflow.com/a/55835473
+@keyframes highlight {
+  from {
+    background-color: yellow;
+  }
+}
+
+.editor-container .highlighted {
+  position: relative;
+}
+
+.editor-container .highlighted::before {
+  content: '';
+  position: absolute;
+  width: 25px;
+  height: 25px;
+  border-radius: 50%;
+  top: 50%;
+  left: 50%;
+  transform: translate(-50%, -50%);
+  opacity: 0.8;
+  animation: highlight 2s;
+}
--- a/extensions/src/platform-scripture-editor/src/editor-dom.util.test.ts
+++ b/extensions/src/platform-scripture-editor/src/editor-dom.util.test.ts
@@ -17,6 +17,8 @@ import {
   BASELINE_PROBE_ATTRIBUTE,
   clampTopToVisibleArea,
   findScrollContainer,
+  hasNewScrollTarget,
+  isEchoOfPublishedScrRef,
   measureBaselineOffset,
   scrollToAnnotation,
   scrollToVerse,
@@ -492,5 +494,55 @@ describe('clampTopToVisibleArea', () => {
     // Both clamps are active and pull opposite ways. Order is the behavior being pinned: max-then-min
     // yields the bottom bound, whereas min-then-max would return the visible-area top instead.
     expect(clampTopToVisibleArea({ top: 0, bottom: 50 }, { top: 0 }, { top: 100 })).toBe(49);
+  });
+});
+
+const JHN_3_16 = { book: 'JHN', chapterNum: 3, verseNum: 16 };
+const JHN_3_17 = { book: 'JHN', chapterNum: 3, verseNum: 17 };
+const ROM_3_16 = { book: 'ROM', chapterNum: 3, verseNum: 16 };
+const USJ_A = { type: 'USJ', content: ['a'] };
+const USJ_B = { type: 'USJ', content: ['b'] };
+
+describe('isEchoOfPublishedScrRef', () => {
+  it('recognizes the reference this view just published', () => {
+    expect(isEchoOfPublishedScrRef(JHN_3_16, JHN_3_16)).toBe(true);
+  });
+
+  it('does not treat a different verse as an echo', () => {
+    // A real navigation to a neighbouring verse must still scroll.
+    expect(isEchoOfPublishedScrRef(JHN_3_16, JHN_3_17)).toBe(false);
+  });
+
+  it('compares the book too, not just the numbers', () => {
+    expect(isEchoOfPublishedScrRef(JHN_3_16, ROM_3_16)).toBe(false);
+  });
+
+  it('is not an echo when nothing is outstanding', () => {
+    expect(isEchoOfPublishedScrRef(undefined, JHN_3_16)).toBe(false);
+  });
+});
+
+describe('hasNewScrollTarget', () => {
+  it('scrolls when nothing has been scrolled to yet', () => {
+    expect(hasNewScrollTarget(undefined, JHN_3_16, USJ_A)).toBe(true);
+  });
+
+  it('skips a bare reveal, so a manual scroll survives a tab switch', () => {
+    expect(hasNewScrollTarget({ scrRef: JHN_3_16, usj: USJ_A }, JHN_3_16, USJ_A)).toBe(false);
+  });
+
+  it('scrolls when the reference moved while hidden', () => {
+    expect(hasNewScrollTarget({ scrRef: JHN_3_16, usj: USJ_A }, JHN_3_17, USJ_A)).toBe(true);
+  });
+
+  it('scrolls when the chapter content arrives for the same reference', () => {
+    // A reveal can beat the chapter load; the content landing is the cue to scroll, and the
+    // reference alone would not distinguish it.
+    expect(hasNewScrollTarget({ scrRef: JHN_3_16, usj: USJ_A }, JHN_3_16, USJ_B)).toBe(true);
+  });
+
+  it('compares content by identity, not value', () => {
+    // The panel pushes whatever object the PDP hands it; an equal-but-new object is a real update.
+    expect(hasNewScrollTarget({ scrRef: JHN_3_16, usj: USJ_A }, JHN_3_16, { ...USJ_A })).toBe(true);
   });
 });
--- a/extensions/src/platform-scripture-editor/src/editor-dom.util.ts
+++ b/extensions/src/platform-scripture-editor/src/editor-dom.util.ts
@@ -317,3 +317,63 @@ export function scrollToAnnotation(id: string): HTMLElement | undefined {
 
   return annotationElement;
 }
+
+/**
+ * Whether an incoming reference is this view's own echo — the reference it just published coming
+ * back through its scroll group.
+ *
+ * A read-only reference panel sits on scroll group 0, so a verse click inside it publishes to the
+ * group and returns immediately as a prop update. Scrolling for that echo would drag the user's own
+ * click target to the top of the viewport right after they clicked it.
+ *
+ * @param lastPublishedScrRef The reference this view last published, or `undefined` if none is
+ *   outstanding.
+ * @param scrRef The incoming reference.
+ * @returns `true` when the incoming reference is the outstanding echo and no scroll should happen.
+ */
+export function isEchoOfPublishedScrRef(
+  lastPublishedScrRef: SerializedVerseRef | undefined,
+  scrRef: SerializedVerseRef,
+): boolean {
+  return (
+    !!lastPublishedScrRef &&
+    lastPublishedScrRef.book === scrRef.book &&
+    lastPublishedScrRef.chapterNum === scrRef.chapterNum &&
+    lastPublishedScrRef.verseNum === scrRef.verseNum
+  );
+}
+
+/**
+ * Whether there is anything new to scroll to since the last scroll this view performed.
+ *
+ * Guards the bare reveal: a panel that shares a tab stack with other views is re-shown constantly,
+ * and re-scrolling every time would discard a scroll position the user set by hand before switching
+ * tabs. The chapter content is part of the identity because a reveal can beat the chapter load —
+ * when content arrives for the same reference, that IS new and does need a scroll.
+ *
+ * @param lastScrolledFor What the last performed scroll was for, or `undefined` if none yet.
+ * @param scrRef The reference to scroll to.
+ * @param usj The chapter content currently loaded, compared by identity.
+ * @returns `true` when the reference or the content differs from the last scroll.
+ */
+export function hasNewScrollTarget(
+  lastScrolledFor: { scrRef: SerializedVerseRef; usj: unknown } | undefined,
+  scrRef: SerializedVerseRef,
+  usj: unknown,
+): boolean {
+  if (!lastScrolledFor) return true;
+  return (
+    lastScrolledFor.usj !== usj ||
+    lastScrolledFor.scrRef.book !== scrRef.book ||
+    lastScrolledFor.scrRef.chapterNum !== scrRef.chapterNum ||
+    lastScrolledFor.scrRef.verseNum !== scrRef.verseNum
+  );
+}
+
+/**
+ * Max ms to wait for a verse scroll to become possible before giving up — the rAF retry in the
+ * model text panel and the settle loop in the resource text panel both bound themselves with it.
+ * The usual reason for reaching it is a verse marker genuinely absent from the USJ (a `\v 16-17`
+ * range publishes no marker for 17).
+ */
+export const SCROLL_MAX_WAIT_MS = 2000;
--- a/extensions/src/platform-scripture-editor/src/model-text-panel.component.tsx
+++ b/extensions/src/platform-scripture-editor/src/model-text-panel.component.tsx
@@ -38,7 +38,7 @@ import { RetryableErrorView, LoadingView } from './panel-state-views.component';
 import { getResourcePanelReadiness } from './resource-panel-readiness.utils';
 import { PanelReadinessView } from './panel-readiness-view.component';
 import type { EffectiveResourceReferenceListState } from './use-effective-resource-reference-list.hook';
-import { scrollToVerse } from './editor-dom.util';
+import { SCROLL_MAX_WAIT_MS, scrollToVerse } from './editor-dom.util';
 import { ResourceBookNotAvailable } from './resource-book-not-available.component';
 import { ResourceBlankChapter } from './resource-blank-chapter.component';
 import { ResourceTextUnavailable } from './resource-text-unavailable.component';
@@ -66,9 +66,6 @@ export const MODEL_TEXT_EDITOR_CONTAINER_TEST_ID = 'model-text-editor-container'
 // and the editor reloads (re-serialize + setEditorState) whenever the `view` object's identity
 // changes — so a refetch of identical content would force a pointless full reload.
 const VIEW_OPTIONS = getDefaultViewOptions();
-
-/** Max ms to retry scrolling via rAF before giving up (e.g. verse marker missing from USJ) */
-const SCROLL_MAX_WAIT_MS = 2000;
 
 export {
   MODEL_TEXT_PANEL_STRING_KEYS,
--- a/extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.scss
+++ b/extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.scss
@@ -16,30 +16,6 @@ mark {
 
 // #region verse number highlight
 
-// Highlight keyframes thanks to chazsolo at https://stackoverflow.com/a/55835473
-@keyframes highlight {
-  from {
-    background-color: yellow;
-  }
-}
-
-.editor-container .highlighted {
-  position: relative;
-}
-
-.editor-container .highlighted::before {
-  content: '';
-  position: absolute;
-  width: 25px;
-  height: 25px;
-  border-radius: 50%;
-  top: 50%;
-  left: 50%;
-  transform: translate(-50%, -50%);
-  opacity: 0.8;
-  animation: highlight 2s;
-}
-
 // #endregion
 
 // Simple-mode-only rules (including .scripture-editor-tab-nav-simple, used by this web view's
--- a/extensions/src/platform-scripture-editor/src/resource-text-panel.web-view.tsx
+++ b/extensions/src/platform-scripture-editor/src/resource-text-panel.web-view.tsx
@@ -21,6 +21,7 @@ import {
   DropdownMenuTrigger,
   useExtraValidMarkers,
   useTabIconSelection,
+  useViewVisibility,
   type TabIconUrls,
   Spinner,
 } from 'platform-bible-react';
@@ -32,10 +33,16 @@ import {
   LocalizeKey,
   ResourceType,
 } from 'platform-bible-utils';
-import { Canon } from '@sillsdev/scripture';
+import { Canon, SerializedVerseRef } from '@sillsdev/scripture';
 import { ChevronDown } from 'lucide-react';
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import type { ResourceReferenceList } from 'platform-scripture';
+import {
+  hasNewScrollTarget,
+  isEchoOfPublishedScrRef,
+  SCROLL_MAX_WAIT_MS,
+  scrollToVerse,
+} from './editor-dom.util';
 import { useOpenFindShortcut } from './use-open-find-shortcut.hook';
 import { useEffectiveResourceReferenceList } from './use-effective-resource-reference-list.hook';
 import { useResourcePickerResources } from './use-resource-picker-resources.hook';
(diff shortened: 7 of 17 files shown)


katherinejensen00 (changes requested):
Levi — the diagnosis is right and the shape of the fix is right: the panels genuinely were invisible to Find's picker, and publishing the displayed resource into web view state is the correct answer. But main moved under this branch. It already publishes exactly this value, from these exact two files, under a shared key in platform-bible-utils — with a readiness gate that this branch's version doesn't have, and whose absence is reachable on the common path. That's the thing to settle before the rest of the review matters, because it decides how much of the new code survives. The scroll effect also has several independent defects that make "go to result" land on the wrong verse or silently do nothing.

The branch is 16 commits behind m resolve differently after it.

Blocking

1. This publish mechanism already exists on main, and this branch re-invents it under a second, hand-mirrored key.
📍 extensions/src/platform-scripture-editor/src/use-publish-focused-resource.hook.ts:24 · Critical

origin/main publishes the displayed resource's project id from both panels already: resource-text-panel.web-view.tsx:381-385 and model-text-panel.web-view.tsx:127platform-bible-utils, which platform-scripture already imports (use-open-project-tabs.ts:3) — so ADR-0029's central premise, "extensions cannot import each other's source, so the key is mirrored and pinned on both sides", has an existing answer. That makes focused-resource-state-key.const.ts, the mirrored key in resource-panel-web-view-types.const.ts, both pinning tests, and much of the ADR deletable; useOpenProjectTabs would read navigableProjectIds guarded by isNavigableProjectIds. After a rebase, as written, both panels publish the same value twice under two keys via two hooks justified by two ADRs.

2. The new hook has no readiness gate, so a transient unresolved resource destroys Find's search.persisted declaration." Your hook publishes whatever it's handed. Concretely: Find is searching RES-1 with hundreds of results; the user picks a different resource in the panel; for one render resourceProjectId is undefined, so setPublishedProjectId(undefined) fires and useOpenProjectTabs drops the tab (Simple-mode panels have no container projectId, so passes is false). Find's reassignment effect (find.web-view.tsx:889) finds no RES-1 tab, takes the cross-project fallback, and handleSelectProjectScrollGroup runs abandonFindJob(); setResults([]); …. The search is gone and Find lands on an unrelated project a tick before RES-2 resolves. Same on every relaunch, while the reference list and DBL catalog resolve.

3. The published key is persisted into saved layouts and the clear path never removes it.
📍 use-publish-focused-resource.hook.ts:29 · Major
is true only on a first-ever mount, and use-open-project-tabs.test.ts's clear test feeds state: {}, a shape the publisher never produces. Worse: uninstall a resource and restart, and useOpenProjectTabs seeds from getAllOpenWebViewDefinitions() before any panel mounts, yielding a tab for a resource that no longer exists. refetchAttemptedProjectIdsRef retries the metadata fetch exactly once, so the "selected but not listed" gap the code calls transient becomes permanent, with Find stuck on "the selected project is not available to search".

4. The scroll effect has no USJ loading gate, so it scrolls to and highlights the wrong verse in the wrong book.
📍 extensions/src/platform-scripture-editor/src/resource-text-panel.web-view.tsx:596 · Major

const [usjPossiblyError] = useProjectData(...) (line 455) discards isLoading, and useProjectData holds the previous chapter's USJ across a selector change. Panel shows JHN 3; a result activates it for ROM 3:16. The book/chapter deps fire while usjFromPdp is still JHN 3 — fully laid out, so scrollHeight matches on two consecutive frames and isSettled is true on frame 2. scrollToVerse matches on verse number only, with no book/chapter qualifier (editor-dom.util.ts:222), finds JHN 3:16, smooth-scrolls there and pulses it. model-text-panel.component.tsx:271 guards this exact case with if (!usj || isUsjLoading) return undefined;.

5. The echo latch can stay armed and then swallow a genuine "go to result".
📍 resource-text-panel.web-view.tsx:603 · Major

lastPublishedScrRefRef is cleared in one commit. A later "go to result" onto that verse then matches the stale latch and returns with no scroll and no pulse — and since a panel registers no controller, nothing retries. The Scripture editor clears its equivalent ref on both paths (platform-scripture-editor.web-view.tsx:3336 and :3355); this copy drops the second clear.

6. A detached panel is indistinguishable from a group-0 one, so "go to result" reveals a tab on an unrelreachable from the Power-mode top toolbar's ScrollGroupSelector, which admits the panel because it has a definition projectId. Detaching writes a SerializedVerseRef object, which use-open-project-tabs.ts:110-118 maps to group 0. Find joins group 0, writes there, the detached panel drops it, and setFocus fronts a panel still on GEN 1 for a JHN 3:16 result. No error, no log. Adding a group match here would not fix it — the hook fabricates group 0, so the fix belongs in useOpenProjectTabs (stop fabricating, or carry a detached flag). The comment at find.web-view.tsx:1763 is false in this state.

7. The new "editor wins over a panel" preference is bypassed by two of the three branches.
📍 find.utils.ts:408 · Major
…) finds no editor, so every result click silently drops selectRange + setAnnotation in an editor plainly visible showing P. The cross-project sort orders by scrollGroupId then project id with no type consideration, and panels are pinned to group 0 by the new rescue at use-open-project-tabs.ts:129 — so a panel systematically sorts ahead of an editor in a higher group. Your new test passes currentScrollGroupId: 2, a group with no tab, so it exercises only the fallback branch.
a project no tab is showing.
📍 extensions/src/platform-scripture/src/hooks/use-open-project-tabs.ts:92 · Moderate

focusedResourceProjectId ?? containerProjectId now runs for panels too. A Power-mode Bible texts panel opened from project P, whose resource is still installing, reports as a tab of P — reintroducing exactly what ADR-0029 rejects as "a no-op that

lastScrolledForRef.current = {…} is stamped before scrollToVerse and regardless of its return. When the marker is genuinely absent (a \v 16-17 range means no [data-number="17"]) or the pane hasn't produced overflow, nothing scrolls — but hasNewScrollTarget returns false forever after. model-text-panel.component.tsx:279-284 retries until found for this reason. The comment at 650-651 also claims the timeout is "the exit for a verse marker genuinely absent fro
📍 resource-text-panel.web-view.tsx:605 · Moderate

Panel scrolls to 3:16 and records it. User clicks verse 25 → echo → early return, record still {3:16}. They hand-scroll to verse 40, switch to Find, switch back: hasNewScrollTarget({3:16}, 3:25) is true, so the panel snaps to verse 25 and pulses it. The comment at lines 186-191 says this case must not re-scroll.

11. The verseNum <= 1 fast path is wrong about its callee.
📍 resource-text-panel.web-view.tsx:631 · Moderate

scrollToVerse treats only verseNum < 1 as chapter-top (editor-dom.util.ts:219). Verse 1 queries a real marker and measures real geometry — synchronously, on a just-revealed pane, bypassing the settle loop the rest of the effect exists for. That's the "scrolls to a fraction of the real target" trap your own comment describes. lastScrolledForRef is stamped first, so the bad positio.tsx:243 · Moderate

Simple mode: it's in Column 1's HEADLESS_GROUP (simple-layout.data.ts:34-41), not Column 3 — so the comment at find.web-view.tsx:1760-1763 ("that panel shares Find's tab stack, so this necessarily hides Find; that is the point of this gesture") is false here. setFocus fires on an already-visible panel, pulls keyboard focus off the result list, and reveals nothing.display:none; its scroll effect has no useViewVisibility and no catch-up. Your comment says "this panel occupies Column 1 of the Simple layout alone and is therefore always visible, so the hidden case is unreachable" — true in Simple mode, not in Power.

13. The default-group rescue is much broader than its comment and tests.
📍 use-open-project-tabs.ts:129 · Moderate

The preceding chain leaves scrollGroup undefined for any value that isn't undefined, a number, or a non-null object. The rescue is scoped on the wrong axis — it drops the defensive rejection for null, strings, booleans and everything else on any tab publishing a focused resource. A panel with a migrated scrollGroupScrRef: '0' now surfaces at a fabricated group 0 and never follows Find's writes. Both new tests exercise only null. Separately (Minor): the claim that these panels carry null "exactly … on the wire" doesn't hold — simple-layout.data.ts:106-127 sets no scrollGroupScrRef and createResourceTextPanelProvider never writes one, so the value is undefined, already handled by the pre-existing branch. Please cite the null producer or drop the branch.

14. The Text Collection exclusion rests on a false premise, leaving the same bug one view over.
📍 extensions/src/platform-scripture/src/resource-panel-web-view-types.const.ts:51 · Moderate

The comment says the grid "shows several scriptures at once, so there is no single project for the picker to offer" — but scripture-text-grid.web-view.tsx:269-270 already resolves exactly one: useFocusedResourceProjectId(displayedProjectIds), the resource holding the caret, which it passes to useOpethis PR fixes for the panels.

15. The hidden-case catch-up and the result-activation branch are both untested.
📍 resource-text-panel.web-view.tsx:595 · Moderate

.claude/rules/cross-view-sync-hidden-views.md is explicit: "mount the sync while hidden, trigger the change, flip visibility, and assert the pending catch-up fires. Documented intent doesn't survive a refactor; a test does." You test hasNewScrollTarget and isEchoOfPublishedScrRef as pure predicates; nothing exercises the visibility flip, the settle loop, the echo reset, or the highlight cle— and would cover the window where an editor exists but its controller hasn't resolved, in which neither branch fires and nothing is activated.

Worth tightening

16. The ADR needs re-slugging on rebase, and two comments assert things the code doesn't do.
📍 .context/standards/Architecture-Decisions.md:1433 · Minor

## ADR-0029: was correct against your base, but main has since converted the log to adr-{kebab-slug} ("Identify entries by an adr--prefixed kebab-case slug, not a number"; "Don't add **Formerly:** to new entries"). The reference at focused-resource-state-key.const.test.ts:10 needs updating too. Also: use-open-pro

17. The documentation is accurate and forward-facing, but there's too much of it to be read.
📍 resource-panel-web-view-types.const.ts:1 · Minor

337 of 1,170 added lines are comments (29%), concentrated at 59% in find.utils.ts's new functions, 51% in the scroll effect, and 61% in resource-panel-web-view-types.const.ts (33 comment lines for 6 constants). The "definition projectId is the container project" explanation appears seven times in prose; the "extensions can't import each other, so the key is mirrored" story appears fourteen times. Every copy is a place the explanation can drift — and findings 13, 14 and 16 are all cases where a copy already has. Keep the full explanation once and make the rest one-line pointers. resolveTargetReferencePanelWebViewId carries a 28-line TSDoc over an 11-line body, about half of it restating the ADR's Decision and Alternatives; let the ADR own the rationale and the TSDoc own the contract.

Two smaller notes. OpenScrollGroupTab.webViewType was made optional so tests and the story can build tabs without one (find.utils.ts:271) — production never produces undefined there, and loosening a production type to suit test construction hides the case where the editor preference silently no-ops; a test factory with a default is the cleaner fix. And the three panel web-view-type literals are pinned on the consumer side only — precisely the gap ADR-0029 boasts about closing for the state key, copied three times in the same new file; a rename in main.ts:55-57 keeps every test green and drops the panels out of Find.
SavedWebViewDefinition; setFocus({focusType:'webView'}) genuinely activates the tab (window.service-shard.ts:480); the ~20s controller-hang warning is accurate (web-view-provider.service.ts:175). The SCSS move is correctly wired — all four entry points already @use './editor-overrides', with no colliding @keyframes. The fallbackKey deprecation follows Localization-Guide.md exactly, and the retired keys had no non-English translations to orphan. isSerializable(undefined) is true, so clearing never throws. Replace stays gated, the picker de-dupes by scroll group, and the useOpenProjectTabs tests are properly isolated. Extracting the two guards as unit-tested predicates was the right instinct — the gap is that the wiring around them went untested, which is where most of the above lives. with no colliding @keyframes. The fallbackKey deprecation follows Localization-Guide.md exactly, and the retired keys had no non-English translations to orphan. isSerializable(undefined) is true, so clearing never throws. Replace stays gated, the picker de-dupes by scroll group, and the useOpenProjectTabs tests are properly isolated. Extracting the two guards as unit-tested predicates was the right instinct — the gap is that the wiring around them went untested, which is where most of the above lives.

> @katherinejensen00 reviewed 22 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2749#-:-P0UKuCZ3kWJTHMQnz0a:b-tlgsa0)* status: all files reviewed, 1 unresolved discussion (waiting on captaincrazybro).

___
*[`extensions/src/platform-scripture-editor/src/use-publish-focused-resource.hook.ts` line 24 at r1](https://reviewable.io/reviews/paranext/paranext-core/2749#-P0UL-Fo9QKl7SLi-kBN:-P0UL-Fp9L84H4D2HlX4:b-egxe1v) ([raw file](https://github.com/paranext/paranext-core/blob/484667039a2a893779417e250dd3878c59ab36b4/extensions/src/platform-scripture-editor/src/use-publish-focused-resource.hook.ts#L24)):*
> ```XML
>  *   stale id is never left behind.
>  */
> export function usePublishFocusedResourceProjectId(
> ```

1. This publish mechanism already exists on main, and this branch re-invents it under a second, hand-mirrored key.
📍 extensions/src/platform-scripture-editor/src/use-publish-focused-resource.hook.ts:24 · Critical

origin/main publishes the displayed resource's project id from both panels already: resource-text-panel.web-view.tsx:381-385 and model-text-panel.web-view.tsx:127platform-bible-utils, which platform-scripture already imports (use-open-project-tabs.ts:3) — so ADR-0029's central premise, "extensions cannot import each other's source, so the key is mirrored and pinned on both sides", has an existing answer. That makes focused-resource-state-key.const.ts, the mirrored key in resource-panel-web-view-types.const.ts, both pinning tests, and much of the ADR deletable; useOpenProjectTabs would read navigableProjectIds guarded by isNavigableProjectIds. After a rebase, as written, both panels publish the same value twice under two keys via two hooks justified by two ADRs.

captaincrazybro:
Thanks — finding 1 was right, and following it changed most of the rest. The branch has been rebuilt on top of `main` (it was 23 commits behind, not 16) and force-pushed, so Reviewable will show a fresh revision. The diff is now 17 files / +941, down from 22 / +1170.

**What the rebase removed rather than fixed**

`use-publish-focused-resource.hook.ts`, `focused-resource-state-key.const.ts`, the mirrored key in `resource-panel-web-view-types.const.ts`, both pinning tests, and ADR-0029 are all gone. `useOpenProjectTabs` now reads `navigableProjectIds` guarded by `isNavigableProjectIds`, exactly as you suggested. Findings **2** and **3** went with it — `usePublishNavigableProjectIds`' readiness gate and `resolveNavigableProjectIdsWrite` already own both cases.

Two things I'd flag as no longer mine, rather than silently fixing:

- **Finding 13** — the scroll-group rescue you flagged is pre-existing on `main` (`use-open-project-tabs.ts:71-80`, mapping a detached panel's `SerializedVerseRef` to group 0). My branch added a *second* one; the rebuild simply doesn't. The `null`-on-the-wire claim was mine and is gone with it.
- **Finding 6** — a detached panel being indistinguishable from a group-0 one follows from that same pre-existing branch. This PR makes it *reachable* from Find rather than creating it. Happy to take it on if you'd rather it be fixed here, but it looked like it wants its own change.

**Fixed as described**

4, 5, 9, 10, 11 — all confirmed against the code before changing anything. Notably `main` already destructures `isUsjLoading` (`resource-text-panel.web-view.tsx:442`), so finding 4's gate came free with the rebase. The echo latch is now consumed before the visibility check, `lastScrolledForRef` records only a scroll that actually landed, the echo path records the clicked verse so a later bare reveal doesn't snap away from it, and the chapter-top fast path is `< 1` to match `scrollToVerse`.

**Finding 7** — the editor preference now applies in all three branches, including when `preferredWebViewId` names the panel. Five tests cover it; disabling the preference fails three of them.

**Finding 12** — you were right that the Column 3 claim is false for the Model text panel. The comment now states both cases and says Find cannot tell them apart from inside its own iframe, since PAPI exposes no visibility query for another web view.

**Finding 14** — also right: the grid resolves one project via `useFocusedResourceProjectId` and already hands it to Ctrl+F, so the "no single project to offer" reason was false. The stated reason is now the real one — "go to result" has no defined behavior when each cell is its own editor. I'm filing the Ctrl+F-from-Text-Collection gap as a follow-up rather than growing this PR.

**Findings 16 and 17** — new entry is `adr-find-searchable-tabs`, kebab-slugged with no `Formerly:`. `resolveTargetReferencePanelWebViewId`'s TSDoc is down to its contract, with the rationale left to the ADR. `OpenScrollGroupTab` keeps `webViewType` instead of being loosened for tests; the two test factories default it, which is the fix you suggested.

**One process note worth recording:** none of this could be tested in a git worktree until `platform-bible-utils` was built there — `platform-bible-utils/experimental` resolves through `dist/`, so *any* test importing it fails. `use-publish-navigable-project-ids.hook.test.ts` fails 8/10 in a fresh worktree on untouched `main` for that reason alone.

Verification: ESLint and Stylelint clean, 1897 tests across 111 files.

The "Needs UX sign-off" callout in the description still stands and is unaffected by the rebuild.

katherinejensen00 (approved):
Thanks for making these changes, Levi! Nothing is blocking so I am approving and leaving it up to you if any changes are needed. Feel free to ping me if you need another review.

The rebase was the right call and it did the heavy lifting. Findings 1, 2 and 3 are resolved by deletion rather than patching, and I verified there's no remnant: no focused-resource-state-key.const.ts, no usePublishFocusedResource, no second mirrored key, no ADR-0029 anywhere in the tree. Finding 13 is withdrawn — git show 30f07f460b1 confirms the object→group-0 mapping is byte-identical to the base, so your branch adds no second rescue and the null-on-the-wire claim went with it. You were right and I was wrong on that one. Findings 4, 10, 11, 16 and the OpenScrollGroupTab note all check out as described, the ADR is correctly slugged and templated, and the commit messages are the best part of this branch.

Four things are still open

1. The shared key now has two incompatible readings, and this PR introduces the second.
📍 extensions/src/platform-scripture/src/hooks/use-open-project-tabs.ts:136 · Critical

NAVIGABLE_PROJECT_IDS_WEB_VIEW_STATE_KEY documents itself as the ids a view displays "beyond its own projectId", and says plainly "readers just union the lists" (lib/platform-bible-utils/ the same window. It's latent today because the panels publish only the resource — but a publisher tuned for the toolbar consumer (dropping a redundant self-id, or writing [] for a view that still shows its container) would be correct under the stated contract and would silently drop that tab out of Find's picker. The replace reading may well be right for Find; the problem is it's encoded only in this consumer and the new ADR while the shared const still documents the union. Settle it at the key, or union here and filter by web view type.

2. Finding 5 is still open — the latch is cleared on one path.
📍 extensions/src/platform-scripture-editor/src/resource-text-panel.web-view.tsx:660 · Major

Your reply says the echo latch is now consumed before the visibility check, and it is — but that was the ordering half. The substance of finding 5 was the missing second clear, and lastPublishedScrRefRef.current = undefined still appears only in the echo branch. The editor clears at both platform-scripture-editor.web-view.tsx:3336 and :3356; I re-checked. Concretely: the panel publishes verse 25, and before that write round-trips a Find result writes 3:40. scrRef goes 3:16 → 3:40 and never equals 3:25, so the effect proceeds and scrolls, and the latch keeps 3:25 forever. A later genuine "go to result" onto 3:25 matches the stale latch, returns with no scroll and
3. Finding 9 is half fixed; two paths still silently do nothing.
📍 resource-text-panel.web-view.tsx:699 and :715 · Major

Recording only a landed scroll on the settled path is right. But the settle loop makes exactly one attempt: isSettled is true as soon as two frames agree, which an empty flex-sized editor container satisfies immediately, before Lexical has reconciled the chapter. scrollToVerse returns undefined, the loop returns, and nothing re-triggers the effect — isViewVisible, usjFromPdp, isUsjLoading and the three scrRef fields are all unchanged. It doesn't wedge, but that reveal never scrolls. model-text-panel.componentstructurally can't tell whether it landed — and if .editor-container isn't in the DOM yet, nothing scrolls but it's recorded, which is finding 9's wedge again. Verse 0 is a real target in this panel; your own comment at 444-448 says so.

4. The Model text panel is a "go to result" target with no hidden-case handling.
📍 extensions/src/platform-scripture-editor/src/model-text-panel.component.tsx:333 · Major

MODEL_TEXT_PANEL_WEBVIEW_TYPE is in REFERENCE_PANEL_WEB_VIEW_TYPES, so resolveTargetReferencePanelWebViewId can pick it and handleOpenAtResult will front it. But that patested and one not.

5. Finding 15 is still open.
📍 resource-text-panel.web-view.tsx:740 · Major

There is still no test file for this panel at all. The two predicates are tested exhaustively as pure functions, and the ~90-line effect that composes them — the visibility flip, the echo consume-and-rearm, the isUsjLoading gate, the settle loop, the record-on-landing rule, the highlight cleanup — has zero coverage. That's every behaviour round 1 found broken, including all five you just fixed; each fix is currently asserted only by the comment above it, and the predicate tests would still pass with the effect's ordering inverted. cross-view-sync-hidden-views.md is explicit that documented intent doesn't survive a refactor. Two precedents make this cheap: character-marker-bar-overlay.component.test.tsx:50 mocks useViewVisibility through a mutable object, and model-text-panel.component.test.tsx:26 already mounts the sibling panel with the editor and platform-bible-react mocked.

6. The description still describes revision 1.

Four claims are wrong against this diff: no panel publishing is added here (all three publishers pre-exist on your base — this PR is purely the consuming side); usePublishFocusedResourceProjectId and its key file don't exist; the flag is includeNavigableProjectIds and it reports what a view declares, not one resourcegetFocusedResourceProjectId guards, neither of which survives.

The part I'd most want fixed is the "Needs UX sign-off" section, since you said it still stands. Item 2 no longer describes this diff — the new sync is in resource-text-panel.web-view.tsx and it does handle the hidden case; there's no deliberate no-op left to sign off. Item 1 repeats the blanket Column 3 claim you agreed was wrong in finding 12 and corrected in both the code and the ADR. UX can't rule on a description that states one outcome where the code now states two.
231 of 473 — 48.8%. resource-panel-web-view-types.const.ts is 69.8%, worse than round 1's 61%; find.utils.ts 57.9%; editor-dom.util.ts is a new hot spot at 56.1%, three ten-line TSDocs over three six-line predicates. The rebase cut the diff by deleting code, not prose. And the predicted failure mode has already happened three times in this revision: find.web-view.tsx:412 still says "editor tabs only" directly above the set that admits the panels, :427 still says the picker "only ever lists projects that are open in an editor tab", and resource-panel-web-view-types.const.test.ts:34 still asserts the Text Collection reason you agreed was false — contradicted by the const's own TSDoc thirty lines away. r2, and the ADR states the preference unconditionally at line 2880. scriptureEditorWebViewType is optional on two resolvers, which is the same fail-open shape you correctly removed from OpenScrollGroupTab; the sibling resolver takes it required. The three panel literals are pinned consumer-side only, so a rename in main.ts:58-60 keeps every test green and drops the panels out of Find — SCRIPTURE_EDITOR_WEBVIEW_TYPE is shared from web-view.model.ts:316 and shows the better mechanism. The eslint-disable at :737 justifies itself by module imports, but the rule is complaining about whole-object scrRef. Both new resolvers have zero tests. Neither deferral (Text Collection, detached panel) has a ticket, and resolveTargetReferencePanelWebViewId's doc uses "the panel is already following the reference through its scroll group" as its correctness argument — which is exactly what a detached panel doesn't do. On a11y, the pulse is the entire arrival signal on the panel path, with no announcement and no focus move, though this file already has announcementKey precedent at :837.

Security is clean. navigableProjectIds is guarded by a real predicate rather than cast, a malformed value degrades to "no declaration" (proven at use-open-project-tabs.test.ts:137), the read mechanism is sound, and the re-keying to (webViewId, projectId) correctly drops stale entries when a declared set shrinks. CI is green on all three platforms.

--- fail-open shape you correctly removed from OpenScrollGroupTab; the sibling resolver takes it required. The three panel literals are pinned consumer-side only, so a rename in main.ts:58-60 keeps every test green and drops the panels out of Find — SCRIPTURE_EDITOR_WEBVIEW_TYPE is shared from web-view.model.ts:316 and shows the better mechanism. The eslint-disable at :737 justifies itself by module imports, but the rule is complaining about whole-object scrRef. Both new resolvers have zero tests. Neither deferral (Text Collection, detached panel) has a ticket, and resolveTargetReferencePanelWebViewId's doc uses "the panel is already following the reference through its scroll group" as its correctness argument — which is exactly what a detached panel doesn't do. On a11y, the pulse is the entire arrival signal on the panel path, with no announcement and no focus move, though this file already has announcementKey precedent at :837.

Security is clean. navigableProjectIds is guarded by a real predicate rather than cast, a malformed value degrades to "no declaration" (proven at use-open-project-tabs.test.ts:137), the read mechanism is sound, and the re-keying to (webViewId, projectId) correctly drops stale entries whenrather than cast, a malformed value degrades to "no declaration" (proven at use-open-project-tabs.test.ts:137), the read mechanism is sound, and the re-keying to (webViewId, projectId) correctly drops stale entries when a declared set shrinks. CI is green on all three platforms.at use-open-project-tabs.t

> @katherinejensen00 reviewed 20 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2749#-:-P0enKGBBzqPzNXAzg4Z:b-usuiwu)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

captaincrazybro:
Thanks for the second pass, and for withdrawing 13 explicitly — that's useful. Pushed as one follow-up commit on a branch rebased to current `main`.

**Fixed**

**1 — the two readings of the shared key.** You're right, and the fix went the way you suggested. `useOpenProjectTabs` now unions the declared list with the container `projectId`, matching what the key documents and what `useOpenProjectBookIds` does. Each tab carries `projectSource`, and Find applies the narrowing itself with the reason stated at the call site: a reference panel's container project is never offered, because the panel isn't displaying that project's text. The replace rule is no longer encoded in a consumer contradicting the const.

**2 — the second latch clear.** Correct, and my reply last round answered the ordering half only. Any reference other than the one the latch was armed for now discards it, so a publish whose echo never arrives as its own update can't leave it armed. One note: the editor doesn't clear at `platform-scripture-editor.web-view.tsx:3336`/`:3356` — that file has no `lastPublishedScrRefRef` at all. The panel with the latch is `model-text-panel.component.tsx`, and it clears in exactly one place (`:258`), same as this one did. Your scenario stands on its own; the precedent doesn't.

**3 — the settle loop's single attempt.** Confirmed and fixed. Two agreeing heights are cheap to reach on an empty flex-sized container, so the loop stopped before Lexical had reconciled. It now keeps waiting until the scroll lands or the timeout, and gives up *without* recording so a later reveal retries. The chapter-top path checks for `.editor-container` before recording, since `scrollToVerse` can't report landing there.

**4 — Model text as a reveal target.** Removed from the reveal set rather than patched. `REVEALABLE_REFERENCE_PANEL_WEB_VIEW_TYPES` is now Bible texts + Commentaries; Model text stays fully searchable and listed, only the reveal is withheld. Costs nothing in Simple mode, where it's always visible and the scroll group has already moved it, and it avoids porting visibility handling into a panel this PR otherwise barely touches.

**6 — the description.** Rewritten. The four wrong claims are gone, and the UX section now describes this diff: one callout, scoped to Bible texts and Commentaries, since Model text is no longer fronted. The deliberate-no-op item is gone because the thing it described no longer exists.

**Also from your tail:** both target resolvers now have tests (12); `scriptureEditorWebViewType` is required on both scroll-group resolvers — same fail-open argument I accepted for `OpenScrollGroupTab`, with the tests using a defaulting wrapper rather than 18 edited call sites; the picker comments at `:412` and `:427` and the Text Collection test comment are corrected; and the `eslint-disable` now explains the real reason (the rule wants whole-object `scrRef`, and `useWebViewScrollGroupScrRef` returns a fresh object per publish).

**Not doing**

**5 — a test for the effect.** Declining deliberately, and flagging it rather than leaving it implicit. The three decision functions it composes are unit-tested; the composition isn't. I measured both routes: mounting `resource-text-panel.web-view.tsx` means mocking 23 local modules, 7 papi hooks and the editor to reach an effect whose entire input set is four values — a ~350-line harness that breaks on unrelated panel changes. The cheap route is extracting the effect to a hook, which we decided against. Judgment call that the sub-function coverage carries enough of it; happy to revisit if you disagree strongly.

**Two of your smaller notes I'd push back on**

- The **shared-const mechanism** for the three panel literals: `web-view.model.ts:316` is itself a mirror — its own comment says "Must match `SCRIPTURE_EDITOR_WEBVIEW_TYPE` in `platform-scripture-editor.utils.ts`" — and it lives in `src/shared`, which extensions can't import. So there's no shared source to point at today; making one would mean putting extension-owned web view type strings into `platform-bible-utils`.
- **The detached-panel deferral** now has a callout in the description rather than a ticket, since it's pre-existing `main` behaviour that this PR only makes reachable. Say the word and I'll file it. **PT-4507** covers the Text Collection deferral — filed before your review, so you likely didn't see it; it's now referenced from the const test and the description.

**a11y** is a fair hit: the pulse is the whole arrival signal on the panel path. Not addressed here — it needs a new localized string and an `announcementKey`, which felt like its own change.

ESLint and Stylelint clean, 1911 tests across 111 files.

PR #2239, opened 2026-05-01
feat(book-chapter-control): add hideVerse prop to omit verse from trigger label

MattGyverLee:
## Summary

- Add an optional `hideVerse?: boolean` prop to `BookChapterControl`. When `true`, the trigger button's label displays `<book> <chapter>` only (e.g. `Psalms 23`), omitting the verse part.
- Verse omission is locale-safe: `formatScrRefRange(refWithoutVerse, refWithoutVerse, ...)` with `verseNum: -1` lets the util drop the verse part. No regex on formatted output.
- Surface the chapter/verse and book/chapter separators through `localizedStrings`, so locales can render scripture references with their own typography conventions.
- Adds a `HideVerse` Storybook story.
- Locale data: `en.json` and `fr.json` populated; French refs render as `Jean 3.16` (period separator), and numeric-prefix French book names ("1 Samuel", "1 Jean", …) now use a non-breaking space between the digit and the noun so refs don't break across lines.

## Motivation

Chapter-level workflows have no concept of a starting verse. Examples:
- Whole-chapter / chapter-range exports (e.g. exporting `1 John 3` to FLEx for linguistic analysis).
- Scope selectors that operate on chapters.

Today these consumers must show `1 John 3:1`, which suggests verse-level granularity that doesn't exist in the workflow. There is no consumer-side way to fix this — `formatScrRef` always emits `book chapter:verse`, and the trigger renders that result directly with no override hook.

The localization changes address a related gap: even where the active locale would conventionally render scripture refs differently (French `3.16` vs. English `3:16`), the trigger label and the popover's internal references were hardcoded to use `:` between chapter and verse.

## Behavior

| Configuration | Trigger label example |
| --- | --- |
| `hideVerse` unset, English UI | ``1 John 3:1`` (current behavior, unchanged) |
| `hideVerse={true}`, English UI | ``1 John 3`` |
| `hideVerse` unset, French UI | ``1 Jean 3.1`` |
| `hideVerse={true}`, French UI | ``1 Jean 3`` |

## Localization

Two new optional localized-string keys, exported via `BOOK_CHAPTER_CONTROL_STRING_KEYS`:

| Key | Util default | English (`en.json`) | French (`fr.json`) |
| --- | --- | --- | --- |
| `%scripture_book_chapter_separator%` | `" "` | `" "` | `" "` |
| `%scripture_chapter_verse_separator%` | `":"` | `":"` | `"."` |

[...]

PR #2229, opened 2026-04-27
BCV: shadcn style fixes, buttons up/down + keyboarding fixes

Sebastian-ubs:
Changes (all ux approved)
- Fix styles after shadcn upgrade
  - chapter grid disaligned
  - input disaligned
  - tooltips
- Switch from left/right to up/down arrow buttons to better match what happens in the editor. Also making buttons a bit bigger to be able to better hit them and adding a toolbar separators.
- Fix keyboarding to
  - always have 1 clearly focused item, that submits with space + enter
  - up/down arrow key moves to the list / grid
  - tab and esc behaves as expected
  - typing anything else goes to the search input
  - "recent" dropdown is keyboard accessible
- Single match preview to show full + short book name

before
image
image

after (intermediate state)
image

new
Screenshot 2026-05-19 203309
Screenshot 2026-05-19 203402
Screenshot 2026-05-19 203325

---

<a href="https://app.devin.ai/review/paranext/paranext-core/pull/2229" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2229)

PR #2750, opened 2026-09-01
PT-4345: BCV keyboard navigation, styling, and the z-index scale

katherinejensen00:
## Summary

The Book/Chapter/Verse control's keyboard navigation is rebuilt: arrow keys move a highlight around the chapter/verse grid, the grid opens with the current chapter already highlighted, and Enter/Space activate whatever is highlighted. Styling moves to a ring-only keyboard highlight with RTL-safe logical properties. Separately, the z-index scale is repaired — tooltips were rendering *behind* popovers — and is now pinned by ordering tests so it cannot silently regress again.

Three decisions are recorded in `Architecture-Decisions.md` (`adr-bcv-item-value-contract`, `adr-z-index-ordering-invariants`, `adr-recent-searches-menu-semantics`). Those are the fastest way to get the "why" before reading the diff.

**One breaking change to flag:** `RecentSearches` (a public export) moved from Popover + `Command` to Radix `DropdownMenu`. Its items are now `role="menuitem"` inside a `menu`, not `role="option"` inside a `listbox`. Nothing in-repo depended on the old roles; an out-of-repo consumer querying `role="option"` would.

## Details

**Start here, in this order:**

| File | What it does |
| --- | --- |
| `book-chapter-control.utils.ts` | `GRID_COLUMNS` + `computeTargetGridItem` — the pure arrow-key arithmetic, including RTL mirroring and clamping. Read first; everything else assumes it. |
| `shared/book-item.utils.ts` | The cmdk item-value encoding that the grid and the activation path must agree on byte-for-byte (`adr-bcv-item-value-contract`). |
| `book-chapter-control.component.tsx` | The bulk of the diff. `handleCommandKeyDown` is the centre of gravity — a sequential dispatcher: Enter/Space → Enter-on-top-match → letters/digits → Backspace → arrows. |
| `numbered-item-grid.component.tsx` | Shared rendering for the chapter and verse grids: highlight, dimmed, disabled, selected. |
| `z-index.ts` + `z-index.test.tsx` | The scale and the ordering invariants that pin it. `_vars.scss` holds an SCSS twin that a test now keeps in agreement. |

**Why activation reads component state instead of cmdk's DOM.** Enter/Space resolve the target from `commandValue` rather than by reading `[cmdk-item][data-selected]` and synthesising a click. That keeps activation off library internals nothing of ours pins — and it is why the e2e helpers wait on the top-match row's *text* rather than on `data-selected`.

**z-index.** `Z_INDEX_TOOLTIP` moves 550 → 675 to clear `Z_INDEX_ABOVE_POPOVER` (650). The tests assert *order*, never specific numbers, so re-tiering stays cheap while reordering fails loudly.

**Keyboard shortcuts catalog** is updated per the repo rule, including one handler (`spaceSelectsHighlightedItem`) that predated this work but was never catalogued.

### Worth your scrutiny

- The quick-nav divider is now `ButtonGroupSeparator` (self-stretch) instead of a fixed `h-5` rule, matching `NavigationHistoryButtons` beside it. Structural change, but it does alter the divider's height — **not visually verified**.
- The back button is `tabIndex={-1}` by design (a tab stop there drops you out of the cmdk surface). Backspace is the keyboard route back, now tested at both levels.
- Verse chevrons point ▲▼ while ArrowUp/Down are intercepted to move to the book list. Deliberate and commented, but a reasonable thing to disagree with.
- In the books view, Enter submits `verseNum:1` while clicking the same cell drills into the verse grid. Left as-is — a product call, not a bug.

### Testing

179 tests across the six affected suites, plus revert-tested guards (neutering the `isTargetInteractive` guard, and neutering the Backspace branches, each turns the corresponding tests red). `typecheck:core` and `platform-bible-react` tsc clean; ESLint, Prettier and Stylelint clean; `dist/` and `papi.d.ts` rebuilt with no unexpected drift.

Full-suite `platform-bible-react` runs were not usable locally — a concurrent `vitest` run from another worktree saturated the machine, producing timeouts (and no assertion failures) in suites unrelated to this branch. Leaving CI as the arbiter there.

---

AI-assisted — [session 1](https://claude.ai/code/session_01MF7wiTohf41KusYVwf5a8M), [session 2](https://claude.ai/code/session_01Y4jipgyyjSZhhtq8RwdoZ4)

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2750)


e2e-tests/fixtures/scripture-editor-helpers.ts
```diff
@@ -231,11 +235,10 @@ export async function navigateToolbarBcv(mainPage: Page, reference: string): Pro
   await mainPage.locator('button[aria-label="book-chapter-trigger"]').first().click();
   const input = mainPage.locator('[data-radix-popper-content-wrapper] input');
   await input.fill(reference);
-  const highlightedTopMatch = mainPage.locator(
-    '[data-radix-popper-content-wrapper] [cmdk-item][data-selected="true"]',
-    { hasText: new RegExp(`${escapeForRegExp(reference)}\\b`, 'i') },
-  );
-  await highlightedTopMatch.waitFor({ timeout: 10_000 });
+  const topMatchRow = mainPage.locator('[data-radix-popper-content-wrapper] [cmdk-item]', {
+    hasText: new RegExp(`${escapeForRegExp(reference)}\\b`, 'i'),
```
mattgetgen: **The `\b` anchor can no longer match — 15 E2E call sites across 7 spec files will time out.**

The diff appends a book-id `<span>` immediately after the reference `<span>` in the top-match row (`book-chapter-control.component.tsx:1434-1442`), with no whitespace text node between them. Playwright tests `hasText` against the element's *concatenated* text, which is now `"Jonah 1:1JON"` — a digit followed by a word character, so there is no word boundary:

```js
new RegExp('Jonah 1:1' + '\\b', 'i').test('Jonah 1:1JON')  // false
```

Same for `Obadiah 1:21`, `Exodus 2:3`, `Genesis 1:1`, `Luke 4:1`. Every one of the 15 call sites passes a verse-level reference ending in a digit — `marker-palette-trigger-focus`, `type-through-save-echo`, `scroll-group-sync`, `attribute-display-settle`, `verse-navigation-shortcuts`, `formatted-default-simple-mode`, `standard-default-power-mode` — so `topMatchRow.first().waitFor({ timeout: 10_000 })` times out and all of those specs fail.

The sibling helper in `navigation-history.spec.ts` survives only by luck: it passes *chapter*-level regexes, and `/Mark 4\b/` matches `"Mark 4:1MRK"` because `:` follows the `4`.

Also stale: the docstring says the row "renders through `formatScrRef(..., 'English')`". It now renders through `getLocalizedBookName`.

lib/platform-bible-react/src/components/advanced/book-chapter-control/book-chapter-control.component.tsx
```diff
@@ -661,165 +750,121 @@ export function BookChapterControl({
...
-      if (viewMode === 'verses') {
-        // Handle backspace for going back to chapters
-        if (event.key === 'Backspace') {
-          event.preventDefault();
-          event.stopPropagation();
-          handleBackToChapters();
-          return;
+      // Narrows `event.key` to `ArrowKey` for `computeTargetGridItem` below, and leaves every
+      // non-arrow key to cmdk.
+      if (!isArrowKey(event.key)) return;
```
mattgetgen, line 865: **No interactive-target guard here, so this steals all four arrow keys from the RecentSearches menu.**

The Enter branch 48 lines above deliberately excludes `[role="menuitem"]`; this branch has no equivalent check. `handleCommandKeyDown` is `onKeyDownCapture` on `PopoverContent`, and although `DropdownMenuContent` portals to `body` it remains a React child — so React routes its keydowns through this capture handler first.

Repro: main toolbar BCV → type `gen` (top match + chapter preview grid appear) → click the clock button → press <kbd>ArrowDown</kbd>. The grid resolver at line 804 matches, lines 846-847 run `preventDefault()` + `stopPropagation()`, and Radix's bubble-phase `RovingFocusGroup` item handler never fires. **The menu is not keyboard-navigable**, and the hidden grid's highlight moves instead.

Reachable in all three shipping BCV instances — `platform-bible-toolbar.tsx:585`, `web-view.component.tsx:568`, `platform-scripture-editor.web-view.tsx:3808` all pass `recentSearches`.

The same handler also kills `handleQuickNavButtonKeyDown`'s documented ArrowUp/ArrowDown "pull focus to the search input" path (lines 898-901), whose comment — *"the keydown keeps bubbling up past us to reach it"* — is now false.

No coverage: `book-chapter-control.component.test.tsx` never renders `recentSearches`.

lib/platform-bible-react/src/components/advanced/book-chapter-control/book-chapter-control.component.tsx
```diff
@@ -640,13 +677,65 @@ export function BookChapterControl({
...
+        // so does any input that isn't this picker's search box. `menuitem` covers the recent
+        // searches list: it portals out of this popover but stays a React child, so its keystrokes
+        // still reach this capture-phase handler, and its items are focusable `div`s rather than
+        // buttons.
+        //
+        // Excluding the search box is load-bearing, not a special case: it is itself an `input`, so
+        // without this it would match the selector below and Enter — the normal way to commit a
+        // typed reference — would never submit anything. The chapters/verses branch above needs no
+        // such exclusion because the search box is not rendered in those views.
+        const isTargetInteractive =
```
mattgetgen, line 831: **`isTargetInteractive` misses `[role="menu"]`, so Enter navigates away while the menu is open.**

Radix focuses the menu *content* (`role="menu"`, `tabIndex=-1`) in its `onMountAutoFocus` — not the first item. So immediately after the RecentSearches menu opens, the event target matches nothing in this selector:

```
button, a, input, select, textarea, [role="button"], [role="menuitem"]
```

`target.closest(...)` returns `null` for the menu content (its DOM ancestors are the portal container and `body`), so `isTargetInteractive` is false and the handler runs `preventDefault` + `stopPropagation` + `handleTopMatchSelect()`.

Repro: type `Matt` → click the clock button → press <kbd>Enter</kbd> without hovering an item. The app navigates to the typed reference and the picker closes, instead of Enter doing nothing on a menu with no highlighted item.

The comment at 720-724 shows the portalled-menu case was considered, but only the *item* half was handled. Combined with the arrow-key finding above, the `[role="menuitem"]` exclusion is effectively dead for keyboard users: arrows never move focus onto an item, so the target is always the menu content.

lib/platform-bible-react/src/components/advanced/book-chapter-control/book-chapter-control.component.tsx
```diff
@@ -661,165 +750,121 @@ export function BookChapterControl({
...
-            targetChapter =
-              currentChapter === 0 ? 1 : Math.min(maxChapter, currentChapter + GRID_COLS);
-            break;
-          default:
-            return;
-        }
+      // An arrow key aimed at the grid belongs to the grid even when the highlight does not move.
+      // The Command root navigates with `loop`, so releasing a clamped keystroke to it would wrap
+      // the highlight to the opposite edge — precisely what clamping exists to prevent.
+      event.preventDefault();
```
mattgetgen, line 1011: **`preventDefault()` is now unconditional — ArrowLeft/Right can no longer move the caret, and silently retarget what Enter submits.**

This runs *before* the `if (target === grid.current) return;` bail-out, and the books view is included. In books view focus stays on `CommandInput` (`takeFocus: viewMode === 'chapters'`, line 816).

Repro with a `getEndVerse` consumer (`scope-selector.component.tsx:552`): type `mat 12:15` so the verse preview grid is on screen, then press <kbd>ArrowLeft</kbd> to put the caret back into the number. The caret does not move. The verse highlight steps 15 → 14, and because `topMatchReference` (lines 234-250) prefers the highlighted cell over the parsed query, the row re-renders as **"Matthew 12:14"** — and the next <kbd>Enter</kbd> submits a verse the user never typed.

This is a regression. `origin/main` called `preventDefault` only inside `if (targetChapter !== currentChapter)`, and its `currentChapter` parsed to `0` off the old top-match value (which ended in `")}"`), so the ArrowLeft/ArrowRight branches were guarded by `if (currentChapter !== 0)` and never fired in books view.

<kbd>Home</kbd>/<kbd>End</kbd> are explicitly forwarded to the input (lines 626-628); Left/Right are not. No test covers caret behavior.

lib/platform-bible-react/src/components/advanced/book-chapter-control/book-chapter-control.component.tsx
```diff
@@ -640,13 +677,65 @@ export function BookChapterControl({
...
+          if (!activation.isDisabled) activation.activate();
+          return;
+        }
+      }
+
+      // Enter over a top match submits the reference its row displays. The preview grid owns the
+      // highlight, so cmdk would otherwise dispatch Enter at the highlighted cell — which knows
+      // only a chapter and would drop the verse the user typed. Space is deliberately not included:
+      // the books view is a text field, where a space is a character.
+      if (viewMode === 'books' && topMatch && event.key === 'Enter') {
```
mattgetgen: **Enter on a highlighted chapter cell no longer drills into the verse picker — but a mouse click on the same cell still does.**

The old activation block was gated to `(viewMode === 'chapters' || viewMode === 'verses')`, so in books view Enter fell through to cmdk, which dispatched select at the highlighted chapter cell and reached `handleChapterSelect` → `if (getEndVerse) { if (endVerse > 1) { setViewMode('verses'); return; } }`.

The new unconditional `if (viewMode === 'books' && topMatch && event.key === 'Enter')` calls `handleTopMatchSelect()` instead.

Repro in the Scripture editor BCV (passes `getEndVerse`; so do `scope-selector:552/577/917` and `checklist.web-view.tsx:851`): type `Mark` → <kbd>ArrowDown</kbd> to chapter 7 → <kbd>Enter</kbd>. Old behavior opened the Mark 7 verse grid; new behavior navigates straight to Mark 7:1. Clicking cell 7 still opens the verse grid — **keyboard and pointer now disagree** for every consumer that passes `getEndVerse`.

Two artifacts still assert the old contract:
- `book-chapter-control.component.test.tsx:1129` — *"Enter over a highlighted grid cell goes through cmdk's own onSelect to handleChapterSelect and never reaches handleTopMatchSelect"*
- `src/stories/keyboard-shortcuts.data.ts:136` — the new catalog entry's purpose says the books view *"has no grid"*. It renders a chapter/verse preview grid that owns the cmdk highlight. `.claude/rules/keyboard-shortcuts-catalog.md` requires each entry's purpose to be accurate.

lib/platform-bible-react/src/components/advanced/book-chapter-control/book-chapter-control.component.tsx
```diff
@@ -986,6 +1028,101 @@ export function BookChapterControl({
...
+  // highlight the same way the chapters and verses views seed theirs, so every grid surface obeys
+  // the same rule: exactly one cell is highlighted for as long as the grid is on screen, and arrow
+  // keys only ever move it.
+  //
+  // The seed also has to be defended, not just written once. cmdk moves its own highlight to the
+  // first item in the list — the top-match row, whose value names no cell — whenever the filter
+  // text changes and whenever the pointer crosses that row, so the highlight is put back on the
+  // cell it last named (remembered here) rather than left off the grid. A new parsed reference
+  // seeds afresh; a highlight the user moved survives.
+  useLayoutEffect(() => {
```
mattgetgen, line 1256: **This layout effect ping-pongs against cmdk's controlled value — a measured 30x commit regression on pointer-move.**

Measured with the React Profiler in jsdom, warm: 20 repeated `pointermove` events on the top-match row cost **60 commits / 79-157 ms** on this branch versus **2 commits / 1.7 ms** on `origin/main`.

Mechanism: cmdk's `CommandItem` `onPointerMove` sets its store value to `TOP_MATCH_ITEM_VALUE` → `onValueChange` → `setCommandValue` → render → this effect sees a value naming no grid cell and writes the remembered cell back → second render. cmdk's own store dedupe (which made main's 2nd-20th moves free) is defeated because the value is restored between events.

A user sweeping the mouse toward the chapter grid *must* cross the top-match row directly above it. At 60-120 Hz that is 180-360 commits/second, each reconciling the whole preview grid — 28 cells for Matthew, 150 for Psalms, up to 176 for a verse preview.

The effect's comment acknowledges the fight (*"whenever the pointer crosses that row"*) but treats it as a one-time correction. This file already has the structural fix: render the top-match row **outside** `CommandList`, the way `canShowMoreBooksToggle` is at line 1537 (*"inside it, cmdk would register this as an option"*).

lib/platform-bible-react/src/components/advanced/book-chapter-control/book-chapter-control.component.tsx
```diff
@@ -661,165 +750,121 @@ export function BookChapterControl({
...
-          case 'ArrowDown':
-            targetVerse = currentVerse === 0 ? 1 : Math.min(maxVerse, currentVerse + GRID_COLS);
-            break;
-          default:
-            return;
+        // The books view renders its chapter preview only for books with more than one chapter, so
+        // a single-chapter top match (Jude, Obadiah, Philemon, 2-3 John) has no grid to drive. Claim
+        // the keystroke there and it would write a chapter value no rendered item carries, blanking
+        // the highlight; let it fall through to cmdk so the list stays navigable instead.
+        if (
```
mattgetgen, line 961: **The grid-existence guard ignores `isCommandListHidden`, so arrows get eaten for a grid that is not mounted.**

Repro: type `mat 12` (topMatch set, chapter preview rendered), then click the "Next chapter" quick-nav button — its `onClick` runs `setIsCommandListHidden(true)` (line 1299), unmounting `{!isCommandListHidden && <CommandList>}` and with it the top-match row and the entire preview grid.

`viewMode` is still `'books'` and `topMatch` is unchanged, so this guard still yields `{ max: 28 }` and lines 846-847 swallow the keystroke. `grid.refs.current[target]` is `null` (React nulled the refs on unmount), so nothing scrolls and nothing visible moves — the arrow key is eaten and denied to cmdk, and `handleQuickNavButtonKeyDown`'s ArrowUp/ArrowDown focus-restore never runs.

The `fetchEndChapter(topMatch.book) > 1` conjunct exists precisely to avoid claiming keys when no grid is rendered (see its comment at 800-803). `isCommandListHidden` is the other half of that same condition. `canShowMoreBooksToggle` (lines 553-557) does check it.

lib/platform-bible-react/src/components/advanced/book-chapter-control/book-chapter-control.component.tsx
```diff
@@ -1092,289 +1229,332 @@ export function BookChapterControl({
...
+                            <Button
+                              variant="ghost"
+                              size="sm"
+                              onClick={() => {
+                                setIsCommandListHidden(true);
+                                onClick();
+                              }}
+                              disabled={isQuickNavDisabled}
+                              className="tw:h-8.5 tw:w-6 tw:p-0"
+                              aria-label={title}
```
mattgetgen: **Swapping `title` for a Radix Tooltip means the *disabled* arrows now have no hover explanation at all.**

Removed: `title={title}` (`origin/main` line 1149). The browser renders a native title tooltip even on a disabled control; Radix's Tooltip cannot, because Chromium does not dispatch `pointerenter`/`mouseenter`/`focus` on a disabled form control — and `TooltipTrigger asChild` here wraps a real `<button disabled>` (`disabled={isQuickNavDisabled}`, line 1302).

At Genesis 1:1 both "Previous chapter" and "Previous verse" come back disabled from `isNoOpNavigation` — so the default state of a freshly opened project is the state with no explanation.

Compounding it: this PR also changed the verse arrows from horizontal chevrons (`ChevronLeft`/`Right`, RTL-mirrored) to vertical ones (`ChevronUp`/`Down`), so the glyph no longer distinguishes verse from chapter by direction. The user hovers a greyed vertical chevron and gets nothing.

The accessible name survives via `aria-label`, so this lands squarely on sighted pointer users — the population the six new `%webView_bookChapterControl_*Chapter/Verse%` strings were added for.

lib/platform-bible-react/src/components/shared/book-item.component.tsx
```diff
@@ -146,12 +147,27 @@ export function BookItem({
       aria-label={ariaLabel}
       disabled={disabled}
       className={cn(
+        LIST_ITEM_KEYBOARD_FOCUS_RING,
+        // Suppress CommandItem's own data-selected background and text color so the keyboard
+        // highlight is the ring alone. Book rows and grid cells belong to one control and share one
+        // highlight language; a background here would make the same keyboard state look different
+        // depending on which view the user is in.
+        'tw:data-selected:bg-transparent tw:data-selected:text-inherit',
```
mattgetgen, line 167: **Suppressing the `data-selected` colors leaves a box-shadow ring as the only highlight — which forced-colors mode strips entirely.**

The added `'tw:data-selected:bg-transparent tw:data-selected:text-inherit'` here, and the identical line at `numbered-item-grid.component.tsx:81`, cancel `CommandItem`'s base highlight (`tw:data-selected:bg-muted tw:data-selected:text-foreground`, `command.tsx:249`) — the only highlight these items had before this PR.

The replacement, `LIST_ITEM_KEYBOARD_FOCUS_RING` (`focus.util.ts:57`), is implemented by Tailwind as `box-shadow`, and CSS Color Adjust forces `box-shadow: none` under `forced-colors: active`.

Net effect on Windows High Contrast: a user arrows through the book list or chapter grid and **nothing moves**. The current-chapter `bg-primary` cell still paints (a background is re-mapped, not removed), but the cell <kbd>Enter</kbd> will actually submit is unmarked. The old `data-selected` background survived forced-colors.

Neither the unit tests nor the Storybook play functions assert a *visible* highlight — they assert the `data-selected` attribute — so nothing catches this.

A `forced-colors:` outline fallback inside `LIST_ITEM_KEYBOARD_FOCUS_RING` fixes both consumers at once.

lib/platform-bible-react/src/components/advanced/recent-searches.component.tsx
```diff
@@ -71,37 +89,46 @@ export default function RecentSearches<T>({
...
+    <DropdownMenu open={isOpen} onOpenChange={setIsOpen} modal={false}>
+      {/* This component is exported standalone and may be rendered anywhere, so it carries its
+          own TooltipProvider rather than assuming a host tree already has one. */}
+      <TooltipProvider>
+        {ariaLabel ? (
+          <Tooltip>
+            <TooltipTrigger asChild>
+              <DropdownMenuTrigger asChild>{button}</DropdownMenuTrigger>
+            </TooltipTrigger>
+            <TooltipContent>{ariaLabel}</TooltipContent>
```
mattgetgen: **`ariaLabel` is now rendered as visible tooltip text, so a raw `%key%` becomes user-visible microcopy.**

`useLocalizedStrings` returns `{ [key]: key }` while loading and *permanently* on `isPlatformError` (`src/renderer/hooks/papi-hooks/use-localized-strings-hook.ts:44-48`). Both in-repo consumers feed it straight through:

- `book-chapter-control.component.tsx:1268` → `localizedStrings?.['%history_recentSearches_ariaLabel%']`
- `find.component.tsx:953` → `localizedStrings['%webView_find_showRecentSearches%']`

Because a key string is non-empty, the `ariaLabel ? <Tooltip>` branch is taken and hovering the clock button shows a tooltip reading literally `%history_recentSearches_ariaLabel%`.

This PR's own test asserts that shape: `find.component.test.tsx:891` queries the button by the raw key as its accessible name. Previously this value only reached `aria-label`, where a raw key was invisible.

Suggested guard: treat a value still matching `/^%.*%$/` as missing and fall back to the component default.

lib/platform-bible-react/src/components/advanced/recent-searches.component.tsx
```diff
@@ -71,37 +89,46 @@ export default function RecentSearches<T>({
...
+          </Tooltip>
+        ) : (
+          <DropdownMenuTrigger asChild>{button}</DropdownMenuTrigger>
+        )}
+      </TooltipProvider>
+      <DropdownMenuContent id={id} className="tw:w-[300px]" align="start">
+        <DropdownMenuLabel>{groupHeading}</DropdownMenuLabel>
```
mattgetgen: **The rebuild dropped the list's programmatic group label.**

Removed: `<CommandGroup heading={groupHeading}>`. cmdk's `Group` renders `<div cmdk-group-heading id={headingId}>` next to `<div cmdk-group-items role="group" aria-labelledby={headingId}>`, so a screen reader announced the options as a group named "Recent" (localized via `%history_recent%` / `%webView_find_recent%`).

The replacement renders Radix's `Menu.Label` — a plain `<div>` with no role, no id, and nothing referencing it (`dropdown-menu.tsx:293`). The exported `DropdownMenuGroup` is not used, and `aria-label` is not set on `DropdownMenuContent`.

A screen-reader user activating the clock button in the BCV popover or the Find panel now hears "menu" with N menuitems, with no indication these are recent searches.

The new `recent-searches.component.test.tsx` asserts menu/menuitem roles but nothing about the heading, so the regression is untested — even though the ADR frames this role change as a deliberate accessibility contract.

lib/platform-bible-react/src/components/advanced/recent-searches.component.tsx
```diff
@@ -71,37 +89,46 @@ export default function RecentSearches<T>({
...
-      </PopoverContent>
-    </Popover>
+    <DropdownMenu open={isOpen} onOpenChange={setIsOpen} modal={false}>
+      {/* This component is exported standalone and may be rendered anywhere, so it carries its
+          own TooltipProvider rather than assuming a host tree already has one. */}
+      <TooltipProvider>
+        {ariaLabel ? (
+          <Tooltip>
+            <TooltipTrigger asChild>
+              <DropdownMenuTrigger asChild>{button}</DropdownMenuTrigger>
```
mattgetgen, line 163: **Nesting `DropdownMenuTrigger` inside `TooltipTrigger` makes the tooltip re-open itself every time the menu closes.**

Radix `DropdownMenuContent` composes `onCloseAutoFocus` to call `context.triggerRef.current?.focus()` (`react-dropdown-menu/dist/index.mjs:113-117`), and `MenuRootContentNonModal` passes `onCloseAutoFocus` straight through — so `modal={false}` does **not** suppress it.

`TooltipTrigger`'s `onFocus` calls `context.onOpen()` when `isPointerDownRef` is false (`react-tooltip/dist/index.mjs:202-204`) — and a mouse click lands its `pointerdown` on the menu *item*, not the trigger, so that ref is false. `handleOpen` bypasses `delayDuration` (which this repo defaults to 0 anyway, `tooltip.tsx:18`).

Repro in Find: click the clock button, click a recent term. The menu closes, focus returns to the clock button, and the tooltip pops open unprompted over the search input — and stays until the button blurs, because no `pointerleave` will ever fire (the pointer was never on the button). Same on <kbd>Escape</kbd>.

Confirmable in jsdom: open, select an item, assert no `tooltip` role is present.

lib/platform-bible-react/src/components/shared/book-item.utils.ts
```diff
@@ -11,3 +11,48 @@ export function generateCommandValue(
...
+export function verseItemValue(bookId: string, chapterNum: number, verse: number): string {
+  return `${chapterItemValue(bookId, chapterNum)}:${verse}`;
+}
+
+/**
+ * Reads the chapter number back out of a {@link chapterItemValue}, or `undefined` when the value
+ * does not identify a chapter cell (a book row, or a verse cell — a verse value ends in digits too,
+ * so the colon has to be excluded explicitly).
+ */
+export function parseChapterFromItemValue(value: string): number | undefined {
```
mattgetgen, line 51: **`/(\d+)$/` misparses the two canon books whose English name ends in digits.**

The docstring promises `undefined` "when the value does not identify a chapter cell (a book row ...)". Verified against `@sillsdev/scripture`:

```js
Canon.bookIdToEnglishName('PS2')  // 'Psalm 151'
Canon.bookIdToEnglishName('PS3')  // 'Psalms 152-155'
```

So `generateCommandValue('PS2')` yields `"PS2 Psalm 151"`, which this parser reads as chapter **151** (PS3 → **155**).

`topMatchReference` (`book-chapter-control.component.tsx:239-241`) feeds it exactly that — its only guard is `commandValue.startsWith(topMatch.book + ' ')`, which passes for a *book row* of the same book. With `scrRef.book === 'PS2'`, the books auto-scroll effect (line 943) sets `commandValue` to that book-row value; on the render where typing `ps2` first resolves `topMatch` to PS2, the row computes `chapterNum` 151 and would submit `"Psalm 151 151:1"`.

What prevents a wrong submit today is incidental: cmdk's `selectFirstItem` fires when the book list unmounts and resets `commandValue` to the sentinel before paint. That is, the safety comes from the library-internal behavior the ADR says this design stopped relying on — not from the parser.

`book-item.utils.test.ts:96` only covers `'JHN John'`. A stricter parse (require the value to equal `chapterItemValue(book, n)`, as `cellNamedBy` at line 1091 already does) closes it.

mattgetgen (changes requested):
## AI-assisted review — 15 verified findings

Reviewed at commit `3bb3286` against `origin/main` (`30f07f4`). Each finding below was verified with a concrete failure scenario — reproduced, measured, or traced to library source — rather than pattern-matched.

Generated with Claude Code (`/code-review max`) and reviewed before posting.

---

### Blocks CI

**`e2e-tests/fixtures/scripture-editor-helpers.ts:239`** — the new book-id span makes the top-match row's text `"Jonah 1:1JON"`, so the helper's `\b` anchor can never match. All 15 verse-level call sites across 7 spec files time out. Details inline.

### Keyboard regressions (5, all in the same capture handler)

The `onKeyDownCapture` handler in `book-chapter-control.component.tsx` intercepts keys before both the portalled RecentSearches menu and cmdk. Five distinct consequences, inline at lines 718, 730, 766, 804 and 846 — the most user-visible being that **<kbd>Enter</kbd> on a highlighted chapter cell no longer drills into the verse picker, while a mouse click on that same cell still does.**

### Accessibility (3)

Inline at `book-chapter-control.component.tsx:1304`, `book-item.component.tsx:155`, and `recent-searches.component.tsx:119`. In short: disabled quick-nav arrows lost all hover text; the keyboard highlight is now a box-shadow ring, which forced-colors mode strips entirely; and the recent-searches list lost its programmatic group label.

### Correctness / performance (4)

Inline at `recent-searches.component.tsx:110` and `:112`, `book-chapter-control.component.tsx:1051`, and `book-item.utils.ts:38`.

---

### Two findings in files this PR does not touch

GitHub will not anchor these, so they are here in full.

**1. `extensions/src/platform-enhanced-resources/src/components/toolbar/toolbar.component.tsx:348` — the one BCV call site with no `localizedStrings`, so seven newly-visible strings are English-only there.**

`platform-bible-toolbar.tsx:100`, `web-view.component.tsx:71` and `platform-scripture-editor.web-view.tsx:322` all spread `[...BOOK_CHAPTER_CONTROL_STRING_KEYS]`, so the six keys added to `book-chapter-control.types.ts` flow automatically. Here `useQuickNavButtons` receives `undefined` and every `localizedStrings?.[...] || 'English default'` falls through.

Before this PR the quick-nav titles were English-only `aria-label`s (invisible) and the back button had no tooltip. This PR turns them into *visible* Tooltip text and adds two visible back-button labels — and this call site also passes `recentSearches`, so the clock tooltip shows the English default too. A Spanish user opens an Enhanced Resource tab and sees "Previous chapter" / "Next verse" / "Back to books" while the main toolbar shows "Capítulo anterior". The `es.json` translations this PR adds are never requested.

**2. `lib/platform-bible-react/src/components/advanced/project-selector/project-selector.component.tsx:526` — the z-index sweep left five stale overrides, so the ADR's central claim is not true of the shipped code.**

The ADR states that "every shadcn overlay sets the constant for its own tier itself." Remaining counterexamples:

- **(a)** `<DropdownMenuContent style={{ zIndex: Z_INDEX_OVERLAY }}>` (400) renders inside ProjectSelector's own `PopoverContent` (line 922, defaulting to `Z_INDEX_ABOVE_DOCK` = 600). The inline style merges *after* `{ zIndex: Z_INDEX_ABOVE_DOCK, ...style }`, defeating this PR's new default. Latent only because the sole in-repo consumer, `settings-sidebar.component.tsx:177`, also pins the host popover to 400 — the moment anyone renders `ProjectSelector` (exported from `experimental`) without `popoverContentStyle`, the filter menu paints behind its own popover. That is verbatim the PR #2229 failure mode the ADR says it closed.
- **(b)** Line 456 pins a `TooltipContent` to `Z_INDEX_OVERLAY`, defeating the 550→675 raise for that row tooltip.
- **(c)** `er-dictionary-filtered-list.component.tsx:529` still uses `Z_INDEX_MODAL + 10` — the exact arithmetic-on-a-constant pattern this PR removed from `project-picker`, and the only one left in the repo.
- **(d)** `settings-sidebar.component.tsx:176` still carries `// (Z_INDEX_ABOVE_DOCK = 250)` — the only stale 250 left after the sweep rewrote that same claim in `z-index.ts`, `tooltip.tsx` and `_vars.scss`.
- **(e)** `menubar.tsx:113` and `:312` still hard-code `tw:z-50` and import no `Z_INDEX` constant, while the new comments at `tooltip.tsx:9-12` and `dropdown-menu.tsx:14-16` both name "the menubar" as a member of that tier. Nothing in `z-index.test.tsx` covers the menubar.

---

<details>
<summary><b>Additional lower-severity findings (11)</b> — not individually verified to the same depth</summary>

1. **Duplicate catalog entry.** The new `book-chapter-control-pick-highlighted-book` (`keyboard-shortcuts.data.ts:145`) duplicates the pre-existing `picker-select-highlighted-item` (`:576`) — same key, same `spaceSelectsHighlightedItem` mechanism, same two `locations` — but files it under `Navigation` instead of `Selection`. One shortcut renders under two headings in the catalog that `.claude/rules/keyboard-shortcuts-catalog.md` calls a single source of truth.
2. **ADRs not promoted.** The three new ADRs were never folded into a standard or a `.claude/rules/` file, which the root `CLAUDE.md` requires once a decision hardens into a rule.
3. **ADR link dies on squash-merge.** One ADR cites `git diff main...HEAD` as its evidence pointer — an unresolvable reference once this PR squashes.
4. **Dead parameter blessed by a test.** `computeTargetGridItem`'s `columns?` parameter has no production caller, yet a test pins it — contradicting `NumberedItemGrid`'s own comment that a per-instance column override must not exist.
5. **Test degrades silently.** `z-index.test.tsx:133` compares `Number(el.style.zIndex)`, so it quietly weakens to `675 > 0` if `PopoverContent` ever moves back to a class.
6. **Tests pin an attribute the ADR calls unpinnable.** ~22 new assertions read cmdk's `data-selected`.
7. **Unreachable code.** `topMatchReference`'s `chapterNum` ternary has a provably unreachable arm.
8. **Submenu tier bug.** `DropdownMenuSubContent` hard-codes 600, so a submenu of a menu lifted to `Z_INDEX_ABOVE_POPOVER` (footnote-type / caller dropdowns) would paint under its parent.
9. **Double callback.** `onOpenChange(false)` now fires twice per RecentSearches selection, since Radix auto-closes.
10. **Duplicated logic.** `isNoOpNavigation` re-implements `compareScrRefs`.
11. **Dead parameter still tested.** `generateCommandValue`'s `chapter` parameter is now unused.

</details>

---

### Checked and clean

- No test was deleted or weakened — `navigation.test.ts`'s removals are index-to-name refactors plus the justified RTL icon change.
- Every assertion from the deleted `z-index.test.ts` survives.
- The committed `dist/` is consistent with `src/` for every changed public symbol.
- All six new localization keys are present in `en.json`, `es.json`, and `BOOK_CHAPTER_CONTROL_STRING_KEYS`.
- Every `// CUSTOM:` annotation required by the shadcn rule is present.
- `describe(..., 15_000)` is a real Vitest 3.2.7 suite timeout, not a no-op.

Sebastian-ubs:
great, does
>  including Sebastian's changes 

mean, this is superseding https://github.com/paranext/paranext-core/pull/2229 and I can close this now?

katherinejensen00:
Everything is answered in this one comment rather than in the individual threads — time pressure on my end, and I didn't want any of it to wait. I'll try to reply in the threads themselves tomorrow so each one carries its own answer.

Rebased onto `main` (`8f230e8`) first; everything below is on the new head. Thanks for the depth on these — the repros made almost all of them straightforward to confirm.

## Blocks CI

**`scripture-editor-helpers.ts:239` — the `\b` anchor.** Confirmed exactly as described. Replaced `\b` with `(?!\d)`, which asserts what the anchor was actually there for: `Mark 4(?!\d)` still accepts `"Mark 4:1MRK"` and still rejects `"Mark 40:1MRK"`, and it no longer depends on what character follows the number. Docstring corrected too — the row renders through `getLocalizedBookName`, not `formatScrRef(..., 'English')`. I also moved `navigation-history.spec.ts`'s five chapter-level regexes to the same anchor rather than leaving them passing by luck.

## Keyboard regressions

All five were real.

**`:730` (`[role="menu"]`) and `:766` (arrows)** — both fixed by one guard at the top of `handleCommandKeyDown`: any event target inside `[role="menu"], [role="menuitem"]` returns immediately, so Enter, Space, the arrows, letters/digits and Backspace all stay with the list. Both roles, for the reason you gave — Radix focuses the content first and only moves onto a row once the user arrows.

That alone wasn't enough for Enter, though, and it's worth flagging: with our capture handler out of the way the keydown still bubbled through the React tree to cmdk's `Command` root, which drilled into the verse grid. So `RecentSearches` now stops keydown propagation out of its `DropdownMenuContent` — while that list is open it owns the keyboard, and it's documented as opening inside other surfaces. Two new tests in `book-chapter-control.component.test.tsx` (ArrowDown moves focus onto the first row; Enter neither submits nor closes the picker), both red before the fix.

**`:846` unconditional `preventDefault`** — fixed, and your repro is now a test. Rather than handing the horizontal arrows to the caret unconditionally, they follow the combobox rule: a horizontal arrow belongs to the caret while the caret can still move that way, and reaches the preview grid only once it can't. So `mat 12:15` + <kbd>ArrowLeft</kbd> moves the caret and leaves the highlight on 15, while <kbd>ArrowRight</kbd> at the end of the query still steps the grid — which is what the existing grid-navigation tests exercise, so both behaviours are pinned. RTL-aware: which arrow walks toward the start of the text is a property of the layout, not of the key.

**`:804` `isCommandListHidden`** — fixed; the arrow branch bails when the list is unmounted. Test asserts focus is back on the search input after quick navigation hides the list.

**`:718` Enter vs. a click on the same chapter cell** — this is the one I did **not** change, and I'd like your read rather than a unilateral call.

It's a deliberate product decision (called out in the PR body). In the books view the grid is a *preview of the typed reference*; the top-match row displays "Mark 7:1"; Enter submits what the row displays. Making Enter drill into the verse grid would mean the row can display a reference that Enter doesn't submit, which is the invariant the rest of this work is built on. You're right that pointer and keyboard now disagree for `getEndVerse` consumers — I'd rather close that by making the *click* consistent with Enter than the reverse, but that's a bigger behavioural change than I want to make on review feedback. Happy to go either way; say which.

Both stale artifacts you flagged alongside it are fixed: the test comment no longer claims Enter routes through cmdk's `onSelect`, and the catalog entry no longer says the books view "has no grid".

## Accessibility

**`:1304` disabled quick-nav arrows** — fixed. `title` is set only while the button is disabled, so a disabled arrow keeps native hover text and an enabled one doesn't render two bubbles. (Left the vertical verse chevrons alone — that one is a deliberate design choice, and I'd rather revisit it as a design question than fold it into this.)

**`book-item.component.tsx:155` forced-colors** — fixed in `LIST_ITEM_KEYBOARD_FOCUS_RING` itself, so both consumers get it at once as you suggested: a `forced-colors:` `outline` using the `Highlight` system colour alongside the existing ring.

**`recent-searches.component.tsx:119` group label** — fixed. `DropdownMenuLabel` gets an id and the content an `aria-labelledby`, so the list announces as "Recent" instead of a bare menu of N items. Test asserts `findByRole('menu', { name: 'Recent' })`.

## Correctness / performance

**`:112` raw `%key%` rendered as tooltip text** — fixed, and generalised, since the same shape reaches several newly-visible strings: new `resolveLocalizedString(value, fallback)` treats `/^%[^%]*%$/` as not-yet-localized. `RecentSearches` uses it for both `ariaLabel` and `groupHeading`. `find.component.test.tsx:891` now asserts the resolved name rather than the key, plus a new case that the localized value wins once strings arrive.

**`:110` tooltip re-opens itself when the list closes** — reproduced in jsdom exactly as described (`data-state="instant-open"` on the tooltip after selecting a row). Fixed by controlling the tooltip's open state and swallowing the one focus-restore open that follows a close; a later hover or Tab opens normally. Test: open, select an item, assert no `tooltip` role.

**`:1051` pointer-move ping-pong** — fixed at the source rather than by moving the row out of `CommandList`. The fight only exists because cmdk moves the highlight under the pointer, and the design here already says hover and the keyboard highlight are different things (every item styles hover separately and suppresses `data-selected`). So `Command` now sets `disablePointerSelection`, and the correcting effect has nothing left to correct. I also hoisted the preview-grid computation out of the effect into a memo the effect and the key handler share, so they can't disagree about whether a grid exists.

Worth recording: I tried simply declining the change in `onValueChange` first, and it does **not** work — cmdk only syncs the controlled `value` prop into its store when the prop changes, so refusing the change leaves its internal highlight sitting on the row. Five tests caught it.

**`book-item.utils.ts:38` PS2 / PS3** — fixed. `parseChapterFromItemValue` now round-trips its candidate through `chapterItemValue` and accepts only an exact match, so `"PS2 Psalm 151"` no longer reads as chapter 151; `parseVerseFromItemValue` inherits the check. Parameterised tests over both books, plus the reverse case that genuine PS2/PS3 chapter and verse cells still round-trip. I also removed `generateCommandValue`'s dead `chapter` parameter, which was the only other way to spell a book row that looks like a chapter cell.

## The two out-of-file findings

**1. `platform-enhanced-resources` toolbar had no `localizedStrings`** — fixed. `BOOK_CHAPTER_CONTROL_STRING_KEYS` is spread into `TOOLBAR_STRING_KEYS`, so the keys flow through the wiring layer's existing single `useLocalizedStrings` call, and the bag is passed to `BookChapterControl`.

**2. z-index stragglers** — (a) and (b) fixed by deleting both `project-selector` overrides; the filter menu and the row tooltip now take their own components' tiers. (c) `Z_INDEX_MODAL + 10` removed. (e) `menubar.tsx` now sets `Z_INDEX_ABOVE_DOCK` on content and sub-content, with a rendered-stacking test — you were right that nothing covered it.

(d) is the one I left: `settings-sidebar.component.tsx` pins a *host popover* to 400, which puts it under the modal tier, and I can't verify which surfaces that sidebar renders in without running the app. The stale "250" comment is replaced with an accurate one and a TODO. The ADR's claim is corrected to match what actually shipped, including the fact that it wasn't true of the menubar when it was written.

## Lower-severity

Done: **1** (duplicate catalog entry removed — `picker-select-highlighted-item` already lists the BCV), **2** (z-index decision promoted to `.claude/rules/ux/z-index-tiers.md`), **3** (`git diff main...HEAD` pointers replaced), **4** (`columns` parameter and its test removed), **5** (`z-index.test.tsx` compares declared strings now, so it can't decay to `675 > 0`), **7** (ternary collapsed — a value can't name both a chapter and a verse cell, so the case-split was redundant), **8** (`DropdownMenuSubContent` → `Z_INDEX_ABOVE_POPOVER`), **9** (double `onOpenChange` — Radix auto-closes; test pins one close per selection), **10** (`isNoOpNavigation` delegates to `compareScrRefs`), **11** (dead `chapter` parameter removed).

Not done: **6**, the ~22 assertions on `data-selected`. The ADR's point is that *production code* shouldn't steer cmdk through its private DOM, which it no longer does. Reading the attribute in a test is how you observe the highlight from outside; if cmdk renames it, those tests go red loudly, which is the cheap failure rather than the expensive one. Happy to be argued out of it.

## Verification

- Rebased onto `8f230e8`; `dist/` rebuilt.
- `platform-bible-react`: 1535 tests across 165 files pass.
- `find.component.test.tsx`: 49 pass.
- `npm run typecheck` clean; Prettier, ESLint and cspell clean on every changed file.
- E2E not run locally (needs a running app). The helper change is the one you flagged, verified by inspection against the row's rendered text.
- Two caveats on my side: `node_modules` was not readable in the environment I worked in, so I could not read the Radix and cmdk internals you cited. I reproduced both behaviours in jsdom instead, and that is what the new tests pin.

katherinejensen00:
@Sebastian-ubs Yes, https://github.com/paranext/paranext-core/pull/2229 can be safely closed. Good idea!

e2e-tests/fixtures/scripture-editor-helpers.ts
katherinejensen00: Confirmed, and it would have taken out all 15 call sites.

Replaced \b with (?!\d), which asserts what the anchor was actually for: the number must have ended. Mark 4(?!\d) still accepts "Mark 4:1MRK" and still rejects "Mark 40:1MRK", and it no longer depends on what character happens to follow.

Docstring corrected — the row renders the book name via getLocalizedBookName, and the app under test runs in English, which is why an English name is still the right thing to pass.

I moved the sibling helper's regexes to the same anchor rather than leaving them passing by luck. Heads up that the spec moved on main since you reviewed: tests/navigation-history/ to tests/isolated/navigation-history/, with a new isolated fixture and extra call sites (JHN 3, ACT 2, REV 1). The fix is applied to main's version, all 12 anchors.

lib/platform-bible-react/src/components/advanced/recent-searches.component.tsx
katherinejensen00: Reproduced in jsdom exactly as described — the tooltip comes back with data-state="instant-open" after selecting a row, and the assertion you suggested is the test I used.

Fixed by controlling the tooltip's open state and swallowing the single focus-restore open that follows a close. A later hover or Tab opens it normally, so focus still returns to the trigger and keyboard users lose nothing — which is why I did not reach for onCloseAutoFocus prevention.

One caveat on my side: node_modules was not readable in the environment I worked in, so I could not check the Radix sources you cite. I took the mechanism from your description and verified the observable behaviour in jsdom instead, which is what the test pins.

There is a related note in replace.spec.ts about the tooltip outliving the Find panel's close and swallowing a later click; that workaround still holds either way, so I left it.

lib/platform-bible-react/src/components/advanced/recent-searches.component.tsx
katherinejensen00: Confirmed, and I generalised it — the same shape reaches several strings this PR made visible, including the six new quick-nav titles.

New resolveLocalizedString(value, fallback) in platform-bible-react/src/utils/localization.util.ts treats /^%[^%]*%$/ as not-yet-localized; RecentSearches uses it for both ariaLabel and groupHeading. The bracketed character class is deliberate so text like "50% complete" or a %s of %s template is not swallowed.

The reason the usual localizedStrings?.[key] || 'Default' idiom silently fails to fall back is worth naming: the seed is key-to-key, and a key is a non-empty string, so || never fires. That is a repo-wide trap, not just this component.

find.component.test.tsx:891 now asserts the resolved English name rather than the key — and I added a case that the localized value wins once the strings arrive, so the fallback cannot quietly become permanent.

lib/platform-bible-react/src/components/advanced/recent-searches.component.tsx
katherinejensen00: Confirmed. The ADR arguing the role change on accessibility grounds while dropping the programmatic name is a fair thing to call out.

DropdownMenuLabel now takes a useId and DropdownMenuContent an aria-labelledby pointing at it, so the visible heading is the accessible name rather than a second, parallel string. The list announces as "Recent" instead of an unnamed menu of N items.

Test added: findByRole('menu', { name: 'Recent' }). It was red before the fix, so the regression is covered rather than just repaired.

lib/platform-bible-react/src/components/advanced/book-chapter-control/book-chapter-control.component.tsx
katherinejensen00: This is the one I have not changed, and I would rather have your read than decide it unilaterally.

It is a deliberate product call rather than an oversight: in the books view the grid is a preview of the typed reference, the top-match row displays "Mark 7:1", and Enter submits what the row displays. Making Enter drill into the verse grid means the row can display a reference that Enter does not submit, which is the invariant the rest of this work is built on.

You are right that pointer and keyboard now disagree for every getEndVerse consumer. My instinct is to close that by making the click consistent with Enter rather than the reverse, but that is a larger behavioural change than I want to make on review feedback. Say which way you want it and I will do it.

Both stale artifacts are fixed regardless: the test comment no longer claims Enter routes through cmdk's onSelect, and the catalog entry no longer says the books view "has no grid" — it now describes the preview grid and says Enter submits the previewed reference while Space stays an ordinary character.

lib/platform-bible-react/src/components/advanced/book-chapter-control/book-chapter-control.component.tsx
katherinejensen00: Correct, and the "effectively dead" observation is the sharpest part of it — the item half could never be reached because arrows never got the user onto an item.

Both roles are now in the single bail at the top of the handler, for exactly the reason you give: Radix focuses the content on open and only moves onto a row once the user arrows.

Test: with the list open and a preview grid on screen, Enter neither calls handleSubmit nor closes the picker. It was red before the fix — and it stayed red after only the capture-handler bail, which is how I found the cmdk bubble path described in the arrow-key thread.

lib/platform-bible-react/src/components/advanced/book-chapter-control/book-chapter-control.component.tsx
katherinejensen00: Real. Fixed with one guard at the top of handleCommandKeyDown rather than a check per branch:

if (eventTarget?.closest('[role="menu"], [role="menuitem"]')) return;
so Enter, Space, the arrows, letters/digits and Backspace all stay with the list.

That alone was not enough, and this is the part worth knowing: with our capture handler out of the way, the keydown still bubbled up the React tree to cmdk's Command root, which drilled into the verse grid. So RecentSearches now also stops keydown propagation out of its DropdownMenuContent — while that list is open it owns the keyboard, and it is documented as opening inside other surfaces.

handleQuickNavButtonKeyDown's ArrowUp/Down path is live again too: in the books view the arrow branch now yields when focus is on a header button, so its comment is true once more.

Coverage added — book-chapter-control.component.test.tsx now has a describe block that renders recentSearches, covering arrow navigation onto the first row and the Enter case in the next thread.

lib/platform-bible-react/src/components/advanced/book-chapter-control/book-chapter-control.component.tsx
katherinejensen00: Correct — it is the other half of the same condition, and framing it that way is what made the fix obvious.

The arrow branch now bails when the list is unmounted, before any grid resolution runs. isCommandListHidden is also part of the shared preview-grid memo mentioned in the ping-pong thread, so the effect and the key handler cannot disagree about whether a grid is on screen.

Test: after quick navigation hides the list, ArrowDown leaves focus on the search input instead of being swallowed — which also exercises the handleQuickNavButtonKeyDown focus-restore you noted was unreachable.

lib/platform-bible-react/src/components/advanced/book-chapter-control/book-chapter-control.component.tsx
katherinejensen00: Confirmed, including the second-order effect — the silent retarget of what Enter submits is the worse half of this.

I did not hand the horizontal arrows to the caret unconditionally, because arrow-driving the preview grid is deliberate and two existing tests exercise it. They follow the combobox rule instead: a horizontal arrow belongs to the caret while the caret can still move that way, and reaches the grid only once it cannot.

So mat 12:15 + ArrowLeft moves the caret and leaves the highlight on 15; ArrowRight at the end of the query still steps the grid. Both directions are now pinned by tests, and the caret position is asserted directly rather than inferred.

It is RTL-aware: which arrow walks toward the start of the text is a property of the layout, not of the key, so the check flips with direction.

Home/End are unchanged — still forwarded to the input.

lib/platform-bible-react/src/components/advanced/book-chapter-control/book-chapter-control.component.tsx
katherinejensen00: Confirmed. Fixed at the source rather than by moving the row out of CommandList.

The fight only exists because cmdk moves the highlight under the pointer, and the design here already says hover and the keyboard highlight are different things — every book row and grid cell styles hover separately and suppresses data-selected. So the Command root now sets disablePointerSelection, which is cmdk's own public prop for exactly this, and the seeding effect has nothing left to correct. That also makes the styling and the behaviour say the same thing, which they previously did not.

Worth recording because it cost me an attempt: declining the change in onValueChange does not work. cmdk syncs the controlled value prop into its store only when the prop changes, so refusing the change leaves cmdk's internal highlight on the row while our state says otherwise — data-selected lands on the wrong element. Five tests caught it.

I also hoisted the preview-grid computation out of the effect into a memo that the effect and the key handler share, so they cannot disagree about whether a grid exists — that disagreement was the root of the isCommandListHidden finding too.

lib/platform-bible-react/src/components/advanced/book-chapter-control/book-chapter-control.component.tsx
katherinejensen00: Confirmed, and the Genesis 1:1 observation is the one that decided it — the default state of a freshly opened project being the state with no explanation is not acceptable.

title is now set only while the button is disabled. A disabled arrow keeps native hover text; an enabled one takes the Radix tooltip and does not render two bubbles.

On the vertical chevrons: I have left those alone. That is a design choice rather than a defect, and I would rather it be revisited as a design question with UX than folded into a review-fix pass. Happy to raise it separately if you think it should change.

lib/platform-bible-react/src/components/shared/book-item.component.tsx
katherinejensen00: Confirmed, and I took your suggested fix — it belongs in the constant, not at either call site.

LIST_ITEM_KEYBOARD_FOCUS_RING now carries a forced-colors: outline alongside the ring, using the Highlight system colour so it tracks whichever high-contrast theme the user has chosen, with a negative outline-offset to match ring-inset. Both consumers pick it up unchanged.

You are also right that nothing in the suite would catch this — the tests assert the attribute, not a painted highlight, and I have not added a test that would, since asserting computed styles under an emulated forced-colors media query is not something this harness can do today. The comment on the constant records why the outline is there so it does not get tidied away.

lib/platform-bible-react/src/components/shared/book-item.utils.ts
katherinejensen00: Confirmed, and the observation that the current safety is incidental — coming from the very cmdk internal the ADR claims to have stopped relying on — is the part that made this worth fixing properly rather than special-casing two book ids.

parseChapterFromItemValue now round-trips its candidate through chapterItemValue and accepts only a byte-identical match, which is the cellNamedBy rule you pointed at. parseVerseFromItemValue inherits it by requiring the part before the colon to be a chapter cell.

I also removed generateCommandValue's chapter parameter. It had no production caller, and it was the only other way to spell a book row that looks like a chapter cell — so removing it closes the same hole from the other end.

Tests are parameterised over both PS2 and PS3, asserting a book row parses as neither, plus the reverse case that genuine PS2/PS3 chapter and verse cells still round-trip. Neutering the round-trip check turns them red.

katherinejensen00 (commented):
> @katherinejensen00 made 13 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2750)* status: 0 of 70 files reviewed, 13 unresolved discussions (waiting on mattgetgen).

lib/platform-bible-react/src/components/advanced/book-chapter-control/book-chapter-control.component.tsx
katherinejensen00: Correcting myself here — my earlier summary comment said this was fixed, and the first fix did not work.

I put title back on the button, gated to the disabled state. That cannot fire: button.tsx's base variants include tw:disabled:pointer-events-none, so a disabled Button is not hit-tested at all and the browser never shows its native tooltip. Your diagnosis was right and my fix was aimed one level too low — the problem is not only that a disabled control dispatches no events, it is that this particular control is not even a pointer target.

What is in now: while disabled, the button is wrapped in a <span title={title}> that is itself hit-tested, so the native tooltip has something to hover. Enabled buttons keep the Radix tooltip and are not wrapped, so there is no duplicate bubble. Genesis 1:1 — your example, and the one that decided it, since that is the default state of a freshly opened project — now explains both greyed arrows.

Worth flagging that this is verified by reading the CSS and the Radix/Chromium behaviour you described, not by hovering a real disabled button: it is a pointer-hover behaviour with no assertion available in jsdom, and I could not run the app in this environment. If you can hover a disabled arrow at Genesis 1:1 when you next have it running, that is the check I could not perform.

The vertical chevrons I have deliberately left alone. That is a design decision rather than a defect, and I would rather it be revisited with UX as its own question than changed inside a review-fix pass.

katherinejensen00 (commented):
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2750)* status: 0 of 70 files reviewed, 13 unresolved discussions (waiting on mattgetgen).

Sebastian-ubs:
Thank you so much Katherine. I spotted following issues
1. my fix had ruled out the option for 2 focus indicators. Focusing a button should remove focus from the list. Now I still/again see 2 focus rings
    Screenshot 2026-09-08 132054
2. the arrow buttons are a button group with rounded borders only at the edge, this looks wrong when tabbing through them - make all rounded
    (see above)
3. hovering the currently active chapter makes it hardly visible.
    Screenshot 2026-09-08 132221
4. pressing TAB on the chapter or verse selection page unexpectedly closes the whole dropdown
5. my fix was based on some older code that did not yet have the "Select chapter" text. Now when typing a match we get 2 rows, we should fix this to only be one. Need to think how to best do it. Maybe just to not show the "Lamantations    Select chapter" row in case a unique match row appears.
    image
6. down arrow does not consistently move cursor down to chapter selection when text is typed into the bcv input field, instead right and left arrow still apply to the text, even after a cursor up/down

lib/platform-bible-react/src/components/advanced/book-chapter-control/book-chapter-control.component.tsx
mattgetgen: Ran the check you asked for. **The fix works.**

Built `b884835` and ran the app with CDP, navigated the toolbar picker to Genesis 1:1, and inspected the disabled **Previous chapter** arrow. A screenshot would prove nothing here — native `title` tooltips are OS-drawn chrome and never appear in one — so I hit-tested instead, which is the actual mechanism:

| Check | Result |
| --- | --- |
| `button.disabled` | `true` |
| `getComputedStyle(button).pointerEvents` | `none` |
| Wrapped in `<span title>` | `true`, title `"Previous chapter"` |
| `document.elementFromPoint()` at the button's center | returns the **span**, carrying `title="Previous chapter"` |

That last row is the conclusive one: the element under the cursor is one with a `title`, which is exactly the condition for Chromium to draw its native tooltip. Your diagnosis and your fix both hold.

The enabled path is clean too — hovering **Next chapter** produced exactly one `role="tooltip"` reading "Next chapter", and that button has no `title` of its own and no span wrapper. No duplicate bubbles. Both titles resolved to real text rather than `%webView_bookChapterControl_previousChapter%`, so `resolveLocalizedString` is doing its job on this path.

### One correction to the premise — mine, not yours

"At Genesis 1:1 both Previous chapter and Previous verse come back disabled" was wrong, and it originated in my review, not your response. Only **Previous chapter** is disabled there. **Previous verse** is enabled and navigates to **Genesis 1:0** — which I confirmed by clicking it.

That is not a regression, and nothing here needs changing: `book-chapter-control.navigation.ts:37-41` documents that the quick-nav buttons deliberately pass no `ScriptureBounds`, so verse navigation keeps its pre-versification floor-at-0 behavior, with PT-4379 tracking the versification-aware version. It does slightly weaken the original argument — a freshly opened project shows one unexplained greyed arrow, not two — but the fix is still correct and still needed for that one.

### The three library claims you had to take on trust

You flagged that `node_modules` was unreadable in your environment. I could read it, so for the record:

1. **`disablePointerSelection` is real** — cmdk 1.1.1 exposes it. It is the right public fix, and better than the "move the row out of `CommandList`" workaround I suggested.
2. **`Button` does carry `tw:disabled:pointer-events-none`** (`button.tsx:22`), confirming why the first `title` attempt could never fire.
3. **Your `stopPropagation()` on `DropdownMenuContent` does not break Escape** — this one you could not have checked, and it was the real risk in that change. Radix registers Escape via `useEscapeKeydown`, which is `document.addEventListener('keydown', handler, { capture: true })`. Capture phase on the document fires before the event reaches your bubble-phase handler, so dismissal is unaffected.

Also confirmed locally: 108 tests pass across the five review-relevant files, including the new caret and "yields keys it does not own" cases.

mattgetgen (approved):
Approving. This supersedes my earlier `CHANGES_REQUESTED` on `3bb3286`.

Every finding from that review is addressed on `b884835`, and the response held up everywhere I checked it. Verification I ran rather than took on trust:

- **The CI blocker is fixed.** `(?!\d)` is the right anchor and asserts what `\b` was there for. Checked against all five reference shapes, including that `Mark 4` still rejects `"Mark 40:1MRK"`.
- **The disabled-arrow tooltip works in the running app** — built this commit, ran it, and hit-tested the greyed arrow at Genesis 1:1. Details in [the thread on `:1304`](https://github.com/paranext/paranext-core/pull/2750#discussion_r3959581359), including a correction to my own premise: only *Previous chapter* is disabled there, not both arrows.
- **The three library claims you could not check** — `disablePointerSelection` exists in cmdk 1.1.1, `Button` really does carry `tw:disabled:pointer-events-none`, and your `stopPropagation()` on the menu content does **not** break Escape (Radix listens with `capture: true` on `document`, ahead of your handler). That last one was the genuine risk in the change and it is clear.
- **Tests**: 108 pass across the five review-relevant files; the full unit project is green.
- **The ADR corrections are honest.** `drawer.tsx` really is still on `tw:z-50` and you recorded it as a known violation instead of claiming the sweep was complete. `Z_INDEX_MODAL + 10` is gone, `menubar.tsx` is on the constant with the required `// CUSTOM:` annotations, and the item you deferred carries an accurate `TODO(PT-4345)`.

Two things I want to call out rather than let the approval bury.

**1. `:718` is still open and this approval does not settle it.** You asked which way to take Enter-vs-click on a chapter cell, and that is a product call I am not making unilaterally. Your reasoning is sound — the row displays what Enter submits, and that invariant is worth more than matching the pointer — so "make the click consistent with Enter" looks right to me, as a separate change. I will answer properly in that thread. Do not read the approval as agreement to leave the divergence permanently.

**2. On the `data-selected` assertions (item 6), you were more right than I was.** I checked: `spaceSelectsHighlightedItem` in `command.tsx:134-140` still does `querySelector('[cmdk-item][data-selected="true"]…')` plus a synthesized `.click()`, and this control opts into it. So production *does* still read that attribute, and your ADR correction says so. Reading it in a test is the cheap failure, exactly as you argued. Dropping it.

Two caveats on what this approval covers. E2E has not actually run — the helper fix is verified by inspection and by checking the regex against every call-site shape, but nothing has exercised those 15 sites against a live app, so let CI be the real gate there. And the pointer-hover behaviour is verified by DOM hit-testing rather than by a human eye on a rendered tooltip bubble; the mechanism is proven, the pixels are not.

Nice work on catching the `ScopeSelector` regression that my own suggested fix caused, and on flagging your first `title` attempt as broken rather than letting it stand.

lib/platform-bible-react/src/components/advanced/book-chapter-control/book-chapter-control.component.tsx
mattgetgen: Go with your instinct: **make the click consistent with Enter**, not the reverse.

Your reasoning is the right way round. "The row displays what Enter submits" is the stronger invariant — it is the one a user can actually see and predict, and the rest of this work is built on it. Letting Enter drill into the verse grid would buy pointer/keyboard symmetry by breaking the thing that makes the books view legible, which is a bad trade.

So the divergence closes by bringing the click up to the same rule, and I agree that is too big to fold into a review-fix pass. Please spin it out as its own work item rather than doing it here — this PR is approved as it stands, and I would rather that change get its own scrutiny (it touches `handleChapterSelect` for every `getEndVerse` consumer: the Scripture editor, `scope-selector` in three places, and `checklist.web-view.tsx`).

Worth noting in that follow-up: today a click on a chapter cell in the books view drills into the verse grid, so making it submit `chapter:1` instead is a user-visible behaviour change for anyone who has learned the current gesture — not just an internal consistency fix. That is an argument for doing it deliberately and separately, which is exactly where you landed.

Thanks for flagging it rather than quietly picking one. Nothing blocking here.

katherinejensen00: Done.

lib/platform-bible-react/src/components/advanced/book-chapter-control/book-chapter-control.component.tsx
katherinejensen00: Done.

katherinejensen00 (commented):
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2750)* status: 0 of 77 files reviewed, 13 unresolved discussions (waiting on mattgetgen).

jolierabideau:
Hi Katherine — I'm Claude, Jolie's coding assistant, posting on her behalf. Recording a decision here after the fact, because this PR turned out to be the one that fixed a defect tracked on a *different* ticket and nothing on the PR says so.

## This PR carried the fix for [PT-4553](https://paratextstudio.atlassian.net/browse/PT-4553) (Todd NTH-4, BCV chapter-hover alignment)

PT-4553's triage explicitly warned that it did **not** look like this PR contained an alignment fix — the title is "keyboard navigation, styling, and the z-index scale", and the diff reads as row-hover styling plus a chevron-column realignment. That warning was wrong, and it is worth knowing why it was easy to get wrong.

The defect: every chapter digit sat **exactly 12.0px left of the centre of its own highlight box**, and outside that box's left edge — so the highlight read as belonging to the neighbouring chapter. Cause was `CommandItem` rendering a trailing `<IconCheck>` unconditionally: invisible (`opacity-0`) but fully laid out at `size-4` plus the flex `gap-2`, i.e. 24px of phantom content in a 16px box, with `justify-center` splitting the overflow — (16+8)/2 = 12px.

Your one-line `tw:[&>svg]:hidden` on the grid cell is the entire fix. It was filed under a different description — the comment calls it hiding "CommandItem's own trailing check icon (a multiselect affordance this grid doesn't use)" — which is accurate but doesn't mention alignment, so nobody connected the phantom icon to the 12px shift. Measured before and after: **12.0px → 0.0px, digit inside its box.**

Decision recorded: **this PR is the merge that closes PT-4553's fix half.** #2229 was closed as superseded (its author's call, since confirmed by an audit of its payload against `main`), and #2239 was found to be orthogonal and stays open on its own merits.

## Two follow-ups, both on Jolie's side, neither asking anything of you

1. **A regression test** — [#2824](https://github.com/paranext/paranext-core/pull/2824) adds a Storybook browser-mode story measuring the geometry, since the fix is a bare utility class and jsdom performs no layout. That matters because the defect arrived via `apply-shadcn-preset-destructive` in `a61ca3b913c`, a script that will be run again: the next baseline could re-add a laid-out child to `CommandItem` with nothing failing.

2. **A heads-up about the Tab tests you added here.** `BookChapterControl keeps Tab inside the picker` passes for a weaker reason than it appears to. `@testing-library/user-event` computes the tab ring itself and wraps within the document, so the ring cycles inside the popover whether or not the component traps the key — verified by collapsing `event.shiftKey ? tabStops[0] : lastStop` to `lastStop` and re-running, green in jsdom **and** in browser mode. The trap logic is right; it just isn't what those tests are proving. Flagging it so it isn't relied on as proof later. Genuinely covering it needs the Playwright e2e harness.

Also: thanks for the `z-index.ts` repair, the SCSS twin and the tiers rule — WI-6 ([PT-4535](https://paratextstudio.atlassian.net/browse/PT-4535)) now consumes that scale as baseline rather than having to introduce it.

[PT-4553]: https://paratextstudio.atlassian.net/browse/PT-4553?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PT-4535]: https://paratextstudio.atlassian.net/browse/PT-4535?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2630, opened 2026-08-03
PT-4059: Show locally-installed resources (non-DBL + all types) in Text Collection

katherinejensen00:
https://github.com/user-attachments/assets/a794ecc2-f04e-4459-9ed3-05cc22ab8ba2

## Summary

**Purpose**: Make locally-installed non-DBL resources (TNN, TND, HBK commentary series, VULGP83, etc.) discoverable in Text Collection's Get Resources dialog and displayable in the Text Collection grid.

The Resource Viewer dropdown previously only listed resources explicitly linked to the current project, and only knew about DBL-catalog entries. Locally-installed non-DBL resources — installed as `.p8z` bundles outside the DBL download flow — were invisible. This PR fixes that gap in two layers:

1. **C# layer** — `ParatextPublishedProjectDataProviderFactory` switches from `ScriptureOnly` to `AllAccessible` enumeration so `.p8z` bundle resources are included in the published-project list.

2. **Extension layer** — `getLocalNonDblResources` returns synthetic `DblResourceData` entries for projects not in the DBL catalog. Commentaries (TNN, TND, HBK series) are typed `CommentaryResource` via a `LOCAL_COMMENTARY_UIDS` whitelist mirroring the C# `CommentariesWhiteList`. Synthetic entries use `dblEntryUid === projectId` as a marker so callers create a `ProjectReference` (resolvable by project folder) instead of a `DblResourceReference`.

Also adds the `downloaded` option to the Scripture Text Grid View Options data layer (default-off, ready for PT-4171) and a `resolveLongName` parameter to display full resource names.

## File guide

| File | What it does |
|---|---|
| `ParatextPublishedProjectDataProviderFactory.cs` *(modified)* | Switches to `AllAccessible` so `.p8z` resources appear in the published-project list |
| `get-local-non-dbl-resources.utils.ts` *(new)* | Pure utility: filters non-DBL projects from metadata and maps to synthetic `DblResourceData`; exports `LOCAL_COMMENTARY_UIDS` whitelist |
| `resource-reference.utils.ts` *(modified)* | Adds `isNonDblResource()` — single named home for the `dblEntryUid === projectId` synthetic-marker check |
| `downloaded-resources.utils.ts` *(new)* | Fetches all locally-installed USJ_Chapter projects via PAPI and resolves display metadata. Uses `Promise.allSettled` so one bad project doesn't blank the list |
[...]

PR #2271, opened 2026-05-14
Add getCachedResources command with background fetch and periodic refresh

captaincrazybro:
- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2271)

# Code Review Summary

**Branch**: feat-cached-resources

**Base**: origin/main

**Date**: 2026-05-20

**Review model**: Claude Sonnet 4.6

**Files changed**: 6

## Overview

These changes migrate the DBL resources data layer from a live subscription (`useData(...).DblResources`) to a command-backed cache (`platformGetResources.getCachedResources`). A new `getCachedResources` command is registered in the `platform-get-resources` extension. On activation, the extension starts a background fetch that retries up to 10 times, persists results to extension user data, and then refreshes on a 12-hour interval. All four call sites — `get-resources.web-view.tsx`, `home.web-view.tsx`, `model-text-panel.web-view.tsx`, and `resource-picker.dialog.tsx` — are updated to use `usePromise` + the new command. The `getCachedResources` implementation also reconciles the `installed` flag against live project metadata on each call, so the UI reflects out-of-band install/removal without a full network refetch.

## API Changes

- **New command**: `platformGetResources.getCachedResources(): Promise<DblResourceData[] | undefined>` — returns DBL resources from in-memory cache; triggers a fetch if cache is empty; returns `undefined` if no fetch has succeeded yet
- `DblResourceData` import added to the `papi-shared-types` augmentation block in `platform-get-resources.d.ts` to support the new command's return type

### Critical — Must address before merge

None.

### Important — Should address before merge

- [x] **`model-text-panel.web-view.tsx:256` — `isLoadingResources` in zero-state guard caused flash** _(fixed by author: condition restructured so `isLoadingResources` only gates the empty-state path, not the editor render)_
- [x] **`platform-get-resources.d.ts` — blank line between JSDoc and command declaration broke JSDoc association** _(fixed during review: blank line removed)_
- [ ] ~~`main.ts` — `hasFetchStarted` and `cachedResources` not reset in `deactivate()`~~ _(author explains: the extension host process fully restarts on extension reload, resetting module-level state. Not an issue in practice.)_
- [x] **`main.ts:39` — `cachedResources = resources` assigned without null guard; unexpected network value could corrupt in-memory cache** _(fixed during review: wrapped in `if (resources) { ... }`)_
- [ ] ~~`main.ts` — new caching/reconciliation logic has no unit tests~~ _(author dismissed: tests not required for this logic at this time)_

[...]

PR #2751, opened 2026-09-01
Fix Get Resources showing Update after a resource is updated

captaincrazybro:
# Code Review Summary

**Branch**: `fix-get-resources-stale-update-flag`

**Base**: `origin/main` (`bc056ca9284`) · **Head**: `d480cc6ad53`

**Date**: 2026-09-10 (round-4 rework; originally 2026-09-01)

**Files changed**: 8 (5 modified, 3 new)

## Verified end to end

**Confirmed against a real resource with a pending DBL update, on `d480cc6ad53`, in
paratext-10-studio.** Clicking Update takes the row to "Installed" in the *open* dialog — no
reopen, no restart. Also exercised: first-time install, uninstall, install failure, opening the
dialog, the resource picker, Home, the scripture text grid, Share Layout, and the cross-surface
case (install from the text grid, then open Get Resources — no stale badge).

**An earlier revision of this section claimed the same thing and was wrong.** The check behind it
did not distinguish "the row cleared in place" from "the row cleared after the dialog was
reopened", and it was in fact the latter — round 4's finding 1 was correct, and the defect was
still present at that point. It is fixed now (`b0780a2`, `c1b94261`), and the re-check above was
run specifically against that distinction.

## Overview

Get Resources kept showing an "Update" button on a resource that had already been updated, for the
rest of the session. `platformGetResources.getCachedResources` is the only data source the Get
Resources page, Home, the resource picker, Share Layout, the scripture text grid and
`use-dbl-resource-catalog.hook` read — six consumers. Its sync reconciled just the `installed` flag
against local project metadata; it never recomputed `updateAvailable`, and its one cache-rewrite
branch fired only when `installed` flipped. Updating an already-installed resource leaves
`installed` true on both sides, so the stale flag survived. The C# provider does fire
`SendDataUpdateEvent(DBL_RESOURCES, …)` after an install, but nothing has subscribed to that data
type since the cache replaced the front end's `useData` subscription (zero `.DblResources(` call
sites repo-wide), so that event refreshes nothing.

The fix adds a no-network C# provider function, `recomputeDblResourcesUpdateStatus`, which
re-evaluates `InstallableResource.IsNewerThanCurrentlyInstalled()` over the catalog snapshot already
in memory. `syncInstalledFlags` applies it alongside the `installed` check; because that sync is
single-flighted by `ensureInstalledFlagsSynced` and runs in the background, the recompute happens at
most once at a time and never on a path a caller waits on. The reconciliation logic moved into a
pure, unit-tested util. A resource absent from the backend's answer keeps its cached value, so a
not-yet-loaded catalog or a busy provider degrades to the previous behavior rather than guessing.

Cost is bounded by gathering the installed DBL uids in **one** pass over the project collection and
consulting ParatextData only for entries in that set. `InstallableResource.ExistingScrText` is a
computed property with no backing field that enumerates the whole collection on every access, so
asking each of ~1850 catalog entries whether it is installed would cost ~1850 full scans — twice
over for an installed entry (once via `Installed`, once inside `IsNewerThanCurrentlyInstalled`), and
gating the loop on `Installed` does not help because that property is the same lookup. An entry
outside the set is reported as having an update available, which is exactly what
`IsNewerThanCurrentlyInstalled` returns for anything uninstalled.

Three review rounds ran against this branch: a four-pass analysis (api / style / compliance / ux), a
follow-up correctness pass, and a 22-finding review from @katherinejensen00 that this revision
answers in full.

## API Changes

- `extensions/src/platform-get-resources/src/types/platform-get-resources.d.ts`: added
  `recomputeDblResourcesUpdateStatus: () => Promise<DblResourceUpdateStatus>` to the exported
  `IDblResourcesProvider` type, and a new exported type `DblResourceUpdateStatus`
  (`{ [dblEntryUid: string]: boolean | undefined }`). Additions only — nothing removed or narrowed.
- `c-sharp/Projects/DigitalBibleLibrary/DblDownloadableDataProvider.cs`: `DblResourcesDataProvider`
  registers a new function `recomputeDblResourcesUpdateStatus` returning `Dictionary<string, bool>`,
  carrying an explicit `[NetworkTimeout]` rather than inheriting the 30 s default. The method and
  the extracted `ProjectUpdateStatus` are `internal` (not `private`) so `c-sharp-tests` can exercise
  them via the existing `InternalsVisibleTo`.
- `lib/platform-bible-react/`, `lib/platform-bible-utils/`, `lib/papi-dts/papi.d.ts`: no changes.
- Extension-internal (not a declared API surface): new module `resources-cache.util.ts` exporting
  `reconcileCachedResources` and `ReconciledCachedResources`.

## Findings

Rounds 1 and 2 (16 findings) were resolved before the third round; their detail is in this PR's
earlier revisions and in `adr-dbl-cache-recompute-on-read`. The list below is round 3.

### Round 3 — @katherinejensen00, 22 findings

**Fixed (14):**

- [x] **1. Branch was 13 (by then 20) commits behind main, which had rewritten this function —
      Critical.** Rebased. Main's PR #2630 had restructured `getCachedResources` into a
      single-flight background `syncInstalledFlags()` and changed the project filter to
      `platform.base`; the fix is re-derived on top of that rather than reverting main's mutex
      guarding. The decision-log conflict was resolved as a slug-keyed insert, not "keep both
      sides".
- [x] **2, 7. The 250 ms gate could not deliver its stated guarantee, and the ADR argued against
      it.** Dissolved by the rebase, as the reviewer predicted. The one caller now sits inside
      `ensureInstalledFlagsSynced`'s single flight, so concurrent sibling recomputes are
      structurally impossible and no caller awaits the result. The wait bought nothing, so the gate
      is non-waiting again — which is what the decision log already described.
- [x] **3, 5. ~1850 live `ScrTextCollection` scans per call, with no behavioral coverage.** Loop
      extracted to `ProjectUpdateStatus(resources, installedDblIds)` with the one-pass change
      described above. The extraction made it coverable: `ExistingScrText` is `public virtual` and
      `InstallableResource` has a protected parameterless constructor, so test doubles need no REST
      client, no catalog and no live collection.
- [x] **6, 14. Single-element tests hid a class of bug; the `projectId` comment asserted an
      invariant the code did not hold.** 14 was a real defect, not just a comment — a row cached
      `{installed: true, projectId: ''}` kept `''` indefinitely. `projectId` is now recomputed
      unconditionally and included in change detection, and main's empty-`dblEntryUid` guard is
      carried into the util.
- [x] **8, 21. One throwing resource discarded the entire map; duplicate-uid handling disagreed with
      `FindResource`.** Per-entry try/catch, and `TryAdd` for first-wins.
- [x] **9. "Never waits" was inaccurate on the public surface.** Reworded, with no number restated
      in TypeScript.
- [x] **11, 13. Per-method `<summary>` blocks against convention; the wire test's suffix match.**
      One fixture-level summary, and the test now pins the full registered name via
      `Client.IsHandlerRegistered`.
- [x] **15. "unlike every other method here" was inaccurate.** Now names the methods it means.
      _(Partial correction to the finding: `InstallDblResource` and `UninstallDblResource` do load
      the catalog, via `EnsureResourcesLoadedCore()` — it is their private `*Core` helpers that do
      not. The claim was still wrong as written, because `IsGetDblResourcesAvailable` does not
      load.)_
- [x] **18. The only wire method without `[NetworkTimeout]`.** Added. Note the value is not "just
      above the gate timeout" as suggested — a non-waiting gate means the bound has to cover the
      recompute itself, which the review measured at up to 713 ms.
- [x] **19. Fast path before `Task.Run`.** Applied, and the `_hasFetchedResources` comment now
      documents that it may be read without the gate as a one-way hint (it is only ever set `true`,
      in `FetchResourcesCore`) rather than silently violating its own stated invariant.
- [x] **22. The one combination that decides the design question.** Settled rather than merely
      pinned: `updateAvailable` is now cleared for any row reconciled as not installed. The backend
      reports `true` for everything uninstalled — nothing installed trivially fails its "is the
      installed copy the newest" test — so taking that at face value persisted a flag that
      described nothing. The flag now means what the list renders it as.
- [x] **12, 16, 17. Decision-log corrections.** Six consumers not three; the garbled `` `adr` ``
      -adjacent PT-4268 cross-reference; and the unrelated failed-install NRE removed from
      Consequences. Entry re-inserted in byte-order slug position per the log's own rule.

**Closed rather than actioned (3):**

- [ ] **4. Five of six callers pay for a value they never render.** The premise — each blocking the
      full 250 ms — is gone now that the recompute runs inside the fire-and-forget background sync,
      and finding 3 removed the CPU cost. An opt-in parameter would now add a cross-language
      contract knob for no measurable gain.
- [ ] **10. The degradation path is untested.** Left to the util's coverage. Post-rebase the
      extraction is larger than costed, since `getLocalProjectMetadata`'s retry logic and the
      `!hasResourceProjects` early return now share the same function.
- [ ] **20. Two independent round trips run sequentially.** Skipped deliberately: the latency is on
      a background path nobody waits on, and parallelizing would fire the recompute even when the
      `!hasResourceProjects` early return discards it — which includes every startup before the C#
      PDPF registers.

**Deferred as pre-existing (1):** finding 17's failed-install `NullReferenceException` is real but
unrelated to this change. Removed from the ADR; needs its own ticket.

#### Shared Regions Modified

None. No `#region shared with` markers exist in any changed or added file.

#### Extension Config Changes

None. The only files touched under `extensions/` are source and type-declaration files inside
`extensions/src/platform-get-resources/src/`.

## Positive Observations

- Extracting the reconciliation into a pure function made previously untestable `main.ts` logic
  unit-testable, following the established `<name>.util.ts` + `<name>.util.test.ts` convention.
- `reconcileCachedResources` preserves object identity for unchanged entries, and a test pins it
  (`expect(resources[1]).toBe(laterUnchanged)`) — the contract `main.ts` relies on to avoid
  rewriting the cache file unnecessarily.
- The degradation design is consistent across the process boundary and documented on both sides: C#
  returns an empty map rather than blocking; TS treats a missing key as "unknown" and keeps the
  cached value.
- The one-pass installed-uid set reaches the same answer ParatextData does by construction, rather
  than by an approximation that could drift: the value assigned to an entry outside the set is
  literally what `IsNewerThanCurrentlyInstalled` returns for anything uninstalled.
- No new user-visible strings and no new project-data write patterns, so localization has nothing
  outstanding and `SendReceiveWriteLockCoverageTests` is unaffected.

## Interview Notes

**Author's stated purpose**: fix Get Resources showing "Update" after a resource has already been
updated, reported as happening in Power mode.

**Key context for the reviewer:**

- **This is not Power-specific.** There is no interface-mode gating anywhere in
  `platform-get-resources`; the defect reproduces wherever the Get Resources page is reachable.
- **It is a regression from the caching layer.** Before PR #2271 the web view subscribed via
  `useData(...).DblResources(...)`, so the post-install `SendDataUpdateEvent` re-ran
  `getDblResources` and recomputed the flag. The cache replaced the subscription without replacing
  the invalidation.
- **The design choice was recompute-on-read over invalidate-on-write**, recorded as
  `adr-dbl-cache-recompute-on-read`.
- **`updateAvailable` genuinely cannot be computed in TypeScript.** The author pushed back and it
  was re-verified: the DBL-side revision is dropped when C# projects `InstallableResource` into
  `DblResourceData`, and the *installed* revision is an entry name under `.dbl/revision/` inside the
  password-protected `.p8z`, read through ParatextData's zip file manager. `Revision` appears
  nowhere in `c-sharp/`. Sending both sides to TS would still need a C# call to read the installed
  half, so it removes no round trip while adding ~10 contract fields and a TS copy of a five-branch
  precedence chain that would drift when the ParatextData pin moves.
- **Correctness of the no-fetch approach rests on decompiled ParatextData, not observation.**
  `ExistingScrText` is a live `ScrTextCollection` lookup on every access, and `InternalInstall` calls
  `scrText.SetFileManager(null)` before overwriting the `.p8z`, so `DBLResourceSettings` is rebuilt
  from the new file. Read from the pinned 9.5.0.24 assembly (ILSpy) on 2026-09-04; ParatextData
  source is not available locally, so this wants re-probing when the pin moves. The ADR now records
  the version and date.
- **The one-pass predicate is deliberately narrower than `ExistingScrText`'s.** It matches only that
  property's main branch, omitting the null-`DBLEntryUid`-matched-by-name branch and the
  `SourceLanguageResource` fallback, because this provider serves only whitelisted DBL catalog
  entries — always uid-carrying, always `ResourceType.DBL`. Documented at the call site.

**Retracted claim — a failed install does *not* report success.** An earlier draft of this review
stated that `InstallDblResourceCore`'s `ScrTextCollection.IsPresent` check passes trivially for an
already-installed resource, so a failed *update* would be recorded as successful. That was wrong.
`IsPresent(ScrText)` resolves to `scrText.Guid != null && scrTextsById.ContainsKey(scrText.Guid)` —
it inspects `InstalledScrText`, not whether the resource exists on disk. `InstalledScrText` is
assigned only as the last statement of `InternalInstall`, after every failure path has returned, so
on a failed install it is `null`, `IsPresent(null)` throws `NullReferenceException`, and that
propagates through JSON-RPC to the web view's `.catch`. **Failures surface** — just as an NRE rather
than the localized message, which is finding 17's deferred ticket.

Related and confirmed correct as written: `Install()` returns `bool`, but the value is set true only
inside `foreach (… ProjectFiles("*.font"))` — it means "fonts were extracted", not "install
succeeded". The existing comment "we don't get any info telling if the installation succeeded or
failed" is accurate and discarding the return value is right.

## In-Review Quality Check

- **Extension tests**: 70 passed across 6 files in `platform-get-resources`; the reconcile util is
  at 14 cases, up from 9. The 5 added cases were mutation-verified — the `isChanged` latch, the
  `projectId` recovery, the empty-`dblEntryUid` guard, and the two `updateAvailable`-when-not-
  installed cases each kill exactly their own mutant.
- **C# tests**: **1724 passed, 0 failed, 6 skipped** (full suite). The 6
  `LocalParatextProjectsWatcherTests` failures reported in the previous round as pre-existing do
  **not** reproduce on this rebased tree.
- **C# mutation checks** on the new `ProjectUpdateStatus`: keying by `Name` instead of
  `DBLEntryUid.Id` → 5 red; dropping the installed-set short-circuit → 3 red; removing the per-entry
  try/catch → 1 red; the indexer instead of `TryAdd` → 1 red.
  - Known-surviving mutant, deliberately: removing the `!_hasFetchedResources` fast path (finding 19)
    leaves all 7 tests green. With `_resources` empty the loop is a no-op either way and the gate
    check immediately after returns the same empty map, so that fast path is a pure optimization
    with no observable behavior. The behavior it fronts — an empty result before the catalog loads —
    is pinned; a test for the dispatch itself would be timing-dependent.
- **C# build**: 0 errors (pre-existing warnings in unrelated test files only).
- **Lint**: `npm run lint` exits 0. One pre-existing warning (0 errors) in
  `unsubscriber-async-list.test.ts`, untouched here.
  - Note for a fresh worktree: lint initially fails on
    `Unable to resolve path to module '../../../release/app/buildInfo.json'`. That file is gitignored
    and generated; run `npx ts-node .erb/scripts/generate-dev-build-info.ts`, then delete
    `.eslintcache` (eslint runs with `--cache` and holds the stale verdict).
- **Formatting**: csharpier and prettier applied; nothing outside the change was reformatted.
- **Typecheck**: skipped per the author's standing preference (lint + build treated as sufficient).

## Suggested Review Focus

- [ ] **Recompute-on-read vs. invalidate-on-write** (`adr-dbl-cache-recompute-on-read`). Still the
      load-bearing design decision, and it collides with PT-4268's scope — that ticket's stated
      premise ("the provider's install path fires the event so the Get Resources UI refreshes") is
      already false and should be re-scoped against this entry.
- [ ] **The narrowed one-pass predicate.** It assumes every entry this provider serves carries a DBL
      uid and is `ResourceType.DBL`. That holds for the whitelisted catalog today; it is the
      assumption that breaks first if the provider's inputs ever widen.
- [ ] **The ParatextData pin.** The no-fetch correctness argument is read off the 9.5.0.24 assembly.
      Worth a look at whether the ADR's recorded version + date is enough of a tripwire, or whether
      this wants a test.
- [ ] **Finding 17's ticket** — a failed install surfaces as an NRE rather than the localized
      `%getResources_errorInstallResource_installationFailed%`. Pre-existing, removed from the ADR,
      not yet filed.
- [ ] **The two deferred pre-existing minors**: the `TODO(PT-4268)` code pointer at the dead
      `SendDataUpdateEvent` site, and the `projectId`-starts-with-`dblEntryUid` invariant duplicated
      in `use-commentary-marker-styles.hook.ts`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2751)

### DIFF
--- a/.context/standards/Architecture-Decisions.md
+++ b/.context/standards/Architecture-Decisions.md
@@ -498,11 +498,18 @@ step, no automation. Just a record.
 - **Decision:** Extend the read-path reconciliation rather than adding a write-path invalidation. A
   new no-network provider function, `recomputeDblResourcesUpdateStatus`, re-evaluates
   `InstallableResource.IsNewerThanCurrentlyInstalled()` over the already-loaded catalog snapshot,
-  and the front end's `syncInstalledFlags` applies it alongside the `installed` check. Because that
-  sync is single-flighted by `ensureInstalledFlagsSynced` and runs in the background, the recompute
-  happens at most once at a time and never on a path a caller waits on, however many views refresh
-  together. Resources absent from the result keep their cached value, so a not-yet-loaded catalog
-  or a busy provider gate degrades to the previous behavior instead of guessing. Skipping the
+  and the front end's `syncInstalledFlags` applies it alongside the `installed` check. The sync is
+  single-flighted by `ensureInstalledFlagsSynced`, so the recompute happens at most once at a time
+  however many views refresh together. Reads do not wait for it — `getCachedResources` answers from
+  the array it already has — which makes a read one refresh behind. That is invisible for
+  `installed`, whose caller already knows what it just did, but it is the whole defect for
+  `updateAvailable`: nothing else about an updated row changes, and there is no data-update event,
+  so the row keeps offering "Update" until the catalog is read a second time. A caller that has
+  just changed local state therefore awaits `refreshResourceFlags` and then re-reads, which is the
+  one path that does wait on the recompute. Resources absent from the result keep their cached
+  value, so a not-yet-loaded catalog or a busy provider gate degrades to the previous behavior
+  instead of guessing — including on that awaited path, where a contended gate leaves the stale
+  flag in place rather than blocking the user. Skipping the
   catalog fetch is sound because the DBL-side revision is the half that should stay fixed;
   ParatextData re-reads the *installed* revision on each call (`ExistingScrText` is a live
   `ScrTextCollection` lookup, and `InternalInstall` nulls the ScrText's FileManager before
--- a/.context/standards/Architecture-Decisions.md
+++ b/.context/standards/Architecture-Decisions.md
@@ -477,6 +477,80 @@ step, no automation. Just a record.
   the only copy a user receives. **Revisit** if this repository ever needs to publish a build to a
   public audience.
 - **Source:** the multi-agent review of #2654, finding 1.
+## adr-dbl-cache-recompute-on-read: The DBL resource cache recomputes derived flags on read, not on write
+
+- **Date:** 2026-09-01
+- **Status:** Accepted
+- **Context:** `platformGetResources.getCachedResources` is the only source the Get Resources page,
+  Home, the resource picker, Share Layout, the scripture text grid and
+  `use-dbl-resource-catalog.hook` read — six consumers. Its background sync reconciled `installed`
+  against local project metadata on every call but never recomputed `updateAvailable`, and its one
+  rewrite branch fired only when `installed` flipped — so updating an already-installed resource
+  left the row reading "Update" for the rest of the session. The C# provider does fire
+  `SendDataUpdateEvent(DBL_RESOURCES, …)` after an install, but nothing has subscribed to that data
+  type since the cache replaced the front end's `useData` subscription (zero `.DblResources(` call
+  sites repo-wide), so the event refreshes nothing. `updateAvailable` cannot be derived in
+  TypeScript: it compares the installed resource's DBL revision against the catalog's, and neither
+  number is reachable there. The DBL side is dropped when C# projects `InstallableResource` into
+  `DblResourceData`; the installed side is an entry name under `.dbl/revision/` inside the
+  password-protected `.p8z` bundle, read through ParatextData's zip file manager — no project
+  setting or metadata field exposes it, and `Revision` appears nowhere in `c-sharp/`.
+- **Decision:** Extend the read-path reconciliation rather than adding a write-path invalidation. A
+  new no-network provider function, `recomputeDblResourcesUpdateStatus`, re-evaluates
+  `InstallableResource.IsNewerThanCurrentlyInstalled()` over the already-loaded catalog snapshot,
+  and the front end's `syncInstalledFlags` applies it alongside the `installed` check. Because that
+  sync is single-flighted by `ensureInstalledFlagsSynced` and runs in the background, the recompute
+  happens at most once at a time and never on a path a caller waits on, however many views refresh
+  together. Resources absent from the result keep their cached value, so a not-yet-loaded catalog
+  or a busy provider gate degrades to the previous behavior instead of guessing. Skipping the
+  catalog fetch is sound because the DBL-side revision is the half that should stay fixed;
+  ParatextData re-reads the *installed* revision on each call (`ExistingScrText` is a live
+  `ScrTextCollection` lookup, and `InternalInstall` nulls the ScrText's FileManager before
+  overwriting the `.p8z`, so `DBLResourceSettings` is rebuilt from the new file). Verified against
+  the pinned ParatextData 9.5.0.24 assembly (ILSpy, 2026-09-04); source is not available locally,
+  so re-probe when that pin moves.
+- **Alternatives:**
+  - **Patch the cache entry after a successful install** — rejected, but not because the install is
+    unverified: `InstallDblResourceCore`'s `ScrTextCollection.IsPresent(InstalledScrText)` guard
+    inspects `InstalledScrText`, which `InstallableResource.InternalInstall` assigns only as its last
+    statement, so a failed install leaves it null and the guard throws rather than reporting success.
+    Rejected instead because the patch would have to be applied by every caller that installs — the
+    Get Resources web view, `platform-scripture-editor`'s install util, and the Send/Receive path
+    that PT-4268 describes, which cannot reach the provider at all — and because it only corrects
+    the one resource this client just installed, leaving flags that changed for any other reason
+    stale. Recomputing on read asks the source of truth instead of inferring from an action.
+  - **Re-fetch the catalog on install** — rejected: `FetchResourcesCore` is an unbounded network
+    download and would stall the list refresh. Subscribing to `DBL_RESOURCES` from the extension is
+    the same alternative in disguise, because `subscribe<data_type>` calls `get<data_type>` and
+    `getDblResources` fetches unconditionally.
+  - **Clear the flag optimistically in the web view** — rejected: leaves the cache wrong for Home and
+    the resource picker, and adds a second "caller must remember to notify" seam of exactly the kind
+    PT-4268 documents.
+  - **Send both revisions to TypeScript and compare there** — rejected: the installed half still
+    requires a C# call to read it out of the encrypted bundle, so it removes no round trip while
+    adding ~10 fields to the contract and a TypeScript copy of
+    `IsNewerThanCurrentlyInstalled`'s five-branch precedence chain (name, language,
+    `IsResourceProject`, `RequiresDBLCheck`, then revision / permissions checksum / manifest checksum
+    plus timestamp). That copy would drift silently when the ParatextData pin moves.
+- **Consequences:** Reads are authoritative for both derived flags, which makes the `DBL_RESOURCES`
+  data-update event dead weight rather than a missing link — PT-4268's stated premise ("the provider's
+  install path fires the event so the Get Resources UI refreshes") is already false, and whoever picks
+  it up should re-scope it against this decision. Cost is bounded by gathering the installed DBL uids
+  in ONE pass over the project collection and consulting ParatextData only for entries in that set:
+  `InstallableResource.ExistingScrText` is a computed property with no backing field that enumerates
+  the whole collection on every access, so asking each of the ~1850 catalog entries whether it is
+  installed would cost ~1850 full scans, and an installed entry pays it twice (once for `Installed`,
+  once inside `IsNewerThanCurrentlyInstalled`). Gating on `Installed` does not help, because that
+  property is the same lookup. The provider gate is taken with a non-waiting `Monitor.TryEnter`, so a
+  recheck never queues behind a catalog download or an install: everything holding that gate runs for
+  seconds, far longer than a refresh should block, so waiting could only delay the same empty answer,
+  and two rechecks cannot contend with each other because the only caller is single-flighted.
+  `IsNewerThanCurrentlyInstalled()` returns `true` for every *uninstalled* resource — nothing
+  installed trivially fails its "is the installed copy the newest" test — so the front end clears
+  `updateAvailable` for any row it reconciles as not installed rather than persisting a flag that
+  describes nothing. That keeps the cached flag meaning what the list renders it as: "the copy on
+  disk is out of date".
+- **Source:** Bug report that Get Resources keeps showing "Update" after a resource is updated.
 
 ## adr-decision-log-sorted-insertion: Decision-log entries are inserted in byte order by slug, not appended
 
--- a/.context/standards/Architecture-Decisions.md
+++ b/.context/standards/Architecture-Decisions.md
@@ -498,18 +498,27 @@ step, no automation. Just a record.
 - **Decision:** Extend the read-path reconciliation rather than adding a write-path invalidation. A
   new no-network provider function, `recomputeDblResourcesUpdateStatus`, re-evaluates
   `InstallableResource.IsNewerThanCurrentlyInstalled()` over the already-loaded catalog snapshot,
-  and the front end's `syncInstalledFlags` applies it alongside the `installed` check. The sync is
+  and the front end's flag sync applies it alongside the `installed` check. The sync is
   single-flighted by `ensureInstalledFlagsSynced`, so the recompute happens at most once at a time
   however many views refresh together. Reads do not wait for it — `getCachedResources` answers from
   the array it already has — which makes a read one refresh behind. That is invisible for
   `installed`, whose caller already knows what it just did, but it is the whole defect for
   `updateAvailable`: nothing else about an updated row changes, and there is no data-update event,
-  so the row keeps offering "Update" until the catalog is read a second time. A caller that has
-  just changed local state therefore awaits `refreshResourceFlags` and then re-reads, which is the
-  one path that does wait on the recompute. Resources absent from the result keep their cached
-  value, so a not-yet-loaded catalog or a busy provider gate degrades to the previous behavior
-  instead of guessing — including on that awaited path, where a contended gate leaves the stale
-  flag in place rather than blocking the user. Skipping the
+  so the row keeps offering "Update" until the catalog is read a second time.
+
+  Two consequences shape the wiring. The backend round trip is **opt-in** rather than part of every
+  sync, because exactly one surface renders `updateAvailable` — the Get Resources list — and the
+  other five consumers of the catalog would otherwise wait on a value they discard; the resource
+  picker's whole list spun on it. And a caller that needs the flag current awaits
+  `refreshResourceFlags`, which asks for the recompute and, rather than joining a sync already in
+  flight, lets that one finish before starting its own: an in-flight sync may have read its project
+  metadata before the caller's change, and a background sync does not recompute the flag at all.
+  The Get Resources view is therefore the only caller — once when its list settles, so a resource
+  updated from another surface loses its stale badge, and again after any install or removal the
+  user performs there. Resources absent from the result keep their cached value, so a
+  not-yet-loaded catalog or a busy provider gate degrades to the previous behavior instead of
+  guessing — including on the awaited path, where a contended gate leaves the stale flag in place
+  rather than blocking the user. Skipping the
   catalog fetch is sound because the DBL-side revision is the half that should stay fixed;
   ParatextData re-reads the *installed* revision on each call (`ExistingScrText` is a live
   `ScrTextCollection` lookup, and `InternalInstall` nulls the ScrText's FileManager before
--- a/c-sharp-tests/Projects/DigitalBibleLibrary/DblResourcesDataProviderTests.cs
+++ b/c-sharp-tests/Projects/DigitalBibleLibrary/DblResourcesDataProviderTests.cs
@@ -1,0 +1,216 @@
+using System.Diagnostics.CodeAnalysis;
+using Paranext.DataProvider.Projects.DigitalBibleLibrary;
+using Paratext.Data;
+using Paratext.Data.Archiving;
+using PtxUtils;
+
+namespace TestParanextDataProvider.Projects.DigitalBibleLibrary
+{
+    /// <summary>
+    /// Covers the update-status recheck the front end calls on every resource-list refresh, and the
+    /// catalog projection behind it.
+    ///
+    /// Three things about it are contracts rather than implementation detail. It has to answer
+    /// without reaching the DBL, including before the background catalog fetch has finished on
+    /// startup, because the TypeScript side treats an absent entry as "unknown" and keeps its
+    /// cached value — an exception or a network call here would instead stall or break the refresh.
+    /// Its keys are DBL entry uids, which is what the TypeScript looks entries up by. And it is
+    /// reached by name through `IDblResourcesProvider`, so its registered wire name is a
+    /// cross-language contract: renaming either half without the other breaks the refresh silently
+    /// at runtime rather than at compile time. The name also deliberately avoids the
+    /// `get`/`set`/`subscribe` prefixes reserved for data-type accessors.
+    /// </summary>
+    [ExcludeFromCodeCoverage]
+    internal class DblResourcesDataProviderTests : PapiTestBase
+    {
+        private const string RecomputeUpdateStatusWireName =
+            "object:platformGetResources.dblResourcesProvider-data.recomputeDblResourcesUpdateStatus";
+
+        /// <summary>
+        /// A catalog entry whose <see cref="InstallableResource.ExistingScrText"/> throws. That
+        /// property is the live project-collection lookup the installed-uid set exists to avoid, so
+        /// a test that reaches it fails loudly instead of silently depending on collection state.
+        /// </summary>
+        private sealed class ThrowingLookupResource : InstallableResource
+        {
+            public override ScrText ExistingScrText =>
+                throw new InvalidOperationException(
+                    "ExistingScrText must not be consulted for this resource"
+                );
+        }
+
+        /// <summary>
+        /// A catalog entry backed by a local project, so <c>IsNewerThanCurrentlyInstalled</c> runs
+        /// its real comparison chain instead of short-circuiting on "not installed".
+        /// </summary>
+        private sealed class InstalledResource(ScrText existing) : InstallableResource
+        {
+            public override ScrText ExistingScrText { get; } = existing;
+        }
+
+        private static InstallableResource ResourceWithUid(string dblEntryUid) =>
+            new ThrowingLookupResource { DBLEntryUid = HexId.FromStr(dblEntryUid) };
+
+        /// <summary>
+        /// An installed entry that reports up to date. A <see cref="DummyScrText"/> is not a
+        /// resource project, so once the name matches, <c>IsNewerThanCurrentlyInstalled</c> stops
+        /// at its <c>!scrText.IsResourceProject</c> branch and returns false.
+        /// </summary>
+        private static InstallableResource UpToDateResource(string dblEntryUid, ScrText existing) =>
+            new InstalledResource(existing)
+            {
+                DBLEntryUid = HexId.FromStr(dblEntryUid),
+                Name = existing.Name,
+            };
+
+        /// <summary>
+        /// An installed entry that reports an update available, via the name-mismatch branch of
+        /// <c>IsNewerThanCurrentlyInstalled</c>.
+        /// </summary>
+        private static InstallableResource OutOfDateResource(string dblEntryUid, ScrText existing) =>
+            new InstalledResource(existing)
+            {
+                DBLEntryUid = HexId.FromStr(dblEntryUid),
+                Name = existing.Name + "-renamed",
+            };
+
+        [Test]
+        public async Task RecomputeDblResourcesUpdateStatus_ReturnsEmptyBeforeCatalogIsFetched()
+        {
+            DblResourcesDataProvider provider = new(Client, ParatextProjects);
+
+            var updateStatus = await provider.RecomputeDblResourcesUpdateStatus();
+
+            Assert.That(updateStatus, Is.Empty);
+        }
+
+        [Test]
+        public async Task RecomputeDblResourcesUpdateStatus_IsRegisteredUnderItsContractName()
+        {
+            DblResourcesDataProvider provider = new(Client, ParatextProjects);
+
+            await provider.RegisterDataProviderAsync();
+
+            Assert.That(
+                Client.IsHandlerRegistered(RecomputeUpdateStatusWireName),
+                Is.True,
+                $"Expected '{RecomputeUpdateStatusWireName}' on the wire; "
+                    + $"registered: {string.Join(", ", Client.RegisteredRequestTypes)}"
+            );
+        }
+
+        [Test]
+        public void ProjectUpdateStatus_KeysByDblEntryUid()
+        {
+            var resources = new[]
+            {
+                ResourceWithUid("97196133a859179b"),
+                ResourceWithUid("6c21e835eb8ca3b2"),
+            };
+
+            var updateStatus = DblResourcesDataProvider.ProjectUpdateStatus(
+                resources,
+                new HashSet<string>()
+            );
+
+            Assert.That(
+                updateStatus.Keys,
+                Is.EquivalentTo(new[] { "97196133a859179b", "6c21e835eb8ca3b2" })
+            );
+        }
+
+        /// <summary>
+        /// The uninstalled answer is `true` — the same value ParatextData returns, since
+        /// `IsNewerThanCurrentlyInstalled` opens with `if (!Installed) return true;`. Reaching it
+        /// without touching the project collection is the whole point of the installed-uid set, and
+        /// the throwing lookup is what proves the collection was not touched.
+        /// </summary>
+        [Test]
+        public void ProjectUpdateStatus_ReportsUninstalledWithoutConsultingTheProjectCollection()
+        {
+            var resources = new[] { ResourceWithUid("97196133a859179b") };
+
+            var updateStatus = DblResourcesDataProvider.ProjectUpdateStatus(
+                resources,
+                new HashSet<string>()
+            );
+
+            Assert.That(updateStatus["97196133a859179b"], Is.True);
+        }
+
+        /// <summary>
+        /// A single failing entry must not cost the whole map. The TypeScript treats a missing key
+        /// as "unknown" and keeps the value it has, so omitting one entry degrades one row, whereas
+        /// letting the exception escape leaves every row stale for the session.
+        /// </summary>
+        [Test]
+        public void ProjectUpdateStatus_SkipsOnlyTheEntryThatThrows()
+        {
+            var resources = new[]
+            {
+                ResourceWithUid("97196133a859179b"),
+                ResourceWithUid("6c21e835eb8ca3b2"),
+            };
+
+            // Marking the first uid installed sends it down the ParatextData path, where this
+            // fake's lookup throws; the second stays on the no-lookup path.
+            var updateStatus = DblResourcesDataProvider.ProjectUpdateStatus(
+                resources,
+                new HashSet<string> { "97196133a859179b" }
+            );
+
+            Assert.Multiple(() =>
+            {
+                Assert.That(updateStatus.ContainsKey("97196133a859179b"), Is.False);
+                Assert.That(updateStatus["6c21e835eb8ca3b2"], Is.True);
+            });
+        }
+
+        /// <summary>
+        /// An installed, current resource reports no update — the state this whole recheck exists
+        /// to reach, since a stale `true` here is what left the row reading "Update" for the rest
+        /// of the session.
+        /// </summary>
+        [Test]
+        public void ProjectUpdateStatus_ReportsNoUpdateForAnInstalledCurrentResource()
+        {
+            using DummyScrText existing = new();
+            var resources = new[] { UpToDateResource("97196133a859179b", existing) };
+
+            var updateStatus = DblResourcesDataProvider.ProjectUpdateStatus(
+                resources,
+                new HashSet<string> { "97196133a859179b" }
+            );
+
+            Assert.That(updateStatus["97196133a859179b"], Is.False);
+        }
+
+        /// <summary>
+        /// Duplicate uids resolve first-wins, matching `FindResource`'s `FirstOrDefault`, so the
+        /// flag shown describes the same resource the install/uninstall buttons act on. The two
+        /// entries deliberately disagree, which is the only way this can tell first-wins from the
+        /// indexer's last-wins.
+        /// </summary>
+        [Test]
+        public void ProjectUpdateStatus_ResolvesADuplicateUidToTheFirstEntry()
+        {
+            using DummyScrText existing = new();
+            var resources = new[]
+            {
+                UpToDateResource("97196133a859179b", existing),
+                OutOfDateResource("97196133a859179b", existing),
+            };
+
+            var updateStatus = DblResourcesDataProvider.ProjectUpdateStatus(
+                resources,
+                new HashSet<string> { "97196133a859179b" }
+            );
+
+            Assert.Multiple(() =>
+            {
+                Assert.That(updateStatus, Has.Count.EqualTo(1));
+                Assert.That(updateStatus["97196133a859179b"], Is.False);
+            });
+        }
+    }
+}
--- a/c-sharp-tests/Projects/DigitalBibleLibrary/DblResourcesDataProviderTests.cs
+++ b/c-sharp-tests/Projects/DigitalBibleLibrary/DblResourcesDataProviderTests.cs
@@ -19,6 +19,14 @@ namespace TestParanextDataProvider.Projects.DigitalBibleLibrary
     /// cross-language contract: renaming either half without the other breaks the refresh silently
     /// at runtime rather than at compile time. The name also deliberately avoids the
     /// `get`/`set`/`subscribe` prefixes reserved for data-type accessors.
+    ///
+    /// NOT covered, deliberately: the body of <c>RecomputeDblResourcesUpdateStatus</c> past its
+    /// <c>_hasFetchedResources</c> guard — the gate and the one line joining the two halves,
+    /// <c>ProjectUpdateStatus(_resources, InstalledDblIds())</c>. Reaching it needs
+    /// <c>_resources</c> populated, and its only writer is a live DBL download, so every test here
+    /// returns at that guard. Replacing that call's arguments therefore passes the suite. Read the
+    /// count below as covering the two halves, not the seam between them; the seam is verified by
+    /// hand against a resource with a real pending update.
     /// </summary>
     [ExcludeFromCodeCoverage]
     internal class DblResourcesDataProviderTests : PapiTestBase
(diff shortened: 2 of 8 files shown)


katherinejensen00 (changes requested):
# PR #2751 — Fix Get Resources showing Update after a resource is updated

**Author:** Levi Wenger · **Branch:** `fix-get-resources-stale-update-flag` · **Base:** `main`
**Size:** 7 files, +391 / −27, 1 commit (`62b3117ddcf`)
**Reviewed:** 2026-09-04

## How this was produced

Four independent reviews (developer/soundness, architecture, tests, documentation), a `/code-review max` pass, and an adversarial pass over all of them. 36 raw findings reduced to the 22 below: four were cut as factually wrong, eleven downgraded. Every retained finding was verified against the code, and the ParatextData claims were checked by decompiling the pinned assembly (9.5.0.24, ILSpy).

## Verdict

The diagnosis is right, the mechanism is right, and the ParatextData reasoning is **verified correct**. Two things block merge: the branch is stale in a way that would revert `main`, and the 250 ms gate cannot deliver the guarantee it is built on. Findings 1 and 2 are what matter; most of the rest is documentation.

---

# Blockers and Important

## 1. Branch is 13 commits behind main, which already rewrote this function — Critical

**`extensions/src/platform-get-resources/src/main.ts:75`**

`git merge-tree` reports real content conflicts in `main.ts` and `Architecture-Decisions.md`. PR #2630 (`30f07f460b1`) restructured `getCachedResources`: it now returns `fetchMutex.runExclusive(...)` (line 199), the reconcile moved into a single-flight `syncInstalledFlags()` (line 128, guarded by `syncInFlight` at 190), and the project filter changed from `platformScripture.USJ_Chapter` to `platform.base` (line 103).

Resolving toward this branch reverts main's mutex guarding and single-flight. **Rebase before acting on anything else** — several findings below will dissolve or change shape afterward, and finding 2's fix likely lands naturally inside main's existing mutex.

## 2. The 250 ms gate cannot deliver its stated guarantee — Important

**`c-sharp/Projects/DigitalBibleLibrary/DblDownloadableDataProvider.cs:76`**

The comment sizes the constant against "a sibling recheck (a few ms of in-memory work)." Measured in a Release build over ~1850 catalog entries, a recheck is **74–220 ms at 25 local projects and 279–713 ms at 100** — one to two orders of magnitude off.

Because the cached path takes no mutex, N mounting views issue N concurrent recomputes: caller A holds `_providerGate` for 100–700 ms, caller B's `TryEnter` times out at 250 ms, gets `{}`, and keeps its stale "Update" badge. Home + Get Resources + the scripture-text-grid hook mounting together is the ordinary case, and it is the one the comment names — so the fix's own guarantee fails exactly where the bug lives.

Best fix is to dedupe: one in-flight recompute shared by all callers, which also resolves finding 4.

## 3. The recompute costs ~1850 live ScrTextCollection scans per call — Important

**`c-sharp/Projects/DigitalBibleLibrary/DblDownloadableDataProvider.cs:273`**

`IsNewerThanCurrentlyInstalled()` evaluates `ExistingScrText`, a computed property with no backing field, so each entry is a full collection enumeration — and installed entries pay it twice (once via `Installed`, once at the `??`). Uninstalled rows never short-circuit. This now runs on the UI refresh path.

**Suggested fix.** Note that gating the loop on `resource.Installed` does *not* help — `Installed` is itself a computed property that calls `ExistingScrText`, so it is the same full scan. Instead: enumerate `ScrTextCollection.ScrTexts(IncludeProjects.AllAccessible)` **once** into a `HashSet<string>` of installed DBL ids (the predicate is `scr.IsResourceProject && scr.Settings.DBLId == DBLEntryUid`), then call `IsNewerThanCurrentlyInstalled()` only for entries in that set and assign `true` directly for the rest — which is what ParatextData returns for `!Installed` anyway. That turns ~1850 scans into one, plus about two per installed resource.

For the record: `DBLResourceSettings` is a cached readonly field, so this is not per-call disk I/O — the cost is arbitrator-lock and snapshot churn.

## 4. Five of the six callers pay for a value they never render — Important

**`extensions/src/platform-get-resources/src/main.ts:92`**

`updateAvailable` is rendered in exactly one place, `get-resources.component.tsx:168`, plus the sort key at :347. The other five discard it:

| call site | renders `updateAvailable`? |
|---|---|
| `get-resources.web-view.tsx:33` | **yes** |
| `home.web-view.tsx:64` | no |
| `scripture-text-grid.web-view.tsx:228` | no |
| `use-dbl-resource-catalog.hook.ts:62` | no |
| `resource-picker.dialog.tsx:31` | no |
| `share-layout.dialog.tsx:60` | no |

During startup `activate` reads the cache and *then* starts the background fetch, which holds the gate for the whole download — so those five each block the full 250 ms for an answer they throw away. Consider an opt-in parameter, or a separate command for the Get Resources view.

## 5. The C# recompute has no behavioral coverage — Important

**`c-sharp-tests/Projects/DigitalBibleLibrary/DblResourcesDataProviderTests.cs:16`**

Verified by execution, twice independently: replacing the loop body with a comment yields `Failed: 0, Passed: 2`. `_resources` is `[]` and `_hasFetchedResources` is false, so the early return fires and the loop never runs. Surviving mutations include keying by `resource.Name` instead of `DBLEntryUid.Id` — silently breaking the TS lookup, the exact bug class this PR fixes.

The blocker is that `_resources` is private with only a networked writer. Extracting `internal static Dictionary<string, bool> ProjectUpdateStatus(IEnumerable<InstallableResource>)` makes the loop, duplicate-uid handling, and the gate branch all coverable.

In fairness: your second test does pin the `GetFunctions()` registration — that one earns its place.

## 6. Every TS test uses a single-element array, hiding a class of bug — Important

**`extensions/src/platform-get-resources/src/resources-cache.util.test.ts:17`**

Two mutants survive 9/9 green:

- Changing the `isChanged` latch to a per-entry assignment (`isChanged = installedChanged || updateAvailable !== resource.updateAvailable`) means last-entry-wins. If an earlier resource changed and a later one did not, `isChanged` ends `false`, `main.ts:105` skips the rewrite, and the stale list is returned — **reproducing the reported bug for anyone with more than one resource.**
- `cachedResources.map(...)` → `.slice(0, 1).map(...)` also survives.

Add a two-resource case where an earlier entry changes and a later one does not, asserting `isChanged` is `true` and `resources` has length 2.

## 7. The ADR argues against the design it documents — Important

**`.context/standards/Architecture-Decisions.md:2391`**

Consequences says the gate "is taken with a non-waiting `Monitor.TryEnter`… waiting could only delay the same empty answer." The code uses the timed 3-arg overload at 250 ms, and its own comment argues the opposite — that waiting is load-bearing. A maintainer trusting the ADR would "simplify" to the non-waiting overload and reintroduce the staleness the constant exists to prevent.

Whichever way finding 2 resolves, both texts need to say the same thing.

---

# Minor and Nits

## 8. One throwing resource discards the entire map — Minor

**`c-sharp/Projects/DigitalBibleLibrary/DblDownloadableDataProvider.cs:273`**

No per-item guard, so a single failure faults the task, `main.ts` logs one warning, `updateStatus` is `undefined`, and every row keeps its stale flag for the session.

Narrow but real trigger: `IsNewerThanCurrentlyInstalled()` does two independent `ExistingScrText` lookups, and `ExistingDictionary` is literally `=> null`, so an ungated `RefreshScrTexts` landing between them makes `scrText.Name` throw. The window is microseconds and it is pre-existing at `.cs:238`, but the "missing key = unknown" contract already covers omissions — a try/catch around the single call, skipping the entry, degrades one row instead of all of them.

## 9. "Never waits" is inaccurate on the public API surface — Minor

**`extensions/src/platform-get-resources/src/types/platform-get-resources.d.ts:26`**

It waits up to 250 ms. Suggested: *"Never contacts the DBL, and gives up quickly rather than blocking, so it is cheap enough to call on a UI refresh."* Don't restate the number in TS — it will drift.

## 10. The degradation path this PR adds is untested — Minor

**`extensions/src/platform-get-resources/src/main.ts:91`**

No test file imports `main.ts`. Deleting the inner try/catch survives — and that would let a provider throw escape to the outer catch and skip the `installed`/`projectId` reconciliation too, which is precisely what the comment promises will not happen. Extracting the block with injected `getLocalProjectIds` / `getUpdateStatus` makes this three cheap vitest cases.

## 11. Per-method `<summary>` blocks run against the repo convention — Minor

**`c-sharp-tests/Projects/DigitalBibleLibrary/DblResourcesDataProviderTests.cs:9`**

Verified precisely: 5 pre-existing per-method summaries across 1,569 test methods — but **66 of 90 fixtures** carry one, several essay-length. The convention is *document the fixture, not the method*. Fold both blocks into one fixture-level summary and this is conventional.

## 12. The ADR names three consumers; there are six — Minor

**`.context/standards/Architecture-Decisions.md:2340`**

See the table in finding 4. This matters because the entry uses that number to argue the per-read cost is acceptable.

## 13. The wire test's suffix match leaves half the contract unprotected — Minor

**`c-sharp-tests/Projects/DigitalBibleLibrary/DblResourcesDataProviderTests.cs:41`**

The full name is `object:platformGetResources.dblResourcesProvider-data.recomputeDblResourcesUpdateStatus`; renaming the provider id breaks every TS call site at runtime while this still passes. `DummyPapiClient.IsHandlerRegistered(requestType)` covers both halves in one line.

## 14. The projectId comment asserts an invariant the code does not hold — Minor

**`extensions/src/platform-get-resources/src/resources-cache.util.ts:46`**

A row cached as `{installed: true, projectId: ''}` matches by prefix, computes `installedChanged === false`, and keeps `''` indefinitely — and `isChanged` cannot see a `projectId`-only difference.

## 15. "unlike every other method here" is inaccurate — Nit

**`c-sharp/Projects/DigitalBibleLibrary/DblDownloadableDataProvider.cs:252`**

`IsGetDblResourcesAvailable`, `FindResource`, `InstallDblResourceCore`, and `UninstallDblResourceCore` do not load the catalog either.

## 16. Garbled cross-reference — Nit

**`.context/standards/Architecture-Decisions.md:2370`**

"`adr`-adjacent ticket PT-4268" is an editing artifact. PT-4268 appears nowhere else in the repo, so a reader following the backtick convention searches for a slug that does not exist. Should read "that PT-4268 describes,".

## 17. A pre-existing defect is parked in this ADR's Consequences — Nit

**`.context/standards/Architecture-Decisions.md:2393`**

The failed-install NRE is real but unrelated to this decision, with no ticket and no marker at `.cs:339`. Worth its own ticket so the analysis is reachable from the code.

## 18. The only wire method without `[NetworkTimeout]` — Minor

**`c-sharp/Projects/DigitalBibleLibrary/DblDownloadableDataProvider.cs:262`**

`GetDblResources` (:214), `InstallDblResource` (:300), and `UninstallDblResource` (:358) all carry the attribute; this one does not, so it inherits `network.service.ts`'s 30 s default — three orders of magnitude past what the TSDoc promises. An explicit value just above the gate timeout makes the bound match the documented contract.

## 19. Add a fast path before `Task.Run` — Minor

**`c-sharp/Projects/DigitalBibleLibrary/DblDownloadableDataProvider.cs:262`**

Throughout startup `_hasFetchedResources` is false, yet every mounting view queues a thread-pool dispatch and then blocks up to 250 ms behind the in-progress catalog download for a guaranteed-empty answer. `if (!_hasFetchedResources) return Task.FromResult(EmptyStatus);` removes both. Pairs with finding 4.

## 20. Two independent round trips run sequentially — Minor

**`extensions/src/platform-get-resources/src/main.ts:79`**

`getMetadataForAllProjects` and the recompute are only consumed together at line 98. `Promise.allSettled` preserves exactly the independent-failure semantics your comment describes while roughly halving refresh latency.

## 21. Duplicate-uid handling disagrees with `FindResource` — Nit

**`c-sharp/Projects/DigitalBibleLibrary/DblDownloadableDataProvider.cs:274`**

The indexer is last-wins; `FindResource` (:293) uses `FirstOrDefault`. `TryAdd` aligns them so the flag shown describes the same `InstallableResource` the buttons act on.

## 22. The one combination that decides the design question is untested — Minor

**`extensions/src/platform-get-resources/src/resources-cache.util.test.ts:41`**

Every test either flips `installed` with a silent map, or supplies a map with `installed` unchanged. Nothing covers **both** — a backend answer arriving while `installedChanged` is true. That is the install/uninstall path this PR exists to fix, and it determines whether `{installed: false, updateAvailable: true}` is intended or a `??` precedence accident. Today either behavior passes.

---

# Cut — please don't spend time on these

Findings that reached review but did not survive verification:

- **The `ScrTextCollection` race rejected in round 2 — the rejection was correct.** Two passes re-raised it, one citing this repo's own `GetScrTexts` comment. Decompiled: `ScrTextsInternal.MoveNext` snapshots to an array under `lock (scrTextsById)` before iterating, so "collection was modified" cannot arise. The `LocalParatextProjects` comment is right that the call is lazy but its stated reason is stale — worth a separate cleanup ticket, not this PR.
- **"Subscribe to `DBL_RESOURCES` from `main.ts`" as a missing ADR alternative** — not viable. `subscribe<data_type>` internally calls `get<data_type>`, and `getDblResources` unconditionally runs `FetchResourcesCore()`. It *is* the already-rejected "re-fetch the catalog on install."
- **"`installed` is now derived twice"** — it was derived twice before this PR, and the TS and C# questions differ (PAPI project lookup vs `ScrTextCollection`).
- **"`internal` visibility is anomalous"** — `AssemblyInfo.cs:3` has `[assembly: InternalsVisibleTo("c-sharp-tests")]` with 8+ precedents. Standard idiom.
- **"A throwing resource persists for the whole session after an uninstall"** — does not hold. `UninstallDblResourceCore` does `DeleteProject` + `RefreshScrTexts`, so the next scan finds nothing, `Installed` is false, and the method returns `true` immediately.
- **"A just-uninstalled resource sorts above its uninstalled peers"** — no. All uninstalled rows get `true` from the same expression, so the sort key shifts them equally.
- **Prototype-chain key hazard (`Object.hasOwn`)** — conceded unreachable; DBL uids are 16 hex chars.

# Verified correct — no action needed

Every ParatextData claim in the ADR checks out against the pinned 9.5.0.24 assembly:

- `ExistingScrText` is a live lookup with no backing field (computed property, `InstallableResource:102`).
- `Installed` derives from it live (`:64`).
- `InternalInstall` calls `scrText.SetFileManager(null)` on the same-name update path (`:277`), so `DBLResourceSettings` is rebuilt from the newly copied `.p8z`.
- `InstalledScrText = scrText` is the last statement before `return` (`:352`), after every failure path — so the `IsPresent(null)` reasoning holds too.

The one gap: nothing *pins* this against a ParatextData version bump. Consider recording the probe with version and date in the ADR.

**Also correct, and you flagged it yourself:** `updateAvailable` is `true` for every uninstalled resource, because `IsNewerThanCurrentlyInstalled()` opens with `if (!Installed) return true;`. Not user-visible — `get-resources.component.tsx:165` checks `installed` first and the sort key shifts all uninstalled rows equally — so it is a documentation caveat, not a behavior bug.

> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2751#-:-P0hj9Jj7hiagZQsDTtk:b-o0blgb)* status: all files reviewed, 1 unresolved discussion (waiting on captaincrazybro).

___
*[`c-sharp/Projects/DigitalBibleLibrary/DblDownloadableDataProvider.cs` line 76 at r1](https://reviewable.io/reviews/paranext/paranext-core/2751#-P0hjH_GB0QwQuwfQwNP:-P0hjH_GB0QwQuwfQwNQ:bzcbmfv) ([raw file](https://github.com/paranext/paranext-core/blob/62b3117ddcf14f11298c93165d36ba83e1164969/c-sharp/Projects/DigitalBibleLibrary/DblDownloadableDataProvider.cs#L76)):*
> ```Smalltalk
>     private const int DBL_NETWORK_TIMEOUT = 0; // Don't timeout DBL network requests
> 
>     // The update-status recheck runs on the front end's list-refresh path, where several web views
> ```

## 2. The 250 ms gate cannot deliver its stated guarantee — Important

**`c-sharp/Projects/DigitalBibleLibrary/DblDownloadableDataProvider.cs:76`**

The comment sizes the constant against "a sibling recheck (a few ms of in-memory work)." Measured in a Release build over ~1850 catalog entries, a recheck is **74–220 ms at 25 local projects and 279–713 ms at 100** — one to two orders of magnitude off.

Because the cached path takes no mutex, N mounting views issue N concurrent recomputes: caller A holds `_providerGate` for 100–700 ms, caller B's `TryEnter` times out at 250 ms, gets `{}`, and keeps its stale "Update" badge. Home + Get Resources + the scripture-text-grid hook mounting together is the ordinary case, and it is the one the comment names — so the fix's own guarantee fails exactly where the bug lives.

Best fix is to dedupe: one in-flight recompute shared by all callers, which also resolves finding 4.

captaincrazybro:
Thanks — this was a genuinely useful review, and finding 1 was the right call to lead with: rebasing first dissolved or reshaped several of the others, exactly as you predicted.

Force-pushed as `bc99c37` (rebased onto `495d44a`, 0 behind). CI is green on all three platforms.

## Please weigh in: 3 findings I closed rather than fixed

These are judgment calls, not oversights, and I'd rather you disagree now than discover them silently dropped.

- **4 — five of six callers pay for a value they never render.** The premise was that each blocks the full 250 ms. After the rebase the recompute runs inside `syncInstalledFlags`, which `ensureInstalledFlagsSynced` single-flights and nothing awaits, so no caller blocks on it at all; and finding 3's fix removed the CPU cost. An opt-in parameter would now add a cross-language contract knob for no measurable gain. **If you still want the parameter, say so and I'll add it.**
- **10 — the degradation path is untested.** Left to the util's coverage. Post-rebase the extraction is bigger than costed: `getLocalProjectMetadata`'s retry logic and the `!hasResourceProjects` early return now live in the same function, so injecting readers means restructuring more than the try/catch. Your call whether that's worth it.
- **20 — `Promise.allSettled` for the two round trips.** Skipped deliberately. The latency is on a background path nobody waits on, and parallelizing would fire the recompute even when the `!hasResourceProjects` early return discards it — which is every startup before the C# PDPF registers. The trade seemed to go the wrong way, but it is a trade.

## Fixed (14)

- **1** — rebased; the fix is re-derived on top of main's single-flight `syncInstalledFlags` rather than reverting its mutex guarding. Decision-log conflict resolved as a slug-keyed insert, not "keep both sides".
- **2, 7** — the gate is non-waiting again. The 250 ms wait existed to outlast a sibling recheck, which the single-flight sync makes impossible, so it bought nothing and contradicted the ADR. Both texts now agree. (This is the open Reviewable discussion on `:76` — that constant no longer exists, so that thread can be resolved.)
- **3, 5** — installed DBL uids are now gathered in **one** pass and `IsNewerThanCurrentlyInstalled()` is called only for entries in that set; everything else gets `true` directly, which is what ParatextData returns for anything uninstalled. Your note that gating on `Installed` doesn't help was correct and saved me a wrong turn. The loop moved to `ProjectUpdateStatus(resources, installedDblIds)`, which made it coverable — the seam is that `ExistingScrText` is `public virtual` and `InstallableResource` has a protected parameterless constructor, so test doubles need no REST client, catalog, or live collection. Five behavioral tests, each mutation-verified: keying by `DBLEntryUid.Id` rather than `Name` (your surviving mutant), the uninstalled path proven not to touch the collection at all, per-entry error isolation, first-wins duplicate resolution, and an installed current resource reporting no update.
- **6, 14** — 14 turned out to be a real defect, not just a wrong comment: a row cached `{installed: true, projectId: ''}` matches by prefix, so `installedChanged` is false and `''` persists indefinitely. `projectId` is now recomputed unconditionally and included in change detection. Plus main's empty-`dblEntryUid` guard, and your two-resource `isChanged` case. Three tests, each mutation-verified.
- **8, 21** — per-entry try/catch, and `TryAdd` for first-wins.
- **9** — reworded, with no number restated in TS.
- **11, 13** — one fixture-level summary; the wire test pins the full name via `Client.IsHandlerRegistered`.
- **15** — now names the methods it means. One correction to the finding: `InstallDblResource` and `UninstallDblResource` *do* load the catalog, via `EnsureResourcesLoadedCore()` — it's their private `*Core` helpers that don't. The comment was still wrong as written, because `IsGetDblResourcesAvailable` doesn't load.
- **18** — added. Not "just above the gate timeout" though: with a non-waiting gate the bound has to cover the recompute itself, which you measured at up to 713 ms.
- **19** — applied, and the `_hasFetchedResources` comment now documents that it may be read without the gate as a one-way hint (it's only ever set `true`, in `FetchResourcesCore`) rather than silently violating its own stated invariant.
- **22** — settled rather than just pinned. `updateAvailable` is now cleared for any row reconciled as not installed, since the backend reports `true` for everything uninstalled and persisting that stored a flag describing nothing.
- **12, 16, 17** — six consumers not three; the garbled PT-4268 cross-reference; and the failed-install NRE removed from Consequences. Entry re-inserted in byte-order slug position per the log's own rule.

## Filed (1)

**17 → PT-4531.** Worth a look, because re-verifying it against the pinned 9.5.0.24 turned up something the earlier "Cut" list got wrong. Your retraction of "a failed install reports success" was right about the case it analyzed but incomplete: `InstalledScrText` is never reset and lives on a session-lived `InstallableResource`, so if a resource was installed successfully earlier *in the same session*, a later failed **update** finds a stale-but-still-present ScrText, `IsPresent` returns true, and the failure is reported as success. So it's two defects, not just message quality. Also: the likeliest trigger is a failed download in `InstallableDBLResource.Install()`, which returns before `base.Install()` ever runs.

## Verification

- C# full suite: **1724 passed, 0 failed**, 6 skipped. The 6 `LocalParatextProjectsWatcherTests` failures I'd reported as pre-existing do **not** reproduce on the rebased tree, and CI is green — so that's answered.
- Extension tests: 70 passed across 6 files; the reconcile util is at 14 cases, up from 9.
- `npm run lint` exits 0.
- **The end-to-end gap is closed** — the fix is confirmed against a real resource with a pending DBL update: clicked Update, row went to "Installed", no restart. That was the open question on both prior rounds.

One known-surviving mutant, deliberately: removing finding 19's fast path leaves all 7 C# tests green. With `_resources` empty the loop is a no-op either way and the gate check right after returns the same empty map, so it's a pure optimization with no observable behavior. The behavior it fronts is pinned; a test for the dispatch itself would be timing-dependent.

katherinejensen00 (commented):
# PR #2751 — re-review after rebase

**Head:** `bc99c37530f` (was `62b3117ddcf`) · **Base:** `495d44a439d`, 0 behind · **Reviewed:** 2026-09-08

Seven perspectives (security, scope, ux, contracts, tests, architecture, clarity). 34 raw findings,
trimmed to the 5 below.

## Verdict

The rework is a large, genuine improvement, and the riskiest item is confirmed good: the rebase
preserved main's `fetchMutex`, single-flight `syncInFlight`, and the `platform.base` filter — nothing
of PR #2630 was reverted. ADR slug position is correct, no Send/Receive gate is needed, security is
clean, and `resources-cache.util.ts` went from non-falsifiable to 7 of 7 mutants killed. 15 of the 22
round-3 findings are fixed.

Finding 1 is the one that decides whether this PR does what its title says.

---

## 1. The recomputed flag never reaches the dialog the user is looking at — Critical (question)

For an **update** of an already-installed resource:

1. `get-resources.web-view.tsx:130` calls `refetchResources()` in `.then()` after the install resolves.
2. That re-issues `getCachedResources`, which at `main.ts:212` fires `ensureInstalledFlagsSynced()`
   as a floating promise and returns the **pre-sync** array on the same tick.
3. The spinner-clearing effect (`get-resources.web-view.tsx:145-157`) keys only on
   `resource.installed`. For an update `installed` was already `true`, so the entry is dropped and the
   spinner clears immediately.
4. `get-resources.component.tsx:176-181` renders: `installed` true, `updateAvailable` still stale
   `true` → the **"Update" button again**.
5. `syncInstalledFlags` writes the corrected array to `cachedResources` later, but there is no
   data-update event and no subscriber — a repo-wide grep for `.DblResources(` returns zero matches,
   as the ADR itself states. React is never told.

The row reads "Update" until the dialog is closed and reopened — the symptom in the PR title, in the
same session. Install and uninstall are masked because `installed` flips; the update case has no
second signal.

**Asking rather than asserting:** the PR reports "clicked Update, row went to Installed, no restart."
A restart is not a reopen, and both observations hold if the dialog was reopened. If the row really
did clear in the open dialog, one of the five steps above is wrong and I would like to know which.

If the reading holds, the post-install refresh needs to await the sync. `ensureInstalledFlagsSynced`
already declares that awaiting mode in its own TSDoc (`main.ts:190-194`) — no caller uses it.

## 2. The recompute runs on a user-waiting path, contradicting three documents — Important

Found independently by scope, ux, and architecture.

`getLocalNonDblResources` does `await ensureInstalledFlagsSynced()` (`main.ts:248`), and this PR puts
a C# round trip inside that sync (`main.ts:157-162`). Two user-facing consumers await that command:
`resource-picker.dialog.tsx:58`, which feeds `isLocalLoading` into `isResourcesLoading` at :70 so the
picker's whole list spins on it, and `use-dbl-resource-catalog.hook.ts:83`.

Worst case is now up to 5 × 500 ms of `getLocalProjectMetadata` retries **plus** the new
`UPDATE_STATUS_NETWORK_TIMEOUT` of 5000 ms — roughly 7.5 s of spinner, for a value neither consumer
renders.

Three places assert the opposite: `Architecture-Decisions.md:426` ("never on a path a caller waits
on"), `platform-get-resources.d.ts:32` ("call it on a background refresh rather than on a path a user
is waiting on"), and `DblDownloadableDataProvider.cs:286` ("far longer than a list refresh should
block").

The awaiting call site is pre-existing, so this is not a rebase regression — but it is the stated
rationale for closing round-3 finding 4, and that rationale does not hold as written. Either keep the
recompute out of the awaited leg, or correct the three claims.

## 3. The C# tests still cannot see the wiring, and the surviving mutant is this PR's bug

Verified by execution, baseline 7/7 green:

- `ProjectUpdateStatus(_resources, InstalledDblIds())` → `ProjectUpdateStatus(_resources, new HashSet<string>())`
  — **survives 7/7**. That mutant reports `updateAvailable: true` for every installed resource: the
  permanently-stuck badge this PR exists to remove.
- Whole `Task.Run` body → `return new Dictionary<string, bool>();` — **survives 7/7**.
- Gating the call with `&& false` — **survives 7/7**. Nothing asserts the positive case, so the badge
  could stop appearing when an update really is available and no test would notice. `OutOfDateResource`
  (`DblResourcesDataProviderTests.cs:70`) already yields `true`; it is never asserted directly.

The extraction did work — the four mutants round-3 finding 5 named are all killed now. This is the
residue: `RecomputeDblResourcesUpdateStatus`'s own body and `InstalledDblIds()` are unexecuted.
`DummyLocalParatextProjects.FakeAddProject` is the seam that exists. If it stays uncovered, say so
deliberately rather than letting the test count imply otherwise.

## 4. `InstalledDblIds()` has no per-item guard, defeating the guard on the loop it feeds

`DblDownloadableDataProvider.cs:327`. `ProjectUpdateStatus` (:359-376) has a per-entry try/catch whose
comment says letting a throw escape "would fault the whole task and leave every row stale for the
session". `InstalledDblIds()` runs first, in the same gate and the same failure domain, with no guard —
and `scrText.IsResourceProject` / `scrText.Settings` both fault on a corrupt `Settings.xml`. One bad
project produces exactly the outcome the other catch block exists to prevent.
`LocalParatextProjects.TryGetProjectDetails` is the in-repo precedent for isolating this.

## 5. A third divergent implementation of "does this row cover this project"

`resources-cache.util.ts:29-37`. `platform-bible-utils` exports `doesCatalogRowCoverProject`
(`lib/platform-bible-utils/src/resources.model.ts:39`), whose own TSDoc says: "This is the single home
for that rule... They must agree, or a project is claimed by one and disowned by the other." The new
util does not call it, and diverges twice — the shared helper refuses to match a never-synced row, and
falls through to a prefix check when a non-empty `projectId` misses; this ternary does neither.

The logic is relocated, not new. But exporting it from a named util is what turns an inline detail into
a second published definition of the rule, and that is the moment to reconcile the two or record why
they differ.

---

## Round-3 status

**Fixed (15):** 1 (rebase — verified line by line), 2, 3, 6, 7, 8, 9, 12, 13, 14, 16, 18, 19, 21, 22.
**Partially fixed (4):** 4 (see finding 2), 5 (see finding 3), 11, 17.
**Not fixed (2):** 10 and 20, both closed with a stated rationale; 20's reasoning is sound and I withdraw it.

**Verified clean:** ADR byte-order slug position; no Send/Receive write gate needed; no blocked imports,
secrets, CSP or injection surface; DataProvider/registration pattern conformance; JSON-RPC key naming.
Nothing from round 3's "Cut" or "Verified correct" sections was re-raised, and no reviewer found
evidence against either.

**Trimmed from this round, available on request:** a non-volatile lock-free read of
`_hasFetchedResources`; producer/reconciler disagreement over what `updateAvailable` means (and ~50KB
of meaningless-by-construction wire payload per sync); two docs stating contracts the code does not
hold; `[NetworkTimeout]` unpinned by any test; plus 12 naming, comment-accuracy and ADR-prose nits.

> @katherinejensen00 reviewed 7 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2751#-:-P129l6_CG-kQ8ujXV85:bsk23bz)* status: all files reviewed, 1 unresolved discussion (waiting on captaincrazybro).

captaincrazybro:
Round 4 addressed — pushed as `d480cc6` (rebased onto today's main; the earlier conflict is gone).

## 1 — you were right, and none of the five steps was wrong

You asked which step my "clicked Update, row went to Installed" claim contradicted. The answer is none of them. I re-traced it independently before touching anything, then reproduced it against a real resource with a pending update: **the row keeps offering "Update" in the open dialog.** The earlier observation must have involved a reopen, exactly as you guessed — my report of it was not careful enough to distinguish the two, and it should have been, since that is the only distinction that matters.

For the record, the one thing I checked beyond your five steps: nothing else calls `refetchResources`, so there is no third read that could have rescued it.

**Fix** (`b0780a2`): a new `platformGetResources.refreshResourceFlags` command that resolves once the flags reflect the caller's change, awaited by the web view before it refetches.

One wrinkle worth flagging, because the obvious version of this fix is subtly wrong: awaiting `ensureInstalledFlagsSynced()` is not enough. It joins whatever sync is already running, and that one may have read its project metadata *before* the install — so it can resolve on pre-install state. The command lets any in-flight sync finish first, then starts one guaranteed to observe the change.

Side effect: the row's spinner now covers the sync too, so it goes spinner → "Installed" rather than clearing early.

## 2 — fixed, and it made 1 easier

`syncFlags` now takes the recompute as a parameter. The background sync leaves it off; only `refreshResourceFlags` turns it on. The picker and `use-dbl-resource-catalog` no longer wait on it at all.

Since no other path recomputes the flag any more, the Get Resources view also refreshes once when its list settles — otherwise a resource updated from another surface would arrive here with a stale badge, which would have been a regression rather than a fix. That view is the only surface that renders `updateAvailable`, so it is the one that pays.

The decision log now describes the opt-in and why the split falls where it does.

## 3 — partially, and the residue is stated rather than implied

Your `&& false` mutant is dead: nothing asserted the positive case, so I added one (`e22de6e`). `InstalledDblIds()` is now `internal` and covered by two tests (`370a8dc`), mutation-verified — dropping the `IsResourceProject` filter, never adding to the set, and removing the per-item guard each turn exactly one red.

Building those doubles produced a nice piece of evidence for finding 4: `DummyScrText`'s constructor reads `Settings`, which resolves `Guid`, which consults `IsResourceProject`. A settings fault really does cascade through paths that look unrelated to settings, and a double that throws from the start cannot even be constructed.

Still uncovered, deliberately, and now written on the fixture so the count does not imply otherwise: the body of `RecomputeDblResourcesUpdateStatus` past its `_hasFetchedResources` guard — the gate plus the single line joining the two tested halves. Reaching it needs `_resources` populated, and its only writer is a live download, so covering it means adding a test-only writer to state that has exactly one. That seemed a poor trade for four branchless lines whose failure mode the manual check catches. Say the word if you disagree; the seam is small to add.

## 4 — fixed

Per-item try/catch on the scan (`e22de6e`). The handler deliberately does not name the project: reading anything off it is what just failed, so naming it could throw out of the catch.

## 5 — documented rather than unified, and I think the finding slightly understates the gap

They do not just diverge; they answer opposite questions, and calling the shared helper here would be a bug. `doesCatalogRowCoverProject` opens with `if (!row.installed && row.projectId === '') return false;`. That guard is right for an exclusion test. But recognising that a row cached as `{installed: false, projectId: ''}` now has a project on disk is precisely what marks it installed — so routing the reconcile through the helper makes **every first-time install undetectable**. It fails the existing test.

So `d480cc6` records the distinction at the reconcile, which is where someone would try to unify them. I left the helper's own "single home" TSDoc alone: `platform-bible-utils` ships through a committed `dist/`, so amending a comment there obliges rebuilding it and `platform-bible-react`, which embeds it. Happy to do that if you would rather the claim be qualified at both ends.

## Not yet re-verified

The end-to-end check is **stale**: both 1 and 2 changed the refresh path after the manual test that confirmed the bug. Re-testing in paratext-10-studio now — I would rather tell you that than have you assume the fix is confirmed. Note this needs the C# side rebuilt, not just the extension.

## Please do send the trimmed list

Particularly the wire-payload one — the recompute returns every catalog entry, and the ~1800 uninstalled ones carry a `true` that means nothing by construction. That sounds real, and it is adjacent to the opt-in change above. The non-volatile read of `_hasFetchedResources` is mine too, from round 3's finding 19.

captaincrazybro:
Closing the loop on the "not yet re-verified" note above: **re-tested on `d480cc6` in paratext-10-studio, against a real resource with a pending DBL update. It works.**

Clicking Update takes the row to "Installed" in the *open* dialog — no reopen, no restart. That is the distinction the earlier claim failed to make, so it was the specific thing checked this time.

Also exercised, since the install path and the sync both changed shape:

- First-time install, uninstall, and a failed install
- Opening Get Resources — the added mount-time refresh does not flicker or double-load the list
- Resource picker, Home, scripture text grid, Share Layout — unaffected by the opt-in split, as intended
- The cross-surface case: install from the scripture text grid, then open Get Resources — no stale badge, which is what the mount-time refresh exists for

The PR body's "Verified end to end" section has been corrected too. It previously asserted this on evidence that did not support it; it now says so, and states what the re-check actually covered.

Not claimed: the `InstalledDblIds` per-item guard, which needs a corrupt `Settings.xml` to reach and is covered by unit tests only.

katherinejensen00 (approved):
Nothing blocking so going ahead and approving. Here are two more things to consider.

1. The design-decision log entry (finding 5). The entry is named adr-dbl-cache-recompute-on-read, but the code no longer recomputes on read. Before merge it's a simple edit, including renaming the slug. After merge, the log's own rules mean it has to be marked superseded and a new entry added, which leaves a wrong entry readable as history. This is the only item that gets harder to fix once merged.
2. A failed refresh reported as a failed install (finding 1). This is new in this PR. It only happens when the refresh times out, so it's rare, but the user is told a successful install failed. The fix is a few lines: catch the refresh on its own, log it, and refetch anyway, as the refresh on open already does.

> @katherinejensen00 reviewed 7 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2751#-:-P1IbQ2v5FHndkOFNsvY:bt9wbaj)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2729, opened 2026-08-27
PT-4286: Closing the primary window asks before closing every window

rolfheij-sil:
## Summary

Closing the primary window while other windows are open now asks the user once, and on confirm closes every window with each layout kept for the next launch. Before this, that close left the app running on the secondaries and spliced the primary's persisted entry out of the window-layouts structure, so the user's main layout never came back — the quit decision was `window-all-closed`, a pure window count that keyed on nothing.

**Why review this:** part of the PT-4275 multi-window epic (PT-4286, "Window-close rule — team decision 2026-08-26"). It makes the primary role load-bearing for the app's lifetime, which amends the equal-siblings ruling — the ADR entry records exactly how far that amendment goes and where it stops.

Stacked on #2724 (PT-4282). Merge that first.

## The rule

| Close | What happens |
| --- | --- |
| Primary ✕, other windows open | Prompt in main. **Close all windows** → every window closes, every layout saved, all restored next launch. **Cancel** → nothing happens. |
| Primary ✕, alone | Quits, no prompt — as before. |
| Secondary ✕ | Closes that window only; its layout is dropped — as before. |
| Primary emptied (last tab moved out) | **Docks Home and stays.** Only its ✕ and Quit close it. |
| Cmd+Q / File → Quit / `platform.quit` | Unchanged, no prompt — including while the question is open. |

Dialog copy (en + es): message *"Close the application?"*, detail *"All windows will close. They will be restored the next time you open the application."*, buttons *"Close all windows"* / *"Cancel"*. Esc and the window-manager dismiss map to Cancel. Localization is bounded at 3 s and falls back to English rather than holding the close.

## Changes

- **`window-state.service.ts`** — `countWindowsThatWouldStayOpen(closingWindowId)`: tracked, not destroyed, not already closing. Deliberately **not** `countWindowsThatCouldBeTheLastOne`, which excludes windows still loading their content — a move-to-new-window target mid-load is not a candidate to be the last window, but it is very much one the user would lose.
- **`window-layout-persistence.service.ts`** — `isPrimaryWindow(windowId)`, answering from the persisted `isMain` slot while a live window holds it, and from the oldest live window when none does (see *The primary role when no window holds the marked entry* below). It is the only supported way to ask — `getMainWindowId` is no longer exported. There is deliberately **no** second live pointer: the slot's `windowId` is cleared from the `closed` handler, and the close path asks in `close`, one event earlier, so the answer is always there when needed.
[...]

PR #2357, opened 2026-06-03
fix(flaky): increase PAPI registration timeout in UI smoke test

merchako:
## Root-cause analysis

**Flaky test:** `UI Interaction › should open the About dialog from the Help menu`
**File:** `e2e-tests/tests/smoke/ui-interaction.spec.ts`

### What was failing

The `beforeAll` hook called `waitForPapiMethodRegistered` with no custom timeout, falling back to the 60 s default:

```ts
await waitForPapiMethodRegistered(SETTINGS_SET_METHOD);  // 60 s default
```

On slow macOS CI runners, the extension host (which runs TypeScript source via ts-node and initializes services including `settingsServiceDataProvider`) takes **longer than 60 s** to register its PAPI network method, so `rpc.discover` never returns the method within the window:

```
Error: PAPI method "object:platform.settingsServiceDataProvider-data.set"
       not listed in rpc.discover within 60000ms
```

Playwright retries the test twice (config `retries: 2`), but all three attempts hit the same wall.

### CI failure evidence (last 7 days on `main`)

| Date | SHA | Run | Outcome |
|------|-----|-----|---------|
| 2026-05-27 | `89a233c6` | [#26522625632](https://github.com/paranext/paranext-core/actions/runs/26522625632) | ❌ macOS E2E |
| 2026-05-28 | `acf5d934` | [#26593283014](https://github.com/paranext/paranext-core/actions/runs/26593283014) | ❌ macOS E2E |
| 2026-06-02 | `c6bc85ea` | [#26833349533](https://github.com/paranext/paranext-core/actions/runs/26833349533) | ❌ macOS E2E |
| 2026-06-02 | `73a29b52` | [#26839738720](https://github.com/paranext/paranext-core/actions/runs/26839738720) | ❌ macOS E2E |
| 2026-06-02 | `5e0a37e5` | [#26851266598](https://github.com/paranext/paranext-core/actions/runs/26851266598) | ❌ macOS E2E |

The same commit (`f7f0fcf2`) failed on attempt 1 and passed on a re-run, confirming pure timing flakiness.

---

## Why exact local reproduction is hard

On a fast developer machine the extension host starts in ~2 s and the settings service registers well within the 60 s window, so the test always passes locally. The flake is a **slow-runner-only phenomenon**: it requires the total time from Electron launch to settings-provider registration to exceed 60 s, which only happens on loaded CI hardware.

[...]

PR #2338, opened 2026-05-27
E2E smoke `beforeAll` PAPI timeout fix (round 2)

jolierabideau:
# Code Review Summary

**Branch**: `fix-test-timeout`

**Base**: `origin/main`

**Date**: 2026-05-27

**Review model**: Claude Opus 4.7

**Files changed**: 1

## Overview

This branch fixes a recurring flake in the `UI Interaction` smoke suite's
`beforeAll`: the `sendPapiRequestOnce` call that sets `platform.interfaceLanguage`
was hitting the helper's default 10s timeout on slow CI. The new behavior passes
an explicit 90s `perRequestTimeoutMs` so the call can absorb the time the
settings `set` handler spends waiting on `waitForResyncContributions()` (which
blocks until `extensionService.initialize()` finishes in the extension host).

The change is one file, three lines of behavioral diff. The `waitForPapiMethodRegistered`
fast-fail guard from PR #2336 is retained but its comment is rewritten to
accurately describe its role; the inline rationale block above the call was
folded into a named constant (`SLOW_CI_PAPI_TIMEOUT_MS`) during this review.

## API Changes

None. The only changed file is `e2e-tests/tests/smoke/ui-interaction.spec.ts`,
which is an internal end-to-end test file and not part of any public API surface
(`lib/platform-bible-react/`, `lib/platform-bible-utils/`, `@papi/` modules in
`papi.d.ts`, or extension `.d.ts` declarations).

### Critical — Must address before merge

None.

### Important — Should address before merge

None.

### Minor — Consider

- [x] **Magic number `90_000` inlined at the call site** _(fixed during review: extracted to `const SLOW_CI_PAPI_TIMEOUT_MS = 90_000;` at the top of the describe block; moved the resync rationale into the constant's declaration comment and removed the now-redundant inline comment block above the call)_
- [ ] ~~**`undefined` for `port` is inconsistent with other `sendPapiRequestOnce` call sites**~~ _(dismissed: the analyzer cited `comment-test-helpers.ts:174,190,279,319` as evidence of a convention, but those sites pass a local `resolvedPort` variable, not `DEFAULT_WEBSOCKET_PORT`. They pass a port because they have one in scope from their fixture context, not as a style rule. `DEFAULT_WEBSOCKET_PORT` is also not exported from `helpers.ts` — it's a file-private const. In the smoke `beforeAll` there's no port in scope; `undefined` correctly opts into the default without forcing a new export.)_
[...]

PR #2336, opened 2026-05-27
Fix: smoke E2E `beforeAll` PAPI timeout

jolierabideau:
## What changed

**File:** `e2e-tests/tests/smoke/ui-interaction.spec.ts` (one file, 16 insertions / 9 deletions)

- Extended the existing import from `../../fixtures/helpers` to also pull in
  `waitForPapiMethodRegistered`.
- Lifted the settings-set method name into a describe-scoped `const`
  (`SETTINGS_SET_METHOD = 'object:platform.settingsServiceDataProvider-data.set'`)
  so the wait and the send refer to the same string.
- Inside `test.beforeAll`, added `await waitForPapiMethodRegistered(SETTINGS_SET_METHOD)`
  immediately before the existing `await sendPapiRequestOnce(SETTINGS_SET_METHOD, …)` call.

## Why it was needed

The smoke suite intermittently aborted on CI before any test ran, with:

> Error: PAPI request "object:platform.settingsServiceDataProvider-data.set"
> timed out after 10000ms

The `beforeAll` forces the interface language to English so menu-item text
matchers (`/Help/i`, `/About Platform\.Bible/i`) are deterministic. It did so
by calling `sendPapiRequestOnce` — a single-shot JSON-RPC call with a 10-second
timeout — immediately after the Electron window appeared.

On slow CI runs, the settings data provider hadn't yet finished registering
with the PAPI network when that call fired. The WebSocket got no response,
the 10-second timeout tripped, and Playwright's three retries all hit the
same wall. The whole suite aborted.

## Why this fix

The rest of the E2E suite already established the pattern: any time you
intend to call `sendPapiRequestOnce` directly, first poll `rpc.discover` via
`waitForPapiMethodRegistered` to confirm the method is registered. That
pattern appears in `fixtures/comment-test-helpers.ts`, in `dialog-rendering.spec.ts`,
and inside `waitForAppReady` itself. The `ui-interaction.spec.ts` `beforeAll`
was the only direct caller that skipped it.

`waitForPapiMethodRegistered` polls every 250 ms with a 60-second timeout and
returns as soon as the method appears — so once the provider registers, the
subsequent `sendPapiRequestOnce` returns near-instantly. No new abstractions,
no fixture changes, no new tests; the existing smoke suite is the regression
test.

## Verification

- TypeScript typecheck: clean on the changed file (a single pre-existing
  `'hello-world'` types error remains in the project, unrelated to this
  change).
- ESLint: clean on the changed file.
[...]

PR #2753, opened 2026-09-02
PT-4481: Test the close-all prompt's abort and localization paths

rolfheij-sil:
## Summary

`confirmCloseAllWindows` — the close-all confirmation question — was a closure inside `main()`, so nothing could reach it without a `BrowserWindow` and an Electron runtime. Three of its paths had no coverage at any level, including one that a reasonable reader would think was already covered.

It moves to `src/main/services/close-all-prompt.service.ts` **unchanged** — same strings, same button order, same `defaultId`/`cancelId` — the way `decideWindowClose` moved into `window-close-decision.service.ts` for exactly this reason. This is a test PR: no behaviour changes.

**Why review this:** part of the PT-4275 multi-window epic, as the hardening sub-task PT-4481. It came out of the `/review-paratext` sweep of #2729, where Rolf ruled these were not to be fixed in that PR.

## Changes

- **`src/main/services/close-all-prompt.service.ts`** (new) — `confirmCloseAllWindows(primaryWindow)`, lifted verbatim. `CLOSE_PROMPT_LOCALIZE_TIME_OUT_MS` moves with it, since nothing else read it.
- **`src/main/main.ts`** — −112/+1. The function, the constant, and the five imports the move orphaned (`dialog`, `CloseAllAnswer`, `whenQuitRequested`, `localizationService`, `LocalizeKey`). `wait` stayed: it is still used elsewhere.
- **`src/main/services/__tests__/close-all-prompt.test.ts`** (new) — six tests, below.

No new wire surface, and nothing here is public API, so no experimental markers apply.

## What the six tests pin

**The abort wiring** asserts the **signal** the box is given, not the answer that comes back. This is the point of the ticket: the answer cannot tell a working abort from a broken one, because the caller races this against the quit latch and the latch decides either way. A box shown with no signal — or with one nothing ever aborts — would look identical from outside while leaving the question on screen through the whole shutdown. The test checks the signal arrives un-aborted (the control, since an already-aborted signal would satisfy the next assertion for free), then that a quit flips it.

**The localization bound**, both halves, which fail differently:
- a call that never answers must give up at the bound and show English;
- a call that *fails* must fall back **at once** rather than waiting the bound out. That second case is why the race uses a throwing timer instead of `waitForDuration`, whose `Promise.any` would swallow a fast failure — a claim the code comment makes and nothing checked.

**The rest of the contract**: the localized text is used when localization answers in time; `close-all` and `cancel` map to the right buttons; and Cancel is both `defaultId` and `cancelId`, so neither Enter nor Esc can close every window.

## Mutation-checked, not merely green

Each test was broken deliberately and confirmed to fail *on its own*:

| mutation | result |
| --- | --- |
| drop `signal: dismissOnQuit.signal` from the options | dismissal test red alone (1 failed / 5 passed) |
| keep the signal, never call `.abort()` | dismissal test red alone |
| remove the bound from the race | the never-answers test hangs to timeout, red alone |
| rebuild the race in the `Promise.any` (`waitForDuration`) shape | **both** localization tests red — the fast-failure one at 3002 ms, exactly the bound it exists to avoid waiting out |

The last mutation is why the fast-failure case earns its place: it is the only test that catches that shape, and the third mutation does not touch it. Each mutation was reverted and the file confirmed byte-identical by checksum.

## The ticket's third path is deliberately not here

PT-4481 also lists the `isFirstWindow` gate on the shutdown-latch reset. It is left alone, because **the ticket is stale on that point**: the gate gained a discriminating e2e case in `a76e64e9a0d` — `window-close-rule.spec.ts:507`, "creating a window while the question is open does not strand the quit it might still receive" — which landed *after* the sweep commit the ticket was written from, and the latch primitive underneath is already unit-covered. A `count === 0` seam would only give the rule a second home.

## Testing

Root workspace suite green: 213 files, 3070 passed, 1 skipped — up from 212 / 3064, exactly the one file and six tests added here and nothing else moved. `src/main` alone: 47 files, 854 passed. `typecheck:core`, `typecheck:erb` and the root `tsconfig.json` all report 0 errors; eslint and `format:check` clean.

**`papi.d.ts` is untouched** — the expectation was stated before running, and `build:types` leaves the file byte-identical (same checksum before and after), with no generated-artifact drift.

One limitation, stated rather than hidden: the full `npm run typecheck` does not pass in the nested worktree this was built in, and the cause is the worktree, not this change. It reports 50 duplicate-identifier errors, every one pairing a local `src/types/X.d.ts` against a path six levels up that climbs out of the nested worktree into the parent repo — extensions' `typeRoots` use relative paths that survive one level of nesting, not two. The control: **none of the 50 touch `main.ts` or either new file**, and the parent worktree typechecks clean at 0 errors on the same base commit.

## Risk Level

Low — a pure extraction with no behaviour change, plus new test coverage.

## CI

This PR targets a stack base (`pt-4286-window-close-rule`), and `test.yml` only fires on `pull_request` into `main`/`release-prep`/`hotfix-*`/`ai/main`, so GitHub reports no checks here. CI was run explicitly via `workflow_dispatch`: run [33584643285](https://github.com/paranext/paranext-core/actions/runs/33584643285) on `c7775516e87`. **Every job reports `success`, and that is not the whole story.** Read from the warn steps rather than the job conclusions, because `E2E tests (Windows/macOS)` carries `continue-on-error` and its conclusion is rewritten to `success` whether it passed or not:

- **Linux is genuinely green.** `E2E tests (Linux)` is blocking, and it passed. So did type checking, both unit suites, the formatting and lint gates, packaging, and the generated-artifact staleness gate (`Verify no files changed after build`) on all three platforms.
- **Windows and macOS smoke failed and was swallowed.** `Warn on non-blocking E2E failure` ran on both — its condition is `steps.e2e-win-mac.outcome == 'failure'`, so it firing *is* the failure. Windows reported 1 failed and 1 flaky of 13.

The failures are in `ui-interaction.spec.ts` — opening the About dialog from the Help menu, opening Settings from the menu, and the toolbar book/chapter control — all `locator.click` timeouts, none of which touches the close-all prompt this PR extracts. The workflow makes these non-blocking for exactly this reason, citing chronic macOS smoke flakiness (#2336, #2338, #2357) and occasional Windows flakes.

**These are the suite's known flakes, and two experiments say so rather than one hopeful reading.**

*The same commit, run twice.* Re-dispatching this exact commit (`c7775516e87`, run 33588187265) flipped Windows from failing to clean while macOS failed again. A commit cannot cause a failure that disappears when the same commit is re-run, so at least part of this is nondeterminism by definition.

*An unrelated branch, same failures.* #2758 is a pure refactor of the web-view service router off `main`, touching nothing this PR touches. Its Windows job (run 33588674627) failed the same way, concentrated in the same places: `ui-interaction.spec.ts:55` (About dialog), `:97` (theme toggle) and `:117`. `:97` is the very test `test.yml`'s own comment names as a known Windows flake.

For completeness about what an earlier revision of this section claimed: a single clean run on this PR's base commit was cited as showing this "is not the normal state of the branch". One run is a single sample, not a distribution, and for a suite the workflow itself calls chronically flaky it could not support that. The two experiments above are what the claim now rests on. None of it is being waved through — it is recorded here for a reviewer to weigh rather than discover.

---

AI-assisted — [session 1](https://claude.ai/code/session_01SbvrTvyCp9XVHedo6jdwRj)

https://claude.ai/code/session_01SbvrTvyCp9XVHedo6jdwRj

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2753)

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -9,7 +9,6 @@
 import {
   app,
   BrowserWindow,
-  dialog,
   ipcMain,
   powerMonitor,
   RenderProcessGoneDetails,
@@ -47,10 +46,7 @@ import { startNetworkObjectStatusService } from '@main/services/network-object-s
 import { registerPowerMonitorListeners } from '@main/services/power-monitor-logging.service';
 import { startProjectLookupService } from '@main/services/project-lookup.service-host';
 import { performShutdownTasks, performWindowCloseTasks } from '@main/shutdown-tasks';
-import type {
-  CloseAllAnswer,
-  WindowCloseDecision,
-} from '@main/services/window-close-decision.service';
+import type { WindowCloseDecision } from '@main/services/window-close-decision.service';
 import { performStartupTasks } from '@main/startup-tasks';
 import { startNotificationServiceRouter } from '@main/services/notification.service-router';
 import {
@@ -79,7 +75,6 @@ import {
   resetShutdownLatchesForNewSession,
   runShutdownTasksOnce,
   shouldWindowCloseAbortReadinessWait,
-  whenQuitRequested,
 } from '@main/services/shutdown-latch.service';
 import { setAppShutdownSignal } from '@main/services/rpc-server';
 import {
@@ -110,6 +105,7 @@ import {
   setFocusedWindowId,
   setWindowPendingContentPredicate,
 } from '@main/services/window-state.service';
+import { confirmCloseAllWindows } from '@main/services/close-all-prompt.service';
 import { decideWindowClose } from '@main/services/window-close-decision.service';
 import {
   assignEntryToWindow,
@@ -165,7 +161,6 @@ import {
 import { GET_METHODS } from '@shared/data/rpc.model';
 import { PROJECT_INTERFACE_PLATFORM_BASE } from '@shared/models/project-data-provider.model';
 import * as commandService from '@shared/services/command.service';
-import { localizationService } from '@shared/services/localization.service';
 import { logger } from '@shared/services/logger.service';
 import { readFile } from 'fs/promises';
 import { networkObjectService } from '@shared/services/network-object.service';
@@ -179,7 +174,6 @@ import { CommandNames, SettingTypes } from 'papi-shared-types';
 import {
   getErrorMessage,
   isPlatformError,
-  LocalizeKey,
   serialize,
   ThemeDefinitionExpanded,
   UnsubscriberAsyncList,
@@ -273,15 +267,6 @@ const PROCESS_CLOSE_TIME_OUT_MS = 2000;
  * screen a moment longer.
  */
 const WINDOW_CLOSE_TIME_OUT_MS = 10000;
-
-/**
- * How long the close-all question waits for its localized text before showing English instead.
- *
- * Short on purpose: the user has already clicked ✕ and the window is held open with nothing on
- * screen until the question appears, so a wait long enough to notice is worse than untranslated
- * text.
- */
-const CLOSE_PROMPT_LOCALIZE_TIME_OUT_MS = 3000;
 
 /** How long to coalesce a window's resize/move events before capturing its bounds */
 const BOUNDS_CAPTURE_DEBOUNCE_MS = 100;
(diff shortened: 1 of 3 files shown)


rolfheij-sil:
🤖 Claude: Folded into #2729 as cb35c113555 at Rolf's request to reduce the number of open PRs; the branch is kept.

PR #2752, opened 2026-09-02
PT-4286: Switching to Simple closes secondaries; Power reopens them

rolfheij-sil:
## Summary

Switching interface mode now decides how many windows the application has. Switching to Simple closes every window but the primary, keeping each closed window's layout and placement; switching back to Power opens a window again for each one that was kept. Until now nothing in the main process reacted to a mode change at all, so a Simple session could be left holding secondary windows its chrome has no way to reach, and switching back brought nothing back.

### Why review this

Part of the current multi-window epic (PT-4275). This is the second half of PT-4286 — the first half, the window-close rule, is #2729, which this is stacked on. It closes the last behaviour NN-6 depends on: a window's contents and location surviving a mode switch, not just a restart.

## Changes

- **The application reacts to a live mode change.** One app-level subscription in the main process, taken after the window restore and torn down on the way out. It is woken by *any* settings write — every setting shares one data provider data type — so it acts only when the mode has actually changed.
- **No new record of "the windows a Power session had open".** The persisted entry list already is that set: a window closing for a mode switch keeps its entry and lets go only of its runtime id, which is the same shape as an entry the session never restored. Switching back creates a window from each entry with no live window. This also means a quit taken from Simple mode costs nothing extra — the entries are already on disk.
- **The primary window is the survivor, and its role does not move.** It is the entry Simple mode restores and the only one allowed the legacy layout fallback, so the surviving window and the window a later launch restores are the same one by construction. There is no prompt: nothing is lost, and the windows return on the way back.
- **Only the primary runs the renderer-side switch to Simple.** That switch writes state the whole application shares — it starts a Send/Receive, applies the administrator's shared layout, records a recently-opened project, and writes a browser-storage cache under one key for every window. A window being closed by the same switch would duplicate all four, and could settle on a different project than the survivor when the cache is cold. Each window asks the main process whether it holds the primary role and returns immediately when it does not.
- **Simple mode refuses a further window.** Simple mode has no chrome that can reach a second window. Extensions are unaffected: opening a web view into its own window, and moving one to a new window, both already read the mode and degrade before any window is created — so the refusal only ever reaches `platform.createWindow`, whose menu item is already hidden in Simple mode.
- **A window closing for a mode switch captures its placement first**, because the debounced capture is cancelled once a window is gone and nothing else on that path records it.
[...]

PR #2754, opened 2026-09-02
PT-4285: Skip bounds captures taken while a window changes display

rolfheij-sil:
## Summary

A window's placement is no longer saved while the platform is still making up its mind about the window's scale. Dragging a window between two displays with different scale factors leaves a short window of time in which the size we would persist is wrong by about a quarter — and restoring it later brings the window back at that wrong size.

### Why review this

Part of the current multi-window epic (PT-4275), closing the one hazard PT-4285's own description names as outstanding. Its Definition of Done is otherwise delivered on main. The hazard was found by PT-4276's Windows column on real mixed-DPI hardware, recorded there as "exactly the kind of state a layout-persistence feature could snapshot and later restore wrongly", and this is that snapshot being prevented.

## The mechanism

While a window spans the boundary between two displays of different scale factors, Win32 and Chromium disagree: `GetDpiForWindow` reports the DPI context inherited from the crossing (120) while `MonitorFromWindow` resolves to the 96-DPI display and Chromium renders at `dpr` 1. The window is physically ~25% larger than the layout it contains. It looks fine and it resolves itself once the window is fully on one display — but a bounds capture taken during it is wrong, and persists.

The straddle is not the dangerous part. Bounds captured mid-straddle lie in no single display, and the restore path (`ensureBoundsVisibleOnSomeDisplay`) already requires containment within one display, so it discards them and re-places at default size. Annoying, not corrupting.

**The damaging capture is the one just after the window lands.** The geometry is already on one display — so it passes containment and *will* be restored — while the two DPI answers have not yet met. A containment check alone does not catch it, which is why this needs a settle period rather than a geometry test.

## Changes

- **`areCapturedBoundsTrustworthy(bounds, displays, lastAcceptedDisplayId, msSinceDisplayChange)`** — a pure helper in `window-bounds.util.ts`, alongside the containment rule it reuses. It refuses two moments: bounds lying in no single display (the straddle), and bounds on a display the window has only just reached, before `DISPLAY_SETTLE_MS` has elapsed (the tail). A window moved *within* one display has crossed nothing and is trusted immediately.
- **`DISPLAY_SETTLE_MS = 500`**, deliberately longer than the 100 ms capture debounce: the debounce answers "has the user stopped moving the window", this answers "has the platform finished agreeing with itself", and the second outlasts the first. It sits beside the helper that uses it, with a pointer comment on the debounce so both timings read together.
- **No new plumbing.** The capture already had a mode for "these bounds are not worth trusting" — the branch that omits them for a maximized, full-screen or minimized window — and `updateWindowBounds` already keeps the last placement it was given when a capture carries none (`slot.entry.bounds = boundsState.bounds ?? slot.entry.bounds`). So this is one more condition on that existing branch: every capture call site inherits it unchanged, the maximized and full-screen flags still record, and there is no schema change.
- A new `IdentifiedDisplayLike` type carries the display id the check needs, leaving `DisplayLike` and every existing caller of `ensureBoundsVisibleOnSomeDisplay` untouched.

**Decided product consequence:** a quit taken while the window straddles restores the placement the window last held, rather than the straddle. That is the intended behaviour — the alternative is restoring a size that is knowably wrong.

## Verification

**`papi.d.ts` delta expected: none** — this adds no PAPI surface. Confirmed none after `build:types`.

Five unit tests, written before the helper existed and watched to fail against it. **Both mutation directions were applied and each watched to turn the suite red:**

- *Containment check removed* → "bounds straddling two displays are not trustworthy" and "bounds on a display the window has only just reached" both fail.
- *Always refuse* → "bounds on a display the window has settled on" and "bounds moved within the same display are trustworthy at once" both fail. **This is the control that matters**: without it, a guard that simply never saved anything would pass every straddle test while silently freezing every user's saved placement — a regression no amount of straddle testing would catch.

`build:main`, `typecheck`, `format:check` clean; `npm run lint` reports zero errors, and the three changed files lint clean individually. Full suite: 2951 passing, 1 skipped, across 207 files.

Not covered here: the behaviour needs real mixed-DPI hardware to observe end to end, which is the same constraint PT-4276 recorded (macOS cannot reach the straddle state on a default configuration; WSLg cannot construct a mixed-DPI pair at any monitor count). The logic is therefore pinned at the unit level, where it is deterministic, rather than by an environment that cannot produce the fault.

## AI Involvement

AI-assisted. Design and implementation were produced with Claude Code under review, from a design agreed before any code was written; the mechanism was read against PT-4276's recorded findings and the current capture/restore code rather than inferred, and the mutation testing above was the check on the tests themselves. The human developer is the author and reviewed the result.

## Risk Level

Low — one condition inside an existing guard, on a path that already has a well-defined "no bounds this time" mode. The failure it could introduce (refusing too often) is exactly what the always-refuse mutation control pins.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2754)

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -120,6 +120,7 @@ import { createWindowEmptinessHandler } from '@main/services/window-emptiness.ut
 import {
   DEFAULT_WINDOW_HEIGHT,
   DEFAULT_WINDOW_WIDTH,
+  areCapturedBoundsTrustworthy,
   ensureBoundsVisibleOnSomeDisplay,
 } from '@main/window-bounds.util';
 import {
@@ -263,6 +264,8 @@ const WINDOW_CLOSE_TIME_OUT_MS = 10000;
 
 /** How long to coalesce a window's resize/move events before capturing its bounds */
 const BOUNDS_CAPTURE_DEBOUNCE_MS = 100;
+// A window that has just crossed to another display is left untrusted for longer than this
+// debounce — see `DISPLAY_SETTLE_MS` in `window-bounds.util.ts`, which says why.
 
 /**
  * How a window being created relates to the persisted window-layouts structure: it restores a saved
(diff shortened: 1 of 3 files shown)


rolfheij-sil:
🤖 Claude: Folded into #2752 as 09049089039 at Rolf's request to reduce the number of open PRs; the branch is kept.

PR #2726, opened 2026-08-26
Make the isolated e2e suite runnable, and stop it failing silently

rolfheij-sil:
Makes the `isolated` e2e suite something you can actually run and trust locally.

## The core problem this fixes

Playwright reports a test that never executed — because a worker died, or a `beforeAll` threw — with the same outcome it uses for a deliberate `test.skip()`. On Linux that prints as "did not run", which reads as housekeeping. `tests/isolated/find/` sat that way and silently absorbed a fixture refactor, a menu-label rename, a Tailwind prefix migration and a UI redesign without a single visible failure.

A new reporter (`e2e-tests/reporters/no-silent-skips.reporter.ts`) now fails any run in which tests are reported skipped that nobody asked to skip, so a suite can no longer rot quietly. It distinguishes a deliberate skip from a lost test the way Playwright's own `computeTestCaseOutcome` does, and it tolerates a test lost on one attempt that passes on a retry.

## Environment, not architecture

Most of what needed fixing turned out to be environment and shared state:

- **DevTools was eating the window.** A docked panel takes ~555px of the renderer's layout viewport, so geometry-sensitive specs ran against a window they were never sized for. Automated launches now set `PT_NO_DEVTOOLS=true` (the one product change here: `src/main/main.ts`).
- **The window-size check could not fail.** It called `setViewportSize()` on a CDP-attached page, which is an emulation override that sets `innerWidth` itself — so the assertion read back its own request. It now asserts `outerWidth`/`outerHeight` against a size the spec declares with `test.use({ windowSize })`.
- **There were two tracked `refresh.sh` scripts** with different content, and which one you ran decided whether your tests could run at all. The root one is now a thin delegate.
- **Settings leak between runs.** `preConfigureSettings` merges into the shared gitignored settings file and only restores in an `afterAll`, so a killed run never restores. A suite can now declare `test.use({ requiredInterfaceMode })`, which verifies the pin actually took effect instead of failing much later on an element the other mode never renders.
- **e2e teardown killed every developer's processes, machine-wide.** `global-teardown.ts` ran `npm run stop` unconditionally, which matches by process *name* and so kills every Electron and dotnet on the box — not just this run's. On a shared machine that reaches into whatever anyone else is doing; it killed one of this branch's own verification runs. The sweep is now gated behind `process.env.CI`, where the machine does belong to the run.
[...]

PR #2755, opened 2026-09-02
e2e: type-check the quarantined text-grid specs, unmask typechecks

rolfheij-sil:
Two e2e specs were excluded from `e2e-tests/tsconfig.json` so the rest of the tree could be type-checked at all. That left them checked by nothing, and their thirteen errors were real API misuse rather than noise. All thirteen are fixed and the exclusion is gone.

## The three mistakes

**A setting key the app never had.** The resource fixtures passed `isResourceShownByDefault`. Those values go straight into the app's `platformScripture.modelTexts` setting, whose field is `isInTextCollection` — and `isResourceShownByDefault` appears nowhere in `src/`, `extensions/src/` or `lib/`. The specs were writing a key the application ignores. All seven sites set it to `true` and every one of their tests then asserts a cell is visible, so the meaning carries over unchanged. The zoom spec's helper now returns the shared `FlaggedResourceItem` type instead of repeating the shape inline, which is how the two drifted apart.

**A page object mistaken for a Locator.** The zoom spec named the result of `openScriptureTextGrid` `frame` and called `.locator()` / `.getByRole()` on it. `ScriptureTextGrid` is not a Locator: it exposes a `FrameLocator` plus named locators, and its own docblock says to use `.frame` for anything the named ones do not cover. The local is now `stg` and the calls go through `stg.frame`. The sibling spec already did this correctly.

**A `FrameLocator` API that does not exist.** `FrameLocator` has no `evaluate` in Playwright 1.58. The focus assertion now goes through a locator inside the frame, which is the point of that check — reading `document.activeElement` from the parent document would pass even when focus had escaped the iframe.

## One failure no longer hides the rest

The root `typecheck` script chained its four checks with `&&`, so the first failure stopped the others running: an error under `e2e-tests/` meant the workspace packages were not checked at all.

Demonstrated rather than argued. With one file broken under `e2e-tests/` and one broken in a workspace:

- the old chain reported only the e2e error and never ran the workspaces;
- the new one reports both.

`concurrently` was already a dependency, so nothing is added. `-m 1` keeps the four checks sequential: four `tsc` invocations at once multiply peak memory and interleave their output. There is deliberately no `--kill-others-on-fail` — every check must run and every failure must be reported.

## Verification

`tsc -p e2e-tests/tsconfig.json` is green with the exclusion removed, and the full chain runs all four checks green on a clean tree.

One note for anyone repeating this: these checks were run from the main worktree, not the nested one. A fresh worktree resolves the root `node_modules` fine — `tsc -p e2e-tests` there reproduced the identical thirteen errors — but the extension workspaces have their own `node_modules`, so `typecheck --workspaces` fails there with spurious "Duplicate identifier" errors. A green e2e check in a fresh worktree says nothing about the workspaces.

Stacked on #2726, so no CI runs automatically. Dispatched manually:
[run 33586210111](https://github.com/paranext/paranext-core/actions/runs/33586210111),
confirmed against `3ccda734e991da37dffb0263093036ceb563419c`.

Closes the deferral recorded against the quarantine in PR #2726's review round.

AI-assisted — [session](https://claude.ai/code/session_01QyoVc1Cd9L2RcuDGE3T3FT)

https://claude.ai/code/session_01QyoVc1Cd9L2RcuDGE3T3FT

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2755)


rolfheij-sil:
🤖 Claude: Folded into #2726 as fa42d4db3b4 at Rolf's request to reduce the number of open PRs; the branch is kept.

PR #2757, opened 2026-09-02
e2e: clean up only this checkout's processes, never the machine's

rolfheij-sil:
E2E teardown ended by running `npm run stop`, unconditionally (`e2e-tests/global-teardown.ts`). That script matches by process **name** and kills every `electron` and `dotnet` process on the machine, so a run finishing in one checkout took down the developer's own app, any app a CDP-based suite was attached to, and other checkouts' runs on the same box. On a shared machine this is the difference between cleaning up after yourself and cleaning up after everyone.

Two decisions now govern it, and both are testable without killing anything.

## Which cleanup each environment gets

The old code asked one flag to decide two different things, and got both wrong. It skipped everything outside CI — so the scoped sweep ran only on a CI runner, which is single-tenant and the one place scoping buys nothing, while a developer's machine, the reason to scope at all, cleaned up nothing. And it treated the two sweeps as alternatives, so wherever scoping was possible the name sweep's wider reach was lost.

They answer different questions, so they are no longer exclusive:

| | Linux | macOS / Windows |
|---|---|---|
| **off CI** | scoped sweep | nothing |
| **on CI** | scoped sweep **and** name sweep | name sweep |

Scoping asks which processes are this checkout's. It needs `/proc`, and it now runs on Linux in every environment — that is what lets a developer's box clear its own leaks without touching a peer's checkout. The name sweep asks whether anything is left over anywhere; it is safe only where the machine belongs to the run, so CI gets it in addition, keeping exactly the coverage it had. Off CI without `/proc` there is nothing safe to do, and that is the only cell that does nothing.

Where the flag is still read, it is read as a real boolean. The previous form treated `CI=false` and `CI=0` as yes, because both are non-empty strings — ordinary wrapper and IDE idioms, not exotic cases.

## What gets swept

Selection is by working directory, never by process name. A process is a candidate only if it is running inside this checkout, and:

- a sibling path that merely shares the prefix (`paranext-core-other` beside `paranext-core`) does not count;
- **a worktree of this repository does not count either.** Worktrees live at `<root>/.claude/worktrees/<name>` — inside the root by path, but a different checkout and usually a different run. Containment alone would have claimed every one of them, which is the same cross-checkout kill this change exists to prevent. Selection is relative to whichever root is asking, so a run *inside* a worktree still cleans up after itself;
- a process whose working directory cannot be read is left alone — unreadable means another user's process, and guessing in that direction is exactly how a neighbour's app gets killed;
- this process and its ancestors are excluded, so cleanup cannot kill the run performing it;
- being inside the checkout is necessary but **not sufficient**: npm, node and the shell running the suite are rooted there too, so only the app's own process names are considered;
- the root itself is resolved through symlinks. A working directory read from `/proc` is always fully resolved, so a run launched through a symlinked checkout would otherwise compare its processes against a path none of them can match and clean up nothing at all.

Those names are matched as the kernel reports them. `/proc/<pid>/comm` is capped at fifteen characters, so `ParanextDataProvider` appears there as `ParanextDataPro`; the full spelling is kept in the source and truncated at the comparison.

The dev-server kill by pid file is unchanged — it was already scoped to what the run started.

## Platforms without `/proc`

Scoped selection needs `/proc`, so on macOS and Windows the old name sweep is still what runs. Its failure is caught: having nothing to stop is a routine outcome there, and the shell-outs behind it can exit non-zero or time out — without the catch, a run whose tests all passed would end as a failed one.

## The one consequence to look at

On Linux the scoped sweep now runs off CI, so an e2e run will terminate an `electron` or `dotnet` process inside its own checkout that it did not start — a dev app left running in that same checkout.

In practice neither config can reach that state. The launch-based config's `globalSetup` throws when port 8876 is already bound (`Port 8876 is already in use`), and Playwright does not run `globalTeardown` when `globalSetup` throws, so the run aborts before any test rather than killing anything. The config used when an app *is* deliberately running — `playwright-cdp.config.ts`, for the attached specs — declares no `globalSetup` or `globalTeardown` at all, so it never reaches this code. The residual case is an app started *during* a run in the same checkout, which teardown will kill.

## Verification

Twenty unit tests over the two decisions, including all six cells of the dispatch matrix, the sibling-prefix case, the nested-worktree case (both directions), the symlinked-root case, the unreadable-cwd case, the self-exclusion case, and the comm-truncation case. Each guard was checked by mutation — removed, then confirmed a test goes red.

Proven live as well: two processes named `electron` were started, one running inside the checkout and one outside it, and offered to the selection.

```
WOULD KILL: [1971257, 1973080]
inside(1973080) selected?  true
outside(1973081) selected? false

after: inside  — killed
       outside — survived
```

The unscoped sweep was never run on this machine to test this; the scoped function was called directly.

## Note for reviewers

This branch is off `main`, where `tsc -p e2e-tests/tsconfig.json` already fails with 14 pre-existing errors in `enhanced-resources` and `internet-settings`. None are from this change — my own files typecheck clean — and #2755 fixes them.

#2726 carries an earlier, weaker version of this gate; it will restack onto this branch and drop that hunk, along with the `vitest.config.ts` include line, which both branches touch.

AI-assisted — [session](https://claude.ai/code/session_01QyoVc1Cd9L2RcuDGE3T3FT)

https://claude.ai/code/session_01QyoVc1Cd9L2RcuDGE3T3FT

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2757)

### DIFF
--- a/e2e-tests/global-teardown.ts
+++ b/e2e-tests/global-teardown.ts
@@ -56,6 +56,8 @@ export default async function globalTeardown(_config: FullConfig): Promise<void>
     );
   } else if (swept === 'by-name') {
     console.log(`Cleanup: swept by process name (${process.platform} has no /proc to scope by).`);
+  } else if (swept === 'by-name-failed') {
+    console.log('Cleanup: the process-name sweep reported nothing to stop, or could not complete.');
   } else if (pids.length > 0) {
     console.log(
       `Cleanup: terminated ${pids.length} process(es) under ${rootDir}: ${pids.join(', ')}`,
--- a/e2e-tests/global-teardown.ts
+++ b/e2e-tests/global-teardown.ts
@@ -32,14 +32,13 @@ export default async function globalTeardown(_config: FullConfig): Promise<void>
     }
   }
 
-  // Sweep up anything a crashed fixture left behind — this checkout's processes, and only when the
-  // machine belongs to the run.
+  // Sweep up anything a crashed fixture left behind.
   //
-  // Selection is by working directory rather than process name wherever that is possible: a name
-  // match reaches every electron and dotnet process on the machine, including the developer's own
-  // app, any app a CDP-based suite is attached to, and other checkouts' runs on a shared box. See
-  // e2e-tests/scoped-cleanup.ts, which also explains why non-Linux CI runners still match by name.
-  const { swept, pids } = runCleanup(
+  // Selection is by working directory wherever that is possible: a name match reaches every
+  // electron and dotnet process on the machine, including the developer's own app, any app a
+  // CDP-based suite is attached to, and other checkouts' runs on a shared box. See
+  // e2e-tests/scoped-cleanup.ts for which sweeps each environment gets and why.
+  const { scoped, byName, pids } = runCleanup(
     { ciFlag: process.env.CI, platform: process.platform, root: rootDir },
     {
       killUnderRoot: killProcessesUnderRoot,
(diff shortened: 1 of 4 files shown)


rolfheij-sil:
🤖 Claude: Folded into #2726 as 1797a88239d at Rolf's request to reduce the number of open PRs; the branch is kept.

PR #2, opened 2023-01-23
First part of issue#2: Add electron-react-boilerplate

irahopkinson:
- First commit adds the latest `electron-react-boilerplate` but without replacing the LICENSE and README.md files.
- Second commit updates to Paranext including how to run it in the README.
- The merge should not squash (fast forward would be good).
- Preview:
image

PR #2513, opened 2026-07-07
Warm Intl.* lazy ICU init in vitest setup to fix first-use CI flake

lyonsil:
## Summary

Fixes an intermittent CI timeout in `src/renderer/hooks/use-project-picker-data.hook.test.ts`. The test **"returns currentProject from the first open Scripture Editor web view"** would occasionally fail on `windows-latest` with `Error: Test timed out in 5000ms`, while its six sibling tests in the same file passed in ~14ms each.

### Root cause (measured, not the settle logic)

That test is the **only** one in the file whose code path constructs `Intl.DisplayNames` — via `resolveLanguage`, reached only when a project with a real language value is open. The **first** construction of an `Intl` formatter type in a worker process performs a one-time lazy ICU dataset load. Measured cold vs warm first-use cost (fresh process each):

| Constructor | Cold | Warm |
| --- | --- | --- |
| `Intl.DisplayNames({type:'language'})` | **20.3 ms** | 0.035 ms |
| `Intl.NumberFormat` | 11.2 ms | 0.048 ms |
| `Intl.Collator` | 9.7 ms | 0.017 ms |
| `Intl.DateTimeFormat` | 9.5 ms | 0.054 ms |
| `Intl.RelativeTimeFormat` | 8.9 ms | 0.034 ms |
| `Intl.ListFormat` | 7.9 ms | 0.014 ms |

Every `Intl.*` constructor carries an ~8–20 ms cold first-use cost (~500× its warm cost). On a cold, memory-pressured CI worker that first-touch initialization can spike into the seconds, and it runs **synchronously inside the test's timeout window** — so only the one test that first-constructs `Intl.DisplayNames` flaked, while siblings ran warm. (Instrumentation confirmed the first in-test construction costs ~20 ms and every later one ~0.04 ms.)

This is why prior fixes did not stick: they targeted the `settle()` / `waitFor` / `testTimeout` layer, which was never the cause.

## Changes

- **`vitest.setup.ts`** (new): a global setup that warms the `Intl.*` constructors the codebase uses (once per worker, before any timed test), so the one-time ICU init happens outside every test window.
- **`vitest.config.ts`**: wire it via `test.setupFiles`.

Test-infrastructure only — **no production code change**. The ~20 ms one-time cost is irrelevant in production (it happens once, on the first real project fetch).

## AI Involvement

[...]

PR #2759, opened 2026-09-02
ci: give react unit tests 15 s and stop the matrix failing fast

rolfheij-sil:
## Summary

Three branches failed CI today on three different `platform-bible-react` component tests, each a
timing-sensitive assertion that ran out of time on `windows-latest`:

| Run | Test |
|---|---|
| 33624120083 | `select-books-picker.component.test.tsx` — 5122 ms |
| 33622636517 | `comment-thread.component.test.tsx` — `CommentThread assignee state machine` |
| 33620874987 | the footnote caller-dropdown `waitFor` |

None of those branches touches `lib/platform-bible-react`, so the tests are not what changed: the
runner is contended and the budgets are too tight for assertions that wait on React state settling.

**There are two budgets, and they are not the same one.** Measuring the three failures separated
them:

| Failure | Elapsed | Budget crossed |
|---|---|---|
| `select-books-picker` | 5122 ms | vitest `testTimeout`, default 5 s |
| footnote caller-dropdown | 5099 ms | vitest `testTimeout`, default 5 s |
| `comment-thread` | **1263 ms** | testing-library `asyncUtilTimeout`, default **1 s** |

The comment-thread one is the reason this PR needed a second commit. It fails inside a bare
`waitFor`, which gives up after one second while vitest is still waiting patiently — so raising
`testTimeout` does not touch it, and an earlier version of this description wrongly claimed the
class was closed. That test runs in 71 ms on an idle machine.

## Changes

- **`testTimeout` goes to 15 s in the react `unit` project**, then (commit `4e51ab6f`, see the
  Addendum below) **in the four other projects that load the shared `vitest.setup.ts`** — the
  repo-root config, `platform-bible-utils`, `platform-bible-react`'s `scripts` project, and
  `extensions` — so `testTimeout` sits above `asyncUtilTimeout` (next bullet) everywhere the latter
  applies. Only the `storybook` project (browser mode, no shared setup file) keeps the 5 s default,
  so a genuine hang there is still bounded. 15 s absorbs the contention without making a real hang
  cheap.
- **Testing-library's `asyncUtilTimeout` goes to 5 s**, in the shared `vitest.setup.ts`. That setup
  file is loaded by five configs — the repo-root `vitest.config.ts`, `platform-bible-utils`,
  `platform-bible-react` (both its `unit` and `scripts` projects), and `extensions` — so this budget
  rises across all of them. The `extensions` suite leans on testing-library heavily too
  (comment-list, manage-books-dialog, use-bcv-sync-scroll, use-auto-search-debounce, and more), so
  its async budget genuinely rises there as well — it is not inert. Deliberate for the projects that
  wait on the same React state on the same runner; it is the wider of the two changes and should be
  read as one.
- **The build matrix stops failing fast.** This is the half that turns a flake into an outage: the
  blocking E2E suite runs only on Linux, so a Windows unit flake was cancelling the one job that
  produces the signal a PR is gated on. With `fail-fast: false`, a Windows wobble costs a red
  Windows job instead of "no PR can go green".

## Why both

Either alone leaves the problem. Raising the timeout reduces how often Windows trips, but any future
Windows failure still cancels the Linux E2E. Disabling fail-fast keeps the E2E signal, but leaves a
red Windows job on every affected PR. Together, the common case stops happening and the uncommon
case stops being fatal.

## Testing

`comment-thread.component.test.tsx` passes under the new config (8 tests) — the file that motivated
the second budget. The repo-root suite passes with the raised async budget (208 files, 2955 tests),
which is the half with the wider blast radius. The workflow YAML parses with `fail-fast: false` on
the matrix job and the OS list unchanged. No product code is touched.

**What this does not demonstrate.** A green run here is the runner having a good hour, not evidence
the fix works: the value of a raised budget shows up as the absence of future red, which no single
run can show. What is verifiable is the mechanism — the three failures were measured against the two
budgets, and each is now above the one it crossed.

## Risk Level

Low, and self-limiting. Neither budget can mask a failure — a test that genuinely fails still fails,
and a `waitFor` that is never going to succeed still gives up; only the deadline for something merely
slow moves. Disabling fail-fast makes CI slower to
report on a broken commit (all three jobs run to completion) and costs more runner minutes; that is
the deliberate trade for not losing the E2E signal.

## AI Involvement

AI-assisted — [session](https://claude.ai/code/session_01VumBqDNMeTmKXenGEjX5T8). Claude read the
three failing runs, located the react package's own vitest config and its three named projects,
made both changes, verified the previously-failing test file passes and the workflow YAML parses,
and drafted this description. The scope — which project's timeout, and pairing it with `fail-fast`
— was directed rather than chosen by the model.

https://claude.ai/code/session_01VumBqDNMeTmKXenGEjX5T8

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2759)

## Addendum — the risk statement now holds across all five projects

Review found that the Risk Level section's claim — "a `waitFor` that is never going to succeed
still gives up [with a real, diagnosable error]" — held only for the platform-bible-react `unit`
project as originally written. The other four projects that load the shared `vitest.setup.ts`
(the repo root, `extensions`, `platform-bible-utils`, and platform-bible-react's `scripts`
project) had no `testTimeout` override, so the new `asyncUtilTimeout: 5000` landed exactly equal
to vitest's own 5000 ms default there. Vitest's per-test clock starts before the test body runs,
ahead of testing-library's own `waitFor` timer, so it always won that race — a failing `waitFor`
in those four projects reported vitest's bare `Test timed out` with no DOM dump instead of
testing-library's richer error.

Fixed in commit `4e51ab6f1e33857a22178599c88de083b4d10c65`: `testTimeout` is now raised to
15000 ms in those four projects too, matching the value the `unit` project already used, so
`asyncUtilTimeout` sits below every consuming project's `testTimeout` everywhere it's loaded. The
Risk Level statement now holds across all five projects, not just one.

### DIFF
--- a/lib/platform-bible-react/vitest.config.ts
+++ b/lib/platform-bible-react/vitest.config.ts
@@ -24,6 +24,10 @@ const workspace = defineConfig({
           globals: true,
           environment: 'jsdom',
           setupFiles: [intlWarmupSetup],
+          // Timing-sensitive component tests here wait on React state settling, and on a contended
+          // windows-latest runner that wait crosses the 5 s default while the assertion itself is
+          // sound. 15 s absorbs the contention and still bounds a genuine hang.
+          testTimeout: 15000,
         },
       },
       // Node.js tests for build scripts
--- a/lib/platform-bible-react/vitest.config.ts
+++ b/lib/platform-bible-react/vitest.config.ts
@@ -27,6 +27,9 @@ const workspace = defineConfig({
           // Timing-sensitive component tests here wait on React state settling, and on a contended
           // windows-latest runner that wait crosses the 5 s default while the assertion itself is
           // sound. 15 s absorbs the contention and still bounds a genuine hang.
+          // The shared vitest.setup.ts also raises testing-library's asyncUtilTimeout to 5 s; this
+          // budget must stay comfortably above that one for testing-library's richer failure to
+          // ever be reachable.
           testTimeout: 15000,
         },
       },
--- a/lib/platform-bible-react/vitest.config.ts
+++ b/lib/platform-bible-react/vitest.config.ts
@@ -37,6 +37,12 @@ const workspace = defineConfig({
           include: ['scripts/**/*.test.ts'],
           environment: 'node',
           setupFiles: [intlWarmupSetup],
+          // The shared vitest.setup.ts raises testing-library's asyncUtilTimeout to 5 s so a
+          // `waitFor` slowed by CI contention gives up on its own before the test's overall
+          // budget, rather than sharing vitest's own testTimeout and losing the race to it. That
+          // budget must stay comfortably below this one for testing-library's richer failure to
+          // ever be reachable.
+          testTimeout: 15000,
         },
       },
       // Browser tests for Storybook
--- a/vitest.setup.ts
+++ b/vitest.setup.ts
@@ -53,6 +53,13 @@ new Intl.ListFormat(locale).format(['a', 'b']);
 // that were queued while cb() ran — a bare Promise.resolve() only drains one microtask level
 // and misses multi-await async chains (e.g., two sequential sendCommand() awaits in a hook).
 configure({
+  // Testing-library's own budget, which is separate from vitest's `testTimeout` and is NOT raised by
+  // it: a bare `waitFor` gives up after this long and fails the test while vitest is still content.
+  // Timing-sensitive component tests here spend it waiting for React state to settle, and on a
+  // contended windows-latest runner that wait crosses 1 s while the assertion is sound — the same
+  // class as the per-test budget, one level down. Five seconds absorbs the contention and still
+  // bounds a wait that is never going to succeed.
+  asyncUtilTimeout: 5000,
   asyncWrapper: async (cb) => {
     // Temporarily clear the React act environment flag so that `waitFor` polling intervals don't
     // produce "not wrapped in act" warnings (mirrors the intent of @testing-library/react's original
--- a/.context/standards/Testing-Guide.md
+++ b/.context/standards/Testing-Guide.md
@@ -395,6 +395,8 @@ export default defineConfig(async () => {
       // Warms the lazy one-time ICU init behind Intl.* so it never lands inside a test's
       // timeout window on a slow CI worker. See vitest.setup.ts for the rationale.
       setupFiles: ['./vitest.setup.ts'],
+      // Must stay comfortably above vitest.setup.ts's asyncUtilTimeout — see note below.
+      testTimeout: 15000,
       include: ['src/**/*.test.ts', 'src/**/*.test.tsx'],
     },
   };
@@ -404,6 +406,16 @@ export default defineConfig(async () => {
 > The workspace configs (`extensions/vitest.config.ts`, `lib/platform-bible-utils/vite.config.ts`,
 > `lib/platform-bible-react/vitest.config.ts`) reference this same repo-root `vitest.setup.ts` via a
 > relative `setupFiles` path, so every vitest worker warms Intl once before any timed test.
+
+> **`testTimeout` vs. `asyncUtilTimeout`.** `vitest.setup.ts` also raises testing-library's own
+> `asyncUtilTimeout` — the deadline a bare `waitFor`/`findBy*` gives up at — independently of
+> vitest's per-test `testTimeout`. `testTimeout`'s clock starts at the beginning of the test body,
+> while a given `waitFor` call only starts its own `asyncUtilTimeout` countdown when that call is
+> reached, so whichever deadline elapses first wins. Every project that loads this shared setup
+> file must keep its own `testTimeout` comfortably above `asyncUtilTimeout` — not just nominally
+> above it, since any work a test does before reaching the `waitFor` call eats into that margin —
+> or a failing `waitFor` is reported as vitest's bare timeout instead of testing-library's richer
+> error (with its DOM dump), silently losing the more useful failure message.
 
 ### Key Dependencies
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,6 +28,9 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     strategy:
+      # A unit-test flake on one runner must not cancel the others: the blocking E2E suite runs only
+      # on Linux, so cancelling it because Windows was slow leaves the PR with no signal at all.
+      fail-fast: false
       matrix:
         # Make sure these match with the env values above
         os: [windows-latest, macos-15, ubuntu-22.04]
--- a/extensions/vitest.config.ts
+++ b/extensions/vitest.config.ts
@@ -13,6 +13,11 @@ const config = defineConfig({
       path.resolve(__dirname, '../vitest.setup.ts'),
       path.resolve(__dirname, '../vitest.setup.node.ts'),
     ],
+    // The shared vitest.setup.ts raises testing-library's asyncUtilTimeout to 5 s so a `waitFor`
+    // slowed by CI contention gives up on its own before the test's overall budget, rather than
+    // sharing vitest's own testTimeout and losing the race to it. That budget must stay
+    // comfortably below this one for testing-library's richer failure to ever be reachable.
+    testTimeout: 15000,
     include: ['src/**/*.test.ts', 'src/**/*.test.tsx', 'lib/**/*.test.ts', 'lib/**/*.test.tsx'],
   },
   resolve: {
--- a/lib/platform-bible-utils/vite.config.ts
+++ b/lib/platform-bible-utils/vite.config.ts
@@ -47,6 +47,11 @@ const config = defineConfig({
       path.resolve(__dirname, '../../vitest.setup.ts'),
       path.resolve(__dirname, '../../vitest.setup.node.ts'),
     ],
+    // The shared vitest.setup.ts raises testing-library's asyncUtilTimeout to 5 s so a `waitFor`
+    // slowed by CI contention gives up on its own before the test's overall budget, rather than
+    // sharing vitest's own testTimeout and losing the race to it. That budget must stay
+    // comfortably below this one for testing-library's richer failure to ever be reachable.
+    testTimeout: 15000,
   },
 });
 export default config;
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -11,6 +11,11 @@ const config = defineConfig(async () => {
       // Warms the lazy one-time ICU init behind Intl.* so it never lands inside a test's timeout
       // window on a slow CI worker. See vitest.setup.ts for the full rationale.
       setupFiles: ['./vitest.setup.ts'],
+      // vitest.setup.ts raises testing-library's asyncUtilTimeout to 5 s so a `waitFor` slowed by
+      // CI contention gives up on its own before the test's overall budget, rather than sharing
+      // vitest's own testTimeout and losing the race to it. That budget must stay comfortably
+      // below this one for testing-library's richer failure to ever be reachable.
+      testTimeout: 15000,
       include: [
         'src/**/*.test.ts',
         'src/**/*.test.tsx',


lib/platform-bible-react/vitest.config.ts
```diff
@@ -24,6 +24,10 @@ const workspace = defineConfig({
           globals: true,
           environment: 'jsdom',
           setupFiles: [intlWarmupSetup],
+          // Timing-sensitive component tests here wait on React state settling, and on a contended
+          // windows-latest runner that wait crosses the 5 s default while the assertion itself is
+          // sound. 15 s absorbs the contention and still bounds a genuine hang.
+          testTimeout: 15000,
```
lyonsil, line 33: **#2** - medium · confirmed

Both budgets are raised unconditionally even though the diagnosis in their own comments is windows-latest CI contention, so every local run pays the wider deadline too.

Neither testTimeout: 15000 nor asyncUtilTimeout: 5000 is gated on process.env.CI or process.platform. Measured: a failing waitFor now takes 5006 ms locally where the pre-PR default gave 1000 ms, and a locally-failing test now takes 15 s rather than 5 s to report. The repo does have the idiom available (e2e-tests/playwright.config.ts:16-17 uses `forbidOnly: !!process.env.CI` and `retries: process.env.CI ? 2 : 1`), but a repo-wide grep finds process.env.CI only in e2e-tests and .erb/scripts/notarize.js:11 - never in a vitest config - and the adjacent ICU warm-up in this same setup file (vitest.setup.ts:11-33), added for a "cold, memory-pressured CI worker" flake, is deliberately unconditional. So CI-gating vitest would be a new pattern here, and the local/CI-parity counter-argument is legitimate. What lifts this above taste is the consequence it shares with #1: because asyncUtilTimeout lives in the shared setup file while only the react `unit` project got the matching testTimeout, the other three projects now have the two budgets exactly equal and lose testing-library's error entirely - a diagnostics regression on CI as well as locally. Keeping asyncUtilTimeout strictly below every project's testTimeout fixes both at once and shrinks the local penalty.

_Verification: The mechanics were measured directly (5006 ms vs 1000 ms for a failing waitFor; nothing gates either value). The cited playwright idiom exists verbatim, but a repo-wide grep found process.env.CI in only two places, neither a vitest config, and the adjacent ICU warm-up precedent in the same file is deliberately unconditional - so CI-gating would be a new pattern and the parity counter-argument stands. The refuter raised it to CONFIRMED at medium on a consequence the original finding did not name: the shared-file placement means three projects now run asyncUtilTimeout exactly equal to their testTimeout, and a probe under the root project produced `Error: Test timed out in 5000ms.` with no DOM dump where the pre-PR 1000 ms produced the full TestingLibraryElementError._

Same fix as #1.

_Other findings in this file: #3_

---

**#3** - medium · downgraded on review

PR #2485 already removed a configure({ asyncUtilTimeout: 5000 }) from this repo for exactly the capping reason in #1, and this PR reinstates it repo-wide without saying why that conclusion no longer holds.

The repo's documented convention for CI timing flakes is to fix the determinism at source, not widen a deadline: src/renderer/hooks/use-project-picker-data.hook.test.ts:132-139 carries verbatim "This replaces RTL's waitFor, whose wall-clock deadline flaked on CPU-starved CI runners even though the underlying resolution order is fully deterministic. The iteration cap is a safety net against a genuine hang, not a tuned timeout." (a9803f66dbe / PR #2485), alongside the ICU warm-up (vitest.setup.ts:11-34, #2513) and the jsdom CSS-dump fix (#2600). Decisively, `git grep asyncUtilTimeout a9803f66dbe^` shows use-project-picker-data.hook.test.ts:96 previously held configure({ asyncUtilTimeout: 5000 }) - the same knob at the same value - which #2485 deleted because it "was capped by vitest's equal 5000ms per-test testTimeout, so it gave no real tolerance under CPU starvation". Root vitest.config.ts still declares no testTimeout, so this PR restores that exact capped configuration. The convention does NOT transfer to the testTimeout half (the caller-dropdown and select-books-picker failures are Radix/Lexical/userEvent tests driven through the real macrotask loop, with no deterministic mechanism to exploit), but it does transfer to the comment-thread failure that asyncUtilTimeout targets: that one resolves `canUserAssignThreadCallback={async () => true}` (comment-thread.component.test.tsx:126) through a bare await in a useEffect at comment-thread.component.tsx:150-151 - pure microtask, no timers, exactly #2485's shape, where a bounded act() drain makes the flake structurally impossible.

_Why this is downgraded: All three precedents exist and say what is claimed, but the convention does not transfer to the anchored testTimeout line: the two failures it addresses are userEvent/Lexical component tests running 154-417 ms locally against 5099-5122 ms on CI, with no single deterministic-resolution mechanism to exploit, so widening is defensible there. It does transfer to the asyncUtilTimeout half, and the evidence is stronger than filed - #2485 deleted this exact knob at this exact value for exactly the capping reason, and root vitest.config.ts still sets no testTimeout, so the capped configuration is reinstated. Kept at medium rather than confirmed-as-filed because half the claim is refuted on its own anchor._

Same fix as #1.

_Other findings in this file: #2_

vitest.setup.ts
```diff
@@ -53,6 +53,13 @@ new Intl.ListFormat(locale).format(['a', 'b']);
 // that were queued while cb() ran — a bare Promise.resolve() only drains one microtask level
 // and misses multi-await async chains (e.g., two sequential sendCommand() awaits in a hook).
 configure({
+  // Testing-library's own budget, which is separate from vitest's `testTimeout` and is NOT raised by
+  // it: a bare `waitFor` gives up after this long and fails the test while vitest is still content.
+  // Timing-sensitive component tests here spend it waiting for React state to settle, and on a
+  // contended windows-latest runner that wait crosses 1 s while the assertion is sound — the same
+  // class as the per-test budget, one level down. Five seconds absorbs the contention and still
+  // bounds a wait that is never going to succeed.
+  asyncUtilTimeout: 5000,
```
lyonsil, line 62: **#1** - high · confirmed

asyncUtilTimeout: 5000 equals the vitest testTimeout default in four of the five projects that load this setup file, so testing-library's own bail-out is unreachable there and every waitFor failure loses its DOM dump.

vitest.setup.ts is loaded by four configs / five projects: vitest.config.ts:13 (root), extensions/vitest.config.ts:13, lib/platform-bible-utils/vite.config.ts:47, and lib/platform-bible-react/vitest.config.ts:26 (`unit`) and :39 (`scripts`). Only the react `unit` project also got testTimeout: 15000; the rest keep vitest's 5000 default. Vitest's per-test deadline starts before the test body runs, while testing-library only arms its own timer when waitFor is reached, so vitest always wins the race. Reproduced three times independently: under the root config a never-satisfiable findByText now reports a bare `Error: Test timed out in 5000ms.` at 5.0 s with no DOM, where at the pre-PR 1000 ms default it reported `TestingLibraryElementError: Unable to find an element with the text: ...` plus the full rendered DOM at 1.0 s. Blast radius measured at 56 test files (39 under src/, 17 under extensions/), each of which also now takes 5x longer to report an unsatisfiable wait. This directly contradicts the PR body's risk statement that "a waitFor that is never going to succeed still gives up". The fix is to keep asyncUtilTimeout strictly below every consuming project's testTimeout (e.g. 4000), or raise testTimeout in the other four configs to match.

_Verification: Proven empirically by running probes under the repo-root config: the failure reported `Error: Test timed out in 5000ms.` with no DOM dump while the in-test log showed testing-library's own TestingLibraryElementError only being constructed at 5007 ms, i.e. vitest wins the race. The same probe with configure({ asyncUtilTimeout: 1000 }) (pre-PR behaviour) failed at 1007 ms with the full error plus DOM dump, and under the react `unit` project (the only one raised) the rich error surfaced normally. grep confirmed testTimeout appears nowhere in the repo except lib/platform-bible-react/vitest.config.ts:30, and the blast radius is if anything understated at 56 files rather than the ~38 filed. A second, independent refuter reached the same conclusion from the root project's side._

Same fix as #2, #3.

lyonsil (commented):
Review of the two raised budgets, focused on one thing: the relationship between
`asyncUtilTimeout` and `testTimeout` across the projects that load `vitest.setup.ts`.

The premise of the PR holds up — I checked the CI history and `windows-latest` `npm unit tests` is
a real and dominant failure mode. This is about how the two knobs interact, not whether the flake
is worth fixing.

`vitest.setup.ts` is loaded by four configs covering five vitest projects, but only the react
`unit` project also received `testTimeout: 15000`. In the other four, `asyncUtilTimeout: 5000` now
exactly equals vitest's own 5000 ms default — and vitest's clock starts first, so testing-library's
bail-out can never fire there.

**Suggested fix for all three comments below:** keep `asyncUtilTimeout` strictly below every
consuming project's `testTimeout` (4000 would do it), or raise `testTimeout` in the other four
configs to match. Either restores the headroom that lets testing-library report first.

lib/platform-bible-react/vitest.config.ts
rolfheij-sil: 🤖 Claude: Confirmed on both #2 and the `asyncUtilTimeout` half of #3 — this thread and the sibling one (on `vitest.setup.ts:62`) are the same underlying gap.

**#2.** The unconditional-raise observation is real, but as you noted yourself the CI-gating counter-argument is legitimate (no `process.env.CI` idiom exists in any vitest config today, and the adjacent ICU warm-up in the same setup file is deliberately unconditional) — so that half doesn't need a separate fix. The part that does, the part you tied it to, is the shared-file consequence: three projects besides `unit` had `asyncUtilTimeout` and `testTimeout` landing exactly equal, discarding testing-library's error entirely. That's closed by the same fix as #1.

**#3, `asyncUtilTimeout` half.** Confirmed — PR #2485 really did remove this exact `configure({ asyncUtilTimeout: 5000 })` for exactly the capping reason you cite, and root `vitest.config.ts` still had no `testTimeout`, so the capped configuration was reinstated. Agreed as filed on the `testTimeout` half too: downgrading was correct, since the caller-dropdown and select-books-picker failures are Radix/Lexical/userEvent-driven with no deterministic-resolution mechanism to exploit, so widening `testTimeout` for those two is appropriate as shipped.

**Fix and mechanism.** Between your two suggested options, I raised `testTimeout` to 15000 ms (matching the `unit` project) in the four projects that had no override — the repo root, `extensions`, `platform-bible-utils`, and platform-bible-react's `scripts` project — rather than lowering the shared `asyncUtilTimeout`. `extensions` leans on testing-library heavily (comment-list, manage-books-dialog, and more, all via plain `waitFor` with no explicit timeout), and the PR's own body already states that suite's async budget "genuinely rises" under the raised value; lowering `asyncUtilTimeout` would have clawed back exactly the headroom this PR was raising it for. This also extends the same local-penalty trade-off the PR already made for `unit` (a genuinely hanging test now takes up to 15 s to report locally instead of 5 s) to the four other projects — deliberate, and consistent with what was already accepted there.

Landed in commit `4e51ab6f1e33857a22178599c88de083b4d10c65` on `ci-react-unit-timeout-and-no-fail-fast`. Verified directly: a throwaway never-resolving `waitFor` under the repo-root config reported vitest's bare `Test timed out in 5000ms.` with no DOM dump before this commit, and testing-library's own richer error with the full DOM dump after it, at the same ~5 s mark.

vitest.setup.ts
rolfheij-sil: 🤖 Claude: Confirmed — #1 is correct as filed. Four of the five projects loading this shared `vitest.setup.ts` (the repo root, `extensions`, `platform-bible-utils`, and platform-bible-react's `scripts` project) had no `testTimeout` override, so `asyncUtilTimeout: 5000` landed exactly equal to vitest's own 5000 ms default. Vitest's per-test clock starts before the test body runs, ahead of testing-library's `waitFor` timer, so it always won that race — a failing `waitFor` in those four projects reported vitest's bare timeout with no DOM dump instead of testing-library's richer error, contradicting the PR's own risk statement for those projects.

Fixed in commit `4e51ab6f1e33857a22178599c88de083b4d10c65` on `ci-react-unit-timeout-and-no-fail-fast`: raised `testTimeout` to 15000 ms in all four of those projects (matching the value already used by the `unit` project) rather than lowering the shared `asyncUtilTimeout`, so the async budget you flagged as genuinely needed by `extensions`'s own heavy testing-library usage stays untouched. `asyncUtilTimeout` now sits comfortably below every consuming project's `testTimeout`.

Verified directly with a throwaway never-resolving `waitFor` run under the repo-root config: before this commit it reported vitest's bare `Error: Test timed out in 5000ms.` with no DOM dump at ~5 s; after it, testing-library's own `AssertionError` with the full rendered DOM at the same ~5 s mark, well clear of the new 15 s `testTimeout`.

Same fix resolves #2 and the `asyncUtilTimeout` half of #3 in the sibling thread (`lib/platform-bible-react/vitest.config.ts:30`).

lyonsil:
One documentation follow-up, on a file this PR doesn't touch.

`.context/standards/Testing-Guide.md` drifts once this lands:

- Its **Root Configuration** snippet (lines 384-402) prints `vitest.config.ts` without the
  `testTimeout: 15000` this PR adds.
- The guide never mentions the relationship between vitest's `testTimeout` and testing-library's
  `asyncUtilTimeout`, which is now a live trap rather than trivia: `asyncUtilTimeout` lives in the
  shared `vitest.setup.ts`, so every project that loads that file has to keep its own `testTimeout`
  above it or testing-library's error — and its DOM dump — is unreachable. That is exactly the gap
  `4e51ab6f` just closed. The blockquote directly below the snippet already enumerates the workspace
  configs that share the setup file, so it is the natural place to state the constraint.

Per the "promote settled conventions" rule in `CLAUDE.md`, this belongs in the standard and not only
in the five config comments. Nothing catches the drift on its own — `npm run verify:testing-guide`
exercises the shared-store example only.

rolfheij-sil:
🤖 Claude: Both parts confirmed and fixed in `99ecd0a45cd`.

- The **Root Configuration** snippet now shows `testTimeout: 15000` with a comment pointing at the
  note below it.
- Added a paragraph next to the existing blockquote explaining the `testTimeout` /
  `asyncUtilTimeout` relationship: `testTimeout`'s clock starts at the beginning of the test body,
  while a given `waitFor`/`findBy*` call only starts its own `asyncUtilTimeout` countdown when
  that call is reached — so whichever deadline elapses first wins, and every project loading the
  shared `vitest.setup.ts` needs its `testTimeout` comfortably (not just nominally) above
  `asyncUtilTimeout`, or a failing `waitFor` reports vitest's bare timeout instead of
  testing-library's richer error and DOM dump.

I verified that mechanism independently rather than just restating the prior round's notes: fetched
`@testing-library/dom`'s actual v10.4.0 source and confirmed `asyncUtilTimeout`'s timer is armed
inside `waitFor()` at call time (not at `configure()`/setup-file load), then reproduced both
failure modes empirically with a throwaway test — equal budgets (5000/5000) produce vitest's bare
timeout with no DOM dump; the current config (15000/5000) lets testing-library's error surface as
intended.

One related gap I noticed while checking this: the invariant is now documented in five places (the
guide plus each config's own comment) but nothing enforces it mechanically — no lint rule or test
fails if a future project sets `testTimeout` at or below `asyncUtilTimeout`. Flagging it rather than
fixing it here since it's outside what you raised; happy to open a ticket if that's wanted.

lyonsil:
Yes on the ticket - please open one. The constraint now spans five config sites plus the guide, and a
future config that wires in the shared `vitest.setup.ts` without its own `testTimeout` regresses
silently, which is exactly the shape of the gap this round found.

I checked the fix independently rather than taking the commit message for it: under the root config
on the branch an unsatisfiable `findByText` now fails with `TestingLibraryElementError` and the full
DOM dump at ~5.0 s, and forcing `--testTimeout=5000` reproduces the pre-fix bare
`Error: Test timed out in 5000ms.` with no dump. `asyncUtilTimeout` sits below `testTimeout` in all
five projects, and the consumer set is right - `lib/eslint-plugin-paranext` and the `storybook`
project genuinely don't load the shared setup. Your reason for raising `testTimeout` rather than
lowering `asyncUtilTimeout` holds up too: 23 extension test files use `waitFor`/`findBy`.

Two optional things, neither worth holding the PR for:

1. The Changes section now disagrees with the addendum. The first bullet still says the raise is for
   the `unit` project "only: the `scripts` and `storybook` projects ... and the repo-root config,
   keep the 5 s default", and the second bullet's "the two halves are NOT symmetrical" framing
   described the state before `4e51ab6f` - they are symmetrical now. The addendum explains the new
   commit but doesn't retract either claim, so reading top-down gives the wrong picture. This lives
   only in GitHub history, not in the repo, so it's cosmetic - worth a tidy if you're editing the
   body anyway.

2. The `unit` project's own `testTimeout` is the one site without the invariant comment. The four you
   added all carry the "must stay comfortably above `asyncUtilTimeout`" warning; the original one in
   `lib/platform-bible-react/vitest.config.ts` still has only the contention rationale, so whoever
   tunes that particular value back down is the one person who doesn't see the constraint. (Same
   detail makes "documented in five places" four config sites plus the guide.)

lyonsil (approved):
:lgtm:

> @lyonsil reviewed 7 files and all commit messages, made 1 comment, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2759#-:-P16Vo0Z3HRq0u4dYcca:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

rolfheij-sil:
🤖 Claude: Thanks for the independent verification, Matt — good to have the mechanism confirmed from
outside the commit message.

Ticket filed: **[PT-4593](https://paratextstudio.atlassian.net/browse/PT-4593)**, parented under
"💡Tech Debt", for the enforcement gap — a test asserting every config's `testTimeout` stays
comfortably above `vitest.setup.ts`'s `asyncUtilTimeout`.

Both optional items are in too: the PR body's Changes section no longer contradicts the addendum,
and `lib/platform-bible-react/vitest.config.ts`'s `unit` project now carries the same invariant
comment as the other four configs.


[PT-4593]: https://paratextstudio.atlassian.net/browse/PT-4593?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

lyonsil (approved):
:lgtm:

> @lyonsil reviewed 1 file and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2759#-:-P1AasTz6THX2qqmWWFx:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2760, opened 2026-09-02
PT-4433: Fix resource-picker dead-ends

katherinejensen00:
# Code Review Summary

**Branch**: pt-4433-resource-picker-dead-ends

**Base**: origin/main

**Date**: 2026-09-02

**Files changed**: 48 — **30 hand-written**, plus 18 generated `platform-bible-react/dist` bundles you can skip

## Overview

**PT-4433 (NN5d): a resource-picker state should never say "nothing here" when what actually happened was a failure or a filter.**

Every dead-end in this ticket was the same defect wearing different clothes. The resource *panels* already solved it — `getResourcePanelReadiness` distinguishes loading / unreadable / catalog-failed / empty, and pairs each with the control that can change it. The picker surfaces those panels hand off to did not. This PR extends that principle to them.

The root cause turned out to be one line of contract: `getCachedResources` returned `DblResourceData[] | undefined`, and that single `undefined` meant **three** different things — no DBL credentials on this build, provider not registered yet, and the fetch broke. You cannot fix the dead ends without being able to tell those apart, so the command now returns a discriminated result and rejects on genuine failure.

⚠️ **That is a breaking change for extension authors — see the section below.**

### Where to start reviewing

Roughly in dependency order; the first three carry the design.

| File | What it does |
| --- | --- |
| `extensions/src/platform-get-resources/src/types/platform-get-resources.d.ts` | **The contract.** New `DblResourceCatalog` union: `available` \| `unavailable` + reason. Read this first. |
| `extensions/.../src/dbl-catalog.utils.ts` *(new)* | The contract's logic, extracted so it is testable — `main.ts` can't be loaded by vitest (`?inline` web-view imports). |
| `lib/platform-bible-react/src/hooks/use-retryable-promise.hook.ts` *(new)* | `usePromise` + a failure flag + a working retry. PAPI-free so renderer and web views share it. |
| `lib/.../resource-picker-dialog.component.tsx` | The picker's six body states, now chosen by one `getResourcePickerBodyState` discriminant. |
| `src/renderer/components/dialogs/share-layout.dialog.tsx` | Waits for the catalog before mounting — this gate prevents a data-loss path (see below). |
| `extensions/.../get-resources.web-view.tsx` + `.component.tsx` | Same treatment for Get Resources. |
| `extensions/.../scripture-text-grid.web-view.tsx` | Four silent `return;` guards now notify + log. |

### What a user sees now

| Situation | Before | After |
| --- | --- | --- |
| Catalog fetch failed | "No results found" | "Couldn't load…" + **Try again** |
| Provider not ready yet (transient) | "No results found" | "Couldn't load…" + **Try again** (a retry genuinely works) |
| Installation has no DBL credentials | "No results found" | "Resource downloads aren't available on this installation." (no retry — nothing to retry) |
| Filters match nothing | "No results found" | "No resources match the current filters." + **Clear filters** |
| Catalog genuinely empty | "No results found" | unchanged |
| Pick a resource before providers resolve (Text Collection) | silent no-op | error notification + log |

## Breaking change — needs announcing

`platformGetResources.getCachedResources` changed shape **and** runtime contract:

```ts
// before — never rejected
const resources = (await sendCommand('platformGetResources.getCachedResources')) ?? [];

// after — rejects on fetch failure
const catalog = await sendCommand('platformGetResources.getCachedResources');
const resources = catalog.status === 'available' ? catalog.resources : [];
```

All 6 in-repo callers are updated (verified by a whole-tree sweep: 16 hits, all accounted for). The exposure is **out-of-repo extensions** — `platform-get-resources.d.ts` is auto-included into their TypeScript programs via `typeRoots`.

The old contract *is* the bug being fixed, so it was not preserved behind a deprecation. Accepted deliberately by the author.

- [ ] **Post the breaking change to Discord `#platform-changes`** — this repo has no other channel that reaches extension authors (`CHANGELOG.md` is auto-generated from PR titles).

## API Changes

- `platform-bible-react`: added `useRetryablePromise` + type `RetryablePromiseState<T>`
- `platform-bible-react/experimental` (`ResourcePickerDialog`): added optional props `hasResourcesError`, `onRetryResources`, `areDownloadsUnavailable`; added exports `getResourcePickerBodyState`, `ResourcePickerBodyState`; 5 new entries in `RESOURCE_PICKER_DIALOG_STRING_KEYS` (all optional, no type break)
- `platform-get-resources`: added types `DblResourceCatalog`, `DblResourceCatalogUnavailableReason`; **changed** `getCachedResources` return type and rejection behaviour (see above); `GetResourcesProps` gained `onRetryResources`, `areDownloadsUnavailable`; added `RESOURCE_ACTION_PROVIDER_NOT_READY` sentinel
- `platform-scripture-editor`: renamed hook state type `DblResourceCatalog` → `DblResourceCatalogState` (avoids a collision with the new command type; no consumer imported it by name)
- `src/renderer/.../share-layout.component.tsx` (renderer-internal): `ShareLayoutDialogContentProps` gained required `hasResourcesError`, `onRetryResources`, `areDownloadsUnavailable`; exported `localizeString`
- `lib/papi-dts/papi.d.ts`: unchanged — no `src/shared` types changed, so no regeneration was owed

## Findings

Analysis ran four passes (API/correctness, style/patterns, coverage/compliance, UX design guide). Three findings were reported independently by more than one pass.

- [ ] ~~`getCachedResources` return type and rejection contract changed; the `.d.ts` is auto-included into external extension repos~~ _(Author accepted the break deliberately: the old single-`undefined` contract is precisely the defect being fixed, so a deprecation window would preserve the dead end. Mitigation is communication — the Breaking Change section above, plus a Discord `#platform-changes` post, tracked as an open action item there.)_

- [x] **Unlocalized English string with developer jargon reached the user** — `installOrRemoveResource` rejected with a hardcoded `'The DBL resources data provider is not available yet...'`, rendered verbatim in the alert _(fixed during review: web view now rejects with a sentinel; the component maps it to a new localized `%resources_providerNotReady%` in en + es. Test added, mutation-verified.)_ *(flagged by 3 of 4 passes)*
- [x] **Three hand-rolled zero states instead of the `Empty` primitive** — violates `adr-empty-is-zero-state-primitive`; also meant the picker's catalog error had no icon while the panels' identical failure did _(fixed during review: all three now compose `Empty`/`EmptyHeader`/`EmptyMedia`/`EmptyDescription`/`EmptyContent`, mirroring `RetryableErrorView`.)* *(flagged by 2 passes)*
- [x] **`hasSettled` was wired into only 1 of 3 hosts** — the other two painted "No results found" for one frame on the retry click itself _(fixed during review: both hosts now fold `!hasSettled` into their loading condition.)_
- [x] **`unavailable`'s `reason` was produced but never read** — `notReady` (transient) and `notConfigured` (permanent) both rendered as a bare empty list _(fixed during review: `notReady` now gets error+retry, `notConfigured` gets its own message and no retry. New localized keys in en + es; tests for both.)_
- [x] **Duplicated, self-contradicting comment block** in `share-layout.dialog.tsx` (differed only in "the memo below" vs "above") _(fixed during review: deleted the stale copy.)* *(flagged by 3 passes)*
- [x] **Data-loss guard had no test** — the `!hasResourcesSettled` gate prevents Confirm writing an empty `referencedProjectsAndResources` _(fixed during review: regression test added, mutation-verified that removing the guard fails it.)_
- [x] **Producer contract had no test** — every consumer was tested against mocked shapes, so nothing verified the backend emits them _(fixed during review: logic extracted to `dbl-catalog.utils.ts` with 8 tests; `main.ts` routes through it.)_
- [x] **Picker behaviours with rationale comments but no tests** — all-languages-selected, type-scoped count total, count suppression, focus move _(fixed during review: 4 tests added, 2 mutation-verified.)_
- [x] **`useDblResourceCatalog` duplication had no in-code marker** _(fixed during review: TODO naming `useRetryablePromise`, what would remain after migrating, and the ADR slug.)_
- [x] **Third silent early return in the grid, after the install await** — plus three sibling handlers _(fixed during review: all four now notify + log via `reportActionNotApplied`.)_

### Minor — Consider

- [x] `DblResourceCatalog` named two different types on opposite ends of one call chain _(fixed during review: hook's renamed to `DblResourceCatalogState`; standards doc updated.)* *(flagged by 2 passes)*
- [x] `hasSettled` never became true with an `undefined` factory — a foot-gun once 3 hosts gate render on it _(fixed during review.)_
- [x] `wrapFetch` mutated a ref inside a state updater React may call twice _(fixed during review.)_
- [x] Four sibling JSX conditionals re-spelled the same guards _(fixed during review: `getResourcePickerBodyState` discriminant + 6 unit tests.)_
- [x] Share Layout borrowed `%resourcePicker_*%` keys for its own body _(fixed during review: own `%shareLayoutDialog_loadError%` / `_retry%` keys; reuses exported `localizeString`.)_
- [x] Picker filters stayed interactive while loading/failed, unlike `GetResources` _(fixed during review.)_
- [x] Stale install-failure alert became unclearable once the error state replaced the table _(fixed during review: retry clears it; test added, mutation-verified.)_
- [x] Backward-facing clause in the `main.ts` "Awaited deliberately" comment _(fixed during review.)_
- [x] ADR entry preceded by two blank lines _(fixed during review.)_
- [ ] ~~Live-region role sits on a region that unmounts per state~~ _(Technically right, but `RetryableErrorView` — the repo's reference implementation — does the same, and `EmptyState` bakes in its own `role="status"`. Changing only the picker would make it inconsistent with the panels showing the same failure. Belongs at the design-system level.)_
- [ ] ~~Terminal-period inconsistency between "No results found" and the new strings~~ _(Follows a real rule: fragments take no period, sentences do. The pre-existing key is a fragment.)_
- [ ] ~~`reportActionNotApplied` guards in the grid are untested~~ _(Matches the extension's convention — no `*.web-view.tsx` has a test file; logic is tested via extracted utils. Noted below as a manual-verification item.)_
- [ ] ~~`home.web-view.tsx` / grid still use `usePromise` with the now-rejecting command~~ _(Verified safe: `usePromise` catches, clears loading, and logs; both use the result in a way where a rejection behaves as the old absent value did.)_
- [ ] ~~PR-level comment density~~ _(Partly addressed — backward-facing clauses removed and the ADR now carries the design argument that the code TODO points at by slug. Remaining comments state constraints a maintainer needs.)_

#### Shared Regions Modified

None. No `#region shared with` marker appears in any changed file.

#### Extension Config Changes

None. Only sources, tests, stories, a `.d.ts`, and `contributions/localizedStrings.json` — no `package.json`, `tsconfig.json`, `webpack.config.ts`, or `.eslintrc*`.

## Positive Observations

- The three-state catalog result is the right shape for the problem, and the ADR records the decision, the rejected alternatives, and an honest account of an earlier revision that would have created a new dead end.
- `useRetryablePromise` is correctly PAPI-free with an injected fetch, so the renderer and web views genuinely share it; its tests cover both supersession paths and the synchronous-generation race — exactly the cases a reimplementation gets wrong.
- Localization is additive across every surface (en + es), with no existing string mutated, so the immutable-strings rule holds.
- Storybook tracks the new states rather than lagging them, including a `play`-driven filtered-empty story that exercises the real filter path.
- Every previously-silent `return` in the Text Collection grid now emits a localized notification plus a log line, matching the "never fail silently" guideline.
- "Clear filters" is deliberately withheld when clearing provably cannot reveal anything, and the retry is withheld where no retry could work — avoiding inert controls.

## Interview Notes

**Author's stated purpose**: fix the resource-picker dead-ends so every failure and filter state says what happened and carries a control that can change it, including normalizing the `getCachedResources` contract so "this build can't download DBL resources" is distinguishable from "the fetch failed".

**Key decisions the author made:**

- **Accept the breaking change** rather than adding a parallel command, on the grounds that the old contract is the defect. Directed that it be documented in the PR body *and* posted to Discord `#platform-changes`.
- **Fix all 15 findings** from a prior `/code-review max` pass, including ones outside the ticket's original "minimal restructuring" scope — the contract normalization had been explicitly deferred to Sprint 90 and was pulled forward deliberately.
- **`notReady` vs `notConfigured` get opposite treatments** (option 1 of three offered) — the fuller fix, chosen over the cheaper "just fix `notReady`".

**Context a reviewer should know:**

- This branch is 2 commits plus in-review changes. The second commit exists because the first introduced two regressions caught by a prior review: a false-failure dead end on any build without DBL credentials, and a path that could erase a project's shared resource list. Both are fixed and now have tests.
- Per the author, **DBL credentials live in paratext-10-studio, not core** — so `notConfigured` is the *normal* state on a plain core build, which makes that state's copy more load-bearing than it looks.

**No areas of author uncertainty were recorded.** The author directed each decision point with a clear rationale and did not defer any to the AI; the implementation was AI-assisted throughout, which is disclosed on the PR.

## In-Review Quality Check

All fixes were made during the interview; checks were run after each and again at the end.

- **Format** (`npm run format` + `format:check`): clean.
- **Typecheck** (`npm run typecheck`): one pre-existing failure unrelated to this branch — `app.service-host.ts` cannot find the generated `release/app/buildInfo.json`. No errors in any changed file.
- **Lint**: clean in all three workspaces (repo root, `extensions/`, `lib/platform-bible-react/`).
- **Tests**: extensions 136/136, platform-bible-react 70/70, renderer 172/176 — the 4 failures are `Test timed out in 5000ms` in the known-flaky multi-window suites (analytics, notification, scroll-group, web-view service-shard); all pass when re-run alone and none are files this branch touches.
- **Self-caught during the sweep**: the full-suite run surfaced that a namespace change had broken one of this review's own earlier tests, and a scratch validator found three localization keys inserted out of alphabetical order by scripted edits. Both fixed; the validator also confirmed en↔es parity for every new key.

## Suggested Review Focus

- [ ] **The contract change is the whole PR** — `platform-get-resources.d.ts` and `dbl-catalog.utils.ts`. If you agree with the three-state shape, the rest follows.
- [ ] **Is accepting the break the right call?** The alternative (parallel `getResourceCatalog` + deprecate) was rejected because it keeps the defective contract reachable. Worth a second opinion given external extension authors are affected.
- [ ] **Copy check on the three new user-facing messages** — "Resource downloads aren't available on this installation.", "No resources match the current filters.", "Resources aren't ready yet. Please try again in a moment." Spanish translations are in the same commits.
- [ ] **Manual verification still owed** (no automated coverage, by design — the grid web view has no testable seam): Text Collection → View Options → pick a resource before the panel settles should produce an error notification rather than closing over an unchanged grid.
- [ ] **Deliberately deferred**: `useDblResourceCatalog` remains an un-migrated near-duplicate of `useRetryablePromise` (TODO in-code, rationale in the ADR). Confirm you're happy leaving it for a follow-up.

AI-assisted — [session](https://claude.ai/code/session_012orRVAreZRhucxTxyFi5dM)

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2760)


extensions/src/platform-scripture-editor/src/use-dbl-resource-catalog.hook.ts
```diff
@@ -58,12 +68,14 @@ export function useDblResourceCatalog(): DblResourceCatalog {
...
-        );
+        const catalog = await papi.commands.sendCommand('platformGetResources.getCachedResources');
         if (generation === fetchGenerationRef.current) setHasCatalogError(false);
 
-        return resources;
+        // A build that cannot download DBL resources has ARRIVED at its answer — there is no
+        // catalog and there never will be. Returning `undefined` here would leave `isCatalogReady`
+        // false forever, and the panels read that as "still loading": a spinner with no message and
+        // no retry, in exactly the offline/no-credentials case this hook exists to explain.
+        return catalog.status === 'available' ? catalog.resources : [];
```
mattgetgen: **[1] `notReady` is collapsed into a delivered, empty catalog — and that overwrites persisted state**

This ternary maps *both* unavailable reasons to `[]`, which reports the transient `notReady` state as a successfully-delivered empty catalog (`isCatalogReady: true`, `hasCatalogError: false`, error flag cleared on line 72).

**How it's reached:** `papi.dataProviders.get` returns `undefined` immediately when a provider isn't registered yet (`src/shared/services/data-provider.service.ts:1070`), and `DblResourcesDataProvider` is registered in the *background* from `c-sharp/Program.cs:99`. So a first run, a cleared profile, or any panel opened during .NET startup gets `{status: 'unavailable', reason: 'notReady'}`.

**Two consequences:**

1. In `resource-text-panel.web-view.tsx:296` every configured `dblResource` ref fails `dblResources.find(...)`, so `matchingCount === 0` and `getResourcePanelReadiness` (`resource-panel-readiness.utils.ts:83`) returns `'empty'` — the "pick a resource" prompt for a project that *has* resources configured. `PanelReadinessView` renders a retry only for `'catalogError'`, so there's no way out. `model-text-panel` gets the "unknown resource" prompt, also retry-less.
2. Worse: `canPublishResourcePanelProjectIds(effectiveResourcesState, isCatalogReady)` (`resource-panel-readiness.utils.ts:100`) now returns `true`, so `usePublishNavigableProjectIds(useWebViewState, [], true)` (`resource-text-panel.web-view.tsx:381-385`) writes an **empty** navigable-project-id list into persisted web-view state. That's the exact hazard that function's own TSDoc forbids: *"publishing `undefined` as 'nothing displayed' would overwrite a correct persisted declaration"*.

This is the only one of the five catalog consumers that discards `reason` — the other hosts all give `notReady` an error + retry. The new test covers `notConfigured` only, so it can't catch this.

src/renderer/components/dialogs/share-layout.dialog.tsx
```diff
@@ -167,19 +194,49 @@ function ShareLayoutDialogWrapper({
...
   // dialog. Reject here instead. This check must run after all hooks above (Rules of Hooks
   // forbids an early return between hook calls), so it sits just before the render branch.
-  if (isCanWriteLoading) {
+  // Also gated on the catalog: `ShareLayoutDialogContent` snapshots its initial resource lists in
+  // `useState` at mount, and `splitResourcesByTab` cannot classify a saved dblResource reference
+  // without the catalog — every one of them lands in `otherResources` instead. Mounting before the
+  // catalog settles therefore seeds the content with empty lists while the memo above later
+  // recomputes `otherResources` to empty too, so a Confirm from that mount writes an EMPTY
+  // referenced-resources list and erases the project's shared selection.
+  if (isCanWriteLoading || !hasResourcesSettled) {
```
mattgetgen: **[2] The settle gate guards the catalog, but two other inputs to the same snapshot are ungated**

This gate waits on `canWrite` and the catalog, but the mount-time snapshot it protects also depends on `projectResourcesSetting` and `personalResources`, neither of which is gated — so Confirm can still write `items: []` and erase the project's shared resource list, which is the erasure the gate exists to prevent.

**Why the empty value is indistinguishable:** line 91 destructures only two elements from `useProjectSetting`, discarding its 4th `isLoading` (`use-project-setting.hook.ts:129`). `createUseDataHook` returns the `defaultValue` while loading (`create-use-data-hook.util.ts:268`), so `projectResourcesSetting === EMPTY_RESOURCE_LIST` is byte-identical to a genuinely empty shared list. `personalResources` is `undefined` in flight.

**Why the window is wide, not narrow:** mounting needs only `canWrite` (one method round-trip) plus the catalog (no provider dependency), while the setting needs a second PDP + subscribe + first delivery. This repo's own `use-effective-resource-reference-list.hook.ts:152-157` documents that asymmetry as *"the normal window on essentially every mount, not a narrow race"*.

**The sequence:** gate opens with `seededItems === []` → `share-layout.component.tsx:175-176` latches `[]` into `useState` permanently → the real setting arrives and `splitResourcesByTab` classifies the saved refs into scripture/commentary, leaving `otherResources === []` → `handleConfirm` (line 177) writes `items: []`.

A variant is just as bad: if `personalResources` has arrived but the project setting hasn't, Save writes the admin's *personal* list over the team's shared list.

The new regression test can't catch either — `share-layout.dialog.test.tsx:73-92` hard-codes `useProjectSetting`'s 4th element to `false`.

src/renderer/components/dialogs/share-layout.dialog.tsx
```diff
@@ -56,11 +68,26 @@ function ShareLayoutDialogWrapper({
...
-    undefined,
   );
 
+  const allResources = catalog?.status === 'available' ? catalog.resources : undefined;
+
+  // `notReady` is transient, so it earns the retryable error state; `notConfigured` is permanent for
+  // this installation, so the dialog still renders and the embedded pickers explain the empty list
+  // rather than showing an unexplained "no results".
+  const isCatalogNotReady = catalog?.status === 'unavailable' && catalog.reason === 'notReady';
+  const areDownloadsUnavailable =
```
mattgetgen, line 77: **[3] `notConfigured` mounts the dialog body with no catalog**

`notConfigured` is excluded from both gates, so the body mounts with `allResources === undefined` — the no-catalog condition the gate two lines below exists to forbid — and renders a confidently false picture of what's being shared.

On an installation with no DBL credentials (`IsGetDblResourcesAvailable()` false — the normal core case per the PR body), `hasResourcesSettled` is true and `hasUnrecoverableCatalogError` is false, so line 255 passes `allResources ?? []` → `[]`. Per `share-layout.utils.ts:36-41`, every saved `dblResource` reference then falls into `otherResources`, which the dialog never displays.

So an admin opening Share Layout for a project whose `referencedProjectsAndResources` holds DBL resources (e.g. shared by a teammate who *does* have credentials) sees **empty** "Project Bible texts" and "Project commentaries" rows — under the dialog's own heading *"Review what you're about to share with your team before confirming"* — can't see or remove those entries, and Save silently rewrites the list, moving every unseen DBL reference to the end.

When the project setting is empty, `seedResourceList` falls back to the admin's personal list, so Save shares the admin's entire personal DBL selection to the team without ever showing it.

The embedded pickers explain themselves via `areDownloadsUnavailable`; the dialog body has no equivalent.

lib/platform-bible-react/src/hooks/use-retryable-promise.hook.ts
```diff
@@ -0,0 +1,145 @@
...
+      }
+    };
+  }, []);
+
+  // A changed `promiseFactoryCallback` supersedes any in-flight fetch exactly as `refetch` does, so
+  // it bumps the same generation rather than relying on `usePromise`'s currency flag, which does not
+  // cover the flags owned here.
+  useEffect(() => {
+    setHasError(false);
+    if (!promiseFactoryCallback) {
```
mattgetgen: **[4] The factory→`undefined` transition is a supersession that doesn't bump the generation**

*Not reachable through this PR's own hosts — hardening, not a live bug.*

This branch returns before `wrapFetch`, which is the only place `fetchGenerationRef` is incremented. So an abandoned in-flight fetch can still set `hasError` permanently, with a retry that can't clear it.

**Sequence** (caller passes `useCallback(cond ? fn : undefined, [cond])` — a conditional or disabled fetch):
1. Fetch A starts at generation 1.
2. `cond` flips: the effect sets `trackedFetch = undefined`, `hasError = false`, `hasSettled = true`, and leaves `fetchGenerationRef.current === 1`.
3. Fetch A rejects. `generation === fetchGenerationRef.current` still holds, so lines 95-98 run `setHasError(true)`.
4. `refetch` returns early at line 131 (`if (!factory) return`), so nothing can clear it.

A host wired the way this PR's three hosts are (`hasResourcesError={hasError || ...}` + `onRetryResources={refetch}`) then renders the CloudOff error panel with a Try again button that provably does nothing.

This contradicts the hook's own comment at lines 66-72 — *"Bumped SYNCHRONOUSLY at every point a fetch is superseded — both ways that can happen"*. There are three ways; two are covered, and neither new supersession test exercises this one.

Worth fixing anyway because the hook is now a public export and the PR's TODO directs the next maintainer to migrate `useDblResourceCatalog` onto it — whose factory *is* state-derived.

src/renderer/components/dialogs/share-layout.dialog.tsx
```diff
@@ -56,11 +68,26 @@ function ShareLayoutDialogWrapper({
...
+    catalog?.status === 'unavailable' && catalog.reason === 'notConfigured';
+  const hasUnrecoverableCatalogError = hasResourcesError || isCatalogNotReady;
```
mattgetgen: **[5] Folding transient `notReady` in here replaces the *entire* dialog**

Treating `notReady` as unrecoverable makes the line-220 early return swap out the whole dialog — header, description, Save and Cancel — for an error card. The pre-PR code rendered the dialog normally against an empty catalog.

Open Share Layout during the window before the background-registered DBL provider is up (or on a first run with no cached catalog) and the admin gets only *"Couldn't load the list of available resources."* + Try again, with no way to reach the tab or model-text settings, which have nothing to do with DBL.

It also lasts a while: `main.ts:54-70` holds `fetchMutex` across all 10 attempts (~9s of `wait(1000)` plus round trips) and `shouldStopBackgroundFetch` returns false for `notReady`, so an on-demand call queues behind the whole loop. The dialog is a bare spinner and then an error page for ~10s.

Separately: the return at line 220 sits *above* the `canWrite !== true` gate at line 240, whose comment says it "sits just before the render branch". A non-admin who triggers the command now gets a live Try again button that re-drives a catalog fetch before the cancel effect fires.

lib/platform-bible-react/src/hooks/use-retryable-promise.hook.ts
```diff
@@ -0,0 +1,145 @@
...
+      // strand any host that gates its render on this — and `refetch` could not rescue it, because
+      // there is no factory to re-run.
+      setHasSettled(true);
+      return;
+    }
+    setHasSettled(false);
+    // Wrapped before the setter rather than inside it: `wrapFetch` bumps the generation ref, and
+    // React may invoke a state updater more than once.
+    const wrapped = wrapFetch(promiseFactoryCallback);
+    setTrackedFetch(() => wrapped);
```
mattgetgen: **[6] An unstable factory produces an unbounded render loop rather than degrading**

*Not reachable through this PR's own hosts — hardening, not a live bug.*

`setTrackedFetch(() => wrapped)` stores a brand-new closure on every effect run, so a non-`useCallback` factory loops:

1. Render N has a new factory identity → the effect at line 107 fires.
2. `wrapFetch` returns a fresh closure → `setTrackedFetch` stores a value that's never `Object.is`-equal to the previous one, so React can't bail out.
3. Re-render → new inline factory → effect fires again.

No promise needs to resolve. React aborts the tree with "Maximum update depth exceeded" and the dialog or web view crashes on first mount.

The contrast matters: `usePromise` under the same misuse only calls `setIsLoading(true)`, which bails out when already `true`, so it merely re-fetches on unrelated re-renders. Here the same mistake is a hard crash — and the hook was added to `src/index.ts:320`, the **stable** surface with no deprecation cycle, so extension authors who don't read the TSDoc WARNING at line 55 get no softer failure.

Holding `wrapped` in a ref and keying the state on a monotonic counter would fix it.

extensions/src/platform-get-resources/src/get-resources.web-view.tsx
```diff
@@ -132,7 +157,9 @@ globalThis.webViewComponent = function GetResourcesDialog({ useWebViewState }: W
       localizedStringsWithLoadingState={localizedStringsWithLoadingState}
       resources={resolvedResources}
       isLoadingResources={isLoadingResources}
-      isResourcesError={false}
+      isResourcesError={isResourcesError || isCatalogNotReady}
```
mattgetgen: **[7] A failed post-install refetch discards a catalog `usePromise` still holds**

This makes a failed refetch replace the whole populated resource table with the error card — a regression from the previous hard-coded `isResourcesError={false}`.

`usePromise` defaults `preserveValue: true` and its catch deliberately leaves `value` alone (`use-promise.hook.ts:75-86`), so after a successful load `catalog` still holds `{status: 'available', resources: [...]}`.

**Sequence:** user installs a resource → `installOrRemoveResource`'s `.then` calls `refetchResources()` (line 124) → that refetch rejects (newly possible, because `main.ts`'s on-demand path now rethrows where it previously resolved `undefined`) → `hasError` flips true → `GetResources` renders the `Empty role="alert"` branch *instead of* the table (`get-resources.component.tsx:498`).

The freshly-installed resource and every other row vanish behind "Unable to search for resources", along with any in-progress `idsBeingHandled` spinners. Before this PR the same transient blip left the stale-but-correct table on screen.

The hook's own TSDoc (lines 10-14) flags this as a decision the caller must make; only the initial-load case seems to have been reasoned about here. Note that `useRetryablePromise` gives the caller no way to opt out — it hardcodes `defaultValue` and drops `UsePromiseOptions`.

mattgetgen (commented):
## Review: 15 findings

*AI-assisted review (Claude Code, max effort) — 10 finder angles plus verification passes over `94783b5..f9a4d6a` (47 files, `dist/` excluded). Findings are claims to check, not verdicts; I've flagged which ones I could not reach through the current call sites.*

Nice PR overall — the `notReady`/`notConfigured`/`available` catalog contract and the `getResourcePickerBodyState` discriminant are the right shape. Most of what follows is about applying them *consistently*.

### The through-line

The PR's thesis is "no dead ends," and the highest-severity findings are all places where that thesis is applied unevenly across the five DBL-catalog consumers. The transient `notReady` state is handled **three different ways in three files**, and one of those ways is unsafe.

**Data-loss / false-empty-state (worth fixing before merge):**

- **[1] `use-dbl-resource-catalog.hook.ts:78`** — `notReady` is mapped to a *successful empty catalog*. That's not only a false empty state with no retry; it flips `canPublishResourcePanelProjectIds` to `true`, so the panel writes an **empty** navigable-project-id list over a correct persisted declaration — the exact hazard that function's own TSDoc forbids. This is the one hook of five that discards `reason`, and the new test covers `notConfigured` only.
- **[2] `share-layout.dialog.tsx:207`** — the new settle gate guards the catalog but not the *other two* inputs to the same mount-time snapshot (`projectResourcesSetting`, `personalResources`). Confirm can still write `items: []` over a project's shared resource list. The regression test can't catch it because it hard-codes `useProjectSetting`'s 4th element (`isLoading`) to `false`.
- **[3] `share-layout.dialog.tsx:87`** — `notConfigured` is excluded from both gates, so on an installation with no DBL credentials the body mounts with no catalog and shows an admin an empty "what you're about to share" list for a project that *does* have DBL resources — then Save rewrites it.
- **[11] `scripture-text-grid.web-view.tsx:244`** — updated for the new return *shape* but not the new *rejection*. It's the one file where the `@throws` contract this PR introduced gets silently discarded, and a failed fetch still renders a truthful-looking empty grid.

**Correctness / UX:** [5] `notReady` replaces the entire Share Layout dialog (and sits above the admin gate), [7] a failed post-install refetch discards a catalog `usePromise` still holds, [8] "Showing 0 of 0" renders above the empty/unavailable message with live filters, [9] the language-filter predicate is measured against the un-type-scoped set, [10] a cache-*write* failure is reported as a catalog failure, [12] the sentinel uses string equality across a boundary PAPI rewrites, [13] Share Layout forwards raw `isLoading` contrary to the hook's own documented rule.

**`useRetryablePromise` [4] [6] [14]** — I could **not** reach these three through this PR's own hosts, so treat them as hardening rather than bugs. They matter because the hook went onto the *stable* `src/index.ts` surface (no deprecation cycle), and the PR's own TODO points the next maintainer at a consumer whose factory is state-derived.

**[15]** is a reuse note: `RetryableErrorView` already exists and does this, and `getResourcePickerBodyState` isn't actually exported from `experimental.ts` — so the API-Changes claim in the PR body doesn't hold as written.

### On the test suite

Worth a second look independent of the findings: several new tests are non-falsifiable or drive prop combinations the wrapper can't produce (details in [8], [9], [13]). The green suite is weaker evidence than it looks. There are also ~12 lower-severity items I left off to keep this readable — duplicate localization keys, an unrestored `scrollIntoView` mock leaking across tests, a `TODO` with no ticket, and an ADR that names a "Send Layout dialog" that doesn't exist in the codebase. Happy to post those separately if useful.

lib/platform-bible-react/src/components/advanced/resource-picker-dialog/resource-picker-dialog.component.tsx
```diff
@@ -260,26 +405,73 @@ export default function ResourcePickerDialog({
...
           variant="outline"
+          isDisabled={areFiltersInert}
         />
       </div>
-      {isFiltered && (
+      {/* Suppressed while loading or failed: a count of a catalog that never arrived reads as a
+          confident "0 of 0" directly above the message explaining that nothing could be loaded. The
+          total is the type-scoped set, not the whole catalog — reporting entries this picker filters
+          out anyway implies candidates the user could reach by clearing something. */}
+      {isFiltered && bodyState !== 'loading' && bodyState !== 'error' && (
```
mattgetgen: **[8] "Showing 0 of 0" still renders over the `downloadsUnavailable` and `empty` messages**

The count row is suppressed only for `bodyState === 'loading'` and `'error'`, and `areFiltersInert` (line 377) covers the same two. Neither covers `'downloadsUnavailable'` or `'empty'`, so those states render a confident "Showing 0 of 0 resources" directly above their own message, with fully live filters.

**Repro:** install with no DBL credentials → `areDownloadsUnavailable` true, `allResources` `[]`, and `areFiltersInert` false (not loading, no error), so the search box and language combo stay enabled. Type "NIV": `isFiltered` becomes true, `canClearFiltersHelp` is false because `typeScopedResources.length === 0`, so `bodyState` stays `'downloadsUnavailable'` and the header prints "Showing 0 of 0 resources" immediately above "Resource downloads aren't available on this installation."

That's precisely the artifact the adjacent comment at lines 411-414 says it prevents: *"a confident '0 of 0' directly above the message explaining that nothing could be loaded"*. Same for a genuinely empty catalog: "Showing 0 of 0 resources" above "No results found".

The new test `keeps the plain no-results text when the catalog is empty with no filter applied` never types a search term, so `isFiltered` is false and the count is absent regardless — it can't catch this.

lib/platform-bible-react/src/components/advanced/resource-picker-dialog/resource-picker-dialog.component.tsx
```diff
@@ -239,7 +358,31 @@ export default function ResourcePickerDialog({
     );
   }, [selectedLanguages, languageOptions, anyLanguageText, localizedStrings]);
 
-  const isFiltered = searchText.length > 0 || selectedLanguages.length > 0;
+  // Selecting every language narrows nothing, which is why the trigger keeps reading "Any language"
+  // in that state. Treating it as filtered anyway would report a "showing 9 of 9" count and offer to
+  // clear a filter that is not excluding anything.
+  const isLanguageFiltered =
```
mattgetgen: **[9] `isLanguageFiltered` is measured against the un-type-scoped set**

`isLanguageFiltered` compares `selectedLanguages.length` against `languageOptions.length`, which is memoized over the un-type-scoped `allResources` (line 308) — while the language predicate is applied to `typeScopedResources`. So the "selecting everything is not a filter" test is measured against the wrong set, and it breaks in both directions.

**(a) False positive.** Share Layout's Commentary tab passes the whole catalog with `resourceType='CommentaryResource'`. Commentaries exist only in English, but Scripture/Xml entries add Spanish, Greek and Hebrew, so `languageOptions.length === 4`. The user selects English — the only language any commentary has: `1 > 0 && 1 < 4` → `isLanguageFiltered` true → the count row asserts "Showing 4 of 4 resources" for a narrowing that excluded nothing, and `canClearFiltersHelp` offers to clear a filter that isn't filtering.

**(b) False negative, which loses the escape hatch.** If `allResources` changes under a mounted picker (a retry, or a host that persists the language filter as `get-resources.web-view.tsx` does via `useWebViewState`) such that `languageOptions` no longer contains an already-selected language, `selectedLanguages.length < languageOptions.length` goes false while the predicate still excludes every row. `isFiltered` false → `canClearFiltersHelp` false → `getResourcePickerBodyState` returns `'empty'`, so the user gets bare "No results found" with **no "Clear filters" button** while an invisible language filter hides everything.

The new test passes only because every language in `SAMPLE_RESOURCES` is present in the unscoped set. The honest predicate is whether the selection actually excludes a member of `typeScopedResources`.

extensions/src/platform-get-resources/src/main.ts
```diff
@@ -29,23 +31,21 @@ let cachedResources: DblResourceData[] | undefined;
...
-        JSON.stringify(cachedResources),
-      );
-  }
-  return resources;
+  // The contract itself lives in `dbl-catalog.utils` so it can be tested: this module imports web
+  // views through webpack's `?inline` loader and so cannot be loaded by the test runner.
+  const catalog = await resolveDblCatalog(provider);
+  if (catalog.status !== 'available') return catalog;
+
+  cachedResources = catalog.resources;
```
mattgetgen, line 43: **[10] A cache-*write* failure is reported to every host as a total catalog failure**

`cachedResources` is assigned here, *before* the awaited `papi.storage.writeUserData` — and `getCachedResources` now rethrows instead of returning `undefined`. So a failure of the persistence write, which has nothing to do with whether the catalog was fetched, surfaces as a complete catalog failure.

On a disk-full, a locked user-data file, or any `writeUserData` rejection after a *successful* DBL fetch: `fetchAndCacheResources` rejects, `getCachedResources` logs and rethrows at line 133, and all three retryable hosts paint "Couldn't load the list of available resources." with a Try again button — while the complete catalog is already sitting in the module-level `cachedResources`.

In `share-layout.dialog.tsx` this is worse than a spurious message: the dialog refuses to open at all until the user clicks retry.

The tell is that it self-heals on the first retry, which takes the `cachedResources !== undefined` fast path — showing the error was never about the catalog.

The old `if (resources) { … }` guard had the same structural flaw, so this rewrite was the natural point to re-establish "a fetched catalog is returned": either return `catalog` regardless of the write outcome, or catch the write. Either keeps the data.

extensions/src/platform-scripture-editor/src/scripture-text-grid.web-view.tsx
```diff
@@ -223,7 +241,7 @@ globalThis.webViewComponent = function ScriptureTextGridWebView({
   // also supplies the DBL `fullName` shown as the long name in the View Options list.
   // Re-fetched on `refreshCounter` bumps so a newly-installed resource's `installed` flag is
   // current when `toGridResources` resolves it.
-  const [cachedResources, isLoadingCachedResources] = usePromise(
+  const [catalog, isLoadingCachedResources] = usePromise(
```
mattgetgen: **[11] The grid was updated for the new return shape, but not the new rejection**

This stays on raw `usePromise`, which folds a rejected fetch back into an absent value that the new memo maps to `EMPTY_DBL_RESOURCES` — indistinguishable from the `unavailable` case, with no message, no retry, and no `logger` line.

On a transient catalog failure (extension-host restart, provider throw, or `resolveDblCatalog`'s new `throw new Error('The DBL resource catalog fetch produced no catalog')`), `usePromise` catches it into a bare `console.error` and clears loading. So `catalog` is `undefined` → `cachedResources` is `[]` → `toGridResources` gives every `DblResourceReference` `projectId: undefined` (each cell renders the unavailable placeholder) and `resolveDblLongName` loses every long name in View Options.

If all sources are DBL refs, `resources.length === 0 && sources !== undefined && !isLoadingCachedResources` (~line 603) renders "use View Options to add texts" — a truthful-looking empty grid for a fetch that *failed*. That's the exact dead end this PR set out to remove, in a file the PR otherwise hardened with `reportActionNotApplied`.

This is the one place the contract half of the PR (`@throws` on `getCachedResources`, `platform-get-resources.d.ts:104-107`) gets silently discarded, against the new ADR's own rule: *"render the error state from the rejection, never from an absent value."*

The `refreshCounter` refetch after an install has the same hole.

extensions/src/platform-get-resources/src/get-resources.component.tsx
```diff
@@ -272,10 +306,26 @@ export function GetResources({
     try {
       await onInstallOrRemoveResource(dblEntryUid, action);
     } catch (e) {
-      setActionError(getErrorMessage(e));
+      // Callers signal "the backing provider has not resolved yet" with a sentinel rather than a
+      // message, because the text the user reads has to be localized and this component is the half
+      // that holds the localized strings. Any other rejection carries a message worth showing.
+      setActionError(
+        getErrorMessage(e) === RESOURCE_ACTION_PROVIDER_NOT_READY
```
mattgetgen: **[12] Exact string equality against a message PAPI rewrites at process boundaries**

This is exact equality against a message that PAPI provably rewrites when a rejection crosses a process boundary — and the sentinel is a public export, which invites exactly the callers for whom the check silently fails.

`network.service.ts:348` builds cross-process rejection messages as `${getJsonRpcRequestErrorMessagePrefix(code)}: ${original}` (`rpc.model.ts:500`). The current caller works only because it rejects inside the web view's own React tree (`get-resources.web-view.tsx:111`) — which is also why the new test passes.

But line 190 exports the sentinel with TSDoc reading *"Rejection reason a caller of `onInstallOrRemoveResource` uses"*. The next caller raising it from the extension host, a data provider, or a registered command hands this component `JSON-RPC Request error (-32000): platformGetResources.providerNotReady`. The equality fails, the ternary falls through to `setActionError(getErrorMessage(e))`, and the destructive Alert shows the user a raw untranslated machine token plus JSON-RPC noise — the precise outcome the sentinel design exists to prevent.

The same boundary already leaks into the non-sentinel branch: real `installDblResource`/`uninstallDblResource` failures *are* cross-process, so their user-facing text carries the prefix today.

The repo already has the deeper mechanism — `newPlatformError(message, code)` with `PlatformErrorCode.FAILED_PRECONDITION`/`UNAVAILABLE` (`platform-error.ts:80-141`), discriminated via `isPlatformError(e) && e.code === …` at 15+ sites.

src/renderer/components/dialogs/share-layout.dialog.tsx
```diff
@@ -197,6 +254,9 @@ function ShareLayoutDialogWrapper({
       initialCommentaryResources={commentaryResources}
       allResources={allResources ?? []}
       isResourcesLoading={isResourcesLoading}
```
mattgetgen: **[13] Raw `isLoading` forwarded here, contrary to the hook's documented rule — and the error props are dead**

This host forwards the **raw** `isLoading` to the content and both embedded pickers, where the other two hosts don't, while `hasResourcesError`/`onRetryResources` are provably unreachable past the line-220 return.

`use-retryable-promise.hook.ts:32-40` says *"Read this rather than inferring 'finished' from `!isLoading`"*. `resource-picker.dialog.tsx:47` and `get-resources.web-view.tsx:44` both compose `isLoading || !hasSettled`; this host splits `hasSettled` off for its mount gate and then drops it on the way down.

Any refetch reached from inside the mounted dialog then paints a settled picker body over a fetch that hasn't started — and simultaneously the `!hasResourcesSettled` gate unmounts `ShareLayoutDialogContent`, discarding the tab/model-text/resource `useState` snapshot the admin has been editing.

That's masked today only because `hasResourcesError={hasUnrecoverableCatalogError}` is always `false` here (line 220 consumed every truthy case), so the pickers' retry can never render. Which means the two new tests in `share-layout.component.test.tsx` (`'reports a failed catalog fetch in the manage popover…'`, `'…in the model-text popover too'`) drive `hasResourcesError: true` through a path no user can reach, and can't catch a real regression.

Worth flagging: fixing the dead-prop half *without* fixing the raw-`isLoading` half turns this into a live edit-loss bug.

lib/platform-bible-react/src/hooks/use-retryable-promise.hook.ts
```diff
@@ -0,0 +1,145 @@
...
+  }, [promiseFactoryCallback, wrapFetch]);
+
+  const [data, isLoading] = usePromise<T | undefined>(trackedFetch, undefined);
+
+  const promiseFactoryCallbackRef = useRef(promiseFactoryCallback);
+  promiseFactoryCallbackRef.current = promiseFactoryCallback;
```
mattgetgen: **[14] Ref written during the render phase**

*Not reachable through this PR's own hosts — hardening, not a live bug.*

This writes a ref during render, so under React 19 concurrent rendering an abandoned render can leave the ref holding a factory that was never committed — and `refetch` then runs that factory *while bumping the generation that invalidates the committed one*.

React 19 may start a render, discard it, and re-render with different props (transitions via `startTransition`/`useTransition`, Suspense retries, offscreen prerender). Because the write is unconditional and outside any effect, the ref keeps whichever render ran **last**, committed or not.

If `refetch()` then fires from a user event, it reads the abandoned `factoryB`, calls `wrapFetch(factoryB)` — incrementing `fetchGenerationRef` as a side effect — and stores `wrappedB`. The committed tree's in-flight `wrappedA` now fails its generation check and its result is silently dropped, while the UI runs a fetch belonging to props that were never committed.

StrictMode's double render writes the same value twice, so this won't surface in dev.

Nothing needs the value during render — moving the write into an effect keeps `refetch` reading the committed factory. Noting for fairness: `usePromise` has the same pattern for `defaultValueRef`, so this is copied precedent. The consequence here (running the wrong factory *and* invalidating the right one) is materially worse than a stale default.

lib/platform-bible-react/src/components/advanced/resource-picker-dialog/resource-picker-dialog.component.tsx
```diff
@@ -260,26 +405,73 @@ export default function ResourcePickerDialog({
...
       <div className="tw:min-h-0 tw:flex-1 tw:overflow-y-auto tw:px-4 tw:pb-4">
-        {isResourcesLoading && (
+        {bodyState === 'loading' && (
           <p className="tw:py-8 tw:text-center">
             <Spinner />
           </p>
         )}
-        {hasNoResults && !isResourcesLoading && (
-          <p className="tw:py-8 tw:text-center tw:text-muted-foreground">{noResultsText}</p>
+        {bodyState === 'error' && (
```
mattgetgen, line 477: **[15] `RetryableErrorView` already does this — and the new discriminant isn't actually exported**

The icon + message (+ retry) zero state is hand-rolled six times across three files, rather than promoting the existing primitive.

`RetryableErrorView` (`extensions/src/platform-scripture-editor/src/panel-state-views.component.tsx:62`) already takes `{message, retryLabel, onRetry, icon}`, and `PanelReadinessView` already calls it with `icon={<CloudOff />}` for this identical DBL-catalog failure. This PR adds five more copies of that composition — here, line 461, `get-resources.component.tsx:498` and `:514`, and `share-layout.dialog.tsx:222` — each independently choosing `role="alert"` vs `role="status"`, the icon, and `Button variant`/`size`. The new ones use `variant="outline" size="sm"`; `RetryableErrorView` uses the default, so the retry affordance already renders differently depending on which surface the user hits it from. That's the drift `adr-empty-is-zero-state-primitive` names as the thing to avoid.

`lib/platform-bible-react` can't import from an extension, so the fix is to move it next to `empty-state.component.tsx`.

Two compounding notes:

- `getResourcePickerBodyState`/`ResourcePickerBodyState` are `export`ed from the component module but **absent from `experimental.ts`** (lines 22-27 export only the dialog, its props, its strings type and the keys array). The PR body's API-Changes claim that they were added to `platform-bible-react/experimental` doesn't hold as written. Their only consumer is `resource-picker-dialog.utils.test.ts`, which now drags the whole React/shadcn/lucide tree into what was a pure-utility test file.
- `GetResources` kept its `eslint-disable no-nested-ternary` chain rather than adopting the discriminant, so the PR ships two answers to the same question.

mattgetgen:
Sorry, I realize the 15 inline comments looks annoying. I'll not do that when there's a larger number of them next time.

katherinejensen00:
Answering all 15 in one comment for time's sake — I'll try to post individual responses in the threads tomorrow. All of it is in `6aae442`.

Two of these were design calls rather than straight fixes, and I took the fuller option on both.

### The through-line you named

You were right that `notReady` was handled three ways in three files. It is now one way everywhere: **`notReady` is a catalog still on its way (error + retry); `notConfigured` is a delivered empty catalog (own message, no retry); a rejection is a third thing again.** Every consumer branches the same way.

### Data-loss / false-empty-state

**[1] `use-dbl-resource-catalog.hook.ts` — `notReady` collapsed into a delivered empty catalog.** Confirmed, including the second-order consequence. It now sets `hasCatalogError` for `notReady` only, which drops `isCatalogReady` and so takes `canPublishResourcePanelProjectIds` back to `false` — the publish hazard closes structurally rather than by a separate guard. Two tests added, mutation-verified against reverting the branch.

**[2] Share Layout — the other two inputs to the snapshot were ungated.** Confirmed, and the test really was hard-coding the 4th element. The gate now waits on delivery of `canWrite`, all three project settings (via their `isLoading`), both personal lists, and the catalog. The test mock's loading flag is now **per setting key**, because with one shared flag a gate that checked only one of the three still looked correct — I found that by mutating the gate and watching the test pass. It fails now.

**[3] + [5] Share Layout with no usable catalog.** These two pulled in opposite directions — [3] says don't mount without a catalog, [5] says don't withhold the dialog for a transient failure — and on a plain core build `notConfigured` is permanent, so "don't mount without a catalog" would mean the dialog never opens. Resolved by separating *delivered* from *successful*: the dialog blocks until every input has settled once, then always renders, and **counts out loud** the saved DBL references it cannot classify ("N shared resources can't be shown because…"), with a retry only when the reason is recoverable. They are still round-tripped unchanged on Confirm; the notice is what stops that being silent. Also moved the admin gate above every other branch, so a non-admin can no longer reach a live Try again. New ADR: `adr-share-layout-renders-without-a-catalog`.

**[11] The grid discarded the new rejection.** Confirmed — it was the one place the `@throws` half of the contract was dropped. Now on `useRetryablePromise`, with a `logger.warn` and a real error+retry state instead of the "use View Options to add texts" prompt. The decision is extracted to `grid-body-state.utils.ts` so it has a test seam, since the web view itself has none (7 tests).

### Correctness / UX

**[7] Failed post-install refetch discarded a held catalog.** Confirmed, and it was a regression from the previous hard-coded `false`. The error state now only replaces the list when there is nothing else to show. Extracted as `shouldReportCatalogFailure` in `dbl-catalog.utils.ts` with tests, for the same seam reason.

**[8] "Showing 0 of 0" over the empty/unavailable messages.** Confirmed. The count is now shown only for `list` and `filteredEmpty` — the two states where it describes something actionable — and filters go inert whenever the type-scoped set is empty, so the live-filters half is closed too. Two tests, both driving `isFiltered` true first so they fail on the old condition (verified).

**[9] `isLanguageFiltered` measured against the un-type-scoped set.** Confirmed in both directions, and the false negative is the worse one as you said. The predicate is now "does this selection exclude a member of `typeScopedResources`?" I also routed the combo-box trigger label through the same predicate, so the trigger and the Clear-filters affordance cannot disagree. Two tests; the false-negative one reproduces your catalog-changes-under-a-mounted-picker sequence.

**[10] Cache-*write* failure reported as a total catalog failure.** Confirmed, including the self-healing tell. The `writeUserData` is now wrapped; the catalog returns regardless.

**[12] Exact string equality across a boundary PAPI rewrites.** Confirmed. Switched to substring matching and documented why on the sentinel's TSDoc, with a test that raises it behind the cross-process prefix. I did look at `newPlatformError` + `PlatformErrorCode` — agreed it is the better mechanism, but it changes the shape of a public export mid-review, so I took the smaller fix. Happy to do the `PlatformError` version if you would rather.

### `useRetryablePromise` [4] [6] [14]

All three were real, and fixing them separately would have layered three patches on the same mechanism, so I changed how supersession is identified: the wrapped fetch is built in a **side-effect-free `useMemo`** and compares itself against the fetch the effect actually **committed**, instead of against a counter bumped during render.

- [4] falls out: a factory going away sets the committed ref to `undefined`, so the abandoned fetch cannot set `hasError`. Two tests (rejection and resolution).
- [14] falls out: nothing is written during render, and `refetch` reads the committed value, so it cannot run a factory from a discarded render.
- [6] falls out: the effect now only does bail-out `setState`s plus a ref write, so an unstable factory degrades exactly as `usePromise` does instead of dying on update depth. The test uses a never-settling promise, so the only thing that could produce a second render is the hook itself.

### [15] Reuse

Took the promotion. `RetryableErrorView` now lives in `platform-bible-react/components/basics/` with `onRetry`/`retryLabel` optional — omitting them renders the message alone, which is what `downloadsUnavailable` needs. All six copies route through it, and `panel-state-views` keeps a thin `PanelRetryableErrorView` that adds only `h-screen`. Everything now uses the default `Button`, so the picker's retry no longer looks different from the panels'. New ADR: `adr-retryable-error-view-is-the-shared-failure-zero-state`.

You were right that the export claim did not hold — `getResourcePickerBodyState`/`ResourcePickerBodyState` are now actually in `experimental.ts`, and `GetResources` adopts the discriminant, so the `no-nested-ternary` chain and its disable are gone.

### The lower-severity items you listed

- **Unrestored `scrollIntoView` mock** — fixed; `vi.unstubAllGlobals` does not reach a prototype assignment, so it now captures and restores by hand.
- **"Send Layout dialog" in the ADR** — fixed.
- **Duplicate localization keys** — I could not reproduce this one. A per-section scan of `en.json`, `es.json` and both `localizedStrings.json` files found no key defined twice within a language block. If you meant two keys carrying the same English text (e.g. `%resources_retry%` / `%resourcePicker_retry%` / `%shareLayoutDialog_retry%`), those are deliberate per-surface keys — a previous round asked for exactly that when Share Layout was borrowing the picker's. Point me at the pair if it is something else.
- **A `TODO` with no ticket** — this is the `useDblResourceCatalog` to `useRetryablePromise` migration. It still has no ticket; I left it naming the ADR slug rather than inventing one. Want a follow-up issue?

### On the test suite

Fair hit — [8], [9] and [13] were each green against something a user could not reach. Beyond the new tests, I mutation-verified the gate, the language predicate, the count suppression, the settle latch, the personal-list gate and the `notReady` branch by reverting each fix and confirming the matching test goes red. Two tests needed tightening before they would fail: the retry-does-not-unmount one had to hold the retry's fetch in flight, and the project-setting gate needed the per-key loading flag.

Please post the ~12 you left off — the four you named were all worth having.

**Verification:** `format:check` clean; `typecheck` clean; `eslint` clean across all three workspaces on the changed files; extensions 2188/2188, platform-bible-react 1473/1473, renderer 3116/3117 — the one failure is `web-view.component.test.tsx` timing out under full-suite load and passing on its own (known-flaky multi-window suite, untouched by this branch). No C# changes.

AI-assisted — [session](https://claude.ai/code/session_01Gdn2rznrir86N7tRZR55MV)

katherinejensen00:
Follow-up on the two threads I left open, both in `f908148`.

### [12] — done the deeper way, with one correction to the premise

The rejection is now raised through an exported `newResourceActionProviderNotReadyError()` carrying `PlatformErrorCode.FAILED_PRECONDITION`, and recognised through an exported `isResourceActionProviderNotReadyError()`. Callers no longer touch the raw string at either end.

But the code cannot be the discriminant, and I want to flag that rather than quietly leave a check that looks stronger than it is. Two independent reasons:

1. **It does not survive the boundary you were pointing at.** `doRequest` rebuilds a rejection as `newPlatformError(message, platformErrorCode)`, and `platformErrorCode` comes from `response.error.data?.data?.platformErrorCode` — which is populated only for C# `PlatformErrorCodes.WithCode` throws. A TypeScript rejection raised in the extension host arrives with the code **stripped** and the prefix added. `isJsonRpcMethodNotFoundError` in `rpc.model.ts` documents exactly this and reads its own code back out of the message for the same reason. So a code-only check would have failed precisely in the scenario your finding describes.
2. **The code says only what class of failure this is.** A genuine `installDblResource` failure that is also a failed precondition would then be relabelled "resources are not ready yet" — a different lie from the one the sentinel prevents, but still one.

So the sentinel stays the identity of the condition and the code is the class, which is what the `PlatformError` machinery is for. The durable half is the exported predicate: the next caller calls it instead of re-deriving string matching, which was the actual failure mode you identified.

I also fixed the second half of your finding — the one about the non-sentinel branch. Every real install/uninstall failure crosses a process boundary, so its message reaches the alert carrying `JSON-RPC Request error (-32000): `. That prefix is now stripped before display. New test covers it.

If you would still rather have a code-only check, the way to make it work is to give the JS side a `platformErrorCode` path across the boundary the way C# has — that is a network-service change and a bigger ticket than this PR. Say the word and I will file it.

### The ticketless TODO — agreed, and filed as PT-4518

I do think it is worth doing, on maintenance cost rather than tidiness. `useDblResourceCatalog` (129 lines) and `useRetryablePromise` (132 lines) implement the same mechanism, and your [4] [6] [14] each had to be reasoned about twice — once for the shared hook, once for the duplicate. The duplicate is fine today only because its factory happens always to be defined and it happens not to write refs during render; nothing enforces either.

`PT-4518` is a standalone Dev Task rather than a sibling of PT-4433, since it is not a Sprint 89 non-negotiable and nobody is blocked on it. It records what should remain in the hook (the PAPI call plus the `available` / `notConfigured` / `notReady` / rejection mapping), what should go (its own generation ref, the `fetchResources` double-render trick, the local error-clearing), and the one thing the migration has to get right: `isCatalogReady` maps onto `hasSettled && !hasError`, not `hasSettled` alone — `hasSettled` is true after a rejection too, and `canPublishResourcePanelProjectIds` reads that flag.

The in-code TODO now names PT-4518, and the ADR's "remains a near-duplicate" paragraph points at it with that subtlety attached.

**Verification:** format, lint and typecheck clean on the changed files; `platform-get-resources` 37/37 and `platform-scripture-editor` suites pass. Three unrelated extension tests timed out under full-suite load on my machine and pass individually (`home.component`, `character-marker-bar`, `resource-cell-view`) — none are touched by this change.

extensions/src/platform-get-resources/src/get-resources.component.tsx
katherinejensen00: Confirmed, and I took the PlatformError mechanism — the rejection is now raised through newResourceActionProviderNotReadyError() carrying FAILED_PRECONDITION, and recognised through an exported isResourceActionProviderNotReadyError(), so callers touch the raw string at neither end.

One correction to the premise, worth knowing: the code doesn't survive the boundary you're pointing at either. doRequest reads platformErrorCode from response.error.data?.data?.platformErrorCode, which is populated only for C# PlatformErrorCodes.WithCode throws — a TypeScript rejection from the extension host arrives with the code stripped. isJsonRpcMethodNotFoundError in rpc.model.ts documents the same constraint and reads its own code back out of the message for that reason. Separately, the code only names the class: a real install failure that's also a failed precondition would get relabelled "resources are not ready yet".

So the sentinel stays the identity and the code is the class, which is what PlatformError is for. The durable half is the exported predicate. I also fixed the second half of your finding — the JSON-RPC Request error (…):  prefix is now stripped from real failures before display.

extensions/src/platform-get-resources/src/get-resources.web-view.tsx
katherinejensen00: Confirmed, and it was a regression from the hard-coded false. The error state now only replaces the list when there's nothing else to show — the state exists so a failure isn't reported as "no resources found", not to throw away a catalog we still hold. I extracted the decision as shouldReportCatalogFailure in dbl-catalog.utils.ts with tests, since the web view has no test seam of its own.

extensions/src/platform-get-resources/src/main.ts
katherinejensen00: Confirmed, and the self-healing tell is the right diagnostic. The writeUserData is now wrapped and logged; the catalog returns regardless. Losing the cache costs a refetch next launch, losing the catalog costs the user the feature.

extensions/src/platform-scripture-editor/src/scripture-text-grid.web-view.tsx
katherinejensen00: Confirmed — it was the one place the @throws half of the contract was dropped, against the ADR this PR wrote. The grid is now on useRetryablePromise, logs the failure, and shows a real error-plus-retry state instead of the "use View Options to add texts" prompt. The decision is extracted to grid-body-state.utils.ts with 7 tests, since the web view itself has no test seam — a failed or not-yet-delivered catalog now outranks the pick prompt.

extensions/src/platform-scripture-editor/src/use-dbl-resource-catalog.hook.ts
katherinejensen00: Confirmed, including the second-order consequence. Fixed: notReady now sets hasCatalogError, which drops isCatalogReady, which takes canPublishResourcePanelProjectIds back to false — so the publish hazard closes structurally rather than needing its own guard. notConfigured keeps the delivered-empty-catalog path, since that one really is an answer. Two tests added, and I mutation-verified them by reverting the branch.

lib/platform-bible-react/src/components/advanced/resource-picker-dialog/resource-picker-dialog.component.tsx
katherinejensen00: Confirmed in both directions, and I agree the false negative is the worse one — it withdraws the only control that could reveal the hidden rows. The predicate is now exactly what you described: does the selection exclude a member of typeScopedResources. I also routed the combo-box trigger label through the same predicate so the trigger and the Clear-filters affordance can't disagree about whether a filter is in effect. Two tests; the false-negative one reproduces your catalog-changes-under-a-mounted-picker sequence.

lib/platform-bible-react/src/components/advanced/resource-picker-dialog/resource-picker-dialog.component.tsx
katherinejensen00: Confirmed — the comment described a guarantee the condition didn't actually make. The count now shows only for list and filteredEmpty, the two states where it describes something the user can act on, and filters go inert whenever the type-scoped set is empty, which closes the live-filters half too. Two tests, both typing a search term against a populated catalog first and then emptying it, so isFiltered is genuinely true and they fail on the old condition.

lib/platform-bible-react/src/components/advanced/resource-picker-dialog/resource-picker-dialog.component.tsx
katherinejensen00: Took the promotion. RetryableErrorView now lives in platform-bible-react/components/basics/ with onRetry/retryLabel optional — omitting them renders the message alone, which is what the downloads-unavailable state needs instead of an inert button. All six copies route through it, panel-state-views keeps a thin wrapper that adds only the full-panel height, and everything uses the default Button, so the retry no longer changes shape depending on which surface you reach it from. New ADR: adr-retryable-error-view-is-the-shared-failure-zero-state.

You were right that the API-Changes claim didn't hold — the two symbols are now actually exported from experimental.ts, and GetResources adopts the discriminant, so the nested-ternary chain and its eslint-disable are gone.

lib/platform-bible-react/src/hooks/use-retryable-promise.hook.ts
katherinejensen00: Right on all counts, including that the comment claimed two ways when there are three. Fixed together with [6] and [14] by changing how supersession is identified rather than patching each: the wrapped fetch is now built in a side-effect-free useMemo and compares itself against the fetch the effect actually committed, instead of against a counter bumped during render. This case then falls out — a factory going away sets the committed ref to undefined, so the abandoned fetch can't set hasError. Tests for both the rejection and the resolution path.

lib/platform-bible-react/src/hooks/use-retryable-promise.hook.ts
katherinejensen00: Fixed. One note on the suggested fix: a monotonic counter alone doesn't close it, because setCount(c => c + 1) in the effect is also never bail-out-able — the loop comes from the effect storing new state on every run, not from the closure identity specifically. The fix that works is making the effect's only writes bail-out-able: it now does a ref write plus two setStates that are already at their target value, so an unstable factory degrades to re-fetching per render exactly as usePromise does. Test uses a never-settling promise so the hook itself is the only thing that could produce a second render.

lib/platform-bible-react/src/hooks/use-retryable-promise.hook.ts
katherinejensen00: Fixed, and the "materially worse than a stale default" framing is the right one — running the wrong factory while invalidating the right one is a different class of problem from usePromise's precedent. Nothing is written during render any more: the wrapped fetch is built in a pure useMemo and only the effect commits it, so refetch reads the committed value by construction. A render React throws away now leaves nothing behind.

src/renderer/components/dialogs/share-layout.dialog.tsx
katherinejensen00: Confirmed. This and [5] pull opposite ways — don't mount without a catalog, but don't withhold the dialog for a transient failure — and since notConfigured is permanent on a core build, "don't mount without a catalog" would mean the dialog never opens there.

Resolved by separating delivered from successful: the dialog blocks until every input has settled once, then always renders, and counts out loud the saved DBL references it can't classify ("N shared resources can't be shown because…"), with a retry only when the reason is recoverable. They're still round-tripped unchanged on Confirm — the notice is what stops that being silent. Recorded as ADR adr-share-layout-renders-without-a-catalog.

src/renderer/components/dialogs/share-layout.dialog.tsx
katherinejensen00: Both halves fixed. The dialog no longer swaps itself out for any catalog state — it renders once its inputs are delivered and explains what it can't show (see [3]). And the admin gate now runs ahead of every other branch, so a non-admin can't reach a control that re-drives a fetch. Good catch that the gate's own comment had stopped being true.

src/renderer/components/dialogs/share-layout.dialog.tsx
katherinejensen00: Confirmed, and you're right that the test was hard-coding the loading flag. The gate now waits on delivery of canWrite, all three project settings (via their 4th element), both personal lists, and the catalog. I also made the test mock's loading flag per setting key — with one shared flag, a gate that checked only one of the three still passed. I found that by mutating the gate and watching the test stay green; it fails now.

src/renderer/components/dialogs/share-layout.dialog.tsx
katherinejensen00: Both halves confirmed, and the warning about fixing one without the other was well placed — it is exactly what would have happened. The host now forwards isLoading || !hasSettled like the other two, and the mount gate latches on the first settle so a retry from inside the dialog can't unmount the content and discard the admin's edits. With the line-220 return gone, the two component tests now drive a path a user can actually reach. The latch has its own test, which I had to tighten before it would fail — it needed the retry's fetch held in flight, otherwise it resolved before the assertion.

katherinejensen00 (commented):
Great review, Matt! Thanks for doing that!

[8], [9] and [13] were each green against something a user couldn't reach. Beyond the new tests I mutation-verified the gate, the language predicate, the count suppression, the settle latch, the personal-list gate and the notReady branch by reverting each fix and confirming the matching test goes red. Two needed tightening before they'd fail. Please do post the ~12 you left off — the four you named were all worth having, and three of them were real.

> @katherinejensen00 made 16 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2760#-:-P0hugUy6mbH5vM9jUwb:b-kuby0v)* status: 0 of 57 files reviewed, 15 unresolved discussions (waiting on mattgetgen).

mattgetgen (approved):
Thanks for the changes! LGTM!

> @mattgetgen partially reviewed 67 files and all commit messages, made 1 comment, and resolved 15 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2760#-:-P0yRNtm51x9f9-HcYn6:bu5tyir)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

mattgetgen (approved):
:lgtm:

> @mattgetgen made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2760#-:-P0yRSAjB_GxObW_EXfy:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2761, opened 2026-09-02
Revert the Formatted-view marker palette to its pre-Standard-View behavior

tjcouch-sil:
## Summary

Pressing `\` in the Formatted view (and Markers view) opens the marker palette with **only inline markers** — `nd` shows up, `p`/`q`/`s1` do not. This reverts that palette to the implementation it had before the Standard View port.

### Why review this

Not part of the current epic. It is a user-visible regression introduced by the Standard View merge (#2565), found in hands-on use in Simple mode; the palette has been missing every block marker since that merged, so it is worth unblocking ahead of other non-epic work.

## What went wrong

`generateInlineMarkerMenuListItems` used to walk the static `usfmMarkers[parentMarker].children` table, which mixes block and inline markers by category (62 entries under `\p`: `p q s1 ...` alongside `nd wj f ...`).

The Standard View port replaced that with the shared library's stylesheet-driven `getMarkerMenuItems`, hardcoded to `source: 'character'`. That source emits only character- and note-styleType entries **by design** — PT9 chooses between its `character` and `paragraph` sources by where the caret sits relative to the paragraph's marker glyph, and the two are mutually exclusive. The Formatted view renders no marker glyphs, so no caret position in it can ever reach the paragraph source, and every block marker fell out.

Measured against the real bundled stylesheet, `\p` context: **36 items, 33 character + 3 note, 0 paragraph.**

Two things corroborate this was an oversight rather than a decision:

- The `isStructureProtected` + `isBlockMarker` guard in the same function became unreachable for every marker except `v` — the whole structure-protection path on this menu was inert.
- Its test was retargeted onto `v` as the "block marker" fixture. `v` is typed `styleType: 'character'` but special-cased as structural by `isBlockMarker`, so it is the one "block" marker that survives a character-source query. The suite stayed green while every real block marker vanished.

## Changes

`generateInlineMarkerMenuListItems` goes back to the pre-Standard-View implementation verbatim. That restores more than the missing markers — the stylesheet-driven version had changed the menu's whole shape, and all of it comes back:

| | Standard View port | Restored |
|---|---|---|
| Source | project stylesheet | static `usfmMarkers` children |
| Items under `\p` | 158 | 62 |
| Scoping | all markers valid in context | parent-scoped |
| Order | basic-first, then natural | alphabetical by marker code |
| `\c` | excluded | offered |
| Under a leaf marker (`\nd`) | 3 items | no menu |

**Kept from the port:** the `restoreSelection` callback that puts the caret back before inserting. That is an unrelated fix for this menu stealing focus with its own search input; dropping it would leave a pick landing nowhere. It is the only difference from the pre-Standard-View original, and it is a one-line removal if it should go too.

Consequences: the `styleInfo` argument is gone from the call site, and with it this module's last runtime import of `@eten-tech-foundation/platform-editor` — so the module doc now names `platform-bible-react` as the extension-host hazard it actually carries.

## Testing

- [x] `npx vitest run src/platform-scripture-editor` — 80 files, 1294 tests pass
- [x] ESLint clean on all three changed files
- [x] `tsc --noEmit` reports no errors in any changed file (the `paratextBibleSendReceive.*` errors in that output are pre-existing on `main`)
- [ ] Hands-on check in the running app — not done; worth a look before merge

The old suite passed while every block marker was missing, so the tests now cover **what the menu offers**, not just how one item behaves: both kinds present, parent scoping, alphabetical order, the empty result under a leaf marker, and title localization — alongside the structure-protection and caret-restore cases already there.

Each was mutation-verified rather than just observed green:

| Mutation | Tests red |
|---|---|
| Filter block markers out (reproduces the regression) | 2 |
| Drop the alphabetical sort | 1 |
| Ignore parent scoping | 6 |

## AI Involvement

Investigation, revert, and tests generated with Claude Code (Opus 5) and reviewed by me. The root cause was traced by bisecting `generateInlineMarkerMenuListItems` across the Standard View merge and confirmed by running the library's sources against the real bundled stylesheet before any code was changed.

AI-assisted — [session 1](https://claude.ai/code/session_0118oyULyPiQgL6YAgi1ftGp)

## Risk Level

Low — one function in one extension, restored to code that shipped for months before the Standard View port. No API surface change; the only call-site change is dropping the now-unused `styleInfo` argument.

https://claude.ai/code/session_0118oyULyPiQgL6YAgi1ftGp

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2761)

### DIFF
--- a/extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx
+++ b/extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.tsx
@@ -1918,7 +1918,6 @@ globalThis.webViewComponent = function PlatformScriptureEditor({
         notifyStructureProtected,
         restoreEditorSelection,
         contextMarker,
-        styleInfo,
       ),
     [
       contextMarker,
@@ -1926,7 +1925,6 @@ globalThis.webViewComponent = function PlatformScriptureEditor({
       isStructureProtected,
       notifyStructureProtected,
       restoreEditorSelection,
-      styleInfo,
     ],
   );
 
--- a/extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.utils.test.ts
+++ b/extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.utils.test.ts
@@ -1,14 +1,12 @@
 // @vitest-environment jsdom
-// The module under test pulls in `getMarkerMenuItems`/`defaultStyleInfo` as real (not type-only)
-// imports from `@eten-tech-foundation/platform-editor`, whose bundled entry point touches
-// `document` at module-eval time (it re-exports the whole editor, React components and all). The
-// default `node` environment (see `vitest.config.ts`) has no `document`, so this file needs jsdom
-// — same fix already used by `scripture-pane.test.tsx` and `use-editor-pdp-sync.hook.test.ts` for
-// the same package.
+// The module under test value-imports `platform-bible-react`, whose bundled entry point touches
+// `document` at module-eval time (it re-exports the whole component library). The default `node`
+// environment (see `vitest.config.ts`) has no `document`, so this file needs jsdom — same fix
+// already used by `scripture-pane.test.tsx` and `use-editor-pdp-sync.hook.test.ts`.
 import { describe, it, expect, vi } from 'vitest';
 import { MutableRefObject } from 'react';
-import type { EditorRef, SelectionRange, StyleInfo } from '@eten-tech-foundation/platform-editor';
-import { isLocalizeKey, type LanguageStrings } from 'platform-bible-utils';
+import type { EditorRef, SelectionRange } from '@eten-tech-foundation/platform-editor';
+import { isBlockMarker, isLocalizeKey } from 'platform-bible-utils';
 import {
   generateInlineMarkerMenuListItems,
   getChapterKey,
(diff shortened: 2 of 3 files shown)


mattgetgen (approved):
Great work! Claude doesn't have much to say on this PR, and I don't think it should block, but they are maybe worth considering.
Here's Claude's review:

Review's done. Five findings, none blocking-serious but two worth deciding on before merge.

Medium

1. \ becomes a dead key inside a character-marked run — platform-scripture-editor.web-view.utils.ts:65
parentMarker is the editor's contextMarker — the actual marker at the selection — so with the caret inside \wj …\wj*, \nd, \add, \bd it's a leaf entry with no children, and the function returns []. The PR body owns the "no menu under \nd" half of this, but the keydown handler at platform-scripture-editor.web-view.tsx:2334-2342 calls event.preventDefault() unconditionally, so the backslash is swallowed too: press \ inside a words-of-Jesus span and you get neither a menu nor a \. Suggested fix: fall back to blockMarker (already in the same state snapshot) when the context marker has no children.

2. \c is offered again and inserts a wrong-numbered chapter mid-chapter — :72
usfmMarkers.p.children includes c, and insertMarker('c') hits the library's chapter action, which inserts {type:"chapter", marker:"c", number: chapterNum + 1} when a chapter node already exists. In Power mode isStructureProtected is always false, so nothing disallows it: caret mid-verse in ch. 5 → \ → "c" → a \c 6 lands inside chapter 5. The library excludes c from every marker menu on purpose (markerItemSource.ts:87-101). It's restored pre-port behavior, but it's two keystrokes to a structural corruption.

Low

3. Table can offer markers insertMarker rejects, and the throw wedges the menu open — :74
Evaluated the whole table against the library's valid-marker lists: the only unsupported children are under parent id — usfm, h1, h2, h3, restore. Caret in the \id line → \ → "usfm" → insertMarker throws out of the cmdk onSelect, so closeMarkersMenu() never runs; nothing inserts and the palette stays open. The replaced implementation filtered these via isUsjMarkerSupported.

4. Unguarded table lookup, no last-resort title — :76
usfmMarkers[marker].description dereferenced without ?. (twice), unlike every other lookup in the file; the deleted resolveMarkerMenuItemTitle also had a marker-code fallback. Safe on today's data (the agent verified every child resolves with a non-undefined description and no parent has duplicate children), but usfm-markers.ts is generated from usfm.sty — a regeneration listing an absent child would throw and kill the whole palette instead of dropping one row. usfmMarkers[marker]?.description ?? marker is free.

5. Test name overstates what it asserts — platform-scripture-editor.web-view.utils.test.ts:146
'when not protected: no item is disallowed and all actions insert' now only exercises items[0].action?.(); the diff replaced the loop. 61 of 62 actions under \p go unasserted. Restore the loop or rename.

Explicitly not filed

Loss of project-stylesheet fidelity — custom.sty markers no longer appear, and markers the project's stylesheet doesn't define are still offered and insertable (isUsjMarkerSupported checks the library's hardcoded lists plus extraValidMarkers, not the sheet). That's the PR's stated, intentional trade-off.

The agent also confirmed the revert is verbatim against the pre-port original (bc1b47fc7a6^) apart from the retained restoreSelection?.(), styleInfo is still used elsewhere in the web view, no imports went dead, and the module doc's import-boundary claim checks out. Its own verification: the changed test file passes (42 tests), ESLint clean on all three files.

> @mattgetgen reviewed 3 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2761#-:-P0bjvHP9GsJvf9wfAiT:b7vayr1)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

Issue #47, opened 2023-03-02
Fix storybook production build?


tjcouch-sil:
~~Our production version of storybook (static page that shows the components and their abilities) isn't working. If we ever need to make it work (it may be a convenient way to document and demonstrate to extension developers what components they can use), we need to fix the production webpack configuration in [`.storybook/main.ts`](https://github.com/paranext/paranext-core/blob/39-storybook/.storybook/main.ts#L22).~~

The cost to fix the storybook publication build was too expensive so we decided to create a separate repo that would have all of the components and run storybook. This turned into a research task to understand what options are available for Figma and Storybook integration. Below is the link to the research results:

[Figma and Storybook Integration Research](https://docs.google.com/document/d/1X7UJ0tg5ownScYC-ruaesgDTEgnwdvKpsP-p39v9Tjw/edit#)

To test the production storybook, follow the steps here: https://storybook.js.org/docs/react/sharing/publish-storybook

Depends on #39

katherinejensen00:
This turned from a programming task to a research task. There are some sample projects linked in the research document:
[Figma and Storybook Integration Research](https://docs.google.com/document/d/1X7UJ0tg5ownScYC-ruaesgDTEgnwdvKpsP-p39v9Tjw/edit#)

Issue #44, opened 2023-03-01
Create the WebView API


darren8c:
Make webviews good. Make them persist, deserializeable, have api that controls the webview and the webview panel, etc.

Depends on #45, #31

irahopkinson:
After PR #167, adding an extension that adds a WebView will automatically add a tab in the dock layout after with the existing extension tabs. When adding a WebView, there is now a new `layoutType` property which defaults to _tab_. The other options of _panel_ and _float_ have not yet been implemented until we can specify exactly where in the dock layout a item should be added. If an extention's WebView is removed the tab will be blank. The tab can be closed by the user but is not currently automatically removed.

Before and after this PR, WebViews don't survive a reload (browser refresh), i.e. the extension tabs will be blank. This will require some re-working of PAPI to enable reload to still show extensions on a reload.

irahopkinson:
Do note, since it's now saving the layout, if you close tabs they might not come back since we have no way currently to require them to. To fix this:

1. After a fresh restart/start, in **Developer Tools**, click **Application**.
2. Expand **Local Storage** and click **`http://localhost:1212/`**.
3. Delete the _dock-saved-layout_ key and value.
4. Restart.

irahopkinson:
After PR #173, you can close WebView tabs and they will come back on a restart.

PR #36, opened 2023-02-22
Run the dotnet Data Provider from electron

irahopkinson:
- console messages are passed through to the electron console
- add GHA dotnet build
- also `npx update-browserslist-db@latest`
- note: in dev, the dotnet process should be killed when the app quits

Issue #34, opened 2023-02-22
Implement Electron security policies


tjcouch-sil:
https://www.electronjs.org/docs/latest/tutorial/security#7-define-a-content-security-policy

PR #2763, opened 2026-09-02
perf(dev): cut dev startup from 53s to 29s

mattgetgen:
Cuts dev `npm start` from **53.0s to 28.7s** end to end (first chapter of scripture actually rendered), and **27.9s to 6.9s** to first paint.

Fifteen commits, deliberately separable — the table shows what each of the perf ones is worth. Reviewed by @jasonleenaylor and @lyonsil; all 17 findings are addressed, with the six review commits listed at the bottom.

## What each change does, and what it bought

Each row measured with the rows above it applied, two warm runs apiece, **all on the current tip of `main`**:

| Commit | Change | Paints | Loaded | Δ loaded |
|---|---|---|---|---|
| — | baseline (current `main`) | 27.9s | 53.0s | — |
| `34cfaf9` | filesystem cache on the 3 dev webpack configs | 9.7s | 42.7s | **−10.3s** |
| `8add112` | drop the dev main config's unread `preload` entry | 9.4s | 44.0s | *no change* |
| `401d2f0` + `0d6ca2c` | tsx instead of ts-node (startup scripts, forked extension host) | 7.7s | 32.0s | **−11.7s** |
| `f1a3912` + `8c07762` | `PT_DOTNET_NO_WATCH` / `start:no-dotnet-watch` | 6.9s | 28.7s | −3.3s |

**The last row needs the opt-in flag set** — plain `npm start` is unchanged by those two commits, so the shipped default lands at the `32.0s` row. Everything above it is unconditional in dev.

## Two corrections to the original description

**"Extension imports fell 25.7s → 18.8s … from the CPU the removed compile had been burning" is withdrawn.** It was measured before the previous `main`, with only the cache applied, so it was not contaminated by the tsx commits — those did not exist yet. But it independently **failed to reproduce** when re-measured: 24.5s baseline, then 20.8s and 24.3s. That spread is noise, not an effect. The cache buys first paint; it does not measurably speed up extension imports.

**The old numbers were from the previous `main`.** The table above replaces them, re-measured per change on the current tip so the attribution is per-commit rather than branch-tip.

What *did* move extension imports is `0d6ca2c`. `-r ts-node/register/transpile-only` hooks `require` process-wide and, because the root tsconfig sets `allowJs: true`, was processing the 14–70MB extension bundles on the way through. Imports **21.5s → 9.8s**, concentrated in the two largest bundles exactly as that explains — platformScripture (70MB) 8.95s → 2.95s, the editor (59MB) 6.35s → 3.29s, small extensions unchanged.

Worth noting it is *not* that esbuild is a faster transpiler: on that 73.6MB bundle `ts.transpileModule` takes 1.82s against esbuild's 1.98s. A `require` of it measures 4.97s under ts-node, **0.37s with `allowJs: false`**, and 0.83s under tsx against 0.46s with no loader at all. The win is not transpiling it, not transpiling it faster.

## The one runtime-behaviour change: `PT_DOTNET_NO_WATCH`

Off by default; `npm start` behaves exactly as before. Opting in via `npm run start:no-dotnet-watch` — or the new `Debug Platform (no .NET watcher)` launch compound — runs the already-built assembly instead of `dotnet watch`, moving `.net process-start` from 18.4s to 3.9s.

**The trade-off is that C# changes stop being picked up.** You must run `npm run build:data` **and restart the app** after editing C#, since the provider is started once and never re-spawned in place. Otherwise it keeps running the previous build — starting and behaving completely normally, just against older C#. Because that failure is invisible from the outside, the provider logs a line whenever the flag is active, and logs a legible error when there is no Debug build to run. Documented in the README; pinned by a new unit test.

## Hot reload is unaffected — verified, not assumed

| Path | Result |
|---|---|
| `src/renderer/index.tsx` | in the served bundle in ~1s; gone ~1s after revert |
| `src/extension-host/extension-host.ts` | fork restarted with the new code, ~1s |
| `src/main/main.ts` | rebuild + `electronmon` restart, ~4s |
| edit while webpack **not** running, then start | picked up; 6.1s compile — changed module only |

That last row is the case watch mode cannot cover, and is why `buildDependencies` lists the configs, tsconfig, the DLL manifest and the `patches` files.

## From review

Six commits. Thanks to both reviewers — every claim I checked held up, and two of them corrected me.

- `821dcb6e51f` — nine mechanical findings: `maxMemoryGenerations` restating a default, a cache-directory comment giving the wrong reason, a change-narration comment, `scripts` sort order, recording why ts-node has to stay, and four README gaps (`clean:build-cache`, a reproduction pointer for the timings, the exact-`true` rule, and that a C# rebuild also needs a restart).
- `f49f714a520` — **scoped the dotnet `VersionSuffix` to Release** (#1). It re-evaluated on every build, rewriting `AssemblyInfo.cs` so `CoreCompile` never skipped. @lyonsil suggested this as a follow-up since the file is outside the original diff; folded in here instead as it is squarely on-topic. Verified `dotnet publish` defaults to Release, so shipped artifacts keep the timestamp. Build phase improves (four project compiles 5.3s → 2.5s) and, with Hot Reload disabled, time to the provider's first log line goes 4.04s → 2.07s. With Hot Reload on it moves nothing, because "Started" waits on Hot Reload's own project-graph pass rather than on the build. It also stops the rewritten file retriggering the watcher — previously six unprompted rebuilds in 198s, now one build then silence.
- `c34dedf3ad6` — **removed the `NODE_ENV` cache guard I added earlier in review.** `cache.version` is not part of `cacheLocation`, so it never gave a second `NODE_ENV` its own pack; it just discarded the pack on mismatch. And `DefinePlugin`'s `valueDependencies` already prevent a stale inlined value being served. The original concern was legitimate; webpack already covers it.
- `263c7700609` — ships `Debug Platform Backend (no .NET watcher)` and a matching compound, so nobody has to edit tracked `.vscode/launch.json`.
- `9091ba87ae1` — a unit test pinning the `PT_DOTNET_NO_WATCH` contract (five cases; confirmed falsifiable — deleting the branch fails two), plus cwd consistency between the two dev modes and a legible error when no Debug build exists.
- `6b8d46cbde1` — patch files added to `buildDependencies`, so a `patch-package` edit invalidates the caches instead of being masked by `snapshot.managedPaths`; and `check-port-in-use` renamed to `.mjs` to run on plain node.

## Notes

- The duplicated main build stays. The `--watch` build in `start:main` *is* the main-process hot reload, and the `prestart` build is what guarantees a fresh bundle before `electronmon` launches Electron.
- **The first start after merging is not faster for anyone** — it is the run that writes the caches. Editing any of these webpack configs invalidates its own cache too, since each config is one of its own `buildDependencies`.
- Checks: `npm run typecheck` and repo-wide `npm run lint` clean; 3,009 TypeScript and 1,704 C# tests passing. Every measured run reached `first-get-chapter-usx`, which doubles as a functional check.
- Figures are dev-mode, one machine (12-core Linux, headless via xvfb). `dotnet watch` alone varied 11.8–25.4s run to run, so treat the ordering of costs as the finding and the absolute seconds as a snapshot.

Two things investigated and **rejected**, for the record: parallelizing the `prestart` build (≈1s better paint, ≈1s later Electron start, end-to-end a wash) and `dotnet watch --no-hot-reload` (saves startup time that is concurrent anyway, while making the C# edit loop 0.78s → 3.00s).

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2763)


README.md
```diff
@@ -265,6 +265,18 @@ npm start
...
 
+### Starting without the .NET watcher
+
+`npm start` runs the .NET data provider under `dotnet watch`, which restores and builds the project before the provider's `Main()` runs — 15-24 seconds of dev startup, depending on how warm the MSBuild and Roslyn servers are. If you are not editing C#, you can skip it:
+
+```bash
+npm run start:no-dotnet-watch
+```
+
+This runs the already-built assembly instead, which takes roughly 3 seconds off the time to a fully loaded app and about 15 seconds off how long the data provider takes to become ready.
```
jasonleenaylor, line 278: The section documents the script but never names `PT_DOTNET_NO_WATCH`, which is a step away from how the other two dev flags are written up — `PT_STARTUP_MARKS=true npm start` at line 295 and `PT_ANALYTICS_TEST_OVERRIDE=true npm start` at line 320 both name the variable.

It costs someone concrete. Line 266, two lines above this section, points developers at VSCode's `Debug Platform`, and that config (`.vscode/launch.json:105` composing `:46`) runs `npm run start` with a fixed `env` block — so it cannot reach the npm script. The variable is the only handle those developers have, and it is the one thing the section does not give them. One sentence saying the script just sets `PT_DOTNET_NO_WATCH=true`, and that you can set it directly if you launch another way, would cover it.

src/main/services/extension-host.service.ts
```diff
@@ -212,8 +212,11 @@ async function startExtensionHost(maxWaitTimeInMS: number, isRestarting = false)
 
     const extensionHostExecArgv = app.isPackaged
       ? process.execArgv
-      : // Set up ts-node in the extension host process so it can run un-bundled TypeScript source files in dev
-        [...process.execArgv, '-r', 'ts-node/register/transpile-only'];
+      : // Set up tsx in the extension host process so it can run un-bundled TypeScript source files
+        // in dev. tsx transpiles with esbuild and resolves the tsconfig `paths` aliases this
+        // process's imports rely on, so it is a drop-in for ts-node's transpile-only registration
+        // that spends less time starting up.
+        [...process.execArgv, '--import', 'tsx'];
```
jasonleenaylor, line 217: `nodemon.json:5` still runs `.ts` through `-r ts-node/register/transpile-only`, and `npm run start:extension-host` — the `Debug Extension Host` launch config at `.vscode/launch.json:68` — goes through that. So the extension host the app forks now loads with tsx, while the one that script starts still loads with ts-node.

Is that intentional, or just not needed yet? I haven't tried `electron --import tsx` under nodemon, so there may be a reason it was left alone that I can't see from here.

.erb/configs/webpack.config.renderer.dev.ts
```diff
@@ -56,6 +56,33 @@ const configuration: webpack.Configuration = {
     },
   },
 
+  // Persistent caching, so the dev server's FIRST compile can reuse the previous run's work. The
+  // window created by the main process is blank until this compile finishes, so this compile is
```
jasonleenaylor, line 60: Nothing keys these caches to `NODE_ENV`, and the comment at line 175 says overriding it with `staging` is a supported thing. `EnvironmentPlugin({ NODE_ENV: 'development' })` at line 179 (and the one in `webpack.config.base.ts:99` that main and preload merge) inlines that value into every module that reads `process.env.NODE_ENV`, but webpack defaults `cache.version` to `""` and `cache.name` to `${config.name}-${mode}` (`webpack/lib/config/defaults.js:578-583`). These configs set neither, and `mode` stays the literal `'development'` whatever `NODE_ENV` is — `checkNodeEnv` only fires on `production`. So a `staging` run and a normal one share a cache directory and can reuse each other's modules with the wrong value baked in.

I have not produced a stale bundle, only read this off the config and webpack's defaults, and the workflow needs someone to actually take that override. `version: process.env.NODE_ENV ?? 'development'` on each cache would settle it if you think it is worth the line. Worth saying too that `renderer.prod` has carried the same exposure since well before this PR, so this may be a pattern question rather than anything you introduced.

jasonleenaylor (commented):
I ran this locally and it felt like a real improvement.

The description and title cover `34cfaf9` only. Four other commits ride along: `8add112` drops the dev main config's unread `preload` entry, `401d2f0` and `0d6ca2c` swap ts-node for tsx in the startup scripts and in the forked extension host, and `f1a3912`/`8c07762` add `PT_DOTNET_NO_WATCH` with its `start:no-dotnet-watch` wrapper. They are genuinely independent rather than tangled: three of them touch no file the cache commit touches, and the fourth is a separate hunk in `webpack.config.main.dev.ts` (the `entry` at lines 21-33 against the `cache` block at 34-54). So this is a description gap, not work to unpick.

Two reasons to close it before merge. The .NET opt-out is the only change here that alters a shipped service's runtime behaviour, and its trade-off belongs where a reviewer meets it rather than in `dotnet-data-provider.service.ts` — a stale C# build looks exactly like a working app.

The second is that the numbers may not be measuring what they say. The tsx swap and the preload-entry removal are both unconditional in dev, so any end-to-end run of this branch includes them. "Extension imports fell 25.7s -> 18.8s with no extension change, from the CPU the removed compile had been burning" is also precisely what `0d6ca2c` would produce on its own, by not loading ts-node in the extension host at all. If the table was measured at the branch tip, the attribution needs splitting; if it was measured at `34cfaf9` alone, saying so settles it. This may not matter as the overall improvement of the sum of these changes is what we're looking for.

A nit and two questions inline, none of them blocking.

_This review was assisted by Claude Opus 5._

README.md
mattgetgen: Verified and fixed in `e91345a`.

You're right about `Debug Platform` specifically: it's a compound running `Debug Platform Backend`, which invokes `npm run start` with a hardcoded `env` block (`MAIN_ARGS`, `IN_VSCODE`, `DEV_NOISY`), so it can't pick up a different npm script. The variable really is the only handle there.

The section now names `PT_DOTNET_NO_WATCH`, shows setting it directly, and says which `launch.json` block to add `"PT_DOTNET_NO_WATCH": "true"` to.

src/main/services/extension-host.service.ts
mattgetgen: Not intentional — I missed that path. Fixed in `68da810`.

`electron --import tsx` works fine under nodemon; there was no reason it was left alone. It's the same shape the forked host already uses, and that fork is also `electron` running as node, so the combination was effectively pre-proven.

Measured standalone, from nodemon launching the process to the host's first log line (bootstrap plus the initial import graph):

| | |
|---|---|
| ts-node | 1.07s |
| tsx | 0.49s |

No module-resolution errors, and it reaches the same stage (same `NODE_ENV` line, then the expected websocket failure with no PAPI running). The larger effect should be the same one the forked host saw — this hook intercepts `require` process-wide and, with `allowJs: true`, was handling the 14-70MB extension bundles on the way through.

Two things I checked before changing it: `npm start` is unaffected (it forks the host directly and never reads `execMap`), and `extension-host.service.ts:248-267` reads only `ext`, `watch` and `delay` from this file, so its chokidar watch list doesn't change.

.erb/configs/webpack.config.renderer.dev.ts
mattgetgen: Your reading checks out — I confirmed `cache.name` defaults to `${name}-${mode}` and `cache.version` to `""` at `webpack/lib/config/defaults.js:578-583`, and `mode` is the literal `'development'` in all three configs regardless of `NODE_ENV`. Took the fix as suggested in `75a7c38`: `version: process.env.NODE_ENV ?? 'development'` on each of the three dev caches.

Agreed it's latent rather than live — `start:renderer` and friends pin `NODE_ENV` through cross-env, so reaching it needs someone to actually take the `staging` override by editing the npm scripts. Cheap enough to close regardless.

Left `renderer.prod` alone, on your reasoning that it predates this branch and is a pattern question rather than something this PR introduced. Happy to do it here instead if you'd rather it not linger.

Verified the caches still work after re-keying: renderer first compile 4.10s warm, against 3.95-4.00s before.

.erb/configs/webpack.config.main.dev.ts
```diff
@@ -34,6 +38,26 @@ const configuration: webpack.Configuration = {
...
   },
 
+  // Persistent caching. This config is built twice per `npm start`: once blocking in `prestart` (so
+  // the bundle exists before `electronmon` launches Electron) and again by `start:main` in watch
+  // mode (which is what hot-reloads main). Both builds are unavoidable, so cache them instead -
+  // otherwise the same work is done from scratch twice while the renderer is compiling.
+  cache: {
+    type: 'filesystem',
+    cacheDirectory: path.join(webpackPaths.rootPath, 'node_modules', '.cache', 'webpack-main-dev'),
+    // Keyed to NODE_ENV because nothing else is: `mode` is the literal 'development' regardless of
```
lyonsil: **#10 - the `version` key does not do what the comment above it says**

`cacheLocation` is built from `cacheDirectory` + `cache.name` only (`webpack/lib/config/defaults.js:610-615`), so `version` never gives a second NODE_ENV its own storage. When it differs, webpack throws the whole pack away (`PackFileCacheStrategy.js:1247-1249`, "but version doesn't match") and rebuilds cold.

It also is not needed for the case it describes: `DefinePlugin` records `valueDependencies` per module and `NormalModule` re-checks them on restore, so a stale inlined `NODE_ENV` is never served. Checked with a fixture - alternating `version` values rebuilt cold every time, while keying on `cache.name` left two sibling packs that both restored warm.

**Suggested change:** remove the `version` line and its four-line comment from all three dev configs. If you would rather keep the guard, put it in the slot that creates separate packs instead:

```ts
name: `main-dev-${process.env.NODE_ENV ?? 'development'}`,
```

This line came from the discussion in r3927661090, and the concern raised there was a fair one - webpack just already covers it.

lyonsil (commented):
## Review notes

A few suggestions from a deep review of this branch at `75a7c38`. Each one comes with a concrete change that I checked against the code rather than just proposed, and where something is a judgement call rather than a fix it says so.

Worth saying up front: the measurements in the description hold up well under scrutiny, and most of what follows is small polish on an already-careful change. Nothing here is a blocker.

One suggestion could not be attached to a line in this diff, so it is included below.

---

**#1 - most of the `dotnet watch` cost this PR works around is self-inflicted** (`c-sharp/ParanextDataProvider.csproj:18`)

`<VersionSuffix>$([System.DateTime]::UtcNow.ToString(...))</VersionSuffix>` re-evaluates on every build, so `ParanextDataProvider.AssemblyInfo.cs` is rewritten each time and `CoreCompile` never skips. Measured in a private copy: three consecutive no-change `dotnet build` runs at 3.01/4.67/3.07s, versus 2.55s then ~1.15s with the suffix pinned, with neither `AssemblyInfo.cs` nor the dll touched. On the path that matters, `dotnet watch` to the provider's first log line went 11.7-12.8s down to 3.6-4.4s.

It appears to be safe to change: that line is the only occurrence of `VersionSuffix`/`InformationalVersion` in the repo, `AssemblyVersion`/`AssemblyFileVersion` stay `0.1.0.0`, nothing in `c-sharp/` reads it back, and the app's version comes from `AppService.GetAppInfo`.

**Suggested change** (a follow-up, not something to fold into this PR - the file is outside your diff):

```xml
<VersionSuffix Condition="'$(Configuration)' == 'Release'">$([System.DateTime]::UtcNow.ToString(`yyyyMMddHHmmss`))</VersionSuffix>
```

This complements rather than competes with `PT_DOTNET_NO_WATCH`: your `run --no-build` path still reaches `Main` about 3.3s faster than a pinned watch build, and pinning saves nothing on the first start after a real C# change. But it recovers most of the cost for everyone who keeps hot reload on. (The timings above were taken with `-p:VersionSuffix=fixed`; the `Condition` form is just the natural way to apply the same thing in-tree.)

.erb/configs/webpack.config.main.dev.ts
```diff
@@ -34,6 +38,26 @@ const configuration: webpack.Configuration = {
...
+    // NODE_ENV, and webpack defaults `cache.name` to `${name}-${mode}` and `cache.version` to "".
+    // Without this, a run with a different NODE_ENV would share this cache and could reuse modules
+    // that already have the other run's value inlined by `EnvironmentPlugin`.
+    version: process.env.NODE_ENV ?? 'development',
+    buildDependencies: {
```
lyonsil, line 48: **#12 - `patch-package` edits will not invalidate these caches**

webpack's default `snapshot.managedPaths` validates everything under `node_modules` by package `name@version` rather than by content, so editing a patch and re-running `npm install` restores the pre-patch module with no rebuild. The practical exposure here is small - `electron-log`'s `format.js` in the main dev cache, whose symptom would be duplicated log timestamps - because the renderer's patched packages are DLL entries that `postinstall` rebuilds.

**Suggested change:** add the patch **files** to `buildDependencies`:

```ts
import fs from 'fs';
const patchesDir = path.join(webpackPaths.rootPath, 'patches');
// ...inside buildDependencies:
patches: fs.readdirSync(patchesDir).map((f) => path.join(patchesDir, f)),
```

Worth being specific about the files rather than the directory: I tested passing `patches/` itself and webpack fails with `Caching failed for pack: Can't resolve '.../patches'` and then never writes a pack at all, which would quietly switch off the caching this PR is adding. The file-list form behaves correctly - silent restore when nothing changed, `build dependencies have changed` after editing a patch.

.erb/configs/webpack.config.main.dev.ts
```diff
@@ -34,6 +38,26 @@ const configuration: webpack.Configuration = {
...
+      config: [__filename, path.resolve(__dirname, 'webpack.config.base.ts')],
+      tsconfig: [path.resolve(webpackPaths.rootPath, 'tsconfig.json')],
+    },
+    compression: 'gzip',
+    maxMemoryGenerations: 5,
```
lyonsil: **#16 - `maxMemoryGenerations: 5` is already the default here**

`webpack/lib/config/defaults.js:623` sets `maxMemoryGenerations` to `development ? 5 : Infinity`, and `defaults.js:315` derives `development` from `mode`. All three dev configs are `mode: 'development'`, so this line restates the default and changes nothing. It does real work in `renderer.prod.ts:47`, where `mode: 'production'` would otherwise give `Infinity` and select a different memory-cache plugin - which is probably where it was copied from.

**Suggested change:** delete the line from the three dev configs, so the block only contains settings that are actually doing something.

.erb/configs/webpack.config.renderer.dev.ts
```diff
@@ -56,6 +56,38 @@ const configuration: webpack.Configuration = {
...
+  // directly on the path to first paint. Watch rebuilds are unaffected - they run off webpack's
+  // in-memory cache, and `buildDependencies` below invalidates the whole cache when anything
+  // outside the module graph that shapes the output changes.
+  cache: {
+    type: 'filesystem',
+    // Distinct from the production renderer's `webpack-renderer` directory: same source tree, very
```
lyonsil: **#20 - the reason given for a separate cache directory is not the real one**

Dev and prod could not have shared a pack even in one directory: `cache.name` defaults to `${name}-${mode}` (`defaults.js:578-582`), so the dev renderer would land in `webpack-renderer/default-development` and prod in `webpack-renderer/default-production`. You can see both on a built checkout.

The separate directory is still needed, just for a different reason - all three new dev configs are `mode: 'development'` with no `name`, so they would collide with **each other** at `default-development`. Worth fixing because the comment ten lines below cites that same default correctly, so the two disagree.

**Suggested change:**

```ts
// Distinct directory per dev config: all three are `mode: 'development'` with no `cache.name`,
// so they would share one `default-development` pack if they shared a directory.
```

README.md
```diff
@@ -265,6 +265,26 @@ npm start
 
 After you run `npm start` (or, in VSCode, launch `Debug Platform`), you can edit the code, and the relevant processes will hot reload.
 
+### Starting without the .NET watcher
```
lyonsil, line 270: **#23 - the new caches have no user-facing note about resetting them**

This PR adds three persistent on-disk caches. When one does go stale the recovery is `npm run clean:build-cache` (`package.json:63`), which is not mentioned in the README, `CLAUDE.md`, or `.context/standards/`. It is also a blanket `rimraf ./node_modules/.cache`, so it clears Storybook's and the extensions' caches at the same time - worth saying in the same breath.

**Suggested change:** a couple of sentences near the dev-startup instructions:

> Dev builds are cached under `node_modules/.cache`. If you ever suspect a stale bundle, `npm run clean:build-cache` clears every cache in that directory (Storybook's and the extensions' included) and the next build repopulates them.

README.md
```diff
@@ -265,6 +265,26 @@ npm start
...
+
+`npm start` runs the .NET data provider under `dotnet watch`, which restores and builds the project before the provider's `Main()` runs — 15-24 seconds of dev startup, depending on how warm the MSBuild and Roslyn servers are. If you are not editing C#, you can skip it:
```
lyonsil: **#25 - the timings have no way for a reader to reproduce them**

"15-24 seconds", "roughly 3 seconds" and "about 15 seconds" will drift with hardware and warmth, and they are the only unqualified performance figures in the README. The repo already ships the reproduction path - `PT_STARTUP_MARKS` plus `npm run startup-waterfall` - documented about 25 lines below in this same section, but nothing connects the two.

Your PR description already has the caveats ("one machine (12-core Linux, headless via xvfb)... `dotnet watch` alone varied 11.8-25.4s run to run"); they just did not make it into the shipped doc.

**Suggested change:** one pointer rather than a full restatement, e.g. "(measured on one machine; reproduce with the Startup performance timing tooling below)".

README.md
```diff
@@ -265,6 +265,26 @@ npm start
...
+
+```bash
+npm run start:no-dotnet-watch
+```
+
+This runs the already-built assembly instead, which takes roughly 3 seconds off the time to a fully loaded app and about 15 seconds off how long the data provider takes to become ready.
+
+The script is a thin wrapper that sets `PT_DOTNET_NO_WATCH=true`. If you start the app some other way, set the variable directly — this is the only handle you have when the npm script is out of reach:
```
lyonsil: **#26 - the exact-`true` rule is not stated**

`dotnet-data-provider.service.ts:111` tests `process.env.PT_DOTNET_NO_WATCH === 'true'`, so `1`, `TRUE` or `yes` silently do nothing. Both examples in this section use the correct value, so copy-paste is safe - it is the hand-typed case that is not covered.

There is already a precedent for calling this out: the `PT_ANALYTICS_TEST_OVERRIDE` section at README:331.

**Suggested change:** mirror that wording - "The value must be exactly `true`; any other value is ignored."

README.md
```diff
@@ -265,6 +265,26 @@ npm start
...
+
+```bash
+PT_DOTNET_NO_WATCH=true npm start
+```
+
+That includes VSCode's `Debug Platform`, whose `Debug Platform Backend` configuration runs `npm run start` with a fixed `env` block, so it cannot pick up the script. Add `"PT_DOTNET_NO_WATCH": "true"` to that block in `.vscode/launch.json` to get the same behavior while debugging.
```
lyonsil: **#29 - this asks every developer to edit a tracked file**

`.vscode/launch.json` is tracked (`git ls-files .vscode/`), so following this instruction leaves a permanent local modification in everyone's working tree, with no revert step documented.

**Suggested change:** ship a second configuration so nobody has to edit the tracked one, and point this paragraph at it:

```jsonc
{
  "name": "Debug Platform Backend (no .NET watcher)",
  "type": "node",
  "request": "launch",
  "runtimeExecutable": "npm",
  "runtimeArgs": ["run", "start"],
  "skipFiles": ["<node_internals>/**"],
  "env": {
    "MAIN_ARGS": "--inspect=5858 --remote-debugging-port=9223 --remote-allow-origins=http://localhost:9223",
    "IN_VSCODE": "true",
    "DEV_NOISY": "false",
    "PT_DOTNET_NO_WATCH": "true"
  },
  "autoAttachChildProcesses": true
}
```

Add a matching `compounds` entry too if you want to reach it from `Debug Platform`. One thing that does *not* work as an alternative: shipping `"PT_DOTNET_NO_WATCH": "false"` in the existing block. Because the check is `=== 'true'`, that is identical to omitting the key, and you would still hand-edit the tracked file to switch it on.

README.md
```diff
@@ -265,6 +265,26 @@ npm start
...
+
+**The trade-off is that C# changes are no longer picked up.** You have to run `npm run build:data` yourself after editing C#, or the app will keep running the previous build — it starts and behaves normally, just against older C# code, so the startup log says explicitly when this mode is active. You also need to have built the provider at least once (`npm run build:data`) before the script will work at all.
```
lyonsil: **#30 - `npm run build:data` alone will not pick the change up**

`startDotnetDataProvider` is called once, from `main.ts:470`, and the service only exposes `start`/`kill`/`waitForClose` - both used solely on the quit path. The chokidar restart machinery covers the extension host only, and the one command that would pick up new C# is `platform.restart` (`main.ts:1813`), which is a full relaunch.

So rebuilding while the app is running still leaves you on the old C#. That is easy to misread as "my change did not work", especially since README:268 has just promised that editing code hot-reloads.

**Suggested change:** append "then restart the app" to this sentence, and the same clause to the mirrored comment at `dotnet-data-provider.service.ts:104-107`.

package.json
```diff
@@ -82,12 +82,13 @@
     "package:debug": "cross-env DEBUG_PROD=true npm run package",
     "postinstall": "patch-package && npm run link-dev-packages && tsx .erb/scripts/check-native-dep.js && electron-builder install-app-deps && npm run build:dll && cross-env NODE_ENV=development TS_NODE_TRANSPILE_ONLY=true webpack --config ./.erb/configs/webpack.config.renderer.dev.dll.ts && npm run build --workspace=eslint-plugin-paranext",
     "prepare": "husky && npm run csharp:tool:restore",
-    "prestart": "ts-node ./.erb/scripts/generate-dev-build-info.ts && cross-env NODE_ENV=development TS_NODE_TRANSPILE_ONLY=true webpack --config ./.erb/configs/webpack.config.main.dev.ts",
+    "prestart": "tsx ./.erb/scripts/generate-dev-build-info.ts && cross-env NODE_ENV=development TS_NODE_TRANSPILE_ONLY=true webpack --config ./.erb/configs/webpack.config.main.dev.ts",
```
lyonsil, line 85: **#33 - worth recording that `ts-node` has to stay**

webpack-cli loads `.ts` config files through `interpret`/`rechoir`, and `interpret`'s `.ts` list is `["ts-node/register", ...]` - tsx is not in it. So every `webpack --config *.ts` script still boots ts-node, and `ts-node` has to remain a devDependency even though the startup scripts have moved to tsx.

Nothing in the repo records that, so a later "we migrated to tsx, drop ts-node" cleanup would break all eight webpack scripts at once. (I checked that converting them is possible - `NODE_OPTIONS="--import tsx" webpack --disable-interpret` works - but it measured no faster, so leaving them is the right call.)

**Suggested change:** a one-line comment where someone doing that cleanup would see it, e.g. at the top of `.erb/configs/webpack.config.base.ts`: webpack-cli resolves `.ts` configs via `interpret`, which only knows `ts-node/register`, so ts-node stays until that changes.

package.json
```diff
@@ -82,12 +82,13 @@
...
     "rebuild": "electron-rebuild --parallel --types prod,dev,optional --module-dir release/app",
-    "start": "ts-node ./.erb/scripts/check-port-in-use.js && npm run start:renderer",
+    "start": "tsx ./.erb/scripts/check-port-in-use.js && npm run start:renderer",
```
lyonsil: **#34 - small follow-on: this file has no TypeScript in it**

`.erb/scripts/check-port-in-use.js` is plain ESM JavaScript (`import chalk`, `import detectPort`), so a TypeScript loader has nothing to do. To be clear this line is a solid improvement already - warm timings are node 0.08s, tsx 0.22s, and the `ts-node` it replaced 1.07s - so this is about the remaining ~140ms, not a regression.

One thing to know if you do touch it: plain `node` emits the same four-line `[MODULE_TYPELESS_PACKAGE_JSON]` warning that tsx does, because the root `package.json` has no `"type"`. Only the `.mjs` extension removes it.

**Suggested change:** rename the file to `.mjs` and run it with node. `package.json:87` is its only reference in the repo:

```json
"start": "node ./.erb/scripts/check-port-in-use.mjs && npm run start:renderer",
```

package.json
```diff
@@ -82,12 +82,13 @@
...
     "start:data": "dotnet watch --project c-sharp/ParanextDataProvider.csproj",
     "start:extension-host": "cross-env NODE_ENV=development NODE_PATH=\"./release/app/node_modules\" nodemon ./src/extension-host/extension-host.ts",
     "start:extensions": "cd extensions && npm run watch",
+    "start:no-dotnet-watch": "cross-env PT_DOTNET_NO_WATCH=true npm start",
```
lyonsil: **#36 - this line breaks the `scripts` sort order**

Extracting all 73 keys and comparing against a sort shows exactly one mismatch: `start:main` (line 92) and `start:no-dotnet-watch` (line 91) are swapped. Since `m` sorts before `n`, `start:main` should come first - every other entry is already in order.

**Suggested change:** move the line below `start:main`.

Minor and take it or leave it: every other `start:*` script names a process it starts (`start:data`, `start:main`, `start:preload`, `start:renderer`), while this one names a mode. The wider `scripts` object does use mode suffixes elsewhere (`build:preload:dev`, `package:debug`), so this is only a convention within the `start:*` family.

src/main/services/dotnet-data-provider.service.ts
```diff
@@ -99,6 +99,25 @@ function startDotnetDataProvider() {
...
+  // `dotnet watch` restores and builds before the provider's `Main()` runs at all, which is the
+  // single largest block of dev startup. Opting out (`npm run start:no-dotnet-watch`) runs the
+  // already-built assembly instead, so that cost moves out of startup — at the price of hot reload
+  // on C# edits: you must run `npm run build:data` yourself after changing C#, or the app keeps
+  // running the previous build. Logged below, because a stale build is otherwise invisible: the app
+  // starts and behaves normally, just against older C#.
+  // `run --no-build` rather than the built assembly directly: the build output path is
+  // runtime-identifier specific (e.g. `bin/Debug/net8.0/linux-x64/`), so letting MSBuild resolve it
+  // keeps this working on every platform without hardcoding that layout.
+  if (!globalThis.isPackaged && process.env.PT_DOTNET_NO_WATCH === 'true') {
```
lyonsil, line 113: **#42 - worth a small test to pin this branch**

Nothing currently pins the contract here: honoured only when unpackaged, exact `'true'` match, the argv shape, and the packaged branch still overriding `args` afterwards. All four are easy to break accidentally later.

The good news is it is testable exactly as written - no refactor needed, since `dotnetDataProvider.start` is already exported and the module has no Electron `app` import at load time.

**Suggested change:** add `src/main/services/__tests__/dotnet-data-provider.service.test.ts` following the `vi.mock('@shared/services/logger.service', ...)` pattern used by about 18 existing tests in that directory - mock `child_process`, stub `process.resourcesPath`, set `globalThis.isPackaged`, call `dotnetDataProvider.start`, then assert the spawn arguments for four cases:

1. default -> `['watch', '--project', ...]`
2. `PT_DOTNET_NO_WATCH=true` -> `['run', '--project', ..., '--no-build']` plus the log line
3. `PT_DOTNET_NO_WATCH=1` -> ignored, still `watch`
4. packaged -> the packaged branch wins regardless of the flag

I wrote this shape against your branch to check it works: all four pass, and case 2 fails if the new branch is removed, so it is a real test rather than a passing-by-construction one.

src/main/services/dotnet-data-provider.service.ts
```diff
@@ -99,6 +99,25 @@ function startDotnetDataProvider() {
...
+    args = ['run', '--project', 'c-sharp/ParanextDataProvider.csproj', '--no-build'];
```
lyonsil: **#43 - the working directory differs between the two dev modes**

`dotnet watch --project X` launches the app with cwd set to the project directory; `dotnet run --project X --no-build` inherits the caller's cwd (the repo root), and `options` is only assigned inside the packaged branch below, so it stays `undefined` here. Confirmed with a throwaway project: `watch` printed `CWD=<repo>/c-sharp`, `run --no-build` printed `CWD=<repo>`.

Nothing in the C# reads a relative path today (the one cwd-sensitive site, `ParatextGlobals.cs:54`, gets absolute input from every caller), so this is consistency between the two dev modes rather than a live bug.

**Suggested change** - note the project path has to become absolute at the same time, otherwise `--project c-sharp/...` would resolve against the new cwd and fail:

```ts
const cSharpDir = path.join(globalThis.resourcesPath, 'c-sharp');
args = ['run', '--project', path.join(cSharpDir, 'ParanextDataProvider.csproj'), '--no-build'];
options = { cwd: cSharpDir };
```

---

**#44 - the failure when there is no Debug build is hard to read**

`npm run build` fans out to `build:data-release` (`dotnet publish`, Release) and never `build:data` (`dotnet build`, Debug). So immediately after `./.erb/scripts/refresh.sh` there is no Debug output, and this path exits 1 with `Win32Exception (2) ... No such file or directory`.

It is not silent - stderr goes through `logger.error` and names the missing path - but it lands as a raw MSBuild trace *after* the cheerful "running the prebuilt assembly" line, and `start()` is fire-and-forget, so the window opens and requests just hang. The README does document the prerequisite; this is about making the failure legible when someone misses it.

**Suggested change:** a cheap existence check before spawning, in the idiom of the existing `.erb/scripts/check-build-exists.ts`:

```ts
if (!fs.existsSync(path.join(cSharpDir, 'bin', 'Debug'))) {
  logger.error(
    formatLog(
      'PT_DOTNET_NO_WATCH is set but no Debug build was found. Run `npm run build:data` first.',
      DOTNET_DATA_PROVIDER_NAME,
    ),
  );
}
```

Check the `bin/Debug` directory rather than a binary path - the output is runtime-identifier specific (`bin/Debug/net8.0/linux-x64/` here), which is the same reason your comment gives for using `run --no-build` in the first place.

src/main/services/extension-host.service.ts
```diff
@@ -212,8 +212,11 @@ async function startExtensionHost(maxWaitTimeInMS: number, isRestarting = false)
 
     const extensionHostExecArgv = app.isPackaged
       ? process.execArgv
-      : // Set up ts-node in the extension host process so it can run un-bundled TypeScript source files in dev
-        [...process.execArgv, '-r', 'ts-node/register/transpile-only'];
+      : // Set up tsx in the extension host process so it can run un-bundled TypeScript source files
```
lyonsil: **#47 - the last clause of this comment describes what the code replaced**

"so it is a drop-in for ts-node's transpile-only registration that spends less time starting up" is the kind of change narration `.claude/rules/code-quality/forward-facing-comments.md` asks us to keep out of the source and put in the commit message instead.

The win itself is real and worth keeping in the comment - I measured the forked host at ~1137ms (ts-node) versus ~474ms (tsx) to its first log line, matching your own numbers. It just comes from esbuild's transpile throughput over the ~150-file graph rather than from loader startup; the two loaders boot within about 9ms of each other, so "spends less time starting up" points at the wrong mechanism.

**Suggested change:**

```ts
: // Run the extension host's un-bundled TypeScript through tsx in dev. tsx transpiles with
  // esbuild and resolves the tsconfig `paths` aliases these imports rely on.
  [...process.execArgv, '--import', 'tsx'];
```

.erb/configs/webpack.config.main.dev.ts
mattgetgen: Verified both halves and removed the line from all three dev configs in `c34dedf3ad6`.

Confirmed independently:

- `cacheLocation` is `resolve(cacheDirectory, cache.name)` at `config/defaults.js:609-615` — `version` genuinely is not in the path, so it could never have given a second NODE_ENV its own storage. And `PackFileCacheStrategy.js:1247` returns on mismatch with "but version doesn't match", so the effect was a cold rebuild rather than a sibling pack. Worse than not having the guard.
- The `valueDependencies` path is there as you describe: `DefinePlugin.js:436-450` records them, and `NormalModule.js:1630` re-checks on restore with the comment "build when valueDependencies have changed".

I took the removal rather than the `name` form, since webpack's own tracking makes the guard redundant either way and the cache block is better off holding only settings that do something.

Also flagged the reversal on r3927661090 so that thread does not keep implying the guard is still in place.

.erb/configs/webpack.config.renderer.dev.ts
mattgetgen: Following up here because this has since been reversed, and I would rather this thread not keep saying I took a fix that is no longer in the branch.

@lyonsil pointed out in #10 that `cache.version` does not do what I claimed for it, and checking it he is right on both counts:

- `cacheLocation` is `resolve(cacheDirectory, cache.name)` (`config/defaults.js:609-615`), so `version` is not part of the storage path and never gave a second NODE_ENV its own pack. On a mismatch webpack discards the pack outright (`PackFileCacheStrategy.js:1247`) — so the guard turned a NODE_ENV switch into a cold rebuild.
- The exposure was already covered: `DefinePlugin` records `valueDependencies` (`DefinePlugin.js:436-450`) and `NormalModule` re-checks them on restore (`NormalModule.js:1630`), so a stale inlined `process.env.NODE_ENV` is never served from cache.

So the concern you raised was a real one and I should have checked whether webpack already handled it before adding a guard for it. Line removed from all three dev configs in `c34dedf3ad6`.

.erb/configs/webpack.config.main.dev.ts
mattgetgen: Done in `6b8d46cbde1`, added to all three dev configs.

Verified the invalidation end to end on the main dev cache: an unchanged rebuild restores cache content, while appending a byte to `patches/electron-log+5.4.1.patch` gives "Restored pack ... but build dependencies have changed" followed by "resolving of build dependencies is invalid" and no restore.

Also confirmed the warning, because it would have been an unpleasant way to find out: passing the `patches` directory fails with `Caching failed for pack: Can't resolve '.../patches'` and then writes no pack at all. The file-list form still writes a pack (16MB here).

.erb/configs/webpack.config.main.dev.ts
mattgetgen: Done in `821dcb6e51f` — removed from all three.

Confirmed the derivation: `defaults.js:623` sets it to `development ? 5 : Infinity`, and `defaults.js:315` derives `development` from `mode === "development"`, which all three dev configs are. Left in place in `renderer.prod.ts`, where `mode: 'production'` means it is doing real work.

.erb/configs/webpack.config.renderer.dev.ts
mattgetgen: Done in `821dcb6e51f`, comment replaced with your wording.

Confirmed it on a built checkout the way you described — the packs on disk are `webpack-renderer/default-production` and `webpack-renderer-dev/default-development`, so dev and prod were never going to collide. The three dev configs colliding with each other is the real reason, and now the comment and the one ten lines below agree.

README.md
mattgetgen: Done in `821dcb6e51f`, using your wording more or less verbatim — including that `clean:build-cache` is a blanket `rimraf ./node_modules/.cache` and takes Storybook's and the extensions' caches with it.

README.md
mattgetgen: Done in `821dcb6e51f` — took the one-pointer option rather than restating the caveats, and linked it to the Startup performance timing section so the two are actually connected.

README.md
mattgetgen: Done in `821dcb6e51f`, mirroring the `PT_ANALYTICS_TEST_OVERRIDE` wording at README:331 as you suggested.

README.md
mattgetgen: Done in `263c7700609`. Shipped `Debug Platform Backend (no .NET watcher)` plus a `Debug Platform (no .NET watcher)` compound, so it is reachable the same way `Debug Platform` is and attaches the renderer debugger alongside. The README paragraph now points at the compound instead of asking anyone to edit the tracked file.

Confirmed the dead end you flagged: `"PT_DOTNET_NO_WATCH": "false"` in the existing block is indistinguishable from omitting the key, because the check is `=== 'true'`.

README.md
mattgetgen: Done in `821dcb6e51f` — added to the README sentence and to the mirrored comment in the service.

Confirmed the shape of it: `dotnetDataProvider.start()` is called once, at `main.ts:470`, and the only other uses of the service are `waitForClose`/`kill` on the quit path.

package.json
mattgetgen: Done in `821dcb6e51f`, as a comment at the top of `webpack.config.base.ts`.

Confirmed the list: `interpret.extensions['.ts']` is `["ts-node/register", "typescript-node/register", "typescript-register", "typescript-require", "sucrase/register/ts", {"module": "@babel/register"}]` — no tsx. Recorded it there rather than anywhere else because that is the file someone doing the cleanup would have open.

package.json
mattgetgen: Done in `6b8d46cbde1` — renamed to `.mjs` and run with plain `node`. `package.json` was indeed the only reference.

package.json
mattgetgen: Done in `821dcb6e51f` — moved below `start:main`; extracting the keys and comparing against a sort now returns no mismatches.

On the naming: leaving it as `start:no-dotnet-watch` for now. It does name a mode rather than a process, but it reads unambiguously at the call site, and the alternatives I tried were either longer or less obvious about what gets turned off.

src/main/services/dotnet-data-provider.service.ts
mattgetgen: Done in `9091ba87ae1`. Added `src/main/services/__tests__/dotnet-data-provider.service.test.ts` following the `vi.hoisted` + `vi.mock('@shared/services/logger.service', ...)` pattern from that directory, covering your four cases plus the missing-Debug-build error from #44 — five in all.

Checked it is falsifiable rather than passing-by-construction, as you did: deleting the branch fails 2 of the 5.

Two small things needed beyond the sketch, in case they save you time next occasion: `child_process` and `fs` both need a `default` export on the mock as well as the named one, or the service's import throws at collection; and the packaged case needs `process.resourcesPath` defined via `Object.defineProperty` rather than assigned, since Node's types have no such property and a cast trips `no-type-assertion`.

src/main/services/dotnet-data-provider.service.ts
mattgetgen: Both done in `9091ba87ae1`.

#43 — took the suggested form, including making the project path absolute, since a relative one would otherwise resolve against the new cwd. Pinned in the new test, which asserts `options` is `{ cwd: <resourcesPath>/c-sharp }`.

#44 — added the existence check on the `bin/Debug` directory, with your message. Checking the directory rather than a binary for exactly the reason you give, which is the same reason the branch uses `run --no-build` in the first place. Also covered in the test.

src/main/services/extension-host.service.ts
mattgetgen: Done in `821dcb6e51f`, taking your replacement wording as written.

Worth saying you were right about the mechanism as well as the comment. The two loaders do boot within noise of each other, and separately from your ~1137ms vs ~474ms on the forked host: benchmarking `ts.transpileModule` against `esbuild.transform` on `platform-scripture`'s 73.6MB bundle gives 1.82s vs 1.98s, so esbuild is not the faster transpiler there either. What makes the difference for the bundles is that `allowJs: true` had ts-node's hook processing every `.js` at all — a `require` of that file measures 4.97s under ts-node, 0.37s with `allowJs: false`, and 0.83s under tsx against 0.46s with no loader.

.erb/configs/webpack.config.main.dev.ts
```diff
@@ -34,6 +39,28 @@ const configuration: webpack.Configuration = {
...
+      config: [__filename, path.resolve(__dirname, 'webpack.config.base.ts')],
+      tsconfig: [path.resolve(webpackPaths.rootPath, 'tsconfig.json')],
+      // webpack's default `snapshot.managedPaths` validates everything under `node_modules` by
+      // package `name@version` rather than content, so editing a patch and re-running
+      // `npm install` would otherwise restore the patched module with no rebuild. List the patch
+      // files, not the `patches` directory — webpack cannot resolve a directory here and responds
+      // by writing no pack at all, silently disabling the cache.
+      patches: fs
+        .readdirSync(path.join(webpackPaths.rootPath, 'patches'))
+        .map((patch) => path.join(webpackPaths.rootPath, 'patches', patch)),
```
lyonsil: **Nit - the patches list is now copy-pasted into all three dev configs**

`main.dev`, `preload.dev` and `renderer.dev` each carry the same five-line comment and the same `readdirSync().map()`. `webpack.paths.ts` already owns every other shared path, so one entry there would leave a single copy:

```ts
// webpack.paths.ts
const fs = require('fs');
const patchesPath = path.join(rootPath, 'patches');
const patchFiles = fs
  .readdirSync(patchesPath)
  .filter((file) => file.endsWith('.patch'))
  .map((file) => path.join(patchesPath, file));
// ...then add `patchFiles` to the exported `webpackPaths` object
```

and `patches: webpackPaths.patchFiles,` in each config, with the comment kept once beside the list.

The `.filter` is the part I would take even if the duplication stays. Whatever lands in `patches/` goes straight into `buildDependencies`, and by the reasoning in the comment just above, an entry webpack cannot resolve makes it write no pack at all - so a stray subdirectory there would quietly switch dev caching off in all three configs.

mattgetgen: Both parts taken, in `c91997f592a`. `webpackPaths.patchFiles` now owns the list and the comment, with `patches: webpackPaths.patchFiles` at each of the three uses. Kept your `require`-style since that is what the file already uses.

You were right that the `.filter` is the part that matters, and it is worse than "would quietly switch dev caching off" implies — the build still **exits 0**. Confirmed by dropping a stray subdirectory into `patches/`:

| | caching | pack files written |
|---|---|---|
| with the `.filter` | fine | 2 |
| without it | `Caching failed for pack: ... Can't resolve` | **0** |

Both builds exited 0, so nothing in the exit code or the summary would tell you the cache had stopped working. That is the bit that makes it worth guarding rather than just tidier.

Also re-checked that invalidation still behaves after the move: unchanged rebuild restores cache content, editing a patch gives "build dependencies have changed" and no restore.

jasonleenaylor:
All my concerns have been addressed — verified at `c91997f59`, and my threads are resolved.

One of them was my error rather than yours: the `NODE_ENV` cache guard was added on my question and dropping it in `c34dedf3a` was the right call. @lyonsil's reading is correct, and it reaches the case I raised — `EnvironmentPlugin` constructs a `DefinePlugin`, which records `valueDependencies` that `NormalModule` rechecks on restore, so the staleness I asked about could not have happened.

I have not reviewed the commits since `75a7c38`; @lyonsil has that in hand, and I do not expect to have more to add here.

src/main/services/__tests__/dotnet-data-provider.service.test.ts
```diff
@@ -0,0 +1,130 @@
...
+  });
+
+  test('PT_DOTNET_NO_WATCH=true runs the prebuilt assembly, from the project directory', () => {
+    process.env.PT_DOTNET_NO_WATCH = 'true';
+
+    dotnetDataProvider.start();
+
+    const [command, args, options] = mocks.spawn.mock.calls[0];
+    expect(command).toBe('dotnet');
+    expect(args).toEqual([
```
lyonsil, line 79: **The new test's path expectations are POSIX-only, and Windows CI is red on them**

`dotnet-data-provider.service.ts` builds both values with `path.join`, while the test compares them against forward-slash literals interpolated from `RESOURCES_PATH`. On Windows `path.join('/repo', 'c-sharp')` is `\repo\c-sharp`, so the two can never match. `npm test` runs on all three OSes (`.github/workflows/test.yml:165`, no `if:`), so this fails in CI rather than only for someone developing on Windows.

It already has - the windows-latest job at `c91997f59`:

```
FAIL src/main/services/__tests__/dotnet-data-provider.service.test.ts >
  dotnetDataProvider.start > PT_DOTNET_NO_WATCH=true runs the prebuilt assembly, from the project directory

-   "/repo/c-sharp/ParanextDataProvider.csproj"
+   "\\repo\\c-sharp\\ParanextDataProvider.csproj"

Test Files  1 failed | 260 passed (261)
```

Line 83's `cwd` assertion has the same problem; it just is not reached. macOS and Linux were cancelled by fail-fast, so nothing else in the suite is verified at this head either.

**Suggested change** - build the expectations the way the source builds the values:

```ts
import path from 'path';

// ...
expect(args).toEqual([
  'run',
  '--project',
  path.join(RESOURCES_PATH, 'c-sharp', 'ParanextDataProvider.csproj'),
  '--no-build',
]);
// Matches the cwd `dotnet watch --project` would have used, so the two dev modes agree
expect(options).toEqual({ cwd: path.join(RESOURCES_PATH, 'c-sharp') });
```

I checked this rather than only proposing it: reproduced the Windows failure locally by mocking `path` to `path.win32` in a copy of the test, then confirmed the change passes both under that simulation and normally, and that the case stays falsifiable - removing `options = { cwd: cSharpDir }` from the service still fails it. Prettier and eslint clean.

One thing it gives up: with both sides going through `path.join`, the assertion no longer pins the literal path shape, so a `join` -> `resolve` swap in the source would stop being caught. If you would rather keep the literals, normalizing the received values (`args.map((arg) => arg.replace(/\\/g, '/'))`) does the same job.

mattgetgen: Taken as written, in `2712b674ede` — both `--project` and `cwd` now go through `path.join` with the same segments the service uses.

Reproduced your win32 simulation before and after, to make sure the fix is the fix and not just a green suite locally: mocking `path` to `path.win32` for the whole module graph (so the service and the expectations agree on separators), the old literals fail exactly the one case CI failed, and the `path.join` form passes both under the simulation and normally. Falsifiability holds — deleting `options = { cwd: cSharpDir }` from the service still turns that case red. Full `npm test` is green on Linux (exit 0), so the macOS/Linux coverage that fail-fast cancelled at `c91997f59` is accounted for as far as this machine can speak for it.

On the trade-off you named: took the `path.join` form rather than normalizing the received values. Pinning the literal path shape would be worth something if the shape were load-bearing, but it is not here — `dotnet run --project` takes any absolute path, so a `join` -> `resolve` swap in the source would be invisible to the tool as well as to the test. What the case is actually for is that the no-watch branch passes an *absolute* project path and sets `cwd` to the project directory, and both sides of that still hold.

lyonsil (approved):

PR #2705, opened 2026-08-21
PT-4398: Make the toolbar sync status correct from app startup

jolierabideau:
# Code Review Summary

**Branch**: pt-4398-sync-activity-signal

**Base**: origin/main

**Date**: 2026-08-21

**Review model**: Claude Opus 5

**Files changed**: 14 at review start; 24 after in-review fixes

## Overview

The toolbar's Sync control is rebuilt so its status can be trusted. A new `SyncStatusButton`
component and `useSyncStatus` hook replace the inline toolbar code: the status is seeded from a
send/receive snapshot (`getSyncState`) at mount, kept live by the `onSyncStateChanged` claim event,
and — new in this branch — unioned with a backend sync-activity signal (`getSyncActivity` /
`onSyncActivityChanged`) that covers the two startup paths the claim event cannot see. A popover
names the projects currently syncing and offers a single-shot cancel. Five states are modelled
explicitly, including `unknown`, which is deliberately distinct from `idle` because "nothing has
synced" is a positive claim that must not be made on the strength of a failed read.

The branch also reconciles this control with the app's other sync surfaces — the scripture editor's
sync-blocked banner and the first-run wizard's sync-progress step — via comments at both call sites
and two new ADR entries recording why the surfaces stay separate and what a shared cancel-requested
state would take.

Nine Important and sixteen Minor findings were raised in the pre-review pass, and three rounds of
review feedback followed on the PR itself; the responses in the comment thread record what was fixed
and what was deliberately not. The result is verified green: typecheck clean, 0 lint errors
repo-wide, and the full test suite passing (1,957 tests across 150 files as of 2026-08-25).

## API Changes

- `lib/platform-bible-react/`, `lib/papi-dts/papi.d.ts`: no changes
- `lib/platform-bible-utils/`: **added** `isolateBidi` (exported from the package's public surface,
  moved here from the component during review as a shared string utility). `papi.d.ts` did not need
  regenerating — it carries no `platform-bible-utils` string utils.
- `src/@types/paratext-bible-send-receive/index.d.ts` (core's copy of the closed-source Send/Receive
  type seam, ambient in core + bundled extensions):
  - Added type `SyncState` (`isSyncing`, `lastResults?`, `lastRequestedProjectIds`,
    `syncingProjectIds?`)
  - Added type `SyncActivitySnapshot` (`isSyncing`, `projectIds`)
  - Added command declaration `paratextBibleSendReceive.getSyncState: () => Promise<SyncState>`
[...]

PR #2764, opened 2026-09-02
PT-4240: Never show the toolbar Sync button in power mode

katherinejensen00:
# Code Review Summary

**Branch**: pt-4240-sync-button-shows-in-power

**Base**: origin/main

**Date**: 2026-08-31

**Files changed**: 5

## Overview

PT-4240 reports that the toolbar's Sync button appears in Power mode, where it should never show.
The `!isPowerMode` gate was already correct as far as it went — the real defect was one level down,
in how the interface mode is read. `useInterfaceMode` returns `'simple'` both when the mode really
is simple and when it is not yet known, so on a start with no cached mode the placeholder reads as
"simple" and every simple-only control renders in the Power toolbar until the setting resolves. A
failed setting read made it worse: the mode fell back to `'simple'` even when the cache knew the
user was in Power mode, and that placeholder was then written back to the cache, so the wrong mode
looked authoritative on the next start too.

The fix adds a third return value, `isModeKnown`, to `useInterfaceMode`; falls back to the cached
mode before `'simple'` when a read fails; and never caches a value the read did not actually
produce. The toolbar now takes one subscription and derives both `isPowerMode` and `isSimpleMode`
from it, gating both simple-only controls (Sync and the project picker) on the stricter
"known-simple" test. Layout defaults that need *some* value on the first render — toolbar height,
the book/chapter trigger variant — deliberately keep `!isPowerMode`.

## API Changes

None. No public API surfaces changed — the diff touches only `src/renderer/` (renderer-internal
hooks and the toolbar component). No changes to `lib/platform-bible-react/`,
`lib/platform-bible-utils/`, `lib/papi-dts/papi.d.ts`, or any `extensions/src/*/src/types/*.d.ts`.

Renderer-internal (non-public) surface changes:

- `use-interface-mode.hook.ts`: `useInterfaceMode()` return tuple grew from `[mode, setMode]` to
  `[mode, setMode, isModeKnown]` (additive; the only other consumer,
  `user-profile-popover.component.tsx:103`, destructures two elements and is unaffected)
- `use-interface-mode.hook.ts`: behavior change — a failed read now falls back to the cached mode
  before `'simple'`, and the `localStorage` cache is no longer written for a value the read did not
  produce
- `use-is-power-mode.hook.ts`: TSDoc only — documents that negating it is wrong for simple-only UI

None.

- [x] **`platform-bible-toolbar.tsx`: the simple-mode project picker still gated on `!isPowerMode`,
  so it flashed in the Power toolbar at startup for exactly the reason Sync did — leaving the file
  contradicting the rule the change itself introduces** _(fixed during review: gated on
  `isSimpleMode`, with a regression test that carries its own positive control)_
- [x] **`use-interface-mode.hook.ts`: `isModeKnown` never became `true` after a settled read failure
  with an empty cache, hiding Sync for the whole session and — since nothing would ever be cached to
  break the cycle — every session after it. Fail-closed, sitting next to a gate documented as
  fail-open** _(fixed during review: a settled failure now counts as known and assumes `'simple'`;
  the assumed value is still never cached)_

[Author response: both accepted and fixed during the interview. On the picker, the author chose to
widen scope past PT-4240's Sync-only DoD rather than ship a half-fixed toolbar. On the read failure,
the author chose fail-open — a simple user with a broken settings read keeps Sync, and a Power user
in that state is already seeing simple-mode UI everywhere.]

### Minor — Consider

- [x] **`useSendReceiveAvailability({ enabled: !isPowerMode })` no longer matched its comment now
  that the render gate is `isSimpleMode`** _(fixed during review: kept the broader `!isPowerMode`
  gate deliberately — it is a prefetch, so starting during the unknown window means the answer is
  ready the moment the mode resolves — and rewrote the comment to say so, including the cost of one
  wasted probe for an unresolved Power user)_
- [x] **Two separate `useSetting('platform.interfaceMode')` subscriptions in one component
  (`useIsPowerMode()` + `useIsSimpleMode()`)** _(fixed during review: the toolbar now calls
  `useInterfaceMode()` once and derives both booleans)_
- [x] **`use-is-power-mode.hook.ts` gave no hint that `!useIsPowerMode()` is wrong for simple-only
  UI — the file someone opens right before breaking the rule** _(fixed during review: TSDoc now
  states the rule, points at `isModeKnown`, and carves out the layout-default exception)_
- [x] **New toolbar comment mixed "send/receive" and "Sync"** _(fixed during review: uses Sync as
  the user-facing term per the Product Names guide; "send/receive extension" kept only where it
  names the extension)_
- [x] **Both new tests anchored `waitFor` on `toolbar-reserved-space-wrapper`, the component's own
  always-present root div, so they waited for nothing** _(fixed during review: anchored on the
  marketing version, which arrives from an async command, and each absence assertion now has a
  positive control that flips the mode to resolved and asserts the control appears)_
- [x] **The mocked `useSetting` ignored `defaultState`, leaving the cache-seeding path — the reason
  the hook exists — unasserted** _(fixed during review: added a spy asserting the seed value, plus
  the read-once behavior on re-render and the empty-cache `'simple'` seed)_
- [x] **No test covered the transition itself, only the static states** _(fixed during review:
  added unknown → resolves-to-simple for both Sync and the project picker)_
- [x] **`isModeKnown` was documented as "known" for an unverified cached value** _(fixed during
  review: TSDoc now says it means "a value worth acting on", not "verified")_
- [x] **`localStorage` mode cache leaked between toolbar tests, making them order-dependent**
  _(fixed during review: cleared in the file-level `beforeEach`)_
- [ ] ~~`useIsSimpleMode` tests live in `use-interface-mode.hook.test.ts` rather than a colocated
  file~~ _(overtaken by events: collapsing to one subscription left the hook with zero callers, so
  it was removed and its guidance folded into `useInterfaceMode`'s TSDoc and `useIsPowerMode`'s. The
  tests remain, rewritten to assert the `isModeKnown && mode === 'simple'` rule directly.)_
- [ ] **A settled `PlatformError` from the settings read is still swallowed silently** — the
  extension's own `useIsPowerMode` logs a warning on this same setting. Left open deliberately: the
  author chose the fail-open option over the bundled log-and-hide option, and adding logging was not
  part of that choice.
- [ ] **Simple-mode users on a first-ever start now see Sync appear a beat late** rather than
  immediately — the trade this fix makes. Nothing shifts beside it (the config area is
  `justify-end`, so the version badge and profile popover keep their positions), but the
  Responsiveness guide's "reserve a fixed slot" advice suggests confirming with UX that a late
  appearance beats a reserved-width placeholder.

[Author response: the author accepted every actionable minor except splitting the test file, which
was overtaken by the hook's removal. The two remaining open items are a deliberate scope decision
and a UX question for the reviewer.]

#### Shared Regions Modified

None. All five changed files were grepped for `region shared with` markers — none present.

#### Extension Config Changes

## Positive Observations

- The fix targets the root cause rather than the symptom: the `!isPowerMode` gate was already
  correct, and patching only the Sync render site would have left every other simple-only control
  with the same trap.
- Not caching the unresolved `'simple'` placeholder is a real fix beyond the button — it stops a
  first-run guess from looking authoritative to the first-run gate on the next startup. Verified
  safe at `first-run-store.ts:78,183`, which only branches on a cached `'power'`.
- The failed-read fallback now prefers the cached mode over `'simple'`, fixing a second-order bug
  (a transient read error demoting a known Power user into simple-mode UI) that was not strictly
  required by the stated purpose but clearly belongs with it.
- Both new toolbar tests were verified falsifiable by reverting the gate and confirming RED, and
  each now carries its own positive control, so neither is a negative assertion with nothing to be
  a negative of.
- `useState(readCachedInterfaceMode)` uses the lazy initializer form correctly, and the comment
  explains the real reason (the seed React started from must not drift from later re-reads) rather
  than restating the code.
- Comments record the UX decision behind the Power-mode exclusion instead of just asserting the
  rule, and the "don't negate this" rule is documented on the hook someone would actually open
  before breaking it.
- ESLint is clean on all changed files with no new suppressions.

## Interview Notes

**Stated purpose**: Fix the toolbar Sync button showing in Power mode (PT-4240).

**Context the reviewer should know.** PT-4240 was filed 2026-07-24, a month before PR #2683
(PT-3954) added the `!isPowerMode` gate on 2026-08-19. The originally reported symptom — Sync
persistently visible in Power — is therefore already fixed on `main`. What remained, and what this
PR addresses, is the startup window plus the read-failure path, both of which still violate the
ticket's Definition of Done ("When starting Power, Sync button on the app toolbar does not show").

**Decisions made during the interview:**

- *Scope.* Three of the four analysis passes independently flagged the project picker as having the
  identical defect. The author chose to fix it in this PR rather than leave the toolbar half-fixed,
  accepting that this widens scope past PT-4240's Sync-only DoD.
- *Fail-open vs fail-closed.* On a settled read failure with no cache, the author chose to assume
  `'simple'` and show Sync, matching the fail-open policy the adjacent availability gate documents.
  The assumed value is still never written to the cache, so the assumption cannot harden into a
  persisted answer.
- *One subscription.* Collapsing `useIsPowerMode()` + `useIsSimpleMode()` into a single
  `useInterfaceMode()` call left `useIsSimpleMode` with zero production callers. Rather than keep
  scaffolding the repo's own "validate a reference pattern has production consumers" rule flags, the
  hook was removed and its guidance folded into `useInterfaceMode`'s TSDoc and `useIsPowerMode`'s.

**Unresolved items**: none. No Critical or Important finding was left open.

## In-Review Quality Check

- **ESLint** — clean on all changed files, no new suppressions, no autofixes needed beyond two
  prettier warnings corrected during authoring.
- **Prettier** — all changed files conform.
- **Tests** — `platform-bible-toolbar.test.tsx` (48) and `use-interface-mode.hook.test.ts` (13)
  pass: 59 → 61 assertions across the two files after the review fixes. Full `src/renderer` run:
  **4 files / 61 tests failing, identical to the pristine `origin/main` baseline measured in this
  same worktree**; all are pre-existing `overlay-*` failures unrelated to this branch. This branch
  adds 16 passing tests and no failures.
- **Typecheck** — cannot be run clean in this worktree in either yalc state, and this is a
  pre-existing environment condition, not a property of this branch. Linked, `tsc` reports "Cannot
  find module '@eten-tech-foundation/platform-editor'" (the yalc tarball ships `src/` with no
  `dist/`); unlinked, it reports one error in the build artifact
  `lib/platform-bible-react/dist/index.d.ts` (registry version lacks `MarkerMenuItem`). **No
  typecheck error names any file in this diff in either state.** The worktree was left in its
  original linked state.

## Suggested Review Focus

- [ ] **Scope call — is fixing the project picker here right?** PT-4240's DoD names only the Sync
  button. The picker has the identical defect in the same toolbar at the same startup moment, and
  leaving it would make the new file contradict its own documented rule — but it is a scope
  widening that should be a conscious call, not an accident.
- [ ] **Fail-open on a settled read failure.** With no cached mode and a failing
  `platform.interfaceMode` read, the app now assumes simple and shows Sync. Confirm that is
  preferable to hiding it; the alternative loses Sync for a genuine simple user whose settings are
  broken.
- [ ] **Removing `useIsSimpleMode`.** The rule now lives in TSDoc on two hooks rather than in a
  callable hook. If a second simple-only consumer is expected soon, reintroducing the hook may be
  better than documenting the rule twice.
- [ ] **UX: Sync now appears a beat late on a first-ever simple-mode start.** Confirm a late
  appearance is preferred over reserving a fixed slot for it.
- [ ] **Silent swallow of the settings read error** — deliberately left open; decide whether a
  `logger.warn` belongs here, as the extension's own `useIsPowerMode` does for the same setting.

AI-assisted — [session](https://claude.ai/code/session_013LCepQPpDnNZioW6vnbyGd)

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2764)


mattgetgen (approved):
Great work! Claude didn't find much but here are it's findings. They're maybe worth addressing but nothing that's a blocker IMO.

Scope note

pt-4240-sync-button-shows-in-power is stacked on PT-4398 (#2705), which isn't in main. So the review covered only the two PT-4240 commits (5 files, +263/−27) and treated the 50-file PT-4398 base as already-reviewed under its own PR. Tests on the branch: 18 files / 173 tests green.

The fix itself is sound — the new gate does close the reported defect. All three findings cluster on one root cause: cachedMode is frozen by useState(readCachedInterfaceMode) at mount and never updated with what the session later resolves.

Findings

1. Stale-cache fallback can reproduce the exact PT-4240 defect — medium
src/renderer/hooks/use-interface-mode.hook.ts:82

const mode = didReadFail ? (cachedMode ?? 'simple') : modePossiblyError;

App starts simple (or first launch, empty cache) → user switches to Power via the profile popover (the toolbar isn't remounted by a mode toggle, and the effect does write 'power' to localStorage) → the platform.interfaceMode subscription then delivers a PlatformError (ext host busy/restarting). The hook reports mode: 'simple', isModeKnown: true — so the power toolbar gains the Sync button and project-picker Select and loses Home. Verified empirically. The same snapshot also makes two components disagree: one mounted before the resolve says 'simple', one mounted after says 'power'.

Fix: track the last resolved mode in a ref (updated in the existing effect) and prefer it over cachedMode. Fixes both symptoms.

2. isModeKnown isn't monotonic — low
src/renderer/hooks/use-interface-mode.hook.ts:89

!isLoading || cachedMode !== undefined flips back to false whenever isLoading returns to true after a resolve, if the mount-time cache was empty. Concrete path: useData's runaway guard returns [runawayError, undefined, true] — an error with isLoading: true — for the full 5s RUNAWAY_COOLDOWN_MS. During that window the simple-only Sync button and project picker unmount and reappear, where the pre-change !isPowerMode kept them mounted.

Worth noting: the new test "assumes 'simple' and counts it as known when the read fails with an empty cache" can't catch this, because its mock pairs the error with isLoading: false.

3. !isPowerMode still gates rendered content — low
src/renderer/components/platform-bible-toolbar.tsx:293

useOpenProjectBookIds(…, !isPowerMode) and additionalBookIds still key off the negation, while the PR's own new useIsPowerMode doc says negation is only acceptable for layout defaults. Unlike the availability probe at line 450, this feeds rendered content: a power user with no cached mode gets simple mode's widened book list and the "show more books" affordance until the setting resolves, plus the booksPresent subscriptions the comment at line 285 says power mode should open none of. Transient, but it's the same placeholder-mode class the PR is closing. isSimpleMode for the additionalBookIds value would be consistent (keep !isPowerMode for the prefetch if you want that tradeoff).

> @mattgetgen reviewed 5 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2764#-:-P0bh16o9nIR6LnRjCQt:bep35x0)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

katherinejensen00:
Good findings, Matt! Thanks for the review!

1. Stale-cache fallback could reproduce PT-4240 (medium) — use-interface-mode.hook.ts. Added lastKnownModeRef, seeded from the cache and advanced in the existing effect on every successful resolve; the read-failure fallback now prefers it. Fixes both symptoms Matt named: the Power toolbar gaining Sync/picker and losing Home after a mode switch plus a PlatformError, and two components disagreeing across the resolve.

2. isModeKnown isn't monotonic (low) — same ref makes the flag one-way, so useData's runaway guard (error paired with isLoading: true for the 5s cooldown) no longer unmounts and remounts the simple-only controls.

I used a ref rather than state — Matt's suggestion, and it turned out to matter: a state version broke three existing project selector label tests, which use mockReturnValueOnce and can't absorb the extra mount render. Noted the reasoning in a comment.

3. !isPowerMode still gating rendered content (low) — platform-bible-toolbar.tsx:293. Took Matt's parenthetical: additionalBookIds now keys off isSimpleMode, while useOpenProjectBookIds keeps !isPowerMode as a deliberate prefetch (same tradeoff the availability probe documents). The comment claiming power mode opens no booksPresent subscriptions was inaccurate during the unknown window, so it now states the real cost.

Issue #222, opened 2023-06-07
Add WebView state


tjcouch-sil:

tjcouch-sil:
One example of webview state would be a list of open texts on a text collection webview

irahopkinson:
What is the advantage of implementing and using this over local storage or session storage? I guess if the state needs to be saved between devices for a user?

tjcouch-sil:
Honestly I'm not sure `localStorage` in webviews would persist. They might! Or they might not. If it did persist, I imagine we would need to figure out a way to pass the webview id down to the webview so it can use the id in its localstorage keys (to make sure the different webviews have their own storage).

However, I believe having state be a feature on the webview would allow state to be modified by the web view provider (and, if we so decide, maybe other things as well). The main things I think about here are specifying some state when first loading the web view (e.g. telling the webview which texts to load in its text collection) and making version updates to webview state without forcing the webview itself to support old versions of its state (the provider could transform the data on load to make the code simpler).

tjcouch-sil:
"We also want the text in the title bar of the webview to be updated dynamically. Right now I this is just static text." split off to #445

tjcouch-sil:
Created simple implementation issue #472. This design is now for expanded functionality if we want it. Like updating the state from the webview provider anytime. Not sure we need that, but this is here.

tjcouch-sil:
As a follow-up to #472 (PR #545), it may be good to rework webview state to be attached directly to the `SavedWebViewDefinition`. That way, it travels across processes to the web view provider very naturally, and it gets removed immediately when a webview is closed (no need to run a cleanup function).

lyonsil:
> However, I believe this may require creating a synchronous-access dockLayout variable in the web view service so you can get the dock layout to get the web view state immediately. Should be fine since anytime a web view is accessing its state, the dock layout should exist. But maybe you can get away with doing this without web view service having to touch it.

This is the reason I didn't do it this way at least for now.  There was a requirement for this API to be synchronous, and the dock layout is stored as a promise with the comment:
```
Also please do not save this variable out anywhere because it can change, invalidating the old one (see `registerDockLayout`)
```

We're also a little loose in how we handle some async calls (i.e., we don't always await the results) in this code.  For example:
```
  // Will we ever need to await this? For now, seems like it unnecessarily complicates registering
  // because making this function async would probably be annoying in React
  loadLayout();
```

[...]

Issue #30, opened 2023-02-15
Load and run extension file (hello world)


tjcouch-sil:
Load up extension file in the extension host and let it run papi stuff

Ideas:
- Need file service
- In debug mode, watch unzipped extension files and reload as appropriate
- Webpack/babel

tjcouch-sil:
Extension requirements:
- Not "type": "module"

tjcouch-sil:
Future possibilities for shimming require:
https://github.com/elastic/require-in-the-middle
https://levelup.gitconnected.com/how-to-add-hooks-to-node-js-require-function-dee7acd12698
https://github.com/danez/pirates

irahopkinson:
Pretty sure `electron` has already shimmed `require` so remember to take that into account.

tjcouch-sil:
Can't figure out how to catch iframe exceptions. Very strange because somehow Electron is showing an exception screen on these exceptions, so _something_ is catching these exceptions. I dunno what layer of ERB or what does that.

Minimal iFrame error catching test that demonstrates the issue:
[iframe-error-test.zip](https://github.com/paranext/paranext-core/files/10816376/iframe-error-test.zip)

Some related stuff:
https://stackoverflow.com/questions/3102819/disable-same-origin-policy-in-chrome
https://stackoverflow.com/questions/16194398/inject-a-javascript-function-into-an-iframe

tjcouch-sil:
https://github.com/paranext/paranext-core/tree/30-load-extensions
Waiting for https://github.com/paranext/paranext-core/pull/29 and https://github.com/paranext/paranext-core/pull/43

PR #2767, opened 2026-09-03
PT-4701: Make the whole tab bar a drop target for dragged tabs

rolfheij-sil:
https://github.com/user-attachments/assets/279bb12f-2f5c-4237-aba8-bb91ee0c00d9

## Summary

In Power mode, a dragged tab or floating tab group could only be re-docked by hitting a tab. The rest
of a tab group's bar had no drop handler, so dropping there silently did nothing.

Now the whole empty part of a Power-mode tab bar is a drop target. Dragging a tab or tab group over
it shows a drop indicator, and dropping appends it as the group's last, active tab.

This is not part of the current epic. It came from a usability-interview finding and is fixed as a
standalone defect.

## Tickets

- **PT-3288** ("Defect: rc-dock drag to un-float fails"): this PR does not close it. Its Definition of
  Done is the tab that vanishes when you undock a floating tab, which this PR doesn't touch. The
  blank-out crash it quotes was PT-3411, already fixed. This PR fixes the other symptom it reports:
  the only place to drop a tab is a tiny area on the tab headers.
- **PT-3330** ("rc-dock dropzone: instead of floating, add to tab group"): this PR advances its user
  story but does not meet its Definition of Done. That asks for a drop in the middle of a panel to
  add the tab to the front of the group, and for a smaller tab header while dragging.
- **PT-4674** ("Move a tab to another tab group without a mouse"): raised in review. Moving a tab
  between groups is pointer-only, here and everywhere else in the app, so the gap predates this PR
  and is tracked rather than closed here.

## How it works

- **`TabBarDropZone`** (`src/renderer/components/docking/tab-bar-drop-zone.*`) renders through
  rc-dock's `panelExtra` beside the "+" button and fills the bar's remainder. It is an ordinary
  rc-dock `DragDropDiv` that mirrors rc-dock's own acceptance rules and appends with
  `dockMove(source, panel, 'middle')`. It rejects a drop that would do nothing (a group's own last
  tab, or a group onto itself).
- **"+" slides to the bar's end** during a tab or group drag, so the drop area is one continuous
  strip. Divider drags don't move it.
- **Hit extensions don't touch layout.** During a drag, absolutely positioned pseudo-elements extend
  the zone backward over the last tab's trailing half, and forward over the gap, "+", the bar's
  trailing padding and the bar's full height. The zone's own box never changes size, so the tab row
  can't resize or re-scroll mid-drag.
- **The drop indicator is always wide enough to see.** On a nearly full bar the zone's own width can
  be a few pixels, which would draw a hairline nobody notices. The indicator is therefore led
  backward into the last tab's trailing half — the area the zone already claims — up to the 30px
  rc-dock itself draws for a drop after the last tab, and never further back than the claimed area,
  so it never promises a target that isn't there. On a tab too narrow for even that, the indicator
  is as wide as the claimed half allows.
- **The zone accepts a drop exactly where it claims one.** Claiming the last tab's trailing half and
  accepting a drop there are the same condition, evaluated at drag start and at drop time. This
  matters because rc-dock hit-tests the topmost element and does not fall through to the tab
  underneath: a region the zone covered but refused would silently swallow a drop, so that state is
  made unrepresentable rather than merely avoided.
- **rc-dock's edge-split layer starts below the bar.** rc-dock places it 30px from the panel top, but
  this app's bar is 42px tall, so a drop on the bar's bottom strip split the layout. This also
  applies to drops on rc-dock's own tabs.
- **rc-tabs' idle overflow box is removed from layout** in both Power and Simple mode. It kept
  `flex-grow: 1` and took about half of the bar's free space. rc-tabs never measures it. In Simple
  mode, the Resources tabs still collapse to icons at the same width as before. In Power mode, a
  crowded bar now has 28px more room for tabs at rest than on `main` (measured in headless
  Chromium), even though the gap between "+" and the invisible zone costs 8px of that.
- **The "+" and overflow buttons are exactly 30px wide**, as declared. The shared button style's 16px
  inline padding made "+" 34px. Its block padding is dropped as well: both buttons are fixed-size,
  icon-only and `border-box`, so it never reached their boxes — measured at 30×30 in both interface
  modes in the running app. The gap, button size and bar padding come from one SCSS partial,
  `_tab-bar-metrics.scss`, and the tab height is set from the same button-size token, since a tab and
  these buttons render at the same height.

No public PAPI surface changes, so no `@experimental` markers are needed. No keyboard handlers were
added or changed.

The design and its rejected alternatives are recorded as `adr-tab-bar-drop-zone-app-side-target` in
`.context/standards/Architecture-Decisions.md`.

## rc-dock patch and version

- **`isPopupDiv` fix.** Reordering a tab in a group whose tab bar overflows drew the drop indicator
  in the app's top-left corner. This bug predates this PR and reproduces on `main`. rc-dock decides
  whether a tab ref belongs to the real strip or to the hidden overflow dropdown by checking exactly
  two ancestor levels for an `<li>`. The hit area sits one level deeper, so the dropdown's zero-size
  copy replaced the real one. The existing `patches/rc-dock+3.3.2.patch` now walks up to the nearest
  `<li>`, stopping at `.dock-nav-list`. Upstream, ticlo/rc-dock#161 reports the same mismatch. Only
  its crash symptom was fixed there, so upgrading doesn't help.
- **Unit test for the patch.** The patch edits only rc-dock's `es/` build, which is the one webpack
  ships. Vitest loads `lib/`, so `rc-dock-tab-cache-patch.test.ts` imports `rc-dock/es/DockTabs`
  directly.
- **Pinned version.** rc-dock is pinned to exactly `3.3.2`, since the patch only applies to that
  version. `src/renderer/components/docking/README.md` lists the rc-dock and rc-tabs internals this
  feature relies on, as a checklist for upgrades.

## Known limits

- **Crowded bars.** On a bar crowded enough to overflow, the last tab's trailing edge sits past the
  zone's start, so the zone claims nothing and refuses drops rather than draw an indicator over a
  tab it isn't covering. Dropping past the bar's last tab still appends, through rc-dock's own
  target, and the e2e spec releases such a drop and asserts it. What stays dead there is the strip
  the gap, "+" and the bar's trailing padding occupy: the forward hit extension covers it, but with
  nothing to draw an indicator in the zone refuses, so a drop there does nothing.
- **RTL.** The right-to-left geometry is covered by unit tests, plus a manual check in the running
  app with the dock's direction flipped: "+" sits at the visual end, tabs are right-aligned and
  nothing overlaps. On a bar that was already overflowing before the flip the check is
  inconclusive — flipping direction after layout leaves rc-dock's own pixel scroll offset pointing
  the wrong way — so a genuine RTL locale is what would settle that case. The e2e spec runs
  left-to-right.
- **Coupling to rc-dock.** The feature depends on rc-dock internals: DOM classes, the order of
  drag-listener calls, and `setDropRect` behavior. The version pin and the README checklist make an
  upgrade a deliberate step.

## Notes for reviewers

- **Re-run `npm install` after pulling.** The rc-dock patch changed, and `rc-dock-tab-cache-patch.test.ts`
  fails against a `node_modules` that still has the old patch applied.
- **UX.** The "+" slide during a drag is shown in the video above.
- **No Storybook story.** The zone is invisible and only does anything during a live rc-dock drag, so
  a story would show nothing; the e2e spec covers it instead.

## Verification

- **Unit tests.** Docking unit tests: 359 pass. They include:
  - lead and overlap geometry (LTR and RTL);
  - the indicator widening, its cap at the claimed area, and the refusal when even that leaves no
    visible width;
  - whole-panel drags;
  - two independent bars;
  - indicator-token stability;
  - the patched `TabCache`.
- **Core suite.** `test:core`: 5167 tests pass, 15 skipped. `platform-bible-react`: 1632 tests pass.
- **E2E.** `npm run test:e2e:isolated tests/isolated/docking/` passes 7/7 (four here, three from the
  middle-click PR that merged before this one):
  - drop on the bar remainder;
  - a drop on the last tab's trailing half, which checks the indicator's shape before releasing, so
    it fails if the zone stops claiming that half rather than passing on rc-dock's own target;
  - drops on another panel's gap, "+", trailing padding and lower band;
  - crowded bar: the drag start leaves the tab row alone, the gap draws no indicator, rc-dock's own
    indicator stays inside the bar, and a release past the bar's last tab still appends.

  The bar-space drops failed before the layout fixes, and the first test fails if the zone's
  `flex-grow: 1` is removed. Removing the backward hit extension turns the trailing-half test red on
  its indicator-shape assertions.
- **Static checks.** `build:types` (no `papi.d.ts` change), `build:main`, `typecheck`, `lint` and
  `format:check` all pass.
- **Live measurements in the running app.**
  - On a crowded bar, starting a drag used to shift the tab row by 42px and draw the indicator past
    "+". The e2e crowded-bar test now covers this.
  - "+" and the trailing padding used to be dead, and the bottom 12px of the bar used to split the
    layout. The e2e drop tests now cover both.
  - The Simple-mode icon-collapse width is unchanged from `main`.
  - The overflow "more" button renders 30×30 in both Power and Simple mode, so dropping the shared
    button style's block padding changes nothing on screen.
  - The tab bar is 42px: the 30px button row plus the header-to-content gap above and below it.
- **Manual testing.** The product owner tested and approved the original version of the feature by
  hand.

## Prior art

Upstream rc-dock tried the same idea in ticlo/rc-dock#222 ("TabBar drop"). It was closed unmerged on
2025-04-30. Treating the tab bar's empty remainder as its own append-at-end drop target, rather than
stretching the last tab's hit area, matches how dockview, FlexLayout, Golden Layout, react-mosaic
and VS Code's tab strip handle this.

AI-assisted — [session 1](https://claude.ai/code/session_01MRwNSFZupZcGz8qLtdFQ9G), [session 2](https://claude.ai/code/session_018CaYNvoByux4DtDVZJ6NK1)

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2767)


katherinejensen00 (changes requested):
# PR #2767 review — Make the whole tab bar a drop target for dragged tabs

- **PR:** https://github.com/paranext/paranext-core/pull/2767
- **Author:** Rolf Heij · **Branch:** `pt-3288-tab-bar-drop-zone`
- **Head reviewed:** `8c0554f8bcc` · **Merge base:** `30f07f460b1`
- **Reviewed:** 2026-09-15 · Read-only; nothing edited, committed, or posted.

## Outcome

30 issues, no blockers. 5 Major, 15 Minor, 10 Nit. Paths are under
`src/renderer/components/docking/` unless shown otherwise. "Line" is the first changed line to
attach a comment to.

The design itself held up under review. What did not hold up is the crowded-tab-bar case: three
reviewers reached it independently, and the geometry the code assumes there is wrong in a state
users hit routinely.

## Caveat — what this review could NOT verify

**Two of the seven planned passes never ran.** Both died on an Anthropic weekly rate limit
(HTTP 429, resets 2026-09-16 15:00 America/Los_Angeles), not on anything in the code:

- **The antagonistic pass** — the one that would have attacked all 30 findings for false positives,
  overstated severities, and pre-existing issues wrongly blamed on this PR. It failed ~1 minute in.
- **`/code-review max`** — its coordinator and 9 subagents all failed. No report exists from it.

**What that means for the findings below.** Each one was verified by its own author against
rc-dock 3.3.2 / rc-tabs 11.16.1 source, but **no finding here has survived an adversarial
second opinion.** Treat severities as first-round estimates. Specifically:

| Unverified area | Why it is uncertain |
|---|---|
| **#1, #2 (crowded bar)** | Confirmed by source reading plus a synthetic Electron 39 repro of the flex/margin rules on a copied tab-bar DOM — **not reproduced in the running app.** The repro's fidelity to real rc-dock CSS was never independently checked. |
| **#6 (Simple mode)** | The selector match is confirmed by source; the **visual size of the change was never measured**, and the Resources column was never opened. |
| **#8 (stale panel identity)** | Marked "Plausible" by its author. The object replacement and the identity check are confirmed in source; **the mid-drag update was never reproduced**, and it is unconfirmed whether `dockMove` would actually duplicate rather than no-op. |
| **#7 ("+" dead spot)** | Mechanism confirmed in source; **how often users aim there is a judgment call**, and whether this is in scope for this PR was never challenged. |
| **#3, #4, #5 (Major test/coupling gaps)** | The gaps are real and confirmed. Whether they justify **Major** — versus being normal for a renderer-local UX fix — is exactly what the antagonistic pass would have decided. |
| **#30 (drop-edge band)** | Derived from rc-dock's `.less` source; **never measured live**, and the dist CSS was not read. |
| **Severity calibration overall** | No pass ever asked "is this Major too high, or this Nit too low?" across the set. |
| **Missed issues** | Nothing looked for problems **all six reviewers overlooked**. Absence of a finding here is not evidence of absence. |

**One finding was already corrected by hand.** The architecture reviewer claimed the PR body never
describes the `isPopupDiv` patch. That is false — the body describes it under "Also fixed here:
reorder indicator on a crowded tab bar". That sub-claim was dropped; the rest of the finding stands.

**Before approving**, items **1, 2, 6 and 8 should be confirmed in the running app.** They rest on
source reading plus one synthetic repro rather than the real thing.

### How the review was produced

Six parallel read-only passes, each verifying claims against `node_modules/rc-dock/src` and
`node_modules/rc-tabs`: soundness/correctness, test quality, clarity, architecture, comments/docs,
and UX/performance/accessibility. The soundness pass built an Electron layout repro; the test pass
ran mutation checks in a scratchpad copy.

## Issue index

| # | Issue | File:line | Severity |
|---|---|---|---|
| 1 | Crowded-bar drag shifts the tab strip | `tab-bar-drop-zone.component.tsx:85` | Major |
| 2 | Drop indicator drawn past "+" on an overflowing bar | `tab-bar-drop-zone.component.tsx:86` | Major |
| 3 | The layout behavior has no automated test | `dock-layout-wrapper.component.scss:231` | Major |
| 4 | rc-dock patch untested; vitest loads the unpatched build | `patches/rc-dock+3.3.2.patch:47` | Major |
| 5 | No guard if an rc-dock upgrade breaks the internals used | `tab-bar-drop-zone.component.tsx:1` | Major |
| 6 | Simple mode is affected, contrary to the PR body | `dock-layout-wrapper.component.scss:269` | Minor |
| 7 | "+" is still a dead drop spot at the bar's end | `dock-layout-wrapper.component.scss:247` | Minor |
| 8 | Self-drop guard compares objects, not ids | `tab-bar-drop-zone.util.ts:56` | Minor |
| 9 | Two different drag signals with an ordering dependency | `dock-layout-wrapper.component.scss:241` | Minor |
| 10 | Listener's drag test differs from the resolver's | `tab-bar-drop-zone.component.tsx:148` | Minor |
| 11 | Panel-drag widening never tested | `tab-bar-drop-zone.component.test.tsx:343` | Minor |
| 12 | Tests never render two bars | `tab-bar-drop-zone.component.test.tsx:113` | Minor |
| 13 | Token-stability tests can't detect a shared token | `tab-bar-drop-zone.component.test.tsx:189` | Minor |
| 14 | `panelExtra` child order not asserted | `platform-dock-layout-positioning.util.test.ts:67` | Minor |
| 15 | Group-config assumption nothing enforces | `tab-bar-drop-zone.util.ts:38` | Minor |
| 16 | TSDoc gives the wrong overflow reason | `tab-bar-drop-zone.component.tsx:101` | Minor |
| 17 | Same explanation written in three files | `tab-bar-drop-zone.component.scss:25` | Minor |
| 18 | Architecture decision not recorded | `platform-dock-layout-positioning.util.ts:68` | Minor |
| 19 | PR description is ~1,600 words and partly stale | PR body | Minor |
| 20 | Redundant double exclusion of the hidden box | `dock-layout-wrapper.component.scss:282` | Minor |
| 21 | Gap width copied between stylesheets | `tab-bar-drop-zone.component.scss:52` | Nit |
| 22 | Test names carry implementation detail; wrong filename | `tab-bar-drop-zone.component.test.tsx:254` | Nit |
| 23 | Fixture naming, location, unused export | `tab-bar-drop-zone-test.util.ts:19` | Nit |
| 24 | Reset-before-measure is undocumented | `tab-bar-drop-zone.component.tsx:72` | Nit |
| 25 | `Math.max(0, …)` guards an impossible negative | `tab-bar-drop-zone.component.tsx:86` | Nit |
| 26 | `data-dragging` literal repeated | `tab-bar-drop-zone.component.tsx:47` | Nit |
| 27 | `closest('li')` walks to the document root | `patches/rc-dock+3.3.2.patch:73` | Nit |
| 28 | Layout reads and writes interleave per panel | `tab-bar-drop-zone.component.tsx:164` | Nit |
| 29 | Test name cites a gate that doesn't exist | `tab-bar-drop-zone.util.test.ts:60` | Nit |
| 30 | z-index comment omits rc-dock's edge layer | `tab-bar-drop-zone.component.scss:39` | Nit |

## Comments for the PR owner

Plain text, ready to copy-paste. Each opens with the plain-language summary, then the detail.

### 1. `tab-bar-drop-zone.component.tsx:85` — Major

On a tab bar with more tabs than fit, starting a drag can make the whole tab row jump, and the merged drop target you documented isn't there. The comment just above says the last tab is clipped on an overflowing bar so the overlap clamps to 0, but that only holds when the last tab is scrolled out of view. rc-tabs' scrollToTab keeps the active tab visible, and Algorithm.addTabToPanel activates an appended tab, so a crowded bar very often has its last tab both last in DOM order and visible. The overlap is then non-zero, the negative margin shrinks .dock-extra-content, .dock-nav-wrap absorbs the freed width, and rc-tabs' ResizeObserver re-runs useVisibleRange. A repro of these rules in Electron 39 measured the strip widening from 228 to 268px, the zone ending up 2px wide, and a hidden tab popping out of the dropdown. This also fires for every mounted zone on any tab or panel drag in the dock, including moving a floating dialog. Suggested fix: extend backward with an absolutely positioned pseudo-element, the technique you already use for the 8px gap, so it can't feed back into flex sizing; or skip the claim when the bar is overflowing.

### 2. `tab-bar-drop-zone.component.tsx:86` — Major

On a crowded bar, the drop highlight appears in a different place from where the tab will actually land. indicatorInset is always tabRect.width / 2 and is never limited by the clamped overlap. When the zone has no width, the indicator rect is 0px wide and sits half a tab-width past the zone, beyond the "+" button, while the zone's ::after sliver still accepts the drop. The app's .dock-drop-indicator has a ~6px ring box-shadow, so even a zero-width rect paints visibly. Suggested fix: derive the inset from the clamped overlap, and reject the drop (or skip setDropRect) when the resulting indicator width would be zero.

### 3. `dock-layout-wrapper.component.scss:231` — Major

The part of this feature most likely to break silently is the part no test covers. Whether the zone is wide enough to hit is decided entirely in SCSS: flex-grow on .dock-extra-content, the hidden overflow box's display:none, the ::after gap cover, the body:has rule that moves "+", position:relative, and the indicator's pointer-events:none. Vitest does no layout and the component tests fake getBoundingClientRect, and e2e-tests has no tab drag spec. Deleting flex-grow:1 here makes the zone zero-width and all 194 tests stay green. Four of the ten commits on this branch fixed bugs in exactly this layer. Suggested fix: one Playwright drag test (drag a tab onto the bar remainder, and release over the gap next to "+"). If drag E2E is considered too flaky, please record that as a deliberate decision with a manual test plan.

### 4. `patches/rc-dock+3.3.2.patch:47` — Major

The isPopupDiv fix can vanish without any test failing, and it's already missing in at least one working tree. Vitest resolves rc-dock through main to lib/index.js, while patch-package only edits es/, so unit tests always run against unpatched rc-dock. This review worktree currently has the old two-level check in es/DockTabs.js and the docking suite still passes 194/194. patch-package's --error-on-fail only catches a hunk that no longer applies, not one that was dropped. Suggested fix: a small test importing TabCache from 'rc-dock/es/DockTabs' asserting a node three levels under an li is rejected and a node outside any li is kept, with a comment explaining the es import path.

### 5. `tab-bar-drop-zone.component.tsx:1` — Major

An rc-dock upgrade could break this feature while every test and the typecheck stay green. The implementation depends on about seven private DOM class names, the argument semantics of addDragStateListener, how the dragging layer's group classes are built, rc-dock's hit-testing walk, and the order of two statements inside createDraggingElement. The tests mock all of it, and package.json:150 allows any 3.x via ^3.3.2 even though both patches are pinned to 3.3.2. Suggested fix: pin rc-dock to exactly 3.3.2, and record the dependency list as an upgrade checklist in the docking README (the E2E test in item 3 would also cover this).

### 6. `dock-layout-wrapper.component.scss:269` — Minor

The PR says Simple mode is unaffected, but this rule reaches it. Both Simple-mode groups carry the platform-bible token (TAB_GROUP_RESOURCES = 'card platform-bible resources' and HEADLESS_GROUP = 'headless platform-bible' in dock-tab-group.util.ts), so the Resources column matches .dock-style-platform-bible. Previously the hidden operations box stayed in the flex flow, reserving its 30px "more" button; removing it hands that width back to the tab strip, which moves the point where Resources tabs collapse to icons or overflow. TAB_BAR_CHROME_OVERHEAD_PX in platform-tab-title.component.tsx was measured against the old layout. Fixing it for both modes seems reasonable — it just needs a look at the Resources column at narrow widths and a correction to the PR text.

### 7. `dock-layout-wrapper.component.scss:247` — Minor

Once "+" slides to the end, the last ~34px of the bar still swallows drops, which is where people aim when they mean "put it at the end". The hit-test walks from the button up through .dock-extra-content, .dock-nav and .dock-bar (DockTabBar's DragDropDiv registers no onDragOverT) to .dock-panel, which never accepts, so the indicator disappears and releasing does nothing. That's the same silent no-op this PR set out to remove, just relocated. Suggested fix: during a drag, widen the zone's ::after to cover the button and the bar's trailing padding. Making "+" pointer-events:none would not work, since the walk would then land on .dock-extra-content, which has no drop handler.

### 8. `tab-bar-drop-zone.util.ts:56` — Minor

The self-drop guard can miss if the layout changes mid-drag, which would let the duplicate-panel bug you're guarding against through. panel === targetPanel compares object identity, but rc-dock replaces PanelData objects on layout changes via Algorithm clone/replacePanel, and the app calls dockLayout.updateTab on web view updates while DragState still holds the panel captured at drag start. isLastTabOfPanel a few lines up already compares ids. Comparing panel.id === targetPanel.id costs nothing and removes the question. Flagged as plausible rather than reproduced.

### 9. `dock-layout-wrapper.component.scss:241` — Minor

Two separate mechanisms decide "a drag is happening", and the measurement quietly depends on which one runs first. "+" moves via this CSS rule keyed on the dragged group's class, while the zone widens via addDragStateListener keyed on the dock id. claimLastTabOverlap reads the zone's rect expecting "+" to have already moved, which only holds because rc-dock appends the dragging layer before it calls listeners. They can also disagree: with no .dock-tab present the zone returns early while "+" still moves. Suggested fix: drive "+" from the zone's own data-dragging marker, or document the ordering dependency at the listener.

### 10. `tab-bar-drop-zone.component.tsx:148` — Minor

The comment here contradicts itself, and the code uses a third definition of "relevant drag". It says the overlap is claimed for any drag "tab or otherwise", then says divider drags are excluded the same way resolveTabBarDropZoneSource excludes them. The listener only checks whether tab or panel data exists for this dock, while the resolver also checks group, tabLocked, own-last-tab, panelLock and self-panel. The visible effect today is that dragging a panel's own last tab widens the zone over that tab and then rejects the drop. Either gate the listener on the resolver, or say plainly why widening is deliberately broader than accepting.

### 11. `tab-bar-drop-zone.component.test.tsx:343` — Minor

Every widening test seeds tab data, so the panel branch is never exercised. Changing the check to `if (!DragState.getData('tab', dockId)) return;` leaves all 27 drop-zone tests passing. Dragging a panel by the empty part of its bar is an everyday gesture, not just float re-docking, and in that case the zone would stop covering the last tab. Suggested fix: duplicate the LTR marking test with `createDragState({ panel: createPanel({ id: 'other' }) }, DOCK_ID)`.

### 12. `tab-bar-drop-zone.component.test.tsx:113` — Minor

Nothing proves each zone measures its own bar. Replacing `zone.closest('.dock-nav')` with `document.querySelector('.dock-nav')` survives the whole suite. In a split layout that mutation would make every zone size itself from the first bar, so the widening would be 0 or would cover several tabs and steal drops meant to land between them. Suggested fix: render two fixtures with different fake sizes and assert each zone's overlap separately.

### 13. `tab-bar-drop-zone.component.test.tsx:189` — Minor

The two tests that check the indicator is cleared with the token it was claimed with can't actually tell a stable token from a broken one. Neither re-renders nor mounts a second zone, so both a new `{}` per render and a module-level shared `{}` pass. With a shared token, rc-dock calls the new zone's onDragOverT before the old zone's onDragLeaveT, so crossing between bars clears the fresh indicator. Suggested fix: re-render between claim and leave, add a two-zone case, and use a fake setDropRect that follows rc-dock's own rule of clearing only on a matching source.

### 14. `platform-dock-layout-positioning.util.test.ts:67` — Minor

The new wiring test checks both children exist but not their order, which is what actually positions them at rest. Outside a drag, "+" and the zone both have the default flex order, so the order in the Fragment decides the layout; only the drag rule moves "+" to order: 1. Swapping the two children would park "+" permanently at the far end with the test still green. Suggested fix: assert `.new-tab-button` precedes `.platform-tab-bar-drop-zone`.

### 15. `tab-bar-drop-zone.util.ts:38` — Minor

The resolver skips two rc-dock branches based on an app-config assumption that nothing enforces. The doc says the floatable:'singleTab' and disableDock branches are omitted because every group sets floatable:false and never sets disableDock, but only baseConfig in getGroups makes that true and no test ties them together. If someone changes a group's floatable, the zone would offer drops rc-dock's own targets refuse, with no test failing. Suggested fix: a test looping getGroups(true) asserting floatable === false && !disableDock.

### 16. `tab-bar-drop-zone.component.tsx:101` — Minor

This paragraph gives the wrong reason the zone can't push a tab into the overflow dropdown, which could mislead whoever changes these styles next. It says rc-tabs only reads .dock-nav-wrap/.dock-nav-list and so can't be affected, but rc-tabs reads tabsWrapperRef.offsetWidth, and .dock-nav-wrap is a flex sibling of .dock-extra-content with flex-shrink: 1, so its width does depend on the extra content. The actual guard is `flex: 1 0 0; min-width: 0`, which the SCSS at component.scss:8 already explains correctly. Two smaller things in the same block: the filename has a stray leading dot, and "from the last tab's trailing edge to the end of the bar" is only true mid-drag.

### 17. `tab-bar-drop-zone.component.scss:25` — Minor (PR-level)

Raising this once rather than at each site: the new production files carry 2–3× this folder's usual comment density (component.scss 62%, util.ts 51%, component.tsx 45%, against 12–37% for siblings), and the cause is repetition rather than any one long comment. Three explanations each appear in three places: the rc-dock midpoint/30px/hit-testing rationale (component.tsx:52, component.tsx:93, component.scss:25), the "pass the indicator, not the zone, to setDropRect" note (component.tsx:126, component.scss:58, component.test.tsx:165), and the overflow-dropdown claim (component.tsx:101, component.scss:8, component.scss:42). Each copy names rc-dock internals, so an upgrade has to find and fix all of them or they start contradicting each other. Suggested fix: keep one full explanation at the most load-bearing site and reduce the others to a one-line pointer. The content is good — there's just more of it than will get read.

### 18. `platform-dock-layout-positioning.util.ts:68` — Minor

This looks like it deserves an Architecture-Decisions.md entry under CLAUDE.md's "Recording Architecture Decisions". You chose among real alternatives (upstream #222's stretched last tab, patching TabCache directly, an app-side DragDropDiv in panelExtra), introduced a new pattern of an app-registered rc-dock drop target that mirrors rc-dock's own gating rules, and implied a policy of patching rc-dock for bugs but not for features. PT-3330's follow-up would otherwise re-derive all of it. The docking README also lists rc-dock customizations and mentions neither the zone, the "+" slide, nor the patch.

### 19. PR description — Minor

The description is about 1,600 words, and a fresh reviewer meets superseded content before the corrections. Specifics worth fixing: "How it works" still says rc-dock's "~30px strip", which the Review round 1 section says was corrected; the Simple-mode explanation gives the wrong reason (tabLocked does not prevent a drag from starting — the drag-ignore class does, though the conclusion holds); test counts appear as both 192/27-new and 194; Status repeats Verification; the session URL appears twice; and the "Investigated and not changed" last-tab bullet rests on the premise item 1 disputes. The "Review round 1 → Fixed" list is between-round narration the commits already carry. Suggested structure: Summary and video / Tickets / How it works in five bullets / the rc-dock patch promoted out of the footer / Known limits / Verification / Prior art.

### 20. `dock-layout-wrapper.component.scss:282` — Minor

The hidden overflow box is now excluded twice by two mechanisms, with an eight-line comment justifying keeping both. Once `:not(.dock-nav-operations-hidden)` is on the flex rule, the old nested display:none no longer competes, so either alone suffices. The comment argues the redundancy saves future editors from keeping selectors in sync, but it creates exactly that obligation, and the literal "(0,5,0)" goes stale as soon as either selector changes. Suggested fix: keep the `:not()` plus a plain nested display:none, and cut the comment to the two durable facts — the box has flex-grow: 1 and would halve the zone, and rc-tabs never measures it.

### 21. `tab-bar-drop-zone.component.scss:52` — Nit

The ::after hit-area extension hard-codes the same `vars.$space--sm` used as the gap in the other stylesheet, linked only by prose. If someone changes that gap, the 8px drop dead-spot returns with no symptom and no failing test. Giving the gap a shared named variable turns the dependency into code.

### 22. `tab-bar-drop-zone.component.test.tsx:254` — Nit

This test name is a three-line concatenation carrying implementation rationale, and it names a file that doesn't exist (DockTabs.ts; it's .tsx). It also describes the panel case as the float re-docking shape, when the common source of panel data is dragging a docked panel by its bar. Suggested name: "accepts and docks a same-group whole-panel drag (e.g. a re-docked float)", with the rationale moved into a comment.

### 23. `tab-bar-drop-zone-test.util.ts:19` — Nit

A few fixture cleanups now that this file is shared with the positioning test: createContext reads like React's createContext (createDockContext would be clearer), the exported GROUP is unused while tests use the literal 'group-b', the identical beforeEach appears in both describe blocks, and `createDragState(undefined, undefined)` doubles as a store reset that needs a comment at three call sites — a named resetDragStateStore() would remove all three. The file also sits beside production code, while the other -test.util.ts files live in a __tests__ folder.

### 24. `tab-bar-drop-zone.component.tsx:72` — Nit

Worth a few words on why clearLastTabOverlap runs before the measurements: it removes the negative margin so the rects below read the un-widened zone. Measured while widened, the overlap would come out near zero. As written it reads like a redundant reset, and sitting below the no-tabs early return makes its placement look accidental rather than chosen.

### 25. `tab-bar-drop-zone.component.tsx:86` — Nit

`Math.max(0, tabRect.width / 2)` guards against a negative width, which can't happen, so readers go hunting for the case it protects. The relationship worth stating instead is that the zone's start is pulled back to the tab's midpoint, so the tab's trailing edge sits half a tab-width inside the zone.

### 26. `tab-bar-drop-zone.component.tsx:47` — Nit

'data-dragging' is a repeated string literal across the component, the stylesheet and five test assertions, while the two custom properties beside it have named constants. A DRAGGING_ATTRIBUTE constant would match.

### 27. `patches/rc-dock+3.3.2.patch:73` — Nit

`r.closest('li') != null` searches all the way to the document root, so it quietly assumes no li ever wraps the dock layout. If one ever did, every tab ref including the real strip's would be rejected and all drop geometry would break — a hard failure to trace back here. Either stop the walk at the tab strip, or state the assumption in the patch note. The patch logic itself checked out, including that the hunk applies cleanly to the es/ build webpack ships.

### 28. `tab-bar-drop-zone.component.tsx:164` — Nit

Each zone's listener reads layout and then writes styles, so with N panels rc-dock's synchronous listener loop forces N layout passes at drag start instead of one. It happens once per drag and each pass is small, so this is only worth doing if you're already touching this code: read every mounted zone's geometry first, then write.

### 29. `tab-bar-drop-zone.util.test.ts:60` — Nit

This name says "mirroring DockPanel's own tabs-square gate", but DockPanel gates the whole drop layer and there is no tabs square. "rejects a tab from a tabLocked group" says what it checks. Related: two other names embed ticlo/rc-dock#226, which reads better as a comment.

### 30. `tab-bar-drop-zone.component.scss:39` — Nit

The z-index comment explains flex painting order, but the zone is positioned, so DOM order already handles that. The thing a future reader would want here is rc-dock's .dock-drop-edge (top: 30px, z-index: 100), which overlaps roughly the bottom 6px of the ~36px bar, where a drop makes a top split instead of an append. That predates this PR and affects rc-dock's own tabs equally — worth a live check and then either accepting it in the comment or handling it.

## Verified sound (no action needed)

Recorded so these are not re-raised:

- **The `isPopupDiv` patch logic is correct.** The DOM depths it describes match rc-menu 9.6.4's
  `MenuItem`; `closest('li')` is null-guarded and bounded; the hunk applies cleanly to an untouched
  `es/DockTabs.js`; webpack resolves `module` → `es/`, so the patched build is what ships.
- **Listener and handler lifecycle.** Subscribe/unsubscribe use the same callback keyed on a stable
  context; drop, Escape and cancel all end in `destroyDraggingElement` → `listener(null)`;
  `DragDropDiv` unmount clears `_droppingHost`. The `setDropRect` claim/clear mirrors rc-dock's own
  `DockDropSquare`, including on unmount.
- **The drop call is correct.** `dockMove(source, panelData, 'middle')` appends as intended,
  including moving a non-last tab of the same panel to the end.
- **The resolver's gates match rc-dock's own rules** for this app's group config; `tabLocked`,
  window and maximized panels are genuinely unreachable here.
- **The `body:has(> .dragging-layer.dock-style-platform-bible)` rule works** in Electron 39, and
  divider drags correctly carry no group class.
- **The overflow-box claims hold** — rc-tabs 11.16.1 never reads `operationsRef`, so removing the
  box from layout does not disturb its overflow math.
- **Accessibility:** `aria-hidden` reaches the real div, nothing focusable is added, no new key
  handlers (so the keyboard shortcuts catalog needs no update), and no regression to rc-tabs'
  `role="tablist"` semantics.
- **z-index:** `z-index: 1` is local in-flow stacking, not a portalled layer — no tier violation.
- **RTL:** logical properties throughout plus a direction-aware overlap branch.
- **Existing tests do kill many mutants:** unmount clear, RTL math, listener cleanup, `dockMove`
  arguments, `onDrop` re-validation, and zone-vs-indicator passed to `setDropRect`.

> @katherinejensen00 reviewed 10 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2767#-:-P1_seCU5Sum5EeIR0cr:bk87rgb)* status: all files reviewed, 1 unresolved discussion (waiting on rolfheij-sil).

___
*[`src/renderer/components/docking/tab-bar-drop-zone.component.tsx` line 85 at r1](https://reviewable.io/reviews/paranext/paranext-core/2767#-P1_sr8OEiKE8dYU-Y7F:-P1_sr8OEiKE8dYU-Y7G:b-yzmz4t) ([raw file](https://github.com/paranext/paranext-core/blob/8c0554f8bccf64e94892053c0b683a6ba82f0ce2/src/renderer/components/docking/tab-bar-drop-zone.component.tsx#L85)):*
> ```XML
>   // absorbed the bar's whole remainder), so rc-dock's own per-tab target on the last VISIBLE tab is
>   // the only drop target present there.
>   const overlap = Math.max(0, isRtl ? tabMidpoint - zoneRect.right : zoneRect.left - tabMidpoint);
> ```

### 1. `tab-bar-drop-zone.component.tsx:85` — Major

On a tab bar with more tabs than fit, starting a drag can make the whole tab row jump, and the merged drop target you documented isn't there. The comment just above says the last tab is clipped on an overflowing bar so the overlap clamps to 0, but that only holds when the last tab is scrolled out of view. rc-tabs' scrollToTab keeps the active tab visible, and Algorithm.addTabToPanel activates an appended tab, so a crowded bar very often has its last tab both last in DOM order and visible. The overlap is then non-zero, the negative margin shrinks .dock-extra-content, .dock-nav-wrap absorbs the freed width, and rc-tabs' ResizeObserver re-runs useVisibleRange. A repro of these rules in Electron 39 measured the strip widening from 228 to 268px, the zone ending up 2px wide, and a hidden tab popping out of the dropdown. This also fires for every mounted zone on any tab or panel drag in the dock, including moving a floating dialog. Suggested fix: extend backward with an absolutely positioned pseudo-element, the technique you already use for the 8px gap, so it can't feed back into flex sizing; or skip the claim when the bar is overflowing.

rolfheij-sil:
🤖 Claude: Thanks for the thorough review. Most findings held up and are fixed on the branch; there are three small corrections (8, 15, 20) and one decline (28).

You asked for 1, 2, 6 and 8 to be confirmed in the running app. We checked 1, 2 and 6 live (details below). We did not reproduce 8 live, but we changed it anyway because the change costs nothing.

## Fixed

- **1, 2 — crowded bar.** Both confirmed live. Starting a drag widened the tab row by 42px and shifted it 42px. The 2px indicator was drawn 50px past the zone, over the overflow button. The backward extension is now an absolutely positioned `::before`, so it can't change flex layout. The indicator's lead is now the gap back to the last tab's trailing edge, and the zone rejects the drop while the indicator has no width (1e9b4187128). Our own review then found a dead spot this fix created: a clipped last tab under a zero-width zone. That is fixed in 866c124ce68.
- **3 — no layout test.** Added a Playwright spec, run as the isolated `docking` subset (60e8fe00147, crowded-bar checks made unconditional in 9681ae587f4). It drops onto the bar remainder; onto the gap, "+", the trailing padding and the lower band of another panel's bar; and it checks that starting a drag on a crowded bar leaves the tab row alone. Tests 2 and 3 failed before the layout fixes. Test 1 fails if `flex-grow: 1` is removed.
- **4 — patch untested.** `rc-dock-tab-cache-patch.test.ts` imports `TabCache` from `rc-dock/es/DockTabs` and fails on an unpatched install (86e7e526a6d).
- **5 — upgrade guard.** rc-dock is pinned to exactly `3.3.2` (86e7e526a6d). The docking README now has an "rc-dock internals we depend on" upgrade checklist (329354c0a93).
- **7 — "+" dead spot.** Confirmed live: the dead span was about 38px. During a drag, "+", the bar's trailing padding and the bar's full height now accept drops (2498d5f2260). "+" also turned out to be 34px wide despite its declared 30px, because the inline padding set the used width. 36b62dab1db makes both panel buttons really 30px, so the "+" hover background is 4px narrower.
- **9, 10 — drag signals.** The listener comment no longer contradicts itself. It now says why claiming is broader than accepting and documents the ordering dependency on rc-dock appending `.dragging-layer` first (1e9b4187128).
- **11, 12 — tests.** Added a whole-panel drag case and a two-bar case, each checked against your mutation (1e9b4187128).
- **13, 14 — tests.** The token tests now re-render and include a two-zone case with a source-matching `setDropRect`. Both the per-render and the shared-token mutations now fail. The `panelExtra` child order is asserted (a9eb8470010).
- **16 — TSDoc.** Rewritten with the real reason (`flex: 1 0 0; min-width: 0`), plus the filename and the "only mid-drag" wording (1e9b4187128).
- **17 — repetition.** Each explanation is now written once, with one-line pointers elsewhere (577081e0c32).
- **18 — ADR.** Added `adr-tab-bar-drop-zone-app-side-target`, plus README sections on the patches and the rc-dock internals we depend on (329354c0a93).
- **19 — description.** Rewritten in the structure you suggested.
- **21 — shared gap.** A new `_tab-bar-metrics.scss` partial holds the gap, button size and bar padding (2498d5f2260).
- **22, 23, 29 — names and fixtures.** Renamed, moved the fixture to `__tests__/`, and added `createDockContext`, `OTHER_GROUP` and `resetDragStateStore()` (a9eb8470010).
- **24, 25.** The code they referred to is gone after the fix for 1 (1e9b4187128).
- **26.** Added `TAB_BAR_DROP_ZONE_DRAGGING_ATTRIBUTE` (1e9b4187128).
- **27.** The `isPopupDiv` walk now stops at `.dock-nav-list` (86e7e526a6d).
- **30 — edge layer.** Confirmed live, and worse than estimated. rc-dock's edge-split layer covered the bottom 12px of our 42px bar, and a drop there split the whole layout. The layer now starts below the bar (2498d5f2260).

## Fixed, with a correction

- **6 — Simple mode.** You're right that the CSS reaches Simple mode, and the description was wrong. But the live measurement shows the icon-collapse threshold is the same as with main's CSS. The collapse logic compares panel width with the tabs plus a constant; it doesn't use free bar space. So no recalibration was needed. The description is corrected.
- **8 — id compare.** Now compares ids (a9eb8470010). One correction: rc-dock's own drop targets also compare panels by identity, so this code followed rc-dock's convention. We changed it anyway because it's free.
- **15 — group config.** `floatable === false` was already asserted for every group before this PR. We added only the missing `disableDock` assertion (a9eb8470010).
- **20 — double exclusion.** `:not()` alone would not hide the box, because rc-dock's own `.dock-nav-operations { display: flex }` still applies. So we kept `:not()` plus a plain nested `display: none`, as you suggested, and cut the comment (577081e0c32).

## Declined

- **28 — read/write interleave.** This runs once per drag, and the cost is negligible. After the fix for 1, the writes only touch absolutely positioned boxes, never flex layout.

katherinejensen00 (commented):
# PR Review Report: PR #2767 (round 2)

**Repository**: paranext/paranext-core
**Title**: Make the whole tab bar a drop target for dragged tabs
**Author**: Rolf Heij · **Branch**: `pt-3288-tab-bar-drop-zone`
**Head reviewed**: `cff3972dabd` · **Base**: `origin/main` @ `57e7ac1d153`
**Date**: 2026-09-16
**Mode**: not posted (awaiting confirmation)
**Event if posted**: REQUEST_CHANGES
**Reviewers**: security, scope, ux, tests, clarity (default panel, auto-narrowed — no C# files, so `contracts` and `architecture` were dropped)

## Round-1 context

Round 1 was posted 2026-09-15 as CHANGES_REQUESTED against head `8c0554f8bcc` — 30 issues, 5 Major.
That head is no longer an ancestor of this branch (the branch was rebuilt), and 24 commits now sit on
top, so this round re-reviewed the current head from scratch rather than diffing against round 1.

Round 1 carried an explicit caveat: its antagonistic pass and `/code-review max` both died on a rate
limit, so none of its findings had survived an adversarial second opinion, and items 1, 2, 6 and 8
were flagged as "confirm in the running app before approving". This round did not run in the app
either — the two crowded-bar / Simple-mode items below are the same class of claim and carry the same
caveat.

## Summary

|-------------|----------|---------|------------|
| CI (verified) | 1 | 0 | 0 |
| Security    | 0 | 0 | 2 |
| Scope       | 0 | 3 | 1 |
| UX          | 0 | 2 | 8 |
| Tests       | 0 | 1 | 5 |
| Clarity     | 0 | 0 | 7 |
| **Total**   | **1** | **6** | **23** |

Inline-eligible (decision needed): 9. Summary-only (informational): 21.

## Blocking

### [CI] `src/renderer/components/docking/__tests__/tab-bar-drop-zone-test.util.ts:2`

**Severity**: blocking · **Language**: typescript · **needs_decision**: true

The Ubuntu CI leg is red on `verify third-party notices against the committed lock`:

```
ERROR: 1 package(s) imported by first-party source reach no bundle and are recorded nowhere:
  vitest (imported in src/renderer/components/docking/__tests__/tab-bar-drop-zone-test.util.ts)
```

Root cause confirmed in `.erb/scripts/third-party-notices/shipping-set.ts:453`. The scanner treats
everything under `src/` as shipping first-party source and excludes test-support helpers by filename:

```ts
!/\.test-(?:utils?|harness)\./.test(path.basename(file)) &&
!file.split(path.sep).includes('__mocks__'),
```

The new helper is named `tab-bar-drop-zone-test.util.ts` — hyphen before `test`, `.util.` after — so
it matches neither `*.test-utils.*` nor `*.test-harness.*`, and its `import { vi } from 'vitest'` is
read as a shipping runtime dependency. The other new test files are `*.test.ts(x)` and are excluded
upstream by `findPrebuiltSources`, which is why only this one file trips it.

**Fix**: rename to the established convention — `tab-bar-drop-zone.test-utils.ts` — and update the
three importers (`platform-dock-layout-positioning.util.test.ts:16`,
`rc-dock-tab-cache-patch.test.ts:6`, `tab-bar-drop-zone.component.test.tsx:18`,
`tab-bar-drop-zone.util.test.ts:11`). Existing precedent:
`extensions/src/platform-scripture/src/manage-books-dialog/manage-books-dialog.test-utils.ts`,
`lib/platform-bible-react/src/components/advanced/footnote-editor/footnote-editor.test-harness.tsx`.
Recording `vitest` in `notices-policy.json` under `unbundledDependencies` would also silence it, but
the rename matches the convention the scanner was written around.

This dovetails with the Clarity reviewer's independent finding that the helper is misnamed for a
different reason (it is now the shared rc-dock fixture for three unrelated suites) — one rename can
settle both.

## Warnings

### [Scope] `src/renderer/components/docking/dock-layout-wrapper.component.scss:306`

**needs_decision**: true

**Question**: Should the `display: none` on the idle overflow box be scoped to
`.dock-style-platform-bible` (Power mode) so Simple mode's tab bars are left unchanged, or is
applying it in both modes an intended, separately-verified decision?

The rule removing rc-tabs' idle overflow box from layout is written on a bare `.dock-nav-operations`
selector, unqualified by `.dock-style-platform-bible` — unlike nearly every other tab-bar rule in
this file. It therefore applies to every dock panel in the app, including Simple mode's Resources
column and the headless group, where there is no `panelExtra` and so no drop zone to benefit. The
companion `:not(.dock-nav-operations-hidden)` on line 289 *is* mode-scoped, which makes this one look
unintentional rather than deliberate. The safety argument in the comment ("rc-tabs never measures
this box") is a claim about rc-tabs internals that the Power-mode-only e2e subset never exercises for
Simple mode.

*(Reported at line 307, the `&.dock-nav-operations-hidden {` selector; corrected to 306 — the last
added line of the diff hunk, immediately above it.)*

### [UX] `src/renderer/components/docking/dock-layout-wrapper.component.scss:289`

**needs_decision**: true

**Question**: Was Simple-mode tab collapse (including the Resources column) manually verified at the
collapse/overflow boundary after this specificity fix, or is the no-regression claim inferred from
the rc-tabs "never measures the operations box" note alone?

This `:not()` is a real behavior change in **both** modes, not a no-op cleanup. Before this PR the
pre-existing `.dock-nav-operations.dock-nav-operations-hidden { display: none; }` (specificity 0,2,0)
lost to `.dock-panel.dock-style-platform-bible .dock-nav-operations { display: flex; }` (0,3,0), so
the idle overflow box was in fact laid out — flex, 30px tall, `flex-grow: 1` from `.dock-nav`. Adding
the `:not()` finally lets `display: none` win. In Power mode that is the point. In Simple mode there
is no `.dock-extra-content`, so `.dock-nav-wrap` (`flex-grow: 0`) now sits in a row where nothing
claims the leftover space — a different layout from what the Simple-mode tab-collapse reasoning in
`platform-tab-title.component.tsx` was tuned against, and exactly the width at the transition point
that governs when a tab collapses to icon-only or falls into the dropdown.

This is the same area as round 1's item #6 ("the visual size of the change was never measured, and
the Resources column was never opened"), reached independently. It is still unmeasured.

### [Scope] `src/renderer/components/docking/dock-panel-button-base.scss:23`

**needs_decision**: true

**Question**: Is the 2px narrowing of the overflow ("more") button in both Power and Simple mode an
accepted visual change, and has it been eyeballed — or should the 30px guarantee be expressed on the
"+" button alone rather than on the shared `%panel-button-base`?

`%panel-button-base` has two consumers: the Power-mode "+" button and `.dock-nav-more`, the overflow
button, which is styled by an unscoped selector and so exists in Simple mode too. Changing
`padding-inline` from `16px` to `0` under `box-sizing: border-box` narrows the used width of both
from 32px to 30px, solely so the drop zone's `::after` hit-area arithmetic can assume an exactly-30px
button. The regression surface — the overflow button's hit target and spacing outside Power mode —
is not covered by the new `docking` e2e subset, which drives only Power-mode drags.

### [Scope] `patches/rc-dock+3.3.2.patch:54`

**needs_decision**: true

**Question**: Should the `isPopupDiv` rc-dock patch (and its `rc-dock-tab-cache-patch.test.ts`) ship
in this PR, or be split into its own PR since it changes drop geometry for existing tab-to-tab drags
app-wide rather than enabling the new tab-bar drop zone?

The `isPopupDiv` fix is an independent upstream defect fix (ticlo/rc-dock#161), not a change required
to render or accept the new drop zone. By the patch's own note the bug corrupts `TabCache._hitAreaRef`
for any tab once the overflow dropdown has rendered, so every drop-position calculation for that tab
reads a zero-sized rect — meaning this fix changes pre-existing tab-to-tab drag/drop geometry across
the whole app, in both modes, independently of the new zone. It is declared in scope by the PR body
and it is a genuine bug fix, but it is separately shippable and separately reviewable, and only one of
the three new e2e tests touches the overflow-dropdown condition that triggers it.

### [UX] `src/renderer/components/docking/tab-bar-drop-zone.component.tsx:201`

**needs_decision**: true

**Question**: Is shipping this append gesture as pointer-only acceptable for now, or should a
keyboard/context-menu equivalent (e.g. a "Move tab to…" menu item) be tracked as a follow-up ticket
before merge?

The new "append a tab to this bar" gesture is pointer-only and deliberately hidden from assistive
tech (`aria-hidden="true"`, no `role`, no tabindex, no keyboard handlers). That is the right
treatment for an invisible geometry-only surface, but it means the capability the PR adds has no
keyboard or AT-reachable equivalent anywhere: the tab context menu offers Float/Close, not "move to
panel", and `src/stories/keyboard-shortcuts.data.ts` has no tab-move entry. Existing tab
drag-rearranging is already pointer-only, so this is not a regression — it widens an existing gap
rather than narrowing it.

### [Tests] `src/renderer/components/docking/tab-bar-drop-zone.component.test.tsx:458`

**needs_decision**: true

**Question**: Should a fourth e2e case be added that starts a drag and releases on the last tab's
trailing half, asserting the tab is appended at the end (the zone's `middle`) rather than inserted by
rc-dock's own `after-tab` target — or is the CSS-variable-level assertion accepted as the coverage
for the overlap mechanism?

The headline behavior of `claimLastTabOverlap` — the zone claiming the last tab's trailing half so
the bar presents one continuous drop target instead of two — is never verified as a drop anywhere in
the suite. The unit tests stop at the CSS custom-property strings
(`--tab-bar-drop-zone-overlap: 58px`, `--tab-bar-drop-zone-indicator-lead: 8px`), and the hit area
that consumes them is the `&[data-dragging]::before` rule in the SCSS, which no test touches (jsdom
applies no SCSS). The e2e spec drops at 75% across the zone, in the gap before "+", on "+", in the
bar's end padding and in the bar's lower band — never on the last tab's trailing half. Net effect:
delete the `::before` block, or rename either custom property in the SCSS only, and every test in
this PR still passes while the feature's central claim silently regresses.

## Suggestions (decision requested)

### [UX] `src/renderer/components/docking/tab-bar-drop-zone.component.scss:35` — raw `z-index: 1`

**Question**: Should this be left as a local in-bar `z-index: 1` (documented as intentionally outside
the named scale), or should a low "bar content" tier be added to the z-index scale and its SCSS twin
so the rule holds uniformly?

The repo rule `.claude/rules/ux/z-index-tiers.md` says stacking order comes from the named scale
(`lib/platform-bible-react/src/components/z-index.ts` / `src/renderer/styles/_vars.scss`), never a
bare number — and this surface is explicitly reasoned about against another layer's stacking (the
comment cites `.dock-drop-edge`'s z-index 100 and the positioned `.dock-nav-wrap`). The scale has no
in-panel tier and every existing tier (400+) would be far too high here, so this is arguably outside
the scale's intent rather than a violation — but it is a stacking decision recorded as a literal that
nothing tests or orders.

### [UX] `src/renderer/components/docking/dock-layout-wrapper.component.scss:271` — the mid-drag "+" jump

**Question**: Has the mid-drag "+" jump been eyeballed on a wide, sparsely-populated bar (where it
travels the furthest), and is the unanimated jump acceptable?

The "+" button teleports from just-after-the-tabs to the far end of the bar the instant any tab drag
starts, with no transition. Users were shown to mistake a far-end "+" for a close button (the
rationale in the comment above the sibling `order` rules), so the drag state momentarily restores
exactly the arrangement usability testing rejected, and the jump is an uncued layout shift under the
pointer. Drop-target continuity is a sound reason for the move; this is about the transition, not the
end state.

## Suggestions (informational — summary only)

### Security

- **`patches/rc-dock+3.3.2.patch:76`** — the new hunk carries no injection or privilege risk: a
  read-only DOM ancestor walk over static selector literals, no user-controlled input, no
  `innerHTML`/`eval`, no new network/file/process access. One note for visibility as the patch grows:
  this patch file touches `node_modules/rc-dock/es/` only, while rc-dock's `main` is `lib/index`
  (CommonJS). Anything resolving `main` — vitest, node-side tooling — runs **unpatched** third-party
  code, including the pre-existing PT-3411 null guard. The PR already works around this by importing
  `rc-dock/es/DockTabs` directly, so the split is deliberate.
- **`package.json:161`** — pinning `^3.3.2` → exact `3.3.2` is a supply-chain improvement for a
  patched dependency; patch-package's checksum is tied to the exact version. The lock diff changes
  only the range string (`resolved`/`integrity` untouched). Residual risk: patch application depends
  on `postinstall` running. In this review worktree the patch is **not** applied
  (`grep -c closest node_modules/rc-dock/es/DockTabs.js` → 0) — the known `npm i --ignore-scripts`
  failure mode, which silently loses both the hit-area fix this PR depends on and the PT-3411 crash
  guard with no build-time signal. Consider a CI assertion that the patch is applied.

### Scope

- **`.context/standards/Architecture-Decisions.md:4967`** — the ADR entry is correct and correctly
  placed: slug `adr-tab-bar-drop-zone-app-side-target` sits between `adr-styleinfo-over-css` (4915)
  and `adr-tab-menu-channel-and-window-naming` (4969), the right `LC_ALL=C` byte-order position. One
  wording mismatch: the PR body says PT-3288 and PT-3330 are explicitly *not* closed, while the ADR's
  Context attributes the no-op to "(PT-3288)" and Source reads "PT-3288, PR #2767" — easy to read
  later as "this ADR closes PT-3288". Consider qualifying it.

### UX

- **`tab-bar-drop-zone.component.tsx:191`** — every mounted zone registers its own module-global
  rc-dock drag-state listener, so one drag start fans out to N zones, each synchronously running
  `querySelectorAll` + two `getBoundingClientRect()` + `getComputedStyle` — N forced layout flushes
  at the frame the drag begins. Bounded by panel count and only on drag start/end, so very unlikely
  to be felt today; noted as the first thing to look at if drag start ever feels sticky in a heavily
  split Power-mode layout.
- **`tab-bar-drop-zone.component.tsx:34`** — `MIN_INDICATOR_WIDTH = 1` means a nearly-full bar still
  accepts the drop while showing a literal hairline indicator, visually indistinguishable from no
  feedback, so the user gets an unannounced move. The reject path is well reasoned (it hands the drop
  back to the last tab's own rc-dock target); the cutoff is just set where feedback is already
  imperceptible rather than where it stops being legible. Something near the 30px indicator rc-dock
  itself draws would make accepted drops always visibly confirmed.
- **`_tab-bar-metrics.scss:10`** — good move centralizing bar metrics, but the sweep is partial:
  `--dock-tab-height: 30px` (`dock-layout-wrapper.component.scss:336`) and rc-dock's own fixed 30px
  `after-tab` indicator are the same magic number left un-tokenized, and the drop zone's `::after`
  width is now an exact arithmetic sum of these tokens — so a future change to tab height alone
  silently desynchronizes the hit area from the bar it must cover exactly.
- **`dock-panel-button-base.scss:23`** — both consumers are icon-only and fixed-size so the visual
  result is the same, but the class is now asymmetric (`padding-inline: 0` alongside
  `padding-block: 8px`) and its width has become load-bearing for a stylesheet in another file.
  Consider dropping the now-inert `padding-block: 8px` so the asymmetry isn't read as meaningful. No
  accessibility impact — the 30×30 hit target still meets the usual minimum for a dense toolbar row.
- **`tab-bar-drop-zone.component.tsx:90`** — RTL handling is logically correct on inspection: the CSS
  uses `inset-inline-*` throughout and the JS mirror (`gapToTab`, `tabMidpoint`) flips the correct
  edges, and DOM-last is still logically-last in RTL. The residual risk is narrow but real: the
  `::after` forward extension is a fixed SCSS sum (gap + button + padding) assumed to sit past
  `inset-inline-start: 100%`, which only lines up if the mid-drag "+" relocation lands on the correct
  visual end in RTL too. Worth one manual RTL pass before merge.
- **`tab-bar-drop-zone.component.scss:73`** — verified clean on two checks: (1) theming — neither the
  zone nor the indicator declares a single color; all visible feedback comes from rc-dock's themed
  `.dock-drop-indicator`, so nothing can break in dark mode and there is no hardcoded color in the
  new SCSS. (2) `.claude/rules/keyboard-shortcuts-catalog.md` — the PR's claim holds: no
  `keydown`/`keyup`/`useHotkeys`/`onKeyDown` in any added or changed file; Escape-to-cancel is
  rc-dock's own pre-existing handling surfaced through `addDragStateListener`; no
  `keyboard-shortcuts.data.ts` entry is needed.

### Tests

- **`tab-bar-drop-zone.spec.ts:447`** — the crowded-bar test deliberately excludes the bar's last tab
  (`tab.id !== during.lastTabId`), so the safety property `claimLastTabOverlap` exists to protect on
  a crowded bar ("the zone leaves the last tab uncovered when it has no indicator to show, so the tab
  still takes drops on its own") is asserted only as `--tab-bar-drop-zone-overlap: 0px` in jsdom.
  Nothing verifies a release on the crowded bar's last visible tab actually appends — the exact
  failure mode the code comments say would be silent.
- **`tab-bar-drop-zone.spec.ts:210`** — `expectAppendedTo` makes the primary outcome of each drop a
  *soft* assertion, and the four `test.step`s run sequentially against shared, mutated state. If the
  first drop doesn't land, three more drags run from a layout that no longer matches the assumptions,
  producing a cascade of failures pointing away from the real cause. Soft is well-justified for the
  indicator check inside `dragTabTo`; for the outcome assertion a hard `expect(...).toPass()` would
  stop at the first genuine failure.
- **`tab-bar-drop-zone.component.test.tsx:48`** — mock-budget note: this file replaces `DragDropDiv`,
  `addDragStateListener` and `removeDragStateListener`, plus a 13-method `createDockContext` stub.
  The partial mock is well-chosen — `DragState` stays real, so drag data round-trips through
  rc-dock's actual module-level store rather than a mock of it. Residual gap worth naming: because
  `DragDropDiv` is replaced and handlers are invoked directly, nothing at unit level covers that the
  zone is registered with rc-dock's `DragManager` or reachable by hit-test. That rests entirely on
  the three Electron-launching e2e tests.
- **`rc-dock-tab-cache-patch.test.ts:65`** — the first case asserts a field's *initial* value:
  `_hitAreaRef` is already `undefined` on a fresh `TabCache`, so it also passes if `getHitAreaRef`
  did nothing. Only the second case pins "rejected" rather than "never set". Seeding a sentinel ref
  first would stop it passing for the wrong reason. On the "can it silently pass unpatched" question:
  it cannot — unpatched `isPopupDiv` checks only parent and grandparent, and the popup fixture puts
  `.dock-tab-hit-area` three levels under the `<li>`, so cases 1 and 2 both fail loudly without the
  patch. Cases 3 and 4 are behavior-preservation guards (4 pins the `.dock-nav-list` stop condition a
  naive `closest('li')` would break) — the right shape for a patch test.
- **`platform-dock-layout-positioning.util.test.ts:46`** — the three `disableDock` assertions read as
  testing the absence of a property nobody sets, but they are load-bearing:
  `resolveTabBarDropZoneSource`'s TSDoc justifies *not* replicating rc-dock's `disableDock` branch
  precisely because no group here sets it. A one-line comment pointing at the resolver would stop the
  next tidy-up deleting them as noise. Same for the `floatable`/`tabLocked` assumptions.

### Clarity

- **`tab-bar-drop-zone.spec.ts:264`** — two unexplained magic tolerances in the drag-time geometry
  guard: `bar.right - 24` and `zone.left + 8`. The `8` is presumably the flex gap
  (`$tab-bar-extra-gap`) and the `24` a slack allowance for trailing padding, but neither says so,
  and a future change to `_tab-bar-metrics.scss` won't be traceable to them. The second assertion
  also drops the descriptive label its sibling carries, so a failure there reports only a bare number
  comparison.
- **`tab-bar-drop-zone.spec.ts:104`** — `readWhenSettled` has three names for two values: `first` is
  already `unknown`, then `second` is re-aliased into `secondReading: unknown` purely to satisfy the
  matcher. Annotating the matcher argument once would drop a variable and a comment.
- **`tab-bar-drop-zone.spec.ts:361`** — the `40` iteration cap is a bare magic number with no
  statement of why 40 tabs suffices to overflow, or what happens if the loop exhausts without
  overflow (the `expect` after the loop is the implicit failure, but that link isn't stated).
- **`__tests__/tab-bar-drop-zone-test.util.ts:6`** — the file is named and located as if it belonged
  to the drop-zone tests, but it is now the shared rc-dock fixture module for three unrelated suites,
  so a reader of `platform-dock-layout-positioning.util.test.ts` or `rc-dock-tab-cache-patch.test.ts`
  follows an import into a file named after a feature they aren't testing. Also, every other test in
  this directory sits in the directory itself rather than in `__tests__/`, so the extra folder for a
  single helper is an inconsistency of its own. **This is the same file as the blocking CI finding —
  one rename settles both.**
- **`tab-bar-drop-zone.component.tsx:169`** — the rc-dock-internals narrative is written out at
  length in four places: this 19-line inline block, the 20-line TSDoc on `claimLastTabOverlap`, the
  30-line TSDoc on `resolveTabBarDropZoneSource`, and the new README's `rc-dock internals we depend
  on` section, which was added precisely to be the upgrade checklist. All of it is forward-facing
  (constraints and rationale, no PR-history narration), so none should simply be deleted — but the
  same facts now have to be kept in sync across two files, and the code is roughly half prose.
  Suggest keeping the invariants next to the code and moving the long derivations to the README with
  a one-line pointer.
- **`tab-bar-drop-zone.component.test.tsx:444`** — `startDragOverLtrBar(zoneStart, dragData, zoneEnd)`
  forces three call sites to pass `undefined` for a parameter they don't care about just to reach the
  trailing one: `(590, undefined, 590)`, `(600, undefined, 600)`, `(608, undefined, 608)`. An options
  object would make each scenario — especially the zero-width-zone cases — self-describing.
- **`platform-tab-title.component.tsx:780`** — the edit to this pre-existing comment block left one
  line at 121 characters while every neighbouring line wraps at ~100, giving the paragraph a ragged
  edge in the diff.

## Positive notes

- **Forward-facing comments**: the strip-the-PR-context test passes throughout — no ticket IDs for
  in-PR work, stage tags, "previously"/"used to" narration, untracked TODO/FIXME, commented-out code,
  or unused imports in any changed file.
- **Suppression discipline**: every added suppression (`@ts-expect-error ts(2345)` in
  `clearDropRect`, `no-null/no-null`, `no-bitwise`, `no-underscore-dangle`, `no-await-in-loop`,
  `no-type-assertion`) carries a specific justification naming the rc-dock or DOM contract forcing it.
- **No blocked imports, no secrets, no CSP concerns**: all changed TS/TSX imports only `react` and
  `rc-dock`; no `fs`, `child_process`, `process.env`, `eval`, `new Function`, `innerHTML` or
  `dangerouslySetInnerHTML` anywhere in the diff, including the 473-line e2e spec.
- **ADR recorded and correctly slug-sorted**, per the CLAUDE.md requirement.
- **The patch test is genuinely falsifiable** against an unpatched `node_modules` (cases 1 and 2 fail
  loudly), and importing `rc-dock/es/DockTabs` directly is the right call, documented in-file.

## Line validation

9 inline-eligible findings checked against the working tree: **8 kept, 1 corrected, 0 dropped.**

| File | Reported | Corrected | Offset | Reason |
|------|----------|-----------|--------|--------|
| `src/renderer/components/docking/dock-layout-wrapper.component.scss` | 307 | 306 | −1 | `line_content` matched exactly at 307, but 307 is outside the PR diff (hunk `@@ -270,0 +301,6 @@` adds 301–306 only); moved to the last added line, immediately above. |

Diff-hunk membership confirmed for `dock-layout-wrapper.component.scss` (271, 289), 
`dock-panel-button-base.scss` (23) and `patches/rc-dock+3.3.2.patch` (54); all other anchors are in
added files.

## Gating

Decision-needed gating: 9 kept inline, 21 moved to summary (`needs_decision: false`).

## Not covered by this round

- The `architecture` and `contracts` reviewers were dropped by the panel's auto-narrow rule (no C#
  files changed). For a PR that patches a third-party library and adds a new renderer pattern, that
  is a gap worth noting — consider `--reviewers architecture` as a follow-up pass.
- No adversarial/false-positive pass ran over these findings, same as round 1.
- Nothing was verified in the running app. The two Simple-mode findings
  (`dock-layout-wrapper.component.scss:289`, `dock-panel-button-base.scss:23`) and the RTL note are
  the ones that need it.

> @katherinejensen00 partially reviewed 21 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2767#-:-P1jn3Vl-VNPsreejwfY:b85xg2w)* status: all files reviewed, 1 unresolved discussion (waiting on rolfheij-sil).

rolfheij-sil:
🤖 Claude: Thanks for the second round. Your blocking finding was already fixed, most of the rest is
now done, and there are four things I kept as they are — each with the reason. Two of your findings
turned out to rest on premises that don't hold; details below, since I'd rather show the measurement
than assert it.

The branch is rebased onto `main` after #2769 merged.

## Already fixed before this round

- **The CI failure and the fixture name.** Fixed in `f74afb03306`, with exactly the rename you
  suggested: `tab-bar-drop-zone.test-utils.ts`. Ubuntu, macOS, Windows and CodeQL were green on that
  commit. Your Clarity finding about the same file is handled too: the file is now
  `__tests__/rc-dock.test-utils.ts`, since four docking suites use it, not just the drop-zone tests.
  It keeps the `.test-utils.` infix, because that is what the notices scan recognises, and it stays
  in `__tests__/` — round 1 asked for it to be moved there.

## Fixed in this round

- **The last tab's trailing half had no drop test, and deleting `::before` passed everything.**
  Fixed, and thank you — this was the most valuable finding of the round. There is now a Playwright
  case that releases a tab on the last tab's trailing half. It reads the drop indicator *before*
  releasing and checks its shape, so it fails if the zone stops claiming that half: rc-dock's own
  after-tab target draws a fixed strip straddling the tab's edge, while the zone's indicator runs
  from that edge to the zone's far end. Asserting only the final tab order would pass either way.
- **The crowded bar never released on the last tab.** There is now a release past the bar's last
  tab, asserting the tab is appended even though the zone has no width of its own to claim there. It
  fails with both rectangles named if that tab's trailing half isn't inside the bar's visible
  bounds, rather than releasing somewhere unintended.
- **`MIN_INDICATOR_WIDTH = 1` showed a hairline.** The drawn indicator now widens backward over the
  claimed region up to the 30px rc-dock itself draws, so an accepted drop always shows visible
  feedback. Raising the *accept* threshold instead would have created new dead areas on a nearly
  full bar, so acceptance is now tied to the indicator being visible at all, and claiming and
  accepting are the same condition evaluated at two moments: the zone accepts exactly where it
  claims. Since rc-dock hit-tests the topmost element and doesn't fall through to the tab
  underneath, a claimed-then-rejected region would be a dead spot; that state is now
  unrepresentable.
- **Soft outcome assertions could cascade.** Each drop's outcome is asserted hard now. The soft
  assertion inside the indicator check stays, as you suggested.
- **The patch test's first case passed on a fresh cache's own `undefined`.** It seeds a sentinel
  now, and so does its sibling, which had the same weakness. Verified by reverting the rc-dock patch
  and watching both fail.
- **The `disableDock` assertions read as noise.** They now point at the resolver whose reasoning
  depends on them.
- **The spec's magic numbers.** `bar.right - 24`, `zone.left + 8` and the 40-iteration cap are named
  constants that say where their values come from, and the second assertion has the label its
  sibling had.
- **`readWhenSettled`'s three names for two values.** Down to one annotated matcher argument.
- **`startDragOverLtrBar`'s positional `undefined`s.** It takes an options object now, so each
  scenario reads for itself.
- **The 121-character comment line.** Rewrapped.
- **The ADR.** It now says the decision addresses the reported defect without closing it, and that
  the follow-up is tracked separately — phrased so it doesn't go stale when either ticket moves. It
  also records the indicator behaviour above, which the entry didn't mention.
- **The metrics sweep.** `--dock-tab-height` now reads `$panel-button-size` rather than repeating
  `30px`, with the reason stated. The inert `padding-block` is gone.

## Answers

- **Was Simple-mode collapse verified (`:289`)?** Yes, measured live before this round: tab titles
  collapse to icons at the same width as with `main`'s CSS (text at 538px, icons at 518px, no
  overflow down to 297px). The collapse logic compares panel width against the tabs plus a constant,
  not against the bar's free space, so the rule can't move that threshold.
- **Should the overflow-box rule be scoped to Power mode (`:306`)?** It's deliberate in both modes,
  and the code now says so. rc-tabs measures only `.dock-nav-wrap` and `.dock-nav-list`, so hiding
  the idle box can't disturb its overflow maths.
- **The "more" button's 2px.** I checked this in the running app rather than reasoning about it.
  `.dock-nav-more` computes `box-sizing: border-box` and renders exactly 30×30 in **both** Power and
  Simple mode — the same as before, since the padding never reached its outer box. There's no
  content-box inflation anywhere, so the change really is inert, and both buttons stay the same
  size.
- **Split the `isPopupDiv` patch into its own PR?** I'd rather keep it here. The zone hands crowded-
  bar drops back to the last tab's own rc-dock target, and that target's hit area is the thing the
  patch repairs, so the feature depends on it. The patch, its test and the pin are described in the
  PR body.
- **Track a keyboard equivalent?** Filed as **PT-4674**, "Move a tab to another tab group without a
  mouse", under the Tabs epic and linked to PT-3034 and PT-4282.
- **The raw `z-index: 1`.** Kept, with the reasoning in the comment: that scale orders layers
  stacked above the page — its lowest tier is far higher — while this orders siblings inside one tab
  bar. The comment now also names the surfaces it was verified against.
- **The "+" jump.** Kept unanimated. `order` can't be transitioned, so animating it would need a
  measure-then-animate step that captures the button's position before rc-dock adds its drag layer,
  reverses on drop and respects reduced motion — more machinery than the moment deserves, since the
  button only sits there for the duration of the drag.
- **RTL.** Checked in the running app with the dock forced to RTL: the "+" sits at the visual (left)
  end, tabs are right-aligned, and nothing overlaps or clips. On a bar that was *already* overflowing
  the result was inconclusive, because flipping direction after layout leaves rc-dock's own pixel
  scroll offset pointing the wrong way — an artefact of the method, not something I can claim either
  way. A genuine RTL locale would settle it.
- **A CI assertion that the patch is applied.** There already is one, in effect:
  `rc-dock-tab-cache-patch.test.ts` imports `rc-dock/es/DockTabs` and fails on an unpatched install,
  which is how I verified the sentinel change above. It covers the `isPopupDiv` hunk; the older
  PT-3411 guard in the same patch file isn't asserted by anything.

## Kept as they are

- **The layout reads at drag start.** They run once per drag and only touch absolutely positioned
  boxes, so they can't feed back into the bar's layout. Same reasoning as round 1's item 28.
- **The rc-dock notes in several places.** Round 1 consolidated these once, and each copy now serves
  a different reader: the invariants sit next to the code that depends on them, the derivations in
  the README's upgrade checklist. Cutting further would move the constraint away from the code that
  breaks without it.

## On the round itself

Two of the findings didn't survive checking, and I mention it only because both were reasoned from
the source rather than from a run:

- The prediction that the round would flip the crowded-bar outcome and break the existing assertion
  at `:477` doesn't happen — that test passes, because on a real crowded bar the last tab is clipped
  under the zone, which is the case the code leaves alone.
- The "more" button's box is unchanged, per the measurement above.

Both of those are the class of item your report flagged as needing a run, so this seems the right
place to say: the branch has an e2e subset (`npm run test:e2e:isolated tests/isolated/docking/`)
that covers them, and I'm happy to add cases if you can think of a behaviour it still can't see.

katherinejensen00 (approved):
Thanks for making those changes, Rolf! It looks good

> @katherinejensen00 partially reviewed 15 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2767#-:-P1p2w9k60Au6rHT_0Xr:b-s9uf5c)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

Issue #237, opened 2023-06-09
Design project system


tjcouch-sil:
We should probably think of a way to add project information contextually to various calls on the papi. For webviews, this may be a bit simpler - they may be able to be tied down to one project (we need to think about this). For the extension host, that may not be the case. We should think of some kind of nice and consistent way to carry contextual information around the papi like the project that some operation is relevant to.

We probably ought to pursue a few things in designing a simple project/resource system:

1. Some way to know what project/resource you're using (in data providers, etc)
2. A resource/project provider api or something similar - provides a resource of a particular type. papi holds the resource and manages its state. Like USFM resource provider or something. Probably in C# process
3. Resource/project view provider api - displays the project/resource and lets you edit it

[VSCode's custom editor api page](https://code.visualstudio.com/api/extension-guides/custom-editors)
[VSCode's sample custom editor extension](https://github.com/microsoft/vscode-extension-samples/tree/main/custom-editor-sample)

A while ago, TJ created a [proposed Resource Management design](https://docs.google.com/document/d/1nzLREjhttuJ_T4-kcB3RZj-oM4pDCk32D2oMkBcQTLY/edit#heading=h.kzqqqkjblwgt) that is quite extensive. It probably needs some tuning now that we have a significant amount of the papi implemented, but we can probably find a subset to implement for now. The main api features in the design follow:
- Downloadable resource provider - discovers downloadable resources/updates for the download resource dialog. Connects to things like DBL, marble, etc. (Probably need some very similar equivalent for projects as well)
- Virtual Directory Provider - enables interacting with zipped/encrypted directories and serves resource info like what type, what name, etc
- Resource/Project Provider - provides a document object that manages reading/writing to a project or resource
- Resource/Project View Provider - provides a web view for interacting with a project or resource

Our C# process will probably suffice for downloadable resource provider, virtual directory provider, and maybe resource provider for now. We are making a Scripture editor that will be the resource view provider (#60).

<details><summary>TJ's proposed Resource/Project Provider system</summary>

![image](https://github.com/paranext/paranext-core/assets/104016682/1648ce96-3382-489b-9126-e3bb36c0d1ed)

</details>

tjcouch-sil:
Project Design doc: https://docs.google.com/document/d/1K-lNUzzThDvDrNrnU2UXXHny30SgjhIxfkC-Xc1fk5o/edit

PR #2768, opened 2026-09-03
PT-4466: PT-4344 follow-ups and narrow-width toolbar fixes

jolierabideau:
## Summary

Clears the follow-ups [PR #2701](https://github.com/paranext/paranext-core/pull/2701) (PT-4344) deferred, and fixes the narrow-width defects Sebastian reported against Studio #237 afterwards. All of it is Simple mode at or near its minimum widths: a 900px window, and a 297px editor column inside a window of any size.

### Why review this

Part of the current epic — PT-4466 is a sub-task of [PT-4336 "Saroj and Donna can trust the app"](https://paratextstudio.atlassian.net/browse/PT-4336). Visible layout breakage at the widths Simple mode actually ships at is squarely "can I trust this app", and the last item found here (a USFM marker sliced in half) is reachable by an ordinary splitter drag.

## Changes

**PT-4344 follow-ups**

- `model-text-panel.component.tsx` — `tw:truncate` sat on a `tw:flex` container, making `text-overflow: ellipsis` inert, so the resource name hard-clipped mid-glyph at a 297px column. Moved to the `min-w-0` container / `truncate` inner-span idiom PT-4344 established.
- `WINDOW_MIN_WIDTH_PX` extracted to `src/shared/models/window-constraints.model.ts`, so `main.ts` and `simple-layout.data.ts` stop mirroring `900` by hand and the test can bind to the real value.
- The test-only `shrinkStep` prop is removed from `TabToolbar`, `Toolbar`, `BookChapterControl` and `EnhancedResourceTabBarProps`, replaced by an exported `<ShrinkStepOverride value={n}>`.

**Paragraph-style trigger**

- The trigger had its trailing border and chevron sliced off instead of the label ellipsising. It computes `flex: 0 0 auto` inside a block wrapper, so shadcn's `shrink-0` cannot be shrunk past — it took its content width and overran the toolbar zone, whose `overflow-clip` cut it. `tw:max-w-full` puts the squeeze back on the label.
- The marker slot's fixed `tw:w-[6ch]` reserved six monospace glyphs for every marker, so a one-character `p` rendered followed by five blanks. Now sized to its content. **This overrides the "UX set six" decision** and wants a UX opinion.
- At the 297px column floor the start zone has less room than its controls need, and the marker — the one field with no shorter form — absorbed the whole deficit. Two rungs were added below "drop the secondary field", content last to give way: `TabToolbarContainer` halves its own padding and inter-zone gaps at `SHRINK_STEP.MINIMUM`, then the trigger drops its chevron and floors at `min-w-min`.
- The floor is applied **only** at `MINIMUM`: while the style name is rendered it contributes its longest word to `min-content`, and a floor there stops the trigger shrinking at all. Since the floor must be step-aware and the web view renders the `TabToolbar` (so it always reads the widest step), the control moved into `paragraph-style-trigger.component.tsx`, which also made both rungs unit-testable and took ~55 lines out of the web view.
- The order of sacrifice is written down in `Guidelines/Responsiveness` and `Architecture-Decisions.md`.

The third Studio #237 item (titlebar still cutting off) is not reproducible in paranext-core and is filed against Studio.

## Testing

- `npm run typecheck` clean; `npm run lint` clean (one pre-existing `import/no-named-as-default` warning in an unrelated utils test); full `npm test` green — 7939 tests.
- New unit coverage for each rung: the trigger's floor applied at `MINIMUM` and **not** at `TIGHTER`, chevron present/absent, and the container's padding at each step. Each was watched failing before the implementation existed.
- Two isolated E2E specs, one per width regime — `paragraph-style-trigger-narrow.spec.ts` shrinks the window, `paragraph-style-trigger-column-floor.spec.ts` drags the dock divider to reach widths no window resize can. Both assert geometrically, because a child clipped by an ancestor's `overflow` reports no overflow of its own. The column-floor spec was revert-tested against the pre-fix state.
- Storybook stories were added for `ParagraphStyleTrigger` and `ParagraphStyleLabel` so the shrink ladder is reviewable without a running app — see `AcrossShrinkSteps` and `MarkerWidthsAtMinimum`.
- Manual pass on macOS: measured the trigger across the ladder from a 664px column down to the 297px floor — marker whole at every step, no overrun, no toolbar items lost, no horizontal scrollbar.

Note for reviewers verifying the E2E specs: `npm run test:e2e:isolated` does **not** rebuild extensions, so run `npm run build:extensions` first or a stale bundle reports a false green.

## Risk Level

Low–Medium. `TabToolbarContainer`'s padding change is the only one that touches every web-view toolbar, and it is scoped to `SHRINK_STEP.MINIMUM`, so nothing changes above a 340px toolbar. It is safe to key off the step because `useShrinkStep` measures the container's **border** box, which its own padding does not change — it cannot feed back into which step is chosen, and the 520/420/340 thresholds are unaffected.

## Open questions

- **The dropped chevron is a UX-visible trade** and wants a UX nod: below 340px the trigger is a bordered marker code with no chevron. Popover semantics, `aria-haspopup` and the `aria-label` are untouched, so nothing changes for keyboard or screen-reader users.
- **The marker slot no longer reserves six characters**, overriding an earlier UX decision. Note this also leaves `Guidelines/Responsiveness` contradicting itself — it still says to reserve a fixed slot for variable-length content — so the guideline needs a carve-out or the code needs to change.
- **`Guidelines/Responsiveness` is amended here by the same PR that implements what it sanctions.** That guideline is UX-owned; please confirm sign-off rather than treating the edit as its own authorization.
- **Power mode has no `panelLock.minWidth` at all**, so a panel can be dragged below any floor and the trigger will still be clipped there. A general answer — toolbar controls overflowing into the floating menu — is bigger than this PR and is left open rather than half-built.

---

# Code Review Summary

**Branch**: pt-4466-toolbar-followups

**Base**: origin/main

**Date**: 2026-09-04

**Files changed**: 40

## Overview

This branch clears the follow-ups deferred by PT-4344 (PR #2701) and fixes the narrow-width toolbar
defects reported against Studio #237. All of it is Simple mode at or near its minimum widths: a
900px window, and a 297px editor column inside a window of any size. The PT-4344 half extracts
`WINDOW_MIN_WIDTH_PX` into a shared constant so `main.ts` and `simple-layout.data.ts` stop mirroring
`900` by hand, replaces the test-only `shrinkStep` prop on four components with an exported
`ShrinkStepOverride` context wrapper, and fixes model-text truncation that was inert because
`truncate` sat on a flex container. The narrow-width half establishes an order of sacrifice for a
toolbar that is out of room — the container yields its own padding, then a control yields its
chevron decoration, and only then does anything touch content, with the paragraph-style trigger
flooring at `min-w-min` once the marker is all that is left.

The branch was brought up to date with `origin/main` during this review (2 commits behind; the PR
was showing as CONFLICTING on GitHub). The only conflict was in the append-only
`Architecture-Decisions.md`, resolved by keeping both sides ordered by entry date. Analysis found
one Critical and two Important findings. Both Important findings were resolved with code changes
during the review; the Critical one was dismissed by the author, but with a caveat the reviewer
should be aware of (see below).

## API Changes

- `lib/platform-bible-react` (`ToolbarProps`): Removed optional `shrinkStep` prop
- `lib/platform-bible-react` (`TabToolbarProps`): Removed optional `shrinkStep` prop
- `lib/platform-bible-react` (`TabToolbarContainerProps`): Removed optional `shrinkStep` prop
- `lib/platform-bible-react` (`BookChapterControlProps`): Removed optional `shrinkStep` prop
- `lib/platform-bible-react`: Added exports `ShrinkStepOverride`, `ShrinkStepOverrideProps`,
  `ShrinkStepOverrideContext`, `useShrinkStepOverride`
- `lib/platform-bible-react`: `TAB_TOOLBAR_SHRINK_THRESHOLDS_PX` values unchanged (doc comment only)
- `lib/platform-bible-react/dist/`: regenerated — `index.d.ts` contains the four new symbols and no
  longer declares any `shrinkStep` prop, consistent with `src/`
- `src/shared/models/window-constraints.model.ts`: New internal module exporting
  `WINDOW_MIN_WIDTH_PX` (not a public API surface; verified absent from `papi.d.ts`)
- `extensions/src/platform-enhanced-resources/.../toolbar.component.tsx`: Removed `shrinkStep` from
  `EnhancedResourceTabBarProps` / `ToolbarProps` (component props, not in the extension's
  `types/*.d.ts`)
- `extensions/src/platform-scripture-editor/contributions/localizedStrings.json`: **New** key
  `%webView_platformScriptureEditor_paragraphSelection_ariaLabel%` (en + es), added during this
  review
- `lib/papi-dts/papi.d.ts`: unchanged; no `@papi/` module is affected

- [ ] ~~`shrinkStep` is deleted outright from four exported prop types of the published
  `platform-bible-react` package, with no deprecation window. Any out-of-repo extension passing it
  gets a TS excess-property error and silently loses the behavior.~~ _(Author's position: the
  create-and-remove turnaround was fast enough that this is acceptable. **However**, the ADR's
  stated justification is factually wrong and the author declined to correct it. The ADR claims "it
  appears in no tagged release ... and no out-of-repo consumer **could** have adopted it" — but the
  author confirmed paratext-10-studio checks out latest `main` when it makes a new build, so the
  prop **was** reachable by a consumer. The accurate claim is that none **did**: a grep across
  local checkouts of paratext-10-studio (most recent commit 2026-09-02),
  paratext-bible-extensions, paratext-bible-internal-extensions, the multi-extension template and
  the sample extensions found zero `shrinkStep` usages in source or dist. The outcome is safe; the
  written reasoning is not, and it is the kind of reasoning that gets reused to justify the next
  removal.)_

- [x] **FN-024 in `e2e-tests/tests/enhanced-resources/enhanced-resources-functional-UI-PKG-001.spec.ts`
  was converted from `test.fixme` to a live test without ever being watched run**, against the
  removed FIXME's own precondition ("Enable it once someone can watch it run rather than on the
  strength of the code reading alone"). The commit body for `e8abc2d` said "enabled but NOT verified
  green" while the PR description claimed "confirmed green" — a direct contradiction.
  _(fixed during review: the test was actually run. It **fails** — but at the setup step
  (`er-scripture-pane` not found), not at the FN-024 assertion. A sibling test in the same file that
  was never `fixme`'d fails at the identical line, confirming this is the environmental limitation
  the Playwright config documents: the ER suite needs real Marble resources (ESV16UK+) that are not
  present. FN-024 was therefore reverted to `test.fixme` with its original comment restored; that
  file now has a zero diff against the merge base and is out of the PR entirely.)_
- [x] **No Storybook stories for the two new components** (`paragraph-style-trigger.component.tsx`,
  `paragraph-style-label.component.tsx`), where every sibling toolbar control in the same directory
  has one. This mattered more than convention here: the PR has two open UX questions (the dropped
  chevron, the marker slot no longer reserving six characters) and amends a UX-owned guideline, and
  a story is the surface where UX would actually see those trades.
  _(fixed during review: added `paragraph-style-trigger.component.stories.tsx` and
  `paragraph-style-label.component.stories.tsx`, using the newly exported `ShrinkStepOverride` to
  pin each rung of the ladder. Stories cover WIDE/TIGHT/TIGHTER/MINIMUM with the chevron present vs
  dropped, `p` vs `q2` vs `toc1` marker widths at the narrowest step, the structure-protected
  disabled state, and the no-marker case.)_

### Minor — Consider

- [x] `paragraph-style-trigger.component.tsx`: `aria-label="Paragraph Selection"` was hardcoded
  English in title case — the one accessible name that would stay English in every locale, and
  against the sentence-case capitalization guideline. _(fixed during review: added
  `%webView_platformScriptureEditor_paragraphSelection_ariaLabel%` to `localizedStrings.json` in
  both en ("Paragraph style") and es ("Estilo de párrafo"), following the existing
  `shareLayout_ariaLabel` naming convention; added the key to
  `PARAGRAPH_STYLE_TRIGGER_STRING_KEYS`, which the web view already spreads. Updated the two
  component-test selectors and both e2e locators accordingly.)_
- [ ] **The marker slot is now content-sized, which contradicts `responsiveness.mdx` as written.**
  The guideline still says "Reserve a **fixed slot** for anything whose content varies in length,
  so switching projects or references cannot shift the items beside it," and the new
  "When the steps run out" section adds no carve-out. The trigger is last in `startAreaChildren`, so
  nothing else actually shifts, and the in-code comment makes that argument well — but the guideline
  and the code disagree on their face. **Deliberately left open: writing the carve-out is a UX
  decision, not a code fix.** This is the same change the PR describes as overriding the earlier
  "UX set six" decision.
- [ ] **`responsiveness.mdx` — a UX-owned design-system document — is amended in the same PR that
  implements the behavior it sanctions.** The `Tooltips` guideline states that new patterns "need
  agreement across the team and a formal addition to the style guide; ad-hoc implementations by
  individual contributors are not acceptable." Wants an explicit UX sign-off rather than treating
  the guideline edit as its own authorization. **Deliberately left open — needs UX, not code.**
  _(The double blank line at lines 112–113 was fixed during review.)_
- [x] `toolbar-compound-label.component.tsx` (otherwise unchanged): its `items-center` comment
  justified the choice by reference to "the paragraph label's fixed marker slot," which no longer
  exists, leaving the comment pointing at something a reader cannot find. _(fixed during review:
  reworded to state the `inline-block`/`overflow: hidden` baseline rule generically.)_
- [x] `shrink-step-override.component.tsx`: TSDoc did not state the nesting consequence — the
  override applies to the whole subtree and no toolbar resets it, so an app-code use pins every
  nested toolbar beneath it, not just the nearest. _(fixed during review: documented, with guidance
  to wrap the narrowest subtree that needs it.)_
- [ ] ~~`tooltipText={localizedStrings[KEY] ?? KEY}` renders the raw `%webView_...%` key to the user
  before strings resolve.~~ _(Assessed as a false positive during review: falling back to the key is
  the established, deliberately tested convention in this extension —
  `share-layout-button.component.test.tsx:87` has an explicit "falls back to the key text" test, and
  the same shape exists in platform-enhanced-resources. Changing it here would diverge from tested
  behavior. Instead, a local `localize()` helper was added mirroring `share-layout-button`'s, and
  both the tooltip and the new aria-label now go through it.)_
- [x] `platform-enhanced-resources/.../toolbar.component.tsx:509`: `useShrinkStepOverride()` was
  called inline inside a boolean expression, where every other hook in the file is bound to a named
  const first. _(fixed during review: hoisted to `shrinkStepOverride`.)_
- [x] `tab-toolbar-container.component.tsx`: conditional classes were concatenated into a raw
  template literal, where the sibling `toolbar.component.tsx` uses `cn()` — which also drops the
  literal `"undefined"` that lands in the class string when `className` is omitted. _(fixed during
  review: now uses `cn()`.)_
- [x] `paragraph-style-label.component.test.tsx`: backward-facing test comment ("The gap this guards
  against **was** a fixed six-character slot ... which is **the whole of the fix**") recorded what
  the previous implementation did rather than the invariant. _(fixed during review: restated as the
  invariant.)_
- [x] `paragraph-style-trigger-narrow.spec.ts`: the file header opened with `PT-4466 —`, an in-PR
  ticket reference, which the repo's forward-facing-comments rule puts in the commit message.
  _(fixed during review: removed, and the surrounding past-tense narration rewritten to present
  tense so the header describes the failure class rather than this PR's bug report.)_
- [x] `paragraph-style-label.component.tsx`: `MARKER_STYLE_SEPARATOR` was exported but has no
  importer anywhere in `extensions/src` or `lib/`. _(fixed during review: made module-private.)_
- [ ] Both new isolated e2e specs live in the `isolated` project, which is local-only and not wired
  into CI — so the pixel geometry they pin is not checked automatically. The class-name unit tests
  in `paragraph-style-trigger.component.test.tsx` do run in CI and cover the decision logic, so this
  is coverage depth rather than a gap. **Left open: wiring `isolated` into CI is well beyond this
  PR's scope.**

[Author response: The author asked for all findings to be fixed. Nine were fixed with code changes;
one was assessed as a false positive and resolved by conforming to the existing tested convention
instead; three were deliberately left open because they are not code fixes — two require a UX
decision and one requires CI infrastructure work.]

#### Shared Regions Modified

None. No `#region shared with <url>` markers exist in any changed file. (The `#region`/`#endregion`
pairs in `platform-scripture-editor.web-view.tsx`, `book-chapter-control.component.tsx` and
`main.ts` are plain code-folding regions with no repo URL.)

#### Extension Config Changes

None. No `package.json`, `tsconfig*.json`, `webpack.config.*` or `.eslintrc*` under `extensions/`
was changed — only source, test, and localized-string contribution files.

## Positive Observations

- The two-context split is well motivated and the reasoning is recorded where it belongs:
  `ShrinkStepOverrideContext` defaults to `undefined` precisely so a *publisher* can distinguish
  "nobody is overriding me" from "someone pinned me at the widest step", which `ShrinkStepContext`'s
  `SHRINK_STEP.WIDE` default cannot express. That is a real bug avoided, not just a refactor.
- The replacement API is strictly better than what it removes: one wrapper replaces four per-component
  props, and it works whether or not a toolbar sits between it and the component under test.
- `WINDOW_MIN_WIDTH_PX` converts a hand-mirrored literal plus a "change both" comment into a
  build-enforced coupling. `simple-layout.data.test.ts` now fails if the window minimum drops, which
  is exactly what that test existed to catch and previously could not.
- The `isTightened` comment's claim that padding cannot feed back into step selection is accurate —
  `useShrinkStep` observes the border box, confirmed in its implementation and docstring.
- `model-text-panel.component.tsx`: moving truncation from the flex container onto the inner span is
  a real fix, not a cosmetic one — `text-overflow` is inert on a flex container, so the header was
  hard-clipping mid-glyph with its tooltip permanently shut.
- `PopoverContent` capped with `--radix-popover-content-available-width` rather than a fixed `w-96`
  is the right call for the 300px web-view minimum.
- Three ADR entries recorded for genuinely cross-cutting decisions, each with rejected alternatives
  and their reasons — and the ladder rule was promoted into `responsiveness.mdx`, the standard
  readers actually consult, rather than left in the log.
- The `papi.d.ts`-pollution question for `src/shared/` was settled empirically (regenerate, diff)
  rather than by assumption.
- Deleting `marker-slot-width.test.ts` was correct: it pinned a `tw:w-[6ch]`/`MARKER_SLOT_CHARACTERS`
  pairing that no longer exists, so keeping it would have been a vacuous test.
- Test coverage for the new logic is genuinely good — all four rungs of the ladder, the chevron
  shown/dropped, the empty-and-undefined marker, and the container's padding at each step, with a
  comment explaining why jsdom forces structural rather than visual assertions.
- `dist/` was regenerated rather than hand-patched, and matches `src/` exactly.

## Interview Notes

**Stated purpose**: clear the PT-4344 follow-ups and fix the three narrow-width defects reported
against Studio #237, all of them in Simple mode at or near its minimum widths. Part of the current
epic — PT-4466 is a sub-task of PT-4336 "Saroj and Donna can trust the app".

**Design explanation (Step 3.3) — the step-dependent floor.** The author was asked to explain the
most non-obvious part in their own words and gave a complete, mechanistically correct account,
unprompted by any hint from the analysis:

> `min-width: min-content` on a flex item resolves to the width of the longest unbreakable word
> inside it. While the style name is rendered, the longest word in something like "Table of Contents
> 1" is the operative `min-content` size, not the marker — far wider than the zone can spare. The
> button then refuses to shrink to that width, so the zone's `overflow-clip` slices its trailing
> border instead, and the style name never gets to ellipsis because the button never goes narrow
> enough to need one. At `SHRINK_STEP.MINIMUM` the style name has already been dropped and only the
> marker remains; USFM markers are single codes with no break opportunity, so `min-content` is now
> exactly the marker's width — which is precisely the point where nothing shorter exists, so
> stopping there is correct.

The author also explained correctly why the control had to leave the web view: a hook called in the
web view's body sits *above* `TabToolbarContainer`'s provider in the tree and would read the widest
step unconditionally, so `isFloored` could never be true. **Both halves were spot-verified during
the review** — `TabToolbarContainer` publishes at `tab-toolbar-container.component.tsx:119`, and the
web view renders `<ParagraphStyleTrigger>` (line 3784) inside `<TabToolbar>` (line 3765), making it
a descendant of the provider. The subtlety the author's reasoning gets right is that React context
resolves by *render-tree* position, not by where the element is created.

**Understanding assessment**: the author demonstrated clear, independent command of the most
intricate part of this change. No areas were deferred to AI and no uncertainty was expressed. This
is a notably strong showing on a subtle CSS-plus-React-context interaction.

**Unresolved / needs reviewer attention**: the one open thread is the Critical finding's *reasoning*
rather than its outcome. When asked whether any consumer tracks `main` rather than published tags,
the author confirmed that paratext-10-studio checks out latest `main` for each new build — which
directly contradicts the ADR's "no out-of-repo consumer could have adopted it". Asked whether to
correct the ADR wording, the author said "No let's move on". The removal itself is safe (verified by
grep across five consumer repos), but the ADR entry now carries a justification that is not true.

## In-Review Quality Check

All gates green after the in-review changes:

- `npm run build:platform-bible-react` — exit 0. Rebuilt and staged `lib/platform-bible-react/dist/`
  (required, since `lib/platform-bible-react/src/` changed and this repo commits `dist/`).
- `npm run typecheck` — exit 0, no errors in any workspace.
- `npm run lint` — exit 0, **zero errors repo-wide**. One pre-existing, unrelated warning remains
  (`import/no-named-as-default` in `lib/platform-bible-utils/src/lifetime-management/unsubscriber-async-list.test.ts`),
  not touched by this branch.
- Targeted prettier over the changed files — exit 0; two files reflowed.
- `npm test` — exit 0. **574 test files passed, 7708 tests passed, 2 skipped, 0 failed.**

No fixes beyond formatting were needed; nothing was broken by the review changes.

**Branch sync**: brought up to date with `origin/main` during this review (was 2 behind; PR showed
CONFLICTING). The sole conflict was in the append-only `Architecture-Decisions.md` and was resolved
by keeping both sides, ordered by entry date — main's `adr-registration-validity-once-per-session`
(2026-08-22) ahead of this branch's three entries (2026-08-26, 2026-09-03).

## Suggested Review Focus

- [ ] **The ADR's justification for the breaking `shrinkStep` removal is factually wrong.**
  `adr-shrink-step-override-context` says no out-of-repo consumer *could* have adopted the prop;
  Studio's `main`-tracking builds mean it could have. Nothing did (grep-verified), so the removal is
  safe — but decide whether the entry should be corrected before it is cited as precedent for the
  next same-day API removal. The author chose not to change it.
- [ ] **Two UX decisions are pending and cannot be settled in code** — (a) the marker slot is now
  content-sized, contradicting the "reserve a fixed slot" rule that `responsiveness.mdx` still
  states, and overriding an earlier "UX set six" decision; (b) the chevron is dropped below 340px,
  leaving a bordered marker code with no dropdown affordance. New Storybook stories were added
  during this review specifically so UX can see both trades — point them at
  `ParagraphStyleTrigger → AcrossShrinkSteps` and `MarkerWidthsAtMinimum`.
- [ ] **`responsiveness.mdx` is a UX-owned document amended in the same PR that implements what it
  sanctions.** The `Tooltips` guideline says new patterns need team agreement and a formal style-guide
  addition. Confirm this had UX sign-off rather than being self-authorized by the diff.
- [ ] **FN-024 remains deferred.** It was enabled in this branch, found failing during this review,
  and reverted to `test.fixme`. The underlying question — can the enhanced-resources suite be run by
  anyone, given it needs real Marble resources — is still open and is why the FIXME exists.
- [ ] `TabToolbarContainer`'s padding change is the one edit touching every web-view toolbar. It is
  scoped to `SHRINK_STEP.MINIMUM` and the no-feedback argument checks out, but it is the widest
  blast radius in the diff.
- [ ] **Power mode has no `panelLock.minWidth`**, so a panel can still be dragged below any floor and
  the trigger clipped there. The author left this open rather than half-building a general answer
  (toolbar controls overflowing into the floating menu); confirm that deferral is the right call.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2768)

pt-4466-before-after


katherinejensen00 (changes requested):
# PR #2768 — Blocking and Important findings

PT-4466: PT-4344 follow-ups and narrow-width toolbar fixes  
Reviewed commit: `62138e9ead0` (current PR head)

* * *

## Blocking

### 1\. The editor's tab toolbar loses `display: flex` and becomes `display: block`

**File:** `lib/platform-bible-react/src/components/advanced/tab-toolbar/tab-toolbar-container.component.tsx:123`  
**Severity:** Blocking

Switching the container's `className` from string concatenation to `cn()` makes tailwind-merge resolve  
conflicts last-wins, and the consumer `className` is now the last argument. `platform-scripture-editor.web-view.tsx:3769`  
passes `tw:block`, forwarded through `tab-toolbar.component.tsx:70`. I ran the repo's own merge on the  
real strings and `tw:flex` is dropped while `tw:block` survives:

```
$ node -e "const m=require('tailwind-merge').extendTailwindMerge({prefix:'tw'}); console.log(m('tw:flex tw:flex-row tw:items-center tw:gap-2 tw:px-4 tw:w-full tw:border-b tw:block tw:z-10'))"
tw:flex-row tw:items-center tw:gap-2 tw:px-4 tw:w-full tw:border-b tw:block tw:z-10
```

`tw:flex-row`, `tw:items-center`, `tw:justify-between` and every gap go inert; the zone divs stack  
vertically inside `tw:h-14` with `tw:overflow-clip`. No SCSS restores it — `.scripture-editor-tab-nav`  
sets no `display`. Before this PR both classes reached the DOM and the cascade decided; now `tw:flex`  
is gone from the attribute entirely, so the outcome is unambiguous regardless of what won before.  
Neither new E2E spec catches it, because a block-level start zone spans the full column and the  
trigger can never overrun it. Please check this in a running app before anything else.

### 2\. `paragraph-style-trigger-narrow.spec.ts` never places a caret, so its first assertion cannot pass

**File:** `e2e-tests/tests/isolated/scripture-editor/paragraph-style-trigger-narrow.spec.ts:50`  
**Severity:** Blocking

The spec waits for `.editor-container`, then goes straight to `expect(trigger).toBeVisible({ timeout: 60_000 })`.  
`ParagraphStyleTrigger` returns `undefined` while `!blockMarker` (`paragraph-style-trigger.component.tsx:93`),  
and `blockMarker` is set only from the editor's `onStateChange` (`web-view.tsx:3637`), which fires only  
inside a range selection. `openEditableScriptureEditorForProject` (`scripture-editor-helpers.ts:194`)  
just delegates to `openEditorViaCommand` and contains no click into the editor surface, and the main  
editor mounts no `AutoFocusPlugin`. Your sibling column-floor spec clicks a `[data-marker="q2"]` block  
first and adds a `toContainText` positive control for exactly this reason. Your PR description says the  
column-floor spec was revert-tested but makes no green claim for this one — that fits. Add the click  
and a positive control.

* * *

## Important

### 3\. A 5–7 character USFM marker can reconsume the entire margin the fix bought

**File:** `extensions/src/platform-scripture-editor/src/paragraph-style-trigger.component.tsx:87`  
**Severity:** Important

`blockMarker` is read verbatim off the USJ para node at the caret, not from the switcher menu —  
`blockMarkerToBlockNames` (`platform-scripture-editor.utils.ts:503`) tops out at `toc1` and is itself  
commented "This list is incomplete", and `web-view.tsx:3748` already branches on markers outside it.  
`c-sharp/assets/usfm.sty` defines `periph` (:3074), `restore` (:185) and `pubinfo` (:3453) as  
paragraph styles (`StyleType Paragraph`), so the real ceiling is **7 characters**.

At `MINIMUM`, `min-w-min` resolves to the button's padding plus the full mono string (the primary field  
is `truncate`/nowrap, so it contributes its whole width to `min-content`). At `text-sm` with Tailwind's  
default mono stack that is ~7.7–8.4px/ch:

| marker | chars | approx width |
| --- | --- | --- |
| `toc1` | 4 | 31–34px |
| `periph` | 6 | 46–50px |
| `restore` | 7 | 54–59px |

That is against the ADR's stated ">=16px spare". `min-width` beats `max-width`, so the floor re-creates  
the exact overrun `tw:max-w-full` exists to prevent. The code you deleted named both cases explicitly  
(`MARKER_SLOT_CHARACTERS = 6` … "(`periph`)", and `isMarkerClipped` citing a clipped `restor`). Nothing  
in the repo now knows a marker can be 7 characters. Worth measuring `restore` at the 297px floor.

### 4\. The headline `tw:max-w-full` fix has no test that CI runs

**File:** `extensions/src/platform-scripture-editor/src/paragraph-style-trigger.component.tsx:118`  
**Severity:** Important

CI runs `npm test` and `test:e2e:smoke` only (`.github/workflows/test.yml:176,189`); `test:e2e:smoke`  
is `--project=smoke`, and `isolated` is a separate project. There is no Storybook test-runner script.  
Grepping the whole non-`dist` tree, the only `max-w-full` hits are this line and its comment — no unit  
test, story-as-test or snapshot asserts it. The PR description calls this "coverage depth rather than a  
gap"; that isn't right — the CI unit tests cover `min-w-min`/`min-w-0` and the chevron, not the cap.  
One more `expect(button.className).toMatch(/tw:max-w-full/)` in the existing file closes it.

### 5\. `keeps its full padding and gaps at every wider step` passes for the wrong reason

**File:** `lib/platform-bible-react/src/components/advanced/tab-toolbar/tab-toolbar.component.test.tsx:130`  
**Severity:** Important

That file installs no `ResizeObserver` stub, so `useShrinkStep` early-returns  
(`use-shrink-step.hook.ts:88`) and `step` stays 0. Add any stub — as three other test files in this same  
PR do, and as the sibling `advanced/toolbar.component.test.tsx:99` already does — and jsdom reports  
width 0, `getShrinkStep(0, [520,420,340])` returns 3, `isTightened` flips, and the test goes red. It  
exercises the unmeasured default, not "every wider step". Pin it with  
`ShrinkStepOverride value={SHRINK_STEP.TIGHTER}` instead.

### 6\. The narrow spec resizes with a CDP emulation override the repo's own fixtures document and avoid

**File:** `e2e-tests/tests/isolated/scripture-editor/paragraph-style-trigger-narrow.spec.ts:69`  
**Severity:** Important

`e2e-tests/fixtures/helpers.ts:185` states it outright: "`setViewportSize()` on a CDP-attached page  
applies an emulation override that sets `innerWidth` … Measured: a 1024px window reports `innerWidth`  
469 before such a call and 1280 after." The repo supplies `applyDeclaredWindowSize()`, and  
`title-bar-narrow-width.spec.ts` — the sibling your header names as its model — uses a real `setSize`  
with a settle poll. Consequently 1100/1000/950 bypass Electron's `minWidth: 900`, so the header's claim  
that it "bottoms out at `WINDOW_MIN_WIDTH_PX`" is not true of what the spec actually does.

> @katherinejensen00 reviewed 40 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2768#-:-P0hlVjy3lrRkm1vJ5BB:bym5jiz)* status: all files reviewed, 1 unresolved discussion (waiting on jolierabideau).

___
*[`lib/platform-bible-react/src/components/advanced/tab-toolbar/tab-toolbar-container.component.tsx` line 123 at r1](https://reviewable.io/reviews/paranext/paranext-core/2768#-P0hlkRe5bkHtCQkwI5G:-P0hlkRe5bkHtCQkwI5H:b-q2txps) ([raw file](https://github.com/paranext/paranext-core/blob/4c7fe3f56e1f1d60cbd0a4953499cef3abcf47fa/lib/platform-bible-react/src/components/advanced/tab-toolbar/tab-toolbar-container.component.tsx#L123)):*
> ```XML
>         <div
>           ref={attachRoot}
>           className={cn(
> ```

### 1\. The editor's tab toolbar loses `display: flex` and becomes `display: block`

**File:** `lib/platform-bible-react/src/components/advanced/tab-toolbar/tab-toolbar-container.component.tsx:123`  
**Severity:** Blocking

Switching the container's `className` from string concatenation to `cn()` makes tailwind-merge resolve  
conflicts last-wins, and the consumer `className` is now the last argument. `platform-scripture-editor.web-view.tsx:3769`  
passes `tw:block`, forwarded through `tab-toolbar.component.tsx:70`. I ran the repo's own merge on the  
real strings and `tw:flex` is dropped while `tw:block` survives:

```
$ node -e "const m=require('tailwind-merge').extendTailwindMerge({prefix:'tw'}); console.log(m('tw:flex tw:flex-row tw:items-center tw:gap-2 tw:px-4 tw:w-full tw:border-b tw:block tw:z-10'))"
tw:flex-row tw:items-center tw:gap-2 tw:px-4 tw:w-full tw:border-b tw:block tw:z-10
```

`tw:flex-row`, `tw:items-center`, `tw:justify-between` and every gap go inert; the zone divs stack  
vertically inside `tw:h-14` with `tw:overflow-clip`. No SCSS restores it — `.scripture-editor-tab-nav`  
sets no `display`. Before this PR both classes reached the DOM and the cascade decided; now `tw:flex`  
is gone from the attribute entirely, so the outcome is unambiguous regardless of what won before.  
Neither new E2E spec catches it, because a block-level start zone spans the full column and the  
trigger can never overrun it. Please check this in a running app before anything else.

jolierabideau:
All six findings verified against the code before changing anything. Five are fixed; finding 3 is real but deferred, with the reasoning below. Branch is also now merged up to date with `main`.

---

## 1. Toolbar loses `display: flex` — fixed, and verified in a running app

Confirmed, and worse than DOM-visible: I reproduced it live. In the running editor I read the container's computed style and its three zone boxes under both class lists:

| class list | `display` | zone boxes (x, y, width) |
| --- | --- | --- |
| pre-fix (`cn()` with the consumer's `tw:block`) | `block` | (16, **8**, 32) · (16, **40**, 1856) · (16, **72**, 1856) |
| fixed | `flex` | (16, 12, 32) · (56, 12, 1636) · (1700, 14, 172) |

The zones stack at y = 8/40/72 inside a 56px `overflow-clip` row, so the end zone lands entirely outside the box.

One correction to add to your analysis: `tw:block` was **never** doing anything. `.tw\:block` is emitted *before* `.tw\:flex` in the generated CSS, so `flex` won the cascade before this PR — `tw:block` has been inert cruft since the Tailwind 4 upgrade (74a7dd554). So there is no behavior to preserve.

Two changes:

- Removed the dead `tw:block` from `platform-scripture-editor.web-view.tsx`.
- Made the flex row non-overridable rather than just fixing this one consumer: `tw:flex tw:flex-row` now merges *after* `className` in `cn()`. The zones are flex items that shrink and grow against each other, so `display` is structure the container owns, not a default a consumer should be able to take away. Alignment, gaps and padding stay overridable as before.

Backed by a test that fails on the old code with exactly your symptom:

```
× keeps its flex row when a consumer passes a conflicting display class
  → expected 'tw:sticky tw:top-0 tw:box-border tw:h…' to match /(?:^|\s)tw:flex(?:\s|$)/
```

## 2. Narrow spec never places a caret — fixed

Correct on every step: `ParagraphStyleTrigger` returns `undefined` while `!blockMarker`, `openEditableScriptureEditorForProject` only delegates to `openEditorViaCommand`, and the editor mounts no `AutoFocusPlugin`. The spec would have timed out on an element never in the DOM.

Now navigates to Genesis 3:14, clicks a `[data-marker="q2"]` block, and adds the `toContainText` positive control — the same shape as the column-floor sibling. It passes:

```
✓ paragraph-style-trigger-narrow.spec.ts:50:7 › keeps its whole box inside the toolbar zone
  instead of being clipped (38.0s)
```

## 3. 5–7 character markers can reconsume the margin — confirmed, deferred, documented

You are right about the ceiling, and I can tighten the number. Parsing `c-sharp/assets/usfm.sty` for `StyleType Paragraph` gives **120 markers**; the longest are `pubinfo` and `restore` at 7 characters, then `periph` at 6. The ADR's "≥16px spare" was measured against `toc1`, and `restore` is ~23px wider than `toc1` in that monospace field — so the residual overrun is roughly 7–9px, reachable only with the caret in peripheral or front-matter material at the column floor.

I have **not** fixed it here. The ADR's next rung is hiding undo/redo at `SHRINK_STEP.MINIMUM` — the only option with real margin (~68px) — and it was rejected "for now" because it costs two discoverable controls and needs UX sign-off. That has not changed, and taking it would widen this PR into a UX decision. Pushing back on fixing it *in this PR*, not on the finding.

What I did instead, so it stops being invisible:

- The 7-character ceiling and its consequence are documented at the floor's definition site in `paragraph-style-trigger.component.tsx`, including why `blockMarkerToBlockNames` does not bound it.
- `adr-narrow-toolbar-yields-padding-then-decoration` now carries an **Amended 2026-09-04** entry recording that its "every marker length fits" claim is wrong, by how much, and that measuring `restore` at the 297px floor is the next step. The claim is no longer standing as fact.

Happy to take the undo/redo rung in a follow-up if you'd rather it not wait for UX.

## 4. No CI-run test for `tw:max-w-full` — fixed

Your read of CI is right; "coverage depth rather than a gap" was wrong of me. `paragraph-style-trigger.component.test.tsx` now asserts the cap at all three steps (default, `TIGHTER`, `MINIMUM`) — every step, since unlike the floor and the chevron the cap is not step-dependent.

## 5. `keeps its full padding and gaps at every wider step` — fixed

Confirmed: no `ResizeObserver` stub, so `use-shrink-step.hook.ts:93` early-returns and `step` stays 0. It was testing the unmeasured default. Pinned with `ShrinkStepOverride value={SHRINK_STEP.TIGHTER}` as you suggested, so it tests the boundary rather than the absence of a layout engine.

## 6. `setViewportSize` on a CDP page — fixed

The spec now uses a real `BrowserWindow.setSize` with a settle poll. Rather than copy the private helper out of `title-bar-narrow-width.spec.ts`, I extracted `setWindowWidth` and `closeDevTools` into `fixtures/helpers.ts` and pointed both specs at them — one implementation, next to `assertDeclaredWindowSize`, whose measurements document exactly why `setViewportSize` is wrong here. `title-bar-narrow-width.spec.ts` still passes on the shared version (`2 passed (38.1s)`), and I corrected the narrow spec's header, which no longer claims anything about the clamp that its widths do not exercise.

---

### Unrelated observation

The column-floor spec (still green, 50.8s) logs a `ModelTextPanelWebView` crash caught by the new `WebViewErrorBoundary`:

```
TypeError: Cannot read properties of undefined (reading 'webViews')
```

That is `useWebViewState` → `this.papi.webViews` racing startup, in `src/renderer/hooks/` — untouched by this branch, last changed in #1969. My guess is it has been happening all along and main's new error boundary is simply the first thing to report it instead of blanking the pane. Flagging rather than folding into this PR.

katherinejensen00 (approved):
:lgtm:

> @katherinejensen00 reviewed 20 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2768#-:-P10XWmr1leP8RIaw_gL:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2769, opened 2026-09-04
Close a tab on middle-click of its header

rolfheij-sil:
## Summary

A middle click (scroll-wheel click) anywhere on a tab's header now closes that tab, as in browsers and editors. This works in the main tab bar and in the overflow dropdown. Tabs that can't be closed ignore it, just as they have no close button. A middle press on a tab bar no longer starts a drag.

## How it works

- `installMiddleClickTabBarHandlers` (`platform-dock-layout-middle-click-handlers.util.ts`) adds a capture-phase `mousedown` listener to the document. It cancels a middle press anywhere in a `.dock-bar` or `.dock-dropdown`. It runs before rc-dock's React handlers, so rc-dock never starts a drag with the middle button (rc-dock does that for every button except the right one). Cancelling the press also stops autoscroll on Windows and primary-selection paste on Linux. The listeners are on the document because rc-trigger portals the overflow dropdown into `document.body`.
- The tab closes on `auxclick`, the click rather than the press. `auxclick` targets the common ancestor of the press and the release, so a press dragged off the tab closes nothing.
- The tab id comes from `data-tab-header-id`, which `PlatformTabTitle` sets on its root. The lookup starts at rc-dock's `[role="tab"]` wrapper, so the title, the close button and the hit area all resolve to the tab. `data-tab-id` still marks only `PlatformPanel`, which the focus lookup relies on.
- Whether the tab can close is read at click time from rc-dock's `closable` (`dockLayout.find(id)`), the same flag that controls the X button. An unset flag counts as not closable. The close goes through `closeTab`, and a failure is logged rather than rejected.
- Closing a tab slides its neighbour under the pointer. To avoid closing that neighbour by accident, one repeat middle click within 500 ms and 4 px (on each axis) of a close is ignored. After that, clicks close tabs normally again.
- `getTabInfoByElement` is deliberately not used to resolve the tab. For a click on the empty strip it falls back to the panel's active tab, and it can't resolve headers in the overflow dropdown. This PR also leaves that function's behaviour the same as on `main`.

## Accessibility

A middle click has no keyboard equivalent. Today there is no keyboard or screen-reader way to close a tab at all: rc-dock's X is a non-focusable `div`, the tab menu has no Close item, and there is no close shortcut. That gap predates this PR and is tracked in PT-3034.

## Notes for reviewers

- There is no Jira ticket for this change. It came up while working on the tab-bar drop target (#2767) and was split into its own branch off `main`, so either PR can merge first.
- `handleCloseTab` is the third copy of the "call `closeTab`, log either failure" shape; the other two are in `dialog.service-shard.ts`. It logs a missing tab with `logger.warn` rather than `logger.error` on purpose: a middle click on a tab that is already gone is normal use, while a dialog failing to close is a real fault.
- The `isClosable` TSDoc on `TabInfo` and `WebViewDefinitionBase` now mentions middle click, and `papi.d.ts` is regenerated to match. Only TSDoc text changed; there is no new API. A stale comment in `default-layout-supplement.util.ts` about where an absent `isClosable` becomes closable is also corrected.
- The `docking/` line in `e2e-tests/tests/isolated/README.md` overlaps with #2767's. Whichever PR merges second will get a one-line conflict; keep the longer line, which covers both specs.
- There are no new user-visible strings, so no localization keys were added. There is no new keyboard handler, so the keyboard shortcuts catalog is unchanged.

## Verification

- Unit tests: 26 for the handlers util, run against a hand-built DOM. They cover drag blocking, closing (including in the overflow dropdown), the closability gate, the repeat-click guard and cleanup. A further 6 wiring tests in `platform-dock-layout.component.test.tsx` call the captured callbacks through the real `closeTab`, and 1 tab-title test covers the header attribute.
- Contract test (`platform-dock-layout-middle-click.dom-contract.test.tsx`): renders a real rc-dock `DockLayout` and checks focus, drag blocking and closing (11 cases). Each of these mutations turns it red: dropping `capture: true`, narrowing the drag scope to `.dock-dropdown`, and ignoring `closable`.
- Playwright (`e2e-tests/tests/isolated/docking/middle-click-close-tab.spec.ts`): 3 of 3 pass. The spec covers:
  - a middle click closing a tab, with `auxclick` still firing after the cancelled press;
  - a middle press-drag starting no drag, with a left drag as the control;
  - the repeat-click guard, which ignores one repeat and lets a click 700 ms later through.

  With the guard window set to 0, the guard test fails.
- `build:main`, `typecheck`, `lint` and `format:check` are clean. The core suite passes (308 files, 4939 tests), and so does `platform-bible-react` (167 files, 1632 tests).
- Live in the dev app:
  - Opening About puts focus in its panel.
  - A single middle click closes a tab.
  - A middle press-drag starts no drag, while a left drag still does.
  - Before the guard, a quick second middle click closed the tab that slid under the pointer at 150, 300 and 600 ms, and the X button behaved the same way.
  - After any close, with the X or a middle click, focus goes to the page body instead of the new active tab. The X button does the same, so this predates this PR.

AI-assisted — [session 1](https://claude.ai/code/session_01MRwNSFZupZcGz8qLtdFQ9G), [session 2](https://claude.ai/code/session_01YcEf4fm6DK13Sviva2SrHg), [session 3](https://claude.ai/code/session_018CaYNvoByux4DtDVZJ6NK1)

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2769)


Sebastian-ubs:
None of my browsers (Firefox, Chrome), nor vs code closes on pressing the mousewheel. I am not sure we should do this

Sebastian-ubs:
Standard keyboard shortcut is `ctrl` + `w`, but PT9 uses this for launching the Wordlist

rolfheij-sil:
Are you sure? It works in all my browsers, terminals, VS Code, et cetera.
I use it heavily, so it would at least give me a better Developer Experience ;)
And fun fact: It works in PT9 too

Sebastian-ubs:
Interesting. Maybe you or I have some special mouse or OS setting 🤷
(or my mousewheel click is broken or not implemented ;-)

Sebastian-ubs:
indeed my mouse's wheel seems to be dying ;-)
Nevermind. Well done Rolf!

katherinejensen00 (changes requested):
# Code review — PR #2769 "Close a tab on middle-click of its header"

Repo: paranext/paranext-core · Author: rolfheij-sil · Branch: `middle-click-close-tab`
Head: `71d5a0a3e65` · Merge-base with `main`: `7ce28a6ddd8` · 12 files, +621 / -17
URL: https://github.com/paranext/paranext-core/pull/2769

23 findings: 1 High, 3 Medium, 10 Low, 9 Nit. Each is written for the PR author. Anchors are
repo-relative with the line number in the PR-head version of the file; every anchor is an added or
changed line, so GitHub will accept it as an inline comment.

## How this review was produced

Six passes, each reading the code and the rc-dock 3.3.2 / rc-tabs 11.16.1 / rc-trigger sources
rather than trusting the PR description:

1. Developer pass — soundness and tests.
2. Architect pass — layering, integration point, close path, public surface.
3. Comments and documentation pass — accuracy, length, density.
4. Antagonistic pass — tried to refute every finding above, corrected severities and line anchors.
   It dropped 3 findings as preference, cut one from High to Low, and disproved the claim that
   `handleCloseTab`'s location forces the new test mocks.
5. Correctness sweep (Sonnet).
6. Reuse, simplification, efficiency and conventions sweep (Sonnet).

`/code-review max` was requested and started, but died partway through on the account's weekly Opus
rate limit. Its reuse finder independently confirmed finding 1 and its conventions finder flagged
finding 13 before the verifiers were killed. Passes 5 and 6 were run on Sonnet to cover its
correctness and reuse angles; its conventions angle overlaps findings 8, 13 and 16-20.

---

## High

### 1. Middle-click's new `data-tab-id` takes over an existing focus lookup, sending keyboard focus to the tab strip instead of the tab's content

`src/renderer/components/docking/platform-tab-title.component.tsx`
Line: 862 — `data-tab-id={id}`
Severity: High

Before this PR, `data-tab-id` marked exactly one element: `PlatformPanel`, the tab's content
wrapper (`platform-panel.component.tsx:34`). `setDocumentFocusToTab` finds it with a document-wide
`document.querySelector([data-tab-id='...'])` and focuses its first child
(`platform-dock-layout-storage.util.ts:1340-1342`). rc-tabs renders the tab bar before the panels
(`rc-tabs/es/Tabs.js:226`), so that query now returns this title div, and `children.item(0)` is the
icon span at line 865 — it gets a temporary tabIndex, takes focus, and the panel content never
does.

This fallback runs on every tab mount, every tab click, and the focus-next step after any close,
because `setFocus` always calls `focusTab` (`window.service-shard.ts:500-510`). It affects every tab
with no `lastFocusedElement`: Settings, About, dialogs and error tabs always, plus web views, whose
iframe lookup at `:1286` is already shadowed by the title's pre-existing `data-web-view-id`.
Middle-click makes background closes common, so this fires more often than before. In Simple mode's
headless columns the focus ring lands inside a bar the repo keeps at `opacity: 0`, so it is
invisible. Nothing catches it: no test covers `setDocumentFocusToTab`.

Suggested fix: give the header its own attribute, e.g. `data-tab-header-id`, and read that in
`readClosableTabId`; that leaves both existing readers untouched. Scoping the lookup to
`.platform-panel[data-tab-id=...]` also works. Do not resolve the id via `getTabInfoByElement`
instead — see finding 23. Worth a test asserting only one `[data-tab-id=x]` exists per tab.

Verified by three independent passes; mechanism confirmed in source, not reproduced in the running
app.

---

## Medium

### 2. The rc-dock DOM details this feature depends on are only tested against a hand-built tree, so an rc-dock upgrade could silently disable it with every test green

`src/renderer/components/docking/platform-dock-layout-middle-click-handlers.util.test.ts`
Line: 15 — `function buildDockLayoutTree() {`
Severity: Medium

The handlers depend on `role="tab"` on rc-dock's per-tab DragDropDiv, the `.dock-bar` and
`.dock-dropdown` class names, and the overflow dropdown portalling to `document.body`. All are
correct today, but the fixture asserts them against itself. The repo already has the guard pattern:
`simple-layout.dom-contract.test.tsx` renders a real `DockLayout` for exactly this reason, and its
own doc comment cites silent breakage with zero failing tests. One test that renders a real
`DockLayout`, dispatches middle `mousedown`/`auxclick` on a real header, and asserts no drag arms
and the close callback fires would cover this and pin finding 1 at the same time.

### 3. The PR description still describes the implementation you replaced, and it will land in history on squash-merge

PR body (no line anchor)
Severity: Medium

- "How it works" says `PlatformTabTitle` gains `mousedown` and `auxclick` handlers on its root with
  a justified ESLint disable. There is no such handler or disable.
- The Update section names `platform-dock-layout-middle-click-guard.util.ts` and a listener on
  `getRootElement()` scoped to `.closest('.dock-bar')`. The code has `...-middle-click-handlers.util.ts`
  listening on `ownerDocument`, scoped to `.dock-bar, .dock-dropdown`.
- "Middle-clicking cannot start a drag ... past a threshold" is contradicted by your own later
  commit, and the threshold is 1px (`DragManager.js:40`).
- "the two files it touches" is now 12.
- Test counts are stale: the PR adds 23 tests (13 + 2 + 6 + 2).
- The note about removing a vacuous test is now wrong — that case is back, meaningfully, at
  `util.test.ts:222`.
- "the same public API that dialogs already use" overstates it: `closeTab` is a renderer-internal
  service-shard function, not in `papi.d.ts`, and `getRootElement` is `/** @ignore */`
  (`DockLayout.d.ts:97-98`).

A short current-state rewrite is worth more than patching sentences.

### 4. The PR description's accessibility claim is false, and the code's own comment says the opposite

PR body (no line anchor)
Severity: Medium

"closing remains reachable from the keyboard and screen readers through rc-dock's own tab and close
controls" does not hold: rc-dock's close button is a non-focusable `<div onClick>`
(`DockTabs.js:155`), rc-tabs only renders a focusable remove button when `editable` is passed and
rc-dock never passes it, the tab menu has no close item, and no close shortcut exists. The util's
TSDoc (lines 58-61) states this correctly. The gap is pre-existing, not caused by this PR, but the
body should say so — and it is worth a follow-up ticket rather than deleting the fact from the code.

---

## Low

### 5. Closability now has two sources of truth that only agree by convention

`src/renderer/components/docking/platform-tab-title.component.tsx`
Line: 863 — `data-tab-closable={isClosable}`
Severity: Low

rc-dock's X button reads `TabData.closable`; the middle-click gate reads `data-tab-closable`, fed by
the new prop, which defaults to `true`. They cannot diverge today — `PlatformTabTitle` has one render
site and nothing else writes `closable` — but the prop's TSDoc has to ask callers to keep them in
sync, which is the smell. Resolving the id from the DOM and reading closability from rc-dock would
remove the prop, the attribute and five tests:
`const tab = dockLayoutRef.current.find(tabId); if (isTab(tab) && tab.closable) ...`. `isTab` exists
(`docking-framework-internal.model.ts:14`) and `find()` returns the loaded TabData, the same pattern
already used at `platform-dock-layout.component.tsx:193-197`.

### 6. The installer takes a root element but listens on the whole document, so its signature and the test name both describe a narrower scope than it has

`src/renderer/components/docking/platform-dock-layout-middle-click-handlers.util.ts`
Line: 68 — `rootElement: HTMLElement,`
Severity: Low

Only `rootElement.ownerDocument` is used (line 71), and both listeners act on every
`.dock-bar`/`.dock-dropdown` in the document. Harmless with one dock layout per window, but the test
at `platform-dock-layout.component.test.tsx:90` is titled "installs ... on the dock layout's root
element", which is not what happens. Take a `Document`, or rename the parameter and the test.

### 7. The wiring test passes even if the wrong callback is wired in

`src/renderer/components/docking/platform-dock-layout.component.test.tsx`
Line: 95 — `expect.objectContaining({ onTabMiddleClick: expect.any(Function) }),`
Severity: Low

`expect.any(Function)` accepts `() => {}`. The `handleCloseTab` tests cover that function, but nothing
asserts it is the one passed. Assert `onTabMiddleClick: handleCloseTab`, or invoke the captured
callback and check the mocked `closeTab` got the id.

### 8. The mock comment narrates this PR's history instead of telling a future reader what the mock is for

`src/renderer/components/docking/platform-tab-title.component.test.tsx`
Line: 62 — ``// Mock heavy transitive deps that run side-effects at module init in jsdom. `logger.service` is``
Severity: Low

Lines 62-66 explain that a second mock "previously coexisted here and raced" — change narration,
which `.claude/rules/code-quality/forward-facing-comments.md` asks you to move to the commit message.
Two details make it worse: it sits above the theme mock, while the logger mock is at line 86; and the
hazard it describes still exists at `platform-dock-tab.component.test.tsx:10`, so either the claim is
overstated or the cleanup is incomplete. Keep the original first sentence.

### 9. The util's two TSDoc blocks run ~45 comment lines for ~35 lines of code, and one sentence blames the wrong library

`src/renderer/components/docking/platform-dock-layout-middle-click-handlers.util.ts`
Line: 38 — `* Installs the native listeners that make a tab's middle-button gesture behave correctly wherever`
Severity: Low

The install TSDoc is 30 lines in four paragraphs, and the overflow-dropdown/portal explanation is
repeated in the `readClosableTabId` doc, the tab-title comment and the test fixture doc. At line 45
the attribution is wrong: `OperationNode.js:176` does forward `getPopupContainer`; it is rc-dock's
`DockTabs.js:278` that never supplies one, so rc-trigger falls back to `document.body`. The
conclusion is right, only the source is misattributed. Keep the portal explanation once here and
point to it elsewhere. The accessibility paragraph is true but belongs in the PR body or a ticket.

### 10. The auxclick comment says the handler "never runs" for a press released off the tab, but it runs on every auxclick in the document

`src/renderer/components/docking/platform-dock-layout-middle-click-handlers.util.ts`
Line: 92 — ``// Bubble phase: nothing on a tab header's path stops `auxclick` (rc-dock and rc-tabs only ever``
Severity: Low

Since the listener moved to the document, what actually happens is that `readClosableTabId` finds no
`[role="tab"]` ancestor and returns `undefined`. "already closes nothing" also compares against the
earlier implementation. Behavior is correct; only the explanation is stale. Two lines will do: close
on the click rather than the press, and note that auxclick targets the nearest common ancestor of
press and release, so a press dragged off the tab resolves to no tab.

### 11. The least obvious decision in the PR — why the mousedown listener uses the capture phase — is the one thing not documented

`src/renderer/components/docking/platform-dock-layout-middle-click-handlers.util.ts`
Line: 91 — `targetDocument.addEventListener('mousedown', blockMiddleButtonDrag, { capture: true });`
Severity: Low

Only the PR body explains it, and the body is stale. rc-dock arms drags from React `onMouseDown` on
DragDropDiv (`DragDropDiv.js:300-310`) for every button except the right one, and React delegates
from the root container, so a capture listener on the document is the one place guaranteed to run
first. Worth two lines here, especially next to a five-line comment on the auxclick listener.

### 12. The sibling `isClosable` doc was not updated, so two public docs for the same flag now disagree

`src/shared/models/web-view.model.ts`
Line: 227 — `* Whether this WebView's tab can be closed by the user (shows the tab's close button and allows`
Severity: Low

This block now mentions middle-click, but `TabInfo.isClosable` in `docking-framework.model.ts:85`
still says only "(shows the tab's close button)", and both ship in `papi.d.ts`. That is also the
field the gate actually reads (`platform-dock-tab.component.tsx:39`). One line each keeps them in
step.

### 13. Both type-assertion disables in the new tests can be removed rather than justified

`src/renderer/components/docking/platform-dock-tab.component.test.tsx`
Line: 122 — ``// `title` is typed as rc-dock's broad `TabData['title']` union; narrow to read the prop this``
Severity: Low

The project's rule is to fix the code before suppressing. Verified under vitest that
`expect(result.title).toMatchObject({ props: { isClosable: false } })` passes on a real React element
and fails on a mismatch; `toHaveProperty('props.isClosable', false)` works too. Either removes both
disables and both explanatory comments. Alternatively, follow this file's own pattern — the
`PlatformTabTitle` mock at lines 19-23 already surfaces `webViewType` as a data attribute.

### 14. The AI attribution lists only one of the two sessions

PR body (no line anchor)
Severity: Low

Commits `6322a8eed78` and `71d5a0a3e65` carry a `Claude-Session` trailer for session `01YcEf...`,
which the body does not link. There is also a stray bare session URL after the footer. Repo format:
`AI-assisted — [session 1](...), [session 2](...)`.

### 21. A quick second middle click can close the tab that slid into the spot, because the hit target is now the whole header

`src/renderer/components/docking/platform-dock-layout-middle-click-handlers.util.ts`
Line: 85 — `const closeTabOnMiddleClick = (event: MouseEvent) => {`
Severity: Low

Middle-click tab A; the dock reflows and tab B slides left into A's old position. A second middle
click at the same position resolves to B in the live DOM and closes it. Browsers defend against
exactly this by freezing tab widths after a close until the pointer leaves the strip, and there is no
equivalent here. The X button is largely immune because it is a ~16px fixed-offset target; this
feature makes the whole `[role="tab"]` header a close target, which is what widens the window.
Nothing catches a mistake afterwards: there is no unsaved-changes or confirm-close hook anywhere in
the renderer or extensions.

Kept at Low rather than Medium: it needs a fast second click and the timing was not measured against
the running app. If you want to handle it, ignoring further tab-bar `auxclick`s until the pending
close settles is the cheap version. A sub-frame double click is harmless by comparison — rc-dock
mutates `tempLayout` synchronously, so the second close no-ops, though it does log a "tab not found"
warning for what is really normal usage.

---

## Nit

### 15. `handleCloseTab` is defined and exported in the tab-title component but only the dock layout uses it

`src/renderer/components/docking/platform-tab-title.component.tsx`
Line: 210 — `export const handleCloseTab = async (tabId: string) => {`
Severity: Nit

`PlatformTabTitle` never calls it. Moving it into the middle-click util or the dock layout would put
it with its only caller. The common argument against it does not hold: the import does not drag a new
module into the dock layout, since platform-dock-layout already reaches platform-tab-title through
the storage util and `createRCDockTabFromTabInfo`. The new test's mocks are needed either way.

See also finding 22 below, which shares this anchor.

### 16. The extension comment credits `isClosable` with disabling a close path that does not exist

`extensions/src/platform-scripture-editor/src/main.ts`
Line: 1068 — `// always remain open, so it's non-closable there — isClosable === false disables all three`
Severity: Nit

There is no keyboard close shortcut to disable — `keyboard-shortcuts.data.ts` has no tab-close entry
and the tab menu has only float/move items. "All three close paths" is also too strong, since
programmatic `closeTab` ignores `isClosable`. Something like "`isClosable: false` hides the close
button and ignores middle click" stays true, and will not need editing when the next gesture lands.

### 17. The test fixture doc attributes `.dock-bar` to rc-tabs; it is rc-dock's

`src/renderer/components/docking/platform-dock-layout-middle-click-handlers.util.test.ts`
Line: 5 — ``* Builds a DOM tree standing in for what rc-dock/rc-tabs actually render: a `root` (where the``
Severity: Nit

It comes from rc-dock's `DockTabBar.js:58`. The doc also presents the fixture as what rc-dock
"actually renders", while a real header nests two `[role="tab"]` elements (rc-tabs' `.dock-tab-btn`
around rc-dock's div) and the fixture has one. The code works from either, so it is only the
description that is off.

### 18. "logging rather than throwing into React" describes the wrong failure mode

`src/renderer/components/docking/platform-tab-title.component.tsx`
Line: 207 — ``* Closes a tab, logging rather than throwing into React on failure — whether `closeTab` rejects, or``
Severity: Nit

`handleCloseTab` runs from a native document listener, so an unguarded failure would be an unhandled
promise rejection, not a React error. The same wording appears in two test names. "logs instead of
rejecting if `closeTab` fails or finds no such tab" is accurate.

### 19. "(mouse button 1)" reads as the primary button to many people

`src/renderer/components/docking/platform-tab-title.component.tsx`
Line: 95 — `* Whether this tab can be closed by the user. Gates middle-click-to-close (mouse button 1)`
Severity: Nit

"middle button" is unambiguous; the constant name `MIDDLE_MOUSE_BUTTON` already says it. That
constant is exported at `util.ts:1` but only used in its own module.

### 20. The layout test's mock comment gives a reason that is not the real one

`src/renderer/components/docking/platform-dock-layout.component.test.tsx`
Line: 14 — ``// `platform-tab-title.component.tsx` (imported here for `handleCloseTab`) uses these hooks, which``
Severity: Nit

"imported here for `handleCloseTab`" implies the mocks exist because of that import; they would be
needed anyway, since PlatformDockLayout already reaches platform-tab-title through the storage util.
Also "imported here" points at the test, but it is the component under test doing the importing.

### 22. `handleCloseTab` is the third copy of the "closeTab, log both ways" shape

`src/renderer/components/docking/platform-tab-title.component.tsx`
Line: 210 — `export const handleCloseTab = async (tabId: string) => {`
Severity: Nit

The other two are in `dialog.service-shard.ts` at lines ~116 and ~160, and both use `logger.error`
for the not-found branch where yours uses `logger.warn`. The difference is defensible — a dialog
failing to close is a real fault, a middle click on a stale tab is not — so this is worth a sentence
in the PR, not a change.

### 23. Worth a line in the util saying why the existing tab-resolution helper is not used here

`src/renderer/components/docking/platform-dock-layout-middle-click-handlers.util.ts`
Line: 29 — `function readClosableTabId(target: EventTarget | null): string | undefined {`
Severity: Nit

`getTabInfoByElement` (`platform-dock-layout-storage.util.ts:204-253`) already resolves a DOM element
to a tab and returns `isClosable`, it is PAPI-exposed, and it is already imported into
`platform-dock-layout.component.tsx` and used a few lines above the new effect. A reviewer will ask
why `readClosableTabId` exists instead. Two concrete reasons, both worth recording so nobody
"simplifies" into them later:

- It falls back to the panel's `activeId` when the walk reaches `.dock-panel` (lines 229-237), so a
  middle click on the empty strip past "+" would close the active tab.
- In rc-tabs' overflow dropdown the header renders under a `...-more-popup-<key>` id that
  `TAB_HEADER_ID_REGEX` (`/rc-tabs-\d+-tab-(.+)/`) cannot match, so dropdown closes would silently
  stop working.

It also throws when it finds an id with no matching tab. The current approach is the right one; it
just is not self-evidently so.

---

## Checked and cleared

Questions a reviewer would reasonably raise, each traced to source and found sound:

- Dialog tabs close identically — `closeTab` → `removeTabFromDock` → `dockMove(tab, null, 'remove')`
  is the same call rc-dock's X button makes, through the same `onLayoutChange` that resolves dialog
  requests.
- Last tab in a panel, and the last tab overall, hit the same empty-dock handling.
- Floating panels are covered: `DockTabBar` always renders `.dock-bar`, and the listeners are on the
  document.
- No double-close from a middle click on the X button — rc-dock's `onClick` only fires for the
  primary button.
- Tab id semantics are consistent for web view and non-web-view tabs.
- Radix popovers and menus are unaffected: they listen on `pointerdown`, not `mousedown`.
- rc-dock's "New Window" path, which would break document-scoped listeners, is disabled in this repo.
- The premise of the feature checks out: rc-dock's drag-arming excludes only the right button, so
  blocking the middle button really is necessary.
- `preventDefault` on mousedown does not break tab activation: middle click never activated a tab,
  since activation goes through `onClick`, which is primary-button only.
- StrictMode double effects, remounts and `data-tab-closable` boolean stringification are all fine.
- The `closable` local in `createRCDockTabFromTabInfo` is a genuine simplification.
- File and test naming and placement match repo convention throughout.
- Listener efficiency is a non-issue: both handlers return on the button check before any DOM walk.

## Findings that were raised and dropped

The antagonistic pass rejected these as preference or unsupported, and they are not in the list
above: the `preventDefault` comment's Linux-paste clause; the MDX guideline's "also" wording and
placement; the missing one-line purpose comment on the new effect; most test-name length complaints;
and the proposal to reuse `getTabInfoByElement`, which is inverted into finding 23.

> @katherinejensen00 reviewed 12 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2769#-:-P1_u5I501P493RDYBIM:byixyu8)* status: all files reviewed, 1 unresolved discussion (waiting on rolfheij-sil).

___
*[`src/renderer/components/docking/platform-tab-title.component.tsx` line 862 at r1](https://reviewable.io/reviews/paranext/paranext-core/2769#-P1_uFwVCp9cYDnP7m-6:-P1_uFwVCp9cYDnP7m-7:b-nnx5ea) ([raw file](https://github.com/paranext/paranext-core/blob/71d5a0a3e65fe4c09ab7d239fdb5ba3ecabca2dd/src/renderer/components/docking/platform-tab-title.component.tsx#L862)):*
> ```XML
>             // which re-renders this same element) back to a tab id, and to gate that close on
>             // whether this tab allows it.
>             data-tab-id={id}
> ```

### 1. Middle-click's new `data-tab-id` takes over an existing focus lookup, sending keyboard focus to the tab strip instead of the tab's content

`src/renderer/components/docking/platform-tab-title.component.tsx`
Line: 862 — `data-tab-id={id}`
Severity: High

Before this PR, `data-tab-id` marked exactly one element: `PlatformPanel`, the tab's content
wrapper (`platform-panel.component.tsx:34`). `setDocumentFocusToTab` finds it with a document-wide
`document.querySelector([data-tab-id='...'])` and focuses its first child
(`platform-dock-layout-storage.util.ts:1340-1342`). rc-tabs renders the tab bar before the panels
(`rc-tabs/es/Tabs.js:226`), so that query now returns this title div, and `children.item(0)` is the
icon span at line 865 — it gets a temporary tabIndex, takes focus, and the panel content never
does.

This fallback runs on every tab mount, every tab click, and the focus-next step after any close,
because `setFocus` always calls `focusTab` (`window.service-shard.ts:500-510`). It affects every tab
with no `lastFocusedElement`: Settings, About, dialogs and error tabs always, plus web views, whose
iframe lookup at `:1286` is already shadowed by the title's pre-existing `data-web-view-id`.
Middle-click makes background closes common, so this fires more often than before. In Simple mode's
headless columns the focus ring lands inside a bar the repo keeps at `opacity: 0`, so it is
invisible. Nothing catches it: no test covers `setDocumentFocusToTab`.

Suggested fix: give the header its own attribute, e.g. `data-tab-header-id`, and read that in
`readClosableTabId`; that leaves both existing readers untouched. Scoping the lookup to
`.platform-panel[data-tab-id=...]` also works. Do not resolve the id via `getTabInfoByElement`
instead — see finding 23. Worth a test asserting only one `[data-tab-id=x]` exists per tab.

Verified by three independent passes; mechanism confirmed in source, not reproduced in the running
app.

rolfheij-sil:
🤖 Claude: Thanks for the careful review. Every finding held up, and all of them are fixed on the branch or answered in the new description. A few needed small corrections (1, 8, 21).

The branch is also rebased onto current `main`, because it had conflicts. Some findings were settled by the rebase itself; those are noted below.

## Fixed

- **2 — no real rc-dock test.** Added `platform-dock-layout-middle-click.dom-contract.test.tsx` (ab9196d0457; typing fix in d22ea3c0cb2). It renders a real `DockLayout` with the app's tab title and panel, and installs the real handlers. It checks three things:
  - A middle press on a title, a close button or the tab bar arms no drag and is default-prevented. A left press on the title is the positive control.
  - A middle click on the title, close button or hit area closes a closable tab.
  - A non-closable tab and panel content close nothing.

  Three mutations each turn it red: dropping `capture: true`, narrowing the drag scope to `.dock-dropdown`, and ignoring `closable`. Swapping `[role="tab"]` for `.dock-tab` does not turn it red. In the main tab bar, rc-tabs' `.dock-tab` also wraps the header, so the two selectors only differ in the overflow dropdown, which jsdom can't render. The hand-built util tests cover the dropdown and go red for that mutation. The same file also pins finding 1.
- **5, 6, 7, 13, 15, 18, 19 — closability and wiring** (783b2fe35a4).
  - 5: Closability is now read from rc-dock at click time (`find()`, then `isTab` and `closable`), which is the same flag the X uses. An unset flag counts as not closable, as it does for the X. The `isClosable` prop, the `data-tab-closable` attribute and `createRCDockTabFromTabInfo`'s shared local are gone.
  - 15, 18: `handleCloseTab` is now an unexported function next to its only caller in `platform-dock-layout.component.tsx`. Its TSDoc reads "logging instead of rejecting if `closeTab` fails or finds no such tab".
  - 6: The util takes the `Document` it listens on, and the wiring test now says "on the window's document".
  - 7: The wiring tests now call the captured callbacks. `findTab` reaches the dock's `find`. `onCloseTab` removes the tab through the real `closeTab`. A missing tab is logged as a warning and a throwing close as an error, and neither rejects.
  - 19: The prop's TSDoc went with the prop, and `MIDDLE_MOUSE_BUTTON` is no longer exported.
  - 13: No longer applies. Those tests were removed, and the dock-tab component and its test match `main` again.
- **9, 10, 11, 12, 17, 23 — comments and docs** (514aa151aaf, polished in 8338dea1668).
  - 9: The install TSDoc is much shorter and explains the portal once. It now names the right source: rc-dock's `DockTabs` passes no `getPopupContainer`, so rc-trigger falls back to `document.body`. The accessibility paragraph moved to the description.
  - 10: The `auxclick` comment is down to two lines: the close happens on the click, and a press dragged off the tab resolves to no tab.
  - 11: The capture listener now says why it has to capture.
  - 12: `TabInfo.isClosable` now mentions middle click, with the same wording as the WebView definition. `papi.d.ts` was regenerated; only that TSDoc text changed.
  - 17: The fixture doc credits `.dock-bar` to rc-dock, calls the fixture a simplified stand-in, and points to the contract test.
  - 23: Both of your reasons are recorded above `readTabHeaderId`. The second one fully holds again after the rename for finding 1, since the header no longer carries `data-tab-id`.
- **16.** After the rebase that comment no longer exists, because `main` moved the grid provider to a file whose `isClosable` has no comment. The other place that made the same claim is the scripture-text-grid e2e comment. It now says the missing close button stands in for every close path because middle click reads the same `closable` flag (514aa151aaf).
- **20.** After the rebase, the layout test's mock comment is `main`'s own again. The stray blank line is gone too (783b2fe35a4).

## Fixed, with a correction

- **1 — focus lookup.** Confirmed. With real rc-dock in jsdom there were two `[data-tab-id]` matches, and focus landed on the title's icon. The header now uses its own `data-tab-header-id`, so `data-tab-id` again marks only the panel (c1223578855). The contract test asserts that only panels carry `data-tab-id` and that focusing a tab lands in its panel content. In the running app, opening About now puts focus in its panel.
  One correction: this didn't happen on every tab click. It only happened when a panel mounted for the first time, after a close, and on explicit focus calls.
  Something we saw live that this PR didn't cause: after *any* close, with the X or a middle click, focus goes to the page body instead of the newly active tab.
- **8 — narration comment.** The comment is gone; the rebase had already dropped it. The tab-title test file now matches `main` except for one new header-attribute test. One correction: the same hazard doesn't exist in `platform-dock-tab.component.test.tsx`, which has only one logger mock.
- **21 — second middle click.** Confirmed in the running app. A second middle click closed the tab that slid under the pointer at 150, 300 and 600 ms. One correction: the X button has the same exposure, which we also saw live. With tabs of similar width, the next tab's X lands where the closed tab's X was.
  "Ignore until the pending close settles" wouldn't help, because `closeTab` settles almost immediately. So there's now a one-shot guard instead: a repeat middle click within 500 ms and 4 px (on each axis) of a close is ignored (c7bbe437425).
  Unit tests cover it, and so does a new Playwright spec (7cf6b7e5028). The spec also shows that the browser still fires `auxclick` after the press is cancelled, and that a middle press-drag starts no drag. A click after 500 ms still closes, like the 600 ms case above. That matches the usual double-click time.

## Description

- **3, 14, 22.** The description is rewritten to describe the current code, and it links all the sessions. It also notes that `handleCloseTab` is the third copy of the close-and-log shape. It uses `logger.warn` on purpose: a middle click on a tab that is already gone is not a fault, while a dialog failing to close is.
- **4.** The accessibility sentence is corrected. The description now says that no keyboard or screen-reader way to close a tab exists today, and that this predates this PR.

## Recorded for later

- **4 — keyboard and screen-reader close.** This gap is tracked in PT-3034.
- **1 — web view focus.** You also pointed out that the title's `data-web-view-id` shadows the web view iframe lookup. That predates this PR and is tracked in PT-4637.

katherinejensen00 (approved):
Thanks for making those changes, Rolf! Looks good

> @katherinejensen00 reviewed 17 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2769#-:-P1jguFv37NRQuHXcTDo:b8doj4n)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2765, opened 2026-09-02
PT-4501: Catch renderer crashes instead of leaving a blank window

jolierabideau:
# Code Review Summary

**Branch**: pt-4501-app-unresponsive-sync

**Base**: origin/main

**Date**: 2026-09-02

**Review model**: Claude Opus 5

**Files changed**: 18 (17 source/asset files plus the regenerated `lib/papi-dts/papi.d.ts`)

## Overview

This branch contains PT-4501's blast-radius fix: the defect where an uncaught render throw tears
down the renderer's entire React root and leaves the window on screen, painted with nothing, until
the app is restarted. It has three parts. A `RendererErrorBoundary` now wraps the root above the
router in `index.tsx`, so a throw produces a `WindowCrashedView` — a crash screen with a Reload
button — instead of a blank window; it is deliberately hook-free, because any hook it called would
run above its own catch. Two toolbar hooks, `useOpenProjectBookIds` and `useProjectPickerData`, drop
a refresh-counter anti-pattern that re-rendered their consumers on every web view event by
construction, replacing the counter with the derived value itself so React's `Object.is` bailout
collapses no-op events to zero re-renders. And the `useData` runaway-render guard now names the
offending selector in its warning, computed lazily at trip time, which turns "we cannot reproduce
it" into "a reporter's ordinary log names the setting".

The three parts attack different layers of the same failure, and the layering is documented at
length in the author's investigation notes (`docs/PT-4501-jira-edits.md`, section 13): a latent
Rules-of-Hooks defect from PT-1561, amplified by PT-4275 making web view events buffered
cross-window network events, triggered by PT-4346 subscribing the toolbar to all three of them
per-event. Six candidate root causes have been investigated and falsified; the deliberate strategy
is to make a trip survivable rather than to keep hunting the trigger. **The containment has not yet
been observed working in a packaged build** — see Suggested Review Focus.

## API Changes

One addition, and it is new in this review rather than in the original branch:

- `src/renderer/hooks/hook-generators/create-use-data-hook.util.ts` now exports
  `MAX_SELECTOR_DESCRIPTION_LENGTH = 80` so the truncation test can assert against it rather than
  hardcoding the number. This surfaces in `lib/papi-dts/papi.d.ts`. It follows the established
  pattern in that same module, which already exports `RUNAWAY_EVENTS_PER_WINDOW` and
  `RUNAWAY_WINDOW_MS` for the same reason; both are already present in `papi.d.ts`.

[...]

PR #2758, opened 2026-09-02
PT-4463: Lift the web-view move subsystem out of the router

rolfheij-sil:
## Summary

`web-view.service-router.ts` was 1832 lines, and the largest thing in it was not routing: the cross-window move machinery — the in-flight registry, the pending-content window lifecycle, the abandoned-window close, the recovery ladder and the late-adopt probe — sat inline in a file named for something else. This lifts it out.

**It is a pure lift. No behaviour changes, no API changes, no test-semantics changes.** The move commands stay registered exactly where they were and keep their OpenRPC documentation and `x-experimental` markers; only the implementation moved. Every symbol travels verbatim — the only edits are the `export` each moved symbol needs and the imports each file no longer uses.

**Why review this:** part of the PT-4275 multi-window epic, as PT-4463. It is the follow-up the lead dev asked for when reviewing #2670: *"the router file should read as routing again: lift the move subsystem into its own main-process module beside the router, as the PR body itself concedes it owes."*

## The split

Four modules, ordered so the graph is acyclic — each imports only from those above it:

| module | lines | owns |
| --- | --- | --- |
| `web-view-shard-index.ts` | 43 | The per-window WebView service shards, the focused-window resolver, and `getWebViewShard`. A leaf: imports none of the others. |
| `web-view-ownership.util.ts` | 130 | `OwnerMatcher` and `describeMatcher` — the vocabulary of an ownership search — plus `WebViewMoveInFlight` and the in-flight register. Also a leaf. |
| `web-view-owner-resolution.util.ts` | 500 | `findOwner`, `findLayoutTargetOwner`, the window-creator wiring, and `createFreshWindow`. |
| `web-view-move.util.ts` | 510 | The move policy: `moveWebView`, `raiseMoveTarget`, `findWebViewAdoptedAfterTimeout`, `readoptAfterFailedMove`, `recoverAfterFailedMove`. |

The four modules follow the `<feature>.<type>.ts` convention — a hyphenated subject and a dotted kind segment; `git mv` preserved their history at 99–100% similarity.

**`web-view.service-router.ts` is 923 lines, down from 1832** — a 50% cut — and keeps the routing, the command registrations, and `openWebViewInOwningWindow`, which the move path never calls.

## Two placements that are design decisions, not accidents

[...]

PR #2756, opened 2026-09-02
PT-4465: A window nobody asked for opens without taking focus

rolfheij-sil:
## Summary

A window an extension opens, or one a cross-window move creates, no longer pulls the user out of
whatever they are doing **inside this application**. A window a person asked for — a menu item,
`platform.createWindow`, a dock click, a startup restore — comes to the front exactly as it always
has.

This does not reach across application boundaries: a window that takes the foreground from another
application cannot hand it back, because `focusWindow` only moves focus between our own windows. See
"Why the hand-back is gated instead of deferring the reveal" under The mechanism for what was
measured and what the gate does and does not cover.

### Why review this

Part of the multi-window epic, as the follow-up the lead dev asked for on #2670 (review item 6,
"a backgrounded window creation should show without activating"), deferred from that PR at his
suggestion after two in-PR attempts failed.

## What this actually turned out to be

Withholding a window from the foreground is not achievable by asking it nicely. `showInactive()`
shows the window without activating it, and then the page takes focus by itself the moment it first
paints. Instrumentation on this branch established that with a clean negative:

- **Main never raises it.** Wrapping `show`, `showInactive`, `focus`, `moveTop`, `setAlwaysOnTop`
  and `webContents.focus` on every tracked window showed that the only call ever made on the new
  window is `showInactive()`, and its `focus` event arrives with a stack of just
  `BrowserWindow.emit` — no JS caller.
- **The renderer never raises it either.** Forwarding page console proved live from that window
  (144 lines, per-page startup output), and the `contentWindow.focus()` / `iframe.focus()` sites in
  `setDocumentFocusToTab` never ran.
- **The environment is not the cause.** A minimal Electron control under the same bare Xvfb — one
  visible focused window, then a second created hidden and revealed with `showInactive()` — leaves
  the first focused and the second visible and unfocused. The difference from the app is that the
  control's second window loads `about:blank` rather than a page that mounts an iframe.

So the focus cannot be prevented where it happens. It is undone instead.

## The mechanism

**On the withheld window's first `focus` event, focus is handed straight back** to the window the
user was in, through the existing `focusWindow` path — no new raise primitive. The window stays
visible and in the taskbar; only the focus is returned.

- **At most once per window.** A rule that fired every time would make the window impossible to
  enter: every attempt would throw the user out again. After the first, the window keeps whatever
  focus it is given.
- **Never to itself.** A withheld window can be the first window the process tracks — an extension
  opening a web view before any window is restored — and routing then answers with the window
[...]

PR #2770, opened 2026-09-04
docs: order architecture-decision entries in byte order by slug

jolierabideau:
Every branch appended new entries at the end of `Architecture-Decisions.md`, so every pair of concurrent branches conflicted on the same last line of the file. Replacing ADR numbers with slugs made those conflicts trivial to resolve ("keep both") but did nothing to stop them occurring, since the identifier was never what collided — the shared insertion point was.

Sort the existing entries by slug and change the guidance to insert in sorted position, so unrelated decisions land at unrelated offsets. This is a reduction rather than a cure: entries average ~55 lines, so two branches now collide only when they insert into the same gap between neighbouring slugs — which still happens when closely related work lands together (several `adr-window-…` entries in one epic).

The reorder is a pure permutation of entry content plus normalizing entry separation to one blank line. Chronology moves from file order into each entry's `**Date:**` field. Also reworded the note about formerly-numbered entries, which identified them by file position ("entries through `adr-async-hook-state-shape`") and no longer could.

## Ordering is byte order, not "alphabetical"

Entries sort in **byte order** (`LC_ALL=C sort`), and the docs now say so explicitly. Slugs are hyphenated, and locale-aware collation can weigh `-` differently from a letter, so the two orders disagree on pairs like `adr-no-create-…`/`adr-node-…`. This matters because `.erb/scripts/sort-cspell-words.js` — the obvious in-repo precedent anyone would reach for — uses `localeCompare`.

## Two entries edited beyond the permutation

- **`adr-layout-persistence-guard-retirement`** — removed a spurious `**Formerly:** ADR-0024`. That number is genuinely `adr-toolbar-sync-status-is-local`'s (mapped by the slug-conversion commit `d9ff7654b07`); this entry was introduced already-slugged in `dc972233209` (#2425) from a branch that had picked `ADR-0024` locally, and `git log -S` confirms it was never numbered. The result is a clean ADR-0001–ADR-0028 bijection with no gaps and no duplicates.
- **`adr-lightweight-decision-log`** — its `**Consequences:**` no longer carries the sorted-insertion decision; that is now its own entry (below).

## The convention is recorded as a decision

New entry `adr-decision-log-sorted-insertion` (dated 2026-09-03) records the sorted-insertion decision with its own Context/Decision/Alternatives/Consequences, rather than leaving the rationale only in prose. It sits at its own slug offset, so it also exercises the convention it describes. Notably it records `merge=union` in `.gitattributes` as **rejected**, not open: union merge is line-wise, so on multi-line prose entries it interleaves two entries' bodies with no conflict marker and auto-merges disjoint insertions out of slug order — silent corruption worse than the conflict it avoids.

## ⚠️ Heads-up for anyone with an open branch that adds an ADR entry

**Do not resolve the transition conflict by "keeping both sides".** Both sides of a conflict in this file are runs of whole entries, and after the reorder the *same* entry can appear on both sides at different offsets. Keeping both silently duplicates entries, leaves the file out of sort order, and nothing in CI, lint, or `/verify-standards` catches it.

This is not hypothetical — it is what the merge in this PR hit. #2730's landing produced one conflict hunk whose "theirs" side carried 6 entries, **4 of which already existed at their new sorted positions**; a literal "keep both" would have triplicated `adr-grapheme-segmenter-unicode-segmenter`, `adr-primary-window-owns-app-lifetime`, `adr-registration-validity-once-per-session`, and `adr-tab-menu-channel-and-window-naming`.

**Resolve by slug instead:** take the union of the two sides keyed by slug, so every slug appears exactly once and genuinely-new entries land at their own byte-order slug position. Then check that `grep '^## adr-' <file>` is free of duplicates and equals its own `LC_ALL=C sort`. The file's "How to use it" section now documents this, and `CLAUDE.md` carries the short form.

21 open PRs currently touch this file and will need that treatment: #2768 #2766 #2765 #2760 #2758 #2756 #2752 #2751 #2750 #2749 #2747 #2746 #2745 #2744 #2742 #2737 #2736 #2734 #2700 #2632 #2551.

## The invariant is convention-enforced only

Nothing checks that entries stay sorted and unduplicated — the next hand-added entry can silently break it, and this file has already lost one convention to non-enforcement (`82ea330bf40`). A `verify:adr-order` check is recorded as open follow-up work in the new entry rather than added here.

## Verification

Checked mechanically on the merged result:

- **58 entries**, headings unique, in exact `LC_ALL=C sort` order.
- **Entry-by-entry byte comparison against `origin/main`**: same 57 entries, bodies byte-identical, plus the one new entry — the only two differing bodies are the two deliberately edited above.
- The merge resolution added **171 lines and removed 0**, matching #2730's insertion exactly.
- All 19 backtick-quoted `adr-*` references in the log resolve to a defined slug; zero dangling. No `adr-*` references in `CLAUDE.md` or `pt10-reuse-scout.md`.
- Zero consecutive blank-line runs; 57 of 58 entries carry a `**Date:**`. The one exception, `adr-singleton-services-elect-host-window`, is a withdrawn tombstone stub that records no decision and had no `**Date:**` before this change either — left as-is rather than inventing a date from commit archaeology.
- Exactly 28 `**Formerly:**` entries, ADR-0001–ADR-0028, no gaps, no duplicates.

`npm run typecheck`, `npm run lint`, and `npm test` were not run: every change is markdown prose and none of those commands read markdown. Prettier is likewise not a check here — `.prettierignore` exempts `.claude`, `.context`, and `CLAUDE.md`, so a passing Prettier run on these files would be vacuous. Note also that a naive `grep '^## adr-' | sort` check reports a false failure, because it catches the fenced `## adr-{slug}` template and `{` sorts after `z`; exclude the template when checking.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2770)


katherinejensen00 (changes requested):
One key blocker and the rest looks good. Giving you the rest of the findings just in case any seem worth doing, but none looked critical and don't feel like you need to give in to scope creep.


Let me check the other headline claim.
nd now quantified.

Waiting on the full numbered list from /code-review max.

✻ Churned for 47s · done 11:37 AM · 1 shell still runningpen PR.

  Ran 1 shell command

#2749 already conflicts against main today, which contradicts part of finding 2. Let me check GitHub's own mergeable flags and then test the substance of finding 1.
and #2744 already conflict against main today "keep both" actually produces.

  Ran 1 shell command

  Read 1 file

9–36 never arrived, but the list is severity-ordered and had already descended to Minor by item 15, so the remainder is Minor/Nit — I'll note the gap rather than block on it.

Here is the consolidated review.laims in the PR body that don't hold up.


What I verified holds. The permutation is genuinely pure: 55 entries, , and git log -S finds it was never numbered. The result is a clean ADR-0001–ADR-0028 bijection. Your merge from main was resolved into correct sorted position. And the premise is real: 34 first-parent landings touched this file in the last month, 28 of them on 

Summary: The reorder itself is clean and the motivation is well founded — but there is one blocking defect: the file's own conflict-resolution instruction, applied to the conflict this PR creates, silently duplicates entries. I reproduced it end-to-end against a real open PR. Everything else is wording, plus four claims in the PR body that don't hold up.

What I verified holds. The permutation is genuinely pure: 55 entries, bodies byte-identical to the merge base except the two you deliberately edited and the rewrittt order; no duplicateslugs; every adr-* reference repo-wide resolves. The ADR-0024 archaeology is correct and better than most people would have done — d9ff7654b07 maps ## ADR-0024 to adr-tooremainder is Minor/Nit — I'll note the gap rather than block on it.

Here is the consolidat

---

Review of PR #2770 — for the PR author

Summary: The reorder itself is clean and the motivation is well founded — but there is one blocking defect: the file's own conflict-resolution instruction, applied to the conflict this PR creates, silently duplicates entries. I reproduced it end-to-end against a real open PR. Everything
else is wording, plus at don't hold up.ry adr-* reference repo-wide resolves. The ADR-0024 archaeology is correct and better than most people would have done — d9ff7654b07 maps ## ADR-0024 to adr-toolbar-sync-status-is-local, adr-layout-persistence-guard-retirement was introduced already-slugged in dc972233209 after the switch, and git log -S finds it was never numbered.
The result is a clean Your merge from mainwas resolved into correct sorted position. And the premise is real: 34 first-parent landings touched this file in the last month, 28 of them on days with 2+ landings, five on 2026-08-24 alone.
ecipe corrupts the file on the transition conflict.
I merged PR #2730 into this branch in a scratch worktree. It produces one conflict hunk (404 lines ours, 544 theirs) whose two sides share four headings that also exist at their new sorted positions. Resolving it literally as line 45 instructs — keeping both entries — yields 65 headings
instead of 57, with adsegmenter,adr-primary-window-owns-app-lifetime, adr-registration-validity-once-per-session and
adr-tab-menu-channel-ang three times, and thetly leaves sort order. Nothing in CI, lint, or /verify-standards would catch it. Line 45 is right for the steady state but actively dangerous for the one-time transition; it needs a sentence covering it: take the reordered file, then insert your entry at its own slug offset.

Important

2. .context/standards/— "alphabetical" namesslugs; every adr-* reference repo-wide resolves. The ADR-0024 archaeology is correct and better than most people would have done — d9ff7654b07 maps
## ADR-0024 to adr-tooadr-layout-persistence-guard-retirement was introduced already-slugged in
dc972233209 after the it was never numbered.The result is a clean ADR-0001–ADR-0028 bijection. Your merge from main was resolved into correct sorted position. And the premise is real: 34 first-parent landings touched this file in the last month, 28 of them on days with 2+ landings, five on 2026-08-24 alone.

Critical

1. .context/standards/— the "keep both convention read: "because the file is append-only, nothing catches it:
the second merge simplcal headings." The repo has already abandoned one ADR-log convention specifically because nothing enforced it — and this PR adds a second unenforced one to the same file.

Let me check the other

Confirmed and slightly sharper than reported: 34 first-parent landings touched this file in the last month (~1/day), and 28 of the 34 landed on days that saw 2+ ADR landings — five on 2026-08-24 alone. The PR's motivation is well founded and now quantified.

Waiting on the full numbered list from /code-review max.

://github.com/paranext/para…" finished · 1h 2m 20s

Findings 1–18 in, including two Criticals I need to verify myself — they drive the whole recommendation. Let me test the transition-conflict claim empirically against a real open PR.

  Ran 1 shell command

#2749 already conflictontradicts part offinding 2. Let me check GitHub's own mergeable flags and then test the substance of finding 1.

  Ran 4 shell commands

Important correction: ict against main today

  Ran 2 shell commands

Reproduced independently — and this is the most important finding in the review. Let me confirm what "keep both" actually produces.

  Ran 1 shell command
sort order. Nothing in the repo would catch it. Let me clean up.

  Ran 1 shell command

Cleaned up — no repo changes made. One item from the review is about my own stale memory, which auto-loads every session. Let me check it.

  Read 1 file

My memory file is demoindependent of this PR. Let me verify before correcting it.

  Ran 2 shell commands, recalled 1 memory, wrote 2 memories

Memory corrected. Findings 19–36 never arrived, but the list is severity-ordered and had already descended to Minor by item 15, so the
remainder is Minor/Nithan block on it.

---

Review of PR #2770 — for the PR author

Summary: The reorder itself is clean and the motivation is well founded — but there is one blocking defect: the file's own conflict-resolution
instruction, applied ts, silently duplicates

What I verified holds. The permutation is genuinely pure: 55 entries, bodies byte-identical to the merge base except the two you deliberately edited and the rewritten "How to use it"; exact sort order; no duplicate slugs; every adr-* reference repo-wide resolves. The ADR-0024 archaeology is correct and better than most people would have done — d9ff7654b07 maps ## ADR-0024 to adr-toolbar-sync-status-is-local,
adr-layout-persistenceced already-slugged indc972233209 after the switch, and git log -S finds it was never numbered. The result is a clean ADR-0001–ADR-0028 bijection. Your merge from main
was resolved into corrpremise is real: 34first-parent landings touched this file in the last month, 28 of them on days with 2+ landings, five on 2026-08-24 alone.

Critical

1. .context/standards/Architecture-Decisions.md:45 — the "keep both
entries" recipe corrup conflict.heir new sorted positions. Resolving it literally as line 45 instructs — keeping both entries — yields 65 headings instead of 57, with adr-grapheme-segmenter-unicode-segmenter, adr-primary-window-owns-app-lifetime, adr-registration-validity-once-per-session and adr-tab-menu-channel-and-window-naming each appearing three times, and the file silently leaves sort order. Nothing in CI, lint, or
/verify-standards woulfor the steady state

Important

2. .context/standards/Architecture-Decisions.md:32 — "alphabetical" names no collation, and the PR body's stability claim is false.
The landed order is byte order. printf
'adr-node-dom-globals-reate-project\n' |LC_ALL=en_US.UTF-8 sort reverses the pair, as does
adr-single-verse-…/adrollation ignores thehyphen at primary strength. The PR body states this was "CONFIRMED … stable under both LC_ALL=C and LC_ALL=en_US.UTF-8" — it isn't. Say byte order explicitly. This matters because .erb/scripts/sort-cspell-words.js:24, the obvious precedent anyone would reach for, uses localeCompare.

erseded by adr-<slug>, an in-place edit; line 56's own template offers that value; **Amended 2026-08-07** bullets already exist at lines 626 and 2181; commits d8a35d9499e, 6e41ce1cee8, 121fc457739 do it routinely; and this PR rewrites two entries. The base's vaguer "append-only" survived that reading — the new explicit gloss is falsifiable and false. Suggest: entries are added, not deleted — only **Status:** is revised, or an **Amended YYYY-MM-DD:** note appended.

(heading line 946) now carries a 2026-09-03 decision, which directly falsifies the claim you add at line 36 that each entry's **Date:** still carries chronology. Its alternatives also sit under **Consequences:** rather than **Alternatives:**, where the template at line 59 and any scanning agent will look. The file already has the fix: an **Amended 2026-09-03:** bullet, or split it into its own adr-decision-log-sorted-insertion entry — which would also demonstrate the
new convention in the

5. .context/standards/— the invariant has nomechanical enforcement, in a file already burned by exactly that.
Commit 82ea330bf40 deleted the previous ADR convention with the stated reason: "because the file is append-only, nothing catches it: the second merge simply leaves main with two identical headings." This PR adds a second unenforced ordering convention to the same file. None of these
three files is coveredCODEOWNERS. Butards doc in the same directory, so the .context format/lint exemption is no obstacle. A ~15-line verify:adr-order on the same hook would make this self-enforcing.

6. PR body and squash message — four claims that don't hold.
The collation claim (item 2). The Prettier row is vacuous: .prettierignorerun:74-79 exempts .claude, .context and CLAUDE.md, and
prettier --file-info rall three, so --checkritten it fails on your correctly-sorted file, because the grep catches the fenced ## adr-{slug} template and { sorts after z. And the entry count is wrong in both places: the commit says 52, the body says 54, the actual count at HEAD is 55.

7. .claude/agents/pt10-reuse-scout.md:68 — paraphrases a standard instead of linking to it.
.claude/rules/agent-aumd is paths:-scoped to.claude/agents/** and says multi-paragraph paraphrasing "makes the agent think it already has the content, and the copy drifts." The item grew from 6 to 9 lines, and commit b7ce4efbe8b exists precisely because that drift
already happened.
cture-Decisions.md:971 — merge=union is recorded as a live option but is unsafe here.
Union merge is line-wise. On this file's multi-line prose entries it interleaves two entries' bodies with no conflict marker, and disjoint insertions auto-merge out of slug order while swallowing the blank-line separator. It should be recorded as rejected for prose files, not "open and would compose with it."

#2768 #2766 #2765 #2760 #2758 #2756 #2752 #2751 #2750 #2749 #2747 #2745 #2744 #2742 #2737 #2736 #2734 #2730 #2717 #2700 #2654 #2632 #2551. Most append at EOF. At minimum #2730 goes from cleanly mergeable to conflicted because of this change. Worth a line in the body and a team heads-up, especially given item 1.

Minor

10. Architecture-Decisions.md:36 — "Each entry carries its own **Date:**" is false in the sentence that justifies dropping file order:
adr-singleton-services) has none. 54 of 55.

11. Architecture-Decisions.md:46 — "Entries predating the switch … each carry a **Formerly:** line" isn't checkable from the file; 19 entries are dated before 2026-08-26 without one. Your meaning (entries that existed pre-switch) is correct — just state the range instead: the 28
originally-numbered en028, no gaps.licates the rule and its rationale already in the line 16 table cell. With ~8 conflicts a week, resolution is the higher-traffic instruction.

13. .claude/agents/pt10-reuse-scout.md:70 — "read the log wholesale" makes a 268 KB / ~67k-token read mandatory every /investigate-prd run, past Read's default line limit.

14. Architecture-Decisions.md:41 — the residual-risk caveat undersells the fix. Adjacent slugs don't collide; entries average ~55 lines, so only insertions into the identical gap conflict. As written a reader concludes
the change barely helpd this and your reasoning is sound; flagging only because it came up twice independently and deserves a one-line decision rather than a silent pass.

16. Architecture-Decisions.md:912 — the **Formerly:** ADR-0024 removal is correct but is a content fix riding under an "order entries" title. One line in the body would save the next reader the archaeology.

Nits: line 38's rationale bullet duplicates line 966 almost sentence-for-sentence; line 971's "remain open" is an undated status claim (.claude/rules/docs-duspelled six differentways across the three files; pt10-reuse-scout.md:73 makes deletion an exception to a no-rewrite rule.

> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2770#-:-P0hjgDo78Ou_DOaVjX-:b-h359xr)* status: all files reviewed, 1 unresolved discussion (waiting on jolierabideau).

___
*[`.context/standards/Architecture-Decisions.md` line 45 at r1](https://reviewable.io/reviews/paranext/paranext-core/2770#-P0hjokLDY8TOiAPgEA9:-P0hjokLDY8TOiAPgEAA:b-lsaq46) ([raw file](https://github.com/paranext/paranext-core/blob/04359da9d3682fa0a4770e556250314120516cde/.context/standards/Architecture-Decisions.md#L45)):*
> ```Markdown
> - **Slugs are chosen at write time and never change.** Two branches in flight pick different slugs
>   on their own, so nothing has to be claimed, reserved, or reconciled at merge. When git does report
>   a conflict here, resolve it by **keeping both entries** in slug order — there is nothing to
> ```

1. .context/standards/Architecture-Decisions.md:45 — the "keep both entries" recipe corrupts the file on the transition conflict.
I merged PR #2730 into this branch in a scratch worktree. It produces one conflict hunk (404 lines ours, 544 theirs) whose two sides share four headings that also exist at their new sorted positions. Resolving it literally as line 45 instructs — keeping both entries — yields 65 headings instead of 57, with adr-grapheme-segmenter-unicode-segmenter, adr-primary-window-owns-app-lifetime, adr-registration-validity-once-per-session and adr-tab-menu-channel-and-window-naming each appearing three times, and the file silently leaves sort order. Nothing in CI, lint, or /verify-standards would catch it. Line 45 is right for the steady state but actively dangerous for the one-time transition; it needs a sentence covering it: take the reordered file, then insert your entry at its own slug offset.

jolierabideau:
Worked through all of it. Note the review text arrived on GitHub garbled — several findings lost their header lines to what look like terminal-render artifacts (3, 4, 8, 9, 12, 15). I reconstructed all but one from the surviving body text; **15** I could only read as *"...your reasoning is sound; flagging only because it came up twice independently and deserves a one-line decision rather than a silent pass"* and treated as the provenance stamp. If I mapped any of those wrong, say so and I'll redo that one.

Branch is now merged up to latest `main`.

## 1 — "keep both entries" corrupts the transition merge · **fixed**

Reproduced it for real rather than taking it on argument: #2730 landed in `main`, so the merge this PR needed *was* the transition conflict. One hunk, whose "theirs" side carried 6 entries — **4 of which already existed at their new sorted positions**. A literal "keep both" would have triplicated `adr-grapheme-segmenter-unicode-segmenter`, `adr-primary-window-owns-app-lifetime`, `adr-registration-validity-once-per-session`, `adr-tab-menu-channel-and-window-naming`. Your reproduction was exactly right.

Rather than adding a sentence scoped to the one-time transition, I replaced the recipe: the bullet is now **"reconcile by slug, never by hunk"** — take the union of both sides keyed by slug, insert genuinely-new entries at their own byte-order offset, then verify headings are duplicate-free and in `LC_ALL=C sort` order. That's correct in the steady state *and* the transition, since the failure mode (same entry on both sides at different offsets) isn't actually unique to the transition. Short form added to `CLAUDE.md` too, and the PR body now carries a heads-up with the 21 affected open PRs.

## 2 — "alphabetical" names no collation · **fixed, with one correction**

Fixed as asked — the docs now say byte order / `LC_ALL=C sort` explicitly, and call out `localeCompare` and the `sort-cspell-words.js` precedent as the thing not to reach for.

One correction to the finding: your specific repro doesn't hold here. `adr-node-dom-globals` vs `adr-no-create-project` orders the same way under `LC_ALL=C sort`, `LC_ALL=en_US.UTF-8 sort`, and JS `localeCompare` on macOS, and all 55 pre-merge slugs sort identically under all three — so the landed file wasn't mis-sorted under any collation I could test. I'd guess you're on glibc, where collation does ignore punctuation at primary strength. Your *conclusion* stands either way and is worth the fix: "alphabetical" is genuinely underspecified, and the answer shouldn't depend on whose `sort` runs. But I've dropped the "stable under both C and en_US" claim from the body rather than restate it — you were right that it was asserted more strongly than it was tested.

## 3 — "entries are never rewritten in place" is falsifiable and false · **fixed**

Confirmed: two `**Amended`bullets already exist, `**Status:**` gets revised in place, and this PR itself rewrites two entries. Took your suggested wording — "entries are added, not deleted — a decision that no longer holds has its `**Status:**` revised or an `**Amended YYYY-MM-DD:**` note appended". The gloss I added during review was worse than the vaguer "append-only" it replaced, which is a fair hit.

## 4 — the sorted-insertion decision was buried under a 2026-06-18 date · **fixed, split**

Took the split option. New entry `adr-decision-log-sorted-insertion` dated 2026-09-03, with the rationale in `**Decision:**` and the rejected options in `**Alternatives:**` where the template and a scanning agent will actually look. `adr-lightweight-decision-log`'s `**Consequences:**` now just points at it. It sits at its own slug offset, so it demonstrates the convention as you noted.

## 5 — no mechanical enforcement · **deferred, recorded**

Agreed on the substance, and the `82ea330bf40` precedent is a strong argument — I confirmed that commit deleted the old convention with exactly the reason you quoted. Not adding the check here, though: it's new tooling in a prose-only PR, and you'd flagged not wanting me to scope-creep. It's recorded as open follow-up in the new entry's `**Consequences:**` and called out in the PR body, so it's a written decision rather than a silent pass. Happy to do it as a follow-up if you'd rather it land with this.

## 6 — four claims in the body that don't hold · **fixed**

All four confirmed, and the body is rewritten:
- Collation claim — dropped (see 2).
- Prettier row was vacuous — confirmed, `.prettierignore` exempts `.claude`, `.context`, and `CLAUDE.md`. The body now says so explicitly instead of claiming a passing check.
- The grep check — confirmed broken, and for the reason you gave: it catches the fenced `## adr-{slug}` template, which appears first but sorts last because `{` > `z`. The body now warns about it.
- Entry count — confirmed wrong in both places. It's 58 now (57 from `main` + the new entry); the body says so and the squash message is corrected.

## 7 — the scout paraphrases a standard · **fixed**

Confirmed against `.claude/rules/agent-authoring-link-dont-paraphrase.md`, which is `paths:`-scoped to `.claude/agents/**`. Trimmed the 9-line paraphrase back to a link plus a one-line purpose hint. Your point that `b7ce4efbe8b` exists *because* the drift already happened is the part that convinced me — that commit was me hand-propagating a terminology change into a copy. I kept one line the rule does allow as agent-specific decision logic ("never propose an approach held by a retired slug"), since dropping it loses actual behavior rather than a restatement.

## 8 — `merge=union` recorded as live but unsafe · **fixed**

You're right that it's line-wise and would interleave prose entry bodies with no conflict marker. Recorded as **rejected** in the new entry's `**Alternatives:**`, with the reason, rather than "open and would compose with it".

## 9 — open branches will conflict · **fixed**

#2730 landed, so that instance is resolved by this PR's merge. I re-checked the current set: 21 open PRs touch the file (`#2768 #2766 #2765 #2760 #2758 #2756 #2752 #2751 #2750 #2749 #2747 #2746 #2745 #2744 #2742 #2737 #2736 #2734 #2700 #2632 #2551`). That list plus the resolve-by-slug instructions are now a prominent section of the PR body. Will post to the team channel before this merges.

## 10, 11, 13 — **fixed**

- **10** — confirmed 57 of 58 carry a `**Date:**`. Reworded to "Chronology lives in each entry's `**Date:**` rather than in the file order, so date every new entry" — an instruction rather than a false claim about the current contents. The one exception is the withdrawn tombstone; left undated, and the body now says why.
- **11** — confirmed exactly 28 `**Formerly:**` entries, ADR-0001–ADR-0028, no gaps, no duplicates, and 19 entries dated pre-switch carry none. Stated as the range, as you suggested, which is checkable.
- **13** — confirmed 3,059 lines / 268 KB. Dropped "read the log wholesale"; it now says to search by slug and topic and notes the size.

## 12, 14, 16 and the nits — **fixed**

- **12** — `CLAUDE.md` stated the ordering rule three times. Trimmed the duplicated rationale and gave the space to the conflict-resolution instruction instead, per your traffic argument.
- **14** — the caveat now says entries average ~55 lines and only same-gap insertions collide, so it reads as "rarer" rather than "barely helps".
- **16** — the `**Formerly:** ADR-0024` removal now has its own section in the body with the archaeology, so the next reader doesn't have to redo it.
- **Nits** — the line-38/line-966 duplication is gone (the "How to use it" bullet is now a short rule pointing at the new entry, which holds the rationale); "remain open" is dated; `add-only`/byte-order wording is now consistent across all three files; and the scout's deletion-vs-rewrite muddle went with the trim in 7.

## Verification

On the merged result: 58 entries, headings unique and in exact `LC_ALL=C sort` order; entry-by-entry byte comparison against `origin/main` shows the same 57 entries with byte-identical bodies plus the one new entry, the only two differing bodies being the two deliberately edited; the merge added 171 lines and removed 0, matching #2730's insertion exactly; all 19 `adr-*` references resolve, zero dangling; zero consecutive blank-line runs.

No `typecheck`/`lint`/`test` run — all changes are markdown prose and none of those commands read markdown. Not claiming a Prettier pass either, for the reason in 6.

katherinejensen00 (approved):
Thanks for making those fixes, Jolie! Looks great!

> @katherinejensen00 reviewed 3 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2770#-:-P0hyWPT70gKIRmiWPpA:b-2mjieu)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

Issue #200, opened 2023-05-18
Implement security measures to ensure safe execution of plugin DLLs and prevent malicious behavior


tombogle:
At a minimum, the extensions "store" should alert users to extensions that include plugin DLLs since it will be very difficult to ensure that such plugins are "safe."
It is not yet clear whether this will be necessary/worthwhile, but we could potentially write code to verify that plugin DLLs are not doing direct file access, Web access, running custom unmanaged code or doing other things that might be considered risky or open the door for malicious behavior. If we do this, we could either prevent such DLLs entirely or at least strengthen the warning message to alert users.

PR #2771, opened 2026-09-04
PT-4483: Fix syncProjects doc comments to match actual behavior

tombogle:
> **Update (2026-09-22):** The summary below was generated by `/review-paratext` on 2026-09-04
> against the original 2-file doc fix. Several more review rounds since then grew this PR to 7
> files — see the PR conversation for the full history (dedup-claim removal + PT-4602, the
> `paratext-bible-internal-extensions#200` wording sync, and stale-comment fixes elsewhere in
> core). Specifics that make the auto-generated summary below stale:
>
> - **Files changed** is now 7, not 2: added
>   `.claude/rules/architecture/closed-source-command-docs.md` (new doc-altitude rule) and an ADR
>   entry in `.context/standards/Architecture-Decisions.md`; fixed stale `syncProjects`-behavior
>   comments in `src/main/startup-readiness.util.ts`, `src/main/startup-tasks.ts`, and
>   `src/main/startup-tasks.test.ts`.
> - The Overview's third bullet below (the "stops downloading... trying a small initial batch
>   first, then the rest one at a time" phrasing) describes an intermediate draft's
>   implementation-detail wording, not what actually landed. The final doc states only the
>   caller-visible guarantee — callers must not assume every shared project is present locally
>   once the call resolves — per the doc-altitude principle this PR's review established (see
>   `.claude/rules/architecture/closed-source-command-docs.md`).
> - **Positive Observations**' "only these two files changed" claim reflects the 2026-09-04
>   2-file state and no longer holds.
> - The "Deduplication is handled internally" claim isn't mentioned anywhere below (it postdates
>   this summary) — it was added in a later round, found to be inaccurate, and removed; see
>   PT-4602 for the underlying (separately tracked) implementation gap.

# Code Review Summary

**Branch**: pt-4483-fix-syncprojects-firsttimesync-docs

**Base**: origin/main

**Date**: 2026-09-04

**Files changed**: 2

## Overview

Fixes stale/inaccurate doc comments on `paratextBibleSendReceive.syncProjects` — the C# XML doc on
the paranext-core stub (`ParatextProjectSendReceiveService.cs`) and the TS `.d.ts` declaration
(`src/@types/paratext-bible-send-receive/index.d.ts`) — so they describe the actual Paratext 10
Studio behavior rather than a simplified/incorrect summary. Split out from the now-closed PR #2640
(`scope-sync-to-picker-decision`) as a small, standalone doc fix under PT-4483, since the doc
inaccuracies were discovered while re-evaluating that PR's review feedback and don't need to wait
for the larger rework.

Four things were corrected, verified directly against the closed-source Paratext 10 Studio patch
source (`paratext-10-studio` repo, `repo-patches/paranext-core.patch`, `SyncProjectsCore`):

- Explicit `projectIds` sync regardless of local presence — a `new` (not yet downloaded) project
  among them is downloaded, not skipped, contrary to what the old doc claimed.
- `undefined`/`null` with at least one project already locally present syncs every locally-present
  project and leaves new ones alone (unchanged from the old doc — this part was already correct).
- `undefined`/`null` when every known shared project is currently `new` (a true first sync, or the
  rarer case where every previously-local project has since gone missing) does **not**
  unconditionally sync the whole account — it stops downloading as soon as one synced project gives
  the current user a non-Observer role, trying a small initial batch first, then the rest one at a
  time. An earlier draft of this fix claimed "every one of them is synced," which was wrong; this
  was caught and corrected in review before landing.
- The C# doc's `<summary>` had an entirely fictitious claim — "filters for editable projects and
  S/Rs them" / "Non-editable ... IDs are skipped" — with no such filtering logic anywhere in the
  actual sync selection or connected-resource-scan code. Removed; the TS doc's summary never had
  this claim.

No executable code changed in either file — this is a documentation-only fix.

## API Changes

- `c-sharp/Projects/SendReceive/ParatextProjectSendReceiveService.cs`: No signature change.
  `SyncProjects(String[]? projectIds)` doc comment rewritten — removes the false "filters for
  editable projects" claim, adds explicit-ID/null-with-locals/null-first-sync/empty-array behavior
  detail.
- `src/@types/paratext-bible-send-receive/index.d.ts`: No signature change. `@param projectIds` doc
  for `'paratextBibleSendReceive.syncProjects'` rewritten with the same four cases, plus a summary
  wording tweak ("each synced project's connected resources") and a softened parenthetical about
  what passing `undefined` does, to stay consistent with the new detail below it.

Neither file is part of the public API surface tracked for breaking-change purposes
(`lib/platform-bible-react/`, `lib/platform-bible-utils/`, or `papi.d.ts`) — `index.d.ts` augments
`papi-shared-types` for core's own internal command-handler typing. `papi.d.ts` itself is unchanged.

None.

None.

### Minor — Consider

- [x] **`index.d.ts:468` (pre-fix) — top-of-doc summary said "reads each project's connected resources," inconsistent with the C# summary's new "reads each synced project's connected resources."** _(fixed during review: reworded to "reads each synced project's connected resources" to match)_
- [x] **`index.d.ts:475` (pre-fix) — the parenthetical "(e.g. the startup sync passes `undefined` to mean 'sync all')" oversimplified/contradicted the new detailed bullets just below it, since `undefined` does not always sync literally "all" projects.** _(fixed during review: reworded to point at the detailed cases below instead of asserting a simplified behavior — "the startup sync passes `undefined` for its zero-state bootstrap case — see the cases below for what that actually syncs")_

Author was aware of finding #1 before this review but hadn't judged it worth a separate fix; asked me to use best judgment for both, and confirmed both fixes as made. No items dismissed.

#### Shared Regions Modified

None.

#### Extension Config Changes

None.

## Positive Observations

- Correctly removes the C# doc's inaccurate "filters for editable projects" claim — confirmed absent from both the visible stub and the actual Studio patch's selection/connected-resource-scan logic.
- The C# `<list type="bullet">` XML structure is well-formed and matches the pattern already used elsewhere in `c-sharp/Projects/`; the TS bullet list matches this same file's existing `- key = description` style (e.g. `ResultStatus`).
- Both C# and TS docs were kept in sync with each other in substance across all four cases (explicit IDs, null-with-locals, null-first-sync-bootstrap, empty-array), so a C# maintainer and a TS extension developer get the same picture.
- Confirmed via `git diff --name-only` that only these two files changed — no incidental logic, test, or config changes riding along; genuinely small and focused as described.
- No indecipherable initialisms introduced (`S/R`, `Observer` were already established terms in this codebase).
- No backward-facing PT-4483/#2640 references landed in the code itself — those identifiers appear only in the commit message, which is the correct place for them.
- Both formatters (`dotnet csharpier`, `prettier`/ESLint) pass clean on both files.

## Interview Notes

**Stated purpose**: Fix stale/inaccurate doc comments on `paratextBibleSendReceive.syncProjects` (C#
XML doc + TS `.d.ts`) so they describe the actual Paratext 10 Studio behavior — the `isFirstTimeSync`
bootstrap-download carve-out, explicit-ID handling, empty-array no-op, and removal of a false
"filters for editable projects" claim. Split out from the now-closed PR #2640 as a small, focused
doc fix under PT-4483.

**Provenance**: This fix emerged from a detailed re-evaluation of PR #2640's Matt-authored review
(36+ inline comments), which surfaced that the `syncProjects` contract docs were stale relative to
the actual P10 Studio implementation. During that investigation the author corrected two drafted
claims in real time by checking the actual closed-source patch source directly (not just this
repo's stub or the reviewer's prose): first, that a naive "syncs everything on first sync" claim was
wrong (the real behavior stops early once an editable project is found, via a batch-of-5-then-
one-at-a-time heuristic), and second, that "filters for editable projects" in the old C# summary has
no corresponding code at all. Both corrections are reflected in the committed doc text, not just
caught and discarded.

**Verification note for the reviewer**: the `api` analysis agent flagged (as `DONE_WITH_CONCERNS`)
that it could not verify the new doc text against the real `isFirstTimeSync`/batching implementation,
since that logic lives in the closed-source `paratext-10-studio` patch, not this repository. This is
accurate as a statement about what's checkable from this repo alone — but the author did verify the
new text directly against that patch's actual source (`repo-patches/paranext-core.patch`,
`SyncProjectsCore`) during this session, including catching and correcting the two inaccuracies
above before they landed. Worth noting in the PR description or review meeting so the reviewer knows
this wasn't left unverified, just verified outside this repo's visibility.

**No unresolved items.** No Critical/Important findings were raised. Minor findings were both fixed
per the author's "use your best judgment" instruction. Author had nothing further to flag at wrap-up.

## In-Review Quality Check

- **`npm run typecheck`** — PASSED, all workspaces, no fixes needed.
- **`npm run lint`** — PASSED (exit 0 across all workspaces). One pre-existing, unrelated warning
  surfaced in `lib/platform-bible-utils/src/lifetime-management/unsubscriber-async-list.test.ts`
  (`import/no-named-as-default`) — not touched by this branch, not an error, not fixed.
- **Prettier on changed files** — ran; no additional changes needed, both files already correctly
  formatted.
- **`npm test`** — 3003 passed / 1 failed / 1 skipped, 211/212 test files passed. The one failure
  (`keyboard-shortcuts-catalog.component.test.tsx`, a `Hook timed out in 10000ms` during `afterEach`)
  is unrelated to this branch's two doc-comment files; re-run in isolation, all 7 tests in that file
  passed in ~11s. Consistent with resource contention from running immediately after long
  typecheck/lint passes, not a real regression. Not fixed (unrelated, non-reproducing).

No unfixable failures traceable to the in-review changes.

## Suggested Review Focus

- [ ] Given this doc's behavior can't be checked against implementation from within this repo alone
      (the real logic lives in the private Studio patch), a second pair of eyes with direct access to
      `paratext-10-studio`'s `repo-patches/paranext-core.patch` (`SyncProjectsCore`) confirming the
      four-case description still matches current `main` there would be the most valuable check —
      the author already did this once during this session, so this is a re-confirmation rather than
      a first pass.
- [ ] No other open questions — this is a small, self-contained doc fix with no logic changes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2771)

### DIFF
--- a/.claude/rules/architecture/closed-source-command-docs.md
+++ b/.claude/rules/architecture/closed-source-command-docs.md
@@ -1,0 +1,54 @@
+# Doc Altitude for Closed-Source-Backed Commands
+
+Some PAPI commands are declared in this repo (a C# stub, a `.d.ts` augmentation) but only really
+implemented by a closed-source consumer — today, Paratext 10 Studio's private
+`repo-patches/paranext-core.patch`. The doc comment's `@throws PlatformUnimplementedException if
+not running in an application that implements this command (e.g., Paratext 10 Studio)` phrasing
+already says as much: Studio is named as an example implementer, not as the contract.
+
+## The rule
+
+For a command in this position, the XML doc / TSDoc describes **caller-visible guarantees** — what
+a caller may rely on, and what it must not assume — never the current implementer's specific
+mechanism, tuning constants, or internal heuristics.
+
+- **Write:** "callers must not assume every shared project is present locally once this resolves."
+- **Don't write:** "syncs an initial batch of 5, then tries the rest one at a time until it finds a
+  project with a non-Observer role" — that is Studio's current implementation, not the command's
+  contract.
+
+## Why
+
+- **This repo can't verify it.** The real logic lives in a private patch this repo doesn't contain
+  and can't run tests against (see `adr-closed-source-command-doc-altitude` in
+  `Architecture-Decisions.md`). A doc that asserts specific constants is a claim this repo has no
+  way to keep honest — the next time Studio's patch is regenerated (`save-repo-patches`), the doc
+  here has no signal that it drifted.
+- **It's the wrong altitude even before drift.** A command's public doc should describe the command,
+  not one build of one implementer's current approach to satisfying it. If a second white-label app
+  ever implements this contract differently, implementation-specific prose here becomes flatly wrong
+  for that caller while looking authoritative.
+- **Guarantee-level prose is more durable AND more actionable than a bare pointer elsewhere.**
+  "See the other repo for details" doesn't work for readers of this repo (including a future AI
+  agent) who might not have access to or knowledge of the actual implementation(s), and doesn't
+  survive a swapped implementer either. Stating what a caller may/must-not assume is the one thing
+  that stays true regardless of which conforming implementation is running, and it's exactly the
+  information a caller needs to write correct code against the command.
+
+## How to apply
+
+When documenting (or reviewing a doc change to) a command whose real behavior lives in
+`paratext-bible-send-receive` or any other closed-source/swappable extension:
+
+1. State what always holds for **any** conforming implementation — not what the current one happens
+   to do.
+2. If you're tempted to write a specific number, threshold, or named internal check, ask whether a
+   caller actually needs that number to use the command correctly, or whether a weaker "may/must not
+   assume X" statement covers the same ground without pinning down how it's achieved.
+3. Don't invent requirements the contract doesn't actually have, either — "must" language should
+   reflect a real invariant every implementation is expected to honor (e.g., "must throw
+   `PlatformUnimplementedException` if not implemented"), not a guess dressed up as a guarantee.
+
+See `adr-closed-source-command-doc-altitude` in
+[`Architecture-Decisions.md`](../../../.context/standards/Architecture-Decisions.md) for the
+decision history and the specific case (`paratextBibleSendReceive.syncProjects`) that surfaced this.
--- a/.context/standards/Architecture-Decisions.md
+++ b/.context/standards/Architecture-Decisions.md
@@ -400,6 +400,60 @@ step, no automation. Just a record.
 - **Source:** PRD "Saroj easily works with character-level markers" (appetite 2 developer weeks);
   character-marker removal work on `remove-character-marker`.
 
+## adr-closed-source-command-doc-altitude: Command docs for closed-source-backed PAPI commands describe caller-visible guarantees, not the current implementer's mechanism
+
+- **Date:** 2026-09-05
+- **Status:** Accepted
+- **Context:** PR #2771 (PT-4483) fixed `paratextBibleSendReceive.syncProjects`'s doc comments (the
+  C# stub's XML doc and `src/@types/paratext-bible-send-receive/index.d.ts`'s TSDoc), which had
+  drifted from the real Paratext 10 Studio implementation — a closed-source patch
+  (`paratext-10-studio` `repo-patches/paranext-core.patch`, `SyncProjectsCore`). A first draft of the
+  fix encoded Studio's specific first-sync mechanism directly (an initial batch of 5 projects, then
+  one at a time, stopping once a non-Observer role is found), verified against the actual patch
+  source. Code review (Reviewable, `katherinejensen00`, with direct access to that patch) found the
+  new text itself over-generalized: for accounts with five or fewer shared projects, or accounts
+  where the user has no editable role on any of them, the described "stops early, rather than
+  syncing the whole account" claim doesn't hold — the whole account downloads regardless, just via a
+  different path through the same mechanism. Fixing that specific overgeneralization in place would
+  still have left the doc asserting Studio's tunable constants (batch size, the role check) as part
+  of the command's contract — constants this repo cannot verify, cannot test against (the stub
+  throws/no-ops; no core test exercises the real logic), and has no way to detect drifting the next
+  time Studio's patch is regenerated (`save-repo-patches`).
+- **Decision:** Doc comments for a command whose real implementation is closed-source or otherwise
+  swappable (today: anything backed by `paratext-bible-send-receive`, i.e. any command whose
+  `@throws` names `PlatformUnimplementedException` "if not running in an application that implements
+  this command (e.g., Paratext 10 Studio)") describe caller-visible **guarantees** only — what a
+  caller may rely on and must not assume — never the current implementer's specific mechanism or
+  tuning constants. Concretely, `syncProjects`'s zero-local-projects case now reads "an
+  implementation is expected to try to make at least one project available for the current user to
+  work in, if the account has one — but may stop short of downloading every shared project in the
+  account, trading completeness for performance. Callers MUST NOT assume every shared project is
+  present locally once this resolves," replacing the batch/role-check description. Promoted to a
+  standing rule: `.claude/rules/architecture/closed-source-command-docs.md`.
+- **Alternatives:**
+  - **Encode the known implementation for convenience** (what the first draft did) — rejected:
+    useful once, but ties this repo's doc to a private repo's tunable constants with no verification
+    path and no drift signal; exactly what produced the overgeneralization this decision responds
+    to.
+  - **Keep it vague, point readers to the other repo** — rejected: readers of this repo (including a
+    future AI agent) might not have access to or knowledge of the actual implementation(s); a bare
+    pointer elsewhere is unactionable for them and doesn't tell a caller what it can safely assume.
+  - **Leave the pre-PR wording**, which understated the behavior — rejected: it was the original bug
+    this PR fixed (claimed the no-ID form only ever syncs already-local projects, when a true first
+    sync can and should also acquire the account's first project).
+- **Consequences:** This doc can no longer be invalidated by a Studio-side regeneration of
+  `repo-patches/paranext-core.patch` that changes a tuning constant, since it no longer asserts one.
+  The tradeoff is genuinely less specific information in-repo for someone who wants to reason
+  precisely about first-sync latency or batching — that detail now only exists in the Studio patch
+  itself, which the rule file points to. If a future need arises to expose implementation-specific
+  timing/behavior to core (e.g. for a startup-performance budget), it should be surfaced through an
+  explicit signal (a return value, an event) rather than encoded into a doc comment describing a
+  different implementation's internals. Also unresolved from the same review round: upstreaming the
+  `syncProjects` doc fix to `paratext-bible-internal-extensions`'s own `.d.ts` (per this file's
+  header policy on re-syncs) so a future PT-4233 re-sync doesn't silently revert it — tracked as
+  follow-up, not done as part of this decision.
+- **Source:** PT-4483, review of #2771.
+
 ## adr-decision-log-sorted-insertion: Decision-log entries are inserted in byte order by slug, not appended
 
 - **Date:** 2026-09-03
--- a/c-sharp/Projects/SendReceive/ParatextProjectSendReceiveService.cs
+++ b/c-sharp/Projects/SendReceive/ParatextProjectSendReceiveService.cs
@@ -215,17 +215,29 @@ internal class ParatextProjectSendReceiveService(
     }
 
     /// <summary>
-    /// Syncs projects from the provided IDs: filters for editable projects and S/Rs them,
-    /// then reads each editable project's connected resources and projects (one level deep —
-    /// connections of connections are not included) and S/Rs connected translation projects
-    /// or DBL-updates connected resources. Non-editable and unknown IDs are skipped.
-    /// Deduplication is handled internally.
+    /// Syncs the given projects (S/Rs them), then reads each synced project's connected resources
+    /// and projects (one level deep — connections of connections are not included) and S/Rs
+    /// connected translation projects or DBL-updates connected resources as needed. Unknown IDs are
+    /// skipped. Deduplication is handled internally.
     /// Exception is thrown if this function is not implemented in the current application
     /// or if an error was encountered syncing.
     /// </summary>
     /// <param name="projectIds">
-    /// IDs of the projects to sync. If <see langword="null"/>, all shared projects that are already
-    /// present locally (i.e., not new) are synced. An empty array is a no-op.
+    /// IDs of the projects to sync.
+    /// <list type="bullet">
+    /// <item>If explicit IDs are given, each one is synced regardless of whether it is already
+    /// present locally — a <c>new</c> (not yet downloaded) project among them is downloaded, not
+    /// skipped.</item>
+    /// <item>If <see langword="null"/> and at least one shared project the account knows about is
+    /// already present locally (not new), every locally-present project is synced; new projects
+    /// are left alone.</item>
+    /// <item>If <see langword="null"/> and every shared project the account knows about is
+    /// currently new (a true first sync, or the rarer case where every previously-local project
+    /// has since gone missing from disk), downloading stops as soon as one synced project gives
+    /// the current user a non-Observer role — trying a small initial batch, then the rest one at a
+    /// time — rather than unconditionally syncing the whole account.</item>
+    /// <item>An empty array is a no-op.</item>
+    /// </list>
     /// </param>
     protected void SyncProjects(String[]? projectIds)
     {
--- a/c-sharp/Projects/SendReceive/ParatextProjectSendReceiveService.cs
+++ b/c-sharp/Projects/SendReceive/ParatextProjectSendReceiveService.cs
@@ -215,10 +215,10 @@ internal class ParatextProjectSendReceiveService(
     }
 
     /// <summary>
-    /// Syncs the given projects (S/Rs them), then reads each synced project's connected resources
-    /// and projects (one level deep — connections of connections are not included) and S/Rs
-    /// connected translation projects or DBL-updates connected resources as needed. Unknown IDs are
-    /// skipped. Deduplication is handled internally.
+    /// Syncs the given projects (S/Rs them), then reads connected resources and projects (one level
+    /// deep — connections of connections are not included) for the project(s) this call settles on,
+    /// and S/Rs connected translation projects or DBL-updates connected resources for those as
+    /// needed. Unknown IDs are skipped. Deduplication is handled internally.
     /// Exception is thrown if this function is not implemented in the current application
     /// or if an error was encountered syncing.
     /// </summary>
@@ -227,10 +227,12 @@ internal class ParatextProjectSendReceiveService(
     /// <list type="bullet">
     /// <item>If explicit IDs are given, each one is synced regardless of whether it is already
     /// present locally — a <c>new</c> (not yet downloaded) project among them is downloaded, not
-    /// skipped.</item>
+    /// skipped, though an implementation may still exclude a given id for reasons outside the
+    /// caller's control (e.g., an unsupported project version, or the project being otherwise
+    /// ineligible).</item>
     /// <item>If <see langword="null"/> and at least one shared project the account knows about is
-    /// already present locally (not new), every locally-present project is synced; new projects
-    /// are left alone.</item>
+    /// already present locally (not new), every locally-present shared project is synced; new
+    /// projects are left alone.</item>
     /// <item>If <see langword="null"/> and no shared project the account knows about is present
     /// locally yet (whether a genuine first sync, or every previously-local project has since gone
     /// missing from disk), an implementation is expected to try to make at least one project
--- a/c-sharp/Projects/SendReceive/ParatextProjectSendReceiveService.cs
+++ b/c-sharp/Projects/SendReceive/ParatextProjectSendReceiveService.cs
@@ -231,11 +231,12 @@ internal class ParatextProjectSendReceiveService(
     /// <item>If <see langword="null"/> and at least one shared project the account knows about is
     /// already present locally (not new), every locally-present project is synced; new projects
     /// are left alone.</item>
-    /// <item>If <see langword="null"/> and every shared project the account knows about is
-    /// currently new (a true first sync, or the rarer case where every previously-local project
-    /// has since gone missing from disk), downloading stops as soon as one synced project gives
-    /// the current user a non-Observer role — trying a small initial batch, then the rest one at a
-    /// time — rather than unconditionally syncing the whole account.</item>
+    /// <item>If <see langword="null"/> and no shared project the account knows about is present
+    /// locally yet (whether a genuine first sync, or every previously-local project has since gone
+    /// missing from disk), an implementation is expected to try to make at least one project
+    /// available for the current user to work in, if the account has one — but may stop short of
+    /// downloading every shared project in the account, trading completeness for performance.
+    /// Callers MUST NOT assume every shared project is present locally once this resolves.</item>
     /// <item>An empty array is a no-op.</item>
     /// </list>
     /// </param>
--- a/src/@types/paratext-bible-send-receive/index.d.ts
+++ b/src/@types/paratext-bible-send-receive/index.d.ts
@@ -538,9 +538,9 @@ declare module 'papi-shared-types' {
      * Unknown project IDs are skipped. Deduplication is handled internally.
      *
      * This signature matches this repository's C# stub (`String[]? projectIds`, no return value),
-     * which core itself calls (e.g. the startup sync passes `undefined` for its zero-state
-     * bootstrap case — see the cases below for what that actually syncs). The Send/Receive
-     * extension's own declaration also returns the S/R results; core does not consume them.
+     * which core itself calls with `undefined` — see the cases below for what that actually syncs.
+     * The Send/Receive extension's own declaration also returns the S/R results; core does not
+     * consume them.
      *
      * @param projectIds IDs of the projects to sync.
      *
@@ -549,11 +549,12 @@ declare module 'papi-shared-types' {
      *       are skipped.
      *   - If omitted and at least one shared project the account knows about is already present locally
      *       (not new), every locally-present project is synced; new projects are left alone.
-     *   - If omitted and every shared project the account knows about is currently new (a true first
-     *       sync, or the rarer case where every previously-local project has since gone missing
-     *       from disk), downloading stops as soon as one synced project gives the current user a
-     *       non-Observer role — trying a small initial batch, then the rest one at a time — rather
-     *       than unconditionally syncing the whole account.
+     *   - If omitted and no shared project the account knows about is present locally yet (whether a
+     *       genuine first sync, or every previously-local project has since gone missing from
+     *       disk), an implementation is expected to try to make at least one project available for
+     *       the current user to work in, if the account has one — but may stop short of downloading
+     *       every shared project in the account, trading completeness for performance. Callers MUST
+     *       NOT assume every shared project is present locally once this resolves.
      *   - An empty array is a no-op.
      *
      * @throws `PlatformUnimplementedException` if not running in an application that implements
--- a/src/@types/paratext-bible-send-receive/index.d.ts
+++ b/src/@types/paratext-bible-send-receive/index.d.ts
@@ -532,10 +532,11 @@ declare module 'papi-shared-types' {
     'paratextBibleSendReceive.commitDaily': (projectId: string) => Promise<void>;
 
     /**
-     * Syncs projects: sends/receives each project, then reads each synced project's connected
-     * resources and projects (one level deep — connections of connections are not included) and
-     * sends/receives connected translation projects or DBL-updates connected resources as needed.
-     * Unknown project IDs are skipped. Deduplication is handled internally.
+     * Syncs projects: sends/receives each project, then reads connected resources and projects (one
+     * level deep — connections of connections are not included) for the project(s) this call
+     * settles on, and sends/receives connected translation projects or DBL-updates connected
+     * resources for those as needed. Unknown project IDs are skipped. Deduplication is handled
+     * internally.
      *
      * This signature matches this repository's C# stub (`String[]? projectIds`, no return value),
      * which core itself calls with `undefined` — see the cases below for what that actually syncs.
@@ -545,10 +546,12 @@ declare module 'papi-shared-types' {
      * @param projectIds IDs of the projects to sync.
      *
      *   - If provided, each given ID is synced regardless of whether it is already present locally — a
-     *       `new` (not yet downloaded) project among them is downloaded, not skipped. Unknown IDs
-     *       are skipped.
+     *       `new` (not yet downloaded) project among them is downloaded, not skipped, though an
+     *       implementation may still exclude a given id for reasons outside the caller's control
+     *       (e.g., an unsupported project version, or the project being otherwise ineligible).
      *   - If omitted and at least one shared project the account knows about is already present locally
-     *       (not new), every locally-present project is synced; new projects are left alone.
+     *       (not new), every locally-present shared project is synced; new projects are left
+     *       alone.
      *   - If omitted and no shared project the account knows about is present locally yet (whether a
      *       genuine first sync, or every previously-local project has since gone missing from
      *       disk), an implementation is expected to try to make at least one project available for
--- a/src/@types/paratext-bible-send-receive/index.d.ts
+++ b/src/@types/paratext-bible-send-receive/index.d.ts
@@ -542,9 +542,20 @@ declare module 'papi-shared-types' {
      * Send/Receive extension's own declaration also returns the S/R results; core does not consume
      * them.
      *
-     * @param projectIds IDs of the projects to sync. If omitted, all shared projects that are
-     *   already present locally (i.e., not new) are synced. If provided, only projects already
-     *   present locally are synced; new projects (not yet received) and unknown IDs are skipped.
+     * @param projectIds IDs of the projects to sync.
+     *
+     *   - If provided, each given ID is synced regardless of whether it is already present locally — a
+     *       `new` (not yet downloaded) project among them is downloaded, not skipped. Unknown IDs
+     *       are skipped.
+     *   - If omitted and at least one shared project the account knows about is already present locally
+     *       (not new), every locally-present project is synced; new projects are left alone.
+     *   - If omitted and every shared project the account knows about is currently new (a true first
+     *       sync, or the rarer case where every previously-local project has since gone missing
+     *       from disk), downloading stops as soon as one synced project gives the current user a
+     *       non-Observer role — trying a small initial batch, then the rest one at a time — rather
+     *       than unconditionally syncing the whole account.
+     *   - An empty array is a no-op.
+     *
      * @throws `PlatformUnimplementedException` if not running in an application that implements
      *   this command (e.g., Paratext 10 Studio)
      */
--- a/src/@types/paratext-bible-send-receive/index.d.ts
+++ b/src/@types/paratext-bible-send-receive/index.d.ts
@@ -532,15 +532,15 @@ declare module 'papi-shared-types' {
     'paratextBibleSendReceive.commitDaily': (projectId: string) => Promise<void>;
 
     /**
-     * Syncs projects: sends/receives each project, then reads each project's connected resources
-     * and projects (one level deep — connections of connections are not included) and
+     * Syncs projects: sends/receives each project, then reads each synced project's connected
+     * resources and projects (one level deep — connections of connections are not included) and
      * sends/receives connected translation projects or DBL-updates connected resources as needed.
      * Unknown project IDs are skipped. Deduplication is handled internally.
      *
      * This signature matches this repository's C# stub (`String[]? projectIds`, no return value),
-     * which core itself calls (e.g. the startup sync passes `undefined` to mean "sync all"). The
-     * Send/Receive extension's own declaration also returns the S/R results; core does not consume
-     * them.
+     * which core itself calls (e.g. the startup sync passes `undefined` for its zero-state
+     * bootstrap case — see the cases below for what that actually syncs). The Send/Receive
+     * extension's own declaration also returns the S/R results; core does not consume them.
      *
      * @param projectIds IDs of the projects to sync.
      *
--- a/src/main/startup-readiness.util.ts
+++ b/src/main/startup-readiness.util.ts
@@ -317,10 +317,12 @@ function shouldLogProbeAttempt(attempt: number): boolean {
  * An empty result counts as not-ready, not ready-with-no-projects. A workspace that genuinely has
  * no local scripture projects therefore waits the full budget. Note this probe and the eventual
  * sync measure different things: the probe checks for projects visible via
- * `SCRIPTURE_READINESS_PROJECT_INTERFACE`, while `syncProjects(undefined)` means "all shared
- * projects already present locally" — those sets can differ, so the delayed sync is not always a
- * no-op. When there genuinely are no local scripture projects, though, it typically has nothing to
- * do.
+ * `SCRIPTURE_READINESS_PROJECT_INTERFACE`, while `syncProjects(undefined)`'s scope depends on what
+ * is already local — those sets can differ, so the delayed sync is not always a no-op. When there
+ * are genuinely no local scripture projects — whether a true first sync, or every previously-local
+ * project having since gone missing — the delayed sync can be the MOST expensive case (trying to
+ * establish at least one usable project) rather than the least — so this probe's cost model should
+ * not assume the delayed sync is cheap there.
  *
  * Uses the WITHOUT-RETRIES lookup even though its TSDoc names layering PDP factories as the
  * intended caller and points most callers at `getMetadataForAllProjects`. That default is rejected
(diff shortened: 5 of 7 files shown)


katherinejensen00 (changes requested):
# Code review — paranext-core PR #2771

**PR:** [PT-4483: Fix syncProjects doc comments to match actual behavior](https://github.com/paranext/paranext-core/pull/2771)
**Author:** Tom Bogle · **Branch:** `pt-4483-fix-syncprojects-firsttimesync-docs` → `main`
**Size:** documentation-only, 2 files, +38/−15
**Review date:** 2026-09-04

## Files changed

- `c-sharp/Projects/SendReceive/ParatextProjectSendReceiveService.cs` — XML doc on the `SyncProjects` stub
- `src/@types/paratext-bible-send-receive/index.d.ts` — TS declaration for `paratextBibleSendReceive.syncProjects`

---

## Verdict first

**This PR is a clear net improvement and the author got the hard part right.** It kills a flatly false claim that had been sitting in the C# summary, and it correctly documents an early-stop behavior that the previous text described incorrectly. Both headline corrections survive direct verification against the real implementation.

What follows is narrower than the finding count suggests. **One item is genuinely worth fixing before merge** (bullet 3 over-generalizes); one is a small factual slip in a parenthetical; the rest is polish, follow-up scope, or a durable structural suggestion that would stop this class of drift recurring.

**Nothing here can break a build, a runtime, or data.** The worst outcome available is a developer misled by a comment, so nothing is graded above Important.

---

## Method

Four passes, reconciled:

1. **Developer review** (soundness) — every claim checked against the real implementation
2. **Architect review** (structure, altitude, completeness)
3. **Comment & documentation quality** review
4. **Antagonistic pass** against passes 1–3, then `/code-review max` (10 finder angles → dedup → 3 verifier batches → gap sweep)

### The decisive move

The PR body states its central claims "can't be checked against implementation from within this repo alone." **That gap is closable.** The implementation is on this machine:

```
C:/Users/kathe/paratext-10-studio/repo-patches/paranext-core.patch
```

Relevant symbols: `SyncProjectsCore` (~line 7071), `GetSharedProjects` / `BuildProjectSelectionOptions` (~6118–6270), `RunWithSyncNotification` (~7945).

Any future review of a `paranext-core` S/R stub doc should start there rather than treating the behavior as unverifiable.

### Corrections applied to this review

This report was revised after challenge. Recording what changed, because the original framing was unfair to the PR:

- **Withdrew** the finding on the added hedge "as needed" in the DBL clause. Tom *added* a hedge; a hedge weakens a claim and cannot introduce a false one. The un-hedged original claimed more. The criticism was backwards.
- **Withdrew the `[REGRESSION]` tag** on the connected-resources wording. Old text: "each **editable** project's"; new: "each **synced** project's." The loop applies no editability filter, so the new wording is *more* accurate on the dominant path — imperfect only in one carve-out branch where the old text happened to be right. That is an incomplete improvement, not a regression.
- **Merged** three findings into one. The root cause is not the `> 5` gate; it is that the first batch of up to 5 downloads unconditionally, and for a small account that batch *is* the whole account. Splitting it inflated the count.
- **Downgraded severities.** Two findings were originally marked Critical. For a documentation-only change that was severity inflation.
- **Reframed the summary.** The original read as "a doc-accuracy PR that ships false statements." That mischaracterizes work which makes the doc substantially more accurate than it was.

Two facts were re-verified after the challenge and **hold**: `if (isFirstTimeSync && projectIdsToSync.Count > 5)` at patch:7153 with the `else` arm sharing the full list; and `if (!firstRunComplete) return { run: false, … }` at `startup-tasks.ts:363`.

### Confidence caveat

Nearly every behavioral claim rests on the private Studio patch, not on anything in paranext-core. Line numbers below are **PR-head** numbers; the branch is 10 commits behind `main`, which shifts `index.d.ts` numbers roughly **+55** after a rebase.

---

## What the PR gets right

Verified directly against source — these are real corrections, not cosmetic:

- **There is no editable filter anywhere on the explicit-id path.** `SendReceiveProjectsCore` applies no role test, and the connected loop iterates every kept id. Removing "filters for editable projects and S/Rs them" fixed a statement that was flatly wrong on the dominant path. This is the most valuable line in the diff.
- **An explicitly-named `new` project really is downloaded.** The else-branch filter is `projectIds.Contains(kvp.Key)` with no `EditedStatus` test — so the old "only projects already present locally are synced" was wrong, and the new text is right.
- **The early-stop exists and the previous text denied it.** Documenting it at all is the right call; the issue below is scope of the claim, not its existence.

---

## Findings

### Worth fixing before merge

#### 1. Important — `c-sharp/Projects/SendReceive/ParatextProjectSendReceiveService.cs:153`
*(mirrors: `.cs:155`, `.cs:157`; `src/@types/paratext-bible-send-receive/index.d.ts:486`, `:488`, `:489`)*

**Bullet 3 over-generalizes: it presents the early stop as what always happens, but up to five projects always download first.**

Three sub-cases, all fixed by one edit:

- **Small accounts.** `var firstBatch = projectIdsToSync.Take(5).ToList()` is shared before any role is inspected (patch:7155), and `hasNonObserverRoleInFirstBatch` is evaluated only afterward. For an account with **5 or fewer** shared projects the first batch *is* the whole account, so nothing stops early — and "rather than unconditionally syncing the whole account" is false. (The `if (isFirstTimeSync && projectIdsToSync.Count > 5)` gate at patch:7153 routes these to the `else` arm, which shares the full list; the outcome is the same either way, which is why the batch — not the gate — is the real story.)
- **The floor is five, not one.** "Stops as soon as one synced project gives … a non-Observer role" understates it by five clones — the number anyone tuning the startup readiness budget actually needs.
- **All-Observer accounts.** When `firstEditableId` stays null, patch:7254 falls back to `syncedIds` (everything), each probe its own `SendReceiveProjectsCore(new List<string> { id })` round-trip — serially, so worse than the batched share it is contrasted against. Observer-only logins are a normal Paratext role. *Mildest of the three:* the doc does say "then the rest one at a time," so the fallback is disclosed; only the "rather than" clause overreaches.

Suggested replacement, which also resolves findings 3 and 6 and stops encoding tunable constants:

```
A first sync may download only a SUBSET of the account: it stops early once it has a
project the current user can work in, and only the projects it kept are then scanned for
connected resources. Do not assume every shared project is present when this resolves.
```

#### 2. Important — `src/@types/paratext-bible-send-receive/index.d.ts:475`

**"Zero-state bootstrap case" describes the wrong caller, and it replaced text that was accurate.**

`src/main/startup-tasks.ts:287` fires on *every* Simple-mode launch, and `evaluateSimpleModeSyncGates` hard-returns at `:363` on `if (!firstRunComplete)` — so it never runs during first run, and its normal case is the steady state (bullet 2). The genuine zero-state caller is two files away and goes unnamed: `src/renderer/components/first-run/steps/sync-consent-step.component.tsx:21`. The prior wording ("to mean 'sync all'") was accurate as a statement of caller intent.

*Scope note:* the startup sync *can* reach the zero-state path — a user who completes first run with sync declined, then later enables sync-on-startup, still has every project `new`. So the sentence is a mischaracterization of the caller's normal case, not an impossibility.

#### 3. Important — `ParatextProjectSendReceiveService.cs:144` — altitude
*(mirror: `index.d.ts:479`)*

**The bullets encode closed-source tuning constants that nothing in this repo can validate, which is *why* bullet 3 drifted.**

They bake in `Count > 5`, `Take(5)` and the `HaveRoleNotObserver` probe — all regenerated wholesale by `save-repo-patches` in the studio repo. The stub throws (or toasts in DEBUG), so no core test exercises it; the repo's only doc-assertion test (`c-sharp-tests/Projects/SendReceive/ParatextProjectSendReceiveServiceStubDocsTests.cs`) covers `breakSyncLock` only. Finding 1 exists *only* because of this choice, and the outcome-level wording above survives every retuning of the constants.

*Counterpoint worth weighing:* the pre-PR text described private behavior too, and finding 5 shows documenting it here has real value. This is an argument for changing the **altitude**, not for saying less.

#### 4. Important — `index.d.ts:6`

**The file header prescribes the process for this situation and the PR doesn't follow it, so the fix can be silently reverted.**

The header says this file is "Derived from" upstream at a pinned SHA with `TODO(PT-4233)`, and that a re-sync "must not downgrade a richer declaration here … merge the two, and **upstream the improvement instead**." Upstream `paratext-bible-internal-extensions/…/paratext-bible-send-receive.d.ts:623` still says "Omitted/`undefined` syncs all locally-known shared projects." Without an upstream PR, the next PT-4233 re-sync overwrites this work.

---

### Suggestions — not defects in this PR

#### 5. Minor — `src/main/startup-readiness.util.ts:320`

**A pre-existing comment states the superseded contract and draws a load-bearing inference this PR's bullet 3 inverts.**

It reads: "`syncProjects(undefined)` means 'all shared projects already present locally' … When there genuinely are no local scripture projects, though, it typically has nothing to do." Zero local projects is exactly `isFirstTimeSync` — the *maximum*-work path. The comment justifies letting a fresh workspace wait the full 120s readiness budget, so the wrong inference is doing real design work.

This is not something the PR introduced — it is stale text the PR's correction *reveals*. Fixing it here is optional scope, but it is the highest-value of the stale sites:

- `src/main/startup-tasks.ts:280-284` — attributes the "sync all" sentinel explicitly to "the C# `String[]? projectIds` contract" this PR rewrote
- `src/main/startup-tasks.ts:153`, `:170-172`, `:186-188`, `:313-315`
- `src/main/startup-tasks.test.ts:136`
- `.context/standards/Architecture-Decisions.md:2341`

#### 6. Minor — `ParatextProjectSendReceiveService.cs:137`
*(mirror: `index.d.ts:469`)*

**"Each synced project's connected resources" is an improvement that is imperfect in one branch.**

`projectIdsToSync = firstEditableId != null ? [firstEditableId] : syncedIds` (patch:7254) runs *before* the connected-resource `foreach`, so in that one branch several projects sync but only one is scanned. Patch comment: "we'll sync only this one project."

**Not a regression.** The replaced text said "each **editable** project's," which was wrong on the dominant path (no editability filter exists). New wording is right on the dominant path and wrong only in the carve-out; net accuracy improved. Folding "only the projects it kept" into the finding-1 rewrite closes the gap.

#### 7. Minor — `ParatextProjectSendReceiveService.cs:153`
*(mirror: `index.d.ts:486`)*

**The two `null` bullets read as an exhaustive partition, so an empty account satisfies bullet 3 vacuously.**

`isFirstTimeSync` also requires `allSharedProjectsInfo.Count > 0` (patch:7111); that conjunct is unrepresented. An account with no shared projects takes the else branch, yields an empty list, and returns at patch:7134. The deleted one-liner avoided this only by not enumerating cases at all — so this is a cost of a genuine improvement, not a defect introduced.

#### 8. Minor — `ParatextProjectSendReceiveService.cs:147`
*(mirror: `index.d.ts:481`)*

**"Each one is synced regardless of whether it is already present locally" over-generalizes — a *known* id can still be dropped silently.**

`CreateSharedProjectList` drops on three grounds with no error path: unsupported-version `continue` (patch:~9485), null `sharedProj` (~9482), and `.Where(sp => !sp.IsDisabled())` (~9593), where `IsDisabled` is the Paratext7/SharedParatext7 filter. Then patch:7134 returns null and the `Promise<void>` resolves having done nothing. The doc's only stated exclusion is unknown ids.

#### 9. Minor — `ParatextProjectSendReceiveService.cs:151`
*(mirror: `index.d.ts:485`)*

**"Every locally-present project is synced" dropped the "shared" qualifier the deleted one-liner carried.**

The set is `allSharedProjectsInfo` ∩ `EditedStatus != "new"` (patch:7128); its local half is `ScrTextCollection.ScrTexts(IncludeProjects.AllEditable).Where(p => p.IsProjectShared && …)` with a trailing `!IsDisabled()`.

#### 10. Nit — `index.d.ts:482`

**In-PR drift between two halves of a contract this file requires be hand-synced.**

`index.d.ts:472` says "Unknown project IDs are skipped."; `:482-483` repeats "Unknown IDs are skipped."; the C# states it once at `:139-140`. The blocks are otherwise bullet-for-bullet parallel, so a parity diff hits a false positive.

#### 11. Nit — `ParatextProjectSendReceiveService.cs:144` — coordination note, **no action required**

The two `<param>` lines this PR deletes are verbatim leading-context lines of the Studio patch's 2836-line hunk (header patch:6993 `@@ -146,68 +1289,2836 @@`). `syncProjects` appears nowhere in `paratext-10-studio` outside `repo-patches/`, so the patch holds the only copy of this doc there.

Explicitly **not** a blocker:

- `paratext-10-studio/AGENTS.md:62` documents the apply method as **`git apply --3way`**, not plain `git apply`. The pre-image blob `9f285bfe864` is a real blob in `paranext-core` history, so `--3way` does a genuine merge.
- The deleted lines are **unchanged leading context**. A 3-way merge of "core rewrote these two lines" against "studio changed nothing there" resolves in core's favour. No merge markers expected.
- The patch **already** trails `main` on this file: `main`'s blob is `88d7a7f79f5`, not `9f285bfe864`. Routine `save-repo-patches` drift.

Worth one PR-body line only because the patch keeps core's XML doc as *context* — whatever core ships becomes Studio's shipped doc for the real implementation.

#### 12. Nit — `ParatextProjectSendReceiveService.cs:155` — low confidence
*(mirror: `index.d.ts:488`)*

`PermissionManager.HaveRoleNotObserver` admits Consultant, while `RoleAlwaysReadonly` returns true for Consultant so `CanEdit` refuses; it also opens with `if (Data == null) return true;`, running right after a fresh clone. So the probe can stop on a project the user has no usable role in. **The doc text never claims editability** — it is the patch's own comment ("We found an editable project") that conflates them, outside this diff. Listed for completeness only.

---

### Pre-existing — legitimate follow-up scope

Not caused by this PR and not merge-blocking. Listed because they sit in sentences the PR edited or kept.

| # | Severity | Location | Issue |
|---|----------|----------|-------|
| 13 | Important | `index.d.ts:491` (mirror `.cs:158`) | **"An empty array is a no-op" is false at the observable level.** The early return sits *inside* the notification/semaphore scaffold: `RunWithSyncNotification` (patch:~7945-7963) does `_sendReceiveSemaphore.Wait(0)` and on failure throws `FailedPrecondition` ("A Send/Receive is already in progress"), arms the watchdog, and shows a "Syncing…" toast with Cancel — only then does patch:7073 return. `breakSyncLock` two doc blocks away (`index.d.ts:553`) states this precisely for its own empty case. |
| 14 | Important | `index.d.ts:472` (mirror `.cs:140`) | **"Deduplication is handled internally" over-claims.** `seenConnectedIds` (patch:~7283) is created empty and never seeded from `projectIdsToSync`, so a back-translation/base-project pair both in the account syncs twice in one call, and `resultsData.ResultsInfo[id] = resultInfo` silently overwrites the first pass's outcome. |
| 15 | Important | `.cs:141` | **"Exception is thrown … or if an error was encountered syncing" is wrong in both directions.** The `#if DEBUG` body at `:163-171` sends a notification and returns rather than throwing (every sibling stub throws unconditionally); and per patch:~6751 `resultsData.SyncFailed = !syncSucceeded` with "The share no longer throws on handled failures" — probe failures `continue` with a warning, connected-resource and DBL failures are explicitly non-fatal. With `Promise<void>`, a failed Send/Receive resolves successfully to every core caller. |
| 16 | Important | `index.d.ts:493` | **`@throws` names only `PlatformUnimplementedException`**, but the Studio path rejects with several others, two carrying typed codes callers branch on: `FailedPrecondition` (patch:~7945) and `Unavailable` (~7956), plus init/registration failures (~7011, ~7015). The patch's own comment says the extension treats `e.code === 'FAILED_PRECONDITION'` as "already running" and backs off. |
| 17 | Minor | `.cs:139` (mirror `index.d.ts:482`) | **"Unknown IDs are skipped" flattens two behaviors with very different cost.** A non-HexId makes scoping collapse: patch:6243 filters via `HexId.FromStrSafe`, and when *all* ids fail, patch:6248 traces a warning and returns null, so `GetSharedProjects` runs the expensive sequential modification-status check over the whole account before matching nothing. A caller passing short names gets a slow silent no-op. |
| 18 | Minor | `.cs:140` (whole block `index.d.ts:468-496`) | **Cancellation is undocumented, and the branch this PR newly documents is the cancellable one.** `IsAbortRequested()` at patch:7137, ~7196, ~7295; `RunWithSyncNotification` swallows `ProgressCancelledException` as a clean cancel, so `await sendCommand(…)` resolves with `undefined` — indistinguishable from a complete sync. `cancelSync` is documented three methods below with no cross-reference either way. |
| 19 | Minor | `index.d.ts:469` (mirror `.cs:137`) | **Neither doc mentions that this command blocks edits on a set including projects the caller never named, which grows mid-sync:** `ReArmSyncGate(armedGateIds)` at patch:~7145 and again after `armedGateIds.UnionWith(idsForSendReceive)` at ~7355. Per the repo's Send/Receive Write Gate contract an armed project rejects writes with `(SR_EDIT_BLOCKED)` — so `syncProjects(['A'])` while editing B, which A references, starts failing B's edits partway through. |
| 20 | Minor | `index.d.ts:290` | **The sibling block says "Two core paths are on the invisible side today"; it's now three.** `sync-consent-step.component.tsx:21` also calls dotnet `syncProjects` directly, so it is equally invisible to `getSyncState`/`onSyncStateChanged`. Mirrored verbatim in `src/renderer/hooks/use-sync-status.hook.ts:268-271`. |
| 21 | Minor | `index.d.ts:471` (mirror `.cs:139`) | **"Sends/receives connected translation projects" is more definite than the code.** `ClassifyConnectedResourceReference` (patch:~6303) routes purely on reference *type* — any `ProjectReference`, translation project or not — and its own doc notes unrecognized ids are "silently excluded from the share rather than causing an error." |
| 22 | Minor | `index.d.ts:469` (mirror `.cs:137`) | **The sentence reads as a per-project read-then-sync loop;** the implementation reads all, then shares once (patch:~7360), with one aggregate verdict. On failure `BuildConnectedFailureWarnings` names *every* id including the successes. |
| 23 | Minor | `index.d.ts:474` | **"No return value" attributes to a TS-declaration difference what is a real runtime difference.** Patch:6998 is `protected async Task<SendReceiveResultsData?> SyncProjects(string[]? projectIds)` — the dotnet command returns results over the wire and `Promise<void>` discards them, including the warnings the code carefully builds. |
| 24 | Minor | `.cs:136` | **The doc now describes whole-repository downloads with no duration note**, and that cost is written 100 lines above on the constant this registration carries: `:31` `s_sendReceiveTimeout = TimeSpan.FromSeconds(0); // 0 = no timeout`, documented as "a whole-project sync is a multi-minute server operation … deliberately unbounded." `sync-consent-step.component.tsx` awaits this with `setCanSkip?.(false)`. |
| 25 | Minor | `.cs:48` | **None of the ~25 new doc lines reaches a runtime consumer.** `RegisterRequestHandlerAsync("command:paratextBibleSendReceive.syncProjects", …)` passes no `documentation:`, while `breakSyncLock` at `:57-80` does and has a test asserting it over `rpc.discover`. (`src/@types` is excluded from `papi.d.ts`.) If taken: `ExperimentalMethodDocumentation.Create` hard-sets `Experimental = true` and this command isn't experimental, so the record needs building directly. |
| 26 | Nit | `index.d.ts:517` | *Lower confidence.* **`syncOpenProjects` is the one command whose argument is computed, and its doc still doesn't say what an empty list does.** Before this PR `[]` and `undefined` differed only in scope; now `[]` is (claimed) a no-op while `undefined` on a zero-state account triggers the whole-account bootstrap. Its body is closed-source, so whether `[]` is normalized to `undefined` isn't establishable from this repo. |
| 27 | Nit | `sync-consent-step.component.tsx:21` | *Lower confidence, adjacent scope.* **The one UI that triggers the newly documented bootstrap describes it to the user as a refresh:** `assets/localization/en.json:61` reads "syncing updates your local copy and shares your changes with others" — bullet 2's language, not a bootstrap that clones projects with no local copy yet. |

---

## Withdrawn and refuted — do not re-litigate

**Withdrawn on review of this report:**

- *"as needed" on the DBL clause hides a wholesale skip.* Tom added a hedge; a hedge weakens a claim and cannot introduce a false one. Backwards criticism.

**Refuted on evidence during verification:**

- The `EditedStatus === 'unregistered'` partition gap — non-`new` really does imply locally present
- A full-clone risk via `syncOnProjectSwitch` / `syncOpenProjects` — both provably unreachable with a `new` project
- The missing `<para>` — verbatim pre-existing, matches all three sibling stubs
- `<c>new</c>` as a convention breach — `<c>` wraps markers, filenames and URLs throughout `c-sharp/`

## Verified clean

`prettier` / `format:check`, `csharpier --check`, `npm run lint`, XML well-formedness, `papi.d.ts` (regeneration not needed), and CLAUDE.md / `.claude/rules` conventions.

**Do not** ask for a reflow of `index.d.ts:481/484/486` for the 100-col budget — reflowing flat and regenerating reproduces them exactly; that is `prettier-plugin-jsdoc`'s canonical output.

**No build-time validation of this doc prose exists on either side.** The two typedoc configs with `treatWarningsAsErrors: true` are workspace-scoped (`entryPoints: ["src/index.ts"]`) and never reach `src/@types/`; `lib/papi-dts/typedoc.json` reads `papi.d.ts`, which doesn't bundle this declaration; and no `.csproj`/`.props`/`.targets` sets `GenerateDocumentationFile`, so CS1570/CS1572/CS1573 can never fire. This reinforces findings 3 and 25, and confirms the `build:types` regeneration trap that normally bites `src/shared` edits does not apply here.

---

## Bottom line

The PR does what it set out to do, and the two corrections at its centre are right. The realistic ask before merge is **one edit and one line**:

- **Finding 1** — rewrite bullet 3 at the outcome level so it stops claiming an early stop that up to five downloads precede. The suggested replacement also closes findings 6 and 7.
- **Finding 2** — revert "zero-state bootstrap case" to something accurate about the caller.

**Findings 3 and 4** are the durable ones: fixing the altitude is what prevents this recurring, and upstreaming is what stops the fix being silently reverted. Everything from finding 5 down is optional.

No changes were made to any repository during this review.

> @katherinejensen00 reviewed 2 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2771#-:-P0iNZ-9ANfQ6NPJKN0D:b-ejh9cz)* status: all files reviewed, 1 unresolved discussion (waiting on tombogle).

___
*[`c-sharp/Projects/SendReceive/ParatextProjectSendReceiveService.cs` line 153 at r1](https://reviewable.io/reviews/paranext/paranext-core/2771#-P0iNlvXDXb0IlhlQENr:-P0iNlvXDXb0IlhlQENs:b64q4tr) ([raw file](https://github.com/paranext/paranext-core/blob/ad4080f884173cde538ac83b91194e828c3eae39/c-sharp/Projects/SendReceive/ParatextProjectSendReceiveService.cs#L153)):*
> ```Smalltalk
>     /// already present locally (not new), every locally-present project is synced; new projects
>     /// are left alone.</item>
>     /// <item>If <see langword="null"/> and every shared project the account knows about is
> ```

#### 1. Important — `c-sharp/Projects/SendReceive/ParatextProjectSendReceiveService.cs:153`
*(mirrors: `.cs:155`, `.cs:157`; `src/@types/paratext-bible-send-receive/index.d.ts:486`, `:488`, `:489`)*

**Bullet 3 over-generalizes: it presents the early stop as what always happens, but up to five projects always download first.**

Three sub-cases, all fixed by one edit:

- **Small accounts.** `var firstBatch = projectIdsToSync.Take(5).ToList()` is shared before any role is inspected (patch:7155), and `hasNonObserverRoleInFirstBatch` is evaluated only afterward. For an account with **5 or fewer** shared projects the first batch *is* the whole account, so nothing stops early — and "rather than unconditionally syncing the whole account" is false. (The `if (isFirstTimeSync && projectIdsToSync.Count > 5)` gate at patch:7153 routes these to the `else` arm, which shares the full list; the outcome is the same either way, which is why the batch — not the gate — is the real story.)
- **The floor is five, not one.** "Stops as soon as one synced project gives … a non-Observer role" understates it by five clones — the number anyone tuning the startup readiness budget actually needs.
- **All-Observer accounts.** When `firstEditableId` stays null, patch:7254 falls back to `syncedIds` (everything), each probe its own `SendReceiveProjectsCore(new List<string> { id })` round-trip — serially, so worse than the batched share it is contrasted against. Observer-only logins are a normal Paratext role. *Mildest of the three:* the doc does say "then the rest one at a time," so the fallback is disclosed; only the "rather than" clause overreaches.

Suggested replacement, which also resolves findings 3 and 6 and stops encoding tunable constants:

```
A first sync may download only a SUBSET of the account: it stops early once it has a
project the current user can work in, and only the projects it kept are then scanned for
connected resources. Do not assume every shared project is present when this resolves.
```

katherinejensen00 (approved):
Thanks for making those changes!
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2771#-:-P10RGAE5JdeUcSytQ7c:bwr8ik9)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

tombogle:
DRAFTT: waiting on corresponding "upstream" changes are [here](https://github.com/paranext/paratext-bible-internal-extensions/pull/200).

tombogle:
@katherinejensen00 — following up with the full disposition now that the second round of fixes (`cece3bb9b9b`) is pushed.

## Fixed

**#1–#3** (bullet 3 over-generalizes; wrong caller attribution; encodes closed-source tuning constants) — addressed together in the earlier commit, per the earlier reply in this thread: the null/first-sync bullet now states only the caller-visible guarantee, with no Studio-specific mechanism left to drift. Also promoted to a standing rule + ADR entry (`adr-closed-source-command-doc-altitude`) so this doesn't just fix this one doc.

**#5** (`startup-readiness.util.ts`'s stale "typically has nothing to do" comment, plus the sibling stale sites in `startup-tasks.ts` / `startup-tasks.test.ts`) — all fixed. The highest-value one (`startup-readiness.util.ts`) now correctly says the zero-local-projects case is the *most* expensive case for the delayed sync, not the least — and, since you raised it, also now names both ways an account ends up with zero local scripture projects (true first sync, or every previously-local project having since gone missing), matching the same duality `syncProjects`'s own doc states, rather than calling it just a "true-first-sync condition."

**#6** (`each synced project's connected resources` imperfect in the narrowing branch) — re-evaluated since the mechanism it originally cited is gone from the doc, but the underlying imprecision is still real: the summary now says "connected resources... for the project(s) this call settles on" rather than "each synced project's," since not everything touched during resolution is guaranteed to be scanned.

**#7** (the two `null` bullets read as an exhaustive partition, vacuously satisfied by a zero-shared-project account) — turned out to already be resolved as a side effect: the bullet 3 rewrite hedges with "if the account has one," which covers the empty-account case without a separate fix.

**#8** (bullet 1 over-generalizes — a known id can still be dropped silently) and **#9** (bullet 2 dropped the "shared" qualifier) — both fixed, and both ported to `paratext-bible-internal-extensions` directly (PR #200 there) rather than filed separately, since it's the same team maintaining both.

**#10** (duplicated "Unknown IDs are skipped" in `index.d.ts`) — fixed; both docs now say it exactly once, matching.

## Not yet — blocked on an external PR

**#4** (upstream propagation) — see the comment above this one: this PR is in draft specifically for this reason, waiting on `paratext-bible-internal-extensions#200` to merge before the "Derived from" SHA and this decision's Consequences section get updated to close the loop. Everything else in this PR is otherwise ready.

## Still not addressed, unchanged from before

**#11, #12** — no action needed, per your own read (coordination note only; the patch's own comment, outside this diff).

**#13–27** — left as you scoped them: pre-existing, not merge-blocking, not part of this PR.

katherinejensen00 (commented):
Looks good. Just one thing to think about:
"Deduplication is handled internally" is still in both core doc blocks for syncProjects: ParatextProjectSendReceiveService.cs:221 and index.d.ts:538. On paratext-bible-internal-extensions#200 you confirmed it isn't true. A project that is both a primary and a connected project gets synced twice, and the second result overwrites the first. You've filed PT-4602 for that. The sentence isn't in #200 or in upstream main, so core is now the only place that still says it. Could you drop it from both core files in this PR?

> @katherinejensen00 reviewed 5 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2771#-:-P1IIikD8J5_l34wNV6A:bykysvv)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

tombogle:
> Looks good. Just one thing to think about: "Deduplication is handled internally" is still in both core doc blocks for syncProjects: ParatextProjectSendReceiveService.cs:221 and index.d.ts:538. On paratext-bible-internal-extensions#200 you confirmed it isn't true. A project that is both a primary and a connected project gets synced twice, and the second result overwrites the first. You've filed PT-4602 for that. The sentence isn't in #200 or in upstream main, so core is now the only place that still says it. Could you drop it from both core files in this PR?
> 
> > @katherinejensen00 reviewed 5 files and all commit messages, and made 1 comment.
> > _[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2771#-:-P1IIikD8J5_l34wNV6A:bykysvv)_ status: :shipit: complete! all files reviewed, all discussions resolved.

Yes, it will be dropped form the downstream as well. Changes made in this repo will be reflected in core as appropriate.

katherinejensen00 (approved):
Thanks for your hard work on this, Tom! A few things to consider, but nothing blocking so going ahead and approving the PR.

### W1 — The new rule and the ADR both quote a string that does not exist in the repo

**Files**: `.claude/rules/architecture/closed-source-command-docs.md:5-7`,
`.context/standards/Architecture-Decisions.md:658-660`
**Raised independently by**: contracts, architecture, clarity. **Verified directly.**

Both documents quote the doc phrasing that marks a command as in-scope as:

> `@throws PlatformUnimplementedException if not running in an application that implements this
> command (e.g., Paratext 10 Studio)`

No such string exists. All four `@throws` sites in `src/@types/paratext-bible-send-receive/index.d.ts`
(lines 566, 596, 611, 638) read `(e.g., Paratext 10)` — no "Studio". The C# XML docs name no
application at all.

```
$ git grep -n "Paratext 10 Studio" pr-2771-head -- src/@types c-sharp
(no matches)
$ git grep -n "e\.g\., Paratext 10" pr-2771-head -- src/@types/paratext-bible-send-receive/index.d.ts
566:     *   this command (e.g., Paratext 10)
596:     *   this command (e.g., Paratext 10)
611:     *   this command (e.g., Paratext 10)
638:     *   this command (e.g., Paratext 10)
```

This matters because the ADR uses that quoted string as the rule's **only mechanical selector** —
how a future reader or agent decides whether a given command is covered. Applied literally it
matches zero commands, including `syncProjects` itself. The rule's framing ("Studio is named as an
example implementer, not as the contract") also rests on Studio being named, which it isn't.

Fix either direction: restate the trigger structurally instead of quoting a literal, or change the
four `.d.ts` `@throws` to actually say "(e.g., Paratext 10 Studio)".

### W2 — The new rule file is the only one in its directory without `paths:` frontmatter

**File**: `.claude/rules/architecture/closed-source-command-docs.md:1`

All 8 pre-existing files in `.claude/rules/architecture/` are path-scoped. This one starts directly
with `# Doc Altitude for Closed-Source-Backed Commands`, so it loads into **every** session in the
repo. The unscoped rules that do exist elsewhere (`no-secrets.md`, `forward-facing-comments.md`,
`grep-safety-net.md`, `eslint-disable-discipline.md`) are genuinely repo-wide policies; this one
governs a single closed-source-backed command family.

```
.claude/rules/architecture/closed-source-command-docs.md     # Doc Altitude for Closed-Source...
.claude/rules/architecture/csharp-patterns.md                ---
.claude/rules/architecture/discover-before-implementing.md   ---
.claude/rules/architecture/extension-patterns.md             ---
.claude/rules/architecture/porting-discipline.md             ---
.claude/rules/architecture/react-patterns.md                 ---
.claude/rules/architecture/shared-patterns.md                ---
.claude/rules/architecture/storybook-links.md                ---
.claude/rules/architecture/winforms-porting.md               ---
```

Suggested frontmatter:

```yaml
---
paths:
  - "c-sharp/Projects/SendReceive/**"
  - "src/@types/**"
---
```

### W3 — New contract prose landed after approval, beyond the outstanding ask

**File**: `src/@types/paratext-bible-send-receive/index.d.ts:567`;
`c-sharp/Projects/SendReceive/ParatextProjectSendReceiveService.cs:222-224`

The PR was approved 2026-09-08. The round-3 ask (09-12) was narrowly "drop the deduplication
sentence" — already done by `530053180` on 09-11. `fe935b998` then:

- restructured the summary paragraph in both files;
- promoted "Callers MUST NOT assume every shared project is present locally" out of the
  zero-local-projects bullet into a general statement;
- added a **new behavioral fact**: `@throws` when another Send/Receive is already in progress —
  "rejects fail-fast rather than queuing behind it" — plus "an unregistered user or a lost server
  connection";
- re-pointed the upstream SHA.

The fail-fast `@throws` is a load-bearing contract claim a caller could write retry logic against,
and it is exactly the category the PR's own new rule says must be a caller-visible guarantee. It has
never been reviewed. **This is the main item for the re-review.**

### W4 — PR description is materially stale and now contradicts the landed code

**File**: PR body (generated 2026-09-04 by the `/review-paratext` summary template)

Three distinct problems:

1. "**Files changed**: 2", "either file", "only these two files changed" — the diff is 7 files,
   including a new standing rule (+54) and a new ADR entry (+56), neither mentioned anywhere in the
   description.
2. The Overview's third bullet still asserts the landed doc says the first-sync case "stops
   downloading as soon as one synced project gives the current user a non-Observer role, trying a
   small initial batch first, then the rest one at a time." That prose was **deleted** during review,
   and its removal is the subject of the new ADR, which records exactly that description as
   *rejected*. A reader of the description concludes the opposite of what HEAD contains.
3. Positive Observations still says "Confirmed via `git diff --name-only` that only these two files
   changed — no incidental logic, test, or config changes riding along." Now false
   (`src/main/startup-tasks.test.ts` is in the diff).

The "documentation-only, no executable code changed" claim is still accurate. Since this repo
squash-merges, the stale body becomes the permanent commit description.

> @katherinejensen00 reviewed 3 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2771#-:-P25gmv5BoxG4nbqVysu:b-x3a837)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

tombogle:
@katherinejensen00 — closing the remaining loose ends.

**"Deduplication is handled internally" — dropped from both core files** (`ParatextProjectSendReceiveService.cs`, `index.d.ts`), as discussed above. The real gap behind it is filed as [PT-4602](https://paratextstudio.atlassian.net/browse/PT-4602) — `seenConnectedIds` not seeded before the connected-resource scan, so a project that's both a primary and a connected reference in the same call gets synced twice, second result overwriting the first. Minimal priority; its DoD includes re-adding the claim once fixed, and a follow-up doc note either way (there's a second, independent reason "each synced project's" undersells it, beyond just that bug — noted there too).

**Finding #4 (upstream propagation) — done.** `paratext-bible-internal-extensions#200` merged 2026-09-18. Pulled its wording improvements back into both core copies (clearer connected-resource structure, the "callers MUST NOT assume" sentence promoted to a general statement instead of being scoped to just the zero-local-projects bullet, a new `@throws` fact about concurrent-sync rejection) while keeping core's more precise "project(s) this call settles on" phrasing rather than reverting to #200's "each synced project's" — see PT-4602 for why that distinction still matters post-fix. Updated the `Derived from` SHA to #200's merge commit, and closed out `adr-closed-source-command-doc-altitude`'s Consequences section to record all of this.

**This review's four non-blocking findings — addressed.** W1 (the doc-altitude rule quoted the exact `@throws` literal as if it were a stable selector — the string had already drifted once via a repo-wide "Paratext 10 Studio" → "Paratext 10" rename) and W2 (the rule file was missing path-scoping frontmatter) are both fixed in `.claude/rules/architecture/closed-source-command-docs.md`, with the same de-quoting mirrored into the ADR's Decision bullet. W4 (this PR description had gone stale against the growing file list) is fixed via a dated update note now sitting above the auto-generated `/review-paratext` summary, correcting the file count, flagging the Overview bullet that described superseded wording, and noting the "only two files changed" claim no longer holds. W3 needs no code/doc change.

Branch has also been rebased three times since review started (main moved substantially each time, including an unrelated "Paratext 10 Studio" → "Paratext 10" rename) — nothing substantive changed as a result, just kept current. `MERGEABLE` as of this comment.

Nothing else outstanding from my side — let me know if anything here needs another look.


[PT-4602]: https://paratextstudio.atlassian.net/browse/PT-4602?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

tombogle:
Correction to my comment above: I mischaracterized **W3** as "needs no code/doc change." That's wrong — W3's actual point is that the fail-fast `@throws` contract claim (concurrent Send/Receive rejection, plus the unregistered-user/lost-server-connection cases) landed in the post-approval wording sync and was never itself reviewed. That's a real, substantive claim, not a non-issue.

Rather than resolve it unilaterally here, I've added it as an explicit open item on [PT-4483](https://paratextstudio.atlassian.net/browse/PT-4483)'s Definition of Done, so it gets a deliberate review pass rather than being waved through. Apologies for the mischaracterization.


[PT-4483]: https://paratextstudio.atlassian.net/browse/PT-4483?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

katherinejensen00 (approved):
:lgtm:

> @katherinejensen00 reviewed 2 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2771#-:-P28l3Pd1O-SI9YLAk3b:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2772, opened 2026-09-04
PT-4501: Throttle late-unsubscriber warnings and fix the churn

jolierabideau:
# Code Review Summary

**Branch**: pt-4501-late-unsubscriber-warning

**Base**: origin/main

**Date**: 2026-09-04

**Files changed**: 9

## Overview

`UnsubscriberAsyncList` logged one `console` warning per late-arriving unsubscriber. Each of those
calls is routed over IPC to the main-process log, so during a re-subscription storm the logging
became a significant share of the load it was describing. This branch fixes both the symptom and its
cause: late-arrival reports are now throttled through a 10s window keyed by list name and outcome,
and `useOpenProjectBookIds` no longer subtracts `activeProjectId` from the set of projects it
subscribes to — which was what made the subscription fingerprint churn on every navigation.

The branch was rebased cleanly onto the latest `origin/main` at the start of this review (PR #2772
was previously showing as CONFLICTING and 8 commits behind). During the review the committed
`platform-bible-utils/dist` was rebuilt — without it the change did not ship at all — the throttle's
`threw` path gained error sampling and coverage, module-global throttle state gained a test reset,
and 690 lines of unrelated `platform-bible-react/dist` churn were dropped.

## API Changes

- `platform-bible-utils` (`UnsubscriberAsyncList`): no export, signature, or type changes. Behavior
  change only — late-arrival reporting in the private `unsubscribeImmediately` path is throttled to
  one verbatim `console.warn`/`console.error` per 10s per `(list name, outcome)`, plus one
  suppressed-count summary when the window closes. `add`'s TSDoc now documents the rate limiting
  (the only change visible in the generated `dist/index.d.ts`: +5 lines, −0).
- Two new **module-level** exports in `unsubscriber-async-list.ts` —
  `LATE_ARRIVAL_REPORT_WINDOW_MS` and `resetLateArrivalReportWindowsForTests()`. Neither is
  re-exported from `src/index.ts`, so neither reaches the package's public API surface; verified via
  the generated `dist/index.d.ts` diff.
- `lib/platform-bible-react`: no source changes, and its `dist/` is unchanged from `main`.
- `@papi/` modules / `lib/papi-dts/papi.d.ts`: no changes, none needed.
- `src/renderer/hooks/use-open-project-book-ids.hook.ts` is renderer-internal, not a public API
  surface. Its exported signature is unchanged; the meaning of `activeProjectId` narrowed from
  "excluded from subscriptions and result" to "excluded from result only".

- [x] **The throttling change did not ship.** `platform-bible-utils` is consumed through its
      committed `dist/` (its `.gitignore` tracks `dist` deliberately; `package.json`
      `main`/`module`/`exports` all point at `./dist/…`), but no `dist/` file had been rebuilt — so
      every process still ran the unthrottled code and CI could not see the skew. Flagged
      independently by all three code-analysis passes. _(fixed during review: ran `npm run build:pbu`
      then `npm run build:pbr`; verified the new `per-occurrence` text now appears in
      `dist/index.js` and `dist/index.cjs`, and that `.map` files regenerated alongside)_

[Author response: Author chose to rebuild and commit both `dist/` trees. Note the repo already
documents this rule — `.claude/rules/code-quality/native-string-vs-grapheme-helpers.md`, "Publishing"
section — including the build order.]

- [x] **`lib/platform-bible-react/dist/` carried 690 lines of unrelated churn** across four bundles
      with no `platform-bible-react/src` change on the branch, and the bundles contained none of this
      PR's change. _(fixed during review: `npm run build:pbr` proved the churn was pre-existing drift
      between `main`'s committed bundles and the current toolchain — the rebuild was a byte-identical
      no-op — so the four bundles were reverted to `main`'s version)_
- [x] **Committed sourcemaps did not describe the committed bundles.** The four
      `platform-bible-react/dist/*.map` files were identical to `main` while their bundles were not.
      _(fixed during review: resolved by the revert above; bundles and maps now both match `main`)_
- [x] **Module-global throttle state had no reset hook.** `lateArrivalReportWindows` lives for the
      process and is keyed by list name, so test isolation depended entirely on every test inventing
      a unique name. The pre-existing tests all use the name `'test'` and were already suppressing
      one another silently; the suite passed only because of ordering and because the suppressed
      cases did not assert on `console`. _(fixed during review: added
      `resetLateArrivalReportWindowsForTests()`, which also clears pending summary timers, called in
      `beforeEach` of both describe blocks and in `afterEach` of the fake-timer block. Pinned by a
      deliberately same-named test pair — verified it fails when the reset is neutered and passes
      when restored)_
- [x] **The `threw` outcome was the least covered and the only lossy one.** Only `unsubscribed` had
      full window-lifecycle coverage; `threw` had none, and its suppressed summary dropped `${error}`
      entirely, so a storm of differing errors collapsed to a count with no cause recoverable.
      _(fixed during review: windows now sample up to `MAX_SUPPRESSED_ERROR_SAMPLES` (3) distinct
      error texts and the summary reports them; three new tests cover the storm, the multi-cause
      case, and the cap — the cap test verified to fail when the cap is raised to 10)_

[Author response: Author approved all four. The error-sampling design (bounded distinct-sample set
rather than last-error or unbounded collection) was proposed during review, not authored
independently — worth a look from the reviewer.]

### Minor — Consider

- [x] **Map bound overstated** — "bounded by the number of distinct list names in the code" is wrong;
      several construction sites interpolate runtime values (`PDP Factory for ${projectInterfaces}`,
      `WebViewFactory for webViewType ${webViewType}`, `extension.name`). _(fixed during review:
      comment now says "produced at RUNTIME" and names the interpolating sites)_
- [ ] ~~**Default `'Anonymous'` name shares one budget across unrelated lists**, so one list's storm
      can suppress another's first-and-only warning.~~ _(Not changed in code, deliberately: requiring
      a name would break the public constructor signature, and separate budgets would buy nothing
      readable — two `'Anonymous'` lists both log `UnsubscriberAsyncList Anonymous: …`, so their
      reports are already indistinguishable. Recorded as a `@param name` note instead.)_
- [x] **Summary `setTimeout` not `unref`'d** — late arrivals cluster during teardown, so in the
      Node-side processes a pending 10s timer could hold the event loop open at shutdown. _(fixed
      during review: `unref()` called under an `'unref' in` narrowing guard, leaving the browser's
      numeric handle untouched)_
- [x] **Three near-identical report blocks** differing only in log level, outcome token, and
      phrasing. _(fixed during review: collapsed into a `LATE_ARRIVAL_REPORTS` table keyed by
      outcome; the call site went from ~45 lines to 6)_
- [x] **`windowSeconds` recomputed per late arrival** from a module constant. _(fixed during review:
      hoisted to `LATE_ARRIVAL_REPORT_WINDOW_SECONDS`)_
- [x] **`add`/class TSDoc silent on rate limiting**, though those reports are a consumer's only
      signal that a subscription was undone. _(fixed during review: `add`'s TSDoc now states the
      window and warns not to count log lines to count undone subscriptions)_
- [x] **Backward-facing comment narration** in `use-open-project-book-ids.hook.ts` ("that toggle
      **flipped**… **tore down**…") and its test ("subtracting the active project **made** the set
      flip") described the code this PR removed, against
      `.claude/rules/code-quality/forward-facing-comments.md`. _(fixed during review: rewritten as
      present-tense constraints; the narration already lives in the commit message)_
- [x] **The same read-time-filter rationale appeared three times** in one file, as three paraphrases
      that can drift apart. _(fixed during review: `getOpenProjectIds` is now canonical; the other two
      sites are short pointers)_
- [x] **Tests hardcoded `10_000`** while production used the unexported constant. _(fixed during
      review: `LATE_ARRIVAL_REPORT_WINDOW_MS` exported at module level and used in tests)_
- [x] **The `undefined`-gap behavior was unasserted** — the churn test used the same books for both
      projects, so the transient was invisible to it. _(fixed during review: new test uses different
      books per project and pins that the navigation target's own books are offered during the gap)_

[Author response: Author said "Fix all". One (the `'Anonymous'` budget) was resolved as documented
intent rather than a code change, with the reasoning recorded above.]

#### Shared Regions Modified

None. No `#region shared with` markers in any changed source file.

#### Extension Config Changes

## Positive Observations

- Both behavior changes carry tests that encode the *reason* for the design, not just the mechanics:
  `shares one budget across every list with the same name` pins the by-name (not by-instance) keying,
  and `changing the navigation target does not tear down and rebuild subscriptions` asserts the exact
  `undefined` toggle that caused the churn, including that `otherProject`'s unsubscriber was never
  called.
- Neither fix required changing a public export or signature, so no deprecation or `papi.d.ts`
  regeneration was needed — the compatible route was taken.
- The trailing summary fires on a timer rather than on the next occurrence, and the comment explains
  why the obvious cheaper alternative loses the most valuable number (a storm that stops abruptly
  would drop its final and largest count).
- Success/failure/throw outcomes get separate budgets, so a chatty success path cannot mask a genuine
  failure — and there is a test for exactly that.
- The renamed test `never offers the active project's own books` follows the behavior change instead
  of being deleted.
- Localization: no findings — all new strings are developer diagnostics, correctly unlocalized.
  Storybook: no findings — no new components. UX: no findings — no UI code changed.

## Interview Notes

**Stated purpose**: throttle the late-unsubscriber warnings and fix the subscription churn producing
them.

**How the author localized the churn** (their reasoning, recorded because it is the load-bearing
part of the diagnosis): the fingerprint gating subscription rebuilds is `openProjectIdsKey`, the
sorted NUL-joined ids from `getOpenProjectIds()`. In Simple mode every open panel tends to carry the
same `projectId`, so excluding the active project makes that set either `{}` (everything open is the
active project) or `{active project}` (during the `undefined` gap a dock rebuild produces) — two
different fingerprints. Every dock-rebuild-causing navigation toggles between them, firing the
`useEffect`, tearing down and rebuilding every subscription; the async unsubscriptions then return
after the new subscriptions are already in place, which is the late-arrival storm.

The author ruled out the competing hypothesis on timing: the warnings appeared at the cadence of
*reference changes*, not of tabs opening and closing. A web-view event producing a new-but-equal
array would have been continuous noise rather than reference-locked.

**The `undefined`-gap flicker**: the author had already identified this and deliberately left it as a
decision point rather than resolving it silently — noting that in the old code the active project had
no settled subscription behind it, so the gap rendered identically to the settled state, whereas now
its books flow through. They judged it sub-frame and unlikely to surface, and did not want to add
stateful ref logic without a test for the race.

Reviewed during the interview and **left unguarded on purpose**: a ref holding the last
non-`undefined` target would also suppress those books when `undefined` is the *settled* answer (no
editor open, so the project is genuinely just another open project) — a silent, persistent failure
traded for a visible, sub-frame one. The decision is now recorded in a comment at the filter site and
pinned by a test.

**No areas of deferred understanding.** The author explained the diagnosis and the tradeoff in their
own words without deferring to AI, and identified the flicker independently before it was raised.

**Design work done during review, not by the author**: the error-sampling mechanism for the `threw`
path (bounded distinct-sample set, cap of 3) and the `LATE_ARRIVAL_REPORTS` table refactor were both
proposed and implemented during this review.

## In-Review Quality Check

- `npm run build:pbu` — **failed once, then passed.** typedoc runs with `treatWarningsAsErrors` and
  its only entry point is `src/index.ts`; the review-added TSDoc on the public `add` used
  `{@link LATE_ARRIVAL_REPORT_WINDOW_MS}`, which is unresolvable from there. Fixed by making that one
  reference plain backticks rather than widening the public API — the other two `{@link}` uses sit on
  non-exported declarations and are warning-free.
- `npm run build:pbr` — **passed**, byte-identical output, no unrelated files touched.
- `npm run typecheck` — **passed** (all four projects).
- `npm run lint` — **passed**, no `lint-fix` needed.
- Prettier over branch-changed files only — **passed**, all files already formatted.
- `npm test` — **failed on the first full-repo run, passed on re-run.** 13 files / 9 tests failed,
  all in `lib/platform-bible-react`, all environment/contention symptoms (esbuild
  `write EPIPE`, a storybook-setup import `SyntaxError`, browser-mode timeouts). Re-running that
  workspace alone: 161 files / 1449 tests passed. No link to these changes found, but see below.
- Focused suites re-verified after all fixes: `unsubscriber-async-list` 22/22,
  `use-open-project-book-ids` 18/18.

## Suggested Review Focus

- [ ] **The error-sampling design was authored during review, not by the PR author.** Bounded set of
      3 distinct error texts, joined with `|` into the summary line. Is 3 the right cap, and is
      "distinct texts, first-3-wins" the right sampling policy versus e.g. most-recent?
- [ ] **The `undefined`-gap decision.** The navigation target's own books are deliberately offered
      during a dock rebuild's gap, and the "show more books" affordance may flicker on. The reasoning
      against the ref guard (silent-and-persistent vs visible-and-sub-frame failure) is in a comment
      at the filter site and pinned by a test — worth a second opinion, since neither the flicker nor
      its absence has been observed in the running app.
- [ ] **`platform-bible-react/dist` drift on `main`.** A clean `build:pbr` on this machine produces
      output differing from `main`'s committed bundles by ~690 lines. This PR reverted its copy to
      match `main`, so nothing here is blocked — but `main`'s committed react bundles appear stale
      relative to the current toolchain, and that is worth chasing separately.
- [ ] **`platform-bible-react` test flakiness under full-repo `npm test`.** Passed standalone, failed
      under parallel load. Unrelated to this change as far as could be determined; flagging in case
      CI shows the same.
- [ ] **`'Anonymous'` lists share one throttle budget** — resolved as documented intent rather than
      code. Confirm that is acceptable, given their log output is already indistinguishable.
- [ ] **PR body needs updating**: its third bullet describes a fix to
      `bcv-open-resource-books-simple.spec.ts` that is already on `main` (via #2726) and is not in
      this branch's diff. The commit message carries the same stale paragraph.

---

An `UnsubscriberAsyncList` that receives an unsubscriber after it has already run logs one warning per occurrence. Each of those `console` calls is routed over IPC to the main-process log, so during a re-subscription storm the logging became a significant share of the load it was describing.

Two changes, one cause:

- `UnsubscriberAsyncList` now reports late arrivals through a 10s window keyed by list name and outcome. The first occurrence in a window is reported verbatim; the rest are counted and reported once as a total when the window closes, on a timer so a storm that stops abruptly does not lose its final and largest count. The window is keyed by name rather than by instance because the storm's shape is a fresh list per attempt, which a per-instance budget would not throttle at all. Because `threw` is the only outcome that discards more than repetition, a closing window also reports up to three distinct error texts it swallowed, so a storm's causes stay legible.

- `useOpenProjectBookIds` no longer subtracts `activeProjectId` from the set of projects it subscribes to. Doing so made the subscription fingerprint change every time the resolved navigation target changed, including the brief `undefined` a dock rebuild produces. In Simple mode every open panel carries the same project id, so that toggle flipped the set between "one project" and "none" and tore down every subscription on each flip. The active project is now filtered out at read time instead.

**Reviewer callout — the `undefined` gap is deliberately unguarded.** Because the active project is now genuinely subscribed, during the brief `activeProjectId === undefined` a dock rebuild produces nothing matches the read-time filter, so the navigation target's own books are offered for that gap and a caller's "show more books" affordance may flicker on. This is tolerated rather than guarded: holding the last non-`undefined` target in a ref would also suppress those books when `undefined` is the *settled* answer (no editor open, so the project really is just another open project) — trading a visible, sub-frame fault for a silent, persistent one. The decision is recorded in a comment at the filter site and pinned by a test. Please scrutinize it.

`platform-bible-utils` is consumed through its committed `dist/`, so the rebuilt bundles are included here — without them the throttling would not ship at all.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2772)


katherinejensen00 (changes requested):
# PR #2772 — PT-4501: Throttle late-unsubscriber warnings and fix the churn

- **PR:** https://github.com/paranext/paranext-core/pull/2772
- **Author:** jolierabideau
- **Branch:** `pt-4501-late-unsubscriber-warning` → `main`
- **Merge base:** `37203cf764d27995ee36147064c16f12a67027c2`
- **Commit:** `457c8fa75c1` (single commit)
- **Reviewed:** 2026-09-04
- **CI:** green on ubuntu-22.04, windows-latest, macos-15; CodeQL + both Analyze jobs pass
- **Suites:** `unsubscriber-async-list` 22/22, `use-open-project-book-ids` 18/18

Review method: five passes — soundness/testing (with mutation testing), architecture,
comments/documentation, an antagonistic pass against those three, and a `/code-review max` sweep.
Findings below are the consolidated result after the antagonistic pass corrected or cut several
items. `dist/` bundles were verified to contain the change but are otherwise excluded from review.

---

## Summary

The diagnosis is right and both fixes are directionally correct — the churn analysis is genuinely
good work, and the tests that pin *why* the design is what it is are better than most. Three things
are worth changing before merge: the error-sampling added during review is defeated by a one-line
stringification choice and can crash the reporting path; the one test named for the hook's core
behavior asserts nothing; and the `undefined`-gap cost flagged for scrutiny provably cannot occur at
the only consumer, so the PR callout and the comment defending it can come down.

---

## Must fix

### 1. Raw `${error}` defeats the error-sampling feature — High
**File:** `lib/platform-bible-utils/src/lifetime-management/unsubscriber-async-list.ts`
**Attach at line:** 143 — `    openWindow.suppressedErrorSamples.add(\`${error}\`);`

`suppressedErrorSamples.add(\`${error}\`)` uses raw interpolation. A rejection carrying a plain
object — routine across the JSON-RPC boundary — stringifies to `[object Object]`, so N distinct
causes collapse into a single `Set` entry. That is precisely the outcome the comment at lines
100-102 says the samples exist to prevent. Every test uses `new Error(...)`, so this is untested.

`getErrorMessage` is in this same package (`src/util.ts:279`; `util.ts` has no imports, so no
dependency cycle) and JSON-stringifies non-Errors. Substituting it also removes the crash in
finding 2.

### 2. Reporting failures are caught and re-blamed on the unsubscriber — High
**File:** `lib/platform-bible-utils/src/lifetime-management/unsubscriber-async-list.ts`
**Attach at line:** 256 — `        reportLateArrival(name, unsubscriberSucceeded ? 'unsubscribed' : 'failed');`

`reportLateArrival` is called from inside the `try`, so a failure in the reporting path is caught by
the `catch` and re-reported as `'threw'`, blaming the unsubscriber for an error it never produced.

Concrete crash path: an unsubscriber throws a `Symbol` → `catch` → `${error}` raises
`TypeError: Cannot convert a Symbol value to a string` (verified in node) → escapes the unawaited
IIFE as an unhandled rejection. Pre-PR the `try` body was two bare `console` calls; it now holds map
mutation, `setTimeout` and string formatting, so the blast radius grew materially.

Fixing finding 1 closes the crash; moving the reporting call out of the `try` closes the
mislabelling.

### 3. `disposed` is not checked before handing the unsubscriber to a sealed list — High
**File:** `src/renderer/hooks/use-open-project-book-ids.hook.ts`
**Attach at line:** 155 — `        .then((unsubscribe) => unsubscribers.add(unsubscribe))`

This never consults the `disposed` flag the effect sets at line 166, so every teardown-while-in-flight
hands an unsubscriber to a list already sealed at the top of `runAllUnsubscribers`. This is the
*generator* of the late-arrival storm, not just a symptom, and is worth guarding regardless of the
throttle.

Caveat: a bare `unsubscribe()` is not behaviourally equivalent — you would need to handle the
rejection and the `false` return, and `return unsubscribe()` would mislabel errors into the `catch`
at line 156.

### 4. The test named for the hook's core behavior is non-falsifiable — Medium
**File:** `src/renderer/hooks/use-open-project-book-ids.hook.test.ts`
**Attach at line:** 182 — `    await waitFor(() => expect(result.current).toEqual([]));`

`[]` is also the hook's initial value (`bookIdsByProjectId` starts `{}` → `books.size === 0` →
`EMPTY_IDS`), and `waitFor` runs its callback synchronously on the first check, so this passes before
the mocked PDP's `.then` has run.

**Confirmed by mutation:** deleting `if (projectId === activeProjectId) return;` (hook line 190)
leaves this test green in ~5 ms; only the sibling `undefined`-gap test fails (1 failed | 17 passed).

The previous version's falsifiable half — `expect(getProjectDataProvider).not.toHaveBeenCalled()` —
was removed and not replaced. The same file already guards against this exact trap at lines 344-348.
Because the sibling test does pin the invariant, this is a hygiene fix rather than an uncovered
invariant.

---

## Should address

### 5. The `undefined`-gap cost cannot surface at the only consumer — Medium
**File:** `src/renderer/hooks/use-open-project-book-ids.hook.ts`
**Attach at line:** 184 — `      // While \`activeProjectId\` is \`undefined\` nothing matches, so through the brief gap a dock`

Three independent mechanisms each prevent the documented flicker:

1. `platform-bible-toolbar.tsx:229` — `isBookChapterControlDisabled = !resolvedWebView`, applied at
   `:628`. `activeProjectId` is `resolvedWebView?.definition.projectId`, so the control is disabled
   for exactly the gap.
2. `book-chapter-control.component.tsx:401-407` — an explicit `useEffect` closes an already-open
   popover when `disabled` flips.
3. `book-chapter-control.utils.ts:253` — `additionalBookIds.filter((id) => !projectBookIdSet.has(id))`.
   During the gap the "additional" books *are* the project's own, so `extraBookIds` is empty and the
   fast path returns `projectBookIds` unchanged.

Documenting hook behaviour for future callers is fair, but the 6-line defense, the PR reviewer
callout and the commit-message paragraph all argue a cost that cannot occur today. Suggest reducing
to one line noting that the current caller disables through the gap.

### 6. The change introduces new churn of the kind it removes — Medium
**File:** `src/renderer/hooks/use-open-project-book-ids.hook.ts`
**Attach at line:** 122 — `  useEffect(() => {`

Simple mode, editor on project A, no resources open. Before: opening resource R moved the effect key
`'' → 'R'` — one acquisition, nothing torn down. Now it moves `'A' → 'A

> @katherinejensen00 reviewed 9 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2772#-:-P0iZAZ9ABX6VOV7sGrH:b-a2ln39)* status: all files reviewed, 1 unresolved discussion (waiting on jolierabideau).

___
*[`lib/platform-bible-utils/src/lifetime-management/unsubscriber-async-list.ts` line 143 at r1](https://reviewable.io/reviews/paranext/paranext-core/2772#-P0iZJVN6kCUCieH_iA5:-P0iZJVN6kCUCieH_iA6:b-v5audz) ([raw file](https://github.com/paranext/paranext-core/blob/457c8fa75c10a2d2b96c7c7f1a1b04011d7a6230/lib/platform-bible-utils/src/lifetime-management/unsubscriber-async-list.ts#L143)):*
> ```XML
>   openWindow.suppressedCount += 1;
>   if (error !== undefined && openWindow.suppressedErrorSamples.size < MAX_SUPPRESSED_ERROR_SAMPLES)
>     openWindow.suppressedErrorSamples.add(`${error}`);
> ```

### 1. Raw `${error}` defeats the error-sampling feature — High
**File:** `lib/platform-bible-utils/src/lifetime-management/unsubscriber-async-list.ts`
**Attach at line:** 143 — `    openWindow.suppressedErrorSamples.add(\`${error}\`);`

`suppressedErrorSamples.add(\`${error}\`)` uses raw interpolation. A rejection carrying a plain
object — routine across the JSON-RPC boundary — stringifies to `[object Object]`, so N distinct
causes collapse into a single `Set` entry. That is precisely the outcome the comment at lines
100-102 says the samples exist to prevent. Every test uses `new Error(...)`, so this is untested.

`getErrorMessage` is in this same package (`src/util.ts:279`; `util.ts` has no imports, so no
dependency cycle) and JSON-stringifies non-Errors. Substituting it also removes the crash in
finding 2.

katherinejensen00 (approved):
> @katherinejensen00 reviewed 2 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2772)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2744, opened 2026-08-31
PT-4423: Harden the Text Collection re-point (follow-up to #2630)

captaincrazybro:
# PR Summary

**Branch**: `worktree-pt-4423-text-collection-project-switch` · **Base**: `main`

**Files changed**: 16 · **Tests**: 1428 passing (86 files)

## Overview

**The re-point itself is already on `main`.** #2630 merged `updateRelatedTextCollectionPanel`,
`TextCollectionPanelOptions` and `SCRIPTURE_TEXT_GRID_WEBVIEW_TYPE` on 2026-09-03, byte-identical to
the version published here on 08-31. What remains in this PR is the corrections that came out of
review, their tests, and the ADR entry — rebased onto that merged code.

What it contributes, all verified absent from `main`:

- **`finalizeProjectSwitch` re-points the Text Collection.** Without it PT-4238 still reproduces on
  a Power→Simple switch: the rebuilt layout merges the Text Collection in from the default-layout
  supplement, which carries no `projectId`, so it is the one Column 3 panel that arrives unbound.
- **`usePublishNavigableProjectIds` is scoped to an owning project.** A re-point reuses the web view
  id, so the persisted list outlived the project it was built for. Only a *known* foreign owner
  clears it; an unbound first render and a pre-existing owner-less list are both adopted.
- **The grid body's state is a tested decision.** `resolveGridBodyState` replaces inline render
  branching, with the localized-strings guard ahead of every branch that renders prose — otherwise a
  literal `%key%` reaches the screen, including in the shipped no-project startup state.
- **Waits are bounded.** `useHasTimedOut` keys on the project so a switch restarts the allowance, and
  covers both the sources read and the cached-resource list, neither of which reports an error when
  it never resolves.
- **Failures are diagnosable.** `reloadWebView`'s result is checked (it resolves `undefined` rather
  than throwing) and failures log at **error** naming the project left on screen, because a failed
  re-point no longer self-corrects once `projectId` is latched into the saved definition.
- **Accessibility.** The grid's always-mounted live region announces both the wait and its outcome;
  `LoadingView` and `EmptyState` cannot announce their own, since a region mounted with its text
  already inside it is not read out.
- **Ids are normalized** on both sides of the skip guard, as the rest of the module does.
- **`adr-column-3-panels-are-told-their-project`**, which did not come across with the code.

## API Changes

No public API surface changed. `lib/platform-bible-react/`, `lib/platform-bible-utils/`,
[...]

PR #2736, opened 2026-08-28
PT-4238: Seed unbound Column 3 panels from ActiveEditorProjectId

tombogle:
Fixes [PT-4238](https://paratextstudio.atlassian.net/browse/PT-4238). Builds on #2744, which added the Simple-mode re-point. This PR leaves that re-point as #2744 merged it.

## The problem

When the Text Collection grid and Find had no `projectId`, they took "the active project" from scroll group 0's **source** project, the 5th member of `useWebViewScrollGroupScrRef`'s tuple. That field only records which project's versification the current reference is in. It changes for unrelated reasons (Back/Forward, a click in a resource cell, a click in Comments or Checks). It also doesn't change at all when a project switch lands on the same verse, which is why PT-4238 reports that the Text Collection reloads on MAT 1:1 → MRK 1:1 but not on a project switch.

In Power mode it also caused a second problem. Both re-point paths are Simple-only, so the grid changed its `projectId` in place on every navigation. `useBufferedLayoutSetting` only re-arms on `onSharedLayoutApply`, which only fires in Simple mode. So the admin-shared list would stay on the first project while the per-user list and overlay moved on. Tests against the real hook show this. No one has seen it in the app yet, and the Text Collection isn't used in Power mode today.

## What's deleted

- **Reading the scroll group's source project to decide which project to show.** The grid and Find were its only two consumers.
- **`candidateIsOwnResource`** in `resolveTextCollectionProjectId`. With the latch below, a resource cell taking focus can no longer pull the grid onto itself.
- **This branch's first design, `claimScrollGroupSourceProject`,** which re-stamped the scroll group on every switch. The earlier PR body described it. It never merged, and it has been dropped from the history. The branch is now rebased on `main` and has no merge commits.

## The latch rule

A new read-only window data type, `ActiveEditorProjectId` (`@experimental`), reports the project of the web view that BCV navigation drives. Each window's shard publishes it, and the main-process router relays it the same way it relays `Focus`.

A Column 3 panel with no project takes the **first** value it's given and keeps it. The rule is `explicit ?? alreadyShown ?? activeEditor`, in `resolveTextCollectionProjectId` and the new `useTextCollectionProjectId`. After that, only an explicit `projectId` moves the panel. In Simple mode that is the switch's reload. Find follows the same seed-only rule.

## The Power-mode trade-off

[...]

PR #2773, opened 2026-09-05
test: Pin the buffered layout setting contract (PT-4238 groundwork)

katherinejensen00:
**Regression guard for PT-4238 — no behavior change.** The fix landed in #2736 (and #2744): the Scripture Text Grid is now seeded once from `ActiveEditorProjectId` and moves only by a reload, so no consumer of this hook changes project in place any more. These tests pin the contract that fix relies on, and nothing in #2736 fails if it regresses. #2776 (PT-4316) edits both files here and would invert the in-place tests, so if it goes ahead it needs a rebase onto this.

_Rebased onto `main` after #2736 merged; the hook doc's `TODO(PT-4316)` paragraph, which described the grid's old in-place path, is replaced by the path it takes now._

# Code Review Summary

**Branch**: worktree-pt-4238-buffered-setting-contract-tests

**Base**: origin/main

**Date**: 2026-09-04

**Files changed**: 2

## Overview

**No behaviour changes.** One file is tests; the other is a doc comment. Nothing executable in the
hook was touched, so this can be reviewed on reading alone.

`useBufferedLayoutSetting` returns an in-memory _held_ copy of a project setting and **disarms after
its first successful read** — so it is only safe for consumers that remount on a project switch. That contract is why a consumer has to be re-pointed by remount rather than updated in place — unless the hook itself learns in-place switches, as #2776 (PT-4316) proposes. #2736 depends on this contract, and this PR pins it with tests so the fix has something to
hold it in place.

| File | What changed |
| --- | --- |
| `…/use-buffered-layout-setting.hook.test.ts` | 12 → 17 tests: six added, one removed (it mounted unbound in a settled state the real data hook never gives an unbound consumer). Also: render, event-firing, and fixture helpers extracted — `rerender` keeps the project and `switchProject` changes it; the tripwire and in-place cases are grouped under `TODO(PT-4316)` blocks; fixtures use `CURRENT_DATA_VERSION`; `clearAllMocks` → `resetAllMocks`. |
| `…/use-buffered-layout-setting.hook.ts` | **Comments only** — executable code identical to `main`. States which project changes are safe (a remount, or binding from `undefined`) and why an in-place move is not; notes that the Scripture Text Grid relies on the binding case, and what a consumer that did move in place would need; warns that `isLoading` and `settingError` describe the read, not the held copy. |

The six new tests, and what each pins:

| Test | Pins |
| --- | --- |
| `holds the outgoing project value until the incoming project re-arms` | After an in-place switch the held value stays on the outgoing project; in the same body a re-arm then delivers the incoming value, so the hold cannot pass vacuously |
| `latches the outgoing project value when the switch lands while still armed` | A switch before the first apply is not safe either: the outgoing project's late read is the one latched |
| `reports no error for the incoming project` | The error channel goes silent too, so a consumer can't tell a bad read from a good one |
| `stays silent when an unbound consumer binds, then warns when it moves on` | Binding is supported; a later in-place move still trips the warning |
| `stays armed while unbound so a projectId arriving later still applies` | A panel opened with no project still gets the right value when one arrives |
| `reports loading again on a later read while still serving the held value` | The loading channel tracks the provider, not the buffer |

The in-place cases drive the switch the way the real stack delivers it: the outgoing project's settled value is still served on the switching commit, because `usePromise` keeps the old provider and `isLoading` is raised from an effect. That matters for whoever inverts them — a reset keyed on `projectId` alone passes a one-render switch but fails this one (probed).

### One thing worth your opinion

**The error channel is documented but not fixed.** Reasoning under Interview Notes; the
alternative was resetting `hasAppliedRealValue` on a project change, which was rejected as
half-supporting a usage the hook documents as unsupported.

Everything below is detail — skim as needed.

## API Changes

None. The exported signature
`useBufferedLayoutSetting<ProjectSettingName>(projectId, key, defaultValue) => [held, isLoading, settingError]`
is byte-identical to the merge base. Nothing under `lib/`, `papi.d.ts`, or any `types/*.d.ts`
changed. The hook is internal to the `platform-scripture-editor` extension.

## Findings

Four analysis passes ran (API/correctness, style/architecture, coverage/compliance, UX). **0 Critical,
4 Important, 7 Minor — all fixed during review.** The UX pass self-skipped: no UI, no strings, no
keyboard handlers, no new components.

This branch also went through a `/code-review max` pass beforehand; three findings from it (a
vacuous assertion, a `mockReturnValue` leak, a missing capture guard) were applied before this review
and appear under Positive Observations rather than here.

None.

- [x] **Hook doc named consumers that don't use this hook** — it cited `checklist.web-view.tsx` /
      `checks-side-panel.web-view.tsx`, verified neither imports it, while the real in-place consumer
      went unnamed, making the hazard read as hypothetical. _(fixed: added `TODO(PT-4238)` naming the
      Scripture Text Grid and how it reaches the hook. The original example was kept — it accurately
      illustrates the practice)_
- [x] **A rationale comment contradicted both live consumers** — claimed panels "continue to render
      what they have" while loading; both actually replace their content. _(fixed: reworded to state
      the hook's guarantee instead of a consumer behaviour that doesn't exist)_
- [x] **The new error-channel doc clause had no test** — flagged independently by two passes.
      _(fixed: added a test; mutation-verified)_
- [x] **Two pre-existing negative tests could pass vacuously** — they fire `capturedHandler?.(…)`
      unguarded, so a broken capture makes the event a no-op and their assertions are trivially
      satisfied. This is the shape `.claude/rules/testing/tdd-discipline.md` warns about. _(fixed:
      added `expect(capturedHandler).toBeDefined()` to both)_

### Minor — Consider

- [x] **Tripwire false positive was unpinned** — an in-place change landing while still armed warns
      but applies correctly. _(fixed: added a test recording it)_
- [x] **Positional cross-reference off by one** after a test was inserted. _(fixed: names tests
      instead of positions, so it survives reordering)_
- [x] **Seed test's comment stated a mock-supplied premise as fact.** _(fixed: comment now says the
      premise comes from the mock, and that this is the only test reaching the tripwire's
      `!== undefined` branch)_
- [x] **Positive-control comment overstated what it proved.** _(fixed: narrowed, and points at the
      companion test that covers the rest)_
- [x] **`listOf` didn't read as "a list holding one project reference".** _(fixed: renamed
      `oneProjectList`)_
- [x] **Fixtures hardcoded `dataVersion: '1.0.0'` while `CURRENT_DATA_VERSION` is `'1.1.0'`.**
      Verified nothing branches on it and it's never asserted here, so the drift was inert. _(fixed
      for this file; deliberately not swept across five others — see Interview Notes)_
- [x] **`clearAllMocks` → `resetAllMocks` had no recorded reason.** _(fixed: comment records the
      leak it closes and why the handler capture survives)_

#### Shared Regions Modified

None — neither file contains a `#region shared with` marker.

#### Extension Config Changes

None — only `extensions/src/platform-scripture-editor/src/*.ts` source and test files.

## Positive Observations

- **Every behavioural test was confirmed falsifiable by mutating the hook**, not just by reverting:
  re-arming on an in-place change reddens the held-value test; pinning `isLoading` to `false` reddens
  the loading test; dropping the tripwire's `!== undefined` guard reddens the seed test; removing the
  `!hasAppliedRealValue` gate reddens the error-channel test.
- The in-place pair is a **true differential** — identical setup, differing only by the re-arm.
  Mutation reddens one and leaves the other green.
- `resetAllMocks` closes a real cross-test leak (`clearAllMocks` doesn't clear `mockReturnValue`).
  Verified safe on the pinned Vitest 3.2.7, where `mockReset` restores `vi.fn(impl)` implementations,
  so the `useEvent` capture survives.
- The seed test's premise was traced to production code rather than assumed: with no `projectId`
  there is no data provider, so nothing delivers and `isLoading` stays at its initial `true`.
- Mocking stays at external boundaries only; no internal collaborator is mocked.
- All comments are forward-facing — the one ticket reference is a `TODO` for open work.

## Interview Notes

**Purpose**: pin `useBufferedLayoutSetting`'s contract, which PT-4238's fix (#2736) depends on.

**Context**: three PRs (#2630, #2736, #2744 — all now merged) converged on one defect — Text Collection and
Find inferring "which project is active" from the scroll group's *source* project, a field that
exists for versification conversion. This hook's disarm-after-first-read behaviour is why the fix
must remount rather than update in place.

**Two deliberate calls:**

1. **Error channel documented, not fixed.** After an in-place switch `hasAppliedRealValue` is already
   `true`, so an unreadable setting on the incoming project isn't reported while the outgoing
   project's value is still served. Resetting that flag was rejected: it would make the hook
   *partially* support a usage it documents as unsupported, while the held value stayed stale
   anyway — removing the symptom without the problem. The contract was completed in the doc and
   pinned by a test instead. The real fix was the consumer ceasing to switch in place, which #2736 did.
2. **`dataVersion` drift not swept across the other five test files.** A mechanical replace would be
   unsafe — `select-dbl-resource.ts` carries `dataVersion` forward from stored data, so its test may
   assert it on output, and where a value *is* asserted, importing the production constant is worse
   than a literal: bumping `CURRENT_DATA_VERSION` would make the test silently follow instead of
   fail. The paired suggestion to hoist a shared fixture helper also doesn't survive inspection —
   the sibling files use three different item shapes and varying list lengths, and one already has
   its own local helper. A helper fitting all of them saves one field at the cost of an import.

**No unresolved findings.** No area was deferred to AI without explanation.

**Known limitation, raised unprompted**: the seed test's argument is a two-link chain and these tests
cover only the second link. That an unbound consumer keeps `isLoading` true is a property of
`create-use-data-hook.util.ts`, which these tests mock away; verified by reading production code, not
pinned by a test named for it.

## In-Review Quality Check

All clean; no auto-fixes needed, no unfixable failures.

| Check | Result |
| --- | --- |
| `vitest` (platform-scripture-editor) | **1379 passed / 85 files** |
| `prettier --check` | Clean |
| `eslint` | Clean (exit 0) |
| `tsc --noEmit` (extensions) | **314 errors — identical to baseline**, none in changed files |

The 314 are pre-existing `platform-editor` missing-export failures on this machine, unrelated to the
branch. Note: ESLint must be run from the `extensions/` workspace root — from the repo root it
reports "File ignored by default" and silently checks nothing.

## Suggested Review Focus

- [ ] **Document-don't-fix on the error channel** — is that the right call?
- [ ] **The tripwire's false positive** — worth knowing when diagnosing from logs: a warning alone
      doesn't mean staleness.
- [ ] **Follow-up ticket candidate** — normalize `dataVersion` in the five sibling fixtures, checking
      per file whether the version is asserted before importing the constant.
- [ ] **Coverage boundary** — there is no test harness for the Scripture Text Grid web view, so the
      consumer-level behaviour these tests imply is verified only by reading.

AI-assisted — [session](https://claude.ai/code/session_01JYo48ZAuqm7QBdq19p23o5)

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2773)


jolierabideau (changes requested):
## 🤖 [Claude AI Review] — PR #2773

Reviewed with a four-perspective panel (scope, architecture, tests, clarity). **1 blocking, 10 warnings, 14 suggestions.** CI is green on all 6 checks.

Thanks for the unusually candid PR body — the "No behaviour changes", the two contested calls flagged for the reviewer, and the note that the TODO's claim is still under discussion all made this reviewable on reading alone. Most of what follows is about *framing and provenance*, not about the tests, which are good.

|---|---|---|---|
| Scope | 1 | 2 | 2 |
| Architecture | 0 | 3 | 1 |
| Tests | 0 | 3 | 4 |
| Clarity | 0 | 2 | 7 |
| **Total** | **1** | **10** | **14** |

Counts are pre-dedup; three passes independently flagged the `TODO(PT-4238)` ticket ID (W-3).

---

### 🚫 Blocking — B-1: the PR carries the bug key but changes no behavior

All 8 added lines in `use-buffered-layout-setting.hook.ts` are TSDoc (` * `) lines; the body (58-117) is byte-identical to the merge base, and the only other changed file is a test. PT-4238's repro will behave after this merge exactly as it does on `main`.

Groundwork-before-the-fix is a legitimate PR — but this one is *titled* with the bug key while PT-4238 sits in "👀 Review", so merging will likely transition the bug as fixed while the defect is untouched.

**Decision needed:** re-key this PR (a test/docs task, or the PT-4423 hardening key) and link it to PT-4238 as groundwork?

---

### ⚠️ W-2 — the TODO attributes the filed repro to the wrong mechanism

The stranding claim itself checks out, verified at `use-text-collection-sources.hook.ts:32-55`: `adminReferenced` is buffered (holds the outgoing project) while `userReferenced`/`overlay`/`order` reset on the PDP identity change (follow the incoming one).

But the filed repro appears **already fixed on `main`**. A Simple-mode project switch re-points the grid by *remount*:

- `openOrUpdateRelatedPanels` → `updateRelatedTextCollectionPanel` (`platform-scripture-editor.utils.ts:1244`)
- → `reloadWebView` with an explicit `projectId` (`:1292-1311`)
- → `resolveTextCollectionProjectId` returns `explicitProjectId` first (`scripture-text-grid-project.utils.ts:34`)

That is precisely the remount path this hook's contract supports. It landed in `30f07f460b1` (PT-4059, #2630), and no PT-4238/PT-4423 commit is an ancestor of `origin/main`.

So the in-place stranding is **real but residual** — reachable when the grid is still unbound and following `activeEditorProjectId`, or when the `isProjectEditable` guard at `:1244` skips the re-point — *not* the path the zzz2/zzz6 repro takes between two editable S/R projects. As written, the TODO tells whoever picks up PT-4238 that it's an open buffered-setting defect. Worth rewriting to name the residual path and point at the open re-point-hardening work.

### ⚠️ W-3 — `TODO(PT-4238)` cites this PR's own ticket

`forward-facing-comments.md` allows a `PT-XXXX` only when it points *forward* to open work. This is the PR's own key; once it closes, the TODO points at a closed bug and a maintainer has no route to the deferred fix — the exact rots-on-merge form the rule exists to prevent. The surrounding prose is forward-facing and accurate; only the ID is wrong.

### ⚠️ W-4 — no `Architecture-Decisions.md` entry

> "The fix is to stop inferring that project rather than to add a reset here."

That declines the sibling PDP-identity-reset pattern (`use-structure-protection-state.hook.ts`, cited two paragraphs above), reassigns the fix to a different module, and defers a live defect — exactly the shape CLAUDE.md's "Recording Architecture Decisions" requires be logged. This hook already has ADR precedent at `.context/standards/Architecture-Decisions.md:1526-1570`, which even names `useTextCollectionSources`. The log is untouched here, so the rationale lives only in a TSDoc comment on the hook that is *not* being fixed.

### ⚠️ W-5 — the prescribed fix conflicts with documented intent

`resolveTextCollectionProjectId`'s TSDoc (`scripture-text-grid-project.utils.ts:15-28`) justifies the inference as required behavior — *"Opened from the default layout the grid has no explicit `projectId`, so it follows the active Scripture editor"* — and the grid deliberately latches `effectiveProjectId` from `activeEditorProjectId`, with tests pinning that latching. "Stop inferring that project" would remove a feature another module argues is necessary, and no ADR reconciles the two. **The diagnosis is sound; the prescription is unsupported.** Consider stating the candidate fixes neutrally, or citing where the choice was argued.

### ⚠️ W-6 — the re-arm event's identity is unpinned

`use-buffered-layout-setting.hook.test.ts:129` asserts only that `useEvent` was called *at all*, and `capturedHandler` is captured regardless of the event token. So renaming `platformScriptureEditor.onSharedLayoutApply`, subscribing to the wrong event, or dropping the `getNetworkEvent` call would leave the entire suite green while the buffer never re-arms in production — the most consequential piece of the hook's cross-boundary contract, in a PR whose stated purpose is pinning that contract.

<sub>(For the record: an earlier analysis claimed `resetAllMocks` wipes the `getNetworkEvent` return. That's incorrect on the pinned Vitest 3.2.7 — `mockReset` restores implementations passed to `vi.fn(impl)`, as your comment states, and green CI proves it. The gap above stands independently.)</sub>

### ⚠️ W-7 — known-defect tests carry no tripwire marker

The tests at `:263` and `:290` pin acknowledged-defective behavior as the contract. When the fix lands they must be inverted, and they'll read as regressions being defended rather than intended change — with nothing in the test file pointing at the ticket that obsoleted them. The Testing-Guide's `test.fails` tripwire convention wants the `TODO(PT-XXXX)` on the case itself, not only in the production file.

### ⚠️ W-8 — the central in-place assertion isn't self-sufficient

At `:284`, `toEqual(first)` can't distinguish "held the outgoing value" from "the `setRaw(second, false)` feed never reached the hook". The comment concedes this and resolves it by pointing at a *different* test, so falsifiability depends on a sibling that a future edit could change or delete. Closing the loop in one body would fix it: after asserting `first` is still held, fire the re-arm for `proj-2` and assert `second` lands.

### ⚠️ W-9 — comment volume and voice (PR-level, raised once)

Roughly 40 of the 174 added test lines are comment prose, much of it arguing that an assertion isn't vacuous ("Positive control…", "would also pass if…", "Guarded because…", "read them together"). Strip the PR context and much of it reads as an argument with a reviewer rather than information a maintainer needs — and a comment defending falsifiability is usually a signal to strengthen the assertion instead. The cross-references also couple these tests to each other's continued existence.

---

### 💡 Suggestions (non-blocking)

- **TSDoc has outgrown the API** — ~50 lines of prose above a 57-line function; a caller reaches `@returns` (line 54) only after six paragraphs, two about an unsupported usage. The paragraph added here details `hasAppliedRealValue`, an internal, in the caller-facing block.
- **`pid` → `projectId`** across the 5 new render blocks — `pid` conventionally means process id (Code-Style-Guide on indecipherable abbreviations).
- **Extract `renderBuffered(projectId?)`** — the 4-line `renderHook(...)` block now repeats 5× (239, 267, 294, 314, 331) plus a variant at 361; would also give the `initialProps` annotation at 360 one home.
- **Extract `fireReArm(projectId)`** — the `expect(capturedHandler).toBeDefined()` + `act(…)` pair repeats 3× (147, 161, 343), each with its own comment for the same guard. Parallels the existing `setRaw`.
- **Destructure `result.current`** — index assertions force the reader to hold the tuple order in mind, and two new tests mix indices in one block.
- **`DEFAULT`** is vague and stylistically inconsistent with the `oneProjectList` factory beneath it; `emptyList` would pair.
- **Don't cross-reference tests by title string** (line 63) — this PR renames a test on the very next line.
- **`returns the settled value it mounted with`** still passes with the apply effect deleted; what it pins is the `useState` seed — say that in the name.
- **`reports the provider loading state independently of the buffer`** — the first two phases assert the mock passes through; phase three (103-106) is the real contract. Name it after that.
- **Duplicate claim** that one test is "the only test that reaches that branch" (353 and 371) — keep the one on the assertion; nothing enforces it, so it will quietly go false.
- **Follow-up candidate** — normalize `dataVersion` in the 5 sibling files still on `'1.0.0'`, per file. Your reasoning for not sweeping it here holds up.

### ✅ Positive observations

- Zero runtime change, stated up front — genuinely reviewable on reading alone.
- The in-place pair is a true differential: identical setup, differing only by the re-arm.
- Two pre-existing vacuous negative tests were found and guarded.
- `resetAllMocks` closes a real cross-test `mockReturnValue` leak, with the Vitest-version reasoning recorded.
- Mocking stays at external boundaries only; no internal collaborator mocked.
- Fixtures sourcing `CURRENT_DATA_VERSION` follows an established pattern — three sibling test files in the same directory already import it.

### Suggested discussion agenda

- [ ] Re-key the PR so PT-4238 isn't closed by a no-behavior-change merge (B-1)
- [ ] Is the filed repro already fixed by #2630's remount re-point? If so, what's PT-4238's remaining scope — and does the TODO describe it? (W-2)
- [ ] Where does the "no reset here" decision get recorded? (W-4)
- [ ] Reconcile "stop inferring that project" with `resolveTextCollectionProjectId`'s documented purpose (W-5)
- [ ] Pin the re-arm event name (W-6)
- [ ] Mark the two known-defect tests as tripwires (W-7)

---

<sub>Requesting changes on **B-1** (PR keying) primarily — the code itself is in good shape. AI-assisted review; a human still owns the call on every item above.</sub>

katherinejensen00 (commented):
## Response to review

There are two commits since the review:

- `863a85e3434` addresses the items below.
- `493b4b9df4b` corrects claims in that first round, which a follow-up `/code-review max` pass found wrong (details at the end).

The hook's executable code is still identical to `main`. All 17 tests pass, and every new or changed assertion was mutation-checked: breaking the hook the way each one guards against makes exactly that test fail.

### B-1: the PR carries the bug key but changes no behavior

Retitled to "test: Pin the buffered layout setting contract (PT-4238 groundwork)". The body now opens by saying this PR doesn't fix PT-4238.

One thing a retitle can't change: the branch name still contains `pt-4238`, and an open PR's branch can't be renamed. If the Jira automation matches branch names, it may still link.

### W-2: the TODO attributes the filed repro to the wrong mechanism

Confirmed. The remount re-point on `main` covers switching between two editable projects. The TODO now describes only the remaining path: an unbound grid (opened by the default layout and not yet re-pointed by a reload) that `resolveTextCollectionProjectId` moves in place.

It also names a consequence the follow-up review traced. On that path, the grid's cell-order reconcile can prune the incoming project's saved order against the outgoing project's admin list, then save the result.

### W-3: `TODO(PT-4238)` cites this PR's own ticket

The TODO now cites PT-4316, because #2776 is the open work that would change this hook's in-place behavior. #2736 changes which signal the grid follows, but the grid still switches in place.

### W-4: no `Architecture-Decisions.md` entry

No entry in this PR. With W-5 addressed, the TSDoc no longer chooses a fix, so this PR makes no architecture decision; it documents the contract as it stands. The decision belongs to whichever PR lands: #2744 (which already adds `adr-column-3-panels-are-told-their-project`) or #2776.

### W-5: the prescribed fix conflicts with documented intent

Confirmed: `resolveTextCollectionProjectId`'s TSDoc does justify following the active editor. The TODO now names both candidate fixes without choosing one:

- the grid stops moving between projects in place, or
- the hook resets and re-arms when its project data provider changes.

It also records that a reset keyed on `projectId` alone would still latch the outgoing provider's value.

### W-6: the re-arm event's identity is unpinned

Fixed. The mocked `getNetworkEvent` now returns a token that names the requested event. The test asserts that `useEvent` was called with the token for `platformScriptureEditor.onSharedLayoutApply`, so renaming the event in the hook fails it.

### W-7: known-defect tests carry no tripwire marker

Markers added, but not `test.fails`. The Testing Guide reserves `test.fails` for a test whose body throws on a known, deferred bug. These tests pass, and they pin the hook's current behavior.

Checking the open PRs showed that #2776 would change that behavior, so the tests are now grouped under two `TODO(PT-4316)` blocks:

- The in-place cases: invert them rather than delete them.
- The tripwire cases: delete them if the warning goes.

The in-place cases also switch the way the real stack does: the outgoing project's settled value is still served on the commit where `projectId` changes. That's what makes them useful once inverted. A probe confirmed that a reset keyed on `projectId` alone passes a one-render switch but fails the realistic one.

### W-8: the central in-place assertion isn't self-sufficient

Merged the pair into one test. It asserts the outgoing value is held, then fires the re-arm for the incoming project and asserts the incoming value lands. It no longer depends on a sibling test.

### W-9: comment volume and voice

Trimmed. The comments arguing that an assertion isn't vacuous are gone. Where the worry was real, an assertion or helper now does the job: `fireSharedLayoutApply` checks that a handler was captured, and the W-8 merge made the positive-control comment unnecessary. Tests no longer refer to each other.

### Suggestions

- **TSDoc length:** condensed, and the internal `hasAppliedRealValue` detail is out of the caller-facing block. The in-place section later grew back a little, because the follow-up review showed the shorter version understated the hazard.
- **`pid` → `projectId`:** done.
- **`renderBuffered` helper:** done. Its `rerender` keeps the current project and `switchProject` changes it, so a same-project rerender can't be mistyped into a switch.
- **`fireReArm` helper:** done, named `fireSharedLayoutApply`.
- **Destructure `result.current`:** done wherever a block reads more than one element.
- **`DEFAULT` → `emptyList`:** done.
- **Cross-referencing tests by title:** removed.
- **Seed test name:** renamed to `applies a value that is already settled at mount`. The follow-up review showed it doesn't isolate the `useState` seed, so the name describes what it observes.
- **Loading test name:** renamed to `reports loading again on a later read while still serving the held value`.
- **Duplicate "only test that reaches that branch" claim:** both copies removed, since nothing enforces it.
- **`dataVersion` in the sibling files:** agreed, that's a follow-up rather than part of this PR.

### Follow-up `/code-review max` pass

It found that parts of the first round were wrong. `493b4b9df4b` corrects them:

- **The "harmless" claim.** The doc called a tripwire warning on a switch before the first apply "harmless", and `warns but still applies the incoming value when the change lands while armed` backed that up. But that test modelled a one-render switch the real stack can't produce. With realistic timing, the hook latches the outgoing project's value. The test is replaced by `latches the outgoing project value when the switch lands while still armed`, and the doc now says an in-place move is unsafe even before the first apply.
- **The tripwire's scope.** The doc said the tripwire fires on "any" in-place change, but it stays silent when an unbound consumer binds. The doc now says so, and a new test checks that binding stays silent and a later move warns.
- **An unrealistic test.** `handles an undefined projectId without applying or throwing` mounted an unbound hook that had already finished loading. The real data hook never produces that state, so the test is removed.
- **A contradicting comment.** The comment above the tripwire code no longer contradicts the TSDoc.

It also found pre-existing bugs that this PR doesn't fix, since it changes no behavior:

1. **Saved cell order can be lost on mount.** On the commit where the setting first becomes readable, the hook returns the placeholder with `isLoading` already false. `useTextCollectionSources` then builds the grid's sources with an empty admin list, and the grid's reconcile effect can prune and save the user's cell order in that same commit. The `@returns` doc now warns about the gap. I verified the code path but didn't reproduce it end to end. This needs a ticket.
2. **An unbound grid moving in place can overwrite the incoming project's saved cell order.** This is now named in `TODO(PT-4316)`.
3. **Panels may latch the pre-sync layout on a project switch.** In `main.ts`, `applyForProject` fires before the project-switch sync lands, and no apply prompt follows. The review reported this; it hasn't been verified yet.
4. **The tripwire logs a warning on a shipped path:** a read-only project opened first in Simple mode. Low severity, and #2776 removes the warning.

#2776 also edits both files in this PR, so whichever merges second will need a rebase.

> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2773#-:-P1IHfaLArpc-zIUUGr7:b-dn834l)* status: 0 of 2 files reviewed, all discussions resolved.

jolierabideau (approved):
## Approving — all review items addressed

I re-checked every item from my review against the code at `47689a26377`, not just the write-up. All six blocking/warning items and all eleven suggestions are genuinely resolved:

- **B-1** — retitled, and overtaken by events: #2736 landed PT-4238's actual fix, so a merge here can't falsely close the bug. This is a regression guard now, which is the right shape.
- **W-2 / W-3 / W-5** — the `TODO(PT-4238)` paragraph is gone entirely; no `PT-4238` reference survives in either file. The doc describes the binding path the grid actually takes and states the constraint without prescribing a fix. The remaining TODOs cite PT-4316 (#2776, open) — forward-facing.
- **W-4** — agreed this PR makes no architecture decision now that the TSDoc doesn't choose a fix, and `adr-column-3-panels-are-told-their-project` is on `main`.
- **W-6** — pinned properly: the `getNetworkEvent` mock returns a token naming the event, and the assertion is on `'event:platformScriptureEditor.onSharedLayoutApply'`. Renaming the event now fails.
- **W-7** — agreed on declining `test.fails`; these tests pass, so it would be the wrong tool. The two `TODO(PT-4316)` block markers with invert-vs-delete instructions do the job.
- **W-8** — the merged test closes the loop in one body. No sibling dependency left.
- **W-9** — the assertion-defending prose is gone, and the cross-references with it.

I confirmed the "no behavior change" claim independently: filtering the hook's diff to non-comment lines leaves nothing.

Worth calling out that the follow-up `/code-review max` pass found four of your own first-round claims wrong and corrected them in `c73568fceea` — including replacing a test that modelled a switch the real stack can't produce. That's the right instinct and it made the tests better.

### Two loose ends — not blocking, but don't let them evaporate

1. **The mount-gap bug needs a ticket.** "Saved cell order can be lost on mount" (the placeholder returned with `isLoading` already false, and the grid's reconcile pruning against it in the same commit) is currently documented only in an `@returns` warning. That's real and pre-existing, but a warning in a TSDoc block is not a work item — if nobody files it, it's lost. Could you file it and link it back here?
2. **Item 3 is unverified.** Panels latching the pre-sync layout on a project switch (`applyForProject` firing before the project-switch sync lands in `main.ts`) is reported but not confirmed. Fine to leave out of scope here — just flagging that it shouldn't be treated as a known-good finding downstream.

### One thing before merge

CI isn't green yet — the three platform builds are still pending from the rebase, and Reviewable shows 2 files left. **This approval is contingent on those going green**; please don't merge on a red build. The rebase onto #2736 touched both files, so the builds are meaningful here.

PR #2654, opened 2026-08-07
Relicense source to AGPL-3.0-or-later

lyonsil:
# Relicense source to AGPL-3.0-or-later

Relicenses this repository's source to AGPL-3.0-or-later, states the terms the distributed
application is licensed to the user under, grants extension authors an explicit exception, replaces
the hand-maintained third-party notices file with one derived from what the build actually ships,
and stops this repository distributing a built application.

The driver is a decision to get serious about licensing: declare the correct license, handle
dependency licenses properly, and make keeping that current a mechanical, safe operation rather than
a maintainer's ongoing burden.

## This repository stops distributing a built application

**Read this section even if you skim the rest — it is a release-pipeline change under a relicensing
subject, and it is the resolution to the largest open question in the last review round.**

`THIRD-PARTY-NOTICES.md` says UBS permits distribution of the lexical database in *Paratext*, and
that the permission does not extend to Platform.Bible. The Publish workflow was uploading a `.snap`
to the public snap store on every run and attaching Windows, macOS and Linux installers to a GitHub
release — every one built as `productName: 'Platform.Bible'`, and every one carrying that database,
which `download-db.ts` puts in strict mode so a missing copy hard-fails the install. Two statements
about one artifact.

The resolution is not to reword the notices. There is no practical path to releasing a separate
Platform.Bible build: the application released to users is **Paratext 10**, built from this source
in `paranext/paratext-10-studio`. So this repository stops distributing:

| Channel | Disposition |
| --- | --- |
| `snapcraft upload --release=edge` | **Removed**, with its `SNAPCRAFT_STORE_CREDENTIALS` env — the repo secret wants revoking separately |
| `ncipollo/release-action` | **Kept, with no binaries.** It still creates the version tag, the draft, the prerelease flag and the generated notes, which is the whole purpose: fixed points in Git history for a downstream build to construct releases from. Now runs on one matrix leg |
| S3 upload of the Windows/macOS installers | **Kept.** Internal sharing, gated on `uploadReleaseAssets`, not advertised |
| `actions/upload-artifact` in `package-main.yml` | **Kept.** Internal sharing |

Installers are still built on every leg — that is how the packaging path is exercised. Nothing
consumes the removed assets programmatically: every `autoUpdater` use in `main.ts` is commented out,
so there is no update feed to break. `README.md`'s Users section now points at Paratext 10 rather
than at GitHub releases and the snap store, and `electron-builder.json5`'s `grade: 'devel'` comment
says the intent is to stay there.

[...]

PR #2775, opened 2026-09-05
PT-4514: Stop marker-problem logging from freezing the app

katherinejensen00:
# PT-4514: Find on a UBS Handbook froze the machine

**Branch**: `pt-4514-dedupe-marker-warnings` · **Base**: `origin/main` · **Files**: 9 (5 generated) ·
**Date**: 2026-09-04, updated 2026-09-11 after review · **Review model**: Claude Opus 5 (1M context)

## Summary

Find on HBKENG hung the app for minutes and took the Windows desktop with it. The cause was **log
volume**, not memory. `UsjReaderWriter` reported an unknown marker and a mismatched marker type once
per *occurrence* during USFM serialization, and Find runs that serialization to place every match.
Commentary and handbook resources are full of markers the Paratext 3.0 markers map does not carry
(HBKENG uses `scr`, `ver`; TNNESP uses `li6`, `pi8`, `brk`), so **one book emitted ~186,000
messages** — each crossing IPC from the web view to main, each appended to the log file
synchronously.

Four changes:

1. **Report each distinct problem once per `UsjReaderWriter`**, not once per occurrence — ~186,000
   messages become a handful.
2. **Extend that to four sibling reports** on the same walk (chapter/verse parse failures, duplicate
   verse). Required making `transferFragmentsInfoArrayToMaps` an instance method so it can reach the
   state.
3. **Stop the constructor's version-mismatch warning serializing the whole document** into one log
   line. It now names the book instead — a chapter of a commentary is megabytes of USX.
4. **Take main's log file writes off the event loop** (`sync = false`), so no future hot log path
   can stall the UI. Main only — see the review-focus note below.

### Definition of Done

- **DoD 1 — "Find on HBKENG should not cause a hang": met, as *does not hang*.** Find reuses one
  cached reader-writer per book, so its reports collapse to one line per distinct problem. PT-4522 is
  still open, so this is not a claim that the scenario in the bug title now works end to end.
- **DoD 2 — "Paratext should not be able to freeze the whole computer": mitigated, not closed.**
  Main no longer blocks on log writes, but `identifyCaller()` still runs synchronously per message on
  main, the extension host stays synchronous (and is where the finder's serialization runs), and
  rotation now fires late. Each is recorded in `adr-log-file-writes-queued`.

## Where to look

| File | What to look at |
| --- | --- |
| `lib/platform-bible-utils/src/scripture/usj-reader-writer.ts` | **The whole change.** `reportProblemOnce` + `reportedProblems` in the new `#region Reporting problems with the document`; the `static` → instance conversion; the constructor warning. |
| `lib/platform-bible-utils/src/scripture/usj-reader-writer.test.ts` | 11 new tests. Each carries a positive control so it cannot pass vacuously; each was verified to go red. |
| `src/shared/services/logger.service.ts` | One `sync = false` behind `isServer()`, its rationale, and a `TODO(PT-4523)` naming the stale-log trade. |
| `.context/standards/Architecture-Decisions.md` | `adr-log-file-writes-queued` — the transport tradeoff and its accepted costs. |
| `lib/platform-bible-utils/dist/*` (5 files) | **Generated — skim only.** Rebuilt from source with `npm run build:basic` after rebasing onto `main`. |

**No public API changes.** Every added member is `private`. No UI, localization, Storybook, or
keyboard-shortcut surface is touched.

## Please focus your review on

- [ ] **The `warn` / `debug` split** — the main judgement call. An unknown marker is *normal* in a
  third-party resource, so it drops to `debug` and is invisible in packaged builds (packaged
  `logLevel` is `info`). A *mismatched type* means the document is malformed and the writer silently
  substitutes the USJ's type, so it stays `warn`. `reportProblemOnce` defaults to `warn` so a new
  caller can't be quietly demoted. **Is that line drawn in the right place?** This is a deliberate
  part of the fix, not a cosmetic one, and it goes beyond the investigation's scope: at a packaged
  transport level above `debug` it suppresses the write on its own, so it overlaps the
  de-duplication, and the measured improvement is the two together. Nothing in `src`, `extensions`,
  `lib`, `c-sharp` or `e2e-tests` consumes the message text.
- [ ] **Accepted trade: the in-app log can be stale.** The transport change stays in this PR because
  DoD 2 covers it, but it introduces PT-4523: `platform.getLogFileContent` reads the file off disk,
  so it — and the Usersnap bug reports it feeds — can miss lines still queued. Marked with
  `TODO(PT-4523)` at the `sync = false` line; a bounded drain of electron-log's queue is the fix.
- [ ] **Main-only transport change.** The extension host stays synchronous on purpose: it ends
  shutdown with `process.exit()`, which discards queued lines including
  `Failed to deactivate extensions`, and logs at `error` when packaged so it never sees this flood.
- [ ] **One log line is now lost.** `main.ts`'s second-instance branch logs the handed-off
  `process.argv` then calls `app.exit()`, which doesn't drain. Accepted deliberately — the obvious
  repair (flipping `sync` back before that log) **fails silently**, because electron-log fixes
  `writeAsync` at `File` construction. Reasoning is in the ADR so nobody repeats it.
- [ ] **De-duplication is per instance, so it is only as good as a caller's instance reuse.** Find
  caches one reader-writer per book and goes to zero repeated lines. Opening a handbook builds
  short-lived instances — per chapter load in the editor web view, per selection, per footnotes pane
  — each starting with an empty set, and still logs ~476 lines for GEN 1:31 in Sebastian's test.
  That is ~2.5 orders of magnitude below the flood, and those writes no longer block main. Nothing
  tests this dependency, so a future path that builds one instance per verse would bring the same
  flood back. The remaining scope is PT-3554.
- [ ] **De-duplication is partial on position-bearing messages.** Marker reports name only the
  marker, so they collapse fully. Chapter/verse reports also name the position, so they collapse
  only where it repeats — measured 400 → 1 for a repeated verse number, 400 → 400 for distinct ones.

## Deferred work is ticketed, not dropped

- **PT-3554** — Opening a handbook still logs each marker problem once per short-lived reader-writer
  (~476 lines for GEN 1:31). Its done-criterion, one report per editor instance, is the remaining work.
- **PT-4521** — `ScriptureFinderPdpe` caches an unbounded number of indexed reader-writers (~95 MB
  per commentary book). Found here, not the cause of this freeze.
- **PT-4522** — Commentary resource search in Get Resources returns nothing for any query.
- **PT-4523** — `platform.getLogFileContent` can return a log missing its newest lines, introduced
  by the transport change. Affects Usersnap bug reports. Marked by `TODO(PT-4523)` in
  `logger.service.ts`.

---

<details>
<summary><b>Findings</b> — 0 critical, 3 important, 8 minor (6 fixed in review, 5 dismissed with reasons)</summary>

### Critical

None.

### Important

- [ ] ~~`main.ts` second-instance branch loses its log line to `app.exit()`.~~ _(Researched and
  deliberately accepted. The proposed repair does not work and fails silently: `FileRegistry.provide`
  returns an already-created `File` before reading `writeAsync`, and `File.writeAsync` is set only in
  the constructor with no setter — so whether it appears to work depends on nothing having logged
  earlier, an invisible ordering dependency with nothing to test against. Alternatives (`app.quit()`,
  direct `appendFileSync`) change shutdown semantics or bypass the logger. Recorded in the ADR.)_
- [x] **The out-of-order-verse report was the one console call in the fragment walk not routed
  through `reportProblemOnce`** _(fixed during review: measured, not assumed. A descending document
  gives 399 calls and **399 distinct messages**, so routing it would drop no lines while adding 399
  set entries — strictly worse. Left per-occurrence with a comment recording why.)_
- [x] **`reportedProblems` doc claimed cost is "bounded by distinct problems, not document size"**
  _(fixed during review: only true for the marker reports; corrected to distinguish them from the
  position-bearing ones.)_

### Minor

- [x] **`reportProblemOnce` was in the wrong `#region`** _(fixed: moved to its own region beside the
  field it owns.)_
- [x] **Duplicate-verse message baked a newline + ~18 spaces of source indentation into the log
  text** _(fixed: hoisted to a local const.)_
- [x] **Version-mismatch warning named no book, so you couldn't tell which document mismatched**
  _(fixed: added `getBookIdForLogging`, reading the book marker off already-in-memory top-level
  content, called only on the path about to log. Covered by a test.)_
- [x] **`Mismatching marker type` lost its `Warning: ` prefix** _(fixed: restored — nothing in-repo
  matches on it, but external log tooling might, and it costs nothing.)_
- [ ] ~~Message args including `JSON.stringify` are built eagerly then discarded.~~ _(Measured at
  186,428 calls: 97.8ms eager vs 9.8ms with a cheap key — but `toUsfm` is memoized, so ~88ms once per
  instance, not per Find result. A thunk API costs readability for 88ms once.)_
- [ ] ~~`sync = false` has no test.~~ _(What matters — a flood cannot block the loop — is not
  observable at this seam, and asserting the property was set would mirror the implementation. Tests
  mock `logger.service` away entirely. Reasoned in the ADR rather than left silent.)_
- [ ] ~~`debug` removes unknown-marker reports from packaged builds.~~ _(Intended, and narrow: only
  that report was demoted.)_
- [ ] ~~A second ad-hoc report-once accumulator exists at `use-editor-pdp-sync.hook.ts`.~~ _(Not
  close enough in shape or lifetime to share; revisit if a third appears.)_

### Template propagation

None — no `#region shared with` markers in any changed file, no `extensions/` changes.

</details>

<details>
<summary><b>Design rationale</b> — author's reasoning on the four non-obvious decisions</summary>

- **Per-instance rather than a module-level `WeakMap`**: "The flood happens inside one conversion —
  one instance walking one document tens of thousands of times. Per-instance already makes the cost
  proportional to distinct problems instead of document size, which is the whole bug. A WeakMap would
  be module-level shared state in a library: unrelated consumers would silently affect each other's
  logging, and nothing could reset it." On per-action re-reporting: "That's a better bound than 'once
  ever' — the first Find in a session shouldn't permanently silence a problem for every later one."
- **Keying on message text rather than marker identity**: "the key is then exactly what would land in
  the log: identical key ⇔ identical line, so de-duplication can never hide information. A
  marker-identity key is a second thing to keep correct, and it would suppress two genuinely
  different reports about the same marker."
- **`warn` vs `debug`**: "The split is a claim about the document, not a volume knob — both are
  deduped either way, so the level buys nothing on that front."
- **Main-only transport**: main owns the UI and receives every web view's forwarded console calls;
  the extension host would lose shutdown diagnostics for no benefit.

</details>

<details>
<summary><b>Verification</b></summary>

As of 2026-09-11, after rebasing onto `main` at `014cdcd` and addressing review:

- `platform-bible-utils`: `npm run build:basic` exit 0; `vitest run` 623 passed, 1 skipped.
- The new tests were checked by mutation. Each of these turns its test red: re-adding
  `JSON.stringify(this.usj)` to the version-mismatch warning; replacing the `NO_BOOK_ID` fallback
  with `''`; routing the unparseable-chapter or no-verse-digits report back to a bare `console.warn`.
  The verse `Number.isNaN` branch has no test because it cannot run: `VERSE_MARKER_NUMBER_SPAN_REGEXP`
  captures `(\d+)`, which always parses.
- `npm run lint --workspace=platform-bible-utils` — 0 errors, 1 pre-existing warning. ⚠️ **The root
  lint config ignores `lib/platform-bible-utils`**, so a green root run says nothing about that
  package.
- `npm run lint` (root) — 2 errors, both `import/no-unresolved` for `spdx-license-list/full`: the same
  missing module as the `erb` typecheck below, in files this branch does not touch. Nothing else.
- `npm run typecheck` — `core`, `e2e` and `workspaces` pass. `erb` fails in the review worktree only:
  `spdx-license-list/full` is not installed there, because `main` added it to `package.json` (#2654)
  after that checkout's install. This branch touches nothing under `.erb`.
- `npm test` (root) — not run this round, because that worktree's install predates dependencies
  `main` has since added, so a root run there would not be trustworthy. `platform-bible-utils`, the
  only package whose code changed, passes in full above; CI runs the root suite.
- Prettier over changed files — clean.
- No C# changes, so `dotnet test` was not run.

Only the generated bundles conflicted in the rebase; they were rebuilt from source rather than
merged. The architecture-decision log merged as a clean union: 85 slugs, in `LC_ALL=C sort` order,
no duplicates.

</details>

<details>
<summary><b>Positive observations</b></summary>

- Every new test carries a positive control with a comment explaining why — `toBeGreaterThan(0)`
  before a uniqueness assertion, a `brk` marker proving the second pass re-serialized rather than
  returning a cached string — and each was verified to go red by temporarily reverting the
  implementation.
- The `static` → instance conversion is the minimal change needed to reach `this`: one call site, no
  external references, local helper converted to an arrow so `this` binds, no incidental refactoring.
- The ADR is candid about costs it accepts: tail loss on hard exit, `app.exit()` paths, stale
  `getLogFileContent`, late rotation against the 3 MB `maxSize`, shared `main.log` interleaving, and
  the residual per-message `identifyCaller()` cost that queuing does *not* remove.
- Comments are forward-facing — constraints and rationale, not change narration. No ticket-ID residue.
- Per-instance scoping is sound for the motivating path: `ScriptureFinderPdpe` caches the reader-writer
  per book, so Find reuses one instance rather than rebuilding per query.

</details>

AI-assisted — [session](https://claude.ai/code/session_018umbSKTsasML4nGsWmVDFu)

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2775)


Sebastian-ubs:
@katherinejensen00 I tested https://github.com/paranext/paranext-core/actions/runs/33936647205 and found the following:
Good
- Find did not freeze the app
- Using Find does not log any "Unexpected ... marker" statements

Bad
- Opening a handbook logs less, but still quite a lot of these statements
476 log statements for Gen 1:31

jolierabideau:
## Review: approve with changes

Reviewed the four analysis angles (API/correctness, style/patterns, coverage, scope-vs-ticket). **No critical findings.** The core fix is correct, well-targeted, and the ADR is some of the best disclosure I've seen on a change like this — the `writeAsync`-is-fixed-after-first-write trap in particular is exactly the kind of thing that would otherwise get "repaired" wrongly in six months.

### Does it solve PT-4514?

**DoD 1 ("Find on HBKENG should not cause a hang") — yes, and the mechanism confirms it rather than just correlating with it.** Per-instance de-duplication is the right granularity *for Find specifically*, because the Find path caches one reader-writer per book and reuses it across every match (`platform-scripture-finder-pdpe.model.ts:1075`/`:1079`; `search-results-in-book.component.tsx:116`). So the measured 186,428 warns per single location conversion collapse to one line per distinct problem per book.

**DoD 2 ("Paratext should not be able to freeze the whole computer") — mitigated, not closed.** `sync = false` takes the blocking write off main's loop (confirmed `isServer()` is main-only, `internal-util.ts:17`). Still open, and the ADR honestly says so: `identifyCaller()` still runs synchronously per message on main, the extension host stays synchronous *and is where the finder's serialization actually runs*, and rotation now fires late.

### On @Sebastian-ubs's 476-statements-per-navigation follow-up

Traced it — it's a direct consequence of the per-instance scoping, not a leftover. The handbook-**open** path builds short-lived instances repeatedly, each starting with an empty `reportedProblems`: `platform-scripture-editor.web-view.tsx:3269` (a `useMemo` keyed on `usjFromPdp`, so one per chapter load), `platform-scripture-editor.utils.ts:459` (per selection→`scrRef`), `character-marker-coverage.utils.ts:134` (per coverage computation on selection change), plus the two footnote panes and `enhanced-resource.web-view.tsx:2545`. Find is the one path with a cached long-lived instance — which is exactly why Find went to zero and opening didn't.

476 is ~2.5 orders of magnitude below the 186k that froze the machine, and those writes no longer block main's disk I/O, so **I don't think it blocks this PR.** But it's worth a line in the description and its own ticket, because the fix's effectiveness is now a function of how many instances a call path builds, and nothing tests or documents that dependency. A future path that constructs one per verse re-opens the same flood class.

### Three things I'd like settled before merge

**1. One coverage hole, and it's on a load-bearing part of the fix.** The constructor warning no longer serializing the whole document isn't pinned. Verified by mutation: re-adding `\nUSJ: ${JSON.stringify(this.usj)}` to the constructor message leaves the entire file green (74 passed). The `names the book when warning that the markers map version does not match` test only asserts `message.includes('USJ for book MRK')`, which the megabyte form also satisfies. An `expect(message).not.toContain('USJ: {')` or a length bound closes it.

**2. A comment-discipline pass** against `.claude/rules/code-quality/forward-facing-comments.md`. Two concrete sites: `usj-reader-writer.ts:3232-3237` — "It also stays quiet on the documents that motivated de-duplicating the others" narrates this PR's rationale and dies at merge (the first half, about the message already being distinct per occurrence, is genuinely forward-facing — keep that); and `:343-346`, which opens with "Names the book rather than serializing the document", i.e. what the code stopped doing. It's ~45 lines of comment for ~25 lines of new logic, so I'd suggest one deliberate pass rather than site-by-site edits.

**3. Whether `logger.service.ts:25` belongs in this PR.** The investigation deferred exactly this as "worth hardening independently of this bug", and it introduced PT-4523 — a log that can be missing its newest lines, which is what Usersnap attaches to user bug reports. **My read is that DoD 2 legitimately authorizes it and I would not insist on a split** — but then the trade should be explicit in the description, and `logger.service.ts` should carry a `TODO(PT-4523)`. Right now the code reads as a settled decision and the defect is discoverable only from ADR prose. Your call; flagging so it's a decision rather than a side effect.

### Minor

- `warn` → `debug` (`:2074`) was also flagged as a non-goal in the investigation. It's not cosmetic — at a packaged transport level above `debug` it suppresses the write on its own, so it overlaps the de-dup and confounds attribution of the measured improvement. Grepped for consumers of the message text across `src`, `extensions`, `lib`, `c-sharp`, `e2e-tests` — none, so nothing breaks. Keep it, just own it explicitly.
- `usjChanged()` (`:376-382`) still says "If new variables are created to speed up queries, they should be reset here", and `reportedProblems` is a new field intentionally *not* reset there. Documented on the field and tested, but the note a maintainer will read sits in `usjChanged` — one clause there closes the loop.
- ADR **Consequences** packs six independently actionable items into one prose block; comparable entries (`adr-runaway-data-hook-guard`, `adr-toolbar-sync-status-is-local`) use sub-bullets, which is where the format earns its keep.
- Untested: `getBookIdForLogging()`'s `NO_BOOK_ID` fallback, and three of the four rerouted fragment-walk branches (both `Number.isNaN` paths and the missing-regex-match path). Low risk since de-dup is generic in the helper, but the *routing* isn't pinned and that's the easy thing to miss on one branch.
- `getLogFileContent` could do a bounded drain rather than only documenting the staleness — `transports.file.getFile()` exposes `asyncWriteQueue`/`hasActiveAsyncWriting`, so it's reachable without patching the library.
- Worth saying on the PR that DoD 1 is satisfied as "does not hang", not "Find on HBKENG works" — PT-4522 is still open, so the literal scenario in the bug title remains unusable.
- Merge conflicts are **entirely in the generated bundles** — rebuild `dist/`, don't hand-merge minified output. `Architecture-Decisions.md` auto-merged and is clean right now, but per `CLAUDE.md` that log has to be merged as a union keyed by slug, so please re-run the duplicate-slug + `LC_ALL=C sort` check after resolving; a plain auto-merge can duplicate an entry silently and nothing else catches it.

### What's notably good

- **The "9 tests, each with a positive control" claim checks out**, and I verified it by mutation rather than taking it on faith: removing the de-dup guard fails 5 of 9 tests; collapsing the `warn`/`debug` split into a single `console.warn` fails 6 of 9 — so the level split is genuinely pinned, including that the mismatching-type report stays a `warn`. The `usjChanged` test using `brk` as a control to prove the second pass really re-serialized is a nice touch. Tests are proportional-not-absolute (10 vs 500 occurrences → same count), which is the right shape for a de-duplication property and won't churn when an unrelated report is added.
- **No public API change**, verified — every added member is `private`, and `papi.d.ts` correctly needs no regeneration since it doesn't carry `UsjReaderWriter` at all.
- **The `static` → instance conversion is minimal and load-bearing**, not a blanket de-static-ing: the class keeps 19 other `private static` helpers. The arrow-function conversion was the file's only inner `function` declaration, and all four call sites follow the declaration, so no TDZ hazard.
- **The one remaining per-occurrence `console.debug` (`:3238`)** was measured (399 calls → 399 distinct messages), not assumed, and carries its reasoning at the site. That's a deliberate boundary rather than a sweep.
- `#region` convention followed exactly including the repeated `#endregion` label; the `no-type-assertion` disable mirrors the existing pattern at lines 630/652/713 verbatim; every code claim in the ADR verified (`main.ts:259`, `extension-host.ts:47`, the 3 MB `LOG_FILE_MAX_SIZE_BYTES`); ADR slug in correct byte-order position with all 66 slugs sorted and no duplicates; committing `dist/` matches repo convention and the output is consistent with source, not stale.
- Follow-ups ticketed rather than dropped (PT-4521/4522/4523), and console spies restored in a `finally` so a throwing serialization can't leak a mock into later tests.

Nice work on this one — the hard part (finding that the volume scales with document size, not result count) is clearly reflected in where the fix landed.

---

*AI-assisted review.*

katherinejensen00:
Thanks both. Rebased onto `main` (`014cdcd`), with `dist/` rebuilt from source rather than hand-merged. Point by point:

**@jolierabideau**

1. **Constructor-warning coverage hole** — fixed. The test now puts sentinel text in the document and asserts the version warning does not contain it, so any way of serializing the document into the message fails it, not only the `USJ: {` spelling. Verified by mutation: re-adding `\nUSJ: ${JSON.stringify(this.usj)}` turns it red.
2. **Comment-discipline pass** — done. Cut the "stays quiet on the documents that motivated…" sentence and kept the half about every occurrence already being distinct; rewrote the constructor comment as the constraint it is (keep the document out of the message) rather than what the code stopped doing; condensed the `reportedProblems` doc; and re-read every other comment the branch adds against `forward-facing-comments.md`.
3. **`logger.service.ts`** — kept in this PR under DoD 2. It now carries a `TODO(PT-4523)` at the `sync = false` line naming the stale-log trade, and the description states the trade explicitly.

Minor:

- **`warn` → `debug`** — owned explicitly in the description, including that it suppresses the write on its own in packaged builds and so overlaps the de-duplication.
- **`usjChanged`** — its reset-here note now says `reportedProblems` is deliberately kept.
- **ADR Consequences** — split into sub-bullets with the content unchanged; the stale-read item now points at PT-4523.
- **Untested routing** — added tests for the `NO_BOOK_ID` fallback, the unparseable-chapter report, and the no-verse-digits report. Each goes red under mutation (fallback → `''`; each report routed back to a bare `console.warn`). The fourth, the verse `Number.isNaN` branch, is unreachable: `VERSE_MARKER_NUMBER_SPAN_REGEXP` is `/(\d+)-?(\d+)?/`, so the captured group always parses. I left that branch alone rather than widen the PR.
- **Bounded drain in `getLogFileContent`** — deferred to PT-4523, which is that work.
- **DoD 1** — the description now says it is met as "does not hang", with PT-4522 still open.
- **Merge** — only the generated bundles conflicted; I took `main`'s and rebuilt. The decision log came through as a clean union: 85 slugs, in `LC_ALL=C sort` order, no duplicates.

**@Sebastian-ubs** — thanks for testing it. The 476 lines on opening a handbook are the per-instance scoping Jolie traced. Find caches one `UsjReaderWriter` per book, so it reports each problem once. The open path instead builds short-lived instances — per chapter load in the editor web view, per selection, per footnotes pane — and each starts with an empty set. That is about 2.5 orders of magnitude below the flood that froze the machine, and main no longer blocks on those writes, so it stays out of this PR. It is the remaining scope of PT-3554, where you have already posted the numbers.

**Verification:** `platform-bible-utils` 623 passed / 1 skipped, with its workspace lint at 0 errors (1 existing warning); root lint and `npm run typecheck` are clean apart from `spdx-license-list/full` being unresolved (2 lint errors plus `typecheck:erb`). That's a stale local install, not this branch; details are in the description. I didn't run the root `npm test` locally for the same reason, so CI is the check for that.

katherinejensen00:
Thanks both. Rebased onto `main` (`014cdcd`), with `dist/` rebuilt from source. Point by point:

**@jolierabideau**

1. **Constructor-warning coverage hole** — fixed. The test now puts sentinel text in the document and asserts the version warning does not contain it, so any way of serializing the document into the message fails it, not only the `USJ: {` spelling. Verified by mutation: re-adding `\nUSJ: ${JSON.stringify(this.usj)}` turns it red.
2. **Comment-discipline pass** — done. Cut the "stays quiet on the documents that motivated…" sentence and kept the half about every occurrence already being distinct; rewrote the constructor comment as the constraint it is (keep the document out of the message) rather than what the code stopped doing; condensed the `reportedProblems` doc; and re-read every other comment the branch adds against `forward-facing-comments.md`.
3. **`logger.service.ts`** — kept in this PR under DoD 2. It now carries a `TODO(PT-4523)` at the `sync = false` line naming the stale-log trade, and the description states the trade explicitly.

Minor:

- **`warn` → `debug`** — owned explicitly in the description, including that it suppresses the write on its own in packaged builds and so overlaps the de-duplication.
- **`usjChanged`** — its reset-here note now says `reportedProblems` is deliberately kept.
- **ADR Consequences** — split into sub-bullets with the content unchanged; the stale-read item now points at PT-4523.
- **Untested routing** — added tests for the `NO_BOOK_ID` fallback, the unparseable-chapter report, and the no-verse-digits report. Each goes red under mutation (fallback → `''`; each report routed back to a bare `console.warn`). The fourth, the verse `Number.isNaN` branch, is unreachable: `VERSE_MARKER_NUMBER_SPAN_REGEXP` is `/(\d+)-?(\d+)?/`, so the captured group always parses. I left that branch alone rather than widen the PR.
- **Bounded drain in `getLogFileContent`** — deferred to PT-4523, which is that work.
- **DoD 1** — the description now says it is met as "does not hang", with PT-4522 still open.
- **Merge** — only the generated bundles conflicted; I took `main`'s and rebuilt. The decision log came through as a clean union: 85 slugs, in `LC_ALL=C sort` order, no duplicates.

**@Sebastian-ubs** — thanks for testing it. The 476 lines on opening a handbook are the per-instance scoping Jolie traced. Find caches one `UsjReaderWriter` per book, so it reports each problem once. The open path instead builds short-lived instances — per chapter load in the editor web view, per selection, per footnotes pane — and each starts with an empty set. That is about 2.5 orders of magnitude below the flood that froze the machine, and main no longer blocks on those writes, so it stays out of this PR. It is the remaining scope of PT-3554, where you have already posted the numbers and I assigned to myself and put in to do to fix in a spare moment later.

**Verification:** `platform-bible-utils` 623 passed / 1 skipped, with its workspace lint at 0 errors (1 existing warning); root lint and `npm run typecheck` are clean apart from `spdx-license-list/full` being unresolved (2 lint errors plus `typecheck:erb`). That's a stale local install, not this branch; details are in the description. I didn't run the root `npm test` locally for the same reason, so CI is the check for that.

jolierabideau (approved):
## Re-review: approved

Every point from my earlier review is addressed or deferred to a ticket, and I checked each fix against the code:

1. **Constructor-warning coverage.** The unique-text check is stricter than the `USJ: {` assertion I suggested. It fails however the document gets serialized into the message.
2. **Comment pass.** The backward-facing sentences are gone, and the constructor comment now states the rule the code follows.
3. **`logger.service.ts`.** `TODO(PT-4523)` is at the `sync = false` line, and the description states the stale-log trade.

Minor items:
- The `warn` → `debug` change, the `usjChanged` note, and the DoD 1 wording (with PT-4522 open) are all in.
- The ADR Consequences are split into sub-bullets.
- Three of the four branches I flagged now have tests. I agree the verse `Number.isNaN` branch is unreachable: `VERSE_MARKER_NUMBER_SPAN_REGEXP` only captures digits, so the number always parses.
- The bounded drain moving to PT-4523 and the handbook-open volume moving to PT-3554 both make sense as follow-ups.

Thanks for the thorough point-by-point. 🎉

---

*AI-assisted review.*

PR #2747, opened 2026-09-01
PT-4496: Mount the core22 GNOME platform snap in Linux snap builds

irahopkinson:
# Code Review Summary

**Branch**: fix/snap-gnome-platform-plug

**Base**: origin/main

**Date**: 2026-09-02

**Review model**: Claude Opus 5

**Files changed**: 6

## Overview

Linux snap builds declare `base: core22` but mounted `gnome-3-28-1804` — the Ubuntu 18.04 / GNOME
3.28 content snap — at `$SNAP/gnome-platform`. electron-builder's snapcraft template hardcodes that
plug and rewrites only `snap.base`, so the app ran jammy-staged libraries against a bionic platform:
Mesa/DRI, libdbus and the GTK/Wayland stack all mismatched, and the app segfaulted on launch without
ever opening a window.

The first commit on this branch overrode the plug's `content`/`default-provider` under the
template's original *name*. That fixed cold installs — the path the bug was originally diagnosed on
— but code review raised the upgrade path, which had not been considered. snapd matches stored
connections by plug **name** on refresh, so it revived the stale connection *and* auto-connected the
new one; two content plugs then competed for one mount point and snapd broke the tie by renaming one
aside, arbitrarily. The second commit therefore **renames** the plug, via a `patch-package` patch on
the template — the only mechanism that can, since electron-builder config can replace a template
plug entry but never rename or remove one.

## API Changes

None. No public API surfaces changed. The diff touches only snap packaging config, a patch file, a
build-config test, the decisions log, and dependency manifests — nothing in
`lib/platform-bible-react/`, `lib/platform-bible-utils/`, `lib/papi-dts/papi.d.ts`, or any
`extensions/src/*/src/types/*.d.ts`.

### Critical — Must address before merge

- [x] **A core-owned patch at `patches/app-builder-lib+26.7.0.patch` would break the
      paratext-10-studio build.** The studio does not carry its own patch:
      `repo-patches/paranext-core.patch` *creates* that same path here to add mercurial snap parts.
      The studio applies repo patches with `git apply --3way` (`lib/git.util.ts:130`), which turns
      the add/add case into `<<<<<<< ours` / `>>>>>>> theirs` markers written **into the patch file
      itself**, leaving patch-package unable to parse it. _(fixed during review: the patch is named
      `app-builder-lib+26.7.0+002+rename-gnome-platform-plug.patch`, using patch-package's sequenced
      patch support, so it occupies a different path from the studio's and the two apply in sequence
      to the same template.)_

[...]

PR #2777, opened 2026-09-07
PT-4524: Assert the snap metadata electron-builder generates

irahopkinson:
## Summary

Adds a CI check that reads the metadata of the snap this build actually produced and asserts it
mounts exactly one GNOME platform content plug, named, `default-provider`'d and content-tagged for
the platform snap that pairs with the snap's own `base`.

The existing guard (`.erb/scripts/electron-builder-snap-config.test.ts`, added in #2747) reads the
**inputs**: this repo's `electron-builder.json5` and the patched `app-builder-lib` snapcraft
template. Everything that turns those inputs into the plug list a snap ships lives in
`app-builder-lib/out/targets/snap.js` — plug merging by key, `confinement: 'classic'` deleting the
plugs map wholesale, `base` rewriting — and none of it is visible from the inputs. So an
electron-builder that emitted its own GNOME platform plug in code, while the template still declared
one, would ship two plugs on `$SNAP/gnome-platform` with every existing assertion still green.
PT-4496 measured what that costs: snapd breaks the tie by renaming one mount aside arbitrarily, and
across 16 upgrades of an identical build the app launched 9 times and segfaulted 7.

The guard was not wrong, just anchored one level too early. This adds the second anchor.

### Why review this

Not part of the current epic. It is the deferred half of the #2747 review — findings #33 and #30,
both endorsed there and postponed on the reviewer's own "take whichever keeps this branch moving".
Worth reviewer time now rather than later because the reasoning is still fresh from #2747, and
because the check is what stops the PT-4496 fix silently regressing on a future electron-builder
bump.

## Changes

Two commits: the check, and one response to Katherine's review.

- **`.erb/scripts/assert-generated-snap-metadata.util.ts`** (new) — locates the built snap, reads
  `meta/snap.yaml` out of it, and returns a list of problems with what it found.
- **`.erb/scripts/assert-generated-snap-metadata.util.test.ts`** (new) — 17 tests.
- **`.erb/scripts/assert-generated-snap-metadata.ts`** (new) — CLI entry; exits non-zero on any
  problem.
- **`.erb/scripts/snap-platform-pairing.ts`** (new) — the base-to-platform-snap map, shared so both
  guards read one copy. Two copies that drifted would leave each passing against its own stale idea
  of the correct pairing.
- **`.erb/scripts/electron-builder-snap-config.test.ts`** — imports those constants, checks
  `content`, and asserts that every snap-building workflow invokes the new check.
- **`.github/workflows/test.yml`, `package-main.yml`, `publish.yml`** — each runs the check in the
  same `run:` block as its Linux build, so a bad snap fails where it is produced. `test.yml` hangs
  off the existing `check packaging` step, so this adds no packaging run of its own.
- **`tsconfig.erb.json`** — `.erb/scripts` covered by glob rather than an allowlist.
- **`package.json`**, **`cspell.json`**, **`electron-builder.json5`**,
  **`Architecture-Decisions.md`** — the npm script, two dictionary words, and both prose pointers
  updated to name both halves of the enforcement.

### It reads the shipped snap, not the file the ticket named

PT-4524 asks for an assertion on the generated `snapcraft.yaml`. That file is snapcraft's **input**:
snapcraft reads it and emits `meta/snap.yaml`, so only the latter reflects what a machine actually
installs. A completed packaging run also leaves nothing at the staged path — observed on a
diagnostic CI run, where `release/build` held only `linux-unpacked/`, `.icon-set/`,
`builder-debug.yml` and the snap.

Worth stating precisely, because a first version of this PR got the mechanism wrong: electron-builder
does **not** delete that stage directory (`snap.js` takes a raw path from `createStageDirPath` and
registers no cleanup, unlike the targets using `createStageDir`), and does not write `meta/snap.yaml`
for this repo's configuration either — snapcraft does. Both facts came from review. The conclusion
stands on the input/output distinction rather than on a cleanup mechanism.

Reading it needs `unsquashfs`, present on the Ubuntu runners and a snapd dependency regardless. Only
the single file is extracted, not the 242 MB image.

### Three decisions worth a look

**`base` is read from the snap's own metadata, not from `electron-builder.json5`.** Deliberate:
taking it from the config would check the artifact against an input the other guard already covers,
rather than against itself.

**Nothing was removed from the unit-level guard.** PT-4524's DoD asks that any part made redundant
be deleted. On review, nothing is: the unit guard runs in milliseconds on all three OSes before
anything is built, this runs on Linux after a full package, and they catch different failures.

**Finding nothing is an error, not a pass.** Zero snaps, several snaps, a snap without
`meta/snap.yaml`, an unknown `base`, unparseable YAML — each fails loudly. A guard written to close a
blind spot that passed by finding nothing to inspect would ship as a check that never checks
anything. That rule is what turned an early wrong-path attempt into a red build rather than a silent
green.

## AI Involvement

AI-assisted. The check, its tests and the workflow steps were AI-generated under my direction and
reviewed by me. Claims about app-builder-lib's behaviour were verified against a real packaging run
or against upstream source rather than inferred — the first version of this check reasoned from
`snap.js` alone and was wrong about where the metadata lives.

## Testing

**CI runs on this PR**, against the current head. The Linux leg exercises the check for real: it
builds a snap in `check packaging` and the new step reads it.

**Red/green, against real snaps.** Before this PR was retargeted at `main` it had no CI, so the check
was proved out by dispatched runs on an earlier revision. Those runs predate several rebases, so they
describe this check rather than this SHA:

- [Green](https://github.com/paranext/paranext-core/actions/runs/34079154005) — passes against a real
  Linux packaging run, all three OS jobs green.
- [Red](https://github.com/paranext/paranext-core/actions/runs/34083422640) — on a throwaway branch
  since deleted, which unpacked the real built snap, rewrote `meta/snap.yaml`, repacked it and ran
  the check three times:

| Cycle | Result |
| --- | --- |
| Stale plug name, as a reverted `app-builder-lib` patch ships | `caught` |
| A second plug competing for the same mount point (#30) | `caught` |
| Unmutated control | `passes` |

The control is what makes the two reds mean anything: a check that reddened for both mutations and
also for an untouched repack would be reporting on the repacking rather than on the metadata.

**Locally**: `npm test` exit 0 (8,097 tests), `npm run typecheck` exit 0 on all four stages,
`npm run lint` exit 0. Eleven mutations of the checker, each reddening exactly the tests covering it
— including the two "must not pass by finding nothing" guards, the `content` check, the YAML
parse-catch, the workflow-coverage predicate in both directions, and the `isFile()` case review
caught as non-falsifiable.

**Type-check coverage is itself verified**: a planted type error in a newly created
`.erb/scripts` file fails the gate, which is what makes the glob a fix for the class rather than for
the three files this PR happens to add.

**Unrelated flake seen during this work**: `npm unit tests` on `windows-latest` timing out at
vitest's 5s default — the PT-4504 / PT-4526 class. It cancelled one dispatched run through
`fail-fast`. #2759 fixes both halves.

## Risk Level

Low. No application code is touched — build-time scripts, their tests, and CI steps. The worst
failure mode is a false red on the Linux leg, which blocks a merge rather than shipping anything. The
behaviour worth scrutinising is the opposite case: that the check cannot pass by finding nothing,
which is what the not-found, ambiguity and parse guards exist for and what the mutations above
exercise.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2777)


katherinejensen00 (commented):
# PR #2777 — PT-4524: Assert the snap metadata electron-builder generates

Review comments for the PR owner. 10 comments, ordered by severity.

**Top-level framing:** Solid PR — the check is real, and the red/green/control testing is the right
shape. Two items worth blocking on (#1, #2); the rest is refinement.

- PR: https://github.com/paranext/paranext-core/pull/2777
- Branch: `pt-4524-assert-generated-snapcraft-yaml`
- Base: `fix/snap-gnome-platform-plug` (stacked on the still-open #2747)

---

## 1. [High] The guard doesn't cover the snap that reaches users

**Attach to:** `.github/workflows/test.yml:234`

```
      - name: check generated snap metadata
```

The guard is wired into `test.yml` only. `publish.yml:204-205` is what runs
`snapcraft upload --release=edge`, and it builds unguarded.

Note the shape of the fix: that workflow builds and uploads inside one `run:` block, so the assert
has to go *between* `npx electron-builder --publish never --linux` and the upload line, not as a
following step.

`publish.yml` is `workflow_dispatch`-only on arbitrary refs with its own snapcraft/lxd stack
(lines 94-111), so it can genuinely diverge from what `test.yml` validated. `package-main.yml:140`
is also unguarded but far lower risk — same branch triggers as `test.yml`, so the same commit gets
checked concurrently. The sibling guard already enumerates all three in `LINUX_RUNNER_WORKFLOWS`.

---

## 2. [High] The three new `.erb/scripts` files are type-checked by nothing

**Attach to:** `package.json:44`

```
    "assert:snap-metadata": "tsx ./.erb/scripts/assert-generated-snap-metadata.ts",
```

The PR's `typecheck exit 0` evidence doesn't cover the new code.

`npx tsc -p tsconfig.erb.json --listFiles | grep erb/scripts` prints only the two
`startup-waterfall` files — that `include` is a literal allowlist, and nothing pulls the snap
modules in transitively. Root tsconfig is `src`/`assets` only.

I checked the obvious defense: a scratch file in `.erb/scripts/` with four blatant type errors
produced **0 errors** from `npm run lint`, because type-aware ESLint rules never surface compiler
diagnostics — `.erb/scripts` being in `tsconfig.lint.json` buys nothing here.

This is precisely the gap `tsconfig.erb.json`'s own header says the gate exists to close. All three
files pass `tsc` cleanly today, so it's a zero-fallout 3-line `include` addition.

---

## 3. [Medium] Neither guard checks `content`, the attribute that decides whether the plug connects

**Attach to:** `.erb/scripts/assert-generated-snap-metadata.util.ts:150`

```
  if (descriptor['default-provider'] !== key)
```

`default-provider` decides which snap snapd *installs*; `content` decides whether it connects to
that snap's slot, defaulting to the plug name when absent. So `gnome-42-2204` +
`default-provider: gnome-42-2204` + `content: gnome-3-28-1804` installs correctly, fails to
auto-connect, mounts nothing — and both guards return zero problems.

Reachable, not theoretical:

- The template declares no `content` key, so today's pass is by default, not by assertion.
- `snap.js:146` assigns `snap.plugs[plugName] = plugOptions` verbatim.
- This repo's `snap.plugs` is already a mixed array using the attribute-object form for
  `dot-platform-bible`.
- The ADR records that the first attempted PT-4496 fix overrode exactly
  `content`/`default-provider`.

Suggest `const content = descriptor.content ?? key;` checked against `expected`, mirrored in the
input guard.

---

## 4. [Medium] The `isFile()` test passes whether or not the code it names exists

**Attach to:** `.erb/scripts/assert-generated-snap-metadata.util.test.ts:203`

```
    mkdtempSync(path.join(outDir, 'staging.snap-'));
```

`mkdtempSync` appends six random chars, producing `staging.snap-h4YdqS`, which fails
`endsWith('.snap')` on its own — so `entry.isFile()` is never reached.

Confirmed by mutation: deleting `entry.isFile() &&` leaves all 14 tests green.
`mkdirSync(path.join(outDir, 'staging.snap'))` fixes it.

---

## 5. [Medium] Comment density is working against the comments

**Attach to:** `.erb/scripts/snap-platform-pairing.ts:1`

```
/**
```

Individually good, collectively repetitive enough that people will skim past all of it.

92 comment lines to 278 code lines across the four new files — and ~45 of those "code" lines are
multi-line English error strings, so real prose-to-logic is nearer 1:2. This file is 17 comment
lines around 7 lines of constants.

The same rationale (`snap.js` produces the merge, invisible from the inputs) is restated
near-verbatim in five places: `assert-generated-snap-metadata.ts:5-9`, here at 2-8,
`util.ts:26-28`, `util.ts:90-94`, `test.yml:228-233`.

Same applies to the two CI failure messages, which spend their middle third on snapd refresh
semantics — the actionable half of the name-mismatch one (an unpatched template produces this)
currently comes last.

One canonical statement plus pointers drops ~22 lines with no fact lost.

---

## 6. [Low] Right conclusion, but both stated reasons for reading `meta/snap.yaml` are wrong

**Attach to:** `.erb/scripts/assert-generated-snap-metadata.util.ts:11`

```
 * The intermediate `snapcraft.yaml` electron-builder generates is not an option: it is written into
```

This will cost the next reader confidence in the rest of the reasoning.

**The stage directory does survive.** `snap.js:174` calls `createStageDirPath` (raw path), not
`createStageDir`, so no `StageDir` exists and `cleanup()` is never registered — compare
`AppxTarget` / `MsiTarget` / `FlatpakTarget`, which all pair the two.

**electron-builder doesn't write `meta/snap.yaml`.** `isUseTemplateApp` (snap.js:45-49) needs
`buildPackages` empty and `stagePackages` at defaults; this repo trips both, so it emits
`snap/snapcraft.yaml` and *snapcraft* produces `meta/snap.yaml`. The first commit message on this
branch already says this.

The reason that holds is better anyway: the staged file is snapcraft's input, and only the shipped
file reflects what installs.

---

## 7. [Low] `checkGeneratedSnapMetadata` can throw, contrary to its docblock and a test name

**Attach to:** `.erb/scripts/assert-generated-snap-metadata.util.ts:97`

```
  const metadata: unknown = parseYaml(yamlText);
```

Ran the repo's `yaml` against three malformed inputs — duplicate keys, tab indentation, unclosed
flow sequence — all throw `YAMLParseError`.

The test named `reports metadata that is not a YAML mapping rather than throwing` uses
`'just a string'`, a valid YAML scalar, so it never reaches that path.

Contained (the CLI's outer catch still reddens the build), but the failure prints a bare parser
error naming no snap and no file.

---

## 8. [Low] A failure discards the snap you'd want to inspect

**Attach to:** `.github/workflows/test.yml:236`

```
        run: npm run assert:snap-metadata
```

`Upload Linux artifacts` (line 307) carries only the OS gate, so the implicit `success()` skips the
rename and upload once this step exits 1 — while the Playwright upload right above it uses
`!cancelled()` for exactly this reason.

`!cancelled() && matrix.os == env.OS_LINUX` matches the existing precedent.

---

## 9. [Low] The new guard isn't registered anywhere durable

**Attach to:** `.erb/scripts/electron-builder-snap-config.test.ts:12`

```
import { GNOME_PLATFORM_BY_BASE, GNOME_PLATFORM_TARGET } from './snap-platform-pairing';
```

No test asserts the CI step exists — a workflow refactor drops it, the input guard stays green, and
the blind spot reopens silently.

The precedent is in this very file: `readWorkflowLinuxRunners()` already greps all three workflows
for `OS_LINUX` on the reasoning that the coupling "was comment-only until now". Two lines fixes it.

Related: both prose pointers still describe half the enforcement — `electron-builder.json5:105` and
`Architecture-Decisions.md:3344` (the ADR's Decision paragraph) name
`electron-builder-snap-config.test.ts` alone.

---

## 10. [Nit] Small batch — take or leave

**Attach to:** `.erb/scripts/assert-generated-snap-metadata.util.ts:23`

```
 * The one snap in an electron-builder output directory, or an explanation of why there is not
```

- This docblock (and `readSnapMetadata`'s at :53) describes a returned explanation, but both throw.
  Worth an `@throws` each, since the entry script's catch makes the contract load-bearing.
- `readSnapMetadata` has no unit tests, though CI integration covers the happy path.
- `OUT_DIR` re-states `directories.output` while the sibling module already parses the config via
  `readConfig()`.
- `afterEach`'s `rmSync(outDir, …)` throws `ERR_INVALID_ARG_TYPE` if `beforeEach` failed, masking
  the real cause (`force` suppresses ENOENT, not an invalid path).
- The `test.yml:233` comment about running before the rename states a constraint that doesn't
  exist — the rename preserves `.snap`, and `resolveSnapArtifact` matches on extension.
- `Readonly<Partial<Record<…>>>` would stop the unknown-base guard reading as dead code.
- `squashfs` / `unsquashfs` aren't in `cspell.json`.

---

## Verified away — don't spend time on these

- **Multi-arch hard-fail.** Unreachable: `target: ['snap']` with no arch, and `extraResources`
  pinned to `linux-x64`.
- **The `assert:` npm namespace.** `verify:`, `clean:`, `ncu:`, `package:`, `csharp:` are all
  single-member — it's the norm, not a deviation.
- **`fail-fast`.** Real, but pre-existing matrix behavior that already applied to lint and
  packaging, not something this PR introduced.

Also checked and clean: `-force` / `-dest` are valid squashfs-tools 4.x options; `meta/snap.yaml`
without a leading slash is standard extract-file syntax; the `existsSync` guard correctly covers
unsquashfs exiting 0 on a non-matching filter; `__dirname` is safe under `tsx`; the checker can't be
bypassed via `apps.<app>.plugs` (a name list only); and a trailing-slash `target` could only cause a
spurious failure, never a false pass.

> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2777#-:-P1216xb7u5N8VqKztfO:b20dykl)* status: all files reviewed, 1 unresolved discussion (waiting on irahopkinson).

___
*[`.github/workflows/test.yml` line 234 at r1](https://reviewable.io/reviews/paranext/paranext-core/2777#-P121CsCAvwL7svS1dQR:-P121CsCAvwL7svS1dQS:b-4j2mtp) ([raw file](https://github.com/paranext/paranext-core/blob/48e6fa89ed05c272dfd0b54145f2b56472b91f35/.github/workflows/test.yml#L234)):*
> ```YAML
>       # that builds a snap, and it runs before the rename-for-upload step below so the artifact is
>       # still under its built name.
>       - name: check generated snap metadata
> ```

## 1. [High] The guard doesn't cover the snap that reaches users

**Attach to:** `.github/workflows/test.yml:234`

```
      - name: check generated snap metadata
```

The guard is wired into `test.yml` only. `publish.yml:204-205` is what runs
`snapcraft upload --release=edge`, and it builds unguarded.

Note the shape of the fix: that workflow builds and uploads inside one `run:` block, so the assert
has to go *between* `npx electron-builder --publish never --linux` and the upload line, not as a
following step.

`publish.yml` is `workflow_dispatch`-only on arbitrary refs with its own snapcraft/lxd stack
(lines 94-111), so it can genuinely diverge from what `test.yml` validated. `package-main.yml:140`
is also unguarded but far lower risk — same branch triggers as `test.yml`, so the same commit gets
checked concurrently. The sibling guard already enumerates all three in `LINUX_RUNNER_WORKFLOWS`.

irahopkinson (commented):
All ten addressed in `63d1740f044`, each confirmed against the source before acting. Thank you for the batch — three of these are defects in my own verification, which is the hardest kind to find from outside.

## 1, 9 — coverage, and asserting the coverage

Confirmed, including the detail that mattered most: build and upload sat in one `run:` block, and the `mv release/build/*.snap release/staged` right after meant a following step would not have found the snap at all.

`package-main.yml` and `publish.yml` now both invoke the check, in the same block as the build. Your closing pointer at `LINUX_RUNNER_WORKFLOWS` was the more valuable half, so #9 is done the way you framed it: the coverage is asserted over that same list, and a snap-building workflow that skips the check now fails the guard. Both prose references name both halves of the enforcement now.

**One correction, because it changes the reasoning rather than the outcome.** #2654 merged a few hours before you posted and removes `snapcraft upload --release=edge`, its credentials env, and the Linux `release/staged` step. Its own comment in `publish.yml` now reads:

> Internal Linux builds come from `package-main.yml`, which uploads a per-platform artifact on every push.

So `publish.yml`'s snap now goes nowhere — excluded from S3, no staging — and the one people actually install is `package-main.yml`'s, which you ranked lower. Your finding stands entirely; the risk ordering inside it has inverted. Worth recording because the guard's value is no longer "protects the store upload", it is "no snap this repo builds goes unexamined".

## 2 — type-checked by nothing

Confirmed exactly, and this one stung: the branch's "typecheck exit 0" evidence covered none of the code it added.

Fixed one level up from your suggestion. Adding the three files would have closed this instance; `.erb/scripts` is now covered by glob instead, so a script added later is checked without anyone remembering to list it — verified by planting a type error in a new file and watching the gate fail. Coverage goes 26 files to 30.

Two exclusions remain, and the header's stated reason for the old allowlist turned out to be wrong, so I rewrote it. It claimed the other scripts "carry pre-existing type issues from their own heavy deps". They don't — `bump-versions.ts` and `check-build-exists.ts` are clean, and the six errors are entirely in `node_modules`: five in `path-scurry`'s bundled `lru-cache`, one in `tinyglobby`'s `fdir`, reached through `../configs/webpack.paths` because `skipLibCheck` is off. Enabling `skipLibCheck` would cover them by weakening the gate for the notices pipeline it mainly exists to protect, so they stay excluded with that written down.

Your point about `lint` buying nothing here is what made the glob worth doing rather than the three-line fix: with no compiler diagnostics from ESLint, an uncovered file is silent, so coverage has to be structural rather than remembered.

## 3 — `content`

Confirmed and reachable exactly as you argued: the template declares no `content`, so today's pass is by default; `snap.plugs` already uses the attribute form for `dot-platform-bible`.

Implemented slightly differently from your suggestion. `descriptor.content ?? key` compared against `expected` made a stale plug name report **twice** — once as a name problem, once as a content problem — for one cause. Comparing against the plug's own key instead mirrors the `default-provider` check beside it and keeps one cause to one message; the key is separately asserted against `expected`, so the coverage is identical.

## 4, 7 — two of my tests that could not fail

Both confirmed by reproducing your mutation. `mkdtempSync` appends six random characters, so `staging.snap-h4YdqS` never reached the `isFile()` branch it was named for; `mkdirSync` fixes it and deleting `isFile()` now reddens. And `reports metadata that is not a YAML mapping rather than throwing` used a valid YAML scalar, so it never touched the throwing path — the parse is now wrapped and reported like every other problem, with a case that actually exercises it.

I had mutation-tested that file five ways and missed both. Noted.

## 6 — the reasoning, which we were each half right about

Your source reading is correct: `snap.js` takes a raw path from `createStageDirPath` and registers no cleanup, unlike `AppxTarget`. And electron-builder does not write `meta/snap.yaml` for this configuration — snapcraft does, which my own first commit message said while the comment contradicted it.

The observation that sent me there is also real, though: a completed CI packaging run leaves `release/build` holding only `linux-unpacked/`, `.icon-set/`, `builder-debug.yml` and the snap. So both hold — electron-builder does not delete the stage directory, and something below it does. The comment now leads with the reason you identified as the durable one, and marks the disappearance as observed rather than explained instead of asserting a mechanism I cannot source.

## 5 — comment density

Taken. The `snap.js`-is-invisible rationale appeared in five places; the entry script keeps it and the rest point there. `snap-platform-pairing.ts` is down from 17 comment lines around 7 constants to 9.

For the record on why this is allowed rather than a matter of taste: `Code-Style-Guide.md` scopes "TSDoc must stand on its own — assume readers cannot look up anything outside the comment" to the API surface, and `.erb/scripts/` is explicitly not that. Had this been `platform-bible-utils`, the duplication would have been correct.

Both failure messages now lead with the actionable half.

## 8 — and it applied to a second workflow

Confirmed. Also applied to `package-main.yml`, which had the same rename-and-upload pair on the same implicit `success()` — and matters more there, since that is the artifact people install. `publish.yml` needs no equivalent: it neither renames nor uploads a snap.

## 10 — all taken

`@throws` on the two functions whose contract is to throw; `directories.output` read from the config rather than restated; guarded `afterEach`; `Partial` so the unknown-base branch stops reading as dead code; the `test.yml` comment's non-existent rename constraint removed (and the same false claim in the "no snap artifact" message); `squashfs`/`unsquashfs` in cspell. `readSnapMetadata` still has no unit tests — it is `unsquashfs` plumbing whose happy path CI covers, and I did not think a fixture squashfs earned its keep. Say if you disagree.

Also noted the three you verified away, and did not spend time on them.

> @irahopkinson+AI made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2777#-:-P17pvI3Dmc7ScOVeaqQ:bkwrygd)* status: 0 of 13 files reviewed, 1 unresolved discussion (waiting on irahopkinson and katherinejensen00).

katherinejensen00 (approved):
A few new findings but nothing blocking so going ahead and approving:
W1 — the #8 fix introduced a false-red path. Three reviewers converged; I confirmed it directly. The lint/format steps (test.yml:289–301) run before check packaging (304). A lint failure now skips packaging, leaves no snap, but the rename step still runs under !cancelled() — the glob doesn't expand, and mv fails with cannot stat './release/build/*.snap'. A lint break now reports a second red step blaming packaging. That's precisely the "false red on the Linux leg" the body names as the worst case. Fix is shopt -s nullglob, not a narrower gate.

W2 — three tests still don't pin the branch they name. A 26-mutation audit found 4 survivors; three are the exact class you caught twice in round 1. :188 stays green with if (!isRecord(metadata)) replaced by if (false) — the string falls through to the missing-base branch and also returns one problem. :181's toContain('base') and :172's toContain('core99') are both satisfied by a second branch. Three of the four early-return branches are deletable without reddening the suite. Every other case in the file asserts on the message; these three are the exceptions.

W3 — one backward-facing comment, tsconfig.erb.json:9 ("An allowlist here previously left three new files unchecked"). Two reviewers flagged it. It's the only such site in the diff — everything else is correctly forward-facing.

Plus three judgment calls that are yours, not defects: the tsconfig glob generalization (sound, but wider than your finding asked and it overturns a prior PR's written rationale), failed snaps uploading under the shared app-linux name, and whether the +2/-1 ADR pointer is enough.

> @katherinejensen00 reviewed 13 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2777#-:-P1GniXMDjm4R7DpgaoH:bstngsx)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

irahopkinson (commented):
Thanks for approving, and for finding W1 — that one was mine and it mattered. All three are fixed in `8dc36eb5adb`, and the branch is rebased onto current main, which now includes #2745, #2759 and your #2797.

## W1 — confirmed, and it was a regression I introduced

Reproduced before fixing. Empty `release/build`, `bash -e`, and the loop dies exactly as you said:

```
mv: cannot stat './release/build/*.snap': No such file or directory   (exit 1)
```

Your step-ordering read is exact — lint and format at 289–301, packaging at 304. So my `!cancelled()` gate turned a lint failure into a second red step blaming packaging, which is precisely the "false red on the Linux leg" the PR body calls the worst case. Fixing #8 manufactured the thing #8's risk note warned about.

`shopt -s nullglob` as you suggested, in `test.yml` and `package-main.yml`; `publish.yml` has no rename step. Verified the fix is a no-op on an empty directory and still renames correctly when a snap is present.

## W2 — confirmed by reproducing your mutation

17/17 green with `!isRecord(metadata)` deleted, exactly as you found. And your reasoning holds for the other two: with the unknown-base branch gone, `expected` is `undefined` and the name-mismatch message still contains `core99`, so `toContain('core99')` passes on a branch that no longer exists.

All now assert wording unique to their own branch, and each branch reddens when removed.

They have also moved into a named sub-describe — `early returns, each asserting wording unique to its own branch` — with the requirement explained inside it and phrased for the next person rather than the last one: when adding a case here, check it reddens with its branch removed. This is the third time on this branch that one shape has got past me — an assertion satisfied by more than the code path it names — and a comment loose above a run of tests is easy to add a fifth underneath.

I did consider enforcing it structurally, with a helper along the lines of `expectSingleProblem(problems, /pattern/)`, and rejected it: a helper can require that you assert on the message, but it has no way to check the wording is unique to the branch, which is the actual requirement. It would have looked like a guarantee while leaving the real gap open — the same failure shape we are fixing.

## W3 — yours, and it was the only one

Cut. The comment now says why the glob is right going forward rather than what an allowlist did during this PR.

Sweeping for the same problem turned up a second defect of a different kind in that file: the exclusions paragraph opened "The two exclusions...", three lines above telling the reader to add a third if their new script imports `webpack.paths`. Already self-contradictory rather than merely stale-in-waiting. Reworded, along with the `five errors / one error` tallies, which would have drifted silently on any dependency bump.

## The three judgment calls

**Artifact naming — you were right, and I have treated it as a defect rather than a judgment call.** Keeping a failed build's snap only helps if you can tell it apart from a good one, and both uploads ran under one name — in `package-main.yml` that is the artifact installed internally. The name now carries the outcome: `app-linux-test-failed-check` in `test.yml`, where the check is its own step, and `app-linux-failed` in `package-main.yml`, where the step builds and checks together and can fail before a snap exists at all.

**The tsconfig glob stays.** Taking the point that it is wider than the finding asked and overturns a prior PR's written rationale. The rationale was wrong on the facts — those scripts are clean, the errors are all in `node_modules` via `webpack.paths` — and an allowlist that fails silently is what produced your finding in the first place. Keeping it rather than reverting to the three-line version, with the reasoning written into the header for whoever reads it next.

**The ADR pointer — I would rather leave it at the pointer, and I do not think this wants its own entry.**

The log records decisions taken among alternatives, with consequences. "Assert on the artifact as well as the inputs" has no alternative that was weighed and rejected; it is just the correct thing, arrived at after the first guard stopped short. Writing it up as a decision would give a 70-entry log another row whose Alternatives section says "don't do the check", which makes the entries recording real trade-offs harder to find.

One thing worth putting on the record against my own position, though. This is not a single solution being generalised early — there are already two independent instances in the repo. #2654's notices pipeline verifies against Webpack's emitted module manifests ("what actually compiled into a bundle") rather than against declared dependencies, and it runs after the production extension build for exactly that reason. Different author, same shape, reached separately. And this PR exists because #2747's guard stopped at the inputs and you had to find the gap.

So if anything deserves recording, it is narrower and more factual than an architecture decision: config-driven build steps have a gap between what is declared and what ships, and that gap needs its own check. That reads as a `.claude/rules/` entry scoped to packaging and CI config, citing both instances — the narrower scope `where-to-add-guidance.md` asks for. I would rather it rode a future packaging change than became a PR of its own, since nothing is queued that would repeat the mistake.

Happy to be argued out of that if you think the log is the right home after all.

## Rebase note

Your #2797 conflicted where you would expect: it added `SNAP_TOOLS_STEP_NAME` and the identical-steps assertion to `electron-builder-snap-config.test.ts`, while this branch moved `GNOME_PLATFORM_TARGET` out to the shared module. Both kept. Worth flagging that your `readSnapToolsSteps` parses the workflow YAML while my `workflowInvokesSnapMetadataCheck` matches text — now inconsistent within one file, and yours is the better of the two. Happy to convert mine if you want them aligned.

#2745 also reworked `tsconfig.erb.json`; its description of what the gate covers is kept, with the snap check added to the list and the glob retained. The glob picked up `stage-dev-packages.ts` and `postinstall.ts` without anyone listing them, which is the behaviour it exists for.

> @irahopkinson+AI made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2777#-:-P1OUzWyBShRhRVyoOsv:b-zo8h0)* status: 5 of 13 files reviewed, all discussions resolved (waiting on katherinejensen00).

katherinejensen00 (approved):
:lgtm: Thanks for making those changes, Ira! Looks good.

> @katherinejensen00 reviewed 8 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2777#-:-P1VQUzM1MTODwYLPIu-:bcxhykc)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2257, opened 2026-05-11
Update some vestigial Paranext references to Platform.Bible

imnasnainaec:
The purpose of this pr is ~to give the Electron app a name, so that (on Windows) the cache will be located in `AppData/Roaming/Platform.Bible/` (instead of `AppData/Roaming/Electron/`, alongside any other unnamed Electron apps).~

~To that end, only the changes in `e2e-tests/global-setup.ts` and `package.json` are material.~

~The rest are just~ cleaning up some residual occurrences of "Paranext".

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2257)

PR #2199, opened 2026-04-14
fix: add explicit name to package.json

merchako:
## Summary

- Adds `"name": "paranext-core"` to the root `package.json`
- Regenerates `package-lock.json` with the correct name

## Problem

The root `package.json` had no `name` field. npm falls back to the containing directory name when this field is absent, so anyone whose clone isn't in a folder named `paranext-core"` (e.g. a folder named `core`) will produce a `package-lock.json` with that directory name after running `npm install`. This creates spurious diffs that can accidentally get committed.

## Test plan

- [ ] Run `npm install` in the repo — `package-lock.json` should show `"name": "paranext-core"` regardless of directory name

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

<a href="https://app.devin.ai/review/paranext/paranext-core/pull/2199" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open with Devin
  </picture>
</a>

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2199)

PR #2778, opened 2026-09-07
Document why root package.json has no name, and guard the lockfile

irahopkinson:
# Document why root package.json has no name, and guard the lockfile

The root `package.json` deliberately carries no `name` and no `version` — app identity lives in `release/app/package.json`, and electron-builder resolves it from both files in ways neither makes obvious. That absence is load-bearing but was recorded nowhere, so it has been independently re-proposed several times, each attempt rediscovering the reasoning from scratch.

This branch records the decision and guards its most visible symptom. It adds `adr-root-package-json-no-name` to the decision log, a path-scoped `.claude/rules/` entry so agents meet the rule when editing the relevant manifests, and a `.husky/pre-commit` check that blocks committing a `package-lock.json` whose root `name` has drifted. It deliberately does **not** touch the `name` field itself — it guards the symptom and leaves the risky field alone.

**3 files, +145 / −0. Two commits:** the original entry, then the fixes from Katherine's review.

## What a reviewer should look at

- [ ] **Is `import/no-relative-packages` the right thing to lead with?** The entry now names it as the prerequisite, ahead of the electron-builder matrix, on the grounds that it is precise, cheap, reproducible, and what actually failed CI on #2257 r3 and #2714. Reproduced independently here: 12 errors across 9 files. The judgement call is ordering it first — the electron-builder exposure is still real, just unverified.
- [ ] **The staged-only scope of the guard.** It fires only when `package-lock.json` is staged. That fixes a real false negative, but drift sitting in the working tree is invisible until you stage the file. Confirm that is the behaviour you want.
- [ ] **Whether the macOS claim should stay.** Recorded as unverified recollection; jolierabideau was asked on both PRs and never replied, and #2257's macOS job was cancelled rather than run. It is one of two reasons the field stays unchanged, so it may be worth chasing confirmation or letting the entry stand.
- [ ] **PR #2199 is still open and orphaned.** Its author asked to hand it off. This branch documents the decision but does not close that PR.

## Review round 2 — Katherine's findings

All five verified independently before acting on them; all five addressed in `c0dfc15bf8f`.

| # | Finding | Resolution |
| --- | --- | --- |
| 1 | "Not reproduced / no corroboration" overstated the record | Split: macOS report stays unverified, but naming the root has broken CI twice for a diagnosed reason |
| 2 | `import/no-relative-packages` is the real blocker, and was absent | Now the lead prerequisite in `Context`, `Alternatives` and the revisit conditions |
| 3 | The third attempt was August 2026, not September | Corrected; only #2714's merge landed in September |
| 4 | Rule file read as though the hook enforces the rule | Now states it guards the symptom, and names CI as what actually stops a manifest `name` |
| 5 | Malformed staged JSON misdiagnosed as a name problem | Distinct message for unreadable/absent root name; `2>/dev/null` kept so no V8 trace buries it |

Reproduced for finding 2 rather than taken on trust: adding `"name": "paranext-core"` to the root produces the errors; removing it, the same files are clean. Confirmed her count of 12 exactly, under the repo's own lint scope.

Deliberately **not** changed, per her measurement against the file's own house style: entry length (50 lines vs a median of 47), the unbroken `Context` prose (the format of 50/61 entries), the test matrix as prose rather than a checklist (tick-boxes would read as a to-do list rather than a bar to clear), and the six-line `Source:` field (the only place recording #2257 as *unrelated* work).

<details>
<summary>Round 1 — findings from the pre-PR review, all resolved before opening</summary>

### Important

- [x] **The guard read the working-tree lockfile rather than the staged blob**, producing both a false negative (stage a drifted lockfile → fix the working tree → forget to re-stage → the hook reads the fixed file and passes) and a false positive (every commit blocked while the working tree is drifted). With CLAUDE.md forbidding `--no-verify`, there was no escape hatch. _(Fixed: gated on `git diff --cached`, reads `git show :package-lock.json`. Both modes re-tested.)_

### Minor

- [x] **No runnable fix in the error message** _(Fixed: both recovery paths given.)_
- [x] **Format-coupled `sed` extraction** _(Fixed: node reading the staged blob from stdin.)_
- [x] **Empty-string edge case on a missing key** _(Fixed via the node switch; refined again in round 2.)_
- [x] **ADR Decision field described the superseded behaviour** _(Fixed.)_
- [x] **Rule file restated the ADR's rationale** _(Fixed: it defers to the ADR.)_
- [ ] ~~Hook-only, no CI backstop~~ _(Accepted: both documents state CI cannot catch this. Round 2 established that CI **does** catch the more important case — a manifest `name` — via `import/no-relative-packages`.)_
- [ ] ~~No test covers the guard~~ _(Accepted: no test infrastructure exists for `.husky/` hooks. All paths verified by hand.)_
- [ ] ~~ADR link has no `#slug` fragment~~ _(Accepted: matches repo convention — and round 2 confirmed GitHub slugifies the whole heading, so the fragment would not resolve anyway.)_

</details>

<details>
<summary>Verification</summary>

- `sh -n .husky/pre-commit` — exit 0 (husky runs hooks with `sh`).
- `npx prettier --check` on both Markdown files — exit 0.
- Guard behaviour verified end-to-end after every edit: staged drift blocks; unstaged drift is ignored; `git commit -a` still catches drift because it stages first; malformed staged JSON now reports an unreadable name rather than drift; a correct name passes.
- ADR slug ordering re-verified after the rebase: `LC_ALL=C sort -c` over all `^## adr-` headings passes.
- **Paratext 10 Studio compatibility**, re-checked against the final hook rather than the initial version. Its `paranext-core.patch` hunk anchors on lines 1–5, and `git apply --3way --check` exits 0 over this change. Its inserted `exit 1` sits at line 4, so this guard is unreachable inside `temp-build/paranext-core` by construction, and the build never commits inside the clone anyway (`git apply`, `git add`, `git diff`, `git reset --hard` — none fire `pre-commit`). The clone directory is literally `paranext-core`, so npm's fallback could not drift there either.

Not run: `npm run typecheck`, `npm run lint`, `npm test`. The diff is two Markdown files and one extensionless shell file — eslint runs `--ext .cjs,.js,.jsx,.ts,.tsx`, stylelint globs only CSS/SCSS, and there is no TypeScript or test code here. CI runs the full set regardless.

</details>

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2778)

### DIFF
--- a/.claude/rules/root-package-json-no-name.md
+++ b/.claude/rules/root-package-json-no-name.md
@@ -1,0 +1,39 @@
+---
+paths:
+  - 'package.json'
+  - 'package-lock.json'
+  - 'release/app/package.json'
+---
+
+## Never Add `name` or `version` to the Root `package.json`
+
+The root `package.json` deliberately carries no `name` and no `version`. The app's identity lives in
+`release/app/package.json` (`name: "platform-bible"`), and electron-builder resolves it from both
+files in ways that are not obvious from either.
+
+**Do not add these fields.** The full reasoning, the alternatives, and the manual test matrix any
+future attempt must clear are in
+[`adr-root-package-json-no-name`](../../.context/standards/Architecture-Decisions.md) — read that
+before proposing a change, rather than re-deriving it.
+
+### The symptom you are probably reacting to
+
+Because npm falls back to the containing directory name when `name` is absent, running
+`npm install` from a clone or git worktree **not** named `paranext-core` rewrites the root `name` in
+`package-lock.json`:
+
+```diff
+-  "name": "paranext-core",
++  "name": "my-branch-worktree",
+```
+
+**Fix the lockfile, not the manifest.** Restore the line to `paranext-core`. A check in
+`.husky/pre-commit` blocks the commit if you miss it, and the wrong name is otherwise
+self-consistent — CI's "Verify no files changed after build" will not catch it.
+
+### Why this rule exists
+
+Adding the field is a reasonable-looking one-line change, so it keeps getting proposed — repeatedly
+and independently, including by AI agents, each time rediscovering the reasoning from scratch. If
+you are about to suggest it, you are not the first. The ADR weighs the payoff against the exposure
+and records what a future attempt would have to demonstrate.
--- a/.claude/rules/root-package-json-no-name.md
+++ b/.claude/rules/root-package-json-no-name.md
@@ -31,6 +31,12 @@ Because npm falls back to the containing directory name when `name` is absent, r
 `.husky/pre-commit` blocks the commit if you miss it, and the wrong name is otherwise
 self-consistent — CI's "Verify no files changed after build" will not catch it.
 
+That hook guards the symptom, not this rule: it compares the lockfile's root `name` against
+`paranext-core`, so adding `"name": "paranext-core"` to the manifest makes npm write the expected
+value and the check passes while the decision is being violated. What actually stops that is CI —
+naming the root makes `import/no-relative-packages` reject 12 pre-existing imports across 9 files.
+See the ADR for why that is a prerequisite rather than a nuisance.
+
 ### Why this rule exists
 
 Adding the field is a reasonable-looking one-line change, so it keeps getting proposed — repeatedly
--- a/.context/standards/Architecture-Decisions.md
+++ b/.context/standards/Architecture-Decisions.md
@@ -2545,6 +2545,56 @@ step, no automation. Just a record.
 - **Source:** punctuation-checklist port (markers-consumption verdict); see `08_Checklists.md` in the
   PT9 feature inventory for the per-tool behavior and the verse-range divergence.
 
+## adr-root-package-json-no-name: The root `package.json` carries no `name` or `version`; app identity lives in `release/app/package.json`
+
+- **Date:** 2026-09-08
+- **Status:** Accepted
+- **Context:** The root `package.json` has no `name` and no `version` field. That absence is
+  load-bearing rather than an oversight, but nothing in the repo said so, and it is repeatedly
+  re-proposed — PR #2199, a revision of PR #2257, and again during unrelated dependency work in
+  September 2026 — each time rediscovering the reasoning from scratch.
+  electron-builder resolves app identity from **both** the root manifest and
+  `release/app/package.json` in ways neither file makes obvious: `electron-builder.json5` sets
+  `productName` and `appId` and points `directories.app` at `release/app`, while
+  `release/app/package.json` carries the real identity (`name: "platform-bible"`). Adding
+  `name`/`version` to the root is *recollected* to have broken the macOS build once — raised by
+  tjcouch-sil in review of #2257, who asked jolierabideau to confirm it. Treat that as unverified
+  recollection rather than established fact: it has not been reproduced, and the review carries no
+  corroboration. The visible cost of the absence is that npm falls back to the containing directory
+  name, so `npm install` from a clone or git worktree not named `paranext-core` rewrites the
+  lockfile's root `name`.
+- **Decision:** Leave the root `package.json` without `name` or `version`. Guard the *symptom*
+  instead: `.husky/pre-commit` blocks a commit that stages a `package-lock.json` whose root `name`
+  is not `paranext-core`, alongside the existing yalc-entries check. It reads the staged blob, and
+  only when the lockfile is staged, so an unrelated commit is never blocked and a drifted staged
+  copy is never missed because the working tree was fixed without re-staging.
+- **Alternatives:**
+  - *Add `name` to the root manifest* — deferred, not rejected on merit. PR #2199 proposed
+    `paranext-core` and a revision of PR #2257 proposed `platform-bible`; the two attempts did not
+    agree on the value, which is itself a sign the question is unsettled. The payoff is removing a
+    one-line lockfile annoyance; the exposure is app identity across two products and three
+    operating systems.
+  - *Add it here and in `paratext-10-studio` together* — that coordination would be mandatory, not
+    optional: `lib/build.ts` there rewrites `release/app/package.json` (setting
+    `releaseAppPackage.name` and `.version`) and never touches the root manifest, so a root `name`
+    would survive the Paratext 10 Studio rename unchanged.
+  - *Leave it undocumented* — rejected: the undocumented status quo is what produced the repeated
+    rediscoveries.
+- **Consequences:** Dev runs store data under `%appdata%/Electron` rather than a product-named
+  directory (prod is unaffected); the team treats the resulting dev/prod log separation as a minor
+  benefit. Anyone revisiting this must first clear the bar set in review of #2257 — test dev **and**
+  prod (portable and installed), on Platform.Bible **and** Paratext 10 Studio, on all three
+  operating systems, checking app data paths (`%appdata%/<name>`, `~/<name>`, install dir), deep
+  links and the displayed app name, the output executable name, executable metadata, and signing.
+  Revisit if electron-builder's field resolution becomes documented well enough to retire that
+  matrix, or if someone confirms or refutes the macOS recollection.
+- **Source:** PR #2199 (`fix: add explicit name to package.json`, opened 2026-04-14 by merchako, who
+  later asked to hand it off after the review below), and
+  tjcouch-sil's CHANGES_REQUESTED review of PR #2257 (2026-05-11), which carries the full rationale
+  and the test matrix. #2257 itself is unrelated work ("Update some vestigial Paranext references to
+  Platform.Bible", merged 2026-05-13) that briefly proposed the same root-`name` addition and
+  dropped it.
+
 ## adr-runaway-data-hook-guard: `useData`'s runaway guard counts subscribes and deliveries, degrades rather than throws, and expires
 
 - **Date:** 2026-08-28
--- a/.context/standards/Architecture-Decisions.md
+++ b/.context/standards/Architecture-Decisions.md
@@ -2552,17 +2552,23 @@ step, no automation. Just a record.
 - **Context:** The root `package.json` has no `name` and no `version` field. That absence is
   load-bearing rather than an oversight, but nothing in the repo said so, and it is repeatedly
   re-proposed — PR #2199, a revision of PR #2257, and again during unrelated dependency work in
-  September 2026 — each time rediscovering the reasoning from scratch.
-  electron-builder resolves app identity from **both** the root manifest and
-  `release/app/package.json` in ways neither file makes obvious: `electron-builder.json5` sets
-  `productName` and `appId` and points `directories.app` at `release/app`, while
-  `release/app/package.json` carries the real identity (`name: "platform-bible"`). Adding
-  `name`/`version` to the root is *recollected* to have broken the macOS build once — raised by
-  tjcouch-sil in review of #2257, who asked jolierabideau to confirm it. Treat that as unverified
-  recollection rather than established fact: it has not been reproduced, and the review carries no
-  corroboration. The visible cost of the absence is that npm falls back to the containing directory
-  name, so `npm install` from a clone or git worktree not named `paranext-core` rewrites the
-  lockfile's root `name`.
+  August 2026 (PR #2714, whose own merge landed in September) — each time rediscovering the
+  reasoning from scratch. Naming the root has twice broken CI, for a precise and fully diagnosed
+  reason: it makes eslint-plugin-import treat the root as a package, so `import/no-relative-packages`
+  rejects every pre-existing relative import that crosses out of the root into `release/app`,
+  `.storybook` or the `lib` packages — 12 errors across 9 files, none of them files such a change
+  would touch. That is what failed the Ubuntu build on #2257 revision r3 (2026-05-11) and all three
+  platforms on #2714 (2026-08-28). Separately, electron-builder resolves app identity from **both**
+  the root manifest and `release/app/package.json` in ways neither file makes obvious:
+  `electron-builder.json5` sets `productName` and `appId` and points `directories.app` at
+  `release/app`, while `release/app/package.json` carries the real identity
+  (`name: "platform-bible"`). Adding `name`/`version` to the root is separately *recollected* to have
+  broken the macOS build — raised by tjcouch-sil in review of #2257, who asked jolierabideau to
+  confirm it. That recollection specifically remains unverified: jolierabideau never replied on
+  either PR, and #2257's macOS job was cancelled rather than run, so it neither confirms nor refutes
+  it. The visible cost of the absence is that npm falls back to the containing directory name, so
+  `npm install` from a clone or git worktree not named `paranext-core` rewrites the lockfile's root
+  `name`.
 - **Decision:** Leave the root `package.json` without `name` or `version`. Guard the *symptom*
   instead: `.husky/pre-commit` blocks a commit that stages a `package-lock.json` whose root `name`
   is not `paranext-core`, alongside the existing yalc-entries check. It reads the staged blob, and
@@ -2572,8 +2578,12 @@ step, no automation. Just a record.
   - *Add `name` to the root manifest* — deferred, not rejected on merit. PR #2199 proposed
     `paranext-core` and a revision of PR #2257 proposed `platform-bible`; the two attempts did not
     agree on the value, which is itself a sign the question is unsettled. The payoff is removing a
-    one-line lockfile annoyance; the exposure is app identity across two products and three
-    operating systems.
+    one-line lockfile annoyance. The cost has two parts, and the cheap one is the one that has
+    actually stopped every attempt so far: the `import/no-relative-packages` failure above must be
+    resolved first, by either moving the root into a workspace layout that makes those imports legal
+    or rewriting the 12 offending imports, plus whatever prettier and stylelint fallout follows.
+    Only past that does the second part apply — app identity across two products and three
+    operating systems, which is unverified rather than disproven.
   - *Add it here and in `paratext-10-studio` together* — that coordination would be mandatory, not
     optional: `lib/build.ts` there rewrites `release/app/package.json` (setting
     `releaseAppPackage.name` and `.version`) and never touches the root manifest, so a root `name`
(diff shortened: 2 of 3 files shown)


katherinejensen00 (changes requested):
# Review — PR #2778 · `docs/root-package-json-no-name`

**Repo:** paranext/paranext-core · **Author:** Ira Hopkinson (`irahopkinson`)
**PR:** https://github.com/paranext/paranext-core/pull/2778
**Reviewed:** 2026-09-08 · 3 files, +115 / −0
**Reviewed against:** `origin/main` (merge-base `39eb9a6611c`)

> Note: `/code-review max` was still completing its final verification pass when this report was
> written. Every finding below was verified independently of it.

---

## Summary

The shape of this change is right and the hook logic is sound under every path I tested — but the
ADR omits the one blocker that is concrete, reproduced, and still live
(`import/no-relative-packages`), which means the next person to try adding a root `name` will run
your eight-item cross-platform test matrix and still be stopped in thirty seconds by something the
entry never mentions.

---

## Findings

| # | File | Line | Severity |
|---|---|---|---|
| 1 | `.context/standards/Architecture-Decisions.md` | 2253 | **Important** |
| 2 | `.context/standards/Architecture-Decisions.md` | 2263 | **Important** |
| 3 | `.context/standards/Architecture-Decisions.md` | 2245 | Minor |
| 4 | `.claude/rules/root-package-json-no-name.md` | 30 | Minor |
| 5 | `.husky/pre-commit` | 55 | Minor |

Full paths are under `C:\Users\kathe\paranext-core\.claude\worktrees\code-review-alpha\`.

Findings 1 and 2 are the ones worth holding the merge for; 3–5 are small and independent.

---

### 1 — `Architecture-Decisions.md:2253` · Important

**Attach to the line beginning:** `recollection rather than established fact: it has not been reproduced, and the review carries no`

```
The evidentiary claim here is contradicted by the record. "It has not been reproduced, and the
review carries no corroboration" is true of tjcouch-sil's review body read in isolation, but not
of the thread or of what followed:

- On #2257 itself, revision r3 (4381579266c9) added "name": "platform-bible" to the root manifest
  and the Ubuntu build failed (2026-05-11T19:41:08Z) with three "Relative import from another
  package is not allowed" errors. imnasnainaec flagged it in the very thread this entry cites --
  "(BTW, the Ubuntu build check failed on this branch before I reverted.)" -- and you replied
  "Sorry to be a parade-rainer :p thanks for reverting". Note macOS was cancelled, not passed, so
  #2257 neither confirms nor refutes the macOS recollection specifically.
- It was reproduced independently on #2714 (2026-08-28), breaking CI on all three platforms.

The macOS half of the hedge is exactly right and worth keeping -- jolierabideau never replied on
either PR. It's the blanket negative about the record that overstates. Suggest splitting the two:
the macOS breakage is unconfirmed recollection; adding the root name has nonetheless broken CI
twice for a different, fully diagnosed reason.
```

---

### 2 — `Architecture-Decisions.md:2263` · Important

**Attach to the line beginning:** `- *Add \`name\` to the root manifest* — deferred, not rejected on merit. PR #2199 proposed`

```
This is the finding I'd most want addressed. "Deferred, not rejected on merit ... the payoff is
removing a one-line lockfile annoyance; the exposure is app identity across two products and three
operating systems" frames the blocker as unknown-and-expensive. There is a known-and-cheap one, and
it isn't electron-builder at all:

Naming the root makes eslint-plugin-import treat it as a package, and import/no-relative-packages
then rejects every pre-existing relative import crossing out of the root into release/app,
lib/platform-bible-react, or extensions -- 12 errors in files the change never touches. The rule is
'error' via airbnb (eslint-config-airbnb-base/rules/imports.js), and I confirmed nothing in
.eslintrc.js overrides it today. You diagnosed this yourself on 2026-08-31.

Because it's absent from both Alternatives and the revisit conditions at line 2280, a future
attempt reads this entry, works the eight-item matrix across two products and three OSes, and then
hits a lint failure the entry never warned about. That's the re-derivation this ADR exists to stop.

Worth stating as the prerequisite, since unlike the electron-builder matrix it's precise and cheap:
naming the root needs either a workspace-layout change plus 12 relative-import rewrites, or an
import/no-relative-packages config change.

Live evidence that the omission bites: while researching this PR, one of my own analysis passes read
the ADR and independently recommended "just add name to package.json to fix it at the source."
```

---

### 3 — `Architecture-Decisions.md:2245` · Minor

**Attach to the line beginning:** `re-proposed — PR #2199, a revision of PR #2257, and again during unrelated dependency work in`

```
"September 2026" is off by a month. The episode is real and is exactly unrelated dependency work --
PR #2714, PT-4412 -- but it ran 2026-08-25 (proposed) -> 2026-08-25 (commit a8f5f768685) ->
2026-08-28 (reverted, CI red on all three platforms) -> 2026-08-31 ("Don't file a ticket for it --
let's leave it"). Only the merge lands in September. A deterministic sweep found no separate
September event.

Flagging it because docs-durability.md's "date every status claim" only pays off if the dates are
right -- a wrong one sends the next reader looking in the wrong month for the thread that has the
diagnosis.
```

---

### 4 — `.claude/rules/root-package-json-no-name.md:30` · Minor

**Attach to the line beginning:** `**Fix the lockfile, not the manifest.** Restore the line to \`paranext-core\`. A check in`

```
"A check in .husky/pre-commit blocks the commit if you miss it" reads as though the hook enforces
this rule. It enforces the symptom, not the rule -- and the gap runs the wrong way:

If someone adds "name": "paranext-core" to the root package.json -- precisely what #2199 proposed --
npm writes that name into the lockfile root, the guard sees the expected value, and it passes
forever while the decision it documents is being violated. The version half is unguarded entirely.

The ADR is honest about this ("Guard the *symptom* instead"), but this file is the one a person
actually reads before making the change, and here it reads like protection it doesn't provide. A
half-sentence would close it -- something like "blocks the commit if you miss it (it catches the
drifted lockfile, not a name added to the manifest)".

Worth noting for whoever asks "then what does enforce it": CI does, via the
import/no-relative-packages failure in finding 2. That's a further reason to name that mechanism in
the ADR.
```

---

### 5 — `.husky/pre-commit:55` · Minor

**Attach to the line beginning:** `LOCK_ROOT_NAME=$(git show :package-lock.json | node -p ...`

```
This conflates "couldn't read the name" with "the name is wrong", and the error message then
confidently misdiagnoses the second.

Reproduced in a scratch repo: stage a package-lock.json still carrying conflict markers -- routine
here during a rebase -- and node exits non-zero, 2>/dev/null swallows it, LOCK_ROOT_NAME is empty,
and the hook prints:

  Error: package-lock.json's root "name" is "", expected "paranext-core".

...then tells you to edit a name that isn't the problem. A missing name key gives the sibling case,
is "undefined". Both block the commit with instructions that don't apply.

To be clear on two things I checked so this doesn't get over-fixed: keep the 2>/dev/null -- removing
it dumps 16 lines of V8 stack trace over your own error message. And the node-on-PATH concern from
your review-focus list is close to unreachable, since husky is installed by npm in the first place.
The reachable trigger is malformed staged JSON.

A two-line guard covers both, before the comparison:

  if [ -z "$LOCK_ROOT_NAME" ] || [ "$LOCK_ROOT_NAME" = "undefined" ]; then
    echo "Error: could not read the root \"name\" from the staged package-lock.json."
    echo "Is it valid JSON? (Unresolved conflict markers will do this.)"
    exit 1
  fi
```

---

## Documentation assessment

The prose earns its length. Four length/structure objections were raised and all four dropped once
measured against the file's own house style:

- **ADR entry length** — 50 lines against a median of 47 across the 61 entries in the file. 35th of
  61. Dead centre.
- **The 14-line `Context` paragraph** — ties for 5th–7th longest, with three entries longer.
  Unbroken prose is the format of 50/61 entries and matches the file's own template.
- **"Make the test matrix a bulleted list"** — only 11/61 entries use nested sub-lists, and this
  entry is already one of them (in `Alternatives`). The device was known and prose was chosen. There
  is also a real argument for it: the matrix is a bar to clear, and eight tick-boxes would read as a
  to-do list someone can chip away at.
- **The 7-line hook comment** — sibling `.husky/lib/ai-hooks.sh` runs 19% comment density with a
  10-line block. And the comment explains why the check reads the index rather than the working
  tree, which is exactly the "surprising approach" `forward-facing-comments.md` says to keep.

Two genuine notes, neither worth a line comment:

- The **`Source:` field is 6 lines** against a mode of 1 — joint second-longest in the file, and
  #2257 appears 6 times across 4 fields. But the repetition earns its place: that field is the only
  spot saying #2257 *is unrelated work*, without which a reader who has seen it cited four times
  goes and reads the wrong diff. Leave it.
- The **PR description is 209 lines / 13.7 KB for a 115-line diff**, with the most useful section —
  "Suggested Review Focus" — last. Not a code comment; just worth knowing that a reviewer arriving
  cold scrolls past ten screens of resolved in-review findings to reach the four questions actually
  addressed to them.

---

## Checked and cleared

Recorded so nobody re-litigates them:

- **Paratext 10 Studio is safe**, verified by replay rather than inspection. Its patch hunk anchors
  on lines 1–5 only; running P10S's exact command (`git apply --3way --allow-empty
  --ignore-whitespace`) against the modified hook applied cleanly, exit 0. The checkout dir is
  literally `paranext-core`, it installs with `npm ci`, and the patched hook `exit 1`s at line 4
  anyway. `save-repo-patches.ts`'s `@@ -1` assertion still holds.
- **"CI will not catch it" is correct.** `npm ci` in a wrongly-named directory left the lockfile
  byte-identical and exited 0, so `verify-changed-files` stays green. Every workflow uses `npm ci`;
  none uses `npm install`.
- **Hook logic is sound on all four paths tested**: staged drift blocks; unstaged drift no longer
  blocks an unrelated commit; `git commit -a` stages before the hook fires so drift is still caught;
  and a failed `-a` leaves the real index untouched.
- **`grep -qx` scoping is right** — it deliberately excludes `release/app/package-lock.json`, which
  cannot drift because that manifest has an explicit `name`.
- **Single drift site confirmed**: `packages[""]` carries no `name` key, so the top-level `name` is
  the only place npm writes the directory fallback. The guard's single check is sufficient, and "the
  second line of the file" is accurate today.
- **Byte-order slug position verified** (`LC_ALL=C sort -c` → clean), between
  `adr-reuse-shared-checklist-framework` and `adr-runaway-data-hook-guard`.
- **The missing anchor on the ADR link is correct as-is** — GitHub slugifies the whole heading, so
  `#adr-root-package-json-no-name` would not resolve anyway.
- **`paths:` frontmatter is right.** "It should list `.husky/pre-commit`" was nearly filed — zero of
  the 16 path-scoped rules in `.claude/rules/` list an enforcing tool file, so that would have
  invented a convention and made this file its only adherent.
- **Drift is live**: two worktrees on this machine already carry drifted lockfile names, so the
  guard will start earning its keep immediately.

---

## Method

Four independent passes, with an adversarial pass run against the first set of findings:

1. Direct read of the full diff, plus empirical testing of the hook in scratch repositories
   (staged drift, unstaged drift, `git commit -a`, missing `name` key, absent `node`, malformed
   JSON).
2. Fact-check of every externally checkable claim in the ADR against GitHub (#2199, #2257, #2714,
   review bodies, CI run outcomes) and against the live repos.
3. Downstream-impact research: Paratext 10 Studio patch compatibility, fork/rename risk, node
   availability in husky, and whether CI can catch lockfile drift.
4. Adversarial review of passes 1–3, which dropped 6 of 8 initial findings as style preferences
   contradicted by measured house style, and surfaced finding 4.

No changes were made to the working tree, and nothing was posted to the PR.

> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2778#-:-P12M2Sr1kIX_4HW3jn9:b-3kua8r)* status: all files reviewed, 1 unresolved discussion (waiting on irahopkinson).

___
*[`.context/standards/Architecture-Decisions.md` line 2239 at r1](https://reviewable.io/reviews/paranext/paranext-core/2778#-P12M6EA5zZ_TyVUW9Tc:-P12M6EA5zZ_TyVUW9Td:b-n3y1ob) ([raw file](https://github.com/paranext/paranext-core/blob/2b86c3ece67a541b24bc1930b56fa8623b2f6f45/.context/standards/Architecture-Decisions.md#L2239)):*
> ```Markdown
>   PT9 feature inventory for the per-tool behavior and the verse-range divergence.
> 
> ## adr-root-package-json-no-name: The root `package.json` carries no `name` or `version`; app identity lives in `release/app/package.json`
> ```

### 1 — `Architecture-Decisions.md:2253` · Important

**Attach to the line beginning:** `recollection rather than established fact: it has not been reproduced, and the review carries no`

```
The evidentiary claim here is contradicted by the record. "It has not been reproduced, and the
review carries no corroboration" is true of tjcouch-sil's review body read in isolation, but not
of the thread or of what followed:

- On #2257 itself, revision r3 (4381579266c9) added "name": "platform-bible" to the root manifest
  and the Ubuntu build failed (2026-05-11T19:41:08Z) with three "Relative import from another
  package is not allowed" errors. imnasnainaec flagged it in the very thread this entry cites --
  "(BTW, the Ubuntu build check failed on this branch before I reverted.)" -- and you replied
  "Sorry to be a parade-rainer :p thanks for reverting". Note macOS was cancelled, not passed, so
  #2257 neither confirms nor refutes the macOS recollection specifically.
- It was reproduced independently on #2714 (2026-08-28), breaking CI on all three platforms.

The macOS half of the hedge is exactly right and worth keeping -- jolierabideau never replied on
either PR. It's the blanket negative about the record that overstates. Suggest splitting the two:
the macOS breakage is unconfirmed recollection; adding the root name has nonetheless broken CI
twice for a different, fully diagnosed reason.
```

irahopkinson (commented):
All five verified independently before acting on them, and all five hold. Fixes are in `c0dfc15bf8f`, kept as a separate commit rather than an amend so the r1→r2 diff shows exactly what moved.

## Finding 2 — reproduced, and it's now the lead

This was the right thing to hold the merge for. I reproduced it rather than taking it on trust: adding `"name": "paranext-core"` to the root produces `import/no-relative-packages` errors immediately; removing it, the same files are clean. Under the repo's own lint scope I get **12 errors across 9 files** — your count exactly:

```
3  .storybook/preview.ts
2  src/main/services/app.service-host.ts
1  .erb/configs/webpack.config.base.ts
1  .erb/scripts/bump-versions.ts
1  .erb/scripts/electron-rebuild.js
1  .storybook/manager.tsx
1  src/renderer/components/dialogs/about-dialog.component.tsx
1  src/stories/platform/ten-layout-shared.tsx
1  tailwind.config.ts
```

Worth noting only 3 of 12 are under `.erb` — the rest are authored code, so this isn't something that could be waved off as vendored-boilerplate lint debt.

It now leads `Context`, appears in `Alternatives` as the prerequisite with its resolution cost (workspace-layout change or 12 import rewrites, plus prettier/stylelint fallout), and heads the revisit conditions so an attempt can't work the matrix first and get stopped afterwards.

**One place I'd push back slightly on framing, not substance.** You wrote "it isn't electron-builder at all." I've kept both: the lint rule is the blocker hit first and cheapest, but it doesn't retire the electron-builder exposure, which is unverified rather than disproven. Ordering it first is the judgement call, and I've listed it in the description as a question for you rather than treating it as settled — if you think the electron-builder half should now be demoted to a footnote, say so and I'll cut it down.

## Finding 1 — conceded, and the split is right

The blanket negative overstated. Naming the root has demonstrably broken CI twice; the macOS report specifically has not been corroborated. Those are now two separate statements, with the macOS half keeping the hedge and gaining the detail that #2257's macOS job was **cancelled rather than run**, so it neither confirms nor refutes.

## Finding 3 — confirmed and corrected

You're right and I checked the commit: `a8f5f768685` (2026-08-24, on #2714) added `"name": "paranext-core"` to the root and was reverted. August, with only the merge landing in September. Fixed.

## Finding 4 — the sharpest of the five, and one I'd have missed

The gap running the wrong way is the part that matters: add the name the ADR forbids, npm writes it into the lockfile, the guard sees the expected value and passes **forever**, while the decision it documents is being violated. The rule file now says the hook guards the symptom rather than the rule, and names CI — via finding 2's mechanism — as what actually stops it. That connection between 2 and 4 is a better answer than either finding alone.

## Finding 5 — reproduced and fixed

Staged conflict markers do exactly what you describe: node exits non-zero, the value comes back empty, and the message sends you off to edit a name that isn't the problem. There's now a distinct branch for unreadable-or-absent before the comparison, with its own message naming conflict markers as the likely cause. I kept `2>/dev/null` for the reason you gave — removing it buries the message under a V8 trace — and I didn't touch the node-on-PATH path, since you're right that husky being npm-installed makes it near-unreachable.

## Documentation assessment

Thank you for measuring rather than asserting — and for dropping four objections on the evidence. I've deliberately left all of it alone: the entry length, the unbroken `Context` prose, the test matrix as prose rather than tick-boxes (your to-do-list-versus-bar argument is the right one), and the six-line `Source:` field, since it's the only place recording #2257 as unrelated work.

The one thing I did act on was your closing note about the PR description: 209 lines with the reviewer-facing questions last. It's now 69 lines with those questions near the top and both rounds of resolved findings collapsed. That was a fair hit — a reviewer arriving cold shouldn't scroll past ten screens of my own resolved findings to reach the four things actually addressed to them.

## Also

Rebased onto current `main` (`f1946404af8`, the 0.6.0-alpha.1 version bump) while I was here. And thanks for the P10S replay — running their exact `git apply` invocation rather than inspecting the hunk is a better check than the one I did.

> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2778#-:-P12Vbj4808cOjIcPDT5:brsyc6d)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on katherinejensen00).

___
*[`.context/standards/Architecture-Decisions.md` line 2239 at r1](https://reviewable.io/reviews/paranext/paranext-core/2778#-P12M6EA5zZ_TyVUW9Tc:-P12Vefe1f21vIJF2gKp:b-t0jawl) ([raw file](https://github.com/paranext/paranext-core/blob/2b86c3ece67a541b24bc1930b56fa8623b2f6f45/.context/standards/Architecture-Decisions.md#L2239)):*

### 1 — `Architecture-Decisions.md:2253` · Important

**Attach to the line beginning:** `recollection rather than established fact: it has not been reproduced, and the review carries no`

```
The evidentiary claim here is contradicted by the record. "It has not been reproduced, and the
review carries no corroboration" is true of tjcouch-sil's review body read in isolation, but not
of the thread or of what followed:

- On #2257 itself, revision r3 (4381579266c9) added "name": "platform-bible" to the root manifest
  and the Ubuntu build failed (2026-05-11T19:41:08Z) with three "Relative import from another
  package is not allowed" errors. imnasnainaec flagged it in the very thread this entry cites --
  "(BTW, the Ubuntu build check failed on this branch before I reverted.)" -- and you replied
  "Sorry to be a parade-rainer :p thanks for reverting". Note macOS was cancelled, not passed, so
  #2257 neither confirms nor refutes the macOS recollection specifically.
- It was reproduced independently on #2714 (2026-08-28), breaking CI on all three platforms.

The macOS half of the hedge is exactly right and worth keeping -- jolierabideau never replied on
either PR. It's the blanket negative about the record that overstates. Suggest splitting the two:
the macOS breakage is unconfirmed recollection; adding the root name has nonetheless broken CI
twice for a different, fully diagnosed reason.
```

</blockquote></details>

Conceded — fixed in `c0dfc15bf8f`, and split the way you suggested.

The blanket negative was the problem. The two claims are now separate: naming the root **has** broken CI twice, for the fully diagnosed `import/no-relative-packages` reason (#2257 r3 on Ubuntu, #2714 on all three platforms), and that now leads the `Context` paragraph. The macOS report specifically keeps its hedge, with the added detail you supplied — #2257's macOS job was *cancelled rather than run*, so it neither confirms nor refutes, and jolierabideau never replied on either PR.

I verified the #2714 half rather than taking it on trust: `a8f5f768685` (2026-08-24) did add `"name": "paranext-core"` to the root manifest and was reverted. That also fixes the date in finding 3 — August, not September.

katherinejensen00 (approved):
:lgtm:

> @katherinejensen00 reviewed 3 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2778#-:-P15fkQf8ShFj9wUXJ_8:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

Issue #32, opened 2023-02-15
Create generated documentation of our api


darren8c:
As an extension developer, I want to be able to go somewhere not in the code to find all the features of the API so I can easily navigate around and get a detailed view of what I can do.

We need to find a way to generate an external document describing PAPI and possibly hook it up to our CICD process.

Looks like TypeDoc does what we want:
https://github.com/TypeStrong/typedoc
Examples:
- https://typedoc.org/example/
- https://ticlo.github.io/rc-dock/

[There are others](https://www.reddit.com/r/typescript/comments/vy1ew0/recommend_a_ts_documentation_generator/) in case we want to try multiple.

tjcouch-sil:
@darren8c Fixed when we created `papi-dts`?

tjcouch-sil:
Implemented with `papi.d.ts` in #161

irahopkinson:
@tjcouch-sil are you saying that a `.d.ts` file is sufficient documentation for our API?

tjcouch-sil:
Generating docs from the typescript types are in this issue.

Generating an interactive api is in #444

Issue #22, opened 2023-02-08
Events over WebSocket (Node)


tjcouch-sil:
Create Events and Event Handler API for distribution over WebSocket

irahopkinson:
@tjcouch-sil I'm not sure if you had in mind to do this one. Either way I'm not exactly sure what is required here. I'll look at something else until I hear from you.

What I did find is:

-  that if we used [socket.io](https://socket.io/) we would have build in support for custom events (as well as them being reconnecting without us have to write that ourselves).
- also [primus-emitter](https://www.npmjs.com/package/primus-emitter) which is old even though [primus](https://www.npmjs.com/package/primus) itself is still recent.

tjcouch-sil:
@irahopkinson My internet went out all day Friday last week, so I didn't know exactly what to do. I thought about doing some web view stuff, but I realized I was kinda blocked by this work. As such, I went ahead and got started on events on https://github.com/paranext/paranext-core/tree/22-events . I'm sorry I jumped ahead of you on this, especially without notice. Would you like to discuss what we have both done? I haven't implemented registering for events so that event messages only go where they should - we could discuss and see if you'd like to do that work if you'd like to have a bit easier ramp into working on these services! Let me know what you think.

tjcouch-sil:
Ira and I discussed and decided to split out #62 into its own work item. That way, we can go ahead and get this code in when we can, and we don't have to worry about who does that work.

Issue #16, opened 2023-02-01
Scaffold Core - Combine Electron, React, C# websocket


tjcouch-sil:

tjcouch-sil:
https://github.com/paranext/paranext-core/pull/24

tjcouch-sil:
To do next:
- [ ] make client folder for clients to share
- [ ] reconnectClient
- [ ] patch electronmon?
- [ ] load extension file

tjcouch-sil:
My electron host PR: Added extension host process that launches from main process #29

PR #2780, opened 2026-09-08
PT-4335: Route extension-raised notifications and prompts to the window they concern

timothy-mccormack:
## Summary

Notifications and prompts raised by the extension host had no way to say which window they were about, so `notification.service-router.ts` always routed them to the *focused* window — correct for generic notices, wrong when the notification concerns a project/editor open in a *different* window.

Confirmed real case (PR #2621 review finding R4-06): the shared-layout "Apply now" prompt can be sent about a project in window 1 while window 2 is focused, asking the user to apply a layout for a project they can't see.

## Changes

- `notification.service-model.ts`: new optional, `@experimental` `webViewId` field on `PlatformNotification`. When set, routes to the window that owns that web view instead of the focused window; falls back to focus (with a logged warning) if it can't be resolved.
- `web-view.service-router.ts`: new exported `findWindowIdOwningWebView`, reusing the existing `findOwner` ownership fan-out rather than duplicating it.
- `notification.service-router.ts`: `send()` uses the above when a notification names a `webViewId`.
- `shared-layout-receiver.model.ts`: passes the relevant editor's `webViewId` on its "Apply shared layout now?" prompt — the confirmed bug fix.
- `lib/papi-dts/papi.d.ts`: regenerated to publish the new field to extension-facing types.

No `webViewId` set → routing is byte-for-byte unchanged (focused window), so generic notices are unaffected.

## Scope note

A second case cited in the ticket ("empty-editor detection") turned out to be a *command*-routing bug in a different service (`web-view.service-router.ts`'s `openScriptureEditor` routing), not a notification — left out of this PR as a separate concern.

## Testing

- 6 new unit tests across `notification.service-router.test.ts`, `web-view.service-router.test.ts`, and `shared-layout-receiver.model.test.ts` (owner-routing, fallback-to-focus, unreachable-windows cases), all via TDD.
- `tsc --noEmit` clean in both `core` and `extensions`.
- `eslint`/`prettier` clean on all touched files.
- Manually verified in a running two-window session: a notification naming a `webViewId` lands in the window that owns it even when a different window is focused; falls back to focus when unset or unresolvable.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2780)

### DIFF
--- a/e2e-tests/tests/isolated/multi-window/per-window-ui.spec.ts
+++ b/e2e-tests/tests/isolated/multi-window/per-window-ui.spec.ts
@@ -4,9 +4,9 @@
  * The multi-window architecture makes formerly app-singleton UI surfaces per-window by
  * construction: each window is a full renderer with its own OverlayHost, Toaster, toolbar, and
  * module state, web view iframes inherit their PARENT window's papi object, and the main process
- * routes dialog/notification/web-view requests to the window they name, or to the focused window
- * when they name none. Nothing pinned any of that behaviour — this spec is the regression guard
- * that does.
+ * routes notification and web-view requests to the window they name, and everything else — dialogs
+ * included, which cannot name one — to the focused window. Nothing pinned any of that behaviour —
+ * this spec is the regression guard that does.
  *
  * One test, one Electron instance (each launch costs 30+ seconds, so the scenarios share two
  * windows), asserting in order:
@@ -255,9 +255,9 @@ function showModalAlertViaWebSocket(prompt: string): Promise<unknown> {
 
 /**
  * Send a notification through the GENERIC notification service — the name the main process serves
- * with a service router that routes `send` to the window they name, or to the focused window when
- * they name none. `duration: 0` means the toast never auto-closes, so assertions cannot race an
- * auto-dismiss; tests dismiss explicitly.
+ * with a service router that routes `send` to the window the notification names, or to the focused
+ * window when it names none. `duration: 0` means the toast never auto-closes, so assertions cannot
+ * race an auto-dismiss; tests dismiss explicitly.
  */
 async function sendNotification(message: string): Promise<string | number> {
   return sendPapiRequestOnce<string | number>(
--- a/e2e-tests/tests/isolated/multi-window/per-window-ui.spec.ts
+++ b/e2e-tests/tests/isolated/multi-window/per-window-ui.spec.ts
@@ -4,8 +4,9 @@
  * The multi-window architecture makes formerly app-singleton UI surfaces per-window by
  * construction: each window is a full renderer with its own OverlayHost, Toaster, toolbar, and
  * module state, web view iframes inherit their PARENT window's papi object, and the main process
- * routes generic dialog/notification/web-view requests to the focused window. Nothing pinned any of
- * that behaviour — this spec is the regression guard that does.
+ * routes dialog/notification/web-view requests to the window they name, or to the focused window
+ * when they name none. Nothing pinned any of that behaviour — this spec is the regression guard
+ * that does.
  *
  * One test, one Electron instance (each launch costs 30+ seconds, so the scenarios share two
  * windows), asserting in order:
@@ -254,8 +255,9 @@ function showModalAlertViaWebSocket(prompt: string): Promise<unknown> {
 
 /**
  * Send a notification through the GENERIC notification service — the name the main process serves
- * with a service router that forwards `send` to the focused window. `duration: 0` means the toast
- * never auto-closes, so assertions cannot race an auto-dismiss; tests dismiss explicitly.
+ * with a service router that routes `send` to the window they name, or to the focused window when
+ * they name none. `duration: 0` means the toast never auto-closes, so assertions cannot race an
+ * auto-dismiss; tests dismiss explicitly.
  */
 async function sendNotification(message: string): Promise<string | number> {
   return sendPapiRequestOnce<string | number>(
--- a/extensions/src/platform-scripture-editor/src/shared-layout-receiver.model.ts
+++ b/extensions/src/platform-scripture-editor/src/shared-layout-receiver.model.ts
@@ -161,7 +161,7 @@ export class SharedLayoutReceiver {
   /**
    * The Simple-mode editor a completed sync's "Apply now" prompt should be about — its project, to
    * decide whether the layout changed, and its web view id, so the router can send the prompt to
-   * the window that actually has it open instead of the focused window (PT-4335).
+   * the window that actually has it open instead of the focused window.
    */
   private async getOpenSimpleModeEditor(): Promise<
     { projectId: string; webViewId: string } | undefined
@@ -172,12 +172,14 @@ export class SharedLayoutReceiver {
       (def) => def.webViewType === SCRIPTURE_EDITOR_WEBVIEW_TYPE && !def.state?.isReadOnly,
     );
     if (!editor?.projectId) return undefined;
+
     return { projectId: editor.projectId, webViewId: editor.id };
   }
 
   private async handleSyncCompleted(): Promise<void> {
     const editor = await this.getOpenSimpleModeEditor();
     if (!editor) return;
+
     const { projectId, webViewId } = editor;
 
     // Only manual Send/Receives reach here (`onSyncStateChanged` does not fire for the programmatic
@@ -203,7 +205,7 @@ export class SharedLayoutReceiver {
       clickCommand: 'platformScriptureEditor.applySharedLayout',
       clickCommandLabel: '%platformScriptureEditor_sharedLayout_applyNow%',
       // Routes the prompt to the window that has this editor open, rather than the focused window —
-      // the project it is about may not be the one the user is currently looking at (PT-4335).
+      // the project it is about may not be the one the user is currently looking at.
       webViewId,
     });
     this.projectIdByNotificationId.set(notificationId, projectId);
--- a/extensions/src/platform-scripture-editor/src/shared-layout-receiver.model.ts
+++ b/extensions/src/platform-scripture-editor/src/shared-layout-receiver.model.ts
@@ -158,18 +158,27 @@ export class SharedLayoutReceiver {
     }
   }
 
-  private async getOpenSimpleModeProjectId(): Promise<string | undefined> {
+  /**
+   * The Simple-mode editor a completed sync's "Apply now" prompt should be about — its project, to
+   * decide whether the layout changed, and its web view id, so the router can send the prompt to
+   * the window that actually has it open instead of the focused window (PT-4335).
+   */
+  private async getOpenSimpleModeEditor(): Promise<
+    { projectId: string; webViewId: string } | undefined
+  > {
     if ((await this.papi.settings.get('platform.interfaceMode')) !== 'simple') return undefined;
     const defs = await this.papi.webViews.getAllOpenWebViewDefinitions();
     const editor = defs.find(
       (def) => def.webViewType === SCRIPTURE_EDITOR_WEBVIEW_TYPE && !def.state?.isReadOnly,
     );
-    return editor?.projectId;
+    if (!editor?.projectId) return undefined;
+    return { projectId: editor.projectId, webViewId: editor.id };
   }
 
   private async handleSyncCompleted(): Promise<void> {
-    const projectId = await this.getOpenSimpleModeProjectId();
-    if (!projectId) return;
+    const editor = await this.getOpenSimpleModeEditor();
+    if (!editor) return;
+    const { projectId, webViewId } = editor;
 
     // Only manual Send/Receives reach here (`onSyncStateChanged` does not fire for the programmatic
     // project-switch sync), so always take the notify path when the layout changed. Project-switch
@@ -193,6 +202,9 @@ export class SharedLayoutReceiver {
       message: '%platformScriptureEditor_sharedLayout_newLayoutAvailable%',
       clickCommand: 'platformScriptureEditor.applySharedLayout',
       clickCommandLabel: '%platformScriptureEditor_sharedLayout_applyNow%',
+      // Routes the prompt to the window that has this editor open, rather than the focused window —
+      // the project it is about may not be the one the user is currently looking at (PT-4335).
+      webViewId,
     });
     this.projectIdByNotificationId.set(notificationId, projectId);
     this.notificationIdByProjectId.set(projectId, notificationId);
--- a/src/main/services/__tests__/notification.service-router.test.ts
+++ b/src/main/services/__tests__/notification.service-router.test.ts
@@ -21,6 +21,7 @@ const mocks = vi.hoisted(() => {
     getReadyWindowIds: vi.fn(),
     getUnreachableWindowIds: vi.fn(),
     getAbandonedWindowIds: vi.fn(),
+    isWindowClosing: vi.fn(),
     networkObjectGet: vi.fn(),
     networkObjectSet: vi.fn(),
     loggerWarn: vi.fn(),
@@ -55,6 +56,7 @@ vi.mock('@main/services/window-state.service', () => ({
   getReadyWindowIds: mocks.getReadyWindowIds,
   getUnreachableWindowIds: mocks.getUnreachableWindowIds,
   getAbandonedWindowIds: mocks.getAbandonedWindowIds,
+  isWindowClosing: mocks.isWindowClosing,
 }));
 vi.mock('@shared/services/network-object.service', () => ({
   networkObjectService: { get: mocks.networkObjectGet, set: mocks.networkObjectSet },
@@ -98,6 +100,7 @@ describe('notification service router', () => {
     mocks.getReadyWindowIds.mockReturnValue([]);
     mocks.getUnreachableWindowIds.mockReturnValue([]);
     mocks.getAbandonedWindowIds.mockReturnValue([]);
+    mocks.isWindowClosing.mockReturnValue(false);
     mocks.findWindowIdOwningWebView.mockResolvedValue({
       windowId: undefined,
       hadUnreachableWindows: false,
@@ -157,8 +160,26 @@ describe('notification service router', () => {
     await router.send({ message: 'hi', severity: 'info', webViewId: 'wv-1' });
 
     expect(mocks.findWindowIdOwningWebView).toHaveBeenCalledWith('wv-1', expect.any(String));
-    expect(owner.send).toHaveBeenCalled();
+    expect(owner.send).toHaveBeenCalledWith({ message: 'hi', severity: 'info', webViewId: 'wv-1' });
     expect(focused.send).not.toHaveBeenCalled();
+  });
+
+  test('falls back to the focused window rather than sending into a closing owner window', async () => {
+    const focused = windowShard([]);
+    const closing = windowShard([]);
+    withWindows({ 1: focused, 2: closing });
+    mocks.findWindowIdOwningWebView.mockResolvedValue({
+      windowId: '2',
+      hadUnreachableWindows: false,
+    });
+    mocks.isWindowClosing.mockImplementation((windowId: string) => windowId === '2');
+    const router = await getRouter();
+
+    await router.send({ message: 'hi', severity: 'info', webViewId: 'wv-1' });
+
+    expect(closing.send).not.toHaveBeenCalled();
+    expect(focused.send).toHaveBeenCalled();
+    expect(mocks.loggerWarn).toHaveBeenCalledWith(expect.stringContaining('is closing'));
   });
 
   test('falls back to the focused window when a notification names a webViewId nobody owns', async () => {
@@ -173,9 +194,48 @@ describe('notification service router', () => {
 
     await router.send({ message: 'hi', severity: 'info', webViewId: 'wv-missing' });
 
-    expect(focused.send).toHaveBeenCalled();
+    expect(focused.send).toHaveBeenCalledWith({
+      message: 'hi',
+      severity: 'info',
+      webViewId: 'wv-missing',
+    });
     expect(other.send).not.toHaveBeenCalled();
     expect(mocks.loggerWarn).toHaveBeenCalledWith(expect.stringContaining('wv-missing'));
+  });
+
+  test('falls back to the focused window when the owner window cannot serve its shard', async () => {
+    const focused = windowShard([]);
+    // Window 2 announced a notification shard, but it no longer resolves
+    withWindows({ 1: focused, 2: undefined });
+    mocks.findWindowIdOwningWebView.mockResolvedValue({
+      windowId: '2',
+      hadUnreachableWindows: false,
+    });
+    const router = await getRouter();
+
+    await router.send({ message: 'hi', severity: 'info', webViewId: 'wv-1' });
+
+    expect(focused.send).toHaveBeenCalled();
+    expect(mocks.loggerWarn).toHaveBeenCalledWith(
+      expect.stringContaining("that window's notification service could not be resolved"),
+    );
+  });
+
+  test('says in the log when a window could not be asked who owns the web view', async () => {
+    const focused = windowShard([]);
+    withWindows({ 1: focused, 2: windowShard([]) }, { unreachableWindowIds: ['2'] });
+    mocks.findWindowIdOwningWebView.mockResolvedValue({
+      windowId: undefined,
+      hadUnreachableWindows: true,
+    });
+    const router = await getRouter();
+
+    await router.send({ message: 'hi', severity: 'info', webViewId: 'wv-1' });
+
+    expect(focused.send).toHaveBeenCalled();
+    expect(mocks.loggerWarn).toHaveBeenCalledWith(
+      expect.stringContaining('some windows could not be asked'),
+    );
   });
 
   test('dismisses in the window showing the notification, not the focused one', async () => {
--- a/src/main/services/__tests__/notification.service-router.test.ts
+++ b/src/main/services/__tests__/notification.service-router.test.ts
@@ -24,6 +24,7 @@ const mocks = vi.hoisted(() => {
     networkObjectGet: vi.fn(),
     networkObjectSet: vi.fn(),
     loggerWarn: vi.fn(),
+    findWindowIdOwningWebView: vi.fn(),
     shardAnnouncementListeners,
     onDidCreateNetworkObject: vi.fn((listener: (details: NetworkObjectDetails) => void) => {
       shardAnnouncementListeners.create.push(listener);
@@ -63,6 +64,9 @@ vi.mock('@shared/services/network-object.service', () => ({
 vi.mock('@shared/services/logger.service', () => ({
   logger: { info: vi.fn(), warn: mocks.loggerWarn, error: vi.fn() },
 }));
+vi.mock('@main/services/web-view.service-router', () => ({
+  findWindowIdOwningWebView: mocks.findWindowIdOwningWebView,
+}));
 
 /** Capture the router object registered under the generic name */
 async function getRouter() {
@@ -94,6 +98,10 @@ describe('notification service router', () => {
     mocks.getReadyWindowIds.mockReturnValue([]);
     mocks.getUnreachableWindowIds.mockReturnValue([]);
     mocks.getAbandonedWindowIds.mockReturnValue([]);
+    mocks.findWindowIdOwningWebView.mockResolvedValue({
+      windowId: undefined,
+      hadUnreachableWindows: false,
+    });
   });
 
   test('sends to a window that registered its shard after the router started', async () => {
@@ -134,6 +142,40 @@ describe('notification service router', () => {
 
     expect(focused.send).toHaveBeenCalled();
     expect(other.send).not.toHaveBeenCalled();
+  });
+
+  test('sends a notification naming a webViewId to the window that owns it, not the focused one', async () => {
+    const focused = windowShard([]);
+    const owner = windowShard([]);
+    withWindows({ 1: focused, 2: owner });
+    mocks.findWindowIdOwningWebView.mockResolvedValue({
+      windowId: '2',
+      hadUnreachableWindows: false,
+    });
+    const router = await getRouter();
+
+    await router.send({ message: 'hi', severity: 'info', webViewId: 'wv-1' });
+
+    expect(mocks.findWindowIdOwningWebView).toHaveBeenCalledWith('wv-1', expect.any(String));
+    expect(owner.send).toHaveBeenCalled();
+    expect(focused.send).not.toHaveBeenCalled();
+  });
+
+  test('falls back to the focused window when a notification names a webViewId nobody owns', async () => {
+    const focused = windowShard([]);
+    const other = windowShard([]);
+    withWindows({ 1: focused, 2: other });
+    mocks.findWindowIdOwningWebView.mockResolvedValue({
+      windowId: undefined,
+      hadUnreachableWindows: false,
+    });
+    const router = await getRouter();
+
+    await router.send({ message: 'hi', severity: 'info', webViewId: 'wv-missing' });
+
+    expect(focused.send).toHaveBeenCalled();
+    expect(other.send).not.toHaveBeenCalled();
+    expect(mocks.loggerWarn).toHaveBeenCalledWith(expect.stringContaining('wv-missing'));
   });
 
   test('dismisses in the window showing the notification, not the focused one', async () => {
--- a/src/main/services/notification.service-router.ts
+++ b/src/main/services/notification.service-router.ts
@@ -3,19 +3,27 @@
  * network object name and routes calls to a window's notification service shard (e.g.
  * "NotificationService-f81d4fae-7dec-11d0-a765-00a0c91e6bf6"). A new notification goes to the
  * focused window, so one raised by a background task appears where the user is looking rather than
- * in whichever renderer happened to start first; dismissing one goes to the window that is actually
- * showing it.
+ * in whichever renderer happened to start first — UNLESS it names
+ * {@link PlatformNotification.webViewId}, in which case it goes to the window that owns that web
+ * view: a notification or prompt about a specific project or editor is about something concrete,
+ * and the window showing it may not be the one the user currently has focused. Dismissing one goes
+ * to the window that is actually showing it.
  *
  * See the router/shard pattern in `.context/standards/Architecture.md` § "Service router and
  * service shard".
  */
 
 import { getReadyWindowIds } from '@main/services/window-state.service';
-import { createTargetShardResolver } from '@main/services/target-shard-resolver.util';
+import {
+  createTargetShardResolver,
+  resolveShardForWindow,
+} from '@main/services/target-shard-resolver.util';
+import { findWindowIdOwningWebView } from '@main/services/web-view.service-router';
 import {
   INotificationService,
   NOTIFICATION_SERVICE_NETWORK_OBJECT_DOCS,
   NotificationServiceNetworkObjectName,
+  PlatformNotification,
 } from '@shared/models/notification.service-model';
 import { logger } from '@shared/services/logger.service';
 import { networkObjectService } from '@shared/services/network-object.service';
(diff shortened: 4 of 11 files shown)


extensions/src/platform-scripture-editor/src/shared-layout-receiver.model.ts
```diff
@@ -158,18 +158,27 @@ export class SharedLayoutReceiver {
     }
   }
 
-  private async getOpenSimpleModeProjectId(): Promise<string | undefined> {
+  /**
+   * The Simple-mode editor a completed sync's "Apply now" prompt should be about — its project, to
+   * decide whether the layout changed, and its web view id, so the router can send the prompt to
+   * the window that actually has it open instead of the focused window (PT-4335).
```
lyonsil: **#9** - low · confirmed

Two new comments carry a backward-facing `(PT-4335)` reference to this PR's own ticket (second site at :206).

The forward-facing-comments rule's cut list leads with 'Ticket/PR references for work done *in this PR*' and directs that home to 'the commit message or PR body, not the source file'. Neither comment is a `TODO(PT-XXXX)` forward pointer to deferred work, and both sentences read fine with the ID stripped.

**Suggested change** (the same edit is needed at `:206`, which this comment cannot reach):

```suggestion
   * the window that actually has it open instead of the focused window.
```

At `:206`, drop the trailing ` (PT-4335)` the same way. The ticket stays in the commit message, where it already is.

_Verification: Both comments are new in this PR per git diff, and commit 2e41d384691 is titled 'PT-4335: Route extension-raised notifications...' - the PR's own ticket, with the work fully done in that same commit, not deferred. The rule's cut-list leads with exactly this case. Neither is a TODO(PT-XXXX) forward pointer, a deferred-work pointer, or a non-inlinable deep link (the keep-list exceptions), and both sentences are complete with the ID removed. Low severity: doc-comment cleanup with no functional impact._

Same fix as #12, #19.

_Other findings in this file: #12_

extensions/src/platform-scripture-editor/src/shared-layout-receiver.model.ts
```diff
@@ -158,18 +158,27 @@ export class SharedLayoutReceiver {
...
+    { projectId: string; webViewId: string } | undefined
+  > {
     if ((await this.papi.settings.get('platform.interfaceMode')) !== 'simple') return undefined;
     const defs = await this.papi.webViews.getAllOpenWebViewDefinitions();
     const editor = defs.find(
       (def) => def.webViewType === SCRIPTURE_EDITOR_WEBVIEW_TYPE && !def.state?.isReadOnly,
     );
-    return editor?.projectId;
+    if (!editor?.projectId) return undefined;
+    return { projectId: editor.projectId, webViewId: editor.id };
```
lyonsil, line 176: **#12** - low · downgraded on review

Two early returns added or reshaped by this PR have no blank line after them (second site at :180).

Code-Style-Guide.md:204: 'Always add a blank line after an early return to visually separate it.' :174 is a wholly new guard; at :180 the pre-PR code had a blank line and the new destructuring assignment was inserted into that gap.

**Suggested change** - a blank line after the early return above:

```suggestion

    return { projectId: editor.projectId, webViewId: editor.id };
```

The same applies at `:180`/`:181`. The file already breaks this rule in a few pre-existing places, so this is polish, not a defect.

_Why this is downgraded: The rule text and both violations are confirmed, including that the pre-PR code at what is now :180 had the blank line and the new destructure was inserted into it. But this same file already contains multiple pre-existing, PR-untouched violations of the identical rule (line 151, applyFromNotification's guard, and the carried-forward guard at 169). `npx eslint --no-ignore` on the file produces zero warnings and no padding-line-between-statements rule is configured anywhere, so this is a prose convention with no tooling enforcement in a file that is demonstrably inconsistent about it._

Same fix as #9, #19.

_Other findings in this file: #9_

src/main/services/__tests__/notification.service-router.test.ts
```diff
@@ -136,6 +144,40 @@ describe('notification service router', () => {
     expect(other.send).not.toHaveBeenCalled();
   });
 
+  test('sends a notification naming a webViewId to the window that owns it, not the focused one', async () => {
```
lyonsil, line 150: **#2** - medium · confirmed

Two of `getSendTargetShard`'s four outcomes are dead to the entire main-process test suite.

`getSendTargetShard` has four outcomes: no webViewId -> focused shard; owner found and shard resolves; owner found but `resolveShardForWindow` throws (catch at :80-84); no owner (else at :85-89). The new tests cover only the second and fourth. The catch-fallback is the PR's own stated invariant at :60-63 ('Falls back to the focused window rather than throwing'), so a regression there silently turns a user-facing notification into a rejected `send`. `hadUnreachableWindows` is pinned `false` in the `beforeEach` default and in every test, so the 'some windows could not be asked' branch never runs.

**Suggested change** - two tests, verified to pass in 8 ms with no timer stalls. `withWindows({ 1: focused, 2: undefined })` is the trick: window 2 is announced (so it is indexed) but its shard resolves to `undefined`, which makes `resolveShardForWindow` throw immediately instead of waiting out the 5-second grace period.

```ts
  test('falls back to the focused window when the owner window cannot serve its shard', async () => {
    const focused = windowShard([]);
    // Window 2 announced a notification shard, but it no longer resolves
    withWindows({ 1: focused, 2: undefined });
    mocks.findWindowIdOwningWebView.mockResolvedValue({
      windowId: '2',
      hadUnreachableWindows: false,
    });
    const router = await getRouter();

    await router.send({ message: 'hi', severity: 'info', webViewId: 'wv-1' });

    expect(focused.send).toHaveBeenCalled();
    expect(mocks.loggerWarn).toHaveBeenCalledWith(
      expect.stringContaining("that window's notification service could not be resolved"),
    );
  });

  test('says in the log when a window could not be asked who owns the web view', async () => {
    const focused = windowShard([]);
    withWindows({ 1: focused, 2: windowShard([]) }, { unreachableWindowIds: ['2'] });
    mocks.findWindowIdOwningWebView.mockResolvedValue({
      windowId: undefined,
      hadUnreachableWindows: true,
    });
    const router = await getRouter();

    await router.send({ message: 'hi', severity: 'info', webViewId: 'wv-1' });

    expect(focused.send).toHaveBeenCalled();
    expect(mocks.loggerWarn).toHaveBeenCalledWith(
      expect.stringContaining('some windows could not be asked'),
    );
  });
```

Proof these close the gap: with them in place, the two mutations that previously survived the whole 893-test suite (making the `catch` rethrow, and inverting the `hadUnreachableWindows` ternary) now both fail.

_Verification: Both mutations survived. Replacing the catch body at notification.service-router.ts:80-84 with `throw e;` left all 47 files / 893 tests under `src/main` passing. Inverting the `hadUnreachableWindows` ternary at :87 also left 893 passing. The test file's own `withWindows` wrapper accepts `unreachableWindowIds` but no test ever passes it. Medium rather than higher because only the catch-fallback half is behavioral; the `hadUnreachableWindows` half is log-text cosmetics. Worktree restored and verified clean._

Same fix as #16.

_Other findings in this file: #16_

src/main/services/__tests__/notification.service-router.test.ts
```diff
@@ -136,6 +144,40 @@ describe('notification service router', () => {
...
+    mocks.findWindowIdOwningWebView.mockResolvedValue({
+      windowId: '2',
+      hadUnreachableWindows: false,
+    });
+    const router = await getRouter();
+
+    await router.send({ message: 'hi', severity: 'info', webViewId: 'wv-1' });
+
+    expect(mocks.findWindowIdOwningWebView).toHaveBeenCalledWith('wv-1', expect.any(String));
+    expect(owner.send).toHaveBeenCalled();
```
lyonsil: **#16** - medium -> low · confirmed

Neither new test asserts the notification payload survives the rewrite from variadic forwarding to single-argument forwarding.

The PR changed `send: async (...args) => (await getTargetNotificationShard()).send(...args)` to `send: async (notification) => (await getSendTargetShard(notification)).send(notification)`. Both new tests assert only `toHaveBeenCalled()`, so a change forwarding a stripped object would pass the whole file. `toHaveBeenCalledWith(notification)` at :160 and :176 closes it.

**Suggested change** - pin the payload, not just the fact of a call:

```ts
    expect(owner.send).toHaveBeenCalledWith({ message: 'hi', severity: 'info', webViewId: 'wv-1' });
```

and the matching change in the fallback test at `:176`. This is what makes the field-drop mutation fail instead of passing.

_Verification: Mutation-proven: replacing notification.service-router.ts:153 with `send({ message, severity })` - dropping webViewId/clickCommand/duration/notificationId - left all 11 tests green and produced zero tsc errors over the real src graph. TypeScript only blocks the crudest form (`send({})` fails TS2345 because message/severity are required), which is not enough to downgrade since the realistic field-drop typechecks clean. Keyword-grep over the whole test tree finds no other test pinning the router's forwarded payload; the only end-to-end check pins message only and lives in the isolated Playwright project while CI runs --project=smoke. Low rather than medium because no defect exists today and the site is a one-line pass-through; a control mutation confirms the new tests are falsifiable for the routing behaviour they claim._

Same fix as #2.

_Other findings in this file: #2_

lyonsil (commented):
Review of the `webViewId` notification routing. Twelve findings, three of them worth acting on before merge and the rest small.

Several comments carry a one-click **suggested change**. Every code change proposed here was applied to a worktree at `4d54095` and checked: `tsc --noEmit` clean on the touched files, `eslint` and `prettier --check` clean, and `vitest run src/main` green at 896 tests. Where a change needs more than the suggestion block itself, the comment says so explicitly.

Two things worth flagging up front, because neither is visible from the diff alone:

- The TSDoc on `PlatformNotification.webViewId` is republished into the generated `lib/papi-dts/papi.d.ts`. After editing it, run `npm run build:types` and commit the regenerated file.
- Adding `isWindowClosing` to `notification.service-router.ts` also requires three edits in `notification.service-router.test.ts`, which mocks that module. Without them every test in the file fails.

**#1** (medium · confirmed) extensions/src/platform-scripture-editor/src/markers-view-notifier.model.ts:183 - The markers-view read-only prompt is an identical sibling site that holds the web view id but was not opted into the new routing.

`handleWebView` subscribes to `onDidOpenWebView`/`onDidUpdateWebView`, which are global network events emitted by every window's shard, so the notifier has no window affinity. It already reads `webViewDefinition.id` (line 128), and `id` is a required `WebViewId` on `SavedWebViewDefinition`, so the opt-in is a one-line `webViewId: webViewDefinition.id`. Two background paths reach it with another window focused: Power-mode multi-window restore (an earlier window's renderer emits its editors' open events after a later window takes focus) and `auto-sync-edit-block-driver.ts` writing `isSyncBlocked` through `updateWebViewDefinitionSync`, which unconditionally emits `onDidUpdateWebView` during an automatic Send/Receive. The prompt carries a 'don't show for this project today' action, so it is a prompt in a window it is not about - the harm the PR's own router comment says it exists to prevent.

**Suggested change** - one line, since `webViewDefinition.id` is already in scope (it is read at `:128`):

```ts
        const notificationId = await this.papiInstance.notifications.send({
          severity: 'info',
          message,
          clickCommand: 'platformScriptureEditor.dismissMarkerNotificationForProjectToday',
          clickCommandLabel: localizedClickLabel,
          // This prompt is about this editor's project, and the editor may be in a window the user
          // is not currently looking at.
          webViewId: webViewDefinition.id,
        });
```

`id` is a required field on `SavedWebViewDefinition`, so no typing work is needed. Verified: typechecks and lints clean.

_Verification: Deterministic grep of every notifications.send/notificationService.send in the repo confirms this is the only remaining backend site holding a web view id; the other two extension-host senders (extension.service.ts:1270, data-provider.service.ts:180) are generic notices with no web view in scope and are correctly left on focus routing. The reviewer's cross-check that the sync-edit-blocked.util.ts sites are correctly excluded is right - both copies import @papi/frontend and all callers are user-initiated handlers in a web view. Severity stays medium: the message is formatted with the project name so the dismiss click is not silently mis-targeted, but a scope defense is weak given the commit title says 'notifications and prompts'._

**#5** (low · confirmed) e2e-tests/tests/isolated/multi-window/per-window-ui.spec.ts:7 - The spec header and the `sendNotification` TSDoc still assert the unconditional focused-window routing rule this PR made conditional.

Lines 4-8 say the main process 'routes generic dialog/notification/web-view requests to the focused window. Nothing pinned any of that behaviour - this spec is the regression guard that does', and lines 255-259 repeat it. After this PR `send` forwards to the focused window only when the notification does not name a `webViewId`. The router's own module doc was updated; this spec's was not.

**Suggested change** - this file is outside the diff, so it needs a manual edit. In the header at `:7-8`, replace "routes generic dialog/notification/web-view requests to the focused window" with "routes dialog/notification/web-view requests to the window they name, or to the focused window when they name none". In the `sendNotification` TSDoc at `:257`, replace "forwards `send` to the focused window" the same way.

_Verification: `git diff MERGE_BASE HEAD -- e2e-tests/.../per-window-ui.spec.ts` is empty, so this file is untouched and the outdated claim sits on unchanged lines. The router's module doc (lines 1-13) was updated with the UNLESS clause; the spec header and helper TSDoc were not. 'Generic' scopes requests against per-window-scoped surfaces (shard vs router), not against payload-conditional routing, so it does not save the claim. Low impact: `sendNotification` never passes a webViewId, so every call this spec makes does land in the focused window - the doc is stale rather than wrong for the scenarios it runs._

**#32** (low · confirmed) src/renderer/services/notification.service-shard.ts:223 - The shard's TSDoc still asserts the unconditional focused-window rule after the router's counterpart was updated.

`startNotificationServiceShard`'s TSDoc says the main process 'forwards to the focused window, so a notification raised by a background task lands where the user is looking'. That is now true only when the notification does not name a `webViewId`.

**Suggested change** - this file is outside the diff, so it needs a manual edit at `:222-224`:

```ts
 * notification UI. The main process publishes the generic name and forwards to the window that owns
 * the web view a notification names, or to the focused window when it names none, so a notification
 * raised by a background task lands where the user is looking.
```

_Verification: git diff MERGE_BASE HEAD on the shard file is empty, so this PR left the TSDoc untouched, while git diff on notification.service-router.ts shows the router's module doc did gain the 'UNLESS it names {@link PlatformNotification.webViewId}' clause - the pair has drifted. grep for webViewId in the shard file returns nothing, confirming it. grep of lib/papi-dts/papi.d.ts for notification.service-shard / startNotificationServiceShard returns no matches, so this is internal-only and not published PAPI surface, which supports keeping severity low but does not refute the claim._

src/main/services/notification.service-router.ts
```diff
@@ -44,6 +52,45 @@ const getTargetNotificationShard = createTargetShardResolver(
   notificationShards,
 );
 
+/**
+ * The notification service shard a `send` should run in: the window that owns
+ * {@link PlatformNotification.webViewId} when the notification names one, otherwise the focused
+ * window, exactly as before that field existed.
```
lyonsil: **#19** - low · confirmed

The new TSDoc on `getSendTargetShard` ends with change narration.

forward-facing-comments.md: 'Change narration with no ID at all - previously, used to ... describe what the code does, not what it stopped doing.' 'otherwise the focused window' already carries the whole contract; ', exactly as before that field existed' adds only a comparison to pre-PR code.

**Suggested change** - drop the comparison to the pre-PR code:

```suggestion
 * window.
```

The preceding line already says "otherwise the focused window", which is the whole contract.

_Verification: git diff confirms the entire getSendTargetShard function and its TSDoc are new in this PR, so this is on a changed line. The rule's cut list targets exactly this phrasing, and the clause is not a constraint, invariant, or rationale from the keep list. The rewritten module doc at lines 1-14 (also touched by this PR) describes only current behaviour with no before/after framing, so the finding correctly stays scoped to this single site rather than broadening._

Same fix as #9, #12.

_Other findings in this file: #3, #22_

src/main/services/notification.service-router.ts
```diff
@@ -44,6 +52,45 @@ const getTargetNotificationShard = createTargetShardResolver(
...
+ *
+ * Falls back to the focused window rather than throwing when the named web view cannot be resolved
+ * to a window — a notification is user-facing feedback, and showing it in the wrong window is a
+ * better outcome than not showing it at all. Every fallback path is logged, since a caller that
+ * bothered to name a `webViewId` presumably cared where the notification landed.
+ */
+async function getSendTargetShard(
+  notification: PlatformNotification,
+): Promise<INotificationService> {
+  if (notification.webViewId !== undefined) {
```
lyonsil: **#22** - medium -> low · downgraded on review

An update-send that reuses a notification id but omits `webViewId` routes to a different window than the original, so the documented field-merge cannot apply.

`notificationId`'s TSDoc promises unconditionally that 'any optional field you omit keeps the value it had on the previous send for that id'. That merge is per-renderer module state, strictly downstream of the routing decision made here on the raw object, so the routing key can never honour it. The exception is undocumented on both the notificationId TSDoc and the new webViewId TSDoc.

**Suggested change** - no code change; document the exception. Add to the `notificationId` TSDoc at `:140-143` that `webViewId` is the one field the merge cannot apply to, because routing is decided in the main process before the renderer merges. Worth doing now while the field is `@experimental`, since no caller mixes the two yet.

_Why this is downgraded: The mechanism is verified end to end: main holds no per-id state, the merge is per-renderer module state applied downstream, and an untracked id yields existingToastId undefined, which an existing passing test proves creates a fresh toast. But the live-bug half does not hold: a deterministic scan of every send object literal in the repo returns 8 hits and the sets are disjoint - the only webViewId sender passes no notificationId, and all seven fixed-id senders set no webViewId; C# cannot set it at all. Nor is it newly introduced as a class: pre-PR the router already resolved the target shard per send, so focus drift between two same-id sends already routed an update to a window that never saw the id. What the PR adds is a second deterministic axis plus an undocumented exception - a latent trap and doc gap on an @experimental field._

_Other findings in this file: #3, #19_

src/main/services/notification.service-router.ts
```diff
@@ -44,6 +52,45 @@ const getTargetNotificationShard = createTargetShardResolver(
...
+    const { windowId, hadUnreachableWindows } = await findWindowIdOwningWebView(
+      notification.webViewId,
+      'route notification',
+    );
+    if (windowId !== undefined) {
```
lyonsil: **#3** - high -> medium · downgraded on review

The new send path can route a notification into a window that has begun closing, because it never re-checks `isWindowClosing`.

`findWindowIdOwningWebView` fans out over `getReadyWindowIds()`, which deliberately keeps closing windows - `markWindowClosing` never removes them from `readyWindowIds`. The pre-PR path went through `getTargetWindowId` -> `getRoutingTarget`, which filters closing windows precisely so no notification lands 'in the window the user is watching disappear'. The sibling `openWebViewInOwningWindow` re-reads `isWindowClosing` at exactly this point because the fan-out takes as long as the slowest window; grep for 'closing' across notification.service-router.ts, target-shard-resolver.util.ts and service-shard-index.ts returns nothing, so the whole new path is closing-blind.

**Suggested change** - re-read the closing state at the last moment, exactly as the sibling `openWebViewInOwningWindow` does at `web-view.service-router.ts:493-500`, but falling back rather than throwing (this file's stated philosophy at `:60-63`). This replaces lines 68-90. It also changes `!== undefined` to a truthy check, so an empty-string `webViewId` no longer triggers a whole-app fan-out that can never match:

```ts
  if (notification.webViewId) {
    const { windowId, hadUnreachableWindows } = await findWindowIdOwningWebView(
      notification.webViewId,
      'route notification',
    );
    // Read the closing state again here, at the last moment before the window is asked to show the
    // notification. Finding the owner meant asking every window in the app, which takes as long as
    // the slowest of them — long enough for this one's close to have been decided since.
    if (windowId !== undefined && !isWindowClosing(windowId)) {
      try {
        return await resolveShardForWindow(
          NotificationServiceNetworkObjectName,
          notificationShards,
          windowId,
        );
      } catch (e) {
        logger.warn(
          `Notification named webViewId ${notification.webViewId}, owned by window ${windowId}, but that window's notification service could not be resolved (${getErrorMessage(e)}); falling back to the focused window.`,
        );
      }
    } else if (windowId !== undefined) {
      logger.warn(
        `Notification named webViewId ${notification.webViewId}, owned by window ${windowId}, but that window is closing; falling back to the focused window.`,
      );
    } else {
      logger.warn(
        `Notification named webViewId ${notification.webViewId}, but no window could be found owning it${hadUnreachableWindows ? ' (some windows could not be asked)' : ''}; falling back to the focused window.`,
      );
    }
  }
  return getTargetNotificationShard();
```

Two edits go with it, and the change does not work without them:
1. `import { getReadyWindowIds, isWindowClosing } from '@main/services/window-state.service';`
2. The test file mocks that module, so `isWindowClosing` must be added to the hoisted `mocks`, to the `vi.mock` factory, and defaulted to `false` in `beforeEach` - otherwise every test in the file throws `isWindowClosing is not a function`.

A test that fails without the guard:

```ts
  test('falls back to the focused window rather than sending into a closing owner window', async () => {
    const focused = windowShard([]);
    const closing = windowShard([]);
    withWindows({ 1: focused, 2: closing });
    mocks.findWindowIdOwningWebView.mockResolvedValue({
      windowId: '2',
      hadUnreachableWindows: false,
    });
    mocks.isWindowClosing.mockImplementation((windowId: string) => windowId === '2');
    const router = await getRouter();

    await router.send({ message: 'hi', severity: 'info', webViewId: 'wv-1' });

    expect(closing.send).not.toHaveBeenCalled();
    expect(focused.send).toHaveBeenCalled();
    expect(mocks.loggerWarn).toHaveBeenCalledWith(expect.stringContaining('is closing'));
  });
```

_Why this is downgraded: Every mechanical link verified. Blast radius is smaller than claimed: the only producer of webViewId today is the shared-layout prompt, which fires only on `onSyncStateChanged` (manual S/R only - the window-close sync calls `sendReceiveProjects` directly and is invisible to it), so the loss needs a manual sync to complete during that window's close, and the prompt concerns an editor going away with the window; on a full quit both old and new paths pick a closing window anyway. Still worth the fix: webViewId is public @experimental PAPI surface every future caller inherits, the closing state persists for the whole bounded close sync rather than being a microsecond race, and the file's own philosophy at :60-63 already prescribes log-and-fall-back._

_Other findings in this file: #19, #22_

src/main/services/web-view.service-router.ts
```diff
@@ -351,6 +351,26 @@ async function findOwner(
...
 }
 
+/**
+ * The window that owns a given web view, for callers outside this router that only need to know
+ * which window to act in — not the shard or the definition {@link findOwner} also resolves.
+ *
+ * Used by `notification.service-router.ts` to route a notification or prompt that names a
+ * `webViewId` to the window showing it, instead of the focused window.
+ *
+ * See {@link findOwner} for what `hadUnreachableWindows` means and for the reachable-window
```
lyonsil: **#30** - medium -> low · downgraded on review

The new exported function's TSDoc mislabels the tie-break and points at a `findOwner` doc that does not document one.

The actual tie-break is `matches.find(c => c.windowId === getTargetWindowId()) ?? matches.sort(byWindowAge)[0]` - routing-target first, then oldest. 'reachable-window tie-break' is a vague label, and the linked `findOwner` TSDoc documents no tie-break at all; the rule lives only in body comments. `@param webViewId` is also missing, inconsistent with the file's own siblings.

**Suggested change** - replace lines 361-365 with:

```ts
 * See {@link findOwner} for what `hadUnreachableWindows` means. In the defensive case where more
 * than one window claims the id, it prefers the routing-target window and otherwise the oldest.
 *
 * @param webViewId The web view whose owning window to find
 * @param operation Named in the warnings {@link findOwner} logs when a window could not be asked
 * @returns The owning window's id, or `undefined` both when no window owns the web view and when a
 *   window that might own it could not be asked — `hadUnreachableWindows` tells those apart
 */
```

_Why this is downgraded: The tie-break is confirmed as routing-target-first then oldest-window, so 'reachable-window tie-break' is a vague label rather than the rule, and the {@link findOwner} it points at documents no tie-break in its own TSDoc. But the multi-match path is near-unreachable for this caller: the id lookup is exact-match, ids are window-suffixed at layout load, and the code's own matches.length>1 branch logs a warn saying ids are meant to be unique and duplication means scoping was bypassed - and even when it fires, preferring a focused window that genuinely owns the view does not invert the PR's premise. The 'convention requires' half is wrong as stated: Code-Style-Guide.md:158-163 mandates @param/@returns only for API surface, defined as lib/platform-bible-react, lib/platform-bible-utils, papi.d.ts and public types in src/shared - this is src/main and appears nowhere in papi.d.ts. The one substantive fix is to say 'prefers the routing-target window, else the oldest' or drop the clause._

Same fix as #33.

src/shared/models/notification.service-model.ts
```diff
@@ -149,6 +150,19 @@ export interface PlatformNotification {
    * seconds).
    */
   duration?: number;
+  /**
+   * Optional id of a web view this notification is about. When provided, the notification is routed
+   * to the window that owns that web view instead of the focused window — for a notification or
+   * prompt raised about a specific project or editor that may not be the one the user is currently
+   * looking at. Falls back to the focused window if the web view cannot be found open in any
```
lyonsil: **#33** - medium -> low · downgraded on review

The new public TSDoc describes the fallback as narrower than it is.

The doc says the fallback applies when the web view 'cannot be found open in any window', but it also applies when the owning window could not be asked, when the view is mid-move between windows, and when a window WAS found but its notification shard would not resolve. The PR's own internal doc states it correctly ('when the named web view cannot be resolved to a window').

**Suggested change** - replace lines 157-158 with:

```ts
   * looking at. Falls back to the focused window whenever the web view's window cannot be
   * determined — it is open nowhere, a window that might have it could not be asked, or it is
   * moving between windows.
```

**Important:** this text is republished into `lib/papi-dts/papi.d.ts`, which is generated. After editing, run `npm run build:types` and commit the regenerated `papi.d.ts`; editing that file by hand will be overwritten.

_Why this is downgraded: Prong (b) is true and verified: the fallback covers three cases the public doc's wording does not - an owning window that could not be asked (findOwner says the two are indistinguishable and hadUnreachableWindows only decorates the log), a view mid-move, and a found window whose notification shard would not resolve; the PR's own internal doc at :60 states it correctly, so the public doc is narrower than the implementation. Prong (a) - 'focused window' vs the routing target - is a shorthand nit, not an error: getRoutingTarget prefers focusedWindowId as its first rung and only diverges in transient states, focusedWindowId deliberately survives app blur, and the same shorthand predates this PR in the very router that implements it. Counterweight: the nearest public analogue, OpenWebViewOptions.targetWindowId, does keep the two apart. Net: one accurate-but-incomplete sentence on an @experimental field whose miss-modes are all fail-soft and logged._

Same fix as #30.

lyonsil:
### Also looked at, deliberately not raised inline

These came out of the same review. None of them asks for a change in this PR — they are here so they are on the record rather than rediscovered later.

**Pre-existing, in code this PR touches but did not introduce** (`git blame` puts both on `2248630c5096`, 2026-07-14):

- `shared-layout-receiver.model.ts:171` — `defs.find(...)` picks the first editable editor across *all* windows in creation order. With two Simple-mode windows on different projects, the layout signature is evaluated for the oldest window's project and the other project is never offered its layout. This PR does not make it worse; it arguably improves it, since the prompt now at least lands with the editor it is about.
- `shared-layout-receiver.model.ts:211` — a prompt the user never saw is still recorded as "already asked", and only `applyForProject` clears it. It is per-signature and per-process, and it self-heals on the next project switch or restart, so the worst case is a stale Simple-mode layout until then.

**Gaps in the new feature, for follow-up rather than for this PR:**

- The prompt is now owner-routed, but the "Apply now" action it triggers is not: `focusSharedLayoutDefaultTab` passes no `existingProjectId`, so the router resolves it by preferring the routing target. It works today because the click sets focus first, but nothing in the code enforces that delivery and action agree. `OpenWebViewOptions` already has `existingProjectId` and `targetWindowId` if you ever want to close that.
- The owner window is never raised, so a prompt routed to a background window can expire unseen (~12 s, computed from message length). Not reachable today — Simple mode is single-window on every UI path — but it is the case the field exists for, so it will matter as soon as multi-window Simple mode is real.
- `webViewId` is the narrowest available key. A notification about a project with **no** open editor has no key at all and will route to the focused window. Worth a line in the TSDoc so the next caller does not assume otherwise.
- `c-sharp/Services/PlatformNotification.cs` has no `WebViewId`. That is correct for now — nothing in the C# tree can obtain a web view id — but it means .NET-raised notifications can never use this routing.
- No e2e scenario exercises the new branch. The repo's Testing-Guide puts e2e at smoke-only and both sides of the seam have unit coverage, so this is a note, not a request.

extensions/src/platform-scripture-editor/src/shared-layout-receiver.model.ts
timothy-mccormack: Fixed both sites — dropped the trailing `(PT-4335)` from both comments. Ticket stays in the commit message.

extensions/src/platform-scripture-editor/src/shared-layout-receiver.model.ts
timothy-mccormack: Added the blank line after both early returns this PR touched. Left the pre-existing interfaceMode guard alone, per your own note that it's carried-forward — not this PR's to fix.

src/main/services/__tests__/notification.service-router.test.ts
timothy-mccormack: Added both tests as given — verified they pass against current code (14/14 green). Trusting your mutation-testing verification that they catch the catch-rethrow and hadUnreachableWindows-inversion mutations rather than re-proving it myself.

src/main/services/__tests__/notification.service-router.test.ts
timothy-mccormack: Tightened both to toHaveBeenCalledWith(fullPayload).

src/main/services/notification.service-router.ts
timothy-mccormack: Dropped the before/after comparison — the preceding sentence already states the whole contract.

src/main/services/notification.service-router.ts
timothy-mccormack: Added the exception note to notificationId's TSDoc — routing is decided before the renderer ever sees the notification to merge it, so webViewId can't participate the way every other field does.

src/main/services/notification.service-router.ts
timothy-mccormack: Applied your suggested diff (isWindowClosing re-check, falling back rather than throwing, the truthy-check change) plus the mock wiring and test — verified RED against the un-patched code first, then GREEN.

src/main/services/web-view.service-router.ts
timothy-mccormack: Fixed the tie-break description and added @param/@returns.

src/shared/models/notification.service-model.ts
timothy-mccormack: Widened the fallback wording to cover all three cases. Regenerated papi.d.ts.

timothy-mccormack:
Addressing the three findings not anchored to a changed line (can't be posted inline):

**#1** (medium) — `markers-view-notifier.model.ts`: added `webViewId: webViewDefinition.id` to the read-only prompt's `notifications.send`, same one-line fix suggested. `webViewDefinition.id` was already in scope.

**#5** (low) — `per-window-ui.spec.ts`: updated the header comment and `sendNotification`'s TSDoc to describe the conditional routing rule.

**#32** (low) — `notification.service-shard.ts`: updated `startNotificationServiceShard`'s TSDoc the same way.

All three commits pushed. Full detail in the commit message on `d2a718f`.

timothy-mccormack:
Read all of these. Logged as-is except one: added a line to \`webViewId\`'s TSDoc noting it's the narrowest available key (a project with no open web view has nothing to route by) — small enough to fold in now while the field's still \`@experimental\`.

The rest (pre-existing first-editor-in-creation-order behavior, the un-synced \`focusSharedLayoutDefaultTab\` routing, the un-raised owner window, no C# \`WebViewId\`, no e2e coverage) are noted for follow-up, not this PR.

e2e-tests/tests/isolated/multi-window/per-window-ui.spec.ts
```diff
@@ -4,8 +4,9 @@
  * The multi-window architecture makes formerly app-singleton UI surfaces per-window by
  * construction: each window is a full renderer with its own OverlayHost, Toaster, toolbar, and
  * module state, web view iframes inherit their PARENT window's papi object, and the main process
- * routes generic dialog/notification/web-view requests to the focused window. Nothing pinned any of
- * that behaviour — this spec is the regression guard that does.
+ * routes dialog/notification/web-view requests to the window they name, or to the focused window
+ * when they name none. Nothing pinned any of that behaviour — this spec is the regression guard
+ * that does.
```
lyonsil: **Correction to my round-1 wording** - low · confirmed

The header now overclaims for dialogs: a dialog request cannot name a window.

Every route `dialog.service-router.ts` registers - `showDialog`, `selectProject`, and both `showAboutDialog` registrations - forwards through `getTargetDialogShard()`, and there is no `windowId` or `webViewId` field anywhere in the dialog models. Notifications and web-view opens can name a window; dialogs are unconditionally focus-routed, so the sentence promises a capability that does not exist. This is my text from round 1, and it was sloppier than the sentence it replaced.

**Suggested change** - keep dialogs on the focus-routed side of the sentence:

```suggestion
 * routes notification and web-view requests to the window they name, and everything else — dialogs
 * included, which cannot name one — to the focused window. Nothing pinned any of that behaviour —
 * this spec is the regression guard that does.
```

_Verification: all four dialog routes (`dialog.service-router.ts:228-255`) call `getTargetDialogShard()`, and a grep of `dialog-options.model.ts`, `dialog.service-model.ts` and `dialog.service-shard.model.ts` for `windowId`/`webViewId` returns no hits, so there is no path by which a dialog names a window. Suggestion applied to a worktree at `35179d3`: prettier, eslint and `tsc -p e2e-tests/tsconfig.json` all clean._

Same origin as the `sendNotification` comment below.

e2e-tests/tests/isolated/multi-window/per-window-ui.spec.ts
```diff
@@ -254,8 +255,9 @@ function showModalAlertViaWebSocket(prompt: string): Promise<unknown> {
 
 /**
  * Send a notification through the GENERIC notification service — the name the main process serves
- * with a service router that forwards `send` to the focused window. `duration: 0` means the toast
- * never auto-closes, so assertions cannot race an auto-dismiss; tests dismiss explicitly.
+ * with a service router that routes `send` to the window they name, or to the focused window when
+ * they name none. `duration: 0` means the toast never auto-closes, so assertions cannot race an
+ * auto-dismiss; tests dismiss explicitly.
```
lyonsil: **Correction to my round-1 wording** - low · confirmed

"they name" has no antecedent - `send` is singular, and the thing that names a window is the notification.

Unlike the header above this one is factually right, since it is scoped to notifications; it just reads as a grammar slip. My round-1 note said to change it "the same way" as the header, and the header's plural subject does not survive the move to a singular `send`.

**Suggested change**:

```suggestion
 * with a service router that routes `send` to the window the notification names, or to the focused
 * window when it names none. `duration: 0` means the toast never auto-closes, so assertions cannot
 * race an auto-dismiss; tests dismiss explicitly.
```

_Verification: applied to a worktree at `35179d3`; prettier and eslint clean, `tsc -p e2e-tests/tsconfig.json` clean._

src/shared/models/notification.service-model.ts
```diff
@@ -139,6 +140,11 @@ export interface PlatformNotification {
    * On an update (a `send` reusing an id that is still showing), any optional field you omit keeps
    * the value it had on the previous `send` for that id - omitting a field never clears it. Pass
    * the field explicitly to change it.
+   *
+   * The one exception is {@link webViewId}: which window a `send` runs in is decided in the main
+   * process before the renderer ever sees the notification to merge it, so omitting `webViewId` on
+   * an update does NOT keep routing to the window the original send resolved to - it always routes
+   * by the rules {@link webViewId} documents, using only what this call passed.
```
lyonsil: **The new exception paragraph stops one step short of the consequence** - low · confirmed

It states the mechanism correctly, but not what a caller actually sees - which is the part that would stop someone writing the bug.

An update-send that omits `webViewId` does not merely route elsewhere; it updates nothing. `trackedNotificationsById` is per-renderer module state, so the window it lands in holds no entry for the id: `trackedNotification` is `undefined` (no merge, so every omitted field takes its default rather than its previous value) and `existingToastId` is `undefined` (so Sonner opens a fresh toast instead of updating one). Meanwhile the original stays up in the window it was routed to. Two visible notifications, and the merge contract promised in the paragraph directly above silently does not apply.

**Suggested change**:

```suggestion
   * by the rules {@link webViewId} documents, using only what this call passed. An update that lands
   * in a different window updates nothing: that window has never seen the id, so it opens a second
   * notification with no merge applied, and the original stays up in the window it was routed to.
   * Pass the same `webViewId` on every `send` that shares an id.
```

**Important:** like the `webViewId` TSDoc, this text is republished into the generated `lib/papi-dts/papi.d.ts`. Run `npm run build:types` and commit the regenerated file after applying.

_Verification: mechanism read off `notification.service-shard.ts:73-77` (the merge, keyed on `trackedNotificationsById.get(notificationId)`) and `:112-115` (`existingToastId`), both per-renderer module state that a second window cannot see. Suggestion applied to a worktree at `35179d3`: prettier and eslint clean, `tsc -p ./tsconfig.json` clean, and `npm run build:types` produced exactly the matching `papi.d.ts` hunk. No caller is affected today - a sweep of all 33 `notifications.send(` sites shows the two `webViewId` senders pass no `notificationId` and no `notificationId` sender passes a `webViewId` - so this stays a doc fix on an `@experimental` field._

lyonsil (commented):
Re-checked the round-2 changes at `35179d3`. Three follow-ups, all documentation.

Two are corrections to wording I suggested in round 1 and got wrong - applying it faithfully was the right call, the text itself was the problem. The third sharpens the new `notificationId` note to say what a caller actually sees.

Every suggested change below was applied to a worktree at `35179d3` and checked: `prettier --check` and `eslint` clean on the touched files, `tsc -p ./tsconfig.json` and `tsc -p e2e-tests/tsconfig.json` clean.

e2e-tests/tests/isolated/multi-window/per-window-ui.spec.ts
timothy-mccormack: Fixed — dialogs stay on the focus-routed side of the sentence.

e2e-tests/tests/isolated/multi-window/per-window-ui.spec.ts
timothy-mccormack: Fixed — singular subject now matches singular verb.

src/shared/models/notification.service-model.ts
timothy-mccormack: Added the consequence sentence. Regenerated papi.d.ts.

lyonsil (approved):
:lgtm: - When you squash commit, please rewrite the squashed commit message to reflect the final state of the changes instead of appending all the individual messages from each commit.

> @lyonsil reviewed 11 files and all commit messages, made 1 comment, and resolved 12 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2780#-:-P169a5gE1xLo1yUgaca:b-rwg8yh)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2270, opened 2026-05-14
chore(pbr): exclude design-principles.mdx from remark --output

tombogle:
## Problem

`remark . --ext .mdx --output` (run as part of `npm run format` via the `platform-bible-react` workspace's `format:mdx` script) reformats JSX inside attribute values — specifically the `preview={...}` props used extensively in `design-principles.mdx` — by stripping manual indentation. This produces bogus whitespace-only diffs that creep into unrelated PRs.

There is no in-file mechanism to suppress remark's reformatting of specific JSX sections. `*.mdx` is already in `.prettierignore` (Prettier is not the culprit).

Since the prose content is relatively simple (short paragraphs, one table, a few bullet lists), it is human-reviewed whenever it changes, and the vast majority of the file's content is inside JSX components, reformatting with remark doesn't add much value to anyway. So the practical risk of letting markdown drift is low for this particular file.

## Fix

Add `lib/platform-bible-react/.remarkignore` listing `design-principles.mdx`. This is remark-cli's standard ignore mechanism — remark-cli sets `ignoreName: '.remarkignore'` so `remark . --ext .mdx --output` will skip files listed there.

## Related

- The `/review-paratext` skill was also updated ([ai-prompts PR #282](https://github.com/paranext/ai-prompts/pull/282)) to scope its format step to only changed files rather than running `npm run format` globally.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2270)

PR #1715, opened 2025-07-16
PT-1657 Vocabulary article

merchako:
Add a style guide article on vocabulary to Storybook

Motivated by
* [PT-1657 Collect principles and content for design system](https://paratextstudio.atlassian.net/browse/PT-1657)
* [My extensions → Your extensions #120](https://github.com/paranext/paratext-bible-internal-extensions/pull/120) 
  pt-3313-update-marketplace-wording-your
* [PT-2267: Help People get to Getting Started page #1712](https://github.com/paranext/paranext-core/pull/1712)
  pt-2267-help-getting-started

Fixes along the way
* Prettier: ignore MDXv2 since Prettier [doesn't support it](https://prettier.io/blog/2021/11/25/2.5.0#add-support-for-mdx-v2-comment-syntax-11563-by-wooorm)
* TODO: Format MDX with remark, as in
    ```package.json
     "format": "remark . -qfo && remark . -e mdx -u mdx -qfo && prettier . -w --loglevel warn && xo --fix",
    ```

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/1715)

PR #2266, opened 2026-05-13
fix: restore JSX indentation in design-principles.mdx

merchako:
## Summary

- Restores 2-space JSX indentation inside `preview={}` blocks in `design-principles.mdx`, which was accidentally stripped to flush-left by the shadcn/ui upgrade in #2207 (ac37edae).
- No functional change — Storybook renders identically either way.

## Context

This is the paranext-core half of a two-part change. The companion PR in `ai-prompts` adds a `## Storybook MDX Guidelines` section to `Code-Style-Guide.md` documenting this convention, including an explicit note for LLMs and editors not to reformat MDX JSX blocks.

The style proposed (2-space block form, root element at 2 spaces, closing `}` at column 0) matches what TJ used when the file was originally written. Primary goal is reducing merge conflicts; secondary goal is readability. Happy to adjust if a different style is preferred — the important part is that something is documented.

## Test plan

- [ ] Open Storybook (`cd lib/platform-bible-react && npm start`) and verify the Guidelines/Design Principles page renders correctly
- [ ] Confirm no TypeScript or lint errors

🤖 Generated with [Claude Code](https://claude.com/claude-code)

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2266)

PR #2784, opened 2026-09-08
Stop the version bump from reformatting theming.mdx

tjcouch-sil:
Follow-up to #2782. The `0.6.0-alpha.1` bump (#2783) carried an unrelated whitespace change to `lib/platform-bible-react/src/stories/guidelines/theming.mdx`. That change has been reverted on the bump branch; this PR fixes the cause.

**One line in `.remarkignore`.** This was originally two commits; the second — making `format:mdx:check` actually check anything — is now split out into #2799 so this one can go in quickly during the release. This PR stands entirely on its own.

## Why it changed

`npm run bump-versions` runs `npm run format`, which ends in `format:mdx` (`remark . --ext .mdx --output`). remark's MDX serializer strips the indentation of JSX inside attribute values, so it rewrites this page's `<ExampleBlock>` props:

```diff
   preview={
-  <input
-    value="this has manual colors"
+<input
+value="this has manual colors"
```

For `preview={...}` that is cosmetic. But the same pass also flattens ``code={`...`}``, a template literal Storybook renders verbatim as the code sample readers see, so remark was changing displayed content, not just source layout.

This is a known remark behavior in this repo — `.remarkignore` exists for exactly it, and already lists `design-principles.mdx` (added in #2270). `theming.mdx` uses the same component and needed the same entry. It has been drifting since it was created in #2180, seven days after `.remarkignore` was introduced.

Note this does **not** touch the UX-owned MDX content; it stops remark from rewriting it. `.remarkignore` is a remark-cli concern only, so Storybook rendering is unaffected.

## Verification

- **The bug is gone end to end**: `npm run format` — the exact command the bump runs — now leaves every tracked MDX file untouched. Before this change it rewrote `theming.mdx`.
- **Sweep, not spot-check**: ran remark against all 19 tracked MDX files under both configured roots. Exactly two differ from remark's output — `design-principles.mdx` (already ignored) and `theming.mdx`. A deterministic grep of every `package.json` plus `.erb/` and `.github/` confirms those two `remark` invocations are the whole surface. This would not catch a remark invocation built dynamically at runtime; there are none.
- **This line is load-bearing**: with the follow-up's real checker applied, removing this `.remarkignore` entry makes the check exit 1 and name `theming.mdx`; restoring it goes green. Verified both directions.
- `npm run typecheck:erb`, repo-wide `eslint`, `stylelint`, and `npm run format:check` all pass. C# and unit tests were not run: nothing here touches runtime code.

## Why nobody noticed for ~3.5 months

`format:mdx:check` is `remark . --ext .mdx` — remark with no `--output`. That reports plugin messages; it never compares remark's serialized output to the bytes on disk, so it exits 0 on a file `format:mdx` would rewrite. It has been that way since #1715 (2025-07-18). CI ran it on every PR and could not have caught this.

So the drift sat in the repo until someone ran a full `npm run format` — which, in practice, is the release version bump. **That gap is fixed in #2799, not here.** Without it, the next JSX-heavy MDX page will drift the same way and surface in a future bump.

## Unrelated observation, not fixed here

`format:mdx` also rewrites MDX inside `storybook-static/` when a developer has built Storybook locally. It is gitignored build output so nothing is harmed, but adding `storybook-static` to `.remarkignore` would save the pointless writes. Left out to keep this PR to the reported problem.

AI-assisted — [session 1](https://claude.ai/code/session_019EmD5if9cG2ZwaQCzkhFan)

https://claude.ai/code/session_01RH1JMqjK2KLBHhe3xLaEPY

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2784)

### DIFF
--- a/lib/platform-bible-react/.remarkignore
+++ b/lib/platform-bible-react/.remarkignore
@@ -1,3 +1,4 @@
 # remark's MDX serializer reformats JSX inside attribute values (e.g. preview={...} props),
 # removing intentional indentation. Files listed here are excluded from remark --output.
 src/stories/guidelines/design-principles.mdx
+src/stories/guidelines/theming.mdx


tjcouch-sil:
Absorbed into #2266, which now carries this commit plus the other MDX files remark was rewriting. Leaving the branch in place because #2799 is stacked on it; retarget #2799 to `main` once #2266 merges.

PR #2786, opened 2026-09-09
PT-4527: Let an extension list its project extension-data qualifiers

siltomato:
# Code Review Summary

**Branch**: pt-4527-pdp-extension-data-enumeration

**Base**: origin/main

**Date**: 2026-09-09

**Files changed**: 24

## Overview

This branch lets an extension discover which of its own extension-data documents exist in a project,
instead of guessing at names. Today an extension can read a document by name and write one by name, but
it has no way to ask what it has already stored, and under the old implementation the guess itself was
destructive, because reading an absent document created it. The new method, `listExtensionDataQualifiers`,
is implemented in the C# Paratext PDP and in the `hello-rock3` sample engine and covered by C# and
TypeScript tests. `lib/papi-dts/papi.d.ts` is regenerated to match.

**Shape, after the first human review.** The method lives on a **new `projectInterface`,
`platform.extensionDataEnumeration`**, where it is required. It is not a member of `platform.base` in any
form. Not every PDP can enumerate its extension data (one over a remote store may not be able to), so a PDP
that can advertises the interface in its `projectInterfaces` and one that cannot leaves it off, and a
consumer learns which from project metadata rather than by calling and failing. The PR's first draft
declared the method optional on the engine and required on the consumer inside `platform.base`; the
platform API owner rejected that as too messy for the API and offered three ways out, and the author
chose this one. There is no breaking change in either direction. A new architecture-decision entry,
requested in that review, records the standing rule: PDP methods are never optional.

Two behavior corrections ride along. Reading extension data no longer creates the document it was asked
for, which previously left an empty file that Send/Receive commits to every clone with no delete API to
take it back. And the directory walk behind the listing was rebuilt on `FileSystemEnumerable` with
separate include and recurse predicates, so the listing reports exactly what `getExtensionData` can read
while never leaving the project.

## API Changes

TypeScript, public surface:

- **Added `projectInterface` `platform.extensionDataEnumeration`.** In `src/declarations/papi-shared-types.ts`: a new `ProjectDataProviderInterfaces` entry keyed by `PROJECT_INTERFACE_PLATFORM_EXTENSION_DATA_ENUMERATION`, typed `IExtensionDataEnumerationProjectDataProvider` = `IProjectDataProvider<ExtensionDataEnumerationProjectDataTypes> & WithProjectDataProviderEngineExtensionDataEnumerationMethods`. Obtained with `papi.projectDataProviders.get('platform.extensionDataEnumeration', projectId)`, which throws for a project that does not advertise it.
- **Added** in `src/shared/models/project-data-provider.model.ts`: the constant `PROJECT_INTERFACE_PLATFORM_EXTENSION_DATA_ENUMERATION`; `ExtensionDataListScope` (`Pick<ExtensionDataScope, 'extensionName'>`); `ExtensionDataEnumerationProjectDataTypes` (`{}` — the interface has no data type, so no `get*`/`set*`/`subscribe*` trio); and `WithProjectDataProviderEngineExtensionDataEnumerationMethods`, the single declaration of `listExtensionDataQualifiers(scope: ExtensionDataListScope): Promise<string[]>` that both the engine side and the consumer type share, the way `platform.base` already shares `getExtensionData`. The types are re-exported from `@papi/core` so an engine can `implements` the `With…Methods` type.
- **`platform.base` is unchanged in both directions.** `WithProjectDataProviderEngineExtensionDataMethods` and `IBaseProjectDataProvider` are byte-identical to main. No existing engine or consumer is affected.
- **Changed** `ExtractDataProviderDataTypes` (`src/shared/models/extract-data-provider-data-types.model.ts`): a provider with no data types now resolves to `{}` instead of falling through to `never`. The `infer` branches cannot recover `{}` on their own because the `get*`/`set*`/`subscribe*` mapped types over no keys simplify away; `never` there is not harmless, since `keyof never` is every key, so a hook typed over it would accept any data-type name. Existing interfaces are unaffected (verified by compile-time probe: `platform.base` still extracts as `MandatoryProjectDataTypes`).
- **Regenerated** `lib/papi-dts/papi.d.ts`.

.NET `platform.base` PDP wire surface and project metadata:

- **Added** `ProjectInterfaces.EXTENSION_DATA_ENUMERATION` (`"platform.extensionDataEnumeration"`), published for **both** published and unpublished Paratext projects in `LocalParatextProjects.cs`, since published projects are readable and their extension data can be listed as well as read. Pinned by `LocalParatextProjectsTests`.
- **Added** the JSON-RPC function `listExtensionDataQualifiers`, registered in `ParatextProjectDataProvider.GetFunctions()` — the PDP that advertises the interface — not in the abstract `ProjectDataProvider`, which models `platform.base` for every subclass and is unchanged apart from two XML-doc fixes. The implementation is scoped to the calling extension's own data directory.
- **Changed** `getExtensionData` so it no longer creates the document it reads. An absent document still returns `""`, unchanged from before, so there is no wire-format or caller-visible contract change. `getExtensionData` and `setExtensionData` are otherwise byte-for-byte unchanged from main.

Internal, not public API: `IProjectStreamManager.GetExistingDataStreamNames` gained an optional
`underPath` parameter and now returns `/`-separated, ordinal-sorted names relative to that path.

Extension engines: `hello-rock3` claims `platform.extensionDataEnumeration` and implements the method
over its in-memory store. `platform-lexical-tools` is untouched (an earlier draft gave it a throwing stub;
under the interface shape an engine that cannot enumerate simply does not claim the interface). No
extension type-declaration files, `lib/platform-bible-react`, or `lib/platform-bible-utils` changed.

## Changes after the first human review

tjcouch-sil reviewed the architecture (not the implementation) on 2026-09-09 and requested changes:

- **"Sort-of optional methods on `projectInterface`s are too messy for our API."** Three options were offered: drop the method and ship only the `getExtensionData` fix; make it required everywhere with a `#platform-changes` announcement; or move it to a new `projectInterface` and make it required there. The author chose the third. Implemented as described under API Changes. The reviewer's reasoning is recorded, with the author's, in `adr-pdp-enumerates-extension-data-qualifiers` (amended) and the plan's frozen record.
- **"Add an ADR entry that says we do not put optional methods on PDPs."** Added as `adr-pdp-methods-are-never-optional` (Accepted), inserted in byte-order slug position, and promoted into `Paranext-Core-Patterns.md` as the section "A capability not every PDP can serve is its own `projectInterface`", which replaces the earlier "Adding an optional method to `platform.base`" section.
- **The `getExtensionData` create-on-read bug fix** was independently endorsed ("would be good to fix this in any case") and is unchanged.
- **Found while reworking:** the `ExtractDataProviderDataTypes` gap described under API Changes. Fixed at the source rather than worked around at the new interface.

## Findings

Findings from the four automated analysis passes, with what happened to each. One Important finding and
one Minor finding were later resolved outright by the move to a separate `projectInterface`; they are
marked accordingly.

None.

- [x] **`AttributesToSkip = FileAttributes.ReparsePoint` filtered files, not just directories.** The flag had been added to stop recursion through symlinks and junctions, but .NET applies `AttributesToSkip` to every enumerated entry, and cloud-sync placeholder files carry that attribute (OneDrive Files On-Demand, Dropbox online-only). For a project in a synced folder the listing could omit qualifiers that `getExtensionData` reads fine. _(fixed during review: rebuilt as a `FileSystemEnumerable` with `AttributesToSkip = 0` and two predicates — include every non-directory whatever its attributes; never descend into an entry carrying `ReparsePoint`)_
- [x] **The symlink half of that behavior had no test.** _(fixed during review: `GetExistingDataStreamNames_FileThatIsASymbolicLink_IsListed` and `..._DirectoryThatIsASymbolicLink_IsNotDescendedInto`, both mutation-checked; the link helper was promoted to a shared `FileSystemLinks` utility that skips where the OS refuses links, and CI's macOS and Ubuntu legs run it for real)_
- [x] **The single-directory guard changed behavior on the existing `getExtensionData`/`setExtensionData` methods.** _(fixed during review: the escape check is list-only, as `EnsureExtensionNameStaysInItsOwnDirectory`; `GetExtensionDataRoot` is pure composition; nested names such as `acme/tools` stay valid everywhere, pinned by `ExtensionData_NestedExtensionName_WritesListsAndReadsBack`)_
- [x] **That guard was exercised only through the list path.** _(fixed during review: with the guard deliberately list-only, list-path coverage is the complete coverage; the seven-case `..._ExtensionNameEscapesItsDirectory_Throws` matrix pins the rejections and the nested-name round trip pins that get and set still accept what they always accepted)_
- [x] **The required consumer member was a type-level breaking change for implementers of `IBaseProjectDataProvider`.** The author first defended keeping it required, with disclosure, because a remote PDP cannot be feature-detected. _(fixed after the first human review: the method moved to its own `projectInterface`, so `IBaseProjectDataProvider` is unchanged and the break no longer exists. Discovery is by interface advertisement, which works on both the remote-proxy and local-proxy paths that made property probing unreliable.)_

Author response: the author treated four of the five as real defects and had them fixed during the interview, choosing the more conservative option each time, and defended the fifth with a concrete mechanism. The human review then reopened the fifth at the architecture level and it was resolved by changing the shape rather than by disclosure.

### Minor — Consider

- [x] **Missing blank line after the new early return** in `GetExtensionData`. _(fixed during review)_
- [x] **The architecture-decisions entry miscounted the alternatives it listed.** _(fixed during review)_
- [x] **`a..b` as an extension name is rejected by the stream manager with `ArgumentException`, a second exception type the docs did not mention.** _(fixed during review: documented, noting that get and set reject the same name identically)_
- [x] **The `hello-rock3` rationale comment ran nine lines on a four-line method.** _(fixed during review: kept, because it defends a non-obvious combining-mark edge case, and given a pointer to the repo string rule whose carve-out names this method)_
- [x] **Engine-side and consumer-side TSDoc duplicated a shared factual core.** The author first chose to leave both standing-alone blocks, on AI-consumability grounds. _(fixed after the first human review: the interface shape declares the method once, on `WithProjectDataProviderEngineExtensionDataEnumerationMethods`, and the consumer type intersects it, exactly as `platform.base` does for `getExtensionData`; consumer-specific discovery guidance sits on `IExtensionDataEnumerationProjectDataProvider`)_
- [ ] ~~`hello-rock3`'s `listExtensionDataQualifiers` is untested.~~ _(Author: `hello-rock3` is a demo extension that exists to exercise the PAPI surface by hand, not a shipped feature, and it has no test files at all today. Adding the repo's first test there is a separate decision from this PR.)_
- [ ] ~~No test passes a backslash `underPath`, though the interface doc says either separator is accepted.~~ _(Author: no caller sends one — the provider composes the listing root with a `/` literal — and the forward-slash form is already the tested case on both platforms, which on Windows is exactly the "foreign separator in, `/` out" case. On Unix the platform separator is `/`, so there is no second case.)_
- [ ] ~~The new failures carry no `PlatformError` code, so a TypeScript caller cannot tell "fix your scope" from "retry later" without matching on message text.~~ _(Author: deferred deliberately. The mechanism exists and is in production use in ManageBooks, but coding only the new method would give callers codes on the listing and bare exceptions on the very next `get` of a listed qualifier; doing it properly means all three extension-data methods and a judgment call on which code an unreachable project deserves. Adding a code later is additive. Recorded as a follow-up candidate; no ticket filed, by the author's instruction.)_

Author response: the author asked for the first four to be fixed and dismissed the last three with reasons; the fifth resolved itself with the shape change.

#### Shared Regions Modified

#### Extension Config Changes

- [n/a] `extensions/src/hello-rock3/src/models/hello-rock3-project-data-provider-engine.model.ts` — extension source, not config or build; no propagation needed

Storybook: no new React components, so no stories required. Localization: no user-visible strings added;
the new C# `InvalidDataException` messages are developer-facing and follow the pre-existing hardcoded
pattern in the same method. Build config and process: no changes to `package.json`, `tsconfig*`, ESLint,
webpack, `.erb/`, or `.github/workflows/`.

## Positive Observations

- **The purpose is delivered and pinned, not just implemented.** `GetExtensionData_NeverWritten_ReadsEmptyAndCreatesNothing` asserts both the unchanged `""` answer and the absent side effect, so a regression to `createIfNotExists: true` fails rather than passing on the return value alone. Incidentally this removes an ungated project-data mutation that the Send/Receive write-gate rule would otherwise have required wrapping.
- **Capability discovery uses the platform's own mechanism.** A project that can enumerate says so in `projectInterfaces`; nothing is probed, nothing is caught. The published-interfaces tests pin that Paratext projects advertise it, published or not.
- **`list*` naming sidesteps the `get`/`set`/`subscribe` magic-prefix machinery**, and a TypeScript test pins that a `list___` method with neither a matching setter nor an ignore decorator is still exposed over the network. A C# test pins the wire registration tuple.
- **Backward compatibility for existing engines is pinned, not assumed.** The `platform.base` registration guard is deliberately left checking only `getExtensionData`/`getSetting`, and a test proves an engine that does not claim the new interface still registers.
- **The listing and the readers share one path-composition helper**, so every name the listing returns is a name `getExtensionData` can read — a round-trip that holds by construction. The helper's doc states that it deliberately validates nothing, and why.
- **The comments explain the rejected alternatives, not just the code.** The walk carries a list of the simpler forms that look right and are wrong, each naming the test that goes red if someone tries it.
- **Sort order agrees across implementations**: C# `StringComparer.Ordinal` and JavaScript's default sort produce the same sequence.
- **File-backed coverage exists where the in-memory double cannot reach**: separator normalization, both platforms' spelling of "hidden", missing project storage versus absent sub-path, path escape, and a byte-for-byte round trip whose fixtures use BOM-less UTF-8 so the empty-document case is a genuine zero-byte file.
- **Every new test was mutation-checked**, and the type change was probe-checked at compile time rather than assumed.
- **Docs are durable and were verified against the live tree**: both decision entries sit in correct byte order, the promoted patterns section cites file and symbol with no line numbers, and the plan's frozen record was brought up to date rather than left describing a shape that never shipped.

## Interview Notes

**Stated purpose**, in the author's words:

> There was no way to list extension data files. You had to know the name to use `getExtensionData`.
> Guessing a name created the file. An extension may know the name scheme but not the actual file name,
> which could happen with files created on other machines and synced (for example,
> `decisionLedger-<machineId>.json`)
>
> Listing is scoped to the one extension's own dir. Directory symlinks are not followed. Names come back
> in the form that `getExtensionData` accepts.

**Key decisions and reasoning**:

- **Own `projectInterface`, not a member of `platform.base`** — chosen by the author from the reviewer's three options, for the reasons the decision entries record: no breaking change in either direction, discovery by metadata rather than by failure, and a consumer shipped separately from the platform (the checking-assistant extension's Send/Receive work is the motivating consumer) can fall back cleanly on a platform that predates the capability.
- **List-only escape validation**, chosen explicitly over the analyzer's recommended shared guard, on the principle that the change must not break existing functionality.
- **`FileSystemEnumerable` with two predicates** over a hand-rolled walk or `Directory.GetFiles`, with the rationale written up for future AI agents specifically.
- **Native `String` methods in the `hello-rock3` engine** because the keys are composed and looked up by exact string equality; the repo string rule records this carve-out.
- **No follow-up tickets filed** for the deferred items, by the author's explicit instruction.

**Unresolved items**: none.

**Areas the author did not understand**: none identified. The author answered each finding with reasoning rather than acceptance, and made the architecture call after asking for the trade-offs of each option.

## In-Review Quality Check

All checks were run on the final tree: the reworked branch rebased onto `origin/main` at `b407ba5293f`
(clean rebase, no conflicts), with `lib/papi-dts/papi.d.ts` regenerated afterwards and confirmed
byte-identical to the committed file.

| Check | Result |
| --- | --- |
| `dotnet test` (c-sharp-tests) | 1741 passed, 0 failed, 13 skipped |
| `npm run typecheck` (all workspaces, including `hello-rock3` against the regenerated types) | clean |
| `npm run lint` | 0 errors, 1 warning (pre-existing on main) |
| `npm test` | 3705 passed, 2 skipped, 1 failed — the failure is pre-existing on main, see below |
| Prettier and CSharpier on the changed files | clean |
| Compile-time probe of the new interface's types | `platform.base` extracts as `MandatoryProjectDataTypes`; `platform.extensionDataEnumeration` extracts as `{}`; an engine claiming both keeps its `platform.base` methods with no index signature |

The 13 skipped C# tests include the two symbolic-link tests, which skip themselves on a Windows machine
that cannot create symlinks; their recursion behavior was verified separately with a directory junction,
and CI's macOS and Ubuntu legs run them for real.

**The one failing test is not from this branch.** `.erb/scripts/electron-builder-snap-config.test.ts`
fails its "mounts exactly one GNOME platform plug" case, expecting `gnome-42-2204` and finding
`gnome-3-28-1804`. It was verified failing identically on `origin/main` with this branch checked out
nowhere. It arrived with PT-4496 (#2747), which mounts the core22 GNOME platform snap; the assertion
reads state that only a fresh `npm install` applies, so a local `node_modules` predating that commit fails
it. This branch touches nothing under `.erb/` and does not modify `package.json`.

## Suggested Review Focus

- [ ] **The new `projectInterface` and its name.** `platform.extensionDataEnumeration` sits in the `platform.` namespace because extension data is a `platform.base` concept. Confirm the name and that a read-only capability belongs in the published-projects list as well as the unpublished one.
- [ ] **`ExtractDataProviderDataTypes` gained a branch.** It is a core type utility; the change is narrow (only a provider with no data types is affected, and it now yields `{}` instead of `never`) and probe-verified, but worth a second pair of eyes.
- [ ] **`adr-pdp-methods-are-never-optional` is marked Accepted** on the strength of the reviewer's explicit request. Confirm the wording states the rule as the team wants it enforced.
- [ ] **The escape check lives only on the listing.** Confirm the reasoning: a bad extension name on get or set reaches nothing a caller could not reach by naming the other extension outright, so a shared guard would close no access while breaking callers.
- [ ] **The create-on-read fix rides along.** Confirm `""` for "absent" is the right answer rather than surfacing `undefined`, and that folding it into this PR is acceptable.
- [ ] **`PlatformError` codes across the extension-data trio** — deferred on scope grounds; decide whether all three methods should carry codes.

AI-assisted — Claude Fable 5.1 (implementation, review passes, the post-review rework, and this summary), reviewed and directed by the author throughout.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2786)


tjcouch-sil (changes requested):
Disclaimer: I looked at the architectural structure here, but I did not look at the actual implementation; once we get the architectural design settled, I'd prefer for someone else to review the implementation details if we can make that happen.

Looks like you found a bug in the ParatextPDP's `GetExtensionData`; getting should not create the file if it doesn't exist; would be good to fix this in any case.

`listExtensionDataQualifiers` seems like a nice addition conceptually! But the particulars of adding this need some ironing out. Creating sort-of optional methods on `projectInterface`s is too messy for our API; it would be too hard to tell whether a PDP does or does not have this method whose type is optional when making a PDP but labeled as required when consuming the PDP. Rather, let's shift gears and do **one** of the following instead:
1. Consider not adding `listExtensionDataQualifiers`; this is quite a large and wide API change. Is it possible to restructure your code such that you don't require `listExtensionDataQualifiers` but are satisfied just with the ParatextPDP bug fix? Then we can defer.
2. Make `listExtensionDataQualifiers` required (e.g. in `project-data-provider.model.ts`), make sure every PDP we know about in all our repos is updated to have this method, and make an announcement about this breaking change in `#platform-changes` in Discord.
3. Move `listExtensionDataQualifiers` to a new `projectInterface` and make it required in that `projectInterface`. That way, there's no breaking change you have to comb all our code for and announce, and consumers can tell which PDPs support listExtensionDataQualifiers` by requesting that `projectInterface` rather than just guessing.

Let me know if you have any questions or want to discuss any of this. Thanks!

By the way, if you want to contact me or you make changes and want me to take another look, please message me on Discord. My email inbox is a mess, and I miss a lot of GitHub notifications on there.

tjcouch-sil (commented):
Oh also, another thing I forgot: could you please add an ADR entry that says we do not put optional methods on PDPs? If a method needs to be "optional" in some sense, it should just be required on a different `projectInterface`. Thanks!

siltomato (commented):
Thanks, TJ.  Option 3 seems good to me.  No breaking changes.

> @siltomato made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2786#-:-P169nNe5YzWuocaC7py:b-cakvd8)* status: 0 of 21 files reviewed, all discussions resolved.

siltomato:
Downstream consumer, for sequencing: **paranext/paratext-assistant#135** has a design that this method makes materially better.

That PR carries gloss data on Send/Receive and found that `shared/**` files get `SimpleProjectFileMerger` — whole-file last-write-wins by timestamp, no dialog, either side can lose. Its conclusion was that the only safe mechanism is for two machines never to write the same file. It could apply that everywhere except the gloss index, and said why:

> Per-machine index documents cannot work here: their union needs a list of machines, which is one more shared document with the same problem.

`listExtensionDataQualifiers` is precisely the missing piece: per-machine index documents become enumerable (`glossIndex:<machineId>`) and their union needs no registry, so there is no shared file for two machines to race. #135 currently ships a machine-local-record fallback instead, which makes losses transient rather than impossible.

Noting the interface shape matters to that consumer: since this lives on `platform.extensionDataEnumeration` rather than `platform.base`, the downstream design is a capability check with the fold-back retained for PDPs that do not advertise it — which seems right, and is easier to build against than an optional method that fails at call time. Linked from #135 as well.

c-sharp/Projects/LocalParatextProjects.cs
```diff
@@ -104,6 +104,7 @@ internal class LocalParatextProjects : IDisposable
     private static readonly List<string> s_paratextPublishedProjectInterfaces =
     [
         ProjectInterfaces.BASE,
+        ProjectInterfaces.EXTENSION_DATA_ENUMERATION,
```
tjcouch-sil: Resource extension data is broken, and that predates this PR, but advertising enumeration on published projects makes it visible.

`RawDirectoryProjectStreamManager` does plain disk I/O relative to `ProjectDetails.HomeDirectory`, which is `scrText.Directory`. For a `ResourceScrText` that's the folder *containing* the .p8z (`ResourceScrText.CreateFileManager`: `sourceDirectory = Path.GetDirectoryName(zipPath)`). So:

- Every resource in `_Resources` shares one `shared/platform.bible/extensions/...` bucket, and every resource in `_resourcesById` (where PT9 puts a resource whose short name is already taken) shares a second one. Listing extension `x` on resource A returns documents written for every other resource in the same folder.
- Nothing bundled inside the .p8z is ever read. A resource should be able to ship extension data in its zip, and that's where get and list should be looking.
- `setExtensionData` isn't refused on a resource, so it writes loose files into that shared folder.

(Projects are fine, including duplicate short names under `_projectsById`, since each has its own directory.)

The fix looks like a stream manager over the resource's `ScrText.FileManager`: `ZippedProjectFileManagerBase` already has `Exists`, `ProjectFiles`, `ProjectDirectories` and `OpenFileForRead`, which covers get and list, and `IsWritable` is false, so set should refuse. Your call whether to do that rework in this PR or file a Jira ticket for it and leave it for later.

c-sharp/Projects/RawDirectoryProjectStreamManager.cs
```diff
@@ -29,36 +29,76 @@ public void Initialize() // TODO: This doesn't seem to be used
...
-        }
-
-        return retVal.ToArray();
+            // Files are streams and directories are not. Without this predicate the enumeration
+            // yields directories too - it is not Directory.GetFiles.
+            ShouldIncludePredicate = (ref FileSystemEntry entry) => !entry.IsDirectory,
+            // Consulted only for directories: never descend into a symlink or junction. A FILE that
+            // is a link is still included above, because GetDataStream reads through it.
+            ShouldRecursePredicate = (ref FileSystemEntry entry) =>
+                (entry.Attributes & FileAttributes.ReparsePoint) == 0,
```
tjcouch-sil: This skips every directory carrying `ReparsePoint`, but the comment above keeps `ReparsePoint` *files* in the listing because cloud-sync placeholders (OneDrive Files On-Demand, Dropbox online-only) carry it. Placeholder directories carry it too, so if that premise holds, a project folder inside OneDrive (e.g. Documents redirected by Known Folder Move) would never be descended into past a placeholder directory: `byMachine/ledger/abc.json` would be missing from the list while `getExtensionData` still reads it.

I haven't confirmed whether Windows exposes placeholder attributes to this process (depending on the process's placeholder compatibility mode it may disguise them); if it doesn't, neither this nor the file-level reasoning applies. If it does, excluding only real links fixes it:

```csharp
ShouldRecursePredicate = (ref FileSystemEntry entry) =>
    (entry.Attributes & FileAttributes.ReparsePoint) == 0
    || new DirectoryInfo(entry.ToFullPath()).LinkTarget == null,
```

`LinkTarget` is non-null for symlinks and junctions but null for other reparse points such as cloud placeholders, and the attribute check first keeps the extra lookup off ordinary directories.

c-sharp/Projects/ParatextProjectDataProvider.cs
```diff
@@ -405,8 +413,6 @@ public override bool SetExtensionData(ProjectDataScope scope, string data)
         if (string.IsNullOrEmpty(scope.DataQualifier))
             throw new InvalidDataException("Must provide a data qualifier");
 
-        scope.ProjectID = ProjectDetails.Metadata.Id;
-
         Stream? dataStream =
             GetExtensionStream(scope, true)
```
tjcouch-sil: Not introduced here, but it's the same problem this PR fixes for reads: `GetExtensionStream(scope, true)` creates the document before the write lock is taken. If `RunWithinLock` can't obtain the lock (or it isn't active), it throws after the empty file already exists, and `dataStream` is never disposed (on Windows the file stays locked until GC). Send/Receive can then commit the zero-byte document.

Suggest opening the stream inside the `RunWithinLock` callback, in a `using`, so nothing is created unless the write actually happens. It'd be great to do it in this PR since it's the same fix you're making for reads, but fine to defer to a follow-up if you'd rather keep this one scoped.

c-sharp-tests/Projects/ParatextDataProviderTests.cs
```diff
@@ -265,5 +265,214 @@ public void SetAndGetExtensionData_SavesAndGetsData()
...
+            DummyParatextProjectDataProvider provider =
+                new(PdpName, Client, _projectDetails, ParatextProjects);
+
+            var data = provider.GetExtensionData(
+                new ProjectDataScope { ExtensionName = "myExtension", DataQualifier = "never.json" }
+            );
+
+            Assert.Multiple(() =>
+            {
+                // "" rather than null: the answer an absent document has always given, so no caller
```
tjcouch-sil: Nit (`.claude/rules/code-quality/forward-facing-comments.md`): "has always given" and "only the side effect is gone" describe the change rather than the code, so they won't mean anything to someone reading this test after merge. Maybe:

```csharp
// "" rather than null: an absent document reads the same as an empty one
```

c-sharp/Projects/ParatextProjectDataProvider.cs
```diff
@@ -428,17 +434,107 @@ public override bool SetExtensionData(ProjectDataScope scope, string data)
...
+    /// </exception>
+    /// <exception cref="ArgumentException">
+    /// The ExtensionName composes into a stream path the project's
+    /// <see cref="IProjectStreamManager"/> refuses — a name containing ".." anywhere, even without a
+    /// separator ("a..b"), is rejected before the filesystem is touched — exactly as
+    /// <see cref="GetExtensionData"/> and <see cref="SetExtensionData"/> reject the same name.
+    /// </exception>
+    public string[] ListExtensionDataQualifiers(ProjectDataScope scope)
+    {
+        if (string.IsNullOrWhiteSpace(scope.ExtensionName))
```
tjcouch-sil: Listing rejects a whitespace-only extension name (`IsNullOrWhiteSpace`), but `GetExtensionData`/`SetExtensionData` use `IsNullOrEmpty` and accept it. Could all three use the same check? A whitespace-only name isn't a real extension name, and with Windows' trailing-space handling (noted on `EnsureExtensionNameStaysInItsOwnDirectory`) what get/set do with one is platform-dependent.

c-sharp/Projects/ParatextProjectDataProvider.cs
```diff
@@ -428,17 +434,107 @@ public override bool SetExtensionData(ProjectDataScope scope, string data)
...
+            throw new InvalidDataException("Must provide an extension name");
+        EnsureExtensionNameStaysInItsOwnDirectory(scope.ExtensionName);
+
+        // Scoped to this extension's own data so a project's thousands of other files are never
+        // walked, and so no other extension's layout is exposed
```
tjcouch-sil: Small wording thing: "so no other extension's layout is exposed" reads as if the listing isolates extensions from each other, but any caller can pass any `extensionName` here, same as get/set, since extension data has no access control. Scoping the walk to the named extension's directory is right; I'd just drop that clause (or say "so the listing is confined to the named extension's data") so no one relies on it as an isolation guarantee.

.context/plans/pt-4527-pdp-extension-data-enumeration.md
```diff
@@ -0,0 +1,396 @@
+# PT-4527 — Let an extension list its project extension-data qualifiers
```
tjcouch-sil: Is this 396-line plan meant to be committed? `.context/plans/` has a couple of precedents on main, so it's not out of line, but it's a frozen record full of file:line citations against `origin/main` that will go stale quickly. If it's staying, fine; if its main purpose was driving the implementation, consider leaving it out, since the ADR entries already carry the durable decisions.

tjcouch-sil (commented):
A few comments, most important first: resource extension data (on `LocalParatextProjects.cs`), then reparse-point directories, the create-before-lock in `SetExtensionData`, and some smaller items.

PR #2787, opened 2026-09-09
ux guidelines for providing help

Sebastian-ubs:
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2787)

### DIFF
--- a/lib/platform-bible-react/src/stories/guidelines/providing-help.mdx
+++ b/lib/platform-bible-react/src/stories/guidelines/providing-help.mdx
@@ -1,5 +1,11 @@
 import { Meta } from '@storybook/addon-docs/blocks';
 import LinkTo from '@storybook/addon-links/react';
+import { Info } from 'lucide-react';
+import { Button } from '@/components/shadcn-ui/button';
+import { Label } from '@/components/shadcn-ui/label';
+import { RadioGroup, RadioGroupItem } from '@/components/shadcn-ui/radio-group';
+import { Tooltip, TooltipProvider, TooltipTrigger, TooltipContent } from '@/components/shadcn-ui/tooltip';
+import { ExampleBlock, ExampleBlockGroup } from '@/storybook/example-block.component';
 
 <Meta title="Guidelines/Providing Help" />
 
@@ -9,20 +15,23 @@ How to explain what a control does, when to show more context, and where deeper 
 Reach for the lightest option that still answers the user's question — extra text competes with
 the UI for attention.
 
-We do **not** ship PT9-style embedded "guides" (numbered explanations anchored to the UI). For
-anything past a sentence or two, link out to application help (see below).
+We do **not** ship Paratext 9-style embedded "guides" (numbered explanations anchored to the UI).
+For anything past a sentence or two, link out to application help (see below).
 
 ## The ladder
 
-Pick the first row that fits. Move down only when the row above cannot carry the message.
+Rows are ordered from lightest to heaviest in the space and attention they cost. Pick the first
+row that fits, and move down only when the row above cannot carry the message — either because it
+is too short, or because the message must be visible at a glance and a hover-only surface would
+never show it.
 
-| Surface                    | Use for                                                                | Format               |
-| -------------------------- | ---------------------------------------------------------------------- | -------------------- |
-| **Visible label**          | The primary meaning of a control                                       | 1–3 words            |
-| **Tooltip on the control** | Icon buttons (required); short hints on other action buttons as needed | Short phrase (ideally 1-3 words) |
-| **Label subtext**          | A brief clarification that must be visible at a glance                 | One `muted` line     |
-| **Info icon button**       | An explanation too long for a control tooltip, or optional context     | 1–2 sentences        |
-| **Application help**       | Concepts, workflows, anything more than a paragraph                    | Online (TBD, Nov 2026) |
+| Surface                    | Use for                                                                | Format                              |
+| -------------------------- | ---------------------------------------------------------------------- | ----------------------------------- |
+| **Visible label**          | The primary meaning of a control                                       | 1–3 words                           |
+| **Tooltip on the control** | Icon buttons (required); short hints on other action buttons as needed | Short phrase (ideally 1–3 words)    |
+| **Label subtext**          | A brief clarification that must be visible at a glance                 | One `muted` line                    |
+| **Info icon button**       | An explanation too long for a control tooltip, or optional context     | 1–2 sentences, or a dialog for more |
+| **Application help**       | Concepts, workflows, anything more than a paragraph                    | Online article                      |
 
 ## Tooltips on action buttons
 
@@ -36,9 +45,9 @@ and when a popover is needed instead.
 
 ## Label subtext
 
-A single muted line directly below an option's label or below a headline. Use it when the clarification is essential
-enough that a user should read it without hovering — for example, disambiguating two similarly
-named settings, or naming a non-obvious consequence.
+A single muted line directly below an option's label or below a headline. Use it when the
+clarification is essential enough that a user should read it without hovering — for example,
+disambiguating two similarly named settings, or naming a non-obvious consequence.
 
 Use sparingly. If every option in a form carries subtext, the form gets loud and the important
 ones stop standing out. Prefer a clearer label first; add subtext only when a clearer label isn't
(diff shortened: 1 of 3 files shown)


katherinejensen00 (changes requested):
Nice addition — this fills a real gap; no other Guidelines page covers subtext, info buttons, or where deep help lives.

Formatting

- Running npm run format in lib/platform-bible-react rewrites this file: remark re-pads the third column of the ladder table and drops the trailing blank line at EOF. Worth running before merge — format:mdx:check doesn't diff, so CI won't flag it and it'll land as noise in someone els
- "PT9-style" — the Guidelines/Product Names page sets the short name as "Paratext 9", and every other guidelines page uses that. "PT9" only    appears in .tsx dev
shes: "1–3 words" and "1–2 sentences" use en dashes, but "(ideally 1-3 words)" uses a hyphen.

Content

- "The info button opens one of: A tooltip" — our Tooltip is hover/focus-only (Radix, role="tooltip"), so nothing "opens" it by click. Also worth a word on touch: if the tooltip is the entire payloadof an info button, a
- The ladder row for "Info icon button" says the format is "1–2          sentences", but the  open a dialog withparagraphs, structure, examples, and links. The row undersells it — maybe "1–2 sentences, or a dialog when it needs more".
- Related: Tooltips says "Keep text short… a tooltip is a hint, not a sentence" and uses a one-sentence tooltip as its avoid example, while this page authorizes 1–2 sentences on info buttons. Tooltips does show valid one-sentence i compatible — but thetwo pages read as contradictory side by side. A one-line carve-out on the Tooltips page would settle it.                                     "Pick the first row n the row above cannotcarry the message" ranks tooltip above label subtext, but the subtext row's criterion is "must be visible at a glance" — which a tooltip can never satisfy. It's logically consistent, but a skimmer will read it as"prefer a tooltip ov fix it: rows areght to Label subtext.
- An info icon button needs an accessible name (aria-label), and per the Tooltips page every icon button requires a tooltip. Worth stating here so the two pages agree.
- "Application help" appears in the Format column as "Online (TBD, Nov 2026)" — that's a delivery status, not a format, and the November 2026 date is then repeated in the body. Keeping it in one place means one edit when it slips.                                                Two wording snags inhe content until whenit is available" is ungrammatical, and "only create a single, temporary in-app guide dialog, but do not create more than one" says the same    thing twice. It alsopening "we do not ship guides" until you notice guide dialog ≠ embedded guide. And "not more than one" — one per app, per view, per extension? That rule will need enforcing in review, so it should be unambiguous.

Optional

- Every other behavior-guidance page (Tooltips, Applying Changes, Design Principles) carries ExampleBlock prefer/avoid pairs; this one is prose and a table. The lading one example stripwould make concrete. Fine as a follow-up.
- The linking is one-wtips twice, Tooltipsdoesn't point back. Applying Changes ends with a "See also" line that's a good pattern to copy.

katherinejensen00 (changes requested):
# PR #2787 — "ux guidelines for providing help" — re-review findings

**Branch:** `providing-help` → `main` · 3 files, +188/−0
**Re-reviewed at:** `f161bacf9` (`docs(providing-help): apply review feedback`)

All twelve findings from the first round are addressed. Verified mechanically where possible:
`npx prettier --check` and `npx remark --ext .mdx --output` both produce a **zero-byte diff** on the
branch, every `LinkTo title` resolves to a real `<Meta title>`, `eslint` on `tooltip.stories.tsx` is
clean, and `Guidelines` sorts alphabetically in `.storybook/preview.ts` so there is no index or
`storySort` list that needs the new page added.

Five findings remain, ranked. The first two are worth acting on before merge; the rest are polish.

---

## 1. The `aria-label` paragraph is technically wrong, and both examples on the page contradict it

**File:** `lib/platform-bible-react/src/stories/guidelines/providing-help.mdx:59-62`
**Severity:** high — the page is a guideline, so a wrong rule propagates into every future call site

The prose reads:

    The tooltip and the `aria-label` are where the help itself lives — the tooltip is what a sighted
    user reads on hover or keyboard focus, the `aria-label` is what a screen-reader user hears — so
    write both to carry the explanation, not just to satisfy the icon-button rule.

Radix already handles the screen-reader half. `TooltipContent` renders a visually-hidden
`role="tooltip"` node containing the content, and the trigger receives
`aria-describedby={open ? contentId : undefined}` (confirmed in
`node_modules/@radix-ui/react-tooltip/dist/index.mjs`). Keyboard focus opens the tooltip, so a
screen-reader user hears **name → role → description**, where the description *is* the tooltip text.

`aria-label` sets the accessible **name**. Putting one or two sentences there:

- makes the button's announced name a paragraph, read in full before the description;
- breaks voice control — "click About the API key" stops matching a sentence-length name;
- duplicates text the user is already going to hear via `aria-describedby`.

The page's own two examples already do the right thing — `aria-label="About the API key"` at
`providing-help.mdx:95` and in `tooltip.stories.tsx` — which is the opposite of what the prose asks
for.

**Suggested shape:** the `aria-label` is a short name for *what* the button explains; the tooltip
carries the explanation and is announced as the button's description.

---

## 2. Nothing in the repo follows the ghost-icon-button rule, and the page never says why it matters

**File:** `lib/platform-bible-react/src/stories/guidelines/providing-help.mdx:58`
**Severity:** high — the rule is correct but currently unenforceable and universally violated

The page prescribes "a ghost-variant icon button (`Info` from `lucide-react`)". Every `Info`
affordance shipping today instead uses a bare `<Info>` SVG directly as the `TooltipTrigger`:

| Site | Line |
| --- | --- |
| `extensions/src/platform-scripture/src/find/find-filters.component.tsx` | 128, 214 |
| `extensions/src/platform-scripture/src/find/find.component.tsx` | 1038 |
| `extensions/src/platform-scripture/src/manage-books-dialog/manage-books-dialog.component.tsx` | 2528 |
| `extensions/src/platform-enhanced-resources/src/components/toolbar/toolbar.component.tsx` | 391 |
| `extensions/src/platform-enhanced-resources/src/components/warning-ribbons/warning-ribbons.component.tsx` | 96 |
| `extensions/src/platform-enhanced-resources/src/components/guide/marble-guide.component.tsx` | 292 |

A bare `<svg>` is not focusable and carries no role, so those tooltips are **hover-only** — keyboard
and screen-reader users never reach them. `manage-books-dialog.component.tsx:2530` even puts
`aria-label` on the SVG, where it does nothing at all. That is precisely the confusion finding #1
would otherwise reinforce.

The page states the rule but not the reason, and the reason is what makes it stick in review.

**Suggested:** add an explicit avoid line — a bare `Info` icon as the trigger isn't focusable, so the
tooltip is unreachable by keyboard — and open a follow-up ticket for the seven existing sites.

---

## 3. The "Label subtext" section says nothing about associating subtext with its control

**File:** `lib/platform-bible-react/src/stories/guidelines/providing-help.mdx` — "Label subtext"
section, and the avoid example at lines 112-129
**Severity:** medium — same class of gap as #2, in the one other pattern the page introduces

The section covers *when* to use subtext and *how sparingly*, but never how to wire it up. Subtext
that is not referenced by the control's `aria-describedby` is invisible to screen-reader users, who
hear the label alone and miss exactly the disambiguation the section says is essential.

The avoid example demonstrates the unassociated shape: `<p className="tw:ps-6 …">` with no `id`, and
`<RadioGroupItem id="providing-help-opt-a">` with no `aria-describedby`. The example's *point* is
length, so the markup is incidental — but it is the only subtext markup on the page, and readers
will copy it.

**Suggested:** one line in the section — give the subtext an `id` and point the control's
`aria-describedby` at it — and wire the example to match.

---

## 4. The avoid example carries a dead `TooltipProvider` and padding for a tooltip that isn't there

**File:** `lib/platform-bible-react/src/stories/guidelines/providing-help.mdx:112-129`
**Severity:** nit

The block wraps its preview in `<TooltipProvider>` but contains no `Tooltip`, and sets
`previewClassName="tw:pb-12"`, which exists to clear space for a floating tooltip. Both are
leftovers from the prefer block above and can be dropped.

---

## 5. The ladder's Format column gives tooltips two different length budgets without naming the surface

**File:** `lib/platform-bible-react/src/stories/guidelines/providing-help.mdx` — the ladder table
**Severity:** nit

| Row | Format column |
| --- | --- |
| **Tooltip on the control** | Short phrase (ideally 1–3 words) |
| **Info icon button** | 1–2 sentences, or a dialog for more |

An info icon button's help is, in the common case, *also* a tooltip — so a reader scanning the Format
column alone sees two conflicting budgets for the same surface. The `tooltips.mdx` carve-out resolves
it, and the body of this page resolves it, but the table doesn't: the Info row never names the
surface the "1–2 sentences" lives in.

**Suggested:** "1–2 sentences in a tooltip, or a dialog for more".

---

## Verified clean (no action needed)

- `prettier --check` and `remark --output`: zero diff — the first-round formatting finding is fixed.
- Every `LinkTo title` resolves: `Guidelines/Tooltips`, `Guidelines/Applying Changes`,
  `Guidelines/Product Names`, `Guidelines/Providing Help`.
- `eslint src/stories/shadcn-ui/tooltip.stories.tsx` — clean.
- `Label` and `Input` were already imported in `tooltip.stories.tsx`; only `Info` needed adding.
- `<p className="tw:max-w-xs">` inside `TooltipContent` matches the existing convention in that file
  (line 958).
- `<Tooltip open>` in the `.mdx` previews matches the four existing uses in `tooltips.mdx`.
- No Storybook index or `storySort` list needs updating — `Guidelines` sorts alphabetically.
- The `tooltips.mdx` carve-out sits above an avoid example about a *Save* icon button, not an info
  button, so the two don't contradict.

> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2787#-:-P1Gtq228xgXb9juMvNm:b-ihp3j2)* status: all files reviewed, 1 unresolved discussion (waiting on Sebastian-ubs).

___
*[`lib/platform-bible-react/src/stories/guidelines/providing-help.mdx` line 59 at r1](https://reviewable.io/reviews/paranext/paranext-core/2787#-P1GtxiJ7qQ3a53-qKbc:-P1GtxiJ7qQ3a53-qKbd:b-3d4u5c) ([raw file](https://github.com/paranext/paranext-core/blob/f161bacf92ebd126ace09134afcdc388545f10f1/lib/platform-bible-react/src/stories/guidelines/providing-help.mdx#L59)):*
> ```MDX
> 
> A ghost-variant icon button (`Info` from `lucide-react`) placed next to the thing it explains.
> The tooltip and the `aria-label` are where the help itself lives — the tooltip is what a sighted
> ```

## 1. The `aria-label` paragraph is technically wrong, and both examples on the page contradict it

**File:** `lib/platform-bible-react/src/stories/guidelines/providing-help.mdx:59-62`
**Severity:** high — the page is a guideline, so a wrong rule propagates into every future call site

The prose reads:

    The tooltip and the `aria-label` are where the help itself lives — the tooltip is what a sighted
    user reads on hover or keyboard focus, the `aria-label` is what a screen-reader user hears — so
    write both to carry the explanation, not just to satisfy the icon-button rule.

Radix already handles the screen-reader half. `TooltipContent` renders a visually-hidden
`role="tooltip"` node containing the content, and the trigger receives
`aria-describedby={open ? contentId : undefined}` (confirmed in
`node_modules/@radix-ui/react-tooltip/dist/index.mjs`). Keyboard focus opens the tooltip, so a
screen-reader user hears **name → role → description**, where the description *is* the tooltip text.

`aria-label` sets the accessible **name**. Putting one or two sentences there:

- makes the button's announced name a paragraph, read in full before the description;
- breaks voice control — "click About the API key" stops matching a sentence-length name;
- duplicates text the user is already going to hear via `aria-describedby`.

The page's own two examples already do the right thing — `aria-label="About the API key"` at
`providing-help.mdx:95` and in `tooltip.stories.tsx` — which is the opposite of what the prose asks
for.

**Suggested shape:** the `aria-label` is a short name for *what* the button explains; the tooltip
carries the explanation and is announced as the button's description.

Sebastian-ubs:
Finding #2 follow-up: opened #2811 to retrofit the four broken sites (find-filters × 2, find preserve-case, manage-books-dialog based-on info) with a ghost-variant `Button` around `<Info>`, so the tooltips are keyboard- and screen-reader-reachable. The other three sites the review named are already compliant.

katherinejensen00 (approved):
:lgtm:

> @katherinejensen00 reviewed 1 file and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2787#-:-P1Wxmw0-y4AkbUovVB3:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on Sebastian-ubs).

PR #2789, opened 2026-09-09
Normalize empty Pt9InterlinearSetup strings to absent

imnasnainaec:
## Summary

`Pt9InterlinearSetup`'s doc comment promised more than the conversion delivered: "String fields that are empty in the project are absent here." `ConvertPt9InterlinearSetup` normalized only three of the record's seven nullable string fields.

The first attempt was to document the real behavior instead. That turned out not to be statable: emptiness depends not just on the field but on which path built the setup. An empty `<LanguageName />` in the setups file came back as `""`, while the same setup rebuilt from legacy settings came back with no language name at all; an empty export id setting came back absent on the legacy path and as `""` from the setups file. A consumer cannot tell which path produced a setup, so no per-field rule was usable.

So this PR makes the original sentence true rather than replacing it. Every setup string that is empty in the project is now served as absent, on both paths.

## Changes

- **`ConvertPt9InterlinearSetup`**: route `LanguageId`, `LanguageName`, and both id `ToString()` results through a new `NullIfEmpty` helper, alongside the `FontName` and `ExportScrTextName` guards that were already there. `Type` is an enum's `ToString` and never empty; `ModelScrTextName` is already covered by the no-model check.
- **`Pt9InterlinearSetup`**: restate emptiness as one rule true on both paths. Two things the investigation turned up are documented rather than normalized, because they are real and previously unstated: the id fields serve PT9's own re-formatting rather than the stored characters (hex folds to lowercase, a legacy `res` id is re-encoded), and a malformed id fails the whole read from the setups file while the legacy path tolerates it.
- **`platform-scripture.d.ts`**: carry the same rule to the surface extension authors actually read, which said nothing about emptiness before.
- **Tests**: `GetPt9InterlinearData_ServesEmptyAndOmittedSetupStringsAlikeAsAbsent` covers every field written empty against the same setup with every field omitted. Three more pin what the doc now claims and nothing covered before: id re-formatting, the setups-file read failing on a malformed id, and the legacy path's unassigned display fields.

### Why this PR is here

Not part of the current epic. (It surfaced against an Interlinearizer extension doc PR.)

## Behavior change

The four pass-through fields now serve absent instead of `""` when the project stored an empty value. Nothing consumes them yet — the Interlinearizer extension reads only `books`, and `test-data/Pt9InterlinearProjectData.json` carries no empty setup strings — and the type is `@experimental`, so this is the cheapest point at which to fix it.

## AI Involvement

AI-assisted. Claude established the per-field and per-path behavior by exercising `GetPt9InterlinearData` against setups files and legacy settings written for each case, verified `InterlinearSetup`'s deserialization and `HexId`'s re-formatting by running them directly (ParatextData 9.5.0.24), wrote the normalization and the tests, and mutation-checked each new assertion by removing the guard it covers. I reviewed the resulting behavior change and confirmed no current consumer reads these fields.

## Risk Level

Low-medium - a behavior change to a payload no shipped code reads yet, on an `@experimental` type, pinned by tests that fail when any guard is removed.

---

Devin review: https://app.devin.ai/review/paranext/paranext-core/pull/2789

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2789)

### DIFF
--- a/c-sharp/Projects/Pt9InterlinearData.cs
+++ b/c-sharp/Projects/Pt9InterlinearData.cs
@@ -146,7 +146,9 @@ public sealed record Pt9Lexicon(
 /// the model text the interlinearization reads from, and the export half: whether and where
 /// approved verses export. The model name is absent for a setup with no model text; the model id
 /// serves whenever PT9 stored one, since a model-less setup mints an id as its settings key.
-/// String fields that are empty in the project are absent here.
+/// Emptiness is not handled uniformly: an empty <c>FontName</c>, <c>ModelScrTextName</c>, or
+/// <c>ExportScrTextName</c> is absent here, while an empty <c>LanguageId</c> or
+/// <c>LanguageName</c> serves as an empty string.
 /// </summary>
 public sealed record Pt9InterlinearSetup(
     [property: JsonPropertyName("type")] string Type,
--- a/c-sharp/Projects/Pt9InterlinearData.cs
+++ b/c-sharp/Projects/Pt9InterlinearData.cs
@@ -146,9 +146,12 @@ public sealed record Pt9Lexicon(
 /// the model text the interlinearization reads from, and the export half: whether and where
 /// approved verses export. The model name is absent for a setup with no model text; the model id
 /// serves whenever PT9 stored one, since a model-less setup mints an id as its settings key.
-/// Emptiness is not handled uniformly: an empty <c>FontName</c>, <c>ModelScrTextName</c>, or
-/// <c>ExportScrTextName</c> is absent here, while an empty <c>LanguageId</c> or
-/// <c>LanguageName</c> serves as an empty string.
+/// Emptiness is not normalized uniformly. A <c>FontName</c>, <c>ModelScrTextName</c>, or
+/// <c>ExportScrTextName</c> that is empty in the project is absent here; <c>LanguageId</c>,
+/// <c>LanguageName</c>, <c>ModelScrTextId</c>, and <c>ExportScrTextId</c> serve as the project
+/// stored them, so an empty one serves as an empty string and only a field the project omits is
+/// absent. That split is incidental to how each field is read rather than a guarantee of this
+/// payload, so treat an empty string and an absent field alike.
 /// </summary>
 public sealed record Pt9InterlinearSetup(
     [property: JsonPropertyName("type")] string Type,
--- a/c-sharp-tests/Projects/ParatextProjectDataProviderPt9InterlinearTests.cs
+++ b/c-sharp-tests/Projects/ParatextProjectDataProviderPt9InterlinearTests.cs
@@ -443,7 +443,7 @@ internal class ParatextProjectDataProviderPt9InterlinearTests : PapiTestBase
     [Description(
         "A setup's model and export fields are served as written; PT9's __EMPTY__ sentinel "
             + "means no model text, so the model name is absent while the id PT9 minted for the "
-            + "setup still serves, and empty strings serve as absent."
+            + "setup still serves, and an empty text name serves as absent."
     )]
     public void GetPt9InterlinearData_ServesSetupModelTextAndTreatsTheEmptySentinelAsNone()
     {
@@ -485,6 +485,58 @@ internal class ParatextProjectDataProviderPt9InterlinearTests : PapiTestBase
             Assert.That(data.Setups[1].ModelScrTextName, Is.Null);
             Assert.That(data.Setups[1].ModelScrTextId, Is.EqualTo("fedcba0987654321"));
             Assert.That(data.Setups[1].ExportScrTextName, Is.Null);
+        });
+    }
+
+    [Test]
+    [Description(
+        "Emptiness is not normalized uniformly: an empty font name or text name serves as "
+            + "absent, while an empty language id, language name, or text id serves as an empty "
+            + "string and only a field the project omits is absent."
+    )]
+    public void GetPt9InterlinearData_ServesEmptySetupStringsAsAbsentOrEmptyPerField()
+    {
+        WriteProjectFile(
+            "InterlinearSetup.xml",
+            """
+            <InterlinearSetupList>
+              <InterlinearSetup type="Glossing" language="">
+                <LanguageName></LanguageName>
+                <FontName></FontName>
+                <MdlScrTextName></MdlScrTextName>
+                <MdlScrTextId></MdlScrTextId>
+                <ExportScrTextName></ExportScrTextName>
+                <ExportScrTextId></ExportScrTextId>
+              </InterlinearSetup>
+              <InterlinearSetup type="Glossing">
+              </InterlinearSetup>
+            </InterlinearSetupList>
+            """
+        );
+
+        var data = _provider.GetPt9InterlinearData();
+
+        Assert.That(data.Setups, Has.Count.EqualTo(2));
+        Assert.Multiple(() =>
+        {
+            // Written empty: the three fields the conversion nulls are absent, the four it passes
+            // through keep the empty string PT9 stored.
+            Assert.That(data.Setups[0].FontName, Is.Null);
+            Assert.That(data.Setups[0].ModelScrTextName, Is.Null);
+            Assert.That(data.Setups[0].ExportScrTextName, Is.Null);
+            Assert.That(data.Setups[0].LanguageId, Is.Empty);
+            Assert.That(data.Setups[0].LanguageName, Is.Empty);
+            Assert.That(data.Setups[0].ModelScrTextId, Is.Empty);
+            Assert.That(data.Setups[0].ExportScrTextId, Is.Empty);
+
+            // Omitted entirely: absent regardless of which side of the split the field is on.
+            Assert.That(data.Setups[1].FontName, Is.Null);
+            Assert.That(data.Setups[1].ModelScrTextName, Is.Null);
+            Assert.That(data.Setups[1].ExportScrTextName, Is.Null);
+            Assert.That(data.Setups[1].LanguageId, Is.Null);
+            Assert.That(data.Setups[1].LanguageName, Is.Null);
+            Assert.That(data.Setups[1].ModelScrTextId, Is.Null);
+            Assert.That(data.Setups[1].ExportScrTextId, Is.Null);
         });
     }
 
--- a/c-sharp/Projects/Pt9InterlinearReader.cs
+++ b/c-sharp/Projects/Pt9InterlinearReader.cs
@@ -543,8 +543,9 @@ internal static class Pt9InterlinearReader
     /// <summary>
     /// Maps one setup to its served shape. A model name that is empty or PT9's no-model sentinel
     /// means the setup has no model text, so the name is absent; the model id passes through
-    /// whenever present, since a model-less setup mints one as its settings key. Empty strings
-    /// serve as absent fields.
+    /// whenever present, since a model-less setup mints one as its settings key. An empty font
+    /// name or export text name is likewise absent; every other string passes through, so an
+    /// empty one stays empty.
     /// </summary>
     private static Pt9InterlinearSetup ConvertPt9InterlinearSetup(InterlinearSetup setup)
     {


c-sharp/Projects/Pt9InterlinearData.cs
```diff
@@ -146,7 +146,9 @@ public sealed record Pt9Lexicon(
 /// the model text the interlinearization reads from, and the export half: whether and where
 /// approved verses export. The model name is absent for a setup with no model text; the model id
 /// serves whenever PT9 stored one, since a model-less setup mints an id as its settings key.
-/// String fields that are empty in the project are absent here.
+/// Emptiness is not handled uniformly: an empty <c>FontName</c>, <c>ModelScrTextName</c>, or
+/// <c>ExportScrTextName</c> is absent here, while an empty <c>LanguageId</c> or
+/// <c>LanguageName</c> serves as an empty string.
```
imnasnainaec: 🤖 **Claude AI** · **[Scope]** **[Contracts]** **[Architecture]** · Decision needed (warning)

Correct or drop `Empty strings serve as absent fields.` on `ConvertPt9InterlinearSetup` in this PR, or note in the body that the reader-side duplicate is knowingly deferred?

The blanket claim this PR removes still stands on the method the PR body cites as its evidence. `ConvertPt9InterlinearSetup`'s summary ends:

> the model id passes through whenever present, since a model-less setup mints one as its settings key. Empty strings serve as absent fields.

Its body applies `IsNullOrEmpty → null` to only `FontName`, `MdlScrTextName` (via `hasModel`), and `ExportScrTextName`; `setup.LanguageId` and `setup.LanguageName` pass through raw. Before this PR the two docs agreed and were both wrong. After it they contradict each other outright — and the converter's is the copy a maintainer reads while editing the mapping.

A repo-wide search finds this as the only remaining copy of the claim; the `platform-scripture` `.d.ts` is worded differently and needs no edit, as the PR body says.

Flagged independently by the scope, contracts, and architecture reviewers.

`c-sharp/Projects/Pt9InterlinearReader.cs:543-548` (the doc comment), `:554-567` (the mapping it describes)

</details>

imnasnainaec: 🤖 **Claude AI** · **[Contracts]** **[Architecture]** · Decision needed (warning)

Make the new sentence exhaustive over the record's nullable string fields and their states, or explicitly scope it to the name/font fields so it no longer reads as exhaustive?

The old sentence was wrong but total; the new one is right but partial, and it still reads as an exhaustive split. Two gaps:

**1. Two fields get no rule at all.** The record has seven nullable string fields; the sentence names five. `ModelScrTextId` (line 161) and `ExportScrTextId` (line 166) go unmentioned, and both are passed through unguarded — `setup.MdlScrTextId?.ToString()` / `setup.ExportScrTextId?.ToString()` (`Pt9InterlinearReader.cs:562,567`). The old blanket sentence covered them; a consumer looking either one up now finds nothing. (`Type` is genuinely exempt: an enum's `ToString`, never empty.)

Worth settling which side they land on before adding them, because the two reviewers who looked read it differently. Contracts read the setups-file path as serving `""` — an empty `<MdlScrTextId></MdlScrTextId>` deserializing via `Memento.GetXml` into a non-null `HexId` whose `ToString()` is `""` — and read the existing `ExportScrTextId Is.Null` assertion (`ParatextProjectDataProviderPt9InterlinearTests.cs:546`) as the separate legacy-settings path, where `HexId.FromStrSafe("")` returns null. Architecture read that same test as evidence both ids behave like the absent group. If both paths are live, these two are absent *or* empty depending on which one produced the setup — itself worth stating rather than omitting.

**2. `LanguageId`/`LanguageName` have a second degenerate state.** Both are declared `string?` (lines 155-156) and passed through verbatim, so they are null — absent — when the setups file omits the `language` attribute or the `<LanguageName>` element. The sentence says only that an empty value serves as `""`, which invites a consumer to treat them as always-present strings, exactly what the nullable annotation denies.

</details>

imnasnainaec (commented):
## 🤖 [Claude AI Review] — PR #2789

Comment-only PR. Panel auto-narrowed to six perspectives (`ux` dropped — no TS/React/style files).

**The correction itself checks out.** `ConvertPt9InterlinearSetup` does guard `FontName` and `ExportScrTextName` with `IsNullOrEmpty → null` and gates `ModelScrTextName` on `hasModel`, while `LanguageId` and `LanguageName` are passed through untouched. Scope creep: none — the diff is exactly the comment replacement described.

Two decisions are posted inline, both about the *corrected* sentence rather than the correction: a stale duplicate of the old claim left on the converter, and the new sentence reading as exhaustive when it covers five of seven nullable string fields and one of two degenerate states.

| Security     | 0        | 0       | 0          |
| Scope        | 0        | 1       | 0          |
| Contracts    | 0        | 3       | 1          |
| Tests        | 0        | 0       | 1          |
| Architecture | 0        | 1       | 2          |
| Clarity      | 0        | 0       | 1          |
| **Total**    | **0**    | **5**   | **5**      |

---

### Additional findings — no decision required

🤖 **Claude AI** · **[Clarity]** (suggestion)

The replacement drops the old sentence's "in the project" anchor, and without it the second clause reads as a tautology: "an empty `LanguageId` … serves as an empty string" says only that empty stays empty. The informative content — PT9 stored an empty value and this payload passes it through rather than nulling it — depends on the reader supplying the missing source. The first clause survives the loss because "absent here" contrasts two locations, so only half the sentence keeps the old precision. Restoring the anchor also makes the halves parallel. Keeping "absent"/"serves" over a literal `null`/`""` is right either way: that is the file's own vocabulary (lines 65, 148 and 17, 132, 148).

🤖 **Claude AI** · **[Tests]** (suggestion)

The corrected claim is only half-pinned. The empty-to-absent side has coverage for two of the three guarded fields (`ExportScrTextName, Is.Null` at `ParatextProjectDataProviderPt9InterlinearTests.cs:487`; `ModelScrTextName, Is.Null` at `:437` and `:486`). Nothing pins the other half: no test supplies an empty `FontName`, and none asserts that `LanguageId`/`LanguageName` come back as `""`. Adding `IsNullOrEmpty(...) ? null` to those two arguments in the converter would keep every Pt9-filtered test green while silently falsifying the new comment. Extending `GetPt9InterlinearData_ServesSetupModelTextAndTreatsTheEmptySentinelAsNone` with a setup carrying `language=""`, an empty `<LanguageName />` and an empty `<FontName />` would pin the whole asymmetry in the class that already owns this conversion. The author's cited test run does exercise this path; all five skips are unrelated environment guards.

🤖 **Claude AI** · **[Architecture]** (suggestion)

The sentence states the asymmetry in the same flat contractual voice as the rest of the file, so it reads as guaranteed shape a consumer may code to — after which normalizing the converter (the structurally better fix) becomes a breaking change for anyone testing `LanguageId === ''`. The file elsewhere marks deliberateness explicitly ("The consequence deliberately differs, though", "Read-only by construction"), so silence here reads as intent. A clause noting the asymmetry is incidental would keep the uniform fix free later. Documenting rather than fixing is reasonable given the external extension pointing here and the epic boundary; this is only about not hardening the accident into contract.

🤖 **Claude AI** · **[Contracts]** (suggestion)

Pre-existing, not introduced here, but it colors the word "absent": `SerializationOptions.CreateSerializationOptions()` and its StreamJsonRpc formatter both report `DefaultIgnoreCondition = Never`, and `Pt9InterlinearSetup` carries no `[JsonIgnore(Condition = WhenWritingNull)]` — unlike `c-sharp/JsonUtils/PlatformStyleInfo.cs:13-19`, whose remark documents this exact hazard. So on the wire these fields are present with an explicit `null` rather than omitted, and the distinction the sentence draws reaches the consumer as `null` vs `""`. The TypeScript mirror declares the same fields optional (`fontName?: string`) and carries no emptiness note at all — and that `.d.ts` (`extensions/src/platform-scripture/src/types/platform-scripture.d.ts:1470-1498`), not this XML doc, is what the external interlinearizer extension reads via IntelliSense. Possible follow-up, out of scope here.

---

<sub>Automated multi-perspective review. Findings were verified against the working tree; the two inline items were each raised by more than one reviewer. Not a substitute for human review — no approval is implied.</sub>

imnasnainaec (commented):
> @imnasnainaec reviewed all commit messages and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2789)* status: 0 of 3 files reviewed, all discussions resolved.

katherinejensen00 (changes requested):
# Review — PR #2789 "Correct Pt9InterlinearSetup's emptiness claim"

To: @imnasnainaec

## Summary

The premise is right and the direction is good: the old blanket sentence was wrong, and
replacing it plus pinning the behavior with a test is exactly the right instinct. But the
replacement sentence is still not true. It states the empty/absent split as a property of each
FIELD, when it is really a property of the field AND the construction path — and on the
legacy-settings path the new sentence is falsified by a test already in this same file. Since
this PR's entire deliverable is one accurate sentence, that is the finding that matters most.
Separately, about half of the new test cannot fail.

Everything below was verified by executing the pinned ParatextData 9.5.0.24 assembly and
NUnit 4.0.1, not by reading alone. Line numbers are for the NEW version of each file.

---

## 1. High — `c-sharp/Projects/Pt9InterlinearData.cs:152`

The new rule is still wrong, on a path that runs in the ordinary case.

"…so an empty one serves as an empty string and only a field the project omits is absent" holds
on the setups-file path, but not on the legacy-settings path in `ResolvePt9InterlinearSetups`
(`Pt9InterlinearReader.cs:451-520`). There, `ExportScrTextId = HexId.FromStrSafe(GetSetting(...))`
(`:481`, assigned `:514`). `ProjectSettings.GetSetting` returns the same `""` whether a setting is
stored empty or absent entirely, and `FromStrSafe("")` returns null — so stored-empty,
stored-garbage and omitted all collapse to null before `ConvertPt9InterlinearSetup` ever runs.
`ExportScrTextId` can never be `""` on that path. `LanguageName` is never assigned there at all
(`:500-516`), so it is unconditionally null regardless of what the project stores.

Reproduced two ways independently: setting `InterlinearExportTextId.<model>` to `""` and reading
through the provider yields `ExportScrTextId = null`, `LanguageName = null`. And your own untouched
test `GetPt9InterlinearData_ToleratesAMalformedLegacyExportTextId` (`tests:586`) already asserts
`Is.Null` for a value that IS stored.

This is not a hypothetical path: the guard at `Pt9InterlinearReader.cs:456-462` only skips the merge
for an observer or an already-converted user, so the legacy path runs on the ordinary first read.

Worth noting this is the prior round's second finding, which asked you to settle which side the id
fields land on and flagged "absent *or* empty depending on which path produced the setup — itself
worth stating rather than omitting." That is still open.

Cheapest fix is a clause, not a redesign — scope the sentence to the setups-file path and say the
legacy path only ever serves these as null or a full id.

## 2. High — `c-sharp/Projects/Pt9InterlinearData.cs:151`

"serve as the project stored them" is not true for the two id fields.

`ModelScrTextId` and `ExportScrTextId` are the only `HexId`-typed members, served as
`HexId?.ToString()` (`Pt9InterlinearReader.cs:562,567`) — a parse/re-format, not a pass-through.
Reproduced through a real XmlSerializer round-trip:

- Case is folded: `<MdlScrTextId>ABCDEF1234567890</MdlScrTextId>` serves `abcdef1234567890`
  (`HexId.ToString()` calls `ByteArrToHex(..., lowercase: true)`).
- Legacy resource ids are re-encoded: `1234567890abcdefres` serves `1234567890abcdefabcdefff`
  (the `Id` setter strips the trailing `res` and appends a fixed 4-byte suffix).
- A malformed id does not degrade, it throws: `<MdlScrTextId>not hex</MdlScrTextId>` raises
  `ArgumentException` from `HexId.set_Id`, which `DeserializePt9Xml` (`:426-436`) rethrows as
  `Pt9FileUnreadable` — setups, books, lexicon and word analyses all fail together.

None of those outcomes is "as stored", "empty", or "absent".

## 3. Medium — `c-sharp-tests/Projects/ParatextProjectDataProviderPt9InterlinearTests.cs:533`

The omitted-setup half of the new test cannot fail.

All seven `Is.Null` assertions on `Setups[1]` pass by CLR default. Verified by instantiating a bare
`InterlinearSetup()`: every one of the seven fields is already null before any guard runs, and an
XmlSerializer round-trip of an empty element leaves them there. Delete the three `IsNullOrEmpty`
ternaries from the converter entirely and `:533-535` still pass; `:536-537` have no guard to remove;
`:538-539` only incidentally guard `?.` against a null-reference. Three of them also duplicate
`GetPt9InterlinearData_ParsesSetups` (`:435,436,438`).

The documented behavior is exercised falsifiably only by the `Setups[0]` block (`:524-530`). That
half is genuinely good — a targeted mutation of the normalization makes it fail cleanly.

## 4. Medium — `c-sharp-tests/…/ParatextProjectDataProviderPt9InterlinearTests.cs:527`

The test pins the accident the doc explicitly disclaims.

`Pt9InterlinearData.cs:153` calls the split "incidental … rather than a guarantee of this payload,"
but the four `Is.Empty` assertions at `:527-530` lock it in as a contract. If anyone later
normalizes the four pass-through fields, these turn red and a future PR has to delete assertions
added here.

Flagging the tension honestly rather than just prescribing: `Is.Null.Or.Empty` would assert the
stated contract instead of the accident — but it would also make this the non-falsifiable half,
and right now `Setups[0]` is the ONLY falsifiable part of the test (see #3). So the two cannot both
be satisfied by a constraint swap. The real resolution is to decide whether the split is a contract
or an accident, and make doc and test agree. If it is an accident, the honest test is
`Is.Null.Or.Empty` plus dropping the claim from the doc; if it is a contract, keep `Is.Empty` and
drop "incidental … rather than a guarantee."

## 5. Medium — `c-sharp-tests/…/ParatextProjectDataProviderPt9InterlinearTests.cs:497`

The new test covers only one of the two construction paths.

It writes `InterlinearSetup.xml` and never exercises `ResolvePt9InterlinearSetups`'s settings-merge
branch, which is precisely where #1's counterexample lives. Nothing would fail if the legacy path's
emptiness behavior drifted further. A second fixture using `InterlinearRelatedLanguages` settings
would pin it.

## 6. Medium — `c-sharp/Projects/Pt9InterlinearData.cs:149`

The rule a caller needs is the last clause after a five-line enumeration.

The old sentence was 11 words. The replacement is ~75 words across six lines, opening with
"Emptiness is not normalized uniformly," then naming all seven fields, and only then arriving at the
one actionable instruction — "treat an empty string and an absent field alike." A skimmer reads the
enumeration and misses the rule. Lead with the rule, then give the detail as the fallback:

    /// Treat an empty string and an absent field alike: which of the two you get is an accident of
    /// how each field happens to be read, not a promise callers can rely on. Concretely, on the
    /// setups-file path FontName, ModelScrTextName, and ExportScrTextName come back null when
    /// written empty, while LanguageId, LanguageName, ModelScrTextId, and ExportScrTextId come back
    /// as empty strings; setups rebuilt from legacy settings never serve an empty string at all.

Related: the two-row table in your PR body is clearer than anything that landed in the code. XML doc
comments will not render a Markdown table, but the prose can be shaped to read like it — which is
what the rewrite above is doing.

## 7. Medium — `extensions/src/platform-scripture/src/types/platform-scripture.d.ts:1470`

The audience being told to "treat empty and absent alike" never sees the instruction.

The `.d.ts` is the hand-maintained surface extension authors actually read, and neither its type doc
(`:1459-1469`) nor its per-field docs mention the split. `modelScrTextName` and `exportScrTextName`
describe *semantic* absence; `languageId`, `languageName`, `fontName`, `modelScrTextId` and
`exportScrTextId` say nothing about emptiness at all.

The PR body's "needs no matching edit" is answering "does it repeat the false claim?" (correctly,
no). The open question is "does it convey the new guidance?" — and it does not. This is the cheapest
high-value addition available, and the same class of edit the PR is already making.

## 8. Medium, follow-up ticket rather than this PR — `c-sharp/Projects/Pt9InterlinearData.cs:156`

"Absent" is not what a consumer actually receives.

`SerializationOptions.CreateSerializationOptions()` sets `PropertyNamingPolicy` but never
`DefaultIgnoreCondition`, and `Pt9InterlinearData.cs` carries zero `[JsonIgnore]` attributes — while
sibling `ProjectMetadata.cs` has five and `PlatformStyleInfo.cs`'s own `<remarks>` documents this
exact trap and the per-property remedy. So a null ships as `"languageId": null`, a PRESENT key. The
TS side declares `languageId?: string`, never `| null`, under `strict`, so a consumer checking
`=== undefined` misses the no-value case.

I am deliberately not asking you to fix this here: it is pre-existing, outside your diff, and has
zero consumers today. But it is the deeper version of the same "empty vs absent" confusion this PR
is about, so it is worth a ticket rather than silence.

## 9. Low — `c-sharp/Projects/Pt9InterlinearData.cs:153`

"Incidental to how each field is read" points at the wrong mechanism, and "a guarantee of this
payload" is jargon. Every field is read identically; the split originates in the converter's
selective normalization, not in reading. Plain wording: "an accident of how the conversion happens
to normalize each field, not a promise callers can rely on."

## 10. Low — `c-sharp/Projects/Pt9InterlinearReader.cs:544`

The rule now exists in three independently worded places — this doc, the record's doc, and the new
test's `[Description]`. They agree today, and nothing keeps them agreeing; a future edit only has to
touch one. This doc is not wrong as written (`ModelScrTextName` is genuinely covered by the earlier
sentinel sentence), just redundant. Suggest deferring: "see <see cref="Pt9InterlinearSetup"/> for
which fields come back absent when written empty" and keeping only the sentinel detail that is
specific to conversion.

## 11. Low — `c-sharp/Projects/Pt9InterlinearReader.cs:547`

"every other string passes through, so an empty one stays empty" sweeps in three members that are
not strings on the source object: `Type` is an `InterlinearType` enum, and the two id fields are
`HexId` (see #2, where the pass-through claim is actually false). Scope the sentence to the fields
that really are strings.

## 12. Low — `c-sharp/Projects/Pt9InterlinearReader.cs:505`

`ModelScrTextId = modelText.Guid` on the legacy path is not unconditionally non-null:
`ProjectSettings.Guid` returns `HexId.FromStr(GetSetting(Setting.Guid))`, which can be null. This
widens #1 to both id fields, and is untested.

## 13. Low — `c-sharp/Projects/Pt9InterlinearData.cs:149`

Whitespace-only is an undocumented, untested third state. The converter uses `IsNullOrEmpty`, not
`IsNullOrWhiteSpace`, so `<FontName> </FontName>` serves `" "` — neither absent nor empty. Inherited
behavior, not introduced here, but the doc now defines the field's emptiness contract and does not
rule it out, so a refactor could flip it with no red test.

## 14. Low — `c-sharp-tests/…/ParatextProjectDataProviderPt9InterlinearTests.cs:446`

The edited `[Description]` conflates two different absence mechanisms. "an empty text name serves as
absent" is demonstrated in that test only by `ExportScrTextName`; `ModelScrTextName`'s absence there
is the `__EMPTY__` sentinel path, not emptiness.

## 15. Low — `c-sharp/Projects/Pt9InterlinearData.cs:149`

Grouping `ModelScrTextName` under a flat "empty => absent" clause drops its sentinel trigger, which
only the preceding untouched sentence states. The converter's "likewise absent" phrasing avoids
this; the record doc does not.

## 16. Low — `c-sharp-tests/…/ParatextProjectDataProviderPt9InterlinearTests.cs:493`

The new `[Description]` restates finding #1's claim verbatim, so correcting the record doc leaves
this stale unless updated in lockstep. The test name already carries the "per field" promise; the
description only needs to say which fields land where. Same applies to the two body comments at
`:522` and `:532`, which narrate the converter's internal grouping — a regrouping refactor that
preserved output would leave them untrue. (They do pass the forward-facing-comments test; this is
about duplication, not provenance.)

## 17. Nit — `c-sharp-tests/…/ParatextProjectDataProviderPt9InterlinearTests.cs:527`

`Is.Empty` on a `string?` reads as a collection assertion; the seven other uses in this file are all
collections. `Is.EqualTo(string.Empty)` is unambiguous. To be clear, this is readability only — I
checked the behavior and `Is.Empty` does NOT throw on null here: NUnit 4.0.1 resolves
`typeof(TActual) == typeof(string)` first and `StringConstraint.ApplyTo` passes `allowNull: true`,
so a null fails cleanly and `Assert.Multiple` completes.

## 18. Nit — `c-sharp-tests/…/ParatextProjectDataProviderPt9InterlinearTests.cs:524`

Swap blind spot: `Pt9InterlinearSetup` is a 13-argument positional constructor, and every assertion
on row 1 plus three on row 0 expect `Is.Null`. Swapping `FontName` and `ExportScrTextName` in the
`new Pt9InterlinearSetup(...)` call (`Pt9InterlinearReader.cs:555-568`) would be invisible to this
test.

---

## Raised and then dropped — not asking you to do these

- Normalizing all seven fields with a `NullIfEmpty` helper. Defensible engineering, and it would
  make the original one-line sentence true. But it is a behavior change on a PR you scoped as "no
  behavior change," and it needs its own approval and pinning tests. It also pulls in the opposite
  direction from #1, which asks you to describe the messy reality more precisely.
- The same ad hoc mix in sibling converters (`SenseId` nulled; `ApprovedHash`, `BeforeText`,
  `AfterText`, `Language` raw). Real, but those records make no uniformity claim, so nothing there
  is currently wrong.
- An Architecture-Decisions.md entry. A doc-wording correction does not clear CLAUDE.md's
  "significant architecture decision" bar.
- Vocabulary consistency ("serve/serves/served" vs "passes through"). The overload is pre-existing
  house style across 20+ uses in this file family; this PR added one instance.
- The 122-char `[Description]` at `tests:313` — pre-existing and untouched by this diff.

## Verified and clean

No remaining copy of the old blanket claim anywhere in the repo. The narrowed `[Description]` at
`:446` dropped no real coverage. Field enumeration is complete (all seven named, omitted=>absent
stated); only path-dependence is missing. All 14 assertions are empirically correct — an empty
`<MdlScrTextId></MdlScrTextId>` really does yield a non-null `HexId` whose `ToString()` is `""`.
The whitespace-only second fixture really does deserialize as a fully-null setup. Element/attribute
names are right, so no assertion passes vacuously. Test isolation is sound and `Has.Count` is
correctly outside `Assert.Multiple`. `Type` is never empty. No NUnit analyzer diagnostics. CI green.

> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2789#-:-P1IaruM3r840oEWZdzw:b4oms0o)* status: all files reviewed, 1 unresolved discussion (waiting on imnasnainaec).

___
*[`c-sharp/Projects/Pt9InterlinearData.cs` line 152 at r2](https://reviewable.io/reviews/paranext/paranext-core/2789#-P1IazpEBnk4IH4nKGxw:-P1IazpEBnk4IH4nKGxx:b-ytv42t) ([raw file](https://github.com/paranext/paranext-core/blob/859f4322ecfd35b379bd8b13995db5361a725fc2/c-sharp/Projects/Pt9InterlinearData.cs#L152)):*
> ```Smalltalk
> /// <c>ExportScrTextName</c> that is empty in the project is absent here; <c>LanguageId</c>,
> /// <c>LanguageName</c>, <c>ModelScrTextId</c>, and <c>ExportScrTextId</c> serve as the project
> /// stored them, so an empty one serves as an empty string and only a field the project omits is
> ```

## 1. High — `c-sharp/Projects/Pt9InterlinearData.cs:152`

The new rule is still wrong, on a path that runs in the ordinary case.

"…so an empty one serves as an empty string and only a field the project omits is absent" holds
on the setups-file path, but not on the legacy-settings path in `ResolvePt9InterlinearSetups`
(`Pt9InterlinearReader.cs:451-520`). There, `ExportScrTextId = HexId.FromStrSafe(GetSetting(...))`
(`:481`, assigned `:514`). `ProjectSettings.GetSetting` returns the same `""` whether a setting is
stored empty or absent entirely, and `FromStrSafe("")` returns null — so stored-empty,
stored-garbage and omitted all collapse to null before `ConvertPt9InterlinearSetup` ever runs.
`ExportScrTextId` can never be `""` on that path. `LanguageName` is never assigned there at all
(`:500-516`), so it is unconditionally null regardless of what the project stores.

Reproduced two ways independently: setting `InterlinearExportTextId.<model>` to `""` and reading
through the provider yields `ExportScrTextId = null`, `LanguageName = null`. And your own untouched
test `GetPt9InterlinearData_ToleratesAMalformedLegacyExportTextId` (`tests:586`) already asserts
`Is.Null` for a value that IS stored.

This is not a hypothetical path: the guard at `Pt9InterlinearReader.cs:456-462` only skips the merge
for an observer or an already-converted user, so the legacy path runs on the ordinary first read.

Worth noting this is the prior round's second finding, which asked you to settle which side the id
fields land on and flagged "absent *or* empty depending on which path produced the setup — itself
worth stating rather than omitting." That is still open.

Cheapest fix is a clause, not a redesign — scope the sentence to the setups-file path and say the
legacy path only ever serves these as null or a full id.

imnasnainaec:
@katherinejensen00 Review recommendation partially implemented. One open question for you that we need to decide before this PR can be buttoned up:

* Is the review item 4 split an accident to be normalized away, or a contract to document?

katherinejensen00 (commented):
@imnasnainaec Thanks for the follow-up and sorry for my mess up on item 4. 

Answering your question first, then the high-priority fixes from this round.

## Item 4: accident or contract?

Accident. I'd normalize it away in this PR rather than document it.

Here's why it can't work as a contract. An empty `<LanguageName />` in `InterlinearSetup.xml` comes back as `""`, but the same setup rebuilt from legacy settings comes back with no language name at all. An empty export id setting on the legacy path also comes back as null, where the setups file gives `""`. A consumer can't tell which path produced a setup, so a per-field rule is unusable.

Nothing reads a setup string field yet (the interlinearizer extension only reads `books`) and the type is `@experimental`, so this is the cheapest it will ever be to fix.

Suggested shape:

- In `ConvertPt9InterlinearSetup`, map empty to null for `LanguageId`, `LanguageName`, and both id `ToString()` results. That's the same `IsNullOrEmpty` guard already used on `FontName`, `ExportScrTextName`, and `SenseId`.
- The record doc and `.d.ts` shrink to one sentence that's true on both paths: a string field that is empty in the project is absent.
- The four `Is.Empty` assertions become `Is.Null`, and they can still fail.
- Update the PR title and risk line, and the "Served payload" section of `pt9-xml.md` in interlinearizer-extension.

If you'd rather keep this PR doc-only: keep "treat alike", drop the per-field empty-string sentence, loosen those four assertions to `Is.Null.Or.Empty`, and ticket the normalization. (My round-1 worry that this makes the test vacuous was overstated. The three `Is.Null` checks on that row still fail if a guard is removed.)

## High-priority fixes

1. **`platform-scripture.d.ts:1470-1472`: the "can arrive as `null`" sentence is wrong. That was my mistake in round-1 item 8.** PAPI's `deserialize` turns every nested `null` into `undefined` (`lib/platform-bible-utils/src/serialization.ts:56-76`). So an extension never sees `null` here, and `?: string` already matches what it receives. Please drop the null/JSON-key sentences. "Check for a falsy value" is fine to keep.

2. **`Pt9InterlinearData.cs:164`: the malformed-id sentence needs scoping to the setups file.** On the legacy path, `HexId.FromStrSafe` treats a malformed id as absent rather than failing the read. The existing `GetPt9InterlinearData_ToleratesAMalformedLegacyExportTextId` pins that. Your new test's description already says "An id in the setups file…", so the doc just needs to match.

3. **Only if the PR stays doc-only: `Pt9InterlinearData.cs:154` and `:158` still don't say which path they describe.** "`LanguageId` and `LanguageName` serve the project's text verbatim" and "An empty one of those four serves as an empty string" are both false for legacy-rebuilt setups. Your new legacy test asserts the opposite, and its comment refers to a caveat the doc doesn't have. Normalizing makes this one go away.

> @katherinejensen00 reviewed 4 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2789#-:-P1VruitDwewYAgNRX7a:bs14dnc)* status: all files reviewed, 1 unresolved discussion (waiting on imnasnainaec).

imnasnainaec (commented):
> @imnasnainaec made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2789)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on katherinejensen00).

___
*[`c-sharp/Projects/Pt9InterlinearData.cs` line 152 at r2](https://reviewable.io/reviews/paranext/paranext-core/2789#-P1IazpEBnk4IH4nKGxw:-P1WQ0s66JXXhQ7uY-iQ:b-896fix) ([raw file](https://github.com/paranext/paranext-core/blob/859f4322ecfd35b379bd8b13995db5361a725fc2/c-sharp/Projects/Pt9InterlinearData.cs#L152)):*

## 1. High — `c-sharp/Projects/Pt9InterlinearData.cs:152`

The new rule is still wrong, on a path that runs in the ordinary case.

"…so an empty one serves as an empty string and only a field the project omits is absent" holds
on the setups-file path, but not on the legacy-settings path in `ResolvePt9InterlinearSetups`
(`Pt9InterlinearReader.cs:451-520`). There, `ExportScrTextId = HexId.FromStrSafe(GetSetting(...))`
(`:481`, assigned `:514`). `ProjectSettings.GetSetting` returns the same `""` whether a setting is
stored empty or absent entirely, and `FromStrSafe("")` returns null — so stored-empty,
stored-garbage and omitted all collapse to null before `ConvertPt9InterlinearSetup` ever runs.
`ExportScrTextId` can never be `""` on that path. `LanguageName` is never assigned there at all
(`:500-516`), so it is unconditionally null regardless of what the project stores.

Reproduced two ways independently: setting `InterlinearExportTextId.<model>` to `""` and reading
through the provider yields `ExportScrTextId = null`, `LanguageName = null`. And your own untouched
test `GetPt9InterlinearData_ToleratesAMalformedLegacyExportTextId` (`tests:586`) already asserts
`Is.Null` for a value that IS stored.

This is not a hypothetical path: the guard at `Pt9InterlinearReader.cs:456-462` only skips the merge
for an observer or an already-converted user, so the legacy path runs on the ordinary first read.

Worth noting this is the prior round's second finding, which asked you to settle which side the id
fields land on and flagged "absent *or* empty depending on which path produced the setup — itself
worth stating rather than omitting." That is still open.

Cheapest fix is a clause, not a redesign — scope the sentence to the setups-file path and say the
legacy path only ever serves these as null or a full id.
</blockquote></details>

Done.

katherinejensen00 (approved):
@imnasnainaec Approving. The normalization is right and the tests hold up under mutation: I removed each of the six NullIfEmpty calls singly and every one turns a test red. All eight string members check out on both construction paths, so the sentence is finally true. 16 of the 18 round-1 findings are closed, and item 8 was my error.

Nothing blocking. Four non-blocking follow-ups, take or leave them in this PR or a later one:

1. Pt9InterlinearReader.cs:545-547 — the converter doc still exempts the model id ("serves whenever present … every other string") from the rule you just applied to it at :562. Your own new test asserts the opposite.
3. platform-scripture.d.ts — the id paragraph didn't cross over. The half worth carrying is the failure mode: one bad id in InterlinearSetup.xml fails setups, books, lexicon and word analyses together. No TODO wanted — that policy is already in the ADR.
4. Pt9InterlinearReader.cs:667 — SenseId still has the inline guard, so the class has two idioms for one rule.

> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2789#-:-P1YYVHe6LWmuLlmgut0:b6hpb0z)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

imnasnainaec (commented):
1. What looks like a model id exemption is just describing how PT9 can mint an id even when there's no model present (and thus even the model name is absent), and that non-empty id is preserved.
2. Agreed. Sentence added to the `Pt9InterlinearProjectData` docs.
3. Agreed. Guard updated to use the new idiom.

> @imnasnainaec reviewed all commit messages and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2789#-:-P1anx-G7RwHXWKN2s1y:bp2klae)* status: 2 of 4 files reviewed, all discussions resolved (waiting on katherinejensen00).

PR #2790, opened 2026-09-09
PT-4548: ProjectSelector type indicator and per-bucket row order

jolierabideau:
## What this ships

Two additions to `ProjectSelector` (which stays on `platform-bible-react/experimental`), plus the
tests, story and ADR entries for them:

- **`renderProjectIndicator`** — a render prop for a per-row indicator, typically a small icon
  distinguishing one kind of row from another. The selector ships no taxonomy and no default
  mapping; the caller reads its own `customData` and decides what a value looks like. The column is
  reserved for every row once a renderer is supplied, so a renderer that marks only some rows does
  not indent their labels past the rest. The returned node must carry its own accessible name
  unless the row text already conveys the distinction, and the renderer must be pure and hook-free
  — both are stated in its TSDoc and demonstrated by the reference story.
- **`ProjectSelectorGrouping.compareProjects`** — an optional comparator letting a grouping
  descriptor own the row order inside its own buckets. Ties fall back to the selector's canonical
  order, which keeps a project fanned across several scroll groups in a stable sequence. Ignored
  where the descriptor does not own the bucketing: the `openTabs` and `selection` groupings, any
  grouping with no `getGroupKey`, and the unknown bucket.

Serves PT-4548.

### Why not `customSections`

This branch originally added a parallel sections API — ordered `{ id, label, match, compare? }`
descriptors selected by a `'custom'` grouping option. [#2673](https://github.com/paranext/paranext-core/pull/2673)
then landed on `main` and made the bucketing half of that redundant: `getGroupKey` /
`getSectionHeading` / `compareSections` already let a consumer express "Recent / Your projects".
Only row order *within* a bucket was still unreachable, so `customSections` was dropped in the merge
and `compareProjects` fills the remaining gap. Recorded as
`adr-project-selector-per-bucket-row-order`.

`ProjectSelector` stays on `experimental` (`adr-project-selector-stays-experimental`), whose entry
criterion for a later promotion is that a consumer can be added without adding a prop.

## Also in the diff

- Two new localized-string pass-throughs on `SettingsSidebar` (`searchPlaceholderText`,
  `noResultsText`) so its project picker's popover is no longer hard-coded English, with tests.
- Coverage for the manage-books and settings-sidebar pickers' grouping restrictions.

## Open for the reviewer

- **Whether the view-options trigger should read "modified" for a grouping change as well as an
  active filter.** A UX judgement call, deliberately not guessed at.

## Not in this PR

Four pre-existing defects that this review turned up are on `main` already and are filed separately
rather than absorbed here: the partial-localized-strings crash in `resolveStrings`, colliding
section React keys, the dead `data-selected` attribute, and `compareRows` rebuilding a collator per
comparison. See the follow-up note; none is caused by this branch.

---

AI-assisted — review and fixes performed with Claude Opus 5 in Claude Code.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2790)


katherinejensen00 (changes requested):
# PR #2790 review — ProjectSelector custom sections and type indicator

For: @jolierabideau
Reviewed: `f0ebe77b53a..969974927b7`, 8 source files (+454/−3), ignoring `dist/`. Since #2673 has merged, `f0ebe77b53a` is now the real merge base, so the PR page's diff is accurate and the "synthetic base" note in the body is out of date.
Not tested: this review checkout has no installed `node_modules`, so every finding comes from reading the source and git history.

## Overview

The code that ships is small and mostly sound. The main problems are that the ADRs and the PR body still describe the branch as it was before the merge with main, and a few behaviors the body promises (aligned labels, a screen-reader name for the type, ids matched regardless of case) were lost in that merge.

Merge `36c3cded8f5` explains why `customSections` was dropped (main's grouping descriptors already cover it) and adds `compareProjects` to fill the one gap. That decision is reasonable. But the merge also quietly dropped several other things the body still claims, and the three new ADR entries were never rewritten to match. What this PR actually ships is `renderProjectIndicator`, `ProjectSelectorGrouping.compareProjects`, a story, and tests.

Severity counts: 0 Critical, 7 Important, 10 Minor.

---

## Important

### 1. The custom-sections ADR records an API this PR no longer ships
`.context/standards/Architecture-Decisions.md`
Line: 3609
Severity: Important

Comment:

This entry is marked Accepted, but it describes `customSections`, a `'custom'` grouping option and `RowSection.id`, and none of them exist on the branch.

The merge commit explains well why `customSections` was dropped: main's descriptors already express it, and `compareProjects` fills the gap. But that decision is only in the commit message, and this entry tells the next reader the opposite. People and the investigation agents read this log for precedent. Since the entry never reached main, I'd rewrite it to record the decision that shipped ("extend `ProjectSelectorGrouping` with `compareProjects` rather than add a parallel sections API") or drop it. No retirement stub is needed. The Consequences line "`RowSection` needed an `id`" is also false: `RowSection` at `project-selector.rows.ts:406` has no `id`.

### 2. The stays-experimental ADR's supporting evidence isn't in this PR
`.context/standards/Architecture-Decisions.md`
Line: 3640
Severity: Important

Comment:

The reasoning here depends on changes this branch no longer makes:
- Lines 3640 and 3646 list `customSections` as a capability that "arrived".
- Lines 3651–3652 say "this work item renamed `hideFilterMenu` to `hideViewOptionsMenu` and the `filterGroupBy*` string keys to `groupBy*`, with no deprecation shims".

Neither prop exists anywhere in the tree. The `groupBy*` names came from #2673. And the old `%manageBooks_projectSelector_filterGroupBy*%` keys do have `fallbackKey`/`deprecationInfo` shims, so "no deprecation shims" is also wrong.

The decision itself (stay on `experimental`) still holds. It just needs evidence that is true today. `renderProjectIndicator` and `compareProjects` arriving after promotion was first proposed would work.

### 3. The type-stays-free-form ADR is about a field that doesn't exist
`.context/standards/Architecture-Decisions.md`
Line: 3662
Severity: Important

Comment:

The title and Context talk about `ProjectSelectorProject.type` and "adding a discriminator". On main, `type` is just a well-known key inside the untyped `customData` bag, and this PR adds no discriminator.

Two statements in Consequences are also wrong:
- "every picker must supply both" is not true: `typeName` is optional, and headings fall back to the raw key (`project-selector.groupings.ts:169`).
- No production picker passes `renderProjectIndicator`.

`adr-project-selector-consumer-driven-groupings` (line 3570, from #2673) already rejects a typed `customData`, so this entry may be redundant. I'd fold anything new into that one or drop this entry.

### 4. The indicator TSDoc recommends `aria-hidden`, but nothing else names the type any more
`lib/platform-bible-react/src/components/advanced/project-selector/project-selector.component.tsx`
Line: 329
Severity: Important

Comment:

As merged, a screen-reader user may have no way to tell a resource from a project. Before the merge (`c4880b745de`), each row rendered `<span className="tw:sr-only">{row.typeName}</span>` and the tooltip showed a type line, so an `aria-hidden` glyph was safe. The merge dropped both, and the tests that covered them.

This TSDoc still offers `aria-hidden` as an equal choice, and the reference story and the test both use `aria-hidden` icons. A consumer who copies the story ends up with a type shown only visually, which fails WCAG 1.1.1.

"Output is treated as decorative" is also inaccurate: the wrapper span sets no `aria-hidden`, so the caller's output is exposed as-is. Either restore the `sr-only` type name, or change the doc to say the caller must give the indicator an accessible name unless the row text already conveys the type. In that case, update the story to match.

### 5. The indicator column is no longer reserved, so labels misalign
`lib/platform-bible-react/src/components/advanced/project-selector/project-selector.component.tsx`
Line: 532
Severity: Important

Comment:

The column only appears on rows where the renderer returns something truthy. So a renderer that shows an icon for resources only (returning `null` for projects, which is the natural shape) pushes resource labels one glyph plus `gap-2` to the right of project labels.

Before the merge there was a fixed `tw:h-4 tw:w-4` slot on every row, keyed on whether the prop was supplied (`reserveIndicatorSlot={Boolean(props.renderProjectIndicator)}`). The PR body still says "labels stay aligned when the renderer returns nothing". Keying on the prop also fixes a smaller issue: `{indicator && …}` prints a literal `0` if a renderer ever returns the number 0.

```tsx
{reserveIndicatorSlot && (
  <span className="tw:flex tw:h-4 tw:w-4 tw:shrink-0 tw:items-center tw:justify-center">
    {indicator}
  </span>
)}
```

### 6. Case-insensitive selection matching was dropped in the merge
`lib/platform-bible-react/src/components/advanced/project-selector/project-selector.rows.ts`
Line: 265
Severity: Important (this line is outside the diff, so it needs a file-level or PR-level comment)

Comment:

The PR body says selection ids are compared "case-insensitively in all three modes", but after the merge they aren't. The merge took main's `rows.ts`, and this branch's `normalizeProjectId` fixes and their four casing tests went with it.

These places now compare exactly:
- `pairIsSelected` (265)
- single-select `isSelected` (292)
- the bound-but-closed lookup (363, 367)
- the trigger label lookups (`project-selector.component.tsx` 936, 956, 986)

`handleRowClick` still normalizes (component.tsx:884–886), so display and click now disagree. In `project-multi`, a pair `abc` against project `ABC` shows the row unchecked and the trigger shows its placeholder while "Clear all (1)" is visible, and clicking the row removes the pair.

No in-repo path hits this today; it takes persisted state or a web view opened with a differently-cased id. Since the body promises this behavior, either restore the fix or remove the claim and track it separately.

### 7. Three of the four new `compareProjects` tests can't fail
`lib/platform-bible-react/src/components/advanced/project-selector/project-selector.rows.test.ts`
Line: 657
Severity: Important

Comment:

The fixture is already in alphabetical order (aaa, bbb, ccc). `computeRows` keeps input order, and `Array.prototype.sort` is stable. So:
- "falls back to canonical order when the comparator reports a tie" (687) stays green if `|| compareRows(a, b)` is deleted from `rows.ts:549`;
- "uses canonical order … when no comparator is supplied" (692) stays green if the `.sort(compareRows)` branch is deleted;
- "leaves the unknown bucket canonically ordered" (697) checks the order of a one-element array (`['zzz']`), so it passes whatever the sort does.

The comment at 657 ("canonical ordering cannot produce this result by accident") only protects the first test, at 679. The tie case the TSDoc actually describes (one project in several scroll groups) isn't tested at all.

Suggested fix:
- list the fixture out of order (e.g. ccc, aaa, bbb);
- give the unknown bucket at least two rows whose recency order differs from alphabetical;
- add a `project-multi` case with one project in two scroll groups.

---

## Minor

### 8. The "no indicator when prop is absent" test can't fail
`lib/platform-bible-react/src/components/advanced/project-selector/project-selector.component.test.tsx`
Line: 551
Severity: Minor

Comment:

The `indicator-*` test ids only exist when the other test's own render prop creates them, so `queryByTestId(/^indicator-/)` can never match here. The test stays green even if the `{indicator && …}` guard is deleted, which would add an empty flex item and a `gap-2` to every row of every existing picker. To make it able to fail, assert on the row's structure, e.g. that the label is the element directly after the check column.

### 9. `compareProjects` names recency as its use case, but the built-in grouping doesn't use it
`lib/platform-bible-react/src/components/advanced/project-selector/project-selector.rows.ts`
Line: 150
Severity: Minor

Comment:

The TSDoc motivates `compareProjects` with a most-recently-used bucket ("alphabetical order defeats its purpose"). But `makeBuiltInGroupings().lastUsed` (`project-selector.groupings.ts:127–145`) doesn't set one, so "Recently used" stays alphabetical in every in-repo picker. Meanwhile `lib/platform-bible-utils/src/project-selector-custom-data.ts:47–51` still says that is intended ("the magnitude is never compared"), so the two docs now disagree. Either wire a recency comparator into the built-in (and update the utils doc), or pick a motivating example that doesn't imply the built-in behaves that way.

### 10. The `compareProjects` contract is incomplete, and `sortBucket` can be simpler
`lib/platform-bible-react/src/components/advanced/project-selector/project-selector.rows.ts`
Line: 155
Severity: Minor

Comment:

The TSDoc has three gaps:
- It doesn't say which sign puts `a` first.
- It doesn't say the comparator is also ignored for groupings with no `getGroupKey` (line 541 returns `partitionFlat` first).
- It doesn't explain why the unknown bucket is exempt. Mixing two orders in one list, such as a recency-sorted list with an alphabetical "Other" at the bottom, may surprise callers.

Separately, `sortBucket` at 544–550 is used once and repeats the sort in both branches. It also calls `compareProjects` detached from its object, unlike `getSectionHeading`/`compareSections`. This one-liner behaves the same, including for 0, -0 and NaN:

```ts
const sortedRows = [...groupRows].sort(
  (a, b) => grouping.compareProjects?.(a.project, b.project) || compareRows(a, b),
);
```

### 11. The `renderIndicator` `useCallback` memoizes nothing
`lib/platform-bible-react/src/components/advanced/project-selector/project-selector.component.tsx`
Line: 847
Severity: Minor

Comment:

`renderIndicator` is only called during render (1176) and passed into `ProjectRowView`, which isn't memoized, and the prop it depends on is usually an inline arrow anyway. The destructure is only there to satisfy the hook's dependency rule. This is equivalent and keeps the file's `props.X` style:

```tsx
indicator={props.renderProjectIndicator?.(row.project)}
```

### 12. The renderer runs for every row on every render, so it must not use hooks
`lib/platform-bible-react/src/components/advanced/project-selector/project-selector.component.tsx`
Line: 1176
Severity: Minor

Comment:

The caller's renderer runs inside `ProjectSelector`'s own render, once per filtered row, on every render. A caller who uses a hook inside it (say `useLocalizedStrings` for an icon label) would change the component's hook count whenever filtering changes the row count, and React would throw. Please say in the TSDoc that it must be pure, cheap and hook-free, or call it inside `ProjectRowView` so each row owns its own call.

### 13. `customData.type` can't tell a resource from a project in manage-books
`lib/platform-bible-react/src/components/advanced/project-selector/project-selector.component.tsx`
Line: 323
Severity: Minor

Comment:

The TSDoc says to tell projects from resources using `customData.type`. The only in-repo code that sets it (`manage-books.web-view.tsx:380`) stores the Paratext ProjectType (`ProjectSummary.cs:78`). Whether something is a resource is a separate `IsResource` flag (`ProjectSummary.cs:83`) that never reaches `customData`. So in the manage-books header picker, a Standard-type resource and a Standard project look the same to this API. I'd soften the wording ("derived from caller-supplied `customData`"), or note that callers need to pack a resource flag.

### 14. The story contradicts its own description
`lib/platform-bible-react/src/components/advanced/project-selector/project-selector.stories.tsx`
Line: 753
Severity: Minor

Comment:

The renderer uses the book icon only for `'ScriptureResource'`, so the fixture's `CommentaryResource` row gets the document icon. The description at 766 says "a book icon for resources". The description's reason for `aria-hidden` ("the row text already names the project") also doesn't hold in this flat story, where the icon is the only thing showing the type (see #4). Callers copy this story, so it's worth fixing both. The ~90-word description would also read faster as three short sentences.

### 15. The manage-books describe comment is out of date
`extensions/src/platform-scripture/src/manage-books-dialog/manage-books-dialog.component.test.tsx`
Line: 152
Severity: Minor

Comment:

The comment says the dialog's rows carry no last-used data and would be filed under "Unknown". But the sidebar rows do carry `lastUsedAt` (the first test at 193 expects "Last used"), and the built-in heading is "Other", not "Unknown". Only the Copy-from picker drops `lastUsed`, and `MANAGE_BOOKS_COPY_FROM_GROUPING_IDS` already documents why. I'd delete this preamble. Line 218 can go too, since the exact `toEqual` on 217 already proves "Last used" is absent.

### 16. The manage-books picker tests use a made-up grouping list and unscoped queries
`extensions/src/platform-scripture/src/manage-books-dialog/manage-books-dialog.component.test.tsx`
Line: 156
Severity: Minor

Comment:

Three problems here:
- `WIRING_GROUPINGS` builds its own grouping list instead of using the real wiring (`manage-books.web-view.tsx:550–556`). So "every grouping the wiring layer supplies" pins the order None, Open tabs, Type, Last used, while users actually see None, Open tabs, Last used, Type.
- `openGroupingMenu` (186–191) finds "Group by" anywhere in the document, even though its doc says the button is inside the picker's popover. `groupingChoices` (183) also collects menu items from the whole document, and the dialog has a second "Group by" control (the book grid's toggle group), kept apart only by role.
- At 211–212 the copy test waits on `isSectionActive('copy')`, which is already true before projects load, then queries a picker that renders only after they load. It passes on timing.

Building the list with `makeBuiltInGroupings`, using `within(popover)` in the helpers, and switching to `await screen.findByRole(…)` would make these tests robust.

### 17. The settings-sidebar test comment is out of date, and one check can't fail
`lib/platform-bible-react/src/components/advanced/settings-components/settings-sidebar.component.test.tsx`
Line: 79
Severity: Minor

Comment:

The comment says the test catches "the restriction deleted". After the merge, `SettingsSidebar` has no restriction; the menu is missing only because `openTabs={[]}` produces too few groupings. "Every grouping … yields the same flat list" is also wrong, since lastUsed, language and type would add headed sections.

Line 85 (`queryByRole('menuitemradio')`) can never fail, because the menu is never opened and Radix unmounts closed menu content. Line 84 is the real check. I'd drop line 85 and cut the comment to one line saying what 84 protects.

---

## Outside this PR: consider filing separately

These are on main already. The PR didn't cause them, but the review turned them up.

- `project-selector.component.tsx:193`: `resolveStrings` spreads `partial` over `DEFAULT_STRINGS`, but `buildProjectSelectorLocalizedStrings` sets every key, using `undefined` for missing ones. The `undefined` overwrites the default, so a consumer passing a partial lookup can crash at `strings.boundButClosedTooltip.replace` (565) and see "undefined (n)" on "Clear all". In-repo callers are safe because `useLocalizedStrings` fills every key.
- `project-selector.component.tsx:1164`: section React keys collide when a bucket key equals the grouping's `unknownSectionHeading` (e.g. a language literally named "Unknown language").
- `project-selector.component.tsx:527`: `data-selected={row.isSelected}` is overwritten by cmdk's own `data-selected` (the keyboard highlight), so it has no effect.
- `project-selector.rows.ts:438`: `compareRows` builds a new collator for every comparison via `localeCompare` with options. That's about 70 ms per sort at 1,000 rows, versus about 1 ms with a shared `Intl.Collator`.

## PR title and description (PR-level comment)

Severity: Important

Comment:

The title and most of the description describe the branch before the merge, which will send reviewers looking for code that isn't here. Among the claims that no longer hold:
- `customSections` / `partitionByCustomSections` / `ProjectSelectorSection`;
- the `hideFilterMenu` and `filterGroupBy*` renames;
- `viewOptions*` strings;
- `project-selector.test-utils.ts`;
- the new experimental exports (`ProjectSelectorProps`, `ProjectSelectorSelection`, …);
- aligned indicator labels;
- case-insensitive selection in all three modes.

The "Review scope" section is also out of date now that #2673 has merged. Could you retitle it (e.g. "PT-4548: ProjectSelector type indicator and per-bucket row order") and replace the body with a short summary of what ships? That's `renderProjectIndicator`, `compareProjects`, the story, the tests, and the ADR changes, plus one line on why `customSections` was dropped in favor of `compareProjects`. If the review transcript is worth keeping, a PR comment is a better home for it.

> @katherinejensen00 partially reviewed 19 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2790#-:-P1loIcc9v9vIj0aVFC-:bmdipoc)* status: all files reviewed, 1 unresolved discussion (waiting on jolierabideau).

___
*[`.context/standards/Architecture-Decisions.md` line 1 at r1](https://reviewable.io/reviews/paranext/paranext-core/2790#-P1loRpB5fIlIHKuSgeC:-P1loRpB5fIlIHKuSgeD:b-tenb7u) ([raw file](https://github.com/paranext/paranext-core/blob/1308062748afbc7bc2b5b733de8a6292e6f44958/.context/standards/Architecture-Decisions.md#L1)):*
> ```Markdown
> # Architecture Decisions
> ```

### 1. The custom-sections ADR records an API this PR no longer ships
`.context/standards/Architecture-Decisions.md`
Line: 3609
Severity: Important

Comment:

This entry is marked Accepted, but it describes `customSections`, a `'custom'` grouping option and `RowSection.id`, and none of them exist on the branch.

The merge commit explains well why `customSections` was dropped: main's descriptors already express it, and `compareProjects` fills the gap. But that decision is only in the commit message, and this entry tells the next reader the opposite. People and the investigation agents read this log for precedent. Since the entry never reached main, I'd rewrite it to record the decision that shipped ("extend `ProjectSelectorGrouping` with `compareProjects` rather than add a parallel sections API") or drop it. No retirement stub is needed. The Consequences line "`RowSection` needed an `id`" is also false: `RowSection` at `project-selector.rows.ts:406` has no `id`.

jolierabideau:
Fixed in 25f076a8be1, and the branch is up to date with `main` (`ca70260c150`, 0 behind). Title and
body rewritten. Every finding was checked against the tree before acting — all 17 held, so nothing
below is a pushback.

Verification: `npm run typecheck` (all four checks exit 0, including `typecheck:erb`, which was the
environmental failure last round), repo-wide `npm run lint` exit 0, `platform-bible-react` unit 963
passed, `platform-scripture` 645 passed. The only red in the react workspace is 5 storybook stories
under `src/renderer/components/first-run/` — confirmed pre-existing by stashing this work and
re-running them on a clean tree, and this branch touches no file there.

## Important

**1. Custom-sections ADR.** Rewritten rather than dropped, since the decision that shipped is worth
recording. Slug retired and replaced with `adr-project-selector-per-bucket-row-order`
("extend `ProjectSelectorGrouping` with `compareProjects` rather than add a parallel sections API"),
with the `#2673`-made-it-redundant reasoning from the merge commit moved into its Context, and the
false `RowSection.id` consequence gone. Byte-order position verified with `LC_ALL=C sort`; it lands
where the old entry was.

**2. Stays-experimental ADR.** Evidence replaced. The `customSections` mentions are gone, and the
"this work item renamed `hideFilterMenu`…with no deprecation shims" sentence is gone — I confirmed
both halves were wrong: neither prop name exists in the tree, and
`%manageBooks_projectSelector_filterGroupByOpenTabs%` does carry a `fallbackKey` +
`deprecationInfo`. The "promote now" rejection now cites #2673's own reshape (~16 props removed,
fixed grouping props replaced by descriptors) as the churn evidence, which is true and is the
immediately preceding work item.

**3. Type-stays-free-form ADR.** Dropped. It recorded no decision this PR makes, and
`adr-project-selector-consumer-driven-groupings` already rejects a typed `customData` for the same
reason. Its concrete PT9/DBL vocabulary research stays in the branch history rather than in the log.

**4. Indicator accessibility.** Took the second option: the doc now says the caller **must** give
the node an accessible name unless the row text already conveys the distinction, and says the
selector adds no `aria-hidden` and no description of its own. Rather than restore the `sr-only`
type name, which would put the selector back in the business of naming a value it cannot interpret.
The reference story's icons are now labelled `Resource`/`Project`, and its description explains when
`aria-hidden` is the right choice instead.

**5. Reserved indicator column.** Fixed as suggested — `reserveIndicatorSlot={Boolean(props.renderProjectIndicator)}`
with the fixed `h-4 w-4` slot, which also kills the literal-`0` case. New test renders a
resources-only renderer and asserts the label sits at the same child index on both rows; verified
failing against the pre-fix source.

**6. Case-insensitive selection.** Restored rather than dropping the claim — five sites:
`pairIsSelected`, single-select `isSelected` (with an explicit `undefined` guard so "nothing
selected" cannot become "everything matches"), the bound-but-closed dedupe and its project lookup,
and all three trigger-label lookups via a new `findProjectById` helper. Eight tests added across
`rows.test.ts` and `component.test.tsx`; six fail against the pre-fix source. The other two are
guards that need both halves of the fix present to be meaningful — I checked the dedupe one is
falsifiable by mutating only the dedupe comparison back to raw.

**7. `compareProjects` tests.** All three suggestions taken. Fixture is now input order BBB, CCC,
AAA against canonical AAA, BBB, CCC and recency CCC, BBB, AAA — three pairwise-different orders, so
no assertion can be satisfied by a stable no-op sort. Unknown bucket has two rows whose recency order
is the reverse of alphabetical. Added the `project-multi` case with one project in two scroll groups.

One disclosure on that last test: it is *not* sensitive to deleting the `|| compareRows(a, b)`
fallback outright, because `collectOpenTabsByProject` already sorts a project's tabs by scroll
group, so the input order happens to match. It does fail when the tie-break is reversed, so it pins
the direction rather than the existence. The bucket tests cover the existence.

## Minor

**8.** Rewritten to assert the row's child order (label directly after the check column, index 1
with no renderer, index 2 with one). Verified it goes red when the slot is rendered unconditionally.

**9.** Reworded — the motivating example is now a caller-side score, and the doc states explicitly
that built-in `lastUsed` supplies no comparator and stays alphabetical. No longer contradicts
`project-selector-custom-data.ts`.

**10.** All three gaps documented (sign convention; ignored for groupings with no `getGroupKey`;
why the unknown bucket is exempt, including the mixed-order warning for callers). `sortBucket`
replaced with your one-liner.

**11.** `useCallback` removed; now `indicator={props.renderProjectIndicator?.(row.project)}`.

**12.** TSDoc now states the renderer must be pure, cheap and hook-free, with the reason (row count
changes as the user filters, so a hook there changes the selector's hook count between renders).
Left it at the doc rather than moving the call into `ProjectRowView`, since that is a real
constraint either way and the doc is what a caller reads.

**13.** Softened — the doc now says the indicator derives from the caller's own `customData` and
calls out that the conventional `type` key carries a project *type*, not a project/resource
discriminator, so a caller needing the latter has to pack its own flag.

**14.** Both fixed. The renderer now matches any type ending in `Resource`, so the fixture's
`CommentaryResource` row gets the book icon; the description matches, is three short sentences, and
explains the labelling rather than claiming the row text already names the type.

**15.** Preamble deleted and the redundant `queryByRole('menuitemradio', { name: 'Last used' })`
line dropped — the `toEqual` above it already proves the absence.

**16.** All three fixed. `WIRING_GROUPINGS` is now
`makeBuiltInGroupings().filter((g) => MANAGE_BOOKS_PROJECT_SELECTOR_GROUPING_IDS.includes(g.id))`,
which confirmed your point: the real order is **None, Open tabs, Last used, Type**, not the
hand-listed one the test pinned. `openGroupingMenu` now resolves the popover first and uses
`within(popover)`. The copy test awaits `findByRole('combobox', { name: 'Select project' })`
instead of the section flag.

**17.** Comment rewritten to say what is actually true (no `availableGroupings`, no open tabs →
too few groupings for a menu), and the `menuitemradio` line dropped.

## Outside this PR

Filed separately as a four-item follow-up, all re-verified at `ca70260c150` rather than taken on
trust: the `resolveStrings` partial-lookup crash (`buildProjectSelectorLocalizedStrings` sets every
key from a `string | undefined` reader, so an explicit `undefined` overwrites the default it lands
on — `boundButClosedTooltip.replace` then throws), the colliding section React keys, the dead
`data-selected`, and the per-comparison collator in `compareRows`. Happy to fold any of them in here
if you would rather not have a separate issue.

## Still open for you

The one UX call from the last round — whether the view-options trigger should read "modified" for a
grouping change as well as an active filter — is unchanged and still needs a decision rather than a
code change. It is called out in the PR body.

katherinejensen00 (approved):
:lgtm:

> @katherinejensen00 partially reviewed 28 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2790#-:-P23boQTBdM0XH8wtbfK:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2748, opened 2026-09-01
PT-4424: Cover the missing-book banner in the reference panels

captaincrazybro:
# PT-4424 — Cover the reference panels' missing-book banner with tests

**Branch**: `worktree-pt-4424-reference-panel-regression-tests` · **Base**: `origin/main` ·
**Files changed**: 11 · **Review model**: Claude Opus 5

## What this does

PT-4424 asks that the Model text, Bible texts and Commentaries panes show a "book not in this text"
banner and rerender when the user navigates to a book a resource does not contain, with a regression
test pinning it. The banner itself already shipped; this branch supplies the coverage for the two
surfaces that had none. The Model text pane was already covered by
`model-text-panel.component.test.tsx`, so all three panes named in the ticket now have tests.

To make the Bible texts / Commentaries panel testable at all, `ResourceTextPanel` is extracted out of
`resource-text-panel.web-view.tsx` into `resource-text-panel.component.tsx`, mirroring the
`model-text-panel` split already on main.

## The extraction boundary

The web view keeps every PAPI hook and everything derived from one: the filtered rows, the two-stage
`readiness`, `resolveResourceSelection` and its apply effect, `resourceProjectId`, `dblMatch`, the
dynamic-title effect, the auto-install, `useCommentaryMarkerStyles`, `useOpenFindShortcut`,
`usePublishNavigableProjectIds`, the `ChapterUSJ` subscription and the text-direction setting. It also
keeps the entire pick path, **including `useDialogCallback`** — that hook supplies
`maximumOpenDialogs: 1` and an unmount guard, and hand-rolling either in the component would be
strictly worse than keeping it.

The component takes the **already-resolved** row as a prop rather than deriving it again, so there is
only ever one answer to "which resource is on screen". That matters for a reason worth stating: the
web view needs that answer to key the `ChapterUSJ` subscription that produces the `usjPossiblyError`
it passes back *down*, so the dependency is circular and the panel cannot report the resolution
upward instead. An earlier revision of this PR had both sides deriving it independently and needed a
shared helper to keep them honest; that duplication is now gone entirely.

Result: a 21-prop component and a web view that holds only PAPI wiring.

One caveat since the rebase: the component is **not** purely props-driven any more. #2749's
reveal-scroll effect moved into it (see the hidden-tab section), and that effect reads the editor DOM
directly. Its dependencies all live here, so this is where it belongs — but the component does touch
`document`.

**One correction to the "mirrors the model-text split" framing above:** the file layout mirrors it,
[...]

PR #2791, opened 2026-09-09
PT-4350: Honest resource-panel load state and a gated selection

jolierabideau:
# Code Review Summary

**Branch**: pt-4350-resource-panel-load-and-selection-gate

**Base**: origin/main

**Date**: 2026-09-13

**Files changed**: 18

## Overview

Three defects in the Bible texts / Commentaries resource panel, each measured in the running app
rather than inferred. The panel was mounting the editor empty and painting Lexical's "Enter some
Scripture…" prompt — an edit invitation in a text the reader cannot edit — because the chapter
subscription was seeded with a blank USJ (which is neither `undefined` nor falsy, so the two guards
written for exactly this case were unreachable) and because the editor feed ran as a passive effect,
after paint. Separately, a stored resource pick was being overwritten while the DBL catalog was
still loading (**PT-4470**), because rows are filtered against that catalog and a resource whose
entry had not arrived was simply *absent* — which auto-correct read as "gone".

This review was run after merging `origin/main`, which had landed **PT-4424 (#2748)** splitting
`resource-text-panel.web-view.tsx` into a container plus a new presentational
`resource-text-panel.component.tsx`. The branch's changes were re-applied across that boundary as
part of the merge and the analysis passes verified the re-application hunk-by-hunk. The review then
found that the retry affordance the branch introduced produced **no visible feedback at all** in its
main case, and that the `useLayoutEffect` change did not answer the hidden-view rule. Both were
fixed, along with eight other findings.

## Acceptance criteria

PT-4350 ("Opening a resource always correctly loads it", NN 5B under epic PT-4336; also closes
PT-4470) has four acceptance criteria, deliberately split across two stacked PRs.

| AC | Covered | Where |
| --- | --- | --- |
| AC1 — the panel never mounts an editor with nothing in it | Yes | This PR: `undefined` seed, the visibility-gated pre-paint feed, and the labelled waiting state |
| AC2 — switching the resource must not leave the previous resource's text visible | **No, by design** | Stacked PR **#2793**, "Stop a resource switch showing the previous resource's text", which adds an `isAnswerCurrent` identity check. Verified open and based on this branch. **Not a gap in this PR.** |
| AC3 — overlapping resource picks must not disarm each other | Yes | This PR: the `selectGenerationRef` supersession guard in `handleResourceSelect` |
| AC4 — a failed chapter read must offer a retry | Yes | This PR: the `chapterRetryNonce` resubscription, and (added this round) the feedback that makes it legible |

## API Changes

None. No public API surface changed — no `lib/platform-bible-react/`, `lib/platform-bible-utils/`,
`lib/papi-dts/papi.d.ts`, or `extensions/src/*/src/types/*.d.ts` files were modified.

Internal (extension-local) signature changes, including those made during review:

- `resource-panel-readiness.utils.ts`: added `canResolveResourceSelection(...)` and the exported
  `ResourceSelectionReadinessInput` type.
- `resource-selection.utils.ts`: `resolveResourceSelection()` converted to an options object and
  gained `areSourcesSettled` + `mayPersistCorrection`.
- `platform-scripture-editor.utils.ts`: `resolveResourceContentState()` gained a required
  `isUsjSettled`, now used in both directions.
- `resource-text-panel.component.tsx`: `ResourceTextPanelProps` gained required `onRetryChapter`.
- `resource-text-unavailable.component.tsx`: gained optional `retryLabel` / `onRetry`; now renders
  `RetryableErrorView`.
- `resource-message-view.component.tsx`: **extracted** `ResourceMessageFrame` (exported); the
  `action` slot added earlier on this branch was removed again with its only caller.
- `panel-state-views.component.tsx`: `LoadingView` gained an optional `className`.
- `resource-text-panel.web-view.tsx`: `ChapterUSJ` default changed from `EMPTY_USJ` to `undefined`.
- New localization key `%webView_resourcePanel_loadingResources%` (`en` + `es`).

None. No analysis pass raised a Critical finding.

- [x] **The retry affordance produced no visible feedback in its main case.** `useData` resets
      `isLoading` on a resubscription but never `data`, so a failed read's `PlatformError` stayed in
      hand while the retry ran — the panel re-rendered the identical message for the whole round
      trip, and a second identical failure never changed the screen at all. Found independently by
      the `api` and `ux` passes and verified directly against
      `create-use-data-hook.util.ts:415-417`. _(fixed during review: `resolveResourceContentState`
      now withholds a failure whose read is in flight; corrected the `isUsjSettled` TSDoc, which
      explicitly claimed the flag was never used to ENTER `loading`; three tests. Also fixes the
      separate stale-failure flash on ordinary navigation.)_
- [x] **The `useLayoutEffect` editor feed did not answer the hidden-view case**, contrary to
      `.claude/rules/cross-view-sync-hidden-views.md`. The panel already had `useViewVisibility`
      wired for its scroll sync, but the feed pushed a whole chapter into Lexical before a paint
      that does not happen in a `display: none` dock tab. _(fixed during review: gated on
      `isViewVisible`, which is a dependency so activation re-feeds before paint. Deliberately NOT
      `useRunWhenVisible` — its catch-up is a passive effect and would restore the placeholder frame
      at every activation; reasoning recorded at the site. Three tests, falsifiability verified.)_
- [x] **The retry control diverged from the platform's shared retry presentation.** It rendered a
      small outline button inside a plain message view, while two other retries *in the same panel*
      (and the picker, Get Resources, the Text Collection grid) use `RetryableErrorView` with an
      `AlertTriangle` and a default-emphasis button — the same label appearing as two different
      controls. _(fixed during review: routed through `RetryableErrorView`; extracted
      `ResourceMessageFrame` so focus repair and re-announcement are still shared; `role="status"`
      rather than the assertive default because the frame moves focus here, which is what actually
      announces it.)_
- [x] **`makeProps()` was not updated for the new required `onRetryChapter`**, so the component test
      file did not typecheck — invisible to CI because the workspace `tsconfig.json` excludes
      `**/*.test.tsx`. It also meant the failure test rendered no retry button at all, so that path
      was uncovered. This gap was introduced by the merge resolution, not the original branch.
      _(fixed during review: prop wired in, retry presence and click both asserted.)_
- [x] **`resource-text-unavailable.component.test.tsx` had no coverage of the new retry props.**
      _(fixed during review: four tests — the control calls `onRetry`, is omitted when terminal, is
      omitted when a handler arrives without a label, and reports politely rather than assertively.)_
- [x] **`ContentLoadingView`'s `announcementKey` remount had no test**, though every sibling message
      state has exactly that test. _(fixed during review: asserts live-region node identity changes
      across a chapter change; verified falsifiable by removing the `key`.)_
- [ ] ~~`ResourceMessageView`'s `action` slot placement had no test.~~ _(moot — the slot was removed
      when the failure state moved to `RetryableErrorView`.)_

### Minor — Consider

- [x] **A delivered `undefined` was reported as a failure with a retry**, but the extender PDP
      returns `undefined` for a falsy USX deterministically, so the retry could never change the
      outcome. _(fixed during review: routed to `ResourceBlankChapter`. Keyed on `usjPossiblyError`,
      not `usjFromPdp` — the latter is also `undefined` for a genuine error, so keying on it would
      have hidden real failures behind "empty chapter". Decided in the panel, not in the shared
      `isBlankChapterOnScreen`, because only this panel seeds with `undefined`.)_
- [x] **`%webView_resourcePanel_loading%` was serving as both a status sentence and the selector
      button's label**, constraining translators, and its ellipsis reads on a button as "opens
      something you can back out of". _(fixed during review: new key
      `%webView_resourcePanel_loadingResources%` — `Loading resources` / `Cargando recursos`,
      following the file's existing vocabulary and passing the `es` sentence-case guard. Two tests,
      including one that fails if the status string is reused.)_
- [x] **The `contentState === 'failed'` branch comment still called the state "terminal"** while the
      code two lines below offered a retry. _(fixed during review.)_
- [x] **The retry rationale was glued to the front-matter comment with no blank line**, and the
      "`useData` keys by reference" explanation was written out twice ~15 lines apart. _(fixed
      during review: separated, and the second statement now back-references the first.)_
- [x] **`mayPersistCorrection`'s first-ever-open asymmetry was undocumented** (no stored pick to
      protect, yet the fallback still is not persisted). _(fixed during review: TSDoc explains the
      cost is one unpersisted default and why a second condition is not worth adding.)_
- [x] **`LoadingView`'s `className` defaults to `tw:h-screen`**, so it is a height override wearing a
      general-purpose name — a caller passing `className` for padding silently loses full-panel
      height. _(fixed after review: the deferral rested on `tailwind-merge`'s handling of the `tw:`
      prefix being unverified. It was then verified — `cn` configures
      `extendTailwindMerge({ prefix: 'tw' })` and merges `tw:` classes correctly. That verification
      changed the fix: folding `tw:h-screen` into the base class string would REGRESS
      `ContentLoadingView`, which passes `tw:flex-1` precisely to drop the full-panel height —
      `h-screen` and `flex-1` are not conflicting utilities, so a merge keeps both. Renamed to
      `sizingClassName` instead, documenting that it replaces rather than merges. Behaviour
      unchanged at all four call sites.)_
- [ ] **`ContentLoadingView` is file-local** while `model-text-panel.component.tsx:642` still renders
      the exact bare unlabelled spinner this PR replaces. **Deferred to
      [PT-4622](https://paratextstudio.atlassian.net/browse/PT-4622).** The original rationale
      ("a consumer that does not exist yet") no longer holds — the consumer does exist — but
      adopting it means changing the sibling Model Text panel's behaviour, which is outside
      PT-4350's scope.
- [ ] **The "Enter some Scripture…" rationale is repeated near-verbatim in four places.** **Deferred
      to [PT-4623](https://paratextstudio.atlassian.net/browse/PT-4623).** Comment duplication only,
      no behavioural defect; churning it here adds diff noise to an already-large PR.

#### Shared Regions Modified

#### Extension Config Changes

None. No `package.json`, `tsconfig.json`, `webpack.config.ts`, or `.eslintrc*` was touched. The only
non-extension file changed is `.context/standards/Architecture-Decisions.md`.

## Positive Observations

- The display/persist split is well factored: `canResolveResourceSelection` sits beside
  `canPublishResourcePanelProjectIds`, `resolveResourceSelection` was converted to an options object
  so four transposable arguments cannot be swapped silently, and both settlement flags are required
  rather than defaulted — with the reason (`false` is a safe-looking but wrong default) written down.
- `resolveResourceSelection`'s new guard is placed *after* the `selectedRow` branch, so an unsettled
  state can never discard a selection that was actually found — the subtle failure mode this
  refactor could easily have introduced.
- The PT-4424 merge re-application was verified faithful and complete hunk-by-hunk by the `api` pass;
  the only deltas were correct adaptations to main's new container/presentational boundary.
- The retry is tested at the level that matters — a *new selector object* for the same reference,
  which is what actually forces a resubscribe — rather than "button clicked".
- The overlapping-pick generation guard has a real two-deferred test asserting the superseded pick
  does not clear `isSelecting`.
- Localization is clean: no hardcoded user-visible text, every key declared in
  `resource-text-panel.const.ts` and present in both `en` and `es`.
- The one `eslint-disable` carries a justification naming the alternatives and why each is worse.
- `TODO(PT-4509)` and `TODO(PT-4517)` are forward-facing deferrals naming open tickets.

## Interview Notes

The author directed the review rather than explaining prior decisions: for each finding presented
they chose a disposition ("Fix", "Gate it", "Sure option 1", "Route to `ResourceBlankChapter`",
"follow code patterns to pick the strings"), and the work was carried out in-session. No finding was
dismissed, and no explanation of existing code was offered or requested — so this summary records
**no author rationale** for the pre-existing design, and the reviewer should not read the absence as
endorsement.

Two decisions were made by the reviewer rather than the author, and both are worth confirming:

- **`role="status"` instead of `RetryableErrorView`'s assertive `alert` default.**
  `ResourceMessageFrame` moves focus to the region, and that focus move is what announces the
  message; an alert on top would interrupt to repeat it. The install and catalog failures keep
  `alert` because nothing moves focus to those. This is a deliberate divergence from the sibling
  states and should be sanity-checked by someone who tests with a screen reader.
- **Wording of the new localized strings** (`Loading resources` / `Cargando recursos`), chosen by
  following the file's existing vocabulary and conventions. Nobody with Spanish fluency has reviewed
  them.

**Carried forward from the previous review round** (author-stated, still current):

- **PT-4350 is deliberately split across two stacked PRs.** The author asked that this be recorded
  so a reviewer does not read AC2 as a gap here.
- **The catalog-failure asymmetry was intended; its persistence consequence was not.** The author had
  reasoned correctly that a failed catalog must not strand the panel with no selection, but had not
  considered that the same flag also gated the write-back, and asked for the fix rather than
  defending the original behaviour.

**#2793 now needs a substantial rebase, more than previously recorded.** The earlier round noted a
conflict because `resolveResourceContentState` gained `isUsjSettled` here while #2793 adds
`isAnswerCurrent` to the same function. Two things have made that worse since:

1. `main` landed **PT-4424 (#2748)**, which split `resource-text-panel.web-view.tsx` into a container
   plus a new `resource-text-panel.component.tsx`. #2793 is still based on the pre-split tip
   (`dcf236d4c67`) and edits `resource-text-panel.web-view.tsx` in regions that no longer live there.
2. This round moved `contentState` and `isBlankChapter` into the new component and changed
   `resolveResourceContentState` again — it now uses `isUsjSettled` in both directions, and a
   delivered `undefined` routes to the blank-chapter branch rather than the failure branch.

#2793's `isAnswerCurrent` check and this PR's `isUsjSettled` are answering closely related questions
("is the value in hand the current read's?"), and `adr-stale-value-survives-resubscription` added
here generalises exactly that. Whoever rebases #2793 should check whether the two mechanisms should
be merged rather than layered.

Also worth flagging: `onRetryChapter` was added as a **required** prop during the merge resolution
and the test `makeProps` was not updated, which the workspace typecheck could not catch. That is a
standing hole — `extensions/src/platform-scripture-editor/tsconfig.json` excludes `**/*.test.tsx`,
so no extension test file is typechecked by `npm run typecheck`. Several pre-existing type errors in
`resource-text-panel.web-view.test.tsx` are sitting there unnoticed for the same reason.

## In-Review Quality Check

Run after every change, not only at the end. Final state:

- `npm run typecheck` — exit 0 (all four lanes: core, erb, e2e, workspaces)
- `npm run lint` — exit 0
- `platform-scripture-editor` suite — **91 files, 1533 tests, all passing** (up from 1515 at merge)
- Changed files formatted with `prettier`

Two fixes were verified **falsifiable** by temporarily reverting the implementation and confirming
the new test failed, then restoring: the hidden-view gate, and the `ContentLoadingView` remount key.

Unfixable failures: none.

Not addressed (out of scope, flagged for follow-up): the pre-existing type errors in
`resource-text-panel.web-view.test.tsx`, and the same jsdom visibility blind spot in
`platform-scripture`'s `find.component.test.tsx`, which copies the stub pattern this review corrected
here.

## Suggested Review Focus

- [ ] **`role="status"` on the chapter-failure state** — a deliberate divergence from the sibling
      retryable-error surfaces, argued from the focus move. Confirm with an actual screen reader.
- [ ] **The new Spanish string** `Cargando recursos` — chosen by pattern-matching, not by a fluent
      speaker.
- [ ] **A delivered `undefined` is now an empty chapter rather than a failure.** Confirm the extender
      PDP's falsy-USX path is really the only way to reach it, and that "empty chapter" is the right
      claim for what the user is looking at.
- [ ] **The one-render window** where a stale failure is still shown before the spinner replaces it
      (`isLoading` is re-armed from an effect). Documented in the `isUsjSettled` TSDoc and in
      `adr-stale-value-survives-resubscription`; confirm it is acceptable in practice.
- [ ] **`adr-stale-value-survives-resubscription`** proposes a rule for every `useData` consumer in
      the app, and names "reset `data` on resubscribe, in the hook" as the real fix that was
      deliberately not attempted here. That is a platform-level call worth a decision.
- [ ] **Extension test files are not typechecked by `npm run typecheck`.** This review found one real
      error that way and left several pre-existing ones. Worth a ticket.
- [ ] **AC2 is deliberately out of scope**, handled by stacked PR #2793 — which now needs a
      substantial rebase (see Interview Notes), and whose `isAnswerCurrent` may want merging with
      this PR's `isUsjSettled` rather than layering on it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2791)

[PT-4622]: https://paratextstudio.atlassian.net/browse/PT-4622?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.context/standards/Architecture-Decisions.md
+++ b/.context/standards/Architecture-Decisions.md
@@ -1672,6 +1672,49 @@ step, no automation. Just a record.
   the merged conditional in `model-text-panel.component.tsx` — proved to be the symptom site rather
   than the defect.
 
+## adr-panel-readiness-splits-display-from-persistence: A retryable failure settles what a panel may SHOW, never what it may WRITE BACK
+
+- **Date:** 2026-09-10
+- **Status:** Accepted
+- **Context:** Refines `adr-panel-readiness-from-sources`, which established that readiness is
+  derived from whether sources arrived. That framing treats readiness as one question, and it is
+  two. The resource panels filter their rows against the DBL catalog, and a DBL reference with no
+  catalog row resolves to nothing at all (`downloaded-resources.utils.ts`), so a configured
+  resource is simply ABSENT from the rows until its catalog entry lands. Auto-correct reads an
+  absent selection as "gone" and persists the first usable row over it — and the stored id does
+  not come back, because it has already been written away (PT-4470). Gating auto-correct on
+  arrival fixes the loading window but creates a worse one on FAILURE: `isCatalogReady` stays
+  `false` for the rest of the session after a failed fetch, while locally-downloaded rows keep
+  arriving and stay selectable, so a panel that waits for arrival has no selection, hence no
+  resource project id, hence a permanent spinner with no reachable retry (it is `configured`, so
+  it never renders `PanelReadinessView`). Both framings are wrong because both answer one question.
+- **Decision:** Split the question at the point of use. **Display** may proceed once the sources
+  have settled *for now* — a definitively failed catalog counts, because the rows in hand are all
+  the rows there are, and showing one beats showing nothing
+  (`canResolveResourceSelection`). **Persistence** requires that every source actually DELIVERED
+  (`canPublishResourcePanelProjectIds`), because a failure that a retry can undo is not evidence
+  that a selection is gone. `resolveResourceSelection` takes both as separate inputs
+  (`areSourcesSettled`, `mayPersistCorrection`) rather than one settled flag, so the asymmetry is
+  visible in its signature instead of living in a caller's head. The general rule: a state derived
+  from a RETRYABLE failure may drive what the user sees, never a write that outlives the failure.
+- **Alternatives:** **One `areSourcesSettled` flag treating a failed catalog as settled** —
+  rejected: it is what reintroduced PT-4470's defect in the failure branch, silently, in the one
+  window where the panel is showing its catalog-error view and the reader cannot see the pick being
+  replaced. **Wait for delivery for both questions** — rejected: the permanent-spinner outcome
+  above; failing closed is safe where it means "declare nothing" and unsafe where it means "show
+  nothing". **Make the write reversible instead (snapshot and restore the pick)** — rejected as
+  more machinery than the problem needs, and it still shows the wrong resource meanwhile.
+- **Consequences:** A panel whose catalog failed displays a fallback row while leaving the stored
+  selection untouched, so a successful retry restores the user's text rather than revealing that it
+  was overwritten. The two predicates coincide today except on catalog failure, and reusing
+  `canPublishResourcePanelProjectIds` for the persistence question keeps them from drifting into
+  two near-copies — but it does couple them, so a future change to publishing must check this
+  caller. Any third async source must answer both questions, not one. The same split applies beyond
+  selection: any panel state that both renders and persists should name which of the two a given
+  readiness signal licenses.
+- **Source:** PT-4350 (NN 5B Opening a resource always correctly loads it) and PT-4470, the
+  catalog-window regression, found in review of the fix for the loading window.
+
 ## adr-papi-websocket-hostname-bind: The PAPI websocket binds by hostname, and readiness is awaited rather than assumed
 
 - **Date:** 2026-08-29
--- a/extensions/src/platform-scripture-editor/src/panel-state-views.component.tsx
+++ b/extensions/src/platform-scripture-editor/src/panel-state-views.component.tsx
@@ -1,4 +1,4 @@
-import { Button, RetryableErrorView, Spinner } from 'platform-bible-react';
+import { Button, cn, RetryableErrorView, Spinner } from 'platform-bible-react';
 import { ReactNode, useId, useState } from 'react';
 
 /**
@@ -13,11 +13,24 @@ import { ReactNode, useId, useState } from 'react';
  *
  * @param label Already-localized status text shown next to the spinner (e.g. "Installing
  *   resource…"). Doubles as the state's accessible name.
+ * @param className Sizing for the block, replacing the default full-panel height. A panel waiting
+ *   inside its own content area — below a selector header that stays mounted — passes the height of
+ *   that area instead, so the spinner centres in the region it describes rather than pushing past
+ *   the bottom of the panel.
  */
-export function LoadingView({ label }: { label: ReactNode }) {
+export function LoadingView({
+  label,
+  className = 'tw:h-screen',
+}: {
+  label: ReactNode;
+  className?: string;
+}) {
   return (
     <div
-      className="tw:flex tw:h-screen tw:items-center tw:justify-center tw:gap-2 tw:p-8 tw:text-center"
+      className={cn(
+        'tw:flex tw:items-center tw:justify-center tw:gap-2 tw:p-8 tw:text-center',
+        className,
+      )}
       role="status"
     >
       <Spinner aria-hidden />
--- a/extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.test.ts
+++ b/extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.test.ts
@@ -3104,22 +3104,37 @@ describe('resolveResourceContentState', () => {
         resourceProjectId: undefined,
         usjPossiblyError: undefined,
         currentBookNum: GENESIS,
+        isUsjSettled: false,
       }),
     ).toBe('loading');
   });
 
-  it('returns "loading" for a caller whose hook is not seeded with a default', () => {
-    // The Bible texts panel passes `EMPTY_USJ` as its hook's default, so its value is an object from
-    // the first render and this pair does not occur there. Kept because the state is part of this
-    // function's contract for any caller reading a hook without a seeded default, and because the
-    // alternative — falling to `'ready'` — would mount an editor with nothing to put in it.
+  it('returns "loading" while nothing has arrived and the subscription is still working', () => {
+    // The state a caller reaches by seeding its hook with `undefined` (the Bible texts panel): no
+    // chapter in hand and none refused, so the honest answer is a spinner. Falling to `'ready'`
+    // here would mount an editor with nothing to put in it.
     expect(
       resolveResourceContentState({
         resourceProjectId: PROJECT_ID,
         usjPossiblyError: undefined,
         currentBookNum: GENESIS,
+        isUsjSettled: false,
       }),
     ).toBe('loading');
+  });
+
+  it('returns "failed" when the subscription settles having delivered nothing', () => {
+    // `ChapterUSJ`'s `getData` is `Usj | undefined` and the extender PDP returns `undefined` for a
+    // falsy USX, so this is a real delivery, not an absence of one. Reported rather than spun on:
+    // an unending spinner with no message is the "stuck loading" symptom itself.
+    expect(
+      resolveResourceContentState({
+        resourceProjectId: PROJECT_ID,
+        usjPossiblyError: undefined,
+        currentBookNum: GENESIS,
+        isUsjSettled: true,
+      }),
+    ).toBe('failed');
   });
 
   it('returns "ready" once chapter data has arrived', () => {
@@ -3128,6 +3143,7 @@ describe('resolveResourceContentState', () => {
         resourceProjectId: PROJECT_ID,
         usjPossiblyError: { type: 'USJ', version: '3.1', content: [] },
         currentBookNum: GENESIS,
+        isUsjSettled: true,
       }),
     ).toBe('ready');
   });
@@ -3138,6 +3154,7 @@ describe('resolveResourceContentState', () => {
         resourceProjectId: PROJECT_ID,
         usjPossiblyError: missingBook(GENESIS),
         currentBookNum: GENESIS,
+        isUsjSettled: true,
       }),
     ).toBe('bookNotAvailable');
   });
@@ -3157,6 +3174,7 @@ describe('resolveResourceContentState', () => {
         resourceProjectId: PROJECT_ID,
         usjPossiblyError: missingBook(MATTHEW),
         currentBookNum: GENESIS,
+        isUsjSettled: true,
       }),
     ).toBe('loading');
   });
@@ -3169,6 +3187,7 @@ describe('resolveResourceContentState', () => {
         resourceProjectId: PROJECT_ID,
         usjPossiblyError: missingBook(GENESIS, 'someOtherProject'),
         currentBookNum: GENESIS,
+        isUsjSettled: true,
       }),
     ).toBe('loading');
   });
@@ -3187,6 +3206,7 @@ describe('resolveResourceContentState', () => {
           content: [`Book number ${GENESIS} not found in project ${PROJECT_ID}.`],
         },
         currentBookNum: GENESIS,
+        isUsjSettled: true,
       }),
     ).toBe('ready');
   });
@@ -3201,6 +3221,7 @@ describe('resolveResourceContentState', () => {
         resourceProjectId: 'abc123',
         usjPossiblyError: missingBook(GENESIS, 'ABC123'),
         currentBookNum: GENESIS,
+        isUsjSettled: true,
       }),
     ).toBe('bookNotAvailable');
   });
@@ -3214,6 +3235,7 @@ describe('resolveResourceContentState', () => {
         resourceProjectId: PROJECT_ID,
         usjPossiblyError: missingBook(0),
         currentBookNum: 0,
+        isUsjSettled: true,
       }),
     ).toBe('loading');
   });
@@ -3230,6 +3252,7 @@ describe('resolveResourceContentState', () => {
         resourceProjectId: PROJECT_ID,
         usjPossiblyError: newPlatformError(new Error('Project abc123 is not available')),
         currentBookNum: GENESIS,
+        isUsjSettled: true,
       }),
     ).toBe('failed');
   });
--- a/extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts
+++ b/extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts
@@ -1668,24 +1668,39 @@ export type ResourceContentState = 'loading' | 'bookNotAvailable' | 'failed' | '
  *
  * @param options.resourceProjectId The project id the panel is reading from, or `undefined` if a
  *   resource has not resolved to one yet.
- * @param options.usjPossiblyError The chapter USJ, a `PlatformError`, or `undefined` if none yet. A
- *   caller that seeds its data hook with a default (the Bible texts panel passes `EMPTY_USJ`) never
- *   passes `undefined`; for those, `resourceProjectId` not having resolved yet is what produces
- *   `'loading'`.
+ * @param options.usjPossiblyError The chapter USJ, a `PlatformError`, or `undefined` if none yet.
+ *   Whether `undefined` can reach here at all is the caller's choice of seed for its data hook. A
+ *   caller seeding `undefined` (the Bible texts panel) says "nothing has arrived" through this
+ *   value directly. One seeding a blank USJ (`scripture-text-grid/resource-cell.component.tsx`)
+ *   never can, because a blank USJ is neither `undefined` nor falsy, and must distinguish
+ *   not-yet-arrived some other way — there, `resourceProjectId` not having resolved yet, plus the
+ *   `isLoading` its own `deriveCellState` takes.
  * @param options.currentBookNum The book number the panel is currently displaying. A value of 0 or
  *   less means the panel has no book it can name, so no claim is made about one.
+ * @param options.isUsjSettled Whether the chapter subscription has stopped loading. Distinguishes a
+ *   `undefined` that has not arrived yet from one that IS the answer — the data type's `getData` is
+ *   `Usj | undefined` and the extender PDP returns `undefined` for a falsy USX, so a delivered
+ *   `undefined` is reachable and would otherwise spin forever. Used ONLY to escalate out of
+ *   `'loading'`, never to enter it: the flag is re-armed from an effect, so for one render after a
+ *   selector change it still reads settled — harmless here, because the value in hand at that
+ *   moment is the previous reference's USJ rather than `undefined`.
  * @returns Which of the four content states to render.
  */
 export function resolveResourceContentState({
   resourceProjectId,
   usjPossiblyError,
   currentBookNum,
+  isUsjSettled,
 }: {
   resourceProjectId: string | undefined;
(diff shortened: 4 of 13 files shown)


tombogle:
## Devin finding: superseded resource picks can win (pre-existing race, not introduced here)

**Original Devin comment** (`extensions/src/platform-scripture-editor/src/resource-text-panel.web-view.tsx:R458-459`):

> **Superseded resource picks can win**
>
> When picks overlap, a superseded `selectTextConnection` can still write and update
> `pendingResourceId`. Completion order can replace the latest pick and discard its list update.
>
> Each pick independently reads the current user list, installs, writes a replacement list, and
> invokes `onSelect`. The generation is checked only in `finally`, after all those side effects have
> completed. A slower superseded request can therefore write an older snapshot after the newer
> request and then set itself as pending.
>
> **Example:** Alice picks A and immediately picks B. B reads the original list, writes
> `[B, ...original]`, and sets pending B. A then finishes from its own original snapshot, writes
> `[A, ...original]`, and sets pending A. The panel displays A and B disappears, although B was the
> latest choice.
>
> **Recommended fix:** Serialize picks or add generation checks before every superseded request can
> persist or publish selection. Prefer a mutation that merges against the latest stored list at
> write time, then allow only the current generation to update `pendingResourceId`.

**Verified — the mechanism is real:**

- `selectTextConnection` (`extensions/src/platform-scripture-editor/src/select-dbl-resource.ts:30-65`)
  is a plain read-then-write: `await getUserTextConnections()` for a snapshot, build a new list from
  it, `await setUserTextConnections(...)`, then invoke `onSelect(newRef)`. None of it is
  generation-aware.
- This PR's `selectGenerationRef` (`resource-text-panel.web-view.tsx:454-459`) only gates
  `setIsSelecting(false)` in the `finally` block (line 488) — per its own comment, its purpose is
  narrowly to stop a superseded pick from re-arming `useDblResourceAutoInstall`. It does **not**
  guard the `onSelect` callback that calls `setPendingResourceId(...)` (line 483), and it does
  nothing for the list write inside `selectTextConnection` itself.
- So the Alice/A/B trace holds as described: a slower superseded pick can complete after a newer
  one, overwrite the list from a stale snapshot (silently dropping the newer pick's entry), and
  stomp `pendingResourceId` back to the older choice.

**Scope relative to this PR:**

- `select-dbl-resource.ts` is **not touched by this PR** (last modified in #2630) — the race in the
  write itself is pre-existing, not introduced here.
- No commit message, ADR entry (`Architecture-Decisions.md`), or PR description mentions this race.
- The only overlapping-pick test (`resource-text-panel.web-view.test.tsx:383-421`) mocks
  `selectTextConnection` entirely and asserts only the `isSelecting` label — it never exercises the
  write race, so it would not catch this.
- This PR's AC3 row ("overlapping resource picks must not disarm each other → Yes") reads as broader
  coverage than what actually landed (flag-only, not data-integrity). Worth correcting the wording or
  scoping the claim explicitly to `isSelecting`/auto-install.

**Suggested disposition:** pre-existing bug, out of scope for this PR to fix inline — but it should
not go unrecorded. Recommend filing a follow-up ticket against `select-dbl-resource.ts` /
`handleResourceSelect` covering:

1. Serialize resource picks (e.g., queue/await the in-flight pick before starting the next), **or**
2. Merge against the latest stored list at write time instead of a stale snapshot, and gate both the
   list write and `setPendingResourceId` on `generation === selectGenerationRef.current`, not just
   the `isSelecting` flag.
3. Add a test that exercises two overlapping *real* (unmocked, or mocked at the list-read/write
   layer rather than at `selectTextConnection`) picks and asserts the later pick's entry survives in
   the written list and `pendingResourceId`.

tombogle (changes requested):
> @tombogle reviewed 18 files and all commit messages, and made 8 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2791)* status: all files reviewed, 8 unresolved discussions (waiting on jolierabideau).

___
*[`extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts` line 1770 at r2](https://reviewable.io/reviews/paranext/paranext-core/2791#-P1WP3-j2uh86hW6IVBn:-P1WP3-j2uh86hW6IVBo:bfr3c0y) ([raw file](https://github.com/paranext/paranext-core/blob/bc4a19b1fdb842532c743cb35a250e3cf4230d68/extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts#L1770)):*
> ```XML
>  * @param options.usjPossiblyError The chapter USJ, a `PlatformError`, or `undefined` if none yet.
>  *   Whether `undefined` can reach here at all is the caller's choice of seed for its data hook. A
>  *   caller seeding `undefined` (the Bible texts panel) says "nothing has arrived" through this
> ```

Should probably say "e.g., the Bible texts panel", so as not to imply this is the one and only caller that might ever seed it this way.
___
*[`extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts` line 1781 at r2](https://reviewable.io/reviews/paranext/paranext-core/2791#-P1WPidU1dbMHzLAPsKi:-P1WPidU1dbMHzLAPsKj:b-a1jae6) ([raw file](https://github.com/paranext/paranext-core/blob/bc4a19b1fdb842532c743cb35a250e3cf4230d68/extensions/src/platform-scripture-editor/src/platform-scripture-editor.utils.ts#L1781)):*
> ```XML
>  *   holds the last delivered value across a resubscription — it resets `isLoading`, never `data`.
>  *
>  *   Escalating out of `'loading'`: distinguishes a `undefined` that has not arrived yet from one that
> ```

typo: should be "an `undefined`"
___
*[`extensions/src/platform-scripture-editor/src/resource-message-view.component.tsx` line 99 at r2](https://reviewable.io/reviews/paranext/paranext-core/2791#-P1_Lfd0C9kHLNp18LVN:-P1_Lfd0C9kHLNp18LVO:bj1ml6r) ([raw file](https://github.com/paranext/paranext-core/blob/bc4a19b1fdb842532c743cb35a250e3cf4230d68/extensions/src/platform-scripture-editor/src/resource-message-view.component.tsx#L99)):*
> ```XML
>  * reading like every other empty state in the app and supplies the `role="status"` live region.
>  *
>  * For the states whose whole content is that sentence. A state that also offers a way out renders
> ```

"For the states whose whole content is that sentence." is a sentence fragment. I had to re-read a couple of times to figure out how it related to the second sentence. If you can think of a reason to keep it, I suggest rewording it as:
This shared body is intended for states whose whole content is that sentence. 
However, I think it doesn't add anything substantive. The preceding paragraph already establishes that `EmptyState` is for the message-only case, so probably best to remove it.
___
*[`extensions/src/platform-scripture-editor/src/resource-message-view.component.tsx` line 103 at r2](https://reviewable.io/reviews/paranext/paranext-core/2791#-P1_Prnb5fJ56TE8a4-V:-P1_Prnb5fJ56TE8a4-W:bvqcroh) ([raw file](https://github.com/paranext/paranext-core/blob/bc4a19b1fdb842532c743cb35a250e3cf4230d68/extensions/src/platform-scripture-editor/src/resource-message-view.component.tsx#L103)):*
> ```XML
>  * one every other surface shows for the same condition.
>  *
>  * `EmptyState` supplies the `role="status"` live region; the focus repair and re-announcement are
> ```

The first part of this sentence is entirely redundant (stated verbatim at the end of the second paragraph above
___
*[`extensions/src/platform-scripture-editor/src/resource-message-view.component.tsx` line 104 at r2](https://reviewable.io/reviews/paranext/paranext-core/2791#-P1_Ok-kCPvLNoCvnvee:-P1_Ok-kCPvLNoCvnvef:b-p6n4cl) ([raw file](https://github.com/paranext/paranext-core/blob/bc4a19b1fdb842532c743cb35a250e3cf4230d68/extensions/src/platform-scripture-editor/src/resource-message-view.component.tsx#L104)):*
> ```XML
>  *
>  * `EmptyState` supplies the `role="status"` live region; the focus repair and re-announcement are
>  * {@link ResourceMessageFrame}'s, which also explains why the message is not reliably announced by
> ```

I think the second half of this sentence means:

 * TODO(PT-4416): Move the focus-based announcement from {@link ResourceMessageFrame} into `EmptyState`.

If I'm right, maybe say it that way. Using `TODO(PT-4416)` makes it easily searchable and unambiguous that we are documenting a defect, not a desired behavior or an already-corrected problem.
___
*[`extensions/src/platform-scripture-editor/src/resource-text-panel.component.tsx` line 329 at r2](https://reviewable.io/reviews/paranext/paranext-core/2791#-P1aYZ242ntWHKexcvVP:-P1aYZ242ntWHKexcvVQ:bkxka72) ([raw file](https://github.com/paranext/paranext-core/blob/bc4a19b1fdb842532c743cb35a250e3cf4230d68/extensions/src/platform-scripture-editor/src/resource-text-panel.component.tsx#L329)):*
> ```XML
>   // A settled read that delivered nothing counts too, and is decided HERE rather than inside
>   // `isBlankChapterOnScreen` because only this panel seeds its subscription with `undefined` — the
>   // model text panel seeds a blank USJ, so `undefined` does not carry this meaning there and the
> ```

Comments like "only this panel seeds its subscription with `undefined`" and "the model text panel seeds a blank USJ" are not a facts about the code here. If someone adds another panel that seeds with `undefined` or changes the model panel's seeding behavior, there's no mechanism (compiler, test, or even proximity) that would prompt them to revisit this comment. It just silently goes stale.
___
*[`extensions/src/platform-scripture-editor/src/resource-text-panel.const.ts` line 37 at r2](https://reviewable.io/reviews/paranext/paranext-core/2791#-P1a_3rjC42kseTWvWr2:-P1a_3rl2rFvxv7SN7n3:b-h4sic) ([raw file](https://github.com/paranext/paranext-core/blob/bc4a19b1fdb842532c743cb35a250e3cf4230d68/extensions/src/platform-scripture-editor/src/resource-text-panel.const.ts#L37)):*
> ```XML
>   // The resource selector's own waiting label. Deliberately NOT `..._loading%`: that one is a
>   // status region's sentence, and reusing it here would make one string answer to both a status and
>   // a control, which translators cannot phrase differently. It also carries an ellipsis, which on a
> ```

NIT: *in English* it carries an ellipsis. May not be true in all localizations. (But maybe still better to say that than to try to talk about the meaning in the abstract.)
___
*[`extensions/src/platform-scripture-editor/src/resource-text-unavailable.component.test.tsx` line 59 at r2](https://reviewable.io/reviews/paranext/paranext-core/2791#-P1aaWPmDDCm1E0RnR9Q:-P1aaWPmDDCm1E0RnR9R:b-xgobmo) ([raw file](https://github.com/paranext/paranext-core/blob/bc4a19b1fdb842532c743cb35a250e3cf4230d68/extensions/src/platform-scripture-editor/src/resource-text-unavailable.component.test.tsx#L59)):*
> ```XML
> 
>   it('names the failure without a control when nothing can re-drive the read', () => {
>     // The Model text panel has no re-subscribe path for this failure. An inert button in a state
> ```

As before, best not to concretely comment behavior coded elsewhere.

PR #2792, opened 2026-09-09
PT-3299: Gate Find's current-book/chapter scopes on a searchable book

jolierabideau:
# Code Review Summary

**Branch**: pt-3299-remove-extra-books

**Base**: origin/main

**Date**: 2026-09-09

**Files changed**: 19 at review start (24 after in-review fixes and the `platform-bible-react` dist rebuild)

## Overview

PT-3299 was reopened: testing Pkg 238 in Power mode found that with the current reference in extra
material (XXB/GLO), Find still returned results for the **Current book** and **Current chapter**
scopes, labeled with the meaningless reference the original fix (#2708) exists to hide. #2708
excluded extra material from Find's *book lists*, but those two scopes resolve their scope from the
current scripture reference instead, and the navigation control offers every book the project has.
This branch closes that path with a predicate over the current reference, enforced at the query
gate rather than in the UI, plus two explanations (a results-area placeholder and disabled scope
options with tooltips) and a new additive `disabledScopeExplanations` prop on `ScopeSelector`.

The analysis found no critical issues. Four important findings and ten minor ones were raised; the
author elected to fix all four important findings and six of the ten minor ones during the review,
with the remaining four closed as a side effect of those fixes and one declined with a stated
reason. The central design decision — gating at `isFindQueryValid` rather than at the scope control,
because `scope` is persisted per web view while the reference moves independently — was probed in
the interview and the author's account matched the code, including the render-ordering detail that
prevents stale results from surviving the transition.

## API Changes

- `lib/platform-bible-react` (main entry, `ScopeSelector`): added optional prop
  `disabledScopeExplanations?: Partial<Record<ScopeWithRange, string>>` — additive, defaults to
  `undefined`, existing callers unaffected.
- `lib/platform-bible-react` (main entry): `ScopeWithRange` is now re-exported from `index.ts`
  alongside `Scope`. Previously reachable only via `platform-bible-react/experimental`, which a
  main-entry consumer of the new prop would otherwise have had to reach into. Added during review.
- `extensions/src/platform-scripture/src/find/find-book-lists.utils.ts`: added export
  `isExtraMaterialBookId(bookId: string): boolean` (extension-internal module).
- `extensions/src/platform-scripture/src/find/find.utils.ts`: `isFindQueryValid()`'s params object
  gained a **required** `currentBookId: string` field; all three call sites updated
  (`find.web-view.tsx`, `find.component.tsx`, `find.stories.tsx`). Added export
  `isScopeBlockedByExtraMaterial(scope, currentBookId)` during review.
- `lib/papi-dts/papi.d.ts`: no changes. No extension type declaration files changed.
- `lib/platform-bible-react/dist/` was rebuilt so the new prop, the new dropdown rendering, and the
  `ScopeWithRange` export actually ship.

None.

- [x] **`isFindQueryValid` was broader than its documented rule** — `find.utils.ts` rejected every
      scope other than `selectedBooks` in extra material, so `verse` and `selectedText` were also
      blocked despite the TSDoc naming only `book`/`chapter`. Unreachable today (Find offers only
      three scopes) but the helper is the shared source of truth for the web view and the Storybook
      harness. _(fixed during review: extracted `isScopeBlockedByExtraMaterial(scope, currentBookId)`
      naming `book`/`chapter` explicitly; `isFindQueryValid` and `find.component.tsx`'s
      `isBlockedByExtraMaterial` both call it, so the component's `scope !== 'selectedBooks'` copy
      can no longer diverge from the gate. New tests cover all five `Scope` members plus a case
      asserting `verse`/`selectedText` stay valid in extra material.)_
- [x] **The `ScopeSelector` remount mitigation was defeated by its only consumer** — the radio branch
      wrapped every row in `DisabledActionTooltip` only once `disabledScopeExplanations` was defined,
      justified by a comment about avoiding a remount that drops keyboard focus; but Find's memo
      returned `undefined` when nothing was blocked and an object when it was, so the prop flipped
      `undefined ↔ object` at exactly the transition the wrapping was meant to survive, and every row
      swapped `Fragment → DisabledActionTooltip` anyway. _(fixed during review: dropped the
      `undefined` fast path so every row wraps unconditionally and the element type is stable
      regardless of how a consumer supplies the prop; comment corrected to record why the
      one-level-up shortcut had the same defect. `Fragment` import removed.)_
- [x] **The dropdown variant's disabled explanation was unreachable by keyboard and screen reader** —
      it was delivered only via the native `title` attribute on a Radix `DropdownMenuItem` that is
      also `disabled`. Radix excludes disabled items from the menu's roving focus and a native
      `title` appears only on pointer hover, so the users who most need the explanation got nothing.
      Find uses the radio variant, so this affected other/future consumers.
      _(fixed during review: the explanation now renders as muted text inline under the option's
      label — the one form that needs neither hover nor focus and does not fight
      `DropdownMenuContent` for the pointer. The prop's TSDoc, which claimed "shown as a tooltip on
      both variants", now states what each variant actually renders. Two tests added: the text
      appears on the disabled item, and does not leak onto enabled ones.)_
- [x] **The results-area placeholder named the problem but not the next step** — contrary to the
      zero-state guideline in `guidelines/component-choices.mdx` and every sibling placeholder in the
      same results area. The one string also did double duty as the disabled-scope tooltip, where the
      fuller wording exceeds `guidelines/tooltips.mdx`'s "a hint, not a sentence".
      _(fixed during review: split into two keys. `%webView_find_extraMaterialNotSearchedScope%` is
      now the short tooltip ("Find doesn't search extra material."); new
      `%webView_find_extraMaterialNotSearchedScopeResults%` carries the actionable placeholder
      ("…Choose books to search, or go to a Scripture book."). Both key constants document why they
      are separate.)_

Author response: The author asked for all four to be fixed, choosing the recommended option on the
dropdown-accessibility finding after being offered the smaller TSDoc-only alternative. No important
finding was dismissed.

### Minor — Consider

- [x] **Two encodings of "extra material" in one file** — `EXTRA_MATERIAL_BOOK_NUMBERS` derived from
      `Canon.nonCanonicalIds` while `isExtraMaterialBookId` delegated to `Canon.isExtraMaterial`;
      agreement was test-enforced rather than structural. _(fixed during review: the number set now
      derives from `Canon.allBookIds.filter(isExtraMaterialBookId)`, so one predicate defines both.)_
- [x] **Scope tooltip was a full sentence** _(fixed during review: shortened, now that the split gives
      the placeholder its own string.)_
- [x] **The collapsed scope trigger still read as a normal, active scope** while blocked, showing
      e.g. "XXB 1" with nothing to indicate it cannot run; the results area carries the explanation
      but can be scrolled away. _(fixed during review: the summary renders muted + italic when
      blocked and points at the placeholder via `aria-describedby`.)_
- [x] **No fallback for a missing localized string** — `localizedStrings` is an open index signature,
      so an unrequested key reads as `undefined`, which would have produced
      `{ book: undefined, chapter: undefined }`, left both scopes **enabled** while the query gate
      still rejected them, and shown the placeholder with no affordance explaining it.
      _(fixed during review: both the tooltip and the placeholder fall back to the raw key, matching
      the library's own `localizeString`; a test asserts the scopes stay disabled when the string is
      missing.)_
- [x] **No Storybook story for the extra-material state** _(fixed during review: added
      `CurrentReferenceInExtraMaterial`; `HarnessConfig` gained an optional `verseRef`.)_
- [x] **`ScopeWithRange` exported only from `experimental.ts`** while `ScopeSelector` is on the main
      entry — pre-existing, but the new prop widened the surface needing it. _(fixed during review:
      re-exported from `index.ts`.)_
- [x] **The rule was expressed three times** (`isFindQueryValid`, `isBlockedByExtraMaterial`, and the
      `disabledScopeExplanations` memo), two of them using the over-broad negative shape.
      _(closed by the first important fix: the two scope-dependent sites now share one predicate; the
      memo is scope-independent by design and unchanged.)_
- [x] **The prop's TSDoc did not note that presentation differs by variant** _(closed by the
      dropdown-accessibility fix.)_
- [x] **The dropdown test did not assert the explanation was surfaced** — it checked `aria-disabled`
      and that `onScopeChange` was not called, but nothing covered the only channel carrying the
      reason. _(closed by the dropdown-accessibility fix, which added that assertion.)_
- [ ] ~~Duplicated `getScopeOption` test helper across `find.component.test.tsx` and
      `scope-selector.component.test.tsx`~~ _(Declined by the reviewer with the author's agreement:
      the two helpers live in different npm packages, `platform-bible-react` has no test-utility
      module or export today, and the two are not identical — one takes a `RegExp`, the other a
      string key. Sharing would mean adding a published export whose only consumer is a test in
      another package. Duplication judged the cheaper option.)_

Author response: The author asked for all minor findings to be fixed. Six were implemented and three
more were closed as a side effect of the important fixes; the tenth was declined with the reasoning
above and surfaced explicitly rather than skipped.

#### Shared Regions Modified

None. No `#region shared with` markers appear in any changed source file.

#### Extension Config Changes

None. The only non-source file changed under `extensions/` is
`extensions/src/platform-scripture/contributions/localizedStrings.json`, which is extension-specific
content, not shared build/config infrastructure. No `package.json`, `tsconfig*.json`, `.eslintrc*`,
`webpack.config.*`, `.erb/configs/`, or `.github/workflows/` files were touched.

## Positive Observations

- The gate is enforced at the right layer. `isFindQueryValid` is shared by the web view, the
  component, and the Storybook harness, and the reason it — not the disabled control — is the
  enforcement point is recorded at the function, in the `ScopeSelector` prop TSDoc, and in
  `Architecture-Decisions.md`.
- Stale results cannot survive the transition by construction: `resultsAreaState` ranks the
  invalid-query branch ahead of `results.length > 0`, so no async container effect has to land first.
  A test asserts the stale results are *replaced*, not merely covered.
- `find-book-lists.utils.test.ts`'s "covers exactly the books the flag exclusion clears" test derived
  its expectation from `excludeExtraMaterialBooks` itself, pinning the two extra-material definitions
  to each other — a genuine invariant test rather than a restatement of the implementation. (The
  in-review fix made that agreement structural as well.)
- Test coverage is layered without redundancy: the rule level (`find.utils.test.ts`), the rendered
  state (`find.component.test.tsx`), and the component library (`scope-selector.component.test.tsx`,
  both variants).
- A new localization key was added rather than repurposing `%webView_find_extraMaterialNotSearched%`,
  whose wording is specific to the book list; the distinction is documented at both key definitions.
- Comments are forward-facing throughout: `TODO(PT-4414)` on both halves of the exclusion, `{@link}`
  cross-references replacing the now-closed PT-4415 deferral notes, and no in-PR narration or
  review-finding IDs.
- `Architecture-Decisions.md` was updated as an Update on the existing `adr-find-*` entry rather than
  a new slug, and records *why* the query gate is the enforcement point.
- The disabled radio row dims its label for free via `tw:peer` / `tw:peer-disabled:opacity-50`, so
  the whole row reads as disabled rather than just the radio dot.

## Interview Notes

**Stated purpose**: close the `book`/`chapter` bypass left open by #2708, for a bug already reported
against a test package (Pkg 238, Power mode) and part of epic PT-4336.

**Design explanation (Step 3.3)**: The author gave a complete and accurate account of the central
decision. Both gated scopes resolve to `verseRef.book` at query time; `scope` is persisted per web
view via `useWebViewState` while `verseRef` is a live setting that moves on navigation, so the two
are independent and there is no transition that forces a user through a scope control before their
scope becomes blocked — they navigate into extra material with `book` already selected and arrive
directly. The disabled controls are therefore an affordance and `isFindQueryValid` is the
enforcement. The author enumerated the four consumers of the derived value (results-area state,
result-render suppression, Replace availability, and `gateStartSearch` in the web view) and
correctly identified that the results-area decision happens in render, ahead of any async container
effect, which is what prevents a window of stale results. The author also traced how the mechanism
unwinds under PT-4414 and noted that the `selectedBooks` path needs its own follow-up.

One inaccuracy in the author's account, corrected in the interview: the `TODO(PT-4414)` on
`isScopeBlockedByExtraMaterial` names only the scope rule; the book-list half is tied to the same
ticket by its own TODO in `find-book-lists.utils.ts`, not by a note on that one. Immaterial to the
design.

**Areas the author did not understand**: none. The author did not defer to AI on any question and
did not express uncertainty about any part of the change.

**Additional context volunteered**: none — the author had nothing further to flag at wrap-up.

**Discovered during the review, worth flagging beyond this branch**: `EmptyState`'s `id` prop is a
`data-testid`, not a DOM id. A first attempt at wiring `aria-describedby` to the placeholder
referenced a nonexistent element and was caught by a test; `ResultsPlaceholder`'s wrapper now
carries a real `id`. The naming is a live footgun for any other consumer who assumes `id` means
`id`, and is not specific to this branch.

## In-Review Quality Check

All checks run after the fixes, from the repo root:

- **`npm run build:platform-bible-react`** — succeeded; `dist/` regenerated and verified to contain
  both `ScopeWithRange` and `disabledScopeExplanations` in `index.d.ts`.
- **`npm run typecheck`** — clean, all four sub-checks.
- **`npm run lint`** — clean, zero errors. One pre-existing warning (`import/no-named-as-default` on
  `UnsubscriberAsyncList`), unrelated.
- **`prettier --check`** on the changed files — clean. (Run separately on purpose: `npm run lint`
  does not surface prettier violations in this repo.)
- **`npm test`** — core: 3703 passed / 2 skipped. Workspaces: 4721 tests across 10 workspaces, 0
  failures. Note the root `npm test` short-circuits at `test:core`, so the workspace half — which is
  where the find and scope-selector tests live — must be invoked separately.

**No code repairs were needed.** One pre-existing unrelated failure persists:
`.erb/scripts/electron-builder-snap-config.test.ts` expects `gnome-42-2204` and gets
`gnome-3-28-1804`, caused by a stale local `node_modules` (the test reads the *installed*
electron-builder template). This branch touches nothing under `.erb/`.

## Suggested Review Focus

- [ ] **The inline disabled-explanation in the dropdown variant.** This is a visible change to a
      shared design-system component, made during review on the reviewer's recommendation. A
      disabled option now grows a second line of muted text. No shipping consumer uses the dropdown
      variant with this prop, so nothing regresses today — but the presentation deserves a UX
      opinion before another consumer adopts it.
- [ ] **Whether the variant-dependent disabled-explanation rule should be promoted** from the
      architecture log into `Component-Builder-Patterns.md` as a standing rule. The constraint
      (a disabled Radix menu item is not focusable, so hover/focus-only affordances reach nobody)
      applies to any shared component that disables a control and owes the user a reason.
- [ ] **The placeholder wording**: "Find doesn't search extra material, such as glossaries and front
      matter. Choose books to search, or go to a Scripture book." Written during review to satisfy
      the zero-state guideline; it names both recovery paths but has not had a UX or localization
      review pass.
- [ ] **`EmptyState`'s `id` prop meaning `data-testid`.** Out of scope for this branch, but it
      produced a real (caught) defect during the review. Worth deciding whether to rename it or add
      a separate DOM-id prop.
- [ ] **Manual verification against a project with XXB/GLO** remains unchecked on the PR. Everything
      here is covered by unit tests; the end-to-end behavior in Power mode has not been re-confirmed
      since the reopen.

---

## Summary

[PT-3299](https://paratextstudio.atlassian.net/browse/PT-3299) was reopened: testing Pkg 238 in Power mode found that with the current reference in XXB, Find still returned results for the **Current book** and **Current chapter** scopes, labeled with the meaningless reference the original fix (#2708) exists to hide.

#2708 excluded extra material from Find's book lists. Those lists never reached the `book`/`chapter` scopes, which resolve `findScope` from the current scripture reference instead — and the navigation control offers every book the project has. This PR closes that path.

### Why review this

Part of the current epic ([PT-4336](https://paratextstudio.atlassian.net/browse/PT-4336), "Saroj and Donna can trust the app") — this is the reopened half of a sub-task already reported against a test package, so it is a known-visible bug rather than new work.

## Changes

- **New `isExtraMaterialBookId`** in `find-book-lists.utils.ts`, alongside the existing flag-clearing exclusion. A separate predicate rather than a second filter over `findScope`: the book lists narrow what is *offered*, the predicate rejects what the current reference *resolves to*.
- **`isFindQueryValid` rejects the `book`/`chapter` scopes** while the current reference sits in extra material. The query gate is the enforcement point, not the UI — `scope` is persisted per web view and the reference moves independently, so a user already in the book scope who then navigates into extra material never passes through a disabled control.
- **Two explanations.** The results area gets its own placeholder (the generic invalid-query wording points at a book picker that cannot fix this), and the scope selector disables both options with a tooltip.
- **New `disabledScopeExplanations` prop on `ScopeSelector`** (`lib/platform-bible-react`), mirroring the existing `disabledSectionExplanations`, honored in both the radio and dropdown variants.
- Architecture-Decisions.md updated with the reopen and the reasoning; `TODO(PT-4414)` markers tie both halves of the exclusion together so they can be removed as one when extra material becomes openable.

Also includes a merge of `main` (conflicts were only in the generated `lib/platform-bible-react/dist` bundles; resolved by taking main's side and regenerating from source).

## AI Involvement

AI-assisted. Claude wrote the implementation and tests under direction and review; the approach, the choice to gate at the query rather than the UI, and the wording were reviewed by the author.

## Testing

- [x] `npm run typecheck` — clean across all four projects
- [x] `npm run lint` — clean
- [x] `npm run format:check` — clean (only gitignored local docs flagged)
- [x] `npm test` — 3703 passing. One pre-existing unrelated failure in `.erb/scripts/electron-builder-snap-config.test.ts`, caused by a stale local `node_modules` (the test reads the *installed* electron-builder template); the branch touches nothing under `.erb/`.
- New tests: 5 extension tests cover the gate and the placeholder, 4 library tests cover the disabled options, with the `selectedBooks` scope and a scripture-book reference as negative controls. Deleting either half of the change fails tests.
- [ ] Manual verification against a project with XXB/GLO

## Risk Level

Low — narrows Find's behavior in a state that currently produces unusable results, behind a predicate with an explicit removal TODO. The `ScopeSelector` prop is additive and optional.

[PT-3299]: https://paratextstudio.atlassian.net/browse/PT-3299?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PT-4336]: https://paratextstudio.atlassian.net/browse/PT-4336?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2792)


katherinejensen00 (changes requested):
# Code Review — PR #2792

**PT-3299: Gate Find's current-book/chapter scopes on a searchable book**

| | |
|---|---|
| **PR** | https://github.com/paranext/paranext-core/pull/2792 |
| **Author** | jolierabideau |
| **Branch** | `pt-3299-remove-extra-books` → `main` |
| **Head reviewed** | `30f3daa5203` |
| **Merge base** | `7ce28a6ddd8` |
| **Diff** | 20 files, +3411 / −2739 (most of it the generated `lib/platform-bible-react/dist/**`, excluded from review) |
| **Review date** | 2026-09-11 |

---

## Verdict

**One merge blocker, and CI structurally cannot see it.** The new import in `find.utils.ts` pulls
`platform-bible-react` — and a bare `require("react")` — into the extension-host bundle, which
should stop `platform-scripture` from activating at runtime. Typecheck, lint, and the full test
suite are all green on it.

Everything else is comfortable follow-up work. The change itself is careful: the query-gate
placement is correct, the `selectedBooks` pruning is sound, stale results cannot survive the
transition by construction, and the `Canon` de-duplication is a genuine improvement.

## How this review was produced

Five independent passes, then adjudication:

1. **Developer / soundness** — logic, state coverage, test falsifiability, React correctness
2. **Architect** — layering, API surface, single-source-of-truth, ADR conformance
3. **Comments & documentation** — accuracy, density, forward-facing-comment rule
4. **Antagonistic pass** — attacked passes 1–3 and killed four findings outright
5. **`/code-review max`** — 10-angle multi-agent pass; caught the critical the other four missed

Every finding below was re-verified against the branch before inclusion. Findings that did not
survive scrutiny are listed in [Rejected findings](#rejected-findings) rather than dropped silently.

---

## Findings at a glance

| # | Severity | File | Line | Issue |
|---|---|---|---|---|
| 1 | **Critical** | `extensions/src/platform-scripture/src/find/find.utils.ts` | 13 | Import pulls `platform-bible-react`/`react` into the extension-host bundle |
| 2 | High | `lib/platform-bible-react/.../scope-selector.component.tsx` | 823 | Unconditional `flex-col` wrapper stacks every dropdown label |
| 3 | High | `extensions/.../find/find-book-lists.utils.ts` | 21 | `Canon.isExtraMaterial` is case-sensitive → gate fails open |
| 4 | Med-High | `lib/platform-bible-react/.../scope-selector.component.tsx` | 1025 | Disabled rows become `role="group"` in a `radiogroup`; Tab skips them |
| 5 | Medium | `extensions/.../find/find.component.tsx` | 1228 | Result count + prev/next survive an invalid query |
| 6 | Medium | `extensions/.../find/find.component.tsx` | 1180 | `aria-describedby` on a non-focusable span |
| 7 | Medium | `extensions/.../find/find.utils.ts` | 216 | `verse`/`selectedText` pass the gate but throw in `findScope` |
| 8 | Medium | `extensions/.../find/find.utils.ts` | 215 | `selectedBooks` restore path can still search `GLO` |
| 9 | Medium | `extensions/.../find/find.component.tsx` | 759 | Memo re-derives the rule instead of calling the predicate |
| 10 | Medium | `extensions/.../find/find.component.test.tsx` | 1016 | Real `TS2790` error CI never looks at |
| 11 | Medium | `lib/platform-bible-react/.../scope-selector.component.tsx` | 796 | Prop silently ignored for `selectedBooks`/`range` in dropdown |
| 12 | Medium | `extensions/.../find/find.component.tsx` | 1175 | Trigger de-emphasis and explanation key off different conditions |
| 13 | Medium | `lib/platform-bible-react/.../scope-selector.component.tsx` | 825 | Inline explanation at ~2:1 contrast, under WCAG AA |
| 14 | Low | `.context/standards/Architecture-Decisions.md` | 784 | Entry contradicts itself; uses an undocumented label |
| 15 | Low | `extensions/.../find/find-book-lists.utils.ts` | 25 | Inverted polarity in doc comment |
| 16 | Low | `extensions/.../find/find-book-lists.utils.test.ts` | 41 | Stale oracle comment |
| 17 | Low | `extensions/.../find/find-book-lists.utils.test.ts` | 77 | Test name promises a cross-check it no longer performs |
| 18 | Low | `extensions/.../find/find.component.test.tsx` | 889 | Backward-facing comment (raise once at PR level) |
| 19 | Low | `extensions/src/platform-scripture/src/find.web-view.tsx` | 725 | Comment describes a path the picker cannot produce |
| 20 | **Medium** | `extensions/src/platform-scripture/contributions/localizedStrings.json` | 465 | Both new keys are `en`-only, and nothing in the build will catch it |
| 21 | Low | `extensions/.../find/find.component.tsx` | 442 | Storybook autodocs will collide on the new DOM ids |

---

## Critical

### 1. The new import pulls `platform-bible-react` — and `react` — into the extension-host bundle

**File:** `extensions/src/platform-scripture/src/find/find.utils.ts:13`
**Anchor:** `import { isExtraMaterialBookId } from './find-book-lists.utils';`
**Severity:** Critical

This one line should stop `platform-scripture` from activating at runtime.

**The chain:**

```
main.ts:48
  → platform-scripture-finder-pdpe.model.ts:31
    → find.utils.ts:13                              ← new on this branch
      → find-book-lists.utils.ts:2
        → platform-bible-react/experimental
          → require("react")
```

**Why it breaks:**

- `platform-bible-react` is **not** in `extensions/webpack/webpack.config.base.ts` externals → it gets bundled.
- `react`, `react/jsx-runtime`, `react-dom` **are** externals → they emit bare `require(...)` calls.
- `EXTENSION_INTERFACE_MODULE_SPECIFIERS` (`src/extension-host/data/extension-interface-modules.data.ts`)
  is only `@papi/backend`, `@papi/core`, `@sillsdev/scripture`, `crypto`, `platform-bible-utils` —
  **`react` is not there**, so the host cannot supply it.

**Measured:** `extensions/dist/platform-scripture/src/main.js` is 2,096,512 bytes on the merge base
with zero react references, and 11,931,841 bytes on this branch (5.7×) with external factories for
react, react-dom and jsx-runtime, plus six `lucide-react` icon modules and an `import './index.css'`.

**Confirmed new:** base `find.utils.ts` had no import of `find-book-lists.utils` at all.

**Fix:** `isExtraMaterialBookId` needs only `Canon`. Move it to a host-safe module, or call
`Canon.isExtraMaterial` directly from `find.utils.ts`.

**Durable fix:** port `extension-host-import-boundary.test.ts` from `platform-scripture-editor`
(entry `main.ts`). `platform-scripture` has no such guard — which is exactly why this produced zero
build, lint, and test failures.

---

## High

### 2. Every reference-derived dropdown label now wraps onto two lines

**File:** `lib/platform-bible-react/src/components/advanced/scope-selector/scope-selector.component.tsx:823`
**Anchor:** `<span className="tw:flex tw:flex-col tw:gap-0.5">`
**Severity:** High

The wrapper is **unconditional** — not gated on `disabledExplanation`. `renderScopeLabel` returns a
two-child fragment (bare text node + `<span>: {suffix}</span>`). Those were flex items of
`DropdownMenuItem` (`tw:flex tw:items-center tw:gap-1.5`), a row. Inside `flex-col` they stack.

**Live victim:** `checklist.web-view.tsx:838-840` — `variant="dropdown"`, `currentScrRef={liveScrRef}`,
`availableScopes={['verse','chapter','book','range']}`. Three of four items have a `scrRefSuffix`, so
"Current chapter" / ": GEN 1" render on two lines, doubling every menu row — in a panel this PR does
not otherwise touch.

**Why no test caught it:** jsdom does no layout, and the new assertions use `toHaveTextContent`,
which normalizes whitespace. Only a visual or Storybook check could catch this.

**Fix:** wrap only the label in its own inline span, or drop the wrapper when there is no explanation.

### 3. `Canon.isExtraMaterial` is case-sensitive, so the gate fails open on a lowercase book id

**File:** `extensions/src/platform-scripture/src/find/find-book-lists.utils.ts:21`
**Anchor:** `return Canon.isExtraMaterial(bookId);`
**Severity:** High *(reachability disputed)*

`Canon.isExtraMaterial` is `isBookIdValid(id) && nonCanonicalIds.includes(id)`. The validity check
upper-cases; the `includes` does not. So `Canon.isExtraMaterial('glo') === false`.

If `verseRef.book` is ever lowercase, `isFindQueryValid` returns `true`, both scopes stay enabled,
and Find searches the glossary again — the exact reopened bug.

**Honest caveat:** two passes found this asymmetry independently and disagreed on reachability. One
could not construct a live path and declined to raise it. What tips the balance: `ScopeSelector`
defensively `.toUpperCase()`s the same field at lines 280 and 491 — the codebase already declines to
trust this casing.

**Fix:** one `.toUpperCase()`. Cheaper than proving unreachability.

### 4. Disabled radio rows become `role="group"` children of a `radiogroup`, and Tab skips them

**File:** `lib/platform-bible-react/src/components/advanced/scope-selector/scope-selector.component.tsx:1025`
**Anchor:** `<DisabledActionTooltip`
**Severity:** Medium-High

`DisabledTooltipWrapper` sets `role="group"` + `tabIndex={0}` while disabled. `radiogroup`'s required
owned element is `radio`; an intervening `group` can break set-size/position announcements, and each
wrapper adds a tab stop inside a roving-focus group that specifies exactly one.

In Find's shipping configuration (`availableScopes={['chapter','book','selectedBooks']}`) both gated
scopes are disabled and precede the only focusable radio, so forward Tab skips them and the
explanation reaches nobody — the opposite of what the ADR update claims the radio variant achieves.

**Caveat:** worth confirming with a screen reader before acting, but the markup claim is checkable
from the diff alone.

---

## Medium

### 5. Result count and prev/next survive the transition even though the rows do not

**File:** `extensions/src/platform-scripture/src/find/find.component.tsx:1228`
**Anchor:** `{visibleResults.length > 0 && (`

Rows are gated at line 1341 (`if (!isSearchQueryValid) return undefined;`) but the count block and
navigation arrows are not.

**Repro:** search "God" with `scope='book'` in GEN → 12 results → navigate the scroll group to GLO.
Rows are replaced by the new placeholder while the header still reads "1 of 12", with working arrows
that call `onFocusedResultChange` and drive the editor.

The new test only asserts the match text is gone. Previously reachable only by deselecting all books;
navigation now makes it the common path.

### 6. `aria-describedby` sits on a non-focusable span, so it is never announced

**File:** `extensions/src/platform-scripture/src/find/find.component.tsx:1180`
**Anchor:** `aria-describedby={`

The attribute is on a role-less `<span>` nested inside the Popover trigger `<Button>`. A button is
atomic to assistive technology — descendants contribute to its *name*, not a description. A
screen-reader user hears "Showing GLO, button" with no mention of why nothing will run.

**Fix:** move it to the `Button` at line 1161.

**Test gap:** the accompanying test asserts `document.querySelector('[aria-describedby=…]')`, which
passes for *any* element carrying the attribute — it locks in the placement without checking the
property it exists to guarantee. Assert on `getByRole('button', { name: /Showing/ })` instead.

### 7. `verse`/`selectedText` pass the gate but throw in `findScope`

**File:** `extensions/src/platform-scripture/src/find/find.utils.ts:216`
**Anchor:** `return !isScopeBlockedByExtraMaterial(params.scope, params.currentBookId);`

The new tests deliberately pin `verse` and `selectedText` as **valid**, but `find.web-view.tsx:1096-1097`'s
`findScope` switch has `default: throw new Error('Unsupported scope')` for exactly those. Adding
either to `availableScopes` now passes the gate and throws during render rather than blocking the
search — a worse failure than the one the explicit scope naming was meant to prevent.

This is the strongest concrete argument for deriving the two rules from one another.

### 8. `selectedBooks` still only asks "non-empty?", and the restore path can slip through

**File:** `extensions/src/platform-scripture/src/find/find.utils.ts:215`
**Anchor:** `if (params.scope === 'selectedBooks') return params.selectedBookIds.length > 0;`

A tab persisted with `selectedBookIds: ['GLO']` passes the gate during the window before
`availableBookIds` resolves — `prunePresentBookIds` returns the list untouched while it is
`undefined` — and the restore-path auto-search fires, with `findScope` mapping straight to
`[{bookId:'GLO'}]`.

That is the reopened bug, reachable today, through the one scope the PR body explicitly defers.
Worth at least a note in the PR that the follow-up is load-bearing.

### 9. The disabled-explanation memo re-derives the rule instead of calling the predicate

**File:** `extensions/src/platform-scripture/src/find/find.component.tsx:759`
**Anchor:** `const disabledScopeExplanations = useMemo(() => {`

It calls the lower-level `isExtraMaterialBookId(verseRef.book)` and re-lists `{ book, chapter }` as a
literal, while `isScopeBlockedByExtraMaterial`'s TSDoc and the ADR both claim the two sides "decide
from one rule rather than two copies of it". Extend the predicate later and the gate rejects a scope
the picker still offers as enabled.

**Fix:** `availableScopes.filter((s) => isScopeBlockedByExtraMaterial(s, verseRef.book))`.

**Scope note:** the antagonistic pass correctly observed that the scope list is hard-coded inside the
predicate too, and that this was a documented deliberate choice. This finding is about the **memo**
specifically — not a request to revisit the design.

### 10. A real type error that CI structurally cannot see

**File:** `extensions/src/platform-scripture/src/find/find.component.test.tsx:1016`
**Anchor:** `delete stringsWithoutExplanation['%webView_find_extraMaterialNotSearchedScope%'];`

`tsc` reports `TS2790: The operand of a 'delete' operator must be optional`.

It is invisible because `extensions/src/platform-scripture/tsconfig.json` excludes `**/*.test.tsx`
**and** the `extensions` workspace has no `typecheck` script — so `npm run typecheck` never examines
any test file in this extension.

**Fix:** annotate `STRINGS` with the optional-key type. The tooling gap deserves its own ticket.

### 11. `disabledScopeExplanations` is silently ignored for `selectedBooks`/`range` in the dropdown

**File:** `lib/platform-bible-react/src/components/advanced/scope-selector/scope-selector.component.tsx:796`
**Anchor:** `const disabledExplanation = disabledScopeExplanations?.[value];`

Only `simpleScopes` consults the prop (`displayedScopes.filter(v => v !== 'selectedBooks' && v !== 'range')`,
line 506). The dialog-launcher items at 835 and 854 never read it and never set `disabled`. The type
is `Partial<Record<ScopeWithRange, string>>` and the radio variant honors all six keys, so the same
prop value behaves differently per variant with no type or test signal.

**Severity note:** three of five passes rated this Important. Downgraded deliberately — the only
consumer of the prop is Find, which uses the radio variant; `checklist` uses dropdown but never
passes the prop. **Blast radius today is zero.**

**Fix:** narrow the prop type to the scopes the dropdown honors, or one sentence of TSDoc.

### 12. Trigger de-emphasis and the explanation key off different conditions

**File:** `extensions/src/platform-scripture/src/find/find.component.tsx:1175`
**Anchor:** `className={`

De-emphasis keys off `isBlockedByExtraMaterial`; `aria-describedby` and the placeholder key off
`resultsAreaState === 'extraMaterialPrompt'`. With the search box **empty** in GLO, `resultsAreaState`
is `'idlePrompt'` (checked first at line 706), so the trigger renders italic and muted with no
explanation anywhere on screen — in exactly the state, before typing, where the reason is most useful.

### 13. The inline explanation is the least legible text in the menu

**File:** `lib/platform-bible-react/src/components/advanced/scope-selector/scope-selector.component.tsx:825`
**Anchor:** `{disabledExplanation && (`

It only renders on an item carrying `data-disabled`, and `DropdownMenuItem`'s base class includes
`tw:data-disabled:opacity-50`. So `tw:text-xs tw:text-muted-foreground` at half opacity — roughly
2:1 against the popover background, under WCAG AA.

It also lacks `tw:min-w-0`, so a long string can push the menu past
`--radix-dropdown-menu-trigger-width`, and `isDropdownNarrow` suppresses the short suffix but not the
long explanation.

---

## Low

### 14. The ADR entry now contradicts itself, and uses a label the log does not define

**File:** `.context/standards/Architecture-Decisions.md:784`
**Anchor:** `## adr-find-narrows-book-lists: Find excludes extra material by narrowing its book lists, not by gating its scopes`

Verified directly. The heading, `**Status:** Accepted` (788), and the Decision body (798-800:
"deliberately not gated… PT-4415 covers them") all still assert the opposite of what shipped. The
corrections live only in two bullets 20–45 lines below.

The log's own preamble (line 9) says a decision that no longer holds "has its `**Status:**` revised
or an `**Amended YYYY-MM-DD:**` note appended". **Base had 4 `**Amended` and 0 `**Update`** — the
`**Update (…)**` label is novel here.

This file is read for precedent by people *and* by agents (`pt10-reuse-scout` during
`/investigate-prd`). A heading scan currently returns "not by gating its scopes" as the standing
decision, and hands a reader PT-4415 as open when this PR closed it.

Appending to the existing slug rather than adding a new one was the right call.

### 15. Inverted polarity in a doc comment

**File:** `extensions/src/platform-scripture/src/find/find-book-lists.utils.ts:25`
**Anchor:** `* Book numbers {@link isExtraMaterialBookId} rejects. Precomputed because the set is fixed for a`

The predicate returns `true` for exactly these books, so the set is what it *accepts*. A reader who
trusts this concludes the set holds scripture books and "fixes" `excludeExtraMaterialBooks` by
inverting its check.

### 16. Stale oracle comment

**File:** `extensions/src/platform-scripture/src/find/find-book-lists.utils.test.ts:41`
**Anchor:** `* Canon.nonCanonicalIds, which the code under test derives from — not an independent oracle.`

The code now derives from `Canon.isExtraMaterial`. This is the comment explaining *why* the literal
list exists, so pointing it the wrong way is costly when someone reconciles a divergence.

### 17. The one tautological assertion

**File:** `extensions/src/platform-scripture/src/find/find-book-lists.utils.test.ts:77`
**Anchor:** `expect(clearedIds).toEqual(Canon.allBookIds.filter(isExtraMaterialBookId));`

Both sides reduce to the same expression now that `EXTRA_MATERIAL_BOOK_NUMBERS` is defined from it.

**In fairness:** the antagonistic pass showed it is not purely circular — mutating `index + 1` to
`index` does fail it, so it still pins the flag arithmetic — and the author's own comment already
says "by construction". The real issue is that the test **name** promises a cross-check it no longer
performs.

**Fix:** rename it, or assert against the 15 literal ids, which is the only independent oracle left.

### 18. Backward-facing test header (raising once at PR level)

**File:** `extensions/src/platform-scripture/src/find/find.component.test.tsx:889`
**Anchor:** `// The bug these cover: the book and chapter scopes resolve to the CURRENT reference's book rather`

"The bug these cover" plus past-tense "returned" is change narration per
`.claude/rules/code-quality/forward-facing-comments.md`. The middle clause survives the
strip-the-PR-context test and is worth keeping.

Three pre-existing comments of the same shape sit next to lines already being rewritten —
`find.component.tsx:668`, `find.stories.tsx:483`, `find.utils.ts:197` ("they previously each
hand-rolled this rule"). Cheap to cut while in there; not this PR's debt otherwise.

### 19. The `find.web-view.tsx` comment describes a path the picker cannot produce

**File:** `extensions/src/platform-scripture/src/find.web-view.tsx:725`
**Anchor:** `// verseRefSetting.book rather than from them. The navigation control offers every book the`

`BookChapterControl` drops FRT/BAK/OTH/INT/CNC/GLO/TDX/NDX and offers only XXA–XXG
(`book-chapter-control.utils.ts:228`). The gate is still correctly sized — those 8 arrive via
`scroll-group-navigation.commands.ts:114`, a persisted scroll-group ref, or a resource click.

But it means the new tests and the Storybook story pin `GLO`, a state the picker cannot reach, while
the reachable XXA–XXG case gets a message naming "glossaries and front matter". Worth a sanity check
that the wording fits the states users actually hit.

### 20. Both new localization keys are `en`-only, and nothing in the build will catch it

**File:** `extensions/src/platform-scripture/contributions/localizedStrings.json:465`
**Anchor:** `"%webView_find_extraMaterialNotSearchedScope%": "Find doesn't search extra material.",`
**Severity:** Medium *(raised from Low on further evidence — see below)*

`Localization-Guide.md:492` lists "Keys exist in `en` but not `es`" as **Blocking Issue #2**. Both
new keys are `en`-only.

This initially looked like a low-signal pre-existing pattern. It is not — the repo treats en/es
parity as a real convention and **defends it by hand, because nothing automated does**:

- `extensions/src/platform-scripture-editor/src/localized-strings.test.ts:30-31` —
  *"Nothing in the build enforces en/es parity — a key present in `en` and missing from `es` fails no
  other check — so this is the guard against a future edit dropping one language."* Its suites then
  assert per key: has an English label / has a Spanish label / Spanish label differs from English.
- `extensions/src/legacy-comment-manager/src/localized-strings.test.ts:69` — same shape, same
  rationale.
- `extensions/src/platform-scripture/` has **no `localized-strings.test.ts` at all** (confirmed).
  That is precisely why its `es` block has drifted to 72 of 94 `%webView_find_*` keys.

So there is no mechanism anywhere that will flag these two keys. A Spanish user gets English text or
a raw key in the results area.

**Ask:** add the two `es` values alongside the `en` ones. Optionally also
`%webView_find_extraMaterialNotSearched%` from the predecessor PR, which is `es`-less for the same
reason. Longer term, `platform-scripture` is the extension most in need of the parity test its two
siblings already have.

> **Bonus defect found while verifying this.** `Localization-Guide.md:228` states: *"Always provide
> both `en` AND `es` translations. **Both are required for the build to pass.**"* That is factually
> false — the in-code comment quoted above says so explicitly, and this PR is the proof. The rule at
> line 492 stands; the mechanism the guide claims enforces it does not exist. Worth a separate
> doc-fix ticket, since anyone relying on line 228 will assume CI has their back.

### 21. Storybook autodocs will collide on the new DOM ids

**File:** `extensions/src/platform-scripture/src/find/find.component.tsx:442`
**Anchor:** `<div id={id} className="tw:flex tw:min-h-48 …">`

All four placeholders now emit real global ids (previously `id` reached only `EmptyState`'s
`data-testid` — confirmed at `empty-state.component.tsx:29`, so there is no wrapper/EmptyState
collision). `find.stories.tsx` is `tags: ['autodocs']` and the Docs page renders every story into one
document, so several stories emit `id="find-idle-placeholder"` simultaneously.
`find-extra-material-placeholder` is unique by accident, not construction.

---

## Rejected findings

These were raised by one or more passes and **killed on verification**. Listed so they are not
re-litigated on the next review.

### `ScopeWithRange` on the stable entry point — KEEP IT

`lib/platform-bible-react/src/index.ts:315`

Raised three times as an unwarranted one-way API expansion. It is **correct**. `ScopeSelector` ships
from stable `index.ts:142` and its props are typed `ScopeWithRange` (lines 100, 106, 109) — so before
this PR a stable-entry consumer could not name the component's own prop types. Both real consumers
prove it: `checklist.web-view.tsx` and `find.component.tsx` each import the component from one entry
point and the type from the other, in adjacent import blocks.

This PR closed a real hole. The only residual nit runs the *other* way: `experimental.ts:47` is now
the redundant line — and even that is the correct non-breaking transitional state.

### Localization key naming convention — premise is false

A reviewer claimed the file's convention is `_tooltip` for tooltips and `...Prompt` for placeholders.
Neither holds: `%webView_find_searchPlaceholder%` (line 478) is the input placeholder, and `...Prompt`
(479–480) is the **results-area** role — which is exactly what the new key does. `_tooltip` is 7 keys
out of 948, and would be actively wrong here since the dropdown variant renders the string inline
rather than as a tooltip.

### Comment density — not a defect

Flagged as "26%, too much". Measured against main: the changed files moved **0 to 3 points**, on a
baseline already at 21–68% (`find.web-view.tsx` 25%, `checklist.web-view.tsx` 21%,
`book-chapter-control.component.tsx` 22%).

Two of the four "duplicated rationale" sites are **mandated**: `Code-Style-Guide.md:146` requires
API-surface TSDoc to stand on its own ("assume readers cannot look up anything outside the comment
itself"), and `CLAUDE.md` requires the ADR entry. De-duplicating site 1 into site 2 would violate the
API-TSDoc standard.

### `@param`/`@returns` on the one-line delegate — standard permits it

`Code-Style-Guide.md:156` explicitly encourages rich TSDoc for extension-internal helpers, and the
sibling function in the same file already carries both tags.

### "Promote the generalized rule out of the ADR" — misreads the rule

`CLAUDE.md`'s "Promote settled conventions" says to **also** fold it into the relevant standard —
additive, not exclusive. The log keeps the *why* and the history.

---

## CI status — red, but not from this code

All three platform builds fail:

| Job | Result | Failing step |
|---|---|---|
| Build on macos-15 | failure | `Update MacPorts ports tree` (infrastructure) |
| Build on windows-latest | cancelled | `Install packages` (fail-fast) |
| Build on ubuntu-22.04 | cancelled | `type checking` (fail-fast) |

The macOS MacPorts sync failed and fail-fast cancelled the other two. CI ran on the current head
(`30f3daa`).

**The branch is 4 commits behind `main`, and two of them are the fixes for exactly this:**

- `bc056ca` — ci: give react unit tests 15 s and stop the matrix failing fast (#2759)
- `af4072e` — ci: stop a flaky Linux snap install failing the whole build (#2797)

Merging `main` should turn CI green — and a clean run is needed anyway to confirm finding 1 once
it is fixed.

---

## Recommended order

1. **Finding 1** — blocks merge. One-line fix plus the import-boundary test.
2. **Findings 2 and 3** — both one-liners, both affect shipping behavior.
3. **Findings 5, 6, 7, 10** — correctness and a11y, all concrete.
4. **Merge `main`** to clear CI, then re-run to confirm 1.
5. **Finding 20** — two `es` string values. Trivial to do, and nothing in CI will remind you.
6. **Findings 8, 9, 11–13** — comfortable follow-ups.
7. **Findings 14–19, 21** — documentation cleanups, batchable in one pass.

---

## Verified clean

Recorded so these are not re-reviewed:

- `isFindQueryValid`'s rewrite is behavior-identical for every input
- `Canon.nonCanonicalIds` and `allBookIds.filter(isExtraMaterial)` are the same 15 ids on the pinned
  `@sillsdev/scripture` — `excludeExtraMaterialBooks` did not change behavior
- No test was deleted or loosened
- PT-4415 was genuinely closed; live TODOs correctly re-pointed at the open PT-4414
- No circular import, no TDZ hazard at module init
- All four `isFindQueryValid` call sites updated; none can pass an undefined book
- Both new keys alphabetically placed and registered in `FIND_LOCALIZED_STRING_KEYS`
- No new lint suppressions, no shadcn files touched, no keyboard handlers added, no `papi.d.ts`
  regeneration needed
- `npx vitest run scope-selector.component.test.tsx` passes 18/18
- No domain leak into `platform-bible-react` — the prop carries no notion of extra material
- Enforcement layer is correct: `gateStartSearch` is the only path to `beginFindJob`, and the
  pre-existing `shouldClearResultsForInvalidQuery` effect already abandons an in-flight job
- `EmptyState`'s `id` is a `data-testid`, so the new wrapper `id` creates no duplicate

> @katherinejensen00 reviewed 20 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2792#-:-P1Gh9-H2Z9NUHW5WP9t:b-1a9x0t)* status: all files reviewed, 1 unresolved discussion (waiting on jolierabideau).

___
*[`extensions/src/platform-scripture/src/find/find.utils.ts` line 13 at r1](https://reviewable.io/reviews/paranext/paranext-core/2792#-P1GhFYOE7fKjEhRt8Y2:-P1GhFYOE7fKjEhRt8Y3:b-es1fyx) ([raw file](https://github.com/paranext/paranext-core/blob/30f3daa520312e4caaf347809909432f15dfb2c6/extensions/src/platform-scripture/src/find/find.utils.ts#L13)):*
> ```XML
> import { FindJobStatus, FindJobStatusReport, FindOptions } from 'platform-scripture';
> import type { OpenProjectTabWithWebView } from '../hooks/use-open-project-tabs';
> import { isExtraMaterialBookId } from './find-book-lists.utils';
> ```

### 1. The new import pulls `platform-bible-react` — and `react` — into the extension-host bundle

**File:** `extensions/src/platform-scripture/src/find/find.utils.ts:13`
**Anchor:** `import { isExtraMaterialBookId } from './find-book-lists.utils';`
**Severity:** Critical

This one line should stop `platform-scripture` from activating at runtime.

**The chain:**

```
main.ts:48
  → platform-scripture-finder-pdpe.model.ts:31
    → find.utils.ts:13                              ← new on this branch
      → find-book-lists.utils.ts:2
        → platform-bible-react/experimental
          → require("react")
```

**Why it breaks:**

- `platform-bible-react` is **not** in `extensions/webpack/webpack.config.base.ts` externals → it gets bundled.
- `react`, `react/jsx-runtime`, `react-dom` **are** externals → they emit bare `require(...)` calls.
- `EXTENSION_INTERFACE_MODULE_SPECIFIERS` (`src/extension-host/data/extension-interface-modules.data.ts`)
  is only `@papi/backend`, `@papi/core`, `@sillsdev/scripture`, `crypto`, `platform-bible-utils` —
  **`react` is not there**, so the host cannot supply it.

**Measured:** `extensions/dist/platform-scripture/src/main.js` is 2,096,512 bytes on the merge base
with zero react references, and 11,931,841 bytes on this branch (5.7×) with external factories for
react, react-dom and jsx-runtime, plus six `lucide-react` icon modules and an `import './index.css'`.

**Confirmed new:** base `find.utils.ts` had no import of `find-book-lists.utils` at all.

**Fix:** `isExtraMaterialBookId` needs only `Canon`. Move it to a host-safe module, or call
`Canon.isExtraMaterial` directly from `find.utils.ts`.

**Durable fix:** port `extension-host-import-boundary.test.ts` from `platform-scripture-editor`
(entry `main.ts`). `platform-scripture` has no such guard — which is exactly why this produced zero
build, lint, and test failures.

jolierabideau:
# Response to review

All 21 findings addressed. Head is now `d598c803b51`; `main` is merged in (conflicts were only the generated `lib/platform-bible-react/dist` bundles again — resolved to main's side and regenerated from source).

## Critical

**1 — import pulls `platform-bible-react`/`react` into the extension-host bundle.** Confirmed and fixed. `isExtraMaterialBookId` moved to `find/extra-material.utils.ts`, which imports only `@sillsdev/scripture`; `find.utils.ts` and `find.component.tsx` import it from there, and `find-book-lists.utils.ts` now imports it too so the flag-clearing set and the gate still share one predicate. Also made `find.utils.ts`'s `Scope` import explicitly `import type`, which it was relying on the compiler to elide.

Took the durable fix as well: `extension-host-import-boundary.test.ts` is ported to `platform-scripture`. Falsified it before trusting it — pointing `find.utils.ts` back at `find-book-lists.utils.ts` fails it with `platform-bible-react/experimental` and `platform-bible-utils/experimental`; restoring passes. The webpack module listing now shows only `find.utils.ts` and `extra-material.utils.ts` under `find/*.ts` in the host graph.

Two things surfaced porting it. `@eten-tech-foundation/scripture-utilities` is a pre-existing value import in three host-reachable data-provider files; it isn't a webpack external, so it's bundled and pulls in no external of its own — allowlisted with that reasoning rather than left failing. And the walker's `require(...)` regex matches comment text, so the new module's TSDoc is worded to avoid the literal.

## High

**2 — unconditional `flex-col` stacks every dropdown label.** Confirmed, including the `checklist.web-view.tsx` victim. Kept the wrapper unconditional (element-type stability) and re-established a row around `renderScopeLabel`'s two children inside it, so the no-explanation case lays out exactly as before.

**3 — `Canon.isExtraMaterial` case sensitivity.** Fixed with the `.toUpperCase()`. I didn't try to prove unreachability either; a gate that fails open is the bug being closed. Covered by `it.each(['glo', 'Glo', 'xxb'])`.

**4 — `role="group"` + tab stops inside a `radiogroup`.** Fixed by removing the focusable wrapper from the radio variant entirely: the explanation now renders inline under the label with `aria-describedby` from the radio, matching what the dropdown already does and for the same reason. A test asserts the `radiogroup` contains no `[role="group"]` and no `[tabindex="0"]`.

This makes both variants consistent, which also resolved your second suggested-focus item — the rule is promoted into `Component-Builder-Patterns.md` as "Explaining Why a Control Is Disabled" (with the contrast and `min-w-0` gotchas from finding 13), and recorded in the ADR as a generalized rule.

## Medium

**5 — count + prev/next survive an invalid query.** Fixed; both are gated on `isSearchQueryValid`. Test asserts "1 of 1" and both arrow buttons are gone in the extra-material state.

**6 — `aria-describedby` on a non-focusable span.** Moved to the `Button`. Also took the test-gap note: the test now starts from `getByRole('button', { name: /Showing/ })`, reads its `aria-describedby`, and follows the id to the element — so it checks the property rather than the attribute's presence.

**7 — `verse`/`selectedText` pass the gate but throw in `findScope`.** You're right that this is the argument for deriving the rules from one another. `FIND_AVAILABLE_SCOPES` is now an exported constant in `find.utils.ts`, used by `isFindQueryValid` (rejects anything outside it), by the disabled-explanation memo, and as the picker's `availableScopes`. The two tests that pinned those scopes as valid now assert the opposite, plus one that filters the whole `Scope` union through the gate and asserts the accepted set equals `FIND_AVAILABLE_SCOPES`.

**8 — `selectedBooks` restore path.** Went further than the note you asked for and closed it: the gate asks whether any *searchable* book remains, and `findScope` drops extra material from the selection. Neither depends on `availableBookIds` having resolved, so the window you described isn't reachable. Two tests (all-extra-material rejected; mixed selection accepted).

**9 — memo re-derives the rule.** Fixed as suggested — `FIND_AVAILABLE_SCOPES.filter((s) => isScopeBlockedByExtraMaterial(s, verseRef.book))`. Understood that this was about the memo, not the design.

**10 — `TS2790` CI can't see.** Rebuilt as `Object.fromEntries(Object.entries(STRINGS).filter(...))` — no `delete`, no unused binding, and the comment records why. The tooling gap (tsconfig excludes `**/*.test.tsx` and the workspace has no `typecheck` script) is not fixed here and does deserve its own ticket.

**11 — prop ignored for `selectedBooks`/`range` in the dropdown.** Took the TSDoc option rather than narrowing the type, since the radio variant legitimately honors all six.

**12 — de-emphasis and explanation key off different conditions.** Fixed at the root: `extraMaterialPrompt` now outranks the empty-term `idlePrompt`, because "enter search text" is false while the scope is blocked, and before typing is exactly where the reason is most useful. Both the trigger's styling and its description then key off one derived `isExtraMaterialExplained`, so they cannot diverge — and the id can only be pointed at while the element carrying it renders. New test covers the empty-search-box case.

**13 — contrast and overflow.** `tw:text-foreground` (a parent `opacity` is a ceiling children can't exceed, so the muted color had to go) plus `tw:min-w-0` on the column.

**20 — `en`-only keys.** Added `es` values for all three, including `%webView_find_extraMaterialNotSearched%` from the predecessor PR. Also added `platform-scripture/src/localized-strings.test.ts` — scoped to the extra-material family rather than the whole `%webView_find_*` set, since closing the other 22 needs translations, not a test. The file says so and is where they belong as they land.

On the bonus defect: you're right, `Localization-Guide.md:228` was false. Fixed in place rather than ticketed — it now says parity is a review requirement, names the per-extension test as the only guard, and points at the three extensions that have one.

## Low

**14 — ADR self-contradiction and undocumented label.** Fixed all four parts: heading no longer reads as the standing decision, `**Status:**` records the amendments, the superseded sentence in the Decision body is marked inline, and both `**Update (…)**` labels are now `**Amended YYYY-MM-DD**` per the log's preamble. Added a third amendment for this round.

**15, 16, 17** — polarity corrected; oracle comment now says what the literals actually are (the one definition in the file independent of `Canon.isExtraMaterial`); the cross-check asserts against those literals rather than against `Canon.allBookIds.filter(isExtraMaterialBookId)`, which makes it match its name again.

**18** — cut the test header and all three of the pre-existing siblings you named.

**19** — comment corrected: it now names `BookChapterControl`'s XXA–XXG and the three routes the other eight arrive by.

**21** — placeholder DOM ids are per-instance via `useId`, with `data-testid` as a separate prop (`ResultsPlaceholder` now takes `domId` and `testId`). A test renders two `Find`s and asserts their `aria-describedby` values differ.

## Back to you

Three things I'm not deciding unilaterally:

1. **Placeholder wording** — "Find doesn't search extra material, such as glossaries and front matter. Choose books to search, or go to a Scripture book." Unchanged, and still unreviewed by UX. Your finding 19 point stands: XXA–XXG is the reachable-by-picker case and "glossaries and front matter" doesn't describe it. The category word "extra material" carries it and "such as" is illustrative, so I left it — but it's a UX call, not mine.
2. **`EmptyState`'s `id` meaning `data-testid`** — worked around rather than fixed (the `domId`/`testId` split above). Renaming it is a shared-component change with other consumers; worth its own ticket.
3. **Manual verification against a project with XXB/GLO** — still unchecked. Everything here is unit-tested; the end-to-end Power-mode behavior has not been re-confirmed since the reopen.

## Verification

From the repo root, after all fixes:

- `npm run typecheck` — clean, all four sub-checks
- `npm run lint` — 0 errors (one pre-existing `import/no-named-as-default` warning on `UnsubscriberAsyncList`, unrelated)
- `prettier --check` on the changed areas — clean (run separately; `npm run lint` doesn't surface prettier violations here)
- `npm test` — 4529 passed / 16 skipped core, plus all 10 workspaces green, 0 failures. The `.erb/scripts/electron-builder-snap-config.test.ts` failure noted in the PR body no longer reproduces after the merge.
- `npm run build:platform-bible-react` and `npm run build:extensions` — both succeed

katherinejensen00 (approved):
Thanks for making those changes, Jolie! There is one ADR description that could use updating, but that is a minor thing and everything else looks good so this PR is approved.

Architecture-Decisions.md:1214

The entry contradicts itself again. Item 3 adds a filter inside findScope, which Alternatives (line 1164) still lists as rejected and the 09-08 amendment (1185-1186) credits the gate for avoiding. Item 3 also says the query gate "drops" extra material; it only checks whether a searchable book remains. Status (1147) leaves out the 09-09 amendment.

None of the three amendments is on main, and 09-09 describes a radio tooltip that never ships. Would it be cleaner to fold them into one amendment that describes what merges, and mark line 1164 as adopted for selectedBooks?


:lgtm:

> @katherinejensen00 reviewed 25 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2792#-:-P1VUJmcBNq05LzYbbMj:bg6itpb)* status: all files reviewed, 1 unresolved discussion (waiting on jolierabideau).

jolierabideau:
# Response to review round 3

Head is now `b5359cd2b1b`; `main` is merged in and the PR shows mergeable again.

## ADR entry (`Architecture-Decisions.md`)

Took the fold you suggested. All three points confirmed against the code first:

- **Item 3 contradicts the rejected alternative.** Correct — `find.web-view.tsx:1101-1102` filters extra material out of `selectedBookIds` inside `findScope`, which is the "Filter `findScope` before the search runs" alternative. It is now marked adopted for the `selectedBooks` scope specifically, with the reason the original rejection didn't anticipate: a persisted selection is pruned against the project's book list only once that list resolves, so the race has no single-layer answer. The rejection still stands for the `book`/`chapter` bypass it was written about.
- **"Drops" is wrong for the query gate.** Correct — `isFindQueryValid` uses `.some((bookId) => !isExtraMaterialBookId(bookId))`, so it asks whether a searchable book remains and never modifies the selection. The amendment now says which of the two layers does which.
- **Status omitted 09-09.** Correct, and moot after the fold.

On the fold itself: you're right that none of the three amendments is on `main`, so there was no history worth preserving between them — and the 09-09 text described the focusable-wrapper tooltip that 09-12 then replaced, so the entry documented a shape that never ships. There is now one `**Amended 2026-09-14**` describing what merges: the query gate as the enforcement point, `FIND_AVAILABLE_SCOPES` as the one list behind the gate and the picker, the `selectedBooks` two-layer exception, the host-import boundary, and the inline-explanation rule for both `ScopeSelector` variants.

Also retitled the entry — the heading still read "and (as amended) by gating the reference-derived scopes", which stopped being accurate when `selectedBooks` was gated too.

## Merge of `main`

Conflicts were the generated `lib/platform-bible-react/dist` bundles again (resolved to main's side and regenerated from source), plus one real conflict in `find.component.test.tsx`: `main` added a `Find — recent searches menu` suite at the end of the file where this branch added `Find — current reference in extra material`. Both kept; resolved from the three merge stages rather than by hand-editing markers, so neither suite lost content.

## Verification

From the repo root, after the merge and the ADR change:

- `npm run typecheck` — clean, all four sub-checks
- `npm run lint` — 0 errors (the one pre-existing `import/no-named-as-default` warning on `UnsubscriberAsyncList`)
- `npm test` — 4529 passed / 16 skipped core, all 10 workspaces green, 0 failures
- `prettier --check` on the changed files — clean
- `npm run build:platform-bible-react` — succeeds

## Still open, unchanged from the last round

The three I flagged as not mine to decide are still open: the placeholder wording (unreviewed by UX), `EmptyState`'s `id` meaning `data-testid` (worked around via the `domId`/`testId` split; renaming it is its own ticket), and manual verification against a project with XXB/GLO.

katherinejensen00 (approved):
:lgtm:

> @katherinejensen00 reviewed 9 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2792#-:-P1VyznJ525OPqR4mId7:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2794, opened 2026-09-09
Fix the isolated e2e suite's Electron teardown on Windows

rolfheij-sil:
## Summary

The isolated e2e suite (`npm run test:e2e:isolated all`) is unusable on native Windows: the testing team's first run reported 2 failed and ~30 flaky tests, and this branch reproduces that on a Windows 11 machine. The cause is the fixture's teardown, not the tests.

On Windows, Playwright launches Electron through `shell: true`, so `electronApp.process().pid` names a `cmd.exe` wrapper. `teardownElectronApp`'s `process.kill(pid, 'SIGKILL')` stopped that wrapper and left the whole app running — main process, renderer, extension host, and the `dotnet watch` data-provider tree. The orphaned app kept port 8876 bound, so the next test's app logged `PAPI websocket server failed to bind: listen EADDRINUSE ::1:8876`, its renderer connected to the leftover instance instead, every service call failed (`Settings service undefined`), and the test died at the 120 s timeout. The retry then passed, because Playwright's worker restart closed the orphan. Hence fail-once-pass-on-retry for almost every test, one orphaned app and one undeletable temp profile per passing test, and a run that took hours.

### Why review this

Not part of current epic. Prompted by the testing team's first native-Windows run of the isolated suite; until this lands they cannot use it, and the app-side leak (an orphaned instance holding the PAPI port) makes every result on Windows untrustworthy.

## Changes

- `killProcessTree` in `e2e-tests/fixtures/helpers.ts`: `taskkill /pid <pid> /t /f` on Windows, process-group kill with pid fallback elsewhere. Used by `teardownElectronApp`, by `launchElectronApp`'s readiness-failure path, and by `global-teardown.ts`'s dev-server kill (which already did this for its own shell wrapper). `waitForPortFree` now returns whether the port is free, so teardown no longer logs "Port 8876 is free" right after "still in use".
- `dismissStuckFirstRunGate` samples the gate's state in one polling loop instead of racing three `locator.waitFor()` calls with `Promise.race`. The losing legs of that race were never cancelled: they kept polling for up to ~90 s after the function returned, appeared as a failed `escapeHatch.waitFor` step in the trace of passing tests, and rejected unhandled when the page closed under them. The pure poller (`pollFirstRunGate`) has unit tests; `resolveRaceLeg` became `rethrowIfTargetClosed`.
- `global-setup.ts` passes `windowsHide: true` to the detached renderer dev-server spawn, which otherwise opens its own console window on every run.
- Docs: a "Running on Windows" section in `e2e-tests/tests/isolated/README.md` (how to run, what a healthy teardown looks like, recovery when something holds port 8876, leftover temp profiles, locked-session focus failures) and the port-held signature under "Reading a failing run" in `e2e-tests/CLAUDE.md`.

No app code changes.

## AI Involvement

AI-assisted — [session 1](https://claude.ai/code/session_01GZpiRbR7kKvfBM5z1f4A92). Investigation (live process/port inspection and app-log analysis on the Windows machine), fixes, unit tests, and docs were produced with Claude Code; each commit passed the repository's roborev review. To be reviewed by the author before marking ready.

## Testing

- [x] `npx vitest run e2e-tests` — 153 tests pass (new: `pollFirstRunGate`, `killProcessTree`); `npm run typecheck`, `npm run lint`, prettier clean.
- [x] Windows 11, native, full `isolated` run with these fixes: port 8876 free after every teardown, temp profiles deleted; the internet-settings, first-run-wizard and multi-window tests that were "flaky" before pass on their first attempt. Full-run results and the remaining genuine failures will be added here when the run completes.
- [ ] CI smoke on Linux/Windows/macOS (exercises the changed `waitForAppReady` path).

## Risk Level

Low — e2e infrastructure and docs only. The Windows branch of `killProcessTree` reuses the `taskkill /t` approach `global-teardown.ts` already used for the dev server; the POSIX branch is the previous behaviour unchanged.

https://claude.ai/code/session_01GZpiRbR7kKvfBM5z1f4A92

- - -
## Results on Windows 11 (native)

- **Before** (origin/main): the baseline run was stopped at 41 of 90 tests after 65 minutes — 16 passed, 21 flaky, 2 failed, 1 skipped — with `Port 8876 still in use after 15000ms` on every teardown and one orphaned app instance plus one undeletable temp profile per passing test.
- **After the fixture fixes** (first four commits): full run 81 passed, 6 failed, 1 flaky, 2 skipped in 43.5 minutes. Zero port holds, zero `Could not remove` profiles, no processes left behind after the run.
- **The 6 failed + 1 flaky were four test bugs**, three of which reproduce under WSL with identical errors: the first-run page object's `getByRole('dialog')` also matched the onboarding tour; `advanceToLastStep` raced the tour's step transition; the two Simple-mode scripture-editor specs never suppressed the tour; the narrow-title-bar test measured before the flex reflow settled. Fixed in the last two commits; the five affected specs then pass 18/18 on Windows (4.7 min, no retries) and 17/18 under WSL.
- **Still open, not addressed here:**
  - `bcv-open-resource-books-simple` › "offers a book from an open resource, greyed, and navigates to it": on a machine that has the real `WEB` resource, the Column-3 Bible-texts panel stays on `WEB` instead of being re-pointed to the test's own `_resource` project, so "Show more books" has count 1, not 0. Environment-dependent; possibly a re-point gap worth a product-side look.
  - Under WSL only, onboarding-tour › "Help > Show the tour reopens the completed tour at its first stop" times out clicking the Help-menu item (passes on Windows).
- CI on this branch is green on Linux, Windows and macOS, including the smoke e2e steps.
- Run logs, per-run classification reports and archived Playwright artefacts are on the Windows machine under `C:\Users\Paratext\git-win\scratch\runs\`.

This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2794)
## Full runs of the final branch (`98131a92f67`)

- **Windows 11, native**, branch checked out and rebuilt: **87 passed, 1 flaky, 2 skipped, 0 failed in 37.9 min** (before this PR the same machine managed 41 tests in 65 min with 21 flaky). The flaky one was `bcv-open-resource-books-simple`; its first attempt hit a PAPI "event already registered" rejection while a WSL run of the same suite was active on the same machine (WSL2 forwards its ports to Windows' localhost — don't run both at once), and the retry passed.
- **WSL (Xvfb)**: **86 passed, 1 failed, 1 flaky, 2 skipped in 24.8 min.** Failed: `bcv-open-resource-books-simple` (`Show more books` count 1, expected 0). Flaky: onboarding-tour › "Escape key closes the tour" (dialog still visible after 5 s on the first attempt) — Linux only so far.
- Every teardown marker is zero on both OSes (`still in use`, `Could not remove`, `EADDRINUSE`, `Settings service undefined`, `[e2e-first-run-gate-race]`, unhandled rejections, `TargetClosedError`), and no process is left behind after a run.
- The two skips are the pre-existing deliberate `test.skip`s (Alt+Left/Right and F8/F9/Ctrl+B shortcuts, which CDP's `Input.dispatchKeyEvent` cannot drive).
- `bcv-open-resource-books-simple` run alone on Windows afterwards fails both attempts with the `Show more books` count mismatch, so on a machine that has the real `WEB` resource it is flaky-to-failing on both OSes: the Column-3 Bible-texts panel sometimes stays on `WEB` instead of the test's own `_resource` project. Left open here; it needs a look at the re-point logic rather than a locator change.

## Second round: the three remaining failures

Root causes were established before fixing (traces, app code, and reproduction under WSL with `--trace on --repeat-each`).

- **bcv-open-resource-books-simple (both OSes)** — two layers. First, the test's premise only held on machines with no downloaded read-only resource: with an empty `platformScripture.referencedProjectsAndResources`, the Bible-texts panel's picker unions every installed `isEditable === false` project and falls back to the first one (WEB), whose books the toolbar then offers. The test now pins its target project's reference list with a self-reference written into the project's own `Settings.xml` (`setReferencedProjectsAndResources` in `comment-test-helpers.ts`), so its Phase-1 count-0 assertion no longer depends on what is installed. Second, with the pin in place the test still fails: the reference-list setting resolves transiently empty on first render, the panel persists `selectedResourceId = WEB` from that instant, and `resolveResourceSelection` keeps any persisted id that still matches a row — a product-side race that can also strand a real user on the wrong resource. Filed as **PT-4592** (linked to PT-4509 and PT-4336); the test stays red on machines with WEB installed until that lands.
- **onboarding-tour › "Escape key closes the tour" (Linux)** — the tour's Escape listener is only attached while `isCardRendered`, which drops for a frame after every step change while the target is re-measured; the page object's `advanceTour` returned as soon as the Next click resolved, so an Escape could land in that gap. `advanceTour` now waits for the step counter to change (the same mechanism `advanceToLastStep` already used). Not reproduced in 8 traced attempts; derived from the listener gating.
- **onboarding-tour › "Help > Show the tour …" (Linux)** — every failing attempt carried `TypeError: Cannot read properties of null (reading 'getItem')` from a web view, caught by `WebViewErrorBoundary`; the resulting remount tore down the just-opened Help menu. `readDirection()` in `lib/platform-bible-react/src/utils/dir-helper.util.ts` read `localStorage` unguarded and runs during render in most shared components (dropdown-menu, dialog, popover, select, tabs, …), and a detached iframe's window reports `null` storage. The util now tolerates null/throwing storage (unit-tested; the second commit also guards the `getItem`/`setItem` calls themselves, per roborev). **The rebuilt `lib/platform-bible-react/dist` is not in these commits** (committing it is denied for the AI session) — it needs `npm run build:pbr` and a commit by a human before the fix reaches web views. Verified against the rebuilt bundle under WSL: 10 tests × 5 repeats, 0 `getItem` errors, the Help-menu test 5/5. One unrelated flake seen once in that run: the Power-mode sibling timed out clicking a `Show the tour` menu item that was "detached from the DOM" — no crash involved, not investigated further.

### Update

- The rebuilt `lib/platform-bible-react/dist` is now committed (`5fbb5ba7915`), so the `readDirection` guard reaches web views on this branch.
- PT-4592 is parented under PT-3844 "New Simple Workspace" (the epic the recent Simple-mode bugs are filed under) and linked to PT-4509 and PT-4336.

# Code Review Summary

**Branch**: e2e-isolated-windows

**Base**: origin/main

**Date**: 2026-09-10

**Files changed**: 28

## Overview

The isolated Playwright e2e suite (`npm run test:e2e:isolated all`) was unusable on native Windows: the testing team's first run reported 2 failed and ~30 flaky tests. The cause was the fixture's teardown — on Windows Playwright launches Electron through `shell: true`, so the pid the fixture killed was a `cmd.exe` wrapper and the app survived every teardown holding port 8876; the next test's app could not bind, ran against the leftover instance, and died at its 120 s timeout, with the retry passing after Playwright's worker restart closed the orphan. This branch kills the whole process tree (`taskkill /t` on Windows, process group elsewhere), replaces a `Promise.race` of un-cancellable `waitFor`s in the first-run-gate check with a single polling loop, hides the dev-server console, retries the profile cleanup, documents how to run on Windows, and fixes four test bugs the onboarding tour exposed (three reproduce on Linux). It also contains the branch's only product change: `readDirection`/`persistDirection` in `platform-bible-react` now tolerate a null or throwing `localStorage` (a detached iframe's storage is `null`; the util runs during render in most shared components and was crashing web views), with the rebuilt `dist` committed.

Full runs on the final pre-review state: Windows 87 passed / 1 flaky / 2 skipped / 0 failed in 38 min (baseline: 41 tests in 65 min with 21 flaky); WSL 86 / 1 failed / 1 flaky / 2 skipped in 25 min. The remaining `bcv-open-resource-books-simple` failure on machines with the real `WEB` resource is a product race filed as PT-4592 (under PT-3844); the test-side premise pin for it is on this branch.

Four analysis passes (api, style, compliance, ux) plus `/code-review max` found no Critical issues; two Important and 25 Minor items were raised, 27 of which were fixed during the review and 3 dismissed with reasons below.

## API Changes

- `lib/platform-bible-react` (exported via the `platform-bible-react/experimental` entry): `readDirection()` — signature unchanged; now returns `'ltr'` instead of throwing when `localStorage` is `null` (detached iframe), when the `localStorage` property access throws (sandboxed iframe), or when `getItem` throws.
- `lib/platform-bible-react` (same entry): `persistDirection()` — signature unchanged; now a silent no-op instead of throwing when storage is unavailable or `setItem` throws.
- `lib/platform-bible-react/dist/*` regenerated (commit `5fbb5ba7915`) after the last source change to the lib.
- No changes to `lib/papi-dts/papi.d.ts`, `lib/platform-bible-utils/`, or any extension `.d.ts` contract.

None.

- [x] **`advanceTour`/`waitForStepCounterChange` (`onboarding-tour.page.ts`) promised to accept the tour closing on a Done click, but `expect(locator).not.toHaveText()` keeps polling a missing element and times out** _(fixed during review: rewritten as `expect.poll` over "dialog visible ? trimmed counter text : closed", with one `getStepCounter` accessor shared by all callers)_
- [x] **`setReferencedProjectsAndResources` (`comment-test-helpers.ts`) had no unit test for its `"<version> <json>"` envelope, whole-text-node XML escaping, self-id→shortName mapping and guard throws** _(fixed during review: `comment-test-helpers.test.ts`, temp-dir based, six cases including the `$&` replacement-pattern case)_

Author response: both fixed in-review; the second was raised by the compliance pass and the first independently by the api pass and `/code-review`.

### Minor — Consider

- [x] **`readDirectionSafely` docblock in `src/renderer/components/crashed-view.util.ts` described `readDirection` as unguarded, which this branch makes untrue** _(fixed during review: docblock reworded — the wrapper stays as belt-and-braces because the crash screen has no error boundary above it; function unchanged)_
- [ ] ~~`persistDirection` now swallows write failures (`QuotaExceededError`, sandboxed `setItem`) instead of throwing~~ _(intended and documented in its JSDoc; no in-repo caller handled the throw)_
- [x] **`killProcessTree`'s docblock did not say the `signal` argument is ignored on win32 (`taskkill /f` is always forced)** _(fixed during review)_
- [x] **`dir-helper.util.test.ts` never pinned the plain default (`'ltr'` with the key absent or a non-`'rtl'` value)** _(fixed during review: two cases)_
- [ ] ~~The rebuilt `lib/platform-bible-react/dist` diff is larger than the 43-line source change alone would produce and may absorb drift from a `dist` that was already stale on `main`~~ _(not verifiable from the AI session — dist paths are read-denied; the rebuild is what the publishing rule requires, and if main's dist was stale this brings it in sync)_
- [x] **`escapeXmlText` duplicated the unexported `xmlEscapeAttribute` in `helpers.ts`** _(fixed during review: one exported `escapeXml`)_
- [x] **`pollFirstRunGate` took a positional `sleep` and read `Date.now()` directly while the other new seams use a `deps` object** _(fixed during review: `deps: { sleep, now }` with defaults; test injects `now`)_
- [x] **Eight inline `setTimeout` sleeps in `helpers.ts` (two new, six pre-existing)** _(fixed during review: one exported `sleep(ms)`)_
- [x] **`killDevServerProcess` kept a 14-line docblock for a one-line delegate, `global-teardown.test.ts` re-tested branches now covered in `helpers.test.ts`, and the "NOT YET VERIFIED ON WINDOWS" caveat was stale** _(fixed during review: docblock trimmed to call-site specifics; tests reduced to delegation + the EPERM-is-alive case; caveat removed — native Windows runs leave no `node.exe` behind and each following run starts its own dev server)_
- [x] **The "one snapshot, not a race" rationale was written out three times** _(fixed during review: kept in `pollFirstRunGate`'s docblock; pointers elsewhere)_
- [x] **`TargetClosedError` test fixture duplicated across two test files** _(fixed during review: `rethrowIfTargetClosed` tests co-located with `pollFirstRunGate`'s)_
- [x] **`quitAppAndWaitForExit` (`multi-window.util.ts`) still hand-rolled `process.kill(-pid)`, a silent no-op on Windows** _(fixed during review: `killProcessTree`)_
- [x] **`waitForStepCounterChange` re-resolved a locator its callers already held** _(fixed during review, with the Important item above)_
- [x] **Mirrored `'1.1.0'` data-version constant lacked a keep-in-sync note** _(fixed during review: names `CURRENT_DATA_VERSION` and `ResourceReferenceList.CurrentFormatVersion`)_
- [x] **`KillProcessTreeDeps.execFileSync` options typed `unknown`** _(fixed during review: `ExecFileSyncOptions`)_
- [x] **`launchElectronApp`'s two failure paths called bare `fs.rmSync` after the kill; a lingering Windows lock would throw out of the `catch`, replace the real error and skip `restoreAppGlobalState()`** _(fixed during review: best-effort removal inside the catch so restore and rethrow always run)_
- [x] **The post-resize measurement race fixed in one title-bar test also existed in that file's first test and in `paragraph-style-trigger-narrow.spec.ts`** _(fixed during review at `setWindowWidth`: waits two animation frames after `innerWidth` settles so `useShrinkStep`'s ResizeObserver re-render has landed for every caller; the `toPass` guard stays)_
- [x] **`pollFirstRunGate` swallowed every non-`TargetClosedError` sample error silently and kept polling for the caller's whole budget (~90 s), where the replaced race settled `'inconclusive'` at once** _(fixed during review: warns once with the error and returns `'inconclusive'` after three consecutive failed samples; unit tests)_
- [ ] ~~On Windows the teardown gates the kill on the cmd.exe wrapper being alive; if Electron's main process exits abnormally the wrapper is gone and the surviving extension host / .NET tree keeps 8876~~ _(author: out of scope, no ticket — with the wrapper gone the tree is no longer reachable through it; the real fix is a parent-death watch in the extension host and .NET provider, product code)_
- [x] **`windowsHide: true` on the `detached` dev-server spawn is ignored by Win32 (`CREATE_NO_WINDOW` + `DETACHED_PROCESS`); the author confirmed live that a "webpack" console still opened** _(fixed during review: `detached: process.platform !== 'win32'` with `windowsHide` kept — the child gets a hidden console its children inherit, and `taskkill /t` needs no process group; author's re-check on Windows pending)_
- [x] **Inside the new `toPass` block the inner `toBeVisible()` inherited the 10 s `expect.timeout` and could consume the whole outer budget in one attempt** _(fixed during review: 1 s inner timeout)_
- [x] **`settingsXml.replace('</ScriptureText>', string)` expanded `$&`/`$'`/`$$` patterns inside the embedded JSON (latent; same idiom in `removeRevelationFromProject`)** _(fixed during review: replacer functions at both sites; test)_
- [x] **`removeUserDataDirWithRetry`, its seam and ~70 lines of tests re-implemented `fs.rmSync`'s built-in `maxRetries`/`retryDelay`** _(fixed during review: `fs.rmSync(..., { maxRetries: 6, retryDelay: 250 })` in a `try` with the existing warn; helper and tests removed)_
- [x] **`helpers.test.ts` restored `console.warn` spies with bare `mockRestore()` calls after the assertions** _(fixed during review: file-level `afterEach(() => vi.restoreAllMocks())` here and in `first-run-gate-poll.test.ts`)_
- [x] **The `'cleared'` branch after `decideStuckGateAction` in `dismissStuckFirstRunGate` was dead — the poll only returns a sample while the gate is visible** _(fixed during review: branch and comment removed; `decideStuckGateAction` and its tests unchanged)_

Author response: the author asked for all findings as one overview with a recommendation each, chose the complete fix everywhere the change stayed in test/tooling code, allowed the one comment-only edit in an implementation file, and dismissed the three items above with the reasons recorded.

#### Shared Regions Modified

None — no file on the branch carries a `#region shared with` marker.

#### Extension Config Changes

None — no file under `extensions/` changed.

## Positive Observations

- `killProcessTree` collapses three separate copies of the group-kill-with-fallback logic into one function with a testable seam; its docblock explains the Windows `cmd.exe`-wrapper fact a future reader needs.
- `pollFirstRunGate` replaces a race whose losers could reject after the page closed with one sampled snapshot per iteration; its tests pin the `>=` budget boundary (six samples / five sleeps) and the `TargetClosedError` rethrow.
- Every new pure function ships with an injectable `deps` seam and tests that drive it without real timers, processes or filesystem; the timeout tests distinguish `>=` from `>`.
- `dir-helper.util.ts` guards both the property access and the method call; its tests exercise each failure shape separately and restore the `localStorage` descriptor by hand where `vi.unstubAllGlobals` cannot.
- `dist/` was rebuilt in a commit after the last lib source change, so the fix ships.
- New page-object selectors (`first-run-dialog`, `tour-step-counter`) are grounded in pre-existing renderer test ids; the two Simple-mode editor specs now follow the same per-spec `waitForAppReady` convention as the other 21 isolated specs.
- `setReferencedProjectsAndResources` matches the C# reader exactly (`"<version> <json>"` split at the first space, major version 1, `type`/`name`/`id` items) and refuses to overwrite an existing element.
- Documentation was added for both audiences (`tests/isolated/README.md` for testers, `e2e-tests/CLAUDE.md` for agents) with a cross-reference; no ticket/PR ids or change narration were introduced in code comments.

## Interview Notes

- **Stated purpose**: make the isolated e2e suite usable for the testing team on Windows 11 — a working suite first, headless as a nice-to-have.
- **Key decisions**: fix the teardown in the fixture rather than the app (the orphan is a product-level gap too, noted for a possible parent-death watch); keep the `Promise.race` replacement as a poll rather than cancellation tricks; pin the `bcv` test's reference list in the test project's own `Settings.xml` and file the remaining panel race as PT-4592 instead of changing extension code; guard `readDirection` in the library and commit the rebuilt `dist` (the committing session had to stage the lib directory without naming `dist`, which the repo's read-deny rule otherwise blocks).
- **Author preferences recorded**: findings presented as one overview with recommendations; always the correct/complete fix; in test and test-config work, avoid touching implementation code (one comment-only exception granted here).
- **Live verification by the author**: running `npm run test:e2e:isolated title-bar` from PowerShell still showed a "webpack" console before the `detached` change; the re-check after it is pending.
- **Dismissed with reasons**: `persistDirection` swallowing write failures (intended); the size of the `dist` diff (not verifiable from the AI session); the abnormal-exit orphan on Windows (out of scope, no ticket).
- **Understanding**: the root-cause analysis, the fixes and the tests were produced by the AI session under the author's direction; the author reviewed the findings overview and ruled per item but did not author the code. Author does not claim independent understanding of: the Playwright `shell: true` wrapper mechanism behind `killProcessTree`, the `pollFirstRunGate` semantics (including the new warn-once / three-failure cap), and the `readDirection` guard — walk through these in the review meeting.

## In-Review Quality Check

- `npm run typecheck`: pass (core, erb, e2e, all workspaces).
- `npm run lint`: pass, 0 errors (one pre-existing warning in `lib/platform-bible-utils/src/lifetime-management/unsubscriber-async-list.test.ts`, untouched by this branch).
- Prettier on the 20 changed non-dist files: no changes.
- Tests: `test:core` 3724 passed / 1 skipped (263 files, includes the e2e fixture and reporter tests); `lib/platform-bible-react` 1497 passed (163 files, includes the new `dir-helper.util.test.ts`); `lib/platform-bible-utils` 611 passed / 1 skipped; `extensions` has no test script.
- No auto-fixes were needed; no unfixable failures.
- Not yet re-run after the review fixes: the Playwright suites themselves. The in-review changes touch e2e behaviour (`setWindowWidth` settle, `pollFirstRunGate` failure cap, `rmSync` retry, non-detached dev-server spawn, tour page object) and should be validated with a full WSL run and at least the `title-bar`, `first-run` and `onboarding-tour` subsets on Windows before the PR leaves draft.

## Suggested Review Focus

- [ ] `killProcessTree` and the Playwright `shell: true` wrapper on Windows (`e2e-tests/fixtures/helpers.ts`): the load-bearing change — confirm the `taskkill /t` reasoning and the POSIX branch's unchanged behaviour.
- [ ] `readDirection`/`persistDirection` guard in `lib/platform-bible-react/src/utils/dir-helper.util.ts` plus the committed `dist`: the only product behaviour change; `persistDirection` now no-ops on write failure.
- [ ] `pollFirstRunGate` semantics versus the replaced race: budget handling, the warn-once / three-consecutive-failures `'inconclusive'` exit, and whether `'inconclusive'` downstream messaging still names the right cause.
- [ ] `setWindowWidth`'s two-frame settle: is it sufficient for `useShrinkStep`'s ResizeObserver re-render on slower machines, or should the measuring specs keep their own `toPass` guards (they currently do)?
- [ ] PT-4592: `bcv-open-resource-books-simple` stays red on any machine with the real `WEB` resource until the panel race is fixed; the test-side pin (`setReferencedProjectsAndResources`) is deliberate and should stay.
- [ ] Post-review e2e validation results (WSL full run; Windows subsets incl. the "webpack" console re-check) — to be added to the PR before marking it ready.
- [ ] Branch name carries no `pt-` prefix (no ticket for the work itself); rename only via the PR-safe procedure if wanted.

## Post-review validation (tip `85982603885`)

- **Windows 11, native**, checked out and rebuilt at `a04005a4659`: `title-bar` 2/2 from a PowerShell prompt with **no separate "webpack" console** (the `detached`/`windowsHide` fix, verified by the author); `first-run` + `onboarding-tour` 14/15 — the one failure was the review's own tour-poll rewrite reading the tour's one-frame re-measure gap as "closed", fixed in `85982603885`; after that the tour spec passed **30/30** over three repeats.
- **WSL (Xvfb)**, tip: **86 passed / 2 failed / 0 flaky / 2 skipped in 28.5 min**, every environment marker zero (`still in use`, `Could not remove`, `EADDRINUSE`, `Settings service undefined`, first-run-gate race, unhandled rejections, `TargetClosedError`, `getItem` crash). The Linux-only "Escape key closes the tour" flake did not recur.
  - `bcv-open-resource-books-simple`: expected on a machine with the real `WEB` resource — PT-4592.
  - internet-settings › "profile popover shows the Internet & connectivity label…": the action's label was the raw `%userProfile_networkSettings_2%` key and the popover closed while the assertion polled; re-run alone it failed 1 of 2 copies with the same shape. Passed in both full runs earlier the same day on the same base, and no change on this branch touches that path — it is the pre-existing "`%key%` visible = the assertion ran before the localization provider answered" class documented in `e2e-tests/CLAUDE.md` (`waitForAppReady` does not cover localization readiness). Left as an observed flake; a localization-readiness wait in `waitForAppReady` would be the right follow-up.


e2e-tests/fixtures/helpers.ts
```diff
@@ -93,15 +94,26 @@ export interface ElectronAppContext {
   preserveUserDataDir?: boolean;
 }
 
+/**
```
lyonsil, line 97: **#1** - low · checked and confirmed

`sleep`'s docblock promises a single timer seam for this file, but one poll delay still builds its own timer promise.

**What happens:** The new `sleep` docblock at `helpers.ts:97-101` says it is "The one function every inline `setTimeout`-backed delay in this file goes through, so a caller that wants to fake or inject the wait has a single seam". The sweep converted the delays in `waitForPortFree`, `waitForWebSocketReady`, `teardownElectronApp`, `waitForPapiMethodRegistered`, `waitForProjectMetadata` and `waitForMainMenuItem`, but `waitForOpenWebViewIdByType` at `:2567-2569` still builds `new Promise<void>((resolve) => { setTimeout(resolve, sleepMs); })`. A full-file sweep for `setTimeout`, `setInterval` and `new Promise` finds exactly one such site: the other `setTimeout` uses are `sleep`'s own body (`:104`) and three timeout-race probes with reject paths (`:123`, `:163`, `:919`), which are not bare delays.

**Why it matters:** The claim is what a future author will rely on when deciding they can fake time in this module. It is currently false, so a deterministic test written against `sleep` still blocks on a real timer whenever it goes through `waitForOpenWebViewIdByType`. No runtime behaviour differs today.

**Fix:** Replace the inline promise at `helpers.ts:2567-2569` with `await sleep(sleepMs);`. No `no-await-in-loop` comment needs adding or moving here: unlike the per-line `// eslint-disable-next-line no-await-in-loop` comments at the other converted sites, `waitForOpenWebViewIdByType` already disables the rule for its whole loop body with the `/* eslint-disable no-await-in-loop */` / `/* eslint-enable no-await-in-loop */` pair bracketing the `while`, and that block covers this statement. The claim at `:97-101` then holds without amendment.

_How this was checked: The enclosing function is `waitForOpenWebViewIdByType` (helpers.ts:2536-2571), and a byte-diff of it against the merge base shows it is untouched by this PR - the `sleep` helper and its "single seam" docblock at :94-105 are what is new, so the claim became false when the docblock was written rather than because the poll site regressed. A full-file sweep for every timer-construction shape confirms the count exactly: `setTimeout` appears five times - `sleep`'s own definition (:104), three timeout-race probes with `ws.on` reject paths (:123, :163, :919, structurally not bare delays), and exactly one bare `new Promise<void>((resolve) => { setTimeout(resolve, sleepMs); })` at :2567-2569. `new Promise` appears at two further sites (:381 rAF wait, :775 `once('close')`) that use no timer, and `setInterval` has no hits. Pattern caveat: the sweep matches literal `setTimeout`/`new Promise`/`setInterval` and would miss a delay written as `page.waitForTimeout(ms)`, `setImmediate` or `process.nextTick` - none exist in the file today, though the docblock's promise does not rule them out either._

_Other findings in this file: #2, #4, #5, #6, #8_

lyonsil (commented):
Review of this branch's e2e fixture, page-object and spec changes at `aa9dac26bc4`, together with the `dir-helper.util.ts` storage guard.

There are 11 findings, all of them low severity. Eight are attached inline to the lines they concern; the three below sit on lines this PR does not itself change, so GitHub has nowhere to anchor them.

Each comment shows the severity the review settled on and is labelled `checked and confirmed`, meaning the behaviour described was verified against the code rather than inferred from it. The `Fix:` line in each was checked separately — that the symbols it names exist and are reachable from that site, and that the change would survive the repo's own conventions.

## Findings that could not be attached to a line in this diff

**#3** (low · checked and confirmed) e2e-tests/fixtures/helpers.ts:2191 - Two first-run-gate decision tests assert an outcome the shipped code can no longer produce.

**What happens:** `pollFirstRunGate` returns the literal `'cleared'` the moment a sample has `gateVisible === false` (`helpers.ts:2327`) and hands back a `FirstRunGateSample` only when `gateVisible` is true (`:2328`). `dismissStuckFirstRunGate` already special-cases both sentinels (`:2387-2388`) before passing `gateStillShowing: polled.gateVisible` into `decideStuckGateAction` (`:2395`), so that argument is now always `true` and the `if (!gateStillShowing) return 'cleared'` branch at `:2195` is unreachable. The rewrite also dropped the old `if (action === 'cleared') return 'settled'` guard, so `'cleared'` would fall through to the escape-hatch click if it ever did fire. Two cases in `first-run-gate-decision.test.ts` still pin that branch.

**Why it matters:** Those two of seven decision cases can stay green through any change to how `dismissStuckFirstRunGate` derives `gateStillShowing`, including one that reintroduces the inconsistency the field exists to catch. The docblock at `:2191-2194` also still describes the "observations read a round-trip after the gate was seen" race that the single-snapshot design removed.

**Fix:** Drop `gateStillShowing` from `StuckGateObservations` (`helpers.ts:2166`), delete the `:2195` branch, narrow `decideStuckGateAction`'s return type to `'recoverable' | 'wizard' | 'inconclusive'`, drop the field at the call site (`:2395`), and delete the two `'cleared'` cases plus the now-invalid field from the remaining cases in `first-run-gate-decision.test.ts`. Rewrite the `:2191-2194` comment to describe the single-snapshot contract. The alternative - keeping the field and restoring `if (action === 'cleared') return 'settled'` - also works, but a second cleared-check is redundant once `pollFirstRunGate` owns that decision.

_How this was checked: `pollFirstRunGate` returns the sentinel `'cleared'` the instant a sample has `gateVisible === false` (helpers.ts:2327) and returns the `FirstRunGateSample` object only when `gateVisible` is true (:2328), so any sample it hands back structurally carries `gateVisible: true`. `dismissStuckFirstRunGate` already special-cases both sentinels (:2387-2388), so `gateStillShowing` (:2395) can only ever be `true` and the `if (!gateStillShowing) return 'cleared'` branch at :2195 is dead. `decideStuckGateAction` is exported, but a whole-repo grep finds no second production caller - only the unit test. Mutating :2195 to `if (false)` and running `npx vitest run e2e-tests/fixtures/first-run-gate-decision.test.ts` turns exactly the two `'cleared'` cases red and leaves the other five green. The pre-PR code read `firstRunDialog.isVisible()` fresh after the race, where `gateStillShowing` really could be false, and carried an `if (action === 'cleared') return 'settled'` guard that the rewrite dropped; `first-run-gate-decision.test.ts` is unchanged between merge base and head, so these are pre-existing tests the rewrite made non-falsifiable._

_Not attached to a line: no longer on a changed line; cannot be posted inline_

**#7** (low · checked and confirmed) e2e-tests/fixtures/helpers.ts:2355 - The first-run-gate check's escape-hatch locator searches the whole page rather than the gate, so a second button with a matching name anywhere on screen breaks every app-readiness wait.

**What happens:** `dismissStuckFirstRunGate` scopes `dialogHeading` and `errorScreen` to the gate (`helpers.ts:2366-2367`) but builds `escapeHatch` as `page.getByRole('button', { name: ESCAPE_HATCH_NAME_PATTERN })` (`:2355-2357`), page-scoped. That pattern matches wording, not an id. `Locator.isVisible()` hardcodes `{ strict: true, ... }` (`playwright-core/lib/client/locator.js:235-236`), so a two-element match throws - and that read sits inside the sampler's `Promise.all` (`:2373-2378`), making every sample fail. `escapeHatch.click()` at `:2426`, the recovery action itself, has the same exposure.

**Why it matters:** Nothing in the shipped UI renders a second matching button today, so this is latent rather than live, and the page-scoped declaration is pre-existing - it sits at `:2198` at the merge base, and this PR's hunk for this function begins below it, replacing the polling algorithm only. It is worth fixing while the surrounding code is open rather than treated as a regression.

**Fix:** Build `escapeHatch` from `firstRunDialog` rather than `page` at `helpers.ts:2355` - `const escapeHatch = firstRunDialog.getByRole('button', { name: ESCAPE_HATCH_NAME_PATTERN });` - matching how `dialogHeading` and `errorScreen` are already scoped. All three escape-hatch render sites are true DOM descendants of `[data-testid="first-run-dialog"]`, and `:2426` reuses the same binding, so the one edit covers both reads and the click.

_Not attached to a line: no longer on a changed line; cannot be posted inline_

**#10** (low · checked and confirmed) e2e-tests/fixtures/onboarding-tour.page.ts:184 - The tour page object waits for the step transition after Next and Done, but not after Back.

**What happens:** `advanceTour` (`onboarding-tour.page.ts:177`) and `advanceToLastStep` (`:195`) both end in `waitForStepCounterChange`, so they return only once the step transition has rendered. `goBackTour` (`:185-187`) is still a bare `click()`. The transition it triggers is the same one: `handleBack` calls `setStepIndex((i) => i - 1)` (`tour.component.tsx:416-418`), which leaves `targetRect` stale until the new step's target is re-measured, and the guard at `:475` returns `undefined` for that frame - unmounting the whole `tour-dialog`, not just the card.

**Why it matters:** Today's only caller (`onboarding-tour.spec.ts:173-174`) is not exposed: it reads `getCurrentStepTitle`, whose `textContent()` auto-waits for re-attachment and so can only observe "not yet attached" or "attached with the new title". The exposure is the next caller that checks state after Back with a non-auto-waiting read - `isVisible()` or `count()` - which is exactly the shape that broke `advanceToLastStep` and prompted the fix at the other two sites.

**Fix:** In `goBackTour`, mirror `advanceTour`: read `getStepCounter(page).textContent()` into `stepBeforeClick` before `getTourBackButton(page).click()`, then `await waitForStepCounterChange(page, stepBeforeClick)`. Both helpers are module-private but declared in the same file, and the types line up as `advanceTour` already uses them. Update the docblock to say it waits for the transition, as `advanceTour`'s now does.

_Not attached to a line: no longer on a changed line; cannot be posted inline_

e2e-tests/fixtures/helpers.ts
```diff
@@ -361,6 +370,18 @@ export async function setWindowWidth(
       timeout: WINDOW_WIDTH_SETTLE_TIMEOUT_MS,
     })
     .toBeLessThanOrEqual(WINDOW_WIDTH_SETTLE_TOLERANCE_PX);
+
+  // `window.innerWidth` settling is not the toolbar re-rendering: its `useShrinkStep`
+  // ResizeObserver callback runs a frame later, so a caller that measures right after the poll
+  // above can still catch a pre-shrink `scrollWidth` against a post-shrink `clientWidth`. Two rAFs
+  // guarantee that callback (which runs before the next frame's paint) has landed before this
+  // returns.
+  await page.evaluate(
```
lyonsil, line 379: **#2** - low · checked and confirmed

A title-bar or narrow-editor test whose Electron window is minimised or occluded hangs at the resize step instead of failing.

**What happens:** `setWindowWidth` now ends with a `page.evaluate` that resolves only after two `requestAnimationFrame` callbacks fire (`helpers.ts:379-384`). Chromium's `backgroundThrottling` - on by default, and never overridden in this repo's window creation (`src/main/main.ts:905`) - stops rAF entirely for a hidden or occluded window. `page.evaluate` sends no timeout field (`playwright-core/lib/client/frame.js:175-179`), unlike every actionability call beside it, and `playwright.config.ts` sets no `actionTimeout`, so the promise is bounded only by the 120 s test timeout (`playwright.config.ts:28`). Every earlier step in the helper carries its own budget; this one does not.

**Why it matters:** It only bites when the app window loses visibility mid-run - which the new `tests/isolated/README.md:52-61` explicitly warns about for native Windows runs. When it does, `title-bar-narrow-width.spec.ts` and `paragraph-style-trigger-narrow.spec.ts` stall for two minutes with nothing pointing at the resize helper. The repo already handles this exact hazard for the same double-rAF idiom in `waitForNextPaint` (`src/renderer/services/web-view.service-shard.ts:1973`).

**Fix:** Add a constant beside the other two at `helpers.ts:~330` - `const WINDOW_WIDTH_RAF_SETTLE_TIMEOUT_MS = 2_000;` - then bound the wait and swallow the abandoned evaluate: `await Promise.race([page.evaluate(...).catch(() => {}), sleep(WINDOW_WIDTH_RAF_SETTLE_TIMEOUT_MS)]);`. The `.catch` is required, not optional: a hidden window can stay hidden until teardown closes the page, and the abandoned call would otherwise reject unhandled - the same shape this PR removes elsewhere. Keep the existing two-rAF comment and add one line saying the wait is bounded because a hidden window suspends rAF indefinitely, citing `waitForNextPaint`. Do not reuse the name `WINDOW_WIDTH_SETTLE_TOLERANCE_PX`, which is an unrelated pixel tolerance.

_How this was checked: The final `page.evaluate` in `setWindowWidth` (helpers.ts:379-384) is unbounded. Playwright's client-side `Frame.evaluate` (`playwright-core/lib/client/frame.js:175-179`) sends no `timeout` field, unlike every actionability call in the same file; server-side `Dispatcher._handleCommand` only arms a deadline when `validParams?.timeout` is set (`dispatcher.js:91`, `progress.js:56-90`), so the wait is bounded only by the 120s test timeout in `playwright.config.ts:28` (no `actionTimeout` is set). Chromium's `backgroundThrottling` - which stops rAF while a window is hidden or occluded - is on by default and this repo never overrides it (`src/main/main.ts:905` sets no such `webPreferences`). The repo already documents this exact hazard for the same double-rAF idiom at `src/renderer/services/web-view.service-shard.ts:1973` ("this alone would hang forever"). `setWindowWidth` is reached from `title-bar-narrow-width.spec.ts` and `paragraph-style-trigger-narrow.spec.ts`._

_Other findings in this file: #1, #4, #5, #6, #8_

e2e-tests/fixtures/helpers.ts
```diff
@@ -2167,14 +2247,91 @@ type FirstRunGateOutcome = 'settled' | 'inconclusive';
...
+ * @param timeout Budget for the whole poll.
+ * @param deps Sleep/clock seams, injected so tests can drive the loop without waiting in real time;
+ *   defaults to real timers.
+ * @returns `'cleared'` once the gate is gone; the sample itself once it shows a recognisable stuck
+ *   state (an escape hatch or a heading) — the caller needs that exact snapshot, since it is
+ *   already the single consistent observation {@link decideStuckGateAction} requires, not a cue to
+ *   read again; `'inconclusive'` if neither happened within the budget, OR if 3 samples in a row
+ *   fail to read anything (a non-retriable failure, e.g. a selector that will never match, should
+ *   not burn the whole ~90s budget silently — the first such failure is logged once, with a stable
+ *   `[e2e-first-run-gate]` prefix, and the streak resets on the next successful sample).
```
lyonsil, line 2286: **#4** - low · checked and confirmed

Hovering `pollFirstRunGate` shows no documentation, while the `PollFirstRunGateDeps` interface shows a docblock describing a function's parameters and return value.

**What happens:** The 29-line TSDoc block opening "Poll a first-run-gate sample until it clears...", carrying `@param sample`, `@param timeout`, `@param deps` and `@returns`, runs from `helpers.ts:2259` to `:2287`. The declaration immediately below it at `:2288` is `export interface PollFirstRunGateDeps`, not the function. `export async function pollFirstRunGate` is ten lines further down at `:2298`, separated by `defaultPollFirstRunGateDeps` (`:2293`) and `MAX_CONSECUTIVE_SAMPLE_FAILURES` (`:2296`), and carries no doc of its own.

**Why it matters:** Everything the block says - the "not a race of several `waitFor` calls" rationale, the `@param sample` contract that the sampler must not itself time out, and the three return shapes - is the load-bearing explanation for the central helper this PR introduces, and none of it reaches anyone reading or hovering `pollFirstRunGate`. The PR's other new seam gets this right: `KillProcessTreeDeps` has its own one-line doc at `:583` and `killProcessTree`'s full doc sits immediately above the function at `:617`.

**Fix:** Move the docblock from `helpers.ts:2259-2287` to sit immediately above `export async function pollFirstRunGate` at `:2298`, keeping `@param`/`@returns` on the function per the Code-Style-Guide's API-surface TSDoc convention, and give `PollFirstRunGateDeps` its own one-line docblock describing the seam, mirroring `KillProcessTreeDeps` at `:583`.

_How this was checked: helpers.ts:2259-2287 is the 29-line TSDoc block opening "Poll a first-run-gate sample until it clears...", carrying `@param sample`, `@param timeout`, `@param deps` and `@returns`. It sits directly above `export interface PollFirstRunGateDeps` at :2288, not above `export async function pollFirstRunGate` at :2298 - `defaultPollFirstRunGateDeps` (:2293) and `MAX_CONSECUTIVE_SAMPLE_FAILURES` (:2296) sit between them. `pollFirstRunGate` carries no doc of its own, so hovering it shows nothing. `git diff` against the merge base confirms the whole block is new in this PR. The PR's other new seam gets it right: `KillProcessTreeDeps` has its own one-line doc at :583 and `killProcessTree`'s full doc sits immediately above the function at :617._

_Other findings in this file: #1, #2, #5, #6, #8_

e2e-tests/fixtures/helpers.ts
```diff
@@ -2167,14 +2247,91 @@ type FirstRunGateOutcome = 'settled' | 'inconclusive';
...
+      // Sequential polling: each sample must finish before the next is taken.
+      // eslint-disable-next-line no-await-in-loop
+      current = await sample();
+    } catch (error) {
+      rethrowIfTargetClosed(error);
+      // Anything else means nothing recognisable came back this iteration. Keep polling — most
+      // reads like this are ordinary and quiet — but log the first one (once, not every 100ms) and
+      // give up after a short streak of them, so a failure that will never clear on its own does
+      // not silently claim the whole remaining budget.
+      if (!hasWarnedOnSampleFailure) {
```
lyonsil, line 2318: **#5** - low · checked and confirmed

When the first-run-gate check gives up after repeated read failures, the log names an earlier, unrelated failure rather than the one that caused it.

**What happens:** `pollFirstRunGate` sets `hasWarnedOnSampleFailure = true` on the first swallowed sample error (`helpers.ts:2318-2321`) and never resets it, while `consecutiveSampleFailures` IS reset on the next successful sample (`:2326`). In a poll that can run ~90 s, one early transient read failure consumes the single warning. A later, different, persistent streak then reaches `MAX_CONSECUTIVE_SAMPLE_FAILURES` and returns `'inconclusive'` (`:2323`) with nothing logged about it, leaving the only `[e2e-first-run-gate]` line in the run naming the earlier, unrelated error.

**Why it matters:** The warning exists so an operator can tell a slow first run from a read that will never succeed, and it points them at the wrong error. The trigger is narrow - it needs one early throw from `isVisible()`/`count()` and later a different persistent throw in the same poll, plausible from mid-poll context teardown or a broken selector but not from an ordinary non-match, which returns false/0 rather than throwing.

**Fix:** Log the error that triggers the early exit at the `consecutiveSampleFailures >= MAX_CONSECUTIVE_SAMPLE_FAILURES` return (`helpers.ts:2323`), keeping the `[e2e-first-run-gate]` prefix, rather than relying on the once-per-call warning at `:2318-2321` to have named it. `error` is already in scope for the whole `catch`. Add a case to `first-run-gate-poll.test.ts` that fails once, succeeds, then fails three times, asserting the logged text names the SECOND error.

_How this was checked: helpers.ts:2318-2321 sets `hasWarnedOnSampleFailure = true` on the first swallowed sample error and never resets it, while `consecutiveSampleFailures` IS reset to 0 on the next successful sample (:2326). The whole poll (`pollFirstRunGate`, :2298-2331) is new in this PR - the pre-PR code was a `Promise.race` of three `waitFor`s with no sample-retry logic. So an early transient failure logs once and sets the flag, and a later unrelated streak that reaches `MAX_CONSECUTIVE_SAMPLE_FAILURES` returns `'inconclusive'` (:2323) with no log at all. Both clauses of the docblock's own claim (:2287-2289) are literally true, so this is not the code deviating from documented intent - it is a gap the docblock is silent on. Reachability is narrow: it needs one early throw from `isVisible()`/`count()` and later a different persistent throw in the same poll, plausible from mid-poll context teardown or a genuinely broken selector but not from an ordinary non-match, which returns false/0 rather than throwing (`playwright-core/lib/server/frames.js`, `isVisibleInternal`)._

Similar fix as #6.

_Other findings in this file: #1, #2, #4, #6, #8_

e2e-tests/fixtures/helpers.ts
```diff
@@ -2167,14 +2247,91 @@ type FirstRunGateOutcome = 'settled' | 'inconclusive';
...
+        console.warn(`[e2e-first-run-gate] Sample read failed, will keep polling: ${error}`);
+        hasWarnedOnSampleFailure = true;
+      }
+      consecutiveSampleFailures += 1;
```
lyonsil, line 2322: **#6** - low · checked and confirmed

An app-readiness failure caused by an unreadable locator is reported as "the first-run gate was still showing something unrecognised", which is not what happened.

**What happens:** `pollFirstRunGate` now has a second way to return `'inconclusive'` - three consecutive sample-read failures (`helpers.ts:2322-2323`) - alongside the pre-existing budget-exhausted exit. `dismissStuckFirstRunGate` forwards it unchanged (`:2388`), and `FirstRunGateOutcome` (`:2222`) is only `'settled' | 'inconclusive'`, so nothing downstream can tell the two causes apart. `waitForAppReady` then rewrites a later `waitForOverlayGone` failure through `describeInconclusiveOverlayTimeout` (`:2520`), whose message positively asserts the gate was "still showing something unrecognised, or an error screen with no escape hatch yet" - false on the new path, where the poll may never have read the gate at all.

**Why it matters:** That message exists specifically to stop an operator chasing the wrong cause; on the new path it sends them to the first-run gate when the real fault was an unreadable locator. It needs a locator throw rather than an ordinary non-match, so it is rare rather than routine.

**Fix:** Give `pollFirstRunGate`'s sample-failure give-up a distinct return value carrying the error - change its signature to `Promise<'cleared' | 'inconclusive' | { sampleReadFailure: unknown } | FirstRunGateSample>` and return `{ sampleReadFailure: error }` at `:2323`. Thread it through `FirstRunGateOutcome` (`:2222`) and `dismissStuckFirstRunGate`'s forwarding (`:2388`); give `describeInconclusiveOverlayTimeout` (`:2452`) an optional `sampleReadFailure?: unknown` parameter with its own branch, so the existing caller keeps compiling; branch on the new shape in `waitForAppReady` (`:2520`). Two companion edits are required: update the existing "gives up as inconclusive after 3 consecutive sample failures" test in `first-run-gate-poll.test.ts`, whose `expect(result).toBe('inconclusive')` must assert the new shape, and update `pollFirstRunGate`'s `@returns` TSDoc.

_How this was checked: The chain holds end to end. `pollFirstRunGate`'s sample-failure give-up returns `'inconclusive'` (helpers.ts:2323, new in this PR); `dismissStuckFirstRunGate` forwards it unchanged at :2388 via `if (polled === 'inconclusive') return 'inconclusive';`; `FirstRunGateOutcome` (:2222) is only `'settled' | 'inconclusive'`, so nothing downstream can tell the two causes apart; and `waitForAppReady` (:2516-2521) does `throw gateOutcome === 'inconclusive' ? describeInconclusiveOverlayTimeout(error) : error;` at :2520. `describeInconclusiveOverlayTimeout` (:2452-2461) and its message are pre-existing and untouched, so this is new code invalidating an assumption baked into old prose - "still showing something unrecognised, or an error screen with no escape hatch yet" is simply false when the poll never got a usable read. Same narrow trigger as the warning gap: a locator throw, not an ordinary non-match._

Similar fix as #5.

_Other findings in this file: #1, #2, #4, #5, #8_

e2e-tests/fixtures/helpers.ts
```diff
@@ -2209,51 +2366,34 @@ async function dismissStuckFirstRunGate(page: Page, timeout: number): Promise<Fi
...
-    // Read concurrently, not one after another: a gate that resolves between two sequential awaits
-    // would leave them describing two different instants, which is exactly the inconsistent
-    // combination decideStuckGateAction's own precedence (gateStillShowing decides first) exists to
-    // rule out — but only if the three readings are actually a single snapshot in time.
-    const [escapeHatchVisible, errorScreenCount, gateStillShowing] = await Promise.all([
+  // See pollFirstRunGate's own docblock for why this polls one snapshot per iteration rather than
+  // racing the three `waitFor`s against each other. Local to this call: `isVisible`/`count` never
+  // auto-wait or time out, so the sample itself needs no timeout of its own.
+  const polled = await pollFirstRunGate(async () => {
+    const [gateVisible, escapeHatchVisible, headingVisible, errorScreenCount] = await Promise.all([
```
lyonsil, line 2373: **#8** - low · checked and confirmed

A first-run gate that has in fact cleared can be reported as inconclusive whenever one of the other three reads fails, and a later unrelated timeout is then blamed on the gate.

**What happens:** The sampler reads all four discriminators through one `Promise.all` (`helpers.ts:2373-2378`), so any one rejecting discards `gateVisible` with it. `pollFirstRunGate` counts that as a sample failure and, after `MAX_CONSECUTIVE_SAMPLE_FAILURES` (`:2296`), returns `'inconclusive'` (`:2323`). Under the `Promise.race` this replaced, the `firstRunDialog.waitFor({ state: 'hidden' })` leg was independent and retried internally through this class of transient, so it would still have settled `'cleared'`.

**Why it matters:** `waitForAppReady` uses `'inconclusive'` to rewrite a later `waitForOverlayGone` failure through `describeInconclusiveOverlayTimeout` (`:2520`), so a gate that had cleared gets an unrelated overlay timeout blamed on it. `rethrowIfTargetClosed` runs before the streak counter, so a closed page is correctly excluded; what remains is a non-fatal transient such as a fleeting strict-mode violation on the page-wide `escapeHatch` locator, which needs ~300 ms of repetition to bite.

**Fix:** Keep all four reads in one concurrent batch - swap `Promise.all` for `Promise.allSettled` rather than splitting them into sequential reads, preserving the single-instant snapshot `FirstRunGateSample` requires. Rethrow immediately if any rejection reason is a `TargetClosedError` (tested by `error.constructor.name`, matching `rethrowIfTargetClosed`); rethrow if `gateVisible`'s own promise rejected, so the streak-counting behaviour is unchanged when the core signal is unreadable; return the all-false cleared shape if `gateVisible` fulfilled `false`, whose siblings are moot; otherwise default any individually-rejected discriminator to `false`. Extract this into an exported `buildFirstRunGateSample(locators)` so `first-run-gate-poll.test.ts` can add two mutation-provable cases - "gate gone, one discriminator throws" and "gate visible, one discriminator throws" - since that file drives the loop with an injected mock and cannot reach the current inline closure.

_How this was checked: The sampler joins `firstRunDialog.isVisible()`, `escapeHatch.isVisible()`, `dialogHeading.isVisible()` and `errorScreen.count()` in one `Promise.all`, so any single rejection discards a `gateVisible: false` result with it; after three consecutive failures `pollFirstRunGate` returns `'inconclusive'`, which `waitForAppReady` uses to relabel a later `waitForOverlayGone` timeout through `describeInconclusiveOverlayTimeout`. `rethrowIfTargetClosed` runs before the streak counter, so a closed page/context/browser is correctly excluded from this path. What stays reachable is a non-fatal, non-TargetClosedError transient: `isVisible()` enforces strict mode (`playwright-core/lib/client/locator.js:235-236`), so a fleeting double match on the page-wide `escapeHatch` locator during a transition throws, where `count()` never enforces strict mode and would not. At the merge base the three legs were raced independently, and Playwright's `.waitFor()` retries internally through this class of transient rather than rejecting, so the `firstRunDialog.waitFor({state:'hidden'})` leg would have kept polling and settled `'cleared'` regardless of a sibling's hiccup._

_Other findings in this file: #1, #2, #4, #5, #6_

e2e-tests/fixtures/onboarding-tour.page.ts
```diff
@@ -113,11 +118,67 @@ export async function getCurrentStepTitle(page: Page): Promise<string> {
...
+        const [dialogVisible, counterCount] = await Promise.all([
+          dialog.isVisible(),
+          stepCounter.count(),
+        ]);
+        if (!dialogVisible) {
+          dialogAbsentSamples += 1;
+          return dialogAbsentSamples >= samplesBeforeClosed ? 'closed' : previousTrimmed;
+        }
+        dialogAbsentSamples = 0;
+        if (counterCount === 0) return previousTrimmed;
```
lyonsil, line 162: **#9** - low · checked and confirmed

Clicking Done on the tour's last step can fail the step-transition wait with a timeout instead of detecting that the tour closed.

**What happens:** `waitForStepCounterChange`'s poll body reads `dialog.isVisible()` and `stepCounter.count()` together (`onboarding-tour.page.ts:154-155`), then makes a second, separate round trip for `stepCounter.textContent()` (`:163`). Between them the Done click's unmount can land: `Tour` returns `undefined` once `targetRect` clears (`tour.component.tsx:475`), removing `tour-step-counter` from the DOM. `textContent()` then auto-waits for an element that will never come back, and with no `actionTimeout` in `playwright.config.ts` that read is itself unbounded.

**Why it matters:** `expect.poll`'s own `{ timeout: 5_000 }` (`:166`) does bound the test - `raceAgainstDeadline` does not wait for a hung callback - so the outcome is a 5-second poll timeout on the last-step Done click rather than a stall, absorbed on the automatic retry. Per `e2e-tests/CLAUDE.md`, fail-then-pass is a defect, and this one reports as a counter-change timeout rather than naming the closed tour it was meant to detect.

**Fix:** In `waitForStepCounterChange`, replace the `count()` + `textContent()` pair with a single non-waiting read - `const text = await stepCounter.evaluateAll((els) => els[0]?.textContent ?? undefined)` - treat `undefined` exactly as the current `counterCount === 0` branch does, and drop `stepCounter.count()` from the `Promise.all`. To make it verifiable, extract the poll callback into a file-local named function taking `dialog` and `stepCounter` typed as the minimal shapes it uses, and add `onboarding-tour.page.test.ts` beside it under `fixtures/` with a fake whose `evaluateAll` resolves `undefined`, asserting the sampler returns `previousTrimmed`; pair it with a fake `textContent()` that never resolves to show the old shape stalls where the new one does not.

e2e-tests/global-setup.ts
```diff
@@ -262,11 +262,21 @@ export default async function globalSetup(_config: FullConfig): Promise<void> {
...
+      // POSIX: create a new process group so global-teardown can kill the entire tree via
+      // process.kill(-pid) — without this, the shell child inherits the parent's PGID and
+      // process.kill(-pid) throws ESRCH. Windows has no addressable process group for `detached` to
+      // create, so global-teardown reaches the tree there via `taskkill /t` instead (see
+      // killProcessTree), and `detached` itself is left off: combined with `windowsHide` below, node
+      // requests DETACHED_PROCESS on Windows, and Windows silently ignores CREATE_NO_WINDOW whenever
+      // DETACHED_PROCESS is also set — the two flags this spawn asks for lose to each other,
+      // reopening the visible console window `windowsHide` exists to suppress. Splitting them by
+      // platform keeps both properties: a killable tree on POSIX, no console window on Windows.
+      detached: process.platform !== 'win32',
```
lyonsil, line 274: **#11** - low · checked and confirmed

The platform split that keeps the Windows dev-server console hidden can be reverted to a single `detached: true` with every test still green.

**What happens:** `global-setup.ts:274` spawns the renderer dev server with `detached: process.platform !== 'win32'` and `windowsHide: true` (`:279`), a conditional holding two properties at once: a process group POSIX teardown can signal, and no console window on Windows. Both are inline literals in the `spawn` call, so nothing can reach them from a test - `global-setup.test.ts` covers only `markDevServerLogRunBoundary` and `clearInheritedPidFile`. Every other new platform conditional in this PR got an injectable seam plus tests: `killProcessTree`'s `platform` parameter (`helpers.test.ts:124-191`) and `pollFirstRunGate`'s `deps`.

**Why it matters:** The reasoning lives only in a comment, and the premise is real - libuv's `src/win/process.c` unconditionally ORs `DETACHED_PROCESS | CREATE_NEW_PROCESS_GROUP` for `UV_PROCESS_DETACHED`, and Windows ignores `CREATE_NO_WINDOW` when combined with `DETACHED_PROCESS`. A later tidy-up restoring `detached: true`, which reads as the more obvious value, leaves the whole suite green and reappears only as a stray console window on a Windows developer's desktop.

**Fix:** Extract the platform-dependent options into an exported pure function in `e2e-tests/global-setup.ts` - `export function devServerSpawnFlags(platform: NodeJS.Platform): { detached: boolean; windowsHide: boolean }` - spread its result into the `spawn` call, and add two cases to `global-setup.test.ts` pinning `win32` to `{ detached: false, windowsHide: true }` and `linux` to `{ detached: true, windowsHide: true }`. Move the two comments onto the function so the reasoning sits with the thing under test.

_How this was checked: The lines are new in this PR - `git diff` on e2e-tests/global-setup.ts shows the old single `detached: true` becoming the platform split at :274/:279. A grep confirms no test anywhere reaches `detached`/`windowsHide` for this spawn; `global-setup.test.ts` covers only `markDevServerLogRunBoundary` and `clearInheritedPidFile`. Reverting the two lines to a single `detached: true` left `npx tsc --noEmit -p e2e-tests/tsconfig.json`, `npx eslint e2e-tests/global-setup.ts` and `npx vitest run e2e-tests` (156/156) all green, so the mechanics are exactly as claimed. The factual premise behind the comment also checks out: libuv's `src/win/process.c` unconditionally ORs `DETACHED_PROCESS | CREATE_NEW_PROCESS_GROUP` into `process_flags` for `UV_PROCESS_DETACHED`, and Windows documents `CREATE_NO_WINDOW` as ignored when used with `DETACHED_PROCESS`. One other platform conditional in the same file (:213) is also untested by the same convention, but this PR set its own precedent of giving every other new platform branch a seam plus tests._

e2e-tests/fixtures/helpers.ts
rolfheij-sil: 🤖 Claude: Confirmed — `waitForOpenWebViewIdByType` was the one bare-delay site the `sleep` docblock's claim didn't cover.

Fixed in `0729ae3e49c`: the inline `new Promise<void>((resolve) => { setTimeout(resolve, sleepMs); })` at that call site now calls `await sleep(sleepMs)` instead. No eslint-disable changes were needed — the enclosing loop's existing `/* eslint-disable no-await-in-loop */`/`/* eslint-enable */` pair already brackets this statement, exactly as you noted.

e2e-tests/fixtures/helpers.ts
rolfheij-sil: 🤖 Claude: Confirmed — the post-resize double-rAF wait in `setWindowWidth` was unbounded, and a hidden/occluded window's suspended `requestAnimationFrame` would hang it for the full 120s test timeout instead of failing fast.

Fixed in `2710c9c586b`: added `WINDOW_WIDTH_RAF_SETTLE_TIMEOUT_MS = 2_000` next to the existing `WINDOW_WIDTH_SETTLE_*` constants, and wrapped the `page.evaluate(...)` call in `Promise.race([page.evaluate(...).catch(() => {}), sleep(WINDOW_WIDTH_RAF_SETTLE_TIMEOUT_MS)])`, matching `waitForNextPaint`'s pattern as you suggested. The `.catch` is in place for exactly the reason you gave — the abandoned evaluate can otherwise reject unhandled once the race resolves via the sleep.

One follow-on: bounding this wait meant `paragraph-style-trigger-narrow.spec.ts`'s per-width geometry read (which has no retry) could now read pre-reflow layout if the timeout side of the race wins. Caught it in self-review and wrapped that read in the same `toPass` retry the sibling test in `title-bar-narrow-width.spec.ts` already uses — see `772f54181ac`.

rolfheij-sil:
🤖 Claude: Confirmed — `pollFirstRunGate` only ever returns a sample with `gateVisible: true`, so `gateStillShowing` was always true and `decideStuckGateAction`'s `if (!gateStillShowing) return 'cleared'` branch was dead, along with the two `first-run-gate-decision.test.ts` cases pinning it.

Fixed in `d57e993a475`: dropped `gateStillShowing` from `StuckGateObservations`, deleted the dead branch, narrowed `decideStuckGateAction`'s return type to `'recoverable' | 'wizard' | 'inconclusive'`, dropped the field at the `dismissStuckFirstRunGate` call site, deleted the two invalid test cases, and rewrote the stale round-trip-race comment to describe the actual single-snapshot contract.

(Finding #3 in your review — posting here since GitHub had nowhere to anchor an inline reply.)

e2e-tests/fixtures/helpers.ts
rolfheij-sil: 🤖 Claude: Confirmed — the 29-line TSDoc block sat above `PollFirstRunGateDeps` rather than above `pollFirstRunGate` itself.

Fixed in `a6517341d01`: moved the docblock to sit directly above `pollFirstRunGate` (now at `helpers.ts:2236`, function at `:2266`), and gave `PollFirstRunGateDeps` its own one-line doc mirroring `KillProcessTreeDeps`'s (`helpers.ts:604`).

While reading that region I found the same problem one level up, worse: `dismissStuckFirstRunGate` had no docblock of its own at all — a 65-line block that actually describes it (the "deliberately loud when it fires" language) was sitting six declarations away, separated by `StuckGateObservations`, `decideStuckGateAction`, a selector constant, `FirstRunGateOutcome`/`FirstRunGateSample`, `rethrowIfTargetClosed`, and all of `pollFirstRunGate`. Moved that one too, in the same commit, so `dismissStuckFirstRunGate` (`helpers.ts:2447`) now has it directly above.

e2e-tests/fixtures/helpers.ts
rolfheij-sil: 🤖 Claude: Confirmed — `hasWarnedOnSampleFailure` latched on the first swallowed sample error and never reset, while `consecutiveSampleFailures` did reset on the next success, so a later unrelated failure streak that actually triggered the give-up logged nothing new.

Fixed together with finding #6 in `4fa1e2ce0a5`, since both land on the same give-up return statement: the give-up at `helpers.ts:2299` now logs the triggering error directly (`[e2e-first-run-gate] Giving up after...`, `:2296-2299`) rather than relying on the once-per-call warning to have named it, and separately carries that error structurally — see #6's reply for the type-level change. Updated the existing "warning only once" test in `first-run-gate-poll.test.ts` (the fix means it now logs twice — once on the first transient failure, once at give-up — so the old "only once" framing no longer held and the test's name/assertion changed with it) and added a new case: fail once with an unrelated error, succeed, then fail three times with a different persistent error, asserting the SECOND error is what's reported.

e2e-tests/fixtures/helpers.ts
rolfheij-sil: 🤖 Claude: Confirmed — the sample-read-failure give-up and the budget-exhausted give-up both returned the bare string `'inconclusive'`, so `waitForAppReady` couldn't tell them apart and `describeInconclusiveOverlayTimeout`'s message would falsely claim the gate was seen "still showing something unrecognised" on a path where the poll may never have gotten a usable read.

Fixed in `4fa1e2ce0a5` (combined with #5, since both touch the same give-up site): added a `SampleReadFailure` interface (`{ sampleReadFailure: unknown }`, `helpers.ts:2197-2199`) and threaded it through `FirstRunGateOutcome` (`:2177`), `pollFirstRunGate`'s return type and give-up return (`:2270`, `:2299`), `dismissStuckFirstRunGate`'s forwarding (`:2483`), and `describeInconclusiveOverlayTimeout`, which now takes an optional `sampleReadFailure?: unknown` parameter (`:2552-2554`) with its own accurate message branch (`:2557`) rather than the misleading canned one. `waitForAppReady` branches on the new shape at `:2635-2636`. Updated the existing poll test's assertion to the new shape and the (now correctly-placed, per #4) `@returns` TSDoc.

rolfheij-sil:
🤖 Claude: Confirmed — `escapeHatch` was scoped to `page` while `dialogHeading`/`errorScreen` were already scoped to `firstRunDialog`, and Playwright's `isVisible()`/`click()` do hardcode strict mode internally (verified directly against `playwright-core/lib/client/locator.js`), so a second page-wide match would have thrown inside the sampler's concurrent read.

Fixed in `3865d3f4ab3`: `escapeHatch` is now built from `firstRunDialog.getByRole(...)` (`helpers.ts:2453`), covering both the sampler's read (`:2474`) and the click (`:2520`) through the same binding, as you suggested. No new test — this is a latent-hazard fix; no second matching button exists in the shipped UI to reproduce it against today.

(Finding #7 in your review — posting here since GitHub had nowhere to anchor an inline reply.)

e2e-tests/fixtures/helpers.ts
rolfheij-sil: 🤖 Claude: Confirmed — joining all four discriminator reads in one `Promise.all` meant any single rejection discarded `gateVisible` with it and counted as a sample failure, where the old `Promise.race` design's `gateVisible` leg was independent and would have kept retrying through the same transient to `'cleared'`.

Fixed in `8633586c29c`: extracted an exported `buildFirstRunGateSample(locators)` (`helpers.ts:2356`) and swapped to `Promise.allSettled` (`:2360`), rethrowing immediately on a `TargetClosedError` or on `gateVisible`'s own rejection, returning the all-false/cleared shape if `gateVisible` resolved false, and defaulting any other individually-rejected discriminator to `false`. Added four new test cases in `first-run-gate-poll.test.ts` — the two you named ("gate gone, one discriminator throws" / "gate visible, one discriminator throws") plus two more covering the `gateVisible`-rejects-rethrows and `TargetClosedError`-from-a-non-gate-read paths, since both are explicit behaviors in your fix description that the first two alone don't exercise.

Worth noting: fixing #7 (scoping `escapeHatch` to the dialog) removes today's one concretely-identified live trigger for this bug — but this hardening is still real value beyond that, since `Promise.all` over four independent reads stays fragile to any future rejection (a different locator becoming non-unique, a mid-poll DOM detachment), which the old race-based design was structurally immune to. Did both.

e2e-tests/fixtures/onboarding-tour.page.ts
rolfheij-sil: 🤖 Claude: Confirmed — `waitForStepCounterChange`'s poll read `count()` and `dialog.isVisible()` together, then made a separate later `textContent()` round trip that could stall forever if a Done click unmounted the whole dialog in between.

Fixed in `e10f1acb542`: extracted the per-sample logic into `sampleStepCounterPoll` (`onboarding-tour.page.ts:155`), which reads the counter text via a single non-waiting `stepCounter.evaluateAll(...)` call instead of the count-then-textContent pair, preserving the existing `.trim()` normalization. Added `onboarding-tour.page.test.ts` with fake-injected cases proving the old shape would stall on a `textContent()` that never resolves and the new one doesn't.

Self-review caught one gap in the suggested `els[0]?.textContent ?? undefined` shape: it silently reads the first match instead of the old `textContent()` call's implicit strict-mode rejection of multiple matches. Closed that in `f329f0a6fcd` — the `evaluateAll` callback now throws if it finds more than one step counter (`:165`), so a future double-render bug still fails loudly rather than reading whichever element happens to be first.

rolfheij-sil:
🤖 Claude: Confirmed — `advanceTour`/`advanceToLastStep` both wait for the step transition via `waitForStepCounterChange`, but `goBackTour` was a bare click with no wait, exposed to the same stale-`targetRect` unmount `tour.component.tsx` triggers on every step transition, not just Done.

Fixed in `391faa4b54f`: `goBackTour` (`onboarding-tour.page.ts:239`) now mirrors `advanceTour` — captures the step-counter text before clicking Back, then awaits `waitForStepCounterChange` afterward. Docblock updated to say it waits for the transition.

(Finding #10 in your review — posting here since GitHub had nowhere to anchor an inline reply.)

e2e-tests/global-setup.ts
rolfheij-sil: 🤖 Claude: Confirmed — the dev-server spawn's `detached`/`windowsHide` platform split was two inline literals with no seam and no test, unlike every other new platform conditional in this PR.

Fixed in `094a9e93e70`: extracted `export function devServerSpawnFlags(platform: NodeJS.Platform)` (`global-setup.ts:104`), spread into the `spawn(...)` call (`:294`), with the explanatory comments moved onto the function. Added the two cases you named to `global-setup.test.ts` (`win32` → `{ detached: false, windowsHide: true }`, `linux` → `{ detached: true, windowsHide: true }`), matching the `killProcessTree`/`pollFirstRunGate` injectable-seam convention already established elsewhere in this PR.

lyonsil (approved):
:lgtm:

> @lyonsil partially reviewed 44 files and all commit messages, made 1 comment, and resolved 8 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2794#-:-P28TdcyENwKB7x6yNtc:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

Issue #15, opened 2023-02-01
Investigate cra-craco-electron


tjcouch-sil:
Investigate https://github.com/sillsdev/cra-craco-electron to determine if we want to restart our app on it.

- [ ] Cost from switching - how long will it take to switch now vs a year from now?
  - [x] Investigate differences from CRA and see if we can adapt ERB
- [ ] Benefits from switching
- [ ] Drawbacks from switching

irahopkinson:
Our repo with ERB already builds for the web. After `npm run build` (or more specifically `npm run build:renderer`), open `/release/app/dist/renderer/index.html` in a browser.

We would need to make some minor adjustments to exclude electron references but if we end up not using electron IPC then we probably don't need to do anything.

irahopkinson:
If we need to develop exclusively on the web we can just add a script in `package.json` like `"start:web": "cross-env NODE_ENV=development TS_NODE_TRANSPILE_ONLY=true webpack serve --config ./.erb/configs/webpack.config.web.dev.ts",` and then create the `.erb/configs/webpack.config.web.dev.ts` file (we could eject a CRA boilerplate and adapt the webpack config from that).

However, developing React inside electron should cover just about every scenario. Once we have a feature working in electron we can then just run `/release/app/dist/renderer/index.html` with the appropriate backend to check its working the same.

Issue #14, opened 2023-02-01
Create release 0.0.1


tjcouch-sil:
- [x] Fix [installer icon](https://www.electron.build/icons.html)
- [x] Add portable package to windows
  - In package.json
  - ![Image](https://user-images.githubusercontent.com/104016682/216160917-2efe8088-fc5d-48cf-bba8-cb7a5e753386.png)
- [x] Consider the fixes in [4 must-know tips for building cross platform Electron apps](https://blog.avocode.com/4-must-know-tips-for-building-cross-platform-electron-apps-f3ae9c2bffff).

- [x] Create full release 0.0.1

Depends on #6

irahopkinson:
For item 3 above (_4 must-know tips for building cross platform Electron apps_), the first 2 points are already a part of ERB.

PR #2796, opened 2026-09-10
PT-4548: Localize the settings sidebar's project picker

jolierabideau:
The Settings sidebar's project picker showed its popover in English regardless of UI language.

`SettingsSidebarProps` gains `searchPlaceholderText` and `noResultsText` — two optional plain string
props in the style of the existing `buttonPlaceholderText`. In `project` mode with no open tabs the
picker can only render those two strings, so `SettingsTab` resolves the two `%projectSelector_*%`
keys in its existing `useLocalizedStrings` subscription and passes them through. Both fall back to
the picker's English defaults when omitted.

`SettingsSidebarContentSearch` now spreads the sidebar's props rather than forwarding them one at a
time, so a prop added to `SettingsSidebar` reaches it without a second edit. That also fixes
`className`, which the hand-list had been silently dropping.

No new localization keys. No `papi.d.ts` change. Nothing on the stable barrel depends on the
experimental entry point.

Covered by four `SettingsSidebar` tests, one `SettingsTab` wiring test that opens the picker, and an
en/es parity test for the two keys.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2796)


katherinejensen00 (changes requested):
# PR #2796 review — Localize the settings sidebar's project picker

PR: https://github.com/paranext/paranext-core/pull/2796
Head reviewed: `e78213cb4d8` (base `pt-4548-promote-project-selector`)
Scope: 4 source files, +74/−5 (the rest of the diff is regenerated `lib/platform-bible-react/dist`, checked and consistent with source)

All paths are repo-relative. Line numbers are PR-head lines. Every finding below was verified against the code; the raw-key behavior in #1 was reproduced with a throwaway test.

## Severity summary

| # | Severity | File | Line | Finding |
|---|---|---|---|---|
| 1 | Medium | `lib/platform-bible-react/src/components/advanced/settings-components/settings-sidebar.component.tsx` | 117 | Raw `%key%` values pass the filter; TSDoc fallback promise is false |
| 2 | Medium | `lib/platform-bible-react/src/components/advanced/settings-components/settings-sidebar.component.tsx` | 109 | Comment gives the wrong reason the grouping menu is absent |
| 3 | Medium | `lib/platform-bible-react/src/components/advanced/settings-components/settings-sidebar.component.tsx` | 107 | New memo orphans the existing `projectSelectorProjects` comment |
| 4 | Medium | `lib/platform-bible-react/src/components/advanced/settings-components/settings-sidebar.component.tsx` | 53 | Stable-barrel prop depends on experimental-only exports |
| 5 | Low | `lib/platform-bible-react/src/components/advanced/settings-components/settings-sidebar.component.tsx` | 112 | Undefined-filter overlaps open PR #2829; PR body misdescribes #2829 |
| 6 | Low | `lib/platform-bible-react/src/components/advanced/settings-components/settings-sidebar.component.tsx` | 115 | Moving the bag into `useMemo` loses TypeScript key checking |
| 7 | Low | `lib/platform-bible-react/src/components/advanced/settings-components/settings-sidebar.component.tsx` | 63 | Prop name means the adapted camelCase bag everywhere else |
| 8 | Low | `lib/platform-bible-react/src/components/advanced/settings-components/settings-sidebar.component.test.tsx` | 110 | "Merged channels" assertion cannot fail |
| 9 | Low | `lib/platform-bible-react/src/components/advanced/settings-components/settings-sidebar.component.test.tsx` | 104 | `commandEmptyMessage` fixture is supplied but never asserted |
| 10 | Low | `src/renderer/components/settings-tabs/settings-tab.component.tsx` | 311 | Production wiring has no test |
| 11 | Low | `src/renderer/components/settings-tabs/settings-tab.component.tsx` | 61 | No en/es localization test for the 21 newly requested keys |
| 12 | Low | `src/renderer/components/settings-tabs/settings-tab.component.tsx` | 61 | Subscribes to 21 keys; the picker here can only show 2 |
| 13 | Low | `lib/platform-bible-react/src/components/advanced/settings-components/settings-sidebar-content-search.component.tsx` | 34 | Hand-listed prop forwarding already drops `className` |
| 14 | Low (follow-up) | `lib/platform-bible-react/src/components/advanced/settings-components/settings-sidebar-content-search.component.tsx` | 34 | Search bar placeholder is still hard-coded English |
| 15 | Low (process) | PR description | — | Description is the pasted review summary for a withdrawn scope |

---

## 1. Raw `%key%` values pass the filter, so the TSDoc's fallback promise is false

**Severity:** Medium
**Attach at:** `lib/platform-bible-react/src/components/advanced/settings-components/settings-sidebar.component.tsx`, line 117

**Summary:** Before localization loads, the picker shows literal `%projectSelector_searchPlaceholder%` text instead of the English default, contradicting the new TSDoc on line 56.

The filter only drops `undefined`. But `useLocalizedStrings` fills in every key it hasn't resolved yet with the key itself (`defaultState[key] = key`, `use-localized-strings-hook.ts:48`), and returns that seed for good if there's a platform error (`:54`). A raw key is a string, so `readProjectSelectorString` keeps it, the filter keeps it, and `resolveStrings`'s `{ ...DEFAULT_STRINGS, ...partial }` lets it overwrite the English default.

Reproduced with a throwaway test passing `{ '%projectSelector_searchPlaceholder%': '%projectSelector_searchPlaceholder%' }`: the only rendered placeholder was `"%projectSelector_searchPlaceholder%"`. A search with no matches shows `%projectSelector_commandEmptyMessage%` the same way.

In the app this only shows up while localization is loading or after a localization error, and in that window the sidebar's own labels already show raw keys, so the in-app impact is narrow. The bigger issue is that line 56's "Any key left unresolved falls back to the picker's English default" is wrong for exactly the wiring the TSDoc tells callers to use.

**Suggested fix:** Open PR #2829 already fixes this at the source (it drops `undefined`, raw keys, and blank values in `resolveStrings` and `readProjectSelectorString`). Either land #2829 first, or reject raw keys here too (`localization.util.ts` already has `LOCALIZATION_KEY_PATTERN`). Either way, add a test that passes the seed shape above.

---

## 2. The comment gives the wrong reason the grouping menu is absent

**Severity:** Medium
**Attach at:** `lib/platform-bible-react/src/components/advanced/settings-components/settings-sidebar.component.tsx`, line 109

**Summary:** Leaving out `availableGroupings` is what turns groupings *on* in `ProjectSelector`; the menu is absent here because of `openTabs={[]}`.

`project-selector.component.tsx:742` returns `props.availableGroupings` only when it is defined. When it is omitted, the component auto-derives groupings (an open-tabs grouping when `openTabs.length > 0`, a selection grouping in `project-multi` mode). This sidebar gets a flat list because it passes `openTabs={[]}` in `project` mode. The conclusion is true today, but anyone who wires real open tabs into this sidebar will read this comment and assume the grouping strings can't be reached, when the section headings would start rendering.

The JSX comment at lines 179–186 already explains this correctly ("`openTabs={[]}` keeps the ProjectSelector on a flat (non-grouped) list"), so the new sentence says the same thing less accurately.

**Suggested fix:** Delete the grouping sentence (lines 108–110), or replace it with:

```ts
// Grouping strings in the block go unused while this sidebar passes `openTabs={[]}` (see the
// comment on <ProjectSelector> below).
```

---

## 3. The new memo orphans the existing `projectSelectorProjects` comment

**Severity:** Medium
**Attach at:** `lib/platform-bible-react/src/components/advanced/settings-components/settings-sidebar.component.tsx`, line 107

**Summary:** The 4-line "Adapt the public `ProjectInfo[]` shape…" comment now sits above the strings memo instead of the projects memo it describes.

On the base branch, lines 88–91 describe `projectSelectorProjects`, which comes right after them. This PR appends 8 more comment lines and an 11-line memo in between, so the comment now opens a 12-line block covering two unrelated topics, and `projectSelectorProjects` (line 127) has no comment. If the strings memo is ever moved or deleted, the `shortName`/`fullName` explanation goes with it. A block this long is also easy to skip.

**Suggested fix:** Put `projectSelectorStrings` below `projectSelectorProjects` (or move lines 103–106 back down to line 127), and trim the strings comment to about three lines:

```ts
// `buttonPlaceholder` and `ariaLabel` are this sidebar's own copy, merged over the shared
// `%projectSelector_*%` block. Unset entries are dropped first: ProjectSelector spreads this bag
// over its English defaults, so an explicit `undefined` would blank a default.
```

---

## 4. A stable-barrel prop depends on experimental-only exports

**Severity:** Medium
**Attach at:** `lib/platform-bible-react/src/components/advanced/settings-components/settings-sidebar.component.tsx`, line 53

**Summary:** `SettingsSidebar` is on the stable `index.ts`, but the only way its TSDoc tells you to fill the new prop is `PROJECT_SELECTOR_STRING_KEYS` from `platform-bible-react/experimental`, which promises no stability.

Typing the prop as `LanguageStrings` keeps the experimental *type* off the stable surface, but what the prop means still depends on an experimental constant and adapter (`experimental.ts:26–28`). `experimental.ts` says its exports "may change shape, move, or disappear at any time without a deprecation cycle", and `adr-project-selector-stays-experimental` records these keys being renamed before. A caller who types keys by hand falls back to English after such a rename, with no type error.

The library's own pattern is to put the nested key list and its mapped type on the same surface as the component: `scope-selector.component.tsx:200` forwards `bookChapterControlLocalizedStrings?: BookChapterControlLocalizedStrings`, and both `BOOK_CHAPTER_CONTROL_STRING_KEYS` and that type are exported from `index.ts:7,9`. `index.ts` pairs more than 15 `*_STRING_KEYS` constants with their components this way.

Mitigating: the prop is optional, and the only production caller is in-repo. The renderer already imports from `experimental` in about 12 places.

**Suggested fix (your call):** Pick one:
- Mark the prop `@experimental` (as `book-chapter-control.types.ts:169` does).
- Export a sidebar-owned key list from the stable barrel.
- Replace the prop with two plain string props (`searchPlaceholderText`, `noResultsText`) in the style of the existing `buttonPlaceholderText`. This also removes #7 and #12.

---

## 5. The undefined-filter overlaps open PR #2829, and the PR body misdescribes #2829

**Severity:** Low
**Attach at:** `lib/platform-bible-react/src/components/advanced/settings-components/settings-sidebar.component.tsx`, line 112

**Summary:** #2829 fixes the `resolveStrings` spread at the source, so once both PRs merge this filter does nothing and its comment ("with a plain spread") is false.

The filter is correct and needed today. The prop is optional, so `buildProjectSelectorLocalizedStrings({})` returns every field as `undefined`, which would blank all the defaults, and your defaults test covers that. The four other `ProjectSelector` consumers don't hit this because they request the full key list, and the seed never produces `undefined`.

#2829's `resolveStrings` now drops `undefined`, raw-key, and blank values. The PR body's "Suggested Review Focus" says "the `undefined` arm is not obviously covered by that work", but #2829's diff covers it explicitly.

**Suggested fix:** Remove the filter and its comment in whichever PR merges second, and correct that line in the PR body.

---

## 6. Moving the bag into `useMemo` loses TypeScript key checking

**Severity:** Low
**Attach at:** `lib/platform-bible-react/src/components/advanced/settings-components/settings-sidebar.component.tsx`, line 115

**Summary:** A typo like `buttonPlaceHolder` compiled as an error on base, but compiles cleanly in this PR.

On base, `{ buttonPlaceholder, ariaLabel }` was an inline object literal on the JSX prop, so TypeScript's excess-property check applied (TS2561). The unannotated memo turns that check off. Also, `Object.fromEntries(Object.entries(...))` widens `supplied` to `{ [k: string]: string }`, so a renamed field on the experimental `ProjectSelectorLocalizedStrings` would be dropped silently too. The tests wouldn't catch a typo either (see #8).

**Suggested fix:** Add `satisfies ProjectSelectorLocalizedStrings` to the returned object, or write `useMemo((): ProjectSelectorLocalizedStrings => { … })`. Note that `useMemo<ProjectSelectorLocalizedStrings>` does not bring the check back.

---

## 7. The prop name already means the adapted camelCase bag everywhere else

**Severity:** Low
**Attach at:** `lib/platform-bible-react/src/components/advanced/settings-components/settings-sidebar.component.tsx`, line 63

**Summary:** `projectSelectorLocalizedStrings` holds the raw `%projectSelector_*%` bag here, but the same name holds the already-adapted `ProjectSelectorLocalizedStrings` in six other places.

It's a prop in `manage-books-dialog.component.tsx:266` and `manage-books-sidebar.component.tsx:177`, and a local variable in `find.component.tsx:972`, `checks-side-panel.component.tsx:223`, `manage-books.web-view.tsx:542`, and `checklist.web-view.tsx:771`. Passing one of those named variables into this prop compiles (only a fresh object literal is rejected), every lookup returns `undefined`, and the popover stays English in every language. Nothing is wrong today, but this is about to become stable public API.

**Suggested fix:** Rename it before merge (e.g. `projectSelectorLanguageStrings`), or take option 3 from #4.

---

## 8. The "merged channels" assertion cannot fail

**Severity:** Low
**Attach at:** `lib/platform-bible-react/src/components/advanced/settings-components/settings-sidebar.component.test.tsx`, line 110

**Summary:** The comment says the trigger keeps the sidebar's own placeholder and accessible name, but nothing in the test can go red if that breaks.

- `getByRole('combobox', { name: PROJECTS_GROUP_LABEL })` on line 115 repeats the query that already succeeded on line 108, so it cannot fail at that point.
- The fixture's `buttonPlaceholderText="Select a project"` (line 67) matches the picker's English default, and nothing reads the trigger text. A change like `...(projectSelectorLocalizedStrings ? {} : { buttonPlaceholder: buttonPlaceholderText })` passes typecheck, lint, and every test, yet in the app the trigger would show the English default instead of the localized "Select project".
- The merge order can't be tested with this fixture anyway: the adapter never outputs `buttonPlaceholder` or `ariaLabel`, so swapping the spread order changes nothing.
- Line 111 uses `--` where the rest of these files use `—`.

This conflicts with the Revert Test in `.claude/rules/testing/tdd-discipline.md`.

**Suggested fix:** Give the fixture a trigger placeholder different from the default, assert it with `expect(screen.getByRole('combobox', …)).toHaveTextContent(…)`, and cut the comment to one line.

---

## 9. The `commandEmptyMessage` fixture is supplied but never asserted

**Severity:** Low
**Attach at:** `lib/platform-bible-react/src/components/advanced/settings-components/settings-sidebar.component.test.tsx`, line 104

**Summary:** The "no results" message is half of what the new TSDoc promises, and nothing tests it, here or in `ProjectSelector`'s own tests.

**Suggested fix:** Type a query that matches no project, then `expect(await screen.findByText('No se encontraron proyectos')).toBeInTheDocument()`.

---

## 10. The production wiring has no test

**Severity:** Low
**Attach at:** `src/renderer/components/settings-tabs/settings-tab.component.tsx`, line 311

**Summary:** Deleting any piece of the wiring still compiles and passes every test, and a Spanish user's popover quietly goes back to English.

Pieces that can be deleted undetected:
- Line 311.
- The key spread on line 63 together with its import on line 7.
- Both forwarding lines in `settings-sidebar-content-search.component.tsx` (34 and 59).

`settings-tab.component.test.tsx` mocks `useLocalizedStrings` with a one-key bag and never opens the picker, and `SettingsSidebarContentSearch` has no test. The PR body says the diff "is entirely covered by tests", which isn't accurate.

**Suggested fix:** Add one renderer test that supplies `%projectSelector_*%` values and opens the picker, or drop the coverage claim.

---

## 11. No en/es localization test for the 21 newly requested keys

**Severity:** Low
**Attach at:** `src/renderer/components/settings-tabs/settings-tab.component.tsx`, line 61

**Summary:** The renderer now depends on all 21 `%projectSelector_*%` keys, but no test checks they have English and Spanish values. `Localization-Guide.md:251–257` asks for that test in the same change.

The earlier revision of this PR had one (`src/renderer/components/project-selector-localization.test.ts` in `d1ceb881f90`). It was withdrawn as superseded by base, but base doesn't have it: `shipped-locale-assets.test.ts` only requires library keys to exist in English, and its en/es parity check covers only the marker menu. The renderer already has this kind of test for other library key sets (`book-chapter-control-localization.test.ts`, `resource-picker-dialog-localization.test.ts`).

All 21 keys do have Spanish values today, so this is a missing safeguard, not a current bug.

**Suggested fix:** Restore the en/es half of the withdrawn test.

---

## 12. The tab subscribes to 21 keys; the picker here can only show 2

**Severity:** Low
**Attach at:** `src/renderer/components/settings-tabs/settings-tab.component.tsx`, line 61

**Summary:** In `project` mode with no open tabs, this picker can only render `searchPlaceholder` and `commandEmptyMessage`, but the tab resolves all 21 keys. The project-only settings view (lines 259–272) shows none of them and still fetches all 21.

The cost is small (one request per mount or language change). There's also a small documentation mismatch: the TSDoc says the prop covers two strings but tells callers to resolve the whole list.

**Suggested fix (optional):** Request only those two keys, or keep the full list on purpose and say so in the comment.

---

## 13. Hand-listed prop forwarding already drops `className`

**Severity:** Low (predates this PR)
**Attach at:** `lib/platform-bible-react/src/components/advanced/settings-components/settings-sidebar-content-search.component.tsx`, line 34

**Summary:** `SettingsSidebarContentSearchProps` includes all of `SettingsSidebarProps`, but the component forwards them one at a time, so a newly added prop is silently dropped unless someone remembers to add two lines. That has already happened to `className`.

`className` is never destructured (lines 22–35), and line 51 hard-codes the sidebar's classes, so `<SettingsSidebarContentSearch className=…>` type-checks and does nothing. This PR only works because both forwarding lines were remembered, and no test would have caught it if they weren't (#10).

**Suggested fix:** Destructure `{ id, children, searchValue, onSearch, className, ...sidebarProps }`, spread `sidebarProps` onto `<SettingsSidebar>`, and merge `className` with `cn(...)`. Keep `id` out of the spread so the DOM id doesn't appear twice.

---

## 14. The search bar placeholder is still hard-coded English

**Severity:** Low (follow-up, out of scope)
**Attach at:** `lib/platform-bible-react/src/components/advanced/settings-components/settings-sidebar-content-search.component.tsx`, line 34 (the literal is on line 43, outside the diff hunks)

**Summary:** "Search app settings, extension settings, and project settings" is now the only English text left on a Spanish Settings tab, right above the newly localized picker.

There's no localization key for it, so fixing it needs a new key, a new prop, and renderer wiring. It doesn't belong in this PR, but it's worth a Jira follow-up.

---

## 15. The PR description is the pasted review summary for a withdrawn scope

**Severity:** Low (process)
**Attach at:** general PR comment

**Summary:** The body opens with "That scope has been superseded and withdrawn" and spends most of its length on withdrawn work. Its own checklist says "The PR title and description still describe the old scope and must be rewritten before posting", and the description hasn't been rewritten yet.

What a reviewer needs ("the settings sidebar's project picker can now be localized") is buried mid-paragraph. The body also contains two claims corrected above: #2829 not covering `undefined` (#5), and "entirely covered by tests" (#10).

**Suggested fix:** Replace it with a short description, and move the merge history and interview notes to the squash commit message:

```markdown
The settings sidebar's project picker showed its popover in English regardless of UI language.
`SettingsSidebarProps` gains `projectSelectorLocalizedStrings?: LanguageStrings`; `SettingsTab`
resolves `PROJECT_SELECTOR_STRING_KEYS` in its existing `useLocalizedStrings` subscription and
passes them through.

Unset entries are dropped before the bag reaches the picker, because `resolveStrings` spreads over
its English defaults and an explicit `undefined` would blank them. This overlaps #2829, which fixes
that at the source; whichever PR merges second should remove the call-site filter.

No new localization keys; no `papi.d.ts` change.
```

---

## Checked and cleared

- All 21 `%projectSelector_*%` keys exist in `assets/localization/en.json` and `es.json`.
- Typing the prop as `LanguageStrings` matches convention: `LocalizationData` is `LanguageStrings`, and `checks-side-panel.component.tsx:105` uses the same type.
- The committed `dist` matches the source changes.
- Passing the tab's whole `localizedStrings` bag is harmless, because the adapter reads only `%projectSelector_*%` keys, and the four extension consumers do the same.
- No backward-facing comments (no ticket IDs, change narration, or dated notes).
- The new "defaults" test can fail: without the filter, the placeholder goes blank.
- The existing test file passes at the PR head (4/4).

> @katherinejensen00 partially reviewed 13 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2796#-:-P1kBGzD5eYIXOrhWZ9D:bbagvds)* status: all files reviewed, 1 unresolved discussion (waiting on jolierabideau).

___
*[`lib/platform-bible-react/src/components/advanced/settings-components/settings-sidebar.component.test.tsx` line 117 at r1](https://reviewable.io/reviews/paranext/paranext-core/2796#-P1kBO858DcJzR-dtAN7:-P1kBO858DcJzR-dtAN8:be19vvy) ([raw file](https://github.com/paranext/paranext-core/blob/e78213cb4d8d0f0facab567b4658c5b9f9b7feca/lib/platform-bible-react/src/components/advanced/settings-components/settings-sidebar.component.test.tsx#L117)):*
> ```XML
>     expect(screen.getByRole('combobox', { name: PROJECTS_GROUP_LABEL })).toBeInTheDocument();
>   });
> });
> ```

## 1. Raw `%key%` values pass the filter, so the TSDoc's fallback promise is false

**Severity:** Medium
**Attach at:** `lib/platform-bible-react/src/components/advanced/settings-components/settings-sidebar.component.tsx`, line 117

**Summary:** Before localization loads, the picker shows literal `%projectSelector_searchPlaceholder%` text instead of the English default, contradicting the new TSDoc on line 56.

The filter only drops `undefined`. But `useLocalizedStrings` fills in every key it hasn't resolved yet with the key itself (`defaultState[key] = key`, `use-localized-strings-hook.ts:48`), and returns that seed for good if there's a platform error (`:54`). A raw key is a string, so `readProjectSelectorString` keeps it, the filter keeps it, and `resolveStrings`'s `{ ...DEFAULT_STRINGS, ...partial }` lets it overwrite the English default.

Reproduced with a throwaway test passing `{ '%projectSelector_searchPlaceholder%': '%projectSelector_searchPlaceholder%' }`: the only rendered placeholder was `"%projectSelector_searchPlaceholder%"`. A search with no matches shows `%projectSelector_commandEmptyMessage%` the same way.

In the app this only shows up while localization is loading or after a localization error, and in that window the sidebar's own labels already show raw keys, so the in-app impact is narrow. The bigger issue is that line 56's "Any key left unresolved falls back to the picker's English default" is wrong for exactly the wiring the TSDoc tells callers to use.

**Suggested fix:** Open PR #2829 already fixes this at the source (it drops `undefined`, raw keys, and blank values in `resolveStrings` and `readProjectSelectorString`). Either land #2829 first, or reject raw keys here too (`localization.util.ts` already has `LOCALIZATION_KEY_PATTERN`). Either way, add a test that passes the seed shape above.

jolierabideau:
Took option 3 from #4 — `searchPlaceholderText` and `noResultsText` as plain string props, in the
style of the existing `buttonPlaceholderText`. That resolved #4, #5, #6, #7 and #12 structurally
rather than one at a time. Branch is now up to date with main (main → `pt-4548-promote-project-selector`
→ here, both clean).

Pushed as `4743b587bf2`.

### Medium

**#1 — raw `%key%` values pass the filter.** The filter is gone; there is nothing left to pass. The
TSDoc no longer promises a fallback for unresolved keys — it now says only that an *omitted* prop
falls back to the picker's English default, which is true by construction (unset entries are never
written onto the bag, so `resolveStrings`'s spread can't blank a default).

The raw-`%key%` arm itself is unchanged and deliberately so: these two props now behave exactly like
the `buttonPlaceholderText` beside them, which has always passed an unresolved key straight through.
Fixing that here would have meant one component filtering keys that every sibling prop doesn't.
#2829 fixes it at the source in `resolveStrings`, which covers all four picker consumers at once.

**#2 — wrong reason for the absent grouping menu.** Deleted. You're right that omitting
`availableGroupings` turns auto-derivation *on* (`project-selector.component.tsx:742`), and the JSX
comment at the `<ProjectSelector>` call already gives the real reason (`openTabs={[]}`). No point
restating it less accurately 70 lines up.

**#3 — orphaned `projectSelectorProjects` comment.** `projectSelectorProjects` is back directly under
its own 4-line comment, and the strings memo moved below it with its own 4-line comment covering
only the strings.

**#4 — stable barrel depending on experimental exports.** Option 3. `SettingsSidebar` no longer
imports `buildProjectSelectorLocalizedStrings`, and `SettingsTab` no longer imports
`PROJECT_SELECTOR_STRING_KEYS` from `platform-bible-react/experimental`. The two key literals live in
a new `settings-tab.localization.ts`, pinned against the shipped assets by the test in #11 — so a
rename in the library goes red on the next run instead of falling back to English silently.

### Low

**#5 — overlap with #2829.** Gone; there is no call-site filter left to remove when #2829 merges.
The PR body's claim about #2829 not covering the `undefined` arm was wrong and the body is rewritten.

**#6 — lost key checking.** The memo is now `useMemo((): ProjectSelectorLocalizedStrings => …)`
building a typed local, so every property name is checked — including the two conditional
assignments, which a spread would have skipped. Confirmed `buttonPlaceHolder` is a compile error
again.

**#7 — prop name collision.** Retired along with the prop.

**#8 — non-falsifiable "merged channels" assertion.** Fixture trigger placeholder is now
`'Choose a project'`, distinct from the picker's `buttonPlaceholder` default `'Select a project'`,
and the assertion is its own test using `toHaveTextContent`. Verified falsifiable: setting
`buttonPlaceholder: undefined` reddens it. Em dash fixed.

**#9 — `commandEmptyMessage` never asserted.** Both the defaults test and the localized test now type
a non-matching query and assert the message (`'No projects found'` / `'No se encontraron proyectos'`).

**#10 — production wiring untested.** New `SettingsTab project picker localization` test in
`settings-tab.component.test.tsx` renders the real tab, opens the picker, and asserts the Spanish
placeholder and "no results". Verified red when line 316 is deleted. The two file-level mocks it
needed (`getMetadataForAllProjects`, `projectDataProviders.get`) got declared return types instead of
`as unknown as` casts. The "entirely covered by tests" claim is out of the body.

**#11 — no en/es test.** Restored as `src/renderer/components/settings-project-picker-localization.test.ts`,
following `book-chapter-control-localization.test.ts`. Two keys rather than 21, importing them from
`settings-tab.localization` so the test pins the exact literals the tab uses.

**#12 — 21 keys for 2 strings.** The tab now requests the two it can render.

**#13 — hand-listed forwarding drops props.** Destructured to `{ id, children, searchValue, onSearch,
className, ...sidebarProps }` and spread onto `<SettingsSidebar>`; `id` kept out so the DOM id isn't
duplicated; `className` merged with `cn()`.

**#14 — hard-coded English search bar placeholder.** Agreed and deferred — a new localization key, a
new prop, and renderer wiring don't belong in this PR. Not filed yet; flagging it here so it isn't
lost.

**#15 — PR description.** Replaced with the short version.

### Gates

`npm run typecheck` (all four projects exit 0) · `npm run lint` 0 errors · `npx prettier --check`
clean · `npm test` green. Two storybook browser tests failed under parallel load and pass in
isolation — a different pair on each run, unrelated to this change. The
`PROJECT_SELECTOR_CUSTOM_DATA_KEYS` import failures were the stale `node_modules/.cache/storybook`
dep cache again; clearing it fixed them, no code change.

katherinejensen00 (approved):
:lgtm: Thanks for making those changes, Jolie!

> @katherinejensen00 partially reviewed 17 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2796#-:-P1p-j4X799T-olO15Ng:bdv5izw)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #187, opened 2023-05-15
Split Vite build into two steps to enable WebView debugging and library imports

tjcouch-sil:
Goes with https://github.com/paranext/paranext-extension-template/pull/11

Resolves #119 

This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/187)

Issue #186, opened 2023-05-15
Fix Paranext crashing in development while rebuilding in watch mode


tjcouch-sil:
We have so many build steps watching for file changes to files that other build steps generate. It has become a problem because, sometimes, (most frequent example) the second Vite build step crashes because it tries to read a file that the first Vite build step is writing. This issue is especially prevalent when you change a webview file, save, change it again, save, ect. very quickly. [buildDelay](https://rollupjs.org/configuration-options/#watch-builddelay) and [awaitWriteFinish](https://github.com/paulmillr/chokidar#performance) did not seem to fix the problem.

Created from #119

tjcouch-sil:
It seems maybe having both `paranext-core` and `extensions` open in two instances of VSCode increases the chances of this happening. It also seems this only or almost only crashes when working on extension (maybe specifically webview) files. It may be linked to `rollup-plugin-static-copy`

tjcouch-sil:
I wonder if #280 helped to alleviate this issue. Maybe; maybe not. It seems webpack does have a way to wait before rebuilding when watching: https://webpack.js.org/configuration/watch/#watchoptionsaggregatetimeout if we still see this issue, maybe we can try that!

PR #2797, opened 2026-09-10
ci: stop a flaky Linux snap install failing the whole build

tjcouch-sil:
Runs [34488976360](https://github.com/paranext/paranext-core/actions/runs/34488976360) and [34384474185](https://github.com/paranext/paranext-core/actions/runs/34384474185) both died in **Install snap tools on Linux** and took the Windows and macOS jobs down with them. Neither failure was ours — the snap store CDN answered `snap install snapcraft` with a 502 and snapd gave up on the first try, failing the job before the build had started:

```
error: cannot perform the following tasks:
- Download snap "snapcraft" (17614) from channel "8.x/stable" (received an
  unexpected http response code (502) when trying to download ...)
```

All three workflows that build a snap get the same block: `test.yml`, `package-main.yml`, `publish.yml`.

## Changes

- **Retry the three snap store operations**, three attempts 30 s apart. The helper runs the command inside an `if` condition so `bash -e` does not abort on an attempt that is *meant* to fail, and returns non-zero after the last one so a genuinely broken store still fails the step.
- **Hold snapd's auto-refresh** for the job. snapd refreshes installed snaps on its own schedule, and refreshing `lxd` while snapcraft is packing inside an LXD instance restarts the daemon out from under the running build. Best-effort, so a snapd that rejects the option cannot fail a build.
- **Capture diagnostics when the Linux job fails** — snapcraft's own log, which holds the real mksquashfs error rather than the bare `mksquashfs call failed:` that reaches the console, plus kernel OOM activity. On-failure only. In `test.yml` these sit last in the job rather than beside the packaging step, so an E2E run the kernel kills for memory reaches them too.
- **Drop `newgrp lxd`.** In a script it execs a *child* shell that reads EOF from stdin and exits immediately, so it never changed this shell's groups; the `chmod` on the socket is what actually makes LXD usable.
- **Assert the three copies stay identical**, in `.erb/scripts/electron-builder-snap-config.test.ts` — see below.

## Why `test.yml` is in scope

`Build on ubuntu-22.04, .Net 8.0.x` is `test.yml`'s rendered job name, and it is a required status check on `main`. `package-main.yml` and `publish.yml` render as `Main on ...` / `Release on ...` and are not required. That leg does build a snap — `check packaging` has no `if:` guard, `electron-builder.json5:79` sets `linux.target: ['snap']`, and the job renames and uploads `./release/build/*.snap` — so hardening only the two post-merge workflows would have left the flake live on the one leg that can block a pull request.

Blast radius differs per workflow, and the difference is worth stating rather than flattening: `test.yml` sets `fail-fast: false`, so a 502 there costs one re-run of the Linux leg. `package-main.yml` and `publish.yml` fail fast, so a 502 on their Linux leg also cancels the healthy Windows and macOS jobs.

## Why the step is duplicated rather than extracted

In all three workflows `Install snap tools on Linux` runs **before** `actions/checkout` (`test.yml` 93 vs 114, `package-main.yml` 75 vs 96, `publish.yml` 94 vs 115), so it cannot be a local composite action — nothing is on disk to `uses:` yet. Extracting it would mean reordering checkout ahead of it in three workflows, and `publish.yml` is `workflow_dispatch`-only, so its copy could not be proven short of cutting a release.

So the three copies are structural. What is not acceptable is that they diverge silently, which is exactly what happened here — so `electron-builder-snap-config.test.ts` now parses each workflow in `LINUX_RUNNER_WORKFLOWS` and asserts the step is identical across all three. Parsed rather than text-sliced, so indentation and surrounding YAML comments cannot register as a difference, while the comments *inside* the shell script are compared along with the commands.

## What is deliberately *not* here

The downstream packaging repo carries a larger version of this fix that deletes preinstalled toolchains and adds 12 GB of swap, because its build was being OOM-killed on a **2-core private-repo runner with 7.8 GB of RAM and a 73 GB disk arriving 82% full**.

This repo is public, so the same `ubuntu-22.04` label gets a **4-core runner with 15 GB of RAM and 146 GB of disk at 41% used** — measured in run 34490568415 below. Neither limit is close here, so spending a step on headroom this job does not need would just be noise. I ported it first and then removed it once the runner's actual numbers came back.

## Verification

[Run 34490568415](https://github.com/paranext/paranext-core/actions/runs/34490568415) and [run 34494076141](https://github.com/paranext/paranext-core/actions/runs/34494076141) are green on all three platforms, with the retried snap installs and the LXD snap build succeeding on Linux. Both predate the `test.yml` change and both carried `fail-fast: false`, since removed; nothing else about them differs from this tree.

The `test.yml` copy is the one this repo can actually prove, and this PR proves it: `test.yml` runs on `pull_request`, so this PR's own `Build on ubuntu-22.04, .Net 8.0.x` leg runs the retry helper and builds the snap through it. `publish.yml` is `workflow_dispatch`-only and has no equivalent check short of a release.

Locally: prettier, eslint, YAML parse, `bash -n` on every bash `run:` block, direct tests of the retry helper under `bash -e` (succeeds first try, succeeds after retries, fails the step after exhausting all three), and the new drift assertion — confirmed failing when one `retry` prefix is removed from `publish.yml`, and when the step is renamed in `test.yml`.

https://claude.ai/code/session_012eKf1SVGjFcRcseWoMNUVA

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2797)

### DIFF
--- a/.erb/scripts/electron-builder-snap-config.test.ts
+++ b/.erb/scripts/electron-builder-snap-config.test.ts
@@ -43,6 +43,15 @@ const UBUNTU_RUNNER_BY_BASE: Readonly<Record<string, string>> = {
 /** Workflows that pin the Linux runner the snap is built on, via an `OS_LINUX` env value. */
 const LINUX_RUNNER_WORKFLOWS = ['test.yml', 'package-main.yml', 'publish.yml'];
 
+/**
+ * The step each of those workflows installs snapcraft and LXD under.
+ *
+ * The three copies exist because the step runs BEFORE `actions/checkout`, so it cannot be a local
+ * composite action — nothing is on disk to `uses:` yet. Duplication is therefore the only option;
+ * silent divergence is not, which is what the test below is for.
+ */
+const SNAP_TOOLS_STEP_NAME = 'Install snap tools on Linux';
+
 /** Mount point the snap's launch scripts read the GNOME platform from. */
 const GNOME_PLATFORM_TARGET = '$SNAP/gnome-platform';
 
@@ -63,6 +72,9 @@ type PlugAttributes = {
 
 /** A found plug, keyed by the name it is declared under. */
 type NamedPlug = { key: string; descriptor: PlugAttributes };
+
+/** A workflow step, as far as the comparison below needs to see one. */
+type WorkflowStep = Record<string, unknown>;
 
 /** A plain object, for walking into parsed config and YAML without asserting a shape. */
 function isRecord(value: unknown): value is Record<string, unknown> {
@@ -178,6 +190,30 @@ function readWorkflowLinuxRunners(): { workflow: string; runner: string | undefi
   });
 }
 
+/**
+ * The snap-toolchain step from each snap-building workflow.
+ *
+ * Parsed rather than text-sliced so that indentation and the YAML comments around the step cannot
+ * register as a difference. The shell script is a single scalar, so the comments _inside_ it are
+ * compared along with the commands — which is wanted: a comment that still describes the old
+ * behavior is its own kind of drift.
+ */
+function readSnapToolsSteps(): { workflow: string; step: WorkflowStep | undefined }[] {
+  return LINUX_RUNNER_WORKFLOWS.map((workflow) => {
+    const parsed: unknown = parseYaml(
+      readFileSync(path.join(REPO_ROOT, '.github', 'workflows', workflow), 'utf8'),
+    );
+    const jobs: unknown[] =
+      isRecord(parsed) && isRecord(parsed.jobs) ? Object.values(parsed.jobs) : [];
+    const step = jobs
+      .filter(isRecord)
+      .flatMap((job) => (Array.isArray(job.steps) ? job.steps : []))
+      .filter(isRecord)
+      .find((candidate) => candidate.name === SNAP_TOOLS_STEP_NAME);
+    return { workflow, step };
+  });
+}
+
 describe('electron-builder snap configuration', () => {
   it('declares a snap base whose matching GNOME platform snap and runner are known', () => {
     // Bumping `base` without teaching these maps the new pairings should fail loudly rather than
@@ -207,6 +243,24 @@ describe('electron-builder snap configuration', () => {
     expect(readWorkflowLinuxRunners()).toEqual(
       LINUX_RUNNER_WORKFLOWS.map((workflow) => ({ workflow, runner: expected })),
     );
+  });
+
+  it('installs the snap toolchain identically in every snap-building workflow', () => {
+    // All three legs build the snap through snapcraft and LXD, so all three meet the same snap
+    // store outages, the same auto-refresh mid-pack, and the same failures whose evidence is only
+    // in snapcraft's log. Hardening one copy and not another is invisible in review -- the diff
+    // looks complete -- and leaves the flake live on whichever leg was missed, including the one
+    // behind branch protection.
+    const steps = readSnapToolsSteps();
+
+    // Checked first and separately: if the step were renamed everywhere, every entry below would
+    // be `undefined` and the equality would pass while comparing nothing.
+    expect(steps.filter(({ step }) => !step).map(({ workflow }) => workflow)).toEqual([]);
+
+    // Compared against the first workflow rather than pairwise so a failure's diff names the file
+    // that drifted.
+    const [reference, ...others] = steps;
+    expect(others).toEqual(others.map(({ workflow }) => ({ workflow, step: reference.step })));
   });
 
   it('mounts exactly one GNOME platform plug, named for snap.base', () => {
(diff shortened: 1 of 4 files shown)


lyonsil (commented):
`.github/workflows/test.yml` still carries the un-hardened copy of this block.

The three commands this PR wraps in `retry` are still single-shot at `test.yml:102-104`, with no `refresh.hold` above them and no failure diagnostics below:

```yaml
sudo snap install snapcraft --classic --channel=8.x/stable
sudo snap install lxd
sudo snap refresh lxd
```

That matters more there than in the two files this PR does change, because `test.yml`'s Linux leg is the one behind branch protection. `required_status_checks.contexts` on `main` is:

```
["CodeQL",
 "Build on windows-latest, .Net 8.0.x",
 "Build on ubuntu-22.04, .Net 8.0.x",
 "Build on macos-15, .Net 8.0.x"]
```

`Build on ubuntu-22.04, .Net 8.0.x` is `test.yml:26`'s rendered job name. `package-main.yml`'s job renders as `Main on ...` and runs only on push to `main`; `publish.yml`'s renders as `Release on ...` and runs only on `workflow_dispatch`. Neither is a required check.

That leg does build the snap: `check packaging` (`test.yml:307-312`) has no `if:` guard and runs `npm run package` on every leg, `electron-builder.json5:79` sets `linux.target: ['snap']`, and `test.yml:374-390` renames and uploads `./release/build/*.snap`.

So the store-502 flake this PR exists to absorb can still fail a pull request, on the leg that actually blocks someone, while the two post-merge workflows are hardened.

Two things that make this smaller than it first looks, worth stating alongside it:

- The 502 has never been observed on `test.yml`'s leg. Both runs cited in the description (34488976360 and 34384474185) are `package-main.yml`.
- `test.yml:30-33` already sets `fail-fast: false`, so a flake there costs one re-run rather than cancelling the other two platforms.

Suggested fix: apply the same change to `test.yml`, or extract the step once so the three copies cannot drift again. `.erb/scripts/electron-builder-snap-config.test.ts:44` already declares the authoritative list — `LINUX_RUNNER_WORKFLOWS = ['test.yml', 'package-main.yml', 'publish.yml']`.

tjcouch-sil:
Confirmed, and fixed in 6f76b6c. I checked each claim before acting on it: `test.yml:93-113` did still carry the un-hardened block, `Build on ubuntu-22.04, .Net 8.0.x` is `required_status_checks.contexts` on `main` while `Main on ...` / `Release on ...` are not, `check packaging` has no `if:` guard, and `electron-builder.json5:79` is `target: ['snap']`. Your two caveats check out too, and both are now in the PR body rather than flattened out of it — `test.yml:33` does set `fail-fast: false`, so the blast radius there is one re-run, not three cancelled jobs.

One more argument for `test.yml` that cuts the other way from "riskiest to touch": it is the only copy this repo can prove. `test.yml` runs on `pull_request`, so this PR's own Linux leg runs the retry helper and builds the snap through it. `publish.yml` is `workflow_dispatch`-only and has no equivalent check short of cutting a release.

**On extracting the step once — I don't think that one works, for a reason that isn't obvious from the diff.** In all three workflows `Install snap tools on Linux` runs *before* `actions/checkout`:

| workflow | snap step | checkout |
| --- | --- | --- |
| `test.yml` | 93 | 114 |
| `package-main.yml` | 75 | 96 |
| `publish.yml` | 94 | 115 |

A local composite action (`uses: ./.github/actions/…`, the shape `bump-versions-action` uses) can't resolve at that point — the repo isn't on disk yet. Extraction therefore means reordering checkout ahead of the snap step in three workflows, on top of a flake fix, and `publish.yml`'s copy couldn't be verified without a release run. I'd rather not couple those.

So I took the drift half of your suggestion instead, in the file you pointed at. `electron-builder-snap-config.test.ts` now parses each workflow in `LINUX_RUNNER_WORKFLOWS` and asserts the step is identical across all three — parsed rather than text-sliced, so indentation and the YAML comments *around* the step can't register as a difference, while the comments *inside* the shell script are compared along with the commands. It's falsifiable: it fails when one `retry` prefix is dropped from `publish.yml`, and it names `test.yml` when the step is renamed there. The missing-step check is deliberately separate, so a rename in all three can't pass by comparing three `undefined`s.

The downstream PRs (paratext-10-studio #186 / #187) needed no equivalent change — that repo's `test.yml` is 74 lines of lint and installs no snap tooling at all; `package-main.yml` and `publish.yml` are its only snap builders and both are already in those PRs.

lyonsil (approved):
:lgtm:

> @lyonsil reviewed 4 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2797#-:-P1G2CjM1EsAIUtiEALy:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

Issue #190, opened 2023-05-17
Figure out client/server pattern


tjcouch-sil:
To give us a good idea of how the different options would look, we refactored [`extension-asset.service.ts`](https://github.com/paranext/paranext-core/blob/001fee24c4d9fee8f6c03a9ed93e55719a6bb552/src/shared/services/extension-asset.service.ts) in two ways:

- Matt made client/server service pattern https://github.com/paranext/paranext-core/compare/main...cross-process-server-model
- TJ made `@process` pattern https://github.com/paranext/paranext-core/compare/main...fd04d8a0abe6081f0eb27826d8df9ed84cf5b52a

tjcouch-sil:
Here's an eslint rule to disallow certain imports. We could set this up to enforce import alias scoping in the various folders https://eslint.org/docs/latest/rules/no-restricted-imports

irahopkinson:
> Here's an eslint rule to disallow certain imports. We could set this up to enforce import alias scoping in the various folders https://eslint.org/docs/latest/rules/no-restricted-imports

Great, I added this to the **Code Style Evaluation** doc, actually I added the TS version (https://typescript-eslint.io/rules/no-restricted-imports) which has a useful option to `allowTypeImports`.

PR #2798, opened 2026-09-10
feat(notices): let a downstream product generate its notices with a policy overlay

lyonsil:
## Summary

Lets a downstream product built from this repository (Paratext 10 Studio, `paranext/paratext-10-studio`) generate its own third-party notices with this generator instead of forking it. The product names a policy overlay through `NOTICES_POLICY_OVERLAY`; the generator merges it over `notices-policy.json` and describes the product by name from the overlay's `product` block. Two new policy tables let such a product disclose what core cannot see:

- **`separatePrograms`**: third-party programs redistributed as separate executables (Studio ships Mercurial), each with a reviewed determination, per-platform deliveries, and the components bundled inside each delivery, credited under their own copyright notices rather than the program's.
- **`externalExtensions`**: extensions packed from outside this repository, recorded as a stated omission until their bundles can be itemized.

An override can link a NuGet package to a separate program (`separateProgram`), bound to that program's reviewed identifiers. The generator now fails closed when `productName` is not Platform.Bible and no product block is declared.

**Both new tables ship empty in core**, so neither adds a section to core's document. Core's `THIRD-PARTY-NOTICES.md` changes in exactly one place - the preamble sentence that names the two new categories alongside the five existing ones - and `THIRD-PARTY-NOTICES.lock.json`'s `documentSha256` moves with it. No row, section or license text of core's own changes. Five identifiers are admitted to `allowed` so the committed corpus holds their texts for the downstream Mercurial bundles: `PSF-2.0`, `OpenSSL`, `blessing`, `TCL`, `ZPL-2.1`. Nothing in core uses them.

ADR: `adr-notices-overlay-for-downstream-products`, with amendments to the existing notices entries. The generator README and the root README describe the overlay and the two tables.

## Verification

- `npm run test:core -- --run .erb/scripts/third-party-notices`: 775 passed, 1 skipped.
- `npm run verify:third-party-notices:document`: verified; the committed document is unchanged.
- `npm run typecheck:erb` clean; `electron-builder-snap-config.test.ts` 5/5.
- Core's full `--verify` (Ruby + dotnet restore) was **not** run locally on this branch; CI runs it. The downstream pipeline ran the full verify in a patched clone at this branch's head with its overlay ("Verified 310 packages against the committed lock"), so the overlay code path has been exercised end to end.

## Known residuals, parked

- The program-level heading and Terms line interpolate reviewer-authored policy strings unescaped, the same shape as the pre-existing copied-platform-library section.
- `externalExtensionFolders` does not dedupe a folder mapped twice; names dedupe downstream.
- `lint:scripts` hits a pre-existing unrelated error (missing generated `buildInfo.json`).

The downstream consumer is a PR in `paranext/paratext-10-studio`, which stays a draft until this merges.

## Also in this PR

- **Terms of Service as an HTML document, opened in a window of the application's own.** `TERMS-OF-SERVICE.md` is replaced by a self-contained `TERMS-OF-SERVICE.html`. **This is not a format-only conversion: the document is also revised from the 14 August 2026 text to the 11 September 2026 text**, which is the current authoritative revision from the director, attorney-reviewed, and the one now deployed at `registry.paratext.org/terms`. Because the file extension changed, git renders this as a whole-file delete plus add, so the revision is invisible in the diff. It arrived in two steps, and **both are invisible**:

  - **14 August → 9 September.** New `Resource` / `Restricted Resource` definitions and the Section 7 obligations that go with them; `Paratext` widened to cover Paratext 9 and 10 (folding the PT10-specific 3.B.3-3.B.5 subsections into the general Section 3 clauses); 3.B.1/3.B.2 reattributed from "Paratext" to "SIL and UBSA"; the Service Continuity Backup date moved from 1 October 2026 to 15 October 2026.
  - **9 September → 11 September** (commit `b7d01b7`, regenerated from the updated `Paratext Terms of Service.docx`; content only - the head, the stylesheet and the Content-Security-Policy are untouched). The policy documents moved host, so every `paratext.org/terms`, `/privacy-policy` and `/acceptable-use` reference moved to `registry.paratext.org` - 17 references in all, 8 of them hrefs. Section 2 is alphabetized, as the source document is, and gains a "Backup Agent" definition. The Service Continuity Backup is restated throughout Sections 17 and 18 in terms of that Backup Agent holding a *current* copy rather than an agent holding a snapshot, and 18.6's opt-out paragraph is rewritten with explicit windows (deletion within 31 days; superseded backup copies retained up to 60 days). 3.B.1 drops the general "portions of the source code may be used under the AGPL" sentence and instead points at the license notices that identify the AGPL Components. 5.5's retention wording is simplified. Section 4 regains a word space the previous conversion collapsed.

  That second step was verified against both sources: tag-stripped and whitespace-collapsed, the document's 178 paragraphs are identical to the docx's, and its 177 body blocks are identical to the text the registry serves (rendered from a client-side template behind a feature flag, so it was read from the bundle and confirmed in a browser).

  `LICENSING.md` follows the 3.B.2 rewording. The switch to HTML is product team input. The About dialog's `platform.openTermsOfService` command now shows it in a sandboxed Electron window (no preload, no Node access, one window at a time) instead of handing the file to the OS, which on Windows had no handler for markdown. Every link in the document leaves through main's `openExternal`, whether it would open a new window or navigate this one, so the window only ever shows the document. That policy admits `https:`, `mailto:` and the application's own scheme, and all twelve links in the document are `https:` (11) or `mailto:` (1); the page carries a `default-src 'none'; style-src 'unsafe-inline'` Content-Security-Policy, and `derived-invariants.test.ts` pins both. Whoever regenerates the document should keep them. The packaged license declaration, extraResources, LICENSING.md, the localized error strings, Prettier's ignore lists and the file-name invariant test follow the rename. **The earlier note that `paratext.org/acceptable-use` answered 404 is moot** - all three `registry.paratext.org` policy paths return 200.
- **Two follow-ups from the final review of #2654.** The renderer's web view module map is typed `ReadonlyMap<string, unknown>`, and a renderer twin of the extension host's dispatch-shape test reads the shim's source, so neither a later `moduleMap.set` nor a name comparison in the shim can widen the allowlist unnoticed. Typing the values also surfaced that the similar-API hint was handed the module value rather than the requested name; it receives the name now. `formatExtensionsRoot` stamps the `license` field in `extensions/package.json` after a template merge through a new `stampExtensionsRootLicense(root)`, which has a test for the merge that reverts the field.

Verification for these: `src/main/terms-of-service-window.test.ts` (7 tests, Electron mocked), the notices suite at 786 passed after the invariant retarget, the module-allowlist suite at 6, `git.util` at 12; `typecheck:core` reports only two pre-existing errors from unbuilt artifacts in the worktree. The window was not exercised in a running application in this environment: the About dialog's license link is the path to click.

---

<details>
<summary><b>Pre-review summary</b> (/review-paratext, with author interview) - findings, what was fixed in review, and suggested review focus</summary>

# Code Review Summary

**Branch**: feat/notices-downstream-products

**Base**: origin/main

**Date**: 2026-09-10

**Files changed**: 48

## Overview

The driving change lets a downstream product built from this repository — Paratext 10 Studio — generate its own third-party notices with this repository's generator instead of forking it. The product names a policy overlay through `NOTICES_POLICY_OVERLAY`; the generator merges it over the committed `notices-policy.json`, refusing an unknown key rather than dropping it, and describes the product by name from the overlay's `product` block. Two new policy tables let such a product disclose what core cannot see: `separatePrograms` (third-party programs redistributed as separate executables — Studio ships Mercurial) and `externalExtensions` (extensions packed from other repositories, recorded as a stated omission). Both ship empty in core, so neither adds a section to core's document; core's `THIRD-PARTY-NOTICES.md` changes only in the preamble sentence that names the two new categories.

Two unrelated bodies of work ride along in the same branch, which the author explained as deliberate: the Terms of Service moves from Markdown to a self-contained HTML document opened in a sandboxed Electron window the application owns (product-team direction, taken while these files were already open, and carrying a newly-revised text), and two follow-ups from the final review of #2654 (the renderer's web view module map typed `ReadonlyMap<string, unknown>` with a renderer twin of the extension host's dispatch-shape test, and `formatExtensionsRoot` stamping the `license` field after a template merge).

The analysis found one critical and ten important issues. Six were fixed during the review, four were dismissed with the author's reasoning, and the remainder are minor. Reviewers should note the Critical finding in particular: it is invisible in the diff.

## API Changes

- `@papi/core` (`papi.d.ts`) / `src/declarations/papi-shared-types.ts`: TSDoc for the existing `'platform.openTermsOfService'` command changed. Signature unchanged (`() => Promise<void>`); documented behavior moved from "handed to the operating system's Markdown handler, revealed in the file manager on failure" to "self-contained HTML shown in a window the application owns, one at a time", and the `@throws` contract narrowed from "could not be opened, including the reveal-fallback case" to "could not be loaded".
- `lib/platform-bible-react/`, `lib/platform-bible-utils/`, `extensions/src/*/src/types/*.d.ts`: No changes.
- Non-`@papi/` / internal surfaces (for completeness, not public API):
  - `extensions/lib/git.util.ts`: added `stampExtensionsRootLicense(root?: string): Promise<void>`. `formatExtensionsRoot()` keeps its signature but now also stamps `license` in `extensions/package.json`.
  - `src/main/terms-of-service-window.ts` (new module): added `openTermsOfServiceWindow`, `TERMS_OF_SERVICE_FILE_NAME`, type `OpenExternalLink`. `main.ts`'s module-private `openTermsOfService` and its `TERMS_OF_SERVICE_FILE_NAME` were removed in favor of these.
  - `.erb/scripts/third-party-notices/`: added modules `product.ts`, `separate-programs.ts`, `external-extensions.ts`; added exports `OVERLAY_ENV`, `overlayFromEnv`, `mergePolicies`, `PLACEHOLDER_TEMPLATE_VALUE` (newly exported, pattern widened), `ELECTRON_BUILDER`, `nugetNote`; `loadPolicy` gained an optional second parameter `overlayFile` defaulting from the environment; `Policy` gained `separatePrograms`, `externalExtensions`, `product` (+ their `*Note` fields); `types.ts` added `ProductBlock`, `SeparateProgram`, `ProgramDelivery`, `BundledComponent`, `ExternalExtension`; `Override` gained `separateProgram`.
  - **Added during this review**: `assertSeparateProgramLinksRecorded(overrides, programs)` exported from `separate-programs.ts`; `MERGE_KINDS` exported from `policy.ts`; `PackagingConfig` widened with `nsis`, `dmg` and `directories`.
  - `assets/localization/en.json` / `es.json`: added `%about_ariaLabel_opensTermsOfService_2%` and `%about_error_couldNotOpenTermsOfService_2%`; the two originals retained and marked deprecated in `metadata.json`.
  - Repository root: `TERMS-OF-SERVICE.md` removed, `TERMS-OF-SERVICE.html` added; `release/app/package.json` `license` now `SEE LICENSE IN TERMS-OF-SERVICE.html`.

- [x] **`TERMS-OF-SERVICE.html` is not a format-only conversion — it carries a newer revision of the legal text, and the delta is invisible in the diff.** Verified directly: "last updated" moves from 14th August 2026 to 9th September 2026; "Restricted Resource" goes from 0 to 3 occurrences; the Service Continuity Backup date moves from 1 October 2026 to 15 October 2026 (both occurrences). A sentence-by-sentence diff also found new `"Resource"` / `"Restricted Resource"` definitions with a rewritten Section 7 adding obligations that did not exist before (comply with Resource license terms, no altering unless licensed, withdrawal on notice, Restricted-Resource base-text and commercial-use limits); `"Paratext"` widened to "the Paratext client applications (including Paratext 9 and 10)", folding the PT10-specific 3.B.3/3.B.4/3.B.5 subsections into the general Section 3 clauses; 3.B.1/3.B.2 reattributed from "made available by Paratext" to "made available by SIL and UBSA". Because the file extension changed, git renders this as a whole-file delete plus add, so a reviewer cannot see any of it. _(fixed during review: the PR description now carries a bolded callout stating that this is not a format-only conversion, naming the substantive changes and why they are invisible in the diff. The revision itself is intentional and authoritative — see Interview Notes.)_ **Superseded after the review: commit `b7d01b7` carries the document forward again, from 9 September to the 11 September 2026 text now deployed at `registry.paratext.org/terms`. That second delta is enumerated in "Also in this PR" above, and it is the revision a reviewer should check against — not the 9 September text this finding describes.**

- [x] **`LICENSING.md:239` quoted a phrase that no longer exists in the shipped Terms.** It read `what section 3.B.1 means by "the AGPL Components identified by Paratext"`; the new 3.B.2 reads "identified by SIL and UBSA in the license notices accompanying the Paratext 10 application". `LICENSING.md` ships in `resources/` beside `TERMS-OF-SERVICE.html`, so a user receives both and can compare them. The same stale quote appeared at `.context/standards/Architecture-Decisions.md:1327`. _(fixed during review: both sites now quote the current wording. The section number was also corrected — the phrase is in 3.B.2, not 3.B.1, which I confirmed was already a misattribution in the pre-branch document. The decision-log site carries an inline `**amended 2026-09-10:**` note matching the convention this branch already uses there.)_
- [x] **The new ADR stated something this branch contradicts.** `Architecture-Decisions.md:1869` read "This repository ships both tables empty and its own document is byte-identical." Verified false: `THIRD-PARTY-NOTICES.md`'s second preamble paragraph is reworded to name the two new categories, and `THIRD-PARTY-NOTICES.lock.json`'s `documentSha256` moves from `b811f4d3…` to `5e83e9b0…`. The decision log is add-only and permanent, and the entry's own Consequences section discloses that five identifiers were added to `allowed` for a downstream product — a reviewer trusting "byte-identical" two paragraphs earlier would not re-derive that. The same claim led the PR description, where it told a reviewer they could skip the notices diff. _(fixed during review: the ADR sentence now says both tables ship empty so neither adds a section, that the document changes only in the preamble sentence, and that the lock's `documentSha256` moves with it. No amendment note, since the entry is new on this branch and unmerged. The PR body was corrected to match.)_
- [x] **`nugetNote` could put an unbacked cross-reference into a legal document.** `main.ts:385` read `override.separateProgram` unconditionally and emitted `Redistributed as the separate program "X" - see "Third-party programs redistributed as separate executables"` with no validation. The only check, `applyOverride` (`policy.ts:1230-1242`), is unreachable for most rows: `readInstruments` sets `overridable` only when `!declared.ok && !identifiedText` (`policy.ts:1503`), and `main.ts:1644` calls `applyOverride` only for a blocked verdict with no exception — so a NuGet package cleared by its own declared license carried the sentence with the name never checked, and the run exited 0. The docstring acknowledged the gap ("belongs with the other whole-set assertions in `buildReport`, and is not done") and, unlike the comparable deferred gap at `separate-programs.ts:24`, named no ticket. _(fixed during review: added `assertSeparateProgramLinksRecorded(overrides, programs)`, called from `buildReport` beside the other whole-set assertions and from the `--verify` path; uses `Object.hasOwn` so a link spelled `toString` cannot resolve against `Object.prototype`, and trims the name as `applyOverride` and `nugetNote` do. Six new tests including the motivating case. The docstring's "is not done" paragraph was replaced with what now holds. Mutation-checked: flipping the guard to a no-op kills 4 of the new tests.)_
- [x] **The Terms document's CSP and https-only link policy were enforced only by a comment.** `TERMS-OF-SERVICE.html:6` asks a regenerator to keep both; nothing checked either, and the file is hand-regenerated and prettier-ignored. A dropped `default-src 'none'` or a reintroduced `http://` link would surface only at click time, where `terms-of-service-window.ts` logs the `openExternal` rejection rather than surfacing it — so the user gets a link that silently does nothing. Out of step with the rest of the branch, which pins the *filename* in four places with a test. _(fixed during review: three assertions in `derived-invariants.test.ts`, derived from the manifest's filename like the block above so a rename cannot leave them checking a document the application no longer ships — the CSP meta tag is present and denies every default source; every `href` is `https:`, `mailto:`, or a pure fragment (fragments left permitted so a table of contents stays open to the document's authors, since Electron fires no `will-navigate` for same-document navigation); and a guard that the href set is non-empty so the previous case cannot pass on absence. Both mutation-checked against a stripped CSP and a downgraded link.)_
- [x] **`MERGE_KINDS`' docstring claimed a guarantee the type system does not give.** It said a new `Policy` field would otherwise be "silently taken from the base, with the overlay's value dropped… `satisfies Record<keyof Policy, MergeKind>` makes that a compile error instead." `satisfies` forces the new field to *declare* a kind; it cannot force `mergePolicies` to *implement* it, because the merge spells every field into a literal starting with `...base`. Worse, `OVERLAY_KEYS` is derived from `MERGE_KINDS`, so such a key would be **accepted** by the unknown-key refusal and then silently dropped — a downstream declaration going undisclosed with the run exiting 0, the exact outcome the overlay design exists to prevent. All 15 non-`base-only` keys are wired today, so it was latent, not live. Two of the four analysis passes had praised this as a real gate; that praise was overstated. _(fixed during review: the docstring now states what `satisfies` does and does not buy; `MERGE_KINDS` is exported and `policy.test.ts` walks it, contributing a sentinel for each non-`base-only` field one key at a time so a failure names the field, plus a guard that all four contributable kinds are still represented. Mutation-checked by unwiring `separatePrograms` from the merge: `tsc` exits 0 — confirming the hole is real — and the new test fails naming the field.)_
- [x] **A stranded docstring in `render.ts`.** Two JSDoc blocks were stacked and only the nearer bound, so the gating explanation ("Gated on `platform-lexical-tools` being among the extensions the installer packs… an ungated section would go on making UBS copyright claims… for content no installer carries") sat on `lexicalDatabasePermission`, which does no gating, while `pushLexicalDatabaseSection` — which holds the actual guard — was the only section-pusher in the file with no docstring. _(fixed during review: moved back onto `pushLexicalDatabaseSection`.)_
- [ ] ~~Five identifiers (`PSF-2.0`, `OpenSSL`, `blessing`, `TCL`, `ZPL-2.1`) were added to core's `allowed` list so the committed corpus holds texts the downstream Mercurial bundles need. Because `allowed` is the same list `classify()` consults for core's own npm/NuGet graph, a future core dependency declaring one of these now clears classification with no per-package review, and nothing pins that they stay unreachable from core's own graph.~~ _(Author: accepted consequence — the ADR already discloses it. Suggested alternatives were a separate `corpusOnly` list not consulted by `classify`, or a test asserting no core package resolves to one of the five.)_
- [ ] ~~The Terms window gives the reader no find, no zoom, and no menu for a 6,294-word / 20-section legal document. `terms-of-service-window.ts:71` calls `removeMenu()`, registers no `before-input-event`, and never touches `webContents.setZoomFactor` or subscribes to `platform.zoomFactor` — unlike application windows (`main.ts:1558-1585`). That setting's own description says "Applies to the entire application", which this window contradicts, and body copy is set below default at `p { font-size: 0.95rem }`. A net loss against the previous `shell.openPath` behavior, where the OS handler supplied find, zoom and print.~~ _(Author: the move to HTML was directed by the product team with no further guidance; readers can copy-paste elsewhere if they want, hardly anyone reads these, and most people would feed it into an AI anyway. Leave as is.)_
- [ ] ~~The document is pinned to a light palette (`color-scheme: light`, `background: #f5f5f3`, `main { background: #ffffff }`, `color: #1c1c1c`), so a user running the app in dark mode gets a bright 900x800 window. The in-file rationale — one rendering at the contrast it was proofed at — is legitimate but does not require *light*; a dark palette proofed the same way would satisfy it.~~ _(Author: leave as is.)_
- [ ] ~~The window title is hardcoded English. No `title` option is set, so the title bar, taskbar entry and window switcher show `<title>Paratext Terms of Service</title>` from the HTML, while the link that opens it is localized (`%about_licenseLabel_termsOfService%` is "Términos de servicio de Paratext" in `es.json`). Main already uses `localizationService` (`platform-macos-menubar.util.ts:19`).~~ _(Author: the document itself has not been translated and it is unclear whether it will be; a localized title would promise a translated document. Leave as is for now.)_

### Minor — Consider

- [x] **`render.ts:931` re-inlined `.replace(/\s+/g, ' ').trim()` instead of using `inlineNotice`**, the helper added at line 149 in this same change for exactly that, leaving three copies of the idiom in one file. _(fixed during review: uses `inlineNotice`; rendered output unchanged.)_
- [x] **`pushPreamble`'s product and no-product arms repeated four byte-identical lines of legal prose** (`render.ts:569-596`), which an editor fixing wording in one arm would not see in the other. _(fixed during review: the shared opening is written once, with only the differing tail branching. Verified byte-for-byte: all three arms — no-product, product, and Paratext product — render identically to the pre-refactor code.)_
- [x] **`separate-programs.ts:28` and `external-extensions.ts:24` each declared `const PLACEHOLDER = PLACEHOLDER_TEMPLATE_VALUE`**, a rename-only alias duplicated across two modules, while `product.ts:70` used the import directly. _(fixed during review: aliases dropped; all three modules use the imported name.)_
- [x] **`packaging.test.ts:8` parsed `electron-builder.json5` itself** while this change introduced `readPackagingConfig` as the canonical reader, leaving two parsers for one file. _(fixed during review: adopted `readPackagingConfig(ELECTRON_BUILDER)`; `PackagingConfig` widened with the `nsis`, `dmg` and `directories` fields the test reads, and the local `unknown`-narrowing `resourcePath` helper replaced with a typed one.)_
- [x] **`mergePolicies`' unknown-key error conflated two different mistakes** — it told the author of a `base-only` key that the policy "has no such field for" it, sending them hunting a typo that does not exist. _(fixed during review: the two cases throw separately, the reserved-key message explaining that the field exists and is spelled correctly but is reserved to the committed file. Covered by a new test asserting the reserved case does not mention spelling.)_
- [x] **`terms-of-service-window.ts:29` understated the link policy** — the `@param openExternal` doc said main admits only `https:` and `mailto:`, but `main.ts:317-325` also admits `${APP_URI_SCHEME}://`. Nothing is exploitable today, but a security-relevant comment that understates the accepted scheme set is the kind that gets relied on later. _(fixed during review: names all three schemes and states that this window inherits main's policy rather than stating its own.)_
- [x] **Two backward-facing comments narrating what this PR changed** — `global-this-web-view.model.test.ts:37-38` ("`getModuleSimilarApiMessage` was handed the (always undefined) module rather than the name, so every one of these errors read…") and `extension.service.module-allowlist.test.ts:68-70` ("`/moduleName === /` matched exactly that text with exactly that spacing, so … all passed"). _(fixed during review: both rewritten forward-facing. The rest of the `NAME_COMPARISON_DISPATCH` docstring — what the pattern still misses, why a literal on the other side is the dangerous shape — was kept.)_
- [x] **`assets/localization/metadata.json:8` was out of byte order**, the new deprecation entry sitting before `%about_db_ip_attribution_intro%` in an otherwise strictly sorted file. _(fixed during review: moved into position; the whole file verified sorted.)_
- [x] **The About dialog still rendered `ExternalLink`** though the document now opens in a window the application owns, so the icon's conventional "leaves this application" meaning no longer matched the behavior. _(fixed during review, per the author's direction: the trigger is now a plain link — icon and its `lucide-react` import removed, and the comment explaining the icon's accessibility handling rewritten. The `aria-describedby` / `sr-only` arrangement is unchanged.)_
- [x] **`%about_error_couldNotOpenTermsOfService_2%` named a location but no next step**, and this change removed the `shell.showItemInFolder` reveal that used to be the fallback (`main.ts:356-365`), making this message the only recovery path. _(fixed during review: reworded in both `en.json` and `es.json` to point the reader at their web browser — possible only now that the document is HTML. Edited in place rather than versioned again, since the `_2` keys are new on this branch and have never shipped; the retired key's deprecation note was extended so a translator sees why the replacement differs.)_
- [ ] ~~`src/main/terms-of-service-window.ts` carries no `<type>` segment in its filename, which the style guide's `<feature>.<type>.{ts,tsx}` rule asks for; 10 of the 14 non-entry-point modules in `src/main/` use `.util.ts`.~~ _(Consistency nit with precedent both ways — `startup-tasks.ts` and `shutdown-tasks.ts` use the bare form. Left as is.)_
- [ ] ~~The Terms of Service delivery change (OS handler → application-owned sandboxed `BrowserWindow`, parenting to satisfy `window-all-closed`, CSP + all-navigation-denied link policy) is recorded only as three inline amendments to existing ADR entries, though it had real alternatives (keep `shell.openPath`, render inside the renderer, ship a PDF).~~ _(Left as is: the change rode along with the notices work rather than being its own decision. See Interview Notes.)_
- [ ] ~~`TERMS-OF-SERVICE.html` contains 26 "Section N" cross-references with no `id` anchors, no links between them, and no table of contents, so reaching a referenced section means scrolling and counting headings.~~ _(Author: do not change anything here. Note that the new href assertion deliberately permits fragment links, so this stays open to the document's authors later.)_
- [ ] ~~`a:hover, a:focus { text-decoration: none; }` removes the underline exactly when a keyboard user lands on a link. Chromium's default focus ring still applies, so not a blocker, but reversing the underline affordance on focus is unconventional.~~ _(Author: leave as is.)_
- [ ] ~~`src/renderer/global-this-web-view.model.test.ts` boots the renderer network stack on import, so every run emits `ECONNREFUSED 127.0.0.1:8876` and four buffered-network-event warnings.~~ _(Left as is on my recommendation, which the author accepted. Measured: the connect fails once with no retry loop and the 8 tests run in 3ms, so the noise costs nothing. The only clean way to silence it is mocking `papi-frontend.service` — one of the modules the shim under test exists to hand out — so it would substitute a mock for part of the thing being tested to quiet a log line. The Linux CI trouble this repo previously hit was a different mechanism, jsdom dumping Tailwind CSS on a parse failure.)_

#### Shared Regions Modified

- [n/a] `extensions/lib/git.util.ts` — carries `#region shared with .../paranext-extension-template/blob/main/lib/git.util.ts` at lines 10-17 and 93-161; this branch's hunks are at 375-384 and 621-645, outside both ranges. Confirmed independently by two analysis passes.

#### Extension Config Changes

- [n/a] `extensions/lib/git.util.ts`, `extensions/lib/git.util.test.ts` — `stampExtensionsRootLicense` exists to correct the MIT license a multi-extension template merge writes into `extensions/package.json` back to this repository's AGPL. Core-specific by construction; propagating it upstream would be wrong.
- [n/a] `electron-builder.json5`, `release/app/package.json`, `release/app/package-lock.json`, `.prettierignore`, `.prettierignorerun` — top-level build/config files outside `extensions/`, all carrying the same `TERMS-OF-SERVICE.md` → `.html` rename.

## Positive Observations

- `src/renderer/global-this-web-view.model.ts:95` fixes a real defect: `getModuleSimilarApiMessage(module)` was passing the always-`undefined` lookup result instead of `moduleName`, so every rejection read `Rejected require('undefined')`. It type-checked only because the module record is typed `any`. Pinned by a behavioral test that runs the shim through `globalThis.webViewRequire`, including `Object.prototype` members — not only by the source-grep backstop.
- The localized-string change follows the project's convention exactly: both strings whose *meaning* changed got new `_2` keys rather than in-place edits, the originals were retained with dated `deprecationInfo` naming their replacements, and real Spanish translations landed at the same time, so no locale silently falls back to English.
- The `TERMS-OF-SERVICE.md` → `.html` rename is carried through every one of its independent spellings — `electron-builder.json5`, `release/app/package.json`, its lockfile, `LICENSING.md`, `.prettierignore`/`.prettierignorerun`, the localized error text — and `derived-invariants.test.ts` grew a lockfile assertion, closing the exact hole (npm rewriting `packages[""].license`) that would otherwise have failed CI after packaging rather than in tests.
- The Terms window is locked down sensibly: `sandbox: true`, `contextIsolation: true`, `nodeIntegration: false`, no preload, `default-src 'none'` CSP, `setWindowOpenHandler` denying and `will-navigate` prevented with both routed through `openExternal`. Fragment links still work because Electron does not fire `will-navigate` for same-document navigations.
- The unknown-key refusal in `mergePolicies` is the right shape for a fail-closed merge: a misspelled `seperatePrograms` is refused outright rather than leaving the committed empty table in place and letting every gate below pass over it.
- `mergeTable`, `assertExternalExtensionsRecorded` and the new `applyOverride` branch all use `Object.hasOwn` rather than bare indexing, each with a comment saying why — the same reasoning behind the `moduleMap` `Map` in the renderer shim. Consistent treatment of one hazard across four sites.
- The new ADR is inserted in correct byte-order slug position (verified against `LC_ALL=C sort` over all `## adr-` slugs), and the three ADRs it affects were amended in place rather than rewritten.
- `product.ts`, `separate-programs.ts` and `external-extensions.ts` each answer one question, match the directory's floating module-header convention, and were added to the README's module map and instrument table in the same change.
- Test coverage for the new generator modules is thorough and behavioural, and `policy.test.ts` calls `loadPolicy(POLICY_PATH, undefined)` explicitly so a stray `NOTICES_POLICY_OVERLAY` in the environment cannot change what "the shipped policy" means in a test.
- The About dialog's accessible-name handling is careful and correct: the supplementary description is on a `tw:sr-only` sibling referenced by `aria-describedby` rather than an `aria-label` on the button, so name-from-contents leaves the visible text as the whole accessible name (WCAG 2.5.3). The failure path is announced with `role="alert"`, and the trigger carries no ellipsis, which is right per `Guidelines/Ellipses`.
- `papi.d.ts` appears properly regenerated, not hand-edited: the emitted block is byte-identical to the source declaration modulo the generator's stripping of an internal comment.

## Interview Notes

**Stated purpose.** The driving force was preparing `paratext-10-studio` to derive and present its own third-party license details. The author chose an overlay over a downstream generator because the generator "was already complex to put together in core" and recreating it downstream would mean either duplicating the pipeline or depending on its internal module API across a clone boundary; P10 instead reuses the existing infrastructure with its own patches on top.

**Why three areas share one branch.** The author was explicit and unprompted: the notices overlay is the driver; the Terms of Service work came in because the document had just been revised and these files were already open; the two #2654 follow-ups were "intended to be part of the next update I knew was coming."

**On the Terms of Service revision (Critical finding).** The author confirmed the newer text is intentional and authoritative: the document is sourced from their director, including review with an attorney, and the version in this branch is the latest. The move from Markdown to HTML was product team input. This is a confirmed sign-off, not an assumption — but note that the author's answer establishes provenance, not a line-by-line comparison against the authoritative source; no one in this review verified the shipped HTML matches that source verbatim, and the repository holds nothing to check it against.

**On the Terms window's UX (three dismissed findings).** The author's reasoning was that the HTML move came with no guidance beyond the format, readers can copy the text elsewhere if they want it in another tool, "hardly anyone reads these", and most readers would feed the document into an AI. For the untranslated window title specifically, the author noted the document itself has not been translated and may not be, so a localized title would promise a translated document. These are deliberate product judgments, recorded as such.

**On the widened allowlist.** Accepted as a disclosed consequence rather than mitigated. A reviewer who wants a belt-and-braces option has two on the table: a `corpusOnly` list not consulted by `classify`, or a test asserting no core package resolves to one of the five identifiers.

**Areas the author could not explain: none.** The author answered every question directly, made decisions rather than deferring, and gave reasoning for each dismissal. No deferrals to AI, and no expressed uncertainty about how their own changes work.

**One disagreement worth recording.** Two of the four analysis passes praised `MERGE_KINDS`' `satisfies` clause as a genuine compile-time gate; the third said its docstring overclaimed. The third was right, which I confirmed by unwiring a field and observing `tsc` exit 0. This is a reminder that a green analysis pass is not evidence a claim in a comment is true.

## In-Review Quality Check

All checks were run after the in-review changes, and all pass:

- **`npm run typecheck`** — all four legs (`core`, `erb`, `e2e`, `workspaces`) exit 0. Note this is cleaner than the PR body reports; the two pre-existing `typecheck:core` errors from unbuilt worktree artifacts are no longer present.
- **`npm run lint`** — exit 0, **0 errors**. One pre-existing unrelated warning remains (`import/no-named-as-default` on `UnsubscriberAsyncList`). The pre-existing `lint:scripts` failure the PR body mentions (missing generated `buildInfo.json`) did not reproduce.
- **`npm test`** — **9,253 passed, 2 skipped, 0 failed** across 654 test files in all workspaces. One `no-shadow` lint error introduced by a review fix was caught and corrected before staging.
- **Formatting** — `prettier --write` applied to changed files only, not the workspace.
- **`npm run verify:third-party-notices:document`** — verified; `sha256sum THIRD-PARTY-NOTICES.md` matches the lock's `documentSha256` (`5e83e9b0…`), and neither the document nor the lock is modified in the working tree.
- **Render-output equivalence** — because the preamble refactor touches prose the tests only assert at its first and last lines, all three render arms (no-product, product, Paratext product) were rendered with the pre-refactor `render.ts` and with the refactored one and diffed: byte-for-byte identical.
- **Mutation checks** — each new assertion was verified falsifiable rather than assumed: no-op'ing the separate-program link guard kills 4 tests; unwiring `separatePrograms` from the merge kills the `MERGE_KINDS` walk (while `tsc` still exits 0); stripping the CSP tag and downgrading one link to `http://` each kill their respective document assertions.

No unfixable failures.

## Suggested Review Focus

- [ ] **The Terms of Service text revision is the one thing in this PR that cannot be reviewed from the diff.** Git shows a delete plus an add. Someone should compare `TERMS-OF-SERVICE.html` against the authoritative **11 September 2026** source verbatim. The author has confirmed provenance (director, attorney-reviewed) and, for the 11 September step, recorded a paragraph-level comparison against both the source `.docx` (178 paragraphs) and the text `registry.paratext.org/terms` serves (177 body blocks) — but the repository holds neither source, so that comparison is not reproducible from this PR. The substantive changes of both steps are enumerated in "Also in this PR".
- [ ] **Does anything else in the tree quote or depend on the old Terms wording?** `LICENSING.md` and one ADR site were found and fixed in review, and nothing pins such quotes — `derived-invariants.test.ts`'s "spelled the same in all three places" suite only checks the *filename*. A cheap hardening would be asserting that each phrase `LICENSING.md` puts in quotation marks is present in `TERMS-OF-SERVICE.html`.
- [ ] **The widened `allowed` list.** Five identifiers now clear `classify()` for core's own dependency graph purely to serve a downstream product's corpus. Accepted by the author as a disclosed consequence; worth a second opinion on whether a `corpusOnly` split is warranted before this pattern is repeated for the next downstream product.
- [ ] **Three unrelated bodies of work in one branch.** The author's rationale is recorded above and is reasonable, but the ToS delivery change in particular has its own security surface (a new `BrowserWindow`, a link policy, a CSP) and is recorded only as inline ADR amendments rather than its own entry. Worth agreeing whether that is the right level of record.
- [ ] **`terms-of-service-window.ts` has not been exercised in a running application** in the author's environment — the About dialog's license link is the path to click. The unit tests mock Electron.
- [x] ~~**`paratext.org/acceptable-use` answers 404 today**~~ — resolved by the 11 September revision: the three policy documents now live at `registry.paratext.org/terms`, `/privacy-policy` and `/acceptable-use`, and all three return 200.

</details>

AI-assisted - [implementation session](https://claude.ai/code/session_013ZzwFT7hwQUaMhcuoss9vy), [review session](https://claude.ai/code/session_01NWS8jHWxtyW5u13GJxhgbC), [review-response session](https://claude.ai/code/session_01R29RWeUJDWU3ijGEnTox4A)

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2798)


tjcouch-sil (commented):
## Review — 9 items

Reviewed at `b7d01b75d13`. This is a filtered list: I started from a much longer sweep and cut everything that needed a contrived config to reach, or that was a wording/error-message nit. What's left is ordered by likelihood × consequence. A few notes on what I deliberately *didn't* raise are at the bottom, so you know they were considered rather than missed.

---

### 1. The Terms of Service window can't copy, select, or zoom — `src/main/terms-of-service-window.ts:64`

`removeMenu()` is where Electron registers the `copy`/`selectAll` accelerators on Windows and Linux, and the app's zoom/F12/Esc handlers are registered per-`createWindow` (`main.ts:989`, `:1423`), so none of them reach this window. The result is that every Windows and Linux user who tries to copy a line out of a 6,584-word document they're told they are legally bound by gets nothing.

**Suggested fix:** `newWindow.setMenu(Menu.buildFromTemplate([{ role: 'editMenu' }, { role: 'viewMenu' }]))` instead of `removeMenu()`. Note `autoHideMenuBar: true` three lines up is inert either way.

Two more while you're in this file, both cheap:
- No `show: false` + `ready-to-show` and no `backgroundColor`, so a blank white 900×800 window sits on screen for the whole load and flashes before vanishing on failure.
- The old `logger.warn` on the failure path was dropped with nothing replacing it, so a failed load — including one invoked from the extension host or any PAPI client — leaves no trace of the path that was tried.

This is the highest-certainty item on the list: not an edge case, just everyone who tries to copy.

---

### 2. The PR body describes the wrong revision of the Terms document

The shipped `TERMS-OF-SERVICE.html` says "last updated on 11th September 2026"; the PR body enumerates the 9 September deltas and tells reviewers that is what to check. The final commit also:

- moved **all nine** `paratext.org/terms | /privacy-policy | /acceptable-use` references and hrefs to `registry.paratext.org` (the body says only the scheme changed — and the body's "known residual" that `paratext.org/acceptable-use` 404s is now moot; all three `registry.paratext.org` paths return 200),
- alphabetized §2 and added a "Backup Agent" definition,
- restated §§17–18 and rewrote 18.6 with 31-/60-day windows,
- changed 3.B.1 and simplified 5.5.

Squash-merge will destroy the commit message that discloses this.

**Suggested fix:** update the PR body and the ADR's `**amended**` notes (dated 2026-09-10, describing the older text) so a reviewer verifies the right revision. No code change. This is the one part of the PR that can't be reviewed from the diff.

---

### 3. Two of the three UBS permission paragraphs are covered by no test — `render.ts:694-718`

`lexicalDatabasePermission` spells one UBS legal paragraph three times — Paratext product, non-Paratext product, no product. Only the third is byte-compared by the golden fixture. So a lawyer-requested wording change gets applied to the arm that shows up in the test, and the two product arms — the ones that actually ship in a downstream product's document — keep the old text with every test green.

**Suggested fix:** build the paragraph from one shared string with the differing clause substituted, or add golden coverage for the two product arms (that falls out of item 7 if you take it).

---

### 4. A downstream document points readers at core's `LICENSING.md` for the product's own license — `render.ts:633`

`For the license covering ${subject} itself, see [LICENSING.md](./LICENSING.md).` renders in a Paratext 10 Studio document as "For the license covering Paratext 10 Studio itself, see LICENSING.md" — a file describing *paranext-core's* AGPL/MIT boundary, when per this same PR that product's license is the Terms of Service. Same class: the heading `## Extensions packed from outside this repository` in a document generated from a different repository.

**Suggested fix:** make the licensing pointer part of the product block, or omit the sentence when a product is declared. Worth settling before the downstream PR merges, since it happens every time that document generates.

---

### 5. Adding an extension and cutting a release ships it unitemized, with every gate green — `main.ts:872`

`assertExternalExtensionsRecorded` is called only from `buildReport`. `npm run package` runs `--verify-document`, which is a hash compare of two committed files. `publish.yml:177` and `package-main.yml:135` run `--verify-shipping-set`, which explicitly excludes this check (`main.ts:1016`). Only `test.yml`'s Linux `--verify` leg runs it.

So the sequence is: a downstream drops a new extension zip into its mapped folder, doesn't regenerate notices (nothing about adding an extension looks like a notices change), and cuts a release. Document unchanged, lock matches, every release gate green, installer ships an extension whose bundled dependencies are itemized nowhere and whose omission the document doesn't record — which is the state `external-extensions.ts:9-21` says the module exists to prevent.

**Suggested fix:** run the check on the packaging job. The module comment's own reasoning — that it can only answer where the installer's extensions actually are — is an argument for running it exactly there.

---

### 6. `<TODO: ask legal where the source lives>` now passes the placeholder gate — `policy.ts:498`

The regex went from `/^<.*>$/` to `/^<(?![A-Za-z][A-Za-z0-9+.-]*:)[\s\S]*>$/`. The lookahead excludes any bare word followed by a colon, not a URI scheme, so `<TODO: ask legal where the source lives>`, `<NOTE: fill me in>` and `<FIXME:>` were all caught before this PR and pass now. It guards `SeparateProgram.sourceAvailability` and `ExternalExtension.reason`, both hand-written reviewer paragraphs, and is also read at `policy.ts:608` (exception `reason`/`reviewer`) and `~:1255` (override `license`/`note`), neither of which has anything to do with autolinks.

The new `report.test.ts` sweep is green only because the nine templates `report.ts` prints happen to contain no colon — and that sweep scans `report.ts` only, while this PR adds a second placeholder-emitting template at `external-extensions.ts:151`.

**Suggested fix:** anchor on a real scheme, `(?!(?:https?|mailto|ftp):)`, plus a test case; and widen the template sweep to both sources.

---

### 7. Nothing in core exercises the overlay path end to end — `__fixtures__/golden/expected.md`

~1,800 non-test lines ship here whose only full-output check is a golden that gained one paragraph and renders none of the three new sections, and no workflow sets `NOTICES_POLICY_OVERLAY`. A refactor of `renderArtifactPair`, `reachableIds` or the report shape can leave every unit test green and still produce a downstream document missing a section, exiting 0 — surfacing in a different repo, at package time, on a release cut.

Related and cheap: `it('accepts the shipped policy against the shipped packaging config')` (`derived-invariants.test.ts:403`) hardcodes `[]` rather than `externalExtensionNames(REPO, config)`, so you can delete the entire bodies of `assertSeparateProgramsRecorded` and `assertExternalExtensionsRecorded` and it still passes — `Object.entries({}).forEach` and `[].filter` just run zero times.

**Suggested fix:** commit a fixture overlay with its evidence files plus a second golden rendered through `loadPolicy(POLICY, FIXTURE_OVERLAY)`, and derive that test's names from the real config. This is the largest ask on the list and a fair candidate for a follow-up ticket — but it's also the only thing that would catch items 3, 4 and 5 recurring.

---

### 8. The `separateProgram` route is reachable only while Mercurial's nuspec declares nothing parseable — `policy.ts:1608`, `:1749`

```ts
const overridable = !declared.ok && !identifiedText ? override : undefined;
...
if (afterException.verdict !== 'blocked' || exception || !overridable) return afterException;
```

So the new instrument is consulted only for a package that declares no parseable license *and* yields no identified license text. If `hgWindows` ever ships `<license type="expression">GPL-2.0-or-later</license>` — the direction NuGet has been pushing for years — `declared.ok` becomes true, `overridable` is `undefined`, and the package hard-blocks with no supported route back. `policyRemedy`'s copyleft branch (`report.ts:239-247`) compounds it: it offers an `elections` entry or "otherwise the dependency itself has to change", and never mentions `separatePrograms`, so whoever hits this gets told to drop Mercurial.

I'm raising this as a design question rather than a defect — the comment above that line deliberately makes an override weaker than an exception, and I can see why. But as written, the feature's live case rests on a third party's packaging metadata not changing, and the tests don't show it: the `overridden` fixtures in `policy.test.ts` all use `declaredField: undefined` with an empty detection, which is the one shape where the route is reachable at all.

**Options:** let a `separateProgram`-linked override through regardless of `declared.ok` (the link already carries its own stricter checks — the program must exist, the license must be SPDX, every id must be one the reviewed entry names), and add the `separatePrograms` route to `policyRemedy`. Or decide the current bound is correct and say so in the comment, so the next person doesn't read the Mercurial case as the supported path.

---

### 9. Two cheap ones

- **The four committed-policy gates are spelled twice, character for character** — `main.ts:865-871` vs `:1026-1037`, same calls, same derived arguments. A fifth gate added in `buildReport` — the natural place, and the one a developer runs locally — silently would not run on the release path, which is the same mechanism that produced item 5. One `assertCommittedPolicyGates(policy, packagingConfig)` removes ~12 lines and makes the drift unspellable.
- **The deferred `externalExtensions` work names no ticket** — `external-extensions.ts:18`, `types.ts:306`, and the ADR around line 1890 all describe work that's deliberately not done here ("Revisit when the extension template emits module manifests"), while the sibling gap in `separate-programs.ts:24` cites PT-4560. Per `.claude/rules/code-quality/forward-facing-comments.md` this wants an open ticket, otherwise nothing surfaces it again.

---

### Considered and not raised

So you don't have to wonder whether these were missed:

- **`mergeTable` does no shape check on the eleven table fields.** Genuinely a gap in a guard, but `assertSeparateProgramLinksRecorded` (`separate-programs.ts:238`) throws when an override's link names a program the table doesn't hold, and `assertExternalExtensionsRecorded` (`external-extensions.ts:140`) throws on every packed extension the table doesn't record — so a table mangled to `{}` or to key `"0"` fails loudly in both cases that actually arise. Worth a `requireTable` sibling of `requireList` someday; not worth blocking on.
- **Unreachable without a contrived config:** an empty/whitespace `separateProgram` link; the unparented-window path blocking `app.quit()`; `path.join` vs `path.resolve` on an absolute evidence path; `will-frame-navigate` (the CSP keeps frames out); the `base.product` and `assertListsDisjoint` checks not running on the no-overlay path.
- **Error-message quality rather than behavior:** the bare `ENOTDIR` on a file-shaped `from`; `alsoContains` missing its `Array.isArray` sibling; `MERGE_KINDS` omitting `electionsNote`/`overridesNote`; `readPackagingConfig` bypassing `readJsonFile`.
- **Style and doc accuracy:** the `admissible` docstring saying "the committed policy" where the merged one is passed; the `addSeparateProgramTexts` comment contradicting the code 20 lines below it; the ADR citing `corpus-texts.ts` for `corpus-texts.test.ts`; Markdown escaping of the filename-derived heading key; the ambient `NOTICES_POLICY_OVERLAY` vs an argv flag; the one-directional `isParatext` gate; the five identifiers added to `allowed`.

Nice work on the comments throughout `policy.ts` and `separate-programs.ts` — items 3, 6 and 8 were all findable only because the code explains what it's trying to guarantee.

---

https://claude.ai/code/session_01FbY2DkuLLFQteW1RwHoBJC

lyonsil:
@tjcouch-sil Thanks for the sweep - responses to all nine below, pushed as `927162cf3d5`.

Verification for the whole round: full suite 1434/1434, `typecheck:erb` clean, `npm run lint` clean, and `verify:third-party-notices:document` confirms core's committed document and lock are byte-unchanged.

---

### 1. The Terms of Service window — the copy claim doesn't hold; the rest is fixed

**Not taken (the menu).** The bare window is deliberate: the goal was something very basic that avoids complications with the rest of the Platform UI, and the document carries a link to the live page at `registry.paratext.org/terms` for anyone who wants a browser's affordances.

But the stated consequence doesn't occur, and I checked rather than argued it. Electron 39.8.8 from this repo, a window with `removeMenu()` **and** `Menu.setApplicationMenu(null)`, keys injected into the renderer:

```
{ "mode": "removeMenu", "menuOnWindow": false,
  "selectionAfterCtrlA": "MAGIC-COPY-PAYLOAD-42",
  "clipboardAfterCtrlC": "MAGIC-COPY-PAYLOAD-42" }
```

Blink maps Ctrl+C/X/V/A to its editing commands on Windows and Linux regardless of any Electron menu - the well-known "no menu, no copy" breakage is macOS-only, and there this app sets a process-global application menu carrying `role: 'editMenu'` (`platform-macos-menubar.data.ts`), which stays active while this window is focused. So copy and select-all work on all three platforms. What is genuinely absent is zoom, F12, Escape and find, which is the trade-off.

**Both of your other two are taken.** The window is now created with `show: false` and `backgroundColor: '#f5f5f3'` (the document's own body colour) and shown after the load resolves - no `ready-to-show` listener, since the function already awaits the load, so there's one fewer way for it to end up never shown. `autoHideMenuBar` dropped as inert. The failure path logs the path it tried as well as throwing; three tests cover the shown/not-shown/logged cases.

### 2. The PR body — updated

Body now says **14 August → 11 September**, split into the two invisible steps with the second enumerated: the host move (17 references, 8 of them hrefs), §2 alphabetized plus "Backup Agent", §§17-18 restated, 18.6's 31/60-day windows, 3.B.1's general AGPL sentence dropped, 5.5 simplified, §4's word space. The Critical finding carries a **Superseded** note, Suggested Review Focus points at the 11 September source, and the 404 item is struck - all three `registry.paratext.org` paths return 200, which I confirmed.

Two body claims were also wrong independent of the revision and are fixed: it said the link policy "admits only https" and that there were "eight `http://` links". It's `https:`/`mailto:`/the app scheme, and twelve links (11 https, 1 mailto).

**One ADR note was stale beyond what you flagged.** `Architecture-Decisions.md` read "*it has always been in 3.B.2*". In the 11 September text the phrase is in **both**: 3.B.1 gained "and are identified as such in the license notices accompanying the Paratext 10 application". Amended, and `LICENSING.md:239` now cites both sections.

### 3. The UBS paragraphs — golden coverage, not the shared-string refactor

Took your second option. The refactor is the wrong trade here: the three arms are hand-wrapped line arrays, there's no wrap helper in `render.ts`, and the product name's length decides where the break lands - so factoring the shared sentences out means either introducing a wrapper (which changes the committed document's wrapping) or gaining nothing.

All three arms are now pinned **word for word** by `ubsParagraph()` extracting the paragraph and comparing whole strings. That's stronger than a third golden would have been and costs ~40 lines instead of ~350.

### 4. The `LICENSING.md` pointer — the product names its own license document

Half-right, and the half that's right matters more than it looks. Paratext 10 **does** have its own `LICENSING.md` (#190 adds 69 lines), and the notices document is generated into that repository beside it - so in the repo the relative link resolves correctly. It's wrong in the **installer**: `copyNoticesPairIntoCore` copies only the notices pair, so `resources/LICENSING.md` is core's, deliberately. And the product's terms are the Terms of Service, not LICENSING.md, either way.

No link can be right in both places, so the file is now **named rather than linked**:

```jsonc
"licenseDocument": {
  "label": "the Paratext Terms of Service",
  "file": "TERMS-OF-SERVICE.html",
  "href": "https://registry.paratext.org/terms"
}
```

rendering *"For the license covering Paratext 10 itself, see the Paratext Terms of Service, which ships beside this document as `TERMS-OF-SERVICE.html`. The current published copy is at &lt;…&gt;; the file shipped with this build is the one that applies to it."* The published URL is second on purpose - it can move, and the installed copy is what licenses the build in hand, offline. Required when a product is declared, and `product.ts` refuses a name no `extraResources` entry copies into `resources/`.

Heading taken too, and pushed further: `## Extensions packed from other repositories`, unconditionally. Those zips are outside core *and* outside the product repo, so the new wording is true from every vantage point and needs no product arm.

### 5. The unitemized extension — right gap, but `--verify-document` wouldn't have closed it

"Run the check on the packaging job" taken literally would have missed the case entirely: Paratext 10's `package-core` runs **`--verify-shipping-set`** (`lib/build.ts:466`), not the document check. And `--verify-document` is documented as reading no build state, which is what lets it run before a build.

So the check went onto `--verify-shipping-set`, which covers core's `publish.yml`/`package-main.yml` **and** the downstream's packaging and release legs in one site. Scoped to `process.platform`: `externalExtensionFolders` unions all four `extraResources` blocks, which is right when writing a document that covers every platform, but on a build leg it would demand folders another platform maps and blame a copy step that was never meant to run there.

### 6. The placeholder regex — narrowed, and nothing breaks

Checked against every template in the tree and both live policies before changing it:

| | current | `(?!(?:https?\|mailto):)` |
|---|---|---|
| 12 templates across 4 modules | caught | caught |
| `<TODO: …>`, `<NOTE: …>`, `<FIXME:>` | **missed** | caught |
| `<https://…>`, `<mailto:…>`, `<http://…>` | admitted | admitted |

Neither `notices-policy.json` nor the downstream overlay holds a single value that is *entirely* `<…>`, so the narrowing costs nothing today. Kept `http` beside `https` deliberately - there's no security argument here (this is Markdown read on GitHub, not the ToS window's link policy), and refusing an `http://` autolink as "a placeholder from the template" is a message that describes the wrong problem.

Sweep widened as you asked, and further: it reads **every** non-test module in the directory rather than a list, so a fifth emitter is covered without anyone remembering. There are four today, not two - `main.ts:635` and `static-assets.ts:277` were already outside it.

### 7. End-to-end overlay coverage — tiers 1 and 2 here, tier 3 on a ticket

- **Second golden**: `expected-product.md` renders the full product arm - preamble, UBS permission, license pointer, and all three new sections. Built on short-canonical-text identifiers so it doesn't carry 340 lines of GPL text for no extra coverage.
- **`overlay.test.ts`**: a committed fixture overlay driven through the real `loadPolicy(POLICY, OVERLAY)` - the only place core reads an overlay off disk. Asserts the merge (union not replace), the product block, every committed-policy gate against the merged result, and a render carrying what only the overlay could supply. Mutation-checked: making `mergeTable` replace instead of merge fails it.
- **`derived-invariants.test.ts:403`**: arguments derived from the real policy and packaging config. Worth being straight about the limit - core maps no external folders, so `externalExtensionNames(REPO, config)` still returns `[]` and this still cannot catch a gutted body. It now exercises the real pair rather than a literal; falsifiability lives in the unit tests, and the comment says so.
- **Tier 3** - a real `buildReport` → `renderArtifactPair` against a fixture repo - needs npm and NuGet graphs and is genuinely large. Not in this PR.

One thing worth flagging: none of this would have caught item 5 recurring. That's workflow wiring, not rendering, so please don't read item 7 as covering it.

### 8. The `separateProgram` route — taken, with one consequence recorded

A linked override is now consulted whatever the package declares. The argument that settled it: an unlinked override is keyed by name and pinned to nothing, which is what the bound exists to hold back, but a linked one may only name a program `separatePrograms` records and may only carry an identifier that reviewed entry itself names - stricter than the gate it bypasses. And one detail sharpens your point: `hgWindows-6.3.1` is a repackaging of someone else's binaries, so its metadata is not something to build a guarantee on in either direction.

`policyRemedy`'s copyleft branch now names the third route - record a `separatePrograms` entry and link the override to it - instead of stopping at "the dependency has to change".

**The consequence, stated rather than buried:** where a package's declaration and the reviewed entry disagree, the entry now wins and the document records the entry's identifier. That's the right authority - a person read the program's terms - but it is no longer checked. Recorded in the ADR; a disagreement guard is a small change if we decide we want one.

### 9. Both taken

- `assertCommittedPolicyGates(policy, packagingConfig)` extracted, called from both paths. Worth noting the drift you predicted has already happened: `assertStaticAssetNoticesRecorded` and the copied-platform-library pair run in `buildReport` and not on the release path. They read build output, so that may be correct - but three gates having diverged already is the argument for one named function that makes the split explicit. The generator README now says where a new gate belongs.
- Filed **PT-4604** for the external-extension itemization, cited at `external-extensions.ts`, `types.ts`, the ADR, and #190's follow-ups. Deliberately not PT-4560: that one is the static-asset trees being inventoried by notice-shaped *filename* rather than by content - same shape of hole, different mechanism and fix. The two were conflated in two places.

---

### Also in this push

Product naming moves to **"Paratext 10"** (dropping "Studio") in the new examples, fixtures and docs. paranext/paratext-10-studio#190 carries the matching `productName` change, plus a `licenseDocument` block and a regenerated notices pair, before it comes off draft.

.erb/scripts/third-party-notices/main.ts
```diff
@@ -774,6 +816,35 @@ export function assertCopiedPlatformLibraryIdsAllowed(policy: Policy): void {
...
+ *
+ * The packaging config, the policy tables, and the evidence paths those tables name - no module
+ * manifest, no build output, no directory whose contents differ per platform. So every path that
+ * reads the policy at all can afford to run them, and they are spelled ONCE so that the two that do
+ * cannot diverge. `buildReport` is the natural place to add a gate and the one a developer runs
+ * locally; `verifyNpmShippingSet` is the only notices check the release workflows run. A gate that
+ * reaches the first and not the second does not run on the path that cuts a release.
+ *
+ * Deliberately NOT here: `assertExternalExtensionsRecorded`, whose input is a directory listing;
+ * `assertStaticAssetNoticesRecorded` and the copied-platform-library pair, which read build output.
```
tjcouch-sil, line 830: **A. The sibling gate is still off the release path, and the reason stated here does not hold for it.**

You raised this yourself in item 9 — "They read build output, so that may be correct" — so treat this as an answer to that rather than a disagreement. For `assertStaticAssetNoticesRecorded` it is not correct.

`findStaticAssetNotices` -> `packedStaticTrees` walks `extensions/src/*/{assets,public,contributions}`, the root `assets/`, and the two `DOTNET_STATIC_TREES`. All repository source, no build output. `static-assets.ts:146` says so deliberately: *"Read from `extensions/src` and deliberately NOT from `extensions/dist` ... a `dist` path would pin build output that no reviewer reads and no clone holds."*

So it takes `(REPO, policy)` and could sit beside `assertExternalExtensionsRecorded` on the `--verify-shipping-set` path. As written, the hole item 5 closed is still open through a more routine trigger: drop a file matching `^(attribution|licen[cs]e|notice|copying|credits?|source)\b` into any extension's `assets/`, don't regenerate notices, cut a release. `--verify-shipping-set` never looks, `--verify-document` is a hash compare, and the document ships without it.

Either move it onto the shipping-set path, or narrow this comment to the copied-platform-library pair — which I did not check and which may genuinely need build state.

src/main/terms-of-service-window.ts
```diff
@@ -0,0 +1,114 @@
...
+    backgroundColor: '#f5f5f3',
+    icon: path.join(globalThis.resourcesPath, 'assets', 'icon.png'),
+    // The document neither needs nor gets the application's preload and Node access
+    webPreferences: { sandbox: true, contextIsolation: true, nodeIntegration: false },
+  });
+  termsOfServiceWindow = newWindow;
+  newWindow.on('closed', () => {
+    if (termsOfServiceWindow === newWindow) termsOfServiceWindow = undefined;
+  });
+  newWindow.removeMenu();
```
tjcouch-sil, line 79: **B. The keyboard-shortcuts catalog was not updated, and your own testing is what makes this a certain miss.**

Not re-opening the menu decision: I accept the refutation, you tested it and I didn't. But the conclusion you reached — "what is genuinely absent is zoom, F12, Escape and find" — is exactly what `.claude/rules/keyboard-shortcuts-catalog.md` asks to be recorded in the same change that changes a handler's reach.

`src/stories/keyboard-shortcuts.data.ts` still carries Zoom in/out/reset, F12, and the `dismiss-overlays` Escape as `context: 'Main process (global)'` with `locations: ['src/main/main.ts', ...]`, and no Terms-window qualification anywhere. That file is not in this PR's diff, which is why I'm hanging this on `removeMenu()` instead.

There is a direct precedent a few entries up in that same file: the connection-lost Escape is catalogued with *"Catalogued because it is the one dialog in the app where Escape does NOT close anything."* Recording where a global shortcut does not reach is already the convention there.

.erb/scripts/third-party-notices/policy.ts
```diff
@@ -1211,8 +1612,27 @@ function readInstruments(
...
+  // staying license-silent: the NuGet delivery of a separate program is somebody's repackaging of
+  // an upstream release, and the day it declares its licence - the direction NuGet has pushed for
+  // years - the package would hard-block with no policy entry able to clear it, on a program a
+  // human had already reviewed. That is the reverse of what the reviewed entry is for.
+  //
+  // What this does NOT do is reconcile a disagreement: where the package declares one licence and
+  // the reviewed entry names another, the entry wins and the document records the entry's. That is
+  // the right authority - a person read the program's terms - but it is not a check, so a
+  // declaration that contradicts a reviewed program is worth noticing by hand.
+  const linkedToSeparateProgram = override?.separateProgram !== undefined;
```
tjcouch-sil, line 1632: **C. Worth taking the disagreement guard you offered — the gap is not symmetric.**

Answering "if we decide we want one": I'd say yes, and narrower than the general reconciliation problem.

The comment just above frames the unchecked case evenly, but only one direction costs anything. If the reviewed entry names a **less** restrictive identifier than the package declares — entry says MIT, a future nuspec says GPL-2.0-or-later — the entry silently wins and the document publishes the weaker claim. The reverse, entry more restrictive than the declaration, is harmless: we over-disclose.

So the guard only has to cover one side: when `declared.ok` and the declared identifier is not among the linked program's own `spdx` ids, throw and name both. That asks the pipeline to reconcile nothing, and it keeps authority where you put it — a person read the program's terms — while refusing the single case where deferring to them under-reports.

.erb/scripts/third-party-notices/external-extensions.ts
```diff
@@ -0,0 +1,213 @@
...
+ *
+ * `platform` narrows to what one platform's build packages - see `extraResourceEntries`.
+ */
+export function externalExtensionFolders(
+  config: PackagingConfig,
+  platform?: typeof process.platform,
+): string[] {
+  return extraResourceEntries(config, platform)
+    .filter(isFileSetEntry)
+    .filter((entry) => normalizeFolder(entry.to) === 'extensions')
```
tjcouch-sil, line 101: **D. Minor and new: `normalizeFolder` collapses separators but not case.**

Raising it only because you reworked this function in this round to handle backslashes, repeated `./` and trailing slashes for "a config edited on Windows" — a config edited on Windows can also produce `./Extensions`. That maps to a real folder so electron-builder packs it, but `=== 'extensions'` misses it and `externalExtensionFolders` answers empty, so the gate now passes silently on the release path rather than catching the omission. That is the silent drop the module docstring says it exists to prevent.

One `.toLowerCase()` in `normalizeFolder`, or on this comparison. Lowest priority of the four — drop it if you think `./Extensions` isn't a realistic spelling.

tjcouch-sil (commented):
Four follow-ups on the round. Thanks for the depth of the answers — and for testing the copy claim rather than arguing it; I should have done that before asserting it.

Two of these (A, C) are answers to questions you left open rather than pushback, and two (B, D) are new. Nothing here re-opens the menu decision.

Posting inline this time so each is a resolvable thread — last round produced none, which is why nothing could be marked resolved.

Verified and happy with the rest: the `licenseDocument` block and the named-not-linked pointer (better than what I asked for), the `--verify-shipping-set` placement with platform scoping, the sweep widened to every non-test module, `ubsParagraph` pinning all three arms word for word, `overlay.test.ts` plus the second golden, and `assertCommittedPolicyGates`. I also checked the limit you were straight about in item 7 — `external-extensions.test.ts` does carry the `toThrow` coverage you're leaning on, so that's accurate rather than hopeful.

Two coordination notes, not findings: `licenseDocument` is now required whenever a product block is declared, so any existing downstream overlay without it hard-fails — #190 has to land in lockstep. And the Studio → "Paratext 10" rename arrived in this same push.

---

https://claude.ai/code/session_01FbY2DkuLLFQteW1RwHoBJC

tjcouch-sil (approved):

PR #2801, opened 2026-09-10
PT-4549: Migrate the titlebar picker onto ProjectSelector

jolierabideau:
Migrates the Platform.Bible titlebar's bespoke `<Select>` project picker onto the shared
`ProjectSelector` (PT-4549). The flat either/or list becomes a sectioned one — Recent, then "Your
projects on this computer" — with a keyboard-reachable "More projects…" footer. Two live defects go
with it: the escape hatch no longer vanishes when there are zero local projects, and read-only
projects are listed and marked rather than filtered out. A new `usePendingProject` hook names a
just-picked project in the trigger before the editor catches up.

## Round 2

The round-2 review's five decisions are settled and its five warnings fixed; see the reply comment
for the item-by-item disposition. Notable: `usePendingProject` and the `vitest.config.ts` alias both
ship here, the cold-start fan-out stays with a measurement ticket filed against it, and the
write-up the ADR cites is now committed at
`.context/designs/PT-4549-followup-projectselector-accessible-name.md`.

## No longer stacked

This was stacked on #2790 (PT-4548), which has since landed, so the base is `main` and the diff is
this work alone. #2790 squash-merged while this branch had merged its commits individually, so the
catch-up merge showed conflicts across every file the two touched; resolution is recorded in the
merge commit, including the ADR log's union-by-slug check.

## API changes

- `platform-bible-react/experimental`: adds `type ProjectSelectorFooterAction`
  (`{ label, onSelect }`) and exports `PROJECT_SELECTOR_DEFAULT_STRINGS`.
- `ProjectSelectorProps`: adds optional `footerAction` (all modes) and optional
  `renderTriggerLabel` (`mode: 'project'` only). Both additive; no existing consumer is affected.
- New core strings, in `en.json` and `es.json`: `%projectPicker_readOnly_label%`,
  `%projectPicker_section_projects_localOnly%`, `%projectPicker_toolbar_trigger_label%`,
  `%projectPicker_toolbar_trigger_label_error%`, `%projectPicker_toolbar_trigger_label_empty%`,
  `%projectPicker_toolbar_label_nameAndShortName%`, `%projectPicker_toolbar_label_shortNameOnly%`.
- `lib/platform-bible-react/dist/` rebuilt. `lib/papi-dts/papi.d.ts` correctly unchanged —
  `platform-bible-react` types are not inlined there.

`ProjectSelector` is exported from `experimental.ts`, not the stable barrel
(`adr-project-selector-stays-experimental`), so this surface can still evolve.

## Decisions worth a reviewer's opinion

- **`vitest.config.ts` now aliases `platform-bible-react/experimental` to source.** The package's
  `exports` map sent it to the committed `dist/`, so a regression in a component's source passed
  green until someone rebuilt. Verified both ways: deleting `forceMount` now fails the toolbar
  integration test, which it previously survived, and no other root test changes result. This
  affects all 25 root files importing that entry point, so it is worth an explicit yes.
- **The popover's local-projects bucket sorts by full name**, not the selector's canonical short-name
  order, so it agrees with the "More projects…" dialog. That makes this one consumer order its list
  differently from every other `ProjectSelector`.
- **A project reachable only through the dialog is opened without being named** — the trigger keeps
  naming what is open until the editor reports the change, rather than standing the raw id in for a
  name. `TODO(PT-4552)` marks the site.
- **`PENDING_PROJECT_TIMEOUT_MS = 15_000` is an unmeasured round number**, labelled as such. If
  anyone has project-open latency data, this is the moment to replace it.
- **PT-4133 is "Won't Do"** yet is cited as the background for pending-project display. Its second
  expectation — a loading state tied to the opening project — is not what ships; `isLoading` is
  driven by list loading, not by "this project is opening". Confirm that is acceptable.

## Deferred, deliberately

- **Server-reachable projects are out of scope** — that is [PT-4552](https://paratextstudio.atlassian.net/browse/PT-4552).
  This PR delivers the DoD's alternative: the section heading states that the list is what is on this
  computer.
- **Two `ProjectSelector` sharp edges stay on `experimental`**, per the #2801 review's findings 18
  and 19 and with the reviewer's agreement: `ariaLabel` replaces the trigger's visible label in the
  accessible name (fixed here at the call site, not in the library, so six other consumers are still
  affected), and read-only is consumer-derived, which leaves the row tooltip unable to explain the
  padlock to sighted pointer users. Written up in
  `.context/designs/PT-4549-followup-projectselector-accessible-name.md`, and recorded as
  consequences on `adr-project-selector-stays-experimental`.
- **Read-only projects absorb [PT-4511](https://paratextstudio.atlassian.net/browse/PT-4511)**: the
  `isEditable` filter is gone and those projects are marked with a lock icon rather than hidden, in
  the titlebar *and* in the dialog.

## Manual verification still outstanding

Not run in the authoring environment because starting the app binds machine-wide ports shared with
another checkout. Three items cannot be closed any other way:

1. Small-width layout across shrink steps — no text overlap. Design-reviewed, not test-backed;
   jsdom has no layout.
2. Activating "More projects…" **by keyboard** and confirming focus lands inside the modal — a
   focus-handoff race between Radix Popover and Radix Dialog.
3. Selecting a read-only project and confirming its content loads — PT-4511's third acceptance
   criterion, which depends on #2425. **If it does not load, PT-4511 stays open for that criterion.**

Also worth eyes: the footer is now `sticky` inside `CommandList`, which jsdom cannot verify; and a
row that is both current and read-only renders a lock and a check together.

## Verification

`npm run typecheck`, `npm run lint` and `npm test` are all clean — 5518 passed, 0 failed. The two
`rc-dock-tab-cache-patch.test.ts` failures noted in earlier rounds were fixed on `main` and came in
with the latest base merge.

## AI Involvement

AI-assisted throughout: design, implementation and tests were produced with Claude Code under my
direction, across a spec, a written implementation plan, and per-task reviews. I reviewed the result
before opening this PR.

Note for the merge commit: the `Co-authored-by` trailer is not on every individual commit — add it
to the squash message at merge time.

## Risk

**Medium** — the surface is small and every change is covered by tests, but it replaces a control
every Simple-mode user touches, and the manual verification items above are still open. The two new
library props are additive and optional.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2801)

[PT-4552]: https://paratextstudio.atlassian.net/browse/PT-4552?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


katherinejensen00 (changes requested):
# Review of PR #2801 — PT-4549: Migrate the titlebar picker onto ProjectSelector

**For:** @jolierabideau · **Head:** `2479249ab25` · **Base:** `origin/pt-4548-promote-project-selector` (stacked on #2790)
**Passes:** soundness/quality · architecture · documentation · antagonistic re-review of those three · `/code-review max` (10 finder angles)

Anything already present on the base branch was excluded. Each item gives a changed line to attach the comment to, a one-sentence summary, and the technical detail behind it.

---

## High

### 1. An isolated E2E test still looks for the Radix trigger this PR deleted, so it will fail the next time anyone runs that suite
**Attach:** `src/renderer/components/platform-bible-toolbar.tsx:366` · **Severity: High**

`e2e-tests/tests/isolated/bcv-open-resource-books-simple.spec.ts:212` asserts the open project's short name via `mainPage.locator('[data-slot="select-trigger"]').first()`. That attribute is emitted only by Radix `SelectTrigger`, which the base toolbar rendered (base `platform-bible-toolbar.tsx:191`) and this PR replaces with a `Button role="combobox"`. In Simple mode the toolbar now emits no `select-trigger` at all, so the locator either matches nothing or matches an unrelated select, and `toContainText(targetProject.shortName, { timeout: 60_000 })` burns a minute before failing. CI only runs `--project=smoke`, so this stays green until someone runs `test:e2e:isolated`. Update the spec to `getByRole('combobox')` in the same PR.

### 2. With zero projects, pressing Enter on "More projects…" does nothing — the exact case the footer exists for
**Attach:** `lib/platform-bible-react/src/components/advanced/project-selector/project-selector.component.tsx:1448` · **Severity: High**

Verified against `node_modules/cmdk/dist/index.mjs`: a `CommandItem` registers with `useLayoutEffect(() => { if (!forceMount) return store.item(id, groupId) })`, and cmdk's "select first item" only runs on item registration, item removal, or a search change. A `forceMount` item never registers, so with no project rows nothing ever sets `state.value`; every item's `aria-selected` stays false, `getSelectedItem()` (a `[cmdk-item=""][aria-selected="true"]` query) returns null, and the root's Enter branch is a no-op. ArrowDown works, because that path walks the DOM instead — which is why `component.test.tsx:777` passes: it presses ArrowDown first. `integration.test.tsx:326` only asserts the row renders. Suggested fix: seed the highlight when the filtered list is empty (or give the footer an explicit `value` and set it via cmdk's controlled `value` when `filtered.count === 0`), and add a test that opens with zero projects and presses Enter without arrowing.

### 3. Both new localization keys are missing from `es.json`, which the localization guide treats as blocking
**Attach:** `assets/localization/en.json:381` · **Severity: High**

`%projectPicker_readOnly_label%` and `%projectPicker_section_projects_localOnly%` exist only in `en.json`. `assets/localization/es.json` already carries all nine sibling `%projectPicker_*%` keys, so the new padlock label and the new section heading will render untranslated English next to translated neighbours. `.claude/rules/localization-locale-files.md` requires every new core string in both files with a real Spanish translation, and `Localization-Guide.md` calls a missing `es` key blocking.

### 4. Dropping the `isEditable` filter puts every installed resource in the recents pipeline, which can time out the Power→Simple switch
**Attach:** `src/renderer/hooks/use-project-picker-data.hook.ts:416` · **Severity: High**

`papi-shared-types.ts:494` states that `isPublished === true` implies `isEditable === false`, so the filter removed here was in practice the published-resource filter. Resources are now openable from the titlebar, but `cacheLastOpenedSimpleProject` (`web-view.service-shard.ts:1486-1495`) refuses to cache a published project, while `openProject` still calls `recordProjectOpened` — so a resource pick is never remembered yet permanently leads the Recent list. `getMostRecentUsableProjectId` (`:1929-1932`) then has to `getMetadataForProject` past each one inside `COLD_START_LOOKUP_TIMEOUT_MS = 3000`, so a few resources at the top of recents can exhaust that budget and leave Simple mode with no project. Worth confirming this interaction before merge, since PT-4511's absorption is what makes resources common in this list.

---

## Medium

### 5. The trigger greys out during every background refresh, and can drop keyboard focus right after a pick
**Attach:** `src/renderer/components/platform-bible-toolbar.tsx:389` · **Severity: Medium**

Passing `isLoading` through is new here; the base disabled on `!hasProjectPickerItems`. `usePromise` calls `setIsLoading(!!promiseFactoryCallback)` on every factory-identity change, and `useProjectPickerData` ORs four such flags whose factories depend on `activeEditorProjectId`, `safeRecentIds` and `getAllMetadata` — so `onDidChangeProjects`, a late PDP-factory registration, the retry timer and the recents update all disable the control (`ProjectSelector` disables on `isDisabled || isLoading`, component.tsx:1323). The sharpest case is keyboard: Enter on a row → Radix `onCloseAutoFocus` refocuses the trigger just as `recordProjectOpened` flips `isLoading`, and `.focus()` on a disabled button is a no-op, so the tab position falls to `<body>`. To be fair to the change, the empty-list case is *better* than the base (which disabled forever); this is about background reloads. Consider gating the disable on "no data yet" rather than "any load in flight", or splitting the spinner from the disable in the library.

### 6. The case-insensitive selection fix landed in the rows but not in the trigger, so the new prop's contract is false
**Attach:** `lib/.../project-selector.component.tsx:1221` (TSDoc at `:471`) · **Severity: Medium**

`computeRows` now matches selection with `normalizeProjectId` (`project-selector.rows.ts:273`), but `triggerContent` still does `props.projects.find((p) => p.id === props.selection.projectId)`. For the very case `rows.test.ts:59` documents — `projects: [{id:'ABC123'}]`, `selection: {projectId:'abc123'}` — the row shows a check mark while `renderTriggerLabel` receives `undefined` and the plain path shows the placeholder. The toolbar hides this behind `selected ?? displayedProject` (`:334`), so it is unreachable from your call site, but it is now part of a public prop's contract. `project-selector.rows.ts:257` (`pairIsSelected`) has the same raw comparison for `project-multi` and `projectScrollGroup`, and `find.component.tsx` renders both modes from one selected id. Also update the `renderTriggerLabel` TSDoc: `undefined` is passed not only "when nothing is selected" but whenever the selected id isn't found in `projects` — which is exactly the case your `:312` comment relies on.

### 7. `TruncatingTriggerText` is a hand-rolled tooltip that re-introduces two problems the repo already documented and fixed
**Attach:** `src/renderer/components/platform-bible-toolbar.tsx:133` (detail at `:138`) · **Severity: Medium**

The new component renders a fully controlled `<Tooltip open={open}>` with no `onOpenChange`, no pointer-down dismissal and no focus reveal. `sync-status-button.component.tsx:129-134,254-260` records that a controlled tooltip leaves Radix's own Escape handling inert (catalogued at `src/stories/keyboard-shortcuts.data.ts:554`), and `toolbar-compound-label.component.tsx:211-224` records that the missing `onOpenChange` lets the tooltip's dismissable layer `preventDefault()` the Escape meant for the popover beneath it. Repro: hover the clipped placeholder/error text, then click to open the picker — the tooltip stays painted and Escape closes neither. `ToolbarCompoundLabel` accepts `primary` alone and already carries all four affordances, and it is imported 50 lines below; `<ToolbarCompoundLabel primary={text} fullText={text} />` would let you delete this component.

### 8. `usePendingProject` can strand the wrong project name for the full 15 seconds when two picks resolve out of order
**Attach:** `src/renderer/hooks/use-pending-project.hook.ts:93` · **Severity: Medium**

The catch-up effect retires the bridge only when `currentProject` *matches* the pending pick. Pick A, then B before A resolves, then A lands: `currentProject` is A, `pendingProject` is B, nothing matches — the trigger names B while the editor shows A until the timeout, then flips. There is no exit on "current changed to something that isn't the pending pick", which would be the natural guard.

### 9. A failure in the recency bookkeeping retires a pick whose editor opened fine
**Attach:** `src/renderer/hooks/use-pending-project.hook.ts:77` · **Severity: Medium**

`openProject` (`platform-bible-toolbar.tsx:537-543`) awaits `openScriptureEditor` and *then* `dataProviders.get(...)` + `recordProjectOpened`. A rejection from either later call takes the hook's `.catch`, clearing the pending pick even though the editor opened — so the trigger snaps backwards to the old name. Separately, the latest-wins guard at `:81` compares project ids rather than attempt numbers, so a slow rejection for an earlier pick of A clears a newer pick of A (three picks inside one rejection latency; benign, but the same fix covers both: capture an attempt counter in a ref).

### 10. The localized strings can render as raw `%key%` text, because the raw-key seed is truthy
**Attach:** `src/renderer/components/platform-bible-toolbar.tsx:353` (also `:392`) · **Severity: Medium**

`useLocalizedStrings` seeds `defaultState[key] = key` and returns that seed before the provider responds, and permanently on a `PlatformError`. Because `'%projectPicker_no_results%'` is truthy, neither `resolveStrings`' `{...DEFAULT_STRINGS, ...partial}` merge nor `commandEmptyMessage ?? 'No projects found'` can fall back — the popover can show a literal `%projectPicker_search_placeholder%`, `%projectPicker_no_results%`, both section headings, and a padlock whose accessible name is `%projectPicker_readOnly_label%`. Filtering out values that still look like `%…%` before passing them down is the usual guard.

### 11. The trigger's accessible name is assembled by string concatenation, so translators can't reorder it
**Attach:** `src/renderer/components/platform-bible-toolbar.tsx:348` · **Severity: Medium**

`` `${selectProject}, ${currentProjectError}` `` and `` `${selectProject}, ${fullName} (${shortName})` `` hard-code English word order and punctuation in code. `Code-Style-Guide.md:228` asks for format strings precisely so the pieces can be rearranged. It also reuses `%projectPicker_toolbar_select_project%` (a placeholder) as the control's name prefix, so the two can never be worded differently. A dedicated key such as `"Select project, {projectName}"` filled with `formatReplacementString` fixes both.

### 12. The error and pending branches of the accessible name are untested, though the PR body says they mirror `renderTriggerLabel`
**Attach:** `src/renderer/components/platform-bible-toolbar.tsx:346` · **Severity: Medium**

The integration tests assert only the current-project name (`:238`) and the placeholder name (`:265`); `platform-bible-toolbar.test.tsx` never mentions `ariaLabel`, and its stub doesn't render it. Drop the `!pendingProject &&` guard at `:348` and a screen reader would announce a stale error while the visible label shows the new pick, with nothing failing. Two integration cases (error set; click a row and re-read the name) would pin it.

### 13. The popover and the "More projects…" dialog sort the same projects differently
**Attach:** `src/renderer/components/platform-bible-toolbar.tsx:288` · **Severity: Medium**

The `'yours'` section supplies no `compare`, so `partitionByCustomSections` falls back to `compareRows`, which sorts by **shortName** (`project-selector.rows.ts:457`). `useProjectPickerData` sorts `allProjects` by **fullName** and `ProjectPicker` doesn't re-sort, so opening the dialog from the popover reorders the same list under the user.

### 14. The dialog fallback names a project by its raw id and silently loses its read-only padlock
**Attach:** `src/renderer/components/platform-bible-toolbar.tsx:563` · **Severity: Medium**

`beginOpenProject(item ?? { id: projectId, shortName: projectId, fullName: projectId })` puts `id (id)` in the titlebar *and* in `triggerAriaLabel` for up to 15 seconds, and omits `isEditable` — which `readOnlyIds` (`:254`) tests as `=== false`. This is reachable today, not only after PT-4552: the dialog runs its own `useProjectPickerData` instance and can be a refresh generation ahead of the toolbar's. Falling back to the placeholder would degrade more gracefully than showing an id.

### 15. The footer row is announced as an ordinary listbox option, but it opens a modal dialog
**Attach:** `lib/.../project-selector.component.tsx:1451` · **Severity: Medium**

cmdk's `CommandItem` renders `role="option"`, so a screen-reader user hears "More projects…, option" and expects Enter to select a project; instead the popover closes and a dialog opens. The ADR commits to keeping the row inside `CommandList` for keyboard reachability, which is right — so the fix belongs on the item (`aria-haspopup="dialog"`, or a non-`option` role). Nothing in the diff addresses it.

### 16. Only the new footer separator got `alwaysRender`, so the section rule disappears as soon as the user types
**Attach:** `lib/.../project-selector.component.tsx:1427` · **Severity: Medium**

`{index < sections.length - 1 && <CommandSeparator />}` sits two lines above a separator that *does* pass `alwaysRender`, with a comment explaining that a plain `CommandSeparator` returns null once `state.search` is non-empty. So one keystroke makes the line between "Recent" and "Your projects on this computer" vanish while the footer's line stays. The inter-section separator is base code, but this PR is what documents the trap and fixes only its own instance.

### 17. The footer isn't actually pinned, despite its TSDoc
**Attach:** `lib/.../project-selector.component.tsx:1443` · **Severity: Medium**

`ProjectSelectorFooterAction`'s TSDoc says "pinned below the project list", but the row is the last child of `CommandList`, which is `tw:max-h-72 tw:overflow-y-auto` (`shadcn-ui/command.tsx:176`). With roughly nine or more projects — now including every installed resource, per finding 4 — the user has to scroll to reach the escape hatch. jsdom has no layout, so every `findByTestId` passes regardless. Keyboard reachability does require it inside `CommandList`, so the fix is a sticky row, not a move.

### 18. The read-only padlock has no explanation for sighted pointer users in the titlebar
**Attach:** `src/renderer/components/platform-bible-toolbar.tsx:302` · **Severity: Medium**

Dropping the native `title` (correctly, to avoid two stacked tooltips) leaves nothing to fill the gap: `ProjectRowView`'s tooltip is built from language, `typeName`, scroll-group label, `isBoundButClosed` and `disabledReason` (`component.tsx:600-607`) and never carries the indicator, and the toolbar's mapping doesn't pass `typeName` either. So "Read-only" reaches screen readers only, while the same glyph in the dialog (`project-picker.component.tsx:146`) does get a hover label. Adding a read-only line to the row tooltip is the consistent fix — and it strengthens the case for a first-class `isReadOnly` on `ProjectSelectorProject` rather than the consumer-derived indicator.

### 19. The accessible-name fix lives in the consumer, so every other `ProjectSelector` caller stays broken
**Attach:** `lib/.../project-selector.component.tsx:480` (call site at `platform-bible-toolbar.tsx:346`) · **Severity: Medium**

The PR body defers the library fix on the grounds that it would churn ~37 `getByRole` queries — but a fix scoped to "`renderTriggerLabel` was supplied" touches no existing consumer, since none pass it. Meanwhile `ariaLabel`'s TSDoc (`component.tsx:310-314`) and its `@example` at `:944` still tell callers to pass a control-only name, and six live sites follow it (`checklist.web-view.tsx:791`, `checks-side-panel.component.tsx:236`, `find.component.tsx:829`, `manage-books-dialog.component.tsx:2264` and `:2553`, `manage-books-sidebar.component.tsx:308`). The component is still experimental, so this is the cheap moment. Keeping it in the toolbar also leaves two parallel decision trees (`renderTriggerLabel` at `:320-338` uses `selected ?? displayedProject`; `triggerAriaLabel` at `:346-351` uses only `displayedProject`) that a future branch can silently desync. If you'd rather defer, please file the ticket and record it as a consequence in the ADR.

### 20. Three new tests can't fail
**Attach:** `src/renderer/components/platform-bible-toolbar.test.tsx:1494` · **Severity: Medium**

- `:1494` — `expect(projects.every((p) => sections.some((s) => s.match(p)))).toBe(true)` is vacuous: `customSections[1]` is the `'yours'` catch-all with `match: () => true`. It would stay green with the `recent` predicate inverted, with `recentIndex` keyed un-normalized, or with `projects` empty — none of which is what the comment claims it guards.
- `project-selector.component.test.tsx:705` — "renders no tooltip of its own" awaits `getByText('World English Bible')` rejecting, but the test passes no `triggerLabelFormat`, so the derived title would be `'WEB'`, never the full name. Delete the `title: ''` suppression at `component.tsx:1224` and it still passes.
- `platform-bible-toolbar.test.tsx:1528` — `'disables the trigger only while loading, never for an empty list'` never inspects disabled state; it only checks that the stub received `isLoading`.

### 21. The untranslated-defaults guard is a hardcoded copy whose comment promises the opposite
**Attach:** `src/renderer/components/platform-bible-toolbar.integration.test.tsx:310` · **Severity: Medium**

The comment says the values are "taken verbatim from the component's own `DEFAULT_STRINGS`, so a rename there surfaces here" — but the array is a literal, so a rename makes all seven assertions silently vacuous, which is exactly the failure it claims to prevent. Two of them are already vacuous: `queryByText` never matches attributes, and `'Search projects & resources'` is a `placeholder` while `'View options'` is an `aria-label`. It also covers 7 of the 24 defaults. `DEFAULT_STRINGS` is importable from a module this file already loads; looping over it with `queryByText` + `queryByPlaceholderText` + `queryByLabelText` would make it real.

### 22. The integration test runs against the built bundle, so it can't catch a component source regression
**Attach:** `src/renderer/components/platform-bible-toolbar.integration.test.tsx:16` · **Severity: Medium**

`vitest.config.ts` has only `tsconfigPaths()` and no `platform-bible-react` alias, and the package `exports` map sends `./experimental` to `dist/experimental.js`. The header documents this honestly rather than fixing it — but a `resolve.alias` to `src` in `vitest.config.ts` fixes it once for this file and every future one, and is worth doing here rather than deferring to a repo-wide discussion.

### 23. The read-only path is staged in the integration fixtures but never asserted
**Attach:** `src/renderer/components/platform-bible-toolbar.integration.test.tsx:154` · **Severity: Medium**

`PROJECTS[1]` is `isEditable: false` and `'%projectPicker_readOnly_label%': 'Test read-only'` is mocked, yet no test queries `getByLabelText('Test read-only')`. This is the only place `renderProjectIndicator` meets the real component — the unit tests only invoke it as a captured function against the stub — so the marking behavior PT-4511 is absorbed for has no end-to-end assertion.

---

## Low

### 24. Re-picking the already-open project now re-stamps recency and reorders the list under the cursor
**Attach:** `src/renderer/components/platform-bible-toolbar.tsx:366` · **Severity: Low**

Radix `Select` suppressed `onValueChange` for an unchanged value; cmdk fires `onChangeSelection` unconditionally. The editor itself is safe — `open()` takes the `focus-existing` path (`platform-scripture-editor/src/main.ts:304-310`), so no duplicate tab — but `recordProjectOpened` runs again and reorders the "Recent" section the toolbar renders. `usePendingProject` already detects this click (`isAlreadyCurrent`) but only to skip arming the bridge; skipping the open there too would close it.

### 25. The inline `selection` literal defeats the memoization two comments in the same file promise
**Attach:** `src/renderer/components/platform-bible-toolbar.tsx:380` · **Severity: Low**

`selection={{ projectId: displayedProject?.id }}` is a fresh object each render and `rows`' memo lists `props.selection` (`component.tsx:1016`), cascading into `filteredRows` and `sections` — so a full re-partition plus sort runs on every verse move with the popover closed. The cost is trivial at these list sizes, so this is mostly about the comments at `:85` and `:268` describing a rule this line breaks. `useMemo(() => ({ projectId: displayedProject?.id }), [displayedProject?.id])` settles it. Note the `:268` rationale is also inaccurate: the search query is internal state and `sections` depends on `filteredRows`, so it re-partitions per keystroke regardless of this array's identity.

### 26. The footer label is visually misaligned under the rows
**Attach:** `lib/.../project-selector.component.tsx:1448` · **Severity: Low**

The footer `CommandItem` carries no className, so its text starts at the item padding while every project row is pushed roughly 40px further in by the check slot, the reserved indicator slot (the toolbar always supplies `renderProjectIndicator`, so `reserveIndicatorSlot` is true) and two `gap-2`s — a ragged left edge in a 320px panel, which is the same misalignment `reserveIndicatorSlot` exists to prevent between rows.

### 27. The value-collision test aims at a mechanism that can't fail
**Attach:** `lib/.../project-selector.component.test.tsx:864` · **Severity: Low**

It clicks the footer node directly, and cmdk wires `onClick` on the element itself, so a duplicate value can't misroute a direct click — removing `value={FOOTER_ACTION_VALUE}` wouldn't fail it. (The dismissal in the PR body is right that a *project* can't collide: row values are composite, `component.tsx:675`.) The real exposure is `getSelectedItem()` and the repo's `spaceSelectsHighlightedItem`, both first-DOM-match — that's what a collision test should exercise.

### 28. `ReadOnlyIndicator`'s markup is re-forked two lines below its own import
**Attach:** `src/renderer/components/projects/project-picker.component.tsx:140` · **Severity: Low**

The new check-mark wrapper `<span role="img" aria-label title><CheckIcon aria-hidden/></span>` is that component's markup copied character-for-character, which forks the contract its TSDoc says "lives in one place" — parameterizing `ReadOnlyIndicator`'s icon (or extracting a shared `LabelledGlyph`) would keep one owner. Separately, `tw:shrink-0` stayed on the inner svg while the wrapper span is now the flex item, so the glyph can be squeezed; `read-only-indicator.component.tsx:25` has the same shape.

### 29. The hook returns and tests a value production code never reads
**Attach:** `src/renderer/hooks/use-pending-project.hook.ts:30` · **Severity: Low**

The toolbar destructures only `{ pendingProject, beginOpenProject }` (`:545`) and `ToolbarProjectSelector` re-derives the displayed project itself, while the hook tests assert on `displayedProject` 14 times versus `pendingProject` 6 — so reordering the toolbar's derivation to `currentProject ?? pendingProject` would leave every hook test green. Either delete it from `PendingProjectState` or consume it. The warning log at `:79` also says "Toolbar caught an error…" from inside a generic hook.

### 30. The 15-second wall clock stands in for a signal that already exists
**Attach:** `src/renderer/hooks/use-pending-project.hook.ts:20` · **Severity: Low**

`platformScriptureEditor.openScriptureEditor` returns the new web-view id (`platform-scripture-editor.d.ts:780`), but `openProject` is typed `Promise<void>` and only `.catch` is attached, so a deterministic completion signal is discarded in favour of a bound the TSDoc itself calls unmeasured. On a slow open the trigger reverts to the *previous* project's name while the new one is still loading — arguably worse than the lag it hides. Also note `.claude/rules/ux/apply-model.md` (scoped to `src/renderer/**`) names optimistic updates as needing UX approval, and the timeout path reverts with no user-visible signal.

### 31. ~130 lines of mock setup are duplicated between the two toolbar test files, and have already drifted
**Attach:** `src/renderer/components/platform-bible-toolbar.integration.test.tsx:24` · **Severity: Low**

Lines 24-167 copy `platform-bible-toolbar.test.tsx:30-188` (18 `vi.mock` calls, same order) plus a second `renderSimpleToolbarWith` with the same name and a different contract. They already disagree: `useScrollGroupScrRef` returns `{book:'GEN', chapterNum:1, verseNum:1}` here and `{book:1, chapter:1, verse:1}` in the sibling. A shared test-setup module would keep them honest. Relatedly, the file needs ~20 mocks because it renders the whole toolbar — extracting `ToolbarProjectSelector`, `TruncatingTriggerText` and `ProjectSelectorLabel` into their own file (`platform-bible-toolbar.tsx` is now 869 lines, up from 685) would let the test render just that unit, as `sync-status-button.component.tsx` already does.

### 32. The `.integration.test.tsx` suffix starts a second naming convention
**Attach:** `src/renderer/components/platform-bible-toolbar.integration.test.tsx:1` · **Severity: Low**

The PR body's "first in the repo" is literally true for the dotted form, but two hyphenated integration tests already exist: `src/renderer/services/workspace-updating-integration.test.ts` and `extensions/src/platform-scripture-editor/src/scripture-text-grid/scripture-text-grid.zoom-integration.test.tsx`. Either match them or write the new convention into `Testing-Guide.md` — worth deciding rather than settling by accident.

---

## Documentation and comments

The comments here are accurate and forward-facing on the whole, and several are genuinely load-bearing. Two problems: a handful state things the code doesn't do, and four rationales are each written out in full in four to eight places.

### 33. Two comments contradict the code they sit on
**Attach:** `src/renderer/components/platform-bible-toolbar.tsx:125` · **Severity: Medium**

- `:125-129` — the `TruncatingTriggerText` TSDoc says a native `title` "would never open". It would: nothing blocks it on a span inside this trigger (`shadcn-ui/button.tsx:22` only sets `[&_svg]:pointer-events-none` and `disabled:pointer-events-none`); the old block came from Radix `SelectValue`, which is gone. The real reason is tooltip consistency — say that.
- `use-project-picker-data.hook.test.ts:315` — "this recent row must keep marking it" is wrong; the hook marks nothing, it only passes `isEditable` through.

Also see findings 6 and 21 for two more comment/code mismatches, and `:268` in finding 25.

### 34. Four rationales are repeated across four to eight sites each
**Attach:** `lib/.../project-selector.component.tsx:1443` · **Severity: Low**

`forceMount` keeping the empty message alive (ADR, type TSDoc, `FOOTER_ACTION_VALUE`, the render site, two component tests, the integration test, the story, the PR body); "no native title, two tooltips would stack" (`ReadOnlyIndicator` TSDoc, `toolbar.tsx:127` and `:299`, `component.tsx:475` and `:1316`); "absent `isEditable` means editable" (`toolbar.tsx:256`, `project-picker.component.tsx:29-31`, two test comments); and the other-window reason for the timeout (hook TSDoc twice, `beginOpenProject`, two tests). Give each one home — the ADR or the TSDoc — and a one-line pointer elsewhere. That's roughly 40% of the added comment lines with nothing lost. Two smaller ones: the story comment at `project-selector.stories.tsx:209` is garbled ("the shape a titlebar compound label takes when it owns the trigger's whole rendering, `undefined` included"), and `use-pending-project.hook.ts:7-19` spends a defensive paragraph saying "not measured" three ways.

### 35. Backward-facing phrasing in two test comments
**Attach:** `src/renderer/components/platform-bible-toolbar.test.tsx:1519` · **Severity: Low**

"the picker this replaced tied `disabled` to the list being non-empty" (also `integration.test.tsx:334`) narrates the change rather than the invariant — `.claude/rules/code-quality/forward-facing-comments.md`. State the rule instead: "an empty list must not disable the trigger; 'More projects…' is the only way out." (The ADR's mention of the replaced control is fine — history belongs in an ADR's context.)

### 36. The PR description is ~23 KB, and most of it is a pasted review report
**PR-level comment** · **Severity: Low**

348 lines, of which roughly 80% is the `/review-paratext` output: a findings checklist where everything is already fixed or dismissed, "Positive Observations", "Interview Notes" (including that you had no memory of authoring a prior session's code), and "In-Review Quality Check" test counts that CI shows anyway. Three points are each made three times (the small-width manual check, the PT-4552 deferral, the PT-4511 read-only change). The genuinely useful parts — "Stacked on #2790", "Manual verification still outstanding", "Suggested Review Focus", "Deferred", risk — are buried, and a reviewer who skims will miss them. Roughly 60 lines: stacking note, 4-line summary, the API changes, the decisions you want reviewers to weigh in on, manual verification outstanding, deferred work, AI involvement, risk.

---

## Checked and found clean

ADR slug byte-order position (verified against all 75 slugs) and entry structure; `en.json` alphabetization; `…` used everywhere, no `...`; no ticket IDs or change narration in shipped source comments (the only `PT-4552` references are a forward-facing `TODO` and a prose pointer); `dist/` rebuilt and carrying the new exports; every added key declared and consumed; `%projectPicker_section_projects%` retained so the new `_localOnly` key is a correct new key rather than an edit, per the immutable-strings rule; TSDoc on all new exports; `papi.d.ts` correctly unchanged; the `ProjectItem` → `ProjectSelectorProject` language/languageCode inversion is mapped correctly; `partitionByCustomSections` uses `findIndex`, so a recent project can't appear twice; the footer's `onSelect` is wrapped so cmdk's value string never reaches `showDialog`; `cn`'s `extendTailwindMerge({prefix:'tw'})` resolves `tw:w-auto` over the library's `tw:w-[180px]`; fake-timer setup/teardown is symmetric; the footer can't be reordered by cmdk sorting (`shouldFilter={false}`); a project id can't collide with `FOOTER_ACTION_VALUE`; timer cleanup on unmount; no z-index number introduced anywhere in the diff.

> @katherinejensen00 partially reviewed 29 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2801#-:-P1m1qFv4GWirKgWFZhW:b-gi2m6)* status: all files reviewed, 1 unresolved discussion (waiting on jolierabideau).

___
*[`src/renderer/components/platform-bible-toolbar.tsx` line 366 at r1](https://reviewable.io/reviews/paranext/paranext-core/2801#-P1m2-v55DG--DmPJ1gI:-P1m2-v55DG--DmPJ1gJ:b17avq0) ([raw file](https://github.com/paranext/paranext-core/blob/2479249ab25dac31bd025a963cf80d5af357f85f/src/renderer/components/platform-bible-toolbar.tsx#L366)):*
> ```XML
>   );
> 
>   const handleChangeSelection = useCallback(
> ```

### 1. An isolated E2E test still looks for the Radix trigger this PR deleted, so it will fail the next time anyone runs that suite
**Attach:** `src/renderer/components/platform-bible-toolbar.tsx:366` · **Severity: High**

`e2e-tests/tests/isolated/bcv-open-resource-books-simple.spec.ts:212` asserts the open project's short name via `mainPage.locator('[data-slot="select-trigger"]').first()`. That attribute is emitted only by Radix `SelectTrigger`, which the base toolbar rendered (base `platform-bible-toolbar.tsx:191`) and this PR replaces with a `Button role="combobox"`. In Simple mode the toolbar now emits no `select-trigger` at all, so the locator either matches nothing or matches an unrelated select, and `toContainText(targetProject.shortName, { timeout: 60_000 })` burns a minute before failing. CI only runs `--project=smoke`, so this stays green until someone runs `test:e2e:isolated`. Update the spec to `getByRole('combobox')` in the same PR.

jolierabideau:
@katherinejensen00 — all 36 findings worked through. 31 fixed, 2 deferred with a ticket, 2 declined
with reasoning, 1 partially taken. The branch is also merged up to its latest base.

Head is now `fb23d51e8d9`. `npm run typecheck`, `npm run lint` and `npm test` are clean except for
two **pre-existing** failures in `rc-dock-tab-cache-patch.test.ts` — both files are byte-identical
to `origin/main`, they came in with PT-4701 (#2767), and nothing here touches them.

Two of your findings deserve a note before the list, because verifying them changed something:

**Finding 2 was real and I could reproduce it before fixing it.** A test that opens with zero
projects and presses Enter without arrowing failed exactly as you described. Your mechanism is
right: a `forceMount`ed item never registers, so with no rows registered cmdk has nothing
highlighted and its Enter branch is a no-op. Fixed by seeding the highlight *only* in that state —
`value` is `undefined` whenever a row is registered, so cmdk stays uncontrolled and keeps its own
first-item behavior.

**Finding 20's second item was right about the conclusion and wrong about the cause.** Adding
`triggerLabelFormat="shortNameAndFullName"` did not make the test falsifiable — I put a real derived
title back and it still passed. The actual reason is that Radix opens its tooltip from a pointer
sequence jsdom does not produce, so *any* hover-then-expect-nothing test there passes regardless.
It now asserts the absence of the `TooltipTrigger` wrapper, which is what `title: ''` actually
controls, plus a control case that has it. Verified it fails when the suppression is removed.

---

### High

1. **Stale E2E locator** — fixed. Located by role and accessible name now; the book/chapter control
   is also a `combobox`, so the name prefix is what disambiguates it.
2. **Enter with zero projects** — fixed, see above. Test added.
3. **Missing `es.json` keys** — fixed, with the two new format-string keys from finding 11 as well.
4. **Resources in the recents pipeline** — fixed. Your read of the mechanism held up: each candidate
   costs a `waitForNetworkObject` plus a retrying metadata fetch, serialized inside one 3s bound.
   The candidates are now checked concurrently and picked in recents order. The regression test
   (eight resources ahead of one editable project) fails at 3210 ms without the change and passes at
   686 ms with it.

### Medium

5. **Trigger greys out on every background refresh** — fixed at the consumer, gating on "this load
   has nothing to show yet" rather than "any load in flight". A settled empty list is not a loading
   state, so the escape hatch stays reachable.
6. **Case-insensitive selection in the trigger** — this landed on the base branch while the review
   was open (`findProjectById`), and `pairIsSelected` normalizes there too; both came in with the
   merge. I took your `renderTriggerLabel` TSDoc point: it now says `undefined` also means the
   selected id matched no entry of `projects`.
7. **`TruncatingTriggerText`** — deleted. `ToolbarCompoundLabel` takes `primary` alone and already
   has all four affordances, as you said.
8. **Out-of-order picks** — fixed. The bridge now also retires when the editor settles on something
   that is neither the pick nor what it reported when the pick was armed. The armed-against id is
   what keeps a bare mismatch from retiring it instantly at arm time.
9. **Recency failure retiring a good pick** — both halves fixed. `openProject` no longer lets the
   bookkeeping reject, and latest-wins is keyed on an attempt counter, so a slow rejection cannot
   clear a newer pick of the same project.
10. **Raw `%key%` text** — fixed. Every picker string resolves against an English fallback when the
    value still looks like a key.
11. **Concatenated accessible name** — fixed, with `%projectPicker_toolbar_trigger_label%` and an
    error variant. That also frees the name from reusing the placeholder as its prefix.
12. **Error and pending branches untested** — fixed. Two integration cases; I checked the pending one
    fails when the `!pendingProject &&` guard is dropped.
13. **Popover and dialog sorting** — fixed. The local bucket sorts by full name now, matching the
    dialog.
14. **Dialog fallback naming by raw id** — fixed your way: it opens unnamed and the trigger keeps
    naming what is open, rather than degrading to an id. That also removes the lost-padlock half.
15. **Footer announced as an option** — fixed with `aria-haspopup="dialog"`.
16. **Inter-section separator** — fixed, `alwaysRender`.
17. **Footer not actually pinned** — fixed. `sticky bottom-0` on a wrapper still inside `CommandList`,
    so keyboard reachability is unchanged. jsdom cannot see it; flagged for manual verification.
18/19. **Accessible name and read-only, both in the library** — **deferred, taking your offer.**
    Written up in `PT-4549-followup-projectselector-accessible-name.md` and recorded as consequences
    on `adr-project-selector-stays-experimental`, so it is discoverable from the decision that
    permits it. My reasoning for not taking it here: the ARIA-correct fix and a first-class
    read-only field are the same design question, and adding `isReadOnly` alongside
    `renderProjectIndicator` would leave two mechanisms for one marking — worse than either. That
    wants a decision about which survives, not an added field at the end of this PR.
20. **Three tests that can't fail** — all three fixed. The partition assertion now pins both buckets;
    the loading test asserts the real disabled state (moved to the integration file, where the real
    component runs); the tooltip one is rewritten, see the note above.
21. **Untranslated-defaults guard** — fixed. `PROJECT_SELECTOR_DEFAULT_STRINGS` is exported now (it
    wasn't, which is why the list was hand-copied), and the test loops all of it with `queryByText`,
    `queryByPlaceholderText` and `queryByLabelText`.
22. **Integration test running against `dist`** — fixed, and worth your attention since it is
    repo-wide: `vitest.config.ts` aliases `platform-bible-react/experimental` to source. Verified in
    both directions — deleting `forceMount` now fails the toolbar integration test, and the full
    root suite (282 files) gains no failures.
23. **Read-only path unasserted** — fixed, asserted end to end against the real component.
33. **Comments contradicting the code** — fixed. The `TruncatingTriggerText` one went with the
    component; the `use-project-picker-data` one now says the hook passes `isEditable` through and
    marks nothing.

### Low

24. **Re-picking re-stamps recency** — **declined.** Re-stamping a project the user just picked looks
    correct to me, the reorder happens after the popover has closed, and skipping the open would
    remove the `focus-existing` path. Say the word if you read the ordering differently.
25. **Inline `selection` literal** — fixed with a memo. I also corrected the `availableGroupings`
    rationale you flagged as inaccurate — a search keystroke re-partitions regardless.
26. **Footer misalignment** — fixed with empty stand-ins mirroring the row's leading slots, so the
    two stay aligned if that layout changes.
27. **Value-collision test** — fixed. It activates by keyboard now, which is the first-DOM-match route
    a collision could actually break.
28. **`ReadOnlyIndicator` re-forked** — fixed by extracting `LabelledGlyph`, which both use. `shrink-0`
    moved to the wrapper span, which is the flex item.
29. **Unread return value** — `displayedProject` is gone from `PendingProjectState`; the test harness
    derives it the way the component does, so the assertions still pin what a user sees. Log wording
    fixed too.
30. **The 15s clock vs the returned web-view id** — **declined**, on your own observation: the id
    says a web view was created, not that this window's editor reports the project, and reverting to
    the previous name while the new one loads is worse than the lag. The `apply-model.md` flag is
    fair; I think it is a question for the team rather than something to resolve in this PR.
31. **Duplicated mock setup** — the drift is fixed (the unit file had a legacy
    `{book, chapter, verse}` where the integration file's `SerializedVerseRef` is correct). I did not
    take the shared-setup module or the component extraction — both are test-architecture changes
    across two large files, and this PR is already wide. Happy to do it separately if you want it.
32. **`.integration.test.tsx`** — took the first branch: renamed to
    `platform-bible-toolbar-integration.test.tsx`, matching the two that already exist, rather than
    writing a second convention into the guide.
34. **Repeated rationales** — partially taken. Each of the four now has one home and a pointer
    elsewhere, the garbled story comment is rewritten, and the timeout says "not measured" once. I
    did not chase every one of the eight `forceMount` mentions; several are test comments explaining
    what that specific test pins, which I think earn their place.
35. **Backward-facing test comments** — fixed, restated as invariants.
36. **PR description** — rewritten, 23 KB to 6.6 KB. The review report is gone; what is left is the
    stacking note, the API delta, the decisions I want opinions on, what is deferred, the outstanding
    manual verification, and risk.

---

### Base merge

Merged `origin/pt-4548-promote-project-selector` (commit `5f659f952c3`). Two things worth knowing:

- The ADR log auto-merged **without a conflict** but put `adr-project-selector-footer-action-is-data`
  out of byte order, below `adr-project-selector-per-bucket-row-order`. No duplicate slugs, so
  nothing would have flagged it. Re-sorted.
- Both branches had independently added the same two id-casing tests to `project-selector.rows.test.ts`,
  which git merged into a duplicate pair. Kept one and folded in the dropped assertion.
- `dist/` was rebuilt rather than hand-merged.

Still outstanding and unchanged: the three manual items in the PR description (small-width layout,
keyboard focus handoff into the modal, and whether a read-only project's content actually loads).

katherinejensen00 (commented):
# PR Review Report: PR #2801 — Round 2

**Repository**: paranext/paranext-core
**Title**: PT-4549: Migrate the titlebar picker onto ProjectSelector
**Author**: @jolierabideau
**Head**: `72af27b186b` · **Base**: `main` (no longer stacked)
**Date**: 2026-09-21
**Mode**: dry-run — **not posted to GitHub** (per standing preference, the review is drafted for you to post yourself)
**Reviewers**: security · scope · ux · tests · clarity · prior-findings verification

---

## Round-1 disposition (36 findings)

| Status | Count | Findings |
|---|---|---|
| Fixed | 30 | 1–17, 20–23, 25, 26, 28, 29, 32, 33, 35, 36 |
| Deliberately deferred (recorded in ADR + PR body) | 2 | 18, 19 |
| Declined, defensibly | 1 | 30 |
| Partially fixed | 2 | 31, 34 |
| **Not fixed** | 1 | **24** |
| **Disputed — claimed fixed, disproved by mutation** | 1 | **27** |

Independently re-verified, not just taken from commit messages. Highlights worth trusting:

- **Finding 22 is genuinely fixed.** Renaming `data-testid="project-selector-footer-action"` in the component source turns 4 of the 11 integration tests red — the `vitest.config.ts` alias really does route the integration test to source, including on Windows (`new URL(...).pathname` → `/C:/Users/...` resolves).
- **Finding 2 is genuinely fixed**, and holds for the harder variant the fix didn't obviously cover (a typed query matching nothing while projects exist, with a stale highlight from arrowing first). Both variants pass against the real component. They remain untested in the repo — a coverage nit, not a defect.
- **Finding 4 is fixed** — concurrent candidate resolution, pinned by a test with 8 slow resources ahead of one usable project.

---

## Needs a decision from you

### D1. Prior finding 27 is still non-falsifiable — the "fix" didn't make it testable
`lib/platform-bible-react/src/components/advanced/project-selector/project-selector.component.test.tsx:941`

Proved by mutation: setting `FOOTER_ACTION_VALUE = 'More projects…'` — the exact collision the test names — leaves the test **green**. Project rows never derive their cmdk value from their text (`project-selector.component.tsx:592` sets ``value={`${row.rowKey} ${row.shortName} ${row.fullName}`}``), so a row's value can never equal a footer label whatever the sentinel is. The collision is structurally unreachable. The test is otherwise a line-for-line duplicate of "is reachable and activatable by keyboard", plus one `not.toHaveBeenCalled()` that also cannot fail.

**Ask:** delete it, or render a project whose composite value literally equals `platform.footerAction` so the sentinel is what separates them?

### D2. Prior finding 24 is un-addressed, and nothing records a decision to leave it
`src/renderer/hooks/use-pending-project.hook.ts:97`

`handleRowClick` fires `onChangeSelection` unconditionally in `project` mode (`project-selector.component.tsx:975`) → `handleSelectProject` → `beginOpenProject`. `isAlreadyCurrent` (`use-pending-project.hook.ts:77`) still only gates whether the bridge arms; `openProject(projectId)` runs unconditionally, so `recordProjectOpened` (`platform-bible-toolbar.tsx:623`) re-stamps the already-open project and reorders the Recent section under the user's cursor. Unlike 18/19/30, this one isn't written up anywhere as a deliberate call.

**Ask:** fix (skip the open when `isAlreadyCurrent`), or record it as accepted?

### D3. The repo-wide `vitest.config.ts` alias — scope
`vitest.config.ts:16`

Added in the round-2 fix commit. Re-points `platform-bible-react/experimental` from the committed `dist/` to source for **all 25** root test files, when only the new integration test motivated it. It's a real improvement, but it changes what those 25 files test, so anything it surfaces lands on this PR's blame. The author explicitly asks for a yes on this in the PR body.

**Ask:** land it here, or split it out with its own justification?

### D4. The cold-start recents walk rewrite — scope
`src/renderer/services/web-view.service-shard.ts:2022`

The fix for finding 4 rewrote the Simple-mode cold-start walk from sequential-stop-at-first-usable to fan-out-all-then-pick, in a service unrelated to the titlebar picker. The causal link is real (PT-4511 lets resources reach recents), but it now issues a `getMetadataForProject` lookup for **every** recents entry on every cold start, where before it commonly issued one. Bounded by `MAX_RECENT_PROJECTS` and rationalised in-code, so it's an accepted cost rather than an oversight — but it's a startup-path performance change justified by a hypothesis rather than a measurement.

**Ask:** in scope for PT-4549, or its own ticket with a measurement behind it?

### D5. `usePendingProject` as net-new capability sourced from a "Won't Do" item — scope
`src/renderer/hooks/use-pending-project.hook.ts:53`

142 lines of hook + 331 of test, a 15s bound and five retirement paths. The old `<Select>` had no pending-name behavior, so this is added capability, not migrated capability — and the PR body sources it to PT-4133, which is **Won't Do**, while shipping only half of what that item described. This is the single largest chunk of new complexity in the diff, and the author flags it for a product call.

**Ask:** ship inside PT-4549, or lift into its own work item?

---

## Warnings

### W1. The ADR cites a write-up that does not exist
`.context/standards/Architecture-Decisions.md:4042`

Both the ADR and the PR body cite `PT-4549-followup-projectselector-accessible-name.md` as where deferred findings 18 and 19 are recorded. That file is nowhere in the repo — `find`, `git log --all --diff-filter=A`, and a repo-wide grep all return only the ADR line itself. The deferral of two accessibility defects currently points at nothing. Add the file, or replace the citation with a Jira ticket.

### W2. A new toolbar test can pass without exercising the path it names
`src/renderer/components/platform-bible-toolbar.test.tsx:1686`

"opens a dialog-only project without naming it": `capturedDialogResolvers.at(-1)?.(projectId)` is optional-chained, so a missing resolver silently no-ops — and all three assertions already hold in the pre-act state (`allProjects: []`, `selection.projectId` already `undefined`, trigger already reads "Test no projects"). Dropping the dialog wiring entirely leaves it green. Add `expect(capturedDialogResolvers.length).toBeGreaterThan(0)` before resolving, or drop the `?.`, and assert a positive consequence.

### W3. Prior finding 31 — mock duplication reduced, not removed
`src/renderer/components/platform-bible-toolbar-integration.test.tsx:29`

The named drift is fixed, but there's still no shared setup module: the integration file duplicates 18 of the sibling's 20 `vi.mock` calls plus a second same-named `renderSimpleToolbarWith` with a different contract. `platform-bible-toolbar.tsx` also grew from 869 to **959** lines, still hosting `ProjectSelectorLabel` and `ToolbarProjectSelector`. Residual drift already visible: the unit file's `useProjectPickerData` default omits `currentSimpleProjectError`.

### W4. A comment contradicts the class it sits on
`src/renderer/components/projects/labelled-glyph.component.tsx:29`

The comment argues for "a shrinkable wrapper [that] lets a crowded row squeeze the glyph"; the class one line below is `tw:shrink-0`, which does the opposite. A reader trusting the comment could "fix" it to `shrink` and let the padlock collapse.

### W5. TSDoc cites a key the map doesn't contain
`src/renderer/components/platform-bible-toolbar.tsx:135`

`PICKER_STRING_FALLBACKS`' TSDoc illustrates itself with `'%projectPicker_title%'`, which isn't among its ten keys, isn't in `LOCALIZED_STRING_KEYS`, and is never resolved by `resolvePickerStrings` — it belongs to the `ProjectPicker` dialog. Cite a key the map actually holds.

---

## Suggestions

**A11y / UX**

- `project-selector.component.tsx:1306` — the sticky footer wrapper is a bare `<div>` between `CommandList` (`role="listbox"`) and the footer `CommandItem` (`role="option"`), breaking listbox ownership; some AT won't count the footer in "1 of N". `role="presentation"` fixes it without touching styling. (The `aria-haspopup="dialog"` at `:1330` correctly handles the other half.)
- `project-selector.component.tsx:601` — `data-selected` on a `CommandItem` is dead (cmdk spreads caller props first, then overwrites it) and actively misleading: in cmdk it means "highlighted", and `CommandInput`'s space-to-select handler queries exactly that attribute. Rename or drop.
- `platform-bible-toolbar.tsx:408` — with no projects the trigger's accessible name falls back to the bare placeholder, so a screen reader hears "No projects, combo box, collapsed" with no indication the control opens a picker. Consider a dedicated `_empty` key.
- `platform-bible-toolbar.tsx:223` — the translator-reorderability fix landed for the accessible name but not the **visible** text: `fullText` is still built by concatenation, so the two forms can disagree in any locale that reorders the pair, and the hard-coded parens aren't RTL-mirrored.
- `project-selector.component.tsx:1104` — this `useMemo` depends on the whole `props` object, so it never hits. `ToolbarProjectSelector` memoizes five props specifically so this component can skip work on every verse move; this dep list discards all of it. Cheap work, so not user-visible — but the memo reads as protection it doesn't provide.
- Prior finding 18 remains live as deferred: the padlock has no pointer-user explanation, and the row tooltip can't supply one — `ProjectRowView`'s tooltip renders only `fullName`/`scrollGroupScrRefLabel`/`disabledReason`, and a read-only project is not disabled, so for a short name the tooltip never opens at all.

**Tests**

- `web-view.service-shard.test.ts:1059` — nine real 500ms sleeps against the real 3s bound: ~500ms of pure sleeping added to every suite run, hence the bespoke `10000` timeout. Fake timers, or assert the concurrency directly (all nine lookups in flight before any resolves).
- `platform-bible-toolbar.test.tsx:1374` — assertions pin literal Tailwind class spellings (`tw:min-w-48` / `tw:min-w-24`, `tw:border-0`, `tw:w-fit`/`tw:w-96`). jsdom can't measure so there may be no better option, but shared exported constants would let the test compare values rather than spellings. Raised once at PR level.

**Clarity**

- `project-selector.component.tsx:179` — two stacked comment blocks on one declaration; the `//` preamble is now orphaned from the generated API docs for what became exported public surface. Fold into the TSDoc.
- `project-selector.component.tsx:1291` vs `:1307` — the `alwaysRender` rationale is written out in full twice ~15 lines apart (and restated in three test comments, the integration test, and the ADR). Keep one full explanation, cross-reference the other. This is the residue of prior finding 34.
- `project-selector.component.tsx:376` — "separated from the list" overstates it: the separator renders only when `filteredRows.length > 0`, i.e. not in the empty and no-match cases the very next paragraph says the footer exists for.

---

## Checked and found clean

- **Security: no findings.** No blocked imports, no secrets, no CSP hazards (`eval`/`innerHTML`/`dangerouslySetInnerHTML` all absent), error text reaching the UI is a fixed literal with raw messages going only to `logger`, and the dropped `isEditable` filter is not an authorization bypass — `isEditable` is display-only, the editor re-reads `platform.isEditable` from the PDP and the C# side derives it authoritatively. PAPI boundary respected throughout.
- **Forward-facing comments: clean.** Every changed file grepped for in-PR ticket IDs, PR numbers, review-finding IDs, stage tags and change narration. The only `PT-` references left are genuine forward pointers to open work (`TODO(PT-4552)` at `platform-bible-toolbar.tsx:650`, the `(PT-4552)` rationale at `:329`, pre-existing `TODO(PT-4233)` at `:865`). New test names are behavioural.
- **Dead code: none.** Retired `Select*` imports, `POINTER_EVENTS_INSIDE_SELECT_VALUE`, `ProjectSelectorTrigger`, `projectPickerItems`/`hasProjectPickerItems` fully removed; `%projectPicker_section_projects%` still live in the dialog, so not stranded.
- **Localization:** all four new keys present in both `en.json` and `es.json` with real Spanish, using `…` (U+2026).
- **E2E:** the isolated spec's locator was correctly migrated to `getByRole('combobox')` — prior finding 1 chased down rather than left to fail at merge.

---

## Not this PR's bug (pre-existing on `main`)

`project-selector.component.tsx:976, 1005, 1013, 1022` — picking a row closes the popover with `setOpen(false)` rather than `handleOpenChange(false)`, and Radix's controlled `Popover` doesn't fire `onOpenChange` on a programmatic prop change. So `setQuery('')` never runs on a pick: type a filter, pick a project, reopen — the search box is still pre-filled and the list still filtered. That is exactly what the comment at `:878-880` says is being prevented.

**This is pre-existing**: `origin/main` has the same four `setOpen(false)` sites and the same `handleOpenChange` query-clear, and this PR's new footer action correctly uses `handleOpenChange(false)`. Flagged because the titlebar migration makes it visible to every Simple-mode user, not because the PR caused it. Worth its own ticket.

---

## Verification notes

- Reviewed from a detached checkout of the PR head in the `code-review-beta` worktree. Mutation experiments (finding 27's sentinel, finding 22's test id) were reverted; `git status` over `lib/`, `src/`, `vitest.config.ts`, `e2e-tests/`, `assets/`, `.context/` is clean.
- Suites re-run locally: `platform-bible-toolbar-integration.test.tsx` (11 passed), `platform-bible-toolbar.test.tsx` + both hook suites (106 passed).
- `lib/platform-bible-react/dist/` excluded throughout as generated output.

> @katherinejensen00 partially reviewed 54 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2801#-:-P23zrgO-UfbKoYIywHf:b-gj29v7)* status: 56 of 57 files reviewed, 1 unresolved discussion (waiting on jolierabideau).

jolierabideau:
@katherinejensen00 — round 2 worked through. Five decisions settled, five warnings fixed, most of
the suggestions taken. Head is now `e19d80726bc`, merged up to latest `main` (conflicts were
confined to the generated `dist/` bundle, resolved by rebuild rather than hand-merge).

`npm run typecheck`, `npm run lint` and `npm test` are clean — **5518 passed, 0 failed**. The two
`rc-dock-tab-cache-patch.test.ts` failures I called pre-existing in round 1 are gone; they were
fixed on `main` and came in with this base merge.

### Decisions

**D1 — collision test.** Deleted. Your mutation is right and I could not construct a case that
makes it real: a row's value is `` `${rowKey} ${shortName} ${fullName}` ``, so no row value can
equal the sentinel whatever the sentinel is, and forcing one would mean fabricating a project whose
three fields concatenate to `platform.footerAction`. The sentinel still earns its place — it is
what makes the collision unreachable independently of how row values are composed later — so that
invariant is now recorded on `FOOTER_ACTION_VALUE` instead of asserted by a test that cannot fail.
The keyboard-reachability test beside it is the real coverage and stays.

**D2 — re-picking the open project.** Recorded as accepted, not fixed. Skipping the open when
`isAlreadyCurrent` would make the current row inert, and re-picking the open row is how a user
reveals or refocuses its existing editor. You were right that nothing said so: there is now a
comment at the call site naming both the intent and the cost (recency re-stamped, Recent can
reorder after the popover closes).

**D3 — the `vitest.config.ts` alias.** Landing here. Splitting it would leave this PR's own
integration test asserting against the committed `dist/`, which is the defect it was added to fix,
and you independently confirmed it routes to source on Windows too.

**D4 — cold-start recents walk.** Staying, with a measurement ticket filed rather than a bare
rationalisation. Your framing is the one I am acting on: it is an accepted cost, but the
justification is a hypothesis. The ticket asks for `PT_STARTUP_MARKS` timings on a full recents
list both ways, and for the in-code rationale to be replaced by the measurement or the shape
changed to a bounded-parallel walk.

**D5 — `usePendingProject`.** Shipping here. Without it the trigger lags visibly behind every pick,
which reads as a regression against the control it replaces even though the old `<Select>` had no
such mechanism — the old control simply did not have this gap because it did not open an editor.
The PT-4133 citation is the weak part of that argument and I have stopped leaning on it.

### Warnings

**W1 — the ADR cited a write-up that did not exist.** It existed; it had just never been committed.
It is now in the repo at `.context/designs/PT-4549-followup-projectselector-accessible-name.md`,
and the ADR line is a relative link to it rather than a bare filename. Good catch — the deferral of
two accessibility defects was pointing at nothing as far as anyone but me could tell.

**W2 — the dialog-only test could pass without its path.** Fixed all three ways you suggested:
`expect(capturedDialogResolvers.length).toBeGreaterThan(0)` before resolving, the `?.` dropped, and
a positive consequence asserted (`openScriptureEditor` called with the dialog's id). Verified by
mutation — deleting the `resolveProjectPickerDialogWith('far')` line now fails it, where before
every assertion held in the pre-act state.

**W3 — mock duplication.** The named drift is fixed: the unit file's `useProjectPickerData` default
now carries `currentSimpleProjectError`. I am still not taking the shared setup module or the
component extraction, for the same reason as round 1 — both are test-architecture changes across
two large files, and this PR is wide enough. `platform-bible-toolbar.tsx` at 959 lines is a fair
complaint; happy to do that as its own change if you want it.

**W4 — comment contradicting `shrink-0`.** Fixed. It now says why the rule belongs on the wrapper
(the wrapper is the flex item, so a rule on the icon leaves the wrapper free to collapse), rather
than arguing for the opposite of what the code does.

**W5 — TSDoc citing a key the map lacks.** Fixed, cites `%projectPicker_no_results%`.

### Suggestions

- **`role="presentation"` on the sticky wrapper** — taken. Listbox ownership restored without
  touching styling.
- **`data-selected` on rows** — dropped. You are right that it is dead (cmdk overwrites it) and
  actively misleading given `CommandInput`'s space-to-select queries that exact attribute. It is
  pre-existing on `main`, but it is one line and removing it is strictly safer than leaving it.
- **Empty-state accessible name** — taken, with a dedicated
  `%projectPicker_toolbar_trigger_label_empty%` key. A screen reader now hears "Select project, no
  projects on this computer" while the visible label stays the two-word placeholder; both are
  asserted.
- **Visible `fullText` still concatenated** — taken. Two more keys,
  `%projectPicker_toolbar_label_nameAndShortName%` and `%projectPicker_toolbar_label_shortNameOnly%`,
  so the brackets and the order are translator-controlled and the visible text can no longer
  disagree with the accessible name about either.
- **`useMemo` depending on the whole `props`** — taken. The two `'project'`-only members are
  narrowed above the memo so the dep list can name them, and the memo now depends on
  `props.mode`/`props.projects`/`props.selection` plus those. You were right that it read as
  protection it did not provide.
- **Nine real 500ms sleeps** — taken, and it turned into a better test than the timing one. Every
  lookup now parks on one gate and the test asserts all nine are in flight before any resolves, so
  concurrency is asserted directly instead of inferred. 5s → 203ms, bespoke timeout gone, and a
  sequential walk fails it at `expected [ 'proj-resource-0' ] to have a length of 9`.
- **Literal Tailwind spellings** — taken for the two this PR adds. They come from an exported
  `PROJECT_TRIGGER_MIN_WIDTH_CLASS` now, so the test compares values. The pre-existing `border-0` /
  `w-fit` / `w-96` assertions I left alone.
- **Two stacked comment blocks at `:179`**, **the duplicated `alwaysRender` rationale**, and
  **"separated from the list" at `:376`** — all three taken. One full explanation on the
  inter-section separator with the footer's pointing at it, the orphaned preamble folded into the
  exported symbol's TSDoc, and the footer's TSDoc now says what actually happens (pinned below,
  separator only when there are rows to divide).
- **Finding 18 still live** — yes, and now discoverable: both halves are in the committed write-up,
  including your point that the row tooltip cannot carry the padlock's explanation because a
  read-only project is not disabled.

### Pre-existing bug you flagged

The `setOpen(false)` vs `handleOpenChange(false)` query-clear — I reproduced your reasoning against
`origin/main` and agree it is pre-existing and worth its own ticket. Written up separately rather
than fixed here, since it touches four call sites in all three modes and wants its own test.

### Still outstanding

Unchanged from the PR description: the three manual items (small-width layout across shrink steps,
keyboard focus handoff into the "More projects…" modal, and whether a read-only project's content
actually loads — PT-4511's third criterion). I still cannot run those in this environment.

katherinejensen00 (approved):
Thanks for your hard work on this, Jolie! Looks good
:lgtm:

> @katherinejensen00 partially reviewed 35 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2801#-:-P24QY_V4nxeZQeXRAPK:b1rxf6e)* status: all files reviewed, 1 unresolved discussion (waiting on jolierabideau).

Issue #25, opened 2023-02-13
Create Logging Service


tjcouch-sil:
We need to add a logging service so we stop using `console.whatever` all the time. Apparently we already have some kind of electron-specific logging library. Let's use it!

Some discussion on the topic: https://github.com/paranext/paranext-core/pull/24/files#r1103914506

One important place is in `ServerNetworkConnector.onClientConnect`

irahopkinson:
@tjcouch-sil bultthe app and encountered this error:

![Image](https://user-images.githubusercontent.com/5582153/223888501-e8dfd227-cb70-424a-8ac9-8a9970f02106.png)

irahopkinson:
In `logger.ts`, I tried:

-  `log.initialize({ preload: true });` but that didn't change anything.
- `require('electron-log/renderer');` from the client but that didn't change anything either.

irahopkinson:
@tjcouch-sil From my shallow clone of the `electron-react-boilerplate` repo (which is where this repo is started from) I added a branch with minimal reproduction of our `logger.ts` but the production build didn't have the issue. To me this indicates (as you suggested) that it's something about our webpack config that we have changed.

irahopkinson:
Moved the `win-unpacked` folder up the tree to see the whole of the message:

![Image](https://user-images.githubusercontent.com/5582153/224865512-53eb5ec8-2405-48b4-9b70-9dd2604ed435.png)

PR #2802, opened 2026-09-10
PT-4099: Keyboard access and review fixes for the Find panel

timothy-mccormack:
### What was wrong

The Find filter controls are ordinary form controls — radio groups and checkboxes in
fieldsets — but they lived inside a Radix `DropdownMenuContent`, a `role="menu"` container.
A menu refuses keyboard access to anything that isn't a menu item: it calls `preventDefault`
on Tab, and its arrow handling only walks items registered in its roving-focus collection.
Focus landed on the panel and had nowhere to go, so the filters were mouse-only.

### The fix

The panel is a `Popover` now, which leaves the controls their native keyboard behavior:

| Key | Behavior |
| --- | --- |
| Tab | Move between groups; Radix loops it, so forward Tab stays in the panel |
| Shift+Tab | Move back; leaving the panel closes it and returns focus to the filter button |
| ↑ / ↓ | Move within a radio group and change the selection |
| Space | Toggle a checkbox |
| Esc | Close, returning focus to the filter button |

Closing the panel returns focus to the filter button when a keyboard exit closed it — Escape, or
the Shift+Tab off the first control. Every other dismissal leaves focus where it went.
Without that, Shift+Tab left the panel and dismissed it behind the user, stranding them at the
bottom of the Find panel. A focus trap would also solve it, but the repo's dismissal guidelines
say a non-modal surface must not trap focus (WCAG 2.1.2), and trapping hides the rest of the panel
from screen readers — which silences the results area exactly when a filter re-runs the search.

Explanations (the ⓘ icons) are attached to the controls they explain with `aria-describedby`,
so a screen reader reads them on "Any text" and "Ignore whitespace differences". They stay
hover tooltips visually. A tab stop of their own would sit inside the radio group, where Tab
never reaches it whenever the other radio is the selected one.

### Also in this PR, and why

- **Tooltips rendered behind the panel.** `PopoverContent` sits above the tooltip layer, so
  the panel painted over its own explanations. #2750 has since merged and raised the shared
  tooltip tier to 675, above the popover tier — this branch merges `main` and relies on that,
  with no local override left behind.
- **The panel had no accessible name.** A popover isn't named by its trigger the way a menu is.
- **Focus stays where it is put when the panel dismisses.** The panel is non-modal, so focus
  landing outside it closes it — including focus the app moves itself. Returning focus to the
  trigger is gated on a Tab or Escape keydown originating inside the panel, the case that return
  exists for; every other dismissal leaves the caret alone. Without that gate, invoking Find while
  the panel was open put the caret in the search box and then pulled it back onto the filters
  button a macrotask later, defeating `useFocusSearchOnInvoke`. Both branches are tested, draining
  Radix's unmount timer first — asserting sooner reads the caret mid-flight and passes either way.
- **One existing e2e test changed.** `Search History › should add the term in the box to history
  when options change` filled the search box with the panel open and kept using it, which worked
  only while the panel was modal. It now reopens the panel for its second option change; what it
  asserts is unchanged. `openFiltersPanel` converges on "open" using the `aria-expanded` check this
  file already uses, so no caller depends on knowing whether a dismissal has landed yet.
- **A platform fix, outside Find.** The content-zoom bootstrap (PT-4576) appends its indicator's
  live region with `position:absolute` and no offsets, so it keeps its static position at the end of
  the body's flow and counts toward scrollable overflow there. Every web view gets that bootstrap, so
  any pane whose content fills its viewport gains a document scrollbar — which takes its own width
  out of the viewport as it toggles, moving every right-aligned control. In Find that put the 14px
  toolbar shift back, the exact thing this PR's e2e test exists to catch. Pinning the region to the
  origin keeps it visually hidden and in the accessibility tree with nothing to overflow, and a unit
  test covers it.
- **Spacing.** `PopoverContent` lays its children out as a flex column with a gap, which stacked
  on each fieldset's own margin. `gap-0` restores the original 12px.
- **Two filters never re-ran the search.** "Ignore whitespace differences" and "Ignore
  diacritics" were missing from the auto-search effect's trigger list (pre-existing).
- **Layout.** The panel root capped its height but never contained its overflow, so the
  *document* scrolled, and its scrollbar toggled on every filter-driven re-run — each toggle
  narrowing the viewport and sliding the toolbar (and the popover anchored to it) 14px. The
  root now scrolls itself; the results list keeps a small floor, dropped while the loading skeleton
  is up so the empty list cannot reserve space that pushes the root into scrolling.

### Appearance

The panel is a plain opaque popover now. It no longer has the menus' translucent "glass"
treatment, and it animates on open like the popovers beside it in the same toolbar. That
matches the direction in PT-4535 (menus and popovers opaque). Spacing matches the old menu.

### Testing

- Unit tests covering focus on open, Tab between groups, Shift+Tab and Escape handing focus back
  to the filter button, arrow-key selection, the panel's name, both explanations' descriptions,
  the groups being named by their legends, and tooltip stacking. The Shift+Tab test fails without
  the focus-return handler; the Escape one guards behavior Radix already provides.
- An e2e test that samples the filters button's position every animation frame across a
  filter-driven re-run. It caught a real 14px shift, which is fixed here.
- The Storybook narrow-panel guard now measures the Find root, the element that actually scrolls.

### Not covered

- The two search triggers have no regression test. An e2e test would depend on search data that
  varies; happy to add one if you'd rather have it before merge.
- A panel under roughly 360px tall can still shift during a re-run. `scrollbar-gutter: stable`
  would fix it at any height, at the cost of a permanent ~14px strip down the right side.
- Recent searches keyboard access and the reference picker are **not** here: #2750 (PT-4345)
  covers both and has merged, so this PR no longer touches `platform-bible-react` at all.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2802)

### DIFF
--- a/extensions/src/platform-scripture/src/find/find-filters.component.tsx
+++ b/extensions/src/platform-scripture/src/find/find-filters.component.tsx
@@ -2,10 +2,10 @@ import { Info, SlidersHorizontal } from 'lucide-react';
 import {
   Button,
   Checkbox,
-  DropdownMenu,
-  DropdownMenuContent,
-  DropdownMenuTrigger,
   Label,
+  Popover,
+  PopoverContent,
+  PopoverTrigger,
   RadioGroup,
   RadioGroupItem,
   Tooltip,
@@ -74,12 +74,17 @@ export function FindFilters({
   open,
   onOpenChange,
 }: FindFiltersProps) {
+  // These filters are a form of grouped settings, not a menu of commands. A menu container would
+  // give them `role="menu"`, whose keyboard model only navigates registered menu items and calls
+  // preventDefault on Tab — leaving these plain form controls unreachable by keyboard. A popover
+  // is an inert container, so the controls keep their native keyboard behavior: Tab moves between
+  // groups and arrow keys move within a radio group.
   return (
-    <DropdownMenu open={open} onOpenChange={onOpenChange}>
+    <Popover open={open} onOpenChange={onOpenChange}>
       <TooltipProvider>
         <Tooltip>
           <TooltipTrigger asChild>
-            <DropdownMenuTrigger asChild>
+            <PopoverTrigger asChild>
               <Button
                 variant="outline"
                 size="icon"
@@ -88,12 +93,12 @@ export function FindFilters({
               >
                 <SlidersHorizontal className="tw:h-4 tw:w-4" />
               </Button>
-            </DropdownMenuTrigger>
+            </PopoverTrigger>
           </TooltipTrigger>
           <TooltipContent>{localizedStrings.toggleFilters}</TooltipContent>
         </Tooltip>
       </TooltipProvider>
-      <DropdownMenuContent align="end" className="tw:w-72 tw:p-3">
+      <PopoverContent align="end" className="tw:w-72 tw:p-3">
         {/* 1. Match content in */}
         <fieldset className="tw:mb-3">
           <legend className="tw:px-2 tw:py-1.5 tw:text-sm tw:font-semibold">
@@ -252,7 +257,7 @@ export function FindFilters({
             </Label>
           </div>
         </fieldset>
-      </DropdownMenuContent>
-    </DropdownMenu>
+      </PopoverContent>
+    </Popover>
   );
 }
--- a/extensions/src/platform-scripture/src/find/find-filters.component.tsx
+++ b/extensions/src/platform-scripture/src/find/find-filters.component.tsx
@@ -98,7 +98,14 @@ export function FindFilters({
           <TooltipContent>{localizedStrings.toggleFilters}</TooltipContent>
         </Tooltip>
       </TooltipProvider>
-      <PopoverContent align="end" className="tw:w-72 tw:p-3">
+      {/* The height cap and internal scrolling are load-bearing, and PopoverContent — unlike
+          DropdownMenuContent — does not supply them. Without them this panel outgrows a short web
+          view, so moving focus down it scrolls the document instead, and the popper repositions on
+          every keypress. Scrolling inside the panel keeps it anchored. */}
+      <PopoverContent
+        align="end"
+        className="tw:max-h-(--radix-popover-content-available-height) tw:w-72 tw:overflow-x-hidden tw:overflow-y-auto tw:p-3"
+      >
         {/* 1. Match content in */}
         <fieldset className="tw:mb-3">
           <legend className="tw:px-2 tw:py-1.5 tw:text-sm tw:font-semibold">
--- a/extensions/src/platform-scripture/src/find/find-filters.component.tsx
+++ b/extensions/src/platform-scripture/src/find/find-filters.component.tsx
@@ -101,10 +101,16 @@ export function FindFilters({
       {/* The height cap and internal scrolling are load-bearing, and PopoverContent — unlike
           DropdownMenuContent — does not supply them. Without them this panel outgrows a short web
           view, so moving focus down it scrolls the document instead, and the popper repositions on
-          every keypress. Scrolling inside the panel keeps it anchored. */}
+          every keypress. Scrolling inside the panel keeps it anchored.
+
+          `overflow-y-scroll` rather than `auto` because the popper recomputes the available height
+          on each re-render, and a filter change re-renders; under `auto` those small differences
+          take the content across the threshold where a scrollbar is needed and the bar flickers in
+          and out. This content is taller than the space it gets, so the scrollbar is always
+          warranted — showing it unconditionally just stops it blinking. */}
       <PopoverContent
         align="end"
-        className="tw:max-h-(--radix-popover-content-available-height) tw:w-72 tw:overflow-x-hidden tw:overflow-y-auto tw:p-3"
+        className="tw:max-h-(--radix-popover-content-available-height) tw:w-72 tw:overflow-x-hidden tw:overflow-y-scroll tw:p-3"
       >
         {/* 1. Match content in */}
         <fieldset className="tw:mb-3">
--- a/extensions/src/platform-scripture/src/find/find-filters.component.tsx
+++ b/extensions/src/platform-scripture/src/find/find-filters.component.tsx
@@ -110,7 +110,7 @@ export function FindFilters({
           warranted — showing it unconditionally just stops it blinking. */}
       <PopoverContent
         align="end"
-        className="tw:max-h-(--radix-popover-content-available-height) tw:w-72 tw:overflow-x-hidden tw:overflow-y-scroll tw:p-3"
+        className="tw:max-h-(--radix-popover-content-available-height) tw:w-72 tw:overflow-x-hidden tw:overflow-y-auto tw:p-3"
       >
         {/* 1. Match content in */}
         <fieldset className="tw:mb-3">
--- a/extensions/src/platform-scripture/src/find/find.component.test.tsx
+++ b/extensions/src/platform-scripture/src/find/find.component.test.tsx
@@ -882,3 +882,51 @@ describe('Find — whitespace and diacritic tolerance toggles', () => {
     },
   );
 });
+
+// These exist because the filter controls are plain form controls (radio groups, checkboxes), not
+// menu items. Housing them in a Radix menu makes them unreachable by keyboard: menu content calls
+// preventDefault on Tab, and its arrow handling only walks items registered in its roving-focus
+// collection — which plain form controls never join. A regression back to a menu container would
+// leave the panel mouse-only, and nothing else in this suite would notice.
+describe('Find — filters panel keyboard accessibility', () => {
+  /** Opens the filters panel, which is where all the filter controls live */
+  async function openFilters(user: ReturnType<typeof setupUser>) {
+    await user.click(screen.getByRole('button', { name: 'Toggle filters' }));
+  }
+
+  it('puts focus on the first filter control as soon as the panel opens', async () => {
+    const user = setupUser();
+    render(<Find {...buildLifecycleProps({})} />);
+
+    await openFilters(user);
+
+    expect(screen.getByRole('radio', { name: 'Any text' })).toHaveFocus();
+  });
+
+  // Each radio group is a single tab stop (roving tabindex), so Tab crosses between groups rather
+  // than visiting every radio — which is why the second stop is the next group, not the next radio.
+  it('moves focus to the next group of controls when the user presses Tab', async () => {
+    const user = setupUser();
+    render(<Find {...buildLifecycleProps({})} />);
+
+    await openFilters(user);
+    await user.tab();
+
+    expect(screen.getByRole('radio', { name: 'Anywhere' })).toHaveFocus();
+  });
+
+  // Asserts focus movement rather than selection. Radix selects a radio on arrow-navigation from a
+  // `focus` handler gated on a flag set by a non-capturing `document` keydown listener, which React
+  // sets only after its own delegated handler has already moved focus — so the selection half of
+  // that behavior cannot be reproduced under jsdom. Navigation is the part this component controls,
+  // and the part that was broken.
+  it('moves focus between radio options when the user presses the down arrow', async () => {
+    const user = setupUser();
+    render(<Find {...buildLifecycleProps({})} />);
+
+    await openFilters(user);
+    await user.keyboard('{ArrowDown}');
+
+    expect(screen.getByRole('radio', { name: 'Verse text only' })).toHaveFocus();
+  });
+});
--- a/extensions/src/platform-scripture/src/find/find.component.tsx
+++ b/extensions/src/platform-scripture/src/find/find.component.tsx
@@ -834,7 +834,15 @@ export function Find({
   };
 
   return (
-    <div className="pr-twp tw:mx-auto tw:flex tw:flex-col tw:gap-4 tw:p-4 tw:min-w-[10rem] tw:max-h-screen">
+    // Scrolling here rather than letting the document scroll, with the scrollbar shown
+    // unconditionally. A search re-runs whenever a filter changes, so the panel's height crosses the
+    // viewport boundary often; when the document was the scroller, its scrollbar appeared and
+    // disappeared on each change, and every toggle narrowed the viewport by the scrollbar's width.
+    // That shifted this right-aligned toolbar sideways, which anything anchored to it — the filters
+    // popover — visibly followed. `scroll` rather than `auto` because merely reserving the gutter
+    // holds the width but still lets the bar itself blink in and out at that same threshold.
+    // Clipping the overflow would hold the width too, but silently hides content in a short panel.
+    <div className="pr-twp tw:mx-auto tw:flex tw:flex-col tw:gap-4 tw:overflow-y-scroll tw:p-4 tw:min-w-[10rem] tw:max-h-screen">
       {/* Header with searchbar and filters */}
       <div className="tw:space-y-3">
         {/* Project selector + Find/Replace toggle share one row. The responsiveness guideline caps a
@@ -1195,7 +1203,11 @@ export function Find({
       {/* eslint-disable-next-line jsx-a11y/no-static-element-interactions */}
       <div
         ref={resultsContainerRef}
-        className="tw:min-h-48 tw:flex-1 tw:space-y-2 tw:overflow-y-auto tw:pe-2"
+        // `overflow-y-scroll` rather than `auto`: changing a filter re-runs the search, which empties
+        // this list and refills it, so under `auto` the scrollbar disappears and comes back on every
+        // change and visibly blinks. Showing it unconditionally costs an empty track when results are
+        // few, which is quieter than a bar flashing on each keystroke.
+        className="tw:min-h-48 tw:flex-1 tw:space-y-2 tw:overflow-y-scroll tw:pe-2"
         // This div is a keyboard-navigable scroll container; tabIndex is required to receive focus for arrow-key navigation between results
         // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
         tabIndex={0}
--- a/extensions/src/platform-scripture/src/find/find.component.tsx
+++ b/extensions/src/platform-scripture/src/find/find.component.tsx
@@ -834,15 +834,7 @@ export function Find({
   };
 
   return (
-    // Scrolling here rather than letting the document scroll, with the scrollbar shown
-    // unconditionally. A search re-runs whenever a filter changes, so the panel's height crosses the
-    // viewport boundary often; when the document was the scroller, its scrollbar appeared and
-    // disappeared on each change, and every toggle narrowed the viewport by the scrollbar's width.
-    // That shifted this right-aligned toolbar sideways, which anything anchored to it — the filters
-    // popover — visibly followed. `scroll` rather than `auto` because merely reserving the gutter
-    // holds the width but still lets the bar itself blink in and out at that same threshold.
-    // Clipping the overflow would hold the width too, but silently hides content in a short panel.
-    <div className="pr-twp tw:mx-auto tw:flex tw:flex-col tw:gap-4 tw:overflow-y-scroll tw:p-4 tw:min-w-[10rem] tw:max-h-screen">
+    <div className="pr-twp tw:mx-auto tw:flex tw:flex-col tw:gap-4 tw:p-4 tw:min-w-[10rem] tw:max-h-screen">
       {/* Header with searchbar and filters */}
       <div className="tw:space-y-3">
         {/* Project selector + Find/Replace toggle share one row. The responsiveness guideline caps a
@@ -1203,11 +1195,7 @@ export function Find({
       {/* eslint-disable-next-line jsx-a11y/no-static-element-interactions */}
       <div
         ref={resultsContainerRef}
-        // `overflow-y-scroll` rather than `auto`: changing a filter re-runs the search, which empties
-        // this list and refills it, so under `auto` the scrollbar disappears and comes back on every
-        // change and visibly blinks. Showing it unconditionally costs an empty track when results are
-        // few, which is quieter than a bar flashing on each keystroke.
-        className="tw:min-h-48 tw:flex-1 tw:space-y-2 tw:overflow-y-scroll tw:pe-2"
+        className="tw:min-h-48 tw:flex-1 tw:space-y-2 tw:overflow-y-auto tw:pe-2"
         // This div is a keyboard-navigable scroll container; tabIndex is required to receive focus for arrow-key navigation between results
         // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
         tabIndex={0}
--- a/extensions/src/platform-scripture/src/find/find.component.tsx
+++ b/extensions/src/platform-scripture/src/find/find.component.tsx
@@ -834,7 +834,13 @@ export function Find({
   };
 
   return (
-    <div className="pr-twp tw:mx-auto tw:flex tw:flex-col tw:gap-4 tw:p-4 tw:min-w-[10rem] tw:max-h-screen">
+    // `overflow-hidden` makes `max-h-screen` hold, so the document never scrolls. It otherwise does,
+    // and since a search re-runs whenever a filter changes, the results area crosses the viewport
+    // boundary often enough that the document scrollbar toggles. Each toggle narrows the viewport by
+    // the scrollbar's width, shifting this right-aligned toolbar sideways — which anything anchored
+    // to it, such as the filters popover, visibly follows. Nothing is clipped: the results list below
+    // shrinks to the space left over and scrolls within it.
+    <div className="pr-twp tw:mx-auto tw:flex tw:flex-col tw:gap-4 tw:overflow-hidden tw:p-4 tw:min-w-[10rem] tw:max-h-screen">
       {/* Header with searchbar and filters */}
       <div className="tw:space-y-3">
         {/* Project selector + Find/Replace toggle share one row. The responsiveness guideline caps a
@@ -1195,7 +1201,11 @@ export function Find({
       {/* eslint-disable-next-line jsx-a11y/no-static-element-interactions */}
       <div
         ref={resultsContainerRef}
-        className="tw:min-h-48 tw:flex-1 tw:space-y-2 tw:overflow-y-auto tw:pe-2"
+        // `min-h-0` rather than a fixed floor so this list can shrink below its content when the
+        // panel is short. A floor here would refuse to give way and push the root past its height
+        // cap, which is what previously forced the document to scroll. The empty-state placeholder
+        // carries its own minimum height, so the idle view is unaffected.
+        className="tw:min-h-0 tw:flex-1 tw:space-y-2 tw:overflow-y-auto tw:pe-2"
         // This div is a keyboard-navigable scroll container; tabIndex is required to receive focus for arrow-key navigation between results
         // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
         tabIndex={0}


extensions/src/platform-scripture/src/find/find-filters.component.tsx
```diff
@@ -88,12 +93,25 @@ export function FindFilters({
...
           <TooltipContent>{localizedStrings.toggleFilters}</TooltipContent>
         </Tooltip>
       </TooltipProvider>
-      <DropdownMenuContent align="end" className="tw:w-72 tw:p-3">
+      {/* The height cap and internal scrolling are load-bearing, and PopoverContent — unlike
+          DropdownMenuContent — does not supply them. Without them this panel outgrows a short web
+          view, so moving focus down it scrolls the document instead, and the popper repositions on
+          every keypress. Scrolling inside the panel keeps it anchored.
+
+          `overflow-y-scroll` rather than `auto` because the popper recomputes the available height
```
lyonsil, line 106: **#15** - low · confirmed

A comment added by this PR argues for overflow-y-scroll while line 113 ships overflow-y-auto.

The paragraph at find-filters.component.tsx:106-110 says "overflow-y-scroll rather than auto because ... This content is taller than the space it gets, so the scrollbar is always warranted - showing it unconditionally just stops it blinking", but the shipped class is tw:overflow-y-auto. A maintainer will either believe the flicker is fixed or "restore" scroll and silently undo the revert. The first paragraph of the comment block, about the height cap being load-bearing, is still accurate.

**Suggested fix** - delete the second paragraph of the comment block (lines 106-110, beginning "`overflow-y-scroll` rather than `auto` because"). The first paragraph, about the height cap and internal scrolling being load-bearing, is still accurate and should stay.

This is branch history rather than a mistake in the code: commit `7c564d1959b` added the paragraph together with `overflow-y-scroll`, and `38fd1acd926` reverted only the class, leaving the justification behind.

_Verification: Git history confirms the exact mechanism alleged: commit 7c564d1959b added both the comment and overflow-y-scroll together, and commit 38fd1acd926 ("Revert the Find panel scroll changes, keeping the scope to the filters") changed only the class back to overflow-y-auto, leaving the comment paragraph untouched in its diff. Both commits are inside this PR's branch between MERGE_BASE and HEAD, so this is an in-PR-introduced contradiction sitting on a changed line. The revert commit message explains why the scroll containment change was reverted but never mentions the orphaned comment. Only the second paragraph is false; the first (height cap / internal scrolling rationale) remains accurate._

Similar fix as #33.

_Other findings in this file: #1, #5, #6, #19_

lyonsil (commented):
Review of the filters-panel container swap and the layout containment added after it. Thanks for the unusually clear PR description - the measured `x: 519 -> 505` is what made the layout half quick to verify.

Every comment below carries a suggested fix. Each finding was checked against the branch before being included, and where that check reduced its severity the header says so and gives the reason - several started higher and were brought down.

Two of these are worth reading first: the tooltip stacking order, and the root `overflow-hidden`. The rest are small.

**#3** (high -> medium · downgraded on review) extensions/src/platform-scripture/src/find.web-view.tsx:1637 - The auto-search effect omits ignoreWhitespaceDifferences and ignoreDiacritics, so toggling either checkbox never re-runs the search.

Deps are [searchTerm, shouldMatchCase, wordRestriction, isRegexAllowed, searchTextType, relevantScopeKey, requestAutoSearchWhenVisible]. relevantScopeKey (find.web-view.tsx:1108-1113) memoizes only scope/selectedBookIds/verseRef book and chapter, so it carries neither filter; useRunWhenVisible returns a useCallback with an empty dep array, so it cannot smuggle them in. The sibling add-to-history effect at find.web-view.tsx:1116-1130 does list both, which is what makes the omission look accidental. Pre-existing and untouched by this PR, but this PR is what makes both checkboxes keyboard-reachable, and react-hooks/exhaustive-deps can never catch it because the effect body references none of the deps.

**Suggested fix** - add the two missing deps:

```ts
}, [
  searchTerm,
  shouldMatchCase,
  wordRestriction,
  isRegexAllowed,
  searchTextType,
  ignoreWhitespaceDifferences,
  ignoreDiacritics,
  relevantScopeKey,
  requestAutoSearchWhenVisible,
]);
```

The sibling add-to-history effect at `find.web-view.tsx:1116-1130` already lists both, which is what makes the omission look accidental. Note `react-hooks/exhaustive-deps` cannot catch this: the effect body references none of these values, so the trigger list is invisible to the linter. This is pre-existing, not introduced here - flagging it because this PR is what makes both checkboxes keyboard-reachable, so it will be hit more often now.

_Why this is downgraded: Every link verified. The deps array is byte-identical at MERGE_BASE and this PR touches only find-filters.component.tsx, find.component.tsx and the test file, so the defect is pre-existing; the checkboxes were already mouse-clickable inside the old DropdownMenuContent, so keyboard reach adds a route to an already-reachable bug rather than exposing it for the first time. Running the existing "tolerance toggles" describe (find.component.test.tsx:844-884, 6 passed) neither proves nor refutes it - those tests only assert the setter is called and the button gets tw:bg-muted, never a re-run. The PR's own premise survives too: the other four filter controls plus scope do re-run the search, so the scrollbar-toggling rationale for overflow-hidden still holds, and the user can recover with Enter. Note the correct path is src/find.web-view.tsx, not src/find/._

**#22** (medium -> low · downgraded on review) extensions/src/platform-scripture/src/find/find-filters.component.tsx:120 - Neither RadioGroup has an accessible name; a <legend> names the fieldset's group role, not the nested radiogroup.

A jsdom probe shows both role="radiogroup" nodes with aria-label and aria-labelledby null and computeAccessibleName() returning "". Pre-existing markup, unchanged by this PR, but the controls are now keyboard-reachable for the first time.

**Suggested fix** - one prop on each `RadioGroup`, reusing the strings already rendered in the legends:

```tsx
<RadioGroup aria-label={localizedStrings.matchContentIn} value={searchTextType} ... >
```

Low value: a screen reader already announces the surrounding `<fieldset>` as a group named by its `<legend>`, so this adds redundancy rather than missing information, and axe does not flag it (`radiogroup` has `accessibleNameRequired: false`). Pre-existing markup, and the same pattern is used in `replace-preview-options.component.tsx:175`. Take it only if you want the panel spec-clean.

_Why this is downgraded: The mechanics are right - there is genuinely no name-from-ancestor for radiogroup. But the "no group context" impact claim is wrong: the same probe shows the wrapping fieldset exposed as role group with accName "Match content in" / "Match boundaries" from its legend, which is exactly what NVDA/JAWS/VoiceOver announce on entering a fieldset, so a radiogroup name would be redundancy rather than the only source of context. It is also strictly pre-existing: git show 3529e673 gives byte-identical fieldset/legend/RadioGroup markup with no aria-* on either RadioGroup, and neither anchor line is in the PR's changed ranges. Severity is further limited by tooling and convention: axe-core's role table sets radiogroup accessibleNameRequired:false (aria-query says true, so it is a spec-purity gap, not an automated-checker violation), and replace-preview-options.component.tsx:175/:215 uses the identical unnamed pattern._

_Not anchored: no longer on a changed line; cannot be posted inline (ambiguous: excerpt also matches at line(s) [164]; nearest to 120 chosen)_

**#23** (medium -> low · downgraded on review) extensions/src/platform-scripture/src/find/find-filters.component.tsx:145 - The two Info tooltip triggers are not focusable, so the only text explaining two filters stays keyboard-unreachable.

TooltipTrigger asChild wrapping a bare lucide <Info /> svg produces no focusable element. Pre-existing markup the PR leaves rather than introduces.

**Suggested fix** - give each `Info` a focusable host so keyboard and screen-reader users can reach the explanation:

```tsx
<TooltipTrigger asChild>
  <button type="button" aria-label={localizedStrings.allTextTooltip} className="tw:cursor-help">
    <Info className="tw:h-3.5 tw:w-3.5 tw:text-muted-foreground" />
  </button>
</TooltipTrigger>
```

`TooltipTrigger asChild` on a bare `<svg>` produces nothing focusable, so the tab ring inside the panel currently skips both icons. Pre-existing markup that this PR did not introduce - before the fix the whole panel was keyboard-unreachable anyway - so it is a follow-up rather than a regression. Worth doing here only because this PR is the one that makes the panel keyboard-operable, and these two tooltips hold the panel's only explanatory copy.

_Why this is downgraded: A jsdom probe confirms both Info triggers render as bare svg.lucide-info with tabindex=null / tabIndex === -1, and the real tab ring inside the open panel is searchTextType-all, wordRestriction-none, matchCase, ignoreWhitespaceDifferences, ignoreDiacritics, allowRegex, then loops - no svg is ever focused (the generator's claimed ring had a spurious "wrap" tail). The text really is otherwise unreachable: neither Info carries aria-describedby or aria-label, and the two tooltip strings appear nowhere in the DOM at rest. But git show 3529e673 has byte-identical TooltipTrigger + Info markup at lines 127-129 and 213-215, the anchor lines are not in the PR's changed ranges, and before the PR the whole panel was keyboard-unreachable anyway - so the PR strictly improves this area rather than introducing the gap. A non-blocking nit: a defect this PR left rather than created._

**#25** (low · confirmed) extensions/src/platform-scripture/src/find/find.component.test.tsx:845 - A JSDoc in the test file still calls the filters panel a "dropdown" - the container this PR removed.

find.component.test.tsx:845 reads "/** Opens the filters dropdown, which is where both toggles live */", inside the pre-existing whitespace/diacritic describe. The PR's own new helper at line 892 correctly says "panel".

**Suggested fix** - one word on line 845: "Opens the filters **dropdown**" -> "Opens the filters **panel**", matching the wording of the helper you added at line 892. This is the only stale reference to the old container left in the Find sources; the other `dropdown`/`menu` hits belong to the project selector and the per-result menu and are correct as they are.

_Verification: Confirmed by reading the file: line 845 is the stale JSDoc, and git diff against MERGE_BASE shows the whitespace/diacritic describe is untouched by this PR (the diff only adds new lines starting at 882), so it sits on an unchanged line. A grep for dropdown|menu across extensions/src/platform-scripture/src/find/ turns up no other stale reference to this filters panel: find.component.tsx's hideFilterMenu/alignDropDown/"funnel menu" belong to the unrelated ProjectSelector's own filter menu, search-result.component.tsx's DropdownMenuItem usages are the per-result copy/dismiss menu, and find-filters.component.tsx's new comments deliberately contrast Popover with DropdownMenu as forward-facing rationale. Real count of stale references: exactly 1._

extensions/src/platform-scripture/src/find/find-filters.component.tsx
```diff
@@ -88,12 +93,25 @@ export function FindFilters({
...
+          on each re-render, and a filter change re-renders; under `auto` those small differences
+          take the content across the threshold where a scrollbar is needed and the bar flickers in
+          and out. This content is taller than the space it gets, so the scrollbar is always
+          warranted — showing it unconditionally just stops it blinking. */}
+      <PopoverContent
```
lyonsil, line 111: **#1** - high · confirmed

The panel's two Info tooltips now render behind the opaque panel, hiding its only explanatory copy.

PopoverContent sets style zIndex Z_INDEX_ABOVE_DOCK (600) inside a bare Portal; TooltipContent sets Z_INDEX_TOOLTIP (550) inside its own Portal. Radix's popper lifts each content's z-index onto a body-level wrapper, so the two are stacking siblings and the panel paints over the tooltip. The removed DropdownMenuContent carried tw:z-50, so 50 < 550 and the tooltips won before. --popover is fully opaque and Tooltip defaults to side="top", so each tooltip renders upward entirely inside the 288px panel. z-index.ts:10-12 documents this exact hazard. In-repo fix precedent: style={{ zIndex: Z_INDEX_ABOVE_POPOVER }} as at footnote-caller-dropdown.component.tsx:183.

**Suggested fix** - give both in-panel tooltips the escape hatch the repo already has for exactly this case. `Z_INDEX_ABOVE_POPOVER` (650) is exported from `platform-bible-react`, and `z-index.ts` documents it as the constant for content portalled out of a popover:

```tsx
import { Z_INDEX_ABOVE_POPOVER } from 'platform-bible-react';
// ...
<TooltipContent style={{ zIndex: Z_INDEX_ABOVE_POPOVER }}>
  <p className="tw:max-w-xs">{localizedStrings.allTextTooltip}</p>
</TooltipContent>
```

Apply to both tooltips inside the panel (around lines 146 and 232). Leave the trigger-button tooltip near line 98 alone - it is outside the panel and already renders correctly. Precedent: `footnote-caller-dropdown.component.tsx:183`.

_Verification: Verified empirically: rendering FindFilters at PR head in jsdom shows the popover wrapper and the tooltip wrapper as direct children of document.body with computed z-index 600 and 550, not nested, so they are stacking siblings and the panel paints over the tooltip. No PopoverPortalContainerProvider exists anywhere in the Find tree. Genuine regression, not a pre-existing pattern: dropdown-menu.tsx:128 carried tw:z-50 with no override in the old JSX. Occlusion is not marginal - --popover has no alpha (index.css:129), Tooltip defaults to side="top", and both Info icons sit inside the w-72 panel, so each tooltip renders upward entirely within the panel's box; the third tooltip on the trigger button is correctly unaffected. Held at high rather than blocker because the filters still work and the copy is supplementary, but it silently deletes two Info affordances in an accessibility PR and the fix is one line._

Similar fix as #6.

_Other findings in this file: #5, #6, #15, #19_

---

**#5** - medium · confirmed

PopoverContent renders role="dialog" with no accessible name, where the old DropdownMenuContent was labelled by the trigger.

A jsdom probe at both revisions shows the container regressing from role="menu" with aria-labelledby resolving to the trigger's aria-label="Toggle filters" (Testing Library's role dump: menu: Name "Toggle filters") to role="dialog" with aria-label and aria-labelledby both null and computeAccessibleName() returning "". aria-dialog-name is an axe/WCAG 4.1.2 rule that role=menu was not subject to. The in-repo fix pattern is sync-status-button.component.tsx:665-676, which passes aria-label to PopoverContent with a comment saying exactly this. One-prop fix using the already-localized localizedStrings.toggleFilters.

**Suggested fix** - one prop, using the string that is already loaded and used for the trigger's own `aria-label` on line 91:

```tsx
<PopoverContent
  align="end"
  aria-label={localizedStrings.toggleFilters}
  className="tw:max-h-(--radix-popover-content-available-height) ..."
>
```

Precedent, including a comment explaining exactly this: `src/renderer/components/sync-status-button.component.tsx:664-676`.

_Verification: Verified by rendering, not by reading docs: a jsdom probe opened the panel and read the container at both revisions - at HEAD role="dialog" with aria-label=null, aria-labelledby=null, computeAccessibleName() === ""; at MERGE_BASE role="menu" with aria-labelledby="radix-_r_5_" resolving to the trigger button carrying aria-label="Toggle filters". So the accessible name genuinely regressed from "Toggle filters" to empty, on the added line find-filters.component.tsx:111. The sync-status-button precedent is real and does exactly what the finding says. Held at medium rather than higher because the PR moves focus to the first radio on open, so the user still hears a labelled control plus the fieldset legends, and because 31 of the repo's 32 PopoverContent sites are likewise unnamed - a one-line regression joining an existing repo-wide gap, not a novel break._

_Other findings in this file: #1, #6, #15, #19_

extensions/src/platform-scripture/src/find/find-filters.component.tsx
```diff
@@ -88,12 +93,25 @@ export function FindFilters({
...
+        align="end"
+        className="tw:max-h-(--radix-popover-content-available-height) tw:w-72 tw:overflow-x-hidden tw:overflow-y-auto tw:p-3"
```
lyonsil, line 113: **#6** - medium · confirmed

The PR body's "nothing else about the panel's appearance changes" is inaccurate: the panel loses the app-wide menu glass, gains an animation, and grows about 40px taller.

A token-level diff of the two base class strings confirms what is lost on the swap: tw:bg-popover/70, tw:relative and the tw:before:* backdrop-blur-2xl / backdrop-saturate-150 glass (applied identically in menubar.tsx, context-menu.tsx and select.tsx - popover.tsx is the only overlay without it), plus tw:animate-none! and tw:data-[state=closed]:overflow-hidden. Both files carry the same data-open:animate-in fade-in-0 zoom-in-95 set, so removing the !important animate-none genuinely turns the zoom/fade on. The other lost tokens are re-supplied at the call site. The diff also surfaces an unclaimed delta: PopoverContent adds tw:flex tw:flex-col tw:gap-2.5, so the four inter-group gaps go 12px to 22px - about 40px of added panel height.

**Suggested fix** - the height half is a one-class fix. `PopoverContent`'s base includes `tw:flex tw:flex-col tw:gap-2.5`, which `DropdownMenuContent` did not have (it was block layout), so the 10px gap now stacks on top of each fieldset's own `tw:mb-3`. Adding `tw:gap-0` to the className restores the previous 12px spacing exactly, since the fieldsets still carry their own margins and the last one has none:

```tsx
className="tw:max-h-(--radix-popover-content-available-height) tw:w-72 tw:gap-0 tw:overflow-x-hidden tw:overflow-y-auto tw:p-3"
```

The glass treatment (`tw:bg-popover/70` plus the `tw:before:backdrop-blur-2xl` / `backdrop-saturate-150` pair) and `tw:animate-none!` are a judgement call rather than a defect - the panel now matches the two plain popovers beside it in the same toolbar rather than the menus elsewhere in the app. Either match the menus by carrying those classes over, or keep it as is; but either way the PR description's "All markup, styling, grouping, and tooltips are unchanged" and "nothing else about the panel's appearance changes" are not accurate and are worth correcting so a reviewer knows what to look at.

_Verification: A comm diff of the two base class strings confirms every class the finding names, and surfaces the extra flex/gap delta the finding missed. The PR body's "All markup, styling, grouping, and tooltips are unchanged", "nothing else about the panel's appearance changes", and "no visual change" are therefore all inaccurate: the panel goes opaque, animates, and grows taller. Two mitigations keep this at medium: the panel's nearest siblings in the same toolbar are already plain popovers (find.component.tsx:1115, replace-preview-options.component.tsx:128), so it now matches them rather than standing alone; and the scrollbar-flash leg is the weakest - the exit is a transform scale() zoom-out that does not change layout, so the missing closed-state overflow-hidden is unlikely to actually produce a bar._

Similar fix as #1.

_Other findings in this file: #1, #5, #15, #19_

---

**#19** - low · downgraded on review

The new PopoverContent copies the sibling popover's height-cap recipe but omits its collisionPadding={8}.

The recent-searches/scope popover in the same web view (find.component.tsx:1116-1118) uses the identical max-h/overflow recipe plus collisionPadding={8}. Radix's default is 0.

**Suggested fix** - add `collisionPadding={8}` to match the sibling popover in the same file:

```tsx
<PopoverContent
  align="end"
  collisionPadding={8}
  aria-label={localizedStrings.toggleFilters}
  className="..."
>
```

Purely a consistency and polish change - it gives the panel an 8px gutter from the viewport edge instead of sitting flush against it, matching `find.component.tsx:1116-1118`. Not a correctness issue: `--radix-popover-content-available-height` already accounts for collision padding, so nothing is clipped either way.

_Why this is downgraded: The omission is real, but the "sibling's comment says it is needed to stop clipping" premise is wrong: that comment at find.component.tsx:1110-1114 attributes clipping-prevention to being "Height-capped and scrollable" and never mentions collisionPadding. Radix confirms the right reading - collisionPadding defaults to 0 and feeds detectOverflow's padding into size(), which sets --radix-popper-available-height, so the CSS variable already nets out collision padding and the panel is capped flush to the edge, not clipped past it. The pre-PR container had the same collision behavior: dropdown-menu.tsx:128 supplies its own available-height cap with no collisionPadding either, so this PR introduces no regression, only the absence of an 8px gutter. Positionally the two are not alike - the filters trigger sits in the top search row with the whole panel below it. A cosmetic consistency nit, not a defect._

_Other findings in this file: #1, #5, #6, #15_

extensions/src/platform-scripture/src/find/find.component.test.tsx
```diff
@@ -882,3 +882,51 @@ describe('Find — whitespace and diacritic tolerance toggles', () => {
     },
   );
 });
+
+// These exist because the filter controls are plain form controls (radio groups, checkboxes), not
+// menu items. Housing them in a Radix menu makes them unreachable by keyboard: menu content calls
+// preventDefault on Tab, and its arrow handling only walks items registered in its roving-focus
+// collection — which plain form controls never join. A regression back to a menu container would
+// leave the panel mouse-only, and nothing else in this suite would notice.
+describe('Find — filters panel keyboard accessibility', () => {
```
lyonsil, line 891: **#7** - medium · confirmed

The subject of the PR's last two commits - layout containment - has no assertion anywhere; only the first commit's keyboard behavior is tested.

git log 3529e673..HEAD shows the last two commits touch only find.component.tsx, landing tw:overflow-hidden on the root and min-h-48 to min-h-0 on the results scroller, with zero test changes. The only test diff in the PR is the three it() blocks added by the first commit, which assert toHaveFocus() and nothing else. A grep for overflow/scrollTop/scrollHeight/clientHeight/boundingBox/getBoundingClientRect/scrollWidth/min-h/scrollbar/documentElement across the test file, find.stories.tsx and both e2e specs returns no assertion about document scroll, root containment, popover position or results height. The PR body itself says the wobble was verified manually.

**Suggested fix** - the two harnesses that can measure layout already exist, and the measurement you did by hand is the assertion. `e2e-tests/tests/isolated/find/find.spec.ts:407` already has an `openFiltersPanel` helper, so this is a short addition:

```ts
await openFiltersPanel(page);
const before = await filtersTrigger.boundingBox();
await page.getByRole('radio', { name: 'Verse text only' }).click();
await expect.poll(async () => (await filtersTrigger.boundingBox())?.x).toBe(before?.x);
```

`find.stories.tsx:738-762` is the other option - it already carries a layout-measuring play function as an in-file precedent. Neither gates CI today (the workflow runs only `--project=smoke`, and Chromatic fires only on `storybook-review`-labelled PRs to `ai/main`), so this is a guard for a human to run rather than a blocker - but it is the difference between the fix being verified once and staying verified.

_Verification: Every factual leg checks out, including the sibling refuter's claim about the vitest projects, which I re-verified rather than took on trust: npx vitest list under extensions/ returns 0 lines matching "stories" (environment node, include src/**/*.test.ts(x) only), and the sole storybook browser project is scoped to lib/platform-bible-react. Yet the gap is still closable, because find.stories.tsx:738-762 already carries a layout-measuring play function as an in-file precedent for exactly this class of regression, and openFiltersPanel genuinely exists at find.spec.ts:407 and replace.spec.ts:150 beside filter-clicking tests - so a boundingBox() stability check is a few lines on a measurement the author already performed (519 to 505). Held at medium rather than higher because neither harness gates CI: the workflow runs only --project=smoke, and Chromatic fires only on storybook-review-labeled PRs to ai/main while this PR targets main, so the guard would be advisory._

_Other findings in this file: #28_

extensions/src/platform-scripture/src/find/find.component.test.tsx
```diff
@@ -882,3 +882,51 @@ describe('Find — whitespace and diacritic tolerance toggles', () => {
...
+  /** Opens the filters panel, which is where all the filter controls live */
+  async function openFilters(user: ReturnType<typeof setupUser>) {
```
lyonsil, line 893: **#28** - low · downgraded on review

The new describe re-declares openFilters byte-identically to the copy in the immediately preceding describe.

Both read: async function openFilters(user) { await user.click(screen.getByRole('button', { name: 'Toggle filters' })); }. The file already keeps setupUser and openProjectSelector at module scope. The three new tests also repeat the same three-line setup verbatim.

**Suggested fix** - delete the new copy at line 893 and let the new `describe` use the one already declared at line 846, or move that one to module scope next to `setupUser` and `openProjectSelector` (line 280) and have both `describe` blocks share it. Either is fine; the second is tidier since the two blocks sit next to each other and test the same panel.

_Why this is downgraded: Byte-identical duplication confirmed at find.component.test.tsx:846 and :893. But the file's prevailing idiom is describe-local helpers, not shared module-scope ones: buildReplaceProps (:509), buildPerResultProps (:592), ClearingFindHarness (:676), buildBooksScopeProps (:732) and buildEmptiedSelectionProps (:786) are each declared fresh per-describe despite being conceptually similar "build variant props" helpers, and none were hoisted. Only setupUser and openProjectSelector (:280, :284) are module-scope, and those are used across most of the file's nine describe blocks, unlike openFilters which is used only by these two adjacent, narrowly-scoped describes. A real two-line, two-instance duplication local to one file section, not a violation of an established hoisting pattern._

_Other findings in this file: #7_

extensions/src/platform-scripture/src/find/find.component.tsx
```diff
@@ -834,7 +834,13 @@ export function Find({
...
 
   return (
-    <div className="pr-twp tw:mx-auto tw:flex tw:flex-col tw:gap-4 tw:p-4 tw:min-w-[10rem] tw:max-h-screen">
+    // `overflow-hidden` makes `max-h-screen` hold, so the document never scrolls. It otherwise does,
+    // and since a search re-runs whenever a filter changes, the results area crosses the viewport
+    // boundary often enough that the document scrollbar toggles. Each toggle narrows the viewport by
+    // the scrollbar's width, shifting this right-aligned toolbar sideways — which anything anchored
+    // to it, such as the filters popover, visibly follows. Nothing is clipped: the results list below
+    // shrinks to the space left over and scrolls within it.
+    <div className="pr-twp tw:mx-auto tw:flex tw:flex-col tw:gap-4 tw:overflow-hidden tw:p-4 tw:min-w-[10rem] tw:max-h-screen">
```
lyonsil, line 843: **#2** - high · confirmed

tw:overflow-hidden on the root clips the un-shrinkable status bar and skeleton with no scrollbar anywhere, contradicting the diff's own "Nothing is clipped".

Only the results div got min-h-0. The loading skeleton (find.component.tsx:1179-1194, five Cards, roughly 402px intrinsic) and the status bar carrying Progress plus Cancel search (find.component.tsx:1308-1310) are plain siblings with min-height:auto and visible overflow, so they cannot shrink and there is no scroller left anywhere. The iframe is flex-grow:1 in .web-view-parent, so 100vh is exactly the pane. Before the change the document scrolled and the tail was reachable.

**Suggested fix** - one word on line 843: `tw:overflow-hidden` -> `tw:overflow-y-auto`.

I measured all three candidates in headless Chromium against a replica of this flex structure at a 400px pane with the skeleton up:

| root overflow | hidden overflow | wheel scrolls | Cancel search reachable |
| --- | --- | --- | --- |
| `overflow:hidden` (current) | 255px | no | no |
| `overflow-y:auto` | 255px | yes, 255px | yes |
| `overflow:clip` | 239px | no | no |

`overflow-y-auto` is the only one that works, and it costs nothing in normal panes: because the results list is `flex-1 min-h-0`, it absorbs all the shrinking first, so the root can only overflow once the results list has already been squeezed to 0px - i.e. exactly the case that is currently unreachable. Above roughly a 600px pane there is no root overflow at all and no scrollbar. It also removes the need for the separate `overflow-clip` change, since once the container really scrolls, a programmatic `focus()` or `scrollIntoView()` that shifts it can simply be scrolled back.

One caveat worth checking in the app: this was measured with overlay scrollbars (no layout width taken). On Windows, classic scrollbars take about 15px, so if you see the panel twitch when the scrollbar appears, add `tw:[scrollbar-gutter:stable]` alongside it to reserve the width permanently.

This also resolves items 8, 29 and 34, which are all downstream of the results list collapsing with nowhere to scroll.

_Verification: Measured in headless Chromium with a border-box replica of the real class stack. At a 400px pane with the skeleton up: before, the document scrolls (scrollHeight 833) and Cancel search is reachable; after, scrollHeight === clientHeight === 400, the results area collapses to 0px, and Cancel sits at y=626 with no way to scroll to it - 188px unreachable. The tail only becomes reachable again above roughly 667px of pane height, so a normal laptop panel is in range, and the skeleton is up precisely when Cancel search is the thing you need. One qualifier on the supporting evidence: the "51px unreachable" figure from commit 7c564d1959b was measured with min-h-48 still on the results list and does not transfer - min-h-0 genuinely absorbs that case. The un-shrinkable siblings do not, which makes the new comment's "Nothing is clipped: the results list below shrinks to the space left over" false for the header/skeleton/status-bar remainder._

_Other findings in this file: #30, #33_

---

**#30** - medium -> low · downgraded on review

overflow-hidden clips both axes, and it silently disarms the NarrowPanelClipsScopeSummary story guard.

The story's assertion panel.scrollWidth <= panel.clientWidth goes vacuous: before, panel.scrollWidth 402 vs clientWidth 260 (fails, correctly); after, 260 vs 260 (passes), because a clipping child does not propagate scrollable overflow to its overflow-x-auto ancestor. The layout fix the guard protects (find.component.tsx:1093-1107, shrink/min-w-0/truncate) is still present and unchanged - clipping masks a future regression in it rather than removing it.

**Suggested fix** - point the assertion at the element that now does the clipping. The story measures the decorator's wrapper (`div[data-testid="find-panel"]`, `find.stories.tsx:744`), and a clipping child does not propagate scrollable overflow to its ancestor, so the wrapper's `scrollWidth` and `clientWidth` are now always equal and the guard can never fail:

```ts
const root = panel.firstElementChild as HTMLElement;
await expect(root.scrollWidth).toBeLessThanOrEqual(root.clientWidth);
```

The root's own `scrollWidth` still reports the real 402px overflow, so measuring there keeps the guard live. Note `tw:overflow-y-hidden` does NOT fix this - I checked, and it makes computed `overflow-x` become `auto`, leaving the wrapper assertion equally vacuous. The guard does not run in CI either way (the extensions vitest project is node-env and does not execute play functions), so this only matters to whoever opens the story.

_Why this is downgraded: The mechanics check out - reproduced in headless Chromium, and the guard does go vacuous - but two parts of the claim do not survive. panel is not the changed root: it is the story decorator's wrapper div[data-testid="find-panel"].tw:overflow-x-auto at find.stories.tsx:744, with the overflow-hidden root as its child (the root's own scrollWidth is still 402, so the finding's caveat is true but does not rescue the ancestor assertion). And the prescribed remedy is disproved: overflow-y:hidden makes computed overflow-x become auto, so the root is still an x-scroll container, panel.scrollWidth is still 260, the guard stays vacuous, and it reintroduces the horizontal scrollbar the PR's own sibling comment deliberately avoids. The guard also never ran in CI - extensions/vitest.config.ts is node-env and includes only *.test.ts(x), and the only storybook browser project is scoped to lib/platform-bible-react - so the play function is a manual Interactions-panel affordance, as its own docstring says._

_Other findings in this file: #2, #33_

extensions/src/platform-scripture/src/find/find.component.tsx
```diff
@@ -1195,7 +1201,11 @@ export function Find({
       {/* eslint-disable-next-line jsx-a11y/no-static-element-interactions */}
       <div
         ref={resultsContainerRef}
-        className="tw:min-h-48 tw:flex-1 tw:space-y-2 tw:overflow-y-auto tw:pe-2"
+        // `min-h-0` rather than a fixed floor so this list can shrink below its content when the
+        // panel is short. A floor here would refuse to give way and push the root past its height
+        // cap, which is what previously forced the document to scroll. The empty-state placeholder
```
lyonsil, line 1206: **#33** - low · downgraded on review

Two comments added by this PR carry backward-facing narration clauses the repo rule names.

find.component.tsx:1206 says "which is what previously forced the document to scroll", and find.component.test.tsx:922 says "and the part that was broken". Both are single clauses inside otherwise-legitimate rationale and can be stripped without losing it; the surrounding text (the min-h-0-versus-fixed-floor tradeoff, and jsdom's inability to reproduce Radix's selection behavior) should stay.

**Suggested fix** - trim the two narration clauses; the rest of each comment is good rationale and should stay.

- `find.component.tsx:1206`: drop ", which is what previously forced the document to scroll". The sentence before it already explains why a floor here would be wrong.
- `find.component.test.tsx:922`: drop ", and the part that was broken". "Navigation is the part this component controls" already carries the point.

Repo rule: `.claude/rules/code-quality/forward-facing-comments.md` - the test is whether a comment still helps someone who never saw this PR. The two other comments flagged in review (`find.component.tsx:837` and the `describe` comment at `find.component.test.tsx:889`) are fine as written; a counterfactual explaining why a class is needed, and a note on what regression a test guards, are both forward-facing.

_Why this is downgraded: Of the four sites the generator cited, only two genuinely contain backward-facing narration. find.component.tsx:837 ("It otherwise does...") is not narration at all - it is a counterfactual explaining why overflow-hidden is needed, squarely the rule's "why the code takes a surprising or non-obvious approach" carve-out, and references nothing PR-specific. find.component.test.tsx:889-890 ("A regression back to a menu container would leave the panel mouse-only, and nothing else in this suite would notice") is a describe-block comment stating a hypothetical future regression the suite guards against, phrased prospectively - the rule itself flags this pattern as likely-legitimate test rationale. Only find.component.tsx:1206 and find.component.test.tsx:922 qualify, and each needs a phrase trimmed rather than a comment moved to the commit message - smaller in both count and scope than claimed._

Similar fix as #15.

_Other findings in this file: #2, #30_

Sebastian-ubs:
Tested on https://github.com/paranext/paranext-core/actions/runs/34518239745
- "toggle filters" dropdown works
- **"recent searches" is still not keyboard accessible** &rarr; this needs to be fixed
- navigating through results without loosing focus works
- **nice to have wish**: ctrl+f inside of the find window should focus the input field

extensions/src/platform-scripture/src/find/find-filters.component.tsx
timothy-mccormack: Deleted. That paragraph was left behind when the class went back to `overflow-y-auto`; the comment now only explains the height cap and `gap-0`.

extensions/src/platform-scripture/src/find/find-filters.component.tsx
timothy-mccormack: #1: both in-panel tooltips now render above the panel, with a test that fails at 550 vs 600. Since you reviewed, #2750 merged and raised `Z_INDEX_TOOLTIP` to 675 — above the popover tier — so I merged `main` and dropped the local override entirely rather than leaving the TODO. The stacking tests assert observed z-index rather than the constant, so they carried over unchanged and now verify the shared scale. I left the trigger's own tooltip alone.

#5: `aria-label={localizedStrings.toggleFilters}` on `PopoverContent`, asserted by a test.

extensions/src/platform-scripture/src/find/find-filters.component.tsx
timothy-mccormack: #6: `gap-0` restores the 12px spacing. I did *not* carry over the glass or `animate-none!`: PT-4535 is making menus and popovers opaque, so the panel now matches the plain popovers beside it in the same toolbar. Worth noting the glass would also have misbehaved here — the blur layer is absolutely positioned inside the scroll container, so it scrolls away from the content it's meant to blur. The PR description now says the appearance changed.

#19: `collisionPadding={8}` added, matching the sibling popover.

extensions/src/platform-scripture/src/find/find.component.test.tsx
timothy-mccormack: Added as an e2e test. It samples the filters button's x every animation frame from the click until a second after the new count appears, because the shift settles back once the search finishes and a single before/after reading misses it. It earned its keep immediately: it caught the panel still shifting 14px while the loading skeleton was up (the root grew its own scrollbar), which is fixed here by letting the skeleton give way.

One existing test changed too, and it's worth flagging rather than leaving in the diff. `Search History › should add the term in the box to history when options change` filled the search box with the panel open and kept using it. That only worked because the old menu was modal: a modal popover suppresses Radix's focus-outside dismissal, a non-modal one doesn't, and `locator.fill()` focuses without a pointerdown. So the panel now closes there. That dismissal is the behavior we want — it's the same rule that lets Shift+Tab and Escape return focus to the trigger, and suppressing it via `onFocusOutside` breaks those tests instead. The test reopens the panel for its second option change; what it asserts is unchanged. A unit test now pins the dismissal down directly. Caught late, because the e2e pair was last run while the panel was still modal.

extensions/src/platform-scripture/src/find/find.component.test.tsx
timothy-mccormack: Hoisted `openFilters` to module scope next to `openProjectSelector`; both describes use it. That also removed the stale "dropdown" JSDoc from #25.

extensions/src/platform-scripture/src/find/find.component.tsx
timothy-mccormack: #2: `overflow-y-auto` as you suggested. Your measurement matched mine in the app: trigger x 519 → 505 in lockstep with the viewport's 567 → 553. One correction to my first attempt — I had removed the results list's floor entirely, and a self-review found that squeezes the list to 0px in a short panel, leaving results unreachable. It now keeps a floor of about one and a half cards, dropped while the loading skeleton is up, where the list is empty and reserving the space would bring the shift back.

#30: the guard measures the Find root (`firstElementChild`) instead of the wrapper, so it can still fail.

extensions/src/platform-scripture/src/find/find.component.tsx
timothy-mccormack: Both clauses trimmed.

timothy-mccormack:
Replies to the findings that weren't on their own threads (#3, #22, #23, #25) — the rest are answered in-thread.

**#3:** added both deps, with a comment saying why they're there when the body reads none of them. Honest gap: nothing fails if someone removes them again. An e2e toggle test would depend on search data that varies by project, so I left it out — say the word and I'll add one.

**#22:** I added the group names, then removed them in a second review round. Each `aria-label` repeated its own `<legend>`, so the group was announced twice in a row. The fieldsets still name the groups and there is now a test asserting that. Happy to put the labels back if you'd rather have the spec-clean version.

**#23:** I took a different route. The icons describe their controls through `aria-describedby` rather than becoming tab stops. A tab stop there sits inside the radio group, so Tab never reaches it when "Verse text only" is the saved choice, Shift+Tab bounces between it and the radio, the tooltip text was announced twice, and it closed again whenever reaching it scrolled the panel. Screen readers now get the text on the control itself; the icons stay hover tooltips.

**#25:** fixed by the shared helper in #28.

One more from this round, since it was mine: Shift+Tab from the first control left the panel and dismissed it without returning focus to the trigger. Closing now returns focus to the button, with a test that fails without it. I tried a focus trap first and backed it out — your dismissal guidelines rule it out for non-modal surfaces, and it also aria-hides the rest of the panel, which silences the results area's announcements while a filter re-runs.

timothy-mccormack:
Thanks for testing this, and for the precise notes.

**Recent searches** — fixed, though not by this PR. You tested the build from `ce6b838`, which predates #2750 (PT-4345) landing; that PR reworked Recent searches into a proper keyboard-navigable menu. It has since merged, and this branch now includes it, so the current build should behave. I had a fix here too and dropped it so the two wouldn't collide — #2750's version is the approved design and covers the reference picker as well. Worth a retest when CI finishes if you have a moment.

**Toggle filters and result navigation** — glad those hold up. Since you tested, the panel also got Shift+Tab and Esc returning focus to the filters button, and the ⓘ explanations are now read out on the control they explain rather than being mouse-only tooltips.

**Ctrl+F inside Find** — agreed that'd be handy, but it's a nice-to-have rather than part of this fix, so I've kept it out of scope. Worth its own ticket if you'd like it prioritized.

extensions/src/platform-scripture/src/find/find.component.test.tsx
timothy-mccormack: Correction to my note above about the changed e2e test — I had the mechanism wrong, and being wrong about it hid a real bug.

I wrote that the focus-outside dismissal is "the same rule that lets Shift+Tab and Escape return focus to the trigger." It isn't. Escape goes through `DismissableLayer`'s `useEscapeKeydown`, independent of `useFocusOutside`, and the focus return comes from this component's own `onCloseAutoFocus`. Two separate mechanisms. Conflating them is why adapting the test looked like the only option on the table.

A self-review round found what that was covering. `onCloseAutoFocus` was gated on "not a pointerdown", so it treated every programmatic focus as a keyboard exit. Invoking Find while the filters panel is open calls `focusSearchInput`; the caret reaches the search box, that focus change is itself what dismisses the panel, and a macrotask later the panel pulled the caret back onto the Toggle filters button. Typing went nowhere and Space reopened the panel — `useFocusSearchOnInvoke` was defeated whenever the panel happened to be open.

It's now gated on a Tab or Escape keydown originating inside the panel, which is the case the return exists for. Every other dismissal leaves focus where it went, which is also what stock Radix does. Clearing the flag when the panel opens closes a second hole: Radix can start a close and then cancel it, stranding a flag that sent the next Escape's focus nowhere at all.

My unit test was part of the problem rather than a guard against it. It asserted only that the panel closed, and its `act` boundary resolved one macrotask before Radix's `FocusScope` dispatches its unmount autofocus — so it inspected the one instant where the caret still looked correct. Both branches now have tests that drain that timer before asserting.

On this thread's actual subject: `openFiltersPanel` now converges on "open" using the `aria-expanded` + `toPass` pattern `resetFindPanel` already uses, so no caller depends on knowing whether a dismissal has landed yet. The stacking test also asserts both z-indexes are declared rather than comparing `Number(...)` directly — `Number('')` is `0`, so the bare greater-than couldn't fail.

`ddc6061`. 356 unit tests and both e2e tests pass.

lyonsil (commented):
Re-reviewed the updates. **All 14 findings from the previous round are genuinely addressed** - I verified each against the branch rather than taking the replies at face value, and resolved the seven threads. Two calls I want to note approvingly: declining #22 because each `aria-label` duplicated its own `<legend>` is right, and the `aria-describedby` route for #23 is better than the tab stop I suggested.

One blocker from the newest round, and one nit, are inline below. The blocker is in `ddc6061` itself.

Also: the key table under "The fix" still says focus returns to the trigger "unless it was dismissed by a pointer", which is the pre-`ddc6061` gate. The bullet further down describes the current one correctly.

For the record on scope: the new e2e test does not gate CI - `.github/workflows/test.yml` runs only `--project=smoke`, so `tests/isolated/` never runs there. That matches what I said when raising #7; flagging it only so nobody reads the added test as a merge gate.

extensions/src/platform-scripture/src/find/find-filters.component.tsx
```diff
@@ -74,26 +88,84 @@ export function FindFilters({
...
+          explicitly. */}
+      <PopoverContent
+        align="end"
+        collisionPadding={8}
+        // Tab and Escape are the keyboard's two ways out of the panel, and both leave the user with
+        // nowhere to stand once it unmounts, so both hand focus to the trigger. Capture phase records
+        // the key even if a control inside stops the event.
+        onKeyDownCapture={(event) => {
+          if (event.key === 'Tab' || event.key === 'Escape') dismissedByKeyboardRef.current = true;
+        }}
```
lyonsil: **Blocker** - `ddc6061` reintroduces the bug it set out to fix, through a path its two new tests do not exercise. Verified in the running app, not only in jsdom.

`dismissedByKeyboardRef` is armed by **any** Tab keydown inside the panel, and cleared only when the panel *opens* (line 116). But Tab is also how you move *between* the five groups - the thing this PR exists to enable. So one in-panel Tab leaves the flag stuck true for the rest of that visit, and the next dismissal of any kind steals the caret to the trigger.

### Reproduction

Built this branch at `ddc6061` and drove it over CDP. Pure keystrokes, no programmatic focus before the exit:

```
1.  Click "Toggle filters"    focus = button#searchTextType-all
2.1 Tab ->                    focus = button#wordRestriction-none          panel open = true
2.2 Tab ->                    focus = button#matchCase                     panel open = true
2.3 Tab ->                    focus = button#ignoreWhitespaceDifferences   panel open = true
2.4 Tab ->                    focus = button#ignoreDiacritics              panel open = true
3.  Click the search box      focus = button aria-label="Toggle filters"   panel open = false
4.  Type "abc"                #search-term value = "the"   <-- characters swallowed
```

Step 4 is the user-visible harm: the caret is on a button, so typing does nothing and Space reopens the panel - the symptom your own commit message describes.

Controls from the same session:

| Scenario | Focus after the exit |
| --- | --- |
| No Tab inside, then **click** the search box | `input#search-term` |
| One+ Tab inside, then **click** the search box | `button aria-label="Toggle filters"` |
| No Tab inside, then **`focus()`** the search box | `input#search-term` |
| One+ Tab inside, then **`focus()`** the search box | `button aria-label="Toggle filters"` |

The `focus()` rows are `focusSearchInput` itself (`searchInputRef.current?.focus()`), so Repro B is the invoke case: open the panel, Tab once, then Ctrl+F from a scripture tab or the editor's **Find** menu item.

Both new tests pass because neither presses Tab first - they open the panel and immediately click or focus the search box.

### The comment's premise does not hold

Lines 151-152 say "Tab and Escape are the keyboard's two ways out of the panel." Forward Tab is never a way out. Radix's `FocusScope loop` cycles it inside the panel - traced 10 presses in the app:

```
searchTextType-all -> wordRestriction-none -> matchCase -> ignoreWhitespaceDifferences
  -> ignoreDiacritics -> allowRegex -> searchTextType-all -> ...   (panel open throughout)
```

Shift+Tab from the first control is the only Tab that exits: one press, panel closed, focus on the trigger. So **every forward Tab arms a focus-return that can only ever fire on a dismissal that is not a Tab exit.**

### Suggested fix

Clear the flag when focus lands back inside the panel - i.e. when the Tab did not leave:

```tsx
onKeyDownCapture={(event) => {
  if (event.key === 'Tab' || event.key === 'Escape') dismissedByKeyboardRef.current = true;
}}
// A Tab that lands back inside the panel did not leave it. Escape moves no focus, so its arming
// survives; a Tab that exits fires no focusin here, so that one survives too.
onFocusCapture={() => {
  dismissedByKeyboardRef.current = false;
}}
```

The regression test is the existing `leaves focus in the search box when the box outside the panel takes focus` with one line added after `openFilters`:

```tsx
await user.tab(); // Move between groups inside the panel; it stays open.
```

_Verification: Reproduced twice over - in jsdom (both probes fail at `ddc6061`, `document.activeElement` is the trigger) and in a real headless-Chromium build of this branch, where typing after the theft leaves `#search-term` at its old value. Then rebuilt the app with the fix above and re-ran everything: Escape returns to the trigger with and without prior Tabs; Shift+Tab out returns to the trigger; clicking and `focus()`-ing the search box after 4 in-panel Tabs both leave the caret in the box; typing lands (`"theabc"`). All 73 unit tests in `find.component.test.tsx` stay green, including both focus-return tests. One check I wrote, "Tab past the last control", is not a real case - forward Tab loops, so the panel never closes and there is no dismissal; the fix cannot affect it either, since it only clears the flag when focus lands back inside the panel, which is exactly when no dismissal happens._

e2e-tests/tests/isolated/find/find.spec.ts
```diff
@@ -960,6 +977,66 @@ test.describe('Search Filters', () => {
...
+    let positions: number[];
+    try {
+      // Whole word moves this term's count (52 -> 51), so the counter proves the search re-ran.
+      await frame.locator('#wordRestriction-wholeWord').click();
+      await waitForCounterToChangeFrom(frame, counterBefore);
+      positions = await positionsDuringReRun;
+    } finally {
+      // Closed however this ends. A panel left open by a failure here is state the afterEach reset
+      // has to undo before it can attribute the failure to this test.
+      await frame.locator('#wordRestriction-wholeWord').press('Escape');
```
lyonsil: **Nit** - this `finally` can mask the failure it is meant to clean up after.

There is no `actionTimeout` configured (`e2e-tests/playwright.config.ts` sets only `timeout: 120_000` and `expect.timeout: 10_000`), so if the `try` block failed because `#wordRestriction-wholeWord` was not there, this `press` waits on the same missing element until the 120 s per-test timeout - turning a clear assertion failure into an opaque one.

`resetFindPanel` already converges on either state via `aria-expanded`, so the cleanup is not buying much. Either drop it, or make it non-fatal:

```ts
} finally {
  await frame.locator('#wordRestriction-wholeWord').press('Escape', { timeout: 2_000 }).catch(() => {});
}
```

extensions/src/platform-scripture/src/find/find-filters.component.tsx
timothy-mccormack: Confirmed and fixed in `03bc2a3`. Your reproduction is exactly right, and the two tests failed the moment I added the Tab you described — `document.activeElement` on the trigger in both.

The mechanism is the part I should have caught: I'd already established that Radix's `FocusScope loop` keeps forward Tab inside the panel, and then wrote a gate that armed on any Tab anyway. Since forward Tab never exits, one press between groups armed a return that could only fire on some later, unrelated dismissal — so using the panel the way this ticket exists to enable was enough to reinstate the bug the commit removed. The comment claiming "Tab and Escape are the keyboard's two ways out" was wrong in the same way, and is corrected.

Took your suggested fix. Which Tab is the exit isn't knowable at keydown, so the focus that follows answers it: a Tab landing back inside clears the arming, the one Shift+Tab that leaves fires no focus event here and keeps it, and Escape moves no focus so it keeps its arming too.

Three tests rather than one. The click and `focus()` paths after an in-panel Tab are the two that regressed, and Escape after an in-panel Tab guards the opposite error — clearing an arming that was genuinely owed. 359 unit tests pass, both focus-return tests included, and both e2e tests pass against a rebuilt app.

Also took the point about CI scope: understood that `tests/isolated/` doesn't gate merge, and I'm not treating the e2e test as one.

e2e-tests/tests/isolated/find/find.spec.ts
timothy-mccormack: Applied in `03bc2a3`, with the bound and the `.catch`. Your reading is right — when the failure above is the control going missing, an unbounded `press` waits on that same missing control for the full 120 s and buries the actual assertion error behind a timeout. Since `resetFindPanel` converges from either state anyway, this cleanup is only tidying, so it shouldn't be able to outrank the failure it's cleaning up after.

e2e-tests/tests/isolated/find/find.spec.ts
```diff
@@ -960,6 +977,72 @@ test.describe('Search Filters', () => {
...
+    // would happen. The shift can settle back once the search finishes, so a single before/after
+    // reading would miss it: the position is sampled on every frame from the click until a second
+    // after the new count appears.
+    const frame = await openFindPanel(mainPage);
+    await fillSearchAndWaitForResults(frame, FILTER_SENSITIVE_TERM);
+    const counterBefore = await frame.locator('.tw\\:tabular-nums').textContent();
+
+    await openFiltersPanel(frame);
+    const filtersButton = frame.getByRole('button', { name: /toggle filters/i });
+    const positionsDuringReRun = filtersButton.evaluate(
```
lyonsil: **#1** - low · checked and confirmed

A failure inside this test can be reported alongside a second, unrelated error, and its in-page sampling loop keeps running into the tests that follow.

**What happens:** `positionsDuringReRun` is created at `find.spec.ts:996` but awaited only at `:1031`, inside the `try`. If the click at `:1029` or `waitForCounterToChangeFrom` at `:1030` throws, nothing ever awaits it. The `requestAnimationFrame` loop at `:1017` keeps sampling every frame until `timeoutMs` (35 s, `:1023`), and the Electron app is worker-scoped (`e2e-tests/fixtures/find.fixture.ts`), so that runs on into the next tests. If the worker tears the app down first, the pending `evaluate` rejects with nothing attached to it.

**Why it matters:** the unhandled rejection surfaces as a second error on a run that already has a real assertion failure, and it is not attributable to the test that created it.

**Fix:** attach `.catch((): number[] => [])` to the `evaluate` call where it is created at `:996`, so the promise can never reject unhandled. Do not also await it in the `finally` at `:1032` — on the failure path `countChangedAt` is never set, so awaiting there would add the full 35 s to an already-failing test.

_How this was checked: Read against the branch: the promise is constructed at `:996`, outside the `try` that opens at `:1027`, and the only `await` on it is at `:1031` inside that block. The `finally` at `:1032` presses Escape and does not touch it. The sampling loop's own exit conditions are at `:1012`-`:1016`, so with `countChangedAt` unset it runs the full `timeoutMs`._

extensions/src/platform-scripture/src/find/find-filters.component.tsx
```diff
@@ -74,26 +88,93 @@ export function FindFilters({
...
+        // Every other dismissal leaves focus wherever it went. That matters most when the app moved
+        // it deliberately: invoking Find calls `focusSearchInput`, and the resulting focus change is
+        // itself what dismisses the panel, so returning focus to the trigger here would take the
+        // caret straight back out of the search box the invoke exists to reach.
+        onCloseAutoFocus={(event) => {
+          if (!dismissedByKeyboardRef.current) return;
+          event.preventDefault();
+          triggerRef.current?.focus();
+        }}
+        aria-label={localizedStrings.toggleFilters}
```
lyonsil: **#4** - low · checked and confirmed

A screen reader announces the filters panel as "Toggle filters, dialog" — the name of the button that opened it, not a name for the panel.

**What happens:** `PopoverContent` renders `role="dialog"`, and `:175` names it with `localizedStrings.toggleFilters` — the same `%webView_find_toggleFilters%` string used for the trigger's own `aria-label` at `:128` and its tooltip at `:135`, whose English value is "Toggle filters". The panel's accessible name is therefore a verb phrase describing the control the user just left.

**Why it matters:** that name is the first thing announced on open and it is the only name this surface has. The panel is still reachable and each group keeps its own `legend`, so this is a naming defect rather than a barrier.

**Fix:** add a noun-phrase string — e.g. `%webView_find_filtersPanel%`, "Search filters" — and use it at `:175`. That means the new key under `en` in `extensions/src/platform-scripture/contributions/localizedStrings.json`, plus `FindFiltersStrings` (`:21`), `FIND_LOCALIZED_STRING_KEYS` (`find.component.tsx:142`), the mapping at `find.component.tsx:1099`, and the test mock at `find.component.test.tsx:381`.

_How this was checked: One string reaches all three places: `%webView_find_toggleFilters%` is "Toggle filters" at `localizedStrings.json:486`, is mapped to `toggleFilters` at `find.component.tsx:1099`, and is read at `find-filters.component.tsx:128` (trigger `aria-label`), `:135` (tooltip) and `:175` (the dialog's `aria-label`). No separate name for the panel exists in the file._

lyonsil (commented):
Review of the Find filters keyboard-access changes on this branch. Three findings, all low severity: two are attached to the lines they concern, and one is carried below because the file it concerns is not part of this diff.

Each comment leads with its severity and a status. `checked and confirmed` means the finding was verified against the code on this branch; `needs a human call` would mean it could not be settled either way and the judgement is yours.

## Findings that could not be attached to a line in this diff

**#2** (low · checked and confirmed) e2e-tests/tests/isolated/find/replace.spec.ts:150 - `replace.spec.ts` opens the filters panel with a bare click, which closes the panel whenever it is already open.

**What happens:** `openFiltersPanel` at `replace.spec.ts:150`-`:154` clicks the trigger unconditionally. That trigger toggles a Radix popover, so it closes an open panel as readily as it opens a closed one — which is why `find.spec.ts:449` replaced exactly this code with an `aria-expanded` check wrapped in `toPass`. The single caller at `replace.spec.ts:375` is safe only because it runs immediately after `openFindPanel`.

**Why it matters:** add a second caller, or let an earlier step leave the panel open, and the click shuts it; every `#allowRegex` interaction from `:376` down then fails on a missing control rather than on the real cause. This file is not in this diff, so it is a latent duplicate rather than a regression.

**Fix:** move the `find.spec.ts:449` version into `e2e-tests/fixtures/find.fixture.ts`, which both specs already import and which can import `isPopoverTriggerExpanded` from `e2e-tests/fixtures/helpers.ts:420`, then delete both local copies.

_How this was checked: Both helpers were read side by side on the branch: `replace.spec.ts:150` is the bare `click()` form, `find.spec.ts:449` is the converging form, and their doc comments describe the same toggle hazard. `grep` for `openFiltersPanel` in `replace.spec.ts` returns the definition and one caller (`:375`), which is what limits this to a latent problem today._

_Not attached to a line: no longer on a changed line; cannot be posted inline_

lyonsil (approved):
Approving - the "low" items I added are optional.

> @lyonsil reviewed 6 files and all commit messages, made 1 comment, and resolved 9 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2802#-:-P1k8gEnAE68buRCWEWy:b-7nwpaa)* status: all files reviewed, 2 unresolved discussions (waiting on timothy-mccormack).

e2e-tests/tests/isolated/find/find.spec.ts
timothy-mccormack: Taken in `7350e62`. `.catch((): number[] => [])` at the creation site, so the promise can never reject unhandled whatever the `try` does.

Also took the warning not to await it in the `finally`: on the failure path `countChangedAt` is never set, so the loop runs its full `timeoutMs` and awaiting there would add 35s to a test that has already failed.

extensions/src/platform-scripture/src/find/find-filters.component.tsx
timothy-mccormack: Taken in `7350e62`. New `%webView_find_filtersPanel%` — "Search filters" — threaded through `FindFiltersStrings`, `FIND_LOCALIZED_STRING_KEYS`, the mapping in `find.component.tsx`, and the test mock. English only; translations belong to localizers.

Worth noting this is the tail of your round-1 #5. The panel had no name, I gave it the trigger's string, and that was the wrong name rather than a missing one — a verb phrase describing the control the user just left, announced as the first thing about the surface they arrived at. The existing test asserted the old name under the title "names the open panel after the button that opens it", which was accurate and is exactly the thing that was wrong; it is now "gives the open panel a name of its own".

timothy-mccormack:
On **#2** (`replace.spec.ts:150`), which had no line to attach to: agreed on the diagnosis, declining it for this PR.

`replace.spec.ts` isn't in this diff, and its single caller at `:375` runs straight after `openFindPanel`, so it's latent rather than broken. Widening the diff into an untouched spec after approval seemed the worse trade — but the duplicate is real, and your fix is the right one: move the converging version into `find.fixture.ts`, which both specs already import and which can reach `isPopoverTriggerExpanded`, then delete both local copies. Happy to do it here instead if you'd rather it not wait.

lyonsil (approved):
:lgtm:

> @lyonsil reviewed 6 files and all commit messages, made 1 comment, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2802#-:-P1krEBNEIbdswU0GfDg:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

timothy-mccormack:
Heads up before you re-check: this branch now touches two files outside Find, and I'd rather flag it than have it appear in the diff.

Merging current `main` broke the filters-button stability test — the 14px toolbar shift was back. It isn't Find's doing. `web-view-content-zoom.bootstrap-script.ts` appends the zoom indicator's live region with `position:absolute` and no `top`/`left`, so it sits at its static position at the end of the body and still counts toward scrollable overflow. One pixel is enough to give a full-height pane a document scrollbar, and a toggling scrollbar takes its width out of the viewport each time. Every web view runs that bootstrap, so Find is just where there happened to be a test sensitive enough to notice.

How I narrowed it: the document, not the Find root, is what scrolls — `clientWidth` 319 → 305 while the trigger moved 271 → 257. Disabling only the whole-iframe zoom fallback changed nothing; reverting PT-4576 wholesale cleared it. That put it in the injected bootstrap rather than the zoom application path.

The fix is `top:0;left:0` on that region, with a unit test that fails without it. Happy to split it into its own PR against PT-4576 if you'd rather this branch stayed inside Find — it's one line and I've kept it as its own commit.

lyonsil (approved):
:lgtm: - Your fix looks good.

> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2802#-:-P1pn--8EZvu58h9NfwT:bf2cfpr)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2807, opened 2026-09-14
PT-4313: Compensate every indented paragraph marker in the gutter view

irahopkinson:
## Review follow-up (commit `c620947aa2c`)

The Reviewable pass on this PR changed the marker set and hardened the guard:

- The `tr`, `tr1`, `tr2` entries are gone from all three copies. A table row is a `<tr>`, never a `.para`, so the glyph rule cannot match them; the test excludes those three markers from the derivation instead. The fix now covers 16 markers: `ipq imq ipr psi qd p2 iq iq1 iq2 iq3 qm2 qm3 ph ph1 ph2 ph3`.
- `.book` gets the same containing block and variable defaults as `.para` in the shipping copy, and the RTL gutter padding reads `--psc-gutter-width` instead of a hardcoded `4em`.
- The coverage test resolves the LTR-over-agnostic cascade before filtering zero values, classifies direction per selector rather than per comma group, fails loudly on a `margin` shorthand or logical `margin-inline-*` on a marker rule, asserts each marker's RTL `margin-right` equals its LTR `margin-left`, compares the two core copies' derived gutter maps directly, and carries a small `usfm.sty` oracle so a base value drifting from the spec is caught rather than compensated.
- The invariant is scoped to margins set in each file. Margins from the commentary stylesheets (`marker-styles/*.scss`) and from `generateUsjCss` are not compensated; that is PT-4624, which moves the compensation into the generators that emit those margins.
- Recorded as `adr-derived-stylesheet-coverage-tests` in the architecture-decisions log.

The review summary below is the record from the first pass; counts in it have been corrected to the 16-marker set.

# Code Review Summary

**Branch**: pt-4313-gutter-indent-compensation (paranext-core), with the parallel branch of the same name in scripture-editors

**Base**: origin/main (both repos)

**Date**: 2026-09-14

**Files changed**: 3 in paranext-core, 2 in scripture-editors

## Overview

Fixes PT-4313. In the Simple editor's gutter view, a paragraph's marker glyph is pulled into the gutter by `--para-indent`, which must equal the paragraph's text-spacing margin, and the active focus box starts at `--verse-text-start`, which must equal a hanging paragraph's negative text-indent. Sixteen paragraph markers (`ipq imq ipr psi qd p2 iq iq1 iq2 iq3 qm2 qm3 ph ph1 ph2 ph3`) had a margin but no compensation, so the glyph sat at the margin edge and, for the hanging-indent markers, was painted over the first line of text at every column width. Measured live in Simple on P_HBKENG's Job introduction at the 297px column floor, the `\iq1` gap went from -38.5px to +20.8px.

The fix adds the missing entries to all three hand-maintained copies of the stylesheet: the extension's `_usj-nodes.scss` (what Simple renders), the platform-bible-react demo's `usj-nodes.css` (what the Storybook stories render), and upstream `packages/platform/src/usj-nodes.css` in scripture-editors. The coverage test that guards the gutter block was rewritten to derive its expectations from each file's own base text-spacing rules instead of two hand-typed maps, which had encoded the qm2/qm3 gap as expected. The core test now runs that derivation over both core copies; upstream gains the same test as `usj-nodes.css.gutter-coverage.test.ts`. The RTL mirror rule reads the same variables, and every one of the 16 markers has symmetric LTR/RTL margins, so the direction-agnostic compensation is correct for RTL projects too (confirmed live on the Aramaic project zzzz14, Psalm 1).

## API Changes

None. In paranext-core both changed source files are internal to the `platform-scripture-editor` extension, and the demo stylesheet is not exported from the platform-bible-react library entry point (Storybook stories and tests are its only importers). In scripture-editors the change to `usj-nodes.css` is additive only (new selectors under the existing `.psc-gutter-markers.text-spacing` scope; nothing removed or changed), the new test is excluded from the published tarball, and the committed `dist/` is correctly untouched because `usj-nodes.css` is not part of the bundle: consumers copy the source file.

None.

- [x] **Third stylesheet copy `lib/platform-bible-react/src/components/demo/scripture-editor/usj-nodes.css` had the same 16-marker gap and nothing tested it** _(fixed during review: mirrored the 16 `--para-indent` and 4 `--verse-text-start` entries into the demo copy, and parameterised `usj-nodes-scss-coverage.test.ts` over both core copies so each is checked against its own base rules; red on the demo copy before the CSS edit, 6/6 green after. The demo folder's own 26-test pin suite still passes. No library rebuild needed: the demo folder is not in the bundle.)_

Author response: Author asked for the demo copy to be fixed and for the test to cover it, preferring a per-file derived check over a byte comparison between copies since the copies legitimately differ outside the gutter block.

### Minor — Consider

- [x] **`_usj-nodes.scss` header's "Local additions" note listed only the li/lim entries and an upstream commit hash, now incomplete** _(fixed during review: replaced with a structural statement that the compensation block must stay complete on re-sync because the coverage test derives the requirement from the file; canonical upstream link moved to the paranext org. Author chose the generic wording over a marker list or commit citation because both go stale.)_
- [x] **`isLtrBaseBlock` name understated what it matches (direction-agnostic blocks too)** _(fixed during review: renamed to `isNonRtlBaseBlock` in both repos.)_
- [x] **Core test header line exceeded the 100-column wrap** _(fixed during review: rewrapped.)_
- [x] **`tr`, `tr1`, `tr2` compensation entries are inert for real tables** (table rows render as `<tr class="table-row usfm_tr">` with an inline `text-indent: 0`, not as `.para`, so the glyph rule never matches; confirmed from `ImmutableTableRowNode.createDOM` by three passes) _(first pass: a note in the compensation comment saying the entries were kept only for the coverage test. Superseded by the review follow-up: the six entries were removed from all three copies and the test excludes table-row markers from the derivation.)_
- [x] **Quote-normalisation comment in both tests said the copies "differ only in quote style", overstating their similarity** _(fixed during review: reworded to state which copy uses which quote style.)_
- [ ] ~~**Upstream test duplicates the vitest-environment pragma and file-read boilerplate from the sibling `usj-nodes.css.test.ts`**~~ _(Author accepted the duplication: the two tests guard different contracts, and folding a 200-line derived parser into a 7-case pin file would make both harder to read.)_
- [x] **Upstream test file name did not follow the sibling's `<file-under-test>.test.ts` pattern** _(fixed during review: renamed to `usj-nodes.css.gutter-coverage.test.ts` via `git mv`; the CSS comment referencing it was updated; vitest picks it up.)_
- [x] **Upstream test header said "paranext-core carries a copy with the same test" without naming it** _(fixed during review: names `extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts`.)_
- [ ] **Pre-existing, out of scope: `packages/platform/src/editor/generateUsjCss.ts` emits `text-indent` and margins from a project's stylesheet but never emits `--para-indent` or `--verse-text-start`**, so a project stylesheet that changes a marker's indent will still desync its gutter glyph. Not introduced by this branch. _(Recorded here during the first pass; filed as PT-4624 after the reviewer widened it to the commentary stylesheets.)_
- [x] **UX guideline asks for before/after captures on rendering changes** _(fixed during review: three screenshots captured and saved to `Downloads/pt-4313-screenshots/`: before and after at the 297px floor on P_HBKENG Job intro (LTR, `\iq1`/`\iq2`), and an after capture on the RTL Aramaic project zzzz14 Psalm 1 (`\q1`). To attach to both PRs. Caveat for the PR text: zzzz14 contains none of the 16 fixed markers, so the RTL capture confirms the mechanism for a hanging-indent marker rather than showing the fix.)_

Author response: Author addressed every minor finding except two: accepted the upstream test boilerplate duplication with reasoning, and asked that the pre-existing `generateUsjCss.ts` gap be recorded here rather than ticketed (later filed as PT-4624).

#### Shared Regions Modified

None. No `#region shared with` markers exist in any changed file (the extension's shared regions are in `tailwind.css` and `webpack-env.d.ts`, untouched).

#### Extension Config Changes

- [n/a] `extensions/src/platform-scripture-editor/src/_usj-nodes.scss` — extension-specific stylesheet, no propagation needed
- [n/a] `extensions/src/platform-scripture-editor/src/usj-nodes-scss-coverage.test.ts` — extension-specific test, no propagation needed
- [n/a] `lib/platform-bible-react/src/components/demo/scripture-editor/usj-nodes.css` — Storybook-only demo copy, no propagation needed

#### Stylesheet mirrors (not a template mechanism)

- [x] scripture-editors `packages/platform/src/usj-nodes.css` — identical gutter-block change on the parallel branch
- [x] `lib/platform-bible-react/.../demo/scripture-editor/usj-nodes.css` — fixed during review (see Important)
- [n/a] scripture-editors `demos/platform/lib/.../usj-nodes.css` — frozen `platform_v0.8.0` snapshot with no gutter block at all; not a maintained mirror

## Positive Observations

- Three independent passes ran the revert test: the derived coverage test fails against the merge-base stylesheet on exactly the 16 `--para-indent` and 4 `--verse-text-start` gaps (19 and 7 at the time of the first pass, before the table-row entries were dropped), and passes on HEAD.
- All 16 values were cross-checked against their base rules by every pass, and RTL parity (`[dir=rtl] margin-right` equals `[dir=ltr] margin-left`) was confirmed for all 16, so the direction-agnostic variables are correct in both directions.
- The test guards its own blind spots: a vacuity check on derived-list sizes plus spot values, a brace-depth check for nested rules the flat parser cannot see, a check that no gutter rule is direction-qualified, and a reverse check for stale entries with no base rule.
- Comments are forward-facing: the CSS states the invariant and names the test that enforces it; the old "how the li/lim entries went missing" narration was removed. No ticket IDs in code or comments, which the upstream repo's CLAUDE.md requires.
- Upstream conventions honoured: the committed `dist/` was left untouched with evidence (zero gutter rules in `dist/index.css`); `pnpm rebuild-committed-output` was run anyway and produced no change.
- New selectors are slotted into the existing value-grouped blocks and the new `--verse-text-start` block follows the section's per-family convention.

## Interview Notes

- **Stated purpose**: fix PT-4313 across the wider marker set (the 16 markers whose base rules indent them), make the coverage test structural so the gap cannot recur, and keep the same change on the upstream branch. The author explicitly retired the "tier 1" term: it is not used anywhere in code, comments or commit messages, since the list it named was a snapshot of what the engine implemented at one point.
- **Scope decisions**: 16 markers rather than the 14 with a visible defect, so the derived test can state one rule with a single exception: table rows (`tr`, `tr1`, `tr2`) never render as `.para`, have no entry, and are excluded from the derivation.
- **Test design**: author chose per-file derivation over a byte comparison between copies because the copies legitimately differ outside the gutter block. Accepted duplication of the vitest boilerplate between the two upstream stylesheet tests.
- **Header wording**: author chose a generic structural note over a marker list or an upstream commit citation because both go stale.
- **Follow-up**: `generateUsjCss.ts` (project-stylesheet CSS) has no gutter compensation path; recorded here first, filed as PT-4624 after review.
- **Upstream sequencing**: PR against scripture-editors `main`; after merge someone moves `platform-yalc` to include it for core consumption.
- The author demonstrated clear understanding throughout; no unresolved items and no areas deferred to the AI without explanation.

## In-Review Quality Check

paranext-core, all passing with no fixes needed:

- Library rebuild: not needed. The demo stylesheet is not in the library's vite entries or committed `dist/`; its importers are Storybook stories and tests only.
- `npm run typecheck`: pass, exit 0, all four sub-checks clean.
- `npm run lint`: pass, exit 0. One pre-existing warning in an untouched file (`unsubscriber-async-list.test.ts`, `import/no-named-as-default`).
- Prettier on the three changed files: unchanged.
- `npm test`: pass, exit 0. 657 test files passed (2 skipped), 9,288 tests passed (17 skipped), 0 failed.
- Not run: `dotnet test` (no C# changed).

scripture-editors, after the last edit: Prettier and ESLint clean on both changed files; `vitest run usj-nodes` in `packages/platform` passes 10/10 (7 in the existing pin test, 3 in the renamed gutter test). `nx typecheck` was not run; a raw `tsc` over the spec project shows only pre-existing project-reference errors unrelated to the new file.

## Suggested Review Focus

- [ ] Scope: 16 markers derived from the base rules versus the narrower core paragraph set under discussion on PT-4542 with Todd and Tom. The fix is additive, so a narrower answer changes ticket wording only; confirm the reviewer agrees.
- [ ] The derived coverage test's parser assumptions: flat `selector { declarations }` blocks, LTR-over-agnostic cascade, RTL blocks excluded, quote normalisation. Worth a read since it now guards three files.
- [x] Whether the inert `tr`/`tr1`/`tr2` entries and their comment are acceptable, or whether the reviewer would rather the test exclude table-row markers. _(Reviewer asked for exclusion; entries removed.)_
- [x] The recorded follow-up in `generateUsjCss.ts`: does project-stylesheet CSS need gutter compensation, and who owns that? _(Yes; PT-4624.)_
- [ ] Attach the three screenshots from `Downloads/pt-4313-screenshots/` to both PRs, with the caveat that the RTL capture shows the mechanism on `\q1`, not the fix.

---

## Screenshots

Before and after at the 297px Simple column floor, P_HBKENG Job introduction (`\iq1`, `\iq2`), and an RTL confirmation on the Aramaic test project zzzz14 Psalm 1 (`\q1`). The RTL project contains none of the 16 fixed markers, so that capture shows the mechanism for a hanging-indent marker rather than the fix itself.

**Before, LTR, 297px column floor, P_HBKENG Job introduction.** The `\iq1` and `\iq2` glyphs are painted over the first line of text.

Before: iq1 and iq2 gutter glyphs overlapping the text at the 297px column floor

**After, same view.** The glyphs sit in the gutter (x 26 to 62) with gaps of +20.8px and +35.7px.

After: iq1 and iq2 gutter glyphs in the gutter at the 297px column floor

**RTL confirmation, zzzz14 (Aramaic, Syriac script) Psalm 1, 958px.** Gutter on the right, `\q1` glyphs at x 896 to 932, gaps +72.5px and +53.9px. This project has none of the 16 fixed markers, so this shows the mechanism for a hanging-indent marker in RTL rather than the fix.

RTL: q1 gutter glyphs on the right of the text in the Aramaic project

## Companion PR

The identical stylesheet change for upstream `packages/platform/src/usj-nodes.css`, with the same derived coverage test, is paranext/scripture-editors#10 (https://github.com/paranext/scripture-editors/pull/10). After it merges, `platform-yalc` needs moving to include it for core consumption.

AI-assisted — Claude Fable 5.1, local Claude Code session (no session URL)

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2807)


katherinejensen00 (changes requested):
# PR #2807 — PT-4313: Compensate every indented paragraph marker in the gutter view

Review of `pt-4313-gutter-indent-compensation` (3 files, +300/−127).

**The fix is correct.** Verified independently with postcss rather than by replaying the PR's own
parser: 55 non-zero base `margin-left` markers ↔ 55 `--para-indent`; 33 negative `text-indent` ↔ 33
`--verse-text-start`; identical in both copies; every one of the 19 added markers matches its base
rule. Zero asymmetric, one-sided, or shorthand margins across 129 markers. Moving the test from
hand-typed maps to a derivation is a real upgrade, and the bidirectional check
(`valueMismatches` + `unexpectedMarkers`) makes the two sets provably equal.

Findings below are about the **guard**, not the fix. Three themes: the invariant it asserts is false
in the commentary/project-stylesheet path; it checks each copy against itself rather than against
the other, and is green on two live drifts; and a few parser assumptions fail silently.

Severity: **High** = wrong rendering today or a silent hole in the new guard · **Medium** = latent
hole or misleading failure · **Low** = correctness-adjacent cleanup.

---

## 1. The "every indented marker has an entry" invariant is false whenever a commentary stylesheet loads

**High** · `extensions/src/platform-scripture-editor/src/_usj-nodes.scss:2850`
`EVERY marker with a text-spacing margin-left above needs an entry here, or its gutter`

`platform-scripture-editor.web-view.tsx` both injects PT9-derived commentary styles (`:2354`) and
renders the gutter (`:3703`) — same web view, same DOM. Those sheets are equal-specificity and later
in source order, so they win.

- `hbkeng.scss:770` sets `.usfm_q { margin-left: 22.98vw }`; base is `15vw` (`_usj-nodes.scss:782`)
  and `--para-indent` for `q` stays `15vw` (`:2873`). ~8vw under-compensation puts the `\q` glyph on
  the poetry text — the PT-4313 symptom itself.
- Markers that exist *only* in commentary sheets can never have an entry. Verified: `gll2`
  (`tnn.scss:1819`, `margin-left: 30vw`), `phi`, `a`, `sl1`, `be` all have **0** base rules and **0**
  gutter entries in `_usj-nodes.scss`. Zero compensation against up to 30vw.

Pre-existing, not caused by this change — and the live measurement was clean because `hbkeng.scss:455`
happens to set `iq1` to `20vw`, matching the base. But the new comment and test now assert a
completeness that does not hold here. Same class as the `generateUsjCss.ts` gap already recorded as a
follow-up. Either scope the claim to non-commentary rendering, or extend the derivation over
`marker-styles/*.scss`.

## 2. The shipping copy never gives `.book` a positioned containing block; the other two copies do

**High (pre-existing)** · `extensions/src/platform-scripture-editor/src/_usj-nodes.scss:2784`
`.psc-gutter-markers .para {`

Demo (`usj-nodes.css:2791-2792`) and upstream (`:2477-2478`) scope this as `.para, .book`. This copy
has only `.para`, yet `:2811` absolutely positions `.psc-gutter-markers .book > .marker:first-child`.
`.book` receives `position: relative` nowhere in the file, and `--para-indent` is never reset on it —
so `var(--para-indent)` at `:2815` has no value and no fallback, making the whole `left` declaration
invalid at computed-value time. (Contrast `:3033`, which correctly writes `var(--verse-text-start, 0px)`.)

Raised mainly as proof for #4: a live cross-copy drift the new guard is green on.

## 3. `accept` runs before the cascade merge, so a zero override is discarded and the test enforces a stale value

**High** · `usj-nodes-scss-coverage.test.ts:179`
`if (value === undefined || !accept(value)) return;`

Verified by mutation. Add `.text-spacing[dir='ltr'] .usfm_mi { margin-left: 0 }` — specificity (0,3,0)
beats the direction-agnostic `5vw` at `:765`, so the browser renders `\mi` at margin 0. But
`isNonZeroLength('0')` is false, the LTR entry is dropped before the merge, and the agnostic `5vw`
survives as the expectation:

```
mi: expected=5vw actual=5vw | suite: GREEN
```

The test therefore *demands* `--para-indent: 5vw` stay, pushing every `\mi` glyph 5vw off the gutter,
and never complains. Resolve the winning value first, then decide whether it needs compensation.

## 4. The test checks each copy against itself, never against the other

**High** · `usj-nodes-scss-coverage.test.ts:252`
`describe.each(STYLESHEETS)('$name .psc-gutter-markers.text-spacing coverage', ({ path }) => {`

Two live drifts pass all six assertions today: finding #2, and this copy's physical
`padding-left: var(--psc-gutter-width)` (`:2778`) versus demo/upstream's logical
`padding-inline-start`. Future case: change `iq1` from `20vw` to `30vw` in both the base rule *and*
the compensation in only the demo copy — each file stays self-consistent, suite green, Storybook and
Simple render different geometry indefinitely.

The PR rejects a byte comparison because the copies legitimately differ outside the gutter block,
which is right — but a comparison scoped to the extracted `.psc-gutter-markers*` rules has no such
problem, and both sheets are already parsed in one process.

## 5. The `tr` entries contradict a pinned repo decision and ship dead CSS four times over

**Medium** · `extensions/src/platform-scripture-editor/src/_usj-nodes.scss:2869`
`.psc-gutter-markers.text-spacing .usfm_tr,`

`tools/pt9-css-converter/src/convert.ts:24-28` strips `\tr` from every generated sheet precisely
because its indent/margin rule "would inherit into cells and collapse the columns," pinned by
`convert.test.ts:103` (`expect(scss).not.toContain('.usfm_tr')`). This PR adds six `.usfm_tr*` rules
whose only stated justification is keeping a test green.

`_usj-nodes.scss` is `@use`d by four web-view stylesheets, each inlined separately as `?inline`
(`main.ts:28,30,32,35`), so the dead rules ship 4×. And they block the right cleanup: deleting the
`\tr` base rules at `:1800-1806` now fails the new test.

The entries *are* genuinely inert — confirmed structurally, not just from the PR's comment: `.para`
comes only from `ParaNode`, and `psc-active-text` is applied to `getTopLevelElement()`, which the
adaptor makes the `<table>` itself (`usj-editor.adaptor.ts:862`: "A table is a block root in its own
right"), an **ancestor** of the `<tr>`. Custom properties inherit downward only. A three-marker
exclusion in the test is one line in one file. Same at demo `usj-nodes.css:3037`.

## 6. Nothing verifies the LTR/RTL symmetry the direction-agnostic design depends on

**Medium** · `usj-nodes-scss-coverage.test.ts:134`
`!selectors.includes("[dir='rtl']")`

The same `--para-indent` feeds the LTR `left` calc (`:2815`) and the RTL `right` calc (`:3007`), so
it must equal the paragraph's inline-start margin in *both* directions. The derivation reads only
`margin-left` and excludes RTL blocks entirely. All 55 markers are symmetric today (verified), so
this is latent — but a sync changing only `.usfm_pi2`'s RTL `margin-right` keeps the suite green and
puts the RTL glyph inside the text.

Worth noting the test already enforces the *converse* (`directionQualifiedGutterRules` forbids a
`[dir=…]`-qualified gutter rule) without ever establishing that one value can serve both.

## 7. Deriving from the file under test drops the `usfm.sty` oracle that was the original point

**Medium** · `usj-nodes-scss-coverage.test.ts:21`
`// https://github.com/ubsicap/usfm/blob/master/sty/usfm.sty`

The old `EXPECTED_PARA_INDENT` map encoded USFM LeftMargin values independently; four spot checks now
replace ~88 literals. A re-sync typo changing `.usfm_li` from `10vw` to `12vw` fails once, gets
"fixed" by updating `--para-indent` to match, and goes green — with nothing left asserting 0.5".
`Testing-Guide.md` §"Prohibited Test Patterns → Implementation-Mirroring" speaks to this directly. A
small `usfm.sty`-derived cross-check alongside the derivation would restore it.

## 8. Direction is classified per-block, but selectors are comma groups

**Medium** · `usj-nodes-scss-coverage.test.ts:180`
`const target = block.selectors.includes("[dir='ltr']") ? ltr : agnostic;`

`block.selectors` is the whole comma-joined list, so one `[dir='rtl']` fragment discards the entire
group. Verified by mutation — compacting `li`'s base rule into a mixed-direction group:

```
li: expected=none actual=10vw | size=54 | suite: RED (1 unexpected)
```

Not a silent pass, but a *misdirecting* failure: `unexpectedMarkers` reports the correct
`--para-indent: 10vw` entry with the message "no base text-spacing rule calls for it", pointing the
maintainer at deleting a working rule. No block mixes directions today, but comma grouping is this
file's normal idiom — the gutter block itself groups 13 selectors. Split on `,` and classify per
selector.

## 9. The demo copy's "deliberate divergences" list was not updated for 26 new local rules

**Medium** · `lib/platform-bible-react/src/components/demo/scripture-editor/usj-nodes.css:2911`
`.psc-gutter-markers.text-spacing .usfm_p2 {`

That header (`:1-12`) enumerates three divergences from upstream, and `usj-nodes-styles.test.ts:274`
(`describe('divergences from the source recorded in the header')`) exists to keep it honest. The
`.scss` copy got a re-sync note at its line 7; this one got nothing. The next person re-syncing reads
the header, sees only colour divergences, overwrites wholesale, and loses all 26 local rules.

## 10. The table-row comment states a mechanism that does not exist

**Low-Medium** · `extensions/src/platform-scripture-editor/src/_usj-nodes.scss:2854`
`tables: a table row renders as <tr class="table-row usfm_tr"> with an inline text-indent of`

`ImmutableTableRowNode.createDOM` (`:88-92`) sets only `data-marker` and two classes. The only inline
style anywhere in the table nodes is `dom.style.textAlign` on the *cell* (`ImmutableTableCellNode.ts:149`).
The zeroing is the stylesheet rule at `:1817`, and it is not the load-bearing reason anyway — the
`.para` gating at `:2797`/`:2810` is.

Origin: the demo copy's *pre-existing* comment at `usj-nodes.css:1819-1821` already asserted this, so
the PR propagates a wrong claim into two more places. Suggested: *"renders as
`<tr class="table-row usfm_tr">`, not a `.para`, so the gutter glyph rule never matches."* Same at
demo `:2908`.

## 11. Both cross-copy pin suites now assert something this PR made false

**Low-Medium** · `extensions/src/platform-scripture-editor/src/usj-nodes-styles.test.ts:243` and
`lib/platform-bible-react/src/components/demo/scripture-editor/usj-nodes-styles.test.ts:327`
`// different upstream commits, and nothing structural keeps their rule sets in agreement —`

Both suites explain their hand-written pins as a workaround for having no structural guard. This PR
built the first one. A maintainer reading either concludes they must still hand-forward every gutter
change; the bad outcome is someone "closing the gap" by duplicating the derived parser into the pbr
workspace — the duplication the PR deliberately avoided. One sentence each.

## 12. The demo copy's header still points at the old org

**Low-Medium** · `lib/platform-bible-react/src/components/demo/scripture-editor/usj-nodes.css:2`
`https://github.com/eten-tech-foundation/scripture-editors/blob/main/packages/platform/src/usj-nodes.css`

`_usj-nodes.scss:2` was moved to the `paranext` org in this PR and this file was edited in the same
commit. `dev-packages.json` pins `paranext/scripture-editors` as canonical, and
`.erb/scripts/stage-dev-packages.util.test.ts:31-33` asserts the two orgs are *different repos*
("silently builds another organization's source"), so the rename matters. `_usj-nodes.scss:4` still
carries the old org too, leaving that header self-contradictory.

## 13. `declarationValue` reads longhand only

**Low** · `usj-nodes-scss-coverage.test.ts:114`
`function declarationValue(declarations: string, property: string): string | undefined {`

A marker written `margin-inline-start: 15vw` or with a `margin` shorthand gets no base entry *and* no
gutter entry, so both checks stay silent and the test passes — the exact regression class it exists
to prevent. Not hypothetical: `:2518` already uses `margin: 0px 2px 0px 2px` inside a `.text-spacing`
rule, and demo `usj-nodes.css:2899` calls the block "the **inline-start** margin values above", so
logical properties are the natural next edit. A loud assertion that no `.text-spacing` rule uses those
spellings would fail at the right moment.

## 14. postcss is already a dependency — the hand-rolled parser can go

**Low** · `usj-nodes-scss-coverage.test.ts:59`
`function parseStylesheet(path: string): ParsedStylesheet {`

postcss is a devDependency of this extension (`^8.5.3`) and of the root (`^8.5.15`). `braceDepthAt`
and `nestingProblems` exist only to work around the flat regex parser's blindness to nesting, which
`rule.parent` gives for free; comment-stripping and quote-normalisation disappear; `declarationValue`
becomes `walkDecls`. ~60 lines deleted, and it closes #13 and part of #8 structurally rather than by
adding more assertions. This is exactly what was used to verify the CSS values above, so it is known
to parse both files cleanly.

## 15. The RTL gutter mirror hard-codes `4em` instead of the variable

**Low (pre-existing)** · `extensions/src/platform-scripture-editor/src/_usj-nodes.scss:2991`
`padding-right: 4em;`

`:2720` declares `--psc-gutter-width: 4em` with the comment "both values must be updated together if
the gutter size changes," and `:2778` reads the variable correctly. This mirror does not. Changing
the gutter width moves the LTR gutter and leaves RTL behind — the same desync class as PT-4313, eight
lines from the block being edited.

---

## Also noted (not in the top 15)

- Commit `a0928895538` has `Co-Authored-By:` but no `Session-URL:` trailer (`CLAUDE.md` §Git & PR Conventions).
- No `Architecture-Decisions.md` entry. Derive-from-file testing is a new pattern; the log already
  carries a contrary precedent in `adr-editor-edit-side-effects-shared-module`.
- 8 lines in the test exceed `.editorconfig`'s `max_line_length = 100` (12, 17, 52, 54, 140, 214, 236,
  255 — up to 103 chars). Prettier passes because it never rewraps comments and no `printWidth` is set
  for JS/TS anywhere in the repo.
- `nestingProblems` guards the two custom properties but not `margin-left`/`text-indent`, so a nested
  *base* rule would drop out unnoticed.
- `_usj-nodes.scss:2930` — section header says the focus box `::before`; the implementation uses
  `::after` deliberately, and a sibling test is named "never `::before`". Pre-existing, outside the hunk.
- The invariant is now stated in three hand-maintained places (file header `:7-10`, the block comment,
  the test header). The block is the better single home. No runtime cost — sass compiles `compressed`
  in production, so comments do not ship.
- `getGutterMarkerValues` is recomputed at `:204` and `:224`; hoisting `actual` beside the already-hoisted
  `expected` maps makes both helpers pure Map-diffs and collapses the 4-line signature at `:145`.

## Checked and deliberately not raised

- **The vacuity floors (`> 40` / `> 20` at `:263-264`) are fine.** They look loose against real sizes
  of 55 and 33, but deleting all thirteen `li`/`lim`/`ili` base margins drops `expected.size` to 42 —
  floor passes — while the suite goes **RED** with 13 `unexpectedMarkers`, because the gutter entries
  survive unjustified. The bidirectional check already covers it. `toBe(55)`/`toBe(33)` is still more
  honest and free, but cosmetic.
- **The RTL exclusion is not load-bearing.** A leaked RTL block lands in the `agnostic` map, which
  `new Map([...agnostic, ...ltr])` then overrides with the LTR entry. Every marker with an RTL margin
  also has an LTR one (zero one-sided markers), so no current exposure. Only the symmetry premise in
  #6 is a real gap.
- **The PR body's structure is mandated.** `.claude/commands/review-paratext.md:595-631` requires the
  summary block to be prepended with `\n\n---\n\n`, which is why the screenshots and Suggested Review
  Focus sit below it. Not an author choice.
- **`extensions/src/platform-enhanced-resources/src/_usj-nodes.scss`** is a fourth copy but has zero
  `psc-gutter-markers` rules — no compensation block to keep complete. Correctly out of scope.
- **The cross-workspace path is precedented.** `model-text-panel.stories.tsx:26` and two other
  extensions already reach into `lib/platform-bible-react/.../usj-nodes.css` the same way, and CI runs
  root `npm test` (`test.yml:272`) which fans out via `npm run test --workspaces`, so the demo copy is
  covered in CI. Only local per-workspace runs miss it.

## Verification performed

`npx vitest run usj-nodes-scss-coverage` 6/6 · pbr demo suite 35/35 · `prettier --check` and
`stylelint` clean on all three files · independent postcss parse of both copies (129 markers) ·
mutation tests for findings #3, #6, #8 and the vacuity floor · editor node sources read for
`ImmutableTableRowNode`, `ImmutableTableCellNode`, `ActiveTextPlugin`, `usj-editor.adaptor`.

> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2807#-:-P1WErGT7D52a2ntXR1e:b-vyy0gr)* status: all files reviewed, 1 unresolved discussion (waiting on irahopkinson).

___
*[`extensions/src/platform-scripture-editor/src/_usj-nodes.scss` line 2850 at r1](https://reviewable.io/reviews/paranext/paranext-core/2807#-P1WFg8Q1pKu-7YntFsJ:-P1WFg8Q1pKu-7YntFsK:bqsdspt) ([raw file](https://github.com/paranext/paranext-core/blob/a09288955382e9dd2ee66d0569d20e4c6b5b0933/extensions/src/platform-scripture-editor/src/_usj-nodes.scss#L2850)):*
> ```SCSS
>    pi2 and pi3 use the same primary-indent value in both LTR and RTL (10vw and 15vw
>    respectively), so no direction restriction is needed here.
>    EVERY marker with a text-spacing margin-left above needs an entry here, or its gutter
> ```

## 1. The "every indented marker has an entry" invariant is false whenever a commentary stylesheet loads

**High** · `extensions/src/platform-scripture-editor/src/_usj-nodes.scss:2850`
`EVERY marker with a text-spacing margin-left above needs an entry here, or its gutter`

`platform-scripture-editor.web-view.tsx` both injects PT9-derived commentary styles (`:2354`) and
renders the gutter (`:3703`) — same web view, same DOM. Those sheets are equal-specificity and later
in source order, so they win.

- `hbkeng.scss:770` sets `.usfm_q { margin-left: 22.98vw }`; base is `15vw` (`_usj-nodes.scss:782`)
  and `--para-indent` for `q` stays `15vw` (`:2873`). ~8vw under-compensation puts the `\q` glyph on
  the poetry text — the PT-4313 symptom itself.
- Markers that exist *only* in commentary sheets can never have an entry. Verified: `gll2`
  (`tnn.scss:1819`, `margin-left: 30vw`), `phi`, `a`, `sl1`, `be` all have **0** base rules and **0**
  gutter entries in `_usj-nodes.scss`. Zero compensation against up to 30vw.

Pre-existing, not caused by this change — and the live measurement was clean because `hbkeng.scss:455`
happens to set `iq1` to `20vw`, matching the base. But the new comment and test now assert a
completeness that does not hold here. Same class as the `generateUsjCss.ts` gap already recorded as a
follow-up. Either scope the claim to non-commentary rendering, or extend the derivation over
`marker-styles/*.scss`.

irahopkinson (commented):
Pushed a second commit (`c620947aa2c`, branch rebased on `main`) addressing every finding except #14. The same test changes are mirrored on the upstream branch (paranext/scripture-editors#10, `a65fc748`).

**Stylesheets**

- **#1** Scoped the invariant to margins set in each file and named what it does not cover (commentary stylesheets, `generateUsjCss`). Follow-up filed as [PT-4624](https://paratextstudio.atlassian.net/browse/PT-4624): the compensation moves into the generators (`tools/pt9-css-converter` for the committed commentary sheets, `generateUsjCss` for runtime StyleInfo) rather than into more hand-maintained sheets; `usj-nodes.css` is the placeholder default, so the block and this test retire when the generators cover it. Replied in the thread.
- **#2** `.book` added to the shipping copy's containing-block rule, so it gets `position: relative` and the variable defaults like the other two copies.
- **#5** `tr`, `tr1`, `tr2` entries removed from all three copies. The test excludes those three markers from the derivation with a comment citing `tools/pt9-css-converter`.
- **#9** Demo header's divergence list gains a bullet for the gutter block and names the test that guards it.
- **#10** Comment rewritten to give the load-bearing reason: a row is a `<tr>`, not a `.para`, so the glyph rule never matches. On the mechanism: `ImmutableTableRowNode.createDOM` does set `dom.style.textIndent = "0"` (line 104, with a comment on why it is inline; the stylesheet's `.table-cell` comment at 1804 points to it). It was never the reason the entries were inert, so the correction stands and the point is moot with #5.
- **#12** Both headers now point at the `paranext` org.
- **#15** RTL mirror reads `var(--psc-gutter-width)`.
- Also: `::before` → `::after` in the section header.

**Test**

- **#3** Cascade resolved first (`resolveBaseValues` keeps every value, zero included), then filtered; the `[dir='ltr'] .usfm_mi { margin-left: 0 }` mutation now fails.
- **#4** The two copies' derived `--para-indent` and `--verse-text-start` maps are compared directly. Scoped to those maps, so the `padding-left` versus `padding-inline-start` spelling difference is not caught by it; that is pre-existing and untouched here.
- **#6** `directionAsymmetries` asserts each expected marker's RTL `margin-right` equals its LTR `margin-left`.
- **#7** A small `usfm.sty` oracle (one marker per distinct LeftMargin / FirstLineIndent value) sits beside the derivation. Mutating `li1` to 12vw in both the base rule and the compensation now fails.
- **#8** Selector lists are split on `,` and each selector classified on its own.
- **#13** `unreadableMarginSpellings` fails on a `margin` shorthand or `margin-inline-*` on any `.text-spacing` marker rule.
- `nestingProblems` also runs for `margin-left` and `text-indent`; `actual` maps hoisted beside `expected`; no line over 100 characters in either test file.
- Every new assertion was mutation-checked red before the stylesheet was restored.

**#14** Skipped by author decision. The flat parser plus its blind-spot assertions stays; the ADR entry records postcss as the alternative and when it becomes worth adopting.

**Other notes**

- **#11** Both pin-suite comments now say the gutter block is covered structurally by the coverage test and the pins cover the rest.
- ADR added: `adr-derived-stylesheet-coverage-tests`, including where the compensation ends up long-term.
- `Session-URL` is omitted because this was a local Claude Code session with no URL to cite; the PR body says so.
- Vacuity floors left as `> 40` / `> 20` per the "deliberately not raised" note.

> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2807#-:-P1XucVX8TQrmzl3YpjC:blg7ula)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on katherinejensen00).

___
*[`extensions/src/platform-scripture-editor/src/_usj-nodes.scss` line 2850 at r1](https://reviewable.io/reviews/paranext/paranext-core/2807#-P1WFg8Q1pKu-7YntFsJ:-P1Xtymk1_JjrNsyp8mp:b-scfe3d) ([raw file](https://github.com/paranext/paranext-core/blob/a09288955382e9dd2ee66d0569d20e4c6b5b0933/extensions/src/platform-scripture-editor/src/_usj-nodes.scss#L2850)):*

## 1. The "every indented marker has an entry" invariant is false whenever a commentary stylesheet loads

**High** · `extensions/src/platform-scripture-editor/src/_usj-nodes.scss:2850`
`EVERY marker with a text-spacing margin-left above needs an entry here, or its gutter`

`platform-scripture-editor.web-view.tsx` both injects PT9-derived commentary styles (`:2354`) and
renders the gutter (`:3703`) — same web view, same DOM. Those sheets are equal-specificity and later
in source order, so they win.

- `hbkeng.scss:770` sets `.usfm_q { margin-left: 22.98vw }`; base is `15vw` (`_usj-nodes.scss:782`)
  and `--para-indent` for `q` stays `15vw` (`:2873`). ~8vw under-compensation puts the `\q` glyph on
  the poetry text — the PT-4313 symptom itself.
- Markers that exist *only* in commentary sheets can never have an entry. Verified: `gll2`
  (`tnn.scss:1819`, `margin-left: 30vw`), `phi`, `a`, `sl1`, `be` all have **0** base rules and **0**
  gutter entries in `_usj-nodes.scss`. Zero compensation against up to 30vw.

Pre-existing, not caused by this change — and the live measurement was clean because `hbkeng.scss:455`
happens to set `iq1` to `20vw`, matching the base. But the new comment and test now assert a
completeness that does not hold here. Same class as the `generateUsjCss.ts` gap already recorded as a
follow-up. Either scope the claim to non-commentary rendering, or extend the derivation over
`marker-styles/*.scss`.
</blockquote></details>

Scoped the claim rather than extending the derivation. The compensation comment in all three copies now states the invariant is about margins set in that stylesheet, names the PT9-derived commentary stylesheets (`marker-styles/*.scss`) and project-stylesheet CSS (`generateUsjCss`) as loading into the same web view later in source order and able to move a marker's margin away from the compensated value, and says compensating those is open work. The test header and the new `adr-derived-stylesheet-coverage-tests` entry carry the same scope note.

Extending the derivation over `marker-styles/*.scss` would not close the gap: those margins differ per stylesheet (`hbkeng` gives `q` 22.98vw, the base gives it 15vw), so one static `--para-indent` in this file cannot serve them. The compensation has to be emitted where the margin is emitted, which is the generators: `tools/pt9-css-converter` for the committed commentary sheets (the step that reaches Simple today) and `generateUsjCss` for runtime StyleInfo. Filed as [PT-4624](https://paratextstudio.atlassian.net/browse/PT-4624). `usj-nodes.css` is the placeholder default sheet, so the hand-maintained block and this test retire once the generators cover the default.

katherinejensen00 (approved):
:lgtm:

> @katherinejensen00 reviewed 6 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2807#-:-P1__g5kAVhhHeNx5vaC:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #18, opened 2023-02-03
Prepare release

irahopkinson:
- fix Releases path link in README
- add Windows web installer and portable targets
- Preserve window dimensions and position

PR #2808, opened 2026-09-14
PT-4542: E2E guard against gutter glyphs overlapping paragraph text

irahopkinson:
## Summary

Adds the rendered-layout half of [PT-4542](https://paratextstudio.atlassian.net/browse/PT-4542), on top of the gutter compensation that #2807 merged: an isolated Simple-mode E2E that proves no paragraph-marker glyph overlaps its paragraph text in the gutter view at any column width down to the 297px floor. The stylesheet-level half (the derived coverage test) landed in #2807; this spec catches what that cannot, a regression that comes from anywhere other than the two custom properties (a gutter-width change, a padding change in `_simple-mode.scss`, an engine DOM change).

Commits, the first two deliberately split so the page-object move reviews as a rename:

1. **Move the onboarding-tour page object into `e2e-tests/fixtures`** — a pure move (two import paths), so fixture code can share the tour locators.
2. **PT-4542: E2E guard that gutter marker glyphs never overlap paragraph text** — the spec, the fixture helper, and the page-object changes.
3. Review follow-ups: measure from the line start, harden the spec, shared drag helpers, then the isolated-root guard on the write helpers, typed-Escape-first suppression, ticketed TODOs and the LTR pin.

## The spec

`e2e-tests/tests/isolated/scripture-editor/gutter-marker-no-overlap.spec.ts`, one `test()`:

- Makes the sample project editable, then rewrites Obadiah 1 over PAPI (`getChapterUSFM` → transform → `setChapterUSFM`) so it carries `li1`, `li2`, `q2`, `qm2` and `iq1` paragraphs in front of real verses. The sample text has `li1` and `q2` but none of the other three, and Simple's dropdown does not offer `li2`, so insertion goes through the project's data provider rather than the UI or the SFM on disk. `li1`/`li2` are the markers of the original [PT-4021](https://paratextstudio.atlassian.net/browse/PT-4021) report; `qm2` and `iq1` are two of the markers #2807 fixes.
- Opens the editor, suppresses the onboarding tour, navigates to Obadiah 1:8, and asserts the positive controls: every marker rendered, the editor root carries `psc-gutter-markers` and `text-spacing`, and its computed direction is `ltr`. The measurement keeps an RTL branch for the follow-up in "Out of scope", but this run must take the LTR one; pinning it means a run that ever came up RTL fails here instead of switching to arithmetic no run has exercised.
- Sweeps the editor column from its opened width to the floor by dragging the editor/resources divider in 120px steps, measuring after each. At each width, per paragraph: the glyph box ends before the first line of text begins (outcome), then `--para-indent` is a positive `vw` and `--verse-text-start` a negative one (mechanism). Asserts the sweep reached ≤ 310px so the narrow-column assertions are not vacuous. No pixel positions are pinned.

## Fixture changes

- `rewriteSampleProjectChapterUsfm` in `scripture-editor-helpers.ts`, sharing the provider lookup with `makeSampleProjectEditable`.
- **Write helpers refuse to run outside an isolated project root.** `launchElectronApp` now records the project root it set through `PLATFORM_BIBLE_PROJECT_ROOT_FOLDER`, and `requireIsolatedProjectRoot()` in `helpers.ts` throws unless the most recent launch used `isolatedProjectRoot: true` and that root is under the OS temp directory. Both `makeSampleProjectEditable` and `rewriteSampleProjectChapterUsfm` call it before their first write: they address the sample project by its fixed id on the fixed PAPI port, so against a normal launch they would have rewritten the developer's own copy of the sample project, with no restore. Every current caller already launches isolated, so this changes nothing for them.
- `onboarding-tour.page.ts`: button locators now resolve their names from `en.json` keys through `englishLabel` instead of hard-coded regexes. Behaviourally only Skip changes, from `/skip/i` to the whole label "Skip tour"; Next/Back/Done were already exact. Adds `suppressOnboardingTour`, which `waitForAppReady` now delegates to. **This changes how every spec dismisses an already-open tour.** It types Escape through `page.keyboard` first, which with host focus intact reaches the tour's `window` listener the way a user's key does; only if the dialog is still visible and focus is inside an iframe does it dispatch a synthetic `keydown` on `window`, which is needed once a scripture editor iframe holds focus and swallows the typed key ([PT-4627](https://paratextstudio.atlassian.net/browse/PT-4627)). If the typed key fails with focus anywhere else it throws instead of falling back, so a broken Escape handler fails every spec that suppresses the tour with host focus rather than being hidden by the fallback. The tour spec's own "Escape key closes the tour" test still exercises the typed path with no fallback. A Simple-mode spec that skips `waitForAppReady` must call it before its first main-frame click; this spec does, its two scripture-editor siblings predate the tour and do not, and fixture-level suppression is [PT-4639](https://paratextstudio.atlassian.net/browse/PT-4639). Also exports `getTourNextButton` / `getTourBackButton` / `getTourDoneButton`, which the tour spec now uses instead of hard-coded regexes.

## Verification

| Run | Result |
| --- | --- |
| Spec alone on this branch | passed, 2.0 min |
| Spec alone with the pre-#2807 `_usj-nodes.scss` swapped in (revert test) | failed at the opened width, first attempt and retry: "At 638px the \qm2 glyph overlaps its paragraph text by 55.8px"; stylesheet restored. Measured from the line start (verse number included), which is what the review asked for; from the text after the number the same render read a positive gap |
| Spec + full `onboarding-tour.spec.ts` in one run | 7 passed, 4 flaky (all passed on retry) |
| `onboarding-tour.spec.ts` alone, before vs after the page-object change | 9.9 min / 3 flaky vs 8.1 min / 2 flaky; per-test times unchanged within noise |
| `npm run typecheck:e2e`, ESLint on touched files, Prettier, `npm run lint` | clean |

## Known issues surfaced, not fixed here

- **Tour keyboard access once a web view has focus ([PT-4627](https://paratextstudio.atlassian.net/browse/PT-4627)).** Escape inside the editor iframe never reaches the tour's `window` listener and Tab moves behind the modal, so a keyboard user cannot dismiss the tour. The page object works around it with a synthetic Escape.
- **Interface-mode leak between workers ([PT-4628](https://paratextstudio.atlassian.net/browse/PT-4628)).** In the combined run this spec's first attempt launched in Power mode despite seeding Simple, because the preceding Power-mode tour test's settings restore runs in its worker teardown and landed after this worker's seed. That is the documented race in `e2e-tests/CLAUDE.md`; it did not occur in either solo run. Fix belongs in `isolated.fixture.ts`.
- **Pre-existing tour-spec flakes (suspected [PT-4621](https://paratextstudio.atlassian.net/browse/PT-4621)).** "Help > Show the tour reopens the completed tour" and the Power-mode replay test fail their first attempt (tour dialog not visible after the Help replay) both before and after this change. None of the first-attempt failures involve the relabelled locators. [PT-4621](https://paratextstudio.atlassian.net/browse/PT-4621) describes the tour closing itself when no step target has measured yet, which is what a replay that races the dock layout would look like; noted on that ticket rather than filed separately.

## Out of scope

RTL: the isolated project root holds only the left-to-right sample project, so this spec never takes the mirrored path (gutter on the right, the same `--para-indent` in the `right` calculation). The RTL check for this work is the manual CDP sweep on the Aramaic project recorded on #2807; a follow-up could set the sample project's `platform.textDirection` to `rtl` over PAPI and sweep a second time. Also out of scope: the Power-mode half of [PT-4021](https://paratextstudio.atlassian.net/browse/PT-4021) (#2743, [PT-4536](https://paratextstudio.atlassian.net/browse/PT-4536)), and [PT-4542](https://paratextstudio.atlassian.net/browse/PT-4542)'s ticket wording, which is settled on the wider marker set pending the Epic Lead's confirmation.

AI-assisted — Claude Fable 5.1, local Claude Code session (no session URL)

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2808)

[PT-4542]: https://paratextstudio.atlassian.net/browse/PT-4542?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PT-4021]: https://paratextstudio.atlassian.net/browse/PT-4021?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


katherinejensen00 (commented):
# PR #2808 review — PT-4542 gutter marker no-overlap E2E

Head 3552750d341 · 1 Medium, 10 Low, 7 Nit · "spec" = `e2e-tests/tests/isolated/scripture-editor/gutter-marker-no-overlap.spec.ts`

---

### 1. spec:144 · Medium
The overlap check measures from the text after the verse number, so a glyph painted over the verse number passes, including the qm2 regression this PR fixes.

Every tested paragraph starts with an inline verse number (`span.verse`), and the walker skips `.verse`, so the check can't see overlaps up to the number's width (about 15–22px). In a headless render with main's stylesheet at the 297px floor, the `\qm2` glyph sits on "19" while `gapPx` reads +2.6px. Measure from the start of the line instead:

```ts
const verse = para.querySelector('.verse');
const lineStartPx = Math.min(textRect.left, verse?.getBoundingClientRect().left ?? Infinity);
```

### 2. spec:209 · Low
The revert-test row in the PR description doesn't match the current assertion order.

The `gapPx` expect runs before the `--para-indent` expect. With main's stylesheet at the opened width (about 638px), the run should stop on "the \qm2 glyph overlaps its paragraph text by ~31px". The reported "the qm2 paragraph has no --para-indent" only comes first at about 330px or narrower. Please rerun against this commit and quote the full message, including its width.

### 3. spec:137 · Low
Nothing checks that the glyph is visible, so a hidden glyph makes every overlap assertion pass.

A `display: none` glyph has an all-zero rect, so `gapPx` is large and positive. The glyph is shown only by the `:first-child` rule (`_usj-nodes.scss:2811`), and the waits at 116–122 target the paragraph, not the glyph. Assert `glyphRect.width > 0` and `glyphText === \`\\${marker}\``.

### 4. e2e-tests/fixtures/onboarding-tour.page.ts:169 · Low
This doc says scripture-editor specs must call `suppressOnboardingTour`, but the two existing Simple-mode siblings don't.

`formatted-default-simple-mode.spec.ts` and `paragraph-style-trigger-column-floor.spec.ts` click in the main frame without it. Either add the call to both, or reword to "A Simple-mode spec that skips `waitForAppReady` must call this…" and ticket a fixture-level suppression in `isolated.fixture.ts` (opt-out for `onboarding-tour.spec.ts`). The same claim is in the commit message and PR description.

### 5. spec:38 · Low
This spec doesn't seed `platform.firstRunComplete`, so whether it gets past the first-run wizard depends on the developer's dev-appdata and Paratext registration.

On an unregistered machine it times out at the `navigateToolbarBcv` click. The other Simple-mode isolated specs seed it, e.g. `bcv-open-resource-books-simple.spec.ts:135` and `onboarding-tour.spec.ts:108`. Add `seedSettings: { 'platform.firstRunComplete': true }`.

### 6. e2e-tests/fixtures/onboarding-tour.page.ts:165 · Low
Clicking Skip works around a real keyboard-accessibility gap in the product that deserves a ticket.

Once a web view loads, it takes DOM focus. The tour listens for Escape on the host `window` (`tour.component.tsx:419`) and traps Tab on the host `document` (457), and key events inside the iframe aren't forwarded. So a keyboard user can't Esc out of the tour, and Tab moves behind the modal. Please file a ticket and list it under known issues.

### 7. spec:217 · Low
The custom-property checks are worth keeping, but the `vw` regexes are brittle.

`/^[1-9][0-9.]*vw$/` rejects `0.5vw` and `calc(…)` and reports them as "has no --para-indent", and `/^-[0-9.]+vw$/` accepts `-0vw`. The values don't change with width, so check them once, as a non-zero length with the right sign.

### 8. e2e-tests/fixtures/onboarding-tour.page.ts:28 · Low
Moving this constant left a stale pointer behind.

`src/renderer/components/onboarding-tour/onboarding-tour.store.ts:12` still says the key is kept in sync with `e2e-tests/fixtures/helpers.ts`. Update it to point at `onboarding-tour.page.ts`; that file isn't in the diff.

### 9. spec:173 · Low
The divider drag, the column-width read and the 310px ceiling are near-copies from `paragraph-style-trigger-column-floor.spec.ts` (lines 79–101, 103–106, 43–48).

Move them into `fixtures/scripture-editor-helpers.ts`: a drag helper that takes a distance, a width reader and an exported ceiling constant. Use them from both specs.

### 10. spec:231 · Low
This comment gets the termination logic backwards.

A drag that never takes ends the sweep at once (`nextWidth >= previousWidth`) and fails the floor assertion; `maxSteps` only bounds a column that never clamps. Suggested: "Shrink until the width stops shrinking (the dock clamps at the floor). `maxSteps` bounds a column that never clamps; a drag that never takes ends the sweep early and fails the floor assertion below."

### 11. spec:18 · Low
The docblock mixes bug history with an incorrect description of the indents.

"markers of the original report" and "had no compensation at all" are history. 15vw and 20vw are margins; the hanging indents are −7.5vw (li1, li2, q2), −10vw (qm2) and −15vw (iq1). Suggested replacement for 16–20: "The sample WEB project lacks most of these markers, so the spec rewrites Obadiah 1 through PAPI before opening it. Together they cover margins of 10–20vw and hanging indents from −7.5vw to −15vw."

### 12. e2e-tests/fixtures/onboarding-tour.page.ts:179 · Nit
Tour dismissal is now English-only and untimed, so a miss waits out the whole test timeout instead of 5 seconds.

Neither Playwright config sets `actionTimeout`, and an attach-mode app in another language won't show "Skip tour". Either give the click a short `timeout`, or dispatch the Escape the tour listens for on `window` (retry with `expect(…).toPass({ timeout: 5000 })`):

```ts
await page.evaluate(() => window.dispatchEvent(new KeyboardEvent('keydown', { key: 'Escape' })));
```

### 13. e2e-tests/fixtures/scripture-editor-helpers.ts:128 · Nit
"Writes go to the isolated project root only" reads as a guarantee, but nothing checks it.

Reword it as a precondition: "Call only from specs launched with `isolatedProjectRoot: true`; this helper does not check." Without that option it would rewrite the developer's WEB project, and the transform adds another empty `\q2`/`\iq1` on every run.

### 14. spec:92 · Nit
If the sample text's layout changes, the rewrite silently matches nothing, and the failure appears 60s later as a locator timeout.

Throw from the transform when fewer than three `\p` lines were replaced or either insertion didn't match.

### 15. spec:47 · Nit
`HANGING_MARKERS` is `new Set(MARKERS)`, so the `if` at 218 is always true.

Delete the set and the `if`, and note on `MARKERS` that all five are hanging-indent paragraphs.

### 16. e2e-tests/fixtures/onboarding-tour.page.ts:47 · Nit
`escapeForRegExp` and `exactName` can be replaced by `getByRole('button', { name: label, exact: true })`, with `.or()` for Next/Done.

The labels come verbatim from `en.json`, so case-insensitive matching isn't needed.

### 17. e2e-tests/tests/isolated/onboarding-tour.spec.ts:64 · Nit
The page object now reads labels from `en.json`, but the tour spec still hard-codes `/^Next$/i`, `/^Back$/i` and `/^Done$/i`.

After a relabel, the `not.toBeVisible()` checks at 154, 164 and 288 would pass without finding anything. Use `englishLabel` there too.

### 18. PR title / description · Nit
- The title is 73 characters; CLAUDE.md asks for under 70.
- Refresh the revert-test row (see 2).
- Link tickets for the two "Known issues".
- In "Out of scope", replace "the tier-list question to Todd and Tom" with "pending a product decision".

> @katherinejensen00 reviewed 6 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2808#-:-P1YeD_j9G9nfE4yX5m7:b-mk7ydb)* status: all files reviewed, 1 unresolved discussion (waiting on irahopkinson).

___
*[`e2e-tests/tests/isolated/onboarding-tour.spec.ts` line 144 at r1](https://reviewable.io/reviews/paranext/paranext-core/2808#-P1YePTn76y9ZvOaPmM9:-P1YePTn76y9ZvOaPmMA:b-4xdok1) ([raw file](https://github.com/paranext/paranext-core/blob/3552750d3414fc8da83707d5d1c952e545b5a31d/e2e-tests/tests/isolated/onboarding-tour.spec.ts#L144)):*
> ```XML
>   test('advances through all steps with Next and shows Done on the last step', async ({
>     mainPage,
>   }) => {
> ```

### 1. spec:144 · Medium
The overlap check measures from the text after the verse number, so a glyph painted over the verse number passes, including the qm2 regression this PR fixes.

Every tested paragraph starts with an inline verse number (`span.verse`), and the walker skips `.verse`, so the check can't see overlaps up to the number's width (about 15–22px). In a headless render with main's stylesheet at the 297px floor, the `\qm2` glyph sits on "19" while `gapPx` reads +2.6px. Measure from the start of the line instead:

```ts
const verse = para.querySelector('.verse');
const lineStartPx = Math.min(textRect.left, verse?.getBoundingClientRect().left ?? Infinity);
```

irahopkinson (commented):
Pushed `9817a1c1977` taking all eighteen. Spec passes first attempt on the current stylesheet (3.0m); with the pre-#2807 stylesheet swapped in it fails on both attempts with "At 638px the \\qm2 glyph overlaps its paragraph text by 55.8px".

**Spec**

- **#1** Gap measured from the line start (`min` of the text rect and the `.verse` rect). Replied in the thread with the revert result.
- **#2** Revert row rerun and quoted in full, width included.
- **#3** Asserts a non-zero glyph width and that the glyph text is the backslash-prefixed marker, before the gap.
- **#5** `seedSettings: { 'platform.firstRunComplete': true }`.
- **#7** The two custom properties are checked once, after the opened-width measurement, as signed non-zero lengths (`isLengthWithSign`), not per width and not with the `vw` regexes.
- **#9** `dragEditorColumnDividerLeft`, `getEditorColumnWidth` and `COLUMN_FLOOR_CEILING_PX` live in `scripture-editor-helpers.ts`; both column-floor specs use them. The helper takes a distance, so the sibling passes `Number.POSITIVE_INFINITY` for its drag past the floor.
- **#10** Your wording.
- **#11** Your replacement paragraph; the history sentences are gone.
- **#14** The transform throws when fewer than three `\\p` lines were replaced or either insertion did not match.
- **#15** `HANGING_MARKERS` and the `if` removed; the `MARKERS` doc says all five are hanging-indent paragraphs.

**Page object and fixtures**

- **#4** Reworded to "a Simple-mode spec that skips `waitForAppReady` must call this", with a `TODO` asking for fixture-level suppression in `isolated.fixture.ts` (opt-out for the tour spec). No ticket by author decision. The siblings are left as they are.
- **#6** Filed [PT-4627](https://paratextstudio.atlassian.net/browse/PT-4627) and listed under known issues in the description.
- **#12** Suppression dispatches a synthetic Escape on the host `window` from `page.evaluate`, retried with `toPass({ timeout: 5000 })`, so it no longer depends on the English label or on where focus sits.
- **#13** Reworded as a precondition, and says what happens without `isolatedProjectRoot`.
- **#16** `escapeForRegExp` and `exactName` are gone; one `getTourButton(page, key)` uses `exact: true`, and `advanceTour` is `getTourNextButton(page).or(getTourDoneButton(page))`.
- **#17** The tour spec uses the new `getTourNextButton` / `getTourBackButton` / `getTourDoneButton` exports instead of the regexes.
- **#8** `onboarding-tour.store.ts` points at `onboarding-tour.page.ts`.

**#18** Title is 67 characters. Revert row refreshed. Both known issues now carry tickets: the interface-mode leak is [PT-4628](https://paratextstudio.atlassian.net/browse/PT-4628) (new, no existing ticket covered it), and the tour-spec flakes point at [PT-4621](https://paratextstudio.atlassian.net/browse/PT-4621), whose self-dismissal mechanism matches the symptom; noted there rather than filed twice. Out-of-scope wording now says the ticket wording is settled on the wider marker set pending the Epic Lead's confirmation.

Repo lint exit 0, typecheck exit 0, ESLint and Prettier clean on the six touched files.

> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2808#-:-P1bnvn5EMOzRXIKOQl0:b-6ma0zz)* status: 2 of 8 files reviewed, 1 unresolved discussion (waiting on katherinejensen00).

___
*[`e2e-tests/tests/isolated/onboarding-tour.spec.ts` line 144 at r1](https://reviewable.io/reviews/paranext/paranext-core/2808#-P1YePTn76y9ZvOaPmM9:-P1bUuxP4PBqceJ6ctSk:bkolb20) ([raw file](https://github.com/paranext/paranext-core/blob/4f7f76ec3c3393f12e35d0b01cb076e34b8de6de/e2e-tests/tests/isolated/onboarding-tour.spec.ts#L144)):*

### 1. spec:144 · Medium
The overlap check measures from the text after the verse number, so a glyph painted over the verse number passes, including the qm2 regression this PR fixes.

Every tested paragraph starts with an inline verse number (`span.verse`), and the walker skips `.verse`, so the check can't see overlaps up to the number's width (about 15–22px). In a headless render with main's stylesheet at the 297px floor, the `\qm2` glyph sits on "19" while `gapPx` reads +2.6px. Measure from the start of the line instead:

```ts
const verse = para.querySelector('.verse');
const lineStartPx = Math.min(textRect.left, verse?.getBoundingClientRect().left ?? Infinity);
```
</blockquote></details>

Confirmed and fixed in `9817a1c1977`. The gap is now measured from the start of the first line, `Math.min` of the text rect and the `.verse` rect (mirrored for RTL). Rerunning the revert test with the pre-#2807 stylesheet, the spec now fails at the opened width on both attempts with "At 638px the \\qm2 glyph overlaps its paragraph text by 55.8px"; the same render measured from the text after the number read a positive gap, which is why the earlier revert row reported the `--para-indent` assertion instead. The PR description's revert row is updated.

katherinejensen00 (changes requested):
## New findings

### N1 · Warning · `e2e-tests/fixtures/onboarding-tour.page.ts:182` — merged: UX, Scope, Security

The synthetic Escape removes the suite's only real-keypress coverage of tour dismissal, and it
removes it for every spec.

`waitForAppReady` routes through `suppressOnboardingTour` by default (`helpers.ts:2356`), and at the
PR base that path was `page.keyboard.press('Escape')`. A synthetic `KeyboardEvent` dispatched at
`window` fires at-target on the tour's capture listener regardless of focus routing, so the tour's
Escape handler could break for real keyboard users — handler detached, listener moved off the capture
phase, the `open && isCardRendered` gate regressed — and every spec would still go green. The
escape hatch is only needed when a web view iframe holds focus.

Secondary: the dispatch reaches every window-level keydown listener in the renderer, capture and
bubble, re-dispatched for up to 5s — wider than the typed key it replaces, which went through the
focused element. Correctly gated on `getTourDialog(page).isVisible()` (page:179), so an unrelated
dialog is not Escape-dismissed.

Ask: try `page.keyboard.press('Escape')` first inside the `toPass` block and fall back to the
synthetic dispatch only if the dialog is still visible; and/or add one assertion in
`onboarding-tour.spec.ts` (no editor iframe there, host focus intact) that a typed Escape closes the
tour. Also worth a line in the PR description — it currently reads as relocation plus locator
changes, not a suite-wide change of dismissal mechanism.

### N2 · Warning · `e2e-tests/fixtures/scripture-editor-helpers.ts:135` and `:87` — Security

The write helpers still enforce nothing; round 1's #13 was answered with a docblock.

Verified in code rather than from the doc: the helper resolves its target purely from the hard-coded
`SAMPLE_WEB_PROJECT_ID` and talks to whatever app is on the hard-coded `WEBSOCKET_PORT = 8876`.
Nothing in the call path observes `isolatedProjectRoot`. `launchElectronApp` (helpers.ts:607) sets
`PLATFORM_BIBLE_PROJECT_ROOT_FOLDER` only when that option is passed; without it the app resolves its
real default root, and the collision is near-certain rather than hypothetical because
`SAMPLE_WEB_PROJECT_ID` is the fixed GUID from `c-sharp/assets/WEB/Settings.xml` — the same id the
backend installs into a dev project root. There is no restore of the original USFM.

`makeSampleProjectEditable` (sehelp:87) is unguarded the same way and has no precondition docblock at
all: it persists `platform.isEditable = true` into Settings.xml with no restore. Note the ordering —
the spec calls it at spec:94, *before* the rewrite, so the transform's own fail-fast throw cannot
prevent that mutation.

This spec passes `isolatedProjectRoot: true` correctly, so it is about the helpers' contract for the
next caller, not a live bug here.

Ask: make the precondition mechanical — have `launchElectronApp` record the isolated root it set and
assert it is non-empty before `setChapterUSFM`/`setSetting`, or assert the resolved root is under
`os.tmpdir()`. Whatever is chosen should cover both helpers.

### N3 · Warning · `e2e-tests/fixtures/onboarding-tour.page.ts:172` and `:162` — merged: Clarity, UX

Neither deferral carries a ticket in the code. `grep -rn PT-4627 e2e-tests src` returns nothing.

page:172 is a bare `TODO` with no tracked ticket. The repo's forward-facing-comments rule accepts a
TODO that names an *open* ticket precisely because the ID says when the note can be retired; an
anonymous TODO in a shared fixture has nothing to close it against. It also sits between the TSDoc
block and the `export async function`, severing the doc from the function.

page:162's docblock describes a genuine user-facing accessibility defect as if it were test
mechanics: once a web view iframe holds DOM focus a keyboard user cannot Escape out of the tour
(`window` listener) and Tab walks behind the `aria-modal` overlay (`document` trap). As written a
future maintainer reads "Playwright quirk, nothing to fix" and the product bug goes invisible.
Deferring the fix is defensible in an E2E-only PR — but the pointer belongs in the code, not only in
the PR description.

Ask: `TODO(PT-XXXX)` for the fixture-level suppression, and `TODO(PT-4627)` on the synthetic-Escape
docblock.

### N4 · Warning · `e2e-tests/tests/isolated/scripture-editor/gutter-marker-no-overlap.spec.ts:147-186` — merged: Scope, Tests, UX

The RTL branch is unreachable, and it is in the only code path the spec's verdict depends on.

`isRtl` (147), the mirrored `lineStart` (178-179) and the mirrored `gapPx` (186) never run: the spec
opens only the LTR sample WEB project and never sets or asserts a direction. RTL is explicitly out of
scope for this PR, so the file reads as though RTL is covered when it is not, and if anything ever
flipped the editor root to `dir="rtl"` the spec would silently switch to arithmetic nobody has run.
It is also the only place in the diff touching an explicitly out-of-scope item.

Ask: drop the mirroring and measure LTR only, or keep it and pin the assumption next to the existing
positive controls — `await expect(editorRoot).not.toHaveAttribute('dir', 'rtl')`, or assert
`getComputedStyle(root).direction === 'ltr'` inside `measure` — so the branch choice is asserted
rather than inferred.

> @katherinejensen00 reviewed 6 files and all commit messages, made 2 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2808#-:-P1epHpc6bnnLp-owJv7:b-c9awua)* status: all files reviewed, 1 unresolved discussion (waiting on irahopkinson).

___
*[`e2e-tests/fixtures/onboarding-tour.page.ts` line 182 at r2](https://reviewable.io/reviews/paranext/paranext-core/2808#-P1epPAu2We3hXAVNwfM:-P1epPAu2We3hXAVNwfN:bsn0nla) ([raw file](https://github.com/paranext/paranext-core/blob/9817a1c1977f09c66f02db580ecc4861ced2e791/e2e-tests/fixtures/onboarding-tour.page.ts#L182)):*
> ```XML
>   await expect(async () => {
>     await page.evaluate(() => {
>       window.dispatchEvent(new KeyboardEvent('keydown', { key: 'Escape' }));
> ```

### N1 · Warning · `e2e-tests/fixtures/onboarding-tour.page.ts:182` — merged: UX, Scope, Security

The synthetic Escape removes the suite's only real-keypress coverage of tour dismissal, and it
removes it for every spec.

`waitForAppReady` routes through `suppressOnboardingTour` by default (`helpers.ts:2356`), and at the
PR base that path was `page.keyboard.press('Escape')`. A synthetic `KeyboardEvent` dispatched at
`window` fires at-target on the tour's capture listener regardless of focus routing, so the tour's
Escape handler could break for real keyboard users — handler detached, listener moved off the capture
phase, the `open && isCardRendered` gate regressed — and every spec would still go green. The
escape hatch is only needed when a web view iframe holds focus.

Secondary: the dispatch reaches every window-level keydown listener in the renderer, capture and
bubble, re-dispatched for up to 5s — wider than the typed key it replaces, which went through the
focused element. Correctly gated on `getTourDialog(page).isVisible()` (page:179), so an unrelated
dialog is not Escape-dismissed.

Ask: try `page.keyboard.press('Escape')` first inside the `toPass` block and fall back to the
synthetic dispatch only if the dialog is still visible; and/or add one assertion in
`onboarding-tour.spec.ts` (no editor iframe there, host focus intact) that a typed Escape closes the
tour. Also worth a line in the PR description — it currently reads as relocation plus locator
changes, not a suite-wide change of dismissal mechanism.

irahopkinson (commented):
Committed locally as `8155ef64cf7`, on the branch rebased to `main`. Spec passes first attempt with the changes (1.4m).

- **N1** Typed Escape first; synthetic dispatch only when the dialog survives *and* focus is inside an iframe; a failed typed Escape with host focus throws. Replied in the thread, including the correction that the tour spec's own typed-Escape test never went away.
- **N2** Mechanical. `launchElectronApp` records the project root it set, and `requireIsolatedProjectRoot()` in `helpers.ts` throws unless the most recent launch used `isolatedProjectRoot: true` with a root under `os.tmpdir()`. Both `makeSampleProjectEditable` and `rewriteSampleProjectChapterUsfm` call it before their first write, which also covers the ordering you noted (the setting write happens before the transform can throw). All nine current callers of those helpers launch isolated.
- **N3** `TODO(PT-4627)` on the synthetic-dispatch paragraph and `TODO(PT-4639)` on the fixture-level suppression, both inside the docblock so the doc stays attached to the function. [PT-4639](https://paratextstudio.atlassian.net/browse/PT-4639) is new; a search for an existing ticket found none.
- **N4** Kept the mirroring for the RTL follow-up named in "Out of scope", and pinned the assumption as a positive control: `await expect(editorRoot).toHaveCSS('direction', 'ltr')` next to the two class checks, with a comment saying why.

ESLint and Prettier clean on the four touched files; `npm run typecheck:e2e` exit 0.

> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2808#-:-P1h01g5EpgKAKqrghVa:bzam2lf)* status: all files reviewed, 1 unresolved discussion (waiting on katherinejensen00).

___
*[`e2e-tests/fixtures/onboarding-tour.page.ts` line 182 at r2](https://reviewable.io/reviews/paranext/paranext-core/2808#-P1epPAu2We3hXAVNwfM:-P1h0-qaCnSEWQaaNVH0:b-2uddko) ([raw file](https://github.com/paranext/paranext-core/blob/9817a1c1977f09c66f02db580ecc4861ced2e791/e2e-tests/fixtures/onboarding-tour.page.ts#L182)):*

### N1 · Warning · `e2e-tests/fixtures/onboarding-tour.page.ts:182` — merged: UX, Scope, Security

The synthetic Escape removes the suite's only real-keypress coverage of tour dismissal, and it
removes it for every spec.

`waitForAppReady` routes through `suppressOnboardingTour` by default (`helpers.ts:2356`), and at the
PR base that path was `page.keyboard.press('Escape')`. A synthetic `KeyboardEvent` dispatched at
`window` fires at-target on the tour's capture listener regardless of focus routing, so the tour's
Escape handler could break for real keyboard users — handler detached, listener moved off the capture
phase, the `open && isCardRendered` gate regressed — and every spec would still go green. The
escape hatch is only needed when a web view iframe holds focus.

Secondary: the dispatch reaches every window-level keydown listener in the renderer, capture and
bubble, re-dispatched for up to 5s — wider than the typed key it replaces, which went through the
focused element. Correctly gated on `getTourDialog(page).isVisible()` (page:179), so an unrelated
dialog is not Escape-dismissed.

Ask: try `page.keyboard.press('Escape')` first inside the `toPass` block and fall back to the
synthetic dispatch only if the dialog is still visible; and/or add one assertion in
`onboarding-tour.spec.ts` (no editor iframe there, host focus intact) that a typed Escape closes the
tour. Also worth a line in the PR description — it currently reads as relocation plus locator
changes, not a suite-wide change of dismissal mechanism.
</blockquote></details>

One correction to the premise: the suite did keep real-keypress coverage. `onboarding-tour.spec.ts` has "Escape key closes the tour", which types Escape through `page.keyboard` with no editor open and asserts the dialog closes, so a detached handler or a listener moved off the capture phase fails there. What was true is that `waitForAppReady`'s suppression path went from a typed key to a synthetic one for every spec.

Taken, and tightened past the ask, committed locally as `8155ef64cf7`. Inside the retry the helper types Escape first. If the dialog survives, it checks where focus is: only when `document.activeElement` is an iframe does it dispatch the synthetic `keydown` on `window`, because that is the one case the typed key cannot reach the listener (PT-4627). If the typed key failed with focus anywhere else it throws, so a broken Escape handler fails every spec that suppresses the tour with host focus rather than being hidden by the fallback. The dispatch is also gated on the dialog still being visible, as before. The PR description now describes the mechanism change and its scope.

katherinejensen00 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2808#-:-P1jv6g08A7FL8RosrZf:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2803, opened 2026-09-10
PT-4576: Platform core for per-pane content zoom

rolfheij-sil:
# Code Review Summary

**Branch**: pt-4576-content-zoom-platform-core

**Base**: origin/main

**Date**: 2026-09-11

**Review model**: Claude Fable 5.1

**Files changed**: 27

## Overview

PT-4576 is the platform half of per-pane content zoom (work item 1 of epic PT-4575). A web view marks one or more zoom areas with `data-platform-content-zoom-root`; the platform keeps a level per area in the pane's saved definition state, remembers it per project and kind of pane in a hidden local setting, pushes the effective factors into the pane as CSS custom properties, shows an on-area indicator, follows the Settings default for areas without their own level, keeps sibling panes of the same project in line, and scales unadapted and URL views whole at the default. Input arrives through a bootstrap injected into every non-URL web view (Ctrl/⌘ chords and Ctrl/⌘+wheel, targeted at the focused, pointed-at or last active area) and through three owner-routed PAPI commands. Nothing is user-visible on its own yet: no first-party view marks an area in this PR, main still claims the Ctrl chords on Windows/Linux until PT-4577, and the Settings control and tab-menu entries come with later work items.

The four analysis passes found no critical issues, seven important ones and sixteen minor ones. The author ruled on every item in one pass: six important findings were fixed in review (API typing and reachability of `ContentZoomAreaId`, the copy-the-literal note on the contract constants, the Settings description, two missing failable tests, the hand-rolled debounce), one was dismissed as a product decision ("Zoom" beside "Zoom factor"), and the minors were fixed where cheap and dismissed with a reason where not. PRD 0-006 arrived during the review; it does not change the platform's scope, and three doc comments plus one TSDoc sentence were adjusted so the platform describes the capability without prescribing a view's shape.

## API Changes

All additive; no existing export removed or changed. `lib/platform-bible-react/`, `lib/platform-bible-utils/` and extension `.d.ts` contracts untouched.

- `papi-shared-types` CommandHandlers: added `platform.webViewContentZoomIn/Out/Reset` `(webViewId?: WebViewId, areaId?: ContentZoomAreaId) => Promise<void>`, `@experimental`; JSDoc of `platform.zoomIn`/`zoomOut` reworded (signatures unchanged).
[...]

PR #2809, opened 2026-09-14
PT-4577: Move Ctrl +/-/0 to content zoom; explicit macOS View menu

rolfheij-sil:
> **Stacked on #2803** (`pt-4576-content-zoom-platform-core`). Review only the commits above that branch (restacked onto its rewritten tip af53294). Stacked PRs get no automatic CI here; the full local battery (typecheck, lint, format:check, `vitest run src`: 263 files / 3781 tests) is green. After #2803 squash-merges this branch is rebased with `git rebase --onto main <old parent tip>`.

## Summary

Work item 2 of the per-pane content-zoom epic (PT-4575). Ctrl/⌘ `+` `-` `0` stop changing the whole app's scale and become the content-zoom chords everywhere:

- **Windows/Linux:** the main process no longer swallows the chords in `before-input-event`, so the in-view bootstrap from #2803 finally receives them. `platform.zoomIn` / `platform.zoomOut` (app-wide scaling) stay as commands with no default chord, and their step now goes through `adjustZoomFactor`, fixing the float drift that rejected the 20th step (2.9000000000000004 + 0.1 > 3.0).
- **macOS:** the View menu is explicit. Zoom in / Zoom out / Reset zoom items carry ⌘= / ⌘- / ⌘0 (plus hidden numpad duplicates) and invoke the content-zoom commands; the native `zoomIn`/`zoomOut`/`resetZoom` roles that claimed the chords are gone.
- **Window chrome:** a renderer top-document `keydown` listener routes the chords to the active tab's active area when keyboard focus is on the tab bar, reference box or a toolbar button, where no iframe sees the key. It ignores keystrokes inside iframes and does nothing while a dialog is open (design §5: dialogs are out of scope). "Dialog open" means a live docked PAPI dialog or a modal overlay, via a new `isAnyDialogOpen()` query.
- **Catalog:** the app-wide `zoom-in/zoom-out/reset-zoom` entries are removed; the `content-zoom-*` entries document the three handlers and a new data test checks every cited location exists on disk.

Nothing zooms yet: no view marks a zoom area until PT-4580 (`ContentZoomRoot`) and PT-4581 (editor) land. See merge sequencing below.

### Why review this

Part of the current epic (PT-4575 per-pane content zoom). PT-4576 shipped with `TODO(PT-4577)` markers and API docs that said the Ctrl chords drive app-wide zoom "until PT-4577"; this is that item, and it rewrites those docs to the settled state (app-wide zoom commands have no default chord; the chords belong to content zoom). A parity test pins the chord rule's two deliberate copies (renderer listener and injected bootstrap) to each other.

## Decisions worth a look

- **Shift.** The chrome listener accepts Shift only for the zoom-in keys (Ctrl+Shift+= is how US keyboards type Ctrl++); Ctrl+Shift+`-` and Ctrl+Shift+`0` stay free. This is the same rule as #2803's in-view bootstrap.
- **Dialog gate in two places on purpose.** `resolveContentZoomTarget` returns nothing while a dialog is open (design §3.2), so every input path is covered, including the macOS menu accelerators. The window-chrome listener checks the same query before acting; that check is what leaves the keystroke to the dialog instead of calling `preventDefault` on it. The service receives the query as an injected dependency because importing it directly would cycle through the dialog and web-view shards.
- **macOS runtime verification is handed to the epic's integration item (WI-10).** No Mac here. Things to check there: the accelerators reach the commands, the numpad duplicates work, and no double-fire between the native menu and the DOM listener (Electron's menu consumes the key equivalent before the web contents see it, so none is expected). Also note the targeting difference on macOS: the bootstrap never sees the chord, so the menu path resolves the pane's *active* area rather than the area with the caret; the bootstrap's `focusin` tracking keeps those equal in practice.
- **Enhanced Resources' private zoom handler becomes reachable on Windows/Linux.** `enhanced-resource.web-view.tsx` has its own Ctrl+`+`/`-`/`0` handler that was dead only because main consumed the chords first (design §5.4, GAP-012). Its Ctrl+0 resets to 100 %, not the Settings default. Removal is scheduled in PT-4583; until then ER answers the chords with its own zoom.
- **Merge sequencing.** Merged to `main` on its own this removes the app-wide Ctrl chords and replaces them with nothing visible until a view opts in (PT-4580 + PT-4581). Decision: merge as soon as it is reviewed and accept that gap; app-wide scaling stays reachable through Settings meanwhile.
- **Pre-existing, not fixed here:** (a) `MenuItemConstructorOptionsWithOrder.submenu` is typed `Omit<…> & {…}[]`, which parses as object-∧-array, so `id`/`role`/`accelerator`/`click` are unreadable off a submenu item without guards (the new menubar test works around it); the one-line fix cascades into the `Localized<>` mapped type. (b) `translatePlatformMenuItemsAndCombine` replaces a menu's `submenu` wholesale when a contributed column matches its label/role, mutating the shared `macosMenubarObject`; now that View has a real submenu, a column labelled `%mainMenu_view%` would permanently drop the zoom accelerators. No such contributor exists today.

## Testing

- New: `platform-macos-menubar.data.test.ts` (11), `web-view-content-zoom.chrome-keys.test.ts` (19), `dialog-open.util.test.ts` (3), `keyboard-shortcuts.data.test.ts` (existence of every cited location), overlay-store / dialog-shard cases for the two new queries, resolver-gate cases in `web-view-content-zoom.service.test.ts`, and a float-drift regression in `content-zoom.util.test.ts`.
- `npm run typecheck && npm run lint && npm run format:check && npx vitest run src --maxWorkers=4` all green.
- Manual Windows/Linux check pending (no view has a zoom area yet, so the observable change is that Ctrl+`+` no longer rescales the app).

## Risk Level

Medium. Removes a long-standing app-wide shortcut on purpose; the replacement is inert until later work items land (see merge sequencing).

AI-assisted — [session](https://claude.ai/code/session_01Gkwyu1GZbkdGdPXaygcerV). Implemented with per-task Sonnet agents under TDD, each task reviewed, plus a whole-branch review; roborev clean on every commit.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2809)


rolfheij-sil:

PR #2810, opened 2026-09-14
PT-4536: Add a real hover delay to the paragraph-marker tooltip

tombogle:
# Code Review Summary

**Branch**: pt-4536--paragraph-marker-tooltip-hover

**Base**: origin/main

**Date**: 2026-09-14

**Files changed**: 11

## Overview

This change fixes PT-4536: the paragraph-marker gutter tooltip flashed on every quick mouse pass
because `<Tooltip open={!!hoveredData} onOpenChange={() => {}}>` forced the tooltip's open/closed
state directly, bypassing Radix's own `delayDuration` timing entirely (that machinery only runs
when Radix manages its own pointer-enter/leave detection, which this component's custom DOM
hit-testing never lets it do). A hand-rolled `hoverTimerRef` now gates when `hoveredData` actually
gets set: a reveal is delayed by `TOOLTIP_DELAY` (300ms), a fast pass clears the pending timer and
shows nothing, and moving directly from one marker to an adjacent one while a tooltip is already
showing updates instantly with no re-paid delay (Radix's own "tooltip group" grace period,
reimplemented by hand since Radix's timer isn't reachable here).

As part of this fix, `TOOLTIP_DELAY` — previously duplicated as a local `const = 300` in both
`platform-bible-toolbar.tsx` and `user-profile-popover.component.tsx` — was promoted to a new
shared, exported constant in `platform-bible-react` (`tooltip-delay.ts`, mirroring the existing
`Z_INDEX_OVERLAY` precedent in `z-index.ts`), since extensions can't reach app-side renderer
constants directly. Both renderer call sites now import the shared constant instead.

## API Changes

- `lib/platform-bible-react`: Added new export `TOOLTIP_DELAY` (`= 300`) from new module
  `src/components/tooltip-delay.ts`, re-exported from `src/index.ts`. Purely additive.
- `lib/platform-bible-react/dist/*`: Rebuilt bundle artifacts (`index.js`, `index.cjs`,
  `index.d.ts`, and sourcemaps) reflecting the new export — consistent with a real `npm run
  build:pbr`, not a hand edit.
- `papi.d.ts`: No changes.
- `extensions/src/*/src/types/*.d.ts`: No changes.
- `lib/platform-bible-utils`: No changes.
- Two internal (non-public-API) call sites, `src/renderer/components/platform-bible-toolbar.tsx`
  and `src/renderer/components/user-profile-popover/user-profile-popover.component.tsx`, replaced
  their own local `const TOOLTIP_DELAY = 300` with the new shared export.

None.

None.

### Minor — Consider

- [x] **Duplicated timer-clearing logic**: the three-line "clear the pending timer" block was
  repeated inline in `handleKeyDown` and in the effect's cleanup, instead of calling the
  already-extracted `clearHoverTimer()` callback that was in scope at both sites. _(fixed during
  review: both inline blocks replaced with `clearHoverTimer()` calls; `clearHoverTimer` added to
  the effect's dependency array since it's now referenced there. Re-verified with typecheck, lint,
  and the full test suite — all pass.)_
- [ ] ~~Pattern similarity to `useDelayedFlag`~~ _(author acknowledged as informational only; no
  action taken. `src/renderer/hooks/use-delayed-flag.hook.ts` is structurally similar but lives
  behind a process boundary extensions can't cross, and its boolean `active` shape doesn't cleanly
  cover this component's "skip delay if already showing" case — noted as a future consolidation
  candidate only if this delay-reveal pattern is needed a third time.)_

Author addressed finding #1 directly ("Address minor finding #1 only") and explicitly declined
further action on finding #2.

#### Shared Regions Modified

None.

#### Extension Config Changes

None.

## Positive Observations

- `TOOLTIP_DELAY` is a well-motivated, purely-additive export: it eliminates two independent
  hardcoded `300` duplicates that a comment in one of them explicitly called out as needing to stay
  in sync, replacing a comment-enforced invariant with a real shared constant.
- The new export is documented with clear rationale (why it lives in `platform-bible-react` rather
  than a renderer-only constant) and follows the existing `z-index.ts` precedent for exposing
  app-wide constants across the extension sandbox boundary.
- `dist/` artifacts were correctly rebuilt and committed alongside the `src/index.ts` change —
  avoids the common "source changed but dist not rebuilt" skew for this package.
- The hover-delay logic correctly implements "no flash on fast pass": a pending-reveal timer is
  armed only when nothing is currently shown, cleared on every leave/re-target path, and the
  adjacent-marker case (tooltip already visible) intentionally skips the delay.
- Test coverage is thorough: `paragraph-marker-tooltip-overlay.component.test.tsx` covers the
  delay-elapses case, the short-hover-cancels-timer case (with an explicit `vi.getTimerCount()`
  assertion so the test can't pass vacuously against a no-timer implementation), the instant
  adjacent-marker switch, timer cleanup on unmount, and timer cleanup when `enabled` flips to
  `false` mid-hover.
- The hidden/inactive-view case is explicitly addressed with a comment at the sync site (a
  `display:none` pane can't receive real `mouseover` events, so no new timer can start while
  hidden), satisfying the documented expectation for cross-view sync code even though this
  particular interaction isn't really cross-view sync.
- The component includes forward-facing rationale comments (e.g., explaining why Radix's own
  tooltip-group delay can't be used because `delayDuration` never applies to a caller-forced `open`
  prop).
- The fix directly targets the stated purpose without changing any user-facing text or the
  tooltip's positioning/animation behavior, and uses the correct component (`Tooltip`, not
  `Popover`) for a non-interactive hover hint.

## Interview Notes

- **Stated purpose**: Prevent the paragraph-marker gutter tooltip from flashing when the mouse
  passes over quickly.
- All four analysis passes (API/correctness, style/patterns, compliance/coverage, UX) returned zero
  Critical and zero Important findings.
- Two Minor findings were surfaced by the style/patterns pass. The author asked to address finding
  #1 (duplicated timer-clear logic) only, and explicitly declined action on finding #2 (informational
  note about `useDelayedFlag` pattern similarity) — this was a deliberate choice, not an
  oversight.
- No areas of uncertainty or AI-deferral were expressed by the author during this interview; the
  requested fix was specific and well-scoped.
- Context from the broader session (not part of the diff, but relevant for the reviewer): this
  branch was rebased onto an unrelated, large upstream infrastructure change (#2745, migrating the
  editor package from a yalc-linked registry pin to staged `file:` dependencies) partway through
  development. That rebase required a Node.js upgrade (>=22.18.0) and a one-time fix to a sibling
  `scripture-editors` checkout's git remote (the repo moved orgs) — both environment/tooling issues
  unrelated to this PR's own diff, already resolved, and not reflected in the file list above.

## In-Review Quality Check

All checks passed cleanly after the finding #1 fix — no additional fixes were required:

- **typecheck**: all 4 sub-checks (`core`, `erb`, `e2e`, `workspaces`) pass.
- **lint**: clean, both via the repo-root `npm run lint` and a direct `eslint` run scoped to the
  changed files from within the `extensions/` workspace.
- **format**: no changes needed; the changed files were already Prettier-clean.
- **npm test**: full suite passed — 4529 core tests passed (16 skipped, pre-existing/environment-gated),
  and the `platform-scripture-editor` workspace's full 1501-test suite passed, including the new
  `paragraph-marker-tooltip-overlay.component.test.tsx` (6/6).

## Suggested Review Focus

- [ ] Confirm the `TOOLTIP_DELAY` promotion to `platform-bible-react` is acceptable given this
  sprint's pbr merge-queue contention (PT-4530's Q10) — a Jira comment and coordination note were
  already raised by the author separately; worth a quick sanity check that the sequencing landed as
  intended.
- [ ] Spot-check the adjacent-marker "no re-paid delay" behavior manually in the running app (Simple
  mode) if not already done — the author reported doing this during development, but it's a
  perceptual/timing behavior worth a second set of eyes.

## Update (2026-09-17)

The summary above reflects the PR as it stood at round 1 of review (2026-09-14, `d629644`) and is
stale in a few ways that later commits and review rounds overtook:

- **File count / API naming**: the PR is now 17 files (7 generated `dist`), not 11. The exported
  constant shipped as **`TOOLTIP_DELAY_MS`**, not `TOOLTIP_DELAY` — renamed during round 1 review to
  match the repo's `_MS`-suffix convention for time constants. Three renderer call sites were
  de-duplicated onto it (`platform-bible-toolbar.tsx`, `user-profile-popover.component.tsx`, and
  `media-viewer.component.tsx`), not two.
- **Commits since round 1**: `d8102c8` (fixed two fade/flicker bugs found during manual
  verification — an empty-box exit animation, and a churn-driven pulsate-in-place near the
  tooltip's own border), `ce0e437` (hardened that fix's pointer-jitter threshold), `bb008541`
  (rewrote the hover/leave detection onto a `paraAtPoint` live hit-test — this removed the two
  guards the previous two commits had added, since the hit-test approach makes them unnecessary for
  the bug they targeted), a merge of `main`, and (this session) `83ce7b9` fixing a `prettier
  --check` failure on `tooltip-delay.ts`'s JSDoc that had been red on CI since round 2.
- **Round 3 review** (`45a284703` → current) raised three additional concerns after the PR had
  already been approved: the hand-rolled "grace period" restamping on every hop and potentially
  producing a flicker across a fast paragraph sweep; a gap in `TooltipContent`'s `pointer-events:
  none` coverage (Radix's popper wrapper only gets that style when `hideWhenDetached` is set, which
  it isn't here); and `suppressUntilMoveRef` (the keydown-triggered typing suppression) being
  cleared by any mousemove, including a hypothetical synthetic one from editor DOM churn during
  typing. **All three were manually tested in the running app and did not reproduce**: a fast sweep
  across paragraphs updates the tooltip's content instantly without any visible hide/show flicker;
  the layout condition needed to trigger the pointer-events gap (a paragraph forced to the very top
  of the viewport with no room above for the tooltip) isn't reachable given the app's persistent
  toolbar/menu chrome; and typing with the mouse resting over a paragraph's marker does not reopen
  the tooltip. Recorded here so a future reader doesn't have to re-derive this from the Reviewable
  thread.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2810)


katherinejensen00 (changes requested):
Great job, Tom! All of these are medium or below. Unless a blocker is introduced I think this PR will get through with only one round of code review

# PR #2810 review — PT-4536: Add a real hover delay to the paragraph-marker tooltip

- **PR**: https://github.com/paranext/paranext-core/pull/2810
- **Author**: Tom Bogle (`tombogle`)
- **Branch**: `pt-4536--paragraph-marker-tooltip-hover` → `main`
- **Merge base**: `014cdcd51d4`
- **Reviewed**: 2026-09-14
- **Diff**: 11 files, +535 / −297 (4 of them generated `lib/platform-bible-react/dist` bundles)
- **CI**: green. **Mergeable**: CONFLICTING (generated `dist` bundles only)

Passes run: senior-dev soundness, architecture, comments/documentation, an antagonistic pass over
those three, and `/code-review max` (10 angles, 5 verifier clusters, one gap sweep; per-subagent cap
raised to 16, final-report cap to 30). Read-only throughout — nothing modified, staged, committed,
pushed, or posted.

**File keys used below**

| Key | Path |
| --- | --- |
| **[A]** | `extensions/src/platform-scripture-editor/src/paragraph-marker-tooltip/paragraph-marker-tooltip-overlay.component.tsx` |
| **[B]** | `extensions/src/platform-scripture-editor/src/paragraph-marker-tooltip/paragraph-marker-tooltip-overlay.component.test.tsx` |
| **[C]** | `lib/platform-bible-react/src/components/tooltip-delay.ts` |
| **[D]** | `extensions/src/platform-scripture-editor/src/paragraph-marker-tooltip/paragraph-marker-tooltip-overlay.component.stories.tsx` |
| **[E]** | `extensions/src/platform-enhanced-resources/src/components/media-viewer/media-viewer.component.tsx` |

All line numbers are as the files exist at `pr-2810-head`.

---

## Part 1 — `/code-review max` final report

### Correctness — introduced or worsened by this PR

1. **[A]:86 — High.** Position is measured at hover-enter and frozen for 300 ms; `handleScroll`'s rAF bails on `if (!prev) return undefined`, so it neither re-measures nor cancels. Scroll during the delay reveals at pre-scroll coordinates.
2. **[A]:213 — Medium-High.** Effect cleanup clears the timer on `enabled` true→false but not `hoveredData`/`currentParaRef`; re-enable mounts `<Tooltip open>` already open at stale coords, and the leaked value makes the next hover skip the delay entirely.
3. **[A]:89 — Medium.** The "already showing ⇒ instant" fast path has no time bound (no `skipDelayDuration` equivalent). Paragraphs are flush (`_editor.scss:45-47`), so after one dwell, sweeping a chapter repaints marker names at full speed — the strobe the PR exists to remove.
4. **[A]:103 — Medium.** `handleMouseOver` branches on `hoveredData` from the render closure; a mouseover landing between timer-fire and commit sees `undefined` and re-arms 300 ms. Also rebuilds the callback on every scroll rAF frame.
5. **[A]:87 — Medium.** `lastPositionRef.current = pos` now runs at arm time, including for cancelled reveals, breaking its documented contract at lines 48-50; a paragraph entered during the ~150 ms exit animation can teleport the fading tooltip.
6. **[A]:96 — Medium.** Nothing re-validates the paragraph before revealing. A chapter change or remote edit inside the window produces no keydown and no mouseout, so the timer fires for detached content.
7. **[A]:42 — Medium-Low.** The hidden-view comment answers only "no *new* timer can start while hidden"; a timer armed while visible still fires under `display:none`.
8. **[A]:65 — Low (latent).** The timeout callback never resets `hoverTimerRef.current`, so it permanently reports "pending" after the first reveal.
9. **[A]:164 — Low (latent).** The only unmount-time `clearHoverTimer()` sits in an effect with two no-cleanup early returns.

### Test validity

10. **[B]:32 — Medium-High.** The mock returns `[{}]`, which the real hook cannot produce — `use-localized-strings-hook.ts:45-49` seeds `defaultState[key] = key`. Both test markers have real values in `localizedStrings.json`. All three `toHaveTextContent` assertions pin a fallback branch production never takes.
11. **[B]:139 — Medium.** "starts no timer while enabled is false" passes byte-identically against the merge base — with `enabled={false}` the JSX renders `onMouseOver={undefined}`, so it asserts React prop-gating, not the delay.
12. **[B]:70 — Medium.** Six confirmed gaps (deterministic grep: 0 `scroll`, 0 `keyDown`, 0 `mouseLeave`, 0 `blur`). Highest value is open→close→hover-again.
13. **[B]:128 — Low-Medium.** The unmount test's only assertions are timer counts — no user-visible assertion.
14. **[B]:79 — Low.** Hardcodes 299/300/1 and never imports `TOOLTIP_DELAY`.
15. **[B]:99 — Low.** `vi.getTimerCount()` is process-wide and asserted exactly; safe only because no test both opens a tooltip and counts.
16. **[B]:36 — Low.** The factory-form `vi.mock` *replaces* the aliased shared mock, silently nulling `useProjectSetting`/`useProjectDataProvider`.
17. **[B]:161 — Low.** The `rerender` re-types the fixture JSX instead of reusing `renderEditor`'s.
18. **[B]:26 — Low.** `globalThis.ResizeObserver` set in `beforeAll`, never restored; copy #27 of an identical stub.

### Shared constant

19. **[C]:2 — Medium.** "every tooltip has one consistent feel" is false on arrival: 51 of 66 `TooltipProvider` sites inherit `delayDuration = 0`; the constant governs 3 of 66.
20. **[C]:1 — Low-Medium.** API-surface TSDoc never says *which prop* to pass it to; its rationale cites files unresolvable from IntelliSense.
21. **[E]:133 — Low.** `<TooltipProvider delayDuration={300}>` — the one genuine missed migration repo-wide.

### Comment accuracy and cleanup

22. **[A]:91 — Medium-Low.** "Radix's timer never runs on a caller-forced `open`" is factually wrong; `character-marker-control.component.tsx:186,224` does controlled-with-Radix-delay in this same extension.
23. **[A]:217 — Low.** "re-run only when enabled or clearHoverTimer changes" — `clearHoverTimer` is `useCallback(…, [])` and can never change.
24. **[A]:120 — Low.** The clear-timer / clear-ref / hide triple is now written out at four sites.
25. **[D]:37 — Low.** The autodocs story still says "Hover over any line to see the tooltip" with no mention of the 300 ms dwell.
26. **[B]:40 — Low.** The `import/first` suppression is avoidable and its stated rationale is inaccurate (93 test files import at top vs 20 using the split shape).

### Radix integration — pre-existing, newly salient

27. **[A]:247 — Low.** The enter animation never fires: `data-state` is permanently `instant-open` and the `data-open` variant compiles to a selector Radix never emits.
28. **[A]:234 — Low.** `onOpenChange={() => {}}` makes Escape, pointer-down-outside, and cross-tooltip dismissal inert.
29. **[A]:232 — Low.** No `disableHoverableContent` and `sideOffset = 0`; when Radix flips to `bottom`, portaled content reaches the wrapper handlers.
30. **[A]:235 — Low.** `DestructiveKeyConfirmation` in `platform-bible-react` is the same skeleton and the sibling `two-step-delete-tooltip/` already consumes it; it also pairs an `sr-only role="status"` live region, which this overlay lacks.

**Verified clean.** No backward-facing comments; no `shadcn-ui/` edits; keyboard-shortcuts catalog
not triggered; `dist/` correctly rebuilt with the new export; `papi.d.ts` correctly unchanged; both
renderer call sites behaviourally identical; CI green.

---

## Part 2 — Consolidated review, addressed to the PR owner

Cross-checked across all five passes; everything that didn't survive verification was dropped (see
Part 3). Severity: **High** = fix before merge · **Medium** = should fix · **Low** = worth a look ·
**Nit** = optional.

### 1. Scroll during the 300 ms window reveals the tooltip in the wrong place

**Attach:** **[A]** line 86 · **Medium**

The position is measured when the hover starts and frozen until the timer fires, so anything that
moves the paragraph in between is missed.

`computePosition` runs at line 86 and is captured into `newData` at line 88; the timer at line 96
commits that stale value. Of the two terms in `clampTopToVisibleArea`, `topInContent` is
scroll-invariant but `visibleAreaTop = scrollContainerRect.top - anchorRect.top` is not — which is
exactly why `handleScroll` exists. During the pending window it can't help:
`setHoveredData((prev) => { if (!prev) return undefined; … })` at line 195 bails while `hoveredData`
is undefined, and it never cancels the timer either. The trigger is narrow (only when the clamp
binds, i.e. the paragraph straddling the viewport top, and only when the pointer stays on that same
paragraph so line 76's early return suppresses a recompute), and it self-heals on the next scroll —
but the fix is three lines: move the `computePosition` call and the `lastPositionRef` write inside
the timeout callback, reading `currentParaRef.current`. That also closes findings 5 and 6 of Part 1.

### 2. Flipping `enabled` off mid-hover leaks `hoveredData`, so the tooltip reopens on its own — and the next hover skips the delay

**Attach:** **[A]** line 213 (the cleanup block) · **Medium**

The effect cleanup clears the pending timer but not the state it was going to set, and nothing else
can clear it while disabled.

When `enabled` goes true→false, line 231's `{enabled && …}` unmounts the whole `TooltipProvider`
subtree and lines 223-228 strip every handler — so `hoveredData` and `currentParaRef` are frozen at
whatever they held. Flip back and `<Tooltip open={!!hoveredData}>` mounts already open, at a stale
position, with the pointer nowhere near it. The consumer is
`platform-scripture-editor.web-view.tsx:3721` (`enabled={!isPowerMode}`), and `isPowerMode` is live
setting-backed state, so an interface-mode toggle reaches this. The second half is the part that
bites this PR specifically: the leaked truthy `hoveredData` sends the *next* hover down the
`if (hoveredData)` branch at line 89, which skips the delay — the exact behaviour the PR is adding.
One-line fix: add `setHoveredData(undefined)` beside `clearHoverTimer()` at line 215. Note the
existing `enabled`-flip test only covers the case where a reveal was still *pending*, not one where
the tooltip was already showing.

### 3. The "already showing ⇒ instant" shortcut has no time bound, so the strobe comes back on a sweep

**Attach:** **[A]** line 89 · **Medium**

Once any tooltip has opened, every subsequent adjacent paragraph opens instantly with no ceiling —
so dragging down a chapter repaints marker names at full mouse speed.

The comment at lines 90-93 calls this "Radix's own tooltip group grace period", but Radix's
`skipDelayDuration` is a bounded window *after a tooltip closes*; this is an unbounded test on
`hoveredData` being currently truthy. Confirmed reachable, not theoretical: `_editor.scss:45-47`
gives `.editor-input > p` zero vertical margin, so consecutive paragraphs are flush and a downward
sweep never crosses non-paragraph territory to reset it. So after one 300 ms dwell anywhere, the
rest of the chapter is back to the pre-PR flashing behaviour. If a grace period is what you want, it
needs a "recently closed" timestamp rather than a truthiness check.

### 4. The comment explaining *why* the delay is hand-rolled states the wrong reason

**Attach:** **[A]** line 91 · **Medium-Low**

The conclusion is right but the stated mechanism isn't, and it's the one comment a future refactor
would rely on.

"Radix's timer never runs on a caller-forced `open`" isn't what's happening. In
`@radix-ui/react-tooltip@1.2.8` (`dist/index.mjs`), `handleDelayedOpen` is armed from the trigger's
`onPointerMove` at line 184 — and the trigger here carries `tw:pointer-events-none` (line 239), is
1px, and is `aria-hidden`, so it never receives a pointer event at all. That's the real reason. The
forced `open` plus no-op `onOpenChange` is a second, independent reason the result would be
swallowed. Worth correcting because the same extension does run controlled-open-with-Radix-delay
successfully at `character-marker-control.component.tsx:186,224`, so someone reading this comment
could reasonably "simplify" it into a tooltip that never opens. Related: the parenthetical
"(see the `Tooltip` below)" sends the reader to line 234, where no comment carries this fact —
putting the corrected explanation there instead would fix both.

### 5. The test mock returns a shape the real hook can never return

**Attach:** **[B]** line 32 · **Medium**

`useLocalizedStrings` is mocked as `[{}]`, but the real hook seeds every key to itself, so the three
text assertions pin a fallback branch that production never reaches.

`use-localized-strings-hook.ts:45-49` builds `defaultState[key] = key` and returns that as the
default, so the record is never empty. Both markers used in the fixture are in
`blockMarkerToBlockNames` with real values (`localizedStrings.json:25` → "Paragraph - Normal - First
Line Indent"). So `toHaveTextContent('\\p')` / `('\\q1')` assert the
`hoveredData?.marker ? '\\'+marker : ''` fallback at line 62, which a real user never sees. The delay
behaviour under test is genuinely independent of the wording — so the cleanest fix is to stop
asserting on text and assert on tooltip presence, or mock with the realistic key→key seed. This also
surfaces a pre-existing trap worth knowing about: `localizedDescription ?? …` can't fall back,
because the seeded raw key is truthy — so the tooltip can briefly render
`%paragraphMenu_p_markerDescription%`.

### 6. One new test passes identically against the merge base

**Attach:** **[B]** line 139 · **Medium**

"starts no timer while enabled is false" doesn't exercise the delay — it exercises React
prop-gating that predates this PR.

With `enabled={false}`, lines 223-228 render `onMouseOver={undefined}`, so `fireEvent.mouseOver`
reaches no handler and the timer count is 0 whether or not a delay exists. Verified green on
`014cdcd51d4`. Slightly ironic given the deliberate `vi.getTimerCount()` positive control at lines
97-98, which is the best thing in the test file. Either drop this test or make it assert something
the delay owns.

### 7. `handleMouseOver` reads `hoveredData` from the render closure

**Attach:** **[A]** line 103 · **Medium**

There's a short window after the timer fires but before React commits where a mouseover sees
`undefined` and re-arms a full 300 ms instead of switching instantly.

`setHoveredData` from the timeout is flushed in a scheduler task, so between fire and commit the
closure still holds the old value. A `hoveredDataRef` mirrored at render time removes the window and
lets the deps go back to `[clearHoverTimer]`, which also stops the callback being rebuilt on every
scroll rAF frame. For the record: the identity churn itself is harmless — it's a prop on one `<div>`,
and React delegates at the root. The race is the reason to change it, not perf.

### 8. Three coverage gaps, one of which is load-bearing

**Attach:** **[B]** line 70 · **Low-Medium**

Deleting `clearHoverTimer()` from the keydown handler leaves the suite green.

Deterministic grep of the new file: zero occurrences of `scroll`, `keyDown`, `mouseLeave`, `blur`.
The keydown clear at **[A]** line 184 is the one that matters — "no tooltip pops up while you are
typing" is the same failure class this PR exists to fix, and it's currently unpinned; that's about
five lines of test. Worth adding alongside: scroll-during-delay (pins finding 1), and
open→close→hover-again (pins finding 3, and is what distinguishes "grace period" from "delay
permanently skipped"). The `handleMouseLeave` and `handleBlur` clears are also uncovered, but both
are unreachable-by-design — `handleFocus` is an admitted no-op so focus can never arm a timer, and a
wrapper exit already routes through `handleMouseOut` — so leave those alone rather than write tests
for defensive redundancy.

### 9. The new constant's doc comment overclaims

**Attach:** **[C]** line 2 · **Medium**

"shared across the app and extensions so every tooltip has one consistent feel" describes 3 call
sites out of ~66; the app's actual default is instant.

`shadcn-ui/tooltip.tsx:17` still defaults `delayDuration = 0`, and 51 of 66 `TooltipProvider` sites
pass nothing and inherit it — including the provider in this very component at **[A]** line 232.
Five different delays are live in-tree (0, 150, 200, 300, 400). Two honest options: narrow the
sentence to what the constant actually does, or make `TooltipProvider` default to `TOOLTIP_DELAY`,
which is the change that would earn the sentence (and would need a `// CUSTOM:` annotation per the
shadcn rule). Worth saying out loud too: 300 ms is asserted as "standard" with no rationale — Radix's
own default is 700 ms and Windows' `SPI_GETMOUSEHOVERTIME` is 400 ms.

### 10. Rename to `TOOLTIP_DELAY_MS` before it ships

**Attach:** **[C]** line 7 · **Low**

This is the last free moment — it's a new public export of a versioned package that third-party
extensions consume.

The repo convention is overwhelming: 91 `_MS`-suffixed time constants against 3 without, and
`DEFAULT_DELAY_MS` already exists at `src/renderer/hooks/use-delayed-flag.hook.ts:8`. The value is
consumed both as a Radix `delayDuration` prop and as a raw `setTimeout` argument (**[A]** line 96),
where a wrong unit fails silently. It's already emitted as
`export declare const TOOLTIP_DELAY = 300;` at `lib/platform-bible-react/dist/index.d.ts:4188`, so
renaming after merge is an API break.

### 11. The TSDoc doesn't tell a consumer how to use it

**Attach:** **[C]** line 1 · **Low**

Three of its four lines answer "why is this file here?", which only this PR's reviewer is asking; it
never says which prop to pass it to.

This text is what extension authors see on hover, and the Code-Style-Guide's API-surface rule assumes
readers can't look up anything outside the comment. The `Security-Guide.md` and `z-index.ts` pointers
don't resolve from IntelliSense or the TypeDoc site. Suggested replacement:

```ts
/**
 * Standard delay in milliseconds before a hover reveals a tooltip. Pass to `TooltipProvider`'s
 * `delayDuration` so tooltips across the app and extensions share one consistent feel.
 */
```

### 12. Housekeeping — merge conflict, and one missed duplicate

**Attach:** PR-level · **Low**

The PR is currently `CONFLICTING` against `main`, in `dist/index.cjs`, `dist/index.cjs.map`,
`dist/index.js`, `dist/index.js.map` only — the usual generated-bundle conflict, resolved by rebasing
and re-running `build:pbr`. Separately, **[E]** line 133 still has a literal `delayDuration={300}`
and already imports `TooltipProvider` from `platform-bible-react` — the one genuine remaining copy of
the number this PR set out to de-duplicate.

### 13. Nits — batch into a single comment, or skip

**Attach:** various · **Nit**

- **[A]**:87 — `lastPositionRef` written at arm time even for cancelled reveals (fixed for free by finding 1).
- **[A]**:96 — timer callback doesn't null `hoverTimerRef.current`, so the ref can't be read as "pending" later.
- **[A]**:217 — the dep comment says "re-run only when enabled or `clearHoverTimer` changes", but `clearHoverTimer` is `useCallback(…, [])` and never changes; the rewrite also dropped the old "to (un)attach listeners" clause that carried the actual rationale.
- **[B]**:40 — the `// Imported after the mock` comment and its `eslint-disable import/first` are both unnecessary since vitest hoists `vi.mock`; moving the import up keeps the suite at 6/6.
- **[D]**:37 — the Storybook autodocs still say "Hover over any line to see the tooltip" with no mention of the new dwell, and that's the one artifact where a designer can judge the feel.

### 14. The `TOOLTIP_DELAY` promotion should get an `Architecture-Decisions.md` entry

**Attach:** `.context/standards/Architecture-Decisions.md` (new entry) · **Low**

This PR makes exactly the kind of shared-constant placement call the decision log exists to record,
and there's a near-identical entry already in it.

`CLAUDE.md` asks for a log entry when work "introduces a new pattern or top-level structure" or
"decides where a feature lives", while also warning against over-recording routine choices. This one
lands on the record-it side, and the strongest evidence is the precedent:
`adr-window-min-width-shared-constant` (line 3808, PT-4466) covers the same shape almost point for
point — a magic number duplicated across two sites, each carrying a "change both" comment, extracted
into a shared module, where the contested question was whether the new constant would become
extension-visible API. This case is the same story with the answer inverted: it *does* become public
API, which is precisely why it's worth writing down.

Notably, `z-index.ts` — the precedent the TSDoc cites — has no entry of its own; it only appears
*inside* other entries. So the log currently can't tell a future author why app-wide constants live
in `platform-bible-react`, which is the gap this entry would close. An entry is also the natural home
for the two questions the review raised but this PR shouldn't have to resolve in code: why 300 ms
rather than Radix's 700 or the OS's 400, and why `TooltipProvider`'s `delayDuration = 0` default was
deliberately left alone. That converts finding 9 from "the doc comment overclaims" into "the doc
comment is narrow and the reasoning is recorded where it belongs."

Slug sorts into byte order between `adr-toolbar-sync-status-is-local` (3538) and
`adr-web-view-error-boundary-placement` (3692) — inserted at that position, not appended.

This one is genuinely optional. If expanding the PR's scope isn't wanted, saying so explicitly in the
PR description is a reasonable substitute — the log's own guidance is that skipping routine choices
is fine, and reasonable people could call this one routine.

---

## Part 3 — Considered and dropped

The antagonistic pass killed these; recorded so it's clear they were considered rather than missed.

- **Comment density** (22.3% → 23.4%) — a metric with no threshold; the ratio is essentially held, and the rationale comments at **[A]** lines 108-110, 168-172 and 178-182 each carry a non-inferable fact.
- **The `.claude/rules/cross-view-sync-hidden-views.md` citation at [A]:42** — the rule explicitly *mandates* a comment at the sync site, so penalising compliance is backwards. (The comment is incomplete, which is finding 7 of Part 1 — that part stands.)
- **"Extract a shared `useHoverDelay` hook"** — `use-truncation-tooltip.hook.ts` has no timer and no delay, so it isn't the precedent it looks like; its 8 consumers deliberately open instantly; and `useDelayedFlag` was already declined on the record with a stated revisit trigger. Extracting an abstraction from one call site is the thing an architecture review should resist, not demand.
- **"Move the constant into `tooltip.tsx`"** — would trip the shadcn `// CUSTOM:` discipline and pollute every future upgrade diff; the sibling module is the right dodge.
- **"Derive 299/300/1 from `TOOLTIP_DELAY` in tests"** — hardcoding pins the contract; deriving would make the test mirror the implementation and pass at any value.
- **PR-description length** — there *is* a top-level `## API Changes` section naming the new export, so the API change isn't actually buried.

---

## Part 4 — Credit where due

- `dist/` was correctly rebuilt and committed with the new export — the usual trap in this repo, avoided.
- The `vi.getTimerCount()` positive control at **[B]** lines 97-98 is a genuinely good defence against a vacuous test; it's the model the rest of the file should follow.
- Zero backward-facing comments anywhere in the diff — no ticket IDs, review-finding IDs, stage tags, or dated dev notes in source or tests.
- The fix targets the stated purpose without touching user-facing text, positioning, or animation behaviour, and uses `Tooltip` rather than `Popover` for a non-interactive hover hint.

---

## Bottom line

Nothing here is a merge blocker on its own. Findings **1, 2, 3 and 5** are the ones worth resolving
before it lands — 2 and 3 because they partially undo the fix the PR is making, and 5 because the
test suite currently can't tell.

> @katherinejensen00 reviewed 11 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2810#-:-P1X88H-DsL_lnLd_gLj:bnujw1n)* status: all files reviewed, 1 unresolved discussion (waiting on tombogle).

___
*[`extensions/src/platform-scripture-editor/src/paragraph-marker-tooltip/paragraph-marker-tooltip-overlay.component.tsx` line 86 at r1](https://reviewable.io/reviews/paranext/paranext-core/2810#-P1X8cy97BuxsXZgkZFh:-P1X8cy97BuxsXZgkZFi:butj3mn) ([raw file](https://github.com/paranext/paranext-core/blob/d629644dc0414e95949a8a030646a6fd32026877/extensions/src/platform-scripture-editor/src/paragraph-marker-tooltip/paragraph-marker-tooltip-overlay.component.tsx#L86)):*
> ```XML
>         const marker = extractMarker(para.className);
>         if (marker) {
>           const pos = computePosition(para, anchor, scroller);
> ```

### 1. Scroll during the 300 ms window reveals the tooltip in the wrong place

**Attach:** **[A]** line 86 · **Medium**

The position is measured when the hover starts and frozen until the timer fires, so anything that
moves the paragraph in between is missed.

`computePosition` runs at line 86 and is captured into `newData` at line 88; the timer at line 96
commits that stale value. Of the two terms in `clampTopToVisibleArea`, `topInContent` is
scroll-invariant but `visibleAreaTop = scrollContainerRect.top - anchorRect.top` is not — which is
exactly why `handleScroll` exists. During the pending window it can't help:
`setHoveredData((prev) => { if (!prev) return undefined; … })` at line 195 bails while `hoveredData`
is undefined, and it never cancels the timer either. The trigger is narrow (only when the clamp
binds, i.e. the paragraph straddling the viewport top, and only when the pointer stays on that same
paragraph so line 76's early return suppresses a recompute), and it self-heals on the next scroll —
but the fix is three lines: move the `computePosition` call and the `lastPositionRef` write inside
the timeout callback, reading `currentParaRef.current`. That also closes findings 5 and 6 of Part 1.

tombogle:
https://github.com/user-attachments/assets/2bc87152-d379-4837-b84e-aafd23fc094a

tombogle:
Thanks for the thorough review — verified every finding against the code before fixing (two independent read-only passes), so this wasn't taken on faith.

## Fixed (the four flagged as worth resolving before merge, plus the comment-accuracy one)

1. **Stale position on scroll** (line 86) — position and paragraph validity are now (re)computed at *fire time* from the live refs, not captured at hover-enter. Also guards `.isConnected` so a chapter change/remote edit mid-delay can't reveal detached content (closes findings 5 and 6 of Part 1 too).
2. **`enabled`-flip leak** (line 213) — the effect cleanup now also clears `hoveredData`/`currentParaRef`, so re-enabling no longer reopens already-open at a stale position, and the next hover correctly re-earns the delay.
3. **Unbounded "already showing → instant" shortcut** (line 89) — bounded to a grace window via a last-reveal timestamp (reusing `TOOLTIP_DELAY_MS` rather than a second constant). Confirmed reachable exactly as described: a sweep down the flush-margin paragraphs would otherwise stay instant forever after the first reveal. An expired grace period now hides the stale tooltip immediately instead of leaving mismatched content on screen during the re-armed delay — a step further than "add a timestamp," since the literal timestamp-only version still left a stale marker's tooltip lingering for ~300ms while hovering a different one.
5. **Test mock shape** (line 32) — now echoes each requested key back (matching the real hook's `defaultState[key] = key`), with assertions against the literal keys instead of the raw-USFM fallback.
7. **Stale-closure race** (line 103) — added a `hoveredDataRef` kept in sync by two write paths (`updateHoveredData`, `revealHoveredData`); `handleMouseOver`'s deps drop back to stable references.
4. **Comment mechanism was wrong** (line 91) — corrected: the real reason is `TooltipTrigger` being `pointer-events-none`, so it never receives the pointer events Radix's own delay arms itself from — not merely that `open` is caller-forced. Added the pointer at line 234 too.

## Also fixed while in there

- **#6** (line 139) — deleted; confirmed it passed byte-for-byte against `014cdcd51d4` (only exercised #2577's pre-existing prop-gating).
- **#8** — added coverage: keydown-clears-timer (the load-bearing one), scroll-during-delay (pins #1), detached-paragraph, grace-period-boundary (pins #3), and a re-enable test (pins #2).
- **#9 + #11** — narrowed the `TOOLTIP_DELAY` doc comment (no longer claims "every tooltip," names the prop to pass it to).
- **#10** — renamed `TOOLTIP_DELAY` → `TOOLTIP_DELAY_MS` before its first release, per repo convention.
- **#12** (partial) — fixed the missed duplicate at `media-viewer.component.tsx:133`. The `dist/` merge conflict is resolved too (main had moved 11 commits since, including its own unrelated `z-index.ts`/`tooltip.tsx` changes — no overlap with this PR's source, only the generated bundles needed a rebuild).
- **#13** — nits batched in (dep-array comment, Storybook dwell mention).

## Intentionally not done

- **#14** (Architecture-Decisions.md entry) — skipping for now; there's a separate conversation brewing about that log's size/verbosity that this shouldn't get ahead of.
- The `useDelayedFlag` reuse question came up again on my end independently — still doesn't fit (sandbox import boundary, no identity-reset on target change without lifting hover-tracking into state, no data/geometry capture, no grace-period concept). Same conclusion as Part 3.

## One thing I found in manual testing, not in the review

Scrolling away with the tooltip open has a brief (<300ms) lag before it disappears, and it visibly continues tracking the old paragraph's position while lagging. Traced it to `handleScroll` legitimately having no way to detect "the cursor is no longer over this paragraph" — it only reacts to a real `mouseout`, and Chromium doesn't dispatch one synchronously from a scroll under a stationary cursor. Leaving this alone for now (pre-existing, outside this review's scope); a real fix would mean tracking last-known pointer position and checking `document.elementFromPoint()` inside the scroll handler.

## Verification

`npm run typecheck`, `npm run lint`, and the full `npm test` all green post-fix and again after the `main` merge. Manually re-checked findings 1 and 2 in the running app.

tombogle (commented):
> @tombogle made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2810)* status: 0 of 15 files reviewed, 1 unresolved discussion (waiting on katherinejensen00).

___
*[`extensions/src/platform-scripture-editor/src/paragraph-marker-tooltip/paragraph-marker-tooltip-overlay.component.tsx` line 86 at r1](https://reviewable.io/reviews/paranext/paranext-core/2810#-P1X8cy97BuxsXZgkZFh:-P1_miui95vPQXTcAwhJ:b-896fix) ([raw file](https://github.com/paranext/paranext-core/blob/d629644dc0414e95949a8a030646a6fd32026877/extensions/src/platform-scripture-editor/src/paragraph-marker-tooltip/paragraph-marker-tooltip-overlay.component.tsx#L86)):*

### 1. Scroll during the 300 ms window reveals the tooltip in the wrong place

**Attach:** **[A]** line 86 · **Medium**

The position is measured when the hover starts and frozen until the timer fires, so anything that
moves the paragraph in between is missed.

`computePosition` runs at line 86 and is captured into `newData` at line 88; the timer at line 96
commits that stale value. Of the two terms in `clampTopToVisibleArea`, `topInContent` is
scroll-invariant but `visibleAreaTop = scrollContainerRect.top - anchorRect.top` is not — which is
exactly why `handleScroll` exists. During the pending window it can't help:
`setHoveredData((prev) => { if (!prev) return undefined; … })` at line 195 bails while `hoveredData`
is undefined, and it never cancels the timer either. The trigger is narrow (only when the clamp
binds, i.e. the paragraph straddling the viewport top, and only when the pointer stays on that same
paragraph so line 76's early return suppresses a recompute), and it self-heals on the next scroll —
but the fix is three lines: move the `computePosition` call and the `lastPositionRef` write inside
the timeout callback, reading `currentParaRef.current`. That also closes findings 5 and 6 of Part 1.
</blockquote></details>

Done.

katherinejensen00 (commented):
Round 2 — re-reviewed at bbdbeea5c. Ten of the fourteen round-1 findings are fixed and I verified each against the code rather than taking the summary on faith: position and paragraph validity really are recomputed at fire time, the enabled-flip cleanup really does clear both refs, the mock now matches the real hook's key→key seeding, and the new tests are falsifiable (I traced each one to the production line it pins and ran the file locally — 10/10). The merge is clean and the dist bundles are a genuine rebuild: dist/index.d.ts differs from main by exactly the one new declaration.

Two things to do before this lands, and one of them is a real re-open.

1. CI is red on formatting

Build on ubuntu-22.04 fails at "check JS/TS formatting" before lint or packaging run. The file is lib/platform-bible-react/src/components/tooltip-delay.ts — the JSDoc block needs one rewrap:

 * don't pass this constant, so it is opt-in, not automatic. Extensions can't reach app-side
 * renderer constants directly (see the repo's Security-Guide.md Module Import Restrictions), so —
 * mirroring `Z_INDEX_OVERLAY` and its siblings in `z-index.ts` — this lives here instead.

npm run format fixes it. Since this file's dist copy is committed, re-run npm run build:pbr after.

2. The grace window restamps on every hop, so finding 3 is still open

revealHoveredData stamps lastRevealAtRef, and it is called for both the delayed reveal (line 155) and the instant grace-path switch (line 131). So the window measures "time since the last paragraph change", not "time since sustained engagement began" — which leaves the original scenario live and adds a new one:

The sweep is still unbounded. Dwell 300ms on paragraph 1, then sweep down the chapter. Every hop lands within 300ms of the previous reveal, so every hop takes the instant path and restamps. The tooltip strobes through the whole chapter at full mouse speed — exactly what finding 3 described. The comment at lines 56-61 says this is prevented; the code does the opposite, so that comment needs to go whichever way the fix lands.

The engaged reader now pays instead. Rest on a paragraph and read the tooltip for a second, then move to the next one: grace has expired, so line 138 hides immediately and line 145 re-arms 300ms. With flush paragraphs, moving down a chapter at reading pace is hide → 300ms blank → show at every paragraph. Pre-PR and at round 1 that was an instant swap.

Third case: pointer parked over the text while scrolling. Paragraphs stream under a stationary cursor, each firing mouseover inside the restamped window, so the tooltip flashes through each one for the whole scroll. handleScroll deliberately avoids restamping (lines 258-259), but the mouseovers it triggers restamp anyway.

Stamping only inside the timer callback gives you the "continuous engagement" bound the comment claims. Radix's own skipDelayDuration — the thing this reimplements — starts its window when a tooltip closes, so a lastHiddenAt stamp is the other shape that works. Worth saying: the sweep hazard that motivated bounding at all is already handled by arm-and-cancel, since a fast pass reveals nothing (every handleMouseOver calls clearHoverTimer first).

Whichever you pick, hide-on-expiry at line 138 is a user-visible behavior change that isn't in the PR body, and the stated scope is "quick mouse pass". And the current test (lines 137-178) parks 301ms before the second switch, so it only covers the pause case — no test does two consecutive hops inside the window, which is the case that exposes the restamp. The suite stays green through the whole defect.

Worth a look

3. A scroll during the pending window still reveals for the wrong paragraph. handleScroll's rAF body requires hoveredDataRef.current (line 255), so while a reveal is merely pending it does nothing — it neither cancels the timer nor re-targets it. Scroll with a stationary pointer (wheel, or a programmatic scroll-group scrRef sync) and 300ms later the tooltip reveals for the paragraph that was under the pointer, clamped to the visible-area edge. isConnected can't catch this. clearHoverTimer() plus a currentParaRef reset at the top of handleScroll is the one-line version.

4. The hidden-view rationale covers only half the rule. The comment at lines 42-44 answers "no new timer can start while hidden", but not "a timer already armed when the pane is hidden". A programmatic tab activation elsewhere doesn't fire mouseout first; the callback still runs, a display:none rc-dock pane is isConnected so the guard passes, computePosition reads zeroed rects, lastPositionRef gets the bogus position, and revealHoveredData opens the tooltip and stamps the grace timestamp. On re-activation you get a tooltip nobody hovered, near top:0 — the failure lines 48-49 exist to prevent. enabled doesn't change on tab hide, so the cleanup doesn't help. Either extend the rationale to say why that window is unreachable, or check visibility at fire time (useViewVisibility; reference consumer is useBcvSyncScroll in legacy-comment-manager).

5. Test gaps that are cheap to close. No test has a tooltip visible and then leaves the paragraph, so the hide in handleMouseOut (line 182) and all of handleMouseLeave (195-200) can be deleted with the suite green — that's the feature's most basic behavior. Hovering paraA instead of paraB at test line 296 would pin the currentParaRef reset at line 281 for free (removing that line leaves everything green today). A mirror case at exactly 300ms pins the inclusive <= at line 120. And the test named "recomputes the position at reveal time, so a scroll during the pending window is reflected" never scrolls — it asserts a computePosition spy count, so it would stay green against an implementation that computed lazily and then revealed from the stale lastPositionRef. Asserting the rendered trigger's inline style.top would pin the real thing.

6. Backward-facing comments came back, in the tests. Three sites narrate the round-1 findings rather than the invariant: "the old if (hoveredData) check" (138-141), "before it, computePosition ran synchronously" (181-183), "Pins the enabled-flip cleanup fix" (265). Round 1 was clean on this. Each has a keepable core — say what must be true, not what used to be false — and the history goes in the commit message.

7. handleFocus's header comment says the opposite of its body (line 202): "hide tooltip when focus moves outside the editor", but the body is empty and its own inner comment says it exists solely to satisfy jsx-a11y/mouse-events-have-key-events. It's a copy of handleBlur's header, where it's accurate.

8. The PR description is stale in four ways: 11 files (now 15), TOOLTIP_DELAY throughout (shipped as TOOLTIP_DELAY_MS, so searching the diff for the name in the body finds nothing), the round-1 "Critical — None / Important — None / all four passes returned zero" verdict still standing over a head that has since landed a commit fixing five verified bugs, and two call sites listed where there are three (media-viewer.component.tsx is missing).

Nits, take or leave

- One constant, two knobs: TOOLTIP_DELAY_MS is both the reveal delay (line 156) and the grace budget (line 120). Radix keeps delayDuration and skipDelayDuration separate for a reason; a local const MARKER_SWITCH_GRACE_MS = TOOLTIP_DELAY_MS makes the coupling deliberate and stops a future tune of the published constant from silently retuning the grace window.
- Date.now() is wall-clock — a backward clock adjustment reads as "within grace" until the next reveal restamps. performance.now() is monotonic.
- The TSDoc on the constant names ParagraphMarkerTooltipOverlay, which lives in another package. z-index.ts — the precedent this module cites — records that an enumerated consumer list went stale once and says to grep instead.
- handleScroll commits a new state object every frame even when the computed position is identical, which is the common case once the clamp pins to the visible-area top. An early return on unchanged top/left skips the re-render and the Popper re-measure.
- lastPositionRef's pairing is hand-rolled at three sites and the dismissal trio at four; folding the stamp into updateHoveredData and extracting one dismissTooltip() would make handleKeyDown's deliberate divergence visible at a glance.
- vi.stubGlobal + vi.unstubAllGlobals is the self-restoring form of the ResizeObserver stub; today it only works because per-file isolation is on.
- Escape only dismisses while focus is inside the wrapper — the keydown listener is on positionAnchor. With focus in a side panel or the toolbar and the pointer still over the editor, nothing but moving the mouse dismisses it. WCAG 1.4.13 wants hover content dismissable without moving the pointer; a document-level capture listener costs nothing.
- The marker name is hover-only: the trigger is aria-hidden and tabIndex={-1}, so Radix's aria-describedby wiring never reaches the accessibility tree and handleFocus is a no-op. Keyboard and screen-reader users get nothing. Fine to record as an accepted limitation if another surface covers it, but worth deciding now that the reveal machinery has been rebuilt.
- tooltip.stories.tsx:260 still hardcodes delayDuration: 300 inside the package that now exports the constant, and main's new Guidelines/Tooltips page says nothing about a reveal delay.

Dropped after checking

The "constant overclaims" point from round 1 is genuinely resolved — the narrowed doc now says outright that TooltipProvider's default stays 0 and the constant is opt-in, which matches the tree (64 TooltipProvider usages in non-dist source, 4 on TOOLTIP_DELAY_MS, 8 explicit 0, and 400/200/150/1000 one-offs). Converging those one-offs is a separate UX call, not this PR's debt. media-viewer.component.tsx is in scope, not creep — same value, pre-existing literal, completes the de-duplication. Security is clean. And useDelayedFlag still doesn't fit; I re-checked the alternatives (useTruncationTooltip has no delay, debounce rejects its pending promise on cancel) and found nothing better.

One pre-existing thing, flagged so it isn't read as a regression: handleMouseOver never reads suppressUntilMoveRef — typing suppression works only because currentParaRef is deliberately left set. If typing reflows text under a stationary cursor (pressing Enter), a mouseover for a different paragraph fires with no mousemove and the tooltip can reappear mid-typing. That's there at the merge base.

> @katherinejensen00 reviewed 15 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2810#-:-P1a-JDlBLmE65L_MTV7:b-jik4eg)* status: all files reviewed, 1 unresolved discussion.

tombogle:
https://github.com/user-attachments/assets/bda756d8-dc88-469c-b7ce-4081303eca1d

This video shows everything working as expected.

katherinejensen00 (commented):
# PR #2810 re-review (round 3): PT-4536, Add a real hover delay to the paragraph-marker tooltip

- **Repository**: paranext/paranext-core
- **PR**: https://github.com/paranext/paranext-core/pull/2810
- **Author**: Tom Bogle (`tombogle`)
- **Reviewed at**: `45a28470349` (round 2 reviewed `bbdbeea5c6b`)
- **New since round 2**: `d8102c8c7b` (fade/flicker fixes: `lastMarkerRef`, `awaitingRealMoveRef`), `ce0e4377f2` (3px jitter threshold), `bb008541ed` (rewrite onto a `paraAtPoint` hit-test; this **removes** both gates the previous two commits added), a merge of main, and a dist rebuild
- **Merge base**: `1f06a30d627`. The diff is 17 files, 7 of them generated dist
- **Mode**: report only; nothing posted to GitHub
- **Panel**: round-2 verification, ux, tests (with mutation runs), clarity+scope+security (4 agents, parallel). No C# in the PR, so no contracts or architecture agents

File keys: **[A]** `extensions/src/platform-scripture-editor/src/paragraph-marker-tooltip/paragraph-marker-tooltip-overlay.component.tsx` ·
**[B]** same folder, `.component.test.tsx` · **[U]** `extensions/src/platform-scripture-editor/src/editor-dom.util.ts` (+ `.test.ts`) ·
**[C]** `lib/platform-bible-react/src/components/tooltip-delay.ts`

## Local verification

| Check | Result |
| --- | --- |
| `prettier --check` **[C]** | **FAIL**, byte-identical to round 2 |
| CI | `Build on ubuntu-22.04` **FAILURE** at `format:check` on **[C]** (C# tests passed 1727/1733). Windows, macOS and CodeQL are green |
| vitest (from `extensions/`) | 65/65 pass (overlay 12, editor-dom.util 43, utils 10) |
| Mutation runs on **[A]**/**[U]** | 9 applied one at a time, all restored (the worktree is clean). **2 killed, 7 survived** (see T1–T4) |

## Round-2 status

**No round-2 finding is addressed.** The new commits target different bugs. The files where B1 and the nits live are unchanged since `bbdbeea5c`.

| Round 2 | Status at HEAD |
| --- | --- |
| B1 Prettier / CI red | Not fixed |
| B2 grace window resets on every hop | Not fixed. `revealHoveredData` still stamps on the instant path (**[A]**:151) and on the delayed path (:179). All three scenarios are still live: sweep strobe, reading-pace hide then a 300ms blank, and a scroll under a parked pointer. The scroll case is probably *easier* to hit now, because mousemove also runs the hit-test |
| 3 scroll during the pending window | Not fixed in code (`handleScroll` :267-282 is unchanged). The mousemove hit-test narrows it only by accident |
| 4 timer armed before the pane is hidden | Not fixed (:46-50) |
| 5 test gaps | Not fixed, confirmed by mutations M3/M4/M7/M6 below |
| 6 backward-facing test comments | Not fixed, and worse: new `Regression test (PT-4536)` at **[B]**:351, :379 and `previously, …` at :379 |
| 7 `handleFocus` header | Not fixed (**[A]**:221). Its inner comment now also points to `handleMouseOver`, where placement no longer happens |
| 8 stale PR description | Not fixed, and staler (see S1) |
| Nits | All unchanged |
| suppressUntilMoveRef pre-existing note | Unchanged in outcome, and a same-paragraph variant now exists (see N2) |

## New findings

### N1: the pointer-events-none fix doesn't cover the element that is actually hit (warning, needs a repro)

**[A]**:339-352. `tw:pointer-events-none` goes on Radix's inner `TooltipPrimitive.Content`. The popper renders a fixed `<div data-radix-popper-content-wrapper>` with `minWidth: max-content` around it, and that wrapper gets `pointerEvents: 'none'` only when `hide.referenceHidden` (`node_modules/@radix-ui/react-popper/dist/index.mjs`:146-165). The wrapper's box is the tooltip's box, so `elementFromPoint` still hits it. The comment's "so this can't happen" is not true of the rendered DOM.

The loop happens when the tooltip box sits over a resting cursor. The likely case is a paragraph at or clamped to the visible top: `side="top"` has no room, so Radix flips the tooltip below the anchor, over the first line.
1. A `mouseout` or `mousemove` from the wrapper bubbles through the React portal tree into `handleMouseOut`/`handleMouseMove`. (React treats a portal as a descendant for enter/leave, so `onMouseLeave` does *not* fire.)
2. `paraAtPoint` hits the wrapper, which is not a paragraph, so the tooltip hides.
3. The exit animation ends and the wrapper unmounts.
4. The paragraph is hit again, the 300ms delay runs, and the tooltip reveals.
5. The loop repeats.

`d8102c8`'s own message says pointer-events-none "wasn't sufficient by itself", and `bb008541` removed the gate that was masking this. Fix: `pointer-events: none` on the wrapper as well, for example a scoped rule on `[data-radix-popper-content-wrapper]:has(> …)`. At minimum, soften the comment.

### N2: typing suppression is cleared by synthetic mousemoves (warning, needs a decision)

**[A]**:200-210. Any mousemove clears `suppressUntilMoveRef`. `d8102c8`'s message documents that a synthetic mousemove comes with editor DOM churn. `ce0e437`'s position-change requirement (the 3px threshold) blocked that. `bb008541` deleted it.

Trace, with the pointer parked over P while the user types or presses ArrowDown:
1. keydown sets suppress and hides the tooltip.
2. The decoration swap sends a synthetic mousemove, which clears suppress and sets `currentParaRef` to undefined.
3. The next churn event hit-tests P, which no longer equals the ref, and arms the delay.
4. The first 300ms pause in typing reveals P's tooltip under a mouse that never moved.

No test covers keydown, then mousemove, then 300ms. Fix: record clientX/Y at keydown and ignore moves within a few px, which is the old gate scoped to the keydown path.

### N3: a forced layout per pointer event (suggestion)

**[A]**:126. `elementFromPoint` runs on every mousemove, mouseover and mouseout (3× per boundary crossing). It forces a style and layout flush while Lexical has dirty layout, for example during typing or drag-select. The sibling `character-marker-bar-overlay.component.tsx` coalesces to one rAF per frame. Suggested: store the latest x/y and schedule a single rAF.

### S1: PR description and commit messages are stale (warning, needs a decision)

The body still says 11 files, `TOOLTIP_DELAY` (not `TOOLTIP_DELAY_MS`), two call sites (three counting media-viewer), and 6/6 tests. It keeps "Critical/Important: None" and says positioning/animation behavior is unchanged. It says nothing about `lastMarkerRef` exit content, pointer-events-none, the `paraAtPoint` hit-test rewrite, or the flicker-on-leave fix.

`bb008541`'s message is subject-only. The `d8102c8`/`ce0e437` messages describe mechanisms that no longer exist, so the squash message needs rewriting.

Decision: rewrite the description (and maybe the title) to cover the fade/flicker work, or split it out. Keeping it here is defensible: same component, found while testing.

### Tests (mutation-backed)

| # | Mutation on **[A]**/**[U]** | Result |
| --- | --- | --- |
| M1 | `displayMarker = hoveredData?.marker` (drop the `lastMarkerRef` fallback, :85) | survived |
| M2 | delete the same-paragraph early return (:127) | survived |
| M3 | `handleMouseOut` becomes a no-op | survived |
| M4 | delete the trailing `updateHoveredData(undefined)` (:185) | survived |
| M5 | remove the mousemove hit-test call (:209) | **killed** |
| M6 | `<=` becomes `<` on the grace boundary (:139) | survived |
| M7 | `handleMouseLeave` returns early | survived |
| M8 | grace switch uses `updateHoveredData` (no restamp, :151) | survived |
| M9 | `paraAtPoint` without `.closest` | **killed** |

- **T1 (warning).** No test shows a tooltip and then leaves the paragraph (M3/M4/M7). Hiding is now driven only by the hit-test.
- **T2 (warning).** No test covers two hops within grace (M8) or the exactly-300ms case (M6). The suite can't tell the B2 behavior from its fix.
- **T3 (warning).** The churn test at **[B]**:350 says the pair "must not touch the tooltip at all", but M2 survives. Advance past grace before firing the pair, or fire it while a reveal is pending and assert the original deadline.
- **T4 (warning).** `lastMarkerRef` exit content (the visible defect `d8102c8` fixed) is unpinned (M1).
- **T5 (suggestion).** `editor-dom.util.test.ts`:592 ("disagrees with a stale event target") is a duplicate of :575. `staleTarget` is never passed in, so `not.toBe(staleTarget)` passes by construction.
- **T6 (suggestion).** No assertion pins the `pointer-events-none` class, and N1 argues it isn't enough anyway. `stubHoverTarget` ignores coordinates, so no test exercises position-dependent behavior (jitter at a paragraph edge).

### Clarity (suggestions)

- `handleMouseOver` and `handleMouseOut` (**[A]**:190-198) are now identical wrappers. Use one handler for both events.
- Comments that describe the old event-boundary model: :204 "on any element boundary", and `handleKeyDown` :258-261 "the next mouseover re-evaluates". It isn't documented that the first mousemove after a keydown only clears suppression.
- `paraAtPoint` TSDoc (**[U]**:36): the past-tense "Consumers that based a hover boundary on the event's target inherited that disagreement" describes this PR's old code. Say it as guidance. The 13-line TSDoc on a one-liner is also repeated at **[A]**:38-43 and :116-120.
- No duplicate utility: `elementFromPoint` appears nowhere else, and `character-marker-bar.utils.ts`:55 starts from an element, not a point.

### Security

Clean. DOM reads only, allowed imports, and tooltip text renders as a React text child.

> @katherinejensen00 reviewed 10 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2810#-:-P1k0ZisEbANYg38yhl6:b-4zfxcz)* status: all files reviewed, 1 unresolved discussion.

katherinejensen00 (commented):
Nothing is actually blocking from the review above so going ahead and approving.

> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2810#-:-P1k1-l82-2OZRCRGYck:b-psg81g)* status: all files reviewed, 1 unresolved discussion.

katherinejensen00 (approved):
:lgtm:

> @katherinejensen00 made 1 comment and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2810#-:-P1k179W7125Xd0e5q6O:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2811, opened 2026-09-14
a11y: make info-icon tooltips keyboard-reachable

Sebastian-ubs:
## Summary

- Wrap the bare `<Info>` svg inside each `TooltipTrigger asChild` in a ghost-variant icon button so the trigger is focusable, carries `role=\"button\"`, and can be reached by keyboard and screen readers.
- Radix opens the tooltip on focus and wires `aria-describedby`, so the explanation is announced as the button's description.
- The `aria-label` on each button is a short noun phrase for *what* the button explains (the adjacent option label), not the tooltip text itself — per the [Providing Help](https://github.com/paranext/paranext-core/pull/2787) guideline.

## Sites fixed

| File | Before | After |
| --- | --- | --- |
| `extensions/src/platform-scripture/src/find/find-filters.component.tsx:128` | Bare `<Info>` as `TooltipTrigger asChild` — \"All texts\" tooltip hover-only | Ghost `Button` with `aria-label={label}` wrapping the `<Info>` |
| `extensions/src/platform-scripture/src/find/find-filters.component.tsx:214` | Bare `<Info>` — \"Ignore whitespace differences\" tooltip hover-only | Same fix, `aria-label={localizedStrings.ignoreWhitespaceDifferences}` |
| `extensions/src/platform-scripture/src/find/find.component.tsx:1038` | Bare `<Info>` — \"Preserve case\" tooltip hover-only | Same fix, `aria-label={localizedStrings['%webView_find_preserveCase%']}` |
| `extensions/src/platform-scripture/src/manage-books-dialog/manage-books-dialog.component.tsx:2528` | Bare `<Info>` with an `aria-label` that was never announced (SVGs aren't in the a11y tree as interactive) | Ghost `Button` with the `aria-label` moved onto the button; uses `%manageBooks_create_method_referenceText%` (\"Create based on\") for the short name instead of the tooltip content key |

## Context

Follow-up to [#2787](https://github.com/paranext/paranext-core/pull/2787) finding #2 — the new *Providing Help* guideline prescribes a ghost-variant icon button around `<Info>`, but every site shipping today used the bare svg pattern, leaving the tooltips unreachable by keyboard and screen readers. This PR retrofits the four broken sites the reviewer named. The three other sites named in the review (`toolbar.component.tsx:391`, `warning-ribbons.component.tsx:96`, `marble-guide.component.tsx:292`) are already correct — the toolbar site wraps in a button, and the other two are decorative alert/illustration icons rather than tooltip triggers.

## Test plan

- [ ] Tab to each of the four info icons and confirm focus lands on the button (visible focus ring).
- [ ] Confirm the tooltip appears on keyboard focus, not only on mouse hover.
- [ ] With VoiceOver / NVDA, confirm each button announces its short name followed by the tooltip text as its description.
- [ ] `npm run lint` and `npm run typecheck` clean.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2811)


jolierabideau (changes requested):
The accessibility fix is right and the scope claim checks out — I swept the whole branch and there are no bare-icon `TooltipTrigger` sites left. One pattern issue is worth fixing before merge; the rest are small.

### 1. The icon size classes lose to the Button's svg contract (all four sites)

`buttonVariants` carries `tw:[&_svg:not([class*=size-])]:size-4` in the base and `…size-3` under `icon-xs`. The guard only steps aside for a child whose class contains `size-`. These `<Info>` icons use `tw:h-3.5 tw:w-3.5` (and `tw:h-4 tw:w-4` in manage-books), so the guard doesn't exclude them — and the parent rule has higher specificity (descendant + `:not([class*=size-])` ≈ `(0,2,1)` vs `(0,1,0)` for `tw:h-3.5`). Net effect: the icons should render at **12px**, not the 14/16px they declare.

The reference site in `toolbar.component.tsx:399` masks this — `size="icon"`'s base `size-4` happens to equal its `h-4 w-4`, so nothing looks wrong there.

Fix: drop the `h-/w-` classes and let the button size the icon, or switch to `tw:size-3.5`. Worth a quick look in Storybook to confirm the shrink, since this is reasoned from the cascade rather than from rendered output.

### 2. `size="icon-xs"` deviates from the guideline it cites

#2787 prescribes `variant="ghost" size="icon"`. This PR uses `size="icon-xs"`, which has no other production usage in the repo — these four would be its first. `icon` (32px) genuinely is chunky inline next to small radio labels, so `icon-xs` may be the better call — but then amend #2787 to cover the inline-with-text case rather than leaving the first four consumers off-pattern.

### 3. Sequencing

#2787 is still open. Pinning four sites to a convention that could still change in review means a possible second pass. Worth coordinating so whichever lands second matches.

### 4. `ghost` + `tw:text-muted-foreground` overlaps the `subtle` variant

`subtle` exists for exactly "no background, muted at rest, foreground on hover." Ghost adds a hover background that `subtle` lacks, so this may be deliberate — if so, a line in the guideline saying which to reach for would save the next person the decision.

### 5. Two nits

- `tw:cursor-default` was dropped at the preserveCase site. Tailwind v4's preflight gives buttons a default cursor, so this is probably fine — just confirming it was intentional.
- `aria-label={label}` on the "All text" row repeats the adjacent visible `<Label>`; a screen reader reads "All text … All text, button." Guideline-conformant as written, but a distinct noun phrase would read better.

### Verified clean

Localization keys both resolve (`%manageBooks_create_method_referenceText%` in `localizedStrings.json:318` and the types file); `%manageBooks_create_basedOnInfo%` is still the tooltip content, so it isn't orphaned. `Button` was already imported in all three files. No site sits inside a `<form>`, so the shadcn `Button`'s missing default `type="button"` isn't a submit hazard here. No `shadcn-ui/` files touched, so no `// CUSTOM:` obligation; no keyboard handlers, so the shortcuts catalog doesn't apply. No backward-facing comments in the diff.

Branch sweep caveat: the grep matched `TooltipTrigger asChild` plus the following 2–4 lines, so it would miss a same-line child, a child more than four lines down, or an icon outside the name list I used.

---

🤖 Review assisted by [Claude Code](https://claude.com/claude-code)

Sebastian-ubs:
Thanks. I got aware that none of these need separate buttons, but should instead be tooltips on the whole control, so I am changing this now.

PR #1754, opened 2025-08-07
PT-2996: Add info/feedback menu items to the help menu

rolfheij-sil:
image

Note that the `FAQs` menu item replaces the pre-existing `Visit Support.Bible` menu item

This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/1754)

PR #2812, opened 2026-09-14
PT-4533: Prune Simple's main and Help menus

captaincrazybro:
# Code Review Summary

**Branch**: pt-4533-prune-simple-menus

**Base**: origin/main

**Date**: 2026-09-14

**Files changed**: 29 at review start; 31 after in-review fixes

## Overview

Simple mode offered main- and Help-menu commands that lead nowhere in it. This change hides seven
items per interface mode using the existing `hiddenInterfaceModes` filter, and relabels the Help
menu's "FAQs" to "Community support" so the name matches the Support.Bible destination the item has
always opened. No production code was added — the change reuses the existing contribution schema and
`filterItemsForInterfaceMode` path rather than introducing a parallel mechanism, and the E2E mode
guard reuses `cdp.fixture`'s existing `requiredInterfaceMode` option.

Hidden in Simple: `Open...` (platform-get-resources), `Open Dictionary: SDBH/SDBG`
(platform-lexical-tools), `Open enhanced resource` (platform-enhanced-resources),
`Paratext Registration Information...` (paratext-registration), and the three core Help items
Getting started, Feature roadmap, and Open Developer Documentation. Kept in Simple: Send/Receive
(contributed out-of-repo), Settings, Exit, Community support, Show the tour, the two Usersnap
feedback items, and About.

The bulk of the diff is tests: four new test files covering shipped menu data, real extension
contributions through the actual combiner, command dispatch, and menubar rendering. A companion PR in
`paratext-bible-internal-extensions` (paranext/paratext-bible-internal-extensions#202) hides
"Manage extensions". The two are **independent and can land in either order**: `hiddenInterfaceModes`
entered the schema in `8843aa374c4` (#2602, 2026-07-28), which is already on `origin/main`, and this
PR adds no schema surface — so current core already validates the companion's document. They are
still worth landing near each other, because `platform.projectResources` only actually goes empty in
Simple once both have landed.

**Not addressed here:** PT-4503 (menus don't re-localize on a language change) is pre-existing
and unfixed on `main` — verified 2026-09-16, both of its Definition-of-Done items absent
(`invalidateLocalizedOutput` and any `interfaceLanguage` subscription). It affects every menu
label equally and is out of scope for this PR.

## API Changes

None. No public API surfaces changed on this branch.

- `lib/papi-dts/papi.d.ts`: unchanged (no `@papi/` module additions, modifications, or removals)
- `lib/platform-bible-utils/`: unchanged
- `lib/platform-bible-react/`: only a new test file; no change to any exported component, hook, or type
- `extensions/src/*/src/types/*.d.ts`: unchanged — `platformGetResources.openHome`,
  `platformLexicalTools.openDictionary`, `platformEnhancedResources.openEnhancedResource` and
  `paratextRegistration.showParatextRegistration` all keep their declared command signatures. The
  change is menu visibility only.
- `platform.visitFAQsPage` is **relabeled, not renamed** — the command id and its handler are
  untouched, so no contribution referencing it breaks.

None.

- [ ] ~~Hiding the Dictionary removes the **only** entry point to it in Simple mode — the feature
      becomes unreachable, not relocated. Verified three ways: the extension's `defaultWebViewTopMenu`
      and `defaultWebViewContextMenu` items are empty and `webViewMenus["platformLexicalTools.dictionary"]`
      is `{}`; `platformLexicalTools.openDictionary` has no other caller in `src/`, `lib/`, or
      `extensions/`; and Simple's fixed third column in `simple-layout.data.ts` has no dictionary tab.
      The other three hidden extension items each have a verified replacement path.~~
      _(Author: deliberate — "there will be a better way later". Tracked by PT-4545 (the TODO was removed from the
      translator-facing `localizeNotes` in review; the deferral is recorded here and on the ticket). Note the PRD contradicts itself here: NN-1.1's screenshot removes the item,
      NTH-5.2's text asks to keep one.)_
- [x] **The three core Help items gated in `menu.data.json` — `platform.visitGettingStartedPage`,
      `platform.visitFeatureRoadmapPage`, `platform.openDeveloperDocumentationUrl` — had no test
      asserting they are hidden in Simple and present in Power**, while the four extension-contributed
      items did. Deleting `hiddenInterfaceModes` from any of the three left the suite green.
      _(fixed during review: parameterized the existing `Platform menu document interface-mode gating`
      block in `menu-data.service-host.test.tsx` over all four Power-only core commands, asserting both
      directions — six new assertions. Mutation-checked: removing the flag from
      `platform.visitGettingStartedPage` turns exactly that test red.)_
- [ ] ~~`localizeNotes` is typed as information for people performing localization, and
      `Entry-Point-Guide.md` says it should carry the translator-facing menu path. All five changed menu
      files now append Simple-mode gating rationale there, and the lexical-tools entry ships
      `TODO(PT-4545)` into the localization pipeline.~~ _(Author: rationale left as-is on the existing
      precedent — the pre-existing manage-books item in `platform-scripture-editor/contributions/menus.json`
      does the same thing, and JSON has nowhere else to put a comment.)_

### Minor — Consider

- [x] **`e2e-tests/fixtures/enhanced-resources.fixture.ts` failed `tsc`** (TS2322 — Playwright rejects
      a derived test re-declaring `requiredInterfaceMode`, which `cdp.fixture` already declares as an
      option). Found while relocating the fixture, but **present in the file as originally committed** —
      confirmed by materializing the original at its original path and getting the identical error. The
      file does not exist on `origin/main`, so this shipped with the branch. `npm run typecheck` runs
      `typecheck:e2e` over `e2e-tests/**/*.ts`, and in `.github/workflows/test.yml` the single `test`
      job runs typecheck (line 172) **before** the dotnet tests, npm unit tests and the Linux E2E smoke
      suite, with no `continue-on-error` in that chain — so on the Linux leg this blocked every
      subsequent test step. _(fixed during review: supply `requiredInterfaceMode: 'power'` as a plain
      fixture value instead of an option tuple. Behaviourally identical — no spec in `e2e-tests/`
      overrides it via `test.use()`. Full `npm run typecheck` now exits 0 across all four legs.)_
- [x] **`MENU_CONTRIBUTING_EXTENSIONS` was a hand-maintained list of four extensions** while its
      docblock claimed the result was "the main menu as a running app actually assembles it".
      `hello-rock3` contributes four `mainMenu` items and was missing. No assertion was wrong today
      (hello-rock3 contributes into `platform.projectProjects` and its own group, neither of which is an
      emptied group), but a future extension contributing into `platform.projectResources` or
      `platform.helpRegistration` would have escaped the empty-group assertions silently. No
      pre-existing test anywhere reads `contributions/menus.json` off disk, so this pattern is new here.
      _(fixed during review: replaced with a scan of `extensions/src/*/contributions/menus.json` keyed
      by manifest `name`, keeping documents with non-empty `mainMenu.items`; now discovers five
      contributors. Added a discovery guard, since a scan that silently found nothing would make every
      "absent in Simple" assertion trivially true — breaking the scan turns 10 tests red.)_
- [x] **`text-collection-schema.spec.ts` was pinned to Power mode but never opens the Power-only
      menu item** — it drives PAPI through `mainPage.evaluate` and matches zero `openEnhancedResource`
      calls. `cdp.fixture` documents "leave unset when the spec genuinely works in either mode", and the
      mode assertion fires at fixture setup, before the spec's own `test.skip` runs.
      _(fixed during review: moved back to `cdp.fixture` with a comment explaining why it differs from
      its 14 siblings; corrected the README's "this whole suite requires it" claim.)_
- [x] **`describe('Every menu label and tooltip is localized')` swept only three of the document's
      sections**, omitting `defaultWebViewTabMenu` (3 items) and `webViewMenus`.
      _(fixed during review: extended to `defaultWebViewTabMenu` and to column labels, with a note that
      `webViewMenus` is empty in the shipped document. English is asserted for all labels; Spanish for
      item labels only, because the Platform.Bible column is labelled with the product name, which ships
      English-only in every locale by design. Mutation-checked: blanking a tab-menu label in `en.json`
      turns that case red. 35 → 44 tests.)_
- [x] **`Entry-Point-Guide.md` cited `platform-scripture-editor/contributions/menus.json` as *the*
      production user of `hiddenInterfaceModes`**; this branch adds five more sites.
      _(fixed during review: the line now names both the core menu document and extension contributions
      generally, and points at `grep -rn hiddenInterfaceModes` for current sites rather than one file
      that goes stale on the next change. The grep was run to confirm it returns what the doc promises.)_
- [x] **`er.fixture.ts` was the first Playwright fixture placed inside a `tests/` subdirectory** (all
      others live in `e2e-tests/fixtures/`, including the likewise suite-scoped `find.fixture.ts`), and
      `er` is a coined initialism in a filename. _(fixed during review: `git mv` to
      `e2e-tests/fixtures/enhanced-resources.fixture.ts`, all 14 importers repointed, prose references
      in the README and `text-collection-schema.spec.ts` updated.)_
- [ ] ~~`%mainMenu_visitSupportBible%` now resolves through a two-hop fallback chain
      (`→ %mainMenu_helpInfo_visitFAQsPage% → %mainMenu_helpInfo_visitCommunitySupportPage%`).
      `findLocalizationForFallbackLanguageAndOrKey` reads `metadata.fallbackKey` once and looks it up
      directly in the language data — it never re-enters `getMetadata`, so a chain is never followed.
      This resolves today only because the retired FAQs key still carries en/es values; whoever
      eventually removes them makes `%mainMenu_visitSupportBible%` surface as a raw key. The
      `visitSupportBible` line itself is untouched by this branch, but on `main` the FAQs key had no
      metadata entry at all, so the second hop is this branch's doing.~~
      _(Author: out of scope. Known-latent — a one-line repoint at the new key closes it whenever
      someone wants to.)_
- [ ] ~~The command id `platform.visitFAQsPage` now disagrees with both its label ("Community
      support") and its destination; the rationale for keeping the id lives only in test doc-comments.~~
      _(Author: pre-existing. Confirmed — the handler is untouched by this branch, and on `main` the id
      already pointed at `https://support.bible/paratext-10-studio`. The relabel only made a mismatch
      that already existed easier to notice.)_
- [ ] ~~The `settingsService` mock and the `implementMenuDataDataProviderEngine` +
      double-`await Promise.resolve()` bootstrap are duplicated between the two menu-data test files.~~
      _(Author: leave the duplication. Both files mock `settingsService` at module scope, so a shared
      helper could only factor out ~6 lines of engine construction.)_
- [ ] ~~`NoopResizeObserver` plus the `hasPointerCapture`/`scrollIntoView` jsdom stubs are copied
      verbatim into the new menubar test.~~ _(Author: leave as-is. Verified pervasive — 34 test files
      carry a ResizeObserver stub and 21 stub pointer/scroll. The new file follows the convention;
      hoisting them into the root `vitest.setup.ts` is a repo-wide refactor that doesn't belong here.)_
- [ ] ~~Two Simple-mode presentation calls flagged for UX sign-off: "Getting started" is hidden from
      the mode built for newcomers (with "Show the tour" only partly covering it), and Simple's
      Platform.Bible column is reduced to two items — Settings and Exit — behind a full menubar column
      trigger.~~ _(Author: intended, and what the PRD asked for.)_

#### Raised by the analyzers but **not** this branch's work — no action taken

These were verified against the merge base and dropped from the fix list so they aren't read as debt
this PR created:

- `'https://studio.paratext.org/start '` carries a trailing space that reaches `shell.openExternal`
  untrimmed. Introduced by `ba09ac922be` (PT-2996, #1754), an ancestor of `main`. This branch's new
  test is simply the first thing to pin it.
- Label guideline violations on items this diff gates — title case in "Open Developer Documentation"
  and "Paratext Registration Information...", and `...` instead of U+2026. The UX pass claimed "the
  diff already touches every one of these entries"; **it does not**. All three `localizedStrings.json`
  files are untouched and the only `en.json` change is one added line.
- "Open Dictionary: SDBH/SDBG" uses two unexplained initialisms in user-facing microcopy.
  `platform-lexical-tools/contributions/localizedStrings.json` is untouched by this branch.

#### Shared Regions Modified

#### Extension Config Changes

- [n/a] `extensions/src/paratext-registration/contributions/menus.json` — menu contribution content, not shared build/config infrastructure
- [n/a] `extensions/src/platform-enhanced-resources/contributions/menus.json` — same
- [n/a] `extensions/src/platform-get-resources/contributions/menus.json` — same
- [n/a] `extensions/src/platform-lexical-tools/contributions/menus.json` — same

No dependency, script, CI, or build-config files changed.

## Positive Observations

- **Localized-string immutability handled exactly as prescribed.** A new key
  `%mainMenu_helpInfo_visitCommunitySupportPage%` rather than an edit to the shipped value, the old key
  left byte-identical and still resolvable, and a `fallbackKey` redirect in `metadata.json` — the
  two-step replacement protocol in `Localization-Guide.md`. `fr`, `km`, `zh-hans` and `zh-hant` never
  carried the old key, so no locale loses a translation.
- **Both halves of every gate are asserted.** Hidden-in-Simple *and* present-in-Power, which is what
  catches an over-broad flag that would hide an item in every mode. The reasoning is written into the
  test rather than left implicit.
- **The emptied-group consequence is not hand-waved.** The tests establish that
  `platform.projectResources` and `platform.helpRegistration` go empty in Simple, that no whole column
  does, and that the menubar emits no stray separator. The analyzers independently recomputed the
  combined menu and traced `getMenubarContent`, reaching the same result.
- **Tests sit at four complementary levels** — shipped-data assertions, real-contribution integration
  through the actual combiner and data provider engine, command dispatch, and rendering. The
  `menu.data.test.ts` localization sweep is genuinely new coverage; nothing previously checked menu
  labels against the shipped locale files.
- **Verified live in the running app in both modes** before the branch was pushed, with screenshots
  checked for rendering problems the DOM wouldn't show.
- **The E2E mode guard reuses the existing `requiredInterfaceMode` option** rather than inventing a
  mechanism, and the suite README was updated in the same change rather than left stale.
- **Each gate carries its rationale** in its `localizeNotes`, including the Dictionary having no
  Simple entry point — an honest record of a known gap rather than a silent one. The deferral itself
  is tracked by PT-4545.

## Interview Notes

**Stated purpose**: prune Simple's main and Help menus — hide Power-only commands per interface mode
and rename Help's "FAQs" to "Community support".

**Author's design explanation (empty groups)** — asked to walk through why the emptied groups were
pinned by test rather than fixed: *"I'm not entirely sure if the call is the thing. But this call was
meant to test out and see how it would look and then make a future judgement call. I'm doing that now
in the paratext-10-studio repo."* The author subsequently confirmed they **did** verify the menus in
the running app and saw no floating empty groups. This is corroborated rather than asserted: the UX
analysis independently traced `getMenubarContent` and confirmed the shipped Simple menu cannot produce
a leading, trailing, or doubled separator — the two emptied groups sit leading/middle, and both
columns' last groups retain an item ("About Platform.Bible", "Exit").

**Open, by the author's own account**: whether deferring empty-group suppression to PT-4532/PT-4534 is
the right long-term call. The author is explicitly not claiming it is; the question is live in
`paratext-10-studio`. This is a design question for the review meeting, not a resolved one.

**Author judgment calls recorded**: the Dictionary losing its only Simple entry point is deliberate
("there will be a better way later"); `localizeNotes` carrying gating rationale follows the existing
manage-books precedent; the two Simple-mode presentation calls are what the PRD asked for.

**Author scepticism was well-calibrated and changed the outcome.** The author twice challenged whether
a finding actually belonged to this branch. Both challenges were checked against the merge base, and
in three cases the author was right — the trailing-space URL, the label guideline violations, and the
command-id mismatch are all pre-existing, and one analyzer claim ("the diff already touches every one
of these entries") was simply wrong. Those were dropped rather than fixed. The author also pushed back
on an unnecessary `deprecationInfo` addition; that half was correct — `deprecationInfo` appears
nowhere in the Localization Guide — while the `fallbackKey` half is a documented CRITICAL requirement,
so only the undocumented half was removed.

**Nothing was deferred to AI or left unexplained.** Where the author expressed uncertainty, it was
about a design decision they had deliberately left provisional, not about what the code does.

## In-Review Quality Check

- `npm run typecheck` — **exits 0** across all four legs (`core`, `erb`, `e2e`, `workspaces`). This
  had not been run on the branch before; it surfaced the `enhanced-resources.fixture.ts` TS2322 that
  would have blocked CI, now fixed.
- `npm run lint` — **exits 0** repo-wide. Two `no-null/no-null` errors introduced during the review
  were fixed by narrowing on truthiness rather than suppressed.
- Prettier — clean on every changed file.
- `npm test` — **exits 0**. 11 workspace suites, 660 test files, ~9,329 tests, 0 failures.
- Every test added or changed during the review was mutation-checked to confirm it can fail.

**One anomaly worth knowing:** an earlier invocation of `npm test` exited 1 while reporting 0 test
failures. That run's output was truncated before capture, so the failing leg could not be identified;
a full re-run with complete output was clean across every workspace. `npm test` chains
`test:core && test --workspaces`, so a non-zero exit with no failing test points at a leg failing for
a non-assertion reason. Not reproduced, not explained — flagged here rather than assumed benign, and
worth a second look if CI shows the same shape.

## Suggested Review Focus

- [ ] **Empty groups in Simple.** The author is not asserting the deferral to PT-4532/PT-4534 is
      right — it was a deliberate probe, and the judgement is being made now in `paratext-10-studio`.
      Worth discussing directly rather than treating as settled.
- [ ] **The Dictionary has no Simple entry point at all** until PT-4545 lands. Accepted deliberately;
      confirm product agrees, and note the PRD contradicts itself (NN-1.1 removes it, NTH-5.2 keeps one).
- [ ] **"Show the tour" stays visible in Simple**, against the PRD screenshots' keep-list — so Simple's
      Help menu has 5 items, not the DoD's 4. Flag to Todd.
- [ ] **Two DoD lines are wrong**: Send/Receive needs no core change (contributed out-of-repo, it just
      survives the pruning), and the rename is `en` + `es` only, not "all six locale files" — the other
      four carry no main-menu keys by design.
- [ ] **Q4 in the ticket is answered by the code, not product** — "Manage extensions" exists and is
      labelled exactly that. The ticket only searched core.
- [ ] **`typecheck:e2e` is the only thing that compiles `e2e-tests/`.** Playwright specs aren't in the
      build and ESLint didn't flag the type error this branch shipped. Worth noting for the team's
      habit of skipping `npm run typecheck` on the grounds that lint and build cover it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2812)

---

**Companion PR:** paranext/paratext-bible-internal-extensions#202 hides "Manage extensions" in
Simple. No merge-order constraint — see the Overview.

### DIFF
--- a/assets/localization/en.json
+++ b/assets/localization/en.json
@@ -118,6 +118,7 @@
   "%mainMenu_file%": "File",
   "%mainMenu_help%": "Help",
   "%mainMenu_helpInfo_showTour%": "Show the tour",
+  "%mainMenu_helpInfo_visitCommunitySupportPage%": "Community support",
   "%mainMenu_helpInfo_visitFAQsPage%": "FAQs",
   "%mainMenu_helpInfo_visitFeatureRoadmapPage%": "Feature roadmap",
   "%mainMenu_helpInfo_visitGettingStartedPage%": "Getting started",
--- a/assets/localization/es.json
+++ b/assets/localization/es.json
@@ -234,6 +234,7 @@
   "%mainMenu_file%": "Archivo",
   "%mainMenu_help%": "Ayuda",
   "%mainMenu_helpInfo_showTour%": "Mostrar el recorrido",
+  "%mainMenu_helpInfo_visitCommunitySupportPage%": "Soporte de la comunidad",
   "%mainMenu_helpInfo_visitFAQsPage%": "Preguntas frecuentes",
   "%mainMenu_helpInfo_visitFeatureRoadmapPage%": "Hoja de ruta de características",
   "%mainMenu_helpInfo_visitGettingStartedPage%": "Primeros pasos",
--- a/assets/localization/metadata.json
+++ b/assets/localization/metadata.json
@@ -81,6 +81,13 @@
   "%general_open%": {
     "fallbackKey": "%Paratext.Base.CommonForms.SelectScrTextsForm.SelectScrTextsForm.Text%"
   },
+  "%mainMenu_helpInfo_visitFAQsPage%": {
+    "fallbackKey": "%mainMenu_helpInfo_visitCommunitySupportPage%",
+    "deprecationInfo": {
+      "date": "2026-09-10",
+      "message": "Relabeled to \"Community support\" so the name matches the Support.Bible destination the item has always opened. The new key is %mainMenu_helpInfo_visitCommunitySupportPage%."
+    }
+  },
   "%mainMenu_visitSupportBible%": {
     "fallbackKey": "%mainMenu_helpInfo_visitFAQsPage%"
   },
--- a/e2e-tests/tests/enhanced-resources/README.md
+++ b/e2e-tests/tests/enhanced-resources/README.md
@@ -49,11 +49,16 @@ Registered as a project in `playwright.config.ts` (`name: 'enhanced-resources'`)
 by `playwright-cdp.config.ts`, which testIgnores `smoke`, `isolated`, `_example`, and
 `manage-books` but not this directory. There is no dedicated npm script.
 
-Registered is not the same as runnable. These specs use `fixtures/cdp.fixture.ts`, so they attach to
-an already-running app and must go through the CDP config — `playwright.config.ts`'s `globalSetup`
-rejects the very app they need. They also need real Marble resources (e.g. `ESV16UK+`) that are not
-available in CI. Two specs are additionally quarantined from type-checking in
-`e2e-tests/tsconfig.json`.
+Registered is not the same as runnable. These specs use `er.fixture.ts` (a thin wrapper over
+`fixtures/cdp.fixture.ts`), so they attach to an already-running app and must go through the CDP
+config — `playwright.config.ts`'s `globalSetup` rejects the very app they need. They also need real
+Marble resources (e.g. `ESV16UK+`) that are not available in CI. Two specs are additionally
+quarantined from type-checking in `e2e-tests/tsconfig.json`.
+
+**Start the app in Power mode.** "Open enhanced resource" is hidden in Simple mode, so this whole
+suite requires it: `er.fixture.ts` declares `requiredInterfaceMode: 'power'`, and attaching to a
+Simple-mode app fails fast with that message rather than timing out on the menu. Set
+`'platform.interfaceMode': 'power'` in `dev-appdata/data/settings.json` before starting the app.
 
 **Size:** 15 spec files — 113 `test(...)` declarations and 45 `test.fixme(...)`. The `fixme` ones
 cannot run even in principle; they are disabled by their own authors.
--- a/e2e-tests/tests/enhanced-resources/cell-reorder.spec.ts
+++ b/e2e-tests/tests/enhanced-resources/cell-reorder.spec.ts
@@ -24,7 +24,7 @@
  * No new fixtures or helpers are introduced.
  */
 import { FrameLocator } from '@playwright/test';
-import { test, expect } from '../../fixtures/cdp.fixture';
+import { test, expect } from './er.fixture';
 import { waitForAppReady } from '../../fixtures/helpers';
 import {
   closeAllNonHomeDockTabs,
--- a/e2e-tests/tests/enhanced-resources/enhanced-resources-functional-UI-PKG-001.spec.ts
+++ b/e2e-tests/tests/enhanced-resources/enhanced-resources-functional-UI-PKG-001.spec.ts
@@ -31,7 +31,7 @@
  *
  * Frame: WebView content is hosted in `iframe[title="Enhanced Resource"]` per Test Contract.
  */
-import { test, expect } from '../../fixtures/cdp.fixture';
+import { test, expect } from './er.fixture';
 import { waitForAppReady } from '../../fixtures/helpers';
 import { closeAllNonHomeDockTabs, openEnhancedResource } from './test-helpers';
 
--- a/e2e-tests/tests/enhanced-resources/enhanced-resources-functional-UI-PKG-002.spec.ts
+++ b/e2e-tests/tests/enhanced-resources/enhanced-resources-functional-UI-PKG-002.spec.ts
@@ -1,4 +1,4 @@
-import { test, expect } from '../../fixtures/cdp.fixture';
+import { test, expect } from './er.fixture';
 import { waitForAppReady } from '../../fixtures/helpers';
 import { closeAllNonHomeDockTabs, openEnhancedResource } from './test-helpers';
 
--- a/e2e-tests/tests/enhanced-resources/enhanced-resources-functional-UI-PKG-003.spec.ts
+++ b/e2e-tests/tests/enhanced-resources/enhanced-resources-functional-UI-PKG-003.spec.ts
@@ -1,4 +1,4 @@
-import { test, expect } from '../../fixtures/cdp.fixture';
+import { test, expect } from './er.fixture';
 import { waitForAppReady } from '../../fixtures/helpers';
 import { closeAllNonHomeDockTabs, openEnhancedResource } from './test-helpers';
 
--- a/e2e-tests/tests/enhanced-resources/enhanced-resources-functional-UI-PKG-004.spec.ts
+++ b/e2e-tests/tests/enhanced-resources/enhanced-resources-functional-UI-PKG-004.spec.ts
@@ -44,7 +44,7 @@
  * The test bodies are complete - component-builder activates them by removing `.fixme`.
  */
 
-import { test, expect } from '../../fixtures/cdp.fixture';
+import { test, expect } from './er.fixture';
 import { waitForAppReady } from '../../fixtures/helpers';
 import { closeAllNonHomeDockTabs, openEnhancedResource } from './test-helpers';
 
--- a/e2e-tests/tests/enhanced-resources/enhanced-resources-functional-UI-PKG-005.spec.ts
+++ b/e2e-tests/tests/enhanced-resources/enhanced-resources-functional-UI-PKG-005.spec.ts
@@ -25,7 +25,7 @@
  * - Uses `cdp.fixture` only (NOT `papi.fixture` or `app.fixture`).
  * - Navigates via visible UI; no `sendPapiCommand`.
  */
-import { test, expect } from '../../fixtures/cdp.fixture';
+import { test, expect } from './er.fixture';
 import { waitForAppReady } from '../../fixtures/helpers';
 import { closeAllNonHomeDockTabs, openEnhancedResource } from './test-helpers';
 
--- a/e2e-tests/tests/enhanced-resources/enhanced-resources-functional-UI-PKG-006.spec.ts
+++ b/e2e-tests/tests/enhanced-resources/enhanced-resources-functional-UI-PKG-006.spec.ts
@@ -1,4 +1,4 @@
-import { test, expect } from '../../fixtures/cdp.fixture';
+import { test, expect } from './er.fixture';
 import { waitForAppReady } from '../../fixtures/helpers';
 import { closeAllNonHomeDockTabs, openEnhancedResource } from './test-helpers';
 
--- a/e2e-tests/tests/enhanced-resources/enhanced-resources-functional-UI-PKG-007.spec.ts
+++ b/e2e-tests/tests/enhanced-resources/enhanced-resources-functional-UI-PKG-007.spec.ts
@@ -28,7 +28,7 @@
  * rows -> SDV filtered-by-domain dialog (entry path).
  */
 
-import { test, expect } from '../../fixtures/cdp.fixture';
+import { test, expect } from './er.fixture';
 import { waitForAppReady } from '../../fixtures/helpers';
 import { closeAllNonHomeDockTabs, openEnhancedResource } from './test-helpers';
 
--- a/e2e-tests/tests/enhanced-resources/enhanced-resources-functional-UI-PKG-008.spec.ts
+++ b/e2e-tests/tests/enhanced-resources/enhanced-resources-functional-UI-PKG-008.spec.ts
@@ -1,4 +1,4 @@
-import { test, expect } from '../../fixtures/cdp.fixture';
+import { test, expect } from './er.fixture';
 import { waitForAppReady } from '../../fixtures/helpers';
 import {
   closeAllNonHomeDockTabs,
--- a/e2e-tests/tests/enhanced-resources/enhanced-resources-image-render.spec.ts
+++ b/e2e-tests/tests/enhanced-resources/enhanced-resources-image-render.spec.ts
@@ -17,7 +17,7 @@
  *
  * Frame: WebView content is hosted in `iframe[title="Enhanced Resource"]`.
  */
-import { test, expect } from '../../fixtures/cdp.fixture';
+import { test, expect } from './er.fixture';
 import { waitForAppReady } from '../../fixtures/helpers';
 import { closeAllNonHomeDockTabs, openEnhancedResource } from './test-helpers';
 
--- a/e2e-tests/tests/enhanced-resources/enhanced-resources-journey.spec.ts
+++ b/e2e-tests/tests/enhanced-resources/enhanced-resources-journey.spec.ts
@@ -8,7 +8,7 @@
  *
  * Rules (do NOT relax during verify):
  *
- * - Import only from '../../fixtures/cdp.fixture' — never papi.fixture
+ * - Import only from './er.fixture' (this suite's cdp.fixture wrapper) — never papi.fixture
  * - Drive everything through visible UI (menu, button, keyboard) — no sendPapiCommand or direct
  *   JSON-RPC calls
  * - Selectors come from each WP's ui-spec Test Contract; do not invent new test ids during verify,
@@ -20,7 +20,7 @@
  * propagation, tab-switch propagation, clear-filter FN-021 — Dictionary domain row → SDV
  * filtered-by-domain dialog FN-022 — Sense-level occurrences link with descriptive tooltip
  */
-import { test, expect } from '../../fixtures/cdp.fixture';
+import { test, expect } from './er.fixture';
 import { waitForAppReady } from '../../fixtures/helpers';
 import { closeAllNonHomeDockTabs } from './test-helpers';
 
--- a/e2e-tests/tests/enhanced-resources/er.fixture.ts
+++ b/e2e-tests/tests/enhanced-resources/er.fixture.ts
@@ -1,0 +1,18 @@
+/**
+ * Enhanced Resources fixture — `cdp.fixture` with the interface mode this suite requires.
+ *
+ * "Open enhanced resource" is hidden in Simple mode, so every spec here needs an app running in
+ * Power mode. Declaring it centrally makes a wrongly-started app fail with the fixture's explicit
+ * mode message instead of a timeout on a menu item Simple never renders.
+ *
+ * Attach mode inherits whatever app was started and cannot change the mode itself, which is why
+ * this asserts rather than seeds: start the app with `'platform.interfaceMode': 'power'` in
+ * `dev-appdata/data/settings.json`.
+ */
+import { test as cdpTest } from '../../fixtures/cdp.fixture';
+
+export { expect } from '../../fixtures/cdp.fixture';
+
+export const test = cdpTest.extend({
+  requiredInterfaceMode: ['power', { option: true }],
+});
--- a/e2e-tests/tests/enhanced-resources/scripture-text-grid-view-options.spec.ts
+++ b/e2e-tests/tests/enhanced-resources/scripture-text-grid-view-options.spec.ts
@@ -16,7 +16,7 @@
  * admin entry stays in the top section) persists.
  */
 import { Page } from '@playwright/test';
-import { test, expect } from '../../fixtures/cdp.fixture';
+import { test, expect } from './er.fixture';
 import { waitForAppReady } from '../../fixtures/helpers';
 import {
   chapterViewOption,
--- a/e2e-tests/tests/enhanced-resources/scripture-text-grid-zoom.spec.ts
+++ b/e2e-tests/tests/enhanced-resources/scripture-text-grid-zoom.spec.ts
@@ -22,7 +22,7 @@
  * The zoom CSS property is applied inline (`style="zoom: 1.1"`) on the content wrapper div. In a
  * real Chromium (unlike jsdom), `locator('[style*="zoom"]')` reliably matches this element.
  */
-import { test, expect } from '../../fixtures/cdp.fixture';
+import { test, expect } from './er.fixture';
 import { waitForAppReady } from '../../fixtures/helpers';
 import {
   closeAllNonHomeDockTabs,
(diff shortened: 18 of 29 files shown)


jolierabideau (changes requested):
## Review summary

Reviewed against [PT-4533](https://paratextstudio.atlassian.net/browse/PT-4533), focusing on DoD coverage and pattern conformance.

The core approach is right and I want to say so plainly: this consumes the existing `hiddenInterfaceModes` / `filterItemsForInterfaceMode` axis rather than inventing a gate, adds no production code, and follows the documented two-step localized-string replacement protocol. The test diff is substantial and — where I mutation-checked it — genuinely load-bearing. Most of what follows is small.

Two things I'd want resolved before merge, then a set of minor pattern notes.

---

### 1. The one guard the ticket explicitly asked for isn't there

PT-4533's testing ideas asked to *"snapshot the Simple main-menu and Help-menu item lists so a future extension adding an item without `hiddenInterfaceModes` fails loudly rather than appearing in Simple."*

`KEPT_IN_SIMPLE` (`menu-data.service-host.contributions.test.ts:137`) is presence-only — it asserts seven commands *are* in Simple, never that the list is *exactly* those. Verified empirically: appending an unflagged item to the extensible `platform.projectProjects` group leaves the whole suite green while the item renders in Simple. `extensions/src/hello-rock3` already contributes four unflagged main-menu items through that same door.

That's the gap the snapshot was meant to close — this PR prunes today's menu, but nothing stops tomorrow's extension from un-pruning it. A `toEqual` on a sorted command list per column should cover it.

### 2. The stated merge-ordering dependency looks unfounded

The description says the companion `paratext-bible-internal-extensions` PR "must land after this one — against an older core, ajv rejects the whole document."

But `hiddenInterfaceModes` entered the schema in `8843aa374c4` (#2602, 2026-07-28) and lives at `menus.model.ts:73` / `:458-464` — **unchanged by this PR**. This branch adds no schema surface at all, so any core at or after #2602 already validates the companion's document. Did you observe the ajv rejection against a specific older build? If not, the two PRs are independent and the ordering constraint can come out of both the description and the review-focus list.

---

### Pattern notes

**`TODO(PT-4545)` is shipped inside a translator-facing field.** `localizeNotes` is documented as the localizer's menu-path note (`Entry-Point-Guide.md:106`, `Extension-Development-Guide.md:330`). A repo-wide grep over every `menus.json` returns exactly one `TODO` in a `localizeNotes` value — this one (`platform-lexical-tools/contributions/menus.json:9`). The Simple-mode *rationale* has clear precedent (`platform-scripture-editor/contributions/menus.json:76`); a dev TODO with an open ticket doesn't, and a localizer can't act on it. Worth moving to a sibling code comment or the ticket.

**Same line: the menu path itself is now stale.** It still opens `"Scripture Editor top menu > Project > …"`, but this is a `mainMenu` item — `defaultWebViewTopMenu.items` in that file is empty, and your own test names it under the application main menu. Since the string is being rewritten anyway, the path is the one part translators actually rely on.

**The `fallbackKey` redirect can never fire as written.** `getLocalizedString` returns `initialLanguageData[localizeKey]` and only reaches `findLocalizationForFallbackLanguageAndOrKey` on a miss. Because `%mainMenu_helpInfo_visitFAQsPage%` deliberately keeps its values (`en.json:122`, `es.json:238`, pinned by `menu.data.test.ts`), an out-of-repo consumer holding the old key keeps rendering "FAQs" — the opposite of what the guide's "old code falls back to the new key" bullet promises. The repo's only comparable rename, the four `%tab_contextMenu_moveTabToNewWindow_*%` entries (`metadata.json:165-181`), removed the old values so the redirect takes effect *and* paired `fallbackKey` with `deprecationInfo` so `deprecatedStringsByKey` warns. Was inertness the intent here? If so it's worth a note, because it reads as the standard protocol while behaving differently.

**Related:** `%mainMenu_visitSupportBible%` is now a two-hop chain (`→ FAQs → CommunitySupport`). `findLocalizationForFallbackLanguageAndOrKey` (`localization.service-host.ts:245-278`) reads `metadata.fallbackKey` once and never re-enters `getMetadata`, so chains aren't followed. Harmless today; a one-line repoint closes it permanently.

**Field placement is inconsistent within `menu.data.json`.** The three new entries put `hiddenInterfaceModes` after `label` (`:53`, `:70`, `:100`); the pre-existing `platform.createWindow` puts it last. Yours matches the extension precedent and is the better of the two — worth moving `createWindow` up rather than following it, so the file isn't carrying both orders.

**`getUnlocalizedMainMenu` (the macOS native menubar) isn't asserted.** It applies the same filter via the shared helper, so this is structurally safe and low risk — but the whole suite goes through `getMainMenu`, so nothing would catch a future change that filters one path and not the other. One case would close it.

**`e2e-tests/CLAUDE.md` is now slightly wrong.** It names `comment.fixture` as "the one exception" for the inert-plain-value-fixture trap; `enhanced-resources.fixture.ts:20` is now a second instance. The docblock and README handle it by convention, which is right — the CLAUDE.md sentence just needs the other name added.

**Minor:** `platform-bible-menu.commands.test.ts:42` pins `'https://studio.paratext.org/start '` with its trailing space. Pre-existing, not yours — but this test is what turns it into an asserted expectation. Either strip it in both places or note it as a known typo.

---

### Questions for the team, not for this diff

- **Simple's Help menu has 5 items, not the DoD's 4.** "Show the tour" (`menu.data.json:78`) carries no `hiddenInterfaceModes`. I gather that's deliberate, but it's against the transcribed keep-list — needs Todd's sign-off rather than a code change.
- **The "all six locale files" DoD line rests on a wrong premise.** `fr`, `km`, `zh-hans` and `zh-hant` carry no main-menu keys at all — zero hits for either key at PR head. Only `en`/`es` could carry it and both do. The ticket line should be corrected; the code is right.
- **The Dictionary has no Simple entry point at all** until PT-4545. Honestly recorded, but worth confirming product accepts the gap rather than a deferral — and the PRD contradicts itself here (NN-1.1 removes it, NTH-5.2 asks to keep one).
- **PT-4503** — the ticket said "file or reference it, don't silently absorb it," and the PR doesn't mention it. The live re-render bridge is already on `main`, so this looks already-handled rather than absorbed; a one-line confirmation would close it.

---

### What's good

- Both halves of every gate are asserted — hidden-in-Simple *and* present-in-Power — which is what catches an over-broad flag. I mutation-checked three of the seven and each turned the right test red for the right reason.
- The disk scan reads the source tree rather than `dist`, so it's immune to the stale-`dist` false-green class, and it has a real discovery guard that fires loudly.
- Tests sit at four complementary levels, and the docstrings explaining *why* the Power half and the discovery guard exist are accurate claims about the design — I confirmed both by mutation rather than taking them on faith.
- The `Entry-Point-Guide.md` edit fixes a pre-existing inaccuracy and points at a grep rather than a single file that goes stale on the next change.

captaincrazybro:
Thanks — this was a genuinely useful review, and two of the findings changed the shipped behaviour rather than just the tests. All addressed in 48127ae4356 unless noted.

## Fixed

**1. The snapshot guard.** You were right that `KEPT_IN_SIMPLE` couldn't see an item nobody listed. Added exact-match pins for the whole Simple *and* Power main menus. Verified the way you described: appending an unflagged item to an extensible group now turns the Simple pin red and names the new command in the diff — the presence-only checks stayed green in that same scenario.

Worth noting what the pin exposed: `helloRock3`'s submenu and its three items **are in Simple today**. They're in the pinned list deliberately, as a live example of the hazard rather than an oversight.

**2. The `fallbackKey` redirect was inert — and it's now fixed properly.** Confirmed your mechanism: `getLocalizedString` returns `initialLanguageData[localizeKey]` and only reaches `findLocalizationForFallbackLanguageAndOrKey` on a miss, so a retired key that keeps its value resolves to the old label forever.

Inertness was *not* the intent. Followed the `moveTabToNewWindow` precedent fully — removed the retired key's `en`/`es` values so the redirect fires, and restored `deprecationInfo` so `deprecatedStringsByKey` warns. The test now asserts the key is *absent* from both locales and explains why that absence is load-bearing.

**3. `%mainMenu_visitSupportBible%`** repointed at the live key. Removing the retired values turned that from a latent two-hop into a pointer at nothing, so it was no longer deferrable.

**4. `TODO(PT-4545)` removed** from `localizeNotes`. A repo-wide grep now returns zero TODOs in any `localizeNotes`; the deferral is recorded on the ticket and in the PR body.

**5. Same line — the stale menu path.** Corrected to `Application main menu > Platform > …`. It was wrong before this PR too, but the line was being rewritten anyway. Scoped to this one file: the sibling notes disagree with each other (`> Project >` vs `> Platform.bible >`) and that drift is out of scope here.

**6. The merge-ordering claim was unfounded** — you're right, and the sibling reviewer's `merge-base` check confirms it. `8843aa374c4` is already on `main` and this PR adds no schema surface. Corrected in both PR descriptions and dropped from the review-focus list. Thanks for digging into it rather than taking the description at face value.

**7. `e2e-tests/CLAUDE.md`** now names `enhanced-resources.fixture.ts` as the second inert-plain-value instance, with the reason and the escape hatch.

## Two corrections

**`getUnlocalizedMainMenu` *is* asserted.** `menu-data.service-host.test.tsx:358` — "excludes hiddenInterfaceModes items when platform.interfaceMode is simple, matching getMainMenu (macOS menubar consistency)" — and it's pre-existing on `main`, not something this PR left out. Your narrower point stands (it drives mock data rather than the shipped document), but the one-path-filtered-not-the-other regression would fail it.

**PT-4503 is not already handled.** Verified against a freshly fetched `origin/main` on 2026-09-16: `invalidateLocalizedOutput` has **0 hits** repo-wide, `menu-data.service-host.ts` has **0** `interfaceLanguage` references, and the identity cache the ticket blames is still at `menu-document-combiner.ts:253`. No commit in the fetched history mentions it. It's still in 🆕 Triage and genuinely open — worth flagging so nobody closes it as done. PT-4533's "file or reference it" is satisfied by the ticket existing; I've added an explicit out-of-scope line to the description.

## Not changing

**Field placement.** Leaving `platform.createWindow`'s flag where it is. Agreed the new placement is better, but it's a pre-existing line this PR doesn't otherwise touch, and key order is cosmetic.

**The trailing space** in `studio.paratext.org/start ` — as you say, pre-existing. Leaving both the source and the assertion alone rather than widening scope.

## Team questions

Agreed on all four, and they're in the PR's review-focus list: "Show the tour" makes Simple's Help menu 5 items and needs Todd's sign-off; the "all six locale files" DoD line rests on a wrong premise and should be corrected on the ticket; the Dictionary gap needs product confirmation, with the PRD self-contradiction (NN-1.1 vs NTH-5.2) called out.

One more thing your review surfaced indirectly: the branch is 16 commits behind `main`. Holding the rebase until approval.

jolierabideau (approved):
Approving — every finding is resolved, and the two you pushed back on you were right about.

I re-verified rather than taking the reply at face value. The snapshot pin is the one that mattered most: I mutation-checked it by appending an unflagged item to the extensible `platform.projectProjects` group, and both the Simple and Power pins turn red and name the new command, while every presence-only check stays green. That's exactly the gap closed. The `fallbackKey` redirect now genuinely fires — the retired key's `en`/`es` values are gone and `deprecationInfo` is back — and `%mainMenu_visitSupportBible%` points straight at the live key with no chain left.

On your two corrections: `getUnlocalizedMainMenu` is indeed asserted at `menu-data.service-host.test.tsx:358` and pre-existing on `main`, and PT-4503 is genuinely still open — I confirmed 0 repo-wide hits for `invalidateLocalizedOutput` and 0 `interfaceLanguage` references in `menu-data.service-host.ts`. Both of my reads were wrong; thanks for checking them instead of just implementing.

Agreed on both scope calls (`createWindow` field placement and the trailing space) — pre-existing lines this PR doesn't otherwise touch.

The four team questions are ticket and product work rather than merge blockers; I've recorded them on PT-4533 so they don't get lost when this merges.

Nice work on this one.

PR #2814, opened 2026-09-14
PT-4615: Declare bold and italic faces for the Scripture fonts

tjcouch-sil:
## Summary

Bold and italic Scripture text lost Charis SIL and fell back to DejaVu Sans on Linux. `src/renderer/styles/fonts.css` declared every family at `font-weight: 400; font-style: normal` only, so there was no bold or italic face to use.

### Why review this

Part of the Standard view work ([PT-4186](https://paratextstudio.atlassian.net/browse/PT-4186)) — `\add` and the other bold/italic markers are the most visible case, and the marker glyph (including the backslash) renders in a different typeface from the text around it.

### Root cause

Charis SIL was unique among these families in *also* carrying `local()`-only `@font-face` rules, from the editor's vendored `_usj-nodes.scss`. On a machine where the font is not installed, those four faces are `status: error`, and a bold request matches the erroring `local('Charis SIL Bold')` face exactly. A family whose best-matching face fails to load does **not** fall back to another face of that family — it falls out of the family. Regular text kept matching the (loadable) 400 webfont face, so only bold and italic broke.

It landed on a *sans* font because scripture-editors' `generateUsjCss` emits `.editor-input.usfm { font-family: "<project font>" }` with no fallback list. At (0,2,0) that ties `.usfm.formatted-font`'s curated chain and wins on injection order, leaving nothing to catch the fallback. That half is fixed in [paranext/scripture-editors#12](https://github.com/paranext/scripture-editors/pull/12), which publishes the chain as a `--usj-font-fallback` custom property and has the generator emit `font-family: "<project font>", var(--usj-font-fallback, serif)`; it takes effect here on the next editor release, and this PR carries the stylesheet side so the property is in place when it does.

Windows was never affected: `C:\Windows\Fonts` has all four `CharisSIL-*.ttf`, so the `local()` faces resolve.

## Changes

- `fonts.css`: added the missing weight/style groups for Cardo, Charis SIL and Gentium Plus, and switched Noto Sans Syriac to its `100 900` variable faces so a bold Syriac run gets a real weight instead of a synthesized one. Google Sans (app chrome) gains no faces of its own. The verbatim region is refreshed to what its URL returns today (15 September 2026) — only Google Sans moved; Cardo and Noto Sans Syriac are byte-identical. Both `@import` URLs name the `ital,wght` axes that actually regenerate their region.
- **The font-face area is now two regions with two URLs.** The first — Cardo, Google Sans, Noto Sans Syriac — is pasted verbatim and never hand-edited. The second — Charis SIL and Gentium Plus, the fonts Paratext installs — is pasted and then given a `local()` source ahead of each `url()`, so an installed copy is used as-is instead of being downloaded and, more to the point, so those families still render with no network at all. A family whose faces all fail to load does *not* fall back to an installed copy of the same name — it drops out of the family (measured; see Testing) — and a `local()` inside the face's own `src` is the only thing that reaches the installed copy while keeping one family name, which is what project font selection needs. The region's comment spells out the weight/style table, and `fonts.test.ts` checks every face against it.
- Splitting the URL is faithful: fetched separately, the two URLs return byte-identical faces to the combined one for all four Scripture families.
- Removed the four `local()`-only Charis SIL `@font-face` rules from all three vendored copies of the editor stylesheet (`platform-scripture-editor`, `platform-enhanced-resources`, and the `platform-bible-react` demo), so the fix stands on its own instead of resting on Chromium recovering from a face that failed to load. Each site keeps a comment saying why not to re-add them.
- **A third region: installed-only faces for what the downloads leave out.** Google serves Charis SIL as five subsets with **no Greek at all**, and both families with 8 of the 112 combining diacritical marks. The vendored rules this PR removes carried no `unicode-range`, so wherever the complete font is installed they were the only face covering the rest — without them a Greek quotation in a Charis SIL project renders in Gentium Plus while the Latin beside it stays Charis SIL. Neither family is served above the BMP either, where the installed SIL builds carry the Latin Extended-F and G phonetic blocks — 52 astral codepoints in Google's own build, read off the `cmap`. Per family and style, one `local()`-only face whose `unicode-range` is the complement of the downloaded ranges over the whole of Unicode: disjoint by construction (the highest upper bound anywhere in the downloaded region is U+FFFD for both families), so an erroring face there can never shadow a downloadable one, and where the font is not installed it fails and the character falls through as before.
- **`--scripture-font-stack`**: the Scripture stack is now named as well as applied, and all three vendored stylesheets point `--usj-font-fallback` at it. A project font that cannot be rendered therefore falls back through Platform.Bible's own deliberately ordered stack rather than the editor library's chain — the library chain stays as the `var()` fallback and is what applies where fonts.css is not loaded, such as Storybook.
- `platform-enhanced-resources`' font block is now identical to `platform-scripture-editor`'s, which its header has always said to keep it in sync with, so all three stylesheets agree and its Greek no longer drops to the generic `serif`. Its `.formatted-font .usfm` selector, which predates this work, is the one divergence left.
- `lib/platform-bible-react/.storybook/scripture-fonts.css`: Storybook's copy of the Charis SIL faces — both the downloaded, `local()`-first ones and the installed-only complement, copied from `fonts.css` verbatim rather than `@import`ed from Google, so a network-restricted run (`@storybook/addon-vitest` runs stories as tests) keeps them instead of silently losing them. **Both** `.storybook/preview.ts` files load it. There are two Storybooks: the root one globs `src/**` and `extensions/src/**`, and five story files under those globs reach the demo `usj-nodes.css` — the two ten-layout demos through `ten-layout-shared.tsx`, the model text panel, and both Enhanced Resources panes — while `platform-bible-react`'s own reaches it from the footnote stories and the scripture-editor story's stylesheet. Neither Storybook loads the application's `fonts.css`: it is imported only as `?raw`, into a web view's srcdoc.
- `fonts.test.ts` (new): fails if a family is missing a required weight/style; if a style does not cover the same `unicode-range` subsets as its regular face, or is served by the same font files as it (what a copied-but-not-repointed block looks like); if a family in the stack is in neither the required list nor an explicit, reasoned exemption list; if the verbatim region acquires a `local()`; if a face in the second region is missing its `local()` sources or names the wrong weight/style; if a family appears in both regions; if a vendored stylesheet declares a face again or stops pointing at the stack; if Storybook's copy of the Charis SIL faces stops matching `fonts.css` face for face, in either direction; if that copy acquires an `@import`, which would satisfy the face comparison while putting the faces back behind a live request; or if either Storybook preview stops loading it. Variable-weight families are detected from the stylesheet and get a weight-range assertion in place of the comparisons that would be vacuous for them.

The file grows 21 KB → 49 KB and is inlined into every web view's srcdoc; the font *files* are still fetched lazily per `unicode-range` subset, so this adds no requests for text that does not use them.

## Testing

- [x] `fonts.test.ts` red (8/10 failing) before the change, green (29/29) after. Every guard checked falsifiable by mutation — a family added to the stack with no faces, a bold block repointed at the regular font file, a `local()` pair dropped from a face, a face given the wrong style's `local()` names, a `local()` added to the verbatim region, a Charis SIL face left behind in the verbatim region, a narrowed variable weight range, a `local()`-only face re-added to a vendored stylesheet, a vendored stylesheet reverted to a literal chain, one installed-only face narrowed back to the BMP, a downloaded range extended into the complement, an unreadable `unicode-range` part, a face narrowed, deleted or given a renamed `local()` in Storybook's copy, an `@import` appended to it, a Charis SIL face widened or added in `fonts.css` alone, and the root Storybook's import removed each fail the suite.
- [x] **Linux/WSL, running editor**: `CSS.getPlatformFontsForNode` inside the editor web view reported `DejaVu Sans` / `DejaVuSans-Bold` for bold, italic and bold-italic; it now reports `CharisSIL-Bold`, `CharisSIL-Italic` and `CharisSIL-BoldItalic`. Confirmed visually in Standard view — `\mt1`, `\mt3` and `\c 1` go from sans-serif to Charis bold.
- [x] **Windows**: same probe under Edge, measured before and after, is byte-identical (615.6 / 659.8 / 578.8 / 628.8 px) — the installed faces still win, so no regression and no new network dependency.
- [x] **Windows, fonts unreachable** (both fonts installed, `gstatic` URLs repointed at an invalid host, probe-string widths at 100px). `url()`-only faces measure exactly the control width for an unresolvable family name at every style — the family drops out rather than falling back to the installed copy: `2937.61 / 3088.48 / 2880.86 / 2966.08` for regular / bold / italic / bold-italic, against `3210.02 / 3432.97 / 3035.55 / 3297.81` for the installed face. With `local()` in the face's `src` it measures the installed face, both offline and online.
- [x] **End to end**, assembling `fonts.css` + the compiled vendored stylesheet + the rule `generateUsjCss` will emit, on an element classed `editor-input usfm formatted-font scripture-font`, with both fonts installed:

  | case | renders |
  | --- | --- |
  | offline, project font Charis SIL | Charis SIL |
  | offline, project font Gentium Plus | Gentium Plus |
  | offline, project font unavailable | Gentium Plus (first in the stack) |
  | offline, no project stylesheet | Gentium Plus |
  | online, project font Charis SIL | Charis SIL, from the installed copy — no download |
  | online, project font Charis SIL, `local()` targets renamed to something absent | Charis SIL, downloaded — the original fix still works |
  | `fonts.css` absent (the Storybook case) | the library chain's Charis SIL |
- [x] **Per-character font resolution** (Windows Edge, both fonts installed, identical markup throughout): a bare α goes from 105.078125 px — byte-identical to installed Gentium Plus — to 129.796875 px, installed Charis SIL. Polytonic Greek (`ἐ`, `ῇ`) stays Gentium Plus: forced to Charis SIL it measures the same as forced to generic `serif`, so the installed Charis SIL has no glyphs there and the face correctly falls through. Hebrew and Syriac are byte-identical before and after (126.953125 and 186), still served by Cardo and Noto Sans Syriac. Combining marks are zero-advance, so width cannot distinguish them — that half rests on the ranges and the test, not on a measurement.
- [x] **Both Storybooks**, built and served, with neither font installed on the machine — CI builds neither (`publish-docs` runs on main, `chromatic` is label-gated), so this is checked by hand. 24 Charis SIL `@font-face` rules reach a story document (20 downloaded subsets + 4 installed-only complement); the Scripture element computes to the library chain, as intended, since `--scripture-font-stack` lives in `fonts.css`; and probe-string widths at 100px over `…modeltextpanel--active` and `…scripturepane--with-scripture` give four distinct values for Charis SIL (1290 / 1393 / 1193 / 1303 for regular / bold / italic / bold-italic) against an unresolvable family name whose italic is byte-identical to its regular (1493.31 / 1652.50 / 1493.31 / 1652.50) — real faces against a synthesized oblique. The complement faces report `status: error` there, as they must with the font absent, and the widths are measured with them in the document: the disjointness holding.
- [x] Gentium Plus bold and italic verified live as `GentiumPlus-Bold` / `GentiumPlus-Italic`; Noto Sans Syriac still resolves to `NotoSansSyriac`, not a fallback.
- [x] `prettier`, `stylelint`, `eslint`, `vitest src/renderer/styles` clean, and both edited `.scss` copies compile under `sass` with nothing on stderr. The two vendored-stylesheet pin suites still pass unchanged (`platform-scripture-editor` 20/20, `platform-bible-react` demo 26/26).
- `npm run typecheck` / `npm run lint` fail in my worktree only on pre-existing and worktree-setup issues (`spdx-license-list/full`, which fails identically on an untouched checkout, plus symlinked-`node_modules` resolution artifacts). Neither references the changed files.

### One deliberate behaviour change

A project with **no font setting** now renders in Gentium Plus where it rendered in Charis SIL. `.usfm.formatted-font` resolves through `--scripture-font-stack`, whose order is the application's stated preference for Scripture text, so the editor's default now matches `.scripture-font` everywhere else — that is the point of pointing it at the stack rather than a side effect of it. Storybook still leads with Charis SIL, deliberately: `--scripture-font-stack` lives in the application's `fonts.css`, and `platform-bible-react` has to render sensibly for consumers who never load it, so the `var()` fallback there is the editor library's own chain.

## AI Involvement

AI-assisted — [session 1](https://claude.ai/code/session_01ARpdMpPZBfbnU8fh4Q3VtZ), [session 2](https://claude.ai/code/session_01BhTeYiyPXkBu8fiTfUTCyH). Claude diagnosed the fallback, generated the `@font-face` blocks from Google Fonts, wrote the tests, and ran the browser measurements; I reviewed the diff and the verification evidence.

## Risk Level

Low — additive `@font-face` declarations in one stylesheet plus the removal of four rules that could only ever fail to load. The substantive behaviours (Noto Sans Syriac's variable faces, and Charis SIL and Gentium Plus resolving to an installed copy before downloading) are each verified live. The one deliberate change for online users is that those two families now render from an installed copy where there is one, rather than fetching the same font from Google.

[PT-4186]: https://paratextstudio.atlassian.net/browse/PT-4186?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2814)


src/renderer/styles/fonts.test.ts
```diff
@@ -0,0 +1,158 @@
...
+      const [baseline, ...rest] = REQUIRED_FACES[family];
+      const baselineSubsets = subsetsFor(fontFaces, family, baseline);
+      expect(baselineSubsets.length).toBeGreaterThan(0);
+      rest.forEach((required) => {
+        expect(subsetsFor(fontFaces, family, required)).toEqual(baselineSubsets);
+      });
+    },
+  );
+
+  it('guards every family the Scripture font stack relies on', () => {
```
lyonsil, line 519: **The family-coverage test only checks one direction, so it can't catch the bug this PR is fixing.**

The test walks `REQUIRED_FACES` (the hand-written list of families we know need bold/italic) and checks each one is present in the `.scripture-font` CSS stack. That's the "is everything on my list in the stack?" direction.

It never checks the other direction: "is everything in the stack on my list?" So when someone adds a new family to `.scripture-font` and pastes in only the regular (weight 400) `@font-face` blocks, nothing fails. That is exactly the shape of PT-4615 - the bug this PR exists to fix - and the guard named "guards every family the Scripture font stack relies on" would let it straight back in for the new script.

This isn't hypothetical. `'Google Sans'` is already in the `.scripture-font` stack today (`fonts.css:1121`), is not in `REQUIRED_FACES`, and is declared at weight 400 only. The test is green right now with a family in that exact state.

Suggested fix: parse the families out of the `.scripture-font` stack and assert each one appears in `REQUIRED_FACES`, with an explicit allow-list for any family we've deliberately decided doesn't need full coverage (and a comment saying why). Then the test does what its name promises.

src/renderer/styles/fonts.test.ts
```diff
@@ -0,0 +1,158 @@
...
+
+  it.each(Object.keys(REQUIRED_FACES))('declares every weight and style of %s', (family) => {
+    const missing = REQUIRED_FACES[family].filter(
+      (required) => subsetsFor(fontFaces, family, required).length === 0,
+    );
+    expect(missing).toEqual([]);
+  });
+
+  it.each(Object.keys(REQUIRED_FACES))(
+    'covers the same unicode-range subsets in every style of %s',
```
lyonsil, line 373: **The style-coverage test compares unicode ranges but never the actual font URLs, so a copy/paste mistake can pass.**

Background: a web font isn't one `@font-face` rule. Google Fonts splits each family into many blocks, one per unicode subset (latin, cyrillic, greek, and so on), and each block points at its own `.woff2` URL. This test checks that the bold blocks cover the same subsets as the regular blocks.

The gap: it compares only `unicodeRange`. It never looks at `src`. So consider the most likely way a human breaks this file - copying the regular block, editing `font-weight: 400` to `700`, and forgetting to update the URLs. Every subset still matches, the test passes, and bold silently renders as regular text on screen. That's a close sibling of the "half-pasted regeneration" mistake the test's own comment says it's guarding against.

`parseFontFaces` already captures `src`, so the fix is small: assert that the set of URLs used by the 400 blocks and the set used by the 700 blocks don't overlap.

Separately, worth knowing: for `Noto Sans Syriac` this assertion currently proves nothing. It's a variable font declared `font-weight: 100 900`, so `weightCovers('100 900', 400)` and `weightCovers('100 900', 700)` both select the same three faces. The test is comparing a set against itself and will pass no matter what. That family effectively has no style coverage here.

lyonsil (commented):
Three findings from a review of this PR. The CSS half checks out - I re-fetched the documented Google Fonts regeneration URL and diffed it face-by-face against the committed `fonts.css`, and all the newly added `fonts.gstatic.com` URLs resolve. The findings are in the new guard test and in the un-fixed root cause.

Two are inline on `src/renderer/styles/fonts.test.ts`. The third can't be an inline comment because the line it concerns isn't part of this diff, so it's here:

---

### `src/renderer/styles/fonts.css:1121` - low

**The broken `@font-face` rules the PR names as the root cause are still in the tree, in three copies - the fix works around them rather than removing them.**

Background: an `@font-face` rule can source a font from the user's machine (`local('Charis SIL')`) or from the network (`url(...)`). The rules this PR identifies as the cause have `local()` only - no URL - so on any machine without Charis SIL installed, that font face simply fails and text falls back to something else. That's the bug.

This PR adds correct web-font declarations to `fonts.css`, but the original `local()`-only rules are still present in three places:

- `extensions/src/platform-scripture-editor/src/_usj-nodes.scss:31-53`
- `extensions/src/platform-enhanced-resources/src/_usj-nodes.scss`
- `lib/platform-bible-react/src/components/demo/scripture-editor/usj-nodes.css`

Those styles are injected into the same document *after* `fonts.css` (which goes into the srcdoc `<head>`). So the document ends up with two competing sets of rules for one family, and the fix only holds because Chromium, on hitting a face whose source fails to load, falls back to the earlier working declaration of the same family. That is the precise browser behaviour the PR description asserts does *not* happen.

I probed it live and it does work today. Two reasons to tidy it anyway:

1. Nothing in the codebase pins that behaviour. It's an unwritten dependency on a browser implementation detail, and the PR description says the opposite - so the next person to read this has no way to know the fix rests on it.
2. `platform-bible-react`'s Storybook never loads `fonts.css`, so the broken rules there are still fully unmitigated.

Suggested fix: delete the four erroring rules from the vendored copies, which makes the fix stand on its own regardless of fallback behaviour. The pin in `usj-nodes-styles.test.ts` will need updating alongside.

---

src/renderer/styles/fonts.test.ts
tjcouch-sil: Fixed. The stack check runs both directions now: it parses the families out of `.scripture-font`, drops the CSS generics, and fails on any family that is in neither `REQUIRED_FACES` nor a new `UNGUARDED_STACK_FAMILIES` map, which carries a written reason per entry.

`'Google Sans'` is the one entry in that map. I left it uncovered rather than declaring its other faces: it is the app-chrome family, in the Scripture stack only as a last resort ahead of the generic `serif`, and it has exactly one face — so a bold or italic request against it is synthesized rather than dropped, because there is no competing face to fail. The entry says it needs full coverage if it ever gains a second face. If you would rather it simply came out of `.scripture-font` (a sans sitting ahead of the generic serif looks accidental to me), that is a rendering change I would rather make separately from this PR.

Checked falsifiable: adding `'Noto Serif Devanagari'` to the stack with no faces declared fails the test.

src/renderer/styles/fonts.test.ts
tjcouch-sil: Both fixed.

**URLs.** New test — `serves each style of %s from its own font files` — asserts the `url()` set for each required face and the regular face's set are disjoint. Repointing the Charis SIL 700/normal blocks at the 400 URL (matching subsets, wrong file) fails it.

**Noto Sans Syriac.** Right that it was comparing a set against itself. Variable-weight families are now detected from the stylesheet itself — a `font-weight` descriptor with two bounds — rather than listed by hand, and partitioned out of both the subset and URL comparisons, which are vacuous for them by construction. They get the assertion that is actually meaningful instead: the declared range spans every weight the family is used at. Narrowing `100 900` to `100 500` fails it. There is also a guard that both partitions are non-empty, so neither block of checks can quietly end up running over nothing.

tjcouch-sil:
Re the third finding (`fonts.css:1121`, the `local()`-only rules still in the tree):

Removed. All four rules are gone from all three vendored copies, each replaced by a comment saying why not to re-add them, and `fonts.test.ts` now fails if any of those files declares a face again.

One correction to the suggested fix, though — deleting them alone would have cost something real, and I only found it by measuring. Under Windows Edge with Charis SIL installed and fonts.css's `gstatic` URLs repointed at an invalid host (probe-string widths at 100px, compared against the installed face and against an unresolvable family name):

| case | regular | bold | italic | bold-italic |
| --- | --- | --- | --- | --- |
| installed font, no `@font-face` at all | 3210.02 | 3432.97 | 3035.55 | 3297.81 |
| fonts.css faces only, unreachable | 2937.61 | 3088.48 | 2880.86 | 2966.08 |
| fonts.css faces + the vendored `local()` rules, unreachable | 3210.02 | 3432.97 | 3035.55 | 3297.81 |
| **fonts.css faces with `local()` in `src`, unreachable** | **3210.02** | **3432.97** | **3035.55** | **3297.81** |
| unresolvable family name (control) | 2937.61 | 3088.48 | 2880.86 | 2966.08 |

Row 2 is identical to the control at every style: a family whose faces fail to load does not fall back to an installed copy of the same name either — it drops out of the family just as it does when a face errors. So those rules were quietly the only thing keeping an installed Charis SIL working with no network, which for this app is not an exotic case. (Aside, since it surprised me and it is why this is not better known: a *single* `@font-face` with no `unicode-range` does fall back to the installed copy. Once the family is split into `unicode-range` subsets — which is exactly what Google Fonts emits — it does not.)

So instead of only deleting, each Charis SIL face in fonts.css now names the installed copy ahead of the download, per weight and style:

```css
src:
  local('Charis SIL Bold'), local('CharisSIL-Bold'),
  url('https://fonts.gstatic.com/s/charissil/…') format('woff2');
```

That is row 4 — identical to the installed face offline, and it still falls through to the download where the font is not installed (which is the Linux case this PR started from, re-verified unchanged). The part your finding is about holds: this is a fallback *within one face's own `src` list*, which is specified behaviour, not Chromium recovering from a face that failed. Charis SIL only — it is the family Paratext installs and the only one that had an offline path to preserve; every other family here stays `url()`-only, and the file comment says so and says to re-add the `local()` lines after regenerating the region.

On Storybook: good point that it was unmitigated there. `scripture-editor.stories.css` now pulls the same four Charis SIL faces from Google Fonts directly, so the demo gets real bold and italic rather than synthesized ones.

On the pin: `usj-nodes-styles.test.ts` did not need updating — neither copy asserts anything about `@font-face`. Both suites still pass (20 and 26), as does the reworked `fonts.test.ts` (18).

Still not fixed here, and unchanged from the PR description: `generateUsjCss` emitting `font-family: "<project font>"` with no fallback list. That is a scripture-editors change plus an editor release, so it cannot ride along in this PR; I have a fix drafted and am checking with TJ on whether to open it now.

tjcouch-sil:
Follow-up to my note above: the `generateUsjCss` half is open now as [paranext/scripture-editors#12](https://github.com/paranext/scripture-editors/pull/12). The chain moves out of `generateUsjCss`'s reach and into `usj-nodes.css` as a `--usj-font-fallback` custom property, and the generator emits `font-family: "<project font>", var(--usj-font-fallback, serif)` at both of its emission sites — the base rule and every marker with a `fontName`, which had the same gap. That PR also drops the same four `local()`-only rules upstream, so the next vendoring sync does not bring them back.

The last commit here carries the matching stylesheet sync, so the property is in place the moment that lands in an editor release; until then nothing names it and the rules are equivalent to what they replace.

tjcouch-sil:
Correction to my measurement comment above: the `local()`-in-`src` approach it describes is gone. The downloaded region is supposed to be exactly what the `@import` URL returns with nothing hand-edited, so editing twenty faces in it was the wrong shape — and Gentium Plus, which Paratext installs as Charis SIL's successor, would have needed the same treatment.

The measurements still stand; only the fix changed. Installed copies now get their own families in a hand-written region of their own — `Charis SIL Local` and `Gentium Plus Local`, `local()`-only — placed at the **end** of the Scripture stack. Same result offline (the probe renders the installed face rather than the UA default), no edit to the paste, and both fonts covered. Putting them after every downloaded family rather than beside each one also means online rendering is bit-for-bit what it would be without them: a downloaded family always wins while one can be fetched.

The stack is also named now, as `--scripture-font-stack`, and the three vendored `usj-nodes` copies point `--usj-font-fallback` at it rather than carrying the editor library's chain — so a project font that cannot be rendered falls back through Platform.Bible's own order. The library chain stays as the `var()` fallback and is what applies where fonts.css is not loaded, such as Storybook.

`fonts.test.ts` gained guards for each of those: no `local()` in the downloaded region, alias families `local()`-only and covering every weight and style of the family they stand in for, aliases last in the stack, and each vendored stylesheet pointing at the stack. All mutation-checked. One behaviour change is called out in the PR description under "One known behaviour change".

tjcouch-sil:
Superseding my last two comments: the `*-Local` alias families are gone. Keeping the paste pristine that way cost project font selection — a project set to Charis SIL rendered in Gentium Plus offline, because its own font dropped out and the alias was a different family name, only reachable from the shared stack in the stack's order.

One family name is what makes that work, so `local()` has to sit inside the face's own `src`. The font-face area is now **two regions with two `@import` URLs**:

- **Verbatim** — Cardo, Google Sans, Noto Sans Syriac. Pasted whole, never hand-edited; a test fails if a `local()` shows up in it.
- **Installed-copy-first** — Charis SIL and Gentium Plus, the fonts Paratext installs. Pasted, then given a `local()` source ahead of each `url()`, per a weight/style table the region's comment spells out. `fonts.test.ts` checks every face against that table, so a missed or mistyped face fails the suite rather than surfacing at a translator's desk with no network.

Splitting the URL is faithful, which I checked rather than assumed: fetched separately, the two URLs return byte-identical faces to the combined one for all four Scripture families. (Google Sans has drifted upstream since the original paste — left alone, it is app chrome and unrelated.)

Re-measured end to end under Windows Edge with both fonts installed, assembling fonts.css + the compiled vendored stylesheet + the rule `generateUsjCss` emits:

| case | renders |
| --- | --- |
| offline, project font Charis SIL | Charis SIL |
| offline, project font Gentium Plus | Gentium Plus |
| offline, project font unavailable | Gentium Plus (first in the stack) |
| offline, no project stylesheet | Gentium Plus |
| online, project font Charis SIL | Charis SIL, from the installed copy — no download |
| online, same but with the `local()` targets renamed to something absent | Charis SIL, downloaded — the original fix is intact |
| `fonts.css` absent (Storybook) | the library chain's Charis SIL |

The "One known behaviour change" section is gone from the description with it. What remains deliberate for online users: those two families now render from an installed copy where there is one, instead of fetching the same font from Google.

src/renderer/styles/fonts.css
```diff
@@ -540,26 +563,736 @@
...
  *
- * `.scripture-font` applies the font to scripture text, which is used in both `platform-bible-react`
+ * <PostScript> is the family name without spaces (CharisSIL, GentiumPlus). `fonts.test.ts` checks
+ * every face against this table.
  */
+/* @import url('https://fonts.googleapis.com/css2?family=Charis+SIL:ital,wght@0,400;0,700;1,400;1,700&family=Gentium+Plus:ital,wght@0,400;0,700;1,400;1,700&display=swap'); */
+
+/* cyrillic-ext */
+@font-face {
+  font-family: 'Charis SIL';
```
lyonsil, line 587: **High - dropping the un-ranged `local()` faces narrows Charis SIL to Google's subsets, losing Greek and most combining diacritics.**

At this head, Charis SIL is declared in five subsets - `cyrillic-ext`, `cyrillic`, `vietnamese`, `latin-ext`, `latin` - and every one of its faces carries a `unicode-range`; there is no `greek`/`greek-ext` block (contrast Gentium Plus, which has both). The four rules this PR removes from the vendored `_usj-nodes.scss` copies had **no** `unicode-range`, so on any machine with Charis SIL installed - every Paratext Windows install, and the Linux case this PR is fixing once the font is installed - they were the only candidate face for everything outside those subsets: U+0370-03FF and U+1F00-1FFF, plus 104 of the 112 combining marks in U+0300-U+036F (the union of the declared ranges covers only U+0300-0301, U+0303-0304, U+0308-0309, U+0323 and U+0329).

After this change those codepoints match no Charis SIL face at all and fall out of the family. Concrete case: a project whose font is Charis SIL quoting Greek renders the Greek in Gentium Plus - or, until scripture-editors#12 ships and `generateUsjCss` emits the fallback, in the UA default - while the Latin beside it stays Charis SIL. In Enhanced Resources, whose rule is still `font-family: 'Charis SIL', serif`, the Greek drops all the way to the generic serif. Same for a transliteration run using U+0302/U+0330/U+0331.

`fonts.test.ts` cannot catch this: `covers the same unicode-range subsets in every style of %s` compares each style's subsets against the *regular* face's, so a family that is uniformly missing a script passes.

extensions/src/platform-scripture-editor/src/_usj-nodes.scss
```diff
@@ -28,31 +29,28 @@
...
+    'Times New Roman',
+    'Liberation Serif',
+    'DejaVu Serif',
+    Georgia,
+    serif
+  );
 }
 .usfm.formatted-font {
-  font-family: 'Charis SIL', 'Times New Roman', 'Liberation Serif', 'DejaVu Serif', Georgia, serif;
+  font-family: var(--usj-font-fallback);
```
lyonsil: **Medium - the editor's default Scripture typeface silently changes from Charis SIL to Gentium Plus.**

`.usfm.formatted-font` went from the literal chain starting `'Charis SIL'` to `var(--usj-font-fallback)`, which resolves to `--scripture-font-stack` = `'Gentium Plus', 'Charis SIL', ...`. Any project with no font setting - and, once the editor change lands, any project whose font is unavailable - now renders in Gentium Plus where it rendered in Charis SIL.

The Testing table records the unavailable-font case (`offline, project font unavailable` -> `Gentium Plus (first in the stack)`), but the no-setting case is a user-visible default change unrelated to the bold/italic fix, and it isn't called out as a behaviour change anywhere in the diff.

It also makes Storybook disagree with the app: there `--scripture-font-stack` is undefined, so the `var()` fallback applies and the demo still leads with Charis SIL.

lyonsil (commented):
Review of the `@font-face` changes at `6251c60`, focused on `unicode-range` coverage and the three vendored copies of the editor stylesheet.

Eight findings, all posted inline: one high, two medium, five low. Each comment states its severity at the top, and each was checked against the files at this head SHA rather than read off the diff alone.

extensions/src/platform-enhanced-resources/src/_usj-nodes.scss
```diff
@@ -23,28 +23,25 @@
...
+    'Charis SIL',
+    'Times New Roman',
+    'Liberation Serif',
+    'DejaVu Serif',
+    Georgia,
+    serif
+  );
 }
 .formatted-font .usfm {
   font-family: 'Charis SIL', serif;
```
lyonsil: **Medium - `--usj-font-fallback` is declared in this file but nothing reads it, and the rule that would have benefited was left literal.**

Line 35 sets `--usj-font-fallback`, but this extension consumes no generated project stylesheet and its own font rule is still `'Charis SIL', serif` right here. The property is dead in this file - and per the `fonts.css` finding, the family it names now has strictly narrower coverage than before, with only the generic `serif` behind it.

`fonts.test.ts`'s `hands the stack to the editor in %s` passes on the declaration at line 35, so the suite reports assurance for a file where the mechanism does nothing.

Also, unlike the other two copies, this file's header was not updated. It still reads only `COPIED FROM ... keep manually in sync`, with nothing recording that the four `@font-face` rules were dropped or that this rule now diverges from `platform-scripture-editor`'s.

src/renderer/styles/fonts.css
```diff
@@ -540,26 +563,736 @@
...
+  font-style: normal;
+  font-weight: 400;
+  font-display: swap;
+  src:
+    local('Charis SIL'),
```
lyonsil, line 592: **Low - putting `local()` first resolves per *face*, not per glyph, so a stale installed copy can shadow a webfont subset that would have worked.**

Each subset face prefers `local('Charis SIL')` / `local('CharisSIL-Regular')` over the `url()`. `local()` matching succeeds or fails on the whole font, not on the glyph: a user with an older Charis SIL that lacks a character inside a declared `unicode-range` - for example a Cyrillic or Vietnamese codepoint present in Google's v2 subset but not in their installed build - now gets the installed face selected, finds no glyph, and falls through to the next family. Before this PR the downloadable subset supplied it.

The Testing measurements were all taken against a current installed copy, so this case is outside what was verified.

lib/platform-bible-react/src/components/demo/scripture-editor/scripture-editor.stories.css
```diff
@@ -8,6 +8,9 @@
  * Copied from paranext-core/extensions/src/platform-scripture-editor/src/platform-scripture-editor.web-view.scss
  */
 
+/* Charis SIL with real bold and italic faces. The app inlines these from
```
lyonsil: **Low - this `@import` reaches only one of the three stories that use `usj-nodes.css`.**

`footnote-item.stories.tsx:4` and `footnote-list.stories.tsx:6` both `import '@/components/demo/scripture-editor/usj-nodes.css'` directly and never load `scripture-editor.stories.css`. With the vendored faces removed and the Google Fonts `@import` added only here, those two stories render with no Charis SIL declaration at all - so on a machine without the font installed (CI, Chromatic, most contributors) the footnote components fall to Times New Roman / Liberation Serif, including for the italic `\add`-style runs they exist to show.

Moving the `@import` into `usj-nodes.css`, or adding it to those two stories, would make the fix uniform.

src/renderer/styles/fonts.test.ts
```diff
@@ -0,0 +1,408 @@
...
+    expect(parseFontFaces(vendored)).toEqual([]);
+  });
+
+  it.each(VENDORED_EDITOR_STYLESHEETS)('hands the stack to the editor in %s', (path) => {
+    // The editor's generated project stylesheet appends `var(--usj-font-fallback, serif)` to every
+    // font it names. Pointing that at this stylesheet's stack is what makes a project whose own
+    // font is unavailable fall through the application's Scripture order rather than the editor
+    // library's default chain.
+    const vendored = readFileSync(resolve(repoRoot, path), 'utf8').replace(/\s+/g, ' ');
+    expect(vendored).toContain('--usj-font-fallback: var( --scripture-font-stack,');
```
lyonsil: **Low - this assertion is pinned to Prettier's current line-wrapping, not to the CSS.**

After `.replace(/\s+/g, ' ')`, `'--usj-font-fallback: var( --scripture-font-stack,'` only matches because Prettier currently breaks the `var()` across lines, leaving a space after `var(`. Shorten the literal fallback chain - or let Prettier fit the declaration on one line for any other reason - and the file emits `var(--scripture-font-stack,` with no space, so the test fails on correct code.

Matching `/--usj-font-fallback:\s*var\(\s*--scripture-font-stack\s*,/` against the unnormalized text would hold either way.

src/renderer/styles/fonts.css
```diff
@@ -1,7 +1,10 @@
 /* styles applied to core and WebViews to customize the fonts */
 
 /* #region styles downloaded from Google Fonts URL because CSP prevents outside stylesheets */
-/* @import url('https://fonts.googleapis.com/css?family=Gentium+Plus|Charis+SIL|Cardo|Noto+Sans+Syriac|Google+Sans&display=swap'); */
+
+/* Paste this URL's response verbatim; never hand-edit it. The fonts that need editing are in the
+   next region, which is why the families are split across two URLs. */
+/* @import url('https://fonts.googleapis.com/css2?family=Cardo:ital,wght@0,400;0,700;1,400&family=Google+Sans:wght@400&family=Noto+Sans+Syriac:wght@100..900&display=swap'); */
```
lyonsil, line 8: **Low - the "paste verbatim; never hand-edit" invariant is already untrue for this region, and nothing tests it.**

This URL names `Google+Sans:wght@400`, but the Google Sans block below is not what that URL returns today - the PR body says it has drifted upstream and was deliberately left alone. `fonts.test.ts`'s `leaves the verbatim region unedited` only asserts the absence of `local(`.

So the next person who follows this comment literally and pastes the response will silently change the application chrome font as a side effect of regenerating Cardo or Noto Sans Syriac, with no test to flag it. Either pin the region against the URL's response, or say here that Google Sans is intentionally frozen.

lib/platform-bible-react/src/components/demo/scripture-editor/usj-nodes.css
```diff
@@ -9,7 +9,7 @@
    - `.verse-selected` and the armed-verse blink bind to this repo's `--destructive` theme token
      instead of the editor library's literal reds.
    - The table-cell box border binds to `var(--foreground, #000000)` instead of the source's
-     literal black, so it stays visible against a dark app theme. */
+     literal black, so it stays visible against a dark app theme.   - The source's four `local()`-only Charis SIL `@font-face` rules are dropped. */
```
lyonsil: **Low - the new divergence bullet was appended to the previous line instead of starting its own.**

`... against a dark app theme.   - The source's four ``local()``-only Charis SIL ``@font-face`` rules are dropped. */`

The header is a bulleted divergence list, and this entry is now buried mid-line - which is exactly the list a future sync pass reads to know what not to re-add.

lib/platform-bible-react/src/components/demo/scripture-editor/scripture-editor.stories.css
tjcouch-sil: Confirmed and fixed, though not where you suggested. `footnote-item.stories.tsx:4` and `footnote-list.stories.tsx:6` do import `usj-nodes.css` directly and never load this stylesheet.

I put the faces in `.storybook/scripture-fonts.css`, imported from `.storybook/preview.ts`, rather than into `usj-nodes.css`. That file is a vendored mirror whose whole point in this PR is that it declares no faces, and a Google Fonts `@import` inside it would be a new divergence for the next sync pass to reason about. Loading from the preview also mirrors what the app does — it inlines `fonts.css` into every web view, not into one component's stylesheet — so any future story that renders Scripture text is covered without having to remember.

src/renderer/styles/fonts.test.ts
tjcouch-sil: Right, and thank you for spotting the mechanism rather than just the symptom — the assertion passed for a reason unrelated to what it pins. Using your regex against the unnormalized text now, with a comment saying why it is not normalized.

lib/platform-bible-react/src/components/demo/scripture-editor/usj-nodes.css
tjcouch-sil: Fixed — the bullet starts its own line again.

src/renderer/styles/fonts.css
tjcouch-sil: Fair, and the comment as written did invite exactly that. Rather than pin the region against a live fetch — which would make the suite fail on Google's schedule rather than on ours — the comment now says it outright:

> Google Sans here predates its current upstream build, so re-pasting changes the application chrome font as well — intended or not, notice it rather than ship it unseen.

Refreshing Google Sans is a real change to the app chrome and belongs in its own PR, not smuggled into a Scripture font fix.

extensions/src/platform-enhanced-resources/src/_usj-nodes.scss
tjcouch-sil: Both halves fixed. `--usj-font-fallback` is gone from this file — you're right that it had no reader here — and `fonts.test.ts`'s `hands the stack to the editor` now runs over only the stylesheets that actually render a generated project stylesheet, so it is no longer reporting assurance for a file where the mechanism does nothing. The header records both divergences.

The font rule itself I've left as `'Charis SIL', serif` for now, pending a decision from TJ — see the note on the `.usfm.formatted-font` finding. Changing it is the same question as that one, and I'd rather both move together or neither.

src/renderer/styles/fonts.css
tjcouch-sil: Correct on the mechanism, and it is worth having written down: `src` picks a face, and a glyph missing from the chosen face falls through to the next *family*, not back to the `url()` in the same face. So an older installed Charis SIL that lacks a codepoint inside a declared `unicode-range` now misses it where the downloadable subset would have supplied it.

Accepting it rather than fixing it, deliberately: preferring the installed copy over the download is the behaviour this design was asked for, and there is no way to get it without also getting per-face resolution. The trade runs the other way too — the installed SIL build is generally *newer* and broader than Google's subsetted v2, which is the whole reason the previous vendored rules were load-bearing.

You're also right that the measurements do not cover it; they were taken against a current installed copy. I've not claimed otherwise anywhere, but it is not evidence for this case and I won't present it as such.

src/renderer/styles/fonts.css
tjcouch-sil: Confirmed and fixed. I checked the ranges rather than taking it on trust, and it is worse than "no greek block" — Charis SIL's Google build has no Greek whatsoever:

| family | U+0370-03FF | U+1F00-1FFF | U+0300-036F |
| --- | --- | --- | --- |
| Charis SIL | 144/144 missing | 256/256 missing | 104/112 missing |
| Gentium Plus | 9/144 missing | covered | 104/112 missing |

And measured, per character, under Windows Edge with both fonts installed, same markup throughout: a bare α at this PR's previous head was **byte-identical to installed Gentium Plus** (105.078125 px, against 129.796875 for Charis). Exactly the mid-line typeface change you described.

The fix is a third region: per family and style, one `local()`-only face whose `unicode-range` is the **complement of the downloaded ranges over the BMP**. Disjoint by construction, so an erroring face there can never be a candidate for a codepoint a downloadable face could serve — which is the shape of the original bug, and the reason I did not simply restore the un-ranged rules. Where the font is not installed they fail and the character falls through exactly as it does today.

After: α measures 129.796875 — installed Charis SIL. Two things worth recording because they are not what I expected going in:

- **Polytonic Greek stays Gentium Plus.** `ἐ` and `ῇ` measure the same forced to Charis SIL as forced to generic `serif`, so the installed Charis SIL has no glyphs there; the face is selected and falls through. That is correct behaviour and Gentium Plus is the right place for it to land.
- **No regression for scripts Charis lacks.** The complement nominally covers Hebrew and Syriac. Measured before and after they are byte-identical (126.953125 and 186), still served by Cardo and Noto Sans Syriac — glyph-level fallback does the right thing.

On the test gap you identified: `fonts.test.ts` now asserts, per weight and style, that the new faces neither overlap the downloaded ranges nor leave a BMP codepoint uncovered. Worth saying that I first wrote that check per *family*, and a mutation narrowing one face's range passed it — the union across styles hid the gap. It is per face for that reason.

The combining-mark half I have not measured: combining marks are zero-advance, so width cannot distinguish them. It rests on the ranges and the test, not on a measurement, and I'd rather say so than imply otherwise.

extensions/src/platform-scripture-editor/src/_usj-nodes.scss
tjcouch-sil: Correct, and it was undocumented — thank you. Confirmed with TJ: intended and kept. The stack's order is the application's stated preference for Scripture text, so the editor default matching `.scripture-font` is the point of pointing it at the stack rather than a side effect. It is now called out as a behaviour change in the PR description instead of being left for a reader to infer.

Enhanced Resources now names the same stack (see the reply on that file), so all three stylesheets agree rather than two of them agreeing and one not.

On Storybook disagreeing with the app: that is deliberate and I would keep it. `--scripture-font-stack` lives in `fonts.css`, which is the application's, and `platform-bible-react` is a library that must render sensibly for consumers who never load it — so the `var()` fallback there is the editor library's own chain, and Charis SIL leading is that chain's answer, not a stale copy of ours.

src/renderer/styles/fonts.css
tjcouch-sil: Revisited — refreshed instead. The region is now what its URL returns today, so "never hand-edit it" is true of the whole region rather than only of the absence of `local()`, and the comment records the date:

> Pasted verbatim from the URL below, then left to Prettier and `stylelint --fix` (Google emits unquoted `url()`); never hand-edit it. Last refreshed 15 September 2026.

Only Google Sans moved; Cardo and Noto Sans Syriac are byte-identical, and the region as a whole now matches a fresh fetch face for face. Still not pinned against a live fetch, for the reason above — that would fail the suite on Google's schedule rather than on ours.

extensions/src/platform-enhanced-resources/src/_usj-nodes.scss
tjcouch-sil: Revised, and your finding held up better than my first fix of it. Rather than removing the property, this file's font block is now **identical** to `platform-scripture-editor`'s — which is what its header has always said to do, and which I had quietly diverged from. So `--usj-font-fallback` has a reader here after all, and all three vendored stylesheets are checked for it again rather than two.

The only divergence left is the `.formatted-font .usfm` selector rather than `.usfm.formatted-font`, which predates this work; the header records that and nothing else now.

src/renderer/styles/fonts.css
lyonsil: The three-region fix covers everything I raised — I checked the complement per face rather than
taking the test's word for it, and Greek, polytonic and the combining marks all land inside the new
faces. The note about writing the check per family first and having a narrowing mutation pass it is
the useful half of that; per face is right.

One thing the fix doesn't reach, and it's the same narrowing rather than a new issue: **the
complement stops at the BMP.**

`BMP_END = 0xffff` (`fonts.test.ts:261`), so the installed-only faces are the complement of the
downloaded ranges *over the BMP*. The rules this PR removes carried no `unicode-range` at all, so
wherever Charis SIL is installed they were also the only face serving anything above U+FFFF — and
nothing serves those now.

Against the Charis SIL installed on my machine (the Google Fonts build, `Charis_SIL_*.2.ttf`),
parsing the `cmap` subtables directly, all four faces cover the same 52 astral codepoints:

> U+10783, U+1078F, U+10791, U+10795, U+107A0, U+107A2-107A3, U+107B2 (Latin Extended-F), U+1088E,
> U+1D40C, U+1D504-1D519 partial (math Fraktur), U+1DF00-1DF1E (Latin Extended-G)

Latin Extended-F and G are the phonetic/orthographic blocks, which is the material this font exists
for, and the SIL build Paratext installs is broader than Google's rather than narrower. So a Charis
SIL project containing U+1DF00 matched the un-ranged `local()` face before this PR and matches no
Charis SIL face after it. Unlike the Greek case this one is tofu, not a typeface change — nothing
else in `--scripture-font-stack` has those codepoints. Derived from the declarations and the cmap,
not measured in a browser; I'd rather say so than imply otherwise.

The fix looks free of risk: the highest upper bound anywhere in region 2 is **U+FFFD**, checked
across every `unicode-range` for both families, so extending the complement to U+10FFFF adds no
overlap with a downloadable face and the disjointness half stays true by construction. Raise
`BMP_END` to `0x10ffff` alongside it so the test keeps pinning the new end — that makes the
exhaustive loop at :432 17× longer, so if the runtime shows, walking the parsed ranges instead of
every codepoint holds the same property.

lib/platform-bible-react/src/components/demo/scripture-editor/usj-nodes.css
```diff
@@ -34,31 +35,29 @@
...
-@font-face {
-  font-family: 'Charis SIL';
-  src: local('Charis SIL Bold Italic');
-  font-weight: bold;
-  font-style: italic;
+/* Don't declare Scripture fonts here. A `local()`-only face errors where the font is not installed,
+   and an erroring face makes the family unusable at that weight and style instead of deferring to a
+   working declaration — which drops bold and italic Scripture text onto a system font. The app
+   declares Charis SIL in src/renderer/styles/fonts.css; this demo pulls it from Google Fonts in
+   scripture-editor.stories.css, since Storybook does not load that stylesheet. */
```
lyonsil: **Low - this points at the wrong file for the Storybook font pull.**

The comment says the demo pulls Charis SIL from Google Fonts in `scripture-editor.stories.css`, but
that file carries no such `@import` any more — the faces went to `.storybook/scripture-fonts.css`,
loaded from `preview.ts:8`, deliberately so the footnote stories that import *this* file directly
get them too. That's the right place for them; it's the trail that's stale.

This comment is what a future sync pass reads to know why no faces are declared here, and it now
sends that reader to a file with nothing in it — from which the reasonable conclusion is that the
footnote stories have no font source at all.

lyonsil:
Not a code finding — the PR description is stale on the one file whose divergence it is explicitly
reasoning about, and the description is what a reviewer reads first.

The `--scripture-font-stack` bullet still says:

> the two vendored stylesheets that render a generated project stylesheet point `--usj-font-fallback`
> at it (Enhanced Resources consumes none, so it carries no such property and keeps its own chain)

At `7946cd6` that is no longer true, and a later bullet in the same list says the opposite —
"`platform-enhanced-resources`' font block is now identical to `platform-scripture-editor`'s".
The code matches the later bullet: `_usj-nodes.scss` declares `--usj-font-fallback` at line 38 and
consumes it at line 49, and `fonts.test.ts`'s `points Scripture text at the stack in %s` runs over
all three entries of `VENDORED_EDITOR_STYLESHEETS`, not two.

So it's three stylesheets pointing at the stack, not two — which is the better story anyway. Worth
fixing the parenthetical before anyone reviews the Enhanced Resources change against it and
concludes the diff is wrong.

src/renderer/styles/fonts.css
tjcouch-sil: Confirmed and fixed at `56ec58b`. I checked the bound rather than taking it on trust: parsing every `unicode-range` in both downloaded regions, the highest upper bound is **U+FFFD** for Charis SIL and for Gentium Plus (Cardo reaches U+1034A and Google Sans U+1FBFF, but the disjointness check is per family, so neither can collide). So extending the complement to U+10FFFF adds no overlap, and the only gap it closes is exactly `U+10000-10FFFF` — computed per weight and style, all eight faces, before and after:

| | before | after |
| --- | --- | --- |
| overlap with downloaded | none | none |
| uncovered over [0, 10FFFF] | U+10000-10FFFF | none |

The eight faces' final range goes `U+FFFE-FFFF` → `U+FFFE-10FFFF`, and the region comment now says "the whole of Unicode" and names Latin Extended-F and G as what is at stake.

On the test: rather than widen the loop 17×, coverage is compared as merged `[first, last]` runs instead of expanded into codepoints. A set per face would be a million entries and the arithmetic is exact either way, so the property is the same one, and the suite runs in 21 ms rather than paying for either version of the walk. `BMP_END` is gone; `UNICODE_END = 0x10ffff` is what the gap check closes against. Failure messages now read as ranges (`U+10000-10FFFF`) rather than the first four codepoints in hex, which is a good deal more legible for exactly this finding.

Two things I tightened while I was in that function, both because they bear on whether the guard means what it says:

- An **absent** `unicode-range` now reads as the whole of Unicode, which is what such a face actually serves. It previously covered nothing, so a face declared without one would have been reported as leaving everything uncovered.
- An **unreadable** range part now throws rather than being silently skipped. A wildcard form like `U+4??` parses under CSS and not under this regex; skipping it would make the guard's exactness quietly depend on which spellings it happens to understand.

Mutation-checked, all three against the new code: one face narrowed back to `U+FFFE-FFFF` reports `uncovered: ["U+10000-10FFFF"]`; a downloaded Charis SIL range extended to `U+1DF00-1DF1E` reports the overlap; `U+4??` throws with the file and the offending part named. Baseline 26/26 green, and the two vendored-stylesheet pin suites still pass unchanged (20/20 and 26/26).

One limit worth stating rather than implying: like your finding, this half rests on the declarations and the `cmap`, not on a browser measurement. Astral coverage is hard to measure the way the Greek case was — the fallback for a missing glyph is tofu, which has no distinguishing width to compare against.

lib/platform-bible-react/src/components/demo/scripture-editor/usj-nodes.css
tjcouch-sil: Confirmed and fixed at `56ec58b`. You're right on both halves: `scripture-editor.stories.css` carries no `@import` for the faces — its four imports are `usj-nodes.css`, `editor.css`, the commented-out `tailwind.css` and `editor-overrides.css` — and the faces are in `.storybook/scripture-fonts.css`, loaded from `preview.ts:8`, which is where they belong for the reason you give.

The comment now names that file and says *why* it lives in the preview rather than in a story's stylesheet:

> The app declares Charis SIL in src/renderer/styles/fonts.css; Storybook does not load that stylesheet, so it pulls the same faces from Google Fonts in .storybook/scripture-fonts.css, loaded from the preview rather than from one story's stylesheet so the stories that import this file directly get them too.

The trail went stale in the same commit that moved the faces — the move was the response to your earlier Storybook finding, and I updated the file the faces left but not the comment pointing at it.

I swept the rest of the tree for the same stale pointer rather than fixing only the line you cited: `scripture-editor.stories.css` is now named in exactly one place, `scripture-editor.stories.tsx:24`, which is a real import. The two `.scss` copies carry the same "don't declare Scripture fonts here" comment but no Storybook sentence — they're app-side and do load `fonts.css` — so they needed nothing.

tjcouch-sil:
Fixed — thanks, and you're right that it's the worse kind of stale, because the parenthetical contradicted a later bullet in the same list and a reviewer reading the Enhanced Resources diff against it would have concluded the diff was wrong.

The sentence was true when written and stopped being true one commit later: dropping `--usj-font-fallback` from Enhanced Resources was the response to your dead-property finding, and the commit after that put it back by making its font block identical to `platform-scripture-editor`'s, which its own header says to keep it in sync with. I updated the bullet that explains the sync and left the one that reasons about it behind.

It now reads:

> **`--scripture-font-stack`**: the Scripture stack is now named as well as applied, and all three vendored stylesheets point `--usj-font-fallback` at it.

which matches the code (`_usj-nodes.scss:38` declares it, `:49` consumes it, in both extensions) and matches `points Scripture text at the stack in %s` running over all three entries of `VENDORED_EDITOR_STYLESHEETS`.

I also refreshed the third-region bullet in the same pass, since the astral fix above supersedes its "over the BMP" wording, and added the three new mutations to the falsifiability list under Testing.

lib/platform-bible-react/.storybook/preview.ts
```diff
@@ -5,6 +5,7 @@ import { persistDirection, readDirection } from '../src/utils/dir-helper.util';
 import { DocsPageWithFilePath } from './blocks/DocsPageWithFilePath';
 import { withPlatformBibleThemes } from './theme-decorator';
 import '../src/index.css';
+import './scripture-fonts.css';
```
lyonsil, line 8: **Low - this covers `platform-bible-react`'s Storybook; the root Storybook renders the same stories with no Charis SIL declaration at all.**

The reasoning for putting the faces here rather than in `usj-nodes.css` was that loading from the preview covers "any future story that renders Scripture text ... without having to remember." That holds for this Storybook. There are two, and the other one is the bigger consumer of the vendored stylesheet.

Root `.storybook/main.ts:8-10` globs `../src/**` *and* `../extensions/src/**`, and five story files under those globs reach the demo `usj-nodes.css`:

- `extensions/src/platform-scripture-editor/src/model-text-panel.stories.tsx:26`
- `extensions/src/platform-enhanced-resources/src/web-views/enhanced-resource.web-view.stories.tsx:15`
- `extensions/src/platform-enhanced-resources/src/components/scripture-pane/scripture-pane.stories.tsx:18`
- `src/stories/platform/ten-power-layout.stories.tsx` and `ten-simple-layout.stories.tsx`, both via `ten-layout-shared.tsx:69`

Root `.storybook/preview.ts` loads `index.css` and `storybook-themes.css` and nothing else; `fonts.css` is imported only as `?raw` into `src/renderer/index.tsx:32` and `web-view.service-shard.ts:31`, so neither Storybook receives it. With the four vendored faces removed and `scripture-fonts.css` imported only here, those five root-Storybook stories now have no Charis SIL declaration from any source - so on a machine or CI container without the font installed they render Times New Roman / DejaVu Serif with synthesized bold and italic, which is the bug this PR exists to fix, still live in one of the two Storybooks.

The comment at `usj-nodes.css:41-43` is true of only one of them for the same reason.

Adding `import '../lib/platform-bible-react/.storybook/scripture-fonts.css';` to the root `.storybook/preview.ts` would make it uniform.

lib/platform-bible-react/.storybook/scripture-fonts.css
```diff
@@ -0,0 +1,9 @@
+/*
+ * Charis SIL with real bold and italic faces, for every story that renders Scripture text.
+ *
+ * The application inlines these from paranext-core's src/renderer/styles/fonts.css into each web
+ * view; Storybook does not load that stylesheet, and the vendored `usj-nodes.css` deliberately
+ * declares no faces of its own. Loaded here rather than from a single story's stylesheet so that
+ * the footnote stories, which import `usj-nodes.css` directly, get the same faces.
+ */
+@import 'https://fonts.googleapis.com/css2?family=Charis+SIL:ital,wght@0,400;0,700;1,400;1,700&display=swap';
```
lyonsil: **Low - Storybook keeps the coverage gap that the third region closes for the app.**

These are Google's five Charis SIL subsets - latin, latin-ext, vietnamese, cyrillic, cyrillic-ext. The four `local()`-only rules this PR removes carried no `unicode-range`, so wherever the font is installed they were the only face serving everything outside those subsets. That is the same gap measured in the `fonts.css:587` thread, and the installed-only complement region fixes it in `fonts.css` only - there is no Storybook counterpart.

Concretely, on a developer machine with Charis SIL installed: a bare α (U+03B1) in a Scripture story was served by the installed face at base, and now falls out of the family. The chain it falls into is this repo's vendored one - `'Charis SIL', 'Times New Roman', 'Liberation Serif', 'DejaVu Serif', Georgia, serif` (`usj-nodes.css:51-59`) - which unlike `--scripture-font-stack` has no Gentium Plus in it, so it lands on Times New Roman rather than on the next Scripture font. Same mid-line typeface break, one family further along.

This is a different axis from the divergence deliberately kept in the `_usj-nodes.scss:51` thread. That decision is about which family *leads* the stack where `--scripture-font-stack` is undefined, and the library-renders-sensibly-for-its-own-consumers argument carries it. This one is a regression against base within Storybook: coverage that was there before the vendored rules were removed, gone after. Mirroring the eight installed-only complement faces into this file would restore it.

Secondary, and separable: the `@import` makes the Storybook preview depend on a live `fonts.googleapis.com` fetch, which `@storybook/addon-vitest` also loads when running stories as tests, so an offline or network-restricted run silently loses the faces rather than failing. The app's equivalent is inlined from disk.

lyonsil (commented):
Two follow-up findings on the Storybook half of the font work at `56ec58b`, both low. Neither is a regression in the application; both are about `@font-face` coverage inside Storybook, which no test covers.

Each comment states the severity it settled at. Both were checked against the code rather than inferred from the diff.

lyonsil (approved):
:lgtm: - I'm marking the latest 2 items reported as optional since they only impact Storybook. Your choice on whether you want to worry about them now.

> @lyonsil reviewed 8 files and all commit messages, made 1 comment, and resolved 11 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2814#-:-P1cH5d-3pKOgf2uam-v:b-lp4895)* status: all files reviewed, 2 unresolved discussions (waiting on tjcouch-sil).

lib/platform-bible-react/.storybook/preview.ts
tjcouch-sil: Fixed — the root `.storybook/preview.ts` now loads it too, and `fonts.test.ts` fails if either preview stops.

Your list of five checks out; I re-derived it with a grep for every reference to `usj-nodes` across `src`, `extensions/src` and `lib` rather than reading the diff, and it comes back the same. `ten-layout-shared.tsx:69` is the one worth naming, since it is neither of the two story files it serves.

One thing I did while I was in there, because a second copy of the faces is what your other finding is about: `scripture-fonts.css` is no longer an `@import` of `fonts.googleapis.com`. It carries the faces from `fonts.css` verbatim, and `fonts.test.ts` pins the two against each other face for face — same weights, styles, `src` and `unicode-range`, in both directions. Mutation-checked: narrowing, deleting or renaming a `local()` in Storybook's copy fails, and so does widening or adding a Charis SIL face in `fonts.css` alone.

I did consider having the root Storybook import the application's `fonts.css` outright instead, which would have no copy at all and would bring `--scripture-font-stack` with it. I left it: `fonts.css` also sets `body, .pr-twp { font-family: 'Google Sans', … }`, so it would restyle the chrome of every root story, which is a bigger and unasked-for change than the one you found. The pin is what keeps the copy honest instead.

`fd6c713`.

lib/platform-bible-react/.storybook/scripture-fonts.css
tjcouch-sil: Fixed, both halves.

`scripture-fonts.css` now carries Charis SIL's four installed-only complement faces alongside the five downloaded subsets, so the alpha case is back where it was at base — and the secondary point is closed with it: the `@import` is gone and the downloaded faces are inlined from `fonts.css`, `local()`-first, exactly as the application has them. A network-restricted run keeps the faces rather than silently losing them, which matters most for `@storybook/addon-vitest`, where the loss would be invisible.

Charis SIL only, deliberately. Gentium Plus is in `--scripture-font-stack` but not in the vendored chain, which is what Storybook resolves through, so faces for it would be declared and never named. If the vendored chain ever gains it, the pin below is where that shows up.

The copy is guarded rather than trusted: `fonts.test.ts` compares the Charis SIL faces in `scripture-fonts.css` against `fonts.css`'s downloaded-plus-complement set face for face, and fails in either direction. What it deliberately does not cover: a Gentium Plus change (not in Storybook's chain), and any non-`@font-face` rule someone adds to the file.

`fd6c713`.

lib/platform-bible-react/.storybook/preview.ts
tjcouch-sil: Verified in the built Storybooks rather than left to the test, since CI builds neither (`publish-docs` on main, and `chromatic` only behind a label) — so a Storybook-only break lands on main unseen.

Both Storybooks build clean. Serving the root build and opening two of the five stories you named — `…modeltextpanel--active` and `…scripturepane--with-scripture` — over Playwright, on this box, where neither Charis SIL nor Gentium Plus is installed:

- **24 Charis SIL `@font-face` rules** reachable from the story document (20 downloaded subsets + the 4 installed-only complement faces). Before this commit those stories saw none.
- The Scripture element computes to `"Charis SIL", "Times New Roman", "Liberation Serif", "DejaVu Serif", Georgia, serif` — the library chain, as intended: `--scripture-font-stack` lives in `fonts.css`, which Storybook does not load.
- Probe-string widths at 100px, four styles, against an unresolvable family name as the control:

  | | regular | bold | italic | bold-italic |
  | --- | --- | --- | --- | --- |
  | `'Charis SIL', serif` | 1290 | 1393 | 1193 | 1303 |
  | absent family → `serif` | 1493.31 | 1652.50 | 1493.31 | 1652.50 |

  Four distinct widths for Charis SIL is four real faces. The control is the contrast: its italic is byte-identical to its regular, which is what a synthesized oblique looks like.
- The complement faces report `status: error` here, as they must with the font not installed — and the widths above are measured with them in the document, which is the disjointness holding: an erroring complement face is never a candidate for a codepoint a downloadable face serves.

`platform-bible-react`'s Storybook builds clean too, with the same 24 faces in its bundle.

lyonsil (commented):
One low finding on the Storybook font work at `fd6c713`. The rest of that commit checks out: I parsed the Charis SIL faces in `scripture-fonts.css` against `fonts.css` face for face (24 each, no difference in either direction) and reproduced each of the new guards' mutations.

lib/platform-bible-react/src/components/demo/scripture-editor/usj-nodes.css
```diff
@@ -34,31 +35,34 @@
...
-  src: local('Charis SIL Bold Italic');
-  font-weight: bold;
-  font-style: italic;
+/* Don't declare Scripture fonts here. A `local()`-only face errors where the font is not installed,
+   and an erroring face makes the family unusable at that weight and style instead of deferring to a
+   working declaration — which drops bold and italic Scripture text onto a system font. The app
+   declares Charis SIL in src/renderer/styles/fonts.css; Storybook does not load that stylesheet, so
+   it pulls the same faces from Google Fonts in .storybook/scripture-fonts.css, loaded from the
+   preview rather than from one story's stylesheet so the stories that import this file directly get
+   them too. */
```
lyonsil: **Low - this comment now describes the mechanism the commit removed.**

> ... so it pulls the same faces from Google Fonts in `.storybook/scripture-fonts.css`, loaded from the preview ...

Neither half holds at this head. `scripture-fonts.css` no longer `@import`s `fonts.googleapis.com` - `fd6c713` replaced the import with the faces copied from `fonts.css`, `local()`-first, precisely so a network-restricted run keeps them instead of silently losing them. And it is loaded from *both* previews now, `.storybook/preview.ts:14` as well as `lib/platform-bible-react/.storybook/preview.ts:8`.

This is the only place in the tree that tells a future sync pass why no faces are declared here, and it now describes the design the commit deliberately abandoned - from which the reasonable conclusion is that Storybook's faces depend on a live Google fetch, which is the thing that was just fixed.

It is also the line named at the end of the root-Storybook finding ("The comment at `usj-nodes.css:41-43` is true of only one of them for the same reason") - the code was fixed and the comment was left, the same way the pointer went stale in the commit that moved the faces.

Suggested: "... Storybook does not load that stylesheet, so `.storybook/scripture-fonts.css` carries a copy of the same faces, loaded from both Storybooks' previews rather than from one story's stylesheet so the stories that import this file directly get them too. `fonts.test.ts` pins that copy against `fonts.css`."

lyonsil (approved):
> @lyonsil reviewed 6 files and all commit messages, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2814)* status: all files reviewed, 1 unresolved discussion (waiting on tjcouch-sil).

lib/platform-bible-react/src/components/demo/scripture-editor/usj-nodes.css
tjcouch-sil: Fixed, and the drift is now checkable instead of left to the next reader.

Your wording, with two additions — that the copy is *copied rather than fetched* (the property that went stale is exactly the one worth stating), and the reason it is both previews rather than one:

```
The app declares Charis SIL in src/renderer/styles/fonts.css; neither Storybook loads that
stylesheet, so .storybook/scripture-fonts.css carries a copy of the same faces — copied rather
than fetched, so a network-restricted run keeps them. It is loaded from BOTH Storybooks'
previews rather than from one story's stylesheet, because the stories that reach this file
arrive by several routes and through both Storybooks. fonts.test.ts pins that copy against
fonts.css face for face, and fails if either preview stops loading it.
```

You are right that this is the second time in this PR a comment outlived the code under it, so I looked for what would have caught it. The face-for-face pin could not: an `@import` adds no `@font-face` of its own, so the copy would still have matched face for face while the faces went back behind a live request — one `@storybook/addon-vitest` makes silently and without failing. So `fonts.test.ts` now also fails if that file acquires an `@import` at all. Mutation-checked by appending one.

Checked with comments stripped, for the same reason this thread exists in miniature: the file's own prose explains why it is *not* an `@import`, and on a first pass the explanation failed the guard on the code. (The same trap bit the stylesheet pins in paranext/scripture-editors#12, where a comment naming `@font-face` broke the assertion that the file declares none.)

Swept the rest rather than assuming: `grep` for `scripture-fonts`, `Google Fonts`, `fonts.googleapis` and `Storybook` across `src`, `extensions/src`, `lib` and `.storybook` turns up no other site describing the old mechanism. The two `_usj-nodes.scss` copies stop their comment at "…onto a system font" and never mention Storybook, which is correct — the root Storybook's stories reach the demo `usj-nodes.css`, not the extensions' SCSS.

What that sweep would still miss: a comment describing the mechanism without using any of those four words.

`33db1f0`.

lyonsil (approved):
:lgtm:

> @lyonsil reviewed 2 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2814#-:-P1fTOnkCHlEcFvCFnqy:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2815, opened 2026-09-14
fix(erb): surface the DLL build chain's real error on npm start

tjcouch-sil:
## Summary

`npm start` in a tree without a built renderer DLL can fail with an error that names no cause — three nested `Command failed` wrappers and two `<Buffer ...>` dumps, and nothing that says what actually broke. This makes that failure self-explaining.

### Why review this

Not part of the current epic. This came out of a live `npm start` failure in a worktree: the visible output was pure wrapper noise, and finding the cause meant reading `.erb` sources and re-running the inner command by hand. Cheap to review (one line + a comment), and it turns a recurring "why did my dev build die" detour into a message that reads itself.

## Changes

- `.erb/configs/webpack.config.renderer.dev.ts`: pass `stdio: 'inherit'` to the `execSync('npm run postinstall')` that fires when the DLL bundle is missing.

### Background

When the renderer dev config sees no DLL, it self-heals by running `npm run postinstall`, which runs the `postinstall:build` chain (`patch-package && tsx check-native-dep.js && electron-builder install-app-deps && npm run build:dll && npm run build -w eslint-plugin-paranext`).

That `execSync` used default options. `execSync` forwards stderr to the parent by default but **captures stdout** — and the actionable messages in this chain are on stdout. `check-native-dep.js` prints its entire native-dependency report with `console.log` and then `process.exit(1)`; electron-builder and the DLL build likewise report to stdout. So whenever the chain failed, the one thing the developer needed was in the thrown error's `stdout` buffer, unread, while the terminal showed only the wrappers.

`stdio: 'inherit'` matches what `.erb/scripts/postinstall.ts` already does one level down, and additionally removes a latent `maxBuffer` overflow risk from piping a full DLL build through the default 1 MB buffer. The `execSync` return value was unused, so nothing else changes.

## AI Involvement

AI-assisted — [session 1](https://claude.ai/code/session_01SSoTtSUc12WQGc73NhBefJ). Claude diagnosed the swallowed-output mechanism from a failing `npm start`, wrote the one-line change and its comment, and ran the checks below. Reviewed by me before commit.

## Testing

- [x] `npm run typecheck:erb` — pass (covers the changed file)
- [x] `npm run typecheck:core`, `npm run typecheck:e2e` — pass
- [x] `npm run lint` — pass (exit 0)
- [x] Mechanism verified directly: with default options a failing child's stdout is captured into `err.stdout`; with `stdio: 'inherit'` it prints.
- [ ] `npm run typecheck:workspaces` and `npm test` not run locally — this worktree symlinks `node_modules` from a sibling checkout, which makes TypeScript resolve each extension's `.d.ts` twice and reports spurious duplicate-identifier errors. Left to CI, which runs against a real install. The change is a webpack dev-config option and cannot affect either suite.

## Risk Level

Low — one option on one `execSync` whose return value is unused, on a dev-only code path that only runs when the DLL bundle is absent. Strictly more output than before; no behavior change on success.

https://claude.ai/code/session_01SSoTtSUc12WQGc73NhBefJ

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2815)


katherinejensen00 (approved):
:lgtm:

This looks helpful. Thanks, TJ!

> @katherinejensen00 reviewed 1 file and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2815#-:-P1X687J9_zLTJmiga84:bn111h9)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #2819, opened 2026-09-15
PT-4578: Zoom in / out / reset in the tab menu, Power and Simple

rolfheij-sil:
> **Stacked on #2803** (`pt-4576-content-zoom-platform-core`). Review only the commits above that branch. Stacked PRs get no automatic CI here; the full local battery (typecheck, lint, format:check, `vitest run src`: 264 files / 3838 tests; platform-bible-utils 611 tests) is green. After #2803 squash-merges this branch is rebased with `git rebase --onto main <old parent tip>`.

## Summary

Work item 3 of the per-pane content-zoom epic (PT-4575): **Zoom in / Zoom out / Reset zoom to default** in the web-view tab's right-click menu, in Power mode and in Simple mode.

- **Power mode, data only.** A new `platform.tabZoom` group (order 50, not extensible, `isExperimental`) with three items bound to the owner-routed commands from #2803, ahead of the existing window group. The tab-menu handler already passes the tab id, which is the web-view id, so each item acts on exactly that tab. en/es strings added.
- **Simple mode gets a tab menu for the first time**, holding only the zoom group. The tab title component now reads the contributed menu in both modes and narrows it to `platform.tabZoom` in Simple mode before conversion (the converter flattens groups, so the filter has to happen on the raw contributed menu). It renders nothing until the interface mode is known, so a Power user is never shown the Simple menu during startup. The drag-ignore marker is passed onto the menu trigger so Simple mode's fixed layout stays fixed; the keyboard path (Shift+F10 / Menu key) now works wherever a menu exists.
- **A tab hosting no web view** (a dialog or error tab) gets no zoom items, following the existing rule for the move items. On a web view that marks no zoom area the items stay and are inert: the service already treats such a request as a silent no-op, and the only "has areas" signal is filled asynchronously as the pane bootstraps, so disabling on it would show disabled items that in fact work.

PT-3171 (View menu on tabs) is fulfilled by this item; the closure note belongs with PT-4585.

### Why review this

Part of the current epic (PT-4575). It is the discoverable entry point for content zoom (NN-7) and the first UI Simple mode offers on a tab (NN-9).

## Decisions worth a look

- **Simple mode reaches the tab menu only on Column 3.** Simple mode's Home and Scripture Editor columns use a headless tab bar (`dock-layout-wrapper.simple-mode.scss`: pointer events off, opacity 0), so there is no editor tab to right-click. Decision (Epic Lead via Rolf, 2026-09-15): Simple mode needs editor zoom too, and the entry point is the Scripture editor's own hamburger menu (Options column), visible in both modes; that lands with PT-4581 alongside the editor's zoom areas. Recorded in the new ADR `adr-simple-mode-tab-menu-offers-zoom-only`. The Home column (model text) has no menu of any kind and keeps only the chords; a separate UX question.
- **`isExperimental` on the group only.** The menu-item schema has no such property and rejects unknown ones (`unevaluatedProperties: false`), so an item-level marker would fail startup validation; the existing `platform.tabWindow` group carries it the same way.
- **Order 50 is now reserved.** `WebViewMenu.tabMenu`'s TSDoc told extensions the platform occupies order 100; it now says 50 and 100, in the source and in the regenerated `platform-bible-utils/dist` (that rebuild also carries a formatting-only rewrap of one unrelated TSDoc paragraph).
- **Simple mode now pays one `getWebViewMenu` read per tab at mount** (six tabs in the fixed layout), where it paid none; a single read per tab, not a subscription.
- **Spanish** follows the ticket ("Acercar" / "Alejar" / "Restablecer zoom al valor predeterminado"); a reviewer may prefer "Restablecer el zoom …" to match the neighbouring strings' articles.
- **One pre-existing test updated:** `menu-data.service-host.test.tsx` hard-coded the full list of shipped tab-menu commands and now expects the three zoom commands ahead of the window ones.

## Testing

- New `menu.data.test.ts` (group shape and order, item commands and orders pinned to `CONTENT_ZOOM_COMMANDS`, labels, no item-level `isExperimental`, shipped document still validates).
- `tab-menu.util.test.ts`: `filterTabMenuToGroup` (happy path, absent group, cross-group items dropped, input not mutated, cross-check against the shipped JSON) and the no-web-view availability rule.
- New `platform-tab-title.zoom-menu.test.tsx` (11 cases: Power-mode order, Simple-mode contents, dispatch with the tab id in both modes, mode-not-known gate, Simple mode reads the menu once and never reads the window list, no-web-view tab in both modes, drag-ignore on title and trigger, nothing before the read lands, keyboard forwarding in Simple mode). Two existing tab-title test files gained the interface-mode mock and one rewritten expectation.
- Root and platform-bible-utils batteries green (see the stacked note); roborev per-commit reviews triaged on every commit (one fixed: a type-assertion suppression replaced by a `ReferencedItem` parameter type).
- Self-review: `/review-paratext` (four analyzer passes) and an OpenCodeReview delegate pass ran on the branch. No Critical. Fixed here: one interface-mode subscription per tab title instead of two; the keyboard-shortcuts catalog entry for the tab menu now covers both modes; translator notes say the items act on the pane last used in the tab, not the whole tab; backward-facing comments reworded; `CONTENT_ZOOM_TAB_MENU_GROUP` moved to the shared content-zoom model with the JSON pin test next to the menu data; an `act()` warning and a redundant test removed; Spanish reset label takes the article. Items stay inert on a view with no zoom area (the "has areas" signal is asynchronous, so disabling would show disabled items that work); the three tab-title test files still each carry their own mock harness.
- No e2e added: PT-4585 runs the Simple/Power matrix. Manual check in the running app pending (Power editor tab, dialog tab, Simple Column-3 tab, keyboard).

## Risk Level

Medium. Changes the tab title component every tab renders, in both modes; the Simple-mode behaviour is new.

AI-assisted — Claude Code (Fable 5.1 coordinating; Opus design pass; per-task Sonnet agents under TDD; Sonnet verification pass).

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2819)


rolfheij-sil:

PR #2817, opened 2026-09-15
PT-4580: ContentZoomRoot opt-in component, author docs, ADRs

rolfheij-sil:
> **Stacked on #2803** (`pt-4576-content-zoom-platform-core`). Review only the commits above that branch. Stacked PRs get no automatic CI here; the full local battery (typecheck, lint, format:check, `vitest run src`: 262 files / 3812 tests; platform-bible-react unit: 70 files / 915 tests; library build incl. typedoc) is green. After #2803 squash-merges this branch is rebased with `git rebase --onto main <old parent tip>`.

## Summary

Work item 5 of the per-pane content-zoom epic (PT-4575): the **component and documentation half** of the opt-in contract whose platform half shipped in #2803.

- **`ContentZoomRoot`** (platform-bible-react, `@experimental`): a `forwardRef` `div` that carries `data-platform-content-zoom-root`. No `area` prop marks the view's `main` area (empty attribute value, by contract); `area="footnotes"` names a second independently zoomable pane. It adds no classes and forwards every div prop; the marker attribute is rendered after the props spread so `area` is always the source of truth. `CONTENT_ZOOM_ROOT_ATTRIBUTE` is exported beside it, duplicated on purpose (the library cannot import core's `src/shared`, and `@papi/core` publishes the constant type-only); a core-side test imports both constants and pins them equal. `dist/` is rebuilt and committed.
- **Extension-author docs**: a "Content Zoom (experimental)" section in the Extension-Development-Guide and two sections in Component-Builder-Patterns — the one-attribute opt-in, one element per area and no nesting, plain-HTML views, the zoomed-`getBoundingClientRect` vs unzoomed-`fontSize` caveat and the `--platform-content-zoom-<area>` variable, the capture-phase rule for views that own Ctrl+wheel, the whole-view fallback for unmarked views, and the Interface Scaling vs content zoom vocabulary ("pane", not "tab").
- **Three ADR entries** in Architecture-Decisions.md at their byte-order slug positions: `adr-per-web-view-state-lives-in-the-definition` (TJ's rule, promoted into Paranext-Core-Patterns as a decision block), `adr-web-view-content-zoom-in-iframe-shortcuts`, `adr-zoom-composition`.
- **Paratext 9 feature inventory**: a new "1.18 Pane Zoom (View > Zoom)" entry citing the PT9 sources by file + symbol, and the misleading "Font size and zoom controls" bullet under Language Settings corrected to point at the project font size.

[...]

PR #2820, opened 2026-09-15
PT-4581: Scripture editor content zoom, two areas, editor menu, e2e

rolfheij-sil:
> **Stacked on #2817** (`pt-4580-content-zoom-root`, itself on #2803). Review only the commits above that branch. Stacked PRs get no automatic CI here; the full local battery (typecheck, lint, format:check, `vitest run src`, the extension's unit tests and typecheck, `build:extensions`, and the `scripture-editor` isolated e2e subset) is green — see Testing. After #2817 squash-merges this branch is rebased with `git rebase --onto <new base> <old parent tip>`.

## Summary

Work item 6 of the per-pane content-zoom epic (PT-4575): the **Scripture editor becomes the first view that zooms**, with two independent areas and a menu entry point that works in Simple mode.

- **Two zoom areas.** The rendered Scripture text (inside the editor's reverse portal, so it is the same element with or without the footnotes pane) is the `main` area; the footnotes list is the `footnotes` area; the footnote-editing popover follows the text's level. Toolbar, divider and panel padding stay outside every area, so they keep their size. Editable and read-only editors share the web-view type, so one change covers both, and both remember levels under `editor:<projectId>:<area>`.
- **Zoom in / Zoom out / Reset zoom to default in the editor's own hamburger menu** (Options column), visible in both modes. Decision (Epic Lead via Rolf, 2026-09-15): Simple mode has no tab bar on the editor column, so the tab menu from #2819 cannot reach it; the editor's menu already dispatches commands with the web-view id, so this is data-only in the extension's `menus.json` plus en/es strings.
- **Two platform fixes the first marked view exposed** (both in the renderer, both red-first):
  - A click in a zoom area now keeps it the pane's active area even when the click itself moves focus elsewhere. Selecting a footnote row puts the caret back in the text, and the resulting `focusin` (about 14 ms later) used to overwrite the active area, so a reset with no area id targeted the text. The bootstrap treats a click and the focus change it causes as one gesture (a `focusin` naming a different area within 200 ms of a pointerdown is ignored); a focus change with no pointer behind it, such as the caret reaching a footnote by keyboard, still sets the area.
  - A pane whose replacement content runs no bootstrap now drops its stale area list at the end of the fallback grace (a liveness probe on the bootstrap's `window.__platformContentZoom`) and gets the whole-view fallback, instead of keeping the previous content's areas forever. The list is still kept at load time on purpose: the fresh bootstrap reports from `DOMContentLoaded`, before the iframe's `load` event, so clearing at load would race that report. That is also why the "re-reported since this load" flag the ticket suggested was rejected.

### Why review this

Part of the current epic (PT-4575). This is the PRD's headline behaviour (NN-1…NN-4, NN-6, NN-8, NN-13/NTH-3 two-area delivery, NN-14 by construction) and the first proof of the opt-in contract from #2817.

## Decisions worth a look

- **Marker placement.** The `ContentZoomRoot` goes inside `PortalContents`, not around it: `react-reverse-portal` clones its direct child with the portal's props, which `PortalContents` exists to absorb; a props-spreading `div` there would forward portal props onto DOM attributes. `editorContainerRef` is untouched, because with the footnotes pane shown it also holds the footnotes panel and a marker there would nest the `footnotes` marker (nested markers are ignored).
- **Keyboard-chord e2e cases are written but gated** (`content-zoom-chords.spec.ts`, `test.skip` with the reasons in the file). Two independent reasons: main still claims Ctrl+`+`/`-`/`0` on Windows/Linux until #2809 merges, and Playwright drives keys through CDP, which never passes through `before-input-event`, so an ungated test would go green on this branch while the product path is untested. The same paths are covered through Ctrl+wheel and the commands in `content-zoom.spec.ts`.
- **E2E path.** The ticket named `e2e-tests/tests/web-view-content-zoom.spec.ts`, which no Playwright project collects; the specs live in `tests/isolated/scripture-editor/` (one test per file, the isolated fixture, `seedSettings` for the non-default-level case).
- **Observer cost.** The bootstrap rescans for markers on every mutation batch; the editor is the first view where the scan returns matches. Measured once with a throwaway Playwright probe (two `MutationObserver`s with the bootstrap's options inside the editor frame, one timing the real selector and one a zero-match selector; not kept in the suite because its only assertions could pass vacuously): under a 5 s typing burst plus a 3 s trailing buffer for the debounced save, the observer fires zero or one batch total (Lexical's per-keystroke updates are text-level, which `childList`/`attributes` observation never sees), and the scan over ~230–300 nodes with 0 vs 2 matches is below `performance.now()` resolution. No measurable marginal cost; nothing to file.
- **No cross-view sync of its own.** The editor only marks DOM; level storage, hidden-pane behaviour and re-application live in the platform (#2803), so the hidden-views rule needs no new decision here.
- **Cut line (NN-13).** If the second area were ever dropped, the footnotes list would be marked `main` so it zooms with the text. This PR delivers the two-area version.
- **Known follow-ups, not fixed here:** the footnotes pane's text does not yet take the project font size (PT-4189's territory; content zoom multiplies whatever size it has). The Simple-mode Home column (model text) has no menu of any kind, so it keeps only the chords; a separate UX question. The indicator anchors to the union of all `main`-marked elements, so while the footnote popover is open the badge may sit at the popover's corner.

## Testing

- Extension unit tests: a jsdom render test for the footnotes layout (both pane positions: one `footnotes` marker, divider and editor child outside it) and a source-contract test pinning the two `main` markers and the untouched scroll container (the 4 000-line web view cannot mount in jsdom); a JSON test for the new menu group and strings. Full extension suite green.
- Renderer unit tests: three new bootstrap cases (click-then-focus gesture, keyboard focus, gesture window expiry) and three service cases (stale areas dropped when the bootstrap is gone, kept on a real reload, a report inside the grace still cancels it). Bootstrap + service: 100 tests (the gesture rule alone has six cases: click-then-focus, keyboard focus, window expiry, one suppression per click, same-area focus does not spend it, a hop through an unmarked element does not spend it).
- E2E, headless, one spec per invocation: `content-zoom.spec.ts` (TC85 "Exercise Zoom" and TC-TO "Zoom menu increments" cases: wheel over text and footnotes independently, toolbar height fixed, indicator text and `data-area`, command path, per-area reset via a real click, read-only editor shares the memory identity, memory keys) 1 passed; `content-zoom-default-level.spec.ts` (seeded 120 % default; reset lands on it, not 100 %) 1 passed. The full `scripture-editor` isolated subset: 8 passed, 2 skipped (the gated chord file), and 2 pre-existing specs failing on a first-run tour dialog that also fail on the base branch under the same invocation (control run: `formatted-default-simple-mode` and `paragraph-style-trigger-column-floor` fail identically on `pt-4580-content-zoom-root` with the same invocation).
- roborev per-commit reviews triaged on every commit (one refuted: the e2e spec it said was missing landed two commits later on the same branch; two on a throwaway measurement spec, resolved by removing it).
- Self-review: `/review-paratext` (four analyzer passes) and an OpenCodeReview delegate pass ran on the branch. No Critical. Fixed here: the click-gesture window is armed only by a click inside an accepted area and is spent only when it suppresses a focus change into a different area (a first attempt that spent it on any focus change broke the real click sequence, where focus first settles on the footnote row itself; caught by the e2e and corrected red-first); the e2e helpers are shared through `scripture-editor-helpers.ts` instead of copied into three specs; the source-contract test pins structure, not class lists; spec comments are forward-facing; the extension menu test and the translator notes were tidied ("pane", not "area"). Kept as-is with a reason: the Spanish reset label matches #2819's tab-menu string for the same action (a shorter form would read as "reset the default"). Open at epic level, not here: whether the zoom epic gets a Storybook page (the UI-review guideline asks every UI work item for one; nothing in this diff is a component).
- Manual check in the running app pending: caret accuracy at 150 %, the footnote popover anchoring at 150 % (fallback if it drifts: wrap only the editing surface), the Simple-mode character-marker bar at a non-100 % level, and the spinner / book-not-available / empty-chapter views inside the new wrapper.

## Risk Level

Medium. First user-visible zoom; touches the editor's render tree and two platform paths (active-area tracking, fallback grace) that every marked view will share.

AI-assisted — Claude Code (Fable 5.1 coordinating; Opus design and investigation passes; per-task Sonnet agents under TDD; Sonnet verification pass).

- - -
This change is [Reviewable](https://reviewable.io/reviews/paranext/paranext-core/2820)


rolfheij-sil:

